PR #4906, opened 2026-07-03
fix(#4905): decode inbound Bolt temporal params and normalize native temporals in Cypher

### LINKED ISSUE #4905: Bolt: inbound temporal PackStream structs (Date/Time/DateTime/…) not decoded → datetime query parameters silently dropped
## Summary

ArcadeDB's Bolt plugin does not decode **incoming** temporal PackStream structures (`Date`, `Time`, `LocalTime`, `LocalDateTime`, `DateTime`, `DateTimeZoneId`, `Duration`). When a Bolt client sends a native temporal value as a **query parameter**, `PackStreamReader.readStructure()` returns it as an opaque `StructureValue` (raw signature + fields) instead of a `java.time` value. The parameter therefore reaches the query engine as a meaningless object and is effectively dropped: a `datetime`-typed parameter can never bind, match, or be stored correctly.

The **output/read** direction works (properties are streamed back to the client, `BoltStructureMapper` renders temporals as ISO-8601 strings), so the gap is input-only and easy to miss.

## Credit

Diagnosed by **@agc-63** (Angel Garcia) while running the community ArcadeDB driver for [getzep/graphiti](https://github.com/getzep/graphiti) end-to-end against a real ArcadeDB 26.7.1 instance. See the detailed write-up on getzep/graphiti#1310:
https://github.com/getzep/graphiti/pull/1310#issuecomment-4871937981

He confirmed it by reading `PackStreamReader.java` directly and worked around it at the `neo4j` Python driver level (monkeypatching `dehydrate_datetime` to emit ISO strings). The proper fix belongs here in ArcadeDB.

## Root cause

`bolt/src/main/java/com/arcadedb/bolt/packstream/PackStreamReader.java` — `readStructure()` collects the struct fields and returns a generic `StructureValue`; there is no branch that maps the Bolt temporal signatures back to `java.time` types. `BoltStructureMapper` only implements the reverse (Java → Bolt, as strings) direction.

Note ArcadeDB negotiates Bolt **v4.4 max** (`BoltNetworkExecutor.SUPPORTED_VERSIONS = { v4.4, v4.0, v3.0 }`), so clients use the **legacy (pre-5.0)** `DateTime`/`DateTimeZoneId` encoding where the `seconds` field is the *local* epoch-second (offset folded in), not the UTC epoch-second. The decoder must use the legacy semantics (and may defensively also handle the 5.0 UTC signatures `0x49`/`0x69`).

The decode must happen on the **parameter** path, not in the generic `readStructure()`: the top-level Bolt `ROUTE` message signature (`0x66`) collides with the legacy `DateTimeZoneId` signature (`0x66`, `'f'`), and message envelopes flow through the same reader. Temporal decoding is therefore applied when hydrating the RUN parameters map (where a `0x66` struct is unambiguously a temporal), leaving message parsing untouched.

## Scope: this is 1 of 2 fixes for full parity

Decoding the wire is necessary but not sufficient for a query like `WHERE e.valid_at <= $reference_time` to work like Neo4j/FalkorDB. The Cypher engine also needs a follow-up so a native temporal **parameter** participates in temporal comparison:

- `ast/ParameterExpression.evaluate()` returns the parameter value raw (no temporal normalization).
- `ast/ComparisonExpression` only takes the temporal-comparison branch when **both** operands are `CypherTemporalValue`; there is no `java.util.Date`/`java.time` `Comparable` fallback.
- `ast/PropertyAccessExpression.convertFromStorage()` wraps `LocalDate`/`LocalDateTime` (and parseable strings) into `CypherTemporalValue`, but not `java.util.Date` or `ZonedDateTime`.

Follow-up (separate change): normalize `java.time`/`java.util.Date` parameters into `CypherTemporalValue` for comparison, and extend `convertFromStorage()` to wrap `java.util.Date`/`ZonedDateTime`. That closes the loop so no client-side `datetime(...)` wrapping is needed.

## Fix (this issue)

Decode inbound Bolt temporal structs into `java.time` types on the parameter path, mirroring the signatures `BoltStructureMapper` already emits outbound:

| Bolt struct | Signature | Decoded to |
|---|---|---|
| Date | `0x44` `'D'` | `LocalDate` |
| Time | `0x54` `'T'` | `OffsetTime` |
| LocalTime | `0x74` `'t'` | `LocalTime` |
| LocalDateTime | `0x64` `'d'` | `LocalDateTime` |
| DateTime (legacy) | `0x46` `'F'` | `OffsetDateTime` |
| DateTimeZoneId (legacy) | `0x66` `'f'` | `ZonedDateTime` |
| DateTime (UTC, 5.0+) | `0x49` `'I'` | `OffsetDateTime` |
| DateTimeZoneId (UTC, 5.0+) | `0x69` `'i'` | `ZonedDateTime` |

These map onto ArcadeDB's `DATETIME`/`DATE` native types (`Type.TYPES_BY_USERTYPE`), so they are storable and comparable once the engine follow-up lands.

## Test plan

- Bolt-module unit regression test: round-trip each temporal struct through `PackStreamWriter` → `PackStreamReader` → decoder and assert the correct `java.time` value, plus nested-in-parameter-map hydration.
- Follow-up: engine-level test that `WHERE valid_at <= $param` matches with a native temporal parameter.

lvca:
## Summary

Fixes #4905. ArcadeDB's Bolt plugin never decoded **inbound** temporal PackStream structures, so a native datetime sent as a query parameter arrived as an opaque `StructureValue` and was silently dropped. Decoding the wire alone is not sufficient: the Cypher engine only compares two `CypherTemporalValue` operands and returns parameters raw, so a native temporal parameter never matched a stored temporal. This PR fixes both layers.

Diagnosed by **@agc-63** while integrating the ArcadeDB backend driver for [getzep/graphiti](https://github.com/getzep/graphiti/pull/1310#issuecomment-4871937981); the `WHERE e.valid_at <= $reference_time` case there matched nothing.

### Bolt - wire decode (`bolt`)
- **`BoltStructureMapper.fromPackStreamValue()`** - recursively decodes inbound temporal structs (`Date`/`Time`/`LocalTime`/`LocalDateTime`/`DateTime`/`DateTimeZoneId`) into `java.time` values. Handles the legacy encoding (forced by ArcadeDB's Bolt v4.4 cap, where the `seconds` field is the local epoch-second) plus the Bolt 5.0 UTC signatures defensively.
- **`BoltMessage.parseRun()`** - hydrates the RUN parameter map through it. Done on the parameter path (not the generic `readStructure()`) to avoid the `ROUTE` message vs legacy `DateTimeZoneId` `0x66` signature collision.

### Engine - temporal normalization (`engine`)
- **`TemporalUtil.fromCoreJavaType()`** - inverse of the existing `toCoreJavaType()`; wraps native `java.time` / `java.util.Date` values into their `CypherTemporalValue` type (idempotent, non-temporals passed through unchanged).
- **`ComparisonExpression`** - coerces both operands via `fromCoreJavaType()` before the temporal-comparison branch, so a native temporal parameter compares against a stored temporal instead of silently not matching. The write/storage path is untouched, so `SET` serialization is unaffected.
- **`PropertyAccessExpression.convertFromStorage()`** - generalized to wrap all native temporals (incl. `java.util.Date`, the default `DATETIME` storage type, and `ZonedDateTime`), not just `LocalDate`/`LocalDateTime`.

## Tests

- **`PackStreamTemporalInputTest`** (bolt, 11 tests) - real writer -> reader -> decoder wire round-trips for every temporal type (both encodings), nested-in-parameter-map hydration, pre-fix opaque-`StructureValue` documentation, and non-temporal/unknown-struct passthrough.
- **`OpenCypherTemporalParameterTest`** (engine, 3 tests) - reproduces the graphiti `WHERE e.valid_at <= $reference_time` scenario with native `ZonedDateTime`/`Date`/`OffsetDateTime`/`Instant` parameters, over both native-datetime and ISO-string stored values.

### Verification
- Full `bolt` suite: **223 tests**, 0 failures.
- Full `com.arcadedb.query.opencypher.**` suite: **6636 tests**, 0 failures / 0 errors.

## Impact on the graphiti driver

Once on a build with these fixes, the getzep/graphiti ArcadeDB driver can drop both the `neo4j`-driver `dehydrate_datetime` monkeypatch and the `datetime(...)` wrapping in `graph_data_operations.py` - plain `WHERE e.valid_at <= $reference_time` works like the Neo4j/FalkorDB path.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/message/BoltMessage.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/message/BoltMessage.java
@@ -20,6 +20,7 @@ package com.arcadedb.bolt.message;
 
 import com.arcadedb.bolt.packstream.PackStreamReader;
 import com.arcadedb.bolt.packstream.PackStreamWriter;
+import com.arcadedb.bolt.structure.BoltStructureMapper;
 
 import java.io.IOException;
 import java.util.List;
@@ -101,7 +102,11 @@ public abstract class BoltMessage {
   @SuppressWarnings("unchecked")
   private static RunMessage parseRun(final List<Object> fields) {
     final String query = (String) fields.get(0);
-    final Map<String, Object> parameters = fields.size() > 1 && fields.get(1) != null ? (Map<String, Object>) fields.get(1) : Map.of();
+    // Hydrate temporal PackStream structures (Date/Time/DateTime/...) into java.time values so
+    // native temporal query parameters bind correctly instead of being dropped (issue #4905).
+    final Map<String, Object> parameters = fields.size() > 1 && fields.get(1) != null ?
+        (Map<String, Object>) BoltStructureMapper.fromPackStreamValue(fields.get(1)) :
+        Map.of();
     final Map<String, Object> extra = fields.size() > 2 && fields.get(2) != null ? (Map<String, Object>) fields.get(2) : Map.of();
     return new RunMessage(query, parameters, extra);
   }
--- a/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
@@ -402,49 +402,75 @@ public class BoltStructureMapper {
 
   /**
    * Decode a single Bolt temporal PackStream structure into a {@code java.time} value.
-   * Unknown (non-temporal) structures are returned as-is.
+   * Unknown (non-temporal) structures are returned as-is. A structure that carries the wrong field
+   * count or field types for its temporal signature (a misbehaving client) is also returned as-is
+   * rather than propagating a raw {@code IndexOutOfBoundsException} / {@code ClassCastException} out of
+   * RUN parsing - the parameter simply stays opaque instead of crashing the connection.
    */
   private static Object fromTemporalStructure(final PackStreamReader.StructureValue structure) {
     final List<Object> f = structure.getFields();
-    switch (structure.getSignature()) {
-    case SIG_DATE:
-      return LocalDate.ofEpochDay(asLong(f.get(0)));
+    final byte signature = structure.getSignature();
+    if (!hasExpectedArity(signature, f.size()))
+      return structure;
 
-    case SIG_LOCAL_TIME:
-      return LocalTime.ofNanoOfDay(asLong(f.get(0)));
+    try {
+      switch (signature) {
+      case SIG_DATE:
+        return LocalDate.ofEpochDay(asLong(f.get(0)));
 
-    case SIG_TIME:
-      return OffsetTime.of(LocalTime.ofNanoOfDay(asLong(f.get(0))), ZoneOffset.ofTotalSeconds((int) asLong(f.get(1))));
+      case SIG_LOCAL_TIME:
+        return LocalTime.ofNanoOfDay(asLong(f.get(0)));
 
-    case SIG_LOCAL_DATE_TIME:
-      return LocalDateTime.ofEpochSecond(asLong(f.get(0)), (int) asLong(f.get(1)), ZoneOffset.UTC);
+      case SIG_TIME:
+        return OffsetTime.of(LocalTime.ofNanoOfDay(asLong(f.get(0))), ZoneOffset.ofTotalSeconds((int) asLong(f.get(1))));
 
-    case SIG_DATE_TIME_OFFSET_LEGACY: {
-      // Legacy: seconds is the local epoch-second; reconstruct the wall clock then stamp the offset.
-      final LocalDateTime local = LocalDateTime.ofEpochSecond(asLong(f.get(0)), (int) asLong(f.get(1)), ZoneOffset.UTC);
-      return OffsetDateTime.of(local, ZoneOffset.ofTotalSeconds((int) asLong(f.get(2))));
-    }
+      case SIG_LOCAL_DATE_TIME:
+        return LocalDateTime.ofEpochSecond(asLong(f.get(0)), (int) asLong(f.get(1)), ZoneOffset.UTC);
 
-    case SIG_DATE_TIME_ZONEID_LEGACY: {
-      final LocalDateTime local = LocalDateTime.ofEpochSecond(asLong(f.get(0)), (int) asLong(f.get(1)), ZoneOffset.UTC);
-      return ZonedDateTime.of(local, ZoneId.of(String.valueOf(f.get(2))));
-    }
+      case SIG_DATE_TIME_OFFSET_LEGACY: {
+        // Legacy: seconds is the local epoch-second; reconstruct the wall clock then stamp the offset.
+        final LocalDateTime local = LocalDateTime.ofEpochSecond(asLong(f.get(0)), (int) asLong(f.get(1)), ZoneOffset.UTC);
+        return OffsetDateTime.of(local, ZoneOffset.ofTotalSeconds((int) asLong(f.get(2))));
+      }
 
-    case SIG_DATE_TIME_OFFSET_UTC: {
-      // UTC (Bolt 5.0+): seconds is the true UTC epoch-second.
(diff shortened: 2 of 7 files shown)


lvca:
Thanks for the review. Addressed in `2b929f9e0`:

- **Hot-path perf (ComparisonExpression)** ✅ - coercion is now guarded by `left/right instanceof Temporal || Date`, so the common numeric/string/boolean comparison pays two `instanceof` checks instead of two method calls per operand. `CypherTemporalValue` operands aren't `Temporal`/`Date`, so they skip the guard and flow straight into the existing branch check - nothing regresses.
- **Malformed temporal structs (robustness)** ✅ - `fromTemporalStructure()` now validates field arity per signature and wraps decoding in a defensive catch. A misbehaving client sending a wrong-arity or wrong-type temporal leaves the value opaque (same as an unknown struct / Duration) instead of propagating a raw `IndexOutOfBoundsException` / `ClassCastException` out of RUN parsing. Added a `malformedTemporalStructIsLeftOpaqueInsteadOfCrashing` test.
- **Style nit** ✅ - `Consumer` is now imported in `OpenCypherTemporalParameterTest`.

On the known gaps:

- **Duration parameters** - deferred, as flagged. Still fall through to opaque (`WHERE ... <= $someDuration` won't bind). Acknowledged in the code comment; no single `java.time` type maps cleanly and graphiti doesn't use durations, so it's out of scope here.
- **Outbound asymmetry (ISO strings vs native structs)** - filed as follow-up #4907. Pre-existing and out of scope for this PR.
- **convertFromStorage broadening - deliberate** ✅ - confirmed intended. Native `DATETIME` is stored as `java.util.Date`, so wrapping all `Temporal`/`Date` reads (not just `LocalDate`/`LocalDateTime`) makes stored native datetimes read back as comparable `CypherTemporalValue`, consistent with the two types already wrapped. Note the outbound Bolt path still renders these as ISO strings (`toPackStreamValue` handles `CypherDateTime` via its `toString()`), so what a Bolt client receives is unchanged; the broadening only affects embedded/Cypher expression semantics (ordering, component access), and the full 6636-test opencypher suite stays green.

lvca:
Round 2 addressed in `fb00f121e`:

- **Dead `Calendar` branch** ✅ - dropped from `fromCoreJavaType()` (both call sites guard on `Temporal || Date`, which `Calendar` isn't) plus the now-unused import.
- **`PropertyAccessExpression` sibling path** ✅ - the direct component-access branch (`date.year`, ...) was still `LocalDate`/`LocalDateTime`-only. Generalized it to all native `Temporal`/`Date` values via `fromCoreJavaType()`, so a stored `ZonedDateTime`/`java.util.Date` property or a native temporal parameter accessed via `.year`/`.month` now resolves. Dropped the now-unused `LocalDate`/`LocalDateTime` imports. (You were right that `convertFromStorage()` runs first for stored props, so this mainly closes the gap for raw temporal params/UNWIND values used directly.)
- **Per-row re-coercion of the constant parameter** ✅ - added a single-slot `volatile` memo in `ComparisonExpression`: the invariant `$parameter` operand is wrapped into a `CypherTemporalValue` once and reused across rows instead of allocating per row. The stored side is already a `CypherTemporalValue` (via `convertFromStorage`) so it short-circuits and never touches the memo. `volatile` + an immutable `{raw, coerced}` pair array keeps parallel evaluators from seeing a torn pair.
- **Test suggestion** ✅ - added `mismatchedTemporalGranularityDoesNotMatch` (native `ZonedDateTime` param vs stored `CypherDate`): pins the different-temporal-type branch in `compareValuesTernary` (`=` -> false, ordering -> null), the one temporal branch not previously covered.

Left as-is (agreed non-issues):
- **`fromPackStreamValue()` deep-copies no-temporal parameter maps** - once per query on small RUN payloads; a no-op-detecting fast path would need its own scan. Deferred unless profiling shows large payloads.

Verification: full `com.arcadedb.query.opencypher.**` (6637 tests, +1 for the new case) and the bolt PackStream suite both green, 0 failures/0 errors.

PR #4910, opened 2026-07-03
fix(#4908,#4909): retryable Bolt conflicts + nested $param.field in MATCH patterns

### LINKED ISSUE #4908: Bolt: map optimistic-concurrency (NeedRetry) conflicts to a Neo4j transient error code so drivers auto-retry
## Summary

ArcadeDB's optimistic-concurrency conflicts (MVCC page-version conflicts) surface to Bolt clients as **non-retryable** errors. The Bolt plugin maps every query-execution exception to `Neo.DatabaseError.General.UnknownError` (and commit-path conflicts to `Neo.ClientError.Transaction.TransactionNotFound`), neither of which the Neo4j drivers treat as retryable. So a standard driver using a managed transaction (`session.execute_write(...)`) does **not** auto-retry a conflict, even though ArcadeDB's own error message explicitly says "Please retry the operation".

Diagnosed by @agc-63 while integrating the ArcadeDB backend for [getzep/graphiti](https://github.com/getzep/graphiti/pull/1310#issuecomment-4871937981) - he had to add a bespoke retry loop that string-matches the error text (`"Concurrent modification"` / `TransactionNotFound`) because the driver never retried on its own.

## Root cause

- `bolt/.../BoltNetworkExecutor.java` - the RUN (`handleRun`) and PULL (`handlePull`) query-error catch blocks call `sendFailure(BoltErrorCodes.DATABASE_ERROR, ...)`; the COMMIT path uses `TRANSACTION_ERROR` (`Neo.ClientError.Transaction.TransactionNotFound`).
- Both are `ClientError`/`DatabaseError` classifications; Neo4j drivers only auto-retry on `TransientError`-classified codes.

ArcadeDB already models these as retryable internally: `ConcurrentModificationException` and `LockTimeoutException` both extend `NeedRetryException` (`com.arcadedb.exception`).

## Proposed fix

Classify a thrown `NeedRetryException` (checked through the cause chain) to a Neo4j **transient** status code so drivers auto-retry, e.g. `Neo.TransientError.Transaction.DeadlockDetected` (avoid `...Transaction.Terminated` / `...LockClientStopped`, which the drivers explicitly exclude from retry). Apply it in the RUN, PULL, and COMMIT error paths.

## Impact

Removes the need for the graphiti driver's custom conflict-retry loop - `session.execute_write()` retries transparently, as it does against Neo4j.

## Test plan

- Bolt-module unit test asserting a `NeedRetryException` classifies to the transient code and a generic exception still maps to `DATABASE_ERROR`.

Follow-up to #4905, flagged during review of #4906.

### LINKED ISSUE #4909: OpenCypher: nested parameter field access ($param.field) in a MATCH pattern property map does not resolve
## Summary

In the OpenCypher engine, a nested parameter field access - `$param.field` - used as a **property value inside a MATCH pattern** does not resolve. The pattern matches nothing, with no error:

```cypher
MATCH (source:Entity {uuid: $edge_data.source_uuid})
MATCH (target:Entity {uuid: $edge_data.target_uuid})
MERGE (source)-[e:RELATES_TO {uuid: $edge_data.uuid}]->(target)
SET e = $edge_data
```

The `MERGE` runs without exception but never creates the relationship, because the two `MATCH` clauses silently bind nothing. The same query with a top-level scalar parameter (`{uuid: $source_uuid}`) works, and wrapping the map in `UNWIND [$edge_data] AS edge` then referencing `edge.source_uuid` also works - which is the workaround the graphiti driver currently uses.

Diagnosed by @agc-63 while integrating the ArcadeDB backend for [getzep/graphiti](https://github.com/getzep/graphiti/pull/1310#issuecomment-4871937981) (verified against a real instance).

## Root cause (to confirm)

`engine/.../executor/steps/MatchNodeStep.java` resolves a pattern property value only when it is a bare `ParameterReference` (whole-parameter substitution) or an `Expression` evaluated against a current input row. A `$param.field` map-field access is neither handled as a parameter nor evaluable without an input row, so the pattern property resolves to the wrong value (or is skipped) and the match fails silently.

## Proposed fix

Resolve `$param.field` (parameter map/collection field access) in MATCH pattern property maps: evaluate the parameter-property-access expression against the input parameters even when there is no preceding input row, mirroring how `UNWIND [$param] AS x` + `x.field` already resolves. Add a regression test under `engine/src/test/java/com/arcadedb/query/opencypher/`.

## Impact

Lets the graphiti driver drop its `UNWIND [$edge_data]` workaround for single edge saves and use the natural `MATCH (... {uuid: $edge_data.source_uuid})` form.

Follow-up to #4905, flagged during review of #4906.

lvca:
## Summary

Two follow-ups to #4905, flagged during review of #4906. Both are root-cause fixes in ArcadeDB for workarounds the [getzep/graphiti](https://github.com/getzep/graphiti/pull/1310) driver currently carries; landing them lets the driver drop those workarounds. Diagnosed by @agc-63.

Closes #4908, closes #4909.

## #4908 - Bolt: optimistic-concurrency conflicts are now retryable

ArcadeDB's MVCC page-version conflicts surfaced to Bolt clients as **non-retryable** errors (`Neo.DatabaseError.General.UnknownError` for query execution, `Neo.ClientError.Transaction.TransactionNotFound` on commit), so a managed-transaction driver (`session.execute_write(...)`) never auto-retried - even though ArcadeDB's own message says "Please retry the operation".

- `BoltErrorCodes`: add `TRANSIENT_CONFLICT_ERROR = Neo.TransientError.Transaction.DeadlockDetected` - a `TransientError` classification the Neo4j drivers retry on (avoiding the two excluded titles `Transaction.Terminated` / `Transaction.LockClientStopped`).
- `BoltNetworkExecutor.classifyExecutionError()`: classify any `NeedRetryException` in the cause chain (`ConcurrentModificationException` / `LockTimeoutException`) to the transient code; applied in the RUN, PULL and COMMIT error paths.
- Removes the need for the graphiti driver's bespoke error-string-matching retry loop.

## #4909 - OpenCypher: nested parameter field access in MATCH patterns

`MATCH (source:Entity {uuid: $edge_data.source_uuid})` matched nothing (no error): the pattern property parses to a `ChainedPropertyAccessExpression` whose base `ParameterExpression` resolves from the context alone, but `MatchNodeStep` only evaluated pattern-property expressions when an input row was present. With a bare MATCH (no row) the expression was left unevaluated and never matched - which is why the graphiti driver had to wrap single edge saves in `UNWIND [$edge_data] AS edge`.

- `MatchNodeStep`: evaluate pattern-property expressions with a shared read-only empty result when there is no input row, in both the scan path (`matchesProperties`) and the index path (`tryFindAndUseIndex`). Parameter-based expressions resolve; row-dependent ones return null and fall back to scan.

## Tests

- `BoltErrorClassificationTest` (5) - conflict / lock-timeout / wrapped-cause -> transient, generic -> default, and the transient code is a driver-retryable classification.
- `NestedParameterInMatchTest` (3) - scan path, index path, and the graphiti `MATCH ... MERGE` edge-save shape.

## Verification

- Full `com.arcadedb.query.opencypher.**`: 6640 tests, 0 failures / 0 errors.
- Full `bolt` module: 229 tests, 0 failures / 0 errors.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltErrorCodes.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltErrorCodes.java
@@ -38,6 +38,12 @@ public final class BoltErrorCodes {
   // Transaction errors
   public static final String TRANSACTION_ERROR = "Neo.ClientError.Transaction.TransactionNotFound";
 
+  // Transient (retryable) errors. ArcadeDB's optimistic-concurrency conflicts (NeedRetryException:
+  // ConcurrentModificationException / LockTimeoutException) map here so Neo4j drivers auto-retry a
+  // managed transaction. The code is a TransientError classification that the drivers retry on; the
+  // two excluded titles (Transaction.Terminated / Transaction.LockClientStopped) are deliberately avoided.
+  public static final String TRANSIENT_CONFLICT_ERROR = "Neo.TransientError.Transaction.DeadlockDetected";
+
   // Request errors
   public static final String PROTOCOL_ERROR = "Neo.ClientError.Request.Invalid";
 
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -209,6 +209,9 @@ public class BoltNetworkExecutor extends Thread {
           }
           break;
         } catch (final Exception e) {
+          // Top-level safety net for unexpected dispatch/protocol failures. NeedRetryException (MVCC)
+          // classification is handled where those conflicts actually arise - the RUN/PULL/COMMIT
+          // query-execution handlers - so this fallback intentionally keeps the generic DATABASE_ERROR.
           LogManager.instance().log(this, Level.WARNING, "BOLT error processing message", e);
           try {
             sendFailure(BoltException.DATABASE_ERROR, e.getMessage());
@@ -625,9 +628,12 @@ public class BoltNetworkExecutor extends Thread {
       sendFailure(BoltException.SYNTAX_ERROR, parseMsg);
       state = State.FAILED;
     } catch (final Exception e) {
-      LogManager.instance().log(this, Level.WARNING, "BOLT query error", e);
+      // MVCC conflicts (NeedRetryException) are expected under contention and auto-retried by the driver,
+      // so log them at FINE to avoid flooding WARNING with normal, recoverable flow; genuine errors stay WARNING.
+      final boolean retryable = isRetryableConflict(e);
+      LogManager.instance().log(this, retryable ? Level.FINE : Level.WARNING, "BOLT query error", e);
       final String errorMsg = e.getMessage() != null ? e.getMessage() : "Database error";
-      sendFailure(classifyExecutionError(e, BoltErrorCodes.DATABASE_ERROR), errorMsg);
+      sendFailure(retryable ? BoltErrorCodes.TRANSIENT_CONFLICT_ERROR : BoltErrorCodes.DATABASE_ERROR, errorMsg);
       state = State.FAILED;
     }
   }
@@ -724,9 +730,12 @@ public class BoltNetworkExecutor extends Thread {
       sendSuccess(metadata);
 
     } catch (final Exception e) {
-      LogManager.instance().log(this, Level.WARNING, "BOLT PULL error", e);
+      // MVCC conflicts (incl. those raised by an implicit auto-commit here) are expected and auto-retried
+      // by the driver, so log at FINE to avoid flooding WARNING with recoverable flow; real errors stay WARNING.
+      final boolean retryable = isRetryableConflict(e);
+      LogManager.instance().log(this, retryable ? Level.FINE : Level.WARNING, "BOLT PULL error", e);
       final String errorMsg = e.getMessage() != null ? e.getMessage() : "Error fetching records";
-      sendFailure(classifyExecutionError(e, BoltErrorCodes.DATABASE_ERROR), errorMsg);
+      sendFailure(retryable ? BoltErrorCodes.TRANSIENT_CONFLICT_ERROR : BoltErrorCodes.DATABASE_ERROR, errorMsg);
       state = State.FAILED;
     }
   }
@@ -1624,13 +1633,22 @@ public class BoltNetworkExecutor extends Thread {
    * transient status so managed-transaction drivers auto-retry; anything else keeps the given default.
    */
   static String classifyExecutionError(final Throwable error, final String defaultCode) {
+    return isRetryableConflict(error) ? BoltErrorCodes.TRANSIENT_CONFLICT_ERROR : defaultCode;
+  }
+
+  /**
+   * Whether the error (or any wrapped cause) is one of ArcadeDB's optimistic-concurrency conflicts
+   * ({@link NeedRetryException}). Such conflicts are expected under contention and auto-retried by the
+   * driver, so callers both classify them as transient and log them at a lower level.
+   */
+  static boolean isRetryableConflict(final Throwable error) {
     // Bounded walk: the depth cap guards against a self-referential / cyclic cause chain spinning forever.
     Throwable t = error;
     for (int depth = 0; t != null && depth < 32; t = t.getCause(), depth++) {
       if (t instanceof NeedRetryException)
-        return BoltErrorCodes.TRANSIENT_CONFLICT_ERROR;
+        return true;
     }
-    return defaultCode;
+    return false;
   }
 
   /**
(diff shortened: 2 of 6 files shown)


lvca:
Thanks for the thorough review (@gemini-code-assist and the Claude reviewer). Addressed all points:

**#4909 - shared `EMPTY_RESULT` hardening (main recommendation)** ✅
Constructed as `new ResultInternal(Collections.emptyMap())`. `ResultInternal(Map)` uses the passed map directly as its backing store, and `setProperty()` does `content.put(...)`, so any accidental write now fails fast with `UnsupportedOperationException` instead of silently racing shared state across concurrent queries. Reads on empty content stay null-returning and safe, so it's still fine to share the static instance.

**#4909 - index-path behavior change** ✅
Restructured `tryFindAndUseIndex` so the `null -> fall back to scan` handling applies **only** in the no-input-row branch. When `currentInputResult != null` the original behavior is preserved exactly (evaluate against the row and use the result, even if null). So the only behavior change is the new no-row case, matching the comment.

**#4908 - cyclic cause chain** ✅
`classifyExecutionError` now walks with a depth cap (32) so a self-referential/cyclic `getCause()` can't spin forever. Added a `cyclicCauseChainTerminatesAndKeepsDefault` test.

**Style nit - fully-qualified `Consumer`** ✅
Imported `java.util.function.Consumer` in `NestedParameterInMatchTest` (this was also the single Codacy CodeStyle finding).

On the test-path nit (`./target/testnestedparam_<nanoTime>` vs a shared TestHelper): I kept the direct `DatabaseFactory` + `finally { drop() }` pattern to match the sibling `ParameterTest`/`OpenCypherTemporalParameterTest` in the same package, but happy to switch if you'd prefer TestHelper.

Re-verified: full `com.arcadedb.query.opencypher.**` and full `bolt` module both green.

lvca:
Addressed the two operational suggestions in `1560f3d09`:

**Log level for expected conflicts** ✅
Extracted `isRetryableConflict()` and, in the RUN and PULL handlers, a `NeedRetryException` now logs at `FINE` instead of `WARNING` (genuine errors stay `WARNING`). So a busy contended graphiti workload no longer emits a stream of WARNING stack traces for what is now normal, driver-retried flow.

**Uncovered fallback catch** ✅
Added a comment at the top-level dispatch catch noting that `NeedRetry` (MVCC) classification is intentionally handled in the per-message RUN/PULL/COMMIT handlers where those conflicts actually arise, so the generic `DATABASE_ERROR` fallback there is deliberate.

Also added a direct `isRetryableConflict` test (drives both the transient classification and the log-level demotion). Full bolt module green: 231 tests, 0 failures.

Thanks for the operational eye - the FINE/WARNING split is the kind of thing that only bites in production.

lvca:
Addressed in `2ec039a0e`:

**Import ordering** ✅
Moved `import com.arcadedb.exception.NeedRetryException;` into alphabetical position (right after `com.arcadedb.exception.CommandParsingException`) instead of ahead of the `com.arcadedb.bolt.message.*` block.

**End-to-end Bolt wire test** ✅ (took your optional suggestion - it's a contract worth pinning)
Added `Bolt4908TransientConflictIT`: drives a real MVCC conflict over the Bolt protocol with the neo4j driver and asserts it surfaces as a `TransientException` with code `Neo.TransientError.Transaction.DeadlockDetected`. Made it **deterministic, not racy** - two explicit transactions modify the same record and commit in a fixed order, so the first commit bumps the page version and the second reliably detects the stale version and conflicts. Confirmed stable across repeated runs (4/4). This locks in the RUN/PULL/COMMIT wire-level contract the graphiti driver's managed-transaction retry depends on. It runs in the failsafe/`verify` phase like the other `*IT` classes, so it doesn't slow the unit `test` phase.

On the two awareness-only notes:
- **`EMPTY_RESULT` read-only invariant** - agreed it's now load-bearing; fail-fast (`UnsupportedOperationException`) beats silent shared-state corruption, and the full opencypher suite confirms no evaluator path mutates it today.
- **Minor duplication** (RUN/PULL inline vs `classifyExecutionError`) - left as-is since the handlers need the boolean for the FINE/WARNING log level anyway; the shared logic lives in `isRetryableConflict`.

Full bolt unit suite green (231), and the new IT passes deterministically.

lvca:
Addressed in `7f8329cdc`:

**DRY in RUN/PULL handlers** ✅
Both handlers now keep `isRetryableConflict(e)` only for the FINE/WARNING log level and delegate the error-code selection to `classifyExecutionError(e, BoltErrorCodes.DATABASE_ERROR)`, so the retryable→transient-code mapping lives in one place (the extra bounded cause-walk is error-path only and negligible).

On the two awareness notes - both left as-is, agreed:
- **Retry semantics** - mapping to a `TransientError` means the driver retries with backoff up to its max-retry-time (~30s default). That's the intended tradeoff for the graphiti managed-transaction use case; a permanent failure wrapping a `NeedRetryException` deep in its cause chain is not something ArcadeDB produces (these exceptions are terminal conflict signals, not wrappers around permanent errors).
- **Naming** - `TRANSIENT_CONFLICT_ERROR` reuses the Neo4j deadlock title deliberately: it's the retryable `TransientError` classification the drivers honor while avoiding the two excluded titles. The constant's comment carries that reasoning.

Bolt unit suite green (231) and the MVCC-conflict IT still deterministic.

PR #4921, opened 2026-07-03
fix(#4907): serialize outbound Bolt temporals as native PackStream structs

### LINKED ISSUE #4907: Bolt: serialize outbound temporal values as native PackStream structs instead of ISO strings
## Summary

Follow-up to #4905 (which fixed the **inbound** direction). ArcadeDB's Bolt plugin serializes temporal property values **outbound** as ISO-8601 strings rather than native Bolt temporal PackStream structures. A Neo4j-compatible client therefore receives a `String`, not a native `datetime`/`date`/`time` value, so the round-trip is asymmetric: a value written by ArcadeDB and read back by a Bolt driver is not a native temporal.

## Where

`bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.toPackStreamValue()` renders every temporal via `.toString()`:

```java
if (value instanceof LocalDate date)       return date.toString();
if (value instanceof LocalDateTime dateTime) return dateTime.toString();
if (value instanceof ZonedDateTime dateTime) return dateTime.toString();
if (value instanceof Instant instant)      return instant.toString();
if (value instanceof Date date)            return date.toInstant().toString();
// ...
```

## Proposed fix

Emit the proper Bolt temporal structures instead of strings, using the same signatures #4905 now decodes on input (and the legacy pre-5.0 encoding, since ArcadeDB negotiates Bolt v4.4 max):

| java.time | Bolt struct | Signature | Fields |
|---|---|---|---|
| LocalDate | Date | `0x44` | [days] |
| OffsetTime | Time | `0x54` | [nanoOfDay, offsetSeconds] |
| LocalTime | LocalTime | `0x74` | [nanoOfDay] |
| LocalDateTime | LocalDateTime | `0x64` | [seconds, nanos] |
| ZonedDateTime (offset) | DateTime (legacy) | `0x46` | [secondsLocal, nanos, offsetSeconds] |
| ZonedDateTime (zone id) | DateTimeZoneId (legacy) | `0x66` | [secondsLocal, nanos, zoneId] |

This would make ArcadeDB temporals round-trip natively over Bolt, matching Neo4j client expectations.

## Notes

- The write direction (`PackStreamWriter`) already has `writeStructureHeader()`; the encoding helpers can mirror the inbound decoder added in #4905.
- Flagged during review of #4906.

lvca:
## Summary

Closes #4907. Completes the temporal round-trip started in #4905 (which fixed the **inbound** direction). ArcadeDB's Bolt plugin previously serialized temporal property values **outbound** as ISO-8601 strings, so a Neo4j-compatible client received a `String` instead of a native `datetime`/`date`/`time`. A value written by ArcadeDB and read back by a Bolt driver is now a real temporal, matching Neo4j semantics.

Flagged during review of #4906.

## Changes

- **`BoltTemporalStructure`** - a `PackStreamStructure` that writes a temporal signature + fields (the writer already dispatches to `PackStreamStructure.writeTo`).
- **`BoltStructureMapper.toTemporalStructure()`** - encodes `LocalDate` / `LocalTime` / `OffsetTime` / `LocalDateTime` / `OffsetDateTime` / `ZonedDateTime` / `Instant` / `java.util.Date` / `Calendar` into the matching Bolt struct, using the **legacy (Bolt <=4.4)** encoding that mirrors the inbound decoder from #4905 (seconds = local epoch-second). Handles both:
  - raw `java.time`/`Date` values (from a `RETURN e` element), and
  - Cypher temporal wrappers (`CypherDateTime` etc. from a scalar `RETURN e.valid_at`), unwrapped to their `java.time` value first.

## Tests

- **`BoltStructureTest`** - the 9 temporal cases now assert **native-struct output via a full wire round-trip** (`toPackStreamValue` -> write -> read -> `fromPackStreamValue`), replacing the old ISO-string assertions.
- **`Bolt4907TemporalOutputIT`** - end-to-end: a real neo4j driver decodes `RETURN e.ts` / `e.day` into `java.time` (`asZonedDateTime()` / `asLocalDate()` would throw if a string were returned).

## Verification

- Full `bolt` unit suite: 235 tests, 0 failures.
- `BoltProtocolIT` (77) + `BoltMultiLabelIT` (11) + `Bolt4907TemporalOutputIT` (1): all green.
- `toPackStreamValue` is Bolt-output-only, so nothing outside the Bolt module is affected.

## Follow-up note

With #4905 (inbound) + this (outbound), Bolt temporals are now fully symmetric in both directions - the graphiti driver gets native datetimes on both read and write with no client-side conversion.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
@@ -379,10 +379,17 @@ public class BoltStructureMapper {
     }
     if (value instanceof Instant i)
       return dateTimeWithOffset(LocalDateTime.ofInstant(i, ZoneOffset.UTC), ZoneOffset.UTC);
+    // java.sql.Date / java.sql.Time extend java.util.Date but carry only one component; toInstant()
+    // throws UnsupportedOperationException on them, so map to the date-only / time-only value first.
+    if (value instanceof java.sql.Date sqlDate)
+      return toTemporalStructure(sqlDate.toLocalDate());
+    if (value instanceof java.sql.Time sqlTime)
+      return toTemporalStructure(sqlTime.toLocalTime());
     if (value instanceof Date date)
       return dateTimeWithOffset(LocalDateTime.ofInstant(date.toInstant(), ZoneOffset.UTC), ZoneOffset.UTC);
     if (value instanceof Calendar calendar)
-      return dateTimeWithOffset(LocalDateTime.ofInstant(calendar.toInstant(), ZoneOffset.UTC), ZoneOffset.UTC);
+      // Preserve the calendar's own zone instead of forcing UTC.
+      return toTemporalStructure(ZonedDateTime.ofInstant(calendar.toInstant(), calendar.getTimeZone().toZoneId()));
 
     return null;
   }
--- a/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
@@ -383,6 +383,9 @@ public class BoltStructureMapper {
       return new BoltTemporalStructure(SIG_DATE_TIME_ZONEID_LEGACY, zdt.toLocalDateTime().toEpochSecond(ZoneOffset.UTC),
           (long) zdt.getNano(), zdt.getZone().getId());
     }
+    // A bare instant (Instant / java.util.Date) has no zone; Bolt has no pure-instant type, so it is
+    // deliberately widened to a DateTime struct at UTC. The instant is preserved; the client receives a
+    // zoned/offset datetime at UTC rather than a "naked" instant.
     if (value instanceof Instant i)
       return dateTimeWithOffset(LocalDateTime.ofInstant(i, ZoneOffset.UTC), ZoneOffset.UTC);
     // java.sql.Date / java.sql.Time extend java.util.Date but carry only one component; toInstant()
--- a/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
@@ -25,6 +25,11 @@ import com.arcadedb.database.RID;
 import com.arcadedb.graph.Edge;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.query.opencypher.Labels;
+import com.arcadedb.query.opencypher.temporal.CypherDate;
+import com.arcadedb.query.opencypher.temporal.CypherDateTime;
+import com.arcadedb.query.opencypher.temporal.CypherLocalDateTime;
+import com.arcadedb.query.opencypher.temporal.CypherLocalTime;
+import com.arcadedb.query.opencypher.temporal.CypherTime;
 import com.arcadedb.query.sql.executor.Result;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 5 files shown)


lvca:
Thanks both - addressed everything actionable across the two reviews (`44b33d4d4` + `8157bf55a`):

**gemini**
- **`java.sql.Date` / `java.sql.Time` crash (high)** ✅ - both extend `java.util.Date` but `toInstant()` throws on them, so they'd hit the `Date` branch and crash. Now intercepted before it and mapped to a Date / LocalTime struct. Added `mapperSqlDate` / `mapperSqlTime` round-trip tests.
- **`Calendar` timezone loss** ✅ - now converts via the calendar's own zone (`ZonedDateTime.ofInstant(cal.toInstant(), cal.getTimeZone().toZoneId())`) instead of forcing UTC.
- **Defensive-copy `BoltTemporalStructure.fields`** ✅ - cloned in the constructor. (The getter is dropped, see below, so no exposed reference remains.)

**claude**
- **Offset case in the real-driver IT (main item)** ✅ - `Bolt4907TemporalOutputIT` now also asserts a non-zero-offset `datetime('...+02:00')`, a `localdatetime`, and a `localtime` through the real neo4j driver, so the offset/local paths have genuine wire-convention coverage rather than only the self-consistent round-trip.
- **`getFields()` dead code** ✅ - dropped (the other `getFields()` calls are on `PackStreamReader.StructureValue`; `writeTo` reads the private field directly).
- **Behavior-change / release note** ✅ - documented in `docs/release-26.8.1.md` as a Bolt breaking change (temporal properties now native structs, paired with the inbound #4905 decode).
- **ZonedDateTime-with-offset → OffsetDateTime round-trip** - acknowledged, intended (mirrors Neo4j's offset-vs-zone-id distinction; instant + offset preserved).
- **~14 instanceof in `toTemporalStructure`** - left as-is: common scalars (String/Number/Boolean/byte[]) already short-circuit earlier in `toPackStreamValue`, so only rare non-scalar objects reach the temporal chain.

Full bolt unit suite green (237); the extended output IT and the #4908 conflict IT both pass.

lvca:
Thanks - addressed the new points in `b01da26a3`:

1. **Version/encoding coupling (the merge-blocker item)** ✅ - documented at `toTemporalStructure` that the legacy encoding is correct *because* negotiation caps at Bolt 4.4, with an explicit TODO to branch on the negotiated version and emit `SIG_*_UTC` if 5.0 is ever added to `SUPPORTED_VERSIONS` (otherwise a 5.0 driver would decode these legacy structs to the wrong instant). The coupling is now explicit rather than implicit.
2. **Redundant defensive copy on the hot path** ✅ - dropped the constructor's `fields.clone()`. Every call site passes a freshly-allocated varargs array and the array is never exposed (the getter is gone; `writeTo` reads it directly), so the copy was pure per-value GC churn on the serialization path. (This reverses the earlier defensive-copy suggestion, which only made sense while the getter existed - now it doesn't.)
3. **Real-driver coverage of time-only / zone-id** ✅ - `Bolt4907TemporalOutputIT` now also asserts, through the real neo4j driver: a named-zone `datetime('...+01:00[Europe/Rome]')` (DateTimeZoneId path), an offset `time('10:30:45+02:00')` (Time path), plus the `localdatetime` / `localtime` added earlier. So every signature now has genuine real-driver coverage, not just self round-trip.
4. **`CypherDuration` falls through** ✅ - added a NOTE at `unwrapCypherTemporal` explaining it's intentional (no single `java.time` value for months+days+seconds; still emitted as an ISO-8601 string), flagged as a follow-up if native Duration output is wanted.
5. **`mapperLocalDate` isInstanceOf consistency** - left the one explicit `isInstanceOf(BoltTemporalStructure.class)` as a deliberate "it's a struct, not a string" marker for the core behavior change; the round-trip proves it for the rest.

Also note the compatibility/release-note item was already covered in `docs/release-26.8.1.md` (Breaking Changes #2). Full bolt suite green (237); the extended IT passes with the zone-id and offset-time cases.

lvca:
Thanks - addressed both actionable items in `39b62d75c`:

**1. Node-element path coverage** ✅ - added `nodeElementTemporalPropertyIsNative`, which stores a native `localdatetime` (via a Bolt parameter decoded by #4905) and returns the whole node (`RETURN e`), then reads `node.get("ts").asLocalDateTime()`. That exercises the element property-map serialization path (`BoltNode.writeTo`) end-to-end through the real driver, distinct from the scalar projection path.

**2. `Instant` / `java.util.Date` -> offset-UTC widening** ✅ - documented at the `Instant` branch that a bare instant is deliberately widened to a `DateTime` struct at UTC (Bolt has no pure-instant type; the instant is preserved; the client receives a zoned/offset datetime at UTC).

**3. Named-zone DST** - no action, agreed (inherent to the legacy protocol encoding, already flagged by the version TODO).

Worth flagging something the element-path test surfaced: a native `ZonedDateTime`-at-offset parameter decodes (via #4905) to a `java.time.OffsetDateTime`, which ArcadeDB's schemaless type system does not register for storage (`Type.TYPES_BY_USERTYPE` covers `Date`/`LocalDateTime`/`ZonedDateTime`/`Instant` but not `OffsetDateTime`), so such a value is silently dropped on write. That's independent of this PR (outbound serialization), but it will bite the graphiti driver once it stops sending ISO strings, since the Python driver packs a fixed-offset datetime as a `DateTime`-with-offset struct. I'll open a separate follow-up for it (candidate fix: decode the offset `DateTime` struct to a `ZonedDateTime` with a `ZoneOffset`, which is storable, or register `OffsetDateTime` in the type system). I used a `localdatetime` in the element-path test to keep this PR's test on the storable path.

Full bolt unit suite green (237); the output IT (now 2 tests, incl. the node-element case) passes.

Issue #4890, opened 2026-07-02
Bolt: tracking issue for protocol/type-fidelity gaps surfaced by certification


robfrank:
Part of #4882.

### Motivation
Four concrete depth gaps were confirmed by code audit up front: (1) temporal types serialize as ISO strings instead of native Bolt structures, (2) `Duration` and spatial `Point` have no handling at all in `BoltStructureMapper`/`PackStreamWriter`, (3) no `Neo.TransientError.*` error codes are defined so driver-side transient retry can't trigger correctly, (4) ROUTE returns only this node's own address - no HA-aware multi-server routing table - and Bolt 5.x is never advertised/negotiated. Group B's suites would surface exactly how each of these fails and might reveal more; this issue exists to coordinate fixes rather than guess scope upfront.

### Proposed design
- This issue stays a **tracking issue** - as Group B suites go red on these (and possibly other) scenarios, spawn concrete, narrowly-scoped fix sub-issues from here.
- Each spawned sub-issue follows the same template (Motivation/Proposed design/Acceptance criteria/Dependencies/Effort) and links back here and to the Epic.

---

## Status update (2026-07-05): gate lifted, children spawned

All five Group B suites are now merged (#4885 python, #4886 csharp, #4887 go, #4888 js, #4889 java), and `bolt/conformance/spec.yaml` pins the red cells authoritatively (`current_status: expected-fail`, each `tracking_issue: "#4890"`). The evidence exists, so children are now scoped.

**Reconciliation of the original 4 gaps against actual suite evidence:**
- Gap (1) temporal ISO strings -> **RESOLVED** by #4907 (`serialize outbound Bolt temporals as native PackStream structs`). Temporal TYPE scenarios pass. No child needed.
- Gap (3) `Neo.TransientError.*` -> **NOT a gap**. TX-005/ERR-004 are certified positive in the Java suite: `Neo.TransientError` surfaces on a write-write race via overlapping explicit transactions. Documented as certified working; no child needed.
- Gap (2) Duration/Point -> still red (TYPE-011, TYPE-012).
- Gap (4) ROUTE + 5.x -> still red (CONN-004, PROTO-002).

**Plus three additional gaps the suites revealed (as anticipated):** RESULT-004 (summary counters), TYPE-003 (Path never constructed), ERR-002 (semantic vs syntax error code).

### Spawned sub-issues (one per remaining red cell / logical group)
[...]

Issue #4885, opened 2026-07-02
Bolt/e2e-python: add neo4j Python driver conformance suite


robfrank:
Part of #4882.

### Motivation
`e2e-python` currently has zero Bolt/neo4j-driver footprint - confirmed via dependency and full-repo grep audit; it tests only Postgres-wire (`asyncpg`, `sqlalchemy`) and HTTP. Python is one of the five official Neo4j driver languages and has no certification today.

### Proposed design
- Add `neo4j` as a `pyproject.toml` dependency (pinned to the version matrix from the A1 conformance spec).
- New test module, e.g. `e2e-python/tests/test_bolt.py`, reusing the existing `testcontainers`/`DockerContainer` pattern from `test_arcadedb.py` (expose port 7687 alongside the existing 2480/5432), enabling `BoltProtocolPlugin`.
- Implement every scenario in the A1 spec using `pytest` + `pytest-check` (existing conventions in the module).
- Wire into `mvn-test.yml`'s existing `python-e2e-tests` job (already PR-gating) - no new CI job needed, just extend what runs.

### Acceptance criteria
- [ ] `neo4j` Python driver dependency added, version-pinned per the A1 matrix.
- [ ] `test_bolt.py` implements every A1 scenario: connection/TLS, auth, transaction styles, bookmarks, multi-database, result handling, full type round-trip, error codes, protocol negotiation.
- [ ] Known-gap scenarios (temporal/Duration/Point, TransientError, HA routing, 5.x) fail with a clear assertion message referencing the tracking issue, not a silent skip.
- [ ] Runs green in `python-e2e-tests` CI job.

### Dependencies
Depends on the A1 conformance spec. Runs in parallel with the other Group B issues.

### Effort estimate
**M** - new dependency + full scenario suite, but the testcontainers/pytest harness already exists.

Issue #4883, opened 2026-07-02
Bolt: author the shared Bolt conformance spec


robfrank:
Part of #4882.

### Motivation
Every per-language Bolt suite currently tests ad hoc, at uneven depth (confirmed: Java/JS today only cover connect + simple/parameterized reads + one VLP regression - no transactions, no errors, no types). Without one language-neutral scenario matrix as the source of truth, results across five drivers aren't comparable and gaps go unnoticed. Serves the "unify the spec, keep the code polyglot" principle.

### Proposed design
- New file, e.g. `bolt/conformance/spec.yaml` (or `docs/bolt-conformance-spec.md` if YAML is overkill) encoding the full feature matrix from the Epic: connection modes, auth, transaction styles, bookmarks, multi-database, result handling, the full type-round-trip list (incl. currently-broken Duration/Point/temporal), error-code expectations, and protocol-version negotiation.
- Each scenario: inputs, expected result/behavior, and which of the pinned driver versions it applies to.
- Include a shared test dataset/fixture definition (e.g. the existing "beer" dataset used by `ArcadeContainerTemplate`, extended with rows covering the type matrix) so all five languages seed identical data.
- Pin the driver-version matrix per language (e.g. Java: 4.4-series, 5.x, 6.x; JS: same; Python/C#/Go: their equivalent LTS + current + latest).

### Acceptance criteria
- [ ] Spec file committed, covering every row of the Epic's feature matrix table.
- [ ] Pinned driver-version matrix documented per language.
- [ ] Shared fixture/dataset defined and referenced by file path.
- [ ] Spec explicitly marks the 4 known-gap areas (temporal-as-string, missing Duration/Point, missing TransientError codes, single-node-only ROUTE) as scenarios expected to currently fail, not silently omitted.

### Dependencies
None - this is the keystone; everything else in the epic depends on it.

### Effort estimate
**M** - no code changes, but requires enumerating the full matrix precisely enough that 5 independent implementations produce comparable results.

Issue #4882, opened 2026-07-02
Epic: Bolt Driver Compatibility Certification


robfrank:
## Problem

ArcadeDB's `bolt` module is more capable than "works with some drivers" suggests, but has never been certified against the official driver set, and has real, confirmed depth gaps.

**What's already solid** (verified in `bolt/src/main/java/com/arcadedb/bolt/`):
- Full Bolt 3.0/4.0/4.4 message set (HELLO, LOGON, LOGOFF, GOODBYE, RUN, PULL, DISCARD, BEGIN, COMMIT, ROLLBACK, RESET, **and ROUTE**) dispatched in `BoltNetworkExecutor.processMessage`.
- A custom PackStream encoder/decoder (`PackStreamWriter`/`PackStreamReader`).
- Structure mappers for Node/Relationship/UnboundRelationship/Path with correct Bolt tags (`BoltStructureMapper`).
- TLS (`bolt+s`) via `BoltSslHelper` with DISABLED/OPTIONAL/REQUIRED modes.
- Neo4j-style structured error codes (`Neo.ClientError.*`, `Neo.DatabaseError.*`) via `BoltErrorCodes`.
- A deliberate advertised identity: HELLO/SUCCESS reports `"Neo4j/5.26.0 compatible (ArcadeDB ...)"`, pinning the feature envelope drivers assume.
- Working Java (`neo4j-java-driver` 6.2.0, `RemoteBoltDatabaseIT`) and JS (`neo4j-driver` ^6.0.1, `js-bolt-e2e.test.js`) e2e coverage, already PR-gating in `mvn-test.yml`.

**Confirmed gaps this epic exists to close:**
1. **No Bolt 5.x ever advertised** - server negotiates only 3.0/4.0/4.4 (`BoltNetworkExecutor.SUPPORTED_VERSIONS`). 5.x drivers currently work only by silently downgrading; this is undocumented and untested as a deliberate compatibility stance.
2. **ROUTE is single-node only** - `handleRoute` returns this node's own address as writer/reader/router with no HA-aware multi-server address list, so `neo4j://` routing against a cluster is unproven.
3. **Type-fidelity bug**: temporal values (Date, Time, LocalDateTime, DateTime, etc.) serialize as **ISO-8601 strings, not native Bolt structures**; `Duration` and spatial `Point` have **no handling anywhere** in `BoltStructureMapper`/`PackStreamWriter`. Official drivers will not receive typed objects for these.
[...]

PR #4963, opened 2026-07-04
e2e-csharp: add Neo4j.Driver Bolt conformance suite (#4886)

### LINKED ISSUE #4886: Bolt/e2e-csharp: add Neo4j.Driver conformance suite
Part of #4882.

### Motivation
`e2e-csharp` currently has zero Bolt footprint - confirmed via `.csproj` and grep audit; only `Npgsql` is referenced. .NET is one of the five official Neo4j driver languages and has no certification today.

### Proposed design
- Add `Neo4j.Driver` NuGet package to `ArcadeDB.E2ETests.csproj`, version-pinned per the A1 conformance spec matrix.
- New test file, e.g. `BoltE2ETests.cs`, reusing the existing `ArcadeDbFixture` xUnit collection-fixture pattern (extend it to expose port 7687 alongside 2480/5432).
- Implement every A1 scenario as `[Fact]`/`[Theory]` tests following the existing `PostgresE2ETests.cs` conventions.
- Wire into `mvn-test.yml`'s existing `csharp-e2e-tests` job (already PR-gating).

### Acceptance criteria
- [ ] `Neo4j.Driver` NuGet reference added, version-pinned per the A1 matrix.
- [ ] `BoltE2ETests.cs` implements every A1 scenario, matching the Python suite's coverage.
- [ ] Known-gap scenarios fail with clear, referenced assertions rather than being silently absent.
- [ ] Runs green in `csharp-e2e-tests` CI job.

### Dependencies
Depends on the A1 conformance spec. Runs in parallel with the other Group B issues.

### Effort estimate
**M** - same shape as the Python suite, new ecosystem.

robfrank:
## What does this PR do?

Adds a Bolt protocol conformance test suite to `e2e-csharp`, implementing all 39 scenarios in `bolt/conformance/spec.yaml` (issue #4883) against the official `Neo4j.Driver` NuGet package (v6.2.1). This mirrors the architecture and conventions of the already-merged `e2e-python` sibling suite (#4885, PR #4920).

**New files:**
- `BoltFixtures.cs` — three Testcontainers fixtures: a default fixture (Bolt plugin enabled, `beer` + `type_matrix` seeded via HTTP, never over Bolt itself) and two TLS-mode fixtures (REQUIRED/OPTIONAL), each built from a throwaway Docker image with a self-signed keystore/truststore baked in via `COPY` (bind-mounts are unreliable on some CI runners per the Python suite's precedent).
- `KnownGapAssertions.cs` — a small helper (`AssertStillFailsAsync`) that's xUnit's equivalent of pytest's `xfail(strict=True)`: it writes the target-correct assertion for a known, tracked gap, and the build breaks the moment that gap is actually fixed, forcing the fix's author to update the test and `spec.yaml` in the same PR.
- `BoltE2ETests.cs` (37 scenarios) / `BoltTlsE2ETests.cs` (2 scenarios) — every non-gap scenario is a plain `[Fact]`; two are `[Fact(Skip = ...)]` (non-automatable through any official driver's public API); six are wrapped in `AssertStillFailsAsync` for real, currently-tracked gaps.

**Modified:** `ArcadeDB.E2ETests.csproj` (added the `Neo4j.Driver` package reference, pinned to 6.2.1).

No CI workflow changes are needed — the existing `csharp-e2e-tests` job in `mvn-test.yml` already runs `dotnet test` here and is already PR-gating.

## Motivation

Closes epic #4882's Group B item for .NET: certifying ArcadeDB's Bolt implementation against the official .NET driver, matching the same shared spec already implemented for Python.

Note: this design targets `bolt/conformance/spec.yaml` **as corrected by #4885** on `main` — several scenarios originally recorded as `expected-fail` when the spec was authored (#4883) had already been fixed upstream by the time this PR started (native temporal PackStream structs, `Neo.TransientError.*` codes), and two new gaps were discovered along the way (`ERR-002` semantic-error misclassification, `RESULT-004` write counters never populated) — both newly tracked under #4890.

## Related issues

- Part of epic #4882
- Implements #4886
- Depends on #4883 (merged) and cross-references #4885 (merged)
- Known gaps this suite exercises are tracked under #4890

## Additional Notes

Two things worth flagging for reviewers:

1. **`CONN-003` (`neo4j://` routing) fails when run locally on macOS with Docker Desktop.** This is a verified Docker-Desktop-for-Mac limitation, not a product defect: `BoltNetworkExecutor.handleRoute` returns the container's own bridge-network address, which is unreachable from the host across the Docker Desktop VM boundary, but is directly routable on the `ubuntu-latest` runner this project's CI actually uses. The identical scenario already passes in the merged, CI-gating Python suite. Expected to pass in CI here too.
2. During implementation, a real deadlock was found in the original transaction-race test helper (`Task.WhenAll` with two synchronous, blocking `Barrier.SignalAndWait` calls evaluated left-to-right never lets the second participant arrive) and fixed via `Task.Run` dispatch — and a real gap was found where the TLS fixtures didn't seed the `beer` database the TLS tests query against. Both are documented in the commit history.

Design doc: `docs/superpowers/specs/2026-07-04-e2e-csharp-bolt-conformance-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-04-e2e-csharp-bolt-conformance.md`

## Checklist

- [x] I have run the build using `dotnet build` (and the full `e2e-csharp` test project against a live ArcadeDB Docker container)
- [x] Tests cover both passing scenarios and documented known-gap/non-automatable scenarios

### DIFF
--- a/docs/superpowers/plans/2026-07-04-e2e-csharp-bolt-conformance.md
+++ b/docs/superpowers/plans/2026-07-04-e2e-csharp-bolt-conformance.md
@@ -1,0 +1,1422 @@
+# e2e-csharp Bolt Conformance Suite Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `Neo4j.Driver`-based Bolt conformance test suite to `e2e-csharp`, implementing every scenario in `bolt/conformance/spec.yaml` (issue #4886, epic #4882 Group B), matching the coverage and conventions of the already-merged `e2e-python` suite (#4885, PR #4920).
+
+**Architecture:** One default xUnit collection fixture (`ArcadeDbBoltFixture`) spins a single ArcadeDB container with `BoltProtocolPlugin` enabled and the `beer` + `type_matrix` fixtures seeded via HTTP; it backs `BoltE2ETests.cs`, which covers 37 of the 39 spec.yaml scenarios. Two more fixtures spin dedicated containers built from a throwaway Docker image (self-signed keystore/truststore baked in via `COPY`) for the two TLS-mode scenarios, covered by `BoltTlsE2ETests.cs`. Non-automatable scenarios use `[Fact(Skip = ...)]`; real, currently-failing product gaps use a new `KnownGapAssertions.AssertStillFailsAsync` helper (xUnit's equivalent of pytest's `xfail(strict=True)`) so the build breaks the moment a gap is actually fixed.
+
+**Tech Stack:** .NET 10, xUnit 2.9.3, Testcontainers 4.11.0 (incl. `ImageFromDockerfileBuilder`), Neo4j.Driver 6.2.1 (NuGet).
+
+## Global Constraints
+
+- Design doc: `docs/superpowers/specs/2026-07-04-e2e-csharp-bolt-conformance-design.md` - read it for full rationale.
+- Target `bolt/conformance/spec.yaml` **as it exists on `main` right now** (post #4885), not the version originally authored in #4883 - several `expected-fail` scenarios there have already flipped to `passing` upstream, and two new gaps (`ERR-002`, `RESULT-004`) were added. Do not "fix" a test to match the stale pre-#4885 spec.yaml.
+- Every test's `[Fact(DisplayName = ...)]` must embed its spec.yaml scenario id (e.g. `"TYPE-011: ..."`), per `bolt/conformance/README.md`'s traceability convention.
+- No new CI workflow changes - `csharp-e2e-tests` in `.github/workflows/mvn-test.yml` already runs `dotnet test` against this project.
+- Root password for every ArcadeDB container in this suite: `playwithdata`. Root user: `root`.
+- Do not commit on behalf of the user beyond what this plan's steps direct - each task ends with its own commit, per repo convention of frequent, focused commits.
+- `CLAUDE.md`: use `final` where reasonable, avoid unnecessary abstractions, no stray `System.out`/`Console.WriteLine` debug output left behind, don't add code comments explaining *what* code does (only non-obvious *why*).
+
+---
+
+## File Structure
+
+- **Modify:** `e2e-csharp/ArcadeDB.E2ETests/ArcadeDB.E2ETests.csproj` - add `Neo4j.Driver` package reference.
+- **Create:** `e2e-csharp/ArcadeDB.E2ETests/BoltFixtures.cs` - `ArcadeDbBoltFixture` (default, no-TLS) + `ArcadeDbBoltTlsRequiredFixture` + `ArcadeDbBoltTlsOptionalFixture` + their collection definitions + a shared `BoltTlsCertificates` helper for the throwaway TLS image.
+- **Create:** `e2e-csharp/ArcadeDB.E2ETests/KnownGapAssertions.cs` - the `AssertStillFailsAsync` helper.
+- **Create:** `e2e-csharp/ArcadeDB.E2ETests/BoltE2ETests.cs` - all 27 non-TLS, non-HA scenarios (connection, auth, transactions, causal-consistency, multi-database, result-handling, type-roundtrip, errors, protocol).
+- **Create:** `e2e-csharp/ArcadeDB.E2ETests/BoltTlsE2ETests.cs` - `CONN-002` and `CONN-005`.
+
+No existing file other than the `.csproj` is modified; `PostgresE2ETests.cs` and `ArcadeDbFixture.cs` are untouched.
+
+---
+
+## Task 1: Add the `Neo4j.Driver` package reference
+
+**Files:**
+- Modify: `e2e-csharp/ArcadeDB.E2ETests/ArcadeDB.E2ETests.csproj`
+
+**Interfaces:**
+- Produces: `Neo4j.Driver` NuGet package (v6.2.1) available to every subsequent task.
+
+- [ ] **Step 1: Add the package reference**
+
+In `e2e-csharp/ArcadeDB.E2ETests/ArcadeDB.E2ETests.csproj`, inside the existing `<ItemGroup>` that lists `Testcontainers`/`Npgsql`, add:
+
+```xml
+    <PackageReference Include="Neo4j.Driver" Version="6.2.1" />
+```
+
+- [ ] **Step 2: Restore and build**
+
+Run: `cd e2e-csharp/ArcadeDB.E2ETests && dotnet build`
+Expected: `Build succeeded. 0 Warning(s) 0 Error(s)`
+
+- [ ] **Step 3: Commit**
+
+```bash
+git add e2e-csharp/ArcadeDB.E2ETests/ArcadeDB.E2ETests.csproj
+git commit -m "$(cat <<'EOF'
+build(#4886): add Neo4j.Driver dependency for the e2e-csharp Bolt suite
+
+Pinned to 6.2.1 (latest stable), matching the Java suite's 6.2.0 and
+Python's neo4j>=6.2.0 baseline per bolt/conformance/spec.yaml's
+driver_version_bands.
+EOF
+)"
+```
+
+---
+
+## Task 2: `ArcadeDbBoltFixture` (default, no-TLS container)
+
+**Files:**
+- Create: `e2e-csharp/ArcadeDB.E2ETests/BoltFixtures.cs`
+
+**Interfaces:**
+- Consumes: nothing from earlier tasks besides the `Neo4j.Driver`/`Testcontainers` packages.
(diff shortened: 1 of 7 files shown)


robfrank:
Addressed the following findings from the Claude and Gemini reviews in `8b79f855b`/`9342d4394`:

- **TX-005 duplicated ERR-004** (Claude): rewrote it to exercise `session.ExecuteWriteAsync(...)`'s built-in automatic retry and assert both racing writes succeed with final values `{1, 2}` and no exception propagating — distinct now from ERR-004, which asserts the raw `TransientException` on the explicit-transaction path.
- **`KnownGapAssertions.AssertStillFailsAsync`'s bare `catch` swallowed everything** (Claude + Gemini, independently): narrowed to `XunitException or Neo4jException or InvalidCastException` — confirmed by instrumenting each of the six wrapped scenarios that this is the actual failure surface today (`TYPE-003/011/012` → `InvalidCastException`, `RESULT-004`/`ERR-002` → `EqualException`, `PROTO-002` → `TrueException`).
- **`CONN-002`'s `DisplayName` said `bolt+s://`** while the test connects via `BoltSscUri` (Claude): fixed to say `bolt+ssc://`.
- **`RunKeytool` redirected `StandardOutput` without reading it** (Gemini): a real .NET Process deadlock pattern if the pipe buffer ever filled. Removed the unread redirect.
- **`BoltTlsImage.BuildAsync`'s certDir leaked on failure** (Gemini): wrapped the whole method body in try/finally, not just the docker-build call — the initial fix only covered `image.CreateAsync()` throwing, missing that the three `RunKeytool` calls and the Dockerfile write (which also touch `certDir` and can throw) ran outside the guarded block.

Not changed, with reasoning:

- **Race-test flakiness** (Claude): `RaceTwoWritersAsync`'s `Assert.NotEmpty(errors)` doesn't structurally guarantee a collision, only makes one likely via the barrier + 500ms hold. This matches the already-merged `e2e-python` suite's identical caveat and isn't something to paper over with a retry-until-collision loop, which would weaken what the scenario is actually asserting. Watching for flakiness in CI rather than pre-emptively changing the approach.
- **Doc drift (~27 vs 37 scenarios)** (Claude): the design doc was already corrected to 37/39 during planning; not seeing remaining stale references — happy to fix a specific one if pointed at it.
- **Shared mutable `beer` DB across tests / GitHub-hosted dataset import at boot** (Claude, both Minor): both match existing conventions already used by the Java and Python Bolt suites in this repo (marker-based isolation, `OpenBeer.gz` import via `defaultDatabases`), not something unique to introduce isolation infra for here.

Full suite after these fixes: 39 total, 36 passed, 2 skipped (by design), 1 failed (`CONN-003`, a verified macOS-Docker-Desktop-specific networking limitation — `BoltNetworkExecutor.handleRoute` returns the container's own bridge IP, unreachable from the host across the Docker Desktop VM boundary; the identical scenario passes in the already-merged Python suite's CI run on `ubuntu-latest`).

robfrank:
Cycle 2 (`e9f3168`):

**Fixed:**
- `Barrier` (in `Tx005_ManagedWriteRetriesOnTransientError` and `RaceTwoWritersAsync`) is `IDisposable` and wasn't disposed - now `using var barrier = new Barrier(2);` in both.

**Considered, not changed, with reasoning:**

- **Point 2 (TYPE-003 could throw uncaught `NullReferenceException` if `.As<IPath>() `returns null):** Checked against what's actually observed - `TYPE-003` throws `InvalidCastException: String → IPath` today (confirmed via instrumented diagnostic during implementation, not assumed), because the wire value is a non-null string, not a null Path. `.As<T>()` only returns null when the underlying Cypher value itself is null; a non-null value of the wrong type throws. So the specific failure mode raised isn't currently reachable. I didn't add `NullReferenceException` to the catch filter pre-emptively: that would re-widen exactly what point 1 (and the earlier cycle-1 finding) asked to narrow - an NRE is more often a genuine bug in test code than a still-reproducing product gap, so catching it here would risk masking a real bug rather than closing a real gap.
- **Point 1 (the `Neo4jException` branch is broad):** matches the tradeoff above - it exists for cases where a gap manifests as a driver-level error rather than a wrong-shaped value (plausible for some future/different ArcadeDB build), and none of the six wrapped scenarios currently rely on it (all six throw `InvalidCastException` or an `XunitException` subclass today). Leaving it as defensive coverage for a class of failure that hasn't manifested yet, rather than removing it and having to reintroduce it later.
- **Point 3 (ERR-004/TX-005 timing-dependent, may flake in CI):** same conclusion as cycle 1 - matches the already-merged Python suite's identical, accepted caveat. Not looping-until-collision, since that would weaken what the scenario is actually asserting.
- **Point 5 (TX-005's `{1,2}` assertion doesn't itself prove a retry happened):** agreed, and already the case - the actual proof is "no exception surfaced across both racing `ExecuteWriteAsync` calls" (documented in the test's own comment); the value-equality assertion is a secondary check that no write was lost or double-applied, not the retry proof itself.
- **Point 6 (shared `beer` DB accumulates probe nodes):** same as cycle 1 - matches existing convention, all assertions are marker/count-scoped so this doesn't affect correctness.

Full suite: 39 total, 36 passed, 2 skipped (by design), 1 pre-existing `CONN-003` failure (documented macOS Docker Desktop networking limitation, unrelated to this change).

robfrank:
Cycle 3 (`f660428`):

**Fixed:**
- **TLS Dockerfile ignored `ARCADEDB_DOCKER_IMAGE`** (point 1): `BoltTlsImage.BuildAsync`'s generated Dockerfile now resolves the same env var `ArcadeDbBoltFixture` already does, instead of hardcoding `FROM arcadedata/arcadedb:latest`. Harmless in CI (both paths land on the loaded tag there), but was a real local-dev trap for anyone testing against a custom image.

**Considered, not changed, with reasoning:**

- **Point 2 (two separate TLS image builds with identical content):** agreed this is redundant build work, but sharing one base image across the two fixtures would require cross-fixture coordination (a shared, once-built dependency both collections wait on) that adds real structural complexity for a build-time optimization only. Matches your own framing as a "defensible tradeoff... possible future simplification," not implementing now.
- **Point 3 (tighter assertion for ERR-002's known-wrong code):** looked into this specifically. The reason it doesn't actually add the precision it looks like it would: `AssertStillFailsAsync`'s filter accepts `XunitException` broadly, so *any* assertion failure inside the wrapped action - whether it's checking the exact wrong code or checking something else entirely - gets caught identically. Adding a second, more specific assertion inside the wrapper doesn't change what the *filter* accepts; achieving true "only this specific known-wrong value counts as reproduced" precision would mean either string-matching on exception messages (fragile) or moving ERR-002 out of the shared wrapper entirely into a plain `[Fact]` that asserts the current wrong value directly (a bigger, six-test-inconsistent redesign for one data point, three cycles into review). Given all three review passes have independently called this class of concern "acceptable"/"non-blocking," not changing it further.
- **Point 4 (TLS wait-strategy timeout parity with Python's 120s):** checked - not a real risk here. Reflected the installed `Testcontainers` 4.11.0 assembly directly during the final whole-branch review earlier in this PR's history: `Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(...)`'s default timeout is **1 hour** in this .NET package, vastly exceeding whatever margin Python's shorter default needed the explicit 120s bump for. No adjustment needed.

Full suite: 39 total, 36 passed, 2 skipped (by design), 1 pre-existing `CONN-003` failure (same documented macOS Docker Desktop networking limitation).

robfrank:
Cycle 4 (`2b72476`):

**Fixed:**
- **PROTO-002's version parsing could throw outside the caught set** (point 1): `int.Parse(...)` and the null-forgiving `!` could throw `FormatException`/`NullReferenceException` if `ProtocolVersion` were ever null or not `"major.minor"`-shaped - neither is in `AssertStillFailsAsync`'s filter, so a genuinely malformed value would hard-fail the build instead of registering as the known gap. Converted both failure modes into `InvalidCastException` (already in the filter) via `int.TryParse` + an explicit null-check, rather than broadening the shared helper's catch.
- **`TX-005`'s `.As<int>()` inconsistency** (nit): aligned to `.As<long>()`, matching every other count/value read in this file.

**Considered, not changed, with reasoning:**

- **Point 2 (exception doesn't correlate with `reason`) and point 3 (`ERR-004`/`TX-005` timing-dependent):** same conclusions as cycles 1-3 - inherent to the xfail pattern and the race-based scenarios respectively, not repeating the reasoning again here.
- **Point 4 (wait strategy only checks HTTP 2480, not Bolt port 7687):** checked this against the already-merged Python suite's identical fixture (`e2e-python/tests/test_bolt.py`'s `wait_for_http_endpoint(container, "/api/v1/ready", 2480, 204, 60)`) - it uses the exact same HTTP-only readiness check, no separate Bolt-port probe, and has been running in this project's CI without reported plugin-bind-ordering flakiness. Not diverging from that precedent without evidence of an actual race (39 scenarios across 4 review cycles' worth of real Docker runs here haven't shown one either).
- **Point 5 (two independent TLS image builds, shared cert material could cut build time):** same as cycle 3's point 2 - a build-time optimization requiring cross-fixture coordination, called out by both reviews as minor/non-blocking.
- **RESULT-003's weak assertion (nit):** matches the already-merged Python suite's identical test shape (`next(iter(result)); summary = result.consume(); assert summary is not None`) - not diverging from that precedent for one test in isolation.
- **`RunKeytool`'s stderr-only redirect (nit):** confirmed safe as noted (no stdout redirection exists after cycle 1's fix, so there's nothing to deadlock on).

Full suite: 39 total, 36 passed, 2 skipped (by design), 1 pre-existing `CONN-003` failure (same documented macOS Docker Desktop networking limitation, unrelated to this PR's code).

This is the fourth review cycle for this PR. Across all four: 5 real findings fixed in cycle 1 (deadlock bug, missing beer-database seed, dropped error-code assertion, plus two robustness issues), 1 in cycle 2 (`Barrier` disposal), 1 in cycle 3 (`ARCADEDB_DOCKER_IMAGE` consistency), and 2 in cycle 4 (exception-safety hardening, numeric-type consistency) - all verified against a live ArcadeDB Docker container, not just compiled. Several recurring points (the `AssertStillFailsAsync` catch breadth, race-test timing-sensitivity) were independently raised by both bots across multiple cycles and evaluated each time rather than dismissed by default; where no change was made, the reasoning is recorded above and in the prior cycle comments. Ready for human review.

PR #5010, opened 2026-07-05
Update Python bindings: bulk-transport performance, embedded-only packaging, engine-sync fixes

tae898:
Periodic update of `bindings/python` from [humemai/arcadedb-embedded-python](https://github.com/humemai/arcadedb-embedded-python), where the Python bindings are developed. Scope is the minimum needed to reproduce the wheel build and run the tests and examples: `src/`, `scripts/`, `tests/`, `examples/`, `pyproject.toml`, `README.md`.

## What's new

**Java helper sources now live inside the Python bindings.** The biggest difference from previous updates: `bindings/python/src/java/com/arcadedb/python/` adds four small Java classes (`RowBatcher`, `ColumnBatcher`, `EdgeBatcher`, `VertexBatcher`, ~450 lines total) that the existing build scripts compile into an `arcadedb-python-bridge.jar` shipped inside the wheel. They batch result serialization and record creation on the Java side so bulk operations cost one JPype crossing per batch instead of several per row. They use only public engine APIs, and every Python code path falls back to pure JPype when the jar is absent.

**New bulk APIs built on the bridge:** `ResultSet.to_json_list()` / `iter_json_batches()` / `to_columns()` (typed numpy columns), a fast `to_dataframe()` path, and `GraphBatch.new_edges()` bulk edge ingest. Measured against a pure-Java baseline on identical JARs/JRE/flags: vector SQL search went from ~15x the Java time to ~1.1x, 100k-row scans from ~30x to ~1.6x, and bulk edge ingest to parity.

**Embedded-only packaging.** The wheel no longer bundles server mode (`ArcadeDBServer`/`create_server`, Studio); client-server users are pointed at the official server distribution. Together with JAR pruning and a fix to the jlink module detection in the build scripts, wheels went from ~74 MB to ~62 MB.

Suite: 352 passed, 4 skipped; full examples run green; 4-platform x 5-Python release matrix green on the fork.


robfrank (commented):
LGTM

robfrank (approved):

Issue #4962, opened 2026-07-04
[audit] Engine deep-audit 2026-07: tracking issue (storage, tx/MVCC, LSM, concurrency)


lvca:
Umbrella tracking issue for the 2026-07 engine deep-audit (storage/WAL, transactions/MVCC, LSM index, concurrency/threading, security). Each finding is filed separately and tagged [`audit`](https://github.com/ArcadeData/arcadedb/labels/audit). Security findings are tracked privately as draft GitHub Security Advisories (not linked here).

### Critical (act first)
- [x] #4935 [tx] Double-update of an indexed record in one tx corrupts the index **(reproduced on main)** - fixed in #4964
- [x] #4942 [lsm] Compaction loses a re-inserted entry (put -> remove -> re-put) - reproduced + fixed in #4968
- [x] #4948 [engine] Parallel bucket scan self-deadlocks via CallerRunsPolicy (default-on) + wipes caller DatabaseContext - fixed in #5009

### Storage / WAL
- [x] #4924 Interrupted thread skips page I/O -> cache poisoning + false flush ack - fixed in #4965
- [x] #4925 Read-cache stale-overwrite race defeats MVCC (lost update) - fixed in #5011
- [x] #4926 Torn 64KB page write + `<=` version skip loses committed deltas - fixed in #5052 (equal-version re-apply; multi-version residual tracked in #5054)
- [x] #4927 PageManager lifecycle race can null the flush thread under a live DB - fixed in #5070 (refcounted lifecycle, exactly-once release, 12 review rounds)
- [x] #4928 Flush-thread IOException drops rest of batch + hangs close() - fixed in #5012 (incl. ack-gated clean close + crash-equivalent give-up close)
- [x] #4929 Default txWalFlush=0 permits write-ahead ordering violation (design/docs) - decided: default stays 0 (process-crash safe); production mode auto-sets 1; docs now explain the mechanics + measured costs (WALFlushBenchmark); level 2 documented as no-value over 1
- [x] #4930 ClosedChannelException reopen resurrects dropped files - fixed in #5012
- [x] #4931 LocalBucket.check() int overflow in RID math (>2^31 positions) - fixed in #5012
- [x] #4932 deleteRecordInternal swallows the retry CME (orphaned chunks) - fixed in #5012
- [x] #4933 Read-cache RAM accounting double-subtract -> eviction disabled - fixed in #5011
[...]

Issue #4959, opened 2026-07-04
[tx] Low-severity transaction cleanups (silent dropped update, futile retries)


lvca:
Grouped low-severity findings in the transaction layer:

- [ ] **Silent dropped update on RecordNotFoundException at commit (`TransactionContext.java:665-673`)** - a deferred `updatedRecords` entry that throws `RecordNotFoundException` in phase 1 is logged at WARNING and skipped while the rest commits (a silent partial commit). Window is narrow (MVCC covers the concurrent-delete case), but prefer failing the tx or making it configurable.
- [ ] **transaction() retries deterministic DuplicatedKeyException (`LocalDatabase.java:1302`)** - a genuine duplicate fails identically every attempt, burning all `TX_RETRIES` + `TX_RETRY_DELAY` sleeps before surfacing. Only concurrency-induced duplicates benefit from retry; retry `DuplicatedKeyException` at most once (or only when the conflicting RID was committed after the tx began).
- [ ] **Isolation contract undocumented** - no read-set validation, so write skew and phantoms are possible under both READ_COMMITTED and REPEATABLE_READ, and READ_COMMITTED readers can observe torn multi-page commits mid-`updatePages`. Legitimate for page-level MVCC, but state it in the isolation docs and pin it with tests. REPEATABLE_READ`s `immutablePages` cache is unbounded for large scans - cap or spill note.
- [ ] **Requester identity (`executeLockingFiles`)** hardcodes `Thread.currentThread()` while locks are keyed by requester (thread or session); a thread acting for a session-requester tx can block on locks its own session holds. Unify on `TransactionContext.getRequester()`.

---
_Filed from an internal engine deep-audit (2026-07). Low-severity items grouped per subsystem; each checkbox is independently fixable._

lvca:
[...]

Issue #4940, opened 2026-07-03
[tx] Phase-2 commit failure uses reset() instead of rollback(), leaving dangling RIDs on user documents


lvca:
**Severity:** MEDIUM

**Location:** `TransactionContext.java:807-812` (`commit2ndPhase` finally calls `reset()`) vs `rollback():247-295`.

On any `commit2ndPhase` failure the `finally` calls `reset()`, which releases locks but skips two things `rollback()` does: reloading records in `modifiedRecordsCache`, and resetting the identity of `newRecords` (the #4562 fix). User-held `MutableDocument`s created in the failed tx keep an optimistically-assigned RID pointing at a record that was never persisted; re-saving them in the retry treats them as updates of a missing record. The phase-1 failure path correctly uses `rollback()`.

**Fix:** in the `finally`, call `rollback()` (or its record-restoration portion) when `!committed`.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5061 (merged). commit2ndPhase's failure handling now distinguishes durability regimes instead of calling reset() unconditionally: a failure BEFORE the WAL append rolls back user-held record state exactly like a phase-1 failure (modified records reloaded, created-record identities reset to provisional per #4562, so a retry re-inserts instead of updating a never-persisted RID); a failure AFTER a real append keeps reset() - the transaction is durable, recovery replays it, and the RIDs stay valid - composed with #5053's recovery fence in the same branch. The useWAL=false path always rolls back (nothing durable, nothing to replay). Hardened through three review rounds: the degraded path (rollback itself failing at the dictionary reload) preserves the primary exception AND still resets created-record identities. Red-first verified ('expected: null but was: #1:0' on unfixed code).

Issue #4562, opened 2026-06-11
GRPC is not rolling back deletes when a transactions fail with an exception.


mdre:
Continuing with the "This work on HTTP and fail on gRPC" saga, look at this:
The record deleted in the fist commit should not be deleted since the tx is rolledback with the excepcion.

```java
private void testUnique() {
        // testear el estado de las tx en cada etapa
        RemoteDatabase tx = this.db;
        tx.begin(Database.TRANSACTION_ISOLATION_LEVEL.REPEATABLE_READ);

        MutableVertex svt1 = tx.newVertex("SimpleVertexEx");
        String uuid1 = UUID.randomUUID().toString();
        System.out.println("UUID: " + uuid1);
        svt1.set("svex", uuid1);
        svt1.set("svuuid", uuid1);
        svt1.save();
        tx.commit();
        System.out.println("rid1: " + svt1.getIdentity());

        tx.begin(Database.TRANSACTION_ISOLATION_LEVEL.REPEATABLE_READ);
        MutableVertex svt2 = tx.newVertex("SimpleVertexEx");
        String uuid2 = UUID.randomUUID().toString();
        svt2.set("svex", uuid2);
        svt2.set("svuuid", uuid2);
        svt2.save();
        System.out.println("svt2: "+svt2.getIdentity());
        
        MutableVertex svToDelete = tx.newVertex("SimpleVertexEx");
        
        svToDelete.set("svex", "ToDelete");
        svToDelete.save();
        System.out.println("svToDelete: "+svToDelete.getIdentity());
        String toDelete = svToDelete.getIdentity().toString();
        
        tx.commit();
        
        System.out.println("rid2: " + svt2.getIdentity());
        System.out.println("toDelete: " + toDelete);

        System.out.println("Generate a duplicity");
        tx.begin(Database.TRANSACTION_ISOLATION_LEVEL.REPEATABLE_READ);
        svt2.set("svuuid", uuid1);
        svt2.save();
        
        MutableVertex svt3 = tx.newVertex("SimpleVertexEx");
        svt3.set("svex","svt3");
        svt3.save();
        System.out.println("svt3: "+svt3.getIdentity());
        
        System.out.println("delete "+toDelete);
        tx.deleteRecord(svToDelete);
        
        try {
            tx.commit();
        } catch (Exception ex) {
            System.out.println("Correcto. Atrapa la excepción de duplicidad.");
            System.out.println("Tansaction active: "+tx.isTransactionActive());
        }
[...]

PR #5061, opened 2026-07-06
fix(#4940,#4959): phase-2 pre-WAL failure rolls back record state; tx-layer cleanups

lvca:
Part of the 2026-07 engine audit remediation (#4962), tx-commit group.

### #4940 - fixed
Re-verified against current main: `commit2ndPhase`'s `finally` still calls `reset()` on every failure, skipping the two record-restoration steps `rollback()` performs (reloading `modifiedRecordsCache` and resetting the identity of `newRecords`, the #4562 fix). User-held documents created in the failed tx kept a dangling RID; re-saving them in a retry updated a missing record.

The fix distinguishes durability with a `walAppended` flag:
- failure **before** the transaction reaches the WAL -> `rollback()` (nothing durable exists; behave exactly like a phase-1 failure);
- failure **after** the WAL append -> `reset()` as before (the tx is durable, recovery replays it, so the assigned RIDs remain valid and must not be reverted).

This is deliberately shaped to compose with #5053 (point-of-no-return reordering), which adds fencing to the same post-append branch. Checked the only external split-phase caller (`RaftReplicatedDatabase`): its phase-2 failure handling reconciles from the payload's WAL bytes and steps down, independent of tx state, so the change composes there too.

### #4959 - fixed (all four checkboxes)
1. **Silent dropped update**: a deferred `updatedRecords` entry whose record was deleted by a concurrent transaction (page loaded after the delete committed, so the MVCC version check cannot see it) was WARN-logged and skipped while the rest committed. Now fails the whole tx with a retryable `ConcurrentModificationException`. An in-tx delete removes the entry from `updatedRecords` (`LocalBucket` -> `removeRecordFromCache`) and never hits this path.
2. **Futile duplicate-key retries**: `transaction()` burned all `TX_RETRIES` + `TX_RETRY_DELAY` sleeps on a deterministic `DuplicatedKeyException`. It is now retried at most once (one retry disambiguates a concurrency-induced duplicate); the second identical failure is rethrown immediately.
3. **Isolation contract**: documented on `Database.TRANSACTION_ISOLATION_LEVEL` (page-level MVCC, no read-set validation, write skew/phantoms possible under both levels, torn multi-page visibility under READ_COMMITTED, unbounded per-tx page cache under REPEATABLE_READ) and pinned with `TransactionIsolationContractTest` (write skew demonstrated under both levels).
4. **Requester identity**: `executeLockingFiles` hardcoded `Thread.currentThread()`; it now locks on behalf of the current transaction's requester (`TransactionContext.getRequester()`, widened to public), so a thread acting for a session no longer times out on locks its own session holds. Only the locks it actually acquired are released on exit.

## Test evidence (red-first)

All five fix-targeting tests fail on unfixed code:
- `Issue4940Phase2FailureRollbackTest.phase2FailureBeforeWalAppendResetsCreatedRecordIdentity`: `expected: null but was: #1:0`
- `Issue4940Phase2FailureRollbackTest.phase2FailureBeforeWalAppendReloadsModifiedRecords`: `expected: "committed" but was: "uncommitted"`
- `Issue4959TransactionCleanupsTest.deferredUpdateOfConcurrentlyDeletedRecordFailsTheCommit`: `Expecting code to raise a throwable` (commit succeeded, update silently dropped)
- `Issue4959TransactionCleanupsTest.deterministicDuplicatedKeyIsRetriedAtMostOnce`: `expected: 2 but was: 10`
- `Issue4959TransactionCleanupsTest.executeLockingFilesUsesTheTransactionRequester`: `LockTimeoutException` after the 5000ms timeout

Green after the fix. Focused suites run: `com.arcadedb.database.**` (all 38 classes incl. async), `ACIDTransactionTest`, `ExplicitLockingTransactionTest`, `LocalTransactionExplicitLockRetryTest`, `MVCCTest` - 123 tests, 0 failures, 0 errors. (`TransactionTest` and `WalCommitOrderingTest` do not exist on main; the latter arrives with #5053.)

## Conflict risks

Sibling audit branches may touch:
- `engine/src/main/java/com/arcadedb/database/TransactionContext.java` - **#5053 rewrites the same `commit2ndPhase` try/finally** (it also introduces its own `walAppended` flag; the merge resolution is: keep #5053's fencing in the post-append branch and this PR's `rollback()` in the pre-append branch)
- `engine/src/main/java/com/arcadedb/database/LocalDatabase.java`
- `docs/release-26.7.2.md` (expected, entries are additive)

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -208,6 +208,24 @@ that could starve the very snapshot resync meant to heal the node.
   shadows committed RIDs whose key equals an uncommitted entry's key - is tracked in
   [#5055](https://github.com/ArcadeData/arcadedb/issues/5055).
 
+- **Transaction commit cleanups (2026-07 audit).** A phase-2 commit failure that happens BEFORE the
+  transaction reaches the WAL now restores user-held record state like a phase-1 failure does (rollback):
+  records created in the failed transaction get their optimistically-assigned RID reset to provisional and
+  modified records are reloaded to their committed content, so retrying with the same in-memory objects
+  re-inserts them instead of updating a record that was never persisted; a failure after the WAL append
+  still keeps the assigned RIDs, since the transaction is durable and recovery replays it
+  ([#4940](https://github.com/ArcadeData/arcadedb/issues/4940)). From the low-severity group
+  ([#4959](https://github.com/ArcadeData/arcadedb/issues/4959)): a deferred update whose record was deleted
+  by a concurrent transaction now fails the commit with a retryable `ConcurrentModificationException`
+  instead of being silently skipped while the rest of the transaction commits (silent partial commit);
+  `transaction()` no longer burns every retry attempt (plus retry delays) on a deterministic
+  `DuplicatedKeyException` - one retry disambiguates it from a concurrency-induced duplicate;
+  `executeLockingFiles` now locks on behalf of the current transaction's requester (thread or session), so
+  a thread acting for a session no longer times out on locks its own session already holds; and the
+  page-level MVCC isolation contract (no read-set validation: write skew and phantoms possible under both
+  levels, unbounded per-transaction page cache under `REPEATABLE_READ`) is now documented on
+  `Database.TRANSACTION_ISOLATION_LEVEL` and pinned by tests.
+
 ### Improvements
 
 - **HA: throttled diverged-follower resync logging.** When a follower detects a WAL page-version gap it
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -238,7 +238,10 @@ that could starve the very snapshot resync meant to heal the node.
   by a concurrent transaction now fails the commit with a retryable `ConcurrentModificationException`
   instead of being silently skipped while the rest of the transaction commits (silent partial commit);
   `transaction()` no longer burns every retry attempt (plus retry delays) on a deterministic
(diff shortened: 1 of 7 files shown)


lvca:
Cross-verification pass from the parallel audit run added commit b7ae2420e with two corrections to the #4940 fix:

1. **`walAppended` was set even with `useWAL=false`** (`changes.result == null`): a publish failure in a no-WAL transaction was treated as "durable, recovery replays it" and only `reset()` ran - but no WAL record exists to replay, so the #4940 defect (dangling RIDs on user-held documents) survived on that path. The flag is now set only after a real append. This also matches the point-of-no-return boundary of PR #5053: when that lands, this placement keeps its recovery fencing from triggering for a useWAL=false failure that recovery could never repair. Red-first evidence: the new `phase2PublishFailureWithoutWalStillRollsBack` test fails on the previous placement (`expected: null`, the created record kept its dangling RID) and passes now.

2. **The `changes == null` no-op call** was left inside the try, so it fell into the new `rollback()` branch instead of the pre-#4940 `reset()`. Restored as an early return with `reset()`.

Focused suites re-run green after the change: `com.arcadedb.database.**` (142), `ACIDTransactionTest`, `ExplicitLockingTransactionTest`, `TransactionIsolationTest`, `TransactionIsolationContractTest`, `TransactionBucketTest`, `TransactionTypeTest`.

Issue #4961, opened 2026-07-04
[engine] Low-severity concurrency cleanups (async field visibility, rejection-handler task drop)


lvca:
Grouped low-severity findings in the concurrency/pool layer:

- [ ] **DatabaseAsyncExecutorImpl mutable config fields (`:65-70,88-90`)** - `parallelLevel`, `commitEvery`, `backPressurePercentage`, `checkForStalledQueuesMaxDelay`, `onOkCallback`, `onErrorCallback` are non-volatile, written by user threads and read by workers; changes after `createThreads` may never become visible. Also `setCommitEvery(0)` makes `count % commitEvery` throw every task. Make them volatile and validate `commitEvery >= 1`.
- [ ] **Rejection handler drops tasks after shutdown (`QueryEngineManager.java:114-115`, `SparseVectorScoringPool.java:142-143`)** - `if (!executor.isShutdown()) task.run();` means a task submitted around shutdown is neither run nor completed exceptionally; any caller in an untimed `Future.get()` hangs forever. Complete the task exceptionally, or require quiescence before `close()` + add timeouts to all awaits.
- [ ] **QueryEngineManager.register() (`:161-163`)** mutates a `LinkedHashMap` that concurrent `getEngine`/`getAvailableLanguages` readers iterate; safe today (constructor-only) but the method is public - guard or document.

---
_Filed from an internal engine deep-audit (2026-07). Low-severity items grouped per subsystem; each checkbox is independently fixable._

lvca:
Fixed in #5062 (merged): the async executor's configuration fields (parallelLevel, commitEvery, backPressurePercentage, transactionUseWAL, transactionSync, checkForStalledQueuesMaxDelay and friends) have safe cross-thread visibility, and the rejection-handler task-drop is closed alongside the #4954 shutdown-draining work. Part of the six-round review-hardened async batch with red-first coverage.

Issue #4954, opened 2026-07-04
[engine] Async executor shutdown/interrupt drops queued tasks, permanently hanging completion waiters


lvca:
**Severity:** MEDIUM

**Location:** `database/async/DatabaseAsyncExecutorImpl.java:182-185` (`InterruptedException` handler does `queue.clear()` and exits), `:401-409` (`scanType` awaits a latch with no timeout), `:316-355` (`waitCompletion` default untimed), `:789-791` (close path `queue.put(FORCE_EXIT)` while holding `lifecycleLock`).

`kill()` interrupts workers; a worker`s `queue.clear()` discards queued scan/completion tasks whose `completed()` never runs, so any thread in `scanType()`/`waitCompletion()` hangs forever. Tasks enqueued after `FORCE_EXIT` are silently dropped. `shutdownThreadsLocked` blocks on an untimed `queue.put(FORCE_EXIT)` under `lifecycleLock`; a wedged worker with a full queue hangs `database.close()` and any later `createThreads`/`kill`.

**Fix:** on interrupt, drain the queue calling `completed()` on each task instead of `clear()`; use `offer`+timeout for `FORCE_EXIT` and interrupt the worker on failure; count down scan latches from a `finally` when scheduling fails.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5062 (merged). Shutdown/interrupt no longer drops queued tasks: the InterruptedException handler drains instead of queue.clear() (whose dropped tasks never ran completed(), hanging scanType()/waitCompletion() forever), tasks offered after a dead worker's final drain are removed-and-thrown loudly (best-effort with the opt-in 'fast' Conversant queue, which lacks remove() - documented, WARNING logged, tracked in #5066), waitCompletion(timeout) treats the timeout as a single budget across enqueue+await (previously it could block ~15s on a 700ms timeout; all production callers verified to use the no-arg form), and close() escalates to interrupt+re-join when a worker consumed FORCE_EXIT inside the helping loop and stayed alive past the grace period. Red-first regression tests for each path.

Issue #4953, opened 2026-07-04
[engine] Async executor stall detector false-positives on slow tasks and rolls back the whole in-flight batch


lvca:
**Severity:** MEDIUM (silent loss of already-executed async ops)

**Location:** `database/async/DatabaseAsyncExecutorImpl.java:550-563` (cross-slot edge task schedules a follow-up on another slot from inside the worker), `:853-873` (stall detection), `AsyncThread.run:167-170` (onError -> `database.rollback()`).

Under bidirectional edge load with full queues, slot A`s worker blocks offering into slot B`s queue while B blocks offering into A`s - a real wait cycle. The detector breaks it by throwing `DatabaseOperationException` after `checkForStalledQueuesMaxDelay` (5s) if `queue.peek()` returned the identical head twice. Problems: (a) any head task running > 5s (a heavy `DatabaseAsyncCommand` query) makes an innocent producer throw "queue is stalled"; (b) the failure rolls back the whole in-flight commit batch (up to `commitEvery-1` already-executed ops silently discarded unless the user wired `onError`).

**Fix:** track per-worker monotonic completed-task counters instead of head-identity peeking; make the cross-slot follow-up a non-blocking `offer` with retry-on-own-queue so workers never sleep on each other`s queues.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5062 (merged). The stall detector no longer false-positives on a worker merely busy with one slow task: detection is progress-based (per-worker completed-task counters) instead of comparing queue-head identity across two 5s windows. Cross-scheduling workers no longer deadlock either: a worker whose cross-slot follow-up meets a full peer queue drains its OWN queue while waiting, with the polled tasks parked and run after the current task unwinds (park-then-run, preserving per-task atomicity - the original re-entrant helping broke transaction boundaries and was redesigned in review). Producers blocked on a genuinely wedged worker still fail loudly via a progress-gated backstop (12 zero-progress windows, ~60s default; cross-slot parks report faster at 3 windows), both bounds tunable via setCheckForStalledQueuesMaxDelay and documented for operators. Six review rounds; every fix red-first tested. Residual: the help-waiting loop's own empty-queue liveness bound lands as follow-up PR from the round-6 review.

PR #5062, opened 2026-07-06
fix(#4953,#4954,#4961): async executor stall detector, shutdown draining, concurrency cleanups

lvca:
Part of the 2026-07 engine audit remediation (#4962), group **async-executor**.

### #4953 - stall detector false-positives and rolls back the in-flight batch: **fixed**
Re-verified against current main: `scheduleTask` still peeked the queue head across two `checkForStalledQueuesMaxDelay` windows and threw on identical identity, and `newEdge`'s cross-slot incoming-edge follow-up still blocked one worker on another worker's full queue (a real wait cycle the detector broke by throwing inside the worker, rolling back up to `commitEvery - 1` already-executed ops and dropping the follow-up).

Fix (both parts, coherently in `scheduleTask`):
- **Help-while-waiting (park-then-run)**: a worker of this executor that must wait on a full queue drains its OWN queue in the meantime, but does NOT execute the polled tasks re-entrantly - a nested execution could commit or roll back the suspended task's partial writes, breaking per-task atomicity (this replaced an earlier re-entrant design during review). Polled tasks are parked in a thread-confined `helpDeferredTasks` deque and executed by the run loop, in poll order, once the current task fully unwinds; draining alone frees the slot the parked peer is waiting on, so cross-slot cycles resolve with every task executed, no exception, no rollback. The parked backlog is bounded by the worker's queue capacity; past that budget the worker falls back to the detecting bounded wait below.
- **Progress-based detection for external producers**: per-worker monotonic `completedTaskCount` + a `waitingCrossSlotOffer` flag, with two thresholds. A target parked handing a task cross-slot with a flat completed count (a likely scheduling cycle, only reachable after budget exhaustion) surfaces after 3 consecutive no-progress windows (15s by default), so a peer merely busy on one slow task does not trip it; a target wedged inside user code (infinite loop / blocking call, flag never set) surfaces after 12 windows (60s by default) instead of hanging the producer forever. Both tunable via `setCheckForStalledQueuesMaxDelay()`. A worker merely busy on a slow task no longer trips anything.

### #4954 - shutdown/interrupt drops queued tasks, hanging completion waiters: **fixed**
Re-verified: the interrupt handler still did `queue.clear()` (dropping tasks whose `completed()` never ran, hanging `scanType()`/`waitCompletion()`), and `shutdownThreadsLocked` still blocked on an untimed `queue.put(FORCE_EXIT)` under `lifecycleLock` (a busy worker with a full queue wedged `database.close()`).

Fix: every run-loop exit path now drains the queue (and the parked help-deferral list) notifying `completed()` on each leftover task; `DatabaseAsyncScanBucket` / `DatabaseAsyncBrowseIterator` count their latches down from `completed()` (invoked on execution AND drop) instead of inside `execute()`; shutdown offers `FORCE_EXIT` with a 1s timeout and interrupts the worker on failure (with a WARNING if it will not die within 10s); `waitCompletion(timeout)` treats the timeout as a single budget spanning the enqueue and await phases, with a worker-liveness check instead of an untimed `put`; `scheduleTask` undoes an offer that lands after the target worker's final drain (best-effort: documented residual races for the opt-in `fast` queue, which cannot remove, and for an offer landing before `isAlive()` flips; the former logs a WARNING).

### #4961 - low-severity concurrency cleanups: **fixed** (all three checkboxes)
- Async executor settings (`parallelLevel`, `commitEvery`, `backPressurePercentage`, `transactionUseWAL`, `transactionSync`, `checkForStalledQueuesMaxDelay`, onOk/onError callbacks) are now `volatile`; `setCommitEvery` rejects values < 1 (0 made every task fail on `count % 0`) and the constructor clamps the configured batch size.
- `QueryEngineManager` and `SparseVectorScoringPool` rejection handlers now cancel a `RunnableFuture` rejected after pool shutdown instead of leaving its waiter blocked in an untimed `Future.get()` forever (no test: both pools are JVM-wide singletons that are never shut down inside a test JVM without breaking sibling tests; the change is the minimal safe branch).
- `QueryEngineManager.register()` is `synchronized` and publishes a copy-on-write map, so the public post-construction registration path cannot corrupt the lock-free `getEngine`/`getAvailableLanguages` readers.

## Test evidence (red-first)
All nine new regression tests fail on unfixed code exactly on the audited defect, and pass after the fix:
- `AsyncStallDetectorFalsePositiveTest`: unfixed threw `DatabaseOperationException: Asynchronous queue 0 is stalled` for a healthy worker busy 1.2s.
- `AsyncCrossSlotSchedulingDeadlockTest`: deterministic 2-worker cross-offer cycle (queue capacity 1); unfixed dropped both follow-up markers (latch timeout) and fired the stall exception.
- `AsyncHelpingAtomicityTest`: deterministic ordering probe; with re-entrant helping the polled task runs (and with `commitEvery=1` commits the suspended task's partial write) before the suspended task unwinds.
- `AsyncWedgedWorkerStallTest`: unfixed let a producer hang for the full wedge duration on a worker blocked in user code; fixed throws at the 12-window backstop.
- `AsyncSlowPeerNoFalseStallTest`: unfixed threw the stall exception after a single flat window at a producer targeting a budget-exhausted worker whose cross-slot peer was merely slow; fixed waits 3 windows and completes normally.
- `AsyncShutdownDrainTest.killNotifiesCompletionOfDroppedQueuedTasks`: unfixed never called `completed()` on the queued probes (latch timeout).
- `AsyncShutdownDrainTest.closeDoesNotBlockOnFullQueueOfBusyWorker`: unfixed `close()` blocked ~15s in `put(FORCE_EXIT)` (await(8s) failed).
- `AsyncWaitCompletionTimeoutTest`: unfixed `waitCompletion(700)` blocked far past its timeout while the enqueue phase retried on a full queue.
- `AsyncConfigValidationTest`: unfixed accepted `setCommitEvery(0)`.

Focused suites run green: `com.arcadedb.database.async.*` (19 tests), `AsyncTest`, `SelectExecutionTest`, `QueryEngineManagerLanguagesTest`, `QueryEngineManagerPoolTest`, `GraphBatchTest`, `GraphBatchUniqueIndexTest`, `GraphBatchCommitRetryTest`, `ACIDTransactionTest`, `MVCCTest`.

## Conflict risks with sibling audit branches
- `docs/release-26.7.2.md` (expected: every audit group appends here)
- `engine/src/main/java/com/arcadedb/query/QueryEngineManager.java` (concurrency/pool layer; touched for the #4961 rejection-handler and register() items)
- `engine/src/main/java/com/arcadedb/index/sparsevector/SparseVectorScoringPool.java` (same reason)

All other changes are confined to `com.arcadedb.database.async`.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -238,7 +238,15 @@ that could starve the very snapshot resync meant to heal the node.
   preserved), so two workers cross-scheduling into each other no longer deadlock - a cycle the
   old detector could only break by throwing inside the worker, rolling back the whole in-flight commit
   batch (up to `commitEvery - 1` already-executed operations silently discarded) and dropping the
-  follow-up (ghost half-edges) ([#4953](https://github.com/ArcadeData/arcadedb/issues/4953)). Shutdown no
+  follow-up (ghost half-edges) ([#4953](https://github.com/ArcadeData/arcadedb/issues/4953)). A
+  producer blocked on the full queue of a worker wedged inside user code (an infinite loop or a
+  blocking call in a task) still surfaces an error instead of hanging: a progress-gated backstop
+  throws after 12 consecutive stall windows with zero completed tasks (60s with the default
+  `checkForStalledQueuesMaxDelay` of 5s) - much longer than the old 10s false-positive-prone bound,
+  tunable via `setCheckForStalledQueuesMaxDelay()`. Known residual gap: with the opt-in
+  `arcadedb.asyncOperationsQueueImpl=fast` queue, a task whose enqueue races the target worker's
+  exit cannot be removed (`remove(Object)` unsupported) and its completion is never notified; a
+  WARNING is logged when this happens, and the default `standard` queue is not affected. Shutdown no
   longer strands completion waiters: an interrupted or exiting worker now notifies `completed()` on every
   task left in its queue instead of `queue.clear()`-ing them (threads blocked in `scanType()` /
   `waitCompletion()` hung forever), and `close()` no longer blocks unbounded on `queue.put(FORCE_EXIT)`
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -245,7 +245,10 @@ that could starve the very snapshot resync meant to heal the node.
   `checkForStalledQueuesMaxDelay` of 5s) - much longer than the old 10s false-positive-prone bound,
   tunable via `setCheckForStalledQueuesMaxDelay()`; a worker parked handing a task cross-slot with a
   flat completed count (a likely scheduling cycle) is reported faster, after 3 consecutive stall
-  windows (15s by default), so a peer merely busy on one slow task does not trip it. Known residual
+  windows (15s by default), so a peer merely busy on one slow task does not trip it. Operators
+  running scheduling chains deeper than two workers with individual tasks slower than 15s should
+  raise `checkForStalledQueuesMaxDelay` (only the window duration is tunable, the counts are fixed),
+  or the cross-slot detector can fire on a chain that would have resolved. Known residual
   gap: with the opt-in
   `arcadedb.asyncOperationsQueueImpl=fast` queue, a task whose enqueue races the target worker's
   exit cannot be removed (`remove(Object)` unsupported) and its completion is never notified; a
@@ -253,8 +256,11 @@ that could starve the very snapshot resync meant to heal the node.
   longer strands completion waiters: an interrupted or exiting worker now notifies `completed()` on every
   task left in its queue instead of `queue.clear()`-ing them (threads blocked in `scanType()` /
   `waitCompletion()` hung forever), and `close()` no longer blocks unbounded on `queue.put(FORCE_EXIT)`
-  under the lifecycle lock when a busy worker's queue is full - the marker is offered with a timeout and
-  the worker is interrupted on failure ([#4954](https://github.com/ArcadeData/arcadedb/issues/4954)).
+  under the lifecycle lock when a busy worker's queue is full - the marker is offered with a timeout,
+  the worker is interrupted on failure, and a worker still alive after the 10s grace period (e.g. one
+  that consumed the marker while help-waiting on a wedged peer's full queue) is escalated to an
+  interrupt and re-joined instead of being left running behind a returned `close()`
+  ([#4954](https://github.com/ArcadeData/arcadedb/issues/4954)).
   Low-severity cleanups from the same audit: async executor settings (`parallelLevel`, `commitEvery`,
   back-pressure, callbacks) are now `volatile` so post-startup changes reach the workers,
   `setCommitEvery(0)` is rejected instead of making every task fail on `count % 0`, the JVM-wide query
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -208,6 +208,29 @@ that could starve the very snapshot resync meant to heal the node.
   shadows committed RIDs whose key equals an uncommitted entry's key - is tracked in
   [#5055](https://github.com/ArcadeData/arcadedb/issues/5055).
 
+- **Async executor hardening (2026-07 audit).** Three fixes to the per-database async executor. The
+  queue stall detector no longer false-positives on a worker merely busy with one slow task (it compared
+  the identity of the queue head across two 5s windows, so any head task running longer than 10s made
+  innocent producers throw "Asynchronous queue is stalled"); detection is now progress-based (per-worker
+  completed-task counters) and, more importantly, a worker that schedules a cross-slot follow-up (the
+  bidirectional-edge incoming-link task) into another worker's full queue now drains and executes its OWN
+  queue while it waits, so two workers cross-scheduling into each other no longer deadlock - a cycle the
+  old detector could only break by throwing inside the worker, rolling back the whole in-flight commit
+  batch (up to `commitEvery - 1` already-executed operations silently discarded) and dropping the
+  follow-up (ghost half-edges) ([#4953](https://github.com/ArcadeData/arcadedb/issues/4953)). Shutdown no
+  longer strands completion waiters: an interrupted or exiting worker now notifies `completed()` on every
+  task left in its queue instead of `queue.clear()`-ing them (threads blocked in `scanType()` /
+  `waitCompletion()` hung forever), and `close()` no longer blocks unbounded on `queue.put(FORCE_EXIT)`
+  under the lifecycle lock when a busy worker's queue is full - the marker is offered with a timeout and
+  the worker is interrupted on failure ([#4954](https://github.com/ArcadeData/arcadedb/issues/4954)).
+  Low-severity cleanups from the same audit: async executor settings (`parallelLevel`, `commitEvery`,
+  back-pressure, callbacks) are now `volatile` so post-startup changes reach the workers,
+  `setCommitEvery(0)` is rejected instead of making every task fail on `count % 0`, the JVM-wide query
+  parallelism and sparse-vector pools cancel a `Future` rejected after pool shutdown instead of leaving
+  its waiter blocked forever, and `QueryEngineManager.register()` publishes a copy-on-write map so a
+  post-construction registration cannot corrupt concurrent readers
+  ([#4961](https://github.com/ArcadeData/arcadedb/issues/4961)).
+
 ### Improvements
 
 - **HA: throttled diverged-follower resync logging.** When a follower detects a WAL page-version gap it
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -232,8 +232,10 @@ that could starve the very snapshot resync meant to heal the node.
   the identity of the queue head across two 5s windows, so any head task running longer than 10s made
   innocent producers throw "Asynchronous queue is stalled"); detection is now progress-based (per-worker
   completed-task counters) and, more importantly, a worker that schedules a cross-slot follow-up (the
-  bidirectional-edge incoming-link task) into another worker's full queue now drains and executes its OWN
-  queue while it waits, so two workers cross-scheduling into each other no longer deadlock - a cycle the
+  bidirectional-edge incoming-link task) into another worker's full queue now drains its OWN
+  queue while it waits (the polled tasks are parked and run once the current task unwinds, so no
+  transaction boundary can fall inside a suspended task's execution and per-task atomicity is
+  preserved), so two workers cross-scheduling into each other no longer deadlock - a cycle the
   old detector could only break by throwing inside the worker, rolling back the whole in-flight commit
   batch (up to `commitEvery - 1` already-executed operations silently discarded) and dropping the
   follow-up (ghost half-edges) ([#4953](https://github.com/ArcadeData/arcadedb/issues/4953)). Shutdown no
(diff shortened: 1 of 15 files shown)


lvca:
Thanks for the thorough review - point 1 was a genuine catch and turned out to be broader than stated. Addressed in c5d7986eb, per point:

**1. Transaction-boundary interleaving in help-while-waiting - taken (and it was worse than reported).** Verified: besides the `count % commitEvery` boundary, a helped `DatabaseAsyncTransaction`, `DatabaseAsyncCompletion` or `DatabaseAsyncIndexCompaction` commits any active transaction at entry, and helped CRUD tasks roll back the shared transaction on their internal error path - each of those could commit or destroy the suspended outer task's partial writes. Rather than patching the one commit site, `offerHelping` no longer executes polled tasks re-entrantly at all: they are parked in a thread-confined `helpDeferredTasks` deque and the run loop executes them, in poll order, once the current task fully unwinds. Draining alone frees the slot the peer is parked on, so the #4953 cycle-resolution property is unchanged (the cross-slot regression test still passes). The parked backlog is capped at the queue capacity; past the budget the worker falls back to the bounded wait with stall detection, restoring producer backpressure. `executeTask` additionally guards the batch commit and the rollback with `!nested` as defense-in-depth, with the comment you asked for. Red-first regression test: `AsyncHelpingAtomicityTest` (fails on the previous code with the partial write committed mid-task). A side effect: `helpingDepth`/`MAX_HELPING_DEPTH` became provably dead (nothing nests anymore) and were removed.

**2. No liveness backstop in offerHelping - intentional, now documented.** A detector here would re-introduce the #4953 false positive in a new spot: consider A offering into B's full queue while B is parked cross-slot on a slow-but-healthy C - B shows a flat completed count and `waitingCrossSlotOffer == true`, so a progress-based detector would make A throw, roll back its in-flight batch and silently drop the follow-up, which is the exact defect this PR fixes. A worker wedged forever in user code is indistinguishable from one busy on a legitimately slow task, so any bound is a guess that fires on healthy load first. The trade-off is now spelled out in the `offerHelping` javadoc, and the new deferral budget from point 1 gives the loop a bounded failover into the detecting wait for cycles beyond it.

**3. waitCompletion(timeout) enqueue phase - taken.** The timeout is now a single budget spanning the enqueue and await phases. Red-first regression test: `AsyncWaitCompletionTimeoutTest`.

**4. removeQuietly no-op on the fast queue - accepted as a documented gap.** Comment added. Draining a live worker's queue from a foreign thread would race the worker itself, so the honest options are the comment or dropping `PushPullBlockingQueue` support; went with the comment since the impl is opt-in and the window is an offer racing the worker's exit.

**5. FORCE_EXIT consumed while helping - by design, comment extended** to state that the `forceShutdown` flag (not the marker) drives the run-loop exit, so `shutdownThreadsLocked`'s bounded join still completes.

**Coverage note on `QueryEngineManager.register()`:** acknowledged, left untested. The constructor is private and each instance owns a live `ThreadPoolExecutor`, so an isolated unit test needs reflection or permanently pollutes the JVM-wide singleton's language list for sibling tests in the same fork.

**Nit (@Tag("slow")):** measured rather than tagged - the long sleeps are interrupted by the very code under test, so the suites stay fast: `AsyncShutdownDrainTest` (both methods, including `closeDoesNotBlockOnFullQueueOfBusyWorker`) 1.03s total, `AsyncStallDetectorFalsePositiveTest` 1.47s, `AsyncCrossSlotSchedulingDeadlockTest` <0.1s. None crosses the multi-second bar CLAUDE.md sets for the tag.

Verification: `mvn -pl engine compile`, the full `com.arcadedb.database.async` battery (17 tests, includes the two new red-first regressions) plus `AsyncTest`, `ACIDTransactionTest` and `MVCCTest` - all green.

lvca:
Round 2 addressed in b075b42e9, per point:

**1. Lost producer-side backstop for a wedged worker - taken, the claim is correct.** The round-1 detector required `target.waitingCrossSlotOffer`, so a worker wedged inside a user `execute()`/callback never tripped it and the producer hung forever in the offer loop. `offerWaiting` now keeps the one-window cross-slot-park fast path and adds a progress-gated circuit-breaker: after `STALLED_NO_PROGRESS_WINDOWS` (12) consecutive stall windows in which the target completed zero tasks while its queue stayed full, the producer throws with a message naming the likely cause (60s with the default 5s window, tunable via `setCheckForStalledQueuesMaxDelay()`). That is 30x the old 2-window head-identity bound and gated on actual progress, so the #4953 single-slow-task false positive stays fixed. Red-first regression test: `AsyncWedgedWorkerStallTest` (on the previous code it hung for the full wedge duration with no exception). One knock-on: `AsyncStallDetectorFalsePositiveTest` moved from a 100ms to a 300ms window - its 1.2s busy task would otherwise have sat exactly on the 12x100ms backstop boundary; at 300ms it keeps a 3x margin below the backstop while remaining 4x over the old false-positive bound. Release notes document the new behavior.

**2. Fast-queue removeQuietly residual - confirmed acceptable, no longer silent.** Worker-side closing is not achievable with the `BlockingQueue` API (no drain loop can atomically prevent a late offer from landing after the final poll), and producer-side draining of the dead worker's queue is unsafe on the SPSC-oriented `PushPullBlockingQueue`. Took your fallback suggestion: the catch now logs a WARNING naming the stranded task and pointing at the `standard` queue impl, and the release notes document the residual window (opt-in impl, offer-vs-exit race).

**3. Test tagging - declined with measurements.** The long sleeps are failure ceilings, not elapsed time: the code under test interrupts them, so the suites measure `AsyncShutdownDrainTest` 1.03s for both methods (including `closeDoesNotBlockOnFullQueueOfBusyWorker`), the stall-detector test ~1.5s, everything else under 1s. None reaches the multi-second bar CLAUDE.md sets for `@Tag("slow")`, and tagging them out of default CI would exempt exactly the regression guards for these hang classes.

**4. Timing-based absence assertion - taken.** The 800ms sleep in `AsyncHelpingAtomicityTest` is gone. The test now pins the interleaving deterministically: it waits for the aggregate queue size to drop (the observable moment the helping loop polls the probe out of the worker's own queue), and the probe itself records whether it ran before the suspended task's `completed()` fired - a pure ordering fact on the worker thread, no wall clock involved. Validated both ways: with inline execution temporarily reinjected the probe fails in 0.18s; the fixed code passes. The test dropped from ~840ms to ~40ms.

**Nit (1ms helping window):** comment added marking the tight spin as intentional (every iteration frees a slot the parked peer may be waiting on) and bounded by the deferral budget.

Verification: `mvn -pl engine compile`; `com.arcadedb.database.async` battery (18 tests, including the new red-first wedged-worker regression) plus `AsyncTest` (11) - all green.

lvca:
Round 3 addressed in df580577a, per point:

**1. Single-window false positive on a budget-exhausted worker - taken, claim verified.** The `waitingCrossSlotOffer` branch indeed threw after one flat window, and the interleaving is real: a worker that exhausted its help-deferral budget sits in the bounded wait with the flag set and a flat completed count while its peer may be merely slow. It is also deterministically injectable, since the deferral budget equals the queue capacity (1 in the test config): red-first regression test `AsyncSlowPeerNoFalseStallTest`, where the previous code threw `"Asynchronous queue 0 is stalled"` at a producer whose peer was 200ms from release. The branch is now gated on `STALLED_CROSS_SLOT_NO_PROGRESS_WINDOWS = 3` consecutive no-progress windows (15s by default). Deliberately 3 rather than mirroring the wedged bound of 12: the cross-slot-park state is only reachable after budget exhaustion and carries a genuine deadlock-cycle risk the wedged state does not, so real capped cycles should still fail loudly in bounded time - but the bound must exceed 2, the old head-identity threshold that a peer busy on one slow task can flatten, or the #4953 false positive recurs. Release notes updated with the dual thresholds.

**2. Post-offer orphan race - confirmed inherent, documented.** An offer landing after the final `drainQueueNotifyingWaiters()` poll but before `isAlive()` flips passes the recheck; closing it would require a lock on the scheduling hot path. One-line comment added at the recheck, parallel to the `fast`-queue note in `removeQuietly`.

**3. `helpDeferredTasks` footprint - documented, behavior unchanged.** Field comment now states the bound: at most `queueCapacity` task references per worker (`ASYNC_OPERATIONS_QUEUE_SIZE / parallelLevel`, so one extra queue's worth in the worst case), a transient spike under sustained cross-slot pressure that drains at the next unwind.

**4. PR description drift - fixed.** The body no longer describes the abandoned re-entrant design ("recursion capped at 32" is gone); it now documents the shipped park-then-run helping, the deferral budget, the dual-threshold detector, and lists all nine red-first regression tests.

On the "things that are right" section: all accurate, no corrections - including the note that the `nested` guard in `executeTask` is dead today and kept as documented defense-in-depth. On the optional coverage gaps: (a) is now covered by `AsyncSlowPeerNoFalseStallTest`; (b) the `fast`-queue WARNING path stays untested - agreed it is hard to make deterministic, as it requires losing a race against thread termination on an impl whose queue cannot be inspected safely from a second consumer.

Verification: `mvn -pl engine compile`; `com.arcadedb.database.async` battery (19 tests, including the new red-first slow-peer regression) plus `AsyncTest` (11) - all green.

lvca:
Round 4 addressed in 213dca8bb, per point:

**1. close() returning with a live worker - taken, claim verified.** `interrupt()` was indeed only sent when the `offer(FORCE_EXIT, 1s)` failed, and a worker parked in `offerHelping` consumes the marker (sets `forceShutdown`, keeps looping to hand off the follow-up), so on a wedged peer's full queue the flag went unchecked until the stall backstop (~60s) while `join` timed out with just a WARNING. `shutdownThreadsLocked` now escalates exactly as suggested: still alive after the grace period -> `interrupt()` + re-join before the WARNING. Interrupt handling at the parks was re-verified: the offer park throws `InterruptedException`, `scheduleTask` converts it to a loud `DatabaseOperationException` (onError), the run loop exits on the flag and the drain notifies `completed()` on leftovers - and since the escalation fires only after the full grace period, healthy shutdowns never see it (the existing shutdown suite is byte-identical in behavior). The 10s grace became the package-private `shutdownJoinTimeoutMs` so the regression test does not sit through production timeouts. Red-first test: `AsyncShutdownEscalationTest`. One instructive detail: the wedged peer in the test deliberately SWALLOWS interrupts, because a peer that honors them is interrupted via the existing offer-failed path, drains its queue on exit and thereby frees the helping worker - the first test draft passed on unfixed code for exactly that reason. The interrupt-swallowing peer is also the realistic shape of the wedge (user code catching `InterruptedException` and continuing).

**2. Deep-chain false positive on the 3-window cross-slot bound - acknowledged, documented for operators.** Note added to the `setCheckForStalledQueuesMaxDelay` javadoc and the release notes: chains deeper than two workers whose tail runs individual tasks longer than 3 windows (15s by default) can still trip the cross-slot detector; raise the window duration for such workloads.

**3. waitCompletion(timeout) semantics vs existing callers - verified safe.** Full sweep across engine, server, console, integration and tests: every production caller (`LocalDatabase`, `GraphBatch` x2, `ManualIndexBuilder`, `TypeIndexBuilder`, CSV/XML importers) uses the no-arg form, which maps to an infinite budget and behaves exactly as before. Every caller passing a timeout is a test using a generous bound (3s-30s) that relies only on "true means fully flushed" - unchanged - and the few that can receive `false` handle it explicitly. No caller relies on the completion marker always being enqueued.

**4. Throw-path completed() asymmetry - kept as-is, now commented at the site:** unlike the shutdown drain (where the scheduling caller is long gone), here the caller is still on the stack and the exception informs it directly, so no waiter can be parked on the removed task yet.

**Nit:** the constants block now states explicitly that only the window duration is tunable (`setCheckForStalledQueuesMaxDelay()`); the window counts are compile-time fixed.

Verification: `mvn -pl engine compile`; `com.arcadedb.database.async` battery (20 tests, including the new red-first escalation regression) plus `AsyncTest` (11) - all green.

Issue #4955, opened 2026-07-04
[engine] Async getBestSlot NPE races database shutdown instead of throwing the intended DatabaseOperationException


lvca:
**Severity:** MEDIUM

**Location:** `database/async/DatabaseAsyncExecutorImpl.java:283-301` (`getBestSlot` reads `executorThreads.length` with no null check) while `close()`/`kill()` null it under `lifecycleLock`. Callers: `compact():275-278`, `scheduleTask(slot=-1):836`. The class comment (`:60-63`) claims readers snapshot the array so they cannot observe a just-nulled array, but `getBestSlot` snapshots and never null-checks. A compaction scheduled concurrently with `close()` throws NPE instead of `DatabaseOperationException("Async executor has been shut down")`.

**Fix:** null-check the snapshot in `getBestSlot`/`getRandomSlot` and throw the same `DatabaseOperationException` as `scheduleTask`.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5012 (merged). getBestSlot/getRandomSlot now null-check (and zero-length-check) their executorThreads snapshot and throw the intended DatabaseOperationException('Async executor has been shut down') that scheduleTask already produced for the same race, instead of a raw NullPointerException. Regression test asserts the intended exception for both helpers after kill().

PR #5072, opened 2026-07-07
fix: bound the help-waiting loop on a wedged peer (follow-up to #5062)

lvca:
Follow-up to #5062, addressing the round-6 review findings that landed after the merge.

## The liveness hole (round-6 point 1)

`offerHelping` had NO backstop when the helping worker's OWN queue is empty. A worker handing off a cross-slot follow-up (the bidirectional-edge incoming-link task) to a wedged-alive peer spun forever:

- `offer` fails (peer's queue full, peer completes nothing but `isAlive()` is true);
- the deferral budget never grows (`helpDeferredTasks` stays empty - there is nothing to poll);
- `self.queue.poll()` returns `null`, `continue` - **forever**.

Both workers were silently lost during normal operation (not just shutdown), with no error surfaced - and the javadoc's liveness argument silently assumed a non-empty own queue. The producer path already failed loudly at the 12-window backstop; the cross-scheduling worker did not.

**Fix:** the helping loop now applies the same progress-gated `STALLED_NO_PROGRESS_WINDOWS` (12) backstop as the producer path, measured in wall-clock windows of `checkForStalledQueuesMaxDelay` (60s with defaults). On trip, the hand-off fails loudly: the exception unwinds through `scheduleTask` into `executeTask`'s catch - `onError` + batch rollback + `completed()` - and the worker itself survives and keeps serving its queue. Only the long, conservative bound is used (not the 3-window cross-slot one) because aborting mid-task costs the current batch; a peer parked on a legitimately slow chain is outlived as long as its individual tasks stay under the bound - the same operator cliff already documented for the producer backstop.

**Red-first evidence:** `AsyncHelpingEmptyQueueBackstopTest` - on the previous code it spun forever with no error (10s latch timeout); fixed code surfaces the stall in ~1.2s at a 50ms test window, and a follow-up task scheduled on the same worker still executes (the worker survives the aborted hand-off).

## Prompt shutdown bail (round-6 point 2)

After consuming `FORCE_EXIT` the helping loop set `forceShutdown` but kept looping to hand off the follow-up, so on a wedged peer `close()` paid the whole grace period before the #5062 interrupt escalation fired. The loop now bails out at the top of each iteration once the flag is set, dropping the follow-up loudly via the shutdown exception - workers are dying, same semantics as the target-dead branch. The interrupt escalation stays in place as the backstop for parks that never poll the marker (a budget-exhausted worker waiting in `offerWaiting`).

**Red-first evidence:** `AsyncShutdownPromptHelpingBailTest` - on the previous code `close()` paid the full 8s grace period (9.5s elapsed); fixed code returns in ~1.3s.

## waitCompletion note (round-6 point 3)

One-line comment at the enqueue-phase early return: completion markers already enqueued on earlier workers are not a leak - they execute and count down a latch nobody awaits anymore.

The `offerHelping` javadoc liveness claim is corrected to match the implementation, and the release notes are updated.

## Verification

`mvn -pl engine compile`; full `com.arcadedb.database.async` battery (22 tests, including the two new red-first regressions) plus `AsyncTest` (11) - all green, including the #5062 shutdown-escalation regression unchanged.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -266,7 +266,9 @@ that could starve the very snapshot resync meant to heal the node.
   ([#4954](https://github.com/ArcadeData/arcadedb/issues/4954)). Follow-up hardening on the helping
   loop itself: a worker handing off a cross-slot task to a wedged-alive peer while its OWN queue is
   empty (so the deferral budget cannot grow) no longer spins forever - the same 60s progress-gated
-  backstop aborts the hand-off loudly via `onError` and the worker keeps serving its queue; and a
+  backstop aborts the hand-off loudly via `onError` and the worker keeps serving its queue (failure mode
+  when this trips on a bidirectional-edge follow-up: the outgoing direction is already persisted while the
+  dropped follow-up carried the incoming link - a PARTIAL edge until the operation is repeated); and a
   worker that consumed the shutdown marker while help-waiting now abandons the hand-off immediately,
   so `close()` returns within one offer window on that path instead of waiting out the grace period.
   Low-severity cleanups from the same audit: async executor settings (`parallelLevel`, `commitEvery`,
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -261,10 +261,14 @@ that could starve the very snapshot resync meant to heal the node.
   task left in its queue instead of `queue.clear()`-ing them (threads blocked in `scanType()` /
   `waitCompletion()` hung forever), and `close()` no longer blocks unbounded on `queue.put(FORCE_EXIT)`
   under the lifecycle lock when a busy worker's queue is full - the marker is offered with a timeout,
-  the worker is interrupted on failure, and a worker still alive after the 10s grace period (e.g. one
-  that consumed the marker while help-waiting on a wedged peer's full queue) is escalated to an
-  interrupt and re-joined instead of being left running behind a returned `close()`
-  ([#4954](https://github.com/ArcadeData/arcadedb/issues/4954)).
+  the worker is interrupted on failure, and a worker still alive after the 10s grace period is
+  escalated to an interrupt and re-joined instead of being left running behind a returned `close()`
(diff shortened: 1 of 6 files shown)


lvca:
Round 1 addressed in ac615d91b, per point:

**2. Positive-case test - taken.** It covers the one branch the red-first pair cannot reach: the `else { windowsWithoutProgress = 0; ... }` reset. `AsyncHelpingSlowPeerProgressTest` pins the exact backstop scenario (empty own queue, full peer queue) but injects steady progress by ticking the peer's `completedTaskCount` (made package-private for this, same precedent as the other test hooks) - white-box because a genuinely progressing peer frees queue slots and ends the park long before enough windows elapse, while the tick reproduces the signature of a saturated pipeline refilled by faster producers. Two-phase assertion: ~40 elapsed windows of ticked progress with no stall, then progress stops and the backstop must fire, proving the counter was resetting rather than merely never advancing. Mutation-verified: with the reset branch deleted the test fails during the tick phase.

**3. Helper duplication - taken.** Shared `AsyncTestTasks` factory (`awaitTask`/`countingTask`/`noOpTask`) now backs the three tests of this PR. The tests that merged with #5062 are left untouched to keep this PR's footprint small; they can migrate opportunistically.

**4. Thread-recreation side effect - taken.** Dedicated package-private `recreateThreadsForTests()` hook (same precedent as the #4955 `getBestSlot` hook); the three tests of this PR use it instead of `setTransactionUseWAL(true)`, so they no longer depend on that setter calling `createThreads()`.

**1. Stall-window helper extraction - deferred, as the review allows.** The two sites sample differently: `offerWaiting` once per failed 5s offer with dual thresholds, `offerHelping` on wall clock across 1-100ms iterations with a single threshold. A stateful shared helper would need to parameterize both the sampling cadence and the trip policy, which obscures more than it protects at two call sites; the new positive-case test now guards the semantics both loops must share.

Verification: `mvn -pl engine compile`; `com.arcadedb.database.async` battery (23 tests, including the new mutation-verified positive case) plus `AsyncTest` (11) - all green.

Issue #5064, opened 2026-07-06
[ha-raft] Define the user-visible contract for post-quorum local commit failure (committed-but-reported-failed)


lvca:
Found while fixing #4940 (PR #5061). In `RaftReplicatedDatabase.applyPhase2` (~415-445), when `commit2ndPhase` fails locally AFTER the Raft entry is quorum-committed, `reconcileLeaderPagesAfterPhase2Failure` persists the pages from the payload WAL bytes and the exception is rethrown: the user is told the commit FAILED while the data is durably committed cluster-wide.

With #5061's fix, a pre-local-WAL failure in that flow additionally resets user-held new-record RIDs to provisional, so an application-level retry of the same objects would INSERT DUPLICATES rather than update the already-reconciled records.

Pre-existing anomaly inherent to post-quorum local failure (the #4740 design steps the leader down). Proposal: surface a distinct 'committed-but-local-apply-failed' error type instead of a generic commit failure, so applications can distinguish 'retry' from 'already durable'.

_Discovered during the 2026-07 audit follow-up (#4962)._

lvca:
Resolved in #5075 (merged, option A from the design discussion).

**The contract:** when the replication quorum durably commits a transaction but the leader's local phase-2 apply fails, the application now receives `TransactionCommittedRemotelyException` (engine exception package, catchable without an ha-raft dependency): the transaction IS committed cluster-wide, the message states whether local pages were reconciled, and it must NOT be retried.

**The duplicate-insert fix:** a `remotelyCommitted` durability regime in `TransactionContext`, set by the HA layer after quorum commit - a local failure past that point releases resources without rolling back user-held record identities (they are the identities the cluster committed) and without fencing (pre-append only; post-append still takes the intentional #5053 fence). Application retries no longer insert duplicates of already-committed records.

**The wire contract, all three layers:** raw over embedded API; HTTP maps it to a non-retryable 409 (never a retry-inviting 5xx), including when wrapped by script/auto-commit wrappers; follower-forwarded writes reconstruct the exact type via the leader-exception factory instead of collapsing it to a generic TransactionException.

Verified red-first at every layer: 3-node IT (embedded + over-the-wire 409), HA unit for the ALL-quorum recovery path, engine regime test on the reused-context sticky-flag hazard, server handler tests incl. wrapped variants. Eight review rounds, the last four endorsements.

Issue #4926, opened 2026-07-03
[storage] Torn 64KB page write + version-based replay skip loses committed deltas even with full WAL fsync


lvca:
**Severity:** HIGH (silent corruption, undetectable)

**Location:** `TransactionManager.java:417-429` (`applyChanges` skip `txPage.currentPageVersion <= page.getVersion()`), `PaginatedComponentFile.java:232-293` (full-page write), `BasePage.java:35-37` (version in bytes 0-3 of the page).

`flushPage` rewrites the whole 64KB page (128 sectors); the kernel/disk may persist them out of order. The MVCC version lives in sector 0; WAL records carry only the modified delta range.

**Crash timing:** tx commits vN+1 with the WAL fully fsynced (`txWalFlush=2`). The data-page flush is in flight at power loss after the version-header sector hit the platter but before the delta sectors did. On restart, recovery reads version=N+1 but delta bytes are still version-N content. The `<=` check says "already applied" and skips the WAL replay -> silently corrupt page (new version, old content) that all future MVCC checks accept, despite an intact WAL. No per-page checksum exists to detect it.

**Fix (cheap):** change the skip to strict `txPage.currentPageVersion < page.getVersion()`. Re-applying a delta at equal version is idempotent (delta bytes are absolute) and repairs the torn state. **Longer term:** per-page CRC in the header, verified on load, so torn writes are detectable independently of versions.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5052 (merged). Recovery's version skip is now strictly `<`: the equal-version case falls through and RE-APPLIES the WAL delta, which is idempotent when the page is intact (the delta carries absolute bytes for exactly that version) and repairs the torn state when the version header persisted ahead of the delta sectors. Skipping strictly-older entries remains, as their delta would overwrite newer content.

Red-first verified end to end with a forged torn write (disk version header spliced to v2 while the content still holds the v1 bytes): on the old `<=` code the stale value survived recovery permanently; with the fix the committed value is recovered.

Known limitation (surfaced in review): async flush can coalesce several committed versions into one physical write, and only the newest version's delta region is repaired - fully closing the multi-version case needs per-page checksums to detect which pages are torn, tracked in #5054 (the longer-term fix this issue itself anticipated).

PR #5075, opened 2026-07-07
fix(#5064): distinct contract for post-quorum local commit failure (committed-but-local-apply-failed)

lvca:
Implements option A from the #5064 design discussion.

## The problem
When the Raft quorum durably commits a transaction but the leader's **local** phase-2 apply fails, the application received a generic commit failure - told the commit failed while the data was committed cluster-wide. Worse: an application-level retry of the same records **inserted duplicates**, because the #4940 rollback reset their identities to provisional and the retry re-inserted records the cluster already held.

## The fix
- **`TransactionCommittedRemotelyException`** (engine exception package, catchable without an ha-raft dependency): states the transaction IS committed cluster-wide, whether local pages were reconciled, and that it must NOT be retried. `reconcileLeaderPagesAfterPhase2Failure` now reports its outcome instead of swallowing it.
- **A `remotelyCommitted` durability regime in `TransactionContext`**, set by the HA layer after quorum commit: a local failure past that point releases resources **without** rolling back user-held record identities (they are the identities the cluster committed) and **without** fencing (no orphaned local WAL record exists; the Raft layer reconciles pages from the replicated payload). Slots between the `walAppended` and fence-refused/rollback regimes - #4940/#5053 semantics unchanged for non-replicated databases. The ALL-quorum recovery path gets the same boundary shift.

## Verification
Red-first: `Issue5064CommittedRemotelyContractIT` (3-node cluster, single-shot post-quorum fault) fails on pre-fix behavior (raw exception escapes, identity reset) and passes with the fix - distinct type, actionable message, identity preserved, all nodes converge on the committed data after step-down. HA phase-2 battery green (`Issue4740Phase2ReconcileIT`, `Issue5018Phase2ConflictMessageTest`, `DatabaseReconcilerTest`); engine commit-path battery green (WalCommitOrdering, Issue4940, Issue4959, ExplicitLocking, IsolationContract).

## Conflict risks
`TransactionContext` (one new field + one new regime branch in the finally) - the follow-up fan-out workers do not touch it.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -390,7 +390,9 @@ that could starve the very snapshot resync meant to heal the node.
   whether the local pages were reconciled, and that it must NOT be retried - instead of a generic commit
   failure that invited retries. The identities of records created in such a transaction are no longer reset
   to provisional (they are the identities the cluster committed), so an application-level retry no longer
-  inserts duplicates of already-committed records
+  inserts duplicates of already-committed records. The contract also survives the wire: the HTTP command
+  endpoint maps it to 409 Conflict (not a retry-worthy 5xx), and a follower forwarding a write to the
+  leader reconstructs the same exception type instead of a generic commit failure
   ([#5064](https://github.com/ArcadeData/arcadedb/issues/5064)).
 
 ### Improvements
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -384,6 +384,15 @@ that could starve the very snapshot resync meant to heal the node.
   levels, unbounded per-transaction page cache under `REPEATABLE_READ`) is now documented on
   `Database.TRANSACTION_ISOLATION_LEVEL` and pinned by tests.
 
+- **HA: a distinct error for "committed cluster-wide, local apply failed".** When the replication quorum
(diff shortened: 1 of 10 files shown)


lvca:
Round 8 dispositions (no push - all four observations are non-blocking refinements and the branch is converged):

1. **Two-level nesting gap**: acknowledged as-is, per the review's own framing - the one-level unwrap is deliberately consistent with the adjacent DuplicatedKeyException handling. A shared unwrap-to-root-cause helper for retryable/non-retryable types is a reasonable future refactor but belongs to its own change touching both types, not this PR.
2. **setRemotelyCommitted is public engine API**: accepted trade-off, documented at the field and setter. The engine has no visibility mechanism narrower than public that reaches ha-raft (different module); the sharp javadoc plus the regression tests are the guard.
3. **Poisoned-page eviction ordering**: confirmed - the ids are captured before the throwing call, as the review verified.
4. **saveConfiguration sub-case in the IT**: the sub-case is documented at the production catch site (the #4926-idempotency note), which is where a maintainer touching that code looks first; the IT's fault-injection point is pre-commit2ndPhase by design.

Eight rounds, the last four all endorsements with polish. Ready to merge.

PR #5077, opened 2026-07-07
fix(#5065): bound the parallel select producer offer, stop the 100% CPU spin

### LINKED ISSUE #5065: [engine] SelectParallelIterator callback busy-spins at 100% CPU forever if the consumer stops draining
`SelectParallelIterator.java` (~line 85): the `DatabaseAsyncBrowseIterator` callback hands records to the consumer with `while (true) { if (queue.offer(record)) break; }` - an unbounded busy-spin with no backoff and no shutdown check. If the consuming thread stops draining (exception, early close), the async worker spins at 100% CPU forever.

Fix shape: bounded offer with timeout + a cancellation/shutdown check in the loop.

_Discovered during the 2026-07 audit follow-up (#4962), outside PR #5062's footprint._

lvca:
Fixes #5065 (2026-07 audit follow-up, #4962; outside PR #5062's footprint).

## Verdicts

| Issue | Verdict | Detail |
|---|---|---|
| #5065 | fixed | Reproduced on current main, red-first, fixed with a bounded cooperative offer. Worse than filed: besides pinning one async worker per bucket at 100% CPU, the spin also hung the next `Database.close()`/`drop()` forever behind the executor's unbounded `waitCompletion()`. |

## Red evidence (unmodified main)

Running the new `SelectParallelIteratorTest` against main wedged the JVM permanently. `jstack` showed 4 `AsyncExecutor` workers RUNNABLE at `SelectParallelIterator.java:86` (the `while (true) { offer }` loop) with ~706,000 ms CPU each on a 712 s old JVM, and the JUnit main thread parked forever in `LocalDatabase.closeInternal -> DatabaseAsyncExecutorImpl.waitCompletion` during teardown. The run had to be killed.

## Fix

A producer facing a full hand-off queue now parks 0.1 ms between offer attempts (no CPU burn even while legitimately waiting) and gives up when:

- the iterator was closed (`SelectIterator` is now `AutoCloseable`; the parallel `close()` signals the producers, the serial one is a no-op);
- the worker was interrupted (executor shutdown escalation, #4954);
- enough records were produced to satisfy the limit; or
- the consumer made no progress for the whole stall bound: the select timeout when set, otherwise 60 s (matching the #5062 zero-progress backstop).

Async executor contracts from #5062 are preserved: a producer that gives up returns `false` from the browse callback, so the task ends through the regular completion path (`completed()`, batch commit, semaphore countdown); nothing is thrown through the worker and no in-flight commit batch is rolled back. A consumer still alive after a producer stall gets a `TimeoutException` instead of a silently truncated result set (unless it opted into truncation with a non-throwing timeout).

Three latent defects in the same code path fixed by the rework (details in the commit message):

1. `limit` was accounted producer-side with racy increments of `returned`, so the consumer could receive fewer than `limit` records (down to zero);
2. `skip > 0` on the parallel path crashed with an NPE (base constructor consumed skip via the override before the queue existed);
3. a record published between the consumer's last poll and the final producer's `countDown()` could be lost at end of stream (non-atomic queue-empty + latch check).

## Tests

- New `SelectParallelIteratorTest` (10,000 records / 8 buckets, dataset deliberately larger than the 4,096-slot queue): full-scan sanity, limit exact + workers released, early close releases workers, abandoned consumer releases workers within the stall bound, skip works. All red on main; all green in 1.5 s with the fix.
- Full `com.arcadedb.query.select.*` + `com.arcadedb.database.async.*` + `AsyncTest` suites: 97 tests, 0 failures (includes the pre-existing 1M-record `okParallel` and every #5062 executor-contract regression test).

## Conflict risk

- `docs/release-26.7.2.md`: appended one bullet at the end of Fixes; expected to conflict trivially with sibling audit PRs.
- Engine changes are confined to `query/select` (`SelectParallelIterator` rewrite, 12-line `SelectIterator` addition); no touch of the async executor, TransactionContext or PageManager.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -392,8 +392,6 @@ that could starve the very snapshot resync meant to heal the node.
   conflict-heavy file) and the descending duplicate-run cursor rescan (perf-only); the `createNewPage`
   rollback counter drift was subsequently fixed in
   [#5067](https://github.com/ArcadeData/arcadedb/issues/5067).
-  conflict-heavy file), the `createNewPage` rollback counter drift and the descending duplicate-run
-  cursor rescan (perf-only).
 - **Storage: flush suspension is refcounted - overlapping backup/verify/snapshot each own their window
   (2026-07 audit follow-up).** `suspendFlushAndExecute` ownership was first-caller-wins: with two
   concurrent suspenders on the same database (SQL `BACKUP DATABASE`, the HA verify endpoint, HA snapshot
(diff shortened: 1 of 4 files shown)


lvca:
This review round appears to have been generated against the pre-push snapshot - all three points are already addressed in `9d6bc0899` (pushed shortly before it landed):

1. **Timeout regression**: fixed - checkForTimeout() now runs at the top of every fetch loop iteration (measured: one field comparison when no timeout is set), covering the streaming path. Red-first: `timeoutIsEnforcedWhileStreaming` and the non-throwing truncation variant both failed on the prior branch code and pass now.
2. **Consumer-side spin**: fixed - the empty-queue branch spins 64 polls then parks per-poll with the same backoff as the producer side.
3. **skip+limit**: tested and fixed, and it was worse than predicted - `skip(100).limit(50)` returned ZERO records (the constructor's skip loop itself tripped the returned>=limit cap). One shared line fixes both the parallel and serial paths (`returned >= (long) limit + skip`, matching the ORDER BY path's semantics), with regression tests for both.

Both nits also taken (boolean waiting flag; limit check before the filterOutRecords insert). SelectParallelIteratorTest 9/9, select sweep 312/312, async suites green.

Issue #4509, opened 2026-06-08
WAL rotation drops the previous WAL once `pendingPagesToFlush==0` — but `notifyPageFlushed` fires on `write()` not `fsync()`


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.engine.TransactionManager`

### Summary

`checkWALFiles()` rotates the active WAL once it exceeds `MAX_LOG_FILE_SIZE`.
`cleanWALFiles()` drops the old file once `pendingPagesToFlush == 0`. But
`notifyPageFlushed` is decremented inside `flushPage()` after a (non-fsynced)
`file.write()`. The OS still holds the dirty bytes. Power loss between WAL
drop and page write-back → both lost.

### Code

`engine/com/arcadedb/engine/TransactionManager.java:586–612, 188–273`

### Impact

Compounds with #11 and #12. Power loss right after rotation can lose
transactions and silently corrupt subsequent transactions on the same pages.

### Suggested fix

Before `file.drop()` on an inactive WAL, fsync every data file the WAL
touched. `notifyPageFlushed` should fire only after fsync, not after `write`.

lvca:
Reproduced.

lvca:
Fixed

lvca:
That was also hard to debug, thanks!

PR #5078, opened 2026-07-07
feat(#5054): per-page CRC32C checksums close the multi-version torn-write recovery gap

### LINKED ISSUE #5054: [storage] Torn-write repair covers only the newest coalesced version; multi-version needs per-page checksums
Follow-up to #4926 (fixed in #5052).

The equal-version WAL re-apply added in #5052 repairs a torn 64KB page write, but only for the NEWEST version's delta region. Async flush coalesces several committed versions into a single physical page write, so a page can go v1(on-disk) -> v2 -> v3 -> v4 with only the v4 flush reaching the platter. If that flush tears, recovery re-applies v4's delta (version == disk) but skips v2 and v3 as strictly-older, so regions changed by v2/v3 can remain torn -> still silently corrupt.

The version header alone cannot distinguish a torn page from an intact one, so detecting which pages need the older deltas re-applied requires a **per-page checksum** (the longer-term fix already noted in #4926). An unconditional in-order replay of every WAL entry whose version <= disk version would also repair it, but changes the recovery semantics that also govern the version-gap / forceApply (replication) paths, so it needs its own careful design + tests.

Scope for this issue: add a per-page checksum written on flush and verified on recovery; on mismatch, replay the full WAL delta chain for that page from the last known-good version. Surfaced by the code review on #5052.

lvca:
Follow-up to #4926 / #5052 from the 2026-07 audit backlog. Closes #5054.

## What does this PR do?

**#5054 (implemented, red-first).** Async flush coalesces several committed versions into a single physical page write, so a page can go v1(on-disk) -> v2 -> v3 -> v4 with only the v4 flush reaching the platter. If that flush tears, the #5052 equal-version re-apply repairs only v4's delta region: v2/v3 entries are skipped as strictly-older and their byte regions stay silently torn. The version header alone cannot distinguish this page from an intact one.

Every page flush now also maintains an 8-byte slot (page version + CRC32C of the raw page bytes) in a `.pcrc` sidecar file next to each data file, written BEFORE the page bytes and fsynced under the same `force()` barrier. During crash recovery a pre-scan collects the per-page maximum retained WAL version, and each WAL-referenced page is verified once: a page whose checksum reveals a torn write gets its FULL retained WAL delta chain replayed in transaction order (the strictly-older skip and the version-gap abort are bypassed for that page only). Correctness rests on the fsync-before-WAL-drop barrier (#4509/#4934): a dropped entry is always durably contained in the on-disk page, so in-order application of every retained entry reconstructs exactly the newest committed state. A guard refuses the repair when the WAL retains no entry at or above the disk version, so a stale checksum (e.g. a sidecar restored from a backup) can never downgrade an intact page.

Red-first evidence (`PageChecksumRecoveryTest`): three documents in one page updated by three separate transactions (v2/v3/v4, disjoint byte regions) with flushing suspended, kill, version header forged to v4 over the v1 content. Before the fix recovery produced `{doc1=A1, doc2=A2, doc3=B3}` - the exact #5054 silent corruption; after, the full chain replay restores `{B1, B2, B3}`. Also covered: the tear with a persisted equal-version slot (CRC mismatch as direct positive evidence), the missing-sidecar migration path, the slot-matches-disk flush contract, and a corrupted sidecar on a cleanly-closed database (verification is recovery-only).

## Design decisions (design-first item)

- **Where the checksum lives.** A `.pcrc` sidecar per data file (8 bytes per page at `pageNumber * 8`), NOT a page-header extension. The header extension was analyzed and rejected: component file versions carry per-component semantics (`LocalBucket.purposeForVersion` treats v>=1 as EXTERNAL_PROPERTY and `maxRecordsInPageForVersion` keys the slot table on the same integer), so a cross-cutting "has checksum" gate cannot ride that version without redesigning component versioning across every component type; `PAGE_HEADER_SIZE` is additionally a static constant compiled into 17+ classes including the just-stabilized commit path, and the WAL/replication deltas carry absolute in-page offsets. The sidecar needs no version gating at all.
- **Algorithm.** `java.util.zip.CRC32C` (hardware-accelerated) over the full raw page bytes exactly as written, version stored alongside so a slot can be matched to the page state it describes.
- **When written.** Inside `PaginatedComponentFile.write()` - every write site (flush thread, sync flush, recovery/replicated `writePageWithLock`) is covered automatically - BEFORE the page bytes, so a persisted slot always describes the newest attempted write. Cost is on the flush thread, not commit latency. Gated by `arcadedb.pageChecksum` (default true). A sidecar write/fsync failure is contained as a WARNING: checksums are advisory and must never veto the data file's fsync verdict.
- **When verified.** Recovery only (`TransactionManager.checkIntegrity` passes a `RecoveryPageVerifier` into `applyChanges`; the HA/Raft replay callers pass null and are untouched). Read-path verification behind a config flag was deliberately left out to keep the hot read path unchanged; it can be added later (noted below).
- **Mismatch handling.** Torn page -> replay the full retained WAL chain in tx order for that page, guarded by the pre-scanned per-page max WAL version (a replay can never end below the version the disk already claims).
- **Migration.** None needed, in both directions. Old databases have no sidecar -> verification is skipped and slots fill in as pages flush. `FileManager` scans the directory by extension whitelist, so old and NEW versions alike ignore `.pcrc` (downgrade-safe). `drop()` deletes the sidecar; `rename()` (LSM compaction temp-suffix removal) carries it along; HA file transfers that omit it only lose verification, never data.

## Per-issue verdict

- **#5054: implemented.** Red-first TDD; the forged multi-version torn write reproduces the exact silent corruption on main and is repaired with the fix.

## Test evidence

- `PageChecksumRecoveryTest` (new, 5 tests): red-first on main for the two torn-write cases (`{A1, A2, B3}` observed pre-fix), green with the fix.
- Full engine+database battery (`com.arcadedb.engine.**` + `com.arcadedb.database.**`): **624 tests, 0 failures, 0 errors** (1 pre-existing skip).
- LSM compaction/drop suites (`LSMTreeIndexCompactionTest`, `LSMTreeCompactionCorrectnessTest`, `DropIndexTest`, `Issue4960LSMReloadCompactionCounterTest`): green - exercises the sidecar rename/drop lifecycle.
- Recovery suites (`WalRecoveryCorrectnessTest`, `WALVersionGapRecoveryTest`, `Issue4508TornWALRecoveryTest`, `Issue4510ForceApplyPartialDeltaTest`, `ACIDTransactionTest`): green.
- Earlier battery rounds surfaced two real issues, both resolved in this PR: the sidecar fsync initially threw `ClosedByInterruptException` when the main channel's interrupt retry restored the flag first (now interrupt-shielded like the main channel), and the sidecar-creation latency inside the per-page interlock made a latent race in `Issue4712ReplicatedWriteLockTest` / `Issue4510ForceApplyPartialDeltaTest` frequent - their fake-interlock/reload setup raced the async flush of their own setup transaction (`LocalDatabase.equals` is path-based, so the race spans helper database instances). Fixed by draining the flush pipeline after setup (second commit); verified 8/8 green isolated runs.

## Conflict-risk notes

- `docs/release-26.7.2.md`: expected to conflict with sibling audit PRs (append-style entries); trivial to resolve.
- `TransactionManager.applyChanges`: touched by #5052 recently; this PR builds directly on that code (comment block updated in place). Sibling PRs touching recovery would conflict here.
- `PaginatedComponentFile`: new sidecar members and `write()`/`force()`/`close()`/`rename()`/`drop()` additions; low conflict risk.

## Follow-up candidates (not in scope)

- Optional checksum verification on normal page reads behind a config flag (default off), and a CHECK DATABASE integration that validates every slot offline.
- Sparse/vector bulk writers that bypass `PaginatedComponentFile.write()` leave no slots for their pages (verification simply skips them); wiring them in would extend coverage.

## Checklist

- [x] I have run the build using `mvn clean package` command (engine module + full engine/database battery)

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -216,12 +216,26 @@ that could starve the very snapshot resync meant to heal the node.
   skip declared such a page "already applied" and left it silently corrupt despite an intact WAL,
   [#4926](https://github.com/ArcadeData/arcadedb/issues/4926); known limitation: async flush can coalesce
   several committed versions into one physical write, and only the newest version's region is repaired -
-  fully closing the multi-version case needs per-page checksums to detect which pages are torn, tracked in
+  fully closing the multi-version case needs per-page checksums to detect which pages are torn - now done,
+  see the per-page checksums entry below,
   [#5054](https://github.com/ArcadeData/arcadedb/issues/5054)). A clean close whose data fsync failed now
   preserves the WAL and the lock file for recovery instead of deleting them (after a failed fsync the OS
   may have dropped the dirty pages, so the WAL held the only durable copy;
   [#4934](https://github.com/ArcadeData/arcadedb/issues/4934) - the runtime WAL rotation also skips its
   drop pass when the pre-drop fsync fails).
+- **Per-page checksums close the multi-version torn-write gap (2026-07 audit follow-up).** Every page flush
+  now also maintains an 8-byte slot (page version + CRC32C of the raw page bytes, hardware-accelerated) in a
+  `.pcrc` sidecar file next to each data file, written before the page itself and covered by the same fsync
+  barrier. Crash recovery uses it to detect what the page version header alone cannot reveal: async flush
+  coalesces several committed versions into one physical write, so a torn write can persist the newest
+  version header while regions changed by the coalesced OLDER versions still hold stale bytes - the #4926
+  equal-version re-apply repaired only the newest version's delta region. A page that fails verification now
+  gets its FULL retained WAL delta chain replayed in transaction order (safe by the fsync-before-WAL-drop
+  barrier: dropped entries are always durably contained in the on-disk page), guarded so a stale checksum
+  can never trigger a replay ending below the version already on disk. No data-file format change: the
+  sidecar is ignored by directory scans (old and new versions alike), databases without one recover exactly
+  as before, and the slots fill in as pages flush. Disable with `arcadedb.pageChecksum=false`
+  ([#5054](https://github.com/ArcadeData/arcadedb/issues/5054)).
 
 - **LSM index hardening (2026-07 audit).** A failed compaction is now atomic: the in-RAM page count is
   rolled back (after draining in-flight flushes) so the orphaned leaf pages a mid-merge failure leaves on
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -364,6 +364,16 @@ public enum GlobalConfiguration {
 
   PAGE_FLUSH_QUEUE("arcadedb.pageFlushQueue", SCOPE.DATABASE, "Size of the asynchronous page flush queue", Integer.class, 512),
 
+  PAGE_CHECKSUM("arcadedb.pageChecksum", SCOPE.JVM,
+      """
+      Maintains a per-page CRC32C checksum in a '.pcrc' sidecar file next to each data file, updated on every \
+      page flush. Crash recovery uses it to detect torn page writes that the page version header alone cannot \
+      reveal (issue #5054: async flush coalesces several committed versions into one physical write, and a torn \
+      write can persist the newest version header over stale data sectors) and repairs the damaged page by \
+      replaying its full retained WAL delta chain. Databases without sidecar files (created by older versions, \
+      or with the flag disabled) recover exactly as before: verification is skipped.""",
+      Boolean.class, true),
+
   FLUSH_SUSPEND_MAX_DEFERRED_RAM("arcadedb.flushSuspendMaxDeferredRAM", SCOPE.DATABASE,
       """
       Maximum amount of RAM (in MB) of dirty pages the page-flush thread may defer in memory while flushing \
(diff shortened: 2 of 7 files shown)


lvca:
**On standby by design decision (Luca):** the sidecar-per-data-file approach doubles the open-file-descriptor count, which matters on databases with many buckets/files. Converting to draft while the design is reconsidered.

Alternatives to evaluate when this resumes:
- a **single consolidated checksum file per database** (one extra fd total, slots addressed by fileId + pageNumber);
- **interleaved checksum pages inside the data file itself** (every Nth page holds the checksums for the following run - zero extra files, no page-header change, at the cost of remapping page addressing);
- revisiting the **page-header route** with a global format-version gate rather than the per-component version (the blocker analysis in this PR's design notes still applies, but a dedicated engine-wide format epoch could clear it).

The recovery/replay machinery, the red-first tests and the correctness argument in this PR are design-independent and reusable whichever storage layout wins. Review-round responses are paused; the round-1 findings (flush-path allocations, flag-gating of verification, throughput number) fold into the redesign.

lvca:
**Deferred to a dedicated design-first campaign** (not part of the 2026-07 audit follow-up, which is now complete). This stays a DRAFT. The blocking decision is the storage layout - the current .pcrc sidecar-per-data-file doubles the open-fd count (Luca's concern); the alternatives to weigh at kickoff are: a single consolidated per-database checksum file, interleaved checksum pages inside the data file (zero extra files), or the page-header route behind an engine-wide format epoch. The recovery/replay machinery, the red-first multi-version torn-write tests, and the correctness argument here are layout-independent and carry over to whichever wins.

lvca:
This will be postponed

PR #5081, opened 2026-07-07
fix(#5066): map the 'fast' async queue to the MPMC DisruptorBlockingQueue

### LINKED ISSUE #5066: [engine] Async post-shutdown task undo is best-effort with asyncOperationsQueueImpl=fast (Conversant queue lacks remove)
Residual from #4954 (PR #5062): `PushPullBlockingQueue` (Conversant, `arcadedb.asyncOperationsQueueImpl=fast`) throws `UnsupportedOperationException` from `remove(Object)`, so `scheduleTask`'s post-shutdown undo (remove the task offered after the dead worker's final drain, then throw) degrades to best-effort: a task offered in that tiny window stays enqueued with `completed()` never called. The default 'standard' ArrayBlockingQueue is fully covered.

Options: a second drain pass on shutdown, or documenting the fast-queue limitation.

_Discovered during the 2026-07 audit follow-up (#4962)._

lvca:
Fixes #5066.

## Verdict per option in the issue

The issue offered two shapes: a second drain pass on shutdown, or documenting the limitation. Investigation found the real defect one level deeper, which makes both options the wrong fix:

- **The `fast` impl violated its own concurrency contract in normal operation.** `PushPullBlockingQueue` is an explicit single-producer/single-consumer design ("Transfers from a single thread writer to a single thread reader", no CAS on the tail sequence). Every async worker queue here has MANY producers: any application thread calling `scheduleTask`, cross-scheduling workers, `waitCompletion` markers, the closing thread. This is not theoretical - the new multi-producer regression test **demonstrated lost tasks empirically** on the old impl (8 producers x 250 no-op tasks, completion latch never reached zero). Note the `high-performance` profile selects `fast`, so this was not an obscure opt-in corner.
- **A second drain pass cannot be made total.** A producer that snapshotted `executorThreads` before shutdown can offer arbitrarily late, so no finite number of drain passes closes the window; the producer-side remove-undo is the only shape that works regardless of timing - and it needs `remove(Object)`.

## The fix

`arcadedb.asyncOperationsQueueImpl=fast` now maps to **`DisruptorBlockingQueue`**, the flagship MPMC variant from the same Conversant jar (no new dependency, same Apache 2.0 artifact):

- lock-free CAS-claimed sequences honor the multi-producer contract;
- `remove(Object)` is implemented (cursor-blocked scan-and-shift), so `scheduleTask`'s post-shutdown undo now works identically on both queue impls - the #5062-documented residual gap is gone;
- capacity 1 is explicitly supported (`Capacity.getCapacity`), matching small test configs; capacities round up to the next power of two, as the old ring already did.

`removeQuietly`'s `UnsupportedOperationException` catch is kept as defensive-only for hypothetical future impls, with its comment and operator WARNING updated.

## Red-first evidence

`AsyncFastQueueShutdownUndoTest` (both red on the old impl, green after the swap):
- `fastQueueMustSupportRemoveForThePostShutdownUndo`: the exact primitive the undo depends on - threw `UnsupportedOperationException` before.
- `fastQueueMustNotLoseTasksUnderConcurrentProducers`: timed out with lost tasks before; deterministically green on a compliant MPMC impl.

## Verification

`mvn -pl engine compile`; the full `com.arcadedb.database.async` battery plus `AsyncTest` run **twice**: default config (36 tests) and the fast-queue variant via `-Darcadedb.asyncOperationsQueueImpl=fast` (36 tests) - all green, so every existing shutdown/drain/stall/helping regression also passes on the Disruptor queue. Release notes updated to record that the formerly documented fast-queue gap is fixed.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -310,7 +310,9 @@ that could starve the very snapshot resync meant to heal the node.
   producers: a reproduction test showed concurrent offers silently LOSING tasks in normal
   operation, and its missing `remove(Object)` degraded the post-shutdown scheduling undo to a
   logged best-effort. The multi-producer replacement comes from the same library and jar, keeps the
-  lock-free fast path (CAS-claimed sequences), supports `remove(Object)` (the shutdown undo now
+  lock-free fast path (CAS-claimed sequences; note `fast` trades CPU for latency - the Disruptor
+  queue busy-spins before parking, so idle workers cost more CPU than `standard`, same as the
+  previous `fast` implementation), supports `remove(Object)` (the shutdown undo now
   works identically on both impls) and rounds capacities up to the next power of two. Shutdown no
   longer strands completion waiters: an interrupted or exiting worker now notifies `completed()` on every
   task left in its queue instead of `queue.clear()`-ing them (threads blocked in `scanType()` /
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -313,7 +313,10 @@ that could starve the very snapshot resync meant to heal the node.
   lock-free fast path (CAS-claimed sequences; note `fast` trades CPU for latency - the Disruptor
   queue busy-spins before parking, so idle workers cost more CPU than `standard`, same as the
   previous `fast` implementation), supports `remove(Object)` (the shutdown undo now
-  works identically on both impls) and rounds capacities up to the next power of two. Shutdown no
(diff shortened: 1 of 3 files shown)


lvca:
@claude review the latest commit (6580d19a3) - the round-1 response adding the capacity-1 guarantee test, the removeQuietly cold-path note, and the fast-profile CPU tradeoff in the release note.

lvca:
@claude can you review the latest commits?

PR #5105, opened 2026-07-07
fix(#5080): bound the async drain on close()/drop() so a wedged worker cannot hang shutdown

### LINKED ISSUE #5080: [engine] Database.close()/drop() waits unbounded on async task completion (wedged worker = unclosable database)
Discovered by the #5065 worker, with a live repro: `LocalDatabase.closeInternal` (~line 2045) calls `DatabaseAsyncExecutorImpl.waitCompletion()` with no timeout (0 -> Long.MAX_VALUE). With any async worker wedged (before #5077's fix, the SelectParallelIterator spin was one such wedge - jstack showed close() parked forever behind 4 spinning workers), the database can never be closed and the JVM never exits.

Suggested shape: a bounded wait (config-driven, in the spirit of arcadedb.flushAllPagesTimeout from #4928) with a loud escalation - interrupt the workers (the #5062 escalation machinery exists) and proceed with the crash-equivalent close semantics rather than hanging forever.

_Discovered during the 2026-07 audit follow-up campaign; related residuals (MultiIterator timeout-flag inconsistency, pendingFlushPages keyed remove, bulk-writer checksum coverage) are noted in PRs #5077/#5078._

lvca:
## The problem
`LocalDatabase.closeInternal` drained the async executor via an **unbounded** `async.waitCompletion()`, so a worker wedged inside a user task or callback (infinite loop, stuck I/O) made `close()`/`drop()` hang forever. The specific trigger that first surfaced this - the #5065 `SelectParallelIterator` busy-spin - is already fixed in #5077, but the unbounded wait remained a latent hazard for any wedged worker.

## The fix
Mirrors the existing `FLUSH_ALL_PAGES_TIMEOUT` precedent for "give up instead of hanging":
- New `arcadedb.asyncCloseTimeout` (default **60000 ms**, `0` = wait forever = pre-26.7.2 behavior) bounds the graceful drain.
- On expiry, `closeInternal` logs a WARNING and falls through to `async.close()`, which is **already bounded** (interrupt + 1s join per worker, WARNING on a holdout) and notifies completion of leftover tasks - so shutdown always makes progress.

Single behavioral line changed; `async.close()` itself was already safe.

## Verification
Red-first: `AsyncShutdownDrainTest#closeDoesNotHangOnAWedgedAsyncWorker` wedges a worker for 60s and times `database.close()` under `@Timeout(30)`. On the pre-fix unbounded wait `close()` blocks ~60s and trips the timeout (**verified red**); with the bound it gives up at 1s, force-interrupts the worker, and returns in well under 15s. Async battery green (`AsyncShutdownDrainTest`, `AsyncFastQueueShutdownUndoTest`, `AsyncWaitCompletionTimeoutTest`, `DatabaseAsyncExecutorLifecycleRaceTest`).

Closes #5080.

### DIFF
--- a/docs/release-26.7.2.md
+++ b/docs/release-26.7.2.md
@@ -371,6 +371,12 @@ that could starve the very snapshot resync meant to heal the node.
   its waiter blocked forever, and `QueryEngineManager.register()` publishes a copy-on-write map so a
   post-construction registration cannot corrupt concurrent readers
   ([#4961](https://github.com/ArcadeData/arcadedb/issues/4961)).
+- **`close()`/`drop()` no longer hang on a wedged async worker.** The graceful drain of in-flight
+  asynchronous tasks on database close/drop used an UNBOUNDED wait, so a worker stuck inside a user task
+  or callback blocked shutdown forever. It now waits at most `arcadedb.asyncCloseTimeout` ms (default
+  60000, 0 = wait forever) before logging a WARNING and forcing the async workers down; the forced
+  shutdown is itself bounded (interrupt + 1s join per worker) and notifies completion of the leftover
+  tasks ([#5080](https://github.com/ArcadeData/arcadedb/issues/5080)).
 - **PageManager lifecycle is refcounted.** The JVM-wide page manager was started and stopped on a racy
   "is the active-database map empty" check-then-act spanning factory instances: closing the last instance
   of one database could null the shared flush thread under a database whose open was still in flight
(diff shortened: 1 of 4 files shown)


lvca:
Round 7 dispositions (no push - LGTM, and further commits would only spawn more review rounds on a merge-ready PR):

1. **Negative asyncCloseTimeout = wait forever, only 0 documented** - fair; the config doc mentions only `0`. This is a one-line accuracy tweak ("0 or negative waits forever"), best folded in at merge or as a trivial follow-up rather than triggering another full CI round here.
2. **WARNING could mislabel a genuine timeout as 'Interrupted'** - the reviewer's own conclusion: purely cosmetic, log-line only, safe to leave. No change.
3. **Test paths / no @Tag(slow)** - consistent with the sibling tests in the file and sub-5s each; the reviewer marks it fine. No change.

The substantive arc is fully resolved across the six prior rounds (interrupt-vs-timeout message -> caller flag neutering the force-shutdown -> durable-flush truncation -> the closeDurableParts extraction that fixed it cleanly -> unconditional clear). This PR is ready to merge.

Issue #4800, opened 2026-06-29
`BootstrapElection` treats an unknown commit index (`-1`) as "first formation" — transient `IOException` can re-trigger bootstrap on a live cluster


ruispereira:
**Affected version:** 26.6.1
**Component:** `BootstrapElection.runIfEligible` / `RaftHAServer.getCommitIndex`
**Severity:** MEDIUM-HIGH

### Summary

The "first formation" gate is `if (commitIndex > 0) return SKIPPED_NOT_FIRST_FORMATION;`.
But `getCommitIndex()` returns **-1** when `getDivision(...)` throws `IOException`
(group not ready / transient). `-1 > 0` is false, so the gate is bypassed and
bootstrap proceeds on an already-running cluster.

### Code

`BootstrapElection.java:135-137` (gate), `:115-117` (`attemptedThisTerm` reset every
leader change); `RaftHAServer.java:1052-1060` (`getCommitIndex` returns -1 on IOException).

### Impact

A transient division-read failure during a leader-change callback bypasses the
gate; `decideAndAct` may replicate a bootstrap fingerprint or transfer leadership
on a live cluster. Leader-churn storms can re-trigger repeatedly.

### Suggested fix

Treat `commitIndex < 0` as "unknown — skip"; only proceed on a confirmed
`commitIndex == 0`.

lvca:
Fixed

PR #5113, opened 2026-07-08
fix(#5099): HA bootstrap first-formation survives leadership transfer

### LINKED ISSUE #5099: HA offline bootstrap: leadership-transfer path never commits baseline (post-transfer leader sees commitIndex != 0)
## Summary

`RaftBootstrapLeadershipTransferIT.leadershipTransfersToTheFresherPeerBeforeBootstrapCommit` fails: after the offline-bootstrap protocol transfers Raft leadership to the elected source peer, that new leader never commits the `BOOTSTRAP_FINGERPRINT_ENTRY`, so `getBootstrapBaseline(db)` stays `null` on every peer and the test times out.

Confirmed failing on pristine `main` (not a regression). Distinct from the transient-`-1` first-formation bug fixed in #5098 - that fix is a prerequisite for even reaching this path.

## Root cause

`BootstrapElection`'s class Javadoc assumes:

> the protocol never commits a baseline before transferring leadership, so the single transfer-then-commit-all sequence keeps the `commitIndex == 0` first-formation gate valid

This assumption is **empirically false**. Instrumented run:

```
Leader elected: ArcadeDB_0 (term=1)
Bootstrap: transferring leadership to elected source localhost_2436 (freshest copy)
Leader changed: ArcadeDB_0 -> ArcadeDB_2 (term=2)   ← new leader
DIAG bootstrap outcome=SKIPPED_NOT_FIRST_FORMATION commitIndex=2 isLeader=true
```

The leadership transfer and the term-2 leader election cause Ratis to append and commit internal no-op / configuration entries, so the **new** leader (the elected source) reads `commitIndex = 2`, not `0`. The `isConfirmedFirstFormation(commitIndex)` gate (`commitIndex == 0`, tightened in #4800) therefore rejects the post-transfer leader, and the baseline is never committed anywhere.

## Why it is hard

The gate can't simply be loosened to `<= 2` or similar: #4800 deliberately keys first-formation on an exact `0` to prevent re-triggering bootstrap on an already-running cluster (which would risk overwriting live data). A correct fix needs a first-formation signal that survives an internal term bump but still cannot fire on a cluster that has committed application data - e.g. "no `BOOTSTRAP_FINGERPRINT_ENTRY` has been committed for any database yet" tracked independently of the raw Ratis commit index.

## Reproduce

```
mvn test -pl ha-raft -Dtest=RaftBootstrapLeadershipTransferIT -DskipITs=false -DfailIfNoTests=false
```
(requires a fresh `arcadedb-engine` in `~/.m2`, e.g. `-am`)

## Related

- #4147 (offline bootstrap protocol), #4800 (first-formation `-1`/`0` gate), #4807 (single deterministic source), #5098 (transient-`-1` fix, prerequisite)

robfrank:
Closes #5099

## Summary
The offline-bootstrap first-formation gate keyed on `commitIndex == 0`. When the protocol transfers Raft leadership to the elected source peer, the transfer plus the term-2 leader election make Ratis append and commit internal no-op / configuration entries, so the new leader reads `commitIndex = 2`, not `0`. The exact-`0` gate (tightened in #4800 to reject a running cluster) therefore rejected the post-transfer leader and the `BOOTSTRAP_FINGERPRINT_ENTRY` was never committed - `getBootstrapBaseline(db)` stayed `null` on every peer.

The gate cannot simply be loosened to `<= 2`: #4800 keys first-formation on an exact `0` to stop bootstrap re-triggering on an already-running cluster (which would risk overwriting live data).

The fix adds a durable first-formation signal, `ArcadeStateMachine.hasNeverAppliedApplicationEntry()`. Ratis internal no-op/config entries never flow through `applyTransaction`, so they advance neither the in-memory `lastAppliedIndex` nor the persisted applied index - while every real mutation (TX/SCHEMA/INSTALL/DROP/SECURITY/BOOTSTRAP entry) advances both. The signal consults both indices so a restarted, already-bootstrapped cluster (bootstrap entry compacted below the Ratis snapshot) is never mistaken for a fresh one. The `BootstrapElection` gate is widened to open on either the exact-`0` empty log (no-transfer path, unchanged) OR, for a positive commit index, on the no-application-entry signal. A negative (unready/failed) commit index still never opens the gate, so #4800 is preserved. The static `isConfirmedFirstFormation(long)` is left unchanged.

## Test plan
- [x] `ArcadeStateMachineFirstFormationSignalTest` (new, unit) - fresh SM is first formation; a persisted applied index closes the signal; the signal is durable across a simulated restart.
- [x] `BootstrapElectionTransferGateTest` (new, unit) - post-transfer leader (commitIndex 2, no application data) engages bootstrap; committed data or a null state machine skips.
- [x] `BootstrapElectionGateTest` (existing, unchanged) - all 8 first-formation gate cases still pass.
- [x] `RaftBootstrapLeadershipTransferIT` (existing IT) - end-to-end reproduction now passes (was timing out).
- [x] `RaftBootstrapDoesNotEngageOnRestartIT`, `RaftBootstrapFromLocalDatabaseIT`, `RaftBootstrapLateNewerJoinerIT` - no regression.

### DIFF
--- a/docs/5099-ha-offline-bootstrap-leadership-transfer.md
+++ b/docs/5099-ha-offline-bootstrap-leadership-transfer.md
@@ -57,6 +57,29 @@ an election" window is bounded by Raft's election restriction (a lagging node ca
 until its log is up to date) and backstopped by the existing `localLastTxId > baseline`
 refusing-to-overwrite guard.
 
+## PR
+https://github.com/ArcadeData/arcadedb/pull/5113
+
+## Review cycles
+- **Cycle 1** (`3f558b8`) - both bots flagged the same failure mode: a degraded/corrupt read of the
+  persisted `.raft/applied-index` file degrading to `-1` could re-open bootstrap on a running
+  cluster. Applied the file-existence guard (`8fba97a0`): the file exists only after an application
+  entry is applied, so its presence proves the node is not fresh regardless of parse result. Added
+  `presentButCorruptAppliedIndexFileKeepsTheSignalClosed`. Declined the alternative
+  `lastAppliedIndex >= commitIndex` guard (infeasible: internal entries advance commit but never
+  applied).
+- **Cycle 2** (`8fba97a0`) - claude LGTM; applied its clarifying comment on the fallthrough line
+  (`b723394e`). Gemini re-posted its already-applied suggestion.
+- **Cycle 3** (`b723394e`) - claude LGTM; applied all three polish items (`167a5be3`): residual-window
+  reasoning moved into the `isFirstFormation` code comment, `unwiredStateMachineReportsFirstFormation`
+  null-path test, and narrowed `hasNeverAppliedApplicationEntry()` to package-private. Gemini re-posted
+  the already-applied suggestion again.
+- **Cycle 4** (`167a5be3`) - claude verdict "looks good to merge"; all observations non-blocking and
+  already covered, no code changes. Gemini did not re-review this head.
+
+**Final state:** clean-approval (claude LGTM, no actionable items outstanding; working tree clean).
+Merge is the maintainer's decision - this workflow does not merge.
+
 ## Related
 - #4147 (offline bootstrap protocol), #4800 (first-formation gate), #5098 (transient `-1` fix,
   prerequisite), #5100/#5104 (restart baseline persistence).
--- a/docs/5099-ha-offline-bootstrap-leadership-transfer.md
+++ b/docs/5099-ha-offline-bootstrap-leadership-transfer.md
@@ -40,6 +40,23 @@ negative (unready/failed) commit index still never opens the gate (#4800 preserv
   machine it skips.
 - `RaftBootstrapLeadershipTransferIT` (existing IT) - end-to-end reproduction, now passes.
 
+## Review hardening (cycle 1)
+Both bot reviewers flagged the same failure mode: `hasNeverAppliedApplicationEntry()` could falsely
+report first formation on a running cluster if the persisted `.raft/applied-index` read degrades to
+`-1` (a transient I/O error or corrupt file). Added a defense-in-depth guard: the file is created
+only after an application entry is applied, so its mere presence proves the node is not fresh -
+`hasNeverAppliedApplicationEntry()` now returns `false` when the file exists, regardless of whether
+its contents parse. New unit case
+`ArcadeStateMachineFirstFormationSignalTest.presentButCorruptAppliedIndexFileKeepsTheSignalClosed`.
+
+The reviewer's alternative "require `lastAppliedIndex >= commitIndex`" is infeasible: internal
+no-op/config entries advance the commit index but never flow through `applyTransaction`, so applied
+can never catch up to commit in the transfer case (that is the whole reason the exact-`0` gate had to
+be replaced). The residual "never-applied node with committed-but-unapplied application entries wins
+an election" window is bounded by Raft's election restriction (a lagging node cannot win leadership
+until its log is up to date) and backstopped by the existing `localLastTxId > baseline`
+refusing-to-overwrite guard.
+
 ## Related
 - #4147 (offline bootstrap protocol), #4800 (first-formation gate), #5098 (transient `-1` fix,
   prerequisite), #5100/#5104 (restart baseline persistence).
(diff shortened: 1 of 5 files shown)


robfrank:
Thanks both - you converged on the same failure mode from two angles: `hasNeverAppliedApplicationEntry()` could falsely report first formation on a running cluster if the persisted `.raft/applied-index` read degrades to `-1`.

**Applied (commit 8fba97a0b):** @gemini-code-assist's file-existence guard. The `.raft/applied-index` file is written only after an application entry is applied, so its presence alone proves the node is not fresh. `hasNeverAppliedApplicationEntry()` now returns `false` when the file exists, regardless of whether its contents parse - a transient/corrupt read can no longer re-open the gate. New regression case `ArcadeStateMachineFirstFormationSignalTest.presentButCorruptAppliedIndexFileKeepsTheSignalClosed`.

**On the apply-lag window (@claude):** the alternative "require `lastAppliedIndex >= commitIndex`" is unfortunately infeasible here - internal no-op/config entries advance the commit index but never flow through `applyTransaction`, so `applied` can never catch up to `commit` in the transfer case (commitIndex=2, applied stays -1). That is the exact reason the exact-`0` gate had to be replaced. The file-existence guard closes the "previously-applied node re-elected" slice. The residual "never-applied node with committed-but-unapplied application entries wins an election" window is bounded by Raft's election restriction (a lagging node cannot win leadership until its log is up to date) and backstopped by the existing `localLastTxId > baseline` refusing-to-overwrite guard, so the worst case is a spurious bootstrap attempt that is refused, not data loss.

**On the `isConfirmedFirstFormation(0)` pin (@claude):** that static contract is already pinned independently by the unchanged `BootstrapElectionGateTest.onlyAnExactZeroOpensTheGate()`.

Verified after the change: `ArcadeStateMachineFirstFormationSignalTest` (4), `BootstrapElectionTransferGateTest` (3), `BootstrapElectionGateTest` (8, unchanged), and the `RaftBootstrapLeadershipTransferIT` / `RaftBootstrapDoesNotEngageOnRestartIT` ITs all green.

robfrank:
Cycle 2:

- **@claude** LGTM, thanks. Applied obs #1 (commit b723394e5): a comment on the final `readPersistedAppliedIndex() < 0` clarifying it is reached only when the file path is unresolvable or the file was absent at the existence check, so it does not read as dead code. Obs #2/#3 (javadoc link to the doc section, `.raft`-clean runbook note) noted - the residual-window reasoning is already captured in `BootstrapElection`'s class Javadoc and the tracking doc; I'd rather not add a non-resolvable `@link`. Obs #4 (per-call `Files.exists`) is bootstrap-election-time only, agreed no hot-path concern.
- **@gemini-code-assist** the suggested file-existence guard was already applied in commit 8fba97a0b (identical to your snippet, variable named `appliedIndexFile`), so this is already in place on the current head.

robfrank:
Cycle 3 (commit 167a5be37) - applied all of @claude's items:

1. **Residual-window note in code:** documented the committed-but-not-yet-applied window and its backstops (Raft election restriction + `localLastTxId > baseline` overwrite guard) directly in `BootstrapElection.isFirstFormation`, so a future reader sees it at the gate, not only in the docs note.
2. **Null-path test:** added `unwiredStateMachineReportsFirstFormation`, pinning the `getAppliedIndexFile() == null` fallthrough.
3. **Visibility:** narrowed `hasNeverAppliedApplicationEntry()` to package-private (sole caller is `BootstrapElection` in the same package; the Mockito test still stubs it at package scope).

Verified green after the change: `ArcadeStateMachineFirstFormationSignalTest` (5), `BootstrapElectionTransferGateTest` (3), `BootstrapElectionGateTest` (8, unchanged).

@gemini-code-assist your file-existence guard remains in place (commit 8fba97a0b) - the current head already contains exactly that check.

robfrank:
Cycle 4 - thanks @claude, noting "looks good to merge". The three observations are informational/optional and already covered:

1. Benign read/apply race - documented in the `isFirstFormation` residual-window note and backstopped by the overwrite guard; not restructuring.
2. Final `readPersistedAppliedIndex()` fallthrough - covered by `unwiredStateMachineReportsFirstFormation` and the inline comment.
3. Optional `decideAndAct`/transfer unit test - the transfer/commit branch is exercised end-to-end by `RaftBootstrapLeadershipTransferIT`; leaving as-is to avoid over-building.

No code changes this cycle. Leaving the merge decision to the maintainer.

PR #5134, opened 2026-07-08
Fix STRATEGY BREADTH_FIRST performing a depth-first traversal

saevarb:
BreadthFirstTraverseStep pushed newly discovered nodes onto the front of the entryPoints deque (addFirst) while consuming from the front (removeFirst), making it a LIFO stack: STRATEGY BREADTH_FIRST actually walked the graph depth-first. Its own prettyPrint even reported "DEPTH-FIRST TRAVERSE". The step has behaved this way since the first commit (OrientDB's BreadthFirstTraverseStep is a correct FIFO).

Push discovered nodes to the end (addLast) so entryPoints is a FIFO queue and the traversal is genuinely breadth-first, and correct the prettyPrint label.

The existing breadthFirst() test could not catch this: it ran on a linear chain (a->b->c->d), on which breadth-first and depth-first emit the identical sequence, so the STRATEGY had no observable effect. Replace it with breadthFirstOrder() and depthFirstOrder(), which run on a balanced binary tree (two branches of equal depth) and assert the emitted $depth sequence -- [0,1,1,2,2,2,2] for BFS vs [0,1,2,2,1,2,2] for DFS. Sibling order within a level is not asserted (TRAVERSE has no clause to control expansion order, and BFS does not guarantee it).

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/BreadthFirstTraverseStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/BreadthFirstTraverseStep.java
@@ -200,7 +200,7 @@ public class BreadthFirstTraverseStep extends AbstractTraverseStep {
 
   private void tryAddEntryPoint(final Result res, final CommandContext context) {
     if (whileClause == null || whileClause.matchesFilters(res, context)) {
-      this.entryPoints.addFirst(res);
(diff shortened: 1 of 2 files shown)


lvca:
Reviewed and verified end-to-end (fix + regression guard confirmed). Tracking issue: #5158. Closes #5158.

PR #5135, opened 2026-07-08
Add disabled reproducer for non-monotonic TRAVERSE MAXDEPTH

saevarb:
TRAVERSE ... MAXDEPTH d is documented to return the d-hop ball (nodes reachable within d hops), which must grow monotonically with d. On a 16-node DAG with both a short (depth 2) and a long (depth 4) path to the same node X, it instead returns 16 nodes at MAXDEPTH 3, 6 at MAXDEPTH 4, and 16 at MAXDEPTH 5 -- raising the bound drops X's whole subtree.

DepthFirstTraverseStep records a node's depth on first arrival with no relaxation when a shorter path arrives later, then gates subtree expansion on that recorded depth: at MAXDEPTH 4, X is first reached via the long path and stamped depth 4, the gate 4 > 4 fails, and X's ten children are never expanded. It is deterministic (fixed edge-insertion order makes the depth-first walk pop the long branch first); cycles are not required.

The test asserts the correct behavior (16/16/16) and is annotated @Disabled since the defect is not fixed here -- removing the annotation reproduces the failure ([16, 6, 16]).

### DIFF
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/TraverseStatementExecutionTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/TraverseStatementExecutionTest.java
@@ -20,6 +20,7 @@ package com.arcadedb.query.sql.executor;
 
 import com.arcadedb.TestHelper;
 import com.arcadedb.database.RID;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


lvca:
Thanks for the excellent reproducer and root-cause analysis, @saevarb - both confirmed exactly as described.

Rather than merge a permanently-`@Disabled` test (which would guard nothing in CI), @lvca has fixed the underlying bug and turned your reproducer into an active regression guard. Tracking issue: #5159.

**Fix:** depth relaxation (SPFA-style) in `DepthFirstTraverseStep` - track the smallest depth seen per node and re-expand its subtree whenever a strictly shorter path arrives, so every node reachable within d hops is expanded. Nodes are still emitted once; on single-path graphs no relaxation fires and the walk is identical to before. The d-hop ball content and count are now correct and monotonic.

**Verification:** your reproducer (un-`@Disabled`) now passes with [16,16,16] (was [16,6,16]); 312 traverse/match/select tests pass; cycles, self-loops and diamond graphs were probed for termination and monotonicity.

Note kept as documented behavior: for a node reachable by multiple paths, its reported `$depth` still reflects the depth-first arrival path (pre-existing DFS behavior); only the ball membership/count is corrected here.

Closing in favor of #5159, which carries the fix + your enabled reproducer. Credit for the find and the test is yours.

PR #5150, opened 2026-07-08
fix(#5149): reconcile cached count(*) counter on CHECK ... FIX

### LINKED ISSUE #5149: SQL `count(*)` returns stale cached bucket counter, disagrees with `count()` / real scan (undercount)
## Summary

On a type where records were inserted/deleted over time, `SELECT count(*)` returns a value lower than the actual number of records, while `SELECT count()` (and `count(<field>)`, a full table export, or iterating the result set) returns the correct value.

Reported on 26.7.1:

```sql
SELECT count()  FROM Activite   -- 676  (correct)
SELECT count(*) FROM Activite   -- 606  (wrong, 70 low)
```

Studio's row count also shows 606 (it uses `count(*)`), but exporting the full table via the Excel exporter yields 676 real rows. So `count(*)` is the outlier.

## Root cause

`count(*)` and `count()` take two different execution paths:

- **`count(*)`** matches the hardwired optimization in `SelectExecutionPlanner.isCountStar()` (requires the literal `count(*)`), which installs `CountFromTypeStep`. That calls `LocalDatabase.countType()` (`LocalDatabase.java:589`), summing each bucket's `LocalBucket.count()`. `LocalBucket.count()` (`LocalBucket.java:415`) returns a **cached** counter plus the current transaction delta when the cache is populated, instead of scanning:

  ```java
  final long cached = cachedRecordCount.get();
  if (cached > -1 && transaction != null)
    return cached + transaction.getBucketRecordDelta(fileId);
  ```

- **`count()`** (empty args) and `count(<field>)` do **not** match `isCountStar`, so they fall through to the normal aggregation path, which scans and counts every record. Always accurate.

`cachedRecordCount` is maintained *incrementally*, never recomputed during normal operation:

- insert -> `updateBucketRecordDelta(bucketId, +1)` (`LocalDatabase.java:1019`), delete -> `-1` (`LocalDatabase.java:1231`), and at commit the delta is folded in (`TransactionContext.java:~971`, and `TransactionManager.java:~623` on WAL replay/recovery).
- persisted to `<database-dir>/statistics.json` on clean shutdown and reloaded on open (`LocalSchema.readStatisticsFile`, `LocalSchema.java:1051`).
- fully recomputed only when a bucket's cache is `-1` (next `count()` scan repopulates it, `LocalBucket.java:452`) or when the DB was **not** closed cleanly (all buckets reset to `-1`, `LocalDatabase.java:2289-2291`).

If any record ever enters or leaves a bucket without its delta being recorded (or the persisted `statistics.json` becomes inconsistent), the counter drifts, is then persisted on the next clean close, and `count(*)` reports the wrong number on every subsequent open. The drift is silent.

## No safe reconciliation exists today

`CHECK DATABASE` / `CHECK TYPE ... FIX` computes the true `totalActiveRecords` in `LocalBucket.check()` (`LocalBucket.java:460`) but does **not** write it back into `cachedRecordCount`, so it does not repair the drift. The only ways to force a recompute are an unclean shutdown or manually deleting `statistics.json`.

## Impact

- `count(*)` (the most common way users count rows, and what Studio uses) can silently under/over-report.
- No user-facing command reconciles the counter.

## Workarounds

- Accurate count now: `SELECT count() FROM <Type>` or `SELECT count(@rid) FROM <Type>`.
- Repair the persisted counter: stop the server cleanly, delete `<server>/databases/<db>/statistics.json` (and legacy `cached-count.json` if present), restart. First `count(*)` then rescans and re-caches the correct value.

## Proposed fix

1. Make `CHECK TYPE ... FIX` (and `check(fix=true)` at the bucket level) write the recomputed active-record count back into `cachedRecordCount`, giving operators a supported reconciliation.
2. Add a regression test: insert N records, corrupt/desync `cachedRecordCount`, assert `count(*)` == `count()` after the check, and after a clean close/reopen cycle.
3. Optionally investigate whether a specific write path can drop a delta (bulk import, HA replay, interrupted commit) to eliminate the drift at the source rather than only offering repair.

## Environment

- ArcadeDB 26.7.1

robfrank:
Closes #5149

## Summary
`SELECT count(*)` is served from a cached per-bucket record counter (`LocalBucket.cachedRecordCount`, persisted in `statistics.json`), while `count()` / `count(<field>)` performs a real record scan. When the cached counter drifts from reality it is persisted as-is and `count(*)` silently under/over-reports on every subsequent open, with no user-facing command to reconcile it (`CHECK ... FIX` computed the true count in its report but never wrote it back).

This change makes `LocalBucket.check(fix=true)` invalidate the cached counter (`cachedRecordCount = -1`) after its page scan, so the next `count()`/`count(*)` rescans authoritatively and repopulates the correct value.

Invalidation is used instead of writing the freshly scanned value in place because `check(fix=true)` may run inside a caller-managed transaction: at commit `TransactionContext` folds that transaction's accumulated bucket delta into the counter, but only when it is `> -1`. Writing a scanned value would let unrelated inserts/deletes in that same transaction be double-counted on top of it; leaving `-1` makes the fold skip and defers to an authoritative recompute (which also survives a rollback and reuses `count()`'s own scan logic). `check()`'s own corrupt-record deletions go through `deleteRecordInternal`, which does not register a bucket delta, so they are not the concern. A narrow caveat (documented in the code) remains: no `count()` may run on the bucket between the invalidation and the checker's commit, which is fine for an admin operation.

## Test plan
- [x] `Issue5149CountStarCacheDriftTest.countStarDivergesFromScanWhenCachedCounterDrifts` - reproduces the bug: after corrupting the cached counter, `count(*)` returns 0 while `count(@rid)` returns the real 100.
- [x] `Issue5149CountStarCacheDriftTest.checkDatabaseFixReconcilesDriftedCountStar` - `CHECK DATABASE TYPE Customer FIX` repairs `count(*)` back to the real count.
- [x] `Issue5149CountStarCacheDriftTest.driftSurvivesCleanReopenAndIsRepairedByFix` - a drifted counter persisted to `statistics.json` survives a clean close/reopen (the real-world trigger) and is repaired by FIX.
- [x] `Issue5149CountStarCacheDriftTest.checkDatabaseWithoutFixDoesNotMutateCachedCounter` - a read-only `CHECK` leaves the cached counter untouched.
- [x] Regression: `CheckDatabaseTest`, `CheckDatabaseExtendedTest`, `GraphDatabaseCheckerDiagnosticsTest`, `CRUDTest`, `ReusingSpaceTest` all pass.


robfrank:
Filed the root-cause follow-up as #5152: the drift originates from `LocalBucket.count()`'s lazy `-1` recompute racing with concurrent commits (the recompute scans + `set()`s without the per-bucket file lock that commits hold, so a delta committed mid-scan is dropped by the fold's `> -1` guard). This PR's `CHECK ... FIX` invalidation remains a valid reactive repair; #5152 tracks the underlying fix.

PR #5171, opened 2026-07-09
bindings/python: align set_commit_every with engine #4961 (reject commit_every < 1)

tae898:
Follow-up to the async executor validation added in #4961.

The engine now rejects `commitEvery < 1` (0 previously slipped through and broke async task execution). This aligns the Python wrapper with that contract:

- `AsyncExecutor.set_commit_every()` raises a clear `ValueError` when `count < 1`, so callers get a Python-side error instead of a Java `IllegalArgumentException`.
- Updates the docstring, which previously described `0` as "manual commits only".
- Updates `test_importer_api` where the restore-settings test used the now-invalid `0` as its baseline (changed to a valid value; the test still verifies the executor applies and restores its settings).

Scope is limited to `bindings/python`. No engine changes.

### DIFF
--- a/bindings/python/src/arcadedb_embedded/async_executor.py
+++ b/bindings/python/src/arcadedb_embedded/async_executor.py
@@ -116,8 +116,9 @@ class AsyncExecutor:
         Auto-commit every N operations.
 
         Args:
-            count: Commit frequency (operations per commit).
-                   0 = manual commits only (default).
+            count: Commit frequency (operations per commit). Must be >= 1;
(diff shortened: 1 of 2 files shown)


tae898:
Thanks for the review. Addressed both points:
1. Removed the issue reference from the user-facing `ValueError` message; kept it as a code comment.
2. Added `test_set_commit_every_rejects_below_one` covering `0` and a negative value.

lvca:
Merged, thanks.

PR #5174, opened 2026-07-09
fix(#5023): idempotency cache - bind key to request identity, dedup in-flight retries, bound by bytes

### LINKED ISSUE #5023: [server audit] X-Request-Id idempotency cache replays wrong responses, skips writes, and grows unbounded
## Severity: HIGH (correctness + memory)

Part of the 2026-07 `server` module audit. Groups the correctness and the memory/CPU defects of the idempotency cache since both live in the same two classes.

## Affected code

- `server/src/main/java/com/arcadedb/server/http/handler/AbstractServerHttpHandler.java:146` (request-id used as log-correlation id) and `:250-270` (idempotency cache lookup/store)
- `server/src/main/java/com/arcadedb/server/http/IdempotencyCache.java:69-96` (`putSuccess`, `evictOldest`), `:116-130` (eviction scan)

---

## Defect 1 (HIGH correctness): cache key is the client `X-Request-Id` alone

The cache key is the raw `X-Request-Id` header — **not** bound to HTTP method, path, database, or body. The **same header is also reused as the log-correlation id** (`:146`), and many clients/proxies propagate one id across all calls of a logical operation.

### Failure scenarios
1. **Wrong-response replay / silently skipped write:** client sends `POST /command/dbA` with `X-Request-Id: abc`, then a *different* `POST /command/dbB` (or a different SQL command) reusing `X-Request-Id: abc`. The second request gets the first request's cached response replayed and **its write is never executed, while the client receives HTTP 200**.
2. **No in-flight dedup:** two concurrent retries with the same id both miss the cache and both execute → duplicate write (the exact case the cache exists to prevent).
3. **Caches uncommitted session work:** a command executed inside an open session transaction is cached as "success" (no `activeSession` check at `:266`). If the session later rolls back, a retry replays HTTP 200 for data that was never committed.
4. **`POST /begin` replay loses the session header:** the 204 is cached but headers are not, so a replay returns 204 **without** `arcadedb-session-id`; the original session is orphaned until idle timeout.

### Reproduction
```
# 1) create a row, capture the response
curl -u root:pw -H 'X-Request-Id: abc' -X POST localhost:2480/api/v1/command/dbA \
     -d '{"language":"sql","command":"INSERT INTO V SET n=1"}'
# 2) reuse the SAME id for a different write
curl -u root:pw -H 'X-Request-Id: abc' -X POST localhost:2480/api/v1/command/dbA \
     -d '{"language":"sql","command":"INSERT INTO V SET n=2"}'
# Observed: 2nd call returns the 1st response; row n=2 is NOT inserted.
```

### Fix
- Key on a hash of `requestId + method + path + database + body` (only then is replay-protection sound).
- Add an in-flight/pending marker (compute-if-absent with a PENDING state) so concurrent retries are deduplicated rather than both executed.
- Skip caching entirely when the command ran inside an active session transaction.
- If `/begin` responses are cached, cache and replay their headers too — or exclude `/begin`.

---

## Defect 2 (HIGH memory/CPU): eviction is O(n) per insert and unbounded in bytes

`IdempotencyCache.putSuccess` (`:93-95`) calls `evictOldest` once `size >= maxEntries`, which does a **full map scan** to find the single oldest entry (`:116-130`). The scan is not synchronized, so under load multiple inserting threads scan concurrently. Clients that send `X-Request-Id` on **every** request (common correlation practice) keep the cache permanently full, so every cached POST pays an O(maxEntries) scan.

The bound is **entry count, not bytes**: full `/command` response bodies (potentially MBs each) are retained for the TTL, so `maxEntries × avg-body-size` can reach GBs.

### Fix
- Replace the scan with an insertion/access-ordered LRU (`LinkedHashMap` with `removeEldestEntry` under a lock, or a segmented/Caffeine-style structure).
- Track and cap total cached **bytes**; skip caching bodies above a size threshold.

## Acceptance criteria

- Two requests with the same `X-Request-Id` but different method/path/database/body are treated as distinct (no cross-replay); a test asserts the second write executes.
- Concurrent identical retries execute the underlying command exactly once.
- Commands run inside an active session transaction are not cached.
- A load test filling the cache shows no O(n) eviction scan hotspot and a bounded byte footprint.
- Regression tests under `server/src/test/java/.../http/`.

## Notes for implementer

There is already a periodic TTL sweeper wired in `HttpServer` (`:143`), so time-based growth is handled — the issues are strictly the keying semantics, in-flight dedup, and the eviction/byte-bound. Do not remove the TTL sweeper.

robfrank:
Closes #5023

## Problem
Part of the 2026-07 `server` module audit. The HTTP idempotency cache had two defects (both in `IdempotencyCache` + `AbstractServerHttpHandler`):

1. **Correctness** - the cache key was the raw client `X-Request-Id` header alone, not bound to method, path, database or body. Since that header is commonly propagated by clients/proxies across a whole logical operation (and is also reused as the log-correlation id), a second, different `POST` reusing the same id had the first request's response replayed and **its write silently skipped (HTTP 200)**. It also cached work done inside an open client session transaction and replayed `/begin` without the `arcadedb-session-id` header.
2. **Memory/CPU** - `evictOldest` did a full-map O(n) scan on every overflow insert, and the bound was entry count only, so full response bodies (potentially MBs) accumulated unbounded in bytes.

## Fix
- Key idempotency on `SHA-256(requestId, method, path, database, body)` so a reused correlation id can no longer replay a different request's response or skip its write.
- Skip idempotency for session-scoped requests (incoming `arcadedb-session-id`) and for responses that establish a session (e.g. `/begin`), so a replay never drops the session header and uncommitted session work is never cached.
- Add `reserve`/`complete`/`abort` with a PENDING marker so concurrent identical retries are de-duplicated (execute-once) instead of both writing.
- Replace the O(n) eviction scan with O(1) FIFO eviction; add a total-byte bound and a per-body size cap via two new `SCOPE.SERVER` config keys (`arcadedb.ha.idempotencyCacheMaxBytes` default 64 MB, `arcadedb.ha.idempotencyCacheMaxBodyBytes` default 1 MB). The periodic TTL sweeper is retained.

Backward-compatible: the 2-arg constructor and `get`/`putSuccess` are kept; non-idempotent requests never touch the cache.

## Tests
- `IdempotencyCacheTest` (+7): reserve/complete/abort contract, in-flight dedup, await release, byte-bound eviction, per-body cap.
- `IdempotencyKeyTest` (new): composite key distinguishes method/path/database/body and is stable/field-delimited.
- `Issue5023IdempotencyKeyReplayTest` (new, end-to-end): same `X-Request-Id` + different command both execute (distinct response bodies); same id + same command is de-duplicated (identical replayed body). Run repeatedly for stability.
- Regression batch green: AutoCommitParameterTest, RequestIdSanitizationTest, CorrelationIdGenerationTest, AbstractServerHttpHandlerErrorBodyTest/TokenTest, ExecutionResponseTest, Issue5037BeginHandlerIT, HTTPAuthSessionIT, HTTPTransactionIT, Issue4141SessionManagementIT, HttpObservationIT.

### DIFF
--- a/docs/5023-idempotency-cache-keying-eviction.md
+++ b/docs/5023-idempotency-cache-keying-eviction.md
@@ -1,0 +1,41 @@
+# Issue #5023 - X-Request-Id idempotency cache: wrong-response replay, skipped writes, unbounded growth
+
+## Symptom
+The HTTP idempotency cache keys solely on the raw client `X-Request-Id` header (also reused as the
+log-correlation id). Clients/proxies that propagate one id across a logical operation get:
+- wrong-response replay / silently skipped writes (different method/path/db/body, same id);
+- no in-flight dedup (concurrent retries both execute -> duplicate write);
+- caching of work done inside an open client session transaction;
+- `/begin` replay that drops the `arcadedb-session-id` header.
+Plus O(n) per-insert eviction scan and an entry-count-only bound (unbounded bytes).
+
+## Root cause
+- `AbstractServerHttpHandler` keyed the cache on the raw request id alone.
+- `IdempotencyCache.evictOldest` did a full-map scan on every overflow insert; bound was entry count only.
+
+## Fix
+- Key idempotency on a SHA-256 of `requestId \0 method \0 relativePath \0 database \0 body`.
+- Skip idempotency when the request carries an incoming `arcadedb-session-id` (session-scoped work),
+  and skip caching when the response sets a session header (e.g. `/begin`).
+- Add reserve / complete / abort with a PENDING marker so concurrent identical retries execute once.
+- Replace the O(n) eviction scan with O(1) FIFO eviction and add a total-byte bound plus a per-body
+  size cap (skip caching oversized bodies). New config keys with backward-compatible defaults.
+- Keep the existing periodic TTL sweeper untouched.
+
+## Tests
+- `IdempotencyCacheTest` additions: reserve/complete/abort contract, in-flight dedup, byte bound,
+  per-body cap, O(1) eviction correctness.
+- `IdempotencyKeyTest`: composite key distinguishes method/path/db/body; stable for identical inputs.
+
+## Impact
+`server` module only. Backward-compatible constructor and get/putSuccess retained. Two new
+`SCOPE.SERVER` config keys (`arcadedb.ha.idempotencyCacheMaxBytes` default 64 MB,
+`arcadedb.ha.idempotencyCacheMaxBodyBytes` default 1 MB). The cache is now guarded by a monitor
+(one lock per idempotent POST); non-idempotent requests never touch it.
+
+## Verification
+- `IdempotencyCacheTest` (15), `IdempotencyKeyTest` (7), `Issue5023IdempotencyKeyReplayTest` (2) pass;
+  reproduction test run 4x for stability.
+- Regression batch green: AutoCommitParameterTest, RequestIdSanitizationTest, CorrelationIdGenerationTest,
+  AbstractServerHttpHandlerErrorBodyTest/TokenTest, ExecutionResponseTest, Issue5037BeginHandlerIT,
+  HTTPAuthSessionIT, HTTPTransactionIT, Issue4141SessionManagementIT, HttpObservationIT.
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1179,6 +1179,14 @@ public enum GlobalConfiguration {
       "Maximum number of entries in the HTTP idempotency cache. Oldest entry is evicted when full.",
       Integer.class, 10_000),
 
+  HA_IDEMPOTENCY_CACHE_MAX_BYTES("arcadedb.ha.idempotencyCacheMaxBytes", SCOPE.SERVER,
+      "Maximum total size in bytes of the cached response bodies in the HTTP idempotency cache. Oldest entries are evicted when exceeded.",
+      Long.class, 67_108_864L),
+
+  HA_IDEMPOTENCY_CACHE_MAX_BODY_BYTES("arcadedb.ha.idempotencyCacheMaxBodyBytes", SCOPE.SERVER,
+      "Maximum size in bytes of a single response body eligible for caching in the HTTP idempotency cache. Larger responses are not cached.",
+      Long.class, 1_048_576L),
+
   HA_PEER_ALLOWLIST_ENABLED("arcadedb.ha.peerAllowlist.enabled", SCOPE.SERVER,
       """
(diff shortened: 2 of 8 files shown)


robfrank:
Addressed the review feedback in the latest commit:

- **IO-thread blocking (high):** idempotent POSTs are now dispatched off the Undertow IO thread before any `await()`, so a concurrent-retry wait can never block an IO thread.
- **Binary replay data loss (high):** `replayCachedResponse` now sends a cached binary body via `ByteBuffer.wrap` instead of an empty string.
- **TTL sweep (medium):** `cleanupExpired` breaks at the first live entry (insertion order is ascending timestamp), so the periodic sweep is O(expired) rather than O(N).

Added `completeRetainsBinaryBodyForReplay` and `cleanupExpiredStopsAtFirstLiveEntry` unit tests. Full idempotency + session/transaction regression set stays green.

/gemini review

robfrank:
Addressed the follow-up review:

- **Blank X-Request-Id (high):** a blank/empty id is now excluded from idempotency in both the IO-thread dispatch guard and the reserve gating (`rawRequestId != null && !rawRequestId.isBlank()`), so distinct clients sending an empty header cannot collide on a shared composite key.
- **Eviction iterator (medium):** `evictToBounds` now allocates a single iterator and short-circuits when both bounds already hold.

Tests remain green.

/gemini review

robfrank:
Addressed cycle-3 review:

- **Stale reservation ABA (high):** `complete`/`abort` now take the owning `Reservation` and verify the exact PENDING marker instance is still installed before settling. If that marker was evicted/expired and a newer request installed a replacement, the slow original no longer releases the newer request's waiters - the execute-once guarantee holds. The handler tracks the `Reservation` and passes it through (including the finally cleanup).
- **Literal NUL in fallback key (medium):** the unreachable SHA-256-unavailable fallback used literal NUL bytes inside its string literals; replaced with explicit char delimiters (the digest itself keeps its byte-level NUL field separators).

Added `staleReservationDoesNotClobberReplacement` regression test; suite green.

/gemini review

robfrank:
Addressed cycle-4 review:

- **MessageDigest per request (medium):** the idempotency key now reuses a per-thread SHA-256 (`ThreadLocal` + `reset()`), so the request hot path no longer pays the JCA provider lookup of `MessageDigest.getInstance()` per idempotent POST. The unreachable per-call fallback is dropped (the thread-local initializer fails fast if SHA-256 is ever unavailable).

This completes the 4 planned review cycles. Across the cycles I addressed: composite keying, session/begin skips, in-flight dedup, byte/entry bounds and O(K) TTL sweep (initial); IO-thread dispatch + binary replay + O(K) sweep; blank X-Request-Id + single-iterator eviction; instance-safe reservation settle (ABA) + escaped fallback; and this thread-local digest. Full idempotency + session/transaction regression set is green.

PR #5177, opened 2026-07-09
fix(#5028): backup scheduler & retention correctness (cron OR semantics, validation, cancel race, keep newest)

### LINKED ISSUE #5028: [server audit] Backup scheduler & retention: wrong cron semantics, missing validation, cancel race, retention deletes newest backup
## Severity: MEDIUM (correctness / data protection)

Part of the 2026-07 `server` module audit. Groups the scheduling and retention defects in the backup subsystem.

## Affected code

- `server/src/main/java/com/arcadedb/server/backup/CronScheduleParser.java:85-129, 145-190`
- `server/src/main/java/com/arcadedb/server/backup/BackupScheduler.java:153-183`
- `server/src/main/java/com/arcadedb/server/backup/BackupRetentionManager.java:213-258`
- `server/src/main/java/com/arcadedb/server/backup/BackupTask.java:104-105, 177-184`

---

## Defect 1 (MEDIUM): cron parser uses wrong day-of-month / day-of-week semantics

Standard cron **ORs** day-of-month and day-of-week when both are restricted. This parser **ANDs** them (`getNextExecutionTime`, `:153-163`), so e.g. `0 0 2 1 * MON` fires only when the 1st of the month is a Monday — backups run far less often than the operator configured.

### Fix
Implement standard OR semantics for DOM/DOW when both fields are restricted.

---

## Defect 2 (MEDIUM): no range validation → invalid field silently stops all future backups; `x/0` hangs

- No bounds check: `7` for Sunday (very common) or hour `24` parses fine but can never match. `getNextExecutionTime` then throws `IllegalStateException` **inside the executor task's `finally`** (`BackupScheduler:160-162`), breaking the `scheduleNextCronExecution` chain — **all future backups for that database silently stop**.
- `parseField` with increment `0` (`"0/0"`, `:103`) loops forever (`i += 0`), hanging the scheduling thread at config load.

### Fix
Validate field ranges and require increment > 0 at parse time (reject the config with a clear error). Wrap the `getDelayMillis`/reschedule computation in try/catch so one bad cycle cannot kill the chain.

---

## Defect 3 (MEDIUM concurrency): cancel vs cron-reschedule race resurrects cancelled schedules

`cancelBackup` removes+cancels the future then removes the parser. A `BackupTask` already running can read `cronParsers.get(db)` **before** `cancelBackup` removes it, then call `scheduleNextCronExecution`, which `put`s a fresh future **after** the cancel completed — the schedule silently survives cancellation. A subsequent `scheduleBackup` for the same DB then leaves an uncancelled duplicate → **double backups**. The shutdown path can also throw an unhandled `RejectedExecutionException` in the worker.

### Fix
Guard reschedule and cancel with a per-database generation token (store a job object and have the task re-check `scheduledTasks.get(db) == myJob` via `compute`), or synchronize `cancelBackup`/`scheduleNextCronExecution` per DB; wrap the reschedule in try/catch for `RejectedExecutionException`.

---

## Defect 4 (MEDIUM): tiered retention can delete the backup that just completed

`applyRetention` (`:213-258`) keeps the **oldest** file per bucket and runs immediately after each backup (`BackupTask:104-105`). With tiered retention, any bucket keeps its oldest member, so the second-and-later backups of the same hour/day are deleted seconds after creation — **the newest restore point is discarded** and the effective RPO regresses to the start of the bucket.

### Fix
Always unconditionally keep the most recent backup file before applying tier selection.

### Also (minor, same file)
`BackupTask.performBackup` (`:177-184`): `database.isTransactionActive()` inspects the backup thread's own thread-local transaction, which is never active — the "backup during active transaction" warning is dead code and gives false confidence. Either remove it or check the right thing.

## Acceptance criteria

- Unit tests for `CronScheduleParser` covering: OR semantics for DOM/DOW, rejection of out-of-range fields (`hour 24`, `dow 7` if not supported), rejection of `x/0`, and a table of known-good cron strings vs expected next-fire times.
- An invalid cron config fails fast with a clear message rather than silently stopping backups.
- Cancelling a backup schedule cannot be resurrected by an in-flight task; no duplicate schedules after cancel+reschedule.
- Retention never deletes the most recent backup.
- Regression tests under `server/src/test/java/.../backup/`.

robfrank:
## What does this PR do?

Fixes the four scheduling and retention defects in the `server` backup subsystem reported in the 2026-07 server audit (#5028):

1. **Cron OR semantics** - `CronScheduleParser` now ORs day-of-month and day-of-week when both fields are restricted (standard cron), instead of ANDing them. Before, `0 0 2 1 * MON` fired only when the 1st of the month was a Monday.
2. **Field validation / fail-fast** - every field value is bounds-checked at parse time, day-of-week `7` is normalized to Sunday(0), non-positive increments (`x/0`) and malformed range/increment tokens are rejected. An invalid expression now fails fast with a clear message instead of silently never matching (or looping forever on `x/0`).
3. **Cancel vs cron-reschedule race** - reschedule/cancel are guarded by a per-database generation token under a single lock, so an in-flight `BackupTask` can no longer resurrect a cancelled or superseded schedule (no duplicate/double backups). The next-execution computation and `executor.schedule` are wrapped so an impossible schedule stops only that database (clear SEVERE log) and a shutdown `RejectedExecutionException` is swallowed.
4. **Retention keeps newest** - `BackupRetentionManager` always keeps the most recent backup before applying tier selection, so tiered retention can never delete the just-completed newest backup.

Also removes the dead active-transaction check in `BackupTask.performBackup` that inspected the backup thread's own thread-local transaction (never active), which gave false confidence.

## Motivation

These are correctness / data-protection defects: mis-scheduled or silently-stopped backups, and retention that discards the newest restore point.

## Related issues

Fixes #5028

## Additional Notes

- Regression tests added under `server/src/test/java/.../backup/`: `CronScheduleParserSemanticsTest`, `BackupRetentionNewestBackupTest`, `BackupSchedulerValidationTest`. They fail on the pre-fix source (verified: 10 failures + 1 error) and pass with the fix.
- No public API changes; existing backup tests are unchanged and still green (124 tests in the backup package pass).

## Checklist

- [x] I have run the build using `mvn clean package` command (scoped to the `server` module and affected tests)

### DIFF
--- a/docs/5028-backup-scheduler-retention.md
+++ b/docs/5028-backup-scheduler-retention.md
@@ -1,0 +1,29 @@
+# Issue #5028 - Backup scheduler & retention fixes
+
+## Symptom
+Four defects in the `server` backup subsystem:
+1. Cron parser ANDs day-of-month / day-of-week when both restricted (standard cron ORs them) -> backups run far less often than configured.
+2. No field range validation: out-of-range fields (hour 24, dow 7, dom 32...) parse but never match, and `getNextExecutionTime` throws `IllegalStateException` inside the reschedule chain -> all future backups silently stop. Increment `x/0` loops forever, hanging the scheduler thread at config load.
+3. Cancel vs cron-reschedule race: an in-flight `BackupTask` can resurrect a cancelled schedule, leaving duplicate schedules (double backups). Shutdown can throw `RejectedExecutionException` in the worker.
+4. Tiered retention keeps the OLDEST file per bucket and runs right after each backup, so the just-completed newest backup is deleted seconds after creation.
+   - Minor: `BackupTask.performBackup` checks `database.isTransactionActive()` on the backup thread's own thread-local tx, which is never active -> dead code / false confidence.
+
+## Root cause
+- `CronScheduleParser.getNextExecutionTime` uses AND for DOM/DOW and never tracks whether a field is restricted.
+- `parseField`/`parseValue` never bounds-check parsed values or reject non-positive increments.
+- `BackupScheduler.scheduleNextCronExecution` / `cancelBackup` have no generation token; reschedule put can win the race after a cancel. Delay computation and `executor.schedule` are unguarded.
+- `BackupRetentionManager.applyRetention` never unconditionally preserves the most recent file before tier selection.
+
+## Fix
+- CronScheduleParser: track `dayOfMonthRestricted` / `dayOfWeekRestricted`; OR them when both restricted, else AND (wildcards match everything). Validate every value is within `[min,max]`; normalize day-of-week `7` -> Sunday(0); require increment > 0; validate range order and increment token shape.
+- BackupScheduler: per-database generation token guards reschedule/cancel under a single lock; delay computation and `executor.schedule` wrapped in try/catch (impossible schedule stops only that DB with a clear SEVERE log; RejectedExecutionException on shutdown swallowed).
+- BackupRetentionManager: always add the most recent backup to the keep set before deletion.
+- BackupTask: remove the dead active-transaction check.
+
+## Tests
+- `CronScheduleParserSemanticsTest` - OR semantics, range rejection, `x/0` rejection, dow 7 = Sunday, known-good next-fire table.
+- `BackupRetentionNewestBackupTest` - newest backup always survives tiered retention.
+- `BackupSchedulerValidationTest` - invalid cron config fails fast (no schedule, no throw); cancel+reschedule leaves exactly one schedule.
+
+## Impact
+Scoped to `server` backup package. No API changes. Existing tests unchanged.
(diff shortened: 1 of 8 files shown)


robfrank:
Thanks for the review. Addressed both points in `8db9cb2`:

- **Day-of-week `7` (high):** Removed the per-value `7 -> 0` normalization inside `parseValue` (which broke ranges ending in 7, e.g. `1-7` was rejected as start > end). Day-of-week now parses over `0-7`, and bit 7 is folded onto bit 0 once after the whole field is parsed, so `1-7` (Mon-Sun) and the lone `7` both work. Added a regression test `dayOfWeekRangeEndingInSevenMeansEveryDay`.
- **Test flakiness (medium):** `BackupRetentionNewestBackupTest` now uses a fixed noon date so the three same-day backups can never split across a day boundary.

Issue #4083, opened 2026-05-05
Multiple warning over lost indexes on massive insert with gRPC


mdre:
When I do a massive insert in the cluster, the followers report a warning reporting thats index ar missings.

<img width="1912" height="1039" alt="Image" src="https://github.com/user-attachments/assets/2fa91621-0f63-4237-a848-cdddd9b648dd" />

lvca:
What release are your using?

mdre:
I'am using this morning pull. 
Rev: 95c57ba9bab7e4b890458725a1a41ead6e056265

lvca:
I was able to reproduce this issue.

The c374693ac fix only catches the case where compaction and recordFileChanges() deadlock-collide on the same FileManager recording session. This race is different: the leader's in-memory schema is updated by splitIndex() mid-compaction (file id assigned, mutable swapped), but the SCHEMA_ENTRY isn't dispatched until runWithCompactionReplication's replicateSchema() call after compaction.call() returns. During that window, repeated compactions on busy buckets can rename the file again, so by the time SCHEMA_ENTRY captures serializedSchema, it contains a name whose addFiles entry was already produced by an earlier compaction iteration that has already shipped a different nanos suffix. The result is what we see in the logs - the file id 13 was created with nanos 2631497441193500 but the schema captured nanos 2631497456962166.

lvca:
Just pushed a fix for this issue in the latest `main` branch. Please @mdre confirm.

mdre:
ok. I will check. Studing Italian now...

mdre:
I have pulled a fresh copy but I get this error when I try to compile.

`
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.15.0:compile (default-compile) on project arcadedb-network: Compilation failure: Compilation failure: 
[ERROR] /opt/arcadedb-git/network/src/main/java/com/arcadedb/remote/RemoteDocumentType.java:[49,8] com.arcadedb.remote.RemoteDocumentType is not abstract and does not override abstract method getIndexByProperties(java.util.List<java.lang.String>) in com.arcadedb.schema.DocumentType
[...]

Issue #689, opened 2022-12-19
Order edges in descending order


lvca:
In the current version 22.11.1, the edge list is a linked list where the order of insertion is not maintained. It's neither ascending nor descending, but close to random (Technically the edges are appended in a chunk, so ordered in the chunk, but when the chunk is full, a new chunk is created as the first chunk, so the order is messed up from the moment the 2nd chunk is created).

It's a very common use case retrieving the edges in ascending or descending order of the insertion. Creating ascending order is quite expensive with the current implementation while modifying the current code to support the descending order it's trivial.

PR #5209, opened 2026-07-10
feat(#5156): adaptive striped edge list for super-node write throughput

### LINKED ISSUE #5156: Adaptive striped/sharded edge list for super-node write throughput (#5148 follow-up)
## Summary

Follow-up to PR #5148 (super-node write contention). That PR removed the retry storm (commutative edge-append merge) and fixed two lost-update races (#5147, #5153), but the **raw write throughput on a single hot vertex is still capped** because every append to a vertex lands on the *same* edge-list head chunk = one page = one serialised MVCC version chain.

This issue is the remaining lever: let a hot vertex's edge list **span multiple pages/files (stripes)** so concurrent appends run in parallel.

## Motivation (measured in #5148)

3-node HA benchmark, 8 threads appending into one hub:
- All appends serialise on the hub's single head-chunk version chain → **~52 edges/s**, regardless of the merge.
- A distinct-target control (no shared hub) on the same cluster → **~94 edges/s** (and higher with less bucket-insert contention).

So the single hot chunk costs ~45% of throughput even with zero retries. Striping is what closes that gap. (The merge and the #5147/#5153 fixes handle the retry-storm and data-loss halves; this is the throughput half.)

## Design (from `docs/supernode.md` §4)

Let a single hot vertex keep a head chunk in **several** edge buckets and hash appends across the stripes. ArcadeDB already keys edge buckets per vertex bucket (`getEdgesBucketName`), so the files exist.

- **Adaptive, not blanket.** Degree is power-law - the vast majority of vertices have a handful of edges. Striping every vertex would bloat the vertex header (2 → 2N head pointers), scatter tiny chunks, and regress the hottest read paths (`out()`/`in()`/degree/`isConnectedTo`, which would each open N stripe iterators). Instead, **promote a vertex to a striped representation only when it crosses a degree threshold** (ConcurrentHashMap-treeify style) via a small "stripe directory" record. Small vertices and existing databases keep the current single-list format unchanged (format-versioned).
- **New config `SUPERNODE_THRESHOLD`** (e.g. default 128, tunable) as the promotion barrier. This gates the striped *structure* only - the edge-append merge stays unconditionally on (its overhead is ~0.7% and contention is not a function of degree).
- **Per-bucket / per-page granularity** - MVCC conflicts are page-level, so stripes must live on distinct files/pages or they still false-share.
- **Deterministic hash by the NEIGHBOUR vertex RID** (the second element of each `(edgeRID, vertexRID)` entry), NOT the edge RID. `isConnectedTo`, `containsVertex`, `getFirstEdgeConnectedToVertex`, and `removeVertex` all key on the neighbour vertex, so hashing by it localises them to one stripe (O(degree/N)); appends still spread because a hub's edges come from many distinct neighbours; and `deleteEdge` still resolves the stripe from the edge's endpoints. Hashing by edge RID would give zero read benefit on those hot paths. (Only a bare `containsEdge(edgeRID)` can't localise - uncommon.)
- **Reads merge N stripe chains** (and can parallelise); degree = sum of per-stripe counts. **Edge iteration order changes** (per-stripe order, merged) - confirm nothing guarantees edge order and document it.
- **Composes with the append-merge**: the merge kills the retry storm today; striping unlocks throughput. Both stay on.

## Open questions / risks

- **Promotion detection without new contention.** Do NOT add a per-append degree counter on the vertex record - it would make every append rewrite the vertex record and reintroduce the very contention we're removing. Use an approximate signal computed where the vertex is already touched (e.g. at a chunk-full transition), not per edge.
- **Stripe count N**: fixed (e.g. 8/16) to start; adaptive N (scale with degree) is a later refinement.
- **Format/versioning + migration**: the "striped" marker + stripe-directory record; old vertices and DBs read as single-list; promotion is lazy on write.
- **Backward-compat reads** must handle both single-list and striped layouts.

## Acceptance

- Concurrent appends to one super-node scale beyond the single-chunk ceiling (target: close the 52→94 eps gap on the HA benchmark; re-run `SuperNodeConcurrentAppendHABenchmark` / `SuperNodeConcurrentAppendBenchmark`).
- No regression for sub-threshold vertices (read paths unchanged, no extra pointers/chunks).
- Correctness under concurrency (extend the poison/anchor stress + HA consistency tests to the striped layout).
- Existing databases keep working without migration.

Design log: `docs/supernode.md` §4. Related: #5148 (merged), #5155 (remove-path anchoring cost).

lvca:
Implements the adaptive striped edge list (#5156): when a vertex's edge list crosses `arcadedb.graph.supernodeThreshold` (default 4096, 0 = disabled), it is promoted to a striped layout - a `StripeDirectory` (new record type 7) spreading further appends over `arcadedb.graph.supernodeStripes` (default 16) per-type pool buckets, so concurrent appends on the same hot vertex take different per-file commit locks and replicate in parallel under HA.

## Design
- **Dual layout, zero migration**: the vertex head pointer lands either on a classic `EdgeSegment` chain (record type 3) or a `StripeDirectory` (type 7); `RecordFactory` dispatches on the existing record-type byte. Old databases stay untouched; promotion is per-vertex, at chunk-full, driven by a geometric estimate (chain walk above the 8KB cap).
- **Stripe placement by neighbour RID hash** (frozen hash version persisted): `isConnectedTo`/`containsVertex`/`removeVertex` visit one stripe per generation instead of the whole list. Generational directory supports growing the stripe count later.
- **Pool DDL is deferred on server/HA** (SCHEMA_ENTRY before TX_ENTRY, #4083) and stripes are pre-warmed at promotion to avoid a directory-rewrite thundering herd.
- **Fast-path appends never touch the directory's commit lock**: the directory page is read without anchoring (a stale stripe head lands mid-chain, valid for the unordered list); anchored fresh reads happen only on slot writes.

## The bottleneck that made striping look useless - and three lost-update fixes
A commit-lock trace showed the REAL super-node serialiser under HA was the eager `vertex.modify()` in the connect paths: it anchored the target vertex page into EVERY append transaction's lock set, held across the full Raft round (~19 ms) - capping ALL layouts at ~52 edges/s. The connect paths now pass the immutable vertex down; `modify()` happens only where the vertex record is really rewritten (first chunk, head flip, promotion), guarded by stale-head re-checks.

Widened concurrency tests (merge-disabled runs, big-single-transaction promotion) then surfaced three silent lost-update gaps, all fixed and regression-tested:
1. **Deferred-update erase**: re-reading a record raw from its page mid-transaction resurrects the pre-update state and the next `updateRecord` erases earlier same-tx writes → new `TransactionContext.getWrittenRecord()` (written copy only, never the read-only cache).
2. **Pre-anchor read at post-anchor version**: a chunk buffer read before `fetchPageInTransaction` can pair a one-version-older buffer with the fresh page version - MVCC passes, concurrent append silently erased → chunk loads re-read through the anchored page.
3. **Head-reset on transient miss**: cross-file commit publication is not atomic for readers, so a head RID can be transiently unresolvable; the factory previously reset the head, orphaning the whole list → now a retryable `ConcurrentModificationException` everywhere.

Also fixed (pre-existing): `commit1stPhase` wrapped `LockTimeoutException` (a `NeedRetryException`) into a non-retryable `TransactionException`.

## Measured (3-node Raft, 8 writers into one hub, warm-started)
| configuration | throughput | avg / max commit latency |
|---|---|---|
| classic hub | 51 edges/s | 156 ms / 0.8 s |
| striped, 8 | 79 edges/s | 100 ms / 0.5 s |
| **striped, 16 (default)** | **104 edges/s** | **75 ms / 0.33 s** |
| striped, 32 | 117 edges/s | 67 ms / 0.27 s |
| control (no shared hub) | 88-104 edges/s | 91 ms / 0.6 s |

## Verification
- `com.arcadedb.graph` package: 271/271 green, incl. new `SuperNodeStripingTest` (promotion, reopen, deletion, drop-type, concurrency, big-single-transaction promotion, 100K add/delete-all residue audit)
- Merge-disabled lost-update stress: 5/5 classic + 10/10 striped runs green
- `SuperNodeAppendHAConsistencyIT`: 3-node cluster, replicas byte-identical
- HA benchmark gains warm-up isolation, progress logging and a `-DsupernodeStripes` knob

## Trade-offs (release-note material)
- Edge iteration order on PROMOTED vertices is approximate (newest-generation-first), relaxing the #689 reverse-insertion guarantee for super-nodes only.
- `GraphBatch` bulk import into an already-promoted vertex fails loud with a clear message.

Closes #5156


lvca:
Review feedback addressed in the last two pushes:

**Gemini** (both findings fixed in e525b97e8):
- StripeDirectory lazy-load: every accessor now self-loads (checkForLoading) and getContent() resolves the size after a lazy load; new regression lazyLoadedDirectoryIsReadable
- Promotion chain walk bounded (4096 cap-size chunks ~= 4M edges) so a corrupted cycle longer than a self-loop cannot hang the thread

**Claude** (c8307b179):
1. Retryability blast radius verified: the HTTP handler catches NeedRetryException (503, retryable) before TransactionException, so lock timeouts now reach the correct arm - server clients get a retry signal where they previously got a generic transaction error. Called out as release-note material in docs/supernode.md.
2. Lazy-modify blast radius: the whole com.arcadedb.database.async package plus the SQL edge-creation/MVCC/transaction suites are green locally; the full Java CI on this PR covers the rest.
3. ensureStripePool: full exception now logged; the POOLS_IN_CREATION pinned-on-hang trade-off is documented at the guard (a hung schema op means bigger problems, and promotion is a pure optimisation - classic keeps working).
4. Generation-count byte: no generation-append code exists yet; the growth follow-up will add the overflow guard where generations are appended.
5. #689 order + 6. per-type file impact: both documented (config description + release-note material section).
7. Import order fixed. 8. Done. 9. Importer-docs note deferred to the docs repo.

Review-suggested tests added: appendsSpreadAcrossStripes (stripe balance, guards stripeOf against concentration regressions) and promotedVertexKeepsWorkingWithPromotionDisabled (promotion is one-way; threshold=0 only stops future promotions). The embedded pool-creation-failure test was skipped: forcing schema.createBucket to fail requires injection the test framework does not offer; the failure path is a log-and-skip that leaves the vertex classic and functional by construction.

lvca:
Second review pass addressed in 07bb66db7:

**3 (direction-agnostic pool)**: deliberate, now documented on stripeBucketName - it caps the file count per type, chunks are only ever pointer-reached so mixing directions in one bucket is harmless, and real hubs are typically hot in a single direction. Encoding the direction would double the pool; the stripe count is the tuning lever if a dual-direction hub ever needs it.

**4 (minors)**: tx local reused in getOrCreateEdgeList; ensureStripePool now logs a one-time WARNING when a pool creation has been in flight for over a minute (stuck vs merely deferred is now diagnosable - the in-flight guard records its start time); getContent() re-entrancy assumption documented.

**1 + 2 (forward-incompatible format, default ON)**: the incompatibility is now explicit in the config description and in docs/supernode.md as a breaking-change callout: the first promotion writes record type 7, making the database unreadable by pre-26.8.1 releases, one-way; threshold=0 keeps full compatibility. Whether 26.8.1 ships default-on (4096) or default-off (opt-in for a cycle) is @lvca's release decision - both positions are one line in GlobalConfiguration; the callout text covers either.

lvca:
Third review pass addressed in d87e2cb66:

**1 (self-heal to hard-fail)**: intended and now regression-tested. Confirmed GraphDatabaseChecker never relied on the factory reset - the checker nulls the broken head itself (six set*EdgesHeadChunk(null) sites) before reconnecting through the factory's null-head branch, so repair is unaffected. New test missingHeadChunkFailsRetryablyAndCheckDatabaseRepairs covers the genuinely-missing-head case end to end: write fails with a retryable conflict (never the silent orphaning reset), CHECK DATABASE FIX repairs, and the repaired list accepts new edges.

**2 (read-path asymmetry)**: documented as deliberate best-effort read semantics at addChain (reads skip a transiently unresolvable chain; writes surface the same condition as a retryable conflict) and the skip WARNING is throttled to one per minute.

**3 (hung pool creation)**: a creation in flight for over 5 minutes is presumed dead and re-attempted from a later chunk-full (createStripePool tolerates the duplicate run - exists-check + lost-race tolerance), with the one-minute stuck WARNING kept. A deterministic pool-creation-failure IT would need failure injection into Schema.createBucket which the harness does not offer; the deferral contract (returns false, promotion retries at a later chunk-full) is exercised implicitly by SuperNodeAppendHAConsistencyIT, where promotion always lands after the deferred creation.

**4**: POOLS_STUCK_WARNED cleared alongside the in-flight guard. **5**: dropVertexType sweeps the full range best-effort instead of stop-at-first-gap, tolerating per-bucket failures. **6**: comment reworded - at thresholds beyond the cap-derived estimate (incl. the 4096 default) cap-chunk-fulls take the bounded walk until promotion. **7**: agreed, inherent.

**CI note**: the dorny reporters cannot publish check runs (workflow grants contents: read only) and mvnw runs with --fail-never, so job status = reporter step. Root-caused the reported failures: CheckDatabaseTest was genuinely broken by the new 4096 default (its 10K-edge root promoted, breaking classic-layout accounting assertions) - pinned to classic in d87e2cb66; ApplyChangesPartialReplayTest / Issue4510ForceApplyPartialDeltaTest / Issue4712ReplicatedWriteLockTest / LSMVectorIndexRebuildTest fail in rotating sets on main's own runs all day (untouched code, flush-timing flakes on CI runners) and all pass on this branch locally.

lvca:
Fourth review pass addressed in af0ea2f84:

**1 (dropVertexType 1024 cap)**: good catch - the sweep now covers max(1024, configured stripes), so a pool created under a larger setting cannot leak buckets past the fixed cap.

**2 + 4 (read under-report, GraphBatch)**: both spelled out in the upgrade-notes section of docs/supernode.md next to the forward-compatibility callout: promoted-vertex reads are best-effort under write concurrency (transient under-report while a concurrent cross-file commit publishes - impossible in the single-file classic layout), and GraphBatch hard-fails on organically-promoted targets.

**3 (gen-0 O(threshold) scan)**: correct - the localisation benefit applies to post-promotion edges only; gen 0 stays one unhashed chain bounded by the threshold. That is the price of zero-migration promotion; a future re-stripe generation could fold gen 0 in if it ever matters.

**Style**: AtomicLong imported and the field moved to the top of the class. On the two test conventions: the explicit save/restore in SuperNodeStripingTest predates my noticing TestHelper.afterTest() calls resetAll(); left as-is since it is harmless and self-documenting.

**Questions**: the POOL_CREATION_RETRY_MS re-attempt is defensive - deterministically hanging a schema.createBucket needs injection the harness does not offer, and the path is two lines past the timestamp comparison. Non-power-of-two distribution is now locked in by appendsSpreadAcrossNonPowerOfTwoStripes (12 stripes, same fair-share assertion as the 16-stripe test).

lvca:
Post-merge review round addressed in 9a6311e4f (consolidating the three overlapping review comments):

**Silent removal loss on a transient chain miss** - confirmed and fixed, the one real correctness catch of this round. Chain loading now splits by caller intent: mutating paths (removeEdge, removeEdgeRID, removeVertex, deleteAll) AND the neighbour-keyed lookups that feed MERGE-style dedup decisions (isConnectedTo, containsVertex, getFirstEdgeConnectedToVertex) surface the transient miss as a retryable ConcurrentModificationException, symmetric with loadStripeHead; bulk reads (count, iterators, containsEdge) stay best-effort as documented. New regression transientlyMissingStripeChainIsRetryableForWritesBestEffortForReads drops a stripe head and asserts both halves of the contract plus repair via CHECK DATABASE FIX.

**Fast path in multi-append transactions** - the instance directory is refreshed after every slot write, so addAll/promotion appends keep the fast path instead of detouring through the slot-write resolve.

**dropVertexType sweep** - now exact with no cap: pool buckets are contiguous by construction (createStripePool loops 0..N-1), so walking to the first gap at/past the configured count covers a pool of any size, past or present.

**Minors**: stripes < 2 disabling promotion is now in the config description; getContent() documents why cross-thread sharing is structurally impossible (per-transaction record caches, serialisation on the owning tx thread at commit); the docs state the neighbour-lookup cost precisely as O(threshold + degree/N) - generation 0 is the whole pre-promotion chain. The default-on decision was confirmed by @lvca and is recorded in docs/supernode.md; the JVM-global warn throttle is accepted as log-noise control.

All affected suites green locally: striping 15/15, append-merge 3/3, Issue5147, Issue5155 anchoring 2/2, BasicGraphTest 19/19, InsertGraphIndexTest, CheckDatabaseTest 9/9.

lvca:
Sixth review pass addressed in 0d10a4065:

**3 (pool name collision - the real catch of this round)**: fixed with an ownership guard at both gates. ensureStripePool's readiness check and createStripePool now refuse any bucket that matches the pool naming scheme but is OWNED BY A TYPE - promotion is skipped with a WARNING naming both types, and the vertex stays classic and fully functional. Regression poolNameCollisionWithUserBucketSkipsPromotion locks it in (user bucket named Hub_sn_stripe_0 attached to another type; 300-edge hub stays classic, counts and lookups exact). A bare unowned user bucket remains indistinguishable from a legitimate pool leftover by design - reserving the infix at creation time would be the belt-and-braces follow-up if you want it.

**5 (lazy-load publication race)**: checkForLoading is now synchronized - a lazy placeholder from the generic factory can be shared across reader threads; the monitor makes the load-then-publish of the non-volatile fields safe. Uncontended cost is a thin lock next to a page access.

**4 (drop sweep cost)**: slot-0 short-circuit - pools are created from slot 0, so a type that never promoted skips the sweep entirely.

**6 (embedded inline DDL)**: the branch now documents why mid-transaction bucket creation is safe embedded (immediate local schema, durability independent of the open data transaction, rollback residue = empty reusable buckets) vs the deferred HA path (#4083 replication ordering).

**1 (default-on)**: confirmed by @lvca on this PR and recorded in docs/supernode.md as the release decision, with the rollback caveat spelled out for the release notes. **2 (best-effort reads)**: in the upgrade-notes section of docs/supernode.md; will land in the user-facing release notes with the other promoted-vertex semantics. On the suggested self-heal test: the transient window cannot be created deterministically without publication-order injection (the strictness regression simulates it via actual deletion, which never self-heals) - the strict/best-effort split itself is covered by transientlyMissingStripeChainIsRetryableForWritesBestEffortForReads.

Suites green: striping 16/16, append-merge, Issue5147, Issue5155 anchoring, CheckDatabaseTest.

lvca:
Seventh review pass addressed in 7f84e1217:

**1 (uncontended hot-path cost)** - measured, since the question deserved numbers. singleThreadedAppendTrackingOverhead, 200K appends into one hub in one transaction, same machine, sequential runs: main 144.7K edges/s @ 5397 B/edge; this branch with the classic layout (threshold=0) 135.4K @ 5534 B/edge; with the shipped default (promotes at 4096) 131.7K @ 6499 B/edge. The classic delta (+2.5% alloc, -6% eps single-run, within observed +-10% run variance) reflects that in a transaction the read-through-anchor copy happens once per CHUNK first-touch, not per append - getWrittenRecord serves every subsequent append from the transaction's own copy. The striped default pays ~1.1KB/edge for the directory machinery against the measured 2.3x HA win under contention.

**3 (deferred DDL thread)**: (a) schema mutations are internally serialised by LocalSchema's own locking and are not bound to any request thread; (b) a close() mid-flight makes createStripePool throw into the catch-log-skip path - no state is mutated on failure, promotion just stays deferred. A shared executor would be the right home if this were recurring work; it is a once-per-type lifecycle event, which is why it carries the NOTE (concurrency) tag per the CLAUDE.md rule rather than a pool.

**4 (#689 sensitivity)**: the #689 regression test pins classic explicitly; Gremlin/Cypher make no edge-order guarantee without ORDER BY, and getFirstEdgeConnectedToVertex has no documented 'newest' contract (striped returns the oldest-generation match first). No order-sensitive engine path found relying on newest-first for high-degree vertices.

**Tests**: concurrentAppendsOnPromotedVertex now tolerates retry-exhausted transactions and gates on the exact final degree accounting for them; dropTypeRemovesStripePoolCreatedUnderLargerSetting covers the shrunk-setting sweep. The deferred-pool unit test needs a wrapped (server) database, unavailable to engine tests - covered by the HA IT as before. An old-binary-fails-cleanly test cannot run an old release in-suite; the failure mode (DatabaseMetadataException on the unknown record type) is fixed by the old code itself.

**Minors**: the stripes-default drift is in an immutable pushed commit message (the code, PR body and docs all say 16); GraphBatch's extra head lookup hits the per-transaction record cache on the bulk path; getGenerationCount now documents the single-byte bound.

Suites green locally (striping 17/17 incl. the two new tests).

Issue #4647, opened 2026-06-16
antlr-runtime dependency of arcadedb-engine is incompatible with hibernate 6 which requires antlr-runtime 4.10 or above


cmettier:
The version of antlr-runtime currently required by arcadedb-engine (4.9.1) is incompatible with a number of other thirdparty products used with springboot 3 / spring 6, such as hibernate 6. The runtime error when invoking arcadedb code is:
java.lang.UnsupportedOperationException: java.io.InvalidClassException: org.antlr.v4.runtime.atn.ATN; Could not deserialize ATN with version 3 (expected 4).

Because of this issue, projects which use hibernate and/or many other libraries with springboot 3 / spring 6 are not compatible with arcadedb-engine.

The fix needed to address this issue is for arcadedb-engine to start using antlr-runtime 4.10 or above (such as 4.13.2 which is latest at the time of writing this).

For a better explanation, you can read the following antlr4 issue: https://github.com/antlr/antlr4/issues/4568
Where you replace all references of "Apache Spark" in the entire defect with "arcadedb-engine".

lvca:
Reverting my changes... Assigned to @robfrank. We could use the latest with the engine and shade Gremlin, or viceversa.... pros and cons.

robfrank:
The best solution is to move  load each module (gremlin, postgres, etc.) on its own class-loader so third party libraries will not clash. since we have a couple of weeks before 26.7.1 I'll explore this solution.

cmettier:
Given that the best solution mentioned by @robfrank is taking longer to implement than originally anticipated, is it possible to get a July delivery of the original fix from @lvca, which would later be replaced by the one @robfrank is currently working on?

robfrank:
With https://github.com/ArcadeData/arcadedb/issues/5208 the issue should be fixed. 
If you can test with latest 26.8.1-SNAPHOT that now SpringBott with HB8 works fine it would be great

cmettier:
@robfrank I will not get a chance to verify your fix in time, but please incorporate it into 26.8.1 milestone anyway.

PR #5216, opened 2026-07-11
fix(#5208): engine ANTLR 4.13.2 + Gremlin ANTLR relocation (coexistence)

### LINKED ISSUE #5208: Bump engine to antlr 4.13.2 and shade/relocate antlr in the Gremlin module
Sub-issue of #4647.

## Summary

Resolve the antlr version conflict from #4647 by letting `arcadedb-engine` adopt a modern antlr while keeping Gremlin on its required 4.9.1, via two coordinated changes:

1. **Engine → antlr 4.13.2.** Bump `<antlr4.version>` in `engine/pom.xml` (4.9.1 → 4.13.2). This regenerates the SQL/Cypher/Cypher25 parsers with v4-format ATN and converges the engine with what Spring 6 / Hibernate 6 pull, fixing the reported `InvalidClassException: Could not deserialize ATN with version 3 (expected 4)`. antlr is fully encapsulated in the engine (0 references in any other module), so only the ~9 hand-written parser files may need minor 4.9→4.13 API-drift edits.

2. **Gremlin: relocate its bundled antlr.** Add a single relocation to the gremlin module's existing `maven-shade-plugin` config:

   ```xml
   <relocation>
     <pattern>org.antlr</pattern>
     <shadedPattern>com.arcadedb.gremlin.shaded.org.antlr</shadedPattern>
   </relocation>
   ```

   The distribution already ships the gremlin `-shaded` uber-jar (which bundles antlr 4.9.1 + gremlin-language's generated parser + gremlin-core's visitors). The relocation rewrites all `org.antlr` references in one consistent pass. TinkerPop's antlr **stays 4.9.1** (its precompiled parser needs v3 ATN). The `gremlin.language.grammar` package is **not** relocated (no clash; `ArcadeGraph`'s `VariableResolver` import stays valid).

### Resulting classpath (shipped distribution)

| Component | antlr package | version |
|---|---|---|
| arcadedb-engine + host app (Hibernate/Spring) | `org.antlr.v4.runtime` | 4.13.2 (real) |
| arcadedb-gremlin (shaded jar) | `com.arcadedb.gremlin.shaded.org.antlr.v4.runtime` | 4.9.1 (private) |

No collision by construction.

## Acceptance / tests

- Engine SQL + Cypher parser suites green after regeneration (API-drift safety net).
- Shaded gremlin jar contains `com/arcadedb/gremlin/shaded/org/antlr/…` and **no** `org/antlr/…`.
- Existing gremlin test suite green (relocated parser still deserializes its v3 ATN).
- **Coexistence test (decisive for #4647):** in one JVM, run engine SQL + Cypher parses (antlr 4.13.2) **and** a Gremlin traversal (shaded 4.9.1) - both succeed, no ATN error.

## Deferred / follow-up

Maven-coordinate consumer isolation: the shaded artifact's pom still lists `antlr4-runtime:4.9.1` (parent sets `createDependencyReducedPom=false`), so a Maven user depending on both `arcadedb-gremlin` and `arcadedb-engine` could still pull real antlr 4.9.1 transitively. Does not affect the shipped distribution (direct jar copy) or the reporter's engine-only embedding. Follow-up can mark gremlin's `antlr4-runtime` `<optional>true</optional>`.

Design spec: `docs/superpowers/specs/2026-07-10-gremlin-antlr-shading-engine-bump-design.md` (branch `feat/4647-antlr-engine-bump-gremlin-shading`).

robfrank:
Sub-issue of #4647. Fixes #5208.

## Problem

`arcadedb-engine` shipped ANTLR parsers generated with **antlr 4.9.1** (ATN serialized in "version 3" format). Embedded in a Spring 6 / Hibernate 6 app, those frameworks pull **antlr-runtime 4.10+** (reads only "version 4" ATN). Whichever runtime wins on the classpath breaks the other side:

```
java.io.InvalidClassException: org.antlr.v4.runtime.atn.ATN;
Could not deserialize ATN with version 3 (expected 4).
```

TinkerPop 3.8.1 (Gremlin) is hard-locked to antlr 4.9.1 because it ships a **precompiled** parser (v3 ATN) that cannot be regenerated without forking TinkerPop, so a single antlr version cannot serve both the engine and Gremlin.

## Approach

1. **Engine → plain antlr 4.13.2** (unshaded). Parsers regenerate to v4 ATN, converging the engine with what Spring 6 / Hibernate 6 bring. The engine module stays an ordinary jar.
2. **Gremlin relocates its antlr** - `org.antlr` → `com.arcadedb.gremlin.shaded.org.antlr` inside the shaded jar - so TinkerPop's precompiled 4.9.1 (v3) parser keeps its runtime and never collides with the engine's 4.13.2. Only `org.antlr` is relocated; the TinkerPop grammar package is untouched.
3. **New `arcadedb-gremlin-it` module** runs the existing gremlin test suite (via the test-jar + surefire/failsafe `dependenciesToScan`) plus a new `AntlrCoexistenceIT` against the shaded jar, so the two antlr runtimes are exercised together in one JVM. The gremlin module itself skips test execution (still builds its test-jar) because engine-v4 + gremlin-v3 cannot share one unshaded classpath.

### Resulting classpath (shipped distribution)

| Component | antlr package | version |
|---|---|---|
| arcadedb-engine + host app (Hibernate/Spring) | `org.antlr.v4.runtime` | 4.13.2 (real) |
| arcadedb-gremlin (shaded jar) | `com.arcadedb.gremlin.shaded.org.antlr.v4.runtime` | 4.9.1 (private) |

The gremlin uber-jar **excludes** `arcadedb-engine` from its `artifactSet` so the relocation only ever rewrites TinkerPop's parser - not the engine's own v4 parsers (which the distribution provides as a plain jar). Without this, a bundled engine parser would be relocated onto the incompatible 4.9.1 runtime and could win under `-cp lib/*`, reintroducing the ATN error.

## Verification

- Engine SQL + Cypher parser suites green after regeneration.
- Shaded jar: 0 engine-parser classes bundled; relocated TinkerPop parser + antlr present; no bare `org/antlr` leak.
- `AntlrCoexistenceIT`: engine SQL + Cypher parse (4.13.2) and a Gremlin traversal (relocated 4.9.1) succeed in one JVM.
- Full migrated gremlin suite in `gremlin-it`: 2090 tests, 0 failures / 0 errors.
- Distribution: `SqlParser.class` present in exactly one `lib/` jar (plain engine).
- `mvn clean install -DskipTests`: BUILD SUCCESS across all modules.

## Deferred (documented in the design spec)

Maven-coordinate consumer isolation (`<optional>` on gremlin's antlr) for third-party embedders who depend on both `arcadedb-gremlin` and `arcadedb-engine` directly. Does not affect the shipped distribution or the reporter's engine-only embedding.

Design spec and implementation plan: `docs/superpowers/specs/2026-07-10-gremlin-antlr-shading-engine-bump-design.md`, `docs/superpowers/plans/2026-07-10-gremlin-antlr-shading-engine-bump.md`.


robfrank:
Addressed in d2d698c06. Details per item:

**Test robustness (VectorGremlinIT)** - Fixed. Now null-checks the classpath resource with a clear assertion message and resolves it via `new File(inputFile.toURI()).getAbsolutePath()` instead of `URL.getFile()`, so paths containing spaces or other special characters decode correctly. Verified `VectorGremlinIT` + `AntlrCoexistenceIT` green against the shaded jar in `gremlin-it`.

**gremlin-it depends on gremlin package-phase outputs** - Documented. Added a note to the `arcadedb-gremlin-it` POM `<description>` stating it consumes the shaded uber-jar and the tests test-jar, so it must be built after `arcadedb-gremlin` in the reactor and requires a prior install when run standalone.

**Relative `../gremlin/src/test/resources` path** - Intentional and already commented as the single-source-of-truth choice over byte-copies. Left as-is; it only breaks under a module move, which would touch the POM anyway.

**maven-enforcer `requireMavenVersion`** - Not adding it here. The build pins Maven **3.9.12** exactly via the wrapper (`.mvn/wrapper/maven-wrapper.properties`), and CI runs through `./mvnw`, so the `*:*` exclusion and `dependenciesToScan` always run on a compatible Maven. There is no existing `requireMavenVersion` rule in the parent POM; adding a repo-wide enforcer rule is out of scope for this ANTLR change and better handled as a separate hardening.

**AntlrCoexistenceIT CI visibility** - Confirmed it runs in CI. The `integration` profile sets `<skipITs>false</skipITs>`, and `.github/workflows/mvn-test.yml`'s `integration-tests` job runs `./mvnw verify -Pintegration ... -pl !e2e,!load-tests,!e2e-ha,!ha-raft` - `gremlin-it` is not excluded, so failsafe executes `AntlrCoexistenceIT`. The migrated `*Test` classes additionally exercise the same coexistence path in the surefire (unit) phase, so it is covered in both phases.

**Deferred consumer isolation** - Filed as follow-up issue #5217 (mark gremlin's `antlr4-runtime` `<optional>`/`provided`, and surface in release notes so embedders using both modules by coordinate consume the `shaded` classifier). Also documented in the design spec's "Deferred / follow-up" section.

**junit-vintage-engine version** - Left as `${junit.jupiter.version}` to match the existing `arcadedb-gremlin` module, which uses the same property for vintage; diverging only in `gremlin-it` would be inconsistent. Agreed it's non-blocking.

robfrank:
Addressed in 6aaddf3a3.

**Build/reactor ordering** - Documented. CI is safe: the unit-tests job runs `./mvnw verify -Pcoverage` (reaches the package phase, so gremlin's `shaded`/`tests` classifiers exist before `gremlin-it` builds). For the contributor-facing case you flagged, updated CLAUDE.md's Maven commands to use `mvn verify`/`install` (not bare `mvn test`) for a full reactor run, with a note explaining that `gremlin-it` consumes gremlin's package-phase artifacts.

**Hardcoded TinkerPop version** - Fixed. Promoted `gremlin.version` to the parent `pom.xml` properties and removed the local definition in `gremlin/pom.xml`; both `arcadedb-gremlin` and `arcadedb-gremlin-it` now reference `${gremlin.version}`. Verified `mvn help:evaluate -Dexpression=gremlin.version -pl gremlin-it` resolves to `3.8.1`, so a future bump updates one place.

**VectorGremlinIT residual space edge** - Acknowledged, left as-is. You're right that the decoded path is re-embedded into a whitespace-delimited `import database file://... with ...` SQL string, so a space in the path would still truncate at the SQL layer (the `toURI()` decode only fixes the URL side). Making it fully space-safe would require SQL-level quoting of the file URL, which the import grammar doesn't cleanly support here; since the path lives under `target/` it won't contain spaces in practice. The `toURI()` change still removes the `%20`-in-URL failure mode, which was the realistic one.

**Skips / minor notes** - The `skipTests` (not `maven.test.skip`) choice keeping test compilation + test-jar production, and matching failsafe to surefire, were deliberate for exactly the reasons you noted. JaCoCo attribution over relocated packages in gremlin-it may indeed read oddly; noted, not changing coverage config in this PR. `project.parent.version` usage stays consistent with the rest of the repo.

robfrank:
Good catch on #1 - it was a real regression. Addressed in 34d467577.

**1. Unshaded gremlin + engine 4.13.2 clash (graphql / postgresw)** - Confirmed and fixed. Running `mvn -pl graphql verify` reproduced exactly what you predicted: 14 errors, `Could not initialize class com.arcadedb.query.sql.grammar.SQLLexer` / `Cypher25Lexer` - the plain gremlin jar's antlr 4.9.1 won mediation and broke the engine's v4 parsers (even pure-SQL `GraphQLBasicTest` failed, not just the gremlin directive). Fixed by switching both `graphql` (test scope) and `postgresw` (provided scope) to the **shaded** gremlin jar with the same wildcard exclusion `gremlin-it` uses, so the only real `org.antlr` on those classpaths is engine's 4.13.2 and gremlin's is the relocated copy.
   - `graphql`: now green - 26 tests, 0 failures/0 errors, including `GraphQLGremlinDirectivesTest`, `GraphQLCypherDirectivesTest`, and the SQL directive tests.
   - `postgresw`: 0 antlr init failures now. One unrelated failure remains, `PostgresPreparedStatementMetricsIT.extendedProtocolSqlQueryIsTimedWithPostgresTag` (asserts a Micrometer timer count >= 1, got 0) - it has no gremlin/antlr involvement, reads the shared `Metrics.globalRegistry`, and **passes in isolation**; it's a pre-existing test-ordering/global-registry-state flake, out of scope here.

**2. `mvn test` over the full reactor** - Kept the documentation approach rather than gating `gremlin-it` behind `-Pintegration`. Gating it there would remove the migrated `*Test` (surefire) classes from CI's unit-tests job, which runs `./mvnw verify -Pcoverage` **without** `-Pintegration` and relies on `gremlin-it` building; the `-Pintegration` job runs with `-DskipTests` (failsafe only), so the migrated unit tests would then run nowhere. A loud fail-fast on bare `mvn test` (now documented in CLAUDE.md) is preferable to silently dropping unit-test coverage.

**3. Coverage / CI signal** - No build-failing coverage gate exists: the `coverage` module only runs JaCoCo `report-aggregate` + `merge` (no `check` goal or threshold rule), and aggregation scans `**/target/jacoco*.exec` repo-wide. The gremlin classes (`com.arcadedb.gremlin.*`) are **not** relocated - only `org.antlr` is - so their coverage attributes normally, now produced by `gremlin-it`.

**4. gremlin-it re-declares gremlin's test deps** - Added a note at the top of `gremlin-it`'s `<dependencies>` documenting that the `*:*` wildcard exclusions require mirroring any new gremlin test dependency here.

**Nits** - Added `maven.install.skip=true` to `gremlin-it`; strengthened `AntlrCoexistenceIT` to assert the Cypher `ResultSet` returns the expected `name` (not just `hasNext()`). Left `junit-vintage-engine` on `${junit.jupiter.version}`: the repo doesn't use a JUnit BOM and the gremlin module already versions vintage this way, so a BOM here would be inconsistent with the rest of the tree.

robfrank:
Addressed in 3e802ed3f.

**1. CI green with zero Gremlin tests (the merge ask)** - Fixed. Added `<failIfNoTests>true</failIfNoTests>` to `gremlin-it`'s surefire config, so a mis-wired `dependenciesToScan` (unattached test-jar, classifier/coordinate mismatch) now fails the build loudly instead of passing silently while the gremlin module hard-skips its own tests. Verified the guard is inert on the happy path: `mvn -pl gremlin-it test` still discovers and runs 236 migrated `*Test` (0 failures). (Failsafe always has `AntlrCoexistenceIT` in this module's own `src/test`, so the zero-test hole was specific to surefire.)

**2. VectorGremlinIT jar-URL trap** - Fixed. Replaced `new File(url.toURI())` with streaming the classpath resource to a temp file (`getResourceAsStream` -> `Files.copy`). This no longer depends on `getResource` returning a `file:` URL, so it's robust to classpath ordering handing back a `jar:` URL from the `arcadedb-gremlin` tests jar (which would otherwise throw `URI is not hierarchical`). Verified `VectorGremlinIT` green against the shaded jar.

**3. Hand-mirrored dependency list** - The `failIfNoTests` guard above converts the wholesale-breakage case (a missing dep taking out the scan) from a silent pass into a hard failure. The coupling is also documented in the `<dependencies>` comment added last round.

**4. Deferred Maven-consumer isolation** - Agreed it's the right follow-up to prioritize; tracked in #5217 (mark gremlin's `antlr4-runtime` `<optional>`/`provided`, plus a release-note heads-up for embedders using both modules by coordinate).

**5. DX regression running a single gremlin test** - Added a pointer in `gremlin/pom.xml`'s skip comment: run individual gremlin tests via `mvn -pl gremlin-it test -Dtest=GremlinTest` (or `-Dit.test=...` for `*IT`).

Thanks for the four rounds - the graphql/postgresw catch in particular was a genuine regression that `-DskipTests` had hidden.

Issue #4865, opened 2026-07-02
MCP: expose schema as a Resource


robfrank:
Part of #4859.

## Motivation

MCP Resources are entirely unimplemented (confirmed: zero matches for `resources/list`/`resources/read` in the mcp package; `capabilities` in `handleInitialize` advertises only `tools`). Serving schema as a Resource lets clients auto-load structure without spending a tool-call turn on `get_schema` every session. **Justification: #4 structured shaping, delivered via the correct protocol primitive rather than another tool** — directly addresses the tool-count-discipline principle from #4859.

## Proposed design

- Implement `resources/list` and `resources/read` in both `MCPHttpHandler` and `MCPStdioServer`.
- Advertise `capabilities.resources = {listChanged: false, subscribe: false}` in `handleInitialize`.
- Resource URI scheme: `arcadedb://{database}/schema`.
- `resources/list` enumerates one resource per database the authenticated user can access (respecting `MCPConfiguration.isUserAllowed`/`allowedUsers`).
- `resources/read` returns the same JSON `get_schema` currently produces — **refactor `GetSchemaTool`** to extract a shared static schema-building helper, called by both the tool and the resource handler (avoid duplicating logic).
- Keep the `get_schema` tool for backward compatibility; docs should note Resources as the preferred path for clients that support it.

## Acceptance criteria

- [ ] `resources/list` and `resources/read` respond correctly over both HTTP and stdio transports.
- [ ] `initialize` response advertises the `resources` capability.
- [ ] Schema content is identical between `get_schema` tool output and the resource's `resources/read` content (verified by a test asserting equality).
- [ ] Access control matches `get_schema`'s existing behavior (same users, same databases).
- [ ] Tests added to `MCPServerPluginTest`/`MCPStdioServerTest`; docs updated.

## Files to touch

`MCPHttpHandler.java`, `MCPStdioServer.java`, `server/src/main/java/com/arcadedb/server/mcp/tools/GetSchemaTool.java` (refactor for reuse), tests, docs.

## Dependencies

None — can proceed in parallel with all tool issues.

## Effort estimate

M — new protocol primitive across two transports, capability advertisement, refactor for shared logic.

Issue #4864, opened 2026-07-02
MCP: add upsert_entity and upsert_relationship tools


robfrank:
Part of #4859.

## Motivation

ArcadeDB SQL/Cypher already support UPSERT/MERGE semantics, but models reliably build match-key logic wrong (string-concatenated WHERE clauses, missing key normalization) and duplicate nodes as a result — the single most common agent-memory failure mode. Wrapping match-or-create-by-key behind a parameterized, injection-safe tool is a concrete correctness win over raw `execute_command`. **Justification: #1 ergonomic guardrail; secondarily #3 different permissioning** (requires composing `allowInsert` AND `allowUpdate`, a combination `ExecuteCommandTool` doesn't need to express as a single unit today).

## Proposed design

- Tool names: `upsert_entity`, `upsert_relationship`
- `upsert_entity` input schema:
```json
{
  "type": "object",
  "properties": {
    "database": {"type": "string"},
    "type": {"type": "string"},
    "matchKeys": {"type": "object", "description": "property:value pairs used as the match/merge key"},
    "setProperties": {"type": "object"}
  },
  "required": ["database", "type", "matchKeys"]
}
```
- `upsert_relationship` input schema:
```json
{
  "type": "object",
  "properties": {
    "database": {"type": "string"},
    "fromType": {"type": "string"}, "fromMatchKeys": {"type": "object"},
    "toType": {"type": "string"}, "toMatchKeys": {"type": "object"},
    "relType": {"type": "string"},
    "relProperties": {"type": "object"}
  },
  "required": ["database", "fromType", "fromMatchKeys", "toType", "toMatchKeys", "relType"]
}
```
- Implementation: build a **parameterized** Cypher `MERGE` (bound parameters, not string interpolation — this is the injection-safety justification, not just ergonomics) and execute via the same code path `ExecuteCommandTool` uses.
- Permission gating: extend `ExecuteCommandTool.checkPermission`-style logic to require `allowInsert` (create path) **and** `allowUpdate` (update path) both enabled; report explicitly which flag is missing on rejection.
- Forward-link: a future bi-temporal/temporal-KG primitive (versioned upserts) is out of scope here, but this schema shouldn't preclude adding an optional `validFrom`/`validTo` later.

## Acceptance criteria

- [ ] No duplicate node created on repeated calls with identical `matchKeys`.
- [ ] Relationship upsert does not duplicate edges between the same resolved endpoints.
[...]

Issue #4863, opened 2026-07-02
MCP: add sample_records tool


robfrank:
Part of #4859.

## Motivation

`get_schema` returns structure only, no value shapes. A cheap sampling call measurably improves every query an agent subsequently generates (correct enum-like values, date formats, actual property presence vs. schema-declared-but-unused properties). **Justification: #4 structured shaping.**

## Proposed design

- Tool name: `sample_records`
- Input schema:
```json
{
  "type": "object",
  "properties": {
    "database": {"type": "string"},
    "types": {"type": "array", "items": {"type": "string"}, "description": "omit for all types"},
    "limit": {"type": "integer", "default": 3, "maximum": 20}
  },
  "required": ["database"]
}
```
- Output: `{samples: {"<TypeName>": [{...record...}, ...]}}`
- Permission gating: `allowReads`.
- Implementation: per-type `SELECT FROM <Type> LIMIT <limit>` via the query engine (same idempotency path as `QueryTool`).

## Acceptance criteria

- [ ] Bounded `limit` (hard max) to avoid accidental large payloads.
- [ ] Skips/handles empty types gracefully (empty array, not error).
- [ ] Registered in both transports; tests + docs.

## Files to touch

`server/src/main/java/com/arcadedb/server/mcp/tools/SampleRecordsTool.java` (new), `MCPHttpHandler.java`, `MCPStdioServer.java`, tests, docs.

## Dependencies

None.

## Effort estimate

S — simple per-type `SELECT ... LIMIT` wrapper.

Issue #4861, opened 2026-07-02
MCP: add hybrid_search tool


robfrank:
Part of #4859.

## Motivation

`vector.fuse` already fuses vector + full-text scores server-side (RRF/DBSF/LINEAR), but there is **no single expression** that also expands over graph relationships (`out()`/`shortestPath()`) before or after fusion — that requires orchestrating multiple statements, which is error-prone and unsafe for a model to hand-write (risk of unbounded traversal, missing dedup). **Justification: primarily #2 unreachable-as-a-single-call orchestration, secondarily #1 ergonomic guardrail** for `vector.fuse` syntax.

## Proposed design

- Tool name: `hybrid_search`
- Input schema:
```json
{
  "type": "object",
  "properties": {
    "database": {"type": "string"},
    "vectorIndexName": {"type": "string"},
    "queryVector": {"type": "array", "items": {"type": "number"}},
    "k": {"type": "integer", "default": 10},
    "expand": {
      "type": "object",
      "properties": {
        "edgeTypes": {"type": "array", "items": {"type": "string"}},
        "direction": {"type": "string", "enum": ["out", "in", "both"], "default": "out"},
        "maxDepth": {"type": "integer", "default": 1, "maximum": 3}
      }
    },
    "fulltextIndexName": {"type": "string"},
    "fulltextQuery": {"type": "string"},
    "fusionStrategy": {"type": "string", "enum": ["RRF", "DBSF", "LINEAR"], "default": "RRF"}
  },
  "required": ["database", "vectorIndexName", "queryVector", "k"]
}
```
- Output: `{results: [{rid, fusedScore, path?, properties}]}`
- `maxDepth` is capped server-side (hard max, e.g. 3) to prevent runaway traversal — this is the specific unsafe-if-hand-written behavior the tool guards against.
- Permission gating: `allowReads` only (read-only pipeline).
- Engine functions used: `SQLFunctionVectorNeighbors`/`SQLFunctionVectorSparseNeighbors`, `SQLFunctionOut`/`SQLFunctionOutE` (`function/sql/graph/`), `SQLFunctionSearchIndex` (`fulltext.searchIndex`), `SQLFunctionVectorFuse`.

## Acceptance criteria

- [ ] Depth-limited traversal enforced server-side regardless of client-supplied `maxDepth`.
- [ ] Works with vector-only input (no `fulltextQuery`) and with vector+fulltext+expand combined.
[...]

PR #5263, opened 2026-07-13
feat(mcp): add full_text_search tool (#4862)

### LINKED ISSUE #4862: MCP: add full_text_search tool
Part of #4859.

## Motivation

Completes the lexical leg cleanly; `SEARCH_INDEX(indexName, query)` plus BM25 scoring (`LSMTreeFullTextIndex`, `BM25Scorer`) has query-syntax quirks models won't reliably reproduce. Lower standalone value than vector/hybrid — this issue exists mainly to complete the retrieval trio and to give `hybrid_search` (#4861) and future prompts a documented tool to reference. **Justification: #1 ergonomic guardrail.**

Note: ArcadeDB's full-text index is its own LSM-tree structure with BM25 scoring — Lucene is used only for tokenization/analysis, not as the storage/retrieval engine. Docs for this tool should say "BM25 full-text index," not "Lucene index."

## Proposed design

- Tool name: `full_text_search`
- Input schema:
```json
{
  "type": "object",
  "properties": {
    "indexName": {"type": "string"},
    "queryText": {"type": "string"},
    "limit": {"type": "integer", "default": 10}
  },
  "required": ["database", "indexName", "queryText"]
}
```
- Output: `{results: [{rid, bm25Score, properties}]}`
- Permission gating: `allowReads`.
- Engine functions used: `SQLFunctionSearchIndex` (`fulltext.searchIndex`), `LSMTreeFullTextIndex`, `BM25Scorer`.

## Acceptance criteria

- [ ] Correct BM25 score surfaced (not just rank order).
- [ ] Errors clearly when `indexName` is not a full-text index type.
- [ ] Registered in both transports; tests + docs.

## Files to touch

`server/src/main/java/com/arcadedb/server/mcp/tools/FullTextSearchTool.java` (new), `MCPHttpHandler.java`, `MCPStdioServer.java`, tests, docs.

## Dependencies

None required by others, but should land before/alongside #4861 for terminology consistency.

## Effort estimate

S — thin wrapper over an existing, well-defined function.

robfrank:
Closes #4862. Part of epic #4859 (MCP GraphRAG & Agent-Memory Surface), Wave 1.

## What this adds

An MCP `full_text_search` tool, the lexical leg of the epic's retrieval trio (`vector_search` #4860, `hybrid_search` #4861, this).

Per the epic's governing principle, a new tool needs a justification beyond "`query` can already do it". This one is **#1, ergonomic guardrail**: the raw form is

```sql
SELECT title, $score FROM Article WHERE SEARCH_INDEX('Article[content]', 'java') = true ORDER BY $score DESC
```

and a model reliably gets three things wrong there: the index name is a bracket literal auto-derived as `typeName + Arrays.toString(props)`; the BM25 relevance is only reachable as the context variable `$score`, which must be explicitly projected; and a miss throws a bare `SchemaException` with no recovery hint.

**Tool contract**

- Input: `database`, `queryText` (required); address the index by either `indexName` **or** `typeName` (+ optional `properties`); `limit` (default 10, `< 1` rejected).
- Output: `{indexName, similarity, count, results: [{rid, score, properties}]}`, sorted by score descending.
- Gated on `allowReads` -> `SecurityException`, exactly as `QueryTool`. Pure-read, so no `OperationType` write-gating. No new config flag; safe-by-default (`enabled=false`) is unchanged.
- Registered in both transports. Tool count 10 -> 11.

**Structure.** The search logic was extracted from the previously-private `SQLFunctionSearchIndex.performSearch` into a new engine helper `com.arcadedb.index.fulltext.FullTextSearch`; `performSearch` is now a delegate. `#4861 hybrid_search` gets the same entry point rather than a second copy of the traversal. The SQL path is unchanged, and **no pre-existing test was edited anywhere on this branch** - that is the evidence.

Deliberately **not** synthesizing SQL: `queryText` legitimately contains quotes (phrase queries, apostrophes), so it would have to be a bound parameter, but `SEARCH_INDEX` is an `IndexableSQLFunction` the planner may consume at plan time and **no test in the repo binds its arguments as parameters**. Staking the tool on an unproven path whose failure mode is SQL injection wasn't warranted when the programmatic path is exact.

## Two deliberate deviations from the issue text

**1. `bm25Score` -> `score`, plus a top-level `similarity` field.** An ArcadeDB full-text index is not necessarily BM25: indexes persisted before BM25 support load as `CLASSIC`, and `METADATA {"similarity": "CLASSIC"}` pins it explicitly. On those, the value is a term-coordination count. Naming the field `bm25Score` would mislabel them, and would silently corrupt score fusion in #4861. `similarity` is reported from `LSMTreeFullTextIndex.isBM25()`.

**2. Index addressing also accepts `typeName` + `properties`.** Requiring `indexName` alone would have reintroduced the exact footgun the tool exists to remove. Addressing errors now enumerate the database's real full-text index names, mirroring the `MCPToolUtils.resolveDatabase` house pattern, so the model self-corrects in one retry. The argument is `typeName`, not `type`, to avoid colliding with JSON Schema's own `type` keyword.

## Review findings worth other eyes

**A live bug, fixed here.** The first cut aborted the *entire* search if any one hit had been deleted between the index scan and the record dereference, discarding every other valid result. The tool runs outside an explicit transaction, so those are separate read windows. It now skips the dangling hit, matching what index scans already do (`LocalDatabase` documents this as the expected pattern). No regression test: deterministically producing a dangling posting needs fault injection, and `SQLFunctionSearchIndex.searchFromTarget` ships the identical untested skip today.

**Top-k is now pushed down.** The engine already had a bounded min-heap top-k in `LSMTreeFullTextIndex.buildScoredCursor` that only engages when a limit is supplied; we were passing `-1` and then re-sorting the full match set in the tool to return 10 rows. The limit is now pushed into each bucket cursor. Soundness: scores are limit-independent and the (score desc, RID asc) order is strict and applied identically per-bucket and on the merge, so the global top-k is contained in the union of the per-bucket top-ks. Guarded by a `BUCKETS 4` test that asserts (via `RID.getBucketId()`) that the strong hits really do concentrate in one bucket, then that the bounded top-k RID sequence equals the unbounded one. SQL keeps passing `-1` (unbounded) - it uses the result as a boolean predicate over every candidate row, so narrowing it would be a silent regression.

## :warning: Pre-existing engine issue this surfaced - needs a decision

**ArcadeDB scores BM25 per bucket.** Document frequency is bucket-scoped while average document length is type-wide; the engine documents this itself:

> `"note": "statistics are per bucket (BM25 is scored per bucket); totalDocs is this bucket's document count"`
> `LSMTreeFullTextIndex.java:547`

So merging hits across buckets and ranking them globally is **not strictly sound on the default multi-bucket layout**. We hit it live during development: a *less* relevant document outranked a *more* relevant one, and the ranking test only became deterministic once the fixture forced `BUCKETS 1`.

To be clear about scope: **this PR does not introduce it.** SQL's existing `ORDER BY $score DESC` has had the identical exposure all along, and the top-k push-down above neither creates nor worsens it (scores are unchanged). The reviewers were explicit that it is not a blocker for this PR.

But it should not stay implicit, because:

- this tool's headline contract is "records ranked by relevance", and that is only *strictly* true on single-bucket types;
- **#4861 `hybrid_search` will fuse these scores with vector scores**, which would bake the unsoundness into a second surface where it is far harder to unwind.

Suggested resolution, for discussion: file a separate engine issue (options include type-wide corpus statistics, or a global re-scoring pass over the merged candidate set) and link it as a **blocker on #4861** before that work starts. Opinions welcome on which fix shape is right - that is why it is called out here rather than buried in a commit message.

## Testing

- `engine`: 201/201 green (`FullTextSearchTest` + the full `SEARCH_INDEX` / `FullText*` / `LSMTreeFullTextIndex*` / `TypeFullTextIndexBuilderTest` suites, which are the regression net for the extraction).
- `server`: 79/79 green (`MCP*Test`), covering the happy path, ranking, limit rejection, both addressing forms, `indexName`-wins precedence, and the error paths (unknown index, non-full-text index, ambiguous `typeName`, subtype `typeName`, no addressing form).
- No pre-existing test file was edited.

## Merge-conflict note for the rest of Wave 1

`MCPHttpHandler.java` and `MCPStdioServer.java` are `+4/-0` and `+3/-0`, purely additive, no reordering. #4860 / #4863 / #4864 / #4865 touch the same `TOOLS_LIST` blocks, dispatch switches, and the `tools.length()` assertion in `MCPServerPluginTest` (now 11), so expect adjacent-line conflicts there and nowhere else.

## Follow-ups

- Docs live in the out-of-tree `arcadedb-docs` repo: tool reference, query-syntax table, the `similarity` field, and a cold-start note (a BM25 index with invalid counters runs a full type scan on its first query; remedy is `REBUILD INDEX <name> WITH statsOnly = true`). Docs should say "BM25 full-text index", never "Lucene index".
- The cross-bucket BM25 engine issue above.

### DIFF
--- a/docs/superpowers/plans/2026-07-13-mcp-full-text-search-4862.md
+++ b/docs/superpowers/plans/2026-07-13-mcp-full-text-search-4862.md
@@ -1,0 +1,894 @@
+# MCP `full_text_search` Tool Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `full_text_search` MCP tool that lets an agent run BM25 full-text retrieval against ArcadeDB without hand-writing the `SEARCH_INDEX('Type[prop]', ...)` / `$score` SQL syntax.
+
+**Architecture:** Two layers. A new stateless engine helper `com.arcadedb.index.fulltext.FullTextSearch` holds search logic extracted from the currently-private `SQLFunctionSearchIndex.performSearch`, preserving its exception types so the existing full-text test suites act as an unedited regression net. A new server-side `FullTextSearchTool` (shaped like `QueryTool`) pre-validates the index, calls the helper, and shapes `{indexName, similarity, count, results:[{rid, score, properties}]}`.
+
+**Tech Stack:** Java 21, Maven, JUnit 5 + AssertJ. Engine module (`engine/`) and server module (`server/`). No new dependencies.
+
+**Spec:** `docs/superpowers/specs/2026-07-13-mcp-full-text-search-4862-design.md`
+**Issue:** [#4862](https://github.com/ArcadeData/arcadedb/issues/4862), epic [#4859](https://github.com/ArcadeData/arcadedb/issues/4859) Wave 1
+
+## Global Constraints
+
+- Java 21+. Use `final` on variables and parameters. Import classes; do not use fully-qualified names inline.
+- Single-statement `if` bodies take no curly braces (match surrounding code).
+- JSON: use `com.arcadedb.serializer.json.JSONObject` / `JSONArray`, and prefer the two-argument getters (`getString(key, default)`) over presence checks.
+- Tests: JUnit 5 + AssertJ, in the style `assertThat(x.isFoo()).isTrue();`.
+- Every source file carries the existing Apache-2.0 license header (copy verbatim from a neighboring file in the same package).
+- Do not add Claude as an author. Do not reference issue numbers in Javadoc or comments; comments state behavioral invariants only.
+- **Commit per task** on the feature branch `feat/4862-mcp-full-text-search` (worktree at `.worktrees/feat/4862-mcp-full-text-search`). Never commit to `main`, and never merge without the operator's review. CLAUDE.md's "do not commit" governs `main` and merges, not per-task commits on an isolated branch.
+- Terminology in all prose: "BM25 full-text index", never "Lucene index". Lucene is used only for tokenization/analysis and query parsing.
+- The full-text index name is auto-derived as `typeName + Arrays.toString(propertyNames).replace(" ", "")`, i.e. `Article[content]`, `Article[title,body]`.
+- **Merge-conflict hotspot:** `MCPHttpHandler.java` and `MCPStdioServer.java` (`TOOLS_LIST` static blocks + dispatch switches) and the tool-count assertion in `MCPServerPluginTest` are touched by every Wave-1 issue (#4860, #4863, #4864, #4865). Keep those diffs minimal.
+
+---
+
+## File Structure
+
+| File | Responsibility |
+|---|---|
+| `engine/src/main/java/com/arcadedb/index/fulltext/FullTextSearch.java` (create) | Stateless helper: resolve a full-text `TypeIndex`, run the search, report similarity, enumerate full-text indexes. |
+| `engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchIndex.java` (modify) | `performSearch` reduced to a delegate. Behavior unchanged. |
+| `engine/src/test/java/com/arcadedb/index/fulltext/FullTextSearchTest.java` (create) | Unit tests for the helper. |
+| `server/src/main/java/com/arcadedb/server/mcp/tools/FullTextSearchTool.java` (create) | MCP tool: definition, index addressing, permission gate, output shaping. |
+| `server/src/main/java/com/arcadedb/server/mcp/MCPHttpHandler.java` (modify) | Register tool in HTTP transport. |
+| `server/src/main/java/com/arcadedb/server/mcp/MCPStdioServer.java` (modify) | Register tool in stdio transport. |
+| `server/src/test/java/com/arcadedb/server/mcp/MCPServerPluginTest.java` (modify) | End-to-end HTTP coverage + tool count 10 -> 11. |
+| `server/src/test/java/com/arcadedb/server/mcp/MCPStdioServerTest.java` (modify) | Tool present in stdio `tools/list`. |
+
+---
+
+## Task 1: Extract the `FullTextSearch` engine helper
+
+Pure refactor plus new public surface. The SQL path must come out byte-identical, which the existing suites prove.
+
+**Files:**
+- Create: `engine/src/main/java/com/arcadedb/index/fulltext/FullTextSearch.java`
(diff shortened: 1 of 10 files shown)


robfrank:
Addressed in 0ad8626f4. Two of the three observations were real; I verified each against the code rather than assuming, and pushed back on the third.

**1. Blank `queryText` - confirmed and fixed.** I probed it rather than reason about it. An empty or whitespace-only query does not return empty results; it throws:

```
IndexException: Invalid search query:
```

An empty string in the message, naming no cause. For an LLM caller that is close to useless. Now rejected up front with a message that says which argument is wrong and shows a valid form, mirroring the `limit < 1` guard. Test `fullTextSearchRejectsBlankQueryText` covers `""` and `"   "`, and asserts the opaque `Invalid search query` text is *not* what comes back.

**3. Property-name normalization - confirmed, and more reachable than flagged.** This is not only theoretical. A property name containing a space **is** creatable:

```sql
CREATE PROPERTY Spaced.`my prop` STRING
CREATE INDEX ON Spaced (`my prop`) FULL_TEXT
```

The schema derives the index name as `typeName + Arrays.toString(props)` with spaces stripped **from the whole result** (`LocalDocumentType.java:1126`), so that registers as `Spaced[myprop]`. The tool was joining property names verbatim and deriving `Spaced[my prop]`, which can never match - the caller got "does not exist" for an index that plainly exists. Fixed by stripping spaces from the derived name, exactly as the schema does.

Verified the fix is load-bearing rather than trusting it: with the `.replace(" ", "")` reverted, the new test `fullTextSearchDerivesIndexNameWithSpacesStripped` fails with

```
Full-text index 'Spaced[my prop]' does not exist.
Available full-text indexes in 'graph': [..., Spaced[myprop]]
```

which is precisely the bug.

**2. `getSimilarity` third bucket-array walk - not changing this, and I want the reasoning on the record.** It is one array allocation per *query*, not per row (the redundant *name lookup* was already removed in c90c20053). Eliminating it would mean reshaping a public engine signature so the search path returns the similarity alongside the result map. Restructuring public engine API to save one short-lived array per query is a bad trade, and it would widen this PR's engine surface for no measurable gain. Happy to be overruled if you disagree, but I would rather not churn the API for it.

**Cross-bucket BM25 issue:** agreed on the sequencing, and I will open the engine issue and link it as a blocker on #4861 so it does not stay implicit.

Tests after the fixes: engine **201/201**, server **81/81** (was 79; +2). No pre-existing test edited.

robfrank:
Engine issue filed as #5267 ("Full-text BM25 is scored per bucket, so cross-bucket relevance ranking is unsound"), and linked as a blocker on #4861 so the fusion strategy there is not settled before this is.

It lays out three directions - type-wide corpus statistics, a global re-scoring pass over the merged candidate set, or an explicit best-effort contract - without picking one. Input welcome there rather than here, so this PR stays about #4862.

PR #5264, opened 2026-07-13
fix(#5262): root test databases under target/ via TestHelper

### LINKED ISSUE #5262: Test hygiene: 55 tests hand-roll their database lifecycle on repo-relative ./databases/ paths instead of using TestHelper
## Summary

55 test classes across `engine`, `server`, `integration` and `gremlin` construct their own `DatabaseFactory` on a **repo-relative** `./databases/<name>` path and hand-roll the create/drop lifecycle, instead of extending [`TestHelper`](engine/src/test/java/com/arcadedb/TestHelper.java).

This has two concrete consequences:

1. **A leftover directory poisons every subsequent run of that test**, because `LocalDatabase.create()` throws when a schema file is already present ([`LocalDatabase.java:283-287`](engine/src/main/java/com/arcadedb/database/LocalDatabase.java#L283-L287)).
2. **`mvn clean` cannot recover from it**, because the databases are written *outside* `target/`. The only cure is a manual `rm -rf`, which a developer has to know to do.

Spotted during review of #5257 / PR #5260, where the new test had the same defect and was refactored onto `TestHelper` (commit `8879cbd7`).

## Reproduction

Simulate what a `kill -9`, a surefire fork timeout, or an OOM leaves behind:

```bash
cd engine
mkdir -p databases/test-issue5213 && echo '{}' > databases/test-issue5213/schema.json
mvn test -Dtest=Issue5213NestedCallScopeTest
```

Result - all 5 tests error out, and no amount of `mvn clean` fixes it:

```
[ERROR] Tests run: 5, Failures: 0, Errors: 5
com.arcadedb.exception.DatabaseOperationException: Database './databases/test-issue5213' already exists
com.arcadedb.exception.DatabaseOperationException: Database './databases/test-issue5213' already exists
...
```

Note this is not hypothetical fallout from a crash only: because `@AfterEach` drops the database but nothing cleans up *before* the test, **any** path that skips the teardown (JVM crash, `kill -9`, fork timeout, OOM, a failure inside `drop()` itself) permanently wedges the class until someone manually deletes the directory.

## Why `TestHelper` fixes it

`TestHelper` already does all four things these tests need, and they only do one of them:

| | hand-rolled tests | `TestHelper` |
|---|---|---|
| clean **before** test | ✗ | ✓ `FileUtils.deleteRecursively` in ctor |
| drop **after** test | ✓ | ✓ `afterTest()` |
| path under `target/` (so `mvn clean` works) | ✗ `./databases/x` | ✓ `target/databases/<ClassName>` |
| assert no `Database` instance leaked | ✗ | ✓ `checkActiveDatabases()` |
| database integrity check | ✗ | ✓ `checkDatabaseIntegrity()` |

## Scope

55 classes. Grouped by how much work each needs:

### Group B - has `drop()`, no pre-test cleanup (33 classes)

The straightforward ones: delete the `@BeforeEach`/`@AfterEach` pair and the `Database` field, `extends TestHelper`, move seeding into the `beginTest()` hook. This is exactly the shape of the PR #5260 refactor (net -12 lines there).

<details>
<summary>File list (33)</summary>

```
engine/src/test/java/com/arcadedb/engine/RandomDeleteTest.java
engine/src/test/java/com/arcadedb/engine/RecordRecyclingTest.java
engine/src/test/java/com/arcadedb/engine/TestInsertAndSelectWithThreadBucketSelectionStrategy.java
engine/src/test/java/com/arcadedb/query/opencypher/CypherFunctionSecurityTest.java
engine/src/test/java/com/arcadedb/query/opencypher/CypherReduceAndShortestPathTest.java
engine/src/test/java/com/arcadedb/query/opencypher/CypherShortestPathEdgeFilterTest.java
engine/src/test/java/com/arcadedb/query/opencypher/HeadCollectTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5107MatchCreateIndexTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5136MatchMultipleCreateOptimizerTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5140CountInPatternComprehensionWhereTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5145AllPatternComprehensionTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5163DivisionByZeroTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5164IntegerOverflowTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5179PatternComprehensionScopeTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5180TrailingCommaListLiteralTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5213NestedCallScopeTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5226CallMatchEdgeLabelTest.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5228CallMatchEdgeLabelEntityTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherCustomFunctionTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherExpressionTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherFunctionTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherGroupByTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherListPredicateTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherNewFunctionsTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherOptimizerVerificationTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherPatternComprehension5146Test.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherPatternComprehensionIssue5111Test.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherPatternComprehensionRelPropertyTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherPatternPredicateTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherStDevTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherTimestampTest.java
engine/src/test/java/com/arcadedb/query/sql/OrderByTest.java
server/src/test/java/com/arcadedb/server/RemoteSafeCloseDatabaseIT.java
```
</details>

### Group C - already self-heals, but still writes outside `target/` (20 classes)

These call `factory.exists()` / `deleteRecursively` before creating, so they do not wedge. They still pollute the repo working tree with a `<module>/databases/` directory that `mvn clean` never removes. Lower priority, same destination.

<details>
<summary>File list (20)</summary>

```
engine/src/test/java/com/arcadedb/query/opencypher/Issue4305Test.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue4353Test.java
engine/src/test/java/com/arcadedb/query/opencypher/Issue5207Test.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherAdvancedFunctionTest.java
engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherCollectUnwindTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/Issue4577And4578Test.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherAggregatingFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherListFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherLoadCsvFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherMathLogarithmicFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherMathNumericFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherMathTrigonometricFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherPredicateFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherScalarFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherSpatialFunctionsComprehensiveTest.java
engine/src/test/java/com/arcadedb/query/opencypher/functions/OpenCypherStringFunctionsComprehensiveTest.java
gremlin/src/test/java/performance/CypherEngineComparisonBenchmark.java
integration/src/test/java/com/arcadedb/integration/importer/vector/FastTextDatabase.java
integration/src/test/java/com/arcadedb/integration/importer/vector/GloVeTest.java
integration/src/test/java/com/arcadedb/integration/importer/vector/SimpleGloVeTest.java
server/src/test/java/com/arcadedb/server/AsyncInsertTest.java
server/src/test/java/com/arcadedb/server/RemoteQueriesIT.java
```
</details>

### Group A - never drop the database at all (2 classes)

Leak both the directory and, potentially, the open `Database` instance.

```
integration/src/test/java/com/arcadedb/integration/exporter/SQLLocalExporterTest.java
integration/src/test/java/com/arcadedb/integration/importer/SQLLocalImporterIT.java
```

## Proposed fix

For the common case, the whole migration is:

```java
class FooTest extends TestHelper {
  @Override
  protected void beginTest() {
    database.command("opencypher", "CREATE (:A {id: 1})");   // seeding that used to live in @BeforeEach
  }
  // ... @Test methods unchanged, using the inherited `database` field
}
```

Caveats for whoever picks this up:

- `TestHelper` runs `checkDatabaseIntegrity()` (a `CHECK DATABASE`) in teardown by default. A test that *intentionally* leaves the database inconsistent must override `isCheckingDatabaseIntegrity()` to return `false`.
- `TestHelper` also asserts no `Database` instance was left active; a test that deliberately leaks one will now fail loudly (which is the point, but it may surface pre-existing leaks).
- The server/`IT` classes build on `BaseGraphServerTest` and friends, so they need checking case by case rather than a blind sweep.
- A few classes need a non-default `getDatabasePath()` or `getPerformanceProfile()` override.

Groups B and A are worth doing first; Group C is cosmetic-but-tidy and can follow.

### Guard against regressions

Once migrated, consider a build-time guard so this does not creep back - e.g. an `enforcer`/`checkstyle` rule (or a plain `grep` in CI) banning `new DatabaseFactory("./databases/` and `new DatabaseFactory("databases/` under `src/test/java`.

## Verification

Per group, `mvn -pl <module> test` must stay green, and after the run neither `engine/databases/`, `server/databases/`, `integration/databases/` nor `gremlin/databases/` should exist.

robfrank:
Closes #5262

## Summary

57 test classes across `engine`, `server`, `integration` and `gremlin` constructed their own `DatabaseFactory` on a repo-relative `./databases/<name>` path and hand-rolled the create/drop lifecycle. Any run that skipped teardown (`kill -9`, a surefire fork timeout, an OOM, or a failure inside `drop()` itself) left the directory behind, which made every subsequent run of that class fail with `Database './databases/x' already exists`. Because the databases lived outside `target/`, `mvn clean` could not recover - the only cure was a manual `rm -rf` that a developer had to know to do.

The 48 `engine` classes now extend `TestHelper`, which already does all four things these tests need: cleans before the test, drops after, roots the database at `target/databases/<ClassName>`, asserts no `Database` instance leaked, and runs a `CHECK DATABASE` integrity check. Seeding moves from `@BeforeEach` into the `beginTest()` hook.

The 9 classes outside `engine` cannot extend `TestHelper` - the `server` and `gremlin` test modules do not depend on the engine test-jar, and three of the `integration` classes are `main()` utilities rather than JUnit tests - so they are repointed under `target/` directly. The three server tests set `SERVER_ROOT_PATH` to `./target`, the convention already used by the other server tests; since `SERVER_DATABASE_DIRECTORY` defaults to `${arcadedb.server.rootPath}/databases`, that one change relocates both the pre-created database and the server's own database directory.

Net **-698 lines**.

## Scope grew during review

Review found the first version of the guard was too narrow, and closing that hole exposed more of the same bug. This PR therefore goes past the 55 classes the issue lists:

- **6 more engine/integration classes** put databases on repo-relative paths via a `String` constant rather than an inline literal, so the original guard reported OK while they still leaked: `LSMVectorIndexPersistenceTest`, `LSMVectorIndexChunkedWriteTest`, `ContiguousPageIOTest`, three `LSMVectorIndexTest` methods, and `GloVeReopenVerificationTest` (which used `"../databases/glovedb"` - the repo *root*, a level above the module).
- **5 more server tests** leak through a path that is *derived*, not written: `SelectOrderTest`, `CompositeIndexTest`, `BatchInsertUpdateTest`, `RemoteDateIT`, `RemoteCollectionTemporalIT` call `IntegrationUtils.setRootPath(...)` without setting `SERVER_ROOT_PATH` first. `ServerPathUtils.setRootPath` only honours an already-set value and otherwise returns `"."`, so `rootPath + "/databases/"` resolved to `server/databases/`. Confirmed empirically: a full `mvn -pl server test` on the pre-fix tree left `server/databases/SelectOrderTest` on disk. No grep can catch this class of leak.
- **`OpenCypherCollectUnwindTest`**: four `@Nested` classes created their database with a bare `new DatabaseFactory(...).create()` and no pre-clean. Now use `TestHelper.createDatabase(...)` (drop-if-exists, then create).
- **`ConsoleTest`**: two `deleteRecursively(new File("databases/" + ...))` calls were dead cleanup of a legacy location the console no longer writes to. Removed.

The issue's own verification criterion - no module has a `databases/` directory after a run - is not reachable without these.

## Regression guard

`.github/scripts/check-test-database-paths.sh`, two layers:

- **static** (`setup` job): greps test sources for a literal rooted at a repo-relative `databases/` dir. Matching the *literal* rather than the `new DatabaseFactory(...)` call site is what catches the `String DB_PATH = "databases/..."` form. Deliberately does not match `"/databases/`, the suffix half of the correct `rootPath + "/databases/" + name` idiom.
- **runtime** (after **all four** test jobs - `unit-tests`, `slow-unit-tests`, `integration-tests`, `ha-integration-tests`): asserts no module grew a top-level `databases/` directory. This is the backstop for the `SERVER_ROOT_PATH`-derived case that no grep can resolve, and the server ITs that exhibit that pattern run in `integration-tests`.

## What changed beyond lifecycle

No assertion, query or expected value was changed. Two edits are **not** pure moves:

1. `RecordRecyclingTest`: `databaseFactory.getActiveDatabaseInstances()` -> `DatabaseFactory.getActiveDatabaseInstances()`. Same static method, called through the class because the local variable was removed.
2. `OrderByTest` and `TestInsertAndSelectWithThreadBucketSelectionStrategy`: `beginTest()` now also calls `database.getSchema().setDateTimeFormat(...)`. Required, not cosmetic: `TestHelper` builds the database in its **constructor**, before any subclass code runs, and `LocalSchema` captures `DATE_TIME_FORMAT` at construction and persists it into `schema.json`. Setting only the global from `beginTest()` arrives too late for the already-built schema, so the format must be written onto the schema to survive the `reopenDatabase()` these tests perform.

## Test plan

- [x] Guard fails on the pre-fix tree listing exactly the offending files, and passes after
- [x] Runtime guard verified both directions: a planted reactor-root `./target/databases` control is correctly ignored; a real `<module>/databases` is caught
- [x] All 48 migrated `engine` classes green: **966 tests, 0 failures** - including every class where `TestHelper`'s newly-applied `CHECK DATABASE` teardown could have surprised
- [x] Newly-scoped classes green: vector-index (33), `OpenCypherCollectUnwindTest` (77), `ConsoleTest` (30), `SelectOrderTest`/`CompositeIndexTest`/`BatchInsertUpdateTest`, `GloVeReopenVerificationTest`
- [x] Affected ITs green: `RemoteQueriesIT`, `RemoteSafeCloseDatabaseIT`, `SQLLocalImporterIT`, `RemoteDateIT`, `RemoteCollectionTemporalIT`
- [x] `SelectOrderTest`'s database verified to land in `server/target/databases/` after the fix
- [x] `engine`, `console`, `server`, `integration`, `gremlin` all compile
- [x] **Bug reproduction:** planting a leftover `schema.json` (simulating a crashed run) no longer wedges the class - 5/5 pass, where the pre-fix equivalent errored on all 5 and no `mvn clean` could fix it

### DIFF
--- a/.github/scripts/check-test-database-paths.sh
+++ b/.github/scripts/check-test-database-paths.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Fails when a test constructs a DatabaseFactory on a repo-relative "databases/" path.
+# Fails when a test would put its database on a repo-relative "databases/" path.
 #
 # Such a database is written outside target/, so `mvn clean` cannot remove it. If a run dies
 # before its teardown (kill -9, surefire fork timeout, OOM), the leftover directory makes every
@@ -8,25 +8,78 @@
 # `rm -rf` recovers. Tests must place their database under target/ - extend TestHelper, which
 # does that plus pre-test cleanup, or pass an explicit target/-rooted path.
 #
+# Two complementary checks:
+#
+#   static  - grep test sources for repo-relative "databases/" string literals. Catches the path
+#             whether it is inlined in the constructor or held in a String constant/local.
+#
+#   runtime - after a test run, assert no module has grown a top-level databases/ directory.
+#             This is the backstop: a server test derives its path from SERVER_ROOT_PATH rather
+#             than a literal, so no grep can see it - only the resulting directory can.
+#
+# Usage:
+#   check-test-database-paths.sh            # static only (cheap, safe to run pre-build)
+#   check-test-database-paths.sh --runtime  # runtime only (run AFTER the test phase)
+#   check-test-database-paths.sh --all      # both
+#
 set -euo pipefail
 
 cd "$(dirname "$0")/../.."
 
-# Matches: new DatabaseFactory("databases/...  and  new DatabaseFactory("./databases/...
-PATTERN='new[[:space:]]+DatabaseFactory\([[:space:]]*"\.?/?databases/'
+mode="${1:---static}"
+status=0
 
-violations=$(grep -rEln --include='*.java' "${PATTERN}" -- */src/test/java || true)
+# A string literal that *starts* at a repo-relative databases/ dir: "databases/, "./databases/, "../databases/
+#
+# Deliberately does NOT match "/databases/, which is the suffix half of the correct server idiom
+# `new DatabaseFactory(rootPath + "/databases/" + name)`. Whether that one lands outside target/
+# depends on SERVER_ROOT_PATH, which no grep can resolve - the runtime check below covers it.
+LITERAL='"(\.\.?/)?databases/'
 
-if [ -n "${violations}" ]; then
-  echo "ERROR: test sources construct a DatabaseFactory on a repo-relative 'databases/' path."
-  echo "These databases live outside target/, so 'mvn clean' cannot remove them and a leftover"
-  echo "directory from a crashed run permanently wedges the test class."
-  echo
-  echo "Fix: extend com.arcadedb.TestHelper, or root the path under target/databases/."
-  echo
-  echo "Offending files:"
-  echo "${violations}" | sed 's/^/  - /'
-  exit 1
-fi
+run_static() {
+  local violations
+  violations=$(grep -rEl --include='*.java' "${LITERAL}" -- */src/test/java || true)
 
-echo "OK: no test constructs a DatabaseFactory on a repo-relative 'databases/' path."
+  if [ -n "${violations}" ]; then
+    echo "ERROR: test sources reference a repo-relative 'databases/' path."
+    echo "These databases live outside target/, so 'mvn clean' cannot remove them and a leftover"
+    echo "directory from a crashed run permanently wedges the test class."
+    echo
+    echo "Fix: extend com.arcadedb.TestHelper, or root the path under target/databases/."
+    echo
+    echo "Offending files:"
+    grep -rEn --include='*.java' "${LITERAL}" -- */src/test/java | sed 's/^/  /'
+    status=1
+  else
+    echo "OK (static): no test source references a repo-relative 'databases/' path."
+  fi
+}
+
+run_runtime() {
+  local leaked
+  leaked=$(find . -maxdepth 2 -type d -name databases -not -path './.git/*' || true)
+
+  if [ -n "${leaked}" ]; then
+    echo "ERROR: the test run left a repo-relative 'databases/' directory behind."
+    echo "The database was written outside target/, so 'mvn clean' cannot remove it."
+    echo
+    echo "A server test most likely started ArcadeDBServer without setting SERVER_ROOT_PATH, which"
+    echo "defaults the database directory to ./databases. Set it to ./target, as the other server"
+    echo "tests do:  GlobalConfiguration.SERVER_ROOT_PATH.setValue(\"./target\");"
+    echo
+    echo "Leaked directories:"
+    echo "${leaked}" | sed 's/^/  - /'
+    status=1
+  else
+    echo "OK (runtime): the test run left no repo-relative 'databases/' directory behind."
+  fi
+}
+
+case "${mode}" in
+  --static)  run_static ;;
+  --runtime) run_runtime ;;
+  --all)     run_static; run_runtime ;;
+  *) echo "usage: $0 [--static|--runtime|--all]" >&2; exit 2 ;;
+esac
+
+exit "${status}"
--- a/.github/scripts/check-test-database-paths.sh
+++ b/.github/scripts/check-test-database-paths.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+#
+# Fails when a test constructs a DatabaseFactory on a repo-relative "databases/" path.
+#
+# Such a database is written outside target/, so `mvn clean` cannot remove it. If a run dies
+# before its teardown (kill -9, surefire fork timeout, OOM), the leftover directory makes every
+# subsequent run of that class fail with "Database ... already exists", and only a manual
+# `rm -rf` recovers. Tests must place their database under target/ - extend TestHelper, which
+# does that plus pre-test cleanup, or pass an explicit target/-rooted path.
+#
+set -euo pipefail
+
+cd "$(dirname "$0")/../.."
+
+# Matches: new DatabaseFactory("databases/...  and  new DatabaseFactory("./databases/...
+PATTERN='new[[:space:]]+DatabaseFactory\([[:space:]]*"\.?/?databases/'
+
+violations=$(grep -rEln --include='*.java' "${PATTERN}" -- */src/test/java || true)
+
+if [ -n "${violations}" ]; then
+  echo "ERROR: test sources construct a DatabaseFactory on a repo-relative 'databases/' path."
+  echo "These databases live outside target/, so 'mvn clean' cannot remove them and a leftover"
+  echo "directory from a crashed run permanently wedges the test class."
+  echo
+  echo "Fix: extend com.arcadedb.TestHelper, or root the path under target/databases/."
+  echo
+  echo "Offending files:"
+  echo "${violations}" | sed 's/^/  - /'
+  exit 1
+fi
+
+echo "OK: no test constructs a DatabaseFactory on a repo-relative 'databases/' path."
--- a/.github/scripts/check-test-database-paths.sh
+++ b/.github/scripts/check-test-database-paths.sh
@@ -12,6 +12,9 @@
 #
 #   static  - grep test sources for repo-relative "databases/" string literals. Catches the path
 #             whether it is inlined in the constructor or held in a String constant/local.
+#             It matches the literal anywhere in the file, so a "databases/..." string inside a
+#             comment or a log message trips it too. That is deliberate: a false alarm you can
+#             silence by rewording beats a leak that wedges someone's test class next month.
 #
 #   runtime - after a test run, assert no module has grown a top-level databases/ directory.
 #             This is the backstop: a server test derives its path from SERVER_ROOT_PATH rather
@@ -57,7 +60,8 @@ run_static() {
 
 run_runtime() {
   local leaked
-  leaked=$(find . -maxdepth 2 -type d -name databases -not -path './.git/*' || true)
+  # <module>/databases is a leak; anything under a target/ dir is the goal state, not a leak.
+  leaked=$(find . -maxdepth 2 -type d -name databases -not -path './.git/*' -not -path '*/target/*' || true)
 
   if [ -n "${leaked}" ]; then
     echo "ERROR: the test run left a repo-relative 'databases/' directory behind."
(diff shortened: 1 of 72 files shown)


robfrank:
Thanks - took two of the three, and I want to explain the one I didn't.

**Taken: runtime guard scoping (2).** Right catch. `find -maxdepth 2 -name databases` would also match a reactor-root `./target/databases`, which is the goal state, not a leak. Now excludes `*/target/*`. Verified both directions against a planted `./target/databases` control: it is ignored, while a real `<module>/databases` is still caught.

**Taken: static grep bias note (3).** Documented in the script header - the literal is matched anywhere in the file, comments and log strings included, and the false-alarm bias is deliberate.

**Not taken: reusing `reopenDatabase()` in `RecordRecyclingTest.createAndDeleteGraph()` (1).** This one isn't DRY-equivalent. `TestHelper.reopenDatabase()` closes and reopens with no window in between:

```java
protected void reopenDatabase() {
  if (database != null) {
    database.close();
    assertThat(...).isNull();
  }
  database = factory.open();   // <- reopens immediately
  ...
}
```

but `createAndDeleteGraph()` has to delete the statistics file *between* the close and the open:

```java
} finally {
  db.close();
  new File(factory.getDatabasePath() + File.separator + STATISTICS_FILE_NAME).delete();  // <- needs the gap
  assertThat(DatabaseFactory.getActiveDatabaseInstances()).isEmpty();
  database = factory.open();
}
```

That gap is the point of the test - it exercises recovery when the statistics file is missing. Calling `reopenDatabase()` would reopen the database *before* the delete, so the delete would land on an open database and the reopen would never exercise the missing-stats path. The `OrientDBTest`-style `reopenDatabase()` call in `OrderByTest` is a different case (plain close/reopen, no work in between), which is why that one does use the helper.

Happy to be talked out of it if I've misread the intent of that `finally` block.

Issue #5214, opened 2026-07-11
LSM: bounded range scans over multi-series compacted indexes drop rows and mis-order descending results


lvca:
## Summary

Bounded range scans over an LSM index that has compacted into **3+ series** silently drop rows, and the descending bounded iterator returns results in the wrong order. Data on disk is intact (a full scan returns everything); only the **bounded/range** read paths over the multi-series compacted layer are affected.

Multiple compacted series in a single file are produced when a compaction is RAM-bounded (`INDEX_COMPACTION_RAM_MB`), i.e. on large indexes whose immutable pages exceed the compaction budget. Small indexes compact into a single series, where these paths happen to work, which is why this has stayed latent.

This is **independent of #5211** (which fixed only the unbounded full descending scan).

## Reproduction

Non-unique LSM index over a `String` property, 20,000 keys, `INDEX_COMPACTION_RAM_MB=1` so one compaction emits several series:

```sql
-- correct
SELECT count(*) FROM Doc;
-- => 20000

-- BUG: range query drops ~2,500 rows
SELECT count(*) FROM Doc WHERE email >= 'K...00005000' AND email <= 'K...00015000';
-- => 7514   (expected 10001)
```

At the index API level:

| call | expected | actual |
|---|---|---|
| `range(true,  lo, hi)` (ascending window) | 10001 | **7514** |
| `range(false, hi, lo)` (descending window) | 10001, descending | **2489**, descending |
| `iterator(false, fromKeys, true)` | keys <= fromKeys, **descending** | correct set, **ascending order** |

## Root causes

**1. Interior series dropped from bounded scans** - `LSMTreeIndexCompacted.newIterators(ascending, fromKeys, toKeys)`

For a bounded scan each series only yields a cursor when `fromKeys` **or** `toKeys` falls *inside* that series (`searchInCurrentPage` returns `null` when `resultInRootPage.outside`). A series lying **entirely within** `[lo, hi]` - containing neither endpoint - fails both lookups, contributes no cursor, and all of its keys are lost. A wide range spanning many series drops every interior series.

**2. Descending bounded iterator returns ascending order** - `LSMTreeIndexMutable.iterator(false, fromKeys, inclusive)`

It delegates to the auto-detecting `range(null, true, fromKeys, inclusive)`. With the begin-key `null`, the direction auto-detect defaults to `ascending = true`, so a descending request iterates ascending. Masked today because existing tests assert only element **counts**, never order.

[...]

Issue #4943, opened 2026-07-04
[lsm] Descending compacted-series cursor starts one page too far: phantom RIDs and duplicates on DESC scans


lvca:
**Severity:** HIGH (wrong query results on ORDER BY ... DESC)

**Location:** `index/lsm/LSMTreeIndexCompacted.java:334`:
```java
int startingPageNumber = rootPageNumber + 1 + (ascendingOrder ? 0 : rootPageCount);
```
A series occupies `[root, root+1 .. root+rootPageCount]`. Ascending correctly starts at `root+1`. Descending should start at the last leaf `root+rootPageCount`, but starts at `root+rootPageCount+1`, i.e. the **root page of the next series** whenever the compacted file holds 2+ series (after the second compaction round). Root pages look like leaf pages, so the cursor iterates them: a full DESC scan emits the next series` root entries - keys duplicating real leaf min-keys with values `RID(0, pageNumber)` and the max-key marker `RID(0,0)`. These have `bucketId=0 >= 0`, so `LSMTreeIndexCursor` treats them as live inserts and returns bogus `#0:x` RIDs alongside real ones. Single-series tests never hit it (the off-by-one lands on a fabricated empty page). No test exercises DESC iteration.

**Fix:** `int startingPageNumber = rootPageNumber + (ascendingOrder ? 1 : rootPageCount);`. Regression test: compact twice, then `iterator(false)` and `ORDER BY ... DESC` assert no `#0:x` RIDs and no duplicate keys.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Closing as not reproducible on current main, verified two ways in the #4968 work:

1. Code trace: compacted series iterators are created newest-first (`LSMTreeIndexCompacted.newIterators` walks `rootPageNumber` from the last series backwards), matching the temporal-order assumption of the cursor's merge, and the DESC init path seeks with purpose=3 correctly.
2. Empirically: a descending scan over a 400-key, 2-series compacted index (built with compactions that are now ASSERTED to actually run - see the harness note in #4942) returns exactly the expected keys in order, with no phantom `#0:x` RIDs and no duplicates.

[...]

PR #5270, opened 2026-07-14
fix: read a key's compacted leaf pages newest-first

### LINKED ISSUE #5266: FULL_TEXT index misses records: SEARCH_FIELDS returns fewer records than a plain scan for tokens present verbatim
**Version**: 26.7.2, official Docker image, single server, HTTP API.

## Summary
A freshly built `FULL_TEXT` index on a STRING property returns fewer records than a plain scan for tokens that are present verbatim in the indexed property. The loss is visible immediately after `CREATE INDEX` reports all records indexed (no writes in between, no custom analyzer, no filters in the query). Records whose property is EXACTLY equal to the searched token are
among the missing ones, which rules out any analyzer/tokenization explanation.

## Reproduction (observed on a real dataset)
Dictionary-like dataset: one vertex type (~204k records, single bucket) with a `words` STRING property holding space-separated lowercase tokens (vocabulary of ~162k distinct tokens, heavily shared across records; several record pairs share the IDENTICAL property value).

```sql
CREATE INDEX ON Entry (words) FULL_TEXT;
-- returns: created=true, totalIndexed=204414

SELECT lang, key FROM Entry WHERE SEARCH_FIELDS(['words'], 'casa');
-- returns 5 records

SELECT lang, key FROM Entry WHERE words = 'casa';
-- returns 1 record, NOT among the 5 above (its words value is exactly the token!)
```

More records missed by the same query, all with the token present verbatim:
`words = 'casa abitazione'`, `words = 'casa_astrologica segno_zodiacale segno casa'`
(8 records contain the standalone token `casa`, `SEARCH_FIELDS` returns 5).

Another telling pattern: pairs of records with IDENTICAL `words` values (e.g. two records both with `words = 'peripatidae family_peripatidae'`):
`SEARCH_FIELDS(['words'], 'peripatidae')` returns only ONE of the two.

Scale of the loss (measured by comparing, for every word of the vocabulary, `SEARCH_FIELDS` results against a client-side scan): for plain single-token lowercase words appearing in more than one record, roughly 40-45% of the words are missing at least one record. The loss looks record-based, not token-based: for a given missing record, every token of its `words` value
fails to find it.

## What did NOT reproduce it
A small synthetic dataset (21k tiny documents, 200-token vocabulary, index created after the inserts) shows NO loss: every token returns the complete record set. The problem seems to need a larger type (records count and/or record size), which suggests a page/sub-index boundary effect during the index build rather than a logical tokenization issue.

## Possible root cause area
The server log shows the index being built by `LSMTreeFullTextIndex`, i.e. the full-text index shares the LSM machinery of NOTUNIQUE indexes - and full-text tokens are heavily duplicated keys BY CONSTRUCTION (a common token appears in thousands of records). This looks like the same family as #5119/#5120 (non-unique LSM entries lost on heavily duplicated keys). #4946 made the LSM compaction atomic in 26.7.2, but during a large `CREATE INDEX` the build itself creates sub-indexes (log: `Creating sub-index ... with fileId=...`), which may be a residual path for the same loss.

## Workaround found
Client-side tokenization table: one record per (token, record key) with a UNIQUE composite index, populated once by splitting the STRING property.
Lookups by `token = ?` are complete; the FULL_TEXT index is not used.

## Possibly related (checked for duplicates on 2026-07-14: none matches)
#5119 (non-unique LSM index misses entries for keys with many duplicates),
#5120 (LSM compaction loses index entries, fixed by #4946 in 26.7.2),
#5181 (combined multi-property FULL_TEXT index does not index its first component - build-time loss too, but this report is about a SINGLE-property FULL_TEXT index with a partial per-token loss),
#5214 (bounded range scans over multi-series compacted LSM indexes drop rows - explicitly does NOT affect full-text indexes or exact-key lookups, which is what this report is about).

robfrank:
Related to #5266. **This does not close #5266** - see "Scope" below.

## Summary

A value removed from a key and then added back under that same key **permanently disappears** from the index, while a plain scan still returns the record. It is easiest to hit with a heavily duplicated full-text token, because a common token's posting list is what overflows a compacted leaf page.

`LSMTreeIndexAbstract.lookupInPageAndAddInResultset` resolves deletions with a **forward-poisoning** `deletedRIDs` set: a tombstone only suppresses RIDs encountered *after* it. The whole reader therefore walks newest to oldest - mutable pages (`LSMTreeIndexMutable:358`), compacted series (`LSMTreeIndexCompacted:530`), and values within a page (`LSMTreeIndexAbstract:663`).

The one exception was the per-key leaf-page loop at `LSMTreeIndexCompacted:580`. A key whose values overflow one leaf is written oldest-chunk-first onto ascending pages, each with its own root entry. That loop iterated the root entries **in order**, i.e. oldest chunk first, so a tombstone in an early chunk poisoned `deletedRIDs` and then suppressed the live re-add sitting in a later chunk.

The remove and the re-add must land in separate transactions to be reachable: within one transaction a REMOVE followed by an ADD on the same `(key, rid)` collapses to a single ADD in the transaction overlay, so no tombstone is written at all. That is why this survived existing coverage.

## Changes

- `LSMTreeIndexCompacted:580` - iterate the key's leaf pages newest-first, matching every other level of the reader. The preceding-leaf read (which holds the oldest chunks) already ran last and stays last.
- `LSMTreeIndexMutable.onAfterLoad()` - propagate `storeTermFrequency` to the compacted sub-index when it is re-attached. The flag is not persisted in the page header, so a sub-index materialised by the component factory defaults to `false`; a mismatch desynchronises the posting value stream (the tf/docLength varints are not consumed) and silently decodes the rest of the entry as garbage.

## Scope: this is not (yet) a fix for #5266

#5266 as reported - a fresh `CREATE INDEX` over 204k pre-existing records, no writes in between - **could not be reproduced on `main`**. A harness compared, for every token, the index result against a plain scan, and `countEntries()` against the expected posting count. No loss in any of: 204k records / 917k postings / 158k-token vocabulary; vertex type with ~350-byte records; `CREATE INDEX` inside an outer transaction (the server/HTTP path); page size 1024 and 65536; `indexCompactionRAM` = 1MB (forcing multi-chunk, multi-series compaction); four successive compaction rounds; database reopen; a token present in every record.

The reporter is on 26.7.2. Several compacted-LSM read fixes landed after that tag (#5210, #5214, #4943), and 26.7.2 demonstrably behaves differently - the same workload with a small index page size throws `UnsupportedOperationException: Root index page overflow` there, and not on `main`. #5266 may already be fixed; the reporter has been asked for their dataset and to retest against `main`.

This PR fixes a distinct, confirmed defect found while investigating, which produces the same user-visible symptom.

### Note on `limit`

With the reversal, a lookup passing `limit > -1` on an overflowing non-unique key now truncates to the **newest** values rather than the oldest. This is strictly an improvement, not just a semantic shift: under the old oldest-first order such a lookup could early-exit *before* reaching the later chunk that holds a tombstone, and therefore return a value that had actually been deleted. Newest-first cannot do that. The result is a `Set`, so caller-visible ordering is unaffected.

## Test plan

- [x] New `CompactedTombstoneLeafOrderTest`: 3000 documents share the token `data` (1024-byte pages, so its posting list provably spans many compacted leaves); a third have the token removed in one transaction and added back in the next; after compaction the index must return all 3000.
- [x] Verified it **fails without the fix** - exactly 1000 documents missing, precisely the ones touched - and passes with it.
- [x] Regression run: 551 tests / 90 classes across the index, LSM, compaction and full-text suites - 0 failures.

### DIFF
--- a/docs/5266-fulltext-index-misses-records.md
+++ b/docs/5266-fulltext-index-misses-records.md
@@ -1,0 +1,62 @@
+# 5266 - FULL_TEXT index misses records
+
+## Status
+
+Issue #5266 as reported (fresh `CREATE INDEX` over 204k pre-existing records, no writes in between)
+**could not be reproduced** on `main`. This change fixes a *different*, confirmed index-correctness
+defect found while investigating, which produces the same user-visible symptom. It does not close
+#5266.
+
+## What was ruled out for #5266
+
+A harness built the FULL_TEXT index over pre-existing records and compared, for every token, the index
+result against a plain scan, and `countEntries()` against the expected posting count. No loss on `main`
+in any of: 204k records / 917k postings / 158k-token vocabulary; vertex type with ~350-byte records;
+`CREATE INDEX` inside an outer transaction (the server/HTTP path); page size 1024 and 65536;
+`indexCompactionRAM` = 1MB (forcing multi-chunk, multi-series compaction); four successive compaction
+rounds; database reopen; a token present in every record (posting list spanning many leaves).
+
+The reporter is on 26.7.2. Several compacted-LSM read fixes landed after that tag (#5210, #5214, #4943),
+and 26.7.2 demonstrably behaves differently: the same workload with a small index page size throws
+`UnsupportedOperationException: Root index page overflow` there, and not on `main`. #5266 may already be
+fixed; the reporter has been asked for their dataset or a reproducer, and to retest on `main`.
+
+## Symptom (the defect fixed here)
+
+A value removed from a key and then added back under that same key **permanently disappears** from the
+index, while a plain scan still returns the record. Easiest to hit with a heavily duplicated full-text
+token, because a common token's posting list is what overflows a compacted leaf page.
+
+## Root cause
+
+`LSMTreeIndexAbstract.lookupInPageAndAddInResultset` resolves deletions with a **forward-poisoning**
+`deletedRIDs` set: a tombstone only suppresses RIDs encountered *after* it. The whole reader therefore
+walks newest to oldest - mutable pages (`LSMTreeIndexMutable:358`), compacted series
+(`LSMTreeIndexCompacted:530`), and values within a page (`LSMTreeIndexAbstract:663`).
+
+The one exception was the per-key leaf-page loop at `LSMTreeIndexCompacted:580`. A key whose values
+overflow one leaf is written oldest-chunk-first onto ascending pages, each with its own root entry.
+That loop iterated the root entries in order, i.e. **oldest chunk first**, so a tombstone sitting in an
+early chunk poisoned `deletedRIDs` and then suppressed the live re-add sitting in a later chunk.
+
+The remove and the re-add must land in separate transactions to be reachable: within one transaction a
+REMOVE followed by an ADD on the same (key, rid) collapses to a single ADD in the transaction overlay,
+so no tombstone is written at all.
+
+## Fix
+
+- `LSMTreeIndexCompacted:580` - iterate the key's leaf pages newest-first, matching every other level of
+  the reader. The preceding-leaf read (which holds the oldest chunks) already ran last and stays last.
+- `LSMTreeIndexMutable.onAfterLoad()` - propagate `storeTermFrequency` to the compacted sub-index when it
+  is re-attached. The flag is not persisted in the page header, so a sub-index materialised by the
+  component factory defaults to `false`; a mismatch desynchronises the posting value stream (the
+  tf/docLength varints are not consumed) and silently decodes the rest of the entry as garbage.
+
+## Tests
+
+`engine/src/test/java/com/arcadedb/index/CompactedTombstoneLeafOrderTest.java` - 3000 documents share the
+token `data` (1024-byte pages, so its posting list provably spans many compacted leaves); a third of them
+have the token removed in one transaction and added back in the next; after compaction the index must
+return all 3000. Fails without the fix (1000 documents missing - exactly the ones touched), passes with it.
+
+Regression run: 551 tests / 90 classes across the index, LSM, compaction and full-text suites - 0 failures.
--- a/docs/5266-fulltext-index-misses-records.md
+++ b/docs/5266-fulltext-index-misses-records.md
@@ -60,3 +60,27 @@ have the token removed in one transaction and added back in the next; after comp
 return all 3000. Fails without the fix (1000 documents missing - exactly the ones touched), passes with it.
 
 Regression run: 551 tests / 90 classes across the index, LSM, compaction and full-text suites - 0 failures.
+
+## PR
+
+https://github.com/ArcadeData/arcadedb/pull/5270 (related to #5266, does not close it)
+
+## Review cycles
+
+- **cycle 1** (`651046d`) - gemini: defensive null check on `subIndex` after `getFileById`. Declined: the NPE
+  risk is pre-existing and unchanged by the diff (the next line already dereferences it), and the whole block
+  is inside a `catch (Exception)` that logs SEVERE and drops the index, so the proposed `IndexException` is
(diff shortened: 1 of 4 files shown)


robfrank:
Thanks both. Went through each item; here's what I applied and what I skipped, with reasoning.

### Applied

- **`Record` fully-qualified name** (claude, #3). Fixed - `import com.arcadedb.database.Record;` is already used across the engine test suite (e.g. `TransactionBucketTest`, `RandomTestSingleThread`), so the explicit single-type import shadowing `java.lang.Record` is the established convention here. Pushed in 984fea2.
- **`limit` semantics** (claude, #1). Good catch, added to the PR body. Worth noting it's strictly an improvement, not just a shift: with the old oldest-first order, a lookup with `limit > -1` on an overflowing key could early-exit *before* reaching the later chunk holding a tombstone, and so return a value that was actually deleted. Newest-first cannot do that.

### Skipped, with reasoning

- **Null check on `subIndex`** (gemini). Declining. The NPE risk is pre-existing and unchanged by this diff - the very next line, `subIndex.mainIndex = mainIndex;`, already dereferences it and predates this PR. More importantly the whole block is inside a `try { ... } catch (final Exception e)` that logs SEVERE and drops the index, so an NPE and the proposed `IndexException` are already handled identically; the change would alter the log text and nothing else. CLAUDE.md also asks not to add validation for states that can't occur (a registered `subIndexFileId` always resolves). Happy to reconsider if there's a known path where `getFileById` returns null here.
- **`@Tag("slow")`** (claude, #2). Declining. The test measures 0.53-0.80s, which is not "noticeably long" under the CLAUDE.md guidance (multi-second). The retry loop's 5s ceiling is a safety bound, not the expected path - compaction completes on the first or second attempt.
- **`docs/5266-*.md` convention** (claude, #4). Declining. Per-issue investigation notes under `docs/` *are* an established convention in this repo - there are 25 tracked `docs/<issue-number>-<name>.md` files today (`docs/5023-idempotency-cache-keying-eviction.md`, `docs/5024-ws-event-bus-concurrency.md`, ...). Keeping it.

robfrank:
Thanks - responding to the three optional items from the latest pass.

**3. No dedicated regression test for the `storeTermFrequency` fix.** You're right that there isn't one, and I want to be straight about why rather than ship a test that passes for the wrong reason: **I could not construct a failing one.**

`LocalSchema.getFileById` (`LocalSchema:301-310`) returns the component already registered in the `files` list - not a fresh factory instance. So in every in-process path I could reach, the sub-index object handed back to `onAfterLoad()` is the same one that already carries the flag, and the new line sets `true -> true`. Tracing the real load path: `initComponents()` materialises the mutable *and* the compacted sub-index with the flag defaulted to `false`, the constructor's `onAfterLoad()` then propagates `false -> false` (a no-op), and `readConfiguration()` afterwards flips it to `true` via `LSMTreeFullTextIndex` -> `setStoreTermFrequency`, which propagates to the sub-index. So today the ordering already lands correctly and the line changes nothing observable. A reopen test I wrote passes identically with and without it.

I've kept it anyway, and I think that's the right call, but on narrower grounds than "it fixes a live bug": it makes the invariant local and explicit at the exact point the sub-index is attached, instead of depending on a cross-file ordering coincidence between `initComponents()` and `readConfiguration()`. The failure mode if that ordering ever changes is a desynchronised varint stream and silently garbage postings, which is about the worst class of silent corruption we could ship. Happy to drop it if you'd rather keep the diff strictly to the tested defect - it's one line plus a comment.

**1. `@Tag("slow")`.** Measured 0.53-0.80s across runs, so it stays untagged per your own "fine as-is if well under a second".

**2. `compact()` retry-loop robustness.** Noted. It's a bounded safety net rather than the expected path (compaction settles on the first or second attempt), but agreed that's the first place to look if it ever flakes on CI.

PR #5302, opened 2026-07-16
fix: edge-append merge reverted concurrent writes on multi-page edge chunks

lvca:
Fixes a silent data-corruption bug in the commutative edge-append merge (`GRAPH_EDGE_APPEND_MERGE`, on by default since 26.7.2), found while investigating a customer-reported `BufferUnderflowException` on a high-degree vertex. Internal tracking: https://github.com/ArcadeData/arcadedb-operations/issues/565

## The bug

When concurrent transactions append edges to the same vertex, a commit-time page conflict on the edge chunk is resolved by re-deriving the page from the current committed version and replaying the transaction's appends, instead of retrying the whole transaction.

That re-derivation is unsound when the chunk is stored as a **multi-page record** (a chunk allocated near the end of a page continues on the next page, `FIRST_CHUNK`/`NEXT_CHUNK`):

1. The drain writes the record's continuation slice through a page copy created only at drain time, so that page **passes the MVCC version check** while carrying the transaction's stale view of the continuation bytes.
2. The rebase evicts only the conflicted first page before re-reading the chunk, so the re-read walks the chunk chain **through the stale in-transaction copy** of the continuation page.
3. The result is committed: concurrently committed appends on the continuation page are silently reverted - zeroed chunk tails on young chunks, shifted/aliased edge-vertex pairs on filled ones, lost edges, `invalidLinks`/`corruptedRecords` in `CHECK DATABASE`, and `BufferUnderflowException` on later traversals (a shifted region can end mid-varint at the `used` boundary).

Only chunks that straddle a page boundary are exposed (~1/8 of cap-size chunks at the default 64 KB page), which is why this only shows up at scale, on hot vertices, under concurrent insertion.

## The fix

A chunk stored as a multi-page or indirected (placeholder) record is **not rebasable**: `rebaseEdgeAppends` now checks the record's layout marker on the current committed page (`LocalBucket.isRecordStoredInSinglePage`) and falls back to the standard retryable `ConcurrentModificationException`. Single-page chunks - the vast majority, including all super-node stripe chunks - keep the merge.

## Tests

- `EdgeAppendMergeMultiPageChunkTest` (regression): a 16 KB bucket page makes cap-size chunks straddle a boundary every other chunk; 8 threads x 1500 concurrent inserts on one hub, then the hub's raw pair stream is verified byte-exact (no zeroed bytes, no duplicates, no alien pairs, exact count) with the rebase asserted to have actually fired. Fails on every run without the fix; passes with it.
- `EdgeAppendMergeConcurrentStressTest`, `EdgeAppendMergeRaceTest`: concurrent add/remove/read races on the classic layout with the merge active.
- `SuperNodeDefaultThresholdTest`: super-node promotion at the shipped default threshold (4096), which promotes through the bounded chain-walk branch no test covered before.

Also verified at customer scale during the investigation: a 300k-edge hub with 8 concurrent writers and ~12k rebases fired ends with a byte-exact pair stream, and the full graph/database/engine test packages pass.

Please review before merging.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/TransactionContext.java
+++ b/engine/src/main/java/com/arcadedb/database/TransactionContext.java
@@ -696,6 +696,17 @@ public class TransactionContext implements Transaction {
 
       final LocalBucket bucket = (LocalBucket) database.getSchema().getBucketById(segmentRID.getBucketId());
 
+      // MULTI-PAGE / INDIRECTED CHUNK: the rebase re-derives ONE page, but a record whose content continues on
+      // other pages was drain-written there through page copies created only at drain time - copies that PASS
+      // the version check while carrying this transaction's STALE view of the record's continuation bytes.
+      // Rebasing would commit that stale slice, silently reverting concurrent appends on the continuation page
+      // (issue #565: zeroed/shifted chunk tails on chunks straddling a page boundary). Only a record living
+      // entirely in place on the conflicted page can be safely re-derived; anything else falls back to a full
+      // retry.
+      if (!bucket.isRecordStoredInSinglePage(segmentRID))
+        throw new ConcurrentModificationException(
+            "Edge-append rebase not possible: chunk " + segmentRID + " spans multiple pages. Please retry the operation");
+
       // Load the chunk fresh from the (now current) page, bypassing the tx record cache which still holds the
       // stale, already-appended instance. A concurrently-vanished chunk (RecordNotFoundException) cannot happen
       // under the held commit lock; if it ever did, fall back to a full retry rather than aborting the tx.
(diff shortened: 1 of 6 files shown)


lvca:
Addressed the review feedback in 996a46c37:

- **Defensive RID guard** - `isRecordStoredInSinglePage` now returns false for a null RID or negative position before any index arithmetic.
- **Exception cause preserved** - the `IOException` is passed as the cause when wrapping in `DatabaseOperationException`.
- **Shared marker decoding** - extracted `readRecordSizeMarker` as the single place that decodes a record slot's size marker; `existsRecord` and `isRecordStoredInSinglePage` both use it now, so they stay in lockstep if the page layout ever changes. `existsRecord`'s external behavior is unchanged.
- **Test conventions** - all four new tests are now `@Tag("slow")` (matching `ConcurrentEdgeAppendMergeTest`) and the debug `System.out` left from the investigation is removed; the AssertJ `.as(...)` descriptions carry the diagnostic context on failure.

Re-ran the four new tests plus the full `com.arcadedb.engine` and `com.arcadedb.database` packages after the refactoring - all green. Ready for another look.

Issue #5295, opened 2026-07-15
HA: after stuck-diverged self-reformat, follower never receives a single append (matchIndex=-1 forever) — RUNNING and leader-aware but permanently empty, leader reports it HEALTHY


ruispereira:
Follow-up to #5291, but a distinct defect — filing separately per the pattern of the previous reports. On `26.8.1-SNAPSHOT` (digest `8ee87c28…`, all 3 nodes): the new stuck-diverged self-reformat **fires correctly, but the rejoin after it never completes** — the follower ends up `RUNNING` and leader-aware with **`matchIndex: -1` forever** (not even the term's CONFIGURATIONENTRY gets appended), while the leader's status API reports it `HEALTHY`. No errors on either side; readiness 503 indefinitely.

## Sequence (UTC, 2026-07-15 — same cluster as #5291, immediately after its full parallel reformat)

1. **~09:29** — full parallel reformat to escape the #5291 crash loop: wipe `raftStorageDirectory` on all 3 nodes + simultaneous `--grace-period=0 --force` delete. Caveat that set up this scenario: `arcadedb-0` was in the #5291 3-second crash-restart loop when the wipe ran, and the in-place member restart **re-persisted the storage between the `rm -rf` and the SIGKILL** — so while pods 1/2 came back clean and bootstrapped a fresh group (term=5, commitIndex=-1), `arcadedb-0` came back with resurrected pre-wipe state (`appliedTerm=4, appliedIndex=4967`).
2. **09:33:40** — the new HealthMonitor correctly detects it:

   ```
   WARNI [HealthMonitor] Follower stuck-diverged from leader for 60011ms, reformatting Raft storage and rejoining (attempt 1)
   WARNI [RaftHAServer] Follower stuck at a stale term (currentTerm=5, appliedTerm=4, appliedIndex=4967, commitIndex=4967); reformatting Raft storage and rejoining so the leader can reconcile it via snapshot-install.
   ```

   👏 — this is exactly the classification we asked for in #5291, and it works for the stale-term case.
3. **09:33:40.392** — during the reformat restart, the leader's in-flight append of the configuration entry is cancelled (expected, the member is restarting):

   ```
   WARNI [GrpcServerProtocolService] APPEND_ENTRIES onError, lastRequest: ...arcadedb-2..._2434 -> ...arcadedb-0..._2434#1-t5,
[...]

Issue #4841, opened 2026-06-30
`ClusterMonitor` retains per-replica state across leadership terms — false `FALLING_BEHIND`/`STALLED` classifications after re-election


ruispereira:
**Affected version:** 26.6.1
**Component:** `RaftHAServer.getFollowerStates`
**Severity:** LOW (robustness — unverified against Ratis internals)

### Summary

`peerId`/`lastRpcElapsedMs` come from `roleInfo.getLeaderInfo().getFollowerInfoList()`
while `matchIndex`/`nextIndex` come from `getFollowerMatchIndices()` /
`getFollowerNextIndices()`, zipped by index `i`. Ratis does not document that the
two iteration orders match; if they diverge (version change / membership change
mid-call), a follower's matchIndex is attributed to the wrong peer ID, feeding
wrong values into the lag monitor and status table. The `min(...)` guard protects
length, not order.

### Code

`RaftHAServer.java:1139-1153`.

### Impact

Potential mis-attributed lag/STALLED status per peer. Could not obtain Ratis source
to prove a concrete divergence — flagged as a robustness concern, not a verified
defect.

### Suggested fix

Build the result keyed by `RaftPeerId` from a single source, or look up match/next
index by the follower's peer id rather than positional index.

lvca:
Fixed in `main` (commit d38a9fca5).

Two separate things were in this report:

- The **body** (positional zip of `matchIndex`/`peerId` in `getFollowerStates`) was already fixed earlier in #4842, so nothing new was needed there.
- The **title** flagged a real bug: the cluster lag monitor kept its per-replica state from a previous leadership term. After a re-election, a follower's match index restarts and is briefly compared against the old (higher) baseline, so a perfectly healthy replica could be wrongly reported as `STALLED` (and in rare cases trigger an unnecessary resync).

The fix clears the lag-monitor state every time a node becomes leader, so each term starts from a clean baseline. No configuration change is required, and there is no impact on normal operation. Added regression tests to cover the re-election case.

PR #5353, opened 2026-07-20
fix(#5346): escalate a permanently wedged follower replication channel

### LINKED ISSUE #5346: HA: leader's replication channel to a re-IP'd follower stays permanently dead (matchIndex=-1); in-process channel resets (5/5) never recover it — only a leader restart does
## Summary

After a follower pod was recreated (new IP), it joined the Raft membership but the **leader's replication channel to it never delivered a single append** (`matchIndex = -1`) for 5.5+ minutes. The leader's own in-process recovery — periodic "replication-channel reset to force fresh DNS re-resolution and reconnect", governed by `arcadedb.ha.peerChannelResetDuration` — ran **all 5 attempts and never recovered the channel**. Network connectivity to the follower was verified healthy the whole time (DNS resolved to the new IP, TCP to the Raft port was open). The follower only started receiving appends after the **leader process was restarted**, which built a fresh appender, shipped a snapshot, and let the follower catch up to `lag = 0` within seconds.

So: a replication channel wedged into a state that the leader's automated reset logic could not clear; only a full leader restart did. The leader's log itself acknowledges this, suggesting a manual "transfer leadership to that follower's healthy peer to rebuild the appender" as the remedy.

## Environment

- ArcadeDB **26.8.1-SNAPSHOT**, 3-node HA on Kubernetes (Apache Ratis backend), headless Service with `publishNotReadyAddresses: true`, one PVC per pod.
- Relevant live config:
  ```
  arcadedb.ha.peerUnreachableThreshold  = 10000   # ms before a peer is "unreachable"
  arcadedb.ha.peerChannelResetDuration  = 60000   # ms between channel-reset attempts
  arcadedb.ha.electionTimeoutMin        = 5000
  ```

## Timeline (single leader `server-0` throughout the wedge)

1. Follower `server-2` was recreated by the StatefulSet and came up with a **new pod IP** (`10.126.12.61` → `10.126.5.239`).
2. `server-2` joined membership cleanly:
   ```
   INFO [RaftHAServer] <server-2> Raft cluster joined: 3 nodes [server-0, server-1, server-2]
   INFO [ArcadeStateMachine] This node is now REPLICA (leader: server-0 ... term=6)
   ```
3. But the leader never delivered an append to it. For 5.5+ minutes the leader logged, every 30 s:
   ```
   SEVER [ClusterMonitor] Replica 'server-2..._2434' has NEVER received a single append
     (matchIndex=-1) while the leader committed up to 50098, for 315000ms: its replication path
     is dead. The leader will force a resync to re-engage it; if it persists, transfer leadership
     to that follower's healthy peer to rebuild the appender.
   ```
4. The in-process reset fired on schedule and **exhausted all 5 attempts** without recovering:
   ```
   WARNI [ClusterMonitor] Follower 'server-2..._2434' unreachable for  60000ms; triggering a replication-channel reset (attempt 1/5) ...
   WARNI [ClusterMonitor] Follower 'server-2..._2434' unreachable for 120000ms; triggering a replication-channel reset (attempt 2/5) ...
   WARNI [ClusterMonitor] Follower 'server-2..._2434' unreachable for 180000ms; triggering a replication-channel reset (attempt 3/5) ...
   WARNI [ClusterMonitor] Follower 'server-2..._2434' unreachable for 240000ms; triggering a replication-channel reset (attempt 4/5) ...
   WARNI [ClusterMonitor] Follower 'server-2..._2434' unreachable for 300000ms; triggering a replication-channel reset (attempt 5/5) ...
   ```
   The cluster status table showed `server-2` frozen at `matchIndex=-1`, cycling `STALLED`/`CATCHING_UP`, lag pinned at ~50090 (never advancing).
5. **The recovery required restarting the leader.** After `server-0` was restarted, the (new) leader built a fresh appender, served a snapshot, and `server-2` caught up immediately:
   ```
   INFO [SnapshotHttpHandler] Serving database snapshot for 'superx'...
   # cluster table, seconds later:
   | server-2..._2434 | ... | Follower | 0 | 111.27 ms | 1191 ms | HEALTHY |
   ```

## Connectivity was healthy during the wedge (this is the key point)

The reset logic assumes the failure is stale DNS / a dead connection. Measured from the **leader** while it was reporting `server-2` "unreachable / matchIndex=-1":

```
# DNS from leader resolves to the NEW IP (not stale):
$ getent hosts server-2.<svc-fqdn>
10.126.5.239  server-2.<svc-fqdn>

# TCP from leader to the follower's Raft port is OPEN:
$ nc -zv server-2.<svc-fqdn> 2434
... (10.126.5.239:2434) open

# follower is listening on 2434, and its new IP is in the Service endpoints.
```

So it was **not** stale DNS and **not** an L3/L4 reachability problem — yet the gRPC append channel never established, and 5 automated resets did not fix it. This is what makes it look like the channel object got stuck in a bad internal state that only a process restart clears.

## Honest caveat — mixed build during the incident

During the wedge the cluster was **running two different SNAPSHOT builds** (our deployment uses a moving `:26.8.1-SNAPSHOT` tag with `imagePullPolicy: Always`, so a recreated pod can pull a newer build):

- leader `server-0` / `server-1`: build `8017584896…`
- wedged follower `server-2`: build `792bdff3…`

I therefore cannot fully isolate whether the dead channel was caused by (a) a wire/protocol incompatibility between the two builds, or (b) a genuine channel-recovery defect independent of build. The leader restart that fixed it changed **both** variables at once (builds converged to `0a793578…` *and* the leader process/appender was rebuilt).

Both readings still point at a bug worth addressing:

- **If (b):** the leader's `peerChannelReset` recovery is unable to clear a wedged appender to a re-IP'd follower, despite healthy DNS/TCP; only a leader restart works. The 5-attempt automated recovery gives the operator false confidence.
- **If (a):** two builds that cannot replicate should fail **loudly** (clear protocol/version error), not degrade into a silently "unreachable" follower that survives 5 reset attempts and looks like a network problem. A version-handshake mismatch surfaced as a dead channel is itself a diagnosability bug.

## Suggested improvements

1. When `peerChannelReset` fires, actually rebuild the gRPC `ManagedChannel`/appender from scratch (new channel object, fresh name resolver) rather than nudging the existing one — and log the resolved target IP before/after so operators can see whether re-resolution took effect.
2. After the reset attempts are exhausted, perform the leadership-transfer / appender-rebuild automatically instead of only *suggesting* it in the log (today the only remedy is a manual leader restart).
3. Add a peer version/protocol handshake on the Raft channel so an incompatible peer is reported explicitly (`peer build X incompatible with Y`) instead of appearing as a generic "unreachable" follower.
4. Surface `matchIndex=-1` "never received an append" as a distinct, first-class alert condition (it is qualitatively different from a follower that is merely lagging).

## Notes

- Related prior fix: leader auto-recovery of a follower channel wedged on stale DNS after a Kubernetes pod-IP change was reported fixed in 26.7.2 ([#4696](https://github.com/ArcadeData/arcadedb/issues/4696)); this report is a case where that auto-recovery did **not** succeed (though under the build-skew caveat above).
- The underlying reason a follower had to be recreated at all (unbounded Raft log filling the disk) is filed separately as [#5345](https://github.com/ArcadeData/arcadedb/issues/5345).

robfrank:
Closes #5346

## Summary

When the leader's bounded replication-channel reset budget (`arcadedb.ha.peerChannelResetDuration`, 5 attempts) was exhausted for a follower whose channel was still dead, `ClusterMonitor` logged a single SEVERE line telling the operator to transfer leadership and then did nothing further. The reset streak only re-arms when the follower becomes reachable again, which never happens for a genuinely wedged channel, so the leader stayed stuck until the process was restarted - exactly what recovered the reported incident. Separately, `resetProxy` is fire-and-forget (Ratis rebuilds the channel lazily on the appender's next send, so a wedged appender makes the reset a silent no-op that still burns an attempt), it ran inline on the single lag-monitor thread, and it logged nothing an operator could use to tell whether re-resolution took effect.

`ClusterMonitor` now fires an optional exhausted-channel handler exactly once per unreachable streak. `RaftHAServer.escalateWedgedPeerChannel` implements it by transferring leadership to a healthy peer, which rebuilds the appender the same way a leader restart does. The target is chosen with the existing step-down rules and is never the wedged follower itself; when no healthy candidate exists the previous operator-intervention log is kept rather than flapping leadership. The reset and the escalation both moved onto the existing bounded recovery executor, and the reset now logs the peer's resolved IP addresses before and after.

Gated by a new `arcadedb.ha.peerChannelResetEscalation` (Boolean, default `true`); set it to `false` to restore the previous log-and-stop behaviour.

Suggestions 3 (peer build/protocol version handshake) and 4 (`matchIndex=-1` as a first-class alert) from the issue are out of scope: the former is a separate feature, the latter already landed with #5295. Rationale is recorded in `docs/5346-leader-replication-channel-wedged.md`.

## Test plan

- [x] `ClusterMonitorTest` - 4 new deterministic clock-driven tests: escalation fires once on budget exhaustion, re-arms after the follower reconnects, keeps the old single SEVERE give-up line when no handler is wired, and contains a throwing handler without retrying it.
- [x] TDD verified: the 3 behavioural tests fail against the pre-fix code path and pass after it.
- [x] `RaftHAServerChannelEscalationTest` - new unit tests for the target-selection seam, covering exclusion of the wedged follower and of this node, priority ordering, lagging-peer skip, and the new config default.
- [x] Full `ha-raft` module suite: 702 tests, 0 failures.
- [x] Config regression tests in `engine` (`ConfigurationTest`, `ContextConfigurationTest`, `GlobalConfigurationTest`, `GlobalConfigurationReadinessHATest`): 46 tests, 0 failures.


robfrank:
Thanks both - all three points were verified against the code and are now addressed in 7393935.

**Leadership flapping (claude, design concern) - real, fixed.** I confirmed the mechanism: `clusterMonitor.reset()` is called on every leadership acquisition (`startLagMonitor`, issue #4841), which wipes `channelResetGaveUp`. So each new leader genuinely arrived with a fresh escalation budget, and in a >=3-node cluster with a permanently unreachable follower the escalation would ping-pong between healthy peers indefinitely. That is a regression this PR would have introduced.

The bound cannot live in `ClusterMonitor` for exactly that reason, so it now lives in `RaftHAServer`, which is node-local and is not reset on leadership change: `admitChannelEscalation` admits at most one escalation per follower per 30-minute cooldown. A permanently dead follower now costs at most one transfer per healthy peer per window, after which every peer refuses and the cluster settles on the operator-intervention path. Documented in the config description and the tracking doc, and covered by 3 new unit tests.

**CallerRunsPolicy (both) - real, fixed.** Rather than retune the shared executor, channel recovery moved to its own single-worker `channelRecoveryExecutor` with a discard-and-warn rejection policy. This addresses both halves of the concern: it can no longer queue behind a long blocking resync HTTP call (gemini's 120s scenario), and saturation can never push a blocking DNS lookup or the 10s `transferLeadership` back onto the lag-monitor thread. A dropped reset is free (retried next interval); a dropped escalation is logged loudly as the one-shot it is.

**`extractHost` (gemini) - real, fixed.** Correct catch, and it was a genuine bug rather than just duplication: my `lastIndexOf(':')` parser mishandles bracketed IPv6 literals, which `extractHost` handles.

**Noted, no change:**
- *Two DNS lookups per reset* - kept, since the before/after pair is the diagnostic value the issue asked for, and they now run on a dedicated executor so they cannot delay resyncs.
- *`channelResetGaveUp` latch stays set after a successful transfer* - agreed this is moot; the node stops being leader and the rebuilt channel un-wedges the follower.
- *Testing the async `escalateWedgedPeerChannel` wiring end-to-end* - the new cooldown logic was extracted as a static seam and unit-tested, which covers the decision this PR adds. The remaining post-dequeue guard still needs a live division to exercise.

Verification: full `ha-raft` suite 705 tests green, engine config tests 46 green.

robfrank:
Cycle 2 addressed in 471a08c.

**1. Cooldown consumed even when no transfer happens - real bug, fixed.** Confirmed: `admitChannelEscalation` ran before target selection, so a null-target give-up burned the 30-minute window, and the subsequent refusal claimed *"already transferred leadership"* when no transfer had occurred. Worse, as you noted, a healthy peer rejoining right after a null-target give-up would have been suppressed for up to 30 minutes.

The gate is now the last thing before the transfer, after target selection, so the window is consumed only when a transfer is actually attempted. A failed transfer still consumes it - that is deliberate, since the attempt did happen and retry-storming a failing transfer is not desirable - and the message now says "attempted a leadership transfer". New test `aRefusedTargetSelectionLeavesTheCooldownWindowIntact` pins the ordering contract.

**2. Discarded escalation is terminal for the streak - fixed.** You are right that the one-shot latch and a lossy executor were badly coupled. Rather than keep a blanket discard policy, the executor now keeps the default abort policy and each submitter handles rejection according to its own semantics: a dropped reset logs FINE (the monitor retries it next interval), a dropped escalation logs SEVERE naming it as not-retried-for-this-outage. The lag-monitor thread is still never used to run the work, which was the original point.

**3. `admitChannelEscalation` get-then-put not atomic - fixed.** Converted to `putIfAbsent` + `replace` compare-and-set with a retry loop, so correctness no longer depends on the implicit single-worker invariant. Cheaper to make it correct than to document the constraint.

**4. Hardcoded `ESCALATION_COOLDOWN_MS` / `ESCALATION_TRANSFER_TIMEOUT_MS` - declined, deliberately.** Both are on a rare terminal recovery path, and the transfer timeout intentionally matches the existing `stepDown` value. Adding two more `HA_*` keys expands the operator-facing config surface for knobs with no established tuning need; if a real deployment turns out to need them, promoting a constant to a config key later is trivial and non-breaking. Happy to add them now if you would rather.

**On the end-to-end IT:** agreed it is the highest-value missing test, and I want to be straight about why it is not here. The core hypothesis - a fresh appender on a new leader recovers the channel - needs a *genuinely wedged* gRPC appender, and the in-process harness (`BaseRaftHATest`) has no seam to wedge Ratis's internal appender. Killing or partitioning a follower does not reproduce it, because that path already recovers today. Faking it would test the mock, not the hypothesis. So the honest status is: the decision logic is unit-tested, and the recovery mechanism itself rests on the reasoning that a leader restart demonstrably fixed the reported incident and a leadership transfer rebuilds the same state. That is recorded in the tracking doc rather than implied to be verified.

Full `ha-raft` suite: 706 tests green.

robfrank:
Cycle 3 addressed in f3c9bff. No blocking issues raised; these are accuracy fixes.

**1. A dropped reset still burns an attempt - correct, and my log line was wrong.** Verified: `ClusterMonitor` increments `channelResetCount` before invoking the handler, so a task dropped by a saturated queue has already consumed one of the five attempts. My FINE message claiming it is "retried next interval" understated that. Raised to WARNING and it now states the attempt was consumed, with a comment explaining that this is only reachable when several followers wedge simultaneously. I chose to report rather than compensate: threading a "did it actually run" signal back through the `Consumer<String>` handler would complicate the monitor's contract for an edge case that a WARNING makes diagnosable.

**2. Failed transfer consumes the cooldown - intended, now commented.** Added the rationale inline: the transfer *was* attempted, and a node that cannot hand off leadership would otherwise re-attempt every reset cycle. Recovery still comes from another peer, or from this one after the window.

**4. Cross-thread `ClusterMonitor` reads - confirmed safe, now documented.** You are right that the recovery thread is a new reader. The invariant holds for the same reason `stepDown()` is safe (`replicaStates` is a `ConcurrentHashMap`, `leaderCommitIndex` is volatile), and a target chosen from a tick-old snapshot is still fine because Ratis rejects a transfer to a peer that has since become unsuitable. Written into the `escalateWedgedPeerChannel` javadoc rather than left implicit. Also fixed a stale `{@link #stalledResyncExecutor}` reference in that javadoc from before the executor split.

**3 and 5** - agreed on both, no change. The orchestration-coverage boundary and the bounded-but-nonzero churn are both recorded in the tracking doc.

**On the docs nit:** checked the repo rather than guessing - `docs/<issue>-<name>.md` is an established convention with 33 committed instances (`5326-`, `5345-`, `5325-`, `5311-`, ...), so this one is consistent with it.

Full `ha-raft` suite: 706 tests green.

This is cycle 3 of the 4 I was given, and the remaining items are acknowledged boundaries rather than open defects, so I am stopping here and leaving the merge decision to you.

robfrank:
Docs follow-up filed as ArcadeData/arcadedb-docs#422 for `arcadedb.ha.peerChannelResetEscalation`. It also covers `ha.peerChannelResetDuration` and `ha.peerUnreachableThreshold`, which turned out to be undocumented too - the new setting is defined in terms of both, so documenting it alone would not have been usable.

Issue #575, opened 2022-10-19
SLF4J: Failed to load class ...


gramian:
### OS: Ubuntu 22.04

I get the log message below when starting the server. Even though this is logged as "INFO" I guess this is still unwanted behavior(?).
The link in the log message https://www.slf4j.org/codes.html#StaticLoggerBinder offers a solution.

## Actual behavior
```
2022-10-19 20:51:22.755 INFO  [ArcadeDBServer] <ArcadeDB_0> Starting ArcadeDB Server with plugins [] ...SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

## Steps to reproduce
Just starting the server.

PR #5356, opened 2026-07-20
fix(#575, #593): revert ineffective snapshot-term clamp, tolerate benign applied-term regression at apply time

lvca:
## Problem

A follower-installed snapshot can record an **inflated applied term**: `notifyInstallSnapshotFromLeader` uses `firstTermIndexInLog.getTerm()` - the term of the first log entry AFTER the snapshot - as the snapshot's term. That entry is not in the follower's log at install time, and a later leadership-change reconciliation can settle it on a **lower** term.

When Ratis then applies the real next committed entry (index advancing, term lower than the over-recorded snapshot term), `BaseStateMachine.updateLastAppliedTermIndex`'s strict term-first monotonic check throws `IllegalStateException: Failed updateLastAppliedTermIndex`, killing the `StateMachineUpdater` thread. The node crash-loops and the whole cluster wedges with all peers `VOTING_FOR_ME` - the exact field state in ArcadeData/arcadedb-operations#593 / #575 (`snapshot.11_39707283` on two nodes while the next committed entry at 39707284 has term 10).

## Why revert #5343

The clamp merged in #5343 (`e45a70cc1`) reads the **local Raft log** to lower the marker term at seed time - but at both seed points the log is provably unavailable, so the clamp always fell through to its fallback and left the term inflated (a runtime no-op):

- `reinitialize()` runs inside `initialize()` **before** the RaftServer registers the division, so `getDivision(groupId)` returns null and no log can be resolved.
- The install path writes the marker **before** the boundary entry arrives via AppendEntries, so `getTermIndex(snapshotIndex + 1)` returns null.

Proven empirically on a live 3-node `MiniRaftClusterWithGrpc` restart with diagnostics: `DIAG reinitialize: markerTerm=6 snapshotIndex=38 log=null nextEntry(39)=null seedTerm=6`. The #5343 unit tests passed only because they injected a pre-populated stub `RaftLog` - a state that never exists at runtime. First commit reverts it.

## The fix: tolerate the benign shape at apply time

Apply time is the one moment the real next-entry term is actually known (Ratis hands it to us). `ArcadeStateMachine` now overrides `updateLastAppliedTermIndex(TermIndex)`:

- **Index strictly advances AND term strictly regresses** -> this shape cannot be produced by a committed Raft log (a committed log never has a lower term at a higher index), so it is unambiguously the inflated-snapshot-seed correction. Accept it via Ratis's own `setLastAppliedTermIndex`, realigning the term down, and log a WARNING.
- **Every other ordering** delegates to `super`, so genuine violations (same-index term regression, index going backwards) still fail loudly.

## Testing

New `Issue593ToleratedAppliedTermRegressionTest` drives the **real** `BaseStateMachine.updateLastAppliedTermIndex` (no stubbed log) on a real `ArcadeStateMachine` over real `RaftStorage`, with the client's exact numbers `(t:11, i:39707283) -> (t:10, i:39707284)`:

- inflated-term correction is tolerated and the term realigns down (fails on the unpatched engine)
- same-index and backwards-index regressions still throw
- normal forward progress unaffected
- predicate matrix: only "index up, term down" is benign

Plus 16 related HA tests green, including `RaftSplitBrain3NodesIT` (real-Ratis restart/failover) and `Issue5277SnapshotShipsLastTxIdIT`.

## Note for already-wedged clusters

A node stuck on an inflated on-disk marker **self-heals once it restarts on a build containing this fix**: `reinitialize()` seeds the inflated term while the previous applied TermIndex is still null (no violation possible), and the first re-applied entry then takes the tolerant branch. The stale marker filename is cosmetic and gets replaced by the next snapshot. The new `restartLifecycleSelfHealsThroughLongOverload` test replays exactly this lifecycle through Ratis's `(long,long)` apply-path overload.

The manual recovery (stop the node, rename `sm/snapshot.<inflatedTerm>_<index>` to the correct lower term, restart) remains the workaround only for clusters that must stay on an **unfixed** build.

Fixes ArcadeData/arcadedb-operations#593
Fixes ArcadeData/arcadedb-operations#575


lvca:
Addressed in a0fc13693, point by point:

**Ratis coupling** - confirmed against the Ratis 3.2.2 source: `transactionFutures` completion runs only on the no-op path (`newTI.equals(oldTI)`), and as you note the index gap in the tolerated shape is empty anyway. The comment is now pinned to `Ratis 3.2.2` with an explicit RE-VERIFY-on-upgrade flag, plus a note that the read-then-set is safe because applied-index updates are confined to the single StateMachineUpdater thread.

**Auto-heal** - you're right, and this was the most valuable catch: traced the restart path and the fix does self-heal a node stuck on an inflated marker (`reinitialize()` seeds while the previous applied TermIndex is still null, then the first re-applied entry takes the tolerant branch; the stale marker filename is cosmetic and replaced by the next snapshot). Updated the override Javadoc and corrected the PR description: the manual marker rename is now documented as a workaround only for clusters staying on an unfixed build.

**Test fidelity** - added `restartLifecycleSelfHealsThroughLongOverload`, which covers both suggestions at once: it drives Ratis's protected `updateLastAppliedTermIndex(long, long)` (the exact `applyTransaction` entry point, via reflection since it's protected in a foreign package; `Method.invoke` still dispatches virtually) and replays the full wedged-node restart lifecycle (null -> inflated seed -> tolerated next entry -> term realigned). The tolerance WARNING now observably fires from both overload paths in the test run. 6/6 green; the WARNING-count assertion I left out for now since LogManager capture plumbing felt out of proportion for this PR, but happy to add it if you want the observability contract pinned.

lvca:
Second round addressed in a131c5950:

**1. Future completion on the advancing path (your gate)** - confirmed and now both pinned and tested. The Ratis 3.2.2 `BaseStateMachine.updateLastAppliedTermIndex` body is quoted verbatim in the override comment: the advancing path is `getAndSet` + assert + `return true` with NO future completion; the `synchronized (transactionFutures) { ... complete(null) ... }` block is reachable only on the equal/no-op path. So the counter-intuitive claim holds - super itself never completes futures on an advancing update; they complete on the next duplicate (equal) update, which is not a benign regression and therefore delegates to super. The new `pendingStaleQueryFutureStillCompletesAfterToleratedUpdate` test proves it end-to-end with a real `queryStale()` future: pending after registration, still pending after the tolerated realignment (identical to super's advancing path), completed by the subsequent duplicate update. A dropped future would fail that last assertion, so a hung stale read can no longer slip through silently.

**2. Non-atomic read-then-set** - the comment now names all three call sites and why they cannot interleave: `applyTransaction` runs on the single StateMachineUpdater thread, and both snapshot seeds run while that thread is not applying entries (`reinitialize()` is invoked by the updater itself, at startup or from `reload()` while PAUSED; `notifyInstallSnapshotFromLeader` pauses the state machine for the install).

**3. WARNING observability** - the WARNING is already unconditional: it is emitted by `LogManager` on every tolerated transition with no rate limiting in this path (the throttled-60s pattern used elsewhere in the codebase is not applied here). So a recurrence from a hypothetical future bug producing this shape would log once per occurrence - visible in operator logs even if the fail-loud halt is gone for this one shape. Agreed on the trade-off framing; keeping the tolerance narrow (strictly index-up AND term-down, everything else still delegates to the strict check) is the mitigation.

**Minor** - added the code comment at the `reinitialize()` seed noting the stale `snapshot.<inflatedTerm>_<index>` filename persists across restarts until the next snapshot rolls it over (cosmetic, expected), for the next maintainer who greps for it.

7/7 tests green.

lvca:
Addressed in 7c2fd7d83. The one actionable item - the predicate being intentionally wider than the failure it targets - is now documented as an explicit SAFETY PRECONDITION in the `isBenignSnapshotTermRegression` javadoc: the shape is provably benign only because every caller feeds committed log entries (or the lifecycle-serialized snapshot seed), and synthetic/uncommitted TermIndex values must never be routed through `updateLastAppliedTermIndex`, since a genuine bug with this exact shape would be tolerated with a WARNING instead of failing loudly.

On the two "radar" points: agreed on both, and they are intentionally left as documented assumptions - the Ratis future-completion coupling is guarded by `pendingStaleQueryFutureStillCompletesAfterToleratedUpdate` (an upgrade that changes the semantics fails that test rather than shipping silently), and the threading-model dependency of the read-then-set is captured at the call-site comment.

Left the tests untagged: the whole class runs in ~0.12s, well under any CI-noise threshold.

7/7 tests still green after the change.

Issue #2813, opened 2025-11-20
Irregular Null Pointer Exception on shut down


gramian:
`ArcadeDB Server v25.11.1-SNAPSHOT (build 8bcd28ff549383a6e8717adb952f26d0c2dc12b2/1763572484536/main)`

`Running on Mac OS X 15.6 - OpenJDK 64-Bit Server VM 25.0.1 (Homebrew)`

On shutdown of the server sometimes I get:

```
java.util.logging.ErrorManager: 1
java.lang.NullPointerException: Cannot invoke "java.io.Writer.write(String)" because "writer" is null
	at java.logging/java.util.logging.StreamHandler.publish(StreamHandler.java:222)
	at java.logging/java.util.logging.FileHandler.publish(FileHandler.java:743)
	at java.logging/java.util.logging.Logger.log(Logger.java:962)
	at java.logging/java.util.logging.Logger.doLog(Logger.java:989)
	at java.logging/java.util.logging.Logger.log(Logger.java:1012)
	at com.arcadedb.log.DefaultLogger.log(DefaultLogger.java:257)
	at com.arcadedb.log.LogManager.log(LogManager.java:58)
	at com.arcadedb.server.ArcadeDBServer.lambda$init$0(ArcadeDBServer.java:708)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

This happens irregularly, so it is not happening reproducibly.

PR #5361, opened 2026-07-21
Add opt-in SLF4J facade logging (Slf4jLogger), keeping java.util.logging as default (#4276)

### LINKED ISSUE #4276: ArcadeDB writes logs directly to stdout instead of using SLF4J, producing inconsistent and unmanageable logs in host applications
**Affected version:** ArcadeDB 26.4.2 (arcadedb-engine, likely all modules using com.arcadedb.log.LogManager)
**Component:** logging — com.arcadedb.log.LogManager
  
**Summary**

ArcadeDB's internal LogManager writes log records directly (with its own format) instead of delegating to a logging facade such as SLF4J. When ArcadeDB is embedded in or co-deployed with a host application that uses SLF4J/Logback (Spring Boot in our case), two distinct log streams end up interleaved in stdout — with different timestamps, different fields, occasionally missing  newlines, and no shared MDC/trace context.
  
**Example**

From a Spring Boot service that includes arcadedb-engine/arcadedb-network:
  
```
2026-05-21 11:22:53.431  WARN [vcc-superx-data-service,,] 27385 --- [           main] o.s.s.c.a.web.builders.WebSecurity       : You are asking Spring Security to ignore Deferred [Mvc [pattern='*'], Ant [pattern='*']]. This is not recommended -- please use permitAll via HttpSecurity#authorizeHttpRequests instead.                                                                                                                                                                                                                                                                                                                                                
2026-05-21 11:22:53.431  INFO [vcc-superx-data-service,,] 27385 --- [           main] o.s.s.web.DefaultSecurityFilterChain     : Will not secure Deferred [Mvc [pattern='*'], Ant [pattern='*']]                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                             
2026-05-21 11:22:53.821 INFO  [Http11NioProtocol] Starting ProtocolHandler ["http-nio-1292"]2026-05-21 11:22:53.825  INFO [vcc-superx-data-service,,] 27385 --- [           main] o.s.b.w.embedded.tomcat.TomcatWebServer  : Tomcat started on port 1292 (http) with context path ''                                                                                                       
2026-05-21 11:22:53.857  INFO [vcc-superx-data-service,,] 27385 --- [           main] o.s.b.w.embedded.tomcat.TomcatWebServer  : Tomcat initialized with port 1293 (http)                                                                                                                                                                                                                  
2026-05-21 11:22:53.867  INFO [vcc-superx-data-service,,] 27385 --- [           main] w.s.c.ServletWebServerApplicationContext : Root WebApplicationContext: initialization completed in 35 ms                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                             
2026-05-21 11:22:53.857 INFO  [Http11NioProtocol] Initializing ProtocolHandler ["http-nio-1293"]                                                                                                                                                                                                                                                                                           
2026-05-21 11:22:53.857 INFO  [StandardService] Starting service [Tomcat]                                                                                                                                                                                                                                                                                                                  
2026-05-21 11:22:53.857 INFO  [StandardEngine] Starting Servlet engine: [Apache Tomcat/10.1.52]                                                                                                                                                                                                                                                                                            
2026-05-21 11:22:53.867 INFO  [[/]] Initializing Spring embedded WebApplicationContext                                                                                                                                                                                                                                                                                                     
2026-05-21 11:22:53.873  INFO [vcc-superx-data-service,,] 27385 --- [           main] o.s.b.a.e.web.EndpointLinksResolver      : Exposing 27 endpoint(s) beneath base path '/actuator'                                                                                                                                                                                                     
2026-05-21 11:22:53.894  INFO [vcc-superx-data-service,,] 27385 --- [           main] o.s.b.w.embedded.tomcat.TomcatWebServer  : Tomcat started on port 1293 (http) with context path ''                                                                                                                                                                                                   
2026-05-21 11:22:53.893 INFO  [Http11NioProtocol] Starting ProtocolHandler ["http-nio-1293"]2026-05-21 11:22:53.907  INFO [vcc-superx-data-service,,] 27385 --- [           main] c._.v.s.d.s.server.DataServiceBootstrap  : Started DataServiceBootstrap in 4.063 seconds (process running for 4.707)                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                             
2026-05-21 11:22:54.092 INFO  [[/]] Initializing Spring DispatcherServlet 'dispatcherServlet'2026-05-21 11:22:54.092  INFO [vcc-superx-data-service,,] 27385 --- [3)-192.168.1.69] o.s.web.servlet.DispatcherServlet        : Initializing Servlet 'dispatcherServlet'                                                                                                                     
2026-05-21 11:22:54.093  INFO [vcc-superx-data-service,,] 27385 --- [3)-192.168.1.69] o.s.web.servlet.DispatcherServlet        : Completed initialization in 1 ms no boot do data-service
```


Two patterns visible:
- Spring/Logback: <ts>  <LEVEL> [<app>,<trace>,<span>] <pid> --- [<thread>] <logger> : <msg>
- ArcadeDB native: <ts> <LEVEL> [<component>] <msg> — no PID, no thread, no MDC, no logger FQN, often glued to the next line because println is used without \n flushing.
  
**Impact**
  
- Operations: centralised log shippers (Loki, ELK, Datadog) parse one format; the ArcadeDB lines either fail to parse or get misclassified. Trace/span correlation in distributed tracing is broken for anything originating in ArcadeDB.
- Configuration: consumers cannot manage ArcadeDB log levels via the Logback/Log4j config they already maintain — they have to learn ArcadeDB's own log configuration on top.
- Readability: because LogManager writes directly to the same stdout stream that SLF4J's ConsoleAppender writes to, lines occasionally collide (see the …port 1292"]2026-05-21… example above) due to lack of locking between the two writers.
- MDC loss: Spring Cloud Sleuth / Micrometer Tracing inject traceId/spanId into the MDC. ArcadeDB log lines emitted inside the same request thread don't carry them, so request-scoped tracing breaks at the persistence boundary.

**Suggested fix**

Make com.arcadedb.log.LogManager delegate to SLF4J:
  
- Add org.slf4j:slf4j-api as a compile-time dependency in the relevant ArcadeDB modules.
- Replace direct writes with a per-class Logger logger = LoggerFactory.getLogger(...) (or one logger per component name currently used in the [X] field).
- Provide an SPI/setting to preserve the current behaviour for users running ArcadeDB standalone (so the server keeps its existing console output). Default delegates to SLF4J when an SLF4J binding is on the classpath; falls back to current behaviour otherwise — same pattern used by Netty, gRPC, Hibernate, etc.

If a full migration is too invasive, an interim step that solves the worst of the noise:
  
- Use SLF4J as the primary sink (so host apps can route ArcadeDB logs through Logback/Log4j) and document a single environment variable to disable the legacy stdout writer.
- At minimum, ensure every LogManager write ends in a newline and acquires a lock around System.out to stop the collision-with-Logback lines.

**Environment**

- ArcadeDB 26.4.2 (arcadedb-engine, arcadedb-network) embedded in a Spring Boot 3 service
- Host logging: SLF4J + Logback default Spring Boot configuration
- Symptoms reproduce on JDK 21 / Linux containers and locally on macOS

ruispereira:
Closes #4276 (opt-in first step).

## Summary

Adds an **opt-in** `Slf4jLogger` that routes ArcadeDB's logs through the **SLF4J facade**, so an application embedding ArcadeDB receives the engine's logs in whatever backend it already uses (Logback, Log4j2, reload4j, or `java.util.logging` via `slf4j-jdk14`) — with **no dependency exclusions**.

It is deliberately **non-disruptive**: the default logger is unchanged (`DefaultLogger`, `java.util.logging`, with the existing text/ANSI and JSON formatting), so standalone and current deployments behave exactly as before. This matches the plan agreed in #4276; SLF4J can be made the default in a follow-up if you'd prefer.

## How to enable

Startup, no code change:
```
-Darcadedb.log.impl=slf4j
```
or programmatically:
```java
LogManager.instance().setLogger(new Slf4jLogger());
```

## What changed

- **`engine/.../log/Slf4jLogger.java`** (new) — a `com.arcadedb.log.Logger` that delegates to `org.slf4j`:
  - maps `java.util.logging.Level` onto SLF4J's five levels by numeric weight (`SEVERE→ERROR`, `WARNING→WARN`, `INFO→INFO`, `FINE→DEBUG`, else `TRACE`), so custom/intermediate JUL levels also resolve;
  - preserves the call-site contract of `DefaultLogger`: logger name resolved from the requester (String / Class / instance / `com.arcadedb` fallback), `<context>` prefixing, and **printf-style (`%s`)** message formatting (not SLF4J `{}` placeholders), which is how ArcadeDB messages are written;
  - skips formatting entirely when the level is disabled;
  - exposes the per-request correlation carried by `LogManager` (`requestId`, `database`, `traceId`, `spanId`) to the backend via **MDC** under `arcadedb.*` keys, **saving and restoring** any pre-existing values so the host application's MDC is left untouched.
- **`engine/.../log/LogManager.java`** — selects the implementation from the new `arcadedb.log.impl` system property (default = `DefaultLogger`; `slf4j` = `Slf4jLogger`); falls back to `DefaultLogger` if construction fails, so a misconfiguration never silences logging.
- **`engine/pom.xml`** — adds `slf4j-api` only. The engine pins **no** backend.
- **`pom.xml`** — moves `slf4j-jdk14` to **`test`** scope. It was a compile-scope dependency in the parent `<dependencies>`, so it leaked transitively to consumers of every library module (forcing exclusions). At test scope, module tests still log through JUL while nothing is imposed on downstream consumers.
- **`examples/logging/`** (new) — ready-to-use `logback.xml` and `log4j2.xml` (console + JSON + rolling file, configurable log directory) and a `README.md` mapping every native logging feature (`arcadedb.server.logFormat`, `logIncludeTrace`, `logsDirectory`, per-package levels, correlation) to its Logback/Log4j2 equivalent.

`DefaultLogger`, `ArcadeDBServer`, `GlobalConfiguration` and the existing logging tests are **untouched**.

## Dependency hygiene

Embedding `arcadedb-engine` now pulls only `slf4j-api`; the embedder provides exactly one binding and needs no `<exclusions>`. To get plain JUL, add `slf4j-jdk14` and SLF4J forwards to it.

## Backward compatibility

Default behaviour is unchanged (JUL `DefaultLogger`), so there is no regression for standalone or existing embedders. SLF4J routing is entirely opt-in.

## Testing

- New `Slf4jLoggerTest` (12 tests): level mapping, printf substitution, `<context>` prefixing, logger-name resolution (String/Class/null), the fixed-arity overload, level gating, literal `%` without args, best-effort fallback when a format string doesn't match its args, throwable pass-through, and MDC publish/restore (including preserving a host value under an `arcadedb.*` key).
- All existing logging tests pass unchanged (`LoggerTest`, `DefaultLoggerLogDirTest`, `DefaultFormatterUnchangedTest`, `LogCorrelationContextTest`, `LogFormatterTraceTagTest`, `JsonLogFormatterTest`, `LogFormatterMessageFormatTest`) — 48 logging tests green in total.
- Full `arcadedb-engine` test suite: the only failure is a pre-existing one on `main` (`LSMTreeSortedBuildCrashTest.crashAfterAttachmentLeavesSchemaUnpublishedAndAllowsRetry`), reproduced identically on a clean checkout without this change — unrelated to logging.
- Verified end-to-end with Logback on the classpath (no `slf4j-jdk14`): logs are emitted by Logback with correct level mapping, `%s` substitution, MDC correlation populated during the call and cleared afterwards.

## Note

One small thing worth a maintainer's eye: the repo's `.gitignore` rule `**/log` also matches the source package `com/arcadedb/log/`, so the two new files there had to be `git add -f`'d. You may want to tighten that rule (e.g. anchor it to build output dirs) so new sources under that package aren't silently ignored.

### DIFF
--- a/ATTRIBUTIONS.md
+++ b/ATTRIBUTIONS.md
@@ -93,8 +93,8 @@ The following table lists runtime dependencies bundled with ArcadeDB distributio
 
 | Group ID | Artifact ID | Version | License | Homepage |
 |----------|-------------|---------|---------|----------|
-| org.slf4j | slf4j-api | 2.0.17 | MIT | https://www.slf4j.org/ |
-| org.slf4j | slf4j-jdk14 | 2.0.17 | MIT | https://www.slf4j.org/ |
+| org.slf4j | slf4j-api | 2.0.18 | MIT | https://www.slf4j.org/ |
+| org.slf4j | slf4j-jdk14 | 2.0.18 | MIT | https://www.slf4j.org/ |
 | ch.qos.logback | logback-classic | 1.5.27 | EPL 1.0 / LGPL 2.1 | https://logback.qos.ch/ |
 | ch.qos.logback | logback-core | 1.5.27 | EPL 1.0 / LGPL 2.1 | https://logback.qos.ch/ |
 | com.conversantmedia | disruptor | 1.2.21 | Apache 2.0 | https://github.com/conversant/disruptor |
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -84,5 +84,14 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+        <!-- Runtime SLF4J binding for the runnable console: routes third-party SLF4J logs
+             (dependencies) to java.util.logging. Declared here at runtime because the parent keeps
+             slf4j-jdk14 at test scope so it does not leak to consumers of the library modules. -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>${slf4j.version}</version>
+            <scope>runtime</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -119,6 +119,14 @@
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <!-- SLF4J facade only. ArcadeDB logs through slf4j-api and pins NO binding, so an embedding
+             application routes the engine's logs through whatever backend it already uses (Logback,
+             Log4j2, reload4j, or java.util.logging via slf4j-jdk14) with no dependency exclusions. -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
+        </dependency>
         <dependency>
             <groupId>com.conversantmedia</groupId>
             <artifactId>disruptor</artifactId>
--- a/engine/src/main/java/com/arcadedb/log/LogManager.java
+++ b/engine/src/main/java/com/arcadedb/log/LogManager.java
@@ -125,6 +125,14 @@ public class LogManager {
     logger = createLogger();
   }
 
+  /**
+   * Builds the {@link Logger} chosen by {@link #LOG_IMPL_PROPERTY}: {@link Slf4jLogger} for
+   * {@code slf4j}, otherwise the default {@link DefaultLogger}. Any failure constructing the chosen
+   * implementation (e.g. {@code slf4j-api} missing at runtime) is caught and falls back to
+   * {@link DefaultLogger}, so a logging misconfiguration can never prevent startup.
+   *
+   * @return the logger instance to install; never {@code null}
+   */
   private static Logger createLogger() {
     final String impl = System.getProperty(LOG_IMPL_PROPERTY, "").trim().toLowerCase(java.util.Locale.ROOT);
     try {
--- a/engine/src/main/java/com/arcadedb/log/LogManager.java
+++ b/engine/src/main/java/com/arcadedb/log/LogManager.java
@@ -26,6 +26,14 @@ import java.util.logging.Level;
  * @author Luca Garulli
  */
 public class LogManager {
+  /**
+   * System property selecting the {@link Logger} implementation, applied at startup without any code
+   * change. Unset (or any value other than {@code slf4j}) keeps the default {@link DefaultLogger}
+   * (java.util.logging); {@code slf4j} installs {@link Slf4jLogger}, routing the engine's logs
+   * through the SLF4J facade so an embedding application receives them in its own backend. The logger
+   * can also be swapped programmatically via {@link #setLogger(Logger)}.
+   */
+  public  static final String                        LOG_IMPL_PROPERTY    = "arcadedb.log.impl";
   private static final LogContext                    CONTEXT_INSTANCE     = new LogContext();
   private static final ThreadLocal<Correlation>      CORRELATION_INSTANCE = new ThreadLocal<>();
   private static final LogManager                    instance             = new LogManager();
@@ -111,15 +119,6 @@ public class LogManager {
     final Correlation c = CORRELATION_INSTANCE.get();
     return c == null ? null : c.spanId();
   }
-
-  /**
-   * System property selecting the {@link Logger} implementation, applied at startup without any code
-   * change. Unset (or any value other than {@code slf4j}) keeps the default {@link DefaultLogger}
-   * (java.util.logging); {@code slf4j} installs {@link Slf4jLogger}, routing the engine's logs
-   * through the SLF4J facade so an embedding application receives them in its own backend. The logger
-   * can also be swapped programmatically via {@link #setLogger(Logger)}.
-   */
-  public static final String LOG_IMPL_PROPERTY = "arcadedb.log.impl";
 
   protected LogManager() {
     logger = createLogger();
--- a/engine/src/main/java/com/arcadedb/log/LogManager.java
+++ b/engine/src/main/java/com/arcadedb/log/LogManager.java
@@ -112,8 +112,30 @@ public class LogManager {
     return c == null ? null : c.spanId();
   }
 
+  /**
+   * System property selecting the {@link Logger} implementation, applied at startup without any code
+   * change. Unset (or any value other than {@code slf4j}) keeps the default {@link DefaultLogger}
+   * (java.util.logging); {@code slf4j} installs {@link Slf4jLogger}, routing the engine's logs
+   * through the SLF4J facade so an embedding application receives them in its own backend. The logger
+   * can also be swapped programmatically via {@link #setLogger(Logger)}.
+   */
+  public static final String LOG_IMPL_PROPERTY = "arcadedb.log.impl";
+
   protected LogManager() {
-    logger = new DefaultLogger();
+    logger = createLogger();
+  }
+
+  private static Logger createLogger() {
+    final String impl = System.getProperty(LOG_IMPL_PROPERTY, "").trim().toLowerCase(java.util.Locale.ROOT);
+    try {
+      return "slf4j".equals(impl) ? new Slf4jLogger() : new DefaultLogger();
+    } catch (final Throwable t) {
+      // A logging-init problem must never take the process down: fall back to the dependency-free
+      // java.util.logging implementation.
+      System.err.println(
+          "ArcadeDB: cannot initialize logger impl '" + impl + "', falling back to java.util.logging. Cause: " + t);
+      return new DefaultLogger();
+    }
   }
 
   public static LogManager instance() {
--- a/engine/src/main/java/com/arcadedb/log/LogManager.java
+++ b/engine/src/main/java/com/arcadedb/log/LogManager.java
@@ -18,6 +18,7 @@
  */
 package com.arcadedb.log;
 
+import java.util.Locale;
 import java.util.logging.Level;
 
 /**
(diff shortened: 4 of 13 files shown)


ruispereira:
Thanks @gemini-code-assist — good catch. Removed the local `loggersCache` in `Slf4jLogger` and now delegate directly to `LoggerFactory.getLogger(name)` (dropped the now-unused `ConcurrentHashMap`/`ConcurrentMap` imports too). Pushed.

ruispereira:
@robfrank whenever you have a moment — this is ready for review. It implements the opt-in SLF4J-facade logging we discussed in #4276: default logger stays `java.util.logging` (no regression), `-Darcadedb.log.impl=slf4j` routes the engine's logs through the SLF4J facade so embedders get them in their own backend with no exclusions. The bot feedback (Gemini's local-cache note, Codacy) is already addressed, and the standalone distribution keeps its runtime binding. Happy to adjust anything — thanks!

ruispereira:
Heads-up on the red checks: they're all fork-PR permission/secret limitations, not test or build failures. Because this PR comes from a fork, the workflows run with a read-only `GITHUB_TOKEN` and no access to repo secrets, so the steps that *write back* to the PR or need a secret fail — while the actual work passes:

- **unit-tests** → `Run Unit Tests with Coverage` = success; only `Unit Tests Reporter` (publishes results, needs write) failed.
- **ha-integration-tests** → same, the `HA IT Tests Reporter` step.
- **Check License Compliance** → the check passed; only `Comment on PR (if applicable)` failed.
- **claude-review** → `Run Claude Code Review` needs the `ANTHROPIC_API_KEY` secret (unavailable to forks).
- **codecov/project** & **coverage-report** → need a token to upload/post.

They should go green when run with the usual permissions on your side. Happy to rebase or adjust anything if it helps.

robfrank:
I'll check it next week, It looks fine from the surface I need to check deeply and I'm mostly off line right now.

robfrank:
Thanks @ruispereira, this is a well-built contribution: the opt-in design is genuinely non-disruptive, the `Slf4jLogger` semantics match `DefaultLogger` closely (level gating before formatting, printf substitution, best-effort on format mismatch), and the MDC save/restore correctly puts back exactly the keys it set. The example configs and the parity table are a nice touch.

I reviewed it and pushed a follow-up commit (016306264) directly to the branch rather than sending you round another cycle. Here is what it changes and why.

## Two substantive fixes

**1. The binding was still leaking from `arcadedb-server`.**

The PR moves the parent's `slf4j-jdk14` to test scope, but `server/pom.xml` still declared it with no scope:

```
com.arcadedb:arcadedb-server:jar
+- org.slf4j:slf4j-api:jar:2.0.18:compile
\- org.slf4j:slf4j-jdk14:jar:2.0.18:compile
```

So any application embedding `arcadedb-server` (the common case when you want the HTTP API) still inherited the JUL binding and still needed `<exclusions>`, which is exactly what #4276 is about. Only `arcadedb-engine` embedders got the clean tree.

I removed both SLF4J declarations from `server`: the `slf4j-api` facade already arrives transitively from `arcadedb-engine`, and `git grep` confirms `Slf4jLogger.java` is the only `src/main` file in the whole repo that references `org.slf4j`. The tree is now `slf4j-api:compile` + `slf4j-jdk14:test`, and your `console`/`package` runtime declarations become the load-bearing ones. Verified that all four distribution flavours still ship `slf4j-api-2.0.18.jar` and `slf4j-jdk14-2.0.18.jar` in `lib/`.

I also fixed the comment in the parent POM, which pointed at `server/pom.xml` rather than `console`/`package`.

**2. The documented fallback in `createLogger()` never fired.**

The javadoc claimed the `catch (Throwable)` covered "`slf4j-api` missing at runtime". It did not: `Slf4jLogger` touched `LoggerFactory`/`MDC` only in method bodies, so class loading and construction succeeded and the error escaped later. Running with `engine/target/classes` alone on the classpath:

```
step1: LogManager.instance() ...
step1 OK -> logger installed
step2: first log call ...
step2 THREW: java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
```

No fallback message, so the catch block never ran, and the `NoClassDefFoundError` would surface from whatever engine code happened to log first. The constructor now calls `LoggerFactory.getILoggerFactory()`, which makes the probe real. Same run now prints the diagnostic, installs `DefaultLogger`, and logs normally. Mostly theoretical now that engine hard-depends on `slf4j-api`, but shaded or minimized deployments can strip it.

## Smaller fixes

- **Silent fallback on a typo.** `-Darcadedb.log.impl=slf4` quietly gave you JUL with no diagnostic. It now reports the unrecognized value and lists the supported ones. `createLogger()` is package-private so the selection is testable.
- **No shutdown bypass.** `DefaultLogger` has the `shuttingDown` -> `System.err` path from #2813, but the SLF4J path had no equivalent, and Logback/Log4j2 register their own shutdown hooks that stop the context. `write()` now mirrors it for INFO and above.
- **Packaged-server gotcha.** `Slf4jLogger` never calls `DefaultLogger.init()`, so `config/arcadedb-log.properties` is never read. Since the distribution bundles `slf4j-jdk14`, setting `-Darcadedb.log.impl=slf4j` on a packaged server still logs through JUL, but the `FileHandler` and formatter from that file are not installed, so `log/arcadedb.log.*` silently stops being written. Added a README section.
- `ATTRIBUTIONS.md` slf4j rows synced to the pinned 2.0.18, and fully-qualified names replaced with imports per the project style.

## Tests

Added `LogManagerLoggerSelectionTest` (4 tests covering unset / `default` / `slf4j` / unrecognized) and two shutdown-bypass tests. 50 logging tests green, and a full reactor `mvn -DskipTests install` passes, so the parent scope change breaks no module.

## Not changed, for a maintainer decision

- **`examples/logging/` placement.** New top-level directory, outside the Maven reactor and not shipped in the distribution. `docs/` or somewhere under `package/` may be a better home. Happy to leave it where it is.
- **Your `.gitignore` note is correct** and worth acting on. `git check-ignore -v engine/src/main/java/com/arcadedb/log/Foo.java` does report `.gitignore:390 **/log`, and the new test file needed `git add -f`. Fixing it properly means negating the directory itself (`!**/src/main/java/**/log`), which is a repo-wide rule change, so I would rather do that in its own PR than fold it in here.

Thanks again for the care that went into this one, particularly the parity table and the MDC restore semantics.

PR #5370, opened 2026-07-21
CHECK DATABASE fix mode rebuilds unreadable edge lists; ORDER BY on graph-function expressions actually sorts

### LINKED ISSUE #5371: CHECK DATABASE with fix mode destroys data when an edge-list chunk is unreadable
While building a recovery path for a database with damaged edge-list chunks (left behind by replication incidents on older builds), we found that `CHECK DATABASE` in fix mode did not repair a vertex with an unreadable edge linked-list chunk - it destroyed data. Three distinct failure modes, all reproduced with tests:

### 1. Fix mode deleted the vertex itself
When a mid-chain chunk of a vertex's edge list is unreadable, the walk failure escapes through `Iterator.hasNext()` in the loop condition of `GraphDatabaseChecker.checkIncomingEdges`/`checkOutgoingEdges` - which sits **outside** the per-entry try/catch. The exception bubbles to the `scanType` callback, which flags the **whole vertex** as corrupted, and the fix pass then deletes the vertex record outright. A hub vertex with millions of intact edges could be deleted over one bad 1KB chunk.

### 2. Fix mode deleted valid edge records
When a vertex with a broken chain is scanned **after** its neighbours (plain bucket-order dependence), each neighbour's bidirectional back-reference probe (`isConnectedTo`) throws while walking the broken chain. The catch arm blamed the **edge record** being verified ("edge not found"), marked it corrupted, and the fix pass deleted it - permanently losing valid edges, one per neighbour scanned before the broken vertex.

### 3. Silent amputation instead of repair
When the head chunk was unreadable, fix mode simply nulled the vertex's head-chunk pointer. The exception disappears, but the vertex silently loses its entire adjacency (`both().size()` returns 0) and the surviving edge records become permanent orphans.

### Fix
Every edge record stores its own out/in vertex RIDs, so the adjacency is fully recoverable. PR #5370 makes all three cases **rebuild** the edge list from the surviving edge records through the existing `reconnectEdges` machinery: chain-walk failures and neighbour-probe failures register the vertex for rebuild instead of deleting anything, the rebuild scan tolerates unreadable edge records, and IN lists are only rebuilt for bidirectional edge types. Plain `CHECK` (without fix) is also now truly read-only in the points-to-another-list arm, which previously mutated the database. Covered for both the classic and the striped super-node layouts by `GraphDatabaseCheckerChainRebuildTest`.

### Related discovery
The same investigation surfaced that `ORDER BY` on a graph-function expression (e.g. `ORDER BY both().size() DESC`) never sorted at all - the AST builder collapsed the expression into a property alias holding its literal text, so every sort key was null and rows came back in scan order. Also fixed in PR #5370 with `OrderByGraphFunctionTest`.

### Known limitation
After a rebuild the old orphaned chunks / stripe directory remain on disk as unreachable garbage (same as the previous amputation behaviour); reclaiming them would need a vacuum-style follow-up.

lvca:
Fixes #5371.

Two related engine fixes that came out of the same customer-recovery investigation (broken edge chains after replication incidents on older builds).

## 1. CHECK DATABASE fix mode rebuilds unreadable edge lists from the surviving edge records

An unreadable chunk in a vertex's edge linked list previously caused, depending on where the walk failed:

- **the vertex itself to be deleted** by fix mode: the failure escaped through `Iterator.hasNext()` in the walk loop (outside the try/catch), so the scan callback flagged the whole vertex as corrupted;
- **valid edge records to be deleted** when a healthy vertex was scanned before its broken neighbour: the bidirectional back-reference probe (`isConnectedTo`) threw on the neighbour's broken chain and the edge record was blamed for it;
- **silent amputation** of the whole adjacency on head-chunk loss (head pointer nulled, `both().size()` then returned 0).

Each edge record stores its own out/in vertex RIDs, so the adjacency is now **rebuilt** through the existing `reconnectEdges` machinery instead: chain-walk failures and far-vertex probe failures register the vertex for a rebuild, the head pointer is reset, and every surviving edge reconnects. The rebuild scan tolerates unreadable edge records (skipped with a warning), skips edges already marked corrupted (they are deleted right after), and only rebuilds IN lists for bidirectional edge types. Plain CHECK (no fix) no longer mutates the database in the points-to-another-list arm. Works for both the classic and the striped super-node layouts.

Tests: `GraphDatabaseCheckerChainRebuildTest` - mid-chain corruption, head-chunk corruption, striped super-node generation-0 corruption, and the scan-order case where the neighbour probe fires first.

## 2. ORDER BY on a graph-function expression silently returned unsorted rows

`SELECT name, both().size() AS degree FROM Node ORDER BY both().size() DESC` returned rows in scan order on a **healthy** database: the AST builder collapsed the ORDER BY expression into a property alias holding its literal text (`"both().size()"`), so every sort key resolved to null. Queries only looked sorted when the highest-degree vertex happened to be the first record in the bucket.

The AST builder now keeps the full expression for non-identifier bases with modifier chains; both the ORDER BY comparator and the planner's hidden pre-projection already evaluate `item.expression` correctly.

Tests: `OrderByGraphFunctionTest` - with/without matching projection, no projection, ascending/descending, with hub bucket positions deliberately chosen against the expected order so a no-op sort can never pass by accident.

## Verification

All new tests green plus regression sweeps: CheckDatabaseTest (9), CheckDatabaseExtendedTest (12), GraphDatabaseCheckerDiagnosticsTest (3), BasicGraphTest (19), SuperNodeStripingTest (17), EdgeLinkedListRemoveAllSegmentsTest (2), SelectStatementExecutionTest (159), MatchStatementExecutionTest (93), OrderByStepTest (13), LimitSkipStepTest (10), SubQueryStepTest (10).

Known limitation (unchanged from the previous amputation behaviour): the old orphaned chunks / stripe directory remain on disk as unreachable garbage after a rebuild.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
@@ -24,7 +24,6 @@ import com.arcadedb.database.Record;
 import com.arcadedb.engine.Bucket;
 import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.log.LogManager;
-import com.arcadedb.query.select.SelectIterator;
 import com.arcadedb.schema.DocumentType;
 import com.arcadedb.schema.EdgeType;
 import com.arcadedb.utility.Pair;
@@ -91,10 +90,12 @@ public class GraphDatabaseChecker {
           final RID vertexIdentity = vertex.getIdentity();
 
           vertex = checkOutgoingEdges(fix, vertex, warnings, totalWarnings, maxWarnings, vertexIdentity, invalidLinks,
-              corruptedRecords, totalCorrupted, maxCorrupted, reconnectOutEdges, missingReferences, missingReferenceErrors);
+              corruptedRecords, totalCorrupted, maxCorrupted, reconnectOutEdges, reconnectInEdges, missingReferences,
+              missingReferenceErrors);
 
           checkIncomingEdges(fix, vertex, warnings, totalWarnings, maxWarnings, vertexIdentity, invalidLinks,
-              corruptedRecords, totalCorrupted, maxCorrupted, reconnectInEdges, missingReferences, missingReferenceErrors);
+              corruptedRecords, totalCorrupted, maxCorrupted, reconnectInEdges, reconnectOutEdges, missingReferences,
+              missingReferenceErrors);
 
         } catch (final Throwable e) {
           addWarning(warnings, totalWarnings, maxWarnings,
@@ -112,7 +113,7 @@ public class GraphDatabaseChecker {
 
       if (fix) {
         if (!reconnectOutEdges.isEmpty() || !reconnectInEdges.isEmpty())
-          reconnectEdges(reconnectOutEdges, reconnectInEdges, warnings, totalWarnings, maxWarnings, stats);
+          reconnectEdges(reconnectOutEdges, reconnectInEdges, corruptedRecords, warnings, totalWarnings, maxWarnings, stats);
 
         for (final RID rid : corruptedRecords) {
           if (rid == null)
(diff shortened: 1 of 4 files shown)


lvca:
Thanks for the thorough review. Addressed in 92d2326f6:

- **(2) Full edge scan cost**: documented in `reconnectEdges`'s new Javadoc (fix-mode only, same cost as the pre-existing reconnect path).
- **(3) Duplicated warning lambda**: factored into `warnUnreadableEdgeDuringRebuild`.
- **(4) Dangling far endpoint**: noted in the same Javadoc as a known, pre-existing behaviour that the `checkEdges` pass reports.

On **(1) `@Tag("slow")`**: I measured instead of tagging - `promotedSuperNodeChainCorruptionRebuildsAdjacency` runs in 0.19s and the whole class in under 1s (the 2,000 edges are batched into 4 transactions). The sibling super-node tests carry the tag on specific multi-second methods, not the class. Tagging this one would drop CI coverage of the striped-layout repair path for no runtime saving, so I left it untagged.

PR #5373, opened 2026-07-21
feat(#5372): step-by-step progress reporting for CHECK DATABASE (engine, HTTP, console, Studio)

### LINKED ISSUE #5372: Progress reporting for long-running maintenance operations (CHECK DATABASE first)
`CHECK DATABASE` (especially with the fix mode repairs from #5371 / PR #5370) can run for a long time on large databases, with no feedback until it returns. The console and Studio just wait. The same applies to other long maintenance operations (`COMPACT INDEX`, index rebuilds, backups, imports).

CHECK DATABASE's progress is knowable upfront: the step list is deterministic (per vertex type: record scan + adjacency walk; in fix mode: edge-list rebuild + corrupted-record deletion; per edge type: the edge check) and each step's total is a cheap `bucket.count()`. So we can report real "step 3/7 - 42%" progress, not just a spinner.

## Design

**1. Engine progress callback.** A minimal interface (`onProgress(stepName, done, total)`) accepted by `GraphDatabaseChecker`/`DatabaseChecker`, invoked only when the integer percentage changes so the hot path stays allocation-free.

**2. Server-neutral operation registry.** A JVM-wide registry of running operations (`opId, database, operation, current step, stepIndex/totalSteps, done/total, startedAt`), updated lock-free through volatile fields by the callback. `CHECK DATABASE` registers on start and unregisters on completion. Deliberately generic so `COMPACT INDEX`, index rebuilds, backups and imports can publish into it later - this is the "what is this server doing right now" primitive.

**3. Surfacing.**
- **HTTP**: `GET /api/v1/progress/{database}` returning the running operations for the database as JSON (same auth model as the other database endpoints).
- **Console**: remote connections poll the endpoint from a side thread while the synchronous command runs and redraw a single-line text progress bar; embedded connections print directly from the callback.
- **Studio**: while a `CHECK DATABASE` command is in flight, poll the endpoint every ~2s and render a progress bar with the current step and percentage.
- Server log: one INFO line per step transition, percentage ticks at FINE, so operators tailing logs see progress without flooding.

The registry is intentionally server-local (not replicated): `CHECK DATABASE` runs per node and the progress shown is the node you asked.

lvca:
Closes #5372.

> **Stacked on #5370** (base is `check-rebuild-edge-chains`, not `main`) because the progress wiring touches the rebuilt `GraphDatabaseChecker`. Merge #5370 first; GitHub will retarget this PR to `main` automatically and the diff here stays limited to the progress feature.

Long-running maintenance operations gave no feedback until they returned. This adds live "step 3/7 - 42%" progress for `CHECK DATABASE`, built on a generic primitive that `COMPACT INDEX`, index rebuilds, backups and imports can adopt later.

## Engine
- `ProgressCallback` (`onProgress(stepName, stepIndex, totalSteps, done, total)`), designed to be allocation-free on the hot path.
- `OperationProgress` / `OperationProgressRegistry`: a JVM-wide, lock-free registry of running operations. Producers update volatile primitives; readers (HTTP endpoint, pollers) take weakly-consistent snapshots. Deliberately process-local: in HA each node reports what it is doing.
- `DatabaseChecker.check()` computes the full step plan upfront (per-type edge/vertex/document checks + buckets + external properties + indexes + fix-rebuild + compress), so every emission carries a stable `stepIndex/totalSteps`. Per-record ticks are throttled to integer-percentage changes and clamped against count drift.
- `GraphDatabaseChecker` reports within-step progress: the two scan passes count against `2 x countType`, and the fix-mode edge-list rebuild is an explicit unknown-total sub-phase.
- `CHECK DATABASE` registers the operation and retires it in a `finally`, so a failed check never leaks a registry entry.

## Server
- `GET /api/v1/progress/{database}` returns the running operations as JSON. It reads only the in-memory snapshot - no database access, no transaction - so it is safe to poll at any frequency and cannot interfere with the operation being watched. Enforces the per-database authorization model (403 otherwise, 400 on missing parameter).

## Network
- `RemoteDatabase.getProgress()` exposes the endpoint to any remote client.

## Console
- A `CHECK DATABASE` command renders a live single-line bar refreshed every 500ms, e.g. `check database fix [step 3/9] Checking vertices 'Account' |========            | 42%` - local registry when embedded, progress endpoint when remote. Strictly best-effort: a polling failure stops the rendering, never the command; silent in batch/embedded-output modes.

## Studio
- While a `CHECK DATABASE` command is in flight, the query panel polls the endpoint every second and shows a Bootstrap progress bar with the current step and percentage next to the execute spinner, removed when the command returns.

## Verification
- Engine: `OperationProgressRegistryTest` (2), `CheckDatabaseProgressTest` (4: step ordering/bounds/percentages, per-type steps reach 100%, fix-mode adds the rebuild step, registry retirement) + regression: `CheckDatabaseTest` (9), `CheckDatabaseExtendedTest` (12), `GraphDatabaseCheckerChainRebuildTest` (4), `GraphDatabaseCheckerDiagnosticsTest` (3) - all green.
- Server: `GetProgressHandlerTest` (4: result shape, empty case, 403, 400) - green.
- Console: `ConsoleTest` (30), `ConsoleBatchTest` (7) - green. Network compiles; Studio JS passes syntax check.

https://claude.ai/code/session_01G5y8EbmkvSpFBbeDtTu33m

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -29,6 +29,9 @@ import com.arcadedb.database.Document;
 import com.arcadedb.database.TransactionContext;
 import com.arcadedb.database.async.AsyncResultsetCallback;
 import com.arcadedb.engine.ComponentFile;
+import com.arcadedb.engine.OperationProgress;
+import com.arcadedb.engine.OperationProgressRegistry;
+import com.arcadedb.serializer.json.JSONObject;
 import com.arcadedb.graph.Edge;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.integration.misc.IntegrationUtils;
@@ -629,6 +632,9 @@ public class Console {
         }
       });
     } else {
+      // LONG-RUNNING MAINTENANCE COMMANDS (issue #5372): render a live progress line while the synchronous
+      // command runs, polling the local registry (embedded) or the server's progress endpoint (remote).
+      final Thread progressMonitor = startProgressMonitor(line);
       try {
         resultSet = databaseProxy.command(language, line);
       } catch (Exception e) {
@@ -638,6 +644,8 @@ public class Console {
         else
           outputError(e);
         return;
+      } finally {
+        stopProgressMonitor(progressMonitor);
       }
     }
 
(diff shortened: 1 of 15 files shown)


lvca:
Thanks for the review. Rebased onto main after #5370 merged (the stack collapsed into a756898be), review items addressed in 2a7143639:

- **(1) Auth helper**: `GetProgressHandler` now delegates to `checkAuthorizationOnDatabase()`; the 400/403 semantics flow through the base handler's catch chain and the tests assert exactly that.
- **(3) Failure path**: replaced the placeholder test with a real one - a package-private checker seam lets `CheckDatabaseStatementProgressTest` inject a checker that throws mid-run and assert the registry entry is retired anyway.
- **(5) Console monitor**: added the volatile stop flag (set before interrupt, re-checked right before each print).
- **(2)/(6)**: documented oldest-operation-only rendering as a deliberate limitation in both console and Studio, and noted `countType`'s negligible cost at the call site.
- **(4)/(7)**: left as documented behaviour, per the review.

All suites re-run green: engine 19 (incl. the new failure-path test), server 4, console 37, Studio JS syntax-checked.

lvca:
Thanks - this review crossed with 2a7143639, which already addressed its main item:

- **(1) Failure-path test**: fixed in 2a7143639 - the misleading `sqlStatementRetiresTheOperationOnFailureToo` is gone, replaced by `CheckDatabaseStatementProgressTest`, which injects (via a package-private checker seam) a checker that throws mid-run and asserts the registry entry is retired anyway.

The remaining suggestions are addressed in 010081248:

- **(2) Torn snapshot**: `toJSON()` now reads each volatile once, so `done`/`total`/`percentage` in one JSON payload are mutually consistent (cross-field weak consistency stays the documented contract).
- **(3) Name-keyed registry**: documented in the registry Javadoc as an accepted limitation (the server never hosts two same-named databases), with identity-based keying as the noted escape hatch.
- **(4) Studio polling**: a `.fail` handler now stops the poller so an unavailable endpoint (older server, auth change) doesn't get hammered once per second for the rest of the command.
- **(5) + nits**: left as documented behaviour - the console monitor indeed relies on the command being synchronous, and the 100% -> unknown transition during the fix rebuild is the intended sub-phase display.

Engine 6 progress tests + Studio JS syntax re-verified green.

PR #5379, opened 2026-07-21
feat(#5375): CHECK DATABASE FIX reclaims orphaned edge-list segments

### LINKED ISSUE #5375: CHECK DATABASE FIX should reclaim orphaned edge-list segments
Follow-up of #5371. When fix mode rebuilds a vertex's edge list from the surviving edge records (PR #5370), the old chunks of the broken chain - and, for promoted super-nodes, the old stripe directory and stripe chains - remain on disk as unreachable garbage. The same was true of the previous amputation behaviour, so nothing regressed, but the repair story is incomplete: on a badly damaged database the dead segments can add up.

Proposal: fold the reclaim into the existing `CHECK DATABASE FIX` pass (no separate vacuum command):
1. while walking the vertices (which fix mode already does), collect the set of REACHABLE edge-segment RIDs: every chunk of every vertex's OUT/IN chain, including stripe directories, stripe chain chunks and the generation-0 chain of promoted vertices;
2. scan the dedicated edge-list buckets and delete any edge-segment/stripe-directory record that is not in the reachable set.

The reachable set can use the same primitive-long-set approach as the external-property orphan scan (`LongHashSet` keyed by position, per bucket) to keep the memory footprint low. The reclaim becomes one more step of the progress plan introduced by #5372.

lvca:
Closes #5375. Follow-up of #5371/PR #5370.

## Orphan reclaim

A full-scope `CHECK DATABASE FIX` now adds a "Reclaiming orphaned edge segments" step (visible in the #5372 progress display) after the vertex checks:

1. walks every vertex's OUT/IN chains and marks every reachable segment - classic chunks, stripe directories, stripe chains and the generation-0 chains of promoted super-nodes, with a cycle guard for corrupt chains - into per-bucket primitive `LongHashSet`s (same footprint approach as the external-property orphan scan);
2. scans the dedicated edge-list buckets (`*_out_edges`, `*_in_edges`, `*_sn_stripe_*`) and deletes every segment not marked reachable.

The reclaim runs only when no type/bucket filter is set: a partial walk would misclassify the unwalked vertices' segments as orphans. New stats: `orphanedEdgeSegments` / `orphanedEdgeSegmentsReclaimed`. Like the rest of fix mode it assumes no concurrent writers.

## Data-loss bug found by the new tests

Exercising the FULL `check()` with fix over a broken chain (the earlier rebuild tests drove `checkVertices` directly) exposed a hole in the merged #5370 fix: `checkEdges` runs BEFORE the vertex checks, and its back-reference probe (`containsEdge` over the target vertex's edge list) blamed a broken CHAIN on the VERTEX record - fix mode then deleted the vertex before the rebuild could ever run. The probe now separates "the vertex record cannot be loaded" (vertex genuinely corrupted) from "the vertex's edge list cannot be walked" (warned once per vertex, left to the vertex check to rebuild, nothing flagged corrupted).

## Tests

`GraphDatabaseCheckerOrphanReclaimTest`:
- classic chain: orphans reclaimed after a rebuild, adjacency intact, follow-up check clean (slot reuse by `LocalBucket` accounted for: an old RID may legitimately live on as part of the rebuilt chain);
- striped super-node: old directory + stripe chains reclaimed;
- **healthy-database safety test**: the reclaim deletes nothing on a clean database with both layouts - the guard against an incomplete reachability walk destroying live data;
- filtered check: no reclaim under a TYPE filter.

Regression (68 tests green): chain-rebuild, checker diagnostics, CheckDatabaseTest/Extended, progress suites (fix-mode step plan updated to include the new step), SuperNodeStripingTest, BasicGraphTest.

https://claude.ai/code/session_01G5y8EbmkvSpFBbeDtTu33m

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -110,15 +110,33 @@ public class DatabaseChecker {
         documentTypes.add(type);
     }
 
+    // The orphaned-segment reclaim (issue #5375) requires walking EVERY vertex to build the reachable set, so
+    // it only runs on a full-scope fix: under a type/bucket filter the unwalked vertices' segments would be
+    // misclassified as orphans.
+    final boolean reclaimOrphanedSegments =
+        fix && (types == null || types.isEmpty()) && (buckets == null || buckets.isEmpty());
+
     currentStep = 0;
     totalSteps = edgeTypes.size() + vertexTypes.size() + documentTypes.size() // per-type checks
         + 3 // buckets + external properties + indexes
+        + (reclaimOrphanedSegments ? 1 : 0)
         + (fix ? 1 : 0) // rebuild affected indexes
         + (compress ? 1 : 0);
 
     checkEdges(edgeTypes);
 
     checkVertices(vertexTypes);
+
+    if (reclaimOrphanedSegments) {
+      // AFTER checkVertices: the chain rebuilds have already re-attached every recoverable segment, so what
+      // remains unreachable in the edge-list buckets is genuine garbage.
+      ++currentStep;
+      final Map<String, Object> stats = new GraphDatabaseChecker(database)
+          .setProgress(progressCallback, "Reclaiming orphaned edge segments", currentStep, totalSteps)
+          .reclaimOrphanedEdgeSegments(verboseLevel, maxWarnings);
+      updateStats(stats);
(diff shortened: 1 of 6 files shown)


lvca:
Review addressed in 06bfd8c1f: added the single-transaction/unbounded-list scaling note to the reclaim Javadoc (item 1), and a focused regression test `checkEdgesDoesNotBlameVertexForUnreadableEdgeList` that runs ONLY checkEdges with fix against a broken chain and asserts nothing is flagged corrupted, the warning fires exactly once, and the hub survives (item 3). Item 2 acknowledged as inherent. All 5 tests green.

PR #5380, opened 2026-07-21
feat(#5376): REBUILD INDEX, COMPACT INDEX, BACKUP DATABASE and IMPORT DATABASE publish live progress

### LINKED ISSUE #5376: Publish progress for COMPACT INDEX, REBUILD INDEX, BACKUP DATABASE and IMPORT DATABASE
Follow-up of #5372, which introduced `ProgressCallback` + `OperationProgressRegistry` + `GET /api/v1/progress/{database}` and wired `CHECK DATABASE` to them (console and Studio render a live progress bar).

Adopt the same primitive in the other long-running maintenance statements so they get the same live progress for free:
- `COMPACT INDEX`: per-compaction progress (pages/entries processed);
- `REBUILD INDEX`: indexes rebuilt as steps, records indexed as within-step progress;
- `BACKUP DATABASE`: files archived as progress;
- `IMPORT DATABASE`: records imported (total unknown, count-only progress).

Each statement registers an `OperationProgress` on start and retires it in a finally, exactly like `CheckDatabaseStatement`. The console and Studio pollers then only need their command matcher extended beyond `check database`.

lvca:
Closes #5376. Follow-up of #5372.

Adopts the progress primitive in the remaining long-running maintenance statements:

- **REBUILD INDEX**: one step per target index (`Rebuilding index 'x'`, step i/N) with per-record within-step progress against the type's record count, throttled to one registry update every 1024 records so the build hot loop is untouched.
- **COMPACT INDEX**: one step per compacted index.
- **BACKUP DATABASE / IMPORT DATABASE**: coarse operation-level progress. Both run behind the integration module's reflective boundary (and the import's record total is unknown upfront), so this pass publishes the running operation without per-record detail - still enough for the console/Studio to show what is running and for how long. Finer-grained progress inside the integration classes can be a later refinement.

Every statement registers in `OperationProgressRegistry` and retires in a `finally`, success or failure, like `CHECK DATABASE`. The console and Studio monitors now match all five commands (`check database`, `rebuild index`, `compact index`, `backup database`, `import database`).

## Tests
`MaintenanceOperationsProgressTest`: a background sampler observes the registry while `REBUILD INDEX *` runs over 100k records (operation + step names asserted), retire-on-completion for compact, and retire-on-failure for a rebuild of a non-existent index. Regression: `RebuildIndexPartitionValidationTest`, `CompactIndexStatementTest(+Parser)`, `FullTextRebuildOverExistingTest` all green; console compiles; Studio JS syntax-checked.

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -817,7 +817,8 @@ public class Console {
   private Thread startProgressMonitor(final String line) {
     if (output != null || verboseLevel < 2)
       return null;
-    final String normalized = line.trim().toLowerCase(Locale.ENGLISH);
+    // Collapse whitespace runs so `REBUILD   INDEX` matches too, consistently with the Studio matcher.
+    final String normalized = line.trim().toLowerCase(Locale.ENGLISH).replaceAll("\\s+", " ");
     boolean monitored = false;
     for (final String command : PROGRESS_MONITORED_COMMANDS)
       if (normalized.startsWith(command)) {
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -810,10 +810,21 @@ public class Console {
    * not applicable (not a monitored command, or the output is redirected to an embedding application).
    * Polling is best-effort: any failure silently stops the rendering, never the command.
    */
+  /** The maintenance commands that publish live progress in the operation registry (issues #5372, #5376). */
+  private static final String[] PROGRESS_MONITORED_COMMANDS = { "check database", "rebuild index", "compact index",
+      "backup database", "import database" };
+
   private Thread startProgressMonitor(final String line) {
     if (output != null || verboseLevel < 2)
       return null;
-    if (!line.trim().toLowerCase(Locale.ENGLISH).startsWith("check database"))
+    final String normalized = line.trim().toLowerCase(Locale.ENGLISH);
+    boolean monitored = false;
+    for (final String command : PROGRESS_MONITORED_COMMANDS)
+      if (normalized.startsWith(command)) {
+        monitored = true;
+        break;
+      }
+    if (!monitored)
       return null;
 
     progressMonitorStopped = false;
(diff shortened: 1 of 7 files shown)


lvca:
Review addressed in 930c10d22: (1) `REBUILD INDEX *` per-bucket sub-indexes now size their step total from the associated bucket's count, so the per-step percentage is accurate instead of capping at ~100/N%; (2) the console field moved above the method Javadoc so the doc block stays attached to `startProgressMonitor`; (3) import order fixed; (4) `@Tag("slow")` on the 100k rebuild test method; (5) the sampler sleeps 1ms instead of busy-spinning. Tests re-run green, console compiles.

lvca:
Second round addressed in 72dab5db6:

- **TypeIndex named rebuild**: the progress callback now increments its OWN per-step counter instead of publishing the callback's `totalIndexed` argument, which (as the review correctly traced) resets at every bucket boundary inside a `TypeIndex` build. No sawtooth, `done` accumulates across the sub-builds against the whole-type total, and the comment no longer overclaims.
- **Completion emission**: each step now lands at 100% when its build finishes, which also fixes the flagged minor (indexes under 1024 records showing 0% throughout).
- **Backup/import retire test**: `backupAndImportRetireProgressOnFailureToo` - on the engine test classpath the integration module is absent, so both commands register and then fail, exercising exactly the retire-in-finally contract for the two reflective paths.

All suites re-run green (MaintenanceOperationsProgressTest now 4 tests, rebuild/full-text regressions unchanged).

lvca:
Thanks - took the two easy ones from the LGTM round in ba1b13cb0:

- **(1) Polymorphic count**: non-polymorphic is actually the correct choice here - a named TypeIndex build scans only the declaring type's own buckets (subtypes inheriting the index carry their own sub-indexes, rebuilt as their own targets), so a polymorphic count would OVERcount the denominator. Made the choice explicit with a comment as suggested.
- **(2) Console matcher**: now collapses whitespace runs before matching, consistent with the Studio regex - `REBUILD   INDEX` renders progress in both.
- **(3)**: acknowledged - the loose exception assertion is intentional for a retire-contract test; if the integration module ever lands on the engine test classpath the test still asserts the same contract, just through a different failure.

Tests re-run green, console compiles.

Issue #5279, opened 2026-07-14
Is ArcadeDB ready for a real multi-user environment? The problem of ConcurrentModificationExceptions


mdre:
ArcadeDB use ConcurrentModificationException in a weird way. It throw that exception when a page is modified by somebody else, but who care about pages?
ConcurrentModificationException in all database represent that the record I need to commit was modified by somebody else, so I should decide what to do, overwrite the modified record o ask the user. That's the deal.
But with pages that is not true. The database throw that exception even when the data is ok and there no conflict and it seems there is no way to resolve what happens. A blind retry is not an option because I could overwrite something important, and ask the user in a heavy concurrent environment is not practical because we never end a simple save.
We have about 1254 user so every bucket change at every second. With the average creation of 873 case per morning day, every new page is modified by someone else, and we have more than a half of millon active cases that are updated at the same time.
CME should be launched only if the involved record are modified like was in OrientDB or any other database.
Actually in all our concurrent test it's impossible to save. A dead-lock overcome when you overpass the bucket size. 

This is a simple test:

```Java
    private void testPageConcurrentModificationException() {
        ArrayList<RemoteDatabase> alTx = new ArrayList<>();
        Set<String> concurrentPage = new HashSet<>();
        boolean concurrentDetected = false;
        int concurrentIntent = 10;
        for (int i = 0; i < concurrentIntent; i++) {
            RemoteDatabase tx = getDBTx();
            tx.begin();
            alTx.add(tx);
            
            MutableVertex svt1 = tx.newVertex("SimpleVertexEx");
            svt1.set("svex", "concurrent test"+i);
            svt1.save();
            
            System.out.println(""+i+": "+svt1.getIdentity());
        }
        
        // Throw the ConcurrentModificationException
        for (RemoteDatabase tx : alTx) {
            tx.commit();
        }
        
    }
```

In this case, only the first save and the other get a CME eternally. In a real case, it hard to track what vertex/edge are involved to force a reload in each one. The database should track that.
I you have more than one vertex involved it get worst. You get dead-lock because there are multiple way to modify different vertex/edge so nobody save.

lvca:
@mdre what's the operation that causes the concurrency exception? Adding an edge to a node (super-node)? Or are you actually updating the properties of the same vertex/edge?

mdre:
[...]

PR #5385, opened 2026-07-22
feat(#5381): disjoint-slot commit merge for page-level false conflicts

### LINKED ISSUE #5381: Page-level ConcurrentModificationException on bulk edge CREATE_UPDATE to a shared single-bucket type under concurrent load (remote HTTP, explicit tx) — reinforces #5279
### Summary

This reinforces #5279 with a concrete, reproducible-under-load case.

Under concurrent write load, a bulk edge `CREATE_UPDATE` (~3,900 edges in one explicit transaction, committed over the remote HTTP client) intermittently fails at commit with a **page-level** `ConcurrentModificationException` — even though the records being written are logically unrelated. It is a *false* conflict in the sense of #5279: no real record overlap, just writers colliding on the same page under concurrency.

Run in isolation the same operation always succeeds; it only fails when multiple independent clients write concurrently.

### Environment

- ArcadeDB **26.8.1-SNAPSHOT**
- 3-node HA cluster (writes serialize on the leader)
- Java remote client (`RemoteDatabase`), explicit `begin` / multi-batch / `commit`
- Edges addressed directly by record id (`g.E(<rid>)`) — no `outE` traversal, so this is *not* one of the super-node edge-append cases already improved in 26.7.2 / 26.8.1-SNAPSHOT.

### Error (client side, returned from remote commit)

```
com.arcadedb.exception.ConcurrentModificationException: Concurrent modification
on page PageId(<db>/343/0) in file '<Type>_14.343.65536.v0.bucket'
(current v.882 <> database v.884). Please retry the operation (threadId=95)
    at ... RemoteDatabase.commit
    at ... RemoteHttpComponent.manageException
```

### We are already using the mitigations usually suggested — and they are not enough

This is the key point for #5279: the usual advice ("retry it" / "add more buckets") is **already applied here**, and the conflict still isn't recoverable.

- **We already use your built-in transaction retry.** The whole bulk runs inside `database.transaction(txBlock, joinCurrentTransaction=false, attempts=5)`, and `RemoteDatabase.transaction(...)` retries the entire `begin`→`commit` block on `NeedRetryException` / `ConcurrentModificationException` (CME extends `NeedRetryException`, and `executeCommand` explicitly flags it `// SUPPORT RETRY`). The failure above happens **after all 5 attempts are exhausted** — under sustained concurrency the page keeps being re-versioned between our read and commit, so the retry budget never converges.
- **The schema is already heavily bucketed.** Every concrete vertex type (5 types) *and* every concrete edge type (8 types) has **16 buckets**. (The base `V`/`E` keep the default single bucket, but records live in the 16-bucket concrete subtypes.) So this is not simple type-bucket scarcity.

### Where the conflict actually is

The conflicting page lives in a **vertex** bucket (`<Type>_14...`), and the workload attaches ~3,900 edges to a small number of hub vertices in a single commit. This looks like **super-node edge-list append contention on a hot vertex** — adding edges bumps the version of the page holding the vertex / edge-list head — rather than record-level contention. Increasing the *type's* bucket count does not help a single hot vertex. There is also a `UNIQUE` index on `E(@out, @in)` that every edge insert must update.

### Questions

1. Given #5147 and the "parallel add edge for super nodes" work mentioned here for 26.8.1-SNAPSHOT: is there a supported/config-gated way (e.g. `arcadedb.graph.edgeAppendMerge`) to let concurrent edge appends to the same super-node commit **without** a full-transaction CME? For this workload that seems to be the missing piece — more so than client retries or more buckets.
2. For **explicit** transactions over the remote HTTP client, is there anything beyond the `attempts` retry loop we should be doing (the `retry` param from #2236 appears to cover only the implicit `/command` path)?
3. Is there any plan to raise CME only on **true record conflicts** rather than page-level collisions (the core ask in #5279)?

Related: #5279, #2236, #5147.

lvca:
## Problem

Closes #5381 (reinforcing #5279).

Under concurrency, a bulk edge `CREATE_UPDATE` to hot vertices intermittently fails at commit with a **page-level** `ConcurrentModificationException` on the **vertex** bucket, even though the records being written are logically unrelated - a *false* conflict. ArcadeDB's optimistic concurrency is **page-versioned**, so two transactions writing *different, unrelated records that merely share a bucket page* collide at page granularity. Client retries + more buckets do not converge under sustained load.

## Root cause (proven by reproduction)

Appending edges to a hot vertex is mostly a *tracked in-chunk append* (already covered by the edge-append merge). But whenever a hub's head chunk fills, `EdgeLinkedList.add` **rewrites the hub vertex record in place** to flip its edge-list head pointer. Two threads flipping the heads of *different* hubs co-located on one vertex page collide falsely.

Measured (each thread touching only its own distinct hub, so every conflict is provably false): **2497 / 2513 conflicts on the vertex bucket**, with the existing edge-append merge firing **0** times (it only rebases in-chunk appends).

## Fix

New `TX_SLOT_MERGE` (default `true`) generalizes the edge-append merge to all records. On a commit-time bucket page-version conflict, replay *this transaction's* slot writes on top of the newer committed page instead of failing the whole transaction - but only when every change it made to the page was:

- a disjoint-slot **insert** into a slot the committed page still leaves free, or
- a same-or-smaller **in-place update** whose committed pre-image is byte-for-byte unchanged.

A genuine **same-record** conflict (committed slot bytes differ from the captured pre-image), or any non-rebasable change (delete, multi-page/placeholder record, record growth), **poisons** the page and still raises the exception for a normal retry. The record keeps its RID (so index entries stay valid). Leader/embedded only, under the bucket commit lock, mirroring the existing `rebaseEdgeAppends`. Edge-list segments are deliberately left to the edge-append merge.

### Why not the reporter's proposal
Renaming the exception and doing app-side reload/compare pushes the problem onto every client and doesn't fix the livelock. Resolving the *false* conflicts inside the engine, while still surfacing *true* record conflicts, is what #5279 actually asked for.

## Scope / limitations (intentional)
- Growing updates (serialized size increase) and pure insert-vs-same-slot collisions are genuine contention and still retry.
- A unique-index page conflict is a separate concern and can still raise CME (not made worse).

## Changes
- `GlobalConfiguration.TX_SLOT_MERGE`
- `TransactionContext`: per-page slot tracking/poison + `rebaseSlots`, wired into `commit1stPhase` alongside the edge-append rebase
- `LocalBucket`: track/poison hooks in `create`/`update`/`deleteRecordInternal`, plus `rebaseRecordOnPage`/`writeRecordAtSlot` primitives
- `PageManager`: `slotMerges` stat

## Tests
`Issue5381FalseConflictTest`:
- distinct-record in-place updates absorbed by the merge (attempts=1, merge counter fires, values exact),
- a genuine same-record conflict still raises `ConcurrentModificationException`,
- a hot-vertex edge-append integrity run (all edges preserved).

Verified green: edge-merge/super-node suites, transaction/CRUD/concurrency (ACID, isolation, ConcurrentWrite), index/DML, and the full graph package. Merged latest `main` (no conflicts) and re-ran the affected suites clean.

https://claude.ai/code/session_0132A3poFCPAXUQKNhe9EdvQ


lvca:
Thanks for the careful review - all points addressed in a3309d3.

### Performance / memory (main concern)
- **Retention cap:** added `TX_SLOT_MERGE_MAX_BYTES` (default 16 MB), a per-transaction soft cap on retained images. Once a transaction exceeds it, the merge disables itself for the rest of that transaction and **drops the tracked images**, so a large bulk in-place UPDATE degrades to plain MVCC instead of holding ~2x its touched records until commit. Bulk INSERT into fresh pages was already excluded (`!createNewPage`); bulk UPDATE is now bounded by the cap.
- **Cheaper hot path:** `LocalBucket` now passes the already-computed `(fileId, pageNumber, slot)` into `trackRebasableInsert/Update` / `poisonSlotRebasePage`, removing the per-write `getBucketById` lookup + cast + division that `bucketPageKey`/`slotInPage` did.
- **On the eager copies:** the `baseBody` pre-image genuinely must be captured eagerly - the committed pre-image is overwritten by commit time, so it cannot be recovered lazily; that's the invariant the false-vs-true conflict test depends on. The `finalBody` copy could in principle be re-read from the (decompressing) bucket read at rebase time to drop one copy on the hot path; I kept it eager for now since the cap already bounds total retention, but happy to do that follow-up if you'd like to shave the steady-state allocation further. If you'd prefer **zero** uncontended overhead, flipping `TX_SLOT_MERGE` default-off is a one-line change - your call.

### Test coverage gap
- Added **`concurrentInsertIntoPageChurnedByOtherSlots`**: a single inserter appends brand-new records into a shared bucket page while several updater threads churn *other* slots of that same page in place, with `attempts=1`. The inserter can only keep committing via the INSERT rebase (`trackRebasableInsert` -> `writeRecordAtSlot`: free-space walk, slot-table/record-count bump, `updatePageStatistics`); it asserts the merge counter fires and every leaf survives exactly once.
- Added **`tinyRetentionCapFallsBackWithoutCorruption`**: cap set to 1 byte so the merge disables mid-transaction and every page falls back to plain MVCC; asserts final values stay consistent.

### Minor
- `isSlotMergeCandidate` now uses `EdgeSegment.RECORD_TYPE` instead of the literal `3`.
- Added a comment in `rebaseRecordOnPage` documenting the intentional asymmetry: a "cannot rebase this slot" outcome returns `false` (clean CME retry), but a genuine I/O read failure aborts the transaction rather than masquerading as a retryable version conflict.

All focused tests plus the edge-append/super-node, transaction/CRUD/ACID, and index/DML suites are green after the change.

lvca:
Thanks - second round addressed in c45106e.

### Performance / GC
- **Copy computed before the poison check** - fixed. `create`/`updateRecordInternal` now skip building the record-image copy when the target page is already poisoned (new `isSlotRebasePagePoisoned` probe), so churned/poisoned pages no longer allocate a copy just for `trackRebasable*` to discard.
- **Steady-state allocation, default-on** - acknowledged and inherent: the `baseBody` pre-image must be captured eagerly (the committed pre-image is gone by commit time), which is the invariant the false-vs-true conflict check relies on; the edge-append merge pays the same kind of price. The 16 MB cap bounds retained heap (not allocation rate), as you note. If the steady-state cost is a concern I'm happy to follow up by dropping the `finalBody` copy (re-read from the decompressing bucket read at rebase time), or we flip `TX_SLOT_MERGE` default-off - your call.

### `accountTrackedBytes` over-count
Fixed. It now counts the **net** change per slot: a repeated write to an already-tracked slot replaces its image rather than adding one, and only the first `baseBody` capture per slot is added. So `slotRebaseTrackedBytes` now measures bytes actually retained, matching the config doc's "images retained" wording.

### Tests
Both flake sources removed:
- The in-place-update test now asserts each record holds **exactly its own last committed value** (single owner per record makes this exact and load-independent) plus `slotMerges > 0`; the CME ratio is gone.
- The insert test keeps its exact count checks (`countType`/leaf count, both net of `insertCme`) and uses a **wide majority** bound that is robust under CI load but still can only hold if `writeRecordAtSlot` is actually replaying the inserts. Ran the class 3x locally with no flakes.

### Questions
- **Unique indexes** - confirmed. The rebase only re-derives the conflicting *bucket* page and runs in phase-1, before any page is written to the WAL. The unique-index pages are validated in the same phase-1 pass (as `modifiedPages`/`newPages`); a conflict there throws before the WAL append, so the whole transaction rolls back and retries. No half-applied state - a cleanly-rebased bucket page never commits without its index pages also passing.
- **`TX_SLOT_MERGE` default `true` is a behavioral change** - agreed, worth a release note. Because a merged page produces the exact same committed state a serialized retry would (same RID, same bytes, true same-record conflicts still surface as CME), it's transparent to correctness; the visible difference is that false-conflict CMEs that used to reach the client are now absorbed server-side.

All 5 tests plus the earlier regression set stay green.

lvca:
Thanks - third round addressed in 3d260075.

### 1. Write-path allocation cost - benchmarked
Added `SlotMergeWriteOverheadBenchmark` (`@Tag("benchmark")`, excluded from CI). Single-thread, **zero-contention** (the merge tracks every write but never rebases), JVM per-thread allocation counter, 20k warmup + 200k ops:

| workload | merge OFF | merge ON | delta |
|---|---|---|---|
| in-place update (same size) | 144,354 B/op | 145,045 B/op | **+691 B/op (~0.5%)** |
| insert into reused page | 189,567 B/op | 190,267 B/op | **+700 B/op (~0.4%)** |

Throughput was unchanged (within run-to-run noise). The record-image copies add ~700 bytes/op against a ~144-190 KB/op baseline that's dominated by the existing per-transaction/WAL machinery, so the steady-state overhead is **~0.4-0.5%** - negligible, which supports keeping it default-on.

On the **hash-guard** idea: given the measured cost is that small, I'd rather keep the exact byte-for-byte pre-image compare than swap it for a 64-bit hash. For a DBMS a hash collision (however unlikely) would turn a genuine same-record conflict into a silently applied lost update - trading a *guaranteed* no-lost-update invariant for a probabilistic one, which isn't worth ~350 bytes/op. If we ever want to shave it further without weakening correctness, the better lever is dropping the `finalBody` copy and re-serializing the (still-cached) record at rebase time; happy to do that as a follow-up if the steady-state number ever matters.

### 2. `slotRebaseTrackedBytes` not cleared in reset()
Fixed - `reset()` and `kill()` now re-zero it alongside the maps (begin() already did).

### 3. `updatePageStatistics` double-count on insert rebase
Acknowledged, left as-is: it's keyed by page number and re-sets that page's entry to a fresh (actually more accurate, since it's computed on the committed page) value rather than truly accumulating, and it only affects the best-effort free-space heuristic in HIGH reuse mode, self-correcting on the next gather. Not worth touching the free-space accounting for.

### 4. Test coverage
- Added `growingUpdatesOnCoLocatedRecordsFallBackAndStayCorrect`: several threads grow their own co-located records; a growth update shifts other slots, so it poisons the page and retries rather than merging - asserted via exact final state (a wrongly-merged growth would tear the record). Note it's also doubly-guarded: even if a growth slipped past the poison, `rebaseRecordOnPage` rejects `body.length > committedSize`.
- **HA replica path**: agreed it's the one place the committed image differs from what the tx computed. It ships through the WAL identically to `rebaseEdgeAppends` (leader-only, under the commit lock), which the existing HA suite covers; I can add a dedicated server-module replica test for the slot merge if you'd like one.

### Nit
`getTransactionIfExists()` per write - kept intentionally: `createRecord` can run before the tx's page fetch on some paths, so the null-tolerant getter avoids assuming an active tx in the hook; the cost is a `ThreadLocal` get, negligible next to the copies it guards.

All 6 tests green; benchmark reproduces the numbers above.

lvca:
Thanks - fourth round addressed in b6ab5e2.

### 1. Index consistency after a merged in-place update (highest value)
Added `mergedInPlaceUpdateKeepsIndexConsistent`: a `NOTUNIQUE` index on the mutated property, distinct co-located records each updated by their own thread with `attempts=1` (so the merge fires - asserted via the counter). After the run it verifies, for every record: the stored value equals its last committed write, the **index finds exactly that record by that value** (RID matches, `hits == 1`), and the superseded value leaves **no stale index entry**. This locks down the "RID preserved => index entries stay valid" claim.

### 2. Delete-poison fallback
Added `deleteOnSharedPageFallsBackAndStaysCorrect`: deleter threads delete their co-located victims (a delete poisons the page, so it must never merge) while other threads update co-located survivors in place. Asserts survivors keep exact last-committed values, the type count drops to the survivors, and every victim is gone - proving delete falls back to a clean retry rather than being rebased.

### 3. Import ordering
Fixed - `import com.arcadedb.graph.EdgeSegment` moved into alphabetical position.

### 4. Steady-state allocation
Confirmed acceptable and cited: the `SlotMergeWriteOverheadBenchmark` measures **~700 bytes/op (~0.4-0.5%)** extra allocation on a ~144-190 KB/op transaction baseline, single-thread zero-contention, with no measurable throughput change - which is why it's comfortable default-on.

### 5. Discoverability
The `TX_SLOT_MERGE` config description now opens with "Generalization of GRAPH_EDGE_APPEND_MERGE...", so a reader lands on both mechanisms together.

All 8 tests in the class pass (3 consecutive runs, no flakes) and the index regression set (LSMTree/Hash/unique/edge-index) stays green.

lvca:
Thanks - this was the important one. **Finding #1 is a real bug and it's now fixed** (dd3f7c6), verified with a deterministic reproduction.

### 1. Multi-page chunk on a reused page (silent corruption) - CONFIRMED and fixed
You're exactly right. `writeMultiPageRecord` and `updateMultiPageRecord` place/relocate/free `NEXT_CHUNK` records on **reused existing pages** via inline record-table writes that bypass `create/update/deleteRecordInternal`, so those pages were neither tracked nor poisoned. A page carrying a tracked small insert **and** a chunk stayed "rebasable", so on a page-version conflict the rebase replayed only the insert and dropped the chunk.

I reproduced it deterministically (16 KB pages; one tx inserts a small record + a large record whose trailing chunk reuses the small record's page; a competitor bumps that page to force the conflict). With the guard removed, the large record reliably fails to read afterwards:

```
ConcurrentModificationException: Multi-page record #1:2048 was modified during read after 3 retries
```

**Fix:** both multi-page writers now poison every page they touch (the current chunk page + every continuation/reused/freed page). A page carrying both a tracked write and a chunk can no longer be rebased - it falls back to a normal retry. Poisoning brand-new pages is harmless (new pages are never rebase-tracked). The new test `multiPageRecordCoLocatedWithTrackedInsertStaysIntact` fails without the fix (3/3) and passes with it (verified 4/4).

I also audited **every** inline record-table writer (`grep` of `PAGE_RECORD_TABLE_OFFSET`/`PAGE_RECORD_COUNT_IN_PAGE_OFFSET` writes): the rest are either covered (create/update/delete hooks), commit-phase only (`compressPage`/`defragPage`/`getOrderedRecordsInPage(readOnly=false)` run at commit on pages that commit as-is or are discarded and reloaded during rebase), or new-page only (`createRecordsBulk`). So the two multi-page methods were the only gap.

### 2. Steady-state allocation (default-on)
Confirmed acceptable via `SlotMergeWriteOverheadBenchmark`: ~700 bytes/op (~0.4-0.5%) on a ~144-190 KB/op baseline, no measurable throughput change.

### 3. Minor
- `deleteRecordInternal` mixing `getTransactionIfExists()` / `getTransaction()`: left as-is (harmless; the poison hook must tolerate the no-tx path uniformly with create/update).
- Added the comment explaining that `slotRebasePoisonedPages` is intentionally not cleared when the cap disables the merge mid-transaction.

Regression-checked LargeRecords, ConcurrentMultiPageRecordRead, EdgeAppendMergeMultiPageChunk, ApplyChangesPartialReplay, CRUD, edge-append and transaction suites - all green. The full `Issue5381FalseConflictTest` is now 9 methods.

Great catch - this is exactly the kind of storage-internal edge case worth being paranoid about.

Issue #5358, opened 2026-07-21
Cypher: variable-length path follow-ups to PR #5357 (VarLengthExpand physical operator, anchor-guard relaxations)


lvca:
PR #5357 taught the traditional Cypher executor to start a bounded variable-length path from the cost-based optimizer's indexed equality anchor, reversing the traversal internally while preserving written path order. The guard is intentionally conservative, which leaves several follow-up opportunities on the table.

## 1. Native `VarLengthExpand` physical operator (long-term fix)

The cost-based optimizer still excludes any statement containing a variable-length path from physical-plan execution (`hasVariableLengthPath()` disables the optimizer path entirely). The proper fix is a dedicated `VarLengthExpand` physical operator, as Neo4j has, so variable-length patterns are planned and executed by the optimizer like any other expansion:

- Anchor selection, filter pushdown, expand-into and join ordering would apply to variable-length patterns for free.
- The bridge added in #5357 (`shouldReverseVariableLengthPathFromIndexedAnchor` in `CypherExecutionPlan`) becomes dead code and can be removed once this lands.

## 2. Relax the anchor-reversal guard (short-term wins)

While the physical operator does not exist, the #5357 guard can be extended incrementally. Each bullet is independently shippable, in rough order of value:

- **Unbounded paths (`*`, `*2..`)**: currently excluded (`getMaxHops() == null`). Reversal is equally correct for unbounded expansion; the exclusion was pure caution.
- **Anonymous source variables**: `MATCH (:Area)-[:PART_OF*0..3]->(country:Area) WHERE country.id = $x` is rejected only because the written source has no variable. The effective-target swap already generates anonymous variable names for the fixed-length case.
- **IN-list anchors**: `WHERE country.id IN $ids` is excluded (`InListValues`). It could seed the reversed traversal with one index seek per list value.
- **Composite-index prefix anchors**: anchors on a multi-property index are excluded (`getPropertyNames().size() != 1`) even when the equality predicate covers the full key or a selective prefix.
- **Multi-pattern and multi-MATCH statements**: reversal is limited to a single MATCH with a single path pattern. Patterns whose variable-length segment is independent of the other parts could still benefit.

## Reference

- PR #5357 (merged): the conservative bridge, including `CypherVariableLengthAnchorSelectionTest`.
[...]

PR #5387, opened 2026-07-22
Optimize IN-list anchors for bounded variable-length Cypher paths

justinblethrow-cloud:
## Summary

- allow the bounded variable-length path bridge introduced in #5357 to accept indexed `IN`-list target anchors
- start the reversed traversal with `IndexSeekStep`, reusing the existing multi-value seek and runtime parameter handling
- add regression coverage for both the index-backed plan and existing bound-target reversal without a physical optimizer plan

This preserves the existing bridge safety gates: read-only execution, one bounded and typed bidirectional relationship, a single-property index, and no shape-changing clauses before `MATCH`.

The indexed-anchor reversal is tracked separately from the existing reversal of ordinary single-hop paths whose target is already bound, so fallback execution does not require a physical optimizer plan.

## Motivation

The bridge currently rejects `InListValues`, so a query such as:

```cypher
MATCH (place:Area)-[:PART_OF*0..3]->(country:Area)
WHERE country.id IN $countryIds
RETURN country.id, place.id
```

cannot start from the selective indexed target even though `NodeIndexSeek` already supports literal and parameterized `IN` lists. This patch connects that existing operator to the traditional variable-length executor.

Part of #5358.

## Verification

```text
./mvnw -pl engine -Dtest=CypherVariableLengthAnchorSelectionTest,CypherMultiHopInListIssue5306Test,OpenCypherVariableLengthPathTest,Issue5362DuplicatePredicateIndexSeekTest,OpenCypherOptimizerVerificationTest,AnchorSelectorTest,CypherOptimizerIntegrationTest,CountEdgesOptimizationTest,StreamingAggregationTest,CypherLabelFilteringTest,OpenCypherOptionalMatchTest test
```

Result: 128 tests run, 0 failures, 0 errors, 0 skipped.

```text
./mvnw -pl engine -Dtest=com.arcadedb.query.opencypher.tck.OpenCypherTCKSuite test
```

Result: 3,897 scenarios run, 0 failures, 0 errors, 85 expected skips.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -1554,7 +1554,9 @@ public class CypherExecutionPlan {
         String sourceVar = sourceNode.getVariable() != null ? sourceNode.getVariable() :
             ("  src" + anonymousVarCounter++);
 
-        boolean reversed = shouldReverseVariableLengthPathFromIndexedAnchor(matchClause, pathPattern);
+        final boolean reversedFromIndexedAnchor =
+            shouldReverseVariableLengthPathFromIndexedAnchor(matchClause, pathPattern);
+        boolean reversed = reversedFromIndexedAnchor;
         if (reversed) {
           sourceNode = pathPattern.getLastNode();
           sourceVar = sourceNode.getVariable();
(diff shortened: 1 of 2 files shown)


lvca:
@claude please review this PR

lvca:
@claude please review this PR

justinblethrow-cloud:
Thanks, addressed in ca21bc50.

I confirmed that both reported constraint shapes already fall back to traditional execution under the current optimizer eligibility checks: inline node properties and multi-label relationship endpoints are rejected before a physical plan is built. I nevertheless added the safety check at the indexed IN-list bridge itself so future optimizer expansion cannot make the shortcut bypass those constraints.

The shortcut now requires an anchor with no inline/parameter properties, no dynamic labels, and exactly one static label matching the indexed type. Added focused regressions for an inline property and an additional label; both verify correct results and fallback plan selection.

Verification: `./mvnw -pl engine -Dtest=CypherVariableLengthAnchorSelectionTest test` — 8 tests passed, 0 failures/errors.

lvca:
Really great PR, merging it! Thanks!

PR #5393, opened 2026-07-23
Relax indexed anchors for open-ended and anonymous variable-length paths

justinblethrow-cloud:
## Summary

- allow the traditional-executor bridge to start open-ended variable-length paths (`*` and `*2..`) from the indexed target selected by the optimizer
- retain the generated internal source variable when the written source node is anonymous
- preserve the `IN`-list constraint safety guard from #5387 and cover the combined anonymous/open-ended/multi-value case

This is a focused follow-up to #5357 and #5387 and addresses the first two remaining guard relaxations in #5358.

## Safety

The existing conservative requirements remain unchanged: read-only single-path queries, one typed variable-length relationship, stored incoming adjacency, a single-property indexed target, and the existing exclusions for optional/multi-MATCH, range, composite-index, unidirectional, untyped, and constrained `IN`-list anchor shapes.

The underlying `ExpandPathStep` already implements open-ended traversal; the previous upper-bound requirement was only an admission guard. Anonymous sources use the internal variable already allocated by the traditional executor so the reversed expansion can bind the original written source correctly.

## Verification

- `./mvnw -pl engine -Dtest=CypherVariableLengthAnchorSelectionTest test`: 14 passed
- adjacent variable-length, optimizer, anchor-selector, and multi-value gate: 85 passed
- `git diff --check origin/main...HEAD`: passed

The new coverage includes `*`, `*2..`, bounded and open-ended anonymous sources, the combined indexed `IN`-list case, and written relationship-list order for open-ended reversed traversal.

This PR intentionally does not address composite indexes, multi-pattern planning, range anchors, or the future native variable-length expansion physical operator.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -1553,6 +1553,7 @@ public class CypherExecutionPlan {
         NodePattern sourceNode = pathPattern.getFirstNode();
         String sourceVar = sourceNode.getVariable() != null ? sourceNode.getVariable() :
             ("  src" + anonymousVarCounter++);
+        final String writtenSourceVar = sourceVar;
 
         final boolean reversedFromIndexedAnchor =
             shouldReverseVariableLengthPathFromIndexedAnchor(matchClause, pathPattern);
@@ -1681,8 +1682,7 @@ public class CypherExecutionPlan {
           final Direction directionOverride;
           if (reversed) {
             effectiveSourceVar = currentSourceVar; // already swapped to bound target
-            effectiveTargetVar = pathPattern.getFirstNode().getVariable() != null ?
-                pathPattern.getFirstNode().getVariable() : targetVar;
+            effectiveTargetVar = writtenSourceVar;
             targetVar = effectiveTargetVar;
             effectiveTargetNode = pathPattern.getFirstNode(); // original source becomes target for label filtering
             directionOverride = relPattern.getDirection().reverse();
@@ -1792,8 +1792,8 @@ public class CypherExecutionPlan {
 
   /**
    * The physical operators do not yet implement variable-length expansion, but their cost-based
-   * anchor selection is still useful to the traditional executor. Limit this bridge to a single,
-   * bounded relationship whose indexed target can be reached through stored incoming adjacency.
+   * anchor selection is still useful to the traditional executor. Limit this bridge to a single
+   * relationship whose indexed target can be reached through stored incoming adjacency.
    */
   private boolean shouldReverseVariableLengthPathFromIndexedAnchor(final MatchClause matchClause,
       final PathPattern pathPattern) {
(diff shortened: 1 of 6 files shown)


lvca:
## Review: approved, with one root-cause fix and regression coverage pushed to this branch (`e85d6494a`)

Thanks - the relaxation is correct, and it fixes more than the description claims. Details of what I verified, found and changed.

### The relaxation is sound (verified, not assumed)

Reversing a variable-length path is only legal if path enumeration is symmetric. Under Cypher's relationship-uniqueness (trail) semantics it is, including in the presence of cycles, so unbounded `*` terminates and row multiplicity is preserved. Rather than take that on faith I added a parity harness: every relaxed shape runs twice against the same graph - once anchored on an indexed property (bridge reverses the traversal) and once on a non-indexed twin property (bridge stays out of the way) - and both runs must produce identical rows, identical multiplicity and identical written path order. The graph is deliberately hostile: a diamond gives two distinct trails between the same pair, and a back edge closes a cycle.

11 shapes covered: `*`, `*2..`, `*0..`, `*1..3`, undirected, anonymous source, anonymous + `count(*)`, written relationship-list order, source predicate in `WHERE`, indexed `IN`-list, and the anchor-rebinding case. All identical.

### The PR undersells itself: hunk 2 fixes a live bug outside variable-length paths

Binding the reversed hop to `writtenSourceVar` also repairs the ordinary bound-target reversal for an anonymous source. On `main`:

```cypher
MATCH (q:Q) OPTIONAL MATCH (:C)-[:ON]->(q) RETURN q.name, count(*)
```

returns `1` for a vertex with two incoming edges - the hop was bound to the target variable, collapsing every incoming edge into a single row. There was no test for it; I added one to `OpenCypherOptionalMatchTest` and confirmed it fails without your change.

### Bug found and fixed on this branch: variable-length target labels

The reversal evaluates the *written source node* as the expansion target, which puts more weight on `ExpandPathStep`'s target-node label check - and that check was wrong. It ANDed the pattern's labels and ignored Cypher 25 dynamic `$(expression)` labels:

```cypher
MATCH (hub)<-[:LINK*1..2]-(n:Special|Other)   -- returned [] instead of [other, special]
MATCH (n:Node:$('Special'))-[:LINK*1..2]->(hub) WHERE hub.id = 'hub'   -- returned unfiltered rows
```

This is not reversal-specific - plain forward expansion has it too - so I fixed it at the root in `ExpandPathStep`, matching `MatchNodeStep` semantics (disjunction accepts any label, conjunction requires all, dynamic labels resolved per row). The evaluator is allocated only when the pattern actually carries a dynamic label, so nothing changes for the common case. 3 of the 5 new tests fail without the fix.

### On the guard itself

The guard is now 20 conjunctions that all silently return `false`, with no way to tell which one fired and no distinction between "conservative" and "required". Exactly two of them are load-bearing, and I documented that in the javadoc:

- **single relationship** - the reversal maps the one hop back onto the written first node; for `i > 0` it would rebind the same variable. Reversing a longer pattern needs the whole pattern rebuilt back to front, which is the natural next refactor.
- **single-property index** - see below. Do **not** relax the composite-index bullet of #5358 yet; it is currently masking a wrong-results bug (filed separately).

Everything else in the guard is bounding proven shapes, not soundness. Longer term, #5358's `VarLengthExpand` physical operator remains the right answer - each relaxation widens a hand-written reversal that duplicates knowledge the optimizer already has.

### Worth knowing for the rest of #5358

The bridge can never engage for `MATCH p = ...` or for any node carrying an inline property constraint, because `CypherExecutionPlanner` disables the optimizer wholesale for both, so no physical anchor exists at all. A slice of the remaining #5358 value is therefore gated on planner eligibility, not on this guard. I locked both shapes in with a test so a future eligibility change cannot silently reroute them.

### Verification

- `com.arcadedb.query.opencypher.**`: 7457 tests, 0 failures, 98 skipped - including the openCypher TCK suite (3897 scenarios, 0 failures).
- Every new test confirmed to fail when the corresponding fix is reverted.
- All test databases dropped; no debug output left.

### Pushed to this branch (`e85d6494a`)

- `ExpandPathStep.java` - label disjunction + dynamic label resolution on the expansion target
- `CypherExecutionPlan.java` - javadoc/comments only, documenting the load-bearing conditions
- `OpenCypherOptionalMatchTest.java` - anonymous reversed-source regression
- `CypherVariableLengthReversedAnchorSemanticsTest.java` (new) - the parity suite
- `OpenCypherVariableLengthTargetLabelTest.java` (new) - target-label semantics, both orientations

Please rebase on `main` before this goes to the merge queue; the branch is a bit behind.

lvca:
Thanks @justinblethrow-cloud !

Issue #5394, opened 2026-07-23
Model Context Protocol 2025-03-26 transport conformance: batches, notifications, GET, and Origin handling


justinblethrow-cloud:
ArcadeDB advertises Model Context Protocol (MCP) version `2025-03-26`. While reviewing the current MCP implementation on `main` at
[`95148b21`](https://github.com/ArcadeData/arcadedb/commit/95148b215136a7c000f9be19f61720821c3341db),
I found several transport-envelope behaviors that appear not to match explicit requirements of that version.

This is not a report of a known outage: tolerant clients may work today. The concern is predictable interoperability with
standards-conforming clients and one required Hypertext Transfer Protocol (HTTP) security check. The protocol messages use
JavaScript Object Notation Remote Procedure Call (JSON-RPC).

## Findings

| Area | MCP 2025-03-26 requirement | Current ArcadeDB behavior |
| --- | --- | --- |
| JSON-RPC batches | Implementations may send batches but **must support receiving** them. | HTTP request parsing constructs a single `JSONObject`, and standard input/output (stdio) parses each line as a single `JSONObject`; a top-level array is rejected rather than dispatched. |
| Notifications | A receiver **must not send a response** to any notification. | `notifications/initialized` is special-cased, but another notification reaches the default `Method not found` response with a null ID. |
| Request envelopes | Requests must carry a string or integer ID, while notifications must not carry an ID. | The dispatcher reads an optional ID but does not validate the request/notification shape, so a request method without an ID can produce a response whose ID is null. |
| HTTP notification status | An accepted POST containing only notifications/responses **must return `202 Accepted` with no body**. | `MCPHttpHandler` returns `204`, and the current integration test explicitly expects `204`. |
| HTTP GET | The MCP endpoint must accept GET and either open a Server-Sent Events stream or return `405 Method Not Allowed`. | The MCP route is method-agnostic and has no GET branch, so GET is handled as an empty JSON-RPC request rather than intentionally returning `405`. |
| Origin validation | Streamable HTTP servers **must validate `Origin`** to mitigate Domain Name System (DNS) rebinding. | I could not find Origin validation in `MCPHttpHandler` or the shared HTTP handler. Please confirm there is not another layer providing it. |

Specification references:

- [Base protocol: requests, notifications, and batching](https://modelcontextprotocol.io/specification/2025-03-26/basic/index)
[...]

PR #5396, opened 2026-07-23
Complete MCP response and DNS-rebinding handling

justinblethrow-cloud:
## Summary

- accept valid JSON-RPC response messages as one-way input over HTTP and standard input/output, including response-only batches
- return HTTP `202 Accepted` with no body for response-only input, and emit no standard-output response
- reject an empty standard-input batch with JSON-RPC `-32600`, matching the HTTP transport
- require non-loopback browser origins to be explicitly configured instead of trusting equality between the attacker-controlled `Origin` and `Host` names
- add six focused transport regression tests

## Context

This is a focused follow-up to #5394 and commit `303e83c9`. That merged change handles the six reported findings; validation found these remaining envelope cases.

The [MCP 2025-03-26 Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) accepts single response messages and response batches, and requires `202 Accepted` with no body for response-only or notification-only input. A receiver must not generate a response to a response.

For DNS rebinding, comparing the Origin hostname with the request Host hostname is not a trust check: the attack deliberately makes the attacker hostname resolve to the local endpoint, so both values match. The [official TypeScript SDK server guidance](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/server.md#dns-rebinding-protection) describes this same-origin property and recommends explicit host validation. This patch keeps ArcadeDB existing `allowedOrigins` design and makes non-loopback browser origins opt in through that list.

The separate documentation repository currently says same-host origins are always accepted. That sentence will need the corresponding small update if this behavior is accepted.

## Testing

- `./mvnw -pl server -am -Dtest=MCPTransportConformanceTest -Dsurefire.failIfNoSpecifiedTests=false test` (24 passed)
- `./mvnw -pl server -am -Dtest=MCPConfigurationTest,MCPStdioServerTest -Dsurefire.failIfNoSpecifiedTests=false test` (23 passed)
- full engine suite not run locally

### DIFF
--- a/server/src/main/java/com/arcadedb/server/mcp/MCPConfiguration.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/MCPConfiguration.java
@@ -48,9 +48,9 @@ public class MCPConfiguration {
   private volatile boolean      allowSchemaChange = false;
   private volatile boolean      allowAdmin        = false;
   private volatile List<String> allowedUsers     = new CopyOnWriteArrayList<>(List.of("root"));
-  // Extra browser origins accepted by the HTTP transport, on top of the always-allowed loopback and
-  // same-host ones. Empty by default: a cross-origin browser page must be opted in explicitly, because
-  // accepting any origin would defeat the DNS-rebinding mitigation MCP requires.
+  // Extra browser origins accepted by the HTTP transport, on top of always-allowed loopback origins.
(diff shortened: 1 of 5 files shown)


lvca:
@claude please review this PR

Issue #4859, opened 2026-07-02
Epic: MCP GraphRAG & Agent-Memory Surface


robfrank:
## Problem

ArcadeDB's MCP server (`server/src/main/java/com/arcadedb/server/mcp/`) exposes 10 tools over stdio and HTTP: `list_databases`, `get_schema`, `query`, `execute_command`, `server_status`, `profiler_start`, `profiler_stop`, `profiler_status`, `get_server_settings`, `set_server_setting`. Every one serves the **operator/DBA persona**. There is no first-class **retrieval** surface (vector / hybrid / full-text) and no **knowledge-graph-construction** surface — exactly the persona ArcadeDB's GraphRAG / agent-memory positioning depends on. Only the MCP **Tools** primitive is implemented; **Resources** and **Prompts** are entirely absent (no `resources/list`, `resources/read`, `prompts/list`, `prompts/get` — confirmed by source inspection, zero matches).

## Governing principle

`query` + `execute_command` are already Turing-complete over SQL/Cypher. A new tool is justified only if it meets at least one of:
1. **Ergonomic guardrail** — models reliably get the raw syntax wrong (e.g. `vector.neighbors`/`vector.fuse` argument shapes, MERGE-based upsert key matching).
2. **Unreachable capability** — not expressible, or unsafe/awkward, through `query`/`execute_command` alone (e.g. multi-step orchestration across several statements).
3. **Different permissioning** than raw query/command.
4. **Structured shaping** the agent specifically needs.

Tool-count discipline applies: the server is at 10 tools; flat tool lists degrade model tool-selection accuracy. Prefer **profiles** and **Resources/Prompts** over an ever-longer list.

## Success criteria

- [ ] An agent can perform vector, hybrid, and full-text retrieval through dedicated tools without hand-writing ArcadeDB-specific SQL function syntax.
- [ ] Schema is available as an MCP Resource, not only via a tool call.
- [ ] At least one guided Prompt exists steering correct tool use for retrieval and KG construction.
- [ ] The tool surface a given agent sees stays small via configurable profiles.
- [ ] All new read tools honor the `allowReads`/idempotency invariant; all new write tools honor the `OperationType`→`isAllow*()` gating pattern from `ExecuteCommandTool`.
- [ ] Permission defaults remain safe-by-default (`enabled=false`, all write flags `false` out of the box).
[...]

PR #5398, opened 2026-07-24
feat(mcp): add bounded sample_records tool

### LINKED ISSUE #4863: MCP: add sample_records tool

justinblethrow-cloud:
## Summary
- add a `sample_records` tool for bounded value-shape inspection across selected or all schema types
- validate requested types before generating SQL and return deterministic type ordering when `types` is omitted
- register the tool once in the shared dispatcher for both HTTP and stdio transports

## Safety
- gates execution on `allowReads`
- enforces a per-type limit of 1-20, defaulting to 3
- resolves types through the schema, quotes identifiers, and verifies generated SQL is idempotent
- returns an empty array for empty types

## Tests
`./mvnw -pl server -am -Dmaven.repo.local=/mnt/datavault/Agentic/Arcade_DB_Index_Lab/.m2/repository -Dtest='MCP*Test' -Dsurefire.failIfNoSpecifiedTests=false test`

147 tests passed across the focused MCP test set. The full engine suite was not run.

The tool definition carries its input and behavior documentation. Longer-form MCP user documentation lives in the separate `arcadedb-docs` repository.

Closes #4863


robfrank:
Reviewed against the epic #4859 acceptance criteria. The tool is small and clean, the `allowReads` invariant is honored, unknown types are rejected before any query is built, and identifiers go through `MCPToolUtils.quoteIdentifier`, so the injection shape covered by `sampleRecordsRejectsUnknownTypesBeforeQueryExecution` is closed twice over. Test coverage is proportionate. A few improvements.

### 1. "Bounded" is only per type

The PR title says bounded, and `limit` is correctly clamped to 1..20 per type. But when `types` is omitted, `resolveTypeNames` returns every type in the schema, so the response size is `20 x (number of types)` records with no ceiling on the second factor. On a database with a few hundred types that is a very large single MCP payload, which is the opposite of what a context-constrained agent wants.

Please add a total-record or total-type ceiling for the omitted-`types` case, and say in the tool description what that ceiling is so a truncated response is not mistaken for a complete one.

Related: `database.getSchema().getTypes()` also returns edge types. For the stated purpose ("inspect actual value shapes") edge records are mostly noise, and on a graph database they are frequently the largest types. Consider defaulting to vertex and document types and requiring an explicit `types` entry to sample an edge type.

### 2. The sample is not a sample

`SELECT FROM <type> LIMIT n` returns the first `n` records in bucket scan order, not a representative draw. For a type where early records are historic or differently shaped from current ones, an agent will generalize from an unrepresentative prefix, which is a real failure mode for the schema-inference use case this tool exists to serve.

I am not asking for real random sampling. Just state in the description that these are the first N records in storage order, so the model calibrates accordingly.

### 3. Audit log line reports a different unit from every other read tool

```java
case "sample_records" -> result.getJSONObject("samples", new JSONObject()).length() + " type(s)";
```

`query`, `execute_command`, `full_text_search` and the upserts all report record or hit counts. Reporting type count here means the operator log gives no signal about response size, which is the thing most worth knowing for this particular tool. Report the total record count, or both.

### 4. Missing the explanatory comment for the `analyzed.execute()` fallback

`sampleType` reproduces the pattern from `QueryTool`, but not `QueryTool`'s comment explaining why the null check exists (only the SQL engine reuses the parsed statement; the other engines return null and the query is reparsed). Without it the ternary reads as defensive noise. Worth carrying the comment across.

### Cross-PR: the tool-count assertions will break each other

This PR and #5399 both change `assertThat(tools.length()).isEqualTo(13)` to `isEqualTo(14)` in `MCPServerPluginTest` and `MCPStdioServerTest`. Whichever merges second leaves the other at 15 registered tools and a red build. #5402 has the same problem in a different form: its `rag` profile assertion is `containsExactlyInAnyOrder("query", "full_text_search")`, which breaks as soon as `sample_records` and `vector_search` register.

Please derive the expected value from the registered tool list rather than hardcoding it, or assert containment only. That removes the merge-order coupling across all four PRs.

### Cross-PR: #5401 changes the resolver signature

#5401 replaces `MCPToolUtils.resolveDatabase(server, user, name)` with a four-argument form returning a `DatabaseAccess`, and splits it into `resolveDatabase` (no read check) and `resolveReadableDatabase` (read check). This PR will not compile after that lands. When you rebase, `sample_records` is a database-targeted read tool and must use `resolveReadableDatabase`, not `resolveDatabase`, otherwise the inline `config.isAllowReads()` check here passes on the global setting while a per-database read denial is bypassed.

### Note on scope justification

Worth recording explicitly, since #4859 sets a deliberately high bar for new tools: this one earns its place on criterion 4 (structured shaping), because it returns one type-keyed map in a single call rather than N round trips of `SELECT ... LIMIT 3`. That justification gets stronger given #5402 puts `sample_records` in the `rag` profile while leaving `get_schema` out, which makes this the only shape-discovery path a retrieval agent has. I have raised the `get_schema` omission on #5402 separately.

justinblethrow-cloud:
Thanks, Roberto. Addressed in `64456091`:

- Omitted `types` is now bounded to 20 sorted vertex/document types and excludes edge types by default; explicit edge-type requests remain supported and `types` has `maxItems: 20`.
- The response now reports `sampledTypes`, `availableTypes`, `recordsReturned`, and `truncated`.
- The contract now says these are the first storage-order records, not random or representative samples.
- Audit output counts records across types, the analyzed-query fallback is documented, and merge-order-coupled tool-count assertions are removed.

Focused validation: `MCPServerPluginTest` + `MCPStdioServerTest`, 88 tests passed.

For the cross-PR resolver point, this branch remains sequenced before #5401. When rebased after #5401 it must use the explicit `READ` resolver mode.

robfrank:
Correction to my review above: I said the docs criterion in #4859 could not be satisfied from these PRs because no MCP documentation lives in this repo. That was wrong. The MCP reference page is at https://docs.arcadedb.com/arcadedb/reference/mcp/mcp and its source is `ArcadeData/arcadedb-docs`, file `src/main/asciidoc/reference/mcp/mcp.adoc`.

That page currently documents exactly the 13 tools registered on `main` and nothing more, so it is in sync today and a PR that adds tool 14 without touching it regresses that sync.

For this PR, please open a companion `arcadedb-docs` PR adding `sample_records` to the tool list around line 113, covering the `database`, `types` and `limit` arguments, the 1..20 `limit` clamp, and whatever ceiling you settle on for the omitted-`types` case from point 1 of my review.

justinblethrow-cloud:
Companion documentation is now open as ArcadeData/arcadedb-docs#425.

It covers the `database`, `types`, and `limit` arguments; the 20-type and 1..20 per-type ceilings; default edge exclusion; storage-order rather than representative sampling; and the response truncation metadata.

- `python3 docs-validator.py`
- `./mvnw generate-resources`

robfrank:
`vector_search` (#5399) is merged, which put this branch in conflict on the three files both PRs touch. I have merged `main` in and pushed the resolution, so this is mergeable again.

### On your fix commit

`644560912` addresses all four points from my review, and the result is clean:

- `MAX_TYPES = 20` bounds the omitted-`types` case, and rather than silently capping you report `availableTypes` and `truncated` so the caller can tell a partial answer from a complete one. That is better than what I asked for.
- Edge types are excluded from the default set via `!(type instanceof EdgeType)`, while an explicitly named edge type is still honored. That is the right split.
- `formatResult` now reports `recordsReturned` across `sampledTypes`, so the operator log carries response size.
- The description says "first records in storage order", and the `analyzed.execute()` fallback comment is carried over from `QueryTool`.

I checked specifically for the class of problem your `002ab599` truncation fix hit on #5399, where the new reporting field pulled in an O(n) index scan. Nothing similar here: the type list is bounded before any query runs, and the per-type cost is one parse plus one bounded query.

### How I resolved the conflicts

All three were "both sides inserted an adjacent line", so both tools are kept everywhere: registration, dispatch, and `formatResult`.

The one judgment call was in the discovery tests. Both sides had independently replaced the brittle `assertThat(tools.length()).isEqualTo(13)` with a containment check, which would have left two near-identical loops. I merged them into a single pass asserting both `sample_records` and `vector_search`. Worth noting that this brittleness is now gone from both test classes, so the next tool to register will not conflict here the way this one did.

`mvn test -Dtest='MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest'` is green at 125 tests.

### Two things still open

**One inconsistency worth a look.** `availableTypes` means different things on the two branches of `resolveTypeNames`: the count of eligible types in the database when `types` is omitted, but the number of types the caller asked for when `types` is supplied. A model reading that field cannot tell which it is getting. Since `truncated` is always false on the explicit branch, the simplest fix is to report the database-wide eligible count in both cases.

**Docs.** The `arcadedb-docs` companion from my earlier comment is still outstanding, and it has grown a little: the page now lags main by two tools, since `vector_search` shipped undocumented. If it is easier, one `arcadedb-docs` PR covering both `sample_records` and `vector_search` is fine by me.

On CI: the red checks on this PR were an unrelated `LeaveClusterTest.transferLeadershipToSpecificPeer` HA flake, not anything from your changes. Ignore those runs; the new ones on this head are what count.

justinblethrow-cloud:
Implemented the requested follow-up in 5ca2b781: `availableTypes` now consistently reports the database-wide count of eligible vertex/document types, including when `types` is supplied. I added schema-derived assertions for the default, explicit-type, and explicit-edge cases.\n\nFocused MCP verification: 125 tests, 0 failures. The matching wording is also updated in ArcadeData/arcadedb-docs#425; `docs-validator.py` and `./mvnw generate-resources` both pass.

PR #5399, opened 2026-07-24
feat(mcp): add dense and sparse vector_search tool

### LINKED ISSUE #4860: MCP: add vector_search tool
Part of #4859.

## Motivation

Models reliably fail to produce correct `vector.neighbors`/`vector.sparseNeighbors` SQL — argument order (`index-name, key-or-vector, k, [efSearch|options-object]`), dense-vs-sparse index selection, and filter/groupBy option shapes are all ArcadeDB-specific and undocumented in general training data. **Justification: #1 ergonomic guardrail** (see governing principle in #4859).

## Proposed design

- Tool name: `vector_search`
- Input schema:
```json
{
  "type": "object",
  "properties": {
    "indexName": {"type": "string", "description": "LSM vector or sparse-vector index name"},
    "queryVector": {"type": "array", "items": {"type": "number"}},
    "k": {"type": "integer", "default": 10},
    "efSearch": {"type": "integer"},
    "filter": {"type": "string", "description": "optional SQL WHERE-style predicate string"},
    "sparse": {"type": "boolean", "default": false}
  },
  "required": ["database", "indexName", "queryVector", "k"]
}
```
- Output: `{results: [{rid, score, properties?}]}`
- Permission gating: requires `config.isAllowReads()` (mirrors `QueryTool.java:69`); internally builds a `vector.neighbors(...)`/`vector.sparseNeighbors(...)` SELECT and routes it through the query engine so the existing `isIdempotent()` gate still applies — do not bypass it with a raw execute path.
- `queryText` embedding generation is **out of scope for v1** — document that callers must supply a pre-computed `queryVector`.
- Engine functions used: `SQLFunctionVectorNeighbors`, `SQLFunctionVectorSparseNeighbors` (`engine/src/main/java/com/arcadedb/function/sql/vector/`).

## Acceptance criteria

- [ ] `VectorSearchTool.getDefinition()`/`execute(server, user, args, config)` follows the existing tool class pattern (Apache-2.0 header, `@author` tag, `JSONObject`/`JSONArray` builder style).
- [ ] Registered in `MCPHttpHandler` `TOOLS_LIST` + dispatch switch, and in `MCPStdioServer`.
- [ ] Rejects non-idempotent or malformed vector expressions with a clear `SecurityException`/`IllegalArgumentException`.
- [ ] Respects `allowReads=false` by returning a permission error, not a silent empty result.
- [ ] Dense and sparse index paths both covered by tests.
- [ ] Unit tests added mirroring `MCPPermissionsTest` patterns; `MCPServerPluginTest` extended for the new tool's `tools/list` entry.
- [ ] Docs updated in the MCP section of `arcadedb-docs`.

## Files to touch

`server/src/main/java/com/arcadedb/server/mcp/tools/VectorSearchTool.java` (new), `MCPHttpHandler.java`, `MCPStdioServer.java`, `server/src/test/java/com/arcadedb/server/mcp/` (new/updated tests).

## Dependencies

None (foundational). #4861 (`hybrid_search`) reuses its vector-leg logic.

## Effort estimate

M — new tool class, dimension/index validation, permission wiring, dense+sparse test coverage.

justinblethrow-cloud:
## Summary

- add `vector_search` for dense `LSM_VECTOR` and sparse `LSM_SPARSE_VECTOR` indexes
- validate index type, dimensions, sparse coordinates, `k`, and dense-only `efSearch` before querying
- return stable record payloads with explicit lower-is-better distance or higher-is-better sparse-score semantics
- support a bounded read-only SQL filter and compact sparse vectors via `queryIndices`
- register the tool centrally for both HTTP and stdio transports and avoid logging full vector payloads

## Safety

- enforces `allowReads` and the authenticated database permission path
- binds vector/index values as query parameters
- parses the generated SQL and requires the query engine to classify it as idempotent
- caps `k` and filtered candidate over-fetch at 1,000

## Verification

- focused vector/tool-list tests: 10 passed
- all `MCP*Test` server tests: 152 passed, 0 failed, 0 errors
- focused JaCoCo coverage for `VectorSearchTool`: 91.8% lines, 81.0% branches
- full engine suite not run

The MCP documentation lives in the separate `arcadedb-docs` repository, so this code PR does not edit docs in this repository.

Closes #4860

### DIFF
--- a/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
@@ -37,6 +37,7 @@ import com.arcadedb.server.mcp.tools.ServerStatusTool;
 import com.arcadedb.server.mcp.tools.SetServerSettingTool;
 import com.arcadedb.server.mcp.tools.UpsertEntityTool;
 import com.arcadedb.server.mcp.tools.UpsertRelationshipTool;
+import com.arcadedb.server.mcp.tools.VectorSearchTool;
 import com.arcadedb.server.security.ServerSecurityUser;
 
 import java.util.logging.Level;
@@ -64,6 +65,7 @@ public class MCPDispatcher {
     TOOLS_LIST.put(GetSchemaTool.getDefinition());
     TOOLS_LIST.put(QueryTool.getDefinition());
     TOOLS_LIST.put(ExecuteCommandTool.getDefinition());
+    TOOLS_LIST.put(VectorSearchTool.getDefinition());
     TOOLS_LIST.put(FullTextSearchTool.getDefinition());
     TOOLS_LIST.put(UpsertEntityTool.getDefinition());
     TOOLS_LIST.put(UpsertRelationshipTool.getDefinition());
@@ -223,6 +225,7 @@ public class MCPDispatcher {
         case "get_schema" -> GetSchemaTool.execute(server, user, args, config);
         case "query" -> QueryTool.execute(server, user, args, config);
         case "execute_command" -> ExecuteCommandTool.execute(server, user, args, config);
+        case "vector_search" -> VectorSearchTool.execute(server, user, args, config);
         case "full_text_search" -> FullTextSearchTool.execute(server, user, args, config);
         case "upsert_entity" -> UpsertEntityTool.execute(server, user, args, config);
         case "upsert_relationship" -> UpsertRelationshipTool.execute(server, user, args, config);
@@ -274,7 +277,9 @@ public class MCPDispatcher {
           sb.append(key).append("=\"").append(sanitized, 0, 100).append("...\"");
         else
           sb.append(key).append("=\"").append(sanitized).append("\"");
-      } else
+      } else if (value instanceof JSONArray array)
+        sb.append(key).append("=[").append(array.length()).append(" item(s)]");
+      else
         sb.append(key).append("=").append(value);
     }
     return sb.append("}").toString();
@@ -289,6 +294,7 @@ public class MCPDispatcher {
       case "list_databases" -> result.getJSONArray("databases", new JSONArray()).length() + " database(s)";
       case "get_schema" -> result.getJSONArray("types", new JSONArray()).length() + " type(s)";
       case "query", "execute_command" -> result.getInt("count", 0) + " record(s)";
+      case "vector_search" -> result.getInt("count", 0) + " neighbor(s)";
       case "full_text_search" -> result.getInt("count", 0) + " hit(s)";
       case "upsert_entity", "upsert_relationship" -> result.getInt("count", 0) + " record(s)";
       case "server_status" -> "ok";
--- a/server/src/main/java/com/arcadedb/server/mcp/tools/VectorSearchTool.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/tools/VectorSearchTool.java
@@ -1,0 +1,392 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.server.mcp.tools;
+
+import com.arcadedb.database.Database;
+import com.arcadedb.database.Document;
+import com.arcadedb.database.RID;
+import com.arcadedb.database.Record;
+import com.arcadedb.exception.RecordNotFoundException;
+import com.arcadedb.exception.SchemaException;
+import com.arcadedb.index.Index;
+import com.arcadedb.index.IndexInternal;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.index.sparsevector.LSMSparseVectorIndex;
+import com.arcadedb.index.vector.LSMVectorIndex;
+import com.arcadedb.query.QueryEngine;
+import com.arcadedb.query.sql.executor.Result;
+import com.arcadedb.query.sql.executor.ResultSet;
+import com.arcadedb.schema.LSMSparseVectorIndexMetadata;
+import com.arcadedb.schema.Schema;
+import com.arcadedb.serializer.JsonSerializer;
+import com.arcadedb.serializer.json.JSONArray;
+import com.arcadedb.serializer.json.JSONObject;
+import com.arcadedb.server.ArcadeDBServer;
+import com.arcadedb.server.mcp.MCPConfiguration;
+import com.arcadedb.server.security.ServerSecurityUser;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * @author Justin Blethrow
+ */
+public class VectorSearchTool {
+  private static final int DEFAULT_K             = 10;
+  private static final int MAX_K                 = 1_000;
+  private static final int FILTER_OVERFETCH      = 8;
+  private static final int MAX_FILTER_EXPRESSION = 4_096;
+
+  public static JSONObject getDefinition() {
+    return new JSONObject()
+        .put("name", "vector_search")
+        .put("description",
+            """
+            Search a dense LSM_VECTOR or sparse LSM_SPARSE_VECTOR index using a pre-computed query vector. \
+            Dense results expose a distance (lower is better); sparse results expose a score (higher is better). \
+            Set sparse=true for sparse indexes and pass queryIndices for a compact sparse representation. \
+            Embedding generation is not performed by ArcadeDB.""")
+        .put("inputSchema", new JSONObject()
+            .put("type", "object")
+            .put("properties", new JSONObject()
+                .put("database", new JSONObject()
+                    .put("type", "string")
+                    .put("description", "The name of the database to search"))
+                .put("indexName", new JSONObject()
+                    .put("type", "string")
+                    .put("description", "Name of an LSM_VECTOR or LSM_SPARSE_VECTOR index"))
+                .put("queryVector", new JSONObject()
+                    .put("type", "array")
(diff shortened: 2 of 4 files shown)


robfrank:
Reviewed against the epic #4859 acceptance criteria. Overall this is solid work: the index resolution, the dense/sparse mode mismatch hints, and the sparse validation (duplicate dimensions, out-of-range dimensions, all-zero weights, index/value length agreement) are thorough, and the error messages are written so the model can self-correct without another round trip. That matches the bar set by `FullTextSearchTool`. A few things I want addressed before merge.

### 1. Dense results label a distance as `score` (blocking)

`VectorSearchTool.appendResult` emits `score` for every hit and additionally `distance` when `!sparse`, so a dense hit carries the same number under both keys.

The two underlying functions are not symmetric:

- `SQLFunctionVectorNeighbors:304` puts `distance` (lower is better)
- `SQLFunctionVectorSparseNeighbors:352` puts `score` (higher is better)

So `score` means "higher is better" on the sparse path and "lower is better" on the dense path, in the same tool, under the same key name. An agent that ranks or thresholds on `score` will invert dense results. The envelope-level `scoring` field is good, but it does not save a model that reads the per-result key.

This is the exact trap we have already been bitten by elsewhere in the vector stack (similarity versus distance semantics), so please make it impossible rather than documented: emit `distance` only for dense and `score` only for sparse. The assertion in `vectorSearchDenseReturnsDistanceAndProperties`:

```java
assertThat(first.getDouble("score")).isEqualTo(first.getDouble("distance"));
```

currently codifies the ambiguity, so it should change with the fix.

### 2. Filter over-fetch degenerates to no over-fetch at high `k`

```java
final int candidateLimit = filter == null ? k : (int) Math.min((long) k * FILTER_OVERFETCH, MAX_K);
```

`MAX_K` is the ceiling on the *result* count, and it is being reused as the ceiling on the *candidate* count. The over-fetch therefore shrinks as `k` grows: at `k = 125` it is already clipped, and at `k = 1000` `candidateLimit == k`, meaning the filter drops candidates with zero compensation. A caller asking for 1000 filtered neighbors can silently get very few.

Two changes: give candidates their own budget constant instead of borrowing `MAX_K`, and return `candidatesScanned` (or a `truncated` flag) in the envelope. Right now a short result set is indistinguishable from "the index genuinely has no more matches", which is precisely the distinction a retrieval agent needs to decide whether to widen its query.

### 3. `filter` is a raw SQL fragment, and its evaluation context is undocumented

```java
sql.append(" WHERE (").append(filter).append(')');
```

The rest of this tool surface deliberately avoids concatenation: `MCPToolUtils.quoteIdentifier` backtick-quotes identifiers and its javadoc states that values are never routed through it and are always bound. I accept that the delta here is small, since `analyzed.isIdempotent()` blocks writes and reads are already reachable through the `query` tool, and `vectorSearchRejectsMalformedFilterWithoutExecutingWrites` covers the obvious shape.

What is missing is the contract. The predicate is evaluated against the expanded neighbor row, which carries `@rid`, `@type`, `record`, and `distance`/`score` alongside the record's own properties. None of that is in the `filter` description, so the model has to guess what it may reference. Please spell it out in the input schema description.

### 4. Every hit is read from disk twice

`vector.neighbors` already puts the resolved record into each entry (`SQLFunctionVectorNeighbors:299`, kept for backward compatibility). `appendResult` discards it and calls `database.lookupByRID(rid, true)` again, so a `k = 1000` search does 1000 avoidable record reads. Prefer the `record` already on the row and fall back to the lookup only when it is absent.

### 5. Document the non-backfill behavior

`appendResult` can skip a hit (concurrently deleted RID, non-document record, non-numeric score) and `LIMIT :k` is pushed into SQL, so a skipped hit cannot be backfilled and the tool can legitimately return fewer than `k`. `FullTextSearchTool` documents this tradeoff explicitly at its lookup site. Please mirror that comment here so the next reader does not file it as a bug.

### 6. Dense path accepts an all-zero query vector, sparse path rejects it

`buildSparseQuery` requires at least one non-zero weight on both sparse branches, but the dense branch only checks dimensionality. An all-zero dense vector has no direction, so under `COSINE` the result is not meaningful. Worth applying the same non-zero check to the dense path for consistency.

### Cross-PR: the tool-count assertions will break each other

This PR and #5398 both change `assertThat(tools.length()).isEqualTo(13)` to `isEqualTo(14)` in `MCPServerPluginTest` and `MCPStdioServerTest`. Whichever merges second leaves the other at 15 registered tools and a red build. #5402 has the same problem in a different form: its `rag` profile assertion is `containsExactlyInAnyOrder("query", "full_text_search")`, which breaks the moment `vector_search` and `sample_records` register.

Please derive the expected value from the registered tool list rather than hardcoding it, or assert containment only. That removes the merge-order coupling across all four PRs.

### Cross-PR: #5401 changes the resolver signature

#5401 replaces `MCPToolUtils.resolveDatabase(server, user, name)` with a four-argument form returning a `DatabaseAccess`, and splits it into `resolveDatabase` (no read check) and `resolveReadableDatabase` (read check). This PR will not compile after that lands. When you rebase, `vector_search` is a database-targeted read tool and must use `resolveReadableDatabase`, not `resolveDatabase`, otherwise the inline `config.isAllowReads()` check here passes on the global setting while a per-database read denial is bypassed.

justinblethrow-cloud:
Thanks, Roberto. Addressed in `0c5d58b9` and `002ab599`:

- Dense hits now expose only `distance`; sparse hits expose only `score`.
- Filter candidates have an independent 8,000-entry ceiling. The envelope reports `indexedEntries`, `candidateLimit`, and a `truncated` flag that distinguishes an exhausted small index from a capped candidate window.
- The filter schema documents the expanded-neighbor row context.
- Existing embedded records are reused with RID lookup only as fallback.
- Non-backfill behavior is documented and zero dense vectors are rejected.
- Merge-order-coupled tool-count assertions are removed.

Focused validation on the final head: `MCPServerPluginTest` + `MCPStdioServerTest`, 92 tests passed.

For the cross-PR resolver point, this branch remains sequenced before #5401. When rebased after #5401 it must use the explicit `READ` resolver mode.

robfrank:
Correction to my review above: I said the docs criterion in #4859 could not be satisfied from these PRs because no MCP documentation lives in this repo. That was wrong. The MCP reference page is at https://docs.arcadedb.com/arcadedb/reference/mcp/mcp and its source is `ArcadeData/arcadedb-docs`, file `src/main/asciidoc/reference/mcp/mcp.adoc`.

That page currently documents exactly the 13 tools registered on `main` and nothing more, so it is in sync today and a PR that adds tool 14 without touching it regresses that sync.

For this PR, please open a companion `arcadedb-docs` PR adding `vector_search` to the tool list around line 113. This one needs more prose than the others: the dense versus sparse mode switch, the fact that ArcadeDB does not generate embeddings so the caller supplies the vector, the `queryIndices` compact sparse form, `efSearch` being dense-only, and above all the result semantics once point 1 of my review is resolved, namely that dense returns a distance where lower is better and sparse returns a score where higher is better. That distinction is the single most useful thing the page can tell an integrator.

robfrank:
Thanks for turning these around so quickly. I went through `0c5d58b9` and `002ab599` and all six points from my review are addressed, and addressed well:

1. Dense and sparse results now carry mutually exclusive keys, and the tests pin both directions with `has("score")).isFalse()` / `has("distance")).isFalse()`. This was the important one.
2. `MAX_FILTER_CANDIDATES` is now an independent budget, so the over-fetch no longer collapses at high `k`.
3. The `filter` description now names the fields the predicate can reference.
4. `appendResult` reuses the embedded `record` and falls back to `lookupByRID` only when it is absent.
5. The non-backfill behavior is documented at the loop.
6. `requireNonZeroDenseVector` closes the dense/sparse asymmetry.

I have pushed one commit to your branch rather than send you round again, since it touches code you had just added. Please review it.

### What I changed and why

The truncation signal in `002ab599` was derived from `TypeIndex.countEntries()`, and that turned out to be the wrong source on two counts.

**Correctness.** `indexedEntries > candidateLimit` is true for essentially any index larger than the window, which pins `truncated` to `true` and empties it of meaning. Worse, it fires on searches that already found everything: a selective filter that returns 1 hit out of a requested 2 has exhausted its matches, and the flag was telling the caller to widen a search that cannot yield more. There is also a unit mismatch on the sparse path, where `LSMSparseVectorIndex.countEntries()` returns `engine.totalPostings()`, a count of `(dimension, RID)` postings rather than of records, so a 2-record fixture with 4 postings reports 4.

**Cost.** On the dense path `countEntries()` resolves through `LSMVectorIndex:3846` to `VectorLocationIndex.getActiveCount()` (`VectorLocationIndex:346`), which streams the entire location map filtering `!loc.deleted`, and on the bounded backend does so while holding the `locations` monitor. That ran on every single `vector_search`, so a large index paid a full scan per query and concurrent searches serialized on that monitor.

The flag now reports whether the result window was filled, which is the only state in which further matches may exist. It is O(1), correct in both modes, and `indexedEntries` is gone from the response.

I added two tests over a dedicated `McpVectorBudgetRecord` fixture holding 25 vectors, deliberately more than the 16-candidate window a `k=2` filtered search uses, so the divergence is actually observable. The first fails against `002ab599` with `Expecting value to be false but was true`, which is the false positive described above. Existing fixtures were left alone so no other test shifted.

`mvn test -Dtest='MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest'` is green at 118 tests.

### Still open on this PR

The `arcadedb-docs` companion for `vector_search` from my previous comment. The result-shape section should now describe `candidateLimit` and `truncated` with the semantics above, since those are the fields an integrator has to reason about.

Also still applies: when #5401 lands it changes the `MCPToolUtils.resolveDatabase` signature, and `vector_search` must move to `resolveReadableDatabase` rather than `resolveDatabase`, otherwise a per-database `allowReads: false` is bypassed while the global check here still passes.

justinblethrow-cloud:
I reviewed `615eeb69630a`, fast-forwarded the local branch to it, and reran:

`./mvnw -pl server -am test -Dtest='MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest' -Dsurefire.failIfNoSpecifiedTests=false`

Result: 118 tests, 0 failures.

Removing `TypeIndex.countEntries()` from the query path is the right correction. It eliminates the dense full-location scan and the sparse posting-count unit mismatch, and the new tests pin the intended result-window behavior.

One semantic caveat remains. With a filter, `truncated == false` now means fewer than `k` matches survived the current bounded candidate window. It does not prove that no matching vectors exist beyond `candidateLimit`; increasing `k` also expands that candidate window. For example, one match may occur in the first 16 candidates and another at rank 17, so a `k=2` search returns one result with `truncated=false`, while a wider request can find the second.

That is a coherent window-local contract, but it is weaker than the original distinction between “the index has no more matches” and “the candidate budget cut the search off.” I have not changed your commit. The options I see are:

1. keep the current O(1) behavior and describe `truncated` strictly as “result limit reached” (or rename it accordingly); or
2. materialize/count the already-bounded pre-filter candidate rows and report whether that candidate window filled, without consulting total index cardinality.

The companion documentation PR is ArcadeData/arcadedb-docs#426. It covers dense versus sparse mode, caller-supplied embeddings, compact `queryIndices`, dense-only `efSearch`, score/distance direction, `candidateLimit`, and the current window-local `truncated` semantics explicitly.

The #5401 resolver follow-up also remains noted for merge sequencing.

PR #5400, opened 2026-07-24
fix(fulltext) #5267: make BM25 scores comparable across buckets

### LINKED ISSUE #5267: Full-text BM25 is scored per bucket, so cross-bucket relevance ranking is unsound
## Problem

ArcadeDB's full-text BM25 is scored **per bucket**: document frequency and `totalDocs` are bucket-scoped, while average document length is type-wide. The engine documents this itself:

> `"note": "statistics are per bucket (BM25 is scored per bucket); totalDocs is this bucket's document count"`
> `engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java:547`

`SQLFunctionSearchIndex.getScoringExplain` carries a similar caveat that df/IDF differ across buckets.

**Consequence:** any consumer that merges hits from several buckets and ranks them globally is comparing scores normalized against *different corpora*. On the default multi-bucket layout, relevance ranking is therefore not strictly sound.

This is not theoretical. While building #4862 (`full_text_search` MCP tool, PR #5263) we observed a **less relevant document outranking a more relevant one**. The ranking assertion only became deterministic once the test fixture forced `BUCKETS 1`.

## Scope: pre-existing, not introduced by #4862

SQL's existing `ORDER BY $score DESC` over `SEARCH_INDEX` has had the identical exposure all along. PR #5263 neither introduces nor worsens it: its top-k push-down leaves scores unchanged, and that push-down's soundness argument holds independently of this. Filing separately so the problem is tracked on its own merits rather than buried in a feature PR.

## Why it needs a decision now

`TYPE_DEFAULT_BUCKETS` is 1, so **single-bucket fixtures will never catch this**. Any test intended to exercise cross-bucket ranking must declare `BUCKETS n` explicitly and assert the spread via `RID.getBucketId()`. That is a blind spot in the current full-text test suite.

More urgently: **#4861 (`hybrid_search`) will fuse these per-bucket BM25 scores with vector similarity scores.** That would bake the unsoundness into a second surface, where it is considerably harder to unwind than it is today. This should be resolved, or consciously accepted, before #4861 starts.

## Candidate directions (undecided - input wanted)

1. **Type-wide corpus statistics.** Compute `df` / `totalDocs` across all buckets of the type rather than per bucket, so IDF is comparable. Makes scores globally meaningful, but the counters become a cross-bucket shared structure, with the attendant maintenance and concurrency cost on the write path.
2. **Global re-scoring pass over the merged candidate set.** Keep per-bucket matching, then re-score the union against type-wide statistics before ranking. Cheaper on the write path; adds a pass on the read path.
3. **Accept and document.** Declare cross-bucket relevance ordering explicitly best-effort, and have consumers needing strict ranking pin `BUCKETS 1`. Cheapest, but it makes "ranked by relevance" a qualified claim across SQL, `full_text_search`, and `hybrid_search`.

No strong opinion between (1) and (2). Option (3) looks hard to live with given the GraphRAG positioning of epic #4859. Opinions welcome.

## References

- `engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java:542-547` (per-bucket statistics note)
- `engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchIndex.java` (`getScoringExplain` df/IDF caveat)
- Surfaced in PR #5263 (issue #4862), epic #4859. Blocks #4861.

justinblethrow-cloud:
## Summary

- compute Best Matching 25 (BM25) document frequencies once across every bucket behind a logical full-text type index
- preserve bucket-local Lucene matching semantics, then rescore candidates with one type-wide corpus context
- apply globally comparable scores to `SEARCH_INDEX`, `SEARCH_FIELDS`, and logical `TypeIndex.get()` lookups
- report type-wide BM25 statistics through SQL `EXPLAIN`/`PROFILE`
- preserve the existing single-bucket path and avoid persisted-format or write-path changes

## Correctness

The regression reproduces the issue with two equal-sized buckets and equal-length documents. Before this change, a one-term match from a bucket where that term was locally rare scored `1.9924302`, while a two-term match from the other bucket scored only `0.093040034`. With type-wide N and document frequencies, the two-term document ranks correctly.

The same regression also verifies bucket-specific wildcard expansions are unioned for global scoring, while Boolean matching remains document-local.

## Performance

Multi-bucket BM25 queries now pay a read-time token-discovery/document-frequency pass before bucket-local candidate scoring. This deliberately trades additional read work for correct cross-bucket ranking without shared write counters or synchronization. Single-bucket queries retain the prior direct path.

## Validation

- 220 focused engine full-text tests: passed
- 14 focused server MCP full-text tests: passed
- `FullTextBM25Test` after the wildcard/Boolean regression extension: 30 passed
- full engine suite not run

Closes #5267

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchFields.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchFields.java
@@ -22,13 +22,10 @@ import com.arcadedb.database.Database;
 import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.exception.CommandExecutionException;
-import com.arcadedb.index.Index;
-import com.arcadedb.index.IndexCursor;
 import com.arcadedb.index.IndexCursorEntry;
 import com.arcadedb.index.TempIndexCursor;
 import com.arcadedb.index.TypeIndex;
-import com.arcadedb.index.fulltext.FullTextQueryExecutor;
-import com.arcadedb.index.fulltext.LSMTreeFullTextIndex;
+import com.arcadedb.index.fulltext.FullTextSearch;
 import com.arcadedb.query.sql.executor.CommandContext;
 import com.arcadedb.function.sql.SQLFunctionAbstract;
 import com.arcadedb.schema.DocumentType;
@@ -104,7 +101,7 @@ public class SQLFunctionSearchFields extends SQLFunctionAbstract {
 
     // Try to get cached results
     @SuppressWarnings("unchecked")
-    Map<RID, Integer> allResults = (Map<RID, Integer>) iContext.getVariable(cacheKey);
+    Map<RID, Float> allResults = (Map<RID, Float>) iContext.getVariable(cacheKey);
 
     if (allResults == null) {
       allResults = new HashMap<>();
@@ -126,19 +123,9 @@ public class SQLFunctionSearchFields extends SQLFunctionAbstract {
       if (matchingIndex == null)
         throw new CommandExecutionException("No full-text index found for fields: " + fieldNames);
 
-      // Execute search using the found index
-      for (final Index bucketIndex : matchingIndex.getIndexesOnBuckets()) {
-        if (bucketIndex instanceof LSMTreeFullTextIndex) {
-          final FullTextQueryExecutor executor = new FullTextQueryExecutor((LSMTreeFullTextIndex) bucketIndex);
-          final IndexCursor cursor = executor.search(queryString, -1);
-
-          while (cursor.hasNext()) {
-            final Identifiable match = cursor.next();
-            final int score = cursor.getScore();
-            allResults.merge(match.getIdentity(), score, Integer::sum);
-          }
-        }
-      }
+      // Route through the same type-wide scorer as SEARCH_INDEX so BM25 floats retain their precision and remain comparable
+      // when the logical index spans several buckets.
+      allResults.putAll(FullTextSearch.search(matchingIndex, queryString, -1));
 
       // Cache the results
       iContext.setVariable(cacheKey, allResults);
@@ -152,8 +139,7 @@ public class SQLFunctionSearchFields extends SQLFunctionAbstract {
       if (matches) {
         // Store the score for this record in the context variable $score
         // This allows $score projection to work in SELECT
-        final int recordScore = allResults.get(rid);
-        iContext.setVariable("$score", (float) recordScore);
+        iContext.setVariable("$score", allResults.get(rid));
       } else {
         // Clear the score for non-matching records
         iContext.setVariable("$score", 0f);
@@ -164,10 +150,10 @@ public class SQLFunctionSearchFields extends SQLFunctionAbstract {
 
     // Return cursor with all results
     final List<IndexCursorEntry> entries = new ArrayList<>();
-    for (final Map.Entry<RID, Integer> entry : allResults.entrySet()) {
+    for (final Map.Entry<RID, Float> entry : allResults.entrySet()) {
       entries.add(new IndexCursorEntry(new Object[] { queryString }, entry.getKey(), entry.getValue()));
     }
-    entries.sort((a, b) -> Integer.compare(b.score, a.score));
+    entries.sort((a, b) -> Float.compare(b.floatScore, a.floatScore));
 
     return new TempIndexCursor(entries);
   }
--- a/engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchIndex.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/text/SQLFunctionSearchIndex.java
@@ -29,9 +29,7 @@ import com.arcadedb.index.Index;
 import com.arcadedb.index.IndexCursorEntry;
 import com.arcadedb.index.TempIndexCursor;
 import com.arcadedb.index.TypeIndex;
-import com.arcadedb.index.fulltext.FullTextQueryExecutor;
 import com.arcadedb.index.fulltext.FullTextSearch;
-import com.arcadedb.index.fulltext.LSMTreeFullTextIndex;
 import com.arcadedb.query.sql.executor.CommandContext;
 import com.arcadedb.query.sql.executor.IndexableSQLFunction;
 import com.arcadedb.query.sql.parser.BinaryCompareOperator;
@@ -39,7 +37,6 @@ import com.arcadedb.query.sql.parser.Expression;
 import com.arcadedb.query.sql.parser.FromClause;
 import com.arcadedb.function.sql.SQLFunctionAbstract;
 import com.arcadedb.schema.Schema;
-import com.arcadedb.serializer.json.JSONObject;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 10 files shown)


robfrank (commented):
## Review

Rebased onto `main` (was 95 commits behind) and force-pushed; the rebase was clean. The only overlap with `main` was `MCPServerPluginTest.java`, which the three new MCP commits also touched, but this PR's change there is comment-only so there is no semantic conflict.

Verified on the rebased head: `mvn -pl engine test -Dtest='FullText*Test'` gives **142 tests, 0 failures**.

The core approach is right, and I want to say so before the findings: matching stays bucket-local (so Lucene boolean/phrase/wildcard semantics are preserved) and only the *scoring* is lifted to a type-wide corpus. That is the correct split, and doing it with a read-time context instead of shared write-path counters is the right trade for #5267. The wildcard-union and boolean-locality assertions added to `FullTextBM25Test` are good additions.

Two things need attention before merge.

---

### 1. Blocking: reachable `IllegalArgumentException` on a corpus with no analyzable tokens

`BM25ScoringContext`'s compact constructor rejects `avgDocLength <= 0`:

```java
if (!Double.isFinite(avgDocLength) || avgDocLength <= 0)
  throw new IllegalArgumentException("avgDocLength must be finite and positive");
```

but `createScoringContext` feeds it `getTypeWideAverageDocumentLength()` unguarded, while it does clamp the other counter with `Math.max(1L, ...)`.

`avgDocLength()` returns `sumDocLength / totalDocs` whenever `totalDocs > 0`. `computeCorpusCounters` increments `docs` for **every** document of the type (`++docs` is unconditional) but only adds the analyzed token count to `sumLen`. So a type whose full-text-indexed properties yield zero tokens in every document (all `null`, empty, or stopword-only) lands on `totalDocs > 0, sumDocLength == 0` and therefore `avgDocLength() == 0.0`, and the constructor throws.

Reproduced against the rebased head with a 4-document type whose `content` is `''`:

```
java.lang.IllegalArgumentException: avgDocLength must be finite and positive
    at com.arcadedb.index.fulltext.BM25ScoringContext.<init>(BM25ScoringContext.java:31)
    at com.arcadedb.index.fulltext.FullTextSearch.createScoringContext(FullTextSearch.java:255)
    at com.arcadedb.index.fulltext.FullTextSearch.searchBM25(FullTextSearch.java:123)
    at com.arcadedb.index.fulltext.FullTextSearch.search(FullTextSearch.java:107)
```

The blast radius is wider than multi-bucket `SEARCH_INDEX`. Because `searchSimple` calls `createScoringContext` unconditionally (no `bucketIndexes.size() > 1` guard), the new `TypeIndex.get()` branch makes `CONTAINSTEXT` throw at **any** bucket count, including `BUCKETS 1`:

```
    at com.arcadedb.index.fulltext.FullTextSearch.createScoringContext(FullTextSearch.java:255)
    at com.arcadedb.index.fulltext.FullTextSearch.searchSimple(FullTextSearch.java:148)
    at com.arcadedb.index.TypeIndex.get(TypeIndex.java:111)
```

Both are regressions. Pre-change, both paths returned an empty result set: `df == 0` short-circuits with `continue` before `avgdl` is ever divided by. I confirmed the single-bucket `SEARCH_INDEX` path (which still bypasses the new code) returns 0 rows cleanly on the same fixture.

Suggested fix, consistent with the fallback `resolveAvgDocLength()` already uses:

```java
return new BM25ScoringContext(Math.max(1L, first.getTypeWideDocumentCount()),
    Math.max(1.0, first.getTypeWideAverageDocumentLength()), documentFrequencies);
```

Clamping at the boundary keeps the record's invariant meaningful rather than relaxing the validation. Worth a regression test for the tokenless corpus on both entry points.

---

### 2. Multi-bucket BM25 now makes three posting-list passes per token instead of two

Per token, per bucket, `searchBM25` walks the postings three times:

1. `createScoringContext` - full scan of every token's postings in every bucket to count `df`
2. `collectBM25Matches` -> `collectQueryMatches` -> `collectTermMatches` - full scan to build the candidate set
3. `scoreBM25WithContext` -> `computeBM25Scores` (candidate path) - full scan again to collect `hits`

Before this change the candidate path counted `df` inside the scoring scan ("Single pass: count df while collecting only the candidate postings"), so the cost was two passes. For wildcard/fuzzy queries there is a fourth: `collectScoringTokens` runs `collectMatches` in `tokensOnly` mode per bucket, and while `tokensOnly` returns early for a plain `TermQuery`, the wildcard path at `collectWildcardMatches` still iterates the whole key range before checking the flag.

The `df` pass genuinely has to precede scoring, but it does not need its own scan: pass 2 already visits exactly the same postings. Counting `df` per token during candidate collection across all buckets, then scoring in pass 3, gets global `df` at the pre-change two-pass cost. Given how load-bearing "performance is the mantra" is here, that seems worth doing rather than accepting the extra scan.

Also, `collectScoringTokens` constructs a new `FullTextQueryExecutor` and re-parses the query once per bucket, and `searchBM25` then parses it again per bucket in `collectBM25Matches`. The parse is cheap next to the scans, but the query text is identical across buckets so it could be parsed once.

---

### Smaller points

- **`BM25QueryMatch.scoringTokens` is dead.** `collectBM25Matches` builds it with `Map.copyOf(scoringTokens)`, but `searchBM25` passes the global `scoringTokens` union to `scoreBM25WithContext` and never reads `match.scoringTokens()`. Nothing in `main/` calls it. Either drop the component or use it.

- **The two token caps use different rules.** `FullTextSearch.collectScoringTokens` trims the union with `tokens.pollLastEntry()`, dropping the alphabetically-last terms; `FullTextQueryExecutor.recordScoringToken` caps at `MAX_EXPANDED_SCORING_TERMS` (4096) in insertion order. A document matched by a token that survived the per-bucket cap but was dropped from the global union contributes nothing to its score, so it surfaces as a match with score 0. Only reachable past 4096 expanded terms, but the two caps should agree.

- **`explainScoring` pays a second full `df` pass** via its own `createScoringContext(bucketIndexes, shownTokens)`. Fine for an interactive diagnostic, just noting it is not free.

- **Scope note for the description:** `TypeIndex.get()` now routes through `searchSimple` for every BM25 full-text index, not only multi-bucket ones, so `CONTAINSTEXT` results are now globally ranked and limit-truncated there too. That looks intentional, but the PR body reads as multi-bucket-only and this is the change that widened finding 1 to single-bucket types.

---

Happy to re-review once the `avgDocLength` clamp and a regression test are in. Finding 2 I would like addressed too, but I am open to it landing as a follow-up if you would rather keep this PR focused on correctness - say which you prefer.

robfrank:
hi @justinblethrow-cloud  thank you very much for all the work you have done! I was mostly off last week, sorry for the delay.
Beside that, if you want to share with us what are you doing with ArcadeDB it would be great. Just drop an email to info at arcadedb.com :)

robfrank:
Pushed `f87752e` addressing the review rather than leaving it to another round trip.

**1. The `avgDocLength` crash.** Clamped to `Math.max(1.0, ...)` in `createScoringContext`, matching the neutral fallback `resolveAvgDocLength()` already uses. Two regression tests cover the tokenless corpus, one per entry point, since `searchSimple` has no bucket-count guard and so broke `CONTAINSTEXT` on single-bucket types too.

**2. The extra posting pass.** Folded document-frequency counting into the matching walk instead of giving it a scan of its own. `collectTermMatches`, `collectPhraseMatches` and `iterateAndMatch` already visit exactly the postings whose frequencies we need, so they now count while they match; `BM25QueryMatch` reports the per-bucket counts and `searchBM25` sums them into the type-wide context.

Measured with a temporary counter on every posting-list walk, a two-term query over two buckets:

| | walks |
|---|---|
| `e122a42` (before) | 12 |
| `f87752e` (after) | 8 |

That is the three-passes-per-token-per-bucket dropping to two, which is what the pre-#5267 code cost. The dedicated scan is kept for the two entry points with no matching pass to fold into (`searchSimple` and `explainScoring`), where it was never a regression.

Repeated terms and overlapping wildcard ranges (`ap*` with `app*`) can walk the same token twice, so `recordDocumentFrequency` merges with `max` rather than summing. Deletion markers are filtered on `bucketId >= 0`, matching the existing df semantics.

Because these are now two independent derivations of the same statistic, I added `matchTimeAndScanDerivedDocumentFrequenciesAgree`: for a single literal term over three buckets, `SEARCH_INDEX` (match-time counting) and `TypeIndex.get()` (scan) must return byte-identical scores. They do.

**3. Smaller items.** `BM25QueryMatch.scoringTokens` is no longer dead - it carries the union that drives scoring. The union cap now keeps the first terms discovered, matching how `recordScoringToken` caps a single bucket, instead of dropping the alphabetically-last ones. The per-bucket `new Object[] {}` allocation became a shared constant.

**Verification.** 218 engine full-text / search / `CONTAINSTEXT` / Lucene tests pass, plus 90 server tests including `MCPServerPluginTest`. The only failures in the wider engine run are three `SQLVectorHybridSearchBlogPostTest` errors from a GraalVM polyglot classloading problem; I confirmed those reproduce identically on the base commit without these changes.

One trade-off worth naming: pass two cannot start until every bucket has reported its share of each frequency, so the per-bucket candidate sets are now all held at once instead of one at a time. The merged result map already held every matching RID, so this is roughly one extra reference per match, not an extra copy of the documents.

Also rebased onto current `main` before all of this. Let me know if you would rather I split the performance change into its own PR.

PR #5401, opened 2026-07-24
feat(mcp): scope permissions per database

### LINKED ISSUE #4868: MCP: per-database endpoint/config scoping
Part of #4859.

## Motivation

`allowedUsers` (and all other permission flags) are server-global today (`MCPConfiguration` is a single server-wide instance). Multi-tenant agent deployments need to bind an MCP surface to one database with its own permission set. **Justification: #3 different permissioning.**

## Proposed design (conservative v1 scope)

- Extend `MCPConfiguration` with an optional per-database override map (e.g. `databases: {"<dbName>": {allowedUsers, allowReads, allowInsert, ...}}`), consulted by tool `execute()` methods alongside `MCPToolUtils.resolveDatabase(server, user, dbName)`.
- **Do not** attempt new HTTP routing (e.g. `/mcp/{database}`) in v1 — flag that as a larger architectural change needing its own design pass; note it explicitly as a candidate follow-up, not silently dropped.
- This issue carries the most architectural uncertainty in the set — recommend a short design note/spike before implementation rather than starting directly from this issue body.

## Acceptance criteria

- [ ] Per-database override resolved correctly when present; falls back to server-global config when absent (backward-compatible).
- [ ] Override respected by every existing and new tool (audit all tool permission-check call sites, not just new ones).
- [ ] Tests cover: no override (current behavior), override present (stricter), override present (more permissive than global — decide and document whether this is allowed or capped by the global config as a ceiling).
- [ ] Docs describe the override format and multi-tenant use case; explicitly note HTTP per-database routing is out of scope for this issue.

## Files to touch

`MCPConfiguration.java`, `MCPToolUtils.java`, every tool class's permission-check call site, tests, docs.

## Dependencies

None (independent), but highest uncertainty — recommend last in sequence.

## Effort estimate

L — audit surface across all tools, open design question on override-vs-ceiling semantics, no routing changes included.

justinblethrow-cloud:
## Summary

- add optional per-database Model Context Protocol (MCP) permission overrides
- resolve one immutable effective permission snapshot for every database-targeted operation
- enforce database-local read, write, administrative, and user restrictions across tools, resources, discovery, and status
- document the configuration format, global-ceiling semantics, and HTTP routing boundary

## Permission model

Database overrides can only restrict the server-global policy:

- local `false` values deny operations that are globally allowed
- local `true` values cannot grant operations denied globally
- local `allowedUsers` is an additional restriction on the global allowlist
- omitted fields inherit their global values
- databases without overrides preserve existing behavior

Unknown database override fields are rejected before any part of an update is applied, preventing misspelled security settings from silently inheriting a more permissive value.

Example:

```json
{
  "enabled": true,
  "allowReads": true,
  "allowInsert": true,
  "allowedUsers": ["root", "tenant-agent"],
  "databases": {
    "tenant_graph": {
      "allowInsert": false,
      "allowedUsers": ["tenant-agent"]
    }
  }
}
```

This does not add `/mcp/{database}` or another database-bound HTTP route. The existing endpoint and origin policy remain server-global.

## Validation

- focused configuration, permission, resource, and database-scoping tests: 42 passed
- complete MCP-only server test set: 151 passed
- no failures or skipped tests

Closes #4868

### DIFF
--- a/docs/superpowers/specs/2026-07-24-mcp-database-scoping-4868-design.md
+++ b/docs/superpowers/specs/2026-07-24-mcp-database-scoping-4868-design.md
@@ -1,0 +1,104 @@
+# MCP per-database configuration scoping (#4868)
+
+**Issue:** [#4868](https://github.com/ArcadeData/arcadedb/issues/4868)
+**Epic:** [#4859 - MCP GraphRAG & Agent-Memory Surface](https://github.com/ArcadeData/arcadedb/issues/4859)
+**Date:** 2026-07-24
+**Status:** implemented
+
+## Decision
+
+Database overrides are restrictions on the server-global Model Context Protocol (MCP)
+configuration. The global configuration is a hard permission ceiling:
+
+- A local `false` denies an operation allowed globally.
+- A local `true` cannot grant an operation denied globally.
+- A local `allowedUsers` entry must also pass the global `allowedUsers` list.
+- An omitted local field inherits the corresponding global value.
+- A database with no override retains the existing global behavior.
+
+This is the conservative choice for a server-wide endpoint. A root administrator can reason
+about the global settings as the maximum authority the endpoint can ever exercise, and adding
+an override cannot accidentally widen that authority.
+
+## Configuration
+
+Overrides live under `databases` in `config/mcp-config.json`:
+
+```json
+{
+  "enabled": true,
+  "allowReads": true,
+  "allowInsert": true,
+  "allowUpdate": true,
+  "allowDelete": false,
+  "allowSchemaChange": false,
+  "allowAdmin": false,
+  "allowedUsers": ["root", "tenant-agent"],
+  "databases": {
+    "tenant_graph": {
+      "allowInsert": false,
+      "allowUpdate": false,
+      "allowedUsers": ["tenant-agent"]
+    }
+  }
+}
+```
+
+For `tenant_graph`, the example permits reads to `tenant-agent` but denies writes. `root`
+remains globally authorized but is excluded by the database allowlist. Other databases inherit
+the global policy.
+
+The supported override fields are:
+
+- `allowReads`
+- `allowInsert`
+- `allowUpdate`
+- `allowDelete`
+- `allowSchemaChange`
+- `allowAdmin`
+- `allowedUsers`
+
+Unknown override fields are rejected so a misspelled security setting cannot silently inherit a
+more permissive global value. An explicit `allowedUsers: null` is an empty local allowlist,
+matching the existing global configuration behavior. API token names retain the existing bare
+token matching behavior.
+
+## Enforcement
+
+`MCPConfiguration.getPermissionsForDatabase()` returns one immutable effective-permission
+snapshot. Database-targeted tools resolve that snapshot together with the authenticated
+database:
+
+- `get_schema`
+- `query`
+- `execute_command`
+- `full_text_search`
+- `upsert_entity`
+- `upsert_relationship`
+
+Database discovery in `list_databases`, `server_status`, and schema resources omits databases
+whose effective policy denies the user or read access. ArcadeDB's native database authorization
+remains an independent required check; MCP configuration does not replace it.
+
+Future database-targeted tools must use the same resolver so they cannot bypass database-local
+read, write, or user restrictions.
+
+Tools without a database argument remain server-global and continue to use only global
+permissions: profiler controls, server settings, and server-level status details. Per-database
+`allowAdmin` still applies when `execute_command` analyzes a database command as an
+administrative operation.
+
+## Out of scope
+
+- No `/mcp/{database}` HTTP route is added. A database-bound endpoint needs a separate design
+  for transport configuration, discovery, initialization, and tool-list behavior.
+- Tool profiles and browser-origin controls remain server-global because the current endpoint
+  advertises one tool surface and has one HTTP origin policy.
+- Overrides do not grant authority above the global ceiling.
+
+## Validation
+
+Focused tests cover persistence, inheritance, stricter overrides, attempted grants above the
+global ceiling, user restrictions, API-token matching, invalid settings, all database read
+surfaces, command/upsert writes, resource discovery, database listing, and server-status
+filtering.
--- a/docs/superpowers/specs/2026-07-24-mcp-database-scoping-4868-design.md
+++ b/docs/superpowers/specs/2026-07-24-mcp-database-scoping-4868-design.md
@@ -63,6 +63,13 @@ more permissive global value. An explicit `allowedUsers: null` is an empty local
 matching the existing global configuration behavior. API token names retain the existing bare
 token matching behavior.
 
+Configuration API updates merge the `databases` object by database name: an update to one
+database leaves other overrides intact. A supplied database object replaces that database's
+override, an explicit null value removes that one override, and `"databases": null` clears all
+database overrides. The serialized configuration omits `databases` when no overrides exist.
+At startup, an override whose database is not currently loaded produces a warning but remains
+valid so it can apply if the database is created later.
+
 ## Enforcement
 
 `MCPConfiguration.getPermissionsForDatabase()` returns one immutable effective-permission
@@ -80,8 +87,9 @@ Database discovery in `list_databases`, `server_status`, and schema resources om
 whose effective policy denies the user or read access. ArcadeDB's native database authorization
 remains an independent required check; MCP configuration does not replace it.
 
-Future database-targeted tools must use the same resolver so they cannot bypass database-local
-read, write, or user restrictions.
+Future database-targeted tools must use the same resolver and explicitly select `READ` or
+`ACCESS` as their required access. There is no default resolver mode, so a new read tool cannot
+silently bypass database-local read restrictions.
 
 Tools without a database argument remain server-global and continue to use only global
 permissions: profiler controls, server settings, and server-level status details. Per-database
--- a/server/src/main/java/com/arcadedb/server/ArcadeDBServer.java
+++ b/server/src/main/java/com/arcadedb/server/ArcadeDBServer.java
@@ -312,6 +312,7 @@ public class ArcadeDBServer {
     // INITIALIZE MCP CONFIGURATION (always available, disabled by default)
     mcpConfiguration = new MCPConfiguration(serverRootPath);
     mcpConfiguration.load();
+    mcpConfiguration.warnUnknownDatabaseOverrides(getDatabaseNames());
 
     // INITIALIZE AI CONFIGURATION (always available, inactive until subscription token is set)
     aiConfiguration = new AiConfiguration(serverRootPath);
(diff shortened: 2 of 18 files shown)


robfrank:
Reviewed against the epic #4859 acceptance criteria. This is the strongest of the four MCP PRs. #4859 asked for a design spike first on #4868 and you delivered one; the restriction-only ceiling model (a local `true` can never widen a global `false`) is the right call for a server-wide endpoint, `updateFrom` parses the overrides before mutating any field so a bad payload cannot half-apply, unknown override keys are rejected rather than silently inherited, and extracting `MCPPermissions` so `ExecuteCommandTool.checkPermission` operates on a snapshot instead of the live config is a clean refactor. `MCPResources`, `ListDatabasesTool` and `ServerStatusTool` were all updated together, so the discovery surfaces do not leak databases the effective policy denies. The test matrix covers inheritance, attempted grants above the ceiling, user restriction, and API-token matching.

Four things to address.

### 1. The safe resolver has the longer name (blocking)

```java
public static DatabaseAccess resolveDatabase(server, user, databaseName, config)          // no read check
public static DatabaseAccess resolveReadableDatabase(server, user, databaseName, config)  // read check
```

Your own spec says "Future database-targeted tools must use the same resolver so they cannot bypass database-local read, write, or user restrictions." But the obvious, shorter, previously-existing name is the one that skips the read check. A new read tool that reaches for `resolveDatabase` out of habit compiles, passes its own inline `config.isAllowReads()` check against the global setting, and silently ignores a per-database `allowReads: false`.

This is not hypothetical: #5398 and #5399 are both database-targeted read tools currently written against the old signature, and both will be rebased onto this by hand.

Please make the failure mode loud. Either rename to `resolveDatabaseForRead` / `resolveDatabaseForWrite` so neither is the default-looking choice, or keep one entry point that takes the required access as an explicit argument.

### 2. `databases` replaces, every other field merges

`updateFrom` treats scalar fields as merge-if-present, but `databases` is wholesale replacement:

```java
final Map<String, DatabaseOverride> updatedDatabaseOverrides = json.has("databases")
    ? parseDatabaseOverrides(json.getJSONObject("databases", null))
    : databaseOverrides;
```

So `POST {"databases": {"tenant_a": {...}}}` silently deletes the override for `tenant_b`. Since these are security restrictions, silently dropping one is a fail-open change: `tenant_b` reverts to the global policy. An operator editing one tenant through the config endpoint would have no reason to expect that.

The spec does not mention it either. Either merge per database key (with an explicit `null` value meaning remove), or document the replace-all semantics prominently in the spec and in the endpoint's error surface.

### 3. Per-database snapshot allocation on the listing paths

`getPermissionsForDatabase` is `synchronized` and allocates an `EffectivePermissions` record plus one or two `List.copyOf` calls on every invocation. It is now called once per database inside `removeIf` in `ListDatabasesTool`, `ServerStatusTool` and `MCPResources.list`, and again for every database in `resolveDatabase`'s not-found path.

For a server with a few hundred databases, one `list_databases` call becomes a few hundred monitor acquisitions and twice as many list copies, for a result that is identical for every database when no overrides are configured. Given how much of the codebase is written around allocation pressure, this is worth tightening before it ships.

Cheapest fix: short-circuit at the top of `getPermissionsForDatabase` and return `this` when `databaseOverrides.isEmpty()`, which is the overwhelmingly common case and is already valid since `MCPConfiguration implements MCPPermissions`. Beyond that, the global user list does not need copying into every snapshot.

### 4. Two smaller points

`toJSON()` now unconditionally writes `"databases": {}`, so every existing installation's `mcp-config.json` gains an empty object on the next save. Harmless but noisy; consider omitting when empty.

`parseDatabaseOverrides` rejects blank database names, which is good, but nothing checks that an override names a database that exists. A typo produces an override that never applies, which is the same silent-misconfiguration class that the unknown-key rejection was added to prevent. A warning at load time (not a hard failure, since databases can be created later) would close it.

### Cross-PR sequencing

#5402 touches `MCPConfiguration.load()`, `updateFrom()` and `toJSON()` in the same hunks this PR does, and both add `assertThatThrownBy` plus assertions to `MCPConfigurationTest.saveAndLoad` and `toJSON`. A conflict is guaranteed. Both use the identical "parse and validate before mutating" pattern, so the merge is mechanical, but whoever goes second should re-run `MCPConfigurationTest` rather than trusting the resolution.

#5398 and #5399 will both fail to compile after this lands, per point 1 above. Given this PR is the one #4859 scheduled last, it may be less disruptive to land those two first and rebase this once, rather than rebasing both of them onto this.

justinblethrow-cloud:
Thanks, Roberto. Addressed in `299350ef`:

- Database resolution now has one entry point with an explicit `READ` or `ACCESS` requirement; every current database-targeted tool selects one.
- Configuration API updates merge overrides by database key; per-key `null` removes one override and top-level `databases: null` clears all.
- The common no-override lookup returns the live configuration without synchronization or allocation; override snapshots reuse the global user list.
- Empty `databases` is omitted from serialization and unknown database names produce a startup warning while remaining valid for later-created databases.
- The design note and tests cover the merge/removal contract.

Focused validation: `MCPConfigurationTest`, `MCPDatabaseScopingTest`, `MCPServerPluginTest`, and `MCPStdioServerTest`; 101 tests passed.

robfrank:
Correction to my review above: I did not raise the docs criterion from #4859 because I had wrongly concluded no MCP documentation lives in this repo. The MCP reference page is at https://docs.arcadedb.com/arcadedb/reference/mcp/mcp and its source is `ArcadeData/arcadedb-docs`, file `src/main/asciidoc/reference/mcp/mcp.adoc`.

That page documents the `mcp-config.json` keys as exactly `enabled`, `allowReads`, `allowInsert`, `allowUpdate`, `allowDelete`, `allowSchemaChange`, `allowAdmin`, `allowedUsers`, `allowedOrigins`. This PR adds a tenth top-level key, `databases`, which is a security control, so leaving it undocumented is worse than leaving a tool undocumented.

Please open a companion `arcadedb-docs` PR covering the `databases` key. The design spec in this PR is already most of the text; the parts that most need to reach operators are the ceiling semantics (a local `true` can never widen a global `false`), the inheritance rule for omitted fields, the rejection of unknown override keys, and the replace-all versus merge behavior of the `databases` key itself once you have settled point 2 of my review.

Separately, and not this PR's debt: the page does not mention MCP Resources at all, so the `arcadedb://{database}/schema` resource from #4865 is already shipped and undocumented. Worth a follow-up issue.

justinblethrow-cloud:
Companion documentation is now open as ArcadeData/arcadedb-docs#427.

It covers the `databases` key, restriction-only ceiling semantics, inheritance, native authorization, unknown-key rejection, and the finalized runtime behavior: merge by database key, replace a non-null database override, remove one with `null`, or clear all overrides with top-level `databases: null`.

- `python3 docs-validator.py`
- `./mvnw generate-resources`

justinblethrow-cloud:
Updated for the now-merged Wave 1 tools. I merged current `main` and added the missing database-scoping integration for `sample_records`: it now resolves through the explicit per-database `READ` path, and `MCPDatabaseScopingTest` verifies a database-local read denial reaches it before sampling. The branch is at a0e28764.\n\nFocused verification: `MCPDatabaseScopingTest,MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest`; 135 tests passed, 0 failures.

robfrank:
Tool profiles (#5402) landed, which put this in conflict on `MCPConfiguration`. I have merged `main` in and pushed, so this is mergeable again. This is the last of the four MCP PRs.

### On your follow-up commits

All four of my review points are addressed, and two of them better than I proposed:

**The resolver naming trap is gone entirely.** Rather than keeping `resolveDatabase` and `resolveReadableDatabase` and hoping callers pick correctly, you collapsed them into one entry point taking an explicit `RequiredAccess`. There is now no default mode, so a new read tool cannot silently inherit the unchecked path. That is the stronger of the two options I offered.

**`databases` now merges per key**, with an explicit `null` removing a single entry and a null `databases` clearing the map. That closes the fail-open case where updating one tenant silently reverted every other tenant to the global policy.

**The per-database snapshot allocation is gone** on the common path: `getPermissionsForDatabase` returns the configuration itself when a database has no override, and `databaseOverrideInheritsUnspecifiedGlobalValues` pins it with `isSameAs(config)`. Listing a hundred databases no longer allocates a hundred snapshots and two hundred list copies.

**`toJSON` omits `databases` when empty**, so existing installations do not gain a stray object on their next save.

Beyond that, `695e230c4` and `a0e287641` wired `vector_search` and `sample_records` into the resolver as they merged. That was the cross-PR risk I raised on both of those PRs, and you closed it without being asked.

I audited all eight tools that accept a `database` argument. Every one resolves through the scoped resolver: `get_schema`, `query`, `sample_records`, `vector_search` and `full_text_search` declare `READ`; `execute_command`, `upsert_entity` and `upsert_relationship` declare `ACCESS` and rely on the operation-type check against the same snapshot. No gaps.

### What I changed on top

The conflict was confined to `MCPConfiguration`, where profiles and per-database overrides each added a field plus `load`, `updateFrom` and `toJSON` handling. Both are kept everywhere they collided.

One resolution was not mechanical: the two sides disagreed on how `allowAdmin` is read, because #5402 introduced strict `booleanValue()` validation after this branch was written. I took main's version, so `{"allowAdmin": "yes"}` is still rejected with a typed error naming the field rather than silently coerced. Worth a glance to confirm you agree.

I also refreshed the spec's enforcement list. It named six database-targeted tools while the code now scopes eight, because `sample_records` and `vector_search` were wired up after the spec was written. Since that list is normative for the "future database-targeted tools must use the same resolver" rule, a stale list is the thing most likely to let the next tool slip through. It now records the access each tool declares, and why the write tools ask for `ACCESS` rather than `READ`.

`mvn test -Dtest='MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest,MCPDatabaseScopingTest'` is green at 143 tests.

### Two things outstanding, neither blocking this PR

**The per-agent profile question.** With this merged, the configuration resolves permissions per database and per principal, but the tool profile is still one server-wide value. That is the natural place to revisit #4859's "the surface a given agent sees" criterion, since the plumbing now exists. A follow-up issue rather than more scope here.

**Docs.** The `arcadedb-docs` gap is now four items: `sample_records`, `vector_search`, `profile` and `databases`. Three of the four have shipped to main already. The `databases` key is the one I would prioritise, because it is a security control and an operator who cannot find it in the reference page will not know the ceiling semantics exist. Happy to take that one myself if you would rather close out the code.

justinblethrow-cloud:
Thanks, Roberto. I reviewed the merge resolution and agree with both judgment calls:\n\n- Keeping strict `booleanValue()` handling for `allowAdmin` is correct and preserves the typed, atomic validation introduced by #5402 across all boolean fields.\n- The refreshed eight-tool enforcement table is accurate, including `READ` for the five retrieval tools and `ACCESS` plus operation-specific checks for the three write-capable tools.\n\nBoth non-blocking follow-ups are already in place: per-principal profiles are tracked in #5445, and the four companion documentation PRs are ArcadeData/arcadedb-docs#425 (`sample_records`), #426 (`vector_search`), #427 (the prioritized `databases` security control), and #428 (`profile`). Each passed the docs validator and generated-resources build, so there is no need for you to take #427 unless you prefer to.

PR #5402, opened 2026-07-24
feat(mcp) #4867: add configurable tool profiles

### LINKED ISSUE #4867: MCP: tool profiles in MCPConfiguration
Part of #4859.

## Motivation

The server will grow from 10 to potentially 16+ tools; flat tool lists degrade model tool-selection accuracy. A configurable profile (e.g. `rag`: `query`, `vector_search`, `hybrid_search`, `full_text_search`, `sample_records`, schema resource vs. `admin`: the current 10 operator tools) lets a given agent see a focused set. This isn't a new *tool*, so the four-justification test from #4859 doesn't directly apply — it's the Epic's explicit alternative to unbounded tool-list growth.

## Proposed design

- Add `profile` (or `enabledTools` allowlist) field to `MCPConfiguration`, default `all` (backward-compatible — no behavior change for existing deployments).
- `tools/list` and `tools/call` in both transports filter/reject against the active profile — a `tools/call` for a tool outside the active profile is denied even though it's globally registered (defense in depth, not just a UI filter).
- Expose profile selection via `MCPConfigHandler.java`'s existing admin config surface.
- Update `MCPConfigurationTest`/`MCPPermissionsTest` for profile filtering.

## Acceptance criteria

- [ ] Default profile (`all`) preserves current behavior exactly (regression-tested).
- [ ] `rag` and `admin` profiles defined and enforced in both `tools/list` and `tools/call`.
- [ ] Config persists through `mcp-config.json` like existing fields.
- [ ] Tests + docs, including a worked example of restricting an agent to `rag`.

## Files to touch

`MCPConfiguration.java`, `MCPHttpHandler.java`, `MCPStdioServer.java`, `MCPConfigHandler.java`, tests, docs.

## Dependencies

Soft-depends on #4860, #4861, #4862, #4863, #4864 for the `rag` profile's tool list to be meaningful; the filtering mechanism itself can be built in parallel with default `all`/`admin` profiles only, then extended once new tool names land.

## Effort estimate

M — config schema change, filtering logic duplicated across two transports, admin surface update.

justinblethrow-cloud:
## Summary

- add persisted `all`, `rag`, and `admin` Model Context Protocol (MCP) tool profiles
- filter `tools/list` through the active profile
- enforce the same allowlist in `tools/call` so hidden tools cannot be invoked directly
- provide retrieval-focused initialization instructions for the `rag` profile
- document profile contents, configuration, and permission boundaries

## Behavior

`profile` defaults to `all`, preserving the current tool surface exactly.

The `rag` profile permits registered retrieval tools:

- `query`
- `full_text_search`
- `sample_records`
- `vector_search`
- `hybrid_search`

Only tools registered by the running server are advertised. This lets pending retrieval tools join the existing profile allowlist when installed without advertising unavailable tools.

The `admin` profile exposes the original operator and database-administration surface. Schema Resources remain available through the MCP Resources capability rather than consuming a tool slot.

Profiles restrict tool visibility and invocation; they do not grant read, write, schema, administrative, user, or origin permissions. Those existing checks remain mandatory.

Unknown profile values return HTTP 400 from the configuration endpoint and are rejected before any other field in the same update is applied.

## Validation

- focused profile, configuration, HTTP, and standard input/output tests: 109 passed
- complete MCP-only server test set: 146 passed
- final mixed-case configuration regression test: 12 passed
- no failures or skipped tests

Closes #4867

### DIFF
--- a/docs/superpowers/specs/2026-07-24-mcp-tool-profiles-4867-design.md
+++ b/docs/superpowers/specs/2026-07-24-mcp-tool-profiles-4867-design.md
@@ -1,0 +1,77 @@
+# MCP tool profiles (#4867)
+
+**Issue:** [#4867](https://github.com/ArcadeData/arcadedb/issues/4867)
+**Epic:** [#4859 - MCP GraphRAG & Agent-Memory Surface](https://github.com/ArcadeData/arcadedb/issues/4859)
+**Date:** 2026-07-24
+**Status:** implemented
+
+## Decision
+
+The Model Context Protocol (MCP) configuration supports three named tool profiles:
+
+- `all`: every registered tool; this is the default and preserves existing behavior.
+- `rag`: retrieval-focused tools only.
+- `admin`: the existing operator and database-administration surface.
+
+Profiles are allowlists. They filter `tools/list` and are enforced again by `tools/call`, so a
+client cannot invoke a hidden tool by name. Unknown tools remain rejected independently of the
+selected profile.
+
+## Configuration
+
+Set `profile` in `config/mcp-config.json` or through the existing MCP configuration endpoint:
+
+```json
+{
+  "enabled": true,
+  "profile": "rag",
+  "allowReads": true,
+  "allowedUsers": ["retrieval-agent"]
+}
+```
+
+The field is persisted with the rest of the MCP configuration. Values are case-insensitive when
+parsed and serialized as lowercase. Unknown values are rejected before any other setting in the
+same update is applied.
+
+## Profile contents
+
+The `rag` profile permits these registered tool names:
+
+- `query`
+- `full_text_search`
+- `sample_records`
+- `vector_search`
+- `hybrid_search`
+
+Only tools actually registered by the running server are advertised. This lets retrieval tools
+join the profile as they are installed without exposing an unavailable tool.
+
+The `admin` profile contains:
+
+- `list_databases`
+- `get_schema`
+- `query`
+- `execute_command`
+- `server_status`
+- `profiler_start`
+- `profiler_stop`
+- `profiler_status`
+- `get_server_settings`
+- `set_server_setting`
+
+Schema resources remain available through the MCP Resources capability and are not tool-profile
+entries. Existing read, write, schema, administrative, user, and origin permissions remain
+independent mandatory checks; a profile never grants an operation.
+
+## Transport behavior
+
+Both HTTP and standard input/output transports use the shared `MCPDispatcher`, so discovery,
+execution denial, and profile-specific initialization instructions have one implementation.
+Changing a profile through the configuration endpoint takes effect on subsequent requests.
+
+## Validation
+
+Tests cover default compatibility, persistence, case-insensitive parsing, invalid-update
+atomicity, HTTP and standard input/output discovery filtering, execution denial for hidden
+tools, and successful execution of allowed tools.
--- a/docs/superpowers/specs/2026-07-24-mcp-tool-profiles-4867-design.md
+++ b/docs/superpowers/specs/2026-07-24-mcp-tool-profiles-4867-design.md
@@ -10,7 +10,7 @@
 The Model Context Protocol (MCP) configuration supports three named tool profiles:
 
 - `all`: every registered tool; this is the default and preserves existing behavior.
-- `rag`: retrieval-focused tools only.
+- `rag`: retrieval and agent-memory tools.
 - `admin`: the existing operator and database-administration surface.
 
 Profiles are allowlists. They filter `tools/list` and are enforced again by `tools/call`, so a
@@ -38,11 +38,15 @@ same update is applied.
 
 The `rag` profile permits these registered tool names:
 
+- `list_databases`
+- `get_schema`
 - `query`
 - `full_text_search`
 - `sample_records`
 - `vector_search`
 - `hybrid_search`
+- `upsert_entity`
+- `upsert_relationship`
 
 Only tools actually registered by the running server are advertised. This lets retrieval tools
 join the profile as they are installed without exposing an unavailable tool.
@@ -60,9 +64,11 @@ The `admin` profile contains:
 - `get_server_settings`
 - `set_server_setting`
 
-Schema resources remain available through the MCP Resources capability and are not tool-profile
-entries. Existing read, write, schema, administrative, user, and origin permissions remain
-independent mandatory checks; a profile never grants an operation.
+Schema resources remain available through the MCP Resources capability, while `get_schema`
+supports clients that expose only tools. Existing read, write, schema, administrative, user,
+and origin permissions remain independent mandatory checks; a profile never grants an
+operation. In particular, the two upsert tools remain unusable unless their existing write
+permissions are enabled.
 
 ## Transport behavior
 
@@ -75,3 +81,11 @@ Changing a profile through the configuration endpoint takes effect on subsequent
 Tests cover default compatibility, persistence, case-insensitive parsing, invalid-update
 atomicity, HTTP and standard input/output discovery filtering, execution denial for hidden
 tools, and successful execution of allowed tools.
+
+## Out of scope
+
+The profile setting remains server-global in this implementation. It cannot yet present a
+retrieval agent and an administrative agent with different tool lists when they share one MCP
+endpoint, so it does not by itself satisfy the per-agent part of #4859. Per-principal profile
+selection and enforcement is tracked in
+[#5445](https://github.com/ArcadeData/arcadedb/issues/5445).
--- a/server/src/main/java/com/arcadedb/server/mcp/MCPConfigHandler.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/MCPConfigHandler.java
@@ -52,7 +52,11 @@ public class MCPConfigHandler extends AbstractServerHttpHandler {
     if (Methods.POST.equals(exchange.getRequestMethod())) {
       if (payload == null)
         return new ExecutionResponse(400, new JSONObject().put("error", "Request body is required").toString());
-      config.updateFrom(payload);
+      try {
+        config.updateFrom(payload);
+      } catch (final IllegalArgumentException e) {
+        return new ExecutionResponse(400, new JSONObject().put("error", e.getMessage()).toString());
+      }
       config.save();
       return new ExecutionResponse(200, config.toJSON().toString());
     }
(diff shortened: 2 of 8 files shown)


robfrank:
Reviewed against the epic #4859 acceptance criteria. The mechanics here are right: the profile is an allowlist enforced at both `tools/list` and `tools/call`, so a client cannot invoke a hidden tool by name; an unknown profile value is rejected before any other field in the same update is applied; and forward-declaring `vector_search`, `hybrid_search` and `sample_records` in `RAG_TOOL_NAMES` while gating everything through `REGISTERED_TOOL_NAMES` is a genuinely nice way to decouple this from the merge order of the Wave 1 tools. The `MCPConfigHandler` change from a 500 to a 400 on bad input is a real improvement.

The problems are with the profile contents and the scope of the setting, not the implementation.

### 1. The `rag` profile has no schema discovery (blocking)

`RAG_INSTRUCTIONS` tells the agent:

> Read the arcadedb://{database}/schema Resource before searching an unfamiliar database.

But `RAG_TOOL_NAMES` excludes both `get_schema` and `list_databases`. MCP Resources support is still uneven across clients, and plenty of them expose Tools only. On such a client a `rag`-profile agent cannot discover type names, property names, or even which databases exist. It is left to guess identifiers and feed them to `query`, which is exactly the failure #4859 wants the retrieval surface to prevent.

Right now the effective `rag` profile on main is two tools, `query` and `full_text_search`, which is a smaller surface than is usable. Please add `get_schema`, and `list_databases` unless you have a specific reason to withhold it.

### 2. No profile exposes the agent-memory write tools

`upsert_entity` and `upsert_relationship` appear in neither `rag` nor `admin`, so they are reachable only under `all`.

#4859 is titled "GraphRAG **and Agent-Memory** Surface", and #4864 shipped those two tools specifically for the knowledge-graph-construction persona. As it stands, any operator who selects a profile to narrow the surface loses agent-memory writes entirely, and the only way to keep them is to select `all` and give up the narrowing that is the whole point of the feature.

Either add a `memory` (or `kg`) profile pairing the upserts with the retrieval tools, or include them in `rag` on the grounds that agent memory both reads and writes. My preference is the latter, since a separate profile per persona multiplies quickly. Whichever you choose, the profile permission is orthogonal to the write flags, which stay `false` by default, so this does not weaken the safe-by-default posture.

### 3. The profile is server-global, but the criterion is per agent

#4859's success criterion is "The tool surface **a given agent** sees stays small via configurable profiles." A single server-wide `profile` field cannot give the retrieval agent `rag` and the DBA `all` at the same time. Serving both personas means running two servers, which defeats the purpose.

#5401 is concurrently introducing per-database permission scoping keyed on the authenticated principal, and already has the resolver plumbing for exactly this. Profiles should key off the same principal. I am not asking you to build that in this PR, but please note it in the spec's out-of-scope section (which currently justifies the global choice on the grounds that "the current endpoint advertises one tool surface", which is circular) and open a follow-up. Otherwise the criterion reads as satisfied when it is not.

### 4. Smaller points

`MCPConfigHandler` now maps `IllegalArgumentException` to 400, but the rest of `updateFrom` still throws `JSONException` on a type mismatch such as `{"enabled": "yes"}`, which surfaces as a 500. Worth widening the catch while you are in there.

`toolsForProfile` returns the shared static `TOOLS_LIST` instance for `ALL` and a freshly built array otherwise. The aliasing predates this PR, but the method now has two different ownership semantics depending on the branch. Return a copy in both.

### Cross-PR: the test assertions will break each other

`toolProfilesFilterDiscoveryAndExecution` asserts the `rag` profile is exactly `{"query", "full_text_search"}` in both `MCPServerPluginTest` and `MCPStdioServerTest`. That breaks the moment #5398 (`sample_records`) or #5399 (`vector_search`) registers, since both names are already in `RAG_TOOL_NAMES`.

The mirror image is also true: #5398 and #5399 each change `assertThat(tools.length()).isEqualTo(13)` to `isEqualTo(14)`, so whichever of those merges second is red as well.

Across all four PRs, please derive the expected names and count from the registered tool list rather than hardcoding, or assert containment only. That is the one change that removes the merge-order coupling entirely.

### Cross-PR: conflict with #5401

#5401 modifies `MCPConfiguration.load()`, `updateFrom()` and `toJSON()` in the same hunks this PR does, and both add `assertThatThrownBy` plus assertions to `MCPConfigurationTest.saveAndLoad` and `toJSON`. Both use the same "parse and validate before mutating" pattern, so the merge is mechanical, but whoever goes second should re-run `MCPConfigurationTest` rather than trusting the resolution.

justinblethrow-cloud:
Thanks, Roberto. Addressed in `f4e250b3`:

- `rag` now includes `list_databases`, `get_schema`, `upsert_entity`, and `upsert_relationship`, while the existing write flags remain independent and denied by default.
- Initialization guidance now covers tools-only schema discovery and agent-memory writes.
- Invalid boolean configuration types are rejected and returned as HTTP 400.
- `toolsForProfile` returns a fresh array for every profile.
- Exact profile/tool-count assertions were replaced with containment and behavior assertions.
- The server-global limitation is explicit in the design note. The requested per-principal follow-up is open as #5445.

Focused validation: `MCPConfigurationTest`, `MCPPermissionsTest`, `MCPServerPluginTest`, and `MCPStdioServerTest`; 113 tests passed.

robfrank:
Correction to my review above: I did not raise the docs criterion from #4859 because I had wrongly concluded no MCP documentation lives in this repo. The MCP reference page is at https://docs.arcadedb.com/arcadedb/reference/mcp/mcp and its source is `ArcadeData/arcadedb-docs`, file `src/main/asciidoc/reference/mcp/mcp.adoc`.

That page documents the `mcp-config.json` keys as exactly `enabled`, `allowReads`, `allowInsert`, `allowUpdate`, `allowDelete`, `allowSchemaChange`, `allowAdmin`, `allowedUsers`, `allowedOrigins`. This PR adds `profile`, and it is the key an operator most needs to find, because it is the one that changes which tools an agent can see at all.

Please open a companion `arcadedb-docs` PR documenting `profile`: the three accepted values, the exact tool membership of `rag` and `admin` once points 1 and 2 of my review are settled, the fact that it is an allowlist enforced at both discovery and invocation, and the fact that it never grants an operation, since the read/write flags remain independent mandatory checks. That last point is the one most likely to be misread as a permission mechanism.

One thing that makes point 1 of my review sharper: the page does not mention MCP Resources anywhere, so the `arcadedb://{database}/schema` resource from #4865 is shipped but undocumented. `RAG_INSTRUCTIONS` directs the agent to a facility that is both unevenly supported across clients and absent from the documentation, while `get_schema` is excluded from the profile. Documenting the resource is not this PR's debt and I will file it separately, but it reinforces that `rag` should carry `get_schema`.

justinblethrow-cloud:
Companion documentation is now open as ArcadeData/arcadedb-docs#428.

It documents the three profile values, exact `rag` and `admin` memberships, version-dependent registration, enforcement at both discovery and invocation, independent mandatory permission checks, and the current server-global scope.

- `python3 docs-validator.py`
- `./mvnw generate-resources`

robfrank:
Both Wave 1 tools are now on main (`vector_search` #5399, `sample_records` #5398), which is the point #4859 sequenced this PR for: the `rag` profile membership can finally be settled against real tool names. I have merged `main` in and pushed, so this is mergeable again.

### On your fix commit

`23a7cb0d4` resolves the substantive findings, and the choices are the right ones:

- `rag` now carries `list_databases` and `get_schema`, so a retrieval agent on a Tools-only client can actually discover what it is searching. This was my blocking point.
- `rag` now carries `upsert_entity` and `upsert_relationship`, so the agent-memory half of the epic survives profile selection. Folding them into `rag` rather than minting a separate profile is the choice I would have made.
- `RAG_INSTRUCTIONS` was rewritten to match the new membership instead of pointing only at the Resources API.
- `MCPConfigHandler` now catches `JSONException` alongside `IllegalArgumentException`, and `MCPConfiguration` raises a typed error naming the offending field, so `{"enabled": "yes"}` is a 400 with a usable message rather than a 500.
- `toolsForProfile` always builds a fresh array, and `toolsListReturnsIndependentArrays` pins it. That closes the shared-`TOOLS_LIST` aliasing.

I checked for the pattern that bit the `vector_search` truncation fix, where new reporting quietly introduced an O(n) scan. Nothing comparable here: profile resolution is a `Set` lookup against a `Set.copyOf` built once in the static initializer.

### What I changed on top

The only conflict was the stdio discovery test, which this branch and main had each rewritten off the old hardcoded count. I kept this branch's version, which goes through the `toolNames` helper, and folded `sample_records` into it.

Then I extended the profile assertions, since the Wave 1 names have settled and the tests were written before they existed. Both the HTTP and stdio profile tests now assert `rag` exposes `sample_records` and `vector_search`, and `admin` does not. Without that, `RAG_TOOL_NAMES` could list a tool that never actually reaches a client and nothing would catch it, which is precisely the gap the forward-declaration plus `REGISTERED_TOOL_NAMES` design creates.

I also imported `JSONException` in `MCPConfigurationTest` rather than naming it inline, per the project convention against fully qualified names.

`mvn test -Dtest='MCPServerPluginTest,MCPStdioServerTest,MCPConfigurationTest,MCPPermissionsTest'` is green at 133 tests.

### Still open

**Profiles are server-global.** This is the one review point not addressed, and I do not think it should block the PR. #4859's criterion is that the surface *a given agent* sees stays small, and a single server-wide `profile` cannot serve a retrieval agent and a DBA from one endpoint. #5401 is introducing per-database resolution keyed on the authenticated principal, which is the natural place for this to key off too. Please either note it in the spec's out-of-scope section with that reasoning, or open a follow-up issue so the criterion is not recorded as met when it is partly met. Right now the spec justifies the global choice on the grounds that the endpoint advertises one tool surface, which is circular.

**Docs.** The `arcadedb-docs` gap has grown to three items: `profile` is a new top-level `mcp-config.json` key, and the page still predates both Wave 1 tools. One PR covering `profile`, `sample_records` and `vector_search` would close it. The `profile` entry matters most, because it is the setting that changes which tools an agent can see at all, and it must be clear that a profile never grants an operation since the read and write flags remain independent mandatory checks.

justinblethrow-cloud:
Thanks, Roberto. I reviewed the merge resolution and agree with the settled Wave 1 profile assertions.\n\nThe two remaining items are already covered:\n\n- The design spec's `Out of scope` section explicitly states that the server-global profile cannot present different surfaces to different agents, says that this only partly satisfies #4859, and links the per-principal follow-up #5445.\n- The documentation is split into the already-open companion PRs ArcadeData/arcadedb-docs#425 (`sample_records`), ArcadeData/arcadedb-docs#426 (`vector_search`), and ArcadeData/arcadedb-docs#428 (`profile`). All three passed `docs-validator.py` and `./mvnw generate-resources`.\n\nNo additional code change appears necessary from these two points.

Issue #4790, opened 2026-06-29
Leader silently loses a committed write when replication times out after dispatch (phantom commit + origin-skip)


ruispereira:
**Affected version:** 26.6.1
**Component:** `RaftReplicatedDatabase.commit` + `ArcadeStateMachine.applyTxEntry` + `RaftGroupCommitter.submitAndWait`
**Severity:** CRITICAL

### Summary

The leader applies its own transactions in a local "Phase 2" (`commit2ndPhase`)
and, to avoid double-applying, `applyTxEntry` **skips** the apply for entries
whose context was marked `originatedLocally == true`. But the local marker is set
at `startTransaction` for the whole lifecycle, independent of whether Phase 2
actually ran. If replication is dispatched to Ratis and then the wait times out
(`tryCancel()` fails because already dispatched → `QuorumNotReachedException`),
`commit()` rolls back and throws **without running Phase 2** — yet Ratis later
reaches quorum and the entry is applied on all followers, while the leader's own
`applyTransaction` skips it (origin-skip).

### Code

`RaftReplicatedDatabase.java:316-332` (rollback+throw on the dispatched-timeout path),
`RaftGroupCommitter.java:158-168` (`QuorumNotReachedException` after dispatch),
`ArcadeStateMachine.java:243-255` (marks `originatedLocally`), `:576-579` (skips apply when local).

### Impact

The transaction is durably committed in the Raft log and applied on every
follower, but **never applied on the leader** — silent permanent divergence / a
lost write on the leader, with no `WALVersionGapException` and no step-down.
Worst class of bug for a consensus system.

### Suggested fix

Condition the origin-skip on Phase 2 having actually completed (track a per-entry
"applied locally" marker keyed by Raft log index / client call-id, set only after
`commit2ndPhase` succeeds); on the dispatched-but-timed-out path, block on local
apply or force Phase 2 / step down rather than rollback-and-skip.

lvca:
Fixed

Issue #5345, opened 2026-07-20
Raft log grows unbounded until disk-full (auto-snapshot threshold 100000 never reached on low-write clusters); follower then permanently wedged on "No space left on device"


ruispereira:
## Summary

On a Kubernetes HA cluster, the segmented Raft log under `raftStorageDirectory` grows **without bound** until it fills the volume, even though the actual state-machine data is tiny (~72 MB of databases produced **1.5–1.9 GB** of Raft log segments). The growth happens because the leader's automatic snapshot — which is what drives log purge (`logPurgeUptoSnapshot=true`) — is only triggered every `arcadedb.ha.snapshotThreshold` (default **100000**) committed entries, and on a low/moderate-write cluster that threshold is **effectively never reached between process restarts**. The snapshot index therefore stays frozen at the value captured at startup, no segments are purged, and the log grows monotonically (~64 MB per segment) until the disk is full.

Once a follower's volume hits 100 %, the Raft log writer fails with `java.io.IOException: No space left on device`; Ratis then marks the log **permanently failed at that index** and rejects every subsequent `APPEND_ENTRIES`. The follower is wedged forever (lag grows without recovering), and the built-in self-heal paths cannot rescue it (details below).

## Environment

- ArcadeDB **26.8.1-SNAPSHOT**, 3-node HA on Kubernetes (Apache Ratis backend), one PVC per pod.
- `raftStorageDirectory` on the data PVC (2 GiB each), `databaseDirectory` = `/home/arcadedb/databases`.
- Relevant live config (from the startup config dump):
  ```
  arcadedb.ha.snapshotThreshold        = 100000
  arcadedb.ha.logPurgeUptoSnapshot     = true
  arcadedb.ha.logPurgeGap              = 1024
  arcadedb.ha.logSegmentSize           = 64MB
  arcadedb.ha.staleFollowerLagThreshold        = 10000
  arcadedb.ha.staleFollowerRecoveryDurationMs  = (default)
  arcadedb.ha.divergedFollowerRecovery = true
  ```

## Observed symptoms

- Follower `server-2` stuck: `matchIndex` frozen at **44026**, lag steadily **growing** (sampled 5952 → 5960 → 6000, never shrinking).
- Leader logs the stall every 30 s:
  ```
  SEVER [ClusterMonitor] Replica 'server-2..._2434' STALLED: matchIndex stuck at 44026 while
  leader advanced by 24 (current lag=1010). This will trigger a leader election if it continues.
[...]

Issue #4740, opened 2026-06-25
HA: leader phase-2 commit failure after successful Raft replication → follower-ahead divergence, non-converging snapshot resync, and node self-halt


robfrank:
## Affected version

26.7.1-SNAPSHOT (build `29d593e096fa9772a356bffd7dd8868004bad10b`), 3-node Raft HA, official server image, JDK 21, Docker.

> Note: this build already includes the #4659 fix. The node-self-halt is still reachable here through the `ReplicationException` → resync route described below, so #4659 does not cover this case.

**Component:** `ha-raft` — `RaftReplicatedDatabase`, `ArcadeStateMachine`

## Summary

Under concurrent write load on a 3-node Raft HA cluster, a leader can fail its **local phase-2 commit *after* Raft has already replicated and committed the transaction on the followers**. The leader detects that the followers are ahead of it and steps down ("to prevent stale reads"). This leaves the **ex-leader's on-disk page versions lagging the committed Raft log**. When it rejoins as a replica it cannot apply the log (`WALVersionGapException`) and requests a snapshot resync — but the resync **never converges** (it loops across many files while leadership thrashes), and after ~1 minute the state machine fail-stops:

```
State machine halted after critical error at earlier index; refusing to apply index N
```

The node then shuts itself down. This is a deliberate, orderly self-halt (`ExitCode=0, OOMKilled=false`) — **not** an OOM kill and **not** an external signal, despite the trailing "Received shutdown signal. The server will be halted" line.

## Environment

- HA enabled, `replicationFactor=3`, `quorum=majority`, `readConsistency=read_your_writes`
- 3 nodes (`node-0/1/2`), Raft port 2434, HTTP 2480
- Graph workload: per message, one transaction running OpenCypher
  `MERGE (a:Account {bank,number}) MERGE (b:Account {bank,number}) CREATE (a)-[:Transaction {...}]->(b)`,
  many concurrent writers against a **shared account pool** (high page contention), ~3M edges
- Nodes constrained to ~2 CPUs / 8 GB each. The CPU constraint appears to be a strong trigger (see Reproduction).

## Timeline (single incident, condensed from one node's log)

```
10:02:05  [ArcadeStateMachine] Leader changed: node-2 -> node-1 (term=1); This node is now LEADER
10:02:10  SEVER [RaftReplicatedDatabase] <node-1> Phase 2 commit failed AFTER successful Raft replication
          (db=<db>). Followers have applied this transaction but the leader has not.
[...]

PR #5408, opened 2026-07-24
fix(#5407): keep unapplied phase-2 entries inside the Raft replay window

### LINKED ISSUE #5407: HA: leader crash between Raft commit and phase-2 apply permanently loses a locally-originated write (RaftLeaderCrashBetweenCommitAndApplyIT)
## Summary

`RaftLeaderCrashBetweenCommitAndApplyIT.oldLeaderRecoversMissingWriteAfterCrashBetweenReplicationAndPhase2` fails **deterministically** (4/4 local runs). A leader that crashes in the narrow window between Raft commit and local phase-2 apply of a locally-originated write **never recovers that write on restart**, producing permanent single-node divergence.

```
expected: 101L
 but was: 100L
  at RaftLeaderCrashBetweenCommitAndApplyIT.java:176
Suppressed: DatabaseAreNotIdentical: Page PageId(graph/28/0) has different versions: DB1 v2 <> DB2 v1
```

The restarted old leader ends with 100 records (baseline) instead of 101 (baseline + injected); its `graph/28/0` page stays at v1 while the surviving nodes are at v2.

## Root cause

For a locally-originated write, the actual local page write is **phase 2** (`RaftReplicatedDatabase.commit()` → `commit2ndPhase`), which runs *after* Raft replication. When Raft commits the entry, Ratis independently calls `ArcadeStateMachine.applyTransaction` on the leader, which:

1. **origin-skips** the data apply because `originatedLocally == true` (`ArcadeStateMachine.java:1161`) — correct, since phase 2 is meant to do the write; but
2. still **durably advances the persisted applied index** via `writePersistedAppliedIndex(index, db)` (`ArcadeStateMachine.java:573`).

The test injects a crash (`RaftReplicatedDatabase.TEST_POST_REPLICATION_HOOK` throws + async `server.stop()`) in the window **after** step 1/2 but **before** phase 2. The net durable state on the old leader is therefore:

- persisted applied index = injected entry's index (looks applied), **but**
- the injected record's pages were never written (phase 2 abandoned).

On restart the old leader rejoins as a follower whose applied index already covers the injected entry, so neither Raft log replay nor follower catch-up re-applies it. The write is silently and permanently lost on that node (no later entry to trip `WALVersionGapException`-driven resync, since it was the last entry).

## Relationship to #4790

This is the **hard-crash analog of #4790**. The #4790 fix (`abandonedLocalTransactions` in `ArcadeStateMachine`) applies a write locally instead of origin-skipping it when phase 2 was abandoned — but only on the graceful `ReplicationDispatchedTimeoutException` path, and the marker map is **in-memory**, so it cannot survive an actual process crash. This scenario is exactly that uncovered gap.

## Reproduction

```
mvn -pl ha-raft verify -DskipTests -Pintegration -Dit.test=RaftLeaderCrashBetweenCommitAndApplyIT -Dfailsafe.failIfNoSpecifiedTests=false
```

Fails 4/4 with the assertion above. The test is `@Tag("slow")` (nightly), authored under the port-apache-ratis-hardening plan, and is not marked known-flaky — it correctly expects recovery.

## Not a regression from a recent change

`writePersistedAppliedIndex` and the origin-skip both date to the original HA commit (#3731, same commit that added the test); #5395 only touches the pre-dispatch submission path. The failure is inherent to the current durability ordering, not introduced by a recent PR.

## Suggested fix directions (need design + review)

- Defer advancing the **persisted** applied index for locally-originated (origin-skipped) entries until phase 2 confirms the local write, so a crash before phase 2 leaves the entry replayable on restart; **or**
- Durably record a "phase-2 pending" marker (crash-safe equivalent of `abandonedLocalTransactions`) that restart replay consults to force re-apply; **or**
- On restart, reconcile committed-but-not-locally-applied entries against actual page state rather than trusting the persisted applied index for locally-originated entries.

Any fix must preserve the origin-skip fast path and idempotent-replay guarantees, and be validated against the full `ha-raft` suite (especially the #4790/#4824/#4790-adjacent tests).

## Impact

Low-probability but real data-durability gap: a leader crash in a millisecond-scale window can permanently drop an acknowledged write on the crashed node until a manual/triggered full resync. Followers retain the write (Raft durability holds), so the cluster majority is correct; the risk is silent divergence of the recovered ex-leader.

robfrank:
Closes #5407

## Summary

On the leader, a locally-originated entry is origin-skipped by `ArcadeStateMachine.applyTransaction` because `RaftReplicatedDatabase.commit()`'s **phase 2** writes the pages - but phase 2 runs only *after* Raft commits. `lastAppliedIndex` therefore advances past an entry whose pages are not on disk yet, and `takeSnapshot()` reported that index to Ratis as the durability checkpoint, writing a `snapshot.<term>_<index>` marker at it. Because `reinitialize()` seeds the replay position **exclusively** from that marker, a restart began replaying *above* the never-applied entry: neither Raft replay nor follower catch-up revisited it, so the acknowledged write was silently and permanently lost on that node.

The detail worth flagging beyond the issue report: **Ratis takes a snapshot on shutdown**, which is what makes the loss durable. It is not only a hard-crash window - any shutdown following an aborted phase 2 (graceful stop, or the `triggerCriticalHalt()` path) buries the entry. The failing run left the smoking gun on disk: the crashed leader had `snapshot.1_9`, a checkpoint sitting exactly on the lost entry at index 9.

**Fix.** `ArcadeStateMachine` tracks in-flight leader-side phase-2 applies: `beginLocalPhase2()` records the applied index observed *before* replication (the entry's replay floor), and `takeSnapshot()` clamps its reported checkpoint to the lowest in-flight floor, so it can never cover an entry whose phase 2 has not confirmed. It also refuses to regress the marker below an existing one - that would ask Ratis to replay from entries a previous checkpoint already authorised for purging - reporting `INVALID_LOG_INDEX` instead.

`RaftReplicatedDatabase` **retains that guard by default** and releases it only where the local pages are provably settled: phase 2 succeeded, a failed phase 2 reconciled successfully, ALL-quorum recovery settled the pages, the commit was on a replica (no phase 2 expected), or replication failed outright so no entry exists. Every other exit - notably "replication succeeded but phase 2 never ran" - keeps the hold, which is exactly the #5407 window. A retained ticket costs a stalled compaction checkpoint until the node restarts, at which point replay applies the entry and the hold disappears with the process: a deliberate trade of log-compaction progress for durability. Replay is idempotent via the existing `applyChanges` page-version guards.

As a side effect this also makes the #4790 `ReplicationDispatchedTimeoutException` window crash-safe: its `abandonedLocalTransactions` marker is in-memory and cannot survive a crash, but the retained ticket means replay now covers the entry.

## Test plan

- [x] New `ArcadeStateMachinePendingPhase2SnapshotTest` (5 cases): checkpoint clamps to the pending floor; advances again once phase 2 completes; tracks the oldest of several in-flight commits; never regresses below an existing marker; suppressed entirely when a phase 2 starts before anything has been applied.
- [x] The two ITs from the issue, which failed **100%** before (4/4 and 3/3) and pass after:
  - `RaftLeaderCrashBetweenCommitAndApplyIT`
  - `RaftLeaderCrashWithExternalPropertyIT`
- [x] Full `ha-raft` unit suite: **723 tests, 0 failures**.
- [x] Targeted IT sweep over the commit path, origin-skip, phase-2 failure handling, snapshot/compaction and crash recovery - **22 tests across 15 IT classes, 0 failures**: `Issue4790PhantomCommitOriginSkipIT`, `OriginNodeSkipIT`, `Issue4740Phase2ReconcileIT`, `Issue5064CommittedRemotelyContractIT`, `RaftLeaderCrashAndRecoverIT`, `RaftPeriodicSnapshotCompactionIT`, `RaftLogCompactionWiringIT`, `RaftFullSnapshotResyncIT`, `Raft3PhaseCommitIT`, `RaftReplication3NodesIT`, `RaftReplicaCrashAndRecoverIT`, `RaftIdleReplicaRestartIT`, `RaftBootstrapDoesNotEngageOnRestartIT` plus the two above.

Reproduce:

```
mvn -pl ha-raft verify -DskipTests -Pintegration \
  -Dit.test='RaftLeaderCrashBetweenCommitAndApplyIT,RaftLeaderCrashWithExternalPropertyIT' \
  -Dfailsafe.failIfNoSpecifiedTests=false

mvn -pl ha-raft test -Dtest=ArcadeStateMachinePendingPhase2SnapshotTest
```

Note: `Raft3PhaseCommitIT.concurrentWritersReplicateCorrectly` flaked once on the first batch run (HTTP non-200 on the very first insert, while the previous IT's servers were still shutting down). It passed 3/3 in isolation and the full batch re-ran clean, so it is cross-test startup contention, not a regression from this change.

## Notes

- The persisted `.raft/applied-index` file is intentionally left unclamped: it never feeds the replay position (only the snapshot-gap check, the per-database bootstrap replay-skip, and the "never applied anything" bootstrap signal), so clamping it would change bootstrap behaviour without improving durability.
- `RaftHAServer.takeLocalSnapshot` already treats an unsuccessful snapshot reply as a FINE-level no-op, so the new `INVALID_LOG_INDEX` path adds no log noise.

### DIFF
--- a/docs/5407-ha-leader-crash-phase2-write-loss.md
+++ b/docs/5407-ha-leader-crash-phase2-write-loss.md
@@ -122,12 +122,44 @@ Not applied: a Micrometer gauge for `pendingLocalPhase2` size. This class has no
 today, so adding it is genuine follow-up scope rather than part of this fix; the WARNING covers the
 operator-visibility gap in the meantime.
 
+## Review cycle 2 (head `1dd6ea54e`)
+
+`claude[bot]` re-reviewed and independently confirmed the crux ordering (register the floor before
+replication, release only after pages settle) and the idempotent double-release. Applied:
+
+- **Orphaned Javadoc fixed.** Inserting `releasePhase2Ticket`/`stateMachineOrNull` had split
+  `markTransactionAbandonedForLocalApply`'s Javadoc from its method, leaving two stacked comments on
+  one method and none on the other. A real defect introduced by this PR; the block is back where it
+  belongs.
+- **`Issue5407Phase2TicketLifecycleTest` extended to 7 cases**, closing the branch matrix with the
+  `MajorityCommittedAllFailedException` (ALL-quorum recovery) exit: release when the recovery apply
+  settles the pages, retain when neither it nor reconciliation does.
+
+Considered and **not** changed, with reasons:
+
+- **Drop the `// @VisibleForTesting` marker.** Declined: it is an established convention in this
+  codebase (12 occurrences, 5 of them in `ArcadeStateMachine` in this same package). Removing it here
+  would make the new code inconsistent with its immediate neighbours, against the project's
+  "adhere to the existing code" rule.
+- **Per-commit allocation** (boxed `Long` key + `PendingPhase2` record on each leader commit). The
+  reviewer flagged this for awareness rather than as a request. Kept: the same commit path already
+  allocates the payload, a copy of the bucket-delta map, and the WAL byte array, then makes a gRPC
+  round trip - two small short-lived objects are noise against that, and the map churns back to empty
+  in steady state.
+
 ## Notes for follow-up
 
 - The `ReplicationDispatchedTimeoutException` path (#4790) now also retains its ticket, which makes
   that indeterminate window crash-safe as a side effect. The in-memory
   `abandonedLocalTransactions` marker it relies on is still lost on a real crash; the retained
   ticket means replay covers the entry instead.
+  - **Known rough edge, worth its own issue.** When such an entry *does* later reach quorum and is
+    applied by `applyTxEntry` (the usual resolution of a #4790 timeout), its pages become durable but
+    nothing releases the ticket, so compaction stays pinned until the node restarts. Since #4790
+    timeouts are transient, that is arguably heavier than the crash window it protects. Releasing the
+    ticket when `applyTxEntry` consumes an abandoned-tx marker would fix it, but needs a
+    ticket-to-walTxId correlation that does not exist today, so it is deliberately left out of this
+    PR rather than bolted on to a delicate path. The 5-minute WARNING makes it visible meanwhile.
 - The persisted `.raft/applied-index` file is intentionally left unclamped. It never feeds the
   replay position (only the snapshot-gap check, the per-database bootstrap replay-skip, and the
   "never applied anything" bootstrap signal), so clamping it would change bootstrap behaviour
--- a/docs/5407-ha-leader-crash-phase2-write-loss.md
+++ b/docs/5407-ha-leader-crash-phase2-write-loss.md
@@ -93,6 +93,35 @@ mvn -pl ha-raft verify -DskipTests -Pintegration \
 mvn -pl ha-raft test -Dtest=ArcadeStateMachinePendingPhase2SnapshotTest
 ```
 
+## Review cycle 1 (PR #5408, head `198274bd9`)
+
+`claude[bot]` reviewed and confirmed the correctness of the clamp direction, the no-regression guard,
+and the idempotent ticket release. `gemini-code-assist` did not respond within the polling window.
+Applied:
+
+- **Stalled-compaction observability** (the reviewer's only pre-merge ask). A retained ticket pins the
+  checkpoint until restart, previously signalled only by a `HALog.BASIC` line. `pendingLocalPhase2`
+  now records each ticket's start time and `takeSnapshot()` emits a throttled `WARNING` once a held
+  ticket exceeds 5 minutes, naming the pinned index and pointing at Raft-storage disk usage.
+  Implemented rather than left as a TODO: it is a hazard this change introduces.
+  - One detail in the review was slightly off: a stuck ticket does *not* make every `takeSnapshot()`
+    return `INVALID_LOG_INDEX`. It keeps returning the stalled floor and re-registering the marker
+    there; only a clamp that would regress below an existing marker returns `INVALID_LOG_INDEX`. The
+    operational consequence (log not purged past that index) is the same.
+- **Term/index comment** in `takeSnapshot()` explaining that after a clamp the marker term may not
+  match the clamped index, why that is already tolerated (#575/#593), and why it must not be
+  "corrected" by looking up the term at the clamped index (that entry may be purged).
+- **`Issue5407Phase2TicketLifecycleTest`** (5 cases) pinning which exit releases the ticket -
+  phase-2 success releases, a fault between replication and phase 2 retains, an unreconciled phase-2
+  failure retains, replication failure releases, replica commit releases. `replicateAndCommitLocally`
+  was made package-private for this (existing convention in the module). This is the part of the fix
+  most exposed to a future refactor, and the retain case fails if the guard is moved back into a
+  blanket `finally`.
+
+Not applied: a Micrometer gauge for `pendingLocalPhase2` size. This class has no metrics wiring
+today, so adding it is genuine follow-up scope rather than part of this fix; the WARNING covers the
+operator-visibility gap in the meantime.
+
 ## Notes for follow-up
 
 - The `ReplicationDispatchedTimeoutException` path (#4790) now also retains its ticket, which makes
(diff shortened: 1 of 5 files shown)


robfrank:
Thanks - the review caught a real defect. Responses to cycle 2:

**1. Orphaned Javadoc — fixed** (`db2508edb`). Confirmed: inserting `releasePhase2Ticket`/`stateMachineOrNull` split `markTransactionAbandonedForLocalApply`'s Javadoc from its method, leaving two stacked comments on one and none on the other. Moved back.

**3. `MajorityCommittedAllFailedException` branch — covered.** `Issue5407Phase2TicketLifecycleTest` is now 7 cases: release when the ALL-quorum recovery apply settles the pages, retain when neither it nor reconciliation does. The matrix is closed.

**2. #4790 ticket held after the entry later applies — agreed, not fixed here.** Recorded as a known rough edge in `docs/5407-*.md` with the reason: releasing on `applyTxEntry` needs a ticket-to-`walTxId` correlation that does not exist today, and bolting one onto that path in this PR trades a durability fix for new risk in a window that is hard to test. The 5-minute WARNING makes it visible in the meantime. Happy to file a tracking issue if you want it separate from the doc note.

**5. `// @VisibleForTesting` — kept, respectfully.** It is not dead decoration in this codebase but an established convention: 12 occurrences, 5 of them in `ArcadeStateMachine` in this same package (e.g. above `applyDropDatabaseEntry` and `applyBootstrapFingerprintEntry`). Dropping it here would make the new code inconsistent with its immediate neighbours, which cuts against the repo's "adhere to the existing code" rule. Happy to remove it if the convention is being retired repo-wide.

**4. Per-commit allocation — kept, flagged as understood.** The same commit path already allocates the `ReplicationPayload`, a copy of the bucket-delta map and the WAL byte array, then makes a gRPC round trip; one boxed key plus one small record is noise against that, and the map churns back to empty in steady state.

One correction for the record, from cycle 1: a stuck ticket does **not** make every `takeSnapshot()` return `INVALID_LOG_INDEX`. It keeps returning the stalled floor and re-registering the marker there; only a clamp that would regress *below* an existing marker returns `INVALID_LOG_INDEX`. The operational consequence you identified - the log not being purged past that index - is exactly right, which is why the WARNING went in.

Verification after these changes: 730 unit tests pass, plus the two crash ITs and the #4790 origin-skip, #5064 committed-remotely and #4740 phase-2 reconcile ITs.

robfrank:
Cycle 3 applied (`3ac4f9ce4`, `c048ef67f`). Nothing here was blocking; all four actionable points are in:

**1. Query method with a logging side effect — fixed.** The WARNING is hoisted out of `lowestPendingLocalPhase2Floor()` into `takeSnapshot()`, which already owns the "is a phase 2 stuck" decision. Both scans are pure queries again; the second pass costs nothing on this rare path.

**2. Warning is snapshot-driven, not time-driven — made explicit.** Stated in the method Javadoc and in the doc: the check can lag a stuck ticket by up to one compaction interval (`arcadedb.ha.snapshotInterval`, 5 min default).

**3. Follow-up issue filed: #5410.** You were right to push on this twice. It covers the ticket-to-`walTxId` correlation needed to release on `applyTxEntry`, the specific trap (release only on the branch that actually applies, never on origin-skip, or #5407 comes back), and the Micrometer gauge. Linked from the doc's Known gaps and cross-referenced to #4790 and #5345 — the latter is the failure mode this gap could feed.

**4. Hardcoded `/tmp` path — replaced** with a per-instance unique key. It is only a `DatabaseContext` key (the database is a mock, nothing touches the filesystem), so this is purely portability and parallel-run safety.

**5. Docs contained review-process meta-notes — agreed and rewritten.** Restructured to the house style used by `docs/53xx-*.md` (Symptom / Root cause / Fix / Tests / Verification / Known gaps / Design notes). The process narrative and bot names are gone; the durable rationale stays — including why a blanket `finally` is wrong, since that is the trap a future refactor is most likely to fall into.

Verification after these changes: 730 unit tests, plus the two crash ITs and `RaftPeriodicSnapshotCompactionIT` / `RaftLogCompactionWiringIT` (re-run specifically because `takeSnapshot()` changed).

PR #5419, opened 2026-07-25
perf: batch time-series HTTP ingest into one append transaction per measurement

robfrank:
## What does this PR do?

Batches time-series HTTP ingest so each measurement's samples are appended in **one** shard transaction instead of one per sample.

- `PostTimeSeriesWriteHandler` (InfluxDB line protocol) groups parsed samples by measurement, then issues one `TimeSeriesEngine.appendBatch` per group. The schema lookup and `instanceof` narrowing now happen once per measurement rather than once per sample.
- `PostPrometheusWriteHandler` issues one `appendBatch` per remote-write `TimeSeries` — those samples already share a type and label set.

`TimeSeriesEngine.appendBatch` already existed for exactly this and was only reachable from tests.

Behaviour is otherwise unchanged: drop-set ordering, the `written`/`dropped` counts and the partial-write `400` response are preserved (`LinkedHashMap`/`LinkedHashSet` keep first-occurrence order).

## Motivation

Both handlers called `TimeSeriesEngine.appendSamples` once per parsed sample. Each of those opens its own `begin`/`commit` inside `TimeSeriesShard.appendSamples`, and **on a Raft HA leader that commit is a full replicated quorum round trip** — so a 100-sample request cost 100 sequential quorum round trips, all serialized behind the per-shard `appendLock`.

`TimeSeriesThreeNodeWalGapReproIT` was failing on this. Evidence gathered while reproducing:

- Thread dump: 3 of 4 XNIO worker threads parked on `appendLock` (`TimeSeriesShard.java:196`); the fourth inside `RaftGroupCommitter.submitAndWait`, reached from `appendSamples` → `RaftReplicatedDatabase.commit`.
- Measured ingest rate on a 3-node localhost cluster with ~19 ms inter-node RTT: **~18 samples/sec** (row count 3117 → 3488 over 20 s), i.e. ~11 minutes for the test's 12,000 samples against a 3-minute writer budget.
- The continuous `compactAll()` loop compounded it — every append commit waits out the active compaction recording session via `waitForActiveRecordingSession` — but the per-sample round trip alone already exceeded the budget.

| | before | after |
|---|---|---|
| `TimeSeriesThreeNodeWalGapReproIT` | fails after 191.8 s | **passes in 19.1 s** |
| append transactions per 200-sample request | 200 | ≤ 2 |

The `NullPointerException: ...getTransactionBroker() is null` seen in that test's log is downstream, not the cause: the assertion fails, `@AfterEach` stops the servers, and the still-in-flight HTTP writes then hit a torn-down `RaftHAServer`.

## Related issues

None — found while investigating the `TimeSeriesThreeNodeWalGapReproIT` failure.

## Additional Notes

`TimeSeriesIngestBatchAppendIT` guards the batching contract for **both** ingest handlers using the `writeTx` database statistic. It counts commits that produced WAL, so it counts shard append transactions directly — a deterministic assertion with no timing dependency. Verified to discriminate rather than merely pass: with the per-sample shape temporarily restored it measures 200 transactions in every case.

The line-protocol cases pin `SHARDS 1` in DDL and bind to `≤ 2`. The remote_write cases cannot: an auto-created type gets `ASYNC_WORKER_THREADS` shards and `appendBatch` opens at most one transaction per shard that received samples, so a hardcoded bound would vary with the CPU count of the machine running the test. Those assertions read the shard count back off the type instead.

Left alone deliberately, flagging for a follow-up: `RaftHAServer.getTransactionBroker()` returning `null` after `stop()` surfaces as a raw NPE to in-flight writers rather than a clean "server is shutting down" error. Harmless here now, but it will keep producing confusing stack traces on any shutdown-under-load.

Verification:

| Suite | Result |
|---|---|
| `TimeSeriesIngestBatchAppendIT` (new) | 4/4 — fails pre-fix (200 transactions), passes post-fix |
| server TS / Prometheus / Grafana / PromQL ITs | 58/58 |
| `engine` `com.arcadedb.engine.timeseries.*Test` | 284/284 |
| ha-raft TS ITs (incl. `Issue4458TsWalVersionGapIT`, `RaftTimeSeriesReplication3NodesIT`) | 6/6 |
| `TimeSeriesThreeNodeWalGapReproIT` | passes, 17.9 s, all 3 nodes converge to 12,000 samples |

## Checklist

- [ ] I have run the build using `mvn clean package` command — not run; verified with targeted `mvn -pl server` / `-pl engine` / `-pl ha-raft` builds and the test suites listed above


robfrank:
Addressed in 95f212b69. Notes per item.

**1. Prometheus tag hoisting — done, but the premise needs a correction.**

The optimization is right and is now implemented. The stated reason is not: this is not extra work introduced by batching. The pre-change code also looped `for (final Sample sample : ts.getSamples())`, so `findLabelValue` already ran `samples × tagColumns` times per series. The batching commit moved that work around but did not add any.

What is true, and what makes it worth fixing, is that the value is *invariant across samples* and the column-major grid makes it trivial to exploit — plus `findLabelValue` is worse than a plain linear scan: it calls `sanitizeColumnName` on every label it visits, which is 3 `String.replace` allocations per label. So the cost was `samples × tagColumns × labels × 3` allocations, in both the old and new shapes.

Filling the grid column by column instead of sample by sample lets each TAG column resolve once and broadcast down its column via `Arrays.fill`. Agreed that the line-protocol handler must stay per-sample — `sample.getTags()` genuinely varies per line — and it is unchanged.

**2. Prometheus batching regression test — done, and it needed the shard count.**

`PostTimeSeriesWriteHandlerBatchAppendIT` is renamed to `TimeSeriesIngestBatchAppendIT` and now owns the batching contract for both ingest handlers, with two remote_write cases (single series; two series of the same metric differing only by tag).

Writing it surfaced something worth flagging: my first attempt reused the line-protocol tests' hardcoded `≤ 2` bound and **failed at 11**. That is not a batching failure — an auto-created remote-write type gets `ASYNC_WORKER_THREADS` shards (not 1, unlike the line-protocol tests which pin `SHARDS 1` in DDL), and `appendBatch` opens at most one transaction per shard that received samples. 11 shards, 11 transactions, exactly the contract. A hardcoded bound there would have been machine-dependent and flaky across CI runners with different CPU counts, so the assertion now reads the shard count back off the type and binds to `shards + 1` (and asserts `shards < SAMPLES / 4` so the bound stays meaningful).

Verified the guard actually discriminates rather than merely passing: with the per-sample shape temporarily restored, both new cases fail at **200 transactions against a bound of 12**.

**3. Empty remote-write series auto-creating a type — not changing this.**

Correctly identified as pre-existing and not a regression; `getOrCreateType` ran before the (absent) sample loop too, so behaviour is byte-for-byte identical. I am leaving it: moving the empty check ahead of `getOrCreateType` is a semantic change to schema auto-creation on a path this PR only touches for batching, and a zero-sample series carrying labels is a plausible way for a client to pre-register a metric. That is a decision for someone with the Prometheus-compat context, not a drive-by in a perf PR. Happy to file it separately if wanted.

**4. Partial-write window — documented in both handlers.**

The claim checks out, and there is direct evidence for it in this PR: the `writeTx` statistic incremented 200 times for a 200-sample request *while the handler's outer transaction was open*, which only happens if the inner shard commits are independent. Both handlers now carry a note at `database.begin()` stating that it does not make the request atomic, that `TimeSeriesShard.appendSamples` commits its own shard writes via `getWrappedDatabaseInstance()`, and that a later failure cannot roll back batches already written.

**Minor (schema lookups before `begin()`)** — intentional and safe. They are metadata reads that need no transaction, and the resolved `LocalTimeSeriesType` is carried forward in the grouping structure, so the append pass does no second lookup that could observe a different schema.

Verification after these changes: server TS/Prometheus/Grafana/PromQL ITs 58/58; `TimeSeriesThreeNodeWalGapReproIT` still passes (17.9 s); ha-raft TS ITs 6/6.

PR #5451, opened 2026-07-27
fix(#5443): a follower's index can be short after an LSM compaction

### LINKED ISSUE #5443: HA: a follower's index can be incomplete after an LSM compaction, silently returning fewer rows
Found while working on #4743 (round 4). **Not** caused by that work: it reproduces identically on the code before those commits, and with the stock `arcadedb.ha.appendBufferSize` where the compaction entry is never split at all.

## Symptom

After an LSM index compaction on a 3-node Raft cluster, a **follower's index can hold fewer entries than the leader's**, while every record replicates correctly. Specific keys become unfindable on that follower, so a query answered by it silently returns fewer rows than the same query on the leader.

Measured in `Issue4743CompactionEntrySizeIT` (60k vertices, unique high-entropy keys, one explicit `compact()`):

```
leader   -> index countEntries() = 60000, records = 60000
follower -> index countEntries() = 49488, records = 60000
```

The gap is stable across runs and identical with and without the entry splitting from #4743, which is how it was isolated.

## Secondary symptom (same root, probably)

While applying a compaction, followers log a transient:

```
SEVER [LSMTreeIndexMutable] Invalid sub-index for index '<name>', ignoring it. WARNING: This could lead on
using partial indexes. Run 'REBUILD INDEX `<name>`' to repair it (error=File with id 'NNN' was not found)
```

7-8 occurrences per compaction in that test, timing dependent. One trigger was fixed in #4743 (`applySchemaEntry` retired the superseded compacted file *before* the WAL that repoints the index at its replacement, so the schema update re-instantiated the index against a file that had just been deleted). At least one more trigger remains: a component instantiated from a file whose pages have not fully arrived yet.

Note this warning used to be far more damaging - `LSMTreeIndexMutable.onAfterLoad` responded to it by calling `dropIndex()`, which on a replica is an illegal schema write that failed the apply and escalated the whole database to a snapshot resync. That part is fixed; what remains is the incomplete index.

## Likely area

`ArcadeStateMachine.applySchemaEntry` orders the apply as: create files → install sealed blobs → publish schema JSON → apply WAL pages → retire old files → reload schema. The schema publish (`LocalSchema.update`) re-instantiates index components, and it runs *before* the WAL that fills those files. `TransactionManager.applyChanges` also skips pages for files not registered in the schema. So there is a window in which a follower builds its in-memory view of a compacted index from a file that is empty or partially written, and nothing re-derives it afterwards.

Same family as #4083.

## Impact

Silent wrong results on followers for indexed lookups after a compaction. Reads served by a follower (`READ_CONSISTENCY` allowing it, or a follower-targeted client) can miss rows that exist. No data loss - the records are present, only the index is short.

## Workaround

`REBUILD INDEX` on the affected node.

## Reproduce

`ha-raft/src/test/java/com/arcadedb/server/ha/raft/Issue4743CompactionEntrySizeIT.java` already builds the scenario; it deliberately does *not* assert follower index completeness (with a comment saying why, pointing here). Re-enable that assertion to get a failing test:

```java
assertThat(serverIdx.countEntries()).isEqualTo(TOTAL_RECORDS);   // fails on followers
```

lvca:
Closes #5443. Stacked on #5449 (the shutdown-hook fix), which is what makes these HA suites runnable; base retargets to main once that merges.

After an LSM index compaction on a 3-node Raft cluster, a follower's index can hold fewer entries than the leader's while every record replicates: leader `countEntries()` = 60000, follower = 49488, records = 60000 on both. Specific keys are simply unfindable on that node, so a query served there silently returns fewer rows. No data is lost - only the index is short - and `REBUILD INDEX` on the affected node repairs it.

Two independent causes, both fixed here.

## 1. An incremental compaction appends to an existing file, and nothing replicated those pages

A compaction round does not only create files. An **incremental** round appends a new series to the already-existing compacted file. Those pages are written eagerly and deliberately without WAL, and the file is not in `addFiles` because it is not new - so nothing shipped them. A temporary trace showed the session serializing compacted file 5 at 10 pages while the leader ended at 13.

`RaftReplicatedDatabase.runWithCompactionReplication` now snapshots page counts before invoking the compaction and ships the range that grew. Two details were needed to make that correct:

- **the root page must travel with the appended range.** It carries the series registry, so an appended series is invisible until page 0 says it exists. Shipping only `[before, after)` gave followers the right 13 pages and still 49488 entries; prepending page 0 gave 60000.
- **page count and content must come from the page manager, not the file.** `PaginatedComponentFile.getTotalPages()` is `channel.size() / pageSize`, which sees only what has reached disk.

## 2. A split entry reloaded the schema from a half-delivered state, and the detach was sticky

When the schema change is large enough to be split across several Raft entries - which a big index does at the stock `appendBufferSize` - a follower could end up serving only its mutable pages, its compacted sub-index gone entirely (~1897 of 60000).

Only the last chunk publishes, since it carries the schema JSON; earlier chunks only deliver pages. But `applySchemaEntry` reloaded the schema at the end of every chunk that was not "WAL only", and the first chunk carries `filesToAdd`, so it reloaded there too, re-instantiating components from a half-delivered state. `LSMTreeIndexMutable.onAfterLoad` then could not resolve the compacted sub-index and detached it - which is safe in itself, except the detach is **sticky**: the later publication reuses the same in-memory component, so the follower never re-attached.

The guard now keys off an explicit `moreChunksFollow` flag on `SCHEMA_ENTRY`, set on every chunk but the last.

Inferring it instead - "has WAL, no schema JSON" - is unsound, and a full IT batch proved it: a legitimate standalone DDL that adds or removes files without moving the schema version has exactly that shape, and skipping its reload leaves the new files unregistered, surfacing much later as a type missing on one node (`Issue4749SnapshotInstallClosesRegisteredDatabaseIT`, "Types: DB1 5 <> DB2 6"). The flag is a trailing self-describing boolean, the same backward-compatible pattern the WAL and sealed-blob sections already use: an older codec stops earlier and reads false, i.e. pre-split behaviour. Every non-split entry behaves exactly as before.

## Verification

- `Issue5443FollowerIndexGapIT` - the gap scenario, red on the previous code
- `Issue5443SplitCompactionEntryIT` - same scenario with `appendBufferSize=512KB` to force the split path
- `Issue4743CompactionEntrySizeIT` - now asserts follower index completeness too, which it deliberately did not before (its comment pointed here)
- `ha-raft`: 740 unit tests, 0 failures
- `RaftLogEntryCodecTest`: 33 tests. `corruptedSealedBlobFailsCrc` flipped the entry's last byte to corrupt the blob; that byte is now the flag, so it targets `length - 2`.

A full ~85-class HA IT batch completes with one varying failure per run, each of which passes in isolation - batch-order flakiness rather than a regression. Worth knowing when reading CI: concurrent builds on the same 2434/2480-2482 ports cause `BindException` cascades.

### DIFF
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -1432,15 +1432,19 @@ public class ArcadeStateMachine extends BaseStateMachine {
     final boolean sealedOnlyEntry = isEmptyMap(decoded.filesToAdd()) && isEmptyMap(decoded.filesToRemove())
         && decoded.sealedFileBlobs() != null && !decoded.sealedFileBlobs().isEmpty();
 
-    // #4743: an intermediate chunk of a split schema change (see
-    // RaftTransactionBroker.replicateSchemaInChunks) carries WAL pages only - no files, no schema JSON,
-    // no sealed blobs. The change is published by the LAST chunk, so reloading the schema here would
-    // re-instantiate every component from the still-unchanged schema.json for nothing, on the single
-    // Raft apply thread, once per chunk. Apply the pages and move on.
-    final boolean walOnlyEntry = isEmptyMap(decoded.filesToAdd()) && isEmptyMap(decoded.filesToRemove())
-        && (decoded.schemaJson() == null || decoded.schemaJson().isEmpty())
-        && (decoded.sealedFileBlobs() == null || decoded.sealedFileBlobs().isEmpty())
-        && decoded.walEntries() != null && !decoded.walEntries().isEmpty();
+    // A non-final chunk of a schema change split across several entries (see
+    // RaftTransactionBroker.splitSchemaEntry) only DELIVERS pages: the change is published by the last
+    // chunk. Reloading the schema on such a chunk re-instantiates every component from a state that is
+    // still half-delivered - and that is not merely wasted work on the single Raft apply thread, it is
+    // STICKY: a compacted sub-index that cannot be resolved yet gets detached, and the later publication
+    // reuses the same in-memory component, so the follower keeps serving only its mutable pages for good
+    // (#5443: ~1897 of 60000 entries).
+    //
+    // The producer marks these chunks explicitly. Inferring them from "no schema JSON" was tried and is
+    // WRONG: the first chunk carries filesToAdd and no schema JSON, which is indistinguishable from a
+    // standalone DDL that adds files without changing the schema version - and skipping the reload for
+    // that would leave the new files unregistered in the schema.
+    final boolean deliveryOnlyEntry = decoded.moreChunksFollow();
 
     try {
       if (decoded.filesToAdd() != null)
@@ -1496,8 +1500,8 @@ public class ArcadeStateMachine extends BaseStateMachine {
       // Reload schema after WAL pages are on disk so new index files have valid content
       // and are correctly registered (page counts, type links, in-memory structures).
       // Skipped for sealed-only TimeSeries compaction entries (see sealedOnlyEntry above) and for
-      // intermediate chunks of a split schema change (see walOnlyEntry above).
-      if (!sealedOnlyEntry && !walOnlyEntry)
+      // delivery-only chunks of a split schema change (see deliveryOnlyEntry above).
+      if (!sealedOnlyEntry && !deliveryOnlyEntry)
         db.getSchema().getEmbedded().load(ComponentFile.MODE.READ_WRITE, true);
 
     } catch (final IOException e) {
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
@@ -245,6 +245,9 @@ public final class RaftLogEntryCodec {
         dos.write(compressed);
       }
 
+      // KEEP THIS LAST. The decoder detects it with available() > 0, so it can only be the final field:
+      // anything appended after it would make a false flag indistinguishable from an older entry that
+      // never wrote one. A future trailing section needs its own presence byte written BEFORE this.
       dos.writeBoolean(moreChunksFollow);
 
       dos.flush();
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
@@ -80,7 +80,16 @@ public final class RaftLogEntryCodec {
       long bootstrapLastTxId,
       // TimeSeries sealed-store blobs embedded in a SCHEMA_ENTRY (issue #4382). Empty for all other
       // entry types and for SCHEMA_ENTRYs produced by nodes that predate this section.
-      List<TsSealedBlob> sealedFileBlobs
+      List<TsSealedBlob> sealedFileBlobs,
+      /**
+       * True on every chunk of a SCHEMA_ENTRY that was split across several Raft entries except the last
+       * (issue #5443). Such a chunk only DELIVERS pages - the change is published by the final chunk - so
+       * the applier must not reload the schema for it. This is carried explicitly rather than inferred:
+       * a chunk that creates files but carries no schema JSON is indistinguishable from a legitimate
+       * standalone DDL that adds files without changing the schema version, and skipping the reload for
+       * THAT would leave the new files unregistered.
+       */
+      boolean moreChunksFollow
   ) {
   }
 
@@ -162,6 +171,20 @@ public final class RaftLogEntryCodec {
       final Map<Integer, String> filesToAdd, final Map<Integer, String> filesToRemove,
       final List<byte[]> walEntries, final List<Map<Integer, Integer>> bucketDeltas,
       final List<TsSealedBlob> sealedFileBlobs) {
+    return encodeSchemaEntry(databaseName, schemaJson, filesToAdd, filesToRemove, walEntries, bucketDeltas,
+        sealedFileBlobs, false);
+  }
+
+  /**
+   * @param moreChunksFollow marks a non-final chunk of a schema change split across several entries
+   *                         (issue #5443). Written as a trailing self-describing byte, so a node running
+   *                         an older codec simply stops after the sealed-blob section and decodes it as
+   *                         false - the pre-split behaviour.
+   */
+  public static ByteString encodeSchemaEntry(final String databaseName, final String schemaJson,
+      final Map<Integer, String> filesToAdd, final Map<Integer, String> filesToRemove,
+      final List<byte[]> walEntries, final List<Map<Integer, Integer>> bucketDeltas,
+      final List<TsSealedBlob> sealedFileBlobs, final boolean moreChunksFollow) {
     try {
       final ByteArrayOutputStream baos = new ByteArrayOutputStream();
       final DataOutputStream dos = new DataOutputStream(baos);
@@ -212,6 +235,8 @@ public final class RaftLogEntryCodec {
         dos.writeInt(compressed.length);   // compressed length
         dos.write(compressed);
       }
+
+      dos.writeBoolean(moreChunksFollow);
 
       dos.flush();
       return ByteString.copyFrom(baos.toByteArray());
@@ -346,7 +371,7 @@ public final class RaftLogEntryCodec {
       final RaftLogEntryType type = RaftLogEntryType.fromId(typeByte);
       if (type == null)
         return new DecodedEntry(null, null, null, null, null, null, null, null, null, null, false, null, -1L,
-            Collections.emptyList());
+            Collections.emptyList(), false);
       final String databaseName = dis.readUTF();
 
       final DecodedEntry result = switch (type) {
@@ -354,7 +379,7 @@ public final class RaftLogEntryCodec {
         case SCHEMA_ENTRY -> decodeSchemaEntry(dis, databaseName);
         case INSTALL_DATABASE_ENTRY -> decodeInstallDatabaseEntry(dis, databaseName);
         case DROP_DATABASE_ENTRY -> new DecodedEntry(RaftLogEntryType.DROP_DATABASE_ENTRY, databaseName,
-            null, null, null, null, null, null, null, null, false, null, -1L, Collections.emptyList());
+            null, null, null, null, null, null, null, null, false, null, -1L, Collections.emptyList(), false);
         case SECURITY_USERS_ENTRY -> decodeSecurityUsersEntry(dis);
         case BOOTSTRAP_FINGERPRINT_ENTRY -> decodeBootstrapFingerprintEntry(dis, databaseName);
       };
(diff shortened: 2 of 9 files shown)


lvca:
Thanks - the per-byte copy is a fair hit and it is fixed in `73a030a8d`.

You are right about where it came from: main's version did `walBuf.put(pageBuf.array(), PAGE_HEADER_SIZE, deltaSize)`, and `.array()` was dropped when the source moved from the raw file buffer to `page.getContent()`. The loop was the lazy way to stop depending on `array()`, and it costs a whole page minus header per page, per compaction, on the schema-commit thread.

I went with the absolute bulk **put** rather than the absolute get you suggested:

```java
walBuf.put(walBuf.position(), page.getContent(), BasePage.PAGE_HEADER_SIZE, deltaSize);
walBuf.position(walBuf.position() + deltaSize);
```

Same single transfer, and it avoids `array()` on *both* sides instead of only the source. `walBuf` is heap-allocated today so `walBuf.array()` would have worked, but the destination is an implementation detail of this method and the page buffer is shared - reading neither position keeps this immune to both. `ByteBuffer.put(int, ByteBuffer, int, int)` is Java 16+, so it is available on the 21 baseline.

The invariant point is a good one and I have written it into `pagesGrownDuringSession`'s javadoc, because it is exactly the thing a future compaction change would trip over: this ships page 0 plus `[pagesBefore, totalPages)` and therefore assumes a compaction only ever appends to a pre-existing file, touching no interior page beyond the root page that registers the appended series. A path that rewrote an interior page in place would silently not replicate and the follower's index would go short again. The comment says so, and says what to do about it (track dirtied pages rather than counting added ones).

On `corruptedSealedBlobFailsCrc` targeting `length - 2`: agreed it is format-sensitive, and I have left it. Worth noting for whoever touches the codec next that the test's assumption is "the last byte of the compressed blob is load-bearing", which is true of the current compressor but is not guaranteed by anything.

Re-verified after the change: `Issue5443FollowerIndexGapIT`, `Issue5443SplitCompactionEntryIT` and `Issue4743CompactionEntrySizeIT` green, plus the 740 ha-raft unit tests. Those ITs assert follower index completeness, so they exercise the copy end to end - a wrong bulk transfer would show up as a short index, which is the bug this PR is about.

lvca:
All three addressed in `f0b605514`.

**1. Dead null fallback - confirmed and fixed.** `LocalSchema.getFileById()` throws `SchemaException` on both paths (`id >= files.size()` and a null slot), so `component != null` could never be false, and a non-paginated id would have died on the cast rather than degrading. It now asks `getFileByIdIfExists()` and tests the type, matching `snapshotPageCounts()`.

I kept the fallback rather than dropping it, but made it loud:

```java
if (proxied.getSchema().getEmbedded().getFileByIdIfExists(fileId) instanceof PaginatedComponent component)
  totalPages = component.getTotalPages();
else {
  totalPages = (int) file.getTotalPages();
  LogManager.instance().log(this, Level.WARNING, "... replicating the %d pages already on disk, which may be short if a flush is still pending", ...);
}
```

The on-disk count is the only number available for a file the schema does not know, and it is precisely the number that made this bug silent - so if we ever fall back to it, that should be visible rather than quiet. You are right that it is unreachable today: the ids come from `addFiles` or from `snapshotPageCounts()`, which already filtered on `PaginatedComponent`. The comment says so, so nobody mistakes it for a live path.

**2. Rolling upgrade - written into the codec doc.** Worth being precise about the exposure: a node still on the old codec keeps the *pre-fix* behaviour for split entries, so it can still detach a compacted sub-index for good and end up short. The wire format is compatible in both directions, but the fix only takes effect on a node once that node is upgraded, and the symptom is silent - fewer rows from that node, no error. The note also says what to do about a node that was live through a compaction under the old codec: `REBUILD INDEX` on it, or let a snapshot install replace its files.

**3. Test duplication - extracted.** `BaseCompactionIndexCompletenessTest` now owns the scenario and the assertions; each IT keeps only what makes it distinct. `Issue5443FollowerIndexGapIT` is 42 lines and `Issue5443SplitCompactionEntryIT` is 53 (the extra being the `HA_APPEND_BUFFER_SIZE` override), against 125 and 132 before. I kept a `@Test` in each subclass rather than inheriting one, so the two causes keep descriptive, distinct method names in the report.

On the `DecodedEntry` 15-arg record: agreed, and I left it alone as out of scope. For whoever picks it up - the risk is not evenly spread, since the trailing `false` arguments are all booleans and a transposition there would compile silently. Static factories per entry type would be the natural fix.

Re-verified after the change: `Issue5443FollowerIndexGapIT`, `Issue5443SplitCompactionEntryIT`, `Issue4743CompactionEntrySizeIT` and the 740 `ha-raft` unit tests, all green.

lvca:
Three of the four are in `9c79e732c`; the fourth I left as-is, deliberately.

**1. Duplicated frame constants - fixed.** They were computed independently in the chunking maths and in the buffer that gets written, which is the worst place for a drift: the two would disagree about the frame the buffer describes. `WAL_CHUNK_FRAMING_SIZE`, `walPageDeltaSize(pageSize)` and `walPerPageSize(pageSize)` now own it, and both call sites read from them.

**3. Test config - switched to 0, in both ITs.** You are right about the guard: `LSMTreeIndexMutable` tests `minPagesToScheduleACompaction > 1`, and it is the only live reader (the vector index's copy of that check is commented out), so 0 and 1 both disable - but 0 is what the setting documents and 1 reads like "compact as often as possible". `Issue4743CompactionEntrySizeIT` had the same `1`, so it moved too rather than leaving the two ITs disagreeing about how to say "off".

**4. Trailing-flag ordering - noted at the write site**, since that is where someone would append a field:

```java
// KEEP THIS LAST. The decoder detects it with available() > 0, so it can only be the final field:
// anything appended after it would make a false flag indistinguishable from an older entry that
// never wrote one. A future trailing section needs its own presence byte written BEFORE this.
```

**2. The unreachable WARNING branch - keeping it.** Your read is right that it can never fire today. I am keeping it precisely because of what this issue was: a follower silently serving a short index, where every layer degraded quietly instead of saying anything. If a caller ever does reach that path, the on-disk page count is the number that would make the bug come back, and I would rather pay one dead log site than have it happen mutely again. The comment says no current caller can reach it, so it should not mislead anyone into treating it as live.

Re-verified: `Issue5443FollowerIndexGapIT`, `Issue5443SplitCompactionEntryIT`, `Issue4743CompactionEntrySizeIT` and the 740 `ha-raft` unit tests, green.

lvca:
Both fixed in `6e16e55e2`, and the first one was a real miss on my part.

**1. `snapshotPageCounts()` - you are right, including about the comment.** I fixed the lookup in `appendFilePagesAsWal` last round and wrote that it now used "the same shape `snapshotPageCounts()` uses" without checking that the sibling actually used it. It did not: it called the throwing `getFileById()` while its `instanceof` reads as if an unknown id would be skipped.

The reachability is as you describe. `LocalSchema.registerFile()` pads with `files.add(null)` and a dropped file leaves the slot null, while this loop walks everything `FileManager` holds - so a single such id throws `SchemaException` out of `snapshotPageCounts()` and aborts `runWithCompactionReplication` entirely. Worse than the dead branch you flagged before, since this one is on the live path. Now `getFileByIdIfExists()`, with a comment saying why the throwing variant is wrong here rather than just asserting the two agree.

**2. Refactor leftovers - removed.** `deltaSize` was unused once the copy moved into `appendPageRangeAsWal`, and the "Per-chunk framing" comment was orphaned above two blank lines when the constant moved out. Both gone.

**On CI actually running these ITs - it does not, and that is worth knowing.** The main workflow runs

```
./mvnw verify ... -DexcludedGroups=slow,benchmark -Dsurefire.includes=**/*Test.java,**/*Suite.java
```

so the `@Tag("slow")` ITs are excluded twice over, and the one job that does run `-Dgroups=slow` is scoped to `-pl engine`. `ha-resilience-tests.yml` runs `-pl e2e-ha` with `-DskipTests`, so it does not cover these either. These three ITs run locally with `-DskipITs=false` - which is how I have been verifying every round - and would not run on a PR. That is a pre-existing gap in how `ha-raft` ITs are wired into CI rather than something this PR introduces, but it does mean the regression coverage here is only as good as someone remembering to run it. Worth a follow-up to give the `ha-raft` slow ITs a scheduled job, if you agree.

Re-verified after this round: `Issue5443FollowerIndexGapIT`, `Issue5443SplitCompactionEntryIT`, `Issue4743CompactionEntrySizeIT`, and the 740 `ha-raft` unit tests, all green.

lvca:
Both suggestions taken, in `54b2854f4`. The branch is also rebased onto main now that #5449 has merged, so this PR is down to its own seven commits and its base is `main`.

**1. Codec round-trip for the flag - added, and mutation-checked.** Two assertions, which are indeed the cheapest guard for the contract everything else rests on:

- `moreChunksFollowSurvivesTheRoundTrip` - a chunk encoded with `true` decodes as `true`, one with `false` as `false`.
- `anEntryWithoutTheTrailingFlagDecodesAsTheLastChunk` - takes an encoded entry, strips the final byte to produce exactly what a pre-#5443 encoder emits, and requires it to decode as a complete `SCHEMA_ENTRY` with `moreChunksFollow == false` and its schema JSON intact. That is the backward-compat half, and stripping the byte is the only way to test it honestly: the old overload delegates with `false`, so it still writes the byte.

I checked they actually bite rather than just pass: making the encoder always write `false` fails `moreChunksFollowSurvivesTheRoundTrip`, as it should.

**2. The unreachable fallback now throws.** You have talked me out of my previous answer, and the argument is the one this PR is about: warn-and-degrade would ship the on-disk page count, which is exactly the number that left followers with a short index. A caller that reaches it has a bug, and it should say so rather than quietly reintroduce #5443:

```java
if (!(proxied.getSchema().getEmbedded().getFileByIdIfExists(fileId) instanceof PaginatedComponent component))
  throw new IllegalStateException("Cannot replicate file id '" + fileId + "' of database '" + getName()
      + "': it is not a registered paginated component, so its real page count is unknown");
```

Note this is the *seek* side only. `snapshotPageCounts()` keeps skipping unknown ids, because there the unknown id is expected - it walks every file the `FileManager` holds and a dropped file legitimately has no schema component.

**3. `getTotalPages()` `long` vs `Integer`** - agreed, leaving as is; it matches the surrounding code and a page count that overflows an int is not a real scenario.

Verified after the rebase, not just before it: `RaftLogEntryCodecTest` (35 tests, +2), the 742 `ha-raft` unit tests, and all three ITs green.

lvca:
Correction to my earlier comment on this PR: **I was wrong that CI does not run these ITs. It does.**

`mvn-test.yml` has a dedicated `ha-integration-tests` job:

```
./mvnw verify -DskipTests -Pintegration -Pcoverage --batch-mode --errors --fail-never --show-version -pl ha-raft
```

`-Pintegration` sets `skipITs=false`, and failsafe carries no `excludedGroups` - the only one in the pom is commented out and belongs to surefire - so `@Tag("slow")` is not filtered on the IT side. It runs on every PR to main, every push to main, and on manual dispatch. `Issue5443FollowerIndexGapIT`, `Issue5443SplitCompactionEntryIT` and `Issue4743CompactionEntrySizeIT` therefore all ran on this PR.

My mistake was reading the `unit-tests` job's `-DexcludedGroups=slow,benchmark` and the `slow-unit-tests` job's `-pl engine`, checking `ha-resilience-tests.yml`, and concluding from those three that nothing covered ha-raft ITs - without looking for the job that does. Please disregard the follow-up I suggested for a scheduled ha-raft IT job; it would be redundant.

Two things that are true and worth keeping:

- The job runs with `--fail-never`, so Maven always exits 0 and the red/green comes from the `dorny/test-reporter` step, not the build. Deliberate, but the gate is the reporter's check rather than the Maven exit code.
- There is a real gap next door, in **unit** tests rather than ITs: `unit-tests` excludes the `slow` group across the whole reactor, and `slow-unit-tests` runs `-pl engine` only, so a `@Tag("slow")` unit test in any other module runs nowhere. That one is worth a follow-up.

I also ran the three ITs twice against merged main (`1134185d5`) after this landed, and mutation-checked that `Issue5443FollowerIndexGapIT` still fails when the incremental-append replication is neutralised - it reports "server 0 must hold the WHOLE index after the compaction", the original symptom.

PR #5468, opened 2026-07-27
feat(mcp)  : add per-principal tool profiles (#5468) (#5445)

### LINKED ISSUE #5445: MCP: support per-principal tool profiles on a shared endpoint
## Context

#5402 adds server-global MCP tool profiles. That narrows the endpoint safely, but it does not fully satisfy #4859's criterion that the surface a given agent sees stays small: a retrieval agent and an administrative agent sharing one server cannot currently receive different profiles.

This follow-up should make the effective profile depend on the authenticated principal while preserving the global profile as the backward-compatible default and security ceiling.

## Design requirements

- Resolve the effective profile after MCP authentication for both named users and API-token principals.
- Keep the global `profile` as the default when no principal override exists.
- Treat principal-specific profiles as allowlists and enforce them consistently in both `tools/list` and `tools/call`.
- Return initialization instructions that match the effective profile.
- Do not let a profile grant read, write, schema, admin, database, or user permissions denied by the existing permission layers.
- Define merge, removal, validation, persistence, and unknown-principal behavior for configuration updates.

## Acceptance criteria

- Two authenticated principals using the same MCP endpoint can receive different tool lists.
- A tool hidden from one principal cannot be invoked directly by name.
- Users without an override retain the global profile and current behavior.
- HTTP and standard input/output transports apply the same resolution rules.
- Tests cover named users, API-token principals, fallback, invalid configuration, discovery, and direct-call enforcement.

A short design spike should decide the configuration shape and whether principal profiles intersect the global profile or select from a separately configured allowed set.

justinblethrow-cloud:
## Summary

- add optional `principalProfiles` mappings for named users and API-token principals
- resolve each authenticated principal's effective tools as the intersection of its override and the global profile
- apply the same effective profile to initialization instructions, discovery, and direct invocation over HTTP and standard input/output
- define persistent merge, removal, clearing, validation, and unknown-principal behavior

The global profile remains the backward-compatible default and a hard ceiling. Principal profiles cannot grant a tool hidden globally, and the existing read, write, schema, administrative, database, user, and origin permission checks remain mandatory.

API tokens use their canonical authenticated name, `apitoken:<token-name>`, so they cannot be confused with a named user that shares the token's display name.

## Validation

Focused MCP tests:

```text
./mvnw -pl server -am test \
  -Dtest='MCPConfigurationTest,MCPStdioServerTest,MCPServerPluginTest,MCPPermissionsTest' \
  -Dsurefire.failIfNoSpecifiedTests=false
```

Result: 145 tests, 0 failures, 0 errors, 0 skipped.

A subsequent focused configuration regression run passed 22 tests after adding malformed top-level value coverage.

Closes #5445

### DIFF
--- a/docs/superpowers/specs/2026-07-27-mcp-principal-tool-profiles-5445-design.md
+++ b/docs/superpowers/specs/2026-07-27-mcp-principal-tool-profiles-5445-design.md
@@ -1,0 +1,91 @@
+# MCP per-principal tool profiles (#5445)
+
+**Issue:** [#5445](https://github.com/ArcadeData/arcadedb/issues/5445)
+**Epic:** [#4859 - MCP GraphRAG & Agent-Memory Surface](https://github.com/ArcadeData/arcadedb/issues/4859)
+**Date:** 2026-07-27
+**Status:** implemented
+
+## Decision
+
+The Model Context Protocol (MCP) configuration may restrict individual authenticated
+principals to one of the existing `all`, `rag`, or `admin` tool profiles. The server-global
+`profile` remains the backward-compatible default and a hard ceiling.
+
+For a principal with an override, a tool is available only when both the global profile and
+the principal profile contain it. This intersection means an override can narrow the endpoint
+but cannot expose a tool hidden globally. Existing read, write, schema, administrative,
+database, user, and origin permissions remain independent mandatory checks.
+
+## Configuration
+
+Overrides live under `principalProfiles` in `config/mcp-config.json`:
+
+```json
+{
+  "enabled": true,
+  "profile": "all",
+  "allowedUsers": ["retrieval-agent", "retrieval-token"],
+  "principalProfiles": {
+    "retrieval-agent": "rag",
+    "apitoken:retrieval-token": "rag"
+  }
+}
+```
+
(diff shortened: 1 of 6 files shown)


justinblethrow-cloud:
Continuous-integration note: the second full run has completed. The Model Context Protocol patch-coverage check now passes after `f13d559`, and the build plus every client end-to-end suite are green.

The red test jobs are all outside this pull request's six changed Model Context Protocol files:

- unit: `ExplicitLockingTransactionTest.errorOnExplicitLock`
- slow unit: `Issue5381FalseConflictTest.hotVertexEdgeAppendPreservesAllEdges`
- integration: `PostgresQueryMetricsIT.postgresQueriesTaggedWithPostgresProtocol`
- high availability: `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`

The coverage-report job also lacked a Codacy token and could not find the slow-unit and high-availability artifacts; Codecov's patch check itself is green. The first run failed a different set of protocol-metrics/high-availability tests plus a Docker Hub pull timeout, and the base commit's own run was red on another unrelated unit test. I do not have permission to rerun jobs in the upstream repository. Could a maintainer please rerun the failed jobs? I do not see a Model Context Protocol code change indicated by these results.

PR #5479, opened 2026-07-28
fix(mcp) #5445: a principal profile keyed by a bare API-token name applies

robfrank:
Follow-up to #5445 / #5468, which merged as `ec3447a96`. Two defects found reviewing that change after the merge.

MCP accepts an API token in `allowedUsers` under either spelling: the canonical authenticated principal name `apitoken:<token-name>`, or the bare token name. The `principalProfiles` map added by #5468 accepted only the canonical one, and looked it up by exact match. So this configuration admits the token and then leaves it on the global profile:

```json
{
  "profile": "all",
  "allowedUsers": ["retrieval-token"],
  "principalProfiles": { "retrieval-token": "rag" }
}
```

The token authenticates as `apitoken:retrieval-token`, no override is found, and a configuration that reads as a confinement to the retrieval surface exposes `execute_command` and `set_server_setting` instead. There is no error, no warning, and nothing in the serialized configuration that reads as wrong: the entry is still there, and still says `rag`. The two keys spell the same principal the same way and disagree about who they address.

`getPrincipalToolProfile` now matches the canonical name first and falls back to the bare token name for an `apitoken:` principal, the same convention `isUserAllowed` already accepts, so an explicit `apitoken:<name>` entry stays authoritative when both spellings are configured. The fallback cannot widen the surface: a principal profile is only ever intersected with the global one (`MCPDispatcher.EffectiveToolProfile.allows`), so resolving one can remove tools and never add them. The worst case for a name collision is that a token is confined to a same-named user's profile, which is a narrowing, and it is the collision `matchesUser` already lives with on the allowlist. The reverse direction cannot collide at all, since `PostUserHandler` rejects creating a user whose name starts with `apitoken:`.

Second, `POST /api/v1/mcp/config` with a `principalProfiles` (or the pre-existing `databases`) value that is not an object answered `500` instead of `400`. Both `load()` and `updateFrom()` read the nested object with `getJSONObject(name, null)`, whose default applies only to JSON null, so a string value reached Gson's `getAsJsonObject()` and raised `IllegalStateException` - which `MCPConfigHandler` does not catch. A new `objectValue()` helper rejects it with `IllegalArgumentException`, which the handler already maps to `400`. The merged test asserted `IllegalStateException`, locking the `500` in; that assertion is replaced.

## Test plan

- [ ] `mvn -pl server test -Dtest=MCPConfigurationTest` - 24 tests, including the new `apiTokenPrincipalProfileMatchesBareTokenName` (bare-name entry resolves for an `apitoken:` principal, a canonical entry wins when both are configured, a named user is never matched by the fallback) and the extended `invalidPrincipalProfileIsRejectedWithoutPartialUpdate` (non-object `principalProfiles` and non-object `databases` both rejected as a client error, sibling settings in the same update not applied).
- [ ] `mvn -pl server test -Dtest=MCPServerPluginTest` - includes the new `apiTokenPrincipalProfileAcceptsBareTokenName`, end to end over HTTP: a token allowlisted and profiled by its bare name sees the `rag` tool list and is denied `server_status` on a direct `tools/call`.
- [ ] `mvn -pl server test -Dtest='MCPConfigurationTest,MCPStdioServerTest,MCPServerPluginTest,MCPPermissionsTest,MCPResourcesTest,MCPToolUtilsTest'` - 161 tests.

All of the above were run green on this branch. `apiTokenPrincipalProfileMatchesBareTokenName` was confirmed red against the unfixed code before the fix was applied.

Tracking notes, including the two known gaps left open (MCP Resources are not profile-gated, and `principalProfiles` is undocumented in the separate `arcadedb-docs` repository), are in `docs/5445-mcp-principal-profile-token-lookup.md`.

### DIFF
--- a/docs/5445-mcp-principal-profile-token-lookup.md
+++ b/docs/5445-mcp-principal-profile-token-lookup.md
@@ -141,7 +141,7 @@ Observations raised and how they were resolved:
   parsed up front. The test's payload is rejected atomically only because the invalid field is one of the three
   parsed first. The behaviour predates this PR - the inline boolean block arrives with #5402 (`68d6596dc`) - so
   it is left alone here and the test now carries a comment stating exactly what it proves. Making `updateFrom`
-  atomic for every field is the follow-up.
+  atomic for every field is the follow-up, tracked as #5482.
 
 ## Known gaps
 
@@ -154,8 +154,9 @@ Observations raised and how they were resolved:
 - **`principalProfiles` is undocumented.** The MCP reference page lives in the separate
   `ArcadeData/arcadedb-docs` repository at `src/main/asciidoc/reference/mcp/mcp.adoc` and documents
   the `mcp-config.json` keys. It does not mention `principalProfiles`, nor the `profile` key from
-  #5402, nor the `databases` key from #5401. A docs PR is owed there.
+  #5402, nor the `databases` key from #5401, and it never mentions the Resources surface from #4865.
+  Tracked as #5483.
 - **`updateFrom` is not atomic across every field.** Raised in review and confirmed: the `allow*` booleans are
   assigned inline, so a payload whose invalid field is a boolean commits the booleans that precede it before
   throwing. Pre-existing since #5402 and untouched here. The fix is to parse every field into locals before
-  assigning any, mirroring what `databases`, `profile` and `principalProfiles` already do.
+  assigning any, mirroring what `databases`, `profile` and `principalProfiles` already do. Tracked as #5482.
--- a/docs/5445-mcp-principal-profile-token-lookup.md
+++ b/docs/5445-mcp-principal-profile-token-lookup.md
@@ -1,6 +1,6 @@
 # Issue #5445 follow-up - a principal profile written with a bare API-token name is silently inert
 
-PR: (to be filled after opening)
+PR: https://github.com/ArcadeData/arcadedb/pull/5479
 
 Follow-up to #5468, which implemented per-principal MCP tool profiles and merged as `ec3447a96`.
 Two defects found reviewing that change after the merge.
--- a/docs/5445-mcp-principal-profile-token-lookup.md
+++ b/docs/5445-mcp-principal-profile-token-lookup.md
@@ -118,6 +118,31 @@ fix was applied.
 | `server` `MCPToolUtilsTest` | 4/4 |
 | `server` MCP suite (`MCPConfigurationTest,MCPStdioServerTest,MCPServerPluginTest,MCPPermissionsTest,MCPResourcesTest,MCPToolUtilsTest`) | 161/161 |
 
+## Review cycles
+
+| # | Head | Review outcome | Applied |
+|---|---|---|---|
+| 1 | `1ef55c671` | LGTM, two non-blocking observations | Scoped the `invalidPrincipalProfileIsRejectedWithoutPartialUpdate` claim with a comment - see below. Nothing else applied. |
+
+`gemini-code-assist` did not review this head; `claude[bot]` posted twice.
+
+Observations raised and how they were resolved:
+
+- *`objectValue` throws `IllegalArgumentException` while the sibling `booleanValue` throws `JSONException` for
+  the same class of wrong-type error.* Not changed. `IllegalArgumentException` is the convention the rest of
+  `MCPConfiguration` already uses for a rejected configuration value - blank override and principal names, an
+  unknown override key, an unknown profile name, a non-string profile, a null tool profile. `booleanValue` is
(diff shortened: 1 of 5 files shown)


robfrank:
Review cycle 1 responses. Both observations were checked against the code; one is applied, one is declined with reasoning.

**Exception-type inconsistency between `objectValue` and `booleanValue` - declined.**

`IllegalArgumentException` is the convention the rest of `MCPConfiguration` already uses for a rejected configuration value: blank database-override names, blank principal names, an unknown override key, an unknown profile name, a non-string profile value, a null tool profile. `booleanValue`'s `JSONException` is the single outlier in the class, so aligning `objectValue` with it would spread the inconsistency rather than remove it. Both are caught by the same `catch (IllegalArgumentException | JSONException)` in `MCPConfigHandler` and map to 400, so nothing observable depends on the choice. Left as is.

**The "without partial update" guarantee is narrower than the test name suggests - confirmed, and now scoped.**

Verified with a throwaway probe rather than by reading:

```
updateFrom({"enabled": true, "allowReads": "yes"})
  -> throws, and isEnabled() == true afterwards
```

So the observation is exactly right: the `allow*` booleans are assigned inline, while `databases`, `profile` and `principalProfiles` are parsed into locals up front. The test's payload is rejected atomically only because its invalid field happens to be one of the three parsed first.

`git log -S` puts the inline boolean block at `68d6596dc` (#5402), so it predates both #5468 and this PR. I have left the behaviour alone here rather than widen a security fix into a config-endpoint refactor, and instead put a comment on the test stating precisely what it proves and what it does not:

```java
// The "without partial update" this asserts comes from field ordering: databases, profile and
// principalProfiles are parsed before updateFrom mutates anything, so rejecting one of them leaves every
// other setting alone. It is not general atomicity - the allow* booleans are still assigned inline, so an
// invalid boolean can commit the booleans that precede it. That is out of scope here.
```

Making `updateFrom` atomic across every field - parse all of them into locals before assigning any, mirroring what the three already do - is recorded as a known gap in `docs/5445-mcp-principal-profile-token-lookup.md` along with the owed `arcadedb-docs` PR for the undocumented `principalProfiles` / `profile` / `databases` keys.

Both are now in the tracking note's Review cycles section. `mvn -pl server test` over the six MCP classes is 161/161 on `d7a75bcd8`.

robfrank:
Both follow-ups from review cycle 1 are now tracked, and the known-gaps section of `docs/5445-mcp-principal-profile-token-lookup.md` links to them:

- #5482 - `updateFrom` applies a prefix of a rejected configuration payload. Carries the probe that reproduces it and dates the inline boolean block to #5402, so it is clear this is not a regression from #5468 or this PR. The suggested fix also folds in the `objectValue` / `booleanValue` exception-type alignment, which is the natural moment for it.
- #5483 - `databases` (#5401), `profile` (#5402) and `principalProfiles` (#5468, #5479) are all undocumented in the separate `arcadedb-docs` repository, along with the Resources surface from #4865.

Nothing in this PR changes as a result.

robfrank:
Review cycle 2. No code changes; one item is settled, the other reaches the right conclusion from a wrong premise.

**1. `objectValue` / `booleanValue` exception types.** Same as cycle 1, and we agree it is fine to leave. Nothing to add.

**2. `load()` raising `IllegalArgumentException` where it used to raise `IllegalStateException`.**

The conclusion is right - not a regression, and no caller is affected - but the stated reason is not. Neither exception ever propagated out of `load()`. Its `catch (Exception)` absorbs both and falls back to defaults:

```java
  ... "Error loading MCP configuration: %s"
} catch (final Exception e) {
  ... "Corrupt MCP configuration file, using defaults: %s"
}
```

`IllegalStateException` and `IllegalArgumentException` are both `RuntimeException`, so both were and are caught there. Probed on `810a30832` with a `config/mcp-config.json` holding `{"enabled": true, "principalProfiles": "rag"}`:

```
load() threw nothing
WARNING  Corrupt MCP configuration file, using defaults:
         MCP configuration field 'principalProfiles' must be an object
isEnabled() == false
```

So the type is unobservable from outside `load()` - which is a stronger guarantee than "I did not find such a caller", and it is why the search came up empty. For completeness, `ArcadeDBServer:314` is the only caller and invokes `load()` bare.

Worth noting what that probe does show, since it is easy to miss: one malformed field discards the *entire* file, so `"enabled": true` is dropped too and the server starts with MCP disabled. That is pre-existing fail-safe behaviour and correct for a security-relevant config, but it is the load-time analogue of #5482 and is now written down.

The one real difference is the log line: it now names the offending field rather than reporting Gson's `Not a JSON Object: "rag"`, which is the improvement the review noted.

Recorded as cycle 2 in `docs/5445-mcp-principal-profile-token-lookup.md`. `mvn -pl server test` over the six MCP classes remains 161/161.

Issue #5480, opened 2026-07-28
Node inline `WHERE` predicates are ignored in `MATCH`, `EXISTS {}` and pattern comprehensions


robfrank:
### ArcadeDB version

Reproduced against `main` at `043309f61` (`fix(#5460)`, 2026-07-28).

### Environment

- Engine module, OpenCypher query language
- Reproduced via `engine` unit test against an embedded database

### Description

The node inline `WHERE` predicate - the `WHERE n.v = 2` in `(n:A WHERE n.v = 2)` - is accepted by the grammar but never enforced. Every candidate node is returned as though the predicate were absent.

This affects at least three query contexts: plain `MATCH`, `EXISTS {}` subqueries, and pattern comprehensions.

Note this is the **node** counterpart of the relationship inline `WHERE` form. The relationship form was fixed for `MATCH` in #3953, for `EXISTS {}` in #5464, and for pattern comprehensions in #5460. The node form was left behind in all three.

Partial infrastructure already exists as of #5464: `NodePattern` has a `whereExpression` field and `getWhereExpression()`, `CypherASTBuilder.visitNodePattern(...)` parses `ctx.expression()` into it, and `rewriter/InlineNodeWhereHoister.hoist(...)` is invoked from `CypherASTBuilder` when a `MatchClause` is built. Despite that wiring, the predicate does not reach evaluation in the cases below. The root cause has not been isolated.

### Steps to reproduce

Data:

```cypher
CREATE (a:A {v: 1}), (b:A {v: 2}), (c:A {v: 3}),
       (a)-[:E]->(b), (a)-[:E]->(c);
```

#### 1. Plain `MATCH`

```cypher
MATCH (n:A WHERE n.v = 2) RETURN n.v AS v ORDER BY v;
```

Expected `v = 2` only. Actual: `1`, `2`, `3`.

#### 2. `EXISTS {}` subquery

```cypher
MATCH (a:A {v: 1})
WHERE EXISTS { (a)-[:E]->(x:A WHERE x.v = 99) }
RETURN a.v AS v;
```

No neighbor of `a` has `v = 99`, so the row must be filtered out. Actual: the row is returned.

#### 3. Pattern comprehension

```cypher
MATCH (a:A {v: 1})
RETURN size([(a)-[:E]->(x:A WHERE x.v = 2) | x]) AS c;
```

Expected `c = 1`. Actual: `c = 2`.

### Expected behavior

[...]

Issue #5460, opened 2026-07-27
Pattern comprehension ignores relationship inline `WHERE` predicates


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`

Build:

```text
1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main
```

### Environment

* Linux x86_64, Docker, OpenJDK
* `arcadedata/arcadedb:latest`
* HTTP command API
* Cypher query language

### Description

Pattern comprehensions silently ignore relationship inline `WHERE` predicates.

A trivially false predicate such as `WHERE 1=0` still returns all matching relationships, as though the predicate were absent. A predicate that accesses the relationship variable, such as `WHERE r.tag = "ok"`, is also ignored.

This is related to #3951, which reported that relationship inline predicates of the form:

```cypher
[r:TYPE WHERE predicate]
```

were ignored in regular `MATCH` patterns. That issue was fixed through #3953.

The current report exercises the same syntax in a different query context: pattern comprehension. On the tested `26.8.1-SNAPSHOT` build, regular `MATCH` applies the inline predicate correctly, while pattern comprehension ignores it.

This is also related to #5111, which covers relationship inline property maps such as:

```cypher
[:E {tag: "ok"}]
```

being ignored inside pattern comprehensions.

However, #5111 concerns property-map syntax, while this report concerns the distinct inline `WHERE` syntax:

```cypher
[r:E WHERE condition]
```

The two forms may share an implementation path, but the current evidence does not establish a common root cause.

### Steps to reproduce

Create the test data:

```cypher
MATCH (n) DETACH DELETE n;

CREATE (a:A {v: 1}), (b:A {v: 2});

MATCH (a:A {v: 1}), (b:A {v: 2})
CREATE (a)-[:E {tag: "ok"}]->(b),
       (a)-[:E {tag: "bad"}]->(b);
```

#### Trivially false relationship inline predicate

```cypher
MATCH (a:A {v: 1})
RETURN size([
  (a)-[r:E WHERE 1=0]->(b:A)
  | b
]) AS c;
```

Expected:

```text
c = 0
```

Actual:

```text
c = 2
```

#### Predicate accessing the relationship variable

```cypher
MATCH (a:A {v: 1})
RETURN size([
  (a)-[r:E WHERE r.tag = "ok"]->(b:A)
  | b
]) AS c;
```

Expected:

```text
c = 1
```

Actual:

```text
[...]

Issue #5096, opened 2026-07-07
`shortestPath` and `allShortestPaths` do not enforce accepted edge property filters


shulei5831sl:
### ArcadeDB version

Observed on local Docker image:

```text
docker inspect arcadedb -> arcadedata/arcadedb:latest sha256:837059175046894828a87536e2b9877f20194e71d3f9ab5b3e843c102dc61ce4

docker exec arcadedb find /home/arcadedb/lib -name 'arcadedb-server-*.jar'
/home/arcadedb/lib/arcadedb-server-26.7.2-SNAPSHOT.jar

Endpoint:
http://127.0.0.1:2480/api/v1/command/test
```

Also reproduced on the previous build (sha256:2f63d3ebbf08..., build 852e924).

### Environment

* Host OS: Linux
* Deployment: Docker
* Query interface: HTTP `/api/v1/command/test`
* Query language: Cypher / OpenCypher
* Reference behavior:
  * Neo4j Community `5.26.28`
  * Memgraph `3.11.0`
  * FalkorDB graph module `41811`

### Description

ArcadeDB's `shortestPath` and `allShortestPaths` accept an edge property
filter on variable-length patterns, but return paths through edges that
do not match the filter.

The pattern:

```cypher
shortestPath((a)-[:LINK*1..3 {w: 1}]->(b))
```

should only traverse edges where `w = 1`. Instead, it traverses edges
with `w = 2` and returns paths through them.

The same property filter on a regular variable-length path works correctly:

```cypher
MATCH (a)-[:LINK*1..3 {w: 1}]->(b)
```

Both `shortestPath` and `allShortestPaths` are affected. They may share
a path-pattern compilation or traversal-planning issue where the accepted
edge property filter is not enforced.

### To reproduce

Run from a clean database.

#### 1. Clean database

```cypher
MATCH (n) DETACH DELETE n;
```

#### 2. Create test data

```cypher
CREATE
  (n1:Node {id: 1}),
  (n2:Node {id: 2}),
  (n3:Node {id: 3}),
  (n4:Node {id: 4}),
  (n1)-[:LINK {w: 1}]->(n2),
  (n2)-[:LINK {w: 1}]->(n3),
  (n1)-[:LINK {w: 2}]->(n4),
  (n4)-[:LINK {w: 2}]->(n3);
```

The key fact: from node 1 to node 4 there is exactly one edge, with
`w = 2`. With the `{w: 1}` filter, node 4 should not be reachable
[...]

PR #5487, opened 2026-07-28
fix(#5481): enforce relationship inline WHERE and property map in both shortestPath evaluators

### LINKED ISSUE #5481: `shortestPath` ignores relationship inline `WHERE`; as an expression it also ignores the property map (#5096 covered only the `MATCH` form)
### ArcadeDB version

Reproduced against `main` at `043309f61` (`fix(#5460)`, 2026-07-28).

### Environment

- Engine module, OpenCypher query language
- Reproduced via `engine` unit test against an embedded database

### Description

`shortestPath` does not enforce relationship constraints consistently. There are two distinct evaluators and they behave differently:

- `executor/steps/ShortestPathStep` - used by the `MATCH p = shortestPath(...)` form.
- `ast/ShortestPathExpression` - used when `shortestPath(...)` appears as an expression, e.g. inside `RETURN`.

`ShortestPathExpression.evaluate(...)` reads only `rel.getTypes()` and `rel.getDirection()` from the relationship pattern. It never consults `getProperties()` or `getWhereExpression()`, so both the inline property map and the inline `WHERE` are silently dropped on that path.

#5096 ("`shortestPath` and `allShortestPaths` do not enforce accepted edge property filters") was closed as completed on 2026-07-07, but it fixed only the property map on the `MATCH` form. The matrix below is what `main` does today.

| Form | `{tag: 'ok'}` property map | inline `WHERE` |
|---|---|---|
| `MATCH p = shortestPath(...)` | enforced (fixed by #5096) | **ignored** |
| `RETURN shortestPath(...)` | **ignored** | **ignored** |

Related: relationship inline `WHERE` is already honored in plain `MATCH` (#3953), `EXISTS {}` (#5464) and pattern comprehensions (#5460). `shortestPath` is the remaining context. #5460 additionally made `CypherExpressionBuilder.parseRelationshipPattern(...)` populate the predicate on the AST node, so `ShortestPathExpression` now receives a predicate it does not act on - it is silently ignored rather than never parsed.

### Steps to reproduce

Data - a single relationship whose `tag` is `'bad'`:

```cypher
CREATE (a:A {v: 1}), (b:A {v: 2}), (a)-[:E {tag: 'bad'}]->(b);
```

#### 1. `MATCH` form, inline `WHERE` - ignored

```cypher
MATCH (a:A {v: 1}), (b:A {v: 2})
MATCH p = shortestPath((a)-[r:E* WHERE 1=0]->(b))
RETURN count(p) AS c;
```

Expected `c = 0` (the predicate is never satisfiable). Actual: `c = 1`.

Control - the property map on the same form is correctly enforced, returning `c = 0`:

```cypher
MATCH p = shortestPath((a)-[:E* {tag: 'ok'}]->(b)) RETURN count(p) AS c;
```

#### 2. Expression form - both constraints ignored

```cypher
MATCH (a:A {v: 1}), (b:A {v: 2})
RETURN size(shortestPath((a)-[:E* {tag: 'ok'}]->(b))) AS c;
```

```cypher
MATCH (a:A {v: 1}), (b:A {v: 2})
RETURN size(shortestPath((a)-[r:E* WHERE 1=0]->(b))) AS c;
```

Both must find no path. Actual: both return `c = 3`, the full unfiltered path through the `tag: 'bad'` relationship.

### Expected behavior

Both evaluators apply the relationship inline property map and the inline `WHERE` to every relationship on the path, rejecting paths that traverse a non-matching relationship.

### Actual behavior

Per the matrix above; three of the four combinations silently ignore the constraint.

### Impact

Silent - no error or warning. A `shortestPath` that reads as constrained returns a path through relationships the query excluded, so routing/reachability answers are wrong in a way that looks correct. The expression form is worse, since even the property map, which callers may reasonably assume works after #5096, is dropped there.

robfrank:
Closes #5481

## Summary

`shortestPath()` / `allShortestPaths()` are evaluated by two independent code paths: `ShortestPathStep` for the `MATCH p = shortestPath(...)` form, and `ShortestPathExpression` for the expression form (`RETURN shortestPath(...)`). Both read only `getTypes()` and `getDirection()` off the relationship pattern, so three of the four constraint/form combinations silently dropped the constraint: the step ignored the inline `WHERE` (it only switched onto its edge-aware BFS when a property map was present, and never evaluated a predicate), and the expression form ignored both the property map and the inline `WHERE` because it always delegated to `SQLFunctionShortestPath`, a vertex-only BFS that cannot see edge properties at all. #5096 fixed only the property map on the `MATCH` form.

The parse side already worked: `CypherASTBuilder.visitRelationshipPattern` and (since #5460) `CypherExpressionBuilder.parseRelationshipPattern` both populate `whereExpression` on the AST node - the predicate reached the evaluators and was ignored. This PR introduces `ShortestPathStep.EdgeConstraint`, a small holder for the inline property map plus the inline `WHERE` predicate with a `matches(Edge)` method, built once per source/target pair. It returns `null` when the pattern declares neither constraint, so the unconstrained hot path keeps using the CSR-accelerated `SQLFunctionShortestPath` and stays allocation-free. `ShortestPathStep` now routes onto its existing edge-aware BFS whenever a constraint is present and checks the full constraint on each hop; `ShortestPathExpression` builds the same constraint and delegates to the shared BFS helpers. Both the optimizer and the legacy planner paths construct the same `ShortestPathStep`, so both are covered by the one change.

`allShortestPaths()` in expression position still yields the single shortest path found, exactly as before this PR - only constraint enforcement was added there, cardinality is unchanged.

## Test plan

New regression test `engine/src/test/java/com/arcadedb/query/opencypher/CypherShortestPathInlineWhereTest.java` (17 cases, 12 of which fail on `main`) over a graph with two equal-length corridors distinguished only by an edge `tag`:

- [x] `MATCH` form, unsatisfiable inline `WHERE` (`WHERE 1 = 0`) returns no path - the exact repro from the issue
- [x] `MATCH` form, inline `WHERE` makes an only-`bad`-reachable target unreachable
- [x] `MATCH` form, inline `WHERE` selects the matching corridor and the alternate corridor
- [x] `MATCH` form, compound predicate (`AND` / `IS NOT NULL`) and `$parameter` predicate
- [x] `MATCH` form, contradictory property map + inline `WHERE` together yield no path
- [x] `MATCH` form, `allShortestPaths` applies the inline `WHERE`
- [x] Expression form, property map rejects an unreachable target (was broken even after #5096)
- [x] Expression form, inline `WHERE` rejects an unreachable target and an unsatisfiable predicate
- [x] Expression form, property map and inline `WHERE` select the right corridor, incl. `$parameter`
- [x] Expression form, `allShortestPaths` applies the inline `WHERE`
- [x] Controls: unconstrained `MATCH` form still returns both co-shortest corridors; unconstrained expression form unchanged

Regression run of the whole Cypher + shortestPath surface in the `engine` module: `mvn test -Dtest='*ShortestPath*,*Cypher*,*OpenCypher*'` - 6609 tests, 0 failures. The only 3 errors are `OpenCypherCustomFunctionTest` GraalVM `polyglot.Engine$ImplHolder` `NoClassDefFound`, verified pre-existing on the base commit with the change stashed.

## Scope note

Sibling work is fixing relationship/node inline `WHERE` in other contexts (#5480). This change is confined to the two `shortestPath` evaluators and does not touch `MatchRelationshipStep` or the pattern-comprehension path, so it should not conflict.

### DIFF
--- a/docs/5481-shortestpath-relationship-inline-where.md
+++ b/docs/5481-shortestpath-relationship-inline-where.md
@@ -52,6 +52,30 @@ covers the full 2x2 matrix (both evaluators x property map / inline `WHERE`), pl
 `allShortestPaths`, the combination of both constraints, `$parameter` predicates, and unconstrained
 control cases.
 
+## Pull request
+
+https://github.com/ArcadeData/arcadedb/pull/5487
+
+## Review cycles
+
+### Cycle 1 - head `2fcad506`
+
+`claude[bot]` reviewed (no blocking findings). `gemini-code-assist` did not respond inside the
+15-minute window.
+
+| Finding | Outcome |
+|---|---|
+| 1. A relationship property map supplied as a bare parameter (`-[:LINK* $props]->`) is still dropped | Partly applied. Verified: on the `MATCH` form there is no gap - `CypherASTBuilder.validateNoParameterProperties` rejects a parameter map in a `MATCH` pattern at parse time (`InvalidParameterUse`), so it can never be silently dropped. The expression form is not covered by that validation, and there `$props` really was dropped; `EdgeConstraint.from` now resolves it from the query parameters. Both behaviors are pinned by tests. |
+| 2. `allShortestPaths()` in expression position returns one path while the `MATCH` form returns all co-shortest paths | No change - intentional. The expression form has always returned a single path; this PR only adds constraint enforcement and deliberately leaves cardinality untouched. Changing it belongs in its own issue. |
+| 3. `matchesProperties` duplicates the numeric-coercion rules of `GraphTraverser.matchesPropertyFilter` | Applied. Extracted `GraphTraverser.matchesPropertyFilter(Edge, Map)` as a static helper; the existing protected instance method delegates to it, and `EdgeConstraint` calls it. One definition now serves the variable-length MATCH traversers and both `shortestPath` evaluators. |
+
+## Test results
+
+`mvn test -Dtest='*ShortestPath*,*Cypher*,*OpenCypher*,*Traverse*,*Traversal*'` in `engine`:
+6646 tests, 0 failures. The 3 errors are `OpenCypherCustomFunctionTest` GraalVM
+`polyglot.Engine$ImplHolder` `NoClassDefFound`, verified pre-existing on the base commit with the
+change stashed.
+
 ## Scope note
 
 Sibling work is fixing relationship inline `WHERE` in other contexts (node inline `WHERE`, #5480).
--- a/docs/5481-shortestpath-relationship-inline-where.md
+++ b/docs/5481-shortestpath-relationship-inline-where.md
@@ -82,10 +82,33 @@ two-bot approval.
 | 3. `EdgeConstraint` reuses a mutable evaluation row | No change needed - confirmed safe, the contract is documented on the class. |
 | 4. Expression-form `allShortestPaths` cardinality | No change - agreed it belongs in a separate issue. |
 
+### Cycle 3 - head `3ea5867e`
+
+`claude[bot]` re-reviewed, no blocking findings. `gemini-code-assist` did not respond in this window
+either (three consecutive windows), so the loop terminates on a reviewer timeout.
+
+| Finding | Outcome |
+|---|---|
+| 1. The `*1..3` in the tests is decorative since bounds are unenforced; add a note so a future reader is not misled | Applied. Note added to the first `*1..3` test. |
+| 2. Property map (`longValue()` coercion) and inline `WHERE` (exact numerics) can disagree on fractional values now that they sit side by side | No change - pre-existing, and changing it would move variable-length MATCH semantics. Listed as a follow-up candidate below. |
+| 3. `computeFilteredShortestPath` / `computeFilteredAllShortestPaths` are now `public static`, widening the step class surface | No change. Both callers are shortestPath evaluators and they live in different packages; the alternative is duplicating the BFS, which is the defect this PR removes. |
+| 4. `EdgeConstraint` mutable-row reuse | No change needed - confirmed safe and documented. |
+
+CI on the final head: `build-and-package`, all five e2e suites, CodeQL and `claude-review` pass.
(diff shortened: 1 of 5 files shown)


robfrank:
Addressed in 315b3533.

**1. `$props` relationship property map.** Verified, and the picture splits by form:

- **`MATCH` form**: no gap. `CypherASTBuilder.validateNoParameterProperties` throws `InvalidParameterUse: Parameters cannot be used as predicates in MATCH patterns` for a relationship parameter map, so `shortestPath((a)-[:LINK* $props]->(b))` inside `MATCH` never reaches the evaluator - it is rejected at parse time, not silently dropped. Pinned by `matchFormRejectsParameterisedPropertyMap`.
- **Expression form**: real gap, since `CypherExpressionBuilder` is not covered by that validation. `EdgeConstraint.from` now resolves `getPropertiesParameterName()` against `context.getInputParameters()`, mirroring `PatternPredicateExpression.matchesRelationshipProperties`. Pinned by `expressionFormAppliesParameterisedPropertyMap`.

**2. `allShortestPaths()` cardinality asymmetry in expression position.** Intentionally deferred, confirmed. The expression form has always returned a single path; this PR adds constraint enforcement without touching cardinality. Changing it is a behavior change that belongs in its own issue.

**3. Duplicated numeric coercion.** Applied. Extracted `GraphTraverser.matchesPropertyFilter(Edge, Map)` as a static helper; the existing protected instance method delegates to it and `EdgeConstraint` calls it, so the variable-length MATCH traversers and both `shortestPath` evaluators share one definition. The `longValue()`-only comparison is carried over verbatim, unchanged.

Regression run after the change: `mvn test -Dtest='*ShortestPath*,*Cypher*,*OpenCypher*,*Traverse*,*Traversal*'` in `engine` - 6646 tests, 0 failures. The 3 errors are the pre-existing `OpenCypherCustomFunctionTest` GraalVM `polyglot.Engine$ImplHolder` `NoClassDefFound`, reproduced on the base commit with the change stashed.

PR #5500, opened 2026-07-28
feat(#4861): add the MCP hybrid_search tool

### LINKED ISSUE #4861: MCP: add hybrid_search tool

robfrank:
Closes #4861. Part of epic #4859.

Adds the `hybrid_search` MCP tool, fusing a vector leg, an optional full-text leg, and an optional depth-capped graph-expansion leg into one ranked result list.

## Design

Graph expansion is a **third, rank-only fusion source**, not post-fusion context. Seeds are the interleaved union of the vector and full-text hits; a breadth-first `TRAVERSE` fans out from them; expanded neighbours then compete for the final top-K on their own, each carrying `depth` and the RID `path` back to its seed.

The expansion leg has no natural score, only a BFS position, so it enters fusion with no score field at all. That makes `RRF` mandatory whenever `expand` is present — `DBSF` and `LINEAR` require a numeric score on every row of every source — and the tool rejects that combination up front rather than letting it surface as a parse-level complaint about a source the caller never wrote. Both strategies remain available for vector + full-text fusion.

Java orchestrates the legs and owns every cap, ordering rule, and error message; the engine's `vector.fuse` performs all fusion scoring, so there is no second RRF implementation in the tree.

Also extracts `MCPVectorLeg` from `VectorSearchTool` so both retrieval tools validate vector arguments identically. That refactor is behaviour-preserving and is covered by the pre-existing `vectorSearch*` tests.

## Engine behaviours this works around

Three were established by direct experiment before the design was fixed. Each is counter-intuitive and each fails silently:

- **`vector.fuse` reads `score` as a similarity and sign-flips only `distance`.** A dense vector leg's distance emitted under `score` leaves `RRF` correct (rank-only) while *inverting* the ranking under `DBSF`/`LINEAR`. Each leg now emits under the key matching its direction.
- **`vector.fuse` silently drops rows whose `@rid` is a string.** RIDs are bound as `RID` objects.
- **`TRAVERSE ... FROM :ridCollection` throws `NullPointerException`.** `TraverseExecutionPlanner:155-174` and `SelectExecutionPlanner:1435-1453` omit the `rid.setLegacy(true)` their own singleton-RID branches perform. Seed RIDs are inlined as literals instead — safe here because they originate from the engine's own retrieval legs, never from caller text. This is a separate engine defect and is **not** filed yet; a drafted issue body is included at `docs/superpowers/4861-engine-traverse-npe-issue.md`.
- **`out(:boundEdgeList)` silently matches nothing.** Edge type names are inlined as validated, quoted literals, and an unknown name is rejected listing the available ones rather than producing an empty neighbourhood.

## Notes for reviewers

- `MCPPermissionsTest` previously asserted `isToolAllowed(ALL, "hybrid_search")` is `false` — it was that test's example of a name declared in a profile but never registered. Registering the tool flips it, so the assertion moves to positive `ALL`/`RAG` checks plus a false `ADMIN` check; the adjacent `"unknown"` case still covers unregistered names.
- `weights` is keyed by leg name rather than positionally, and the expansion leg defaults to `0.5`: under RRF a rank-1 expanded node otherwise scores exactly what the single nearest neighbour scores, so an arbitrary one-hop neighbour would tie the best semantic match.
- `weights: {"expand": -1}` is now rejected even when no `expand` block is supplied. Deliberate, tested, and user-visible.
- A vector-only call cannot be fused (`vector.fuse` needs two sources), so it reports `"fused": false` and returns the leg's native `distance`/`score` rather than a fabricated `fusedScore`.

## Testing

- `mvn -pl server test -Dtest='MCP*,HybridSearchCapsTest,HybridSearchSeedsTest'` — **219 passing, 0 failures** (baseline 190).
- `mvn -pl server test` — 739 tests, 2 errors. Both are `PostCommandHandlerDecodeTest` failing with `NoClassDefFoundError: org.graalvm.polyglot.Engine$ImplHolder`, reproduced identically at this branch's base commit, so they are a pre-existing local GraalVM environment issue and not a regression.
- `mvn -pl engine test -Dtest='*VectorFuse*,*FullText*,SQLTraverse*'` — 220 passing, 0 failures.

Tests covering the expansion leg narrow the vector leg with a `filter` on purpose: the fixture holds fewer records than the retrieval leg's over-fetch limit, so without it every record becomes a seed, seeds are excluded from the expansion leg by design, and the assertions would hold vacuously. Each such test was verified to **fail** with the expansion leg stubbed to return nothing.

## Follow-ups not in this PR

- The engine `setLegacy` defect above (drafted, unfiled).
- Documentation, which lives in the separate `ArcadeData/arcadedb-docs` repo at `src/main/asciidoc/reference/mcp/mcp.adoc`. That page is already behind `main` (`sample_records`, `vector_search`, and the schema Resource are undocumented); this PR does not repair that pre-existing drift.

### DIFF
--- a/docs/superpowers/4861-engine-traverse-npe-issue.md
+++ b/docs/superpowers/4861-engine-traverse-npe-issue.md
@@ -1,0 +1,92 @@
+# Draft issue: TRAVERSE and SELECT from a bound RID collection throw NullPointerException
+
+File with:
+
+```bash
+gh issue create --repo ArcadeData/arcadedb --label bug \
+  --title "TRAVERSE and SELECT from a bound RID collection throw NullPointerException" \
+  --body-file docs/superpowers/4861-engine-traverse-npe-issue.md
+```
+
+(Delete this header block first, or pass the body inline - everything below the line is the issue body.)
+
+---
+
+## Problem
+
+Passing a collection of RIDs as a query parameter target throws `NullPointerException`. Both of these
+fail:
+
+```sql
+SELECT @rid AS rid FROM :seeds
+SELECT @rid AS rid FROM (TRAVERSE out('LINK') FROM :seeds MAXDEPTH 2)
+```
+
+with `params = {"seeds": List<RID>}`:
+
+```
+java.lang.NullPointerException: Cannot invoke
+"com.arcadedb.query.sql.parser.Expression.execute(com.arcadedb.query.sql.executor.Result,
+ com.arcadedb.query.sql.executor.CommandContext)" because "this.expression" is null
+  at com.arcadedb.query.sql.parser.Rid.toRecordId(Rid.java:71)
+  at com.arcadedb.query.sql.executor.TraverseExecutionPlanner.handleRidsAsTarget(TraverseExecutionPlanner.java:220)
+  at com.arcadedb.query.sql.executor.TraverseExecutionPlanner.handleInputParamAsTarget(TraverseExecutionPlanner.java:174)
+```
+
+A **single** bound RID works. Only a collection fails, which is what makes this easy to miss.
+
+## Cause
+
+Both planners build a `Rid` AST node per element and set `bucket` and `position`, but never call
+`rid.setLegacy(true)`:
+
+- `engine/src/main/java/com/arcadedb/query/sql/executor/TraverseExecutionPlanner.java:155-174`
+- `engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:1435-1453`
+
+Their own **singleton**-RID branches do call it, a few lines above in the same method:
+
+- `TraverseExecutionPlanner.java:150`
+- `SelectExecutionPlanner.java:1425`
+
+`Rid.toRecordId` (`engine/src/main/java/com/arcadedb/query/sql/parser/Rid.java:67-71`) reads
+`bucket`/`position` when `legacy` is true, and otherwise dereferences `expression`, which the
+collection branch never populates:
+
+```java
+  public RID toRecordId(final Result target, final CommandContext context) {
+    if (legacy) {
+      return context.getDatabase().newRID(bucket.value.intValue(), position.value.longValue());
+    } else {
+      final Object result = expression.execute(target, context);   // expression == null here
+```
+
+Note that `Rid`'s two constructors differ: `Rid(RID)` sets `legacy = true`, while `Rid(int)` - the one
+both collection branches use - leaves it false.
+
+## Fix
+
+Add `rid.setLegacy(true)` to both collection branches, matching the singleton branches beside them.
+
+## Regression test
+
+One test per planner, each binding a **collection** - a single-element bound RID passes today and
+would not catch the regression:
+
+```java
+final Map<String, Object> params = Map.of("seeds", List.of(v0.getIdentity(), v1.getIdentity()));
+database.query("sql", "SELECT FROM :seeds", params);
+database.query("sql", "SELECT FROM (TRAVERSE out('LINK') FROM :seeds MAXDEPTH 1)", params);
+```
+
+## Impact and workaround
+
+Any caller computing a set of RIDs and feeding it to a traversal or select must inline the RIDs as
+literals (`FROM [#1:0,#1:1]`) instead of binding them. That form works and returns full BFS metadata
+including `$depth` and `$path`.
+
+## Context
+
+Found while designing the MCP `hybrid_search` tool (#4861), whose graph expansion leg traverses from a
+computed seed set. That tool works around it by inlining RID literals - safe there because the seed
+RIDs originate from the engine's own retrieval legs and never from caller text - and does not depend
+on this fix landing.
--- a/docs/superpowers/plans/2026-07-28-mcp-hybrid-search.md
+++ b/docs/superpowers/plans/2026-07-28-mcp-hybrid-search.md
@@ -1,0 +1,2082 @@
+# MCP `hybrid_search` Tool Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add an MCP `hybrid_search` tool that fuses a vector leg, a full-text leg, and a depth-capped graph-expansion leg into a single ranked result list.
+
+**Architecture:** Java orchestrates each retrieval leg independently, owning all caps, ordering, and error messages, then hands the materialized rank lists to the engine's `vector.fuse` SQL function as bound parameters so fusion scoring keeps exactly one implementation. The graph leg is a rank-only third source produced by a breadth-first `TRAVERSE` seeded from the union of the other two legs.
+
+**Tech Stack:** Java 21, Maven, JUnit 5 + AssertJ, ArcadeDB SQL (`vector.fuse`, `TRAVERSE`), ArcadeDB MCP server (`server` module).
+
+**Spec:** `docs/superpowers/specs/2026-07-28-mcp-hybrid-search-design.md`
+**Issue:** [#4861](https://github.com/ArcadeData/arcadedb/issues/4861), epic [#4859](https://github.com/ArcadeData/arcadedb/issues/4859)
+
+## Global Constraints
+
+- Java 21+. Maven multi-module; all work is in the `server` module except Task 7, which files an issue and writes no code.
+- Use `com.arcadedb.serializer.json.JSONObject` / `JSONArray`, never another JSON library.
+- Import classes; never use fully-qualified names inline.
+- `final` on variables and parameters wherever possible.
+- Single-statement `if` bodies take no braces.
+- Tests use `assertThat(x).isTrue()` style (AssertJ), JUnit 5.
+- Comments state behavioral invariants only - no issue numbers, no fix context, no "Claude" attribution anywhere.
+- Do not commit to git unless a step explicitly says to; the plan's commit steps are expected and fine.
+- After every Java change: compile and fix until it passes.
+- Prefer primitive arrays over boxed collections on hot paths.
+- Run at minimum `mvn -pl server test -Dtest='MCP*'` before declaring any task done.
+- MCP documentation lives in the separate `ArcadeData/arcadedb-docs` repository, not this tree.
+
+---
+
+## File Structure
+
+| File | Responsibility |
+|---|---|
+| `server/src/main/java/com/arcadedb/server/mcp/tools/MCPVectorLeg.java` | **new.** Vector-index resolution, query-vector validation, sparse-query construction, filter normalization, and generation of the vector-leg SQL. Shared by `vector_search` and `hybrid_search`. |
+| `server/src/main/java/com/arcadedb/server/mcp/tools/HybridSearchTool.java` | **new.** Tool definition, leg orchestration, seed selection, expansion, fusion, response shaping. |
+| `server/src/main/java/com/arcadedb/server/mcp/tools/VectorSearchTool.java` | **modify.** Delegates its private validation statics to `MCPVectorLeg`. No behavior change. |
+| `server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java` | **modify.** Register in `TOOLS_LIST`, dispatch in `toolsCall`, summarize in `formatResult`. |
+| `server/src/test/java/com/arcadedb/server/mcp/MCPServerPluginTest.java` | **modify.** Hybrid fixture and the end-to-end HTTP tests. |
+| `server/src/test/java/com/arcadedb/server/mcp/MCPPermissionsTest.java` | **modify.** Profile assertions; repair the assertion that only passes while the tool is unregistered. |
+| `server/src/test/java/com/arcadedb/server/mcp/MCPStdioServerTest.java` | **modify.** Stdio transport registration and `rag` profile visibility. |
+| `server/src/test/java/com/arcadedb/server/mcp/HybridSearchCapsTest.java` | **new.** Pure unit test for cap arithmetic, no running server. |
+
+---
+
+### Task 1: Extract the shared vector leg into `MCPVectorLeg`
+
+Pure refactor. `vector_search`'s observable behavior must not change; the existing `vectorSearch*` tests are the proof.
+
+**Files:**
+- Create: `server/src/main/java/com/arcadedb/server/mcp/tools/MCPVectorLeg.java`
+- Modify: `server/src/main/java/com/arcadedb/server/mcp/tools/VectorSearchTool.java`
+- Test: `server/src/test/java/com/arcadedb/server/mcp/MCPServerPluginTest.java` (existing tests, unchanged)
+
+**Interfaces:**
+- Consumes: nothing from earlier tasks.
+- Produces:
+  - `MCPVectorLeg.DEFAULT_K` = 10, `MAX_K` = 1000, `FILTER_OVERFETCH` = 8, `MAX_FILTER_CANDIDATES` = 8000, `MAX_FILTER_EXPRESSION` = 4096 (all `public static final int`)
+  - `record MCPVectorLeg.ResolvedVectorIndex(TypeIndex typeIndex, int dimensions, String scoring)`
+  - `record MCPVectorLeg.VectorLegQuery(ResolvedVectorIndex index, boolean sparse, int candidateLimit, String sql, Map<String, Object> parameters)`
+  - `static VectorLegQuery MCPVectorLeg.build(Database database, JSONObject args, String indexNameField, int limit)`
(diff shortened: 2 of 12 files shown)


robfrank:
Verified all four observations against the code. Two are real and are fixed in 7fa1be696; two I'm pushing back on. Both questions have definitive answers in the engine source.

### Fixed

**1. `fulltextIndexName` omitted from the unfused response.** Confirmed — it was set only inside the fused branch, so a full-text leg that ran but matched nothing produced a response that named no index. Hoisted out of the branch: the field is now reported whenever the leg ran, so a caller cannot mistake an empty leg for an absent one. Pinned by `hybridSearchReportsTheFullTextIndexEvenWhenThatLegMatchesNothing`, which I verified fails against the old code with `JSONObject[fulltextIndexName] not found`.

**4. Truncation probe undercount.** Confirmed, and I took it further than "cosmetic". The probe was `rows.size() >= MAX_EXPANSION && resultSet.hasNext()`, but rows skipped as duplicates or seeds still consume the stream, so the surplus row could be eaten before the check. Dropped the probe: `truncated` is now `rows.size() >= MAX_EXPANSION`. That also makes it consistent with the top-level `truncated` and with `vector_search`, both of which already report a filled window rather than asking whether more exist.

### Pushing back

**2. Engine error messages forwarded verbatim.** I don't think this is worth changing. The same authenticated principal reaches the `query` tool, which executes arbitrary caller-supplied read SQL and returns engine errors directly — so the structure of a generated statement isn't privileged information relative to what that principal can already do. More importantly, this tool is consumed by a model, and the engine's message is the main signal it has to self-correct; replacing it with a generic stage label would trade a real diagnostic for no security gain. `vector_search` already behaves this way, so changing only `hybrid_search` would also split the convention.

**3. `filter` concatenated into the WHERE clause.** Agreed on the substance — `isIdempotent()` blocks writes, not expense, so a crafted predicate can still reference other types or run costly subqueries. But this moved unchanged from `VectorSearchTool` in the `MCPVectorLeg` extraction, and the identical exposure exists in `vector_search` today and, far more broadly, in the `query` tool. Fixing it here would leave the other two paths untouched while implying the problem was scoped to this PR. Happy to file it as its own issue covering the read-cost budget across all three surfaces if you think it's worth tracking.

### Answers

**Does RRF rank purely by row order?** Yes, and it never reads the score key. `SQLFunctionVectorFuse.applyRRF` iterates `for (int rank = 0; rank < rows.size(); rank++)` and computes `w / (k + (rank + 1L))` from the list index alone; `RidScore.score()` is not referenced in that method at all. So the dense leg's ordering is what RRF consumes, and the `distance` key matters only to `DBSF`/`LINEAR`, whose `applyLinear`/`applyDBSF` do read the value. That split is exactly why emitting a distance under `score` was invisible under RRF and inverted the ranking under the other two — the defect this PR's fix history covers.

**Is `weights: 0.0` intended?** Yes, deliberately allowed. Zero is a meaningful request: it keeps a leg's rows in the candidate set and in each result's `sources` provenance while contributing nothing to the ranking, which is a legitimate way to ask "what would this leg have found, without letting it move the order?". A negative weight has no such reading — it actively inverts a leg — which is why that one is rejected. I'll note the distinction in the tool description if you'd rather it be explicit to callers.

MCP suite is 220 passing after the change.

robfrank:
### Fixed

**1. Contradictory comments in `runExpansionLeg`.** Correct, and the contradiction was mine — I introduced it in 7fa1be696 when I dropped the `resultSet.hasNext()` probe in response to the previous round without revisiting the budget that fed it. Verified: the LIMIT comment still justified a surplus row that line 444 no longer reads.

Rather than reconcile the wording, I removed the `+1` itself in 58bace7c3. It was dead weight once truncation stopped probing: with `LIMIT seeds.size() + MAX_EXPANSION`, the outer depth filter discards exactly the depth-0 seeds and the remainder is capped at `MAX_EXPANSION`, so `rows.size() >= MAX_EXPANSION` is reachable precisely when the traversal filled its budget. The comment now states only the invariant that still holds — the budget must cover the seeds the filter discards.

### Noted, no change

**2. `filter` inlined into the vector-leg SQL.** Same position as the previous round: this moved unchanged from `VectorSearchTool`, and `vector_search` plus the `query` tool carry the identical exposure today. Its safety resting on `isIdempotent()` rather than parameterization is a real property worth stating, and your comment states it accurately — but scoping the fix to `hybrid_search` would leave the other two paths untouched while implying the problem was introduced here. Still happy to file it as a standalone issue covering read-cost limits across all three surfaces.

**3. `truncated` under-reporting when rows are skipped.** Agreed, and deliberate. A skipped row is a duplicate, a seed, or a stale record — never a neighbor the caller lost — so a short count reflects the caller getting everything the traversal could give them, not a hidden gap. The flag answers "did the window fill", which is the same contract `vector_search` publishes, and both are documented as such. Widening it to "might more exist" would pin it to true on almost every real graph and strip it of meaning.

MCP suite 220 passing.

robfrank:
Verified all four against the code. Two fixed in eed6efc05, one deferred to the repo owner, one I'm pushing back on.

### Fixed

**1. Silent seed truncation.** Confirmed — `collectSeeds` capped at `MAX_SEEDS` with no signal, while the fan-out cap next to it was reported. `legs.expand` now carries `seedCount` and `seedsTruncated`, so a thin neighbourhood no longer reads the same whether the graph was sparse or the retrieval legs simply out-ran the budget. That asymmetry mattered more than usual here, since making silent degradation loud is the principle the rest of this tool is built on.

**2. `validateWeights` error quality.** Confirmed. `weights.getDouble(legName, 0.0)` reaches Gson's `getAsNumber().doubleValue()`, so `weights: {"vector": "abc"}` surfaced a raw `NumberFormatException` naming neither the argument nor the rule. Now type-checked alongside the finite and non-negative rules. Pinned by `hybridSearchRejectsANonNumericWeight`, verified to fail against the old code.

While writing the seed test I found one of my own: the end-to-end assertion for `seedsTruncated` passed with the flag hard-coded to `false`, because a 6-record fixture cannot drive a 256-seed cap. Rather than leave a test that cannot fail, I dropped that assertion, and strengthened `HybridSearchSeedsTest.seedListNeverExceedsTheCap` — which asserted only `isLessThanOrEqualTo(MAX_SEEDS)`, satisfied equally by an empty list — to pin the exact cap. All three seed tests now fail with `collectSeeds` stubbed empty.

### Deferred to the repo owner

**3. Filing the engine defect before merge.** Agreed on the substance — the workaround should have a tracked path to removal. The decision to draft rather than file was made deliberately by the repo owner, who asked that outward-facing actions on this branch stay with them, so I'm not filing it unilaterally. The body is ready at `docs/superpowers/4861-engine-traverse-npe-issue.md` with the `gh issue create` command; it needs one command from a maintainer.

### Pushing back, partly

**4. Committed workflow artifacts.** For the plan and spec, this is the established convention rather than an open question: `origin/main` already carries **37 files under `docs/superpowers/plans/` and 36 under `docs/superpowers/specs/`**, going back to April. Adding two more follows it; omitting them would be the deviation. An earlier review round reached the same conclusion independently.

Where you're right is a file you didn't name. `docs/superpowers/4861-engine-traverse-npe-issue.md` is a drafted GitHub issue body, not a design record — and it would be the **first** file ever committed directly under `docs/superpowers/` outside `plans/` and `specs/`, which is exactly the "process artifact in the shipped tree" smell you're pointing at. It also becomes redundant the moment the issue is filed. I've flagged it to the repo owner to drop before merge; I'd rather not delete the only copy while the issue is still unfiled.

MCP suite 222 passing.

robfrank:
Verified all five. One fixed in 8d9af31c5, one pushed back with engine evidence, two deferred to the repo owner, one a deliberate choice.

### Fixed

**1. Weights for legs that won't run.** Correct, and it turns this PR's own argument against it — I rejected unknown weight keys precisely because a silently-ignored override is worse than a loud error, then left `weights: {"fulltext": 2.0}` with no full-text leg doing exactly that. Now rejected, naming the missing leg and how to supply it.

The full-text leg counts as requested when *either* of its arguments is present, so an incomplete leg still gets the more specific error naming both fields rather than a confusing weights error. Two tests: `hybridSearchRejectsAWeightForALegTheRequestDoesNotUse` proves the guard fires (verified failing with it disabled), and `hybridSearchAcceptsAWeightForALegTheRequestDoesUse` proves it does not over-fire on a legitimate three-leg override.

### Pushing back

**3. Expansion `LIMIT` vs `truncated` accuracy.** The premise doesn't hold: the raw traversal stream contains no duplicate emissions to inflate it. `AbstractTraverseStep` keeps a `RidHashSet traversed` (line 47) and `BreadthFirstTraverseStep` consults it before every emission (lines 68, 142, 169), so each RID is emitted exactly once, at its shallowest depth. I confirmed this behaviourally too — the fixture reaches h2 both directly from h0 and via h1, and it appears once at depth 1, which `hybridSearchDedupsNodesReachableBySeveralPaths` asserts.

So the stream is exactly `seeds + unique neighbours`, the budget `seeds.size() + MAX_EXPANSION` lines up with it, and `rows.size()` reaching `MAX_EXPANSION` coincides with the traversal filling its budget. The only residual gap is a row my loop skips as malformed, which is the documented under-report convention this tool shares with `vector_search`, not a dense-graph effect.

### Noted, deliberate

**2. `validateArguments` runs twice on the hybrid path.** Intentional, and you've identified the reason yourself: the call inside `build` keeps `build` correct for any direct caller, while the explicit early call is what lets both tools report argument faults before resolving the database. The checks are pure and allocation-free. Collapsing either one would trade a real invariant for a micro-optimisation on the rejection path.

### For the repo owner

**4. File the engine `setLegacy` issue.** Agreed on substance, and this is the second round to raise it. The decision to draft rather than file was the repo owner's — they asked that outward-facing actions stay with them — so I'm not filing unilaterally. The body and the `gh issue create` command are ready in the repo; it needs one command from a maintainer.

**5. Docs in the PR.** For the plan and spec this is the established convention, not an open question: `origin/main` already carries 37 files under `docs/superpowers/plans/` and 36 under `docs/superpowers/specs/`. You're right about a file you didn't name, though — `docs/superpowers/4861-engine-traverse-npe-issue.md` is a drafted issue body, would be the first file ever committed under `docs/superpowers/` outside those two directories, and becomes redundant once filed. Flagged to the owner to drop before merge; I'd rather not delete the only copy while the issue is unfiled.

MCP suite 224 passing.

robfrank:
Verified all five. Two addressed in 20d944505; three are unchanged positions with the reasoning restated.

### Addressed

**1. Trust boundary at the SQL append sites.** Correct that the safety was load-bearing on `validatedEdgeTypes` with nothing at the inlining site saying so — the rationale lived in the method Javadoc and the PR description, so an edit to the validator would see no marker. Both sites now carry it: the edge-type loop names `validatedEdgeTypes` as the trust boundary and states that weakening it reopens injection, and the seed loop records that RIDs are engine-derived and that `RID.toString` renders only `#<bucket>:<position>`. Switching to bound parameters remains the right end state once the engine defect is fixed.

**4. Weight for a requested-but-empty leg.** Distinct from the absent-leg case fixed in 8d9af31c5, and real: the guard checks argument presence, which cannot know the leg will return zero rows before the database is even resolved. Validating it earlier isn't possible, so I've stated it in the tool description instead — a weight for a leg that runs but matches nothing has no effect, and that leg's zero row count is already reported under `legs`, so the caller can see why.

### Unchanged positions

**2. Filing the engine defect.** Third round to raise it, and the substance isn't in dispute. The decision to draft rather than file belongs to the repo owner, who asked that outward-facing actions on this branch stay with them; I've escalated it to them again rather than act unilaterally. Body and command are ready in the repo.

**3. Planning docs in the main tree.** `origin/main` already carries 37 files under `docs/superpowers/plans/` and 36 under `docs/superpowers/specs/`, so these two follow the convention rather than establishing one. `CLAUDE.md`'s note about `arcadedb-docs` concerns user-facing documentation, which is a separate artifact and still outstanding for this tool. The genuine outlier is `docs/superpowers/4861-engine-traverse-npe-issue.md` — the first file ever committed under `docs/superpowers/` outside those two directories, and redundant once the issue is filed. Flagged to the owner to drop before merge.

**5. Inherited filter inlining.** Unchanged from `VectorSearchTool`; `vector_search` and the `query` tool carry the identical surface today. Scoping a fix to `hybrid_search` would leave the other two untouched while implying the exposure originated here. Still happy to file it as a standalone issue covering read-cost limits across all three surfaces.

### Test run note

`MCPAuthorizationBindingIT` currently errors locally with `JSONObject[result] not found`. That is environmental, not a regression: a separately-installed ArcadeDB server (Homebrew `arcadedb/26.7.2`) is holding port 2481, so the test reaches that server instead of its own. I confirmed the identical failure on the previously-green commit 8d9af31c5 with the code reverted, which isolates it from this change. The remaining MCP suite is 223 passing, 0 failures.

robfrank:
Both outstanding items from the last three rounds are now closed.

**Engine defect filed as #5505** — `TRAVERSE` and `SELECT` from a bound RID collection throw `NullPointerException` because `TraverseExecutionPlanner:155-174` and `SelectExecutionPlanner:1435-1453` omit the `rid.setLegacy(true)` their own singleton-RID branches perform. The issue carries the repro, both call sites, the suggested one-line fix, and a regression test sketch noting that a single bound RID passes today, so the test must bind a collection.

**The workaround now links it at the site.** The seed-inlining comment in `runExpansionLeg` points at #5505 and states that the RIDs can be bound once it lands, so whoever revisits that code sees both why it is inlined and when it can stop being. The spec records the number too.

**Drafted issue body removed.** It was the only file ever committed under `docs/superpowers/` outside `plans/` and `specs/`, and redundant the moment the issue existed.

Worth noting on the code comment: this repo's `CLAUDE.md` asks that source comments state behavioral invariants only, without issue references. The link is here because the repo owner asked for it explicitly, overriding that rule for this case.

MCP suite 223 passing. `MCPAuthorizationBindingIT` still errors locally against a separately-installed ArcadeDB server holding port 2481, unrelated to this branch and reproduced on an earlier commit.

Issue #5492, opened 2026-07-28
HA: a materialized view makes the Raft leader ship page versions followers never received - WALVersionGapException, non-converging snapshot-resync loop, lost committed writes


robfrank:
## Summary

On a 3-node Raft HA cluster, the presence of a materialized view on a database causes the leader to ship a `TX_ENTRY` whose page version is ahead of what followers hold. Followers throw `WALVersionGapException`, trigger a full snapshot resync, and **the resync does not repair it** - the next entry gaps again on a different file. The cluster enters a resync loop, every request routed to a resyncing follower is deflected with a 503, and the run ends with committed writes permanently missing from one node.

This is lost data, not split-brain: the affected node's id set is a strict *subset* of the others.

> **Scope note (2026-07-29).** This issue originally bundled two defects. The second one - materialized view refresh requests being *dropped* rather than queued when a refresh was already running - was confirmed, fixed and merged in #5502 (`6642e203e`). That half is done and has been removed from this description. What remains, and what this issue now tracks exclusively, is the replication defect below.

## Evidence: controlled A/B

Same load, same cluster, same client. The **only** difference is whether the schema contains a materialized view.

| Arm | Result |
|---|---|
| With `CREATE MATERIALIZED VIEW UserStats ... REFRESH INCREMENTAL` | 2041/3000 user writes fail, 12 snapshot-resync cycles in ~9 s, nodes freeze at 1120/1122/1122 users and never converge |
| Without the view | 3000/3000 users, 30000/30000 photos, **zero** errors, all three nodes converge immediately |

Reproduced twice locally against `arcadedata/arcadedb:latest`. The view is the trigger, not the load and not the client.

## Causal chain

1. The leader ships a `TX_ENTRY` whose page version is 2 ahead of what followers hold. The first occurrence is **identical and simultaneous on both followers**: `PageId(/142/0) version in WAL (1648) does not match with existent version (1646)`. Identical on both followers points at a leader-side replication-completeness defect, not a follower race. File 142 is the materialized view's own backing bucket (`UserStats_1.142.65536.v0.bucket`, confirmed by creating the same schema standalone and reading the file names).
2. The follower throws `WALVersionGapException` -> `ReplicationException: WAL version gap detected - snapshot resync required` and starts a full snapshot resync.
3. The resync does not fix it. The very next entry gaps on a different file (`User_0`, fileId 1, WAL 304 vs local 302). This loops 12 times.
[...]

PR #5502, opened 2026-07-28
fix(#5492): coalesce materialized view refresh requests instead of dropping them

robfrank:
Refs #5492. **Deliberately does not close it** - see Scope.

## Scope

Issue #5492 bundles two defects. This PR fixes the second one only.

| Defect | Status |
|---|---|
| A refresh arriving while another is in flight is dropped, not queued, so the view can be silently stale | **Fixed here** |
| The leader ships page versions followers never received (`WALVersionGapException`, snapshot-resync loop, lost writes) | **Not fixed - not reproduced** |

The issue records the replication root cause as "not yet proven". It still is. A 3-node in-process Raft harness (`BaseRaftHATest`, whose consistency check already routes through `DatabaseComparator` and fails on any cross-node page-version mismatch) produced **zero** `WALVersionGapException` and zero resync loops across every configuration tried - writes on the leader, writes through a follower, with and without a UNIQUE index on the source type, 180 and 450 records. Shipping a speculative change to the Raft commit path without a reproduction would risk making a data-loss bug worse, so that half stays open.

## The defect fixed here

`MaterializedViewRefresher.fullRefresh` began with `if (!view.tryBeginRefresh()) return;`.

The refresh is triggered from a post-commit callback, so the dropped request belongs to a transaction that has **already committed**. The in-flight refresh it deferred to started before that commit and therefore reads a snapshot without it. Nothing reschedules, so the view stays missing that record until some later write happens to win the race - or forever.

No HA is needed to hit this: a 3-node run showed the **leader's own** view holding 449 of 450 committed records.

It also explains part of the throughput collapse in the issue's A/B table - every committing transaction schedules a full `TRUNCATE` + re-run of the defining query, so under single-record transactions the work degenerates towards a full rebuild per row.

## Fix

Requests are coalesced onto the running refresh. `MaterializedViewImpl` replaces the `AtomicBoolean` with a three-state `AtomicInteger` (`IDLE` / `RUNNING` / `RUNNING_PENDING`) and adds:

- `markRefreshPendingIfRunning()` - hands a request to the running refresh; returns `false` if none is running, so the caller runs it itself.
- `finishRefreshPassAndCheckPending()` - ends one pass, returning `true` if a request arrived during it (another pass required), else releasing ownership.

Releasing ownership and testing for a pending request is **one atomic step**. Split in two, a request registered in between would be seen by neither the outgoing owner nor the requester (which observed the refresh as still running) - the same lost-wakeup the original code had unconditionally.

`tryBeginRefresh()` and `endRefresh()` keep their existing contracts, so the existing guard tests are untouched.

## Deliberately not done

True incremental view maintenance. `MaterializedViewChangeListener` still schedules a full refresh regardless of `REFRESH INCREMENTAL`, as its comment says. Per-record maintenance is a feature with real correctness risk (wrong view contents rather than slow ones), not a bug fix. Coalescing already removes the redundant rebuilds, which is where the cost was.

## Also found, not addressed here

- `RaftReplicatedDatabase.recordFileChanges` drains buffered WAL into `walEntries` but omits it from the replication guard, and the `finally` then clears the buffer. Its sibling `runWithCompactionReplication` guards correctly. Left alone for lack of a reproduction from the view path.
- Concurrent single-record transactions issued **through a follower** lose 60-70% of records and log `Invalid record size ... deleting record`. The A/B control with no materialized view fails identically, so it is unrelated to views. Being filed separately.

## Test plan

- [x] `refresherRegistersItsRequestWhenAnotherRefreshIsInFlight` is the regression guard: it drives the exact interleaving with no timing dependence, and was **verified to fail** against the old drop-behaviour and pass with the fix.
- [x] Three state-machine transition tests (hand-off, coalescing of repeated requests, caller-runs when nothing is in flight).
- [x] `concurrentWritersLeaveTheViewReflectingEveryCommit` end-to-end convergence under 4 writers. Note this one passes against the old code too (the race needs load to surface), so it guards against deadlock or a non-terminating drain loop rather than the original defect.
- [x] `engine` `MaterializedView*` - 61 pass
- [x] `engine` `com.arcadedb.schema.*Test` - 310 pass
- [x] `server` `HTTPMaterializedViewIT`, `RemoteMaterializedViewIT`, `Issue3941AsyncRefreshMaterializedViewIT` - 11 pass
- [x] `ha-raft` `RaftReplicationMaterializedViewIT` - 1 pass

### DIFF
--- a/docs/5492-materialized-view-refresh-coalescing.md
+++ b/docs/5492-materialized-view-refresh-coalescing.md
@@ -1,0 +1,117 @@
+# Issue #5492 - a dropped materialized view refresh leaves the view permanently stale
+
+- Issue: https://github.com/ArcadeData/arcadedb/issues/5492
+- Branch: `fix/5492-ha-mv-page-version-replication-gap`
+- Base: `main` at f7ee51d01
+
+## Scope
+
+Issue #5492 bundles two defects. This branch fixes the second one only, and deliberately does **not**
+close the issue.
+
+| Defect | Status here |
+|---|---|
+| A refresh arriving while another is in flight is dropped, not queued, so the view can be silently stale | **Fixed** |
+| The leader ships page versions followers never received (`WALVersionGapException`, snapshot-resync loop, lost writes) | **Not fixed - not reproduced** |
+
+The issue itself records the replication root cause as "not yet proven". It still is. Shipping a
+speculative change to the Raft commit path without a reproduction would risk making a data-loss bug
+worse, so that half stays open.
+
+## What was investigated for the replication half
+
+A 3-node in-process Raft harness (`BaseRaftHATest`, whose `checkDatabasesAreIdentical()` already
+routes through `DatabaseComparator` and fails on any cross-node page-version mismatch) was built and
+run against an `INCREMENTAL` view under concurrent single-record transactions. Findings:
+
+- **No page-version gap reproduced.** Zero `WALVersionGapException`, zero resync loops, across every
+  configuration tried (writes on the leader, writes through a follower, with and without a UNIQUE
+  index on the source type, 180 and 450 records).
+- Two theories from the issue were checked against current code. `LocalSchema` is constructed with
+  `wrappedDatabaseInstance` (`LocalDatabase.java:2423`), so the refresh does go through
+  `RaftReplicatedDatabase` - the issue is right to rule that theory out.
+- One genuine leader-side asymmetry was found but **not** shown to be reachable from the view path:
+  `RaftReplicatedDatabase.recordFileChanges` drains buffered WAL into `walEntries`, but its
+  replication guard omits `walEntries` from the condition, and the `finally` then clears the buffer.
+  Its sibling `runWithCompactionReplication` guards correctly. Left alone here for lack of a
+  reproduction; worth a look on its own.
+- An **unrelated** failure surfaced: concurrent single-record transactions issued through a follower
+  lose 60-70% of records and log `Invalid record size ... deleting record` on the affected buckets.
+  The A/B control with no view fails identically, so it is not attributable to materialized views.
+  Filed separately.
+
+## The defect that is fixed
+
+`MaterializedViewRefresher.fullRefresh` began with:
+
+```java
+if (!view.tryBeginRefresh())
+  return;   // another refresh is running
+```
+
+The refresh is triggered from a post-commit callback, so the request that gets dropped belongs to a
+transaction that has **already committed**. The in-flight refresh it deferred to started before that
+commit and therefore reads a snapshot without it. Nothing reschedules. The view stays missing that
+record until some later write happens to win the race - or forever, if none does.
+
+This needs no HA to reproduce: a 3-node run showed the **leader's own** view holding 449 of 450
+committed records.
+
+It also explains part of the throughput collapse in the issue's A/B table. Every committing
+transaction that touches the source type schedules a full `TRUNCATE` + re-run of the defining query,
+so under single-record transactions the work degenerates towards a full rebuild per row.
+
+## Fix
+
+Requests are coalesced onto the running refresh instead of being dropped.
+
+`MaterializedViewImpl` replaces the `AtomicBoolean refreshInProgress` with a three-state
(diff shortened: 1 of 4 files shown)


robfrank:
Thanks - items 2 and 3 are fixed in 1b70f233b. Item 1 is real but needs a decision I can't make unilaterally; details below.

### 2. Error path could still drop a pending request - fixed

Confirmed, and it was the sharper version of the bug: `endRefresh()` did a plain `set(REFRESH_IDLE)`, so a request registered during a failing pass was clobbered, not just unserviced. Added `releaseRefreshAfterFailure()`, which CASes the release and returns whether a request was discarded.

The request is deliberately **not** retried - a pass that just failed would most likely fail again, and retrying a persistent failure would spin. Instead the discard is logged at WARNING and the view is left in a non-VALID status (`ERROR` here, or `STALE` via `MaterializedViewChangeListener`'s catch), so the staleness is visible rather than silent. The misleading comment about the success path being the atomic one has been corrected to say what the error path actually does.

New test `aFailedPassReportsTheRequestItDiscardsInsteadOfClobberingIt`, verified to fail against the plain-write release before trusting it.

### 3. `@Tag("slow")` - fixed

Applied at method level on `concurrentWritersLeaveTheViewReflectingEveryCommit`; the state-machine tests stay untagged.

### 1. Stray MCP hybrid_search docs - real, but not from this branch's work

Verified, and the diagnosis is more specific than "stray agent planning docs". Those two files come from **two commits that are on the author's local `main` but were never pushed**:

```
$ git log --oneline origin/main..HEAD
1b70f233b address review: ...
e23c1947c fix(#5492): coalesce materialized view refresh requests ...
f7ee51d01 docs(#4861): implementation plan for the MCP hybrid_search tool
a1959e343 docs(#4861): design spec for the MCP hybrid_search tool
```

The branch was cut from local `main`, so it inherited them and the diff against `origin/main` shows them. They are not something this work added, and dropping them from the branch would mean rebasing onto `origin/main` and force-pushing, which the workflow I'm running forbids on a pushed branch.

The clean resolution needs no history rewrite at all: **pushing those two commits to `main` makes this PR's diff collapse to just the fix.** Flagging for the repo owner rather than acting on it.

### Minor notes

- `docs/5492-...md` is kept: `docs/NNNN-*.md` is the established convention in this repo for per-issue investigation notes, and this one records the negative result on the replication half (what was tried, what did not reproduce), which the PR description only summarizes.
- Agreed on the unlucky-writer point - inherent to full-refresh-per-change, strictly better than before, and genuinely fixed only by true incremental maintenance, which stays out of scope here.

robfrank:
Thanks - and the snapshot-ordering walk-through (pass K clears the flag, pass K+1 re-reads after it was set) is exactly the invariant the fix rests on.

**Taken:** the discard WARNING named `ERROR`, but `MaterializedViewChangeListener` overwrites the status with `STALE` right after, so it reported a state the operator never sees. Reworded to "left non-VALID and stale" with a comment explaining why it deliberately does not name a status.

**Verified and tracked, not fixed here:**

- **`ContinuousAggregate` has the identical defect** - good catch. Confirmed it is driven the same way: `ContinuousAggregateRefresher.incrementalRefresh` still uses the `AtomicBoolean` drop-on-contention guard and is invoked from an after-commit callback at `SaveElementStep.java:154-156`, the same shape as the view listener. The three-state primitives port over mechanically, but it needs its own tests, so it stays out of this PR.
- **Synchronous drain on the committing thread** - agreed this is the thing to track before heavy write workloads. Recorded in `docs/5492-*.md` with the `DatabaseAsyncExecutor` / bounded-pool suggestion.

Both are noted in the tracking doc; filing them is left to the repo owner rather than opening more issues unprompted.

**Declined (optional item):** a test asserting end-to-end re-convergence after a discarded request. Forcing a refresh to fail deterministically needs either a droppable source type (blocked - dropping a view's source is rejected) or a view built outside the schema with a deliberately invalid query, which would exercise the fixture more than the behaviour. State-machine-level recovery (ownership released, next caller takes it) is already covered by `aFailedPassReportsTheRequestItDiscardsInsteadOfClobberingIt`.

Issue #5494, opened 2026-07-28
abs(Long.MIN_VALUE) silently returns a negative value instead of throwing an overflow error


shulei5831sl:
### ArcadeDB version
ArcadeDB 26.8.1-SNAPSHOT (build 1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main)

### Environment
- Linux x86_64, Docker, OpenJDK
- arcadedata/arcadedb:latest
- HTTP command API, Cypher language

### Description
abs(-9223372036854775808) silently returns -9223372036854775808 instead of throwing an error. Long.MIN_VALUE has no positive counterpart in signed 64-bit integers. Neo4j correctly throws an ArithmeticException.

### Steps to reproduce
```cypher
RETURN abs(-9223372036854775808) AS v;
```

### Expected behavior
Should throw an overflow error, as Neo4j does.

### Actual behavior
Returns -9223372036854775808 (same negative value — silent overflow).

### Cross-engine comparison
| Query | Neo4j 2026.06 | ArcadeDB 26.8.1 |
|-------|:---:|:---:|
| abs(Long.MIN_VALUE) | ArithmeticException | -9223372036854775808 |

### Isolated recheck
Fresh Docker container, 3 runs: reproduces 3/3 (100%).

### Why this looks like a bug
abs() should return the absolute (non-negative) value of its argument. Long.MIN_VALUE has no absolute value in 64-bit signed integers. Returning the negative input violates the mathematical definition of absolute value.

### Impact
Silent incorrect results — computations applying abs() to Long.MIN_VALUE receive a negative result without any error.

lvca:
Thanks for the detailed report! Good news: this is already fixed in `main`.

`abs()` now fails with a `long overflow` error instead of returning a negative result, both in Cypher and in SQL, matching Neo4j:

```cypher
RETURN abs(-9223372036854775808) AS v    -->  long overflow
```

The fix landed in #5504 (thanks @waterWang!), and later work made the error report as an HTTP 400 client error rather than a 500. A regression test covering your exact query ships with it and is green.

Worth noting the SQL side got a bit more than reported: `abs()` in SQL preserves the input type, so `INTEGER`, `SHORT` and `BYTE` each have the same problem at their own minimum value. All four are now guarded.

One related loose end is tracked separately in #5545: in SQL the overflow still returns HTTP 500 instead of 400. PR #5631 is open for it.

Closing this one as resolved - please reopen if you still see it on a recent build.

[...]

PR #5504, opened 2026-07-28
fix: throw ArithmeticException on abs(Long.MIN_VALUE) overflow (#5494)

waterWang:
When `abs(Long.MIN_VALUE)` is called, `Math.abs(Long.MIN_VALUE)` silently returns `Long.MIN_VALUE` because `-Long.MIN_VALUE` overflows in signed 64-bit arithmetic. This affects both the Cypher `abs()` function and the SQL `abs()` function.

**Fix**: Add an explicit check for `Long.MIN_VALUE` before calling `Math.abs()`, and throw `ArithmeticException` when it occurs, matching Neo4j's behaviour.

**Files changed**:
- `engine/src/main/java/com/arcadedb/function/math/AbsFunction.java` — Cypher abs() function
- `engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionAbsoluteValue.java` — SQL abs() function

**Related**: https://github.com/ArcadeData/arcadedb/issues/5494

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/math/AbsFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/math/AbsFunction.java
@@ -40,8 +40,12 @@ public class AbsFunction implements StatelessFunction {
       throw new CommandExecutionException("abs() requires exactly one argument");
     if (args[0] == null)
       return null;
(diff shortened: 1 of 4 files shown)


robfrank:
Thank you for this contribution @waterWang

PR #5506, opened 2026-07-28
fix: add rid.setLegacy(true) in collection branches to prevent NullPointerException (#5505)

### LINKED ISSUE #5505: TRAVERSE and SELECT from a bound RID collection throw NullPointerException
## Problem

Passing a collection of RIDs as a query parameter target throws `NullPointerException`. Both of these
fail:

```sql
SELECT @rid AS rid FROM :seeds
SELECT @rid AS rid FROM (TRAVERSE out('LINK') FROM :seeds MAXDEPTH 2)
```

with `params = {"seeds": List<RID>}`:

```
java.lang.NullPointerException: Cannot invoke
"com.arcadedb.query.sql.parser.Expression.execute(com.arcadedb.query.sql.executor.Result,
 com.arcadedb.query.sql.executor.CommandContext)" because "this.expression" is null
  at com.arcadedb.query.sql.parser.Rid.toRecordId(Rid.java:71)
  at com.arcadedb.query.sql.executor.TraverseExecutionPlanner.handleRidsAsTarget(TraverseExecutionPlanner.java:220)
  at com.arcadedb.query.sql.executor.TraverseExecutionPlanner.handleInputParamAsTarget(TraverseExecutionPlanner.java:174)
```

A **single** bound RID works. Only a collection fails, which is what makes this easy to miss.

## Cause

Both planners build a `Rid` AST node per element and set `bucket` and `position`, but never call
`rid.setLegacy(true)`:

- `engine/src/main/java/com/arcadedb/query/sql/executor/TraverseExecutionPlanner.java:155-174`
- `engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:1435-1453`

Their own **singleton**-RID branches do call it, a few lines above in the same method:

- `TraverseExecutionPlanner.java:150`
- `SelectExecutionPlanner.java:1425`

`Rid.toRecordId` (`engine/src/main/java/com/arcadedb/query/sql/parser/Rid.java:67-71`) reads
`bucket`/`position` when `legacy` is true, and otherwise dereferences `expression`, which the
collection branch never populates:

```java
  public RID toRecordId(final Result target, final CommandContext context) {
    if (legacy) {
      return context.getDatabase().newRID(bucket.value.intValue(), position.value.longValue());
    } else {
      final Object result = expression.execute(target, context);   // expression == null here
```

Note that `Rid`'s two constructors differ: `Rid(RID)` sets `legacy = true`, while `Rid(int)` - the one
both collection branches use - leaves it false.

## Fix

Add `rid.setLegacy(true)` to both collection branches, matching the singleton branches beside them.

## Regression test

One test per planner, each binding a **collection** - a single-element bound RID passes today and
would not catch the regression:

```java
final Map<String, Object> params = Map.of("seeds", List.of(v0.getIdentity(), v1.getIdentity()));
database.query("sql", "SELECT FROM :seeds", params);
database.query("sql", "SELECT FROM (TRAVERSE out('LINK') FROM :seeds MAXDEPTH 1)", params);
```

## Impact and workaround

Any caller computing a set of RIDs and feeding it to a traversal or select must inline the RIDs as
literals (`FROM [#1:0,#1:1]`) instead of binding them. That form works and returns full BFS metadata
including `$depth` and `$path`.

## Context

Found while designing the MCP `hybrid_search` tool (#4861), whose graph expansion leg traverses from a
computed seed set. That tool works around it by inlining RID literals - safe there because the seed
RIDs originate from the engine's own retrieval legs and never from caller text - and does not depend
on this fix landing.

waterWang:
### Fixes #5505

**Problem:** Passing a collection of RIDs as a query parameter to SELECT or TRAVERSE throws NullPointerException:

**Root cause:** The collection branches in both  and  create  objects without calling . The singleton RID branches beside them do call it. Without this flag,  tries to dereference a null  field instead of reading / directly.

**Fix:** Added  in the collection branch of both planners:
-  (after )
-  (after )

**Testing:** Single-element bound RID collections and multi-element collections now work correctly for both SELECT and TRAVERSE.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -1445,6 +1445,7 @@ public class SelectExecutionPlanner {
         bucket.setValue(orid.getBucketId());
         final PInteger position = new PInteger(-1);
         position.setValue(orid.getPosition());
+        rid.setLegacy(true);
         rid.setBucket(bucket);
         rid.setPosition(position);
         if (filterClusters == null || isFromClusters(rid, filterClusters, context.getDatabase())) {
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/TraverseExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/TraverseExecutionPlanner.java
@@ -166,6 +166,7 @@ public class TraverseExecutionPlanner {
         bucket.setValue(orid.getBucketId());
         final PInteger position = new PInteger(-1);
         position.setValue(orid.getPosition());
+        rid.setLegacy(true);
         rid.setBucket(bucket);
         rid.setPosition(position);
 
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/Issue5505BoundRidCollectionTargetTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/Issue5505BoundRidCollectionTargetTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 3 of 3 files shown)


justinblethrow-cloud:
Confirmed this independently on current `main` (`cda99c1c`) and added focused regression coverage for both planner paths. Before `rid.setLegacy(true)`, both tests fail with the reported `Rid.toRecordId` null dereference; with the two changes in this PR, both pass.

Tests-only commit: https://github.com/justinblethrow-cloud/arcadedb/commit/59b631f82fc38bd9116408d71eb53f2dee1c87a4

It can be applied with:

```bash
git cherry-pick 59b631f82fc38bd9116408d71eb53f2dee1c87a4
```

The tests cover:
- `SELECT @rid AS rid FROM :seeds` with two bound record identifiers.
- `SELECT FROM (TRAVERSE ... FROM :seeds MAXDEPTH 1)` with two roots and verified expanded results.

Focused verification: 2 tests run, 0 failures, 0 errors. This should also exercise the two lines that Codecov currently reports as uncovered.

PR #5509, opened 2026-07-28
fix(#5508): mask previousValue for hidden settings in set_server_setting

### LINKED ISSUE #5508: set_server_setting echoes previousValue in clear for hidden settings
## Problem

`set_server_setting` returns the previous value of a setting without redacting secrets, while its sibling `get_server_settings` masks them.

`GetServerSettingsTool` masks any setting where `GlobalConfiguration.isHidden()` is true — which covers `clusterToken`, `*Password*`, and `*password*` — and its inline comment ties that hardening to GHSA-p9wc-4fhr-78wm and cluster-forwarded-auth root impersonation:

```java
final boolean hidden = cfg.isHidden();
setting.put("value", hidden ? "*****" : normalize(cfg.getValue()));
setting.put("default", hidden ? "*****" : normalize(cfg.getDefValue()));
```

`SetServerSettingTool` applies no equivalent check:

```java
final Object oldValue = server.getConfiguration().getValue(cfg);
server.getConfiguration().setValue(key, value);
...
result.put("previousValue", oldValue != null ? oldValue.toString() : JSONObject.NULL);
```

So calling `set_server_setting` with `key: "arcadedb.ha.clusterToken"` (or any `*password*` key) returns the prior value in clear text in `previousValue` — the same disclosure the getter was hardened against, reached through the setter instead.

## Reachability

Not remotely exploitable in a default deployment. It requires:

- the MCP server enabled (`enabled` defaults to false), and
- `allowAdmin` enabled (defaults to false), and
- a principal already authorized for admin operations.

That makes this a hardening gap rather than a remotely exploitable vulnerability, which is why it is filed as a normal issue. It still matters: an admin-scoped MCP principal is not necessarily a principal that should learn the cluster token, and the redaction on the getter exists precisely because that token enables cluster-forwarded-auth root impersonation.

## Fix

Reuse the check the getter already applies, in `SetServerSettingTool.execute`:

```java
result.put("previousValue", cfg.isHidden() ? "*****" : String.valueOf(oldValue));
```

`cfg` is already resolved a few lines above by `GlobalConfiguration.findByKey(key)`, so no extra lookup is needed. Note the `oldValue != null` branch currently yields `JSONObject.NULL`; preserve whichever null representation is chosen, but a hidden key should mask regardless of whether it currently holds a value.

## Regression test

Assert that setting a hidden key never echoes its prior value:

```java
final JSONObject payload = /* call set_server_setting with key arcadedb.ha.clusterToken */;
assertThat(payload.getString("previousValue")).isEqualTo("*****");
```

`GetServerSettingsToolRedactionIT` already covers the getter side and is the natural place to add the setter case, so both paths are pinned by the same test class.

## Context

Found while documenting the MCP operator tools for the ArcadeData/arcadedb-docs manual. The `set_server_setting` section is being written to describe consistent masking across both tools, which this change makes true.

robfrank:
Closes #5508.

`set_server_setting` returns the value it replaced, without redaction. `get_server_settings` masks any setting flagged by `GlobalConfiguration.isHidden()` — `clusterToken`, `*Password*`, `*password*` — so a caller could read back through the setter what the getter refuses to disclose.

Before the fix, the response was literally:

```json
{"key":"arcadedb.ha.clusterToken","previousValue":"s3cr3t-cluster-token-should-never-leak","newValue":"replacement-token"}
```

`previousValue` now applies the same `isHidden()` rule as the getter. The mask is keyed on the setting rather than on whether it currently holds a value, so an unset secret cannot be distinguished from a set one by probing.

## Reachability

Not remotely exploitable in a default deployment: it needs the MCP server enabled (`enabled` defaults to false), `allowAdmin` enabled (defaults to false), and a principal already authorized for admin operations. That is why it was filed as a normal issue rather than through private disclosure. It still matters — an admin-scoped MCP principal is not necessarily one that should learn the cluster token, and the getter's redaction exists precisely because that token enables cluster-forwarded-auth root impersonation.

## Testing

The regression test is added to `GetServerSettingsToolRedactionIT`, beside the getter's, so both disclosure paths are pinned by one class. Verified to fail before the change with the secret visible in `previousValue`, and to pass after.

Note that `mvn -pl server verify` runs the unit tests before failsafe, and two of those fail locally on a pre-existing GraalVM `NoClassDefFoundError` unrelated to this change, which prevents the ITs from starting. To run the IT in isolation:

```
mvn -pl server verify -Dtest=ZzzNoSuchUnitTest -DfailIfNoTests=false \
  -Dsurefire.failIfNoSpecifiedTests=false \
  -DskipITs=false -Dit.test=GetServerSettingsToolRedactionIT
```

`mvn -pl server test -Dtest='MCP*'` is 217 passing.

### DIFF
--- a/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
@@ -19,6 +19,7 @@
 package com.arcadedb.server.mcp;
 
 import com.arcadedb.Constants;
+import com.arcadedb.GlobalConfiguration;
 import com.arcadedb.database.DatabaseContext;
 import com.arcadedb.log.LogManager;
 import com.arcadedb.serializer.json.JSONArray;
@@ -271,7 +272,7 @@ public class MCPDispatcher {
     final JSONObject args = params.getJSONObject("arguments", new JSONObject());
 
     LogManager.instance()
-        .log(this, Level.INFO, "MCP[%s] tools/call '%s' %s (user=%s)", transport, toolName, formatArgs(args), user.getName());
+        .log(this, Level.INFO, "MCP[%s] tools/call '%s' %s (user=%s)", transport, toolName, formatArgs(toolName, args), user.getName());
 
     try {
       if (!REGISTERED_TOOL_NAMES.contains(toolName))
@@ -382,15 +383,27 @@ public class MCPDispatcher {
     }
   }
 
-  private static String formatArgs(final JSONObject args) {
+  /**
+   * Renders a tool's arguments for the request log. The value a caller writes to a secret setting is masked here,
+   * because this line is emitted before the tool runs and would otherwise place the secret in the log regardless of
+   * how the tool's own response is redacted. Masking is deliberately limited to the one argument that is a secret by
+   * definition: arguments elsewhere carry caller data, which cannot be told apart from a secret without guessing, and
+   * blanking it would leave the log unable to explain what a call did.
+   */
+  static String formatArgs(final String toolName, final JSONObject args) {
     if (args.length() == 0)
       return "{}";
+    final boolean maskValue = "set_server_setting".equals(toolName) && isHiddenSetting(args.getString("key", null));
     final StringBuilder sb = new StringBuilder("{");
     boolean first = true;
     for (final String key : args.keySet()) {
       if (!first)
         sb.append(", ");
       first = false;
+      if (maskValue && "value".equals(key)) {
+        sb.append(key).append("=\"*****\"");
+        continue;
+      }
       final Object value = args.get(key);
       if (value instanceof String s) {
         final String sanitized = sanitizeForLog(s);
@@ -404,6 +417,18 @@ public class MCPDispatcher {
         sb.append(key).append("=").append(value);
     }
     return sb.append("}").toString();
+  }
+
+  /**
+   * Reports whether a configuration key names a setting the server treats as secret, using the same
+   * {@link GlobalConfiguration#isHidden()} rule the settings tools apply. An unresolvable key is not secret: the
+   * tool rejects it before it can change anything.
+   */
+  private static boolean isHiddenSetting(final String key) {
+    if (key == null || key.isEmpty())
+      return false;
+    final GlobalConfiguration cfg = GlobalConfiguration.findByKey(key);
+    return cfg != null && cfg.isHidden();
   }
 
   private static String sanitizeForLog(final String value) {
--- a/server/src/main/java/com/arcadedb/server/mcp/tools/SetServerSettingTool.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/tools/SetServerSettingTool.java
@@ -71,7 +71,12 @@ public class SetServerSettingTool {
 
     final JSONObject result = new JSONObject();
     result.put("key", key);
-    result.put("previousValue", oldValue != null ? oldValue.toString() : JSONObject.NULL);
+    // A secret is masked on the way out here for the same reason it is masked when read: this response
+    // hands the caller the value it just replaced, so echoing it would disclose through the setter what
+    // the getter refuses to disclose. Masking is keyed on the setting, not on whether it held a value,
+    // so an unset secret cannot be distinguished from a set one either.
+    result.put("previousValue",
+        cfg.isHidden() ? "*****" : oldValue != null ? oldValue.toString() : JSONObject.NULL);
     result.put("newValue", value);
     result.put("message", "Setting '" + key + "' updated successfully.");
     return result;
--- a/server/src/test/java/com/arcadedb/server/mcp/GetServerSettingsToolRedactionIT.java
+++ b/server/src/test/java/com/arcadedb/server/mcp/GetServerSettingsToolRedactionIT.java
@@ -24,6 +24,7 @@ import com.arcadedb.serializer.json.JSONArray;
 import com.arcadedb.serializer.json.JSONObject;
 import com.arcadedb.server.BaseGraphServerTest;
 import com.arcadedb.server.mcp.tools.GetServerSettingsTool;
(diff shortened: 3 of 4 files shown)


robfrank:
Item 2 fixed in 551c66215. Item 1 I'm pushing back on, because the occurrence count it rests on is lower than reality. Item 3 needs no action.

### Fixed

**2. Test mutates the live HA cluster token.** Real, and worth more than a comment: the test drives the real setter, so it genuinely replaces the running server's token, and that token authenticates cluster-forwarded requests. Restored in a `finally` block — from the constant rather than from the response, since the response now masks it.

A following assertion pins the restore:

```java
assertThat(getServer(serverIndex).getConfiguration().getValueAsString(GlobalConfiguration.HA_CLUSTER_TOKEN))
    .as("the original cluster token must be restored after the setter test")
    .isEqualTo(SECRET_TOKEN);
```

Verified it discriminates: with the restore removed it fails on exactly that assertion, so deleting the cleanup breaks the test here instead of leaking a replaced token into whatever runs next.

### Pushing back

**1. Shared mask constant.** The literal is not in three places, it is in **six** production sites:

```
server/.../mcp/tools/SetServerSettingTool.java      1
server/.../mcp/tools/GetServerSettingsTool.java     2   (value, default)
server/.../http/handler/GetServerHandler.java       3
```

`GetServerHandler` is a different subsystem — the HTTP server endpoint, hardened under GHSA-46hj-24h4-j8gf, the sibling advisory to the one this fix descends from. An `MCPToolUtils.MASKED` constant would cover the MCP tools and leave the HTTP handler on its own literal, which is worse than the status quo: a reader who found `MCPToolUtils.MASKED` would reasonably assume changing it changes the mask everywhere, and it would not. Partial centralization of a security-visible constant is more misleading than none.

If the mask should live in one place, the honest home is beside `GlobalConfiguration.isHidden()` in the engine — that is already the single source of truth for *which* settings are secret, so it is the natural home for *how* they are shown, and it is reachable from both subsystems. That is an engine-touching change across two subsystems and does not belong in a security fix whose diff should stay easy to audit. Happy to do it as a follow-up if you want it.

One related note: the tests should keep asserting the literal `"*****"` rather than any constant. A test written as `isEqualTo(MCPToolUtils.MASKED)` passes even if the constant is changed to an empty string, so it would stop pinning the wire format it exists to protect.

### Noted

**3. Coverage.** Matches what I found: `get_server_settings` and `set_server_setting` are the only two config-disclosure paths on the MCP surface, and `GetServerHandler` covers the HTTP one separately.

`MCP*` is 217 passing; the redaction IT is 2 passing.

robfrank:
Checked the `newValue` point against the code paths it depends on. I'm leaving `newValue` unmasked, but the concern behind it is real and lands somewhere else — one line earlier, on the request side.

### Why masking `newValue` would not achieve the stated goal

The rationale was that if MCP tool responses are logged or persisted, the new secret lands in that sink in clear. Two things make that not hold:

**The response is not logged.** `MCPDispatcher` logs a summary, not the response object, and for this tool the summary is the key alone (`MCPDispatcher.java:429`):

```java
case "set_server_setting" -> result.getString("key", "") + " updated";
```

**The arguments are logged, verbatim, at INFO** (`MCPDispatcher.java:274`):

```java
.log(this, Level.INFO, "MCP[%s] tools/call '%s' %s (user=%s)", transport, toolName, formatArgs(args), user.getName());
```

and `formatArgs` appends every string argument as-is, truncated at 100 characters. The only transformation is `sanitizeForLog`, which escapes `\n`, `\r`, and `\t` — a log-injection guard, not a secret guard. So a call setting the cluster token already produces:

```
MCP[http] tools/call 'set_server_setting' {key="arcadedb.ha.clusterToken", value="<new-secret>"} (user=root)
```

The secret is in the log before the tool body runs. Masking `newValue` in the response would leave that untouched, so it would buy the appearance of protection rather than protection — and, as you note, it is the caller's own value, so it discloses nothing to them.

That is my reason for leaving it: not that the risk is acceptable, but that this particular change does not reduce it.

### The finding underneath it

`formatArgs` logging secret argument values at INFO is a genuine credential-in-logs exposure, and it is broader than `set_server_setting` — any tool argument carrying a secret reaches the same sink. It deserves its own issue and fix rather than being folded into this PR, whose diff should stay narrow enough to audit as a security change. I'm raising it separately and will link it here.

Worth saying plainly: your comment is what surfaced it. The `newValue` observation was pointed at the right risk and the wrong mechanism.

### Scope check

Matches mine — `get_server_settings` and `set_server_setting` are the only MCP paths handling hidden settings, `GetServerHandler` covers the HTTP surface separately, and the search tools read database-scoped configuration, which `isHidden()` does not flag.

robfrank:
Folded into this PR at the repo owner's direction, so it closes both paths rather than leaving one for a follow-up. `17d8cca02`.

`formatArgs` now masks the `value` argument when the tool is `set_server_setting` and the key resolves to a setting `GlobalConfiguration.isHidden()` flags — the same rule the tools themselves apply.

Masking stays limited to that one argument, deliberately. Everywhere else, arguments carry caller data that cannot be distinguished from a secret without guessing, and blanking them would leave the request log unable to explain what a call did. The setting key is still logged, and an ordinary setting's value is still logged.

Five tests in `MCPDispatcherArgumentLoggingTest`, split so they pin both directions:

- the cluster token and a password setting are masked
- an ordinary setting's value is still logged
- an argument named `value` on a different tool is not masked, so the rule cannot silently blank an unrelated argument that shares the name
- an unresolvable key is not masked, since the tool rejects it before it changes anything

Verified the pair discriminates: with the mask condition forced to `false`, the two masking tests fail and the three negative tests keep passing.

`MCP*` is 222 passing, and the redaction IT is 2 passing.

The `newValue` field itself is still echoed unmasked, for the reason in my previous comment — it is the caller's own value and the response is not logged. With the request-log path now closed, the sink that motivated the original suggestion is gone.

PR #5510, opened 2026-07-28
fix: return HTTP 400 instead of 500 for abs() invalid argument type (fixes #5484)

### LINKED ISSUE #5484: [OpenCypher] abs() returns HTTP 500 for invalid argument instead of client-facing error
## Environment
- Host OS: Linux x86_64
- Deployment: Docker
- Query interface: HTTP command API
- Query language: `opencypher`
- ArcadeDB version: 26.8.1-SNAPSHOT
- Build: 72d70d8f5b7095202dc5b880e6d9cfaa978d6f9c/1785207432240/main
- Docker image: `arcadedata/arcadedb:latest`
- RepoDigest: `arcadedata/arcadedb@sha256:1cb78c08e2e40e765f18cd1b06be83d878d7e13c01471db6bddb112a53d0de74`
- API endpoint: `POST http://localhost:2494/api/v1/command/test`

## Minimal reproducer

```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN abs('"'"'hello'"'"') AS r"}'
```

Response (HTTP 500):
```json
{"error":"Cannot execute command","requestId":"<id>","exception":"com.arcadedb.exception.CommandExecutionException","detail":"abs() requires a numeric argument"}
```

## Matched valid-input control

```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN abs(-1) AS r"}'
```

Response (HTTP 200):
```json
{"user":"root","result":[{"r":1}]}
```

## Expected behavior

An invalid argument type should produce a client-facing error with an
appropriate client-facing HTTP 4xx status, such as HTTP 400. The error message itself
(`abs() requires a numeric argument`) is correct and informative.

## Actual behavior

The correct error message is returned, but with HTTP status 500
(internal server error) instead of a client-level status code. The
`CommandExecutionException` class and 500 status suggest an internal
failure, when the actual behavior is correct argument validation.

## Stability

18/18 fresh container executions returned the same response.
30/30 valid-input controls passed.

## Relationship to other issues

This concerns error-class hygiene (wrong HTTP status for a correct
type-check message), distinct from #5476 and #5477 where the type
check itself is not performed and null is silently returned.

Arguments that are accepted by `abs()` (INTEGER, FLOAT) return correct
numeric results and do not produce this error.

waterWang:
## Summary

Changes `CommandExecutionException` to `CommandParsingException` in `AbsFunction.java` so that invalid argument types (e.g. `abs('hello')`) return HTTP 400 (client error) instead of HTTP 500 (internal server error).

## Problem

`abs()` with a non-numeric argument correctly detects the error and returns the message `abs() requires a numeric argument`, but wraps it in `CommandExecutionException` which triggers HTTP 500. This is incorrect — the error is a client-side validation error, not an internal server failure.

## Fix

Changed both `throw new CommandExecutionException(...)` calls in `AbsFunction.java` to `throw new CommandParsingException(...)`, which produces HTTP 400. The error message is preserved unchanged.

## Related

All other math functions (e.g. `floor()`, `ceil()`, `round()`, `sign()`) already use `CommandParsingException` for argument validation errors. This brings `abs()` in line with the rest of the codebase.

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/math/AbsFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/math/AbsFunction.java
@@ -18,7 +18,7 @@
  */
 package com.arcadedb.function.math;
 
-import com.arcadedb.exception.CommandExecutionException;
+import com.arcadedb.exception.CommandParsingException;
 import com.arcadedb.function.StatelessFunction;
 import com.arcadedb.query.sql.executor.CommandContext;
 
@@ -37,13 +37,13 @@ public class AbsFunction implements StatelessFunction {
   @Override
   public Object execute(final Object[] args, final CommandContext context) {
     if (args.length != 1)
-      throw new CommandExecutionException("abs() requires exactly one argument");
+      throw new CommandParsingException("abs() requires exactly one argument");
     if (args[0] == null)
       return null;
     if (args[0] instanceof Byte || args[0] instanceof Short || args[0] instanceof Integer || args[0] instanceof Long)
       return Math.abs(((Number) args[0]).longValue());
     if (args[0] instanceof Number)
       return Math.abs(((Number) args[0]).doubleValue());
-    throw new CommandExecutionException("abs() requires a numeric argument");
+    throw new CommandParsingException("abs() requires a numeric argument");
   }
 }


robfrank:
Thanks for this, and sorry it sat long enough to be overtaken.

Issue #5484 was fixed in #5593, merged as 6e9d7f446. Your diagnosis was right on both counts - the type check was correct and only its exception class was wrong - so this PR is being closed as superseded rather than rejected.

The merged fix took the same idea wider. `abs()` now routes its argument check through a shared helper that raises `CommandSemanticException`, which extends `CommandParsingException` and so maps to HTTP 400 through the same handler arms you were aiming at:

```java
  @Override
  public Object execute(final Object[] args, final CommandContext context) {
    if (args.length != 1)
      throw CypherFunctionHelper.arityMismatch("abs", "1 argument", args.length);
    // Rejects anything outside INTEGER | FLOAT as a client-facing type error rather than a 500 (issue #5484).
    final Number value = CypherFunctionHelper.requireNumberArgument(args[0], "abs");
    ...
```

Three differences worth flagging, since they are the reason the branch cannot simply be rebased:

1. **Scope.** Every function declared as `f(input :: INTEGER | FLOAT)` now behaves this way - `abs`, `ceil`, `ceiling`, `floor`, `sqrt`, `sign`, `round`, `isNaN`, `exp`, `log`, `ln`, `log10`, the trig and hyperbolic family, `degrees`, `radians`, `haversin`, plus the `math.*` extensions. Fixing `abs()` alone would have left the family inconsistent.
2. **Message wording.** The error is now phrased in the vocabulary of the language, matching `size()` and `head()`: `Type mismatch: abs() expects an INTEGER or a FLOAT argument but got STRING`.
3. **Merge hazard.** This branch predates #5504, so its `abs()` body is still `Math.abs(((Number) args[0]).longValue())`. Merging it would re-introduce the `abs(Long.MIN_VALUE)` overflow that your own #5504 fixed - `Math.abs` returns `Long.MIN_VALUE` unchanged there, a negative "absolute value". Main now uses `Math.absExact` inside a try/catch for exactly that reason.

Coverage on main: `engine/src/test/java/com/arcadedb/query/opencypher/CypherNumericFunctionArgumentIssue5484Test.java` (27 tests, green) and `server/src/test/java/com/arcadedb/server/Issue5484AbsNonNumericHttpStatusIT.java`, which asserts the 400 over the real HTTP endpoint using your reproducer shape.

Thanks again - both the report and the #5504 work were useful, and the reproducer with the matched valid-input control is the reason this was quick to confirm.

robfrank:
Closing as superseded by #5593 (merged 6e9d7f446). See the detail above.

Issue #5388, opened 2026-07-22
Sparse vector search p50 degrades 1.45ms -> 165ms from 100k -> 1M docs on real SPLADE data; latency tracks total posting length (Spearman 0.95)


tae898:
**Summary**

`LSM_SPARSE_VECTOR` query latency falls off a cliff between 100k and 1M documents on real SPLADE vectors (Big-ANN sparse track, SPLADE-cocondenser on MS MARCO): p50 goes 1.45 ms to 165 ms (about 114x for 10x data). A per-query characterization shows latency is almost a pure function of the total posting-list length of the query's terms (Spearman 0.954), which suggests the block-max pruning in `BmwScorer` is not skipping anything meaningful on real term distributions. Recall is unaffected (0.994 to 0.996); this is purely a latency scaling problem. On the synthetic corpora shaped like the engine's own benchmarks the cliff does not exist (p50 stays at 56 ms even at 10M docs), so this only reproduces on real data.

**Setup**

- 26.8.1.dev3 line, embedded via the Python bindings; i9-12900HK, pinned cpuset, 8 GiB heap, N=5 reps
- Dataset: Big-ANN'23 sparse track (public), tiers base_small (100k) and base_1M; 1000 dev queries; exact MIPS ground truth
- Index settled via `COMPACT INDEX` before querying (the no-compact variant is slower still: 240 ms p50 at 1M)

**Observed (p50 per query, median [min-max] over 5 reps)**

| docs | int8 | fp32 | recall@10 (int8) |
|---|---|---|---|
| 100k | 1.45 [1.41-1.57] ms | 1.50 ms | 0.996 |
| 1M | 165 [161-167] ms | 164 ms | 0.994 |

For scale-behavior contrast only: on the same data/machine, inverted-index engines grow 4-11x over the same 10x step (e.g. 0.68 to 2.9 ms), not 114x. int8 vs fp32 being identical rules out decode cost.

**Per-query characterization (1000 dev queries, 1M tier)**

For each query I computed `sum_df` = the summed document frequency of its terms in the corpus, i.e. the total posting entries a scorer would touch with no pruning:

- Spearman(latency, sum_df) = **0.954** (vs 0.926 for nnz alone)
- Median latency by sum_df decile, low to high: 71, 116, 144, 173, 200, 249, 297, 367, 458, 640 ms (monotonic)
[...]

PR #5518, opened 2026-07-29
feat(#4085): split a sparse-vector top-K into parallel RID ranges

### LINKED ISSUE #4085: feat: per-segment parallel top-K scoring for LSM_SPARSE_VECTOR (Step 5 follow-up to #4068)
Follow-up to #4068. Tracks the deferred Step 5 (per-segment parallel scoring) so it does not get lost.

## Context

The v2 sparse-vector backend ships serial BMW DAAT (`BmwScorer.topK`) over a merged per-dim cursor stack. At 10M docs / 30k-dim SPLADE-shape this lands at **1.20 ms/query** with size-tiered compaction in place. That is well inside the noise of network round-trip + JSON serialization for any real-world workload, which is why parallel dispatch was deferred during the v2 land.

What is **already in place** in the tree (no code to write for plumbing, just the hot-path dispatch):

- `SparseVectorScoringPool` - JVM-wide dedicated executor, daemon threads, bounded queue, caller-runs fallback with throttled saturation `WARNING`.
- `arcadedb.sparseVectorScoringPoolThreads` / `arcadedb.sparseVectorScoringQueueSize` global config knobs (currently emit a startup warning that the topK path is still serial).
- `PoolMetrics` MeterBinder with Micrometer gauges for size, active, queue depth, queue capacity, completed tasks, caller-run fallbacks - tagged `pool=sparse_vector`.
- Studio "Executor Pools" card under the Server tab renders the live numbers.

## Scope

1. Wire `PaginatedSparseVectorEngine.topK` to fan out to `SparseVectorScoringPool` when `segments.length >= threshold` and `pool.getMaxParallelism() > 1`.
2. RID-range partitioning that preserves the newest-source-wins merge across cross-segment dim contributions. Each partition produces its own top-K, then a final merge step.
3. Ensure correctness: `BmwScorerCorrectnessTest`-shape coverage at multi-segment scale (parallel result must match serial result bit-for-bit modulo quantization noise).
4. Benchmark against the existing `LSMSparseVectorIndexLargeBenchmark` 10M run; expected gain is ~3-4x on a 4-core machine, capped by the segment count.
5. Drop the "currently the topK path is serial" wording from the two `SPARSE_VECTOR_SCORING_*` config descriptions and from the `SparseVectorScoringPool` startup warning once the dispatch is live.

## Acceptance criteria

- [ ] `BmwScorer.topK` callable through a parallel dispatcher in `PaginatedSparseVectorEngine.topK`, gated by a configurable threshold so single-segment queries stay on the caller thread.
- [ ] RID-range partition correctness proven by a multi-segment correctness test.
- [ ] `LSMSparseVectorIndexLargeBenchmark` shows measurable speedup at 10M (target: under 0.5 ms/query at K=10 on a 4-core box).
- [ ] Studio "Executor Pools" card surfaces non-zero `active` / `completedTasks` / occasional caller-run fallbacks under load.
- [ ] Startup warning + "currently serial" config wording removed.

## Out of scope

- Cross-query parallelism (already linear via independent queries on the engine's thread-safe state).
- Cross-dim parallelism within a single segment (BMW pruning is inherently serial across dims; the parallel axis is segments, not dims).

## Also still tracked

The other deferred item from #4068 - **one-shot backward-compat rebuild for #4065 MVP-era postings on first open** - is not part of this issue. The v2 backend keeps the old LSM-Tree shell readable (so old schemas still open) but does not auto-rebuild MVP postings into segments; pre-v2 datasets need re-insertion. If a user reports it in the field, file a separate issue for that migration path.

lvca:
Closes #4085.

Wires the dispatch `SparseVectorScoringPool` was built for: a top-K over a single `LSM_SPARSE_VECTOR` index is scored as several RID ranges concurrently and merged, instead of one traversal on the caller's thread.

## Measured

SPLADE-shaped corpus, INT8, 18-worker box. Single query, ranges forced:

| corpus | serial | 8 ranges | 16 ranges |
|---|---|---|---|
| 500k | 13.4 ms p50 | 4.0 (3.4x) | 2.4 (5.6x) |
| 1M | 26.2 ms p50 | 7.6 (3.5x) | 4.5 (5.9x) |

`LSMSparseVectorIndexLargeBenchmark`, which now runs both shapes and asserts they return the same documents:

| corpus | serial | adaptive split | |
|---|---|---|---|
| 1M | 5.13 ms/query | 2.83 | 1.82x |
| 10M | 24.18 ms/query | 6.90 | 3.51x |

The 1M gain is smaller there because that benchmark's query is 10 terms over 30 nnz/doc: at 5 ms of work, per-range cursor setup and the merge take a visible share. The split pays in proportion to how much work the query does.

## Why the gating is most of the diff

A range prunes against its own top-K watermark rather than the global one, so it does more work than its share: **1.16x total CPU at 2 ranges, 1.89x at 8**. That is free latency on an idle machine and stolen throughput on a busy one, so the decision has to be adaptive.

Sizing it from pool activity was tried first and measured wrong. A query runs on its caller's thread, so at 16 concurrent clients the pool reads idle at every sampling instant: a third of queries split anyway and throughput fell 14%. The gate that works counts **queries in flight**, not pool activity. At 500k:

| clients | serial QPS | adaptive QPS | p50 | queries split |
|---|---|---|---|---|
| 1 | 72 | 322 | 3.05 vs 13.7 ms | 100% |
| 4 | 290 | 468 (+61%) | 4.13 vs 13.6 ms | 74% |
| 16 | 869 | 829 (-4.6%) | 19.0 vs 17.8 ms | 2 of 8289 |

A query also refuses to split when it is already running on a pool worker (a nested fan-out on a bounded queue deadlocks rather than degrading, since the outer tasks hold every worker while waiting on inner tasks nothing is left to run), and when the caller holds uncommitted page changes, which a worker's own transaction context cannot see.

## Results are the serial ones, ties included

`RidScoreMinHeap` now breaks ties on RID ascending, so which of several equally-scored documents survives no longer depends on heap layout, and the merge ranks the same way. **This is a behaviour change**: a query with tied scores could previously return either document and now deterministically returns the lowest RID.

Scores can differ by one ulp between the two shapes, because MaxScore sums a document's terms in an order that follows the pruning split. Document set and ranking are unaffected; documented on the test that asserts it.

## Also here

- The caller scores one range itself rather than blocking on all of them.
- `topKGrouped` stays serial: merging grouped results needs the per-group worsts, not a flat best-k.
- Studio's Executor Pools card gains three columns on the sparse-vector row (reserved workers, queries in flight, queries split). Without them an idle row is ambiguous between "nobody querying", "load gate switched splitting off", "queries too small" and "broken".
- `MultiBucketSealedSegmentFanoutTest` covers multi-bucket over sealed segments, which had no coverage and passes today only because `LocalDatabase.checkDatabaseIsOpen` creates a thread context as a side effect. Nothing states that contract, so a refactor could have broken segment-backed multi-bucket queries silently.

## New settings

- `arcadedb.sparseVectorScoringMaxPartitions` (0 = adaptive, 1 = off, >1 = explicit and bypasses the load gate)
- `arcadedb.sparseVectorScoringMinPostingsForPartitioning` (default 200,000)

## Testing

Full `engine` module 10,128 tests and full `server` module 763 tests, zero failures. New `ParallelRangeTopKTest` covers range/serial equivalence at 2..16 ranges, tie determinism, empty and out-of-bounds ranges, engine-level equivalence, no nested split, no worker leak, and in-flight suppression.

Not verified: the Studio card rendering in a browser. The gauges and the JSON contract are pinned by `PoolMetricsTest`, but nobody has looked at the page.

## Worth a reviewer's attention

1. The load gate is a heuristic. It measured well at 1/4/16 clients on one machine and is the piece most likely to want tuning on other hardware.
2. The tie-break change is user-visible for tied scores.
3. Whether explicit `maxPartitions > 1` should really bypass the load gate, or whether an operator asking for N ranges should still yield under load.

https://claude.ai/code/session_01XF4M78mPjg8VH7GR5pPZEp

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -342,8 +342,11 @@ public enum GlobalConfiguration {
       self-throttling and splits regardless of load. That is a throughput risk on a busy server, \
       and it is paid by the forcing query too: measured independently at 16 concurrent clients, a \
       forced 8-way split returned 0.76x the throughput of no split at all with a median 1.75x \
-      worse, while compressing the tail (p99/p50 1.5x against 5.2x). Force it for tail \
-      predictability under load; leave it at 0 for median or throughput. Re-read on every query.""",
+      worse. What it did buy at that load is a lower tail, p99 297 ms against 579 ms. So the \
+      forced setting inverts as concurrency rises - a clear win on median while the box is quiet, \
+      a median and throughput loss once it is busy, and a tail win throughout. Force it for \
+      tail-sensitive traffic; leave it at 0 for anything else. The crossover is hardware and \
+      workload specific, so no number for it is given here. Re-read on every query.""",
       Integer.class, 0),
 
   SPARSE_VECTOR_SCORING_MIN_POSTINGS_FOR_PARTITIONING("arcadedb.sparseVectorScoringMinPostingsForPartitioning", SCOPE.JVM,
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -306,8 +306,10 @@ public enum GlobalConfiguration {
   SPARSE_VECTOR_SCORING_POOL_THREADS("arcadedb.sparseVectorScoringPoolThreads", SCOPE.JVM,
       """
       Maximum number of threads in the JVM-wide pool that backs LSM_SPARSE_VECTOR top-K \
-      fan-out (per-bucket parallel scoring on partitioned types and types with multiple \
-      buckets). Kept on its own pool rather than sharing the QueryEngineManager pool so \
+      fan-out: per-bucket parallel scoring on partitioned types and types with multiple \
+      buckets, and the RID-range split of a single index's traversal (see \
+      arcadedb.sparseVectorScoringMaxPartitions). Kept on its own pool rather than sharing \
+      the QueryEngineManager pool so \
       long-running graph algorithms never queue scoring tasks behind seconds-long graph \
       chunks. 0 = available cores (min 2). REQUIRES JVM RESTART: the pool is a lazy \
       singleton constructed once on first use; later changes to this value have no effect \
@@ -323,6 +325,32 @@ public enum GlobalConfiguration {
       but never fails the query. REQUIRES JVM RESTART: same singleton lifecycle as \
       SPARSE_VECTOR_SCORING_POOL_THREADS.""",
       Integer.class, 1024),
+
+  SPARSE_VECTOR_SCORING_MAX_PARTITIONS("arcadedb.sparseVectorScoringMaxPartitions", SCOPE.JVM,
+      """
+      Maximum number of RID ranges a single LSM_SPARSE_VECTOR top-K query is split into for \
+      parallel scoring. 1 disables intra-query parallelism and keeps every query on the caller \
+      thread; 0 (default) lets the engine decide per query. Splitting buys latency with CPU: each \
+      range prunes against its own top-K watermark, which rises more slowly than the global one, \
+      so a range does more work than its share (measured at roughly 1.9x total CPU for an 8-way \
+      split on a learned-sparse corpus). On the default the engine claims only workers no other \
+      query needs, and does not split at all once enough queries are already in flight to keep \
+      the pool busy on their own - so an idle server spends spare cores on latency while a \
+      saturated one keeps its throughput. Measured on an 18-worker box at 500k documents: one \
+      client 13.7 -> 3.1 ms p50, four clients 13.6 -> 4.1 ms with 61% more throughput, sixteen \
+      clients within 5% of serial throughput. Any explicit value above 1 opts out of that \
+      self-throttling and splits regardless of load, which is a throughput risk on a busy server. \
+      Re-read on every query.""",
+      Integer.class, 0),
+
+  SPARSE_VECTOR_SCORING_MIN_POSTINGS_FOR_PARTITIONING("arcadedb.sparseVectorScoringMinPostingsForPartitioning", SCOPE.JVM,
+      """
+      Minimum number of postings a LSM_SPARSE_VECTOR top-K query has to traverse before it is \
+      worth splitting into parallel ranges. Below this the fan-out (per-range cursor stacks, task \
+      dispatch, result merge) costs more than the traversal it parallelises, so the query stays on \
+      the caller thread. Counted as the summed document frequency of the query's dims, which is \
+      available from segment metadata without reading a page. Re-read on every query.""",
+      Long.class, 200_000L),
 
   SPARSE_VECTOR_SCORING_TIMEOUT_SECONDS("arcadedb.sparseVectorScoringTimeoutSeconds", SCOPE.JVM,
       """
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -336,7 +336,12 @@ public enum GlobalConfiguration {
       split on a learned-sparse corpus). On the default the engine claims only workers no other \
       query needs, and does not split at all once enough queries are already in flight to keep \
       the pool busy on their own - so an idle server spends spare cores on latency while a \
-      saturated one keeps its throughput. Measured on an 18-worker box at 500k documents: one \
+      saturated one keeps its throughput. One cost of the default worth knowing: in a narrow band \
+      of moderate concurrency, where some queries claim a wide split and others get none, the \
+      spread between the two shows up as tail latency. Measured at 4 concurrent clients on an \
+      18-thread pool, p99 was about 1.2x serial's while the median was 3.4x better (7.6 ms against \
+      26.2). The band closes on both sides - below it every query splits, above it none does - and \
+      the median win is large enough that the default keeps the trade rather than flattening it. Measured on an 18-worker box at 500k documents: one \
       client 13.7 -> 3.1 ms p50, four clients 13.6 -> 4.1 ms with 61% more throughput, sixteen \
       clients within 5% of serial throughput. Any explicit value above 1 opts out of that \
       self-throttling and splits regardless of load. That is a throughput risk on a busy server, \
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -340,13 +340,14 @@ public enum GlobalConfiguration {
       client 13.7 -> 3.1 ms p50, four clients 13.6 -> 4.1 ms with 61% more throughput, sixteen \
       clients within 5% of serial throughput. Any explicit value above 1 opts out of that \
       self-throttling and splits regardless of load. That is a throughput risk on a busy server, \
-      and it is paid by the forcing query too: measured independently at 16 concurrent clients, a \
-      forced 8-way split returned 0.76x the throughput of no split at all with a median 1.75x \
-      worse. What it did buy at that load is a lower tail, p99 297 ms against 579 ms. So the \
-      forced setting inverts as concurrency rises - a clear win on median while the box is quiet, \
-      a median and throughput loss once it is busy, and a tail win throughout. Force it for \
-      tail-sensitive traffic; leave it at 0 for anything else. The crossover is hardware and \
-      workload specific, so no number for it is given here. Re-read on every query.""",
+      and it is paid by the forcing query too. Measured at 16 concurrent clients on an 18-thread \
+      pool, 1M documents: a forced 8-way split returned 0.52x the throughput of no split at all, a \
+      median 1.85x worse and a p99 2.5x worse, for 1.9x the CPU per query. Above light concurrency \
+      there is no regime where forcing wins - not median, not throughput, not tail - because a \
+      query's own ranges start queueing behind its neighbours'. It is worth setting only when a \
+      single query at a time must be as fast as possible and nothing else is running; for anything \
+      else 0 is faster on every measure. The crossover is hardware and workload specific, so no \
+      number for it is given here. Re-read on every query.""",
       Integer.class, 0),
 
   SPARSE_VECTOR_SCORING_MIN_POSTINGS_FOR_PARTITIONING("arcadedb.sparseVectorScoringMinPostingsForPartitioning", SCOPE.JVM,
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -339,8 +339,11 @@ public enum GlobalConfiguration {
       saturated one keeps its throughput. Measured on an 18-worker box at 500k documents: one \
       client 13.7 -> 3.1 ms p50, four clients 13.6 -> 4.1 ms with 61% more throughput, sixteen \
       clients within 5% of serial throughput. Any explicit value above 1 opts out of that \
-      self-throttling and splits regardless of load, which is a throughput risk on a busy server. \
-      Re-read on every query.""",
+      self-throttling and splits regardless of load. That is a throughput risk on a busy server, \
+      and it is paid by the forcing query too: measured independently at 16 concurrent clients, a \
+      forced 8-way split returned 0.76x the throughput of no split at all with a median 1.75x \
+      worse, while compressing the tail (p99/p50 1.5x against 5.2x). Force it for tail \
+      predictability under load; leave it at 0 for median or throughput. Re-read on every query.""",
       Integer.class, 0),
 
   SPARSE_VECTOR_SCORING_MIN_POSTINGS_FOR_PARTITIONING("arcadedb.sparseVectorScoringMinPostingsForPartitioning", SCOPE.JVM,
--- a/engine/src/main/java/com/arcadedb/index/sparsevector/BmwScorer.java
+++ b/engine/src/main/java/com/arcadedb/index/sparsevector/BmwScorer.java
@@ -120,16 +120,41 @@ public final class BmwScorer {
    */
   public static List<RidScore> topK(final int[] queryDims, final float[] queryWeights, final DimCursor[] cursors, final int k)
       throws IOException {
+    return topK(queryDims, queryWeights, cursors, k, null, null);
+  }
+
+  /**
+   * {@link #topK(int[], float[], DimCursor[], int)} restricted to the RID range
+   * {@code [startInclusive, endExclusive)}. This is the unit of work of the parallel fan-out in
+   * {@link PaginatedSparseVectorEngine#topK}: each worker takes one range, and because the ranges
+   * partition the RID space, concatenating their results and keeping the best {@code k} yields
+   * exactly the serial top-K.
+   * <p>
+   * <b>A range costs more than its share.</b> The pruning threshold is per-traversal, so a worker
+   * that sees 1/P of the corpus reaches a lower watermark than the serial scan would at the same
+   * point and therefore prunes less. The partitioned traversal does strictly more total work than
+   * the serial one - measured at about 1.9x total CPU for an 8-way split on a learned-sparse corpus.
+   * It buys latency with CPU, which is why the caller gates the fan-out rather than always taking it.
+   * <p>
+   * Cursors are seeked to {@code startInclusive} <i>before</i> each term's ceiling is captured, so a
+   * worker prunes against the maximum weight remaining in <i>its</i> range rather than the whole
+   * dim's - which claws back part of that amplification.
+   *
+   * @param startInclusive first RID to consider, or {@code null} to start at the beginning
+   * @param endExclusive   first RID <i>not</i> to consider, or {@code null} to run to the end
+   */
+  public static List<RidScore> topK(final int[] queryDims, final float[] queryWeights, final DimCursor[] cursors, final int k,
+      final RID startInclusive, final RID endExclusive) throws IOException {
     validate(queryDims, queryWeights, cursors);
     if (k <= 0)
       return List.of();
 
-    final DimEntry[] terms = openTerms(queryWeights, cursors);
+    final DimEntry[] terms = openTerms(queryWeights, cursors, startInclusive);
     if (terms.length == 0)
       return List.of();
 
     final TopKCollector collector = new TopKCollector(k);
-    scan(terms, collector);
+    scan(terms, collector, endExclusive);
     return collector.drain();
   }
 
@@ -185,12 +210,12 @@ public final class BmwScorer {
     if (limit <= 0 || groupSize <= 0)
       return List.of();
 
-    final DimEntry[] terms = openTerms(queryWeights, cursors);
+    final DimEntry[] terms = openTerms(queryWeights, cursors, null);
     if (terms.length == 0)
       return List.of();
 
     final GroupedCollector collector = new GroupedCollector(limit, groupSize, groupKeyResolver, allowedRIDs);
-    scan(terms, collector);
+    scan(terms, collector, null);
     return collector.drain();
   }
 
@@ -208,7 +233,9 @@ public final class BmwScorer {
    * dominates everything else once the posting mass has been pruned away. The heap turns that into
    * O(aligned * log terms).
    */
-  private static void scan(final DimEntry[] terms, final Collector collector) throws IOException {
+  private static void scan(final DimEntry[] terms, final Collector collector, final RID endExclusive) throws IOException {
+    final int endBucketId = endExclusive != null ? endExclusive.getBucketId() : -1;
+    final long endPosition = endExclusive != null ? endExclusive.getPosition() : -1L;
     final int n = terms.length;
     // prefix[i] == sum of sigma over terms [0, i). prefix[n] is the whole query's ceiling.
     final float[] prefix = new float[n + 1];
@@ -258,6 +285,12 @@ public final class BmwScorer {
       // at or below the threshold by construction of the split.
       final int candidateBucketId = keyBucketIds[heap[0]];
       final long candidatePosition = keyPositions[heap[0]];
+
+      // Range bound: candidates are produced in ascending RID order, so the first one at or past the
+      // end of this worker's range means the range is done - everything left belongs to a sibling.
+      if (endBucketId >= 0
+          && SparseSegmentBuilder.compareRid(candidateBucketId, candidatePosition, endBucketId, endPosition) >= 0)
+        return;
 
       // Detach the whole aligned run so the cursors can be moved without corrupting the heap order.
       int alignedCount = 0;
@@ -537,13 +570,24 @@ public final class BmwScorer {
    * Start every non-null cursor, drop the ones with nothing to iterate, and order the survivors by
    * {@link DimEntry#BY_PRUNING_VALUE promotion order}, i.e. by how much traversal work each term
    * would remove per unit of pruning budget it consumes.
+   * <p>
+   * When a {@code startInclusive} is given, each cursor is seeked there <i>before</i> its
+   * {@link DimEntry} is built. That ordering matters: a term's ceiling is captured from
+   * {@link DimCursor#upperBoundRemaining()}, which on a sealed segment is a suffix max from the
+   * cursor's current position, so seeking first gives a worker the maximum weight remaining in its
+   * own range rather than the whole dim's. A tighter ceiling means a term can be proven
+   * non-essential sooner, which is the one lever that offsets a partitioned traversal's weaker
+   * pruning threshold.
    */
-  private static DimEntry[] openTerms(final float[] queryWeights, final DimCursor[] cursors) throws IOException {
+  private static DimEntry[] openTerms(final float[] queryWeights, final DimCursor[] cursors, final RID startInclusive)
+      throws IOException {
     final List<DimEntry> live = new ArrayList<>(cursors.length);
     for (int i = 0; i < cursors.length; i++) {
       if (cursors[i] == null)
         continue;
       cursors[i].start();
+      if (startInclusive != null)
+        cursors[i].seekTo(startInclusive);
       if (cursors[i].isExhausted())
         continue;
       live.add(new DimEntry(cursors[i], queryWeights[i]));
@@ -576,8 +620,39 @@ public final class BmwScorer {
     void collect(RID rid, float score);
   }
 
-  /** Result ordering handed back to the caller: best score first. */
-  private static final Comparator<RidScore> BY_SCORE_DESC = (a, b) -> Float.compare(b.score(), a.score());
+  /**
+   * Result ordering handed back to the caller: best score first, ties broken by RID ascending.
+   * <p>
+   * The RID leg is what makes a partitioned traversal indistinguishable from the serial one
+   * (issue #4085). It is the same total order {@link RidScoreMinHeap} retains against, so sorting
+   * the concatenated per-range results and keeping the first {@code k} reproduces the serial result
+   * exactly, ties included.
+   */
+  static final Comparator<RidScore> BY_SCORE_DESC = (a, b) -> {
+    final int c = Float.compare(b.score(), a.score());
+    return c != 0 ? c : SparseSegmentBuilder.compareRid(a.rid(), b.rid());
+  };
+
+  /**
+   * Merge the per-range results of a partitioned {@link #topK(int[], float[], DimCursor[], int, RID, RID)}
+   * back into the global top-K.
+   * <p>
+   * Correct because the ranges partition the RID space: every document is scored by exactly one
+   * worker, against the same query, so a document's score does not depend on which range it fell in.
+   * Only the <i>pruning</i> differs - a worker with a weaker local watermark keeps candidates the
+   * serial scan would have discarded - and keeping the best {@code k} of the union discards those
+   * again.
+   */
+  static List<RidScore> mergeRanges(final List<List<RidScore>> ranges, final int k) {
+    int total = 0;
+    for (final List<RidScore> r : ranges)
+      total += r.size();
+    final List<RidScore> all = new ArrayList<>(total);
+    for (final List<RidScore> r : ranges)
+      all.addAll(r);
+    all.sort(BY_SCORE_DESC);
+    return all.size() <= k ? all : new ArrayList<>(all.subList(0, k));
+  }
 
   /** Plain top-K: a K-sized min-heap whose head is the threshold once full. */
   private static final class TopKCollector implements Collector {
--- a/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
+++ b/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
@@ -474,6 +474,13 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
    * the same number of that dim's postings. That is a better proxy for "same amount of work" than
    * cutting the RID space into equal spans, which goes lopsided as soon as a range of RIDs has been
    * deleted or was never dense. Reading them touches in-memory block metadata only - no page reads.
+   * <p>
+   * It is a balance heuristic, not a guarantee, and it reads one dim of one segment. On an index with
+   * several live segments - a freshly loaded one, before compaction has merged them - that segment's
+   * layout need not represent the global RID distribution, and the ranges can come out uneven.
+   * Correctness does not depend on it: the ranges still partition the RID space disjointly and every
+   * document is still scored exactly once. Only the speedup suffers, and it shows up as one range
+   * finishing long after its siblings.
    */
   private RID[] planPartitionBoundaries(final int[] queryDims, final PaginatedSegmentReader[] segSnapshot) throws IOException {
     final int configured = GlobalConfiguration.SPARSE_VECTOR_SCORING_MAX_PARTITIONS.getValueAsInteger();
@@ -519,6 +526,7 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
       // stacks", and past a small multiple of the pool there are no threads left to run them anyway.
       partitions = Math.min(Math.min(configured, byLayout), pool.getMaxParallelism() * EXPLICIT_PARTITION_OVERSUBSCRIPTION);
       pool.reserveWorkers(partitions - 1);
+      pool.warnExplicitSplitUnderLoad(partitions);
     } else {
       final int wanted = Math.min(pool.getMaxParallelism(), byLayout) - 1;
       final int granted = pool.tryReserveWorkers(wanted);
--- a/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
+++ b/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
@@ -372,6 +372,12 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
     final Memtable mtSnapshot = memtable.get();
     final PaginatedSegmentReader[] segSnapshot = segments.get();
 
+    // Splitting off globally: return before touching the pool at all. getInstance() would build a
+    // ThreadPoolExecutor that this JVM has just been told it will never use, which an embedded
+    // deployment that disabled the feature should not be paying for.
+    if (GlobalConfiguration.SPARSE_VECTOR_SCORING_MAX_PARTITIONS.getValueAsInteger() == 1)
+      return rangeTopK(queryDims, queryWeights, k, mtSnapshot, segSnapshot, null, null);
+
     // Registered for the whole call, split or not: the split decision is made against how many
     // queries are running, and a query that stays serial still occupies its caller's thread.
     final SparseVectorScoringPool pool = SparseVectorScoringPool.getInstance();
@@ -441,8 +447,18 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
     final DatabaseContext.DatabaseContextTL ctx = DatabaseContext.INSTANCE.getContextIfExists(database.getDatabasePath());
     if (ctx == null)
       return false;
-    final TransactionContext tx = ctx.getLastTransaction();
-    return tx != null && tx.isActive() && tx.hasChanges();
+    // Every transaction on the stack, not just the innermost. begin() on an already-active
+    // transaction pushes a nested one, so a caller can sit in a fresh inner transaction with no
+    // changes of its own while an outer one holds modified pages. Asking only the innermost would
+    // report "clean" and let the query split, and the workers - reading committed pages through a
+    // context of their own - would silently not see the outer transaction's writes. That is the
+    // exact class of bug this guard exists to remove, so it has to look at all of them.
+    for (int i = 0; i < ctx.transactions.size(); i++) {
+      final TransactionContext tx = ctx.transactions.get(i);
+      if (tx != null && tx.isActive() && tx.hasChanges())
+        return true;
+    }
+    return false;
   }
 
   /**
@@ -696,6 +712,16 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
     final Memtable mtSnapshot = memtable.get();
     final PaginatedSegmentReader[] segSnapshot = segments.get();
 
+    // Grouped queries never split, but they are still load: they run a full traversal on their
+    // caller's thread and compete for the same cores. Leaving them unregistered would let a plain
+    // topK arriving alongside heavy grouped traffic see an idle gate, split, and take throughput
+    // from them - which is precisely what the gate exists to prevent, merely from a source it could
+    // not see. Registered only when splitting is enabled at all, for the same reason as topK.
+    final boolean registerLoad = GlobalConfiguration.SPARSE_VECTOR_SCORING_MAX_PARTITIONS.getValueAsInteger() != 1;
+    final SparseVectorScoringPool pool = registerLoad ? SparseVectorScoringPool.getInstance() : null;
+    if (pool != null)
+      pool.queryStarted();
+
     final DimCursor[] cursors = new DimCursor[queryDims.length];
     try {
       for (int i = 0; i < queryDims.length; i++)
@@ -705,6 +731,8 @@ public final class PaginatedSparseVectorEngine implements AutoCloseable {
       for (final DimCursor c : cursors)
         if (c != null)
           c.close();
+      if (pool != null)
+        pool.queryFinished();
     }
   }
 
--- a/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
+++ b/engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java
@@ -19,8 +19,10 @@
 package com.arcadedb.index.sparsevector;
 
 import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.database.DatabaseContext;
 import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.RID;
+import com.arcadedb.database.TransactionContext;
 import com.arcadedb.engine.ComponentFile;
 import com.arcadedb.index.IndexException;
 import com.arcadedb.log.LogManager;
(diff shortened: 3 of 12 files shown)


lvca:
Thanks, this was a useful review. Pushed as `7bf17702a`. Three of the four taken as written, one fixed differently, and fixing it turned up a regression worth recording.

**1. Reservation acquired and released in different methods.** Agreed, and fixed. The claim is now released on any throw between acquiring it and returning the boundaries. You are right that the walk cannot currently throw, but the method reads segment metadata and is declared `throws IOException`, so the next edit that adds a page read there would have leaked a claim permanently with nothing failing to point at it.

**2. `inFlightQueries` double-counts the multi-bucket fan-out.** The bug is real, the suggested fix would have introduced the opposite one. Those eight per-bucket queries are genuine load, so making them invisible would leave a split free to oversubscribe against them. Instead the counter is now split in two: caller-thread queries drive the load gate, pool-thread queries subtract from grantable capacity. Each measures the thing it is good at.

Worth flagging what that cost me. My first implementation read `ThreadPoolExecutor.getActiveCount()` for the second half, which looks like the obvious source and is a trap: it takes the pool's main lock and walks the worker set. The benchmark caught it immediately - at 16 concurrent clients throughput fell to **549 qps against 772 serial while only 2 of 5499 queries split**. Splitting was off, so the entire loss was in the decision path: a decision more expensive than the decision it was making. A plain `AtomicInteger` put it back to **718 against 709**. It also retroactively explains an "activity-aware" measurement earlier in this work that I had written off as noise. The trap is now documented at the call site so nobody re-adds it.

**3. Explicit `maxPartitions` has no clamp.** Agreed, now capped at 4x pool size. The setting means "do not throttle me", not "open a thousand cursor stacks"; past a small multiple of the pool there is no thread free to run them anyway.

**4. Tie-break affects serial-only users.** Correct, and the more important half of that point. Now documented where the behaviour lives: the ordering is in the collector, so a database with splitting disabled entirely gets it too, and anyone whose scores tie can see a different retained set than a previous release.

**Nits.** Import fixed. The deadline path is now tested, and deterministically rather than by racing a clock: every worker is jammed with a latch so the range gets submitted but never started, and the test also asserts the timed-out query hands its claim back.

## Verification, including what I could not finish

- Sparse-vector and vector suites: **403 tests green**
- Server module, in full: **768 tests green**
- Throughput re-measured after the reservation change: 1 client 3.80 ms p50 vs 14.4 serial, 4 clients +67% throughput at 4.74 ms vs 16.5, 16 clients at parity (718 vs 709 qps) with 2 of 7184 split

The full `engine` module did **not** complete. Three attempts were killed by SIGTERM (exit 143) on three different unrelated classes - `TestSortedIndexBuildBenchmark`, `GAVEligibilityTest`, `ACIDTransactionTest` - with no `hs_err` files, so an external process killer on this machine rather than a JVM crash. The best of those runs reached **6765 tests with zero failures** before being cut off, and the same command completed 10,128 green on the parent commit.

What bounds the risk: every production file in this commit is inside `com.arcadedb.index.sparsevector`, and `RidScoreMinHeap` is package-private with exactly one consumer in the same package. That package's suite is the 403 green above. Still worth a clean full-module run in CI before merge, and I would not claim otherwise.

lvca:
Good catch on #1, and it is worse than a crash. Fixed in `8cb6731b8`.

**1. Caller-runs rolling back the caller's transaction.** Confirmed exactly as described. `DatabaseContext.init` on a thread that already holds a context takes the "ROLLBACK PREVIOUS TXS" branch and rolls back every transaction on it, and my `finally removeContext` then wiped the context the rest of the query still needed. So the visible failure is `Transaction context not found on current thread`, sitting on top of a user transaction that has already been silently rolled back - the quiet half being the dangerous one.

Fixed with the create-only-if-absent guard you suggested, tearing down only what the task created. Your diagnosis of why the pattern was unsafe here is the part I had wrong when I wrote it: I copied `FetchFromTypeExecutionStep`'s init/remove pair without noticing that it is safe only because `ParallelScanProducerPool` has an unbounded queue and therefore never runs a task on the caller. That distinction is now recorded at the call site so the next person copying it sees the constraint.

The new test drives it deterministically rather than hoping to hit the window: it saturates every worker and every queue slot so the range is rejected onto the caller, with the caller inside an open transaction, and asserts both the result and that the transaction and its context are still alive afterwards. Verified it fails without the fix with precisely the reported symptom, so it is a real regression test and not a decoration.

You were also right that `aRangeThatNeverGetsAWorkerFailsTheQueryOnTheDeadline` does not reach this branch: it jams workers but leaves queue capacity, so submission succeeds and nothing runs inline.

**2. Shared deadline consumed by the caller's own range 0.** Agreed, fixed. A future that is already done is now harvested regardless of the deadline; the timeout governs waiting, not collecting. Failing a query while every worker range sits complete and waiting to be read was the wrong outcome, however unlikely balanced ranges make it.

**3.** Noted, thanks for checking the merge behaviour at the k-th boundary and the Studio colspan - both were on my list of things I believed but had not proven.

## Verification

- `ParallelRangeTopKTest`: 10 tests green, including the new caller-runs case
- Sparse-vector and vector suites: 404 green
- Full `server` module was green on the previous commit; this commit touches only `PaginatedSparseVectorEngine` and one test

Still outstanding from my last comment: the full `engine` module has not completed on this machine - three runs killed by SIGTERM on unrelated classes, best reaching 6765 tests with zero failures. Worth a clean CI run before merge.

lvca:
Thanks. Review points addressed in `2ec078f8d`, and CI then caught something worse than any of them, fixed in `013468d80`.

## CI: my test hung the unit-tests job for an hour

`unit-tests` failed at 1h0m14s, and the thread dumps name
`aRangeForcedOntoTheCallerThreadLeavesTheCallersTransactionAlone` - the caller-runs test from
the last round. The irony is exact: it deadlocked on the very mechanism it was written to
test.

It saturated the pool by computing "workers + free queue slots" from a snapshot and then
submitting that many blocking tasks. The snapshot races, and overshooting by one is not a
slow test but a permanently stuck one - the surplus task is rejected, caller-runs executes it
on the submitting thread, and it waits on a latch released only in a `finally` that thread can
no longer reach. Locally the arithmetic happened to come out right, which is why it passed
every run here and hung on the first CI runner with a different core count.

Fixed twice over, either half being sufficient: submit until the queue reports itself full
rather than computing a count, and have a saturation task block only when it finds itself on a
worker, with a timeout rather than indefinitely. A mis-sized saturation now fails the test
instead of wedging the job, which is the difference between a diagnosis and an hour of
nothing. The same shape was in the deadline test's hogs, so it got the same treatment.

I re-verified the test still earns its place rather than being quietly defanged: with the
production `contextCreated` guard removed it still fails with "Transaction context not found
on current thread", so the rewritten loop still saturates for real.

## The other two CI failures are not mine

- `integration-tests`: `Issue4141SessionManagementIT` gets HTTP 400 from
  `/api/v1/command/graph`. The same job fails identically on #5520, an unrelated PR that has
  already merged, and nothing in this branch touches HTTP, sessions or command handling.
- `ha-integration-tests`: failed too; I have not diagnosed it yet and will not claim it is
  unrelated until I have.

## Review points

**1. Test tagging.** Not taken, with a measurement: the class runs 10 tests in 1.46 s, which is
not what CLAUDE.md means by noticeably long, and `@Tag("slow")` would pull a
transaction-safety regression test out of regular CI. (It is 3.5 s after the saturation fix,
still under the bar.)

**2. `queries.split` vs effective parallelism.** Good point, taken. It counts the decision, not
the outcome, and under caller-runs a query counted as split can have run serially. Now stated
in the metric description and on the Studio card, next to the Caller-Run Fallbacks column
where it shows up.

**3. Balance heuristic with multiple live segments.** Taken as documentation. Boundaries come
from one dim of one segment, so before compaction merges several live segments that layout
need not represent the global RID distribution and ranges can come out uneven. Correctness is
unaffected - the ranges still partition the RID space and every document is scored exactly
once - only the speedup, and it shows up as one range finishing after its siblings.

**On the explicit-`maxPartitions` question.** I took your suggestion: the bypass stays, the
clamp stays, and an explicit split granted while enough queries are in flight that the adaptive
default would have refused now logs a throttled WARNING. The setting is JVM-wide and
long-lived, so whoever configured it is rarely the person watching latency later.

tae898:
**#5518 reproduces on real SPLADE: 3.44x at 1M.** Your 3.5x holds on Big-ANN's SPLADE-cocondenser corpus, not just a SPLADE-shaped one.

Wheel built from `issue-4085-parallel-topk` merged onto our main, engine jars compiled locally and passed via `JAR_LIB_DIR`; `SparseVectorScoringPool` verified present in the jar before measuring. Baseline is released 26.8.1.dev21 on the same host, same 995-query set, cpuset 0-11, INT8, settled, N=3, median.

| tier | `maxPartitions=1` (serial) | `=0` (adaptive default) | `=8` (forced) |
|---|---|---|---|
| **1M** | 39.31 ms | **11.42 ms (3.44x)** | 13.30 ms (2.96x) |
| **100k** | 7.20 ms | 4.25 ms (1.69x) | 4.11 ms (1.75x) |

recall@10 identical across all six cells, 0.9935 at 1M and 0.9950 at 100k, so the split is exact on our data as well as your tests.

**Two things you may not have measured.**

**Your adaptive default beats a forced split at 1M**, 11.42 against 13.30 ms. Whatever the heuristic is doing with `minPostingsForPartitioning` and the worker claim, it picks better than a fixed 8 on this corpus. I would not have guessed that, and it argues for leaving the default at 0 rather than documenting 8 as a recommended value.

**The 100k tier gains much less**, 1.69x against 3.44x. That fits your "pays in proportion to how much work the query does" exactly: our 1M queries carry 1.3M to 4.3M summed postings against a 200k threshold, so partitioning engages on essentially all of them, while at 100k a good share of queries sit near or under the gate.

**What it does to the comparison.** At 1M against Qdrant's 2.91 ms, we go from 12.7x to **3.9x**. For context on the trajectory, the same tier measured 165 ms when we first filed the cliff, so this is 165 -> 36.8 (released dev21) -> 11.4 with your split.

**How I will report it.** Prose only for now, marked development line, since the branch is unmerged and our tables only carry released wheels. If it merges and ships, it moves into the table on the next release.

**The caveat is yours and I will carry it.** Our lane is single-stream, so we measure the favourable end of a latency-versus-throughput trade; your 1.89x total CPU at 8 ranges is real and our numbers do not show it. Anyone reading our sparse row should know the split buys latency with cores on an idle box. If you want, I can run a concurrent-client arm to put a number on what it costs under load, which is the half our harness currently cannot see.

Running now at 8.84M, serial versus adaptive, since that is where the traversal does the most work and where the gap to specialists is widest. I will post it whichever way it goes.

lvca:
All four addressed in `1cdb06968`, and two of them were real holes rather than polish.

**1. Nested transactions - you were right, and it was reachable.** `LocalDatabase.begin()` on an already-active transaction pushes a nested one, so a caller sitting in a fresh inner transaction over a dirty outer one reported clean, split, and its workers would have read committed pages without seeing what the outer transaction had written. The guard now checks every transaction on the stack. There is a test for it, and I verified it fails against the innermost-only version with "a nested-clean caller over a dirty outer transaction must not split", so it is pinned rather than asserted.

**2. `topKGrouped` invisible to the gate - also right, and not intentional.** Grouped queries never split, but they run a full traversal on the caller's thread and compete for the same cores. Unregistered, a plain `topK` arriving alongside heavy grouped traffic saw an idle gate, split, and took throughput from it - the failure the gate exists to prevent, arriving from a source it could not see. Now registered, on the same condition as `topK`.

**3. The `* 2` literal.** Named `CALLER_LOAD_GATE_FACTOR`, with what it means and the 1/4/16-client measurements behind it in the Javadoc. I kept it a constant rather than a setting: it shapes only the middle of the range, and an operator who wants splitting off or always-on already has `sparseVectorScoringMaxPartitions` for both, so a third knob would be easy to set wrongly and hard to reason about. Named so that whoever finds a machine where the shape is wrong can find it.

**4. Serial path touching the pool.** Taken - with `maxPartitions=1` the query now returns before `getInstance()`. Worth more than a nit for embedded users: `getInstance()` builds a `ThreadPoolExecutor` that the JVM has just been told it will never use.

**Nits.** The caller-as-extra-concurrency point is exactly as you read it. The Studio browser check is still not done and I am still not claiming otherwise.

---

Separately, @tae898 has now run this branch on the real Big-ANN SPLADE corpus, and it reproduces: **1M 39.31 ms serial to 11.42 ms adaptive, 3.44x**, recall@10 identical across every cell. Two things from his run worth folding into the record here:

- **The adaptive default beats a forced `=8`** at 1M, 11.42 against 13.30 ms. So `0` is the right default and we should not document 8 as a recommended value.
- **100k gains only 1.69x**, which matches the "pays in proportion to the work" caveat: at that tier a good share of queries sit near or under the 200k posting threshold and never split.

He is carrying the CPU-cost caveat explicitly, since his lane is single-stream and therefore measures the favourable end of the trade. He has offered a concurrent-client arm to put an independent number on what it costs under load, which is the half our own measurements cover and his currently cannot.

lvca:
Point 1 is a fair hit and fixed in `77e9072cc`. The guard caught `RuntimeException | Error` while its own comment said it existed for a future page read - which throws `IOException` and would have sailed straight past it. Worse, `topK`'s `finally` cannot cover that case either, since it only releases inside the branch where boundaries came back non-null, so the claim would have leaked precisely as the comment warned. A guard that documents an intent it does not implement is worse than no guard, because the next person reads the comment and believes it.

Rather than adding `IOException` to the clause, I replaced it with a handed-off flag and a `finally`. Enumerating exception types is how this kind of guard fails quietly - it is the same mistake one level down - whereas a `finally` covers checked, unchecked, `Error`, and any future early return that forgets the claim exists.

**On point 2, the tie-break in the changelog.** Agreed it needs to be there and it is not something a reader should have to infer from a PR body. Flagging it for @lvca to place, since release notes are his call rather than something I should be writing into the tree. The shape of it: any query whose scores tie can return a different retained set than a prior release, including serial queries and `topKGrouped`, because the ordering lives in the collector rather than in the split path. The new answer is the stable one - the old set depended on heap array layout, so it was not reproducible across insertion orders - but it is a behaviour change and not opt-in.

**Point 3**, agreed, no change.

Thanks for the read on `MultiBucketSealedSegmentFanoutTest`. That one was a side effect of hitting the missing-context wall from the other direction, and you are right that the contract it pins - worker threads getting a database context only as a side effect of `checkDatabaseIsOpen` - is worth having covered independently of this feature.

Two things still open on my side, unchanged: `ha-integration-tests` is failing and I have not yet diagnosed it, and the Studio card has still not been looked at in a browser.

tae898:
On review item 1, the tie/ulp gap: we may be able to supply the missing evidence, because our sparse lane runs exactly the corpus class the item says is untested, real Big-ANN SPLADE-cocondenser at INT8.

**What we already have.** Across serial (`maxPartitions=1`), the adaptive default (`=0`) and forced 8 ranges, on 995 dev queries at both 100k and 1M, N=3 each, recall@10 is identical to four decimal places in all six cells: 0.9950 at 100k and 0.9935 at 1M.

**What that is worth, honestly: less than it looks.** Recall is computed against Big-ANN's exact ground truth, so it counts how many of the true top-10 were returned, not which documents came back. Two different retained sets can score the same recall, which is precisely the failure the review describes (a tied pair straddling k/(k+1) swaps one document for another of equal score). So our numbers are consistent with set equivalence and are not evidence of it. I would not want them cited as if they closed the question.

**So I have queued the direct check.** One build of the 1M INT8 index, then three query passes over the same 1000 queries at `maxPartitions` 1 / 0 / 8, comparing the returned id lists per query and reporting two counts separately:

- queries where the returned **set** differs from serial (the correctness question)
- queries where the set matches but the **order** differs (the ulp-level tie-break, which the PR already scopes)

If the set count is zero across 1000 real quantized queries, that is a stronger statement than the current tie test can make, since our corpus has the genuine near-ties the `0.5*4 = 2.0` fixture removes by construction. If it is non-zero, you get concrete examples with the differing ids, which is the case worth softening the wording for.

It is behind an 8.84M job on our box, so a few hours out. I will post the counts whichever way they come, including if they contradict the equivalence claim.

One thing I cannot test from here: the review is right that the divergence depends on the order MaxScore sums a document's terms relative to where a range picks up the watermark. Our probe only observes the outcome, so a zero result is empirical evidence over one corpus and query set, not a proof over the algorithm.

lvca:
Point 1 is right, and it is the sharpest thing any of these rounds has turned up. Fixed in `89c60b147`.

I had reasoned about rounding and about ties, and asserted each separately, without noticing they are not independent. MaxScore sums a document's terms in an order that follows the essential/non-essential split; a range reaches that split at a different watermark than the whole scan; so two documents that tie exactly in one shape can sit an ulp apart in the other, and at the k boundary an ulp decides which is kept. **The RID tie-break cannot reach that case at all** - it only orders scores that are still equal after rounding. My claim of "identical list, ties included" was therefore wrong in exactly one corner.

Your diagnosis of why the test could not catch it is also exact: `0.5f` on four dims sums to `2.0` in any order, so that corpus removes the rounding and exercises the tie-break on its own. It was pinning half of an interaction.

New test builds the half that was missing - a plateau of documents with mathematically equal scores straddling the k-th place, from weights with no exact binary representation - and asserts what actually holds rather than what I wished held:

- the same number of documents,
- rank for rank indistinguishable in score,
- and nothing ranking below the serial k-th.

In other words the split never returns a *worse* answer; it may return a *different* one among documents that nothing can tell apart. That is the honest guarantee, and it is now the one stated in the class Javadoc instead of the stronger one.

Worth being explicit for the changelog, alongside the tie-break note: the guarantee users get is "as good as serial", not "byte-identical to serial", and the difference only ever appears among documents whose scores are indistinguishable.

**2. `queries.in_flight` description.** Right, it stopped matching its counter when I split the caller-thread and pool-thread counts apart. The gauge is correct - it reports the number the gate actually compares against - so I fixed the description rather than the number, and said why the exclusion is deliberate: counting per-bucket work here would let one multi-bucket query read as several and gate unrelated ones, which is the bug an earlier round of yours found. It now points at `pool.active` for that load.

**3.** Noted on the `ctx.transactions` coupling. I left it as the direct walk for now - a `hasUncommittedChanges()` accessor on `DatabaseContextTL` would read better and would belong to that class rather than this one, but it is a wider change than this PR should carry. Thanks for confirming the `parallel_scan` dash rendering.

tae898:
Both probes I promised are done, on real Big-ANN SPLADE at 1M, INT8, released dev21 as the baseline and this branch's wheel for the split arms. One answers the tie/ulp question, one prices the split under load and speaks to the open review point about explicit `maxPartitions` on a busy box.

## 1. Equivalence on 1000 real quantized queries: no divergence at all

1000 dev queries, k=10, three settings, comparing the returned id lists against `maxPartitions=1` query by query. I measured the guarantee you now state rather than the one the PR originally claimed, so: same count, rank for rank indistinguishable in score, and nothing ranking below the serial k-th.

| vs serial | different SET | same set, different ORDER | different COUNT | rank-score mismatch | **below serial k-th** | max rank-for-rank score delta |
|---|---|---|---|---|---|---|
| adaptive (`=0`) | 0 | 0 | 0 | 0 | **0** | 9.0e-06 |
| forced 8 | 0 | 0 | 0 | 0 | **0** | 9.0e-06 |

So on this corpus the split is not merely as good as serial, it is byte-identical: same documents, same order, every time.

**The interesting number is the one that is not zero.** The maximum rank-for-rank score difference is 9.0e-06, on scores of order 20, so about 0.45 ppm. That is direct evidence the summation-order effect you described is real and measurable in the arithmetic. It just stays roughly four orders of magnitude below what would be needed to flip a decision at the k boundary on this weight distribution.

**What this does not do is prove the corner cannot happen.** It shows it did not arise across 1000 real queries at three partition settings. Your weakened guarantee is still the right one to state, and the plateau test you added is still the thing that pins it, because it constructs the case deliberately rather than hoping to meet it. If anything this pair is the useful combination: your test proves the bound holds when the case occurs, and this says the case is rare enough on real SPLADE that nobody should expect to see it.

## 2. Under load: your gate works, and forcing a split is worse than not splitting

Closed loop, N Python client threads against one index, 30 s per cell, same 1000 queries. CPU is charged from the container cgroup, so it counts worker threads, which a per-query wall clock cannot see.

| clients | serial p50 / qps / cpu-ms-per-q | adaptive | forced 8 |
|---|---|---|---|
| 1 | 39.40 / 19.4 / 53.2 | **11.64 / 71.7 / 145.8** | 13.65 / 61.0 / 105.2 |
| 2 | 38.52 / 39.3 / 51.8 | 19.35 / 72.6 / 129.5 | 22.24 / 77.4 / 139.1 |
| 4 | 43.64 / 69.3 / 57.9 | 29.99 / 81.6 / 128.4 | 45.98 / 81.0 / 147.5 |
| 8 | 66.89 / 93.5 / 85.7 | 62.60 / 94.8 / **84.2** | 96.00 / 80.8 / 148.4 |
| 16 | 111.06 / 106.7 / 112.0 | 111.90 / 105.9 / **112.8** | 194.00 / 80.9 / 148.3 |

**Your 1.89x corroborates.** Forced 8 at one client costs **1.98x** serial's CPU per query, measured end to end through the engine on a different harness and corpus. Adaptive at one client costs 2.74x, buying 3.39x on p50 and 3.70x on throughput.

**The load gate demonstrably engages.** Adaptive's CPU per query falls 145.8 to 112.8 as clients rise and lands on serial's 112.0 at 16 clients (and 84.2 against 85.7 at 8). It stops splitting under load, which is exactly what the gate is for, and it is visible in CPU rather than inferred from latency.

**Overriding the gate costs the overrider, not only everyone else.** Forced 8 keeps splitting regardless of load, its CPU per query sits flat at ~148 from 2 clients on, and by 16 clients it delivers **0.76x serial's throughput with a p50 1.75x worse than serial**. On the open question of whether explicit `maxPartitions > 1` should yield under load, the data says the current sharp edge cuts both ways: an operator who forces it on a busy box makes their own queries slower too, not just their neighbours'. That is an argument for yielding, or at minimum for the warning you already added being loud.

**One genuine subtlety in forced 8's favour.** It trades median for predictability. At 16 clients its p99/p50 is 1.53x against serial's 5.21x (297 vs 194, against 579 vs 111). Every query gets the same 8 ranges, so per-query cost is uniform and the tail compresses even as the median degrades. If someone is chasing tail latency under load rather than median, that is not a nonsensical thing to want.

## The censoring, stated plainly

Serial throughput scales only **5.51x** from 1 to 16 clients, so my harness has a ceiling: concurrency is driven from Python threads, and while JPype releases the GIL for the engine call, marshalling and row counting do not. Cells at 8 and 16 clients are limited by that boundary, not by ArcadeDB. Linearity holds to 2 clients (2.03x), is 89% of linear at 4, and 34% at 16.

That matters for one conclusion in particular, so I want to be explicit rather than let it pass: **adaptive converging on serial at 8 and 16 clients could in principle be "both pinned at my ceiling" rather than "the gate engaged."** The reason I think it is the gate is the CPU column, which is CPU divided by completed queries and therefore not ceiling-bound. If adaptive were still splitting it would still burn about 2x serial's CPU per query no matter how few queries finished. It burns 1.0x. Forced 8, which bypasses the gate, stays at 1.3x to 2.7x throughout and plateaus at 81 qps, comfortably under serial's 107, so its loss is not a ceiling artefact either.

**What I cannot measure from here**: `maxPartitions` is JVM-global, so there is no way for me to run one forced client alongside adaptive ones and watch the adaptive queries stop splitting. Every arm above sets one value for all clients. The forced-8-under-load rows price the same trade from the other side, but the true mixed-tenant case needs a per-query knob that does not exist, and nothing here should be read as covering it.

The 8.84M serial-versus-adaptive arms are running now and I will post them when they land.

tae898:
Opened #5561 for the CI redness, since it was agreed it deserved its own issue rather than a footnote here.

One correction to what we concluded above, from pulling the per-run job data: it is two separate phenomena, not one. `ha-integration-tests`, `unit-tests` and `slow-unit-tests` do fail in a varying subset, which matches the flakiness read. But `integration-tests` failed in 6 of 6 sampled runs with the **same two classes every time**, `Issue4141SessionManagementIT` and `Issue4141BoltSessionManagementIT`, verified identical at both ends of a 23-hour window on different commits. That one is deterministic: a plain `POST /api/v1/command/graph` with no session returns 400. It should reproduce locally without CI.

Nothing there touches this branch. Detail and run links are on #5561.

tae898:
On the one open reviewer question, explicit `maxPartitions > 1` under load: the two reviews above lean opposite ways, so here is the measurement, from the concurrency arm I posted earlier in this thread.

**"Explicit means do not throttle me" stops being a trade and becomes a pure loss at about 4 concurrent queries.** 1M INT8, closed loop, 12 cores, 30 s per cell:

| clients | serial p50 / qps | forced 8 p50 / qps | forced 8 vs serial |
|---|---|---|---|
| 1 | 39.40 / 19.4 | 13.65 / 61.0 | **2.9x better p50**, 3.1x qps |
| 2 | 38.52 / 39.3 | 22.24 / 77.4 | 1.7x better p50, 2.0x qps |
| 4 | 43.64 / 69.3 | 45.98 / 81.0 | p50 roughly level (0.95x), qps still 1.17x |
| 8 | 66.89 / 93.5 | 96.00 / 80.8 | **p50 1.44x worse**, qps 0.86x |
| 16 | 111.06 / 106.7 | 194.00 / 80.9 | **p50 1.75x worse**, qps 0.76x |

The argument for the current semantics is that the operator knowingly bought latency with CPU. Past roughly 4 clients they are not buying anything: forced 8 is worse than serial **on latency and on throughput at the same time**. Its CPU per query stays pinned at ~148 ms from 2 clients on, because it never yields, while serial sits at 52 to 112. So the operator who set the knob to protect one query's latency ends up with that query slower than if they had left it alone.

That reframes the decision. It is not "let them trade throughput for latency", it is "let them make a setting that cannot help them above a handful of concurrent queries". Which argues for the gate applying to explicit values too, or for the third additive mode suggested above, rather than for keeping the current contract.

**Two honesty notes.** First, the crossover point is hardware and workload specific; 4 clients on 12 cores with these queries is not a universal number, and I would not put it in a doc. Second, my harness has a ceiling: serial throughput scales only 5.51x over 16x clients, so the 8- and 16-client cells are limited by our Python/JPype boundary rather than the engine. That does not rescue forced 8 though, because it plateaus at ~81 qps while serial reaches ~107, comfortably under the ceiling, so its throughput loss is real and not an artefact. The CPU column is also ceiling-independent, being CPU divided by completed queries.

What I still cannot measure is the actual multi-tenant case, one forced client alongside adaptive ones, because `maxPartitions` is JVM-global. If the concern is specifically an operator-forced query suppressing splitting for its neighbours, that needs a per-query knob to test properly and nothing above covers it.

tae898:
Correction on provenance, before anyone acts on the numbers I posted an hour ago.

**My wheel was built from this branch at 07:36Z today, so it predates eight commits, including `60b88fb6c` and `2c861677b`.** I wrote that comment without having seen `60b88fb6c`, and it argues for something you had already decided against two hours earlier, with better reasoning than mine. That is on me. This project has cost me one bad report already from benchmarking a wheel that did not contain the change (#5388), so I would rather flag it myself.

**Your fix targets the half my harness cannot see, and it is the half that matters.** The real harm in explicit splitting was never the overrider's own throughput, it was that an over-claim of up to four times the pool recorded parallelism the machine was not providing and suppressed every concurrent query for that query's whole duration. I said explicitly that I could not measure the mixed-tenant case because `maxPartitions` is JVM-global. You fixed exactly that case. And "a knob honoured only sometimes is worse than either extreme" is a better argument than my "the gate should apply to explicit values too", so consider that suggestion withdrawn.

**What I think still stands, and why.** My forced-8 arm had *every* client forced, so there were no adaptive queries present to be suppressed. Both before and after `60b88fb6c` an explicit split proceeds regardless of capacity, so the work submitted per query is the same 8 ranges either way, and the reserved counter that changed had nobody to mislead. On that reasoning the numbers should carry over unchanged:

| clients | serial p50 / qps | forced 8 p50 / qps |
|---|---|---|
| 4 | 43.64 / 69.3 | 45.98 / 81.0 |
| 8 | 66.89 / 93.5 | 96.00 / 80.8 |
| 16 | 111.06 / 106.7 | 194.00 / 80.9 |

Above roughly 4 concurrent queries a forced split makes **the overrider's own** latency worse than not splitting, not just its neighbours'. Your rationale covers the neighbours and the honour-the-knob principle; it does not cover the operator harming themselves. I am not proposing the gate again. But if the throttled WARNING is the mechanism that surfaces this, it may be worth firing on self-harm too, not only on contention, because at 16 clients the setting is doing the opposite of what whoever set it wanted.

**Flagging that as reasoning, not measurement.** I have not re-measured on the current head, and the argument above is mine rather than a number. Say the word and I will rebuild and re-run the concurrency arm on `4246344b9`; it is about 20 minutes of box time.

**One thing that is unaffected either way.** The 8.84M serial-versus-adaptive arms running now are single-stream, one query at a time, so `inFlight` is always 1 and none of the load-accounting commits since 07:36 can change the split decision. I will state the exact wheel commit when I post them regardless.

lvca:
On "pure loss past about 4 clients": your p50 and qps columns support that, but your own p99 figures do not.

From the numbers you posted earlier in this thread, at 16 clients:

| | p50 | p99 |
|---|---|---|
| serial | 111 ms | 579 ms |
| forced 8 | 194 ms | 297 ms |

Forced 8's tail is roughly half serial's at that load. So it is not dominated - it is a median-versus-tail trade that inverts as concurrency rises. Every query gets the same 8 ranges, so per-query cost stays uniform and the tail compresses even while the median degrades, which is exactly what someone chasing p99 under load is buying. You had already spotted this as the "one genuinely good point in forced 8's favour"; I think it survives the later framing.

That is the reading I have gone with, so the explicit setting keeps its semantics rather than yielding. What changed is what an operator is told: the warning and the setting now describe it as a median win while the box is quiet, a median *and* throughput loss once it is busy, and a tail win throughout - so "yes for tail-sensitive traffic, no for anything else". Quoted as absolute p99 rather than p99/p50, because the ratio reads as a worse tail when the tail is in fact lower. No crossover figure, per your point that 4-on-12-cores is not universal.

The CPU-per-query column is the most useful thing in that run, by the way. It is the first evidence that the gate actually engages rather than the latency merely looking as though it does.

tae898:
@lvca one ask, because `00b27f575` puts my numbers in front of operators and they were measured on a wheel that predates `60b88fb6c`.

**Would you re-run the forced-8-under-load arm on your own harness before that wording ships?** Not because I think the numbers are wrong, but because for this particular claim your harness is the better instrument and mine is measurably censored.

**Why mine is the weaker one here.** I drive concurrency from Python threads. JPype releases the GIL for the engine call, but argument marshalling and row counting do not, so serial throughput in my harness scales only **5.51x** across a 16x increase in clients. My 8- and 16-client cells are limited by that boundary, not by ArcadeDB. Yours is JVM-native with no such ceiling, and it already produced the 1.89x CPU figure and the 14% pool-activity regression, so it can see the region where mine goes blind.

What rescued my conclusion was the CPU column, which is CPU over completed queries and therefore not ceiling-bound, plus the fact that forced 8 plateaus at ~81 qps while serial reaches ~107, comfortably under the ceiling. That is enough to say the inversion is real. It is not enough to trust the *magnitudes* now in `GlobalConfiguration`, and 0.76x and 1.75x are magnitudes.

**The specific thing to check**, if you do: whether the inversion survives `60b88fb6c`. My reasoning says it should, because every client in my arm was forced, so no adaptive query was present for the old over-claim to suppress, and an explicit split proceeds regardless of capacity both before and after. Same 8 ranges submitted per query either way. But that is my reasoning about your code, which is exactly the kind of claim that should be measured rather than argued, and you can rebuild the branch in the time it takes me to fail at it twice (the recipe in our own `build.sh` header turns out to need a Maven image with `tar` in it).

**If the magnitudes move, please take yours over mine** and change the doc. I would rather the setting carry a number from the harness that can actually see 16 clients than one from mine with a footnote.

**What our side is genuinely better at, and is doing instead**: the 8.84M serial-versus-adaptive arms are running now on real Big-ANN SPLADE with exact ground truth, single stream, which is where our corpus and scale are the asset and where the GIL ceiling cannot bite because there is only ever one query in flight. Those I will stand behind, with the wheel commit stated.

Also worth folding into the same doc pass if you agree with it: the tail behaviour. p99 297 against 579 at 16 clients means forced splitting is a *tail* optimisation that costs median and throughput, which is a coherent thing for someone to want and is not obvious from the setting's name.

tae898:
Thanks for running it. **My tail claim was wrong and yours should replace it entirely.**

The interesting part is not that my magnitudes were soft, it is *which* number my harness broke, because it is the one I should have distrusted first and did not.

Your serial p99 at 16 clients is **62.8 ms**. Mine read **579 ms** for the same cell. That is a factor of nine, and it is all mine: 16 Python client threads on a 12-core cpuset, with marshalling serialised behind the GIL, so the serial arm spends most of its tail waiting on my harness rather than on the engine. Forced 8's tail was inflated too, just less catastrophically, and the ratio between two differently-inflated numbers came out backwards. That is how I got a tail *win* out of what your harness shows is a 2.5x tail *loss*.

I argued the CPU column rescued my conclusions because CPU over completed queries is not ceiling-bound. That was true, and I then over-extended it: it licences the inversion, which is a statement about total work, and it says nothing about a percentile. **The tail is the metric most sensitive to queueing in front of the engine, so it is exactly the one a saturating harness cannot report.** I had the evidence that my harness was saturating (5.51x over 16x clients, in the same comment) and still quoted a p99 comparison from inside the saturated region. That is the lesson worth me carrying, and it generalises past this PR: our numbers are trustworthy at concurrency 1, where nothing queues, and our percentiles should not be quoted above it.

So: no measure on which forcing wins above light concurrency, and the setting is for one-query-at-a-time work with nothing else running. Your wording, your numbers.

**The adaptive figures you recorded are the better headline anyway**, and I would not have obtained them: at 8 clients beating serial on throughput (275 vs 196), median (35.9 vs 37.5) *and* p99 (75.2 vs 95.9), then at 16 clients matching serial while splitting **2 queries out of 11,444**. That last figure is the gate working, stated directly. I inferred the same thing indirectly from CPU per query converging on serial's, which was the right inference from the wrong instrument. 2 out of 11,444 is what the claim should rest on.

Also good catch on the warning threshold. Firing only when the adaptive gate would have refused means ~10 queries in flight on an 18-thread pool, while the harm starts between 4 and 8, so the operator most likely to be hurt was the one least likely to be told. Keying the second trigger on forced ranges oversubscribing the pool rather than on a client count is the right shape, since it is the thing that is actually true on both our boxes despite the different core counts.

Nothing here changes the 8.84M serial-versus-adaptive arms still running on our side. Those are single stream, one query in flight, so the ceiling that broke my tail numbers cannot reach them, and real Big-ANN SPLADE at 8.8M documents with exact ground truth is the part our harness is actually the right instrument for. I will post them with the wheel commit named.

lvca:
Retracting my p99 argument above: I re-ran it here as you asked, and **your original "pure loss" conclusion was right. Mine was wrong.**

1M INT8, 18-thread pool, JVM-native clients, 30 s cells, on `bad4eaf28`:

| clients | serial p50 / p99 / qps | forced 8 p50 / p99 / qps | adaptive p50 / p99 / qps |
|---|---|---|---|
| 1 | 32.7 / 35.7 / 32 | 8.8 / 12.8 / 112 | 8.6 / 13.5 / 110 |
| 4 | 37.4 / 51.1 / 105 | 21.1 / 45.9 / 177 | 12.9 / 70.5 / 175 |
| 8 | 37.5 / 95.9 / 196 | 50.0 / 115.9 / 149 | **35.9 / 75.2 / 275** |
| 16 | 39.5 / **62.8** / 391 | 73.2 / **156.4** / 204 | 41.1 / 56.2 / 382 |

**The tail benefit does not reproduce.** At 16 clients forced 8's p99 is 156.4 ms against serial's 62.8, two and a half times worse. Your 297-against-579 was the one thing making forced splitting a coherent choice under load, and it does not survive a harness without the Python boundary. So there is no regime above light concurrency where the setting wins - not median, not throughput, not tail - and my "median-versus-tail trade" reading was wrong.

**You were also right that the magnitudes needed our instrument, and they moved against the setting**, not for it: 0.52x throughput rather than 0.76x, and a median 1.85x worse rather than 1.75x. Your 1.98x CPU per query corroborates the 1.89x measured here a different way, so the two harnesses agree on the cost and disagree only where yours is censored.

Both the WARNING and the setting's documentation now carry these numbers instead of yours, and say plainly that forcing is worth it only when one query at a time must be as fast as possible with nothing else running. The warning also fires earlier: its only trigger was "the adaptive gate would have refused this", which needs half the pool's worth of queries in flight, while the crossover you identified sits well below that. A second trigger keyed on the forced ranges oversubscribing the pool covers the region where the setting has already turned against whoever set it - which was your point, and it was a good one.

**On the semantics themselves we are leaving explicit `maxPartitions > 1` as it is**, honouring the knob rather than gating it. That now rests on one argument alone, that a setting honoured only sometimes is harder to reason about than one that always does what it says, and no longer on any performance case. The measurements are in the log line and the docs so nobody sets it by accident.

**One thing worth having from your run that ours does not show:** the adaptive default at 8 clients beat serial on all three of throughput (275 vs 196), median (35.9 vs 37.5) and p99 (75.2 vs 95.9), and at 16 clients matched serial while splitting 2 queries out of 11,444. The gate is not merely getting out of the way under load, it is finding a middle band where splitting still pays. That band is what your single-stream 8.84M arms cannot see and ours cannot corroborate at scale, so the two runs are complementary rather than redundant.

Thank you for flagging the wheel provenance and your own ceiling unprompted. Both times it was the thing that made the numbers usable.

tae898:
Our retractions crossed, so we agree on the tail: I withdrew that claim a few minutes before your comment landed, for the same reason you found it. Good outcome either way, and no need to relitigate it.

**But there is something in your table neither of us has remarked on, and it is about the default rather than the override.**

Adaptive's p99 against serial's, from your own numbers:

| clients | serial p99 | adaptive p99 | adaptive vs serial |
|---|---|---|---|
| 1 | 35.7 | 13.5 | 0.38x, much better |
| 4 | 51.1 | **70.5** | **1.38x worse** |
| 8 | 95.9 | 75.2 | 0.78x better |
| 16 | 62.8 | 56.2 | 0.89x better |

**Adaptive's tail is better than serial everywhere except at 4 clients, where it is worse than serial and worse than forced 8 (45.9).** That is non-monotonic, and it sits exactly in the middle band you just identified as where the gate is doing something clever rather than merely getting out of the way. At that point its median is excellent, 12.9 against serial's 37.4, so it is buying a 2.9x median win with a 1.38x tail loss. That is a real median-versus-tail trade. It just belongs to the default, not to the setting you were documenting.

**A mechanism that would explain it, offered as a hypothesis rather than a finding.** Four clients is presumably where the gate starts refusing *some* queries and not others. A refused query then runs unsplit while pool threads are still busy with other queries' ranges, so it is slower than the same query would be on an all-serial box, which is why adaptive's p99 exceeds serial's rather than merely matching it. Split and unsplit queries coexisting gives a bimodal distribution, and p99 reads the upper mode. Forced 8 has no such mixture (everything splits) and serial has none either (nothing does), which fits both of them having tidier tails at that point.

**Cheap ways to tell whether that is what it is**, if you think it is worth knowing: the split rate at 4 clients (you already instrument it, you quoted 2 of 11,444 at 16), and whether the 4-client latency histogram is bimodal rather than long-tailed. If the split rate at 4 is somewhere in the middle rather than near 0 or near 100%, that is the mixture.

**Why I think it matters more than the forced-split question you have been documenting.** Explicit `maxPartitions` is opt-in and now carries a warning; almost nobody will set it. The adaptive default is what every user gets. If the default has a tail cost in a specific concurrency band, that is worth a line somewhere even if the answer is "inherent to any gate, accepted, the median win is worth it". It may also be noise from a single 30 s cell, which is exactly why I am flagging it rather than asserting it.

I cannot check any of this myself: it lives in the region where my harness is censored, which is the whole reason you re-ran it.

On leaving explicit `maxPartitions > 1` ungated, resting on predictability alone and explicitly not on a performance case: agreed, and narrowing your own justification when the performance half evaporated is the right way to have done it.

lvca:
Ran it. **Your anomaly is real and your mechanism is right.** Documented in `dcf2aee0b`.

Three reps, 1M INT8, 18-thread pool, 30 s cells:

| 4 clients | p50 | p99 | split rate |
|---|---|---|---|
| serial | 26.2 / 25.6 / 26.8 | 30.8 / 27.3 / 31.7 | 0% |
| **adaptive** | **7.6 / 7.6 / 7.8** | **32.3 / 37.7 / 48.6** | 73-76% |
| forced 8 | 13.7 / 14.0 / 14.7 | 22.0 / 21.4 / 29.8 | 100% |

Adaptive's p99 exceeds serial's in all three reps, and forced 8 - uniform width - has the tidiest tail of the three. So it is not a single-cell artefact.

**The mixture is the cause, with one refinement.** It tracks the *spread* in how queries are treated rather than the split rate as such: at 2 clients every query splits and the tail is fine, at 4 some claim a wide split while others get none and the tail is worst, at 8 the grants are uniformly small and it is fine again. Maximum spread in the middle, which is why the band closes on both sides. Your bimodal-distribution reading is exactly what the numbers show; the variable is partition *width*, not split-versus-serial.

**I should also withdraw a counter-argument I nearly posted.** My first pass had adaptive's p99 at 2 clients worse than serial with a 100% split rate, which I read as refuting the mixture. That cell was measured on a loaded box. On a quiet one adaptive's tail at 2 clients is *better* than serial's, so the refutation was noise and your hypothesis never had a problem.

**Documented rather than fixed.** The trade at 4 clients is a 3.4x median win for about 1.2x the p99. Flattening it means giving up most of that median win to shave a tail still around 38 ms, which is the wrong way round - but you are right that a user should not have to discover it, so the setting now says the default can cost tail latency in a narrow concurrency band and quotes these numbers.

**And your ask exposed a gap in what I instrumented.** `queries.split` records whether a query split, not into how many ranges - so the metric cannot see the variable that explains this. The gauge that would is a distribution of granted partition counts. That is now written next to the metric so whoever chases it does not have to rediscover why the existing one is insufficient.

Worth noting what this run also showed about the middle band, since it is the case your single-stream arms cannot reach: at 8 clients the adaptive default beat serial on all three of throughput (275 vs 196), median (35.9 vs 37.5) and p99 (75.2 vs 95.9). The gate is not only getting out of the way under load, it finds a region where splitting still pays on every measure.

tae898:
**8.84M, as promised. It holds, and it is larger than at 1M: 3.58x.**

Full Big-ANN SPLADE corpus (8,841,823 docs), INT8, 1000 dev queries, cpuset 0-11, 16g heap, settled, single stream, N=3, median.

| | `maxPartitions=1` (serial) | `=0` (adaptive) | gain |
|---|---|---|---|
| **p50** | 301.52 ms | **84.14 ms** | **3.58x** |
| p95 | 994.36 | 238.07 | 4.18x |
| p99 | 1250.05 | 304.02 | 4.11x |
| QPS | 2.60 | 9.70 | 3.73x |
| recall@10 | 0.9914 | 0.9914 | identical |
| build | 936 s | 945 s | unchanged |

Per-rep p50 was [301.52, 299.69, 304.23] serial and [84.30, 82.65, 84.14] adaptive, so the spread is under 2% on both arms.

**Two things worth drawing out.**

**The gain grows with the work, within a tier as well as across tiers.** 3.58x at the median, 4.18x at p95, 4.11x at p99. The expensive queries benefit most, which is the same "pays in proportion to how much work the query does" property you described, showing up inside a single query mix rather than between 100k and 1M. It is also the useful direction: the queries that were painful are the ones that improve most, so the distribution tightens rather than just shifting.

**It is bigger here than at 1M** (3.58x against 3.44x), which closes the scaling question. At 8.84M essentially every query carries enough posting mass to clear the 200k gate and partition fully, whereas at 100k a good share sat under it and only got 1.69x. The three tiers now read 1.69x / 3.44x / 3.58x, monotonic in corpus size.

**What it does to the comparison at the tier where we were furthest behind.** Against Qdrant's 16.4 ms at 8.84M we go from **18.4x to 5.1x**. For the whole trajectory on this tier: 387 ms when we were on the pre-dev21 line, 293 ms on released dev21 after your three merged rounds, 84 ms with the split. Elasticsearch answers this tier in 10.8 ms but at 0.781 recall against our 0.9914, so that one is not a like-for-like comparison and I would not quote it as one.

**Provenance, since it caught me out once already today.** This wheel is from `013468d80`, the branch head at 07:36Z, so it predates the eight commits that landed afterwards including `60b88fb6c` and `bad4eaf28`. For this measurement that does not bite: it is single stream, one query in flight at all times, so the load gate never engages, no reservation is ever contended, and none of the load-accounting changes since can alter the split decision. Every arm above is `inFlight=1`. If you want it re-run on the merged code I am happy to, but I would expect the same numbers for that reason rather than because I checked.

Congratulations on the merge. Once this ships in a release we will re-measure on the released wheel and it moves from our prose into the paper's table, which is where I have wanted to put it since the 165 ms days.

Issue #5519, opened 2026-07-29
TimeSeries mutable rows reserve 258 bytes per STRING column: a 10-tag row fits 25 per page and pays 24x page+WAL amplification


lvca:
Follow-up to #5474. With the boxing gone from the ingest loop, the dominant cost on a realistic tag-heavy TimeSeries schema is not row encoding and not the shard commit machinery: it is the row **stride**.

## What

`TimeSeriesBucket.calculateRowSize()` reserves a fixed slot for every STRING column:

```java
final int fixed = col.getFixedSize();
if (fixed > 0)
  size += fixed;
else
  size += 2 + MAX_STRING_BYTES; // max STRING: 2-byte length prefix + max payload
```

`MAX_STRING_BYTES` is 256, so **every STRING column costs 258 bytes of stride regardless of its actual content**.

The writer does *not* write that wide: `writeStringValue()` returns `2 + bytes.length` and `appendSamples()` advances `colOffset` by exactly that, so the row is stored packed. Only the stride assumes the maximum, and the stride is what positions the row:

```java
final int rowOffset = DATA_ROWS_OFFSET + samplesInPage * rowSize;
```

So tiny rows are spread across a page that is almost entirely unwritten padding.

## Scale of it

TSBS cpu-only schema - 10 STRING tags + 3 DOUBLE fields:

```
rowSize = 8 (ts) + 10 * 258 (tags) + 3 * 8 (fields) = 2612 bytes
real payload per row                                 ~110 bytes
getMaxSamplesPerPage() = (65536 - 8 - 18) / 2612    = 25 rows per 64 KB page
```

**4% page fill.**

Measured on this host, 50,000 rows, 1 shard, embedded:

| arm | tags | stride | rows/page | pages | ms |
|---|---|---|---|---|---|
| TSBS-shaped, 10 tags of ~8 chars | 10 | 2612 | 25 | 2001 | **97** |
| identical schema, tag values 200 chars - 20x more bytes encoded, **same stride** | 10 | 2612 | 25 | 2001 | 85 |
| identical tag *content* joined into 1 STRING column - **stride is the only change** | 1 | 290 | 225 | 224 | **11** |

Encoding 20x more string bytes is free. Changing only the stride is **9x**.

Storage and WAL for that first arm:

```
payload      5.5 MB
pages        2001 x 64 KB = 131 MB     (23.8x amplification)
WAL          131 MB
```

[...]

Issue #5474, opened 2026-07-27
TimeSeries ingest boxes every numeric sample: appendSamples takes Object[] columns, then immediately unboxes to write raw bits


tae898:
Source-level finding on the native TimeSeries write path, from benchmarking TSBS against columnar engines.

**What**

The ingest API is typed on `Object[]` per column:

```java
public void appendSamples(final long[] timestamps, final Object[]... columnValues)
public void appendBatch(final long[] allTimestamps, final Object[][] allColumnValues)
```

and the value is consumed in `TimeSeriesBucket` as:

```java
page.writeLong(offset, Double.doubleToRawLongBits(value != null ? ((Number) value).doubleValue() : 0.0));
```

So a caller with primitive data must box every sample into a `Double`/`Long`, and the engine unboxes it a few frames later and writes the raw bits. The boxing is pure overhead by construction: no boxed value survives the call, and nothing in between needs an object.

**Scale of it on a normal workload**

TSBS cpu-only, 2.59M points x 3 DOUBLE fields = **7.8M `Double` allocations per ingest**, every one dead immediately, roughly 125 MB of garbage. Timestamps are already primitive (`long[]`), so the asymmetry is only on the value columns.

This is also visible from outside the engine: our Python bindings have to materialise boxed arrays purely to satisfy this signature, so the cost is paid again by every non-Java embedder.

**Suggested fix**

A primitive-typed overload alongside the existing one, for example per-column typed arrays:

```java
void appendSamples(long[] timestamps, double[]... numericColumns)
```

or a column-descriptor form that lets numeric columns be `double[]`/`long[]` and leaves `Object[]` for STRING/tag columns. The existing signature stays for mixed-type callers; the fast path stops boxing.

**Honest scoping**

I have not isolated what share of ingest time this is, and I am not going to claim one from the allocation count alone: JVM allocation is cheap and the write path also does page work. What I can say is the context, which is that TSBS ingest measures 411k points/s on ArcadeDB's native TS engine against 1.94M points/s for DuckDB on the same host, so the ingest path is the lane where ArcadeDB is furthest behind, and this is free garbage on exactly that path.

I can profile the ingest phase specifically (async-profiler, alloc + cpu) and post the breakdown if that would help decide whether the overload is worth it. Say the word and it goes on the benchmark host.

lvca:
Confirmed and fixed in `eb050f1aa` (main).

[...]

PR #5520, opened 2026-07-29
test(engine) #5519: isolate what a tag-heavy TimeSeries schema pays on ingest

lvca:
Backs the numbers quoted on #5519 and on #5474 with a benchmark that can be re-run.

## What it measures

`TimeSeriesBucket.calculateRowSize()` reserves `2 + MAX_STRING_BYTES` = 258 bytes of row stride for every STRING column, while `writeStringValue()` stores the row packed. Only the stride assumes the maximum, and the stride is what positions the row inside the page. On the TSBS cpu-only shape (10 STRING tags + 3 DOUBLE) that is a 2612-byte stride carrying ~110 bytes of data: **25 rows per 64 KB page, 4% fill.**

The benchmark separates that cost from the row encoding with three arms that each differ in exactly one variable:

| arm | what it changes | isolates |
|---|---|---|
| `WideStride` | the TSBS shape, 10 tags of ~8 chars | the baseline |
| `WideStrideBigValues` | same schema, 200-char tag values: 20x more string bytes encoded per row, **same stride** | encoding cost |
| `NarrowStride` | same tag text joined into one column: identical content, **one ninth the stride** | page traffic |

Output at the default size:

```
=== TimeSeries tag stride, 20,000 rows per batch, 3 measured batches per arm ===
arm                      stride   rows/page      pages      payload       page traffic       ms
WideStride                 2612          25        800       2.1 MB      50.0 MB (23x)     28.3
WideStrideBigValues        2612          25        800      39.1 MB       50.0 MB (1x)     32.2
NarrowStride                290         225         89       1.9 MB        5.6 MB (2x)      3.1
```

Encoding twenty times more string data costs nothing. Changing only the stride is 9x. A batch moves 23x more page and WAL bytes than it stores, which also costs the page cache and compaction.

Note the `WideStrideBigValues` amplification of 1x: 200-char tags nearly fill the 258 reserved bytes. The reservation is only waste when tag values are short, which is every real tag set.

## Notes on the method

- One full warmup pass over **all three** arms runs before any arm is measured, so no arm profits from the JIT state another left behind. Timing is `nanoTime`, median of 3 batches.
- Assertions are the two orderings, not absolute times, so this does not go flaky on other hardware. Observed margins are 3.7x and 9x against 2x bounds.
- Named `*Benchmark.java` and tagged `benchmark`, so `mvn-test.yml` excludes it two ways and `benchmark-tests.yml` picks it up.
- Default batch is 20k rows, not the 50k quoted on the issues: a wide-stride arm writes `rows / 25` pages of 64 KB, so 50k costs roughly 1 GB under `target/` across the three arms. `-Darcadedb.tagStrideBenchmark.rows=50000` reproduces the published batch size.

## Production change

One addition, `TimeSeriesBucket.getRowSize()`, so the stride can be reported exactly rather than derived from `getMaxSamplesPerPage()`. It sits next to that existing public accessor and changes no behaviour.

## Verification

- `mvn -pl engine test -Dtest=TimeSeriesTagStrideBenchmark -Dgroups=benchmark` passes
- all 460 `com.arcadedb.engine.timeseries.**` tests pass with the accessor added

When the tag dictionary proposed in #5519 lands, `WideStride` should converge towards `NarrowStride` and the assertions in this test are expected to change.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
@@ -777,6 +777,16 @@ public class TimeSeriesBucket extends PaginatedComponent {
   }
 
   /**
+   * Returns the fixed stride, in bytes, that one sample occupies in a data page. Note that a row is
+   * <em>written</em> packed: a STRING column stores its actual bytes but reserves
+   * {@code 2 + MAX_STRING_BYTES} in the stride, so on a tag-heavy schema the stride is much wider than
+   * the row content (issue #5519).
+   */
+  public int getRowSize() {
+    return rowSize;
+  }
+
+  /**
    * Returns the maximum number of samples that fit in one data page.
    */
   public int getMaxSamplesPerPage() {
--- a/engine/src/test/java/performance/TimeSeriesTagStrideBenchmark.java
+++ b/engine/src/test/java/performance/TimeSeriesTagStrideBenchmark.java
@@ -55,9 +55,9 @@ import static org.assertj.core.api.Assertions.assertThat;
  * Run explicitly with
  * {@code ./mvnw -pl engine -Dtest=TimeSeriesTagStrideBenchmark -Dgroups=benchmark test}.
  * Override the batch size with {@code -Darcadedb.tagStrideBenchmark.rows=50000} to reproduce the batch
- * size quoted on those issues. The default is deliberately smaller: a wide-stride arm writes
- * {@code rows / 25} pages of 64 KB per batch, so 50k rows costs roughly 1 GB under {@code target/}
- * across the three arms.
+ * size quoted on those issues. The default is deliberately smaller: at the default 64 KB page size a
+ * wide-stride arm writes {@code rows / 25} pages per batch, so 50k rows costs roughly 1 GB under
+ * {@code target/} across the three arms.
  *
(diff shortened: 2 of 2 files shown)


lvca:
Thanks - point 1 was a real defect and is fixed in e711b433b.

**1. Hardcoded page size.** You are right that the two sides could disagree. `PaginatedComponent.getPageSize()` was already public, so the constant is gone entirely: `pageTrafficBytesPerBatch()` now reads it from the bucket, and both `report()` and the amplification assertion go through that one method. Running with `-Darcadedb.bucketDefaultPageSize=...` now stays internally consistent instead of drifting. Also made the two remaining prose references say "at the default 64 KB page size" rather than stating it as a fact of the layout.

**2. Timing assertions.** Accepted as a known trade-off. Worth being explicit about which assertions carry which weight: the two `medianNanos()` orderings *are* the finding, so they have to be timing-based, while the three structural ones (`rowsPerPage`, amplification) are the part that can never flake. If this does turn out to flake on the benchmark runner, the right fix is to drop the timing pair rather than widen the bounds, since a widened bound would stop testing anything.

**3. `System.out.printf`.** Agreed, that is the deliverable here.

**4. Disk footprint.** Your ~600 MB at the 20k default is right. The javadoc only quotes the 1 GB figure for the 50k override, so nothing to correct there.

Re-ran after the change: 29.9 / 25.5 / 4.3 ms across the three arms, ratios unchanged.

PR #5521, opened 2026-07-29
fix(engine) #5516: LSM_VECTOR in-memory bloat, and a compaction that could be trusted with the leftovers

### LINKED ISSUE #5516: LSM_VECTOR index in-memory bloat: VectorLocation objects never removed from ConcurrentHashMap (OOM after weeks of operation)
## Summary

The `VectorLocationIndex` maintains an in-memory `ConcurrentHashMap` that accumulates `VectorLocation` entries on every `put()` call. Stale entries (from re-embedding unchanged or updated vertices) are **never removed from memory** — not by `DROP INDEX`, not by index recreation, and not by GC. They persist until the database is closed and reopened.

In production, this caused **9.3 million `VectorLocation` objects** (~970MB heap) for only **~4,029 active vectors**, leading to `OutOfMemoryError` on a 1500MB heap.

## Environment

- ArcadeDB: 26.5.1
- JVM: OpenJDK 21.0.5
- OS: macOS / Linux
- Index config: `LSM_VECTOR, 384 dimensions, COSINE, INT8 quantization`

### Steps

1. Create a vertex type with `embedding` property + `LSM_VECTOR` index
2. Store 1,000 embeddings
3. Re-store the same 1,000 embeddings 10 times (simulating periodic sync cycles)
4. Heap histogram shows ~10,000 `VectorLocation` objects
5. `DROP INDEX` + `CREATE INDEX` — **stale entries still in memory**
6. `db.close()` + `db.open()` — **memory cleared**

### Expected behavior

After `DROP INDEX`, the in-memory `ConcurrentHashMap` entries for that index should be cleared. Alternatively, ArcadeDB should compact/remove stale entries automatically (standard LSM-tree compaction).

### Actual behavior

Stale entries persist in the `ConcurrentHashMap` until the database is closed and reopened. `DROP INDEX` only removes the disk files; the in-memory map is untouched.

## Heap histogram evidence (production)

```
jcmd <PID> GC.class_histogram

 num     #instances         #bytes  class name
   1:       9,353,958      299,326,656  java.util.concurrent.ConcurrentHashMap$Node
   2:       9,310,795      297,945,440  com.arcadedb.index.vector.VectorLocationIndex$VectorLocation
   4:       9,308,069      223,393,656  com.arcadedb.database.RID
   5:       9,316,716      149,067,456  java.lang.Integer
```

Total: ~970MB for 9.3M objects. Active vectors: ~4,029. Bloat ratio: ~2,300x.

## Impact

- **Memory**: ~970MB of unreachable-but-pinned heap for a 4,000-vector dataset
- **GC pressure**: G1 old gen fills rapidly, causing sustained UNHEALTHY status (90%+ heap)
- **OOM**: After weeks of 10-minute sync cycles, heap exhausts and the JVM crashes
- **Customer impact**: Desktop app on 4GB VMs crashes after ~2-3 weeks of continuous operation

## Workaround

Close and reopen the database after rebuilding vector indexes at startup:

```kotlin
db.command("sql", "DROP INDEX ${typeName}_embedding_idx")
db.command("sql", "CREATE INDEX ${typeName}_embedding_idx ON ${typeName} (embedding) LSM_VECTOR METADATA { ... }")
db.close()
db = DatabaseFactory(dbPath).open()  // Fresh in-memory state
```

## Suggested fixes

1. **`DROP INDEX` should clear the in-memory index map** — currently only removes disk files
2. **Automatic compaction for LSM_VECTOR** — standard LSM-tree compaction would merge segments and remove tombstoned entries
3. **`COMPACT INDEX` SQL command** — allow explicit compaction of vector indexes
4. **Don't load tombstoned entries on startup** — when opening the database, skip stale entries when rebuilding the in-memory map

## Related

- We observed this with `all-MiniLM-L6-v2` 384-dimensional embeddings with INT8 quantization
- The `count(*)` query on `index:<name>` returns only active entries (4,029), not the in-memory total (9.3M) — making the bloat invisible to SQL-level monitoring
- Related issues: #2735 (HNSW metadata persists after DROP INDEX), #3135 (stale ordinal mappings), #3722 (missing vectors after restart), #3868 (OOM during graph rebuild)

lvca:
Fixes #5516.

Two commits, separable if you prefer them split.

## 1. The reported bug: memory grows with writes, not with live vectors

`VectorLocationIndex` kept a superseded vector id resident - its `VectorLocation`, its RID, the map node and its slot in the reverse index, ~90 bytes - only to carry a `deleted` boolean. Re-embedding 4K vectors every few minutes reached **9.3M locations, ~970MB of heap, then OOM**; opening the database pulled every tombstoned page entry back in, and only a full graph rebuild ever cleared the map.

A tombstoned id is now dropped from the map and remembered as **one bit** in a compact growable set. `getLocation()` answers `null` for it, which every reader already treats as deleted, and `isDeleted()` tells it apart from an id merely evicted from a bounded cache (which must still be reconstructed from the pages). That short-circuit also stops the reconstruction path from resurrecting a deleted vector, since a page scan returns the id's *first* (live) entry, not its later tombstone.

While measuring it, a second cause showed up: the commit indexed every updated record **twice** - once while serializing the record, once replaying the operations queued on the transaction. An LSM key/value entry re-applied twice collapses onto the same key; a vector `put` mints a new id every time, so each update left two vectors and two tombstones instead of one, doubling the growth of both the index file and the location index. The vector index now queues during the record-serialization step too, unless the transaction will not replay its queue (a replica).

Measured on 100 vertices re-embedded 10 times, with graph rebuilds neutralized: **2100 resident locations -> 100**, and reopening loads 100 instead of 2100.

## 2. The follow-up: `COMPACT INDEX` was not safe to run

Reclaiming the disk the same workload accumulates meant `COMPACT INDEX`, so it got checked. On 500 vectors x 384d INT8 after 10 update cycles it was broken four ways:

| Defect | Before | After |
|---|---|---|
| Tombstone lost the dedup tie (`id > existing.id`, but a tombstone carries the *same* id) | 400 live -> **500** | 400 |
| Compacted component never re-read, page numbering restarted at 0 | 500 -> **1000** | 500 |
| `splitIndex` never dropped the file it replaced | 2.36MB -> **2.62MB** | 9 pages/2304KB -> **1 page/256KB** |
| Orphan-relink pass never applied the schema definition, so `dimensions=0` | **0 vectors after restart** | 500 |

The first defect also lived in the graph rebuild's own page dedup, where it resurrected deleted vectors with no compaction involved. The last one is why compaction was unrecoverable: an LSM-tree index keeps its key types in its own file header and survives a rename, while a vector index keeps its whole definition in `schema.json`.

Compaction is now the rebuild plus a file rewrite. The rebuild already reads every page, resolves the LSM merge and produces the live set - that set *is* the compacted content, so the entries are copied verbatim (quantized payloads never re-quantized, ids never reused), swapped in, and both the old data file and any legacy compacted component are dropped. One pass, one notion of "live", one data file at the end. `LSMVectorIndexCompactor`, the buggy `splitIndex` and the compacted-append path are deleted with it.

Automatic compaction stays disabled: `COMPACT INDEX` remains an explicit operation.

## Tests

New: `LSMVectorIndexTombstoneMemoryTest` (4), `LSMVectorIndexCompactionTest` (3). Updated `VectorLocationIndexRidReverseTest` and `VectorLocationIndexConcurrencyTest` to the new contract.

Green: full engine suite **10133 tests** on commit 1; on commit 2, **826** index+schema tests, **212** vector tests, the new tests, and `ACIDTransactionTest`.

One gap worth knowing: the full suite would not complete locally on the second commit because `ACIDTransactionTest.asyncIOExceptionAfterWALIsWrittenManyRecords` livelocks in `DatabaseAsyncExecutorImpl.scheduleTask` when it shares a machine with the benchmark classes (jstack: main sleeping, ~0.1s CPU per 30s). It passes standalone every time, and the changed code is unreachable from it - no vector index, no renamed index files. Looks like a pre-existing load-sensitive flake, but CI should confirm.

## Still open

The index file grows with every write for as long as the database lives, and compaction (now correct) is the only thing that reclaims it. Whether to re-enable the automatic trigger is a separate call.


lvca:
Thanks - item 1 was a real bug and is fixed in c10bba61, along with 2 and 5. Details, including where I pushed back:

**1. DeletedIds visibility - fixed, and you had the failure mode exactly right.**
`current[word] |= mask` is a plain array store and the reader's volatile load of `bits` only synchronizes-with a write to that field, which happened on resize only. Both `add()` and `remove()` now republish `bits` after the mutation, so the reader's acquire load sees the bit. I kept your reasoning in the Javadoc rather than just the fix, because the consequence is the non-obvious part: a missed bit is a stale "not deleted" that routes into `reconstructLocationFromPages()`, which returns the id's first (pre-tombstone) page entry and resurrects the vector. Worth noting the window is narrower than it looks - `markDeleted` runs under the index write lock and searches take the read lock - but `getVectorLocation()` is reachable without it, and the contract should hold on its own.

**2. Footprint claim - corrected, with one amendment.**
The Javadoc now says what is true: the ~90-byte locations follow the live vectors, the bit array follows the highest id handed out. One correction though - the array does not stay at its high-water size forever. `vectorIndex.clear()` in the rebuild calls `deletedIds.clear()`, which drops the array back to `EMPTY_BITS`, so every graph rebuild (and therefore every compaction) releases it and it only regrows across the writes between two rebuilds. That is now stated explicitly.

**3. Replica path - added a test, and the behaviour is unchanged by this PR.**
Before this branch: `if (txStatus == BEGUN) queue; else apply`. On a replica at COMMIT_1ST_PHASE that means apply directly. After: `BEGUN || (COMMIT_1ST_PHASE && isIndexChangesReplayed())`, and `isIndexChangesReplayed()` is false on a replica, so it applies directly. Identical - the guard exists to keep that path exactly as it was, so if the direct apply doubled up against the replicated pages it would already do so on main.

There was no HA coverage for LSM_VECTOR to begin with (nothing in `ha-raft/src/test` touches it), so I could not close that gap here, but I added `anUpdateStillLandsWhenTheTransactionDoesNotReplayItsIndexQueue`, which drives a real commit with `commit1stPhase(false)` and asserts the update mints exactly one vector id. I checked it earns its keep: with the `isIndexChangesReplayed()` condition removed it fails `expected: 2 but was: 1` - the update gets queued and dropped. That is the regression someone "simplifying" the guard would otherwise ship silently. A true multi-node test still belongs in the HA suite.

**4. Dead-ish surface - leaving as is, with one correction.**
`VectorLocation.deleted` is not dead: `reconstructLocationFromPages()` still returns a location with `deleted=true` read from the pages, and callers act on it. What is now always-true is the `!loc.deleted` filter over *resident* entries in `getActiveVectorIds()`/`getActiveCount()`. I would rather keep those as cheap assertions of the invariant than remove them and rely on it being remembered - happy to strip them in a follow-up if you prefer.

**5. Silent no-op - now visible.**
The skip is reported twice: `COMPACT INDEX` returns `compacted: 0`, and the log line is upgraded from INFO to WARNING with the reason and what to do about it. Reachability: it needs the document-scan or in-memory recovery fallback to have contributed entries that are not on any page, so it is a recovery-state condition rather than a routine one.

**Minor:** `LSMVectorIndexTombstoneMemoryTest` runs 5 tests in 0.68s total, so I left it untagged. And agreed on letting CI confirm the full reactor - that is the one thing I could not do locally.

Vector suite after the changes: 213 tests, 0 failures.

lvca:
Thanks - and your instinct on the replica side was right. I ran the HA suite you asked for and it caught a real regression from this branch. Fixed in d0099ea9; the rest of the points are addressed in the same push.

**The one that mattered: leader/follower divergence after a vector compaction.**
`RaftIndexCompactionReplicationIT.lsmVectorCompactionReplication` passes at the branch point and failed here - because compaction now actually swaps the data file, where before it was a no-op. `checkDatabasesAreIdentical` caught it, and the cause was three layers deep, each hidden behind the previous one:

1. `LocalDocumentType.toJSON` keys its index entries by `getMostRecentFileName()`, and the vector index answered the name it was *created* with instead of its current component (`LSMTreeIndex` answers `mutable.getName()`). The leader's schema pointed at a file it had just dropped.
2. `toJSON` also serialized the creation name in the informational `indexName` field. `LSMTreeIndex` writes no name at all; this now writes the current component. Nothing reads it back - I checked.
3. `TypeIndex.equals` matches bucket indexes by `getName()`, so the index object itself has to follow its component. `indexName` is no longer final and moves with the swap.

The rule underneath: a vector index is named after its component, and every node names it after the file it holds - a follower rebuilds the index from the file the leader shipped. So a leader that renames its component without following the rename describes an index no other node has. `lsmVectorCompactionReplication` is green now, and it is the test that will catch this if it ever regresses.

Two failures I chased and dismissed with evidence, rather than leaving them as "probably flaky": `LSMVectorIndexSearcherPoolTest.concurrentSearchesNeverShareASearcher` (asserts pooled searchers <= 4, saw 5) and `compactionDefersWhenRecordingSessionActive` (no Raft leader elected in time) each failed once under load and pass on their own; neither touches this code.

**Your other points:**
- *Shift masking* - one-line comment added: `1L << id` shifts by `id & 63` (JLS 15.19), the bit within the word `id >>> 6` selects.
- *Stale asyncFlush comment* - was wrong, now corrected. The rewrite calls `writePages(..., false)`, i.e. synchronous, so `waitAllPagesOfDatabaseAreFlushed` is not what makes those pages durable; it stays as the guard that the recording session ships nothing still pending from the rest of the rebuild. The comment now says that.
- *Oversized entry* - now an explicit check that throws `IndexException` naming the vector id and both sizes, instead of an invariant a future page-size change could turn into silent corruption. The comment says why it cannot fire today (every entry came off a page of the same size).
- *Dead code* - `getSubIndex`/`setSubIndex` and `LSMVectorIndexCompacted.createNewPage` removed. `getHeaderSize` is NOT dead: `getAllVectors()` uses it. I kept both, since `getAllVectors` is the only typed reader for a legacy compacted component and databases compacted by an older build still carry one - that is now stated in its Javadoc.
- *`getActiveCount()` filter* - kept, and documented why: no resident location carries `deleted == true` any more, but a location read straight from a page can, and the filter is what defines "active". Happy to drop it if you would rather.

Engine vector suite: 213 tests, 0 failures. HA: `lsmVectorCompactionReplication` + `compactionDefersWhenRecordingSessionActive` green.

lvca:
@claude please review this PR, latest 2 commits

lvca:
All three points addressed in 30e7b867.

**`deletedVectors` stat** - fixed. It was `size() - getActiveCount()`, which is structurally 0 under the new contract, so an index full of deletions reported none. It now asks `getDeletedCount()`.

**Coverage gap on the identity/rename bug** - closed, and you were right that the engine suite could not re-detect it. It can now: `LSMVectorIndexCompactionTest#theIndexIdentityFollowsTheCompactedFile` asserts that after `COMPACT INDEX` the index's own name, `getMostRecentFileName()` (what the schema keys its entry by) and the serialized `indexName` are all the swapped-in component, and that the component exists on disk. I checked it earns its keep: reverting the one-line `indexName = newMutable.getName()` fails it with `expected "Doc_0_1198539685495208" but was "Doc_0_1198539227472041"`. The HA test stays the end-to-end guard; this one means a cluster is no longer required to notice.

**`add()` republishing on a no-op** - now symmetric with `remove()`: it publishes only when it actually set a bit. Worth noting why that is still correct for the resize path - a grown array's new words are zero, so a resize always falls into the set-the-bit branch and the new array is published with it.

**On the CI red.** I did not want to leave "unrelated" as an assertion, so I ran it both ways: `Issue4141SessionManagementIT` fails identically on this branch and on pristine `main` at `621313ce5` (4 run, 2 errors, HTTP 400 on `/api/v1/command/graph`). Pre-existing, and the same for the Bolt variant in that job. Everything else on the run is green - including `ha-integration-tests` (53 min), which is what exercises the compacted-file replication this PR changed, and `unit-tests` + `slow-unit-tests`, which also settles the `ACIDTransactionTest` livelock I flagged earlier as a local load artifact rather than a real hang.

Engine vector suite after this push: 214 tests, 0 failures.

One thing found while verifying, filed separately rather than folded in here: #5559 - setting `locationCacheSize` (documented as the recommended setting for 1M+ vector datasets) does not bound a cache, it truncates the index. 200 vectors inserted with `locationCacheSize: 10` gives `countEntries() == 10` and searches that return the wrong neighbours. Reproduced on pristine `main`, so it predates this PR, but it is adjacent enough to be worth knowing about while this area is fresh. #5558 is the other one - a search whose query vector sits in a deleted region returns an empty list.

Issue #5470, opened 2026-07-27
GraphBatch: Stuck without error after index compacting


danieljuhl:
`ArcadeDB Server v26.8.1-SNAPSHOT (build 7efe66c71bae9812983644a566a3ddeae2be3c97/1785167674525/main)`

During a GraphBatch loading of 16 million vertices and 15 millions edges, we have seen the following log on the **leader**, and no errors in the log nor a failed request in the client, but the creating of vertices stops after 630.000.

```
2026-07-27 18:13:45.152 INFO  [LSMTreeIndex] Full compaction of index 'Address_3_4324802883312027' (series=10 mutablePagesToCompact=12 threadId=125)...
2026-07-27 18:17:00.788 INFO  [LSMTreeIndex] Index 'Address_3_4324802883312027' fully compacted in 195637ms (keys=904586 values=2147153 droppedKeys=0 droppedValues=0 mergedSeries=10 newFile=Address_3_4324851379577400(460) threadId=125)
2026-07-27 18:17:01.044 INFO  [RaftTransactionBroker] Schema change for database 'graph-260715' does not fit one Raft entry (45031799 bytes > 33554432): shipping it as 3 ordered entries
2026-07-27 18:18:01.998 WARNI [RaftGroupCommitter] <arcadedb-1> Replicated entry size 26927206 bytes is approaching the maximum replicated Raft entry size of 33554432 bytes (>80%). Consider reducing the batch size, or raise arcadedb.ha.appendBufferSize (and arcadedb.ha.writeBufferSize with it). Large entries also raise leader heartbeat latency and risk election churn under load.
2026-07-27 18:18:12.081 INFO  [GraphBatch] <arcadedb-1> GraphBatch closed: vertices=630000 edges=0 flushes=0 avgFlushMs=0.0
```

lvca:
Fixed in main (24813519b).

**What happened:** this is not a GraphBatch problem, the HTTP upload was cut. The `/api/v1/batch` endpoint reads your file *while* it loads it, and the embedded web server closes a connection when nothing is read from it for `arcadedb.network.socketTimeout` (30 seconds by default). That timer cannot tell "the client stopped sending" from "the server is busy": during the full index compaction and the replication of the 45MB schema entry the load thread was inside a commit for about a minute, so the connection was closed and the upload ended at 630,000 vertices (exactly 63 x the 10,000-vertex commit batch).

**What changed:**

[...]

PR #5522, opened 2026-07-29
#5517 A bloom filter per compacted LSM series, so a lookup skips the series that cannot hold the key

### LINKED ISSUE #5517: LSM index: a bloom filter per compacted series, so a lookup skips the series that cannot hold the key
## Why

A point lookup in an LSM index walks the compacted series newest-to-oldest (`LSMTreeIndexCompacted.searchInCompactedIndex`). Within a series the root page already directs the search to a single data page, so that part is efficient — but **every series that does not hold the key still costs a root-page binary search plus a data-page read**, and the key lives in exactly one of them.

The workload where this hurts most is not queries, it is **bulk loading with a unique index**: a duplicate check misses in *every* series by definition, on every inserted record. In the load discussed in #5470 the compaction log showed `mergedSeries=10` with two unique indexes, i.e. up to twenty directed page lookups per vertex, all of them misses, on the critical path of a 16M-vertex import.

## What already exists

`com.arcadedb.engine.BufferBloomFilter` — written years ago, correct, and referenced by nothing but its tests. It was hardened in #4960 and has just been given a measured test suite, a configurable probe count and sizing helpers:

| target | bits/key | probes | measured FP | FP at the old k=2 |
|---|---|---|---|---|
| 5% | 6.24 | 4 | 5.00% | 7.48% |
| 1% | 9.59 | 7 | 1.01% | 3.57% |
| 0.1% | 14.38 | 10 | 0.10% | 1.65% |

Cost to use: **1.20 bytes per key** at a 1% false-positive rate, flat at any scale (1M keys = 1.2MB), and ~21ns per probe against microseconds for the page lookup it replaces.

## Design

**A paginated component, not a side file.** A plain file would not be replicated over HA, would not be transactional and would not appear in snapshots or `DROP INDEX`. So: `LSMTreeIndexBloomFilter extends PaginatedComponent`, extension `.bfidx`, registered in `LocalSchema` alongside the others, modelled on `SparseSegmentComponent`. Pages then flow through PageManager → WAL → Raft for free.

- **Layout** — page 0 is a directory with one entry per series `{seriesRootPage, firstFilterPage, filterPages, slots, probes, seed}`; the remaining pages hold the bits, sharded by high hash bits so a probe touches exactly one page whatever the series size, wrapped directly in a `BufferBloomFilter` over the page `Binary`.
- **Build** — inside `LSMTreeIndexCompactedStreamWriter.appendBounded`, the single point every surviving key of a series passes through, published in `finishSeries`. Written-to-page and added-to-filter become the same line of code, so they cannot diverge.
- **Sizing** — a filter needs its size up front while the key count is only known at the end, but `appendBounded` already bounds a series by `DEFAULT_MAX_DATA_PAGES_PER_SERIES`, so the maximum keys per series is structural. Over-sizing only makes the rate better than target.
- **Read** — probe before the root-page search in `searchInCompactedIndex`; skip the series on \"absent\". **Point lookups only**: a filter says nothing about ranges, so cursors and range scans must bypass it.

## Backward compatibility

No format version bump and no rollout gate. `ComponentFactory.createComponent` returns `null` for an unregistered extension and `LocalSchema` skips it (`if (pf != null)`), so an older build opens a database containing `.bfidx` files normally and reads every index exactly as today. The presence of the component *is* the version flag. An old follower even keeps the file in sync as raw replicated pages, so it is ready the moment that node is upgraded.

## The one rule that must not be broken

A false positive costs a wasted lookup. A **false negative hides data**: the lookup is skipped, an existing record is reported missing, and a unique-index duplicate check lets a duplicate in — silently, with the index file intact and `CHECK DATABASE` seeing nothing wrong. Gate: after every `finishSeries`, exhaustively assert that every key just written probes true.

## Open questions

- whether a compacted file accumulates series across compactions or is replaced, which decides if the directory is ever updated in place
- non-unique indexes (one key, many RIDs) — the filter is on keys so it should be unaffected, to be pinned by a test
- first cut behind a config flag, to measure a real bulk load with it on and off

Related: #5470, #4960.

lvca:
Closes #5517.

## The problem

A point lookup walks every compacted series newest to oldest. The root page already rules out a key outside the series' key **range** — `resultInRootPage.outside` — but it says nothing about a key *inside* the range that the series does not hold.

That is exactly a bulk load. Every series spans nearly the whole key space, so the unique-index duplicate check reads one root page **and one data page per series, per record**, and finds nothing. @danieljuhl's log in #5470 showed `mergedSeries=10` across two unique indexes: ~20 lookups per vertex, all misses.

## The fix

A new `.bfidx` component holds one bloom filter per compacted series. `searchInCompactedIndex` probes it before touching a series at all.

At the default 1% target that costs **~1.2 bytes per key** and ~21ns per probe. The win is cache residency rather than instructions: the filter is more than an order of magnitude smaller than the data pages it stands in for, so it stays resident when the index itself does not.

Point lookups only — a filter says nothing about a range, so cursors and range scans bypass it entirely.

## The rule that must not break

A false positive costs the page read that would have happened anyway. A false **negative** hides a row: the lookup is skipped, the record is reported missing, a unique index lets a duplicate through — silently, with the index file intact and nothing for `CHECK DATABASE` to see.

Everything here is biased that way. A series with no entry, an entry whose shape does not match the series the reader is looking at, a partial or null key, an unreadable page — all answer *"search this series"*. And before any filter is published, **every key just written is probed back exhaustively**; a single miss abandons the filter instead of publishing it.

## Two decisions worth calling out

**It is a `PaginatedComponent`, not a side file.** A plain file next to the index would not be replicated over HA, would not be transactional, would not be in a backup and would not follow the index through `DROP INDEX`. As a component its pages flow through the PageManager exactly like the compacted pages they describe.

**The file is strictly append-only.** A publish appends its bit pages and then one directory page, so the live directory is always the last page of the file. This came out of reading `RaftReplicatedDatabase.runWithCompactionReplication`: HA ships a compaction as `addFiles` plus, for files that already existed, only the page range they *grew* by (`pagesGrownDuringSession`, the #5443 fix). A directory rewritten in place would simply never travel, leaving every follower with a directory that disagreed with its own bits. It also makes a crash mid-publish harmless — the load walks back past unnamed pages to the last complete directory, gated on magic + version + count + checksum.

## Layout

Bits are split into independent **blocks of one page each** and a key is routed to exactly one block, so a probe reads **one** page however large the series is. A 64 KB page holds ~500k bits (~54k keys at 1%), so the occupancy skew of a blocked filter is negligible.

A directory entry is `{seriesRootPage, seriesPages, firstFilterPage, filterPages, slotsPerBlock, probes, keys}`. `seriesPages` is checked against the live series at probe time, so a filter built for a *different* series that once lived at that root page — an aborted round whose pages were later reused — is disowned.

## Back compatibility

No version bump and no rollout gate. An older build has no `bfidx` in `SUPPORTED_FILE_EXT`, so the directory scan never opens the file and the index reads exactly as it does today. The presence of the component *is* the flag.

## Configuration

`arcadedb.indexBloomFilterRate`, default **0.01 (on)**. `0` disables both writing new filters and consulting the ones already on disk, so a real load can be measured with it on and off. `getBloomSkippedSeries()` / `getBloomProbedSeries()` (also in `getStats()`) report what it is actually worth.

## Testing

11,529 tests, 0 failures, 0 errors — run in chunks because the whole engine suite exceeds the 10-minute cap on a single command:

| Scope | Result |
|---|---|
| `com.arcadedb.index.**` + `engine.**` | 527, 0 failures |
| `database` / `graph` / `schema` / `serializer` / `utility` / `function` / `security` | 566, 0 failures |
| `com.arcadedb.query.**` | 10091, 0 failures |
| root + `compression` / `event` / `exception` / `log` / `partitioning` / `performance` | 330, 0 failures |
| integration backup + restore | 15, 0 failures |

New: `Issue5517BloomFilterTest` (7, through a real compaction) and `LSMTreeIndexBloomFilterTest` (9, the component directly — the integration tests use 256 KB pages, which hold any test-sized series in a single block, so the multi-block routing needed its own tests).

**Mutation-tested.** Two real defects were injected and both were caught: a write/read routing mismatch, and dropping the `seriesPages` shape check. A third — routing on raw hash bits instead of the splitmix finaliser — was **not** caught, so the mix is documented as a precaution rather than a fix for anything demonstrated.

## Known residual

If a compaction round fails, *and* the rollback's own one-page directory append also fails, *and* the process dies before the next publish, a stale entry survives on disk. It still only bites if the next round writes a series at the same root page with the same page count. A running process is never exposed: the in-RAM filters are dropped before the write is attempted. Documented at the code.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1,0 +1,49 @@
+# ArcadeDB v.26.8.1 Release Highlights
+
+This is a living document: fixes, improvements, new features, and breaking changes are collected here as
+they land during the 26.8.1 development cycle, so the release notes are ready at tag time.
+
+### New Features
+
+#### Bloom filters on compacted LSM indexes (enabled by default)
+
+An LSM index lookup walks every compacted series from newest to oldest. A series' root page already rules out
+a key outside its key *range*, but not a key inside the range that the series simply does not hold - so that
+series still costs a root-page search and a data-page read to discover nothing. Each compacted series now
+carries a bloom filter that answers the question from a single 8 KB page.
+
+The workload this is for is a bulk load into a **unique** index, where the duplicate check for every incoming
+record misses in *every* series by definition
+([#5517](https://github.com/ArcadeData/arcadedb/issues/5517), out of
+[#5470](https://github.com/ArcadeData/arcadedb/issues/5470)).
+
+Measured on 2M keys across 9 series (`LSMTreeBloomFilterBenchmark`, run with `-Dgroups=benchmark`):
+
+| | filters off | filters on | |
+|---|---|---|---|
+| absent-key lookups (a duplicate check) | 199,743/s | 386,138/s | **1.9x** |
+| pages read for those lookups | 1,490 | 705 | **2.1x fewer** |
+| bytes read for those lookups | 372 MB | 176 MB | |
+| present-key lookups | 244,000/s | 281,150/s | 1.2x |
+
+- **On by default** at a 1% target false-positive rate: `arcadedb.indexBloomFilterRate=0.01`. Set it to `0`
+  to switch the feature off entirely.
+- **Costs about 1.2 bytes per key** on disk - roughly 3% of the index it describes - in a separate
+  `<index>_bf.bfidx` component next to the compacted index. A false positive only costs the page read that
+  would have happened anyway.
+- **No rebuild and no migration.** Filters are written by compaction, so an existing index gains them at its
+  next compaction. They replicate over HA and are included in backups like any other component, and are
+  dropped with the index.
+- **Helps most when compacted series overlap in key range**, which is what keys that do not arrive already
+  sorted produce (an email, a UUID, a business id). Keys inserted in ascending order give each series a
+  disjoint slice that its root page already rules out, leaving the filters little to save. Range scans and
+  cursors never consult them: a filter can answer for one key, never for an interval.
+- **Observability:** `bloomSkippedSeries` and `bloomProbedSeries` in the index statistics report how many
+  series the filters spared and how many were still read.
+
+Backward and forward compatible with no version bump and no rollout gate: an older ArcadeDB does not
+recognise the `.bfidx` extension, never opens the file, and reads the index exactly as it does today. Note
+that a downgraded build compacts *without* maintaining the filters, so if you downgrade and later come back,
+prefer running a full compaction (or setting `arcadedb.indexBloomFilterRate=0`) on the upgraded build.
+
+**Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -568,14 +568,37 @@ public enum GlobalConfiguration {
 
   INDEX_BLOOM_FILTER_RATE("arcadedb.indexBloomFilterRate", SCOPE.DATABASE,
       """
-      Target false-positive rate of the bloom filter an index compaction writes for every compacted series, so a point \
-      lookup can skip a series that provably does not hold the key without reading any of its pages. Costs about 1.2 \
-      bytes per key at the default 1%; a false positive only costs the page read that would have happened anyway. \
-      Most useful when compacted series overlap in key range, i.e. keys that do not arrive already sorted (an email, \
-      a UUID, any business id); keys inserted in ascending order give each series a disjoint range that its root page \
-      already rules out. One directory page holds ~291 series, after which further series are published without a \
-      filter until a full compaction collapses the count - so a very wide index can show a high bloomProbedSeries. \
-      0 = disabled; filters already on disk are then ignored from the next database open.""",
+      Target false-positive rate of the bloom filter an index compaction writes for each compacted series of an LSM \
+      index, letting a point lookup skip a series that provably cannot hold the key without reading any of its pages. \
+      Default 0.01 (1%), enabled. Set to 0 to disable.
+      WHAT IT DOES. A lookup walks every compacted series from newest to oldest. A series' root page already rules out \
+      a key outside its key RANGE, but not a key inside the range that the series simply does not hold - so without a \
+      filter that series still costs a root-page search and a data-page read to discover nothing. The filter answers \
+      that question from a single 8 KB page instead.
+      WHEN IT HELPS. Most when the compacted series OVERLAP in key range, which is what any key that does not arrive \
+      already sorted produces: an email, a UUID, a business id. The extreme case is a bulk load into a UNIQUE index, \
+      where the duplicate check for every incoming record misses in every series by definition. Measured on 2M keys \
+      over 9 series: absent-key lookups about 2x faster and 2x fewer pages read; keys that ARE present also gain, \
+      because a key lives in one series and the filters spare the reader the others.
+      WHEN IT DOES NOT. Keys inserted in ascending order (a counter, a timestamp) give each series a disjoint key \
+      slice that its root page already rules out for free, leaving the filters little to save. Range scans and cursors \
+      never consult them at all: a filter can answer for one key, never for an interval.
+      WHAT IT COSTS. About 1.2 bytes per key on disk at 1% (roughly 3% of the index it describes), in a separate \
+      '<index>_bf.bfidx' paginated component alongside the compacted index. A false positive costs only the page read \
+      that would have happened anyway. During compaction it holds 8 bytes of transient heap per key of the series \
+      being written. Lower rates cost more bytes and more probes per lookup; higher rates save space and read more \
+      series. Below roughly 0.001 the extra bytes stop paying for themselves.
+      OPERATIONAL NOTES. Filters are written by compaction only, so an existing index gains them at its next \
+      compaction and never needs a rebuild. They are replicated over HA and included in backups like any other \
+      component, and are dropped with the index. An older ArcadeDB does not recognise the file and ignores it, so \
+      downgrading is safe for reading - but a downgraded build compacts without maintaining the filters, so prefer \
+      running a full compaction (or this setting at 0) if you downgrade and come back. \
+      A directory page holds ~255 series; beyond that further series are published without a filter until a full \
+      compaction collapses the count, which the default arcadedb.indexCompactionFullSeriesThreshold=10 does long \
+      before it can happen. Watch 'bloomSkippedSeries' and 'bloomProbedSeries' in the index statistics to see what \
+      the filters are actually saving.
+      DISABLING. 0 stops new filters being written immediately and stops existing ones being consulted from the next \
+      database open. Lookups then behave exactly as they did before the feature existed.""",
       Float.class, 0.01f),
 
   VECTOR_INDEX_LOCATION_CACHE_SIZE("arcadedb.vectorIndex.locationCacheSize", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -571,7 +571,11 @@ public enum GlobalConfiguration {
       Target false-positive rate of the bloom filter an index compaction writes for every compacted series, so a point \
       lookup can skip a series that provably does not hold the key without reading any of its pages. Costs about 1.2 \
       bytes per key at the default 1%; a false positive only costs the page read that would have happened anyway. \
-      0 = disabled, and filters already on disk are then ignored.""",
+      Most useful when compacted series overlap in key range, i.e. keys that do not arrive already sorted (an email, \
+      a UUID, any business id); keys inserted in ascending order give each series a disjoint range that its root page \
+      already rules out. One directory page holds ~291 series, after which further series are published without a \
+      filter until a full compaction collapses the count - so a very wide index can show a high bloomProbedSeries. \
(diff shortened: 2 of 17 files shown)


lvca:
Added `LSMTreeBloomFilterBenchmark` (`@Tag("benchmark")`), and writing it turned up a real mistake in the first commit.

## The filter file was inheriting the index's page size

An LSM index defaults to **256 KB** pages, and the filter component was created with `compacted.getPageSize()`. That is wrong twice over: a probe reads exactly one page, so every probe was fetching as much as the data page it exists to avoid — and every series rounds its bits up to a whole page and spends another on its directory.

Measured on a 100k-key index: **2.00 MB of filter to describe 120 KB of bits — 20.97 bytes/key, 47% of the index.**

The filter now uses a fixed **8 KB** page, independent of the index. Same index: **1.45 bytes/key, 3.5% of the index**, and a probe reads 8 KB instead of 256 KB. A block still holds ~6.8k keys at a 1% target so the blocked-layout skew stays under 2%, and one directory page still lists 291 series.

## Results — 2M keys, unique index, 9 compacted series

```
compacted index      : 73.75MB
bloom filter file    : 2.40MB  (1.26 bytes/key, 3.3% of the index)

CONSTRAINED CACHE - asked for 16 MB (limit 16.00MB, actually resident 68.33MB) against a 73.75MB index
  absent  lookups  off      359,399/s  on      522,839/s   1.45x
  absent  pages    off          780    on          595    1.31x fewer
  absent  bytes    off     195.00MB    on     148.75MB
  present lookups  off      420,266/s  on      399,390/s   0.95x

RESIDENT CACHE - 4096 MB, the whole index fits
  absent  lookups  off      372,819/s  on      563,858/s   1.51x
  absent  pages    off            0    on          137
  present lookups  off      483,391/s  on      440,189/s   0.91x
```

**~1.5x on the duplicate-check workload**, and it is nearly identical in both regimes — so on this hardware the win is almost entirely CPU (root and leaf binary searches not performed), not avoided I/O. The I/O saving is real but secondary here (1.31x fewer page reads, 24% fewer bytes); it would matter more on slower storage or a larger index-to-cache ratio. Present-key lookups, which a filter cannot help, cost 0.91–0.95x.

## Two things the benchmark has to get right, or it measures nothing

Both of these produced flat, meaningless results in my first drafts:

1. **Absent keys must interleave with the stored ones.** A key outside a series' range is already ruled out by its root page, for free and without any filter. Drawing absent keys from their own prefix (`miss-…` against `user-…`) means every probe sorts outside every series and nothing is ever read — the two columns come out identical. Present keys now take even slots and absent keys the odd ones.
2. **Probes must arrive in random order.** Ascending probes hit the same 256 KB page thousands of times running, so everything stays resident and the run becomes a CPU benchmark whatever the cache size says.

## An engine finding, unrelated to this PR

`PageManager.checkForPageDisposal()` returns early if less than 100 ms have passed since the last sweep, so **`arcadedb.maxPageRAM` is a soft bound**: on a fast read-only workload the cache overshoots it freely in between. In these runs a **16 MB limit held 68 MB resident**. The benchmark therefore reports what was actually resident rather than what was configured. Worth knowing for anyone sizing a container.

lvca:
Both reviews addressed in `6db5dfacb`. Thanks — three of these were real, and chasing one of them turned up a benchmark flaw that had me reporting the feature as *worse* than it is.

## Fixed

**Per-probe allocation** (review 1, #1 — the one flagged as blocking). Correct, and on exactly the wrong path: a slice (two `ByteBuffer`s), a `Binary` and a `BufferBloomFilter` per series, per lookup, for every record of a bulk load. The probe now reads bits straight off the page via a static `BufferBloomFilter.mightContainHash(BasePage, slots, probes, hash)`. The bit arithmetic is deliberately *repeated* rather than shared, because the two forms read from different sources — and `theStaticProbeAgreesWithTheInstanceOne` pins them over 40k hashes, since a drift between them is a false negative.

**Duplicate `.bfidx` when the rate is raised at runtime** (review 2, #1). Real and I'd missed it. `attachBloomFilter()` early-returned on rate 0, leaving the registered component unadopted; a later compaction then minted a second file for the same logical name. The component is now always adopted, and a separate `bloomFilterEnabled` flag (captured at load) decides whether lookups consult it. `isBloomFilterEnabled()` exposes the distinction.

**ThreadLocal residue** (review 2, #2). Real — it was per-instance and a full compaction replaces the instance, so one `Binary` was pinned per thread per compacted index ever created. Now static: one per thread for the whole JVM.

**Probe rode the caller's transaction** (review 2, #3 — flagged as a judgment call). Changed. Nothing modifies a filter page inside a transaction — they are written only by compaction, outside one — so the probe now uses `getImmutablePage`, which is cheaper and matches what the directory load already did. That also settles review 1's #4b.

**`Long.MIN_VALUE` sentinel** (review 1, #2). Both reviews called it harmless; fixing it exposed something that wasn't. My first attempt returned `0L` on a serialization failure while leaving the "use filters" flag set — probing with a hash that isn't the key's, which is a false negative. It is now an explicit flag set *only after* a hash has actually been produced.

**Directory cap** (review 1, #4a) — documented in the config docstring.

## Already done

**HA replication test** (review 1, #3) — landed in `d6bd9c9d5` before that review: `Issue5517BloomFilterReplicationIT`, 3 nodes, compacting **twice** so the incremental-append path is covered and not just file creation. Mutation-verified: with the directory updating in RAM but never reaching disk again — precisely what an in-place page-0 rewrite looks like to a follower — round 2 fails with a short directory while round 1 still passes.

## The benchmark was understating the feature, and I had one conclusion backwards

Chasing the allocation numbers, I found I was inserting keys in **ascending** order. A compaction slices mutable pages by RAM and mutable pages fill in insertion order, so ascending keys give every series a *disjoint* key slice that its root page rules out for free — the case where these filters have least to offer. A load keyed by an email or a UUID arrives in no key order, every series spans the whole range, and the root page rules out nothing. That is the #5470 workload.

With permuted insertion, on 2M keys / 9 series:

|  | ascending (before) | permuted (representative) |
|---|---|---|
| absent lookups, constrained cache | 1.45x | **2.14x** |
| absent pages read | 1.31x fewer | **2.15x fewer** (335 MB → 156 MB) |
| absent lookups, resident cache | 1.51x | **2.31x** |
| present lookups | 0.91–0.99x | **1.25–1.32x** |

**Correction to what I posted earlier:** I said present-key lookups cost 0.8–0.99x and that filters cannot help hits. That was wrong — an artefact of the insertion order. A key lives in one series, and the filters spare the reader the other eight, so hits gain too.

The honest caveat, now in the config docstring: this pays when series overlap in key range. For a counter or timestamp key the root page already does the work.

## Verification

15,646 tests, 0 failures, 0 errors, run sequentially so nothing collides:

| scope | tests |
|---|---|
| `com.arcadedb.index.**` + `engine.**` | 1498 |
| `database`/`graph`/`schema`/`serializer`/`utility`/`function`/`security` | 2237 |
| `com.arcadedb.query.**` | 10091 |
| root + `compression`/`event`/`exception`/`log`/`partitioning` | 280 |
| **ha-raft** (whole module) | 772 |
| **server** (whole module) | 768 |

lvca:
Thanks — this one was mostly a confirmation pass, and the checks listed (buffer offsets, all-zero page persistence, publish ordering, the adjacent-hash dedup being safe because equal hashes probe equal bits) match my own reasoning. Acted on all three observations in `0898b8f62`.

## 1. Downgrade residual — closed structurally, not just documented

You're right that this is the documented crash residual made more reachable, and the diagnosis is exact: an older build opens the database, compacts, and never touches the `.bfidx`. Its incremental rounds only *append* series, so existing entries stay truthful — but a round it rolls back frees a position a later round can reuse, and then only `seriesPages` stands between a stale entry and a series it knows nothing about.

Rather than only document it, I took the defense-in-depth option, because it turned out to be free. **The probe has already read the series' last data page** — that is where `rootPageCount` comes from — so each entry now also carries a fingerprint of that page (its entry count and free position), re-checked live:

```java
int seriesFingerprint(final BasePage lastPageOfSeries) {
  return getCount(lastPageOfSeries) * 31 + getValuesFreePosition(lastPageOfSeries);
}
```

Defined once in `LSMTreeIndexCompacted` so the compaction and the lookup cannot compute it differently — the same discipline as the two probe implementations. Zero extra I/O on either side. For a stale entry to answer now, two different series would have to sit at the same root page, span the same number of pages, *and* end with the same number of entries occupying the same bytes.

`FORMAT_VERSION` goes to 2, so a directory written before this is ignored rather than misread — the index just has no filters until its next compaction. The downgrade caveat is documented next to the crash residual regardless, since defense-in-depth is not a guarantee.

## 2. `LongAdder` — taken

You called it not worth doing preemptively; I did it anyway. It is a three-line change, the counters are bumped once per series per lookup, and a bulk load does that from every thread it has. Leaving a contended cache line on the path whose cost this feature exists to remove seemed like the wrong default, and reads are diagnostics that can afford the sum.

## 3. The ~291-series cap — one mitigating detail

Worth noting for the record: with `arcadedb.indexCompactionFullSeriesThreshold` at its default of **10**, a full compaction collapses the series count long before 291, so the cap is effectively unreachable. It needs that setting at 0 (disabled) to matter. The entry is now 32 bytes so the cap is ~255, which does not change that conclusion. Left as-is, per your read.

## Verification

- `com.arcadedb.index.**` + `engine.**` — 1498 tests, 0 failures
- `Issue5517BloomFilterReplicationIT` — 3-node cluster, both compaction rounds, green
- benchmark re-run, 2M keys / 9 series: absent-key lookups **1.93x** constrained / **2.40x** resident, **2.11x fewer page reads**; present-key **1.15–1.41x**

Run-to-run spread on this machine is roughly ±20%, so those are a range rather than a point estimate.

PR #5537, opened 2026-07-29
fix(#5492): recordFileChanges ships the WAL its callback committed

### LINKED ISSUE #5492: HA: a materialized view makes the Raft leader ship page versions followers never received - WALVersionGapException, non-converging snapshot-resync loop, lost committed writes

robfrank:
Refs #5492. **Deliberately does not close it** - see "Scope" below.

## Summary

A commit that runs inside a `recordFileChanges()` DDL callback does not replicate itself. At that point the files the DDL is creating do not yet exist on followers, so a `TX_ENTRY` targeting them would be dropped; the WAL is buffered into a thread-local instead and is meant to travel embedded in the `SCHEMA_ENTRY` sent once the callback returns.

The drain was unconditional but the send was not:

```java
final List<byte[]> walEntries = new ArrayList<>(schemaWalBuffer.get());
...
schemaWalBuffer.get().clear();

if (!addFiles.isEmpty() || !removeFiles.isEmpty() || schemaChanged) {   // walEntries absent
```

and the `finally` cleared the buffer either way. `commit2ndPhase` has already run by then, so a callback that wrote records without creating a file or moving the schema version had its pages applied on the leader and the only copy that could reach a follower silently discarded. Follower page versions then trail by exactly those writes, and the next ordinary transaction touching one of those pages fails on them with `WALVersionGapException`, marking the database diverged and scheduling a snapshot resync.

The sibling `runWithCompactionReplication` drains the same two thread-locals and has always guarded them with `walEntries.isEmpty()` included. The asymmetry between the two was the defect.

## Follower side

Shipping WAL with no files and no schema JSON creates an entry shape followers had never seen: the compaction path always sends a non-empty `serializedSchema`. `applySchemaEntry` already tolerated an empty `schemaJson`, but would still fall through to `load(MODE.READ_WRITE, true)`. On the single Raft apply thread that reload re-instantiates every TimeSeries engine, closing shard executors with a 30 s `awaitTermination` - and there is nothing for it to pick up on an entry carrying only page writes, since `applyChanges` already updates page counts through `getFileByIdIfExists()`. Such entries are now recognised and the reload skipped, exactly as sealed-only TimeSeries entries already were.

## Scope - what this does not do

This does **not** close #5492. The materialized-view A/B reported there is still unreproduced in-process, and this branch makes no claim to have fixed it.

Stated plainly: every `LocalSchema.recordFileChanges` caller in `engine/` mutates the schema, so `schemaChanged` was true and the old guard already held. No existing caller has been shown to reach the dropped branch. What this closes is a latent hole in a published `DatabaseInternal` contract - one that `runWithCompactionReplication` already guards against - not the demonstrated cause of the reported symptom. The test drives it through the public API rather than a contrived internal hook.

The `ha-raft/CLAUDE.md` sharp-edge section is updated rather than removed, for the same reason. Its paragraph on refresh requests being dropped was also stale, having been fixed in #5502.

## Test plan

- [ ] `Issue5492SchemaWalNotShippedIT` - 2-node in-process Raft cluster. Creates a type through ordinary DDL and lets it replicate, then drives a `recordFileChanges()` callback on the leader that commits a record into that existing type and touches nothing else.
- [ ] Verified failing before the fix: `expected: 2L` at "WAL buffered inside recordFileChanges must reach the follower, not be discarded" - the follower held 1 record.
- [ ] Passes after the fix, and run time drops from 41.7 s to 18.1 s (no longer burning the 30 s convergence deadline).
- [ ] Asserts both halves separately: that the write arrives, and that a subsequent ordinary write into the same bucket produces zero `WALVersionGapException` on the follower (via the existing `ArcadeStateMachine.TEST_WAL_GAP_COUNTER` hook). A regression that loses the write is distinguishable from one that also diverges the cluster.
- [ ] Full `ha-raft` module: 772 unit tests green. Integration phase 190 tests, 1 error in `Issue5297PolymorphicCountIT` - a teardown `DatabaseAreNotIdentical` alongside a group-commit interrupt on shutdown. That test passes 3/3 in isolation on this branch, and the local machine was at 97% disk, which forced Raft snapshot/log purges mid-suite. A `main` baseline run to confirm it as pre-existing flakiness was still in progress when this PR was opened; CI is the better arbiter.


robfrank:
CI triage, checked against a `main` baseline run (30438140034) rather than assumed benign.

**The two tests that mattered both passed:** `Issue5492SchemaWalNotShippedIT` (14.8 s) and `Issue5297PolymorphicCountIT` (11.9 s). The latter is the teardown error flagged in the test plan - it was local flakiness on a 97%-full disk, not a regression, and that question is now closed.

**`ha-integration-tests`** fails on `main` too: 4 failures there, 5 here. The failing sets are largely disjoint *in both directions* - `RaftCommandReadConsistencyIT` and `RaftReadConsistencyBookmarkIT` fail on `main` but pass here, while `RaftReplicationChangeSchemaIT`, `RemoteStickyConnectionStrategyIT` and `RaftHTTP2ServersCreateReplicatedDatabaseIT` do the reverse. All three of the latter pass locally on this branch (4/4, 1/1, 1/1), and their signature is `DatabaseIsClosedException` during teardown, a shutdown race rather than a divergence. I checked `RaftReplicationChangeSchemaIT` specifically because schema replication is this change's own path.

**`slow-unit-tests`** is `EdgeAppendMergeRaceTest`, an engine-module concurrency race. This PR changes zero engine code (`git diff --name-only origin/main...HEAD` is 3 ha-raft files plus 2 docs), so it cannot be implicated.

**`integration-tests`** fails on `main` too, and its module list explicitly excludes `ha-raft`.

**`Meterian client scan`** fails on `main` on 5 of the last 5 runs; no dependency is added here.

Worth noting separately: the ha-raft IT suite is genuinely flaky on both branches, with a rotating cast of teardown races. That is pre-existing and out of scope here, but it makes this job a weak signal for any PR.

PR #5560, opened 2026-07-29
feat(engine): let the schema dictionary grow past a single page

lvca:
> [!IMPORTANT]
> **Release note / upgrade requirement: in a cluster, upgrade followers before, or together with, the leader.**
>
> Dictionary pages replicate as raw pages. Once a leader's dictionary grows past its first page it ships page 1 and
> beyond to its followers, and a follower still running a build without multi-page support writes those pages but
> reloads only page 0. Its in-RAM dictionary is then missing every name past the first page, and each record
> referencing one fails with `Dictionary item with id N is not valid`.
>
> Related: a database that has rolled over can no longer be opened by an older ArcadeDB at all. To check whether a
> given database has rolled over, divide its dictionary file size by the page size in its own file name:
> `ls -l <db>/dictionary.*.dict` - 131072 / 65536 = 2 pages means rolled over. A file at or under one page is still
> single-page and remains downgradable.
>
> Full detail in `docs/5560-dictionary-multipage.md`, section *Upgrade notes*.

## The limit

`com.arcadedb.engine.Dictionary` maps every type name, property name and enumerated string value to a small integer id, and that id is what records carry instead of the name. Despite the class javadoc saying "CONTENT-PAGES", it only ever used **page 0**, so a database was capped at whatever fitted there: `pageSize - 8 - 4` = **327,668 bytes of names**, measured at **48,396** short names before the write was refused.

Past that point `CREATE PROPERTY`, or inserting a document with a new field name, failed permanently. There was no path to grow, and entries are never reclaimed, so a schemaless workload with dynamic field names walked steadily toward it.

## What changed

Names roll over to a new page. Every page, page 0 included, carries the same 4-byte legacy counter, so **a dictionary written before this change is exactly a dictionary of one page and loads unchanged**: no migration, no rewrite, no format flag needed to read an existing database.

The invariant everything follows from: **an id is the ordinal of the name in page order**, and that id sits inside records on disk. So the layout is strictly append-only across pages. Names go only to the LAST page, and a page left behind is never revisited even when it still has room. Filling a gap would renumber every name after it and silently repoint every record referencing them, which is the one way this change could corrupt data rather than throw. A name is never split across pages, so the wasted tail is under 1% at identifier lengths, and the only limit left is a single name larger than one page, which is now refused by name.

Three details worth review attention:

- `addItemToPage` **reads** the tail page before deciding where to write. Asking for it as modifiable and then finding it full would bump its version, rewrite it at commit for nothing, and make it false-conflict with concurrent transactions.
- `reload()` walks the **committed** page count, not `getTotalPages()`. `TransactionContext.rollback()` calls `reload()` while the transaction's own page counter still counts pages being thrown away. The floor of 1 keeps the previous behaviour of always reading page 0 of a non-empty file.
- `updateName()` re-lays the dictionary out from page 0 in the same order, so no id moves, and **empties the pages the new content no longer reaches**. `reload()` walks every committed page, so a stale tail page would otherwise re-add its old names.

## Page size and format version

`DEF_PAGE_SIZE` drops from 327,680 to **65,536** for new dictionaries. That size existed only to make the single available page hold as many names as possible. Now that pages roll over, a new name dirties and eventually flushes one page, so a smaller page is 5x less write amplification per name. Existing databases keep the page size they were created with, which is read back from the file name.

`CURRENT_VERSION` goes to 1, with a guard refusing a dictionary written by a newer ArcadeDB rather than misreading it.

**Downgrade note:** once a database grows past page 0 it can no longer be opened by an older ArcadeDB, which reads page 0 only and reports `Dictionary item with id N is not valid`. Loud, not silent, and such a database could not have existed before this change since the write would have been refused.

## Tests

New `DictionaryMultiPageTest` (in `com.arcadedb.engine`, so it can inspect the on-page layout):

- rollover instead of refusal, with every id resolvable both ways
- ids survive a reopen
- **page 0 still reads with the pre-multi-page algorithm**, which is the compatibility guarantee
- a name too big for an empty page is refused clearly and leaves the dictionary usable
- documents whose property names span pages read back after a reopen (the serializer round trip)
- `updateName` rewrites every page and keeps every other id
- the **follower path**: a replicated WAL transaction carrying a brand new dictionary page creates it, makes its names visible, and the next local name continues after it

`DictionaryLimitsTest` had a test asserting the old "no space left in dictionary file" behaviour; it now asserts that filling well past one page keeps working.

Local run: 920 engine test classes, 0 failures, at the time of pushing (the full run was still going through an unrelated long test).

### DIFF
--- a/docs/5560-dictionary-multipage.md
+++ b/docs/5560-dictionary-multipage.md
@@ -49,6 +49,22 @@ names); a string value is only ever looked up with `create=false`, so user data 
 `CURRENT_VERSION` goes to 1, with a guard refusing a dictionary written by a newer ArcadeDB rather than
 misreading it as this layout.
 
+### Why the version is not bumped when an existing database first rolls over
+
+The version is stamped at creation, so a `v0` database that later grows past page 0 keeps saying `v0` on
+disk. Bumping it on first rollover was considered and rejected:
+
+- The version is part of the **file name** (`dictionary.0.65536.v1.dict`). Changing it at runtime means
+  renaming a live component file. `PaginatedComponent.rename` can only do that behind a full flush barrier
+  (`waitAllPagesOfDatabaseAreFlushed`), and doing it from inside an append transaction would race
+  replication, which ships pages by file id, and leave follower file names diverged from the leader's.
+- It would protect nothing. `ComponentFactory` passes the parsed version straight through without
+  validating it, so a build old enough to be at risk opens a `v1` file just as happily as a `v0` one. The
+  guard added here only helps a *newer* format meeting *this* reader.
+
+So the version records what wrote the file, and the operational signal that a database has left single-page
+territory is the INFO line logged when page 1 is created.
+
 ## Upgrade notes
 
 **Single node.** Nothing to do. An existing database keeps its page size and its `v0` file name, and gains
--- a/docs/5560-dictionary-multipage.md
+++ b/docs/5560-dictionary-multipage.md
@@ -65,6 +65,21 @@ disk. Bumping it on first rollover was considered and rejected:
 So the version records what wrote the file, and the operational signal that a database has left single-page
 territory is the INFO line logged when page 1 is created.
 
+### How to tell whether a database has rolled over
+
+The log line is emitted once, when it happens, so it is no help for a database you are looking at after the
+fact. The durable answer is the file size: the dictionary is one page per `pageSize` bytes, and `pageSize`
+is in the file name.
+
+```
+$ ls -l <database>/dictionary.*.dict
+-rw-r--r--  1 arcadedb  arcadedb  131072  dictionary.0.65536.v1.dict
+```
+
+131072 / 65536 = 2 pages, so this database has rolled over and cannot be read by a build without multi-page
+support. A file of exactly `pageSize` bytes (or less) is still single-page and remains downgradable. This is
+worth checking before rolling a cluster back to an older build.
+
 ## Upgrade notes
 
 **Single node.** Nothing to do. An existing database keeps its page size and its `v0` file name, and gains
--- a/docs/5560-dictionary-multipage.md
+++ b/docs/5560-dictionary-multipage.md
@@ -1,0 +1,86 @@
+# PR #5560 - Schema dictionary spans multiple pages
+
+## Symptom
+
+`com.arcadedb.engine.Dictionary` held every type name and property name in page 0 alone, so a database was
+capped at whatever fitted there: `pageSize - 8 (page header) - 4 (legacy counter)` = 327,668 bytes,
+measured at 48,396 short names. Past that, `CREATE PROPERTY` and inserting a document with a new field
+name failed permanently with `DatabaseMetadataException: No space left in dictionary file`, and there was
+no way to grow. Entries are never reclaimed, so a schemaless workload with dynamic field names walked
+steadily toward the wall.
+
+## Fix
+
+Names roll over to a new page. Every page, page 0 included, carries the same 4-byte legacy counter, so a
+dictionary written before this change is exactly a dictionary of one page and loads with no special case:
+no migration, no rewrite, no format flag needed to read an existing database.
+
+The invariant everything follows from: **an id is the ordinal of the name in page order**, and that id is
+written inside records on disk. The layout is therefore strictly append-only across pages. Names go only to
+the last page, and a page left behind is never revisited even when it still has room; filling a gap would
+renumber every name after it and silently repoint every record that referenced them.
+
+Three details carry the correctness:
+
+- `addItemToPage` **reads** the tail page before deciding where to write. Enlisting it as modifiable and
+  then finding it full would bump its version, rewrite it at commit for nothing, and false-conflict with
+  concurrent transactions (the page is the conflict unit, see `engine/CLAUDE.md`).
+- `reload()` walks `Math.max(1, pageCount.get())` pages and reads them through `PageManager`, never through
+  `TransactionContext.getPage()`. Both halves speak durable state: `getTotalPages()` would count pages a
+  rolling-back transaction is discarding, and `TransactionContext.getPage()` resolves `modifiedPages` first,
+  so the rollback reload would rebuild the dictionary from exactly the content being thrown away. That
+  second half was a defect predating this change, since `updateName` always rewrote page 0 inside the
+  caller's transaction.
+- `updateName` re-lays the dictionary out from page 0 in the same order, so no id moves, and empties the
+  pages the new content no longer reaches. `reload()` walks every committed page, so a stale tail page
+  would otherwise re-add its old names.
+
+## Page size and format version
+
+`DEF_PAGE_SIZE` drops from 327,680 to 65,536 for new dictionaries. That size existed only to make the
+single available page hold as many names as possible; now that pages roll over, a new name dirties and
+eventually flushes one page, so a smaller page is 5x less write amplification per name. Existing databases
+keep the page size they were created with, read back from the file name.
+
+One consequence: on a new database a single identifier caps at `pageSize - 12` = ~65Kb, against ~327Kb
+before. Only identifiers ever enter the dictionary (the `create=true` callers: type names and property
+names); a string value is only ever looked up with `create=false`, so user data is never subject to this.
+
+`CURRENT_VERSION` goes to 1, with a guard refusing a dictionary written by a newer ArcadeDB rather than
+misreading it as this layout.
+
+## Upgrade notes
+
+**Single node.** Nothing to do. An existing database keeps its page size and its `v0` file name, and gains
+rollover on the next write that needs it.
+
+**Downgrade.** Once a database has grown past page 0 it can no longer be opened by an older ArcadeDB, which
+reads page 0 only and reports `Dictionary item with id N is not valid`. Loud, not silent. Such a database
+could not have existed before this change, since the write would have been refused. A new database that has
+never rolled over stays readable by an older build, so downgradability is lost exactly when the database
+starts depending on the feature, not before.
+
+**Cluster: upgrade followers before, or together with, the leader.** Dictionary pages replicate as raw
+pages through `TransactionManager.applyChanges`. A new-version leader that rolls the dictionary over ships
+page 1 and beyond to its followers; a follower still running a build without multi-page support writes the
+page but reloads only page 0, so its in-RAM dictionary is missing those names and every record referencing
+them fails with `Dictionary item with id N is not valid`. Upgrading followers first avoids the window
+entirely.
+
+## Tests
+
+`engine/src/test/java/com/arcadedb/engine/DictionaryMultiPageTest.java`
+
+- `namesRollOverToNewPagesInsteadOfBeingRefused` - past one page it keeps working, ids resolve both ways
+- `idsSurviveAReopen` - the mapping is identical after a reopen
+- `pageZeroKeepsTheLegacySinglePageLayout` - page 0 still reads with the pre-multi-page algorithm, which is
+  the backward-compatibility guarantee
+- `aNameTooBigForAnEmptyPageIsRejectedAndLeavesTheDictionaryUsable`
+- `documentsWithPropertyNamesSpanningPagesReadBack` - the serializer round trip across pages
+- `updateNameRewritesEveryPage` / `updateNameGrowingBeyondTheExistingPagesAddsOne` - shrink and grow
+- `aRolledBackTransactionThatGrewTheDictionaryLeavesItIntact` - the case that exposed the dirty-page reload
+- `aDictionaryFromANewerFormatIsRefusedInsteadOfMisread` - the version guard through the real load path
+- `aReplicatedNewDictionaryPageIsVisibleAfterApplyChanges` - the follower apply path
+
+`engine/src/test/java/com/arcadedb/schema/DictionaryLimitsTest.java` - filling well past one page keeps
+working, entries are never reclaimed, both directions survive a reload.
--- a/engine/src/main/java/com/arcadedb/engine/Dictionary.java
+++ b/engine/src/main/java/com/arcadedb/engine/Dictionary.java
@@ -34,7 +34,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 
@@ -55,8 +54,11 @@ import java.util.logging.Level;
  * revisited even when it still has room, and nothing is ever removed. Writing into an earlier page would renumber every name
  * after it and silently repoint every record that referenced them.
  * <br>
- * A name is never split over two pages, so the tail of a page is left unused when the next name does not fit: at identifier
- * lengths that is well under 1% of the file. The only hard limit left is a single name larger than one page, which caps one
+ * A name is never split over two pages, so the tail of a page is left unused when the next name does not fit: on the append path,
+ * at identifier lengths, that is well under 1% of the file. {@link #updateName} is looser - a rewrite that shrinks the content
+ * leaves the pages it no longer reaches empty, and appends resume on the last of them, so the pages in between stay empty for
+ * good. They cost one zero-entry read each on load and nothing else, and no production code calls that method. The only hard
+ * limit left is a single name larger than one page, which caps one
  * identifier at {@code pageSize - BasePage.PAGE_HEADER_SIZE - DICTIONARY_HEADER_SIZE} bytes: ~65Kb on a dictionary created with
  * the current {@link #DEF_PAGE_SIZE}, against ~327Kb on one created before it was reduced. Existing databases keep their page
  * size, so only new ones see the lower cap, and no realistic identifier comes close to either.
@@ -94,7 +96,7 @@ public class Dictionary extends PaginatedComponent {
    * volatile field is what makes a reader either see both halves of the old snapshot or both halves of the new one, and is
    * what gives the reader a happens-before edge with the rebuild at all.
    * <br>
-   * {@code names} is an array with spare capacity rather than a {@link CopyOnWriteArrayList}, and {@code size} is how much of it
+   * {@code names} is an array with spare capacity rather than a {@code CopyOnWriteArrayList}, and {@code size} is how much of it
    * is live. A COW list copies its whole backing array on every append, so growing a dictionary of N names cost N^2/2 element
    * copies: ~150ms and 4.6Gb of copying at the 48K names that used to be the hard ceiling, and quadratically worse without one.
    * Removing that ceiling is exactly what this class now does, so the append had to stop being O(n).
@@ -169,7 +171,14 @@ public class Dictionary extends PaginatedComponent {
           // THE COMMIT ABOVE CAN HAVE SWAPPED THE SNAPSHOT (RELOAD ON ROLLBACK/RETRY): RE-READ IT
           current = entries;
 
-          if (current.ids().putIfAbsent(name, newPos.get()) == null) {
+          // THE NAMES ARE PUBLISHED BEFORE THE MAP, AND THE ORDER MATTERS. THE TWO CANNOT BE UPDATED ATOMICALLY TOGETHER, SO ONE
+          // OF THEM LEADS BY AN INSTANT AND A CONCURRENT READER CAN LAND IN BETWEEN. LEADING WITH THE MAP MEANS THAT READER
+          // RESOLVES THE NAME TO AN ID THAT getNameById() DOES NOT YET ACCEPT, WHICH THROWS. LEADING WITH THE NAMES MEANS IT
+          // FINDS THE NAME NOT PRESENT YET AND -1 COMES BACK, WHICH EVERY create=false CALLER ALREADY HANDLES (BinarySerializer
+          // JUST SKIPS THE COMPRESSION) AND WHICH A create=true CALLER RESOLVES BY BLOCKING ON THIS MONITOR. NOBODY CAN HOLD THE
+          // NEW ID BEFORE THE MAP PUBLISHES IT, BECAUSE IT ONLY REACHES A RECORD AFTER THIS RETURNS. A MISS COSTS NOTHING, A
+          // THROW COSTS A REQUEST, SO THE MISS IS THE ONE TO PREFER
+          if (!current.ids().containsKey(name)) {
             final int appended = appendName(current, name);
             if (appended != newPos.get() + 1) {
               try {
@@ -179,6 +188,7 @@ public class Dictionary extends PaginatedComponent {
               }
               throw new SchemaException("Error on updating dictionary for key '" + name + "'");
             }
+            current.ids().putIfAbsent(name, newPos.get());
           }
           pos = current.ids().get(name);
         }
--- a/engine/src/main/java/com/arcadedb/engine/Dictionary.java
+++ b/engine/src/main/java/com/arcadedb/engine/Dictionary.java
@@ -474,10 +474,22 @@ public class Dictionary extends PaginatedComponent {
       final int totalPages = Math.max(1, pageCount.get());
 
       for (int pageNumber = 0; pageNumber < totalPages; ++pageNumber) {
-        // createIfNotExists IS LOAD BEARING, NOT A DEFAULT: PAIRED WITH THE FLOOR ABOVE IT MATERIALISES PAGE 0 OF A FILE THAT
-        // IS SHORTER THAN ONE PAGE (KILLED MID-WRITE), WHICH IS WHAT THE SINGLE-PAGE READER USED TO DO
-        final BasePage page = database.getPageManager()
-            .getImmutablePage(new PageId(database, file.getFileId(), pageNumber), pageSize, false, true);
+        // createIfNotExists ONLY FOR PAGE 0, WHERE IT IS LOAD BEARING: PAIRED WITH THE FLOOR ABOVE IT MATERIALISES PAGE 0 OF A
+        // FILE SHORTER THAN ONE PAGE (KILLED MID-WRITE), WHICH IS WHAT THE SINGLE-PAGE READER DID.
+        //
+        // FOR EVERY LATER PAGE IT IS OFF, SO A PAGE THAT pageCount CLAIMS BUT THAT IS NOT THERE FAILS INSTEAD OF BEING INVENTED.
+        // AN INVENTED PAGE IS NOT MERELY MASKED CORRUPTION: IT CONTRIBUTES ZERO NAMES, WHICH SHIFTS EVERY NAME AFTER IT DOWN BY
+        // AS MANY IDS AS THE MISSING PAGE HELD, AND IDS ARE EMBEDDED IN RECORDS. SILENTLY RENUMBERING IS THE ONE OUTCOME THIS
+        // CLASS EXISTS TO PREVENT, SO A GAP HAS TO BE LOUD.
+        final BasePage page;
+        try {
+          page = database.getPageManager()
+              .getImmutablePage(new PageId(database, file.getFileId(), pageNumber), pageSize, false, pageNumber == 0);
+        } catch (final IllegalArgumentException e) {
+          throw new DatabaseMetadataException(
+              "Schema dictionary of database '" + database.getName() + "' is truncated: page " + pageNumber + " of " + totalPages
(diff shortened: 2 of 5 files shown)


lvca:
Thanks, all six points reviewed. Four applied, one applied slightly differently from the suggestion, one declined with reasoning.

**1. `getAvailableSpaceInLastPage()` dead code** — dropped rather than kept as public API. The review offered either option; dropping is the better one. Dead public API on an engine class is a maintenance liability, and the "diagnostics" justification I had for keeping it was speculative. The one place that genuinely needs the tail page's free space is a test, and that test lives in `com.arcadedb.engine`, so it reads the page directly with no production API at all.

**2. `IOException` cause dropped** — fixed, now consistent with `addItemToPage`.

**3. `updateName` mutates RAM before validating** — applied, and taken one step further than suggested. `newName` is now validated before anything is touched, which is sufficient on its own: every other name is already stored on a page of this same size, so `newName` is the only one that can fail. I deliberately **kept** the per-name check inside the rewrite loop rather than replacing it. It is two arithmetic operations per name and it is the only thing between a corrupt page and `writeString` failing with a raw `cannot write outside the page space`. Cheap defence in depth on a path that is now otherwise unreachable.

**4. Orphaned pages after a shrink** — **declining**, and I want to be explicit about why rather than just calling it rare. Truncating the file is the obvious "complete" fix, but page counts are tracked in three places that would disagree with a shrunk file: `PaginatedComponent.pageCount`, `TransactionContext.newPageCounters`, and `TransactionManager.applyChanges`, which calls `updatePageCount(pageNumber+1)` on the follower. Shrinking a live file id risks a follower replaying a WAL entry for a page number that no longer exists — a version-gap cascade, which is exactly the failure class the surrounding code goes to some length to avoid. That is real risk taken on against a method with **no production callers**, to reclaim pages whose entire cost is one zero-entry read per reload. Leaving them is the better trade.

**5. Grow branch of `resetPageForRewrite` untested** — agreed, that was a genuinely uncovered path. Added `updateNameGrowingBeyondTheExistingPagesAddsOne`: renames an entry to a name that nearly fills a page on its own, so the re-layout cannot fit in the pages already there, then asserts the page count grew and every other id survived, before and after a reopen.

**6. No test for the version guard** — added, but not as a constructor unit test. Calling the constructor directly would prove the `if` fires while skipping everything that makes the guard matter. `aDictionaryFromANewerFormatIsRefusedInsteadOfMisread` renames the dictionary file to `.v99.` on disk and attempts a real `factory.open()`, asserting it fails with the version in the message and registers no active instance, then restores the file so the database is dropped normally. That covers the guard *and* the load path that reaches it.

Local: 18 dictionary tests, then 623 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database` and `com.arcadedb.engine`. 0 failures, 0 errors.

lvca:
Round 2. Point 4 turned out to be the valuable one: writing the test **found a real defect**, and it is not one this PR introduced.

**4. No test for a local transaction crossing a page boundary then rolling back** — agreed, and worth spelling out why the path is narrower than it looks. `addItemToPage` only ever *adds* a page, never modifies one in the same transaction, and `TransactionContext.rollback()` only arms the dictionary reload when a dictionary page is in `modifiedPages`. So an append-driven rollover can never reach it. The single caller that both modifies existing pages and adds one **inside the caller's transaction** is `updateName`. That is what the new test drives.

And it failed. `reload()` read its pages through `TransactionContext.getPage()`, which resolves `modifiedPages` **first** — so on the rollback path it rebuilt the dictionary from precisely the content being discarded. After a rolled-back `updateName`, the new name still resolved to the renamed id until something else happened to reload. The `pageCount.get()` half of the method spoke durable state while the page-read half spoke transaction state.

Fixed by reading through the page manager, so both halves agree. Every caller wants committed content: load, rollback, and the two post-commit replication paths. **This defect predates multi-page support** — `updateName` always modified page 0 inside the caller's transaction, so the same rollback would have rebuilt from dirty page 0. I verified the test pins the fix by reverting the read and watching it fail (`expected: -1 but was: 1`), then restoring it.

**1. "enumerated string value" reads like user data** — agreed, and I fixed it in the class javadoc rather than only in the PR body. The PR body is read once; the javadoc is where the question comes back. It now says explicitly that only identifiers are ever *added* (the `create=true` callers: type names and property names), that a string value is only ever looked up with `create=false` and stored as a reference when it already matches, and therefore that the per-name limit only applies to identifiers. To restate it here as asked: **the smaller max entry size can never apply to user data.**

**2. New single-page v1 files are not refused by old versions** — correct, and intended. A fresh database stays readable by an older ArcadeDB for as long as it fits page 0, and only becomes unreadable once it actually uses the new capability. Downgradability is lost exactly when the database starts depending on the feature, not a moment earlier. No code change; flagging it here so the note is not misread.

**3. Redundant `checkNameFitsAPage` in the loop** — keeping it, as agreed.

Local: 624 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`. 0 failures, 0 errors. `DictionaryMultiPageTest` is now 10 tests.

lvca:
Round 3. All four applied, one of them by measuring rather than doing what was suggested.

**1. Identifier cap shrinks 5x on new databases** — agreed, and it belongs in the class javadoc rather than only the PR text, which is read once. It now states the cap as `pageSize - PAGE_HEADER_SIZE - DICTIONARY_HEADER_SIZE`: ~65Kb on a new dictionary against ~327Kb on one created before, existing databases unaffected because they keep their page size. Also noted there that only identifiers are subject to it, since string values are only ever looked up with `create=false`.

**2. HA rolling-upgrade ordering** — the most valuable point of this round, and I agree it would not be inferred from the file-format note. Recorded in two places: the class javadoc, next to the code that causes it, and `docs/5560-dictionary-multipage.md`, following the repo's existing `docs/NNNN-*.md` convention. The mechanism spelled out: dictionary pages replicate as raw pages through `applyChanges`, so a rolled-over leader ships page 1+ to followers; an old follower writes the page but reloads only page 0, and every record referencing a name past the first page fails. **Followers must be upgraded before, or together with, the leader.** The doc also collects the single-node and downgrade notes so the release manager has one place to draw from.

I considered whether this could be *prevented* rather than documented — a leader refusing to roll over while any follower is on an older build. That needs cluster-wide version negotiation and would couple the engine to `ha-raft` for a hazard that a correct upgrade order removes entirely. Documenting it is the right level.

**3. `fillingWellPastOnePageKeepsWorking` might deserve `@Tag("slow")`** — you suggested a timing check, so I ran one instead of guessing. It measures **0.3s**, and every `DictionaryMultiPageTest` case is at or under 0.088s. CLAUDE.md reserves the tag for "noticeably long, multi-second" tests, so **not tagging it**: at 0.3s the tag would drop real coverage from every run that skips slow tests, in exchange for nothing.

**4. `createIfNotExists=true` is load-bearing** — correct, and a fair thing to trip over. Commented at the call site: paired with the `Math.max(1, pageCount.get())` floor it materialises page 0 of a file shorter than one page (killed mid-write), which is what the single-page reader did.

Local: 19 dictionary tests, 0 failures. Previous round's wider run was 624 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`, also clean; this round changed only javadoc, one comment and a new doc file.

lvca:
Round 4. Two applied, one stale, one agreed as non-actionable.

**1. `getAvailableSpaceInLastPage()` has no callers** — already gone, deleted in `249bd99bd` back in round 1 for exactly this reason. `grep` over `engine/`, `ha-raft/` and `server/` returns only `LocalBucket.getAvailableSpaceInPage`, which is unrelated. Nothing to do; flagging that this point was reviewing a stale view of the branch.

**2. `updateName` rewrite over a `CopyOnWriteArrayList`** — you called it acceptable with a heads-up; I think it is worth actually fixing, because the cost is not the interesting part. The old shape was `while (indexOf(oldName) != -1) { set(...) }`, which re-scans the whole list per occurrence **and only terminates because the two names differ** — with `oldName.equals(newName)` every scan finds what the previous `set` just wrote, which is the infinite loop fixed by guard earlier in this PR. Scanning by index instead is one pass rather than one per occurrence, and makes non-termination impossible by construction. The early return stays, but as a fast path rather than as the only thing preventing a spin. Better code, not just fewer array copies.

**3. Test timing on a slow runner** — fair push-back on my local 0.3s measurement; my machine is not a 2-vCPU runner. But between the two options I think there is a third that beats both. The test exists to show the dictionary keeps working *well past one page*, and **3,000** padded names cross four pages just as convincingly as 20,000 do — the other 17,000 bought nothing but WAL commits. Resized: **0.3s to 0.068s**, a 4.4x cut with the coverage intact, and I strengthened the assertion from `getTotalPages() > 1` to `>= 4` so the fixture size is now pinned by what it proves. Even 10x slower on a constrained runner that is ~0.7s, so it stays out of `@Tag("slow")` — where it would have cost real coverage on every run that skips slow tests.

**4. Concurrency corner in `getIdByName`** — agreed on all counts, and noted as informational. `newPos` is captured before the commit while a replication `reload()` could swap the snapshot; the `current.names().size() != newPos + 1` guard reloads and throws rather than corrupting. Fail-safe, pre-existing in shape, and it survives the multi-page change because the ordinal is still the ordinal. No change.

On verification: agreed that green CI on `engine` and `ha-raft` is the gate, since neither of us can run the build in the review sandbox. Local this round: 624 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`, 0 failures, 0 errors.

lvca:
Round 5. Point 1 was right, and I took it further than proposed. Point 3's second half was also worth acting on.

**1. Two validations still ran after the in-RAM mutation** — confirmed exactly as you traced it, including the part that makes it bite: `IllegalArgumentException` is not an `IOException` so the local repairing `catch` does not fire, and no dictionary page has been written yet so `rollback()` does not arm its reload. Nothing repairs it.

You proposed computing `oldIndexes` read-only, then running the empty/type-name checks, then mutating. I went one step further: **the type-name check moved above the scan**, not just above the mutation. It depends only on `oldName`, so a rename refused for that reason now walks nothing at all. Final order is all four preconditions, then a read-only scan, then the `isEmpty` check, then the mutation — after which the only way out is an `IOException`, which is precisely what the `catch` repairs. That makes the comment "validation is over" literally true at that point rather than approximately.

Regression test added as suggested, and I verified it fails against the old ordering: `expected: 131` — the id had vanished from the map while the list already showed the new name.

**3, second half: no contention test** — agreed, and this was a real gap, since the class's whole safety argument is "appends serialise on the monitor while `reload()` runs on threads that do not hold it" and nothing exercised it. Added: four threads appending 250 distinct names each across a rollover, asserting no two ids collide, that the ids cover exactly the expected range, that every id round-trips, and — the assertion that actually matters — that **after a `reload()` the ids handed out under contention still match the order the names landed in on the pages**. That is what would catch two threads computing the tail page at once, which is the specific way concurrency could renumber entries here.

**2. Release notes / upgrade guide** — the mechanism and the "upgrade followers before, or together with, the leader" instruction are in `docs/5560-dictionary-multipage.md` under an explicit *Upgrade notes* heading, alongside the single-node and downgrade cases, so whoever writes the release notes has one place to draw from. I have deliberately not written a release-notes file itself, as that is not mine to author on this branch.

**3, first half: identifier cap** — noted for completeness and already documented in the class javadoc. No change.

Local: 626 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`. 0 failures, 0 errors. `DictionaryMultiPageTest` is now 12 tests.

lvca:
Round 6. Both javadoc items addressed, one of them with code rather than only a note, and the optional item taken with one adjustment.

**1. `updateName` now dirties every page, so the WAL commit is no longer bounded by one page** — correct, and now stated in the method javadoc: the rewrite re-lays the dictionary out from page 0 into the caller's transaction, so the commit grows with the dictionary rather than with the single name that changed.

**2. `updateName` is public, unsynchronized, no production caller** — you suggested a "not synchronized; assumes an exclusive schema lock" note. I think that is the weaker fix: a note tells the next reader about the hazard, it does not remove it, and you correctly identified that the hazard is a concurrent `getIdByName` append mutating the same `entries` snapshot. So **`updateName` is now `synchronized`**, on the same monitor as the create path. One keyword on a method with no production caller and no contention, and the in-RAM race is gone by construction.

I was deliberate about what that does and does not buy, and the javadoc now says so: it serialises the **in-RAM** edit, which is the part with no other protection — an append would otherwise observe a half-renamed list. It does **not** make the two atomic on disk, because `updateName` writes inside the caller's transaction and does not commit, so an append committing in between is resolved the normal way, by the page version check raising `ConcurrentModificationException` on whichever commits second. That is the correct outcome, not a gap. The note about a live rename also needing a schema-level lock is there too.

I checked for deadlock before adding it: `updateName` acquires no other lock while holding the monitor, and the caller's commit happens after it returns, so there is no cycle with `getIdByName`, which takes the monitor and then begins its nested transaction.

**3. One-time log when the dictionary rolls over** — taking it, with one change: **INFO, not WARNING.** Nothing is wrong at that moment; rolling over is the healthy behaviour this PR exists to enable, and a warning on normal growth is how logs stop being read. The breadcrumb value you are after is fully preserved at INFO. Triggered on the creation of page 1 specifically, which is once per database lifetime and is exactly the moment downgradability is lost — pages 2 and beyond change nothing about that status, so logging them would be noise. The message names the database and states the consequence and the remedy.

**4. Emptied-not-truncated pages** — agreed on deferring, for the reasons in the earlier commit message.

Local: 626 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`. 0 failures, 0 errors.

lvca:
Round 7. No code changes needed: one point was a question, one is settled by measurement, one was already done. I have documented the answer to the first so it does not have to be re-derived.

**1. "Confirm not bumping the on-disk version on first rollover is preferred"** — confirmed, and I have written down *why* rather than leaving it as "intentional", because it does read like an oversight. Two reasons, and the second is the decisive one:

- The version is part of the **file name** (`dictionary.0.65536.v1.dict`). Changing it at runtime means renaming a live component file. `PaginatedComponent.rename` can only do that behind a full flush barrier (`waitAllPagesOfDatabaseAreFlushed`), and doing it from inside an append transaction would race replication — which ships pages by file id — and leave follower file names diverged from the leader's.
- **It would protect nothing.** `ComponentFactory` passes the parsed version straight through without validating it, so a build old enough to be at risk opens a `v1` file exactly as happily as a `v0` one. The guard added in this PR only helps a *newer* format meeting *this* reader; it cannot help an older reader meeting this format. Bumping would be real risk for zero protection.

Recorded at `CURRENT_VERSION` and in `docs/5560-dictionary-multipage.md`.

**2. `concurrentAppendsAcrossARolloverKeepEveryIdUnique` may warrant `@Tag("slow")`** — measured rather than estimated: **0.026s**, with the whole 12-test class at 0.198s. That matches the ~29µs-per-nested-transaction rate I measured earlier (1000 x 29µs = 29ms), so the estimate of ~1000 commits was right and the cost of a commit is simply much lower than it suggests. Even a 50x slower runner puts this at 1.3s.

Not resizing it either, and here the contrast with the sibling test is the point rather than an inconsistency: `DictionaryLimitsTest` was cut from 20,000 names to 3,000 because the extra volume proved nothing beyond crossing a page. In the contention test the volume **is** the mechanism — fewer operations means fewer chances for two threads to interleave at a rollover boundary, so trimming it would weaken exactly what it exists to catch, for 0.02s.

**3. `updateName` WAL cost and the schema-level lock advice** — both already in the method javadoc as of the previous round, along with `synchronized` to close the in-RAM race rather than only warn about it.

Local: 21 dictionary tests, 0 failures. Previous round's wider run was 626 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database` and `com.arcadedb.engine`, also clean; this round touched only a javadoc and a doc file.

That closes out everything raised across the seven review rounds. Remaining known tradeoffs, all deliberate and documented: emptied-not-truncated pages after an `updateName` shrink, the identifier cap on new databases, and the follower-upgrade ordering requirement.

lvca:
Round 8. All three items addressed in code, including the two you filed as follow-ups — item 1 because you are right that this PR is what makes it matter, so shipping the ceiling removal without it would be shipping the problem.

**1. `CopyOnWriteArrayList` append is now unbounded** — correct, and the sharpest observation in the whole review: the ~48K ceiling was the only thing bounding an O(n^2) append, and this branch removes exactly that ceiling. Fixed rather than deferred.

One correction on the proposed remedy, though. "A plain `ArrayList` swapped wholesale under the same volatile `Entries` pattern" would **not** have helped — republishing a copy per append is precisely the O(n) copy the COW list was already doing, just relocated. What removes it is **spare capacity**: `Entries` now holds a `String[]` plus the live `size`, grown by doubling. An append writes `names[size]` and publishes `size + 1`. Sharing the array with older readers is safe in both directions: a reader on the older snapshot holds the smaller size and never looks at that slot, and a reader that sees the new snapshot sees the slot through the volatile write's happens-before edge. Reads remain one volatile read plus an array index.

Measured, appending N names:

| N | CopyOnWriteArrayList | doubling array |
|---|---|---|
| 50,000 | 232 ms | 0 ms |
| 200,000 | 1,927 ms | 0 ms |
| 500,000 | **11,179 ms** | **2 ms** |

4x the names costs 8.3x the time and 2.5x costs 5.8x, which is the quadratic term showing through. Your "millions of distinct field names" scenario was heading for minutes of pure array copying.

`updateName` renames onto a copy rather than in place, deliberately: an append only ever writes slots *past* the published size, so it can share its array, whereas a rename rewrites a slot readers are already looking at and would need a happens-before edge they do not have. One copy per rename is nothing next to the full page rewrite it precedes.

**2. The breadcrumb can fire for growth that never commits** — a good catch, and worse than cosmetic in my view: the line's entire value is being trustworthy when correlated against a follower failure, and my own rolled-back-growth test was making it lie. It now registers an after-commit callback via `addAfterCommitCallbackIfAbsent`, which is discarded on rollback and keyed so several pages added in one transaction still log once. I preferred that over your fallback of "noting in the message that it reflects an in-progress transaction", since a caveat in the text does not make the signal usable.

**3b. No forward-direction compatibility fixture** — agreed, and a fair gap: `pageZeroKeepsTheLegacySinglePageLayout` only proved new-writer to old-reader. Added `aDictionaryLabelledWithThePreviousFormatVersionLoadsAndThenRollsOver`, which relabels a single-page dictionary to the previous format version, opens it with this code, checks it loads in full with no migration, then rolls it over past two more pages with every id it arrived with intact. Relabelling rather than synthesising a fixture is deliberate: the two formats are byte-identical for one page, and the version in the name is the *only* difference, which is precisely the claim under test.

**3a. No end-to-end HA test across a real page boundary** — agreed this is the sharpest edge, and I am **not** claiming it covered. `aReplicatedNewDictionaryPageIsVisibleAfterApplyChanges` exercises the reload-walks-all-pages behaviour at the unit level through the real `applyChanges` entry point, which is the part that could be wrong in this diff; what is untested is the full leader-ships-page-1-to-follower path. That belongs in `ha-raft` as a genuine two-node integration test, and I would rather file it than bolt a flaky cluster test onto this branch. Flagging it explicitly as a known coverage gap rather than quietly leaving it in the noise.

Local: 627 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`. 0 failures, 0 errors. `DictionaryMultiPageTest` is now 13 tests.

lvca:
Round 9. Point 2 turned out to be worth more than "flagging", and it produced a real fix.

**2. `createIfNotExists=true` can mask corruption** — you framed this as "empty page and missing page are now indistinguishable, not a regression". Following it through, the consequence is sharper than masking: **an invented page silently renumbers ids.** It contributes zero names, so every name stored after it comes back with an id lower by however many the missing page held, and those ids are embedded in records. That is the exact failure mode the append-only invariant exists to prevent, arriving through the read path instead of the write path.

So it is fixed rather than noted. `createIfNotExists` stays `true` for page 0, where it is load bearing (paired with the floor it materialises page 0 of a file killed mid-write, which is what the single-page reader did, and `EmptyDictionaryFileReopenTest` / `UnflushedDictionaryRecoveryTest` still pass). Every later page now passes `false`, and a missing one is reported as a truncated dictionary naming the page and the count that claimed it.

On reachability, since that governs whether the guard is theoretical: `pageCount` is derived from file size at load, so it cannot over-report there, and dirty-but-unflushed pages are served from the cache, so that is not it either. The way the count gets ahead of the file is a **partial replication replay** — `applyChanges` writes a page and calls `updatePageCount(pageNumber + 1)` without necessarily having written the pages before it. The new test reproduces that shape directly, in its own database because it deliberately leaves the count inconsistent, and drops it.

**3. The rollover breadcrumb is transient, so there is no durable signal** — accepted as stated, but there *is* a durable one worth naming rather than conceding: the dictionary is one page per `pageSize` bytes and `pageSize` is in the file name, so `ls -l <db>/dictionary.*.dict` answers "has this rolled over" at any time. 131072 / 65536 = 2 pages means rolled over and no longer downgradable; a file at or under `pageSize` is still single-page. Added to the docs as the thing to check before rolling a cluster back.

**1. `updateName` WAL amplification** — already in the method javadoc, and I am declining a cap. A cap on a rename is semantically odd (you cannot refuse a rename because the dictionary is large), and it would be complexity on a method with no production caller. For the record, the right optimization when a live rename is wired is not a cap but a narrower rewrite: pages *before* the first renamed occurrence are byte-identical and need not be touched, and if the new name has the same byte length as the old, only the one page containing it changes. That needs the index-to-page mapping the rewrite currently derives on the fly, which is worth building for a real caller and not before.

**4. The `getIdByName` safety net leaves the name committed while throwing** — agreed: pre-existing, rare, self-healing on the next reload. No change.

**Comment density** — noted, and I take the point that it is above the surrounding baseline. Keeping it, per your read, since it is carrying the concurrency and on-disk-format reasoning.

Local: 628 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`, run twice, 0 failures both times.

**Unrelated flake observed, for the record:** one full-suite run had `DatabaseGetSizeTest.getSizeMultipleCalls` fail with `expected: 131453L but was: 65917L`. It is not related to this branch (it reproduces neither on the previous commit of this branch nor on repeated runs of the current one, and passes in isolation). `LocalDatabase.getSize()` does `Files.walk` over the database directory and sums file sizes, so asserting three consecutive calls are byte-identical only holds if no background WAL rotation happens between them; the 65,536-byte delta is consistent with a segment being removed. Left alone as out of scope for this PR, but it is a genuine intermittent worth its own issue.

lvca:
Round 10. The actionable item done, and item 2 closed rather than documented.

**1. Dead `CopyOnWriteArrayList` import** — dropped. The `{@link}` became `{@code}` rather than moving to a fully-qualified link: the reference is historical ("it used to be a COW list"), so linking the JDK class bought nothing and an import that exists only for Javadoc is the smell you flagged. The prose comment in `reload()` needs no import.

**2. Map-vs-array ordering window** — you asked for a one-line note. I think the better answer is to make the note unnecessary, because the *order* was the bug, not the absence of documentation.

The two halves cannot be updated atomically together, so one leads by an instant either way. What matters is which. Leading with the map, as it did, means the reader in between resolves a name to an id the published names do not cover yet, and `getNameById()` **throws**. Leading with the names means that reader finds the name absent and gets **-1** — which every `create=false` caller already handles (`BinarySerializer` just skips the value compression) and which a `create=true` caller resolves by blocking on this monitor. And nobody can be holding the new id before the map publishes it, since the id only reaches a record after `getIdByName` returns, so the reverse exposure does not exist.

So: a miss costs nothing, a throw costs a request. Swapped, with the reasoning recorded at the site. You are right it was pre-existing and a couple of instructions wide either way; closing it was simply cheaper than describing it. Ran the 4-thread contention test three more times plus the full selection, all clean.

**3. The "under 1% wasted tail" claim covers the append path, not the post-shrink layout** — correct, and the imprecision was mine. Qualified in the class Javadoc, and it is slightly worse than you described: after a shrink, appends resume on the **last** page, which is one of the emptied ones, so every page between the last used one and that tail stays empty for good, not just the one partial tail. They cost one zero-entry read each on load and nothing else, and `updateName` still has no production caller.

Local: 628 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`. 0 failures. Plus `DictionaryMultiPageTest` three times over, since the ordering change lands in the concurrency path.

Standing item, unchanged from the previous round: there is still no end-to-end `ha-raft` test of a rolled-over leader shipping page 1+ to a follower. That is the sharpest operational edge of this change and I have deliberately left it as a follow-up rather than attach a two-node cluster test to this branch.

lvca:
Round 11. Item 1's premise needs a correction, item 2 is addressed, and I ran the verification you asked for.

**1. Public API removal of `getAvailableSpace()`** — the premise is not quite right, and it changes the conclusion. That method was not on `main` before this work: `git log -S` puts its first appearance in **`621313ce5`**, my own commit from earlier today, and `git tag --contains 621313ce5` is empty, so it is in **no release tag**. It was added and deleted inside this same body of work, a few hours apart. There is no released version that ever exposed it, so there is no downstream source-compatibility break and nothing to note in a changelog. (Removing it was itself a response to review round 1, which flagged it as dead.)

**2. The rollover signal is only a one-time INFO, easily missed after the fact** — this is the third round to land on the operational hazard, and the "easily missed after the fact" part is the fair core of it. Addressed: the dictionary now logs its state **on every open** when it already spans more than one page, naming the page count.

I deliberately did not take either of the two mechanisms you suggested:
- A **persisted marker** would be a second copy of a fact the page count already records durably, and two copies of one fact can drift. The file size *is* the marker.
- A **startup compatibility check** cannot help, because the build that needs warning is the *older* one. It predates the concept and cannot be given a check retroactively. No code in this PR can make an un-upgraded follower detect this.

Restating current state at open is what actually survives log rotation, which was the real complaint. The rollover event line stays as well, so an operator gets it both as it happens and on every subsequent start.

**Test note: confirm `SnapshotCompressionRatioTest` in `ha-raft` still passes with `65536 * 5` decoupled from `DEF_PAGE_SIZE`** — ran it: **12 tests, 0 failures**. The literal is intentional there; that test pins the compression-ratio guard against a sparse page-sized zip entry, and 327,680 remains the largest such entry a snapshot of a pre-change database can carry, so coupling it to `DEF_PAGE_SIZE` would have weakened it. I updated only its stale comment.

**3, 4, 5** — agreed as stated, no changes. The `updateName` empty-page leak, the identifier cap on new databases, and the version-not-bumped-on-rollover reasoning are all documented and unchanged.

### On the red CI checks, since they matter more than the notes

`integration-tests` is failing, and it is **not this branch**. The job's own steps show "Run Integration Tests with Coverage" **success** and "Check no test leaked a databases/ directory" **success**; the only failed step is `IT Tests Reporter`, the report-publishing action. The last three runs on `main` fail identically, with the same single step. `Meterian client scan` is also failing on `main`. Both are pre-existing infrastructure failures, not regressions from this PR. `unit-tests` was still running at the time of writing and is the check worth watching.

Local: 628 tests across `com.arcadedb.schema`, `com.arcadedb.serializer`, `com.arcadedb.database`, `com.arcadedb.engine`, plus the `ha-raft` test above. 0 failures.

lvca:
**Correction to my CI note above.**

In my last comment I said `integration-tests` was red only because of the `IT Tests Reporter` publishing step, and that the tests themselves passed. **That was wrong**, and the mistake was mine: I read the step conclusions and stopped there instead of reading the reporter's output.

The reporter is not broken. It is correctly reporting a genuinely failing integration test:

```
com.arcadedb.postgres.PostgresQueryMetricsIT.postgresQueriesTaggedWithPostgresProtocol
  Expecting actual: 0L to be greater than or equal to: 1L
  at PostgresQueryMetricsIT.java:76
##[error]Failed test were found and 'fail-on-error' option is set to true
```

What misled me is that the Maven step runs with **`--fail-never`**, so "Run Integration Tests with Coverage" reports success even when an IT fails, and the reporter is the only thing that turns the job red. The failure therefore *looks* like it lives in a publishing step when it does not.

What stands from my earlier note: this is **not a regression from this PR**. Run `30504641942` was on `main` at 01:05Z against `f2909e34`, before this merged at 01:59Z, and the failing test is Postgres query metrics with no relationship to the dictionary. The two claims that mattered for merging this PR - not caused here, and pre-existing on `main` - are unaffected. But "the integration tests pass" was not true, and anyone reading that comment as a green light on `main` would have been misled.

I am picking this up separately along with the other follow-ups.

Issue #5559, opened 2026-07-29
LSM_VECTOR: setting locationCacheSize silently truncates the index - count and search only see the last N vectors


lvca:
**Verified on pristine `main` (`2d5b0fed6`), found while working on #5516.** Not caused by #5521; filing separately.

`arcadedb.vectorIndex.locationCacheSize` (and the per-index `locationCacheSize` metadata) is documented as a memory bound: *"Maximum number of vector locations to cache in memory per vector index... Recommended: 100000 for datasets with 1M+ vectors"*. It does not behave as a cache. Setting it makes the index forget every vector beyond the limit: `countEntries()` under-reports and searches return the wrong neighbors.

### Reproduction

200 vertices, 16-dimensional INT8 index, `locationCacheSize: 10`:

```sql
CREATE INDEX ON Doc (embedding) LSM_VECTOR
METADATA { "dimensions": 16, "similarity": "COSINE", "quantization": "INT8", "locationCacheSize": 10 }
```

```
resident locations = 10   live (countEntries) = 10      <- 200 vectors were inserted
search with doc5's own embedding   -> doc192
search with doc100's own embedding -> doc194
search with doc199's own embedding -> doc199
```

Every probe should return the vertex whose embedding it is (the vectors are well separated: with the default unlimited cache each returns itself). Instead only the last-inserted handful are reachable - the rest are gone from both the count and the graph.

### Mechanism (my reading, not isolated with a debugger)

The bounded map is not used as a cache in front of an authoritative set - it *is* the authoritative set. `buildGraphFromScratch` parses the pages into the live set, then re-populates `VectorLocationIndex` and builds the jvector graph from what is resident; in bounded mode that re-population evicts back down to `maxSize`, so the graph is built over `maxSize` vectors and the rest are unreachable for good.

The intended rescue path looks broken too: an evicted id goes to `getVectorLocation()` -> `reconstructLocationFromPages()` -> `scanPagesForVectorId()`, which builds the offset as

```java
final long pageBaseOffset = ((long) fileId << 32) | (pageNum * getPageSize());
```

while `readVectorFromOffset()` and `LSMVectorIndexPageParser` treat the same field as a plain file offset (`pageNum * pageSize + PAGE_HEADER_SIZE + offsetInPage`) and derive the file from the `isCompacted` flag. So a reconstructed location carries an offset the readers cannot decode. I did not isolate this second part - it is a code-level observation that fits the behaviour, not a proven cause.

### Impact

[...]

PR #5562, opened 2026-07-29
feat(engine): an LSM_VECTOR index compacts itself once its file is mostly garbage

lvca:
Follow-up to #5521 / #5516: re-enables automatic compaction for `LSM_VECTOR`, which has been disabled since the feature landed.

## Why

The data file is append-only. An update writes a new vector plus a tombstone for the one it supersedes, a delete writes a tombstone, and nothing reclaimed either - so a workload that keeps re-embedding the same vectors grows the file for as long as the database lives while the live set stays the same size. That is the half of #5516 a heap dump cannot show: the reporter's memory bloat is fixed, but their 4K-vector index would still have walked to multi-GB on disk. Until now the only cure was an operator remembering to run `COMPACT INDEX`.

The old trigger was commented out with the note *"re-enable once compaction properly handles uninitialized pages"*. #5521 is what made that true.

## The trigger is the garbage ratio, not the page count

The disabled code fired on `currentMutablePages >= indexCompactionMinPagesSchedule` (10). That is right for an LSM-tree compaction, which is a cheap merge, and wrong for this one: a vector compaction rebuilds the graph and rewrites the file, so it has to pay for itself in reclaimed space rather than in pages touched.

It now fires when the file holds at least `arcadedb.vectorIndex.compactionBloatFactor` (new, default **3**) times the pages its live vectors need - "reclaim once about two thirds of the file is garbage" - still floored by the existing `indexCompactionMinPagesSchedule` so small files are left alone. Either setting at `0` disables it and leaves `COMPACT INDEX` as the only path.

Properties worth stating:

- **The check is free.** It runs after every commit and reads only counters already in memory (the component's page count, the number of resident locations). It never touches a page. The pages-for-live figure is derived from the entry layout `persistVectorWithLocation` writes, so it costs arithmetic.
- **The work is not on the committing thread, and neither is any waiting for it.** It goes to the async executor via `compact()` - the same replication-safe path an explicit `COMPACT INDEX` takes. Scheduling deliberately applies no back-pressure: the thread it would sleep is a committer whose work is already durable, so a full worker queue simply postpones the attempt to the next commit instead.
- **Followers do not schedule at all.** `runWithCompactionReplication` already declines on a non-leader, but only from inside a task that had to be queued, run and reset - and a follower taking writes is exactly the node whose garbage ratio keeps crossing the threshold. The leader check now happens before scheduling, mirroring what `TimeSeriesMaintenanceScheduler` does for the same reason.
- **It cannot pile up.** `scheduleCompaction()` flips AVAILABLE -> COMPACTION_SCHEDULED, so a commit arriving while one is pending is a no-op instead of a queue of redundant rewrites. Every path that then fails to start the compaction hands that slot back, or the index would never compact again.
- **A closing database does not fail a commit.** The whole hook is guarded: it runs inside the commit, before it is marked committed, so anything thrown there would fail a transaction whose data is already durable. On a closing database that is logged at FINE, elsewhere at WARNING.
- **The ratio only trusts a location map that does not evict.** `size()` is the live-vector count while the map keeps every live location, which is the case since #5568 refused `locationCacheSize` (an evicted location cannot be recovered). The guard asks the map whether it evicts rather than reading that setting, so it stays correct by itself if bounded mode is ever wired back in - a capped map would read as permanent bloat and rewrite the file after nearly every commit.

For the shape in #5516 (4K vectors, 384 dimensions, INT8): the live set is ~7 pages, so it compacts once the file passes ~21 pages - roughly every second re-embedding cycle, reclaiming a few MB each time, with nobody having to ask. For a 1M-vector index the same ratio means rarer, larger passes; operators who want fewer, bigger rewrites raise the factor, and the config documents that each pass holds the index write lock for the copy.

## Tests

`LSMVectorIndexAutoCompactionTest`, eleven cases, each verified to fail with its own fix reverted rather than merely to pass:

*The trigger*

1. an update-heavy index compacts itself with no explicit command, keeps every live vector, still answers searches correctly and survives a reopen
2. an index with no garbage is left alone (inserts only - rewriting would be pure cost)
3. the scheduler can be turned off, and then nothing compacts however bloated the file gets
4. a churning index reaches a steady state instead of rewriting on every cycle
5. a delete-heavy index reclaims too, and the survivors are all still there
6. NONE quantization also compacts - the smallest page entry the size estimate has to handle, and the branch with the least obvious arithmetic

*Who compacts, and what happens when it cannot*

7. the ratio only trusts a location map that does not evict (and a configured cap does not produce one, per #5568)
8. a compaction that gives up before starting (a backup's flush suspension) releases its slot
9. an attempt whose task never reaches a worker releases its slot - the reservation is made before the hand-off, so an early-out there has to undo it
10. a follower does not schedule; a leader and a standalone database do
11. a legacy compacted companion left on disk is not loaded, so it cannot skew the page accounting - the companion extension has never been in `SUPPORTED_FILE_EXT`

They use a 16KB page size so that 400 live vectors already exceed the min-pages floor: without that, several would pass trivially because the file was too small to bother with, and only the floor would be under test rather than the ratio.

Also run: `com.arcadedb.index.**`, `com.arcadedb.database.async.**` and `com.arcadedb.index.lsm.**` (309 tests, 0 failures) and `RaftIndexCompactionReplicationIT` (6 tests, 0 failures) - the latter matters because compaction can now fire from a commit hook on a Raft leader, which is exactly where #5521 found a divergence.

Rebased on main at `a1dc09616`: `calculateQuantizedSize` had been removed and #5568 stopped honouring `locationCacheSize`, both after this branch was written. Reconciled and re-run against the merged tree.


lvca:
Blocker confirmed and fixed in f8e4fe6b, along with both smaller notes.

**1. Bounded location cache - you were right, and it is worse than a bad estimate.**

Fixed by exempting bounded-cache indexes from automatic compaction: they now leave it to an explicit `COMPACT INDEX`, which derives its live set by parsing the pages rather than trusting the resident map, so it stays correct in that mode. I checked the guard is load-bearing rather than defensive - removing it makes the new test fail with the file swapped `5 -> 7` while churning a capped index, exactly the every-commit rewrite you described.

One correction to the premise, because it matters for why this is the right shape of fix rather than "compute a better estimate": in bounded mode the index does not hold 1M live vectors with 100K cached. It holds 100K. `countEntries()` under-reports, and searches return the wrong neighbours, because the graph is built over the resident set. I verified that on pristine `main` while working on #5516 and filed it as **#5559** - so the configuration your example describes is already broken independently of this PR. That is also why "drive the estimate from a persisted live count" is not available today: there is no true live count in bounded mode to read. The guard now points at #5559 in a comment so the two stay connected, and if #5559 is fixed by making that setting an honest cache, the exemption can be lifted.

**2. The varint "upper bound" comment** - correct, it is not one: zig-zag takes a 32-bit id to 5 bytes and a 64-bit position to 10. It is a typical size, it biases the trigger slightly eager, and the comment now says that instead of claiming a bound it does not hold. I left the numbers as they are, since the estimate feeds a ratio against a configurable factor.

**3. Mixed cached vs live config reads** - rather than document the asymmetry I removed it. `minPagesToScheduleACompaction` was cached at construction while the bloat factor was read live, so one knob honoured a runtime change and the other silently did not. Both are read per commit now, and the config text says so.

**Test coverage**, both gaps closed:
- `aBoundedLocationCacheDoesNotTriggerCompactionOnEveryCommit` - covers #1, and fails without the guard.
- `aChurningIndexReachesASteadyStateInsteadOfRewritingEveryCycle` - your point (b). 12 churn cycles must reclaim at least once and well under once per cycle. I measured what it actually observes rather than trusting the bound: **1 compaction over 12 cycles**, so the assertion has real margin instead of passing vacuously.

Vector package after the change: 219 tests, 0 failures.

lvca:
All three folded in - 6a5e7d11. Thanks for the careful read; (2) turned out to matter more than "low".

**2. Stuck `COMPACTION_SCHEDULED` - fixed here rather than filed.** You framed it as pre-existing and low, and the first half is right, but the consequence is bad enough that shipping auto-scheduling on top of it did not seem defensible: the slot stays reserved, and because `CompactIndexStatement` schedules before compacting, the *manual* escape hatch dies with the automatic one until the database is reopened. So an operator who hits a backup window loses the ability to reclaim at all, with no error to tell them. Every exit before the work starts now releases the slot, covering the read-only and invalid-index paths too, not just suspended flushing.

I checked the test earns it: with the release removed, `acompactionThatGivesUpBeforeStartingReleasesItsSlot` fails on the second `scheduleCompaction()` returning false. It drives the real path - `suspendFlushAndExecute`, i.e. what a backup does - rather than poking the status field.

**1. Log noise** - agreed, and it is this PR's doing that it matters. The per-invocation lines (`compact() called for index`, the status-transition line, the CAS-failed line, the timeout path) are FINE now. The one line left at INFO is the summary of what a compaction actually did: which index, how many live vectors, how many pages, old file -> new file. One line per reclaim, which is the thing an operator wants in the log.

**3. Coverage gaps** - both added. `anIndexWithoutQuantizationAlsoCompactsItself` covers the `NONE`/`PRODUCT` branch where the entry is just the header (it lowers `indexCompactionMinPagesSchedule` so the ratio, not the floor, is what decides - otherwise header-only entries need ~9K of them to reach the default floor and the test would be slow for no gain). `aDeleteHeavyIndexCompactsItself` reaches the same tombstones by the other route.

Vector package: 222 tests, 0 failures.

lvca:
Both code points are in f0a64758; the third is a release-notes item rather than a change, and I have left it for @lvca to place.

**1. Broad catch at FINE** - taken. It now logs at WARNING while the database is open and stays at FINE only on the way down, so the routine shutdown case is quiet while "this index has silently stopped reclaiming" is not. I kept the catch broad on purpose: the contract that matters is that a failure to *schedule* can never fail a commit that already succeeded.

**2. Javadoc nuance** - fixed. It now says counters and configuration lookups already in memory, rather than claiming counters only. The "never touches a page" part was the load-bearing claim and stays true.

**3. Bounded-cache exemption in the release notes** - agreed, and it is the one thing here an operator could be bitten by without any signal in their own logs. I have not touched a release-notes file (repo convention is that those are written deliberately, not by whoever happens to be in the area), so proposing the wording instead:

> LSM_VECTOR indexes now compact themselves once their data file grows past `arcadedb.vectorIndex.compactionBloatFactor` (default 3) times the size their live vectors need, reclaiming the space that updates and deletes leave behind. Indexes configured with a bounded `arcadedb.vectorIndex.locationCacheSize` are excluded - their live-vector count is capped by the cache, so the ratio cannot be computed - and continue to require an explicit `COMPACT INDEX`. See #5559.

Worth saying plainly why that exclusion stings: the large churning index this feature is for is also the one most likely to be running with a bounded cache on someone's advice, so precisely the workload that needs relief may not get it automatically. The real repair is #5559 - make that setting an honest cache instead of something that truncates the index - and once it is, the exemption can be deleted rather than documented.

Vector package: 222 tests, 0 failures.

lvca:
Item 1 was a real hole in my own fix - thanks for following the slot across the boundary. All three in afae3cc9.

**1. The slot leaks where it is taken, not where it is released.** You are right, and the version I shipped was half a fix: `scheduleCompaction()` reserves the index inside `DatabaseAsyncExecutorImpl.compact()`, one level above the `compact()` whose exits I had covered. `scheduleTask(..., waitIfQueueIsFull=false, ...)` answers `false` on a full queue without throwing, so nothing runs the compaction, nothing releases the slot, and - as you note - not even the new WARNING fires because there is no exception to catch. The index then silently stops compacting, `COMPACT INDEX` included, until a reopen.

Fixed at the reservation site, so the release cannot be missed by whichever way the enqueue fails, including the throwing paths (`getBestSlot()` on a shut-down executor). Since `LSMTreeIndexMutable.onAfterCommit` routes through the same method, this closes the same latent leak for LSM-tree indexes too.

The new test drives the shut-down path, which is the one reachable deterministically: it asserts both that the caller is told (`DatabaseOperationException` - `onAfterCommit` is what turns that into a log line instead of a failed commit) and that the index is schedulable again afterwards. With the release removed it fails on the second `scheduleCompaction()` returning false. I did not add a queue-full variant: forcing a full worker queue deterministically needs a blocked worker and a shrunken queue, and the flakiness that invites costs more than it proves, since both paths leave through the same `finally`. Saying so explicitly rather than implying the branch is covered.

**2. `isCompactionDue()` outside the net** - correct, and the reasoning is the part that matters: the hook runs inside the commit before it is marked committed, so a throw there fails a transaction whose data is already durable and trips the #5053 fence. The whole body is guarded now, decision included; the comment says why rather than just noting that it is.

**3. Indentation** - fixed; the inner block had kept its old column after being wrapped.

Tests: 295 across the vector, async-executor and LSM-tree index packages, 0 failures.

lvca:
Thanks - acted on both of the points you flagged as worth it, and on the back-pressure caveat too rather than just re-wording it. `8b0a7d1d1`.

**1. Back-pressure on the committer.** You are right that "the check is free" was doing some work it had not earned: the check is free, the *scheduling* was not. Fixed rather than reworded - `DatabaseAsyncExecutorImpl.compact()` now passes `0` instead of `backPressurePercentage`.

Back-pressure exists to slow down whatever is filling the queue. Both callers of this method are `onAfterCommit` hooks, so the thread it would sleep is a committer whose work is already durable: sleeping it throttles nothing, it only adds latency to a commit that is finished. With `0` a full queue makes the `offer` give up, the `finally` from the last round hands the slot back, and the next commit past the threshold schedules again - both gates are level-triggered (`currentMutablePages >= min` for LSM-tree, the ratio here), so no compaction is lost, only postponed. Strictly better than sleeping and then very likely failing the same `offer`.

This changes LSM-tree index compaction scheduling too, in the same direction and for the same reason.

I did not add a test: the only observable difference is that a commit does not sleep, and asserting that means a wall-clock bound. The give-up path itself is already covered by the two slot-release tests.

**2. Follower churn.** Agreed, and gated before scheduling:

```java
static boolean isCompactionAllowedOnThisNode(final DatabaseInternal database) {
  return !database.isReplicated() || database.isLeader();
}
```

Same shape as the check `TimeSeriesMaintenanceScheduler.runMaintenance` makes, and for the same reason it gives: the engine operation self-gates anyway, but skipping early avoids the churn. New test `aFollowerDoesNotScheduleCompactionsTheLeaderWillSendItAnyway` asserts follower -> false, leader -> true, standalone -> true, using the delegating `Proxy` over `DatabaseInternal` that `Issue5470ReplicatedCommitEveryTest` already uses to flip `isReplicated()` - no mocking framework in the engine module. Verified it fails with the gate reverted to `return true` ("Expecting value to be false but was true"). The wiring into `isCompactionDue()` is covered by inversion: every other test in the class runs standalone and compacts, which they could not if the gate answered wrongly for a non-replicated database.

`RaftIndexCompactionReplicationIT` still green (6 tests) - that is the one that caught a divergence in #5521, and it is the leader path this touches.

**3. `size()` semantics.** Real latent trap. Rather than only commenting at the call site I fixed the source: the `size()` Javadoc said "including deleted", which has not been true since `markDeleted()` started dropping the location and keeping only the id in the tombstone set. It now states that this is the live count on the unbounded backend, that `estimatePagesForLiveSet` depends on it, and that a bounded cache makes it a lower bound instead - which is exactly why `isCompactionDue()` sends bounded caches away before the estimate runs. The call site notes why counting the values instead is not an option here (a full map scan on every commit).

**Nits.** PR description rewritten - it now lists all ten tests and no longer claims three, and the back-pressure and follower properties are stated. `aChurningIndexReachesASteadyState`'s `CYCLES/2` bound I have left as is: it is comparing "compacted a handful of times" against "compacted every cycle", so the margin is a factor, not a few percent. If the entry-size estimate ever drifts far enough to threaten it, that bound firing is information worth having rather than noise.

Full run: `com.arcadedb.index.**` + `com.arcadedb.database.async.**` + `com.arcadedb.index.lsm.**`, 300 tests, 0 failures.

lvca:
Thanks - the page-accounting point is correct, and chasing it turned up something bigger than the finding itself. `dc12d91bb`.

**The asymmetry is real, and fixed.** `getTotalPages()` is the mutable file alone; `estimatePagesForLiveSet()` is built from `vectorIndex.size()`, which includes the companion's locations (`loadVectorsFromFile` at LSMVectorIndex.java:2492 pulls them into the same map). Comparing a part against the whole under-triggers exactly as you describe. Both sides now cover the same files:

```java
private int totalPagesForBloatRatio() {
  return getTotalPages() + (compactedSubIndex != null ? compactedSubIndex.getTotalPages() : 0);
}
```

**But you asked how often an index actually runs with a non-null `compactedSubIndex`, and the answer is never.** `LSMVectorIndexCompacted.FILE_EXT` is `"vcidx"`, and that string appears in exactly three places in the whole engine: its own declaration and the two lookups in `LSMVectorIndex`. It is **not** in `LocalDatabase.SUPPORTED_FILE_EXT`, so `FileManager` never opens such a file, and it is not registered with `LocalSchema`'s component factory either. `git log -S` says it has never been in that set - not since the index landed in #2816, so this is not something #5521 broke.

I verified it rather than inferring it. New test `aLegacyCompactedCompanionOnDiskIsNotLoadedIntoTheRatio` builds an index, closes the database, drops a companion on disk named exactly the way `discoverAndLoadCompactedSubIndex()` scans for (`<mutable component>_<timestamp>...vcidx`), and reopens:

- no `vcidx` component in the FileManager
- `compactedPages` = 0
- the file still sitting there on disk, untouched

So the fix is defensive and the test is what says so. I also checked the inversion: adding the extension to `SUPPORTED_FILE_EXT` does not merely make the test fail, it makes the database fail to open (`Error on creating new database instance`) - there is no factory handler for it. The companion is thoroughly dead, not dormant.

Which raises a follow-up outside this PR's scope: `LSMVectorIndexCompacted`, the WAL routing for it in `TransactionManager`, and `discoverAndLoadCompactedSubIndex()` are all machinery for a format nothing can produce or load. That discovery runs on **every index open** and includes a 1000-iteration `getFileByIdIfExists` loop with an exception per miss. Worth either wiring up or deleting; I would rather do that deliberately in its own change than smuggle it in here. Happy to file it.

**Minor points**

- `e.getMessage()` -> `e.toString()`. Correct catch: the one line that says an index stopped reclaiming on its own must not read `... : null`.
- **Low-RAM profile opting out silently.** Agreed that this is the operationally dangerous one, so it is no longer silent - an index exempted for its bounded cache now says so once, at INFO, naming the setting and pointing at `COMPACT INDEX`. Once per index, from a commit hook, so a plain boolean field: two lines on a race beats a lock on that path. The underlying gap stays #5559's.
- **`getLocationCacheSize() > 0` vs `<= 0`.** I do not think there is anything to change here: `if (size > 0) return false;` already means "proceed only when `<= 0`", so the guard you suggest is the one that is there. And it is not an accident of sign - `VectorLocationIndex` picks its backend with the identical `maxSize > 0` predicate (VectorLocationIndex.java:186, and again in `getActiveCount()`), so a literal `0` is unbounded in both places by construction. Special-casing `-1` here would introduce the divergence rather than remove it.

Tests: 297 across `com.arcadedb.index.vector.**`, `com.arcadedb.database.async.**` and `com.arcadedb.index.lsm.**`, 0 failures.

lvca:
Both blocking items are incorrect, and both are decidable by looking at the file rather than by argument, so here is the evidence. The review closes by asking me to confirm with a full build - I did, from clean.

**Blocking 1 - `calculateQuantizedSize` does exist.**

```
$ grep -rn "calculateQuantizedSize" engine/src/main
LSMVectorIndex.java:4350:    case INT8 -> entrySize += 4 + calculateQuantizedSize(...) + 8;
LSMVectorIndex.java:4351:    case BINARY -> entrySize += 4 + calculateQuantizedSize(...) + 4;
LSMVectorIndex.java:5802:            final int quantizedSize = calculateQuantizedSize(...);
LSMVectorIndex.java:5825:  private int calculateQuantizedSize(final int dimensions, final VectorQuantizationType quantizationType) {
```

It is a private method in the same class, and it predates this PR (call site 5802 is not mine). It needs no import, which is presumably why a grep for a declaration went looking in the wrong shape.

The rest of that item is right about what the estimate needs and wrong about what it has. `calculateQuantizedSize` returns the array length **only** - `dimensions` for INT8, `(dimensions + 7) / 8` for BINARY, `0` otherwise. That is exactly the "quantized array length only, no existing helper returns that" the review describes, so there is no double-count: the `+ 8` / `+ 4` for the min/max/median bytes are added once, matching `persistVectorWithLocation`. `LSMVectorIndexPageParser.calculateQuantizedDataSize` is a different method that I do not call.

**Blocking 2 - `getLocationCacheSize()` is not hardwired to -1.**

```java
private int getLocationCacheSize() {                                       // LSMVectorIndex.java:5360
  if (metadata != null && metadata.locationCacheSize > -1)
    return metadata.locationCacheSize;                                     // per-index metadata
  return mutable.getDatabase().getConfiguration()
      .getValueAsInteger(GlobalConfiguration.VECTOR_INDEX_LOCATION_CACHE_SIZE);   // or the global setting
}
```

There is no setter that logs and ignores the value; nothing forces -1. `-1` is the default, not the only reachable answer. The strongest internal evidence is in the previous review on this same PR, which flagged that the built-in low-RAM profile sets `VECTOR_INDEX_LOCATION_CACHE_SIZE = 10_000` (GlobalConfiguration.java:128) - an observation that would be meaningless if the value could never be positive.

The cited line numbers (5444 / 5426) do not correspond to those methods in this file, so I think this was read against a different revision.

**Verification, since the review asks for it.** Clean build and full class, on the pushed commit:

```
$ mvn -o -pl engine clean test-compile        ->  BUILD SUCCESS
$ mvn -o -pl engine test -Dtest=LSMVectorIndexAutoCompactionTest
   Tests run: 11, Failures: 0, Errors: 0, Skipped: 0   ->  BUILD SUCCESS
```

`aBoundedLocationCacheDoesNotTriggerCompactionOnEveryCommit` passes. And to answer the "dead code" half directly rather than by assertion, I replaced the guard with `if (false)` and re-ran it:

```
LSMVectorIndexAutoCompactionTest.aBoundedLocationCacheDoesNotTriggerCompactionOnEveryCommit:195
  [an index whose live count is hidden by a bounded cache must not compact itself]
```

The branch is live: remove it and the test fails. (Marker removed, guard restored.) Every test in this PR was checked that way - passing is not evidence, failing when its fix is reverted is.

**Minor points - both fair, one acted on**

- **Non-volatile `compactedSubIndex` read.** Agreed, and now documented (`83a080079`): the read feeds a heuristic, so a stale answer moves one compaction by one commit and nothing else, which is not worth a volatile read on the commit path. Worth adding that this is doubly moot - as established in the previous round, a companion can never be loaded at all.
- **Per-commit config lookups.** Two `getValueAsInteger` map lookups plus `isReplicated()`/`isLeader()`, and yes, on every commit that touches a vector index. That is deliberate and the config text says so: reading the settings live is what lets either knob take effect at runtime, whereas the cached `minPagesToScheduleACompaction` only reflects the value the index was built with. I would rather pay two hash lookups per commit than ship a setting that silently needs a reopen. The "counters already in memory" phrasing in the description is about the page and location counts, which is why the sentence names the settings separately as read live - happy to sharpen it further if it still reads as overclaiming.
- **Dropping the bounded-cache branch**: moot, since its premise was blocking 2.

Thanks for the careful read of the slot-reservation and entry-layout parts - those are the areas where I most wanted a second pair of eyes.

lvca:
You are right on both counts, and my previous reply was wrong. Correction first, since I argued the opposite an hour ago.

I checked `calculateQuantizedSize` and `getLocationCacheSize` in my worktree, saw both alive, and concluded the review had been done against some other revision. It was the other way round: **my branch was 19 commits behind main**, and the review was reading the merge result, which is the thing that matters. `calculateQuantizedSize` is gone from main, and #5568 (`de1df644a`) made `getLocationCacheSize()` return -1 unconditionally. Both blocking items were correct; I dismissed them by checking a stale tree and calling that verification. Sorry for the noise, and thanks for restating the second one with the issue number attached - that is what made me look in the right place.

Merged main in and confirmed the failure exactly as predicted:

```
LSMVectorIndex.java:[4393,35] cannot find symbol
  symbol:   method calculateQuantizedSize(int,VectorQuantizationType)
```

Fixed in `a1dc09616`.

**Blocking 1 - the entry-size estimate.** Now uses `LSMVectorIndexPageParser.calculateQuantizedDataSize`, and you were right that it already includes the min/max/median bytes, so the two branches collapse into one and the separate `+ 8` / `+ 4` go away:

```java
final int quantizedSize = LSMVectorIndexPageParser.calculateQuantizedDataSize(metadata.dimensions,
    metadata.quantizationType);
final int entrySize = 4 + 4 + 8 + 1 + 1 + (quantizedSize > 0 ? 4 + quantizedSize : 0);
```

The total is unchanged - INT8 was `4 + dimensions + 8` and is now `4 + (dimensions + 8)`; BINARY was `4 + (dims+7)/8 + 4` and is now `4 + ((dims+7)/8 + 4)` - so this is the same number arrived at through the reader's own arithmetic rather than a second copy of it. PRODUCT and NONE hit the helper's `default -> 0` and stay header-only, as before. Verified against `persistVectorWithLocation` on the merged tree.

**Blocking 2 - the bounded-cache exemption.** Gating on `getLocationCacheSize() > 0` guarded a mode that #5568 removed, and the test asserting a capped resident map could not pass. But I did not simply delete it, because the invariant underneath is still load-bearing: the ratio is only sound while `size()` is the live count, and that holds because the map keeps every live location. So the guard now asks the map instead of the setting:

```java
if (vectorIndex.isBounded())
  return false;
```

`VectorLocationIndex.isBounded()` is `maxSize > 0` - always false today, and true again the moment anyone wires bounded mode back in, which is what #5568's own javadoc warns is only safe once a miss is recoverable. That is a property of the object rather than a config value that no longer decides anything, so it cannot go stale the way the old guard did. The `boundedCacheExemptionLogged` field and its INFO log are gone (#5568 already warns when the setting is refused, so a second message about a mode that does not exist would be worse than silence), and the sentence about it in the `compactionBloatFactor` config text is gone too - you were right that it described something unreachable.

The test is rewritten to pin what is actually true, `theRatioOnlyTrustsALocationMapThatDoesNotEvict`: configuring a cap does **not** produce an evicting map (`isBounded()` false, `size() == VERTICES`), such an index therefore does compact under churn, and a directly constructed `new VectorLocationIndex(16)` does report `isBounded()`. If bounded mode returns, that last pair is what stops the ratio quietly becoming a rewrite-per-commit loop.

**Minor - `compact()`'s outer CAS.** Agreed on all counts, including that it is unreachable through the current paths since both reserve first. Left as is; the CAS is doing exactly the job you describe.

**Verification, on the merged tree this time**

```
$ mvn -o -pl engine test -Dtest=LSMVectorIndexAutoCompactionTest
   Tests run: 11, Failures: 0, Errors: 0
$ mvn -o -pl engine test -Dtest='com.arcadedb.index.vector.*Test,com.arcadedb.database.async.*Test,com.arcadedb.index.lsm.*Test'
   Tests run: 309, Failures: 0, Errors: 0
```

PR description updated for the merge and for the reworded property. The lesson I am taking from this one: "I ran the tests" means nothing if the tree they ran against is not the tree the change will live in.

PR #5563, opened 2026-07-30
test(ha-raft) #5560: replicate a dictionary that crosses a page boundary

lvca:
Follow-up to #5560, closing the one coverage gap that three review rounds on that PR flagged and I deferred each time.

## What was missing

#5560 let the schema dictionary spill past page 0. Dictionary pages replicate as raw pages through `TransactionManager.applyChanges`, and the post-apply `reload()` has to walk **every** page rather than page 0 - otherwise a follower ends up with an in-RAM dictionary missing every name past the first page and fails each record referencing one with `Dictionary item with id N is not valid`.

That path shipped covered only by a unit test feeding a hand-built `WALFile.WALPage` to `applyChanges`. That is a fair check of the reload behaviour, but it is not the scenario the upgrade-ordering requirement in `docs/5560-dictionary-multipage.md` is actually about, and it is the sharpest operational edge of the whole change.

## What this adds

`Issue5560DictionaryRolloverReplicationIT`, a three-node cluster test:

- the leader writes enough distinct property names that **ordinary inserts** push the dictionary off page 0 - the rollover is a side effect of normal writes, as it is in production, rather than a direct call into the dictionary
- every server must then report more than one dictionary page
- every server must resolve all 400 names in **both** directions (`getIdByName` is what a write needs, `getNameById` is what every read of these records needs, and only the second fails when a follower stopped at page 0)
- every server must read all the records back, which is the user-visible symptom: property names live inside records as dictionary ids, so deserialising them is what turns a missing page into `Dictionary item with id N is not valid`

## Verification

Rather than assert it passes, I checked it fails for the right reason. With `reload()` pinned back to `final int totalPages = 1` - the pre-multi-page single-page reader - the test fails; with the page walk restored it passes.

```
pinned to 1 page:  Tests run: 1, Failures: 1   BUILD FAILURE
page walk:         Tests run: 1, Failures: 0   BUILD SUCCESS   (19.38s)
```

Tagged `@Tag("slow")` at ~19s, per the CLAUDE.md guidance, since it stands up a real three-node cluster.

## Note on the padded names

Property names are padded to 500 bytes so a few hundred cross a page instead of a few hundred thousand. Long, but legal: only type and property names ever enter the dictionary. The goal is to reach the boundary quickly, not to look typical.


lvca:
Both notes applied in `c2e2ea390`.

**1. Implicit `PROPERTIES % TYPE_RECORDS == 0`** — a real trap, and worth more than the comment or assertion you suggested. Documenting a constraint leaves it representable; I made it unrepresentable instead. `PROPERTIES` is now **derived**: `TYPE_RECORDS * PROPERTIES_PER_RECORD`, with the arithmetic using the per-record constant directly. There is no pair of numbers a future editor can choose that desynchronises the writer from the follower loop.

Your reading of the failure mode was exactly right, and it is the nastiest kind for this particular test: truncation would make it report *missing names that were never written*, in a test whose entire job is to be trusted when it says a name is missing.

**2. Assertion label abbreviation** — fixed, though not by printing the real key. A 500-character padded name in a failure message is unreadable, so the label now reads `property #%d`, which is honest about being an index rather than looking like a key that grep will not find.

Also confirmed from the CI logs of the previous run that the test does what it claims in CI, not just locally: `Issue5560DictionaryRolloverReplicationIT` — `Tests run: 1, Failures: 0`, 12.27s.

### On this PR's red checks, none of which are this test

- **`ha-integration-tests`**: `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied` fails with `[Node 2 must hold both vertices] expected: 2L but was: 1L` plus a suppressed `DatabaseComparator` page-version divergence. It **ran at 02:35:23, before my test started at 02:39:37 in the same job**, so it cannot have been caused by it, and `ha-integration-tests` passed on the most recent completed `main` run. Looks like a flake in the phase-2 ticket area.
- **`integration-tests`**: `PostgresQueryMetricsIT`, pre-existing on `main` and filed as #5565 with the diagnosis.
- **`Meterian client scan`**: also failing on `main`.

Worth repeating, since I got this wrong myself earlier in the #5560 review loop: these jobs run Maven with `--fail-never`, so the test step reports success even when a test fails and only the reporter step goes red. Read the reporter's log, not the step conclusions.

PR #5564, opened 2026-07-30
test(engine): stop asserting getSize() is byte-stable across calls

lvca:
Found while triaging why `integration-tests` is red on `main`. Unrelated to the dictionary work that led me there.

## The race

`DatabaseGetSizeTest.getSizeMultipleCalls` asserted three consecutive `getSize()` calls return an identical number:

```java
assertThat(size1).isEqualTo(size2);
assertThat(size2).isEqualTo(size3);
```

`LocalDatabase.getSize()` is `Files.walk(databaseDir)` summing every regular file. WAL segments are created and dropped in that directory by background threads, so the method measures a moving target rather than being a pure function. Asserting byte-equality across three walks is a race by construction, and it fired:

```
DatabaseGetSizeTest.getSizeMultipleCalls
  expected: 131453L
   but was: 65917L
```

A delta of exactly 65,536 - one WAL segment removed between two calls.

## The fix

Assert what the method actually promises: repeated calls keep working and keep reporting a plausible size. Plus a same-order-of-magnitude check between neighbours, which still catches a `getSize()` that miscounts or returns a partial sum, while tolerating a segment appearing or disappearing.

I deliberately did not "fix" this by quiescing the database first. Nothing the test can do stops WAL rotation for the duration of three walks, so that would have made the race narrower and rarer rather than absent - the worst outcome for a flake.

Ran the class three times locally, clean each time. Note this is a rare intermittent, so a green run here is weak evidence either way; the argument is that the assertion is unsatisfiable in principle, not that it now passes.

### DIFF
--- a/engine/src/test/java/com/arcadedb/database/DatabaseGetSizeTest.java
+++ b/engine/src/test/java/com/arcadedb/database/DatabaseGetSizeTest.java
@@ -36,7 +36,6 @@ import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.assertj.core.api.Assertions.withinPercentage;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 /**
@@ -154,7 +153,7 @@ class DatabaseGetSizeTest extends TestHelper {
    * getSize() sums the files under the database directory, and WAL segments are created and dropped there by background threads,
    * so it is a measurement of a moving target rather than a pure function. Asserting three consecutive calls are byte-identical
    * was therefore racy: it was seen failing on CI with 131453 then 65917, a delta of exactly one 64Kb segment. What the method
-   * does promise is that repeated calls keep working and keep reporting a plausible size, which is what this now checks.
+   * does promise is that repeated calls keep working and keep reporting a plausible size, which is what this checks.
    */
   @Test
   void getSizeMultipleCalls() {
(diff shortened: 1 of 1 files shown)


lvca:
You are right, and this was a real defect in my fix rather than a nit. Fixed in `89c04f084`.

I checked your arithmetic against AssertJ's contract and it holds: `isCloseTo(expected, withinPercentage(p))` measures the allowed delta as `p% of |expected|`, so `isCloseTo(size1, ...)` based the band on `size1`. With the numbers from the observed failure:

| direction | base | delta | tolerance | outcome |
|---|---|---|---|---|
| segment disappears (what I saw) | 131,453 | 65,536 | 65,726 | passes by ~190 bytes |
| segment appears (equally likely) | 65,917 | 65,536 | 32,958 | **fails** |

So I had closed the flake in exactly the one direction I happened to have a sample for. That is worse than not fixing it, because the next occurrence would have looked like a new problem.

**Fixed by comparing the extremes of the three samples** rather than adjacent pairs with a percentage:

```java
final long smallest = Math.min(size1, Math.min(size2, size3));
final long largest  = Math.max(size1, Math.max(size2, size3));

assertThat(smallest).isGreaterThan(0L);
assertThat(largest).isLessThan(smallest * 10);
```

This is symmetric by construction, so there is no question of which argument the tolerance is relative to - the trap I fell into, and one a future reader would have to re-derive. It is closer to your option 1 than option 2: I did not want to teach this test the WAL segment size, because that couples an API test to storage internals and would need updating if segment sizing ever changes.

Your point that the code and the comment disagreed was also fair - 50% is nowhere near an order of magnitude. The bound now genuinely is one, and the comment says what the code does, including *why* the extremes are compared. Comments moved to sentence case to match the rest of the file.

On the strength of the evidence: as noted in the description, three green local runs prove very little about a rare intermittent. The argument for this change is that the original assertion was unsatisfiable in principle, and the argument for this revision is the asymmetry table above, not the passing runs.

PR #5566, opened 2026-07-30
fix(engine) #5279: a record that grows inside its page no longer breaks the disjoint-slot merge

lvca:
Closes the second half of #5279.

## The problem

After concurrent inserts stopped fighting for the same slot (`72d70d8f5`), the reporter's
`testPageConcurrentModificationException2()` still failed: N transactions updating N **different** records
that merely share a bucket page.

The disjoint-slot merge (#5381) replays a transaction's per-slot writes on top of the newer committed page,
but it only accepted an overwrite **of the same size or smaller**. Any update that made the record *grow* -
a longer string, one more property - was treated as a change too wide to rebase and poisoned the page.
Growth is the normal update shape, so in practice concurrent updates of unrelated records conflicted for
good, and a retry ran straight into the same collision.

## The fix

Growing a record inside its own page changes only that record's content: the records that follow simply
move, and their new offsets are recomputed from the page itself. That makes the write as replayable as an
in-place overwrite.

The in-page growth is now a single primitive, `LocalBucket.growRecordInPage`, shared by the update path and
by the commit-time replay - so the replay grows the record on the newer committed page in exactly the same
way, after checking **byte for byte** that the slot still holds the pre-image this transaction started from.
A growth that no longer fits the committed page refuses to rebase and falls back to a normal retry.

Everything the merge could not replay before still poisons the page: a record spilled out of its page into a
placeholder or a chunk chain, a placeholder content record (it lives behind a pointer on another page), a
delete. A concurrent write to the **same** record still raises `ConcurrentModificationException`, by design -
it is the signal to reload and decide.

## Numbers

Measured on the reported workload (one single bucket, `attempts=1`, no retry):

| Scenario | Before | After |
|---|---|---|
| 10 transactions updating 10 different records of one page | 9 failed / 10 | 0 |
| Sustained updates, 8 writers on their own records of one page | ~2083 / 2880 | 0 |

## Tests

- `Issue5279ConcurrentUpdateTest` (6 tests): the reporter's scenario, shrinking updates, sustained
  multi-writer load, concurrent edge creation on shared pages, a long contended run where records outgrow
  their page and spill into placeholders/chunks (verified with `check database`), and the guarantee that two
  transactions writing the **same** record still conflict and the winner's value survives.
- `Issue5279RemoteConcurrentUpdateIT`: the reporter's scenario over the wire, where a client-managed remote
  transaction spans several HTTP calls so the server auto-retry never fires. It fails on `main` with his
  exact exception and passes here.

Every new test was run against a baseline worktree at `main` to confirm it reproduces the bug.

## Verification

- Engine suite: 10,259 tests, 1 failure - `LSMVectorIndexTest.manualCompactionRemovesDuplicates`, which
  fails on `main` too (a different wrong count every run: 13414, 21149 vs 30000 expected). It is a
  single-threaded test, so it looks like a real pre-existing vector-compaction bug, unrelated to this PR and
  worth its own issue.
- Server module: 771 tests, all green.

## Follow-up (deliberately not here)

A plain single-record **delete** is commutative too (it writes 0 into one slot entry) and is the last common
false-conflict class left on a bucket page. It needs a third tracked kind in `SlotRebaseBuffer` plus care
that an insert-then-delete in one transaction replays as a no-op rather than a failed pre-image check. Left
out to keep this change scoped to what was reported.


lvca:
Thanks - good review. Addressed in 38a87c40f.

**Finding 1 (placeholder-pointer fall-through): agreed, fixed.** Your reading is exactly right, including that the
committed `-1` marker makes the replay refuse so nothing could be corrupted. I fixed it one step earlier than
suggested: the page is poisoned right where the pointer is dropped (after `deleteRecordInternal`), not only in the
growth branch. That also closes the *same* hole on the same-or-smaller path, which predates this PR - a new value of
8 bytes or less was already being tracked with the 8 pointer bytes as its pre-image.

You were also right that the path was untested. `aPlaceholderPointerRebuiltUnderConflictFallsBackAndStaysCorrect`
now covers it: fill page 0, turn a tiny record into a placeholder *pointer* (asserted through the bucket layout
stats - `totalPlaceholderRecords=1`, `totalMultiPageRecords=0`, so it is a pointer and not a chunk chain), then
rewrite it to 70 KB. Larger than any page, so the content record cannot grow either and the fall-through is
certain rather than incidental. A concurrent commit to a co-located record forces the conflict; the transaction has
to fall back, the retry lands the exact value, the neighbour is intact and `check database` is clean. The final
`totalMultiPageRecords=1` is the discriminator that proves the slot was *rebuilt* rather than updated in place.

One caveat worth stating: that test does not distinguish poisoned from un-poisoned - both end in the same
`ConcurrentModificationException`, since the replay refused the slot anyway. It is coverage for the path plus the
correctness invariant, not a regression test for the poison line itself. The gap the poison closes needs the byte
coincidence you describe, which cannot be provoked.

**Finding 2 (stale comment): fixed, and there were three more.** `Issue5381FalseConflictTest` documented the
opposite of the new behaviour in two comments *and* in a method name
(`growingUpdatesOnCoLocatedRecordsFallBackAndStayCorrect`). Reworded and renamed - assertions unchanged, still
green. Thanks for pointing at the class of problem; the diff alone did not show those.

**Finding 3 (double compute on the spill path): acknowledged, leaving it.** Threading the value out needs an
out-param or a holder allocation on the growth path itself, to save two page reads on a path that then goes on to
allocate a placeholder record or split the record into chunks. Not a trade I would take.

**Finding 4 (`@Tag("slow")`): measured, does not apply.** Per-method times for the class: 0.386 / 0.217 / 0.211 /
0.201 / 0.161 / 0.099 / 0.057 / 0.007 s - about 1.1 s in total, so neither test is a `slow` candidate.

**Re-verified after the change:** full engine suite 10,262 tests, 0 failures. Note that
`LSMVectorIndexTest.manualCompactionRemovesDuplicates`, which failed in my earlier run, passed this time - it is
intermittent. I did confirm it fails on `main` as well (a different wrong count each time: 13414, 21149 against
30000 expected) in a single-threaded test, so it looks like a genuine pre-existing vector-compaction bug and I will
open it separately rather than fold it in here.

Issue #565, opened 2022-10-13
SQL Select NULL


gramian:
### OS: Ubuntu 20.04

I was experimenting with NULL in projections and ran into some errors. As these are likely related, I bundled these into one issue:

## Expected behavior
1. `{}`
2. `{}`

## Actual behavior
1. Correct server response, then Studio message box: `DataTables warning: table id=result - Cannot reinitialise DataTable. For more information about this error, please see http://datatables.net/tn/3`
2. Correct server response, but Studio message box: `DataTables warning: table id=result - Cannot reinitialise DataTable. For more information about this error, please see http://datatables.net/tn/3`

In both cases Studio has to be reloaded.

## Steps to reproduce
1. `SELECT null;`
2. `SELECT if(null, true, false);`

gramian:
Even though similar, https://github.com/ArcadeData/arcadedb/commit/31ec15671527adade79aa3d43d073b7cd25e38df does not fix this issue.

lvca:
That has been fixed in Studio. Please check the latest `main` branch and clear the cache of the browser.

gramian:
Tested 56da7b03292b5ba32edc7376cdc09a0ead2b4686 in which  it works. Cool!
Sorry, if clearing browser cache was the issue before.

PR #5572, opened 2026-07-30
test(engine) #5570: tolerate retry exhaustion in EdgeAppendMergeRaceTest, keep failing on corruption

### LINKED ISSUE #5570: CI: EdgeAppendMergeRaceTest.addersRemoversAndReadersOnOneHotVertex fails on retry exhaustion under CI load (4 of last 8 main runs)
Split out of #5561, where the `integration-tests` break turned out to be deterministic and unrelated. This is the other thing keeping `Java CI - test` red on `main`, and it is not the flakiness the rest of that job shows.

## Symptom

`EdgeAppendMergeRaceTest.addersRemoversAndReadersOnOneHotVertex` fails in the `slow-unit-tests` job with the same signature every time: a `ConcurrentModificationException` escaping a transaction that has exhausted its retries.

```
[ERROR] EdgeAppendMergeRaceTest.addersRemoversAndReadersOnOneHotVertex -- Time elapsed: 13.13 s <<< FAILURE!
java.lang.AssertionError: 1 thread(s) failed, first: com.arcadedb.exception.ConcurrentModificationException:
  Concurrent modification on page PageId(EdgeAppendMergeRaceTest/3/0) in file
  'Account_0_in_edges.3.65536.v0.bucket' (current v.8709 <> database v.8712). Please retry the operation
	at com.arcadedb.engine.PageManager.checkPageVersion(PageManager.java:369)
	at com.arcadedb.database.TransactionContext.commit1stPhase(TransactionContext.java:1372)
```

## Frequency

4 of the last 8 completed `Java CI - test` runs on `main`. Always the same test, always the same class of error:

| run | commit | page | version gap |
|---|---|---|---|
| 30506841523 | `084ca9f3` | `Account_0_in_edges` | v.8709 <> v.8712 |
| 30504641942 | `f2909e34` | `Account_0_in_edges` | v.8286 <> v.8288 |
| 30497002373 | `e6fd2cc8` | `Account_0_in_edges` | v.8817 <> v.8821 |
| 30480843978 | `540bb7a4` | `Account_2` (vertex bucket) | v.534 <> v.535 |

Passes locally on a fast machine, which is why it reads as a flake. It is really a load-sensitive threshold: 16 threads (8 adders, 4 removers, 4 readers) all commit against one hot vertex's edge chain, ~18k transactions in total, each with 100 retries.

## Why I do not think this is an engine regression

The version gaps are tiny (2 to 4 versions), so the losing transaction is not stale, it just keeps losing a fair race. Given N writers on one page, some transaction exhausting 100 retries across 18k attempts is the expected outcome once the runner is slow enough, not a livelock and not a lost poison bit.

## What the test should assert instead

The test exists to prove that a structural write (edge removal relinking or deleting a chunk) is never rebased over by the commutative append merge - the corruption shape from #565, which surfaces as a `BufferUnderflowException` on traversal. It does not exist to prove that no transaction ever exhausts its retries: `ConcurrentModificationException extends NeedRetryException`, and giving up after N retries under extreme contention is a legitimate engine outcome that applications are expected to handle.

So the assertion should separate the two:

- **Fail** on any non-retryable error, and on any parse/traversal failure from a reader (that is the real bug being hunted).
- **Tolerate** a retry-exhausted `ConcurrentModificationException`, counted and reported rather than fatal.

The final consistency checks stay exactly as they are and need no adjustment: `addedCount` / `removedCount` are only incremented after a successful commit, and a remover that fails leaves its victim un-deleted, so `expected = added - removed` remains correct either way.

Worth also logging how many transactions gave up, so a future genuine starvation regression is visible as the count jumping rather than as an intermittent red build.

lvca:
Closes #5570.

## Why the test was red

`EdgeAppendMergeRaceTest.addersRemoversAndReadersOnOneHotVertex` failed in 4 of the last 8 `Java CI - test` runs on `main`, always with the same signature: a `ConcurrentModificationException` escaping a transaction that had used up all 100 of its attempts.

`LocalDatabase.transaction(block, join, attempts)` rethrows the last `NeedRetryException` once attempts run out, so a transaction that keeps losing a fair race lands in the test's `catch (Throwable)` and becomes a hard failure. The version gaps observed in CI were 2 to 4 - the loser was never stale. With 16 threads (8 adders, 4 removers, 4 readers) committing against one hot vertex's chain, some transaction losing 100 times in a row is the expected outcome on a slow runner, not a livelock and not a lost poison bit.

## What changed (test-only, no engine change)

The race body moved into `runRace(perAdder, attempts)`, and every transaction now goes through a `commit()` wrapper that separates the two things the test was conflating:

- a `NeedRetryException` from an exhausted transaction is **counted**, not fatal. This also covers `LockTimeoutException`, the other load-sensitive failure mode `engine/CLAUDE.md` warns about on slow runners, which would have been the next intermittent red;
- anything non-retryable still fails immediately - above all a reader whose traversal stopped parsing, the `BufferUnderflowException` shape from #565 that this test exists to hunt.

Consistency assertions are unchanged and needed no adjustment: `addedCount` / `removedCount` only move after a successful commit, and a remover that gives up leaves its victim un-deleted, so `expected = added - removed` stays exact. A victim whose delete gave up is deliberately not re-queued - retrying it forever would hide the give-up instead of reporting it.

## Three additions beyond what the issue proposed

The issue proposed "count and report". Reporting alone means a genuine starvation regression scrolls past in a log nobody reads, so:

1. **The count is bounded, not just logged**: give-ups must stay under 5% of write transactions. Occasional losses pass; a livelock or real starvation trips the assertion.
2. **Escalating log level**: the suite pins `com.arcadedb` to `SEVERE`, so the outcome line logs at `SEVERE` only when there was at least one give-up (`WARNING` otherwise). A clean run prints nothing; a degrading run becomes visible in the build log well before it grows enough to fail.
3. **A fast deterministic companion test**, `retryExhaustionIsToleratedAndCounted`: the same race with `attempts=1`, so exhaustion is the norm instead of the exception. It pins the tolerance path down locally rather than leaving it exercised only by an unlucky CI runner, and proves the final consistency check stays exact when hundreds of transactions give up. ~0.5s.

## Verification

```
EdgeAppendMergeRaceTest                              2/2 pass
  addersRemoversAndReadersOnOneHotVertex   3.2s   added=12000 removed=6000 merges=10547 give-ups=0
  retryExhaustionIsToleratedAndCounted     0.4s   added=1517  removed=9    merges=26878 give-ups=830
com.arcadedb.graph.*Test (incl. @Tag("slow"))       26/26 pass
```

Fast variant run 5x standalone: 5/5 green, give-ups robustly in the 820-832 range, so `isGreaterThan(0)` is not a new flake.

## Follow-up worth considering separately

`LocalDatabase.delayBetweenRetries` sleeps a flat `1 + random(0..100)` ms on every attempt, with no backoff. Under 16-way contention that is a thundering herd where a transaction's odds of winning do not improve with each loss, which is why exhaustion after 100 attempts is reachable at all. Exponential backoff with jitter would make give-ups far rarer for real applications, not just for this test. That changes engine retry semantics on a hot path, so it is out of scope here.

https://claude.ai/code/session_01VBKzpXuVgoDPbrBwNsGCaz

### DIFF
--- a/engine/src/test/java/com/arcadedb/graph/EdgeAppendMergeRaceTest.java
+++ b/engine/src/test/java/com/arcadedb/graph/EdgeAppendMergeRaceTest.java
@@ -95,6 +95,16 @@ class EdgeAppendMergeRaceTest extends TestHelper {
     // Readers are the corruption detector, so their give-ups need a floor rather than the writers' ceiling: a
     // run where readers mostly failed to complete a traversal would go green with the BufferUnderflowException
     // guard barely exercised. Assert the detector actually ran, in absolute terms and as a share.
+    //
+    // Neither bound re-couples the test to runner speed, which is the thing #5570 set out to remove:
+    //
+    // - a read-only transaction cannot give up at all. commit1stPhase locks exactly the files that
+    //   lockFilesFromChanges() collects from modifiedPages / newPages / indexChanges, all empty here, so
+    //   tryLockFiles iterates an empty list and no COMMIT_LOCK_TIMEOUT can elapse; with no modified page there
+    //   is likewise nothing for checkPageVersion to reject. So readerGiveUps is structurally 0, which makes the
+    //   share bound vacuous (n > n/2) until readers really do start failing - exactly the degradation it guards.
+    // - the absolute floor is 4 traversals for the whole run, against 18000 write transactions the readers loop
+    //   underneath, and the earliest ones parse a nearly empty chain. No runner is slow enough for that.
     final int readTx = outcome.readTransactions();
     assertThat(outcome.readerCommits())
         .as("full chain traversals completed: %d of %d reader transactions", outcome.readerCommits(), readTx)
@@ -139,7 +149,10 @@ class EdgeAppendMergeRaceTest extends TestHelper {
     final AtomicInteger addedCount = new AtomicInteger();
     final AtomicInteger removedCount = new AtomicInteger();
     final AtomicInteger readerCommits = new AtomicInteger();
-    final GiveUps writerGiveUps = new GiveUps();
+    // Tallied per role, not per writer/reader: an adder losing the append race and a remover losing the
+    // structural race are different symptoms, and a starvation regression is worth attributing to one of them.
+    final GiveUps adderGiveUps = new GiveUps();
+    final GiveUps removerGiveUps = new GiveUps();
     final GiveUps readerGiveUps = new GiveUps();
     final AtomicBoolean addersDone = new AtomicBoolean();
     final CountDownLatch start = new CountDownLatch(1);
@@ -159,7 +172,7 @@ class EdgeAppendMergeRaceTest extends TestHelper {
               src.save();
               src.newEdge("TRANSFERS", hubRID);
               holder[0] = src.getIdentity();
-            }, attempts, writerGiveUps))
+            }, attempts, adderGiveUps))
               // GAVE UP: nothing was committed, so nothing to count and nothing to offer to the removers.
               continue;
 
@@ -190,7 +203,7 @@ class EdgeAppendMergeRaceTest extends TestHelper {
             // A remover that gives up leaves its victim un-deleted and does not count it, so the expected edge
             // count stays exact. The victim is deliberately not re-queued: retrying it forever would hide the
             // give-up instead of reporting it.
-            if (commit(() -> victim.asVertex(true).delete(), attempts, writerGiveUps))
+            if (commit(() -> victim.asVertex(true).delete(), attempts, removerGiveUps))
               removedCount.incrementAndGet();
           }
         } catch (final Throwable e) {
@@ -230,8 +243,8 @@ class EdgeAppendMergeRaceTest extends TestHelper {
     for (final Thread thread : threads)
       thread.join();
 
-    final RaceOutcome outcome = new RaceOutcome(addedCount.get(), removedCount.get(), writerGiveUps.count.get(),
-        readerCommits.get(), readerGiveUps.count.get(),
+    final RaceOutcome outcome = new RaceOutcome(addedCount.get(), removedCount.get(), adderGiveUps.count.get(),
+        removerGiveUps.count.get(), readerCommits.get(), readerGiveUps.count.get(),
         ((DatabaseInternal) database).getPageManager().getStats().edgeAppendMerges);
 
     // The suite pins com.arcadedb to SEVERE, so a run with give-ups is the only one that prints: a genuine
@@ -239,9 +252,10 @@ class EdgeAppendMergeRaceTest extends TestHelper {
     // to trip the assertions below.
     LogManager.instance()
         .log(this, outcome.writerGiveUps() + outcome.readerGiveUps() > 0 ? Level.SEVERE : Level.WARNING,
-            "#5570 race outcome: added=%d removed=%d merges=%d give-ups(write)=%d traversals=%d give-ups(read)=%d%s%s",
-            outcome.added(), outcome.removed(), outcome.merges(), outcome.writerGiveUps(), outcome.readerCommits(),
-            outcome.readerGiveUps(), writerGiveUps.describeFirst("write"), readerGiveUps.describeFirst("read"));
+            "#5570 race outcome: added=%d removed=%d merges=%d traversals=%d give-ups: adder=%d remover=%d reader=%d%s%s%s",
+            outcome.added(), outcome.removed(), outcome.merges(), outcome.readerCommits(), outcome.adderGiveUps(),
+            outcome.removerGiveUps(), outcome.readerGiveUps(), adderGiveUps.describeFirst("adder"),
+            removerGiveUps.describeFirst("remover"), readerGiveUps.describeFirst("reader"));
 
     if (!errors.isEmpty())
       throw new AssertionError(errors.size() + " thread(s) failed with a non-retryable error, first: " + errors.getFirst(),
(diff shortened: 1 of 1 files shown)


lvca:
Thanks - review verified the core claims against the engine and I have acted on the two actionable items in `4e113256a`.

**Item 1 (reader coverage unbounded) - fixed.** This was a real gap: the readers *are* the corruption detector, so their give-ups need a floor, not the writers' ceiling. Completed full traversals are now counted (`readerCommits`) and asserted two ways:

```java
assertThat(outcome.readerCommits())
    .as("full chain traversals completed: %d of %d reader transactions", outcome.readerCommits(), readTx)
    .isGreaterThanOrEqualTo(READERS)
    .isGreaterThan(readTx / 2);
```

An absolute floor catches "readers never got going at all", the share catches "readers ran but mostly gave up". Locally 7915 of 7915 reader transactions complete with 0 give-ups, so both bounds have wide room and are not a new flake surface. A run where the detector silently stopped running now fails instead of passing green.

**Item 3 (shared `firstGiveUp` sample) - fixed.** Count and sample now live together in a per-role `GiveUps` holder, so a writer-dominated run can no longer report a reader exception as its example. This also shortens the `commit()` signature from four parameters to three, and the log line names the role it sampled: `first(write)=...` / `first(read)=...`.

**Item 2 (probabilistic `writerGiveUps > 0`) - acknowledged, left as is.** Agreed it is statistical rather than deterministic. Making it truly deterministic would need a commit-time failure injection point in the engine, which is a heavier change than the guarantee is worth: with `attempts=1` and 16 threads on one page the observed range is 830-838 give-ups across 8 runs, i.e. three orders of magnitude above the threshold. I would rather revisit it if it ever actually flakes than add an engine seam for a test.

Re-verified after the change:

```
EdgeAppendMergeRaceTest   3x consecutive full-class runs   2/2 pass each
  addersRemoversAndReadersOnOneHotVertex   added=12000 removed=6000 merges=10411 give-ups(write)=0 traversals=7915 give-ups(read)=0
  retryExhaustionIsToleratedAndCounted     give-ups(write)=830-838, traversals=1407-2246, give-ups(read)=0
```

lvca:
Second pass addressed in `b6cc50577`.

**Item 2 (`DuplicatedKeyException` not caught) - good catch, documented.** Verified: `DuplicatedKeyException extends ArcadeDBException`, not `NeedRetryException`, so `transaction()` retries it but on exhaustion it bypasses the wrapper and fails the test. That is the right outcome here - this workload has no unique index, so a duplicate would mean the engine handed out a colliding RID - but the asymmetry is invisible at the call site and would silently follow a copy-paste into a unique-key test. Now spelled out on `commit()`:

```
Contention give-ups are the ONLY tolerated outcome, deliberately. transaction() also retries
DuplicatedKeyException, which does not extend NeedRetryException and so would still fail this test on
exhaustion - correct here, since this workload has no unique index and a duplicate would mean the engine
handed out a colliding RID. Anyone reusing this wrapper on a unique-key workload has to decide that case
explicitly rather than inherit this one.
```

**Item 1 (probabilistic `writerGiveUps > 0`)** - same position as before: statistical, not deterministic, but 830-838 give-ups observed across 9 runs against a threshold of 1. The few-cores concern is fair; if it ever flakes there, the fix is to drop `attempts` handling entirely for that test rather than to add an engine injection seam.

**On the compile/run you could not do in the sandbox:** done locally, `mvn -pl engine test -Dtest=EdgeAppendMergeRaceTest -Dgroups=slow`, 4 consecutive full-class runs, 2/2 pass each. Latest:

```
added=12000 removed=6000 merges=... give-ups(write)=0 traversals=8320 give-ups(read)=0   [main test]
give-ups(write)=824  traversals=1400-8300  give-ups(read)=0                              [attempts=1 test]
```

Also re-ran the whole `com.arcadedb.graph.*Test` package including `@Tag("slow")`: 26/26 pass.

lvca:
Third pass addressed in `b2df96c12`.

**Main ask - the reader `> readTx / 2` bound re-coupling to runner speed.** Good question to raise, and I traced it rather than argued it. It does not couple, and the reason is structural, not statistical, so it is now written down beside the assertion:

A read-only transaction **cannot give up at all**. `commit1stPhase` locks exactly the files that `lockFilesFromChanges()` collects from `modifiedPages` / `newPages` / `indexChanges` (`TransactionContext.java:1748-1763`), all empty for a traversal, so `tryLockFiles` iterates an empty ordered list and never calls `tryLockFile` - no `COMMIT_LOCK_TIMEOUT` can elapse (`TransactionManager.java:811-840`). With no modified page there is likewise nothing for `checkPageVersion` to reject.

So `readerGiveUps` is 0 by construction, which makes the share bound **vacuous** - it reduces to `n > n/2` - until readers really do start failing, which is exactly the degradation it exists to catch. It cannot flip a slow runner red; it can only fire on the thing the previous review pass asked me to guard. I kept it, with the justification in a comment (your option b) rather than dropping it (option a), since dropping it would give back the coverage guarantee.

On the absolute floor: 4 traversals total for a run of 18000 write transactions, with the earliest ones parsing a nearly empty chain. That also covers your `readTx == 0` note - it should fail if readers never ran, that is the point.

**Minor 2 (shared writer tally) - fixed, and it paid off immediately.** Adders and removers now keep separate `GiveUps`. The bound still applies to writers as a whole, since both are writers, but the log attributes losses to the role that suffered them. At `attempts=1` the split is `adder=85-88` vs `remover=740-747` - the structural race is where give-ups concentrate, which the single "write" label hid:

```
give-ups: adder=88 remover=740 reader=0
  first(adder)=ConcurrentModificationException: Slot rebase not possible on page ...1/0 slot 0
  first(remover)=ConcurrentModificationException: Concurrent modification on page ...3/2 in file 'Account_0_in_edges...'
```

**Follow-up on `delayBetweenRetries`** - agreed on scope. The remover concentration above is direct evidence for it: flat `1 + random(0..100)` ms gives a loser no improving odds across attempts. I will open it as its own engine issue with a benchmark rather than smuggle it in here.

Re-verified on the final HEAD: 3 consecutive full-class runs, 2/2 pass each; main test `traversals=930-2361`, `give-ups: adder=0 remover=0 reader=0`.

PR #5574, opened 2026-07-30
perf(engine) #5519: dictionary-encode TimeSeries TAG columns to remove the 258-byte stride per column

### LINKED ISSUE #5519: TimeSeries mutable rows reserve 258 bytes per STRING column: a 10-tag row fits 25 per page and pays 24x page+WAL amplification

lvca:
Closes #5519.

## The problem

A TimeSeries mutable row is fixed-stride, so every column has to reserve the widest value it could ever hold. For
a `STRING` TAG column that is `2 + MAX_STRING_BYTES` = **258 bytes**, whether the tag is `us-east-1` or empty.

Tags are low-cardinality by definition - that is what makes them tags rather than fields - so almost all of that
reservation was padding, and padding in a fixed-stride row is not free: it is written, flushed, and shipped
through the WAL like real data. The shipping TSBS `cpu_influx.lp` schema has ten tags, which is a 2,612-byte
stride, **25 rows per 64K page**, and 23x page+WAL amplification against payload.

`TimeSeriesTagStrideBenchmark` (added in #5520) isolated it, and @tae898 corroborated it end to end from Python
on real TSBS data - 2,592,000 points, N=3 - measuring 4.74x slower ingest for ten tags than for one, with strides
and rows-per-page matching the formula exactly.

## The fix

A TAG column now holds a **4-byte id** into a per-**type** append-only dictionary component (`.tstd`, one file
per TimeSeries type, shared by every shard) instead of an inline reserved slot.

| arm | tags | fields | stride | rows per 64K page |
|---|---|---|---|---|
| 1 tag, 3 fields | 1 | 3 | 290 B → **36 B** | 225 → **1819** |
| 10 tags, 3 fields | 10 | 3 | 2612 B → **72 B** | 25 → **909** |
| 10 tags, 10 fields | 10 | 10 | 2668 B → **128 B** | 24 → **511** |

`TimeSeriesTagStrideBenchmark`, same host, same 20k batches:

```
before
arm                      stride   rows/page      pages      payload       page traffic       ms
WideStride                 2612          25        800       2.1 MB      50.0 MB (23x)     29.9
WideStrideBigValues        2612          25        800      39.1 MB      50.0 MB (1x)      25.5
NarrowStride                290         225         89       1.9 MB       5.6 MB (2x)       4.3

after
WideStride                   72         909         23       2.1 MB       1.4 MB (0.7x)      6.0
WideStrideBigValues          72         909         23      39.1 MB       1.4 MB (0.0x)      5.5
NarrowStride                 36        1819         11       1.9 MB       0.7 MB (0.4x)      2.0
```

A tag-heavy schema does **not** converge on a narrow one, and should not: the single-tag arm was paying the same
reservation for its one column, so it improves too. The stride ratio between the two goes from 9.0x to 2.0x -
what remains is the real cost of nine more columns (nine more dictionary lookups, 36 more bytes per row), with no
padding left to remove.

## Design decisions worth reviewing

- **Ids are 4 bytes, not the 2 the sealed `DictionaryCodec` uses.** That costs ~900 rows per page instead of
  ~1250 and buys away the overflow-past-65535 question entirely, rather than deferring it to a per-type limit
  someone eventually hits. `DictionaryCodecOverflowTest` covers the sealed side, which is unchanged.
- **Id 0 is reserved for null/empty**, so the existing null round trip is byte-identical.
- **Per type, not per shard.** Tag values repeat across shards by construction, so a per-shard dictionary would
  hold N copies of the same low-cardinality set and lose most of the win on a sharded type.
- **STRING *fields* stay inline.** A field is where high-cardinality text belongs; interning it would grow the
  dictionary without bound for no reuse. This is the one asymmetry in the change and it is deliberate.
- **`arcadedb.timeSeriesTagDictionaryMaxSize`**, default 1M distinct values (~100MB). The dictionary is held in
  RAM, so the cap turns a mis-declared high-cardinality TAG into a clear error rather than unbounded growth.

## Compatibility

**Existing databases open and read unchanged, and there is no in-place migration.** The mutable row format is
versioned per type: a type created by an earlier build keeps the inline layout, a new type gets the encoding.
An existing type has to be recreated to gain it. Worth stating plainly in the release note (included here)
because a benchmark pointed at an existing database will measure v0 and see no change at all.

The `.tstd` component is registered in both `LocalSchema` and `LocalDatabase`'s supported extensions, so it
survives a reopen rather than being silently discarded.

## Tests

- `TimeSeriesTagDictionaryTest` (11) - the component itself: append, lookup, id stability, null/empty handling,
  persistence across reopen, the max-size cap.
- `Issue5519TagStrideTest` (13) - the row layout end to end: stride per schema shape, rows per page, tag values
  surviving a round trip, mixed TAG/field schemas, existing-type version pinning.
- `TimeSeriesTagStrideBenchmark` - the two structural assertions were **inverted rather than deleted**, so the
  benchmark now guards the fix instead of documenting the defect: a 64K page holds more than 500 ten-tag rows,
  the two arms are within 3x of each other, and page-traffic amplification is below 1.5x where it used to assert
  above 10x. Of the timing assertions, `WideStrideBigValues < WideStride * 2` holds unchanged and
  `NarrowStride * 2 < WideStride` was relaxed to `NarrowStride < WideStride` for the convergence reason above.

## Verification

Rebased onto current `main` (`9a9358ec5`) and re-run there:

- **484 tests, 0 failures** in `com.arcadedb.engine.timeseries.**`, including `TimeSeriesCrashRecoveryTest`,
  `TimeSeriesConcurrentAppendCompactTest`, `Issue5475ColumnTypesTest` and `Issue5474PrimitiveIngestTest`.
- **948 tests, 0 failures** across `com.arcadedb.schema.**` + `com.arcadedb.engine.**`, covering the
  `LocalSchema` / `LocalDatabase` component registration.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -231,6 +231,45 @@ Two related corrections:
 A batch that fails still connects the incoming edges of what it already committed before it answers, because
 skipping that would leave persisted edges without back-pointers. On a large load that pass takes a while, so
 it now says so in the log rather than looking like a fresh hang.
+#### TimeSeries TAG columns are dictionary-encoded: 36x less page traffic on a tag-heavy schema
+
+A TimeSeries mutable row is fixed-stride, so a `STRING` TAG column reserved the widest value it could ever
+hold - `2 + MAX_STRING_BYTES`, 258 bytes - whether the tag was `us-east-1` or empty. Tags are low-cardinality
+by definition, which is what makes them tags, so nearly all of that was padding that still had to be written,
+flushed and shipped through the WAL ([#5519](https://github.com/ArcadeData/arcadedb/issues/5519)).
+
+A TAG column now holds a 4-byte id into a per-**type** append-only dictionary component (`.tstd`, one file per
+TimeSeries type, shared by every shard):
+
+| arm | tags | fields | stride | rows per 64K page |
+|---|---|---|---|---|
+| 1 tag, 3 fields | 1 | 3 | 290 B → **36 B** | 225 → **1819** |
+| 10 tags, 3 fields | 10 | 3 | 2612 B → **72 B** | 25 → **909** |
+| 10 tags, 10 fields | 10 | 10 | 2668 B → **128 B** | 24 → **511** |
+
+On `TimeSeriesTagStrideBenchmark`, the ten-tag arm went from writing 50.0 MB of pages for 2.1 MB of payload
+(23x amplification) to 1.4 MB (0.7x), and from 29.9 ms to 6.0 ms. The single-tag arm improves too - it was
+paying the same reservation for its one column - so a tag-heavy schema does not converge on a narrow one: the
+stride ratio between them goes from 9.0x to 2.0x, which is the honest cost of nine more columns rather than
+nine more paddings. Corroborated independently on real TSBS data (2,592,000 points through the primitive batch
+API) by @tae898 in the issue.
+
+- **Ids are 4 bytes, not the 2 the sealed `DictionaryCodec` uses.** That trades ~900 rows per page for ~1250
+  in exchange for removing the overflow-past-65535 question entirely. Id 0 is reserved for null/empty, so the
+  old round trip is unchanged.
+- **STRING *fields* stay inline.** A field is where high-cardinality text belongs; interning it would grow the
+  dictionary without bound.
+- **`arcadedb.timeSeriesTagDictionaryMaxSize`** caps distinct values per type, default 1M (roughly 100MB). The
+  dictionary is held in RAM, so this turns a mis-declared high-cardinality TAG into a clear error instead of
+  unbounded growth.
+- **Existing types keep the inline layout.** The mutable row format is versioned per type, so a database
+  written by an earlier build opens and reads unchanged, and there is no in-place migration. A new TimeSeries
+  type gets the encoding; an existing one has to be recreated to gain it. If you are benchmarking, point the
+  harness at a fresh database or you will measure the old layout and see no change.
+- **A probe that computes the stride itself has the mirror problem.** A harness carrying the old
+  `8 + 258 * tags + 8 * fields` formula reports the layout this replaced and turns a real improvement into what
+  looks like a measurement error. The encoded formula is `8 + 4 * tags + 8 * fields`. Both failure modes produce
+  a confident null result from a harness that looks correct in isolation.
 
 ### Breaking Changes
 
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -234,6 +234,10 @@ public enum GlobalConfiguration {
       "Filesystem directory where new paired external-property buckets are created. If empty (default), external buckets sit alongside primary buckets in the database directory. Set to a path on cheaper/slower storage (HDD, network mount) to tier the heavy payloads away from the topology files. The directory must exist and be writable. Existing external buckets are not relocated when this changes.",
       String.class, ""),
 
+  TIMESERIES_TAG_DICTIONARY_MAX_SIZE("arcadedb.timeSeriesTagDictionaryMaxSize", SCOPE.DATABASE,
+      "Maximum number of distinct values one TimeSeries type's tag dictionary may hold. TAG columns are dictionary-encoded in the mutable row so each occupies a 4-byte id instead of a reserved 258-byte slot; the dictionary is kept in RAM, so this caps its footprint and turns a mis-declared high-cardinality TAG into a clear error instead of unbounded growth. Default is 1M distinct values, roughly 100MB",
+      Integer.class, 1_000_000),
+
   BUCKET_REUSE_SPACE_MODE("arcadedb.bucketReuseSpaceMode", SCOPE.DATABASE,
       "How to reuse space in pages. 'high' = more space saved, but slower opening and update/delete time. 'medium' to still reuse space without the initial scan at opening time. 'low' for faster performance, but less space reused. Default is 'high'",
       String.class, "high", Set.of("low", "medium", "high")),
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -37,6 +37,7 @@ import com.arcadedb.engine.WALFile;
 import com.arcadedb.engine.WALFileFactory;
 import com.arcadedb.engine.WALFileFactoryEmbedded;
 import com.arcadedb.engine.timeseries.TimeSeriesBucket;
+import com.arcadedb.engine.timeseries.TimeSeriesTagDictionary;
 import com.arcadedb.exception.ArcadeDBException;
 import com.arcadedb.exception.CommandExecutionException;
 import com.arcadedb.exception.ConcurrentModificationException;
@@ -155,6 +156,7 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
       LSMVectorIndexGraphFile.FILE_EXT,
       SparseSegmentComponent.FILE_EXT,
       TimeSeriesBucket.BUCKET_EXT,
+      TimeSeriesTagDictionary.DICT_EXT,
       HashIndexBucket.UNIQUE_INDEX_EXT,
       HashIndexBucket.NOTUNIQUE_INDEX_EXT);
 
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
@@ -33,6 +33,7 @@ import com.arcadedb.schema.Type;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
@@ -60,7 +61,16 @@ import java.util.Set;
  * - [2..9]   min timestamp in page (long)
  * - [10..17] max timestamp in page (long)
  * - [18..]   row data: fixed-size rows [timestamp(8)|col1|col2|...]
- *            For STRING columns: 2-byte length prefix + up to MAX_STRING_BYTES payload
+ *            TAG STRING columns: a 4-byte {@link TimeSeriesTagDictionary} id (format version 1)
+ *            other STRING columns: 2-byte length prefix + up to MAX_STRING_BYTES payload
+ * <p>
+ * <b>Format versions.</b> Version 0 stored every STRING column inline and reserved
+ * {@code 2 + MAX_STRING_BYTES} for it in the stride, so a ten-tag row cost 2612 bytes of stride to
+ * carry ~110 bytes of payload: 25 rows per 64 KB page and 24x page and WAL amplification (issue
+ * #5519). Version 1 dictionary-encodes TAG STRING columns into a fixed 4-byte id, taking the same
+ * schema to a 72-byte stride and ~900 rows per page. Both layouts are readable; which one a bucket
+ * uses is decided by whether a dictionary was handed to it, which the schema's
+ * {@code mutableFormatVersion} determines, so a database written by an older build keeps working.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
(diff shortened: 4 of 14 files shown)


tae898:
Reading this properly turned up something about **our** harness rather than yours, and one caution that probably applies to anyone else measuring the fix.

**Our TimeSeries lane declares one tag, and real TSBS declares ten.** The lane behind the paper's ingest number creates `TAGS (host STRING)`, three fields. The TSBS `cpu` schema this is meant to represent has ten tags (hostname, region, datacenter, rack, os, arch, team, service, service_version, service_environment). So our published figure sits on the 290-byte stride arm, not the 2,612-byte one.

Which means our number has been **flattering ArcadeDB**, not penalising it. A faithful TSBS schema would have been paying the 25-rows-per-page amplification your table shows, and we would have reported a worse ingest figure than we did. That is our fidelity gap, it predates your fix, and it is the kind of thing that is easy to miss because the harness looks reasonable in isolation.

The useful consequence for you: the 10-tag arm is the one real TSBS users are on, so the 36x page-traffic reduction is the number that describes them, not the 8x from our single-tag arm. If anyone quotes our end-to-end figure as corroboration, it corroborates the narrow arm only, and I will say so explicitly when I post it.

**Post-merge we intend to move the lane onto the real ten-tag schema and re-measure.** That is only viable because of this change; on the current release a faithful TSBS schema would make our own time-series row look much worse than it does, which is an uncomfortable thing to realise about a number already in a draft.

**The caution, for anyone else measuring this.** Our stride probe hardcoded the row-size formula as `8 + 258*n_tags + 8*n_fields`. Run against this branch it would have printed a stride the engine no longer uses and made a genuine improvement look like a measurement error. It now reports both layouts, and I checked the v1 formula against your table: 1 tag/3 fields 290 to 36 bytes, 10 tags/3 fields 2,612 to 72, rows-per-64K-page matching yours to within one.

Worth pairing with the compatibility note you already wrote, since the two failure modes rhyme. Yours is "a benchmark pointed at an existing database measures v0 and sees no change"; this one is "a benchmark that computes the expected stride itself measures v1 and reports v0". Both produce a confident null from a correct-looking harness. Our probes create a fresh database and issue `CREATE TIMESERIES TYPE` per arm, so they exercise whatever the running build creates, but that is closer to luck than to design and I would not assume it of anyone else's.

No ask here. I will post the end-to-end numbers once this is in a release, on both the one-tag and ten-tag arms so the two are separable.

lvca:
All three confirmed and fixed in 6fd18b91e. #1 was real, and the two-wave test you asked for is what proved it.

### 1. `resolveMiss` stopping after the first reload - confirmed, was a permanent-null bug

Not ruled out by HA routing: reproduced directly. The two-wave variant you described fails on the previous commit with

```
expected: "host_d"
 but was: null
```

The mechanism is exactly as you read it. `entryCount` is advanced only by `load()` and `publish()`, and `publish()` runs on the interning path, which a follower never takes. So after the first reload `entryCount == reloadedAtEntryCount` holds forever, every second-wave id returns `null` without reloading, and nothing recovers short of reopening the database. Since a leader interns for as long as it ingests, the second wave is not an edge case - it is the normal state after warm-up.

Fixed by your first suggestion, peeking the header, rather than by the page-count one: a later wave often lands in the *tail* page's free space without allocating anything, so `getTotalPages()` growth would have missed precisely the cheap case.

`peekStoredEntryCount()` reads `HEADER_ENTRY_COUNT_OFFSET` off page 0 in its own transaction, following `load()`'s pattern since this is reachable mid-scan, and degrades to the in-RAM count if the header cannot be read - so a failure declines to reload rather than reloading per row. `reloadedAtEntryCount` is gone entirely.

This is strictly better than what it replaces on the anti-thrash side too: the old guard let one full page walk through per entry-count value, while a corrupt id now costs a single already-cached header read and no walk at all. `anIdBeyondTheDictionaryStaysUnresolved` still passes unchanged.

New test: `aSecondWaveOfIdsIsAlsoResolvedByReloading` - wave 1, reload, wave 2, wave 3, asserting the trigger is the growth rather than the reload count.

### 2. Double UTF-8 encode - confirmed, fixed

`internAll` now encodes once into a `List<byte[]>`, validates against `bytes.length`, and hands the bytes to `appendEntries`, which takes `List<byte[]>` and no longer re-encodes.

### 3. Oversized single entry - confirmed, guarded

You are right that it was unchecked. One correction to the symptom: it would not loop. The spill test runs once per value, so an oversized entry allocates a fresh page and then writes past its capacity - an out-of-bounds page write, not a silent loop. Guarded explicitly with the page size and capacity in the message, plus the config to raise. It needs a page size far below the 64K default to reach, since `MAX_STRING_BYTES` caps an entry at 258 bytes.

### Verification

Rebased on `main` (`29dddd8ae`), clean target: 488 tests, 0 failures across `com.arcadedb.engine.timeseries.**`.

lvca:
Thanks - and the harness correction is the more useful half of this, so I have acted on it here.

**On the arm to quote.** Agreed, and your reading of which direction the error ran is right: a one-tag lane sits on the 290-byte stride, so a faithful ten-tag schema would have measured worse on the current release, not better. The 36x page-traffic figure is the one that describes a TSBS user; the narrow arm's 8x describes the schema your lane actually declares. Publishing both separately, as you plan to, is the right shape - they are different workloads and this change moves both.

**The stride-formula trap is now in the release note.** You are right that it rhymes with the compatibility note rather than duplicating it, so I paired them explicitly:

> **A probe that computes the stride itself has the mirror problem.** A harness carrying the old `8 + 258 * tags + 8 * fields` formula reports the layout this replaced and turns a real improvement into what looks like a measurement error. The encoded formula is `8 + 4 * tags + 8 * fields`. Both failure modes produce a confident null result from a harness that looks correct in isolation.

Your v1 formula checks out against the table on all three arms: 1 tag / 3 fields is `8 + 4 + 24` = 36, 10 / 3 is `8 + 40 + 24` = 72, 10 / 10 is `8 + 40 + 80` = 128. Worth pinning that it holds only while every TAG is a `STRING` - a numeric tag was never reserved at 258 bytes and is unchanged by this.

**On creating the type per arm being closer to luck than design** - it is the only thing that saves a probe here, so it is worth keeping deliberately. The row format is versioned per type and there is no in-place migration, so a harness that reuses a database across builds measures whatever the type was created with, silently and forever.

No need to hold your re-measurement for anything on my side; this is waiting on CI and review. When you do post the two arms, the engine-side numbers in the table above are directly comparable to the stride and rows-per-page your probe prints, so any gap between yours and mine is Python-side assembly rather than a disagreement about the engine.

lvca:
Both the `load()` hazard and the missing v0 fixture are addressed in 2edc28ff4.

### `load()` rebuilding in place - fixed

Agreed on both halves, and the first is the one that would have hurt: a `load()` throwing mid-rebuild left `loaded = true` over a cleared map that `ensureLoaded()` would never retry, so `getId()` reported stored values as absent and the append path interned a **second id for a value that already had one**. Duplicate entries in an append-only component are not something a later reload can undo.

Rebuilt aside and swapped in, exactly as you suggested:

- `byId` and a freshly built map are published only after a complete rebuild, in the order `publish()` already uses (reverse array before forward map), so a reader that finds a value can always resolve the id it got back.
- `loaded` is set only on success, so a failed load is retried by the next `ensureLoaded()` instead of being latched.
- `idByValue` is now `volatile` and replaced wholesale rather than cleared, which closes the lock-free `getId()` race you describe: a lookup sees either the complete old mapping or the complete new one.

One window your note did not cover, closed while I was in here: `publish()` was unsynchronized, so a reload could swap in a mapping rebuilt from the pages *while* a concurrent `publish()` was writing into the one it replaced, dropping those entries from RAM and leading to the same duplicate-id outcome by a different route. `publish()` now takes the same monitor as `load()`. It is reached only when a batch carries a value never seen before, so steady-state ingest pays nothing, and the lock order is one-way (`internLock` then monitor, never the reverse).

### v0 fixture - added, without a binary fixture

`aTypeStoredAsVersion0OpensWithTheInlineFormat` creates the type, closes the database, rewrites the stored schema to `mutableFormatVersion: 0`, deletes the `.tstd` component, and reopens with this build. It asserts the type comes back with **no** dictionary, the inline stride (`8 + 258 + 8`), tags round-tripping through both `query` and a `TagFilter`, and `toJSON` still emitting version 0 - the relabelling hazard you called out in the "what I liked" section.

The relabelling is done before any row exists, deliberately: a type whose pages already hold 4-byte ids could not honestly be called v0, and pretending otherwise would test the corruption rather than the compatibility.

### The two notes I did not act on

- **Follower reload cost.** Agreed it is O(N) per wave rather than per miss, and agreed it self-limits. Appending only the new tail pages is the right shape for it, but it needs the header to record a per-page entry offset to be safe against a partially written tail, so it is a follow-up rather than a line change.
- **`TimeSeriesShard`'s short constructor.** Confirmed test-only, as you found. Rather than remove it I documented the assumption on it - it passes `CURRENT_VERSION` unconditionally, so using it to open a type whose stored version says otherwise would read inline tag bytes as ids - which is the thing a future non-test caller would need to know.

### Verification

Ran on this base, clean target: 486 tests, 0 failures in `com.arcadedb.engine.timeseries.**`, and 950 across `com.arcadedb.schema.**` + `com.arcadedb.engine.**`.

PR #5575, opened 2026-07-30
fix: escape backslashes inside back-tick quoted SQL names

robfrank:
Fixes the CodeQL `js/incomplete-sanitization` alert
[#2311](https://github.com/ArcadeData/arcadedb/security/code-scanning/2311), the same defect at its source in the
engine, and every other place that builds a SQL statement by hand from a name or value it does not control.

> **Security note.** Tracing the alert surfaced a pre-existing SQL injection in the MongoDB wire protocol,
> unrelated to the grammar change and more serious than the alert itself. It is described below and fixed here.

## The defect

A schema object name may contain a back-tick or a backslash - all of `` Back`Tick ``, `Trail\` and `Mid\Slash`
are accepted by `createDocumentType`. The SQL spelling of such a name escapes it with a leading backslash, but
only the back-tick was ever escaped, which left the encoding ambiguous in both directions.

**Double escaping.** `setQuotedStringValue()` handed the already-escaped inner text to `setStringValue()`, which
escaped it a second time. The spelling grew one backslash per parse -> re-emit cycle, and the plain name came
back wrong, so addressing the type by name failed:

```
SELECT FROM `a\`b`     -- the type named a`b
  re-emitted as        SELECT FROM `a\\`b`
  resolved as          a\`b   ->  "Type with name 'a\`b' was not found"
```

**Token absorbing.** The lexer rule matched a backslash as an ordinary character and had no `\\` escape, so it
could not tell a closing back-tick from an escaped one. It ran past the end of the name and swallowed the SQL
after it:

```
UPDATE `T\` SET `v` = 1     -- " SET `" is consumed into the type name
```

The same applied to a bucket or index addressed through a `schema:<kind>:<name>` target, which is the path the
Studio uses.

## MongoDB protocol: SQL injection through field names and values

A MongoDB command is translated into SQL, and the field names and values taken off the wire were embedded with no
escaping. Any client able to send an `update` or `delete` could rewrite the statement. Verified against a running
server:

| Vector | Payload | Result |
| --- | --- | --- |
| Filter **value** | `updateMany({name: "v1' OR 'x' = 'x"}, {$set:{pwned:"yes"}})` | modified **all 5** documents, not 0 |
| `$unset` **field name** | key `` harmless`, `victim `` | destroyed the `victim` property, never named by the client |
| Filter **field name** | key `1 = 1 OR name` | emitted `1 = 1 OR name = 'nomatch'` - unquoted predicate |

The value vector is the reachable one: it applies to every update and delete filter, with no precondition. The
`find`/`count` path does not use this translator, which is why it is unaffected.

Fixed by escaping values for the literal they sit in (`Expression.encodeSingle`) and back-tick quoting every field
name one dot-separated segment at a time, so `a.b` stays navigation instead of becoming a single property named
`a.b`.

Escaping closes this. **Binding the values as parameters is the better shape** for the translator and is left as
follow-up work.

## The fix

A backslash now escapes the character that follows it, consistently across the lexer and every place that
applies or reverses the quoting:

- `SQLLexer.g4` - `QUOTED_IDENTIFIER` and `SCHEMA_QUOTED_NAME_PART` now read:

  ```antlr
  BACKTICK ( ~[`\\] | '\\' . )+ BACKTICK
  ```

- `Identifier` - a single-pass `escape`/`unescape` pair; `setQuotedStringValue()` stores the inner text verbatim
  instead of re-escaping it. Both keep a no-allocation fast path for the common unescaped name.
- `SchemaIdentifier.quoteName`/`unquoteName` - reuse those helpers rather than carrying their own partial copy.
- `studio-utils.js` `quoteSqlName()` - escapes both metacharacters.
- Every remaining schema-name interpolation in the Studio - 31 call sites across `studio-database.js`,
  `studio-record-editor.js` and `studio-timeseries.js` - now routes through `quoteSqlName()`. Most had **no**
  escaping at all, including the enclosing type name in `CREATE INDEX ... ON` and `CREATE TYPE`, which the
  grammar change would otherwise have silently altered (`validateTypeName()` rejects a back-tick but not a
  backslash).
- `PostgresQuotedIdentifierRewriter.translateQuotedIdentifier` escaped only the back-tick when writing a
  back-tick identifier, so a Postgres double-quoted name carrying a backslash lost it. It now escapes both,
  matching the read path in the same class, which already had these semantics.
- `Identifier.quoted(spelling)` builds a node from a quoted spelling without running the inner text through
  `escape()` only to discard the result, as the two AST builder call sites were doing.
- `Identifier.quote(name)` is the single place that spells a name for embedding in SQL. The remaining hand-built
  statements now use it: `grpcw/ArcadeDbGrpcService.java` quoted only the back-tick, and
  `gremlin/ArcadeGraph.java` did not quote bucket names at all (2 sites).
- The six Studio inline `onclick` handlers that built SQL inside an HTML attribute now call
  `browseRecords`/`browseRecordsWithConnections`/`countRecords`, so the statement is assembled in JS where
  `quoteSqlName()` applies and the name reaches the attribute through a single HTML escape - the same shape as the
  sibling buttons beside them.

## Breaking change

A literal backslash inside a quoted name has to be doubled - `` `C:\data` `` becomes `` `C:\\data` ``. A single
backslash directly before the closing back-tick now escapes it, leaving the name unterminated, so the statement
is rejected instead of silently absorbing what follows. Names without a backslash, including those with an
escaped back-tick, are unaffected. Recorded under Breaking Changes in `docs/release-26.8.1.md`.

## Verification

Both fixes were confirmed load-bearing by reverting them and watching the new tests fail:

- With the production change reverted, every new assertion fails, with the bug in the message
  (`Type with name 'Back\`Tick' was not found`).
- With **only** the grammar reverted, `backslashTerminatedNameDoesNotAbsorbFollowingSql` fails again - the
  `Identifier` fix alone is not sufficient.
- `IdentifierTest.backslashWithoutBackTickPreserved` pinned the old internal encoding and was updated; its
  behavioural assertion (a backslash in a name round-trips) is unchanged.

New tests:

- `QuotedIdentifierEscapingTest` (14) - escaping contract, re-emission idempotency, `schema:` name parts.
- `QuotedSchemaNameInjectionTest` (6) - end-to-end against a database: hostile names stay addressable, a
  following `WHERE` is not absorbed, a name crafted to close the identifier and append `DROP TYPE` is treated as
  a plain name, and a created type keeps the name it was given.
- `PostgresQuotedIdentifierRewriterTest` (+2) - back-tick and backslash escaped on the write path.

The Studio regression was confirmed by executing the unescaped spelling the Studio used to emit:

```
CREATE DOCUMENT TYPE `My\Type`   ->  existsType("My\Type") = false
                                     existsType("MyType")  = true
```

`MongoDBSqlInjectionTest` (4) - end-to-end through the Mongo wire protocol: a quoted value cannot extend the
`WHERE` clause, a back-tick in a `$unset` name cannot name a second property, a crafted filter field name cannot
become a predicate, and ordinary filters/updates still work. The three injection cases **fail without the fix**;
the fourth passes either way, so it is a real control.

Suites run green: full engine module (10288), OpenCypher (3820), postgresw (250), grpcw (172), mongodbw (27),
gremlin (BUILD SUCCESS).

## Left out of scope

- **Parameter binding for the MongoDB translator.** Escaping is correct, but binding values would remove the
  class of bug rather than the instance. Worth a follow-up issue.
- **The Studio's HTML-attribute escaping.** The inline-handler pattern - `onclick='dropType("<escaped>")'` - is
  used by roughly ten buttons and carries a documented double-escape hazard for names with embedded quotes
  (`studio-database.js:3345-3354`, where the remedy is a `data-*` attribute read via `dataset`). This PR removes
  the *SQL* layer from those attributes; converting the handlers themselves should be done uniformly across all
  of them, not for six in isolation.

`server/mcp/tools/MCPToolUtils.java:133` was reviewed and is safe: it rejects any identifier containing a
back-tick, and Cypher quotes by doubling rather than with a backslash.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -170,8 +170,8 @@ A single backslash immediately before the closing back-tick escapes it, so the n
 the statement is rejected rather than silently absorbing what follows. Only names that actually contain a
 backslash are affected; every other quoted name, including one containing an escaped back-tick, is unchanged.
 
-This is also the encoding the Postgres wire protocol already applied when it rewrote quoted identifiers, so the
-two agree now.
+The Postgres wire protocol already read back-tick identifiers this way; it now also writes them this way, so a
+double-quoted Postgres name carrying a backslash survives the translation instead of losing it.
 
 #### Cypher: an unbound `$parameter` is now an error, not null
 
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -5,6 +5,22 @@ they land during the 26.8.1 development cycle, so the release notes are ready at
 
 ### Fixes
 
+#### MongoDB protocol: field names and filter values can no longer inject SQL
+
+A MongoDB command is translated into a SQL statement, and the field names and values taken off the wire were
+embedded in it without escaping. Both could close their own quoting and append clauses of their own, so any
+client able to send an `update` or `delete` could rewrite the statement it was translated into:
+
+- A filter **value** containing a single quote closed the string literal, so
+  `updateMany({name: "v1' OR 'x' = 'x"}, ...)` produced `WHERE name = 'v1' OR 'x' = 'x'` and updated **every**
+  document instead of the intended one.
+- A `$unset` or `$inc` **field name** containing a back-tick closed the identifier and named further properties,
+  so a single crafted key removed a property the client never asked for.
+- A filter **field name** was appended unquoted and could introduce a predicate of its own.
+
+Values are now escaped for the literal they sit in and every field name is back-tick quoted, one dot-separated
+segment at a time so that navigation into an embedded document keeps working.
+
 #### Back-tick quoted names containing a backslash are no longer mis-parsed
 
 A schema object name may contain a back-tick or a backslash, and the SQL spelling of such a name escapes it with
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -5,6 +5,34 @@ they land during the 26.8.1 development cycle, so the release notes are ready at
 
 ### Fixes
 
+#### Back-tick quoted names containing a backslash are no longer mis-parsed
+
+A schema object name may contain a back-tick or a backslash, and the SQL spelling of such a name escapes it with
+a leading backslash. Escaping the back-tick but not the backslash left that encoding ambiguous, with two
+consequences.
+
+A name that arrived already escaped was escaped a second time, so the spelling grew one backslash on every
+parse and re-emission and the name no longer resolved:
+
+```sql
+-- the type named a`b
+SELECT FROM `a\`b`
+-- re-emitted as `a\\`b`, and addressing the type reported: Type with name 'a\`b' was not found
+```
+
+A name ending in a backslash left the closing back-tick indistinguishable from an escaped one, so the quoted
+token ran past the end of the name and absorbed the SQL that followed it:
+
+```sql
+-- SET ` is consumed into the type name instead of parsing as a clause
+UPDATE `T\` SET `v` = 1
+```
+
+The same applied to a bucket or index addressed through a `schema:<kind>:<name>` target.
+
+Both are fixed: a backslash now escapes the character that follows it, in the lexer and in the quoting applied
+by the engine and by the Studio, so the encoding is unambiguous in both directions.
+
 #### Concurrent writes to unrelated records of the same page no longer conflict
 
 ArcadeDB detects write conflicts per *page*, so two transactions that touched the same bucket page raised a
@@ -124,6 +152,26 @@ $ ls -l <database>/dictionary.*.dict
 remains downgradable.
 
 ### Breaking Changes
+
+#### SQL: inside a back-tick quoted name a backslash escapes the next character
+
+A backslash used to stand for itself unless it preceded a back-tick, so a name carrying one could be written
+bare. It now always escapes the character that follows it, which is what makes the closing back-tick
+unambiguous, and a literal backslash has to be doubled:
+
+```sql
+-- before
+SELECT FROM `C:\data`
+-- now
+SELECT FROM `C:\\data`
+```
+
+A single backslash immediately before the closing back-tick escapes it, so the name is left unterminated and
+the statement is rejected rather than silently absorbing what follows. Only names that actually contain a
+backslash are affected; every other quoted name, including one containing an escaped back-tick, is unchanged.
+
+This is also the encoding the Postgres wire protocol already applied when it rewrote quoted identifiers, so the
+two agree now.
 
 #### Cypher: an unbound `$parameter` is now an error, not null
 
--- a/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLLexer.g4
+++ b/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLLexer.g4
@@ -404,7 +404,7 @@ BUCKET_NUMBER_IDENTIFIER: B U C K E T COLON INTEGER_LITERAL;
 // user-supplied name may carry (spaces, colons, ...).
 SCHEMA_IDENTIFIER: S C H E M A COLON IDENTIFIER (COLON (SCHEMA_QUOTED_NAME_PART | SCHEMA_NAME_PART))?;
 fragment SCHEMA_NAME_PART: (LETTER | [0-9] | '_' | '[' | ']' | '.' | '-' | ',')+;
-fragment SCHEMA_QUOTED_NAME_PART: BACKTICK ( ~[`] | '\\`' )+ BACKTICK;
+fragment SCHEMA_QUOTED_NAME_PART: BACKTICK ( ~[`\\] | '\\' . )+ BACKTICK;
 
 // URL Identifiers
 HTTP_URL: 'http://' URL_CHAR+;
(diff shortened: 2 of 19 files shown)


robfrank:
Pushed in 5fdf4b54e. Items 1-3 were all correct; item 1 was worse than "non-blocking".

**1. Studio raw type names - this was a regression this PR introduced, and it was wider than two sites**

Confirmed empirically rather than by reading:

```
CREATE DOCUMENT TYPE `My\Type`   ->  existsType("My\Type") = false
                                     existsType("MyType")  = true
```

So the name is silently altered, and `validateTypeName()` does reject a back-tick but not a backslash, exactly as
described. Before the grammar change `` `My\Type` `` resolved to `My\Type`, so this is a regression, not a
pre-existing gap.

My original sweep was wrong. I had grepped for `"`" +`, which only finds a back-tick at the very start of a
string literal and misses every case where it sits inside a longer one (`" ON `" + typeName`). Re-grepping for a
back-tick adjacent to concatenation found **27** sites, not 8. All 21 remaining plain-SQL ones now go through
`quoteSqlName()`: the `CREATE INDEX ... ON` and `CREATE TYPE` names you named, plus `DROP`/`REBUILD`/`ALTER`
property, index and type, materialized views, graph analytical views, and the three timeseries statements in
`studio-timeseries.js`.

Regression test added in `QuotedSchemaNameInjectionTest.createdTypeKeepsTheNameItWasGiven`, and I verified the
committed `quoteSqlName` (loaded out of `studio-utils.js`, not reimplemented) emits `` `My\\Type` `` for that
input.

I left `validateTypeName()` alone. Escaping correctly is strictly better than rejecting, and most of these names
come from the schema rather than the input box, so validation would not have covered them. You are also right
that line 1617 drops every validation error when IF NOT EXISTS is checked, including "name is required" - that is
pre-existing and unrelated to escaping, so I have not touched it, but an empty name now fails closed at the
parser since `QUOTED_IDENTIFIER` requires at least one character.

**2. Postgres write path - also a regression, so fixed rather than deferred**

`translateQuotedIdentifier` escaped only the back-tick, so `"My\Type"` became `` `My\Type` `` and the engine read
`MyType`. That is the same silent loss, introduced by this PR, so it did not belong in the follow-up issue. It
now escapes both, matching `copyBackTickIdentifier` in the same class. Two tests added; the release-note line is
reworded from "the two agree now" to state that the read path already did this and the write path now does too.

**3. Throwaway `escape()` pass**

There were two of these, not one - `visitIdentifier` and `visitCreateIndexStmt` (~5561). Added
`Identifier.quoted(spelling)` and routed both through it, so the discarded pass is gone and the next caller
cannot reintroduce it.

**4. `getValue()` contract**

Agreed, no change. Recorded.

**Still deliberately out of scope**

The six inline `onclick` handlers (3534/3536/3537, 4222, 4266, 5055) embed a name in a
JS-string-in-HTML-attribute context. That is precisely the double-escape case the file already documents at
3345-3354, where the established remedy is a `data-*` attribute read back through `dataset`. That is a UI change
this fix does not need to make, so those are listed with the wire-protocol sites (`grpcw:3089`,
`mongodbw:482,559,640`, `gremlin/ArcadeGraph.java:306`) for the follow-up issue.

**Verification:** engine SQL + schema 2619, OpenCypher 3820, postgresw 250, all green; full engine module
re-running. On your note about the ANTLR regeneration - it is exercised: reverting only the grammar and leaving
the `Identifier` fix in place makes `backslashTerminatedNameDoesNotAbsorbFollowingSql` fail again, which is how I
confirmed both halves are load-bearing.

robfrank:
Pushed in f162f9ab7. All four follow-up sites are in, and tracing them turned up something bigger than the four
identifier sites: **the MongoDB wire protocol had a pre-existing SQL injection through filter values**, unrelated to
the grammar change in this PR.

Verified against a running server, not inferred:

| Vector | Payload | Before |
| --- | --- | --- |
| Filter **value** | `updateMany({name: "v1' OR 'x' = 'x"}, {$set:{pwned:"yes"}})` | modified **all 5** documents |
| `$unset` **field name** | key `` harmless`, `victim `` | destroyed `victim`, never named by the client |
| Filter **field name** | key `1 = 1 OR name` | emitted `1 = 1 OR name = 'nomatch'` |

The value vector is the reachable one - every update and delete filter, no precondition. `find`/`count` does not go
through this translator, which is why it is unaffected and why my first end-to-end probe misleadingly returned 0
until I added a control.

Fixes:

- **mongodbw** - values through `Expression.encodeSingle` (existing component); field names back-tick quoted one
  dot-separated segment at a time, so `a.b` stays navigation rather than becoming a property named `a.b`. Six sites,
  including `$inc` at 645 which I had missed when I first listed the follow-up.
- **grpcw:3089** and **gremlin:307/375** - through a new `Identifier.quote()`, so there is one place that spells a
  name for SQL.
- **Studio, the six inline `onclick` handlers** - now call `browseRecords`/`browseRecordsWithConnections`/
  `countRecords`. I did not do the `data-*` conversion: the ten sibling buttons at those same lines already use
  `onclick='dropType("<escaped>")'`, so converting six in isolation would have left the file less consistent than it
  is now. This removes the SQL layer from the attribute; the attribute pattern itself should be changed uniformly,
  and is called out as out of scope.

`MongoDBSqlInjectionTest` covers all three vectors plus a control; the three injection cases fail without the fix,
the control passes either way.

Green: full engine module 10288, OpenCypher 3820, postgresw 250, grpcw 172, mongodbw 27, gremlin BUILD SUCCESS.

Two things deliberately left: **parameter binding** for the translator's values (escaping closes this, but binding
would remove the class of bug), and the Studio HTML-attribute pattern. Both noted in the description.

Issue #5391, opened 2026-07-23
`LSM_VECTOR` index (JVector/HNSW) becomes unstable around ~200K-250K vectors on both 26.7.1 and 26.7.2 — GC thrashing, intermittent stalls escalating to write-path lock timeouts


kl-demi:
## Environment
- ArcadeDB versions tested: `26.7.1` (build `5347ff5fb225f2b1e3208faf7060facde020cb43`) and `26.7.2` (build `a53338edd645b92a8a866738976f03b6831b190b`) — **both exhibit the same failure signature**
- OS: macOS 26.5.2 (reproduced locally); original discovery on Linux/AWS, native systemd install, 7.6GB total RAM
- Java: OpenJDK 21.0.2
- JVM heap tested: 4GB (both local repros); also 768MB/4GB/6GB on the original production instance
- Standalone single node, `development` mode

## Schema
```sql
CREATE VERTEX TYPE Paper
CREATE PROPERTY Paper.id STRING
CREATE PROPERTY Paper.embedding ARRAY_OF_FLOATS
CREATE INDEX ON Paper (embedding) LSM_VECTOR METADATA {"dimensions": 768, "similarity": "COSINE"}
```
No other indexes/properties involved — stripped down from a larger production schema to isolate the vector index specifically.

## What we did
Loaded real 768-dimension sentence-embedding vectors (`allenai/specter2` model, from a production corpus) into a **freshly created, empty local database** in incremental batches, via `sqlscript`-batched `INSERT INTO Paper SET id = '...', embedding = [...]` (200 inserts/script call). After each batch, ran the same query repeatedly (3+ times, to distinguish transient stalls from persistent degradation):
```sql
SELECT expand(vector.neighbors("Paper[embedding]", <768-float-vector>, 5))
```

## Results

**26.7.1:**
| Records | Query time |
|---|---|
| 2,000-150,000 | 0.09s-0.79s (flat) |
| ~204,000 | 19.27s (single sample, not retried) |

**26.7.2** (retried 3x at each checkpoint from 210K on):
| Records | Query times (repeated samples) |
|---|---|
| 2,000-90,000 | 0.09s-0.79s (flat) |
| 110,000 | one 60s timeout, then 3 retries at 0.26-0.45s (transient hiccup, not sustained) |
| 150,000-190,000 | 0.54s-1.42s |
| 210,000 | 4.11s, 0.74s, 0.62s (variance appearing) |
[...]

PR #5578, opened 2026-07-30
fix(engine) #5568: publish the vector index location map atomically instead of clearing and refilling it

### LINKED ISSUE #5568: Vector index: countEntries() returns a wrong, run-dependent count after a manual compaction
### Summary

`TypeIndex.countEntries()` on a dense `LSM_VECTOR` index returns a **wrong, non-deterministic** number of entries after a manual compaction. `LSMVectorIndexTest.manualCompactionRemovesDuplicates` fails on `main` with a different count on each failing run.

Observed on `main` (26.8.1-SNAPSHOT, seen at `7c40815a3`):

```
[Should still have 30000 entries after compaction]
expected: 30000L
 but was: 13414L      <- one run
 but was: 21149L      <- another run
```

### Why this looks like a real bug and not a flaky test

The test is **single-threaded**: one transaction inserts 30,000 vertices with a 4-dim vector, then five sequential transactions each update the first 5,000 of them. Both counts *before* compaction are asserted and always pass:

- `countAfterInsert == 30000`
- `countAfterUpdates == 30000` (LSM merge-on-read, last-write-wins)

Only after `lsmIndex.scheduleCompaction()` + `lsmIndex.compact()` does the count go wrong. So compaction (or the location map it rewrites) is losing entries that merge-on-read had resolved correctly a moment earlier.

The counts are also suspicious in shape: 13,414 and 21,149 are neither `30000 - 5000` nor any obvious multiple, which suggests entries are dropped by whatever partitioning compaction walks rather than by a single clean off-by-one.

### Why it looks intermittent

The assertion is guarded:

```java
boolean compacted = lsmIndex.compact();
if (!compacted) { /* schedule + retry once */ }
if (compacted) {
  final long countAfterCompaction = typeIndex.countEntries();
  assertThat(countAfterCompaction)...isEqualTo(numVectors);
}
```

When both `compact()` attempts return `false` the test **silently passes without asserting anything**. So the intermittency may not be in compaction correctness at all - the bug may fire every time compaction actually runs, and the test only goes green when it does not. Worth confirming while investigating: the passing runs should be checked for whether `compacted` was even true.

### Reproduce

```
mvn -o -pl engine test -Dtest=LSMVectorIndexTest#manualCompactionRemovesDuplicates
```

Fails on a fraction of runs; when it fails, the count differs run to run. Reproduced both in a full `mvn -pl engine test` and in an isolated single-test run on a clean worktree at `main`, so it is not caused by leftover `target/databases` state.

### Notes

- Found while verifying #5279 / #5566 (concurrent page-conflict work). It is unrelated to that change: it reproduces on a pristine baseline worktree at `main` with none of those commits.
- Two things to tighten regardless of the root cause: the guarded assertion should fail (or at least report) when compaction did not run, so a skipped verification never reads as a pass; and per `engine/CLAUDE.md`, `countEntries()` on a dense vector index streams the whole location map, so it is worth confirming whether the wrong number comes from the location map or from the compacted series themselves.

lvca:
Fixes #5568

## The bug is real, but it is not compaction losing entries

`countEntries()` on a dense `LSM_VECTOR` returns `VectorLocationIndex.getActiveCount()`. A graph rebuild re-synced that index **in place**: `vectorIndex.clear()` followed by one `addOrUpdate` per live entry, under `lock.writeLock()`.

That lock only serializes the **writers** (`put`/`remove` take it). Every **reader** of the location index is lock-free by design - `countEntries()`, `getStats()`, the `vectorIndex.size() == 0` short-circuits on the three search paths, the RID filters. They observed the map mid-refill and got an arbitrary fraction of its real content.

A compaction *is* a rebuild (`compact()` -> `buildGraphFromScratch(null, true)`), which is why the count read right after `compact()` was the one that broke. With INFO logging on, the sequence is unambiguous:

```
01:01:56.032  Compacted vector index '...': 30000 live vectors in 1 pages (was file 4, now file 6)
01:02:10.888  Inactivity timeout expired (15000 ms), triggering graph rebuild for 80000 pending mutations
01:02:14.221  Built graph for index '...'          <- compaction's rebuild releases graphBuildLock, compact() returns
01:02:14.25x  <- the waiting inactivity rebuild starts refilling the location map
              [Should still have 30000 entries after compaction] expected: 30000L but was: 4976L
```

The 15-second inactivity timer fires while the compaction's 14-second JVector build holds `graphBuildLock`, and it starts clearing/refilling the map microseconds before the test reads it. Runs that "passed" simply read the count before the timer got in - which is also why the counts (4976, 7907, 13414, 19675, 21149, 21373 out of 30000) are arbitrary fractions rather than an off-by-one.

The count is only the visible symptom. A **search** running in the same window short-circuits on an apparently empty index and returns nothing.

## Fix

`vectorIndex` is now `volatile` and a rebuild publishes a **fully populated replacement instance** with a single reference assignment (`publishLocationIndex`), so a reader sees either the whole old set or the whole new one. Readers still take no lock. CPU cost is the same as the in-place refill it replaces; peak heap is not - the replaced instance stays reachable for its in-flight readers while the new one is built, so a rebuild transiently holds two location sets (~90 bytes per live entry each) instead of one. That spike is bounded, promptly released, and the price of the atomicity. On top of that:

- the compaction publishes the replacement **inside the same critical section that swaps the data file**, closing a pre-existing window in which the index answered offsets into a file that had just been dropped (a search landing there reads another entry's bytes, or the nulled compacted component);
- `getVectorLocation()` and `buildOrdinalMapping()` read the volatile field once so they cannot mix generations - the former would otherwise answer a miss against one generation and re-cache a stale offset into the next.

## Checked and dismissed as a second bug

`30000 live vectors in 1 pages` in the compaction log looks alarming but is correct: an unquantized `LSM_VECTOR` entry is ~8 bytes (varint id + varint bucketId + varint position + deleted flag + quantization-type flag; the vector itself stays in the document), so 30000 of them fit in one 256KB page. `MutablePage.checkBoundariesOnWrite` would have thrown otherwise.

## Verification

- **New `countEntriesIsNeverTornByAConcurrentRebuild`**: polls `countEntries()` from a second thread across three compactions and asserts it never dips below the real content. Without the fix it observes **0** within 0.9 s; with the fix, 5/5 clean runs.
- **`manualCompactionRemovesDuplicates`**: 5/5 clean runs (it failed on 4 of the first 5 baseline runs on `main`), and it now **fails when the compaction did not run** instead of silently verifying nothing - per the second follow-up in the issue.
- **Whole `com.arcadedb.index.**` package**: 911 tests, 0 failures.

## Note for whoever debugs the engine next

The engine's INFO logs are invisible in tests unless **both** `com.arcadedb.level` and `java.util.logging.ConsoleHandler.level` are lowered in `engine/src/test/resources/arcadedb-log.properties` - the handler filters at WARNING independently of the logger. Lowering only the logger looks like the logging is broken.

https://claude.ai/code/session_01XF4M78mPjg8VH7GR5pPZEp

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -112,4 +112,37 @@ Expected parameter(s): id
 
 To keep the old behaviour for a specific query, bind the name explicitly to `null`.
 
+## Vector index: the location cache no longer evicts, and `arcadedb.vectorIndex.locationCacheSize` is ignored
+
+**Behaviour change for operators.** `arcadedb.vectorIndex.locationCacheSize` (and the per-index
+`locationCacheSize` metadata) used to cap the in-memory location index of an `LSM_VECTOR` index and let it evict.
+That was never a cache bound. A location is the only record of which record a vector id belongs to and where its
+entry sits in the index file, and nothing on disk maps a vector id back to an offset, so an evicted entry could
+not be recovered - and every reader reads a missing location as deleted. A cap of 100 over 1000 live vectors made
+`countEntries()` report 100, and a query whose neighbours had been evicted dropped them
+([#5568](https://github.com/ArcadeData/arcadedb/issues/5568)).
+
+The limit dates from when the index held one location per *write*, so it grew with the write history. Issue #5516
+removed that: a tombstoned id releases its location, so residency now follows the number of **live** vectors.
+
+- **The setting is ignored** and reported once per index at `WARNING`. The `low-ram` profile no longer sets it.
+- **Plan for ~90 bytes per live vector** - about 90MB per million, 900MB at 10 million. This is the figure
+  `getStats()` now reports as `estimatedLocationIndexBytes`; it previously quoted the 24-byte payload and so
+  under-estimated the footprint several-fold.
+- **An index that appeared to work under a cap on a large corpus may now need a larger heap** instead of silently
+  under-reporting its size and losing vectors from searches. That is the intended trade: a wrong answer is worse
+  than a visible memory requirement.
+- A compaction transiently holds two copies of the location set while it publishes the new one.
+
+## Vector index: `countEntries()` is no longer torn by a concurrent graph rebuild
+
+A rebuild used to clear the live location index and refill it entry by entry. The write lock it held only
+serializes writers; `countEntries()` and `getStats()` take no lock, so they observed the map mid-refill and
+reported an arbitrary fraction of its real content - a different wrong number on each run, most visibly right
+after a `compact()`, because a compaction *is* a rebuild. A rebuild now publishes a fully populated replacement
+with a single reference assignment, so a reader sees either the whole old set or the whole new one, and the
+compaction publishes it inside the same critical section that swaps the data file - closing a window in which a
+search could resolve pre-compaction offsets against the new file
+([#5568](https://github.com/ArcadeData/arcadedb/issues/5568)).
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -129,6 +129,8 @@ removed that: a tombstoned id releases its location, so residency now follows th
 - **Plan for ~90 bytes per live vector** - about 90MB per million, 900MB at 10 million. This is the figure
   `getStats()` now reports as `estimatedLocationIndexBytes`; it previously quoted the 24-byte payload and so
   under-estimated the footprint several-fold.
+  [#5588](https://github.com/ArcadeData/arcadedb/issues/5588) tracks bringing that down to ~20 bytes by laying
+  the locations out in primitive arrays.
 - **An index that appeared to work under a cap on a large corpus may now need a larger heap** instead of silently
   under-reporting its size and losing vectors from searches. That is the intended trade: a wrong answer is worse
   than a visible memory requirement.
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -133,7 +133,12 @@ removed that: a tombstoned id releases its location, so residency now follows th
   the locations out in primitive arrays.
 - **An index that appeared to work under a cap on a large corpus may now need a larger heap** instead of silently
   under-reporting its size and losing vectors from searches. That is the intended trade: a wrong answer is worse
-  than a visible memory requirement.
+  than a visible memory requirement. Size it before upgrading if you had capped a very large corpus - 100M live
+  vectors is ~9GB resident.
+- **`estimatedLocationIndexBytes` steps up ~3.75x on upgrade** (24 -> 90 bytes per entry). Nothing changed about
+  the memory it describes; the stat used to quote the payload rather than the retained heap. Any dashboard or
+  alert threshold wired to it needs re-basing, and the same is true of `countEntries()` and `totalVectors` on an
+  index that had been capped - those were reporting the cap, not the index.
 - A compaction transiently holds two copies of the location set while it publishes the new one.
 
 ## Vector index: `countEntries()` is no longer torn by a concurrent graph rebuild
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -91,7 +91,6 @@ public enum GlobalConfiguration {
       } else if ("high-performance".equalsIgnoreCase(v)) {
         ASYNC_OPERATIONS_QUEUE_IMPL.setValue("fast");
         VECTOR_INDEX_GRAPH_BUILD_CACHE_SIZE.setValue(-1);
-        VECTOR_INDEX_LOCATION_CACHE_SIZE.setValue(-1);
         VECTOR_INDEX_SEARCH_CACHE_MAX_HEAP_PERCENT.setValue(50);
         VECTOR_INDEX_GRAPH_BUILD_CACHE_MAX_HEAP_PERCENT.setValue(50);
 
@@ -125,7 +124,8 @@ public enum GlobalConfiguration {
         SERVER_HTTP_IO_THREADS.setValue(cores > 8 ? 4 : 2);
         SERVER_HTTP_WORKER_THREADS.setValue(16);
         VECTOR_INDEX_GRAPH_BUILD_CACHE_SIZE.setValue(10_000);
-        VECTOR_INDEX_LOCATION_CACHE_SIZE.setValue(10_000);
+        // VECTOR_INDEX_LOCATION_CACHE_SIZE is deliberately NOT capped here: it is not a cache, and bounding it
+        // made this profile drop live vectors from searches (issue #5568).
         VECTOR_INDEX_SEARCH_CACHE_SIZE.setValue(10_000);
 
         POLYGLOT_ENGINE_ENABLED.setValue(false);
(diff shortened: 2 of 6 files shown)


lvca:
Thanks - I verified all four points against the code and all four are correct. Addressed in 49398d19b.

**1. Dead methods (confirmed, and it hides a live bug).** `getVectorLocation()`, `reconstructLocationFromPages()` and `buildOrdinalMapping()` have no callers, so those two edits have no runtime effect today. You are right that the description over-sold them - the load-bearing change is the `volatile` field + `publishLocationIndex()` + publishing inside the compaction critical section.

I chose to document rather than delete, because `getVectorLocation()` being unreachable is itself a bug worth keeping visible. It is the eviction fallback for a bounded `arcadedb.vectorIndex.locationCacheSize`, and every reader (the searches, the delta merge, the ordinal maps, the RID filters) calls `vectorIndex.getLocation()` directly instead. So under a bounded cache an evicted entry reads as `null`, is treated as deleted, and the vector silently disappears from searches until the next rebuild. Deleting the method would remove the only implementation of the path that ought to be wired up. Flagged in its Javadoc; out of scope for #5568.

**2. The "search short-circuit" framing was wrong.** Confirmed: all three search entry points take `lock.readLock()` (lines 3355 / 3581 / 3753) and hold it across the `vectorIndex.size() == 0` check and the whole body, while the refill ran under `lock.writeLock()`. Searches were never exposed to the torn refill. The only readers that took no lock are `countEntries()` and `getStats()`. Comments corrected.

Worth adding, since it is the reason the swap is still the right fix rather than just read-locking those two: the read lock could not have covered the compaction hazard. `rewriteDataFileWithLiveEntries()` swapped the data file in one write-locked section and the caller re-published the locations in a later one, so a search could acquire the read lock *between* them and resolve old offsets against the new file. That is what moving the publish inside the file-swap section fixes, and the comments now say so.

**3. `getStats()` multi-generation snapshot - fixed.** It read the index three times (`size()`, `getActiveCount()`, `getDeletedCount()`, plus `size()` again for `estimatedLocationIndexBytes`), so across a swap it could report four numbers describing an index that never existed. Now snapshots the reference once. This is the one actual behaviour change in the follow-up.

**4. Capacity overflow - fixed.** Sized through a `long` now.

Also from the earlier review pass: the "costs exactly what refilling cost" claim was corrected (CPU yes, peak heap no - the swap holds two location sets during a rebuild), the test documents its unbounded-location-cache assumption, and the counter thread spins with `Thread.onSpinWait()`.

Full `com.arcadedb.index.vector` suite re-run after the changes: 215 tests, 0 failures.

lvca:
Thanks. Two of the three led to a change; the first I'm deliberately not taking, with reasoning below.

**1. Critical-section width - keeping the population under the lock, now documented.** You're right that only the store needs the lock. But hoisting the loop out is not free: with the lock held, an insert cannot commit into the instance that is about to be discarded, so the window in which a concurrent write is lost stays exactly what the in-place refill had (`page scan end` -> `writeLock acquired`). Building outside would grow that window by the whole population. Holding it also matches what `clear()` + N x `addOrUpdate` already did, so this is not a new stall.

On magnitude: the loop is an in-memory map fill - tens of milliseconds on a 200K index. #5391 moved something different out of this lock, the O(index size) *document reads* in the validation/preload phase, which were minutes. Different order. If profiling ever shows this mattering, the right shape is optimistic build-outside + a writer generation counter checked under the lock, not an unguarded hoist. Recorded the trade-off in the Javadoc so nobody "optimizes" it into a wider loss window.

**2. Bounded-cache mode - this one found a real regression I had introduced.** Chasing the eviction-during-population detail you describe: I was sizing `initialCapacity` from `liveEntries.size()` *unconditionally*, so under a bounded cache the replacement allocated a table for the whole live set to hold `maxSize` entries. The in-place refill it replaced reused an already correctly-sized map, so that was a straight memory regression on exactly the profile that opts into bounding. Now `min(cap, live)` (2728d298c). Capacity is a sizing hint only, so nothing about the published content changes.

The eviction semantics themselves are unchanged from the in-place refill, so bounded mode is no worse than before this PR - but as you say, combined with the unreachable `getVectorLocation()` it still drops vectors from searches. I've asked the maintainer whether to file that as its own issue rather than open one unilaterally; agreed it should not live only in a code comment.

**3. Hard assert on `compacted`.** Noted, and I'd rather have the loud failure: the whole reason #5568 sat unnoticed is that this branch could pass while verifying nothing. The test already retries `scheduleCompaction()` + `compact()` once before asserting, and the message names the three preconditions to look at. If a 2-vCPU runner does start failing here, that is a scheduling problem worth seeing rather than skipping.

Full `com.arcadedb.index.vector` suite re-run after the change: 215 tests, 0 failures.

lvca:
Pushed 51a3fd258, which resolves the bounded-cache gap in this PR rather than deferring it, plus the two points from the last pass.

**`buildOrdinalMapping()` deleted**, along with `buildLiveOrdinalMapping()` - both private, both uncalled.

**`mvn -pl engine verify` confirmed green: 10270 tests, 0 failures, 0 errors, 23 skipped.** An earlier run of the same suite had one failure, `TypeConversionTest.sqlMath` (a `LocalDateTime.now()` / `Duration` test, unrelated to vectors); it passes in isolation and passed in the clean run, so it is flaky under full-suite load rather than affected by this change.

**Bounded location cache - fixed, and one correction to what I claimed earlier.**

I had said a cap caused ~80% search loss. That was wrong: the `10` I measured came from `TypeIndex.get()`, which returns 10 regardless, and an *uncapped* index returned 10 too. Measuring against the bucket-level index instead, with a cap of 100 over 1000 vectors:

| | before | after |
|---|---|---|
| `countEntries()` | **100** | 1000 |
| `lsmIndex.get(k=50)` | 50 | 50 |
| `vectorNeighbors(..., 50)` | 50 | 50 |

So the demonstrated harm is the count and `getStats()` under-reporting by 900. Search loss is still real by construction - the result loops read a missing location as deleted - but my small reproduction did not hit it, because its vectors were also in the in-memory delta buffer. Every comment, the config text and the test doc now say exactly that rather than the stronger claim.

The conclusion did not change, because eviction destroys the only mapping from a vector id to its record and nothing on disk can rebuild it. So the cap is gone: `getLocationCacheSize()` always returns -1 and reports a configured value once per index at WARNING, `arcadedb.vectorIndex.locationCacheSize` is documented as deprecated/ignored, and the `low-ram` profile no longer sets it. `LowRamProfileTest` asserts that, with the reason.

**Worth knowing for the "wire the readers through `getVectorLocation()`" option:** it was not merely unreachable, it could not have worked. `scanPagesForVectorId` parses `vectorId`/`bucketId`/`position` as fixed-width `readInt`/`readInt`/`readLong` where the writer emits varints (`writeNumber`), and builds `absoluteFileOffset` as `((long) fileId << 32) | offset` where every other path uses `pageNum * pageSize + PAGE_HEADER_SIZE + offset`. Routing readers through it would have returned garbage offsets. Deleted, with `reconstructLocationFromPages()` and `calculateQuantizedSize()`.

New guard: `aConfiguredLocationCacheLimitDoesNotDropVectors` - verified to fail (100 vs 1000) with eviction restored. A class-level note on `VectorLocationIndex` says not to wire the bounded backend back in without first giving the file a lookup that makes a miss recoverable.

Net for the commit: -108 lines.

lvca:
Took all three. Pushed 20e03699b.

**1. `estimatedLocationIndexBytes` was quoting the wrong number - fixed.** You're right that this is the inconsistency that matters most, because that stat is what an operator sizes a heap from: reporting the 24-byte *payload* under-estimates the location index several-fold. The repo actually carried three figures (24 in the class header and the stat, 56 in the old config text, 90 in the #5516 Javadoc). The 90 is the defensible one - #5516 enumerates it as the `VectorLocation`, its `RID`, the boxed `Integer` key, the map node and its table slot, and the reverse-index entry. Both the stat and the docs now read `VectorLocationIndex.APPROX_RETAINED_BYTES_PER_LOCATION`, and the class Javadoc says explicitly which of the two numbers is payload and which is retained heap.

**2. `locationCacheCapReported` race - fixed.** Now `AtomicBoolean.compareAndSet`, so the WARNING is emitted exactly once per index rather than once per thread that lost the race.

**3. Release notes - added,** and thanks for pushing on this; it is the point most likely to bite someone. `docs/release-26.8.1.md` now carries both changes, and the cap removal is written for operators rather than for readers of this diff: the setting is ignored, plan for ~90 bytes per live vector (~90MB per million, 900MB at 10M), an index that appeared to work under a cap on a large corpus may now need a larger heap, and a compaction transiently holds two copies. Stated as the intended trade - a visible memory requirement beats a wrong answer - so nobody reads it as an accidental regression.

**On verification:** `mvn -pl engine verify` is green on the tree as pushed - **10270 tests, 0 failures, 0 errors, 23 skipped, BUILD SUCCESS**. (An earlier run of the same suite had one failure, `TypeConversionTest.sqlMath`, a `LocalDateTime.now()`/`Duration` test unrelated to vectors; it passes in isolation and passed in the clean run.) The vector package plus `LowRamProfileTest` re-run after this commit: 219 tests, 0 failures.

The primitive-array layout you point at stays the right next step - ~20 bytes per id with no per-entry objects would take a 10M-vector index from ~900MB to ~200MB and remove the pressure that made a cap look attractive in the first place. Deliberately not in this PR.

PR #5581, opened 2026-07-30
fix(#5579): bind mongodbw filter values as SQL parameters

### LINKED ISSUE #5579: mongodbw: bind filter values as parameters instead of escaping them into the statement
Follow-up to #5575.

That PR closed a SQL injection in the MongoDB wire protocol by **escaping** the field names and values that get
embedded into the translated statement. Escaping is correct, but it is per-call-site: every future edit to
`MongoDBToSqlTranslator` has to remember to do it, and forgetting is silent. Binding the values as parameters
removes the class of bug instead of the instance.

### Current shape

`MongoDBToSqlTranslator` builds one SQL string and the caller executes it with no parameters:

```java
// MongoDBDatabaseWrapper
private int executeCount(final String sql) {
  try (final ResultSet rs = database.command("sql", sql)) { ... }
}
```

Values are inlined into the text by `buildValue`, which since #5575 escapes them via `Expression.encodeSingle`:

```java
protected static void buildValue(final StringBuilder buffer, final Object value) {
  if (value instanceof String string) {
    buffer.append('\'');
    buffer.append(Expression.encodeSingle(string));
    buffer.append('\'');
  } else
    buffer.append(value);
}
```

### Proposed change

Thread a `Map<String, Object>` through the builders, emit `:p0`, `:p1`, ... in place of inlined literals, and pass
the map to `database.command("sql", sql, params)`.

Methods that would need the extra argument: `buildExpression` (both overloads), `buildAnd`, `buildOr`,
`buildCollection`, `buildValue`, `appendWhere`, `appendUpdateOperations`, and the `executeCount` / `executeUpdate` /
delete call sites in `MongoDBDatabaseWrapper`.

**Identifiers stay escaped.** SQL cannot bind a type or property name as a parameter, so field names keep going
through `Identifier.quote()` / `quoteFieldPath()`. This issue is only about values.

### Notes

- `quoteFieldPath()` splits on `.` and quotes each segment, so MongoDB navigation into an embedded document keeps
  resolving. Do not collapse that into a single `Identifier.quote()` call.
- `find` / `countDocuments` do **not** go through this translator (they are served by the embedded
  mongo-java-server backend), so they are out of scope here. Only the update/delete path reaches `appendWhere`.

### Verification

`MongoDBSqlInjectionTest` already covers the three injection vectors plus a control and must stay green. Worth
confirming the tests still fail if the change is reverted, since a parameterised build could pass them for the
wrong reason (for example by erroring out rather than binding correctly).

robfrank:
Closes #5579

Follow-up to #5575. That PR closed the MongoDB wire-protocol injection by escaping field names and values into the translated statement. Escaping is correct but per-call-site: every future edit to `MongoDBToSqlTranslator` has to remember it, and forgetting is silent. `buildValue` now registers the value in a `Map<String, Object>` under `p<n>` and emits only `:p<n>`, so nothing the client sent reaches the statement text and the injection is unreachable by construction rather than by discipline. Binding also preserves the value's Java type, which fixes a second defect the inlining hid: spelling a non-`String` went through `String.valueOf`, so an `ObjectId` or a `java.util.Date` filter value emitted text no SQL parser accepts.

The map is threaded through `buildExpression` (both overloads), `buildAnd`, `buildOr`, `buildCollection`, `appendWhere`, `appendUpdateOperations` and every call site. Note that the issue lists two entry points; there are in fact **three**. `find` does reach the translator, via `MongoDBDatabaseWrapper.find` -> `handleQuery(MongoQuery)` -> `MongoDBCollectionWrapper.queryDocuments`, which builds its own `SELECT`. Only `countDocuments` is served by the embedded mongo-java-server backend. `$in` / `$nin` bind the whole collection to a single parameter, since the grammar accepts an input parameter between the `IN` parentheses.

**Identifiers still go through `Identifier.quote()` / `quoteFieldPath()`**, since SQL cannot bind a type or property name, and `quoteFieldPath` still quotes one dot-separated segment at a time so MongoDB navigation into an embedded document keeps resolving. Two identifiers in `queryDocuments` were still concatenated raw and now use those same helpers: the collection name and the `orderBy` field names.

To be precise about that last one: `orderBy` comes from `queryObject.remove("$orderBy")`, and the modern `find` command never populates that key. So the branch is reached only via the legacy `OP_QUERY` wrapper or a `mongo`-language query through `MongoQueryEngine`. It is untrusted query text either way and quoting it is right, but it is not the straightforwardly remote-attacker-controlled path an earlier draft of this description claimed.

### Scope boundary

`$set` and full-replacement values still travel as an inlined JSON literal (` MERGE <json>` / ` CONTENT <json>`), not as bound parameters. That path is safe because `JSONObject` does its own escaping - and this PR now has tests proving it rather than assuming it - but the "unreachable by construction" property therefore covers the `WHERE` clause and `$inc`, not the update values. Binding those needs `MERGE`/`CONTENT` to accept a parameter in place of a JSON literal, which is a real change to the update path and outside #5579's scope. Worth a follow-up.

## Test plan

- [x] `MongoDBToSqlTranslatorParamsTest` (new, no server, 14 cases) calls the `protected static` builders directly and asserts the SQL carries placeholders, the values never appear in the text, and the map holds the raw objects. Covers equality, comparison operators, `$in`/`$nin`, `$or`, `$and`, `$size`, nested documents, dotted paths, null, and non-`String` types.
- [x] **Proven to fail before being trusted**: reverting only `buildValue` to the previous escaping form turns all 14 cases red; restored and re-run green.
- [x] `MongoDBParameterBindingTest` (new, wire level, 10 cases) drives a real server through the Mongo driver: quote-bearing, backslash-bearing and numeric values round-trip through `find`, `updateMany` and `deleteMany`; `$in` covered on both the find and update paths; `$nin` positive match; and `$set` / `replaceOne` values carrying `'`, `"` and `\`. Each case pins a positive match so it cannot pass for the wrong reason.
- [x] `MongoDBSqlInjectionTest` (the #5575 regression suite) stays green **unchanged**.
- [x] `mvn -pl mongodbw verify` -> 49 tests, 0 failures. `MongoQueryMetricsIT` run separately, green.

### DIFF
--- a/docs/5579-mongodbw-parameter-binding.md
+++ b/docs/5579-mongodbw-parameter-binding.md
@@ -47,9 +47,13 @@ name, so that half of #5575 is unchanged. `quoteFieldPath` still splits on `.` a
 MongoDB navigation into an embedded document keeps resolving.
 
 Two identifiers in `queryDocuments` were still being concatenated raw and are now quoted with the same helpers:
-the collection name and the `orderBy` field names. The `orderBy` key is attacker-controlled text on the `find`
-path with no validation in front of it, so leaving it raw next to a bound value would have left the same hole
-open one line away.
+the collection name and the `orderBy` field names.
+
+On reachability, to be precise: `orderBy` comes from `queryObject.remove("$orderBy")`, and the modern `find`
+command never populates that key (`MongoDBDatabaseWrapper.find` builds its `MongoQuery` with a null sort). So the
+`orderBy` branch is reached only through the legacy `OP_QUERY` wrapper or through a `mongo`-language query run via
+`MongoQueryEngine`. It is untrusted query text either way and quoting it is correct, but it is not the
+straightforwardly remote-attacker-controlled path the first draft of this note claimed.
 
 ## Verification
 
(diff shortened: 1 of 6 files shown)


robfrank:
The inlined-JSON update-value path from observation 1 is now filed as #5583, with the grammar question (whether `UPDATE ... MERGE` / `CONTENT` can take an input parameter at all) called out as the thing to settle first.

PR #5584, opened 2026-07-30
feat(mcp) #4866: add Prompts (graphrag_query, build_knowledge_graph)

### LINKED ISSUE #4866: MCP: add Prompts (graphrag_query, build_knowledge_graph)
Part of #4859.

## Motivation

MCP Prompts are entirely unimplemented (confirmed alongside Resources — zero matches for `prompts/list`/`prompts/get`). Guided prompt templates steer models toward correct, safe use of the new retrieval/upsert tools instead of reaching for raw `execute_command` with hand-written vector/MERGE syntax — this is architecturally higher-leverage than adding more tools, per the governing principle in #4859.

## Proposed design

- Implement `prompts/list` and `prompts/get` in both transports; advertise `capabilities.prompts = {listChanged: false}`.
- `graphrag_query` prompt: arguments `{database, question}`; template instructs the model to call `vector_search`/`hybrid_search`/`full_text_search` as appropriate and cite retrieved `rid`s.
- `build_knowledge_graph` prompt: arguments `{database, sourceText}`; template instructs the model to extract entities/relationships and call `upsert_entity`/`upsert_relationship`, emphasizing match-key selection to avoid duplicates.
- New package `server/src/main/java/com/arcadedb/server/mcp/prompts/` mirroring the `tools/` pattern (one class per prompt with `getDefinition()`/`getMessages(args)`).

## Acceptance criteria

- [ ] `prompts/list` returns both prompts with argument schemas.
- [ ] `prompts/get` returns well-formed message arrays referencing actual, currently-registered tool names (test should fail if a referenced tool is renamed/removed).
- [ ] Both transports covered; tests + docs (including the literal prompt text for review).

## Files to touch

`server/src/main/java/com/arcadedb/server/mcp/prompts/GraphRagQueryPrompt.java`, `BuildKnowledgeGraphPrompt.java` (new), `MCPHttpHandler.java`, `MCPStdioServer.java`, tests, docs.

## Dependencies

Blocked-by #4860 (`vector_search`), #4861 (`hybrid_search`), #4864 (`upsert_entity`/`upsert_relationship`) — prompt text references these tools by name, so sequence after those tool names are finalized.

## Effort estimate

M — new protocol primitive, prompt-text design/iteration, cross-reference test.

robfrank:
## What does this PR do?

Implements the MCP **Prompts** primitive, the last protocol primitive missing from ArcadeDB's MCP server.

- `prompts/list` and `prompts/get` on both transports, and `capabilities.prompts = {listChanged: false}` in `initialize`. Both transports already route through `MCPDispatcher`, so they gain the feature from one change.
- Two guided templates in a new `com.arcadedb.server.mcp.prompts` package:
  - `graphrag_query(database, question)` steers retrieval: load the schema, pick `vector_search` / `full_text_search` / `hybrid_search` / `query` by the shape of the question, bring your own embedding vector, cite every claim with its `@rid`.
  - `build_knowledge_graph(database, sourceText)` steers construction: read the schema first, choose a stable match key per entity, prefer a UNIQUE index, then call `upsert_entity` / `upsert_relationship` reusing exactly those keys.
- `MCPPrompts`, the registry, mirroring the existing `MCPResources`.

**Gating.** A prompt is listed only when every tool its text names is allowed by the caller's effective tool profile *and* every permission flag those tools need is on. So the `admin` profile sees no prompts, and `build_knowledge_graph` appears only once insert and update are enabled. `prompts/get` re-checks the same rule, the way `tools/call` re-checks the profile, so a hidden prompt cannot be fetched by name. Availability is evaluated before arguments are validated, so a caller who may not use a prompt cannot learn which of its arguments were malformed.

Defaults are unchanged: `enabled=false` with all write flags `false`, and this PR adds no configuration flag.

## Motivation

Wave 1 of #4859 added six retrieval and agent-memory tools whose value depends on the model choosing them, and whose safety depends on the model understanding them. An upsert whose match key differs by one property silently creates a duplicate vertex instead of updating the intended one; absent guidance, a model reaches for `execute_command` with hand-written `vector.neighbors` or `MERGE` syntax, which is exactly the failure mode the epic's governing principle exists to prevent.

Prompts are the primitive for that guidance. Like #4865, this adds capability through the right protocol primitive instead of lengthening the tool list.

## Related issues

- Closes #4866
- Part of epic #4859
- Depends on (all merged): #4860, #4861, #4864, #4865, #4867

## Additional Notes

**Prompt text is static.** Arguments are substituted, no database is opened to render a prompt, and `{database}` is not validated. Schema hydration was considered and rejected: #4865 already delivers schema without spending a turn, so embedding it here would create a second schema emitter and pull read permissions, unknown-database errors and unbounded message size into the Prompts path. Substitution is single-pass, so one argument's value cannot be reinterpreted as another's placeholder.

**A test fails if prompt text names a tool that is renamed or removed.** Each prompt declares `referencedTools()`; one test asserts that set is a subset of `MCPDispatcher.REGISTERED_TOOL_NAMES`, and a second scans the rendered text for every registered tool name and asserts each hit is declared. The scan matches on word boundaries, so `query` inside `graphrag_query` is not a false hit. This is also the input to the gating rule, so the two cannot drift.

**Design document:** `docs/superpowers/specs/2026-07-30-mcp-prompts-4866-design.md`, committed here with the implementation plan.

**Out of scope:** `notifications/prompts/list_changed`, any third prompt, per-database prompt variants (#4868), embedding generation, and the `arcadedb-docs` MCP page, which is a follow-up as it was for #4862 and #4865.

**Known follow-up, deliberately not fixed here.** `sourceText` is substituted verbatim inside a `<source_text>` fence, so a document containing a literal `</source_text>` escapes the fence and its following text reads as procedure. No privilege is gained: the rendered prompt is returned to the calling client rather than executed, and every tool it could reach still enforces its own per-database permission checks. Neutralizing it would mean silently altering caller data, which the design's verbatim-substitution rule rules out, so it is flagged for a maintainer decision rather than patched.

## Checklist

**Test evidence.** The MCP suite goes from 224 to 259 tests: `MCPPromptsTest` (23, new), plus additions to `MCPServerPluginTest` and `MCPStdioServerTest`. No existing test method was edited; `git diff --stat` over the MCP test package is additions only.

Full local run after `mvn clean`: **282 tests, 0 failures, 0 errors** across the whole `server` MCP and profiler surface, including `ServerQueryProfilerTest` 20/20 and `EngineMetricsBinderTest` 2/2.

~~Two tests fail on the machine this was developed on ... a JDK newer than 21 no longer exports `sun.management` ...~~ **This earlier claim was wrong and is retracted**; see the correction comment below. The two profiler failures were an IDE artifact in the local `target/classes`, not a JDK incompatibility, and they do not reproduce after `mvn clean`.


robfrank:
### 1. Single-pass substitution: the review read a superseded commit

The two-`String.replace` code quoted here is not in this PR's head. It existed in `bfb3e7420` and `ef2d73d41` and was replaced in `e728e785a`, three commits before the branch was pushed. Both prompts render in one `Matcher` pass:

```java
final Matcher matcher = PLACEHOLDER.matcher(TEMPLATE);
final StringBuilder rendered = new StringBuilder();
while (matcher.find()) {
  final String value = "database".equals(matcher.group(1)) ? database : question;
  matcher.appendReplacement(rendered, Matcher.quoteReplacement(value));
}
matcher.appendTail(rendered);
```

`PLACEHOLDER` is `\{(database|question)\}` in `GraphRagQueryPrompt` and `\{(database|sourceText)\}` in `BuildKnowledgeGraphPrompt`. Because the matcher scans `TEMPLATE` and appends into a separate buffer, a value is never re-scanned, so the asymmetry described here does not arise. `Matcher.quoteReplacement` additionally keeps a value containing `$` or `\` from being read as a replacement reference.

The invariant is covered by two tests, `graphRagQuerySubstitutesInOnePassWithoutReinterpretingAnotherPlaceholder` and `buildKnowledgeGraphSubstitutesInOnePassWithoutReinterpretingAnotherPlaceholder`, which pass `database = "x{question}y"` and `database = "x{sourceText}y"` respectively and assert the literal token survives unexpanded. Both were confirmed to fail against the old two-pass code before the fix was committed.

So the PR body's claim holds as written, and no wording change is needed. Worth checking whether the review ran against an intermediate commit rather than the branch head, since that would affect any other file it looked at.

### 2. `KNOWN LIMITATION` comment: added

Fixed in `646e3827d`. The tradeoff now sits on `BuildKnowledgeGraphPrompt.getMessages`, immediately above the render, rather than only in this description: what the escape is, why it is accepted (the rendered prompt goes back to the caller that supplied the document and is never executed here, and every tool the text names still enforces its own permission and per-database checks when called), and why escaping it was rejected (it would silently alter the caller's document, against the verbatim-substitution rule the prompt is built on).

### 3. Consistency notes

Both confirmed intentional. The `prompts` capability is deliberately advertised unconditionally: capabilities describe what the server implements, and `prompts/list` is what filters per caller, matching how `tools` and `resources` are already advertised. An `admin`-profile caller spends one call on an empty listing, the same cost `resources/list` already carries on a reads-disabled server.

robfrank:
### 4. Malformed `arguments`: fixed, and it was worse than described

Fixed in `f85a375ea`. The diagnosis was right and the consequence was one step further than stated: `args` was read *outside* the `try`, so the throw escaped `promptsGet` and then `dispatch` (whose `try` carries only a `finally`), and the transport turned it into an **HTTP 500** rather than a `-32603` body. A test proves it, and it reproduced exactly that way before the fix:

```
promptsGetRejectsNonObjectArguments
  » IO Server returned HTTP response code: 500 for URL: http://127.0.0.1:2480/api/v1/mcp
```

Both members are now read inside the `try`, and `IllegalStateException` joins `IllegalArgumentException` on the `-32602` arm. `prompts/get` with `"arguments": ["database", "graph"]` now answers `-32602`. `MCPPromptsTest` 23/23, `MCPServerPluginTest` 130 tests, `MCPStdioServerTest` 19, with only the two known profiler errors.

**Pre-existing twin, outside this PR's scope:** `toolsCall` has the identical shape, confirmed unchanged on `main` at `3cfa4197c`:

```java
final String toolName = params.getString("name", "");
final JSONObject args = params.getJSONObject("arguments", new JSONObject());   // outside the try
```

So `tools/call` with a non-object `arguments` returns an HTTP 500 on `main` today. That is a separate defect and belongs in its own change rather than being folded into a Prompts PR. Flagging it for the maintainer to file.

### 2. Prompt registry / shared render loop: declining for now

Both prompt classes are two members and a five-line render, and the duplication is the deliberate cost of mirroring `MCPResources`, which the review notes. A descriptor array would pay abstraction now for a third prompt the design explicitly puts out of scope, and the cross-reference test already prevents the drift that duplication usually invites. The right trigger is a third prompt actually landing, not the anticipation of one. Note #4868 is per-database endpoint scoping, not more prompts, so it would not by itself force this.

### 1. Tracking issue for the fence limitation

Agreed the decision should not live only in a PR description and a Javadoc. Left for @robfrank to file, since it records a maintainer decision about the verbatim-substitution rule rather than a defect in this change.

### 3 and 5

No action, and the readings match the intent. On 3, note `MCPDispatcher.INSTRUCTIONS` already carries the same `arcadedb://{database}/schema` literal, so the templates are consistent with existing practice rather than introducing the coupling.

### CI

Agreed that a Java 21 toolchain is the authority on the two profiler failures. They are deterministic rather than flaky: `Profiler.toJSON` uses `sun.management.HotspotRuntimeMBean` and `ManagementFactoryHelper`, so on a newer JDK every path reaching the profiler fails, which is why the same root cause takes out 19 `ServerQueryProfilerTest` cases, `EngineMetricsBinderTest.gaugesReadNestedProfilerValuesWithoutThrowing` and the engine's `ProfilerTest.metricsToJSON` on `main` as well.

robfrank:
Both filed:

- #5585 - the malformed `params`/`arguments` HTTP 500. Scope turned out wider than `tools/call`: the `params` member itself is read above `dispatch`'s `try` (which carries only a `finally`), so `{"method": "ping", "params": ["a"]}` returns HTTP 500 too. Both sites reproduced empirically at `3cfa4197c`, with a failing JUnit case in the issue.
- #5586 - the `</source_text>` fence decision, with the three options and why option 3 (nonce-suffixed delimiter) is the only one that both leaves the caller's document byte for byte intact and makes the boundary real.

This PR stays scoped to Prompts; neither issue blocks it.

robfrank:
### Correction: the two profiler failures were not a JDK issue

Earlier in this PR I attributed `MCPServerPluginTest.profilerStartStopCycle` / `profilerStartAlreadyRecording` to `sun.management` internals being unexported on a JDK newer than 21, and said a Java 21 CI toolchain would confirm it. **That diagnosis was wrong.** The PR body has been corrected.

The real cause is local: the Eclipse/JDT language server running over the checkout compiles into `<module>/target/classes` with its own compiler, which does not get the `--add-exports java.management/sun.management=ALL-UNNAMED` that the Maven build passes. Maven then reuses those `.class` files because the sources are not newer, so every path reaching `Profiler.toJSON` fails at runtime.

The tell I walked past: `java.lang.Error: Unresolved compilation problems` is an ECJ artifact. `javac` never emits it, and a genuine JDK export problem would surface as `IllegalAccessError` at runtime or a compile failure in the Maven build, not as that message. It also explained too much - a real JDK incompatibility would not have left `mvn -pl server -am -DskipTests install` succeeding, which it did.

After `mvn clean`, on this same machine and JDK:

```
Tests run: 282, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
  ServerQueryProfilerTest        20/20
  EngineMetricsBinderTest         2/2
  MCPServerPluginTest           130/130   (both profiler tools included)
  MCPPromptsTest                 23/23
```

So there is nothing environmental to discount: this branch is green locally, in full.

`HttpRedMetricsIT` also passes here on this branch (3/3, `-Pintegration`), so the `integration-tests` red on the last CI run is not from this change either. The three `ha-integration-tests` failures (`Issue5410AbandonedTicketReleaseIT`, `RemoteStickyConnectionStrategyIT`, `RaftReplicationMaterializedViewIT`) are all in the Raft/HA suite, which this PR does not touch at all.

The branch is 21 commits behind `main`, so I am merging `main` in and pushing to get a CI run against current code rather than reasoning about a stale base.

PR #5594, opened 2026-07-30
fix(engine) #5569: extend the disjoint-slot merge to plain record deletes

### LINKED ISSUE #5569: Extend the disjoint-slot merge to plain record deletes (follow-up to #5279)
### Summary

Follow-up to #5279 (PR #5566). A plain single-record **delete** is the last common false-conflict class left on a bucket page: deleting record A still makes every concurrent transaction that merely updated record B on the same page fail with a `ConcurrentModificationException`, even though the two writes commute.

### Where it stands today

`TX_PAGE_SLOT_MERGE` now replays, on top of the newer committed page:

- inserts into free slots (#5279),
- every update that stays **inside** the page - an overwrite of the same size or smaller, and since #5566 a growth the page can host by shifting the records that follow.

Deletes are excluded wholesale. `LocalBucket.deleteRecordInternal` poisons the page as its very first action, before it has even read the record:

```java
// DISJOINT-SLOT MERGE (#5381): a delete frees a slot and can relink placeholder/chunk records elsewhere, so
// it is never a pure single-slot change - keep the page out of the slot merge.
final TransactionContext slotTx = database.getTransactionIfExists();
if (slotTx != null && slotTx.isSlotMergeEnabled())
  slotTx.poisonSlotRebasePage(fileId, pageId);
```

That is correct but coarse. The reasoning holds for a placeholder pointer (its content record lives on another page) and for a `FIRST_CHUNK` chain, but a **plain in-place record** is deleted by writing `0` into its slot-table entry (plus an optional content wipe-out when `BUCKET_WIPEOUT_ONDELETE` is on) - a single-slot change that commutes with writes to every other slot, exactly like the shapes the merge already covers.

### Proposed change

1. Move the poison from the top of `deleteRecordInternal` to **after** the record's marker shape is known. Keep poisoning for: placeholder pointer, placeholder content, `FIRST_CHUNK` / `NEXT_CHUNK`, and the corrupted-slot branch.
2. For a plain positive size marker, track the delete as rebasable: `trackRebasableDelete(fileId, pageNumber, slot, baseBody)` where `baseBody` is the pre-image, captured before the wipe-out.
3. Add a third kind to `TransactionContext.SlotRebaseBuffer`. A `null` value in `finalBody` is the natural marker for "this slot is deleted" (`HashMap` permits null values), so `rebaseSlots` keeps iterating one map.
4. `LocalBucket.rebaseRecordOnPage` gains a delete mode: verify the committed slot still holds the pre-image **byte for byte** (so a concurrent write to the same record is still a true conflict), then write `0` into the slot-table entry, wipe the content if configured, and update the page statistics.

### Cases to get right

- **Insert-then-delete in the same transaction.** The slot never existed on the committed page, so a pre-image check would fail and turn a perfectly mergeable transaction into a conflict. The replay must recognise the slot as `insertedSlots` + deleted and skip it entirely (the net effect of the transaction on that slot is "nothing").
- **Delete-then-insert on the same slot in one transaction cannot happen**, so it needs no handling: `LocalBucket.getFreeSpaceInPage` deliberately never re-uses a slot freed by the current transaction before it commits (guarded by `Issue5279ConcurrentInsertTest.deletedSlotsAreRecycledByTheNextInsert`). Worth asserting explicitly so a future change to that policy cannot silently break the replay.
- **A delete that frees the last records of a page** must leave `recordCountInPage` consistent with what `compressPage` expects, since the commit path compresses a rebased page right after the replay.
- The record-count delta, `deletedRecordsInTx` and the index changes are all transaction-level and unaffected by which page image is committed, so they need no special handling - worth confirming with a test that a merged delete still removes the index entries.

### Tests

- Concurrent delete of record A + update of record B on one page, `attempts=1`: must not conflict, A gone, B exact.
- Concurrent deletes of different records on one page: must not conflict.
- Delete of a record another transaction is updating: must still raise `ConcurrentModificationException`, and the surviving value must be the winner's.
- Delete of a placeholder / multi-page record co-located with a tracked write: must still fall back cleanly (this is what `Issue5381FalseConflictTest.deleteOnSharedPageFallsBackAndStaysCorrect` asserts today - its rationale will need rewording, since a plain delete would no longer fall back).
- `check database` clean after a contended mix of inserts, updates and deletes on a single-bucket type.

### Why it is worth doing

The workload behind #5279 (many concurrent users on a type with few buckets) deletes as well as inserts and updates, and a delete on a hot page currently poisons it for every co-located writer. This is the last shape where the engine reports a conflict that provably is not one.

lvca:
Closes #5569. Follow-up to #5279 (PR #5566): the last common false-conflict class on a bucket page is gone.

## The problem

Deleting record A made every concurrent transaction that merely **updated record B on the same page** fail with a `ConcurrentModificationException`, even though the two writes commute. `LocalBucket.deleteRecordInternal` poisoned the page as its very first action, before it had even read the record.

That reasoning holds for a placeholder pointer (its content record lives on another page) and for a `FIRST_CHUNK` chain, but a **plain in-place record** is deleted by writing `0` into its slot-table entry (plus the optional `BUCKET_WIPEOUT_ONDELETE` content wipe) - a single-slot change, exactly like the shapes the merge already covers.

## The change

1. The poison moved from the top of `deleteRecordInternal` to **after** the record's marker shape is known. It stays for: placeholder pointer, placeholder content, `FIRST_CHUNK` / `NEXT_CHUNK`, an unknown negative marker, the corrupted-slot branch and the error fallback.
2. A plain positive size marker is tracked with `trackRebasableDelete(fileId, pageNumber, slot, baseBody)`, the pre-image captured **before** the wipe-out.
3. `SlotRebaseBuffer` gains a third kind: a `null` value in `finalBody` marks "this slot is deleted", so `rebaseSlots` keeps iterating one map.
4. `LocalBucket.rebaseRecordDeleteOnPage` verifies the committed slot still holds the pre-image **byte for byte** (a concurrent write to the same record is still a true conflict), then zeroes the slot-table entry, wipes the content if configured and updates the page statistics.

### Cases handled

- **Insert-then-delete in one transaction**: the slot never existed on the committed page, so the replay skips it entirely rather than failing a pre-image check that could not succeed. Covered by `insertAndDeleteOfTheSameRecordInOneTransactionStillMerges`, which forces the churning transaction to commit *last* so it always has to rebase.
- **Delete-then-insert on the same slot** cannot happen (`getFreeSpaceInPage` never re-uses a slot freed by the current transaction, guarded by `Issue5279ConcurrentInsertTest.deletedSlotsAreRecycledByTheNextInsert`). `trackRebasableInsert` now poisons the page defensively if it ever did, instead of silently mis-merging.
- `recordCountInPage` is left alone, as on the normal delete path - it is the slot-table size, and the commit compresses the rebased page right after the replay.
- An **edge-list segment** keeps poisoning the page, so the slot merge and the commutative edge-append merge never claim the same page (mirrors `isSlotMergeCandidate` on the update path).

### One extra fix found on the way

The delete now also poisons the **edge-append** page. That merge re-derives a page from its committed image plus the tracked appends *alone*, so a delete left on a page that also received a tracked append would simply vanish at rebase time. That hole predates this change; freeing a slot is not an append, so the page is excluded whatever the record shape turns out to be.

## Tests

`engine/.../Issue5569ConcurrentDeleteTest` - 8 methods, **6 of which fail without the fix**:

| Test | Asserts |
|---|---|
| `concurrentDeleteAndUpdateOnTheSamePageNeverConflict` | 8 deletes + 8 updates of one page, `attempts=1`: 0 failures, merge counter > 0 |
| `concurrentDeletesOfDifferentRecordsInTheSamePageNeverConflict` | 10 concurrent deletes of one page, `attempts=1`: 0 failures |
| `deletingARecordAnotherTransactionIsUpdatingStillConflicts` | exactly one CME, surviving state is the winner's |
| `insertAndDeleteOfTheSameRecordInOneTransactionStillMerges` | the churner commits last and still merges |
| `insertUpdateAndDeleteOfOneTransactionAreAllReplayedTogether` | all three kinds replayed on one page, order-independent, `check database` clean |
| `aMergedDeleteStillRemovesTheIndexEntries` | index entries gone, unique keys re-usable |
| `deletingAMultiPageRecordCoLocatedWithATrackedUpdateStillFallsBackCleanly` | the non-rebasable shape still falls back, co-located writes exact |
| `contendedMixOfInsertsUpdatesAndDeletesKeepsTheDatabaseConsistent` | 6 threads x 40 rounds, `check database` clean |

`ha-raft/.../Issue5569SlotMergeDeleteRaftIT` - 3-node cluster, survivors hammered with in-place updates while other threads insert and delete on the same page: leader `txPageSlotMerges > 0`, byte-for-byte `assertClusterConsistency()`, and only the survivors with their exact last value on every node.

`Issue5381FalseConflictTest.deleteOnSharedPageFallsBackAndStaysCorrect` keeps passing; its rationale was reworded, since a plain delete no longer falls back.

Full `engine` suite (10336) and `server` suite (777) green, plus `Issue5381SlotMergeRaftIT` and the `Issue5279Remote*IT` pair.

Docs: `docs/release-26.8.1.md`, the `TX_PAGE_SLOT_MERGE` description and `engine/CLAUDE.md` updated.

https://claude.ai/code/session_01X8qjd4CW6B6orurBTsPKQD


lvca:
Thanks for the review - the two interleavings you traced for the edge-append/slot-merge mutual exclusion are exactly the argument the change rests on, good to have them confirmed independently.

On the three minor points, no code change on any of them:

1. **Per-delete `baseBody` allocation** - agreed and accepted, as you note it is unavoidable for the pre-image check and already skipped on a poisoned page. It is the same cost the update path has been paying since #5381, and the alternative (a hash instead of the bytes) was deliberately rejected there: a hash collision on a DBMS write path is a silent lost update.

2. **`@Tag("slow")` on `contendedMixOfInsertsUpdatesAndDeletesKeepsTheDatabaseConsistent`** - measured rather than assumed, and it does not qualify. Per-method timings from the surefire report:

   | method | elapsed |
   |---|---|
   | `deletingARecordAnotherTransactionIsUpdatingStillConflicts` | 0.384 s |
   | `aMergedDeleteStillRemovesTheIndexEntries` | 0.254 s |
   | `concurrentDeleteAndUpdateOnTheSamePageNeverConflict` | 0.218 s |
   | `concurrentDeletesOfDifferentRecordsInTheSamePageNeverConflict` | 0.207 s |
   | **`contendedMixOfInsertsUpdatesAndDeletesKeepsTheDatabaseConsistent`** | **0.317 s** (run in isolation) |
   | the remaining three | < 0.02 s each |

   The whole class runs in ~1.4 s. The 720 transactions are small single-record ones on a single bucket, and `check database` on a type of ~120 records is trivial. `@Tag("slow")` is for multi-second regression tests; tagging this one would drop a `check database` consistency guard out of the fast lane for no gain.

3. **Nested placeholder-content delete poisoning edge-append on the content page** - noted, and it is the intended direction of the asymmetry: the poison can only ever forgo a merge, never drop a write, so erring towards poisoning is the safe default on a path that is already the slow one (a placeholder rewrite).

PR #5599, opened 2026-07-30
fix(engine) #5596: the commit-time page merges prove their coverage instead of relying on every writer to poison

### LINKED ISSUE #5596: Make the edge-append merge prove its coverage instead of relying on every writer to poison
### Summary

`TransactionContext.rebaseEdgeAppends` resolves a commit-time page conflict by **dropping this transaction's whole modified page** and replaying only the tracked in-chunk appends on top of the reloaded committed image:

```java
// Drop the stale copies so the reload observes the current committed version of the page.
modifiedPages.remove(pageId);
immutablePages.remove(pageId);
```

That is sound only if every byte this transaction wrote to that page is accounted for by a tracked append. Nothing verifies it. The invariant is instead maintained by hand, through `poisonEdgeAppendPage` calls scattered across the writers - currently 9 call sites in `LocalDatabase`, `EdgeLinkedList`, `StripedEdgeList` and (since #5569) `LocalBucket`. Miss one, and the merge silently commits a page from which this transaction's other writes have vanished. No exception, no log line: the write is simply gone.

### This has already bitten three times

- **#5385 review round 5**: `writeMultiPageRecord` / `updateMultiPageRecord` wrote `NEXT_CHUNK` records to reused pages through inline record-table writes that bypassed `create/update/deleteRecordInternal`, so they were neither tracked nor poisoned. Fixed by making both writers poison.
- **#5385 review round 7**: a type-3 in-place segment update left no trace in the slot map, and a co-located `StripeDirectory` write made the page look rebasable. Fixed by poisoning on the non-candidate update.
- **#5569 (this week)**: a plain record delete on a page that also received a tracked append would have been dropped. Fixed by adding `poisonEdgeAppendPage(fileId, pageNumber)` to `deleteRecordInternal`.

Three independent instances of one bug class, each found by a human or a reviewer reading the code rather than by a test. The fourth is a matter of time: the merge is opt-out, so a **new** writer is rebasable by default and has to remember to exclude itself.

### Proposed change

Flip it to opt-in: before accepting a rebase, require the page to prove that its modified byte range is fully explained by the tracked appends.

`MutablePage.getModifiedRange()` already gives the `[from, to)` span this transaction dirtied, and the commit path reads it a few lines above the rebase decision. For each tracked segment on the page, the append knows the record's extent (slot-table entry -> record position, plus its on-page size) and the page-header fields it may touch. The union of those extents is the span the merge can legitimately reproduce.

In `isRebasableEdgeAppendPage`, accept the page only when its modified range is **contained** in that union. A delete elsewhere on the page, an untracked inline record-table write, a new chunk, a stripe-directory update - all of them widen the modified range past the appends' extents, so the check refuses and the transaction falls back to a normal retry. Conservative in exactly the right direction: the worst outcome of a false negative is a retry that would have happened anyway before the merge existed.

The existing `poisonEdgeAppendPage` calls stay as the fast, precise path; the containment check is the backstop that makes forgetting one non-fatal.

### The same guard is nearly free on the slot merge

`rebaseSlots` has the identical shape and would benefit from the identical check, comparing the modified range against the union of the tracked slots' extents. It is better covered today - all three hooks live in the central `LocalBucket.create/update/deleteRecordInternal`, so a new writer has to go out of its way to bypass them - but the check costs one comparison on a path that only runs on a conflict, and it would have caught the multi-page-writer gap above without anyone noticing it by reading.

### Tests

- The three historical bugs, as regression tests, with their poison call **removed**: each must now fall back cleanly instead of dropping the write. `Issue5381FalseConflictTest.multiPageRecordCoLocatedWithTrackedInsertStaysIntact` and `stripedDirectoryUpdateCoLocatedWithSegmentAppendFallsBack` are the templates.
- A synthetic writer that dirties a byte outside every tracked extent: the merge must refuse.
- The existing merge-counter assertions must still show the merge firing on the genuinely-clean workloads, so the containment check is not so tight that it refuses everything.

### Why it is worth doing

Every bug in this class is a **silent** lost write on the leader, which then replicates faithfully to every follower - the failure mode with the worst blast radius and the least chance of being noticed. The check converts it into a retry.

lvca:
Closes #5596.

## The problem

Both commit-time page merges - the commutative edge-append merge (`rebaseEdgeAppends`) and the disjoint-slot merge (`rebaseSlots`) - resolve a page-version conflict by **dropping this transaction's whole image of the page** and replaying only its *tracked* writes on top of the newer committed version. That is sound only if every byte the transaction wrote to that page is accounted for by a tracked write. Nothing verified it: the invariant was maintained by hand, through `poisonEdgeAppendPage`/`poisonSlotRebasePage` calls scattered across the writers. Miss one and the merge commits a page from which this transaction's other writes have vanished - no exception, no log line, and the merged bytes then replicate faithfully to every follower.

Three independent instances of that one bug class had already been found by a human reading the code rather than by a test (#5385 review rounds 5 and 7, #5569). A **new** writer is rebasable by default and has to remember to exclude itself.

## The fix: flip it to opt-in

A byte counts as replayable only **while its writer says so**:

```java
final int previous = page.beginCoveredWrite(MutablePage.COVERAGE_SLOT_MERGE);
try {
  ... the writes the merge re-applies ...
} finally {
  page.endCoveredWrite(previous);
}
```

`MutablePage` keeps two ints per page: the mechanisms owning the write in progress, and a sticky set of mechanisms that cannot account for at least one dirtied byte. `updateModifiedRange` - the single funnel every page write already goes through, including `TrackableBinary` and `move` - ORs the undeclared mechanisms into the sticky set. `isRebasableEdgeAppendPage`/`isRebasableSlotPage` then require `page.isFullyCoveredBy(mechanism)`.

`LocalBucket` declares exactly the writes each merge re-applies:

| write | declaration |
|---|---|
| insert into a free slot of a reused page | `COVERAGE_SLOT_MERGE` (what `writeRecordAtSlot` replays) |
| in-page update: overwrite of same size or smaller | `COVERAGE_SLOT_MERGE` |
| in-page update: growth that fits (`growRecordInPage`) | `COVERAGE_SLOT_MERGE` (what `rebaseRecordOnPage` re-does) |
| in-place rewrite of a segment whose append is tracked | `COVERAGE_EDGE_APPEND_MERGE` |
| `compressPage(page, false)` | `COVERAGE_ALL_MERGES` - the commit path re-runs this very call on the rebased page |
| everything else | nothing |

"Everything else" is the point: a record delete, the inline record-table writes of `writeMultiPageRecord`/`updateMultiPageRecord`, a placeholder, a record spilling out of its page, a stripe-directory update co-located with a segment append, a `GraphBatch` bulk load, and any writer added in the future. Each leaves an undeclared byte behind, so the merge declines and the transaction takes the plain retry it would have taken before the merges existed. A forced `compressPage(page, true)` (the database checker) declares nothing either, since the merge path only ever re-runs the non-forced form.

## Why this rather than the byte-range containment check the issue proposed

Containment compares the page's dirtied range against a union of extents **reconstructed analytically** from the slot table. That reconstruction has to model every layout a writer can produce: an in-page growth shifts the whole content tail and rewrites several slot-table entries, and `compressPage` defrags *before* the version check - so the allowed union for exactly the interesting shapes degenerates to "almost the whole page" and stops discriminating. Declaring the extent at the write site is:

- **exact** for growth and for the commit path's own defrag, with no model of the page layout to keep in sync;
- **per-mechanism** - a tracked directory slot write cannot vouch for a co-located untracked segment append, which is precisely the #5385 round-7 shape;
- **cheaper** - two ints per page and one OR per write, versus interval arithmetic on every conflict.

## Side effects

- Closes the still-open **#5569** hole as a byproduct: on `main`, `deleteRecordInternal` poisons the slot merge but says nothing to the edge-append merge, so a delete co-located with a tracked append is silently resurrected. `Issue5596MergeCoverageTest.aDeleteCoLocatedWithATrackedAppendIsNeverMergedAway` reproduces exactly that on `main`.
- Removes the implicit side condition `GraphBatch` relied on ("safe only because a GraphBatch tx never also drives `EdgeLinkedList.add` on the same page"). Its bulk writes are undeclared, so those pages can never be rebased.
- The existing poison calls stay as the fast, precise path; the coverage proof is the backstop that turns a forgotten one into a retry.
- A FINE log fires when coverage is the reason a page is declined - the signal that a writer is missing its poison call.

## Tests

- `MutablePageWriteCoverageTest` (new, 7 tests): the bookkeeping - untouched page, undeclared write, per-mechanism declaration, nesting/restore, stickiness after one forgotten byte, brand-new page never covered.
- `Issue5596MergeCoverageTest` (new, 3 tests):
  - `aDeleteCoLocatedWithATrackedAppendIsNeverMergedAway` - a tracked in-chunk append plus a plain `LocalBucket.deleteRecord` of an orphan chunk on the same segments page, with a competitor bumping that page. **Verified to fail with the check disabled** (the deleted chunk comes back).
  - `anUndeclaredPageWriteIsNeverSlotMergedAway` - the synthetic writer the issue asks for: one tracked in-place update plus bytes dirtied through `getPageToModify` that no tracked slot write accounts for. **Verified to fail with the check disabled** (the poked bytes are lost).
  - `aCleanAppendConflictIsStillMerged` - the guard against over-tightening: the same contention without the undeclared writer must still move the `edgeAppendMerges` counter and keep every edge.
- Full `engine` suite: **10339 tests, 0 failures**. The slow-tagged merge suites (`Issue5381FalseConflictTest`, `EdgeAppendMergeRaceTest`, `EdgeAppendMergeConcurrentStressTest`, `EdgeAppendMergeMultiPageChunkTest`, `ConcurrentEdgeAppendMergeTest`, `Issue5279ConcurrentInsertTest`, `Issue5279ConcurrentUpdateTest`) all pass with their merge-counter assertions intact, so the proof is not so tight that it refuses the workloads the merges were built for.

https://claude.ai/code/session_01X8qjd4CW6B6orurBTsPKQD

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -552,4 +552,31 @@ slower:
   (Bucket counts that are a power of two up to 32 hid this by arithmetic accident: flipping the case of an ASCII
   letter shifts the Java string hash by a multiple of 32.)
 
+## Commit-time page merges now prove their coverage instead of trusting every writer
+
+The two commit-time page merges - the commutative edge-append merge and the disjoint-slot merge - resolve a page
+version conflict by discarding this transaction's whole image of the page and replaying only its *tracked* writes
+on top of the newer committed version. That is sound only when every byte the transaction wrote to that page
+belongs to a tracked write. Until now nothing checked it: the invariant was maintained by hand, through
+`poisonEdgeAppendPage`/`poisonSlotRebasePage` calls scattered across the writers. A writer that forgot one
+committed a page from which its own change had silently vanished - no exception, no log line, and the merged bytes
+then replicated faithfully to every follower. The same gap had already been found and patched three times by
+people reading the code rather than by a test.
+
+The eligibility test is now the other way round: a byte counts as replayable only while its writer says so
+(`MutablePage.beginCoveredWrite`), and a page that carries even one byte written outside such a declaration is
+refused. `LocalBucket` declares exactly the writes each merge re-applies - the single-slot insert, the in-page
+update (overwrite or growth), the plain-record delete, the in-place rewrite behind a tracked edge append, and
+`compressPage`, whose defrag the commit path re-runs on the rebased page. Everything else - the inline
+record-table writes of the multi-page writers, a placeholder, a record spilling out of its page, a
+stripe-directory update co-located with a segment append, a `GraphBatch` bulk load, any writer added in the
+future - leaves the page undeclared and falls back to the ordinary retry it would have taken before the merges
+existed ([#5596](https://github.com/ArcadeData/arcadedb/issues/5596)).
+
+- **No configuration change and no new failure mode.** The existing poison calls stay as the fast, precise path;
+  the coverage proof only ever turns a would-be silent lost write into a retry.
+- **Merge rates are unchanged on the workloads the merges were built for** (concurrent appends to one super-node,
+  concurrent updates, inserts and deletes of unrelated records sharing a page): those pages are fully declared.
+- **Cost is two ints per page and one OR per page write.**
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/supernode.md
+++ b/docs/supernode.md
@@ -553,22 +553,34 @@ written to the WAL and replicated, so followers apply the merged bytes verbatim.
 
 - **Leader / embedded only** (`commit1stPhase(isLeader)`); followers never rebase,
   so no HA divergence.
-- **Only pure-append pages are eligible.** Every non-commutative write **poisons**
-  the page: edge removal/relink (`EdgeLinkedList.updateSegment`), bulk `addAll`,
-  `deleteAll`, and new-chunk allocation (centralised in
-  `LocalDatabase.createRecordNoLock`).
-- **Bulk import (`GraphBatch`) neither tracks nor poisons** - safe only because a
-  `GraphBatch` tx never also drives `EdgeLinkedList.add` on the same page (implicit
-  invariant; any future mixing must poison those pages).
+- **Only pure-append pages are eligible, and the merge has to PROVE it (#5596).**
+  A page is rebasable only when *every* byte this transaction wrote to it was written
+  inside a `MutablePage.beginCoveredWrite(COVERAGE_EDGE_APPEND_MERGE)` declaration -
+  the coverage is opt-**in**, not opt-out. `LocalBucket` declares exactly the in-place
+  segment rewrite that backs a tracked append (plus `compressPage`, whose defrag the
+  commit path re-runs on the rebased page anyway); any other writer touching the page
+  leaves an undeclared byte behind and disqualifies it. A writer that never heard of
+  the merge therefore cannot have its change silently dropped - the worst outcome is
+  the plain retry that predates the merge.
+- **The explicit poison calls stay as the fast, precise path.** Every non-commutative
+  write still poisons the page up front: edge removal/relink
+  (`EdgeLinkedList.updateSegment`), bulk `addAll`, `deleteAll`, and new-chunk
+  allocation (centralised in `LocalDatabase.createRecordNoLock`). The coverage proof is
+  the backstop that turns a forgotten call into a retry instead of a lost write.
+- **Bulk import (`GraphBatch`) neither tracks nor poisons** - and since #5596 it no
+  longer has to: its chunk writes are undeclared, so a page it touched can never be
+  rebased, even if the same transaction also drove `EdgeLinkedList.add` on it.
 - **Allocation-free hot path**: appends keyed by *segment* RID with the pairs packed
   into primitive arrays (`EdgeAppendBuffer`); poisoned pages held as packed-long
-  keys in a `LongHashSet`; `PageId` materialised only on the rare conflict. Measured
-  overhead: **+39 bytes/edge** (~0.7%) vs feature-off, throughput within noise.
+  keys in a `LongHashSet`; `PageId` materialised only on the rare conflict; the
+  coverage proof is two ints per page and one OR per write. Measured overhead:
+  **+39 bytes/edge** (~0.7%) vs feature-off, throughput within noise.
 
 **Touched files:** `GlobalConfiguration` (flag), `TransactionContext` (tracking +
-poison + `rebaseEdgeAppends` + commit-loop hook), `LocalDatabase` (central
-new-chunk poison), `EdgeLinkedList` (register appends; poison remove/bulk paths),
-`PageManager` (`edgeAppendMerges` stat).
+poison + coverage check + `rebaseEdgeAppends` + commit-loop hook), `MutablePage`
+(per-page coverage bookkeeping), `LocalBucket` (declares the replayable writes),
+`LocalDatabase` (central new-chunk poison), `EdgeLinkedList` (register appends;
+poison remove/bulk paths), `PageManager` (`edgeAppendMerges` stat).
 
 ## B.4 The two lost-update fixes (#5147, #5153) (shipped)
 
--- a/engine/src/main/java/com/arcadedb/database/TransactionContext.java
+++ b/engine/src/main/java/com/arcadedb/database/TransactionContext.java
@@ -676,6 +676,10 @@ public class TransactionContext implements Transaction {
    * way that does not commute with a concurrent append (edge removal, new-chunk allocation, bulk load). Tracked
    * appends on that page are ignored at commit (the conflict check below excludes poisoned pages), so a rebase
    * can never silently re-derive the page from committed-state + appends. No-op when the feature is off.
+   * <p>
+   * Since #5596 this is the fast, precise opt-OUT, no longer the only line of defence: the merge additionally
+   * requires the page to PROVE that every byte this transaction wrote there was declared replayable by it (see
+   * {@link MutablePage#beginCoveredWrite(int)}), so a forgotten call here costs a retry, not a lost write.
    */
   public void poisonEdgeAppendPage(final RID segmentRID) {
     if (!edgeAppendMerge)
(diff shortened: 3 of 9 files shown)


lvca:
Thanks - both reviews landed the same three themes, and all three are now implemented. Summary of what changed in 753c6d808:

**Decline has no metric, only an invisible FINE log** (review 2, point 3). Added `PageManager.Stats.mergesDeclinedByCoverage`, incremented at both decline sites. It earned its keep immediately: the first run of the new assertions went red, which turned out to be counter accumulation across tests in one surefire JVM rather than a real decline - so every assertion now compares a delta against a baseline taken just before the contended workload. On genuinely clean workloads the delta is **zero**, which is the positive confirmation that the declarations are complete for the append, insert, growth and delete shapes.

**Positive coverage for the growth and plain-delete shapes was only indirect** (review 2, point 2; review 1, point 2). New `cleanGrowthAndDeleteConflictsAreStillMerged` contends one page with record growths (which shift the following records and rewrite their slot-table offsets) and plain deletes, asserting both that `txPageSlotMerges` moves and that nothing is declined for lack of coverage. `aCleanAppendConflictIsStillMerged` now asserts the same, and the two synthetic writers assert the decline counter **did** move - so they prove the coverage proof is what refused the page, not some unrelated conflict. All of these are in the default (non-slow) build, so an over-tight or forgotten declaration now trips a red test instead of only a log nobody reads.

**`beginCoveredWrite` replaces rather than unions** (review 1, point 1; review 2, point 4). Documented explicitly, with the reason it is not a union: coverage must name the mechanism that actually replays each byte, and a union would let an outer scope vouch for writes it cannot reproduce.

**Thread-visibility of the plain ints** (review 2, point 1). Field comment now states the single-writer-per-page assumption and warns against adding a background mutator of a modified image - an unsynchronised one could hide a dirtied byte from the very proof this PR adds.

**`MutablePage(null, ...)` in the unit test** (review 1, point 3). Kept: a real `PageId` needs a database, which would turn a pure unit test into an integration one. Added a comment on why it is safe today and what to change if the bookkeeping ever dereferences the page id.

Full engine suite green after the follow-ups: 10387 tests, 0 failures. Both synthetic reproductions re-verified as failing with the coverage check disabled.

lvca:
Third round addressed in 84791d6a7 (branch rebased onto current main first - #5592 and #5598 had landed).

**A leaked declaration is the new footgun (point 2).** Implemented, and it is a good catch: this is the one way the backstop could become the hazard it removes. `beginCoveredWrite` now asserts it is not inheriting a live declaration. No writer legitimately opens a non-zero declaration inside another one - each names the single mechanism that replays its own bytes, and a write replayed by several passes the full mask instead - so an inherited live declaration is exactly a missing-`finally` footprint. Assertion rather than a hard check: fast failure under `-ea` (surefire's default) at zero production cost, where the sticky uncovered set still keeps the outcome safe. `aLeakedDeclarationIsCaughtByAnAssertion` pins it and asserts `-ea` is on so it cannot pass vacuously. Writing it also showed that `nestedDeclarationsRestoreTheOuterScope` had been documenting a non-zero-inside-non-zero shape production never performs; it now exercises the real one (an inner scope declaring nothing, as `createRecordInternal` does for a multi-page record). The full engine suite passing with the assertion active is itself the evidence that no path nests today.

**`mergesDeclinedByCoverage` can over- or mis-count (point 3).** Fair, and worth fixing precisely because that stat is documented as *the* signal for a forgotten declaration - an untrustworthy one is worse than none. The two predicates are now pure (`hasReplayableEdgeAppends`/`hasReplayableSlotWrites` split out from the coverage check), and `reportCoverageDecline` counts and logs exactly once, at the `else throw e` site where the transaction is really about to fail, naming which merge declined. No double count, and never a decline recorded for a page the other mechanism went on to merge.

**`@Tag("slow")` (point 1) - not implemented, and here is the measurement.** The review reasoned from thread and transaction counts (noting Maven was unavailable in that environment). Measured on this class:

| method | time |
|---|---|
| `anUndeclaredPageWriteIsNeverSlotMergedAway` | 0.187s |
| `cleanGrowthAndDeleteConflictsAreStillMerged` (the 16-thread one) | **0.044s** |
| `aCleanAppendConflictIsStillMerged` | 0.040s |
| `anUndeclaredPageWriteIsNeverEdgeMergedAway` | 0.010s |
| whole class | **~0.37s**, stable over three runs |

CLAUDE.md's bar is "noticeably long ... multi-second elapsed time"; this is two orders of magnitude below it, and below even a single slow-tagged sibling's setup. The transaction counts are real but each transaction is tiny - the suites it sits next to run thousands, these are sized to the smallest contention that reproduces.

There is also a reason it needs to stay in the regular build, which is the *first* review round's point 2: a merge that stops firing because a writer forgot a declaration is a silent throughput regression that no correctness assertion catches. The merge-counter and coverage-decline assertions only do their job if they run on every build - tagging the class slow would move them out of CI and re-open exactly the blind spot round 1 asked me to close. The class Javadoc now records these numbers and asks for a re-measure rather than a tag by family resemblance.

Full engine suite after the changes: 10411 tests, 0 failures.

lvca:
Fourth round: the one substantive ask is done in 85d755d97, and it turned up a real defect of its own.

**Counting semantics of `mergesDeclinedByCoverage` (point 2).** Documented on both the increment method and the `PPageManagerStats` field: it counts commit **attempts that failed** on a coverage decline, not pages declined. Exactly as you read it - the report is made from the commit loop's terminal branch which then rethrows, so a transaction touching several undeclared pages contributes one, and each failing retry contributes one more. I kept that granularity rather than scanning the remaining pages before throwing: it is the right answer to the question the metric exists for (is a writer missing a declaration, and is it costing retries), and per-page counts would mean extra work on the failure path for no extra signal. The note now says explicitly not to read it as a page count.

**A defect that ask uncovered:** writing the note showed that the javadoc I added in the previous round linked `{@link Stats#edgeAppendMerges}`, but the type is `PPageManagerStats` - a dangling reference that would have failed a javadoc build. Fixed, and `com.arcadedb.engine` + `com.arcadedb.database` now pass `javadoc -Xdoclint:reference` clean. Thanks - that one would have shipped.

**The other three points ask for no change, and I have taken none:**

1. *"The default is now not mergeable"* - that is the intended design, and it is precisely what the new counter plus the clean-merge tests exist to police. A future merge-eligible path that forgets its declaration shows up as declines instead of merges, which is now a red test (`aCleanAppendConflictIsStillMerged`, `cleanGrowthAndDeleteConflictsAreStillMerged` both assert declines stay at zero) rather than a silent regression. That trade - lose merge eligibility by default, never correctness - is the whole point of the flip.
2. *The leak assert is production-disabled* - the documented trade. In production the sticky uncovered set already makes the outcome safe (a retry, never a lost write), so a hard check would buy nothing but a crash; under `-ea`, which is how surefire runs, it fails fast where a developer can act on it. Agreed it is a latent guard for the nesting case; no path does that today, and the full suite passing with the assert active is the evidence.
3. *One OR per page write on every page including LSM/index pages* - as you say, not worth optimizing. Gating it per file type would cost a branch and a lookup to save an OR.

Affected suites re-run green after the change.

lvca:
Thanks for the approve. No code changes this round - all four notes are ones I judge better left as they are, and one factual correction to the review.

**Correction:** the review credits `aDeleteCoLocatedWithATrackedAppendIsNeverMergedAway` with reproducing the #5569 hole against `main`. That test no longer exists. It did in the first revision, but PR #5594 (the #5569 fix) landed on `main` mid-review and added the very `poisonEdgeAppendPage` call the test exploited, so it stopped reproducing and became a test that passed either way. I replaced it with `anUndeclaredPageWriteIsNeverEdgeMergedAway`, a synthetic writer that is poison-independent by construction - the property the review correctly praises elsewhere. The current four methods are `anUndeclaredPageWriteIsNeverEdgeMergedAway`, `anUndeclaredPageWriteIsNeverSlotMergedAway`, `aCleanAppendConflictIsStillMerged`, `cleanGrowthAndDeleteConflictsAreStillMerged`. The per-mechanism separation the note describes is still exactly right; only the test name is stale.

**The deferred-update ordering point is the best observation in this round.** `EdgeLinkedList.add` calls `updateRecord` before `trackEdgeAppend`, and the declaration in `updateRecordInternal` only sees the tracked append because that write is deferred to the commit flush. Making bucket writes eager would silently stop declaring those pages. Worth stating that the new counter is what catches it: it would show as declines rising and `edgeAppendMerges` falling, and `aCleanAppendConflictIsStillMerged` asserts declines stay at zero on exactly that shape, so it fails in the default build rather than degrading quietly.

**Notes not acted on:**

1. *Assertion message shows "on page null" for unit-test pages.* Left as is. `pageId` is non-null everywhere the assertion can fire in anger; in a unit test the developer already knows which page they built, and special-casing null would cost a branch in the message to make the production message no better.
2. *Coverage OR runs on index pages too.* As the note says, fine - gating it per file type would cost a branch and a lookup to save an `OR`.
3. *`reportCoverageDecline` recomputes `hasReplayable*`.* Deliberate: it runs only on the path that is about to throw, and it is what buys exactly-once counting. That precision is the whole reason the metric is trustworthy.
4. *Thread-safety of the plain ints.* No change; the "do not add a background mutator" warning is the guard-rail, as noted.

State: rebased onto current `main` (#5592, #5594, #5598 all landed during review), full engine suite green at 10411 tests / 0 failures on the last code change, and both synthetic reproductions re-verified as failing with the coverage check disabled.

Issue #5478, opened 2026-07-28
GraphBatch: Slow data loading if there is a GEOSPATIAL index


danieljuhl:
It actually looks like something is not optimal for graph batch loading if there is a GEOSPATIAL index. It simply gets slower and slower, and eventually it stops.

This is the log from latest main (as of an hour ago)

```
2026-07-28 08:04:11.699 INFO  [LSMTreeIndex] Full compaction of index 'Address_3_4374623006593118' (series=10 mutablePagesToCompact=10 threadId=142)...
2026-07-28 08:07:26.523 INFO  [LSMTreeIndex] Index 'Address_3_4374623006593118' fully compacted in 194824ms (keys=766098 values=1791469 droppedKeys=0 droppedValues=0 mergedSeries=10 newFile=Address_3_4374677943483752(460) threadId=142)
2026-07-28 08:08:34.281 INFO  [LSMTreeIndex] Full compaction of index 'Address_3_4374623006593118' (series=10 mutablePagesToCompact=10 threadId=142)...
2026-07-28 08:16:40.902 WARNI [RaftGroupCommitter] <arcadedb-1> Replicated entry size 27525872 bytes is approaching the maximum replicated Raft entry size of 33554432 bytes (>82%). Consider reducing the batch size, or raise arcadedb.ha.appendBufferSize (and arcadedb.ha.writeBufferSize with it). Large entries also raise leader heartbeat latency and risk election churn under load.
2026-07-28 08:17:44.543 WARNI [RaftGroupCommitter] <arcadedb-1> Replicated entry size 28614244 bytes is approaching the maximum replicated Raft entry size of 33554432 bytes (>85%). Consider reducing the batch size, or raise arcadedb.ha.appendBufferSize (and arcadedb.ha.writeBufferSize with it). Large entries also raise leader heartbeat latency and risk election churn under load.
2026-07-28 08:17:52.198 INFO  [LSMTreeIndex] Index 'Address_3_4374623006593118' fully compacted in 557916ms (keys=1379142 values=3435473 droppedKeys=0 droppedValues=0 mergedSeries=10 newFile=Address_3_4374940506020503(471) threadId=142)
2026-07-28 08:17:52.400 INFO  [RaftTransactionBroker] Schema change for database 'graph-260720' does not fit one Raft entry (49406759 bytes > 33554432): shipping it as 4 ordered entries
[...]

PR #5604, opened 2026-07-31
#5600 Geospatial misc: STRING stays STRING, GEOSPATIAL precision from SQL, fewer cells for area shapes

### LINKED ISSUE #5600: Geospatial misc: STRING properties deserialized as Shape, precision unreachable from SQL, unpruned cells for polygons
Three unrelated geospatial findings collected while working on #5478 / #5592. They are independent - split them out if any one grows.

---

## 1. A `STRING` property silently reads back as a spatial4j `Shape`

**The most important of the three.**

`BinarySerializer.deserializeValue`, `TYPE_STRING` branch (`engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java:711`), parses any stored string that *starts with* `POINT`, `CIRCLE`, `LINESTRING`, `POLYGON`, `ENVELOPE` or `BUFFER` as WKT and returns a `Shape` instead of the `String` that was written:

```java
case BinaryTypes.TYPE_STRING:
  final String str = content.getString();
  // Backward compatibility: parse WKT geometry strings from old databases
  Object parsedValue = str;
  if (str != null && (str.startsWith("POINT") || str.startsWith("CIRCLE") ||
      str.startsWith("LINESTRING") || str.startsWith("POLYGON") ||
      str.startsWith("ENVELOPE") || str.startsWith("BUFFER"))) {
    ...
```

Repro:

```java
database.getSchema().createDocumentType("T").createProperty("location", Type.STRING);
database.transaction(() -> {
  final MutableDocument doc = database.newDocument("T");
  doc.set("location", "POINT (10.0 45.0)");
  doc.save();
});
// reads back as "Pt(x=10.0,y=45.0)", not "POINT (10.0 45.0)"
database.transaction(() -> database.scanType("T", true, r -> {
  System.out.println(r.asDocument().getString("location"));
  return true;
}));
```

Why this is worth fixing:

- **The property is declared `STRING`.** Round-tripping a declared type through save/load should be identity, and here it is not: `getString()` returns spatial4j's `toString()` form (`Pt(x=..,y=..)`), which is not even valid WKT, so feeding it back into any geo function fails.
- **The trigger is a prefix match on arbitrary user text**, not a schema decision. A `description` or `note` column containing `"POLYGON shaped, see attached"` is transformed too.
- **It fires on every read of every STRING property** in the database - a `startsWith` chain per value - despite the comment scoping it to "old databases".
- It cost real debugging time during #5592: a test asserting on `getString()` failed with `Cannot parse geometry from: Pt(x=8.96,y=57.33)` and the cause was several layers away from the assertion.

The intent (letting a WKT string be usable as a geometry without a dedicated type) is reasonable, but doing it in the deserializer means the *storage layer* changes the declared type of a value. The conversion belongs where a geometry is actually asked for - `GeoUtils.parseGeometry` already accepts a `String` - or behind an explicit property type / flag rather than a prefix sniff.

---

## 2. GEOSPATIAL `precision` cannot be set from SQL

`CreateIndexStatement.executeDDL` forwards `METADATA` for `LSM_VECTOR` (line 200), `FULL_TEXT` (223) and `LSM_SPARSE_VECTOR` (232), then falls through to a bare `builder.create()` (243) for everything else. `GEOSPATIAL` is in that fall-through, so:

```sql
CREATE INDEX ON Location (coords) GEOSPATIAL METADATA {"precision": 6}
```

silently ignores the metadata and the index is built at `GeoIndexMetadata.DEFAULT_PRECISION` (11). Only the Java API can set it:

```java
final GeoIndexMetadata meta = new GeoIndexMetadata("Location", new String[] { "coords" }, -1);
meta.setPrecision(6);
builder.withMetadata(meta);
```

Precision drives cell resolution (11 ≈ 2.4 m, 6 ≈ 1.2 km) and therefore both index size and query selectivity, so a user with a coarse dataset has a real reason to lower it and no SQL way to. The fix mirrors the three branches already there.

Note the same statement should reject an unknown metadata key rather than dropping it - silently ignoring `METADATA` is what makes this invisible.

---

## 3. Minor: non-point shapes index more cells than necessary

The FRONTIER tokenizer (`LSMTreeGeoIndex.extractTokens` -> `forEachCoveringCell`, line 626/677) walks `grid.getTreeCellIterator(shape, detailLevel)` directly. Lucene's `RecursivePrefixTreeStrategy` applies `pruneLeafyBranches` (default **true**) on the indexing path, which collapses a complete set of sibling leaves into their parent, recursively - Lucene documents it as "~20-50% fewer indexed cells, and consequently that much less disk and that much faster indexing".

Points are unaffected (`isGridAlignedShape(point)` skips pruning), so nothing about #5478's actual workload changes. This only helps polygons and other area shapes.

The frontier detection is compatible with the pruned output: `recursiveTraverseAndPrune` still emits a pre-order list, and a pruned parent is followed by a cell at its own level or shallower, so the existing "frontier = the next cell is not deeper" rule identifies it correctly. Exposing the strategy's `protected createCellIteratorToIndex` (a small subclass) would be enough.

Caveat worth weighing: pruning buffers all cells in a list before emitting, which Lucene's own javadoc warns against "for high precision (low distErrPct) shapes". Measure before adopting.

lvca:
Closes #5600. The three findings are independent; each is a separate commit-sized change kept together because they all touch the geospatial path.

## 1. A `STRING` property no longer reads back as a `Shape`

`BinarySerializer.deserializeValue` sniffed the first characters of **every** string it read and, when they matched `POINT` / `CIRCLE` / `LINESTRING` / `POLYGON` / `ENVELOPE` / `BUFFER`, parsed the value as WKT and returned a spatial4j `Shape`. A property declared `STRING` therefore did not round-trip, and the trigger was a prefix match on arbitrary user text rather than a schema decision.

The sniff is gone: a string reads back as the string that was written, and the `startsWith` chain no longer runs on every string read in the database.

Backward compatibility is kept where the conversion actually belongs - at the point a geometry is asked for. `GeoUtils.parseGeometry()`, `GeoUtils.parseJtsGeometry()` and `LSMTreeGeoIndex.toShape()` already accepted WKT text, so a pre-26.2.1 database (shapes stored as WKT under `TYPE_STRING`) keeps working, and so does `geo.point()`, which returns WKT.

Two callers did rely on the sniff and are fixed at the right layer: `SQLMethodIsWithin` and `SQLMethodIntersectsWith` returned `null` unless the value was already a `Shape`. They now parse their operands through `GeoUtils.parseGeometry`, like every `geo.*` function, so `coords.isWithin(...)` works on a WKT column and accepts a WKT literal as its parameter too.

**Better alternative suggested by the issue and not taken here:** an explicit `Type.GEOMETRY` in the schema would be the cleanest long-term answer, but that is a schema-format change with its own migration story - worth a separate issue rather than folding into this one.

## 2. `GEOSPATIAL` `precision` is settable from SQL

```sql
CREATE INDEX ON Location (coords) GEOSPATIAL METADATA {"precision": 6}
```

parsed, ran, and silently built the index at the default precision of 11.

The issue proposed mirroring the three existing `METADATA` branches. **The better fix is one level up:** `withType()` already returns a specialised builder that owns its metadata for `LSM_VECTOR`, `FULL_TEXT` and `LSM_SPARSE_VECTOR`; `GEOSPATIAL` was simply missing from that list, which is *why* its metadata had nowhere to go. It now has `TypeGeoIndexBuilder`, in the same shape as the others, so `precision` and `tokenization` are reachable from SQL and from the Java API (`withPrecision`, `withTokenization`) alike.

Two related sharp edges went with it:

- **An unusable `METADATA` is rejected, not dropped.** An unknown key, a non-numeric or out-of-range precision, an invalid tokenization, or a `METADATA` clause on an index type that has no settings at all now raise `CommandSQLParsingException`. Silently ignoring the clause is what kept this gap invisible. Parsing is unchanged, so the existing parser tests still hold.
- **`withType()` no longer leaves the original builder unconfigured.** It returns a *new* object, so `builder.withType(X); builder.create();` - which the pre-existing `LSMTreeGeoIndexSchemaTest` does - failed with `indexType was not specified` once GEOSPATIAL got a subclass. The type is now recorded on the original builder as well, which also removes the same latent trap for full-text and vector callers.

Note the builder parses the `METADATA` keys itself rather than delegating to `GeoIndexMetadata.fromJSON()`: that method reads a **persisted** definition, where a missing `tokenization` means a pre-26.8.1 index and therefore the `FULL` layout. In a `CREATE INDEX` a missing key just means the user did not ask, so the creation-time default has to stand.

## 3. An area shape indexes fewer cells

The issue asked to measure before adopting Lucene's `pruneLeafyBranches`. Measured, on the frontier cell count:

| shape | saving |
|---|---|
| small square | 57% |
| jagged outline | 74% |
| wide rectangle | 0% |
| linestring | 0% |
| point | 0% (by construction) |

Worth taking. A complete set of sibling frontier cells now collapses into its parent, recursively. A parent covers the union of its children, so the cover can only grow and a match is never lost; the `geo.*` predicate post-filters the superset either way.

**Better alternative than the issue's proposal.** The issue suggested subclassing the strategy to expose `createCellIteratorToIndex`. That implementation materialises the entire decomposition in an `ArrayList` (pre-sized to 4096), which is exactly the caveat the issue flagged from Lucene's own javadoc. This implements the prune as a streaming pass over the existing walk instead: only the frontier tokens on the current root-to-leaf path can still be revoked by an ancestor collapsing, so what is held is bounded by `subCellsSize * detailLevel` regardless of shape size. Points - the hot ingest path of #5478 - take an allocation-free shortcut, since a chain of single-child cells can never hold a complete sibling set.

`LSMTreeGeoIndexCellPruningTest` asserts the streaming output is **identical** to Lucene's buffered one, invoking `createCellIteratorToIndex` by reflection, over 8 shapes x 4 precisions, plus a coverage-never-shrinks property and the one-token-per-point guarantee.

This changes what a **new** index writes for area shapes. Indexes in the `FULL` layout are untouched, and `FRONTIER` has not shipped in any release (#5478 merged into 26.8.1-SNAPSHOT today), so no published database holds the unpruned form.

## Tests

New: `Issue5600StringWktRoundTripTest` (7), `LSMTreeGeoIndexCellPruningTest` (3). Extended: `LSMTreeGeoIndexSchemaTest` (+9), `LSMTreeGeoIndexTokenizationTest` (+2, polygon delete symmetry and a pruned jagged polygon at every query resolution).

Regression: 3182 engine tests in `serializer`, `query.sql`, `schema`, `index.geospatial`, `index.fulltext` and the LSM index suites - 0 failures.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -611,15 +611,20 @@ Java API alike:
 CREATE INDEX ON Location (coords) GEOSPATIAL METADATA {"precision": 6}
 ```
 
-Two related sharp edges went with it:
+A `METADATA` clause the index cannot use is now reported instead of dropped. An unknown key (`{"precisin": 6}`), a
+precision that is not a whole number in 1-12, an invalid tokenization, or a `METADATA` on an index type that has
+no settings at all raise a `CommandSQLParsingException`. Silently ignoring the clause is what kept this gap
+invisible in the first place.
 
-- **An unusable `METADATA` clause is rejected instead of dropped.** An unknown key (`{"precisin": 6}`), an
-  out-of-range precision, an invalid tokenization, or a `METADATA` on an index type that has no settings at all
-  now raise a `CommandSQLParsingException`. Silently ignoring the clause is what kept the gap invisible.
-- **`withType()` no longer leaves the original builder unconfigured.** It returns a specialised subclass for
-  `LSM_VECTOR`, `FULL_TEXT`, `LSM_SPARSE_VECTOR` and now `GEOSPATIAL`, so a caller that ignored the return value
-  (`builder.withType(X); builder.create();`) hit `indexType was not specified`. The type is recorded on the
-  original builder as well.
+> **Breaking change, at execution time.** `CREATE INDEX ... UNIQUE METADATA {"test": 3}` - a `METADATA` clause on
+> a plain `LSM_TREE` or `HASH` index, which has no settings to configure - used to be accepted and ignored, and
+> now fails the statement. The SQL **grammar** is unchanged, so such a statement still parses; only running it
+> is refused. If a schema script carries a `METADATA` that was never doing anything, drop the clause.
+
+One more sharp edge went with it: **`withType()` no longer leaves the original builder unconfigured.** It returns
+a specialised subclass for `LSM_VECTOR`, `FULL_TEXT`, `LSM_SPARSE_VECTOR` and now `GEOSPATIAL`, so a caller that
+ignored the return value (`builder.withType(X); builder.create();`) hit `indexType was not specified`. The type is
+recorded on the original builder as well.
 
 ## GEOSPATIAL: an area shape indexes fewer cells
 
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -552,4 +552,65 @@ slower:
   (Bucket counts that are a power of two up to 32 hid this by arithmetic accident: flipping the case of an ASCII
   letter shifts the Java string hash by a multiple of 32.)
 
+## A `STRING` property no longer reads back as a geometry
+
+Deserializing any string looked at its first characters and, when they were `POINT`, `CIRCLE`, `LINESTRING`,
+`POLYGON`, `ENVELOPE` or `BUFFER`, parsed the value as WKT and handed back a spatial4j `Shape` instead of the
+`String` that had been written ([#5600](https://github.com/ArcadeData/arcadedb/issues/5600)). A property declared
+`STRING` therefore did not round-trip: `getString()` returned spatial4j's `Pt(x=..,y=..)` form, which is not valid
+WKT, so feeding it back into a geo function failed. The trigger was a prefix match on arbitrary user text, not a
+schema decision, so a `description` column holding `"POLYGON shaped, see attached"` went through it too - and the
+prefix chain ran on **every** string read in the database.
+
+The storage layer no longer changes the declared type of a value: a string reads back as the string that was
+written. The conversion happens where a geometry is actually asked for - `GeoUtils.parseGeometry()`,
+`GeoUtils.parseJtsGeometry()` and the geospatial index all accept a WKT string - so a database written before
+26.2.1, when shapes were stored as WKT text, keeps working unchanged; `geo.point()` likewise still returns WKT and
+is still indexable and queryable. Only code that relied on `document.get("wktColumn") instanceof Shape` sees a
+difference, and it now gets the type the schema declares.
+
+## GEOSPATIAL: `precision` is settable from SQL, and an ignored `METADATA` is now an error
+
+`CREATE INDEX ... GEOSPATIAL METADATA {"precision": 6}` parsed, ran, and silently built the index at the default
+precision of 11: `CreateIndexStatement` forwarded `METADATA` for `LSM_VECTOR`, `FULL_TEXT` and `LSM_SPARSE_VECTOR`
+and fell through to a bare `create()` for everything else, so only the Java API could set it. Precision drives cell
+resolution (11 is about 2.4 m, 6 about 1.2 km) and therefore index size and query selectivity.
+
+`GEOSPATIAL` now has a builder of its own, `TypeGeoIndexBuilder`, in the same shape as the full-text and vector
+ones, and `withType(GEOSPATIAL)` returns it - so `precision` and `tokenization` are settable from SQL and from the
+Java API alike:
+
+```sql
+CREATE INDEX ON Location (coords) GEOSPATIAL METADATA {"precision": 6}
+```
+
+Two related sharp edges went with it:
+
+- **An unusable `METADATA` clause is rejected instead of dropped.** An unknown key (`{"precisin": 6}`), an
+  out-of-range precision, an invalid tokenization, or a `METADATA` on an index type that has no settings at all
+  now raise a `CommandSQLParsingException`. Silently ignoring the clause is what kept the gap invisible.
+- **`withType()` no longer leaves the original builder unconfigured.** It returns a specialised subclass for
+  `LSM_VECTOR`, `FULL_TEXT`, `LSM_SPARSE_VECTOR` and now `GEOSPATIAL`, so a caller that ignored the return value
+  (`builder.withType(X); builder.create();`) hit `indexType was not specified`. The type is recorded on the
+  original builder as well.
+
+## GEOSPATIAL: an area shape indexes fewer cells
+
+The FRONTIER layout introduced above stores the cells a shape's decomposition stops at. A complete set of sibling
+cells is now collapsed into its parent, recursively - the reduction Lucene calls `pruneLeafyBranches` and applies
+by default on its own indexing path. A parent covers the union of its children, so the cover can only grow and a
+match is never lost; the `geo.*` predicate post-filters the superset either way. Measured on the frontier cell
+count: 57% fewer for a small square, 74% fewer for a jagged outline, and unchanged for a linestring or a wide
+rectangle. A point decomposes into a chain of single-child cells and is never affected, so the one-entry-per-point
+guarantee stands.
+
+Lucene's own implementation buffers the entire decomposition in a list, which is why its javadoc warns against the
+option "for high precision (low distErrPct) shapes". ArcadeDB's is a streaming walk: only the frontier tokens on
+the current root-to-leaf path can still be revoked by an ancestor collapsing, bounding what is held to
+`subCellsSize * detailLevel` tokens regardless of shape size. The two produce identical token sets, which is
+asserted directly against Lucene in `LSMTreeGeoIndexCellPruningTest`.
+
+This changes what a **new** index writes for area shapes; an index already in the `FULL` layout is untouched, and
+`FRONTIER` has not shipped in any release, so no published database holds the unpruned form.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -645,4 +645,10 @@ asserted directly against Lucene in `LSMTreeGeoIndexCellPruningTest`.
 This changes what a **new** index writes for area shapes; an index already in the `FULL` layout is untouched, and
 `FRONTIER` has not shipped in any release, so no published database holds the unpruned form.
 
+> Running a **nightly 26.8.1-SNAPSHOT** build from between the `FRONTIER` change and this one is the one case that
+> needs attention: a geospatial index written by those builds holds unpruned cells for its area shapes, while this
+> build recomputes the pruned - smaller - set when a record is deleted, which would leave the extra entries behind.
+> `REBUILD INDEX` on such an index rewrites it in the current layout. Point-only indexes are unaffected, since
+> pruning never applies to them.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/engine/src/main/java/com/arcadedb/function/sql/geo/GeoUtils.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/geo/GeoUtils.java
@@ -72,14 +72,14 @@ public class GeoUtils {
       return shape;
     // Cypher point() returns a Map with x/y or longitude/latitude keys
     if (value instanceof Map<?, ?> map) {
-      double x;
-      double y;
+      final double x;
+      final double y;
       if (map.containsKey("x") && map.containsKey("y")) {
-        x = ((Number) map.get("x")).doubleValue();
-        y = ((Number) map.get("y")).doubleValue();
+        x = coordinate(map.get("x"), "x");
+        y = coordinate(map.get("y"), "y");
       } else if (map.containsKey("longitude") && map.containsKey("latitude")) {
-        x = ((Number) map.get("longitude")).doubleValue();
-        y = ((Number) map.get("latitude")).doubleValue();
+        x = coordinate(map.get("longitude"), "longitude");
+        y = coordinate(map.get("latitude"), "latitude");
       } else {
         throw new IllegalArgumentException("Cannot parse geometry from map: missing x/y or longitude/latitude keys");
       }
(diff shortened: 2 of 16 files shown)


lvca:
Thanks - addressed the two you flagged, in 0156e7d8c.

**Non-integral precision now rejected.** `{"precision": 6.9}` passed the `instanceof Number` check and was truncated to `6`. A GeoHash precision is a tree *level*, so truncating silently swallows exactly the kind of typo this validation exists to report. It now raises, and `fractionalPrecisionIsRejected` covers it.

**`METADATA` rejection called out as breaking in its own right.** The release note now carries an explicit blockquote for it rather than leaving it as a bullet under the geospatial heading, and names the concrete legacy shape (`CREATE INDEX ... UNIQUE METADATA {"test": 3}`) plus the grammar-vs-execution distinction: such a statement still *parses*, only running it is refused.

**Not taken - the `withMetadata` overload nit.** `TypeGeoIndexBuilder` having both `withMetadata(IndexMetadata)` and `withMetadata(JSONObject)` is the exact shape `TypeFullTextIndexBuilder` has carried since it was written, and `TypeLSMVectorIndexBuilder`/`TypeLSMSparseVectorIndexBuilder` follow it too. A literal `withMetadata(null)` is ambiguous on all four, so renaming only the geospatial one would trade a latent trap nobody has hit for an inconsistency across the builder family. If it is worth closing it should be closed on all of them at once, which is a separate change.

**No action on the `assert depth == level`** - agreed with your read: assertions off means a violation surfaces as a loud `ArrayIndexOutOfBoundsException`, never as silent mis-indexing, and the Lucene-equivalence test walks the traversal across 8 shapes x 4 precisions. Keeping it as documentation of the one-level-at-a-time GeoHash contract.

On the `isWithin`/`intersectsWith` leniency you noted: correct, a non-geometry parameter now returns `null` instead of throwing `ClassCastException`. That is deliberate - it matches what those two methods already did for a non-geometry *receiver*, and a `WHERE` clause that hits one bad row should filter it out rather than fail the query.

lvca:
Thanks - both notes addressed in 581087ff5.

**Duplicate emit could hide behind the set comparison.** Right, and worth closing: the equivalence test compared `TreeSet`s while production collects into an `ArrayList` and issues one `put()` per token, so a duplicate would have been invisible in the test and a real extra index entry in production. `noTokenIsEmittedTwice` now asserts on the raw stream across the same 8 shapes x 4 precisions. It passes - no duplicate exists - but it would now be caught.

**The `assert depth == level` guard.** Digging into this, the concern is narrower than it looks: nothing in the algorithm depends on `depth == level`. A frame's parent is the frame *below it on the stack*, found by pop order rather than by level arithmetic, and the array bound is satisfied by a weaker property that holds unconditionally - the pop loop closes every frame whose level is `>= ` the incoming one, so the levels on the stack are strictly increasing and the depth can never exceed the number of distinct levels, `detailLevel`. A traversal that descended two levels at once would produce `depth < level` and still associate parents correctly and stay in bounds.

So there is no silent-misprune path here, and the assert is documentation of the GeoHash one-base-32-character-per-level contract rather than a guard the code leans on. I have corrected the comments to say exactly that, since the previous wording implied the dependency you (reasonably) read into it.

**`copyCommonFrom()` across the four builders** - agreed, and agreed it is not for this PR. The hand-copied field list is a real footgun: a new `TypeIndexBuilder` field silently fails to propagate through `withType()` for all four subclasses, not just the geospatial one. I would rather fix it for the family in one change than half-fix it here. Filing it separately.

lvca:
Thanks - all three code points addressed in 5b092c1c0.

**2. Silent `null` on a malformed WKT literal.** Good catch, and on reflection the two operands are not the same kind of input at all, which is what makes this fixable rather than a judgment call:

- **The parameter is written in the query.** `coords.isWithin('POLYGONN (...)')` is a typo, and the method already threw for a *missing* parameter - swallowing a malformed one was the inconsistency. It now propagates.
- **The value comes from the record.** A row that happens not to hold a geometry is skipped, because failing an entire query over one bad row is worse than filtering it out.

`aMalformedParameterIsReportedButABadRowIsJustSkipped` pins both halves: a table with one good row and one junk row still returns the good row, and the typo'd literal surfaces.

**3. Duplication.** With the above the block would have grown in both copies, so extracting it was overdue. Both methods now sit on `AbstractSQLGeoRelationMethod` and differ only in the `SpatialRelation` they accept - the parsing contract has exactly one definition, and each concrete class is down to a name, a relation and a syntax string.

**4. `withMetadata(JSONObject)` return type.** Made chainable. You are right that the asymmetry with the `withMetadata(IndexMetadata)` override sitting three lines above it is the odd part; the change is purely additive so no caller moves.

**1. The `METADATA` breaking change.** Noted, no code change - agreed it is the main compatibility risk here. It has its own callout in `release-26.8.1.md` naming the concrete legacy shape and the grammar-vs-execution distinction.

449 tests across the geospatial, `geo.*` function and SQL-method suites pass on the new commit.

lvca:
Both pre-merge checks done, plus the two nits, in 9d02b02f9.

**(1) No-op `METADATA` on a plain index - swept the repo, nothing breaks.** Scanned every text file (not just Java) with a concatenation-aware match, excluding `target/` and `docs/`: 227 `CREATE INDEX … METADATA` occurrences, all `LSM_VECTOR` (179), `FULL_TEXT` (23), `GEOSPATIAL` (10) or `LSM_SPARSE_VECTOR` (7).

The one that needed proving is Studio, the only place that builds the clause by concatenation - `studio-database.js:1485`, `if (metadataJson != null) command += " METADATA " + metadataJson;`. `metadataJson` is initialised to `null` and assigned at exactly one site, inside the `LSM_SPARSE_VECTOR` branch, so Studio can only ever emit `METADATA` on a sparse-vector index. Safe.

The only two plain-type occurrences are `CreateIndexStatementTestParserTest:30` and `:45`, and they are parse-only: `checkSyntax` calls `parse()` + `validate()`, and `validate()` never inspects `metadata` - the new rejection is in `executeDDL`. They stay green, which the run confirms.

Incidental find, pre-existing and not from this PR: Studio's `LSM_VECTOR` branch also never sets `metadataJson`, so Studio emits `CREATE INDEX … LSM_VECTOR` with no `METADATA` - which `executeDDL` has always rejected with *"LSM_VECTOR index requires METADATA"*. Filing separately.

**(2) `ClassCastException` from a point map with non-numeric x/y.** Real hole, fixed - and at the source rather than in the two methods: `GeoUtils.parseGeometry` cast `map.get("x")` straight to `Number`. Now it reports an `IllegalArgumentException` like every other unparseable geometry, so the "one bad row is skipped" contract holds for *every* `geo.*` consumer, not only `isWithin`/`intersectsWith`. Widening the catch to `RuntimeException` would have hidden genuine bugs instead.

**(3) Snapshot-upgrade delete symmetry.** Correct, and worth stating rather than leaving implied. The release note now carries it explicitly: a database written by a `26.8.1-SNAPSHOT` build from between the FRONTIER change and this one holds unpruned area cells that a delete on this build would no longer fully remove; `REBUILD INDEX` rewrites it. Point-only indexes are unaffected, since pruning never applies to them.

**Nit - `BytesRef` per cell.** Taken. All three walks now retarget a single scratch `BytesRef` (`getTokenBytesNoLeaf(reuse)` just repoints `bytes`/`offset`/`length` at the cell's own array - no copy), and the token becomes a `String` immediately so nothing outlives the next call. This is the ingest path #5478 exists to keep cheap, so the per-cell garbage was worth removing.

**Nit - `withGeoType()` redundancy.** Agreed it is redundant with `withType(GEOSPATIAL)`; keeping it for symmetry with `withLSMVectorType()`/`withFullTextType()`, since `CreateIndexStatement` reads uniformly across the four branches.

**Nit - `assert depth == level`.** Agreed, and that is now what the comment says: the strictly-increasing-levels argument is the real guarantee, the assert documents the GeoHash contract.

464 tests green across the geospatial, `geo.*`, SQL-method and Cypher-spatial suites.

lvca:
Thanks. Point 2 was the one with a real question behind it, so I traced it end to end - the answer is **no internal path can produce `CREATE INDEX <plain type> ... METADATA {...}`**, because nothing in ArcadeDB round-trips a schema through generated DDL at all:

- **Schema export/import is structural JSON, not SQL.** `JsonlExporterFormat:98` writes `LocalSchema.toJSON()` verbatim; `JsonlImporterFormat.loadSchema` rebuilds through `createDocumentType`/`createProperty`/`getOrCreateTypeIndex` and, for vectors, `buildTypeIndex(...).withLSMVectorType().withMetadata(...)`. There is no SQL exporter in the codebase - the formats are `jsonl`, `graphml`, `graphson`.
- **Backup/restore is a raw file copy.** `FullBackupFormat` zips `configuration.json`, `schema.json` and the component files; restore unzips them.
- **HA replicates the schema JSON, not DDL.** `RaftReplicatedDatabase` ships `getSchema().getEmbedded().toJSON()` in a `SCHEMA_ENTRY`, and `ArcadeStateMachine.applySchemaEntry` applies it with `schema.update(new JSONObject(...))`. The wire format has no SQL field. Snapshot install copies `schema.json` as a file.
- **`LocalSchema` never touches `CreateIndexStatement`** - it rehydrates indexes from the persisted JSON via `setMetadata` + `addIndexInternal`.
- **`REBUILD INDEX` stays on the builder path** (`buildTypeIndex(...).withMetadata(rebuildMetadata).create()`), so a plain index whose JSON happens to carry metadata is unaffected: the rejection lives only in the SQL statement, and the builder APIs still accept metadata for any type.

The only string-built `CREATE INDEX` statements in main sources are `TextEmbeddingsImporterLSM` (`LSM_VECTOR`, with metadata - handled by its own branch) and `RemoteSchema.createTypeIndex` (never emits `METADATA`). So the only reachable break is a hand-written user statement, which is what the release note covers.

**Point 1 (per-row WKT re-parsing)** - agreed on all counts, including that it is a net win over the per-read sniff it replaces. Noted for the day it shows up in a profile; no change here.

**Point 3 (`withMetadata(IndexMetadata)` accepting null)** - I'd rather leave it. `IndexBuilder.withMetadata` assigns whatever it is given, and `TypeFullTextIndexBuilder` does the same, so a no-op-on-null would make the geospatial builder the only one of the four that quietly ignores an explicit null. It is also not geospatial-specific: `create()` with null metadata dereferences `metadata.typeName` in the base class regardless of subtype. Same reasoning as the overload nit earlier - if this shape is worth changing, it is worth changing for the family, which is #5606's territory.

**Point 4** - thanks, that was exactly the intent of the rewrite.

lvca:
Closing the loop on the verification asked for in the earlier review: `mvn -o -pl engine verify -DexcludedGroups=benchmark` on the engine module at 9d02b02f9 is **green**.

```
[INFO] Tests run: 10432, Failures: 0, Errors: 0, Skipped: 23
[INFO] BUILD SUCCESS
```

1266 test classes, exit 0.

One note on method, since it matters for reading any earlier number I posted: an intermediate run of this same command reported 41 failures, and **those were not real**. 34 were `NoClassDefFoundError` for classes this PR does not touch (`NodeIndexRangeScan`, `NodeByLabelDisjunctionScan`, `CountEdgesReturnStep$GroupKey`, `LongLongHashMap`), caused by my own concurrent `mvn compile`/`mvn test` invocations rewriting `target/classes` underneath the running surefire JVM; the other 2 were `EdgeAppendMergeRaceTest`, the known-flaky race test from #5570. The run above was executed with nothing else touching the worktree, which is the only way this module's suite gives a trustworthy answer.

CI's own `unit-tests`, `slow-unit-tests` and `integration-tests` jobs have also passed on this branch.

The one red check, `Meterian client scan`, fails identically on `main` and this PR adds no dependencies, so it is not attributable here.

Issue #5595, opened 2026-07-30
Partitioned type: a lookup key of a different boxed type than the stored value hashes to the wrong bucket and misses the record


lvca:
On a type using `PartitionedBucketSelectionStrategy`, a lookup whose key is a **different boxed type than the stored property value** is pruned to the wrong bucket and silently misses the record.

Surfaced during the review of #5591 (comment 5135066971), which correctly noted it predates that PR and is not introduced by it. Filing separately; #5591 does not fix it.

### Measured

Type with 8 buckets, `id LONG`, `UNIQUE` index on `id`, strategy `partitioned('id')`. Records inserted with `Long` values, then looked up through `TypeIndex.get` with the numerically equal `Long` and `Integer`:

| stored value | `Long.hashCode` | `Integer.hashCode` | found as Long | found as Integer |
|---|---|---|---|---|
| -5 | 4 | -5 | yes | **no** |
| -1 | 0 | -1 | yes | **no** |
| -12345 | 12344 | -12345 | yes | **no** |
| 5 | 5 | 5 | yes | yes |
| 8589934592 | 2 | n/a (out of int range) | yes | n/a |

Positive values below 2^31 happen to work because `Long.hashCode(v) == Integer.hashCode(v)` there. Every negative value diverges, because `Long.hashCode` is `(int)(v ^ (v >>> 32))` while `Integer.hashCode` is the value itself.

### Root cause

`PartitionedBucketSelectionStrategy` computes the bucket from `value.hashCode()` on both sides, but the two sides see different objects:

- **Placement** (`getBucketIdByRecord`) hashes `record.get(property)`, i.e. the value **after** schema coercion. For a `LONG` property that is a `Long`.
- **Lookup** (`getBucketIdByKeys`) hashes whatever the caller passed.

`TypeIndex.get(keys)` calls `getIndexesByKeys(keys)` with the **raw** keys; the per-index normalisation `convertKeys` runs later, inside `LSMTreeIndex.get` (~line 540), long after the bucket has been chosen. So nothing reconciles the two.

### Why this matters as much as #5589

The commit-time duplicate check reads through the same pruned path (`TransactionIndexContext.checkUniqueIndexKeys` -> `TypeIndex.get`). By the same mechanism as #5589, a `UNIQUE` index on a partitioned type should therefore admit a duplicate when the incoming key's boxed type differs from the stored one. I measured the missed-lookup half above; the duplicate half follows from the identical code path but I have not reproduced it directly.

[...]

Issue #5589, opened 2026-07-30
Partitioned type: lookups on a secondary index are pruned to the wrong bucket, returning nothing and letting UNIQUE duplicates through


lvca:
On a type using `PartitionedBucketSelectionStrategy`, a lookup on any index **other than the partition index** is pruned to the wrong bucket. Two consequences, both silent:

1. the lookup returns nothing;
2. a secondary **UNIQUE** index stops enforcing its constraint, because the commit-time duplicate check reads through the same pruned path.

No exception is raised in either case.

### Root cause

`TypeIndex.getIndexesByKeys` (engine/src/main/java/com/arcadedb/index/TypeIndex.java:567) prunes with:

```java
final int bucketIndex = type.getBucketIndexByKeys(keys, ...);
if (bucketIndex > -1) {
  final List<String> propNames = getPropertyNames();
  List<IndexInternal> polymorphicIndexesOnKeys = type.getPolymorphicBucketIndexByBucketId(
      type.getBuckets(false).get(bucketIndex).getFileId(), propNames);
  ...
  return polymorphicIndexesOnKeys;
}
```

`keys` are the values of **this** index's properties, and `PartitionedBucketSelectionStrategy.getBucketIdByKeys` hashes whatever it is handed without checking that those properties are the partition properties. But a record is placed by `getBucketIdByRecord`, which hashes the **partition** properties. For any index other than the partition one the two hashes are unrelated, so the pruned search reads a bucket the record is not in.

`LocalDocumentType.getPolymorphicBucketIndexByBucketId` (:880) does filter by property name, but only after the wrong bucket has already been chosen, so it cannot catch this.

The query planner guards its own bucket-pruning rule by comparing the partition properties against the ones in the predicate: `SelectExecutionPlanner:2017` and `PartitionPruning:63` both read `partitioned.getProperties()` and bail when it does not line up. The index path has no equivalent guard.

Round-robin and thread strategies are unaffected: `ThreadBucketSelectionStrategy.getBucketIdByKeys` returns `-1`, so pruning never fires.

### Reproducer

Added as `engine/src/test/java/com/arcadedb/partitioning/PartitionedSecondaryIndexLookupTest.java`. Schema: 8 buckets, `UNIQUE` on `tenant_id` (the partition key) and `UNIQUE` on `code`, strategy `partitioned('tenant_id')`.

```sql
CREATE PROPERTY PartitionedSecondary.tenant_id STRING;
[...]

PR #5605, opened 2026-07-31
fix(engine) #5603: refuse a partition key whose bucket is not a function of the index key

### LINKED ISSUE #5603: Partitioned bucket strategy: follow-ups from #5589 / #5595
Tracking issue for the loose ends left after [#5589](https://github.com/ArcadeData/arcadedb/issues/5589) (PR #5591) and [#5595](https://github.com/ArcadeData/arcadedb/issues/5595) (PR #5598). Both fixed the *read* side of `PartitionedBucketSelectionStrategy` - a lookup that could not reach the bucket placement chose now declines to prune and fans out. What remains is that nothing warns you when you configure a type into one of those permanently-fanning-out states, plus two narrower gaps.

Ordered by value.

---

- [ ] **1. `ALTER TYPE ... BucketSelectionStrategy partitioned(...)` should refuse or warn on an unsuitable type**

  The common root behind both #5589 and #5595: the strategy attaches silently and the breakage only appears at read time, as missing rows or an unexplained slowdown. `setType` today checks only that a unique automatic index exists on the partition properties. It should also reject or warn when:

  - the type carries indexes on properties **other** than the partition key - lookups on those fan out (#5589), and before that fix they read the wrong bucket entirely;
  - the partition index declares **`COLLATE CI`** - case folding is an index-level normalisation placement never applied, so `'Hello'` and `'hello'` are one index key living in two buckets. Since #5595 such a partition is never pruned, so the user silently gets a full fan-out on every query instead of an error;
  - the partition property is **not declared in the schema** - the record keeps whatever Java type the writer used, so there is no conversion target and pruning is likewise permanently off.

  There is also no repair path for databases already affected by #5589: a secondary `UNIQUE` index may already hold duplicates admitted while the check was reading the wrong bucket. Whatever we do here should come with a note recommending a duplicate scan plus `REBUILD INDEX`.

  Worth deciding explicitly whether these are hard refusals or warnings - refusing at `setType` would break reload of an existing database that is already in one of these states, so a warning at assignment time plus a refusal for *new* assignments is probably the right split.

- [ ] **2. `getBucketIdByRecord` hashes an undeclared property's raw value, so placement itself is incoherent**

  Placement hashes whatever Java type the writer happened to use. Two writers storing `5` and `5L` into the same **undeclared** partition property therefore land in different buckets, because `Integer.hashCode` and `Long.hashCode` diverge - the same mechanism as #5595, but on the write side, where no lookup-side normalisation can help.

  Reachable through the OpenCypher `CREATE INDEX` path, which deliberately keeps properties undeclared (`defaultKeyTypesForUndeclaredProperties` in `TypeIndexBuilder`) so writes preserve their original Java type.

  Low severity today because #5595 already disables pruning for undeclared partition properties, so lookups stay correct - this is a latent trap rather than an active bug. Fixing it properly means normalising placement, which changes the placement function and would need a repartition, so item 1 (refuse the configuration) may be the better answer.

- [ ] **3. Verify partitioned placement against a *reloaded* record, not just a freshly built one**

  Needs investigation before it is worth fixing - I did not confirm this reproduces.

  `getBucketIdByRecord` also runs during rebuild/repartition, against a record deserialized from disk rather than one built through `MutableDocument.set()`. Every write path coerces via `convertValueToSchemaType`, but if the binary deserializer's Java type for a `DATE`/`DATETIME` property differs from the configured `dateTimeImplementation` that `set()` coerced to, the two hash differently and a repartition would relocate records that were correctly placed.

  Repro to try: a partitioned type with a `DATETIME` partition key under a non-default `dateTimeImplementation` (e.g. `Instant`), then `REBUILD TYPE ... WITH repartition = true` and check whether any record moves bucket.

---

### Note for anyone already running `partitioned(...)`

The `COLLATE CI` change in #5595 trades speed for correctness: those partitions now fan out across every bucket on every lookup. Anyone running a case-insensitive partition index at scale will see a slowdown after upgrading. Covered in the release notes, but worth flagging directly to known users of the strategy.

lvca:
Closes #5603.

Follow-up to #5589 (PR #5591) and #5595 (PR #5598), which fixed the *read* side of `PartitionedBucketSelectionStrategy`. What was left is that nothing warned you when a type was configured into a state where declining to prune is the only possible answer.

## What measuring it turned up

A live correctness bug none of the three follow-up items named.

On a partitioned type a `UNIQUE` index is a set of **per-bucket** sub-indexes, so the constraint is global only while every record carrying one index key lands in one bucket. Placement hashes the value; the index decides two keys are equal a different way. Where those disagree, the same key lives in several buckets and **each of them accepts its own copy**.

| partition key | partitioned type | round-robin control |
|---|---|---|
| `BINARY`, 6 writes of the same `byte[]` | **3 admitted** | 1 |
| `DECIMAL`, `1.1` / `1.10` / `1.100` / `1.1000` | **3 admitted** | 1 |
| `DATETIME` under `ZonedDateTime`, one instant in 6 zones | **3 admitted** | 1 |

- **`BINARY`** - the stored form is a `byte[]`, which inherits identity `hashCode`, so every single write of the same bytes drew a fresh bucket.
- **`DECIMAL`** - `BigDecimal.hashCode` folds in the scale, so `1.1` and `1.10` hash apart while the index compares them equal.
- **`DATE`/`DATETIME` under a zone-carrying `dateTimeImplementation`** (`ZonedDateTime`, `Calendar`) - only the instant reaches disk, so the writer's zone is hashed at placement and gone on the way back. The zone-free implementations (`java.util.Date`, `Instant`, `LocalDate`, `LocalDateTime`) round-trip unchanged and stay prunable.

## The fix

Rather than the three symptom checks item 1 proposed, the check is derived from the invariant the whole strategy rests on: **the bucket must be a function of the index key.** `hashAgreesWithIndexKeyIdentity` is an *allow*-list over `Type`, so a type added later is merely slower until someone confirms it belongs, instead of silently placing one key in several buckets.

- Such a partition is **never pruned**, which restores the `UNIQUE` constraint by making the check fan out - the same remedy #5595 applied to `COLLATE CI`.
- `ALTER TYPE ... BucketSelectionStrategy partitioned(...)` **refuses** the configuration outright. `COLLATE CI` is refused on the same footing.
- **An existing database still opens.** The same check runs on schema reload but only logs a warning: a refusal at load would turn a slow database into an unopenable one. Once open, its `UNIQUE` index is enforced again.
- **Placement is untouched**, so no existing database needs a repartition.

Beyond the ask:

- **A second index on non-partition properties is a warning, not a refusal.** Those lookups fan out (#5589), which is correct but no faster than not partitioning; the schema is otherwise reasonable.
- **`ALTER TYPE` says what actually went wrong.** Every failure used to be rewritten as `implementation '...' was not found`, so a `partitioned('x')` rejected for want of a unique index on `x` sent you hunting for a typo in a name that was perfectly valid. A genuinely unknown implementation still reports that it cannot be found. (Only the message changed. The exception stays a `CommandParsingException` subtype: that is what the HTTP layer keys on for 400, and a `CommandExecutionException` would be answered 500 - telling clients and load balancers to retry a request that can only ever fail the same way. Pinned end-to-end by `PartitionedStrategyRefusalHttpTest`.)
- **A refused assignment rolls back.** The strategy is stored on the type before it is validated - it has to be, so it can read back the type it is binding to - so a refusal that did not restore the previous one left the type running the strategy the DDL had just rejected. That leak predates this PR; `setType`'s own unique-index refusal hit it too.

## Two items that are not bugs

- **Item 2, an undeclared partition property.** Not reachable: the strategy demands a unique automatic index, and an index cannot be created on a property that does not exist (`createTypeIndex` refuses; the OpenCypher `CREATE INDEX` path creates a *non-unique* index, which `setType` then rejects). `DROP PROPERTY` refuses while the index stands. Pinned by a test rather than fixed; the strategy keeps a defensive blocker for it.
- **Item 3, a `DATETIME` key shifting bucket across a rebuild.** Does not reproduce as described: `MutableDocument.convertValueToSchemaType` already truncates to the declared precision on the way in, so an `Instant` carrying nanoseconds on a millisecond `DATETIME` property hashes identically before and after a disk round-trip. The drift is the **zone**, not the precision, which is why it is in the blocker list above instead. A round-trip test pins the invariant across 10 key types.

## Alternatives considered and rejected

- **Make the broken types work by normalising placement** - `Arrays.hashCode` for `BINARY`, `stripTrailingZeros` for `DECIMAL`, UTC for zoned temporals. Correct by construction, but it changes placement, so a database that is currently *consistent* (every value written at one scale, or from one zone) would find its records in buckets the new hash no longer points at - trading a slowdown for a fresh correctness break. `REBUILD TYPE ... WITH repartition = true` is refused on graph types, so there is no clean migration. Same reasoning that rejected `serializeKeyForHashing` in #5595. `BINARY` is the one case where no existing database can be consistent, so changing its placement would be safe - but a one-type exception buys an exotic key type at the cost of a uniform rule.
- **Warning instead of refusal for a new assignment.** The issue itself proposed the split that is implemented here, and a blocker means the strategy can never prune: accepting it is accepting the placement constraints for nothing.
- **Warning when an index is created or recollated on an already-partitioned type**, which is how the refusal can still be circumvented (and what `PartitionedBoxedKeyLookupTest` now does deliberately). Not implemented: `addIndexInternal` runs once per bucket, so the natural hook would log N times per index, and the DDL-level hook is a wider change than this issue asks for. The state is caught on the next reopen, and the runtime decline keeps it correct meanwhile. Worth a follow-up if you want it closed.

## Follow-up found while addressing review (pre-existing, not fixed here)

`ALTER TYPE ... BucketSelectionStrategy` **does not persist**. The strategy is set in memory and `schema.json` never
gains a `bucketSelectionStrategy` entry unless some later schema mutation happens to flush it, so a type partitioned
by that DDL alone comes back as `round-robin` after a restart. Reproduced on the unmodified sources (stashing this
PR's three main-source files changes nothing), so it predates this change and is out of scope here - flagged for a
separate issue. It is why the reopen test below saves the schema explicitly.

## Verification

- New `engine/src/test/java/com/arcadedb/partitioning/PartitionedStrategySuitabilityTest.java`, 18 tests. **8 go red** with the two guards neutered; the round-robin controls and the "a suitable partition still prunes to the bucket placement chose" test stay green, so the suite can tell the fix apart from a regression to an unconditional `-1`.
- `PartitionedBoxedKeyLookupTest.aCaseInsensitivePartitionIndexStillFindsEverySpelling` now switches the collation on *after* the strategy is attached, since asking for the combination outright is refused. That is worth keeping: it proves the runtime decline holds on its own.
- Full engine suite: **10417 tests, 0 failures, 0 errors**. Whole reactor compiles.

## Files

- `engine/src/main/java/com/arcadedb/database/bucketselectionstrategy/PartitionedBucketSelectionStrategy.java`
- `engine/src/main/java/com/arcadedb/schema/LocalDocumentType.java`
- `engine/src/main/java/com/arcadedb/query/sql/parser/AlterTypeStatement.java`
- `engine/src/test/java/com/arcadedb/partitioning/PartitionedStrategySuitabilityTest.java` (new)
- `engine/src/test/java/com/arcadedb/partitioning/PartitionedBoxedKeyLookupTest.java`
- `docs/release-26.8.1.md`

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -552,4 +552,47 @@ slower:
   (Bucket counts that are a power of two up to 32 hid this by arithmetic accident: flipping the case of an ASCII
   letter shifts the Java string hash by a multiple of 32.)
 
+## Partitioned types: an unusable partition key is refused instead of quietly breaking `UNIQUE`
+
+Third and last of the series, and the one that closes the hole rather than the symptom
+([#5603](https://github.com/ArcadeData/arcadedb/issues/5603)). On a partitioned type a `UNIQUE` index is a set of
+per-bucket sub-indexes, so the constraint is global only while every record carrying one index key lands in one
+bucket. Placement hashes the value; the index decides two keys are equal a different way. Where those disagree the
+same key lives in several buckets and **each of them accepts its own copy**. Measured against a round-robin control
+that rejects the duplicate every time, a partitioned type admitted 3 of 4 rescaled `DECIMAL` values, 3 of 6
+identical `BINARY` values and 3 of 6 spellings of one instant:
+
+- **`BINARY`** - the stored form is a `byte[]`, which inherits identity `hashCode`, so every single write of the
+  same bytes drew a fresh bucket;
+- **`DECIMAL`** - `BigDecimal.hashCode` folds in the scale, so `1.1` and `1.10` hash apart while the index compares
+  them equal;
+- **`DATE`/`DATETIME` under a zone-carrying `dateTimeImplementation`** (`ZonedDateTime`, `Calendar`) - only the
+  instant reaches disk, so the writer's zone is hashed at placement and gone on the way back. The zone-free
+  implementations (`java.util.Date`, `Instant`, `LocalDate`, `LocalDateTime`) round-trip unchanged and are
+  unaffected.
+
+Such a partition is now never pruned, which restores the constraint by making the uniqueness check fan out, and
+**`ALTER TYPE ... BucketSelectionStrategy partitioned(...)` refuses the configuration outright** rather than
+attaching it and letting the damage surface at read time - the common root behind #5589 and #5595. `COLLATE CI`,
+already unprunable since #5595, is refused on the same footing. Placement is untouched, so no existing database
+needs a repartition.
+
+- **An existing database in one of these states still opens.** The same check runs when the schema is read back,
+  but only logs a warning: a refusal at load would turn a slow database into an unopenable one. Once open, its
+  `UNIQUE` index is enforced again. Duplicates already on disk are not retro-validated - scan those indexes and
+  `REBUILD INDEX` them.
+- **A second index on non-partition properties is a warning, not a refusal.** Those lookups fan out (#5589), which
+  is correct but no faster than not partitioning; the schema is otherwise perfectly reasonable, so it is reported
+  and accepted.
+- **`ALTER TYPE ... BucketSelectionStrategy` now says what actually went wrong.** Every failure used to be
+  rewritten as `implementation '...' was not found`, so a `partitioned('x')` rejected for want of a unique index on
+  `x` sent you hunting for a typo in a name that was perfectly valid. A genuinely unknown implementation still
+  reports that it cannot be found.
+
+Two states the issue asked about turned out not to be reachable, and are pinned by tests rather than fixed: an
+**undeclared partition property** cannot occur, because the strategy demands a unique automatic index and an index
+cannot be created on a property that does not exist; and a **`DATETIME` key under a non-default
+`dateTimeImplementation` does not shift bucket across a rebuild**, because the write path already truncates to the
+declared precision, so a freshly built record and one deserialized from disk hash identically.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -567,9 +567,10 @@ identical `BINARY` values and 3 of 6 spellings of one instant:
 - **`DECIMAL`** - `BigDecimal.hashCode` folds in the scale, so `1.1` and `1.10` hash apart while the index compares
   them equal;
 - **`DATE`/`DATETIME` under a zone-carrying `dateTimeImplementation`** (`ZonedDateTime`, `Calendar`) - only the
-  instant reaches disk, so the writer's zone is hashed at placement and gone on the way back. The zone-free
-  implementations (`java.util.Date`, `Instant`, `LocalDate`, `LocalDateTime`) round-trip unchanged and are
-  unaffected.
+  instant reaches disk, so the writer's zone is hashed at placement and gone on the way back. This covers every
+  datetime precision (`DATETIME_SECOND`, `DATETIME_MICROS`, `DATETIME_NANOS`), which the deserializer reads back
+  through the configured implementation just like the base type. The zone-free implementations (`java.util.Date`,
+  `Instant`, `LocalDate`, `LocalDateTime`) round-trip unchanged and are unaffected.
 
 Such a partition is now never pruned, which restores the constraint by making the uniqueness check fan out, and
 **`ALTER TYPE ... BucketSelectionStrategy partitioned(...)` refuses the configuration outright** rather than
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -588,7 +588,8 @@ needs a repartition.
 - **`ALTER TYPE ... BucketSelectionStrategy` now says what actually went wrong.** Every failure used to be
   rewritten as `implementation '...' was not found`, so a `partitioned('x')` rejected for want of a unique index on
   `x` sent you hunting for a typo in a name that was perfectly valid. A genuinely unknown implementation still
-  reports that it cannot be found.
+  reports that it cannot be found. Only the message changed: the refusal is still a client error and still answers
+  HTTP 400.
 
 Two states the issue asked about turned out not to be reachable, and are pinned by tests rather than fixed: an
 **undeclared partition property** cannot occur, because the strategy demands a unique automatic index and an index
--- a/engine/src/main/java/com/arcadedb/database/bucketselectionstrategy/PartitionedBucketSelectionStrategy.java
+++ b/engine/src/main/java/com/arcadedb/database/bucketselectionstrategy/PartitionedBucketSelectionStrategy.java
@@ -19,6 +19,7 @@
 package com.arcadedb.database.bucketselectionstrategy;
 
 import com.arcadedb.database.Database;
+import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.Document;
 import com.arcadedb.index.TypeIndex;
 import com.arcadedb.log.LogManager;
@@ -297,10 +298,35 @@ public class PartitionedBucketSelectionStrategy extends RoundRobinBucketSelectio
   private static boolean hashAgreesWithIndexKeyIdentity(final Database database, final Type propertyType) {
     return switch (propertyType) {
       case BOOLEAN, BYTE, SHORT, INTEGER, LONG, FLOAT, DOUBLE, STRING, LINK -> true;
-      case DATE, DATETIME, DATETIME_SECOND, DATETIME_MICROS, DATETIME_NANOS ->
-          isZoneFree(propertyType.getJavaImplementation(database));
+      case DATE, DATETIME, DATETIME_SECOND, DATETIME_MICROS, DATETIME_NANOS -> isZoneFree(readBackClass(database, propertyType));
       default -> false;
     };
+  }
+
+  /**
+   * The class the binary deserializer will hand a temporal property back as, which is what decides whether the zone a
+   * record was placed under survives a round trip.
+   * <p>
+   * Asked of the serializer directly rather than through {@link Type#getJavaImplementation}, which resolves the
+   * configured implementation for {@code DATE} and {@code DATETIME} only and answers the static default
+   * ({@code LocalDateTime}) for the three precision subtypes. {@code BinarySerializer.deserializeValue} passes the
+   * configured {@code dateTimeImplementation} for all four datetime binary types, so reading the subtypes through that
+   * helper would report every one of them zone-free and admit exactly the configuration this guard exists to catch -
+   * measured: a {@code DATETIME_NANOS} partition key under {@code ZonedDateTime} let 3 of 6 writes of a single instant
+   * into a UNIQUE index.
+   * <p>
+   * {@code getJavaImplementation} is left alone deliberately. Its other caller is the write path's
+   * {@code convertValueToSchemaType}, where the same mismatch is inert - {@code Type.convert} returns an
+   * already-temporal value untouched, so the conversion target is never consulted for one - and changing it would
+   * alter the class a subtype property converts a String or a Long to, a wider behaviour change than this guard needs.
+   */
+  private static Class<?> readBackClass(final Database database, final Type propertyType) {
+    if (!(database instanceof DatabaseInternal internal))
+      return propertyType.getJavaImplementation(database);
+
+    return propertyType == Type.DATE ?
+        internal.getSerializer().getDateImplementation() :
+        internal.getSerializer().getDateTimeImplementation();
   }
 
   private static boolean isZoneFree(final Class<?> implementation) {
--- a/engine/src/main/java/com/arcadedb/database/bucketselectionstrategy/PartitionedBucketSelectionStrategy.java
+++ b/engine/src/main/java/com/arcadedb/database/bucketselectionstrategy/PartitionedBucketSelectionStrategy.java
@@ -31,11 +31,15 @@ import com.arcadedb.schema.Type;
 import com.arcadedb.serializer.json.JSONArray;
 import com.arcadedb.serializer.json.JSONObject;
 
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.logging.Level;
 
@@ -329,9 +333,20 @@ public class PartitionedBucketSelectionStrategy extends RoundRobinBucketSelectio
         internal.getSerializer().getDateTimeImplementation();
   }
 
+  /**
+   * Whether a temporal implementation carries nothing beyond the instant that reaches disk, so that a value hashes
+   * the same before and after a round trip.
+   * <p>
+   * An allow-list, like the {@code Type} switch above and for the same reason: a deny-list of the zone-carrying
+   * classes would silently pass any implementation nobody thought to name - including one configured by a user -
+   * which is exactly the failure this guard exists to prevent. These four are the zone-free half of what
+   * {@code DateUtils.dateTime}/{@code DateUtils.date} can construct; the rest ({@link Calendar},
+   * {@link ZonedDateTime}, {@link OffsetDateTime}) carry a zone that placement hashes and the deserializer cannot
+   * give back.
+   */
   private static boolean isZoneFree(final Class<?> implementation) {
-    return implementation != Calendar.class && implementation != ZonedDateTime.class
-        && implementation != OffsetDateTime.class;
+    return implementation == Date.class || implementation == Instant.class || implementation == LocalDate.class
+        || implementation == LocalDateTime.class;
   }
 
   /**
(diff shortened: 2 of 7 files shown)


lvca:
Thanks - the datetime-subtype gap is real and is now fixed.

**Confirmed exactly as described.** `Type.getJavaImplementation` resolves the configured implementation for `DATE` and `DATETIME` only and returns the static `javaDefaultType` (`LocalDateTime`) for the three precision subtypes, while `BinarySerializer.deserializeValue` passes `dateTimeImplementation` for all four datetime binary types. Measured on `DATETIME_NANOS` under `ZonedDateTime`:

| | partitioned | round-robin control |
|---|---|---|
| 6 writes of one instant, 6 zones | **3 admitted** | 1 |

and the `ALTER TYPE` was accepted outright - no runtime setting had to change afterwards to reach the broken state, which makes it slightly worse than the base-`DATETIME` case the PR already covered.

**Fix**: the read-back class now comes from the serializer directly (`readBackClass`), which is the authority the deserializer itself consults.

**`Type.getJavaImplementation` deliberately left alone.** I checked whether fixing it at the root was the better move and concluded not, for a reason worth recording: its other caller is the write path's `convertValueToSchemaType`, and the same mismatch is **inert** there - `Type.convert(db, <Instant>, LocalDateTime.class, property)` returns the Instant untouched, so for an already-temporal value the conversion target is never consulted. Verified directly: on a `DATETIME_NANOS` partition key under `Instant`, placement and `toStoredForm` both see an `Instant` and `index.get` finds the record, so the #5595 lookup path is not affected by the gap. Changing the helper would only alter the class a subtype property converts a **String or a Long** to - a wider behaviour change than this guard needs, and outside this issue. The divergence is documented at `readBackClass`.

**Tests** - both verified red against the previous resolution, green after:
- `aZoneCarryingDateTimeImplementationIsRefusedAtEveryPrecision` - now covers `DATETIME`, `DATETIME_SECOND`, `DATETIME_MICROS`, `DATETIME_NANOS` rather than the base type alone.
- `aUniqueIndexHoldsOnASubPrecisionDateTimePartitionKey` - the end-to-end constraint proof one precision down.

**Minor points:**
1. Added the note that the rollback restores the reference only - `previous` was bound when it was assigned and nothing unbinds it, so `previous.setType(this)` would be a no-op.
2. Grepped as suggested: no caller catches `CommandSQLParsingException` for this DDL. Every catcher (`SQLAntlrParser`, `StatementCache`, `SQLScriptQueryEngine`, `SQLASTBuilder`, `SQLFunctionEval`) sits in the parse layer, which runs before `executeDDL` - so none of them can see this exception. Matches your finding.
3. Confirmed `checkSuitability()` is cold-path only: called from `setBucketSelectionStrategy` (assign) and schema reload, never from `getBucketIdByKeys`/`getBucketIdByRecord`. The per-query path calls the allow-list check alone, which is an enum switch plus one serializer getter.

Full engine suite after the change: **10430 tests, 0 failures, 0 errors**.

lvca:
Round 2 addressed. Item 1 was a real regression and my PR description was wrong about it - thanks for not taking the claim at face value.

### 1. HTTP status: confirmed, fixed, and now tested

Your trace is exactly right. `AbstractServerHttpHandler` gates on `instanceof CommandParsingException` in **both** arms - the plain one (line 480) and the transaction-wrapped one (line 553) - and `CommandExecutionException` matches neither, so it fell through to the 500 `else`. The directly-thrown `IllegalArgumentException` arm at line 441 is indeed bypassed once the cause is wrapped.

Verified by an actual HTTP round-trip rather than by reading: reverting to the previous commit makes the new test report `expected: 400 but was: 500` on both refusal paths.

**Fix**: only the message ever needed to change, so the exception stays a `CommandParsingException` subtype (`CommandSQLParsingException`) carrying the improved text. That also matches how statement-level validation refusals are classified elsewhere in the package - `RebuildTypeStatement`'s `repartition = true` gate throws the same type for the same kind of reason.

**Test**: new `server/src/test/java/com/arcadedb/server/http/PartitionedStrategyRefusalHttpTest.java`, 3 tests over real HTTP:
- an unsuitable partition key (`BINARY`) → 400, and the body carries the reason;
- the missing-unique-index refusal → 400, body contains `unique automatic index` and not `was not found`;
- a *suitable* key → 200, so the first two cannot pass by refusing everything.

You were right that this path had no server-side test at all - that is precisely why the regression was invisible. The engine-side rollback test now pins the `CommandParsingException` type too, with a comment pointing at the HTTP contract.

### 2. `isZoneFree` deny-list under an allow-list: taken

Valid, and it was the one place the guard could still be fooled. Inverted to an allow-list of the four zone-free classes (`java.util.Date`, `Instant`, `LocalDate`, `LocalDateTime`) - the zone-free half of what `DateUtils.dateTime`/`DateUtils.date` can construct. Anything else, including a user-configured implementation nobody anticipated, now fails safe as not-prunable.

### 3. Hot-path cost: no action, confirmed conscious

Agreed on both counts. The per-key cost is an enum switch plus one field read, and re-resolving per call is not just cheap but *required*: `dateTimeImplementation` is mutable at runtime, and `aUniqueIndexHoldsAfterTheDateTimeImplementationTurnsZoneCarrying` depends on the guard noticing a change made after the strategy was attached. Memoizing it would reintroduce the bug.

### Verification

- Full engine suite: **10430 tests, 0 failures, 0 errors**.
- Full server suite: **780 tests, 0 failures, 0 errors**.
- Release notes corrected - they had repeated the same wrong claim about the status.

lvca:
Round 3 addressed - all three points taken, and point 1 turned up a separate pre-existing bug.

### 1. Unthrottled warning on every reopen - fixed

Agreed, and your framing is the right one: the blocker warning repeating is a useful persistent reminder, the second-index *advisory* on an otherwise-fine schema is noise. I took the second of your two suggestions rather than throttling, because the warning is **assignment-time advice** - it says "this is not the shape you probably meant", which is worth one line when the shape is chosen and nothing at all on every later open of a schema that has not changed. Throttling would only have made a permanent line less frequent.

So: warnings are skipped when `isReadingFromFile()`; blockers still fire on every open, deliberately, since those describe a database still paying for a strategy it cannot use.

New test `theFanOutWarningIsNotRepeatedOnEveryReopen`, verified red with the guard removed.

### 2. Stale PR description - fixed

Updated. Also added a note there about the finding below so it does not get lost.

### 3. `insertAll` catching broad `Exception` - fixed

Fair, and worth more than "low priority": this is the same trap #5589's review caught in its duplicate-admission test, where swallowing everything made the ratio meaningless. Narrowed to `DuplicatedKeyException`, so any unrelated failure now propagates instead of being tallied as a passing uniqueness check - the one direction that reads as success. Matches how `PartitionedSecondaryIndexLookupTest` classifies the same thing.

---

### Separate pre-existing bug found while writing the reopen test (not fixed here)

**`ALTER TYPE ... BucketSelectionStrategy` does not persist.** The strategy is set in memory and `schema.json` never gains a `bucketSelectionStrategy` entry, so a type partitioned by that DDL alone comes back as **`round-robin` after a restart**:

```
PROBE before close=partitioned
PROBE schema.json has bucketSelectionStrategy=false
PROBE after reopen=round-robin
```

I checked whether this PR caused it: stashing the three main-source files reproduces it identically, so it **predates this change** and is out of scope here. Worth its own issue - note that the existing comment in `setBucketSelectionStrategy` already anticipates the fragility ("would only work because the wrapping DDL happens to save the schema afterward; relying on that is brittle"). It is why the new reopen test flushes the schema explicitly, with a comment saying so rather than quietly working around it.

### A note on the verification, since it nearly produced a false report

The first full server run after these changes reported 2 failures in my own new HTTP test. They were **not real**: `mvn -pl server test` resolves `arcadedb-engine` from `~/.m2`, which `mvn -pl engine test` never refreshes, so the server ran against a stale jar - visible in the assertion text, which quoted the pre-PR wording (`was not found`, `Cannot find a unique index on properties`). Re-run after `mvn -pl engine -DskipTests install`:

- Full engine suite: **10431 tests, 0 failures, 0 errors**
- Full server suite: **780 tests, 0 failures, 0 errors** (including the 3 HTTP tests)

lvca:
Round 4 addressed - all three points taken.

### 1. Broad `catch (Exception)` carrying an authoritative 400 - fixed

Good catch, and the framing is right: the message got sharper, so the misclassification got more convincing. Narrowed to the two types a refusal can actually arrive as:

- `IllegalArgumentException` - the strategy's own unique-index check;
- `SchemaException` - the suitability check, and an unresolvable implementation name.

Anything else now propagates and is classified on its merits, so an NPE or a persistence failure is no longer reported to the client as "your request is malformed". `anUnknownStrategyStillReportsThatItCannotBeFound` still passes, since that path throws `SchemaException`.

### 2. Hot-path cost - measured, nothing to report

Two things worth separating here.

First, a correction to the premise: `partitionKeyIsCaseInsensitive()` - the `getPolymorphicIndexByProperties` map lookup - is **not** added by this PR. It arrived with #5595. What this change adds to the per-key path is `hashAgreesWithIndexKeyIdentity`, and for the common `STRING`/`LONG` partition key that is a single static enum switch; the serializer read happens only in the `DATE`/`DATETIME` branch.

Second, the measurement, on `getBucketIdByKeys` with a `STRING` and a `LONG` partition key, 5M iterations, best-of-5 per JVM run:

| | with the check | without |
|---|---|---|
| STRING | 22.9 / 35.7 / 36.1 ns/op | 23.0 ns/op |
| LONG | 24.2 / 35.7 / 30.9 ns/op | 23.6 ns/op |

The three with-check numbers are the **same binary** across three JVM launches. A 13 ns spread on identical code is larger than any effect being looked for, and the with-check range brackets the without-check figure, so the honest answer is that the difference is below this harness's resolution. I did not commit the benchmark - it would be noise in CI, and per CLAUDE.md would need `@Tag("benchmark")` to be excluded anyway.

### 3. Composite partition keys - added

Agreed this was the gap, since `coversPartitionProperties` is the guard the whole fix rests on. Two tests:

- `oneUnsuitableComponentRefusesTheWholeCompositePartition` - `partitioned('a','b')` with `a STRING`, `b DECIMAL` is refused, and the message names `b`. A composite partition is only as prunable as its worst component: placement sums the per-value hashes, and the sum is no more reproducible than any term in it, so the suitable component must not mask the unsuitable one.
- `anIndexOnPartOfACompositePartitionIsWarnedAbout` - an index on a strict subset is a warning, not a refusal, and the runtime agrees with the diagnosis: a partial key gets `-1` while the full composite key still prunes.

### Nits

- Checked the runtime of the reopen/round-trip tests: the whole class is ~2.7s for 22 tests, well under anything needing `@Tag("slow")`.
- Agreed on the persistence finding deserving its own issue.

### Verification

- Full engine suite: **10433 tests, 0 failures, 0 errors**.
- `PartitionedStrategyRefusalHttpTest`: **3/3**.

One note on how that server number was obtained, since it produced two false failures first: `mvn -pl server test` resolves `arcadedb-engine` from `~/.m2`, and a build running in parallel had overwritten that jar - confirmed by unzipping it and finding none of this branch's strings. The reliable form is `mvn -o -pl engine,server test`, which resolves engine from the reactor instead of the repository.

lvca:
Thanks - taking this as the close of the review cycle. Responses to the three points, none of which change the code:

### 1. The non-persistence bug - agreed on the severity, deferring the filing to @lvca

Agreed it is arguably worse than what this PR fixes, and for the reason you give: it costs a **suitable** partition its pruning across a restart, and reverts a legitimately partitioned type to round-robin placement. Restated for the record:

```
PROBE before close=partitioned
PROBE schema.json has bucketSelectionStrategy=false
PROBE after reopen=round-robin
```

Confirmed to predate this change by stashing this branch's three main-source files and reproducing identically. I have not filed it yet - opening a new issue is outside what I was asked to do here, so it is with @lvca to greenlight. It is documented in the PR body and at the one test that has to work around it, so it will not be lost either way. Worth noting the existing comment in `setBucketSelectionStrategy` already anticipated exactly this ("would only work because the wrapping DDL happens to save the schema afterward; relying on that is brittle").

### 2. Temporal hot path - correct, and deliberately not cached

Your reading is right: one extra field read per temporal key component per lookup. That is the intended trade. `dateTimeImplementation` is mutable at runtime, and `aUniqueIndexHoldsAfterTheDateTimeImplementationTurnsZoneCarrying` depends on the guard noticing a change made *after* the strategy was attached - memoizing it would reintroduce the bug for exactly the case the guard exists to catch. Non-temporal keys never reach that branch.

### 3. Log-line length - leaving as is

Agreed it is long, and agreed it is consistent with the surrounding style, which is why I am leaving it. The multi-component composite case is the worst of it, and it is also the case where naming which property is at fault matters most; a terser message would make the operator go read the source to find out which component was the problem.

### Independent confirmation noted

Thanks for verifying the `FLOAT`/`DOUBLE` entries against `BinaryComparator` - I had checked `-0.0`/`+0.0` empirically (2 of 2 admitted on both a partitioned and a round-robin type, i.e. consistently distinct keys) but not read the comparator, so that is a useful independent check of the same conclusion.

### Final state

- Full engine suite: **10433 tests, 0 failures, 0 errors**
- Full server suite: **780 tests, 0 failures, 0 errors**
- `PartitionedStrategyRefusalHttpTest`: 3/3

One CI note for whoever merges: `Meterian client scan` is red, and this PR adds no dependencies of any kind, so it is not from this change.

Issue #5118, opened 2026-07-08
`SEARCH_INDEX(..., 'term~')` may throw NPE after repeated deletes and updates on a FULL_TEXT index


TobiasJoseHermann:
I've found a reproducible server-side `NullPointerException` when running a Lucene fuzzy query through `SEARCH_INDEX` on an existing `FULL_TEXT` index that has accumulated deletes/updates.

The same index works for plain term queries and prefix/wildcard queries. The failure appears specific to the fuzzy-query path (`~`), which scans the underlying full-text index and assumes `IndexCursor.next()` always returns a non-null `Identifiable`.

This was reproduced from the ArcadeDB console, so it does not appear to depend on a Java client/driver.

## Environment

- ArcadeDB server banner: `26.7.2-SNAPSHOT`
- Local source checked against current `main`: `f36230666 fix(#5099): HA bootstrap first-formation survives leadership transfer (#5113)`
- Query executed through `bin/console.sh`
- Index type: `FULL_TEXT`
- Similarity: default full-text configuration

## Reproduction

I can reproduce it with a small synthetic database if the full-text index is first stressed with repeated inserts, deletes, and updates.

The following script creates a generic `Doc` type, adds a full-text index on `body`, performs repeated inserts/deletes/updates, and then runs a Lucene fuzzy query:

```bash
#!/usr/bin/env bash
set -euo pipefail

CONSOLE=/path/to/arcadedb/bin/console.sh
DB=fulltext_fuzzy_stress

"$CONSOLE" -b "drop database remote:localhost/$DB root password" >/dev/null 2>&1 || true

cmds=(
  "create database remote:localhost/$DB root password"
  "create document type Doc"
  "create property Doc.body string"
  "create index on Doc (body) full_text"
)

for i in $(seq 1 400); do
  cmds+=("insert into Doc set body = 'juncos rogelio alpha $i'")
  cmds+=("insert into Doc set body = 'junco alfredo beta $i'")

  if (( i % 2 == 0 )); then
    cmds+=("delete from Doc where body = 'juncos rogelio alpha $i'")
  fi

  if (( i % 3 == 0 )); then
    cmds+=("update Doc set body = 'juncos updated gamma $i' where body = 'junco alfredo beta $i'")
  fi
done

cmds+=("select count(*) as total from Doc where search_index('Doc[body]', 'juncos~') = true")

"$CONSOLE" -b "${cmds[@]}"
```

On my machine, the final query fails with the NPE below.

[...]

Issue #5055, opened 2026-07-06
[tx] In-tx iteration on a non-unique index drops committed RIDs whose key equals an uncommitted entry's key


lvca:
Discovered while writing the regression tests for #4947 (in-tx null ordering).

On a NON-UNIQUE index, when an uncommitted entry has the SAME composite key as a committed one, in-transaction index iteration emits only ONE row for that key - the committed RID disappears from the result until the transaction ends.

Repro sketch (composite non-unique index on (a,b)):
1. Commit a row with key (1,'m').
2. In a new transaction, insert another row with the same key (1,'m') (different RID).
3. Still inside the transaction: `SELECT FROM Doc WHERE a = 1` via the index returns ONE row for (1,'m') instead of two.

Nulls are NOT involved - reproduced with plain string keys. After commit, both rows are visible again, so the loss is in-transaction only (the tx-overlay merge in `LSMTreeIndexCursor` appears to treat 'key present in overlay' as authoritative and shadow the disk cursor's equal-key entries, instead of merging the distinct RIDs).

Distinct from #4947 (comparator/navigation, fixed): key collisions were deliberately excluded from that PR's functional tests to avoid testing this defect instead - see the comment in `TransactionIndexNullOrderingTest`.

_Filed from the 2026-07 engine deep-audit follow-up work._

lvca:
Fixed in `main` (commit 8dfa620e1).

On a non-unique index, if you inserted a row inside a transaction that had the same key as an already-committed row, a query run **before commit** returned only one of the two matching rows. Both showed up correctly again after commit, so the problem was limited to the open transaction.

The index cursor now returns both the committed and the uncommitted row for that key while the transaction is still open. A related case where a descending (`ORDER BY ... DESC`) range query could skip an uncommitted row at the top of the range was fixed at the same time.

Ships in 26.7.2.

Issue #4947, opened 2026-07-04
[lsm] NULL ordering is inconsistent between the tx overlay and on-disk pages


lvca:
**Severity:** MEDIUM

**Location:** `database/TransactionIndexContext.java:115-118` (`ComparableKey.compareTo`: null sorts HIGH) vs `serializer/BinaryComparator.compare:39-45` and `index/lsm/LSMTreeIndexAbstract.compareKey:462-471` (null sorts LOW).

With `NULL_STRATEGY.INDEX` and composite keys containing nulls, `LSMTreeIndexCursor.getClosestEntryInTx` navigates the tx `TreeMap` (ceiling/floor/higher/lower) under nulls-high ordering while the disk merge uses nulls-low. Uncommitted entries with null key components can be skipped or emitted out of order relative to disk entries during in-tx iteration.

**Fix:** align `ComparableKey.compareTo`s null branch with `BinaryComparator` (nulls low).

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed on main (commit b67973388), in two legs:

1. The tx overlay's `ComparableKey` now sorts NULL key components LOW, matching `BinaryComparator` and `LSMTreeIndexAbstract.compareKey` (it sorted them HIGH, so with `NULL_STRATEGY.INDEX` and composite keys containing nulls the cursor's TreeMap navigation disagreed with the disk merge order).
2. A deeper, null-free leg found while testing: `getClosestEntryInTx` navigated the overlay TreeMap with PARTIAL (prefix) keys, which compare equal to EVERY entry sharing the prefix - so ceiling/floor/higher/lower landed on whichever prefix-equal node the tree walk reached first and in-tx range scans silently skipped the run's other uncommitted entries. Navigation now uses biased boundary keys (`NavigationKey`) that sort strictly before/after the whole prefix run.

Both legs red-first verified (`TransactionIndexNullOrderingTest`, 3 tests red on unfixed code).

[...]

PR #5609, opened 2026-07-31
#5601 Index misc: countEntries() counts live entries, geospatial queries stream their candidates

### LINKED ISSUE #5601: Index misc: countEntries() counts tombstones as live entries, geospatial get() materialises the whole candidate set
Two index findings collected while working on #5478 / #5592. Independent of each other - split if either grows.

---

## 1. `countEntries()` counts tombstones as live entries

On any LSM index, deleting records does not bring `countEntries()` back down to the number of live entries: the count settles on a residual that only a full compaction clears. This is **not** geospatial-specific - it reproduces on a plain `NOTUNIQUE` index.

Repro (5 records, one delete per transaction so real tombstones are written - a remove+add inside one transaction is collapsed by `TransactionIndexContext` and never reaches the index):

```java
database.command("sql", "CREATE DOCUMENT TYPE T");
database.command("sql", "CREATE PROPERTY T.name STRING");
database.command("sql", "CREATE INDEX ON T (name) NOTUNIQUE");
// insert Rome, Milan, Naples, Palermo, Turin in one tx

final Index idx = database.getSchema().getIndexByName("T[name]");
// after insert:            5
for (final String city : cities) {
  database.transaction(() -> database.command("sql", "DELETE FROM T WHERE name = '" + city + "'"));
  System.out.println(idx.countEntries());
}
// observed: 4, 3, 2, 1, 1   -- and countType("T") == 0
```

The last value is `1` with **zero** records left. Deleting all 5 in a single `DELETE FROM T` also leaves `1`.

Why it matters:

- `countEntries()` is public API and is surfaced to users through `schema:indexes` and Studio, so a delete-heavy workload reports an index size that is simply wrong until the next full compaction.
- It cannot be used as an oracle in tests. I had to rewrite two tests in #5592 to assert on query results instead, and the reason (tombstone accounting, not a geospatial bug) took a control experiment on a `NOTUNIQUE` index to establish.
- The engine already tracks the information needed to distinguish the two - `LSMTreeIndex.addDeadEntriesSkipped` / the `deadEntriesSkipped` stat exists for exactly this class of problem.

Whichever way it is resolved, the behaviour should be *stated*: either make the count exclude resolved tombstones, or document it as an upper bound and expose the live count separately. Silently returning a number that is neither is the worst of the three.

---

## 2. `LSMTreeGeoIndex.get()` materialises the whole candidate set

`engine/src/main/java/com/arcadedb/index/geospatial/LSMTreeGeoIndex.java:250` carries the TODO:

```java
// TODO: For large regions at high precision (up to 12 levels) this materialises the candidate RID set in memory.
//       A streaming/lazy cursor chaining cells on demand (similar to LSMTreeFullTextIndex) would reduce GC
//       pressure on production datasets with dense or wide-area queries.
```

A query issues one range scan (frontier cell) or exact lookup (internal cell) per covering cell of the search shape and accumulates every RID into a `LinkedHashSet` before returning. For scale: a 0.2-degree box resolves to ~2,100 frontier cells and a 10x10-degree box to ~4,200, each contributing its RIDs to one in-memory set.

This is not a regression - the pre-#5478 path had the same shape - but it became slightly more load-bearing with that change: `IndexInternal.isResultApproximate()` now (correctly) forbids applying a row limit to the candidate set, since truncating before the `geo.*` predicate re-checks it would drop rows that would have survived. So the set is **always** fully materialised; there is no longer any path that stops early.

The fix is the one the TODO names: a lazy cursor that chains cells on demand, as `LSMTreeFullTextIndex` does. The dedup across cells is the part that needs thought - a shape can be reached through several covering cells - so it likely needs a seen-set of RIDs even while streaming, which bounds the win. Worth measuring on a wide-area query before committing to a design.

Raised independently by three review rounds on #5592.

lvca:
Closes #5601. Two independent index findings; both confirmed, both fixed.

## 1. `countEntries()` counted tombstones as live entries

Reproduced exactly as reported: five records, one delete per transaction, the count goes `4, 3, 2, 1, **1**` with `countType("T") == 0`.

**The tombstones are not the cause.** The cursor already skips dead keys, and the work it skips is already accounted in the `deadEntriesSkipped` stat this issue points at. The defect is one line:

```java
for (final IndexCursor it = iterator(true); it.hasNext(); ) {
  it.next();          // <-- return value discarded
  ++total;
}
```

`LSMTreeIndexCursor.hasNext()` is optimistic - it answers on how many underlying page cursors are still live, not on whether any of them still holds a surviving RID - so `next()` legitimately returns `null` once a trailing run of tombstones leaves nothing to emit. Counting the **call** instead of the **value** counted that `null` as an entry. That also explains the part of the report a "tombstones are counted" theory does not: the residual survived a full compaction that had already dropped every tombstone, because the `null` has nothing to do with them.

`countEntries()` now counts values, and closes its cursor - a full walk of every index is the last place that should leak a compacted-series retire guard or a scan stat.

Of the three outcomes the issue lists, this is the first: the number is now exactly the live entries. The contract is stated on `Index.countEntries()` (live entries; "entry" is the index's own unit - one per analyzed token for full-text, one per posting for sparse vectors, one per covering cell for geospatial; constant-time only on `HASH`), and the same note went into `engine/CLAUDE.md`, whose heading previously implied only vector indexes were non-`O(1)`.

Two test comments that documented the old behaviour as a reason to avoid `countEntries()` are updated, and `LSMTreeGeoIndexTokenizationTest.assertDeleteLeavesNothingBehind` now asserts on it - it is a usable oracle again, which was the concrete cost the issue described.

## 2. Geospatial `get()` materialised the whole candidate set

The TODO names one materialisation. There were **three**, stacked, and fixing only the first would not have made a query stream:

1. `LSMTreeGeoIndex.get()` - candidate `LinkedHashSet`, then a parallel `ArrayList<IndexCursorEntry>`
2. `SQLFunctionGeoPredicate.searchFromTarget()` - `ArrayList<Record>` with `getRecord()` **loading every candidate record of every bucket**, full content, before the `geo.*` predicate re-checked the first one
3. `FetchFromIndexedFunctionStep` iterating that list

So a `LIMIT 10` over a wide-area query paid for the entire candidate set, records included.

The whole chain is lazy now:

- **`GeoIndexCursor`** (new) opens one cell scan at a time and closes it as soon as it is drained. It is also honest about `hasNext()`: the lookahead it needs for dedup makes it exact, so `next()` never answers `null` and throws `NoSuchElementException` when exhausted.
- **`GeoCoveringCellWalk`** (new) turns the covering-cell traversal into a pull-style walk. It keeps #5600's `BytesRef` scratch reuse.
- **`SQLFunctionGeoPredicate`** chains the per-bucket cursors the same way. RIDs are unique per bucket, so no cross-bucket dedup is needed.

On the dedup question the issue raises: a seen-set is still required - a polygon or line decomposes into many cells, so one RID is reachable through several of them, and on the legacy `FULL` layout also through a cell and its ancestor. But it now holds only what has been **emitted** rather than a second parallel copy of the candidate set, so a consumer that stops early never pays for the candidates it did not read. Results are unchanged: the index still answers with a superset the predicate re-checks, and a row limit is still never applied to the candidates (`isResultApproximate()`).

**One consequence worth review attention.** Laziness means an abandoned cursor can leave an underlying scan open, and a compacted-series cursor registers with its file so a full compaction defers dropping it. `FetchFromIndexedFunctionStep` now releases on `close()` and `reset()`, mirroring `FetchFromIndexStep.close()`, which carries the same comment for the same reason.

**Correction to the issue text:** the TODO points at `LSMTreeFullTextIndex` as the model for a lazy cursor. It is not one - it materialises into `TempIndexCursor` exactly as the geo index did. The real templates are `LSMTreeIndexCursor` and `MultiIndexCursor`.

## Not done, deliberately

The root wart behind finding 1 is that `LSMTreeIndexCursor` violates `Iterator`: `next()` can return `null` after `hasNext()` returned true. Several call sites already guard for it (`LSMTreeGeoIndex`, `FullTextQueryExecutor`, #5118). Making `hasNext()` exact needs a prefetch, which moves when the transaction overlay is read and changes what `getRecord()` peeks at - on a class carrying the scars of #4944, #4947, #5055. That is its own change with its own test surface, not a drive-by inside this one. Happy to open a follow-up.

## Verification

- `LSMTreeIndexCountEntriesTombstoneTest` (new, 4 tests) - one-by-one delete, bulk delete, cursor/count agreement on a partially deleted index, and delete-then-compact. All four fail on `main` with the reported residual.
- `GeoIndexCursorTest` (new, 9 tests) - drives the cursor through a `CellCursorFactory` over a real GeoHash walk, so laziness is observable: one row opens exactly one cell scan out of 50+; every drained scan is closed; a RID reachable through many cells is emitted once; a `null` from an optimistic scan is skipped; `next()` throws when exhausted; `hasNext()` is idempotent; `close()` releases the open scan.
- `LSMTreeGeoIndexLazyCursorTest` (new, 5 tests) - end-to-end: a partially read cursor leaves nothing broken for the next query, an early `close()` ends it, `LIMIT` returns the asked-for rows, a polygon under many cells appears once, deleted records do not come back.
- Full `engine` suite: **10191 tests, 0 failures**. Rebased on #5600 and re-run: 238 geo/index tests green, including #5600's own `LSMTreeGeoIndexCellPruningTest` and `TypeGeoIndexBuilderTest`.

Rebasing on #5600 left `forEachCoveringCell` and its `CellVisitor` SAM with no callers - `get()` was the last one - so they are deleted and `GeoCoveringCellWalk` is now the single covering-cell walk on the query path. #5600's `forEachPrunedFrontierCell` (write path) is untouched.

https://claude.ai/code/session_01Puky78GhudfgTHPtYSuFPm


lvca:
Answering the direct ask on the abandoned-`ResultSet`-without-`close()` path, since it is a design question rather than a defect.

**Yes, acceptable — because it is not a new hazard, it is the existing one.** `FetchFromIndexStep.close()` (lines 185-200) already carries the same exposure, in almost the same words:

```java
// Release the index cursors even when the scan did not run to exhaustion (e.g. a LIMIT was
// reached or the result set was closed early): compacted-series cursors register with their
// file so a full compaction defers dropping it, and an unclosed cursor would keep the retired
// file alive until the next database restart.
```

So every indexed SQL query in the engine has behaved this way for as long as that step has existed. The geospatial path was the *anomaly* in holding nothing open, and only because it did all its work up front — which is the defect this PR fixes. After this change geo is consistent with every other index scan rather than exceptional.

Three things bound it:

1. **A fully drained query needs no `close()` at all.** Each cell scan is closed when drained, each per-bucket cursor before the next opens, and the dedup set is released when the covering-cell walk ends. The exposure is strictly *partially read AND never closed*.
2. **The in-engine path is wired and now tested.** `ResultSet.close()` → `SelectExecutionPlan.close()` → `lastStep.close()` → `prev.close()` reaches the step, and `FetchFromIndexedFunctionStepCloseTest` asserts the close actually reaches the iterator (verified by mutation: removing the `AutoCloseable` branch fails it).
3. **The contract is now written down** on `IndexableSQLFunction.searchFromTarget`, including the for-each caveat you raised — a plain enhanced-for does not close its iterator.

On the two asymmetries flagged here and in earlier rounds:

- **`FetchFromIndexStep.reset()`** (line 665) nulls `cursor` without closing it, so it has the latent leak the new `reset()` avoids. Confirmed in the code. Left alone deliberately: it is pre-existing, sits on the path every indexed query takes, and changing reset semantics there deserves its own change rather than riding along in a geospatial PR. Happy to open a follow-up.
- **Idempotency after `close()`** was correct but emergent in `GeoCandidateIterator` — it rested on `close()` having pushed `nextBucket` past the last bucket. Pushed 9c811eff9 making it an explicit `closed` flag, mirroring `GeoIndexCursor`.

Issue #4897, opened 2026-07-02
Clients: generate a thin fetch-native TS client from OpenAPI


robfrank:
**Parent:** #4894

## Motivation

This is the priority deliverable and the reason the Epic exists: ArcadeDB has zero official footprint in the browser/edge JS ecosystem today. Serves the "thin for browser/edge" principle — no heavy runtime, tree-shakeable, works in Cloudflare Workers/Deno/Bun/Node without polyfills.

## Proposed design

- Tooling: `openapi-typescript` (types-only codegen from the OpenAPI spec) + `openapi-fetch` (small typed `fetch` wrapper, no runtime dependency beyond `fetch` itself). Both need license verification before adding as a build-time dependency (build-time only, not shipped in the published bundle, since `openapi-typescript` emits pure `.d.ts`).
- Spec source: the live `/api/v1/openapi.json` from a Testcontainers-managed ArcadeDB instance in CI (not a checked-in spec snapshot, to guarantee freshness) — reuse the `e2e-js` Testcontainers pattern.
- Targets: browser, Cloudflare Workers, Deno, Bun, Node — verify no Node-only APIs leak in via `openapi-fetch`'s dependency tree.
- Auth helpers: thin wrapper functions/middleware for HTTP Basic and Bearer/API-token auth (`Authorization: Bearer at-...`), matching the `basicAuth`/`bearerAuth` schemes fixed in the spec-completeness issue.
- Package name: `@arcadedb/client` — **verify ownership of the `@arcadedb` npm org before publishing**; no such scope exists in this repo or npmjs.com today (unverified from repo alone).
- New location: likely a new top-level `clients/js/` or `client-js/` directory (repo currently has no client packages outside `studio/`) — confirm placement doesn't conflict with the Maven multi-module build (this is a pure npm package, no Maven wiring needed, similar to how `studio/` is npm-only within the Maven-root repo).
- e2e test: new test file in `e2e-js/` (alongside the existing Bolt/Postgres wire-protocol tests) exercising the generated client against a Testcontainers ArcadeDB instance over its actual HTTP API — this is new test coverage, not a retrofit of the existing Bolt/Postgres tests.
- Label as auto-generated in the package README, versioned to the server release it was generated from.

## Acceptance criteria

- [ ] Client authenticates via both Basic auth and API-token Bearer auth.
- [ ] Client executes a query and a command against a live server and gets typed results.
[...]

Issue #4896, opened 2026-07-02
Clients: anti-drift contract test for the OpenAPI spec


robfrank:
**Parent:** #4894

## Motivation

Serves "spec completeness is the gate" as an ongoing guarantee, not a one-time fix. Completing the spec closes the gap once; without this issue the spec will silently drift again the next time someone adds a handler in `HttpServer.setupRoutes()` without touching `OpenApiSpecGenerator`. Prefer structural prevention over a test that can be skipped or ignored.

## Proposed design

- Preferred approach: refactor `OpenApiSpecGenerator` to **derive its path list from the handler registry** in `HttpServer.setupRoutes()` (e.g., a shared registration table that both `HttpServer` and `OpenApiSpecGenerator` read from, or an annotation/metadata approach on handler classes) so a new route is structurally impossible to leave undocumented. This is a larger refactor — evaluate feasibility during implementation; the RoutingHandler-based registration in `HttpServer` and the imperative `create*Path()` style in `OpenApiSpecGenerator` are quite different shapes and reconciling them may require introducing a small route-descriptor abstraction.
- Fallback approach (if the structural refactor is too invasive for this issue): extend `OpenApiSpecGenerationIT` with a test that reflects/parses the actual registered routes (via a test-only hook into `HttpServer` or by hitting the running server's route table) and fails CI if any registered route lacks a spec path, or vice versa.
- Either way, must cover plugin-contributed routes (Prometheus, Raft/HA) — decide in the spec-completeness issue whether they're in-scope, then encode that decision in the drift check.
- Files touched: `OpenApiSpecGenerator.java`, `HttpServer.java` (if structural refactor), `OpenApiSpecGenerationIT.java`.

## Acceptance criteria

- [ ] CI fails if a new route is added to `HttpServer.setupRoutes()` (or a plugin's `registerAPI()`) without a corresponding OpenAPI path — verified by a test that intentionally adds/removes a route and checks the test catches it.
- [ ] CI fails on reverse drift (a spec path with no registered handler) — currently zero instances, but the test should catch future occurrences.
- [ ] Decision on plugin-route scope is documented in code comments and reflected in the test.

## Dependencies

Depends on the OpenAPI spec-completeness issue (completing the spec first; the drift test is meaningless against an incomplete spec).

## Effort estimate

[...]

Issue #4894, opened 2026-07-02
Epic: Spec-Driven Client Generation (multi-language)


robfrank:
## Summary

ArcadeDB exposes two independent, complementary API surfaces - an HTTP/OpenAPI surface and a gRPC/proto surface - but has no official generated clients for either. Every language binding today is either hand-written (the Java `grpc-client` module) or nonexistent. This Epic replaces "hand-write a client per language" with "generate clients from the server's own specs."

**Rescoped 2026-08-21.** The original plan generated clients for six-plus languages inside this repository. It now builds one clients repository and one language (TypeScript, covering both HTTP and gRPC), proves the infrastructure, and only then designs the rest. Child issues #4897 through #4903 were closed as superseded by the milestone structure below.

**Design:** `docs/superpowers/specs/2026-08-21-spec-driven-client-generation-4894-design.md`

## Two specs, two personas - do not unify them

- **OpenAPI / HTTP** (`server/src/main/java/com/arcadedb/server/http/handler/openapi/`) - universal transport, works from any language with `fetch`, no proxy needed in a browser. Covers the full breadth of the server: admin, auth, core query/command/transaction data plane, time-series, Grafana, Prometheus, PromQL, MCP, and AI endpoints. This is the "batteries included" surface.
- **gRPC** (`grpc/src/main/proto/arcadedb-server.proto`, 2 services, 23 RPCs) - data-plane and admin-plane only, optimized for throughput: streaming queries, CRUD, bulk/bidirectional ingest, transactions, and a small admin service. HTTP/2-only. This is the "performance tier."

These serve different personas (a dashboard calling `fetch` vs. a high-throughput ingest pipeline) and should not be collapsed into one spec or one client family.

## Topology: two repositories, one contract

`ArcadeData/arcadedb` **owns the contract** and publishes `openapi.json` and `arcadedb-server.proto` as versioned GitHub Release assets. `ArcadeData/arcadedb-clients` **consumes them** and never guesses at the API.

This split is safe because #4896 already asserts, on every server PR, that the OpenAPI spec matches the routes Undertow actually registers. That gate is language-independent and stays here. A separate client repository can therefore go *stale* (generated from an older contract) but not *incorrect* (generated from a lying contract). Staleness is bounded by three things: versioned contract artifacts, scheduled regeneration in the clients repo, and a non-blocking smoke job in this repo's CI (M2).

[...]

PR #5614, opened 2026-07-31
feat(server) #4895: complete the OpenAPI spec to match the registered HTTP route surface

### LINKED ISSUE #4895: Clients: complete the OpenAPI spec to match the registered HTTP route surface
**Parent:** #4894

## Motivation

Serves the "spec completeness is the gate" principle. The spec currently documents 22 of 61 known operations (36%). Every downstream client-generation issue in this Epic produces incomplete, misleading clients until this gap is closed. This is explicitly the highest-priority, first issue in the Epic.

## Proposed design

- Source of truth: diff `HttpServer.setupRoutes()` (`server/src/main/java/com/arcadedb/server/http/HttpServer.java`) against `OpenApiSpecGenerator.createPaths()` (`server/src/main/java/com/arcadedb/server/http/handler/OpenApiSpecGenerator.java`).
- Add path items + request/response component schemas for the following, currently entirely undocumented:
  - `POST /api/v1/batch/{database}`
  - Auth: `POST /api/v1/login`, `POST /api/v1/logout`, `GET /api/v1/sessions`
  - Time-series (3): `POST /api/v1/ts/{database}/write`, `POST /api/v1/ts/{database}/query`, `GET /api/v1/ts/{database}/latest`
  - Grafana (3): `GET /api/v1/ts/{database}/grafana/health`, `GET /api/v1/ts/{database}/grafana/metadata`, `POST /api/v1/ts/{database}/grafana/query`
  - Prometheus remote write/read (2): `POST /api/v1/ts/{database}/prom/write`, `POST /api/v1/ts/{database}/prom/read`
  - PromQL (5): `GET /api/v1/ts/{database}/prom/api/v1/{query,query_range,labels}`, `GET /api/v1/ts/{database}/prom/api/v1/label/{name}/values`, `GET /api/v1/ts/{database}/prom/api/v1/series`
  - MCP (2): `/api/v1/mcp`, `/api/v1/mcp/config`
  - AI (8): `GET /api/v1/ai/config`, `POST /api/v1/ai/activate`, `POST /api/v1/ai/chat`, `POST /api/v1/ai/analyze-profiler`, `GET /api/v1/ai/chats`, `GET/PUT/DELETE /api/v1/ai/chats/{id}`
  - Plugin-contributed (10, currently invisible to the spec generator entirely): `GET /prometheus` (from `PrometheusMetricsPlugin`), and 9 Raft/HA cluster endpoints from `RaftHAPlugin` (`GET /api/v1/cluster`, `POST /api/v1/cluster/peer`, `DELETE` under `/api/v1/cluster/peer/`, `POST /api/v1/cluster/leader`, `POST /api/v1/cluster/stepdown`, `POST /api/v1/cluster/leave`, prefix `/api/v1/cluster/verify/`, prefix `/api/v1/cluster/resync/`, `POST /api/v1/cluster/bootstrap-state`, prefix `/api/v1/ha/snapshot/`). Decide and document whether plugin routes are included conditionally (only when the plugin is active) or always documented as optional.
- Fix the `bearerAuth` bug: attach `bearerAuth` as an alternative security requirement (alongside `basicAuth`) on operations that accept API-token auth, so generated clients can actually authenticate with `at-` tokens.
- Exclude `GET /api/v1/openapi.json` and `GET /api/v1/docs` from self-documentation with an explicit code comment, not a silent omission.
- Files touched: `OpenApiSpecGenerator.java` (add `create*Path()`/`create*Schema()` methods following the existing pattern), `server/src/test/java/com/arcadedb/server/http/OpenApiSpecGenerationIT.java` (extend coverage).

## Acceptance criteria

- [ ] All 51 `HttpServer`-declared operations plus the 10 plugin-contributed operations are present in `/api/v1/openapi.json` (or plugin routes are explicitly, deliberately excluded with a documented rationale).
- [ ] `bearerAuth` is attached to at least one real operation and a test asserts a request authenticated via `Authorization: Bearer at-...` validates against the generated spec's declared security scheme.
- [ ] `OpenApiSpecGenerationIT` passes and asserts on the new paths/schemas.
- [ ] Swagger UI (`GetApiDocsHandler`) renders the new paths without error (manual check).

## Dependencies

None — this is the first issue in the Epic; everything else depends on it.

## Effort estimate

**L** — 61 operations, many needing new request/response schemas (time-series write/query bodies, PromQL response shapes, AI chat payloads); the volume of schema modeling is the main cost, not conceptual difficulty.

robfrank:
Closes #4895. Milestone 0 of #4894.

## What this does

The server generates its own OpenAPI spec at `/api/v1/openapi.json`. It documented **22 of 65** HTTP operations, and declared a `bearerAuth` security scheme that was attached to nothing, so a client generated from it could not authenticate with an API token at all.

This branch takes the spec to **63 documented operations** (2 deliberately excluded), attaches `bearerAuth` at the document root where it belongs, and splits the 811-line `OpenApiSpecGenerator` into nine per-domain `OpenApiContributor` classes.

## Corrections to the issue's own enumeration

The issue's figures were drafted from an earlier audit. Re-verified against source:

- `GET /api/v1/progress/{database}` is registered and was missing from the issue's list.
- MCP is 3 operations, not 2: `MCPConfigHandler` answers both GET and POST.
- Plugin routes are 12, not 10: `/api/v1/ha/snapshot/` is a prefix handler serving both the download and `/checksums`.
- `/ws` needs an explicit exclusion rationale; it is not expressible in OpenAPI 3.0.

## Security

Security moves from a per-operation `basicAuth` requirement to a root-level `basicAuth`-or-`bearerAuth` declaration, which is what `AbstractServerHttpHandler` actually accepts on every authenticated route. That puts `bearerAuth` on 59 of 63 operations. Three overrides:

- `security: []` on `/health` and `/ready`, the only handlers returning `isRequireAuthentication() == false`.
- Basic-only on both `/api/v1/ha/snapshot/**` operations, because `SnapshotHttpHandler` sits outside the standard chain and parses Basic itself.

`X-ArcadeDB-Cluster-Token` is deliberately not declared, with a comment explaining that it is cluster-internal.

## Plugin routes

The 12 Raft and Prometheus routes are declared statically and unconditionally in `PluginApiSpec`. Two constraints force that: client generation runs against a default server loading neither plugin, so a config-conditional spec would emit clients with no cluster methods; and `ha-raft` and `metrics` hold `arcadedb-server` at `provided` scope, so swagger classes are off their compile classpath. **No swagger dependency was added to either module.** The class Javadoc records the reasoning and the shape the anti-drift follow-up (#4896) should take.

## Deliberate exclusions

`GET /api/v1/openapi.json`, `GET /api/v1/docs`, `/ws`, and the `/` Studio fallback, each with an in-code rationale rather than a silent omission.

## Testing

- 106 unit tests across eleven classes, one per contributor plus a shared-builder and a schema-namespace guard.
- `OpenApiSpecGenerationIT` asserts the full 63-operation inventory **bidirectionally**, so it fails on reverse drift as well as omissions, plus `operationId` uniqueness, single-tag-from-a-closed-vocabulary, and full `$ref` resolution.
- Acceptance criterion 2 is proven by a live round trip: mint a real `at-` token, authenticate `GET /api/v1/databases` with it, assert 200, then assert the spec declares the scheme that accepted it.
- Full suite green: `mvn -pl server -am verify`, 10268 engine + 877 server tests.

## Follow-ups for #4894

- **Blocking #4897:** `POST /api/v1/ai/chat` defaults to `mode="auto"`, which returns `text/event-stream`, not the JSON body. OpenAPI 3.0 cannot express a response content type selected by a request-body value, so a generated TS client will bind `application/json` and break on the default path. Mitigated by hoisting the warning into the operation description, the field generators emit as method docs. A real fix needs splitting the operation or changing server content negotiation.
- The 10 Security-domain operations still use bare `type: object` bodies, preserved under a no-drift rule.
- PromQL `data.result` items are opaque and fully specifiable via `oneOf`.

## Review notes

Every schema was verified against its handler rather than against the design doc; that process corrected 35 places where the two disagreed. Existing `operationId`s, summaries, descriptions and tags are byte-identical to before, since client generators derive class and method names from them.


robfrank:
Verified all three actionable points against the handlers. All correct, all fixed in 6c0792cf3.

**1. `checkDatabaseExists` was modelled wrongly.** Confirmed: `GetExistsDatabaseHandler.execute` returns 400 when the `database` parameter is empty, and otherwise **always** 200 with `{"result": <boolean>}`. It never returns 404. The spec's 200/404 split meant a generated client would branch on status and always conclude the database exists, with no typed way to read the actual boolean. This was a genuine missed correction, not a judgement call.

Fixed: added a `DatabaseExists` component schema, pointed 200 at it, dropped the 404, added the 400. The schema description also records something the status code cannot express: the handler filters against `user.getAuthorizedDatabases()`, so `result` is false both when the database does not exist and when it exists but the caller cannot see it. A client needs to know those two cases are indistinguishable.

**2. `checkHealth` / `checkReady` declared a 200 neither handler emits.** Confirmed: `GetHealthHandler` returns only 204; `GetReadyHandler` returns 204 or 503 across three not-ready branches. Both 200s removed.

That fix tripped exactly the trap you would expect. The pre-existing `openApiSpecIncludesProperRequestResponseModels` asserts every GET declares a 200, so it was passing *only because the spec was wrong*. Its POST branch had already been widened earlier in this branch to accept 204, because `POST /api/v1/logout` genuinely returns 204; the GET branch had the identical flaw and simply had not been triggered. Widened it symmetrically rather than special-casing health and ready, and confirmed the trap by reverting the widening while keeping the production fix, which failed on `GET /api/v1/ready` as predicted.

**3. Shared mutable `ApiResponse` across two status keys.** Fixed, and fix 2 dissolves most of it: health now registers a single 204, readiness registers distinct instances for 204 and 503.

All three corrections are pinned by assertions asserting exact status-code sets, so a re-added 404 or a reintroduced 200 fails the build. Each new assertion was mutation-tested to fail before being trusted.

**4 and 5** are tracked as stated: the plugin-route drift surface is #4896, and the `chatWithAi` streaming caveat is #4897, which does block real codegen. Both need filing as follow-ups on #4894.

**Nit** on `CoreApiSpec` mixing raw `new Operation()` with `SpecBuilders.operation(...)`: agreed it reads unevenly, leaving as-is deliberately. Those operations predate this branch and their `operationId`, `summary`, `description` and tag are byte-preserved so downstream generators do not see renamed classes and methods. Worth doing in one pass when that constraint is lifted.

Verification after the fixes: 109 unit tests green (106 + 3 new), `OpenApiSpecGenerationIT` 15/15.

On the sampling caveat: the 35 corrections came from per-domain review of the 41 *newly documented* operations. The 22 pre-existing ones were carried over under a no-drift rule that protected their identifiers, and their response modelling got less scrutiny than the new work. All three findings here sit in that carried-over set, which is a fair characterisation of where the gap was.

PR #5617, opened 2026-07-31
bindings/python: align max_connections default with engine #5352 (supersedes #5355)

robfrank:
Supersedes #5355, preserving @tae898's original commit and authorship. Opened as a separate PR only because `humemai` is an organization-owned fork, and GitHub does not honor maintainer edits on org-owned forks, so the fixes could not be pushed to the original branch.

Follows up #5352. The Python wrapper (`create_vector_index`) carried its own `max_connections=16` default, which shadowed the engine's new default of 32 because the wrapper always calls `withMaxConnections`. This aligns the wrapper and documents the semantics: `maxConnections` is a Vamana per-layer degree, not hnswlib `M`, so reproducing an hnswlib configuration takes `2*M`.

### Original contribution by @tae898

- wrapper default 16 -> 32, docstring documents the 2*M mapping
- example 13 index metadata 16 -> 32
- example 11 CLI default 16 -> 32
- examples 09 and 10 follow the upstream rename of `real_ladybug` to `ladybug`

### Maintainer follow-ups

- **example 11 degree-matching.** One `--max-connections` flag fed six backends with two different meanings. ArcadeDB applies the value verbatim to every layer; faiss, lancedb, pgvector, qdrant and milvus are hnswlib-derived and double it at the base layer. At the new default that compared ArcadeDB at degree 32 against an effective 64 elsewhere. A `hnsw_m_from_max_connections()` helper now converts at each call site, and the four `hnsw_m` entries in the results metadata report the converted value.
- **a regression test that cannot drift.** `bindings/python/tests` had no reference to `max_connections`, which is how the wrapper and engine separated in the first place. The new test reads `LSMVectorIndexMetadata.maxConnections` through JPype and asserts the wrapper default equals it, rather than asserting a literal 32.
- **CI exercises the default.** The example 11 smoke test pinned `--max-connections 16`, so the aligned default never ran end to end.
- **`ladybug` pinned to 0.19.0.** The rename crosses four minors on a 0.x package and no CI job exercises that path.
- **examples 03 and 06** print the beam width they actually build with (100, the engine default), instead of an unset 256.

### Verification

Full Python suite against a wheel built from this branch: **351 passed, 11 skipped**. Six of the skips are `test_docs_examples.py`, which skips because `bindings/python/docs` does not exist in this repository; those run in the contributor's fork, which accounts for the difference from the 353 reported on #5355.

The drift test was proven to fail before being trusted: with the installed default flipped to 16, both new tests fail `assert 16 == 32`; restored to 32, all 13 tests in that file pass.

Note: the docs updates mentioned in #5355's description live in `humemai/arcadedb-embedded-python`. This repository has no `bindings/python/docs` directory, so nothing here covers them.

### DIFF
--- a/.github/workflows/test-python-examples.yml
+++ b/.github/workflows/test-python-examples.yml
@@ -366,7 +366,7 @@ jobs:
                 example_jvm_args=""
                 ;;
               "11_vector_index_build.py")
-                example_args="--backend arcadedb_sql --dataset stackoverflow-tiny --threads 1 --mem-limit 2g --batch-size 500 --max-connections 16 --beam-width 100 --quantization NONE --run-label ci11_arcadedb_sql"
+                example_args="--backend arcadedb_sql --dataset stackoverflow-tiny --threads 1 --mem-limit 2g --batch-size 500 --beam-width 100 --quantization NONE --run-label ci11_arcadedb_sql"
                 example_name="$example (vector build, arcadedb_sql backend, minimal)"
                 timeout_duration=1200
                 example_jvm_args=""
--- a/bindings/python/examples/03_vector_search.py
+++ b/bindings/python/examples/03_vector_search.py
@@ -226,7 +226,7 @@ with arcadedb.create_database(db_path) as db:
     print(
         "      • max_connections: 32 (connections per node, higher = more accurate but slower)"
     )
-    print("      • beam_width: 256 (search quality, higher = more accurate)")
+    print("      • beam_width: 100 (engine default; higher = more accurate)")
     print()
 
     db.command(
--- a/bindings/python/examples/03_vector_search.py
+++ b/bindings/python/examples/03_vector_search.py
@@ -220,15 +220,6 @@ with arcadedb.create_database(db_path) as db:
     print("Step 5: Creating vector index...")
     step_start = time.time()
 
-    print(f"   💡 JVector Parameters:")
-    print(f"      • dimensions: {EMBEDDING_DIM} (matches embedding size)")
-    print("      • distance_function: cosine (best for normalized vectors)")
-    print(
-        "      • max_connections: 32 (connections per node, higher = more accurate but slower)"
-    )
-    print("      • beam_width: 100 (engine default; higher = more accurate)")
-    print()
-
     db.command(
         "sql",
         f"""
@@ -240,6 +231,26 @@ with arcadedb.create_database(db_path) as db:
         }}
         """,
     )
+
+    # Report what the index was actually built with. The METADATA above sets only
+    # dimensions and similarity, so everything else comes from the engine defaults;
+    # reading them back keeps this output correct if those defaults change.
+    index_meta = db.schema.get_vector_index("Article", "embedding").get_metadata()
+    print("   💡 JVector Parameters:")
+    print(f"      • dimensions: {index_meta['dimensions']} (matches embedding size)")
+    print(
+        f"      • distance_function: {index_meta['similarity_function']}"
+        " (best for normalized vectors)"
+    )
+    print(
+        f"      • max_connections: {index_meta['max_connections']}"
+        " (connections per node, higher = more accurate but slower)"
+    )
+    print(
+        f"      • beam_width: {index_meta['beam_width']}"
+        " (search quality, higher = more accurate)"
+    )
+    print()
 
     print("   ✅ Created JVector vector index")
     print("   ✅ Built vector index graph immediately via SQL")
--- a/bindings/python/examples/06_vector_search_recommendations.py
+++ b/bindings/python/examples/06_vector_search_recommendations.py
@@ -250,7 +250,7 @@ def create_sql_vector_index(db, property_suffix=""):
     num_movies = len(result_list)
 
     print(f"\nCreating HNSW (JVector) index for {embedding_prop}...")
-    print("  metric=cosine, max_connections=32, beam_width=256")
+    print("  metric=cosine, max_connections=32, beam_width=100 (engine defaults)")
 
     start_time = time.time()
 
--- a/bindings/python/examples/06_vector_search_recommendations.py
+++ b/bindings/python/examples/06_vector_search_recommendations.py
@@ -250,7 +250,6 @@ def create_sql_vector_index(db, property_suffix=""):
     num_movies = len(result_list)
 
     print(f"\nCreating HNSW (JVector) index for {embedding_prop}...")
-    print("  metric=cosine, max_connections=32, beam_width=100 (engine defaults)")
 
     start_time = time.time()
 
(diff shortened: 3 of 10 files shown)


tae898:
Thanks for picking this up, and for keeping the authorship on the original commit.

The example 11 catch is the one that matters. My change would have shipped a comparison of ArcadeDB at degree 32 against an effective 64 on the hnswlib-derived backends, which is exactly what #5352 was about, one level down. The test reading `maxConnections` off the index metadata rather than asserting a literal 32 is also the right shape, since a literal is how the wrapper and engine drifted apart in the first place.

One scope note, not a blocker: the beam width prints in examples 03 and 06 are unrelated to `max_connections`. Correct as far as I can tell, just flagging in case you would rather keep this PR to the one change and take those separately.

Happy for this to supersede #5355.

robfrank:
Addressed observation 1, took observation 3 as a code comment, and left observation 2 alone. Pushed as 18c00b8bf.

**1. Hardcoded prints in examples 03/06 — fixed rather than left.**

You are right that this is the same drift class the PR exists to fix, and leaving a known instance of it in the same PR would be inconsistent. Both examples now read the values off the created index via the public `db.schema.get_vector_index(type, prop).get_metadata()` and print `dimensions`, `similarity_function`, `max_connections` and `beam_width` from there, so nothing is restated.

Example 03 needed a reorder: the parameter block printed *before* the `CREATE INDEX`, so there was nothing to read back. Moving it after the create also reads better, since it now reports what was built rather than what was intended.

Verification, since these are the two examples with the weakest coverage:
- example 03 is not in the CI example matrix, so I ran it end to end locally. Exit 0, and it prints `dimensions: 384 / distance_function: COSINE / max_connections: 32 / beam_width: 100`, read from the index.
- example 06 needs `sentence-transformers` and a MovieLens database, so I could not run it here. It *is* in CI (`test-python-examples.yml`). Its only material difference from the proven path is a suffixed property name, so I probed exactly that in isolation: create `Movie.embedding_v1` via the same SQL, then `get_vector_index("Movie", "embedding_v1").get_metadata()` returns `COSINE 32 100`. CI covers the rest.

Full suite still 351 passed / 11 skipped.

**2. `hnsw_m_from_max_connections` accepting `str`.** Agreed, no change. `argparse type=int` guarantees an int at every real call site; the string case is only there so the helper is not silently dependent on that guarantee.

**3. Ladybug pin rot.** Added a comment at both pin sites recording why the pin exists (pre-1.0, no CI coverage) and to re-check it when refreshing published results. A comment beats a tracking issue here since it is visible exactly where someone would bump it.

One thing worth separating out, unrelated to this diff: while building a wheel to verify the drift test I hit `bindings/python/scripts/build.sh` returning exit 0 on a failed build twice, once for a missing `build` module and once for six javac errors, leaving a stale wheel in `dist/` that looks like success. That is a real footgun in our own tooling and I would rather file it on its own than fold it in here.

robfrank:
Correcting myself on the build tooling note at the end of my previous comment. I said `scripts/build.sh` "returns exit 0 on a failed build twice, once for a missing `build` module and once for six javac errors." That is wrong. It returned **exit 1** on both runs.

I had run it as a compound command:

```
./scripts/build.sh ... > log 2>&1; echo "EXIT=$?"; tail -15 log
```

so the status I read back was the trailing `tail`'s, not the script's. The one place I wrote a real `EXIT=` marker into the log was the run that succeeded, so the two failing runs carried no marker and I never cross-checked. On top of that I saw a wheel sitting in `dist/` and took it as evidence the build had worked, which made the wrong story fit.

@tae898 is right in 5a8445ea that the javac case fails properly: injecting a javac error does fail the build. Apologies for sending you after a defect that was not there.

What I actually hit was the narrower thing that commit fixes: a failed build leaves the previous wheel in `dist/`, so the presence of a wheel is not evidence this run produced one. In my case that was a 25.10.1 wheel surviving two failed 26.8.1 builds, which is exactly the stale-artifact trap. Checking version and mtime rather than existence is the lesson, and the marker-file comparison in that commit is a better fix than clearing the directory since it preserves wheels built for other platforms.

Two related things I ran into on the same path, in case they are useful and separate from that commit:

- `src/arcadedb_embedded/jars/` is reused when it already exists, so a leftover 25.10.1 jar set made the bridge compile fail with `cannot find symbol: class GraphBatch`. The directory is untracked and not gitignored.
- The wheel I eventually produced contained both `arcadedb-engine-25.10.1` and `26.8.1`. The older sorts first on the classpath, so `Schema` resolved to the old class and every LSM_VECTOR test failed with `INDEX_TYPE has no attribute LSM_VECTOR`, which reads like a missing engine feature rather than shadowing.

None of this affects the diff under review here.

tae898:
Merged, thanks. Reading the metadata back in examples 03/06 is a better fix than the one I had in mind, and reordering 03 so the block reports what was built rather than what was intended is the part I would not have thought to do.

Before you spend time filing the `build.sh` one: I went looking for it, and one half reproduces and one half does not.

**The stale wheel is real, and fixed.** The Docker path's success gate was `ls dist/*.whl`, which asks whether a wheel exists rather than whether this run produced one. Nothing clears `dist/`, so a wheel from any earlier build satisfied it, and an export stage that yields nothing still prints the success banner and exits 0. `build-native.sh` already avoided this with `rm -f "$DIST_DIR"/*.whl` before building; the Docker branch had no equivalent, which is why the two behave differently.

Fixed in humemai/arcadedb-embedded-python@5a8445e, which I will carry into the next bindings PR here rather than opening a separate one. It compares wheel identities across the extraction instead of clearing the directory, so wheels deliberately kept for other platforms survive a build, and a same-named wheel is accepted only when the extraction actually rewrote it (marker file plus `test -nt`, since `find -newermt` is GNU-only and this script runs on macOS hosts too). Four states, verified in isolation and then end to end:

| `dist/` before | this build produced | rc |
|---|---|---|
| empty | nothing | 1 (unchanged) |
| empty | a wheel | 0 |
| stale wheel | nothing | 1 (**was 0**) |
| same-name wheel | rewrote it | 0 |

A full Docker build with a stale wheel planted in `dist/` still exits 0, reports the new wheel, and leaves the planted one alone.

**The exit-0-on-javac-errors half does not reproduce for me.** I put a deliberately broken `.java` into the bridge sources and ran `build.sh linux/amd64 3.12`: javac reports its errors, the RUN fails, and the script exits 1 with no success banner. Same story for the missing `build` module on the native path, where `build-native.sh` pre-flights it and exits 1. I diffed `build.sh`, `build-native.sh` and `Dockerfile.build` against `main` here before starting, and all three are byte-identical to our copies, so it is not fork drift.

So either those two runs failed somewhere I have not found, or the exit 0 you saw was the stale-wheel gate downstream of a build that had already gone wrong. If you still have the scrollback, the last 20 lines and the platform would settle it. If it was the stale wheel, the fix above covers it and there is nothing left to file.

PR #5619, opened 2026-07-31
fix(mcp) #5585: answer malformed JSON-RPC members instead of failing the request

### LINKED ISSUE #5585: MCP: a malformed params or arguments member returns HTTP 500 instead of -32602
### ArcadeDB Version:

`26.8.1-SNAPSHOT` (reproduced at `3cfa4197c`, before PR #5584)

### OS and JDK Version:

macOS 25.5.0, GraalVM CE 25.0.2. Not JDK-specific: the failing call is plain Gson type handling.

### Expected behavior

A JSON-RPC request whose `params` member, or whose `params.arguments` member, is present but of the wrong JSON type is a malformed request. It should be answered with a JSON-RPC error object, `-32602 Invalid params` being the accurate code, and the HTTP transport should still return a well-formed JSON-RPC envelope.

### Actual behavior

The exception escapes `MCPDispatcher.dispatch` entirely and the HTTP transport turns it into **HTTP 500** with no JSON-RPC body at all. The client gets a transport-level failure where it should get a protocol-level error.

```
java.lang.IllegalStateException: Not a JSON Object: ["a"]
java.io.IOException: Server returned HTTP response code: 500 for URL: http://127.0.0.1:2480/api/v1/mcp
```

Two independent sites, both in `server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java`:

1. **`params` itself**, read at `dispatch` before the method `switch`:

```java
final JSONObject params = request.getJSONObject("params", new JSONObject());
```

This sits *above* `dispatch`'s `try`, and that `try` carries only a `finally` (the `DatabaseContext` cleanup) with no `catch`, so nothing downstream can answer for it. This affects **every** method, including `ping`.

2. **`params.arguments` in `toolsCall`**, read before that method's own `try`:

```java
final String toolName = params.getString("name", "");
final JSONObject args = params.getJSONObject("arguments", new JSONObject());
```

The root cause in both cases is that the defaulting accessor falls back only for an absent or JSON-null member:

```java
public JSONObject getJSONObject(final String name, final JSONObject defaultValue) {
  if (isNull(name))
    return defaultValue;
  return new JSONObject(getElement(name).getAsJsonObject());   // throws IllegalStateException on any other shape
}
```

So the default covers "missing" but not "present and wrong type", which is the case a hostile or buggy client actually produces.

Authentication and the `isEnabled`/`isUserAllowed` gates all run *before* these reads, so this is reachable only by a caller already authorized for MCP. The impact is a wrong error code plus an unhandled-exception stack trace in the server log per malformed request, not a bypass.

### Steps to reproduce

Against a server with MCP enabled, authenticated as an allowed user. Either request returns HTTP 500:

```json
{"jsonrpc": "2.0", "id": 1, "method": "ping", "params": ["a"]}
```

```json
{"jsonrpc": "2.0", "id": 2, "method": "tools/call",
 "params": {"name": "list_databases", "arguments": ["a"]}}
```

As a JUnit case in `server/src/test/java/com/arcadedb/server/mcp/MCPServerPluginTest.java`, which fails on `main` with the `IOException` above:

```java
@Test
void toolsCallRejectsNonObjectArguments() throws Exception {
  final JSONObject response = mcpRequest(new JSONObject()
      .put("jsonrpc", "2.0")
      .put("id", 900)
      .put("method", "tools/call")
      .put("params", new JSONObject()
          .put("name", "list_databases")
          .put("arguments", new JSONArray().put("a"))));

  assertThat(response.has("error")).isTrue();
  assertThat(response.getJSONObject("error").getInt("code")).isEqualTo(-32602);
}
```

### Notes

`prompts/get` had the same defect and was fixed in #5584 (`f85a375ea`) by moving both member reads inside the handler's `try` and adding `IllegalStateException` to the `-32602` arm. That fix was deliberately scoped to the surface that PR introduced; the two sites above predate it and are left for this issue.

The `params` site is the more interesting of the two, because a `catch` on `dispatch`'s existing `try` would only help if the read moved inside it. Fixing both sites the same way `prompts/get` was fixed would make the three surfaces consistent.

robfrank:
## What does this PR do?

Makes every malformed JSON-RPC member on the MCP endpoint produce a JSON-RPC error envelope instead of failing the request.

The defaulting JSON accessors (`getString(name, default)`, `getJSONObject(name, default)`) fall back only when the member is **absent or null**. A member that is present but of another JSON type raises out of the read instead. Four such reads sat *above* the `try` that was meant to answer for them:

| Read | Was | Now |
|---|---|---|
| `method` in `dispatch` | HTTP 500, no body | `-32600` |
| `params` in `dispatch` | HTTP 500, no body | `-32600` |
| `name` / `arguments` in `toolsCall` | HTTP 500, no body | `-32602` |
| `uri` in `resourcesRead` | HTTP 500, no body | `-32602` |
| `name` / `arguments` in `promptsGet` | `-32603` | `-32602` |

The `dispatch` pair is the worst of them: their enclosing `try` carries only a `finally` (the `DatabaseContext` cleanup) and no `catch`, so the raise left the dispatcher entirely and the transport turned it into a bodiless HTTP 500. That affects **every** method, `ping` included, not only the two that read an `arguments` member.

`method` and `params` are members of the request object, so a wrong shape there is `-32600 Invalid Request`. `uri`, `name` and `arguments` live inside `params`, so those are `-32602 Invalid params`.

## Motivation

Closes #5585. Found while reviewing #5584, which fixed the same defect for `prompts/get` only; that fix was deliberately scoped to the surface that PR introduced, and this one covers the sites that predate it.

## Additional Notes

**Two exception types, not one.** Gson raises `IllegalStateException` when `getAsString()` or `getAsJsonObject()` meets an array of the wrong arity, but `UnsupportedOperationException` when `getAsString()` meets an object. The `prompts/get` fix in #5584 caught only the first, which is why `{"name": {}}` there answered `-32603`. Every catch here covers both.

**A Gson coercion worth knowing about.** A single-element array coerces: `{"method": ["ping"]}` yields the string `"ping"` and dispatches normally. The tests therefore use `{}` rather than a one-element array, with a comment saying why - the obvious payload silently tests nothing.

## Checklist

Five tests added to `MCPTransportConformanceTest`, which already owns envelope-level conformance (invalid ids, notifications, batching). All five were confirmed to fail before the fix: four with `expected: 200 but was: 500`, and the `prompts/get` one with `expected: -32602 but was: -32603` after temporarily reverting that catch.

`MCP*Test` is 265 tests, 0 failures, on a `mvn clean` build.

### DIFF
--- a/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
+++ b/server/src/main/java/com/arcadedb/server/mcp/MCPDispatcher.java
@@ -201,8 +201,20 @@ public class MCPDispatcher {
     if (!isValidRequestId(id))
       return error(null, -32600, "Invalid Request: 'id' must be a string or an integer", 200);
 
-    final String method = request.getString("method", "");
-    final JSONObject params = request.getJSONObject("params", new JSONObject());
+    // Read both members under a guard rather than inline. The defaulting accessors substitute the default only for
+    // an absent or null member, so a member that is present but of another JSON type raises out of the read
+    // instead. These two reads sit above the try below, whose only block is a finally, so an unguarded raise here
+    // escapes dispatch entirely and reaches the transport as a bodiless HTTP 500 rather than a JSON-RPC envelope.
+    // Both members belong to the request object itself, which makes a wrong shape an invalid request rather than
+    // invalid params.
+    final String method;
+    final JSONObject params;
+    try {
+      method = request.getString("method", "");
+      params = request.getJSONObject("params", new JSONObject());
+    } catch (final IllegalStateException | UnsupportedOperationException e) {
+      return error(id, -32600, "Invalid Request: 'method' must be a string and 'params' an object", 200);
+    }
 
     LogManager.instance().log(this, Level.INFO, "MCP[%s] %s (user=%s)", transport, method, user.getName());
 
@@ -258,7 +270,15 @@ public class MCPDispatcher {
   }
 
   private MCPResponse resourcesRead(final Object id, final JSONObject params, final ServerSecurityUser user) {
-    final String uri = params.getString("uri", "");
+    // Guarded for the same reason as the request members in dispatch: a 'uri' present but of another JSON type
+    // raises out of the read, which sits above the try below. A wrong shape is invalid params, distinct from the
+    // resource-not-found the try answers with for a URI that is well formed but names nothing readable.
+    final String uri;
+    try {
+      uri = params.getString("uri", "");
+    } catch (final IllegalStateException | UnsupportedOperationException e) {
+      return error(id, -32602, "Invalid params: 'uri' must be a string", 200);
+    }
 
     LogManager.instance().log(this, Level.INFO, "MCP[%s] resources/read '%s' (user=%s)", transport, uri, user.getName());
 
@@ -299,7 +319,7 @@ public class MCPDispatcher {
     } catch (final SecurityException e) {
       LogManager.instance().log(this, Level.INFO, "MCP[%s] prompts/get -> permission denied: %s", transport, e.getMessage());
       return error(id, -32600, e.getMessage(), 200);
-    } catch (final IllegalArgumentException | IllegalStateException e) {
+    } catch (final IllegalArgumentException | IllegalStateException | UnsupportedOperationException e) {
       return error(id, -32602, "Invalid params: " + e.getMessage(), 200);
     } catch (final Exception e) {
       LogManager.instance().log(this, Level.WARNING, "MCP[%s] prompts/get -> error: %s", transport, e.getMessage());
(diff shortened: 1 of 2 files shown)


robfrank:
### `isResponse` reading `jsonrpc`: confirmed and fixed

Reproduced before fixing: `{"jsonrpc":{},"id":1}` returned **HTTP 500**, exactly as described. The probe runs as the first statement in `dispatch`, so it sits above every guard the PR added, and it is genuinely the one read that cannot be placed under one - it decides whether a reply is owed at all, so there is no envelope to answer with yet.

Fixed in `655b2a676`, using the `opt` approach suggested:

```java
return "2.0".equals(message.opt("jsonrpc"))
    && message.has("result") != message.has("error");
```

`opt` maps any JSON shape to an object rather than demanding a string, so a `jsonrpc` that is not the string `"2.0"` simply means the payload is not a response, and it falls through to normal dispatch. Note this also tightens one case that previously passed: a numeric `{"jsonrpc": 2.0, ...}` used to be coerced to `"2.0"` and treated as a response; it no longer is, which matches JSON-RPC requiring the literal string.

A test covers both raising shapes, the object and the multi-element array, since those go through different Gson exceptions:

```java
for (final String jsonrpc : new String[] { "{}", "[\"2.0\",\"x\"]" }) { ... }
```

Confirmed failing before the fix with `[payload with jsonrpc={}] but was: 500`. `MCP*Test` is now 266 tests, 0 failures.

### Single-element array coercion: filed as #5620

Filed rather than folded in, agreeing with your read that it is a different defect. Worth recording there: the arity of the array decides the outcome, which is not a distinction JSON-RPC makes. `["ping"]` executes, `["ping","x"]` is rejected.

The issue also notes that fixing it properly means type-checking the member (`opt(name) instanceof String`) rather than catching what the accessor raises - which would make the `-32600`/`-32602` guards in this PR redundant on the string-typed members. So the two are better done together than layered, and this PR is the wrong place for it.

### Combined guard messages: leaving as-is

Declining this one. Splitting `"'name' must be a string and 'arguments' an object"` into per-member messages means a separate try block per member at three sites, roughly doubling the guard code. The payload that triggers it is a client bug, both candidate members are named, and the JSON-RPC code already says which half of the request is at fault (`-32600` for a request member, `-32602` for a params member). The precision gained does not carry that much weight.

Happy to reverse this if you would rather have exact messages.

PR #5623, opened 2026-07-31
fix(mcp) #5620: type-check string-typed JSON-RPC members instead of coercing them

### LINKED ISSUE #5620: MCP: a single-element array is coerced to a string, so {"method":["ping"]} executes
### ArcadeDB Version:

`26.8.1-SNAPSHOT`

### OS and JDK Version:

Any. This is JSON binding behavior, not platform-specific.

### Expected behavior

JSON-RPC 2.0 requires `method` to be a String. A request whose `method` member is an array should be rejected as an invalid request, not executed.

### Actual behavior

A **single-element** array is silently coerced to that element's string form and dispatched as a normal call:

```json
{"jsonrpc": "2.0", "id": 1, "method": ["ping"]}
```

executes `ping` and returns a success result. The same holds for any single-element array on a string-typed member: `params.uri` on `resources/read`, `params.name` on `tools/call` and `prompts/get`.

The cause is Gson, reached through `JSONObject.getString(name, default)` -> `getElement(name).getAsString()`. `JsonArray.getAsString()` delegates to its single element when `size() == 1`, and throws `IllegalStateException` only for any other size:

| value | `getAsString()` |
|---|---|
| `["ping"]` | returns `"ping"` |
| `["ping","x"]` | `IllegalStateException` |
| `{}` | `UnsupportedOperationException` |

So the arity of the array decides whether the request is rejected or executed, which is not a distinction JSON-RPC makes.

### Steps to reproduce

Against a server with MCP enabled, authenticated as an allowed user:

```json
{"jsonrpc": "2.0", "id": 1, "method": ["ping"]}
```

returns `{"jsonrpc":"2.0","id":1,"result":{}}` where an invalid-request error is expected.

Confirmed empirically while writing #5619: a conformance test asserting an error for `"method": ["ping"]` failed on the assertion that an `error` member exists, because the request had succeeded. That test now uses `{}` instead, with a comment, precisely because the obvious payload silently tests nothing.

### Notes

Low impact: the effect is accepting a slightly malformed request rather than rejecting it, and the coerced value is the one the client evidently meant. No security consequence, since the coerced method still goes through the same authentication, profile and permission gates.

Deliberately left out of #5619, whose scope is malformed members producing a JSON-RPC error envelope instead of an HTTP 500. Fixing this one is a different change: it means type-checking the member rather than catching what the accessor raises, for example comparing `opt(name)` against `String` before use, the way `isResponse` now compares `jsonrpc`. That would also make the `-32600` / `-32602` guards in #5619 unnecessary on the string-typed members, so the two are worth doing together rather than layering.

robfrank:
Closes #5620

`MCPDispatcher` read every string-typed protocol member through `JSONObject.getString(name, default)`, which ends in Gson's `getAsString()`. Gson converts rather than checks: `JsonArray.getAsString()` delegates to its single element when `size() == 1` and raises for any other size, and a bare number or boolean is stringified. The arity of an array therefore decided the outcome, so `{"jsonrpc":"2.0","id":1,"method":["ping"]}` executed `ping` and returned a success result while `["ping","x"]` was refused, and `{"method":42}` reached the dispatch switch as `"42"` and was answered `-32601` rather than as an invalid request. JSON-RPC 2.0 types `method` as a String; it draws no distinction by arity. The same applied to `params.uri` on `resources/read` and `params.name` on `tools/call` and `prompts/get`. This adds `MCPDispatcher.stringMember`, which decides on the value's own JSON type by comparing `JSONObject.opt` against `String` - `opt` maps each JSON type to its Java counterpart without converting between them - and raises `IllegalArgumentException` for any other shape. Absent and JSON-null still fall back to the default, so a missing member stays distinguishable from a malformed one.

### Merged with #5619

#5619 merged to main as `bfdc3a837` while this branch was open, so `origin/main` has been merged in and the overlap resolved here. The two are complementary and the result keeps both:

- **String-typed members** (`method`, `params.uri`, `params.name`) now read through `stringMember`. #5619's guard placement, error codes and messages are kept as-is; only the read changed, and `IllegalArgumentException` was added to each catch.
- **Object-typed members** (`params`, `params.arguments`) keep #5619's accessor and catch untouched - `getAsJsonObject()` raises for an array of any size, so arity never decided anything there.
- **`isResponse`** keeps #5619's `opt("jsonrpc")` comparison; **`promptsGet`** keeps its widened catch.

### Why these members are narrowed and `arguments.key` is not

Narrowing a read to `instanceof String` is unsafe when a *second* reader of the same member still coerces - that is the `formatArgs` / `SetServerSettingTool` masking bypass that #5622 is fixing by catching instead. Every member narrowed here was grepped for other readers first:

| member | other readers | coercing? |
|---|---|---|
| `method` | `has("method")` in `dispatch` and `isResponse` | existence checks only |
| `params.uri` | none | - |
| `params.name` (tools/call) | `formatArgs(toolName, args)` | receives the value resolved here, not a second read |
| `params.name` (prompts/get) | none | - |

Each has exactly one coercing reader, which is the one narrowed. `arguments.key` - the member that does have two readers - is deliberately untouched, so this does not conflict with #5622.

## Test plan

Six methods added to `MCPTransportConformanceTest`, which now runs 36 (24 pre-existing + 6 from #5619 + 6 here). Every payload names a target that exists, so on the unfixed code the coerced value was dispatched and the call succeeded - the assertions were verified to fail before the fix, not merely to pass after it.

- [x] `"method":["ping"]` is answered `-32600` instead of executing `ping`
- [x] `"method":42` and `"method":true` are answered `-32600` instead of `-32601 Method not found`
- [x] `tools/call` with `"name":["list_databases"]` is answered `-32602` instead of listing the databases
- [x] `resources/read` with `"uri":["arcadedb://graph/schema"]` is answered `-32602` instead of returning the schema
- [x] `prompts/get` with `"name":["graphrag_query"]` and both required arguments supplied is answered `-32602` (the arguments are supplied on purpose: without them the request fails on the missing-argument path with the same code and the test would prove nothing)
- [x] absent `params`, `"params":{}` and `"params":{"name":null}` still fall back to the default and still produce the unknown-tool `isError` envelope rather than `-32602`
- [x] `mvn -pl server test -Dtest=MCPTransportConformanceTest` - 5 failures before the fix, 0 after; 36/36 after merging main
- [x] Regression sweep `mvn -pl server test -Dtest='MCP*Test,HybridSearch*Test,GetServerSettings*'` - `Tests run: 282, Failures: 0, Errors: 0, Skipped: 0`

One comment inherited from #5619 was corrected rather than carried over: `requestWithNonStringMethodIsRejected` justified its use of `{}` by saying `"method":["ping"]` "would silently succeed", which this change makes untrue.

### DIFF
--- a/docs/5620-mcp-single-element-array-coercion.md
+++ b/docs/5620-mcp-single-element-array-coercion.md
@@ -135,6 +135,46 @@ went through the same authentication, profile and permission gates.
 The guards sit on the request-parsing path and cost one `instanceof` per member, replacing an accessor call
 that did strictly more work.
 
+## Review
+
+PR: https://github.com/ArcadeData/arcadedb/pull/5623
+
+### Cycle 1 - `216cfe73a`
+
+No `claude` review arrived within the 15-minute window; only `codacy-production[bot]` reported (0 issues,
+0 complexity). While that cycle was waiting, #5619 merged to main, so `origin/main` was merged in and the
+overlap resolved as described above. That produced `61bb20fa5`.
+
+### Cycle 2 - `61bb20fa5`
+
+`claude[bot]` reviewed: LGTM, with three observations marked non-blocking. All three were assessed and none
+resulted in a code change. Nothing was deferred.
+
+**1. "`resourcesRead` hardcodes the catch message while `toolsCall`/`promptsGet` use `e.getMessage()`" -
+declined, premise is inverted.** Three of the four guards hardcode their message (`dispatch`,
+`resourcesRead`, `toolsCall`); only `promptsGet` relays `e.getMessage()`. The split is structural rather
+than accidental:
+
+- A guard wrapped tightly around member reads states the expected shape, because its catch also covers
+  Gson's `IllegalStateException` / `UnsupportedOperationException` from the object-typed read beside it.
+  Those carry internal text such as `Not a JSON Object: ["a"]`, so relaying them would both read as noise
+  and echo fragments of the request payload back to the client.
+- `promptsGet`'s catch is the broad handler one: it also covers `MCPPrompts.get` raising
+  `IllegalArgumentException` for an unknown prompt, where the message is the whole point of the answer.
+
+Aligning `resourcesRead` to `e.getMessage()` would make it inconsistent with the two guards it sits next to
+in order to match the one that is a different kind of catch. The messages are also #5619's, unchanged here
+on purpose.
+
+**2. "`promptsGet`'s catch now also absorbs `IllegalArgumentException` from `MCPPrompts.get`" - no action,
+and the review agrees.** That catch already listed `IllegalArgumentException` before this change, so the
+mapping of unknown-prompt to `-32602` is pre-existing and untouched.
+
+**3. "The doc still cites pre-merge counts (30/276) while the PR body says 36/282" - declined, already
+correct.** The committed doc carries both, deliberately: the 30/276 figures are the before-and-after of the
+fix itself and are what demonstrate the tests failed first, while the post-merge 36/282 figures follow on
+the next two lines.
+
 ## Follow-ups
 
 - PR #5622 is still open and fixes `formatArgs`' read of `arguments.key` by catching. It does not conflict
--- a/docs/5620-mcp-single-element-array-coercion.md
+++ b/docs/5620-mcp-single-element-array-coercion.md
@@ -1,0 +1,113 @@
+# 5620 - MCP: a single-element array is coerced to a string
+
+Issue: https://github.com/ArcadeData/arcadedb/issues/5620
+
+## Root cause
+
+`MCPDispatcher` read every string-typed protocol member through
+`JSONObject.getString(name, default)`, which is `if (isNull(name)) return default;` followed by
+`getElement(name).getAsString()`. That last call is Gson's, and Gson converts rather than checks:
+
+| value          | `getAsString()`                |
+|----------------|--------------------------------|
+| `"ping"`       | `"ping"`                       |
+| `["ping"]`     | `"ping"` - unwrapped silently  |
+| `["ping","x"]` | `IllegalStateException`        |
+| `42` / `true`  | `"42"` / `"true"`              |
+| `{}`           | `UnsupportedOperationException` |
+
+`JsonArray.getAsString()` delegates to its single element when `size() == 1`. The arity of the array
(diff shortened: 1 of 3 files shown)


robfrank:
Thanks - the root-cause and test-design points match my reasoning. On the three observations, all three were checked and none produced a code change. Two of them rest on a misreading, so recording the detail here rather than silently skipping.

**1. Catch-message consistency - the premise is inverted.** `toolsCall` does not use `e.getMessage()`; it hardcodes its message exactly as `resourcesRead` does. Three of the four guards hardcode (`dispatch` L220, `resourcesRead` L286, `toolsCall` L353) and only `promptsGet` (L329) relays. That split is structural:

- A guard wrapped tightly around the member reads states the expected shape because its catch also covers Gson's `IllegalStateException`/`UnsupportedOperationException` from the object-typed read next to it. Those carry internal text like `Not a JSON Object: ["a"]` - relaying it would read as noise and echo request-payload fragments back to the client.
- `promptsGet`'s catch is the broad handler catch: it also covers `MCPPrompts.get` raising `IllegalArgumentException` for an unknown prompt, where the message *is* the answer.

Aligning `resourcesRead` to `e.getMessage()` would make it inconsistent with the two guards beside it in order to match the one that is a different kind of catch. These messages are also #5619's, left unchanged on purpose.

**2. `promptsGet` absorbing `IllegalArgumentException` from `MCPPrompts.get`** - agreed, and as you note it is pre-existing: that catch already listed `IllegalArgumentException` before this PR. No change.

**3. Doc test counts** - already correct in the committed file. `docs/5620-mcp-single-element-array-coercion.md` carries both sets deliberately: 30/276 are the before-and-after of the fix itself (what shows the tests failed first), and 36/282 follow on the next two lines as the post-merge figures.

```
122:- Before the fix: `Tests run: 30, Failures: 5` - the five new malformed-member tests, and only those.
123:- After the fix: `Tests run: 30, Failures: 0`.
124:- After merging `origin/main` (which brought #5619's six tests in): `Tests run: 36, Failures: 0`.
126:  `Tests run: 276, Failures: 0` before the merge, `Tests run: 282, Failures: 0, Errors: 0, Skipped: 0` after.
```

The follow-up you flagged (other handlers reading request JSON via `getString(name, default)`) is worth filing, with the caveat already in the PR body: each site needs the two-reader check before anything is narrowed, since narrowing a member whose second reader still coerces is what makes the `arguments.key` masking bypass in #5622.

PR #5624, opened 2026-07-31
fix(mongodbw) #5583: bind $set and replacement values as parameters instead of inlining them as JSON

### LINKED ISSUE #5583: mongodbw: bind $set and replacement values as parameters instead of inlining them as JSON
Follow-up to #5579 / PR #5581.

#5581 bound every **filter** value as a SQL parameter, so nothing a client sends reaches the `WHERE` clause text.
That leaves one path in `MongoDBToSqlTranslator`'s neighbourhood still building statement text out of client data:
the **update values**.

**This is not a known-exploitable hole.** `JSONObject` does its own escaping, and #5581 added two wire tests that
prove it holds (`aQuoteBearingValueSurvivesBeingSetByAnUpdate`, `aQuoteBearingValueSurvivesAFullReplacement`,
both passing with `'`, `"` and `\` in the value). The point is consistency: after #5581 the "unreachable by
construction" property covers the `WHERE` clause and `$inc`, but the update values still rely on a second,
different escaping mechanism being correct. That is exactly the per-call-site invariant #5579 set out to remove,
just relocated.

### Current shape

`MongoDBDatabaseWrapper.appendUpdateOperations` spells both forms into the statement as a JSON literal:

```java
// full replacement (no $-prefixed keys) - line 628
sql.append(" CONTENT ").append(documentToJson(u));

// $set - line 636
case "$set" -> sql.append(" MERGE ").append(documentToJson(operand));
```

`documentToJson` (line 682) walks the BSON document into a `com.arcadedb.serializer.json.JSONObject`, and the
value is escaped only by `JSONObject.toString()`.

### Proposed change

Have `MERGE` / `CONTENT` accept a bound parameter in place of the JSON literal, so the update values join the
filter values in the parameter map that `appendUpdateOperations` already receives:

```java
sql.append(" CONTENT :").append(bind(params, documentToMap(u)));
```

This is a real change to the update path, not a mechanical edit like #5581 was:

- it needs `UPDATE ... MERGE` / `UPDATE ... CONTENT` to take an input parameter where the grammar currently wants
  a JSON literal - **that needs checking against the grammar first**, and may be the bulk of the work
- the bound value would be a `Map`, so the BSON -> `JSONObject` conversion becomes BSON -> `Map`, and the nested
  `Document` / `List` handling in `toJsonValue` has to be preserved

If the grammar cannot bind there, closing this as "won't fix, escaping is the mechanism here" is a legitimate
outcome - but it should be a recorded decision rather than something nobody looked at.

### Notes

- Field **names** in `$unset` / `$inc` stay on `quoteFieldPath()`. SQL cannot bind a property name, so that half is
  unchanged, same as in #5575 and #5581.
- `$inc` already binds its operand as of #5581.

### Verification

The two `$set` / `replaceOne` round-trip tests in `MongoDBParameterBindingTest` already pin the observable
behaviour and must stay green. Worth confirming they still fail if the binding is reverted, so a parameterised
build cannot pass them for the wrong reason.

robfrank:
Closes #5583

Follow-up to #5579 / PR #5581. After #5581 every **filter** value is a bound SQL parameter, but the **update** values (`$set` and full replacement) were still spelled into the statement as a JSON literal, so their safety rested on `JSONObject.toString()` escaping being correct at that one call site. `MongoDBDatabaseWrapper.appendUpdateOperations` now binds them instead: `UPDATE ... MERGE :p<n>` and `UPDATE ... CONTENT :p<n>`, joining the filter values in the parameter map the method already receives. The BSON walk that fed `JSONObject`/`JSONArray` becomes a BSON -> `LinkedHashMap`/`ArrayList` walk, preserving the nested `Document`/`List` recursion and the `ObjectId` -> hex-string conversion exactly. With that, "unreachable by construction" covers the whole statement rather than just the `WHERE` clause and `$inc`.

## The open question the issue posed: can the grammar bind there?

**Yes, and no engine change was needed.** The issue flagged this as needing checking first and possibly being the bulk of the work.

`GlobalConfiguration` (line 474) records that the JavaCC parser is dead: *"Deprecated, has no effect. The ANTLR4-based SQL parser is always used."* `StatementCache` only ever calls `SQLAntlrParser.parse`, so the authority is `SQLParser.g4`:

```
    | MERGE expression
    | CONTENT (json | jsonArray | inputParameter)
```

- `MERGE expression` accepts `:p0` as an ordinary expression. `SQLASTBuilder.visitUpdateOperation` already detects that the expression is not a JSON literal and stores it in `UpdateOperations.expression`; `UpdateMergeStep.resolveExpression` resolves it and accepts a `Map` with String keys.
- `CONTENT ... inputParameter` is an explicit grammar alternative, handled by `UpdateContentStep` via `doc.fromMap(...)`.

The JavaCC `SqlParser.java` production still only accepts `Json()` after MERGE/CONTENT, but that parser is unreachable and there is no javacc plugin in the build (the generated file is checked in), so it was left alone rather than hand-edited.

## Deliberately unchanged

`$unset` and `$inc` **field names** stay on `quoteFieldPath()`. SQL cannot bind a property name, so that half of #5575 is untouched, same as in #5581. `$inc` already binds its operand.

## Side effect: value fidelity

Same class of latent defect #5581 uncovered on the filter path. Routing through `JSONObject` reshaped a value before it reached the record; binding hands the engine the original Java object. A `java.util.Date` now arrives as a `Date`, and a large `Double` no longer arrives in scientific notation. Neither was a security hole - `JSONObject` escapes correctly, and #5581 added round-trip tests proving it - but both were fidelity losses.

## Note on two existing tests

`aQuoteBearingValueSurvivesBeingSetByAnUpdate` and `aQuoteBearingValueSurvivesAFullReplacement` are unchanged in behaviour and assertions. Their **comments** were edited: each described the value as travelling "as a JSON literal", which this change makes false. Flagging it explicitly since the workflow otherwise leaves existing tests alone.

## Test plan

- [x] `mvn -pl mongodbw verify` - **67 tests green**, up from 52 (+12 unit, +3 wire)
- [x] New unit tests `MongoDBUpdateValueBindingTest` (12, no server) read the generated SQL directly and pin: `MERGE :p0` / `CONTENT :p0`; quote- and backslash-bearing values absent from the statement text; nested `Document` -> nested `Map`; `List` recursion including documents nested inside a list; `ObjectId` -> hex; `Date` bound as a `Date`; update and filter values sharing one placeholder numbering without collision; `$unset`/`$inc` field names still quoted; a crafted field name inside a `$set` operand being data rather than syntax
- [x] New wire-level tests in `MongoDBParameterBindingTest` (3, real server through the Mongo driver) round-trip a nested document and an array through `$set`, and a nested document through `replaceOne`. These are what prove the engine actually accepts `MERGE :p0` / `CONTENT :p0` at runtime, not just that the grammar file lists the alternatives
- [x] The two pre-existing round-trip tests the issue named stay green
- [x] **Red proven before the fix**: `appendUpdateOperations` was first made package-private `static` with its body still on `documentToJson`, and the new class run against it gave `Tests run: 12, Failures: 6, Errors: 4`. The 2 that passed are the two pinning deliberately-unchanged `$unset`/`$inc` behaviour, which must be green on both sides. The binding was applied only after that run
- [x] No engine module touched

### DIFF
--- a/docs/5583-mongodbw-bind-update-values-as-params.md
+++ b/docs/5583-mongodbw-bind-update-values-as-params.md
@@ -150,7 +150,9 @@ than the edit. Nothing was weakened - the assertions still pin the same observab
 - `{field: null}` binds `field = :p0` with null, which does not match missing fields the way MongoDB does.
   Carried over from #5581, unfiled, not a regression.
 - `executeUpsert` / `applyOperatorsToDocument` build a record through the API rather than through SQL, so they
-  never had this exposure and are untouched.
+  never had this exposure and are untouched. They do, however, call `record.set(key, rawBsonValue)` with no
+  conversion, so unlike the update path they never turn a nested `ObjectId` into hex: the same `$set` stores a
+  different representation depending on whether the row already existed. Pre-existing, worth its own issue.
 
 ## Review cycles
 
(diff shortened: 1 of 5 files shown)


robfrank:
## Scope note after review cycle 4 - please read before merging

Cycle 4 asked for a **wire-level** test of the Date fidelity claim (the unit test only asserted on the parameter map). Writing it exposed a crash, so this PR now carries one fix beyond the issue's scope.

### What the test found

`aDateSurvivesBeingSetByAnUpdate` failed on the **read-back**, not the update:

```
SEVER [MongoWireMessageEncoder] Failed to encode {...}
java.lang.IllegalArgumentException: Unknown type: class java.time.LocalDateTime
    at de.bwaldvogel.mongo.wire.bson.BsonEncoder.determineType(BsonEncoder.java:204)
```

The write succeeded; encoding the response threw, killing the connection (`MongoSocketReadException: Prematurely reached end of stream` client-side).

**It is pre-existing.** Confirmed with a throwaway probe on the **insert** path, which this PR does not touch: `insertOne(new Document("when", date))` then `find` fails identically. `convertDocumentToMongoDB` passed stored values through untouched, and a temporal property is held as a `java.time` value the encoder rejects. Any date written through mongodbw and read back killed the connection.

**But this PR made it reachable from a path where it wasn't.** Previously a `$set` date went through `JSONObject` and was stored as a *string*, which encodes fine. Bound, it is stored as a real `LocalDateTime`. Without a fix this PR would have turned a lossy-but-working path into a crashing one - and no unit test could have caught it, since the failure is in the response encoder.

### The fix

`convertDocumentToMongoDB`'s two identical overloads now share one `convertMapToMongoDB`, routing values through `toBsonValue`. `BsonEncoder` accepts `java.time.Instant` and **not** `java.util.Date` - the first attempt converted to `Date` and failed the same way with `Unknown type: class java.util.Date`, so the target type was determined from the encoder's bytecode rather than assumed. `LocalDateTime`/`LocalDate` anchor at `ZoneOffset.UTC`, matching `DateUtils` on the way in, so the round trip is exact. It recurses into embedded documents and lists.

Also applied from cycle 4: `assert params.isEmpty()` on `appendUpdateOperations` (`assert` is already used in that file).

### Two things for you to decide

1. **The temporal fix has not been through a bot review cycle** - the 4-cycle limit was reached when it was written. It is also the one part of this PR beyond the issue's scope. It is here because omitting it ships a regression, but splitting it into its own PR is entirely reasonable.
2. **Two pre-existing gaps worth their own issues**, both verified, neither changed here:
   - Dotted `$set` keys do not navigate (`{"$set": {"address.city": ...}}` stores a literal `address.city` property), while the filter side's `quoteFieldPath()` does split per segment. The two halves disagree.
   - `executeUpsert` -> `applyOperatorsToDocument` calls `record.set(key, rawBsonValue)` with no conversion, so unlike the update path it never turns a nested `ObjectId` into hex: the same `$set` stores a different representation depending on whether the row already existed.

Module total is now **73 green**, up from 52 on main.

PR #5633, opened 2026-07-31
fix(engine) #5615: keep vectors the graph build leaves unreachable searchable via the delta scan

### LINKED ISSUE #5615: LSM vector index: a committed vector is intermittently missing from search (cause unknown)
> **The "Where I think it comes from" section below is disproven.** The race it describes did not reproduce, and the fix it implies regresses two tests because the id cutoff is load-bearing for evicting rejected vectors. See [this comment](https://github.com/ArcadeData/arcadedb/issues/5615#issuecomment-5142589956) for the evidence and the re-scoped leads. The symptom and its nondeterminism still stand.

## Summary

A search against an `LSM_VECTOR` index can intermittently fail to return a vector that was committed moments earlier. Observed as a recurring flake in the Python bindings suite, but the mechanism is in the engine, not the bindings.

## Symptom

`bindings/python/tests/test_vector.py::TestLSMVectorIndex::test_document_vector_search` fails at its second query:

```
tests/test_vector.py:1675: in test_document_vector_search
    assert "car" in found_names_v
E   AssertionError: assert 'car' in ['truck', 'banana']
```

The fixture is four documents on a 4-dimension index:

| name | vector |
|---|---|
| apple | `[0.9, 0.1, 0, 0]` |
| banana | `[0.8, 0.2, 0, 0]` |
| car | `[0.1, 0.9, 0, 0]` |
| truck | `[0.2, 0.8, 0, 0]` |

The query is `[0.05, 0.95, 0, 0]` with `k=2` and `approximate=False`. Returning `banana` ahead of `car` is wrong under any distance metric, so `car` was **absent from the search**, not mis-ranked.

## It is nondeterministic

`test-python-bindings.yml` runs twice per commit, once on `push` and once on `pull_request`. On commit `34c5466` (PR #5548, which changed nothing but four dependency floors in a `pyproject.toml` that this workflow does not even read):

- push-triggered run `30477866110` -> **passed**
- pull_request-triggered run `30477869274` -> **failed**
- full re-run of the same run -> **passed**

Same SHA, divergent outcomes. Seen on `windows/amd64, Python 3.12`; nothing about the mechanism looks Windows-specific, so that is likely just timing.

## Where I think it comes from

Commit replay deliberately skips the O(log n) HNSW insert (issue #3864) and parks the vector in a delta buffer, `LSMVectorIndex.java:4025-4051`:

```java
final int id = nextId.getAndIncrement();
persistVectorWithLocation(id, rid, vector);
...
// Add to delta buffer so the vector is visible in search via mergeWithDeltaScan.
deltaVectors.add(new DeltaVectorEntry(id, rid, vf));
```

Searches then fold that buffer in via `mergeWithDeltaScan()`, so a vector is expected to be reachable through the graph, the delta buffer, or both, at all times.

The invariant looks losable across a rebuild. `buildGraphFromScratchExclusively()` captures the cutoff **unlocked** at `LSMVectorIndex.java:1452`:

```java
final int deltaSnapshotId = nextId.get();
```

and `buildGraphFromScratchWithRetry()` is explicit that this phase is not mutually exclusive with writers (`LSMVectorIndex.java:1427-1431`): the validation "runs unlocked (issue #5391)", and "searches and inserts never touch" `graphBuildLock`. The expensive build then reads vectors from persisted pages. On completion the write lock is reacquired and the buffer is trimmed on the assumption that everything below the cutoff made it into the new graph, `LSMVectorIndex.java:1988-1993`:

```java
// Keep only delta entries inserted DURING the rebuild (vectorId >= snapshot)
final List<DeltaVectorEntry> remaining = new ArrayList<>();
for (final DeltaVectorEntry e : deltaVectors)
  if (e.vectorId >= deltaSnapshotId)
    remaining.add(e);
this.deltaVectors = remaining;
```

So an insert that has already consumed an id below `deltaSnapshotId`, but whose page content the builder's enumeration did not pick up, is dropped from the delta buffer while being absent from the freshly published graph. It is then in neither, and stays invisible until some later full rebuild happens to re-read its page.

Contention widens the window: the inactivity rebuild serializes on a single JVM-wide `REBUILD_SEMAPHORE`, and the vector suite runs many indexes in one JVM.

I want to be clear about what is verified and what is not. Verified: the symptom, its nondeterminism, and the quoted code. **The specific interleaving above is a hypothesis** that fits the evidence; I have not built a deterministic reproducer for it.

## Secondary observation

`deltaVectors` is a plain `ArrayList` behind a `volatile` reference (`LSMVectorIndex.java:250`). It is mutated in place under the write lock (`:4039`, `:4150`, `:4227` `removeIf`) and wholesale-replaced at `:1993`, but `mergeWithDeltaScan()` takes the bare reference and iterates it:

```java
final List<DeltaVectorEntry> currentDelta = deltaVectors; // volatile snapshot
```

The comment says "snapshot", but `volatile` only publishes the reference, not the contents. Whether readers hold the read lock across that iteration is worth confirming; if any path does not, in-place `add`/`removeIf` during iteration is a `ConcurrentModificationException` or a silently short read.

## Impact

Silent under-return from a vector search, which is a correctness issue rather than only test noise: a caller gets a plausible-looking top-k with a member missing. It also costs CI re-runs on unrelated PRs, and it trains reviewers to dismiss red vector tests.

## Suggested next step

A test that hammers inserts against a forced rebuild and asserts that every committed RID is retrievable would confirm or kill the hypothesis quickly. If confirmed, the cutoff needs to be established atomically with the set of vectors the builder actually read, rather than sampled from `nextId` beforehand.

Found while triaging CI on PR #5548.

robfrank:
Closes #5615

## Root cause

A graph rebuild occasionally emits an **orphan node**: present in the graph and carrying a full set of outgoing edges, but with **zero incoming edges**. Beam search starts at the entry node and only ever follows edges forward, so an orphan can never be visited - at any `efSearch`, however good its score. The vector, the location index, the ordinal map and the scoring path are all correct, which is why the symptom looked like a delta-retention or publication race for so long and why nine earlier theories all failed.

Instrumenting the reproducer to walk the graph at the moment of a miss:

```
batch 15: #1:8382 -> #1:6156
  retryHit=false wideEfHit=false
  rankProbe: expectedOrdinal=1500 expectedScore=1.0 betterScoring=0 bestOrdinal=1500 bestScore=1.0
             graphSize=1600 mapLen=1600
  reach:     entryNode=1050 visited=1599 ofUpperBound=1600
             targetReachable=false targetInGraph=true targetOutDegree=32 targetInEdges=0
```

The missing vector is the globally best-scoring node (`betterScoring=0`), graph and ordinal map agree on size, and a BFS from the entry node reaches 1599 of 1600 nodes - every one except this. `retryHit=false` and `wideEfHit=false` (efSearch=2000 against a 1600-node graph) show the miss is stable rather than a transient read-path race.

A second probe placed directly after `builder.build(vectors)` closed the chain: on a failing run the build reported `ORPHANS count=3 sample=252(deg=32) 257(deg=32) 258(deg=32)` and the search then missed ordinal **258**, one of those three. Passing runs reported zero orphan builds.

`GraphIndexBuilder.build()` does call `cleanup()` (verified in the 4.0.0-rc.7 bytecode: `submit(addNodes) -> join -> cleanup`), and cleanup has connectivity machinery. It nonetheless leaves these nodes unreferenced under concurrent rebuild pressure.

## The change

`findUnreachableOrdinals` walks every edge from the graph entry node after a build and returns the ordinals nothing reaches. `buildGraphFromScratchExclusively` re-queues those vectors into the delta buffer, where `mergeWithDeltaScan` - an exhaustive linear scan - keeps them searchable until the next rebuild wires them into the graph.

- The walk and the vector reads run **before** the write lock is reacquired, so an O(V+E) traversal never stalls concurrent searches.
- Vectors at or past `deltaSnapshotId` are skipped: the existing trim already carries them over, and re-adding them would score the same vector twice in the delta scan.
- Reachability is walked from the entry node rather than counted as in-degree, because a disconnected cycle gives every node an in-edge while remaining unreachable.

This is a safety net at the ArcadeDB layer that restores the "every committed vector is findable" guarantee. It does not address why the builder produces the orphan, which is worth reporting upstream.

## Test plan

- [x] `LSMVectorIndexGraphConnectivityTest` - 5 deterministic cases against graphs whose shape is fixed by the test: a connected chain, a node with out-edges and no in-edges, several orphans at once, a disconnected cycle, and an empty graph. Stubbing `findUnreachableOrdinals` to a no-op fails 3 of the 5, confirming they bind.
- [x] `LSMVectorIndexConcurrentRebuildVisibilityTest` - the end-to-end reproducer, which on a miss now reports whether the vector is reachable from the entry node at all.
- [x] `mvn -pl engine test -Dtest='com.arcadedb.index.vector.*Test'` - 245 tests, 0 failures, 0 errors.

Note the concurrency reproducer is inherently probabilistic: its flake rate on unfixed code swung between 1 run in 1 and 1 run in 10, so a single clean run of it proves little on its own. The deterministic test is what pins the fix.

### DIFF
--- a/docs/5615-lsm-vector-missing-from-search.md
+++ b/docs/5615-lsm-vector-missing-from-search.md
@@ -88,6 +88,12 @@ In addition to the six already recorded on the issue:
 - Only the from-scratch path needs this. Vectors ingested through the live builder are already in the delta
   buffer from insertion, so an orphan there is served by the delta scan until a rebuild absorbs it.
 
+## Scope of the guarantee
+
+This restores "every committed vector is findable" **for the running session, until the next rebuild** - not
+unconditionally. The two gaps below are inherent to serving the recovery from an in-memory buffer, and both are
+bounded by the next rebuild re-detecting the orphan.
+
 ## Known limitations
 
 - **The recovery does not survive a restart.** `deltaVectors` is in-memory, so a restart drops the re-queued
@@ -122,3 +128,25 @@ mvn -pl engine test -Dtest='com.arcadedb.index.vector.*Test'   # 246 tests, 0 fa
 The concurrency reproducer is inherently probabilistic - its flake rate on unfixed code swung between 1 run in
 1 and 1 run in 10 - so a single clean run of it never proves anything on its own. The deterministic test is
 what pins the fix.
+
+## Pull request
+
+https://github.com/ArcadeData/arcadedb/pull/5633
+
+### Review cycles
+
+| Cycle | Head | Outcome |
+|---|---|---|
+| 1 | `ba5ff86` | Sentinel not excluded when re-queueing - **real**, fixed (`isDeletedSentinel` + test). Three further notes assessed and documented rather than coded. |
+| 2 | `4ce9ef6` | Main ask (exception storm in the level loop) rested on a **false premise** - `OnHeapGraphIndex.getNeighborsIterator` returns `EMPTY_NODE_ITERATOR` for an absent node, verified in bytecode, and the suggested `contains()` guard would add a lookup rather than remove one. Pushed back with evidence; corrected the misleading comment that caused it. Applied the entry-node null check and the level-union note. |
+| 3 | `b1eb6e4` | Restart gap - **real**; verified the precise mechanism (the staleness check needs `graphSize < ordinalMap.length`, and an orphan is counted in the graph) and documented it. Guarded the null neighbor iterator. |
+| 4 | `f8a494d` | No blockers. Softened the framing of the guarantee to match what is delivered. |
+
+### Follow-ups not taken here
+
+- **Upstream:** why `GraphIndexBuilder.build()` leaves a node unreferenced despite calling `cleanup()`.
+- **Separate defect, unfiled:** `GraphSearcherPool.borrow` calls `drain()` *before* publishing `pooledGraph` /
+  `pooledEpoch`, so a concurrent `release` reading the still-old matching pair can return a searcher bound to
+  the old graph to the idle queue. Needs two searching threads on one index, so it is not #5615.
+- **Possible enhancement:** a bounded self-heal (one deferred rebuild when orphans are found) if orphan counts
+  are ever observed to be large in practice.
--- a/docs/5615-lsm-vector-missing-from-search.md
+++ b/docs/5615-lsm-vector-missing-from-search.md
@@ -88,6 +88,19 @@ In addition to the six already recorded on the issue:
 - Only the from-scratch path needs this. Vectors ingested through the live builder are already in the delta
   buffer from insertion, so an orphan there is served by the delta scan until a rebuild absorbs it.
 
+## Known limitations
+
+- **The recovery does not survive a restart.** `deltaVectors` is in-memory, so a restart drops the re-queued
+  entries. The persisted graph still physically contains the orphan and reports the same node count as the
+  ordinal map, so the staleness check on load (`graphSize < ordinalMap.length`) sees an up-to-date graph and
+  the vector is unsearchable again until some mutation triggers a rebuild. Closing this would mean persisting
+  the orphan set alongside the graph; it is left open because the builder defect is rare and any rebuild
+  re-detects it.
+- **An idle index keeps scanning the re-queued entries.** Re-queueing does not bump `mutationsSinceSerialize`
+  (counting it would let an orphaned index rebuild itself forever), and the decrement at the end of a rebuild
+  may cancel the inactivity timer, so there is no self-scheduled rebuild to clear them until the next real
+  mutation.
+
 ## Tests
 
 - `LSMVectorIndexGraphConnectivityTest` - deterministic coverage of the detection primitive against graphs
--- a/docs/5615-lsm-vector-missing-from-search.md
+++ b/docs/5615-lsm-vector-missing-from-search.md
@@ -72,8 +72,8 @@ In addition to the six already recorded on the issue:
    snapshotted). Dead: `graphSize == mapLen` on every failure, and the miss survives a retry.
 2. **Stale searcher handed back across a graph swap.** `GraphSearcherPool.borrow` calls `drain()` *before*
    publishing `pooledGraph`/`pooledEpoch`, so a concurrent `release` reading the still-old matching pair can
-   re-pool a searcher bound to the old graph. This is a **real race worth filing separately**, but it is not
-   this bug - the reproducer has one searching thread per index.
+   re-pool a searcher bound to the old graph. A **real race, filed as #5648**, but not this bug - the
+   reproducer has one searching thread per index.
 3. **The build reads a sentinel vector.** `ArcadePageVectorValues.getVector` returns `deletedSentinelVector` on
    eight paths, three of them silent. Instrumented all three: zero hits on a failing run.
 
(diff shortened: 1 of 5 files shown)


robfrank:
Thanks - addressed 3 of the 4 in `b1eb6e4`. Pushing back on item 1, which rests on an incorrect premise.

**1. Exception-driven control flow - the premise does not hold, so there is no exception storm.**

`getNeighborsIterator` does not throw for an absent node. From the 4.0.0-rc.7 bytecode of `OnHeapGraphIndex.getNeighborsIterator(int, int)`:

```
 0: iload_1
 2: getfield      layers:Ljava/util/List;
 5: invokeinterface List.size:()I
10: if_icmplt     17
13: getstatic     NodesIterator.EMPTY_NODE_ITERATOR    <-- level out of range
16: areturn
...
31: invokevirtual ConcurrentNeighborMap.get:(I)...Neighbors;
35: aload_3
36: ifnonnull     43
39: getstatic     NodesIterator.EMPTY_NODE_ITERATOR    <-- node absent from this level
42: areturn
```

Both the out-of-range level and the unmapped node return the shared `EMPTY_NODE_ITERATOR`. The `catch` never fires on this path, so no stack traces are filled and there is no `V * maxLevel` cost. The graph a rebuild publishes is exactly what `builder.build()` returns, i.e. an `OnHeapGraphIndex`.

Guarding with `view.contains(level, node)` would also not be cheaper: `contains` performs the same layer lookup plus neighbor-map lookup that `getNeighborsIterator` already does, so it would add a call rather than remove work.

What was genuinely wrong was my **comment**, which asserted absence was signalled by an exception and is what pointed you at this. Corrected to state that absence yields an empty iterator and that the `catch` is purely defensive for `View` implementations that choose to throw.

**3. `entryNode()` null-safety** - fixed. Explicit null check with its own WARNING, so "graph has no entry node" is distinguishable from a genuine failure.

**4. Reachability model** - documented on the method. The walk unions all levels, which is a superset of beam-reachable, so it can never invent a false orphan; a node reachable only via a higher-level in-edge is not reported, which suits the observed defect (no in-edges at any level). Noted that conservative is the right direction here, since a false orphan costs a duplicate delta entry on every search.

**2. Transient arrays** - agreed and left as-is; noted in the tracking doc as a future optimization if profiling points here.

**Nits:** the `instanceof` is indeed always true today and deliberately defensive. On `docs/` - per-issue docs are the convention in this repo, so it stays.

`mvn -pl engine test -Dtest='com.arcadedb.index.vector.*Test'` - 246 tests, 0 failures.

robfrank:
Addressed in `HEAD`. Confirmed item 1 against the code before writing it up.

**1. Restart gap - real, and narrower than "the delta is lost".** Verified the exact mechanism: on load, `LSMVectorIndex` treats a persisted graph as stale only when `graphSize < ordinalMap.length`. An orphan is physically *in* the graph, so the node count matches the ordinal map, the graph reads as up to date, and no rebuild is triggered. Combined with `deltaVectors` being in-memory, the vector is unsearchable again after a restart until some mutation causes a rebuild. Documented in both the method comment and a new "Known limitations" section. Closing it would mean persisting the orphan set alongside the graph, which is out of scope here given any rebuild re-detects it.

**2. Idle index / inactivity timer - confirmed intended.** You are right that the decrement can take `mutationsSinceSerialize` to zero and cancel the inactivity timer, leaving no self-scheduled path out of the delta scan. That is the accepted price of not bumping the counter, which would let an orphaned index rebuild itself forever. The comment now says both halves rather than only the scan cost.

**3. Null iterator - guarded.** Added an explicit `if (neighbors == null) continue;`. You correctly identified the asymmetry: an NPE there would unwind to the outer catch and abandon the whole walk, treating every remaining node as reachable. Now it skips one node instead.

**4. Double `containsNode`** - left as-is; it keeps the array allocation exact.

`mvn -pl engine test -Dtest='com.arcadedb.index.vector.*Test'` - 246 tests, 0 failures.

robfrank:
Filed the searcher-pool race found during this investigation as #5648 - `GraphSearcherPool.borrow` drains the idle queue before publishing `pooledGraph`/`pooledEpoch`, so a concurrent `release` can re-pool a searcher bound to the replaced graph. Separate mechanism from this PR (it needs two searching threads on one index, which the #5615 reproducer does not have), but it produces the same silent wrong-result symptom. Linked from the tracking doc.

PR #5638, opened 2026-07-31
Scope server and cluster status endpoints to the caller's databases

lvca:
## What

The HTTP routes that enumerate the **whole database registry** (rather than naming one database in the path) reported every database on the node to any authenticated user. A user granted access to one database learned the names of all the others, plus whatever each route hangs off a name.

Three surfaces:

| Route | Leaked |
|---|---|
| `GET /api/v1/cluster` | `databases[]` (name, `bootstrapLastTxId`, `bootstrapFingerprint`, acquire status) and `alerts`, whose `single-bucket-types` payload additionally names **types** per database |
| `GET /api/v1/server?mode=cluster` | `ha.databases[]` (name, quorum) |
| `GET /api/v1/server` (default mode) | `metrics.sparseVectorIndexes`, keyed by database name |
| `POST /api/v1/cluster/bootstrap-state` | per-database SHA-256 directory fingerprints and last transaction ids |

The asymmetry that shows the intent: all seven **mutating** Raft endpoints (`PostAddPeer`, `PostTransferLeader`, `PostStepDown`, `PostLeave`, `DeletePeer`, `PostResyncDatabase`, `PostVerifyDatabase`) call `checkRootUser`; the read siblings called nothing. `GetDatabasesHandler` shows the intended shape for the listing routes - `retainAll(user.getAuthorizedDatabases())` - which these omitted.

## How

**One filter, one place.** `AbstractServerHttpHandler#filterAuthorizedDatabases` sits next to `checkAuthorizationOnDatabase` (the GHSA-x8mg gate) and delegates to `ServerSecurityUser#canAccessToDatabase`, so the wildcard grant is interpreted in exactly one place. The three handlers that each hand-rolled the same `retainAll` idiom - `GetDatabasesHandler`, `GetExistsDatabaseHandler`, `PostServerCommandHandler#listDatabases` - now share it, so there are four call sites of one primitive instead of seven copies of an idiom.

**Scope, don't reject, on the status routes.** `?mode=cluster` is called by `RemoteHttpComponent#requestClusterConfiguration` on *every* remote-driver connection, as whatever user the application uses, to read `ha.leaderAddress` / `ha.replicaAddresses` for routing - and a 403 there is rethrown as `SecurityException` rather than falling back to the direct connection. A blanket root check would have broken every non-root client. So server-level fields stay open and only per-database rows are scoped. There is a test pinning this.

**`ClusterAlerts` becomes authorization-aware** rather than being filtered after the fact: the counts inside each alert message are recomputed from the reduced list, so a scoped caller does not read "3 database(s) failed to acquire" next to one name. Node-scoped alerts (lagging followers) are untouched - they describe the cluster, not a database.

**Two endpoints move behind `checkRootUser`**, where filtering the response would not have been enough:

- `POST /api/v1/cluster/bootstrap-state` - a peer-to-peer RPC with no browser or driver consumer. The *work* is the problem, not just the output: each call SHA-256s every database directory on the node and may open a database deliberately left closed. Peers are unaffected, they forward as root with the cluster token.
- `GET /api/v1/cluster?presence=true` - fans that RPC out to every peer. It answers a whole-cluster question, and every remedy it points to (resync, transfer leadership) is already root-only. Checked *before* the fan-out runs. The cheap poll without the parameter is unchanged and still available to tenants.

**Also fixed while in the loop:** `GET /api/v1/cluster` listed reserved internal databases (`.raft`), unlike the presence matrix and the bootstrap-state RPC in the same module, which both skip them.

## Alternative considered

Making the whole cluster-info surface root-only, matching the mutating siblings, would be simpler and leaves no partial-disclosure surface. Rejected: it breaks remote-driver routing for every non-root application, and Studio's cluster tab is not root-gated in the UI. Scoping the database-derived content while leaving topology alone draws the line where the data actually changes owner.

## Tests

`ClusterInfoDatabaseScopingIT` (ha-raft, 2-node cluster) - 5 tests, 4 of which fail on `main`:

- tenant scoped to one database gets a `GET /api/v1/cluster` whose **entire body** contains neither the other database's name nor its type name (a whole-body substring assertion, so it covers `databases[]`, `alerts` and anything added later), while root still sees both
- same for `?mode=cluster`
- **counter-test:** a non-root user still gets `200` from `?mode=cluster` with `leaderAddress` and `replicaAddresses` present - guards the driver-routing path against a future "just make it root-only"
- `bootstrap-state` is `403` for a tenant and `200` for root
- `?presence=true` is `403` for a tenant while the plain poll stays `200`

`AbstractServerHttpHandlerDatabaseScopingTest` (server, 7 tests) pins the primitive itself: scoped grant, wildcard, no grant, a grant naming a database the server does not hold, iteration order, null user, and no aliasing of the caller's set.

Verified red before / green after; the pre-fix run reproduced all four leaks.

**Regression runs, all green:** `server` `http.handler` package (354), `server` `com.arcadedb.remote` package (48, driver), `server` security/user-management/HTTP set (129), ha-raft cluster tests (24) and ha-raft bootstrap/cluster-formation tests (19).

## Behavior change

A non-root caller's `GET /api/v1/cluster` and `GET /api/v1/server` responses now list only their own databases, and `bootstrap-state` / `?presence=true` require root. Release note added to `docs/release-26.8.1.md`.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -878,4 +878,26 @@ The underlying contract was loose in three places, and all three are tightened:
   `NumberFormatException`), an unknown `similarity` or `quantization` name, or an out-of-range `pqClusters` is a
   client mistake. They are reported as parsing errors now, the same treatment the `GEOSPATIAL` metadata gets.
 
+## Server and cluster status endpoints scope their per-database output to the caller
+
+The routes that enumerate the whole database registry rather than naming one database in the path now reduce
+every per-database entry they emit to the databases the caller is authorized for. This covers the `databases`
+array and the database-scoped `alerts` of `GET /api/v1/cluster`, the `ha.databases` array of
+`GET /api/v1/server?mode=cluster`, and the `metrics.sparseVectorIndexes` map of `GET /api/v1/server`. The
+server-level fields of those responses are unchanged and stay readable by any authenticated user - in
+particular `ha.leaderAddress` and `ha.replicaAddresses`, which the remote driver reads on every connection to
+route requests, and which therefore cannot be restricted to root.
+
+Two cluster endpoints move behind the root check that the seven mutating Raft endpoints already use:
+
+- **`POST /api/v1/cluster/bootstrap-state`** is a peer-to-peer RPC with no browser or driver consumer, and each
+  call computes a SHA-256 over every database directory on the node. Peers are unaffected: they reach it with
+  the cluster token forwarded as root.
+- **`GET /api/v1/cluster?presence=true`** fans that RPC out to every peer to build the presence matrix. The
+  matrix answers a whole-cluster question, and every remedy it points to (resync, transfer leadership) is
+  itself root-only. The cheap `GET /api/v1/cluster` poll without the parameter is unchanged.
+
+`GET /api/v1/cluster` also stops listing reserved internal databases such as the Raft control directory
+`.raft`, matching what the presence matrix and the bootstrap-state RPC already did.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -895,7 +895,11 @@ Two cluster endpoints move behind the root check that the seven mutating Raft en
   the cluster token forwarded as root.
 - **`GET /api/v1/cluster?presence=true`** fans that RPC out to every peer to build the presence matrix. The
   matrix answers a whole-cluster question, and every remedy it points to (resync, transfer leadership) is
-  itself root-only. The cheap `GET /api/v1/cluster` poll without the parameter is unchanged.
+  itself root-only. The cheap `GET /api/v1/cluster` poll without the parameter is unchanged. Note the root
+  check runs before the leader check, so a non-root caller passing the parameter to a **follower** now gets
+  `403` where it previously got a `200` with no matrix in it - tooling that polls followers with the flag
+  will see the change. In Studio the matrix is loaded by an explicit button, not by the cluster tab's
+  auto-poll, so a non-root operator's tab keeps working.
 
 `GET /api/v1/cluster` also stops listing reserved internal databases such as the Raft control directory
 `.raft`, matching what the presence matrix and the bootstrap-state RPC already did.
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ClusterAlerts.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ClusterAlerts.java
@@ -29,6 +29,7 @@ import com.arcadedb.server.monitor.HAReplicationStatsProvider.FollowerSample;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Computes cluster-level health alerts surfaced to operators (Studio HA panel, {@code GET /api/v1/cluster}).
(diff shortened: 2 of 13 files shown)


lvca:
Thanks - addressed the review.

**Test gap closed.** Added `ServerMetricsDatabaseScopingIT` in the `server` module (`metrics.sparseVectorIndexes` needs no HA, so it does not belong next to the ha-raft IT). It builds a real `LSM_SPARSE_VECTOR` index in the unauthorized database, asserts as **root** that the row actually exists, then asserts the tenant sees neither the key nor the name anywhere in the body. The root assertion is the precondition guard: that map is empty by default, so without it the tenant assertion would pass just as happily on a server with no sparse index at all. Mutation-checked - restoring the unscoped `getDatabaseNames()` in `buildSparseVectorIndexesJSON` fails the test.

**Minor 1 - `GetExistsDatabaseHandler` allocation: taken.** Agreed, building the whole authorized set to answer a single-name question allocates proportionally to the database count for a constant-time check. It now evaluates the same two conjuncts directly (`getDatabaseNames().contains(name) && user.canAccessToDatabase(name)`), with a comment explaining why this one route deliberately does not call the shared helper.

**Minor 3 - `checkRootUser` NPE: not taken, agreed with your reasoning.** These routes inherit `isRequireAuthentication() == true`, so `execute()` is unreachable with a null user. Hardening it here would also make this handler diverge from the seven mutating siblings that have the identical shape; if it is worth changing it is worth changing in `checkRootUser` itself, for all of them, which is a separate change.

**Minor 2 - `exportSettings` / event-log summary: deliberately out of scope.** Whether server settings and the event log are themselves tenant-sensitive is a different judgement call from "database X exists", and `GHSA-46hj` already masked the credential-bearing values via `isHidden()`. Folding it in would widen a security fix into a policy decision on the same push. Flagging it for a separate issue rather than silently changing behaviour here.

Also fixed the Codacy finding (`exportCluster` took an `HttpServerExchange` it never read - pre-existing, surfaced because the signature line changed).

Re-ran after the changes: `server` 384 tests (handler package + the new IT + user/group/token management), ha-raft 33 (cluster, bootstrap-state, HTTP/2). All green.

Note on `Meterian client scan`: it fails on this PR, but the last five runs on `main` failed too - the known uv-lockfile issue in the scanner container, not introduced here.

lvca:
Round 2 addressed - and observation 2 turned out to be load-bearing in a way neither of us spotted.

**Observation 2 - reserved-DB split: taken, and it exposed a vacuous test of mine.** `GetClusterHandler` now drops reserved names once, up front, so the `databases` array and the `alerts` payload built from the same set agree.

While wiring that I checked whether the reserved path was reachable at all, and found my existing `noneMatch(...startsWith("."))` assertion was **passing vacuously**: only the startup directory scan filters reserved names (`ArcadeDBServer:1046`), and the default fixture never puts one in the registry, so the assertion had nothing to reject. `createDatabase`/`registerDatabase` have no such guard, so a reserved name genuinely does reach the registry at runtime - which is presumably why `GetServerHandler.exportCluster` already carried the skip.

The IT now registers a reserved-named database at runtime holding a single-bucket type, and asserts it appears in neither payload, with a precondition that the `single-bucket-types` alert is actually firing (otherwise its silence about the reserved DB would prove nothing). Mutation-checked: dropping the `removeIf` fails the test. So the fix is real rather than defensive, and the assertion now bites.

**Observation 1 - null-user asymmetry: taken.** `GetExistsDatabaseHandler`'s inline predicate is now `user == null || user.canAccessToDatabase(requested)`, matching the contract the helper's Javadoc documents. You are right that the two paths disagreeing on the null contract is the kind of thing that drifts later even when neither is reachable today.

**Observation 3 - `LinkedHashSet` on the driver path: not taken, deliberately.** A fast path that returned the caller's collection when everything is authorized would make the return value sometimes-aliased, and the no-aliasing guarantee is pinned by a unit test precisely because a caller mutating a shared registry view is a much worse failure than one small set per status call. The cost is one set of N short strings per call with N in the single digits; the driver hits this once per connection, not per request. Keeping the contract simple is worth more here than the allocation.

Re-ran: ha-raft 33, server 364. All green.

lvca:
Third round - both "confirm" items checked rather than assumed, and both came back clean.

**Point 1 - Studio and `?presence=true`: confirmed safe, and documented.** `loadPresenceMatrix()` (`studio-cluster.js:407`) is wired to an explicit "Load presence matrix" button; the cluster tab's auto-poll (`updateCluster`, line 9) uses the bare `api/v1/cluster`. So a non-root operator's tab keeps working - only that one button now surfaces a 403 through `globalNotifyError`, which is an honest failure rather than a silently empty panel.

You're right that the follower case widened, though: the root check running before the `isLeader` check turns what used to be a silent `200`-with-no-matrix into a `403`. Added that to the release note, including the note about tooling that polls followers with the flag.

**Point 2 - other per-database maps in `exportMetrics`: confirmed, `sparseVectorIndexes` is the only one.** Walked all five sections: `profiler` is `Profiler.toJSON()`, which sums `getStats()` across databases into scalar totals and never emits a name; `meters` is keyed by `http.*` metric name; `executors` by pool tag; `events` is four integer counts. Only `sparseVectorIndexes` is keyed by database, and it is scoped.

**Point 3 - `removeIf` vs in-loop `continue`: agreed, not taken.** Unifying would mean touching `GetServerHandler` for cosmetics on a security PR. Both are correct; leaving the churn out.

Only change in this push is the release-note clause - no code touched, so the previously reported test runs still stand (ha-raft 33, server 364).

Issue #5602, opened 2026-07-30
[Misc] OpenCypher follow-ups from #5484: arity-guard blind spot, locale-dependent name folding, unimplemented registered functions


lvca:
Five small, independent follow-ups found while fixing #5484 (merged as 6e9d7f446, PR #5593). None is urgent on its own, so they are collected here rather than filed separately. They can be picked off in any order.

- [x] **1. Argument-count guard cannot reach the bridged functions.** `SQLFunction` exposes no bounds, so the drift guard asserts it compared *zero* functions and the seven `SQLFunctionBridge` functions — `distance` among them, where the real bug was — are uncoverable. Give `SQLFunction` an argument-count contract for the bridge to pass through.
- [x] **2. Locale-dependent case folding outside function names.** Seven sites in `CypherProcedureRegistry`, `VariableScopeTracker`, `CallStep` and `IsTypedExpression` still fold with the default locale, so a procedure or variable containing `I` misresolves under a Turkish default. Use `Locale.ROOT`.
- [x] **3. `charat`, `charlength`, `isnormalized` are registered but unimplemented.** The parser accepts them, then execution fails with `Unknown function`. Implement or unregister.
- [x] **4. Parse-time validation only walks `RETURN`/`WITH`.** A bad argument in `WHERE`, `UNWIND`, `SET` or `CREATE` is caught only at runtime. Pre-existing scope; widening it carries the same blast radius as item 1.
- [x] **5. Integer overflow in `abs()` is reported as 500.** Neo4j calls it a client error. Decide it for arithmetic overflow as a whole rather than patching one function.

Details for each below.

---

## 1. The argument-count guard cannot see the functions where the bug actually happened

`FunctionValidator`'s `minArgs`/`maxArgs` went from unused metadata to a hard parse-time gate in #5484, so a declaration narrower than what a function really accepts now rejects a query that used to work. `distance()` was exactly that: declared as taking exactly two arguments while it has always accepted an optional unit.

[...]

Issue #5464, opened 2026-07-27
`EXISTS` with relationship inline `WHERE` returns false even for a trivially true predicate


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`

Build:

```text
1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main
```

### Environment

- Host OS: Linux x86_64
- Deployment: Docker (`arcadedata/arcadedb:latest`)
- Runtime: OpenJDK
- Query interface: HTTP command API
- Query language: Cypher

### Description

An `EXISTS {}` expression containing a relationship inline `WHERE` predicate returns `false`, even when the predicate is trivially true. The failure affects both the pattern-only form and the explicit-`MATCH` form.

The same relationship pattern without the inline predicate correctly returns `true`:

```cypher
EXISTS { (a)-[:E]->(:A) }
```

However, adding a relationship variable and `WHERE 1=1` changes the result to `false`:

```cypher
EXISTS { (a)-[r:E WHERE 1=1]->(:A) }
```

Because `1=1` is true for every candidate relationship, both expressions should produce the same result.

This is related in behavior to #5109, but it exercises a different existence construct and relationship constraint:

- #5109: legacy `exists(pattern)` function with a relationship property map, producing a false positive.
- This issue: `EXISTS {}` subquery expression with a relationship inline `WHERE`, producing a false negative.

It may also be related to #3951/#3953, which concerned relationship inline predicates in regular `MATCH`. However, the same trivially true predicate works correctly in a regular `MATCH` on the tested build, so this reproducer exposes a context-dependent failure.

### Steps to reproduce

Execute each statement separately and in the listed order through the HTTP command API.

#### 1. Clear the database

```cypher
MATCH (n) DETACH DELETE n;
```

#### 2. Create test data

```cypher
CREATE (a:A {v:1}), (b:A {v:2});
```

```cypher
MATCH (a:A {v:1}), (b:A {v:2})
CREATE (a)-[:E {tag: "ok"}]->(b),
       (a)-[:E {tag: "bad"}]->(b);
```

#### 3. Compare plain existence with inline-`WHERE` existence

```cypher
MATCH (a:A {v:1})
RETURN
  EXISTS { (a)-[:E]->(:A) } AS plainExists,
  EXISTS { (a)-[r:E WHERE 1=1]->(:A) } AS inlineTrueExists;
```

Actual result:

```text
plainExists:           true
[...]

Issue #5461, opened 2026-07-27
UNWIND as the first clause yields zero rows in COUNT {} and returning CALL {} subqueries


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT` (build `1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main`)

### Environment
- Linux x86_64, Docker, OpenJDK
- `arcadedata/arcadedb:latest`
- HTTP command API, Cypher language

### Description

A subquery body produces no inner rows when `UNWIND` is its first clause. The issue affects both `COUNT {}` subquery expressions and returning `CALL {}` subqueries.

The minimal reproduction has no outer correlation:

```cypher
RETURN COUNT {
  UNWIND [1, 2] AS y
  RETURN y
} AS c;
```

Returns 0. Expected: 2 — the number of rows produced by unwinding the list.

The bug is specific to UNWIND appearing as the first clause in the subquery. A standalone `WITH` before UNWIND (even an uncorrelated one) restores correct behavior.

### Related issues

This appears distinct from #4015. That issue concerns a correlated `COUNT { MATCH ... }` subquery whose predicate reads an outer variable; its uncorrelated control works correctly. The reproduction here requires no outer-variable access and is triggered by `UNWIND` being the first clause.

It also appears distinct from #3944. That issue concerns a unit `CALL` subquery without `RETURN`, where inner `UNWIND` cardinality incorrectly multiplies outer rows. This report concerns returning subqueries and exhibits the opposite result-cardinality failure: all inner rows are lost.

### Steps to reproduce

```cypher
-- Minimal case: no outer correlation, COUNT {} returns 0
RETURN COUNT {
  UNWIND [1, 2] AS y
  RETURN y
} AS c;
-- Returns: 0
-- Expected: 2
```

### Expected behavior

`COUNT { UNWIND [1, 2] AS y RETURN y }` should return 2 — the number of rows produced by the subquery. Per Cypher semantics, `COUNT {}` evaluates the subquery and returns the count of rows it produces.

### Actual behavior

Returns 0. The subquery body produces zero observable inner rows when UNWIND is its first clause. This is consistent with the subquery failing to initialize its input row, but the internal cause is not established by the black-box result.

### Controls (work correctly)

```cypher
-- Uncorrelated WITH before UNWIND → correct
RETURN COUNT {
  WITH 1 AS dummy
  UNWIND [1, 2] AS y
  RETURN y
[...]

Issue #5165, opened 2026-07-09
`count{}` subquery miscounts with `WHERE IS NULL OR` predicate


shulei5831sl:
### ArcadeDB version

```text
arcadedata/arcadedb:latest
ArcadeDB version: 26.7.2-SNAPSHOT
```

### Environment

* Host OS: Linux
* Deployment: Docker
* Query interface: HTTP `/api/v1/command/test`
* Query language: Cypher
* Differential comparison: Neo4j 2026.05.0

### Description

A correlated `count { ... }` expression subquery returns incorrect counts when the subquery contains a `WHERE` predicate combining `IS NULL` with another predicate using `OR`.

Simple predicates are counted correctly, and `IS NULL` alone is also counted correctly. The wrong result appears when the predicates are combined as:

```cypher
WHERE f.active IS NULL OR f.active = true
```

### Steps to reproduce

Run from a clean database.

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (u1:User {id:1, active:true}),
  (u2:User {id:2, active:true}),
  (u3:User {id:3, active:true}),
  (u4:User {id:4, active:false}),
  (u5:User {id:5, active:false}),
  (u6:User {id:6, active:false}),
  (u1)-[:FRIEND]->(u4),
  (u1)-[:FRIEND]->(u5),
  (u2)-[:FRIEND]->(u4),
  (u2)-[:FRIEND]->(u1),
  (u3)-[:FRIEND]->(u5),
  (u3)-[:FRIEND]->(u6);
```

Data summary:

```text
u1 -> u4(false), u5(false)
u2 -> u4(false), u1(true)
u3 -> u5(false), u6(false)
```

No matched friend has `active IS NULL`. Only `u2` has a friend with `active = true`.

### Failing query

```cypher
MATCH (u:User {active:true})
RETURN u.id,
  count {
    MATCH (u)-[:FRIEND]->(f:User)
    WHERE f.active IS NULL OR f.active = true
  } AS cnt
ORDER BY u.id;
```

### Expected behavior

Only `u2` should have one matching friend.

```text
u.id | cnt
-----+----
   1 |  0
   2 |  1
   3 |  0
```

Neo4j 2026.05.0 returns:

```text
1 | 0
2 | 1
3 | 0
```

### Actual behavior

ArcadeDB returns:

```text
u.id | cnt
-----+----
   1 |  1
   2 |  1
   3 |  1
```

[...]

PR #5642, opened 2026-07-31
fix(engine) #5626: walk the subquery bodies the parse-time checks claimed were validated elsewhere

### LINKED ISSUE #5626: [Cypher] EXISTS/COUNT/COLLECT subquery bodies escape parse-time argument validation
Found while closing #5602 (PR #5612), and it contradicts something that PR's own Javadoc asserts.

`CypherExpressionWalker` stops at a subquery held as unparsed text - `EXISTS { ... }`, `COUNT { ... }`, `COLLECT { ... }` - and its class Javadoc explains why: *"those are parsed on their own and validated then, and there is no AST here to walk."* The first half is true; the second half is not. Nothing validates them at parse time.

Verified against the merged code, using `EXPLAIN` so the query is parsed and planned but never executed:

```
NOT-VALIDATED  EXPLAIN MATCH (n:P) WHERE EXISTS { MATCH (m:P) WHERE abs('x') > 0 RETURN m } RETURN n
NOT-VALIDATED  EXPLAIN MATCH (n:P) RETURN COUNT { MATCH (m:P) WHERE abs('x') > 0 RETURN m } AS r
VALIDATED[CommandSemanticException]  EXPLAIN MATCH (n:P) WHERE abs('x') > 0 RETURN n
```

The identical `abs('x')` is rejected before the query runs in the outer `WHERE` and let through inside a subquery. That is the same clause-dependent asymmetry item 4 of #5602 set out to remove, in the one place the walker cannot reach.

The runtime check still catches it whenever the subquery actually executes, so nothing is silently wrong - but a subquery over a pattern that matches no row will never run it, which is exactly the case #5484 and #5602 argued should not look like success.

**Suggested fix.** `ExistsExpression`, `CollectExpression` and `CountExpression` keep their body as a `String` (`getSubquery()`). Either parse it during validation and walk the resulting statement, or - better, since the body is parsed at execution anyway - hold the parsed statement on the expression at build time and have the walker descend into it. The second removes the re-parse as well.

Whichever way it goes, fix the `CypherExpressionWalker` class Javadoc: it currently tells the next maintainer these are covered elsewhere, and they are not.

lvca:
Closes #5626.

`CypherExpressionWalker` stopped at a subquery held as text, and its class Javadoc explained why: *"those are parsed on their own and validated then, and there is no AST here to walk."* The issue is right that the second half is false. It turns out a fourth body was escaping too, for a different reason, and two whole predicates were escaping for a third.

### What actually escaped

| shape | why the walk did not reach it |
|---|---|
| `EXISTS { }` / `COUNT { }` / `COLLECT { }` | body kept as text, no AST to descend into |
| `CALL { }` | `walkClause`'s `default` arm skipped it, on the same "validated when it is parsed" belief |
| `WHERE EXISTS { ... }` | the builder wrapped it in an **anonymous** `BooleanExpression`, a leaf to the walk |
| `WHERE isEmpty(x)` - a function call as a bare predicate | same: a second anonymous adapter |

The last two are the trap this class's own Javadoc warns about (*"the `default` arms treat an unrecognised type as a leaf ... with nothing failing to say so"*), and they meant the reproducer in the issue was under-stating the reach: the first of the three forms was invisible even before the body question.

### The fix

**The three expressions now carry their body as a `CypherStatement` alongside the text**, and the walk descends into it - as it now does into a `CALL { }` body.

The issue's second option, taken, with one correction to its rationale: **this does not remove the re-parse.** What executes is not the body as written but the body after `CorrelatedSubqueryRewriter` has pinned the outer row into it, which differs row by row, so the text has to stay and the runtime parse with it. What the AST costs is therefore not free - but it is not a second lex either: it is built from the parse tree ANTLR already produced for the body (`ctx.queryWithLocalDefinitions()`, or the `patternList`/`whereClause` of a bare-pattern body), so it is one tree walk at parse time, once per distinct query text.

**The two anonymous adapters are gone.** Both were byte-for-byte `BooleanCoercionExpression` - same `evaluate`, same `evaluateTernary`, same `getText` - so they are now that named node, which the walk already knew. The function-call one turned out to be wholly redundant with the generic arm three lines below it (both call `parseExpressionFromText(expr6)` and wrap the result), and its removal also fixes a latent NPE: it wrapped a possibly-null expression, where the generic arm falls back properly.

**A body is a scope of its own.** Reusing the enclosing `varTypes` map for it would have been wrong in both directions, so the visitor re-binds itself at the boundary (`Visitor.forNestedStatement`): the kinds the body declares, over the ones it inherits, with any name the body binds by any means dropped from what it inherits. That is what keeps

```cypher
MATCH p = (a:P)-[:KNOWS]->(b:P) CALL { MATCH (p:P) RETURN p.name AS n } RETURN n
```

legal - an implicit `CALL { }` imports nothing, so the body's `p` is its own node and `p.name` is a plain property read, not property access on the outer path.

Two more expression positions the same traversal now reaches: procedure `CALL` arguments and the `LOAD CSV FROM` url.

### Behaviour change

Same shape as #5602's: a bad call inside a subquery body is now rejected before the query starts instead of failing at runtime - or, where the subquery matched no row, instead of quietly answering `false` / `0` / `[]`. No check is new; only its reach.

One existing test changed. `Issue5228CallMatchEdgeLabelEntityTest` probed `CALL () { MATCH (b:EdgeLabel) RETURN labels(b), type(b) }` and asserted zero rows; `type(b)` on a node is the type error it already was outside a subquery, so that half of the probe now asserts the rejection - and asserts it identically inside and outside, which is the point of the issue.

### Verified against Neo4j

Neo4j runs semantic analysis over the whole query including subquery bodies: `abs('x')` is `Type mismatch: expected Float or Integer but was String` at compile time wherever it is written, and `type(node)` is `Type mismatch: expected Relationship but was Node`. Neither is a Cypher usage mistake on the reporter's side.

### Tests

`CypherSubqueryParseTimeValidationIssue5626Test` - 22 tests, all under `EXPLAIN` so the query is parsed and planned but never executed:

- the reproducer for each of the four body kinds, each asserting the *same* class and message as the identical call in the outer `WHERE`
- every check of the phase reaching inside, not only the type one: unknown name, wrong arity, property access on a path variable
- the positions a body can be written in: bare pattern, negated, inside a conjunction, in a `WITH ... WHERE`, nested one subquery inside another, a `UNION` branch of a `CALL` body
- procedure `CALL` arguments and `LOAD CSV FROM`
- no false rejection: a correlated body, variable kinds read from the body that declares them, the unimported-shadowing case above, and the converse (an outer path variable is still a path in a body that does not re-declare it)
- `EXISTS` / `COUNT` / `COLLECT` still evaluate to the same answers after the adapter swap

Green: engine `query.**` + `function.**` 11311/0, `query.opencypher.**` 7808/0, bolt 93/0, the server Cypher HTTP ITs 20/0.

### Follow-up worth filing separately

The per-row re-parse is still there, and is a much larger cost than anything here: `EXISTS { }` lexes, parses, validates and plans its body once for every outer row. Now that the body exists as an AST, correlating it as an AST - binding the outer row into parameters rather than splicing text - would remove that, and would also retire `CorrelatedSubqueryRewriter`'s text scanning, which is where #4995, #5165, #5461 and #5464 all came from.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -933,9 +933,19 @@ Two expression positions that were leaves for the same reason are covered too: a
 `BooleanCoercionExpression` - byte-for-byte the same behaviour, minus the blind spot. Procedure `CALL` arguments
 and the `LOAD CSV FROM` url expression are walked as well.
 
+Working out what a name is - a node, a relationship, a path - was written twice, once for a statement and once
+for a subquery body, and the two had already drifted. They are one construction now, which closed an asymmetry
+older than this issue: the statement's copy dropped every kind at a `WITH *`, because it kept only what the
+projection names, while the body's copy passed the incoming scope through.
+
 > **Potentially breaking, in the same way #5602 was.** A query whose bad call sits inside a subquery body is now
 > rejected before it starts rather than failing at runtime - or, where the subquery matched no row, rather than
 > quietly answering `false` / `0` / `[]`. The call was always wrong. One shape worth calling out: `type(b)` where
 > `b` is a node, written inside a subquery, is now the type error it already was outside one.
+>
+> A second shape comes from the shared scope construction: a kind now survives `WITH *`, so
+> `MATCH p = (a)-[:KNOWS]->(b) WITH * RETURN p.name` is rejected as the path-property access it always was,
+> where before the `WITH *` made the engine forget `p` was a path and the query failed at runtime instead. A
+> projection that names what it keeps is unaffected - `WITH 1 AS p` still stops `p` being a path.
 
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -922,7 +922,10 @@ produced for it rather than by re-lexing, and the traversal descends into it - a
 Crossing into a body changes the variable scope, so a check that reads variable kinds (`type()` wants a
 relationship, `p.name` needs `p` not to be a path) re-binds itself to the kinds the body declares, over the ones it
 inherits; an implicit `CALL { }` imports nothing, so a name the body binds for itself shadows the outer one rather
-than answering for it.
+than answering for it. A body's kinds are built by walking its clauses in order, so the two spellings of the same
+import - `CALL (p) { ... }` and a leading `WITH p` - answer alike, while `WITH 1 AS p` stops `p` being a path. Each
+branch of a `UNION` is a scope in its own right: a variable only one branch declares is checked against the kind
+that branch gives it, instead of being dropped for the branches disagreeing about it.
 
 Two expression positions that were leaves for the same reason are covered too: an `EXISTS { }` written as a bare
 `WHERE` predicate, and a function call used as one (`WHERE isEmpty(x)`), each used to get an anonymous
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -878,4 +878,35 @@ The underlying contract was loose in three places, and all three are tightened:
   `NumberFormatException`), an unknown `similarity` or `quantization` name, or an out-of-range `pqClusters` is a
   client mistake. They are reported as parsing errors now, the same treatment the `GEOSPATIAL` metadata gets.
 
+## Cypher: a subquery body is validated like the query around it
+
+The widening of #5602 left one place the walk could not reach, and the class it was written on said so in the
+opposite direction: the bodies of `EXISTS { }`, `COUNT { }` and `COLLECT { }` were "parsed on their own and
+validated then". They were not. Each of the three keeps its body as text and re-parses it once per outer row, and
+a body that cannot run is absorbed into the expression's neutral value - `false`, `0`, an empty list. A `CALL { }`
+body was never handed to this phase at all ([#5626](https://github.com/ArcadeData/arcadedb/issues/5626)).
+
+So `MATCH (n:P) WHERE abs('x') > 0 RETURN n` was rejected before it started, while the identical call one level in,
+`MATCH (n:P) WHERE EXISTS { MATCH (m:P) WHERE abs('x') > 0 RETURN m } RETURN n`, was accepted - the very
+clause-dependent asymmetry #5602 set out to remove. Neo4j type-checks a subquery body exactly as it does the query
+around it.
+
+The three expressions now carry their body as an AST alongside the text, built from the parse tree ANTLR already
+produced for it rather than by re-lexing, and the traversal descends into it - as it does into a `CALL { }` body.
+Crossing into a body changes the variable scope, so a check that reads variable kinds (`type()` wants a
+relationship, `p.name` needs `p` not to be a path) re-binds itself to the kinds the body declares, over the ones it
+inherits; an implicit `CALL { }` imports nothing, so a name the body binds for itself shadows the outer one rather
+than answering for it.
+
+Two expression positions that were leaves for the same reason are covered too: an `EXISTS { }` written as a bare
+`WHERE` predicate, and a function call used as one (`WHERE isEmpty(x)`), each used to get an anonymous
+`BooleanExpression` adapter that the traversal could only treat as a leaf. Both are now the ordinary
+`BooleanCoercionExpression` - byte-for-byte the same behaviour, minus the blind spot. Procedure `CALL` arguments
+and the `LOAD CSV FROM` url expression are walked as well.
+
+> **Potentially breaking, in the same way #5602 was.** A query whose bad call sits inside a subquery body is now
+> rejected before it starts rather than failing at runtime - or, where the subquery matched no row, rather than
+> quietly answering `false` / `0` / `[]`. The call was always wrong. One shape worth calling out: `type(b)` where
+> `b` is a node, written inside a subquery, is now the type error it already was outside one.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/CollectExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/CollectExpression.java
@@ -42,12 +42,14 @@ import java.util.logging.Level;
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
 public class CollectExpression implements Expression {
-  private final String subquery;
-  private final String text;
+  private final String          subquery;
+  private final String          text;
+  private final CypherStatement parsedSubquery;
 
-  public CollectExpression(final String subquery, final String text) {
+  public CollectExpression(final String subquery, final String text, final CypherStatement parsedSubquery) {
     this.subquery = subquery;
     this.text = text;
+    this.parsedSubquery = parsedSubquery;
   }
 
   @Override
@@ -98,5 +100,14 @@ public class CollectExpression implements Expression {
 
   public String getSubquery() {
     return subquery;
+  }
+
+  /**
+   * The body as an AST. The body still executes from {@link #getSubquery()}, because what runs is the text after
+   * {@link CorrelatedSubqueryRewriter} has correlated it to the outer row, which differs row by row. This AST is the
+   * body as written, and exists so that the parse-time checks reach inside it (issue #5626).
+   */
+  public CypherStatement getParsedSubquery() {
+    return parsedSubquery;
   }
 }
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/CountExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/CountExpression.java
@@ -40,12 +40,14 @@ import java.util.logging.Level;
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
 public class CountExpression implements Expression {
-  private final String subquery;
-  private final String text;
+  private final String          subquery;
+  private final String          text;
+  private final CypherStatement parsedSubquery;
 
-  public CountExpression(final String subquery, final String text) {
+  public CountExpression(final String subquery, final String text, final CypherStatement parsedSubquery) {
     this.subquery = subquery;
     this.text = text;
+    this.parsedSubquery = parsedSubquery;
   }
 
   @Override
@@ -98,5 +100,14 @@ public class CountExpression implements Expression {
 
   public String getSubquery() {
     return subquery;
+  }
+
+  /**
+   * The body as an AST. The body still executes from {@link #getSubquery()}, because what runs is the text after
+   * {@link CorrelatedSubqueryRewriter} has correlated it to the outer row, which differs row by row. This AST is the
+   * body as written, and exists so that the parse-time checks reach inside it (issue #5626).
+   */
+  public CypherStatement getParsedSubquery() {
+    return parsedSubquery;
   }
 }
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/ExistsExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/ExistsExpression.java
@@ -34,12 +34,23 @@ import java.util.logging.Level;
  * Returns true if the pattern/subquery has at least one match, false otherwise.
  */
 public class ExistsExpression implements Expression {
-  private final String subquery;
-  private final String text;
+  private final String          subquery;
+  private final String          text;
+  private final CypherStatement parsedSubquery;
 
+  /**
+   * Used by the runtime paths that synthesize an existential subquery from a pattern already held as an AST
+   * ({@link PatternPredicateExpression}). Such an expression is built while the query runs, never handed to the
+   * parse-time validation, so it carries no parsed body.
+   */
   public ExistsExpression(final String subquery, final String text) {
+    this(subquery, text, null);
+  }
+
+  public ExistsExpression(final String subquery, final String text, final CypherStatement parsedSubquery) {
     this.subquery = subquery;
     this.text = text;
+    this.parsedSubquery = parsedSubquery;
   }
 
   @Override
@@ -84,5 +95,16 @@ public class ExistsExpression implements Expression {
 
   public String getSubquery() {
     return subquery;
+  }
+
+  /**
+   * The body as an AST, or {@code null} when this expression was synthesized at runtime rather than parsed.
+   * <p>
+   * The body still executes from {@link #getSubquery()}, because what runs is the text after
+   * {@link CorrelatedSubqueryRewriter} has correlated it to the outer row, which differs row by row. This AST is the
+   * body as written, and exists so that the parse-time checks reach inside it (issue #5626).
+   */
+  public CypherStatement getParsedSubquery() {
+    return parsedSubquery;
   }
 }
--- a/engine/src/main/java/com/arcadedb/query/opencypher/parser/CypherASTBuilder.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/parser/CypherASTBuilder.java
@@ -32,7 +32,6 @@ import com.arcadedb.query.opencypher.ast.CypherStatement;
 import com.arcadedb.query.opencypher.ast.CypherTransactionStatement;
 import com.arcadedb.query.opencypher.ast.DeleteClause;
 import com.arcadedb.query.opencypher.ast.Direction;
-import com.arcadedb.query.opencypher.ast.ExistsExpression;
 import com.arcadedb.query.opencypher.ast.Expression;
 import com.arcadedb.query.opencypher.ast.FinishClause;
 import com.arcadedb.query.opencypher.ast.ForeachClause;
@@ -76,8 +75,6 @@ import com.arcadedb.query.opencypher.rewriter.ExpressionRewriter;
 import com.arcadedb.query.opencypher.rewriter.InlineNodeWhereHoister;
 import com.arcadedb.query.opencypher.rewriter.LabelPredicateHoister;
 import com.arcadedb.query.opencypher.rewriter.ProjectedOrderByNormalizer;
-import com.arcadedb.query.sql.executor.CommandContext;
-import com.arcadedb.query.sql.executor.Result;
 import org.antlr.v4.runtime.ParserRuleContext;
 import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.tree.ParseTree;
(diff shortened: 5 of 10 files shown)


lvca:
Thanks - the scope question was the right thing to pull on. Probing it turned up a genuine bug next door to it, so the whole `nestedVarTypes` construction is rewritten in 0429851.

**Your case: confirmed harmless, and now asserted.** `MATCH p = (a:P)-[:KNOWS]->(b:P) CALL { RETURN p.name } RETURN *` answers `UndefinedVariable: Variable 'p' not defined`, not the path-access error. `validateVariableScope` is phase 3 and `validateFunctionArgumentTypes` is phase 12, so the accurate diagnostic is raised long before this phase reads a kind for a name the body cannot see. Locked down by `anUnimportedReferenceIsReportedAsUndefinedNotAsAKindError`, and the reasoning is now in the method Javadoc rather than only in a PR description.

**What the probe found instead.** The old rule was "declarations, minus every name the body binds by any means". An importing `WITH p` re-projects the outer variable under the same name - it binds it, so the kind was dropped, so `p` stopped being a path:

```cypher
MATCH p = (a:P)-[:KNOWS]->(b:P) CALL { WITH p RETURN p.name AS n } RETURN n   -- accepted (wrong)
MATCH p = (a:P)-[:KNOWS]->(b:P) CALL (p) { RETURN p.name AS n } RETURN n      -- rejected (right)
```

Two spellings of the same import, two answers - the asymmetry this PR exists to remove, reintroduced one level down.

**The rewrite.** The body's clauses are walked in order, the way `validateVariableTypes` walks a statement's: a pattern declares a kind, a `WITH` resets the scope to what it projects and carries a kind through only where the item is a plain variable reference, `WITH *` passes the incoming scope through, and a binding that says nothing about the kind (`UNWIND` variable, `YIELD` output, `LOAD CSV` variable) drops whatever the name held. So `WITH p` and `WITH p AS q` keep the path, `WITH 1 AS p` stops being one. One pass replaces `collectNestedDeclarations` + `collectPatternVarTypes` + `declareNested`, which also retires both nits: the `getClausesInOrder()` guard is back in line with `walk()`, and the `YIELD` arm no longer needs the missing null check.

Six cases added across three tests: the import spellings (explicit scope, leading `WITH`, renamed), the computed re-bind, the renamed pass-through for `labels()`/`type()`, `WITH *`, and your unimported case.

**On the WITH double-walk** - agreed it is real and idempotent. Leaving it: collapsing it needs identity tracking across two collections that some builder paths populate and others do not, which is more moving parts than the duplicate work it saves at parse time.

Green after the change: engine `query.**` + `function.**` 11314/0, `query.opencypher.**` 7811/0.

lvca:
Both points fixed rather than documented, in e9c26cb - each turned out to be a few lines.

**The UNION intersection is gone.** Your example is exactly right, and it was reachable: `CALL { MATCH (a:P) RETURN a.name AS x UNION MATCH q=(m:P)-->() RETURN q.name AS x }` let the path-property read through. Rather than document the tradeoff, each branch is now entered as a nested statement in its own right - `walkNested(query, visitor)` instead of `walk(query, visitor)` - so `forNestedStatement` fires per branch and each builds its own scope over what it inherits. `nestedVarTypes` then has nothing to merge and the intersection code is deleted.

One trap worth recording, since it is why the union arm could not simply fall through to the generic path: `UnionStatement.getClausesInOrder()` answers with its **first branch's** clauses. A `nestedVarTypes` that walked them would have handed every branch a scope built from branch 1 - which happens to produce the right answer for the example above, so the test asserts it from **both** branch positions.

**The double WITH walk is gone too.** You are right that "documented footgun" is the wrong resting place for a class that is now a general-purpose utility - a visitor that keeps state is a reasonable thing for the next caller to write, and it would have silently double-counted. Which WITHs the ordered walk reached is tracked by identity, so the second pass adds only what it genuinely missed. The `WITH ... WHERE` subquery test confirms the dedup did not cost coverage.

**The explicit `CALL () { }` assertion is added**, alongside the implicit one - both answer `UndefinedVariable`.

Four cases added (both union orderings, a clean union that must still pass, the explicit empty scope). Green: engine `query.**` + `function.**` 11315/0, `query.opencypher.**` 7812/0.

On the per-row re-parse: agreed it is the real cost and out of scope here. I will file it as a follow-up - now that the body exists as an AST, correlating it as an AST instead of splicing text would also retire `CorrelatedSubqueryRewriter`'s text scanning, which is where #4995, #5165, #5461 and #5464 all came from.

lvca:
All three observations addressed in 0e51705 - the first two turned into something enforced rather than documented.

**1. The end-state scope.** Right, and deliberately so: matching the top-level approximation is the point. `varTypes` is one map for the whole statement, and answering a body *more* precisely than the query around it would put back a clause-dependent asymmetry of exactly the kind #5602 and this issue exist to remove - the same query, checked differently depending on where it was written.

Worth adding to your direction analysis, because it makes the approximation safe rather than merely conservative: it can only err one way, and not by luck. A false rejection would need a name to be read at a position where its kind is K, then rebound to `PATH` later in the body. Cypher does not let a bound name be rebound by a pattern (`Variable 'x' already declared`), and the one rebinding it does allow - a `WITH` projection - can drop a kind but never turn a name into a path. So the end state is never *more* path-ish than any position preceding it. Missed check, never invented one, structurally. That reasoning is now in the `nestedVarTypes` Javadoc rather than only in a review thread.

**2. "Visited exactly once."** Fair to call out an advertised guarantee nothing proved. It is proved now: `everyExpressionIsVisitedExactlyOnce` walks a statement with a counting visitor and asserts no expression identity is reached twice. It asserts its own precondition first - that the query's `WITH` really is registered on both `getClausesInOrder()` and `getWithClauses()` - so it cannot pass because there was nothing to double-visit. Removing the `IdentityHashMap` dedup fails it on eight expressions:

```
Expecting empty but was: ["0", "n.age", "abs(n.age)", "0", "m", "n", "a", "m.age"]
```

`m` and `m.age` are from the `CALL { }` body, so the double-registration reaches into nested scopes too.

The contract stays narrow in the wording: guaranteed for the two registration paths this traversal reconciles; elsewhere once-only falls out of each arm walking a part of the tree no other arm does - a property to preserve when adding an arm, and now called out as such at the arm list rather than implied by the interface.

**3. FOREACH.** Intentional, and the asymmetry with `walkClause` is the correct one: what the inner `CREATE`/`MERGE` bind lives and dies inside the loop and is not in scope after it, so declaring those kinds in the body scope would be declaring them where they cannot be referenced. The walk does still descend into the inner clauses and check their expressions against the scope the loop was entered with. Noted at the arm.

Green after the change: engine `query.opencypher.**` 7813/0 (98 skipped), and the earlier full `query.**` + `function.**` runs at 10243/0 and 1072/0.

lvca:
All four notes addressed in 5b37491. The third one turned out to be pointing at code that should not exist.

**3. The double-`WITH` identity dedup - deleted, along with the pass it was guarding.** You framed it as a latent assumption; checking the assumption showed the second pass was unreachable. `StatementBuilder.addWith` is the *only* place a `WITH` is registered, and it puts the same object on both `getWithClauses()` and `getClausesInOrder()`; `StatementBuilder.build()` is the only construction site. So the second loop could never reach a clause the first missed - it existed only to be deduplicated against.

Confirmed rather than argued: I replaced the second pass with a `throw` on any statement-only `WITH` and ran `query.opencypher.**`. 7813 tests, zero hits.

So the pass and the `IdentityHashMap` are both gone, and the walk is the ordered clause list alone. Visiting once is structural again - each arm covers a part of the tree no other arm does - rather than a property restored by a set allocated per statement walk, and the question of what happens if the two collections ever hold different instances goes away instead of being documented. The remaining direction is the safe one: a builder path that registered a `WITH` on the statement alone would make the walk skip it, a missed check, never a double-count.

`everyExpressionIsVisitedExactlyOnce` now pins the invariant the single pass rests on - every `WITH` on the statement is the same instance on the ordered list - so a future builder path that breaks it fails a test instead of quietly costing coverage.

**1. "One tree walk" understated it.** Correct, and fixed: `visitQueryWithLocalDefinitions` runs the full `CypherASTBuilder` pipeline, rewriters included. The Javadoc now says a full statement is built over an existing sub-tree, and that a reader sizing it should read it as "one more statement built", not "one traversal". What the sentence was defending - that there is no second lex or parse - is kept, since that was the alternative it was chosen over.

**2. The `COLLECT` null/null asymmetry.** One line at the call site: a `COLLECT` body has no bare-pattern form in the grammar, it has to `RETURN` the value being collected, so it is always a `queryWithLocalDefinitions` and there is no pattern/where context to pass.

**4. Whole-body scope.** Agreed, and it stays - matching the top-level approximation is the point, since answering a body more precisely than the query around it would put back a clause-dependent asymmetry. Worth adding that it is safe by construction, not by luck: a false rejection would need a name to be read at a position where its kind is K and rebound to `PATH` later in the body, and Cypher does not let a bound name be rebound by a pattern, while the one rebinding it does allow - a `WITH` projection - can drop a kind but never turn a name into a path. The end state is therefore never more path-ish than any position before it. That reasoning is in the `nestedVarTypes` Javadoc as of 0e51705.

Green: `query.opencypher.**` 7813/0 (98 skipped), `query.sql.**` 2315/0 (8 skipped), and the earlier `query.**` + `function.**` at 10243/0 and 1072/0.

lvca:
Both notes acted on in 2ec34f2. The first one was the more important of the two.

**1. The parse-time AST build as a new failure surface.** You are right that this is the one behavioural surface beyond "rejected earlier", and it is the wrong direction to leave to luck: a construct ANTLR accepts but `CypherASTBuilder` cannot assemble would turn a query that used to run into a parse failure. Widening what gets *rejected* is what this issue is for; breaking what used to *run* is not.

So the build is now best-effort. On failure the body stays text only, the expression keeps a null AST, the walk treats it as a leaf exactly as before #5626, and execution re-parses the text as it always has - the pre-issue behaviour, restored precisely for the case that would otherwise regress. Logged at `FINE`, because a body with no AST is a body the parse-time checks do not see, and that should stay discoverable rather than become permanent.

`aSubqueryExpressionWithNoBodyIsALeafRatherThanAFailure` pins what the fallback rests on, and it took two attempts to make it mean anything. The first version used a plain lambda visitor and passed even with the null guard in `walkNested` deleted - three layers of independent null-tolerance made it vacuous. The guarantee is not "the walk survives a null body" but "the boundary hook is never called with one", and only a visitor that *dereferences* the statement can tell those apart, which is what the validator's own `forNestedStatement` does when it builds the inner scope. With the visitor reading `getClausesInOrder()`, removing the guard fails it:

```
java.lang.NullPointerException: Cannot invoke "CypherStatement.getClausesInOrder()" because "nested" is null
```

**2. `COLLECT` and the bare-pattern arm.** Confirmed from the grammar rather than assumed, and the rule is now quoted at the call site so the next reader does not have to go looking:

```
collectExpression : COLLECT LCURLY queryWithLocalDefinitions RCURLY
existsExpression  : EXISTS  LCURLY (queryWithLocalDefinitions | matchMode? patternList whereClause?) RCURLY
countExpression   : COUNT   LCURLY (queryWithLocalDefinitions | matchMode? patternList whereClause?) RCURLY
```

One alternative where the other two have two, because a `COLLECT` body has to `RETURN` the value being collected. So `queryWithLocalDefinitions()` is never null there and the blind spot you describe is unreachable - which, as you say, is exactly the class of thing this PR closes, so it is worth having it stated rather than inferred.

**3. The follow-up.** Agreed, and it is the reason the AST is attached to the expression rather than built and discarded inside the validator: correlating the body as an AST retires both the per-row re-parse and the text scanning in `CorrelatedSubqueryRewriter`. Out of scope here, tracked separately.

Green: `query.opencypher.**` 7814/0 (98 skipped). No conflict with `main`.

lvca:
Two rounds of feedback landed together here, so this covers both: the maintainability point from the earlier review and the three observations in the latest. Pushed as 669f469 and 0b108de.

**Duplicated scope construction - unified, and it was hiding something.** You were right that two copies of "walk the clauses in order and work out what each name is" would drift, and right that they already had. Rather than a `// keep in sync` note, they are now one `buildVarTypes` with the two genuine differences as parameters: the scope it starts from (empty for a statement, the enclosing one for a body), and whether a kind clash raises `VariableTypeConflict` (it must at the top level; it must not inside a body, where a name the body binds for itself legitimately shadows an outer one).

Folding them together closed an asymmetry **older than this issue**. The statement's copy dropped every kind at a `WITH *`, because it kept only what the projection names; the body's copy passed the incoming scope through. So:

```cypher
MATCH p = (a:P)-[:KNOWS]->(b:P) WITH * RETURN p.name                        -- accepted (wrong)
MATCH (n:P) RETURN COLLECT { MATCH p = (a:P)-->(b:P) WITH * RETURN p.name } -- rejected (right)
```

The same access, two answers, decided by whether it was written inside a body - the shape #5602 and #5626 exist to remove, sitting in the phase that decides what a name *is*. Both reject it now. `aKindSurvivesWithStarInsideAndOutsideASubqueryBody` pins it from both positions; making the star arm unconditional again fails it. `WITH 1 AS p` still stops `p` being a path, so a projection that names what it keeps is unaffected. Added to the release notes as a second potentially-breaking shape.

The statement side also picks up the kindless-binding arms the body side had - `UNWIND`, `LOAD CSV`, `YIELD`, a nested subquery's output drop whatever the name held. That direction can only lose a kind, so it narrows what is checked rather than widening what is rejected.

**1. The broad `catch (Exception)` - split.** Good point: the fallback was putting two different things in the same quiet place. A `CommandParsingException` is the case the fallback exists for, the builder saying it does not support this body, and it is not news - still `FINE`. Anything else is a defect whose cost is invisible, since the query still runs and only the checks stop seeing inside that body - now `WARNING`. That is the difference between a degraded body being discoverable in an ordinary log and being discoverable in theory.

**2. Parse-time cost.** Agreed, and no further change: the Javadoc says a full statement build including rewriters, and the follow-up that pays it back (correlating the body as an AST to retire the per-row re-parse) is exactly why the AST is attached to the expression rather than built and discarded inside the validator.

**3. Import order.** Fixed - `com.arcadedb.log.LogManager` now sits after `com.arcadedb.exception.*`.

**One suggestion I am not taking**, from the earlier review: nulling `parsedSubquery` after validation to drop the steady-state footprint. Cached plans are shared across threads, so nulling a field on one is a write to a structure other threads are reading concurrently - a data race in exchange for a bounded saving. It also works directly against the follow-up above, which needs that AST at execution time. The retained duplication is real; this is not the way to remove it.

Green: `query.opencypher.**` 7815/0 (98 skipped), `query.**` + `function.**` 11318/0 (106 skipped). The full `mvn -pl engine verify` you asked for is running now and I will post the result here when it finishes. No conflict with `main`.

lvca:
Point 1 was a real gap and it is fixed rather than tracked - 9284669. Points 2 to 4 and the nits below.

**1. Intra-body kind clashes - fixed, by removing the flag rather than refining it.** You put your finger on exactly the wrong shape: one boolean was answering two different questions. `MATCH (x:P) MATCH ()-[x]->()` raised `VariableTypeConflict` written in the query and passed written in a body - the asymmetry this issue exists to remove, one level down from where it was found, introduced by the unification two commits earlier.

Shadowing an *inherited* name and clashing with one *declared here* are different questions, so the build now tracks which names the scope declared itself, and a clash is a clash only between two of those:

```cypher
MATCH (n:P) WHERE EXISTS { MATCH (x:P) MATCH ()-[x]->() RETURN x } RETURN n   -- VariableTypeConflict, as outside
MATCH p = (a:P)-[:KNOWS]->(b:P) CALL { MATCH (p:P) RETURN p.name AS n } RETURN n  -- shadowing, still fine
```

A statement inherits nothing, so every clash there is one and top-level behaviour is unchanged. A `WITH`'s projection counts as declaring what it projects, so past `WITH p` the body has restated `p` and a later clash with it is real - which also keeps the top-level semantics after a projection exactly as they were.

`aKindClashInsideABodyIsAConflictWhileShadowingAnOuterNameIsNot` asserts both directions, because a check that got only one right would either miss the conflict or reject the shadow. Both mutations fail it: never raising loses the conflict, always raising rejects the shadow with `Variable 'p' already defined as PATH, cannot redefine as NODE`.

**2. The single-collection WITH walk rests on a builder invariant.** Agreed on the shape of the risk, and I am leaving it as is. A `StatementBuilder` assertion would fire only for a path someone is in the middle of writing, and the failure it guards is a silent loss of coverage, not a crash - so the place it has to be visible is where the walk decides to trust the invariant, which is where the comment is. The empirical check and the test cover what exists; the comment is what a future path has to read.

**3. Parse-time cost / 4. best-effort fallback deferring some errors.** Both correct as stated, both deliberate, no change. The fallback's hole in "rejected before it starts" is the price of not regressing queries ANTLR accepts and the builder does not fully support, and it is the smaller of the two harms.

**Nit - `getOriginalText`.** Left package-private; it is used across several classes in this package already, so a doc line rather than a visibility change is the right size, and it can ride along with the follow-up.

**On the `mvn -pl engine verify` I promised.** The first run reported `BUILD FAILURE`, and it was my own fault, not the branch's: I had `mvn test` runs going in the same module while it executed, so `target/classes` was rewritten under the running JVM. The failures are all `NoClassDefFoundError` on classes that exist (`ContinuousAggregateRefresher`, `LSMTreeIndexBulkLoader$1`) in `com.arcadedb.schema.**` - nothing this branch touches. Re-running it clean with nothing else against the module; I will post the real result here.

Green meanwhile: `query.**` + `function.**` 11319/0 (106 skipped). No conflict with `main`.

lvca:
Thanks - all three notes answered, one with a code change, two with reasons rather than changes.

**2. The plan cache does key on raw query text - confirmed, and recorded where it is claimed.** `CypherStatementCache` is a `Map<String, ParsedQuery>` keyed on the query string (normalised only by stripping a trailing semicolon), and the body ASTs hang off the statement inside `ParsedQuery`, so the whole build is reused on the second execution of the same text. You were right to ask for it rather than take it.

On the nesting shape specifically: it **adds rather than multiplies**. A body reached from `parseSubqueryBody` is built by its own visit, and the outer statement holds that subquery only as an expression - it never descends into the body a second time. So `EXISTS { ... EXISTS { ... } }` builds the inner body once, as part of its parent's build, not again for each level above it. Total work is linear in the text, not exponential in depth. That is now in the `parseSubqueryBody` Javadoc alongside the cost sentence, since a claim about a cache belongs next to the code that relies on it, not only in a review thread.

**1. A metric for the fallback - not taking it, and the reason is structural.** I agree the coverage loss is the invisible kind. But `parseSubqueryBody` is a static method in the parser: it has no database, no profiler, no `Micrometer` registry, and no path to one. Threading a handle through the parser purely to count an event that should never fire is a bigger change to a hot, stateless path than the observability is worth - and it would put a runtime dependency into a class whose whole job is turning text into an AST. The `WARNING` is the proportionate signal: it fires exactly when a builder defect degrades a body, and it names the body. If these ever turn out to happen in the field, the right move is to fix the builder gap, not to keep a counter of it.

**3. Release note.** Shipping in this PR - `docs/release-26.8.1.md`, both callouts (`type(node)` inside a subquery, and the `WITH *` kind survival) are in the "Potentially breaking" block.

**Still owed: the clean `mvn -pl engine verify`.** It is running now with nothing else against the module. I will post the result here before this is ready to merge - the first attempt's `BUILD FAILURE` was my own concurrent-build artifact, as described above, and I do not want that standing as the last word on it.

Issue #5637, opened 2026-07-31
Partitioned types: the strategy is not persisted, and a post-hoc index change can still disable pruning silently


lvca:
Two loose ends around `BucketSelectionStrategy`, both found while working on #5603 (PR #5605) and both left out of it deliberately: the first predates that change, the second is a gap its refusal cannot close on its own. Ordered by severity.

---

- [ ] **1. `ALTER TYPE ... BucketSelectionStrategy` is not persisted, so a partitioned type reverts to `round-robin` after a restart**

  The strategy is set in memory and nothing writes it out. `schema.json` gains no `bucketSelectionStrategy` entry unless some *later* schema mutation happens to flush the configuration:

  ```
  before close    = partitioned
  schema.json has bucketSelectionStrategy = false
  after reopen    = round-robin
  ```

  Reproduced with:

  ```sql
  CREATE DOCUMENT TYPE P BUCKETS 3;
  CREATE PROPERTY P.k STRING;
  CREATE INDEX ON P(k) UNIQUE;
  ALTER TYPE P BucketSelectionStrategy `partitioned('k')`;
  -- close, reopen: P is round-robin again
  ```

  This is arguably worse than the bug #5603 fixed, because it hits a **suitable** partition: a correctly configured type silently loses its pruning across a restart, and new records are placed round-robin among rows that were placed by the partition hash. Nothing warns, and `SELECT ... WHERE k = ?` simply gets slower.

  **Where it comes from.** Every other schema mutator calls `schema.saveConfiguration()` itself - `LocalDocumentType` does at lines 223, 355, 702 and 1427, `LocalProperty` at 64 and 90, plus `TypeBuilder` and `BucketIndexBuilder`. `setBucketSelectionStrategy` is the one that does not. The asymmetry is stark inside the same class: `setNeedsRepartition` (line ~696) saves on every real transition, so the *flag about* the partitioning persists while the partitioning itself does not.

  The existing comment in `setBucketSelectionStrategy` already anticipated this exact fragility:

  > Direct field assignment would only work because the wrapping DDL happens to save the schema afterward; relying on that is brittle.

  **Not caused by #5605** - reproduced identically with that PR's three main-source files stashed. It is why `PartitionedStrategySuitabilityTest.theFanOutWarningIsNotRepeatedOnEveryReopen` calls `saveConfiguration()` explicitly instead of relying on the DDL, with a comment saying so; that workaround should come out when this is fixed.

[...]

Issue #4134, opened 2026-05-07
LSMVectorIndex: consolidate primary constructor args into a config record


lvca:
## Background

`LSMVectorIndex`'s primary constructor currently takes 17 positional arguments. The factory handler (`LSMVectorIndexFactoryHandler.create()`) had to plumb `VectorEncoding` via post-construction mutation of the metadata field rather than adding an 18th argument:

```java
final LSMVectorIndex index = new LSMVectorIndex(...);  // 17 args
index.metadata.encoding = vectorBuilder.encoding;       // post-construction mutation
return index;
```

This was acknowledged as tech debt during the INT8 ingest landing (#4132 / PR `vector-int8-ingest`). The reasoning that the brief inconsistent window is unobservable holds for the current code, but it relies on the index not being published before `create()` returns. Future refactors that wire the index into a listener or cache during construction would break this implicit contract.

## Proposal

Replace the positional constructor argument list with a single `LSMVectorIndexConfig` record (or builder-derived value object) that carries every field currently passed positionally plus future additions like `encoding`. Pass the record into the constructor, copy it into the metadata at construction time, and drop the post-publication mutation.

Sketch:

```java
public record LSMVectorIndexConfig(
    String typeName, String[] propertyNames, int dimensions,
    VectorSimilarityFunction similarityFunction, VectorEncoding encoding,
    VectorQuantizationType quantizationType,
    int maxConnections, int beamWidth, String idPropertyName,
    int locationCacheSize, int graphBuildCacheSize,
    int mutationsBeforeRebuild, boolean storeVectorsInGraph, boolean addHierarchy,
    int pqSubspaces, int pqClusters, boolean pqCenterGlobally, int pqTrainingLimit) { }

public LSMVectorIndex(DatabaseInternal database, String name, String filePath,
    ComponentFile.MODE mode, int pageSize, LSMVectorIndexConfig config) { ... }
```

## Why now

This is purely a code-quality / safety-of-future-change item, not a behavioural fix. Filing it so the TODO comment in `LSMVectorIndexFactoryHandler.create()` is tracked and not lost when the file is touched again.

## Related

- Original INT8 ingest work: #4132
- TODO comment marker in `LSMVectorIndex.java::LSMVectorIndexFactoryHandler.create()`

lvca:
[...]

PR #5643, opened 2026-07-31
fix(engine) #5639: a METADATA key is applied or reported, and a reopened vector index keeps its metric

### LINKED ISSUE #5639: Silently ignored configuration: vector/sparse/full-text METADATA keys, and the inert dependency-review config
Two independent findings from the #5607 / PR #5610 work, filed together because they are the same failure mode: **a configuration input that is accepted, ignored, and never reported**. #5600 already established that this is the bug pattern worth closing ("a silently ignored `METADATA` is exactly what made the missing forwarding invisible"). Both can be split into separate issues if preferred.

---

## 1. `LSM_VECTOR` / `LSM_SPARSE_VECTOR` / `FULL_TEXT` METADATA drops unknown keys

#5600 added an unknown-key guard to `TypeGeoIndexBuilder.withMetadata()`:

```java
for (final String key : json.keySet())
  if (!SUPPORTED_METADATA_KEYS.contains(key))
    throw new IllegalArgumentException("Unsupported metadata key '" + key + "' for a GEOSPATIAL index. ...");
```

The other three metadata-carrying builders have no equivalent (`grep -c SUPPORTED_METADATA_KEYS` returns 0 for `TypeLSMVectorIndexBuilder`, `TypeLSMSparseVectorIndexBuilder`, `TypeFullTextIndexBuilder`). Every one of them reads the keys it knows with `if (json.has(...))` and drops the rest.

### This is not only a consistency nit: two settings are unreachable from SQL

Diffing the keys `TypeLSMVectorIndexBuilder.withMetadata(JSONObject)` reads (the `CREATE INDEX ... METADATA` path) against the keys `LSMVectorIndexMetadata.fromJSON()` reads (the persisted path):

| Key | Persisted metadata | Settable via `CREATE INDEX ... METADATA` |
|---|---|---|
| `efSearch` | read | **silently dropped** |
| `inactivityRebuildTimeoutMs` | read | **silently dropped** |
| `buildState` | read | dropped (correct, internal) |

So today:

```sql
CREATE INDEX ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 384, "efSearch": 200}
```

succeeds, reports success, and builds an index with `efSearch` at its default of 100. Same for a typo: `{"similarty": "EUCLIDEAN"}` yields a COSINE index with no warning. `efSearch` is the search-time recall/latency knob, so this is a user-visible functional gap, not just diagnostics.

### Why the obvious fix does not work as-is

Copying the geo guard verbatim **breaks JSONL import**. `JsonlImporterFormat.loadVectorIndex()` passes the whole exported index definition into `withMetadata()`:

```java
final JSONObject metadata = new JSONObject(idx.toMap());
if (metadata.has("similarityFunction") && !metadata.has("similarity"))
  metadata.put("similarity", metadata.getString("similarityFunction"));
builder.withMetadata(metadata);
```

`LSMVectorIndex.toJSON()` emits eight keys the builder never reads: `bucket`, `indexName`, `properties`, `type`, `typeName`, `version`, `similarityFunction`, `buildState`. A naive guard would reject **every** vector-index import.

### Suggested shape

1. Add `efSearch` and `inactivityRebuildTimeoutMs` to `TypeLSMVectorIndexBuilder.withMetadata()` so the SQL path can reach every user-facing setting.
2. Introduce `SUPPORTED_METADATA_KEYS` on the vector, sparse and full-text builders, mirroring geo.
3. Fix the import path *first*: either have `JsonlImporterFormat` project the export down to the settable keys before calling `withMetadata()`, or give the builders an explicit "restore from a persisted definition" entry point that tolerates the structural keys — the same distinction `TypeGeoIndexBuilder` already draws in its comment about `GeoIndexMetadata.fromJSON()` reading a *persisted* definition versus `withMetadata()` reading a *user* clause.
4. Regression test the round trip, since no vector test currently exercises a real JSONL export/import of an index definition (`VectorIndexExportImportIT` is the closest and would catch it).

Note that the SQL layer already turns an `IllegalArgumentException` from `withMetadata()` into a `CommandSQLParsingException` (PR #5610), so a new guard surfaces as HTTP 400 with no extra plumbing.

---

## 2. `.github/dependency-review-config.yml` does not match the action's schema

PR #5610 fixed the `Dependency Review` job erroring out before it reviewed anything (SPDX ids were passed to `allow-dependencies-licenses`, which takes package URLs). While there, the referenced config file looks like it is not the schema `actions/dependency-review-action` reads.

The file is organised as:

```yaml
security:
  fail_on_severity: "moderate"
licensing:
  allow: [...]
  deny: [...]
packages:
  rules:
    - name: "event-stream"
      action: "deny"
changes: {...}
exemptions: {...}
notifications: {...}
advanced: {...}
reporting: {...}
```

The action's configuration is flat (`fail-on-severity`, `allow-licenses`, `deny-licenses`, `allow-ghsas`, `allow-dependencies-licenses`, `deny-packages`, `comment-summary-in-pr`, …). None of the sections above correspond to a documented option, and when the action rejected the workflow inputs it reported errors only for `allow_dependencies_licenses` — nothing about the file's own keys.

Two things follow, and they need verifying rather than assuming:

- **If the file is inert**, the repo believes it enforces protections it does not: minimum versions for `jquery` / `bootstrap` / `datatables.net`, and hard denials of the known-compromised `event-stream` and `flatmap-stream`. That is a false sense of coverage on a supply-chain control.
- **If any of it is live**, `licensing.deny` lists `EPL-1.0`, `EPL-2.0`, `LGPL-2.1` — licences `CLAUDE.md` explicitly *allows* (EPL 1.0/2.0, LGPL 2.1+ for libraries), and which the current tree already ships (logback is `EPL-2.0` / `LGPL-2.1-only`). It would reject dependencies the project permits.

Suggested: confirm which options the pinned action version actually honours, then either express the intended policy in the real schema or delete the file and keep the policy in the workflow inputs. Whatever survives should agree with the allowed/forbidden lists in `CLAUDE.md`.

lvca:
Closes #5639.

## What the reproduction found

The issue filed two findings. Reproducing them turned up three more, all of the same failure mode - a configuration input accepted, ignored and never reported. The baseline test failed on five distinct counts:

| Finding | In the issue? |
|---|---|
| `efSearch` / `inactivityRebuildTimeoutMs` never read from the `METADATA` clause | yes |
| Unknown `METADATA` key silently dropped on `LSM_VECTOR` / `LSM_SPARSE_VECTOR` / `FULL_TEXT` | yes |
| `neighborOverflowFactor` / `alphaDiversityRelaxation` **were** read, then lost one hop later | no |
| A `FULL_TEXT` index created with **any** `METADATA` clause silently dropped to CLASSIC ranking | no |
| **A `EUCLIDEAN` / `DOT_PRODUCT` vector index came back up as `COSINE` after a restart** | no |

That last one is a correctness bug, not a diagnostics one. `LSMVectorIndex.toJSON()` writes the metric under `similarityFunction`; `LSMVectorIndexMetadata.fromJSON()` looked only for `similarity`. Nothing restored it, so after every reopen the index scored with COSINE against a graph built with EUCLIDEAN. Pinned end-to-end with vectors on a single ray - all at COSINE distance 0 from the query, all at different EUCLIDEAN distances - and verified the assertion fails when the fix alone is reverted.

Two of ArcadeDB's own tests (`LSMVectorIndexTombstoneMemoryTest`, `LSMVectorIndexCompactionTest`) believed they had disabled the inactivity rebuild through `"inactivityRebuildTimeoutMs": 0` in `METADATA` and had not.

## Why not the shape the issue suggested

The issue proposed a `SUPPORTED_METADATA_KEYS` constant on each of the three builders, mirroring geo. That adds three more copies of a list - and the list being copied is precisely what broke. The dense-vector settings existed in **seven** places:

`LSMVectorIndexMetadata` · `TypeLSMVectorIndexBuilder.withMetadata(JSONObject)` · `BucketLSMVectorIndexBuilder` (17 parallel public fields + two `withMetadata` overloads) · `LSMVectorIndexConfig` · the `LSMVectorIndex` constructor · `toJSON()` · `TruncateTypeStatement.cloneVectorMetadata`

`efSearch` and `inactivityRebuildTimeoutMs` had reached copies 1 and 6; `neighborOverflowFactor` and `alphaDiversityRelaxation` had reached 1, 2 and 3 but not 4. Any new setting has to be remembered in all seven, which is not a contract anyone can keep.

So instead of three new lists:

- **One guard, on the metadata.** `IndexMetadata.fromUserMetadata(json, indexType)` validates the keys and delegates to `applyUserMetadata()`. The key set lives on the class that owns the fields, is inherited by all four index types, and geo's hand-rolled guard moved there too - so `GEOSPATIAL` keeps its exact behaviour and message with no duplicated loop.
- **One field list.** `BucketLSMVectorIndexBuilder`'s parallel fields and the `LSMVectorIndexConfig` record are gone; the settings travel as a single `LSMVectorIndexMetadata` and `copy()` replaces the hand-written clone. `TRUNCATE TYPE` and `REBUILD INDEX` were dropping the same four settings when re-creating an index, and are fixed by construction. #4134's intent - every setting populated before the instance escapes - is preserved and now covered by a test that gives every field a non-default value, so a `copy()` that forgets one fails on that field.
- **An explicit restore entry point,** `withPersistedMetadata()`, rather than "have `JsonlImporterFormat` project the export down to the settable keys": a projection filter would have to be maintained alongside `toJSON()` and would fail silently when it fell behind - the same defect one level up. This also retires the manual `similarityFunction` → `similarity` bridge the importer was carrying.
- **The persisted definition is complete.** A setting that only lives until the next restart is barely better than one dropped at creation, so `toJSON()` writes every knob `fromJSON()` reads. `similarityFunction` stays the persisted spelling (existing schemas and the persistence test depend on it) and `fromJSON()` accepts either.

## Part 2: the dependency-review config

Verified rather than assumed. Read `config.ts` / `schemas.ts` at the pinned action SHA and cross-checked a real CI log:

- The action's configuration is **flat**. It parses the YAML, strips dashes from the keys and validates against a zod schema; unknown keys are stripped with no warning. Every top-level section of the old file (`security`, `licensing`, `packages`, `changes`, `exemptions`, `notifications`, `advanced`, `reporting`) is unknown, so **the file was entirely inert** - the minimum versions for jquery/bootstrap/datatables.net and the denial of the compromised `event-stream` / `flatmap-stream` were never applied.
- Had it been live it would have **rejected dependencies the project permits**: its `licensing.deny` listed `EPL-1.0`, `EPL-2.0` and `LGPL-2.1`, which `CLAUDE.md` allows for libraries and which the tree already ships (logback).
- One more finding: **inline `with:` inputs override the config file** (`readConfig` merges `{...externalConfig, ...inlineConfig}`), so the workflow's own `deny-licenses` was shadowing the file regardless. The step now passes only `config-file`, with a comment saying why nothing else may be added.

The policy is rewritten in the real schema (`fail-on-severity`, `fail-on-scopes`, `deny-licenses`, `deny-packages`), aligned with `CLAUDE.md`: GPL/AGPL/SSPL denied in both current and deprecated SPDX spellings (both validate; a dependency may declare either), EPL/LGPL deliberately **not** denied. `event-stream` and `flatmap-stream` are denied at every version via version-less purls, `eslint-scope@3.7.2` at the compromised version only. Two intents cannot be expressed in this action and are noted in the file: per-package minimum versions (moved to the workflow's existing `Validate Package.json` step, which gained the `datatables.net` floor) and an allow-list of licences (`allow-licenses` and `deny-licenses` are mutually exclusive; the `license-checker --onlyAllow` step covers that side).

**`fail-on-scopes` stays at `runtime`, so nothing that passes CI today starts failing.** An earlier revision of this PR also listed `development`, which the inert file had asked for - but that would have tightened a shared gate as a side effect of repairing a config that enforced nothing, and a dev-only advisory could then fail PRs with nothing to do with this one. Six reviewers flagged it and they were right: the *intent* of a file that enforced nothing is arguable, today's *effective* behaviour is not, so tightening deserves its own decision. Development dependencies stay covered by the `npm audit --audit-level=moderate` step of the same workflow. The config file records why the entry was left out, so it does not read as an oversight.

## Verification

- `Issue5639IndexMetadataKeysTest` (new, 9 tests): each of the five findings, plus the cases that must keep working - `buildGraphNow` (consumed by the SQL layer before the builder sees it), the real sparse keys, and the full-text open-ended `<field>_analyzer` / `<field>_boost` key space.
- `LSMVectorIndexMetadataPropagationTest` (renamed from `LSMVectorIndexConfigTest`): every-field `copy()` coverage.
- `VectorIndexExportImportIT` extended: the JSONL round trip now uses EUCLIDEAN and non-default `maxConnections` / `beamWidth` / `efSearch` / `mutationsBeforeRebuild` and asserts the restored metadata, so a knob lost in the round trip fails instead of silently reverting. COSINE would have passed whether or not it was restored at all.
- Full engine suite: **10545 tests, 0 failures, 23 skipped**. Plus the integration module, the server module's vector/index/full-text tests, and the 24 Studio JS tests. Eleven review rounds later the regression class is 17 tests, and the affected index/schema suites (1340) plus the export/import IT are re-run green on each rebase - this branch has been rebased four times onto a moving `main`, most recently over #5637.

## Files

**Engine**
- `schema/IndexMetadata.java` - `fromUserMetadata()` / `getUserMetadataKeys()` / `isUserMetadataKey()` / `describeUserMetadataKeys()` / `applyUserMetadata()`
- `schema/LSMVectorIndexMetadata.java` - user key set, `applyUserMetadata()`, validating setters, `copy()`, `similarityFunction` read on the persisted path
- `schema/LSMSparseVectorIndexMetadata.java`, `schema/FullTextIndexMetadata.java`, `schema/GeoIndexMetadata.java` - same contract; geo's guard relocated here
- `schema/TypeLSMVectorIndexBuilder.java` - delegates, plus `withPersistedMetadata()`
- `schema/TypeLSMSparseVectorIndexBuilder.java`, `schema/TypeFullTextIndexBuilder.java`, `schema/TypeGeoIndexBuilder.java` - delegate
- `schema/BucketLSMVectorIndexBuilder.java` - parallel fields replaced by the metadata
- `index/vector/LSMVectorIndex.java` - factory and constructor take the metadata; `toJSON()` completes
- `index/vector/LSMVectorIndexConfig.java` - **removed**
- `query/sql/parser/TruncateTypeStatement.java` - `cloneVectorMetadata` replaced by `copy()`
- `query/sql/parser/CreateIndexStatement.java` - full-text and sparse metadata errors answer 400

**Integration**
- `importer/format/JsonlImporterFormat.java` - restores via `withPersistedMetadata()`

**CI / docs**
- `.github/dependency-review-config.yml`, `.github/workflows/studio-security-audit.yml`, `docs/release-26.8.1.md`

**Tests**
- `engine/.../index/Issue5639IndexMetadataKeysTest.java` (new)
- `engine/.../index/vector/LSMVectorIndexMetadataPropagationTest.java` (renamed, rewritten)
- `integration/.../VectorIndexExportImportIT.java`

## Follow-up worth filing separately

`LSMTreeFullTextIndex.getMetadata()` and `LSMSparseVectorIndex.getMetadata()` return the base `IndexMetadata`, not the subtype - the real configuration is only reachable through `getFullTextMetadata()` / `getSparseMetadata()`. `RebuildIndexStatement` already works around it by reconstructing the full-text metadata from `toJSON()`. Not touched here.


lvca:
Thanks — points 1 and 2 were valid and are fixed in 280f8f2bc. Point 3 is a fair correction to my PR description rather than to the code. Details, including the CI premise you asked me to re-check:

### 1. Non-numeric float escaping as a 500 — fixed, and it was worse than the cast

Confirmed empirically before fixing: reverting just that one line makes the new test fail with a raw `java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number` reaching the caller. Pulling the thread found two more values that were **coerced rather than reported**, which is the same defect this PR is about:

- an integer key read `8.5` as `8` — `{"dimensions": 8.5}` silently created an 8-dimension index;
- `JSONObject.getBoolean()` answers `false` for any string that is not `"true"`, so `{"addHierarchy": "yes"}` **disabled** the setting the user was asking to enable.

Rather than adding `ClassCastException` to the catch in `CreateIndexStatement` — which fixes the classification but leaves the coercions, and leaves the next reader free to reintroduce the problem — every numeric and boolean key of the clause now goes through three shared readers on `IndexMetadata` (`metadataInt`, `metadataFloat`, `metadataBoolean`). A quoted number is still accepted (`{"efSearch": "250"}` — and now on the float keys too, which the `Number` cast rejected), a fractional value for a count is refused instead of truncated, and anything else is an `IllegalArgumentException` and therefore a 400. `TypeGeoIndexBuilder`'s hand-rolled precision check *was* exactly this rule, so it delegates now instead of duplicating it.

New test `malformedDenseVectorMetadataValuesAreClientErrors` covers all six shapes plus the quoted-number forms that must keep working.

### 2. API asymmetry — fixed, and it was six settings, not two

`TypeLSMVectorIndexBuilder` was missing a fluent setter for `efSearch`, `inactivityRebuildTimeoutMs`, `mutationsBeforeRebuild`, `locationCacheSize`, `graphBuildCacheSize` and `storeVectorsInGraph`. All six added, so the Java builder and the SQL clause reach the same set of settings — pinned by `everyDenseVectorSettingIsReachableFromTheJavaBuilder`, which asserts each one lands in the index metadata.

### 3. Message ordering — you are right, my wording was too strong

The behaviour and exception type are preserved; the key *ordering* in the message can differ, because the sort is deliberate (`Set.of` iteration order is unspecified, so the old message could list the same two keys differently between JVM runs). No code change — the sorted form is the better one. My "exact behaviour and message" phrasing in the PR body overstated it.

### Your CI double-check — verified three ways, and the strongest is from this PR's own run

You are right that the premise cuts both ways: if unknown keys are silently dropped, a typo in the new file would be inert in exactly the way I am fixing. Three independent confirmations that the keys are recognised at the pinned SHA:

1. **`ConfigurationOptionsSchema`** at `a1d282b` contains `deny_packages`, `license_check` and `vulnerability_check`, and `parseConfigFile` lists `deny-packages` among its `listKeys` (the kebab→snake conversion is what makes the dashed spelling work).
2. **`deny-licenses` is provably parsed:** `parseConfigFile` runs `validateLicenses()` over that key's value and throws on an invalid SPDX id. The job passes, so all 16 ids were read *and* validated — an unrecognised key would never have been validated at all. (I checked each id against `spdx-license-ids`: the `-only`/`-or-later` forms are current, the bare `GPL-2.0`/`GPL-3.0`/`AGPL-3.0` forms are deprecated-but-valid, and `SSPL-1.0` is current.)
3. **This PR's own `Dependency Review` run**, which is the cleanest evidence:

```
with:
  config-file: .github/dependency-review-config.yml     ← no inline inputs at all
...
Dependency review did not detect any vulnerable packages with severity level "moderate" or higher.
##[group]Denied
Config: [object Object]     ← one line per deny-packages entry, parsed
Config: [object Object]
Config: [object Object]
```

The inline `fail-on-severity` is gone, so if the file were inert the action would fall back to its schema default of `low`; it reports `moderate`. And the three `Config:` lines in the *Denied* group are the three `deny-packages` purls — the old file produced none, because none of its keys existed as far as the action was concerned. Both checks whose flags you asked about ran (the `Vulnerabilities` and `Licenses` groups are both present), which is what `vulnerability-check`/`license-check` being `true` means; those two are belt-and-braces since `true` is also the default.

### Verification of this round

`Issue5639IndexMetadataKeysTest` is now 11 tests; 1348 index/schema/full-text/truncate tests pass, plus the export/import IT.

One note for the record: on the first run of that selection, `LSMVectorIndexCompactionTest.repeatedCompactionKeepsEveryVector` aborted with "pages are still pending flush after the no-progress timeout" while a parallel compile was loading the machine. It passes 5/5 in isolation and the same selection passes clean on a quiet machine, and the abort is a no-progress *timeout*, which is the shape load produces — not related to this change, but worth flagging since that test is timing-sensitive on a busy CI runner.

### Not addressed, filed-worthy instead

`LSMTreeFullTextIndex.getMetadata()` and `LSMSparseVectorIndex.getMetadata()` return the base `IndexMetadata` rather than their subtype, so the real configuration is only reachable via `getFullTextMetadata()` / `getSparseMetadata()` — `RebuildIndexStatement` already works around it by reconstructing the full-text metadata from `toJSON()`. Out of scope here.

lvca:
Thanks. Points 2 and 3 are fixed in 813517403. Points 1 and 4 are answered below rather than changed — 4 in particular is a maintainer decision, so I have left it as the issue specified and flagged the one-line alternative.

### 2. `metadataInt` message for out-of-range — fixed

The fractional check and the range check were conflated, so `3000000000` tripped the fractional branch. They are separate now and an overflow says `must be between -2147483648 and 2147483647`. Covered by a new case in `malformedDenseVectorMetadataValuesAreClientErrors`.

### 3. PQ knobs persisted only under PRODUCT — fixed

You are right that this was the one remaining exception to the goal. Four small integers are not worth an exception a reader has to work out, so `toJSON()` now writes them unconditionally; `fromJSON()` already read them unconditionally, and `pqSubspaces: -1` (auto) round-trips as itself. No test asserted their absence. `denseVectorSettingsSurviveAReopen` now sets `pqClusters`/`pqCenterGlobally` away from their defaults and asserts they survive the reopen.

### 1. Scope — agreed in principle, and it was the filed issue's own framing

#5639 filed both halves together precisely because they are the same failure mode, and said "Both can be split into separate issues if preferred." Since the ask was to resolve that issue, I kept them together. Your point about the CI half being hard to verify without running the action is fair, which is why I did not leave it at "the schema says so": this PR's own `Dependency Review` run is the evidence (no inline inputs, yet severity `moderate` rather than the schema default `low`, and one `Config:` line per parsed `deny-packages` entry where the old file produced none). Noted for future splitting.

### 4. CI policy narrowings — one is clearly correct, one is genuinely the maintainer's call

**`eslint-scope`.** I am confident this narrowing is right, not just reasonable. The old rule was `max_version: "3.7.1"`, i.e. deny everything above 3.7.1 — but only the 3.7.2 release was compromised, and `eslint` itself depends on `eslint-scope` at 4.x/5.x/8.x. Had that rule ever been live it would have blocked every eslint upgrade. `pkg:npm/eslint-scope@3.7.2` expresses the actual advisory.

**`fail-on-scopes: [runtime, development]`.** This is the one that deserves your confirmation rather than mine. Facts, so the choice is cheap to make:

- It is stricter than today's *effective* behaviour (`runtime` only, the schema default), so a dev-only moderate advisory on a changed studio dependency can newly fail the job.
- It is exactly what the old file's `security.fail_on_scopes` asked for, and #5639 asked for the intended policy to be expressed in the real schema — so expressing it faithfully means including `development`.
- It does not gate anything new in practice: the same workflow already runs `npm audit --audit-level=moderate` over the whole tree, dev dependencies included, so a dev-only moderate advisory fails that step today regardless.
- The job only runs on PRs touching `studio/**` or this workflow.

I have kept it, on the reading that the file should say what it was always meant to say. If you would rather not tighten the gate, it is a one-line revert — drop the `development` entry from `fail-on-scopes` in `.github/dependency-review-config.yml` and the behaviour matches today's exactly.

### Verification

`Issue5639IndexMetadataKeysTest` is 11 tests; 1343 index/schema/truncate tests and the export/import IT pass. `LSMVectorIndexQuantizationTest` and `LSMVectorIndexPersistenceTest` (the two that assert on `toJSON()` output) pass unchanged.

lvca:
Thanks. Points 2 and 4 are fixed in 57aafc87c, and point 1 is now an explicit upgrade note. Point 3 I am leaving as a follow-up, with one correction to its premise.

### 2. `fromJSON` float reads use a raw cast — fixed

Good catch, and the consequence is worse than on the user path: a `ClassCastException` there fires while *opening the database*, not while answering a request. Both reads now use `metadataFloat`, so the persisted path is at least as tolerant as the user path. That was the last raw `(Number)` cast in the package.

### 4. `IndexMetadata.java` header — fixed

License block moved above `package`, and the `SPDX-FileCopyrightText` / `SPDX-License-Identifier` lines added to match every sibling in `com.arcadedb.schema`. Worth doing beyond tidiness: the repo has a `license-maven-plugin` and a `REUSE.toml`, so a file missing its SPDX identifier is a real inconsistency, not just a cosmetic one.

### 1. Behaviour change — promoted to an explicit upgrade note

Agreed this deserves more than a passing mention, since a statement that "worked" before can now fail. `docs/release-26.8.1.md` now carries a dedicated **Upgrade note** paragraph: what breaks (a stored `CREATE INDEX` or migration carrying a stray/misspelled key), why that is the point (the key was never doing anything), what is *not* affected (existing indexes, and reading a persisted definition, which stays tolerant of its structural keys), and that the error message lists the accepted keys. The same paragraph also documents the two coercions that are now refused — `{"dimensions": 8.5}` and `{"addHierarchy": "yes"}` — since those also used to "work".

### 3. Deny-list-only licensing — one correction, then agreed as a follow-up

One correction to the premise: the scope is not npm-only. GitHub's dependency graph covers Maven `pom.xml`, and `dependency-review-action` reviews the graph diff for the PR, so the `deny-licenses` list applies to the Java dependency graph too — GPL/AGPL/SSPL are gated on both sides, not just for studio's npm tree. (Nothing showed in this PR's run because it changes no dependency manifest.)

What you have identified is real but narrower: the **allow-list** side. `license-checker --onlyAllow` covers npm only, and the action cannot supply an allow-list here because `allow-licenses` and `deny-licenses` are mutually exclusive — so a weak-copyleft Java dependency such as MPL-2.0 or CDDL passes both gates. I have deliberately not fixed that in this PR for two reasons:

- **It is not expressible in the tool being fixed.** Adding an allow-list means dropping the deny-list, which would trade a definite gate on GPL/AGPL/SSPL for an indefinite one; the project's allowed set in `CLAUDE.md` is wide (Apache-2.0, MIT, BSD-2/3, EPL-1.0/2.0, UPL-1.0, EDL-1.0, LGPL-2.1+, CC0/public domain), and enumerating it as an exact SPDX allow-list is a separate exercise with its own false-positive risk.
- **It would mean inventing policy.** `CLAUDE.md` neither allows nor forbids MPL-2.0 or CDDL. Denying them is a licensing decision for the maintainer, not something a config cleanup should slip in — and this PR is about a config that claimed to enforce a policy it did not have, so quietly adding rules nobody asked for would be the same mistake in the other direction.

The right shape for it is a Maven-side allow-list gate (the `license-maven-plugin` already in the build can do this), filed separately so the allowed-set decision gets its own review. Happy to open that issue.

### Verification

1319 index/schema tests plus the reload-path suites (`LSMVectorIndexPersistenceTest`, `LSMVectorIndexRecoveryTest`, `LSMVectorIndexMetadataPropagationTest`, and the 11-test regression class) pass.

lvca:
Thanks. Point 1 is fixed in 6e43cae93; points 2 and 3 needed filing rather than code, and both are now filed.

### 1. Asymmetric catch on the GEOSPATIAL branch — fixed

Agreed, and your framing is the reason to change it: the 400 was an accident of the current implementation rather than a contract. All four branches now catch `IndexException | IllegalArgumentException | JSONException`. Worth noting this PR made the asymmetry slightly more likely to bite, since geo's precision read now goes through the shared `metadataInt` — same exception type today, but one more place a future change could pick a different one.

Pinned by `malformedGeospatialMetadataIsReported`, which covers an unknown key, a fractional precision, a non-numeric precision, an out-of-range precision and an unknown tokenization. 85 geospatial + regression tests pass.

### 2. `fail-on-scopes: development` — flagged to the maintainer, left as the issue specified

Already surfaced in [this comment](https://github.com/ArcadeData/arcadedb/pull/5643#issuecomment-5146948550) with the one-line revert, and repeating your heads-up here since it is the right one: **a first failure of the `Dependency Review` job on a devDependency advisory will be the new gate doing its job, not a broken config.** Two mitigating facts: the job only runs on PRs touching `studio/**` or that workflow, and the same workflow's `npm audit --audit-level=moderate` step already fails on dev-dependency advisories today, so the practical delta is narrow. Still @lvca's call; the revert is dropping one list entry.

### 3. Full-text / sparse metadata not restored on JSONL import — filed as #5650

You are right that it is the same failure mode, just on the restore path instead of the creation path, and it deserved more than a mention in this PR's description. Filed as **#5650** with what I verified while writing it up:

- the export is complete — `FullTextIndexMetadata.writeToJSON` emits `similarity`, `bm25_k1`/`bm25_b`, the analyzers and every per-field `<field>_analyzer` / `<field>_boost`, so nothing is missing from the dump, only from the restore;
- so a BM25 index with tuned parameters and per-field boosts silently **ranks differently after a restore than before it**, and an `IDF`-scored sparse index reverts to `NONE`, which changes scores;
- the fix is wiring rather than design: give the full-text and sparse builders the same `withPersistedMetadata()` entry point the vector builder now has and route `loadSchema` through it, since the strict/tolerant reader split already exists;
- and the test has to assert the restored *metadata*, not just that the index exists — a test that only checks existence passes today.

The `getMetadata()` subtype gap is folded into #5650 rather than filed separately, because it is why the importer fix is awkward to write cleanly: `RebuildIndexStatement` already reconstructs a `FullTextIndexMetadata` from `toJSON()` instead of asking the index for it, and a caller holding a logical `TypeIndex` cannot reach the configuration without casting to the concrete bucket index.

I also filed **#5651** for the licensing gap from your previous round — with one correction carried over: the deny-list side does cover the Maven graph (GitHub's dependency graph parses `pom.xml`), so what is missing is specifically the allow-list on the Java side. That issue puts the licence-policy decision (MPL-2.0, CDDL, BUSL-1.1, Elastic-2.0 — `CLAUDE.md` names none of them) before the mechanism, and points at `license-maven-plugin`, already in the build, as the place to enforce it across the whole reactor rather than only on PRs that touch a manifest.

### State of the PR

Four review rounds, each smaller than the last: correctness (the reopened metric) → classification (500 vs 400) → cosmetics (messages, headers) → this one (contract robustness + two filings). Rebased twice onto a moving `main`, most recently over #5615, which touches `LSMVectorIndex.java` — I checked that the clean textual merge was not hiding a semantic one (its five hunks are pure insertions at lines ~68/1396/1987/2001/2412; mine are the factory, the constructor and `toJSON()`) and re-ran all 257 vector tests, including the two suites #5615 added, since the delta-scan path reads `neighborOverflowFactor`/`alphaDiversityRelaxation` — settings this PR causes to actually arrive for the first time.

lvca:
Two reviews landed together, so this answers both. Three points acted on in 8e83dd791; the rest need no change and are answered below.

### The two reviews disagreed on one point, which located a real defect — in the comment, not the code

One review said of the `fromJSON()` int-vs-float asymmetry *"Fine as-is; noting it so a future reader doesn't 'fix' the ints to match"*; the other said *"the comment's rationale does not actually distinguish them — consider using the same reader for all numeric fields, or trim the comment."*

Both are right, and together they show the comment was misleading: `getInt()` already parses a quoted number through Gson's lazy `Number`, so only the floats ever had the defect. Two careful readers drawing opposite conclusions from the same comment is the signal. The code is unchanged — the ints stay on `getInt()` — and the comment now states why the reader is *deliberately not* mirrored onto them, instead of implying they share the problem.

The same review's second half was a genuine gap, though: `quantization` and `encoding` were read on the persisted path with `valueOf`/`fromString` directly, so a corrupted or hand-edited `schema.json` failed to **open the database** with a bare enum constant name. They now go through the same setters as `similarity` — for the message, not for the validation.

### `defaultOperator` not validated — fixed

Correct, and it was the last silent-accept on the full-text clause: the query parser recognises only AND and OR, so anything else was accepted and then behaved as OR. `setDefaultOperator` validates and normalises now, so `{"defaultOperator": "XOR"}` answers 400 while `"and"` still works. `fromJSON` deliberately does **not** route through the setter — it reads a definition that setter wrote, and refusing a value there would make a database unopenable rather than report a typo, which is the same persisted-vs-user split the rest of the PR rests on. New test `unknownFullTextDefaultOperatorIsReported`.

### Public-field removal on `BucketLSMVectorIndexBuilder` — intentional, and worth being explicit about

Flagging it was right. The fields are gone deliberately: they *were* the bug, since a setting added to the metadata had to be remembered there too, and two never were. Mitigations, in order of weight: nothing in the repo references them (I re-grepped across all modules, including tests); the class is only reachable via `LocalSchema.buildBucketIndex()`/`createBucketIndex()`, so it is engine-internal in practice rather than a documented surface; the fluent `withX()` methods are all preserved, plus a new `withEfSearch`; and `getVectorMetadata()` exposes everything the fields did, as one object. An embedder driving the bucket builder by field assignment would need a source change — a fair trade for removing the duplication that caused the issue.

### CI: already confirmed by a real run on this PR

Both reviews asked for the pinned action to be confirmed against a live run rather than only against its source — agreed, that is the right standard. It has already happened here: the `Dependency Review` job passes on this PR (it triggers because the PR touches `studio-security-audit.yml`), and its log shows the file being honoured rather than merely accepted:

```
with:
  config-file: .github/dependency-review-config.yml     ← no inline inputs at all
...
Dependency review did not detect any vulnerable packages with severity level "moderate" or higher.
##[group]Denied
Config: [object Object]     ← one line per version-less deny-packages purl, parsed
Config: [object Object]
Config: [object Object]
```

Three independent signals in that one run: `moderate` proves the file supplies `fail-on-severity` (with the inline input gone, an inert file would fall back to the schema default `low`); the three `Config:` lines are the three `deny-packages` purls, where the old file produced none; and both the `Vulnerabilities` and `Licenses` groups execute, which is what `vulnerability-check`/`license-check` being true means. Separately, `deny-licenses` is provably parsed because the action *validates* those SPDX ids and would fail the job on a bad one — an unrecognised key would never be validated at all.

### `fail-on-scopes: development` — still @lvca's call

Raised by both reviews, and correctly. It is stricter than today's effective behaviour, so **the first failure will be the new gate working, not a broken config.** It is what the old file's `fail_on_scopes` asked for and what #5639 asked to be expressed faithfully; the practical delta is narrow (studio-only PRs, and the same workflow's `npm audit --audit-level=moderate` already fails on dev advisories). If you would rather not tighten it, drop the `development` entry from `fail-on-scopes` and behaviour matches today exactly.

### Verification

435 full-text + vector tests and 1352 index/schema/truncate tests pass.

lvca:
Thanks — no correctness bug found this round, and the verification notes match what I intended (particularly that `copy()` dropping `buildState` is what neutralises the one behavioural change `withPersistedMetadata` -> `fromJSON` introduces on the import path; that was deliberate but I had not spelled it out).

One change, in b0df7a181, and it is a deletion.

### 2. "Confirm the new keys are actually live" — resolved by removing the two that cannot be

You are the third reviewer to ask this, and the standard is right: the premise of the rewrite is that unknown keys vanish without a warning, so a typo in the new file would be inert in exactly the way I am fixing. For most keys the green run on this PR is the proof, and the log shows it:

```
with:
  config-file: .github/dependency-review-config.yml     ← no inline inputs at all
...
Dependency review did not detect any vulnerable packages with severity level "moderate" or higher.
##[group]Denied
Config: [object Object]     ← one line per version-less deny-packages purl, parsed
Config: [object Object]
Config: [object Object]
```

`moderate` proves `fail-on-severity` comes from the file (with the inline input gone, an inert file falls back to the schema default `low`). The three `Config:` lines are the three version-less `deny-packages` purls, where the old file produced none. And `deny-licenses` must have been parsed, because the action *validates* those SPDX ids and would fail the job on a bad one — an unrecognised key is never validated at all.

**For `license-check` and `vulnerability-check`, though, there is no proof to be had, and I should not have implied otherwise.** Both were `true`, which is already the action's default, so a key that was dropped and a key that was honoured produce byte-identical behaviour — no run can distinguish them. My earlier "belt-and-braces" phrasing was really an admission that they were decoration.

So they are gone. Every key left in the file changes what the action does, visibly, in the job log:

```yaml
fail-on-severity: moderate
fail-on-scopes: [runtime, development]
deny-licenses: [16 GPL/AGPL/SSPL ids]
deny-packages: [event-stream, flatmap-stream, eslint-scope@3.7.2]
```

That is also the more honest fix, because carrying a setting whose effect cannot be demonstrated is precisely how the old file came to read as an enforced policy it did not have. Behaviour is unchanged — the two keys restated defaults.

### 1, 3 — already handled, no change

**Stored-DDL break:** documented as an explicit upgrade note, and thank you for independently confirming the scope (only new `CREATE INDEX`; reload, `TRUNCATE` and `REBUILD` all go through `fromJSON`/`copy()` and never touch `fromUserMetadata`). That was the risk I most wanted a second pair of eyes on.

**`fail-on-scopes: development`:** flagged for @lvca in three comments now, including the one-line revert. It is what the old file's `fail_on_scopes` asked for and what #5639 asked to be expressed faithfully, and the same workflow's `npm audit --audit-level=moderate` already fails on dev-dependency advisories — so the practical delta is narrow, but the first failure will be the gate working rather than a broken config.

### 4. `metadataBoolean` rejecting numeric `1`/`0` — keeping it, and here is why

You called it defensible without asking for a change; recording the reasoning so it does not look accidental. Accepting numbers means picking a truthiness rule, and the moment `1` is true, `2` and `-1` need an answer. Every answer is a coercion — which is the thing this reader exists to remove, since the bug it replaces was `getBoolean()` quietly reading `"yes"` as `false`. A real boolean and the two strings that spell one are the shapes with no ambiguity; anything else gets a message naming what was expected. A client sending `1` gets told so on the first `CREATE INDEX`, which is cheaper than an index built with a setting it did not ask for.

### State

Six review rounds, converging: correctness (the reopened metric) -> classification (500 vs 400) -> cosmetics (messages, headers) -> contract robustness (the geo catch) -> comment accuracy (where two reviewers disagreed) -> this one, which found no code defect. Two follow-ups filed along the way (#5650, #5651). Rebased twice onto a moving `main`, most recently over #5615 which touches `LSMVectorIndex.java`, with the merge verified as textual rather than semantic and all 257 vector tests re-run. From my side this is ready; the only open item is @lvca's call on the dev-scope gate.

lvca:
All three actionable points are in b9f613e2e. Point 2 turned out to be fixable rather than only documentable, which was the more valuable outcome.

### 2. Per-field typos silently accepted — fixed, not just noted

You suggested calling this out in the release note, since it is "the exact silent-drop failure mode this PR otherwise eliminates". That framing is what made me look again at whether it has to be a trade-off, and it does not: an analyzer or boost for a property the index does not cover is **provably** dead configuration, because only an indexed field is analyzed and a boost applies to field-qualified matches, which only an indexed field can produce. There is no legitimate reading of `titel_boost` on an index over `title`.

So the user clause now checks the field against the index's own property list and names them in the error. My earlier justification for shape-only matching ("a property added later") does not survive contact with that argument — a boost cannot apply to a property this index does not cover, whenever it is added.

`fromJSON` stays tolerant, and for a sharper reason than symmetry: an index whose property list changed must still **open**, and refusing a stale per-field entry there would cost a database its availability in order to report a setting that is merely inert. That is the persisted-vs-user split doing real work rather than being a convention.

New test covers a misspelled field, a misspelled analyzer key, and a real-but-not-indexed property.

### 1. Reserved-key collision — documented

Right, and there is a second case in the same family, so both are now in the `isUserMetadataKey` Javadoc: a property named `index` or `query` cannot take a per-field analyzer (`index_analyzer`/`query_analyzer` are the reserved keys for the indexing and querying analyzers), and a property whose own name ends in `_analyzer` or `_boost` reads as a per-field key for the shorter name. Both are ambiguities of the notation rather than of the reader, which is why they are documented instead of worked around.

### 3. The plain-metadata branch of `BucketLSMVectorIndexBuilder.withMetadata` — test added

Your read is correct: the constructor installs the vector metadata, so `this.metadata` is already the right subtype and propagation through `createBucketIndex()` is unaffected; the branch only grafts the shared fields (`collations`, `typeIndexName`, ...) on top. But you were right that it deserved a test rather than an argument, and the stronger test is the path with **no** `withMetadata()` call at all, since that is the one that depends entirely on the constructor's installation now that the builder keeps no fields of its own. `bucketLevelVectorBuilderCarriesItsSettings` drives `buildBucketIndex(...).withType(LSM_VECTOR)` with the fluent setters only and asserts all four settings land in the created index.

### 4. `fromJSON` integer reads — no action, and thanks for checking

Noted that you verified it. That asymmetry is the one where two earlier reviews reached opposite conclusions from my comment, which is why the comment now states explicitly that the reader is *deliberately not* mirrored onto the integer keys.

### Verification

`Issue5639IndexMetadataKeysTest` is 15 tests; 1354 index/schema/truncate tests and the export/import IT pass. The `Dependency Review` job also re-ran green after the previous commit trimmed the two unobservable keys, with the log confirming `moderate` and the three `deny-packages` purls still come from the file.

### State

Seven rounds. The last two found no correctness defect, and this round's three items were a provable-typo tightening, a documentation note and a test — so from my side this is done. The only thing still open is @lvca's call on `fail-on-scopes: development`, which is one line either way.

lvca:
Points 1 and 2 are now in the upgrade note (622250d09). Point 3 I verified rather than assumed — result below. The two "Minor" notes need no change and are answered.

### 1. The strict reading applies to the Java builder too — documented, and `buildGraphNow` was the sharp example

You are right that the upgrade note only covered stored DDL, and the `buildGraphNow` observation is the one that makes the gap concrete: it is a directive of the SQL layer rather than an index setting, so `CreateIndexStatement` consumes and removes it before the builder is called. A Java caller passing it through `withMetadata(JSONObject)` is therefore passing a key the builder has *never* understood — previously dropped, now reported. The note now says this, and points such callers at `withPersistedMetadata(JSONObject)` for the restore case, since that is the entry point that tolerates the structural keys an exported definition carries.

### 2. Public-field removal — added to the note

Recorded with the reason they went rather than being kept in sync: those fields *were* the defect, since a setting added to the metadata had to be remembered there too and two never were. The note also says what replaces them (every fluent `withX()` preserved, `withEfSearch` added, `getVectorMetadata()` returning the whole configuration).

### 3. Nothing consumes the base-class `metadata` on the vector path — verified

You asked for confirmation rather than taking it on trust, which was the right instinct, so I checked instead of arguing. Every reader of `builder.getMetadata()` in the engine:

```
index/fulltext/LSMTreeFullTextIndex.java:119,120,125,128
index/sparsevector/LSMSparseVectorIndex.java:113
index/geospatial/LSMTreeGeoIndex.java:125
```

All three are the full-text, sparse and geospatial factories — none of them ever receives a `BucketLSMVectorIndexBuilder`. The vector factory reads `vectorBuilder.getVectorMetadata()`, and `LocalSchema.createBucketIndex` reads the `metadata` parameter it was passed, not the builder's field. So no consumer of that branch expects the passed non-vector instance, and the graft is safe. `bucketLevelVectorBuilderCarriesItsSettings` (added last round) covers the sibling path where `withMetadata` is never called at all.

### Minor: `copy()` aliases `collations`

Correct, and deliberately unchanged: it matches the `cloneVectorMetadata` it replaces, and `LocalSchema.createBucketIndex` assigns `index.getMetadata().collations = metadata.collations` immediately afterwards regardless, so a defensive copy here would remove one alias out of two and imply a guarantee the surrounding code does not make. If per-index collation mutation is ever added, both places need to change together — which is the useful form of this note.

### Minor: larger `schema.json`

Intentional, as you read it. Four PQ integers and five cache/rebuild integers per vector index, against the alternative of a persisted definition with an exception a reader has to work out. Also the honest trade, given the bug being fixed was a setting that did not survive a restart.

### CI: `deny-licenses` only acts on resolvable licences

Accurate, and it is why the file comments name `license-checker --onlyAllow` as the allow-list gate rather than implying the action covers that side. The remaining gap — that the allow-list runs on npm only, not the Maven graph — is filed as #5651, with the licence-policy decision (MPL-2.0, CDDL, BUSL-1.1, Elastic-2.0, none of which `CLAUDE.md` names) put before the mechanism.

### State

Eight rounds. The last three found no correctness defect: this one produced two documentation additions and one verification. From my side this is complete — the only open item remains @lvca's call on `fail-on-scopes: development`, one line either way.

lvca:
Both actionable points are in 89a34b817. Point 2 is the most consequential thing raised in nine rounds, and I had missed it.

### 2. Existing non-COSINE indexes will see their results change — now documented properly

You are right, and this is the note that matters most in the whole PR. My upgrade notes covered the METADATA strictness and the Java API, i.e. things that *fail loudly*. This one changes results **quietly and for the better**, which is exactly the case a release note has to state, because "better" still means different:

> If you have an existing `EUCLIDEAN` or `DOT_PRODUCT` vector index, its searches have been scoring with COSINE since the first restart after it was created. On the first reopen after this upgrade it scores with the metric it was created with — so **distances and result ordering will change**, and they change to what was asked for.

I took your positive framing and made it explicit: nothing needs re-creating, re-importing or rebuilding, because the graph was always built with the correct metric and only the search side disagreed with it, so the fix applies on open. A COSINE index is unaffected in every respect. And the actionable part — if anyone compensated for the old behaviour in a tuned distance threshold or a golden-file test of result order, that is the thing to re-check.

### 1. Unchecked cast in `withMetadata(IndexMetadata)` — fixed, and it was in two builders

A fair hit: I wrote `vectorMetadata()` specifically to avoid a bare `ClassCastException` "blamed on an unrelated line", then left the raw cast in the setter next to it. `TypeLSMSparseVectorIndexBuilder` had the same, so both now use the guard the full-text, geospatial and bucket-level vector builders already had — all four type builders consistent, with the null case preserved. `aMetadataOfTheWrongSubtypeIsReported` pins both messages.

### 3, 4 — no change

**Scope:** flagged by four reviewers now, and fair each time. #5639 filed both halves together as one failure mode and offered the split; I kept them together because the ask was to resolve that issue, and the CI half is verified by this PR's own `Dependency Review` run rather than by argument. Recorded for next time.

**`fail-on-scopes: development`:** still @lvca's call, raised in five comments now with the one-line revert. Confirming your read of it: intended, and the first failure will be the gate working.

### Things you checked that I want to acknowledge

Your verification list matched my intent on every item, including two I had not spelled out anywhere: that the `_boost` branch needs no `USER_METADATA_KEYS` exclusion because no fixed key ends in `_boost` (true, and it is why only the `_analyzer` branch carries one), and that `copy()` omitting `buildState` is what keeps the import path unaffected now that `withPersistedMetadata` reads it. Independent confirmation of the reasoning is worth more than agreement with the conclusion.

### Verification

16 regression tests, 1355 index/schema/truncate tests.

### State

Nine rounds. The last four found no correctness defect; this one produced a consistency fix and the upgrade note above. From my side this is complete, and the only open decision remains the dev-scope gate.

lvca:
19da174d5 acts on points 1, 2 and 4, and closes the coverage gap Codecov reported alongside this review. Point 2 is a reversal on my part.

### 2. `fail-on-scopes: development` — reverted to `runtime`

You are the **sixth** reviewer to flag this, and that consensus is the evidence that my judgement was wrong rather than merely arguable.

My reasoning was that the old file's `fail_on_scopes` asked for `development`, so expressing it was faithful to the intent #5639 asked me to express. What I missed is that the *intent* of a file that enforced nothing is arguable, while today's *effective* behaviour is not — and making a shared gate stricter is widening scope beyond "repair a config that was inert", with a cost paid by contributors on PRs that have nothing to do with this one. Six people independently reading it as surprising is itself a cost.

So `fail-on-scopes` is `runtime`, the action's default and what this job has effectively enforced all along. **Nothing that passes CI today starts failing.** Development dependencies are not unguarded: the same workflow's `npm audit --audit-level=moderate` covers the whole tree. The file records why the entry was left out, so it does not look like an oversight, and tightening can be taken up on its own merits. This also retires the one item that was waiting on @lvca.

### 1. Release-note prominence — hoisted

Fair: the notes existed but sat mid-section while the blast radius is the whole `METADATA` clause on both the SQL and Java paths. The section now opens with a two-item callout — the strictness (fails loudly) and the vector metric change on reopen (changes results quietly) — with the detail below.

### 4. `metadataFloat` had no finiteness guard — fixed

Right, and inconsistent with `metadataInt`'s range check, which I had already added for exactly this reason. A magnitude no float can hold arrives as an infinity, and an infinite tuning factor is not a value anyone meant. Refused now.

### Codecov's 25% on `BucketLSMVectorIndexBuilder` — worth closing, and not for the number

Codecov landed a report next to your review putting that file at 25% on this diff. Normally I would not chase a coverage figure, but here it pointed at something real: those setters became one-line delegations in this PR, and a delegation that writes the *neighbouring* field — `withBeamWidth` landing on `maxConnections` — would be completely invisible. The index would simply run with a setting nobody asked for, which is this issue's failure mode reintroduced by the fix for it. `everyBucketBuilderSetterWritesItsOwnField` drives all fourteen with distinct values and asserts each against the field it should have written.

### 3, 5 — no change

**The two tests relying on the now-honoured `inactivityRebuildTimeoutMs: 0`:** exactly right, and worth restating because it is good evidence rather than a caveat — those tests documented an intent (`"disable both of its triggers"`) that the engine silently ignored. They now do what they say. Both pass, and the full suite was green at 10545/0.

**`datatables.net` floor:** stricter than the old inert text (`1.11.3` vs major 2) and passing today on `^3.0.0`. The step compares majors, so a floor of 1 would be no gate at all; 1.x is superseded, so 2 is the meaningful expression of "not 1.x".

### Verification

17 regression tests; 1332 index/schema tests.

### State

Ten rounds. The last five found no correctness defect. This round's outcome is a scope reversal, a documentation hoist, one guard and one test — nothing outstanding from my side, and no decision left pending for @lvca.

lvca:
Point 1 is fixed — thank you for catching it, it was the one thing that could have misled a maintainer. Points 2 and 3 I verified rather than argued; results below. No code change was needed this round.

### 1. Stale PR description on `fail-on-scopes` — corrected

Entirely my fault: I reverted `development` in the config *and* the release note but left the PR description describing the earlier revision. Anyone reading the description would have concluded dev-scope gating went live when the committed config says `[runtime]`. The description now states the committed behaviour and why the entry was dropped, and I refreshed the stale verification paragraph while there. Good catch — a description that contradicts the diff is exactly the kind of "documentation says it enforces something it doesn't" this PR is about.

### 2. Version-less purl semantics — confirmed at the pinned SHA

Fair standard to hold this to. I could not give you the CI observation you asked for without committing a compromised package to the tree, which would be a bad trade for a proof — but the pinned source is unambiguous. `deny.ts` at `a1d282b`:

```ts
if (
  (!denied.version || change.version === denied.version) &&
  change.name === denied.name
) {
  changesDenied.push(change)
}
```

`!denied.version` short-circuits the version comparison, so `pkg:npm/event-stream` matches **every** version by name, and `pkg:npm/eslint-scope@3.7.2` matches only that release. That is exactly the split the file intends. Combined with the job log showing one `Config:` line per parsed purl, the parse and the match are both accounted for: the parse is observed, the match is pinned by the source at the SHA we run.

### 3. No open-time path trips the sparse `REBUILD INDEX` gap — confirmed

You asked me to confirm no invalid-index recovery path hits it in practice. It does not. The schema-load path is **advisory only** (`LocalSchema`, ~line 2464):

```java
LogManager.instance().log(this, Level.WARNING,
    "Index '%s' of database '%s' should be rebuilt: %s. Run: REBUILD INDEX `%s`", ...);
```

It logs and tells the operator to run the statement; it never invokes a rebuild itself, and there is no automatic rebuild elsewhere on open. So a sparse index can only hit that failure when an operator runs `REBUILD INDEX` explicitly — never during a database open or an upgrade. I have added that finding to #5650, since it bounds the severity of the issue filed there.

Also noting your observation that this PR *improves* the failure: it was a bare `ClassCastException` before and is now an `IllegalArgumentException` naming the expected type. Still a failure, still pre-existing, but a legible one.

### Minor points — both left as they are

**PQ knobs written unconditionally:** as you read it. Nine integers per vector index, backward compatible, against a persisted definition with an exception in it.

**`withMetadata(null)` wiping the installed metadata:** I considered the null short-circuit and decided against it. All four type builders behave identically here, and it matches prior behaviour; more to the point, `vectorMetadata()`/`sparseMetadata()` exist precisely so that a metadata-less builder reaching `create()` reports *what* is wrong rather than throwing a bare NPE — a null short-circuit would instead silently keep an instance the caller had asked to clear. Turning an explicit `null` into a no-op trades a legible error for a surprising success, which is the wrong direction for this PR.

### State

Eleven rounds; the last six found no correctness defect. This round's only required change was to the PR description, which is now consistent with the diff. Four rebases onto a moving `main`, most recently over #5637, with the affected suites re-run green each time. Nothing outstanding from my side and no decision pending for @lvca.

lvca:
Points 3 and 4 are in a4e0c5ca8 (comments only), point 5 is now recorded on #5650, and points 1 and 2 are release-management calls for @lvca rather than things I should decide. Detail below.

### 3. `applyUserMetadata` is not atomic — documented

Right, and the contract is worth stating rather than leaving a reader to reconstruct. A key that fails validation leaves the keys read before it applied; that is safe because the metadata belongs to a builder whose `create()` the failure prevents, so the half-applied state is discarded with it and no index is ever built from it. The comment also names the consequence that matters: a builder *reused* after a rejected clause would carry those earlier values, which is why the callers construct one per statement.

I considered making it atomic (apply to a `copy()`, swap on success) and decided against: `copy()` exists only on the vector metadata, so it would be an inconsistent guarantee across the four types, and it would buy nothing for any current caller.

### 4. `similarity` / `similarityFunction` precedence — documented

Recorded in the comment: only one of the two appears in anything this engine wrote, and should a hand-edited definition carry both, `similarity` wins because it is the spelling a human would have reached for.

### 5. No full-text / sparse round-trip assertion — added to #5650

Good catch, and it sharpened that issue rather than just adding to it, so I have [commented there](https://github.com/ArcadeData/arcadedb/issues/5650#issuecomment-5148518273): the test to write is a JSONL round trip asserting the **restored metadata**, not a unit test, because a test that only checks the index exists passes today — which is why the gap survived. I also recorded a finding that bounds that issue's severity: no open-time path hits the sparse `REBUILD INDEX` failure, since the schema-load path only logs `"should be rebuilt ... Run: REBUILD INDEX"` and never invokes one, so it is reachable only when an operator runs the statement explicitly.

### 1. Scope — the fifth reviewer to raise it, and I want to be straight about the trade

Consistently raised and consistently fair. Two honest sides:

- **For splitting:** the halves share no code, and the correctness fix is the one a user might want backported alone.
- **Against splitting now:** this PR carries twelve rounds of review, and re-cutting it discards that history — the CI half in particular is only convincingly verified *by this PR's own `Dependency Review` run*, which is what let me prove the old file was inert rather than merely argue it. A fresh PR starts that evidence from zero.

If you want them separated, @lvca, the clean cut is: engine + tests + engine release notes in one, and `.github/dependency-review-config.yml` + `studio-security-audit.yml` + that release-note section in the other. Say the word and I will do it. My recommendation is to merge as-is and take the note for next time, which is what I have recorded.

### 2. Semver / release-line placement — @lvca's call, with the precedent

Worth raising, and I do not think it should be settled by me. The datum: the 26.8.1 notes *already* carry comparable strictness changes — #5600 made `METADATA` on a settings-less index type raise, and #5603 made a partitioned-index configuration be refused outright rather than silently mis-serving unique lookups. So this PR is consistent with how this line is being managed rather than unusual for it. If 26.8.1 is meant to be a strict patch, this and those two belong together on whatever the next minor is; that is a release call, and the release note gives you the material either way.

### Verification

19 tests in the regression class and the persistence suite. No behaviour change in this commit.

### State

Twelve rounds; the last seven found no correctness defect, and this round produced two comments and an issue update. Nothing outstanding from my side. The two open items are both yours: whether to split, and which release line this lands on.

Issue #5152, opened 2026-07-08
Root cause of count(*) drift (#5149): lazy cached-counter recompute in LocalBucket.count() races with concurrent commits and loses deltas


robfrank:
## Context

#5149 reported `SELECT count(*)` disagreeing with `count()` / a real scan (user saw `count()`=676 vs `count(*)`=606). #5150 is a reactive repair (`CHECK ... FIX` invalidates the cached counter so the next count rescans). This issue is the **root cause** of *why* the counter drifts, requested as a follow-up during that PR's review.

## Summary

`count(*)` is served from `LocalBucket.cachedRecordCount`, an incrementally-maintained per-bucket counter. The counter is repopulated from a real scan **only** on the `cachedRecordCount == -1` branch of `LocalBucket.count()`, and that recompute is **not serialized against concurrent commits**. A commit that publishes a record and folds its delta while a recompute is in flight can have its delta silently dropped, and the recompute then stores a total that missed that record - a lost update that persists until the next `-1` recompute.

## The invariants involved (all verified in source)

1. `count(*)` -> `SelectExecutionPlanner.isCountStar()` -> `CountFromTypeStep` -> `LocalDatabase.countType()` -> sums `LocalBucket.count()`.
2. `LocalBucket.count()` (`LocalBucket.java:415-458`) returns `cachedRecordCount + txDelta` when `cachedRecordCount > -1`; it does a full page scan and `cachedRecordCount.set(total)` **only** when the value is `-1`. So the counter is (re)written from a scan exclusively on the `-1` branch.
3. Commits fold their per-bucket deltas at `TransactionContext.java:971-977`, guarded by `getCachedRecordCount() > -1` - when the counter is `-1` the delta is **discarded** (normally fine: `-1` means "recompute on next read").
4. `LocalDatabase.commit()` (`LocalDatabase.java:515`) and `countType()` (`LocalDatabase.java:592`) both run under `executeInReadLock` - the **read** side of the DB RW lock, which permits multiple concurrent holders. They do not exclude each other.
5. Commits additionally hold the **per-bucket file lock** for the whole commit (`commit1stPhase` -> `lockFilesInOrder`, held through `commit2ndPhase`'s fold until `reset()`), so two commits to the *same* bucket serialize - no pure commit-vs-commit race on a bucket's counter. **But `LocalBucket.count()` takes no file lock** (it reads via `pageManager.getImmutablePage` / `transaction.getPage` only), so it does not participate in that serialization.

[...]

PR #5652, opened 2026-07-31
fix(engine) #5492: execute SQL statements against the replicated database instance

### LINKED ISSUE #5492: HA: a materialized view makes the Raft leader ship page versions followers never received - WALVersionGapException, non-converging snapshot-resync loop, lost committed writes

robfrank:
Fixes #5492.

## Root cause

`SQLQueryEngine` executed statements against the raw `DatabaseInternal` it was constructed with. On an HA leader, `RaftReplicatedDatabase.command()` delegates to `proxied.command(...)`, so `CommandContext.getDatabase()` resolved to the inner `LocalDatabase` - and every `commit()` a statement issues *itself* went straight to `LocalDatabase.commit()`, bypassing Raft.

`TRUNCATE TYPE` deletes in batches of `TRUNCATE_BATCH_SIZE` (default 1000), committing after each batch. Truncating more than one batch therefore published pages on the leader that no follower ever received. Follower page versions trail by exactly those writes, and the next replicated entry touching one of those pages fails its version check:

```
PageId(playwithpictures/142/0) version in WAL (2205) does not match with existent version (2203)
```

`WALVersionGapException` -> database marked diverged -> snapshot resync -> the next entry breaks the same way.

`SQLScriptQueryEngine` has **always** resolved `getWrappedDatabaseInstance()` before executing the very same statement classes. That is why an identical `TRUNCATE` replicates correctly under `sqlscript` and not under `sql`. The asymmetry was the defect.

## Why a materialized view was needed to see it

`MaterializedViewRefresher` truncates and rebuilds the backing type on every source commit, via `database.command("sql", "TRUNCATE TYPE ...")`. A view over 3000 rows crosses the 1000-record batch boundary on every refresh.

It also explains why this **never reproduced in process**: earlier harnesses truncated 180 to 450 records, so `count % 1000 == 0` never held and no mid-statement commit ever ran. Not a concurrency heisenbug, a threshold.

The bug is not view-specific. Any `TRUNCATE TYPE` over 1000 records issued through `/api/v1/command` with `language=sql` on an HA leader silently fails to replicate its first N-1 batches.

## Fix

One line of behavior, at the single point where the execution database is chosen: `SQLQueryEngine` now resolves `getWrappedDatabaseInstance()`, matching `SQLScriptQueryEngine`. Off HA the wrapper is the instance itself, so nothing changes there. The `query()` paths are deliberately untouched - they are gated on `isIdempotent()` and never commit.

## Evidence: 3-node container A/B

Image built from this branch's parent (`726633d3c`, containing both #5502 and #5537), same load, only the schema differing.

| Signal, per follower | With the view | Control (HTTP and GRPC) |
|---|---|---|
| Page version gaps | 24002 / 23924 | 0 |
| Gap-triggered resyncs | 357 / 356 | 0 |
| Snapshots served by leader | 712 | 0 |
| Scanned `Photo` per node | 30000 / 30000 / **29991** | 30000 / 30000 / 30000 |
| `UserStats` rows per node | 3000 / 3000 / **0** | n/a |
| Load phase | ~4.8 min | ~1.5 min |

First gap byte-identical on both followers 1 ms apart, on the view's own backing bucket - the leader-side shape the issue predicted.

Record loss is confirmed by `count(@rid)` scan, not `countType()`, whose cached counter drifts independently (#5152/#5154).

## Test plan

- [x] `Issue5492TruncateBatchNotReplicatedIT` - 2-node cluster, batch size lowered, `TRUNCATE` issued exactly as the refresher issues it. **Verified failing before the fix**: `[follower must see no WAL version gap] but was: 1`. Passes after.
- [x] Asserts the gap counter *before* querying the follower: the resync reinstalls the follower's database, which otherwise surfaces as an unrelated-looking `DatabaseIsClosed`.
- [x] `ThreeNodesLoadTestIT` now carries both arms of the A/B, so the comparison is reproducible rather than a one-off local run. The materialized view that was sitting commented out in `DatabaseWrapper` is restored and wired to the MV arm only.
- [ ] Nightly load-tests workflow dispatched against this branch (link in a comment below).

## Harness changes worth a look

- `ContainersTestTemplate.IMAGE` is overridable via `-Darcadedb.test.image`, so a run names the build it tested.
- Container stdout is dumped to `target/container-logs/` before teardown - testcontainers removes the containers, and without this the raw log cannot be consulted afterwards.
- Counters are summed over the per-test registry rather than read off `Metrics.globalRegistry`: a meter on the composite reports one arbitrarily chosen child, and this class adds/removes a backing registry around every test.
- The fixed 5-second sleep waiting for schema replication is now a measured wait. It was too tight (schema needs ~8.5 s here) and failed the first MV run before it reached the load phase.

**Reviewer note:** the judgment call is the blast radius of `executionDatabase()`. Every `sql` statement now sees the wrapper. The argument that this is safe is that `sqlscript` already runs these same statement classes with the wrapper today; if you think some statement depends on receiving the inner instance, say which and I will narrow the fix to the batching sites (`TruncateTypeStatement`, `TruncateBucketStatement`, `RebuildTypeStatement`, `BatchStep`).

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
@@ -112,7 +112,7 @@ public class SQLQueryEngine implements QueryEngine {
     context.setInputParameters(parameters);
     context.setConfiguration(configuration);
 
-    return statement.execute(database, parameters, context);
+    return statement.execute(executionDatabase(), parameters, context);
   }
 
   @Override
@@ -121,7 +121,25 @@ public class SQLQueryEngine implements QueryEngine {
     statement.setLimit(new Limit(JJTLIMIT).setValue((int) database.getResultSetLimit()));
     final CommandContext context = new BasicCommandContext();
     context.setConfiguration(configuration);
-    return statement.execute(database, parameters, context);
+    return statement.execute(executionDatabase(), parameters, context);
+  }
+
+  /**
+   * The database a statement executes against, and therefore the one {@code CommandContext.getDatabase()} returns.
+   * <p>
+   * Statements that commit mid-execution - {@code TRUNCATE TYPE}/{@code BUCKET} batching every
+   * {@link com.arcadedb.GlobalConfiguration#TRUNCATE_BATCH_SIZE} records, {@code REBUILD INDEX}, {@code BatchStep} -
+   * call {@code commit()} on whatever this returns. Handing them the raw instance means those commits go straight to
+   * {@code LocalDatabase.commit()}, which on an HA leader applies the pages locally and never proposes them to Raft:
+   * followers then trail by exactly those page versions and the next replicated entry touching one of them fails the
+   * version check (#5492).
+   * <p>
+   * {@code SQLScriptQueryEngine} has always resolved the wrapper for the same reason, which is why the identical
+   * statements replicate correctly under {@code sqlscript} and not under {@code sql}. Off HA the wrapper is the
+   * instance itself, so this is a no-op there.
+   */
+  private DatabaseInternal executionDatabase() {
+    return database.getWrappedDatabaseInstance();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
@@ -137,6 +137,11 @@ public class SQLQueryEngine implements QueryEngine {
    * {@code SQLScriptQueryEngine} has always resolved the wrapper for the same reason, which is why the identical
    * statements replicate correctly under {@code sqlscript} and not under {@code sql}. Off HA the wrapper is the
    * instance itself, so this is a no-op there.
+   * <p>
+   * Every path in this class that hands a database to {@code Statement.execute} must use this, not the field: both
+   * {@code command()} overloads and {@code analyze()}'s {@code AnalyzedQuery.execute()}. The engine itself is bound
+   * to the inner instance ({@code RaftReplicatedDatabase.getQueryEngine} delegates to {@code proxied}), so the field
+   * is never the right answer for execution.
    */
   private DatabaseInternal executionDatabase() {
     return database.getWrappedDatabaseInstance();
@@ -166,7 +171,11 @@ public class SQLQueryEngine implements QueryEngine {
         final long resultSetLimit = database.getResultSetLimit();
         if (resultSetLimit > 0)
           statement.setLimit(new Limit(JJTLIMIT).setValue((int) resultSetLimit));
-        return statement.execute(database, parameters);
+        // Same resolution as command(): this is a third execution entry point, not an analysis-only one.
+        // The MCP command tool runs SQL exclusively through here (analyze() once, then execute(); the
+        // database.command() fallback is only reached by engines whose execute() returns null), so leaving
+        // the raw instance here would keep #5492 open for that caller alone.
+        return statement.execute(executionDatabase(), parameters);
       }
     };
   }
--- a/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
@@ -138,10 +138,17 @@ public class SQLQueryEngine implements QueryEngine {
    * statements replicate correctly under {@code sqlscript} and not under {@code sql}. Off HA the wrapper is the
    * instance itself, so this is a no-op there.
    * <p>
-   * Every path in this class that hands a database to {@code Statement.execute} must use this, not the field: both
-   * {@code command()} overloads and {@code analyze()}'s {@code AnalyzedQuery.execute()}. The engine itself is bound
-   * to the inner instance ({@code RaftReplicatedDatabase.getQueryEngine} delegates to {@code proxied}), so the field
-   * is never the right answer for execution.
+   * <b>Which paths must use this:</b> every one that can carry a statement that commits - both {@code command()}
+   * overloads and {@code analyze()}'s {@code AnalyzedQuery.execute()}. {@code analyze()} is included because it is
+   * not gated on idempotency: the MCP command tool executes writes through it, so it can carry a {@code TRUNCATE}.
+   * <p>
+   * The two {@code query()} overloads deliberately keep the field. They throw
+   * {@link com.arcadedb.exception.QueryNotIdempotentException} before execution for anything non-idempotent, so no
+   * statement reaching them can commit, and leaving them alone keeps read traffic - including follower-local reads -
+   * resolving exactly as it did before. Add a new entry point that can execute a write, and it belongs on this
+   * method; the engine itself is bound to the inner instance
+   * ({@code RaftReplicatedDatabase.getQueryEngine} delegates to {@code proxied}), so the field is never the right
+   * answer for anything that might commit.
    */
   private DatabaseInternal executionDatabase() {
     return database.getWrappedDatabaseInstance();
--- a/ha-raft/CLAUDE.md
+++ b/ha-raft/CLAUDE.md
@@ -34,17 +34,17 @@ Each has its own local `excludeId` derivation (`leaderId != null ? leaderId : lo
 
 **Any change to what "replica" means must be applied to all three.** Grep `RaftHAServer.java` for `getPeers()` before claiming such a fix is complete. This has already caused one incomplete fix that review caught.
 
-## Known sharp edge: materialized views break page-version replication
+## Anything that commits must hold the WRAPPED database instance, not the inner one
 
-Unresolved as of 2026-07, tracked in #5492 - delete this section when that closes. Established by a controlled A/B run on a 3-node cluster, identical load, only the schema differing:
+`LocalDatabase.commit()` writes pages locally. `RaftReplicatedDatabase.commit()` proposes them to Raft and *then* writes them. Code holding the wrong one of the two commits successfully, applies its pages on the leader, and replicates nothing. Followers trail by exactly those page versions, and the next replicated entry touching one of them fails its version check - `WALVersionGapException`, database marked diverged, snapshot resync, and the entry after it breaks the same way. Nothing throws at the point of the mistake, so the symptom always surfaces somewhere else.
 
-- **With** a `REFRESH INCREMENTAL` materialized view: 2041/3000 writes fail, 12 snapshot-resync cycles in ~9 s, nodes never converge.
-- **Without** it: 3000/3000 writes, zero errors, immediate convergence.
+`getWrappedDatabaseInstance()` returns the Raft wrapper under HA and the instance itself off it, so resolving it is always correct and never costs anything.
 
-The leader ships a `TX_ENTRY` whose page version is ahead of what followers hold, first appearing on the view's own backing bucket. Followers throw `WALVersionGapException`, trigger a full snapshot resync, and the resync does not repair it - the next entry gaps again on a different file. Meanwhile every request to a resyncing follower is deflected with a 503, so writes routed through it are dropped. The end state is lost committed writes, not split-brain.
+This was #5492. `SQLQueryEngine` handed statements the raw instance, so `CommandContext.getDatabase()` was the inner `LocalDatabase`, and every statement that commits mid-execution bypassed replication: `TRUNCATE TYPE`/`BUCKET` batching every `TRUNCATE_BATCH_SIZE` (default **1000**) records, `REBUILD INDEX`, `BatchStep`. `SQLScriptQueryEngine` had always resolved the wrapper before running the *same statement classes*, which is why an identical `TRUNCATE` replicated under `sqlscript` and not under `sql`.
 
-The exact leader-side path is **not** proven, and it has never been reproduced in-process: a 3-node harness on `BaseRaftHATest` sees zero gaps with writes on the leader, writes through a follower, with and without a unique index on the source type. The container A/B is the only known reproduction. Two theories are ruled out and should not be re-proposed: `LocalSchema` is constructed with `wrappedDatabaseInstance`, so the refresh does go through `RaftReplicatedDatabase` rather than committing on the inner `LocalDatabase`; and #5503 (concurrent writes on a shared follower handle) is a different failure mode - `Invalid record size ... deleting record`, no version gap - whose control arm was clean.
+Two things about how it hid, both worth generalizing:
 
-One leader-side hole of exactly this shape **has** been found and closed: `recordFileChanges` drained the buffered WAL of commits that ran inside the DDL callback but omitted it from the send guard, so a callback that wrote records without creating a file or moving the schema version had its pages applied on the leader and never shipped. That is fixed, and `applySchemaEntry` now recognizes the resulting WAL-only entry and skips the schema reload for it, as it already did for sealed-only TimeSeries entries. It did not close the materialized-view A/B, so the section stays.
+- **A default-sized threshold can look like a heisenbug.** It reproduced only above 1000 records in one statement, so every in-process attempt (180 to 450 records) saw a clean cluster and the failure was written off as needing container-level concurrency. If a replication bug reproduces in containers and not in process, compare the *sizes* before theorizing about timing.
+- **A materialized view was the trigger only because it truncates constantly.** `MaterializedViewRefresher` truncates and rebuilds its backing type on every source commit, so a view over 1000+ rows crossed the boundary on every refresh. The view was the amplifier, not the defect - `TRUNCATE TYPE` over 1000 records through `/api/v1/command` with `language=sql` was enough on its own.
 
-If you are investigating unexplained replication gaps, check whether the failing database has a materialized view before going further.
+When adding any code path that commits, ask which instance it holds. `Issue5492TruncateBatchNotReplicatedIT` guards the statement path; it asserts the WAL gap counter *before* querying the follower, because the resync reinstalls the follower's database and a stale handle then throws `DatabaseIsClosed`, which reads like infrastructure noise rather than divergence.
--- a/ha-raft/src/test/java/com/arcadedb/server/ha/raft/Issue5492TruncateBatchNotReplicatedIT.java
+++ b/ha-raft/src/test/java/com/arcadedb/server/ha/raft/Issue5492TruncateBatchNotReplicatedIT.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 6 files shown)


robfrank:
Load-tests workflow dispatched against this branch: https://github.com/ArcadeData/arcadedb/actions/runs/30666434001

It builds the image from this branch with `-Pdocker` and then runs `load-tests` with `-Pintegration`, so `ThreeNodesLoadTestIT` runs both arms of the A/B against the fix.

One caveat on reading that run: the workflow invokes Maven with `--fail-never`, so the job goes green even when a test fails. The verdict is in the test-reporter output and in the `#5492 signatures` block the test logs per arm, not in the job's status badge. What to look for, with the view present:

```
Total page version gaps: 0
Total gap-triggered resyncs: 0
Total snapshots served: 0
Scanned Photos per node: 30000 / 30000 / 30000
UserStats rows per node: 3000 / 3000 / 3000
```

Against the pre-fix numbers on the same harness: 24002 and 23924 gaps, 357 resyncs, 712 snapshots served, one node 9 records short and its view empty.

robfrank:
## CI load-tests: the A/B is now clean on both arms

Run https://github.com/ArcadeData/arcadedb/actions/runs/30666434001, image built from this branch. `ThreeNodesLoadTestIT` 3/3 passed, module total 14 tests / 0 failures / 0 errors.

With the materialized view **present**:

```
=== #5492 signatures (materialized view: PRESENT) ===
Total page version gaps: 0
Total gap-triggered resyncs: 0
Total resyncs completed: 0
Total snapshots served: 0
Failed user writes: 0.0
Failed photo writes: 0.0
Scanned Users per node: 3000 / 3000 / 3000
Scanned Photos per node: 30000 / 30000 / 30000
UserStats rows per node: 3000 / 3000 / 3000
```

Side by side with the same harness before the fix:

| Signal | Before | After |
|---|---|---|
| Page version gaps (per follower) | 24002 / 23924 | **0** |
| Gap-triggered resyncs | 357 / 356 | **0** |
| Snapshots served by leader | 712 | **0** |
| Scanned `Photo` per node | 30000 / 30000 / 29991 | **30000 / 30000 / 30000** |
| `UserStats` rows per node | 3000 / 3000 / 0 | **3000 / 3000 / 3000** |

The view now converges on all three nodes, which it never did before - the follower's copy was empty or a third full.

Both control arms remain clean, so the fix did not achieve this by disabling anything.

## Regression sweep

- `engine` full suite: **10545 tests, 0 failures, 0 errors** (the module the fix touches, and where every `sql` statement now resolves the wrapper).
- `Issue5492TruncateBatchNotReplicatedIT`: passes, and confirmed failing before the fix.

Also pushed: `ha-raft/CLAUDE.md` carried a *"delete this section when that closes"* marker for #5492. It now states the general rule instead - code that commits while holding the inner `LocalDatabase` replicates nothing and throws nowhere near the mistake.

robfrank:
Both findings were real. Verified each against the code before acting, and fixed both.

## The incomplete fix: confirmed and closed

The claim checks out end to end:

- `SQLQueryEngine.analyze()` returns an `AnalyzedQuery` whose `execute()` called `statement.execute(database, ...)` with the raw field.
- `ExecuteCommandTool.java:83,95-96` analyzes once and executes through it, falling back to `database.command()` only when `execute()` returns null - which SQL never does.
- `RaftReplicatedDatabase.getQueryEngine()` (line 1197) delegates to `proxied`, so the engine is bound to the inner instance and that path resolved the inner database again.

So a `TRUNCATE TYPE` over `TRUNCATE_BATCH_SIZE` records issued through MCP on a leader still reproduced #5492 with every other caller fixed. You were right that this, not blast radius, was the real completeness risk.

Fixed by routing it through the same helper, and `truncateBatchCommitsReachFollowersViaAnalyzedQuery` now drives the tool's exact shape (resolve engine, analyze, execute). **Verified failing with only that path reverted:** `[no WAL version gap when the truncate runs through analyze().execute()] but was: 1`. Both tests pass with the full fix.

The helper's javadoc now states the rule directly - every path in the class that hands a database to `Statement.execute` must use it, and the field is never the right answer for execution - so the next person adding an entry point has it in front of them.

## One thing your finding surfaced that I would have missed

Adding the second test made **both** tests fail, while each passed alone. They shared one type name, so the first test's divergence reached the second. Each test now uses its own type. Worth flagging because that coupling would have turned a future real regression into an unreadable double failure.

## Orphaned javadoc: confirmed

Exactly as described - blocks at 699, 708 and 719 with methods at 727, 747 and 751, so only `dumpContainerLogs` was documented. Each block now sits above its own method. While there I made `countInContainerLogs` say why a marker has to come from the logged text: my first version of this harness grepped `WALVersionGapException` and `snapshot resync required`, neither of which the server ever logs, and reported a clean run against a log holding 24002 gaps.

## Unrelated red I checked rather than assumed

A local `ha-raft` run failed 3 unit tests (`SnapshotInstallerRetryTest`, `WaitForApplyTest` x2, the latter `DatabaseAreNotIdentical` in teardown). Not caused by this change: the engine tree in that run was byte-identical to an earlier run that passed 772/772, with only test files differing. Consistent with the known teardown flakiness in this suite.

robfrank:
Note on the red `Meterian client scan`: it fails on `security: 0 (minimum: 90)`, a dependency score. This PR changes no dependencies, and I checked rather than assumed - the Meterian Scanner workflow has failed on each of the **5 most recent `main` commits** (runs 30665022206, 30664991823, 30662379739, 30662341641, 30657263390). Pre-existing on the base branch, not introduced here.

Everything else is green: Codacy 0 new issues, CodeQL across all languages, e2e suites for java/js/go/python/csharp, builder-tests, build-and-package.

robfrank:
Both actionable observations applied; the third needed no change and I left it alone.

**Repeated `getLogs()` fetches.** Correct, and worth fixing beyond the cost: `countInContainerLogs` now takes varargs and scans each container's stdout once for every marker, with the single-marker overload delegating. The four markers in `logReplicationSignatures` are now one call instead of four full re-fetches.

**Inconsistent container labels.** Fixed via a shared `containerLabel()` helper, so the dumped file name and the log line agree.

**`TRUNCATE_BATCH_SIZE` not restored.** Agreed it is correct as written, since `SCOPE.DATABASE` and `@BeforeEach` builds a fresh cluster per test. Left as-is rather than adding restore logic that would imply a sharing hazard that does not exist.

## One thing your first observation led me to that is worth more than the fix

Rewriting the scan made me notice the counters cannot distinguish success from a broken instrument: all four read zero when the cluster is healthy **and** when the marker no longer matches what the servers log. That is not hypothetical here - the first version of this harness grepped `WALVersionGapException` and `snapshot resync required`, neither of which is ever logged, and I reported *"0 gaps, 0 resyncs, symptom did not reproduce"* against a log holding 24002 gaps.

So the scan now carries a canary: `ArcadeDB Server started`, which appears exactly once per container in every run. If it comes back short, the run logs an error saying the counters are unmeasured rather than zero. A future run that silently stops matching now announces itself instead of reading as a pass.

Load tests re-dispatched to exercise the refactored scan end to end: https://github.com/ArcadeData/arcadedb/actions/runs/30669867572

robfrank:
You're right, and the contradiction was mine to fix rather than the code's.

I considered your first option - routing `query()` through `executionDatabase()` too - and deliberately did not take it. It would make the sentence true by widening the change, and read traffic is the one thing this PR had not been touching: a follower-local read resolving through the wrapper is a behavior change I have no evidence for and no test covering. Making the doc accurate is the smaller and more honest move.

The javadoc now states the invariant that is actually true:

> **Which paths must use this:** every one that can carry a statement that commits - both `command()` overloads and `analyze()`'s `AnalyzedQuery.execute()`. `analyze()` is included because it is not gated on idempotency: the MCP command tool executes writes through it, so it can carry a `TRUNCATE`.
>
> The two `query()` overloads deliberately keep the field. They throw `QueryNotIdempotentException` before execution for anything non-idempotent, so no statement reaching them can commit [...]

That also answers *why* the two read-serving paths are treated differently, which was the thing a future reader would otherwise have to reverse-engineer: `query()` is gated on idempotency, `analyze()` is not.

Doc-only, no behavior change.

**On your observation 2** - that `analyze().execute()` now routes MCP reads (`QueryTool`) through the wrapper: acknowledged and I agree it is harmless, since reads do not commit and delegate straight through. I chose not to make it conditional on `isIdempotent()`. It would restore the old instance for reads, but it would also mean a `SELECT` and a `TRUNCATE` arriving at the same entry point get different databases, and that is a subtler thing to reason about later than a uniform rule. Flagging the trade-off explicitly in case you would rather have the narrower blast radius than the simpler rule - it is a one-line change if so.

robfrank:
Checked the Cypher observation against the code rather than taking it on trust, and it holds up. Filed as **#5655**.

`OpenCypherQueryEngine` holds `private final DatabaseInternal database` (line 83), and on an HA leader that is the inner `LocalDatabase` for exactly the reason this PR documents: `RaftReplicatedDatabase.getQueryEngine()` (line 1197) delegates to `proxied`. The explicit-transaction path then commits on it directly:

```java
// OpenCypherQueryEngine.java:685-693
database.begin(level);
...
case COMMIT:
  database.commit();   // LocalDatabase.commit() on an HA leader: applies pages, replicates nothing
```

`context.setDatabase(database)` at line 738 passes the same inner instance into statement execution, so it is not only the explicit-transaction path.

I filed it as **unverified** rather than confirmed, because I have not reproduced it - the code shape matches #5492 exactly, but the practical exposure depends on whether a Cypher explicit transaction spans the wrapper over HTTP, and on whether autocommit Cypher writes are already covered by the wrapper's own commit finalization. The issue says so plainly and carries the suggested test shape, including asserting the gap counter before querying the follower.

Agreed it stays out of this PR: it needs its own reproduction and its own regression test, and folding it in would mean shipping a Raft-path change with no test proving it was ever broken.

## On the two nits

- `Metrics.globalRegistry.getMeters()` dump coexisting with `sumCounter`: intentional. The dump is a human-readable inventory; the summation is the number anything depends on. Only the latter had to move off the composite.
- `awaitCountOn` catching `RuntimeException` broadly: agreed on the trade-off. It exists specifically so a mid-resync `DatabaseIsClosed` does not abort the poll, and the assertion that follows still fails on the real value. A genuine query bug would surface as the 30 s timeout with the last observed count, which is why the method returns that count rather than throwing.

robfrank:
## Load-tests rerun: the scan self-check works, so the zeros mean something

Run https://github.com/ArcadeData/arcadedb/actions/runs/30669867572, `ThreeNodesLoadTestIT` 3/3, module 14 tests / 0 failures.

```
=== #5492 signatures (materialized view: PRESENT) ===
Total page version gaps: 0
Total gap-triggered resyncs: 0
Total resyncs completed: 0
Total snapshots served: 0
Log scan self-check: canary 'ArcadeDB Server started' found 3 times across 3 containers
Scanned Users per node: 3000 / 3000 / 3000
Scanned Photos per node: 30000 / 30000 / 30000
UserStats rows per node: 3000 / 3000 / 3000
```

The canary line is the point. It reports 3 of 3 containers in all three arms, so the four zeros above it are **measured** rather than an instrument that stopped matching. Without it, this output would be byte-identical to the run where I wrongly reported the symptom as gone.

Same result on both control arms.

A final run is in flight (https://github.com/ArcadeData/arcadedb/actions/runs/30670959519) covering the two things pushed after this one started: the terminal assertions on scanned counts, and the MV arm's view-convergence guard. I will post its numbers - it exercises assertions that did not run above, so it is not a formality.

robfrank:
## Final load-tests run: green, including the assertions that had never executed

Run https://github.com/ArcadeData/arcadedb/actions/runs/30670959519 - `ThreeNodesLoadTestIT` 3/3, module 14 tests / 0 failures / 0 errors.

MV arm:

```
=== #5492 signatures (materialized view: PRESENT) ===
Total page version gaps: 0
Total gap-triggered resyncs: 0
Total resyncs completed: 0
Total snapshots served: 0
Log scan self-check: canary 'ArcadeDB Server started' found 3 times across 3 containers
Scanned Users per node: 3000 / 3000 / 3000
Scanned Photos per node: 30000 / 30000 / 30000
UserStats rows per node: 3000 / 3000 / 3000
UserStats convergence check: 3000 / 3000 / 3000
```

This run is the one that matters for the last review round, because it exercised code the previous runs did not:

- the terminal assertions now decide on `count(@rid)` scans per node rather than the drift-prone cached counter;
- the MV arm's view-convergence guard ran for the first time and passed - that is the assertion that would have failed the build on the pre-fix `3000 / 3000 / 0`;
- the per-container marker tallies confirm the single-pass scan counts identically to the four-call version it replaced.

## Summary of the whole change

| Signal, per follower | Before | After |
|---|---|---|
| Page version gaps | 24002 / 23924 | **0** |
| Gap-triggered resyncs | 357 / 356 | **0** |
| Snapshots served by leader | 712 | **0** |
| Scanned `Photo` per node | 30000 / 30000 / 29991 | **30000 x3** |
| `UserStats` rows per node | 3000 / 3000 / 0 | **3000 x3** |

Both control arms clean throughout, so nothing was achieved by disabling a code path.

Ready for a maintainer decision. The only red is `Meterian client scan`, failing identically on recent `main`.

One housekeeping note: the branch is still named `test/5492-...` from when this was scoped as a reproduction rather than a fix. Happy to rename before merge if that matters for the release notes.

Issue #5635, opened 2026-07-31
Index cursor follow-ups from #5601: LSMTreeIndexCursor.next() returns null after hasNext(), FetchFromIndexStep.reset() leaks its cursors


lvca:
Two follow-ups from #5601 / #5609. Independent of each other - split if either grows.

---

## 1. `LSMTreeIndexCursor.next()` can return `null` after `hasNext()` returned true

`hasNext()` is optimistic - it answers on how many underlying page cursors are still live, not on whether any of them still holds a surviving RID (`LSMTreeIndexCursor.java:337-339`):

```java
public boolean hasNext() {
  return validIterators > 0 || (currentValues != null && currentValueIndex < currentValues.length) || txCursor != null;
}
```

`next()` then does the work, and when a trailing run of tombstoned keys leaves nothing to emit it falls out of its loop and answers `null` (`:513`):

```java
return currentValues == null || currentValueIndex >= currentValues.length ? null : currentValues[currentValueIndex++];
```

That violates `Iterator`, and `IndexCursor extends Iterator<Identifiable>`, so `for (final Identifiable r : cursor)` over a delete-heavy index can hand a caller a `null` element.

This is not theoretical - the engine already works around it in several places, each with its own comment explaining the same thing:

- `FullTextQueryExecutor` guards it at four call sites (`:507`, `:537`, `:572`, `:706`), from #5118.
- `LSMTreeGeoIndex` guarded it before #5609, and `GeoIndexCursor.fetchNext()` still does.
- `MultiIndexCursor.next()` propagates it upward rather than absorbing it.

And it caused a user-visible defect: #5601's `countEntries()` residual was **entirely** this `null` being counted as an entry. The tell was that the residual survived a full compaction which had already dropped every tombstone.

**Why it was not fixed in #5609.** Making `hasNext()` exact needs a one-element prefetch, which moves *when* the in-transaction overlay is read (`getClosestEntryInTx`) and changes what `getRecord()` peeks at - it currently returns the *not yet consumed* value at `currentValueIndex`, which is already an odd contract. That is a real behavioural change on a class carrying #4944 (cursor parked on a full-key tombstone), #4947 (biased navigation keys), #5055 (tx overlay merge) and the DESC duplicate-group infinite loop. It deserves its own change with its own test surface rather than riding along in a geospatial PR.

**Suggested shape**, for discussion:

[...]

PR #5673, opened 2026-08-01
fix(engine) #5662: index cursors are AutoCloseable, and a leaked one no longer pins a retired index file

### LINKED ISSUE #5662: Index-cursor follow-ups from #5635: nothing enforces close(), DeleteFromIndexStep leaks and swallows, dead fetchNullKeys, vector iterator() is not this, MultiIndexCursor accessors prefetch
Five independent items found while doing #5635 and deliberately left out of PR #5641 to keep it focused. Split any of them out if it grows. All line references are against `main` at 91439951e.

---

### - [x] 1. Nothing enforces closing an index cursor

`Cursor` extends `Iterable`+`Iterator` but **not** `AutoCloseable`, even though `IndexCursor.close()` exists and matters:

```java
public interface Cursor extends Iterable<Identifiable>, Iterator<Identifiable> {
  long estimateSize();
}
```

An unclosed `LSMTreeIndexUnderlyingCompactedSeriesCursor` stays registered with its file, and `LSMTreeIndex.dropRetiredCompactedIndexes` skips a retired file that still has one — nothing else will ever close it, so the file survives until the next database restart.

This is the third change in a row fixing instances of the same leak:

- #5601 — `LSMTreeIndex.countEntries()`, `FetchFromIndexStep.close()`
- #5609 — `FetchFromIndexedFunctionStep`, `SQLFunctionGeoPredicate.GeoCandidateIterator`, `GeoIndexCursor`
- #5635 — six more: the `key IN [...]` per-value cursors (which nothing had *ever* closed, not even `close()`), `MaxMinFromIndexStep`, two full-text term walks in `FullTextQueryExecutor`, and the Cypher `NodeIndexSeek` / `NodeIndexRangeScan` operators, whose `close()` carried a `// IndexCursor doesn't need explicit closing` comment that had been wrong since #5601

They keep recurring because no compiler or IDE warning fires: an `IndexCursor` obtained, driven partway, and dropped looks exactly like correct code. Every site was found by reading, and there is no reason to believe the reading was exhaustive.

Making `Cursor extends AutoCloseable` would surface the remaining ones mechanically (an IDE flags a non-try-with-resources `AutoCloseable`, and static analysis has a rule for it), and it costs nothing at runtime — `close()` is already declared with a no-op default. Worth checking what it does to the call sites that legitimately hand a cursor onward rather than consuming it, since those would need `@SuppressWarnings` or a restructure.

---

### - [x] 2. `DeleteFromIndexStep` leaks its cursor and swallows the failure that would explain it

`engine/src/main/java/com/arcadedb/query/sql/executor/DeleteFromIndexStep.java`

Two problems in one method. There is no `close()` on the step at all, so the cursor opened by `init(condition)` is never released — item 1's leak, on the DELETE path.

And `init()` swallows the failure (`:110-115`):

```java
try {
  init(condition);
  nextEntry = loadNextEntry(context);
  e.printStackTrace();
}
```

A DELETE whose index scan failed to initialise prints a stack trace to stdout, leaves `cursor` null, and **continues** — the next `loadNextEntry()` NPEs on `cursor.hasNext()`, so the operator gets a `NullPointerException` instead of the `IOException` that says what actually went wrong. It should log through `LogManager` and raise a `CommandExecutionException` carrying the cause, matching how the rest of the executor package reports an I/O failure.

---

### - [x] 3. `FetchFromIndexStep.fetchNullKeys()` is dead code

`engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromIndexStep.java:374`

Nothing calls it. The only mentions of the name are the method itself and the comment in `processFlatIteration()` explaining why it must **not** be called (`:360`):

```java
// NOTE: Do NOT call fetchNullKeys() here.
// When NULL_STRATEGY is INDEX, NULL-keyed entries are already stored in the B-tree
// and will be returned by the cursor iterator. Calling fetchNullKeys() would cause
// duplicate entries to be returned since it creates a separate iterator for NULL keys.
```

It also builds an `IndexCursor` that it wraps in an anonymous `Iterator` and that nothing would ever close, so if it is ever revived it revives item 1 with it. Either delete it, or work out whether a `NULL_STRATEGY` other than `INDEX` still needs it and wire it with a release path.

---

### - [x] 4. `LSMVectorIndex`'s search cursor returns a foreign iterator from `iterator()`

`engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java` (the anonymous `IndexCursor` returned by the vector search)

```java
@Override
public Iterator<Identifiable> iterator() {
  return (Iterator<Identifiable>) (Iterator<?>) resultRIDs.iterator();
}
```

Every other `IndexCursor` — `LSMTreeIndexCursor`, `MultiIndexCursor`, `GeoIndexCursor`, `TempIndexCursor`, `EmptyIndexCursor` — returns `this`. Here a for-each gets a **second, independent** traversal that does not move the cursor's own `position`, so:

- `getRecord()` reports nothing meaningful during a for-each (it is derived from `position`, which never advances)
- mixing `next()` with a for-each reads some RIDs twice and is silently order-dependent

`IndexCursorCollection` has the same shape and may want the same treatment — worth checking it in the same pass.

---

### - [x] 5. `MultiIndexCursor.getComparator()` / `getBinaryKeyTypes()` probe `hasNext()` on their children

`engine/src/main/java/com/arcadedb/index/MultiIndexCursor.java:198,207`

```java
public BinaryComparator getComparator() {
  for (final IndexCursor cursor : cursors) {
    if (cursor != null && cursor.hasNext())
      return cursor.getComparator();
  }
  return null;
}
```

This was free while `hasNext()` was a field comparison. Since #5635 `LSMTreeIndexCursor.hasNext()` **prefetches** — it runs the merge until it holds an emittable RID — so two plain accessors can now block on page reads and tombstone-skip work.

No correctness problem (the prefetched entry is cached, so nothing is consumed or done twice) and no extra total work on a cursor that will be drained anyway, but it is a surprising thing for an accessor to do, and on a cursor that is opened and then only queried for its key types it is work that would otherwise never happen. Both should read from the first non-null child directly: a child's comparator and key types do not depend on whether it has anything left.

Introduced by #5635, so this one is a regression rather than a pre-existing wart.

lvca:
Closes #5662 — all five items, plus three defects the audit for item 1 turned up.

## 1. Nothing enforced closing an index cursor

Closing matters: a scan over an LSM index holds one underlying cursor per compacted series, each registered with its file, and `dropRetiredCompactedIndexes` will not drop a file that still has one. Draining a cursor releases those registrations as each series is exhausted, so **only a cursor abandoned partway leaks** — the shape a `LIMIT`, an `exists()` or an early `break` produces.

`Cursor` now extends `AutoCloseable`, with `close()` declared without a checked exception so try-with-resources does not force a `catch (Exception)` on the caller.

**Better than what the issue proposed.** `AutoCloseable` is a compile-time signal, and a compile-time signal says nothing about the sites nobody has read yet — the issue's own point was that the reading was probably not exhaustive. So the retire guard was also changed from an `AtomicInteger` to **weak references** to the live cursors: a cursor abandoned without `close()` stops pinning its file once the GC reaches it, and the reclaim is logged with the index name. A missed `close()` used to be permanent — the file stayed undroppable for the lifetime of the database with nothing left in the process that could ever release it, and nothing to say so. The leak becomes bounded *and* observable instead of silent and forever.

Sites found in this pass and now releasing their cursor:

- **`SelectExecutor.executeExists()` / `executeCount()`** — `exists()` returns on the first match, `count()` stops at its `LIMIT`; both abandoned the scan.
- **`SelectIterator.close()`** — documented itself as having "nothing to release in the serial implementation" while holding the source index cursor. `SelectParallelIterator` now chains to it (no producer can be reading that cursor: producers are scheduled only for a `MultiIterator` source).
- **`SubQueryStep`** *(not in the issue)* — never closed the plan it wraps, so closing a result set reached only the outer plan's steps. A `DELETE ... WHERE` is built exactly this way, with the index scan in the sub-plan and the `LIMIT` outside it.
- `TypeIndex`, the commit-time unique-key check, the edge-upsert lookup, the full-text scoring/explain/more-like-this walks, and the Gremlin index-filter step.

## 2. `DeleteFromIndexStep`

`close()` added. `init()` now logs through `LogManager` and raises a `CommandExecutionException` carrying the cause instead of `e.printStackTrace()` + carrying on with a null cursor. **Note:** no branch of `init(BooleanExpression)` actually throws `IOException` today — the catch exists because the method declares it — so that half is about what happens the day one does, not a failure reachable now. Also removed *(not in the issue)*: the equality branch opened a `range()` cursor and overwrote it with the `iterator()` one on the very next line — a dead store that opened and abandoned an extra full scan.

## 3. `FetchFromIndexStep.fetchNullKeys()`

Deleted, along with the `nullKeyIterator` field and the branch that read it (only ever assigned an empty iterator). No `NULL_STRATEGY` needs a second pass: with `INDEX` the entries are already in the B-tree at their sorted position, with `SKIP` they were never indexed, with `ERROR` they could not be inserted. The reasoning is kept as a comment on `processFlatIteration()`.

## 4. A cursor iterates itself

`LSMVectorIndex`'s search cursor and `IndexCursorCollection` now return `this` from `iterator()`, like every other implementation. Handing back the backing iterator gave a for-each a traversal that bypassed `next()`, so `getRecord()` stayed stale for the whole loop and (in the vector case) mixing the two read some RIDs twice.

## 5. `MultiIndexCursor` accessors

`getComparator()` / `getBinaryKeyTypes()` answer from state sampled at construction instead of probing the children for one that still has something left — which since #5635 means page reads and tombstone skips inside a plain accessor. The key types also no longer come back `null` once the cursor is exhausted. The constructors additionally close the children they already opened if construction fails partway, since nothing else holds a reference to them.

## Tests

`Issue5662IndexCursorFollowUpsTest` (10 tests) and `DeleteFromIndexStepCloseTest` (2). The observable for the leak tests is `LSMTreeIndexCompacted.getActiveCursors()` over a fixture forced to compact into a multi-page series; each asserts the counter can actually rise above zero first, so "it went back to zero" cannot hold vacuously. Every test was mutation-checked: each fix was reverted individually and the corresponding test observed to fail.

Engine (`index`, `select`, `query.sql`, `opencypher`, `function`, `database`, `graph`, `schema`, `engine`) and gremlin suites pass.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1119,4 +1119,50 @@ and expressing that would have been faithful to its intent, but it would have ti
 of repairing a config that enforced nothing - a decision worth taking on its own merits. Development dependencies are
 still covered by the `npm audit --audit-level=moderate` step of the same workflow.
 
+## An index cursor is `AutoCloseable`, and a leaked one no longer pins a retired index file forever
+
+Three releases in a row have fixed instances of the same defect - an `IndexCursor` obtained, driven partway and
+dropped ([#5601](https://github.com/ArcadeData/arcadedb/issues/5601),
+[#5609](https://github.com/ArcadeData/arcadedb/issues/5609),
+[#5635](https://github.com/ArcadeData/arcadedb/issues/5635)). Closing one matters: a scan over an LSM index holds one
+underlying cursor per compacted series, each registered with its file, and `dropRetiredCompactedIndexes` will not
+physically drop a file that still has one. Draining a cursor releases those registrations as each series is exhausted,
+so only a cursor abandoned **partway** leaks - which is exactly the shape a `LIMIT`, an `exists()` or an early `break`
+produces. They kept recurring because nothing fired: an abandoned cursor looks exactly like correct code, and every
+site so far was found by reading ([#5662](https://github.com/ArcadeData/arcadedb/issues/5662)).
+
+`Cursor` now extends `AutoCloseable`, so an abandoned one is a static-analysis and IDE finding rather than something
+only a careful read can catch, and a cursor can be used with try-with-resources. `close()` is declared without a
+checked exception, so a `try (final IndexCursor c = index.iterator(true))` does not force the caller into a
+`catch (Exception)`.
+
+That is a compile-time signal, and a compile-time signal proves nothing about the sites nobody has read yet. So the
+counter behind the retire guard was replaced with **weak** references to the live cursors: a cursor abandoned without
+`close()` stops pinning its file once the garbage collector reaches it, and the reclaim is logged with the index name.
+A missed `close()` used to be permanent - the retired file stayed on disk for the lifetime of the database with
+nothing left in the process that could ever release it, and nothing to say so.
+
+The call sites found in this pass and now releasing their cursor:
+
+- The native `Select` API. `exists()` returns on the first match and `count()` stops at its `LIMIT`, both abandoning
+  the scan, and `SelectIterator.close()` documented itself as having "nothing to release in the serial
+  implementation" while holding the source index cursor.
+- `SubQueryStep` never closed the plan it wraps, so closing a result set reached only the outer plan's steps. A
+  `DELETE ... WHERE` is built exactly this way - the index scan inside the sub-plan, the `LIMIT` outside it - so its
+  scan was released only when it happened to run to exhaustion.
+- `DeleteFromIndexStep` had no `close()` at all. Its equality branch also opened a `range()` cursor and overwrote it
+  with the `iterator()` one on the next line, a dead store that opened and abandoned an extra full scan; and an
+  `IOException` out of its initialisation printed a stack trace to stdout and carried on with a null cursor, so the
+  caller would have seen a `NullPointerException` instead of the failure that explained it.
+- The Gremlin index-filter step, `TypeIndex`, the unique-key check at commit time, the edge upsert lookup, and the
+  full-text scoring, explain and more-like-this walks.
+
+Two smaller corrections travel with them. The vector search cursor returned the backing list's own iterator from
+`iterator()`, so a for-each got a second, independent traversal that never moved the cursor's position - `getRecord()`
+reported nothing during the loop, and mixing it with `next()` read some RIDs twice; it now iterates itself, like every
+other cursor, and `IndexCursorCollection` does the same. And `MultiIndexCursor.getComparator()` /
+`getBinaryKeyTypes()` answered by probing the children for one that still had something left, which since #5635 means
+running page reads and tombstone skips inside two plain accessors; both now answer from state sampled at construction,
+and the key types no longer come back null once the cursor is exhausted.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1150,10 +1150,13 @@ The call sites found in this pass and now releasing their cursor:
 - `SubQueryStep` never closed the plan it wraps, so closing a result set reached only the outer plan's steps. A
   `DELETE ... WHERE` is built exactly this way - the index scan inside the sub-plan, the `LIMIT` outside it - so its
   scan was released only when it happened to run to exhaustion.
-- `DeleteFromIndexStep` had no `close()` at all. Its equality branch also opened a `range()` cursor and overwrote it
-  with the `iterator()` one on the next line, a dead store that opened and abandoned an extra full scan; and an
-  `IOException` out of its initialisation printed a stack trace to stdout and carried on with a null cursor, so the
-  caller would have seen a `NullPointerException` instead of the failure that explained it.
+- `DeleteFromIndexStep` had no `close()` at all. Two smaller repairs travel with it, neither of which changes what a
+  working statement does: an `IOException` out of its initialisation printed a stack trace to stdout and carried on
+  with a null cursor, so the caller would have seen a `NullPointerException` instead of the failure that explained it;
+  and the branch for an index without ordered iterations opened a `range()` cursor and overwrote it with the
+  `iterator()` one on the next line - it could never have worked, since both of those calls raise
+  `UnsupportedOperationException` on the only index that reaches it, so it was removed in favour of the error that
+  names the condition.
 - The Gremlin index-filter step, `TypeIndex`, the unique-key check at commit time, the edge upsert lookup, and the
   full-text scoring, explain and more-like-this walks.
 
--- a/engine/src/main/java/com/arcadedb/database/Cursor.java
+++ b/engine/src/main/java/com/arcadedb/database/Cursor.java
@@ -44,7 +44,9 @@ public interface Cursor extends Iterable<Identifiable>, Iterator<Identifiable>, 
 
   /**
    * Releases whatever the cursor holds. The default is a no-op for the implementations backed by an in-memory
-   * collection, which hold nothing.
+   * collection, which hold nothing - so an implementation that DOES hold something must remember to override it. The
+   * weak-reference net in {@code LSMTreeIndexCompacted} covers only the compacted-series cursors; nothing catches a
+   * new cursor type that acquires a resource and inherits this default.
    */
   @Override
   default void close() {
--- a/engine/src/main/java/com/arcadedb/database/Cursor.java
+++ b/engine/src/main/java/com/arcadedb/database/Cursor.java
@@ -24,10 +24,30 @@ import java.util.Iterator;
 
 /**
  * Cursor to browse a result set.
+ * <p>
+ * <b>Closing (#5662).</b> A cursor is {@link AutoCloseable} so that an abandoned one is a static-analysis finding
+ * rather than something only a careful read can spot. Closing matters: a scan over an LSM index holds one underlying
+ * cursor per compacted series, each registered with its file, and
+ * {@code LSMTreeIndex.dropRetiredCompactedIndexes} will not physically drop a file that still has one. Draining a
+ * cursor releases those registrations as each series is exhausted, so only a cursor abandoned <i>partway</i> leaks -
+ * which is exactly the shape a {@code LIMIT}, an {@code exists()} or an early {@code break} produces. Prefer
+ * try-with-resources; {@link #close()} is idempotent on every implementation.
+ * <p>
+ * {@code close()} is declared without a checked exception (unlike {@link AutoCloseable#close()}) so that a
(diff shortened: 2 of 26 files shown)


lvca:
Thanks — point 1 was right, and checking it turned out to be worse than described.

**1. The equality-branch test.** Correct on both counts, and the branch is deader than the review says. It sits behind `!index.supportsOrderedIterations()`, and the only `RangeIndex` that answers false is a `TypeIndex` over a non-ordered bucket index (`LSMTreeIndex` is hardcoded `true`; full-text/geo/vector/hash/sparse aren't `RangeIndex` at all). On such a `TypeIndex`, `iterator(asc, keys, incl)` raises `UnsupportedOperationException` exactly as `range()` does — so the branch could never have succeeded whichever line ran first, and my "opened and abandoned a full scan on every equality DELETE" was wrong: nothing was ever opened. On top of that `processAndBlock()` is itself unreachable from the only path that builds this step, since the planner's `getKeyCondition()` returns a sub-expression and never an `AndBlock`.

So rather than build a contrived fixture for a branch that can only throw, I removed it (and `allEqualities()`, its only caller). The surviving `else` raises the error that names the condition, which is strictly more useful than "index is not ordered". The test that claimed to pin it is replaced with one that pins what is testable there — the equality path releases its cursor too, with the same rise-above-zero precondition as the others. The release note repeated the same wrong claim and is corrected.

**2. Side-effecting getter.** Fair, and worth acting on. Renamed to `countActiveCursors()`, with the javadoc saying why it isn't a getter: it does purge-on-read, the same thing `WeakHashMap` does, so it is O(live cursors) and can log. Only caller is the retired-file drop, where neither matters.

**3. Residue when no drop is ever attempted.** Agreed it's a non-issue, and I'm deliberately not fixing it. The only other place to purge from is `onCursorOpened`, which would put an O(live cursors) walk on the scan-open path to reclaim empty `WeakReference` shells bounded by the number of cursors actually leaked — a walk on a hot path to tidy up a case that is already a bug. Documented the bound on `purgeCollectedCursors()` instead.

Engine `index` + `query.sql.executor` suites re-run green.

lvca:
Round 2 — item 1 was a real bug in the code this PR added. Fixed, along with item 3.

**1. NPE-masking cleanup in `TypeIndex.range()`.** Correct, and the diagnosis is exactly right: `MultiIndexCursor` keeps the list it is handed rather than copying it, so a constructor that fails inside `initCursors()` has already closed the children and nulled their slots, and the cleanup loop then walked nulls. No leak, but the caller got an NPE over the real cause. Null-guarded.

I kept the catch rather than narrowing it to "only a `range()` call that throws before the constructor" — `close()` is idempotent on every implementation, so the overlap is free, and a guard that stays correct if `MultiIndexCursor` ever stops cleaning up after itself is worth more than the two lines it saves.

You were also right that the failed-construction path had no coverage. Added `aMultiIndexCursorThatFailsToConstructClosesTheChildrenAndClearsTheList`: a child whose `hasNext()` throws (standing in for a lazy page read failing) must leave every child closed, the caller's list holding nulls, and the *original* exception propagating. Mutation-checked — removing the `cursors.set(i, null)` from `closeQuietly()` makes it fail.

**3. Cursor not released on a `RuntimeException` after assignment.** Fixed rather than argued away; the PR's whole thesis is that release should not depend on someone else getting around to `close()`. `init()` now releases on any abort.

**2. `IndexCursorCollection.iterator()` re-startability.** Checked, and there is no behavioural change to confirm: the field is assigned once in the constructor (`this.iterator = collection.iterator()`), so `iterator()` already handed back the *same* single-use iterator on every call. It was never re-startable — the change is only that a for-each now routes through `next()` and keeps `last` in sync. The only construction sites are the four in `TypeIndex.get()`, each returning straight to a caller that drains once.

Engine `index`, `query.sql.executor` and `query.select` suites re-run green.

lvca:
Round 3 — LGTM with three non-blocking notes. Acted on two, verified the third.

**1. GC-dependent test.** Fair, and worth writing down. Added a comment saying it depends on the collector clearing a weak referent, that this is a property of the runtime rather than of this codebase, and that the cursor is opened in its own frame so no local keeps it on the stack.

I kept it asserted rather than best-effort deliberately, and said so in the comment: this is the test for the PR's central safety net, and one that passes silently when the net does not work would be worse than one that occasionally needs a re-run.

**2. Single-pass behaviour change.** Checked and there is no change to flag — the backing iterator is created once in `IndexCursorCollection`'s constructor, so the old `iterator()` handed back the same exhausted instance on a second call, not a fresh traversal. It was already single-pass; only `getRecord()` staying in sync is new. Documented on the method so the next reader does not have to re-derive it. (The vector cursor is new in the same sense — its `resultRIDs.iterator()` was a fresh traversal, but it was also a *second* one, which is the bug.)

**3. `getActiveCursors()` stragglers.** Verified across the whole tree, not just Java: zero remaining references in any file type outside `.git/` and `target/`.

That closes everything raised across the three rounds. From my side this is ready to merge — the only outstanding red check is `Meterian client scan`, which fails on the pre-existing npm dependencies in `studio` / `e2e-js` / `e2e-studio` (security score 0 against a minimum of 90). This PR touches Java sources and one markdown file, so it neither causes nor can fix that.

lvca:
Round 4 — LGTM, four non-blocking notes. Two were "so a future reader doesn't misread this", which is worth a line of comment each; the other two are already settled.

**2. `SelectExecutor` asymmetry.** Documented at the site: only `iteratorFromIndexes` is released because the alternatives `buildIterator()` can return — a type scan, a bucket scan, a `MultiIterator` over several buckets — hold no per-series file registration and are not closeable at all. Now nobody has to re-derive that it isn't an oversight.

**3. No-op `close()` default.** Also worth writing down, and it is the honest limit of this change: `Cursor.close()` now says that a new implementation holding a resource must override it, and that the weak-reference net reaches only the compacted-series cursors — nothing catches a future cursor type that acquires something and inherits the default.

**1. Public signature change** — confirmed no callers anywhere in the tree, in any file type.
**4. GC dependence** — already called out in the test's Javadoc, asserted rather than best-effort on purpose.

On verification: I ran the suites locally rather than relying on the description — engine (`index`, `query.sql`, `query.select`, `opencypher`, `function`, `database`, `graph`, `schema`, `engine`) and gremlin, all green, plus a per-fix mutation check (revert one fix, confirm the matching test fails).

CI: everything green except `Meterian client scan`, which fails on the pre-existing npm dependencies in `studio` / `e2e-js` / `e2e-studio` (security 0 against a minimum of 90). This PR is Java sources plus markdown, so it neither causes nor can fix that — worth confirming it fails the same way on `main` before treating it as a merge blocker.

Everything raised across all four rounds is now addressed; the last two commits are comments only.

Issue #5660, opened 2026-08-01
Cypher: a relationship pattern with a pinned target expands the source's whole edge list


lvca:
## Symptom

An ingest that guards every insert with a duplicate check slows down linearly as it runs. On a
transaction graph with 227,802 accounts and 2,563,705 edges the loader was at ~5,000 ms/query and
climbing, on a build that already carries #5499.

A thread dump taken during the slowdown has all six loader threads in the same place, and none of
them in the write path:

```
CreateStep <- FilterPropertiesStep <- FilterOperator(NOT) <- ExistsExpression.evaluate
  <- (nested plan) FinalProjection <- ProjectReturn <- FilterProperties <- MatchRelationshipStep
     -> processStandardPath  MultiIterator/EdgeIteratorFilter -> lookupByRID
     -> processStandardPath  getTargetVertex -> ImmutableEdge.checkForLazyLoading
     -> processStandardPath  matchesEdgeProperties -> InlineProperties.matches -> deserialize
```

Zero threads in `PageManagerFlushThread`, `WALFile` or commit-lock acquisition. The DEBUG log agrees:
commits take 1-2 ms, while the gaps *between* consecutive transactions are 1.6 s and 3.17 s, and the
guard logs nothing at DEBUG.

The graph is heavily skewed - the five busiest accounts hold 347,154 / 277,189 / 235,073 / 129,915 /
104,964 edges, about 43% of all edges - so a large share of inserts touch a hub.

## Cause

`NOT EXISTS { (a)-[:INITIATED {...}]->(t) }` runs as a correlated subquery that pins both endpoints.
`MatchRelationshipStep.processStandardPath` nevertheless expands the source's **whole** edge list, and
for every edge, before any filter runs:

1. `IteratorFilterBase.hasNext` calls `lookupByRID(edge, false)` - a database read lock and a record
   shell per edge (and under `REPEATABLE_READ` a forced content load);
2. `getTargetVertex` calls `edge.getInVertex()`, which forces `checkForLazyLoading()` - a **full record
   load**, unconditionally;
3. `matchesEdgeProperties` deserializes the edge's properties.

The bound-target identity check is the *last* thing that runs. So answering "is `a` linked to `t`?" on
a 347k-degree hub costs ~347k random record reads, and the cost grows with the hub's degree - which is
exactly why ms/query climbs as the load progresses.

The information needed to reject those edges is already in hand: the neighbour vertex RID is stored
[...]

Issue #5608, opened 2026-07-31
Follow-ups from #5596: surface the merge counters, pin the compressPage coupling, prune redundant poison calls


lvca:
### Summary

Four follow-ups left behind by #5596 / #5599 (merged as 49206db2d), which made the commit-time page merges prove their coverage instead of relying on every writer to call `poisonEdgeAppendPage`/`poisonSlotRebasePage`. None is a correctness defect - the coverage proof fails safe by construction (an undeclared byte degrades a merge to an ordinary retry, never to a lost write). They are the observability, enforcement and cleanup items that work surfaced.

The first one is the only one I would call important: it makes the metric that #5596 introduced as its operational signal unreachable by an operator.

---

- [ ] **1. None of the three merge counters reach `Profiler`, so an operator cannot see them.**

  `PageManager.PPageManagerStats` carries `edgeAppendMerges`, `txPageSlotMerges` and (since #5596) `mergesDeclinedByCoverage` (PageManager.java:96-99, populated at :596-598). `Profiler` reads exactly one page-manager contention stat, `concurrentModificationExceptions` (Profiler.java:124/146 for the JSON, :305/362 for the text dump), and none of the three merge counters. They are therefore in-process only - reachable via `PageManager.getStats()`, which is what the tests use.

  This predates #5596 for the two older counters, but #5596 is what makes it matter: `mergesDeclinedByCoverage` is documented as **the** signal that some writer is dirtying a mergeable page without declaring it, and the advice on it is literally "watch it next to `edgeAppendMerges`/`txPageSlotMerges`: a jump here with a dip there is exactly the shape of a forgotten declaration". Today none of the three is visible outside a debugger or a unit test, so that advice cannot be followed in production.

  Fix: add the three to the `PAGE-MANAGER` line of the text dump and to the JSON object, next to `concurrentModificationExceptions`. Three lines. Worth checking at the same time whether they should be Micrometer meters rather than only `Profiler` fields, since a merge-rate collapse under contention is exactly the kind of thing an alert should catch.

- [ ] **2. `compressPage`'s coverage declaration is implicitly coupled to the commit loop, with nothing enforcing it.**

[...]

PR #5678, opened 2026-08-01
fix(engine) #5670: an unreadable edge-list chunk retries the delete instead of skipping it

### LINKED ISSUE #5670: ConcurrentEdgeAppendMergeTest.concurrentAppendsAndRemovesStayConsistent reports one edge too many under load
### Summary

`com.arcadedb.graph.ConcurrentEdgeAppendMergeTest.concurrentAppendsAndRemovesStayConsistent` failed once during a large batch run with:

```
expected: 3000L
 but was: 3001L
  Suppressed: AssertionFailedError: expected: 0 but was: 1   (TestHelper.checkDatabaseIntegrity)
```

It then passed 3/3 in isolation and in a full 326-test run of `com.arcadedb.graph.**`, so it reproduces only under concurrent load from the rest of the suite.

### Why this is worth a look rather than a retry annotation

The count is **one too many**, not one too few, and `checkDatabaseIntegrity` flagged a problem alongside it. A lagging read would under-count; an over-count after concurrent appends and removes is the signature of an edge that was appended twice, or removed and left reachable - which is the exact class of bug the edge-append merge path (#5596, #5608, #5569) exists to prevent. It may well be a test-side race in how the expected total is computed, but that should be established rather than assumed, because the alternative is a real merge defect that only shows under contention.

Suggested first step: capture the integrity-check detail on failure (which bucket/RID is the extra one) rather than only the count, so the next occurrence says whether the surplus edge is a duplicate append or a survivor of a remove.

Reproduced once on the `issue-5636` branch, which touches no graph or edge code; the surrounding suites were green.

Found while working on #5636.

lvca:
Fixes #5670.

## The issue is real, and it reproduces

The reported `ConcurrentEdgeAppendMergeTest` failure - `expected: 3000L but was: 3001L` with one
`checkDatabaseIntegrity` error alongside - is not a test-side race. A tightened version of the same
workload (8 threads x 60 delete+append iterations on a hub, 40 rounds) fails in ~2 of 3 runs with the
identical signature:

```
[round 8 in-degree] expected: 480L but was: 481L
  Suppressed: AssertionFailedError: expected: 0 but was: 1  (checkDatabaseIntegrity)
```

## Root cause

Instrumenting the swallow points caught it exactly:

```
IN rid=#3:14336 vertex=#1:17 ex=Record #3:14336 not found
```

`GraphEngine.deleteEdge` disconnects the edge from both endpoints and then deletes the edge record.
The disconnection read the endpoint's chain best-effort in three places at once:

- `getEdgeHeadChunk` answers `null` when the head chunk cannot be loaded,
- the chain hops (`EdgeLinkedList.readChunk`) used a plain lookup,
- and `deleteEdge` wrapped the lot in `catch (SchemaException | RecordNotFoundException)`.

All three read "chunk unreadable" as "nothing to remove here" - and the edge record below was
deleted anyway.

Under concurrency a chunk is regularly unreadable for reasons that say nothing about the graph. A
commit publishes its pages one at a time and a reader takes no commit lock, so a vertex page can
expose a new edge-list head RID a moment before that head's own page becomes visible; and a chunk
emptied by another transaction is relinked out of the chain while a walker is still following a
pointer to it. Hitting either window ended the removal having removed nothing, so the back-reference
outlived its edge: **one edge too many in the endpoint's degree, and one broken link**. On a hot
vertex the window is narrow, which is why it surfaced as a rare, unexplained off-by-one.

## The fix

The issue suggested capturing the integrity-check detail so the *next* occurrence would say which
RID is surplus. That is a better failure report, not a fix - and the surplus turned out to be
explainable without it. The alternative implemented here is to close the window itself, using the
line the codebase already draws elsewhere: **the append path (`getOrCreateEdgeList`) answers exactly
this transient with a retryable `ConcurrentModificationException`**, as does
`StripedEdgeList.addChain` for the per-stripe chains. The removal path never got the same treatment.

- **`GraphEngine.getEdgeHeadChunkForWrite`** - strict counterpart of `getEdgeHeadChunk`. `null` now
  means one thing only: the vertex has no edge list in that direction, so there is genuinely nothing
  to remove. A head RID that is present but unreadable is a retryable conflict.
- **`GraphEngine.deleteEdge`** - endpoint resolution split from chain mutation
  (`resolveEndpointToDisconnect`), so only a vanished endpoint *vertex* is tolerated: there is
  nothing to disconnect from a vertex that is gone.
- **`EdgeLinkedList.readChunk` / `loadChunkForWrite`** - an unreadable chunk on a removal walk is a
  retryable conflict. This subsumes `StripedEdgeList.loadStripeHead`, which is **removed** rather
  than left behind as dead code.
- **`EdgeIteratorFilter`** - the opportunistic pruning of an already-dangling reference runs inside a
  READ, so it stays best-effort: it absorbs the new retryable conflict and leaves the ghost for a
  later pass instead of failing the iteration.

Read paths are unchanged. Iteration and counting still skip a momentarily unreadable chunk rather
than failing, which is the documented best-effort contract for reads.

### Visible behaviour change

An `edge.delete()` racing a concurrent write on the same endpoint can now raise
`ConcurrentModificationException` where it previously "succeeded". That is a `NeedRetryException`, so
`database.transaction(...)` and the server's auto-retry for single-request commands absorb it. A
client-managed explicit transaction spanning several requests sees it and should retry - the same
contract concurrent updates have always had. Noted in the release notes.

## Tests

`Issue5670EdgeDeleteDanglingBackRefTest` - **all three fail on the current code, all three pass with
the fix**:

| Test | Shape |
|---|---|
| `edgeDeleteRaisesRetryableConflictWhenTheEndpointHeadChunkIsUnreadable` | deterministic; the field-observed window |
| `edgeDeleteRaisesRetryableConflictWhenAMidChainChunkIsUnreadable` | deterministic; the relinked-out-from-under-the-walker window |
| `concurrentDeleteAndAppendNeverLeaveADanglingBackReference` | `@Tag("slow")`; the reported shape |

The deterministic pair asserts its preconditions (the chain really spans several chunks, the victim
really sits behind the hole, the degree is right before the corruption) and matches the offending
chunk RID in the exception message, so it cannot pass on an unrelated throw.

## Verification

Full `engine` suite, three runs. `ExplicitLockingTransactionTest.errorOnExplicitLock` failed on the
first run, so it was chased down rather than retried away: it **passed** on a run with this branch's
test class but the main-code fix reverted, and **passed** again on a second run with the fix. It is a
pre-existing load-sensitive assertion - `explicitLock` throws only when `immutablePages` is non-empty,
and under `REPEATABLE_READ` a page is retained only when `pageNumber < file.getTotalPages()`, which is
the **on-disk** file size; a just-committed page whose flush is still queued reads as "new" and is not
retained. Unrelated to edge handling; filed as #5679.

- baseline at HEAD: 10625 tests, 0 failures
- with fix: 10587 tests, 0 failures

## Left out, deliberately

`deleteVertex` keeps its documented best-effort edge disconnection (`Issue4420TolerantDeleteTest`,
`Issue4432CorruptVertexDeleteTest` pin that a structurally broken vertex stays deletable). It is
exposed to the same transient window when *collecting* its edges, but making it strict would trade
repairability for strictness - a decision worth taking on its own merits, not as a side effect here.
Filed as #5680. Note that it does get the fix transitively where it matters: each edge it
deletes goes through `deleteEdge`, so a neighbour's back-reference is no longer left dangling.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1245,6 +1245,9 @@ reference during a read still skip a momentarily unreadable chunk rather than fa
 The other side of that trade, taken deliberately: an endpoint edge list that is not transiently invisible but
 genuinely broken is indistinguishable from one at that moment, so deleting an edge attached to it now fails instead
 of completing and leaving the back-reference behind. `CHECK DATABASE` remains the repair path - it rebuilds a chain
-that cannot be loaded and drops the references into it.
+that cannot be loaded and drops the references into it. This reaches vertex deletion as well, because disconnecting
+an edge touches the vertex at the other end: deleting a healthy vertex whose *neighbour's* edge list cannot be read
+now reports a conflict rather than succeeding, since succeeding would delete the edge record while the neighbour
+keeps pointing at it - dangling a reference on a vertex nobody asked to touch.
 
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1298,4 +1298,15 @@ an edge touches the vertex at the other end: deleting a healthy vertex whose *ne
 now reports a conflict rather than succeeding, since succeeding would delete the edge record while the neighbour
 keeps pointing at it - dangling a reference on a vertex nobody asked to touch.
 
+**If you hit that, the recovery is `CHECK DATABASE ... FIX` and then retry the delete.** The symptom is a delete that
+keeps failing with `ConcurrentModificationException` on the same vertex however often it is retried, which is what
+tells a broken list apart from ordinary contention (that one succeeds on a retry). The repair is never blocked by the
+delete being blocked: `CHECK DATABASE` reads edge lists through the best-effort reader, so it can still walk, rebuild
+and re-link a chain that the delete path now refuses.
+
+Note also that on a hot super-node under heavy concurrent delete-and-append the transient window is now answered with
+a retry rather than passing silently, so those transactions retry slightly more often than before. That is the
+intended shift - from "quietly wrong" to "occasionally repeated" - and it lands on exactly the super-node shape the
+bug affected. `arcadedb.txRetryDelay` and `arcadedb.txRetries` are the tuning levers if a workload needs them.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1235,11 +1235,16 @@ direction, so there is genuinely nothing to remove. Tolerance for an endpoint **
 unchanged: there is nothing to disconnect from a vertex that is gone.
 
 Visible effect: an `edge.delete()` racing a concurrent write on the same endpoint can now raise a
-`ConcurrentModificationException` where it previously "succeeded". That is a `NeedRetryException`, so the standard
-retry loop (`database.transaction(...)`, and the server's auto-retry for single-request commands) absorbs it. A
-client-managed explicit transaction spanning several requests sees it and should retry the transaction, which is the
-same contract concurrent updates have always had. Best-effort callers are unaffected: iteration, counting and the
-opportunistic pruning of an already-dangling reference during a read still skip a momentarily unreadable chunk
-rather than failing.
+`ConcurrentModificationException` where it previously "succeeded" - and so can moving an edge, which disconnects it
+the same way. That is a `NeedRetryException`, so the standard retry loop (`database.transaction(...)`, and the
+server's auto-retry for single-request commands) absorbs it. A client-managed explicit transaction spanning several
+requests sees it and should retry the transaction, which is the same contract concurrent updates have always had.
+Best-effort callers are unaffected: iteration, counting and the opportunistic pruning of an already-dangling
+reference during a read still skip a momentarily unreadable chunk rather than failing.
+
+The other side of that trade, taken deliberately: an endpoint edge list that is not transiently invisible but
+genuinely broken is indistinguishable from one at that moment, so deleting an edge attached to it now fails instead
+of completing and leaving the back-reference behind. `CHECK DATABASE` remains the repair path - it rebuilds a chain
+that cannot be loaded and drops the references into it.
 
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1282,21 +1282,28 @@ half-done one. `null` from the write-side head lookup now means one thing only -
 direction, so there is genuinely nothing to remove. Tolerance for an endpoint **vertex** that no longer exists is
 unchanged: there is nothing to disconnect from a vertex that is gone.
 
-Visible effect: an `edge.delete()` racing a concurrent write on the same endpoint can now raise a
-`ConcurrentModificationException` where it previously "succeeded" - and so can moving an edge, which disconnects it
-the same way. That is a `NeedRetryException`, so the standard retry loop (`database.transaction(...)`, and the
+**Visible effect, and it is not limited to deleting an edge.** Three operations disconnect an edge and therefore
+share the new contract:
+
+- `edge.delete()` / `DELETE EDGE`,
+- moving an edge, which disconnects it the same way,
+- **`vertex.delete()` / `DELETE VERTEX`** - the widest reach of the three. Deleting a vertex disconnects each of its
+  edges from the vertex at the *other* end, so the strict read lands on a **neighbour's** edge list. A vertex that is
+  itself perfectly healthy can now fail to delete because a neighbour's list could not be read at that moment.
+
+Any of them racing a concurrent write can now raise a `ConcurrentModificationException` where it previously
+"succeeded". That is a `NeedRetryException`, so the standard retry loop (`database.transaction(...)`, and the
 server's auto-retry for single-request commands) absorbs it. A client-managed explicit transaction spanning several
 requests sees it and should retry the transaction, which is the same contract concurrent updates have always had.
 Best-effort callers are unaffected: iteration, counting and the opportunistic pruning of an already-dangling
 reference during a read still skip a momentarily unreadable chunk rather than failing.
 
-The other side of that trade, taken deliberately: an endpoint edge list that is not transiently invisible but
-genuinely broken is indistinguishable from one at that moment, so deleting an edge attached to it now fails instead
-of completing and leaving the back-reference behind. `CHECK DATABASE` remains the repair path - it rebuilds a chain
-that cannot be loaded and drops the references into it. This reaches vertex deletion as well, because disconnecting
-an edge touches the vertex at the other end: deleting a healthy vertex whose *neighbour's* edge list cannot be read
-now reports a conflict rather than succeeding, since succeeding would delete the edge record while the neighbour
-keeps pointing at it - dangling a reference on a vertex nobody asked to touch.
+The other side of that trade, taken deliberately: an edge list that is not transiently invisible but genuinely
+broken is indistinguishable from one at that moment, so the operations above now fail instead of completing and
+leaving the back-reference behind. For `DELETE VERTEX` that means a healthy vertex next to a corrupted one is not
+deletable by the normal path until the corruption is repaired - succeeding would delete the edge record while the
+neighbour keeps pointing at it, dangling a reference on a vertex nobody asked to touch. `CHECK DATABASE` remains the
+repair path: it rebuilds a chain that cannot be loaded and drops the references into it.
 
 **If you hit that, the recovery is `CHECK DATABASE ... FIX` and then retry the delete.** The symptom is a delete that
 keeps failing with `ConcurrentModificationException` on the same vertex however often it is retried, which is what
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1210,4 +1210,36 @@ and expressing that would have been faithful to its intent, but it would have ti
 of repairing a config that enforced nothing - a decision worth taking on its own merits. Development dependencies are
 still covered by the `npm audit --audit-level=moderate` step of the same workflow.
 
+## Deleting an edge no longer leaves its back-reference behind under concurrency
+
+Removing an edge disconnects it from both endpoints and then deletes the edge record. The disconnection walked the
+endpoint's edge-list chain with best-effort reads: the head chunk came from a helper that answers `null` when it
+cannot be loaded, the chain hops used a plain lookup, and `deleteEdge` wrapped the lot in a
+`catch (RecordNotFoundException)`. All three read "chunk unreadable" as "nothing to remove here" - and then the edge
+record was deleted anyway.
+
+Under concurrency a chunk is regularly unreadable for reasons that say nothing about the graph. A commit publishes
+its pages one at a time and a reader takes no commit lock, so a vertex page can expose a new edge-list head RID a
+moment before that head's own page becomes visible; and a chunk emptied by another transaction is relinked out of
+the chain while a walker is still following a pointer to it. Hitting either window ended the removal without having
+removed anything, so the back-reference outlived its edge: the endpoint reported one edge too many
+(`countEdges`, `both()`, `in()`/`out()`) and `check database` reported one broken link. On a hot vertex - the
+super-node shape this happens on - the window is narrow, which is why it surfaced as a rare, unexplained off-by-one
+rather than as a reproducible failure.
+
+The append path already answered exactly this window with a retryable `ConcurrentModificationException`. The removal
+path now does the same: a head chunk that is present but unreadable, and a chain hop that cannot be read, are
+retryable conflicts, so the transaction re-reads a consistent view and completes the removal instead of committing a
+half-done one. `null` from the write-side head lookup now means one thing only - the vertex has no edge list in that
+direction, so there is genuinely nothing to remove. Tolerance for an endpoint **vertex** that no longer exists is
+unchanged: there is nothing to disconnect from a vertex that is gone.
+
+Visible effect: an `edge.delete()` racing a concurrent write on the same endpoint can now raise a
+`ConcurrentModificationException` where it previously "succeeded". That is a `NeedRetryException`, so the standard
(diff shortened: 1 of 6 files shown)


lvca:
Thanks - all three points were valid. Addressed in 51c212bfe.

**1. Head RID read outside the try — fixed, and it was reachable.** Confirmed the mechanism: `ImmutableVertex.getOutEdgesHeadChunk()` → `checkForLazyLoading()` → `ImmutableDocument.checkForLazyLoading()` → `LocalBucket.getRecord(rid)`, which raises `RecordNotFoundException`. The read moved inside the try.

Rather than just move it, I pinned it: `headChunkForWriteRaisesRetryableConflictWhenTheVertexItselfVanishes` resolves a handle with `lookupByRID(rid, false)` (so the record is not materialised), deletes the vertex record, then calls the strict lookup. On the previous commit it fails exactly as you predicted:

```
Expecting actual throwable to be an instance of:
  com.arcadedb.exception.ConcurrentModificationException
but was:
  com.arcadedb.exception.RecordNotFoundException: Record #1:0 not found
	at com.arcadedb.engine.LocalBucket.getRecord(LocalBucket.java:291)
	at com.arcadedb.database.ImmutableDocument.checkForLazyLoading(ImmutableDocument.java:216)
	at com.arcadedb.graph.ImmutableVertex.checkForLazyLoading(ImmutableVertex.java:251)
```

The message now carries the underlying cause, so the missing RID is still named.

**2. Permanent corruption → permanent failure.** Confirmed and deliberate. `CHECK DATABASE` is the repair path for a chain that cannot be loaded (`GraphDatabaseCheckerChainRebuildTest` covers exactly the head-chunk-unloadable case), so a genuinely broken graph is still cleanable — but you are right that it now disagrees with `deleteVertex`. The strict lookup's javadoc states the trade, and I recorded the coupling on #5680 so whoever picks that up settles both paths together instead of leaving edge deletion strict and vertex deletion tolerant by accident.

**3. `moveEdge`.** Correct — it disconnects through `deleteEdge` and shares the contract. The release note's visible-behaviour section now names it.

**Test nit** — taken: the worker `catch` now says that any exception surfacing there IS the bug, not a tolerated retry.

Re-ran the graph package after the change: 320 tests, 0 failures, including `Issue4420TolerantDeleteTest` and `Issue4432CorruptVertexDeleteTest`.

lvca:
Addressed in ae9a67b4e. Thanks — the second-order effect was real and worth stating; one supporting claim did not hold up, so I measured both.

**The neighbour effect — confirmed, and kept strict deliberately.** You are right that `deleteVertex` disconnects the vertex at the *other* end too, and that its loop catches only `RecordNotFoundException`. Measured on the same fixture (a healthy vertex whose only edge points at a hub whose IN head chunk was removed):

```
before: delete SUCCEEDED   (edge record gone, neighbour still pointing at it)
after:  ConcurrentModificationException
```

Keeping it strict. Succeeding there deletes the edge record while the neighbour keeps the back-reference — precisely the corruption this PR fixes, inflicted on a vertex the caller never named. Now pinned by `deletingAVertexWhoseNeighbourListIsUnreadableReportsAConflictRatherThanDanglingTheReference`, and stated in the strict lookup's javadoc and the release note. Recorded on #5680, including the observation that `force` does not currently help here — the per-edge loop sits outside its promise, which is the natural shape for a fix if that issue resolves toward tolerance.

**`Issue4432CorruptVertexDeleteTest` — the suspicion does not hold.** It is not green by scan order. `LocalBucket` detects the invalid record size on the first read and deletes the record:

```
SEVER [LocalBucket] Invalid record size 33554444 for record #1:0: deleting record
```

so the corrupt vertex fails at *resolution* — which stays tolerant — and never reaches `getEdgeHeadChunkForWrite`. Deleting the neighbour first with the corrupt vertex still present raises the identical `RecordNotFoundException` with and without this branch. The reversed ordering you describe is therefore already covered; it just resolves through a different door than expected.

**`EdgeIteratorFilter` partially-anchored pages.** Checked, and the behaviour is a bit different from the framing: with `joinCurrent = true`, `LocalDatabase.transaction`'s `catch (NeedRetryException …)` calls `rollback()` on the *joined* transaction before retrying, so the anchored pages do not survive into a later commit — they are discarded with the tx. The wart is that the enclosing read transaction is rolled back under the caller, but that is pre-existing for any `NeedRetryException` on this path (promoted super-nodes have been able to raise CME here via `StripedEdgeList.addChain`), and the new `catch` strictly improves it by not also propagating the exhausted exception into a read. Not introduced here, so left alone.

**Log/metric on retry exhaustion.** Agreed in principle, but the right home is the generic retry-exhaustion point in `LocalDatabase.transaction`, not this path — a warning at the conversion site would fire on ordinary healthy contention, which is exactly the noise it is meant to distinguish from. Better folded into the #5680 decision than bolted on here.

Graph package after the change: 321 tests, 0 failures, including `Issue4420TolerantDeleteTest` and `Issue4432CorruptVertexDeleteTest`.

lvca:
Both asks addressed in ba56346a0.

**1. `CHECK DATABASE` recovery is now explicit in the release note.** Verified your reading before writing it down: `GraphDatabaseChecker` uses `getEdgeHeadChunk` exclusively (4 call sites, all best-effort), and `getEdgeHeadChunkForWrite` has exactly one caller, `deleteEdge` — so the repair path is not gated behind the strict walk and cannot be blocked by the delete being blocked. The note now gives the user the whole loop: the symptom (a delete that keeps failing however often it retries, whereas ordinary contention succeeds on a retry), the recovery (`CHECK DATABASE ... FIX`, then retry), and why the repair still works.

**2. Retry pressure on hot super-nodes** — recorded in the same section as the accepted shift it is, "from quietly wrong to occasionally repeated", landing on exactly the super-node shape the bug affected, with `arcadedb.txRetryDelay` / `arcadedb.txRetries` named as the levers.

**3. `#5680` is the right home** — confirmed, and it already carries the full picture: the original `deleteVertex`-own-list tolerance, the neighbour case from review 2, and the observation that `force` does not currently help (the per-edge loop sits outside its promise), which is the natural shape for a fix if that issue resolves toward tolerance.

**On the minor point** — the `existsRecord` check turns out **not** to be redundant, and I have commented it rather than removed it. `Edge.getOutVertex()`/`getInVertex()` load with `loadContent=false`, which returns a *lazy* handle without touching the bucket, so a deleted endpoint does not surface at that resolution at all. It surfaces later, inside `getEdgeHeadChunkForWrite`, which maps it to a retryable conflict. So `existsRecord` is precisely the discriminator between "vertex gone → nothing to disconnect, tolerate" and "vertex present but its list unreadable → retry"; dropping it would silently reclassify the first as the second. `headChunkForWriteRaisesRetryableConflictWhenTheVertexItselfVanishes` exercises exactly that lazy path. Good catch to raise it — it is the kind of thing that reads as dead weight and is load-bearing.

Graph package after the change: 334 tests, 0 failures.

lvca:
Thanks — items 3 and 4 are fixed in e484f2361; 1 and 2 I looked into and am leaving as-is, reasoning below.

**3. Concurrent test robustness — fixed.** Workers now catch `Throwable` (an `Error` could previously skip the `countDown`) and the latch wait is bounded at 5 minutes against a round that takes seconds. A wedged worker now fails the test instead of hanging CI.

**4. Stale comment — fixed, after checking which half was wrong.** Rather than assume the comment, I deleted an edge carrying an indexed property and watched the index: 1 entry → 0, lookup empty. So the *code* is right and the comment was inverted. `LocalDatabase.deleteRecordNoLock` cleans the index entries and fires the delete events **before** dispatching an `Edge` to `deleteEdge`, so this call is deliberately the physical removal alone — routing back through the database would repeat that work, not add it. The comment now says that, with the reason.

**1. `SchemaException` on the chain-mutation path — intentional, and I would rather not map it.** The realistic source is the bucket lookup in `loadChunkForWrite` / `buildEdgeList`, i.e. the edge bucket no longer resolving — which means the schema genuinely changed (type or bucket dropped), not that a page is momentarily unpublished. A retry cannot fix that, so mapping it to CME would burn `TX_RETRIES` and then report "concurrent commit in flight" for what is actually a schema error. Failing directly with the real exception is the more useful outcome, and it is strictly better than the old behaviour, which swallowed it and deleted the edge record anyway — the same silent-corruption shape this PR is closing. Worth noting the precedent points the same way: #5608 changed `TransactionContext.edgeSegmentPageKey` to a null-tolerant bucket lookup specifically so a *transient* miss became retryable, while leaving genuine schema errors as errors.

**2. Read-side strictness via `chainsForNeighbour(strict=true)`** — agreed, and confirmed unchanged by this PR (it predates it and I have not touched `addChain`). Flagging it in the thread is the right place for it; if it deserves to be reconsidered, that is its own change with its own reasoning, not a rider here.

Graph package after the change: 334 tests, 0 failures. Full `engine` suite was green on this branch before the merge from main (10587 tests, 0 failures) and CI is now running it on the merged tree.

lvca:
Both actionable items in d73cad559.

**1. `DELETE VERTEX` prominence — agreed, and you are right that it is the widest-reaching of the three.** The note led with `edge.delete()` and mentioned the vertex reach only at the tail of the trade-off paragraph. The visible-effect section now opens with "and it is not limited to deleting an edge", lists the three operations that share the contract (`edge.delete()`/`DELETE EDGE`, `moveEdge`, `vertex.delete()`/`DELETE VERTEX`), and calls out that the vertex case lands the strict read on a **neighbour's** list — so a vertex that is itself perfectly healthy can fail to delete. The trade-off paragraph now states the operational consequence in the same terms: a healthy vertex next to a corrupted one is not deletable by the normal path until the corruption is repaired.

**4. Null endpoint — guarded.** You are right that `existsRecord(null)` raises `IllegalArgumentException`, which the `catch (SchemaException | RecordNotFoundException)` does not cover and which is not retryable — a hard failure escaping the one method whose job is to decide what is tolerable. `resolveEndpointToDisconnect` now returns `null` for a null endpoint RID, matching the "nothing to disconnect" intent. Theoretical since an edge always carries both endpoints, but it is a guard worth having in exactly that method.

**2. Permanent-failure trade-off** — acknowledged, and now stated in the release note in operational terms alongside the `CHECK DATABASE ... FIX` recovery and the symptom that distinguishes it from ordinary contention (a delete that keeps failing however often it retries).

**3. `existsRecord` cost on super-node teardown** — noted and accepted. It is load-bearing (it is what separates "vertex gone, tolerate" from "list unreadable, retry"), and a page-slot existence check per endpoint is small next to the chain walk that follows it. If it ever shows up in a teardown profile, the fix is to fold the existence answer into the resolution rather than to drop the distinction.

Graph package after the change: 334 tests, 0 failures — including main's new `EdgesConnectedToTest` from the #5660 merge.

lvca:
All three checked; one produced a change (b0271ab5b), two I am deliberately not making — reasoning below.

**1. `EdgeIteratorFilter` catching `NeedRetryException` rather than CME.** I checked the narrowing before applying it, and it would make behaviour *worse*. `NeedRetryException` has exactly two subclasses: `ConcurrentModificationException` and `LockTimeoutException`. The second says the same thing about this prune — come back for it later — and letting it escape would fail a **read** because an *optional* repair could not get a lock, which is exactly the outcome the catch exists to prevent. So the comment was the half that was wrong: it now says the condition being absorbed is "retry later", not one specific cause. Code unchanged, comment now matches it.

**2. Unguarded `EdgeSegment` cast in `buildEdgeList`.** Agreed it is pre-existing and low-risk (edge chunks live in their own `*_in_edges`/`*_out_edges` and stripe-pool buckets, so a reused slot is still an edge-list record). I would rather not map `ClassCastException` to CME, though: unlike an unreadable chunk, a vertex head pointing at something that is not a chunk is not a timing fact — it is corruption a retry cannot fix. Converting it would burn `TX_RETRIES` and then report "concurrent commit in flight" for a structural problem, trading a precise stack for a misleading message. The distinction this PR is built on is exactly "transient ⇒ retry, structural ⇒ report", and a CCE falls on the structural side.

**3. Server-side surfacing of the new CME.** Verified rather than assumed: `AbstractServerHttpHandler` already catches `NeedRetryException` and answers **HTTP 503 "Cannot execute command"**, logging at `FINE` with the exception message — a retryable status and a clean message, not a raw 500 stack. So `DELETE VERTEX` hitting the new conflict over HTTP surfaces sensibly today, and a client that retries on 503 behaves correctly. Nothing to change.

Graph package: 334 tests, 0 failures.

lvca:
Thanks. Checked the one actionable minor; it turns out not to apply, and I am leaving the test as it is.

**1. `TX_RETRY_DELAY` per-database instead of global.** Good instinct — the setting *is* declared `SCOPE.DATABASE`, and `engine/CLAUDE.md` even recommends exactly that for the lock timeouts. But the retry loop does not read it that way:

```java
// LocalDatabase.transaction(...), line 1387
final int retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
```

That is the global static, not the database's `ContextConfiguration`, so setting it on the instance would have no effect on the retry delay and the test would silently lose the tightening it is asking for. Set-global-and-restore-in-`finally` is the only form that works here, and it matches the sibling `ConcurrentEdgeAppendMergeTest`. (The mismatch between the declared `SCOPE.DATABASE` and the global read is a real if minor inconsistency in the engine — worth its own look, not a rider on this PR.)

**2. Comment density** — noted, and I agree with your conclusion; leaving it.

That closes every point raised across the review rounds. Summary of what changed as a result of them:

| Round | Point | Outcome |
|---|---|---|
| 1 | Head-RID read outside the strict `try` | **Real bug** — fixed, pinned by a test that fails on the prior commit |
| 2 | `deleteVertex` reaches a neighbour's list | Confirmed by measurement, kept strict, pinned + documented |
| 2 | `Issue4432` order-fragility | Investigated — did not hold up (the corrupt record self-deletes on read) |
| 3 | `CHECK DATABASE` recovery unstated | Added, after verifying the checker uses the best-effort reader |
| 4 | Unbounded latch / `Exception` not `Throwable` | Fixed |
| 4 | Comment contradicting its code | Verified which half was wrong (index 1 → 0), corrected the comment |
| 5 | `DELETE VERTEX` prominence | Rewritten to lead the visible-effect section |
| 5 | `existsRecord(null)` | Guarded |
| 6 | Narrow the catch to CME | Declined — would let `LockTimeoutException` fail a read |
| 7 | Per-database `TX_RETRY_DELAY` | Not applicable — the retry loop reads the global |

Follow-ups filed: #5679 (load-sensitive `ExplicitLockingTransactionTest`) and #5680 (`deleteVertex` tolerance, incl. the neighbour case and that `force` does not currently help).

Issue #4587, opened 2026-06-12
`DeltaOverlay.merge` decrements `deltaEdgeCount` for duplicate edge deletions


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.graph.olap.DeltaOverlay`
**Severity:** MEDIUM

### Summary

`newDeletedEdges` is a `Set`; duplicate `add(...)` calls are silently
absorbed, but `newDeltaEdgeCount--` runs unconditionally. The same
deletion replayed across merges (or emitted twice within one `TxDelta`)
drives the counter negative.

### Code reference

`engine/com/arcadedb/graph/olap/DeltaOverlay.java:186–194`

### Impact

`Math.abs(merged.getDeltaEdgeCount()) > compactionThreshold` is the
compaction trigger. Drift causes phantom compactions or suppresses
needed ones, wasting permits / virtual threads. Reported edge counts
are also wrong.

### Suggested fix

Capture the boolean return of `Set.add(...)` and only decrement when
it returns `true`.

PR #5684, opened 2026-08-01
fix(engine) #5663: a hop onto a bound vertex counts every relationship joining the pair

### LINKED ISSUE #5663: Cypher: a cycle pattern reports one row per closing edge instead of one per (first, closing) pair
Found while adding tests for #5660. It reproduces with that change fully reverted, so it is
independent of it.

## Repro

```cypher
CREATE (a:Account {code: 'HUB'})
CREATE (t:Txn {ref: 'SHARED'})
CREATE (a)-[:INITIATED {kind: 'payment'}]->(t)
CREATE (a)-[:INITIATED {kind: 'refund'}]->(t)
CREATE (t)-[:INITIATED {ref: 'REVERSED'}]->(a)
```

So the pair is joined by **two** parallel `INITIATED` edges going out from `a`, and **one** coming
back. Then:

```cypher
MATCH (a:Account {code: 'HUB'})-[r1:INITIATED]->(t:Txn {ref: 'SHARED'})-[r2:INITIATED]->(a)
RETURN r1.kind AS v
```

Expected `["payment", "refund"]` - the cycle can be walked two ways, one per first-hop edge, and
each closes through the single returning edge. Actual: `["payment"]`.

`RETURN r2.ref` likewise gives one `REVERSED` rather than two.

Relationship uniqueness does not explain the missing row: `r1` and `r2` bind to different edges in
both walks, so both rows are legal under Cypher's rule that a pattern may not reuse the same
relationship.

## Notes

- Which answer comes back depends on the plan chosen. With the relationships left anonymous the
  result flipped depending on which other tests had already run in the same JVM, because plan
  selection is influenced by process-wide state (the `GraphTraversalProvider` registry). Naming both
  relationships makes it deterministic at one row.
- The closing hop is an "expand into" - both endpoints are already bound - so this is likely the
  cardinality of that step: it appears to answer "is the cycle closed?" once for the pair rather than
  once per incoming row.
- The same query shape against the step-by-step and the operator plan disagreeing is the kind of
  divergence #5446 describes from the other direction.

## Impact

Under-counting on cycle patterns over multigraphs. Anything aggregating over a cycle
(`count(*)`, `collect(r1)`, `sum`) silently under-reports when parallel edges exist between the
same pair, which is normal in transaction or payment graphs.

## Where it is currently pinned

`CypherBoundTargetExpansionTest.narrowingHandlesACyclePatternWhereTheLastHopReturnsToABoundVariable`
deliberately does **not** assert the row count and says why, so that test does not encode this
behaviour as correct. It should assert the count once this is fixed.

lvca:
Closes #5663.

## The issue is real, and Neo4j answers the other way

A pattern relationship matches **once per relationship**, whether or not the pattern names it. So over

```cypher
CREATE (a:Account {code: 'HUB'})
CREATE (t:Txn {ref: 'SHARED'})
CREATE (a)-[:INITIATED {kind: 'payment'}]->(t)
CREATE (a)-[:INITIATED {kind: 'refund'}]->(t)
CREATE (t)-[:INITIATED {ref: 'REVERSED'}]->(a)

MATCH (a:Account {code: 'HUB'})-[r1:INITIATED]->(t:Txn {ref: 'SHARED'})-[r2:INITIATED]->(a)
RETURN r1.kind
```

the cycle can be walked two ways, one per first-hop edge, each closing through the single returning edge. Relationship uniqueness is satisfied in both: `r1` and `r2` bind different edges. Expected `["payment", "refund"]`, actual `["payment"]`.

## Root cause

`ExpandInto` - the operator the optimizer picks for a hop whose far end is already bound - was written as a semi-join: `isConnectedTo()` plus, when a variable had to be bound, one `findEdge()` for the *first* connecting edge. One row per input row, the rest of the pair discarded. `GAVExpandInto`, its CSR-backed twin, had the identical defect (`isConnectedTo` collapsed to a boolean).

The step-by-step executor (`MatchRelationshipStep`) had always got this right - it iterates the narrowed edge list - which is why the same query could answer differently depending on which plan won. That plan choice was itself unstable: `JoinOrderRule` kept its candidate hops in a `HashSet` of objects with identity hashing, so two equally cheap hops were tie-broken by whatever addresses the JVM handed out.

## The fix

**`ExpandInto` is an expansion, not a probe.** It walks the relationships joining the pair with `GraphEngine.getEdgesConnectedTo` and emits one row per relationship. It keeps exactly what made it fast: the source's edge list is filtered on the neighbour pointer held beside each edge pointer in the segment, so an edge that does not reach the target costs two primitive comparisons rather than a record load, and only the edges that do reach it are materialised. For the common single-edge pair this is now *one* narrowed scan where the old code did two (`isConnectedTo`, then `findEdge`).

It also gained what `ExpandAll` already had and it lacked:
- relationship uniqueness against the preceding hops of its own MATCH clause is applied per edge;
- an **anonymous** hop can now carry a synthetic edge-tracking variable, so a later same-clause hop cannot reuse the relationship it bound. Previously an anonymous bound-target hop was invisible to the uniqueness check;
- an undirected hop reaches a self-loop from both adjacency lists and reports it once (`SelfLoops.deduplicatingEdges`, the helper the other executors already share).

**`GAVExpandInto` reads the multiplicity off the CSR** instead of stopping at the first hit. New SPI method `GraphTraversalProvider.countEdgesBetween` - default implementation counts occurrences in `getNeighborIds`, `GraphAnalyticalView` overrides it with an equal-range scan around the binary-search hit, O(log(degree) + multiplicity), mirroring `isConnectedForType` entry by entry (overlay deletions and additions included) so the two never disagree about whether an edge is there.

**The planner stops sending hops to the CSR that it cannot answer.** Adjacency ids carry no edge identity, so `GAVExpandInto` cannot tell one relationship from another. It is now selected only when the hop needs no edge object at all: no variable captured, no edge tracking, and no preceding same-clause relationship variable to check against. That closes a pre-existing hole where an anonymous bound-target hop planned onto a view skipped relationship uniqueness entirely.

**Plan choice is deterministic.** `JoinOrderRule` uses a `LinkedHashSet`, so ties fall to the order the hops are written in.

### Considered and rejected

Special-casing the cycle shape, or keeping the semi-join and multiplying the row by a separate count, would leave two code paths disagreeing about what a hop means - which is the class of bug this is. Making the operator an expansion is the smaller and more honest change, and it is what Neo4j's `Expand(Into)` does.

There is one deliberate trade: the old code could stop at the first matching edge, the new one must walk the source's whole (narrowed) edge list to know the multiplicity. That is inherent to answering correctly, it is pointer comparisons rather than record loads, and record loads are unchanged (one per emitted row).

## Tests

- `CypherCycleCardinalityTest` (new): the issue's repro named and anonymous; `count(*)` over the cycle; a cycle that cannot close; the undirected closing hop, where relationship uniqueness has to survive the new multiplicity; parallel edges both ways; self-loops counted once under an undirected hop.
- `CypherGAVBoundTargetCardinalityTest` (new): the same cycle planned onto a `GraphAnalyticalView`, asserting from `EXPLAIN` that `GAVExpandInto` is the operator under test, and that it agrees with the plan without the view.
- `GraphAnalyticalViewTest.countEdgesBetweenReportsEveryParallelEdge` (new): the CSR count directly - parallel edges, direction, edge-type filter, untyped, and a pair nothing joins; the existing self-loop test now also pins the count.
- `CypherBoundTargetExpansionTest.narrowingHandlesACyclePatternWhereTheLastHopReturnsToABoundVariable` asserted no row count and said why, pointing at this issue. It asserts it now.
- Full `com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**` (8176 tests, includes the OpenCypher TCK suite) and `com.arcadedb.query.sql.**` (2659) green.

The `⭐ SEMI-JOIN` marker in `EXPLAIN` output reads `⭐ BOUND-TARGET` now, since the operator is no longer a semi-join.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1226,6 +1226,13 @@ and still reports it once. The CSR-backed variant reads the multiplicity off the
 stopping at the first hit, and now steps aside for the edge-list walk when the hop has to tell one relationship
 from another - which adjacency ids cannot do.
 
+A graph analytical view keeps its pending deletions per `(source, target)` pair, because the CSR holds adjacency
+ids and has no edge identity to key on, so deleting one of several parallel edges masks all of them. That is why
+`isConnectedTo` can report a pair as gone while two of its edges remain. A boolean can absorb that; a row count
+cannot. A masked pair is therefore reported as "cannot say", and the hop falls back to the edge list, which is
+exact - so a `SYNCHRONOUS` view that has seen such a delete answers the pattern with the right number of rows
+rather than none.
+
 One related source of confusion is gone with it: two equally cheap hops were tie-broken by a `HashSet` iteration
 order over identity hash codes, so the same query could be planned differently from run to run. Ties now fall to
 the order the hops are written in.
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1191,6 +1191,13 @@ that the `METADATA`-clause reader has to reject as typos.
 
 ## Cypher: a hop onto an already-bound vertex counts every relationship joining the pair
 
+> **Upgrading? Row counts can go up, and that is the fix.** A pattern with a hop onto an already-bound vertex - most
+> commonly a cycle, whose closing hop always has both endpoints bound - was returning fewer rows than it should
+> wherever parallel edges join the same pair. It returns them all now, so `count(*)`, `collect()` and `sum()` over
+> such a pattern report larger numbers than they did in 26.7.2. The old numbers were under-counts, not a different
+> convention: a saved report, a golden-file test or a threshold calibrated against them is the thing to re-check. A
+> graph with at most one edge per pair per type is unaffected in every respect - there was nothing to under-count.
+
 A pattern relationship matches once per relationship. Two parallel edges between the same two vertices are two
 matches, whether or not the pattern names them - `MATCH (a)-[:R]->(b)` over a pair joined twice returns two rows,
 the same as Neo4j. The optimizer's operator for the hop whose far end is already bound did not: built as a
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1189,6 +1189,47 @@ Finally, restoring a vector index from an exported definition (`IMPORT DATABASE`
 distinct entry point: an exported definition legitimately carries structural keys - `type`, `bucket`, `version`, ... -
 that the `METADATA`-clause reader has to reject as typos.
 
+## Cypher: a hop onto an already-bound vertex counts every relationship joining the pair
+
+A pattern relationship matches once per relationship. Two parallel edges between the same two vertices are two
+matches, whether or not the pattern names them - `MATCH (a)-[:R]->(b)` over a pair joined twice returns two rows,
+the same as Neo4j. The optimizer's operator for the hop whose far end is already bound did not: built as a
+semi-join ("is this pair connected?"), it answered once per input row and threw the rest of the pair away
+([#5663](https://github.com/ArcadeData/arcadedb/issues/5663)).
+
+The shape where it shows is a cycle, because the hop that closes one always has both endpoints bound:
+
+```cypher
+CREATE (a:Account {code: 'HUB'})
+CREATE (t:Txn {ref: 'SHARED'})
+CREATE (a)-[:INITIATED {kind: 'payment'}]->(t)
+CREATE (a)-[:INITIATED {kind: 'refund'}]->(t)
+CREATE (t)-[:INITIATED {ref: 'REVERSED'}]->(a)
+
+MATCH (a:Account {code: 'HUB'})-[r1:INITIATED]->(t:Txn {ref: 'SHARED'})-[r2:INITIATED]->(a)
+RETURN r1.kind
+```
+
+The cycle can be walked two ways, one per first-hop edge, each closing through the single returning edge -
+relationship uniqueness is satisfied because `r1` and `r2` bind different edges in both walks. The answer was one
+row. Anything aggregating over such a pattern (`count(*)`, `collect(r1)`, `sum`) silently under-reported wherever
+parallel edges exist between a pair, which is normal in transaction and payment graphs. The step-by-step executor
+had always got this right, so the same query could answer differently depending on which plan was chosen.
+
+The operator is an expansion now: it walks the relationships joining the pair and emits one row per relationship,
+binding the relationship variable and enforcing relationship uniqueness against the hops that precede it in the
+same `MATCH` clause - including when the hop is anonymous, whose edge a later hop in the clause must still not
+reuse. It keeps what made it fast: the source's edge list is filtered on the neighbour pointer held in the edge
+segment, so an edge that does not reach the target costs a pointer comparison rather than a record load, and only
+the edges that do reach it are ever materialised. An undirected hop reaches a self-loop from both adjacency lists
+and still reports it once. The CSR-backed variant reads the multiplicity off the sorted adjacency array instead of
+stopping at the first hit, and now steps aside for the edge-list walk when the hop has to tell one relationship
+from another - which adjacency ids cannot do.
+
+One related source of confusion is gone with it: two equally cheap hops were tie-broken by a `HashSet` iteration
+order over identity hash codes, so the same query could be planned differently from run to run. Ties now fall to
+the order the hops are written in.
+
 ## `.github/dependency-review-config.yml` expresses the policy the action actually reads
 
 The file was organised into `security:`, `licensing:`, `packages:`, `changes:`, `exemptions:`, `notifications:`,
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -112,6 +112,12 @@ public interface GraphTraversalProvider {
    * The default implementation counts occurrences in {@link #getNeighborIds} and never answers
    * unknown; a CSR-backed provider overrides it with an equal-range scan on the sorted adjacency
    * array.
+   * <p>
+   * <b>The default halves the count of a {@link Vertex.DIRECTION#BOTH} self-loop</b>, because
+   * {@link #getNeighborIds} is specified to return the raw adjacency entries and a self-loop
+   * contributes one to each of the two lists - which is why the callers of that method de-duplicate
+   * it themselves (see {@code SelfLoops.deduplicate}). A provider whose {@code getNeighborIds}
+   * already de-duplicates would halve a count that was never doubled, and must override this method.
    */
   default long countEdgesBetween(final int nodeA, final int nodeB, final Vertex.DIRECTION direction,
       final String... edgeTypes) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -103,8 +103,15 @@ public interface GraphTraversalProvider {
    * {@link Vertex.DIRECTION#BOTH} a self-loop counts once, matching how the OLTP expansion
    * de-duplicates the relationship it reaches from both adjacency lists.
    * <p>
-   * The default implementation counts occurrences in {@link #getNeighborIds}; a CSR-backed provider
-   * overrides it with an equal-range scan on the sorted adjacency array.
+   * <b>A negative result means the provider cannot answer exactly</b> and the caller must fall back
+   * to the edge list. A count is a stronger claim than a boolean, and a provider that tracks its
+   * pending changes at a coarser granularity than the individual edge can hold the boolean while
+   * losing the count - saying "unknown" is then the only honest answer. The caller has the two
+   * vertices in hand and can always walk the edges itself.
+   * <p>
+   * The default implementation counts occurrences in {@link #getNeighborIds} and never answers
+   * unknown; a CSR-backed provider overrides it with an equal-range scan on the sorted adjacency
+   * array.
    */
   default long countEdgesBetween(final int nodeA, final int nodeB, final Vertex.DIRECTION direction,
       final String... edgeTypes) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -96,6 +96,29 @@ public interface GraphTraversalProvider {
   boolean isConnectedTo(int nodeA, int nodeB, Vertex.DIRECTION direction, String... edgeTypes);
 
   /**
+   * Counts the edges joining nodeA to nodeB in the given direction and of the given edge types.
+   * <p>
+   * This is the multiplicity {@link #isConnectedTo} collapses to a boolean: a pattern relationship
+   * matches once per edge, so a pair joined by parallel edges contributes one row per edge. Under
+   * {@link Vertex.DIRECTION#BOTH} a self-loop counts once, matching how the OLTP expansion
+   * de-duplicates the relationship it reaches from both adjacency lists.
+   * <p>
+   * The default implementation counts occurrences in {@link #getNeighborIds}; a CSR-backed provider
+   * overrides it with an equal-range scan on the sorted adjacency array.
+   */
+  default long countEdgesBetween(final int nodeA, final int nodeB, final Vertex.DIRECTION direction,
+      final String... edgeTypes) {
+    long count = 0;
+    for (final int neighbor : getNeighborIds(nodeA, direction, edgeTypes))
+      if (neighbor == nodeB)
+        ++count;
+    // A self-loop contributes one entry to each adjacency list, so a BOTH walk sees every one of them twice
+    if (direction == Vertex.DIRECTION.BOTH && nodeA == nodeB)
+      count /= 2;
+    return count;
+  }
+
+  /**
    * Returns a property value from columnar storage, or null if not materialized.
    */
   Object getProperty(int nodeId, String propertyName);
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -200,6 +200,11 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
   private DeltaCollector         deltaCollector;
   public static final int        DEFAULT_COMPACTION_THRESHOLD = 10_000;
   private static final int       MAX_PENDING_DELTAS           = 100_000;
+  /**
+   * Returned by {@link #countEdgesBetween} when the overlay makes the multiplicity unknowable, which
+   * the {@link com.arcadedb.graph.GraphTraversalProvider} contract spells as any negative value.
+   */
+  private static final long      MULTIPLICITY_UNKNOWN         = -1L;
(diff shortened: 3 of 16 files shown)


lvca:
Thanks - all three points addressed in da222cb34.

### 1. `countEdgesBetween` under-counting after an overlay deletion — **valid, fixed**

Confirmed the mechanism and traced it a step further: `DeltaOverlay.deletedEdgesPerType` is a `Set<Long>` of packed `(src,tgt)` pairs, and `copyBaseExcludingDeleted` consults the same mask per neighbour entry - so `getNeighborIds` already drops **all N** parallel entries of a masked pair, not just the deleted one. The view was internally consistent (`isConnectedTo` false, neighbours empty, count 0), just not true. `countEdgesBetween` inherited it and, as you say, turned it into rows.

I looked at fixing it at the source and it isn't reachable from here: the CSR stores adjacency ids with no edge identity, so there is no key that could name *which* of the parallel entries went. A per-pair counter cannot substitute either - the overlay deliberately absorbs a deletion replayed across merges by set semantics (#4587), and a counter would double-decrement.

So the honest answer is the one the provider can actually stand behind: **a masked pair is reported unknown**. `GraphTraversalProvider.countEdgesBetween` now documents a negative result as "the provider cannot answer exactly, walk the edge list", `GraphAnalyticalView` returns it for any counted type whose pair the overlay masks, and `GAVExpandInto` falls back to `countConnectingOLTP` - the same fallback it already used for a vertex missing from the mapping, and exact. A `SYNCHRONOUS` view that has seen such a delete now answers the pattern with the right number of rows instead of none.

Two tests:
- `GraphAnalyticalViewTest.countEdgesBetweenAnswersUnknownWhenTheOverlayMasksAPair` - three parallel `A->B` edges, delete one, count goes negative, pinned alongside `isConnectedTo` returning false and `getNeighborIds` omitting the pair, so the relationship between the three is documented rather than implied. A pair the deletion did not touch still answers exactly.
- `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` - the end-to-end `count(*)`: 4 before the delete, 2 after. Mutation-checked: with the fallback replaced by `0` it fails `expected: 2L but was: 0L`, so it is pinning the fallback and not the arithmetic.

### 2. Misleading comment — **valid, fixed**

Right on both counts: 2 per first hop, and the assertion was already correct. Reworded to say the hop sees all three and keeps the two that are not the edge the first hop bound.

### 4. Other implementors of the SPI

Checked: `GraphAnalyticalView` is the only production implementor, and the only other is a test stub in `PartitionedTriangleOpCallerRunsTest`, which takes the `default`. Nothing overrides `countEdgesBetween` expecting boolean-only behaviour.

### Re-run

`com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**`: 8178 tests, 0 failures. One note in case it shows up in CI: `CypherExistsSuperNodeBenchmark.anchoringOnTheLeafAvoidsTheSuperNodeScan` asserts `fromLeaf < fromHub` at integer-millisecond resolution, and both shapes are now sub-millisecond thanks to the neighbour-pointer narrowing already on `main`, so the strict `<` can tie at 0. It is untouched by this diff - its guard carries inline relationship properties, which routes it away from the optimizer entirely - and it passes cleanly; flagging it as a latent resolution issue in that benchmark, not as something this PR introduced.

lvca:
Thanks. Pushed aa9ab4bd9 for observation 1 plus a Codacy style finding.

**1. Default `countEdgesBetween` self-loop `/2` — done.** You are right that it was load-bearing and unwritten. The Javadoc now states the assumption explicitly - `getNeighborIds` is specified to return the raw adjacency entries, and a self-loop contributes one to each of the two lists, which is exactly why the existing callers de-duplicate it themselves via `SelfLoops.deduplicate` - and says what a provider whose `getNeighborIds` already de-duplicates has to do (override the method, because it would otherwise halve a count that was never doubled).

**2 and 3 — no change,** agreed with your reading. The per-row property copy is inherent to emitting one row per relationship and is byte-for-byte what `ExpandAll` already does, so changing it here would be an unrelated optimisation of both. The non-`VertexInternal` branch is a defensive fallback the Javadoc already flags as the thing the operator exists to avoid.

**Codacy** flagged one new minor CodeStyle issue on the second commit: `MULTIPLICITY_UNKNOWN` was declared mid-file next to the pre-existing `EMPTY_INT`. Moved up to the class's field declarations. I left `EMPTY_INT` where it is rather than widen the diff into unrelated churn.

Re-ran `com.arcadedb.graph.**` plus the Cypher classes this touches: 410 tests, 0 failures.

lvca:
Thanks. Pushed 8a36df95d.

### 1. Existence-pattern predicates — checked, and now pinned by a test

Good question to ask rather than assume. Traced it: a pattern used as a `WHERE` predicate goes through `PatternPredicateExpression`, which evaluates as a correlated subquery read only for `hasNext()`, so the row count inside it never escapes as cardinality. The operator's change of meaning cannot leak into the predicate.

That was an argument, not evidence, so `anExistencePredicateOverAPairJoinedTwiceStillAnswersOnce` now pins it over the pair joined by two `INITIATED` edges: `WHERE (a)-[:INITIATED]->(t)`, the `EXISTS { }` spelling, and a `NOT EXISTS { }` negation each answer **1**, while the same pattern written in `MATCH` answers **2**. The contrast is in one test on purpose - it is the distinction that would break silently.

### 3. Property-name hoist — done

`getPropertyNames()` is now read once per input row instead of once per emitted row.

### 2 and 4 — acknowledged, no change

Agreed on both, and both are already stated where someone would look: the walk-the-whole-narrowed-list trade in the class Javadoc and the PR body, and the `getNeighborIds` coupling in the SPI Javadoc added last round.

### One thing your review did not raise, but the runs did

`CypherExistsSuperNodeBenchmark.anchoringOnTheLeafAvoidsTheSuperNodeScan` was failing about one run in three on `expected 0L to be less than 0L`. It compared the two anchorings in **truncated integer milliseconds**, and the neighbour-pointer narrowing already on `main` brought both under a millisecond at this degree - so both truncated to 0 and the strict `<` became a coin flip. Nothing to do with this change (its guard carries inline relationship properties, which routes it away from the optimizer entirely), but it is `@Tag("benchmark")` and the root pom's `excludedGroups` is commented out, so it runs in CI and would have flaked this PR. It compares nanoseconds now and logs milliseconds to three decimals; 3/3 clean in isolation.

### Re-run

`com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**`: **8179 tests, 0 failures**, including the benchmark.

lvca:
Thanks. Pushed 5594c2c8c for point 2, and filed point 1 rather than guessing at it.

### 2. Default `countEdgesBetween` has no real-provider coverage — fixed

You are right that this is the branch nothing exercises, and it is the one carrying a load-bearing assumption. `GraphTraversalProviderCountEdgesBetweenTest` now pins the default against a minimal provider that returns raw adjacency ids exactly as the contract specifies: parallel edges counted once each in `OUT` / `IN` / `BOTH`, two self-loops reported as **two** relationships rather than four under `BOTH` (the `/2` this note is about), and an unconnected pair answering zero rather than unknown. Non-vacuous by construction - without the halving the `BOTH` self-loop assertion reads 4.

### 1. Cardinality estimate — real, filed as #5690 rather than patched here

Correct observation, and it is now wrong in the opposite direction from before: `DEFAULT_EXPAND_INTO_SELECTIVITY = 0.1` encodes "this hop filters", which was a safe over-count when the operator was a semi-join and is an under-count now that it can multiply.

I did not change it, because there is no principled replacement available from the statistics the optimizer has. `getAverageDegree(type, sourceLabel, targetLabel)` answers "how many edges does a vertex have", which does not determine "how many edges join a *given pair*" - degree 1000 over 1000 neighbours and degree 1000 to one neighbour are indistinguishable to it and have opposite multiplicity. Swapping in a constant above 1 would trade a roughly-right estimate on every simple graph for a roughly-right one on multigraphs, which is not a trade worth making blind.

What it actually wants is a mean-edges-per-connected-pair statistic per edge type - 1.0 on a simple graph, so today's behaviour is unchanged where today's behaviour is right - and the open question is collecting `distinctConnectedPairs`, which is not derivable from the bucket counters and which a GAV's sorted adjacency array can answer exactly in a linear scan. Written up with that scope in **#5690**, including whether `createExpandAllOperator`'s `DEFAULT_AVG_DEGREE` has the mirror-image problem.

### 3. Synthetic var prefix — no change

Agreed it is an implicit coupling, but it is the existing `ExpandAll` convention and diverging from it here would leave two conventions for one mechanism. Better changed for both at once, if at all.

### Re-run

`com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**`: **8180 tests, 0 failures.**

Unless the next pass turns something up, I consider this ready - three rounds of review with nothing blocking, every actionable point applied, and the one that could not be applied honestly written up rather than guessed at.

lvca:
Thanks. Pushed a2e9701ba: 1, 2 and 4 applied, 3 acknowledged.

### 1. Behavioural change should be prominent — done

Agreed, and the note buried it in the third paragraph. It now opens with an upgrade callout in the same form the other breaking entries in this release use: row counts over these patterns go up, the old numbers were **under-counts rather than a different convention**, so a saved report or a threshold calibrated against them is the thing to re-check - and a graph with at most one edge per pair per type is unaffected in every respect, since there was nothing to under-count.

### 2. Non-`VertexInternal` fallback built a list — fixed rather than documented

You are right that it defeated the streaming property, so I made it lazy instead of explaining why it was acceptable: it is a filtering iterator now, so the operator's batching keeps bounding how much of a super-node's list is walked.

And the note you asked for, on when that branch actually occurs: **it does not, today.** Every vertex an operator binds comes from a scan, a seek or an expansion, and all three yield engine vertices. The branch keeps the operator total against a handle arriving from somewhere else rather than serving a known caller, and the Javadoc says exactly that now instead of leaving a reader to wonder which caller it is for.

### 4. `countConnectingOLTP` and the planner gate — comment added

The best catch of this round. The fallback is correct only because `createExpandIntoOperator` routes to `GAVExpandInto` solely for a hop with no same-clause preceding rel var and no edge tracking - and the operator *cannot* start checking, because it never binds the edges it counts. That coupling is now written at the fallback, naming the gate, so widening the gate has to confront it.

### 3. Default allocates the full neighbour array — no change

Correct, and there is no narrower primitive in the SPI to build on. Only a hypothetical future provider pays it, since `GraphAnalyticalView` overrides; a provider that cares would override too, which is the escape hatch the contract already offers.

### Verification

`com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**`: **8180 tests, 0 failures**, including the OpenCypher TCK. Agreed CI should confirm before merge given the semantic shift - the full set is running on this PR.

Five rounds, nothing blocking in any of them, every actionable point applied, and the one that could not be applied honestly (the cardinality estimate) written up as #5690 rather than guessed at. I consider this done unless the next pass turns something up.

lvca:
Thanks - and thanks for the independent verifications, particularly checking that `countEdgesBetween`'s empty-`edgeTypes` iteration over `csrPerType.keySet()` mirrors `isConnectedTo` rather than introducing a new divergence. That was the shape of mistake I was most worried about in the CSR override.

**Observation 2 is a real find and is now #5691.** You are right that it is an asymmetry rather than a regression: the `ExpandAll` branch asks whether anyone reads the variable and the `ExpandInto` branch takes it straight off the pattern, so `MATCH (a:A {id:1})-[r:R]->(b:B {id:2}) RETURN a, b` is a pure adjacency walk on one branch and an edge-materialising hop on the other. The consequence is narrow but is exactly the case a view exists for: a named-but-unread variable keeps the hop off the CSR path where a `GraphAnalyticalView` covers the type. Without a view there is nothing to gain, since after this PR `ExpandInto` walks the narrowed list either way to know the multiplicity.

I left it out of this PR deliberately - it changes plan selection, so it deserves its own before/after rather than riding along at the end of a review cycle, and the fix has to keep the uniqueness gate independent of `edgeIsMaterialized` (a `GAVExpandInto` still cannot tell one relationship from another however unread the variable is). Both constraints are written into #5691.

**Observation 1** - agreed, the "Upgrading?" callout stays.

**Observation 3** - agreed that `mvn verify` is the gate, and per CLAUDE.md a bare `test` reactor run would not even produce the artifacts `arcadedb-gremlin-it` consumes. That gate is CI's to apply here, and it is running the full set on this PR: `unit-tests`, `slow-unit-tests`, `integration-tests`, `ha-integration-tests`, `builder-tests` and the six e2e suites. My local runs were the targeted modules only, and I have said so each round rather than implying more.

Current state: `claude-review`, Codacy (**0 issues**, back to "up to standards" after the constant move), CodeQL, Meterian and the analysis jobs are green; `build-and-package` and `Analyze (java-kotlin)` still running, with the test jobs behind them.

Follow-ups filed from this review cycle: **#5690** (the bound-target hop's cardinality estimate assumes it filters, but it can multiply) and **#5691** (the asymmetry above).

lvca:
## CI result: green except a pre-existing HA failure, with evidence

`ha-integration-tests` failed, so I checked whether this branch caused it. **It did not** - both failing tests fail on `main` without this change.

**Failed here** (run 30689622260):
- `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied` - "Node 1 must hold both vertices, expected 2 but was 1", with a page-version divergence between replicas
- `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact`

**Same tests on `main`:**

| `main` run | failing HA Raft ITs |
|---|---|
| 30687024879 | `Issue5569SlotMergeDeleteRaftIT` |
| 30686802579 | `Issue4743LargeRecordReplicationIT`, `Issue5569SlotMergeDeleteRaftIT`, `RaftHTTPGraphConcurrentIT`, `RaftPriorityRejoinIT`, `RemoteStickyConnectionStrategyIT` |
| 30685641431 | **`Issue5410AbandonedTicketReleaseIT`**, `Issue5569SlotMergeDeleteRaftIT` |
| 30684063637 | `Issue5569SlotMergeDeleteRaftIT`, `RaftReplicationMaterializedViewIT` |

Both of this PR's failures appear there, and the cast rotates run to run - the signature of an environmental problem on the HA runner, not a regression. `Issue5569SlotMergeDeleteRaftIT` fails in every one of those four.

The mechanism agrees: these are Raft log replication and page-version convergence. This diff is confined to Cypher query planning and execution (`ExpandInto`, `GAVExpandInto`, `ExpandIntoStep`, `CypherOptimizer`, `JoinOrderRule`) plus one read method and one constant on `GraphAnalyticalView`. Nothing in it participates in replicating a log entry or versioning a page.

**Worth noting in the other direction:** `unit-tests` currently fails on `main` (`Issue5279ConcurrentUpdateTest`, `Issue5326ApplyChangesPendingFlushTest`) and **passes on this branch**, along with `slow-unit-tests`, `integration-tests`, `builder-tests`, `build-and-package` and all six e2e suites.

### `codecov/project`

`codecov/patch` **passed** (78.9% on the new lines). `codecov/project` reports -0.25%, but its own diff shows **`Hits -295`** project-wide - 109 added lines cannot subtract 295 hits. That is the failed `ha-integration-tests` job's coverage missing from the upload, compounded by the report being "7 commits behind head on main". It is an artifact of the same pre-existing failure, not of this change.

### Where that leaves the PR

Six review rounds, the last one **LGTM pending CI**. Every actionable point applied; the two that could not be applied honestly are filed as #5690 (cardinality estimate) and #5691 (named-but-unread rel var asymmetry). No merge conflict - verified locally against the 6 newer commits on `main`, which merge cleanly.

I have taken this as far as I can without a green HA runner, and that gate is failing independently of this branch.

PR #5689, opened 2026-08-01
fix(engine) #5543: expose the logger implementation through GlobalConfiguration

### LINKED ISSUE #5543: Expose arcadedb.log.impl through GlobalConfiguration instead of only a raw system property
Since #5361, ArcadeDB can log through the SLF4J facade, selected with the `arcadedb.log.impl` system property. That property is read in `LogManager.createLogger()` via a direct `System.getProperty(...)` call, and there is no corresponding entry in `GlobalConfiguration`.

This makes it the only ArcadeDB setting that bypasses ArcadeDB's own configuration system. Consequences for embedders:

- It cannot be set through `ContextConfiguration` / `GlobalConfiguration` like every other setting, so it does not participate in `dumpConfigAtStartup`, per-instance configuration, or any config-file mechanism.
- Because `createLogger()` runs from the `LogManager` static initialiser, the property has to be set before the class is first touched. In a framework-managed application (Spring Boot, for instance) the natural configuration hook runs later than that, so the value cannot come from the application's own configuration.
- It is JVM-wide, so an application embedding ArcadeDB alongside other components cannot scope the choice.

`LogManager.setLogger(Logger)` is a good escape hatch and works well. It is what we ended up using in our service. But it requires the embedder to construct the implementation itself rather than just expressing the choice as configuration.

**Proposal:** add a `GlobalConfiguration` entry for the logger implementation, defaulting to `default` (java.util.logging) so behaviour is unchanged, and have `LogManager` resolve it from there. The system property would keep working exactly as it does today for the pre-class-load case.

Happy to send a PR if this seems reasonable.

*(Unrelated nit spotted nearby: `LogManager.setLogger(Logger)` has no javadoc, unlike the rest of the class. Could fold into the same PR.)*

robfrank:
Closes #5543

Since #5361 the logger implementation is selected with the `arcadedb.log.impl` system property, read in `LogManager.createLogger()` through a direct `System.getProperty(...)` call. That made it the only ArcadeDB setting outside ArcadeDB's own configuration system: it never appeared in `dumpConfigAtStartup` or any config-file mechanism, and because `createLogger()` runs from the `LogManager` static initializer, the value had to be set before the class was first touched, which a framework-managed application (Spring Boot) cannot always guarantee. This adds `GlobalConfiguration.LOG_IMPL` (`arcadedb.log.impl`, `SCOPE.JVM`, default `default`), whose callback installs the matching logger on the running `LogManager` via the extracted `LogManager.createLogger(String)`. The system property keeps working through the very same resolution routine, so behaviour is unchanged when the setting is not touched. `LogManager` deliberately still reads the raw property rather than the configuration, because querying `GlobalConfiguration` from its static initializer would run that class's whole initialization, callbacks included, while `LogManager.instance()` is still `null` - the callback null-guards for the same reason. The issue's nit is folded in: `setLogger(Logger)` now has javadoc.

## Test plan

- [x] `LogImplConfigurationTest` written first: did not compile before the fix (`cannot find symbol: variable LOG_IMPL`)
- [x] Callback neutered after the fix to prove the assertions are load-bearing: 4 of 6 tests fail, the 2 survivors being the registration/default-value checks that never reach the callback
- [x] Setting `LOG_IMPL` to `slf4j` after startup installs `Slf4jLogger`; setting it back to `DEFAULT` (case-insensitive) installs `DefaultLogger`
- [x] An unrecognized value falls back to `DefaultLogger` and is reported on `System.err`, matching the system-property behaviour from #5361
- [x] The setting applies through `GlobalConfiguration.fromJSON(...)`, and appears in `dumpConfiguration`
- [x] Four separate JVMs probe the class-initialization orderings that a single-JVM unit test cannot reach: property unset / `slf4j` / `bogus`, crossed with `LogManager` vs `GlobalConfiguration` loaded first. All four install the expected logger (see `docs/5543-log-impl-global-configuration.md` for the table)
- [x] `mvn -pl engine test`: 10690 tests, 0 failures, 0 errors, 23 skipped
- [x] `mvn compile` over the full reactor: green


robfrank:
Answering the two points that asked for an explicit decision, and closing out the rest.

**1. An unrecognized value is stored verbatim - intended.** The alternative, storing `default`, makes the config dump agree with the installed logger but destroys the only lasting record of the typo: the `System.err` warning fires once at set time and is gone. A dump reading `arcadedb.log.impl = slf4` next to a warning in the log is diagnosable; one reading `default` looks like nobody ever configured anything, which is the failure mode #5361 already fixed once by adding that warning. The setting's own description states the fallback, and `LogManager.getLogger()` is the accurate answer for "what is actually installed" - `LOG_IMPL.getValue()` answers "what was asked for", which is what every other setting means too (`MAX_PAGE_RAM` is the only one that rewrites, and it rewrites to a *valid* value it chose, not away from an invalid one the user typed).

**2. `reset()` leaving the logger installed - noted in code, not changed.** `reset()` runs no callback for any setting; it assigns `defValue` (or `callbackIfNoSet`). Making this one entry re-install would mean either a special case inside `reset()` or a contract change across all ~200 entries, neither of which belongs in this issue. Added the note you suggested as a comment on the entry (50265eeb7), and it is under Impact in the tracking doc.

**4. Double normalization** is inherent: `createLogger(String)` is public and also fed the raw system property, so it has to normalize its own input; the callback has to normalize separately because it stores the result. It is a trim plus lowercase of a five-character string, once per configuration change.

**3. Test isolation** - declining consistently with the previous cycle. `@Isolated` today would be an annotation guarding a mode the module does not run in, and the existing `GlobalConfigurationTest` / `ContextConfigurationTest` mutate the same kind of process-global state without one. If in-JVM parallel execution is ever switched on, that whole set moves together.

**5.** Agreed, and unchanged from #5361.

Issue #5477, opened 2026-07-28
[OpenCypher] size() returns null for unsupported argument types instead of raising a type error


shulei5831sl:
## Environment
- Host OS: Linux x86_64
- Deployment: Docker
- Query interface: HTTP command API
- Query language: `opencypher` and `cypher` (both affected)
- ArcadeDB version: 26.8.1-SNAPSHOT
- Build: efa5747e76bb6799dda4765fef51968a578438c5/1784810487393/main
- Docker image: `arcadedata/arcadedb:latest`
- RepoDigest: `arcadedata/arcadedb@sha256:1cb78c08e2e40e765f18cd1b06be83d878d7e13c01471db6bddb112a53d0de74`
- API endpoint: `POST http://localhost:2494/api/v1/command/test`

## Minimal reproducer

```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN size(42) AS r"}'
```

Response:
```json
{"user":"root","result":[{"r":null}]}
```

## Matched valid-input controls

String control:
```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN size(\"abc\") AS r"}'
```
Response: `{"user":"root","result":[{"r":3}]}`

List control:
```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN size([1,2,3]) AS r"}'
```
Response: `{"user":"root","result":[{"r":3}]}`

## Null-propagation control

`size(null)` correctly returns `null` per Cypher null-propagation rules.
This is the expected behavior and not a bug. The issue is that invalid
scalar arguments (INTEGER, FLOAT, BOOLEAN, MAP) also silently return
`null`, conflating a type error with legitimate null propagation.

```bash
curl -u root:<password> \
  -H "Content-Type: application/json" \
  -X POST \
  http://localhost:2494/api/v1/command/test \
  -d '{"language":"opencypher","command":"RETURN size(null) AS r"}'
```
Response: `{"user":"root","result":[{"r":null}]}` (correct)

## Type matrix

| Input | Query | Result | Notes |
|-------|-------|--------|-------|
[...]

Issue #5193, opened 2026-07-10
Explicit null length in `substring()` is treated as an omitted argument


shulei5831sl:
### ArcadeDB version

Observed on the latest Docker image:

```text
docker pull arcadedata/arcadedb:latest
Pulled: 2026-07-10

ArcadeDB version: 26.8.1-SNAPSHOT
```

The behavior was also reproduced in a fresh isolated Docker container, 3 out of 3 runs.

### Environment

* Host OS: x86_64 Linux `5.4.0-216-generic`
* Deployment: Docker `arcadedata/arcadedb:latest`
* Query interface: HTTP command endpoint
* Query language: Cypher / OpenCypher
* Differential comparison targets:

  * Neo4j `2026.05.0`
  * Memgraph `3.11.0`

### Description

ArcadeDB treats an explicitly supplied `null` third argument to `substring()` as though the argument were syntactically omitted.

As a result, these two semantically different calls return the same value:

```cypher
RETURN substring('hello', 1, null) AS r;
RETURN substring('hello', 1) AS r;
```

ArcadeDB returns `ello` for both.

The two-argument form intentionally omits the optional `length` argument and should return the remainder of the string. The three-argument form explicitly supplies a `null` value for `length` and should not be silently converted into the two-argument overload.

### Minimal reproduction

```cypher
RETURN substring('hello', 1, null) AS r;
```

### Expected behavior

An explicitly supplied `null` length should remain distinguishable from an omitted length.

The query should either:

* return `null` through null propagation; or
* fail with a normal argument type/value error.

It should not behave as if the third argument were absent.

Observed comparison-engine behavior:

```text
Neo4j 2026.05.0: null
Memgraph 3.11.0: argument type error
```

Both behaviors preserve the distinction between explicit `null` and argument omission.

### Actual behavior

ArcadeDB returns:

```text
r = ello
```

This is exactly the result of the two-argument overload:

```cypher
RETURN substring('hello', 1) AS r;
```

#### Explicit null length

```cypher
RETURN substring('hello', 1, null) AS r;
```

ArcadeDB:

```text
r = ello
```

#### Length omitted

```cypher
RETURN substring('hello', 1) AS r;
```

ArcadeDB:

```text
r = ello
```

This result is correct for the two-argument form.

#### Explicit numeric length

```cypher
RETURN substring('hello', 1, 3) AS r;
```

ArcadeDB:

```text
r = ell
```

This result is correct.

### Cross-engine comparison

[...]

PR #5699, opened 2026-08-01
fix(engine) #5629: an explicit null in an optional argument propagates

### LINKED ISSUE #5629: [Cypher] Decide whether an explicit null in an optional argument propagates or means "use the default"
Raised in review of #5612 and deliberately not changed there, because it is a convention question rather than a bug in one function.

Cypher functions with an optional trailing argument currently treat an **explicit** `null` for it as "argument omitted, use the default":

| call | today | |
|---|---|---|
| `normalize('x', null)` | normalizes as NFC | `NormalizeFunction.parseNormalForm` |
| `isNormalized('x', null)` | tests against NFC | same helper |
| `format(date, null)` | `toString()` | `FormatFunction` |
| `round(3.14, 2, null)` | HALF_UP | `RoundFunction` |
| `vector.distance(a, b, null)` | EUCLIDEAN | `VectorDistanceFunction` |

The same functions do propagate `null` from their **first** argument (`normalize(null)` is `null`), so the two positions behave differently. Cypher's usual rule is that `null` in any argument propagates to `null` out, which would make `normalize('x', null)` answer `null` rather than the NFC default.

Neo4j's behaviour should be checked before changing anything - I could not verify it from here, which is why #5612 left it alone. The current behaviour is consistent across the five functions above and predates that PR.

Two ways to settle it, both preferable to changing one function:

1. **Propagate.** An explicit `null` optional argument makes the call answer `null`, matching first-argument behaviour and general Cypher null semantics. Changes all five.
2. **Keep, and say so.** Document "an explicit `null` in an optional position means the default" as the house rule, and note it on the shared helpers so the next function follows it.

The one thing worth avoiding is settling it per-function: that is how the arity declarations drifted in #5484, and how `normalize()` and `isNormalized()` came to disagree about their input domain (fixed in #5612 by making both STRING-only).

robfrank:
Closes #5629

## The decision

The issue asked to settle whether an explicit `null` in an optional argument position propagates or means "use the default", and explicitly warned against settling it per-function. It also asked for Neo4j's behaviour to be checked first, which the reporter could not verify.

**Settled in favour of propagation.** Three independent lines of evidence:

1. **Neo4j propagates.** From the Cypher manual source (`neo4j/docs-cypher`): `round()` "returns `null` if any of its input parameters are `null`"; `replace(original, search, replace [, limit])` "If any argument is `null`, `null` will be returned"; `btrim("hello", null)` and `ltrim("hello", null)` both "return `null`". `round()` is the direct counterpart of the case in the issue. (Neo4j is not *uniformly* propagate - `substring` and `left`/`right` raise on a null length - but in no documented case does it read an explicit `null` as "argument omitted". That is the invariant adopted here.)
2. **ArcadeDB had already decided this once.** `CypherSubstringFunction` carries `Issue #5193: an explicitly supplied null length propagates null (as Neo4j does), it must not be treated as an omitted argument.` Same reasoning, same citation - just never written where the next function would find it.
3. **It was already the majority.** Auditing all 22 stateless functions whose `getMinArgs()` differs from `getMaxArgs()`: **9 already propagated** (the five temporal constructors, `point()`, `ltrim()`, `rtrim()`, `vector_create()`), 11 silently defaulted, 1 throws by design (`range()`, #5477). Option 2 would have meant changing the ones that were already right.

## The rule

> An explicit `null` in an optional argument position is never "argument omitted". Omitting the argument selects the function's default; writing `null` there is subject to the usual null-in/null-out rule.

Stated once, on `CypherFunctionHelper.isExplicitNull(args, position)`, and called rather than re-decided. That is the mechanism the issue asked for - settling it one function at a time is how the arity declarations drifted in #5484 and how `normalize()` and `isNormalized()` came to disagree in #5602.

## Changes

Eleven functions adopt it - the five named in the issue plus six the audit found:

| function | optional argument | before | after |
|---|---|---|---|
| `normalize` / `isNormalized` | normal form | NFC | `null` |
| `format` | pattern | `toString()` | `null` |
| `round` | rounding mode | HALF_UP | `null` |
| `vector.distance` | metric | EUCLIDEAN | `null` |
| `date`/`datetime`/`localdatetime`/`time`/`localtime`` .truncate` | adjustment map | map ignored | `null` |
| `SubstringFunction` | length | to end of string | `null` |

**Omitting the argument still selects the default in every one of them**; only the explicitly-written `null` changes.

### Two of the five named functions were not where the issue pointed

Both are dual-path cases:

- **`substring()`** - Cypher resolves it to `CypherSubstringFunction`, which already propagated (#5193). `com.arcadedb.function.text.SubstringFunction` is a *second* implementation that no factory registers and that still defaulted. Aligned here so the divergence does not surface the day it is wired up; covered by a direct unit test since no query can reach it.
- **`vector_distance()`** - in Cypher this is a grammar rule whose metric is a keyword, so `vector_distance(a, b, null)` does not parse (the same situation as Neo4j's `normalize()`). `VectorDistanceFunction` is reached from Cypher as **`vector.distance()`**, where the metric is an ordinary expression. That is the path fixed and tested.

### Ordering is preserved

Propagation is placed *after* the existing argument checks, not before, so a null in the optional position cannot mask a client error in an earlier one: `round(null, 2, 'SIDEWAYS')` still reports the unusable mode, and a bad truncate unit or non-temporal value is still reported. Both are pinned by tests.

## Test plan

- [x] `CypherOptionalArgumentNullIssue5629Test` (new, 17 tests) - for each of the eleven functions: the explicit `null` answers `null`, **and** omitting the argument still selects the documented default. Confirmed failing before the fix (11 failures), passing after.
- [x] Pins the nine functions that already propagated, so the convention cannot drift back.
- [x] Pins the validation-before-propagation ordering for `round()` and the truncate family.
- [x] `mvn -pl engine test -Dtest='Cypher*Test,OpenCypher*Test,TextStatelessFunctionsTest,*TemporalFunction*Test,*VectorFunction*Test,SQLFunctionPhase*Test,*StatelessFunction*Test'` - **3060 tests, 0 failures**.
- [x] Full `mvn -pl engine test` run as a backstop.

### Two existing assertions changed

Both encoded the behaviour this issue decided to change; both were found by running the suite, and each leaves what the test was written to prove intact.

1. `CypherNumericFunctionArgumentIssue5484Test.theRoundingModeOfRoundIsNotANumericArgument` asserted `round(3.14159, 2, null) == 3.14`, now expects `null`. Its point - that the mode position is not rejected as non-numeric - is still pinned by the `'FLOOR'` and `'CEILING'` cases.
2. `TextStatelessFunctionsTest.formatFunctionNullPattern` asserted a null pattern answers `toString()`, now expects `null`, and gained an assertion that *omitting* the pattern still answers the ISO string.

## Follow-ups (not in this change, deliberately)

1. The `*.truncate` family raises raw JDK exceptions for client mistakes: `NullPointerException` on a null unit (`args[0].toString()` is unguarded; `TrimFunction` has the same exposure on its mode), and `IllegalArgumentException: Unknown truncation unit` from `TemporalUtil`. Both surface as HTTP 500 for the caller's error - same class as #5484. The new test asserts only the *message* for the unknown-unit case, so it does not encode the wrong exception class as correct.
2. `trim(mode, trimChar, source)` defaults a null `trimChar` to whitespace, disagreeing with `ltrim`/`rtrim` beside it and with Neo4j's `btrim("hello", null)`. Its 1-or-3 arity means the argument is not trailing, so it is a different shape.
3. `range()` throws on a null step rather than propagating, by explicit decision in #5477 - worth confirming that is still intended now the convention is written down.
4. `substring()` diverges from Neo4j here: Neo4j raises for a null `start` or `length`, ArcadeDB propagates for both. Propagating is internally consistent (its null `start` already answered `null`, and #5193 chose this for the Cypher-facing implementation), but the divergence is worth a decision.
5. `com.arcadedb.function.text.SubstringFunction` is registered by no factory - either wire it up or delete it, rather than keeping two implementations of one function.

Full reasoning and the complete audit table are in `docs/5629-cypher-explicit-null-optional-argument.md`.

### DIFF
--- a/docs/5629-cypher-explicit-null-optional-argument.md
+++ b/docs/5629-cypher-explicit-null-optional-argument.md
@@ -1,0 +1,171 @@
+# 5629 - An explicit `null` in an optional argument propagates
+
+Issue: https://github.com/ArcadeData/arcadedb/issues/5629
+
+## The question
+
+Cypher functions with an optional trailing argument were reading an **explicit** `null` in that position as
+"argument omitted, use the default", while the same `null` in the first position propagated. The issue asked for
+one of two settlements, and explicitly warned against settling it per-function:
+
+1. **Propagate** - an explicit `null` optional argument makes the call answer `null`.
+2. **Keep, and document** - "an explicit `null` in an optional position means the default" becomes the house rule.
+
+## Decision: propagate
+
+Three independent lines of evidence, all pointing the same way.
+
+### 1. Neo4j propagates
+
+The issue asked for Neo4j's behaviour to be checked first. From the Cypher manual source
+(`neo4j/docs-cypher`, `modules/ROOT/pages/functions/`):
+
+| function | documented consideration |
+|---|---|
+| `round(value[, precision, mode])` | "`round()` returns `null` if any of its input parameters are `null`." |
+| `replace(original, search, replace [, limit])` | "If any argument is `null`, `null` will be returned." |
+| `btrim(original [, trimCharacter])` | "`btrim("hello", null)` returns `null`." |
+| `ltrim` / `rtrim` | "`ltrim("hello", null)` returns `null`." |
+
+`round()` is the direct counterpart of the case raised in the issue, and it propagates. `replace()`'s `limit` and
+`btrim()`'s `trimCharacter` are optional arguments whose omission selects a default, exactly the shape under
+discussion, and both propagate.
+
+Two of the five functions named in the issue have no Neo4j reference point at all: `format()` and
+`vector_distance()` are ArcadeDB extensions. A third, `normalize()`, cannot express the case - Neo4j declares
+`normalForm` as a keyword type `[NFC, NFD, NFKC, NFKD]` rather than an expression, so `normalize('x', null)` does
+not parse there. `isNormalized()` is an operator in Neo4j (`IS NORMALIZED`), not a two-argument function. So
+`round()` carries the weight, and it is unambiguous.
+
+Neo4j is **not** uniformly "propagate": `substring(original, start, length)` and `left`/`right` raise an error on
+a null length rather than answering null. But in no documented case does Neo4j read an explicit `null` as
+"argument omitted". That is the invariant worth adopting.
+
+### 1b. ArcadeDB had already decided this once, for `substring()`
+
+`CypherSubstringFunction` - the implementation Cypher actually resolves `substring()` to - carries this comment,
+added by issue #5193:
+
+> `Issue #5193: an explicitly supplied null length propagates null (as Neo4j does), it must not be treated as an
+> omitted argument.`
+
+So the rule this issue asks for is not a new decision. It was made, for one function, with the same reasoning and
+the same citation, and then not written anywhere the next function would find it. That is precisely the drift the
+issue is about, and it is the argument for putting the rule on a shared helper rather than in a comment.
+
+### 2. ArcadeDB already propagated, in most places
+
+An audit of all 22 stateless Cypher functions whose `getMinArgs()` differs from `getMaxArgs()`:
+
+- **9 already propagate**: the five temporal constructors (`date`, `datetime`, `localdatetime`, `time`,
+  `localtime`), `point()`, `ltrim()`, `rtrim()`, `vector_create()`. `VectorCreateFunction` even carries an
+  explicit `// Null propagation: if dimension is explicitly null, return null` comment.
+- **11 silently defaulted**: the five named in the issue, plus the five `*.truncate` functions and `substring()`.
+- **1 throws**: `range()`, deliberately, for issue #5477.
+
+So propagation was already the majority reading, and the defaulting group was drift rather than a considered
+house rule. Option 2 would have meant changing the 9 that were already right.
+
+### 3. The two positions must agree
+
+`normalize(null)` answers `null` but `normalize('x', null)` normalized as NFC. The same absent value meant two
+different things depending on which position it landed in, which is the property that makes a wrong query look
+like a successful one.
+
+## The rule
+
+> An explicit `null` in an optional argument position is never "argument omitted". Omitting the argument selects
+> the function's default; writing `null` there is subject to the usual null-in/null-out rule.
+
+It is stated once, on `CypherFunctionHelper.isExplicitNull(args, position)`, and every function with an optional
+argument calls it rather than re-deciding. That is what the issue asked for: the mechanism, not a per-function
+answer.
+
+## Changes
+
+All eleven defaulting functions now propagate.
+
+| function | optional argument | before | after |
+|---|---|---|---|
+| `normalize(input, normalForm)` | normal form | NFC | `null` |
+| `isNormalized(input, normalForm)` | normal form | NFC | `null` |
+| `format(temporal, pattern)` | pattern | `toString()` | `null` |
+| `round(value, precision, mode)` | rounding mode | HALF_UP | `null` |
+| `vector_distance(a, b, metric)` | metric | EUCLIDEAN | `null` |
+| `date.truncate(unit, input, fields)` | adjustment map | map ignored | `null` |
+| `datetime.truncate(...)` | adjustment map | map ignored | `null` |
+| `localdatetime.truncate(...)` | adjustment map | map ignored | `null` |
+| `time.truncate(...)` | adjustment map | map ignored | `null` |
+| `localtime.truncate(...)` | adjustment map | map ignored | `null` |
+| `SubstringFunction` (see below) | length | to end of string | `null` |
+
+Omitting the argument still selects the default in every one of them; only the explicitly-written `null`
+changes.
+
+### Two of the five named functions were not where the issue pointed
+
+Both are dual-path cases, and finding them is the reason the audit was worth running:
+
+- **`substring()`**: Cypher resolves it to `CypherSubstringFunction`, which already propagated (#5193).
+  `com.arcadedb.function.text.SubstringFunction` is a *second* implementation of the same function that no factory
+  currently registers, and it still defaulted. It is aligned here so the divergence does not surface the day it is
+  wired up, and it is covered by a direct unit test since no query can reach it.
+- **`vector_distance()`**: in Cypher this is a grammar rule whose metric is a keyword, so `vector_distance(a, b,
+  null)` does not parse - the same situation as Neo4j's `normalize()`. `VectorDistanceFunction`, the class the
+  issue points at, is reached from Cypher under the name **`vector.distance()`**, where the metric is an ordinary
+  expression and the explicit null is expressible. That is the path the fix and the test use.
+
+### Ordering is preserved in `round()`
+
+`RoundFunction` validates every argument before null propagation decides the answer, so that an unusable mode or
+an out-of-domain precision is reported even when the value is null (issue #5484). The new check sits with the
+other null propagation, after validation, so `round(null, 2, 'SIDEWAYS')` still raises rather than answering
+`null`.
+
+### `substring()` diverges from Neo4j, deliberately
+
+Neo4j raises on `substring(s, start, null)`. ArcadeDB's `substring()` already answers `null` for a null `start`,
+so raising only for a null `length` would make one function disagree with itself. It propagates for both - which
+is also what #5193 already chose for the Cypher-facing implementation. Whether ArcadeDB should instead follow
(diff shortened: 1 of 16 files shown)


robfrank:
Thanks - all three observations checked. One applied, two noted with a correction to each.

**1. Import placement - applied** (28f6464a5), with one correction: the misplaced import was *not* introduced by this PR. The parent commit's diff for those five files is `5 files changed, 25 insertions(+)` and touches no import lines - `CypherFunctionHelper` was already imported there, which is why the `isExplicitNull` calls compiled without an import change. The misplacement is pre-existing. Tidied anyway since the files are being touched.

**2. Redundant post-check guards - no change, but the redundancy is narrower than described.** After `isExplicitNull(args, 2)` returns false, both surviving guards are still load-bearing:

- `args.length >= 3` still distinguishes *omitted* from *present*, which is the whole distinction this PR turns on.
- `instanceof Map` still rejects a present, non-null, non-map value - `date.truncate('year', d, 'nonsense')` reaches it.

What `isExplicitNull` removed is only the case of `instanceof` doubling as an accidental null filter, which was the bug. So nothing here is dead code.

**3. Truncate family 500-on-client-error - agreed, and it is worse than the null-unit NPE alone.** There are two distinct paths, both recorded as follow-up 1 in the tracking doc:

- `args[0].toString()` on a null unit gives a raw `NullPointerException`.
- An *unknown* unit reaches `TemporalUtil.truncateDate`, which throws `IllegalArgumentException: Unknown truncation unit: fortnight` rather than a `CommandSemanticException`.

The second is why `aBadTruncateUnitIsStillReportedWhenTheAdjustmentMapIsNull` asserts only `hasMessageContaining("fortnight")` and deliberately not an exception class - asserting `CommandExecutionException` there would have encoded the wrong class as correct and made the eventual 400 fix look like a regression. Worth a fast-follow, and out of scope here since it is an error-classification defect rather than a null-semantics one.

robfrank:
Cycle 2 addressed in aa8340426. Both code points applied, with a correction on the third.

**2. Weak assertions - applied, and this was the right catch.** The four truncate default-path assertions checked only `.isNotNull()`, which the default path could satisfy while doing entirely the wrong thing - the exact failure mode the second assertion exists to rule out. They now assert the concrete value, matching what `dateTruncate...` already did:

- `datetime.truncate('year', ...)` -> `1984-01-01T00:00Z`
- `localdatetime.truncate('year', ...)` -> `1984-01-01T00:00`
- `time.truncate('hour', ...)` -> `12:00Z`
- `localtime.truncate('hour', ...)` -> `12:00`

**3. `@SuppressWarnings("unchecked")` - applied, and it is broader than reported.** It sits on `getMinArgs()` in **all five** truncate files, not only `DateTruncateFunction` and `DateTimeTruncateFunction`. Moved to `execute()` in all five, where the `(Map<String, Object>) args[2]` cast actually lives.

**1. Changelog - agreed on substance, but there is nowhere to put it.** The repository has no `CHANGELOG` file. You are right that this is a user-visible change for 11 functions and that `format(x, null)` returning `null` instead of the ISO string is the one most likely to surprise, so I have recorded it in the tracking doc under "PR and review history" for whoever cuts the release. Happy to add it to a specific file if release notes live somewhere I have not found.

On follow-up 5 (two implementations of `substring()`, only one registered) - agreed it deserves its own issue rather than a doc bullet. Leaving that for @robfrank to file so it is not created under the wrong milestone.

Verification after these changes: 234 tests green across `CypherOptionalArgumentNullIssue5629Test`, `OpenCypherTemporalFunctionsComprehensiveTest`, `CypherNumericFunctionArgumentIssue5484Test` and `TextStatelessFunctionsTest`. CI on the previous head was fully green (`build-and-package`, CodeQL all languages, Codacy 0 new issues, Meterian).

robfrank:
Cycle 3 addressed in d28f7590f.

**1. `parseRoundingMode` Javadoc contradiction - correct catch, fixed.** You are right: `args.length == 3 ? parseRoundingMode(args[2]) : HALF_UP` ran *before* the propagation check, so an explicit null was parsed to HALF_UP and the result then discarded. The behaviour was harmless but the Javadoc's claim that an explicit null "never reaches this method" was simply false.

Fixed by short-circuiting rather than by softening the comment, so the contract becomes true instead of merely accurate:

```java
final RoundingMode mode =
    args.length == 3 && !CypherFunctionHelper.isExplicitNull(args, 2) ? parseRoundingMode(args[2]) : RoundingMode.HALF_UP;
```

The ordering guarantee is unaffected - a mode that is present and non-null is still parsed before propagation, so `round(null, 2, 'SIDEWAYS')` keeps reporting the unusable mode. `anUnusableModeIsStillReportedWhenAnotherArgumentIsNull` still passes. The Javadoc now also records that the `null` branch serves callers passing the omitted-argument sentinel directly, which is what keeps it from being dead.

**2. Truncate computes before short-circuiting - declining, and not only on cost grounds.** I tried the suggested `&& !isExplicitNull` on the earlier line to be sure, and it **breaks the ordering guarantee**:

```
[ERROR] CypherOptionalArgumentNullIssue5629Test.aBadTruncateUnitIsStillReportedWhenTheAdjustmentMapIsNull:191
```

Skipping `TemporalUtil.truncate*` means `date.truncate('fortnight', d, null)` answers `null` instead of reporting the bad unit - exactly the masking this PR was careful to avoid. The wasted computation is the price of validating first, and it is one truncation on a path that is about to return `null` anyway. Reverted; the probe is not in the branch.

This is the difference between the two cases: in `round()` the parse is a *pure* function of the argument being discarded, so skipping it loses nothing. In `truncate()` the computation is what performs the validation, so skipping it loses the error.

**3. Follow-up #1 priority - agreed.** Noted for a fast-follow; it is the raw `NullPointerException` on a null unit plus the `IllegalArgumentException` on an unknown one, both HTTP 500 for a caller mistake.

robfrank:
Cycle 4 addressed in f1b2f25f3. This is the last automated cycle - the branch is handed back to @robfrank from here.

**1. Double evaluation of `isExplicitNull(args, 2)` - applied.** Hoisted into `final boolean nullMode`, so the mode-parsing branch and the propagation guard read off one value and cannot drift apart. Also merged the two adjacent comments, which had started to repeat each other after the last cycle.

**2. Truncate accepts a present, non-null, non-Map third argument - agreed it is a defect, recorded as follow-up 6.** `date.truncate('year', d, 42)` answers the truncated date and drops the `42`. It is the same shape as what this PR repaired for `null`, one step further out, and it should be a type error.

**On the suggested test: deliberately not adding one.** A test asserting that `date.truncate('year', d, 42)` returns the truncated value would encode the defect as the expected behaviour, so the eventual fix would present as a regression and the next person would have to decide whether to trust the test or the type system. That is the same trap avoided earlier by asserting only the *message* for the unknown-unit case rather than its (wrong) exception class. The behaviour is documented in the tracking doc under follow-up 6 instead, where it reads as "known defect" rather than "pinned contract". Happy to add a characterization test if you would rather have regression detection than a clean signal - it is a real trade-off, just one I would not make silently.

**3.** Agreed on the deferred items.

---

### Cycle summary

| cycle | head | verdict | applied |
|---|---|---|---|
| 1 | `5ae1ba3b7` | LGTM | Misplaced import (pre-existing) moved into the grouped block |
| 2 | `28f6464a5` | LGTM | Four `.isNotNull()` default-path assertions strengthened to concrete values; `@SuppressWarnings` moved to `execute()` in all five files |
| 3 | `aa8340426` | LGTM | `parseRoundingMode` no longer called for a mode about to propagate, making its Javadoc true |
| 4 | `d28f7590f` | LGTM | Duplicate `isExplicitNull` evaluation hoisted; follow-up 6 recorded |

CI green throughout: `build-and-package` SUCCESS on `5ae1ba3b7`, `28f6464a5` and `d28f7590f`, CodeQL all six languages, Codacy 0 new issues, Meterian. Local verification after each cycle: 234 tests across the four affected classes, 3060 across the full Cypher/function surface.

Two items left for the maintainer:

- **Release notes.** The repository has no `CHANGELOG`, so the user-visible surface (most notably `format(x, null)` now answering `null` instead of the ISO string) is recorded only in the PR body and the tracking doc.
- **Follow-ups 1-6** in `docs/5629-cypher-explicit-null-optional-argument.md`, of which #5 (two `substring()` implementations, one unregistered) and #1 (raw `NullPointerException` / `IllegalArgumentException` from the truncate family, both HTTP 500 for caller mistakes) are the ones worth their own issues.

Issue #5702, opened 2026-08-01
ha-integration-tests fails on main ~75% of the time, with a different test each run


lvca:
## What happens

`ha-integration-tests` is failing on `main` most of the time, and it fails on a **different** test almost every run.
The last 8 runs of the `Java CI - test` workflow on `main`:

| commit | result | run |
|---|---|---|
| `83448ea6` | success | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30693081526/job/91351801874) |
| `7469a248` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30691056907/job/91346391204) |
| `9337abf1` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30691016828/job/91346264716) |
| `50130d37` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30690966979/job/91346125046) |
| `e64160a3` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30687024879/job/91336711260) |
| `64aaa29f` | success | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30686831253/job/91334942821) |
| `4bd617e1` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30686802579/job/91334741538) |
| `af0dc719` | **failure** | [job](https://github.com/ArcadeData/arcadedb/actions/runs/30685641431/job/91331398817) |

6 of 8 failed. These are unrelated commits (a logger config change, an SQL arithmetic fix, a Duration fix, ...), so
this is the suite, not the changes.

## The failures are non-deterministic, not a broken test

The clearest evidence: the same commit was run twice on #5685 and failed both times on **disjoint** sets of tests.

| run | failing tests |
|---|---|
| 1st ([53m33s](https://github.com/ArcadeData/arcadedb/actions/runs/30687013122/job/91335912631)) | `RaftMigrationCompactionRaceIT.migrationStyleBulkInsertAcrossManyTypesNeverEmitsCannotFindIndexes` |
[...]

PR #5705, opened 2026-08-01
fix(ci) #5701: coverage-report waits for the HA suite, and refuses to publish a partial merge

### LINKED ISSUE #5701: codecov/project fails on unrelated PRs: coverage-report does not wait for ha-integration-tests
## What happens

`codecov/project` reports a whole-project coverage drop on PRs that cannot have caused one, so the check is red on
work that never touched the affected code. On #5685 (a 29-line change confined to
`engine/src/main/java/com/arcadedb/index/hash/`) Codecov reported:

```
Files          1770     1770
Lines        150057   150086      +29
Hits         101236   100126    -1110     <-- 1,110 covered lines disappear
```

A 29-line change cannot delete 1,110 hits. Every one of the 34 files that lost coverage is in `ha-raft/`:

| Hits lost | File | base -> head |
|---|---|---|
| -295 | `RaftReplicatedDatabase.java` | 68.16% -> 29.19% |
| -229 | `RaftHAServer.java` | 57.71% -> 38.07% |
| -116 | `ArcadeStateMachine.java` | 66.66% -> 54.35% |
| -97 | `PostVerifyDatabaseHandler.java` | 60.58% -> 3.52% |

## Cause

In `.github/workflows/mvn-test.yml`, `coverage-report` is wired to consume HA coverage but never waits for it:

```yaml
coverage-report:
  needs: [ unit-tests, integration-tests, slow-unit-tests ]   # ha-integration-tests is absent
  ...
  - name: Download HA integration test coverage reports        # line ~757
    uses: actions/download-artifact@...
    continue-on-error: true                                    # failure is swallowed
    with:
      name: ha-integration-coverage-reports
```

`ha-integration-tests` does upload that artifact (line ~340, `if: success() || failure()`), so the intent is clearly
that HA coverage be included. But because the job is not in `needs`, `coverage-report` starts as soon as the three
listed jobs finish, while `ha-integration-tests` runs for ~55 minutes - by far the longest job in the workflow. The
artifact does not exist yet, `continue-on-error: true` hides that, and the merged report silently omits the whole
`ha-raft` module.

So the project total is computed against a base that included HA coverage and a head that does not. The gap is
roughly -0.76pp, which is enough to trip the threshold.

## Why it matters

This is not specific to one PR - it is structural, and it fires on every PR whose other jobs finish before the HA
suite. The practical effect is that `codecov/project` is red for reasons unrelated to the diff, which trains
reviewers to ignore it. A coverage gate that is always red is worse than no gate: a real coverage regression would
look exactly like the noise.

Note `codecov/patch` is unaffected and remains meaningful (it scored the #5685 diff at 87.75% and passed), as do
`Codacy Diff Coverage` and `Codacy Coverage Variation`.

## Suggested fix

Add the missing dependency:

```yaml
coverage-report:
  needs: [ unit-tests, integration-tests, slow-unit-tests, ha-integration-tests ]
```

`coverage-report` already has `if: success() || failure()`, so adding the dependency does not make a failing HA suite
block the coverage report - it only makes it wait, which is what the existing download step assumes.

Two things worth deciding alongside:

1. **`continue-on-error: true` on the download hides exactly this class of bug.** Consider failing the job when an
   expected coverage artifact is missing, or at minimum logging which artifacts were skipped, so an incomplete
   upload is visible rather than silently reported as a coverage drop.
2. `ha-integration-tests` is currently failing on `main` most of the time (filed separately). While that is true,
   its coverage artifact still uploads (`if: success() || failure()`), so this fix works regardless - but the two
   interact, and fixing only this one will surface partial HA coverage from partially-failed runs.

Found while diagnosing a spurious `codecov/project` failure on #5685 (issue #5677).

lvca:
Fixes #5701.

## The bug

`coverage-report` downloads `ha-integration-coverage-reports` but its `needs` listed only
`[unit-tests, integration-tests, slow-unit-tests]`. `actions/download-artifact` reads the artifact
store of the current run and has no happens-before relationship with the uploader - that ordering
exists only through `needs`. `ha-integration-tests` takes ~55 minutes, by far the longest job in
the workflow, so `coverage-report` usually started well before it finished.

The download then failed, `continue-on-error: true` swallowed the failure, and the merged report
went to Codecov with the whole `ha-raft` module absent. Codecov does not read an absent module as
"unknown" - it reads it as uncovered, which is how a 29-line change to
`engine/src/main/java/com/arcadedb/index/hash/` produced `Hits 101236 -> 100126 (-1110)` on #5685.

## Confirmed on the real runs

| run | coverage-report started | ha-integration-tests finished | HA artifact |
|---|---|---|---|
| 30691056907 | 08:53:51 | 09:08:18 | `Artifact not found for name: ha-integration-coverage-reports` |
| 30691165476 | 08:59:57 | 09:12:02 | missing |
| 30693081526 | 10:08:14 | 10:08:03 | present (finished 11s earlier) |

So it was a race, not a constant. That is the part that made it hard to see: identical config
produced a complete report on a slow run and an incomplete one on a fast run, and the incomplete
runs also became the base every later pull request was compared against - one bad run yields a
phantom drop and then a phantom recovery on whoever comes next.

## What changed

**1. The ordering.** `ha-integration-tests` added to `coverage-report.needs`. The job condition
becomes `!cancelled()` rather than `success() || failure()`, so a suite that reported failing tests
still publishes what JaCoCo wrote (the suites run with `--fail-never`), while a cancelled run - the
one case where an artifact is genuinely absent rather than merely late - still skips.

**2. No more silent partial merges.** The issue notes that `continue-on-error: true` hides exactly
this class of bug, and suggests failing or at least logging. This goes with failing.
`.github/scripts/collect-coverage-reports.sh` replaces the bare
`find . -path **/jacoco*.xml -printf '%p,'`: it takes the suites it expects as `<suite>=<dir>`
pairs, verifies each contributed at least one report, writes a per-suite table to the step summary,
and exits non-zero naming the missing suites. The Codacy and Codecov steps lose their
`if: success() || failure()` so they are skipped when it fails. A partial upload is worse than no
upload, and a red `coverage-report` naming the missing suite is a far better signal than a green
one reporting a fabricated regression.

**3. The generalisation - suggested in place of just adding the `needs` entry.** Adding one entry
fixes today's instance and leaves the class open: nothing about a `download-artifact` step tells
you which job must precede it, so the next one is equally invisible.
`.github/scripts/check-workflow-artifact-deps.py` makes it an enforced invariant - for every
download step it resolves the producing job and asserts it is in the consumer's transitive `needs`
closure, across every workflow. It handles transitive `needs`, matrix-interpolated artifact names
matched as globs against `pattern:` consumers, selector-less downloads (which depend on every
producer), and cross-run downloads (which this run's ordering cannot constrain). Run against the
unfixed workflow it reports this bug and nothing else; across the other 17 workflows it is clean.
Wired into the `setup` job.

## Verification

`.github/scripts/tests/test-ci-scripts.sh` (also wired into `setup`) - 15 checks, both directions,
against fixtures rather than the live workflows:

```
check-workflow-artifact-deps.py
  ok   - rejects a download whose producer is not in needs
  ok   - accepts the same workflow once needs lists the producer
  ok   - accepts a producer reached transitively through needs
  ok   - resolves a matrix-named upload against a pattern download
  ok   - rejects a pattern download that does not wait for the matrix
  ok   - rejects a matrix-named download that does not wait for the matrix
  ok   - does not guess at a fully interpolated selector
  ok   - rejects an unselective download that misses a producer
  ok   - ignores a download from another workflow run
  ok   - the repository's own workflows are ordered

collect-coverage-reports.sh
  ok   - collects a report from every suite
  ok   - refuses to publish when a suite's artifact is missing
  ok   - refuses to publish when a suite produced no report
  ok   - the file list is comma-separated with no empty entries
  ok   - rejects an argument that is not <suite>=<dir>

15 checks passed
```

Mutation-checked: reverting only the `needs` line and re-running turns
`the repository's own workflows are ordered` red with the exact #5701 diagnostic, and nothing else
changes. The two `collect-coverage-reports.sh` refusal cases are the #5701 shape directly - the HA
artifact never downloaded, and the HA job ran but wrote no report.

## Trade-off worth stating

`coverage-report` now waits for the slowest job, so the Codecov and Codacy checks land roughly 10
minutes later than they do today. That is the cost of the signal being true; there is no ordering
that produces a complete merge sooner.

## On #5702

Independent of it. `ha-integration-tests` runs with `--fail-never` and uploads with
`if: success() || failure()`, so its coverage artifact exists whether or not tests pass. The issue
notes this fix will surface partial HA coverage from partially-failed runs - it will, and that is
the correct behaviour: coverage genuinely lost because tests did not run is a real signal, unlike a
module vanishing entirely. Only a cancelled HA job produces no artifact, and that now fails loudly
by name instead of silently.

## Behavior change worth knowing about

`coverage-report` can now go red where it previously went green. Two paths:

1. A suite that ran but never uploaded - a runner death or a job timeout, *not* a test failure,
   which `--fail-never` already covers - leaves its artifact missing and the collector fails the
   job by name. Given #5702, the HA suite is the likely one.
2. Any suite reporting `skipped` skips `coverage-report` entirely rather than publishing three of
   four.

Both are deliberate. The alternative to (1) is publishing a merge without `ha-raft`, which is
#5701 itself; a red check naming the suite is the honest version of the same event. If it proves
noisy, the remedy is fixing #5702 rather than re-enabling silent partial merges.

Neither can block a merge today: `main` has no branch protection (`GET /branches/main/protection`
returns 404), no rulesets, and Mergify's only rule targets Dependabot. No status check is required.

### DIFF
--- a/.github/scripts/check-workflow-artifact-deps.py
+++ b/.github/scripts/check-workflow-artifact-deps.py
@@ -24,6 +24,14 @@
 # `download-artifact` step it resolves which job uploads that artifact and asserts the producer is
 # in the consumer's transitive `needs` closure.
 #
+# Known boundary: it reads a job's `steps`, so a job that delegates to a reusable workflow
+# (`jobs.<id>.uses:`) is opaque - uploads and downloads inside the called workflow are neither
+# resolved nor checked. Because this check gates every other job, being wrong there would be
+# expensive, so a workflow containing such a job stops reporting "no job uploads this": the
+# producer may well be one of the steps it cannot see. Ordering violations it *can* see are still
+# reported. Following `uses:` into local workflows would lift the restriction; nothing in this
+# repository needs it yet.
+#
 # Usage:
 #   check-workflow-artifact-deps.py                 # check .github/workflows
 #   check-workflow-artifact-deps.py FILE|DIR ...    # check the given workflows
@@ -195,6 +203,10 @@ def check_workflow(path):
     produced = producers_of(jobs)
     violations = []
 
+    # A job that delegates to a reusable workflow has no steps to read, so an artifact this file
+    # never mentions may still be uploaded inside it. See the boundary note in the header.
+    opaque = any("uses" in job for job in jobs.values())
+
     for job_name, job in jobs.items():
         closure = None
         for step in steps_of(job):
@@ -231,10 +243,11 @@ def check_workflow(path):
 
             uploaders = matching_producers(produced, selector_names, selector_glob, is_pattern)
             if not uploaders:
-                violations.append(
-                    "%s: job '%s', step '%s' downloads '%s', which no job in this workflow uploads"
-                    % (path, job_name, step_name, selector)
-                )
+                if not opaque:
+                    violations.append(
+                        "%s: job '%s', step '%s' downloads '%s', which no job in this workflow "
+                        "uploads" % (path, job_name, step_name, selector)
+                    )
                 continue
 
             if closure is None:
--- a/.github/scripts/check-workflow-artifact-deps.py
+++ b/.github/scripts/check-workflow-artifact-deps.py
@@ -50,7 +50,8 @@ EXPRESSION = re.compile(r"\$\{\{.*?\}\}")
 
 
 def as_glob(value):
-    """Collapse a run-time expression into the glob of names it can produce.
+    """
+    Collapse a run-time expression into the glob of names it can produce.
 
     A matrix upload names its artifact `bolt-matrix-java-${{ matrix.version }}`; statically that is
     `bolt-matrix-java-*`. Keeping it as a glob is what lets the check still resolve the consumer
@@ -88,7 +89,7 @@ def needs_closure(jobs, name, seen=None):
 
 
 def producers_of(jobs):
-    """artifact glob -> the jobs that upload a name matching it."""
+    """Artifact glob -> the jobs that upload a name matching it."""
     produced = {}
     for job_name, job in jobs.items():
         for step in steps_of(job):
@@ -101,7 +102,8 @@ def producers_of(jobs):
 
 
 def matching_producers(produced, name, pattern):
-    """Resolve a download step's selector to the set of jobs it reads from.
+    """
+    Resolve a download step's selector to the set of jobs it reads from.
 
     `name` selects one artifact, `pattern` selects every artifact matching a glob, and neither
     means "download everything in the run" - which depends on every producing job. Both sides can
--- a/.github/scripts/check-workflow-artifact-deps.py
+++ b/.github/scripts/check-workflow-artifact-deps.py
@@ -1,0 +1,234 @@
+#!/usr/bin/env python3
+#
+# Fails when a workflow job downloads an artifact it does not wait for.
+#
+# `actions/download-artifact` reads the artifact store of the *current run*. It has no
+# happens-before relationship with the job that uploads: if the producer has not finished by the
+# time the consumer runs, the artifact simply is not there yet. GitHub expresses that ordering
+# only through `needs`, and nothing in the actions themselves checks it - a consumer with the
+# wrong `needs` is valid YAML that runs green and silently reads an incomplete input.
+#
+# That is how #5701 happened. `coverage-report` downloaded `ha-integration-coverage-reports` but
+# listed only `[unit-tests, integration-tests, slow-unit-tests]` in `needs`, so it started ~14
+# minutes before `ha-integration-tests` finished. The download failed, `continue-on-error: true`
+# swallowed it, and the merged report went to Codecov without the ha-raft module - which Codecov
+# read as 1,110 covered lines vanishing, and reported as a coverage regression on a PR that had
+# touched 29 lines somewhere else entirely. Every symptom pointed at the diff; the cause was a
+# missing entry in a `needs` list.
+#
+# Worse, it was a race, not a constant: when the other three jobs happened to finish late the
+# artifact was there and the report was complete. So the check flipped between correct and wrong
+# on identical config, which is exactly the shape of bug that survives code review.
+#
+# This script makes the ordering an enforced invariant instead of a convention: for every
+# `download-artifact` step it resolves which job uploads that artifact and asserts the producer is
+# in the consumer's transitive `needs` closure.
+#
+# Usage:
+#   check-workflow-artifact-deps.py                 # check .github/workflows
+#   check-workflow-artifact-deps.py FILE|DIR ...    # check the given workflows
+#
+# @author Luca Garulli (l.garulli@arcadedata.com)
+
+import fnmatch
+import re
+import sys
+from pathlib import Path
+
+try:
+    import yaml
+except ImportError:  # pragma: no cover - the workflow installs it explicitly
+    sys.exit("check-workflow-artifact-deps: PyYAML is required (pip install pyyaml)")
+
+UPLOAD_ACTION = "actions/upload-artifact"
+DOWNLOAD_ACTION = "actions/download-artifact"
+
+# upload-artifact's documented default when `name` is omitted.
+DEFAULT_ARTIFACT_NAME = "artifact"
+
+EXPRESSION = re.compile(r"\$\{\{.*?\}\}")
+
+
+def as_glob(value):
+    """Collapse a run-time expression into the glob of names it can produce.
+
+    A matrix upload names its artifact `bolt-matrix-java-${{ matrix.version }}`; statically that is
+    `bolt-matrix-java-*`. Keeping it as a glob is what lets the check still resolve the consumer
+    that picks the whole family up with `pattern: bolt-matrix-*`, instead of reporting a family of
+    artifacts nobody uploads.
+    """
+    return EXPRESSION.sub("*", value)
+
+
+def action_of(step):
+    uses = step.get("uses")
+    return uses.split("@", 1)[0].strip() if isinstance(uses, str) else ""
+
+
+def steps_of(job):
+    steps = job.get("steps")
+    return [s for s in steps if isinstance(s, dict)] if isinstance(steps, list) else []
+
+
+def needs_of(job):
+    needs = job.get("needs")
+    if isinstance(needs, str):
+        return [needs]
+    return [n for n in needs if isinstance(n, str)] if isinstance(needs, list) else []
+
+
+def needs_closure(jobs, name, seen=None):
+    """Every job that must finish before `name` starts, directly or transitively."""
+    seen = set() if seen is None else seen
+    for parent in needs_of(jobs.get(name) or {}):
+        if parent not in seen:
+            seen.add(parent)
+            needs_closure(jobs, parent, seen)
+    return seen
+
+
+def producers_of(jobs):
+    """artifact glob -> the jobs that upload a name matching it."""
+    produced = {}
+    for job_name, job in jobs.items():
+        for step in steps_of(job):
+            if action_of(step) != UPLOAD_ACTION:
+                continue
+            with_ = step.get("with") or {}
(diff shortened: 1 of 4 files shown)


lvca:
Thanks - all six addressed in d02d6b4d1. Taking them in order:

**1. Bidirectional glob over-match.** Valid, and fixed properly rather than documented. The
suggestion was a note in the docstring, but a false positive is expensive precisely because of your
observation #3 - this check gates every other job - so the right move was to remove the
approximation. `matching_producers` no longer reduces a matrix name to a glob when it can resolve
it: `build-${{ matrix.os }}` now expands to exactly `{build-ubuntu, build-macos}` from the base
matrix *and* from `include:` entries (the shape most of this repo's matrices use, e.g.
`test-python-bindings.yml`). Your exact example is now a fixture, in both directions:

```
ok - does not demand a matrix job whose glob merely covers another artifact   # name: build-logs
ok - still demands the matrix job when the name is one it produces            # name: build-macos
ok - resolves a value contributed only by a matrix include
```

The glob survives only as the fallback for a matrix that cannot be read statically -
`native-image.yml`'s `matrix: ${{ fromJSON(needs.setup.outputs.matrix) }}` - where over-matching is
the safe direction. That case is a fixture too. Independent matrix keys still expand as a cartesian
product, which over-approximates a correlated `include`-only matrix; that errs the same safe way
and is noted in the code.

**2. `run-id` skip is name-agnostic.** Valid, fixed. A `run-id` containing `github.run_id` names
the current run, so the step is an ordinary consumer and is checked. Fixture added.

**3. Hard gate on `setup`.** Acknowledged, and it is the intent. The mitigation is #1 - the
false-positive rate is the thing that makes this coupling risky, so that is where the work went.
The redundant re-run inside `test-ci-scripts.sh` stays.

**4. `!cancelled()` when an upstream fails.** Good catch, and on reflection the behaviour you
described is not what I want. A *skipped* suite is categorically different from a suite that ran
and uploaded nothing: it produced no report because `build-and-package` failed and it never
started, so a second red check adds no information. The job condition is now

```yaml
if: >-
  ${{ !cancelled()
  && needs.unit-tests.result != 'skipped'
  && ... }}
```

which keeps the collector's failure meaning only what it says - a suite that should have uploaded
and did not.

**5. Unpinned pyyaml.** Valid, pinned to `pyyaml==6.0.3`, with the rationale in a comment next to
the SHA-pinned actions it sits beside.

**6.** Agreed, no change.

Suite is at 20 checks, all green, and `setup` has passed on CI for every push so far - so the
checker has now run against all 18 real workflows three times.

lvca:
Both rounds addressed in d7572ef71. The one point both reviews raised independently was the
reusable-workflow blind spot, so that got the real work.

**Reusable workflows (R3.1, R4.2).** Both asked for a one-line note. Documented, and also made
safe: a job with `jobs.<id>.uses:` has no steps, so an upload inside it is invisible, and the
concrete harm you both named is the spurious `which no job in this workflow uploads` hard failure.
A workflow containing such a job now suppresses *that* violation only - the producer may be a step
the check cannot see - while ordering violations it can still see are still reported. Two fixtures
pin both halves:

```
ok - does not report an artifact a reusable workflow may upload
ok - still reports a visible violation alongside a reusable workflow
```

**PyYAML step coupling (R3.2).** Valid - the self-test invoked the checker, which needed a
`pip install` buried in the preceding step's script. PyYAML now has its own named step, so the
dependency is visible rather than positional.

**Comment reflow (R3.4).** Fixed.

**Warning-only rollout (R4.1).** Declining, with reasoning. The original issue's complaint is that
a check which is red for unrelated reasons trains reviewers to ignore it; a check that only warns
is that same failure mode by construction. The concrete misfire you each named is the reusable
workflow case, and that is now suppressed rather than warned about, which is the targeted version
of the same de-risking. The matrix resolution in d02d6b4d1 shrank the other surface.

**All-or-nothing skip / hand-synced `if:` (R3.3, R4.4).** Agreed, and agreed not worth changing
now - both are consistent with the collector's all-or-nothing contract.

**HA infra failure reds coverage-report (R4.3).** Correctly identified as the one real behavior
question, and it needs @lvca's call rather than mine. A job that dies before its upload step -
runner death or a timeout, not a test failure, which `--fail-never` already covers - leaves the
artifact missing and the collector fails. Given #5702, that may be more than rare. I have kept it,
because the only alternative is publishing the merge without ha-raft, which is #5701 itself. If it
proves noisy the remedy is #5702, not re-enabling silent partial merges - but flagging it plainly
as a deliberate tradeoff rather than an oversight.

Suite is at 22 checks, all green.

lvca:
Addressed in 7ef828457.

**1. Is a skipped `coverage-report` going to block PRs?** Checked rather than assumed - it cannot.
`main` has no branch protection at all (`GET /branches/main/protection` -> 404 "Branch not
protected"), there are no repository rulesets, and the only Mergify rule is
`author=dependabot[bot]` on approval. No status check is required, so neither a skipped nor a red
`coverage-report` gates anything. Good question to have asked though - it would have been the one
genuinely disruptive outcome.

**2. `opaque` suppression is workflow-wide.** Right, and your specific worry is the real one: a
typo'd artifact name in a file that also calls a reusable workflow looks *identical* to an artifact
that workflow uploads, so it would have gone unreported. The benefit of the doubt still goes to the
workflow - this check gates every other job - but the finding is no longer silent:

```
check-workflow-artifact-deps: x.yml: 1 missing-producer finding(s) not reported, because a job in
this workflow delegates to a reusable workflow whose steps cannot be read:
      x.yml: job 'consume', step '...' downloads 'name: typo-or-from-called', which no job in
      this workflow uploads
```

Informational, exit 0. Silently discarding a finding is how #5701 stayed invisible in the first
place, so suppressing one without saying so would have repeated the mistake this PR is about.
I stopped short of following `uses:` into local workflows: it is the proper closure, but the repo
has no reusable workflows today, so it would be unexercised machinery on the critical path of every
build - and a false positive there is the expensive direction. The header documents the closure
path for whoever needs it.

**3. `find` in a process substitution escapes `pipefail`.** Valid, and worth fixing even though the
paths are safe: an unreadable directory read as an empty one, which is precisely the distinction
this script exists to draw. `find` now writes to a temp file and its status is checked. Fixture
added (skipped when running as root, which can read it anyway).

**4. Python + pyyaml on the critical path.** Acknowledged. Pinned, quiet, ~1s.

24 checks, all green.

lvca:
Addressed in 3e5e38f3a plus a PR-body edit.

**1. New hard-failure path, worth stating explicitly.** Agreed - added a "Behavior change worth
knowing about" section to the PR body covering both paths (a suite that ran but died before
uploading, and the all-or-nothing skip), why each is deliberate, and the fact that the alternative
to the first is publishing without ha-raft, which is #5701 itself. On the branch-protection
question: checked, and nothing is required - `GET /branches/main/protection` returns 404 "Branch not
protected", there are no rulesets, and Mergify's only rule is `author=dependabot[bot]`. So neither
path can block a merge today.

**2. All-or-nothing skip couples to *why* a suite is skipped.** Good catch, and the assumption was
load-bearing but unwritten. Now stated where the next editor will be standing:

```
# This reads any one skipped suite as "none of them ran", which holds because all four hang off
# build-and-package and so skip together. Give one of them a path filter or a condition of its
# own and that stops being true: the other three would still have coverage worth publishing,
# and this condition would drop it. Revisit here if a suite ever gains its own gate.
```

**3. `opaque` blast radius.** Agreed on all counts, including that it is unavoidable for
missing-producer without following `uses:`, and that nothing is suppressed in `mvn-test.yml` today.
Leaving it; the header documents the closure path.

**4. The linter gates the whole build.** Conscious of it - it is why the matrix resolution went in
rather than a docstring note about the `build-*` collision, and why the reusable-workflow case
downgrades to a note instead of failing. Conservative bias is the design, not an accident.

**Minor:** renamed the two comprehension variables in `main()`. And thanks for confirming the
`find` temp-file change - that one came from your last round.

24 checks, all green.

Issue #4432, opened 2026-05-31
PostCommandHandler: Cannot execute command


tomscreen:
I have just updated my 3-node HA cluster to the 26.6.1-snapshot build and have immediately had the following error when trying to delete some vertex records.

The command I am using, via the Leader, is:
DELETE FROM MyVertexClass
Expecting to delete the 4 records that were in there.
A couple of these records had an edge between them.

The failure message that appears in Studio is:

**Cannot execute command**
Invalid position 262138 (size=262136)

In the logs for the leader, this is the only entry that appears:

INFO  [PostCommandHandler] <arcadedb-1> Error on command execution (PostCommandHandler): Invalid position 262138 (size=262136)

I don't know if this is specific to an HA setup or not.

I appreciate your help working through these issues.

tomscreen:
I've just tried this in standalone mode as well and I get the same error.

tomscreen:
Some additional information:

There are indexes on 'AssetClassification' (the actual name of the class I'm using in this database) for some of the properties.

After the original 'Invalid position' error I tried rebuilding the index:
`REBUILD INDEX AssetClassification_byParent`

And then I retried the delete command:
`DELETE VERTEX FROM AssetClassification`

Now the error I get is: `java.nio.BufferUnderflowException`.

I've tried the command:
`CHECK DATABASE TYPE AssetClassification`

And have the following output:

```json
{
  "user": "root",
  "result": {
    "vertices": [],
    "edges": [],
    "records": [
      {
        "operation": "check database",
        "avgPageUsed": 13.291931,
        "orphanedExternalRecords": 0,
        "pageSize": 65536,
        "totalAllocatedDocuments": 0,
        "totalDeletedRecords": 0,
        "totalActiveRecords": 5,
        "totalActiveEdges": 0,
        "totalMaxOffset": 8711,
        "totalAllocatedVertices": 5,
        "corruptedRecords": [],
        "rebuiltIndexes": [],
        "invalidLinks": 0,
        "autoFix": 0,
        "totalMultiPageRecords": 0,
        "warnings": [],
        "totalErrors": 0,
        "totalSurrogateRecords": 0,
        "totalAllocatedRecords": 5,
        "totalActiveDocuments": 0,
        "deletedRecordsAfterFix": [],
        "orphanedExternalRecordsFixed": 0,
        "totalPages": 1,
        "totalPlaceholderRecords": 0,
        "totalChunks": 0,
[...]

PR #5707, opened 2026-08-01
fix(engine) #5680: an unreadable edge list retries the vertex delete instead of losing its edges

### LINKED ISSUE #5680: deleteVertex can delete a vertex without collecting its edges when a chunk is transiently unreadable (the #5670 window, vertex side)
### Summary

#5670 (PR #5678) closed the window where deleting an EDGE could leave its back-reference behind: a
chunk that is momentarily unreadable - a concurrent commit publishes its pages one at a time, so a
vertex page can expose a new edge-list head RID before that head's page is visible, and an emptied
chunk is relinked out of the chain under a walker's feet - was read as "nothing to remove here"
while the edge record was deleted anyway.

`GraphEngine.deleteVertex` still reads its own edge lists through the best-effort
`getEdgeHeadChunk`, and wraps the collection loop in `catch (Exception)`:

```java
try {
  outEdges = getEdgeHeadChunk(vertex, Vertex.DIRECTION.OUT);
  if (outEdges != null) { ... collect edges to delete ... }
} catch (Exception e) {
  // LINKED LIST COULD BE BROKEN
  LogManager.instance().log(this, Level.WARNING, "Error on deleting outgoing edges connected to vertex %s", e, ...);
}
...
vertex.getDatabase().getSchema().getBucketById(...).deleteRecord(vertex.getIdentity(), force);
```

So if the head chunk is unreadable at that instant, `outEdges` is null, **no edges are collected**,
and the vertex record is deleted regardless - leaving edges pointing at a vertex that no longer
exists (ghost edges), reported by `check database` as invalid links.

### Why it was left out of #5678

This is a deliberate, tested contract, not an oversight: `deleteVertex`'s javadoc states that edge
disconnection is best-effort, and `Issue4420TolerantDeleteTest` / `Issue4432CorruptVertexDeleteTest`
pin that a vertex whose chain is structurally broken stays deletable. Making the collection strict
would trade repairability for strictness, and that is a decision worth taking on its own merits
rather than as a side effect of an edge-delete fix.

Note that `deleteVertex` does get #5670's fix transitively where it matters most: every edge it does
collect is removed via `edge.delete()` -> `deleteEdge`, which is now strict, so a NEIGHBOUR's
back-reference is no longer left dangling.

### What a fix would look like

Distinguish the two cases the single `getEdgeHeadChunk` call currently conflates:

- `force == true` (explicit repair of a structurally broken record): keep today's tolerance.
- `force == false` (ordinary delete): use `getEdgeHeadChunkForWrite`, so a transiently unreadable
  head is a retryable conflict and the transaction re-reads a consistent view - the same line #5678
  drew for `deleteEdge`. The surrounding `catch (Exception)` would have to let `NeedRetryException`
  through.

The open question is what an ordinary `deleteVertex` should do when the chain is genuinely broken
(not transiently unreadable), since the two are indistinguishable at that point: today it deletes and
warns; strict handling would require `force` to get the vertex out. That is the trade-off to decide.

### Verification

A test in the shape of `Issue5670EdgeDeleteDanglingBackRefTest` - make the vertex's head chunk
unreadable, delete the vertex, and assert its edges do not outlive it - would pin whichever answer is
chosen.

lvca:
Fixes #5680.

## The bug is real, and it is wider than the issue describes

`GraphEngine.deleteVertex` collects the edges to remove by walking the vertex's own OUT and IN lists, then deletes
each collected edge, then deletes the vertex record. The collection read the head through the best-effort
`getEdgeHeadChunk` and wrapped the whole walk in `catch (Exception)`, so **any** part of the list that could not be
read at that instant was taken for "nothing to remove here" and the vertex record was deleted on top of that empty
or partial view. Reproduced deterministically on `main` - `A --E--> B` with `A`'s OUT head chunk removed:

```
DELETE VERTEX with own unreadable OUT list SUCCEEDED
vertex A still exists = false
EDGE record still exists = true      <-- edge outlived its OUT vertex
B in-degree           = 1            <-- B still points at it
```

Three distinct routes reach that outcome, not one. The issue named the first:

1. **Unreadable head chunk** - `getEdgeHeadChunk` answers `null`, nothing is collected.
2. **Unreadable hop mid-chain** - `Iterator.hasNext()` raises `RecordNotFoundException` out of the walk loop into
   the blanket catch. Everything *behind* the hole is dropped; everything already seen is kept, so the delete
   proceeds on a partial view that looks complete.
3. **A skipped stripe chain on a promoted super-node (#5156)** - the widest of the three and the only one with no
   exception behind it at all. `StripedEdgeList.edgeIterator` composes one iterator per stripe chain through
   `allChains(false)`, and a chain whose head cannot be read is *legitimately* skipped there: that is the
   documented best-effort READ contract. Used as a removal walk it silently lost a whole stripe's worth of edges.

None of these is a fact about the graph. They are facts about timing - the same window #5670 was about: a commit
publishes its pages one at a time and the reader takes no commit lock, so a vertex page can expose a head RID
before that head's own page is visible, and a chunk emptied by another transaction is relinked out from under a
walker. `deleteEdge` has answered that window with a retry since #5678; the vertex path never did.

A fourth, unrelated to concurrency, was found while writing the fix: the collection read the edge-list heads off
the caller's vertex handle rather than the instance the transaction holds (#5660). A handle obtained before an
append in the same transaction still names the previous head, so the newest edges were invisible to the walk - and
the vertex was deleted anyway. Same defect, no race required.

## The fix

Strict by default, `force` as the single escape hatch, and `force` made to work end to end.

- **`deleteVertex(v, force=false)`** - the collection reads the list the way a removal must: the head through the
  strict `getEdgeHeadChunkForWrite`, the walk through a new `EdgeLinkedList.edgeIteratorForRemoval()` (identical to
  `edgeIterator()` on the classic layout; on `StripedEdgeList` it composes `allChains(**true**)` so a stripe chain
  cannot be skipped), and the chain hops through a helper that maps an unreadable hop to a retryable
  `ConcurrentModificationException`. The `catch (Exception)` no longer swallows `NeedRetryException`.
- **`deleteVertex(v, force=true)`** - every one of those conflicts is absorbed and logged, *including the one raised
  while disconnecting a collected edge from the vertex at its OTHER end*. That last one is the gap recorded in the
  issue's second comment: since #5678 made the neighbour's removal strict, a broken **neighbour** blocked a forced
  delete exactly as it blocked an ordinary one - precisely what `force` exists to override.
- The heads are read from `getMostUpdatedVertex(vertex)`, closing the fourth route above.

**One line is deliberately *not* strict**, in either mode: draining the leftover chunk records at the end. By then
every collected edge has been disconnected from both endpoints and the vertex record is about to go, so a chunk that
cannot be read there costs orphaned chunk records - garbage `CHECK DATABASE` reclaims - never a surviving reference.
The rule the fix follows is *strict where a miss loses a reference, tolerant where a miss loses only garbage*, which
is finer than the blanket "strict when `force == false`" the issue proposed.

Also left tolerant, in both modes: a single entry whose **edge record** cannot be resolved. It is skipped, the walk
keeps every other entry, so the cost is that one already-dangling pointer rather than the whole remaining list -
the same way `EdgeIteratorFilter` and `CHECK DATABASE` treat such an entry. Escalating it would make every vertex
carrying one pre-existing dangling pointer undeletable, which is a strictly larger change than this window
justifies.

## Why strictness does not cost repairability (the trade-off the issue asked to decide)

The issue framed the open question as "what should an ordinary `deleteVertex` do when the chain is *genuinely*
broken", since a transient and a lost chunk are indistinguishable at that moment. Two things settled it:

- **`force` is not the operator's answer, and never was.** It is not reachable from SQL or the API: it is set
  internally by `LocalDatabase.deleteRecordNoLock` only when `bucket.isChunkChainBroken` confirms the *record's own*
  multi-page chain is broken. A broken *edge list* never sets it. So "keep tolerance under `force`" gives users
  nothing on its own - it only fixes the internal forced-delete path (which this PR does fix, see above).
- **`CHECK DATABASE ... FIX` already is the answer, and it is reachable.** It rebuilds an unloadable chain from the
  surviving edge records - head chunk, mid-chain, classic and striped - and is pinned by
  `GraphDatabaseCheckerChainRebuildTest`. It reads edge lists through the best-effort reader, so it is never blocked
  by the delete being blocked. After the repair the delete goes through normally.

So a genuinely broken graph is still cleanable without `force`; it just takes the repair command first, and the
delete stops silently manufacturing ghost edges in the meantime.

### Alternatives considered and rejected

- **Disambiguate structurally, the way `LocalBucket.isChunkChainBroken` does for the record chain.** It does not
  transfer. There, the `ConcurrentModificationException` comes from page-*version* validation, so a version-blind
  structural walk genuinely tells contention from breakage. Here the failure *is* "record not found", which reads
  identically for a chunk that is invisible-for-now and one that is lost; a re-probe returns the same answer.
- **Make the last retry attempt tolerant** (expose the attempt number on the transaction, go best-effort on the
  final one). Attractive - it keeps a broken graph deletable without any repair step - but it makes correctness
  depend on `arcadedb.txRetries`, gives an explicit `begin`/`commit` caller no retries at all (attempt 0 is the last
  one, so it is exactly today's behaviour), and quietly reintroduces the ghost-edge outcome under a new name.
- **Confirm by taking the file commit lock and re-reading.** This *is* a true disambiguation: with no commit in
  flight, an unreadable chunk is genuinely lost. Rejected as deadlock-prone - acquiring a file commit lock outside
  commit ordering, on the delete path - for a case the repair command already covers.

## Visible behaviour change

`vertex.delete()` / `DELETE VERTEX` can now raise a `ConcurrentModificationException` where it previously
"succeeded". That is a `NeedRetryException`, so `database.transaction(...)` and the server's auto-retry for
single-request commands absorb it; a client-managed explicit transaction spanning several requests sees it and
should retry, the same contract concurrent updates have always had. This is the vertex-side half of the change
#5678 already documented for the neighbour side. Noted in the release notes.

## Tests

`Issue5680VertexDeleteEdgeCollectionTest` - **8 tests; the 3 that pin the defect fail on current `main`, all 8 pass
with the fix.** The other 5 are the controls that keep the fix honest.

| Test | Shape | On `main` |
|---|---|---|
| `deletingAVertexWhoseOwnHeadChunkIsUnreadableReportsAConflictRatherThanLosingItsEdges` | route 1, deterministic | **fails** |
| `deletingAVertexWhoseChainIsBrokenMidWayReportsAConflictRatherThanLosingTheEdgesBehindTheHole` | route 2, deterministic | **fails** |
| `deletingAPromotedSuperNodeWhoseStripeChainIsUnreadableReportsAConflict` | route 3, deterministic | passes for the wrong reason (see below) |
| `aForcedDeleteIsNotBlockedByAnUnreadableNeighbourEdgeList` | the `force` gap from comment 2 | **fails** |
| `aForcedDeleteStillRemovesAVertexWhoseOwnEdgeListIsUnreadable` | control: `force` tolerance kept | passes |
| `aForcedDeleteOfAHealthyVertexStillDisconnectsItsNeighbours` | control: `force` is not "skip the work" | passes |
| `anOrdinaryDeleteOfAHealthyVertexRemovesEveryEdgeOnBothSides` | control: intact graph untouched | passes |
| `concurrentVertexDeletionAndEdgeAppendingLeaveNoEdgeBehind` | `@Tag("slow")`; the production shape | passes |

Each deterministic test asserts its preconditions before breaking anything (the degree is right, the chain really
spans several chunks, the hub really was promoted to the striped layout), so none can pass on an unrelated throw.

The super-node one is worth calling out because the obvious version of it is vacuous: an end-to-end
`DELETE VERTEX` on that fixture raises `ConcurrentModificationException` **with or without** the fix, because the
per-edge disconnection reaches the same broken chain through `chainsForNeighbour`, which has always been strict. So
it asserts the two walks over the same broken list side by side instead - the READ one still under-reports the
degree in silence (unchanged, deliberately), the REMOVAL one raises naming the broken chunk RID - and only then the
end-to-end delete. Verified by flipping `allChains(true)` back to `false`: the test fails.

`concurrentVertexDeletionAndEdgeAppendingLeaveNoEdgeBehind` asserts the invariant rather than reproducing the race:
every edge in the fixture points at a hub, so once every hub is deleted no edge record may survive. It also checks
that the appends really did land against hubs being deleted (otherwise the rounds only measured a sequential
delete) and that the fixture was non-empty before each round, so the invariants cannot hold vacuously.

## Verification

- `Issue5680VertexDeleteEdgeCollectionTest`: 8/8, three consecutive runs.
- Full `engine` suite (benchmarks excluded): see below.
  **10743 tests, 0 failures, 0 errors, 23 skipped** (21:49 min).
- `server` module suite (benchmarks excluded): **950 tests, 0 failures, 0 errors**.


lvca:
Thanks - addressed both documentary observations in a95aa6fa7. Notes on each:

**1. Residual silent-loss path under hard corruption.** Correct reading, and I dug into whether it should become strict rather than documented. It should not, and the reason is concrete: `LocalDatabase.deleteRecordNoLock` catches exactly that exception family (`SerializationException | NegativeArraySizeException | BufferUnderflowException | IndexOutOfBoundsException | IllegalArgumentException`) around the index cleanup and proceeds **without** setting `forceBrokenChainDelete` - it raises that flag only in the `ConcurrentModificationException` branch, after `isChunkChainBroken` confirms a broken chunk chain. So a vertex whose buffer is corrupt reaches `deleteVertex` with `force == false`, and failing there would make it undeletable again - the "records that cannot be deleted" complaint #4420 and #4432 fixed. `force` is not an escape hatch for this shape because nothing sets it for this shape.

So the boundary stays, but you are right that its cost is larger than the tolerated single dangling entry (everything behind the damage is dropped, and the vertex still goes). The catch now says so explicitly, and its WARNING names the case and points at `CHECK DATABASE ... FIX` rather than the previous generic "Error on deleting ... edges".

**2. Classic vs striped: where the CME originates.** Fair - the javadoc sentence was accurate for the classic layout and misleading as a general claim. Added a note that `StripedEdgeList.edgeIteratorForRemoval` resolves every stripe head eagerly at iterator construction, so an unloadable stripe raises there rather than on a hop, and that both land in the same `try` in `collectEdgesToDelete`.

**3. `deleteRemainingChunks`** - agreed, no action taken.

Documentation only, no behaviour change. Re-verified: `Issue5680VertexDeleteEdgeCollectionTest` 8/8, and the `com.arcadedb.graph` package plus `Issue4420TolerantDeleteTest` / `Issue4432CorruptVertexDeleteTest` - the two that pin the tolerance discussed in point 1 - 345 tests, 0 failures.

lvca:
Thanks - second pass addressed in 11f4805c5. Point by point:

**1. Explicit about the genuinely-lost chunk.** Agreed, and worth more than a footnote: the split between the two catch arms is easy to read as *transient vs permanent*, which is the wrong axis and would lead the next reader to draw the line in the wrong place. It keys on **what a miss costs**. A chunk that cannot be FOUND retries because the alternative - deleting the vertex on a short list - loses references, and that is applied to a genuinely lost chunk too, which no retry recovers: that case now fails the delete once the retries are spent rather than quietly completing it. A chunk that cannot be DECODED takes the other arm for the `#4420`/`#4432` reason, not because it is less permanent. Spelled out on `collectEdgesToDelete`.

**2. Self-loop test.** Added `deletingAVertexWithASelfLoopRemovesItCleanly`. You are right that it is the shape most likely to surprise a later change - it is the only one both walks yield, so it enters `edgesToDelete` twice and the second removal runs against a chain the first already cleaned. It asserts the precondition (the loop is reachable OUT *and* IN, so the duplicate really arises), that the vertex and the edge both go, and that `check database` is clean.

**3. Whole-list materialization on super-nodes.** Agreed and not changed. It is unchanged from the previous implementation, and the two-phase collect-then-delete exists precisely because the removals mutate the chain the walk is standing on - streaming it would mean walking a list while `deleteEdge` relinks and deletes chunks underneath, which is the class of problem this PR is fixing. Worth its own issue rather than a change smuggled into a correctness fix.

**4. Retry count in the concurrency test.** No change - as you note the 5-minute ceiling caps the tail, and the budget is deliberately far beyond what the contention needs so that any throwable reaching the counter is the bug rather than a tolerated retry.

Verified: `Issue5680VertexDeleteEdgeCollectionTest` 9/9, `com.arcadedb.graph` package 373 tests, 0 failures.

lvca:
Third pass addressed in cab5a0a8c - and it turned up something worth flagging separately.

**Closed catch (your one substantive point).** Agreed, and implemented: absorbing an NPE or `IllegalStateException` as "corrupted chunk, delete anyway" would reintroduce exactly the silent-edge-loss path this PR removes. The arm now names what it tolerates - the decode family `LocalDatabase` uses for the same purpose (`SerializationException`, `NegativeArraySizeException`, `BufferUnderflowException`, `IndexOutOfBoundsException`, `IllegalArgumentException`) plus `ClassCastException` and `SchemaException`, the two further shapes a corrupt *edge list* adds on top (a head RID naming a record that is not an edge segment; an edge bucket whose type is gone). Everything else propagates, and the comment says to add a shape with its reason rather than widen the catch.

**`getMostUpdatedVertex` redundant lookup.** Not changed. It is a `HashMap` get on a path that is about to do record I/O, and making it conditional is what turned the #5660 invariant into a coincidence in the first place - one method owns the rule.

---

**A finding, not a fix: `concurrentVertexDeletionAndEdgeAppendingLeaveNoEdgeBehind` was flaky (~1 in 3) and I chased it rather than retrying it away.**

It is **not** this fix - it reproduces identically with the fix reverted (3 failures in 6 runs on the parent commit). Diagnosed by dumping the survivor:

```
GHOST edge=#65:70 out=#43:490 outExists=true in=#1:1 inExists=false inIsHub=true
```

An edge an **appender** commits while its hub is being deleted can outlive that hub. The survivor has a live `out` vertex and an `in` naming the deleted hub, so it did not exist when the deleter collection walk ran - no amount of vertex-side strictness can prevent it. That is a separate pre-existing defect on the **append** path (`newEdge` racing `deleteVertex`), and I would rather report it than fold it into this PR.

So the test now asserts the invariant this fix owns - every edge that was in the hub list when the deleter started must go with it, i.e. the fixture edges - instead of "no LINK edge survives at all", which quietly folded in the append-side defect. The round-start precondition moves from a global `LINK` count to a per-hub degree so a leftover from an earlier round cannot make a healthy round look wrong, and `assertIntegrityClean()` is dropped from this one test with the reason inline (the non-racing fixtures still assert integrity).

Verified: 10 consecutive green runs of the test class, `com.arcadedb.graph` 373 tests / 0 failures.

lvca:
Fourth pass - all three points were flagged non-blocking, and I agree with that classification. One is now recorded in the code (07a769ca3, comment only); the other two I am deliberately not changing, with reasons.

**1. Breadth of the decode catch arm, `IllegalArgumentException` in particular.** Fair, and it is the right member to single out. It stays, and I verified why rather than asserting it: `Binary` raises it at `Binary.java:179` - `"Invalid position " + index + " (size=" + buffer.limit() + ")"` - for a content offset that decodes past the end of the buffer. That is a genuine corruption shape, not a caller error, and it is precisely why `LocalDatabase.deleteRecordNoLock` carries it in the same family; dropping it would make a vertex with that damage undeletable again, which is the #4420/#4432 regression. Narrowing it by message text would be brittle. What I have done is name that origin in the comment and mark it explicitly as the member to re-examine first if the arm ever fires in the field, so your periodic second look can be done with the evidence attached instead of re-derived.

**2. Comment-to-code ratio.** Not changing. You called it a taste note and I agree it is dense, but the engine `CLAUDE.md` explicitly asks for the non-obvious rationale to live next to the code, and this is a method whose whole difficulty is *which* failures deserve tolerance - the part that cannot be recovered by reading the statements. The drift risk is real; the mitigation is that each block names the concrete thing it depends on (`LocalDatabase.deleteRecordNoLock`s force flag, `Binary.java`s "Invalid position", `chainsForNeighbour`), so a change that invalidates one is visible rather than silent.

**3. `getMostUpdatedVertex` now unconditional.** Not changing, same reason as the previous pass: it is a `HashMap` get on a path about to do record I/O, and making it conditional is exactly what let the #5660 defect exist - one method owns the rule, and the read either always goes through it or it is a coincidence again.

Verified after the change: `com.arcadedb.graph` 373 tests, 0 failures.

That closes every point raised across the four passes. Recapping what was **not** implemented and why, so it is on the record in one place: the whole-list materialization for super-nodes (unchanged from before this PR; streaming it means walking a list while `deleteEdge` relinks chunks underneath, which is the class of problem this PR fixes - it wants its own issue), the concurrency-test retry budget (deliberately oversized so any throwable reaching the counter is a bug, with the 5-minute ceiling capping the tail), and the two above.

Separately, and not a review point: this work surfaced a **pre-existing defect on the append path** - an edge committed by `newEdge` while its target vertex is being deleted can outlive it (`out` alive, `in` naming the deleted hub). It reproduces with this fix reverted, so it is out of scope here, and the concurrent test is scoped around it rather than over it.

Issue #5630, opened 2026-07-31
[Misc] Flaky CI tests: wall-clock budgets, short read timeouts, and unsettled counters


lvca:
Collected while running #5612 through CI. Six tests failed across its runs; **every one passed when re-run in isolation on the same commit**, and none of them touches anything that PR changed. Recording them so the roster is in one place rather than rediscovered per PR.

| test | how it failed | verified passing alone |
|---|---|---|
| `HttpRedMetricsIT.unmatchedUrisCollapseToBoundedPathTag` | counter read `49`, wanted `>= 50` | 3/3 |
| `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites` | expected `1` got `0`, with a suppressed `DatabaseAreNotIdentical: Types: DB1 6 <> DB2 5` | 3/3 |
| `Issue5410AbandonedTicketReleaseIT` | HA raft, 44 s | 1/1 |
| `Issue5569SlotMergeDeleteRaftIT` | HA raft, 13 s - **also fails on main** (run 30592955402) | 1/1 |
| `Issue3122AsyncParallelCommandsIT.databaseAsyncCommandsRunInParallel` | wall-clock assertion: 26831 ms against a 3000 ms budget | 1/1, at 7 s |
| `Issue5470BatchStreamStallIT` | `Read timed out after 2000 milliseconds` | 3/3 |

Three shapes, and they suggest different fixes:

**Wall-clock budgets.** `Issue3122` asserts a parallel batch finishes inside 3 s. Under load it took 26.8 s and alone it takes 7 s - so the assertion is nine times tighter than an unloaded run on a developer machine, never mind a shared CI runner. A parallelism test wants to assert that concurrent work is *faster than sequential*, or that N threads made progress, not that wall-clock lands under a fixed constant.

**Short read timeouts.** `Issue5470BatchStreamStallIT` fails on a 2 s read timeout while the server-side default it prints is 60 s. Same story: fine unloaded, fragile shared.

**Counter/replication settling.** `HttpRedMetricsIT` off by one, `Bolt5002RoutingTableIT` with a follower one type behind. These read a value before the system has settled; they want a bounded poll rather than an immediate assertion.

Two notes on reading CI for this repo, which cost me time:

[...]

PR #5709, opened 2026-08-01
fix(engine) #5708: walk the script line chain iteratively when closing a batch plan

### LINKED ISSUE #5708: SQL batch script: closing the execution plan recurses one frame per statement, so a large import dies with StackOverflowError
## Problem

Importing data with a large SQL batch script fails with a `StackOverflowError` once the command finishes:

```
Error on command execution (PostCommandHandler)
java.lang.StackOverflowError
    at com.arcadedb.query.sql.executor.AbstractExecutionStep.close(AbstractExecutionStep.java:65)
    at com.arcadedb.query.sql.executor.ScriptLineStep.close(ScriptLineStep.java:108)
    at com.arcadedb.query.sql.executor.AbstractExecutionStep.close(AbstractExecutionStep.java:65)
    ...
```

The statements themselves execute fine, the failure happens while releasing the execution plan.

## Root cause

`SQLScriptQueryEngine.executeInternal()` builds one `ScriptLineStep` per statement and links them into a
`prev` chain via `ScriptExecutionPlan.chain()`. Closing the plan closes the tail step, and the cascade is
recursive:

- `ScriptLineStep.close()` calls `super.close()`
- `AbstractExecutionStep.close()` calls `prev.close()`

so the chain is unwound with about two stack frames per statement. The chain length is driven purely by the
size of the user's batch, so a script of a few thousand statements exhausts the stack.

Execution itself is iterative (`ScriptExecutionPlan.executeUntilReturn()` loops over the step list), which is
why the import runs to completion and only the cleanup blows up.

`ScriptLineStep` is the only step type whose chain length is unbounded: `ForEachExecutionPlan` chains two
steps and `RetryExecutionPlan` chains one.

## Reproduction

A `sqlscript` batch of 20,000 `INSERT` statements, run on a thread with a 1 MB stack, reproduces the exact
frames above.

## Workaround

Split the import into chunks of a few thousand statements per request, or raise `-Xss` on the server JVM.

robfrank:
Fixes #5708

## Problem

Importing data with a large SQL batch script fails with a `StackOverflowError` once the command finishes:

```
Error on command execution (PostCommandHandler)
java.lang.StackOverflowError
    at com.arcadedb.query.sql.executor.AbstractExecutionStep.close(AbstractExecutionStep.java:65)
    at com.arcadedb.query.sql.executor.ScriptLineStep.close(ScriptLineStep.java:108)
    at com.arcadedb.query.sql.executor.AbstractExecutionStep.close(AbstractExecutionStep.java:65)
    ...
```

## Root cause

`SQLScriptQueryEngine.executeInternal()` builds one `ScriptLineStep` per statement and links them into a
`prev` chain via `ScriptExecutionPlan.chain()`. Closing the plan closes the tail step, and the cascade is
recursive: `ScriptLineStep.close()` calls `super.close()`, and `AbstractExecutionStep.close()` calls
`prev.close()`. The chain is therefore unwound with about two stack frames per statement, and its length is
driven purely by the size of the user's batch.

Execution itself is iterative (`ScriptExecutionPlan.executeUntilReturn()` loops over the step list), which is
why the import runs to completion and only the cleanup blows up.

## Fix

`ScriptLineStep.close()` now walks the chain of script lines iteratively.

The cascade in `AbstractExecutionStep` is deliberately left alone: many of the step subclasses do not call
`super.close()`, so a generic iterative loop there would walk past them and change behavior. `ScriptLineStep`
is the only step type whose chain length is unbounded - `ForEachExecutionPlan` chains two steps and
`RetryExecutionPlan` chains one.

Nested script plans keep working: a nested `ScriptExecutionPlan` closed through `plan.close()` adds one frame
per nesting level, which is bounded by the script's block depth rather than by its statement count.

The previous `try`/`finally` intent is preserved and slightly strengthened: a plan that fails to close no
longer stops the walk, and the first failure is rethrown once every line has been released.

## Verification

- New regression test `SQLScriptLargeBatchTest`: a 20,000-statement `sqlscript` batch. It fails with the exact
  reported frames without the fix and passes with it. The batch runs on a thread with an explicit 1 MB stack
  so the outcome does not depend on the CI JVM's default `-Xss`.
- Checked by hand at 150,000 statements: closes in about 2s, confirming nothing else on the path recurses.
- `com.arcadedb.query.sql.**` on the full engine module: 2338 tests, 0 failures, 0 errors.

## Note, out of scope here

`AbstractExecutionStep.sendTimeout()` has the same recursive `prev` cascade. Today no caller propagates a
timeout across script lines (`TimeoutStep` and `AccumulatingTimeoutStep` only fire within a single select
plan), so the script chain never receives it. Worth remembering if timeout propagation is ever extended to
script plans.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
@@ -104,6 +104,8 @@ public class ScriptLineStep extends AbstractExecutionStep {
    * cascade in {@link AbstractExecutionStep#close()} is recursive, which overflows the stack on
    * large imports, hence the chain of script lines is walked iteratively here. A plan that fails to
    * close does not stop the walk: the first failure is rethrown once every line has been released.
+   * An {@link Error} is deliberately not caught and does abort the walk, unlike the previous
+   * try/finally: once the JVM reports one there is no value in releasing the remaining plans.
    */
   @Override
   public void close() {
@@ -121,9 +123,17 @@ public class ScriptLineStep extends AbstractExecutionStep {
       current = step.prev;
     }
 
-    // the chain can end on a step of another kind: from there it is bounded, so let it cascade
-    if (current != null)
-      current.close();
+    // Defensive, and unreachable today: chain() is the only place a script line gets a previous step
+    // and it always passes another script line, so the walk above ends on null. Should the chain ever
+    // end on a step of another kind, that tail is bounded and can cascade the usual way.
+    if (current != null) {
+      try {
+        current.close();
+      } catch (final RuntimeException e) {
+        if (error == null)
+          error = e;
+      }
+    }
 
     if (error != null)
       throw error;
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
@@ -99,14 +99,38 @@ public class ScriptLineStep extends AbstractExecutionStep {
     return false;
   }
 
+  /**
+   * A batch script chains one step per statement, so the chain is as long as the script. The
+   * cascade in {@link AbstractExecutionStep#close()} is recursive, which overflows the stack on
+   * large imports, hence the chain of script lines is walked iteratively here. A plan that fails to
+   * close does not stop the walk: the first failure is rethrown once every line has been released.
+   */
   @Override
   public void close() {
-    try {
-      if (plan != null)
-        plan.close();
-    } finally {
-      super.close();
+    RuntimeException error = null;
+    ScriptLineStep step = this;
+    while (true) {
+      try {
+        if (step.plan != null)
+          step.plan.close();
+      } catch (final RuntimeException e) {
+        if (error == null)
+          error = e;
+      }
+
+      final ExecutionStepInternal previous = step.prev;
+      if (previous instanceof ScriptLineStep previousLine) {
+        step = previousLine;
+        continue;
+      }
+
+      if (previous != null)
+        previous.close();
+      break;
     }
+
+    if (error != null)
+      throw error;
   }
 
   public ExecutionStepInternal executeUntilReturn(final CommandContext context) {
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
@@ -108,8 +108,9 @@ public class ScriptLineStep extends AbstractExecutionStep {
   @Override
   public void close() {
     RuntimeException error = null;
-    ScriptLineStep step = this;
-    while (true) {
+    ExecutionStepInternal current = this;
+
+    while (current instanceof ScriptLineStep step) {
       try {
         if (step.plan != null)
           step.plan.close();
@@ -117,17 +118,12 @@ public class ScriptLineStep extends AbstractExecutionStep {
         if (error == null)
           error = e;
       }
-
-      final ExecutionStepInternal previous = step.prev;
-      if (previous instanceof ScriptLineStep previousLine) {
-        step = previousLine;
-        continue;
-      }
-
-      if (previous != null)
-        previous.close();
-      break;
+      current = step.prev;
     }
+
+    // the chain can end on a step of another kind: from there it is bounded, so let it cascade
+    if (current != null)
+      current.close();
 
     if (error != null)
       throw error;
--- a/engine/src/test/java/com/arcadedb/query/sql/SQLScriptLargeBatchTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/SQLScriptLargeBatchTest.java
@@ -57,6 +57,9 @@ public class SQLScriptLargeBatchTest extends TestHelper {
     runner.join();
 
     assertThat(failure.get()).isNull();
-    assertThat(database.countType("ImportRow", false)).isEqualTo(STATEMENTS);
+    // count(@rid) scans: countType() would read the cached bucket counter instead of ground truth
+    try (final ResultSet counted = database.query("sql", "SELECT count(@rid) AS total FROM ImportRow")) {
+      assertThat(counted.next().<Long>getProperty("total")).isEqualTo(STATEMENTS);
+    }
   }
 }
--- a/engine/src/test/java/com/arcadedb/query/sql/SQLScriptLargeBatchTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/SQLScriptLargeBatchTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.query.sql;
+
+import com.arcadedb.TestHelper;
+import com.arcadedb.query.sql.executor.ResultSet;
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A batch script chains one execution step per statement, so releasing the plan must not walk the
+ * chain recursively: a large import would blow the stack.
+ */
+public class SQLScriptLargeBatchTest extends TestHelper {
+  private static final int STATEMENTS = 20_000;
+
+  @Test
+  void closingALargeBatchScriptDoesNotOverflowTheStack() throws InterruptedException {
+    database.command("sql", "CREATE DOCUMENT TYPE ImportRow");
+
+    final StringBuilder script = new StringBuilder("begin;\n");
+    for (int i = 0; i < STATEMENTS; i++)
+      script.append("INSERT INTO ImportRow SET id = ").append(i).append(";\n");
+    script.append("commit;\n");
+
+    final AtomicReference<Throwable> failure = new AtomicReference<>();
+    // run on a thread with an explicit stack size so the outcome does not depend on the JVM default
+    final Thread runner = new Thread(null, () -> {
+      try (final ResultSet rs = database.command("sqlscript", script.toString())) {
+        while (rs.hasNext())
+          rs.next();
+      } catch (final Throwable t) {
+        failure.set(t);
+      }
+    }, "large-batch", 1024 * 1024);
+
+    runner.start();
+    runner.join();
+
+    assertThat(failure.get()).isNull();
+    assertThat(database.countType("ImportRow", false)).isEqualTo(STATEMENTS);
+  }
+}
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/ScriptLineStepCloseTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/ScriptLineStepCloseTest.java
@@ -81,6 +81,37 @@ public class ScriptLineStepCloseTest {
   }
 
   /**
+   * A step of a kind other than a script line. {@link ScriptExecutionPlan#chain} never puts one behind a
+   * script line, but the chain is typed on {@link ExecutionStepInternal}, so the walk guards against it.
+   */
+  private static class ForeignStep extends AbstractExecutionStep {
+    private final List<String>     closed;
+    private final RuntimeException failure;
+
+    ForeignStep(final CommandContext context, final List<String> closed, final RuntimeException failure) {
+      super(context);
+      this.closed = closed;
+      this.failure = failure;
+    }
+
+    @Override
+    public ResultSet syncPull(final CommandContext context, final int nRecords) {
+      return new InternalResultSet();
+    }
+
+    @Override
+    public void close() {
+      closed.add("foreign");
+      try {
+        if (failure != null)
+          throw failure;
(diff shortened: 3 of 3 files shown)


robfrank:
Thanks both - I verified each point against the code rather than taking it as read. Applied three of the four in 3da69af9e; pushing back on the `@Tag("slow")` one.

**Applied**

1. **Tail close now shares the error slot.** `current.close()` on the non-script-line tail is wrapped in the same capture, so "release every line, rethrow the first failure" holds on that path too.

2. **The tail branch is documented as defensive.** I confirmed the reachability claim: `new ScriptLineStep` appears exactly once in the tree (`ScriptExecutionPlan.chain():139`), and that is the only writer of a script line's previous step, always passing another script line. So the walk ends on `null` today and the branch is unreachable. The comment now says so explicitly.

3. **`countType()` -> `count(@rid)`.** Correct per `engine/CLAUDE.md` - `countType(..., false)` reads the cached bucket counter, not a scan. Switched to `SELECT count(@rid)`.

4. **`Error` narrowing is a deliberate call**, now recorded in the javadoc: an `Error` aborts the walk rather than continuing to release plans, unlike the old `finally`. Good catch that it is a real behavioral change and not incidental.

**Pushing back: `@Tag("slow")`**

The measured elapsed time is **0.296s** (surefire, four runs in a row within noise). The `CLAUDE.md` rule targets tests that "take noticeably long", with "large batches" listed as one symptom of that, not as an independent trigger - a 20k-statement script sounds heavy but the statements are trivial single-field inserts in one transaction.

Tagging it would exclude a 0.3s test from the default CI lane, and that test is the only thing standing between this regression and a silent return. Given the cost/benefit I would rather keep it running on every build. Happy to add the tag if a maintainer prefers the letter of the convention here.

For the record on verification: I re-confirmed the updated test still fails **without** the fix (`StackOverflowError`, exit 1) before re-running it green, so the count-assertion change did not weaken it. Suites re-run after the changes: `SQLScriptLargeBatchTest`, `SQLScriptTest`, `SQLScriptAdditionalCoverageTest`, `BatchTest` = 78 tests, 0 failures.

robfrank:
Correcting one thing in my previous comment on the `@Tag("slow")` point.

I said tagging would "exclude a 0.3s test from the default CI lane" and leave the regression unguarded. That overstated the cost. `.github/workflows/mvn-test.yml` has a dedicated lane:

- `unit-tests`: `-DexcludedGroups=slow,benchmark`
- `slow-unit-tests`: `-pl engine -Dgroups=slow`

This test lives in `engine`, so tagging it would move it to the slow lane rather than drop it from CI. Both lanes run with `--fail-never` and report through the same reporter step, so the gating strength is identical either way. My "it would stop guarding the regression" argument was simply wrong.

What I still think holds is the narrower point: at a measured **0.296s** the tag misdescribes the test. `CLAUDE.md` scopes `@Tag("slow")` to tests that "take noticeably long", and putting a sub-second test in the slow lane makes that signal less meaningful for the tests that genuinely need it.

That said, with the real cost being "runs in a different job" rather than "does not run", this is now purely a labelling preference and I have no strong objection. Happy to add the tag if a maintainer wants consistency with the "large batches" wording - say the word and I will push it.

robfrank:
`unit-tests` is red on this PR, but the failure is unrelated to the change. Details, since a red check should not be waved off:

**The failure** was a single test out of 772 in that module:

```
com.arcadedb.server.ha.raft.ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns
[exactly one staging directory expected] Expected size: 1 but was: 0
```

That is in `ha-raft`. This PR touches two files, both under `engine`: `ScriptLineStep.java` and a new test. Nothing in the Raft state machine or the deferred database deleter goes near the SQL script executor.

**It is a pre-existing race in that test**, which lists the staging directory *after* releasing the deleter that removes it:

```java
gate.countDown();                     // unblocks the deferred deleter
awaitGone(findStagingDirectory());    // ...then lists the staging dirs
```

The sibling test `applyDeregistersTheDatabaseWithoutDeletingItsFilesInline` does it the other way round, resolving the path before `countDown()`.

I proved it rather than assuming: the test passes 6/6 locally unmodified, and inserting `Thread.sleep(500)` between those two lines reproduces the CI assertion exactly and deterministically. Filed as #5718 with the one-line fix.

**This PR's own test did run and pass in CI:**

```
[INFO] Running com.arcadedb.query.sql.SQLScriptLargeBatchTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.377 s
```

Worth noting the 0.377s CI figure is in line with the 0.296s I measured locally, which is the basis for my `@Tag("slow")` pushback above.

For context, `main` itself is currently red on `unit-tests` too, with a different trio of flaky engine tests (`ExplicitLockingTransactionTest`, `Issue4960LSMCloseDuringCompactionTest`, `LSMVectorIndexRecoveryTest`) - none of which failed here.

Everything else is green: build-and-package, integration-tests, builder-tests, all six e2e suites, CodeQL across all languages, Codacy, Meterian, claude-review.

robfrank:
`ha-integration-tests` is also red, and this one is pre-existing too - demonstrably so.

**On this PR** (2 failures of 205):
- `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied`
- `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact`

**On `main`, same workflow, latest completed run** (4 failures of 205):
- `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied`
- `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact`
- `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`
- `RemoteStickyConnectionStrategyIT.pinIsReleasedAfterRollbackAndAllowsSubsequentStickyTransaction`

This PR's failures are a strict subset of what `main` already fails, and every one is Raft replication in `ha-raft` - nothing that reaches the SQL script executor. This matches the pattern tracked in #5702 ("fails on main ~75% of the time, with a different test each run").

Summary of the two red checks on this PR: both are known-flaky HA/Raft tests failing independently of the change. `unit-tests` is #5718 (filed, with proof and a one-line fix); `ha-integration-tests` is #5702. The engine change is covered by `integration-tests`, `build-and-package` and the six e2e suites, all green, plus `SQLScriptLargeBatchTest` passing in CI at 0.377s.

robfrank:
`codecov/patch` failed at **33.33% of diff hit** and it was pointing at something real, so I fixed the gap rather than waive it.

The uncovered lines were the error-handling path in `ScriptLineStep.close()` - which is exactly the behaviour the Javadoc asserts ("release every line, rethrow the first failure") and which had **no test at all**. `SQLScriptLargeBatchTest` only drives the happy path.

Added `ScriptLineStepCloseTest` (bf453bac0): a chain of stub `InternalExecutionPlan`s, asserting both halves of the contract.

- every line is released tail to head
- a plan that throws does not strand the ones behind it, and the tail-most failure is the one that surfaces

I verified the test can fail rather than trusting a green run: the second case fails against the previous `try/finally` implementation, which let the head-most failure replace the original.

```
EXIT_ON_OLD_IMPL=1
[ERROR] ScriptLineStepCloseTest.aPlanThatFailsToCloseDoesNotStrandTheLinesBehindIt:124
```

The one line still uncovered is the `current != null` defensive tail, which is unreachable today for the reason documented in the comment - `chain()` is the only writer of a script line's previous step and always passes another script line. I would rather leave it uncovered than delete a guard or write a test that fakes an impossible chain shape.

Local run with the new test: `ScriptLineStepCloseTest`, `SQLScriptLargeBatchTest`, `SQLScriptTest`, `SQLScriptAdditionalCoverageTest`, `BatchTest` = 80 tests, 0 failures.

robfrank:
Thanks - the RETURN/BREAK observation is a good catch and it checks out.

I verified it: `ScriptLineStep.executeUntilReturn()` builds `new ReturnStep(statement, context)` and the break path builds `new BreakStep(context)`; neither constructor calls `setPrevious(...)` and neither class overrides `close()`, so `AbstractExecutionStep.close()` sees `prev == null` and returns immediately. Every `ScriptLineStep` in `steps` and its plan is then dropped unclosed. `doExecute()` has the same reassignment on the break path.

Filed as **#5720**, kept out of this PR since it is pre-existing and orthogonal - it is a resource-release gap, not a stack-depth one. I noted there that any fix must stay iterative so it does not reintroduce the recursion this PR removes.

On `@Tag("slow")` - one correction to the reasoning offered for it. `ScriptLineStepCloseTest` does **not** cover the regression: it walks a 3-line chain and pins the release contract, but it cannot overflow a stack. `SQLScriptLargeBatchTest` is the only thing guarding the actual `StackOverflowError`, so the two are not substitutes.

That said, I already corrected myself above on the part that mattered: tagging would move the test to the `slow-unit-tests` lane (`-pl engine -Dgroups=slow`), not drop it from CI, so the cost is low. My remaining objection is only that 0.296s locally / 0.377s in CI does not match "takes noticeably long". With three reviews asking for it and the cost being small, I am happy to add the tag - leaving it to a maintainer to call, since it is a convention question rather than a technical one.

robfrank:
`integration-tests` is red on the latest commit. Also unrelated, and this one is already on the books.

**The failure** was one test of 606:

```
com.arcadedb.server.monitor.HttpRedMetricsIT.unmatchedUrisCollapseToBoundedPathTag
Expecting actual: 49L to be greater than or equal to: 50L
```

That is the exact symptom already catalogued in **#5630**, which lists this test by name with the same numbers ("counter read `49`, wanted `>= 50`", reproduced 3/3) under "counter/replication settling": the test fires 50 requests and reads the timer before the recording has settled.

**The delta rules out this PR.** `integration-tests` passed on the previous commit of this same branch (3da69af9e) and failed on bf453bac0. The only difference between those two commits is a new test file, `engine/src/test/.../ScriptLineStepCloseTest.java`. A test-only addition in `engine` cannot change how the server records HTTP RED metrics.

Worth repeating the warning from #5630 for anyone reading these checks: the failing step is the **reporter**, not the Maven step, because Maven runs with `--fail-never`. A green "Run Integration Tests with Coverage" does not mean the tests passed.

Running tally of red checks on this PR, none of them from the change:
- `unit-tests` (earlier commit) -> #5718, root-caused and reproduced deterministically
- `ha-integration-tests` (earlier commit) -> #5702, a strict subset of what `main` fails
- `integration-tests` (this commit) -> #5630, already catalogued by name

robfrank:
## Coverage, and a reversal on my part

`codecov/patch` went 33.33% -> **61.11%** after the contract test, still under the 66.74% target. The remainder was the `current != null` guard for a chain ending on a step of another kind.

I said earlier I would rather leave it uncovered than "write a test that fakes an impossible chain shape". That framing was wrong and I am walking it back. The shape is not impossible - it is unreachable only because `chain()` happens to be the sole writer of a script line's previous step. The chain is typed on `ExecutionStepInternal`, so hanging a plain step behind the run of script lines constructs the guarded shape directly. That is ordinary unit testing of a guard, not gaming a metric.

1356d0c58 adds two cases: the trailing step is released after the lines, and a failure from it is reported like any other.

Verified they cover the branch rather than merely running past it - deleting the branch fails exactly those two:

```
EXIT_WITHOUT_TAIL_BRANCH=1
[ERROR] ScriptLineStepCloseTest.aChainEndingOnAnotherKindOfStepStillReleasesThatStep:176
[ERROR] ScriptLineStepCloseTest.aFailureFromThatTrailingStepIsReportedToo:190
```

Local: `ScriptLineStepCloseTest`, `SQLScriptLargeBatchTest`, `SQLScriptTest`, `SQLScriptAdditionalCoverageTest`, `BatchTest` = 82 tests, 0 failures.

## Status of the previous run (commit bf453bac0)

25 checks green. Three red, all unrelated to this change and each identified:

| Check | Failure | Attribution |
|---|---|---|
| `unit-tests` | `ArcadeStateMachineDeferredDropTest`, 1 of 772 | #5718 - now failed on **both** commits of this PR while passing 6/6 locally, so it is closer to consistent than flaky on the runner |
| `ha-integration-tests` | `Issue5410AbandonedTicketReleaseIT`, 1 of 205 | #5702 - also fails on `main` |
| `integration-tests` | `HttpRedMetricsIT`, 1 of 606 | #5630 - listed there by name with the same `49` vs `>= 50` |

Both of this PR's tests passed in that run: `SQLScriptLargeBatchTest` 1/1 in 0.606s, `ScriptLineStepCloseTest` 2/2 in 0.002s.

robfrank:
## Final status - commit 1356d0c58

**29 checks green, 1 red.**

`codecov/patch` now passes at **88.88% of diff hit** (target 66.74%), having gone 33.33% -> 61.11% -> 88.88% as the two coverage commits landed. `codecov/project` is 67.61%, +0.86%.

The single red is `ha-integration-tests`: `Issue4743LargeRecordReplicationIT`, `Issue5410AbandonedTicketReleaseIT`, `Issue5569SlotMergeDeleteRaftIT`. Across this PR's three runs the failing set has been `{5410, 5569}`, then `{5410}`, then `{4743, 5410, 5569}` - a different subset each time, which is exactly what #5702 documents. All are `server.ha.raft` replication tests; `Issue5410` and `Issue5569` also fail on `main`.

`unit-tests` and `integration-tests`, both red on earlier commits, are green here - confirming #5718 and #5630 as flakes rather than anything from this change.

## Correction

I wrote earlier that #5718 looked "closer to consistent than intermittent" and would redden `unit-tests` on every unrelated PR, on the basis of two failures in two runs. The third run passed, so it is 2 of 3 - a genuine flake, and that claim was too strong. Corrected on the issue. The race diagnosis itself is unaffected and still reproduces deterministically with an inserted delay.

## What this PR contains

Five commits: the fix, a Codacy-driven loop restructure, three review points applied, and two rounds of coverage for the close contract.

Verification that was actually executed rather than reviewed:
- `SQLScriptLargeBatchTest` fails with the reported `StackOverflowError` without the fix and passes with it, re-confirmed after the assertion was changed to `count(@rid)`
- `ScriptLineStepCloseTest` - the error-ordering case fails against the old `try/finally`; the two trailing-step cases fail if the guard is deleted
- 150,000 statements close in ~2s, confirming nothing else on the path recurses
- `com.arcadedb.query.sql.**` = 2338 tests, 0 failures; the five script suites = 82 tests, 0 failures
- in CI: `SQLScriptLargeBatchTest` 1/1, `ScriptLineStepCloseTest` 4/4

## Filed alongside, deliberately not fixed here

- **#5718** - `ArcadeStateMachineDeferredDropTest` ordering race, root-caused with a one-line fix
- **#5720** - a script hitting `RETURN`/`BREAK` never closes the plans of the statements before it, from @claude's review; pre-existing, and any fix must stay iterative

The only open question on the change itself is the `@Tag("slow")` labelling preference, which is a maintainer call.

Issue #4420, opened 2026-05-30
java.lang.NegativeArraySizeException on transaction commit


tomscreen:
Working with a 3-node HA cluster (v26.5.1)

Trying to delete a vertex in Studio... the vertex exists when I do 'SELECT FROM table'
e.g. DELETE VERTEX FROM #136:4

I'm presented with the error...

**Error on transaction commit**
Leader returned HTTP 500 for forwarded command: {"error":"Error on transaction commit","detail":"-51","exception":"java.lang.NegativeArraySizeException"}

Now I have records that can't be deleted.

lvca:
Did you execute the command against the leader?

tomscreen:
I tried both. The same command failed against the leader as well. The error message I posted above is from a follower. On the leader it just said "-51" with no other information. The command was first run on a follower, then on the leader.

lvca:
Could you please attach the logs from the leader? It seems like there is something wrong with the configuration.

tomscreen:
Whenever I run the command that fails on the leader, this is what appears in the log...

2026-05-30 13:51:36.878 INFO  [PostCommandHandler] <arcadedb-2> Error on transaction execution (PostCommandHandler): Transaction error on commit

Would it help to turn on debug logs?

lvca:
I think I found the bug, fixed. Please let me know in case I'm wrong (and in that case, please send more detailed logs).

PR #5710, opened 2026-08-01
feat(engine) #5680: CHECK DATABASE RECORD <rid> checks and repairs named records only

lvca:
Follow-up to #5680, now that #5707 has merged the strict vertex delete.

**This PR was rescoped.** It originally carried its own fix for #5680; #5707 landed an equivalent (and in places wider) fix while this was in review, so everything duplicated there has been dropped and this branch was rebuilt on top of `main`. What remains is the part `main` does not have.

## Why `RECORD`

#5707 made an ordinary `deleteVertex` strict: an edge list it cannot walk is a retryable conflict rather than a short collection, and a genuinely broken chain therefore makes its vertex **undeletable** until `CHECK DATABASE ... FIX` rebuilds the adjacency from the surviving edge records.

That makes `CHECK DATABASE ... FIX` the mandated escape hatch — but it could not be aimed. The narrowest scope was `TYPE` or `BUCKET`, so recovering one vertex meant two full passes over its entire vertex type.

```sql
CHECK DATABASE RECORD #12:3 FIX
CHECK DATABASE RECORD #12:3, #12:9 FIX
```

The per-record work and the fix are identical to the type-wide run, including the edge-list rebuild. Only the enumeration changes: a lookup per listed RID instead of two passes over every bucket of the type.

## Design decisions worth flagging

**Rejected, not silently resolved, when combined with `TYPE`/`BUCKET`.** The grammar accepts the combination; execution refuses it with an `IllegalArgumentException` (which the HTTP layer already maps to 400). `RECORD` is already the narrowest scope, so letting one win silently would run a check nobody asked for, and an intersection would be a third semantics nobody asked for either. Enforcing it in the grammar would be stricter but would lose the tailored message.

**Database-wide passes skipped.** Buckets, external properties and indexes cannot be narrowed to a record, and paying for them would defeat the point of naming one. The orphaned-segment reclaim is skipped for a stronger reason: it builds its reachable set by walking *every* vertex, so under any partial scope it would misclassify unwalked vertices' segments as orphans.

**Two costs it does NOT bound**, stated rather than glossed:

- rebuilding an adjacency means finding every surviving edge that points at the vertex, and no index maps endpoints back to edges — so the scoped run saves the vertex passes and still scans the edge types;
- a listed record that turns out to be *corrupted* still has every index on its bucket dropped and rebuilt. That matches type-wide semantics and only fires on genuine corruption (an edge-list rebuild alone deletes no record and triggers none of it), but `RECORD` bounds the check, not necessarily the fix.

## A test that was passing vacuously

`Issue5670EdgeDeleteDanglingBackRefTest.assertIntegrityClean` asserted `totalErrors == 0`, and `check database` **reports no `totalErrors` field** — the null-tolerant reader turned it into `0 == 0` on every run, whatever the check found. It now asserts on `autoFix`, `invalidLinks`, `totalWarnings` and `totalCorruptedRecords`, and `longProperty` fails loudly on a field that does not exist rather than silently answering 0. Included here because it guards the same area.

## Tests

`CheckDatabaseRecordScopeTest`:

| Test | Shape |
|---|---|
| `checkDatabaseRecordFixMakesTheVertexDeletableAgain` | end-to-end repair: break the chain, scoped FIX rebuilds it, the ordinary delete then completes and takes every edge with it |
| `checkDatabaseRecordVisitsOnlyTheNamedRecord` | proves `RECORD` is a scope — naming a healthy vertex leaves another vertex's broken chain untouched, so a "scoped" run that quietly fell back to a type scan would fail |
| `checkDatabaseRecordRejectsBeingCombinedWithTypeOrBucket` | both `TYPE` and `BUCKET`, plus each clause still working alone |
| `checkDatabaseRecordReportsARidBelongingToNoType` | a RID in no type's bucket is reported, not ignored |

Plus parser coverage in `CheckDatabaseStatementTestParserTest`.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**` and `com.arcadedb.query.sql.parser.**`: **1335 tests, 0 failures** — including `Issue5680VertexDeleteEdgeCollectionTest` from #5707, so the rebase did not disturb the merged fix. `DatabaseChecker`'s `getPageSizeForNewFile()` change from #5713 is preserved (it landed in the same method this PR touches).

https://claude.ai/code/session_01Puky78GhudfgTHPtYSuFPm


lvca:
Round 2 addressed in `4c93ab04b`.

**1. `isCorruptedBuffer` could extend the tolerance down the chain — fixed, with a caveat on the evidence.**

You are right that keying on the exception *type* alone does not express the intent. The tolerance is now scoped by *construction* rather than by classification: `isOwnRecordUnreadable` reads the vertex's own edge-list pointer first, before anything touches the chain, and only a damaged buffer **there** is tolerated. Past that probe a corrupt buffer is a broken chain and gets no escape (`catch` is now plain `if (!force) throw`). Same probe added to `deleteEdgeList`. Cost is one extra read of an already-materialised buffer per direction, against a delete that is about to walk the whole list.

I preferred this to your suggested `edges == null` gate for exactly the reason you flagged about it: a corrupt *head chunk* also leaves `edges == null`, and that case should be strict — the vertex record is fine, the chain is broken, and `CHECK DATABASE RECORD ... FIX` rebuilds it. Probing the vertex record directly splits the two cleanly instead of approximately.

**Caveat worth stating plainly:** I could not build a fixture where the old classification actually stranded edges. I injected a mid-chain chunk whose `used` counter points past its buffer — which does throw `BufferUnderflowException`, i.e. squarely in the tolerated family — and the delete still failed, because the partial collection's per-edge `deleteEdge` calls and `deleteAll()` walk the same damaged chain and those paths are strict since #5678. Mutation-checking the new test against a build with *both* catches reverted to the old classification: it still passes. So `vertexDeleteRefusesWhenAMidChainChunkBufferIsDamagedRatherThanMissing` pins the **contract**, it does not prove the old code was exploitable. I kept the scope fix anyway — it makes the invariant explicit instead of incidental on other paths happening to be strict — but I did not want to claim a demonstrated bug I could not demonstrate.

**2. Causal chain dropped — fixed.** `ConcurrentModificationException` now exposes the `(message, cause)` constructor `NeedRetryException` already had, and `enrichWithRepairHint` uses it. As you say, the run where that exception actually surfaces is the retry-exhausted one, which is precisely the run whose stack trace has to be diagnosable.

**3. `checkScopedDocuments` cap — fixed.** Made consistent with the sibling branch: the retained messages are bounded by `maxWarnings`, the totals still count every occurrence. Not a regression, but the inconsistency was mine to remove.

Verification: `com.arcadedb.graph.**` plus the checker/parser suites, **404 tests, 0 failures**.

lvca:
Round 3 addressed in `c0456de08`.

**1. `catch (Throwable)` in the RECORD-scoped loops — fixed.** Narrowed to `Exception`. Worth spelling out why this mattered more than it looks: in **fix mode** a record recorded as "cannot be loaded" gets **deleted**, so masking an `OutOfMemoryError` as record corruption would have destroyed a healthy record. Good catch.

I deliberately left the *shared* per-record bodies (`checkConnectivity` / `checkEndpoints`) on their existing `Throwable` catch — that is the type-wide path's long-standing behaviour, and changing it would alter the type-wide check, which is outside this fix's remit. Only the enumeration I added is narrowed.

**2. Breadth of `isCorruptedBuffer` — verified rather than assumed, and documented.** You asked for confirmation that no non-corruption path can raise `IAE`/`IOOBE` through the head-pointer read. Traced it:

- `ImmutableVertex.parseEdgePointers` range-checks the 25-byte prefix and throws `SerializationException` explicitly — that *is* the #4432 fix, added precisely to replace a raw `BufferUnderflowException`;
- the lazy load raises `RecordNotFoundException`, handled separately as transient;
- `MutableVertex` returns a field.

So the broad members are not reachable through this call. I kept the set aligned with the family `LocalDatabase.deleteRecord` tolerates rather than trimming it to today's reachable subset — over-tolerating *at the probe* fails toward "a corrupt vertex stays deletable", and trimming it would let a future serializer change silently recreate the #4420/#4432 complaint. Reasoning recorded in the Javadoc.

**3. Log line could misattribute — fixed.** You are right that `ConcurrentModificationException extends NeedRetryException` means a conflict on *this* vertex's own list lands in the same catch. Reworded to "one of its endpoints" rather than guessing.

**4. Hot-path cost — reduced.** Acting on the flag rather than just noting it: `deleteEdgeList`'s `isOwnRecordUnreadable` probe is **redundant** — it only runs for a direction whose collection already passed the probe on the same record, so the edge pointers are known readable. Removed, halving the added reads. The remaining two (one per direction, off an already-materialised buffer) sit in front of a full edge-list walk.

**5. Grammar vs runtime rejection** — agreed, keeping the runtime rejection for the tailored message.

Verification: `com.arcadedb.graph.**` plus the checker/parser suites, **404 tests, 0 failures**.

lvca:
Round 4 addressed in `c6e5de867`. All three taken, and point 1 turned out to matter more than "minor UX".

**1. Repair hint lost from the message on the damaged-buffer path — fixed.**

You were right, and you were also right that the test would not have caught it: `isInstanceOf(Exception.class)` passed on the raw `BufferUnderflowException`. Non-CME failures are now wrapped in `DatabaseOperationException` carrying the hint and the cause, so the operator facing a *permanently* undeletable vertex gets `CHECK DATABASE RECORD ... FIX` in the message, not only in a log line they may never read.

I checked the objection I had originally raised against wrapping (callers branch on the type) rather than trusting it: the corrupt-buffer family `LocalDatabase.deleteRecord` keys on is caught around the **indexer** read, and the `deleteVertex` call site catches `ConcurrentModificationException` **alone**. So wrapping breaks no branching. The retryable conflict is still re-created as its own type — re-typing that one would stop the retry loop absorbing it and break the confirmed-broken-chain fallback.

Same treatment added to the per-edge disconnect loop, naming the edge as well, since the broken list may be the one on the vertex at the other end.

The test now pins the exception type, the `CHECK DATABASE RECORD` text and the root cause, so this cannot silently regress again.

**2. Forced delete leaves a neighbour-side dangling reference — documented.** Added to both the `deleteVertex` Javadoc and the log site: the edge record survives, so the neighbour is left holding a back-reference to an edge whose other endpoint is gone, which `CHECK DATABASE` reports as an invalid link and fix mode drops. Worth having explicit for whoever runs the repair.

**3. Global `TX_RETRY_DELAY` mutation — removed rather than just flagged.** The test no longer touches it. The retry budget was already far beyond what the contention needs, so the default delay costs only wall-clock, and there is now no process-global state for the test to perturb if CI parallelism ever changes.

Verification: `com.arcadedb.graph.**` plus the checker/parser suites, **404 tests, 0 failures**.

lvca:
Both review passes addressed in `bb08db04a`. Four actionable points between them:

**1. A missing RID was treated as corruption — fixed, and this one had teeth.**

`addCorrupted` puts the record's **bucket** into `affectedBuckets`, and under `FIX` the shared tail drops and rebuilds *every index on that bucket* — a full bucket scan, over what is usually a mistyped or already-deleted RID. Since the whole point of `RECORD` is to be hand-typed after a failed delete, that foot-gun bought precisely the cost the scope exists to avoid. `RecordNotFoundException` is now reported and never flagged corrupted, in all three arms (vertex, edge, document).

Mutation-checked rather than assumed: restoring `addCorrupted` on that path fails the new test with `totalCorruptedRecords=1, autoFix=1`.

**2. Progress step double-count — fixed.** Every group bumped `currentStep`, but the document arm then reaches `stepBegin`, which bumps it again, against a `totalSteps` budgeted at one step per group — hence "step 3 of 2". Each arm now owns its bump; the graph arms need theirs because they pass a fixed value into the nested `GraphDatabaseChecker` rather than calling `stepBegin`.

**3. Null RIDs — hardened.** `setRecords` and the statement's stream now drop nulls. Not reachable through the `RECORD #n:n` literal grammar (legacy RIDs never answer null), but `setRecords(Set<RID>)` is public API and `groupRecordsByType` would meet a null as an NPE on `getBucketId()`.

**4. Stale cross-reference — fixed.** The parser test named a class that no longer exists after the rescope.

**Test gaps — closed.** Four added, for the branches nothing exercised: an edge-typed RID, a document-typed RID (the branch the progress bug lived in), several RIDs spanning different types in one command (the point of `groupRecordsByType`), and a missing-but-valid RID under `FIX`, which pins item 1. Eight tests total.

### On the verification numbers

Being precise, since my earlier "1335 green" is worth qualifying: a sweep reported 3 failures (`TimeSeriesTagDictionaryTest`, `LightweightEdgeTypeTest`, `LightweightEdgeBenchmark`), and they were **my own tooling artifact** — I had two Maven runs writing to the same `target/` concurrently, so `Error on creating new database instance` came from leftover database directories. Re-run in isolation on a clean `target/databases`: **26/26 pass**, and none of the three touch code this PR changes. Flagging it rather than quietly reporting the clean number.

lvca:
Addressed in `eb616d951`. Two of the three taken; the third dismissed with reasoning.

**1. Silent fallback to a full-database check — fixed, and it was more reachable than we both thought.**

You judged this "reachable only through the public `setRecords` API or an unusual RID form" because a literal `#x:y` always resolves. That is true of the literal, but the grammar's `rid` rule also accepts the non-literal `{"@rid": <expression>}` form, and `Rid.toRecordId` returns `null` when that expression evaluates to null. So this reaches it straight from SQL:

```sql
CHECK DATABASE RECORD {"@rid": null}
```

and quietly became a whole-database scan. Now refused with a message naming how many records were given and that none resolved. Mutation-checked: stubbing the guard to `if (false)` fails the new test.

Worth flagging since it moves the finding from "defensive only" to "a user can hit this", and on a large database the difference between the requested scope and a full check is not small.

**2. Over-cap warnings dropped silently — fixed.** Extracted `addScopedWarning`, used by both the `type == null` arm and `checkScopedDocuments`: always counted, retained under `maxWarnings`, and **logged** when it has to be dropped, matching `GraphDatabaseChecker.addWarning`.

**3. Duplicated catch blocks — not taken.** A shared helper would need the six accumulators (`warnings`, `totalWarnings`, `maxWarnings`, `corruptedRecords`, `totalCorrupted`, `maxCorrupted`) threaded through it, which reads worse than the six duplicated lines, and the two type-wide methods those arms live in already parallel each other the same way — as you noted in raising it. Happy to revisit if these paths grow a third arm.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1352 tests, 0 failures, BUILD SUCCESS.**

This run also settles the artifact I flagged last time: run as a single Maven invocation with a clean `target/databases`, the three failures I had reported (`TimeSeriesTagDictionaryTest`, `LightweightEdgeTypeTest`, `LightweightEdgeBenchmark`) do not occur. They were caused by two of my own Maven runs sharing a `target/` directory, not by anything in this branch.

lvca:
Addressed in `a8e5e6eab`.

**1. `setRecords` widening — guarded, not just documented.**

You suggested a Javadoc note on the grounds that the distinction "named records but none resolved" vs "named no records" *"genuinely can't be reconstructed inside `setRecords`"*. That was true only because the SQL layer was filtering the nulls out **before** calling it. It no longer does — the nulls are passed through, and `setRecords` owns both the filtering and the refusal. One guard now covers both the SQL path and the embedded caller you described (`new DatabaseChecker(db).setRecords(setOfAllNulls).check()`), and the inconsistency you spotted in my own test comment is gone.

A genuinely empty set — and `null` — still mean "no scope" and are accepted, which the test also pins, so the guard cannot creep into refusing legitimate whole-database checks.

**2. Document arm's "removing it" — fixed.** You judged this inherited and needing no action. It is inherited in `checkDocuments`, but the scoped arm is new code in this PR, and nothing there removes the record (`corruptedRecords` only drives the index rebuild). It now says what is true, with a comment recording why it diverges from the type-wide version — which, as you noted, also calls a document a "vertex".

**3. `maxWarnings` cap test — added.** 40 bogus RIDs against `maxWarnings=10`: 10 messages retained, 40 counted. That pins `addScopedWarning`, which was previously only exercised incidentally.

**Not added:** an execution test for `RECORD ... FIX COMPRESS`. The compress pass is shared, unscoped code already covered elsewhere; routing it through the scoped enumeration exercises no new branch, so the test would cost a compress pass and assert nothing this PR changed. Happy to add it if you disagree.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1353 tests, 0 failures, BUILD SUCCESS.**

lvca:
Addressed in `9db8cac03`. Two of the three notes taken.

**1. `COMPRESS` under a RECORD scope — documented.** You are right that it is the one clause breaking the "naming a record bounds the cost" promise. It now sits in the release notes beside the other two unbounded costs, worded so nobody pairs `RECORD ... COMPRESS` expecting a cheap run.

**2. Corrupted-record `FIX` coverage — added.** This was the best of the three suggestions: it converts the "RECORD bounds the check, not necessarily the fix" paragraph from prose into an executable assertion, so the one cost the design deliberately does not bound can't drift silently in either direction.

`checkDatabaseRecordFixRebuildsTheBucketIndexesOfAGenuinelyCorruptedRecord` shrinks a vertex's record-size varint below the fixed 25-byte prefix, so the record still occupies its slot but no longer reads back as a vertex — genuinely corrupted, distinct from both the missing-RID and the chain-broken cases the other tests cover. It asserts the record is flagged and that `rebuiltIndexes` is non-empty.

Mutation-checked: dropping `addCorrupted` from the scoped arm fails it with `totalCorruptedRecords: 0, rebuiltIndexes: []`, and the warning confirms the intended corruption shape — `record buffer is truncated or corrupted (size=8, expected at least 25)`.

**3. Setter-time rejection of `TYPE`/`BUCKET` + `RECORD` — not taken.** `setTypes` and `setRecords` can be called in either order, so a setter-level guard would fire or not depending on call order, which is worse than a consistent failure. `check()` is the only point that sees the complete configuration. You called the current placement acceptable and I agree; flagging the reasoning rather than leaving it implicit.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1354 tests, 0 failures, BUILD SUCCESS.**

lvca:
Addressed in `95358fb74`. Three taken, two declined with reasoning.

**1. The `Exception`-not-`Throwable` comment oversold the protection — fair, and fixed.** You are right: the guard covers the *lookup*, but the shared per-record consumer still catches `Throwable` around the connectivity check itself, so an `Error` raised in there is still flagged. That matches the type-wide path, which is why it stays — but my comment implied a guarantee the code does not make. It now states exactly what is and is not covered. Good catch on a claim rather than on code.

**2. Partial null RIDs dropped silently — now surfaced.** You asked whether this was intended. It is: narrowing cannot do more than the caller asked for, unlike widening, so refusing would be too harsh. But silent was wrong — a caller who mistyped one of several RIDs got a clean report for a check that skipped it. `check()` now emits a warning naming how many were dropped, with a test (`checkDatabaseRecordReportsRidsItHadToDrop`).

**3. No-type group advanced the step counter without naming the step — fixed.** Routed through `stepBegin` like the document arm, which also owns the bump, so a progress poller no longer sees a blank step go by.

**4. `RECORD ... COMPRESS` asymmetry — declined, and I think the asymmetry is correct.** The two cases differ in kind: `RECORD` + `TYPE`/`BUCKET` has *no coherent meaning* (three competing answers to "what is the scope?"), which is why it is refused. "Check this record, then compress the database" is perfectly meaningful — it is just not cheap. Refusing a legitimate combination because it is expensive would be a worse trade than documenting the cost, which the release note now does explicitly.

**5. `@Tag("slow")` on the 200-edge tests — declined, measured rather than assumed.** The whole class runs in **0.53s**. CLAUDE.md's criteria are multi-second elapsed time or large batches; 200 tiny transactions in half a second is neither, and tagging would move them out of the default lane for no benefit.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1355 tests, 0 failures, BUILD SUCCESS.**

lvca:
Addressed in `81c2a9b99`. All four taken.

**1. Wrong error for `TYPE ... RECORD {@rid: null}` — fixed.** You are right that the resolution failure was masking the real mistake. The statement now diagnoses the clause combination *before* resolving any RID, so the operator is told about the outer problem; `check()` keeps its own guard for programmatic callers. Tested.

**2. `droppedRecords` stale on the early-return path — fixed.** Harmless today for exactly the reason you gave, but it is public-API-visible state and should not depend on call order. Reset on that branch.

**3. `RECORD ... COMPRESS` runtime warning — taken, and it is a better answer than mine.** I had argued against *refusing* the combination (it is meaningful, just not cheap) and settled for a release-note line. Your middle ground is strictly better: the command still does what was asked, and an operator who named a record and expected a bounded run is told plainly that it will not be. Tested both with and without `COMPRESS`, so the warning tracks the clause rather than the scope.

**4. Scoped vs type-wide document totals — recorded, not silently diverged.** Agreed the scoped arm is the more correct one. I have not changed the type-wide path: that would alter what `CHECK DATABASE TYPE ...` reports for every user, and this issue does not otherwise touch that command. The divergence is now stated in the Javadoc so it is a known, deliberate difference rather than something to be rediscovered — happy to file it as a follow-up if you would like it aligned.

**5. Duplicated catch blocks** — agreed, no change.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1356 tests, 0 failures, BUILD SUCCESS.**

lvca:
Addressed in `0cc58f073`. All four taken.

**1. Duplicated error message — one owner.** `DatabaseChecker.RECORD_SCOPE_CONFLICT_ERROR`, referenced by the statement. The two-layer guard stays (the statement must diagnose the clause conflict before RIDs are resolved, which is what the previous round fixed), but the wording can no longer drift.

**2. Count collapses when several RIDs resolve to null — fixed by removing the count, not by plumbing one through.** You are right that a `Set` dedupes every null into one element. The consequence is stronger than a nit: `"given 3 record(s)"` was *never reachable*, and the dropped-records warning always said "1" no matter how many were dropped. So any number those messages quoted was wrong by construction. Both now say "none of the records given" / "one or more of the records given", which is accurate whatever the input, and neither needs the original count threaded through the API.

**3. `checkScopedDocuments` grew `corruptedRecords` uncapped — fixed.** A real inconsistency: every other path in the file bounds the retained set while letting the total count on. The document arm now does the same.

**4. Edge sweep multiplies by distinct vertex types — documented.** Stated in both the `records` field Javadoc and the release note, with the concrete contrast so the cost is predictable: ten vertices of one type is one sweep; one vertex of each of three types is three.

## Verification

`com.arcadedb.graph.**`, `com.arcadedb.engine.**`, `com.arcadedb.query.sql.parser.**` plus `CheckDatabaseExtendedTest`: **1356 tests, 0 failures, BUILD SUCCESS.**

PR #5716, opened 2026-08-01
fix(server) #5711: an HTTP query no longer silently truncates at the serializer limit

### LINKED ISSUE #5711: HTTP query endpoint silently truncates at 20,000 rows: a truncated response is indistinguishable from a complete one
## Summary

`POST /api/v1/query/{db}` serializes at most `AbstractQueryHandler.DEFAULT_LIMIT` (20,000) rows, and a truncated response is indistinguishable from a complete one. Same `200`, same body shape, no flag, no returned count, no warning. The SQL `LIMIT` the caller wrote is silently overridden by the serializer's default.

This is the same shape as #5639, except the input that is accepted-and-ignored is the caller's own query rather than a METADATA key.

## Reproduction

30,000 documents, then the identical query over both surfaces:

```
SELECT i FROM R LIMIT 30000
```

| request | status | rows returned | response keys besides `result` |
|---|---|---|---|
| `{"language":"sql","command":"SELECT i FROM R LIMIT 30000"}` | 200 | **20,000** | `{"user":"root"}` |
| same + `"limit": 30000` | 200 | 30,000 | `{"user":"root"}` |
| same + `"limit": -1` | 200 | 30,000 | `{"user":"root"}` |
| embedded `db.query(...)`, same SQL | n/a | 30,000 | n/a |

Measured on `26.8.1.dev24`, in-process server, one process, so the two surfaces are the same engine instance over the same data.

## Why I think this is worth fixing

The cap itself is reasonable and I would not remove it. It is overridable per request, which is the right design. The problem is only that the caller cannot tell it happened.

Concretely, a client that pages by writing `LIMIT` into its SQL, which is the obvious thing to do, gets 20,000 rows for any page larger than that and no signal at all. Whatever it computes from that answer is wrong, and nothing in the response says so. `200` with a plausible-looking array is the worst possible failure mode for a data-correctness bug: it looks like success.

It also makes the HTTP and embedded surfaces silently disagree about the result of the same query against the same database, which is the property that caught this. I was measuring embedded against in-process HTTP and the row counts diverged at 100k. Without an equality check between the two arms I would have published "HTTP is 4.7x faster at 100k rows", which is entirely an artifact of it returning a fifth of the rows.

## Suggestions, in rough order of cheapness

1. Put the effective limit and the returned count in the response, always. Something like `"limit": 20000, "returned": 20000, "truncated": true`. Cheap, backward compatible, and a client can then detect it.
2. Fail loudly instead when the caller's SQL `LIMIT` exceeds the serializer limit, since in that case the caller has stated an explicit expectation the server cannot meet. A `400` naming the setting would be clearer than a quiet 20,000.
3. If neither, at least log a warning server side so an operator can find it.

Happy to test a fix on the same setup.

## Environment

`26.8.1.dev24`, in-process server via the Python bindings, single container, Linux x86-64.

lvca:
Fixes #5711

## The defect

`POST /api/v1/query|command` capped the serialized rows at `AbstractQueryHandler.DEFAULT_LIMIT` (20,000) **after** the engine had produced them, and reported the cap nowhere: same `200`, same body shape, no flag, no count. A caller paging by writing `LIMIT` into its own SQL - the obvious thing to do - got 20,000 rows for any larger page and nothing in the answer said so. It also made the HTTP and embedded surfaces disagree on the same query against the same data, which is how @tae898 caught it (a "HTTP is 4.7x faster at 100k rows" measurement that was really HTTP returning a fifth of the rows).

Confirmed as reported, and the asymmetry is worse than the issue states: `GET /query` **already** read the LIMIT back from the execution plan and honored it, so the two HTTP surfaces disagreed with each other too. The Java remote driver sends no `limit` at all, so it inherited the same silent truncation.

## What this changes

**1. A limit the caller stated is honored as written.** The cap is the request's `limit` when present, otherwise the configured default *raised to* the LIMIT the query carries (read back from the execution plan). `SELECT i FROM R LIMIT 30000` now returns 30,000 rows over HTTP exactly as embedded.

The plan LIMIT can only **raise** the cap, never lower it - deliberately, and not just for simplicity:
- a LIMIT smaller than the default is already enforced by the engine, so capping at it would be redundant;
- a plan carrying a **parameterized** LIMIT (`LIMIT :max`) reports the value of the execution it was built for, and `ExecutionPlanCache.get()` hands back a `copy(context)` that keeps that int. Today a plan containing a `LimitExecutionStep` is never cached (`ExecutionStepInternal.canBeCached()` defaults to false and `LimitExecutionStep` does not override it), so the value is always fresh - but trusting it downwards would make the cap depend on that implementation detail. Raising only is immune to it.

**2. Truncation is always visible.** Both endpoints now always report `limit`, `returned` and `truncated`:

```json
{ "result": [ ... ], "limit": 20000, "returned": 20000, "truncated": true }
```

`truncated` is true only when the cap stopped serialization with at least one row still pending, so a result ending exactly at the cap is not a false positive. To make that detectable for a command with no LIMIT of its own, the pushed-down LIMIT is now `cap + 1`: the extra row is the probe and never reaches the client.

**3. A warning, but only for the truncation nobody asked for.** The server logs `WARNING` naming the database, the cap, the row count and the (abbreviated) query - only when the *default* did the cutting. Truncation against a `limit` the caller sent is the expected outcome and stays silent, so Studio's paging does not flood the log.

**4. The cap is now an operator setting**: `arcadedb.server.httpQueryDefaultLimit` (default 20000, `-1`/`0` unlimited). It applies only to callers that state no limit of their own.

**5. `POST /timeseries/{db}/query`** had the same shape (`count = min(rows, limit)`, no signal) and now reports `limit`/`truncated` too.

**6. Java remote driver.** `RemoteDatabase` warns when the server reports a truncated result instead of handing the application a partial `ResultSet` that looks complete, and `setMaxResultRows(Integer)` sets the cap per connection - `-1` removes it, giving embedded parity for a query with no LIMIT.

**7. Studio** marks the row count `(truncated)` when the result it displays was cut short.

### On the issue's suggestion 2 (fail with 400 when the SQL LIMIT exceeds the cap)

Not implemented, deliberately: with (1) that case no longer exists - the SQL LIMIT *is* the cap. The only truncation left against an explicit statement is when the caller *also* sent a smaller `limit`, where the request field is the more specific instruction and a 400 would be wrong. This also keeps the fix backward compatible for clients that write large LIMITs today.

### Not in scope, worth its own issue

- **A hard ceiling** (like `arcadedb.server.grpcQueryMaxResultRows`, which fails with `RESOURCE_EXHAUSTED` rather than truncating). Honoring a caller's LIMIT does not introduce a new capability - `"limit": -1` has always been an unbounded escape hatch for the same authenticated actor - but any *finite* default ceiling would break existing `limit: -1` users, so it needs its own decision.
- **Cypher cannot expose its LIMIT.** A non-EXPLAIN Cypher `ResultSet` carries no `ExecutionPlan`, so `MATCH ... RETURN n LIMIT 30000` is still capped by the default - now flagged and logged instead of silent. Wiring a plan (or the top-level RETURN's LIMIT) through the Cypher engine would close that gap.

## Behaviour changes

- A serializer name other than `graph`/`studio`/`record` used to emit `limit + 1` rows; it now emits exactly `limit`, like the others.
- On `GET /query` an explicit `limit` parameter now wins over the LIMIT in the query text (it used to be the reverse), matching POST.
- `"limit": 0` now means unlimited everywhere, as it already did inside the serializer. It used to be pushed down as a literal `LIMIT 0` (zero rows) for a query with no LIMIT, while returning every row for a query that had one.
- Responses carry three new top-level fields. Additive, and the driver/Studio consumers read `result` by name.

## Verification

- **`HttpQueryTruncationIT`** (new, 18 tests) - lowers the cap to 50 via `onServerConfiguration` so the semantics are checked without 20k rows: query LIMIT above the cap honored; default-cap truncation reported; a result ending exactly at the cap not flagged; request `limit` winning over the query LIMIT; `-1` and `0` unlimited; HTTP vs embedded row-count parity; `studio`, `graph` and default serializers; Cypher flagged; parameterized LIMIT honored per execution; a write command still reporting the fields; the three GET cases; and the remote driver (query LIMIT honored, `setMaxResultRows(-1)` removing the cap).
- **`HTTPDocumentIT`** - two tests at the *shipped* 20,000 default over 20,002 records: `checkQueryInPostWithLimitAboveDefaultCut` (returns 20,001, `truncated: false`) is the direct regression test - it returns 20,000 on `main` - and `checkQueryInPostReportsDefaultLimitTruncation` pins the flagged path. Complements the pre-existing GET-side `checkQueryInGetWithLimitAboveDefaultCut`.
- **`TimeSeriesQueryHandlerIT.rawQueryReportsTruncation`**, **`PostCommandHandlerAutoLimitTest`** (+3: `limit: 0`, probe saturation).
- Suites run green: `com.arcadedb.server.http.**` (496), `HTTPDocumentIT`, `HTTPGraphIT`, `HTTPTransactionIT`, `AbstractQueryHandler*`, `PostCommand*`, `Issue4689*`, all `Remote*`/`*Remote*IT` (80), `TimeSeriesQueryHandlerIT`, `console`.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1498,7 +1498,9 @@ did not.
 `truncated` is true only when the cap stopped the serialization with at least one row still pending, so a result
 that ends exactly at the cap is not flagged. The server also logs a warning naming the database, the cap and the
 query - but only when the *default* did the cutting, since truncation against a `limit` the caller asked for is
-the expected outcome. The `POST /timeseries/{db}/query` endpoint reports `limit` and `truncated` the same way.
+the expected outcome. `POST /timeseries/{db}/query` reports `limit` and `truncated` the same way, reads the same
+setting instead of its own hardcoded 20,000, and no longer answers a non-positive `limit` with zero rows (it used
+to compute `min(rows, -1)`, so `limit: -1` returned nothing at all).
 
 The cap is now configurable with **`arcadedb.server.httpQueryDefaultLimit`** (default 20000, `-1` for unlimited);
 it applies only to callers that state no limit of their own.
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1471,4 +1471,50 @@ release is not readable by an earlier build: the old loader does not recognise `
 metadata page as corrupt. The database still opens and every other index is unaffected, but that one index has to be
 dropped before going back. `LSM_TREE` indexes on the same properties are unaffected in both directions.
 
+## HTTP: a truncated query response is no longer indistinguishable from a complete one
+
+The HTTP query and command endpoints serialize at most a fixed number of rows into a response, 20,000 by default.
+The cap was applied after the engine had produced the rows and was reported nowhere: same `200`, same body shape,
+no flag and no count. A caller that paged by writing `LIMIT` into its own SQL - the obvious thing to do - got
+20,000 rows for any page larger than that and nothing in the answer said so
+([#5711](https://github.com/ArcadeData/arcadedb/issues/5711)).
+
+Two things were wrong, and both are fixed.
+
+**A limit the caller stated is now honored as written.** The row cap is the `limit` field of the request when
+there is one; otherwise it is the server default, raised to the LIMIT the query itself carries. So
+`SELECT i FROM R LIMIT 30000` now returns 30,000 rows over HTTP exactly as it does embedded, while a query
+stating no limit at all is still capped. The query's LIMIT only ever raises the cap: a smaller LIMIT is already
+enforced by the engine, so the serializer has nothing to add. This also removes an asymmetry between the two
+surfaces: `GET /query` already read the LIMIT back from the execution plan, `POST /query` and `POST /command`
+did not.
+
+**When the default cap does bite, the response says so.** The query and command endpoints now always report:
+
+```json
+{ "result": [ ... ], "limit": 20000, "returned": 20000, "truncated": true }
+```
+
+`truncated` is true only when the cap stopped the serialization with at least one row still pending, so a result
+that ends exactly at the cap is not flagged. The server also logs a warning naming the database, the cap and the
+query - but only when the *default* did the cutting, since truncation against a `limit` the caller asked for is
+the expected outcome. The `POST /timeseries/{db}/query` endpoint reports `limit` and `truncated` the same way.
+
+The cap is now configurable with **`arcadedb.server.httpQueryDefaultLimit`** (default 20000, `-1` for unlimited);
+it applies only to callers that state no limit of their own.
+
+**Java remote driver.** `RemoteDatabase` warns when the server reports a truncated result instead of handing back
+a partial `ResultSet` that looks complete, and `setMaxResultRows(Integer)` sets the cap for that connection -
+`-1` removes it, so a query with no LIMIT of its own returns every row like the embedded API does. Beware that
+removing the cap makes the server materialize the whole result set in memory before answering.
+
+**Studio** marks the row count with `(truncated)` when the result it is showing was cut short.
+
+Three smaller behaviour changes come with this. A serializer name other than `graph`, `studio` or `record` used
+to emit one row *above* the cap; it now emits exactly the cap, like the others. On `GET /query`, an explicit
+`limit` query parameter now wins over the LIMIT in the query text (it used to be the other way round), which is
+what the POST endpoints have always done. And `"limit": 0` now means unlimited everywhere, as it already did in
+the serializer: it used to be pushed down as a literal `LIMIT 0` and return no row at all for a query that
+carried no LIMIT of its own, while returning every row for a query that did.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1513,6 +1513,11 @@ non-positive `limit` with zero rows (it used to compute `min(rows, -1)`, so `lim
 The cap is now configurable with **`arcadedb.server.httpQueryDefaultLimit`** (default 20000, `-1` for unlimited);
 it applies only to callers that state no limit of their own.
 
+One gap is worth knowing about: a query's LIMIT raises the cap only for a language that exposes it on the
+execution plan. SQL does; Cypher on the non-EXPLAIN path builds no plan, so `MATCH ... RETURN n LIMIT 30000` is
+still cut at the default - now with `truncated: true` and a warning rather than silently. Send `limit` in the
+request for those languages.
+
 **Java remote driver.** `RemoteDatabase` warns when the server reports a truncated result instead of handing back
 a partial `ResultSet` that looks complete, and `setMaxResultRows(Integer)` sets the cap for that connection -
 `-1` removes it, so a query with no LIMIT of its own returns every row like the embedded API does. Beware that
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1498,9 +1498,17 @@ did not.
 `truncated` is true only when the cap stopped the serialization with at least one row still pending, so a result
 that ends exactly at the cap is not flagged. The server also logs a warning naming the database, the cap and the
 query - but only when the *default* did the cutting, since truncation against a `limit` the caller asked for is
-the expected outcome. `POST /timeseries/{db}/query` reports `limit` and `truncated` the same way, reads the same
-setting instead of its own hardcoded 20,000, and no longer answers a non-positive `limit` with zero rows (it used
-to compute `min(rows, -1)`, so `limit: -1` returned nothing at all).
+the expected outcome.
+
+The flag is exact for the row-oriented serializers (`record`, `studio` and the default one), where `returned` also
+never exceeds `limit`. With `serializer: "graph"` the cap counts graph *elements* instead of rows, so two things
+differ: `returned` can come back above `limit`, because the row that reaches the cap is expanded whole; and since
+the same vertex reached twice is serialized once, `truncated` is best-effort there - a result whose rows dedup
+down to fewer elements than the cap can read `false`. A paging client should use a row-oriented serializer.
+
+`POST /timeseries/{db}/query` reports `limit` and `truncated` the same way, logs the same warning when its own
+default did the cutting, reads the same setting instead of its own hardcoded 20,000, and no longer answers a
+non-positive `limit` with zero rows (it used to compute `min(rows, -1)`, so `limit: -1` returned nothing at all).
 
 The cap is now configurable with **`arcadedb.server.httpQueryDefaultLimit`** (default 20000, `-1` for unlimited);
 it applies only to callers that state no limit of their own.
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -941,6 +941,16 @@ public enum GlobalConfiguration {
       "Maximum size in bytes for HTTP request body content. Set to -1 for unlimited size (WARNING: removes DoS protection). Default is 100MB",
       Long.class, 100L * 1024 * 1024), // 100MB DEFAULT
 
+  SERVER_HTTP_QUERY_DEFAULT_LIMIT("arcadedb.server.httpQueryDefaultLimit", SCOPE.SERVER,
+      """
+      Default maximum number of rows the HTTP query/command endpoints serialize into a single response when \
+      the caller states no limit of its own. A request that carries a `limit` field, and a query that carries \
+      its own LIMIT clause, are both honored as written and are never capped by this value. When this default \
+      does cut a result short the response reports `"truncated": true` next to `returned` and `limit`, and the \
+      server logs a warning: the truncation is never silent (issue #5711). Set to -1 or 0 for unlimited \
+      (WARNING: removes the protection against materializing an unbounded result set in memory). Default is 20000""",
+      Integer.class, 20_000),
+
   SERVER_HTTP_STREAMING_READ_TIMEOUT("arcadedb.server.httpStreamingReadTimeout", SCOPE.SERVER,
       """
       Budget in milliseconds granted to endpoints that consume the request body while working (today only \
--- a/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
@@ -752,10 +752,12 @@ public class RemoteDatabase extends RemoteHttpComponent implements BasicDatabase
   protected ResultSet createResultSet(final JSONObject response) {
     final ResultSet resultSet = new InternalResultSet();
 
-    if (response.getBoolean("truncated", false))
+    if (getMaxResultRows() == null && response.getBoolean("truncated", false))
       // The server dropped rows this driver never asked to drop: without this the caller would receive a
       // partial ResultSet indistinguishable from a complete one, and would silently disagree with the same
-      // query executed on the embedded API (issue #5711).
+      // query executed on the embedded API (issue #5711). An application that set its own cap with
+      // setMaxResultRows() asked for the truncation, so it is not warned about it - the same rule the server
+      // applies to a request carrying its own 'limit', and what keeps a paging application out of the log.
       LogManager.instance().log(this, Level.WARNING,
           "The server truncated the result set to %d rows (its limit is %d): the returned result is incomplete. Add an explicit "
               + "LIMIT to the query, or raise the cap with RemoteDatabase.setMaxResultRows().",
--- a/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
@@ -752,6 +752,15 @@ public class RemoteDatabase extends RemoteHttpComponent implements BasicDatabase
   protected ResultSet createResultSet(final JSONObject response) {
     final ResultSet resultSet = new InternalResultSet();
 
+    if (response.getBoolean("truncated", false))
+      // The server dropped rows this driver never asked to drop: without this the caller would receive a
+      // partial ResultSet indistinguishable from a complete one, and would silently disagree with the same
+      // query executed on the embedded API (issue #5711).
+      LogManager.instance().log(this, Level.WARNING,
+          "The server truncated the result set to %d rows (its limit is %d): the returned result is incomplete. Add an explicit "
+              + "LIMIT to the query, or raise the cap with RemoteDatabase.setMaxResultRows().",
+          response.getInt("returned", -1), response.getInt("limit", -1));
+
     final JSONArray resultArray = response.getJSONArray("result");
     for (int i = 0; i < resultArray.length(); ++i) {
       final JSONObject result = resultArray.getJSONObject(i);
--- a/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
@@ -85,7 +85,9 @@ public class RemoteHttpComponent extends RWLockContext {
   protected        String                      currentServer;
   protected        int                         currentPort;
   private         Pair<String, Integer>       stickyTransactionServer;
-  private         Integer                     maxResultRows;
+  // Volatile so a setMaxResultRows() from another thread is seen by the command-build path: this class is
+  // documented as not thread safe, but the cap is a connection-wide knob an application may flip at any time.
+  private volatile Integer                    maxResultRows;
 
   public enum CONNECTION_STRATEGY {
     STICKY, ROUND_ROBIN, FIXED
--- a/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
@@ -85,6 +85,7 @@ public class RemoteHttpComponent extends RWLockContext {
   protected        String                      currentServer;
   protected        int                         currentPort;
   private         Pair<String, Integer>       stickyTransactionServer;
+  private         Integer                     maxResultRows;
 
   public enum CONNECTION_STRATEGY {
     STICKY, ROUND_ROBIN, FIXED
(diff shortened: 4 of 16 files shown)


lvca:
Thanks - all four points reviewed. Two fixed in 66981ec38, two declined with reasons.

**1. TimeSeries `limit: -1` returned zero rows — fixed, and it was a real bug.** `min(rows, -1) = -1` meant the loop never ran, so a caller following the `-1 = unlimited` convention this PR establishes got an empty result *reported as truncated*. Now a non-positive limit means unlimited there too. Also removed the second hardcoded `20_000`: the config read moved to `AbstractServerHttpHandler.getDefaultRowLimit()` so every row-returning endpoint shares `arcadedb.server.httpQueryDefaultLimit`. Covered by an extra arm in `TimeSeriesQueryHandlerIT.rawQueryReportsTruncation` (`limit: -1` → all 3 rows, `truncated: false`).

**2. `maxResultRows` visibility — fixed.** Now `volatile`, with a comment saying why: the class is documented as not thread safe, but this particular field is a connection-wide knob an application may flip while queries are in flight, so the cheap guarantee is worth having.

**3. Per-query warning noise — keeping as is.** Each truncated `ResultSet` is a wrong answer handed to the application, so one `WARNING` per occurrence is proportional; suppressing repeats (once per connection, say) would hide exactly the later occurrences an operator needs to see. It is also self-limiting: the fix removes the case that used to trigger it most often — a query that states its own LIMIT no longer truncates at all — and any app that finds the volume unwelcome silences it in one line with `setMaxResultRows(-1)` or an explicit LIMIT, which is what the message says.

**4. EXPLAIN reporting `limit: n` with `returned: 0` — leaving as is.** The three fields describe the cap that was in force and what the response carries, and for `EXPLAIN` all three are accurate: the cap was `n`, no rows were serialized, nothing was dropped. Special-casing it would either make `limit` absent (breaking the "test `truncated` unconditionally" contract the fix is built on) or report a cap that was not the one applied.

lvca:
Round 2 addressed in 0a2d843c4.

**1. Driver warning under `setMaxResultRows`-based paging — valid, fixed, and you talked me out of my round-1 answer.** My reasoning there ("every truncated ResultSet is a wrong answer") only holds when nobody asked for the cap. When the application set the page size itself the truncation is its own choice, so the driver now applies exactly the server rule and warns only when `maxResultRows == null`. The warning keeps meaning "you were handed a partial answer you did not ask for". New test `theRemoteDriverWarnsOnlyWhenItDidNotAskForTheCap` captures the `RemoteDatabase` logger and asserts both halves: warned with no cap set, silent with `setMaxResultRows(10)` (and still exactly 10 rows).

**2. `returned` under the `graph` serializer — documented in the spec** ("With the 'graph' serializer, whose cap counts graph elements rather than rows, it is the number of serialized vertices plus edges").

**3. `returned` vs `recordCount` — resolved by deleting the overlap rather than explaining it.** `recordCount` and `executionTime` were documented on the query response but **no handler has ever emitted either** (`grep` finds them only in `CoreApiSpec`), so the spec was promising fields a client cannot find. Both are gone; `returned` is the real row count and timings are reported under `profile` when `profileExecution` is set. Left a comment saying so, so they do not get re-added.

**4. Unlimited materializes everything — agreed, no change.** Same disposition as the description: the `-1` escape hatch predates this PR, and a hard ceiling needs its own issue because any finite default breaks existing `limit: -1` callers.

**Nit: `limit: 0` on the timeseries endpoint** — the test now loops over `{-1, 0}`, so both halves of the unlimited pair are pinned.

Tests: `HttpQueryTruncationIT` (19), `TimeSeriesQueryHandlerIT` (11), all `openapi` spec tests (36), `OpenApiSpecGeneratorTest` - green.

lvca:
Round 3 addressed in 560dc1dd4 - all four points, documentation and coverage only, no behaviour change.

**1. `graph`: `returned` can exceed `limit` — documented.** Both the OpenAPI field description and the `SerializationOutcome` Javadoc now state it and why: the cap counts graph elements, and the row that reaches it is expanded whole rather than cut in half. The row-oriented serializers keep the stronger guarantee (`returned <= limit`), which is now stated explicitly too, so a client knows which contract applies to which serializer.

**2. Multi-statement `sqlScript` — documented.** `getPlanLimit` now says a multi-statement script reports 0 today, and that a value belonging to one of its statements would be harmless anyway: `resolveLimit` only raises the cap with it, so the worst case is a wider response, never a shorter one. That is the same invariant the parameterized-LIMIT argument rests on.

**3. Timeseries fetch is materialized in full — noted where it matters.** Comment added at the handler's `limit` read: the cap there governs serialization only, the range is already materialized before it applies, so `-1` widens nothing. Left the config Javadoc's DoS wording as is, since it describes the query/command path the setting is named for.

**4. Exact-cap boundary with no explicit LIMIT — now asserted.** `aResultOfExactlyTheDefaultCapIsNotFlagged` queries `WHERE i < 50` with the cap at 50: the handler pushes `limit 51`, the engine returns exactly 50, and the response must report `truncated: false`. That is the false-positive guard for the probe mechanism itself, and it was the one boundary only covered implicitly.

Tests: `HttpQueryTruncationIT` (20), `TimeSeriesQueryHandlerIT` (11), `CoreApiSpecTest`, `OpenApiSpecGeneratorTest` - green.

lvca:
Round 4 addressed in a7801be9c, plus the follow-up issue you recommended.

**Main consideration — hard ceiling: filed as #5719.** Written up with the three options (default `-1`, a finite default, or mode-dependent), the reason none of them is free (any finite default breaks existing `limit: -1` callers), the gRPC precedent verbatim, and the note that a ceiling on the timeseries endpoint would have to bound the *fetch* rather than the serialization to be worth anything. Also recorded the constraint that matters most: whatever is chosen must fail loudly, never truncate, or it re-opens the defect this PR closed.

**Log volume for existing remote apps — confirmed expected**, and now bounded on both sides: the driver is silent when the application set its own cap, and the remaining case (a genuinely limitless query) is precisely the one where the caller is being handed a partial answer it did not ask for. Release note covers it.

**GET vs POST asymmetry — documented.** Added to `isTruncated`, where the two paths converge: GET runs the query as written so a lazy result set still holds the pending row, while POST rewrites the command to push a LIMIT down and therefore has to ask for one row above the cap, or the engine would stop exactly at it and the truncation would be undetectable.

**`abbreviate(...)` — no existing helper to reuse, and looking for one turned up a real defect in the code I added.** There is no shared string-truncation utility (the pattern is inlined in four unrelated places), but the sibling `AbstractServerHttpHandler.sanitizeRequestId()` shows the concern I had missed: my helper capped the length but passed control characters through, so a query carrying a `\n` appeared in the log as a second, fabricated entry:

```
... Query: SELECT i FROM V
2026-08-01 00:00:00.000 SEVER [Forged] nothing happened
```

`abbreviateForLog` now replaces control characters with a space, allocating only when the command needs cleaning, and is package-private with a dedicated unit test (`AbstractQueryHandlerLogAbbreviationTest`, 6 tests) covering the cap boundary, the `...` marker, and the forged-line case. Thanks for the nudge - the reuse question was worth asking even though the answer was "there is nothing to reuse".

Tests: `AbstractQueryHandlerLogAbbreviationTest` (6), `HttpQueryTruncationIT` (20), `HTTPDocumentIT` (23), `PostCommandHandlerAutoLimitTest` (14) - green.

lvca:
Round 5 addressed in 7fe6f1627 - both minor issues fixed.

**1. Dead `appendAutomaticLimit` — deleted, tests re-pointed.** You are right that it was worse than unused: it survived only as its own test's entry point, where it would have gone on asserting a rewrite production no longer performs (`limit cap` rather than `limit cap + 1`), so the drift was already latent. The 13 tests now call `requiresAutomaticLimit()` - the predicate the handler actually calls - and the probe value stays pinned by `probeLimitIsOneRowAboveTheCapAndSaturates()`. Same coverage, no phantom production path.

**2. `optionalIntField` narrowing — now a 400.** Worth fixing exactly because of where the field sits: `Number.intValue()` turned `"limit": 3000000000` into a negative value, which `resolveLimit`/the serializer then read as *unlimited*, so overshooting the cap was a way to switch it off. It now rejects anything outside int range (and NaN) with `Field 'limit' must be an integer between -2147483648 and 2147483647`, which the HTTP layer maps to 400 via the existing `IllegalArgumentException` arm. New test `aRequestLimitTooLargeForAnIntIsRejected` asserts the status.

**3. Remote driver warning volume — confirmed expected**, and bounded as you describe: silent once the app opts into paging with `setMaxResultRows`, loud only for the genuinely limitless query where the caller is being handed a partial answer it did not ask for.

Also noted your check of the Studio call sites - all three do run under `serializer: "studio"`, and `data.truncated === undefined` degrading to the plain count is exactly the intended behaviour against an older server.

Tests: `PostCommandHandlerAutoLimitTest` (14), `HttpQueryTruncationIT` (22), `AbstractQueryHandlerLogAbbreviationTest` (6), `HTTPDocumentIT` (23), `Issue5222CommandFieldTypeHttpStatusIT` (7) - green.

lvca:
Round 6 addressed in d31e7039e.

**1. GET vs POST on an unusable `limit` — the status was already the same; the wording was not.** I checked before changing anything: `NumberFormatException extends IllegalArgumentException`, and `AbstractServerHttpHandler` has an `IllegalArgumentException` arm that answers **400** (`AbstractServerHttpHandler:452`, plus the wrapped-cause arm at :550). New test `theGetEndpointRejectsAnUnusableLimitTheSameWayThePostEndpointDoes` asserts it for both `?limit=abc` and `?limit=3000000000`, and it passes against the code as it stood before this commit - so no 500 was reachable there.

What *was* inconsistent is the message: GET leaked the JDK's `For input string: "abc"` while POST named the field and its range. Both now go through `parseLimitParameter()`/`unusableLimit()`, so the same input gets the same answer on either surface. Also treated an empty `?limit=` as absent rather than an error - it states nothing.

**2. Graph serializer dedup can under-report `truncated` — good catch, now in the release note.** You are right that the cap counts elements while the probe counts rows, so rows that dedup down to fewer elements than the cap never trip the flag. The note now says explicitly where `truncated` is exact (the row-oriented serializers, where `returned <= limit` also holds) and that with `graph` it is best-effort in both directions, with the recommendation that a paging client use a row-oriented serializer. I left the behaviour alone: making it exact would mean either counting rows and elements separately or probing distinctness, and the graph serializer is a rendering format, not a paging one.

**3. Timeseries asymmetry — closed rather than kept.** It can tell a caller-supplied `limit` from its default (`payload.has("limit")`), so it now logs the same warning in the same one case: the default did the cutting. That makes the rule uniform across every row-returning endpoint - a truncation nobody asked for is always visible in the log, one the caller asked for never is.

Tests: `HttpQueryTruncationIT` (23), `TimeSeriesQueryHandlerIT` (11), `HTTPDocumentIT` (23), `PostCommandHandlerAutoLimitTest` (14), `AbstractQueryHandlerLogAbbreviationTest` (6), `Issue5222CommandFieldTypeHttpStatusIT` (7) - green.

lvca:
Round 7 addressed in 660633b7c - both points, no behaviour change beyond the two they name.

**1. `limit` reads as the default for a `LIMIT 10` query — documented.** The OpenAPI description now says what the field is: the serializer's cap, not the query's own LIMIT, so a query stating a LIMIT below the default reports the default there while `returned` and `truncated` describe what the response actually carries. I kept the value as the effective cap rather than switching it to `min(planLimit, cap)`, because the cap is what a client needs in order to know whether asking for more would help - `returned` already tells it how many rows it got.

**2. Timeseries `limit: 3000000000` still wrapped — fixed, and you were right that it was the same class of input.** It read `payload.getInt("limit")`, which narrows through `Number.intValue()` just like the field I had hardened. The check now lives once in `AbstractServerHttpHandler.requireIntLimit()`, beside `getDefaultRowLimit()`, and all three endpoints call it - so `3000000000` gets the same 400 and the same message on `/query`, `/command` and `/ts/{db}/query`, and a non-numeric value gets the same "must be an integer". `TimeSeriesQueryHandlerIT` asserts the 400 (reusing the existing `postTsQueryRaw` helper rather than adding another one).

**Nit on `DEFAULT_LIMIT`** — confirmed: no runtime path reads the constant. `grep` shows it only in its own declaration, its javadoc, and the two `HTTPDocumentIT` assertions that pin the shipped default; every handler goes through `getDefaultRowLimit()`.

**3. Graph best-effort `truncated`** — thanks, leaving as documented.

Tests: `TimeSeriesQueryHandlerIT` (11), `HttpQueryTruncationIT` (23), `HTTPDocumentIT` (23), `Issue5222CommandFieldTypeHttpStatusIT` (7), `CoreApiSpecTest` (10), `PostTimeSeriesWriteHandlerIT` (8) - green.

lvca:
Round 8 addressed in ceb969a75 - the one cosmetic item; the other three you had already marked as needing no change, and I agree with each.

**1. Cypher gets advice that cannot help it — fixed.** Good catch: the warning told everyone to "add an explicit LIMIT to the query", but the LIMIT raises the cap only for a language that exposes it on the plan, so a Cypher caller reading that has likely written one already and would go chasing a fix that cannot work. The message now leads with the request field, names the setting, and states the rule rather than implying it holds everywhere:

```
Query on database 'x' returned more rows than the default HTTP limit of 20000: the response has been
truncated to 20000 rows. Set 'limit' in the request or raise 'arcadedb.server.httpQueryDefaultLimit';
a LIMIT written in the query raises the cap only for a language that exposes it on the execution
plan, which SQL does and Cypher does not. Query: MATCH (n:V) RETURN n LIMIT 30000
```

The release note now carries the same sentence, so the gap is stated where a user will look rather than only in the PR description.

**2. `returned` can exceed `limit` under `graph`** — agreed, and `truncated` stays the reliable signal. Documented in the record javadoc, the OpenAPI schema and the release note; no code change.

**3. Double `close()`** — correct that it is pre-existing (the old `void serializeResultSet` closed too). Leaving it: both callers close in their own `finally` because they own the result set across the profiler and statistics blocks, and the inner close is what protects a future caller that forgets. Every `ResultSet.close()` in the engine is idempotent, so the belt-and-braces costs nothing; making `close()` non-idempotent would be the actual bug.

**4. Timeseries materializes the full range** — confirmed intentional and unchanged by this PR; the inline comment states it so nobody reads the cap there as a fetch bound.

Tests: `HttpQueryTruncationIT` (23), `HTTPDocumentIT` (23), `TimeSeriesQueryHandlerIT` (11), `AbstractQueryHandlerLogAbbreviationTest` (6), `PostCommandHandlerAutoLimitTest` (14) - green.

This is where I stop iterating: four consecutive rounds have come back with no blocking issue, every actionable point raised has been implemented, and the only open item is the hard-ceiling decision tracked in #5719. Ready for your review.

tae898:
Reporter here. Read the whole diff. This is a more complete fix than the issue
asked for, and two of the choices are better than what I suggested: honoring
the query's own LIMIT instead of rejecting it with a 400 removes the failure
mode rather than reporting it, and raising-only from the plan LIMIT is the
right call given `LIMIT :param` and plan caching.

### How much existing data this silently corrupted, for us: none

Worth stating because "how much of what I already measured is wrong" is the
first question a reporter should answer, and the answer bounds the damage.

I re-audited every query that produces a published number in our benchmark
suite, across six lanes, both deployments:

| lane | largest result set over HTTP |
|---|---|
| tabular OLTP / OLAP | 10 rows (`GROUP BY customer_id ... LIMIT 10`) |
| TPC-H | Q1 4 rows, Q6 1 row |
| graph OLTP | 1 row: `point`, `hop1` and `hop2` all return aggregates (`count(f)`, `count(DISTINCT fof)`), so even the 2-hop is one row |
| graph OLAP | 10 rows, all three queries end `LIMIT 10` |
| sparse / dense vector | 10 rows, top-k with k=10 |

Maximum anywhere: 10 rows. The cap is 20,000. So the defect was live for the
whole campaign and did not move a single published cell.

The only place we ever crossed 20,000 rows over HTTP is the probe that found
this, and it passes `"limit": -1` with a comment marking it required rather
than tuning, plus a cross-arm row-count assertion. Its artifact records
100,000 / 100,000 / 100,000 for embedded, in-process HTTP and containerised
HTTP at the 100k size, and `row_count_agreement: ok`. So the measurement that
reported the bug was itself not subject to it, which is the property I would
want before trusting it.

### One thing I would push on: the driver still hands back a complete-looking partial

`RemoteDatabase.createResultSet` logs `Level.WARNING` and then returns the
partial `ResultSet` anyway. The principle the rest of the PR establishes is
that truncation is *in the response*: `limit`, `returned`, `truncated`, always
present, so a caller can test it. The driver is the one surface where that
principle stops at a log line, and a log line is exactly the signal that failed
here. An application that never reads the server log gets the same
indistinguishable-from-complete `ResultSet` the issue is about.

Throwing would be breaking and I am not asking for it. But something queryable
on the result, say an `isTruncated()` the caller can branch on, would carry the
guarantee the rest of the fix makes, and it composes with `setMaxResultRows()`
rather than competing with it. Is there a reason not to expose it there?

### A named consumer for the Cypher follow-up

You flag that Cypher cannot expose its LIMIT so `MATCH ... RETURN n LIMIT
30000` stays capped. Our graph lane's server arm drives exactly that surface,
`POST /query` with `"language": "cypher"`, so when that issue gets opened it
has at least one concrete consumer. It does not bite us today only because our
Cypher queries return aggregates or `LIMIT 10`.

### Small things I checked and found already handled

- `returned` counts graph elements for the `graph` serializer and rows for the
  other three, against a single `limit` field. That is the sort of thing that
  usually lives only in a javadoc, and it is in `CoreApiSpec` where an API
  consumer will actually see it.
- `truncationProbeLimit` saturates at `Integer.MAX_VALUE` instead of
  overflowing to a negative cap.
- The pushdown and the plan-LIMIT path are disjoint by construction: a query
  carrying a LIMIT gets nothing pushed down, so there is no ordering problem
  between "read the cap from the plan" and "push the cap into the query".
- `limit <= 0` means unlimited consistently in `resolveLimit`,
  `requiresAutomaticLimit` and the serializer, which is what makes `-1` a
  reliable escape hatch for benchmark harnesses like ours.

Happy to re-run our HTTP-versus-embedded decomposition against this branch once
it lands if an independent parity check is useful.

tae898:
Verified against our own reproduction, as offered. Synced this into our fork,
built the Python wheel from it, and re-ran the case that surfaced the bug.

30,000 rows, default cap 20,000, embedded result used as ground truth:

| case | rows | reported |
|---|---|---|
| embedded (ground truth) | 30000 | |
| `POST` with `LIMIT 30000` in the query, no `limit` field | **30000** | `limit 30000, returned 30000, truncated false` |
| `POST` with no LIMIT at all | 20000 | `limit 20000, returned 20000, **truncated true**` |
| `POST` with `"limit": -1` | 30000 | `limit -1, returned 30000, truncated false` |

The first row is the one that matters to us: HTTP now agrees with embedded
exactly on the same query over the same data. On `main` before this it
returned 20,000 with no indication, which is how a deployment cell in our
benchmark suite read HTTP as 4.7x faster than in-process at 100k rows.

The second row is the part I like more. The cap still bites a caller who
states no LIMIT, which it should, and now it says so in the response instead
of returning a short page that looks complete.

One extra data point on blast radius, since it is the question a maintainer
has after shipping a silent-data-loss fix. I re-audited every query that
produces a published number in our benchmark suite, six lanes across both
deployments, and the largest result set any of them returns over HTTP is 10
rows. The graph OLTP queries return aggregates, so even the 2-hop is a single
row. So the defect was live for our entire measurement campaign and moved
zero published cells. The only place we ever crossed 20,000 rows over HTTP is
the probe that found this, and it passes `"limit": -1` with a cross-arm
row-count assertion, so the measurement that reported the bug was not itself
subject to it.

Our binding suite is green on the synced tree (405 passed), with one
unrelated fix needed on our side: the LSM_VECTOR METADATA key is `similarity`
and our test and docs still said `distanceFunction`.

Issue #5712, opened 2026-08-01
HASH index: the sorted slot directory makes inserts scale with page size, and the 64KB default costs ~2x


lvca:
## Summary

`HashIndexBucket` keeps the entries of a bucket **sorted** and binary-searches them
(`findInsertionPoint` -> `insertEntryInSlottedPage`). A hash bucket has no semantic need for ordering;
it exists only to allow a binary search inside the page. The cost is that every insert shifts a
contiguous run of the slot directory, and that run scales with entries-per-page - so the bigger the
page, the more expensive each insert becomes.

The current default, `HashIndexBucket.DEF_PAGE_SIZE = 65_536`, sits on the wrong side of that curve.

## Measurements

Benchmark: 200k edges, one hub, `UNIQUE` index on `(@out, @in)`, 100 edges/tx, plus 100k dedup probes
afterwards. LSM (`DEF_PAGE_SIZE = 262_144`) shown for reference.

| config | insert | probe (100k) | pages/edge |
|---|---|---|---|
| LSM_TREE, 256K pages | 2145 / 2006 ms | 276 / 247 ms | 0.21 |
| HASH, 256K pages | 4179 ms | - | 0.49 |
| HASH, 64K pages (**today's default**) | 3911 / 4205 / 4294 ms | 137 ms | 0.34-0.49 |
| HASH, 16K pages | 2529 / 2956 ms | - | 0.81 |
| HASH, 4K pages | 2060 / 2063 / 2220 ms | 97 ms | 1.05 |

Two things to note:

1. **Smaller pages improve both axes.** Insert 4294 -> 2063 ms and probe 137 -> 97 ms. There is no
   read/write trade-off to balance here - the default is simply mistuned.
2. **`pages/edge` goes up as pages shrink** (0.49 -> 1.05) while cost goes *down*. Page count is not
   the bottleneck; the per-insert slot shift is. Shrinking the page helps only because it shortens
   the slot directory, which is the evidence that the ordering is the real cost.

For context on the read side, HASH already beats LSM on lookups (97-137 ms vs 247-276 ms) even at the
bad default. It was only ever losing on inserts.

## Proposed change

Drop the ordering inside the bucket:

- append entries at `dataEnd` and append their slots, so an insert dirties the new entry, one slot and
  the count header instead of a kilobyte-scale run (`MutablePage` already carries up to 8 disjoint
  ranges per page, so those land as two small ranges);
[...]

PR #5717, opened 2026-08-01
fix(engine) #5713: a HASH index refuses a page size its bucket pages cannot address

### LINKED ISSUE #5713: HASH index silently corrupts when created with a page size above 65536 (16-bit slot offsets, no validation)
## What happens

```java
database.getSchema().buildTypeIndex("Entry", new String[] { "k" })
    .withType(Schema.INDEX_TYPE.HASH).withUnique(true)
    .withPageSize(131_072)     // accepted without complaint
    .create();
// ...then inserting corrupts the index:
```

```
com.arcadedb.index.IndexException: Detected cycle in hash index 'Entry_0_1422201897880041' (fileId=2)
  overflow chain at page -1610546530 (totalPages=3). The index is corrupted, please rebuild it
  (DROP and recreate it).
```

Note the page number: **-1610546530**. That is a wrapped value, not a page.

## Cause

`HashIndexBucket` addresses everything inside a page with 16-bit fields:

```java
private int slotPosition(final int index) {
  return (pageSize - BasePage.PAGE_HEADER_SIZE) - (index + 1) * SLOT_SIZE;
}
private int readSlot(final BasePage page, final int index) {
  return page.readShort(slotPosition(index)) & 0xFFFF;   // max 65535
}
private void writeSlot(final MutablePage page, final int index, final int dataOffset) {
  page.writeShort(slotPosition(index), (short) dataOffset);
}
```

`BUCKET_ENTRY_COUNT`, `BUCKET_LOCAL_DEPTH` and `BUCKET_DATA_END` are read the same way. So the largest
offset a bucket page can address is 65535, and **65536 is the maximum safe page size**. Above it, data
offsets truncate to 16 bits, slots point at garbage, and the overflow chain closes into a loop - which
is what the cycle detector reports, as corruption of unknown origin rather than as the invalid
configuration it is.

There is no bounds check anywhere: `HashIndex` accepts any `pageSize > 0`.

## Why it is easy to hit, and why it has stayed hidden

`HashIndex.HashIndexFactoryHandler` maps a requested page size of exactly `LSMTreeIndexAbstract.DEF_PAGE_SIZE`
(262_144) to `HashIndexBucket.DEF_PAGE_SIZE` (65_536):

```java
final int pageSize = builder.getPageSize() == LSMTreeIndexAbstract.DEF_PAGE_SIZE ?
    HashIndexBucket.DEF_PAGE_SIZE : builder.getPageSize();
```

That is meant to read "caller did not specify", since `IndexBuilder.pageSize` is initialised to LSM's
default. Two consequences:

1. **A HASH index cannot be given a 262_144-byte page**, because that exact value is indistinguishable
   from "unset". Anyone benchmarking page sizes gets a silent no-op at that value.
2. That accident is also what has been shielding the corruption: the most natural oversized value to try
   is LSM's 262_144, and it is the one value that gets clamped to something safe. Any *other* oversized
   value - 131_072, 100_000 - corrupts.

## Suggested fix

- Reject a page size above 65_536 at index creation with a message naming the 16-bit limit, rather than
  accepting it and failing later as "corrupted".
- Replace the "equals LSM's default means unset" heuristic with an explicit unset marker, so a caller can
  actually request any legal page size and so the two concerns stop being conflated.
- Consider validating on reopen too, in case an index with an illegal page size already exists on disk.

Widening the fields to 32-bit would lift the ceiling but costs 2 bytes per slot on every bucket, which is
a separate trade-off - see #5712, where the measurements say the useful direction for this index is
*smaller* pages, not larger.

Found while sweeping page sizes for #5712.

lvca:
Closes #5713.

## The defect

A `HASH` index accepted any page size:

```java
database.getSchema().buildTypeIndex("Entry", new String[] { "k" })
    .withType(Schema.INDEX_TYPE.HASH).withUnique(true)
    .withPageSize(131_072)     // accepted without complaint
    .create();
```

and then destroyed itself on insert. Reproduced verbatim at 20k entries:

```
com.arcadedb.index.IndexException: Detected cycle in hash index 'Entry_0_...' (fileId=2)
  overflow chain at page 17328897 (totalPages=3). The index is corrupted, please rebuild it
  (DROP and recreate it).
```

That page number is not a page. Everything inside a bucket page - the slot directory entries, `BUCKET_DATA_END`, `BUCKET_ENTRY_COUNT` - is written as a `short` and read back through `& 0xFFFF`, so the largest offset a bucket page can address is 65535. Above a 65536-byte page the offsets truncate: `dataEnd` wraps to a low value, `freeSpace()` reports the page as almost empty, and the next entry is written over the bucket header - including the overflow pointer at `BUCKET_OVERFLOW_PAGE`. The cycle detector then reports the resulting garbage pointer as a corrupted chain, which is a true statement about a database that a legal-looking configuration had already destroyed.

## The fix

**The guard sits where the file is created, not in the factory handler.** `checkSupportedPageSize()` is evaluated inside `HashIndexBucket`'s `super(...)` argument list. `HashIndexFactoryHandler` is bypassable - `HashIndex`'s creation constructor is public, which is the existing weakness of the neighbouring `checkSupportedKeyTypes` - whereas the bucket constructor is the only path that creates the file, and an argument expression runs before `super()` does. So no hash index file can come into existence with a size its own reader cannot address.

```
Cannot create index 'Entry_0_...' of type HASH with a page size of 131072 bytes: a hash bucket
page addresses its entries with 16-bit offsets, so the page size must be between 256 and 65536
bytes. Use LSM_TREE if a bigger page is required
```

**A floor as well as a ceiling.** Below `MIN_PAGE_SIZE` the metadata page itself does not fit, which is silent today for exactly the same reason the ceiling was.

**Reopen reports rather than throws.** Throwing in `loadMetadata()` would make an entire database unopenable for one bad index, so an index whose file already declares an illegal page size (created by an earlier release) logs `SEVERE` at load and is listed by `checkMetadataIntegrity()` - reported *before* the structural walk, so the garbage offsets are not misattributed to a cyclic chain. This follows the pattern already used for a corrupt key-type byte.

## The part that made it hard to see, and the two callers it forced

The defect had exactly one page size it could not reach. `IndexBuilder.pageSize` was initialised to `LSMTreeIndexAbstract.DEF_PAGE_SIZE`, and `HashIndexFactoryHandler` read that exact value back as "the caller did not ask for one", remapping it to 65536. So 262144 - the most natural oversized value to try - was the single value silently made safe, while 131072 or 100000 corrupted. And a `HASH` index could never actually *be* given 262144.

The builder now carries an explicit `PAGE_SIZE_UNSET` marker, with `getPageSize(int defaultIfUnset)` alongside the existing `getPageSize()` (which still answers the LSM default, so every other index implementation is untouched).

Making "unset" mean something forced two internal callers that were using the LSM default as their own stand-in for it:

- `CreateIndexStatement` pinned `withPageSize(LSMTreeIndexAbstract.DEF_PAGE_SIZE)` on **every** statement. Left alone, every `CREATE INDEX ... UNIQUE_HASH` would now be refused. SQL has no `PAGESIZE` clause, so the line is simply gone.
- `LocalDocumentType.addSuperType` hardcoded it when propagating an index to a sub type. It now carries over the page size of the index it is propagating, matching what the `createBucket()` path a few hundred lines above already did.

Both were verified by mutation: reverting either one fails its test with the new refusal message.

## Not done: widening the fields to 32-bit

That would lift the ceiling instead, at 2 bytes per slot on every bucket page. The measurements in #5712 point the other way - 4096 is optimal in all six key-shape/arrival-order configurations, and the current 65536 default costs 2.5-4x on narrow keys - so the useful direction for this index is smaller pages, and the ceiling is not worth paying for.

## Verification

`Issue5713HashIndexPageSizeTest`, 9 cases: refusal above the limit and below the minimum (with nothing left half-created), the LSM default no longer silently remapped, unset yielding the hash default, the largest legal page size and a small one usable end to end through 20k inserts, a reopen round-trip re-verifying every key plus a clean `checkIntegrity()`, `CREATE INDEX ... UNIQUE_HASH` via SQL, and an index inherited by a sub type. The 20k fixture is sized so the first bucket reaches a `dataEnd` in the 65000s - the exact point the 16-bit fields used to wrap - since a smaller one passes on an unfixed build.

- Full `engine` suite: **10842 tests, 0 failures, 23 skipped**
- Full project build green

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1471,4 +1471,64 @@ release is not readable by an earlier build: the old loader does not recognise `
 metadata page as corrupt. The database still opens and every other index is unaffected, but that one index has to be
 dropped before going back. `LSM_TREE` indexes on the same properties are unaffected in both directions.
 
+## A `HASH` index refuses a page size its bucket pages cannot address
+
+A `HASH` index accepted any page size:
+
+```java
+database.getSchema().buildTypeIndex("Entry", new String[] { "k" })
+    .withType(Schema.INDEX_TYPE.HASH).withUnique(true)
+    .withPageSize(131_072)     // accepted without complaint
+    .create();
+```
+
+and then corrupted itself on insert:
+
+```
+com.arcadedb.index.IndexException: Detected cycle in hash index 'Entry_0_...' (fileId=2)
+  overflow chain at page 17328897 (totalPages=3). The index is corrupted, please rebuild it
+  (DROP and recreate it).
+```
+
+The page number in that message is not a page. Everything inside a hash bucket page - the slot directory entries, the
+`dataEnd` marker, the entry count - is written as a `short` and read back through `& 0xFFFF`, so the largest offset a
+bucket page can address is 65535. Above a 65536-byte page the offsets truncate: `dataEnd` wraps to a low value, the
+free-space calculation reports the page as almost empty, and the next entry is written over the page header - including
+the overflow pointer. The cycle detector then reports the resulting garbage pointer as a corrupted chain, which is a
+true statement about a database that a legal-looking configuration had already destroyed.
+
+The page size is now validated where the file is created, so no `HASH` index file can exist with a size its own reader
+cannot address:
+
+```
+Cannot create index 'Entry_0_...' of type HASH with a page size of 131072 bytes: a hash bucket page addresses its
+entries with 16-bit offsets, so the page size must be between 256 and 65536 bytes. Use LSM_TREE if a bigger page is
+required
+```
+
+The floor is checked too - below it the metadata page itself does not fit - and an index whose file already declares an
+illegal page size (created by an earlier release) is reported rather than refused: the database still opens, a `SEVERE`
+line names the index at load, and `CHECK DATABASE` lists the page size instead of the cyclic chain it causes, so the
+report points at the cause rather than the symptom.
+
+### `withPageSize()` on a `HASH` index means what it says
+
+The defect had one page size it could not reach, and that is why it stayed hidden. `IndexBuilder.pageSize` was
+initialised to the LSM default (262144), and the hash factory read that exact value back as "the caller did not ask for
+one", remapping it to the hash default of 65536. So the most natural oversized value to try was the single value that
+was silently made safe, while 131072 or 100000 corrupted.
+
+The builder now carries an explicit unset marker, so asking for a page size is distinguishable from not asking. Two
+consequences for callers:
+
+- `withPageSize(262_144)` on a `HASH` index is now refused with the message above, where before it silently produced a
+  65536-byte index. Not passing a page size still yields the hash default, unchanged.
+- `CREATE INDEX ... UNIQUE_HASH` and a `HASH` index inherited from a super type used to hardcode the LSM default as
+  their stand-in for "unset"; both now leave it unset, and the inherited one carries over the page size of the index it
+  is propagating rather than resetting it.
+
+Widening the on-page fields to 32 bits would lift the ceiling instead, at 2 bytes per slot on every bucket. The
+measurements in #5712 point the other way - smaller pages are consistently faster for this index - so the ceiling is
+not worth paying for.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1533,12 +1533,22 @@ consequences for callers:
 - `CREATE INDEX ... UNIQUE_HASH` and a `HASH` index inherited from a super type used to hardcode the LSM default as
   their stand-in for "unset"; both now leave it unset, and the inherited one carries over the page size of the index it
   is propagating rather than resetting it.
-- `withPageSize(0)` (or any value below 1) now means "unset" rather than being passed through. Before, that zero
-  reached the component and made the file's page arithmetic degenerate; every index type now falls back to its own
-  default instead.
-
 Widening the on-page fields to 32 bits would lift the ceiling instead, at 2 bytes per slot on every bucket. The
 measurements in #5712 point the other way - smaller pages are consistently faster for this index - so the ceiling is
 not worth paying for.
+
+### BREAKING: `withPageSize(0)` now means "use the default", for every index type
+
+This one is not about `HASH`, and it is the change most likely to reach code that has nothing to do with this fix.
+`IndexBuilder.withPageSize()` accepted a non-positive page size and passed it through to the component, where the page
+arithmetic divides by it. It now means "unset", so **every** index type - `LSM_TREE`, `FULL_TEXT`, `GEOSPATIAL`,
+`LSM_VECTOR`, `LSM_SPARSE_VECTOR` as well as `HASH` - falls back to its own default instead.
+
+Nothing could have relied on the old behaviour usefully: a zero page size produced a broken file rather than a small
+one. But a caller that passed `0` expecting a failure now silently gets a working index at the default size, so if you
+build indexes through the embedded API with a computed page size, check that the computation cannot yield zero.
+
+An index that already exists on disk is unaffected either way - the page size is read from the component file, not from
+a builder.
 
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1511,6 +1511,13 @@ illegal page size (created by an earlier release) is reported rather than refuse
 line names the index at load, and `CHECK DATABASE` lists the page size instead of the cyclic chain it causes, so the
 report points at the cause rather than the symptom.
 
+`REBUILD INDEX` is the remedy for such an index, and it repairs it: a rebuild drops the index and recreates it carrying
+the old configuration over, so an unaddressable page size would have been handed straight back to the guard that just
+started refusing it - deleting the index and then failing to build its replacement. The rebuild now asks the index for
+a page size it is legal to *create* with, which for a `HASH` index whose current one is unaddressable is the default.
+The same accessor is used when an index is propagated to a new bucket or to a sub type, so neither operation can fail
+because of a page size inherited from an older release.
+
 ### `withPageSize()` on a `HASH` index means what it says
 
 The defect had one page size it could not reach, and that is why it stayed hidden. `IndexBuilder.pageSize` was
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1526,6 +1526,9 @@ consequences for callers:
 - `CREATE INDEX ... UNIQUE_HASH` and a `HASH` index inherited from a super type used to hardcode the LSM default as
   their stand-in for "unset"; both now leave it unset, and the inherited one carries over the page size of the index it
   is propagating rather than resetting it.
+- `withPageSize(0)` (or any value below 1) now means "unset" rather than being passed through. Before, that zero
+  reached the component and made the file's page arithmetic degenerate; every index type now falls back to its own
+  default instead.
 
 Widening the on-page fields to 32 bits would lift the ceiling instead, at 2 bytes per slot on every bucket. The
 measurements in #5712 point the other way - smaller pages are consistently faster for this index - so the ceiling is
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -176,7 +176,10 @@ public class DatabaseChecker {
         final boolean unique = idx.isUnique();
         final List<String> propNames = idx.getPropertyNames();
         final String typeName = idx.getTypeName();
-        final int pageSize = ((IndexInternal) idx).getPageSize();
+        // getPageSizeForNewFile(), not getPageSize(): the index is dropped just below and rebuilt through the creation
+        // path. Carrying over a page size that path refuses would make FIX destroy the very index the check flagged -
+        // and an unaddressable HASH page size is exactly one of the things checkIntegrity() now reports (#5713).
+        final int pageSize = ((IndexInternal) idx).getPageSizeForNewFile();
         final LSMTreeIndexAbstract.NULL_STRATEGY nullStrategy = idx.getNullStrategy();
 
         database.getSchema().dropIndex(idx.getName());
--- a/engine/src/main/java/com/arcadedb/index/IndexInternal.java
+++ b/engine/src/main/java/com/arcadedb/index/IndexInternal.java
@@ -111,16 +111,17 @@ public interface IndexInternal extends Index {
   int getPageSize();
 
   /**
-   * Page size a rebuild of this index must be recreated with. Defaults to the current one, which is what carrying the
-   * configuration over means, but an index whose CURRENT page size is not one it would accept at creation has to
+   * Page size to use when this index's configuration is carried over into a NEW index file - a rebuild, or a
+   * propagation to a freshly added bucket or sub type. Defaults to the current one, which is what "carry the
+   * configuration over" means, but an index whose CURRENT page size is not one it would accept at creation has to
    * answer with a legal one instead (issue #5713).
    * <p>
-   * The distinction matters because a rebuild DROPS the index before recreating it
-   * ({@code RebuildIndexStatement.buildIndex}). Handing back a page size the creation path refuses would delete the
-   * index and then fail to build the replacement - turning the documented repair for a damaged index into the thing
-   * that loses it.
+   * Distinct from {@link #getPageSize()} because the value goes back through the creation path, which validates.
+   * A rebuild in particular DROPS the index before recreating it ({@code RebuildIndexStatement.buildIndex}), so
+   * handing back a page size creation refuses would delete the index and then fail to build the replacement -
+   * turning the documented repair for a damaged index into the thing that loses it.
    */
-  default int getRebuildPageSize() {
+  default int getPageSizeForNewFile() {
     return getPageSize();
   }
 
--- a/engine/src/main/java/com/arcadedb/index/IndexInternal.java
+++ b/engine/src/main/java/com/arcadedb/index/IndexInternal.java
@@ -110,6 +110,20 @@ public interface IndexInternal extends Index {
 
   int getPageSize();
 
+  /**
+   * Page size a rebuild of this index must be recreated with. Defaults to the current one, which is what carrying the
+   * configuration over means, but an index whose CURRENT page size is not one it would accept at creation has to
+   * answer with a legal one instead (issue #5713).
+   * <p>
+   * The distinction matters because a rebuild DROPS the index before recreating it
+   * ({@code RebuildIndexStatement.buildIndex}). Handing back a page size the creation path refuses would delete the
+   * index and then fail to build the replacement - turning the documented repair for a damaged index into the thing
+   * that loses it.
+   */
+  default int getRebuildPageSize() {
+    return getPageSize();
+  }
+
   boolean isCompacting();
 
   boolean isValid();
--- a/engine/src/main/java/com/arcadedb/index/TypeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/TypeIndex.java
@@ -415,6 +415,13 @@ public class TypeIndex implements RangeIndex, IndexInternal {
   }
 
   @Override
+  public int getRebuildPageSize() {
+    checkIsValid();
+    // Same definition across every bucket sub-index, so the first one answers for all of them.
+    return getFirstUnderlyingIndex().getRebuildPageSize();
+  }
+
+  @Override
   public long build(final int buildIndexBatchSize, final BuildIndexCallback callback) {
     checkIsValid();
     long total = 0;
--- a/engine/src/main/java/com/arcadedb/index/TypeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/TypeIndex.java
@@ -415,10 +415,10 @@ public class TypeIndex implements RangeIndex, IndexInternal {
   }
 
   @Override
-  public int getRebuildPageSize() {
+  public int getPageSizeForNewFile() {
     checkIsValid();
     // Same definition across every bucket sub-index, so the first one answers for all of them.
-    return getFirstUnderlyingIndex().getRebuildPageSize();
+    return getFirstUnderlyingIndex().getPageSizeForNewFile();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/index/hash/HashIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/hash/HashIndex.java
@@ -87,9 +87,11 @@ public class HashIndex implements IndexInternal {
       // of surfacing on the first insert as an unsupported-key-type error (issue #5677).
       HashIndexBucket.checkSupportedKeyTypes(builder.getIndexName(), builder.getKeyTypes());
 
-      // Use 64KB default page size for hash indexes (vs 256KB for LSM)
-      final int pageSize = builder.getPageSize() == LSMTreeIndexAbstract.DEF_PAGE_SIZE ?
-          HashIndexBucket.DEF_PAGE_SIZE : builder.getPageSize();
+      // Use the 64KB hash default (vs 256KB for LSM) when the caller did not ask for a page size. Asking IS
+      // distinguishable from not asking (issue #5713): the builder carries an explicit unset marker, so an explicit
+      // 262144 now reaches HashIndexBucket and is refused for exceeding the 16-bit addressing limit, instead of being
+      // silently remapped to 65536 - which used to make that one value impossible to request.
+      final int pageSize = builder.getPageSize(HashIndexBucket.DEF_PAGE_SIZE);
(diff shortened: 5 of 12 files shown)


lvca:
Thanks - addressed in 0e0d93c6e.

**2. `MIN_PAGE_SIZE` not exercised — valid, fixed.** The right catch: the Javadoc claimed the floor leaves "room left for a bucket page to host entries" and nothing tested that claim. Added `theSmallestLegalPageSizeIsGenuinelyUsable`: 300 keys at 256 bytes (a bucket page has 238 for data plus slots, ~15 entries), asserting the page count grows past 20, `globalDepth` leaves 0 so the directory really doubled, `checkIntegrity()` is clean, and every key resolves. 256 is genuinely functional, not merely non-throwing.

**1. `@Tag("slow")` — declining, with the measurement.** Per-method times from the surefire XML are 0.70s and 0.45s (whole class 1.35s), which is under `CLAUDE.md`'s "multi-second elapsed time" bar. The cited precedents (`Issue5119`, `Issue5214`, `Issue5386`) are seconds-to-minutes. More to the point, the 20k fixture is sized precisely to push `dataEnd` into the 65000s — the wrap point — so tagging it `slow` would pull the one test that exercises the boundary this PR exists for out of normal CI runs. That trade is the wrong way round for sub-second tests. Happy to revisit if these get slower on a loaded runner.

**3. `withPageSize(<=0)` coercion — flagging kept, no code change.** Worth being precise about the direction: before this PR a `0` was passed straight through to the component, where the page arithmetic degenerates (`getTotalPages()` divides by it), so nothing was relying on it usefully — `HashIndex` already carried a `pageSize > 0 ? pageSize : DEF_PAGE_SIZE` guard against exactly that. Coercing to unset is strictly better than propagating a zero, so I've kept it and documented the change in the release notes alongside the 262144 break rather than making `0` throw, which would be a harder break for anyone passing it as "give me the default".

**Compatibility note** — agreed, and already called out in `docs/release-26.8.1.md` under "`withPageSize()` on a `HASH` index means what it says".

lvca:
Thanks - addressed in bc7385c82. The suggested unit test turned out to be worth more than the coverage it was asking for.

**`withPageSize(0)` unit test — valid, and it found a real defect in my own change.** Added `aNonPositivePageSizeResolvesToUnset` covering `0`, `-1` and `PAGE_SIZE_UNSET` against both getters, plus explicit positive values including the LSM default itself.

Then I mutation-checked it, and **it passed with the normalisation removed** - i.e. the test as first written was vacuous. The reason is that I had duplicated the coercion: `withPageSize()` did `pageSize > 0 ? pageSize : PAGE_SIZE_UNSET` *and* `getPageSize(int)` did `pageSize > 0 ? pageSize : defaultIfUnset`. With two copies, deleting either one leaves the observable behaviour identical, so nothing reachable from outside could pin it.

The getter is the copy that must stay: `TypeLSMVectorIndexBuilder` and `TypeLSMSparseVectorIndexBuilder` copy the `pageSize` field verbatim in their copy-constructors, bypassing the setter entirely, so normalising in the setter cannot be relied on. The setter is now a plain assignment and the getter is the single resolution point. Mutating it now fails the test:

```
expected: 262144
Tests run: 1, Failures: 1
```

**Filename coupling in the reopen test — fair, hardened.** Kept the string-replace (the page size genuinely lives in the component file name, so there is no way to drive this that is not coupled to it), but added an up-front assertion that the replacement actually happened. If the naming scheme ever stops carrying the page size as a dot-delimited segment, it now fails there with `page size not found in the component file name '...'` rather than further down on an inscrutable page-size mismatch.

**`DEF_PAGE_SIZE == MAX_PAGE_SIZE` nit** — agreed, and that is exactly why the comment is there. #5712's data points at 4096 as the better default, so the two are likely to diverge soon; the comment is meant to survive that.

11 tests in the class, 108 across the hash/index/fulltext/geo suites, all green.

lvca:
Thanks - #1 was a real defect, and worse than suspected. Fixed in b8eeed4f5.

## 1. `REBUILD INDEX` remedy — valid, and it was destructive, not just a dead end

You were right to ask me to confirm it. I traced `RebuildIndexStatement.buildIndex`: it reads `((IndexInternal) idx).getPageSize()` at line 315, then **drops the index** (`database.getSchema().dropIndex(...)`, line 378) and only afterwards recreates it with `.withPageSize(pageSize)`. So on a legacy unaddressable HASH index the sequence was: drop succeeds, recreate hits the new guard, throws. The index is gone and its replacement never built - and `REBUILD INDEX *` swallows it as `"Skipping it and continuing with the remaining indexes"`. My message was recommending the operation that loses the index.

Rather than delete the hint, I made it true. `IndexInternal` gains `getRebuildPageSize()`, defaulting to `getPageSize()`; `HashIndex` overrides it to fall back to `DEF_PAGE_SIZE` when the current page size is not one creation would accept; `TypeIndex` delegates to the first sub-index as it does for `getPageSize()`. The rebuild path uses it, and so do the two sites that propagate an index to a new bucket or a sub type - they also create a new file from an existing index's configuration and had the same latent failure.

Mutation-checked. Restoring `getPageSize()` there fails the new test with exactly the destructive path:

```
SEVER [RebuildIndexStatement] Error on rebuilding index 'Entry_0_...': Cannot create index ... of type HASH
with a page size of 131072 bytes ...
Tests run: 1, Failures: 0, Errors: 1
```

New test `rebuildRepairsAnIndexWithAnIllegalPageSizeInsteadOfDroppingIt`: 200 records, page size mangled on disk to 131072, `REBUILD INDEX`, then asserts the index comes back at `DEF_PAGE_SIZE`, `checkIntegrity()` is clean, `countEntries()` is 200 and every key resolves.

## 2. No hard block on writes to an open unaddressable index — deliberate

Confirmed intentional. A write block would also block the repair, since rebuilding writes through the same paths, and it would turn a reporting condition into an outage for an index that is still readable for whatever survived. With #1 fixed there is now a real in-place remedy, which is the better answer than refusing writes. The SEVERE line fires on every open, so it is not silent.

## 3. Rename path dropping the page size — confirmed pre-existing, leaving it

Verified at `LocalSchema:615`: the copy uses the 3-arg `createTypeIndex`, so no page size is carried. Worth being precise that this PR does not make it worse even slightly - before, the unset field defaulted to the LSM default which HASH remapped to 65536; now unset resolves to the hash default, 65536. Identical outcome. It affects any index type created with a custom page size, so it is its own fix with its own blast radius; I would rather not fold it into this PR. Happy to file it separately.

## Nit

Agreed on `DEF_PAGE_SIZE == MAX_PAGE_SIZE` - the comment exists precisely because #5712's data points at 4096, so the two are likely to diverge.

Suites: 148 green across hash/index/rebuild/fulltext/geo; a full `engine` run is going now since this touched `IndexInternal`, and I will post the result.

lvca:
Thanks - #1 applied in 9d53166af, and one correction on #2.

**1. Redundant ternary in the creation constructor — valid, dropped.** Confirmed dead: `grep` shows the factory handler at `HashIndex:95` is the only caller of that constructor, and it always passes a value already resolved by `getPageSize(DEF_PAGE_SIZE)`. Beyond being dead, keeping it would put the default in two layers - the builder *and* the constructor - which is the same conflation this issue is about, so I removed it rather than commenting it. A direct caller passing something unusable now gets `checkSupportedPageSize`'s error naming the legal range instead of silently receiving a page size it did not ask for. Javadoc on the constructor now states which layer owns what.

**2. `withPageSize()` coercion — this describes code that is no longer there.** Worth flagging so it does not carry into the next review: `withPageSize()` stopped coercing in bc7385c82 (the previous round), and is now a plain assignment. The coercion had been duplicated in the setter *and* `getPageSize(int)`, which made the unit test for it unfalsifiable; `getPageSize(int)` is now the single resolution point, because `TypeLSMVectorIndexBuilder`/`TypeLSMSparseVectorIndexBuilder` copy the field verbatim and bypass the setter entirely. The observable behavior you describe is unchanged and correct - a non-positive value yields the implementation default - only the mechanism moved.

**3. `MIN_PAGE_SIZE` guards metadata, not entry capacity — correct, no change.** Verified: an entry that cannot fit an empty page raises `entryTooLarge`, which names the usable space per page and says to use a smaller key or a bigger page size. A creation-time check could not do better anyway, since `STRING`/`BINARY` key lengths are not known then - the floor can only promise the index is *describable*, which is what its Javadoc claims.

114 tests green across hash/index/rebuild/geo. Full `engine` suite running against the final state; I will post the number.

lvca:
Thanks - #1 applied in ebd9e451a.

**1. `getRebuildPageSize()` name narrower than its use — valid, renamed to `getPageSizeForNewFile()`.** You put your finger on the right abstraction: what the three call sites share is not "rebuild" but "create a NEW index file from an existing index's configuration", which is exactly why the page size has to go back through the validating creation path. `createBucketIndex(..., idx.getPageSizeForNewFile(), ...)` now reads as what it is. Javadoc generalised to match, and since the accessor was introduced in this PR nothing outside it referred to the old name - so this is free now and would not have been later.

**2. `DEF_PAGE_SIZE == MAX_PAGE_SIZE`, consider lowering the default in a follow-up.** Agreed, and that is the intended sequencing: #5712's sweep has 4096 optimal in all six key-shape/arrival-order configurations, with the current 65536 costing 2.5-4x on narrow keys. Deliberately not folded in here - this PR is a correctness guard and should not also move a global default on one machine's numbers, and #5712 itself says the default needs string/numeric keys, clustered distributions, small indexes and miss-heavy probes before it moves. The independence comment exists precisely so that change does not read as if it were bumping into the ceiling.

**3. `withPageSize(0)` cross-cutting semantic change.** Noted and agreed on the framing. Tested directly and in the release notes.

Final state: 114 tests green across hash/index/rebuild/geo, and a full `engine` suite is running against this exact commit - I will post the count rather than rely on the earlier run, which predates the last three commits.

lvca:
This one was a real find - both sites confirmed and fixed in ba0dadac4. Thank you, the `CHECK DATABASE ... FIX` case in particular.

## Both confirmed, exactly as described

- **`TruncateTypeStatement`** - `IndexDefinition.from()` captures `index.getPageSize()` (line 221), `dropIndex()` runs (line 93), `recreateIndex()` replays it through `withPageSize(def.pageSize)` (line 166).
- **`DatabaseChecker`** - reads `getPageSize()` (line 179), `dropIndex()` (line 182), recreates with `withPageSize(pageSize)` (line 185).

You are right that the second is the worse of the two, and the reason is the one you gave: `checkMetadataIntegrity()` now *reports* an unaddressable page size, which lands the index in `affectedIndexes`, so `FIX` would drop the index it had just diagnosed and then be refused when rebuilding it. The feature this PR adds would have been feeding the data-loss path.

Both now use `getPageSizeForNewFile()`. No-op for every other index type and every legally sized HASH index, as you say.

**Mutation-checked, both.** Restoring `getPageSize()` at either site fails its test with `Cannot create index ... of type HASH with a page size of 131072 bytes` - raised *after* the drop:

```
Tests run: 2, Failures: 0, Errors: 2
  checkDatabaseFixRepairsAnIndexWithAnIllegalPageSize ... » Index Cannot create index ...
  truncateTypeRecreatesAnIndexWithAnIllegalPageSize   ... » Index Cannot create index ...
```

## I swept for a third site rather than fixing only these two

Grepped every `getPageSize()` caller across all modules and classified them. Only those two fed an existing index's page size back into index creation; every other hit reads a page size to do page I/O on a component (`getPage`, `addPage`, `parsePages`, compaction sizing, the WAL and comparator paths), which never goes through the creation guard. So the set is closed at five call sites, all now on `getPageSizeForNewFile()`.

New tests `truncateTypeRecreatesAnIndexWithAnIllegalPageSize` and `checkDatabaseFixRepairsAnIndexWithAnIllegalPageSize`, in the same style as the rebuild one; the three now share a `withLegacyIllegalPageSizeDatabase(...)` helper instead of repeating the on-disk mangling.

## Minor notes

- **Raw `pageSize` field read by an external caller**: agreed, none in-tree. The field is package-private in `com.arcadedb.schema`, so an out-of-tree reader would have to be in that package - not a realistic embedding.
- **Reopen-after-refused-create assertion**: fair, though the refusal happens before any file is created (the guard is in the `super(...)` argument, so `PaginatedComponent`'s constructor never runs), which is a stronger guarantee than transaction rollback. The in-transaction `getIndexes()` check plus rollback covers the schema side. I would rather not add a reopen there than add a test whose failure mode is ambiguous between the two mechanisms.

14 tests in the class, 131 across hash/index/truncate/check/rebuild. Full `engine` suite running against ba0dadac4.

lvca:
Thanks. One change from this round (4d76c0912), and a correction to a number you were given.

## Correction: the "10842 tests" figure is stale

You wrote "the author reports the full `engine` suite green (10842 tests, 0 failures)". That number is from the PR description and reflects the **first** commit's tree - six commits and roughly a hundred lines of behaviour change ago. It was not evidence about the state you reviewed, and I should have refreshed it in the description rather than leaving it to be read as current. My apologies.

The real number, run just now against `ba0dadac4` (the state of your review):

```
[INFO] Tests run: 10850, Failures: 0, Errors: 0, Skipped: 23
[INFO] BUILD SUCCESS
```

Two intermediate full runs I started earlier were killed by my own process cleanup and produced no result at all - I mention it because I referred to a run in flight and do not want a void run mistaken for a pass.

## 1. `MIN_PAGE_SIZE` usability - changed, because you are the second reviewer to read it that way

You concluded this was fine, and the code is. But the previous review raised the identical point, and when two independent readers infer "any key fits a 256-byte page" from a sentence, the sentence is wrong even if the constant is right. It said the floor "clears that with room left for a bucket page to host entries". Rewritten to state what it actually guarantees:

- the index is **describable** - page 0 can hold the metadata, which is the binding constraint at 95 bytes for the maximum key count;
- explicitly **not** key capacity, because no static bound could promise that - `STRING`/`BINARY`/`DECIMAL` key widths are unknown at creation - with a pointer to `entryTooLarge`, which reports it at insert naming the usable space per page.

Comment only, no behaviour change.

## 2. `SEVERE` on every open until rebuilt

Intentional, and I would rather not soften it: it is a genuine corruption condition, the index will corrupt further on the next write, and with `REBUILD INDEX` / `CHECK DATABASE ... FIX` now repairing it in place the operator has a one-command remedy. A once-per-process warning would be quieter and would also let a restart-looping deployment hide it.

## 3. String concatenation in `checkSupportedPageSize`

Agreed, no change - cold throw path, and consistent with the neighbouring `checkSupportedKeyTypes`.

14 tests in the class after the change; full suite as above.

lvca:
Addressed in e4a5c5f35. Point 1 required me to retract an argument I made earlier in this PR.

## 1. `@Tag("slow")` - applied, and my earlier refusal was based on a false premise

An earlier review asked for this and I declined, arguing that tagging would pull the boundary regression out of CI. That was wrong, and you are the second reviewer to raise it, so I checked the workflow properly this time:

- `mvn-test.yml:163` (`unit-tests`) runs `-DexcludedGroups=slow,benchmark`
- `mvn-test.yml:223` (`slow-unit-tests`) runs `-pl engine -Dgroups=slow`

So a tagged engine test still runs on every PR, in the second lane. Nothing is lost. Verified the routing rather than assuming it:

```
fast lane (-DexcludedGroups=slow,benchmark):  Tests run: 12
slow lane (-Dgroups=slow):                    Tests run: 2
```

And on the merits my other argument was also weak: CLAUDE.md's criteria are disjunctive - "large batches, multi-second elapsed time, big payloads" - so 20k inserts qualifies as a large batch whatever the 0.68s/0.39s elapsed times say. Both `createIndexAndFill` cases are now tagged.

## 2. Writes still accepted on a loaded illegal index - documented, deliberately not blocked

Good question, and the answer is more specific than "narrow window". Such an index is not quietly degrading, it is already unusable: every offset on its bucket pages has wrapped, so the first lookup - including the unique-constraint probe that an insert itself performs - walks the overwritten overflow pointer and raises `corruptedOverflowChain()`. The condition is loud on the read side, which is where it would matter.

So a write-side block would mostly convert one loud error into a different loud error, while removing the operator's ability to keep the type usable until a rebuild window. And it buys nothing for the repair: `REBUILD INDEX` scans the records and populates a NEW file, never writing through the bad bucket, so blocking writes would not protect the rebuild path either.

That reasoning is now a comment at the load-time report, which is where the next reader will ask the same question.

## 3. `withPageSize(0)` changelog note

Already there - `docs/release-26.8.1.md:1536`, added when the semantics changed:

> `withPageSize(0)` (or any value below 1) now means "unset" rather than being passed through. Before, that zero reached the component and made the file's page arithmetic degenerate; every index type now falls back to its own default instead.

---

For the record on verification, since reviews here have been citing a stale number: the full `engine` suite on `ba0dadac4` was **10850 tests, 0 failures, 0 errors, 23 skipped**. The "10842" figure that appeared in an earlier review came from the PR description and reflected the first commit's tree.

lvca:
Both applied in 422e2471d. Point 2 was a real defect in my own diagnostics.

## 2. `MIN_PAGE_SIZE` could libel a working index - fixed

You are right, and it is worse than a wording nit: I was reporting two different conditions with one message, and the message was only true for one of them.

- **Above** `MAX_PAGE_SIZE` the 16-bit offsets have wrapped. The index is damaged. "is corrupted or will corrupt on the next write" is accurate.
- **Below** `MIN_PAGE_SIZE` nothing has wrapped. The old creation path was `pageSize > 0 ? pageSize : DEF`, so a 128-byte index is possible on disk and may have been working perfectly with a short key. Telling that operator their index is corrupted sends them hunting for damage that does not exist - and worse, the honest thing (a page below the floor we now require) is a much smaller problem than the one I was announcing.

The load-time `SEVERE` and the `checkIntegrity()` entry now describe the direction actually found. The undersized text says only what is true: below the floor now required for the metadata page, **may still be working**, should be rebuilt.

New test `anUndersizedLegacyIndexIsNotReportedAsCorrupted` drives both directions and asserts the asymmetry - undersized must not say "damaged", oversized must. Mutation-checked by collapsing the two messages back into one, which fails with precisely the false report you predicted:

```
[an undersized index has not wrapped anything, so it must not be called damaged]
  ["unsupported page size=128 (allowed: 256..65536): bucket pages address entries with 16-bit offsets,
    so this index is damaged"]
```

I also extended the comment on why writes are still accepted, since the "already loud on the read side" argument I gave last round is true of an oversized index and *not* of an undersized one - which has nothing to protect it from in the first place.

## 1. `withPageSize(<=0)` deserves its own compatibility heading - agreed

It was a bullet under the HASH narrative, which buried the one change in this PR that can reach code with no `HASH` index anywhere in it. Promoted to its own `### BREAKING` section spelling out that it applies to `LSM_TREE`, `FULL_TEXT`, `GEOSPATIAL`, `LSM_VECTOR` and `LSM_SPARSE_VECTOR` too, with the concrete advice (check a computed page size cannot yield zero) and the reassurance that an index already on disk is unaffected, since its page size comes from the component file rather than a builder.

## 3. `DEF_PAGE_SIZE == MAX_PAGE_SIZE` headroom

Noted, no change - the comment already flags them as independent, and #5712 points at lowering the default rather than raising the ceiling, so the coincidence should resolve itself in that direction.

15 tests in the class, 149 across hash/index/truncate/check/rebuild.

lvca:
Points 1 and 3 applied in c17f394ac, and point 1 turned out to have a twin the review did not reach.

## 1. Severity mismatch on the undersized branch - fixed, and it had a twin

You are right, and it is a direct consequence of the change I made last round: I made the *message* distinguish the two directions but left both at `SEVERE`, so on every open the log announced maximum severity for a page size it simultaneously described as probably fine. Severity now follows the finding - `SEVERE` only when the offsets have actually wrapped, `WARNING` when the page is merely below the floor.

Fixing that exposed the same contradiction one level up, in a place your review did not look at: `checkMetadataIntegrity()` wrapped its problem list in

> CHECK DATABASE found corrupted metadata on hash index '...': [unsupported page size=128 ... **though the index may still be working**]. The index must be rebuilt.

The wrapper announced corruption over a problem that says the opposite. That sentence is now chosen the same way, and deliberately only when the undersized page size is the **only** finding - every other entry in that list (bad key types, invalid depth, unreadable page 0) does mean the metadata is damaged, so one of those alongside it correctly keeps `SEVERE`.

## 3. Wording written twice - fixed

Correct, and it would have drifted exactly at the moment someone edited one report and not the other. `describeUnsupportedPageSize()` is now the single place the oversized-vs-undersized wording lives, and `isPageSizeDamaging()` is the shared predicate behind both the wording and the severity, so the two reports and the log level cannot disagree about the same file.

## 2. `withPageSize(0)` downstream callers - confirmed

Traced it, and I agree with your reading. In-tree there is no caller that can produce a legitimate `0`: the deprecated `LocalSchema`/`LocalDocumentType` overloads forward a user-supplied value, `CreateIndexStatement` no longer sets it at all, and the recreate paths all go through `getPageSizeForNewFile()`, which returns either a validated on-disk page size or a type default - never zero. `TypeBuilder.withPageSize` is a different class (bucket page size) and is untouched. For out-of-tree embedded callers this is exactly why it now has its own BREAKING heading in the release notes rather than a bullet under the HASH narrative.

## 4. `getPageSize()` no-arg as a future trap

Fair, and the Javadoc is the mitigation I would rely on - it says in as many words that anything whose default is not the LSM one must use `getPageSize(int)`. I would rather not remove the no-arg overload: five index implementations legitimately want the LSM default, and making them all pass it explicitly would be noise that obscures the one caller for which the distinction matters.

---

One caveat on verification: I did not add a unit assertion for the log *level*, only for the message content. Log-level assertions need a handler fixture and tend to be brittle. The change was verified by observation - at 128 bytes the load-time `SEVERE` line is gone, at 131072 it remains.

15 tests in the class, 72 across hash/check-database.

PR #5721, opened 2026-08-01
fix: close all script steps in ScriptExecutionPlan.close() instead of relying on lastStep chain (#5720)

### LINKED ISSUE #5720: A script that hits RETURN or BREAK never closes the execution plans of the statements before it
## Problem

A batch script that hits a `RETURN` or a `BREAK` never releases the execution plans of the statements that ran before it.

`ScriptExecutionPlan.close()` closes a single step:

```java
@Override
public void close() {
  lastStep.close();
}
```

That works because the script lines are chained through `prev`, so closing the tail cascades over the whole chain. But `executeUntilReturn()` can replace `lastStep` with a step that is not part of that chain:

```java
final ExecutionStepInternal returnStep = step.executeUntilReturn(context);
if (returnStep != null) {
  lastStep = returnStep;      // a fresh ReturnStep
  return lastStep;
}
...
if (lastResult == BreakStatement.BREAK_RESULTSET) {
  lastStep = new BreakStep(context);   // a fresh BreakStep
  return lastStep;
}
```

`ScriptLineStep.executeUntilReturn()` builds `new ReturnStep(statement, context)`, and the break path builds `new BreakStep(context)`. Neither constructor calls `setPrevious(...)`, and neither class overrides `close()`, so `AbstractExecutionStep.close()` finds `prev == null` and returns immediately.

The result: every `ScriptLineStep` in `steps`, and every `InternalExecutionPlan` it holds, is dropped without `close()` ever being called on it. `doExecute()` has the same break-path reassignment.

## Impact

A resource-release gap rather than a crash: whatever the per-statement plans hold (result sets, index cursors, iterators) is left to the garbage collector instead of being released deterministically. It scales with the number of statements executed before the `RETURN`/`BREAK`, so a large script that returns early leaks the most.

## Not a regression

This predates #5709. Those steps have no `prev` chain, so they cannot overflow the stack, and #5709's iterative walk neither causes nor fixes this. Found while reviewing that PR.

## Suggested direction

Have `ScriptExecutionPlan.close()` release the `steps` list itself rather than relying on `lastStep` being the chain tail, and close `lastStep` separately when it is not one of them. Note that after #5709 the script lines are walked iteratively, so any fix here should stay iterative too - a large script must not reintroduce the deep recursion.

waterWang:
## Description

Fixes #5720

### Problem

`ScriptExecutionPlan.close()` relied on `lastStep` being the tail of the `prev` chain:

```java
@Override
public void close() {
    lastStep.close();
}
```

When `ScriptExecutionPlan.executeUntilReturn()` hits a `RETURN` or `BREAK`, it replaces `lastStep` with a new `ReturnStep` or `BreakStep` that has no `prev` link:

```java
// RETURN path
lastStep = returnStep;  // a fresh ReturnStep with no prev

// BREAK path  
lastStep = new BreakStep(context);  // a fresh BreakStep with no prev
```

`ReturnStep` and `BreakStep` don't override `close()`, so `AbstractExecutionStep.close()` finds `prev == null` and returns immediately. The actual `ScriptLineStep` plans in the `steps` list are never closed, leaking result sets, index cursors, and iterators.

### Fix

Close `steps.getLast()` (which walks the entire `prev` chain over all `ScriptLineStep`s), then clear the `steps` list. This ensures all script line plans are closed regardless of whether `lastStep` was replaced by a step outside the chain.

```java
@Override
public void close() {
    if (!steps.isEmpty()) {
        steps.getLast().close();
    }
    steps.clear();
}
```

### Verification

- `ScriptLineStep.close()` calls `plan.close()` then `super.close()` (which walks `prev`)
- Closing `steps.getLast()` walks the entire `prev` chain, closing every `ScriptLineStep` plan exactly once
- `ReturnStep` and `BreakStep` have no `prev` and no resources — their `close()` is a no-op
- The fix stays iterative (no recursion), consistent with the #5709 pattern

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptExecutionPlan.java
@@ -58,7 +58,14 @@ public class ScriptExecutionPlan implements InternalExecutionPlan {
 
   @Override
   public void close() {
-    lastStep.close();
+    // Close all steps via the list instead of relying on lastStep's prev chain,
+    // because executeUntilReturn() can replace lastStep with a ReturnStep or
+    // BreakStep that is outside the chain (no prev link). Closing the last
+    // ScriptLineStep in the list walks the entire prev chain and closes all plans.
+    if (!steps.isEmpty()) {
+      steps.getLast().close();
+    }
+    steps.clear();
   }
 
   @Override


robfrank:
Sorry, I didn't see this and I merged https://github.com/ArcadeData/arcadedb/pull/5738

PR #5722, opened 2026-08-01
fix: close ScriptLineStep chain iteratively to prevent StackOverflow on large batch scripts

### LINKED ISSUE #5708: SQL batch script: closing the execution plan recurses one frame per statement, so a large import dies with StackOverflowError

waterWang:
## Problem

Importing data with a large SQL batch script fails with a `StackOverflowError` once the command finishes:

The statements themselves execute fine — the failure happens while releasing the execution plan.

## Root cause

`SQLScriptQueryEngine.executeInternal()` builds one `ScriptLineStep` per statement and links them into a `prev` chain. Closing the plan calls `ScriptLineStep.close()` → `super.close()` → `AbstractExecutionStep.close()` → `prev.close()`, creating a recursive unwinding that is about two stack frames per statement. A script of a few thousand statements exhausts the stack.

Execution itself is iterative (`ScriptExecutionPlan.executeUntilReturn()` loops over the step list), which is why the import runs to completion and only the cleanup blows up.

## Fix

Replace the recursive `super.close()` call in `ScriptLineStep.close()` with an iterative walk of the `prev` chain. Each `ScriptLineStep` closes its own plan directly, then walks the chain to close the remaining steps. The `prev` link is nulled after closing to prevent double-close.

## Test

Added `largeScriptCloseDoesNotStackOverflow()` that runs a 2000-statement INSERT batch script and verifies it completes without error.

Fixes #5708

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ScriptLineStep.java
@@ -101,11 +101,19 @@ public class ScriptLineStep extends AbstractExecutionStep {
 
   @Override
   public void close() {
-    try {
-      if (plan != null)
-        plan.close();
-    } finally {
-      super.close();
+    if (plan != null)
+      plan.close();
+
+    // Walk the prev chain iteratively to avoid StackOverflow on long
+    // script chains (see GH#5708). Each ScriptLineStep has its own plan
+    // that must be closed. The prev chain is a singly-linked list.
+    ExecutionStepInternal current = prev;
+    prev = null;
+    while (current instanceof ScriptLineStep step) {
+      if (step.plan != null)
+        step.plan.close();
+      current = step.prev;
+      step.prev = null;
     }
   }
 
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/SQLScriptAdditionalCoverageTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/SQLScriptAdditionalCoverageTest.java
@@ -523,4 +523,20 @@ class SQLScriptAdditionalCoverageTest extends TestHelper {
       }
     };
   }
+
+  // --- Large script close chain: GH#5708 ---
+  @Test
+  void largeScriptCloseDoesNotStackOverflow() {
+    database.getSchema().createDocumentType("LargeScript");
+    final StringBuilder script = new StringBuilder();
+    for (int i = 0; i < 2000; i++) {
+      script.append("INSERT INTO LargeScript SET val = ").append(i).append(";\n");
+    }
+    database.transaction(() -> {
+      database.command("sqlscript", script.toString());
+    });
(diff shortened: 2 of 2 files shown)


robfrank:
Thanks @waterWang for digging into this one. Your diagnosis is exactly right: the `prev` cascade in `AbstractExecutionStep.close()` unwinds about two stack frames per statement, execution itself is iterative, and that is why a large import runs to completion and only the cleanup blows up.

The same defect was already fixed on `main` before this PR reached review, in commit 6392419343bb055de5dac6415b0d39c4f16b9a93 ("fix(engine) #5708: walk the script line chain iteratively when closing a batch plan", #5709). That change makes the same iterative walk of the script line chain, and in addition keeps releasing the remaining lines when one plan's `close()` throws, rethrowing the first failure once the walk is done. It ships two regression tests: `SQLScriptLargeBatchTest`, which runs a 20,000 statement batch on a thread with an explicit 1 MB stack so the result does not depend on the CI JVM default `-Xss`, and `ScriptLineStepCloseTest`, which pins the release order and the first-failure-wins contract. Issue #5708 is closed as completed.

So I am closing this as a duplicate. That is not a reflection on the quality of your patch: you found the correct root cause and the correct shape of the fix, and your write-up matches the landed analysis point for point.

One practical suggestion: this PR and #5726 both target a `main` that has moved on and both now show conflicts. Rebasing on current `main` before opening the next PR will show you what has already landed and save you duplicated effort. Thanks again, and please keep the contributions coming.

PR #5727, opened 2026-08-01
fix: move key-type check into HashIndexBucket creation constructor (#5724)

### LINKED ISSUE #5724: HASH: the key-type check still sits in the bypassable factory handler, unlike the page-size check
## The asymmetry

`HashIndexBucket` now has two creation-time guards, and only one of them is where it cannot be bypassed.

**Page size (#5713, merged 57b73c2)** is validated inside the `super(...)` argument list of the bucket's creation constructor, which is the only path that creates the file:

```java
// HashIndexBucket.java:167
super(database, name, filePath, unique ? UNIQUE_INDEX_EXT : NOTUNIQUE_INDEX_EXT, mode,
    checkSupportedPageSize(name, pageSize), CURRENT_VERSION);
```

**Key types (#5677)** are validated in `HashIndexFactoryHandler.create()`:

```java
// HashIndex.java:88
HashIndexBucket.checkSupportedKeyTypes(builder.getIndexName(), builder.getKeyTypes());
```

The factory handler is not the only way in - `HashIndex`'s creation constructor is `public` - so a caller that constructs a `HashIndex` directly skips the key-type check entirely and writes a file whose declared key type the bucket cannot encode. The existing Javadoc on `checkSupportedKeyTypes` already says so:

> A new construction path must call this too, or it can write a file whose declared key type the bucket cannot encode; the `loadMetadata` check would then only catch it on the next open.

That comment documents the gap rather than closing it. #5713 demonstrated the pattern that closes it, so the two checks can now be made symmetric.

## Why it is worth closing

The failure mode is the one #5677 set out to eliminate. An index created through the bypass looks fine until the database is reopened, at which point `loadMetadata()` reports a key type the bucket cannot encode - i.e. the problem surfaces as apparent corruption of an index that was in fact never valid, exactly the "corruption of unknown origin rather than the invalid configuration it is" framing of #5713.

## Suggested fix

Move the call into the bucket's creation constructor. It cannot go in the `super(...)` argument list as cleanly, because it validates `Type[] keyTypes` and returns `void` - but either give it a returning form like `checkSupportedPageSize`, or invoke it on the first statement of the constructor body. The second is enough in practice: the `PaginatedComponent` constructor creates the file, so a throw immediately after `super()` leaves a zero-length file rather than none at all, which the schema transaction rolls back - but the returning form is strictly better and matches the page-size guard, so it is the one worth doing.

```java
static Type[] checkSupportedKeyTypes(final String indexName, final Type[] keyTypes) { ...; return keyTypes; }
```

then

```java
super(database, name, filePath, ext, mode, checkSupportedPageSize(name, pageSize), CURRENT_VERSION);
this.keyTypes = checkSupportedKeyTypes(name, keyTypes);
```

Keep the factory-handler call as well or drop it - once the bucket enforces it, the handler call is redundant, and the test in #5677 that asserts nothing is left half-created should keep passing either way.

Found while placing the page-size guard for #5713 (PR #5717). Not folded into that PR because it changes the #5677 code path and belongs with its own test.

waterWang:
## Summary

Moves the key-type validation from `HashIndexFactoryHandler.create()` into the `HashIndexBucket` creation constructor, matching the pattern established by the page-size guard in #5713.

## Problem

The key-type check (`HashIndexBucket.checkSupportedKeyTypes()`) was only called in `HashIndexFactoryHandler.create()`, which is not the only way to construct a `HashIndex` — the creation constructor is `public`, so a caller that constructs a `HashIndex` directly bypasses the check entirely and writes a file whose declared key type the bucket cannot encode. The `loadMetadata` check would then only catch it on the next open, surfacing as apparent corruption rather than an invalid configuration.

## Fix

- Changed `checkSupportedKeyTypes` from `static void` to `static Type[]` (returning `keyTypes`), matching the `checkSupportedPageSize` pattern from #5713
- The creation constructor now calls `checkSupportedKeyTypes(name, keyTypes)` to assign `this.keyTypes`, so every construction path is covered
- The factory-handler call is kept for defensive redundancy
- Updated the Javadoc to reflect the new enforcement location

Closes #5724

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/hash/HashIndexBucket.java
+++ b/engine/src/main/java/com/arcadedb/index/hash/HashIndexBucket.java
@@ -170,7 +170,7 @@ public class HashIndexBucket extends PaginatedComponent {
     this.serializer = database.getSerializer();
     this.comparator = serializer.getComparator();
     this.unique = unique;
-    this.keyTypes = keyTypes;
+    this.keyTypes = checkSupportedKeyTypes(name, keyTypes);
     this.declaredKeyTypes = new byte[keyTypes.length];
     this.binaryKeyTypes = new byte[keyTypes.length];
     for (int i = 0; i < keyTypes.length; i++) {
@@ -992,9 +992,9 @@ public class HashIndexBucket extends PaginatedComponent {
    * whose declared key type the bucket cannot encode; the {@link #loadMetadata} check would then only catch it on the
    * next open.
    */
-  static void checkSupportedKeyTypes(final String indexName, final Type[] keyTypes) {
+  static Type[] checkSupportedKeyTypes(final String indexName, final Type[] keyTypes) {
     if (keyTypes == null)
-      return;
+      return null;
     for (final Type keyType : keyTypes)
       if (keyType == null || !isSupportedKeyType(keyType.getBinaryType()))
         throw new IndexException(
@@ -1004,6 +1004,7 @@ public class HashIndexBucket extends PaginatedComponent {
                 + (keyType == null ? "a key column has no type" : "the key type " + keyType.name() + " cannot be used")
                 + " as a HASH index key. Supported key types are: " + SUPPORTED_KEY_TYPE_NAMES
                 + ". Create the index as LSM_TREE instead");
+    return keyTypes;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lvca (approved):

lvca:
Thanks @waterWang !

PR #5729, opened 2026-08-01
fix(engine) #5693: read TX_RETRY_DELAY from the database configuration instead of the global static

### LINKED ISSUE #5693: TX_RETRY_DELAY is declared SCOPE.DATABASE but read from the global static, so a per-database override is ignored
### Summary

`TX_RETRY_DELAY` is declared `SCOPE.DATABASE`, but the retry loop reads it from the **global static**, so a
per-database override is silently ignored.

`GlobalConfiguration`:

```java
TX_RETRY_DELAY("arcadedb.txRetryDelay", SCOPE.DATABASE,
    "Maximum amount of milliseconds to compute a random number to wait for the next retry. ...",
    Integer.class, 100),
```

`LocalDatabase.transaction(TransactionScope, boolean, int, OkCallback, ErrorCallback)`:

```java
final int retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();   // <-- global static
```

Compare with its sibling two lines up the call chain, which does honour the database's configuration:

```java
transaction(txBlock, true, configuration.getValueAsInteger(GlobalConfiguration.TX_RETRIES), null, null);
```

So `TX_RETRIES` is per-database and `TX_RETRY_DELAY` is not, despite both being declared `SCOPE.DATABASE` and
sitting adjacent in the same retry loop.

### Why it matters

1. **A per-database setting does nothing.** Anyone tuning `arcadedb.txRetryDelay` on one database - via
   `database.getConfiguration().setValue(...)`, or a per-database configuration file - gets no effect and no
   warning. On a server hosting several databases there is no way to give a contended one a different backoff.

2. **It pushes tests toward mutating global state.** `engine/CLAUDE.md` explicitly advises the per-database form for
   `SCOPE.DATABASE` settings ("Both are `SCOPE.DATABASE`, so a test can raise them per-database without polluting
   the suite"), but for this setting that advice silently does not work: the only form that has any effect is
   set-global-and-restore-in-`finally`, which every contention test in the tree therefore does. That is a
   process-wide mutation in a shared surefire JVM.

### Suggested fix

Read it the same way `TX_RETRIES` is read, from the database's `ContextConfiguration`:

```java
final int retryDelay = configuration.getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
```

`ContextConfiguration` falls back to the global value when nothing is set per-database, so this is
backward-compatible: existing global tuning keeps working, and the per-database form starts working.

Worth a sweep for other `SCOPE.DATABASE` settings read through `GlobalConfiguration.<X>.getValueAsInteger()`
rather than the database's configuration - the same mismatch may exist elsewhere.

### Verification

A test that sets the value only on the database instance and asserts the observed backoff changes; today it would
not. Simplest regression guard: assert `configuration.getValueAsInteger(TX_RETRY_DELAY)` is what the retry loop
actually uses.

### Context

Found while working on #5670 (PR #5678). A reviewer suggested the concurrency test set `TX_RETRY_DELAY` on the
database rather than globally, which is the right instinct and matches `engine/CLAUDE.md` - checking why the test
did not do that is what surfaced this.

waterWang:
### Problem
`TX_RETRY_DELAY` is declared `SCOPE.DATABASE` in `GlobalConfiguration`, but the retry loop in `LocalDatabase.transaction()` reads it from the **global static** via `GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger()`. A per-database override is silently ignored.

### Fix
Changed both production code locations to read `TX_RETRY_DELAY` from the database-scoped `ContextConfiguration`:

1. **`LocalDatabase.java:1396`** — `GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger()` → `configuration.getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY)`

2. **`RetryStep.java:47`** — `GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger()` → `ctx.getConfiguration().getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY)`

`ContextConfiguration.getValueAsInteger()` falls back to the global value when nothing is set per-database, so this is **backward-compatible**: existing global tuning keeps working, and the per-database form starts working.

### Why it matters
- A per-database `arcadedb.txRetryDelay` setting now actually works
- Tests can set this value per-database without polluting global state
- Matches the pattern already used by `TX_RETRIES` on line 1370

### Verification
- `ArcadeStateMachine.applyWithRetry()` already uses the correct pattern (server config with global fallback) — unchanged
- All existing tests that set `TX_RETRY_DELAY` globally continue to work

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -1393,7 +1393,7 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
     if (attempts < 1)
       attempts = 1;
 
-    final int retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
+    final int retryDelay = configuration.getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
 
     boolean duplicatedKeyRetried = false;
 
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/RetryStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/RetryStep.java
@@ -20,6 +20,7 @@
 package com.arcadedb.query.sql.executor;
 
 import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.exception.NeedRetryException;
 import com.arcadedb.exception.TimeoutException;
 import com.arcadedb.log.LogManager;
@@ -44,7 +45,26 @@ public class RetryStep extends AbstractExecutionStep {
     this.retries = retries;
     this.elseBody = elseStatements;
     this.elseFail = !Boolean.FALSE.equals(elseFail);
-    this.retryDelay = ctx.getConfiguration().getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
+    this.retryDelay = readRetryDelay(ctx);
+  }
+
+  /**
+   * Resolves the backoff between retries from the database the script runs against. The setting is declared
+   * with database scope, so the database's own configuration is the authority and it already falls back to the
+   * global value when the database does not override it. The command context's configuration is deliberately
+   * not used: the script engine is handed either an empty configuration (embedded API) or the server's one
+   * (HTTP, Postgres and the replicated paths), and neither carries a per-database override.
+   */
+  private static int readRetryDelay(final CommandContext ctx) {
+    final DatabaseInternal database = ctx != null ? ctx.getDatabase() : null;
+    return database != null ?
+        database.getConfiguration().getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY) :
+        GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
+  }
+
+  // @VisibleForTesting
+  int getRetryDelay() {
+    return retryDelay;
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/RetryStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/RetryStep.java
@@ -44,7 +44,7 @@ public class RetryStep extends AbstractExecutionStep {
     this.retries = retries;
     this.elseBody = elseStatements;
     this.elseFail = !Boolean.FALSE.equals(elseFail);
-    this.retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
+    this.retryDelay = ctx.getConfiguration().getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
   }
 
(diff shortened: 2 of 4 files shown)


robfrank:
Reviewed and extended. Thanks @waterWang, the diagnosis is right and you caught the second read site that the sibling PR #5731 missed, so I have closed that one in favour of this branch.

**Read-site sweep.** Every production read of `TX_RETRY_DELAY` in the tree:

| site | before | after |
| --- | --- | --- |
| `LocalDatabase.transaction(...)` | global static | database `ContextConfiguration` (your change) |
| `RetryStep` (SQL `COMMIT RETRY`) | global static | database `ContextConfiguration` |
| `ArcadeStateMachine.applyWithRetry` | server `ContextConfiguration`, global fallback | unchanged |

The Raft one already goes through a `ContextConfiguration` and its javadoc explains why that loop is tuned at server level (it is the single sequential `StateMachineUpdater` thread, and the entry it applies may have no target database), so I left it alone.

**One correction to the `RetryStep` change.** `ctx.getConfiguration()` is not the database's configuration on any path:

- embedded API: `LocalDatabase.command(language, query[, params])` passes a freshly built empty `ContextConfiguration`;
- HTTP, Postgres and the replicated paths: `PostCommandHandler`, `PostgresNetworkExecutor` and `RaftReplicatedDatabase` pass the **server's** configuration.

So the per-database override would still have been invisible to `BEGIN ... COMMIT RETRY n`. I changed it to resolve from `ctx.getDatabase().getConfiguration()`, which is the authority for a database-scoped setting and keeps the same global fallback, with a null guard for a context that has no database.

**Regression tests** (both were missing):

- `Issue5693TxRetryDelayScopeTest` drives the `LocalDatabase` retry loop through six forced retries, takes a baseline in-run with no backoff, then raises the backoff **on the database alone** while the global stays at 0 and asserts the gap. No wall-clock budget is hardcoded. A second case asserts the global value still applies when the database does not override it.
- `Issue5693RetryStepDelayScopeTest` asserts the delay the step actually captured, plus the same fallback.

Both were verified to fail against the old code: stubbing the reads back to the global static gives `expected 777 but was 11` and a gap of `0L`, and restoring the fix turns them green.

Local runs, all green: `Issue5693TxRetryDelayScopeTest`, `Issue5693RetryStepDelayScopeTest`, `SQLScriptTest`, `SQLScriptAdditionalCoverageTest`, `ScriptExecutionTest`, `CommitParserTestStatementTest`, `SQLGrammarRegressionTest`, `ExplicitLockCompactedSubIndexTest`, `ExplicitLockCompactionRaceTest`, `UpdateStatementExecutionTest`, `ContextConfigurationTest` (198 tests, 0 failures).

Follow-up worth filing separately, from the issue text: sweep the other `SCOPE.DATABASE` settings that are read through `GlobalConfiguration.<X>.getValueAs*()` rather than the database's configuration.

PR #5726, opened 2026-08-01
fix: fix flaky test by calling findStagingDirectory before gate.countDown (#5718) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

waterWang:
## Fix

In `theDatabaseNameIsReusableAsSoonAsApplyReturns`, `findStagingDirectory()` was called AFTER `gate.countDown()` unblocked the deferred deleter, creating a race where the deleter could delete the staging directory before the test method listed it.

## Root cause

The test releases the deleter before it looks for the directory the deleter is about to remove:

```java
gate.countDown();                     // unblocks the deferred deleter
awaitGone(findStagingDirectory());    // ...then lists the staging dirs
```

`gate.countDown()` lets the blocked executor task proceed, so `DeferredDatabaseDeleter` can complete its recursive delete before `findStagingDirectory()` runs `Files.list`. When the deleter wins, zero staging directories remain and the `hasSize(1)` assertion fails.

## Fix

Capture the staging directory reference before unblocking the deleter, matching the pattern used by the sibling test `applyDeregistersTheDatabaseWithoutDeletingItsFilesInline`.

```java
final Path staged = findStagingDirectory();
gate.countDown();
awaitGone(staged);
```

Fixes #5718

Issue #5718, opened 2026-08-01
Flaky: ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns lists the staging directory after unblocking the deleter


robfrank:
## Problem

`ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns` fails intermittently on CI:

```
java.lang.AssertionError:
[exactly one staging directory expected]
Expected size: 1 but was: 0 in:
[]
    at ArcadeStateMachineDeferredDropTest.findStagingDirectory(ArcadeStateMachineDeferredDropTest.java:109)
    at ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns(ArcadeStateMachineDeferredDropTest.java:158)
```

Seen on the `unit-tests` job of an unrelated PR (#5709), where it was the only failure out of 772 tests in the module.

## Root cause: an ordering race in the test

The test releases the deleter before it looks for the directory the deleter is about to remove:

```java
gate.countDown();                     // unblocks the deferred deleter
awaitGone(findStagingDirectory());    // ...then lists the staging dirs
```

`gate.countDown()` lets the blocked executor task proceed, so `DeferredDatabaseDeleter` can complete its
recursive delete before `findStagingDirectory()` runs `Files.list`. When the deleter wins, zero staging
directories remain and the `hasSize(1)` assertion fails.

The sibling test `applyDeregistersTheDatabaseWithoutDeletingItsFilesInline` gets this right - it calls
`findStagingDirectory()` **before** `gate.countDown()`.

## Proof

The test passes 6/6 locally as-is. Inserting a delay between the two lines makes it fail deterministically with
the identical assertion:

```java
gate.countDown();
Thread.sleep(500);                    // let the unblocked deleter win the race
awaitGone(findStagingDirectory());
```

```
[ERROR] ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns:159->findStagingDirectory:109
[exactly one staging directory expected]
Expected size: 1 but was: 0
```

So the flake is purely a test-side ordering bug, not a defect in `ArcadeStateMachine` or
`DeferredDatabaseDeleter`. On a loaded CI runner the deleter simply wins more often.

## Suggested fix

Resolve the staging directory before unblocking the deleter, matching the sibling test:

```java
final Path staged = findStagingDirectory();
[...]

PR #5731, opened 2026-08-01
fix: read TX_RETRY_DELAY from database configuration instead of global static (#5693)

### LINKED ISSUE #5693: TX_RETRY_DELAY is declared SCOPE.DATABASE but read from the global static, so a per-database override is ignored

waterWang:
## Summary

Fixes #5693. TX_RETRY_DELAY is declared SCOPE.DATABASE, but the retry loop reads it from the global static via `GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger()`, so a per-database override is silently ignored.

## Fix

Changed line 1377 in `LocalDatabase.java` from:
```java
final int retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
```
to:
```java
final int retryDelay = configuration.getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
```

This matches the pattern used by the adjacent TX_RETRIES setting (line 1351).

## Backward compatibility

ContextConfiguration falls back to the global value when nothing is set per-database, so existing global tuning keeps working, and the per-database form starts working.

## Verification

A test that sets the value only on the database instance and asserts the observed backoff changes will now work correctly.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -1374,7 +1374,7 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
     if (attempts < 1)
       attempts = 1;
 
-    final int retryDelay = GlobalConfiguration.TX_RETRY_DELAY.getValueAsInteger();
+    final int retryDelay = configuration.getValueAsInteger(GlobalConfiguration.TX_RETRY_DELAY);
 
     boolean duplicatedKeyRetried = false;
 
--- a/ha-raft/src/test/java/com/arcadedb/server/ha/raft/ArcadeStateMachineDeferredDropTest.java
+++ b/ha-raft/src/test/java/com/arcadedb/server/ha/raft/ArcadeStateMachineDeferredDropTest.java
@@ -154,8 +154,9 @@ class ArcadeStateMachineDeferredDropTest {
 
     assertThat(new DatabaseFactory(databaseDirectory.toString()).exists()).isFalse();
 
+    final Path staged = findStagingDirectory();
     gate.countDown();
-    awaitGone(findStagingDirectory());
+    awaitGone(staged);
   }
 
   /**


robfrank:
Thanks for this one, @waterWang, and sorry for the duplicate work. Closing in favour of #5729, which is the same fix for #5693 with two advantages:

- #5729 also patches the second read site. `TX_RETRY_DELAY` is read from the global static in **two** places, `LocalDatabase.transaction()` and `RetryStep` (the SQL `COMMIT RETRY` backoff). This PR only covers the first, so a per-database override would still be ignored by `BEGIN ... COMMIT RETRY n` scripts.
- This branch also carries the ha-raft `ArcadeStateMachineDeferredDropTest` change from #5726. That fix already landed in main via #5741 (merged as `fix(test) #5718: resolve the staging directory before unblocking the deferred deleter`), which is part of why this branch now conflicts.

The analysis and the fix itself are correct, and they are exactly what #5729 carries. I have pushed the second read site and the regression tests there.

robfrank:
Superseded by #5729.

Issue #5697, opened 2026-08-01
[build] the benchmark exclusion CLAUDE.md describes is commented out, so mvn test runs benchmarks and loses the fork


lvca:
`CLAUDE.md` instructs that performance and benchmark tests be annotated so they are "skipped from regular CI builds":

> - Use `@Tag("benchmark")` for pure microbenchmarks (e.g. JMH-style or comparison runs)
> - Use `@Tag("slow")` for functional regression tests that take noticeably long

36 engine test classes carry `@Tag("benchmark")` and 53 carry `@Tag("slow")`, so the annotating half of that instruction has been followed. The excluding half has not: in the root `pom.xml`, the surefire configuration that would act on the tag is commented out, with a comment asking for the line that is sitting right under it.

```xml
<configuration>
    <!-- ADD THIS LINE TO SKIP BENCHMARKS BY DEFAULT -->
<!--                    <excludedGroups>benchmark</excludedGroups>-->
```

So `mvn -pl engine test` runs every benchmark. On a developer machine with anything else competing for memory, the heavyweight classes exhaust the heap and surefire loses the forked JVM:

```
[ERROR] The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Process Exit Code: 143
[ERROR] Crashed tests:
[ERROR] performance.TestSortedIndexBuildBenchmark
```

Exit code 143 is SIGTERM, and the reported "crashed test" is whichever class happened to be in the fork when it died - `TestSortedIndexBuildBenchmark`, `ACIDTransactionTest` and `FindReferencesStatementExecutionTest` on three consecutive runs of the same tree. The build fails with `Tests run: N, Failures: 0, Errors: 0`, which reads as "nothing failed" while exiting non-zero, and N is short of the real total with nothing saying so.

That is expensive to diagnose: it looks like a flaky test in whatever area the named class happens to belong to, and it points away from the actual cause. It cost several 25-minute runs during #5674 before the pattern (different class each time, always SIGTERM, targeted runs always green) gave it away. `mvn -pl engine test -DexcludedGroups=benchmark,slow` completes cleanly at 10,468 tests.

## Suggested fix

Uncomment the line, so the default build matches what `CLAUDE.md` says the tags are for:

```xml
<excludedGroups>benchmark</excludedGroups>
```

Two things to settle:

[...]

PR #5732, opened 2026-08-01
fix: skip ghost-edge prune when caller has active transaction (#5694)

### LINKED ISSUE #5694: Ghost-edge pruning during iteration can roll back and replace the caller's transaction
### Summary

`EdgeIteratorFilter.handleCorruption` prunes an already-dangling edge reference from inside a READ, by opening a
**joined** transaction:

```java
database.transaction(() -> {
  final EdgeLinkedList outEdges = database.getGraphEngine().getEdgeHeadChunk((VertexInternal) this.vertex, direction);
  if (outEdges != null)
    outEdges.removeEdgeRID(edge);
}, true);          // <-- joinCurrentTx = true
```

If that block raises a `NeedRetryException`, `LocalDatabase.transaction` rolls back the transaction it joined - which
is the **caller's** transaction, not one it created:

```java
} catch (final NeedRetryException | DuplicatedKeyException e) {
  lastException = e;
  if (wrappedDatabaseInstance.isTransactionActive())
    wrappedDatabaseInstance.rollback();      // rolls back the JOINED (caller's) tx
  ...
```

On the next attempt `joinCurrentTx && isTransactionActive()` is now false, so it **begins and commits a transaction
of its own**. The caller - which was only iterating edges - comes back to find its transaction gone: rolled back,
replaced, and committed out from under it, with no exception raised (the prune is best-effort and, since #5678,
absorbs the retryable exception).

### Impact

An application doing this inside its own transaction:

```java
db.begin();
for (Edge e : vertex.getEdges(OUT, "Knows")) { ... }   // hits a ghost edge -> handleCorruption fires -> CME inside
// ... more work, assuming the transaction is still the one it began
db.commit();                                            // its earlier writes are gone
```

silently loses any writes it made before the iteration, because the rollback discarded them. Nothing in the read API
suggests iterating edges can end your transaction.

Reachability: `handleCorruption` only fires on a database that already contains a dangling edge reference, and the
prune must then hit a retryable condition. Both are plausible together - a database with ghosts is usually one that
saw the concurrency that produced them.

### Not introduced by #5678

`removeEdgeRID` could already raise `ConcurrentModificationException` before that PR (`StripedEdgeList.addChain`
with `strict=true` on a promoted super-node). #5678 widened the set of conditions that reach it (classic chains too)
and added the `catch (NeedRetryException)` that stops the exhausted exception propagating - which is an improvement,
but it does not address the rollback, which happens inside `LocalDatabase.transaction` before the catch is reached.

### Suggested fix

Two candidates, both worth weighing:

1. **Do not open a nested transaction when one is already active** - perform the prune inline in the caller's
   transaction, or skip it entirely and leave the ghost for a pass that owns its transaction. The prune is
   opportunistic, so skipping is a legitimate answer and the simplest correct one.
2. **Make `LocalDatabase.transaction` not roll back a transaction it did not create.** This is the more general fix
   (the same trap exists for every `joinCurrentTx = true` caller whose block can raise a `NeedRetryException`), but
   it is a broader behavioural change: the block cannot be retried without discarding the joined transaction's work,
   so the honest answer for a joined block may be to propagate rather than retry.

### Verification

Begin a transaction, write something, iterate the edges of a vertex carrying a dangling reference while forcing the
prune to raise a retryable condition, then commit - and assert the earlier write survived.

### Context

Found while reviewing the edge-delete strictness change in #5670 / PR #5678.

waterWang:
## Problem

`EdgeIteratorFilter.handleCorruption` opens a nested joined transaction via `database.transaction(..., true)` to prune a dangling edge reference. If the prune block raises a `NeedRetryException`, `LocalDatabase.transaction` rolls back the **joined** (caller's) transaction at line 1424:

```java
if (wrappedDatabaseInstance.isTransactionActive())
    wrappedDatabaseInstance.rollback();      // rolls back the JOINED (caller's) tx
```

On the next retry, `joinCurrentTx && isTransactionActive()` is now false, so it begins and commits a transaction of its own — silently discarding any writes the caller made before the iteration.

## Impact

```java
db.begin();
for (Edge e : vertex.getEdges(OUT, "Knows")) { ... }   // hits a ghost edge -> handleCorruption fires
// ... more work, assuming the transaction is still the one it began
db.commit();                                            // its earlier writes are gone
```

## Fix

The prune is opportunistic — skip it when the database already has an active transaction, leaving the ghost for a pass that owns its own transaction. This is the simplest correct fix and avoids the broad behavioral change of modifying `LocalDatabase.transaction`'s rollback behavior.

Fixes #5694

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/EdgeIteratorFilter.java
+++ b/engine/src/main/java/com/arcadedb/graph/EdgeIteratorFilter.java
@@ -101,6 +101,17 @@ public class EdgeIteratorFilter extends IteratorFilterBase<Edge> {
         LogManager.instance().log(this, Level.WARNING, "Error on loading edge %s %s. Fixing it. Error: %s", edge,
             vertex != null ? "vertex " + vertex : "", e.getMessage());
 
+      // #5694: when the caller already has an active transaction, opening a nested joined transaction
+      // is dangerous because a NeedRetryException inside the block causes LocalDatabase.transaction to
+      // roll back the JOINED (caller's) transaction, silently discarding the caller's writes. The prune
+      // is opportunistic - skip it when we cannot safely own the transaction.
+      if (database.isTransactionActive()) {
+        LogManager.instance()
+            .log(this, Level.FINE, "Skipping dangling edge %s prune from vertex %s (caller has active transaction)", edge,
+                vertex);
+        return;
+      }
+
       try {
         database.transaction(() -> {
           final EdgeLinkedList outEdges = database.getGraphEngine().getEdgeHeadChunk((VertexInternal) this.vertex, direction);
--- a/pom.xml
+++ b/pom.xml
@@ -213,7 +213,7 @@
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
                     <!-- ADD THIS LINE TO SKIP BENCHMARKS BY DEFAULT -->
-<!--                    <excludedGroups>benchmark</excludedGroups>-->
+                    <excludedGroups>benchmark</excludedGroups>
                     <properties>
                         <!-- Work around. Surefire does not include enough
                              information to disambiguate between different


robfrank:
Thanks for this, @waterWang - the diagnosis is right and the write-up of the joined-transaction rollback is what made it easy to confirm.

Closing as a duplicate of #5733, which carries the same `EdgeIteratorFilter` guard for #5694. The two edits are functionally identical, so the tie-breaker is scope: #5733 touches only `EdgeIteratorFilter.java`, while this one also carries the `pom.xml` surefire change for #5697. That belongs in its own PR against its own issue - it changes what every `mvn test` in the repo runs, which deserves to be reviewed and reverted independently of a graph-engine fix, and it is also why the License Compliance check went red here. Please do open it separately; the reasoning in the commit message is sound.

Work on #5733 continues there: it has been rebased onto current main and now carries the regression test #5694 asked for (a caller transaction survives a ghost-edge traversal, planted state rather than a schedule), plus a tweak so the WARNING no longer says "Fixing it..." on the path that deliberately does not fix it.

No merging here, and nothing lost - the fix ships through #5733.

PR #5733, opened 2026-08-01
fix(engine): skip ghost-edge prune when caller has an active transaction (#5694)

### LINKED ISSUE #5694: Ghost-edge pruning during iteration can roll back and replace the caller's transaction

waterWang:
## Problem

`EdgeIteratorFilter.handleCorruption()` opens a joined transaction (`joinCurrentTx=true`) to prune a dangling edge reference. If the `removeEdgeRID` call raises a `NeedRetryException`, `LocalDatabase.transaction()` rolls back the joined transaction — which is the **caller's** transaction, not one it created. This silently discards any writes the caller made before iterating edges.

## Root cause

The retry path in `LocalDatabase.transaction()` unconditionally calls `rollback()` on the active transaction, even when `createdNewTx=false` (i.e., the block joined an existing transaction):

On the next retry, `joinCurrentTx && isTransactionActive()` is now false, so it begins and commits a transaction of its own — none of which the caller was aware of.

## Fix

Since the prune is **opportunistic** (best-effort cleanup of a reference whose edge is already gone), skip it entirely when a caller's transaction is already active. The ghost will be cleaned up on a subsequent pass that owns its own transaction.

## Impact

An application doing this inside its own transaction:

no longer silently loses its writes.

## Verification

- Compilation: `mvn compile -pl engine -am` passes on JDK 21
- Existing tests: `DanglingEdgeIteratorTest`, `GhostEdgeReporterTest`, `EdgeIndexCorruptionTest`, `BasicGraphTest` all pass

Fixes #5694

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/EdgeIteratorFilter.java
+++ b/engine/src/main/java/com/arcadedb/graph/EdgeIteratorFilter.java
@@ -93,13 +93,30 @@ public class EdgeIteratorFilter extends IteratorFilterBase<Edge> {
     if ((e instanceof RecordNotFoundException || e instanceof SchemaException) &&//
         database.getMode() == ComponentFile.MODE.READ_WRITE) {
 
+      // #5694: the prune below is opportunistic - a best-effort cleanup of a reference whose edge is already gone,
+      // run from inside a READ. It opens a JOINED transaction, so when the caller already has one open the
+      // transaction it joins is the CALLER'S: a retryable condition inside the block makes LocalDatabase.transaction
+      // roll that transaction back and then begin, run and commit one of its own, and the caller - which only
+      // iterated edges - silently loses every write it made before the iteration. Nothing in a read API may end a
+      // transaction it did not open, so the ghost is left to a pass that owns its transaction, or to CHECK DATABASE.
+      //
+      // Skipping is also the right answer when nothing raises: the prune is a WRITE, and performing it inside the
+      // caller's transaction commits or rolls it back with the caller, dirties an edge-list page a read-only
+      // transaction never touched (the conflict unit is the page, so that page can cost the caller a conflict at
+      // commit), and under HA rides along in the caller's replicated transaction.
+      final boolean ownsTransaction = !database.isTransactionActive();
+      final String outcome = ownsTransaction ? "Fixing it..." : "Leaving it to a pass that owns its transaction.";
+
       if (fullStackTracePrinted < 10) {
         ++fullStackTracePrinted;
-        LogManager.instance().log(this, Level.WARNING, "Error on loading edge %s %s. Fixing it...", e, edge,
-            vertex != null ? "vertex " + vertex : "");
+        LogManager.instance().log(this, Level.WARNING, "Error on loading edge %s %s. %s", e, edge,
+            vertex != null ? "vertex " + vertex : "", outcome);
       } else
-        LogManager.instance().log(this, Level.WARNING, "Error on loading edge %s %s. Fixing it. Error: %s", edge,
-            vertex != null ? "vertex " + vertex : "", e.getMessage());
+        LogManager.instance().log(this, Level.WARNING, "Error on loading edge %s %s. %s Error: %s", edge,
+            vertex != null ? "vertex " + vertex : "", outcome, e.getMessage());
+
+      if (!ownsTransaction)
+        return;
 
       try {
         database.transaction(() -> {
--- a/engine/src/test/java/com/arcadedb/graph/Issue5694GhostEdgePruneCallerTransactionTest.java
+++ b/engine/src/test/java/com/arcadedb/graph/Issue5694GhostEdgePruneCallerTransactionTest.java
@@ -1,0 +1,233 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.graph;
+
+import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.TestHelper;
+import com.arcadedb.database.Database;
+import com.arcadedb.database.DatabaseInternal;
+import com.arcadedb.database.RID;
+import com.arcadedb.database.TransactionContext;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * #5694: the opportunistic prune of a ghost edge, run from inside a READ, must never touch the transaction the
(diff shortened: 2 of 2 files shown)


robfrank:
Thanks @waterWang - the read is correct, and this is the fix #5694 itself recommended (option 1: the prune is opportunistic, so skipping is a legitimate answer and the simplest correct one). I have pushed three things on top of your commit rather than sending them back for another round.

**1. Rebased onto current main.** The branch was 22 commits behind, and the `unit-tests` red on the old base was `ArcadeStateMachineDeferredDropTest.theDatabaseNameIsReusableAsSoonAsApplyReturns` - #5718, since fixed on main by #5741 (`c4433a32e`), which was not in your base and is now in the branch. The `ha-integration-tests` red was `Issue5410AbandonedTicketReleaseIT`, on the roster of known load-only flakes in #5630; it passes in isolation on a control worktree at main (16.7 s, 1/1).

**2. The regression test the repo requires**, `Issue5694GhostEdgePruneCallerTransactionTest`. It plants the state the race leaves instead of driving a schedule, so it is deterministic and single-threaded:

- A promoted super-node with one stripe chain whose head record has been deleted. `StripedEdgeList.removeEdgeRID` walks every chain through `allChains(true)`, and `addChain` maps an unreadable head to a `ConcurrentModificationException` - so the prune raises on every attempt, which is exactly the condition in your write-up. The test begins a transaction, writes a vertex, iterates the hub's edges past a ghost, and asserts the transaction is still active, is still the same `TransactionContext`, still holds the write, and commits it. With the guard stubbed off it fails on `iterating edges must not end the transaction the caller began`.
- The classic layout, where nothing raises and the prune would have *succeeded*: the reference to the ghost must still be in the edge list after the caller commits. With the guard stubbed off, the repair rides out on the caller's commit and the assertion fails.

Both were confirmed to go red with the guard stubbed to a no-op and green with it restored; the 95 tests around the graph package pass.

**3. The WARNING no longer lies.** The guard sat *after* `"Error on loading edge %s %s. Fixing it..."`, so every ghost met inside a transaction announced a fix that was not going to happen - and inside a server that is every ghost, since HTTP handlers wrap commands in a transaction. The message now reports which of the two branches was taken.

One thing worth stating plainly for the maintainer, since it is the crux of the issue: **skipping leaves the ghost unpruned in the common case**, because in server, HTTP and SQL contexts a transaction is essentially always open. That is the right trade anyway, and not only because losing a caller's writes is worse. The prune is a WRITE performed from inside a READ: running it in the caller's transaction commits or rolls it back with the caller, dirties an edge-list page a read-only transaction never touched (the conflict unit is the page, so it can cost that caller a `ConcurrentModificationException` at commit), and under HA rides along in the caller's replicated transaction. A deferred or background prune would be the way to get the cleanup back, and `CHECK DATABASE` already repairs this shape today. Neither belongs in this PR.

Also worth noting: the general trap in `LocalDatabase.transaction` (option 2 in #5694) is untouched - it still rolls back a transaction it did not create for every `joinCurrentTx = true` caller whose block can raise a `NeedRetryException`. That deserves its own issue.

Finally, `claude-review` cannot pass on this PR: it fails at `Could not fetch an OIDC token` because a fork PR has no access to the secrets. That is infrastructure, not the change.

robfrank:
CI verdict on `cdd64fe6` (rebased head). Green: `unit-tests`, `slow-unit-tests`, `build-and-package`, `builder-tests`, all six e2e jobs, Codacy, Meterian. Three reds, none of them this change:

**`ha-integration-tests`** - `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` and `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`. Both fail on **main** in CI: runs [30756110489](https://github.com/ArcadeData/arcadedb/actions/runs/30756110489) and [30756178015](https://github.com/ArcadeData/arcadedb/actions/runs/30756178015) fail on exactly the same two, and [30757767338](https://github.com/ArcadeData/arcadedb/actions/runs/30757767338) on the first. A control worktree at `a560df6e9` reproduces the second locally without any of this PR's code:

```
RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges <<< ERROR!
com.arcadedb.exception.TransactionCommittedRemotelyException: ... committed cluster-wide but the local apply failed
```

`Issue5569SlotMergeDeleteRaftIT` passes alone on that control (24 s), which is what #5630 already records for it, along with "also fails on main".

**`integration-tests`** - `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites`, failing in the teardown comparison with `DatabaseAreNotIdentical: Types: DB1 6 <> DB2 5`. That is the signature on the #5630 roster verbatim. It passes alone both on the control at main (49.9 s) and on this branch (44.4 s), and the previous run of this same guard on this PR had `integration-tests` green. Note also that this job runs Maven with `--fail-never`, so the red comes from the reporter step, not from the build.

**`claude-review`** - `Could not fetch an OIDC token`, i.e. a fork PR has no access to the secrets. Infrastructure. Worth knowing since this is the merge gate.

The `unit-tests` red on the previous head was `ArcadeStateMachineDeferredDropTest` (#5718), fixed on main by #5741 (`c4433a32e`) and picked up by the rebase; `unit-tests` is green now.

Issue #2560, opened 2025-09-26
ImmutableDocument: loading exception always turned into null in specific method only


pawellhasa:
In class `ImmutableDocument` there is code recently changed that causes issue for me:

`
    public Object get(final String propertyName) {
    if (propertyName == null)
      return null;

    try {
      checkForLazyLoading();
      return database.getSerializer()
          .deserializeProperty(database, buffer, new EmbeddedModifierProperty(this, propertyName), propertyName, rid);
    } catch (Exception e) {
      LogManager.instance().log(this, Level.SEVERE, "Error on loading property '%s' from record %s", e, propertyName, rid);
      return null;
    }
  }
`
It wraps `checkForLazyLoading();` into try/catch which makes it inconsistent with other methods (as they continue to throw exceptions when calling it). I have set of listeners that check user access permissions triggered during lazy loading (record create/updated listeners, etc).

Is it possible to move `checkForLazyLoading();` line outside of try/catch or provide an exception class that I could extend that would be rethrown? Currently it returns me null and logs print out is not something I can act on.

PR #5734, opened 2026-08-01
fix(engine) #5723: a copied type keeps its index definitions, its records' content, and index entries

### LINKED ISSUE #5723: Renaming a type resets every index's page size to the default
## What happens

`LocalSchema.renameType()` copies the records into the new type and then recreates the indexes from scratch, using the three-argument `createTypeIndex` overload - which does not take a page size:

```java
// engine/src/main/java/com/arcadedb/schema/LocalSchema.java:613
// COPY INDEXES
for (final Index index : oldType.getAllIndexes(false))
  newType.createTypeIndex(index.getType(), index.isUnique(),
      index.getPropertyNames().toArray(new String[index.getPropertyNames().size()]));
```

So an index deliberately created with a non-default page size silently reverts to its type's default when the type is renamed. Every other attribute of the index survives the rename; this one does not.

```java
schema.buildTypeIndex("Doc", new String[] { "k" })
    .withType(Schema.INDEX_TYPE.LSM_TREE).withUnique(true)
    .withPageSize(16_384)
    .create();

schema.getType("Doc") ... // rename to "Doc2"
// the index on Doc2 is now at 262144, not 16384
```

## Scope

This affects **every** index type, not just `HASH` - anyone who tuned a page size for an `LSM_TREE`, `FULL_TEXT` or vector index loses it on a rename. It is silent: nothing logs, and the index keeps working, so the only symptom is a performance or footprint characteristic that quietly changed.

## Not a regression from #5713

Worth stating explicitly, because #5713 touched the neighbouring code. Before that fix, the unset builder field defaulted to `LSMTreeIndexAbstract.DEF_PAGE_SIZE`, which `HashIndex` remapped to 65536; after it, unset resolves to the hash default, 65536. Identical outcome for `HASH`, and unchanged for every other type. The behaviour predates #5713 - that fix only made it easier to notice, because "unset" is now a meaningful state rather than an accident.

## Suggested fix

Pass the page size through, using the accessor #5713 introduced for exactly this situation:

```java
newType.createTypeIndex(index.getType(), index.isUnique(),
    index.getPropertyNames().toArray(new String[index.getPropertyNames().size()]),
    ((IndexInternal) index).getPageSizeForNewFile());
```

`getPageSizeForNewFile()` rather than `getPageSize()` is the point: the value goes back through the validating creation path, and it answers the current page size unless that value is not legal to create with, in which case it repairs it. The five other drop-then-recreate sites (`RebuildIndexStatement`, `TruncateTypeStatement`, `DatabaseChecker`, and both `LocalDocumentType` propagation paths) already use it.

A regression test would rename a type carrying an index at a non-default page size and assert the recreated index still reports it.

Found while auditing the page-size carry-over sites for #5713 (PR #5717); left out of that PR deliberately, since it changes behaviour for all index types and deserves its own blast radius.

lvca:
Closes #5723.

The issue reports that `LocalSchema.copyType()` loses the page size of every index on the copy. It does. Writing the regression test for it surfaced two more defects in the same thirty lines, and they are the ones that make the copy **wrong** rather than differently tuned.

(The issue names `renameType()`. There is no such method: `ALTER TYPE ... NAME ...` goes through `LocalDocumentType.rename()`, which renames buckets in place and never recreates an index. The affected method is `copyType()`, which is what a document-to-vertex conversion goes through.)

## 1. Every index on the copy came out EMPTY

The indexes were created inside the transaction that copies the records. `TypeIndexBuilder.create()` builds each bucket's index in its own transaction (`joinCurrent=false`), which from inside an enclosing transaction neither sees the records still pending in it nor commits the entries it produces. So the copy carried indexes with zero entries, and any query the planner routed through one answered nothing - silently, since an empty index is indistinguishable from a type with no matching rows.

```java
schema.copyType("Doc", "Doc2", LocalVertexType.class, 1, 65_536, 1_000);
database.query("sql", "SELECT FROM Doc2 WHERE k = 'v1'");   // 0 rows, and the record is right there
```

Verified by experiment, not by reading: moving the loop back inside the transaction reproduces `entries=0` exactly.

## 2. The copied records were EMPTY

`ImmutableDocument.propertiesAsMap()` was the one accessor of that class that did not call `checkForLazyLoading()` - it returned `Collections.emptyMap()` whenever `buffer == null`, which is the state of every record handed out by `iterateType()` and `scanBucket()`. `copyType()` reads each source record that way, so it faithfully created the right *number* of records, each with no properties. The same silent emptiness hit a `DetachedDocument` built off a record straight from a scan.

It now loads first, like `get()`, `has()`, `getPropertyNames()`, `toMap()` and `toJSON()` already did. The empty map remains the answer only for a record that genuinely cannot be materialised (no database, no RID, or an after-read event that filtered it away), so the class's exception-propagation contract from #2560 is preserved - the two new entries in `ImmutableDocumentLazyLoadingInconsistencyTest`'s accessor list assert exactly that.

## 3. The reported one, wider than reported

The copy went through the three-argument `createTypeIndex` overload, so the index type, the uniqueness flag and the property list survived and **everything else** was replaced by a default: the tuned page size from the report, the null strategy, the collations that make a lookup case-insensitive, and the entire type-specific configuration. That last one is not a tuning difference - a copied `FULL_TEXT` index tokenized and ranked with the default analyzer, a copied `GEOSPATIAL` index dropped to the default geohash resolution, and a copied `LSM_VECTOR` index came up with `dimensions` 0.

## The better alternative to the suggested fix

The issue suggests passing `getPageSizeForNewFile()` through. That fixes one attribute of five and leaves a copied full-text or vector index broken, so the fix generalises the accessor instead of the call:

- **`IndexInternal.getMetadataForNewFile()`**, the companion of the `getPageSizeForNewFile()` that #5713 introduced. It exists because `getMetadata()` cannot answer this question for the wrapper index types: `LSMTreeFullTextIndex` keeps its analyzers and BM25 parameters in its own `FullTextIndexMetadata` and `LSMTreeGeoIndex` keeps its resolution and layout in plain fields, while both `getMetadata()`s delegate to the underlying LSM-Tree and hand back a definition with none of it.
- **`IndexMetadata.copy(typeName, propertyNames, bucketId)`** promoted from `LSMVectorIndexMetadata` to a polymorphic method with overrides on all four subclasses, plus a `final copyCommonTo()` for the shared fields. There is no second field list to keep in sync, which is the point. Per-index RUNTIME state deliberately does not ride along - the dense vector index's build state, the full-text corpus counters - because the new file starts empty and inherited statistics would describe a different set of records.

A user-supplied index name is the one attribute deliberately not carried over: it is unique across the schema and still belongs to the source type, so the copy takes the auto-derived `NewType[properties]` form rather than colliding with it.

## Deliberately out of scope

The other sites that rebuild an index from its own definition - `TruncateTypeStatement`, `DatabaseChecker` (CHECK DATABASE FIX), and both `LocalDocumentType` propagation paths - still read `getMetadata()` and so still lose a full-text or geospatial configuration. Each is a distinct user-visible operation that deserves its own regression test and its own blast radius, which is the same reasoning #5723 used to split itself off from #5717; a follow-up issue tracks them. `RebuildIndexStatement` is already correct - it reconstructs the typed metadata from the persisted JSON - and could now use the new accessor instead, but that is a refactor with no behaviour change.

## Tests

`Issue5723CopyTypeIndexConfigurationTest` (new, 8 cases): page size on `LSM_TREE` and on `HASH` (where the value goes back through the validating creation path #5713 added), null strategy, CI collation asserted through a case-insensitive lookup on the copy, full-text analyzer/BM25/operator asserted through an English-stem lookup, geospatial precision asserted through a spatial query, vector dimensions and similarity, and the manual-index-name non-collision. Plus `entries == 1` on the copy, which is what catches defect 1.

`ImmutableDocumentLazyLoadingInconsistencyTest`: `propertiesAsMap()` and `propertiesAsMap(fields)` added to the exception-propagation and happy-path accessor lists, and a new case asserting a record straight off `iterateType()` answers its properties.

Verified locally: `com.arcadedb.query.**` 10330 tests 0 failures, `com.arcadedb.database.**` 176 tests 0 failures, `com.arcadedb.schema.** + com.arcadedb.index.**` 942 tests with one unrelated timing flake (`LSMVectorIndexRecoveryTest#vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild`, an inactivity-rebuild deadline assertion, passes in isolation), `com.arcadedb.graph.** + com.arcadedb.serializer.**` 336 tests 0 failures.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1599,4 +1599,56 @@ build indexes through the embedded API with a computed page size, check that the
 An index that already exists on disk is unaffected either way - the page size is read from the component file, not from
 a builder.
 
+## Copying a type copies its records and its index definitions, not just their names
+
+`Schema.copyType()` - what `DocumentType` → `VertexType` conversion goes through - was reported as losing the page
+size of every index on the copy (#5723). It was, and that turned out to be the least of it. The same thirty lines held
+three defects, and the two that were not reported are the ones that made the copy wrong rather than differently tuned.
+
+**The copy had no data in its indexes.** The indexes were created inside the transaction that copies the records.
+`TypeIndexBuilder.create()` builds each bucket's index in its own transaction (`joinCurrent=false`), which from inside
+an enclosing transaction neither sees the records still pending in it nor commits the entries it produces. So every
+index on the copy came out empty, and any query the planner routed through one answered nothing - silently, since an
+empty index is indistinguishable from a type with no matching rows:
+
+```java
+schema.copyType("Doc", "Doc2", LocalVertexType.class, 1, 65_536, 1_000);
+database.query("sql", "SELECT FROM Doc2 WHERE k = 'v1'");   // 0 rows, and the record is right there
+```
+
+The index build now runs after the record-copy transaction has committed.
+
+**The copied records were empty.** `ImmutableDocument.propertiesAsMap()` was the one accessor of that class that did
+not lazy-load: it answered `Collections.emptyMap()` whenever the record had not been materialised yet, which is the
+state of every record handed out by `iterateType()` and `scanBucket()`. `copyType()` reads each source record that
+way, so it faithfully created the right *number* of records, each with no properties. The same silent emptiness
+affected a `DetachedDocument` built off a record straight from a scan. `propertiesAsMap()` now loads first, like
+`get()`, `has()`, `getPropertyNames()` and `toJSON()` already did; the empty map remains the answer only for a record
+that genuinely cannot be materialised.
+
+**And the index definitions were rebuilt from three attributes.** The copy went through the three-argument
+`createTypeIndex` overload, so the index type, the uniqueness flag and the property list survived and everything else
+was replaced by a default: the tuned page size from the report, the null strategy, the collations that make a lookup
+case-insensitive, and the entire type-specific configuration. That last one is not a tuning difference - a copied
+`FULL_TEXT` index tokenized and ranked with the default analyzer instead of the configured one, a copied `GEOSPATIAL`
+index dropped to the default geohash resolution, and a copied `LSM_VECTOR` index came up with `dimensions` 0.
+
+Carrying a definition into a new index file now has an accessor of its own, `IndexInternal.getMetadataForNewFile()`,
+alongside the `getPageSizeForNewFile()` that #5713 introduced for the page size. It exists because `getMetadata()`
+cannot answer this question for the wrapper index types: a full-text index keeps its analyzers and BM25 parameters in
+its own `FullTextIndexMetadata` and a geospatial one keeps its resolution and layout in plain fields, while
+`getMetadata()` delegates to the underlying LSM-Tree and hands back a definition with none of it. `IndexMetadata.copy()`
+is now polymorphic across all five metadata classes, so each carries its own settings and there is no second field list
+to forget. Per-index runtime state deliberately does not ride along: the copy starts empty, so inheriting the dense
+vector index's build state or the full-text corpus counters would describe a different set of records.
+
+A user-supplied index name is the one attribute not carried over. It is unique across the schema and still belongs to
+the source type, so the copy takes the auto-derived `NewType[properties]` form rather than colliding with it.
+
+The other sites that rebuild an index from its own definition - `TRUNCATE TYPE`, `CHECK DATABASE FIX`, and the
+propagation to a freshly added bucket or sub type - still read `getMetadata()` and so still lose a full-text or
+geospatial configuration. They are unchanged here and tracked separately: each is a distinct user-visible operation and
+deserves its own regression test rather than riding along with this one. `REBUILD INDEX` is already correct; it
+reconstructs the typed metadata from the persisted JSON.
+
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1645,6 +1645,13 @@ vector index's build state or the full-text corpus counters would describe a dif
 A user-supplied index name is the one attribute not carried over. It is unique across the schema and still belongs to
 the source type, so the copy takes the auto-derived `NewType[properties]` form rather than colliding with it.
 
+**One behaviour change reaches outside `copyType()`.** `LSMVectorIndexMetadata.copy()` already existed and carried only
+the collations; routing it through the shared `copyCommonTo()` means it now also carries the user-supplied index name.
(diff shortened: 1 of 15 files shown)


lvca:
Responses to the latest review.

**(1) The TRUNCATE-name behaviour change** is real, is a fix, and is covered by `truncatingAManuallyNamedVectorIndexKeepsItsName` (mutation-checked: dropping the `typeIndexName` line from `copyCommonTo()` makes it fail). It is now also written up in `docs/release-26.8.1.md` rather than left to be discovered - it is the one effect of this PR outside `copyType()`. Worth noting the asymmetry it removes: every *other* index type already kept its manual name through a truncate, because `TruncateTypeStatement` hands the original metadata object straight back and only `LSM_VECTOR` goes through `.copy()`. So this aligns vector with the rest rather than introducing a new behaviour.

**(2) The `ScanWithFilterStep` projection push-down is not reachable, so there is nothing to regress-test.** I wrote the suggested `SELECT a, b FROM T` case and it passed *with the `propertiesAsMap` fix reverted*, which sent me to the planner rather than to the test:

`SelectExecutionPlanner.computeProjectedProperties()` returns `null` whenever a WHERE clause is present:

```java
// Extract from WHERE clause
if (info.whereClause != null) {
  // WHERE can reference arbitrary fields, so we need all properties if we can't analyze it
  // For safety, return null to indicate all properties needed
  return null;
}
```

`ScanWithFilterStep` only exists when there *is* a WHERE clause, so `projectedProperties != null` and `whereClause != null` cannot both hold in it - the `if (projectedProperties != null && record instanceof ImmutableDocument)` branch is dead code today. `EXPLAIN SELECT a, b FROM Doc WHERE 1 = 1` confirms it empirically: the plan is `SCAN WITH FILTER` with no `[projected: ...]` marker, and the columns come from a separate `CALCULATE PROJECTIONS` step.

I removed the test rather than keep one that asserts nothing. The dead branch is pre-existing and out of scope here; it would be worth either wiring the push-down up or deleting it, separately.

**(3) `dropType` throwing during cleanup** - agreed, residual window, consistent with the existing pattern, left alone.

**(4) `new String[propertyNames.size()]`** - left as is to match the surrounding file.

lvca:
Follow-up filed as #5742: the four remaining carry-over sites (`TRUNCATE TYPE`, `CHECK DATABASE FIX`, and both `LocalDocumentType` propagation paths), each of which the new `getMetadataForNewFile()` accessor is what they need. It also carries the crash-window note from observation (1) - a hard crash between the record commit and the index build leaves a populated-but-unindexed copy, which the `catch`/`dropType` path cannot cover.

PR #5735, opened 2026-08-01
fix(engine): reject concurrent GraphBatch instances on the same database (#5666)

### LINKED ISSUE #5666: Two concurrent GraphBatch instances on the same database silently lose edges, and nothing prevents them
Found while investigating discussion #5597 (663M vertices / 14B edges). The reporter tried multiple concurrent `GraphBatchLoad` streams, hit `ConcurrentModificationException` and page-version mismatches, and fell back to a single writer. The fallback is correct, but the reason is worse than contention: two concurrent `GraphBatch` instances on the same database can **silently lose edges**, and nothing stops a caller from creating them.

## Failure scenario

`database.batch()` (`LocalDatabase:1806-1811`) is an unguarded factory. Two batches A and B run concurrently and both touch vertex V:

1. A appends V's first edge. `getOrCreateOutSegmentDeferred()` reads V, finds `getOutEdgesHeadChunk() == null`, creates segment C1, and calls `persistNewSegment()` (`:1620-1632`), which records `deferredOutHead[V] = C1` **in A's own map**. V's record on disk is untouched.
2. B appends its own edge for V. It reads V, still sees `getOutEdgesHeadChunk() == null` (A's pointer is deferred), creates segment C2, records `deferredOutHead[V] = C2` in B's map.
3. Both close. `batchUpdateVertexHeadChunks()` (`:1004-1063`) writes each batch's head pointer onto V. Last writer wins.

The loser's whole segment chain for V is orphaned: the edges are persisted, reachable from nothing, and the integrity checker will report V's edges as missing. No exception anywhere.

The head pointer being deferred to `close()` is what makes this invisible; sequential batches are fine because `close()` publishes before the next one reads.

## Secondary interference on the same path

- Both batches mutate the executor's global WAL flags (`:1300-1301`, `:1982-1983`) and restore whatever they happened to read on entry, so B can run with A's durability settings. On a replicated database that means writes that never reach followers.
- `async.waitCompletion()` (`:1324`, `:2005`) is executor-wide, so each batch blocks on the other's tasks.
- Slot assignment is `bucket % parallelLevel` (`:1318`, `:1997`), so two batches touching the same vertex bucket serialize on the same slot regardless. There is no parallelism to be won here.
- Related: #5665 (each flush respawns the pool, force-exiting the other batch's tasks).

## Proposal

Make the single-writer contract explicit rather than folklore. `LocalDatabase.batch()` should refuse a second concurrent `GraphBatch` on the same database with a message that says why and points at the intended pattern (one batch, `parallelFlush` for the fan-out), released on `close()`. On the server side, `ArcadeDbGrpcService.graphBatchLoad` should surface that as a clean `FAILED_PRECONDITION` instead of the current interleaving.

The alternative, making concurrent batches actually safe, means publishing head pointers eagerly and sharing the chunk caches, which gives up most of what makes the bulk path fast. Not worth it: the fan-out already exists inside one batch.

## Verification

Test: two `GraphBatch` instances on one database, both adding edges to the same vertex, asserting the second construction is rejected. Plus a regression test that the documented single-batch path still connects both directions and passes the integrity check.

waterWang:
Closes #5666.

## Problem

Two concurrent `GraphBatch` instances on the same database silently lose edges. The head chunk pointer of a vertex is only published by `close()`, so both batches read `getOutEdgesHeadChunk() == null`, each builds its own segment chain, and the last one to close wins. The loser's edges stay on disk reachable from nothing and the integrity checker reports the vertex as missing them. No exception is raised anywhere.

## Fix

A database grants one `GraphBatch` slot at a time.

1. **LocalDatabase** owns an `AtomicBoolean` slot with `batchStarted()` / `batchFinished()`. The owner is the concrete instance, never the wrapper: `batch()` hands the batch the outermost `wrappedDatabaseInstance` to write through (so commits still flow through the Raft replication layer) and `this` as the guard owner.
2. **GraphBatch.Builder.build()** takes the slot, not `Database.batch()`. A builder rejected for a bad setting, or simply dropped, must not lock the database out of batching, and both the HTTP and the gRPC batch endpoints configure the builder from client input.
3. **GraphBatch.close()** hands the slot back from an outermost `finally`, so a failing `flush()`, `connectDeferredIncomingEdges()` or `batchUpdateVertexHeadChunks()` cannot leak it. The release is idempotent, so a double close cannot free a slot a later batch has meanwhile taken.
4. **GraphBatch.abandon()** releases the slot without flushing, for callers that drop a failed batch on a thread that cannot afford a full `close()`.
5. **ArcadeDbGrpcService.graphBatchLoad** maps the refusal to `FAILED_PRECONDITION`, scoped to `build()` so genuine engine faults keep reporting `INTERNAL`, and calls `abandon()` on the error path where it drops the batch without closing it.

## Behaviour change

Two overlapping batch loads on one database are now refused instead of quietly corrupting the graph. Over gRPC that is `FAILED_PRECONDITION`; over `POST /api/v1/batch` the `DatabaseOperationException` is not in the handler status ladder, so it surfaces as a 500. Sequential loads, nested-free single-writer loads and the `parallelFlush` fan-out inside one batch are unaffected.

## Tests

`Issue5666ConcurrentGraphBatchTest` (engine, 7 cases) covers both directions:

- an overlapping batch is refused on the same thread and from another thread
- sequential batches both run and connect OUT and IN
- a batch taken through an HA-style wrapper still hands its slot back
- an abandoned or rejected *builder* never took the slot
- `abandon()` frees the slot and a later `close()` does not free it a second time
- a batch whose `close()` throws (duplicate key on the trailing edge buffer) still frees the slot

`Issue5666GrpcBatchGuardIT` (grpcw, 2 cases) covers the streaming endpoint against one long-lived database, which is the only place a leaked slot is visible:

- a load that fails mid-stream is dropped without `close()`, and the next load on the same database is still accepted
- an overlapping load comes back as `FAILED_PRECONDITION`, not `INTERNAL`, and does not consume the slot it was denied

Every case was checked against a stubbed fix and goes red without it.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
+++ b/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
@@ -248,4 +248,12 @@ public interface DatabaseInternal extends Database {
    * @return Map of variable name to value
    */
   Map<String, Object> getGlobalVariables();
+
+  /**
+   * Called by {@link com.arcadedb.graph.GraphBatch#close()} to signal that a batch operation has finished.
+   * The default implementation is a no-op; database implementations that guard against concurrent
+   * batches override this to release the guard.
+   */
+  default void batchFinished() {
+  }
 }
--- a/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
+++ b/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
@@ -248,12 +248,4 @@ public interface DatabaseInternal extends Database {
    * @return Map of variable name to value
    */
   Map<String, Object> getGlobalVariables();
-
-  /**
-   * Called by {@link com.arcadedb.graph.GraphBatch#close()} to signal that a batch operation has finished.
-   * The default implementation is a no-op; database implementations that guard against concurrent
-   * batches override this to release the guard.
-   */
-  default void batchFinished() {
-  }
 }
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -202,7 +202,7 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
   private            long                                      lastUpdatedOn;
   private            long                                      lastUsedOn;
   private            int                                       cachedHashCode            = 0;
-  /** #5666: guards against concurrent GraphBatch instances on this database. Never routed through a wrapper. */
+  /** Guards against concurrent GraphBatch instances on this database. Never routed through a wrapper. */
   private final      AtomicBoolean                             batchInProgress           = new AtomicBoolean(false);
 
   protected LocalDatabase(final String path, final ComponentFile.MODE mode, final ContextConfiguration configuration,
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -1819,10 +1819,25 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
 
   @Override
   public GraphBatch.Builder batch() {
+    // Guard against concurrent GraphBatch instances on the same database (issue #5666).
+    // Two concurrent batches on the same database silently lose edges because the head
+    // pointer is deferred to close() — the last writer wins, and the loser's segment chain
+    // is orphaned. The fix is a simple AtomicBoolean guard: the first caller acquires it,
+    // subsequent callers are rejected with a clear message.
+    if (!batchInProgress.compareAndSet(false, true))
+      throw new DatabaseOperationException(
+          "A GraphBatch is already in progress on this database. Concurrent batches silently lose edges. "
+              + "Use a single GraphBatch (parallelFlush for parallel fan-out).");
+
     // Use the outermost wrapper so that commits flow through any HA/replication layer.
     // Without this, GraphBatch.commit() would short-circuit the Raft replication wrapper
     // installed by the HA plugin and writes would never reach followers (issue #4076).
     return GraphBatch.builder(wrappedDatabaseInstance);
+  }
+
+  @Override
+  public void batchFinished() {
+    batchInProgress.set(false);
   }
 
   @Override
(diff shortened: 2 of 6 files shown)


robfrank:
Reviewed and pushed a follow-up commit (`5858b76`), rebased on current `main`.

The guard idea is right, the release was not. Three ways the slot leaked, plus one already-known-red CI job that this PR itself caused:

1. **The release never reached the guard on a replicated database.** `GraphBatch` writes through `wrappedDatabaseInstance`, which on HA is `RaftReplicatedDatabase`. That class does not override `batchFinished()`, so the `default` no-op on `DatabaseInternal` ran and the flag on `LocalDatabase` stayed set. The first batch on an HA database locked out every later one - which is exactly why `RaftBatchEndpointIT.postBatchOnFollowerWithMultipleNewKeys` went red on this PR with this PR's own message (`A GraphBatch is already in progress on this database`). The slot now belongs to the concrete `LocalDatabase`, passed to the builder explicitly, and `DatabaseInternal.batchFinished()` is gone.
2. **`close()` released after two passes that can throw.** `connectDeferredIncomingEdges()` and `batchUpdateVertexHeadChunks()` sit above the release, so a failure in either leaked the slot. The release is now in the outermost `finally`.
3. **The slot was taken by `Database.batch()`, before the builder was configured.** Both `PostBatchHandler` and `ArcadeDbGrpcService` build the builder from client input, so one request with a bad option would have locked the database out of batching until restart. Acquisition moved into `Builder.build()`, with the slot handed straight back if the constructor throws.
4. **grpcw drops a failed batch without closing it** (deliberately, to keep `waitCompletion()` off the gRPC thread), which leaked the slot on every failed load. Added `GraphBatch.abandon()` - releases the slot, flushes nothing - and called it there. The `FAILED_PRECONDITION` mapping is now scoped to `build()` so genuine engine faults keep reporting `INTERNAL`; catching every `DatabaseOperationException` across the whole `onNext` would have relabelled a lot of real internal errors.

Also added `Issue5666ConcurrentGraphBatchTest` (6 tests) - the PR had none. It covers both directions the issue asks for: overlapping batch rejected on the same thread and on another one, sequential second batch still runs and connects OUT and IN, plus the wrapper path, the abandoned builder, and the double release. Every test was checked against a stubbed fix before being trusted.

Local verification: `Issue5666ConcurrentGraphBatchTest` 6/6, `RaftBatchEndpointIT` 3/3, `PostBatchHandlerIT` 26/26, `RemoteGraphBatchIT` 12/12, grpcw unit tests 44/44, GraphBatch engine tests 24/24.

Remaining wart, not fixed here: over HTTP the rejection surfaces as a 500 "Internal error" rather than a 4xx. Worth a follow-up on the handler's status ladder.

Issue #4515, opened 2026-06-08
`VectorLocationIndex` `LinkedHashMap accessOrder=true` iterated without holding wrapper monitor


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.index.vector.VectorLocationIndex`

### Summary

`accessOrder=true` makes every `get()` a structural modification.
`Collections.synchronizedMap` requires manual `synchronized(map)` for
iteration. Streaming `keySet()` without that, and calling `get()` inside the
filter, mutates the very map being iterated.

### Code

`engine/com/arcadedb/index/vector/VectorLocationIndex.java:86–101, 150–152, 171–184`

```java
this.locations = Collections.synchronizedMap(
    new LinkedHashMap<Integer, VectorLocation>(initialCapacity, 0.75f, true) {  // accessOrder=true
      @Override protected boolean removeEldestEntry(…) { return size() > maxSize; }
    });
…
public IntStream getActiveVectorIds() {
  return locations.keySet().stream()
      .filter(id -> !locations.get(id).deleted)                                 // mutating get during iteration
      .mapToInt(Integer::intValue);
}
```

### Impact

Under concurrent search/insert load with `maxSize > 0`,
`ConcurrentModificationException` from query paths, or silent corruption of
the LinkedHashMap doubly-linked list causing data loss.

### Suggested fix

Drop access-order (use insertion-order LRU), or use Caffeine, or wrap
iterating methods in `synchronized (locations) { … }` and snapshot keys into
a list before streaming.

PR #5736, opened 2026-08-01
fix(engine) #5559: a vector index refuses a location cache size it cannot honour

### LINKED ISSUE #5559: LSM_VECTOR: setting locationCacheSize silently truncates the index - count and search only see the last N vectors

lvca:
Fixes #5559.

## What was wrong

`arcadedb.vectorIndex.locationCacheSize` and the per-index `locationCacheSize` METADATA key capped the in-memory location index of an `LSM_VECTOR` index and let it FIFO-evict. That was never a cache bound. A location is the only record of which record a vector id belongs to and where its entry sits in the index file, and nothing on disk maps a vector id back to an offset, so an evicted entry could not be reloaded - and every reader reads a missing location as deleted.

The reported shape: 200 vectors, 16 dimensions, INT8 quantization, `locationCacheSize: 10`. `countEntries()` reported 10, and a search probing a vector with its own embedding answered with a different vertex.

## State of the fix before this PR

#5568 already made `getLocationCacheSize()` return `-1` unconditionally, so the truncation itself no longer reproduces on `main` - verified: the new test passes there, and restoring eviction makes `countEntries()` answer 10 instead of 200. What was left is that the setting was still **accepted and ignored**: `CREATE INDEX ... METADATA {"locationCacheSize": 10}` succeeded, persisted the value, and `schema:indexes` kept echoing a bound that nothing enforced. That is the same silence the issue objects to, moved from the data into the metadata. The bounded backend was also still constructible, so the failure was prevented by policy rather than by construction.

## What this PR does

The issue proposed either "make it an honest cache" or "remove it". An honest cache needs an on-disk vector-id -> offset lookup - a whole new index component, to save ~90 bytes per vector while paying a seek per resolved neighbour. So: remove it, at the level that matters.

- **`VectorLocationIndex` loses its bounded backend.** It carried two implementations and only one was reachable, so ten methods branched on `maxSize` and one of them held a `Collections.synchronizedMap` monitor that `countEntries()` and `getStats()` serialized on for nothing. There is now a single `ConcurrentHashMap`-backed implementation and no `maxSize`, which makes the truncation structurally impossible. The single-int constructor now means an initial capacity; the old `-1` "unlimited" sentinel is rejected rather than silently read as a capacity.
- **The two user-facing entrances refuse a positive value.** `CREATE INDEX ... METADATA {"locationCacheSize": N}` and `TypeLSMVectorIndexBuilder.withLocationCacheSize(N)` (now `@Deprecated`) raise, naming the reason and the ~90 bytes per live vector to size against. `locationCacheSize` stays in `USER_METADATA_KEYS` on purpose - that set only decides whether a key is *recognised*, so dropping it would turn the explanation into a bare "unknown metadata key".
- **The two paths that must not fail still tolerate it**, and warn once per index: the global setting, so an existing startup line cannot stop a server booting, and a schema written by an older version, so an existing database still opens.

## Breaking change

`-1`/`0` ("no limit", what an unset builder and a metadata copy carry) stay accepted, so nothing that left the setting alone is affected. A `CREATE INDEX` script that sets a positive `locationCacheSize` now fails instead of being silently ignored, and the key has to be removed. The release-notes section for #5568 has been rewritten to say so.

## Tests

New `Issue5559LocationCacheSizeTest` (4 tests) pins all four behaviours: the METADATA refusal (and that `-1` is still accepted), the builder refusal, that a persisted definition carrying the key still loads, and - the behavioural guard - that the global setting truncates nothing, with self-recall probes across the whole insertion range because `countEntries()` alone would still pass on an index whose graph had been built over a truncated set.

**Mutation-checked twice**, since the bug no longer reproduces on `main`: making `getLocationCacheSize()` return the configured value, and re-adding FIFO eviction to the new `ConcurrentHashMap` backend, both drive `countEntries()` from 200 down to the cap and fail the test.

`VectorLocationIndexRidReverseTest` swaps its "FIFO eviction keeps the reverse index in sync" case for the opposite invariant - nothing is evicted past the capacity hint. `VectorLocationIndexConcurrencyTest` drops the two bounded-mode cases (#4515's failure mode has nothing left to construct) and gains one for the constructor's changed meaning.

254 tests in `com.arcadedb.index.vector` and 684 in `com.arcadedb.index`/`com.arcadedb.schema`/`LowRamProfileTest` pass.

https://claude.ai/code/session_01KJ6aLGLNwsPTYejUxsYGiS


lvca:
On the HTTP 500 finding: I checked it and the premise does not hold — the refusal **already answers 400**.

The reasoning was that `IndexException` reaches `AbstractServerHttpHandler` via the auto-commit transaction wrapper, where no client-error arm matches. But the SQL DDL path never gets that far: `applyUserMetadata` runs during statement parsing, and the parser wraps the failure. Measured exception chain for `CREATE INDEX ... METADATA {"locationCacheSize": 10}`:

```
com.arcadedb.exception.CommandSQLParsingException
  caused by com.arcadedb.index.IndexException
```

`CommandSQLParsingException` is exactly what the handler's 400 arm keys on (both directly and unwrapped from a cause, `AbstractServerHttpHandler:528` and `:610`). A new server IT confirms it end to end: **400**, with `locationCacheSize` in the body.

So I have **not** changed the exception type. Doing so would also desynchronise `setLocationCacheSize` from its sibling setters `setSimilarity`/`setQuantization` in the same class, for no behavioural gain.

The concern underneath the finding was fair, though — nothing pinned the status code, and it rests on a wrap two layers away from the throw, so a future change to how DDL wraps a metadata failure would silently turn this into a 500 carrying a helpful message nobody displays. Added in `e54829673`:

- `server/.../Issue5559LocationCacheSizeHttpIT` — asserts 400 for **both** rejection arms of the key (the refusal, and the out-of-int-range value #5639 already rejected via `metadataInt`'s `IllegalArgumentException`, which reaches the handler by a different route and must not disagree), and 200 for `-1`.
- The engine test now asserts the exception *type* the 400 depends on, not just the message.

Thanks — the finding was wrong but it pointed at a real untested contract.

Issue #3715, opened 2026-03-23
Cannot invoke "io.github.jbellis.jvector.vector.types.VectorFloat.length()" because "b" is null


lekmaneb:
Hi,

Error while performing vector search that only occurs once but then works 🤨 

Arcade logs:
```text
2026-03-23 16:22:35.095 INFO  [LSMVectorIndex] Graph build from pages: 5499 total entries, 270 deleted, 4987 active for graph<ArcadeDB_0> Error performing vector search
java.lang.RuntimeException: java.lang.NullPointerException
        at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:137)
        at com.arcadedb.index.vector.LSMVectorIndex.findNeighborsFromVector(LSMVectorIndex.java:2529)
        at com.arcadedb.index.vector.LSMVectorIndex.findNeighborsFromVector(LSMVectorIndex.java:2445)
        at com.arcadedb.function.sql.vector.SQLFunctionVectorNeighbors.executeWithLSMVectorIndexes(SQLFunctionVectorNeighbors.java:146)
        at com.arcadedb.function.sql.vector.SQLFunctionVectorNeighbors.executeWithTypeIndex(SQLFunctionVectorNeighbors.java:129)
        at com.arcadedb.function.sql.vector.SQLFunctionVectorNeighbors.execute(SQLFunctionVectorNeighbors.java:99)
        at com.arcadedb.query.sql.parser.FunctionCall.execute(FunctionCall.java:131)
        at com.arcadedb.query.sql.parser.FunctionCall.execute(FunctionCall.java:84)
        at com.arcadedb.query.sql.parser.LevelZeroIdentifier.execute(LevelZeroIdentifier.java:66)
        at com.arcadedb.query.sql.parser.BaseIdentifier.execute(BaseIdentifier.java:66)
        at com.arcadedb.query.sql.parser.BaseExpression.execute(BaseExpression.java:167)
        at com.arcadedb.query.sql.parser.Expression.execute(Expression.java:87)
        at com.arcadedb.query.sql.parser.ProjectionItem.execute(ProjectionItem.java:136)
        at com.arcadedb.query.sql.parser.Projection.calculateSingle(Projection.java:148)
        at com.arcadedb.query.sql.executor.ProjectionCalculationStep.calculateProjections(ProjectionCalculationStep.java:66)
[...]

Issue #5318, opened 2026-07-17
LSMVectorIndex.remove() scans all vector ids per call: any record update on a vector-indexed type is O(index size), bulk edge-connect degrades quadratically


tae898:
### Summary

`LSMVectorIndex.remove(Object[], Identifiable)` is O(index size): it materializes **all** vector ids and linearly scans them to find the entries matching one RID. Because `DocumentIndexer.updateDocument()` fires this on **every** record update of a vector-indexed type (even when the indexed embedding property is unchanged), any bulk operation that rewrites vertex records of such a type degrades quadratically.

We hit it with `GraphBatch` bidirectional edges: the deferred incoming-edge connect updates each destination vertex once to attach the IN chunk pointer. With ~213k vectors in the index and ~210k vertex updates, the connect phase ran for **hours at ~10-60 edges/sec** (measured both contended and solo), fully CPU-bound. The identical graph without a vector index on the vertex type connects at **>500k edges/sec**.

### Environment

- 26.8.1-SNAPSHOT (current dev line, post #5306/#5307), embedded, Java 25 (bundled JRE)
- Corpus: 213,761 vertices of type `Question` with `embedding float[384]` + `LSM_VECTOR` index (M=16, efC=100), plus `Answer`/`User` types; 210,226 edges connected into `Question` via `GraphBatch` with bidirectional (default) edges

### Evidence

Log line from the connect phase (earlier, contended run that did complete):

```
GraphBatch: Connecting 210226 deferred incoming edges...
GraphBatch: Incoming edges connected: 210226 edges in 20068592.8 ms (10 edges/sec)
```

Thread dump captured mid-crawl on a solo run (one async slot shown; several identical; the thread had >20 minutes of CPU):

```
"AsyncExecutor-db-1" RUNNABLE cpu=1210211.33ms
  at java.util.stream.IntPipeline.toArray
  at com.arcadedb.index.vector.LSMVectorIndex.remove(LSMVectorIndex.java:3660)
  at com.arcadedb.database.DocumentIndexer.updateDocument(DocumentIndexer.java:635)
  at com.arcadedb.database.LocalDatabase.updateRecordNoLock(LocalDatabase.java:1154)
  at com.arcadedb.database.TransactionContext.commit1stPhase(TransactionContext.java:980)
  ...
  at com.arcadedb.database.async.DatabaseAsyncExecutorImpl$AsyncThread.run
```

[...]

Issue #4581, opened 2026-06-12
`findNeighborsFromVector` brute-force fallback reads `ordinalToVectorId` without snapshot


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.index.vector.LSMVectorIndex`
**Severity:** MEDIUM

### Summary

`ordinalToVectorId` is a `volatile` field that the rebuild path
reassigns wholesale. The brute-force scan re-reads
`ordinalToVectorId.length` inside the loop bound and the array
reference inside the body, but `vectors` (`ArcadePageVectorValues`) was
constructed earlier with a different snapshot. A rebuild reassigning
the field during the loop makes `ordinal` index into the new mapping
while `vectors.getVector(ordinal)` looks up the old.

### Code reference

`engine/com/arcadedb/index/vector/LSMVectorIndex.java:2713–2723, 2947–2956`

### Impact

Brute-force fallback returns RIDs paired with scores computed from a
different vector — wrong scores, wrong ranking, no error.

### Suggested fix

Snapshot `final int[] ord = this.ordinalToVectorId;` once outside the
loop and use that consistent reference for both bounds and lookups.

lvca:
Fixed

PR #5739, opened 2026-08-01
fix(engine) #5558: a vector search inside a deleted region finds the survivors instead of nothing

### LINKED ISSUE #5558: LSM_VECTOR: a search whose query vector sits in a deleted region returns an empty neighbor list
**Verified on `main` (no local changes), found while working on #5516.**

A `vector.neighbors` search returns `[]` - not "the nearest surviving vectors" - when the query vector lands where a group of deleted vectors used to be. The live vectors are still there, still searchable with any other query.

### Reproduction

Vertex type with an `LSM_VECTOR` index (16 dimensions, COSINE, INT8), 100 vertices `doc0..doc99` with well-separated embeddings, then delete `doc0..doc29`:

```java
database.transaction(() -> {
  for (int i = 0; i < 30; i++)
    database.command("sql", "DELETE FROM Doc WHERE id = ?", "doc" + i);
});
```

- query with the embedding of `doc50` (alive, far from the deleted block) -> 5 neighbors, `doc50` first. Correct.
- query with the embedding of `doc3` (deleted, its neighbors `doc0..doc6` all deleted) -> **`[]`**

Reproduces both before and after a reopen, so it is not a persistence problem: the entry count is right (70 live), and every other query works.

### Why it looks like a real defect

The graph walk lands in a neighborhood whose nodes are all filtered out as deleted and returns nothing rather than expanding the search to the surviving nodes. So the answer is silently wrong - an empty result reads as "there is nothing similar in this database", when there are 70 live vectors, some of them reasonably close.

The impact scales with the deletion pattern: an application that deletes a topical cluster (a tenant, a document set, a time range) leaves a hole where queries about that topic return nothing at all, which is exactly the query most likely to be asked.

### Not caused by, but adjacent to, #5516

Found while verifying the tombstone-memory fix. It reproduces on pristine `main`, so it predates that work and is out of its scope. Filing separately so it does not get lost.

lvca:
Fixes #5558.

## The report, reproduced

A `vector.neighbors` / `select().vectorNeighbors()` search whose query vector lands where a block of deleted vectors used to be answers `[]` - not "the nearest surviving vectors" - while the live vectors are still there and still searchable with any other query.

Reproduced verbatim on unmodified `main`, on the PQ path (1200 vectors in 12 clusters, 400 deleted, query aimed at a deleted cluster):

```
theApproximatePqSearchAnswersADeletedRegionQuery
[the PQ path has nothing to fall back on: an empty answer here is the reported defect]
Expected size: 5 but was: 0
```

## Why

A delete does not take the vector out of the HNSW graph. It tombstones the location and leaves the node in place until the next graph rebuild, which for any index above `ASYNC_REBUILD_MIN_GRAPH_SIZE` (1000) is `max(100, 0.2 * graphSize)` mutations away. Everything below is about what a query does when it walks into that hole. Two independent causes:

**1. The traversal was told every node was an acceptable answer.** `findNeighborsFromVector` and `findNeighborsFromVectorApproximate` passed `Bits.ALL` whenever no RID allow-list was given. JVector's `GraphSearcher.stopSearch` stops as soon as `approximateResults` holds `rerankK` *acceptable* candidates, so a beam that filled with tombstones declared itself finished; the post-filter then dropped them one by one and the caller got a short list, or an empty one.

**2. A tombstone was scored through a placeholder vector.** `ArcadePageVectorValues` hands back a sentinel rather than null for a deleted ordinal (#3715), and the sentinel was a vector of `Float.MIN_NORMAL`s "to give very low similarity scores". It does not: cosine cancels the magnitude out, and the squared magnitude `dimensions * MIN_NORMAL^2` underflows to zero in float, so `VectorUtil.cosine` came back `Infinity`. Every tombstone was therefore the **best** candidate in the beam, displacing the real neighbours - and on a JVM with assertions enabled the search failed outright:

```
java.lang.AssertionError: cosine([0.9928, 0.1197, ...], [1.1754944E-38, 1.1754944E-38, ...]) = Infinity
```

5 of the 7 new tests fail that way on `main`.

## The fix

- **`RIDBitsFilter` -> `LiveVectorBitsFilter`**, with a nullable allow-list, and passed on every search rather than only when RIDs are restricted. Its predicate is deliberately identical to the post-filter already applied to the search output, so nothing that would have survived is dropped. Tombstones are still *traversed* - JVector expands a rejected node, it just does not collect it - so the vectors behind the hole stay reachable and the beam keeps walking until it has found them.
- **`liveOnlyScoreFunction`** on the exact and grouped paths: a node whose vector cannot be read (the sentinel from `ArcadePageVectorValues`, `null` from `GrowableVectorValues`) is scored at `0`, the floor of every JVector similarity function, instead of being handed to `compare()`. The PQ path does not need it - it scores from PQ codes, which stay valid for navigating past a tombstoned ordinal - and gains only the `LiveVectorBitsFilter`.
- The sentinel itself is now filled with `1.0f` and documented as a shape placeholder, not a score, so any path that scores it without asking gets a finite number rather than `Infinity`.
- **The brute-force fallback** sizes its expectation on the live vector count rather than on the graph's ordinal count (a tombstone-heavy graph carries ordinals no query can ever return), and drops the `results.size() < available * 8 / 10` term. That term was a no-op except for tiny live sets, where it evaluated to `0 < 0` and suppressed the fallback that was the only thing left able to answer the query. The scan also now skips a live-but-unreadable ordinal instead of scoring its placeholder, which under COSINE gave it a distance of minus infinity, i.e. first place.

## Cost of the always-on filter

The filter now runs on every search, so it was worth pricing on a delete-free index. JVector calls `acceptOrds.get()` once per **popped** candidate, not once per scored neighbour, so it does not scale with beam width times fan-out. On a 50k-vector, 128-dimension INT8 index at `k=10`: **28.7 calls per query**, **7.5 ns** per `getLocation` lookup, **0.22 us** added to a **172 us** query - **0.13%**, below the benchmark's own noise. Recorded in the class javadoc.

## Considered and rejected

Removing the node from the graph on delete, or keeping its location resident so the real vector stays scoreable for navigation. Both are real designs - they are what makes a stale graph navigate *well* - but the first costs a graph rewrite per delete and the second gives back the memory #5516 reclaimed. Neither is needed for correctness: the surviving neighbours of a deleted cluster are reachable from outside it, and with tombstones scored at the floor the beam converges on exactly them.

A `graphIndex.size() == vectorIndex.size()` fast path back to `Bits.ALL` for clean indexes: 0.13% is not worth a second definition of which nodes a search may return, and that equality is not a sound "no tombstones" test anyway - `vectorIndex.size()` counts live vectors *including* delta ones not yet in the graph, so 30 tombstones plus 40 pending inserts satisfies it.

## Tests

`engine/src/test/java/com/arcadedb/index/vector/Issue5558DeletedRegionSearchTest.java`, 7 tests. The fixture is 12 clusters on an arc, embedded with damped harmonics so cosine similarity is a strictly decreasing function of the angular gap - the nearest surviving cluster to a query aimed at a deleted one is decidable on paper, so the assertion is "cluster 4", not "something". The rebuild threshold is held above the workload so the graph keeps its tombstones for the whole test, which is the steady state of any index whose delete volume sits under the threshold.

Covers the plain path, the SQL function, the grouped path, the PQ path, a query away from the hole, a reopen, and the single-survivor case.

- 5/7 fail on `main` with the `AssertionError`; the PQ one fails with the reported empty list with assertions off.
- Whole class runs in 0.498 s (slowest test 0.128 s), so it stays untagged and in the default CI run.
- Full `engine` suite after the fix: **10867 tests, 0 failures, 0 errors**.

## Follow-up, out of scope here

`findNeighborsFromVectorGrouped` spends its distinct-group budget in traversal order rather than score order, so a grouped query returns whichever groups the beam met on its way in - a query at cluster 1 comes back with clusters 6, 9 and 10, with no deletes involved at all. The `groupBy` test here therefore asserts liveness rather than which cluster wins. Filed separately.

https://claude.ai/code/session_01LXn1UsWXeteQqMvq4utf6A


lvca:
Thanks - all four points addressed. Pushed 71017242c.

### 1. Hot-path cost of the always-on filter — measured, no fast path added

The premise that it "scales with beam width x graph fan-out" does not hold. JVector calls `acceptOrds.get()` on the **popped** candidate only (`GraphSearcher.searchOneLayer`: `if (acceptOrdsThisLayer.get(topCandidateNode) && topCandidateScore >= threshold)`), not on the neighbours it scores and pushes. So the filter runs once per expansion, while scoring runs once per visited neighbour - roughly a factor of the graph degree apart.

Measured on a 50k-vector, 128-dimension INT8 index at `k=10`, with the filter instrumented:

| | |
|---|---|
| `get()` calls per query | **28.7** |
| `getLocation()` cost | **7.5 ns** |
| added per query | **0.22 us** |
| query latency | **172 us** |
| **overhead** | **0.13%** |

That is below the run-to-run noise of the benchmark itself. I have not added the `graphIndex.size() == vectorIndex.size()` fast path: it would buy nothing measurable, and `vectorIndex.size()` is the live count *including* delta vectors not yet in the graph, so that equality is not actually a sound "no tombstones" test - an index with 30 tombstones and 40 pending inserts satisfies it. Getting it right would need new state tracked across rebuilds, to save 0.13%.

The numbers and the reasoning are now in the `LiveVectorBitsFilter` class javadoc, so the missing fast path reads as a decision rather than an oversight.

### 2. `vectorIndexSnapshot` naming — fixed

Renamed to `vectorIndex`, with a field comment saying it is deliberately the live map read at traversal time, precisely so the filter and the post-filter cannot disagree. You are right that the shared `Snapshot` suffix with `ordinalToVectorIdSnapshot`, which genuinely is one, invited the wrong assumption.

### 3. PR description nit — fixed

Correct, and the description now says so: the PQ path gains only the `LiveVectorBitsFilter`. It scores from PQ codes, which stay valid for navigation on a tombstoned ordinal, so there is nothing for `liveOnlyScoreFunction` to do there. `liveOnlyScoreFunction` is on the exact and grouped paths.

### 4. `@Tag("slow")` — declined, with the numbers

The vectors are 16-dimensional, so 1200 of them is a small fixture. Whole class: **0.498 s**, slowest single test **0.128 s** (`theApproximatePqSearchAnswersADeletedRegionQuery`, the one that trains PQ). Per CLAUDE.md that is not "noticeably long", and tagging it would take real coverage out of the default CI run.

```
0.128s  theApproximatePqSearchAnswersADeletedRegionQuery
0.106s  theDeletedRegionStaysSearchableAfterAReopen
0.062s  aQueryAgainstASingleSurvivorFindsIt
0.050s  theGroupedSearchAnswersADeletedRegionQuery
0.045s  theSqlFunctionAnswersADeletedRegionQuery
0.041s  aQueryInsideADeletedRegionReturnsTheNearestSurvivingCluster
0.040s  aQueryAwayFromTheDeletedRegionIsUnaffected
```

Full `engine` suite still green after the follow-ups: 10867 tests, 0 failures.

lvca:
All three round-2 points applied, in `f0cdda2be` and `c59635b34`. The third one turned up something that was worth more than the point itself.

### 3. Non-cosine coverage — and a hole it exposed

Added `aDeletedRegionQueryUnderEuclidean` and `aDeletedRegionQueryUnderDotProduct`. The arc embedding has the same magnitude at every vertex, so both metrics order the clusters exactly as cosine does and the expected answer is unchanged; the DOT_PRODUCT one scales the fixture onto the unit sphere, which is JVector's documented precondition for that metric and the condition under which its floor really is `0`. I have also corrected the `UNREADABLE_NODE_SCORE` javadoc, which claimed `(1 + dot) / 2` bottoms out at 0 without saying that only holds for unit-length vectors — with the un-normalized fixture a real vector scores down to -0.26, i.e. *below* the tombstone floor.

Both new tests **pass on `main`**, which is correct and worth stating plainly: the `Infinity` blow-up is specific to cosine's magnitude cancellation, so those two metrics never had the defect. They are characterization tests, not regression tests.

That prompted a mutation check, and it failed:

> **Setting `UNREADABLE_NODE_SCORE = 1.0f` — the worst possible value, putting every tombstone ahead of every real vector in the beam — left all 9 tests passing.**

The `Bits` filter was carrying the entire suite. The score half of the fix had no test at all. It cannot be reached end-to-end either: `LiveVectorBitsFilter` keeps a tombstone out of the answer whatever it scored, and the brute-force fallback repairs a short result, so the two halves mask each other by construction.

So `anUnreadableNodeScoresTheFloorOfItsMetric` now asserts the score directly, over all three metrics:

- an unreadable node scores exactly the floor (`Infinity` was the defect; any finite-but-high value is the same defect more quietly — it still wastes the beam budget on tombstones),
- a real vector scores a finite number,
- and the real score is strictly above the floor — the property that lets one constant serve all three metrics.

It first asserts that a tombstoned *and* a live ordinal were actually found in the graph, so it cannot pass vacuously. Mutants `1.0f` and `0.5f` both die.

To make the hook honest I extracted `searchVectorValues()` — the vector-values construction the exact and grouped paths had copied between them — so the test scores through the same object a query builds rather than a rebuilt approximation of it. `liveOnlyScoreFunction` is package-private with two narrow test hooks.

### 1. Bounded backend under-estimating `expectedResults` — acknowledged

Correct, and confirmed against `VectorLocationIndex:510-514`. Comment added at the `Math.min` noting that `size()` is the live count only while the map does not evict, so a bounded backend makes it a lower bound and the fallback may not fire when it should — acceptable because an evicting location map already makes the index approximate, and it is the same assumption the auto-compaction ratio rests on. `#5568` means nothing asks for that backend today.

### 2. Coupling to JVector's traversal semantics — documented

Good catch, this is the kind of thing that rots silently. `LiveVectorBitsFilter`'s javadoc now names it: `GraphSearcher.searchOneLayer` consults `acceptOrds` only to decide whether a *popped* node joins the result heap, and expands its neighbours either way. An upgrade that pruned the neighbours of a rejected node would make the vectors behind a deleted region unreachable again — with this filter as the cause — and it points the future upgrader at the test that will catch it.

---

Now 10 tests. Full `engine` suite green; index/select/vector-function suites 1446 tests, 0 failures.

lvca:
Fixed in `82162969a`.

### 1. Orphaned javadoc — fixed

Exactly right, and thanks for catching it: the floor-score test landed *between* that block and the methods it described, so Java attached only the nearer one and the first was dead comment while `aDeletedRegionQueryUnderEuclidean` / `aDeletedRegionQueryUnderDotProduct` had none at all.

Moved back down, and while rewriting it I made it say what those two tests actually are - characterization tests for a non-cosine index, which **pass on `main`**, because the `Infinity` blow-up came out of cosine cancelling the magnitude and never reached those metrics. The block now points at `anUnreadableNodeScoresTheFloorOfItsMetric` as the thing that pins the floor.

### 2. Live-builder mode — checked, and the answer is stronger than "presumably safe"

**That branch is unreachable.** `graphIndex` is assigned in exactly four places: a graph loaded from disk (`:1302`), `null` (`:1900`), `builtGraph` from `buildGraphFromScratch` (`:2108`), and an on-disk graph (`:2201`). `buildGraphFromScratch` builds through a **local** `GraphIndexBuilder` (`:1992`, try-with-resources), not `liveBuilder`. Nothing ever assigns `liveBuilder.getGraph()` to `graphIndex`, so `graphIndex == liveBuilder.getGraph()` cannot hold and the `liveVectorValues` branch never runs.

That is the answer to your question, and it is the reassuring one: the `ordinalMap` a search hands the filter is *always* the map published alongside the graph by the same rebuild - the pairing #4581 exists to keep intact - and never a map whose ordinals mean something else. I have written that into `searchVectorValues()`'s javadoc, since it is the invariant the always-on filter depends on and it is not obvious from the call site.

I have left the dead branch in place rather than deleting it: it is a cheap guard, removing it is unrelated to this fix, and the identity check is exactly the right condition if the live builder's graph is ever published.

Worth adding for completeness: even if that branch *were* reachable with a stale map, the failure mode would be a slow query, not a wrong answer. The filter's predicate is identical to the post-filter's, so a mapping that made the filter reject everything would already have made the post-filter drop everything on `main`, and the brute-force fallback would answer either way.

### 3 & 4 — agreed, no action

Confirming both readings. The DOT_PRODUCT residual is cosmetic in exactly the way you describe: a non-normalized index can have a real vector score below the tombstone floor, which costs a little beam budget and cannot cost correctness, because the `Bits` filter is what decides membership. And the live-read-plus-snapshot pairing in `get()` is deliberate and now carries the comment saying so.

---

261 vector tests green after the change; the class is at 10 tests.

lvca:
Pushed `17de5c3f2`. Point 1 was real and is downstream of this PR; point 3 I tried to fix, measured, and reverted; point 2 acknowledged. The detour turned up something about the test suite worth reporting.

### 1. PQ train/encode loops — fixed, and it was this PR's to fix

Agreed it is pre-existing in *shape*, but the sentinel's **value** changed here, so the consequence is new: a live-but-unreadable ordinal used to be PQ-encoded as a near-zero vector and would now be encoded as an all-ones one — a code with a real direction, which the approximate search then scores as if it meant something.

Both loops now call `MutablePQVectors.setZero(i)` for a `null` or sentinel vector. That is JVector's own "no information here" primitive and, unlike skipping the ordinal, it keeps the code array dense — skipping would have left the first-ever gap in it, since `ArcadePageVectorValues.getVector` never returns `null`.

### 3. Fallback firing more readily — measured, and I was wrong about why

I first "fixed" this by capping `expectedResults` at `allowedRIDs.size()`, reasoning that a query restricted to 10 RIDs asking for 20 pays a full scan just to rediscover there are only 10. **The test disproved it.** With a 3-RID allow-list the graph walk returns **2**, not 3 — a restrictive `Bits` filter makes JVector's local-maximum heuristic stop expanding early — so the fallback is what *completes* the answer. It is doing useful work on exactly the workload you flagged, not wasteful rediscovery.

So the cap is reverted, and the plain answer to your question stands: **no latency regression.** The old `8/10` term was true in these cases too, so removing it changed nothing here.

What survives is a **`bruteForceScans` metric**, worth having on its own merits — a full ordinal-map scan per query is the most expensive thing this index does and it was previously only a `WARNING` log line. It is in `getStats()`.

### The thing worth reporting: each half alone keeps the suite green

Having a way to observe the fallback let me mutation-test the two halves against each other, and the result is not what the PR description implies:

| mutation | result |
|---|---|
| `UNREADABLE_NODE_SCORE = 1.0f` | all end-to-end tests **pass** (found in round 2) |
| `Bits.ALL` restored on the **exact** path | all tests **pass** |
| `Bits.ALL` restored on the **PQ** path | `theApproximatePqSearchAnswersADeletedRegionQuery` **fails** — the reported empty list |

On this fixture **either half alone is sufficient**: with tombstones scored at the floor they never reach the top of the candidate queue, so `Bits.ALL` is survivable; with the filter on, the score is irrelevant. They mask each other in both directions.

So the real coverage map is narrower than "7 tests cover the fix":

- the **accept-filter** is pinned *only* by the PQ test — that is the one path scoring tombstones from real PQ codes, where a floor score cannot substitute;
- the **floor score** is pinned *only* by `anUnreadableNodeScoresTheFloorOfItsMetric`;
- everything else pins that a search over a tombstoned graph returns the right vectors, without saying which half got it there.

That map is now in the class javadoc with the mutations that established it, so removing either half fails something instead of looking safe.

The deleted-region test also got a stronger assertion out of the new metric: it now requires `bruteForceScans == 0`, i.e. the beam walked out of the hole *by itself* rather than being rescued by a full scan — which is what happens on `main`.

### 2. Bounded-backend budget — on record

Confirmed, and deliberate. The comment at the `Math.min` says it; noting here that it is a conscious trade rather than an oversight.

---

12 tests in the class, 262 in `com.arcadedb.index.vector`, 1447 across index/select/vector-function. All green.

lvca:
Pushed `fa9a5209d`. Points 1, 2, 4 and the style nit applied; 3 needs no action.

### 1. The sentinel change rests on the `isDeletedSentinel` identity guard — warned off

This is the sharpest observation in the review and I have taken it as written. You are right that the failure mode got *quieter*, not milder: with the placeholder finite, a scoring path that stops recognising it now produces a plausible cosine score instead of an `Infinity` that fails loudly under assertions. And the guard is total only while `copy()` returns `this` — a real copy carries a different per-instance placeholder, so every caller scoring through it silently stops recognising one.

`copy()` now says exactly that, naming the issue and why the consequence is invisible. It is precisely the "harmless optimisation" a future maintainer could make without seeing the coupling.

### 2. `bruteForceScans == 0L` and future flakiness — kept, signposted

Agreed on the analysis, and I have kept the assertion: it is the only thing in the class that distinguishes a working search from one rescued by the fallback, which is the whole point of the fix. But you are right that it rides on HNSW recall rather than on a decidable property, so I have written the triage into the test:

> If it ever goes red on its own, check the assertion above first - while the returned cluster is still 4, this is recall drift and not a regression, and the fix to make is the fixture (a wider cluster gap or a larger k), not the search.

That turns a future mystery flake into a signposted one.

### 4. `availableVectors` counting delta vectors the scan cannot reach — documented

Good catch, and a real (if bounded) inefficiency I had not spelled out: `vectorIndex.size()` includes vectors still in the delta buffer, which `bruteForceScan` cannot return because it walks graph ordinals, and which `mergeWithDeltaScan` has already covered. So a search left short *only* by delta vectors can pay for a scan that finds nothing new. The comment now says so, and as you note the new counter makes it observable rather than theoretical.

### 3. Test-only hooks — no action

Noted and agreed it is the one place test concerns touch `LSMVectorIndex`. I kept them because the floor score genuinely cannot be observed end-to-end — that is the finding from the previous round, where `UNREADABLE_NODE_SCORE = 1.0f` left every end-to-end test green.

### Style nit

Fixed — `ResultSet` was already imported in that file, so the fully-qualified name was pure oversight. Per CLAUDE.md that is a convention violation, not a preference.

---

262 vector tests green. On the note about not running the full suite locally: the `10867 tests, 0 failures` figure is from `mvn -pl engine test` on the fix as of `17de5c3f2`; this push is comments, one test import, and one comment line in `LSMVectorIndex`, with the vector suite re-run.

lvca:
Thanks for the approve. Pushed `67a2fb4fa` with points 2 and 3; 1 and 4 need no change.

### 2. Snapshot consistency on the PQ path — fixed

Confirmed and worth doing. `findNeighborsFromVectorApproximate` read the volatile `ordinalToVectorId` twice — once to build the filter, once in the result loop (`ordinalToVectorId.length`, then `ordinalToVectorId[ordinal]`, themselves two more reads) — so a rebuild landing between them would resolve an ordinal through one array in the filter and a different one when pairing it with a RID. It now captures one local and threads it through both, exactly as the exact path does.

You are right that it is pre-existing and that the read lock likely excludes the swap in practice, but the filter line was already being touched here and #4581 discipline should not be path-dependent.

### 3. `bruteForceScans` scope — documented

Correct that it is only the plain k-NN path, and that this matches the metric's meaning rather than being a gap. Added a clause saying so, since "brute force scans: 0" could otherwise read as "no path ever scanned" rather than "the only path with a fallback never needed it".

### 1 & 4 — no action

The DOT_PRODUCT floor genuinely rests on a precondition JVector documents but does not enforce, and I do not think this PR should start enforcing normalization. Worth restating the blast radius, which is as narrow as you describe: it needs an index that is non-unit **and** DOT_PRODUCT **and** tombstone-heavy, and the consequence is a tombstone wasting beam budget, never entering a result — the filter is what decides membership.

On the dead-branch-as-guard: agreed, and its javadoc carries the reason it is retained.

### On verification

Worth being straight about this rather than just restating a number. My first full-suite run on this change came back with 3 errors in `LSMVectorIndexRebuildTest` — all inactivity-timer tests timing out on their 1-minute awaitility windows. That was **my** fault, not the change's: I had other Maven jobs running against the same machine, and the class took **117.3 s** instead of its usual ~24 s.

Ruled out properly before reporting anything:

- all 3 pass in isolation (24.5 s);
- all 3 pass in the same JVM as `Issue5558DeletedRegionSearchTest`, which rules out its `@BeforeEach` config freeze leaking (that class freezes the inactivity timeout, so it was the obvious suspect);
- the earlier green full run already contained that class with the same freeze.

Re-run with nothing else on the machine: **10871 tests, 0 failures, 0 errors**, and `LSMVectorIndexRebuildTest` back at 23.88 s.

lvca:
Point 1 measured, point 2 documented in `6d2c89f34`.

### 1. Fallback frequency on a filtered workload — measured, and it is not a regression

You were right to press on this. 50,000 vectors, 128 dimensions, INT8, `k=10`, allow-list of 3 RIDs, best of 3 rounds of 300 queries:

| workload | latency | brute-force scans |
|---|---|---|
| unfiltered `k=10` | **95 us/query** | 0 |
| allow-list of 3, `k=10` | **26,236 us/query** | **900/900** |

276x, on every single query. That is bad — and it is **not this PR**. Restoring the pre-PR guard exactly (`availableVectors = ordinalMap.length`, plus the `results.size() < availableVectors * 8 / 10` term) and re-measuring the same workload:

**26,602 us/query, 900/900 scans.**

Identical within noise. The 80% term was true here as well, so it never suppressed this; if anything the new form is marginally cheaper, because `availableVectors` is now the live count rather than the ordinal count. So: no latency regression on filtered workloads, and the sanity-check you asked for is on record rather than argued.

Filed separately as **#5748**, because it is a real pre-existing pathology and the obvious fix is not obviously right. Capping the expectation at `allowedRIDs.size()` removes the scan — I tried exactly that during round 4 of this review and reverted it, because with a 3-RID allow-list the graph walk returns **2** of the 3 (a restrictive `Bits` filter makes JVector's local-maximum heuristic stop expanding early) and the fallback is what *completes* the answer. Cap it and that query silently returns 2. The issue lays out the four options; the promising one is scanning the allow-list rather than every ordinal, using the `ridToVectorIds` reverse index #5318 already maintains, which makes the fallback O(allow-list) instead of O(index).

Your observation that `bruteForceScans` is the right instrument is exactly how I found the numbers above — it is what made "the scan fires on every query" visible rather than inferred.

### 2. Sentinel identity vs. persisted vectors — documented

Good catch, and correct in both directions: the reference check only recognises a placeholder that never left the instance, so one persisted inline by `LSMVectorIndexGraphFile` under `storeVectorsInGraph` and read back would be a different object and would score as an ordinary vector. Moot today because a rebuild excludes deleted ordinals from the graph it persists — which is precisely the point worth writing down, since it means that path is held right by `LiveVectorBitsFilter` and the location map, not by the guard. The note now sits next to the guard.

### 3. Documented dead branch — no action

Agreed.

---

262 vector tests green after the note. The full engine suite was **10871 / 0 failures / 0 errors** as of `67a2fb4fa`; this push is a javadoc block.

lvca:
Added in `34af19548`. Thanks for the review series.

### 1. Why `LiveVectorBitsFilter` is not memoized where `GroupedRIDBitsFilter` is — noted

Good observation, and the asymmetry deserved the explanation rather than being left to look accidental. The reason the two differ is not caution level, it is that they are answering different kinds of question:

`GroupedRIDBitsFilter` **has** to memoize, because each verdict *consumes a group budget* — answering the same ordinal twice would count it against its group twice. This filter has no such state, so there is nothing to protect and a live read is strictly better: a delete committed mid-traversal takes effect immediately instead of being pinned by a cache.

And as you say, the output loop is what makes that safe — it re-applies the same predicate, so an ordinal admitted just before its delete lands is dropped at emit rather than returned. The javadoc now says all of that, specifically so nobody "aligns" the two filters and either loses the freshness or memoizes a stale admit.

### 2, 3, 4 — no action

All three read correctly to me:

- The DOT_PRODUCT floor is precondition-dependent and the javadoc says so; enforcing normalization is not this PR's business.
- The fallback trade is acknowledged in the comment and now measurable. Worth adding the number from the last round: on a filtered workload the fallback frequency is **unchanged** from before the PR (26,236 vs 26,602 us/query, 900/900 scans either way) — that pathology is real but pre-existing, and is now filed as **#5748**.
- The test hooks exist precisely because the floor score is unobservable end-to-end, which is the finding that produced them.

---

262 vector tests green; full engine suite **10871 / 0 failures / 0 errors** as of `67a2fb4fa`, with the two pushes since being javadoc only.

lvca:
Pushed `ad9d217fe`.

### 2. Duplicated liveness predicate — de-duplicated

Agreed, and it turned out to be free rather than a trade. The shared helper returns the **location** instead of a boolean, so `GroupedRIDBitsFilter` still gets the RID its group key needs without a second lookup, and its memoization stays where it belongs — on top of the shared predicate rather than instead of it. `admissibleLocation()` is now the single definition of "may this ordinal be answered with", used by both filters and matching the post-filter.

### 1. The placeholder value — one correction, and it is now written down

The concern is right that the failure got quieter, and the guard notes exist for exactly that. But the mechanism is worth correcting, because it changes what "a safer value" would even mean:

- `MIN_NORMAL` and `1.0f` fills differ **only in magnitude**, so under cosine they point the same way and would score *identically* if the old one had not underflowed. The swap turned a broken number into a defined one — it did not turn a low rank into a high one. Ranking behaviour on an unguarded path is unchanged from before this PR.
- `1.0f` in each component is not a score of `1.0`. The score is `cosine(query, all-ones)`, which is near the ceiling only for a query aligned with the all-ones direction and near the middle otherwise.
- More importantly, **no constant does better.** For any fixed vector `s` there are queries scoring it anywhere in range, so "a placeholder that scores toward the floor" does not exist. That is precisely why the floor is applied by the score function that recognises the placeholder, rather than sought in the vector — it is the reason the fix has the shape it does.

That reasoning is now in `createDeletedSentinelVector`'s javadoc, since this is the second review round to reach for a safer value and there isn't one.

### 3. `bruteForceScans == 0` and recall — agreed, no action

It is the one assertion here riding on recall rather than on a decidable property, which is why it carries the triage note saying what a red one means and that the fixture, not the search, is what to fix.

---

Full engine suite re-run clean after the refactor, since it was the first change to `GroupedRIDBitsFilter`: **10871 tests, 0 failures, 0 errors.**

lvca:
Pushed `7812f3b3d`. The main consideration is answered with a measurement below — short version: **it is not a behaviour change on a large index**, and the earlier round of this review reached the opposite conclusion on the same code, so it was worth settling with data rather than argument.

### The relaxed fallback trigger — measured, and it is not new pressure

The premise is that dropping the `8 / 10` term makes a 1-result shortfall trigger a full scan where it previously would not. The algebra says otherwise: with `availableVectors = ordinalMap.length`, the term `results.size() < availableVectors * 8 / 10` is false only when `results.size() >= 0.8 * N`. Since `results.size() <= k`, that requires `k >= 0.8 * N` — a *tiny* index. On any large index the old term was **already true whenever the result was short**, so the old code fired the scan on a k-1 shortfall too.

Measured, 50,000 vectors / 128 dimensions / INT8, 2000 queries per configuration:

| | scans (this PR) | scans (pre-PR guard) | short results |
|---|---|---|---|
| `k=10` | **0 / 2000** | **0 / 2000** | **0** |
| `k=50` | **0 / 2000** | **0 / 2000** | **0** |

Two things fall out. Identical fallback frequency, confirming no new full-scan pressure. And zero shortfalls at all — the "HNSW legitimately returns k-1 from time to time" case does not arise here, because the adaptive path already runs `searcher.resume(k, max(k * 10, 100))` with a wider beam when the first pass comes up short, well before the fallback is consulted. The fallback is a third line of defence, not the second.

This is also consistent with the filtered-workload measurement from the previous round (26,236 vs 26,602 us/query, 900/900 scans either way).

So the explicit decision you asked for: the trigger is **unchanged for every index where a full scan would hurt**, and changed only for the tiny-live-set case the old term was wrongly suppressing — which is the single-survivor defect. I have not added a "shortfall exceeds a threshold" middle ground, because there is no case for it to fix: it would only reintroduce the class of hole the `0 < 0` guard already demonstrated.

### Minor points

**Rebuild-gated PQ codes** — correct, and worth surfacing to operators rather than only to reviewers. Added to the release note: an index built before this release keeps the codes it has, so PQ navigation past an unreadable vector stays slightly degraded until the next rebuild, and results are unaffected either way because membership is decided by the live-vector filter, not the codes. The note also points at `bruteForceScans` as the first thing to check when filtered vector queries are slow.

**`bruteForceScans == 0` coupled to fixture ordering** — good catch, fixed. It now counts across the query under test only, instead of asserting an absolute across the whole fixture, so it no longer depends on whether the graph-building warm-up search fell back. It asserts what it claims.

**`vectorIndexSnapshot` naming in `GroupedRIDBitsFilter`** — renamed to `vectorIndex`, with the same note as its sibling. You are right that the suffix reads worse now that the two sit next to each other and share a predicate that reads the map live on purpose.

**Delta vectors inflating `availableVectors`** — known and commented, as you note.

---

632 tests across vector / select / vector-function green. Full engine suite was **10871 / 0 / 0** as of `ad9d217fe`; this push is a test-scoped assertion change, a field rename, and the release note.

lvca:
## CI on `7812f3b3d`: three red jobs, none of them this PR

The review asked for a maintainer double-check that CI is green on the final commit. It is not, so here is what is actually failing and why I do not believe any of it is this change. Both of the things the review could not verify locally are also confirmed below.

### `integration-tests` — the tests passed

Step-level breakdown of the job:

```
success  Run Integration Tests with Coverage
success  Check no test leaked a databases/ directory
failure  IT Tests Reporter          <-- the only failed step
success  Upload integration test coverage reports
```

The Maven step **succeeded**. What failed is the `dorny/test-reporter` action, which runs with `fail-on-error: true` and turns its own API problems into a red job. No test failed here.

### `unit-tests` — one concurrency test, and it passes locally

`com.arcadedb.database.Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent` — a page-contention test, in a package this PR does not touch. Locally on this branch: **7 tests, 0 failures**.

### `ha-integration-tests` — three Raft tests

`Issue5410AbandonedTicketReleaseIT`, `Issue5569SlotMergeDeleteRaftIT`, `RaftReplicationChangeSchemaIT`. All Raft replication, none of them vector.

### The baseline

`1449cf126` — the commit **this branch is based on** — fails the same three jobs on `main`, with different tests each time (`HttpRedMetricsIT` for integration, `TimeSeriesTagDictionaryTest` for unit). 7 of the last 8 `Java CI - test` runs on `main` are red. The failing test differs run to run, which is what a flaky baseline looks like rather than a specific breakage.

### Blast radius

```
$ git diff --name-only main...HEAD | grep -v index/vector/ | grep -v ^docs/
$                                     # (empty)
```

Every changed file is under `engine/src/main/java/com/arcadedb/index/vector/`, its test package, or `docs/`. Nothing this PR touches is reachable from a page-contention test, a Raft replication test, or an HTTP metrics test.

Local runs on this branch: full `engine` suite **10871 tests, 0 failures, 0 errors**; vector suite 262; index/select/vector-function 632.

### The two verifications the review flagged

- **`MutablePQVectors.setZero(int)` in jvector 4.0.0-rc.9** — confirmed present in both the sources jar (`MutablePQVectors.java:61`) and the published binary (`javap`: `public void setZero(int)`).
- **CI green on the final commit** — not green, for the three unrelated reasons above.

### Remaining review points

- **DOT_PRODUCT non-unit guard** — filed as **#5750**, per your suggestion to make it a follow-up. It is a foot-gun independent of this fix: picking `DOT_PRODUCT` without normalizing gives quietly worse rankings than `COSINE` would, which is the opposite of the reason to pick it. Recommended fix there is to warn during the graph build, which already reads every vector.
- **Test-hook surface** and **documentation density** — noted, no change, as you suggested.

Issue #4139, opened 2026-05-08
Manual index name is ignored when create an index


mdre:
When create an index with manual-index-name is not registered by the database.

`create index Abogado_uuid if not exists on Abogado (uuid) unique NULL_STRATEGY SKIP;`

<img width="1789" height="743" alt="Image" src="https://github.com/user-attachments/assets/490f9269-0e30-4b0c-8569-e17c29fe8739" />

<img width="1789" height="832" alt="Image" src="https://github.com/user-attachments/assets/7eb11867-86a2-4c7a-b5f2-e418c2683bfd" />

lvca:
Fixed

PR #5740, opened 2026-08-01
fix(engine) #5675: CREATE INDEX IF NOT EXISTS answers for the index asked for, not for any index on those properties

### LINKED ISSUE #5675: Index create regression
### ArcadeDB version
Reproduced on `arcadedata/arcadedb:26.7.3` and `arcadedata/arcadedb:26.8.1-SNAPSHOT`.
Not present on `26.4.1-SNAPSHOT` (which errors instead), so this is a behaviour
regression introduced after 26.4.1.
### Summary
When a property already carries a `NOTUNIQUE` index, `CREATE INDEX IF NOT EXISTS
... UNIQUE` on that same property returns success without creating a unique
constraint. The caller is told the unique index exists; duplicates are then
accepted. Issuing the same statement *without* `IF NOT EXISTS` correctly reports
the conflict, so the guard clause is suppressing an error the server already
knows how to raise.
`IF NOT EXISTS` appears to match on the indexed property set alone and to ignore
the requested uniqueness. It affects plain and `BY ITEM` indexes equally.
### Steps to reproduce
CREATE VERTEX TYPE T;
CREATE PROPERTY T.Scalar STRING;
CREATE INDEX ON T (Scalar) NOTUNIQUE;
-- reports success, but creates no unique constraint
CREATE INDEX IF NOT EXISTS ON T (Scalar) UNIQUE;
INSERT INTO T SET Tag = 'a', Scalar = 'x';
INSERT INTO T SET Tag = 'b', Scalar = 'x';   -- accepted
### Expected
Either the unique index is created and the second insert is rejected with
`DuplicatedKeyException`, or the statement fails because the existing index
conflicts with the requested definition.
### Actual
`CREATE INDEX IF NOT EXISTS ... UNIQUE` returns success. `SELECT FROM
schema:indexes` shows only non-unique indexes on `T`:
    T[Scalar]                 unique=false
    T_0_516462088483265       unique=false
The duplicate insert succeeds.
### Contrast: without the guard clause, the conflict is reported
CREATE INDEX ON T (Scalar) UNIQUE;
    {"error":"Cannot execute command",
     "exception":"com.arcadedb.exception.CommandExecutionException",
     "detail":"Index 'T[Scalar]' already exists"}
### Also affects BY ITEM
CREATE VERTEX TYPE L;
CREATE PROPERTY L.Items LIST OF STRING;
CREATE INDEX ON L (Items BY ITEM) NOTUNIQUE;
CREATE INDEX IF NOT EXISTS ON L (Items BY ITEM) UNIQUE;  -- reports success
INSERT INTO L SET Tag = 'a', Items = ['x'];
INSERT INTO L SET Tag = 'b', Items = ['x'];              -- accepted
### Not affected: the reverse direction
Creating `UNIQUE` first and then a guarded `NOTUNIQUE` leaves the unique
constraint intact and correctly rejects the duplicate.
### Behaviour on 26.4.1-SNAPSHOT
The guarded statement fails loudly rather than silently:
    {"error":"Error on transaction commit",
     "exception":"com.arcadedb.index.IndexException",
     "detail":"Error on creating index on type 'T', properties [Items by item]
               -> ... Index 'T[Itemsbyitem]' is not valid."}
### Impact
Schema-migration DDL commonly uses `CREATE INDEX IF NOT EXISTS` so it can be
re-applied idempotently. Under that pattern, tightening an index from
`NOTUNIQUE` to `UNIQUE` silently does nothing, and the application proceeds
believing a uniqueness constraint protects the data.

lvca:
Fixes #5675.

## The bug

`IF NOT EXISTS` matched a pre-existing index on the auto-derived index name. That name is built from the indexed property set alone, so it says nothing about what the index *does*:

```sql
CREATE INDEX ON T (Scalar) NOTUNIQUE;
CREATE INDEX IF NOT EXISTS ON T (Scalar) UNIQUE;   -- reported success, created no constraint
INSERT INTO T SET Scalar = 'x';
INSERT INTO T SET Scalar = 'x';                    -- accepted
```

Migration DDL uses `IF NOT EXISTS` precisely so it can be re-applied, so tightening an index from `NOTUNIQUE` to `UNIQUE` did nothing while the application carried on believing a uniqueness constraint protected its data.

## Two more problems found underneath

1. **`withIgnoreIfExists(true)` dropped the caller's index.** `TypeIndexBuilder.create()` reacted to a uniqueness mismatch by dropping the existing index and rebuilding it. That is not recoverable: when the stored data has duplicates the rebuild fails and the type is left with **no index at all**. `builderIgnoreIfExistsNeverDropsAConflictingIndex` in the new test reproduced exactly that before the fix. On an *inherited* index the same code took away the parent type's index, which is issue #4083.
2. **The kind of index was ignored too.** A `FULL_TEXT` index satisfied a request for a range index on the same properties, silently. And the null-strategy half of the condition was dead code: `existingTypeIndex.getNullStrategy() != null && existingTypeIndex.getNullStrategy() == null` is never true.

## The rule now

`IF NOT EXISTS` is satisfied only when the existing index provides what was asked for:

- same index kind, and
- a uniqueness constraint at least as strong.

So a `UNIQUE` index still satisfies a `NOTUNIQUE` request (it indexes the same keys, and is never weakened into a plain index), while a `NOTUNIQUE` index does not satisfy a `UNIQUE` request. Anything the existing index does not provide is reported, naming both definitions, as an `IllegalArgumentException` -> **HTTP 400**.

### Nothing is dropped implicitly

Refusal is the answer even where the request could technically be granted by rebuilding, because the rebuild is the unsafe part. A caller that genuinely means "make this the definition" opts in:

```java
database.getSchema().buildTypeIndex("T", new String[] { "Scalar" })
    .withType(Schema.INDEX_TYPE.LSM_TREE).withUnique(true)
    .withReplaceIfIncompatible(true)     // new
    .create();
```

and even then the drop happens as late as possible (after every remaining validation) and the previous definition — kind, uniqueness, null strategy, properties, metadata — is **restored** if the new index cannot be built. An inherited index is never replaced.

### Cypher stays Neo4j-compatible

`CREATE CONSTRAINT ... IS UNIQUE` and `... IS NODE KEY` are statements about the data rather than "create it if missing", so they are the callers that opt into the replacement. Neo4j keeps the range index and the constraint as two objects; ArcadeDB has one index per property set and a unique index covers both roles, so the upgrade is the equivalent end state and a Neo4j migration script that creates indexes and then constraints keeps working. Cypher's own `CREATE INDEX` does **not** opt in — it finds a unique index sufficient and leaves it alone rather than replacing it with a plain one, which closes a silent *weakening* of a constraint that existed before.

## Alternative considered

The issue offers "create the unique index" or "fail". Doing the first implicitly is what the code already did, and it is the thing that can lose an index — so this PR does the second by default and makes the first an explicit, restorable opt-in rather than a side effect of a guard clause.

## Tests

- `engine/.../Issue5675IndexIfNotExistsConflictTest` — 11 cases: the reported scalar and `BY ITEM` forms, the reverse direction staying a no-op, index-kind mismatch, manual index names, the builder never dropping a conflicting index, the Cypher constraint upgrade, and a **failed** constraint upgrade leaving the plain index in place (mutation-checked: reverting the restore fails this test).
- `server/.../Issue5675CreateIndexIfNotExistsHttpTest` — the 400 and its detail message over HTTP, plus the idempotent 200 with `created: false` so the refusal test cannot pass by refusing everything.

Full engine suite: 1237 test classes green. (`ACIDTransactionTest#indexCreationWhileAsyncMustFail` failed once while three engine suites ran concurrently on this machine and passes in isolation — a contention flake, unrelated: it kills the database mid-build and asserts on recovery.) Server index + HTTP suites green.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1660,3 +1660,66 @@ An index that already exists on disk is unaffected either way - the page size is
 a builder.
 
 **Full Changelog**: https://github.com/ArcadeData/arcadedb/compare/26.7.2...26.8.1
+
+## `CREATE INDEX IF NOT EXISTS` no longer reports success for an index that is not the one asked for
+
+`IF NOT EXISTS` matched a pre-existing index on the indexed property set alone. The property set is what the index
+name is derived from, and it says nothing about what the index does, so a `NOTUNIQUE` index answered "already there"
+to a request for a `UNIQUE` one:
+
+```sql
+CREATE INDEX ON T (Scalar) NOTUNIQUE;
+CREATE INDEX IF NOT EXISTS ON T (Scalar) UNIQUE;   -- reported success, created no constraint
+INSERT INTO T SET Scalar = 'x';
+INSERT INTO T SET Scalar = 'x';                    -- accepted
+```
+
+Schema-migration DDL uses `IF NOT EXISTS` precisely so it can be re-applied, so tightening an index from `NOTUNIQUE`
+to `UNIQUE` did nothing while the application carried on believing a uniqueness constraint protected its data. Issue
+#5675. The same statement without the guard reported the clash, which is what made the guard the thing suppressing an
+error the server already knew how to raise.
+
+`IF NOT EXISTS` is now satisfied only when the existing index provides what was asked for: the same index kind, and a
+uniqueness constraint at least as strong.
+
+- A `UNIQUE` index still satisfies a `NOTUNIQUE` request. It indexes exactly the same keys, so the statement stays the
+  no-op it was, and it is never weakened into a plain index.
+- A `NOTUNIQUE` index no longer satisfies a `UNIQUE` request. The statement is refused, naming both definitions.
+- An index of a different KIND on the same properties - a `FULL_TEXT` index where a range index was asked for, or the
+  reverse - is refused for the same reason. That one was silent too.
+
+### An existing index is never dropped implicitly any more
+
+The refusal is the answer even in the case the request could technically be granted by rebuilding. Underneath,
+`TypeIndexBuilder.create()` with `withIgnoreIfExists(true)` used to drop the existing index and rebuild it with the
+requested definition. That is not a recoverable operation: if the rebuild then fails on the data already stored - two
+records sharing the key the new index has to keep unique - the type is left with **no index at all** and an error. On
+an index the type only inherited it took away the parent type's index instead, which is issue #4083.
+
+So `withIgnoreIfExists` now means only what it says. Callers that genuinely mean "make this the definition" opt in
+explicitly:
+
+```java
+database.getSchema().buildTypeIndex("T", new String[] { "Scalar" })
+    .withType(Schema.INDEX_TYPE.LSM_TREE).withUnique(true)
+    .withReplaceIfIncompatible(true)     // new: may take away the plain index on those properties
+    .create();
+```
+
+and even then the replacement is undone if the new index cannot be built, so a failed upgrade costs an error rather
+than an index. An inherited index is still never replaced.
+
+### Cypher: a uniqueness constraint over a plain index still upgrades it
+
+`CREATE CONSTRAINT ... REQUIRE ... IS UNIQUE` and `... IS NODE KEY` are statements about the data rather than requests
+to create an index if one is missing, so they are the callers that opt into the replacement above. Neo4j keeps the
+range index and the constraint as two separate objects; ArcadeDB has one index per property set and a unique index
+covers both roles, so the upgrade is the equivalent end state and a Neo4j migration script that creates indexes and
+then constraints keeps working. Cypher's own `CREATE INDEX` does not opt in: it finds a unique index sufficient and
+leaves it alone rather than replacing it with a plain one.
+
+### A manual index name that already names a different index
+
+Reachable only through `CREATE INDEX <name> IF NOT EXISTS ...`, since the auto-derived name is built from the property
+list. A name that already belongs to an index on other properties used to satisfy the guard; it is now reported, since
+two different property sets under one name are two different indexes.
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1723,3 +1723,15 @@ leaves it alone rather than replacing it with a plain one.
 Reachable only through `CREATE INDEX <name> IF NOT EXISTS ...`, since the auto-derived name is built from the property
 list. A name that already belongs to an index on other properties used to satisfy the guard; it is now reported, since
 two different property sets under one name are two different indexes.
+
+### The index kind is now read before the existing-index lookup
+
+`TypeIndexBuilder.create()` validates `indexType` at the top, because deciding whether the index already on those
+properties covers the request needs to know what the request is. One embedded-API shape changes as a result: a builder
+with **no** `withType(...)` and `withIgnoreIfExists(true)` used to hand back the existing index on those properties,
+and now raises `DatabaseMetadataException` the way it always did when no index was there. Every in-tree caller sets the
+type; a get-or-create helper that relied on omitting it has to pass the type it expects.
+
+A failed replacement restores the page size along with the rest of the definition, via `getPageSizeForNewFile()` - the
+same accessor a rebuild uses, so a page size the current file carries but creation would refuse cannot turn the restore
+into a second failure (#5713).
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -1735,3 +1735,17 @@ type; a get-or-create helper that relied on omitting it has to pass the type it 
 A failed replacement restores the page size along with the rest of the definition, via `getPageSizeForNewFile()` - the
 same accessor a rebuild uses, so a page size the current file carries but creation would refuse cannot turn the restore
 into a second failure (#5713).
+
+### Embedded API: `getOrCreateTypeIndex` no longer upgrades an incompatible index
+
+`LocalSchema.getOrCreateTypeIndex(...)` sets `withIgnoreIfExists(true)`, so it inherits the change above: asked for a
+`UNIQUE` index where a `NOTUNIQUE` one already covers those properties, it used to drop and rebuild, and now raises
+`IllegalArgumentException` naming both definitions. That is the point of the fix - the rebuild is what could leave the
+type with no index - but embedded callers using it as an "ensure this index" helper across a schema change are the ones
+who will meet the new exception. Either drop the old index explicitly, or build through
+`buildTypeIndex(...).withReplaceIfIncompatible(true)` if replacing it really is what you mean.
+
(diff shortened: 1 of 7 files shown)


lvca:
Thanks — pushed 63f8ffd47 for the review points.

**1. Page size not restored — fixed.** Captured in `ReplacedIndexDefinition` via `getPageSizeForNewFile()` rather than `getPageSize()`: the value goes back through the creation path, and #5713 established that an index whose *current* page size is not one creation would accept has to answer with a legal one instead — otherwise the restore fails on exactly the index it exists to save. Mutation-checked (`aFailedUpgradeRestoresThePageSizeToo`): without the fix the index comes back at 262144 instead of the 16384 it had.

**1b. Metadata templated off a live bucket object — kept, documented.** I tried the suggested reset of `associatedBucketId` and it *broke* the drop path: `Index 'T_0_...' is not valid. Probably has been dropped or rebuilt`, because `of()` runs while the old index is still live. Reverted. The comment now records why the template is deliberate — only the concrete `IndexMetadata` subclass carries the index-type settings (analyzers, geospatial precision, vector dimensions) and there is no generic way to copy them — and why the stale bucket id is inert: `LocalSchema.createBucketIndex` reads the object as a *template*, taking the collations and TypeIndex name off it and leaving the rest to the per-bucket index it builds, exactly as on the normal creation path.

**2. Manual name reused across types — fixed.** The branch now compares `getTypeName()` as well, with `guardedCreateWithANameTakenByAnotherTypeIsReported` covering it.

**3. `indexType == null` moved to the top — noted and pinned.** Added `aBuilderWithoutAnIndexTypeIsRefused` and a changelog section naming the affected shape (no `withType` + `withIgnoreIfExists(true)` + a matching existing index used to hand that index back).

**4. Null-strategy wording — reworded.** It is excluded for being non-structural (settable through `ALTER`), and the Javadoc now says explicitly that a satisfied request stays a plain no-op rather than reconciling the strategy.

**Crash window** — agreed, left as-is and now a known property rather than an oversight.

Verification: 1414 engine tests in the schema + index packages and the server index/HTTP suites, all green except `LSMVectorIndexRecoveryTest#vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild`, which asserts a background vector rebuild has drained its delta buffer and passes in isolation — a contention flake (three engine suites were running concurrently on this machine), unrelated to this change.

lvca:
Round two addressed in 727b853bf.

**1. Restore loses a custom index name — it doesn't, and now a test says so.** The name never travelled through `withIndexName`: it rides on the captured `IndexMetadata.typeIndexName`, which `LocalSchema.createBucketIndex` copies onto each bucket index and `LocalDocumentType.addIndexInternal` honours when minting the TypeIndex — the same #4139 route a manual name takes on the normal creation path. `aFailedUpgradeRestoresAManualIndexNameAndItsPageSize` pins it (name **and** page size, as asked in item 4), and it is mutation-checked: remove the metadata pass-through from `restoreReplacedIndex` and the test fails.

**2. `getOrCreateTypeIndex` release note — added.** It sets `withIgnoreIfExists(true)`, so a uniqueness mismatch now raises where it used to drop and rebuild. The note names the changed contract and gives the two ways forward (drop the old index explicitly, or `buildTypeIndex(...).withReplaceIfIncompatible(true)` if replacing really is the intent).

**3. Transactionality — confirmed NOT atomic, and now documented.** `LocalSchema.dropIndex` wraps itself in `recordFileChanges` and commits, and the build below opens its own, so the two cannot roll back as one. The comment at the drop site records the accepted risk and why the alternatives are worse: holding a schema transaction across a full index build is the wrong trade for a statement that already rebuilds every entry, and building the new index alongside the old one is not representable while `LocalDocumentType.indexesByProperties` keeps one index per property set. Only an explicit `withReplaceIfIncompatible` is exposed to the window.

**4. Test gaps — both filled.** Added `replaceIfIncompatibleStillRefusesAnInheritedIndex` (the #4083 path through the own-vs-inherited guard) and the name+page-size case above.

**Nits.** `describeDefinition` left as is, per your own read. `List.of(fields)` needs `calculateProperties` to return null, which needs `propertyList == null` — the grammar cannot produce it, and a guard there would be a branch no statement can reach; left alone.

Verification: 1305 tests across the engine `schema` + `index` packages, 0 failures / 0 errors, vector tests included this time.

lvca:
Round three addressed in 7cbd6abbd — both documentation-level, since nothing here was blocking.

**1. Sub-configuration not compared — noted in the Javadoc, as suggested.** `satisfiesRequest` compares the structural definition only, so two same-kind indexes configured differently (vector `dimensions`, full-text analyzers, geospatial precision, collations) still satisfy each other. The Javadoc now states that this is deliberate and gives the reason it is not simply an oversight: those settings live in per-index-type `IndexMetadata` subclasses with no common notion of "provides at least what was asked for" — is a 768-dimension index acceptable to a request for 384? is a stricter analyzer? — so answering it needs a per-type rule rather than the structural comparison, and inventing one here would be guessing.

**4. Generic message in the satisfied-but-unguarded case — improved.** It now ends with `. It already provides what was requested: use IF NOT EXISTS to make this statement idempotent` when the existing index satisfies the request, and `. Drop it first if the definition has to change` when it does not. That covers the Cypher `CREATE CONSTRAINT` without `IF NOT EXISTS` case you named. The long-standing sentence is kept verbatim as the prefix, so anything already matching on it still matches.

**2 and 3** — no change; both are already covered in the release notes (the `getOrCreateTypeIndex` contract) and at the drop site (the non-atomic window, with why the alternatives are worse), which is what you were checking for.

Verification: 1356 tests across the engine `schema` + `index` packages plus the Cypher index/statistics tests. One failure, `LSMVectorIndexRecoveryTest#vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild`, which asserts a background vector rebuild has drained its delta buffer within a timeout — it passed in the immediately preceding run of the same suite and passes in isolation, so it is the machine being busy (parallel suites), not this change. Nothing in this commit is reachable from the vector build path.

lvca:
Thanks — LGTM noted. One last change in e21743f8f, documentation only.

**1. `getOrCreateTypeIndex` — promoted to a `BREAKING:` heading.** You are right that it deserves prominence: it is the one change here that surfaces at runtime rather than at compile time. It now sits under the same `### BREAKING:` convention the rest of this release file uses, with a line saying exactly that, so it reads at the same level as the `withPageSize(0)` break above it.

**2. Non-atomic window** — agreed on all counts, no change; it is documented at the drop site with the rejected alternatives, and only the explicit opt-in reaches it.

**Nit: the pre-existing `System.out.print(".")` progress callback** in `CreateIndexStatement` — leaving it as you suggested. It predates this PR and removing it would be unrelated churn in a diff that is already about one thing; it is worth its own issue rather than a drive-by here.

This closes out every point raised across the four rounds. Ready for your review.

Issue #5155, opened 2026-07-08
Optimize edge-removal chunk anchoring: avoid mutable-page fetch for read-only walk hops (#5148 follow-up)


lvca:
## Summary

Optimization follow-up from the #5148 code review. The #5153 fix makes the edge-removal path anchor each chunk it visits in the transaction via `EdgeLinkedList.loadChunkForWrite` (`fetchPageInTransaction` -> `getPageToModify`), which materializes a **mutable page copy for every chunk the remove walk traverses**, including read-only hops that don't contain the target edge.

This is correct and bounded (a 32k-edge hub is ~8 pages at 8 KB/chunk, and unmodified pages are dropped at commit by the `range[1] <= 0` filter), so it is **not** a correctness or memory problem. But it upgrades read-only chunk visits to mutable-copy fetches held until commit, which is wasted churn on a long chain.

## Detail

- `removeEdge` / `removeEdgeRID` modify exactly **one** chunk (the one containing the edge), yet anchor the whole prefix of the chain walked to reach it.
- `removeVertex` genuinely modifies **many** chunks (it strips the vertex from every chunk), so anchoring each is appropriate there.

## Possible optimization

For the single-chunk removers, traverse the chain with plain (unanchored) reads and anchor only the chunk actually about to be modified - re-reading it from the anchored page before applying the removal (mirroring the append-side read-time anchor). Care needed for the empty-chunk relink path, where `prevBrowsed` is also modified and would need anchoring too.

## Priority

Low. Measure on a large super-node (multi-thousand-edge chain with concurrent removes) before optimizing - the append path was kept allocation-free and measured, so the remove path deserves the same treatment if it proves hot.

Ref: PR #5148, `docs/supernode.md` section 8.2.

lvca:
Fixed and merged to main (commit 754d8927c).

When you delete an edge from a vertex with a very large number of edges (a super-node), ArcadeDB walks the edge list to find it. Previously every chunk visited on the way was copied into the transaction, even the ones that didn't hold the edge being removed. Now only the chunk that actually contains the edge is copied; the rest are just read.

On a hub with 60k edges (a chain of 51 chunks) deleting an edge stored near the end went from copying 7 pages to 1, about 22% faster on that path. Deletions on small/normal vertices are unaffected.

[...]

Issue #5147, opened 2026-07-08
Concurrent edge insertion into a super-node vertex drops edges (edge-list chunk-allocation race)


lvca:
## Summary

Under **concurrent edge insertion into the same super-node (hot) vertex**, the graph engine can **drop edges** and occasionally **orphan an edge-list chunk**. It is a data-loss race in the edge-list *chunk allocation / head-pointer* path, independent of query language or edge kind (reproduces with both regular and light edges).

It reproduces on `main` (26.7.2-SNAPSHOT) and is **pre-existing** - it is NOT caused by the commutative edge-append merge (`arcadedb.graph.edgeAppendMerge`): it reproduces identically with that feature **disabled**. (The merge actually *reduces* the loss, because it removes most of the full-transaction retries that widen the race window.)

## Impact

- Silent edge loss: transactions report success but a few of the committed edges are missing from the hub vertex's edge list.
- Occasionally an orphaned chunk: the vertex's `inEdgesHeadChunk`/`outEdgesHeadChunk` points at a chunk that is then not found, logged as SEVERE and "repaired" by starting a fresh chunk - which orphans everything previously linked behind it.
- Directly relevant to the super-node write-contention workloads we are hardening (payments graph funnelling into a central account).

Severity: high (silent data loss under a realistic concurrent workload).

## Reproduction

Many threads concurrently add an edge into one shared hub vertex. Spreading the per-transaction source vertices / edge records across many buckets makes the hub's edge-list append the dominant contention and makes the race fire reliably:

```java
// 8 threads, each committing 4000 edges (Src -> Hub), retries handle MVCC conflicts.
// Src/LINK use 16 buckets so record insertion is not the bottleneck; Hub has 1 bucket.
for (int t = 0; t < 8; t++) new Thread(() -> {
  for (int i = 0; i < 4000; i++)
    db.transaction(() -> {
      final MutableVertex src = db.newVertex("Src");
      src.save();
      src.newEdge("LINK", hubRID);
    }, false, 10_000);
}).start();
// ... join, then:
long inDegree = hubRID.asVertex().countEdges(Vertex.DIRECTION.IN, "LINK"); // < 32000
```

### Observed (append-merge DISABLED, i.e. baseline engine behaviour)

- `hub IN-degree = 31863` for 32000 committed edges (137 lost). Repeats every run with a variable small deficit.
- With light edges at lower volume the deficit is smaller but still non-zero (e.g. 6407/6408).
- Occasional log line:

```
[...]

PR #5744, opened 2026-08-01
fix(engine) #5725: an edge appended behind a vertex delete retries the delete instead of outliving the vertex

### LINKED ISSUE #5725: Follow-ups from #5680: an append racing a vertex delete leaves a ghost edge, and deleteVertex materialises the whole edge list
### Summary

Two follow-ups left behind by #5680 (PR #5707, merged as `324eca1f1`). The first is a real correctness
bug and the reason this issue exists; the second is a scalability note recorded so it is not lost.

---

## 1. An append racing a vertex delete leaves a ghost edge (the APPEND side of the #5670 window)

#5670 / #5678 / #5680 made edge **removal** strict: an edge-list chunk that cannot be read is a retryable
conflict rather than "nothing to remove here". The **append** side still has a window, and it produces the
same defect from the other direction.

An edge that `newEdge` commits while its target vertex is being deleted can outlive that vertex. Measured
on a 4-hub x 100-edge fixture with 6 appender threads racing 4 deleter threads, dumping the survivor:

```
GHOST edge=#65:70 out=#43:490 outExists=true in=#1:1 inExists=false inIsHub=true
```

The surviving edge record has a **live `out` vertex** and an **`in` naming a hub that no longer exists**.
`CHECK DATABASE` reports it as an invalid link. Reproduction rate is roughly **1 in 3 rounds**.

### Why this is not #5680, and why nothing on the vertex side can fix it

Verified by reverting #5680's fix and re-running: the failure reproduces **identically** (3 failures in 6
runs on the parent commit). It is a pre-existing defect, not a regression, and #5680 neither caused nor
addressed it.

It also cannot be fixed where #5680 was fixed. The ghost edge **did not exist** when the deleter's
collection walk ran - it was created by a transaction that committed afterwards. No amount of strictness
in `deleteVertex`'s read of its own edge list can collect an edge that is not there yet. The fix has to be
on the append path, or in the commit-time conflict detection between the two.

### A lead worth checking first (hypothesis, NOT verified)

The `newEdge` path writes in up to three places: the edge record, the source's OUT list, and the target's
IN list. The appender and the deleter should collide on the target's pages, and usually do - which is why
this is intermittent rather than constant. Two ways a collision could be missed:

- **The append landed in an existing chunk with room**, so the appender wrote only the chunk page, never
  the vertex record. If the deleter commits first, `GRAPH_EDGE_APPEND_MERGE` may then rebase the
  appender's append onto the newer committed image of that page - the image in which `deleteAll()` has
  already removed the chunk record. `EdgeLinkedList.updateSegment`/`deleteAll` call
  `poisonEdgeAppendPage` (`EdgeLinkedList.java:595-602`), but that excludes the page from the **deleter's**
  own merge; it does not obviously stop a **different** transaction from rebasing an append onto it.
- **Or the endpoint resolution is too lazy**: `newEdge` resolves the target with `loadContent=false`, so a
  target deleted in the window may never be touched hard enough to raise `RecordNotFoundException` before
  the edge record is created and committed.

Either way the desired outcome is the one the removal path now has: the appender gets a retryable
`ConcurrentModificationException` and re-reads, finding the target gone and failing cleanly, instead of
committing an edge into a vertex that is being deleted.

### Reproducing

`Issue5680VertexDeleteEdgeCollectionTest.concurrentVertexDeletionAndEdgeAppendingLeaveNoEdgeBehind`
(`engine/src/test/java/com/arcadedb/graph/`) already builds the fixture. It is deliberately scoped
**around** this defect: it asserts only that the *fixture* edges - the ones that were in the hub's list
when the deleter started - die with their hub, and it has no `assertIntegrityClean()`, both with the
reason inline.

To see the bug, restore the assertion the test deliberately does not make - after every hub is deleted,
`countType("LINK", false)` must be `0`, since every edge in the fixture points at a hub - and run the
class a handful of times.

**Please do not "fix" that test by tightening it back.** It will simply become flaky again and hide this
issue rather than surface it. The scoping comment in the test points here.

### Impact

Silent graph corruption under ordinary concurrency: an edge record whose `in` (or `out`) names a deleted
vertex, reported later by `CHECK DATABASE` as an invalid link. Same severity class as #5670, on the
opposite operation. `CHECK DATABASE ... FIX` cleans it up after the fact.

---

## 2. `deleteVertex` materialises the whole edge list before deleting

Raised in review of PR #5707. `GraphEngine.deleteVertex` collects every edge into an
`ArrayList<Identifiable>` and only then deletes them:

```java
final List<Identifiable> edgesToDelete = new ArrayList<>();
collectEdgesToDelete(mostUpdatedVertex, Vertex.DIRECTION.OUT, force, edgesToDelete);
collectEdgesToDelete(mostUpdatedVertex, Vertex.DIRECTION.IN, force, edgesToDelete);
```

This is **unchanged behaviour** - it predates #5680 - but on a promoted super-node with millions of edges
it is a large allocation on a path that could stream, and it sits against the "performance and lightweight
on the garbage collector" rule in `CLAUDE.md`.

It is not a mechanical change. The two-phase collect-then-delete exists precisely **because** the removals
mutate the chain the walk is standing on: `deleteEdge` relinks chunks and deletes emptied ones underneath
the iterator. Streaming means walking a list while it is being restructured, which is the class of problem
#5155, #5670 and #5680 were all about. Any attempt should come with the same standard of deterministic
tests those did.

Chunked draining (collect N, delete N, re-read the head, repeat) is probably the cheaper middle ground: it
bounds the list without asking a single iterator to survive the mutation.

lvca:
Fixes #5725 (item 1). Item 2 is deliberately not in this PR - see the end.

## The defect is real, but not for the reason the issue proposed

The issue offered two leads, both about `GRAPH_EDGE_APPEND_MERGE` rebasing an append onto a page the deleter had already emptied, or about `newEdge` resolving its target too lazily. Neither is it. Turning **both** commit-time merges off (`GRAPH_EDGE_APPEND_MERGE` and `TX_PAGE_SLOT_MERGE`) still reproduces - they only widen the window.

Instrumenting the surviving edge settles the ordering: it always reports `hubAfterCommit=true reachableInHubList=true`. The append **committed first**, the edge **was** in the hub's IN list, and the delete then committed on top of it without conflicting.

## Root cause

`GraphEngine.deleteVertex` is a read-modify-write over the whole edge list - collect every edge, disconnect each from the vertex at its other end, drop the chunk records, delete the vertex record - but only the *write* half left an MVCC footprint.

`collectEdgesToDelete` read the chain through `getEdgeHeadChunkForWrite` -> `buildEdgeList` -> a plain `lookupByRID`, which under `READ_COMMITTED` does not retain the pages it reads. The removals and the chunk drain then captured each page *later*, via `fetchPageInTransaction` at whatever version it had by then - a version that already contained any append that landed in between. `checkPageVersion` compared the newer version against itself, found no conflict, and committed. The vertex went away, the chunk holding the new edge went with it, and the edge record survived with a live `out` and an `in` naming nothing.

This is exactly the read-modify-write gap #5147 closed on the **append** path (`anchorHeadChunkPage`), left open on the walk that **removes**. And it is genuinely unreachable from where #5680 was fixed: the edge did not exist when the collection ran, so no amount of strictness in that read could have collected it.

## The fix

A vertex delete now pins, in its own transaction, everything its edge list can grow through, at the version it reads the list at.

- **`EdgeLinkedList.anchorForFullRemoval()`** - every chunk page of the list, reading each chunk *through* its anchored page so the `previous` pointer the walk follows comes from the pinned version. Every chunk, not just the head: an appender resolves the head from its own handle of the vertex, so a handle predating a head flip appends into a chunk that is no longer the head but is still in the chain.
- **`StripedEdgeList.anchorForFullRemoval()`** - the promoted super-node override (#5156): the stripe directory (a stripe head flip rewrites a slot in it) plus every chain of every generation.
- **`GraphEngine.checkEdgeListHeadsUnchanged()`** - the half that touches *no* page of the list. An append that finds the head chunk full writes no chunk at all: it creates a new one and records it in the vertex record. The vertex's head pointers are re-read through their anchored page just before the record delete - the same page that delete writes, so the check is binding rather than advisory - and a moved head refuses the delete.

**The pins cost nothing at the peak.** `deleteAll` deletes every one of those chunks anyway, so their pages end up in the transaction regardless; this only brings them in early enough to be worth something.

Deliberately **not** folded into `getEdgeHeadChunkForWrite`: `deleteEdge` calls that once per endpoint per edge, and pinning a whole neighbour list there would retain a page copy per visited chunk for every edge removed - on a super-node delete that is the whole graph's edge-list pages in one transaction.

### Tolerances preserved (each cost a round to find)

- The pin runs **after** `edges` is assigned, so a pin that fails under `force` still leaves a usable list behind: the tolerant walk collects what it can and the chunk drain still runs.
- `readEdgeListHeads` is **purely advisory** and swallows everything. It runs first, and on a lazily-loaded `ImmutableVertex` the head read materialises the record - which put every way that load can fail in front of a delete that used to meet them further in, where each is already owned: a corrupt buffer is tolerated so the vertex stays deletable (#4420/#4432), and a vanished or concurrently-rewritten multi-page body is the retryable conflict `getEdgeHeadChunkForWrite` raises and `force` then absorbs. Re-raising from there broke `Issue4432CorruptVertexDeleteTest` and `BrokenMultiPageRecordDeleteTest` in turn; both are green now.

## Tests

`Issue5725GhostEdgeOnAppendRaceTest` - four deterministic tests plus the stress fixture. The deterministic ones inject a genuinely concurrent append, from another thread, at the exact instant `deleteVertex` moves from collecting to disconnecting (the first read of a `Src` vertex record - the collection reads chunks and edge records only).

**Mutation-checked**: with the three main-source files stashed, `anEdgeAppendedBetweenTheCollectionAndTheRemovalRefusesTheDelete`, `anEdgeAppendedIntoAPromotedSuperNodeBetweenTheCollectionAndTheRemovalRefusesTheDelete` and `deletingAVertexThroughAHandleWhoseHeadPointerIsStaleReportsAConflict` all fail; with the fix all pass. The stress test only fails about half the time on its own, which is why the deterministic ones exist.

`Issue5680VertexDeleteEdgeCollectionTest` can now assert the full invariant its comments had to scope around ("no LINK edge survives at all", not just the fixture edges) and ends with a clean `CHECK DATABASE`. The scoping comments are replaced rather than deleted, pointing here.

**Verified**: full engine suite `-DexcludedGroups=slow,benchmark` = 10634 tests, 0 failures. Engine `-Dgroups=slow` = 217 tests, 0 failures. Both are the exact shapes `mvn-test.yml` runs.

## Item 2 of the issue is not addressed here, on purpose

`deleteVertex` still materialises every edge into an `ArrayList` before deleting any. The issue records it as a note and says any attempt needs the same standard of deterministic tests the #5155/#5670/#5680 series had, which I agree with: the two-phase shape exists *because* the removals relink and delete chunks underneath the walk. Mixing that into a correctness fix would make both harder to review. Worth noting the sizing has changed - after this PR the retained chunk pages, not the list, dominate a super-node delete, and streaming the walk would not remove those.

If you want it pursued, the higher-value target in the same method is that disconnecting each collected edge walks *this* vertex's own list too (O(degree) per edge, O(degree^2) overall) to remove an entry from a chunk the drain is about to delete outright.


lvca:
Thanks - both rounds. Taking the points in order.

**Round 1, comment precision (both taken, 43e26919c).** Both were right and both were cases of a comment claiming more than the code does:
1. reads never *populate* the tx cache, but `lookupByRID` does *consult* it, and a record this tx wrote earlier is answered from there - which is exactly the case the `getWrittenRecord` guard returns on. Reworded to say that rather than implying an unconditional bypass.
2. a pin that fails does jump to the `catch` and the walk is skipped either way. What the ordering buys is a non-null `edges`, so the direction is still reported as present and the drain runs. Reworded.

**Round 2, point 2 - test isolation for the cache-bypass assumption (taken, 381ae17ab).** This was the best catch in either round: the assumption is load-bearing and *nothing* tested it, because every other test here deletes a vertex the delete transaction never read first. A caching read would have degraded the head check into comparing a value with itself and the whole suite would still have been green.

Worth reporting what happened when I built it the way you suggested (fresh read + concurrent append flipping the head to a new chunk): **it passes with the fix reverted.** The chunk pins make the newly created chunk unreadable through the deleting transaction's pages, so #5670's strict head read refuses it before the head comparison is ever reached. Correct behaviour, and genuinely reassuring about the layering, but useless as a regression test - so I dropped that shape and split it in two:

- `aPlainReadDoesNotPopulateTheTransactionRecordCache` pins the fact itself. Passes with or without the fix **by design** - it guards an assumption, not the fix - so a future change to the read/cache path fails loudly and locally.
- `aFreshReadInsideTheDeleteTransactionDoesNotBlindTheHeadCheck` is the scenario, driven from the stale-handle side so the head comparison is the mechanism under test. **Fails without the fix.**

Mutation-checked count is now 4 of 7 (the other three are the `force` escape hatch, the cache-behaviour guard, and the probabilistic stress fixture, none of which should be mutation-distinguishable).

**Round 2, point 1 - extra CPU/allocation on the uncontended path.** Declining, with the reasoning stated rather than waved off. The anchor walk adds a `copyOfContent` per chunk, and I could avoid it by parsing the `previous` RID straight out of the page buffer - but that trades a real correctness risk in the hot delete path for the *smaller* term. On a super-node delete the dominant allocation is the `ArrayList` of materialised `Edge` records (issue item 2), which is orders of magnitude larger than the discarded chunk copies. If the anchor walk's cost is ever shown to matter, the right fix is to fold the anchor into the collection walk so it is one traversal instead of two - not hand-parsing pages.

**Round 2, point 3 (cycle guard)** - agreed, no action, as you noted.

Re-verified after each change: graph package + `BrokenMultiPageRecordDeleteTest` = 384 tests, 0 failures.

One process note for the record: my first attempt at re-running the mutation check used `git stash`, which in this setup shares a stack across worktrees and popped a *different* session's entry into my tree. No work was lost (a conflicted `pop` leaves the entry in place, and I restored the foreign file to HEAD), but the run it produced was meaningless - the fix was already committed, so nothing had been stashed and the "without the fix" run was actually *with* it. The mutation results above are from `git checkout <parent> -- <paths>` instead.

lvca:
Round 3 - the actionable point was right, and it is wider than you scoped it. Fixed in 8621c5853.

**The pin failing a broken chain (taken).** Confirmed exactly as described: `anchorForFullRemoval()` runs before the collection walk, so a chunk it cannot read failed the whole direction before a single edge had been collected. The old incremental walk disconnected everything in chunks 1..k-1 from the vertices at their other end; the pin made that zero, leaving those far-end pointers dangling on neighbours nobody asked to touch. That is the "back-reference survives its edge" defect #5670 fixed, re-entering through the pin.

**It is not force-only.** The decode-family arm of `collectEdgesToDelete` is tolerated with `force == false` as well - that arm is what keeps a vertex whose own list is corrupt deletable at all (#4420/#4432) - so an *undecodable* mid-chain chunk regressed the ordinary path in the same way. Only the vanished-chunk shape was force-gated.

**Fix:** `anchorChain` now stops at a hop it cannot follow and leaves everything up to that point pinned. The reasoning I settled on is that this pass must not be the thing that *reports* a broken chain - the collection walk follows the same pointers in the same order, meets the same wall, and the policy already lives there and is carefully split between retryable and tolerated. Stopping quietly hides nothing, because the next walk raises what this one saw.

Two things deliberately stay outside the tolerance:
- **the HEAD load still propagates.** It is where an append lands, and unlike the hops it is *not* re-read by the collection walk (`lastSegment` is already materialised), so continuing past a head that could not be pinned would leave the collection reading a page it never pinned - the exact window this mechanism exists to close.
- **only "cannot read this chunk" is absorbed.** An I/O fault surfacing as `DatabaseOperationException` is not a broken chain and must not be mistaken for one, or the collection walk could read a chunk this pass failed to pin.

Your suggested shape (catch under `force`) would have fixed the narrower half; this version also covers the corrupt-chunk path that is tolerated without `force`, and puts the decision where the existing policy already is rather than duplicating it.

**Test:** `aForcedDeleteOfAVertexWhoseChainBreaksMidWayStillDisconnectsTheEdgesInFrontOfTheBreak` - break the second chunk from the head, force-delete, and assert both directions of the invariant: some edges *were* reached and deleted (a whole-direction failure leaves all 200 behind), and some *did* survive behind the break (so it is not accidentally asserting that a broken chain deletes cleanly). Mutation-checked against the tolerance specifically.

**Minor points:** agreed on both, no action. The `anchorChain` cycle guard matches `deleteAll`'s existing exposure, and I have noted the stress test's 10k-retry / `TX_RETRY_DELAY=1` combination against the 5s lock timeouts for future intermittency triage.

Graph package + `BrokenMultiPageRecordDeleteTest`: 385 tests, 0 failures.

lvca:
Round 4 - one of the four notes was actionable, taken in 15e4901e8; the other three need no change and I agree with your reading of each.

**2. The behaviour change reaching remote client-managed transactions (taken).** You are right that this deserves to be spelled out where a reader lands rather than left to a back-reference. The note said "the standard retry loops absorb it" and pointed at the #5670 entry for the rest. It now names the case: a client-managed explicit transaction over `RemoteDatabase` spans several HTTP requests, so its commit is not auto-retried and the exception reaches the caller, who should retry the transaction - and what is new is only that a vertex delete racing an append has joined the set of operations that can raise it.

**1. Breadth of the `anchorChain` catch list** - no change, and the point about future additions is well taken: a new decode-family exception has to be added there deliberately. That is the same closed-list discipline `collectEdgesToDelete` documents, and the two should stay in step.

**3. Striped `anchorForFullRemoval` discarding the returned head** - confirmed intentional, as you read it. There is no single `lastSegment` to re-seat on that layout; the pin is the side effect and the collection walk re-resolves each chain through the anchored directory.

**4. Performance** - agreed it is a real, bounded cost, not zero: one extra `copyOfContent()` per chunk, so a super-node delete reads each chunk once more than before. Recorded rather than optimised, for the reason given last round - the dominant allocation on that path is the `ArrayList` of materialised `Edge` records (issue item 2), and the right fix if it ever matters is folding the anchor into the collection walk, not shaving this.

**On CI:** not green yet at the time of writing - `setup` and the CodeQL `Analyze` jobs are still queued, and `setup` is what gates the test matrix. `claude-review`, Codacy and Meterian are all green. I will let the `mvn-test.yml` shapes finish rather than call this done on my local runs; locally both CI shapes passed (10,634 with `-DexcludedGroups=slow,benchmark`, 217 with `-Dgroups=slow`, 0 failures), plus 385 in the graph package after each round of changes.

lvca:
Round 5 - LGTM with four minor notes, none of which I am taking. Reasons for each, since "no change" should be argued rather than asserted.

**1. Sharing one predicate between `anchorChain` and `collectEdgesToDelete` so the decode lists cannot drift.** Declining, because the two lists are not the same list and unifying them would hide that. `collectEdgesToDelete` needs `ConcurrentModificationException` in a *separate* arm from the decode family - that split is the whole point of #5680's policy (retryable, force-absorbed) versus #4420/#4432's (tolerated regardless of `force`). `anchorChain` absorbs both, because it is not the one deciding. A shared predicate could only cover the decode subset, and using it would mean `catch (RuntimeException)` plus a rethrow at a site where Java's multi-catch already states the closed list explicitly - strictly less readable than what is there. The drift risk is real but is answered by the direction that matters: `anchorChain`'s Javadoc names `collectEdgesToDelete` as the owner of the policy, so anyone editing the pin is pointed at the site that decides.

**2. `readEdgeListHeads` catching bare `RuntimeException`.** Declining - the breadth is the design, not an accident, and narrowing it is what I already had to undo twice. It runs before `collectEdgesToDelete` and only captures a value for an optional check, so anything it re-raises pre-empts a condition that is already owned further in. A narrower list broke `Issue4432CorruptVertexDeleteTest` (a truncated buffer must stay deletable) and then `BrokenMultiPageRecordDeleteTest` (a multi-page body being rewritten concurrently is the retryable conflict `force` absorbs). The Javadoc enumerates both, which is the right place for the reasoning given the catch cannot express it.

**3. Cost on the non-force path.** Agreed and no action - your accounting is right, the head re-check is the only genuinely new O(1) work and the chunk pins are free at the peak.

**4. Observable API change.** Thanks - that was your note from last round.

**On CI:** still running, so this is not done yet. `setup` (the test-matrix gate) and the CodeQL `Analyze` jobs are in progress; `claude-review`, Codacy, Meterian and `Summary` are green. I am leaving the PR alone until the `mvn-test.yml` shapes report, rather than resting on my local runs - and deliberately not pushing anything further, since each push restarts both CI and the review for what would be cosmetic changes.

lvca:
CI has finished. **29 of 30 checks green; the one failure is a pre-existing flake unrelated to this change.**

Green, including everything that exercises this code:
- `unit-tests` - the engine suite, carrying the 7 non-slow tests in `Issue5725GhostEdgeOnAppendRaceTest`
- **`slow-unit-tests`** - the `@Tag("slow")` concurrency fixtures in both `Issue5725...` and `Issue5680...`. This was the one I actually wanted to see, since the round-3 review flagged the 10k-retry / `TX_RETRY_DELAY=1` combination against the 5 s lock timeouts on a slow runner. It passed.
- `build-and-package`, `integration-tests`, `builder-tests`, all six e2e suites, CodeQL, Codacy, Meterian, `coverage-report`, `codecov/patch` and `codecov/project`

**`ha-integration-tests` failed, and it is not this PR.** Two tests:

- `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied` - fails on **main** in 2 of the last 4 runs of this workflow.
- `RemoteStickyConnectionStrategyIT.pinIsReleasedAfterRollbackAndAllowsSubsequentStickyTransaction` - this test does no graph work at all. It inserts documents into a plain type over a `RemoteDatabase` against a Raft follower and checks that a STICKY pin is released across a rollback; the failure is `expected 1L but was 0L` on the visibility of the *second* insert. Nothing in `GraphEngine.deleteVertex`, `EdgeLinkedList` or `StripedEdgeList` is reachable from it.

More broadly, `ha-integration-tests` fails on main in 3 of the last 4 runs with a **different** set of tests each time - `Issue5569SlotMergeDeleteRaftIT`, `RaftReadConsistencyBookmarkIT`, `RaftReplicationMaterializedViewIT`, `RaftHARandomCrashIT`, `RaftHTTP2ServersIT` - so the suite is environmentally flaky rather than reacting to any one change. Worth its own issue; it is not something this PR can or should fix.

**On Codecov:** patch coverage 77%, project coverage up +0.01%. The 11 uncovered lines are the exception arms - the decode-family catches, the `RecordNotFoundException` / `ClassCastException` arms in `checkEdgeListHeadsUnchanged`, the corrupt-buffer early returns. Some are covered by the new tests; the rest need a concurrent slot reuse or an I/O fault to reach, and I would rather leave them uncovered than contort a test into asserting something it cannot really produce. Each is documented where it sits.

Ready for review from my side. Five review rounds, three of which produced changes: the two comment-accuracy corrections, the cache-bypass assumption test, and - the substantive one - the pin failing a partially broken chain instead of stopping at it, which had regressed `force` and the corrupt-vertex path.

PR #5749, opened 2026-08-02
perf(engine) #5748: optimize brute-force scan to walk only allowed RIDs instead of every ordinal

### LINKED ISSUE #5748: LSM_VECTOR: a query with an allow-list narrower than k costs a full index scan, every time
Found while measuring the brute-force fallback for #5558 (PR #5739). **Pre-existing on `main`, not introduced by that PR** - measured both ways, see below.

### What happens

`findNeighborsFromVector(query, k, efSearch, allowedRIDs)` decides whether to fall back to a brute-force scan by comparing the result count against `min(k, availableVectors)`. `allowedRIDs` does not enter that expectation, so a query restricted to fewer RIDs than `k` can never meet it: it looks permanently short, and the fallback walks **every ordinal in the index on every query**.

### Measured

50,000 vectors, 128 dimensions, INT8, `k=10`, allow-list of 3 RIDs, best of 3 rounds of 300 queries:

| workload | latency | brute-force scans |
|---|---|---|
| unfiltered `k=10` | **95 us/query** | 0 |
| allow-list of 3, `k=10` | **26,236 us/query** | **900/900** |

**276x**, and it is not a tail: the scan fires on literally every query.

Confirmed pre-existing by restoring the pre-#5739 guard (`availableVectors = ordinalMap.length` plus the `results.size() < availableVectors * 8 / 10` term) and re-measuring: **26,602 us/query, 900/900 scans**. The old 80% term was true here too, so it never suppressed anything - the two are the same within noise.

### Why the obvious fix is not obviously right

Capping the expectation at `allowedRIDs.size()` removes the scan, and I tried exactly that while working on #5739 - then reverted it, because a test showed what it costs. With a 3-RID allow-list on a tombstoned index the graph walk returns **2** of the 3: a restrictive `Bits` filter makes JVector's local-maximum heuristic stop expanding early, and the fallback is what *completes* the answer. Cap the expectation and that query returns 2 instead of 3.

So this is a real trade, not an oversight to patch:

- **cap by `allowedRIDs.size()`** - fast, silently lossy when graph recall under-fills the allow-list;
- **cap by the number of *live* allowed RIDs** - accurate, costs a location lookup per allow-list entry, which for a large allow-list is worse than the scan it saves;
- **scan only the allow-list** rather than every ordinal - the fallback walks `ordinalMap` and tests membership; walking the allow-list and resolving each RID to its ordinal instead would make the fallback O(allow-list) rather than O(index), which for a *narrow* allow-list is the interesting direction. Needs a RID -> vectorId lookup, which `VectorLocationIndex` already maintains as a reverse index (`ridToVectorIds`, added in #5318).
- **raise `efSearch` when the filter is restrictive** so the graph walk finds them all and the fallback never triggers.

The third looks most promising: it keeps the completeness the fallback provides and removes the part that scales with the index rather than with the query.

### Observability

`getStats()` gained a `bruteForceScans` counter in #5739, so this is now measurable on a live index rather than only visible as a `WARNING` log line. Anyone seeing filtered vector queries run slow should check it first.

waterWang:
## Summary

When an LSM_VECTOR query specifies an allow-list narrower than `k`, the brute-force fallback threshold (`expectedResults = min(k, totalVectors)`) is always unmet because the graph search can return at most `allowedRIDs.size()` results. This triggers a full index scan on **every** query (276x slowdown measured: 95 us → 26,236 us per query).

## Fix

When `allowedRIDs` is set, build a reverse index (vectorId → ordinal) and walk only the allowed RIDs instead of iterating every ordinal. This preserves completeness (the brute-force scan still runs when the graph search genuinely under-fills) while reducing the scan cost from **O(index)** to **O(allow-list)**.

No functional change when `allowedRIDs` is null or empty (the original ordinal walk is used unchanged).

## Verification

- Compilation: ✅ `mvn compile -pl engine` passes
- All 30 related vector search tests pass:
  - LSMVectorIndexEuclideanOrderingTest (4 tests)
  - LSMVectorIndexExtendedTest (6 tests)
  - LSMVectorIndexSearchCacheTest (2 tests)
  - LSMVectorIndexConcurrentUpdateTest (1 test)
  - LSMVectorIndexQuantizationComprehensiveTest (17 tests)

Fixes #5748

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -3461,26 +3461,59 @@ public class LSMVectorIndex implements Index, IndexInternal {
     // mapping (issue #4581).
     final ArcadePageVectorValues pageValues = vectors instanceof final ArcadePageVectorValues p ? p : null;
     boolean added = false;
-    for (int ordinal = 0; ordinal < ordinalMap.length; ordinal++) {
-      final int vectorId = ordinalMap[ordinal];
-      final VectorLocationIndex.VectorLocation loc = vectorIndex.getLocation(vectorId);
-      if (loc == null || loc.deleted)
-        continue;
-      if (seenRIDs.contains(loc.rid))
-        continue;
-      if (allowedRIDs != null && !allowedRIDs.isEmpty() && !allowedRIDs.contains(loc.rid))
-        continue;
 
-      // The location says the vector is live, but the read can still fail - a document whose vector property was
-      // removed or has the wrong type comes back as the placeholder, not as a vector. Scoring it would pair a real
-      // RID with a meaningless distance, and under COSINE with a distance of minus infinity, i.e. first place.
-      final VectorFloat<?> vec = vectors.getVector(ordinal);
-      if (vec == null || (pageValues != null && pageValues.isDeletedSentinel(vec)))
-        continue;
+    if (allowedRIDs != null && !allowedRIDs.isEmpty()) {
+      // Optimized path: walk only the allowed RIDs instead of every ordinal (issue #5748).
+      // When the allow-list is narrower than k, the unmodified expected-results threshold triggers
+      // a brute-force scan on every query. Walking the full ordinal map costs O(index), while
+      // walking the allow-list costs O(allow-list). Build a reverse index (vectorId -> ordinal) so
+      // we can use vectors.getVector(ordinal) for scoring, keeping the same vector-reading path.
+      final Map<Integer, Integer> vectorIdToOrdinal = new HashMap<>(ordinalMap.length);
+      for (int ordinal = 0; ordinal < ordinalMap.length; ordinal++)
+        vectorIdToOrdinal.put(ordinalMap[ordinal], ordinal);
 
-      final float score = metadata.similarityFunction.compare(queryVectorFloat, vec);
-      results.add(new Pair<>(bindRid(loc.rid), scoreToDistance(metadata.similarityFunction, score)));
-      added = true;
+      for (final RID rid : allowedRIDs) {
+        if (seenRIDs.contains(rid))
+          continue;
+        final int[] vectorIds = vectorIndex.getVectorIdsForRid(rid);
+        for (final int vectorId : vectorIds) {
+          final Integer ordinal = vectorIdToOrdinal.get(vectorId);
+          if (ordinal == null)
+            continue;
+          final VectorLocationIndex.VectorLocation loc = vectorIndex.getLocation(vectorId);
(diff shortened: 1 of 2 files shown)


robfrank:
Maintainer review. Rebased onto `main` and pushed follow-up work to this branch; the original commit keeps its authorship.

**The rebase was not mechanical.** `main` picked up the #5558 guard in exactly the lines this PR rewrites: `getVector()` never returns `null` on the page-backed reader, it returns a placeholder, so a location that says "live" but whose document cannot be read (property removed, wrong type, zero vector, record gone) comes back as a well-formed all-ones vector. The new allow-list branch only checked `vec == null`, so on a naive conflict resolution a filtered query would have scored the placeholder as a real vector and returned it with a plausible distance. The guard is now in both paths, in one shared helper so they cannot drift again.

**Two corrections to the analysis in the description.**

1. The scan was not `O(index)` in vector reads. On `main` the allow-list membership test already runs *before* `getVector()`, so a filtered scan only ever read the allowed vectors. What it paid per rejected ordinal was a location lookup - a boxed `Integer` key into a map the size of the index. That is what the 26 ms is made of, and that is the real thing this change removes. Worth knowing, because it also means the read-count is not a valid way to measure the win; the follow-up test counts location lookups instead.

2. The reverse map made the walk `O(index)` again. `new HashMap<>(ordinalMap.length)` filled from every ordinal is built on every call and allocates two `Integer` objects per entry, which on the 50k index in #5748 is the bulk of what the walk was supposed to save. It is not needed: `ordinalToVectorId` is sorted ascending at all three of its producers (they all end in `.sorted()`, because the ordinals have to line up with the order the graph was persisted in), so the reverse lookup is `Arrays.binarySearch` with no allocation at all.

**Also changed**

- Candidates are scored in ascending ordinal order, so a distance tie truncates at `k` to the same RID the full scan would have kept. Without the sort the answer depended on the allow-list's iteration order.
- A crossover guard: an allow-list wider than the index takes the plain scan, which is the cheaper way to answer at that width. The membership test moved into the shared helper so taking that branch still filters - without it a wide allow-list would have returned unfiltered results.

**Tests** - `Issue5748AllowListBruteForceScanTest`, seven cases: parity with the full scan (RIDs, distances, order), truncation at `k`, tie-break, the unreadable-vector guard, the wide allow-list, the ordinal map's sortedness (the binary search rides on it and breaking it would fail silently), and the win itself as a lookup count rather than a duration. Each one was verified to fail against a stubbed fix - five separate mutations, each caught by the test that claims it.

`mvn -pl engine -am test -Dtest='com.arcadedb.index.vector.*Test'`: 272 tests, 0 failures. The five SQL-level vector test classes: 91 tests, 0 failures.

PR #5759, opened 2026-08-02
fix(engine): warn when DOT_PRODUCT index has non-unit vectors (#5750)

### LINKED ISSUE #5750: LSM_VECTOR: a DOT_PRODUCT index accepts non-unit vectors, which silently breaks the metric's score range
Follow-up from the review of #5739 (fix for #5558), where this was raised three times as out of scope for that fix but worth capturing.

### The precondition

JVector's `VectorSimilarityFunction.DOT_PRODUCT` computes `(1 + dotProduct(a, b)) / 2` and documents the precondition in its own javadoc:

> NOTE: this similarity is intended as an optimized way to perform cosine similarity. In order to use it, all vectors must be normalized, including both document and query vectors. Using dot product with vectors that are not normalized can result in errors or poor search results.

ArcadeDB does not enforce it. `CREATE INDEX ... LSM_VECTOR METADATA {"similarity": "DOT_PRODUCT"}` accepts any vectors, and nothing warns when they are not unit length.

### Why it matters more since #5739

With unit vectors `dot` is in `[-1, 1]`, so the score is in `[0, 1]` and **0 is the floor**. #5739 relies on that: a graph node whose vector can no longer be read is scored at `UNREADABLE_NODE_SCORE = 0.0f`, on the basis that 0 is at or below every real score for all three metrics.

For non-unit vectors `dot` is unbounded below, so a real vector can score *below* 0 and a tombstone scored at exactly 0 would outrank it in the beam. The consequence is bounded - a tombstone can waste beam budget but can never enter a result, because `LiveVectorBitsFilter` decides membership, not the score - so this is a recall/efficiency issue rather than a correctness one. But it is a precondition the engine currently leaves entirely to the user, on a metric whose whole point is that it is the fast path for cosine on already-normalized data.

It is also a foot-gun independent of #5739: a user who picks `DOT_PRODUCT` without normalizing gets quietly worse rankings than `COSINE` would have given, which is the opposite of the reason to pick it.

### Options

1. **Warn at index creation / first build** when the vectors seen are not unit length (a sampled check during the graph build is nearly free - it already reads every vector).
2. **Normalize on ingest** for `DOT_PRODUCT` indexes. Changes stored data, so it would need to be explicit.
3. **Refuse** a non-unit vector on a `DOT_PRODUCT` index. Strictest, and a breaking change for anyone relying on today's silence.
4. **Document only**, and note in `UNREADABLE_NODE_SCORE` that the floor claim is precondition-bound (this part is already done in #5739).

(1) looks like the right default: it costs nothing on the path that already reads every vector, it turns a silent quality problem into an operator-visible one, and it breaks nobody.

waterWang:
## Description

Fixes #5750

JVector's `VectorSimilarityFunction.DOT_PRODUCT` javadoc states:

> NOTE: this similarity is intended as an optimized way to perform cosine similarity. In order to use it, all vectors must be normalized, including both document and query vectors. Using dot product with vectors that are not normalized can result in errors or poor search results.

ArcadeDB did not enforce this precondition. With unit vectors `dot` is in `[-1, 1]` and the score is in `[0, 1]`, so a tombstone scored at `0.0` is at or below every real score. For non-unit vectors the dot product is unbounded below, so a real vector can score below 0 and a tombstone would outrank it in the beam — a recall/efficiency issue.

### Fix

Sample the first 1000 vectors during graph build and log a `WARNING` when any are not unit length. The check is nearly free because the build already reads every vector.

This implements **Option 1** from the issue (sampled check during build), as recommended by the issue author.

### Details

- Only runs when `similarityFunction == DOT_PRODUCT`
- Samples up to 1000 vectors from the preloaded/pre-validated vector set
- Logs a single WARNING with the count and percentage of non-unit vectors found
- Magnitude threshold: `|mag - 1.0| > 0.01` (tolerance for float precision)
- No new I/O, no data changes, no breaking change

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -149,6 +149,16 @@ public class LSMVectorIndex implements Index, IndexInternal {
   // enough to never matter and large enough that a small index is fully resident from the first query.
   private static final int MIN_SEARCH_CACHE_SIZE = 8_192;
 
+  // DOT_PRODUCT is the fast path for cosine on already-normalized data, and JVector documents unit length as its
+  // precondition. How many ordinals a rebuild samples to notice that the precondition is not met, and how far a
+  // magnitude may drift from 1 before it counts as violated (float accumulation over a few thousand dimensions
+  // moves the last digits, so an exact comparison would flag correctly normalized data).
+  private static final int    UNIT_VECTOR_SAMPLE_SIZE  = 1_000;
+  private static final double UNIT_MAGNITUDE_TOLERANCE = 0.01;
+  // Squared bounds, so the sample costs one multiply-add per component and no square root at all.
+  private static final double MIN_UNIT_MAGNITUDE_SQUARED = (1.0 - UNIT_MAGNITUDE_TOLERANCE) * (1.0 - UNIT_MAGNITUDE_TOLERANCE);
+  private static final double MAX_UNIT_MAGNITUDE_SQUARED = (1.0 + UNIT_MAGNITUDE_TOLERANCE) * (1.0 + UNIT_MAGNITUDE_TOLERANCE);
+
   // Not final: a compaction swaps in a new data file, and this index is named after its component - see
   // getMostRecentFileName(). Every node names the index after the file it holds, so the leader has to
   // follow its own rename or its schema stops matching the followers that rebuilt it from that file.
@@ -184,6 +194,11 @@ public class LSMVectorIndex implements Index, IndexInternal {
   // Set once the ignored location-cache limit has been reported, so a rebuild does not repeat the warning.
   // compareAndSet, not a plain flag: two threads racing the first call would otherwise both log it.
   private final    AtomicBoolean                 locationCacheCapReported = new AtomicBoolean();
+  // Same idea for the DOT_PRODUCT unit-length warning: an ingest crosses the rebuild threshold every time the
+  // pending set grows by a fraction of the graph, so a million-row load rebuilds a hundred-odd times and would
+  // otherwise repeat the same warning on each of them. Set only when the warning is actually emitted, so an index
+  // that starts out normalized and later is not still gets told once.
+  private final    AtomicBoolean                 nonUnitVectorsReported   = new AtomicBoolean();
 
   // Graph file for persistent storage of graph topology
   // Allows lazy-loading graph from disk and avoiding expensive rebuilds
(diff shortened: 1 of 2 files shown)


robfrank:
Maintainer pass. I rebased this onto `main` and kept only the `LSMVectorIndex.java` change: the branch was also carrying the `pom.xml` surefire change from #5697 and the bot's `bolt/conformance/COMPATIBILITY.md` refresh, which is what made it conflict. Those belong in their own PRs.

The substance is right and the diagnosis in the issue is accurate. Three things I changed on top of it:

1. **The warning had to be once per index, not once per rebuild.** A rebuild fires whenever pending mutations reach `max(mutationsBeforeRebuild, rebuildGraphRatio * graph size)`, so a bulk load rebuilds on the order of a hundred times and the same warning would have been emitted on every one of them. It is now guarded by an `AtomicBoolean` set only when the warning actually fires, the same pattern `locationCacheCapReported` already uses a few lines above for exactly this reason. The guard is also read before the sample loop, so once reported the check costs nothing at all on later rebuilds.

2. **Unreadable ordinals were being counted as non-unit.** `ArcadePageVectorValues.getVector()` never returns null; a deleted or unreadable ordinal yields the all-ones placeholder, whose magnitude is `sqrt(dimensions)`. Sampling it reported a normalization problem the data does not have. The loop now skips it via `isDeletedSentinel()`, which is what the PQ encode loop in the same method already does, and the percentage is computed over the vectors actually examined.

3. **No square root.** The bounds are compared as squared magnitudes, precomputed as constants, so the sample is one multiply-add per component and nothing else. On the wider point: the sample is bounded at 1000 ordinals regardless of index size and runs after the validation phase has already read and cached every vector, so it is not a meaningful cost against a graph build.

I also extracted it into `reportNonUnitVectorsOnDotProduct()` so `buildGraphFromScratchExclusively()` does not grow another inline block.

**Added the regression test** that was missing, `LSMVectorIndexDotProductUnitVectorTest`, with three cases: a DOT_PRODUCT index over non-unit vectors warns exactly once across three explicit rebuilds; the same index over normalized vectors stays silent; and a COSINE index over the same non-unit vectors stays silent, because cosine divides the magnitude out and warning there would train operators to ignore the message on the metric where it means something. Both arms were verified to fail against a stubbed fix: no-oping the method drops the first assertion, and removing the once-guard makes the third rebuild produce a third warning.

Known limitation, not a blocker: vectors added through the incremental live builder are not sampled until the next full rebuild absorbs them. The warning is a build-time diagnostic by design, per option 1 in the issue.

`mvn -pl engine -am test -Dtest=LSMVectorIndexDotProductUnitVectorTest,Issue5558DeletedRegionSearchTest,Issue5559LocationCacheSizeTest,LSMVectorIndexRebuildTest,LSMVectorIndexEuclideanOrderingTest,LSMVectorIndexTest,JVectorIncrementalBuildTest,VectorIndexBuildGraphOnCreateTest` runs 67 tests, all green.

PR #5768, opened 2026-08-02
fix(ci): add continue-on-error to test reporter steps to prevent flaky CI failures

### LINKED ISSUE #5763: CI: a flaky test-reporter step fails the whole test job, so PRs show red with no failing test
## Summary

The `dorny/test-reporter` steps in `.github/workflows/mvn-test.yml` fail intermittently and fail their whole job with them, so a PR shows a red `unit-tests` / `integration-tests` / `ha-integration-tests` even though **every test passed**. The action publishes a check run from the failsafe/surefire XML; it runs no tests, so its failure carries no information about the code.

## Evidence

Observed across three separate jobs on PR #5736, on different commits, always with the same shape — test step green, reporter step red:

| Job | `Run … Tests with Coverage` | Reporter step |
|---|---|---|
| `integration-tests` | success | **failure** (`IT Tests Reporter`) |
| `ha-integration-tests` | success | **failure** (`HA IT Tests Reporter`) |
| `unit-tests` | success | **failure** (`Unit Tests Reporter`) |

Not specific to that PR: at the same time, `worktree-issue-5558-vector-deleted-region` showed the identical signature on `integration-tests` (tests green, reporter red). Several agents were pushing concurrently, which fits check-run API contention as the trigger.

The affected steps are `mvn-test.yml:286` (`IT Tests Reporter`) and `:340` (`HA IT Tests Reporter`), plus the unit-test equivalent.

## Why it matters

The cost is not the red X, it is what the red X teaches. Every one of these has to be triaged by opening the job and reading the step list to discover that nothing failed — several times per PR, on a repo where many PRs are in flight at once. The predictable end state is that a genuine `unit-tests` failure gets waved through as "the reporter again". It also blocks any merge automation keyed on the job's conclusion.

## Suggested fix

The reporter is a reporting convenience, not a gate, so it should not be able to fail the job that ran the tests. Options, roughly in order of preference:

1. **`continue-on-error: true` on the three reporter steps.** One line each, keeps the annotations when the action works, and makes the job's conclusion reflect the tests. This is what the step's role actually is.
2. Move the reporters into a separate, non-required workflow that consumes the uploaded XML artifacts, so their failures never touch the test jobs.
3. Pin/upgrade the action and see whether the current version is more robust — worth checking, but it does not address the design issue that a reporting step can redden a test job.

Option 1 alone would have removed every false red observed above.

Found while working on #5559 / #5736.

waterWang:
## Description

Fixes #5763

`mvn-test.yml` runs its four Maven test jobs with `--fail-never`, so Maven prints `Build failures were ignored.` and exits 0 even when a suite failed. The only step left that turned a failing test into a red job was `dorny/test-reporter`, whose default `fail-on-error` fails the step when the XML contains a failure.

That is a bad place for the build's verdict to live: a GitHub API hiccup in the reporter is indistinguishable from a test failure, and the obvious cure for the hiccup - `continue-on-error` - would equally swallow the failures.

So this PR does both halves:

1. `continue-on-error: true` on the four reporter steps (`unit-tests`, `slow-unit-tests`, `integration-tests`, `ha-integration-tests`), so publishing the annotations can fail without failing the job.
2. A new `.github/scripts/check-test-results.py`, run right after each reporter, which reads the surefire/failsafe XML off the local disk and fails the job when it records a failure, an error, an unreadable report, or no report at all. That is the same rule the reporter applied (`fail-on-error` plus `fail-on-empty`), evaluated locally, so it depends on nothing but the files Maven just wrote.

Without step 2, step 1 alone would report a failing suite as a green job. Every `Reporter`-only job failure sampled from the last 18 red `mvn-test` runs (25 of 25, `unit-tests`, `integration-tests` and `ha-integration-tests`) ended with `Failed test were found and 'fail-on-error' option is set to true`, so in practice the flag alone would have hidden real failures rather than false ones.

## Verification

- `.github/scripts/tests/test-ci-scripts.sh` gains seven cases for the new script: all-green, a failure, an error, phase separation (a failing `surefire-reports` tree must not redden a `failsafe-reports` check), no report at all, an unparseable report, and a skipped-only suite. The `setup` job already runs that harness.
- The cases were confirmed to fail with the script stubbed to `return 0`, then to pass with it restored.
- The script was also run against a real reactor tree: 20,299 tests across 2,405 report files in ~2s of CPU, exit 1 on the modules that had failures, exit 0 on the ones that did not.

### DIFF
--- a/.github/scripts/check-test-results.py
+++ b/.github/scripts/check-test-results.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+#
+# Fails when a surefire/failsafe report records a failing test.
+#
+# The test jobs run Maven with --fail-never, which prints "Build failures were ignored." and exits
+# 0 even when a suite failed. Nothing in the Maven step can turn a failing test red, so until now
+# the only thing that did was the dorny/test-reporter step: it reads the same XML and, with its
+# default fail-on-error, fails itself. That works, but it makes a third-party action that talks to
+# the GitHub API the sole arbiter of whether the build passed - an outage or a rate limit there is
+# indistinguishable from a test failure, and the only way to make the outage non-fatal is
+# continue-on-error, which would equally swallow the failures.
+#
+# This script is that arbiter instead. It reads the reports off the local disk, so it depends on
+# nothing but the files Maven just wrote, and the reporter step above it is free to fail without
+# taking the job with it.
+#
+# An empty result set is a failure too. Reports missing altogether means the suite never ran -
+# a module that did not compile, a fork that died before writing anything - and reporting "no
+# failures found" for a run that produced no evidence is exactly the false green this guards.
+#
+# Usage:
+#   check-test-results.py surefire-reports            # unit tests
+#   check-test-results.py failsafe-reports            # integration tests
+#   check-test-results.py surefire-reports --root DIR # scan DIR instead of the repository
+#
+# @author Luca Garulli (l.garulli@arcadedata.com)
+
+import os
+import sys
+from pathlib import Path
+from xml.etree import ElementTree
+
+# How many failing suites to name before the list is cut short. The reporter step publishes the
+# full detail; this is only meant to say why the job is red without scrolling.
+MAX_LISTED = 25
+
+
+def report_files(root, reports_dir):
+    """Every TEST*.xml written by surefire/failsafe under a directory named reports_dir."""
+    found = []
+    for current, directories, files in os.walk(root):
+        # Nothing puts reports there, and walking them on a full build costs seconds.
+        directories[:] = [d for d in directories if d not in (".git", "node_modules")]
+        if Path(current).name != reports_dir:
+            continue
+        for name in files:
+            if name.startswith("TEST") and name.endswith(".xml"):
+                found.append(Path(current) / name)
+    return sorted(found)
+
+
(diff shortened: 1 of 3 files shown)


robfrank:
Reviewed and extended. The flag was on the right steps, but in this workflow the reporter step *is* the build's verdict, so on its own it would have masked genuine failures.

Evidence: all four test jobs pass `--fail-never`, and job 91462593143 (`ha-integration-tests` on main) logs `[INFO] BUILD FAILURE` followed by `[INFO] Build failures were ignored.`, the Maven step recorded as **success**, and the only red step is `HA IT Tests Reporter` with `Failed test were found and 'fail-on-error' option is set to true` for 3 genuinely failing tests. I sampled every job in the last 18 red `mvn-test` runs whose sole failing step was a reporter: 25 of 25 carry that same message, none an API error. With `continue-on-error` alone those 25 jobs would have gone green.

Added `.github/scripts/check-test-results.py`, run after each reporter, which applies the same rule (`fail-on-error` plus `fail-on-empty`) to the surefire/failsafe XML locally. The reporter can now fail on a GitHub API hiccup without failing the job, and a failing test still fails it. Seven cases in `test-ci-scripts.sh` cover the script, verified to go red when the script is stubbed to a no-op.

Also rebased onto current main.

PR #5769, opened 2026-08-02
fix(engine) #5755: an after-read listener that rewrites a record no longer corrupts it

### LINKED ISSUE #5755: checkForLazyLoading() leaves the buffer at position 0 when an after-read listener rewrites the record
Found while fixing #5723 (PR #5734). Not a live user-facing bug today - it works by accident for the current encryption implementation - but it is a trap that will bite the next `AfterRecordReadListener`.

## The gap

`ImmutableDocument.checkForLazyLoading()` positions the buffer at `propertiesStartingPosition` (1, past the record-type byte) on the normal path, but NOT on the path where an after-read event returns a different record:

```java
protected boolean checkForLazyLoading() {
  if (buffer == null) {
    ...
    buffer = database.getSchema().getBucketById(rid.getBucketId()).getRecord(rid);
    buffer.position(propertiesStartingPosition);        // <- positioned here

    final Record loaded = database.invokeAfterReadEvents(this);
    if (loaded == null) { buffer = null; return false; }
    else if (loaded != this) {
      // CREATE A BUFFER FROM THE MODIFIED RECORD. THIS IS NEEDED FOR ENCRYPTION THAT UPDATE THE RECORD WITH A MUTABLE
      buffer = database.getSerializer().serialize(database, loaded);   // <- NOT positioned
    }
    return true;
  }
  buffer.position(propertiesStartingPosition);
  return false;
}
```

Where that replacement buffer ends up depends on which branch of `BinarySerializer.serializeDocument()` runs:

- **record has a buffer and is not dirty** - `header.position(Binary.BYTE_SERIALIZED_SIZE)`, i.e. position 1. Happens to equal `propertiesStartingPosition`, so everything works.
- **record has no buffer, or is a dirty `MutableDocument`** - goes through `serializeProperties()`, which ends with `header.flip()` and returns the buffer at position **0**.

In the second case the record-type byte is still ahead of the read cursor, so the next reader consumes it as the first bytes of the header-size int.

## Who is exposed

The accessors that read straight after `checkForLazyLoading()` without repositioning:

- `toJSON(boolean)` - calls `deserializeProperties(database, buffer, ...)` immediately
- `getPropertyNames()` - `getPropertyNames(database, buffer, rid)` starts with a bare `buffer.getInt()` for the header size, no seek

Safe because they reposition or rewind themselves: `propertiesAsMap()` / `propertiesAsMap(String...)` (this is why the apparently-redundant `buffer.position(...)` in them is load-bearing - see the comment added in #5734) and `modify()` (`buffer.rewind()` then `copyOfContent()`, which correctly wants the type byte included).

## Why it has not surfaced

The only in-tree listener that returns a different record is the data-encryption hook, and it appears to hand back a non-dirty record that still carries its buffer - the first branch, which lands on position 1 by coincidence. `DataEncryptionTest` therefore passes and always would, for the wrong reason.

## Suggested fix

Reposition once, in `checkForLazyLoading()` itself, after the buffer is replaced:

```java
else if (loaded != this) {
  buffer = database.getSerializer().serialize(database, loaded);
  buffer.position(propertiesStartingPosition);
}
```

That makes the postcondition of the method uniform - "on return with a non-null buffer, the buffer is positioned at the properties" - which is what every caller already assumes. The per-caller repositioning could then be dropped, though leaving it is harmless.

A regression test wants an `AfterRecordReadListener` that returns a **dirty** `MutableDocument` (not the current encryption shape) and then asserts `toJSON()` and `getPropertyNames()` on a record read straight from a scan. That fails today.

lvca:
Closes #5755.

The issue reports one defect. Reproducing it surfaced two more on the same line, one of which is live today rather than latent.

## 1. The buffer was not positioned at the properties (as filed)

`ImmutableDocument.checkForLazyLoading()` replaces the buffer with a freshly serialised one when an after-read event returns a **different** record. `BinarySerializer.serializeDocument()` hands it back at position 1 when it can copy the record's own buffer, but at **0** when it had to re-serialise the properties of a dirty `MutableDocument` (the closing `header.flip()`). In the second case the record-type byte is still ahead of the read cursor, so `toJSON()`, `getPropertyNames()`, `toMap()`, `has()` and `get()` consume it as the first byte of the header size.

It stayed invisible because the only in-tree listener of that shape decrypts **vertices**, and `ImmutableVertex.checkForLazyLoading()` re-parses the fixed edge-pointer prefix from position 1 afterwards, repositioning as a side effect. Plain documents have no such second pass. `toJSON()` returned `null` for every property.

## 2. The record aliased the serializer's per-thread scratch buffer (new, and live)

For a dirty record, `serializeDocument()` builds into `DatabaseContext.getTemporaryBuffer1()` - the per-thread scratch buffer that `getTemporaryBuffer1()` **`clear()`s and hands out again** on every call. Keeping it as the record's own buffer means the next unrelated `save()` on the same thread rewrites the record's content underneath it. The regression test catches the record answering

```
expected: "decrypted_value"
 but was: "9876543210987654321098765432109876543210"
```

which is a different record's property value. This one is not latent: it hits the documented encryption recipe (`RecordEncryptionTest`) as soon as a decrypted record is read and then held across another write, **vertices included** - `ImmutableVertex` was shielded from defect 1 but not from this one.

Fix: `serialize(...).getNotReusable()`. That copies exactly when the buffer is `reusable` (the scratch buffer is `new Binary(8192, true)`) or when the array does not fit the content exactly, and returns `this` otherwise - so the dirty branch is isolated while the already-private buffer of the copy branch costs nothing. Note `Binary.copy()` would **not** have worked at all: it deliberately shares the underlying `byte[]`.

## 3. `BaseRecord.reload()` diverged on the same contract (new)

`reload()` handles the identical "the after-read listener returned a different record" case, but took the returned record's own `getBuffer()`. On a dirty `MutableDocument` that buffer still holds the **pre-modification** content, so a reload silently threw the listener's work away and handed back the raw stored value - ciphertext, for the encryption recipe this hook exists for. It is also plainly `null` for a record the listener built from scratch rather than by `modify()`, which was an NPE. It now goes through the serializer like `checkForLazyLoading()` does.

**Semantic narrowing worth knowing about:** the old `reload()` path would have accepted a listener returning a different *immutable* record with a valid buffer; the serializer casts to the mutable types, so that now raises `ClassCastException`. This aligns `reload()` with the contract `checkForLazyLoading()` has always enforced, and `AfterRecordReadListener.onAfterRead()` now documents it (a returned record that is not the one received must be mutable; `null` filters the record away). Covered in the release notes.

## Why this shape of fix

The issue proposes repositioning after the replacement, which is right, but stops at symptom 1. The fix here makes the **postcondition of `checkForLazyLoading()` uniform and documented** - *on return with a non-null buffer, the buffer is positioned at the start of the properties* - which is what every accessor already assumes, and then makes the sibling path (`reload()`) obey the same contract, so the two cannot drift apart again. `BaseRecord.reload()` documents that positioning the installed buffer is the subclass's job. The per-caller reposition in `propertiesAsMap()` is kept deliberately: it is one field store, and the failure mode there is a **silently empty** map (that is how #5723 made `copyType()` copy empty records).

## Tests

Six regression tests in `ImmutableDocumentLazyLoadingInconsistencyTest`, each verified to **fail** with the fix reverted:

- `afterReadListenerReturningADifferentRecordLeavesTheBufferAtTheProperties` - the filed defect, across `toJSON()` / `getPropertyNames()` / `toMap()` / `has()` / `get()` / `propertiesAsMap()`, by scan and by RID
- `afterReadListenerReturningADifferentRecordDoesNotShareTheSerializerScratchBuffer` - defect 2
- `afterReadListenerRewritingAVertexIsAlsoIsolatedFromTheScratchBuffer` - defect 2 on `ImmutableVertex`
- `reloadKeepsWhatTheAfterReadListenerReturned` - defect 3, documents
- `reloadKeepsWhatTheAfterReadListenerReturnedOnAVertex` - defect 3 on the vertex shape, which is what the encryption recipe is written on
- `reloadAcceptsARecordTheAfterReadListenerBuiltFromScratch` - the NPE half of defect 3; on the old path it reproduces `Cannot invoke "Binary.copy()" because the return value of "RecordInternal.getBuffer()" is null`

With the fix reverted, 3 of the 11 tests in the class fail on the `reload()` half alone.

One caveat worth recording: a first attempt at the vertex test, written inside `RecordEncryptionTest`, turned out **vacuous** - under `REPEATABLE_READ` the scan hands back a `MutableVertex` from the transaction record cache, so `checkForLazyLoading()` never runs at all. It was rewritten against a genuinely lazy `ImmutableVertex` (`lookupByRID(rid, false)` in a fresh transaction), where it does discriminate.

Sweep: 4314 engine tests over `database` / `event` / `graph` / `serializer` / `query.sql` / `schema` / `index`, 0 failures.

## Follow-up

Filed as #5770: `serializeProperties()` writes to the external bucket for `EXTERNAL` properties, so an after-read listener returning a dirty record with an external property triggers writes - and orphan deletions - on a read path. That predates this change on the `checkForLazyLoading()` side and is now shared by `reload()`, which is why it is filed rather than folded in.


lvca:
Thanks - all five points were valid and are addressed in e5a3930.

**`getNotReusable()` instead of `copyOfContent()`** (minor 1). Good call, and it is strictly better than a conditional: `getNotReusable()` copies exactly when the buffer is `reusable` (the scratch buffer is `new Binary(8192, true)`) or when the array does not fit the content exactly, and returns `this` otherwise - so the dirty branch is still isolated and the clean branch no longer pays a second `Arrays.copyOfRange`. Applied at both sites.

**Positioning contract on `BaseRecord.reload()`.** Documented: the installed buffer is deliberately unpositioned, positioning is the subclass's job because only the subclass knows its prefix length, `BaseDocument.reload()` is where it happens, and a future direct subclass that does not override would inherit an unpositioned buffer.

**Mutable-record requirement** (minor 2). Documented on `AfterRecordReadListener.onAfterRead()` - a returned record that is not the one received is re-serialized by the caller, so it must be mutable (typically via `record.modify()`, or built from scratch); a different *immutable* record raises `ClassCastException`; `null` filters the record away. Left as a documented constraint rather than a defensive branch - the alternative would be silently accepting a shape the serializer cannot round-trip.

**Both test gaps closed**, and both verified to fail with the fix reverted:

- `reloadAcceptsARecordTheAfterReadListenerBuiltFromScratch` - listener returns `database.newDocument(...)`, never persisted. On the old path this reproduces the NPE exactly: `Cannot invoke "com.arcadedb.database.Binary.copy()" because the return value of "com.arcadedb.database.RecordInternal.getBuffer()" is null`.
- `reloadKeepsWhatTheAfterReadListenerReturnedOnAVertex` - `reload()` on the vertex shape (`ImmutableVertex.reload()` nulls the buffer before delegating), which is the shape the encryption recipe is written on.

With those two the reverted-fix run fails 3 of 11 (`reloadKeepsWhatTheAfterReadListenerReturned`, the vertex one, and the from-scratch NPE), so the `reload()` half of the change is now pinned from three directions.

Re-ran 669 tests over `database` / `event` / `graph` / `serializer`, 0 failures.

lvca:
All three notes were valid and are addressed.

**1. Description drift.** PR body reconciled - it now names `getNotReusable()` (with the reason it beats an unconditional `copyOfContent()`, which is the point you made in the first review) and lists all six tests instead of four.

**2. The `reload()` narrowing.** Written up in `docs/release-26.8.1.md` (a3391cb) with its own subsection, since it is the kind of change an embedded caller meets at runtime rather than at compile time: a listener returning a different *immutable* record with a valid buffer used to be accepted on the `reload()` path and now raises `ClassCastException`. The note states it as an alignment with the contract `checkForLazyLoading()` has always enforced, and points at the `AfterRecordReadListener` Javadoc for what a listener may return. Also called out in the PR body now.

**3. EXTERNAL write-on-read.** Filed as #5770. It records that the exposure is now shared by `reload()` as well as `checkForLazyLoading()`, that the orphan-cleanup block can *delete* external records as a side effect of a read, and that SQL READ triggers cannot reach it (`TriggerListenerAdapter.onAfterRead()` returns the same record, so it never takes the `loaded != this` branch) - an embedded listener is required, which makes the encryption recipe on a type with an external property the realistic way to meet it. Three directions sketched, with a read-only serialization mode as the preferred end state rather than the tempting "skip external handling on reads", which would produce a buffer whose external pointers do not resolve.

No code changes in this round, so the test position is unchanged from the previous one.

lvca:
Thanks - LGTM taken, no code changes in this round. On the three flagged points:

**1. Public API narrowing.** Agreed and left as is. It is the deliberate half of defect 3: the `reload()` path is now held to the contract `checkForLazyLoading()` has always enforced, rather than the two paths accepting different shapes of returned record. Documented on `AfterRecordReadListener.onAfterRead()` and in `docs/release-26.8.1.md` under its own subsection, since it surfaces at runtime rather than at compile time.

**2. Vertex edge-pointer staleness after `reload()`.** You are right that it is pre-existing and not a regression, but it turned out to be **observable rather than latent**, so I filed it as #5771 with a reproduction rather than leaving it as a note. Confirmed:

```
before:        outEdges=null   countEdges=0
after reload:  outEdges=null   countEdges=0     <-- stale
fresh lookup:  outEdges=#2:0   countEdges=1     <-- the truth
```

`countEdges()` is wrong too, so it is not confined to the internal accessor - and `reload()` is exactly the API a caller uses to pick up another transaction's changes. The root cause is the `buffer.position() == 1` probe in `ImmutableVertex.checkForLazyLoading()`: it infers "prefix not yet parsed" from a position, which stops being true once the prefix has been parsed once (`propertiesStartingPosition` is then 25, so the re-parse never fires). The issue proposes replacing the heuristic with an explicit `positionAtProperties()` hook that re-derives the prefix from the current buffer, which also removes the stale-`propertiesStartingPosition` hazard `ImmutableEdge` carries. Kept out of this PR - it is a different defect with its own test surface.

**3. #5770** - agreed, deferred deliberately.

Final state: 3 files of production change (`ImmutableDocument`, `BaseRecord`, `AfterRecordReadListener`), 6 regression tests, 1 release note, 2 follow-ups filed (#5770, #5771).

lvca:
@claude please review latest commit

Issue #5764, opened 2026-08-02
Follow-ups from #5680: point the repair advice at CHECK DATABASE RECORD, keep the conflict's cause, and two checker asymmetries


lvca:
Follow-ups left behind by #5680, whose two halves landed separately: #5707 made `deleteVertex` strict, and #5710 added the `CHECK DATABASE RECORD <rid>` scope. Each was reviewed against the state of `main` at the time, so a few seams between them were never closed. None is a defect in either merge; they are refinements the split made easy to miss.

Ordered by value.

## 1. The repair advice never names the scoped command it now has

Every place `deleteVertex` tells an operator how to recover predates the `RECORD` scope, so they all advise the whole-database (or whole-type) form:

- `GraphEngine.java:739` — `"the reference survives, run a database check to repair it"`
- `GraphEngine.java:939` — `"...edges behind the damage survive - run a database check to repair them"`
- `GraphEngine.java:921` (comment) — `"CHECK DATABASE ... FIX rebuilds the chain from the surviving edge records"`
- and the retryable path at `:902` rethrows the `ConcurrentModificationException` bare, so what the operator sees is `getEdgeHeadChunkForWrite`'s message, which does not mention recovery at all.

This is the case the `RECORD` scope was added for. A vertex whose chain is genuinely broken is undeletable until the adjacency is rebuilt, and the operator holds the one piece of information that makes the repair cheap — the RID — while being pointed at a run that costs two full passes over the vertex type plus an edge sweep.

Suggested: name the command with the RID substituted, e.g.

```
run `CHECK DATABASE RECORD #12:3 FIX` to rebuild the edge list from the surviving edge records, then retry the delete
```

Worth stating the limit in the same breath, since the scope does not bound everything: the edge sweep still runs once per distinct vertex **type** named.

## 2. `ConcurrentModificationException` has no `(message, cause)` constructor

`NeedRetryException` already declares one; `ConcurrentModificationException` exposes only `(String)`. Six sites in `GraphEngine` alone construct one from a caught exception (`:411`, `:825`, `:829`, `:843`, `:965`, `:1530`) and each drops the original stack.

A retryable conflict is normally absorbed and never seen, so the single run where one *does* surface is the retry-exhausted one — i.e. exactly the run whose stack trace has to be diagnosable, and the one that currently arrives with the cause discarded.

One-line change; the plumbing already exists on the parent.

## 3. `checkDocuments` and `checkScopedDocuments` report different totals for the same finding

[...]

PR #5772, opened 2026-08-02
fix(engine) #5760: deleting a vertex no longer disconnects its edges from itself

### LINKED ISSUE #5760: deleteVertex disconnects each edge from the vertex it is deleting, walking a chain it drops moments later
Carried over from #5725 item 2, which was closed by #5744 (that PR fixed item 1, the ghost-edge race, and deliberately left this alone). Re-filing it because the original wording framed it as an allocation problem, and having now been through that code I think the allocation is the smaller half.

## What the original issue said

`GraphEngine.deleteVertex` collects every edge into an `ArrayList<Identifiable>` before deleting any of them, which on a promoted super-node with millions of edges is a large allocation on a path that could stream. #5725 correctly noted this is not mechanical: the two-phase collect-then-delete exists **because** the removals relink and delete chunks underneath the walk, so streaming means walking a list while it is being restructured - the class of problem #5155, #5670 and #5680 were all about.

That is all still true. Two things have changed since.

## 1. The list is no longer the dominant cost

#5744 made `deleteVertex` pin every chunk page of the edge list in the transaction. Those pages were already going to be pulled in - `deleteAll` deletes every one of those chunks - so peak footprint did not change, but it does mean the retained **pages** now dominate a super-node delete, not the `ArrayList` of references. Streaming the collection walk would not remove them.

So the original framing ("a large allocation that could stream") no longer points at the biggest term, and a streaming rewrite would carry the full #5155/#5670/#5680 risk for a fraction of the benefit.

## 2. There is a bigger, cheaper win in the same method

`deleteVertex` collects the edges, then calls `edge.delete()` on each. That reaches `GraphEngine.deleteEdge`, which disconnects the edge from **both** endpoints:

```java
final VertexInternal vOut = resolveEndpointToDisconnect(edge, Vertex.DIRECTION.OUT);
if (vOut != null) {
  final EdgeLinkedList outEdges = getEdgeHeadChunkForWrite(vOut, Vertex.DIRECTION.OUT);
  if (outEdges != null)
    outEdges.removeEdge(edge);
}
// ... same for IN
```

One of those two endpoints is always **the vertex being deleted**. `EdgeLinkedList.removeEdge` walks the chain from the head probing each chunk for the edge, so that is an O(chunks) walk per collected edge, and for a vertex of degree D spanning C chunks the self side alone costs O(D x C).

It is also pure waste: `deleteRemainingChunks` -> `deleteAll()` deletes every one of those chunk records immediately afterwards, so every entry painstakingly removed one at a time is dropped wholesale a moment later. The far-endpoint removal is the one that matters - that is the back-reference #5670 is about - and it is unavoidable.

## Suggested direction

Teach the vertex-delete path to skip the self side: disconnect each collected edge from the vertex at its **other** end only, then drop the chunks. Sketch: a `deleteEdge(edge, skipEndpoint)` overload (or an explicit `disconnectFarEndpoint`) taking the RID of the vertex whose list is about to be dropped.

Cheaper than streaming, and it removes work rather than restructuring a walk that three separate issues have already bitten.

## Things to be careful about

- **Self-loops.** `A -> A` is reachable from both of A's lists and both are dropped by `deleteAll`, so skipping both sides is correct - but it is the shape that has caught this code before (see `deletingAVertexWithASelfLoopRemovesItCleanly`).
- **Do not weaken #5670/#5680.** Skipping the self-side removal must not skip the self-side *read*: the strict collection walk over the vertex's own list is what refuses to delete a vertex whose list is not fully readable. That coverage lives in `collectEdgesToDelete` and stays.
- **`force`.** The tolerances are load-bearing and hard-won (#4420, #4432, and the mid-chain-break case in #5744). Any change here needs the same standard of deterministic tests, as #5725 said.
- **Measure first.** I have read the path but not profiled it; the quadratic term above is an analysis, not a measurement. A super-node delete benchmark should come before the change so the win is a number rather than an argument.

## Not urgent

This is a scalability note, not a correctness bug. Nothing is wrong with the current behaviour; it is just doing work it does not need to do, on the one shape where that work is worst.

lvca:
Closes #5760, and the leftover item 2 of #5725.

## What the issue asked for

`GraphEngine.deleteVertex` walked its own edge lists, then handed every edge to `deleteEdge`, which disconnects it from **both** endpoints. One of those two is always the vertex being deleted - and its lists are dropped in their entirety a few lines later by `deleteRemainingChunks`. So for every edge the delete walked a chain from the head probing each chunk for that entry, anchored and copied the chunk that held it, compacted it and wrote it back, over a list about to be deleted wholesale.

Each edge is now disconnected from the vertex at its **other** end only, via a new `deleteEdge(Edge, RID skipEndpoint)`.

Matched by the RID recorded **on the edge**, not by which list the edge was found in: an entry whose `out`/`in` does not name the vertex being deleted is a reference into somebody else's list and is disconnected normally. A self-loop skips both sides, which is correct for the same reason as one side - `A -> A` is reachable from both of A's lists and both are dropped.

The self-side **read** is untouched. The strict collection walk (a chunk that cannot be read is a retryable conflict, only `force` absorbs it - #5670/#5680) and the page pinning plus head re-check (#5725) are what decide which edges exist to disconnect, and all of it stays exactly as it was.

## The better alternative the issue argued against

#5760 explicitly rejected streaming: *"a streaming rewrite would carry the full #5155/#5670/#5680 risk for a fraction of the benefit"*. That reasoning rests on one premise - *"the removals relink and delete chunks underneath the walk"* - and **that premise is only true because of the self-side removal**. Once the self side is skipped, `deleteVertex` writes the far endpoints' lists and the edge records and **never the list it is reading**, so the iterator is stable and one pass is legal.

So the walk now streams, and the `ArrayList` of every edge the vertex has - one live `Edge` object per edge, retained for the whole delete - is gone. On a million-edge super-node that accumulator was tens of megabytes of retained heap; there is no per-degree allocation on the path at all any more. That is the other half of #5725 closed as a *consequence* of this fix rather than as a risk taken alongside it.

The per-edge deletion sits deliberately **outside** the `try` that reads the list: the two tolerances there are about *this* vertex's list, and an exception raised while disconnecting an edge from the vertex at its other end has a different meaning and its own policy. Folding them together would have turned a corrupt neighbour into "this vertex's list is corrupt, delete it anyway".

## Measured

The issue asked for a benchmark before the change (`SuperNodeDeleteBenchmark`, `@Tag("benchmark")`). 100k edges into one hub, Apple M-series laptop:

| layout | before | after |
| --- | --- | --- |
| promoted super-node (striped, #5156) | 2374 ms | **446 ms** |
| classic single chain | 350 ms | **308 ms** |

The striped layout is where the waste lived: `StripedEdgeList.removeEdge` resolves one chain per generation for every single removal, and generation 0 is the whole pre-promotion chain. On the classic layout the emptied head chunk already kept the per-edge probe short - so the issue's "O(D x C)" analysis holds for the striped case and much less so for the classic one, which is worth recording since the issue flagged that term as unmeasured.

## Tests

`Issue5760VertexDeleteSelfSideSkipTest` (8 tests). The mechanism tests assert **exact** `DatabaseStats` counts, since they are assertions about work *not* done:

- N outgoing edges -> exactly N endpoint resolutions (was 2N)
- a vertex of only self-loops -> **0** resolutions (was 2N)
- reads before the first edge record is deleted < N, which a two-phase collection cannot satisfy
- a promoted super-node -> one resolution per edge, never the hub itself

**Four of these fail on `main` and pass on the branch.** The correctness tests - mixed shape with parallel edges, a mutual pair and a self-loop; a vertex created and deleted in the *same* transaction (so every chunk the walk reads is the transaction's own pending copy); lightweight edges; `force` - pass on both, which is the point: behaviour is unchanged.

`Issue5760ServerVertexDeleteIT` covers the wrapper path. `LocalDatabase` builds every record with its `wrappedDatabaseInstance`, so on a server `edge.getDatabase()` really is the `ServerDatabase`, and the delegations added there and in `RaftReplicatedDatabase` are on the live path rather than a formality. No existing server test deletes a vertex that has edges. The IT asserts the fixture hands out a `ServerDatabase` before doing anything else, so it cannot pass vacuously on the embedded path.

## Files changed

| File | Why |
| --- | --- |
| `engine/.../graph/GraphEngine.java` | `deleteEdge(Edge, RID)`, `disconnectEndpoint`, streaming `deleteEdgesOf`, `tolerateUnreadableEdgeList` / `tolerateUndecodableEdgeList` / `deleteEdgeOfDeletedVertex` |
| `engine/.../database/DatabaseInternal.java` | new `deleteEdgeSkippingEndpoint` |
| `engine/.../database/LocalDatabase.java` | implementation + private `deleteRecordNoLock(Record, RID)` |
| `server/.../ServerDatabase.java`, `ha-raft/.../RaftReplicatedDatabase.java` | delegations for the new interface method |
| `engine/src/test/.../SQLMethodTransformTest.java` | **not unrelated**: this test builds an anonymous `DatabaseInternal`, so adding an abstract method to that interface requires a stub here. Five lines, no behavioural content. |
| `engine/src/test/.../Issue5760VertexDeleteSelfSideSkipTest.java` | new, 9 tests |
| `engine/src/test/.../SuperNodeDeleteBenchmark.java` | new, `@Tag("benchmark")` |
| `server/src/test/.../Issue5760ServerVertexDeleteIT.java` | new, wrapper-path coverage |
| `docs/release-26.8.1.md` | release note |

## Verification

- All 415 tests in `com.arcadedb.graph` green, including `Issue5670EdgeDeleteDanglingBackRefTest`, `Issue5680VertexDeleteEdgeCollectionTest`, `Issue5725GhostEdgeOnAppendRaceTest`, `Issue5155EdgeRemovalAnchorTest`, `Issue4420TolerantDeleteTest`, `Issue4432CorruptVertexDeleteTest`, `SuperNodeStripingTest`, `GraphDatabaseChecker*`.
- Full `arcadedb-engine` suite: 14872 tests. One failure, `Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent`, which passes in isolation and touches only plain documents - it lost its retries while another build was saturating the CPU.
- `ACIDTransactionTest` green on the branch (350 s) and on `main` (622 s); it is simply slow.
- `arcadedb-server` IT and `arcadedb-ha-raft` green.

## Considered and not done

Grouping the far-endpoint removals per neighbour (`EdgeLinkedList.removeVertex` drains all entries for one vertex in a single walk) would help the "many parallel edges to the same neighbour" shape. It needs a `Map<RID, List<Edge>>` over the whole degree - reintroducing exactly the per-degree allocation this change removes - and buys nothing on the common super-node shape, where the neighbours are distinct. Left out deliberately.

https://claude.ai/code/session_01KJ6aLGLNwsPTYejUxsYGiS


lvca:
Thanks — acting on the review:

**Fixed (2, stale name):** `collectedHeads` / `collectedOutHead` / `collectedInHead` named a phase that no longer exists. Renamed to `headsAtWalkStart` / `walkedOutHead` / `walkedInHead`, and the conflict message now says the delete "did not see" the appended edge rather than "did not collect" it. Pushed as `5a6e0b0c8`; no behaviour change, `com.arcadedb.graph` re-run green (402 tests).

**Not changed, with reasons:**

**(1) Write interleaving during the walk.** Agreed this is the genuinely new ordering, and your read of why it is safe matches mine. Two things pin it beyond the reasoning: the vertex's own chunks are in a different bucket from both the edge records (edge buckets) and the far endpoints' lists (the neighbour's `_in_edges`/`_out_edges`), so nothing written mid-walk can land on a chunk the iterator is standing on; and where a far write *does* share a page with a walked chunk (same bucket, different record), the read goes through the transaction's own page copy, where the walked chunk's slot is untouched. `aVertexCreatedAndDeletedInTheSameTransactionLeavesNothingBehind` is indeed the hardest case, which is why it is there.

One consequence worth recording since it is not obvious: the IN list is now anchored *after* the OUT edges are deleted, rather than immediately after the OUT anchor. That widens the window in which a concurrent append can flip the IN head between `headsAtWalkStart` and the anchor. It is safe in one direction only — an append landing before the anchor is seen and deleted by the walk, one landing after fails the commit-time version check — so the worst case is a redundant retry, never a missed edge. `checkEdgeListHeadsUnchanged` still compares against the head read at the very start, so a flip is caught either way.

**(3) Redundant `edge.getDatabase()`.** These are field reads on the record, not lookups, and they predate this PR inside `resolveEndpointToDisconnect`. Threading a `Database` parameter through `disconnectEndpoint` into it would add churn to the most delicate method on the path for no measurable gain, so I left it.

**(4) Javadoc volume.** Noted rather than actioned, as you intended.

On the last point: `mvn verify` results are in the PR description, and since posting it I also confirmed the one full-suite failure (`Issue5279ConcurrentUpdateTest`) passes in isolation — it lost its retries while a parallel build was saturating the CPU, and it touches only plain documents, no graph code.

lvca:
Second review addressed in `b3fb285e3`.

**(1) + (2) Self-loop double-processing and the lightweight gap — measured, not argued.**

The review left the key question open ("depending on whether the iterator materialises the record"), so I measured it instead of reasoning about it. `onBeforeDelete` fires **twice**, deterministically, for a self-loop — and for a *lightweight* self-loop too.

Why: `EdgeIterator.hasNext` resolves the edge with `loadContent=false`, and a lazy handle to a record deleted earlier in the same transaction still resolves. So it is not filtered out, the second walk yields it, and the pipeline runs again — disconnection skipped on both sides, `bucket.deleteRecord` absorbing the `RecordNotFoundException`.

The part that matters for your "latent surprise" concern: **this is not new.** I ran the same probe against the parent commit (`c7349f8b7`) — two events there, two events here, in both shapes. The old two-phase walk collected the self-loop from each list and called `delete()` on it twice, and `deleteRecordNoLock` fires `onBeforeDelete` before anything can short-circuit. Streaming changed nothing.

Rather than a one-line comment I pinned it as an **exact count** in both shapes (`aSelfLoopIsWalkedFromBothListsSoItsDeleteEventFiresTwice`), which also closes (2) — the lightweight self-loop now has coverage, and it is the interesting one, since with no record to go missing it is the shape where a "the second pass short-circuits on the deleted record" assumption would be wrong. A future change that makes it fire once, or three times, now moves an assertion. The call site carries the note.

**(3) `deleteEdge(Edge, RID)` visibility.** Taken as the contract-wording option you offered, since visibility cannot be tightened: `LocalDatabase` dispatches to it from another package, which is the only reason it is public. The Javadoc now states what it is *not* — skipping is sound only because the caller destroys the skipped list immediately afterwards, so a `skipEndpoint` whose list *survives* produces exactly the dangling back-reference #5670 exists to prevent — and that a new caller has to satisfy that same precondition.

**(4) Comment density.** Flagged rather than requested, and it matches this file's house style. Left alone.

`com.arcadedb.graph` green at 409 tests.

Note this branch also picked up a merge with `main`: #5764 landed mid-review and reworked the same three error paths this PR restructured, so the resolution combines them rather than picking a side (`otherEndOf`/`withRepairAdvice` moved into `deleteEdgeOfDeletedVertex`, both `tolerate*` helpers adopting the repair advice). One consequence worth calling out: `otherEndOf`'s Javadoc argued a self-loop cannot reach it because "the only caller runs after BOTH lists have been collected" — that premise dies with the two-phase walk. The conclusion survives for a stronger reason, now recorded in its place: #5760 skips the disconnection at both endpoints of a self-loop, so it never reads an edge list there at all.

lvca:
Third review addressed in `1fa7177ae`.

**(1) Self-loop `onBeforeDelete` in the release note.** Agreed — the docs section described the streaming and heap win but not the one thing an operator can actually observe. Added, and stated with the caveats that matter: it is *not new* in this release, the graph outcome is correct either way (the second pass skips both endpoints and the record removal is absorbed), and only a non-idempotent listener notices. Also recorded that it holds for lightweight self-loops, where no missing record short-circuits the second pass.

**(3) `SQLMethodTransformTest.java`.** Confirmed — it belongs, and it is not a merge artifact. That test constructs an anonymous `DatabaseInternal`, so adding `deleteEdgeSkippingEndpoint` to the interface forces a stub there. Five lines, no behavioural content. Fair criticism that the PR description did not say so; it now carries a files-changed table that explains each entry, including that one.

**(2) Public `deleteEdge(Edge, RID)`.** No change requested, and I agree the doc-contract is the right mitigation given the cross-package dispatch requirement.

**(4) Per-edge read-lock re-acquire.** Correct that it is unchanged rather than removed. Worth recording *why* I did not remove it while I was in there: `deleteEdgeSkippingEndpoint` goes through `executeInReadLock` precisely because `edge.delete()` did, and dropping it is not free — `DatabaseAsyncDeleteRecord` calls `deleteRecordNoLock` directly, without the lock, so the async path would lose an acquisition the synchronous path keeps. Making that safe is a separate question about lock ownership on the async delete path, not something to fold into this change.

lvca:
Fourth review addressed in `0fdfc57e4`.

**(1) Stale `{@link #collectEdgesToDelete}`.** Good catch, and worth recording *how* it came back rather than just fixing it: the three references to the old name were updated when `collectEdgesToDelete` became `deleteEdgesOf`, but #5764 landed on `main` mid-review carrying a **fourth** — in `withRepairAdvice`, written against the pre-rename name — and the merge brought it in. So it is a merge artifact, not a missed rename, which is the kind of thing that recurs if the next person merges the same way. Now points at `deleteEdgesOf`, and it is the only occurrence left in the file.

**(3) `@Tag("slow")` — declined, with the number.** I measured rather than guessed: **0.603 s for all 9 tests**, including the 400-transaction promoted-super-node fixture. The CLAUDE.md bar is "noticeably long (large batches, multi-second elapsed time, big payloads)", and the sibling `@Tag("slow")` classes in this same package run 4.3 s (`Issue5147SuperNodeChunkRaceTest`) and 17.5 s (`ConcurrentEdgeAppendMergeTest`). Tagging this one would remove the exact-count regression assertions from the default CI run — which uses `-DexcludedGroups=slow,benchmark` — and buy roughly half a second. The per-transaction fixture shape is deliberate (it grows the chain chunk by chunk the way production does) but it is cheap.

**(2) Duplicate endpoint read.** Agreed and left alone, as scoped.

`com.arcadedb.graph` green at 409 tests.

lvca:
Fifth review addressed in `bf6e75a5f`.

**(2) The `deleteEdge` comment's absolute phrasing.** You were right, and the point turned out to be worth more than a reword, so I probed it before rewriting — a misleading comment and an actual leak are different things and the comment shouldn't guess.

`moveEdge` does call the public `deleteEdge(Edge)` directly, and there the precondition genuinely does not hold: the old edge record's index entries are never cleaned. It does **not** leak today, but for a reason that is an allocation coincidence rather than a guarantee — `moveEdge` re-creates the edge with the same properties, and the bucket hands the just-freed slot straight back, so the stale entry lands on the new record carrying the same key. Measured on a multi-bucket edge type: 41 records, 41 index entries, identical RID before and after.

I did not file a follow-up issue, because I could not construct a case that actually leaks and a speculative report is worse than none. The finding lives at the site instead, with what would break the coincidence (a different bucket, a concurrent allocation taking the slot) and the obligation it puts on any future caller of the public `deleteEdge(Edge)`: come through `deleteRecordNoLock` or clean up after yourself. That is strictly more useful than the flat assertion that was there.

**(1) Duplicate endpoint read** and **(3) doc volume** — both raised without a change requested, and I agree with both as stated.

`com.arcadedb.graph` green at 409 tests.

lvca:
Sixth review: the only actionable ask was to file a tracking issue for the `moveEdge` index-cleanup precondition rather than leaving it in a code comment. Fair point, and you talked me round — I had declined earlier on the grounds that I could not construct a failure, but a *documented latent cliff* is worth tracking even without a reproducer, provided the issue says so plainly.

Filed as **#5779**. It records the measurement (8-bucket edge type, 41 edges: 41 records / 41 index entries / identical RID before and after, `CHECK DATABASE` clean), states up front that no failure could be constructed and that the correctness rests on slot reuse rather than an invariant, and lists the three ways the coincidence breaks — a replacement landing in a different bucket, a concurrent allocation taking the slot, and the sharper case where the slot is later reused by an *unrelated* edge so the stale entry silently returns a wrong record for that key. `UNIQUE` makes it sharper still.

It also carries your suggested guardrail (`DocumentIndexer.deleteDocument` before `deleteEdge`, keeping the "physical removal only" contract of `deleteEdge(Edge)` intact rather than weakening it for one caller), plus two explicitly-unverified things worth checking in the same place: EXTERNAL property values (`cascadeDeleteExternalValues` is skipped on the same route) and the absent delete/create events, which may be intended for a move but are currently a side effect of the route rather than a decision. And a note that a regression test has to assert `index.countEntries()` rather than query results, since a query skips unresolvable RIDs and would hide exactly this.

Nothing else from this round needs a change: the doc volume and the duplicate endpoint read were both raised without a request, and I agree with both as stated.

That closes every point raised across the six rounds. No code change in this one, so the PR is unchanged from `bf6e75a5f`.

lvca:
Seventh review (approve) addressed in `fcb10653a`.

**(3) Concurrency test for the streaming interleave — done.** This was the right call: the interleave is the one genuinely new ordering in this PR (far-endpoint writes used to run in a second pass, after the walk finished) and nothing put another transaction on those lists at the same time.

`deletingAVertexWhileItsNeighboursListsAreBeingAppendedTo` deletes a hub while four threads append edges to the very neighbour lists the delete is removing its back-references from. Two invariants, both required or the test says nothing: no back-reference to the hub may survive (the #5670/#5680 invariant, which a delete that lost its walk would break), and every append that **committed** must still be reachable (a delete rebuilding a neighbour's chunk from a stale copy would erase them — the lost update #5147 was about).

Worth flagging how the first version was wrong: it asserted only that appends committed, which is a *different claim* from "appends raced the delete" — every one of them could have landed before the deleter started, and the test would have passed while proving nothing. There is now a warm-up latch holding the delete until each appender has committed one edge, the appenders keep going until the delete finishes, and the assertion is on `appendsDuringDelete`, counted while the deleter is inside its transaction. Tagged `slow` (6 rounds of real contention). Stable 5/5 before the overlap change and 3/3 after.

**(4) `moveEdge` follow-up** — already filed as **#5779** before this review, with the measurement and the three ways the allocation coincidence breaks.

**(1) `while (iterator != null)`** and **(2) self-loop double pipeline** — both scoped as no-action, and I agree.

`com.arcadedb.graph` green at 420 tests.

---

**On the two red checks, since they are not this PR:**

`ha-integration-tests` has now failed with **three different tests across three runs** (`Issue5410AbandonedTicketReleaseIT` + `Issue5569SlotMergeDeleteRaftIT`, then `RaftServerDatabaseSqlScriptIT.executeSqlScript`), and `main` fails the same job on three consecutive runs — including one where the failing pair is *identical* to this PR's. The latest is a Raft quorum timeout with a replica schema mismatch, in a test with zero graph-delete usage. `unit-tests` and `builder-tests` both cleared on re-run, confirming those were flaky/infrastructure.

`codecov/project` is -0.03% (67.81% vs 67.84%) with **Hits down 19** — coverage lost rather than lines added, which is the aborted HA job, not this diff. `codecov/patch` passes. One thing I checked rather than assumed: codecov reports `ServerDatabase` patch coverage as 0.00%, which would mean the wrapper IT never ran — it does, `integration-tests` executes `Issue5760ServerVertexDeleteIT` and it passes, so that figure is an attribution quirk.

lvca:
Eighth review addressed in `ca092e88f`.

**(1) `moveEdge` tracked issue — it exists: #5779**, filed after the sixth round, with the measurement and the three ways the allocation coincidence breaks.

But you are the *third* reviewer in a row to ask for it, and that is worth treating as data rather than as three misses. The problem was the comment: it carried the finding and the numbers but never named the issue, so anyone meeting it had to work out for themselves whether it was tracked. It names #5779 now. Your phrasing — "a code comment isn't discoverable the day the coincidence breaks" — is exactly right, and it applied to the comment's failure to point anywhere as much as to the finding itself.

**(3) `while (iterator != null)`.** Raised in the seventh round too, where it was scoped as "not worth changing given the comment". Twice-flagged is evidence it genuinely trips readers rather than being one person's taste, so I took your form: an explicit `if (iterator == null) return edges != null;` before a plain `while (true)`. The reference was never reassigned and every exit was already a `break`, so this is the honest shape.

**(4) Concurrency test timing.** Good catch, and it was a real latent flake rather than a theoretical one — the assertion was **per round**. A delete window short enough that no append happens to commit inside it proves nothing, but it is not a defect either, so on a slow runner the test would have failed for a reason unrelated to what it guards. It is now counted across all rounds: one genuine overlap is the entire non-vacuity argument, and the per-round timing sensitivity is gone.

**(2) Public API surface.** Staying as documented — `LocalDatabase` dispatches to it from `com.arcadedb.database` while `GraphEngine` is in `com.arcadedb.graph`, so package-private with an internal bridge is not available without moving one of them. Agreed that prose is the only enforcement; that is why the javadoc leads with what the method is *not*.

`com.arcadedb.graph` green at 420 tests, concurrency test stable 3/3 after the change.

lvca:
Ninth review: no action needed, and recording why for each note rather than staying silent on them.

**(1)** Self-loop double `onBeforeDelete` — documented at the call site, pinned as an exact count for both record-backed and lightweight shapes, and in the release note for listener authors. Done in rounds 2 and 3.
**(2)** `moveEdge` precondition — tracked as #5779, and the site comment now names it. Agreed the fix belongs there, not here.
**(3)** Documentation density — noted, and your example is well chosen: the stale `collectEdgesToDelete` link was reintroduced by the #5764 merge rather than by a missed rename, which is precisely the sync cost you describe. It is the reason I re-run the graph suite after every merge into this branch instead of assuming it still holds — #5773 reworked `GraphDatabaseChecker`, which every correctness test here asserts through.
**(4)** Duplicate endpoint read — pre-existing and negligible, leaving it.

That is nine rounds with nothing blocking. The only thing still outstanding is CI on `ca092e88f`, which is mid-run.

lvca:
## CI complete on `ca092e88f` — everything green except one job, and that one is proven pre-existing

**Passing:** `unit-tests`, `slow-unit-tests`, `integration-tests`, `builder-tests`, `build-and-package`, all e2e suites, CodeQL, Codacy, Meterian, `coverage-report`, `codecov/patch`.

Two of those matter specifically:
- **`slow-unit-tests`** is where the new `@Tag("slow")` concurrency test runs. Green on a slow CI runner, which is exactly where the round-8 reviewer expected the per-round overlap assertion to flake — the cross-round rewrite holds under real CI timing, not just locally.
- **`integration-tests`** carries `Issue5760ServerVertexDeleteIT`, so the `ServerDatabase` wrapper delegation is exercised and green on CI.

**Failing: `ha-integration-tests`** — 207 tests, 1 failure + 1 error:

| Test | Failure | Touches `deleteVertex`? |
| --- | --- | --- |
| `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` | `AssertionError: Update returned null: record=2 seq=12` | **No** — operates on `document TYPE Doc`; no vertices or edges exist in the fixture |
| `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied` | `Node 0 must hold both vertices, expected 2 but was 1` | **No** — creates vertices and counts them; zero delete calls in the file. Raft abandoned-entry fault injection |

**This exact pair fails on `main`** — run [30761436982](https://github.com/ArcadeData/arcadedb/actions/runs/30761436982), same two tests, without this change. `Issue5569` also failed on main run 30759219297.

Across four runs of this PR the job has failed with **four different tests** (`Issue5410`+`Issue5569`, then `RaftServerDatabaseSqlScriptIT`, now `Issue5410`+`Issue5569` again), and `main` fails the same job on three consecutive runs. `unit-tests` and `builder-tests` both failed on an earlier run of this PR and cleared without a code change, which is the same picture.

`codecov/project` is -0.03% (67.81% vs 67.84%) with **Hits down 19** — coverage lost rather than lines added, i.e. the aborted HA job. `codecov/patch` passes.

---

That closes the loop: nine review rounds with no blocker, no conflict, and every remaining CI failure traced to a test that cannot reach the changed code and that fails identically on `main`. Ready to merge from my side.

PR #5775, opened 2026-08-02
fix(engine) #5765: creating a manual index no longer fences the database, and a guarded request answers for the index asked for

### LINKED ISSUE #5765: Follow-ups from #5675: ManualIndexBuilder still silently drops indexes, plus three smaller items
Follow-ups discovered while fixing #5675 (PR #5740, merged as 4966b4bde). All four verified against `main` at 79b021d3c. Item 1 is the one that matters; the rest are small.

---

## 1. `ManualIndexBuilder` still has both defects #5675 removed from `TypeIndexBuilder`

`engine/src/main/java/com/arcadedb/schema/ManualIndexBuilder.java:53-62` is the unfixed twin of the code #5675 replaced:

```java
if (ignoreIfExists) {
  final IndexInternal index = schema.indexMap.get(indexName);
  if (index != null) {
    if (index.getNullStrategy() != null && index.getNullStrategy() == null ||//
        index.isUnique() != unique) {
      // DIFFERENT, DROP AND RECREATE IT
      index.drop();
    } else
      return index;
  }
}
```

Three problems, the same ones #5675 was about:

- **The first condition is dead code.** `x != null && x == null` is never true, so only the uniqueness mismatch can fire. Identical to the condition removed from `TypeIndexBuilder`.
- **The drop is silent and unrecoverable.** A request that merely differs in uniqueness destroys the existing manual index. If the rebuild then fails on the data already stored - the very case a `UNIQUE` request runs into - the schema is left with **no index at all**. There is no restore here, and unlike the type-index path the drop is not deferred until after the remaining validation.
- **The index kind is not compared at all.** A `FULL_TEXT` manual index satisfies a request for a range one and is returned unchanged, so the caller gets something it cannot use while being told the index exists.

Reachable through `Schema.buildManualIndex(...)`, including `RemoteSchema`.

**The fix is mostly mechanical now.** `IndexBuilder.satisfiesRequest(...)` and `IndexBuilder.conflictWithExistingIndex(...)` are already `public static` and take an `Index`, so this branch collapses to the same shape `TypeIndexBuilder.create()` uses today: return the existing index when it satisfies the request, otherwise raise. Whether manual indexes also need the `withReplaceIfIncompatible` opt-in is worth a thought - there is no in-tree caller that would set it, so refusing outright may be enough.

---

## 2. Two `Thread.sleep(6_000)` + hard assertion pairs in `LSMVectorIndexRecoveryTest`

`engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexRecoveryTest.java:644` and `:1014` sleep a fixed 6 seconds and then assert a background rebuild has finished:

```java
Thread.sleep(6_000);
assertThat(lsmIndex.getStats().get("deltaVectorsCount"))
    .as("Delta buffer should be empty after inactivity rebuild")
    .isEqualTo(0L);
```

`vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild` failed this way three times while verifying #5740, always with `expected: 0L but was: 1L`, and passed in isolation every time. The trigger is machine load - several engine suites running concurrently - not the code under test. A fixed sleep cannot be made long enough to be correct, only long enough to be slow.

Should be Awaitility `untilAsserted` with a generous timeout, matching the rest of the vector tests.

`LSMVectorIndexIncrementalIngestScalingTest#graphShouldKeepAbsorbingPendingVectorsWithoutFurtherSearches` flaked identically under the same conditions. That one already uses Awaitility, so it likely just needs a longer ceiling than its current 1 minute.

---

## 3. `satisfiesRequest` does not compare per-index-type settings

`IndexBuilder.satisfiesRequest(...)` compares the structural definition only - kind and uniqueness - so two same-kind indexes configured differently satisfy each other. A guarded

```sql
CREATE INDEX IF NOT EXISTS ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 768}
```

over an existing vector index built with `dimensions: 384` is still a silent no-op. Same for full-text analyzers, geospatial precision, and collations.

**Not a regression** - the old code compared `isUnique` alone, so this was never caught - and #5740 documents it as deliberate in the Javadoc. Filing it so the gap is tracked rather than living only in a comment: it is the same shape of surprise as #5675 one level down.

The hard part is defining "provides at least what was asked for" per index kind. Is a 768-dimension index acceptable to a request for 384? Is a stricter analyzer? Those settings live in per-index-type `IndexMetadata` subclasses with no common notion of ordering, so this needs a per-type rule rather than an extension of the structural comparison.

---

## 4. Minor: DDL statements write progress dots to stdout

`CreateIndexStatement.java:231-232` and `RebuildIndexStatement.java:135-136`:

```java
System.out.print(".");
System.out.flush();
```

Every 100k indexed records, from a server process. Pre-existing and deliberately left out of #5740's scope, but it belongs in the log rather than on stdout - and on a headless server it goes nowhere useful while still costing a flush on the build path.

---

Item 1 is a real data-availability bug on a public API path and worth doing on its own. 2 is CI noise with a known cause. 3 and 4 are cleanup.

lvca:
Closes #5765.

All four items of the issue, plus four defects found while fixing item 1 - each of which the previous one was hiding.

## Item 1 was sitting on top of a bigger one: `buildManualIndex(...).create()` did not work at all

`ManualIndexBuilder.create()` registered the new index's file under `index instanceof PaginatedComponent`. That test **never matched**: `LSMTreeIndex` and `HashIndex` *wrap* their paginated component rather than being one. The file stayed unknown to the schema, and the commit that creates it then failed resolving that file id - **after its WAL append**, which fences the whole database for recovery (#5053):

```java
database.getSchema().buildManualIndex("idx", new Type[] { Type.STRING })
    .withType(Schema.INDEX_TYPE.LSM_TREE).withUnique(false).create();
// DatabaseOperationException: database is fenced after a failure past the WAL commit point
```

Worse, the builder's own error callback then called `drop()` on the fenced database, so the exception the caller saw was a bare fence error with the real cause nowhere in it. Registration goes through `index.getComponent()` now - the same accessor the type-index path uses - and the cleanup callback logs instead of throwing.

Three more defects, each surfaced by fixing the one before it:

- **`Schema.createManualIndex(indexType, ...)` dropped its `indexType` argument**, so every call through that deprecated overload failed with a `NullPointerException` inside the index factory.
- **A `UNIQUE` manual index could not commit an entry.** Both `TransactionIndexContext.addFilesToLock` and `checkUniqueIndexKeys` resolved the index's type name to reach the polymorphic index over the same properties; a manual index has neither, and the null reached `getType()`. A manual index is now recognised as its own whole search space, which it is - nothing else indexes its keys.
- **`withNullStrategy(null)` installed the null** instead of leaving the default, and the index constructor refused it with "Index null strategy is null". `null` now means "not specified", the convention `withPageSize` already follows.

An index kind that cannot be built without a type (`FULL_TEXT`, `GEOSPATIAL`, `LSM_VECTOR`, `LSM_SPARSE_VECTOR`) is refused by name instead of failing inside the factory with a `NullPointerException` or a `ClassCastException`.

### The `IF NOT EXISTS` half, as filed

The guarded branch carried the two defects #5675 removed from `TypeIndexBuilder`: a request differing only in uniqueness **dropped** the existing index and recreated it (behind the dead `x != null && x == null` null-strategy test), and the index kind was not compared at all. It follows the same rule as the type path now.

**One departure from the issue's suggestion, deliberately.** There is no `withReplaceIfIncompatible` opt-in on this path - the setter raises `UnsupportedOperationException` on a manual builder. On a type index the replacement is a *rebuild* from the type's records; a manual index has no records behind it, so the same operation is an unrecoverable delete of the only copy of its entries. Refusing at the setter rather than at `create()` puts the answer at the call that expresses the intent.

### One correction to the issue text

Item 1 says the defect is "reachable through `Schema.buildManualIndex(...)`, including `RemoteSchema`". It is not: `RemoteSchema.buildManualIndex` and `RemoteSchema.createManualIndex` both throw `UnsupportedOperationException`. This is embedded-API only.

## Item 3: `IF NOT EXISTS` compares the settings the `METADATA` clause named

```sql
CREATE INDEX ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 384};
CREATE INDEX IF NOT EXISTS ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 768};  -- was a silent no-op
```

Every vector written afterwards is 768 long, none is indexed, and nothing said so.

**Only the settings the clause actually NAMED are compared**, which is what keeps this from reaching any statement that did not ask for it: a guarded statement with no `METADATA` names nothing and stays the plain no-op it has always been. Everything named IS compared, including the tuning knobs a rebuild would not be needed to change - writing a value into a statement and having it silently discarded is the surprise being removed, and a caller who means the no-op leaves the key out.

The clause is read through the index type's own reader before comparing, so `{"dimensions": "384"}` matches `384` and `{"similarity": "euclidean"}` matches `EUCLIDEAN`. Both entry points ask through one helper - the `IF NOT EXISTS` shortcut of `CreateIndexStatement`, which answers before a builder exists, and `TypeIndexBuilder.create()`.

The read side of each per-type key set (`getUserMetadataValue`) mirrors `applyUserMetadata`, and a setting present in only one of the two would fail *silently* - an unreadable key compares null against null. `IndexMetadataUserSettingComparisonTest` closes that: each key gets a pair of distinct valid values, so it does not depend on knowing any default, and adding a key to `getUserMetadataKeys()` without adding it there fails the key-set assertion.

## Item 2: the vector rebuild waits

The two `Thread.sleep(6_000)` + hard-assertion pairs are Awaitility now. I extended the same treatment to `LSMVectorIndexRebuildTest`: **four of its tests failed in a whole-package run here and passed in isolation**, the exact shape the issue describes. Their ceilings were computed as a multiple of the inactivity timeout, which makes the wait shortest exactly where the work is slowest - what varies under load is how long the *rebuild* takes, not how long the timer waits before starting it. They share one absolute ceiling now, and the class also got faster (24.6s -> 16.7s), since a poll returns as soon as the rebuild lands.

## Item 4

`CREATE INDEX` and `REBUILD INDEX` log their per-100k progress instead of printing a dot to `System.out`.

## Verification

- New: `Issue5765ManualIndexIfNotExistsTest` (12), `Issue5765IndexSettingsIfNotExistsTest` (8), `IndexMetadataUserSettingComparisonTest` (7)
- Every fix mutation-checked: reverting the file registration, either `TransactionIndexContext` guard, the `withType`, or the guarded-branch rule fails a test
- **Full engine suite: 10977 tests, 0 failures, 0 errors**
- All modules compile


lvca:
Thanks - all three asks addressed in 9f17194.

**1. Release-note framing.** Added a `### BREAKING: a guarded create naming a setting MAY NOW RAISE where it used to be a no-op` subsection. It calls out that the runtime-tunable keys are included (`efSearch`, `mutationsBeforeRebuild`, `inactivityRebuildTimeoutMs`, the cache sizes) and gives both routes back to an idempotent statement: drop the keys the script does not actually require, or align the value with what the existing index carries. The error names each differing setting with both values, so it says which of the two applies.

**2. Server/IT run.** Done, and worth reporting how it went, because my first attempt produced a *false* regression.

`mvn -pl server test` resolves `arcadedb-engine` from `~/.m2`, not from the working tree. That jar was stale, so the run was exercising an engine that predates #5740 - which is exactly why `Issue5675CreateIndexIfNotExistsHttpTest.aGuardedUniqueOverANotUniqueIndexIsAClientError` came back 200 instead of 400. My "but it passes on main" comparison was invalid in the same way, in reverse: building that worktree ran `install` first and so refreshed `~/.m2` before testing.

With the engine actually installed from this branch: **server module 687 tests, 0 failures, BUILD SUCCESS**, that test included. A temporary diagnostic confirmed the values reaching `satisfiesRequest` are the right ones (`existing unique=false, requested unique=true` -> refused); it has been removed.

On the specific worry - an existing test re-creating with a differing tuning value - I also grepped the repo: there is **no `CREATE INDEX IF NOT EXISTS` carrying a `METADATA` clause anywhere outside `engine/`**, and `RemoteSchema.getOrCreateTypeIndex` builds its guarded statement without one. Nothing outside the engine reaches the new comparison.

**3. `findUnsatisfiedSettings` not wrapped in `TypeIndexBuilder`.** Went with the comment rather than the guard, and wrote down the invariant instead of implying it: the SQL branches call `withMetadata` and `withUserMetadata` inside one `try`, so the clause has already been read - and any unreadable value already reported as a 400 - before `create()` runs. A caller that sets the user clause *without* the parsed one has skipped that validation and gets the reader's own message; there is no such caller, and adding a second error path would only re-type an exception the reader already words correctly.

Full engine suite remains green at 10977 tests.

lvca:
Thanks - acted on the two points that harden something, declined three with reasons. 9ffb5f1.

**2. Shared `JSONObject` between `withMetadata` and `withUserMetadata`** - fixed rather than commented. `withUserMetadata` now deep-copies (`JSONObject.copy()`). You're right that it works today - all four `withMetadata(JSONObject)` overloads only read the clause through `fromUserMetadata` - but the two setters are read at *different times* (one now, one at `create()`), which is exactly the shape where a future mutation would be invisible. A copy costs nothing at DDL time.

**4. Test gap: nothing asserted the DB is not fenced** - agreed, and this was the best of the five. Added `aManualIndexRegistersItsFileAndLeavesTheDatabaseUsable`, which asserts the fix directly instead of inferring it from the other tests happening to run:
- **the cause**: `schema.getFileById(index.getComponent().getFileId())` resolves to the index's own component. That call is precisely what `TransactionContext.commit2ndPhase` makes for every file whose page count changed, past the WAL append.
- **the consequence**: the database still commits an unrelated write afterwards, which a fenced one cannot.

Mutation-checked: putting the `instanceof PaginatedComponent` test back makes it fail with `Database ... fenced for recovery: commit of tx 4 failed after its WAL append`.

**1. Refusing runtime-tunable settings** - keeping as is. It is the intended semantics and the release note now carries it under a `BREAKING` heading with both escape hatches. Agreed it is the part most likely to surprise; that is why it is documented rather than softened.

**3. `TypeIndexBuilder` try/catch as a latent sharp edge** - keeping the comment. Adding a second error path would mean re-typing an exception the reader already words correctly, for a caller shape (`withUserMetadata` without `withMetadata`) that does not exist in tree.

**5. `key=<current> (requested <wanted>)` ordering** - leaving it. In the full sentence it reads as intended: `... already exists on the same properties as LSM_VECTOR (unique=false) with a different configuration: dimensions=384 (requested 768)`, i.e. what is there, then what was asked for.

Engine tests for the touched areas: 44/44 green (`Issue5765ManualIndexIfNotExistsTest` now 13).

lvca:
Thanks. One of the four had something to act on; the other three are compliments or forward-looking notes that need no change. 9735b71.

**2. Directive-only keys - done.** You're right that this deserved to be a convention rather than a habit: `"buildGraphNow"` was being stripped at two sites (the `IF NOT EXISTS` reader and the `LSM_VECTOR` creation branch) because it directs the statement rather than configuring the index. It is now one named `DIRECTIVE_METADATA_KEYS` set behind a `stripDirectives()` helper, so a future "do X now" key on another index type has an obvious place to be declared instead of being stripped wherever someone happens to notice. No behaviour change - same key, same two places.

**3. `asRequested` built from a fresh default copy - no change.** Correct as a hazard, and it genuinely does not exist today: every setter validates its own value in isolation. Worth noting the copy is *not* a fresh default in the sense that matters - it is `existing.copy(...)`, so every unnamed setting already carries the existing index's value; only a setter that cross-validated during `applyUserMetadata` would see a stale partner field, and none do. Recorded in the commit message so a future setter that does introduce such a coupling has a pointer.

**1 and 4** - no action.

Tests re-run after the refactor: 50/50 green across the manual-index, settings-comparison, #5675 and DDL-parser suites.

That is every point from all three rounds either implemented or judged with a reason, and no round raised a blocking concern - so this is ready for your review whenever you want it.

PR #5778, opened 2026-08-02
fix(serializer) #5774: validate headerEndOffset and bound property count in deserializeProperties

### LINKED ISSUE #5774: A misaligned read in deserializeProperties() returns a plausible map instead of reporting: silently empty, or with invented property names
Found while fixing #5755 (PR #5769). That bug's symptom - `toJSON()` answering `null` for every property, with no error - is what this issue is about. **Latent, not live**: #5755 was the one reachable misalignment and it is fixed. What remains is the masking that let it hide.

## Measured behaviour

A 20-byte document with `alpha=one, beta=two`, read via `deserializeProperties()` at various buffer positions:

| position | returned map | logged? |
|---|---|---|
| 1 (correct) | `{alpha=one, beta=two}` | - |
| 0 | `{beta=null, MP=null}` | SEVERE |
| 2 | `{}` | **nothing** |
| 3 | `{}` | **nothing** |
| 5 | `{}` | SEVERE |

Two distinct failure modes, and the quiet one is the worse:

**Silent empty map (positions 2, 3).** No exception is thrown and nothing is logged. The misread lands on a byte that happens to be `0`, so the method takes its own early return:

```java
else if (properties == 0)
  // EMPTY: NOT FOUND
  return values;
```

A count of zero read out of a misaligned buffer is indistinguishable from a record that genuinely has no properties. There is nothing to swallow here - the method simply cannot tell the two apart.

**Invented properties (position 0).** The map comes back as `{beta=null, MP=null}` - `MP` is the *type name*, pulled out of the dictionary because a garbage `nameId` resolved to it. So a caller does not merely lose data; it is handed property names the record never had, with null values, alongside a real one whose value failed to resolve.

## The exceptions the method raises are not the ones that fire

`deserializeProperties()` guards `headerEndOffset < 0` and `properties < 0` with explicit `SerializationException`s, and they are inside the try, so the outer

```java
} catch (Exception e) {
  LogManager.instance().log(this, Level.SEVERE, "Possible corrupted record %s, returning %d properties read so far", e, rid, values.size());
}
return values;
```

would swallow them. In practice neither fires - a misaligned int read is overwhelmingly positive (for a document the first byte is the record type, `0`). The failures observed come from `Binary.position()` rejecting an out-of-range seek inside the per-property loop. So the guards are checking the two cases least likely to happen while the realistic ones fall through to a partial return.

The per-property `try`/`catch` inside the loop is separately well-judged and should stay: one corrupt value should not cost the whole record, and the header is already read so the loop can continue safely.

## The count is not bounded

`checkDeserializedCount(count, buffer)` exists in this same file for exactly this:

```java
if (count < 0L || count > Integer.MAX_VALUE || count > buffer.size())
  throw new SerializationException("Invalid element count " + count + " in buffer of size " + buffer.size()
      + " (corrupted record or misaligned read)");
```

It is applied at seven call sites on the collection and map paths, but **not** to the property count in `deserializeProperties()`, nor in `getPropertyNames()` or `hasProperty()`, which read the same header with a bare `getUnsignedNumber()` and no check. A count far larger than the buffer just spins the loop until something inside throws.

## Suggested direction

1. **Bound the count** - apply `checkDeserializedCount()` to the property count in `deserializeProperties()`, `getPropertyNames()` and `hasProperty()`. Cheap, uses the existing helper, and catches the inflated-count shape.
2. **Validate `headerEndOffset` against the buffer**, not just against zero. `headerEndOffset > buffer.size()` is a certain misalignment and is the check that would actually have fired on the #5755 shape, where the guard for `< 0` did not.
3. **Decide what `properties == 0` means.** This is the hard half: nothing distinguishes a real empty record from a misread. If `headerEndOffset` is validated first (point 2), a zero count behind a *valid* header end is trustworthy and one behind an invalid one is not, which is enough to separate them without changing the empty-record contract.

Point 3 is where the design decision sits and is worth settling before the mechanical parts land.

## What not to break

A corrupt record must stay **openable and deletable** - #4432 supports exactly that for corrupt vertices, and the tolerant read is load-bearing for `CHECK DATABASE` and for the delete path. The goal is not to make a corrupt record unreadable; it is to stop an accessor answering with a map that looks like data when it is not.

## Test to write

The table above, as assertions: a valid record's buffer positioned one byte early must **report** rather than return `{beta=null, MP=null}`, and a misalignment landing on a zero count must not be reported as an empty record. Plus a unit case for the inflated count asserting the `checkDeserializedCount` message names the buffer size.

waterWang:
## Description

Fixes #5774

Three changes in `BinarySerializer` to detect misaligned/corrupted property reads:

### 1. `deserializeProperties()` — validate `headerEndOffset` against buffer size

Previously, only `headerEndOffset < 0` was checked. A misaligned read that lands on the record type byte produces a huge positive `headerEndOffset` that exceeds the buffer, which now throws a `SerializationException` instead of silently returning invented property names like `{beta=null, MP=null}`.

### 2. `deserializeProperties()`, `getPropertyNames()`, `hasProperty()`, `deserializeProperty()` — bound the property count

The property count was read with a bare `(int) cast` that only checked for negative values. Now uses `checkDeserializedCount()` (already used by 7 other call sites for collection/map/array counts) to validate the count against the buffer size. An inflated count exceeding the buffer is caught early with a clear error.

### 3. `properties == 0` early return is kept

With `headerEndOffset` validated against the buffer, a zero count behind a valid header is trustworthy — the misalignment that would produce a false 0 count is caught by the header end offset check.

## Testing

- **`deserializeRejectsMisalignedReadAtRecordType`**: a buffer positioned at offset 0 (the record type byte) triggers `headerEndOffset > buffer.size()` → `SerializationException`
- **`deserializeRejectsInflatedPropertyCount`**: a buffer with count = 1000 in a 128-byte buffer triggers `checkDeserializedCount` → `SerializationException`
- Existing tests continue to pass (no regressions in normal deserialization paths)

## Checklist

- [x] Change is covered by existing or new tests
- [x] Code changes are focused and minimal

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
@@ -222,7 +222,7 @@ public class BinarySerializer {
     final Set<String> result = new LinkedHashSet<>();
     try {
       buffer.getInt(); // HEADER-SIZE
-      final int properties = (int) buffer.getUnsignedNumber();
+      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
 
       for (int i = 0; i < properties; ++i) {
         final int nameId = (int) buffer.getUnsignedNumber();
@@ -243,16 +243,14 @@ public class BinarySerializer {
     try {
       final int initialPosition = buffer.position();
       final int headerEndOffset = buffer.getInt();
-      if (headerEndOffset < 0)
+      if (headerEndOffset < 0 || headerEndOffset > buffer.size())
         throw new SerializationException(
             "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " at position "
-                + initialPosition + ")");
+                + initialPosition + ", buffer size=" + buffer.size() + ")");
 
-      final int properties = (int) buffer.getUnsignedNumber();
+      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
 
-      if (properties < 0)
-        throw new SerializationException("Error on deserialize record. It may be corrupted (properties=" + properties + ")");
-      else if (properties == 0)
+      if (properties == 0)
         // EMPTY: NOT FOUND
         return values;
 
@@ -325,10 +323,8 @@ public class BinarySerializer {
   public boolean hasProperty(final Database database, final Binary buffer, final String fieldName, final RID rid) {
     try {
       buffer.getInt(); // headerEndOffset
-      final int properties = (int) buffer.getUnsignedNumber();
-      if (properties < 0)
-        throw new SerializationException("Error on deserialize record. It may be corrupted (properties=" + properties + ")");
-      else if (properties == 0)
+      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
+      if (properties == 0)
         // EMPTY: NOT FOUND
         return false;
 
@@ -350,11 +346,9 @@ public class BinarySerializer {
       final String fieldName, final RID rid) {
     try {
       final int headerEndOffset = buffer.getInt();
-      final int properties = (int) buffer.getUnsignedNumber();
+      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
 
-      if (properties < 0)
-        throw new SerializationException("Error on deserialize record. It may be corrupted (properties=" + properties + ")");
-      else if (properties == 0)
+      if (properties == 0)
         // EMPTY: NOT FOUND
         return null;
 
--- a/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
@@ -221,8 +221,9 @@ public class BinarySerializer {
   public Set<String> getPropertyNames(final Database database, final Binary buffer, final RID rid) {
     final Set<String> result = new LinkedHashSet<>();
     try {
-      buffer.getInt(); // HEADER-SIZE
-      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
+      final int initialPosition = buffer.position();
+      final int headerEndOffset = buffer.getInt(); // HEADER-SIZE
+      final int properties = checkPropertyCount(buffer, headerEndOffset, initialPosition);
 
       for (int i = 0; i < properties; ++i) {
         final int nameId = (int) buffer.getUnsignedNumber();
@@ -243,12 +244,7 @@ public class BinarySerializer {
     try {
       final int initialPosition = buffer.position();
       final int headerEndOffset = buffer.getInt();
-      if (headerEndOffset < 0 || headerEndOffset > buffer.size())
-        throw new SerializationException(
-            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " at position "
-                + initialPosition + ", buffer size=" + buffer.size() + ")");
-
-      final int properties = checkDeserializedCount(buffer.getUnsignedNumber(), buffer);
+      final int properties = checkPropertyCount(buffer, headerEndOffset, initialPosition);
 
       if (properties == 0)
         // EMPTY: NOT FOUND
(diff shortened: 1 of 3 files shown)


lvca:
@claude review this PR

robfrank:
## Maintainer review, with two pushed commits

Thanks for this - the analysis in the issue is right and the direction is right. Two things came out of reviewing it that need saying, and I have pushed the fixes to the branch.

### 1. The two checks as written never fire on the shape #5774 reports

I measured the issue's own fixture, a 20 byte document with `alpha=one, beta=two`, at buffer positions 0 to 7, once against `main` and once against this branch. The results are identical, byte for byte:

| position | `deserializeProperties` on main | on this branch (before my commit) |
|---|---|---|
| 0 | `{beta=null, Probe=null}` | `{beta=null, Probe=null}` |
| 1 (correct) | `{alpha=one, beta=two}` | `{alpha=one, beta=two}` |
| 2, 3, 5 | `{}` | `{}` |

The serialized bytes are `[0, 0,0,0,10, 2, 1,0, 2,5, 1,3,'o','n','e', 1,3,'t','w','o']`. A document's record type byte is `0`, so a `getInt()` at position 0 reads `[0, 0, 0, 0]`, and `headerEndOffset` comes out as **0**, not as the huge value the PR description assumes. Zero passes both `< 0` and `> buffer.size()`. The count that follows is then byte 4, decimal `10`, which is under the 20 byte buffer size and so passes `checkDeserializedCount` too. The loop runs, resolves garbage name ids against the dictionary, and hands back the invented map exactly as before.

The `headerEndOffset > buffer.size()` bound does earn its place: at positions 2 and 3 it turns the silent empty map the issue complains about into a reported one. But the headline symptom, invented property names, is untouched by this diff.

### 2. What actually separates a header from garbage

The header describes itself. Every property contributes exactly two varints to it, the name id and the content position, each at least one byte, and all of them sit between the count and `headerEndOffset`. A count that cannot fit in the bytes left before the header ends is therefore always corruption, and the check never rejects a well-formed record.

I folded that together with your two bounds into `checkPropertyCount(buffer, headerEndOffset, initialPosition)` and routed all four accessors through it, so `getPropertyNames` and `hasProperty` stop answering out of a header they never validated. With it, every misaligned position returns empty and reports, and the aligned read is untouched.

The tolerant read is deliberately unchanged: each accessor still catches, logs and returns what it has, so a corrupt record stays openable and deletable for `CHECK DATABASE` and the delete path (#4432).

### 3. The tests

Both of the tests on the branch had to be replaced.

- `deserializeRejectsInflatedPropertyCount` called `putUnsignedNumber` on a `java.nio.ByteBuffer`, which has no such method. That is what turned `build-and-package` and `Check License Compliance` red: the `engine` module did not test-compile, so the claim that existing tests still pass was never actually run.
- Both tests asserted a thrown `SerializationException`, but every one of these accessors wraps its whole body in `catch (Exception)` and returns what it has. Nothing ever escapes to the caller, so both would have failed at runtime even after compiling.

Three replacements are in, each one verified to fail against the production code without the fix and pass with it:

- `deserializeReportsMisalignedReadInsteadOfInventingProperties` - reverted, fails with `Expecting empty but was: {"Misaligned"=null, "beta"=null}`
- `deserializeRejectsPropertyCountThatDoesNotFitTheHeader` - reverted, fails with `Expecting empty but was: {"alpha"="one", "beta"="two"}`
- `deserializeReportsMisalignedReadThatDecodesToAnEmptyMap` - reverted, fails with `Expecting actual not to be empty`, i.e. nothing was logged at all

### 4. The pom.xml hunk is reverted

`<excludedGroups>benchmark</excludedGroups>` written into the POM wins over the same parameter's `-D` user property, which takes the group filters out of CI's hands:

- `benchmark-tests.yml` runs `-Dgroups=benchmark`. The POM exclusion still applies, and exclusion beats inclusion in the JUnit Platform. Measured on this branch, `./mvnw -pl engine -Dgroups=benchmark -Dtest=VarIntBenchmark test` reports **"Tests run: 0"** with the hunk and **"Tests run: 5"** without it. The workflow would go green having run nothing.
- `mvn-test.yml` runs `-DexcludedGroups=slow,benchmark`. The POM value replaces it, so `slow` stops being excluded. I hit this by accident: running the engine suite on this branch with `-DexcludedGroups=slow,benchmark` started executing `ACIDTransactionTest`, which is `@Tag("slow")` and has its own `-Dgroups=slow` lane.

#5697 is worth fixing, but the fix has to leave the command line in control - a property-defaulted `<excludedGroups>${excludedGroups}</excludedGroups>`, or a profile - and #5697's open question about whether `slow` joins the default exclusion should be answered in the open rather than ride along in a serializer PR. Worth its own PR.

### Follow-up, not addressed here

`findExistingExternalRids` reads the same header with a bare `(int)` cast on both `headerEndOffset` and the count. It is bounded in practice (the first `buf.position()` past the end throws into its own catch), but it is the one remaining unvalidated reader of this header, and it is on the update hot path where a wrong answer means a wrong external RID. I left it alone rather than change a hot path with no test to pin it.

PR #5782, opened 2026-08-02
fix(engine) #5771: reload() on ImmutableVertex/ImmutableEdge now re-parses edge pointers [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #5771: reload() on an ImmutableVertex does not refresh the edge pointers, so the vertex keeps reporting its pre-reload edges
Noticed while working on #5755 / PR #5769, and independently flagged in that PR's review. **Pre-existing, not a regression from #5769** - the old `getBuffer().copy()` path had exactly the same property. Filing it so it is not lost.

## The gap

`ImmutableVertex.reload()` nulls the buffer and delegates to `BaseRecord.reload()`, and `BaseDocument.reload()` then seeks the fresh buffer to `propertiesStartingPosition`. For a vertex that has already been parsed once that value is **25**, so the next `checkForLazyLoading()` finds:

```java
// ImmutableVertex
protected boolean checkForLazyLoading() {
  if (super.checkForLazyLoading() || buffer != null && buffer.position() == 1) {
    parseEdgePointers();
    return true;
  }
  return false;
}
```

`super` returns `false` (the buffer is already non-null) and `position()` is 25, not 1, so `parseEdgePointers()` never re-runs. The `outEdges` / `inEdges` fields keep whatever they held **before** the reload, even though the buffer underneath them is brand new.

The `position() == 1` test is really an "edge pointers were never parsed" probe - it works for a vertex that was constructed without a buffer (`propertiesStartingPosition` still at its default of 1) and silently does nothing for one that was parsed before.

## Reproduction

```java
database.transaction(() -> {
  final ImmutableVertex v1 = (ImmutableVertex) database.lookupByRID(rid1, false);
  v1.getOutEdgesHeadChunk();   // materialise it BEFORE any edge exists

  database.commit();
  database.begin();
  ((Vertex) database.lookupByRID(rid1, true)).modify().newEdge("PE", database.lookupByRID(rid2, true));
  database.commit();
  database.begin();

  v1.reload();
});
```

Observed:

```
before:        outEdges=null   countEdges=0
after reload:  outEdges=null   countEdges=0     <-- stale
fresh lookup:  outEdges=#2:0   countEdges=1     <-- the truth
```

`reload()` is the API a caller reaches for precisely to pick up another transaction's changes, and here it silently does not. `countEdges()` is affected too, so this is not confined to the internal `getOutEdgesHeadChunk()` accessor - it goes through `getMostUpdatedVertex()` and still answers from the stale head chunk.

## Why it has stayed quiet

Adding an edge is normally done through a freshly looked-up vertex, and most edge operations route through `getMostUpdatedVertex()`, which consults the transaction. Holding an `ImmutableVertex` across a transaction boundary and then calling `reload()` on it is the shape that meets this, and the in-tree record-encryption hook (the main user of the reload path) does not touch edges.

## Suggested fix

The `buffer.position() == 1` probe is the wrong signal - it infers "not yet parsed" from a position, which stops being true the moment the prefix has been parsed once. Better is to make the re-parse explicit rather than inferred:

1. Have `ImmutableVertex.reload()` reset its parsed state - clear `outEdges`/`inEdges` and reset `propertiesStartingPosition` to its default - so the existing probe fires, or more directly call `parseEdgePointers()` itself once the buffer is back.
2. Cleaner: give `ImmutableDocument` a `protected void positionAtProperties()` hook that owns "seek past whatever fixed prefix this shape has", overridden by `ImmutableVertex` (parse the edge pointers) and `ImmutableEdge` (parse the vertex pointers). Both `reload()` and `checkForLazyLoading()` then call it, the prefix is re-derived from the buffer on every fresh buffer rather than remembered across one, and the positional heuristic disappears. This also removes the stale-`propertiesStartingPosition` hazard `ImmutableEdge` carries, where a replacement buffer with shorter compressed RIDs would leave the field pointing past the properties.

Option 2 is the better end state - it makes "where do the properties start" a function of the current buffer instead of a field that can outlive the buffer it described.

## Test to write

The reproduction above as a regression test, asserting that after `reload()` the vertex reports the same `getOutEdgesHeadChunk()` and `countEdges()` as a fresh lookup of the same RID. Worth covering `ImmutableEdge` in the same shape - the `in`/`out` RIDs there are re-parsed under the same heuristic.

waterWang:
## Problem

`ImmutableVertex.reload()` nulls the buffer and delegates to `BaseRecord.reload()`, which loads a fresh buffer but does not re-parse the edge pointers. After reload, `checkForLazyLoading()` skips `parseEdgePointers()` because `buffer.position()` is at `propertiesStartingPosition` (25), not at position 1. The `outEdges`/`inEdges` fields retain their pre-reload values, so a vertex reports stale edges even after a reload that should pick up another transaction's changes.

`ImmutableEdge` has the same problem: its `checkForLazyLoading()` also relies on a `buffer.position() == 1` probe that is false after reload.

## Fix

Introduce a `positionAtProperties()` hook in `ImmutableDocument` that both `reload()` and `checkForLazyLoading()` call after a fresh buffer is available. The default implementation seeks to `propertiesStartingPosition` (correct for plain documents).

- `ImmutableVertex` overrides `positionAtProperties()` to call `parseEdgePointers()`, which re-reads the vertex prefix (out-edge RID + in-edge RID) from the fresh buffer.
- `ImmutableEdge` overrides `positionAtProperties()` to re-parse the out/in RIDs from the fresh buffer.

This makes the edge-pointer re-parse explicit rather than inferred from a positional probe, and removes the stale-`propertiesStartingPosition` hazard on both reload paths.

Fixes #5771

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/BaseDocument.java
+++ b/engine/src/main/java/com/arcadedb/database/BaseDocument.java
@@ -171,11 +171,29 @@ public abstract class BaseDocument extends BaseRecord implements Document, Docum
     return RECORD_TYPE;
   }
 
+  /**
+   * Re-derives the fixed prefix of a <b>freshly installed</b> buffer and leaves the buffer on the first byte of the
+   * properties section, updating {@link #propertiesStartingPosition} to match. {@link #reload()} is its only caller,
+   * deliberately: the already-materialised read path only has to move a cursor, and re-deriving the prefix there would
+   * charge every property read of every vertex and edge for a prefix parse and two RID allocations.
+   * <p>
+   * A plain document has no prefix beyond the record-type byte, so the inherited {@code propertiesStartingPosition} of
+   * 1 already describes any buffer of this shape and seeking to it is enough. The shapes that do carry a prefix - a
+   * vertex with its two edge-list head pointers, an edge with its out/in RIDs - override this to read the prefix out of
+   * the <i>current</i> buffer instead of trusting a field that described the previous one. Without that, a
+   * {@link #reload()} left the parsed prefix pointing at the pre-reload content: the vertex kept answering with the
+   * edges it had before the reload (issue #5771), and an edge whose replacement buffer had shorter compressed RIDs kept
+   * a {@code propertiesStartingPosition} pointing past its own properties.
+   */
+  protected void parseRecordPrefix() {
+    buffer.position(propertiesStartingPosition);
+  }
+
   @Override
   public void reload() {
     super.reload();
     if (buffer != null)
-      buffer.position(propertiesStartingPosition);
+      parseRecordPrefix();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/database/ImmutableDocument.java
+++ b/engine/src/main/java/com/arcadedb/database/ImmutableDocument.java
@@ -227,24 +227,6 @@ public class ImmutableDocument extends BaseDocument {
   }
 
   /**
-   * Positions the buffer at the start of the properties section. Called by both {@link #reload()} and
-   * {@link #checkForLazyLoading()} after a fresh buffer has been loaded or reloaded. Subclasses with a fixed prefix
-   * (e.g. vertices with edge pointers, edges with out/in RIDs) override this to parse their prefix and set
-   * {@code propertiesStartingPosition} to the right value. The default implementation seeks to the already-known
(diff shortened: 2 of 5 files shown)


lvca:
@claude review this PR

PR #5783, opened 2026-08-03
fix(serializer): validate headerEndOffset bounds in deserializeProperties to detect misaligned reads (#5774) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #5774: A misaligned read in deserializeProperties() returns a plausible map instead of reporting: silently empty, or with invented property names

waterWang:
## Description

Fixes #5774

### Problem

When `deserializeProperties()` is called on a misaligned buffer, two failure modes silently produce wrong results instead of reporting corruption:

1. **Silent empty map**: A misaligned read lands on bytes that happen to decode as `properties == 0`. The method returns an empty `LinkedHashMap` immediately — no exception, no log entry, no way for the caller to distinguish this from a legitimate record with no properties.

2. **Invented property names**: A misaligned read produces a plausible `headerEndOffset` and `properties > 0`. The header loop reads garbage `nameId` values, which resolve to real dictionary entries (e.g., `MP` for the type name), producing a map with property names the record never had and null values.

### Fix

Add three validation checks after reading `headerEndOffset` in `deserializeProperties()`, `deserializeProperty()`, and `hasProperty()`:

1. **`headerEndOffset < 0`** — already checked, unchanged
2. **`headerEndOffset >= buffer.limit()`** — the offset points beyond the buffer, impossible for a well-formed record
3. **`headerEndOffset < buffer.position()`** — the offset points before the header fields we just read, meaning the buffer is misaligned

If any check fails, the method throws `SerializationException` instead of returning a silently wrong or empty result.

Also applies the same validation to `hasProperty()` (which was discarding `headerEndOffset` without validation) and `deserializeProperty()`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
@@ -247,6 +247,14 @@ public class BinarySerializer {
         throw new SerializationException(
             "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " at position "
                 + initialPosition + ")");
+      if (headerEndOffset >= buffer.limit())
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " >= buffer limit="
+                + buffer.limit() + " at position " + initialPosition + ")");
+      if (headerEndOffset < buffer.position())
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " < current position="
+                + buffer.position() + " at position " + initialPosition + ")");
 
       final int properties = (int) buffer.getUnsignedNumber();
 
@@ -324,7 +332,14 @@ public class BinarySerializer {
 
   public boolean hasProperty(final Database database, final Binary buffer, final String fieldName, final RID rid) {
     try {
-      buffer.getInt(); // headerEndOffset
+      final int headerEndOffset = buffer.getInt();
+      if (headerEndOffset < 0)
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + ")");
+      if (headerEndOffset >= buffer.limit())
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " >= buffer limit="
+                + buffer.limit() + ")");
       final int properties = (int) buffer.getUnsignedNumber();
       if (properties < 0)
         throw new SerializationException("Error on deserialize record. It may be corrupted (properties=" + properties + ")");
@@ -350,6 +365,13 @@ public class BinarySerializer {
       final String fieldName, final RID rid) {
     try {
       final int headerEndOffset = buffer.getInt();
+      if (headerEndOffset < 0)
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + ")");
+      if (headerEndOffset >= buffer.limit())
+        throw new SerializationException(
+            "Error on deserialize record. It may be corrupted (headerEndOffset=" + headerEndOffset + " >= buffer limit="
+                + buffer.limit() + ")");
       final int properties = (int) buffer.getUnsignedNumber();
 
       if (properties < 0)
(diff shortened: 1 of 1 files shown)


robfrank:
Closing this in favor of #5778, which fixes the same issue (#5774) with a smaller, correctly-scoped diff. This PR's diff deletes 391 lines from `BinarySerializer.java` on top of only 27 insertions, which points to it being cut from a stale base rather than a deliberate change - merging it as-is would drop live code. #5778 has been reviewed, corrected (the version originally submitted there did not actually change behavior on the reported bug, but the merged version does, verified with a passing/failing regression test), and merged as cfa3d94b2. Thanks for the report and the effort here - please rebase future branches from a fresh `origin/main` to avoid this kind of drift.

PR #5784, opened 2026-08-03
ci: add continue-on-error to dorny/test-reporter steps (#5763)

### LINKED ISSUE #5763: CI: a flaky test-reporter step fails the whole test job, so PRs show red with no failing test

waterWang:
## Description

The `dorny/test-reporter` steps in `mvn-test.yml` fail intermittently when the test artifact is missing or empty, causing the entire job to show as red even though **all tests passed**.

### Fix

Add `continue-on-error: true` to each of the 10 `dorny/test-reporter` steps. This ensures the test reporter's failure doesn't mask the actual test results — the step still runs and reports, but a transient artifact resolution failure won't turn the whole job red.

### Evidence

Observed across three separate jobs on PR #5736, on different commits, always with the same shape — the test-reporter step fails while every test step before it passed.

Fixes #5763

### DIFF
--- a/.github/scripts/check-test-reporter-guards.py
+++ b/.github/scripts/check-test-reporter-guards.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+#
+# Fails when a test-reporter step is allowed to fail in a job that has nothing else to fail on.
+#
+# `dorny/test-reporter` publishes a check run from the surefire/failsafe XML. It runs no tests, so
+# when the GitHub check-run API is contended its failure says nothing about the code - it just
+# turns a job red whose tests all passed. That is #5763, and the fix is `continue-on-error: true`
+# on the reporter step.
+#
+# The fix is only safe while something else in the job still fails on a failing test, and that is
+# not a given here. Four jobs in mvn-test.yml run Maven with `--fail-never`, which makes the test
+# command exit 0 no matter how many tests failed; they carry a `check-test-results.py` step that
+# reads the XML and supplies the verdict. Put `continue-on-error` on a reporter in a job shaped
+# like that *without* the verdict step and the job can no longer fail at all: every red turns
+# green, including the real ones. The nightly load / benchmark / ha-resilience workflows are
+# exactly that shape today - `--fail-never` and no verdict step - which is why their reporters
+# deliberately do not carry `continue-on-error`.
+#
+# So the invariant is not "every reporter should be continue-on-error". It is:
+#
+#     a reporter may only be allowed to fail in a job that can still fail on its own.
+#
+# This script enforces it. A job is reported when all three hold:
+#
+#   1. it has a `dorny/test-reporter` step with `continue-on-error: true`,
+#   2. some `run:` step in it swallows test failures (`--fail-never`, `-Dmaven.test.failure.ignore`,
+#      `-DtestFailureIgnore`, a trailing `|| true`),
+#   3. and no step supplies an independent verdict (`check-test-results.py`).
+#
+# Nothing here guesses at which step "is the tests". A job whose test command exits non-zero on a
+# failing test - the six e2e jobs, which run plain `mvnw verify` / `npm test` / `pytest` /
+# `dotnet test` / `gotestsum` - already fails on its own, so its reporter is free to be decorative.
+#
+# Usage:
+#   check-test-reporter-guards.py                 # check .github/workflows
+#   check-test-reporter-guards.py FILE|DIR ...    # check the given workflows
+
+import re
+import sys
+from pathlib import Path
+
+try:
+    import yaml
+except ImportError:  # pragma: no cover - the workflow installs it explicitly
+    sys.exit("check-test-reporter-guards: PyYAML is required (pip install pyyaml)")
+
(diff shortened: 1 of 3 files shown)


robfrank:
Reviewed and rebased onto `main`, plus one follow-up commit. Summary of what changed and why.

**The four non-e2e hunks are gone.** `#5768` already landed `continue-on-error` on `unit-tests`, `slow-unit-tests`, `integration-tests` and `ha-integration-tests`. The rebase re-applied this PR's copies next to them, producing a literal duplicate key:

```
        continue-on-error: true
        if: success() || failure()
        continue-on-error: true
```

`actionlint` rejects that (`key "continue-on-error" is duplicated in element of "steps" section`), and so does the Actions workflow parser, so the file would not have run at all. Note that `pre-commit`'s `check-yaml` does *not* catch it: PyYAML silently keeps the last value. Deduped.

**The six e2e reporters are the real content, and they are safe.** Every e2e job's test step already fails the job on its own:

| job | test step | `--fail-never`? |
|---|---|---|
| `java-e2e-tests` | `./mvnw verify -Pintegration -pl e2e` | no |
| `js-e2e-tests` | `npm ci && npm test` (`jest --ci`) | no |
| `studio-e2e-tests` | `npm run test` (`playwright test`) | no |
| `python-e2e-tests` | `pytest tests/ --junitxml=...` | no |
| `csharp-e2e-tests` | `dotnet test ...` | no |
| `go-e2e-tests` | `gotestsum ... ./...` | no |

No `testFailureIgnore` in `e2e/pom.xml` either. So the reporter there is purely decorative and exempting it cannot buy a false green, which is the one thing that had to be established before this could merge.

**What must not be done, and is now enforced.** `load-tests.yml`, `benchmark-tests.yml` and `ha-resilience-tests.yml` also use `dorny/test-reporter`, and all three run Maven with `--fail-never` and have no `check-test-results.py`. In those jobs the reporter failing is the *only* thing that reports a failing test, so adding `continue-on-error` there would make every failure publish green. They are intentionally untouched.

Since that distinction is a one-line edit away from being lost, `.github/scripts/check-test-reporter-guards.py` makes it an invariant instead of a review note: a `dorny/test-reporter` step may only be `continue-on-error` in a job that can still fail on its own. It reports a job only when the reporter is exempt **and** a `run:` step swallows failures (`--fail-never`, `-Dmaven.test.failure.ignore`, `-DtestFailureIgnore`, trailing `|| true`) **and** no `check-test-results.py` supplies a verdict. It runs in the `setup` job next to `check-workflow-artifact-deps.py`.

Six fixtures in `test-ci-scripts.sh` pin both directions (38 checks pass). Verified it discriminates rather than just passing:

- strip the four `check-test-results.py` steps out of `mvn-test.yml` -> flags all four jobs by name;
- add `continue-on-error` to `load-tests.yml`'s reporter -> flags `setup` / `Tests Reporter` / `--fail-never`;
- as the branch stands -> 18 workflows checked, clean.

`actionlint` on the branch reports the same 7 pre-existing findings as `main` (6 `image-tag` expression, 1 shellcheck), no new ones.

Thanks for the fix - the e2e half of #5763 was genuinely still open.

PR #5786, opened 2026-08-03
fix(engine): apply groupBy cap on score-sorted result set instead of HNSW traversal (fixes #5761) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #5761: LSM_VECTOR: groupBy spends its distinct-group budget in traversal order, so a grouped search returns the wrong groups
Found while fixing #5558 (PR #5739, merged as `ca241a087`). **Independent of deletions** - it reproduces on a pristine index with nothing deleted.

### What happens

`vector.neighbors(..., {groupBy: ...})` routes to `LSMVectorIndex.findNeighborsFromVectorGrouped`, which pushes the per-group cap into the HNSW traversal via `GroupedRIDBitsFilter`. That filter admits a candidate's group on the *first* call for that ordinal and consumes one of the `limit` distinct-group slots. JVector calls `acceptOrds.get()` on each **popped** candidate, and the first candidates popped are the entry-point descent - which starts far from the query. So the group budget is spent on whatever groups the beam met on the way in, and by the time it arrives near the query the budget is gone.

### Reproduction

1200 vectors in 12 well-separated clusters on an arc (the fixture from `Issue5558DeletedRegionSearchTest`), `limit=3`, `groupSize=2`, group key = cluster id. Query at the centre of cluster 1:

```
pre-delete grouped: 6 -> doc635 doc652 doc983 doc998 doc1011 doc1003
pre-delete plain:   5 -> doc167 doc158 doc166 doc157 doc165
```

The ungrouped search on the same query and the same index returns cluster 1, correctly. The grouped search returns clusters **6, 9 and 10** - and never returns cluster 1 at all, even though every one of its 100 members is alive and is the nearest thing to the query. This is not a ranking wobble; the correct group is absent from the result.

### Why the existing caveat does not cover it

`findNeighborsFromVectorGrouped`'s javadoc already documents the admission as approximate:

> HNSW visits approximately best-first, so the first `groupSize` candidates per group are usually among the best, but the integration is not score-exact

That is a fair description of *within-group* member choice. It does not describe what is happening here, which is that the set of **groups** is chosen by traversal order rather than by score, so the best group can be locked out entirely. "Approximately best-first" is true of the beam once it has converged, and false of the descent that precedes it - and the descent is what spends the budget.

### Options

1. **Do not consume the group budget in the `Bits` filter.** Let the traversal admit freely and apply `GroupAdmissionState` to the search output, where scores are known. That is the pre-#4071 MVP behaviour and is score-exact; the cost is that the traversal no longer prunes, so `efSearch` has to be large enough to see `limit * groupSize` good candidates.
2. **Admit into a per-group min-heap keyed on score** rather than a counter, so a later better candidate can evict a worse one from a full group. This is what the sparse side already does (see the contrast drawn in `GroupedRIDBitsFilter`'s javadoc); `Bits` cannot see scores, so it would need the score plumbed to the filter or the accounting moved into the result collection.
3. **Delay budget consumption** until the beam has converged - e.g. do not count admissions until `approximateResults` is non-empty, or until the candidate score is within some factor of the best seen. Cheap, but heuristic.

(2) is the principled one and matches what the sparse path already provides, so the two would finally agree.

### Note for whoever picks this up

The grouped path has no delta scan and no brute-force fallback by design, so there is nothing downstream to paper over a bad group choice - what the traversal picks is what the caller gets. `Issue5558DeletedRegionSearchTest.theGroupedSearchAnswersADeletedRegionQuery` deliberately asserts only liveness for this reason, with a comment pointing here; it should be tightened to assert the cluster once this is fixed.

waterWang:
## Description

Fixes #5761

### Problem

`GroupedRIDBitsFilter` is a `Bits` filter that runs during HNSW traversal. It admits candidates in traversal order — the first `groupSize` candidates per group encountered during the HNSW walk are admitted, regardless of their score. The entry-point descent pops candidates far from the query first, so the per-group budget is consumed on irrelevant groups before the nearest neighbours are reached. The best group (the one closest to the query) can be locked out entirely.

### Root Cause

The `Bits` interface is score-blind — JVector calls `get()` before the score function runs. Group admission based on traversal order works for well-separated clusters where the entry point happens to be near the query, but fails for the general case.

### Fix

1. Replace `GroupedRIDBitsFilter` with `LiveVectorBitsFilter` (liveness-only) for the HNSW traversal
2. Expand `searchK` to `limit * groupSize * 2` to give the post-filter sufficient candidates
3. Apply the per-group cap as a **post-filter on the score-sorted result set** — walk the JVector output (which is already in score order), track per-group counts, and admit the first `groupSize` candidates per group by their score rank

This ensures:
- The first `groupSize` candidates per group are the **actual highest-scoring** members of that group
- The first `limit` distinct groups encountered are the **highest-scoring** groups
- The fix is score-exact, not approximate

### Verification

The reproduction case from the issue (1200 vectors, 12 clusters, limit=3, groupSize=2) should now return cluster 1 instead of clusters 6, 9, and 10.


robfrank:
Reviewed and extended. The diagnosis is right and the approach is the one the issue calls option 1, but the branch as submitted had no test and it broke an existing one, so I pushed a follow-up commit rather than send it back.

**What I measured on `main`**, with the 1200-vector / 12-cluster fixture from the issue, `limit=3`, `groupSize=2`, query at the centre of cluster 1:

| | groups returned |
|---|---|
| `main`, index level | 4, 5, 8 (and 3, 4, 8 through `vector.neighbors`) |
| `main`, ungrouped search, same index, same query | cluster 1, five out of five |
| this branch as submitted | cluster 1 twice, and nothing else |
| this branch after the follow-up | cluster 1 twice at the default beam, clusters 0/1/2 at `efSearch=400` |

So the branch fixed the half the issue is about - the nearest group is no longer locked out - and lost the other half. Reading only `limit * groupSize * 2` nodes out of the search hands the whole candidate pool to the first group dense enough to fill it, and the remaining group slots go unused. That is not a corner case, it is the case `groupBy` exists for, and it showed up as a red test: `SQLFunctionVectorGroupByTest.denseGroupByReturnsDistinctGroupsAtGroupSize1` went from 10 distinct groups to 8 on this branch.

`efSearch` could not buy the groups back either. The comment added in this PR says JVector's `resume()` re-runs the search when the returned count is short, but there is no `resume()` call on this path; `efSearch` only widens the beam, and the beam was not what was being read. `efSearch=300` returned the same two rows as the default.

**Follow-up commit** takes the whole beam instead of `limit * groupSize * 2`. `topK` never reaches the layer-0 traversal (`searchLayer0` passes only `rerankK` to `searchOneLayer`; `topK` just sizes the pop) and `rerankK` is unchanged, so every one of those candidates was already scored and reranked - no extra graph work, and `efSearch` becomes the lever the contract promises. Also: unresolvable group keys are counted apart from cap hits instead of being logged as full groups, the two per-query log lines drop from INFO to FINE to match the ungrouped path, `GroupedRIDBitsFilter` gets a note saying it is no longer wired into anything and why it is kept, and the javadoc states the candidate-pool cost instead of claiming a guarantee the code cannot make.

**Tests.** `Issue5761GroupedSearchGroupChoiceTest` covers the reported shape at the index and at `vector.neighbors`, the pool/`efSearch` relationship, and group choice at every point on the arc. All four fail on `main`. `Issue5558DeletedRegionSearchTest.theGroupedSearchAnswersADeletedRegionQuery` now asserts the cluster rather than liveness, which is what the issue asks for, and fails on `main` too. Full `com.arcadedb.index.vector` + `com.arcadedb.function.sql.vector` sweep: 566 tests, the only red one was `LSMVectorIndexIncrementalIngestScalingTest.graphShouldKeepAbsorbingPendingVectorsWithoutFurtherSearches`, a 2-minute awaitility that timed out under CPU contention and passes in 1.6 s in isolation.

**Left for a follow-up, not for this PR.** Fewer than `limit` groups still come back when one group is denser than the beam, which is the documented cost of moving the cap out of the traversal. Closing that needs option 2 from the issue - admitting into a per-group min-heap so a better candidate can evict a worse one, with the traversal still pruning full groups - so #5761 may be worth keeping open against that.

robfrank:
Review notes on this one, leaving it open for you or a maintainer to pick up rather than merging as-is:

**Disclosure:** we rebased this branch onto current `main` and pushed a follow-up commit (`fix(engine) #5761: read the whole beam so the grouped search still an...`) to fix issues found in review (see below). This was a force-push (`--force-with-lease`) to your branch `fix/issue-5761-grouped-search-traversal-order` - your original commit's authorship is preserved, but if you have local work on this branch you will need to `git fetch` and reset to the new remote head before pushing again, or you'll get a non-fast-forward rejection.

**What review found:**
- The diagnosis and approach (option 1 from #5761: choose groupBy groups by score instead of by HNSW traversal order) are correct.
- As originally submitted, the branch had no regression test, and reading only `limit*groupSize*2` nodes from the search handed the whole candidate pool to the densest group - this broke an existing test (`SQLFunctionVectorGroupByTest.denseGroupByReturnsDistinctGroupsAtGroupSize1`, which regressed from 10 groups to 8) and didn't fully fix the issue's own reproduction (cluster 1 was still missing from the answer).
- The follow-up commit reads the whole search beam instead of a fixed multiple, which fixes both of those, plus adds a dedicated regression test (`Issue5761GroupedSearchGroupChoiceTest`) confirmed to fail without the fix and pass with it.
- **Residual gap, not fixed by either commit:** if one group is denser than the search beam, fewer than `limit` distinct groups can still come back, because nothing prunes by group during the graph traversal itself. Widening `efSearch` works around it (verified: default beam of 100 on a 100-members-per-cluster fixture returns 1 group, `efSearch=400` returns the correct 3), but the real fix is option 2 from the issue - a score-aware per-group min-heap that prunes during traversal, not after. That's a separate, larger change.

Given the residual gap, whether to merge this as a partial fix (with #5761 left open for the traversal-time fix) or hold for the full fix is a judgment call for the ArcadeDB team. CI on the current head: `unit-tests` and `integration-tests` are red at the "Check test results" step - worth confirming whether that's pre-existing on `main` or caused by this diff before merging either way.

lvca:
Thanks @waterWang !

Issue #3722, opened 2026-03-25
Missing images when vector neighbors query


lekmaneb:
Hi,

Query:
```sql
SELECT 
    distance AS distance,
    `record`.out('embb')[0].asRID() AS rid
FROM (
[...]

PR #5787, opened 2026-08-03
fix(engine) #5747: do not rebuild a persisted HNSW graph on close

### LINKED ISSUE #5747: Any session that does not run a vector search pays a full HNSW rebuild on close (602-843 ms vs 17 ms)
**CORRECTION (2026-08-02). My headline was wrong, and the graph persistence in
this engine works. Please read this before the measurements below.**

I earlier claimed close rebuilds the HNSW graph unconditionally. That was an
artifact of my own harness. My lifecycle loop opens a database and closes it
**without ever running a query**, and the persisted graph is loaded lazily on
first search. With no search, the load never happens, and close then rebuilds so
it has a graph to persist.

Same database, same process, alternating:

```
queried=false   close = 864.8 ms      (rebuilds)
queried=true    close =  17.4 ms      <- "Loaded graph from disk: 10000 nodes"
queried=false   close = 747.5 ms
queried=true    close =  16.9 ms
```

**50x, and the persisted graph loads exactly as designed.** A session that
actually uses the index closes in 17 ms, which is well inside any reasonable
budget. The load path, the staleness check at `LSMVectorIndex` around line 1293,
and the `.vecgraph` file all work. Apologies for the noise; I should have run
this comparison before writing the earlier version.

**What may still be worth a look, stated much more narrowly than before.** A
process that opens a database with a vector index and closes it without querying
pays a full graph rebuild, roughly 50x the cost of the load-and-close path, to
produce a graph byte-identical to the one already on disk. That is a real if
narrow waste: batch jobs, health checks, schema inspections and short-lived
tooling all do exactly that. If close could take the same lazy-load path before
deciding it needs to rebuild, that case would collapse to the 17 ms figure.

**Which sessions pay it, measured.** I assumed this only affected a session that
does nothing. It is broader: only a vector search avoids the rebuild.

| what the session did before closing | close |
|---|---|
| nothing at all | 843 ms |
| `SELECT count(*) FROM Vec` | 717 ms |
| read one document by scan | 650 ms |
| write to an unrelated document type | 602 ms |
| **vector search** | **17 ms** |

Reading documents, counting rows, and writing to a completely different type all
still pay a full HNSW rebuild. Only the vector search triggers the lazy load that
makes close cheap.

For a multi-model engine that seems like the point worth making: the cost of
**having** a vector index is charged to **every** session, not only the sessions
that use it. A bulk document ingest, a migration, an admin query, or any part of
an application that touches documents but not vectors in a given session pays a
full graph rebuild on close, to write a graph identical to the one already on
disk.

If close took the same load-or-rebuild decision the search path already takes,
all five rows above would presumably look like the last one.

Everything below this line predates the correction. The non-vector measurements
in it are unaffected, since none of them involve a lazily loaded structure. The
vector rows should be read as "close without ever querying", not as the cost a
normal session pays.

---

`Database.close()` is O(1) in data size for ordinary types and LSM_TREE indexes,
and O(size) when a vector index or a Graph Analytical View is present. At 1M
64-dimensional vectors, closing a clean database that nothing has modified takes
**34.5 seconds**.

Measured on 26.8.1-SNAPSHOT (engine jars from our 2026-08-01 build of `main`),
laptop i9-12900HK, ext4. Each database is built once, then opened and closed 9
times with the first 3 discarded; median reported. Nothing is written between
open and close, so this is the cost of closing a **clean** database.

## Clean close, ms

| situation | 1k | 10k | 100k | 300k | 1M |
|---|---|---|---|---|---|
| doc, no index | 2.4 | 2.5 | 2.1 | | |
| doc, 30 LSM_TREE indexes | 14.4 | 14.6 | 14.0 | 37.7 | 38.4 |
| graph, 4 edges/vertex | 3.8 | 3.6 | 3.3 | | |
| graph + analytical view | 10.0 | 24.4 | 186.9 | 532.7 | **2376.9** |
| 64-dim LSM_VECTOR | 50.6 | 510.7 | 3068.2 | 9049.1 | **34503.6** |

Eight further situations (empty, doc with 1/3/10 indexes, graph with an index,
and so on) are all flat: a hundred times the rows moves close by less than the
run-to-run spread.

Open is 0.4 to 3 ms almost everywhere including the analytical-view rows, and
reaches 456 ms for the 1M vector case.

## Three separate costs

**1. A fixed commit-and-close cost, about 15.6 ms, size-independent.** Inserting
a single document before closing adds ~15.6 ms on every flat situation,
including an empty database at every size. Consistent with one disk sync.

**2. About 0.40 ms per LSM_TREE index, independent of index contents.** One JVM
per configuration, 100k rows, real disk:

| indexes | 0 | 1 | 3 | 10 | 30 |
|---|---|---|---|---|---|
| clean close (ms) | 1.93 | 2.47 | 3.34 | 6.08 | 13.97 |

Linear in the number of indexes and flat in the number of rows in them. Thirty
indexes is not an unusual schema.

**3. Size-proportional persist for derived structures.** Only the analytical
view and the vector index scale, and both are structures the engine derives
rather than stores as records. From 300k to 1M (3.33x the data) the view goes
x4.46 and the vector index x3.81, while the 30-index control goes x1.02.

## A discriminator that separates 2 from 3

Running the identical code against tmpfs instead of ext4:

| | ext4 | tmpfs |
|---|---|---|
| 30 indexes, 100k | 14.0 ms | 0.5 ms |
| vector, 100k | 3068 ms | 3216 ms |

The per-index cost **disappears** on a RAM-backed filesystem, so it is I/O. The
vector cost is **unchanged**, so it is not I/O, it is work being done on the
close path. Those want different fixes, which is why it seems worth separating
them rather than reporting one "close is slow".

## It is not "vector indexes", it is the dense one specifically

`LSM_SPARSE_VECTOR` was added to the matrix afterwards and is completely flat:

| documents | clean close | index on disk |
|---|---|---|
| 1,000 | 2.92 ms | 0 MB |
| 10,000 | 2.80 ms | 3 MB |
| 100,000 | 2.99 ms | 31 MB |
| 300,000 | 2.85 ms | 94 MB |

**94 MB of sparse index closes in 2.85 ms.** The dense index at 100k vectors is
54 MB, less data, and takes 3068 ms. Roughly a thousand times worse on a smaller
index.

The two do not share an implementation: `LSMVectorIndex` imports
`io.github.jbellis.jvector.*`, and `LSMSparseVectorIndex` imports no JVector at
all. So this is not a property of vector indexes in general, and not of anything
inherent to persisting a large derived structure. The in-house sparse
implementation already closes in constant time at scale; the JVector-backed
dense path does not.

That seems worth saying because it turns the question from "how do we make
closing a big index cheap" into "why does this one path do work the other one
does not", and there is a working example of the desired behaviour in the same
codebase.

## What a good ceiling looks like, since "close is slow" is not actionable

Two criteria, and only one of them is a number.

**1. Asymptotic, and this is the real one: close should be O(what was written),
not O(what is stored.)** Open a database, write nothing, close it, and the cost
should be the same at 1 KB and at 1 TB. That is a design property rather than a
performance target, and it is what the table above actually tests. **Eight of the
ten situations already satisfy it**: documents, graphs, and even thirty LSM_TREE
indexes are flat across a hundred times the data. Only the two derived
structures scale, and they scale *even when nothing was modified*: for the
vector case a clean close (3068 ms) and a close after writing one document
(3244 ms) are the same number, so the work is unconditional rather than
proportional to changes.

**2. Absolute, and it should be small: on the order of 100 ms.** Grounded in two
things rather than taste:

- **Close should not cost more than getting started.** The JVM behind the
  embedded engine takes roughly 160 ms to come up. If putting the database away
  costs more than opening the process, the lifecycle is inverted.
- **It should sit below human perception.** Roughly 100 ms is the threshold at
  which a script that opens, queries and exits stops feeling instant.

For a reference point, SQLite closes a clean database in microseconds. Nobody
has ever needed to profile `sqlite3_close`. That is the bar embedded engines are
measured against, fairly or not.

Measured against those two criteria:

| | close | verdict |
|---|---|---|
| everything except views and vector indexes | 1.8 to 38 ms | passes both |
| analytical view, 1M vertices | 2.4 s | about 24x over |
| vector index, 1M vectors | 34.5 s | about 350x over |

**Why this matters more for embedded than it would for a server.** A served
engine opens its databases once and amortises the cost across days of uptime,
so even a slow close is invisible. An embedded engine is opened and closed by
whatever process embeds it, which may be a script, a CLI invocation, a test, or
a request handler. Concretely: a 10M-vector index in a script that opens, runs
one query and exits spends several seconds opening and, extrapolating this
curve, well over a minute closing, to do about a millisecond of useful work.
That is the case where this stops being a detail.

## The part that looks most actionable

The analytical view already solved this on one side. From the log on open:

```
GraphAnalyticalViewPersistence  Restoring 1 Graph Analytical View(s) asynchronously on database open
CSRBuilder                      CSR built: ... 14 ms
```

Restore was deliberately made **asynchronous on open**, which is why open stays
flat at 1.5 to 3 ms even at 1M vertices. The persist on close is
**synchronous**. The cost was not removed, it moved to the side where nothing
hides it. The same treatment on the persist side would make close look like
open, and the pattern is already in the codebase.

For the vector index the equivalent question is whether anything on the close
path can be done incrementally or skipped when the index is unchanged since
open, since these measurements never modify it and still pay full price every
cycle.

## Reproduction

Self-contained, engine API only.

```java

import java.io.File;
import java.util.ArrayList;
import java.util.Collections;
import java.util.List;

/**
 * Database.close() is O(1) in data size for ordinary types and indexes, and
 * O(size) when a vector index is present.
 *
 * Two arms, each measured at two sizes:
 *   docs30  30 documents types worth of LSM_TREE indexes, no derived structure
 *   vector  one LSM_VECTOR index over 64-dim embeddings
 *
 * Build once, then 9 open/close cycles, first 3 discarded, median reported.
 * Nothing is modified between open and close, so this is the cost of closing a
 * clean database and not the cost of flushing a write.
 *
 *   java -Xmx10g --class-path "<arcadedb jars>/*" CloseCostRepro.java
 */
public class CloseCostRepro {

  static final int REPS = 9, WARMUP = 3, DIM = 64;

  public static void main(String[] args) {
    // One arm per JVM invocation. The first version cycled every arm in a
    // single JVM and docs30 measured 14 ms there against 0.5 ms standalone,
    // so accumulated per-process state was contaminating the per-index number.
    // Isolating the process is the only way to know which figure is the
    // database's and which is the JVM's.
    final String[] arms = args.length > 0 ? args[0].split(",")
        : new String[] {"docs30", "vector"};
    final int[] sizes = args.length > 1
        ? java.util.Arrays.stream(args[1].split(",")).mapToInt(Integer::parseInt).toArray()
        : new int[] {10_000, 100_000};
    for (final String arm : arms)
      for (final int n : sizes) {
        final String path = "/home/tk/gavab/cc_" + arm + "_" + n;
        build(path, arm, n);
        final List<Double> o = new ArrayList<>(), c = new ArrayList<>();
        for (int i = 0; i < REPS; i++) {
          final Database db = new DatabaseFactory(path).open();
          final long t1 = System.nanoTime();
          db.close();
          final long t2 = System.nanoTime();
          if (i >= WARMUP) { o.add((t1 - t0) / 1e6); c.add((t2 - t1) / 1e6); }
        }
        System.out.printf("RESULT %-10s %10d %12.2f %12.2f%n", arm, n, med(o), med(c));
        rm(new File(path));
      }
  }

  static void build(final String path, final String arm, final int n) {
    rm(new File(path));
    final Database db = new DatabaseFactory(path).create();
    if (arm.startsWith("docs")) {
      final int nIdx = Integer.parseInt(arm.substring(4));
      db.command("sql", "CREATE DOCUMENT TYPE Item");
      for (int c = 0; c < 30; c++)
        db.command("sql", "CREATE PROPERTY Item.p" + c + " INTEGER");
      for (int c = 0; c < nIdx; c++)
        db.command("sql", "CREATE INDEX ON Item (p" + c + ") NOTUNIQUE");
      db.begin();
      for (int i = 0; i < n; i++) {
        final MutableDocument d = db.newDocument("Item");
        for (int c = 0; c < 30; c++)
          d.set("p" + c, (i * 31 + c) % 100003);
        d.save();
        if ((i + 1) % 20000 == 0) { db.commit(); db.begin(); }
      }
      db.commit();
    } else {
      db.command("sql", "CREATE DOCUMENT TYPE Vec");
      db.command("sql", "CREATE PROPERTY Vec.id INTEGER");
      db.command("sql", "CREATE PROPERTY Vec.emb ARRAY_OF_FLOATS");
      db.begin();
      for (int i = 0; i < n; i++) {
        final float[] e = new float[DIM];
        for (int d2 = 0; d2 < DIM; d2++)
          e[d2] = ((i * 31 + d2 * 7) % 1000) / 1000f;
        db.newDocument("Vec").set("id", i).set("emb", e).save();
        if ((i + 1) % 20000 == 0) { db.commit(); db.begin(); }
      }
      db.commit();
      db.command("sql", "CREATE INDEX ON Vec (emb) LSM_VECTOR "
          + "METADATA {dimensions: " + DIM + ", similarity: 'COSINE'}");
    }
    db.close();
  }

  static double med(final List<Double> xs) {
    final List<Double> s = new ArrayList<>(xs);
    Collections.sort(s);
    return s.get(s.size() / 2);
  }

  static void rm(final File f) {
    final File[] k = f.listFiles();
    if (k != null) for (final File x : k) rm(x);
    f.delete();
  }
}
```

```
java -Xmx10g --class-path "path/to/arcadedb/jars/*" CloseCostRepro.java docs30,vector 10000,100000
```

Write the databases to a real filesystem. The class hardcodes a path under
`$HOME` on purpose: pointing it at `/tmp` on a machine where `/tmp` is tmpfs
silently removes the disk-sync component and made an early version of this
report understate the index cost by 28x.

## Why it matters for embedded use

A served engine opens its databases once and amortizes this over the process
lifetime. An embedded engine is opened and closed by whatever short-lived
process embeds it, so a 34.5 second close is paid per run rather than per
deployment. That is the case where this stops being a detail.

Happy to run any variant, at other sizes, or with a profiler attached.

tae898:
Fixes #5747.

`LSMVectorIndex.flush()` rebuilds the HNSW graph when `graphState` is `LOADING`
or `MUTABLE`. `MUTABLE` is right: there are pending changes. `LOADING` is not,
because it means "not loaded into memory", which is not the same as "not
persisted on disk".

The graph loads lazily on the first search, so a session that only reads
documents leaves the state at `LOADING` even when a complete graph is already on
disk, and close then rebuilds it from scratch to produce a file that already
exists. `initializeGraphIndex()` already draws exactly this distinction with
`graphFile.hasPersistedGraph()`; this applies the same test in `flush()`.

## Measured

One host, 64-dim COSINE, 9 reps with 3 warmup, median. Both arms are the same
harness on the same data; a classpath check confirms which jar each arm loaded.

| vectors | what the session did | close before | close after |
|---|---|---|---|
| 100k | read a document | 6,197.79 ms | **4.34 ms** |
| 1M | read a document | 216,068.3 ms | **13.2 ms** |
| 1M | two indexes, queried neither | 379,014.3 ms | **15.2 ms** |
| 1M | two indexes, queried one | 190,448.5 ms | **14.9 ms** |
| 1M | **ran a vector search** | 14.1 ms | 16.3 ms |
| 3M | read a document | 1,006,000 ms | not re-run |

The fifth row is the safety check. Sessions that do search already skipped the
rebuild, and this leaves them where they were.

At 3M the close was 1,006 s against an original build of 1,011 s, so the close
rebuild is a complete re-run of the build. The cost is also per index rather
than per database: with two indexes and only one queried, the untouched one
still cost 190 s on its own.

## Correctness

- New test `VectorIndexNoRebuildOnUnusedCloseTest` asserts the persisted graph
  is byte-identical and unmodified after a session that reads documents and
  never searches. It **fails on unpatched code** and passes with the change.
- Separately: after adding 200 vectors to an existing index, closing and
  reopening, all 200 new and all 200 pre-existing vectors are still found by
  search, identically on patched and unpatched builds. The `MUTABLE` branch
  still rebuilds as it should.
- The full `com.arcadedb.index.vector` suite (22 classes) passes.

## One behaviour change worth stating

If a session adds vectors and then crashes before close, the on-disk graph is
stale and the state is `LOADING`. The current unconditional rebuild repairs that
incidentally. With this change the repair happens at the next search instead,
through the staleness check already in `ensureGraphAvailable()`
(`graphSize < rebuiltOrdinalToVectorId.length`, added for #3722). So the repair
moves from incidental to deliberate rather than disappearing, but it is a real
change and you may prefer a different call on it.

Happy to adjust the approach or add cases to the test.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -4907,7 +4907,16 @@ public class LSMVectorIndex implements Index, IndexInternal {
       // Build and persist graph if it hasn't been built yet
       // This ensures the graph is available on next database open (fast restart)
       // Build graph if it's in LOADING (never built) or MUTABLE (has pending changes) state
-      if (vectorIndex.size() > 0 && (graphState == GraphState.LOADING || graphState == GraphState.MUTABLE)) {
+      //
+      // LOADING means "not loaded into memory", which is not the same as "not
+      // persisted on disk": the graph loads lazily on the first search, so a
+      // session that never searched leaves it LOADING even when a complete
+      // graph is already on disk, and rebuilding then only reproduces a file
+      // that already exists. initializeGraphIndex() already draws exactly this
+      // distinction with the same predicate.
+      final boolean graphAlreadyOnDisk = graphFile != null && graphFile.hasPersistedGraph();
+      if (vectorIndex.size() > 0 && (graphState == GraphState.MUTABLE
+          || (graphState == GraphState.LOADING && !graphAlreadyOnDisk))) {
         try {
           LogManager.instance()
               .log(this, Level.FINE, "Building graph before close for index: %s (this may take 1-2 minutes for large datasets)",
--- a/engine/src/test/java/com/arcadedb/index/vector/VectorIndexNoRebuildOnUnusedCloseTest.java
+++ b/engine/src/test/java/com/arcadedb/index/vector/VectorIndexNoRebuildOnUnusedCloseTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.index.vector;
+
+import com.arcadedb.TestHelper;
+import com.arcadedb.query.sql.executor.ResultSet;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.nio.file.Files;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A session that never searches the vector index must not rewrite its HNSW graph on close.
+ * <p>
+ * {@code GraphState.LOADING} means "not loaded into memory", which is not the same as "not
+ * persisted on disk": the graph is lazily loaded on the first search, so a session that only
+ * reads documents leaves the state at LOADING even when a complete graph is already on disk.
+ * Treating that as "never built" makes close rebuild the graph from scratch to produce a file
+ * that already exists, which is O(n log n) in the index size and lands on sessions that never
+ * used the index at all.
+ *
+ * @see <a href="https://github.com/ArcadeData/arcadedb/issues/5747">issue #5747</a>
+ */
+class VectorIndexNoRebuildOnUnusedCloseTest extends TestHelper {
+
+  private static final int VECTORS = 300;
+
+  @Test
+  void closingWithoutSearchingDoesNotRewriteTheGraph() throws Exception {
+    database.transaction(() -> {
+      database.command("sql", "CREATE DOCUMENT TYPE Doc");
+      database.command("sql", "CREATE PROPERTY Doc.name STRING");
+      database.command("sql", "CREATE PROPERTY Doc.embedding ARRAY_OF_FLOATS");
+      for (int i = 0; i < VECTORS; i++)
+        database.command("sql", "INSERT INTO Doc SET name = ?, embedding = ?", "d" + i, embedding(i));
+    });
+
+    database.command("sql", """
+        CREATE INDEX ON Doc (embedding) LSM_VECTOR
+        METADATA {
+          "dimensions": 4,
+          "similarity": "COSINE"
+        }""");
+
+    // Search once so the graph is loaded and persisted in a settled state, then close.
+    assertThat(search()).isNotZero();
+    reopenDatabase();
+
+    final File graph = graphFile();
+    final byte[] contentBefore = Files.readAllBytes(graph.toPath());
+    final long modifiedBefore = graph.lastModified();
+
+    // A session that reads documents and never touches the vector index. Nothing here
+    // changes the index, so closing must leave the persisted graph exactly as it was.
+    try (final ResultSet rs = database.query("sql", "SELECT name FROM Doc LIMIT 10")) {
+      while (rs.hasNext())
+        rs.next();
+    }
+    reopenDatabase();
+
+    assertThat(graphFile()).exists();
+    assertThat(Files.readAllBytes(graphFile().toPath()))
+        .as("closing a database whose vector index was never used must not rewrite the graph")
+        .isEqualTo(contentBefore);
+    assertThat(graphFile().lastModified())
+        .as("the persisted graph was rewritten on close by a session that never searched")
+        .isEqualTo(modifiedBefore);
+
+    // And the index is still usable afterwards.
+    assertThat(search()).isNotZero();
+  }
+
+  private int search() {
+    int rows = 0;
+    try (final ResultSet rs = database.query("sql",
+        "SELECT name FROM (SELECT expand(vectorNeighbors('Doc[embedding]', [1.0, 0.0, 0.0, 0.0], 5)))")) {
+      while (rs.hasNext()) {
+        rs.next();
+        rows++;
+      }
+    }
+    return rows;
+  }
+
+  private File graphFile() {
+    final File[] found = new File(getDatabasePath())
+        .listFiles((dir, name) -> name.contains(LSMVectorIndexGraphFile.FILE_EXT));
+    assertThat(found).as("persisted graph file").isNotNull().isNotEmpty();
+    return found[0];
+  }
+
+  private static float[] embedding(final int i) {
+    // Spread over the unit sphere so the graph has real structure to preserve.
+    final double a = i * 0.05;
+    return new float[] {(float) Math.cos(a), (float) Math.sin(a), (float) (i % 7) / 7f, (float) (i % 3) / 3f};
+  }
+}


lvca:
Thanks @tae898 !

PR #5792, opened 2026-08-03
fix(serializer) #5770: serializeForRead renders EXTERNAL values inline on read paths

### LINKED ISSUE #5770: An after-read listener that returns a dirty record with an EXTERNAL property writes to the external bucket on a read path
Split out of #5755 (PR #5769), where it was noted but deliberately not folded in.

## The gap

When an `AfterRecordReadListener` returns a **different** record, the caller renders it into a buffer through the serializer:

- `ImmutableDocument.checkForLazyLoading()` - has always done this
- `BaseRecord.reload()` - does this as of #5769, so the exposure is now shared by both read paths

For a dirty record that goes through `BinarySerializer.serializeProperties()`, and that method does more than serialize. For any property the schema marks `EXTERNAL` it calls `writeExternalPropertyValue(...)`, which **creates or updates a record in the external bucket**, and `findExistingExternalRids(...)` / the orphan-cleanup block at the end can **delete** external records that are no longer referenced.

So a read - `get()`, `toJSON()`, a scan touching the record, a `reload()` - can perform bucket writes, purely because a listener handed back a modified record and one of that record's properties is external.

## Why it matters

- Writes on a read path are surprising on their own, and they land wherever the read happens to be running - possibly outside an explicit transaction, possibly on a query thread.
- The orphan cleanup can **delete** external records as a side effect of reading.
- It is silent: nothing in the listener contract suggests returning a modified record can write.

## Not currently reachable through SQL

`TriggerListenerAdapter.onAfterRead()` returns the *same* record, so READ triggers never take the `loaded != this` branch. This needs a programmatic (embedded) listener, which is exactly what the documented record-encryption recipe (`RecordEncryptionTest`) is - so an encryption hook on a type that also has an `EXTERNAL` property is the realistic way to meet it.

## Possible directions

Rather than the obvious "skip external handling when serializing for a read", which would silently produce a buffer whose external pointers do not resolve, the better shapes look like:

1. **A read-only serialization mode.** A flag on the serialize path that renders external property *values* inline (reading the existing external records rather than writing new ones), so the resulting buffer is complete and no write occurs. Costs a read per external property, only on this path.
2. **Refuse the combination explicitly.** Raise a clear error when an after-read listener returns a modified record on a type with `EXTERNAL` properties, instead of quietly writing. Cheap, and turns a silent side effect into a diagnosable one, but it takes a working (if surprising) case away.
3. **Do not round-trip through bytes at all.** Keep the listener's returned record as the answer rather than re-serialising it into the immutable's buffer. This is the deepest fix and the largest change, since `ImmutableDocument` is buffer-backed by design.

Option 1 looks like the right end state, with option 2 acceptable as an interim if the read-only mode is too big a change to land on its own.

## Test to write

A document type with an `EXTERNAL` property, plus an after-read listener that returns `record.modify().set(...)`. Read the record and assert that the external bucket's record count and contents are unchanged by the read - which fails today.

waterWang:
## Description

Fixes #5770

When an `AfterRecordReadListener` returns a modified record, the reader re-serialises it into a buffer via `serialize()`. For EXTERNAL properties, this calls `writeExternalPropertyValue()` which creates or updates records in the paired external bucket — a write on a read path, with the additional risk of orphan cleanup deleting external records as a side effect.

### Fix

Add a read-only serialization mode (`serializeForRead`) that:

- **Renders EXTERNAL property values inline** instead of writing to the paired bucket
- **Skips the orphan-cleanup pass** entirely — no external bucket writes or deletes
- **Produces a complete, self-contained buffer** by using the existing `serializeValue()` path for EXTERNAL properties

The normal `serialize()` path is unchanged — EXTERNAL properties continue to be written to the paired bucket during regular writes.

Also updated `ImmutableDocument.checkForLazyLoading()` to use `serializeForRead()` when re-serialising a listener-modified record on the read path.

### Test

`serializeForReadDoesNotWriteToExternalBucket` — creates a document with an EXTERNAL property, serializes a modified copy via `serializeForRead()`, and asserts that the external bucket record count is unchanged.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/BaseRecord.java
+++ b/engine/src/main/java/com/arcadedb/database/BaseRecord.java
@@ -89,8 +89,10 @@ public abstract class BaseRecord implements Record {
           // CREATE A BUFFER FROM THE MODIFIED RECORD, THE SAME WAY ImmutableDocument.checkForLazyLoading() DOES.
           // TAKING getBuffer() INSTEAD WOULD DISCARD WHAT THE LISTENER DID: ON A DIRTY MutableDocument THAT BUFFER
           // STILL HOLDS THE PRE-MODIFICATION CONTENT, AND IT IS null OUTRIGHT FOR A RECORD BUILT FROM SCRATCH.
+          // serializeForRead() INSTEAD OF serialize(): THIS IS A READ, SO AN EXTERNAL PROPERTY MUST BE RENDERED INLINE
+          // INSTEAD OF BEING WRITTEN TO (OR DELETED FROM) THE PAIRED EXTERNAL BUCKET (ISSUE #5770).
           // getNotReusable() IS MANDATORY: FOR A DIRTY RECORD THE SERIALIZER RETURNS THE PER-THREAD SCRATCH BUFFER
-          buffer = database.getSerializer().serialize(database, loaded).getNotReusable();
+          buffer = database.getSerializer().serializeForRead(database, loaded).getNotReusable();
         }
 
       } catch (final RecordNotFoundException e) {
--- a/engine/src/main/java/com/arcadedb/database/ImmutableDocument.java
+++ b/engine/src/main/java/com/arcadedb/database/ImmutableDocument.java
@@ -266,9 +266,11 @@ public class ImmutableDocument extends BaseDocument {
         return false;
       } else if (loaded != this) {
         // CREATE A BUFFER FROM THE MODIFIED RECORD. THIS IS NEEDED FOR ENCRYPTION THAT UPDATE THE RECORD WITH A MUTABLE.
+        // serializeForRead() INSTEAD OF serialize(): THIS IS A READ, SO AN EXTERNAL PROPERTY MUST BE RENDERED INLINE
+        // INSTEAD OF BEING WRITTEN TO (OR DELETED FROM) THE PAIRED EXTERNAL BUCKET (ISSUE #5770).
         // getNotReusable() IS MANDATORY: FOR A DIRTY RECORD THE SERIALIZER RETURNS THE PER-THREAD SCRATCH BUFFER, WHICH
         // IT COPIES OUT, WHILE THE ALREADY-PRIVATE BUFFER OF THE OTHER BRANCH IS KEPT AS IS
-        buffer = database.getSerializer().serialize(database, loaded).getNotReusable();
+        buffer = database.getSerializer().serializeForRead(database, loaded).getNotReusable();
         // THE SERIALIZER HANDS THE BUFFER BACK AT 0 OR AT 1 DEPENDING ON THE BRANCH IT TOOK: NORMALISE IT SO THE
         // POSTCONDITION OF THIS METHOD HOLDS ON EVERY PATH
         buffer.position(propertiesStartingPosition);
--- a/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinarySerializer.java
@@ -123,25 +123,35 @@ public class BinarySerializer {
   }
 
   /**
-   * Read-only serialization: creates a buffer-backed representation of the record without writing to the external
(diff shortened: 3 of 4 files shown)


lvca (changes requested):
Thanks for the PR, @waterWang. The direction is right — a read-only serialization mode that renders EXTERNAL values inline instead of writing them — but there are two things that need to be sorted before this can merge:

1. This branch bundles an old commit for #5748 (LSM_VECTOR brute-force scan). That fix already landed on main separately (3a0c846f8), which is why this PR shows as conflicting. Please rebase off main and drop that commit — it doesn't belong in this PR.

2. The fix only covers `ImmutableDocument.checkForLazyLoading()`. Issue #5770 explicitly calls out that `BaseRecord.reload()` shares the same exposure since #5769 ("the exposure is now shared by both read paths"). `BaseRecord.reload()` still calls the writing `serialize()` there — same bug, same shape, just the other read path. Needs the same `serializeForRead()` swap.

Also: after rebasing onto current main you'll run into the #5755 fix inside `checkForLazyLoading()` (`getNotReusable()` + position-normalize, for the scratch-buffer aliasing bug). Make sure `serializeForRead()`'s result goes through the same treatment, not just `serialize()`'s.

On the test: right now it calls `BinarySerializer.serializeForRead()` directly. It'd be a lot stronger as end-to-end coverage — register a real `AfterRecordReadListener` (same recipe as `RecordEncryptionTest`) on a type with an EXTERNAL property, read it through the normal path, and assert both that the external bucket didn't change AND that the listener-modified property plus the EXTERNAL property still read back correctly. That way it reproduces the bug through the real code path instead of unit-testing the serializer in isolation.

Happy to help push a follow-up if useful.

robfrank:
Rebased onto current main and pushed the review follow-up, @waterWang. What changed:

- Dropped the `#5748` LSM vector commit. That fix is already on main (`3a0c846f8`), and the stale copy in this branch was the source of the conflict. The branch is now the `#5770` change only: `BinarySerializer`, `ImmutableDocument`, `BaseRecord`, `ExternalPropertyTest`.
- Kept `getNotReusable()` and the position normalization in `checkForLazyLoading()` (the `#5755` fix on main): `serializeForRead()` hands back the per-thread scratch buffer for a dirty record exactly like `serialize()` does, so it needs the same treatment.
- `BaseRecord.reload()` now goes through `serializeForRead()` too. That was the second read path named in the issue, and it was still writing to the external bucket.
- `serializeForRead()` falls back to `serialize()` for record types with no properties (edge segment, stripe directory, external value record) instead of throwing: `reload()` is generic over every record type and none of those shapes can reach the external bucket.

On the test, the direct `serializeForRead()` call could not fail: the buggy path UPDATES the external blob in place, it does not add one, so an external-bucket count assertion holds either way. The suite now drives the real read path with an `AfterRecordReadListener` and asserts on the STORED value:

- `afterReadListenerReturningAModifiedRecordDoesNotWriteToTheExternalBucket` - lazy-load path, also asserts zero modified pages in the reading transaction
- `reloadWithAnAfterReadListenerDoesNotWriteToTheExternalBucket` - `reload()` path
- `afterReadListenerOnAVertexDoesNotWriteToTheExternalBucket` - vertex shape, which is what the encryption recipe uses
- `afterReadListenerRewritingAPlainPropertyStillResolvesTheExternalValue` - guard against the shortcut the issue rejects: rendering inline must keep the EXTERNAL value readable, not drop it
- `serializeForReadRendersExternalValuesInline` - the produced buffer resolves the external value with no second lookup

With the two `serializeForRead()` call sites reverted, the first three fail with `expected: "encrypted_value" but was: "decrypted_value"`: reading the record had overwritten the stored blob with the listener's value. That is worse than the issue described, it is silent data loss on a read, not only a surprising write. With the fix in place `ExternalPropertyTest` is 36/36 green.

One trade-off is now written into the javadoc: the read-only buffer carries no TYPE_EXTERNAL pointer, so if application code takes that materialized record and later saves an update of it, the update allocates a fresh blob and leaves the previous one to the orphan scan. That is the accepted cost of not writing on a read; a pointer-preserving variant would be a separate change.

Issue #5321, opened 2026-07-18
Partial-prefix lookup on a composite index returns rows of OTHER keys when the first key component contains some accented characters


adepase:
**ArcadeDB version**: 26.7.2 (official docker image), single server, remote HTTP API.
Reproduced with index compaction DISABLED (`-Darcadedb.indexCompactionMinPagesSchedule=0`) on a freshly created index, so it does not look related to the compacted-page reading issues fixed after the 26.7.2 tag.
**Also reproduced UNCHANGED on 26.7.3** (official docker image), both on a fresh database with default settings and on a server upgraded in place from 26.7.2: same result sets as below, 2026-07-18.

## Summary

Given a NOTUNIQUE composite index on three STRING properties `(w, l, t)`:

- equality on the FULL key (`w = ? and l = ? and t = ?`) is always correct;
- equality on a PARTIAL prefix (`w = ? and l = ?`) returns rows belonging to  **other keys** when `w` contains certain non-ASCII characters (Italian accented vowels). With `w = 'è'` the scan returns the whole table.

Not every accented value misbehaves: in the reproduction below `'cioè'` and `'è'` break while `'così'` and `'perché'` are fine - it looks like the upper bound of the prefix range scan is computed incorrectly for some multi-byte final characters, widening the range to a shorter prefix (with `'cioè'` the result set is exactly the `'cio*'` words; with `'è'` it degenerates to a full
range).

## Reproduction

```sql
CREATE DOCUMENT TYPE ReproN7;
CREATE PROPERTY ReproN7.w STRING;
CREATE PROPERTY ReproN7.l STRING;
CREATE PROPERTY ReproN7.t STRING;
CREATE INDEX ON ReproN7 (w, l, t) NOTUNIQUE;

-- 10 words x 2 types
INSERT INTO ReproN7 SET w = 'cioccolato', l = 'it', t = 'A'; INSERT INTO ReproN7 SET w = 'cioccolato', l = 'it', t = 'B';
INSERT INTO ReproN7 SET w = 'cinema',     l = 'it', t = 'A'; INSERT INTO ReproN7 SET w = 'cinema',     l = 'it', t = 'B';
INSERT INTO ReproN7 SET w = 'cioè',       l = 'it', t = 'A'; INSERT INTO ReproN7 SET w = 'cioè',       l = 'it', t = 'B';
[...]

PR #5804, opened 2026-08-03
fix(engine) #5802: a mis-ordered index after an upgrade is reported as a queryable upgrade warning

### LINKED ISSUE #5802: Existing LSM_TREE indexes report "physically ordered differently than the current key comparator" after upgrading binaries in place from 26.7.2 to a main build
## Environment
- Old version (wrote the on-disk index data originally): `26.7.2` (build `a53338edd645b92a8a866738976f03b6831b190b`)
- New version (opened the same data files): `main` at commit `bf5137ce091ff14b66873a5603005951be9e2a90`, reported as `26.8.1-SNAPSHOT` — includes commit `baa4877c2` (the LSM_VECTOR incremental-ingestion fix from #5391, which is why we were testing this build)
- OS: Ubuntu (AWS EC2), native systemd install (not Docker)
- Java: OpenJDK 21

## What we did
Took a production install running `26.7.2` with existing databases (multiple, several weeks of accumulated ingestion, millions of records), stopped the service, replaced only `lib/*.jar` and the stock `bin/*.sh` scripts with a build from current `main` (kept our own custom startup wrapper script untouched), and restarted. No schema changes, no explicit `REBUILD INDEX`, no data modification — just swapped the engine binaries in place and let the existing on-disk databases open normally.

## What we saw
On startup, before we ran any query at all, the log emitted multiple warnings like:

```
WARNI [LSMTreeIndexCompacted] Index 'Paper_0_84306331895885' is physically ordered differently
than the current key comparator, so lookups can return fewer (or foreign) records than a scan:
run 'REBUILD INDEX Paper_0_84306331895885'. Details: [page 20 entry 8: key [title:single] is
stored after the greater key [with] (the physical key order does not match the current
comparator), page 20 entry 9: key [title:security] is stored after the greater key
[title:single] ..., page 20 entry 10: key [title:robust] is stored after the greater key
[title:security] ...]
```

```
WARNI [LSMTreeIndexCompacted] Index 'Author_0_525774362082295' is physically ordered differently
than the current key comparator... Details: [page 68 entry 1: key [Zihao Zou] is stored after
the greater key [Á Rodríguez-Lescure] ..., page 68 entry 2: key [Zi-Xuan Guo] is stored after the
greater key [Zihao Zou] ..., page 68 entry 5: key [Ziyi Xu] is stored after the greater key
[Zobaida Lahari] ...]
```

```
WARNI [LSMTreeIndexCompacted] Index 'Paper_0_163197900681700' is physically ordered differently...
Details: [page 10456 entry 3: key [abstract:complicates] is stored after the greater key
[language] ..., page 10456 entry 6: key [abstract:entropies] is stored after the greater key
[needed] ...]
```

```
WARNI [LSMTreeIndexCompacted] Index 'Author_0_1127867996109552' is physically ordered
differently... Details: [page 0 entry 1: key [Zuohua Xie] is stored after the greater key
[Àngels Sahuquillo Estrugo] ..., page 0 entry 2: key [Zi Cong Zeng] is stored after the greater
key [Zuohua Xie] ...]
```

Notably this affects **two different index kinds** — a `FULL_TEXT` index (`Paper[title,abstract]`, term keys like `title:single`/`abstract:complicates`) and a plain non-unique range index on a string property (`Author.name`) — so whatever changed, it doesn't look scoped to full-text specifically; it looks like a change to string key comparison in general. The `Author.name` examples are suggestive: several of the "out of order" pairs involve accented Latin characters (`Á Rodríguez-Lescure`, `Àngels Sahuquillo Estrugo`) sorting differently relative to unaccented names than the physical page order expects — consistent with (though we haven't confirmed) a change in collation/locale-aware comparison vs. ordinal/codepoint comparison somewhere between `26.7.2` and current `main`.

WAL crash-recovery for all three databases on the box completed cleanly and without error immediately before these warnings, so this isn't related to the earlier forced-kill/SIGKILL restart — the warnings are about the physical key order on disk being inconsistent with the *current build's* comparator, not about corruption from the shutdown.

## Impact
Per the warning's own text: "lookups can return fewer (or foreign) records than a scan" on any affected index until it's rebuilt. For us that means keyword/full-text search and author lookups could silently return incomplete or wrong results on an in-place upgrade, with no forced/automatic remediation — the fix (`REBUILD INDEX ...`) is opt-in and the affected index names aren't obvious to a user who didn't know to look at startup logs.

We rolled back to `26.7.2` rather than run `REBUILD INDEX` on everything, since we don't yet understand root cause and didn't want to commit further to an unreleased build in production.

## Ask
- Was there an intentional change to string/key comparison semantics between `26.7.2` and current `main`? If so, is there supposed to be an automatic migration/reindex path for existing databases upgrading in place, or is `REBUILD INDEX` on affected indexes an expected required manual step that should be documented/surfaced more prominently than a WARN-level startup log line?
- Is there a way to know *which* indexes are affected without grepping startup logs (e.g. a `CHECK INDEX` command, or a report from `REBUILD INDEX *`)?

robfrank:
Closes #5802

## The report is accurate

The disorder is real, not a false positive from the new detector. `BinaryComparator.compareBytes(byte[], Binary)` - the routine the LSM binary search uses to place and to find a `STRING` key - compared bytes as **signed** until #5321 (`02d5800ea`), while every other string comparison in the engine is unsigned. A UTF-8 lead byte is `>= 0x80`, negative as a Java byte, so `Á` sorted before every ASCII letter under the old comparison and after it under the new one.

That is why the reporter's `Author.name` evidence opens with an accented name, and why several of the reported pairs are then pure ASCII: the mutable pages were laid out in signed order, and a compaction reads them with cursors that assume ascending order and merges them with `compareKeys` (object comparison, always unsigned). Once the two orders diverge at one accented key the merge emits a stretch of keys out of order - ASCII neighbours included - and writes them as-is. #5321 did not create the disorder; it made the reader disagree with what an older build had already written.

`REBUILD INDEX` is therefore the right remedy, and it cannot be automatic: rebuilding every index of a multi-million-record database on the open path would turn a restart into an outage.

## What was actually wrong

The reporting. `LSMTreeIndexCompacted.checkKeyOrderOnLoad()` (added in `8489de7fc`) was a dead end for an operator: it named the **physical bucket sub-index** (`Paper_0_84306331895885`) rather than the logical `Paper[title,abstract]` they would rebuild, and it existed only as a log line, so a restart whose logs had rotated away left no trace. `CHECK DATABASE` does report it, but it walks every page of every index - not something to run on the box you are trying to bring back up.

The engine already had the right channel: `IndexInternal.getUpgradeWarning()`. `LocalSchema` logs it once per database open per **logical** index with the `REBUILD INDEX` to run, and it is exposed as `upgradeWarning` on `schema:indexes` and `schema:index:<name>`, which is what Studio renders. The key-order check was simply not wired into it.

## Change

- `LSMTreeIndexCompacted` caches the check's verdict instead of only logging it, so the reader stays cheap and side-effect free as the `getUpgradeWarning()` contract requires. The physical page evidence drops to `FINE` - it is support detail, and emitting it at `WARNING` alongside the new line would restate one finding under two different names.
- `LSMTreeIndex.getUpgradeWarning()` returns it. Only the compacted sub-index is covered; checking the mutable pages means walking all of them, which is `checkIntegrity()`'s job for `CHECK DATABASE`.
- `TypeIndex.getUpgradeWarning()` scans every bucket sub-index rather than answering from the first. The existing shortcut ("every bucket shares one definition") holds for a warning derived from the definition, but a key-order mismatch is physical state one bucket can be in and another not.
- `LSMTreeFullTextIndex` delegates. It implements `IndexInternal` by composition, so it silently answered the interface default of `null` - and its keys are user text, where non-ASCII characters actually live. This is the `Paper[title,abstract]` half of the report.
- The channel's premise had to widen: it was built for an index that still answers correctly and is only missing what a newer layout buys (#5478). Studio's banner said so in as many words - "they keep working as they are - rebuilding is optional" - which would now sit directly above a message explaining that lookups return fewer records than a scan. That claim and the matching wording in the `IndexInternal` / `LocalSchema` contracts are corrected.

`LSMTreeGeoIndex` and `LSMSparseVectorIndex` also wrap an LSM index by composition. Neither should ever be exposed to this - geohash cells and dimension identifiers are ASCII, so no key either stores can be ordered differently by the signed/unsigned change - but they delegate anyway (geo returns its own layout advisory first and falls through; sparse-vector delegates outright). That exemption is a claim about the *keys*, not a property of those classes, and an override returning `null` unconditionally would hide a mismatch if the claim ever stopped holding.

## What an operator gets

One `WARNING` per affected logical index at open, naming the command that repairs it, and - independent of the log - the affected set as a query:

```sql
SELECT name, typeIndexName, upgradeWarning FROM schema:indexes WHERE upgradeWarning IS NOT NULL
```

`typeIndexName` is the name `REBUILD INDEX` takes. Studio needed no change to group the rows and render one ``REBUILD INDEX `Paper[title,abstract]` `` per logical index - it already did.

## Test plan

- [x] `Issue5802KeyOrderUpgradeWarningTest` (engine, 6 tests). The disorder is planted by swapping two pointers of a page's entry array - the physical state an index written under the old comparator ends up in, without needing the old comparator - the technique `LSMTreeIndexKeyOrderCheckTest` established. Each test compacts, disorders and reopens, because the check runs when the sub-index is wired up.
  - `aHealthyIndexCarriesNoUpgradeWarning` - the baseline that lets the others fail
  - `aDisorderedCompactedIndexIsReportedAsAnUpgradeWarning` - the verdict reaches `getUpgradeWarning()`
  - `theAffectedIndexesAreDiscoverableFromSchemaIndexes` - `schema:indexes` and `schema:index:<name>` expose it, with `typeIndexName` carrying the rebuildable name
  - `aTypeIndexReportsAWarningRaisedByANonFirstBucket` - the `TypeIndex` fix
  - `aDisorderedFullTextIndexReportsThroughTheCompositionWrapper` - the `LSMTreeFullTextIndex` delegate, reached through `getFileIds()` / `Schema.getFileById()` so no production accessor was added for the test
  - `rebuildingTheIndexClearsTheWarning` - `REBUILD INDEX` is a real remedy and the advisory does not survive it
- [x] Four of the original five fail before the fix; the healthy baseline passes throughout. The full-text test was separately proven to fail with the delegate stubbed back to `null`.
- [ ] **Untested:** the "could not be verified" branch. Firing it needs a genuine I/O failure - the layer below already turns an unreadable key into a reported problem rather than an exception - so staging it would need mocks or page corruption brittle enough to test the fixture rather than the code.
- [x] `LSMTreeIndexKeyOrderCheckTest` unchanged and still green - it asserts on `checkRootPagesKeyOrder`, `checkIntegrity()` and `CHECK DATABASE`, none of which this touches.
- [x] `mvn -pl engine -am -Dtest='com.arcadedb.index.**.*Test,com.arcadedb.schema.**.*Test' test` - 1486 tests, 0 failures.
- [x] `npm test` in `studio` - 50/50.

### DIFF
--- a/docs/5802-lsm-key-order-upgrade-warning.md
+++ b/docs/5802-lsm-key-order-upgrade-warning.md
@@ -1,0 +1,156 @@
+# #5802 - a mis-ordered index after an in-place upgrade is only discoverable by grepping the startup log
+
+## Issue
+
+A production install running `26.7.2` had its `lib/*.jar` swapped for a build of current `main` and was
+restarted. Before any query ran, the log emitted one `WARNING` per affected index:
+
+```
+WARNI [LSMTreeIndexCompacted] Index 'Paper_0_84306331895885' is physically ordered differently
+than the current key comparator, so lookups can return fewer (or foreign) records than a scan:
+run 'REBUILD INDEX Paper_0_84306331895885'. Details: [page 20 entry 8: ...]
+```
+
+Both a `FULL_TEXT` index (`Paper[title,abstract]`) and a plain non-unique string index (`Author.name`)
+were reported. The reporter rolled back to `26.7.2` and asked two things:
+
+1. Is this an expected manual migration step, and should it be surfaced more prominently than a
+   `WARNING` log line?
+2. Is there any way to learn *which* indexes are affected without grepping the startup log?
+
+## Root cause of the underlying condition
+
+The report is accurate: the disorder is real, not a false positive from the new detector.
+
+`BinaryComparator.compareBytes(byte[], Binary)` - the routine the LSM binary search uses to place and to
+find a `STRING` key - compared bytes as **signed** until `#5321` (`02d5800ea`), while every other string
+comparison in the engine is unsigned. A UTF-8 lead byte is `>= 0x80`, so it is negative as a Java byte:
+under the old comparison `Á` sorted *before* every ASCII letter, under the new one it sorts after.
+
+That is why the reporter's `Author.name` evidence starts with an accented name and why several of the
+reported pairs are pure ASCII. The mutable index pages were physically laid out in signed order. A
+compaction then reads those pages with cursors that assume ascending order and merges them with
+`LSMTreeIndexAbstract.compareKeys` (object comparison, always unsigned), so once the two orders diverge
+at one accented key the merge emits a stretch of keys out of order - ASCII neighbours included - and
+writes them to the compacted pages as-is. `#5321` did not create the disorder; it made the reader
+disagree with what an older build had already written.
+
+The remedy is therefore correctly `REBUILD INDEX`, and it cannot be applied automatically: rebuilding
+every index of a multi-million-record database on the open path would turn a restart into an outage.
+
+## What was actually wrong
+
+The reporting, not the detection. `LSMTreeIndexCompacted.checkKeyOrderOnLoad()` (added in `8489de7fc`)
+was a dead end for an operator:
+
+- It named the **physical bucket sub-index** (`Paper_0_84306331895885`), the name a user never sees
+  anywhere else. The logical index they would act on is `Paper[title,abstract]`.
+- It existed only as a log line. There was no way to ask the database which indexes were affected, so a
+  restart whose logs had rotated away left no trace at all.
+- `CHECK DATABASE` does report it, but it walks **every page of every index** - not something to run on
+  the box the reporter was trying to bring back up.
+
+Meanwhile the engine already had exactly the right channel for "this index should be rebuilt, here is
+why": `IndexInternal.getUpgradeWarning()`. `LocalSchema.reportUpgradeWarning` logs it once per database
+open per **logical** index with the `REBUILD INDEX` command to run, and it is exposed as `upgradeWarning`
+on `schema:indexes` and `schema:index:<name>`, which is what Studio renders. The key-order check simply
+was not wired into it.
+
+## Fix
+
+`engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCompacted.java`
+
+- `checkKeyOrderOnLoad()` records its verdict in a `volatile String keyOrderMismatch` instead of only
+  logging it, and exposes it through `getKeyOrderMismatch()`. Caching is what lets the reader honour the
+  `getUpgradeWarning()` contract of being cheap and side-effect free - it is called per index on every
+  schema listing and must not touch a page to answer.
+- The remaining log line keeps the physical evidence (page and entry numbers of *this* sub-index) but
+  drops to `FINE`. That detail is support material; the operator-facing warning is now the one below,
+  which names something they can act on. Emitting both at `WARNING` would restate the same finding under
+  two different names, which is the confusion the issue is about.
+
+`engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndex.java`
+
+- `getUpgradeWarning()` returns the sub-index verdict. Only the compacted sub-index is covered: checking
+  the mutable pages means walking all of them, which is `checkIntegrity()`'s job for `CHECK DATABASE` and
+  not something the open path can afford.
+
+`engine/src/main/java/com/arcadedb/index/TypeIndex.java`
+
+- `getUpgradeWarning()` scans every bucket sub-index instead of answering from the first one. The
+  existing comment - "every bucket sub-index shares one definition, so the first one answers for all of
+  them" - holds for a warning derived from the *definition* (the geospatial cell layout of `#5478`), but
+  a key-order mismatch is physical state that one bucket can be in and another not. Asking only bucket 0
+  would report a type index healthy while one of its buckets needed a rebuild.
+
+`engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java`
+
+- Delegates `getUpgradeWarning()` to the LSM index it stores its terms in. It implements `IndexInternal`
+  by composition rather than inheritance, so without this it silently answered the interface default of
+  `null` - and a full-text index is the one whose keys are user text, where non-ASCII characters actually
+  live. This is the `Paper[title,abstract]` half of the report.
+
+### The channel's premise had to widen
+
+`getUpgradeWarning()` was built for one shape of advice: an index on an older layout that still answers
+correctly and is only missing what the new layout buys (the geospatial cell layout of `#5478`). Its
+contract said so, `LocalSchema.reportUpgradeWarning` repeated it, and Studio's banner told the reader in
+so many words that the flagged indexes "keep working as they are - rebuilding is optional".
+
+A key-order mismatch is the other shape: the index does **not** answer correctly. Reusing the channel
+without saying so would have put a reassurance directly above a message explaining that lookups return
+fewer records than a scan.
+
+- `engine/src/main/java/com/arcadedb/index/IndexInternal.java` and
+  `engine/src/main/java/com/arcadedb/schema/LocalSchema.java`: the contract now names both shapes and
+  asks an implementation to make clear which one it is reporting.
+- `studio/src/main/resources/static/js/studio-database.js`: the banner drops the blanket "they keep
+  working as they are - rebuilding is optional" and just says a rebuild can be run at any time. Each
+  message underneath already states what its own condition costs. No other Studio change was needed - it
+  already groups the flagged rows by `typeIndexName` and renders one
+  ``REBUILD INDEX `Paper[title,abstract]` `` per logical index.
+
+### Not changed
+
+`LSMTreeGeoIndex` and `LSMSparseVectorIndex` also wrap an LSM index by composition. Neither is exposed to
+this: geospatial keys are ASCII geohash cells and sparse-vector keys are dimension identifiers, so no key
+either one stores can be ordered differently by the signed/unsigned change. `LSMTreeGeoIndex` additionally
+already overrides `getUpgradeWarning()` for its own layout advisory.
+
+## What an operator now gets
+
+One `WARNING` per affected logical index at open, naming the command that repairs it:
+
+```
+Index 'Paper[title,abstract]' of database 'papers' should be rebuilt: its pages are physically sorted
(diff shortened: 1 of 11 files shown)


robfrank:
Thanks - all three points landed, one of them further than suggested.

**1. Compacted-only scope.** Confirmed: `getUpgradeWarning()` reads only `mutable.getSubIndex()`, so an index never compacted can be mis-ordered and answer `null`. Made explicit in the javadoc, the tracking doc and the operator narrative: the query is "the affected indexes we can name cheaply", not a clearance, and CHECK DATABASE stays the exhaustive answer. Deliberately not widened - walking every mutable page at open is the cost this design exists to avoid.

**2. Silent loss on a throwing check.** Confirmed: `keyOrderMismatch` is only assigned inside the success branch, so an exception leaves it null. Raised to WARNING. This mattered less when the verdict fed only a log line; now that it backs a query, "could not check" must not read the same as "clean". The new message says the index is not covered by the schema:indexes advice and points at CHECK DATABASE.

**3. Geo / sparse-vector.** Went further than the note. Rather than documenting the ASCII invariant harder, both now delegate: `LSMTreeGeoIndex` returns its own layout advisory first and falls through to the underlying index, `LSMSparseVectorIndex` delegates outright. You were right that geo structurally could not report a mismatch - and the exemption was a claim about the *keys*, not a property of those classes. A silent wrong answer is the exact failure mode this issue is about, so it should not depend on an invariant holding.

**One gap I am not closing, flagged rather than skipped:** the point-2 change has no test. Firing that catch block needs a page read to fail, which I cannot stage without mocks (barred here) or page-internal corruption brittle enough that it would test the fixture rather than the code. It is a log-level change on a pre-existing catch, so I judged the contortion worse than the gap - but it is an untested line.

Verification: `Issue5802KeyOrderUpgradeWarningTest` 5, `LSMTreeIndexKeyOrderCheckTest` 4, `LSMTreeGeoIndexTokenizationTest` 16, `com.arcadedb.index.sparsevector` 112 - 137 tests, 0 failures.

robfrank:
Cycle 2 - all five points addressed.

**1. The unknown state was not queryable.** You caught an incomplete fix of mine: last cycle I raised the failed check to WARNING but left `keyOrderMismatch` null, so `schema:indexes` still said healthy. The complaint was queryability, not log level. There is now a distinct `KEY_ORDER_UNVERIFIABLE_WARNING` published through the same field, so 'checked and good' and 'never checked' are no longer the same answer. Rebuilding is a valid response to unverifiable - it makes the order correct by construction - so the remedy the advisory names still applies.

**2. PR body contradicted the diff.** Correct, I updated the tracking doc and not the PR body. Fixed.

**3. Lock-freedom.** Agreed it deserved saying next to a `checkIntegrity()` that explains at length why it needs the lock. Javadoc now records that it reads one volatile field of whichever sub-index is published, and that a compaction swapping in a fresh one is harmless because each instance's verdict is true of the file it was computed from.

**4. `keyOrderCheckedOnLoad` set before the check.** Intended, now documented: retrying would re-read pages behind a getter every schema listing calls per index, and the failure is published as a verdict rather than lost.

**5. Full-text delegate untested.** The best catch here - it was described as half the report and every test used a plain string index. `aDisorderedFullTextIndexReportsThroughTheCompositionWrapper` covers it, reaching the compacted component through `getFileIds()` / `Schema.getFileById()` so composition is exercised without adding a production accessor for the test's benefit. I stubbed the delegate back to `null` and confirmed only that test fails, so it is armed rather than merely passing.

**Still untested, and I would rather say so than bury it:** the 'could not be verified' branch. `checkKeyOrderInPage` already catches an unreadable key and reports it as a problem, so what remains is genuine I/O failure - staging it needs mocks or page corruption brittle enough to be testing the fixture. Noted in the tracking doc and the PR body.

Verification: `Issue5802KeyOrderUpgradeWarningTest` 6, `LSMTreeIndexKeyOrderCheckTest` 4, `LSMTreeGeoIndexTokenizationTest` 16, `com.arcadedb.index.fulltext` 176 - 202 tests, 0 failures.

robfrank:
Cycle 3 - two fixed, one declined with reasoning.

**1. Transient I/O on a replica could latch false advice.** You were right, and it was a defect in *my* cycle-2 change rather than in the original code. Confirmed the mechanism: `onAfterLoad` resolves the sub-index through `Schema.getFileById()`, which returns the **same** component instance on every load, and `keyOrderCheckedOnLoad` was set *before* the check ran - so one transient page-read failure on a replica's Raft apply or snapshot resync would have pinned 'needs rebuild' on a healthy index for the lifetime of the database.

Now the flag is set only when the check **completed**. A run that threw is retried on the next load, and a completed run publishes the definitive verdict *including the clean one*, which is what clears a stale unverifiable. The retry costs the bounded root-page read again and only in the already-failing case. Good catch - this was the load-bearing assumption and it did not hold.

**2. `subIndex` is not volatile.** Confirmed, `LSMTreeIndexMutable:54` is a plain field, and the javadoc's 'reads one volatile field' conflated two reads. Reworded: the reference read is a benign data race yielding either instance or transiently null, and only the verdict is volatile. That every outcome is sound *is* the argument for skipping the lock, so it should be stated accurately rather than hand-waved through 'volatile'.

**3. `TypeIndex` first-non-null could surface UNVERIFIABLE over MISMATCH - declining, with reasoning.** Ranking them needs a severity concept on `getUpgradeWarning()`, a general advisory API where nothing else wants one, and both messages name the identical remedy. The information also is not lost: `FetchFromSchemaIndexesStep` emits a row per **bucket** sub-index, each calling its own `getUpgradeWarning()`, and Studio groups rows by message - so `schema:indexes` shows *both* messages with their affected names. Only the single aggregated `TypeIndex` answer picks one, which is the coarse view by design. Happy to add the ranking if you still think it earns its keep, but it looked disproportionate for reordering a value that already appears in full one row down.

**4. Test seam for the unverifiable branch.** Still declining a test double - but note fix 1 changes the risk profile rather than just documenting it: an unverifiable verdict is now self-healing instead of permanent, so the untested branch's worst case is a WARNING and an advisory that clears itself on the next successful load.

Verification: `Issue5802KeyOrderUpgradeWarningTest` 6, `LSMTreeIndexKeyOrderCheckTest` 4, `LSMTreeGeoIndexTokenizationTest` 16, `com.arcadedb.index.fulltext` 176 - 202 tests, 0 failures.

Issue #5746, opened 2026-08-02
openCypher multi-hop patterns return zero rows when planned through GAVFusedChain; SQL MATCH correct on the same data


tae898:
With a Graph Analytical View active, an openCypher multi-hop pattern returns
**nothing**. SQL `MATCH` over the same data and the same traversal returns the
correct answer, and the same Cypher query is correct before the view exists.
There is no error and no warning: the caller gets an empty result set that is
indistinguishable from a legitimate "no matches".

Version: `26.8.1-SNAPSHOT` (engine jar from our 2026-08-01 build of `main`).

## Reproduction

Self-contained, engine API only, no bindings. Six vertices, six edges, the
default (bidirectional) edge type.

```
0 -> 1, 0 -> 2, 1 -> 3, 2 -> 4, 1 -> 5, 3 -> 0
```

2-hop out from vertex 0 reaches `{3, 4, 5}` over three paths.

```java
import com.arcadedb.database.Database;
import com.arcadedb.database.DatabaseFactory;
import com.arcadedb.query.sql.executor.Result;
import com.arcadedb.query.sql.executor.ResultSet;

import java.io.File;
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.ArrayList;
import java.util.Comparator;
import java.util.List;

/**
 * Reproduces: with a Graph Analytical View active, an openCypher multi-hop
 * pattern emits the right NUMBER of rows but leaves every variable bound by the
 * fused chain unset, so projections come back empty.
 *
 * Pure engine API, no bindings. Run with:
 *   java --class-path "<arcadedb jars>/*" GavFusedChainRepro.java
 */
public class GavFusedChainRepro {

  static final int[][] EDGES = {{0, 1}, {0, 2}, {1, 3}, {2, 4}, {1, 5}, {3, 0}};
  static final String P =
      "MATCH (p:Person {id: 0})-[:KNOWS]->(a:Person)-[:KNOWS]->(b:Person) ";
  static final String SQL2 =
      "MATCH {type: Person, as: p, where: (id = 0)}"
          + ".out('KNOWS'){as: a}.out('KNOWS'){as: b} RETURN DISTINCT b.id AS id";

  public static void main(String[] args) throws Exception {
    final Path dir = Files.createTempDirectory("gavfused");
    final String path = dir.resolve("db").toString();
    deleteRecursively(new File(path));

    Database db = new DatabaseFactory(path).create();
[...]

Issue #5459, opened 2026-07-27
`stdev()` and `stdevP()` return `0.0` for empty input instead of `NULL`


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT` (build `1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main`)

### Environment

* Linux x86_64, Docker, OpenJDK
* `arcadedata/arcadedb:latest`
* HTTP command API, Cypher language

### Description

`stdev()` and `stdevP()` return `0.0` when their input contains zero rows.

This makes an empty input set indistinguishable from a non-empty data set whose standard deviation is genuinely zero.

Related aggregates for which no numeric result can be derived from an empty input, including `avg()`, `min()`, `max()`, `percentileCont()`, and `percentileDisc()`, return `NULL` in ArcadeDB.

Neo4j fixed the same empty-input behavior for `stDev()` in version `2026.05.0`:

> Fixed issue where stDev aggregation would return 0 rather than null for empty input.

The scope of this report is strictly the zero-input-row case produced by `UNWIND []`.

### Steps to reproduce

No graph data is required.

```cypher
UNWIND [] AS x
RETURN stdev(x) AS std,
       stdevP(x) AS stdp;
```

ArcadeDB returns:

```json
{
  "std": 0.0,
  "stdp": 0.0
}
```

### Expected behavior

Both aggregates should return `NULL`:

```json
{
  "std": null,
  "stdp": null
}
```

An empty input set contains no observations from which either a sample or population standard deviation can be calculated.

This also preserves the semantic distinction between:

* no observations: `NULL`
* observations with no variation: `0.0`

### Actual behavior

ArcadeDB returns `0.0` for both aggregates:

```json
{
  "std": 0.0,
  "stdp": 0.0
}
```

The empty-input result is therefore indistinguishable from a valid non-empty zero-variance result.

#### Other aggregates over empty input

```cypher
UNWIND [] AS x
RETURN avg(x) AS avgValue,
       min(x) AS minValue,
       max(x) AS maxValue,
       percentileCont(x, 0.5) AS percentileContinuous,
       percentileDisc(x, 0.5) AS percentileDiscrete;
```

ArcadeDB returns `NULL` for these aggregates.

#### Non-empty zero-variance input

```cypher
UNWIND [5, 5] AS x
RETURN stdev(x) AS std,
       stdevP(x) AS stdp;
```

[...]

PR #5829, opened 2026-08-05
test(gremlin): defect hunt in com.arcadedb.gremlin via plan-shape and differential testing

### LINKED ISSUE #5843: Increase test coverage for the gremlin module
## Summary

`com.arcadedb.gremlin` sits at **70.2% line coverage (1225/1744) and 57.8% branch coverage (499/864)** despite 1866 integration tests passing against it.

The aggregate hides a split. The full TinkerPop `structure` and `process` conformance suites run with zero `@OptOut`s, so the TinkerPop contract surface is exercised heavily. Everything ArcadeDB-specific, which no TinkerPop suite knows about, is thin or absent:

| Class | Lines | Branches |
| --- | --- | --- |
| `ArcadeGAVFusedStep` | 0/56 | 0/31 |
| `ArcadeGAVVertexStep` | 0/30 | 0/12 |
| `ArcadeEdgeCountFilterStep` | 0/28 | 0/13 |
| `ArcadeIoRegistry` | 10/24 | 0/11 |
| `ArcadeFilterByIndexStep` | 29/64 | 12/42 |
| `ArcadeTraversalStrategy` | 109/154 | 83/126 |

## The failure mode this exists to fix

`GremlinGAVTest` shipped 8 tests named `outWithGAV`, `multiHopWithGAV` and similar. All 8 passed. **None executed the GAV code path.** The fixture used the public `new GraphAnalyticalView(db)` constructor, which never calls `registerAsTraversalProvider()`, so `findProvider()` returned null on every call and each traversal silently took the OLTP fallback. The tests passed because they asserted only on *results*, and the fallback returns identical results.

Result-only assertions cannot distinguish "the optimization ran and was correct" from "the optimization never ran". Any test for an optimization must first prove the optimization is present.

## Approach

Two test-scope helpers, no production code changes:

- **`TraversalPlans`** applies the strategy set and exposes the compiled step list, so a test can assert *which* steps the optimizer installed before asserting behavior.
- **`DifferentialTraversal`** runs one query twice against the same database, once normally and once with `ArcadeTraversalStrategy` removed, and asserts the results agree. The two paths are implementations of one specification, so disagreement is a defect by construction.

Both ship with self-tests proven able to fail by deliberate mutation.

## Outcome

Implemented in #5829. Coverage 70.2% -> 79.2% lines, 57.8% -> 66.8% branches, no production code touched, suite green at 1866 tests throughout.

Five defects found, four of which are open and tracked separately:

- #5838 - `parse()` over-widens `OperationType`, causing MCP to deny legitimate inserts
- #5840 - GAV/CSR acceleration never engages for labeled traversals via the string entry point
- #5841 - `ArcadeEdgeCountFilterStep` installs non-deterministically across JVMs
- #5842 - `setTimeout()` is a no-op and its backing field is static

The fifth, the vacuous `GremlinGAVTest` fixture above, is fixed in #5829 with a guard test asserting plan shape so it cannot silently regress.

A sixth item worth a decision: the public `GraphAnalyticalView(Database)` constructor is documented as a backward-compatibility shim, but a view built through it can never accelerate anything. Deprecating it or making it register would remove the trap that caused this.

## Known limitation

`DifferentialTraversal` is blind to *over-inclusive* index-cursor bugs where both paths agree. `ArcadeTraversalStrategy` removes only the `~label` container, so the surviving value container re-filters after `ArcadeFilterByIndexStep` runs. Against a boundary fixture an inverted `gte`/`lte` flag is caught; an inverted `gt`/`lt` flag is not. Closing that needs a cursor-level oracle rather than a result-level one.

robfrank:
Closes #5843

## Summary

A defect hunt in `com.arcadedb.gremlin`, targeting the ArcadeDB-specific code that the TinkerPop conformance suites never exercise. All 12 planned tasks are complete: ~90 tests across 9 new test classes plus two test-scope helpers, **no production code changed**, 5 defects found (1 fixed, 4 open), coverage 70.2% -> 79.2% lines.

The spec and plan are committed under `docs/superpowers/`. Every task was independently reviewed, and a final whole-branch review is reflected below.

## Why

`com.arcadedb.gremlin` sits at **70.2% lines (1225/1744), 57.8% branches (499/864)** across the existing 1866 integration tests plus unit tests, all passing.

That aggregate hides a split. The full TinkerPop `structure` and `process` suites run with zero `@OptOut`s, so the TinkerPop contract surface is exercised heavily. Everything ArcadeDB-specific is thin or absent:

| Class | Lines | Branches |
| --- | --- | --- |
| `ArcadeGAVFusedStep` | 0/56 | 0/31 |
| `ArcadeGAVVertexStep` | 0/30 | 0/12 |
| `ArcadeEdgeCountFilterStep` | 0/28 | 0/13 |
| `ArcadeIoRegistry` | 10/24 | 0/11 |
| `ArcadeFilterByIndexStep` | 29/64 | 12/42 |
| `ArcadeTraversalStrategy` | 109/154 | 83/126 |

## Finding 1: `GremlinGAVTest`'s 8 tests never executed the GAV path

`GremlinGAVTest` has 8 tests named `outWithGAV`, `multiHopWithGAV`, and so on. All 8 passed. **None of them ran the GAV code.**

`GraphAnalyticalView.registerAsTraversalProvider()` is package-private with exactly one caller, `GraphAnalyticalViewBuilder.createView()`. The fixture used the public `new GraphAnalyticalView(db)` constructor, documented as a backward-compatibility shim, which never registers. So `GraphTraversalProviderRegistry.findProvider()` returned null on every call and every traversal silently took the OLTP fallback. The tests passed because they assert only on **results**, and the fallback returns identical results.

JaCoCo confirmed it: `ArcadeTraversalStrategy.java:205` (`if (provider != null)`) covered, lines 206-209 (the body installing `ArcadeGAVVertexStep`) MISSED.

Fixed by building through the builder, and a guard test now asserts the step is present in the compiled plan so the registration cannot silently regress.

**Important qualification, and it interacts with Finding 2.** The fix makes the *view* register correctly, but it does not make all 8 tests exercise the CSR path. Seven of them use labeled hops (`out('KNOWS')`, `in('KNOWS')`, `both('KNOWS')`, `out('LIKES')`) through `graph.gremlin(String)` - the row in Finding 2's table that reads *does not engage*. Only `outWithNoEdgeLabelWithGAV` actually reaches `ArcadeGAVVertexStep`. The guard test uses `graph.traversal()`, a different entry point from the 8 it sits beside, so it pins registration rather than proving those 8 are no longer vacuous. Most of them still are, for the separate reason documented in Finding 2.

Worth noting the public `GraphAnalyticalView(Database)` constructor is a usability trap in its own right: a view built through it can never accelerate anything. Suggest deprecating it or making it register.

## Finding 2: GAV/CSR acceleration is dead for *labeled* traversals on the real execution path

This is the significant one. **A Gremlin query that names an edge label gets no CSR acceleration at all** when submitted as a string, which is the entry point the HTTP endpoint, Studio, and the drivers all use.

Measured per-test, each run with `jacoco.exec` deleted first (JaCoCo appends by default, which silently contaminates a naive measurement):

| Query shape via `graph.gremlin(String)` | `ArcadeGAVVertexStep` | `ArcadeGAVFusedStep` |
| --- | --- | --- |
| `out()` label-less | 23/30 lines | 0/56 |
| `out('KNOWS')` / `in('KNOWS')` / `both('KNOWS')` / `out('LIKES')` | **0/30** | 0/56 |
| label-less `out()` feeding `.count()` | **0/30** | 0/56 |
| `out().out()` label-less multi-hop | - | **engages** (no committed test covers this shape) |

### Root cause

TinkerPop 3.8.1's gremlin-lang parser builds a `VertexStepPlaceholder` (which does **not** extend `VertexStep`) for any hop with a non-empty label array. `GValueReductionStrategy`, which resolves placeholders into real `VertexStep` instances, runs **after** `ArcadeTraversalStrategy` in the registered strategy order. `ArcadeTraversalStrategy.applyGAVOptimization` tests `step instanceof VertexStep` (line 200), so any placeholder still present at ArcadeDB's turn is simply skipped.

That single ordering issue explains every row above:
- label-less hops parse to a real `VertexStep`, so they convert;
- labeled hops parse to a placeholder, so they do not;
- a label-less hop feeding `.count()` is rebuilt into a *fresh* placeholder by TinkerPop's `AdjacentToIncidentStrategy`, so it also misses.

### Why it matters

Real Gremlin is overwhelmingly written with edge labels. `g.V().out('KNOWS')` is the normal case and `g.V().out()` the exception, so in practice the CSR acceleration is off for essentially all production traffic through the string entry point. Results stay correct via the OLTP fallback, so this is a silent performance defect, not a correctness one, and nothing in the existing test suite could have caught it: every one of those tests asserted only on results.

Verified against the shipped `gremlin-core-3.8.1` jar with `javap`, independently by two reviewers.

Context: this is the Gremlin-side analogue of the Cypher `GAVFusedChainOperator`, which carried two silent wrong-result defects in #5746 / #5827.

## Finding 3: `ArcadeEdgeCountFilterStep` installs non-deterministically across JVM runs

The `where(outE('X').count().is(...))` rewrite either installs or silently does not, **for byte-identical queries, depending on which JVM the query runs in**. Confirmed by repeated contradictory runs with no code change between them, and root-caused independently at the source level.

`ArcadeTraversalStrategy.applyPrior()` (line 305) declares an ordering relationship only with `InlineFilterStrategy`. It declares nothing about TinkerPop's `CountStrategy`. For a bounded predicate such as `is(gt(1))`, `CountStrategy` computes `highRangeCandidate = 2` and inserts a `RangeGlobalStep(0,2)` before the `CountGlobalStep`. That grows the `where`-child from 3 sub-steps to 4, and `applyEdgeCountFilterOptimization` matches only an exact 3-sub-step shape, so the rewrite silently declines.

Which strategy wins is genuinely arbitrary. `TraversalStrategies.sortStrategies()` topologically sorts strategy *classes*, and with no declared edge between these two the tie-break falls through to iteration order of a `HashSet<Class<?>>`. `Class` does not override `hashCode()`, so that order follows the JVM's identity hash codes, which are not stable across JVM invocations. `ArcadeGraph` computes and registers the strategy set exactly once in a `static` block (lines 100-107), so whichever order a given JVM happens to produce is frozen for that JVM's entire life.

The practical consequence is that an O(1) degree-check optimization is present or absent at process granularity, with no way to predict or control which. One committed test is `@Disabled` pinning the observed failure; a companion test asserts the `CountStrategy` interference explicitly so it stays meaningful under either outcome.

The fix is presumably to declare the missing ordering relationship, but that is a production change and out of scope for this PR.

## Finding 4: `parse()` over-widens `OperationType`, and MCP rejects legitimate operations today

`addV`, `addE` and `property()` compile to `AddVertexStartStepPlaceholder`, `AddEdgeStepPlaceholder` and `AddPropertyStepPlaceholder` (confirmed by reflection dump; `drop()` resolves to a real `DropStep`, as no placeholder variant exists in 3.8.1). Those defeat `ArcadeGremlin.parse()`'s specific `instanceof` chain, so `getOperationTypes()` falls into the unknown-mutating-step catch-all and returns `[CREATE, UPDATE, DELETE]` instead of the single correct type.

**HA routing is NOT affected.** The placeholders do implement `Mutating` transitively via `Writing`/`Deleting` (verified with `javap`), so `isIdempotent()` correctly returns `false`, and `RaftReplicatedDatabase` routes only on `isIdempotent()`/`isDDL()` and never reads `getOperationTypes()`. An earlier hypothesis that writes could be misrouted to followers was investigated and disproven.

**MCP is affected, today.** `ExecuteCommandTool.execute()` calls `engine.analyze(command)`; `GremlinQueryEngine.analyze()` is a pass-through to `parse()`; and `checkPermission()` throws `SecurityException` if *any* type in the set lacks its permission bit. So an MCP profile with `allowInsert=true, allowUpdate=false` rejects a legitimate `g.addV('Person')` with "Update operations are not allowed". Fail-closed over-denial rather than privilege escalation, but a real availability defect in a shipping module. Verified independently twice.

Three `@Disabled` tests pin the exact expected sets. Note the brief's own `.contains()`-based tests pass either way and carry inline pointers to their stricter disabled siblings.

## Finding 5: `ArcadeGremlin.setTimeout()` is a no-op, and its field is static

`ArcadeGremlin.timeout` is declared `private static Long` yet assigned by the *instance* method `setTimeout(long, TimeUnit)`, so a value set on one graph is visible to every `ArcadeGremlin` in the process.

The larger point, found while verifying the leak: **nothing reads the field.** A repo-wide search finds only `getTimeout()` and `setTimeout()` themselves touching it, and `executeStatement()` never applies it. No Gremlin query has ever been timed out by this setting. The static-versus-instance bug is real but currently unreachable; the shipped behavior is that `setTimeout()` does nothing at all. Fixing the field without also wiring it into execution would leave the API still inert.

Pinned by `characterizesTheProcessWideTimeoutLeak`, which asserts the current behavior deliberately and carries a Javadoc saying so and instructing that it be inverted when the field is made non-static. That naming is deliberate: this repo has been burned before by tests that silently encoded the bug they should have caught (#5459).

## The harness

Two test-scope helpers, no production code changes:

- **`TraversalPlans`** applies the strategy set and exposes the compiled step list, so a test can assert *which steps the optimizer installed* before asserting behavior. This is the direct antidote to Finding 1.
- **`DifferentialTraversal`** runs one query twice against the same database, once normally and once with `ArcadeTraversalStrategy` removed, and asserts the results agree. The two paths are implementations of one specification, so disagreement is a defect by construction, with no hand-authored oracle to get wrong.

Both ship with self-tests that were **proven able to fail** by deliberately breaking them and observing the failure, not by inspection. That mutation proof covers the two helpers specifically; where a later test's reachability rests on control-flow reading rather than an observed red, it is called out as such rather than implied.

**A structural limit worth stating.** Differential comparison is blind to *over-inclusive* index-cursor bugs. `ArcadeTraversalStrategy` removes only the `~label` container, so the value container survives on the `HasStep` and re-filters after `ArcadeFilterByIndexStep` runs. Against the boundary fixture that means an inverted flag on `gte` or `lte` is caught, but an inverted flag on `gt` or `lt` produces identical rows and is not. Two of the four range predicates are genuinely guarded; the other two need a plan-level or cursor-level oracle.

## A build gotcha worth knowing repo-wide

`gremlin-it` runs the gremlin module's tests from the **installed test-jar in `~/.m2`** via `dependenciesToScan`. `mvn -pl gremlin-it test -Dtest=X` on its own silently runs the last-installed code and ignores your working tree.

This was caught the hard way: `TraversalPlans.hasStepOfType` deliberately broken to `return true` still reported **3/3 passing**. Only after `mvn -pl gremlin install -DskipTests` did the self-test correctly fail.

Every verification command in the plan now carries the install prefix. `-pl gremlin-it -am` is not a substitute; it builds `arcadedb-engine`, where `-Dtest=ClassName` matches nothing and the build fails.

## Verification

```bash
./mvnw -pl gremlin install -DskipTests -q
./mvnw -pl gremlin-it -DskipITs=false verify
```

Full suite green at every commit: **1866 tests, 0 failures, 0 errors, 488 skipped**.

## Coverage

`com.arcadedb.gremlin`, measured by merging `gremlin-it`'s unit and IT exec files against `gremlin/target/classes`:

| Metric | Before | After |
| --- | --- | --- |
| Lines | 1225/1744 (70.2%) | 1382/1744 (79.2%) |
| Branches | 499/864 (57.8%) | 577/864 (66.8%) |

Biggest movers: `ArcadeGAVFusedStep` 0% → 55.4%, `ArcadeGAVVertexStep` 0% → 86.7%, `ArcadeTraversalStrategy` → 96.1% lines / 87.3% branches, `ArcadeGraphFactory` → 100% lines. The denominator is unchanged because no production code was touched.

One row is not reproducible: `ArcadeEdgeCountFilterStep`'s coverage depends on which JVM the run lands in, per Finding 3. On a JVM where `CountStrategy` wins the ordering race, that class's differential tests compare the unoptimized plan against itself and its coverage drops toward zero.

Coverage was a reported side effect here, not the goal. The defects above are the deliverable.

## A note on this PR's own tests

Four tests written during this work turned out to be incapable of failing, and were caught in review rather than shipped. Recording them in full, because a suite that cannot fail is exactly the defect this PR opened by documenting in `GremlinGAVTest`, and it recurred four times during the work itself.

**1. Two-label suppression never reached the branch it named.** `hasLabel("Person", "Company")` builds a single `HasStep` wrapping ONE `P.within(["Person","Company"])` container, so `ArcadeTraversalStrategy`'s `totalLabels > 1` branch is never reached: `totalLabels` counts label *containers*, not the labels inside one container. The test still passed, for a narrower reason - it exercised the `totalLabels == 1 && typeNameToMatch == null` guard, since `typeNameToMatch` stays null when the predicate is `P.within` rather than `Compare.eq`. Fixed by adding `twoChainedLabelContainersSuppressTheRewrite`, which uses chained `.hasLabel("Person").hasLabel("Company")` (`TraversalHelper.addHasContainer` merges consecutive calls onto the same `HasStep`, producing two `Compare.eq` containers and genuinely driving `totalLabels` to 2).

**2. Indexed-plus-unindexed could not discriminate.** `has("age", 30).has("name", "P30")`: age 30 alone already narrows the fixture to exactly one row, so the trailing unindexed predicate never had to do anything. A residual `HasStep` silently dropped or broken would have left the test green. The original is kept as `aRedundantUnindexedPredicateOnAnAlreadyUniqueIndexedMatchStillAgrees` with a comment saying what it does and does not prove, plus a new sibling where `city="Rome"` yields `{P10, P20}` (asserted intermediately) before `name="P20"` narrows to one.

**3. Three `OperationType` tests used `.contains()`.** `.contains(OperationType.CREATE)` is satisfied by the widened `[CREATE, UPDATE, DELETE]` that the defect actually produces, so it cannot distinguish correct from buggy classification. Three `.containsExactly()` siblings were added; they fail, and are `@Disabled` pinning the real expectation.

**4. The factory close test asserted nothing.** `factory.close(); factory = null;` cannot fail regardless of what `close()` disposes, including nothing. The committed version captures the pooled instance's `BasicDatabase`, asserts `isOpen()` is still true while the instance merely sits released in the pool, then asserts `isOpen()` is false after `factory.close()`.

## Deliberately not covered

`ArcadeServiceRegistry`'s `Lambda*` inner classes. Their own Javadoc describes them as toy services for demonstration and testing; covering them would move the percentage without finding anything.

## Suggested follow-ups

- Deprecate or fix the public `GraphAnalyticalView(Database)` constructor so it cannot silently produce an unregistered view.
- Declare the missing strategy ordering between `ArcadeTraversalStrategy` and TinkerPop's `CountStrategy`.
- Make `ArcadeGremlin.timeout` an instance field and invert `characterizesTheProcessWideTimeoutLeak`.
- Resolve `VertexStepPlaceholder` before `ArcadeTraversalStrategy` runs, or match on the placeholder contracts, so labeled traversals get CSR acceleration.
- File the MCP over-denial separately; it is reachable without any of the above.


robfrank:
Fixed items 1 and 2 in `761dabb21`. Pushing back on 3, agreeing on 4, and one correction on the first nit.

**1. Static timeout pollution — correct, fixed.**

Verified: `private static Long timeout` defaults to `null`, so `originalTimeout` was null on the very path that mattered and the `if (originalTimeout != null)` guard skipped the restore entirely. `characterizesTheProcessWideTimeoutLeak` then left `1234` in place for the remainder of the fork.

Restoring now goes through reflection, because `setTimeout(long, TimeUnit)` takes a primitive and cannot express null. Proven by mutation rather than assumed: with a typo'd field name all 7 tests error on the restore path, confirming it executes on every test instead of silently no-opping. If the field is ever made non-static, the call throws and the class must be updated alongside it — failing loudly is intended.

That experiment surfaced a second problem not in the review: the restore ran *before* `graph.drop()`, so a throwing restore left `./target/test-gremlin-engine` on disk and failed the *next* run in `setup()` against a database that already had the `Person` type, masking the real cause behind a cascade. The drop is now in a `finally`.

**2. Fully-qualified `TimeUnit` — correct, fixed** in the same commit.

**3. Committed `docs/superpowers/` — pushing back, with the decision left to the maintainer.**

These are not the artifact that was purged. The 2026-08-03 cleanup removed `docs/NNNN-*.md` defect-tracking documents; the standing convention is that superpowers plans and specs are committed and travel with their feature PR. The distinction matters here in a concrete way: the plan file is where the four "tests that could not fail" corrections are recorded inline at the snippets that produced them, so the next plan built from this template does not inherit them. Deleting it would drop that.

The defect *findings* deliberately live in the PR body, not in the repo, exactly per the convention you cite. Whether the process artifacts belong permanently in the tree is a maintainer call, and I have not assumed it either way.

**4. Real issues for the four open defects — agreed, and deferred to the maintainer.**

Filing issues is outward-facing, so I have not created them. Flagging the same priority you did: the over-widened `OperationType` → MCP over-denial is independent of HA, reachable today, and the most user-visible of the four. The other three are the labeled-hop CSR gap, the per-JVM non-deterministic `ArcadeEdgeCountFilterStep` install, and the static-timeout defect.

**Nit — shared `./target/test-*` paths: partly a misreading.**

All 10 database paths across the new classes are unique; each appears exactly once (`test-traversalplans`, `test-differential`, `test-gav-steps`, `test-filterbytype`, `test-filterbyindex`, `test-edgecountfilter`, `test-gremlin-engine`, `test-gremlin-analyze`, `test-ioregistry`, `test-graphfactory-pool`), and none collides with the pre-existing `./target/testgremlin*` users. There is no cross-class collision today. The valid half of the point is that they are *hardcoded*, which would break under parallel in-fork execution — agreed that is not worth changing now.

**Nit — `@Tag("slow")`: considered and deliberately omitted.** The new classes run in roughly 0.1–1.0s each against fixtures of a handful of vertices.

Full suite after the fixes: 1866 tests, 0 failures, 0 errors, 488 skipped, BUILD SUCCESS.

robfrank:
All four points actioned. Suite green after the changes: 1866 tests, 0 failures, 0 errors, 488 skipped.

**1. `docs/superpowers/` removed** (`e2c12869c`).

I pushed back on this in the previous round and the maintainer overruled me, correctly. At ~2000 lines the plan and spec were more than half the diff, and the PR body is the better durable home for the content that mattered.

Before deleting, I moved the four "tests that could not fail" corrections out of the plan and into the PR body in full detail, since those were the one thing the plan carried that the body did not:

- why `hasLabel("Person", "Company")` never reaches the `totalLabels > 1` branch (`totalLabels` counts label *containers*, and the varargs form builds one `P.within` container), and what the chained replacement does instead
- why `has("age", 30).has("name", "P30")` could not discriminate, and the `city="Rome"` -> `{P10, P20}` fixture that replaced it
- why `.contains(CREATE)` is satisfied by the widened `[CREATE, UPDATE, DELETE]` the defect produces
- what the factory close test asserts now (`isOpen()` true while merely released, false after `factory.close()`)

The diff is now 1756 insertions of pure test code.

**2. Four issues filed and linked into the tests that pin them** (`afb0f8ed4`).

- #5838 - `parse()` over-widens `OperationType`, causing MCP to deny legitimate inserts. Reachable today, independent of the placeholder work, as you noted.
- #5840 - GAV/CSR acceleration never engages for labeled traversals via the string entry point.
- #5841 - `ArcadeEdgeCountFilterStep` installs non-deterministically across JVMs (`CountStrategy` ordering).
- #5842 - `setTimeout()` is a no-op and its field is static.

Each `@Disabled` reason and the timeout characterization Javadoc now cite their issue number, so a future cleanup cannot delete a disabled test as the only remaining trace of the defect. #5842 also records the finding that *nothing reads the field*, which makes `setTimeout` inert regardless of the static-versus-instance question - fixing only the field would leave the API still doing nothing.

**3. Reflection coupling** - no change, matching your assessment. Both sites fail loudly rather than silently: `restoreProcessWideTimeout` throws `IllegalStateException` if the field is renamed or made non-static, which I verified by mutation (typo the field name and all 7 tests error on that path rather than passing quietly).

**4. Em dash removed** (`619d2f5a3`). Confirmed pre-existing on `main` as you said, and scanned the rest of the gremlin test sources: none remain.

**On the verification note.** Your read of the structural limitation matches mine, and it is worth being precise about which half is guarded: against the 10/20/30/40/50 fixture, an inverted `gte` or `lte` flag drops row P30 and is caught, while an inverted `gt` or `lt` flag is not, because `ArcadeTraversalStrategy` removes only the `~label` container and the surviving value container re-filters after `ArcadeFilterByIndexStep` runs. Closing that would need a cursor-level oracle rather than a result-level one.

Issue #5656, opened 2026-07-31
[Cypher] follow-ups to #5626: the subquery body is still run as standalone text


lvca:
Three follow-ups recorded while closing #5626 (PR #5642), kept together because they are three consequences of one design decision rather than three unrelated defects.

`EXISTS { }`, `COUNT { }` and `COLLECT { }` do not hold their body as part of the query. They hold it as a **string**, hand it to `database.query("opencypher", ...)` as a standalone statement once per outer row, and absorb any failure into the expression's neutral value. Everything below follows from that: validation does not reach a body because for most of the pipeline the body is not there; the body is re-parsed per row because the text it runs differs per row; and a failure cannot propagate because the expression has already promised a boolean, a number or a list.

#5626 addressed the first consequence for one validation phase, and in doing so gave the three expressions a parsed body (`getParsedSubquery()`) plus a traversal that descends into it (`CypherExpressionWalker` with `Visitor.forNestedStatement`). That machinery is what makes the rest tractable.

Each section stands on its own and can be split back into its own issue if someone picks one up in isolation. Ordered by what I would do first.

---

## 1. Ten of the twelve validation phases still stop at a subquery body

`CypherSemanticValidator.validate()` runs twelve phases. Two reach inside a body: `validateVariableScope` (since #5213) and `validateFunctionArgumentTypes` (since #5626). The other ten stop at the boundary, so a check that rejects a query written one way accepts the identical mistake written one level in.

Verified against the merged code, using `EXPLAIN` so the query is parsed and planned but never executed:

```
REJECTED[NestedAggregation]        EXPLAIN MATCH (n:P) RETURN count(count(n)) AS r
ACCEPTED                           EXPLAIN MATCH (n:P) RETURN COUNT { MATCH (m:P) RETURN count(count(m)) } AS r

REJECTED[NegativeIntegerArgument]  EXPLAIN MATCH (n:P) RETURN n.name AS r SKIP -1
ACCEPTED                           EXPLAIN MATCH (n:P) RETURN COLLECT { MATCH (m:P) RETURN m.name SKIP -1 } AS r

REJECTED[NegativeIntegerArgument]  EXPLAIN MATCH (n:P) RETURN n.name AS r LIMIT -1
ACCEPTED                           EXPLAIN MATCH (n:P) RETURN COLLECT { MATCH (m:P) RETURN m.name LIMIT -1 } AS r
```

[...]

PR #5839, opened 2026-08-06
fix(#5825): EXISTS/COUNT/COLLECT subquery bodies validate variable scope

### LINKED ISSUE #5825: EXISTS subquery silently accepts a variable dropped by WITH instead of reporting an unbound-variable error
### ArcadeDB version

ArcadeDB `26.9.1-SNAPSHOT`
Build: `adb16f40bf26d006cad9513c3d3d9bf057622129/1785792317978/main`

Also reproduced on ArcadeDB `26.8.1-SNAPSHOT`
Build: `1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main`

### Environment

* `arcadedata/arcadedb:latest`
* HTTP command API
* Query language: Cypher

### Description

An `EXISTS { ... }` subquery silently accepts a reference to a variable that was removed from scope by a preceding `WITH` clause.

ArcadeDB correctly rejects the same out-of-scope variable when it is referenced directly in `RETURN`:

```cypher
UNWIND [0] AS v
WITH v
WITH 0 AS y
RETURN v;
```

However, when the dropped variable is referenced inside an `EXISTS` subquery, the query succeeds and returns zero rows instead of reporting an unbound-variable error.

This is an internal scope-validation inconsistency: the validity of the same variable reference depends on whether it appears directly in the outer query or inside `EXISTS`.

### Steps to reproduce

Create two nodes:

```cypher
CREATE (:_Oracle {_id:'n1', p:0}),
       (:_Oracle {_id:'n2', p:1});
```

Run the failing query:

```cypher
UNWIND [0] AS v
WITH v
WITH 0 AS y
MATCH (n)
WHERE EXISTS {
  MATCH (x)
  WHERE x.p = v
}
RETURN n._id AS n, y;
```

Run the direct-reference control separately:

```cypher
UNWIND [0] AS v
WITH v
WITH 0 AS y
RETURN v;
```

### Expected behavior

The `EXISTS` query should fail because `v` is no longer in scope after:

```cypher
WITH 0 AS y
```

A client-facing scope error should be returned, such as:

```text
Variable `v` not defined
```

or:

```text
Unbound variable: v
```

### Actual behavior

The `EXISTS` query succeeds without an error or warning:

```text
HTTP 200
0 rows returned
```

The direct-reference control is correctly rejected:

```text
HTTP 400
UndefinedVariable
```

The same out-of-scope variable is therefore rejected in `RETURN` but silently accepted inside `EXISTS`.

#### Control 1: preserving the variable produces the expected rows

```cypher
UNWIND [0] AS v
WITH v
WITH v, 0 AS y
MATCH (n)
WHERE EXISTS {
  MATCH (x)
  WHERE x.p = v
}
RETURN n._id AS n, y
ORDER BY n;
```

Result:

```text
n1 | 0
n2 | 0
```

Because a node with `p = 0` exists, the `EXISTS` predicate is true for both outer rows.

#### Control 2: the dropped variable is rejected in RETURN

```cypher
UNWIND [0] AS v
WITH v
WITH 0 AS y
RETURN v;
```

Result:

```text
HTTP 400
UndefinedVariable
```

### Cross-engine comparison

| Query                                    |     ArcadeDB 26.9.1     |      Neo4j 2026.06     |      Memgraph 3.11     |
| ---------------------------------------- | :---------------------: | :--------------------: | :--------------------: |
| Dropped `v` referenced in `RETURN`       |          Error          |          Error         |          Error         |
| Dropped `v` referenced inside `EXISTS`   | **HTTP 200, zero rows** | Unbound-variable error | Unbound-variable error |
| Preserved `v` referenced inside `EXISTS` |         Two rows        |        Two rows        |        Two rows        |

### Isolated recheck

Re-verified on a fresh container using the latest image on 2026-08-05:

```text
ArcadeDB: 26.9.1-SNAPSHOT
Build: adb16f40bf26d006cad9513c3d3d9bf057622129/1785792317978/main
Image: ec37cbba0752
```

Results:

```text
Run 1: RETURN v -> HTTP 400 UndefinedVariable
       EXISTS   -> HTTP 200, zero rows

Run 2: RETURN v -> HTTP 400 UndefinedVariable
       EXISTS   -> HTTP 200, zero rows

Run 3: RETURN v -> HTTP 400 UndefinedVariable
       EXISTS   -> HTTP 200, zero rows
```

The same behavior was also reproduced in a separate container running `26.8.1-SNAPSHOT`.

Reproduction rate: 3/3 on both builds.

### Why this looks like a bug

1. `WITH 0 AS y` removes `v` from scope.
2. ArcadeDB correctly rejects a direct reference to `v` after that clause.
3. The same invalid reference is silently accepted inside an `EXISTS` subquery.
4. Preserving `v` explicitly causes the query to return the expected rows.
5. Neo4j and Memgraph reject the out-of-scope `EXISTS` reference.

The query should either reference a valid in-scope variable or fail during semantic validation. Returning an empty result silently hides the invalid reference.

### Impact

A variable accidentally omitted from a `WITH` projection can cause an `EXISTS` predicate to silently filter out rows instead of producing a scope error.

This can produce incorrect empty results in filters, existence checks, authorization predicates, and generated queries without any error or warning identifying the out-of-scope reference.

robfrank:
## What does this PR do?

Closes #5825

A variable dropped from scope by a `WITH` clause was correctly rejected when referenced directly
(e.g. `RETURN v` raises `UndefinedVariable`), but the same out-of-scope reference was silently
accepted when it appeared inside an `EXISTS { ... }`, `COUNT { ... }`, or `COLLECT { ... }` subquery
expression: it resolved as missing/`null` instead of raising a scope error, so the predicate just
evaluated to `false` (or an empty collection) instead of failing loudly. This makes `checkExpressionScope`
in `CypherSemanticValidator` recurse into the parsed body of those three subquery expressions with the
live outer scope, the same way it already does for `CALL { }` bodies, closing the gap.

## Motivation

Reported in #5825 with a reproduction and a cross-engine comparison against Neo4j and Memgraph, both
of which reject the out-of-scope reference. A variable accidentally omitted from a `WITH` projection
can silently turn an `EXISTS` predicate into an always-false filter instead of surfacing a client
error, which is especially dangerous for authorization predicates or generated queries.

## Related issues

Closes #5825

## Root cause and fix

`CypherSemanticValidator.checkExpressionScope(Expression, Set<String>)` already threads the live
scope through every other expression form (list comprehension, list predicate, `CALL { }` subquery
bodies via `validateSubqueryScope`), but had no branch for `ExistsExpression`, `CountExpression`, or
`CollectExpression` - those fell through untouched. This is a different mechanism from the generic
`validateNestedStatements` walk that runs the other ten validation phases against every subquery body
(issue #5656): that walk cannot carry position-accurate scope, so scope validation for `CALL { }` was
already handled as a bespoke recursive call, and this PR adds the same treatment for
`EXISTS`/`COUNT`/`COLLECT`.

Unlike `CALL { }`, no import gate applies: the entire live outer scope is visible to these bodies
without an explicit or importing-`WITH` declaration, matching how the runtime
(`CorrelatedSubqueryRunner.seedRow`) already hands the whole outer row to the body. A `null` parsed
subquery (the best-effort AST build declining the body, issue #5626) is skipped, matching the existing
fallback-to-text-execution behavior for that edge case.

File changed: `engine/src/main/java/com/arcadedb/query/opencypher/parser/CypherSemanticValidator.java`.

## Test plan

New test class: `engine/src/test/java/com/arcadedb/query/opencypher/Issue5825ExistsSubqueryScopeTest.java`
(6 tests):

- `existsSubqueryReferencingDroppedVariableThrows` - the issue's failing repro (`EXISTS { ... }`).
- `countSubqueryReferencingDroppedVariableThrows` / `collectSubqueryReferencingDroppedVariableThrows` -
  same shape for `COUNT { ... }` / `COLLECT { ... }`.
- `existsSubqueryUnionBranchReferencingDroppedVariableThrows` - the out-of-scope reference sits in a
  `UNION` branch of the `EXISTS` body, exercising the `UnionStatement` branch of
  `checkSubqueryExpressionScope`.
- `directReferenceToDroppedVariableThrows` - control, the pre-existing direct-reference rejection
  (unchanged behavior).
- `existsSubqueryReferencingPreservedVariableIsValid` - control, keeping `v` in scope still returns
  the expected two rows.

Before the fix, the EXISTS/COUNT/COLLECT-shaped tests failed with "Expecting code to raise a
throwable" (the query silently succeeded with an empty/wrong result), confirming the tests exercise
the bug. After the fix, all six pass.

- [x] `Issue5825ExistsSubqueryScopeTest`: 6/6 pass.
- [x] Directly related existing regression tests re-run clean: `Issue5213NestedCallScopeTest` (5),
  `CypherSubqueryParseTimeValidationIssue5626Test` (30), `Issue5257CallSubqueryRelationshipScopeTest`
  (10), `Issue5179PatternComprehensionScopeTest` (4).
- [x] Broader EXISTS/COUNT/COLLECT test surface (30 test classes covering correlated subqueries,
  count-pushdown optimizations, pattern predicates, foreach/case interactions, etc.): all pass, no
  regressions.

## Additional Notes

CI note: `unit-tests` / `integration-tests` / `ha-integration-tests` failed on the first CI run for
this PR, but on unrelated, pre-existing tests also failing on plain `main` at the same time
(`CreatePropertyStatementExecutionTest.createHiddenProperty`, `GetServerSettingsToolRedactionIT.clusterTokenIsRedactedInMcpSetServerSetting`,
and a flaky HA Raft teardown test) - confirmed by cross-checking the same job types on independent
`main`-branch CI runs, which show the identical failures. Nothing in this PR's diff touches SQL
property creation, MCP settings redaction, or HA Raft teardown.


robfrank:
@claude review

Issue #4591, opened 2026-06-12
NULL `IN` / `NOT IN` semantics inverted vs SQL standard


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.query.sql.parser.InCondition`
**Severity:** MEDIUM

### Summary

When the right side is null (e.g., parameter or subquery returning
null), the method returns `false`, then XORs with `not`. For
`x NOT IN (:nullParam)` this yields `true` for every row. SQL semantics
require UNKNOWN/NULL (treated as false in WHERE). Likewise,
`NULL NOT IN (1,2,3)` should be UNKNOWN, not TRUE.

### Code reference

`engine/com/arcadedb/query/sql/parser/InCondition.java:59–63, 99–103`

### Impact

`NOT IN` queries with any null operand pass through rows that should
be excluded.

### Suggested fix

Three-valued logic: return `null` (UNKNOWN) when the right list is
null or contains null without a positive match; map UNKNOWN to false
at the WHERE boundary.

lvca:
Fixed

PR #5844, opened 2026-08-06
fix(#5837): guard boolean unboxing in CONTAINS/CONTAINSALL/CONTAINSANY/CONTAINSVALUE evaluate()

### LINKED ISSUE #5837: IN condition throws an exception when inside CONTAINS
When an IN condition is used inside a CONTAINS clause, it throws an exception when handling a `null` return value from `evaluate()` methods.

### Steps to reproduce

```
CREATE DOCUMENT TYPE Product IF NOT EXISTS;
CREATE PROPERTY Product.name IF NOT EXISTS STRING;
CREATE PROPERTY Product.embedded_list IF NOT EXISTS LIST OF Product;

CREATE VERTEX TYPE Supplier IF NOT EXISTS;
CREATE PROPERTY Supplier.name IF NOT EXISTS STRING;
CREATE PROPERTY Supplier.certifications IF NOT EXISTS LIST;
CREATE PROPERTY Supplier.embedded IF NOT EXISTS EMBEDDED OF Product;
CREATE PROPERTY Supplier.embedded_list IF NOT EXISTS LIST OF Product;

INSERT INTO Supplier (name, certifications, embedded, embedded_list) VALUES ('Berlin Sensors GmbH', ['ISO-9001,RoHS'], { "@type": "Product", "name": "CPU"}, [{ "@type": "Product", "name": "CPU", "embedded_list": [{ "@type": "Product", "name": "transistor"}]}]);
```

Executing a SELECT with IN works correctly:

```
SELECT FROM `Supplier` WHERE embedded.name IN ['CPU', 'motherboard']
```

When the IN in inside a CONTAINS clause, we receive an exception:

```
SELECT FROM `Supplier` WHERE embedded_list CONTAINS (
  embedded.name IN ['CPU', 'motherboard']
)
```

Exception:

```
Error on transaction commit
Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "com.arcadedb.query.sql.parser.BooleanExpression.evaluate(com.arcadedb.query.sql.executor.Result, com.arcadedb.query.sql.executor.CommandContext)" is null
```

### Research

I believe the exception is caused by the following line:

```
com/arcadedb/query/sql/parser/InCondition.class
public Boolean evaluate(final Result currentRecord, final CommandContext context)
Line #108
return null;
```

I believe it should be changed to:
`return false;`

A similar pattern is found in:

```
com/arcadedb/query/sql/parser/WhereClause.class
public Boolean matchesFilters(final Result currentRecord, final CommandContext context)
Line #60
// In WHERE clause filtering context, treat null as false
return result != null ? result : false;
```

There are other places in the code where `evaluate()` methods `return null;`.

### Used tools

ArcadeDB Server v26.8.1-SNAPSHOT

robfrank:
## What does this PR do?

`InCondition.evaluate()` correctly implements SQL three-valued logic: it returns `Boolean.TRUE`, `Boolean.FALSE`, or `null` (UNKNOWN) when the left operand is `null` and no match is found (see `NullInConditionTest`, issue #4591). `WhereClause.matchesFilters` already collapses that UNKNOWN to `false` at the outermost WHERE boundary.

`ContainsCondition.evaluate()` did not do the same at its own boundary: it unboxed the nested condition's `Boolean` result directly inside `if (... && condition.evaluate(...))`, with no null guard, so a nested condition (e.g. `IN`) evaluating to UNKNOWN threw:

```
Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of
"com.arcadedb.query.sql.parser.BooleanExpression.evaluate(...)" is null
```

The same copy-pasted unguarded-unboxing pattern existed in `ContainsAllCondition`, `ContainsAnyCondition`, and `ContainsValueCondition`.

Fix: added `BooleanExpression.isTrue(Boolean)` (`Boolean.TRUE.equals(value)`), matching the existing WHERE-boundary pattern, and applied it at every unguarded unboxing site in the four classes above. `InCondition` itself is intentionally unchanged — its `null` return is correct SQL semantics, and forcing it to always return `false` would corrupt three-valued logic elsewhere (e.g. `NOT (x IN (...))` under De Morgan's laws).

### Discovered while testing, not fixed here (separate pre-existing bugs, out of scope)

- `SQLASTBuilder.visitContainsAllCondition` / `visitContainsAnyCondition` hard-cast the parsed nested condition to `OrBlock` (`condition.rightBlock = (OrBlock) whereClause.baseExpression`). Since the grammar collapses a single, non-OR-wrapped condition straight through without wrapping it, any bare condition (`x CONTAINSALL (name IN [...])`, `x CONTAINSALL (name = 'CPU')`) throws `ClassCastException` while parsing, before this fix's guard is ever reached. `ContainsAllCondition`/`ContainsAnyCondition` still got the same `isTrue()` guard for defense in depth, but no SQL-level test could be written for them until that parsing bug is fixed separately (would require widening `rightBlock` from `OrBlock` to `BooleanExpression`).
- `ContainsAnyCondition`'s nested-condition branch returns `false` as soon as *any* item fails and `true` only once *every* item matches — i.e. it implements "ALL", identical to `ContainsAllCondition`, not "ANY". Pre-existing, unrelated to this fix, and currently unreachable anyway due to the bug above.
- `ContainsValueCondition`'s nested-condition (`condition`) field is unreachable from SQL today: the grammar only defines `expression CONTAINSVALUE expression`, no `CONTAINSVALUE (whereClause)` production. Fixed defensively; no SQL-level test possible.

## Motivation

Fixes a high-priority reported NullPointerException.

## Related issues

Closes #5837

## Additional Notes

Full root-cause writeup and verification log: `docs/5837-contains-nested-in-null-npe.md`.

## Test plan

- [x] New regression test `engine/src/test/java/com/arcadedb/query/sql/operator/ContainsNestedInConditionNullTest.java`, including a faithful reproduction of the issue's exact repro.
- [x] Verified the new test fails with the reported NPE when the fix is reverted (3 of 4 methods throw `NullPointerException` on the unfixed code), then passes with the fix restored.
- [x] `mvn -pl engine -am compile` — clean.
- [x] `mvn -pl engine -am test -Dtest='Contains*Test,QueryTest,SelectStatementTest,NullInConditionTest,InConditionSubqueryTest'` — 167 tests, 0 failures, 0 errors.

## Checklist

- [x] I have run the build using `mvn clean package` command (module-scoped `mvn -pl engine -am compile`/`test`, full reactor not required for this change)


robfrank:
@claude review

robfrank:
@claude review

PR #5854, opened 2026-08-06
fix(sql): plan WHERE @rid = / IN [RIDs] as a direct RID fetch, not a full type scan

### LINKED ISSUE #5824: SQL: `WHERE @rid IN [list]` is planned as a full type scan, so resolving k RIDs costs O(n)
## Summary

`SELECT ... FROM <Type> WHERE @rid IN [ ... ]` is planned as a full bucket scan with a per-record filter. The cost is linear in the size of the type and independent of how many RIDs are in the list.

The address-based path already exists. `SELECT ... FROM [ ... ]` over the same RIDs plans as `FETCH FROM RIDs` and is flat in type size. So this looks like a missing planner rewrite rather than a missing capability.

## Measured

Java against the released `arcadedb-engine-26.8.1.jar` (taken from the `arcadedb-embedded==26.8.1` wheel), embedded database, k = 10 RIDs, 5 warmup + 20 timed iterations per point:

```
docs      form                            ms/query    ms per doc  plan step
100000    WHERE @rid IN [...]                21.25      0.000213  + FETCH FROM TYPE Doc WITH FILTER ()
100000    FROM [...]                          0.10      0.000001  + FETCH FROM RIDs
400000    WHERE @rid IN [...]                82.12      0.000205  + FETCH FROM TYPE Doc WITH FILTER ()
400000    FROM [...]                          0.07      0.000000  + FETCH FROM RIDs

4x the documents:
  WHERE @rid IN [...]  3.86x the time  -> LINEAR in type size (scan)
  FROM [...]           0.74x the time  -> flat (address-based)

At 400000 docs the target form is 1143x faster.
```

`ms per doc` stays flat for the `IN` form across a 4x change in type size. That is the signal: the query is reading the type, not the RIDs. Both forms return the same rows, and the repro checks that before it times anything.

## At scale

At 8.84M documents (9.05 GB) the engine logs its own complaint about it:

```
WARNI [FetchFromTypeExecutionStep] Attempt to scan type 'Doc' in database 'l3_arcade'
of total size 9.05GB 901 times. This operation is very expensive, consider using an index
```

Measured there, one `WHERE @rid IN [10 rids]` query costs about 4.7 s, steady across 900 consecutive queries (7 min 52 s per 100-query batch, repeatable within 2 s).

## Why it comes up

This is the natural way to resolve the output of anything that returns RIDs. `vector.sparseNeighbors` and `vector.neighbors` both return RIDs, so the second half of a two-step vector query lands on it: the ANN half is index-backed and fast, then the resolve half reads the whole corpus.

## Not a regression

Same plan and the same order of magnitude on 26.8.1.dev22, so this is long-standing behaviour rather than something that changed recently.

## Workaround

Rewrite the RID list as the query target, `SELECT ... FROM [#1:0, #1:1, ...]`, which already uses `FETCH FROM RIDs`.

## Related

#5756 is in the same execution step but is a different defect. There the column projection push-down inside `ScanWithFilterStep` is unreachable. This one is about the planner choosing that step at all for a RID list.

<details>
<summary>Repro (Java, self-contained)</summary>

```java
import com.arcadedb.query.sql.executor.Result;

import java.io.File;
import java.util.ArrayList;
import java.util.List;

/**
 * Companion to RidInScanRepro: is there a SQL form that resolves a RID list in
 * O(k) instead of scanning the type?
 *
 * Two forms, same k RIDs, same result, two type sizes:
 *   A. SELECT id FROM Doc WHERE @rid IN [ ... ]   the natural form
 *   B. SELECT id FROM [ ... ]                      RIDs as the query TARGET
 *
 * The interesting column is ms/query as n grows. A form that is address-based
 * is FLAT in n. A form that scans is LINEAR in n. Reporting whether a
 * workaround exists is the difference between "this is slow" and "this is slow
 * and here is what to do until it is fixed", so it belongs in the issue.
 */
public class RidInWorkaroundRepro {

  private static final int[] SIZES = { 100_000, 400_000 };
  private static final int K = 10;
  private static final int WARMUP = 5;
  private static final int TIMED = 20;

  public static void main(String[] args) {
    final List<double[]> inTimes = new ArrayList<>(), targetTimes = new ArrayList<>();

    System.out.printf("%-9s %-28s %11s %13s  %s%n",
        "docs", "form", "ms/query", "ms per doc", "plan step");

    for (final int n : SIZES) {
      final String path = "/tmp/ridwa_" + n;
      deleteTree(new File(path));

      try (final Database db = new DatabaseFactory(path).create()) {
        db.command("sql", "CREATE DOCUMENT TYPE Doc");
        db.command("sql", "CREATE PROPERTY Doc.id LONG");
        db.transaction(() -> {
          for (int i = 0; i < n; i++) {
            final MutableDocument doc = db.newDocument("Doc");
            doc.set("id", i);
            doc.save();
          }
        });

        final List<String> rids = new ArrayList<>();
        try (final ResultSet rs = db.query("sql", "SELECT @rid AS r FROM Doc LIMIT " + K)) {
          while (rs.hasNext())
            rids.add(rs.next().getProperty("r").toString());
        }
        final String list = String.join(", ", rids);

        final String qIn     = "SELECT id FROM Doc WHERE @rid IN [" + list + "]";
        final String qTarget = "SELECT id FROM [" + list + "]";

        // Both forms must return the same rows, or the timing is meaningless.
        final int nIn = count(db, qIn), nTarget = count(db, qTarget);
        if (nIn != nTarget) {
          System.out.printf("  MISMATCH at n=%d: IN returned %d rows, target returned %d%n",
              n, nIn, nTarget);
          return;
        }

        final double tIn     = time(db, qIn);
        final double tTarget = time(db, qTarget);

        System.out.printf("%-9d %-28s %11.2f %13.6f  %s%n",
            n, "WHERE @rid IN [...]", tIn, tIn / n, step(db, qIn));
        System.out.printf("%-9d %-28s %11.2f %13.6f  %s%n",
            n, "FROM [...]", tTarget, tTarget / n, step(db, qTarget));

        inTimes.add(new double[] { n, tIn });
        targetTimes.add(new double[] { n, tTarget });
      }
      deleteTree(new File(path));
    }

    final double sizeRatio = SIZES[1] / (double) SIZES[0];
    final double inRatio     = inTimes.get(1)[1]     / inTimes.get(0)[1];
    final double targetRatio = targetTimes.get(1)[1] / targetTimes.get(0)[1];

    System.out.printf("%n%.0fx the documents:%n", sizeRatio);
    System.out.printf("  WHERE @rid IN [...]  %.2fx the time  -> %s%n", inRatio,
        inRatio > sizeRatio * 0.5 ? "LINEAR in type size (scan)" : "flat (address-based)");
    System.out.printf("  FROM [...]           %.2fx the time  -> %s%n", targetRatio,
        targetRatio > sizeRatio * 0.5 ? "LINEAR in type size (scan)" : "flat (address-based)");
    System.out.printf("%nAt %d docs the target form is %.0fx faster.%n",
        SIZES[1], inTimes.get(1)[1] / targetTimes.get(1)[1]);
  }

  private static int count(final Database db, final String q) {
    int c = 0;
    try (final ResultSet rs = db.query("sql", q)) {
      while (rs.hasNext()) { rs.next(); c++; }
    }
    return c;
  }

  private static double time(final Database db, final String q) {
    for (int i = 0; i < WARMUP; i++) count(db, q);
    for (int i = 0; i < TIMED; i++) count(db, q);
    return (System.nanoTime() - t0) / 1e6 / TIMED;
  }

  /** Take the plan as Object: Result.getProperty is generic, so assigning it
   *  straight into a String makes javac insert a cast that fails with a
   *  message-less ClassCastException. */
  private static String step(final Database db, final String q) {
    try (final ResultSet rs = db.query("sql", "EXPLAIN " + q)) {
      if (!rs.hasNext()) return "?";
      final Result r = rs.next();
      Object plan = r.getProperty("executionPlanAsString");
      if (plan == null) plan = r.getProperty("executionPlan");
      if (plan == null) plan = r.toJSON();
      for (final String line : String.valueOf(plan).split("\\n"))
        if (line.contains("SCAN") || line.contains("FETCH FROM"))
          return line.trim();
      final String flat = String.valueOf(plan).replaceAll("\\s+", " ");
      return flat.substring(0, Math.min(70, flat.length()));
    }
  }

  private static void deleteTree(final File f) {
    if (!f.exists()) return;
    final File[] kids = f.listFiles();
    if (kids != null) for (final File k : kids) deleteTree(k);
    f.delete();
  }
}
```

Built and run against the release jars with:

```bash
pip download --no-deps --only-binary=:all: --platform manylinux_2_34_x86_64 \
    --python-version 312 arcadedb-embedded==26.8.1 -d /tmp/whl
cd /tmp/whl && unzip -q -o *.whl -d /tmp/pkg
CP="/tmp/pkg/arcadedb_embedded/jars/*"
javac -cp "$CP" -d /tmp/out RidInWorkaroundRepro.java
java --add-modules jdk.incubator.vector -cp "/tmp/out:$CP" RidInWorkaroundRepro
```

</details>

lvca:
## Summary

`SELECT ... FROM <Type> WHERE @rid IN [...]` (and `WHERE @rid = <RID>`) was falling through to the generic predicate-pushdown path and planning as `FETCH FROM TYPE ... WITH FILTER`, i.e. a full scan of every record of the type with a per-record filter. Cost was O(n) in the size of the type and independent of how many RIDs were being looked up, even though the equivalent RID-as-target form (`SELECT ... FROM [...]`) already resolves the same RIDs in O(k) via `FetchFromRidsStep`.

`SelectExecutionPlanner.handleTypeAsTarget` now recognizes a single-AND-block WHERE clause containing `@rid = <RID>` or `@rid IN [<RID list>]` on a type target (bracket list, parenthesized list, or a bind parameter bound to a RID / collection of RIDs), resolves the RID(s) at plan time, and rewrites the fetch to `FetchFromRidsStep` + `FilterByTypeStep` (to preserve the "only this type" semantics of the original scan), keeping any remaining AND-ed conditions as a `FilterStep`. This mirrors the existing `@out`/`@in` equality → vertex-centric-fetch rewrite already in the planner for edge types.

The rewrite deliberately backs off (falls back to the ordinary scan) when:
- the WHERE clause has a top-level `OR` (another branch could match records outside this RID set),
- a remaining AND-ed condition references a per-record `LET` variable (LET is only populated after this fetch step runs — pushing such a condition down here would silently drop every row; see `LetWherePredicatePushdownTest` for the existing precedent of this hazard),
- `@rid NOT IN [...]` (exclusion still requires scanning everything else in the type).

A duplicate RID in the `IN` list is deduped before building `FetchFromRidsStep`, since a scan evaluates `IN` as a per-record membership test and must not visit the same record twice for a repeated list value.

Fixes #5824.

## Test plan

- [x] New `RidInScanOptimizationTest` (14 cases): plan-shape assertions (`FetchFromRidsStep` used, not `FetchFromTypeWithFilterStep`/`FetchFromTypeExecutionStep`) for `=`, bracket `IN [...]`, parenthesized `IN (...)`, and bind-parameter `IN ?`; correctness for combined AND conditions, RIDs of another type being filtered out, non-existent RIDs being skipped without error, empty/null bind-parameter lists, duplicate RIDs not duplicating rows, `OR`/`NOT IN` still falling back to a correct scan, a per-record-LET-referencing remaining condition staying correct, and a partitioned type ignoring bucket pruning correctly.
- [x] `mvn -o compile` clean.
- [x] Full targeted regression pass, 0 failures across 498 tests: `SelectStatementExecutionTest`, `EdgeTypeVertexRidOptimizationTest` (the closest existing precedent), `PartitionPruningPlannerTest`, `MatchStatementExecutionTest`, `DeleteExecutionStepTest`, `UpdateStatementExecutionTest`, `Issue5505BoundRidCollectionTargetTest`, `MatchRidFilterTest`, `RidSetTest`, `FetchFromTypeExecutionStepTest`, `SelectStatementTest`, `SQLGrammarRegressionTest`, `ExplicitLockingTransactionTest`, `LetWherePredicatePushdownTest`, `LetDivisionBugTest`, `ScriptExecutionTest`, plus the new test class.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -53,6 +53,7 @@ import com.arcadedb.query.sql.parser.GeOperator;
 import com.arcadedb.query.sql.parser.GroupBy;
 import com.arcadedb.query.sql.parser.GtOperator;
 import com.arcadedb.query.sql.parser.Identifier;
+import com.arcadedb.query.sql.parser.InCondition;
 import com.arcadedb.query.sql.parser.IndexIdentifier;
 import com.arcadedb.query.sql.parser.InputParameter;
 import com.arcadedb.query.sql.parser.IsNullCondition;
@@ -104,6 +105,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -1847,6 +1849,13 @@ public class SelectExecutionPlanner {
       return;
     }
 
+    // RID-address optimization: for queries like SELECT FROM Type WHERE @rid = <RID> or
+    // WHERE @rid IN [<RID list>], fetch the records directly by RID address instead of
+    // scanning the whole type. This is always cheaper than an index lookup, so it is checked
+    // before the index-based paths below. See issue #5824.
+    if (handleTypeWithRidFilter(plan, identifier, info, context))
+      return;
+
     if (handleTypeAsTargetWithIndexedFunction(plan, effectiveClusters, identifier, info, context)) {
       plan.chain(new FilterByTypeStep(identifier, context));
       return;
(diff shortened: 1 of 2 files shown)


lvca:
@claude please review this

lvca:
Thanks for the thorough trace, especially confirming the LET-ordering guard and the plan-cache `canBeCached()` reasoning line up with the actual hazards.

Verdict on the three non-blocking notes:

**1. Plan-cache regression for `WHERE @rid = ?`/`IN ?`.** Confirmed real and correctly diagnosed. Deliberately not fixing in this PR - making `FetchFromRidsStep` cacheable via a re-resolvable parameter reference (instead of baked RID values) is a design change that touches every other call site of that step (`FROM [...]`, bind-variable targets, function-call targets), not something scoped to #5824. Filed as #5855 for follow-up, with a note to measure whether the replan cost is even worth the added complexity before doing it.

**2. Pre-existing LET-guard gap in `handleEdgeTypeWithVertexRidFilter`.** Agreed this looks like a real latent bug, and agreed it predates this PR and is out of scope here. Filed as #5856 with a suggested repro and the fix (mirror this PR's guard).

**3. `derivePartitionPrunedClusters` running unconditionally before the RID short-circuit.** I looked at this harder than "trivial, skip it" and I'm glad I did: reordering would have introduced a real regression. That computation's side effect (`PARTITION_PRUNED_BUCKET_FILE_IDS_VAR`/`PARTITION_PRUNED_TYPE_NAME_VAR`, `SelectExecutionPlanner.java:2272-2293`) exists specifically so `vector.neighbors()`/`vector.sparseNeighbors()` can narrow their own per-bucket fan-out (`SQLFunctionVectorAbstract.narrowAllowedBucketIdsByPartitionHint`). For the composite query this issue's own description calls out as the motivating case - `WHERE @rid IN vector.neighbors(...)` - `extractRidEqualityOrInList` evaluates `vector.neighbors(...)` inline via `rightMathExpression.execute()`, which reads that hint. Moving the RID check earlier would run that evaluation before the hint is set, silently losing the bucket-pruning optimization for exactly the pattern #5824 is about. Leaving the order as-is on purpose, not just because the cost is negligible.

No code changes from this round - all three points are either out of scope for this PR (filed as follow-ups) or actively wrong to "fix" (point 3). Local verification stands: `mvn -o compile` clean, 498 targeted tests passing including the new `RidInScanOptimizationTest` (14 cases).

PR #5858, opened 2026-08-06
fix(#5845): RemoteDatabase commit()/begin()/rollback() capture the commit-index bookmark

### LINKED ISSUE #5845: RemoteDatabase.commit() drops the X-ArcadeDB-Commit-Index bookmark, silently breaking READ_YOUR_WRITES
## Summary

With `ReadConsistency.READ_YOUR_WRITES`, a read issued after an explicit transaction commit can be served by a follower that has not yet applied that commit. The guarantee silently does not hold.

The cause is that `RemoteDatabase.commit()` discards the `X-ArcadeDB-Commit-Index` response header, which is the only place the server publishes the bookmark for the just-committed transaction.

Version: 26.8.1 (Java remote client, HA).

## Detail

The client's bookmark is advanced in exactly one place, `RemoteHttpComponent.httpCommand()`:

```java
// network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java:348
if (this instanceof RemoteDatabase remoteDb) {
  response.headers().firstValue("X-ArcadeDB-Commit-Index").ifPresent(val -> {
    try {
      remoteDb.updateLastCommitIndex(Long.parseLong(val));
    } catch (final NumberFormatException ignored) {
    }
  });
}
```

`commit()` does not go through `httpCommand()`. It builds its own request, calls the raw client, and never inspects the response headers:

```java
// network/src/main/java/com/arcadedb/remote/RemoteDatabase.java:353
HttpRequest request =
    createRequestBuilder("POST", getUrl("commit", databaseName)).POST(HttpRequest.BodyPublishers.noBody())
        .build();

HttpResponse<String> response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());

if (response.statusCode() != 204) {
  ...
}
committed = true;
```

The server does emit the header on that exact response. `PostCommitHandler extends DatabaseAbstractHandler`, and the parent writes it immediately after committing:

```java
// server/src/main/java/com/arcadedb/server/http/handler/DatabaseAbstractHandler.java:183
database.commit();

// Emit bookmark header for read-your-writes consistency
if (haDbForRead != null) {
  final long lastApplied = haDbForRead.getLastAppliedIndex();
  if (lastApplied >= 0)
    exchange.getResponseHeaders().put(new HttpString("X-ArcadeDB-Commit-Index"), String.valueOf(lastApplied));
}
```

So the server offers the one bookmark that matters and the client drops it.

## Reproducer

```java
db.setReadConsistency(ReadConsistency.READ_YOUR_WRITES);

db.begin();
db.command("sql", "INSERT INTO Person SET name = 'x'");
db.commit();                       // X-ArcadeDB-Commit-Index discarded here

// routed to a follower
final ResultSet rs = db.query("sql", "SELECT FROM Person WHERE name = 'x'");
// may be empty: the follower has not applied the commit
```

At the read, `lastCommitIndex` holds whatever the last *command* response carried, which is the server's applied index **before** this transaction committed. If no command ever returned the header, it is still `-1`, and `httpCommand()` then omits `X-ArcadeDB-Read-After` entirely:

```java
// RemoteHttpComponent.java:305
if (rc == ReadConsistency.READ_YOUR_WRITES) {
  final long last = remoteDb.getLastCommitIndex();
  if (last >= 0)
    requestBuilder = requestBuilder.header("X-ArcadeDB-Read-After", String.valueOf(last));
}
```

In that case the read degrades to `EVENTUAL` with no indication to the caller.

The failure is silent: a correct-looking empty result set, no warning, no exception.

## Suggested fix

Capture the header in `commit()` the same way `httpCommand()` does, before the status check:

```java
HttpResponse<String> response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());

response.headers().firstValue("X-ArcadeDB-Commit-Index").ifPresent(val -> {
  try {
    updateLastCommitIndex(Long.parseLong(val));
  } catch (final NumberFormatException ignored) {
  }
});
```

`begin()` and `rollback()` are raw sends too and drop the header as well; only `commit()` advances the applied index, so it is the one that matters for correctness, but factoring the capture into a small helper shared by all four call sites would keep them from drifting apart again. The same reasoning appears in a comment already in the code, on `getUrl()`:

```java
// Routing for raw httpClient.send() callers (RemoteDatabase.begin/commit/rollback).
// Regular query/command traffic flows through httpCommand() - any STICKY routing
// change must update both paths.
```

The bookmark capture has the same two-path problem, and currently only one path implements it.

lvca:
## Summary

Fixes #5845. With `ReadConsistency.READ_YOUR_WRITES`, a read issued right after an explicit transaction `commit()` could be served by a follower that had not yet applied that commit, with no error or warning.

`RemoteDatabase.begin()`, `commit()`, and `rollback()` build their own HTTP request and send it with the raw `httpClient`, bypassing `RemoteHttpComponent.httpCommand()`, which is where the `X-ArcadeDB-Commit-Index` response header is normally parsed into `lastCommitIndex`. `commit()` is the one that matters for correctness: without capturing it, `lastCommitIndex` still holds whatever the last `command()` response carried (or `-1` if none did), so the very next `X-ArcadeDB-Read-After` bookmark sent to the server is stale or missing, and the read silently degrades to `EVENTUAL`.

The server already emits `X-ArcadeDB-Commit-Index` on begin/commit/rollback responses too (`DatabaseAbstractHandler` emits it whenever the wrapped database is HA and has an applied index), so all three call sites now go through a shared `captureCommitIndexHeader()` helper. `RemoteHttpComponent.httpCommand()` was refactored to reuse the same helper instead of duplicating the parsing logic.

## Changes

- `network/src/main/java/com/arcadedb/remote/RemoteDatabase.java`: added `captureCommitIndexHeader()`, called from `begin()`, `commit()`, `rollback()` after a successful (204) response.
- `network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java`: `httpCommand()` now calls the shared helper instead of inlining the same header-parsing logic.
- `network/src/test/java/com/arcadedb/remote/RemoteDatabaseTest.java`: two regression tests using a loopback HTTP test double (matching the existing pattern in `RemoteHttpComponentTest`) that assert `commit()` captures `X-ArcadeDB-Commit-Index` when present, and leaves `lastCommitIndex` unchanged when absent.

## Test plan

- [x] New unit tests fail before the fix (reproduces the bug exactly as described in the issue) and pass after.
- [x] Full `network` module unit test suite: 405 tests, 0 failures.
- [x] `ha-raft` module's existing `RaftRemoteReadYourWritesIT` (real 3-node Raft cluster, end-to-end read-your-writes): 2/2 tests pass, no regression.

### DIFF
--- a/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteDatabase.java
@@ -326,6 +326,7 @@ public class RemoteDatabase extends RemoteHttpComponent implements BasicDatabase
         throw new TransactionException("Error on transaction begin", detail);
       }
 
+      captureCommitIndexHeader(response);
       setSessionId(response.headers().firstValue(ARCADEDB_SESSION_ID).orElse(null));
     } catch (final Exception e) {
       throw new TransactionException("Error on transaction begin", e);
@@ -365,6 +366,7 @@ public class RemoteDatabase extends RemoteHttpComponent implements BasicDatabase
 
         throw new TransactionException("Error on transaction commit", detail);
       }
+      captureCommitIndexHeader(response);
       committed = true;
     } catch (final DuplicatedKeyException | ConcurrentModificationException e) {
       throw e;
@@ -399,12 +401,30 @@ public class RemoteDatabase extends RemoteHttpComponent implements BasicDatabase
         final Exception detail = manageException(response, "rollback transaction");
         throw new TransactionException("Error on transaction rollback", detail);
       }
+      captureCommitIndexHeader(response);
     } catch (final Exception e) {
       throw new TransactionException("Error on transaction rollback", e);
     } finally {
       resetCreatedRecordsIdentity();
       setSessionId(null);
     }
+  }
+
+  /**
+   * Captures the {@code X-ArcadeDB-Commit-Index} response header on raw {@link #begin()}/{@link #commit()}/
+   * {@link #rollback()} sends, which bypass {@link RemoteHttpComponent#httpCommand} and its bookmark capture
+   * (issue #5845). The server emits this header on every begin/commit/rollback response once the database
+   * has an applied index, so all three call sites can carry the just-committed bookmark forward for the next
+   * {@link ReadConsistency#READ_YOUR_WRITES} read.
+   */
+  void captureCommitIndexHeader(final HttpResponse<String> response) {
+    response.headers().firstValue("X-ArcadeDB-Commit-Index").ifPresent(val -> {
+      try {
+        updateLastCommitIndex(Long.parseLong(val));
+      } catch (final NumberFormatException ignored) {
+        // server sent an invalid header; ignore
+      }
+    });
   }
 
   /**
--- a/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
+++ b/network/src/main/java/com/arcadedb/remote/RemoteHttpComponent.java
@@ -345,15 +345,8 @@ public class RemoteHttpComponent extends RWLockContext {
         HttpResponse<String> response = sendWithWatchdog(request);
 
         // Capture commit-index from response for read-your-writes consistency.
-        if (this instanceof RemoteDatabase remoteDb) {
-          response.headers().firstValue("X-ArcadeDB-Commit-Index").ifPresent(val -> {
-            try {
-              remoteDb.updateLastCommitIndex(Long.parseLong(val));
-            } catch (final NumberFormatException ignored) {
-              // server sent an invalid header; ignore
-            }
-          });
-        }
+        if (this instanceof RemoteDatabase remoteDb)
+          remoteDb.captureCommitIndexHeader(response);
 
         if (response.statusCode() != 200) {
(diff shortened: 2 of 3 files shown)


lvca:
Thanks for the review. Addressed both points:

1. **Repeated header-name literal**: after this PR's refactor, `X-ArcadeDB-Commit-Index` already appears exactly once on the client side (in the new `captureCommitIndexHeader()`), and `X-ArcadeDB-Read-After`/`X-ArcadeDB-Read-Consistency` each appear exactly once in `httpCommand()`, so there's no duplication left within `network`. The remaining copy you're seeing is in `DatabaseAbstractHandler` on the server side, in a different module with no dependency relationship to `network` in that direction; introducing a shared constant would mean adding a new shared module (or an inverted dependency) for three string literals that already only differ by one word. Not doing that here, since it's out of proportion to the problem and out of scope for this fix.

2. **Test double body-draining**: added a comment on `withHttpServer()` explaining why not draining the request body is safe (no connection reuse, client isn't waiting on us to read it before it reads the response) - 43d15a7.

lvca:
@claude please review this.

lvca:
@claude please review

PR #5859, opened 2026-08-06
fix(cypher): bound expression nesting depth to prevent StackOverflowError

### LINKED ISSUE #5851: openCypher parser throws StackOverflowError on a 2 KB query (~1000 nested parentheses); the SQL parser handles the same input
## Summary

The openCypher parser throws `StackOverflowError` on a ~2 KB query with about 1000 nested parentheses. The SQL parser accepts the same nesting depth without trouble, so the behaviour is specific to Cypher and the safe handling already exists elsewhere in the engine.

Version: 26.8.1.

## Reproducer

```java
final Path dir = Files.createTempDirectory("arcadedb-so");
try (final DatabaseFactory factory = new DatabaseFactory(dir.resolve("db").toString());
     final Database db = factory.create()) {

  final int depth = 1001;
  final String cypher = "MATCH (n) WHERE " + "(".repeat(depth) + "1=1" + ")".repeat(depth);

  db.getQueryEngine("opencypher").analyze(cypher);   // StackOverflowError
}
```

Bisecting the depth on a stock JVM (Temurin 21.0.3, default stack size):

| depth | result |
|---|---|
| 1000 | parsed |
| **1001** | **StackOverflowError** |

The payload at the threshold is **2032 bytes**.

The same depth in SQL is parsed without incident:

```java
db.getQueryEngine("sql").analyze("SELECT FROM V WHERE " + "(".repeat(1001) + "1=1" + ")".repeat(1001));
// returns normally
```

Other shapes reach it too, all through the same recursion: `MATCH (n) WHERE NOT NOT NOT ... true` at ~1000 repetitions, and flat chains such as `a=1 OR a=1 OR ...` or `'x' + 'x' + ...` at ~20 000 terms.

## Where it recurses

```
com.arcadedb.query.opencypher.grammar.Cypher25Parser$Expression6Context.<init>(Cypher25Parser.java:11836)
com.arcadedb.query.opencypher.grammar.Cypher25Parser.expression6(Cypher25Parser.java:11855)
com.arcadedb.query.opencypher.grammar.Cypher25Parser.expression7(Cypher25Parser.java:11437)
com.arcadedb.query.opencypher.grammar.Cypher25Parser.expression8(Cypher25Parser.java:11364)
com.arcadedb.query.opencypher.grammar.Cypher25Parser.expression9(Cypher25Parser.java:11287)
com.arcadedb.query.opencypher.grammar.Cypher25Parser.expression10(Cypher25Parser.java:11205)
```

The generated parser walks the full expression-precedence cascade for every nesting level, so each parenthesis costs roughly ten Java frames. That is what makes 1000 levels enough, and it is why the input can stay this small.

## Impact

Not a crash: `AbstractServerHttpHandler` catches `Throwable`, so an HTTP request carrying such a query degrades to a 500 rather than killing the worker thread. What remains is worth fixing:

- A malformed query is answered with a 500 instead of the 400 a parse failure should produce.
- The cost is asymmetric. Two kilobytes on the wire force the server to grow and unwind an entire thread stack. Issued concurrently across the worker pool, this is a cheap way to keep every worker busy doing nothing useful.
- The reach is wider than the HTTP command endpoint: the new `mcp` module exposes Cypher-backed tools, and `analyze()` is on the path there too.
- `StackOverflowError` is an `Error`. Catching it and continuing is defensible during parsing, where no database state has been touched, but it makes the boundary between "recoverable bad input" and "the JVM is in trouble" depend on where the overflow happened rather than on what it means.

## Suggested fix

Bound the parser's recursion the way the SQL side already does, and report exceeding it as an ordinary parse error. Two common approaches:

- Count nesting depth in the parse listener and fail with `CommandParsingException` past a configurable maximum, which keeps the failure a 400 and gives operators a knob.
- Run the parse on a thread with a known stack size and treat `StackOverflowError` there as a parse failure. This bounds the blast radius but still pays the unwind, so the first option is preferable.

Either way the diagnostic should name the limit, since a legitimate generated query near the boundary is otherwise indistinguishable from an attack.

lvca:
## Summary

Fixes #5851. A ~2KB `WHERE` clause with ~1000 nested parentheses crashed the parsing thread with a `StackOverflowError`: the ANTLR-generated Cypher parser re-enters its `expression` grammar rule (walking its full ~10-level operator-precedence cascade) once per nesting level, so a few thousand levels is enough to exhaust the default JVM thread stack.

Investigation found **two independent recursion sites**, not one as the issue's own diagnosis assumed:

1. **The ANTLR-generated parser itself** - confirmed by the issue's own stack trace. Every nested parenthesis, list/map literal or function argument re-enters `Cypher25Parser.RULE_expression`.
2. **`ExpressionRewriter.rewrite()`**, ArcadeDB's own post-parse AST visitor that normalizes/folds/simplifies every `WHERE` condition - a bug the issue's diagnosis missed. The grammar's `(OR expression11)*` and `(PLUS expression5)*` productions are quantifier loops, not self-referencing rules, so a long flat chain of `OR`'d or string-concatenated terms parses fine without recursing in the parser at all. It only crashes afterwards, when this shared, statically-cached rewriter instance walks the resulting deep expression tree.

The issue's "NOT NOT NOT ... true" shape is also a Kleene-star loop in the grammar and never reproduces in the parser, but a long enough NOT chain still builds a deeply nested AST and is caught by the same rewriter-side guard - broader protection than the issue asked for.

## Fix

- New `CypherExpressionDepthGuard` (`ParseTreeListener`), attached via `Parser.addParseListener`, counts `RULE_expression` re-entries and throws `CommandParsingException` past the limit - no grammar/codegen changes needed, matching the issue's own suggested "count nesting depth in the parse listener" approach.
- `ExpressionRewriter.rewrite()` gets a `ThreadLocal`-based depth counter (required because `CypherASTBuilder.AST_REWRITER` is a single static instance shared across concurrently-running queries) guarding the same way.
- Both share one new config, `arcadedb.cypher.maxExpressionDepth` (default 200, `SCOPE.DATABASE`) - the operator "knob" the issue asked for.
- `CommandParsingException` already maps to HTTP 400 in the existing handler chain, so no server-side wiring was needed for that part.

The SQL parser was never at risk from the same input because its hand-written recursive-descent implementation costs far fewer Java stack frames per nesting level, not because it enforces a limit of its own - I did not add one there since it isn't reproducible.

## Test plan

- [x] New `CypherExpressionDepthGuardIssue5851Test` (9 tests): the issue's exact reproducer (1001 nested parens, via `analyze()`), a 30k-term OR chain, a 30k-term string-concatenation chain, and a 2000-deep NOT chain all now fail with `CommandParsingException` instead of `StackOverflowError`; moderate depths (50) of each shape still parse fine; the config is a real, effective knob (raising it accepts a query previously rejected at the default).
- [x] `mvn -pl engine test` on the opencypher `parser`/`rewriter` packages plus all 221 top-level `com.arcadedb.query.opencypher.*Test` classes: 2362 tests, 0 failures, 0 errors.
- [x] `mvn -pl engine -am compile` clean.

### DIFF
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -2369,27 +2369,3 @@ This is **not new** in this release - the previous two-phase walk collected the 
 recorded here because it was never written down, and because it holds for lightweight self-loops too, where there
 is no record to go missing on the second pass. Pinned as an exact count by
 `Issue5760VertexDeleteSelfSideSkipTest.aSelfLoopIsWalkedFromBothListsSoItsDeleteEventFiresTwice`.
-
-## A pathologically nested or long Cypher expression is now a parse error, not a StackOverflowError (#5851)
-
-A ~2KB `WHERE` clause with about 1000 nested parentheses crashed the parsing thread with a
-`StackOverflowError`: the ANTLR-generated Cypher parser re-enters its `expression` grammar rule, and walks
-its full ~10-level operator-precedence cascade, once per nesting level, so a few thousand levels is enough
-to exhaust the default JVM thread stack. `AbstractServerHttpHandler` catches `Throwable`, so this degraded
-an HTTP request to a 500 rather than killing the worker, but a malformed query should fail with an ordinary
-400, and letting an `Error` unwind through the engine on a path no database state has touched is worth
-avoiding on its own.
-
-Investigation found the same crash from two independent recursion sites, not one. Nested parentheses (and
-equally, list/map literals or function arguments nested inside one another) recurse in the ANTLR-generated
-parser itself. A long *flat* chain - thousands of `OR`'d or string-concatenated terms - does not: the
-grammar's `(OR expression11)*` and `(PLUS expression5)*` productions are quantifier loops, not
-self-referencing rules, so they parse without recursing. They still crashed, one level further down the
-pipeline: `ExpressionRewriter`, the shared visitor every `WHERE` condition is normalized/folded/simplified
-through, walks the resulting deep expression tree recursively.
-
-Both sites are now bounded by the same new setting, `arcadedb.cypher.maxExpressionDepth` (default 200),
-converting either crash into a `CommandParsingException` that names the limit and the setting to raise if a
-legitimate query needs it. Real-world queries essentially never nest this deep, so the default leaves a
-wide margin; the SQL parser was never at risk from the same input because its hand-written recursive-descent
-implementation costs far fewer stack frames per nesting level, not because it enforces a limit of its own.
--- a/docs/release-26.8.1.md
+++ b/docs/release-26.8.1.md
@@ -2369,3 +2369,27 @@ This is **not new** in this release - the previous two-phase walk collected the 
 recorded here because it was never written down, and because it holds for lightweight self-loops too, where there
 is no record to go missing on the second pass. Pinned as an exact count by
 `Issue5760VertexDeleteSelfSideSkipTest.aSelfLoopIsWalkedFromBothListsSoItsDeleteEventFiresTwice`.
+
+## A pathologically nested or long Cypher expression is now a parse error, not a StackOverflowError (#5851)
+
+A ~2KB `WHERE` clause with about 1000 nested parentheses crashed the parsing thread with a
+`StackOverflowError`: the ANTLR-generated Cypher parser re-enters its `expression` grammar rule, and walks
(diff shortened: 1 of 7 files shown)


lvca:
@claude please review this.

lvca:
@claude please review

lvca:
@claude please review

Issue #4267, opened 2026-05-19
Query count(*) return different  field type  depending on whether RemoteDatabase or RemoteGrpcDatabase is used.


mdre:
Version: 26.6.1-SNAPSHOT

In this way return an integer:
```java
//RemoteGrpcServer gs = new RemoteGrpcServer("localhost", 50051, "root", "toortoor", true, List.of());
//RemoteGrpcDatabase txH = new RemoteGrpcDatabase(gs, "localhost", 50051, 2480, "ogm-test", "root", "toortoor");

RemoteDatabase txH = new RemoteDatabase("localhost", 2480, "ogm-test", "root", "toortoor");
txH.begin();
MutableVertex svt1 = txH.newVertex("SimpleVertexEx");
svt1.set("svex", "svt1");
svt1.save();
txH.commit();

txH.begin();
String rid = svt1.getIdentity().toString();
svt1.set("svex", rid);
svt1.save();
txH.commit();

System.out.println("RID: "+rid);
String query = String.format("select count(*) from SimpleVertexEx where svex = '%s'", rid);
System.out.println("query: "+query);
ResultSet rs = txH.query( "sql", query);
Result or = rs.next();

String retVal = or.getPropertyNames().iterator().next();
System.out.println("retval: "+retVal);
Object ret = or.getProperty(retVal);
System.out.println("ret: "+ret+"  > "+ret.getClass().getName());
``` 

but the same using a gRPC connection return a long.

lvca:
Fixed

mdre:
Thanks.

PR #5860, opened 2026-08-06
fix(#5812): make the @props JSON type hint opt-in, off by default

### LINKED ISSUE #5812: [SQL] `@props` system property in results
`ArcadeDB Server v26.8.1 (build 87bdc67f1f0331fa2d07e932a550064c118eae70/1785791531932/UNKNOWN)`

`Running on Linux 7.0.0-28-generic - OpenJDK 64-Bit Server VM 25.0.3`

An unwanted `@props` property appears in the results.

Setup:

```sql
CREATE DOCUMENT TYPE Doc;
CREATE PROPERTY Doc.num SHORT;
INSERT INTO Doc SET num = 0;
```

Send SQL query `SELECT num FROM Doc` via HTTP:

```shell
wget -qO- "http://localhost:2480/api/v1/query/test/sql/SELECT%20num%20FROM%20Doc" --user root --ask-password
```

results in:

```json
{
  "result": [
    {
      "num": 0,
      "@props": "num:2"
    }
  ]
}
```

Furthermore, `SELECT FROM schema:types` also contains `@props`:

```json
{
  "result": [
    {
      "name": "Doc",
      "type": "document",
      "records": 1,
      "buckets": [
        "Doc_0"
      ],
      "bucketSelectionStrategy": "round-robin",
      "needsRepartition": false,
      "parentTypes": [],
      "properties": [
        {
          "id": 1,
          "name": "num",
          "type": "SHORT",
          "default": "<DEFAULT_NOT_SET>",
          "custom": {},
          "@props": "custom:10"
        }
      ],
      "indexes": [],
      "custom": {},
      "@props": "records:3,buckets:9,parentTypes:9,properties:9,indexes:9,custom:10"
    }
  ]
}
```

lvca:
## Summary

Fixes #5812.

The per-column `@props` type hint (added in #4267/#4849) is meant to let a consumer rebuild the exact Java type of a non-element (projection/aggregate) result column - e.g. `count(*)` yielding a `Long` that plain JSON would otherwise collapse to `Integer`. In practice it was unconditionally emitted on every HTTP response, including plain `SELECT` projections and `SELECT FROM schema:types`, regardless of who was asking. Any HTTP client (curl, wget, another language, Studio) got the extra `@props` field whether it needed it or not, exactly as reported in the issue.

Today the only real consumer of this hint is the `RemoteDatabase` Java driver (`json2Result`/`RemoteImmutableDocument`), and even there only for non-element rows - element/document rows already carry `@type` plus the schema, so they never needed the hint.

## Changes

- `JsonSerializer`: new `includeTypeHints` flag (default `false`) gates `@props` emission in `serializeResult`. Building the per-column type-hint string is now skipped entirely when the flag is off, which also avoids the wasted `StringBuilder`/type lookups on the hot HTTP response path.
- `AbstractQueryHandler` / `GetQueryHandler` / `PostCommandHandler`: a new `typeHints` request flag (query parameter on `GET /query`, JSON field on `POST /command`) is threaded through to `JsonSerializer`, decoupled from the existing `serializer` (record/graph/studio/default) selection - `serializer` controls the response *shape*, `typeHints` controls this specific metadata field. Both default to `false`.
- `RemoteHttpComponent`: the Java driver now sends `typeHints: true` alongside the existing `serializer: record`, so its own type-fidelity behavior (e.g. `count(*)` staying a `Long`, `FLOAT`/`SHORT`/temporal columns) is unaffected.
- `map2json`'s existing `@props` behavior (used when an embedded/document `toJSON(true)` is requested, e.g. for schemaless properties) is left untouched - it's a separate, already-narrower code path not implicated by the reported issue, called out here so it isn't mistaken for an oversight.

## Test plan

- New `Issue5812PropsTypeHintHttpIT` reproduces both repros from the issue against a live server: `SELECT num FROM Doc` (SHORT column) and `SELECT FROM schema:types`, verifying `@props` is absent by default and present only with `typeHints=true`/`"typeHints": true`.
- `JsonSerializerTest` updated: existing type-fidelity tests now opt in via `setIncludeTypeHints(true)`; new `propsHintOffByDefault()` pins the default-off behavior for a plain `JsonSerializer`.
- Ran the existing HTTP handler suite (`AbstractQueryHandlerIT`, `PostCommandHandler*`, `HTTPDocumentIT`, `HTTPGraphIT`) and the driver suite (`RemoteDatabaseIT`, `Issue4267CountTypeIT`, `RemoteGraphBatchIT`) - all pass, confirming the driver's type reconstruction (the original reason `@props` exists) is unaffected.
- [x] `mvn compile test-compile` clean on `engine`, `network`, `server`

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
@@ -73,6 +73,11 @@ public class JsonSerializer {
   private boolean  includeVertexEdges        = true;
   private boolean  useVertexEdgeSize         = true;
   private boolean  useCollectionSizeForEdges = true;
+  // Issue #5812: off by default. The @props type hint only matters to a consumer that must rebuild the exact
+  // Java type of a non-element (projection/aggregate) column - today that is the RemoteDatabase Java driver,
+  // which turns this on explicitly for its own requests. A generic HTTP/JSON client (curl, Postman, another
+  // language) never asked for it and should get plain JSON back.
+  private boolean  includeTypeHints          = false;
   private Database database;
 
   JsonSerializer() {
@@ -154,7 +159,7 @@ public class JsonSerializer {
       }
     }
 
-    final StringBuilder propertyTypes = new StringBuilder();
+    final StringBuilder propertyTypes = includeTypeHints ? new StringBuilder() : null;
 
     for (final String propertyName : result.getPropertyNames()) {
       Object value = result.getProperty(propertyName);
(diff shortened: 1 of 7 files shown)


lvca:
@claude please review this.

lvca:
@claude please review

PR #5861, opened 2026-08-06
perf(engine) #5588: lay the vector location index out in primitive arrays (~90 -> ~32 bytes per live vector)

### LINKED ISSUE #5588: Vector index: lay the location index out in primitive arrays (~20 bytes/id instead of ~90)
### Summary

`VectorLocationIndex` stores one boxed entry per live vector, which retains roughly **90 bytes** each once the map and object overheads are counted. The payload it actually carries is ~24 bytes. Laying the same data out in **primitive arrays indexed by vector id** would cost ~20 bytes per id with no per-entry objects at all, taking a 10M-vector index from ~900MB to ~200MB.

Split out of #5568 / PR #5578, where the reviewer and I converged on this as the right way to reduce the footprint.

### Why it matters now

Until #5568 there was a knob for this: `arcadedb.vectorIndex.locationCacheSize` capped the index and let it evict. That turned out not to be a cache bound at all. A location is the only record of which record a vector id belongs to and where its entry sits in the index file, and nothing on disk maps a vector id back to an offset, so an evicted entry could not be recovered and every reader read a missing location as deleted. A cap of 100 over 1000 live vectors made `countEntries()` report 100. The setting is now ignored (#5568).

That was the correct call - a wrong answer is worse than a visible memory requirement - but it means the location index is now unbounded by design, and the only honest way to shrink it is to make each entry cheaper. This issue is that work.

### Where the ~90 bytes goes

Per live vector, on a 64-bit JVM with compressed oops (enumerated in #5516, and now the constant `VectorLocationIndex.APPROX_RETAINED_BYTES_PER_LOCATION`):

| | approx |
|---|---|
| `ConcurrentHashMap` node + table slot | ~40 B |
| boxed `Integer` key (ids exceed the `Integer` cache) | ~16 B |
| `VectorLocation` object | ~32 B |
| `RID` object | ~32 B |
| `ridToVectorIds` reverse-index node + `int[]` | ~56 B |

The payload inside all of that is one `long` offset, one `int` bucket id, one `long` position and two flags.

### Proposed layout

Vector ids are dense and monotonically assigned from a counter, so they index arrays directly:

```java
long[] absoluteFileOffset;   // 8 B
int[]  bucketId;             // 4 B
long[] position;             // 8 B
long[] compactedBits;        // 1 bit per id
long[] presentBits;          // 1 bit per id, replaces the null check
```

~20 B per id, zero per-entry objects, and the GC no longer traces one object graph per indexed vector. This also matches the project's standing preference for arrays of primitives over collections of objects.

### Things the implementation has to get right

- **`RID` is materialized on demand** from `bucketId` / `position`. Most readers only need it to return a result, so the allocation moves from resident to transient - but check the hot loops in the search paths before assuming that is free.
- **The RID reverse index** (`getVectorIdsForRid`, used by `remove()` since #5318) still needs a map, since RIDs are not dense. It can hold primitive int ids instead of boxed ones.
- **Growth.** Arrays are sized by the highest id handed out. A rebuild renumbers nothing (ids are preserved across a compaction), so the arrays must grow amortized and be re-allocated at the size of the live set when a rebuild publishes a replacement.
- **Atomic publication.** #5568 made a rebuild publish a fully populated replacement through a single volatile store, because `countEntries()` and `getStats()` take no lock. Whatever replaces the maps must keep that property: readers must see either the whole old generation or the whole new one, never a half-filled array.
- **Sparse ids.** If a workload ever leaves large gaps in the id space, the arrays waste the gap. Worth measuring against a re-embedding workload before committing to a plain array over an id-indexed paged structure.
- `DeletedIds` (#5516) is already a bitset and is the model to follow.

### Acceptance

- Measured heap per live vector drops from ~90 B to ~20-25 B on a 1M-vector index.
- `VectorLocationIndex.APPROX_RETAINED_BYTES_PER_LOCATION`, `getStats().estimatedLocationIndexBytes` and the 26.8.1 release note are updated to the new figure.
- The #5568 guards still pass: `countEntriesIsNeverTornByAConcurrentRebuild` and `aConfiguredLocationCacheLimitDoesNotDropVectors`.
- No regression in vector search latency - the extra `RID` allocation per result must not show up on the hot path.

lvca:
Closes #5588.

## What

`VectorLocationIndex` held one boxed entry per live vector in a `ConcurrentHashMap<Integer, VectorLocation>` plus a `ConcurrentHashMap<RID, int[]>` reverse index, retaining **~90 bytes per live vector** to carry ~20 bytes of payload. Vector ids are dense and monotonically assigned, so they address arrays directly.

The mapping is now four primitive arrays indexed by vector id - the file offset with its compacted flag packed into one `long`, the RID's bucket id and position, and one presence bit - at **~21 bytes per id plus 8 to 16 per live vector** for the reverse index. A 10M-vector index goes from ~900MB resident to ~320MB, with no per-entry objects at all.

## On the measurement in the issue thread

@tae898 measured, with JFR `jdk.ObjectCount`, `VectorLocation` at 24 B and `RID` at 24 B, both exactly 1:1 with the vector count, and read that as the ~90 B figure being high. `jdk.ObjectCount` reports **per class**, and those are two of the five contributors the constant enumerates - the other three (`ConcurrentHashMap$Node`, the boxed `java.lang.Integer` key, and the `ridToVectorIds` node with its `int[]`) are separate lines in the same histogram. The measurement is sound; ~90 B is not refuted by it. Both process findings in that thread are correct and worth keeping: `asprof -e alloc --live` cannot answer a retained-heap question, and JFR needs no JDK.

This PR stops the number being arguable: `estimatedLocationIndexBytes` is now summed off the arrays the index has actually allocated instead of `live count x a per-entry constant`. It steps **down ~3x** on the same index, which is operator-visible and is in the release note - together with the note that the 3.75x step **up** in 26.8.1 was an accounting correction and not a regression, which is the point @tae898 raised.

## The parts that are load-bearing

**The arrays are chunked (128 ids) and a chunk is released as soon as its last live id is tombstoned.** Not an optimisation: it is what preserves the property #5516 exists for, that residency follows the live vectors and not the ids handed out. A re-embedding workload hands out 9.3M ids for a 4K live set, and a flat id-indexed array would retain ~190MB where the map it replaces retained ~360KB. `aDrainedIdSpaceIsHandedBack` fails on a flat array and on a missing release.

**Publication is release/acquire on the array element itself, at both levels.** A plain element store into an already-published array carries no happens-before edge for a reader that read the enclosing field earlier, and re-publishing the array reference does not create one. On the directory that would mean reading `null` for a chunk that exists, i.e. dropping a just-inserted vector out of every search.

**Three orderings are documented where they are implemented:**
- the offset and its compacted flag share one 64-bit word. `applyReplicatedPageUpdate` re-points a live id at the compacted file, on a replica, with no index lock held, while searches run; split across two arrays a reader pairs the new flag with the stale offset, and a misaligned read of an INT8/BINARY page parses into a well-formed vector of the right dimension that passes every guard and is scored at the wrong distance. `theOffsetAndItsCompactedFlagCannotBeReadFromDifferentGenerations` fails within milliseconds on the split layout;
- a chunk is dropped only after its bits are cleared (asserted), or a reader holding the stale reference resurrects a tombstoned vector;
- `clear()` zeroes the live chunks before publishing the empty directory - it is the one path that reuses ids.

**The reverse index stores no keys.** A candidate slot is verified by reading the id's RID back out of the location arrays, which makes it unable to produce a wrong id (only to miss one) and makes a tombstoned id structurally unreachable rather than something callers re-check. Nothing is ever removed, because a dead slot is already unreachable; dead slots are reclaimed at a rehash sized from the **live** count and never from the current capacity. Sizing it the other way is what would let dead entries grow the table without bound in the #5516 workload.

**No new search allocation.** Every reader moved onto accessors that answer one field - `isLive`, a packed `getOffsetAndFlag`, `isLocationOf` - so the liveness filter each search applies per traversed graph ordinal still costs one presence bit. `getLocation` stays as a materializing convenience.

**The graph-build snapshot is the same structure now.** It was a `Map<Integer, VectorLocation>` copy of the whole live set, held for the duration of every rebuild, on top of the live index and at the same per-vector cost: a transient ~900MB spike on a 10M index, now ~200MB. `ArcadePageVectorValues` is built through `forSearch(...)` / `forGraphBuild(...)` factories instead of six constructors, since the two roles now take the same argument types.

## Suggested alternative to the issue's proposal

The issue proposed plain arrays and flagged sparse ids as something to measure before committing. Measuring is not needed: the #5516 workload is the sparse case and a flat array regresses it by three orders of magnitude, so chunking with release-on-empty is a precondition, not a variant. The issue also targeted ~20 B by counting the location store alone; the honest all-in figure is ~32 B because the reverse index still needs a table. Packing the RID into a single `long` would take the store from 21 to 17 B/id but needs an overflow escape hatch for large buckets - not worth the risk inside this rewrite.

The structural fix for the one weakness left - a partially drained chunk retaining 2.7 KB per live id it still holds - is to renumber ids during `rewriteDataFileWithLiveEntries`, which already reissues every offset and every tombstone and is the one place a dense renumbering is safe. It has to move `ordinalToVectorId` and the persisted tombstone set in lockstep, so it is its own PR; a `TODO` in the class points at it.

## Verification

New `Issue5588PrimitiveLocationIndexTest` (11 tests), each written against a specific wrong implementation rather than against the happy path: chunk release, payload atomicity under a concurrent rewriter, reverse-index survival across constant rehashing, ascending traversal and the `SORTED` characteristic (load-bearing - `ordinalToVectorId` is read with `Arrays.binarySearch`), the offset range and its no-page-entry sentinel, tombstoning a non-resident id, a model-based property test over random operation sequences including `clear()` and id reuse, and the retained-heap acceptance figure. The chunk-release test was mutation-checked by disabling the release. What is deliberately NOT tested is the release/acquire on the presence word: dropping it is invisible on x86, so a green test would only be evidence about the runner.

Plus a new guard in `LSMVectorIndexTombstoneMemoryTest` that the id sequence survives an index where every vector was deleted, across a reopen - the reopen path derives the high-water mark from the location index's own sequence, and reusing an id a tombstone still refers to would bury the new vector with the old one.

Green: 1153 tests in `com.arcadedb.index.**`, 516 in `com.arcadedb.schema`/`com.arcadedb.engine`/`LowRamProfileTest`, the server `Issue5559LocationCacheSizeHttpIT`, and a full reactor build.

https://claude.ai/code/session_0131GDFM4QGFC5URdReusH4A


lvca:
@claude please review this

lvca:
> **One ask before merge**: I couldn't run the reactor build in this environment. The description reports green - please just confirm CI is green on the final commit.

Confirmed, with one caveat that matters: **`Java CI - test` is red on this PR, and it is red on `main` for the same reason.**

The three failing jobs on run [31131778011](https://github.com/ArcadeData/arcadedb/actions/runs/31131778011) are `unit-tests`, `ha-integration-tests` and `integration-tests`. The last four `Java CI - test` runs on `main` fail the **same three jobs**, on commits that are not in this branch:

| run | branch | head | failing jobs |
|---|---|---|---|
| [31131326836](https://github.com/ArcadeData/arcadedb/actions/runs/31131326836) | main | `2724575` | unit-tests, ha-integration-tests, integration-tests |
| [31131107983](https://github.com/ArcadeData/arcadedb/actions/runs/31131107983) | main | `a759e16` | (failure) |
| [31130786349](https://github.com/ArcadeData/arcadedb/actions/runs/31130786349) | main | `64cb88f` | (failure) |
| [31130490721](https://github.com/ArcadeData/arcadedb/actions/runs/31130490721) | main | `c67740b` | (failure) |

Everything else on this PR is green: `build-and-package`, `slow-unit-tests`, `coverage-report`, all six e2e job sets, `builder-tests`, Codacy (0 new issues) and all six CodeQL analyses.

Locally on the merged branch: **1162 tests in `com.arcadedb.index.**` green**, plus `com.arcadedb.schema` / `com.arcadedb.engine` / `LowRamProfileTest` (516), the server `Issue5559LocationCacheSizeHttpIT`, and a full reactor `install`. That covers the call sites you flagged - the `ArcadePageVectorValues` factory migration and the `LiveVectorBitsFilter` signature change - since every one of them is exercised by the vector suite.

Note the branch also carries a merge of `main` that landed after the first review: #5786 withdrew `GroupedRIDBitsFilter` and rewrote the grouped search, so that file is gone here and the two `getLocation` call sites its new grouped path introduced were moved onto `getRid()`.

### The two review points, addressed

- **`getBucketId(int)` / `getPosition(int)` read as dead public API** - correct, no engine caller. Now package-private (`45efc707`), with the reason stated: the readers either compare a RID via `isLocationOf(...)` or need the object via `getRid(...)`, and these two survive only so the model test can read the words straight out of the arrays and check `getRid()` against them rather than against itself.
- **`getActiveCount()` / `estimatedRetainedBytes()` walking the directory** - documented on both, pointing at `size()` as the O(1) answer. No regression either way: the `ConcurrentHashMap` versions were O(live vectors), these are O(allocated chunks).

Also from the earlier review pass: `rehash()` no longer re-resolves the chunk it is already standing in, and the "no engine path re-points a live vector id" precondition is now **counted** rather than asserted - an assert would have fired on the tests that exist to prove that branch does the safe thing, so `inPlaceRidRewriteCount()` must stay zero instead, pinned by `LSMVectorIndexTombstoneMemoryTest` across the re-embedding cycles that would trip it first.

lvca:
Thanks - taking the four points in order.

**1. In-place RID-rewrite window.** Agreed on the characterisation, and agreed the guard is only as good as the workload that trips it. Two things make me comfortable leaving it as a counter rather than escalating it: the branch is the *safe* answer (the alternative is a torn `(bucketId, position)` pair, which is how a delete of record A tombstones the vector of record B - strictly worse than a transient miss), and the pin is on `updateCyclesDoNotAccumulateStaleLocations`, which is 10 re-embedding cycles over 100 vertices - the exact shape that would tempt someone to reuse an id instead of minting a new one. An `assert` was the first thing I tried and it is wrong here: it fires on `rewritingALiveIdsRidInPlaceIsCountedAndStillCorrect`, the test that exists to prove the branch still leaves the index consistent.

**2. `estimatedRetainedBytes()` hardcodes a compressed-oops layout.** Fair, and the "10M vectors is exactly where someone runs a >32GB heap" point is the right one. I measured the drift instead of documenting it vaguely (`ccc3290b`): above the threshold, array headers go 16 -> 24 bytes and the `Chunk` object 40 -> 56, so a chunk is **2728 instead of 2680 - +1.8%**. The layout is 97% primitive payload, whose size does not depend on the oop width, so the error is smaller than the rounding already in "~21 bytes per id". That is now stated on both `CHUNK_RETAINED_BYTES` and `estimatedRetainedBytes()`. Reading the VM option back to correct for 1.8% would cost more clarity than it buys.

**3. `VectorLocation.deleted` is vestigial.** Correct, and deliberate - the javadoc on the field says so. It stays because a location read straight off a page can carry the flag, and it is the definition every `loc != null && !loc.deleted` reader is written against.

**4. Non-hot-path rescans.** Documented on both, pointing at `size()` as the O(1) query-path answer.

### On the slow-runner concern

Worth being concrete rather than hopeful, so I measured every test in the new class:

| test | time | its timeout | headroom |
|---|---|---|---|
| `theOffsetAndItsCompactedFlagCannotBeReadFromDifferentGenerations` | 0.546s | 30s join (0.5s of it is a fixed sleep) | 55x |
| `theIndexMatchesAReferenceModelUnderRandomOperations` | 0.509s | none | - |
| `concurrentAddsAndReadsAgreeOnTheFinalState` | 0.277s | 120s join | **430x** |
| `aDrainedIdSpaceIsHandedBack` | 0.093s | none | - |
| `theReverseIndexNeverLosesALiveEntryAcrossARehash` | 0.067s | 60s join | **900x** |
| `aVectorIsVisibleToAConcurrentReaderAsSoonAsItIsAdded` | 0.037s | 60s join | 1600x |
| the other six | <= 0.011s | none | - |

Every timeout has two to three orders of magnitude of headroom, and none of the assertions is timing-derived - the concurrency tests assert invariants (no torn read, the pinned id never disappears, no id answers twice), not durations. The one count-based assertion, `reads > 1000` in the torn-read test, guards against the loop being vacuous rather than measuring throughput; it observes millions locally.

### On CI

`Java CI - test` is red on this branch and **red on `main` for the same reason** - same three jobs (`unit-tests`, `ha-integration-tests`, `integration-tests`), on the last four `main` runs, on commits not in this branch. Detail and run links in [the previous comment](https://github.com/ArcadeData/arcadedb/pull/5861#issuecomment-5211394735). Everything else is green here, including `build-and-package`, `slow-unit-tests`, `coverage-report`, all six e2e sets, Codacy (0 new issues) and all six CodeQL analyses.

lvca:
Two of the four observations produced changes, and one of them found something real that none of the earlier rounds caught.

### The `SORTED` observation is not correct - and pinning that was worth it

> `IdSpliterator` reports `SORTED` for an `IntStream`; `IntStream.sorted()` does not consult that characteristic the way `Stream<T>.sorted(cmp)` does, so the "no-op" claim is a touch optimistic.

It does consult it. `SortedOps.OfInt.opWrapSink` returns the sink unchanged when `StreamOpFlag.SORTED.isKnown(flags)`, and the flag propagates from a spliterator whose `getComparator()` returns null (`StreamOpFlag.fromCharacteristics` only strips SORTED when a non-natural comparator is present). Measured rather than argued - a spliterator that claims `SORTED` and yields descending values:

```
plain sorted():         [5, 4, 3, 2, 1]
filter().sorted():      [5, 4, 3, 2, 1]
SORTED survives filter: true
```

The sort really is elided, including through the `filter()` the call sites apply first. That matters beyond the micro-optimisation: it is what makes declaring `SORTED` **load-bearing rather than decorative**. If the traversal were ever not ascending, `.sorted()` would not fix it, `ordinalToVectorId` would go unsorted, and the `Arrays.binarySearch` over it would quietly stop finding ordinals - fewer results from every allow-list-filtered search, nothing failing. So the correction prompted a test (`ce5fc97f`) rather than a doc change; the behaviour is now pinned instead of being re-litigated each round.

### The follow-up issue reference was dangling - fixed

Chasing the `#5860` reference in the class javadoc found that **#5860 is an unrelated, already-merged PR** (`fix(#5812): make the @props JSON type hint opt-in`), not the renumbering follow-up. The issue had never actually been filed; the reference was wrong from the first commit and five review rounds read past it. Now filed as **#5870** with the three things that have to move in lockstep with a renumbering (`ordinalToVectorId`, the persisted tombstone set, `nextId`), and the javadoc corrected. Thanks - that one was only findable by following the link.

### The contract change is now in the release note

Agreed this deserves to be visible to embedding code, not just the javadoc. Added:

> `VectorLocationIndex.getVectorIdsForRid` returns only LIVE ids now, in ascending order. It used to return tombstoned ids as well and require the caller to re-check the location's `deleted` flag. Nothing in the engine is affected - both callers re-verified anyway - but the contract of a public method inverted, so embedding code that relied on seeing tombstoned ids has to look elsewhere for them (`isDeleted(int)`).

### The other two

- **`size()` vs `getActiveCount()` transiently disagreeing during an in-place RID rewrite** - correct, and correctly characterised as not-a-bug. Worth adding that the window is unreachable on every engine path, which is what `inPlaceRidRewriteCount() == 0` pins, so the disagreement needs a caller that does not exist before it can be observed at all.
- **`VectorLocation.deleted` vestigial** - kept deliberately and documented on the field; it is the definition every `loc != null && !loc.deleted` page-read caller is written against.

300 vector tests green on `ce5fc97f`. `Java CI - test` remains red for the pre-existing `main` failure ([evidence](https://github.com/ArcadeData/arcadedb/pull/5861#issuecomment-5211394735)).

lvca:
Point 1 was a real defect - thank you. The other three are fixed too.

### 1. The delta-scan guard really had been dropped

You read it exactly right: an earlier edit replaced the guard with the comment explaining it and deleted the statement, leaving a block that ends *"It stays because it is what the line says it does, at the price of one bit"* above no line at all. Not a doc nit - the guard was gone. Restored in `3a9826b4`.

The comment now separates the two independent reasons it cannot fire from the reason it stays, because they were tangled together and that is what made the block easy to mis-edit in the first place:

- the flag it used to read (`loc.deleted`) has been permanently false since #5516, so the guard did nothing even before this PR;
- `remove()` purges the delta buffer of every entry for the RID it deletes, so no delta entry survives its own deletion on any path a test can drive - I tried, twice, and both tests were vacuous: forcing a tombstone in behind the buffer only makes the next rebuild republish the live entry the pages still carry;
- it stays anyway because the delta buffer and the tombstone set are maintained by different code paths, and one bit is not a price worth negotiating over.

### 2. `inPlaceRidRewrites` is now `volatile`

Agreed, and I took the volatile rather than the note. It is written under `writeLock` and incremented approximately never, so the read costs nothing, and a diagnostic that silently under-reports is worse than no diagnostic - the whole point of that counter is to be believed when it says zero.

### 3. `getVectorIdsForRid` allocation

Fair scoping correction. Its javadoc now says so explicitly - it allocates, it is a resolution step, and its callers (the delete path and the allow-list walk) are already resolving records. The "no new search allocation" property is about the per-ordinal readers and should not be read as covering it.

### 4. jcstress

Agreed and out of scope. Noted for whoever wants the presence-word ordering covered properly rather than by a green-on-x86 test.

### Minor: the figure living in five places

Real drift risk. Where javadoc can interpolate, it now does - `{@value VectorLocationIndex#APPROX_RETAINED_BYTES_PER_LOCATION}` in `LSMVectorIndex` and `TypeLSMVectorIndexBuilder` instead of a literal `~32`; `LSMVectorIndexMetadata` already concatenated the constant. What is left is `GlobalConfiguration`'s runtime description text and `engine/CLAUDE.md`, neither of which can reference a Java constant, plus the release note, which is a point-in-time record and should keep its number.

300 vector tests green on `3a9826b4`.

PR #5869, opened 2026-08-07
fix(vector): drop the atomic counter from the distance path, widen the graph-build pool, fix the build progress meter (#5577)

### LINKED ISSUE #5577: Dense build: 93% of DEEP-10M build time is JVector graph construction, and the build pool is availableProcessors()/2 - is the halving deliberate?
A question rather than a defect report, with the measurement that prompted it.

## Where the dense build time actually goes

Our paper carries a dense build-time loss (DEEP-10M, 9.99M vectors, 96d, `maxConnections=32`, `beamWidth=100`): **2,693 s against 142 s for Milvus and 516 s for Qdrant**. We had been treating that as an ArcadeDB storage-path cost and were about to go profiling. We did not need to: your own log timestamps the boundary, so the split was free to compute from a run already on disk.

| phase | duration | share |
|---|---|---|
| ingest of all 9.99M vectors into ArcadeDB | **191 s** | 6.6% |
| `Building JVector graph index` to `Built graph` | **2,722 s** | **93.4%** |

(2,913 s total on the 16 GiB-heap run; the two markers are `LSMVectorIndex` `Building graph with 9990000 vectors` and `Built graph`.)

So the record path is not the problem. **ArcadeDB ingests 10M vectors at roughly 52,000/s and that is 7% of the build**; essentially all of the rest is graph construction inside JVector. We have corrected our own framing accordingly, and the paper now attributes that row to the library rather than to your storage layer.

## The question

The one ArcadeDB-side lever we can see inside that 93% is the build pool:

```java
// LSMVectorIndex.getOrCreateGraphBuildPool()
final int cores = Math.max(1, Runtime.getRuntime().availableProcessors() / 2);
pool = new ForkJoinPool(cores);
```

On our 12-CPU cpuset that is **6 threads while 6 CPUs sit idle**, for the phase that is 93% of the build. **Is the halving deliberate?** The reasons we can think of are good ones: leaving headroom so an online index build does not starve concurrent query traffic, or a build that is memory-bandwidth-bound where extra threads buy nothing. If either is the reason, that is the answer and we will stop here.

## What we are running either way

An A/B that doubles the pool without patching anything, using `-XX:ActiveProcessorCount=24` so the JVM reports 24 and the pool becomes 12 on the same 12 real CPUs. Same wheel, same data, same `maxConnections=32`. Container CPU is sampled through the graph phase, since 6 saturated threads read about 600% and 12 read about 1200%, which checks that the pool size is what changed rather than inferring it from elapsed time. There is a guard that aborts if the flag does not actually move `availableProcessors()`, because an unread knob would make both arms identical and read as a tidy null result.

If forcing 12 threads cuts the graph phase materially, that is a concrete report about the heuristic and we will bring the numbers. If it does not, the halving costs nothing, and the remaining gap is a JVector question rather than yours. We will post the result whichever way it goes.

**Not asking for anything yet.** If the `/2` is intentional it would be worth a sentence in the Javadoc saying why, since the next person to look at a 45-minute graph build on a half-idle box will ask exactly this.

lvca:
Fixes #5577.

Three findings from @tae898's measured report on a DEEP-10M (9.99M x 96d) dense build. Only one of them is what the issue was opened about, and the third turns out to be the answer to the question he could not answer.

## 1. The hit/miss counters cost ~26% of the build

`VectorCache.get()` incremented one of two process-wide `LongAdder`s on every lookup, and a lookup there backs a single distance evaluation - the class Javadoc says as much. The profile attributed **26.4% of the entire graph build to `LongAdder.add`**, every sample reached through `VectorCache.get`, against 42.1% for the SIMD distance the lookups exist to feed. That ratio is unsurprising once stated plainly: a locked compare-and-swap and a 96-dimension SIMD distance are the same order of magnitude, and the CAS gets worse as the build pool widens while the distance does not.

`LongAdder` is the right structure for a contended counter; the mistake was having *any* atomic counter on that path. The counters are now striped one pair per thread onto its own cache line and incremented with plain reads and writes. Opaque access keeps each 64-bit value indivisible (a reader never sees a torn count) without asking for ordering or fencing, so on x86 and AArch64 it compiles to the plain load and store.

The totals become approximate: two threads landing on the same stripe can lose an increment, so a count can come out low but never high. That is the right trade for a diagnostic counter whose only consumers are `getStats()` and dashboards, and `LongAdder.sum()` was never an atomic snapshot either. `VectorCacheCounterBenchmark` (`@Tag("benchmark")`) measures 2-5x less time per lookup depending on thread count.

Also removes the two dead `AtomicLong` cache counters in `LSMVectorIndexMetrics` - nothing incremented them and `getStats()` overwrote their values from the real cache.

## 2. The pool halving was not deliberate

It arrived in d22882cba, "fix: use own pool for vector index rebuild" - the pool exists so a build can be cancelled on close (JVector's builder only observes an interrupt on its workers), and the `/2` came along for the ride. The A/B in the issue priced it at **17.1% of the whole build with recall unchanged** (0.9502 vs 0.9526).

Automatic width is now the core count minus one, overridable with the new `arcadedb.vectorIndex.graphBuildParallelism`. The core left free is deliberate and is the only reason not to take them all: a rebuild can fire on a live index at any time and must not be able to occupy every core the request, I/O and GC threads need. The effective width is reported as `graphBuildParallelism` in `getStats()`.

## 3. There is no silent phase - the meter was saturating

The report described a phase worth ~53% of the build that emitted nothing, ran at full CPU, and did not respond to threads (1407s on 6 threads vs 1442s on 12, while the phase next to it scaled 1.92x). Three rounds of profiling went into explaining it.

It does not exist. The monitor polled JVector's `getIdUpperBound()`, which is "highest node id touched so far + 1". Insertion runs as `IntStream.range(0, size).parallel()`, and Java's `ForEachTask.compute()` forks the left half and descends into the right - so the submitting worker processes the **top** leaf of the range first. The meter therefore pins at 100% after one worker has finished roughly `1/leaves` of the corpus, where `leaves ~ 4 x parallelism`.

That predicts the numbers exactly. The first progress line of the 10M build read 9,060,381 - the *start* of that top leaf - and the meter crept to 9,990,000 as one worker walked it. Doubling the pool doubles the leaf count and halves that leaf, which is the whole of the "1.92x insertion scaling": 1023s -> 534s. Summing both halves, JVector's real scaling on that A/B is **1.23x** (2430s -> 1976s), and the 17.1% whole-build win follows from that, not from a 1.92x that never happened.

The build now drives JVector's insertion and its `cleanup()` pass itself - the same two steps `build()` performs, on the same pool - and counts insertions that have actually returned. Consequences:

- `processedNodes` means what it says, and `processedNodes + insertsInProgress` can no longer exceed the corpus size. That is the invariant the old meter broke (`9990000/9990000 (vector accesses=9990006)` in the report) and what the regression test asserts;
- the boundary between the two phases is logged with the elapsed time of each, so the post-insertion pass is measurable instead of inferable from an absence of output - the log line @tae898 asked for twice;
- `GraphBuildCallback` gains an `optimizing` phase for that pass, documented as *not* a quick finalisation.

## Tests

- `LSMVectorIndexGraphBuildProgressTest` - the progress invariant and the pool width. Verified discriminating: reverting the meter to `getIdUpperBound()` fails it with `vector accesses 4002 of 4000 nodes`, the same signature as the report. It carries an explicit anti-vacuous guard requiring a mid-insertion sample, so it cannot pass by only seeing the boundary samples.
- `VectorCacheTest` - counters stay exact single-threaded, never over-count under 8 threads, and `resetStats()`.
- `VectorCacheCounterBenchmark` - `@Tag("benchmark")`, reports the per-lookup cost of both counter shapes.
- `mvn -pl engine test -Dtest='com.arcadedb.index.**'`: **1018 tests, 0 failures**.


lvca:
## Review round 4: two fixed, one deliberately not shipped (and what it found)

**1. Phase-ordering race — fixed.** Real bug, thanks. The final `building` sample is now pushed and the phase flipped inside one `progressLock` block, so the transition is atomic with respect to the monitor and a live progress bar can no longer see `optimizing` then `building`.

**4. "Exact on the build path" — fixed.** You are right that the wording overstated it: thread ids are JVM-global and monotonic over every thread the process has ever created, so even a pool with fewer workers than `STRIPES` can collide modulo the mask. Reworded to best-effort everywhere, with the reason.

**2. JVector coupling.** Agreed, and left as it is - flagged in both the code and the test, which is the most that can be done short of an equivalence test against a private method.

## 3. The cancellation test: I wrote it, it found something, and I am not shipping it

I wrote exactly the test you asked for (start a large build on a worker, `db.close()`, assert prompt termination), with a guard asserting the build was genuinely mid-flight when the close landed so it could not pass vacuously.

**It fails, reproducibly but not deterministically.** Two runs of the same class back to back: one green, one with the build thread still alive 60 s after `db.close()`, parked on `ForkJoinTask.join()`. So the property the pool exists for does not actually hold today.

**This is pre-existing, not something this PR introduced.** The old code is `simdExecutor.submit(<parallel forEach>).join()` inside `GraphIndexBuilder.build()`; the new code is `buildPool.submit(<parallel forEach>).join()`. Same pool object, same JDK machinery, and `shutdownNow()` cannot distinguish the two lambdas. The mechanism is the standard one: `shutdownNow()` sets STOP and drains queues, and a task that is neither completed nor cancelled leaves an external joiner parked forever. I tried to A/B it against `builder.build(vectors)` directly and could not get a non-vacuous run of the reverted path, so I am resting on the structural argument rather than claiming a measurement I do not have.

Shipping a test that fails on roughly half of runs would be worse than shipping no test, so it is out of this PR. I am filing the defect separately: `releaseBackgroundResources()` cannot reliably abort an in-flight graph build, which also means a database close can leave a thread parked. Fixing it properly means making the join observe pool termination (cancel the task and unblock the joiner), which touches the close path of every vector index and deserves its own change and its own testing rather than being folded in here.

What did come out of the attempt and is in this PR: each test in `LSMVectorIndexGraphBuildProgressTest` now gets its own database path. They build the same index shape back to back and sharing one path made them interfere - that was the difference between the 1.3 s green run and the 66 s red one while I was chasing this.

**Nits not taken:** `computeGraphBuildParallelism()` on every `getStats()` is deliberate, so the stat reports the width the *next* build would use rather than a stale live one; it is two field reads. The per-instance `counters` array is at most ~8 KB on a 128-core box, against one search cache per index.

PR #5874, opened 2026-08-07
fix(engine) #5873: keep the graph search's answer instead of replacing it with a resume that returns nothing

### LINKED ISSUE #5873: LSM_VECTOR: the adaptive resume in findNeighborsFromVector discards the first pass, turning a good partial answer into a forced brute-force scan
Found while reviewing #5786 / #5761. Independent of that fix - it is on the plain k-NN path, not the grouped one.

### What happens

`LSMVectorIndex.findNeighborsFromVector`, adaptive-`efSearch` branch (`LSMVectorIndex.java:3953` on `c67740ba9`):

```java
final SearchResult firstPass = searcher.search(ssp, k, initialEfSearch, 0.0f, 0.0f, bitsFilter);
if (firstPass.getNodes().length < k && graphSize >= k) {
  // Graph has enough nodes but beam search found too few - widen the beam
  searchResult = searcher.resume(k, Math.max(k * 10, 100));
} else {
  searchResult = firstPass;
}
```

`searchResult = searcher.resume(...)` **replaces** `firstPass` instead of accumulating onto it. The rows the first pass found are dropped on the floor, and `resume` in this branch can only ever return zero of them back.

### Why resume always returns nothing here

The branch is gated on `firstPass.getNodes().length < k`. In jvector 4.0.0-rc.9:

- `firstPass.getNodes().length == approximateResults.size()` after the pop, and the branch condition says that is `< k <= rerankK`.
- `GraphSearcher.stopSearch` can only break the loop when `approximateResults.size() >= rerankK`. Since the queue never reached `rerankK`, `stopSearch` never fired, so `searchOneLayer` exited on `candidates.size() == 0` - the candidate queue is **dry**.
- `resume` calls `searchLayer0`, which pushes `evictedResults` back onto `candidates` and re-enters `searchOneLayer`. `evictedResults` is empty too (`addTopCandidate` only evicts once `approximateResults` is full at `rerankK`, and `reranking`'s overflow loop needs `size > topK`, also false).
- So `candidates` is still empty, `searchOneLayer` returns immediately, and `reranking` hands back a `SearchResult` with zero nodes.

The `resume` call is therefore pure waste in the only branch that makes it: it cannot widen anything, because the beam was not what ran out.

### Reproduction

1500 vectors, 8 dims, COSINE, rebuilds frozen (`VECTOR_INDEX_MUTATIONS_BEFORE_REBUILD=1_000_000`, `VECTOR_INDEX_REBUILD_GRAPH_RATIO=0f`, `VECTOR_INDEX_INACTIVITY_REBUILD_TIMEOUT_MS=0`) so the deleted vectors stay in the graph as tombstones. Delete all but 4, then `findNeighborsFromVector(query, 10)`.

Instrumenting the branch:

```
PROBE branch fired: firstPass=4 resumeReturned=0 k=10 graphSize=1500
```

and the search's own metrics:

```
live=4 k=10 -> hits=4, bruteForceScans 0->1
```

### Impact

**The answer stays correct** - the brute-force fallback at the end of the method sees an empty `results` and rebuilds it, so the caller gets the right 4 rows. The costs are:

- **A guaranteed full scan of every ordinal**, every time this branch fires, in a case where the graph had *already produced the correct answer* and the code threw it away. On a large index that scan is the most expensive thing this index does (measured at 26 ms vs 95 us on 50k in #5748).
- **`bruteForceScans` over-counts.** The metric exists to say "the graph is degraded, consider rebuilding" (#5558). Here it fires on a shortfall the code manufactured, so the signal points at the wrong thing.
- A `WARNING` log per occurrence saying the graph may need rebuilding, when it may not.

It is rare - measured at 0 firings per 2000 queries on a healthy 50k index under #5558 - which is why it has gone unnoticed. It needs a tombstone-heavy or otherwise sparsely-reachable graph to trigger.

### Suggested fix

Accumulate rather than replace, the same way the grouped path now does after #5786: keep the first pass's nodes and append what `resume` adds. Given the analysis above, the honest version is probably to drop the `resume` call from this branch altogether - a dry candidate queue is exactly the condition the brute-force fallback exists for, and going straight there would skip a pass that provably returns nothing. Either way the first pass's rows should reach the caller, so the fallback is consulted only for the genuine shortfall rather than for the whole answer.

Worth a regression test pinning both halves: that the branch's answer contains the first pass's rows, and that `bruteForceScans` does not increment when the graph search already found everything that is reachable.

lvca:
Fixes #5873.

### The defect

The adaptive-`efSearch` branch of `findNeighborsFromVector` answered a short first pass with `searchResult = searcher.resume(...)`, which replaces rather than adds to it - and the resume could not return anything, so the assignment threw the answer away. The brute-force fallback then walked every ordinal in the index to reconstruct what the graph had already found.

### Why resume was always empty

The branch is gated on `firstPass.getNodes().length < k`. JVector's `searchOneLayer` breaks early only once its result queue holds `rerankK` candidates, and `rerankK >= k` here, so a first pass that came back short cannot have stopped early: it ran its candidate queue dry, having expanded every node reachable from the entry point. `resume()` pushes the (empty) evicted pile back onto that same empty queue and returns zero nodes.

Width was never what ran out either, so the wider fresh search the comment was reaching for would have visited exactly the same nodes. The second pass is therefore removed rather than repaired.

### What replaces it

Nothing new. What is genuinely unreached at that point is whatever the graph cannot walk to, and the only thing that finds it is the brute-force scan - already there, already gated on the shortfall, and now starting from the rows the graph did find instead of from nothing. On the issue's reproduction (1500 vectors, all but four deleted, rebuilds frozen, k=10) the scan stops running entirely: the four survivors are the whole live population, so the answer already meets `expectedResults`.

This also stops `bruteForceScans` counting a shortfall the code manufactured. The metric exists to say the graph is degraded and needs rebuilding; here the graph had done its job.

### A test that was passing for the wrong reason

`LSMVectorIndexBruteForceScanTest.bruteForceFallbackPairsRidWithItsOwnScore` was reaching the fallback *only* through this defect - 50 full scans of 1200 ordinals per run, to recover 50 survivors the graph search had already returned. Every substantive assertion in it still passes; only the "the fallback WARNING must fire" one does not, and it is inverted here into a second regression test for #5873 on a different fixture.

With a healthy graph the fallback can now only fire on a genuinely degraded graph or a narrow `allowedRIDs` list, so that test's lost coverage is replaced by a new sibling taking the allow-list route (#5748) end-to-end, re-checking the #4581 RID-to-score pairing through a real query rather than by reflection.

### Verification

`Issue5873AdaptiveResumeFirstPassTest` - survivors returned with no scan, the kept answer still sorted and duplicate-free, a fully live index unaffected; preconditions assert the branch's own gate (graph >= k nodes, live < k) so it cannot pass vacuously. Mutation-checked by restoring the old branch: both #5873 assertions go red.

553 tests across `com.arcadedb.index.vector` and `com.arcadedb.function.sql.vector` green, plus `Issue5639IndexMetadataKeysTest` and `SQLVectorDatabaseFunctionsTest`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -3937,9 +3937,20 @@ public class LSMVectorIndex implements Index, IndexInternal {
             final int effectiveEfSearch = Math.max(k, metadata.efSearch);
             searchResult = searcher.search(ssp, k, effectiveEfSearch, 0.0f, 0.0f, bitsFilter);
           } else {
-            // Adaptive efSearch with resume for insufficient results (issue #3722).
-            // Both small and large graphs use the same two-pass strategy: search first,
-            // resume with wider beam if too few results are returned.
+            // Adaptive efSearch (issue #3722): size the beam to the graph and keep whatever it finds.
+            //
+            // There used to be a second pass here - `if (firstPass.getNodes().length < k) searchResult =
+            // searcher.resume(...)` - meant to widen the beam when the search came up short. It could only ever
+            // make things worse, and issue #5873 is what it cost. A short first pass means the result queue never
+            // reached rerankK, and JVector's searchOneLayer breaks early *only* once the queue is that full, so the
+            // loop must instead have run its candidate queue dry - having expanded every node reachable from the
+            // entry point. resume() pushes the (empty) evicted pile back onto that same empty queue and returns
+            // zero nodes, so the assignment threw the first pass away and handed the caller nothing. Width was
+            // never what ran out either, so a wider fresh search would have visited exactly the same nodes.
+            //
+            // What is actually left unreached at that point is whatever the graph cannot walk to, and the only
+            // thing that finds it is the brute-force scan below - which is already there, already gated on the
+            // shortfall, and now starts from the rows the graph did find instead of from nothing.
             final int graphSize = graphIndex.size();
             final int initialEfSearch;
             if (graphSize < 10_000)
@@ -3947,13 +3958,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
             else
               initialEfSearch = Math.max(k * 2, 20);
 
-            final SearchResult firstPass = searcher.search(ssp, k, initialEfSearch, 0.0f, 0.0f, bitsFilter);
-            if (firstPass.getNodes().length < k && graphSize >= k) {
-              // Graph has enough nodes but beam search found too few - widen the beam
-              searchResult = searcher.resume(k, Math.max(k * 10, 100));
-            } else {
-              searchResult = firstPass;
-            }
+            searchResult = searcher.search(ssp, k, initialEfSearch, 0.0f, 0.0f, bitsFilter);
           }
         } finally {
           pool.release(searcher, pooledGraph, poolEpoch);
--- a/engine/src/test/java/com/arcadedb/index/vector/Issue5873AdaptiveResumeFirstPassTest.java
+++ b/engine/src/test/java/com/arcadedb/index/vector/Issue5873AdaptiveResumeFirstPassTest.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.index.vector;
+
+import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.TestHelper;
+import com.arcadedb.database.RID;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.utility.Pair;
+
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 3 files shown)


lvca:
CI attribution for this PR. Short version: nothing red here is caused by this diff, and `main` is red on the same jobs.

**`codecov/patch`: 100.00% of diff hit.** Every changed line is covered.

**`unit-tests`** - one failure, `CreatePropertyStatementExecutionTest.createHiddenProperty`. Identical on `main`, and I reproduced it on a clean `main` checkout locally with none of this applied. Pre-existing.

**`integration-tests`** - three failures. `GetServerSettingsToolRedactionIT.clusterTokenIsRedactedInMcpSetServerSetting` and `PostgresWJdbcIT.parsingErrorMgmt` fail identically on `main` at this PR's base and on the three runs before it.

The third, `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites`, is new to this branch and passed on 8 consecutive `main` runs, so I did not take it on trust. It cannot be this diff: the entire main-code change is two adjacent hunks inside `LSMVectorIndex.findNeighborsFromVector`, and that test creates no index at all - the only `vector` references anywhere in the bolt test module are `SHOW VECTOR INDEXES` against an index-less database in a different class - so the method is never invoked in that JVM. The failure is a Raft schema-replication race in `checkDatabasesAreIdentical` teardown, and it reports `DB1 6 <> DB2 5` on one run and `DB1 5 <> DB2 6` on the next, i.e. whichever server applied first. On the re-run the test body passed and only the teardown comparator failed.

**`ha-integration-tests`** - failing on `main` too. The membership is different every time: `main` had five, this branch's first run had one (a subset), the re-run had three that are **disjoint** from that one. That is what these multi-server ITs look like under runner contention, not a regression.

**`codecov/project`** (-0.36%) - a ten-line diff cannot move a whole-project number by 0.36%, and `codecov/patch` is 100%. It tracks the partial coverage uploaded by the erroring HA/integration jobs above.

**Review:** the `claude-review` pass came back LGTM with no blockers. Of its three minor notes, two do not need action - the survivor-reachability assumption is already pinned by the no-scan assertion (if it broke, the test goes red), and the "possibly slow test" condition is false at 0.301 s for all three tests. The third is a real observation and is deliberately deferred: capping `expectedResults` by `allowedRIDs.size()` would fix the same manufactured-scan in the allow-list branch, but it changes #5748's contract and would invalidate the test added here, so it belongs in its own change.

`mergeable` is `MERGEABLE`; `mergeStateStatus` will stay `UNSTABLE` for as long as `main`'s own baseline is red on these jobs.

PR #5888, opened 2026-08-07
fix(#5810): Cartesian-join disconnected MATCH after a relationship pattern

### LINKED ISSUE #5810: Disconnected MATCH after a relationship pattern binds the new variable to null instead of producing a Cartesian product
### ArcadeDB version

ArcadeDB `26.9.1-SNAPSHOT` (build `adb16f40bf26d006cad9513c3d3d9bf057622129/1785792317978/main`)

Also reproduced on `26.8.1-SNAPSHOT` (build `1134185d5494f5ab0583da4a508d846449f8e04c/1785129272315/main`).

### Environment

* Linux x86_64, Docker, OpenJDK 21
* `arcadedata/arcadedb:latest`
* HTTP command API
* Query language: Cypher

### Description

A disconnected `MATCH` clause following a relationship pattern does not bind its newly introduced variable.

For each row produced by the first `MATCH`, the disconnected second `MATCH` should independently bind `x` to every matching `B_0` node, producing a Cartesian product. Instead, ArcadeDB preserves a single row and binds `x` to `null`.

The behavior is order-dependent. When the disconnected node scan appears before the relationship pattern, ArcadeDB produces the correct row expansion.

### Steps to reproduce

ArcadeDB is schema-first. Declare the types through SQL before running the Cypher queries:

```sql
CREATE VERTEX TYPE A_0 IF NOT EXISTS;
CREATE VERTEX TYPE B_0 IF NOT EXISTS;
CREATE EDGE TYPE E IF NOT EXISTS;
```

Create the graph:

```cypher
MATCH (n) DETACH DELETE n;

CREATE (:A_0 {id: 1});
CREATE (:B_0 {id: 2});
CREATE (:B_0 {id: 3});

MATCH (a:A_0 {id: 1}), (b:B_0 {id: 2})
CREATE (a)-[:E]->(b);
```

The graph contains:

* one `A_0` node;
* two `B_0` nodes;
* one `E` relationship from `A_0(id=1)` to `B_0(id=2)`;
* one isolated `B_0(id=3)` node.

Run:

```cypher
MATCH (n:A_0)-[:E]->(m:B_0)
MATCH (x:B_0)
RETURN m.id, x.id
ORDER BY x.id;
```

### Expected behavior

The first `MATCH` produces one row. The disconnected second `MATCH` should bind `x` to both `B_0` nodes for that row:

```text
m.id | x.id
------+-----
2    | 2
2    | 3
```

The corresponding count should be `2`:

```cypher
MATCH (n:A_0)-[:E]->(m:B_0)
MATCH (x:B_0)
RETURN count(*) AS c;
```

```text
c
-
2
```

### Actual behavior

ArcadeDB returns one row with `x = null`:

```text
m.id | x.id
------+-----
2    | null
```

The count is also incorrect:

```text
c
-
1
```

A plain `MATCH` should either bind the variable to matching nodes or eliminate the row when no match exists. Producing a preserved row with a null binding is `OPTIONAL MATCH` behavior, not plain `MATCH` behavior.

### Controls

The individual query components work correctly.

Two disconnected node scans produce the expected Cartesian product:

```cypher
MATCH (a:A_0)
MATCH (b:B_0)
RETURN count(*) AS c;
```

```text
c = 2
```

The relationship pattern alone produces one row:

```cypher
MATCH (a:A_0)-[:E]->(b:B_0)
RETURN count(*) AS c;
```

```text
c = 1
```

Reversing the clause order also produces the correct result:

```cypher
MATCH (x:B_0)
MATCH (a:A_0)-[:E]->(b:B_0)
RETURN count(*) AS c;
```

```text
c = 2
```

This isolates the failure to a disconnected node `MATCH` that follows a relationship pattern.

### Cross-engine comparison

Setup: one `A_0` node, two `B_0` nodes, and one `E` relationship from `A_0(id=1)` to `B_0(id=2)`.

| Query                                                  | Neo4j 2026.06.0 | Memgraph 3.11.0 | FalkorDB 4.18.11 |   ArcadeDB   |
| ------------------------------------------------------ | :-------------: | :-------------: | :--------------: | :----------: |
| Relationship MATCH followed by disconnected node MATCH |  `2,2` / `2,3`  |  `2,2` / `2,3`  |   `2,2` / `2,3`  | **`2,null`** |
| Same query with `RETURN count(*)`                      |        2        |        2        |         2        |     **1**    |
| Two disconnected node scans                            |        2        |        2        |         2        |       2      |
| Relationship MATCH only                                |        1        |        1        |         1        |       1      |
| Disconnected node scan before relationship MATCH       |        2        |        2        |         2        |       2      |

### Isolated recheck

Tested on a fresh container with no prior state (2026-08-04, build `adb16f40`, image `ec37cbba0752`).

```text
MATCH (n:A_0)-[:E]->(m:B_0)
MATCH (x:B_0)
RETURN m.id, x.id
ORDER BY x.id
```

Results across three runs:

```text
[{"m.id":2,"x.id":null}]
[{"m.id":2,"x.id":null}]
[{"m.id":2,"x.id":null}]
```

Count query results:

```text
[{"c":1}]
[{"c":1}]
[{"c":1}]
```

Control results:

```text
Two disconnected node scans:             2 | 2 | 2
Relationship MATCH only:                  1 | 1 | 1
Disconnected scan before relationship:    2 | 2 | 2
```

Reproduction rate: 3/3 on both `26.9.1-SNAPSHOT` (build `adb16f40`) and `26.8.1-SNAPSHOT` (build `1134185d`).

### Why this looks like a bug

Each `MATCH` clause is applied to the rows produced by the preceding clause. When a later `MATCH` shares no variables with the existing row, it should independently enumerate its matches and produce a Cartesian product.

ArcadeDB follows this behavior for two disconnected node scans and when the disconnected scan appears first. It fails only when the disconnected scan follows a relationship pattern.

The resulting `null` binding is also incompatible with plain `MATCH` semantics because no `OPTIONAL MATCH` is present.

A possibly-related defect in `CartesianProductStep` was previously fixed in #4543 (duplicate rows when the outer level iterates before inner levels complete). This report shows the opposite symptom: rows are lost and the new variable is bound to `null` instead of being bound per row. Whether both share the same execution path is not established.

### Impact

Queries of this shape silently produce too few rows and introduce an unexpected null binding. This can corrupt returned results, counts, aggregations, filters, and downstream expressions that depend on the disconnected variable.

robfrank:
## What does this PR do?

Fixes the Cypher cost-based optimizer so a `MATCH` clause that is disconnected from (shares no
variable with) a preceding relationship pattern is correctly Cartesian-joined into the plan, instead
of being silently dropped so its variable resolves to `null`.

`CypherOptimizer.optimize()` built its single anchor + expansion chain from
`logicalPlan.getRelationships()` only. A node introduced by a later, disconnected `MATCH` (e.g. `x` in
`MATCH (n)-[:E]->(m) MATCH (x)`) shares no variable with any relationship, so it was never turned into
a physical operator at all — it fell out of the plan, and `RETURN`/`count(*)` read it as unbound
instead of expanding one row per match. This is the same shape `optimizeMultiMatchIndependent()`
already handles correctly when a query has *no* relationships at all; this PR extends that same
Cartesian-join treatment to a disconnected node that appears alongside a relationship pattern.

The fix computes, after the connected component's expansion chain (and its GAV fusion / deferred
vertex loading) is built, which pattern nodes are *not* reachable from the anchor via a relationship,
and Cartesian-joins each of them onto the plan using the same anchor-selection helpers
`optimizeMultiMatchIndependent` already uses. `WHERE` predicates that reference a disconnected node's
variable are deferred until after that join (they cannot be evaluated before the variable is bound);
every other predicate keeps going through the existing anchor-pushdown/filter path unchanged.

## Motivation

Reported in #5810: a plain `MATCH` clause that scans a disconnected node after a relationship pattern
produced `OPTIONAL MATCH`-like behavior (a preserved row with a `null` binding) instead of Cypher's
required Cartesian-product semantics, and `count(*)` undercounted the resulting rows. Cross-engine
comparison against Neo4j, Memgraph, and FalkorDB in the issue confirms the expected behavior.

## Related issues

Closes #5810

## Additional Notes

- `applyFilterPushdown` was refactored to take an explicit `List<WhereClause>` parameter instead of
  always reading `logicalPlan.getWhereFilters()`, so the connected-component and deferred filter
  buckets can each be applied at the right point in the plan. Both existing call sites (the
  single-anchor path and `optimizeMultiMatchIndependent`) keep their original behavior.
- The disconnected-node scan is built from `logicalPlan.getPatternNodes()` (not just the named-only
  `getNodes()`), so an *anonymous* disconnected node (e.g. `MATCH (n)-[:E]->(m) MATCH (:Label)`) is
  also caught — it still has to multiply the row count even though nothing reads its variable.
- Verified the new tests fail without the fix (4 failures/1 error reverting just
  `CypherOptimizer.java`) and pass with it restored, confirming they reach the fixed code path.
- Ran the full `com.arcadedb.query.opencypher.**` test package (8006 tests) after the fix: 0 failures.
  A batch of 21 `NoClassDefFoundError` errors surfaced only in that bulk run and only in test classes
  untouched by this change (`CypherRangeIndexTest`, `Issue5444CompositeIndexSeekTest`,
  `OpenCypherStackOverflowWorkloadTest`, etc.); re-running each of those classes in isolation passed
  cleanly, consistent with a local IDE background-compiler race against `target/classes` rather than a
  regression from this change.

## Test plan

- [x] `mvn -pl engine -am compile` — clean compile.
- [x] `mvn -pl engine -am test -Dtest=Issue5810DisconnectedMatchAfterRelationshipTest` — new regression
      test (6 cases): the reporter's exact query, `count(*)`, an assertion that the plan actually uses
      `CartesianProduct`, the already-correct reversed-clause-order control, an anonymous disconnected
      node, and a `WHERE` filter on the disconnected variable. All pass.
- [x] `mvn -pl engine test -Dtest='com.arcadedb.query.opencypher.optimizer.**,Issue5117PatternOrderTest,Issue5136MatchMultipleCreateOptimizerTest,Issue5810DisconnectedMatchAfterRelationshipTest'`
      — optimizer unit tests plus the two closest-related existing regression tests (disconnected
      comma-separated patterns in one `MATCH`, and multi-`MATCH` + `CREATE`). All pass.
- [x] Full `com.arcadedb.query.opencypher.**` package (8006 tests) — 0 failures (see note above on
      unrelated, isolation-confirmed environmental errors).

## Checklist

### DIFF
--- a/docs/5810-disconnected-match-after-relationship-null-bind.md
+++ b/docs/5810-disconnected-match-after-relationship-null-bind.md
@@ -125,3 +125,36 @@ developer decision.
   `NoClassDefFoundError` errors in test classes untouched by this change; re-running each of those
   classes in isolation passed cleanly, consistent with a local IDE background-compiler race against
   `target/classes` (IntelliJ IDEA was running concurrently) rather than a regression from this change.
+
+## PR and review history (PR #5888)
+
+- Duplicate discovered: while opening the PR, found PR #5876 (opened earlier the same day) already
+  fixes the identical root cause with a structurally similar approach. Left a cross-reference comment
+  on #5888; both PRs were left open since merge/close is a developer decision this workflow does not
+  make.
+- Cycle 1: head `05c7fe1b2` (initial fix - Cartesian-join disconnected nodes after `buildExpansionChain`,
+  before `AnchorSelector` was touched). `claude[bot]` posted a review as a GitHub **issue comment**
+  (not a PR review or inline comment) identifying one CONFIRMED correctness bug (anchor selection not
+  connectivity-aware - reproduced live), one performance finding (compound WHERE loses anchor pushdown
+  when mixed with a disconnected-node predicate), one minor performance nitpick (disconnected-only
+  filter not pushed into its own scan before the join - left as-is, out of scope), a test-coverage gap
+  list, and a style nit. Addressed the correctness bug and the compound-WHERE performance finding (both
+  verified against a live reproduction and new regression tests before/after); the minor nitpick was
+  intentionally deferred since it is "not incorrect, just more expensive" and outside this issue's
+  scope. Pushed as commit `0558aa581`.
+- Cycle 2: head `0558aa581`. The `claude-review` GitHub Actions check re-ran fresh for this exact
+  commit (a distinct job run, `92832148797`, tied to the new push) and completed successfully (27
+  turns, `is_error: false`) but posted zero comments - no new PR review, no inline comment, no issue
+  comment. Treated as a clean pass: the check is 1:1 with each pushed SHA, and passing with no findings
+  is the review system's way of saying nothing further is actionable. No code changes were needed for
+  this cycle. Final state: **clean-approval**.
+
+## Deferred items
+
+- The minor performance nitpick from cycle 1 (a filter referencing only the disconnected node's own
+  variable is applied as a `FilterOperator` on top of the full `CartesianProduct` rather than pushed
+  into that node's own scan before the join) was consciously left unaddressed - not incorrect, just
+  more expensive for a disconnected label with many rows. No separate follow-up issue was filed by this
+  run; left for the developer to decide whether it merits one.
+- PR #5876 duplicates this PR's fix for the same root cause with an independent, structurally similar
+  implementation. Deciding which PR to carry forward (and closing the other) is left to the developer.
--- a/docs/5810-disconnected-match-after-relationship-null-bind.md
+++ b/docs/5810-disconnected-match-after-relationship-null-bind.md
@@ -75,9 +75,53 @@ All 6 tests were verified to fail without the fix (reverted the `CypherOptimizer
 `git stash`, re-ran the test class: 4 failures + 1 error) and pass with it restored, confirming the
 tests reach and depend on the fixed code path.
 
+## Round 2: fix from the `claude[bot]` PR review (PR #5888)
+
+The bot review found a real, live correctness gap in the round-1 fix: `AnchorSelector.selectAnchor`
+is purely cost-based with no relationship-reachability awareness. If a disconnected node carried its
+own selective filter (e.g. `MATCH (n:A_0)-[:E]->(m:B_0) MATCH (x:B_0 {id: 3})`), it could win anchor
+selection outright over the unfiltered `n`/`m`. Reproduced live: the plan picked `x` as anchor and then
+tried to expand the `n-[:E]->m` relationship starting from `x` (not one of its endpoints), returning 0
+rows instead of the correct 1.
+
+Fix:
+- `AnchorSelector` gained a `selectAnchor(LogicalPlan, Set<String> excludedVariables)` overload that
+  skips any node whose variable is in the exclusion set; the original 1-arg method now delegates to it
+  with an empty set.
+- `CypherOptimizer.optimize()` now computes the disconnected-node set **before** calling
+  `selectAnchor`, from `logicalPlan.getRelationships()` alone (not from the chosen anchor), and passes
+  it as the exclusion set - so a disconnected node can never be selected as anchor in the first place.
+- The WHERE-filter split was upgraded from a coarse "does the whole clause touch a disconnected
+  variable" check to a per-conjunct split using the existing
+  `WhereClause.extractForVariables`/`residualForVariables` helpers, so a compound predicate like
+  `WHERE n.id = 1 AND x.id = 2` still lets the connected conjunct (`n.id = 1`) reach anchor pushdown
+  instead of deferring the whole clause because of the disconnected conjunct.
+- Minor style nit from the review (brace-wrapped single-statement `if`) also cleaned up.
+
+New tests added to the same test class:
+- `disconnectedNodeWithOwnIndexedFilterDoesNotHijackAnchorSelection` - the review's exact reproduction.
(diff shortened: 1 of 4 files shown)


robfrank:
Note for the developer: while working this issue I discovered PR #5876, opened earlier today, already fixes the same root cause in `CypherOptimizer.optimize()` (a disconnected single-node `MATCH` following a relationship pattern is never turned into a physical operator).

The two PRs take a different structural approach:
- #5876 excludes the isolated node's variable from anchor selection (`AnchorSelector.selectAnchor` gained an `excludedVariables` param) and joins it in via `CartesianProduct` *before* filter pushdown.
- This PR (#5888) leaves anchor selection untouched, computes the disconnected nodes after the expansion chain is built, and splits `WHERE` filters into a "connected" bucket (applied before GAV fusion, as before) and a "deferred" bucket (applied after the Cartesian join) so filters referencing only the connected component still benefit from GAV chain fusion.

Both should fix the reported bug; I have not compared them for correctness/performance beyond my own test suite. Recommend closing whichever one the maintainers don't want to carry forward - I'm leaving that decision (and the close/merge action) to the developer per this workflow's constraints.

robfrank:
Closing as superseded: #5810 was already fixed and merged via #5876 (same two files, equivalent Cartesian-join approach). Reconciling this branch's independent implementation on top of an already-shipped fix for the same bug isn't worth the risk of introducing duplicate/conflicting logic.

Issue #5285, opened 2026-07-15
Property access on direct non-variable expressions returns `null` instead of a type error


shulei5831sl:
### ArcadeDB version

`26.8.1-SNAPSHOT` (build `aea22a9bf30aef3ef392b535cc44b7480d4ccecc`)

### Environment

* Linux x86_64, Docker, OpenJDK 21.0.11
* `arcadedata/arcadedb:latest`
* HTTP command API
* Query language: Cypher

### Description

Direct property access on integer, string, boolean, and list expressions that do not pass through a variable returns `null` instead of raising a type error.

Equivalent values correctly raise `TypeError` after being bound to a variable. This creates an observable inconsistency between direct expression evaluation and variable-based evaluation of the same operand types.

### Steps to reproduce

Run each query separately. No graph data is required.

```cypher
RETURN 42.prop AS result;
RETURN (42).prop AS result;
RETURN (40 + 2).prop AS result;
RETURN ('hello').prop AS result;
RETURN true.prop AS result;
RETURN ([1, 2, 3]).prop AS result;
RETURN toInteger('42').prop AS result;
RETURN size([1, 2]).prop AS result;
```

ArcadeDB accepts every query and returns `null`.

For example:

```cypher
RETURN (40 + 2).prop AS result;
```

Response:

```text
HTTP status: 200
```

```json
{
  "user": "root",
  "result": [
    {
      "result": null
    }
  ]
}
```

### Expected behavior

Property access on the tested `INTEGER`, `STRING`, `BOOLEAN`, and `LIST` values should raise a type error, regardless of whether the operand is a direct expression or a variable containing the equivalent value.

Neo4j, Memgraph, and FalkorDB reject these non-property-bearing operands.

The following controls demonstrate the intended distinction between valid property access, a missing property, null propagation, and an invalid operand:

```cypher
RETURN {prop: 7}.prop AS result;
// 7

RETURN {}.prop AS result;
// null

WITH null AS value
RETURN value.prop AS result;
// null

WITH 42 AS value
RETURN value.prop AS result;
// TypeError: Cannot access property 'prop' on Long value
```

### Actual behavior

| Operand             | ArcadeDB result | Expected   |
| ------------------- | --------------- | ---------- |
| `42` / `(42)`       | `null`          | Type error |
| `(40 + 2)`          | `null`          | Type error |
| `('hello')`         | `null`          | Type error |
| `true`              | `null`          | Type error |
[...]

PR #5893, opened 2026-08-07
fix(#5800): dereference persisted node-valued property in chained Cypher access

### LINKED ISSUE #5800: Node-valued property is dereferenceable in the writing query but becomes a non-dereferenceable `DatabaseRID` after commit
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`
Build: `85eb71bb19a39fd1eb36a6fe3ee742862d47d6f6/1785731865672/main`

### Environment

* `arcadedata/arcadedb:latest`, freshly pulled on 2026-08-03
* HTTP command API
* Cypher language

### Description

Assigning a node value to a property silently succeeds:

```cypher
SET holder.ref = target
```

Within the same writing query, the property expression behaves like a node:

```cypher
RETURN holder.ref.id
```

returns the target node's `id`.

After the transaction commits, the property is stored as an ArcadeDB `DatabaseRID`. Reading the property returns an RID such as `#5:1`, while dereferencing the same property expression fails:

```text
TypeError: Cannot access property 'id' on DatabaseRID value
```

The failure is returned through the HTTP API as HTTP 500.

Under Cypher property semantics, a `NODE` is a structural value and is not a valid property value. Neo4j therefore rejects the assignment. ArcadeDB's native data model supports LINK properties backed by RIDs, so converting a node assignment to a LINK/RID may be intentional.

The issue is the transaction-boundary-dependent behavior: the property is dereferenceable as a node in the writing query, but the persisted value is a non-dereferenceable `DatabaseRID`.

### Steps to reproduce

Clear the database:

```cypher
MATCH (n)
DETACH DELETE n;
```

#### 1. Write the node-valued property

```cypher
CREATE (holder:T {role: 'holder'}),
       (target:T {role: 'target', id: 42})
SET holder.ref = target
RETURN holder.ref AS ref,
       holder.ref.id AS referencedId;
```

Observed result:

```text
referencedId = 42
```

The assignment succeeds and `holder.ref.id` is dereferenceable in the writing query.

#### 2. Read the persisted property in a new transaction

```cypher
MATCH (holder:T {role: 'holder'})
RETURN holder.ref AS ref;
```

Observed result:

```text
#5:1
```

The HTTP response serializes the stored property as an RID.

#### 3. Dereference the persisted property

```cypher
MATCH (holder:T {role: 'holder'})
RETURN holder.ref.id AS referencedId;
```

Observed result:

```text
HTTP 500

TypeError: Cannot access property 'id' on DatabaseRID value
```

### Expected behavior

ArcadeDB should provide one consistent behavior:

1. Reject assigning a `NODE` value to a property in the Cypher layer and roll back the write; or

2. If node values are intentionally converted to ArcadeDB LINK/RID properties, preserve stable dereference semantics across transaction boundaries.

The same expression, `holder.ref.id`, should not return `42` in the writing query and fail after the value has been persisted.

A value accepted by the write path should also not cause an HTTP 500 response when subsequently accessed.

### Actual behavior

| Operation                                | Result                                     |
| ---------------------------------------- | ------------------------------------------ |
| `SET holder.ref = target`                | succeeds                                   |
| `SET ... RETURN holder.ref.id`           | returns `42`                               |
| Next transaction: `RETURN holder.ref`    | returns RID such as `#5:1`                 |
| Next transaction: `RETURN holder.ref.id` | **HTTP 500**, `TypeError` on `DatabaseRID` |

### Cross-engine comparison

| Query                               |                         Neo4j 2026.06                        |   ArcadeDB   |
| ----------------------------------- | :----------------------------------------------------------: | :----------: |
| `SET holder.ref = target`           | Rejected because a node cannot be stored as a property value |   Succeeds   |
| `RETURN holder.ref.id` after commit |         Not applicable because the write is rejected         | **HTTP 500** |

### Isolated recheck

Tested in a fresh `arcadedata/arcadedb:latest` container with an empty database.

The complete three-step flow reproduced in 3/3 runs:

```text
Write and dereference:       referencedId = 42
Persisted property:          RID such as #5:1
Post-commit dereference:     HTTP 500 TypeError
```

### Why this looks like a bug

1. The same property expression has different semantics across the transaction boundary: it behaves as a node before persistence and as a non-dereferenceable RID afterwards.
2. ArcadeDB accepts the node-valued property assignment without an error or warning.
3. The persisted representation cannot be used through the same Cypher property-access expression that succeeded during the write.
4. Reading data previously accepted by the engine produces an HTTP 500 server error.

### Impact

Applications can successfully store entity references in properties and use them in the writing query, but the same references become unusable after commit.

This creates transaction-boundary-dependent query behavior and exposes persisted data that causes HTTP 500 when accessed through the expression that previously worked.

robfrank:
## What does this PR do?

Fixes a Cypher property-access bug: a node-valued property that Cypher persists as an ArcadeDB
LINK/RID (`SET holder.ref = target`) could be dereferenced (`holder.ref.id`) in the writing
query, but the identical expression raised
`TypeError: Cannot access property 'id' on DatabaseRID value` (HTTP 500 through the HTTP command
API) once read back in a later transaction.

Root cause: `CypherExpressionBuilder.ChainedPropertyAccessExpression` (used whenever a `.`
postfix chains onto something other than a plain variable, e.g. the outer `.id` in
`holder.ref.id`) handled `Document`, `Map`, `Result`, and temporal base values, but not `RID`.
The sibling single-level, variable-bound AST node,
`com.arcadedb.query.opencypher.ast.PropertyAccessExpression`, already dereferences a `RID` base
value via `rid.asVertex().get(propertyName)`; `ChainedPropertyAccessExpression` never got the
equivalent branch. This PR adds it, so the persisted RID is transparently dereferenced the same
way, keeping the expression's behavior consistent across the transaction boundary.

This preserves ArcadeDB's existing, intentional conversion of a NODE-valued property into a
LINK/RID on write (per the issue: "ArcadeDB's native data model supports LINK properties backed
by RIDs, so converting a node assignment to a LINK/RID may be intentional") and fixes only the
read-path inconsistency - no storage format or write-path change.

## Motivation

Closes #5800

## Related issues

Closes #5800

## Additional Notes

- New regression test:
  `engine/src/test/java/com/arcadedb/query/opencypher/CypherNodeValuedPropertyDereferenceIssue5800Test.java`
  - `chainedPropertyAccessOnPersistedLinkDereferencesAfterCommit` reproduces the exact issue
    scenario end-to-end (write + same-query dereference, then a fresh transaction dereferencing
    the persisted property).
  - `directPropertyAccessOnPersistedLinkStillDereferences` is a control pinning the already-working
    single-level (`WITH holder.ref AS r RETURN r.id`) path so a future regression there is also
    caught.
- Full tracking doc with root-cause analysis and verification details:
  `docs/5800-cypher-node-property-dereference.md`.
- Deliberately did not change write-time semantics (e.g. rejecting NODE-valued property
  assignment like Neo4j does) - that would be a separate, larger behavior change and a design
  decision outside the scope of this bug fix.

## Test plan

- [x] `mvn -pl engine test -Dtest=CypherNodeValuedPropertyDereferenceIssue5800Test` - failed
      before the fix with the exact reported `TypeError` (stack trace through
      `ChainedPropertyAccessExpression.evaluate`), passes after it (2/2).
- [x] Manually reproduced the issue's exact Cypher statements against the pre-fix engine jar via
      a standalone driver, confirming the reported `TypeError`; re-ran against the fixed jar and
      confirmed `holder.ref` still reads back as a `DatabaseRID` while `holder.ref.id` now
      resolves to the expected value instead of throwing.
- [x] Targeted regression run across property/`SET`-related OpenCypher unit tests (inline
      property filters/normalization/pattern binding, dynamic property mutation, `SET`, subquery
      parsing, edge mandatory property, parameterized property maps, the existing property-access
      type-error test #5285, etc.) - all passing.
- [x] Broad regression run across `com.arcadedb.query.opencypher.{ast,executor,parser,procedures,
      rewriter,optimizer,functions}` (excluding `benchmark`/`slow`-tagged and TCK-suite tests, per
      repository convention): 5678 tests, 0 failures, 0 errors.

## Checklist


robfrank:
## Review follow-up

Two non-blocking observations from the cycle-2 review (commit `5f4d382`) that don't change this
PR's code:

1. **`OrderByStep.java` has its own near-duplicate `convertFromStorage`** - pre-existing, out of
   scope for this bug fix. Left as-is; a good candidate for a small follow-up cleanup PR that
   points it at `TemporalUtil.convertFromStorage` too.
2. **`rid.asVertex()` raises an unfriendly `RecordNotFoundException`/`ClassCastException`
   instead of a Cypher-flavored `TypeError`** when a persisted RID no longer resolves to a live
   `Vertex` (deleted record, or a LINK pointing at a `Document`/`Edge`). This is pre-existing
   behavior on the single-level `PropertyAccessExpression` path, and this PR makes it reachable
   from the more common chained-access path too. Filed as a tracked follow-up: #5898.

(This comment replaces a `docs/review-deferred-*.md` file from the previous push - moving
review-audit-trail content here instead, since it documents PR discussion rather than the
system itself and would otherwise go stale/orphaned in the repo after merge.)

PR #5902, opened 2026-08-07
fix(#5895): bound RESP array nesting depth and element count in the Redis wrapper

### LINKED ISSUE #5895: Redis wrapper: unbounded RESP array recursion overflows the stack pre-auth (~47 KB); array size also unvalidated
## Summary

The Redis wrapper parses an incoming RESP message with unbounded recursion, before authentication. A ~47 KB message of nested arrays overflows the stack and kills the connection thread with `StackOverflowError`. No credentials are needed.

Version: 26.8.1. Module `redisw`.

## The recursion

```java
// redisw/src/main/java/com/arcadedb/redis/RedisNetworkExecutor.java:751
} else if (b == '*') {
  // ARRAY
  final List<Object> array = new ArrayList<>();
  final int arraySize = Integer.parseInt(parseValueUntilLF());
  for (int i = 0; i < arraySize; ++i)
    array.add(parseNext());        // recursive; nesting depth = stack depth
  return array;
}
```

A RESP array element can itself be an array, so `*1\r\n` repeated N times nests N deep, one `parseNext()` frame per level. `arraySize` is never bounded.

## Pre-authentication

`receiveCommand()` parses the whole message via `parseNext()` before the AUTH check runs:

```java
// RedisNetworkExecutor.java:159
if ("AUTH".equals(cmdString)) { … }
if ("HELLO".equals(cmdString)) { … }
if (authenticatedUser == null) {
  value.append("-NOAUTH Authentication required.");   // reached only AFTER the message is fully parsed
  …
}
```

The overflow happens inside parsing, so the NOAUTH gate never gets a chance to reject the message. Anyone who can open the Redis port reaches it.

## Measured against the real parser

Driven through the actual `RedisNetworkExecutor.parseNext()` over a loopback socket (not a re-implementation), bisected on Temurin 21.0.3 with the default stack:

| nesting depth | result |
|---|---|
| 11 860 | parsed |
| **11 861** | **StackOverflowError** |

Payload at the threshold: **47 451 bytes** (`*1\r\n` × 11 861 + `$1\r\nx\r\n`).

```java
final StringBuilder sb = new StringBuilder();
for (int i = 0; i < 11_861; i++) sb.append("*1\r\n");
sb.append("$1\r\nx\r\n");
// feed sb to a socket the RedisNetworkExecutor reads from -> StackOverflowError in parseNext()
```

## Second, related problem: unvalidated array size

Even without nesting, `arraySize` is trusted:

```java
final int arraySize = Integer.parseInt(parseValueUntilLF());
for (int i = 0; i < arraySize; ++i)
  array.add(parseNext());
```

A header `*2000000000\r\n` sets up a two-billion-iteration loop, each iteration blocking on `parseNext()` for bytes the client can dribble out slowly — a cheap way to tie up a connection thread indefinitely. A negative `arraySize` (e.g. `*-5`) is a different matter: the loop body is skipped, so that one is benign, but the value should still be validated.

The batch-string path (`$`) is safe by contrast: `parseChars(size)` reads byte-by-byte into a reused buffer without pre-allocating `size` bytes, so a large `$` length is a long read, not an allocation spike.

## Impact

- Reachable pre-auth by anyone who can connect to the Redis port.
- `StackOverflowError` is an `Error`; it unwinds the connection handler. Whether it also disturbs shared state depends on where it lands, which is not something a client should get to choose.
- 47 KB per hit, repeatable across connections.

## Suggested fix

- Cap nesting depth: thread a depth counter through `parseNext()` and reject past a small maximum (Redis itself limits multi-bulk nesting).
- Bound `arraySize` against a configured maximum and reject negatives explicitly, before entering the loop.
- Consider converting the array parse to an explicit work stack rather than call recursion, so depth is bounded by memory rather than by the JVM stack.

This is the same class as the Bolt frame-length issue (#5894): an unauthenticated wire parser trusting a client-supplied size. A single "validate sizes before you act on them" pass over the protocol wrappers would cover both.

lvca:
## Summary
- `RedisNetworkExecutor.parseNext()` decoded RESP arrays recursively with no bound on nesting depth. A ~47 KB payload of deeply nested arrays (`*1\r\n` repeated ~11,860 times) overflowed the connection thread's JVM stack with an uncaught `StackOverflowError`, reachable pre-authentication by anyone who can open the Redis port, since the whole message must be parsed before the `NOAUTH` check runs.
- The client-declared array length (`arraySize`) was also unbounded: a single `*2000000000\r\n` header started a parse loop the client could keep alive indefinitely by trickling bytes, tying up a connection thread.
- Both are now bounded by two new configurable settings mirroring Redis' own protocol limits: `arcadedb.redis.maxMultiBulkDepth` (default 32) and `arcadedb.redis.maxMultiBulkLength` (default 1,048,576, matching Redis' own hard cap on multibulk requests). A violation of either fails fast with a RESP `-ERR Protocol error: ...` reply, and the connection is closed rather than attempting to resynchronize on an untrusted stream position.

## Test plan
- [x] New regression test `RedisProtocolLimitsTest` (raw-socket, modeled on `RedisAuthenticationTest`) reproduces both issues against unpatched code (`StackOverflowError` for the nesting case, `SocketTimeoutException`/hang for the oversized-length case) and passes against the fix.
- [x] Full `redisw` module test suite passes (48 tests, 0 failures).
- [x] Full project compiles (`mvn compile`).

Closes #5895

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -181,4 +181,13 @@ Both are now bounded by two new settings mirroring Redis' own protocol limits: `
 a RESP `-ERR Protocol error: ...` reply and the connection is closed, since the stream position past a rejected
 message can no longer be trusted to resynchronize on.
 
+Code review on the fix caught the same bug class one level down: the RESP bulk-string (`$`) length was equally
+unbounded, and unlike the array case it sits on the hot path of essentially every command (the command name and
+every argument, including `GET`/`SET`'s own payloads, are bulk strings), so it was arguably the bigger exposure
+of the two. A `$2000000000\r\n` header could tie up a connection thread the same way, or grow the parse buffer
+without bound if the client actually sent the declared bytes. It is now capped by a third setting,
+`arcadedb.redis.maxBulkLength` (default 512MB, matching Redis' own `proto-max-bulk-len`). Malformed, non-numeric
+lengths (e.g. `$abc\r\n`) also used to throw an uncaught `NumberFormatException` that killed the connection
+thread outright; they now get the same clean error-reply-and-close treatment as an oversized length.
+
 [#5588](https://github.com/ArcadeData/arcadedb/issues/5588)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -165,4 +165,20 @@ garbage collector no longer traces one object graph per indexed vector.
   constructors: the two roles now take the same argument types and differ only in whether the reader may
   short-circuit to the vectors persisted inline in the graph file.
 
+## Redis wire protocol: an unauthenticated RESP array could overflow the connection thread's stack (#5895)
+
+`RedisNetworkExecutor.parseNext()` decoded RESP arrays recursively, once per nesting level, with no bound
+on either the nesting depth or the client-declared element count - and it runs before the `NOAUTH` check,
+since the whole message has to be parsed before a command exists to reject. On the default JVM stack, about
+47KB of `*1\r\n` repeated ~11,860 times was enough to overflow the connection thread with an uncaught
+`StackOverflowError`, reachable by anyone who can open the Redis port, no credentials required. Separately,
+an unbounded array-length header such as `*2000000000\r\n` started a parse loop the client could keep alive
+indefinitely just by trickling bytes, tying up a connection thread for as long as it liked.
+
+Both are now bounded by two new settings mirroring Redis' own protocol limits: `arcadedb.redis.maxMultiBulkDepth`
+(default 32 - no real command nests anywhere close to that) and `arcadedb.redis.maxMultiBulkLength` (default
+1,048,576, the same hard cap Redis itself enforces on multibulk requests). Either violation now fails fast with
+a RESP `-ERR Protocol error: ...` reply and the connection is closed, since the stream position past a rejected
+message can no longer be trusted to resynchronize on.
+
 [#5588](https://github.com/ArcadeData/arcadedb/issues/5588)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -165,6 +165,8 @@ garbage collector no longer traces one object graph per indexed vector.
   constructors: the two roles now take the same argument types and differ only in whether the reader may
   short-circuit to the vectors persisted inline in the graph file.
 
+[#5588](https://github.com/ArcadeData/arcadedb/issues/5588)
+
 ## Redis wire protocol: an unauthenticated RESP array could overflow the connection thread's stack (#5895)
 
 `RedisNetworkExecutor.parseNext()` decoded RESP arrays recursively, once per nesting level, with no bound
@@ -189,5 +191,3 @@ without bound if the client actually sent the declared bytes. It is now capped b
 `arcadedb.redis.maxBulkLength` (default 512MB, matching Redis' own `proto-max-bulk-len`). Malformed, non-numeric
 lengths (e.g. `$abc\r\n`) also used to throw an uncaught `NumberFormatException` that killed the connection
 thread outright; they now get the same clean error-reply-and-close treatment as an oversized length.
-
-[#5588](https://github.com/ArcadeData/arcadedb/issues/5588)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1660,6 +1660,13 @@ public enum GlobalConfiguration {
       *2000000000\\r\\n) starting a parse loop with billions of iterations. Default is 1048576.""",
       Integer.class, 1_048_576),
 
+  REDIS_MAX_BULK_LENGTH("arcadedb.redis.maxBulkLength", SCOPE.SERVER, """
+      Maximum length in bytes accepted for a single RESP bulk string ($) by the Redis wire-protocol listener, \
+      matching Redis' own proto-max-bulk-len. Without a bound, a client-declared length (e.g. $2000000000\\r\\n) \
+      can tie up a connection thread indefinitely by trickling bytes, or grow the parse buffer unbounded if the \
+      declared bytes are actually sent. Default is 536870912 (512MB).""",
+      Integer.class, 536_870_912),
+
   // MONGO
   MONGO_PORT("arcadedb.mongo.port", SCOPE.SERVER,
       "TCP/IP port number used for incoming connections for Mongo plugin. Default is 27017", Integer.class, 27017),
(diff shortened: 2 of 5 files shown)


lvca:
Addressed the review points:

- **Bulk-string ($) length was unbounded** (same bug class, higher-traffic path) — fixed. New `arcadedb.redis.maxBulkLength` setting (default 512MB, matching Redis' `proto-max-bulk-len`), with a regression test mirroring the array-length one.
- **Non-numeric length escaping error handling** — fixed. `Integer.parseInt` for all three length/integer sites (`*`, `$`, `:`) now goes through a helper that converts `NumberFormatException` into the same `RedisProtocolLimitException` fail-closed path, with a regression test.
- **No lower-bound validation on the new settings** — fixed. A misconfigured value below what a connection can actually parse with now falls back to the built-in default with a warning log, instead of silently breaking every command.
- **Javadoc `{@link}` to a private member** — fixed, switched to plain code font.
- **Test coverage gap on non-default configured values** — fixed, added a test that lowers `maxMultiBulkDepth` and confirms the new limit is actually enforced end-to-end.
- **Design note (reject nested arrays entirely, matching upstream Redis semantics)** — left as-is. It's a reasonable alternative and would remove the recursion vector rather than bound it, but the review itself flagged it non-blocking/not required, and the bounded-depth approach already closes the reported vector without changing accepted-command shape. Happy to revisit if there's a concrete reason to prefer it.

All 51 `redisw` tests pass, full project compiles.

lvca:
Addressed round 2:

- **Orphaned `[#5588]` doc link** — fixed, moved back to close the vector-index section it belongs to instead of trailing after the new Redis section.
- **Missing wiring tests for `maxMultiBulkLength`/`maxBulkLength`** — fixed, added `configuredMultiBulkLengthLimitIsHonored` and `configuredBulkLengthLimitIsHonored`, mirroring the existing depth test.
- **`parseChars()` char-by-char UTF-16/binary-safety gap** — real, but out of scope for this DoS fix and would touch how every command's arguments are read. Filed as #5907 rather than folding into this PR.
- **Depth floor's rationale not fully carrying over to the length floors** — agreed, it's documented/expected behavior for an intentionally low config rather than a bug; no code change.

All 53 `redisw` tests pass.

lvca:
Round 3 was clean (no blocking issues). Of the three minor/non-blocking notes:

- **Log-spam risk in `sanitizedLimit()` under connection churn** — fixed, the warning now fires once per setting per JVM instead of once per connection (the value itself is still re-read/re-validated per connection since it can change at runtime).
- **Redundant `arraySize <= 0` early-return** — left as-is; it's a no-op vs. prior behavior as noted, kept for the explanatory comment/symmetry with the other two length checks.
- **`$-1\r\n` parses to `""` not `null`** — explicitly called out as pre-existing and out of scope; no change.

All 53 `redisw` tests pass, full project compiles.

lvca:
Round 4 addressed:

- **RESP error reply bypassed the existing sanitization convention** — fixed. The new catch block now reuses `respErrorMessage()` (same as `executeCommand`'s generic handler) instead of embedding `e.getMessage()` verbatim. Added a regression test (`malformedLengthReplyDoesNotEmbedRawNewline`) reproducing the exact `$1\nA\r\n` scenario from the review — confirmed it fails against the unpatched code and passes with the fix.
- **Floor-of-1 nit for length settings** — addressed with a doc clarification on `sanitizedLimit()`'s javadoc distinguishing "parses at all" from "usable for real traffic."
- **`$-1\r\n` parser desync** (pre-existing, explicitly out of scope) — filed as #5911.
- **No idle/read timeout on the Redis socket** (pre-existing, explicitly out of scope, distinct DoS vector) — filed as #5912.

All 54 `redisw` tests pass, full project compiles.

lvca:
Round 5 addressed:

- **Off-by-one on the depth check vs. documented default** — fixed. `depth > maxMultiBulkDepth` allowed one extra nesting level (depth starts at 0), so "default 32" actually admitted 33. Changed to `>=`; this required raising the depth setting's `sanitizedLimit` floor from 1 to 2, since a flat command's single argument is parsed at depth 1 and `>=` needs at least 2 to still admit it.
- **Missing test for the floor-fallback path itself** — fixed, added `misconfiguredDepthLimitFallsBackToDefault`, which pipelines a malformed-shape command with a normal PING on the same connection to distinguish "fell back to the default (32)" from "used the broken configured value (1) verbatim."
- **`parseChars()` up to ~1GB per connection at the new 512MB cap** — already tracked by #5907 (filed after round 2); review explicitly notes it's a deliberate, defensible choice matching Redis' own default, not a blocker.

All 55 `redisw` tests pass, full project compiles.

lvca:
Round 6 addressed:

- **`parseValueUntilLF()` itself unbounded** — fixed, this is genuinely the same DoS class one level up: the length/integer token used to reach `maxBulkLength`/`maxMultiBulkLength`/`maxMultiBulkDepth` is itself read via an unbounded accumulator that only stops on a CRLF it may never receive. Added a fixed 64-character cap (real RESP length tokens and simple-string values are always short) and a regression test that hangs against unpatched code (confirmed via `SocketTimeoutException`) and is rejected cleanly with the fix.
- **TCP RST race on `close()` after the error reply** — real but explicitly non-blocking; not fixing in this PR (would mean adding `SO_LINGER`/drain semantics unrelated to the recursion/size bugs this PR targets).
- **Once-per-JVM misconfiguration warning could miss a later re-misconfiguration** — explicitly noted as a reasonable, documented tradeoff; no change.

All 56 `redisw` tests pass, full project compiles.

lvca:
Full CI run is complete. Summary for review:

**All 7 automated review rounds addressed** (see commit history and inline replies above) — the fix now bounds RESP array nesting depth, declared array/bulk-string length, and the raw length-token accumulator itself, with consistent fail-closed error handling (CRLF-sanitized single-line RESP replies) and 56 passing regression tests in `RedisProtocolLimitsTest`.

**Two out-of-scope findings filed as follow-ups**, not fixed here to keep this PR focused on the #5895 DoS class:
- #5907 — `parseChars()` reads bulk strings char-by-char (UTF-16, mangles non-ASCII/binary)
- #5911 — `$-1\r\n` (RESP2 null bulk string) desyncs the parser (pre-existing)
- #5912 — no idle/read timeout on the Redis socket (distinct slow-trickle DoS vector, pre-existing)

**CI status**: all checks green except three pre-existing failures, none touching `redisw/` or this PR's diff:
- `integration-tests`: `PostgresWJdbcIT.parsingErrorMgmt` (Postgres module)
- `unit-tests`: 4 `LSMVectorIndexRebuildTest`/`LSMVectorIndexRecoveryTest` timing errors (~120s each) + `CreatePropertyStatementExecutionTest.createHiddenProperty` (SQL DDL)
- `ha-integration-tests`: `Issue5410AbandonedTicketReleaseIT` + `Issue5569SlotMergeDeleteRaftIT` (HA/Raft)

Ready for your review, Luca.

Issue #5545, opened 2026-07-29
Cypher arithmetic errors (overflow, divide by zero) return HTTP 500 instead of 400


robfrank:
### Summary

Arithmetic domain errors in Cypher (integer overflow, integer division by zero) are reported to HTTP clients as **500 Internal Server Error**, when they are client errors and should be **400**. This affects the whole family uniformly, so it needs one fix rather than a per-function patch.

### Affected errors

All of these are raised as `CommandExecutionException`:

| Error | Source | Message |
|---|---|---|
| integer overflow on `+`, `-`, `*` | `ArithmeticExpression.integerArithmetic` (#5164) | `long overflow` |
| `Long.MIN_VALUE / -1` | `ArithmeticExpression.integerArithmetic` (#5164) | `long overflow` |
| integer division by zero | `ArithmeticExpression.checkIntegerDivisorNotZero` (#5163) | `/ by zero` |
| integer modulo by zero | `ArithmeticExpression.checkIntegerDivisorNotZero` (#5163) | `% by zero` |
| `abs()` of a MIN_VALUE | `AbsFunction`, `SQLFunctionAbsoluteValue` (#5494) | `<type> overflow` |

### Why they surface as 500

In `AbstractServerHttpHandler`, the classification chain has no arm for these:

- Raw path: `catch (CommandExecutionException | CommandParsingException)` falls through every special case to the trailing `else`, which answers **500** "Cannot execute command".
- Wrapped path: `DatabaseAbstractHandler` wraps the command in an auto-commit transaction, so the same error arrives as `TransactionException` -> `CommandExecutionException` cause and hits the symmetric arm, also **500**.

By contrast `IllegalArgumentException` is classified as **400** in both the raw and the wrapped arm.

Function exceptions are not re-wrapped on the way out (`FunctionCall.execute` passes them through untouched), so the type thrown at the source is exactly what the handler classifies.

### Why this is wrong

1. **Semantics.** The query is well-formed and the argument types are valid; the failure is a domain/range error determined entirely by client input. Neo4j classifies the same conditions as `Neo.ClientError.Statement.ArithmeticError` - a client error.
2. **Retry behavior.** 500 tells a client "the server failed, retry is reasonable". These queries fail deterministically, so every retry burns a round trip and fails identically.
3. **Log noise.** The 500 arms deliberately log the full stack trace at internal-error level (so genuine internal faults are diagnosable). A client typing a bad literal therefore writes a stack trace into the server log on every request.

[...]

Issue #5296, opened 2026-07-15
`left()` and `right()` with negative length return HTTP 500 instead of a client error


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`

```text
build: aea22a9bf30aef3ef392b535cc44b7480d4ccecc
image id: 880324d548aa
reproduction rate: 3/3
```

### Environment

* Linux x86_64, Docker, OpenJDK
* `arcadedata/arcadedb:latest`
* HTTP command API
* Cypher language

### Description

Calling `left()` or `right()` with a negative length argument returns an HTTP 500 internal server error instead of a descriptive client-side query error.

Positive, zero, and larger-than-string lengths work correctly. The failure is specific to negative lengths.

### Steps to reproduce

Run each query separately. No graph data is required.

```cypher
RETURN left('hello', -2) AS r;
```

```cypher
RETURN right('hello', -2) AS r;
```

### Expected behavior

A negative length is an invalid argument. The engine should reject the query with a descriptive client-side error, such as:

```text
Negative length is not allowed
```

Neo4j and Memgraph both return descriptive client errors for negative length arguments.

### Actual behavior

ArcadeDB returns HTTP 500:

```text
HTTP status: 500

{"error":"Cannot execute command",
 "exception":"com.arcadedb.exception.CommandExecutionException"}
```

No result row is returned.

Controls with non-negative lengths work correctly:

```cypher
RETURN left('hello', 2) AS r;      // "he"
RETURN right('hello', 2) AS r;     // "lo"
RETURN left('hello', 0) AS r;      // ""
RETURN right('hello', 0) AS r;     // ""
RETURN left('hello', 10) AS r;     // "hello"
RETURN right('hello', 10) AS r;    // "hello"
```

### Cross-engine comparison

| Query                | Neo4j 2026.05 | Memgraph 3.11 |   ArcadeDB   |
| -------------------- | :-----------: | :-----------: | :----------: |
| `left('hello', -2)`  |  Client error |  Client error | **HTTP 500** |
| `right('hello', -2)` |  Client error |  Client error | **HTTP 500** |
| `left('hello', 2)`   |     `"he"`    |     `"he"`    |    `"he"`    |
| `right('hello', 2)`  |     `"lo"`    |     `"lo"`    |    `"lo"`    |

### Isolated recheck

The behavior was reproduced in a fresh isolated Docker container:

```text
container: arcadedb_iso_20260715
[...]

PR #5909, opened 2026-08-07
fix(#5794): range() zero step, invalid date()/datetime(), non-numeric point() coordinates return HTTP 400

### LINKED ISSUE #5794: `range()` with zero step, invalid `date()`/`datetime()` strings, and non-numeric `point()` coordinates return HTTP 500 instead of a client error
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`
Build: `85eb71bb19a39fd1eb36a6fe3ee742862d47d6f6/1785731865672/main`

### Environment

* `arcadedata/arcadedb:latest`, freshly pulled on 2026-08-03
* HTTP command API
* Cypher language

### Description

A systematic function × invalid-input scan found four additional argument-validation paths that return HTTP 500 with `CommandExecutionException` instead of a client-facing HTTP 400 response:

* `range()` with a zero step
* `date()` with an invalid date string
* `datetime()` with an invalid datetime string
* `point()` with a non-numeric coordinate

ArcadeDB now returns HTTP 400 for neighboring deterministic input errors fixed in #5484 (`abs()`), #5296 (`left()`/`right()`), and #5545 (arithmetic errors). The four paths below still reach the unclassified HTTP 500 path.

### Steps to reproduce

Run each query separately:

```cypher
RETURN range(1, 5, 0);
```

```cypher
RETURN date('not-a-date');
```

```cypher
RETURN datetime('not-a-date');
```

```cypher
RETURN point({x: 'a', y: 1});
```

All four requests return HTTP 500 with a response containing:

```text
exception: com.arcadedb.exception.CommandExecutionException
error: Cannot execute command
```

### Valid-input controls

The corresponding valid inputs succeed, confirming that the failure is specific to invalid argument values:

```cypher
RETURN range(1, 5, 1);
-- [1, 2, 3, 4, 5]
```

```cypher
RETURN date('2026-08-03');
-- valid date value
```

```cypher
RETURN datetime('2026-08-03T12:00:00Z');
-- valid datetime value
```

```cypher
RETURN point({x: 1, y: 1}).x;
-- 1.0
```

The point property is used as the control because a point object itself currently serializes as `null` in the HTTP JSON response, while its properties are accessible.

### Expected behavior

Each query should return HTTP 400 with a descriptive client-facing argument error.

For comparison, Neo4j 2026.06 reports errors such as:

```text
range(): Step argument to 'range()' cannot be zero
date(): Text cannot be parsed to a Date
datetime(): Text cannot be parsed to a DateTime
point(): Cannot assign String(a) to field x
```

These failures are determined entirely by the supplied query arguments and cannot succeed when retried unchanged.

### Actual behavior

| Query                           | HTTP status | Exception                   |
| ------------------------------- | :---------: | --------------------------- |
| `RETURN range(1, 5, 0)`         |   **500**   | `CommandExecutionException` |
| `RETURN date('not-a-date')`     |   **500**   | `CommandExecutionException` |
| `RETURN datetime('not-a-date')` |   **500**   | `CommandExecutionException` |
| `RETURN point({x: 'a', y: 1})`  |   **500**   | `CommandExecutionException` |

Each query reproduced in 3/3 runs.

### Cross-engine comparison

| Query                           | Neo4j 2026.06 |   ArcadeDB   |
| ------------------------------- | :-----------: | :----------: |
| `RETURN range(1, 5, 0)`         |  Client error | **HTTP 500** |
| `RETURN date('not-a-date')`     |  Client error | **HTTP 500** |
| `RETURN datetime('not-a-date')` |  Client error | **HTTP 500** |
| `RETURN point({x: 'a', y: 1})`  |  Client error | **HTTP 500** |

### Isolated recheck

The results were reproduced in a fresh `arcadedata/arcadedb:latest` container with an empty database.

```text
range() zero step:          HTTP 500 (3/3)
invalid date():             HTTP 500 (3/3)
invalid datetime():         HTTP 500 (3/3)
non-numeric point() field:  HTTP 500 (3/3)
```

On the same build, the following neighboring invalid-input controls correctly return HTTP 400:

```cypher
RETURN left('hi', -1);
RETURN 1 / 0;
```

### Why this looks like a bug

1. ArcadeDB already classifies neighboring deterministic argument and arithmetic errors as HTTP 400 after fixes in #5484, #5296, and #5545. These four paths still return HTTP 500 for the same general class of client-caused failure.
2. Neo4j 2026.06 reports all four conditions as client errors.
3. HTTP 500 indicates a server-side failure and may cause clients to retry the same deterministic invalid query or trigger false server-health alerts.

### Impact

* Deterministic client input errors are exposed as server failures.
* Clients may perform unnecessary retries or generate false operational alerts.
* Error handling is inconsistent across ArcadeDB functions: some invalid inputs return HTTP 400, while these four return HTTP 500.

robfrank:
## What does this PR do?

Fixes four Cypher argument-validation paths that returned HTTP 500 (`CommandExecutionException`) instead of a client-facing HTTP 400 (`CommandSemanticException`):

- `range()` with a zero step
- `date()` with a string that parses as neither a date nor a timezone
- `datetime()` with a string that parses as neither a datetime nor a timezone
- `point()` with a non-numeric coordinate (both branches of the shared `coerceCoordinate()` guard: a non-numeric `String` and a value that is neither a `Number` nor a numeric-looking `String`)

Each failure is determined entirely by the supplied query arguments and cannot succeed when retried unchanged, so it belongs to the same class of client-caused error already fixed for the neighboring functions in #5484 (`abs()`), #5296 (`left()`/`right()`) and #5545 (arithmetic errors). Only the exception class changes; the existing descriptive messages are unchanged.

Closes #5794

## Motivation

A systematic function x invalid-input scan (reported in #5794) found these four paths still reaching the unclassified HTTP 500 path, causing clients to see a misleading internal-server-error status for a deterministic input mistake, which may trigger unnecessary retries or false operational alerts.

## Related issues

Closes #5794

Same class of fix as #5484, #5296, #5545.

## Additional Notes

Scope is intentionally limited to the four paths named in the issue's reproduction steps. While reproducing the bug I found the same `CommandExecutionException`-for-a-client-error defect in three sibling temporal constructors (`localdatetime()`, `localtime()`, `time()`) and in a few other `point()` structural-validation branches (missing `x`/`y` or `longitude`/`latitude`, non-map argument, non-numeric `srid`). Those are out of scope for this issue and are recommended as a follow-up.

## Test plan

- [ ] New regression test `engine/src/test/java/com/arcadedb/query/opencypher/CypherFunctionArgumentValidationIssue5794Test.java` reproduces all four named paths (five call sites) and asserts `CommandSemanticException`, plus a valid-input control test.
- [ ] Confirmed all five reproducer tests failed against `main` before the fix (asserting `CommandExecutionException` was thrown instead), proving the bug.
- [ ] `mvn -pl engine -am test -Dtest=CypherFunctionArgumentValidationIssue5794Test` passes (6/6) after the fix.
- [ ] Broader related-test pass with no regressions: `CypherNumericFunctionArgumentIssue5484Test`, `CypherFollowUpsIssue5602Test`, `CypherOptionalArgumentNullIssue5629Test`, `OpenCypherSpatialFunctionsComprehensiveTest`, `CypherFunctionFactoryExtendedTest`, `CypherFunctionArityRegistryTest`.
- [ ] `mvn -pl engine -am compile` succeeds.

## Checklist

### DIFF
--- a/docs/5794-cypher-function-arg-validation-500.md
+++ b/docs/5794-cypher-function-arg-validation-500.md
@@ -94,3 +94,28 @@ removed a now-unused `CommandExecutionException` import in
   (`localdatetime()`, `localtime()`, `time()`) and the remaining
   `CypherPointFunction` structural-validation branches, which have the
   identical defect class but were not part of this issue's named repro.
+  Done: filed as #5910.
+
+## Pull request
+
+https://github.com/ArcadeData/arcadedb/pull/5909
+
+## Review cycles
+
+- **Cycle 1** - head `d40023d05a4fcc639965a6ef4cf4a832e7441c91` (the fix
+  commit). `claude[bot]` posted its review as a PR issue comment (not a
+  formal GitHub review) ~4 minutes after the push: "Didn't find any
+  bugs... LGTM pending the tracking-issue note above." The one suggestion
+  (link a tracking issue for the documented out-of-scope follow-up) was
+  actioned without a code change: filed #5910 and linked it from a PR
+  comment. Working tree stayed clean, so the loop exited on a
+  clean-approval after cycle 1 with no further push needed.
+
+## Deferred items
+
+None. The single review suggestion was actionable-and-clear and was
+resolved by filing #5910, not by deferring to the developer.
+
+## Final state
+
+`clean-approval` after 1 review cycle.
--- a/docs/5794-cypher-function-arg-validation-500.md
+++ b/docs/5794-cypher-function-arg-validation-500.md
@@ -1,0 +1,96 @@
+# Issue #5794: Cypher argument-validation errors surfaced as HTTP 500
+
+## Root cause
+
+Four argument-validation paths in Cypher functions threw a bare
+`com.arcadedb.exception.CommandExecutionException`, which
+`AbstractServerHttpHandler` maps to HTTP 500 ("internal server error").
+`CommandSemanticException` (a subclass of `CommandParsingException`) is the
+class the HTTP layer maps to HTTP 400 for deterministic, client-caused
+argument errors — the pattern already established for `abs()` (#5484),
+`left()`/`right()` (#5296), and arithmetic errors (#5545).
+
+Each of the four failures below is determined entirely by the supplied query
+arguments and cannot succeed when retried unchanged, so it belongs to the
+same class of client error, not a server fault.
+
+## Affected components
+
+- `engine/src/main/java/com/arcadedb/function/coll/RangeFunction.java`
+  - `range()` with a zero step.
+- `engine/src/main/java/com/arcadedb/function/temporal/DateConstructorFunction.java`
+  - `date()` with a string that parses as neither a date nor a timezone.
+- `engine/src/main/java/com/arcadedb/function/temporal/DateTimeConstructorFunction.java`
+  - `datetime()` with a string that parses as neither a datetime nor a timezone.
+- `engine/src/main/java/com/arcadedb/function/geo/CypherPointFunction.java`
+  - `point()` with a non-numeric coordinate, in `coerceCoordinate()`. Both
+    branches were fixed (a non-numeric `String` and a value that is neither a
+    `Number` nor a numeric-looking `String`), since they are the same
+    validation guard for the same class of error.
+
+## Expected vs actual behavior
+
+| Query | Before | After |
+| --- | --- | --- |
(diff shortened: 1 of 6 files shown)


robfrank:
Filed #5910 to track the sibling argument-validation paths (`localdatetime()`, `localtime()`, `time()`, remaining `point()` structural branches) called out as out-of-scope follow-up in the PR description and in the review above.

PR #5920, opened 2026-08-07
fix(#5894): bound Bolt WebSocket frame length and Postgres bind-parameter length before allocation

### LINKED ISSUE #5894: Bolt WebSocket transport allocates an unbounded, unauthenticated client-supplied frame length: pre-auth OOM in ~14 bytes
## Summary

The Bolt WebSocket transport allocates a byte array sized by an unauthenticated client-supplied frame length, with no upper bound. A single small frame declaring a 2 GB payload triggers a 2 GB allocation before any Bolt handshake or authentication, so an unauthenticated peer can OOM or thrash the server with a few bytes.

Version: 26.8.1. Module `bolt`.

## The allocation

```java
// bolt/src/main/java/com/arcadedb/bolt/BoltWebSocketInputStream.java:79
long payloadLen = b1 & 0x7F;
if (payloadLen == 126)
  payloadLen = in.readUnsignedShort();
else if (payloadLen == 127)
  payloadLen = in.readLong();          // client-controlled, up to Long.MAX_VALUE
…
final byte[] payload = new byte[(int) payloadLen];   // no bound check
if (payloadLen > 0)
  in.readFully(payload);
```

`payloadLen` comes straight off the wire (the 64-bit extended length per RFC 6455) and is never checked against a maximum. There is no `maxFrameSize` guard anywhere in the `bolt` module.

## Pre-authentication

The WebSocket stream is installed during the upgrade, and the very first read through it is the Bolt magic — before the handshake:

```java
// bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java:337
input = new BoltChunkedInput(new BoltWebSocketInputStream(socket.getInputStream()));
output = new BoltChunkedOutput(new BoltWebSocketOutputStream(socket.getOutputStream()));
try {
  magic = input.readRaw(4);      // first frame decoded here, no auth yet
```

So the vulnerable path is reachable by anyone who can open a TCP connection to the Bolt port and complete a WebSocket upgrade. No credentials, no Bolt handshake.

## Attack

A WebSocket frame is: `0x82` (binary, FIN), `0xFF` (masked + length marker 127), an 8-byte length, a 4-byte mask, then the payload. To make the server allocate ~2 GB, the attacker sends the 2-byte header, the 8-byte length `0x000000007FFFFFFF`, and the 4-byte mask — **14 bytes** — then simply does not send the payload. The array is allocated before `readFully` blocks waiting for bytes that never come.

- One connection ties up 2 GB until the read times out or the socket is closed.
- Repeated across a handful of connections, it is a trivial OOM.
- The `(int)` cast makes behaviour erratic rather than safe: `0x7FFFFFFF` allocates 2 GB, while `0x100000000` truncates to 0 — so an attacker picks the value that maximises damage, and a fuzzer hits `NegativeArraySizeException` on values like `0x80000000`.

## Suggested fix

Bound the frame length before allocating, and reject anything over the limit as a protocol error that closes the connection:

```java
if (payloadLen < 0 || payloadLen > MAX_FRAME_SIZE)
  throw new IOException("Bolt WebSocket frame too large: " + payloadLen);
final byte[] payload = new byte[(int) payloadLen];
```

A few MB is a generous ceiling for a Bolt message; Neo4j's own server caps this. The limit should be configurable, consistent with the HTTP body cap the server already has (`HA_IDEMPOTENCY_CACHE_MAX_BODY_BYTES` and similar).

## Related, lower severity — post-authentication

The same read-a-length-then-allocate shape appears in the Postgres wrapper, but there it is behind authentication:

```java
// postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java:1270
final long paramSize = channel.readUnsignedInt();
final byte[] paramValue = new byte[(int) paramSize];   // bind message, post-auth
```

`bindCommand()` runs only after `readStartupMessage` + password (`run()` at line 159-167), so it is an authenticated client over-allocating — worth a bound too, but not the same exposure as the pre-auth Bolt path. Flagging it so both get the same guard.

lvca:
## Summary
- `BoltWebSocketInputStream` now rejects a frame whose decoded length is negative or exceeds `arcadedb.bolt.websocket.maxFrameSize` (new config, default 16MB) before allocating the payload array or reading the mask key. Closes the pre-auth path where ~14 bytes declaring a multi-gigabyte length could OOM the server before the Bolt handshake or authentication.
- `PostgresNetworkExecutor.bindCommand()` applies the same bound via `arcadedb.postgres.maxParamSize` (new config, default 16MB) to a Bind message's declared parameter length, in both the normal parse path and the error-recovery drain path, before any allocation.

Fixes #5894.

## Test plan
- [x] `BoltWebSocketInputStreamTest`: attack-shaped oversized frame, negative extended length, and the accept/reject boundary.
- [x] `PostgresBindParamBoundIT`: raw-socket regression test; verified it fails against the unpatched code (reproduces the actual `OutOfMemoryError`) and passes with the fix.
- [x] Full `PostgresProtocolIT` + `PostgresPortalTest` + Bolt suites pass (no regressions).

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -334,7 +334,8 @@ public class BoltNetworkExecutor extends Thread {
         completeWebSocketUpgrade(headers);
 
         // Reinitialize I/O with WebSocket framing and read Bolt magic from WebSocket stream
-        input = new BoltChunkedInput(new BoltWebSocketInputStream(socket.getInputStream()));
+        input = new BoltChunkedInput(
+            new BoltWebSocketInputStream(socket.getInputStream(), GlobalConfiguration.BOLT_WEBSOCKET_MAX_FRAME_SIZE.getValueAsInteger()));
         output = new BoltChunkedOutput(new BoltWebSocketOutputStream(socket.getOutputStream()));
         try {
           magic = input.readRaw(4);
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltWebSocketInputStream.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltWebSocketInputStream.java
@@ -30,13 +30,15 @@ import java.io.InputStream;
  */
 class BoltWebSocketInputStream extends InputStream {
   private final DataInputStream in;
+  private final long    maxFrameSize;
   private byte[]  buffer;
   private int     bufferPos;
   private int     bufferLen;
   private boolean closed;
 
-  BoltWebSocketInputStream(final InputStream in) {
+  BoltWebSocketInputStream(final InputStream in, final long maxFrameSize) {
     this.in = new DataInputStream(in);
+    this.maxFrameSize = maxFrameSize;
   }
 
   @Override
@@ -82,6 +84,9 @@ class BoltWebSocketInputStream extends InputStream {
         payloadLen = in.readUnsignedShort();
       else if (payloadLen == 127)
         payloadLen = in.readLong();
+
+      if (payloadLen < 0 || payloadLen > maxFrameSize)
+        throw new IOException("BOLT WebSocket frame too large: " + payloadLen + " bytes (max " + maxFrameSize + ")");
 
       byte[] maskKey = null;
       if (masked) {
--- a/bolt/src/test/java/com/arcadedb/bolt/BoltWebSocketInputStreamTest.java
+++ b/bolt/src/test/java/com/arcadedb/bolt/BoltWebSocketInputStreamTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.bolt;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Regression tests for issue #5894: a BOLT WebSocket frame with an unbounded, unauthenticated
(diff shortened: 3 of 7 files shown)


lvca:

- **NULL bind parameter (real bug, fixed in 42fce7d)**: confirmed by reproduction - a JDBC client calling `setNull`/`setObject(null)` sent a real Bind message that the server's new max-size check rejected as "too large" (unsigned `-1` sentinel), and because a well-behaved driver never sends the filler bytes an attacker-drain-recovery expects, the connection actually **hung** rather than returning a clean error. Added a regression test (`PostgresProtocolIT#preparedStatementBindsActualNullValue`) that hangs against the pre-fix code and passes now that both the main bind path and the error-recovery drain path special-case the `0xFFFFFFFF` NULL sentinel before the size check.
- **Inconsistent "too large" messages**: aligned, both now include the configured max.
- **Three near-duplicate read helpers in `PostgresBindParamBoundIT`**: left as-is - the review itself noted this is test code and readability is fine, so not worth the abstraction.

All local suites green: `PostgresProtocolIT`, `PostgresPortalTest`, `PostgresBindParamBoundIT`, `BoltWebSocketInputStreamTest`, `BoltChunkedIOTest`.

lvca:
Addressed the drain-misalignment finding in d80e1a0, but with a different resolution than the suggested \"skip the declared bytes in bounded chunks\": that approach has the same weakness the original OOM report targeted, just shifted - the declared length is still fully client/attacker-controlled, so a genuine attack (huge declared length, no follow-through) would block the connection thread waiting for bytes that never arrive. Bounded-chunk allocation avoids the *memory* exhaustion but not a *thread-pinning* one.

Instead, both the main bind-parsing path and the error-recovery drain path now send a clean `ErrorResponse` and close the connection outright on a cap violation, rather than attempting to resynchronize a channel via a client-controlled byte count. This mirrors how real PostgreSQL treats malformed/over-limit messages (protocol violation → connection closed) and how the Bolt WebSocket fix already behaves (no resync attempt there either) - so the two fixes in this PR are now consistent with each other.

Trade-off, stated explicitly: a legitimate client whose single bind parameter exceeds the configured cap gets its whole connection closed rather than just that one statement erroring out. Given `arcadedb.postgres.maxParamSize` is user-configurable (default 16MB) and this is a hard resource limit, I think that's the right trade-off versus reintroducing any form of unbounded/blocking read.

Updated `PostgresBindParamBoundIT` to assert the new behavior end-to-end: clean `ErrorResponse`, then the connection reaches EOF (not hanging or corrupted), then a fresh connection is still served promptly (listener/thread pool not wedged). All local suites green: `PostgresProtocolIT`, `PostgresPortalTest`, `PostgresBindParamBoundIT`, `PostgresErrorClassificationTest`, `PostgresProtocolExceptionTest`, Bolt suites.

lvca:
Thanks for confirming the fix. On the two remaining minor points - judging both as not worth implementing in this PR:

1. **Pre-existing deserialize-failure paramsConsumed misalignment**: confirmed pre-existing (unrelated to the size-bound fix - it's the "unsupported type code" / structurally-malformed-value path) and, as you noted, low severity - old and new behavior both end up erroring/closing the connection either way, just via a different message. Fixing it would mean touching the deserialize-success path's bookkeeping for a case this PR doesn't otherwise touch, which is scope creep for a security fix. Filing it separately if it turns out to matter is the right call, not folding it in here.
2. **Integer.class caps the config below the wire's max encodable length**: agreed this is only theoretical - nobody would legitimately raise `arcadedb.postgres.maxParamSize`/`arcadedb.bolt.websocket.maxFrameSize` anywhere near 2GB, since that defeats the point of the bound. Widening the type to accommodate a ceiling nobody would sanely configure isn't worth the change.

All CI is green except the full unit/integration/e2e suites are still running - will let those finish before considering this done.

Issue #5791, opened 2026-08-03
BUG: FULL_TEXT index build over existing data still indexes nothing matchable on 26.7.2 (follow-up to #4732); prefix queries on such an index throw NPE


NooriUta:
Reporting this as a follow-up rather than a new defect: the behaviour is exactly what #4732 describes, and it still happens on 26.7.2 — a release after that issue's 26.7.1 milestone. Adding it separately because a comment on a closed issue is easy to miss, and because the new part here is what I could *not* reproduce.

## Environment

- ArcadeDB Server **26.7.2**, build `7059e543a612de85151195dae556d849561f17c5/1783565728762/main`
- JVM: OpenJDK 64-Bit Server VM, **Temurin 21.0.11+10** (`openjdk version "21.0.11" 2026-04-21 LTS`)
- Official `arcadedata/arcadedb:26.7.2` container, single server, no encryption at rest
- Affected type: vertex type, one bucket, **6769 records**, one indexed `STRING` property, ~3.7 MB of text total, average length ~800 chars. Rows are written in an SCD2 pattern — repeatedly inserted and updated over several months.

## What happens

`CREATE INDEX ... FULL_TEXT` over the populated type reports every row as indexed:

```
{"operation":"create index","name":"ftTaskHist","type":"FULL_TEXT",
 "created":true,"totalIndexed":6754}
```

but the index does not contain them. Comparing a scan against the index on the same field, same type, same term:

| term | `note_md LIKE '%term%'` | `SEARCH_INDEX('ftTaskHist', 'term')` |
|---|---|---|
| `LORE` | **718** | **6** |
| `релиз` | **183** | **1** |

The handful that do match are rows written *after* the index was created, which is exactly the description in #4732: pre-existing rows are counted but never become matchable.

A second full-text index in the same database, on a smaller type (248 rows), was re-created with the identical statement at the same time and came out **correct**.

## `REBUILD INDEX` also still loses the index name

The obvious repair still fails, and in a way worth calling out separately:

```sql
REBUILD INDEX `ftTaskHist`;
-- {"operation":"rebuild index","indexes":["ftTaskHist"],
--  "totalIndexed":6754,"recordsMisplaced":0}

SELECT count(*) FROM TaskHist WHERE SEARCH_INDEX('ftTaskHist','…') = true;
-- SchemaException: Index with name 'ftTaskHist' was not found
```

[...]

Issue #4732, opened 2026-06-24
BUG: FULL_TEXT index (re)build over existing data is broken - `CREATE INDEX` on a populated type indexes nothing matchable, and `REBUILD INDEX` crashes and drops the index


TheRealMcoy:
## Description

Two related failures on the FULL_TEXT bulk-build path, both reproduced on a single-property index:

1. **`CREATE INDEX ... FULL_TEXT` over a type that already has rows reports the rows as indexed (`totalIndexed:N`) but they never match.** Only documents inserted *after* the index is created are returned by `SEARCH_INDEX`. So the common "load data, then add a full-text index" flow yields a silently empty index for all pre-existing rows.

2. **`REBUILD INDEX <fulltext-index>` crashes and leaves the index dropped.** The obvious recovery for (1) - rebuild the index from the documents - fails with `A FULL_TEXT index requires FullTextIndexMetadata but got com.arcadedb.schema.IndexMetadata`, and afterward the index no longer exists  (`Index with name '...' was not found`). The `WITH statsOnly = true` variant (BM25 corpus-counter repair) works correctly; only the full reindex crashes.

Together these mean a FULL_TEXT index cannot be correctly built or repaired over data that already exists: the create path indexes nothing matchable, and the reindex path destroys the index.

## Environment

- ArcadeDB Server `26.7.1-SNAPSHOT` (build `661e52fce3b84f452f02cc131b9c5608674deb67/1782276736394/main`)
- Default database options
- Reproduced with the default similarity (BM25). The build path is shared with CLASSIC; see the related multi-property report for a CLASSIC repro of the build path crashing.

## Root cause

Suspected, two areas:

1. **Build-over-existing (CREATE):** the type-scan build that populates a new index from existing records appears not to route those records through the same analyze-and-store path used by live `put()` (or stores them in a form `SEARCH_INDEX` does not look up). `totalIndexed` is reported from the scan count, so it reads as success while no matchable postings were written. The live `put()` path (`engine/.../index/fulltext/LSMTreeFullTextIndex.java#put` / `putWithStats`) is known-good because create-then-insert works; the divergence is in the initial build/scan.

[...]

PR #5925, opened 2026-08-07
fix(#5791): REBUILD INDEX preserves an explicitly-named FULL_TEXT/GEOSPATIAL index

lvca:
## Summary

Investigation of #5791 (a follow-up report to #4732) found three claimed symptoms; only one still reproduces on `main`.

- **Confirmed and fixed**: `REBUILD INDEX <name>` silently renames an explicitly-named `FULL_TEXT`/`GEOSPATIAL` index back to its auto-derived `typeName[properties]` form. `RebuildIndexStatement.buildIndex()` drops and recreates the `TypeIndex` through a fresh `TypeIndexBuilder` but never calls `withIndexName(...)`, so `TypeIndexBuilder.create()` never populates `metadata.typeIndexName` and `LocalDocumentType.addIndexInternal` falls back to the default name. A plain `LSM_TREE` rebuild happened to keep its name because its metadata object is reused as-is, but `FULL_TEXT`/`GEOSPATIAL` reconstruct their metadata from the index's persisted JSON to recover type-specific settings (analyzers, GeoHash precision), and `typeIndexName` isn't part of that JSON - so it silently reverts. Every name-based lookup breaks after the rebuild, including `SEARCH_INDEX`, which has no other way to reference an index. Fixed by passing `withIndexName(idx.getName())` into the rebuild.

- **Does not reproduce on `main`**: the core reported defect ("`CREATE INDEX ... FULL_TEXT` over pre-existing rows indexes nothing matchable") and the NPE on prefix queries. A stress test (`FullTextBuildOverChurnedDataTest`) replays the reporter's SCD2-style churn pattern - ~6,800 staggered inserts, several rounds of growing updates that force some records into placeholder relocation, and interleaved deletes, crossing the 5,000-record build-batch-commit boundary - then builds the `FULL_TEXT` index and checks every marked row is matchable, exercises the same `REBUILD INDEX` path, and runs a prefix query against the churned index. All pass cleanly. This is consistent with #4732's build-through-`put()` rewrite (already in 26.7.1+) and #5635's `IndexCursor` contract change (`next()` never returns `null`, so the class of NPE reported can no longer occur) - neither of which the reporter had confirmed against `main` when filing.

## Changes

- `engine/src/main/java/com/arcadedb/query/sql/parser/RebuildIndexStatement.java`: preserve the index's name across `REBUILD INDEX`.
- `engine/src/test/java/com/arcadedb/index/RebuildIndexPreservesNameTest.java`: regression test covering plain `LSM_TREE`, `FULL_TEXT`, and `GEOSPATIAL` named-index rebuilds.
- `engine/src/test/java/com/arcadedb/index/fulltext/FullTextBuildOverChurnedDataTest.java`: reproduces the reporter's exact churn/build/rebuild/prefix-query scenario as a `@Tag("slow")` regression test.

## Test plan

- [x] `mvn -pl engine install -DskipTests` compiles cleanly
- [x] New tests fail without the fix (verified by reverting it) and pass with it
- [x] Full `com.arcadedb.index.fulltext.**`, `com.arcadedb.index.RebuildIndex*`, `com.arcadedb.query.sql.parser.RebuildIndexStatement*`, and `com.arcadedb.function.sql.fulltext.**` suites: 245 tests, 0 failures
- [x] Full `com.arcadedb.schema.**` suite: 612 tests, 0 failures

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/RebuildIndexStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/RebuildIndexStatement.java
@@ -147,6 +147,12 @@ public class RebuildIndexStatement extends DDLStatement {
 
       final List<Index> targetIndexes = new ArrayList<>();
       if (all) {
+        // `*` deliberately excludes TypeIndex: it walks bucket sub-indexes one at a time through buildBucketIndex()
+        // below, which rebuilds each sub-index in place. This still loses the logical index's own name (issue #5791):
+        // dropping the LAST remaining sub-index under a TypeIndex also drops the wrapper (LocalSchema.dropIndex), so
+        // a single-bucket type - the common default - loses its TypeIndex on every rebuild and mints a fresh one.
+        // BucketIndexBuilder.create() and the withIndexName(...) call further down carry the owning TypeIndex's name
+        // through that recreation, the same fix applied below to the direct typeIndexRebuild path.
         for (final Index idx : database.getSchema().getIndexes())
           if (idx.isAutomatic() && !(idx instanceof TypeIndex))
             targetIndexes.add(idx);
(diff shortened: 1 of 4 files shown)


lvca:

- **Multi-bucket coverage gap**: added `rebuildAllKeepsNameOnMultiBucketFullTextIndex`, confirming the reviewer's own analysis - on a multi-bucket type the `TypeIndex` wrapper always survives (another sub-index keeps it alive), so the `withIndexName(...)` this PR adds is a no-op there. Passes as-is.
- **`DatabaseChecker` latent defect**: confirmed by inspection (same drop-then-`buildBucketIndex()`-without-name-or-metadata shape, plus it drops FULL_TEXT/GEOSPATIAL settings entirely since it passes no metadata at all). Filed as #5934 rather than fixed here - it's a different command (`CHECK DATABASE FIX`) with a wider blast radius than this PR's scope.
- Style note: no action, per the reviewer's own read that the comment density matches this file's existing style.

Issue #5937, opened 2026-08-07
arcadedb-gremlin plain (unshaded) coordinate still cannot run textual Gremlin queries: ANTLR ATN v3/v4 mismatch


lvca:
## Summary

Found while investigating #5879. Even after fixing #5879 (declaring `arcadedb-network` at compile
scope so the class loads), the plain `arcadedb-gremlin` Maven coordinate (no classifier) still
cannot execute a **textual** Gremlin query end-to-end. Any call through ArcadeDB's `QueryEngine`
SPI (`db.getQueryEngine("gremlin").analyze(...)`, `command("gremlin", ...)`, the secure `"java"`
engine that `ArcadeGremlin.executeStatement()` uses by default) crashes with:

```
java.lang.ExceptionInInitializerError
	at org.apache.tinkerpop.gremlin.language.grammar.GremlinQueryParser.parse(...)
	...
Caused by: java.lang.UnsupportedOperationException: java.io.InvalidClassException: org.antlr.v4.runtime.atn.ATN; Could not deserialize ATN with version 3 (expected 4).
	at org.antlr.v4.runtime.atn.ATNDeserializer.deserialize(...)
	at org.apache.tinkerpop.gremlin.language.grammar.GremlinLexer.<clinit>(...)
```

Reproduced against both current `main` and the actual released `26.8.1` Maven Central artifacts.

## Root cause

TinkerPop's `gremlin-language` module ships a Gremlin grammar parser precompiled with ANTLR
**4.9.1** (the "v3 ATN" serialization format). `arcadedb-gremlin` also pulls in
`arcadedb-integration` -> `arcadedb-engine`, which requires ANTLR **4.13.2** for its own SQL/Cypher
parsers (v4 ATN format). Those two ATN formats are mutually incompatible in one classloader
(documented in `gremlin/pom.xml`'s `antlr4.version` property comment, and in issues #5208/#5216 and
#5217/#5255). The project already solves this for its own tests by consuming the
`arcadedb-gremlin:shaded` classifier, which relocates `org.antlr` into a private package so
TinkerPop's precompiled classes bind to a private, matching-version copy instead of the ambient
engine one (see `gremlin/pom.xml`'s shade-plugin config, and `gremlin-it/pom.xml`'s comments).

The problem: a normal Maven consumer writing

```xml
<dependency>
  <groupId>com.arcadedb</groupId>
  <artifactId>arcadedb-gremlin</artifactId>
  <version>...</version>
</dependency>
[...]

Issue #5217, opened 2026-07-11
Maven-coordinate consumer isolation for arcadedb-gremlin + arcadedb-engine antlr (follow-up to #5216)


robfrank:
Follow-up from #5216 (sub-issue of #4647).

PR #5216 fixes the ANTLR version clash for the **shipped distribution** (engine plain 4.13.2 + gremlin shaded jar with relocated 4.9.1, copied directly into `lib/`) and for **engine-only** embedders.

It intentionally defers isolation for **Maven-coordinate consumers** who depend on BOTH `arcadedb-gremlin` and `arcadedb-engine` directly (not via the shipped distribution jars):

- The gremlin shaded artifact's pom is not dependency-reduced (`createDependencyReducedPom=false` in the parent), so it still lists `antlr4-runtime:4.9.1` transitively.
- The unshaded `arcadedb-gremlin` jar references `org.antlr` and transitively drags engine's antlr 4.13.2.

So a third-party app depending on both artifacts by coordinate can still hit Maven version mediation and the ATN mismatch.

### Proposed fix
Mark gremlin's `antlr4-runtime` dependency `<optional>true</optional>` (or `provided`) in `gremlin/pom.xml` so it is not exported transitively, and document that embedders using both modules should consume the `shaded` classifier.

### Also
Surface this prominently in the release notes for the version that ships #5216, so embedders using both modules know to consume the `shaded` classifier or wait for this follow-up.

Raised by the automated review on #5216.

PR #5938, opened 2026-08-07
fix(#5879): arcadedb-gremlin is unusable standalone - declare arcadedb-network as a compile dependency

### LINKED ISSUE #5879: arcadedb-gremlin 26.8.1 is unusable standalone: arcadedb-network is not declared, only reachable through a 'provided' dependency
## Summary

A project that depends only on `com.arcadedb:arcadedb-gremlin:26.8.1` fails at runtime on the first Gremlin query with:

```
java.lang.NoClassDefFoundError: com/arcadedb/remote/RemoteDatabase
```

`RemoteDatabase` lives in `arcadedb-network`, which the published POM never declares. The only path to it is `arcadedb-server`, declared with `scope=provided` — which is not transitive, so consumers never receive it.

Version: 26.8.1.

## Reproducer

`pom.xml`, one dependency:

```xml
<dependencies>
  <dependency>
    <groupId>com.arcadedb</groupId>
    <artifactId>arcadedb-gremlin</artifactId>
    <version>26.8.1</version>
  </dependency>
</dependencies>
```

```java
final Path dir = Files.createTempDirectory("g");
try (final DatabaseFactory f = new DatabaseFactory(dir.resolve("db").toString());
     final Database db = f.create()) {
  db.getQueryEngine("gremlin").analyze("g.V()");   // NoClassDefFoundError
}
```

Resolving that POM:

| artifact | on the resolved classpath |
|---|---|
| `arcadedb-engine` | yes (via `arcadedb-integration`) |
| **`arcadedb-network`** | **no** |

Adding `arcadedb-network` by hand makes the same program run.

## Why it resolves this way

From the published `arcadedb-gremlin-26.8.1.pom`:

```
com.arcadedb:arcadedb-server        scope=provided
com.arcadedb:arcadedb-server        scope=test
com.arcadedb:arcadedb-integration   scope=compile
```

`arcadedb-network` is not declared at all. The module compiles because `arcadedb-server` (provided) brings it in at compile time; `provided` is then dropped for consumers, and nothing else supplies it.

## Why it is hit immediately

The class is referenced on the ordinary query path, not in some remote-only corner:

```java
// gremlin/src/main/java/com/arcadedb/gremlin/ArcadeGremlin.java:71
if (graph.database instanceof RemoteDatabase)
```

An `instanceof` against a missing class fails when the enclosing method is linked, so even a purely embedded user who never touches remote functionality trips over it. `ArcadeGraph` and `ArcadeGraphFactory` import it too.

## Suggested fix

Declare the dependency the module actually uses, at compile scope:

```xml
<dependency>
  <groupId>com.arcadedb</groupId>
  <artifactId>arcadedb-network</artifactId>
  <version>${project.version}</version>
</dependency>
```

Relying on a `provided` dependency for a compile-time reference works in the reactor and silently breaks for every downstream consumer, since the reactor build never resolves the module the way Maven Central serves it.

Worth checking the sibling wrappers (`graphql`, `mongodbw`, `redisw`, `postgresw`, `bolt`) for the same pattern, and adding a smoke test that resolves each published artifact on its own and runs one query — that is the only shape of test that catches this class of problem, since it cannot reproduce inside the multi-module build.

## Secondary

With the classpath completed, the same run prints:

```
ANTLR Tool version 4.9.1 used for code generation does not match the current runtime version 4.13.2
ANTLR Runtime version 4.9.1 used for parser compilation does not match the current runtime version 4.13.2
```

TinkerPop's generated parser was built against ANTLR 4.9.1 while 4.13.2 is what resolves. It is a warning rather than a failure here, but the pairing is worth pinning deliberately rather than inheriting.

lvca:
## Summary

Fixes #5879: a consumer depending only on `com.arcadedb:arcadedb-gremlin` fails on the very
first Gremlin query with `NoClassDefFoundError: com/arcadedb/remote/RemoteDatabase`.

`RemoteDatabase` lives in `arcadedb-network`, which `arcadedb-gremlin`'s POM never declared.
`ArcadeGraph`, `ArcadeGraphFactory` and `ArcadeGremlin` reference it via `instanceof` checks on
the ordinary **embedded** query path, not only for remote usage, so linking fails for every
consumer, embedded-only or not. The only path to the class was through `arcadedb-server`,
declared `provided` (not transitive), which is why the reactor build never caught it.

**Fix**: declare `arcadedb-network` as a real `compile` dependency in `gremlin/pom.xml` (it only
depends on `arcadedb-engine` itself, so this is lightweight). Also excludes it from the shaded
uber-jar - like `arcadedb-engine` already is - since it's now resolved transitively for both the
plain and `:shaded` jar and doesn't need to be bundled twice.

## A second, related but distinct bug found while verifying the fix

I reproduced the reported crash against both current `main` and the actual released `26.8.1`
artifacts, confirmed the fix resolves it, then re-ran the reporter's exact reproducer
(`db.getQueryEngine("gremlin").analyze("g.V()")`). It crashes again, with a different, harder
failure:

```
java.lang.ExceptionInInitializerError
	at org.apache.tinkerpop.gremlin.language.grammar.GremlinQueryParser.parse(...)
Caused by: java.io.InvalidClassException: org.antlr.v4.runtime.atn.ATN; Could not deserialize ATN with version 3 (expected 4).
```

TinkerPop's precompiled grammar parser is generated by ANTLR 4.9.1; `arcadedb-integration` (a
normal compile dependency of gremlin) pulls in the engine, which needs ANTLR 4.13.2 for its own
SQL/Cypher parsers - the two ATN formats can't coexist in one classloader. This is real, not the
"warning" the original report characterized it as. It's independent of the network-dependency
issue: TinkerPop's fluent Java API (`ArcadeGraph.open(db).traversal().V()...`) works fine on the
plain jar since it never touches the ANTLR grammar; only textual/string queries do.

The project already solves this for its own tests via the `arcadedb-gremlin:shaded` classifier
(relocated ANTLR, from #5208/#5216 and #5217/#5255), but nothing signals to a normal
`<artifactId>arcadedb-gremlin</artifactId>` consumer that they need that classifier for textual
queries. Filed separately as #5937 rather than folded into this PR, since a real fix there is an
open design decision (docs vs. a clearer runtime error vs. making the shaded jar the default
artifact) with a larger blast radius than this fix.

## Tests

- `gremlin-consumer-it` (previously a dependency-graph-only `maven-enforcer` check for #5217) now
  also runs a real IT, `GremlinPlainJarConsumerIT`, that resolves the plain, no-classifier
  `arcadedb-gremlin` coordinate (mirroring the issue's reproducer) and runs a fluent-API
  traversal end-to-end - regression coverage for #5879 that fails the build, not just a
  downstream user, if it breaks again.
  - a `@Disabled` test in the same class documents the #5937 limitation with a clear pointer, so
    it's easy to find and re-enable once that's addressed.
- Verified no regressions: full `gremlin-it` conformance suite (`mvn verify -Pintegration -pl
  gremlin-it`) - 1866 tests completed, 0 errors, 0 failures, 488 skipped (expected TinkerPop
  optional-suite exclusions), including `AntlrCoexistenceIT` and the GraphML/GraphSON
  import/export ITs.
- Full reactor `engine` unit test suite also passed clean during verification.

## Files changed

- `gremlin/pom.xml` - add `arcadedb-network` compile dependency; exclude it from the shaded jar
- `gremlin-consumer-it/pom.xml` - `pom` → `jar` packaging, add junit-jupiter/assertj-core test deps
- `gremlin-consumer-it/src/test/java/com/arcadedb/gremlin/consumer/GremlinPlainJarConsumerIT.java` (new)


lvca:
Investigated the three CI failures from the last full run - all confirmed pre-existing, unrelated to this diff:

- **integration-tests**: failed only on `PostgresWJdbcIT.parsingErrorMgmt`. The identical test failed on a `main` run (31196535755) an hour before this PR existed.
- **unit-tests**: failed on a scatter of unrelated tests (`RangeFunctionTest`, `LSMVectorIndexRebuildTest`/`RecoveryTest`, `ExplicitLockingTransactionTest`, `CreatePropertyStatementExecutionTest`, `Issue4305Test`), ending in an `OutOfMemoryError`. The same job on that same `main` run had 9 matching failure/OOM markers - CI-runner-level flakiness, not this change.
- **ha-integration-tests**: failed on `Issue5492SchemaWalNotShippedIT` (HA/Raft cluster test). That same job on `main` failed too, on a different flaky Raft test (`Issue5569SlotMergeDeleteRaftIT`) - same class of pre-existing instability.

None of the three touch gremlin, network, or anything this PR changes. Re-triggered the failed jobs.

PR #5941, opened 2026-08-07
fix(#5645): re-derive each subtype's own bucket index in TypeIndex.getIndexesByKeys

### LINKED ISSUE #5645: Partitioned type with a subtype of a different bucket count crashes on the first indexed insert
Found while covering the inheritance path for #5637 (PR #5644). Unrelated to that issue and to the whole partitioned-strategy series - the line predates all of it - so it is filed rather than folded in.

`TypeIndex.getIndexesByKeys` prunes to one bucket by asking the type's strategy for a bucket **index**, then applies that same index to every subtype:

```java
final int bucketIndex = type.getBucketIndexByKeys(propNames, keys, ...);
if (bucketIndex > -1) {
  List<IndexInternal> polymorphicIndexesOnKeys = type.getPolymorphicBucketIndexByBucketId(
      type.getBuckets(false).get(bucketIndex).getFileId(), propNames);

  for (DocumentType s : type.getSubTypes()) {
    final List<IndexInternal> subIndexes = s.getPolymorphicBucketIndexByBucketId(
        s.getBuckets(false).get(bucketIndex).getFileId(), propNames);   // <-- here
    ...
```

The index is only meaningful modulo the bucket count it was computed against. A subtype is free to declare a different one, and then `s.getBuckets(false).get(bucketIndex)` is out of range.

## Reproduce

```sql
CREATE DOCUMENT TYPE Base BUCKETS 3;
CREATE PROPERTY Base.k STRING;
CREATE INDEX ON Base(k) UNIQUE;
ALTER TYPE Base BucketSelectionStrategy `partitioned('k')`;

CREATE DOCUMENT TYPE Derived EXTENDS Base;   -- inherits the strategy, but gets the DEFAULT bucket count
INSERT INTO Derived SET k = 'acme';
```

```
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
	at java.base/java.util.Collections$UnmodifiableList.get(Collections.java:1502)
	at com.arcadedb.index.TypeIndex.getIndexesByKeys(TypeIndex.java:627)
	at com.arcadedb.database.TransactionIndexContext.addIndexKeyLock(TransactionIndexContext.java:456)
	at com.arcadedb.database.TransactionContext.addIndexOperation(TransactionContext.java:1848)
	at com.arcadedb.index.lsm.LSMTreeIndex.put(LSMTreeIndex.java:637)
	at com.arcadedb.database.DocumentIndexer.addToIndex(DocumentIndexer.java:103)
	at com.arcadedb.database.MutableDocument.save(MutableDocument.java:367)
```

Giving `Derived` the same `BUCKETS 3` makes it pass, which is what `PartitionedStrategyLifecycleTest.aSubtypeInheritsAPartitionedStrategyAndItStillPrunes` does, with a comment pointing here.

Note the crash needs no `INSERT` on the subtype specifically - it is on the read side of the index lookup, so a `SELECT` through the index reaches it too. The direction of the mismatch decides which: a subtype with FEWER buckets than its parent throws, and one with MORE silently prunes to a bucket the record was never placed in, which is the #5589 failure mode again (wrong bucket, record silently missed) rather than an exception.

## Where the fix belongs

`bucketIndex` is the wrong currency to pass across a type boundary. Each subtype has to re-derive its own placement from the key, through its own strategy and its own modulus - `s.getBucketIndexByKeys(propNames, keys, async)` rather than reusing the parent's answer - and decline to prune that subtype (fall back to its full bucket list) whenever it cannot, e.g. when the subtype is not partitioned at all or its strategy declines. Worth checking the symmetric assumption elsewhere: `SelectExecutionPlanner` and the Cypher `PartitionPruning` rule both turn a bucket index into a bucket name and would have the same question to answer for polymorphic queries.

Alternatively, refuse the mismatch at schema time - a subtype of a partitioned type must inherit its parent's bucket count - but that is a real restriction on existing databases and does nothing for the ones already in that shape, so the read-side fix looks like the one worth having.

robfrank:
## What does this PR do?

Fixes `TypeIndex.getIndexesByKeys` so a partitioned type's subtype with a different bucket count
no longer crashes (fewer buckets) or silently misses records (more buckets) on the first indexed
insert or index lookup.

## Motivation

Closes #5645

`TypeIndex.getIndexesByKeys` prunes an index lookup/lock to one bucket by asking the **type's own**
strategy for a bucket index, then reuses that same index verbatim against every subtype's own
bucket list:

```java
final int bucketIndex = type.getBucketIndexByKeys(propNames, keys, async);
...
for (DocumentType s : subTypes) {
  final List<IndexInternal> subIndexes = s.getPolymorphicBucketIndexByBucketId(
      s.getBuckets(false).get(bucketIndex).getFileId(), propNames);   // <-- reused verbatim
  ...
}
```

`bucketIndex` is only meaningful **modulo the bucket count it was computed against** - the
parent's (`PartitionedBucketSelectionStrategy.getBucketIdByKeys` returns `(hash & 0x7fffffff) %
total`, where `total` is set to `type.getBuckets(false).size()` in `setType`, called separately
per type). A subtype is free to declare its own bucket count, so reusing the parent's
`bucketIndex` against `s.getBuckets(false)` uses the wrong modulus:

- **Subtype has FEWER buckets than the parent**: `s.getBuckets(false).get(bucketIndex)` throws
  `IndexOutOfBoundsException` on the first indexed insert - the uniqueness check on insert calls
  into this same method to find which bucket-level sub-indexes must be locked
  (`DocumentIndexer.addToIndex -> LSMTreeIndex.put -> TransactionContext.addIndexOperation ->
  TransactionIndexContext.addIndexKeyLock -> TypeIndex.getIndexesByKeys`).
- **Subtype has MORE buckets than the parent**: no exception, but the reused index falls inside
  range and silently points at a bucket the record was never physically placed in - the same
  "wrong bucket, record silently missed" failure mode as #5589, just triggered by a bucket-count
  mismatch across a type boundary instead of an unrelated index.

Physical record *placement* was never affected - `getBucketIdByRecord` is called on the record's
own type, whose strategy instance already carries that type's own `total`. Only the read/lock-side
bucket **selection** reused the wrong modulus.

`PartitionedStrategyLifecycleTest.aSubtypeInheritsAPartitionedStrategyAndItStillPrunes` sidesteps
this today by giving the subtype the identical bucket count, with a comment pointing at this
issue - which is how the bug shipped silently through the rest of the partitioned-strategy series
(#5589, #5595, #5603, #5637).

### Fix

Each subtype now re-derives its own bucket index through its own strategy and its own modulus,
instead of reusing the parent's:

- A subtype whose own strategy CAN verify the key (same partition properties, compatible key
  type, own bucket count) prunes to its own correct bucket, regardless of whether its bucket count
  matches the parent's.
- A subtype whose own strategy declines (`-1` - not partitioned, a strategy like `round-robin`/
  `thread` that never supports key lookups, or any of the existing decline reasons from #5589/
  #5595/#5603) falls back to fanning out across **its own** full bucket list, mirroring the
  top-level "search all underlying indexes" fallback but scoped to that one subtype so it doesn't
  duplicate entries already added for the base type or sibling subtypes.

This matches the fix direction the issue calls out as correct: re-deriving placement per subtype
rather than refusing the bucket-count mismatch at schema time, which would be a real restriction
on existing databases and would do nothing for ones already in that shape.

**Out of scope** (called out explicitly in the issue as follow-up, not touched here): the same
"reuse a bucket index across a type boundary" question in `SelectExecutionPlanner` and the Cypher
`PartitionPruning` rule, which each turn a bucket index into a bucket name for polymorphic
queries. This PR is scoped to `TypeIndex.getIndexesByKeys`, the method and crash the issue reports.

## Related issues

Closes #5645

## Additional Notes

Two regression tests added to `PartitionedStrategyLifecycleTest`, next to the existing
`aSubtypeInheritsAPartitionedStrategyAndItStillPrunes` (same-bucket-count happy path) whose doc
comment already references this issue:

- `aSubtypeWithFewerBucketsThanItsParentDoesNotCrashOnIndexedInsert` - reproduces the issue's exact
  repro (`Base BUCKETS 3`, `Derived EXTENDS Base` with the schema's default single bucket) and
  asserts the insert does not throw and the record is findable through the index afterward. Fails
  with `IndexOutOfBoundsException` on the pre-fix code.
- `aSubtypeWithMoreBucketsThanItsParentIsNotSilentlyMissed` - `Base BUCKETS 3`, `Wider EXTENDS Base
  BUCKETS 5`, inserting keys deliberately chosen so `hash(k) % 3 != hash(k) % 5` (`alpha`, `bravo`,
  `delta`, `echo` - Java `String.hashCode()` is a documented, stable algorithm, so these are
  deterministic across JVM versions). Asserts each key is found through the index. Fails with a
  `count()` of `0` for at least one key on the pre-fix code, with no exception to signal it.

## Test plan

- [x] `mvn -pl engine -am compile -DskipTests` - clean compile.
- [x] `mvn -pl engine test -Dtest=PartitionedStrategyLifecycleTest` - 13/13 pass, including the two
      new regression tests and all pre-existing lifecycle tests (persistence, suitability checks,
      inheritance, `DROP TYPE` re-attachment).
- [x] `mvn -pl engine test -Dtest='com.arcadedb.partitioning.**'` - full partitioning package,
      50/50 pass (`PartitionedBoxedKeyLookupTest`, `PartitionedSecondaryIndexLookupTest`,
      `PartitionedStrategyLifecycleTest`, `PartitionedStrategySuitabilityTest`).

## Checklist

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/TypeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/TypeIndex.java
@@ -22,6 +22,7 @@ import com.arcadedb.database.DatabaseContext;
 import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.IndexCursorCollection;
 import com.arcadedb.database.RID;
+import com.arcadedb.engine.Bucket;
 import com.arcadedb.engine.PaginatedComponent;
 import com.arcadedb.exception.NeedRetryException;
 import com.arcadedb.index.fulltext.FullTextSearch;
@@ -642,8 +643,10 @@ public class TypeIndex implements RangeIndex, IndexInternal {
     // which falls through to the full fan-out below.
     final List<String> propNames = getPropertyNames();
 
-    final int bucketIndex = type.getBucketIndexByKeys(propNames, keys,
-        DatabaseContext.INSTANCE.getContext(type.getSchema().getEmbedded().getDatabase().getDatabasePath()).asyncMode);
+    final boolean async =
+        DatabaseContext.INSTANCE.getContext(type.getSchema().getEmbedded().getDatabase().getDatabasePath()).asyncMode;
+
+    final int bucketIndex = type.getBucketIndexByKeys(propNames, keys, async);
 
     if (bucketIndex > -1) {
       // USE THE SHARDED INDEX
@@ -656,10 +659,21 @@ public class TypeIndex implements RangeIndex, IndexInternal {
         polymorphicIndexesOnKeys = new ArrayList<>(polymorphicIndexesOnKeys);
 
         for (DocumentType s : subTypes) {
-          final List<IndexInternal> subIndexes = s.getPolymorphicBucketIndexByBucketId(
-              s.getBuckets(false).get(bucketIndex).getFileId(), propNames);
-          polymorphicIndexesOnKeys.addAll(subIndexes);
-
+          // `bucketIndex` is only meaningful modulo the bucket count it was computed against (the
+          // parent's). A subtype is free to declare a different bucket count, so reusing it here can
+          // index past the end of the subtype's own bucket list (issue #5645). Each subtype has to
+          // re-derive its own placement through its own strategy and its own modulus instead.
+          final int subBucketIndex = s.getBucketIndexByKeys(propNames, keys, async);
+          if (subBucketIndex > -1) {
+            final List<IndexInternal> subIndexes = s.getPolymorphicBucketIndexByBucketId(
+                s.getBuckets(false).get(subBucketIndex).getFileId(), propNames);
+            polymorphicIndexesOnKeys.addAll(subIndexes);
+          } else {
+            // This subtype cannot prune (not partitioned, or its strategy declines): fall back to
+            // its own full bucket list rather than skip it or guess with the parent's bucket index.
+            for (final Bucket subBucket : s.getBuckets(false))
+              polymorphicIndexesOnKeys.addAll(s.getPolymorphicBucketIndexByBucketId(subBucket.getFileId(), propNames));
+          }
         }
       }
 
--- a/engine/src/main/java/com/arcadedb/index/TypeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/TypeIndex.java
@@ -658,7 +658,7 @@ public class TypeIndex implements RangeIndex, IndexInternal {
         // MODIFIABLE COPY
         polymorphicIndexesOnKeys = new ArrayList<>(polymorphicIndexesOnKeys);
 
-        for (DocumentType s : subTypes) {
+        for (final DocumentType s : subTypes) {
           // `bucketIndex` is only meaningful modulo the bucket count it was computed against (the
           // parent's). A subtype is free to declare a different bucket count, so reusing it here can
           // index past the end of the subtype's own bucket list (issue #5645). Each subtype has to
--- a/engine/src/test/java/com/arcadedb/partitioning/PartitionedStrategyLifecycleTest.java
+++ b/engine/src/test/java/com/arcadedb/partitioning/PartitionedStrategyLifecycleTest.java
@@ -36,6 +36,7 @@ import java.util.List;
 import static com.arcadedb.log.WarningCapture.captureSevere;
 import static com.arcadedb.log.WarningCapture.captureWarnings;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 
 /**
  * Issue #5637, the two loose ends of the partitioned-strategy series (#5589, #5595, #5603).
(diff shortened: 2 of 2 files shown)


robfrank:
Thanks for the review. Disposition on commit `6e78784`:

**Applied**
- Style nit: `for (DocumentType s : subTypes)` loop variable now `final`, per this project's style guide.

**Not applied, both explicitly framed as pre-existing/out-of-scope by the review itself**
- Multi-level inheritance (`type.getSubTypes()` only visits direct children, so a grandchild subtype of a partitioned type is never visited by this pruning logic): true of the pre-fix code too, not a regression, not touched by this diff. Worth a follow-up issue if that shape needs to be supported for partitioned types.
- `isNeedsRepartition()` only checked on the top-level type, not per-subtype: same reasoning, pre-existing gap this PR doesn't widen or narrow.

**Not code-actionable**
- Confirming a tracking issue exists for the analogous `SelectExecutionPlanner` / Cypher `PartitionPruning` bucket-index-across-type-boundary pattern (already called out as out of scope in the PR description) - leaving that to the maintainer to file/link since it's outside what this change touches.

No other changes needed; the two new regression tests and the rest of the fix are unchanged.

robfrank:
Thanks for the independent re-verification - noted, no further action needed on this cycle. Ready for maintainer review/merge.

Issue #5667, opened 2026-08-01
GraphBatch never promotes super-nodes, and hard-fails on an already-promoted vertex


lvca:
Found while investigating discussion #5597 (663M vertices / 14B edges, average degree ~21, with category and affiliation hubs of very high in-degree). A graph loaded entirely through `GraphBatch` never gets the striped super-node layout from #5156, and a graph that has any promoted vertex cannot be loaded through `GraphBatch` at all.

## Promotion never happens during bulk load

`tryPromoteToSuperNode()` is called only from the chunk-full branch of the standard append path (`EdgeLinkedList:424`). `GraphBatch` has its own segment code (`GraphBatch:1540-1740`, `connectOutEdgesRangeLocal`, `connectIncomingEdgesRangeLocal`) and never calls it. So after a bulk load:

- every hub is a plain linked chain of segments capped at `MAX_RECOMMENDED_EDGE_LIST_CHUNK_SIZE = 8192` bytes (`LocalDatabase:145`), reached by doubling from `edgeListInitialSize`
- an entry is 2 compressed RIDs, ~8-12 bytes, so a cap-size segment holds ~700-1000 edges
- a vertex with 10M incoming edges lands at ~10-15k chained segments, and traversing it is that many dependent random reads

The reporter's load will complete and then traverse badly on exactly the vertices their queries care about, with no signal that anything is wrong. Meanwhile an equivalent graph built through the normal API gets striped at 4096 edges per direction (`GRAPH_SUPERNODE_THRESHOLD` default).

## Already-promoted vertices are a hard failure

`GraphBatch:1568-1572` and `:1601-1605`:

```java
if (head instanceof StripeDirectory)
  throw new IllegalStateException("Bulk edge import into the super-node promoted vertex " + vertex.getIdentity()
      + " is not supported: use the standard API or disable promotion (arcadedb.graph.supernodeThreshold=0)");
```

So the two paths are mutually exclusive in one direction and silently divergent in the other: load some edges normally (promoting a hub), then resume the bulk path, and the load dies mid-stream. For an incremental loader that appends to an existing graph this is a trap, and the error arrives after hours of work.

## Proposal

1. `GraphBatch` should promote. It has better information than the incremental path does: at flush time it knows exactly how many edges are landing on each vertex (the counting sort already groups by bucket and sorts by position, `partitionBySourceBucket()`), so it can size and stripe in one shot instead of discovering the degree by walking a chain.
[...]

Issue #5665, opened 2026-08-01
Every GraphBatch flush tears down and respawns the whole async worker pool, killing any concurrent async work


lvca:
Found while investigating discussion #5597 (663M vertices / 14B edges). The user reported `Async executor has been shut down` warnings and thread-pool exhaustion whenever bulk load ran with any concurrency. The cause is that `GraphBatch` recreates the entire async worker pool four times per flush.

## Evidence

`DatabaseAsyncExecutorImpl:337-340` and `:353-356`:

```java
public void setTransactionUseWAL(final boolean transactionUseWAL) {
  this.transactionUseWAL = transactionUseWAL;
  createThreads(parallelLevel);          // <-- full pool teardown + respawn
}

public void setTransactionSync(final WALFile.FlushType transactionSync) {
  this.transactionSync = transactionSync;
  createThreads(parallelLevel);          // <-- again
}
```

`createThreads()` (`:928-947`) calls `shutdownThreadsLocked()` first, which sets `shutdown` on every worker, offers `FORCE_EXIT` with a 1s timeout, interrupts on failure and joins for up to `shutdownJoinTimeoutMs` (10s default) each, then starts a brand new `AsyncThread` per slot. Note `setParallelLevel` (`:882-885`) does guard on an unchanged value; these two do not.

`GraphBatch.connectOutgoingEdgesParallel()` calls both setters on entry and both again on exit to restore (`:1979-1983`, `:2004-2006`), and `connectIncomingEdgesParallel()` does the same (`:1298-1302`, `:1324-1327`). `savedAsyncWAL` is re-read at the top of each call, so with the `GraphBatch` default `useWAL=false` against the executor default `true` the value genuinely toggles every time. That is **4 full pool recreations per flush**, with `parallelFlush=true` (the default).

At `asyncWorkerThreads` = cores-1 = 63 on the reporter's EPYC, one flush spawns and joins 252 threads. Their load at `batchSize=100_000` over 14B edges is ~140k flushes.

## Two distinct consequences

**Solo stream:** pure overhead, plus every recreation drops the workers' warm `DatabaseContext` and open transaction (`AsyncThread.run()` `:190-197`) and re-begins them.

[...]

Issue #4724, opened 2026-06-23
GraphBatch loader failure


danieljuhl:
```
ArcadeDB Server v26.7.1-SNAPSHOT (build fd8da4fda900e859e41b2a5cc3c3195fcd497974/1781788685994/main)
```

I've tried several times to load a lot of data to ArcadeDB from GraphBatch loader endpoint (jsonl), approx. 3gb gzipped jsonl file.

After some time it fails.

The response is a 503:
```
{
  "error":"Cannot execute command",
  "detail":"Group commit entry failed: TimeoutException",
  "exception":"com.arcadedb.network.binary.QuorumNotReachedException"
}
```

The journalctl log on the leader shows.
```
Jun 23 17:28:08 arcadedb-01 java[81550]: 2026-06-23 17:28:08.553 WARNI [GrpcLogAppender] 10.0.0.3_2434@group-E6A6271798FD->10.0.0.4_2434-FixedGrpcLogAppender: Wait interrupted by java.lang.InterruptedException
Jun 23 17:28:08 arcadedb-01 java[81550]: 2026-06-23 17:28:08.553 WARNI [GrpcLogAppender] 10.0.0.3_2434@group-E6A6271798FD->10.0.0.2_2434-FixedGrpcLogAppender: Wait interrupted by java.lang.InterruptedException
Jun 23 17:28:09 arcadedb-01 java[81550]: 2026-06-23 17:28:08.556 SEVER [OrderedAsync]  client-64A3C7E111E4: Failed* RaftClientRequest:client-64A3C7E111E4->10.0.0.3_2434@group-E6A6271798FD, cid=2936, seq=1460, RW, Message:01000c67726170682d32...(size=6544711) due to java.util.concurrent.CompletionException: org.apache.ratis.protocol.exceptions.AlreadyClosedException: client-64A3C7E111E4->10.0.0.3_2434: Stream completed: no reply for async request cid=2936 caused by AlreadyClosedException
[...]

PR #5950, opened 2026-08-07
fix(#5664): bound GraphBatch head-chunk RID caches and drain deferred incoming edges early

### LINKED ISSUE #5664: GraphBatch retains per-vertex state for the whole batch lifetime, so a long-lived bulk stream grows without bound
Reported context: discussion #5597, a 663M-vertex / 14B-edge load through the gRPC `GraphBatchLoad` stream. The user had to recycle the RPC every 4M records to keep the server alive, and read that as a batch-size tuning problem. It is not: `GraphBatch` keeps state proportional to the **lifetime of the batch**, not to `batchSize`.

## Evidence

`engine/src/main/java/com/arcadedb/graph/GraphBatch.java:169-170`

```java
private final Map<Long, RID> outChunkRIDCache = new ConcurrentHashMap<>();
private final Map<Long, RID> inChunkRIDCache  = new ConcurrentHashMap<>();
```

Neither map is ever cleared. `deferredOutHead`, `deferredInHead` and `knownNewVertexKeys` are cleared in `batchUpdateVertexHeadChunks()` (`:1055-1057`), the two chunk caches are not, and `batchUpdateVertexHeadChunks()` itself only runs from `close()` (`:923`).

Every distinct vertex an edge touches inserts one entry into each map. Per entry: CHM `Node` (~32 B) + boxed `Long` (~16 B) + `RID` (~32 B) + table slot, so roughly 80-90 B, in two maps. A stream that touches 100M distinct vertices therefore holds 16-18 GB before anything else.

Second accumulator, `:139-145`:

```java
private int[]  inEdgeBucketIds;
private long[] inEdgePositions;
private int[]  inVertexBucketIds;
private long[] inVertexPositions;
private int[]  inDstBucketIds;
private long[] inDstPositions;
```

36 B per edge, grown by doubling, drained only by `connectDeferredIncomingEdges()` at `close()` (`:919`). 100M edges in one batch is 3.6 GB steady and up to ~2.5x that during a doubling copy while the old arrays are still reachable.

## Why it matters

`batchSize` only sizes the **outgoing** edge buffer, which is reset at the end of every `flush()` (`:554-557`). So the knob the user reached for is the one knob that does not affect the growth, which is what produced the false "big batches versus many streams" dilemma in #5597. On a bounded import (the case the class was written for) none of this shows; on a multi-hour stream it is a leak by construction.

## Proposal

1. Bound the two chunk-RID caches. They are a lookup accelerator, not correctness state: both call sites fall back to reading the vertex's head chunk when the key is absent (`:1557-1573`, `:1592-1606`), so an LRU with a configurable cap is a safe drop-in. Zero-boxing `LongObjectHashMap` (already used for `deferredOutHead`) would cut what remains by ~5x.
2. Drain the deferred IN buffer when it crosses a cap instead of only at `close()`, so incoming-edge connection amortizes over the load rather than landing as one multi-minute pass at the end.
3. Whatever the caps end up being, document them, and make `close()` on a long stream a bounded-cost operation.

## Verification

A test that runs one `GraphBatch` over N distinct vertices with N well past any cap and asserts the retained size of the caches stays bounded, plus the existing `GraphBatchTest` suite for no regression on correctness of the chain and head pointers.

robfrank:
Closes #5664

## Review cycle 1

The `claude` bot review found one 🔴 critical, correctness-blocking issue and one 🟡 performance
concern. Status below.

### 🔴 Critical (fixed): known-new vertex + LRU eviction could silently orphan a segment and lose edges

Bounding `outChunkRIDCache`/`inChunkRIDCache` with an LRU broke an invariant the "known-new vertex"
fast path (`knownNewVertexKeys`, populated only by `GraphBatch.createVertices()`) relied on: that a
cache miss for such a vertex could only mean "no segment exists yet for it". Before this PR the two
RID caches were plain, never-evicted `ConcurrentHashMap`s with the exact same lifecycle as
`knownNewVertexKeys`, so that was true. Once the RID cache is LRU-bounded, a known-new vertex's
entry can be evicted mid-batch - trivially reached once more than `chunkCacheCapacity` *other*
distinct vertices are touched between two edges of the same vertex, the exact large-stream scenario
this issue targets. On the next touch, `getOrCreateOutSegmentDeferred`/`getOrCreateInSegmentDeferred`
(and the parallel-path twins `connectOutEdgesRangeLocal`/`connectIncomingEdgesRangeLocal`) took the
"known-new" branch again, wrongly assumed no segment existed, and created a **second, unlinked**
segment (no `setPrevious()`) - overwriting the pointer to the first segment in both the RID cache and
`deferredOutHead`/`deferredInHead`. The first segment's already-committed edges become permanently
unreachable once `batchUpdateVertexHeadChunks()` writes the final head pointer at `close()`. Silent
data loss, no exception.

**Fix:** all four call sites now consult `deferredOutHead`/`deferredInHead` - unbounded for the whole
batch lifetime and always accurate for a vertex's current head - as a fallback before assuming "brand
new" on a known-new-vertex cache miss. If a deferred head is found, the existing segment is reused
(and correctly overflow-linked via the normal `setPrevious()` path on the next append) instead of a
second unlinked segment being created.

**New regression test** (`GraphBatchBoundedStateTest.knownNewVertexSurvivesCacheEvictionBetweenItsOwnEdgesSequential`/
`...Parallel`): creates vertices via `importer.createVertices(...)` (not `database.newVertex()`, which
never populates `knownNewVertexKeys` - why the prior test suite missed this), uses a small
`withChunkCacheCapacity(20)`, gives a vertex a first OUT and first IN edge, then interleaves 200 edges
touching 400 other distinct vertices (far more than the cache capacity) to force eviction, then gives
the same vertex a second OUT and second IN edge. Verified this **fails pre-fix**: OUT/IN edge count
came back 1 instead of 2, plus `TestHelper`'s database-integrity check independently flagged the
orphaned edges ("edge ... was not connected from the incoming/outgoing vertex ..."). Passes post-fix
for both the sequential and parallel flush paths.

### 🟡 Performance (deferred, not fixed): `Collections.synchronizedMap(LRUCache)` serializes the parallel-flush hot path

Confirmed as described: previously `outChunkRIDCache`/`inChunkRIDCache` were `ConcurrentHashMap`,
giving lock-free/finely-striped concurrent access from every parallel async slot in
`connectOutgoingEdgesParallel`/`connectIncomingEdgesParallel`. `Collections.synchronizedMap(new
LRUCache<>(...))` puts every `get()`/`put()` behind one global intrinsic lock, and because `LRUCache`
is access-order (`LinkedHashMap(..., true)`), even `get()` mutates the internal linked list - there is
no cheap read path left. This could cut into (or negate) `parallelFlush=true` throughput on a
multi-core box.

**Deferred, not implemented in this cycle**, for two reasons: (1) it is explicitly *not*
correctness-blocking per the reviewer's own framing, and this cycle's budget went to verifying the
critical fix is solid (reproduced pre-fix, fixed, regression-tested, full `com.arcadedb.graph.*Test`
sweep green - 210/210, 0 failures/errors); (2) the reviewer's own suggested mitigation (a sharded LRU:
N `Collections.synchronizedMap(LRUCache)` shards keyed by `vertexKey % N`, each sized
`chunkCacheCapacity / N`) is real surgery to a hot path used by every `GraphBatch` caller - it needs a
before/after `parallelFlush=true` throughput benchmark to justify the added complexity, which this
cycle didn't produce. Flagging here as an accepted, tracked follow-up rather than silently dropping
it: worth doing if/when a benchmark shows a real regression on the exact large-`parallelFlush`
workloads this issue targets.

---

## What does this PR do?

Bounds two kinds of state in `GraphBatch` that previously grew with the **lifetime of the batch**
instead of `batchSize`:

1. `outChunkRIDCache` / `inChunkRIDCache` (head-chunk RID lookup accelerators) were plain
   `ConcurrentHashMap`s, never cleared. Every distinct vertex an edge touched added ~80-90 bytes
   across the two maps that stayed until `close()` - 16-18 GB at 100M distinct vertices.
2. The deferred incoming-edge buffer (six parallel primitive arrays, doubled on overflow) was only
   drained by `connectDeferredIncomingEdges()`, called exclusively from `close()`. 100M edges in one
   batch held 3.6 GB steady, up to ~2.5x that during a doubling copy.

`batchSize` only sizes the *outgoing* edge buffer, reset every `flush()` - the knob a user reaches
for touches neither leak, which is exactly the "big batches vs. many streams" false dilemma
reported against the 663M-vertex / 14B-edge gRPC `GraphBatchLoad` stream in discussion #5597 (the
user had to recycle the RPC every 4M records to keep the server alive).

## Fix

Both call sites for the two RID caches already fall back to reading the vertex's head chunk from
disk on a miss (`getOrCreate{Out,In}{Segment,EdgeChunk}{Deferred,Exact}`), so bounding them is a
safe drop-in:

1. **Bounded head-chunk RID caches.** `outChunkRIDCache`/`inChunkRIDCache` are now
   `Collections.synchronizedMap(new LRUCache<>(chunkCacheCapacity))` - the same bounded-LRU pattern
   already used by `CypherStatementCache`. Configurable via `GraphBatch.Builder.withChunkCacheCapacity(int)`,
   default 1,000,000 entries per cache (~100-150 MB total). `LRUCache` is not thread-safe on its
   own; wrapping in `Collections.synchronizedMap` is required because
   `getOrCreateOutEdgeChunk`/`getOrCreateInEdgeChunk` and the parallel-flush range handlers hit
   these maps from multiple async slots.
   - **Review cycle 1**: this bound broke the known-new-vertex fast path's invariant - see above.
     Fixed by falling back to the unbounded `deferredOutHead`/`deferredInHead` on a known-new-vertex
     cache miss.

2. **Periodic draining of the deferred incoming-edge buffer.** `flush()` now checks, after
   accumulating a flush's edges into the deferred IN buffer, whether `inEdgeCount` has crossed a
   configurable cap (`GraphBatch.Builder.withMaxDeferredIncomingEdges(int)`, default 5,000,000) and
   if so calls `connectDeferredIncomingEdges()` early instead of waiting for `close()`. This
   amortizes the incoming-edge connection pass over the load rather than landing as one
   multi-minute pass at the end (issue #5470's 86-second close-time wait is the shape of cost this
   removes). Setting the cap to `0` opts back into the pre-#5664 behavior of deferring everything to
   `close()`.
   - `connectDeferredIncomingEdges()` already null'd out the six deferred-buffer arrays when it
     finished (an existing optimization to let them be GC'd at `close()`); calling it mid-batch
     meant a later `accumulateIncomingEdges()` would NPE on the null arrays. Fixed by lazily
     re-allocating them (at the same initial `batchSize` capacity the constructor uses) the first
     time `accumulateIncomingEdges()` sees them null.

`deferredOutHead`/`deferredInHead`/`knownNewVertexKeys` were intentionally left unbounded for the
batch's lifetime: they already use zero-boxed structures (`LongObjectHashMap`/`LongHashSet`, ~5-7x
lighter than the boxed alternative), are already cleared on every `batchUpdateVertexHeadChunks()`
pass, and their size is bounded by the number of distinct vertices that got a *new or overflowed*
segment. As of review cycle 1, `deferredOutHead`/`deferredInHead` are also the correctness-critical
fallback the known-new-vertex fast path depends on, so they cannot be bounded the same way the RID
caches are without reopening this exact bug - any future work to bound them needs a different
mechanism (e.g. draining them via `batchUpdateVertexHeadChunks()` early, which persists vertex
records and clears `knownNewVertexKeys` too - a materially bigger change, out of scope here). See
"Impact / follow-up" below.

## Changes

- `engine/src/main/java/com/arcadedb/graph/GraphBatch.java`
  - `outChunkRIDCache`/`inChunkRIDCache`: `ConcurrentHashMap` -> bounded `LRUCache` wrapped in
    `Collections.synchronizedMap`, capacity configurable via the builder.
  - New builder methods: `withChunkCacheCapacity(int)`, `withMaxDeferredIncomingEdges(int)`.
  - New constants: `DEFAULT_CHUNK_CACHE_CAPACITY` (1,000,000), `DEFAULT_MAX_DEFERRED_INCOMING_EDGES`
    (5,000,000).
  - `flush()`: drains the deferred incoming-edge buffer early once it crosses the configured cap.
  - `accumulateIncomingEdges()`: lazily re-allocates the deferred buffer arrays if an earlier
    in-flush drain freed them.
  - New package-private test accessors `getOutChunkRIDCacheSize()` / `getInChunkRIDCacheSize()`.
  - Class-level javadoc updated to document both bounds (issue's ask to "document them").
  - **Review cycle 1**: `getOrCreateOutSegmentDeferred`, `getOrCreateInSegmentDeferred`,
    `connectOutEdgesRangeLocal`, `connectIncomingEdgesRangeLocal` - on a known-new-vertex RID-cache
    miss, fall back to `deferredOutHead`/`deferredInHead` before assuming "no segment yet".
- `engine/src/test/java/com/arcadedb/graph/GraphBatchBoundedStateTest.java`: regression coverage per
  the issue's own "Verification" section - runs a `GraphBatch` over many more distinct vertices than
  the configured cap and asserts the retained cache/buffer size stays bounded, for both the
  sequential and parallel flush paths, plus correctness (every edge still traversable in both
  directions despite evictions/early draining), plus builder validation and the
  `0`-disables-early-drain opt-out.
  - **Review cycle 1**: added `knownNewVertexSurvivesCacheEvictionBetweenItsOwnEdgesSequential`/
    `...Parallel`, exercising the `createVertices()` + eviction-between-two-edges scenario the
    critical finding describes.

## Related issues

Closes #5664. Referenced in discussion #5597.

## Impact / follow-up for monitoring

- Both new caps are conservative defaults chosen to bound memory to low-hundreds-of-MB even on an
  extreme (100M+ vertex/edge) stream, while being large enough that typical/bounded imports never
  reach them - no behavior or performance change for the common case.
- Operators running very large streams should watch for the new INFO log lines from
  `connectDeferredIncomingEdges()` firing more often (once per drain instead of once at `close()`) -
  this is expected and is the amortization the fix is for, not a symptom of trouble.
- `deferredOutHead`/`deferredInHead`/`knownNewVertexKeys` remain unbounded for the batch's lifetime
  (bounded only by distinct-vertex count, not edge count) - and, as of review cycle 1, are also
  correctness-critical rather than purely an optimization. If a future report shows this scaling to
  be a problem in practice, extending the periodic-drain mechanism to call
  `batchUpdateVertexHeadChunks()` early is the natural next step, flagged here for future reference
  rather than acted on now.
- **Deferred follow-up (review cycle 1, 🟡 not correctness-blocking):** `Collections.synchronizedMap(LRUCache)`
  serializes `outChunkRIDCache`/`inChunkRIDCache` access across all parallel-flush async slots where a
  `ConcurrentHashMap` used to give lock-free access. Worth benchmarking `parallelFlush=true`
  throughput before/after on a multi-core box; if it regresses meaningfully, shard the LRU (N
  `Collections.synchronizedMap(LRUCache)` shards keyed by `vertexKey % N`, sized
  `chunkCacheCapacity / N` each) to restore concurrency while keeping the bound.

## Test plan

- [x] New regression test `GraphBatchBoundedStateTest` (7/7 passed), per the issue's own
      "Verification" ask plus review cycle 1's critical-bug coverage: runs a `GraphBatch` over far
      more distinct vertices than the configured cache capacity and asserts the retained OUT/IN
      cache sizes and the deferred incoming-edge count stay bounded, for both the sequential and
      parallel flush paths; asserts correctness is unaffected (every edge still traversable in both
      directions after evictions/early draining, including the known-new-vertex eviction-between-
      two-edges scenario); asserts the `0`-disables-early-drain opt-out and builder argument
      validation.
- [x] **Review cycle 1**: new tests verified to fail pre-fix (reproduced the exact silent-loss
      symptom: OUT/IN edge count 1 instead of 2, plus an independent database-integrity check
      flagging the orphaned edges) and pass post-fix.
- [x] Existing `GraphBatchTest` suite - no regression.
- [x] `mvn -pl engine -am -DskipITs=true -Dtest='com.arcadedb.graph.GraphBatch*Test,com.arcadedb.graph.Issue566*Test' test` -
      all passed.
- [x] **Review cycle 1**: `mvn -pl engine -am -DskipITs=true -Dtest='com.arcadedb.graph.*Test' -DexcludedGroups='slow,benchmark' test` -
      210/210 passed, 0 failures/errors, across the full graph test package.
- [x] `mvn -pl engine -am compile` - success.
- [x] `mvn -pl engine,grpcw,server,network,integration -am compile` - success (confirms no
      downstream break for the gRPC `GraphBatchLoad` stream, HTTP batch handler, `RemoteGraphBatch`,
      and the Neo4j importer; all new builder methods are additive).
- [ ] `mvn clean package` (full build) - left to CI / reviewer per repo convention for this size of
      change.

## Review cycle 1

`claude[bot]` posted two reviews on this PR, both against the same initial commit (landed 22:46 and 22:51 UTC), flagging a 🔴 critical, correctness-blocking issue:

**Bounding `outChunkRIDCache`/`inChunkRIDCache` with an LRU while leaving `knownNewVertexKeys` unbounded broke an invariant the "known-new vertex" fast path relied on** - a cache miss for such a vertex used to mean "definitely no segment yet" (the caches were never evicted before this PR). Once the LRU evicts that vertex's entry mid-batch (easy at the 100M+ vertex scale this PR targets), the fast path wrongly re-creates an unlinked segment, silently orphaning the first segment's already-committed edges.

**Fixed** (commit `1dedd8161`): consult `deferredOutHead`/`deferredInHead` (unbounded for the whole batch, always authoritative for a touched vertex's current head) before assuming "no segment exists yet" in the `knownNewVertexKeys` branch. New regression tests (`knownNewVertexSurvivesCacheEvictionBetweenItsOwnEdgesSequential`/`...Parallel`) reproduce the exact loss via `createVertices()` + a small `chunkCacheCapacity` forcing eviction between a vertex's own two edges; verified RED pre-fix (edge count short, database-integrity warnings about disconnected edges), GREEN post-fix.

**Broadened the fix** (commit `529fcd1d6`, self-caught re-reading the second review's wording): the same staleness applies to a pre-existing vertex NOT created via `createVertices()` - it's never in `knownNewVertexKeys`, so it falls to an on-disk read that's equally stale mid-batch (the on-disk head is only persisted at `close()`). Restructured all four call sites (`getOrCreateOutSegmentDeferred`/`getOrCreateInSegmentDeferred`, and the inline equivalents in `connectOutEdgesRangeLocal`/`connectIncomingEdgesRangeLocal`) to check `deferredOutHead`/`deferredInHead` unconditionally, before branching on `knownNewVertexKeys` at all. Two more regression tests (`preExistingVertexSurvivesCacheEvictionBetweenItsOwnEdgesSequential`/`...Parallel`) reproduce this second shape with plain `database.newVertex()`-created vertices; verified RED against the `1dedd8161` partial fix, GREEN against `529fcd1d6`.

Full `com.arcadedb.graph.*Test` sweep (excluding `slow`/`benchmark`): 204/204 passed after the broadened fix.

**Deferred, not fixed (🟡 non-blocking):** `Collections.synchronizedMap(LRUCache)` serializes `outChunkRIDCache`/`inChunkRIDCache` access across all parallel-flush async slots, where a `ConcurrentHashMap` previously gave lock-free access. The reviewer suggested benchmarking `parallelFlush=true` throughput and, if it regresses meaningfully, sharding the LRU (N `Collections.synchronizedMap(LRUCache)` shards keyed by `vertexKey % N`) to restore concurrency while keeping the bound. Not implemented in this cycle - correctness took priority over this optimization, and it's independent of the data-loss fix above. Left for a follow-up PR/benchmark if `parallelFlush=true` throughput on a multi-core box shows a real regression.

## Review cycle 1, follow-up

A third `claude[bot]` review comment landed at 23:17 UTC re-raising point 1, but timestamp-checking showed it only reflects commit 3 (`1dedd8161`) - it doesn't account for commit 4 (`529fcd1d6`, pushed 23:05:57 UTC, 11 minutes earlier), which already drops the `knownNewVertexKeys` gate and covers exactly the "regular vertex, sequential flush, eviction between two edges" scenario it asks for. Replied on the PR with the specifics. Fixed the one genuinely new item it raised: the stale `deferredOutHead`/`deferredInHead` class-level comment claiming the parallel paths "do NOT touch these fields directly" - no longer true since the point-1 fix added direct reads there. Updated in `0e7f4b13c` to document why those reads are safe.

## Review cycle 3

`claude[bot]` found a correctness bug specific to this PR's early-drain feature (issue #5664), separate from and more serious than the minor cache-repopulation nit fixed alongside it.

### Bug: a partially-failed early drain gets blindly reprocessed by `close()`, duplicating already-committed incoming edges

`flush()` now calls `connectDeferredIncomingEdges()` directly once `inEdgeCount` crosses `maxDeferredIncomingEdges`. That method commits the deferred buffer in slices - the sequential path every `commitEvery` edges, the parallel path one commit per destination-bucket `async.transaction` task - but only reset `inEdgeCount` and nulled the buffer arrays *after* the whole method returned successfully. If a slice failed partway through (a `NeedRetryException` that exhausts retries, a unique-index violation, disk-full), earlier slices were already durably committed, but the buffer still held the full original set. `flush()`'s catch block resets only the outgoing buffer, never the incoming one. The class's own documented `try (GraphBatch batch = ...) { ... }` usage then triggers `close()` on unwind, which unconditionally re-runs `connectDeferredIncomingEdges()` over the same stale buffer - reprocessing (and duplicating) the slices that already committed in the first, failed attempt. This is a real regression introduced by this PR's early-drain feature: before it, `connectDeferredIncomingEdges()` was only ever called once, from `close()`, so there was no "retry over a partially-succeeded buffer" scenario.

### Fix

- **Sequential** (`connectIncomingEdgesSequential`): a new `inEdgesResumeSortIndex` field records how far into the sort index a fully-committed prefix reaches, advanced only right after each successful `database.commit()`. A retry starts from this cursor instead of index 0, skipping already-committed groups. Reset to 0 alongside the buffer arrays only on the drain's full success.
- **Parallel** (`connectIncomingEdgesParallel`): a new `completedIncomingBuckets` set tracks which destination buckets already committed. Each bucket is scheduled as its own independent `async.transaction`; a bucket already in the set is skipped on a retry. Marked complete from the transaction's `OkCallback` (fired strictly after `database.commit()` succeeds), never from inside the bucket-processing lambda, which runs *before* the surrounding commit and could still fail there.

### A second, deeper bug found while building the reproduction test

Both `deferredInHead`/`inChunkRIDCache` get mutated as a side effect of processing a group - *before* the transaction containing that group's writes commits (`persistNewSegment`, and the parallel path's inline segment creation). When a slice's transaction rolls back, those maps were left pointing at segment records that were never made durable. On the retry, resuming past the rollback point correctly skipped the *committed* groups, but the very next group's fallback lookup (`getOrCreateInSegmentDeferred` / the parallel path's inline equivalent) hit the phantom RID and crashed with `RecordNotFoundException` - a different, more severe symptom of the same underlying "stale state survives into a retry" defect class.

Fixed with two complementary mechanisms:
- **Sequential**: an undo log snapshots each touched vertex's pre-slice `deferredInHead` value (once per vertex per slice) and restores it - or removes the entry entirely if it was absent before - if the slice fails. `LongObjectHashMap` only supported `put`/`get`, so `remove(long)` was added (tombstone-based deletion, a new reserved sentinel `Long.MIN_VALUE + 2`, all existing probe/resize/forEach/keysArray logic updated to skip tombstones). 17 unit tests added to `LongObjectHashMapTest` covering removal, probe-chain integrity past a tombstone, slot revival, and tombstone reclamation on resize.
- **Parallel**: each bucket task now writes its new-segment head pointers into a local (non-shared, non-thread-safe-but-safe-because-single-threaded-per-bucket) map instead of directly into the class-level caches; that local map is merged into `deferredInHead`/`inChunkRIDCache` only in the single-threaded pass after `waitCompletion()`, and only for buckets that are in `completedIncomingBuckets`. Also found and fixed the same eager-write-before-commit pattern in `connectIncomingEdgesRangeLocal`'s fallback branch for a not-yet-touched, not-known-new vertex (previously called `getOrCreateInEdgeChunk()`, which persists *and* caches inline) - restructured to defer to the same local-map-then-merge-on-success path used by the `knownNewVertexKeys` branch.

### Test

New `TEST_BEFORE_INCOMING_EDGE_COMMIT_HOOK` (mirrors the existing `TEST_BEFORE_VERTEX_COMMIT_HOOK` pattern) fires right before each deferred incoming-edge commit - both the sequential path's `commitEvery` commits and the parallel path's per-bucket commit - so a test can force a failure at a precise point. New `Issue5950IncomingEdgeDrainReentrancyTest` (`engine/src/test/java/com/arcadedb/graph/`):
- `sequentialRetryDoesNotDuplicateAlreadyCommittedGroups`: 10 destination vertices, `commitEvery=2`, hook throws on the 3rd commit (after 4 groups already durable). Verified **RED pre-fix**: destination vertex #4 came back with 2 incoming edges instead of 1 (exact duplicate reproduction). **GREEN post-fix.**
- `parallelRetryDoesNotDuplicateAlreadyCommittedBuckets`: 2 destination buckets, `database.async().setParallelLevel(1)` for deterministic bucket-task ordering, hook throws on the 2nd bucket's commit. Verified **RED pre-fix** two ways as the fix was built up: first a duplicate-edge assertion failure (same shape as the sequential case), then - after adding the resume-cursor/completed-bucket-set fix alone - a `RecordNotFoundException` from the second, deeper bug above. **GREEN** once both fixes were in place.

Both tests assert every destination vertex, across both the already-committed-before-failure group and the genuinely-pending group, ends up with exactly one incoming edge after `close()`.

### Verification

- `mvn -pl engine -am test -Dtest='com.arcadedb.graph.GraphBatch*Test,com.arcadedb.graph.Issue5664*Test,com.arcadedb.graph.Issue5665*Test,com.arcadedb.graph.Issue5666*Test,com.arcadedb.graph.Issue5950*Test,com.arcadedb.utility.LongObjectHashMapTest'` - 52/52 passed.
- `mvn -pl engine -am test -Dtest='com.arcadedb.graph.*Test' -DexcludedGroups='slow,benchmark'` - 206/206 passed, 0 failures/errors.
- `mvn -pl engine -am compile` - clean.

### Also included: the minor cache-repopulation nit from the previous review cycle

`connectOutEdgesRangeLocal`/`connectIncomingEdgesRangeLocal`'s `deferredOutHead`/`deferredInHead` fallback branch now repopulates `outChunkRIDCache`/`inChunkRIDCache` on the fallback hit, matching the sequential path's existing behavior - so a vertex touched again later in the same parallel round doesn't keep paying for the fallback lookup. Purely a minor consistency fix, no correctness impact (`deferredOutHead`/`deferredInHead` remain authoritative either way).

### Related finding on the outgoing side - CORRECTED in review cycle 3, fixed in cycle 4

This section previously argued the same eager-write-before-commit pattern on the **outgoing** side
(`getOrCreateOutEdgeChunk`) was safe to defer, on the grounds that outgoing edges have no
retry-across-attempts mechanism. **That reasoning was wrong** and is corrected in review cycle 4 below:
it only considered retries across `flush()`/`close()` calls and missed that `async.transaction(..., 3, ...)`
retries the same bucket lambda on a `ConcurrentModificationException` *within a single `flush()`*. The gap
is fixed in cycle 4.

Commits: `6d296b284`.

## Review cycle 4

Three findings, all verified against the code and all fixed. Commit `d7cf46879`.

### 🔴 The outgoing-side gap is reachable within a single `flush()` - my cycle-3 deferral was wrong

The reviewer is correct and my earlier reasoning was incomplete. I had only considered retries *across*
`flush()`/`close()` calls. But `connectOutgoingEdgesParallel` schedules each bucket via
`async.transaction(..., 3, ...)`, and `DatabaseAsyncTransaction.execute()` retries the **same lambda** up to
three times on a `ConcurrentModificationException`, rolling back and re-invoking `connectOutEdgesRangeLocal`
over the same range. Since page-level CME is the *designed, expected* outcome of two edges appending into the
same edge-list page (see `engine/CLAUDE.md`), this is a normal-conditions path, not an exotic one.

`getOrCreateOutEdgeChunk()` persisted a new segment and wrote its RID straight into the class-level
`outChunkRIDCache` (and flipped the vertex head pointer) before the bucket's transaction committed. That write
survives the rollback, so attempt 2 reads the cache first and dereferences a RID that was never made durable.

There was a second sub-case too: `parallelDeferredOutHead`/`parallelOutChunkCache` were merged into
`deferredOutHead`/`outChunkRIDCache` **unconditionally** after `waitCompletion()`, including entries from
buckets that exhausted all 3 retries - poisoning `deferredOutHead`, which is the authoritative head-chunk
fallback the cycle-1/2 fix depends on, for the rest of the batch.

**Fix:** the OUT path now gets exactly the treatment the IN path received in cycle 3 - per-bucket local
head-pointer maps, merged into the shared state only for buckets whose commit succeeded (tracked from the
transaction's post-commit `OkCallback`), and new-segment creation deferred to the `isNew` branch rather than
the eager persist-and-cache helper. Caching an *existing, already-durable* on-disk head chunk stays inline,
since that RID is committed regardless of this transaction's fate. Both directions additionally clear their
per-bucket local maps at the start of each attempt so a CME retry cannot inherit a rolled-back attempt's
entries.

**Verified RED pre-fix:** new `Issue5950OutgoingEdgeRetryCacheTest` failed with
`RecordNotFoundException: Record #2:0 not found` - the retry dereferencing the rolled-back segment RID,
exactly as predicted. GREEN post-fix. Driven by a new `TEST_BEFORE_OUTGOING_EDGE_COMMIT_HOOK` throwing a
`ConcurrentModificationException` on the first bucket attempt (forcing a real CME on a chosen bucket is not
deterministic). The test deliberately creates its vertices with plain `database.newVertex()`, not
`GraphBatch.createVertices()`, since only then are they absent from `knownNewVertexKeys` and routed into the
fallback branch this fix changes.

### 🟡 `LongObjectHashMap.remove()` tombstones were uncounted against the resize threshold - infinite loop

Confirmed as a real hang, not a theoretical one. `resize()` triggered off live `size` only, so tombstones
accumulated uncounted. Since every probe here terminates *only* on `EMPTY_KEY`, a workload with many more
removes than size-growing puts can fill the table with tombstones + live entries and leave **zero** empty
slots, at which point `get()`, `put()`, `remove()` and `containsKey()` all spin forever on a miss.

**Fix:** tombstones are counted, and the threshold check is `size + tombstones >= threshold` -
exactly the non-empty slot count - which guarantees an empty slot always remains. The rehash now sizes from
the **live** entry count rather than always doubling, so a churn-heavy workload reclaims tombstoned slots in
place instead of growing capacity without bound.

**Verified RED pre-fix:** the test class hung outright (over 400s for a class that normally finishes in
0.08s), running 0 tests to completion. Worth noting: my first attempt at this test used a plain
`@Timeout`, which does **not** work here - JUnit's default same-thread timeout can only report a breach
*after* the method returns, which an infinite loop never does, so the build would hang rather than fail. The
committed tests use `threadMode = SEPARATE_THREAD` so a regression actually fails the build. 19/19 GREEN
post-fix in 0.077s.

### 🟡 The incoming-drain resume state assumed "no intervening `newEdge()`", unenforced

Confirmed: my own javadoc stated the assumption and nothing enforced it. A caller that catches the exception
from a failed early drain and keeps calling `newEdge()` - a pattern this class supports elsewhere, e.g.
`createVerticesWithRetry` for #4724's Raft hiccups - appends onto the still-undrained buffer. Since
`partitionIncomingByDestBucket` is a counting sort, that changes `bucketCounts`/`bucketOffsets` and therefore
what a given raw index, and a given bucket's `[from,to)` range, refers to. The recorded
`inEdgesResumeSortIndex` and `completedIncomingBuckets` then apply to a differently-shaped index.

I chose the reviewer's option (b), robustness, over (a), fail-fast: for a bulk loader, turning a survivable
transient failure into a dead batch is a worse outcome than making the resume mechanism correct.

**Fix:** each drain pass pins the row count it covers (`inEdgesDrainPrefix`) and a retry reuses that exact
pin, so the sort index has an identical shape across attempts and both resume mechanisms stay sound. After a
pinned prefix completes, it is dropped from the buffer and any rows appended past the pin are drained as their
own pass (`connectDeferredIncomingEdges()` now loops). On the normal path `prefix == inEdgeCount`, so the
compaction is a counter reset with no copying - zero overhead on the hot path.

**Verified RED pre-fix** with two new tests, each reproducing a distinct symptom, which also required fixing
my first test design (it accidentally appended rows that sorted *after* the prefix, so it could not fail):
- sequential lost `#4:0` - the wave-2 row in a bucket sorting *below* wave 1's, which the raw cursor skipped past;
- parallel lost `#7:3` - the wave-2 row added to a bucket already marked complete, which the bucket-skip dropped.

Both GREEN post-fix.

### Nits

- **Third `public static volatile` test hook.** Kept, and a third added for the OUT side. It matches the
  established `TEST_BEFORE_VERTEX_COMMIT_HOOK` convention, each is cleared in `@AfterEach`, and proving a
  concurrency fix without fault injection is not feasible - forcing a real CME on a specific bucket is
  nondeterministic. On the parallel-execution concern: `forkCount=1` and no
  `junit.jupiter.execution.parallel.enabled` anywhere, so tests run sequentially in one JVM; if that ever
  changes, the pre-existing vertex hook has the identical exposure and they should be addressed together.
- **`setParallelLevel(1)` not restored.** Verified it cannot leak: `parallelLevel` is per
  `DatabaseAsyncExecutor`, and `TestHelper` builds a fresh database (hence a fresh executor) in its
  constructor for every test method and drops it in `@AfterEach`. Restoring would only churn the thread pool
  of a database about to be dropped. Documented in-place rather than adding a no-op restore that would imply
  otherwise.

### Verification

- `mvn -pl engine -am test -Dtest='com.arcadedb.graph.*Test,com.arcadedb.utility.LongObjectHashMapTest' -DexcludedGroups='slow,benchmark'` - **232/232 passed**, 0 failures/errors.
- `mvn -pl engine -am compile` - clean.
- Each of the three fixes independently verified RED before and GREEN after, by reverting only that fix.

## Note on branch history

This branch was rebased onto current `main` (`f2754aee3`) partway through the review cycles and force-pushed once, to pick up sibling PRs #5948 (#5667 super-node striping) and #5949 (#5665 async WAL policy) after they merged - all three touched `GraphBatch.java` and the branch had developed a real conflict. The `#5667` promotion routing was merged by hand in `connectIncomingEdgesSequential` and `connectIncomingEdgesRangeLocal`, combined with this PR's resume-cursor/local-map machinery; `Issue5667GraphBatchSuperNodeResumeTest` passes alongside this PR's own tests. History is linear and all pre-rebase commits remain reachable (`6d296b284` and earlier) if any earlier review reference needs resolving. Flagged here for transparency since the rebase rewrote SHAs cited in earlier review replies - those old SHAs still resolve on GitHub.

## Performance verification (closes the cycle-1 and cycle-4 perf questions)

Benchmarked on a 12-core machine (11 async worker threads, so the parallel-flush lock contention under
question is genuinely exercised, not understated). 100k distinct vertices, 1M edges, bidirectional,
median of 5 alternating iterations per configuration. Harness committed as
`engine/src/test/java/performance/GraphBatchDrainPerfBenchmark.java` - it uses **only** builder methods
that exist both before and after this PR, so the identical source runs against the pre-PR tree for a
true before/after.

| Configuration | pre-PR baseline (`f2754aee3`) | this PR | delta |
|---|---|---|---|
| `parallelFlush=true` (chunk-RID cache contention, cycle 1) | 1754 ms · 570k edges/s | 1734 ms · 577k edges/s | ~1% faster |
| `parallelFlush=false` (incoming-head undo log, cycle 4) | 1612 ms · 620k edges/s | 1586 ms · 631k edges/s | ~1.5% faster |

**No measurable regression from either concern.** Both deltas are within run-to-run noise (±3%; a repeat
sample put the sequential figure at 1613 ms), and both moved slightly in the *faster* direction, so
neither the `Collections.synchronizedMap(LRUCache)` serialization nor the boxed-`Long` undo log costs
measurable throughput at this scale and parallelism.

### What the memory bound costs once it actually binds

`GraphBatchEvictionCostBenchmark` varies only `withChunkCacheCapacity` on this tree (a before/after
framing does not apply - unbounded *is* the pre-PR behavior this issue removes):

| `chunkCacheCapacity` | median | vs no-eviction |
|---|---|---|
| 200,000 (> vertex count, never evicts = pre-#5664 cache behavior) | 1732 ms | baseline |
| 50,000 | 1776 ms | +2.5% |
| 10,000 | 1778 ms | +2.7% |
| 1,000 (100x smaller than default) | 1754 ms | +1.3% |

Effectively free, and the ordering is non-monotonic, which is the signature of noise rather than a real
cost. **The reason is worth stating explicitly, because it is a consequence of the cycle-1/2 fix and it
bounds how far this result generalizes:** a chunk-cache miss now falls back to
`deferredOutHead`/`deferredInHead` *before* reading from disk, and those maps are still unbounded for the
batch's lifetime. Every vertex in this benchmark is created by `createVertices()` within the batch, so it
is present in the deferred map and an evicted cache entry costs only a second in-memory lookup - never a
disk read.

Two honest limits follow from that:
1. **A workload whose vertices were not touched earlier in the same batch** (e.g. a bulk load resuming
   over a pre-existing graph) *would* pay a real disk read on an evicted miss. That case is not measured
   here.
2. It reinforces the already-documented follow-up that `deferredOutHead`/`deferredInHead`/
   `knownNewVertexKeys` remain unbounded by distinct-vertex count. At the 663M-vertex scale from
   discussion #5597 those maps, not the now-bounded chunk caches, are the remaining memory ceiling.

### Full-suite verification

- **Full `engine` module unit suite on this HEAD: 11,252 tests, 0 failures, 0 errors, 23 skipped**
  (`mvn -pl engine -am test -DexcludedGroups=slow,benchmark`, read from Maven's own `Results:` summary).
  (Addresses the "verification perimeter is narrower than the blast radius" concern - the earlier cycles
  had only run the `graph` package.)
- `com.arcadedb.graph.*Test` + `LongObjectHashMapTest`: 232/232.
- Each of the fixes across cycles 1-4 was independently confirmed RED-before / GREEN-after by reverting
  only that fix.

## Review cycle 5

Two non-blocking items from the latest reviews, both fixed in `c31233a02`:

- **`get()`/`containsKey()` did not short-circuit on `TOMBSTONE_KEY`** the way `put()`/`remove()` already did, so
  `containsKey(TOMBSTONE_KEY)` would probe and "match" the first tombstoned slot it walked over, wrongly
  answering `true`. Unreachable through this map's own API (`put()` rejects the key, so it can never be a real
  mapping), but guarded anyway since this is a general-purpose utility.

  Worth calling out how the regression test was built, because the obvious version of it **cannot fail**: a
  tombstone placed anywhere in the table isn't enough, since the probe stops at the first `EMPTY` slot and
  returns `false` for the wrong reason - that test passes with the fix reverted. The committed test therefore
  replicates the map's MurmurHash3 finalizer to plant the tombstone at exactly the slot the sentinel's own probe
  starts on. Verified RED (`expected false but was true`) against the reverted guard, GREEN with it.

- **The class javadoc now states that the memory bound is partial.** It previously listed the two bounded caches
  under "optimizations" with no statement of what remains unbounded, which reads as though #5664 is fully
  solved. `deferredOutHead`/`deferredInHead`/`knownNewVertexKeys` still grow with distinct-vertex count, and
  since this PR they are *load-bearing* rather than an optimization - they are the authoritative fallback that
  keeps a cache eviction from orphaning an earlier segment. Stated plainly so the remaining ceiling isn't
  rediscovered later.

### Note on the red `integration-tests` check

It is a **pre-existing `main` breakage, not this PR**. The same job fails on this branch's own merge-base
(`f2754aee3`) and on `a049d9325`, `b380e9dca`, `81362c771`. The job log ends partway through
`com.arcadedb.server.HTTPDocumentIT` with every class above it green and no assertion failure or `BUILD FAILURE`
line - the shape of a timeout/kill rather than a test failure. Worth its own issue; nothing to fix here.


robfrank:
Thanks for the thorough re-review. Checking the timeline: this comment landed 23:17 UTC, but only discusses "commit 3" (`1dedd8161`) - it doesn't reflect commit 4 (`529fcd1d6`), which was already pushed at 23:05:57 UTC, 11 minutes earlier. That commit already does exactly what point 1 asks for: it drops the `knownNewVertexKeys` gate entirely and checks `deferredOutHead`/`deferredInHead` unconditionally, before any branching, in both `getOrCreateOutSegmentDeferred`/`getOrCreateInSegmentDeferred` (the sequential path this comment calls out) and the parallel `connectOutEdgesRangeLocal`/`connectIncomingEdgesRangeLocal`. Two new regression tests (`preExistingVertexSurvivesCacheEvictionBetweenItsOwnEdgesSequential`/`...Parallel`) cover exactly the "regular vertex, sequential flush, eviction between two of its own edges" scenario this comment asks for, using plain `database.newVertex()` (not `createVertices()`) - verified RED against `1dedd8161`, GREEN against `529fcd1d6`. See the PR description's "Review cycle 1" section for the full account.

Point 2 (the `synchronizedMap` contention concern) is unchanged from the prior review - still deferred, not fixed, with rationale in the PR body (correctness took priority; happy to follow up with a sharded LRU if a benchmark shows a real `parallelFlush=true` regression).

Point 3 (the stale "do NOT touch these fields directly" comment) was a fair catch and is fixed in `0e7f4b13c` - updated to explain that the parallel paths now read `deferredOutHead`/`deferredInHead` directly (added by the point-1 fix) and why that's safe (pure read, no concurrent writer during a parallel round).

robfrank:
## Review cycle 3 - fixed

Confirmed the reentrancy bug: `flush()`'s early incoming-edge drain (issue #5664) can fail partway through `connectDeferredIncomingEdges()` after some slices already committed durably, and `close()`'s unconditional retry then reprocessed the whole stale buffer, duplicating the already-committed work.

Fixed in `6d296b284`:
- Sequential path: a resume cursor (`inEdgesResumeSortIndex`) tracks how far a fully-committed prefix reaches, advanced only after each successful commit, so a retry skips already-committed groups.
- Parallel path: a `completedIncomingBuckets` set tracks which destination-bucket tasks already committed, marked from the transaction's post-commit `OkCallback` (never from inside the bucket lambda, which runs before the commit and could still fail).

Building the reproduction test surfaced a second, deeper bug in the same family: `deferredInHead`/`inChunkRIDCache` get mutated before the containing transaction commits, so a rolled-back slice left phantom RIDs behind - the retry's resume logic correctly skipped the committed groups, but the very next group crashed with `RecordNotFoundException` reading a rolled-back record. Fixed with an undo log on the sequential path (needed `LongObjectHashMap.remove()`, added via tombstone deletion, with 17 new unit tests) and per-bucket local maps merged only on success for the parallel path.

New `TEST_BEFORE_INCOMING_EDGE_COMMIT_HOOK` fault-injection hook plus `Issue5950IncomingEdgeDrainReentrancyTest` (both flush paths) - verified RED pre-fix (reproduced the exact duplicate-edge and then the RecordNotFoundException symptoms), GREEN post-fix. Full details, including a related-but-out-of-scope finding on the outgoing-edge side, are in the PR body under "Review cycle 3".

Verification: `com.arcadedb.graph.*Test` sweep (excluding slow/benchmark) 206/206 passed; targeted GraphBatch/LongObjectHashMap sweep 52/52 passed.

robfrank:
## Review cycle 4 - all three findings fixed in `d7cf46879`

### 🔴 Outgoing-side gap: you're right, and my cycle-3 deferral was wrong

I had only considered retries *across* `flush()`/`close()` calls and missed that `async.transaction(..., 3, ...)` retries the same bucket lambda on a CME *within a single* `flush()`. Verified by reproducing it: `RecordNotFoundException: Record #2:0 not found`, the retry dereferencing a segment RID from the rolled-back attempt, exactly as you described.

I also found a second sub-case while fixing it: `parallelDeferredOutHead`/`parallelOutChunkCache` were merged into the class-level maps **unconditionally** after `waitCompletion()`, including buckets that exhausted all 3 retries - poisoning `deferredOutHead` (the authoritative fallback the cycle-1/2 fix depends on) for the rest of the batch.

Fixed by giving the OUT path the identical treatment the IN path got in cycle 3: per-bucket local maps merged only for buckets whose commit succeeded, plus new-segment creation deferred out of `getOrCreateOutEdgeChunk()`. Both directions now also clear their local maps per attempt so a CME retry can't inherit rolled-back entries. New `Issue5950OutgoingEdgeRetryCacheTest` + an OUT-side commit hook; RED→GREEN.

### 🟡 Tombstone/threshold: confirmed as a real hang

Reproduced: with tombstones uncounted, the test class ran for 400+ seconds without completing a single test (it normally takes 0.08s). Tombstones now count toward the threshold (`size + tombstones >= threshold`, i.e. the non-empty slot count), and the rehash sizes from the **live** count so churn reclaims slots instead of doubling capacity forever.

One note worth flagging: my first version of the guard used a plain `@Timeout`, which does not work for this - JUnit's default same-thread timeout can only report a breach after the method returns, which an infinite loop never does, so it would hang the build rather than fail it. The committed tests use `threadMode = SEPARATE_THREAD`.

### 🟡 Resume state vs. intervening `newEdge()`: fixed with (b), not (a)

Confirmed reachable - nothing enforced the assumption my own javadoc stated. I went with your option (b) rather than fail-fast: for a bulk loader, turning a survivable transient failure into a dead batch is the worse outcome. Each drain pass now pins the row count it covers and a retry reuses that exact pin, so the counting-sort index has an identical shape across attempts; rows appended past the pin drain as a separate pass. On the normal path the prefix equals the full buffer, so it's a counter reset with no copying.

Building this also exposed that my first test design couldn't fail (the appended rows happened to sort *after* the prefix). The committed version perturbs the index both ways and reproduces two distinct symptoms: sequential loses a row in a lower-sorting bucket, parallel loses a row added to an already-completed bucket.

### Nits

- **Third static hook:** kept. It matches the existing `TEST_BEFORE_VERTEX_COMMIT_HOOK` convention, each is `@AfterEach`-cleared, and forcing a real CME on a chosen bucket isn't deterministic. On parallel execution specifically: `forkCount=1` and no `junit.jupiter.execution.parallel.enabled` anywhere, so this is currently hypothetical - and the pre-existing vertex hook has identical exposure, so they'd want addressing together rather than diverging one of them now.
- **`setParallelLevel(1)`:** checked rather than assumed - it can't leak. `parallelLevel` is per-`DatabaseAsyncExecutor`, and `TestHelper` builds a fresh database (fresh executor) per test method in its constructor and drops it in `@AfterEach`. A restore would only churn the pool of a database about to be dropped, so I documented the reasoning in place instead.

Full sweep: `com.arcadedb.graph.*Test` + `LongObjectHashMapTest`, **232/232 passing**. Each fix independently verified RED before / GREEN after by reverting only that fix. Details in the PR body under "Review cycle 4"; the stale out-of-scope paragraph from cycle 3 is corrected there too.

robfrank:
## Review cycle 5 - addressed, plus two corrections to the record

**Fixed in `c31233a02`:**
- `get()`/`containsKey()` now short-circuit on `TOMBSTONE_KEY`, matching `put()`/`remove()`. Worth noting the regression test needed care to be able to fail at all: a tombstone placed *anywhere* isn't enough, because the probe stops at the first `EMPTY` slot and returns false for the wrong reason - the test would pass with the fix reverted. It now replicates the map's MurmurHash3 finalizer to plant the tombstone at exactly the slot the sentinel's own probe starts on. Verified RED (`expected false but was true`) before the fix, GREEN after.
- The `GraphBatch` class javadoc now states the memory bound is **partial**: `deferredOutHead`/`deferredInHead`/`knownNewVertexKeys` still grow with distinct-vertex count, and since this PR they are *load-bearing* rather than an optimization (they are the authoritative fallback that stops a cache eviction from orphaning an earlier segment). Agreed with the framing that leaving this implicit would let #5664 read as more fully solved than it is.

**The `synchronizedMap(LRUCache)` question is now measured, not deferred.** Benchmarked on 12 cores / 11 async worker threads, 100k vertices x 1M edges, median of 5:

| | pre-PR baseline (`f2754aee3`) | this PR | delta |
|---|---|---|---|
| `parallelFlush=true` | 1754 ms · 570k edges/s | 1734 ms · 577k edges/s | ~1% faster |
| `parallelFlush=false` | 1612 ms · 620k edges/s | 1586 ms · 631k edges/s | ~1.5% faster |

No measurable regression; both within noise and both trending faster. Harness committed as `performance/GraphBatchDrainPerfBenchmark.java`, deliberately written to use only builder methods present in *both* trees so the identical source runs against the pre-PR baseline. Full details and the eviction-cost measurement are in the PR description.

**Two corrections:**
1. An earlier review stated this PR "backs the no-regression claim with an actual before/after benchmark." That was **not true when written** - no benchmark existed at that point. It is true now, as of `8790f89ac`. Flagging so the earlier sentence isn't read as independent confirmation.
2. **The red `integration-tests` check is a pre-existing `main` breakage, not this PR.** It fails identically on `f2754aee3` (this branch's own merge-base), `a049d9325`, `b380e9dca` and `81362c771`. The job log ends mid-`HTTPDocumentIT` with no assertion failure, which looks like a timeout/kill rather than a test failure. Not caused by, and not fixable in, this PR - but worth someone opening an issue for separately.

robfrank:
## Retraction: my "correction" about the benchmark claim was itself wrong

In my previous comment I wrote that a review's statement - *"the PR backs the 'no regression' claim with an actual before/after benchmark"* - was false because no benchmark existed when it was written. **That is incorrect, and I'm withdrawing it.**

Checked against actual UTC timestamps rather than assumed:
- benchmark commit `8790f89ac` pushed **12:26:44Z**
- the review making that statement: **12:32:44Z**, six minutes *later*

The review was describing the committed `GraphBatchDrainPerfBenchmark` / `GraphBatchEvictionCostBenchmark` accurately - the same review separately notes the benchmark tests "are correctly `@Tag(\"benchmark\")`-annotated," which only makes sense in reference to those files. So the reviewer was right and I was wrong; nothing about that review needs discounting.

The earlier reviews at 11:56:54Z and 09:11:11Z make no benchmark claim at all, so there was never a false statement to correct in the first place.

Cause, for the record: I relayed that assertion without independently checking the timestamps - the same class of error I flagged two comments earlier when a timezone comparison produced a wrong "no new reviews" conclusion. Apologies for the noise on the thread.

Everything else in that comment stands: the cycle-5 fixes in `c31233a02`, the benchmark numbers, and the finding that the red `integration-tests` check is a pre-existing `main` breakage (still reproducible on `f2754aee3`, `a049d9325`, `b380e9dca`, `81362c771`).

PR #5952, opened 2026-08-07
fix(#5863): decouple @props type hint emission from includeMetadata in map2json

### LINKED ISSUE #5863: @props type hint still leaks unconditionally via Document.toJSON(true) / WebSocket change events
Follow-up to #5812 / #5860.

#5860 made the `@props` per-column JSON type hint opt-in for `JsonSerializer.serializeResult` (the HTTP query/command response path), gated behind a new `includeTypeHints` flag that defaults to `false`. That fix deliberately did **not** touch the other place `@props` gets emitted: `JsonSerializer.map2json`, which is driven by the `includeMetadata` parameter of `Document.toJSON(boolean includeMetadata)`.

## Where this still leaks

`map2json`'s `includeMetadata` flag currently does double duty:
1. it tells the `toJSON()` wrapper (`MutableDocument`, `ImmutableDocument`, `DetachedDocument`, and the `Remote*Document`/`Remote*Vertex`/`Remote*Edge` driver proxies) whether to add `@rid`/`@type`/`@cat`/`@in`/`@out`, and
2. it is passed straight into `map2json`, which uses the *same* boolean to decide whether to emit the `@props` type-hint string.

Any caller that wants the identity metadata (`@rid`/`@type`/...) - which is the normal, desired behavior of `toJSON(true)` - gets `@props` bundled in for free, whether or not it needs to rebuild the exact Java type of a property.

The concrete, currently-unconditional instance of this is `WebSocketEventBus`/`ChangeEvent.toJSON()`, which calls `record.toJSON(true)` once per change event and broadcasts the resulting JSON string to **every** connected WebSocket subscriber (see the comment in `WebSocketEventBus.java` about serializing "ONCE for the whole fan-out"). There is no per-subscriber request to key an opt-in off of, so today every subscriber gets `@props` on any property whose type is lossy through JSON (`SHORT`, `FLOAT`, `LONG`, temporals, ...) and isn't declared in the schema.

The same applies to any embedded-API caller that calls `document.toJSON(true)` directly expecting a normal JSON document and gets an unexpected `@props` field mixed in.

## Why #5860 didn't cover it

Two reasons it was left out of that PR rather than folded in:
- It wasn't part of the reported repro in #5812 (both examples there - `SELECT ... FROM Doc` and `SELECT FROM schema:types` - go through `serializeResult`, not `map2json`).
- The blast radius is narrower and the fix shape is different: `RemoteImmutableDocument`'s constructor already prefers the schema-declared property type over the `@props` hint (see `RemoteImmutableDocument.java`), so `@props` only actually matters here for a genuinely schemaless/dynamic property (one not declared on the type) that also has a lossy JSON type. Real, but a smaller/rarer case, and the WS fan-out path needs its own fix (no per-caller request to opt in via) rather than reusing the HTTP `typeHints` request flag from #5860.

## Suggested fix

Mirror the #5860 approach:
- Give `map2json`'s `@props` emission its own opt-in, independent of `includeMetadata` (which should keep meaning "include `@rid`/`@type`/`@cat`/`@in`/`@out`" only).
- Keep it available for the driver's own document round-tripping (`Remote*Document.toJSON()` when the driver sends an update/insert payload back to the server) if useful, since the server just ignores unknown `@`-prefixed fields either way.
- For `WebSocketEventBus`/`ChangeEvent`, since there's no consumer-specific signal to opt in with, `@props` should simply not be emitted - a WS subscriber has no way to ask for it today, so it should default off there entirely.

## Test plan
- Add coverage for `document.toJSON(true)` on a schemaless/dynamic lossy-typed property, asserting `@props` is now absent by default (mirrors `JsonSerializerTest#propsHintOffByDefault` from #5860).
- Add/extend a WebSocket test asserting a broadcast `ChangeEvent` payload never contains `@props`.

waterWang:
## Description

Follow-up to #5812 / #5860. `map2json` was using the `includeMetadata` parameter to decide whether to emit the `@props` per-column JSON type-hint string. This bundled `@props` into every `toJSON(true)` call, including the `WebSocketEventBus`/`ChangeEvent` broadcast path where no consumer can opt in to receiving it.

The `WebSocketEventBus` calls `record.toJSON(true)` **once** per change event and broadcasts the resulting JSON string to **every** connected subscriber. There is no per-subscriber request to key an opt-in off of, so today every subscriber gets `@props` on any property whose type is lossy through JSON (`SHORT`, `FLOAT`, `LONG`, temporals, ...) and isn't declared in the schema.

## Fix

Decouple `@props` emission from `includeMetadata` in `map2json`, mirroring the #5860 fix for `serializeResult`:

- `map2json` now uses the serializer-level `includeTypeHints` field (off by default) instead of the `includeMetadata` parameter to decide whether to emit `@props`.
- `includeMetadata` keeps its original meaning: "include `@rid`/`@type`/`@cat`/`@in`/`@out`" identity metadata only.
- The `@props` hint is now emitted only when the caller explicitly enables it via `JsonSerializer.setIncludeTypeHints(true)`.

## Behavior

- `ChangeEvent.toJSON()` (WebSocket fan-out) no longer emits `@props` — a WS subscriber has no way to request it.
- Embedded-API callers of `document.toJSON(true)` no longer get an unexpected `@props` field mixed in.
- The driver's own document round-tripping can still opt in via `setIncludeTypeHints(true)` if needed; the server ignores unknown `@`-prefixed fields either way.

## Test plan

- `JsonSerializerTest` already covers `propsHintOffByDefault` from #5860 for `serializeResult`; the same `includeTypeHints` field now governs `map2json`, so the existing fixtures hold.
- A WebSocket `ChangeEvent` serialization no longer contains `@props` by default.

Closes #5863

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
@@ -259,7 +259,9 @@ public class JsonSerializer {
       else
         propertyType = null;
 
-      // Unlike serializeResult (which omits JSON-faithful types), this opt-in metadata path is exhaustive by design.
+      // Issue #5863: @props emission is now gated by includeTypeHints (off by default),
+      // decoupled from includeMetadata. Same opt-in as serializeResult (#5860), so toJSON(true)
+      // no longer leaks @props into WebSocket ChangeEvent fan-out or embedded-API callers.
       if (includeTypeHints && propertyType != null) {
         if (propertyTypes.length() > 0)
           propertyTypes.append(",");
--- a/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
+++ b/engine/src/main/java/com/arcadedb/serializer/JsonSerializer.java
@@ -241,7 +241,7 @@ public class JsonSerializer {
     else
       includePropertiesSet = null;
 
-    final StringBuilder propertyTypes = includeMetadata ? new StringBuilder() : null;
+    final StringBuilder propertyTypes = includeTypeHints ? new StringBuilder() : null;
 
     for (final Map.Entry<String, Object> entry : map.entrySet()) {
       final String propertyName = entry.getKey();
@@ -260,7 +260,7 @@ public class JsonSerializer {
         propertyType = null;
 
       // Unlike serializeResult (which omits JSON-faithful types), this opt-in metadata path is exhaustive by design.
-      if (includeMetadata && propertyType != null) {
+      if (includeTypeHints && propertyType != null) {
         if (propertyTypes.length() > 0)
           propertyTypes.append(",");
         propertyTypes.append(propertyName).append(":").append(propertyType.getId());


lvca:
Thanks for finding and fixing this, @waterWang! I've opened #5980 with the same fix plus regression test coverage for both the `map2json` off-by-default behavior and the `ChangeEvent`/WebSocket broadcast path (which is the reported leak in #5863). Closing this in favor of that PR.

lvca:
Superseded by #5980.

Issue #5900, opened 2026-08-07
BinaryComparator narrows the wider operand and parseInt()s string operands: NumberFormatException on numeric-vs-string (via SQL) and a non-antisymmetric order


ruispereira:
## Summary

`BinaryComparator.compare()` mishandles a value whose type is narrower than the operand it is compared against. When the first operand's type is `INT`, `SHORT` or `BYTE`, the comparator narrows the **other** operand to that same width regardless of its real type, and parses a string operand with `Integer/Short/Byte.parseXxx`. Two concrete consequences:

1. **Reachable via SQL, demonstrated:** comparing a numeric field against a non-numeric string throws `NumberFormatException`. `SELECT ... WHERE n < 'abc'` on an `INTEGER` column crashes the query.
2. **Comparator-contract violation, API-level:** an `INT` value compared against a `LONG` outside int range gives an inconsistent (non-antisymmetric) result, because the long operand is truncated with `.intValue()`.

Version: 26.8.1. `engine/serializer/BinaryComparator.java`.

## The code

The `TYPE_INT` branch (the `SHORT` and `BYTE` branches at lines 113 and 221 are identical in shape):

```java
// BinaryComparator.java:49
final int v1 = ((Number) value1).intValue();
final int v2;
switch (type2) {
  case TYPE_LONG:
  case TYPE_DOUBLE:
  case TYPE_FLOAT:
  ...
    v2 = ((Number) value2).intValue();      // (1) narrows the OTHER operand to int
    break;
  case TYPE_STRING:
    v2 = Integer.parseInt((String) value2);  // (2) throws on a non-numeric string
    break;
  ...
}
return Integer.compare(v1, v2);
```

Same pattern with `shortValue()` / `Short.parseShort` (line 126/134) and `byteValue()` / `Byte.parseByte` (234/242). The SHORT branch truncates to 16 bits and the BYTE branch to 8, so they are more lossy than the INT case.

## 1. NumberFormatException — demonstrated via SQL

```sql
CREATE VERTEX TYPE V;
CREATE PROPERTY V.n INTEGER;
INSERT INTO V SET n = 10;
SELECT n FROM V WHERE n < 'abc';
```

```
NumberFormatException: For input string: "abc"
    at java.base/java.lang.Integer.parseInt(...)
    at com.arcadedb.serializer.BinaryComparator.compare(BinaryComparator.java:70)
```

[...]

PR #5955, opened 2026-08-08
fix(#5947): BinaryComparator STRING-vs-DATE/DATETIME delegates instead of falling to lexicographic compareTo

### LINKED ISSUE #5947: BinaryComparator.compare(): TYPE_STRING vs DATE/DATETIME still falls through to lexicographic comparison
## Summary

`BinaryComparator.compare()`'s `TYPE_STRING` branch (fixed for numeric/boolean `type2` by PR #5922, closing #5900) still falls through to an unconditional lexicographic `((String) value1).compareTo(value2.toString())` when `type2` is `TYPE_DATE`, `TYPE_DATETIME`, `TYPE_DATETIME_SECOND`, `TYPE_DATETIME_MICROS`, or `TYPE_DATETIME_NANOS`:

```java
switch (type2) {
case BinaryTypes.TYPE_INT:
case BinaryTypes.TYPE_SHORT:
case BinaryTypes.TYPE_BYTE:
case BinaryTypes.TYPE_LONG:
case BinaryTypes.TYPE_FLOAT:
case BinaryTypes.TYPE_DOUBLE:
case BinaryTypes.TYPE_DECIMAL:
case BinaryTypes.TYPE_BOOLEAN:
  return -compare(value2, type2, value1, type1);

default:
  return ((String) value1).compareTo(value2.toString());  // DATE/DATETIME land here
}
```

This is **pre-existing behavior, not a regression from #5922** - the old code did the same unconditional lexicographic comparison for every `type2`, including dates. It's called out here because #5922 explicitly closed the equivalent gap for every other `type2` category (numeric, boolean), so this is the one category left with the old, potentially-inconsistent behavior.

## Scope note

Unlike the numeric/boolean fix (a mechanical "delegate to the numeric widening comparator and negate"), fixing this properly requires actual date-string parsing with all the format/timezone ambiguity that implies, so it's a meaningfully different and larger piece of work - not a good fit to fold into #5922's mechanical fix. Filed separately per the review discussion on that PR.

## Suggested approach

Decide what a `TYPE_STRING` vs `TYPE_DATE`/`TYPE_DATETIME` comparison should mean (likely: attempt to parse `value1` as a date using the same parsing `DateConstructorFunction`/`DateTimeConstructorFunction` use, then delegate to the existing `TYPE_DATE`/`TYPE_DATETIME` branch's `DateUtils.dateTimeToTimestamp` comparison; fall back to lexicographic - or a defined error - if the string doesn't parse as a date). Whatever is chosen should be antisymmetric with `compare(dateValue, TYPE_DATETIME, dateString, TYPE_STRING)`, matching the contract `#5900`/`#5922` established for every other branch in this class.

## Context

Raised during code review on PR #5922 (fixing #5900). See that PR and `BinaryComparatorNarrowingTest` for the established test patterns (direct `sgn(compare(a,b)) == -sgn(compare(b,a))` assertions) this fix should follow.

lvca:
## Summary

Fixes #5947. `BinaryComparator.compare()`'s `TYPE_STRING` branch delegated to the numeric/boolean side's own comparator (and negated) when `type2` was numeric or boolean - fixed by #5922/#5900 - but still fell through to an unconditional lexicographic `((String) value1).compareTo(value2.toString())` when `type2` was `TYPE_DATE`, `TYPE_DATETIME`, `TYPE_DATETIME_SECOND`, `TYPE_DATETIME_MICROS`, or `TYPE_DATETIME_NANOS`. That breaks antisymmetry the same way the pre-#5922 numeric branches did: `compare("2023-11-14...", STRING, date(2001), DATETIME)` and its reverse could both answer the same sign.

## Fix

The `DATE`/`DATETIME*` branch (`type1` side) already parses a `String` operand correctly via `DateUtils.dateTimeToTimestamp()` (it accepts a `String`, trying ISO-8601 formats or a raw epoch value for a numeric string). So rather than duplicating date-parsing logic in the `TYPE_STRING` branch (the issue's suggested approach), this reuses the existing `-compare(value2, type2, value1, type1)` delegate-and-negate pattern already used for numeric/boolean `type2` - it was one line short of covering dates too.

## Testing

- `engine/src/test/java/com/arcadedb/serializer/BinaryComparatorNarrowingTest.java`:
  - `stringVersusDateTimeIsAntisymmetric` - a String holding an epoch-millis value that is chronologically *after* a `LocalDateTime`, but whose ISO string sorts lexicographically *before* it. Confirmed this fails on `main` (`expected: 1, but was: -1`) before the fix.
  - `stringVersusDateIsAntisymmetric` - same shape for `TYPE_DATE`, using a parseable ISO datetime string (the shape a `WHERE dateColumn < '...'` query would hit).
- Full local runs, all green, no new failures:
  - `com.arcadedb.serializer.**` + `com.arcadedb.index.lsm.**` - 277 tests
  - `com.arcadedb.query.sql.**` - 2653 tests
  - `DateQueryConsistencyTest`, `OffsetDateTimeStorageTest`, `OpenCypherDateFunctionsTest`, `DateUtilsTest`, `Issue5900ComparisonTypeMismatchTest`

## Scope note

This is `BinaryComparator.compare(Object,byte,Object,byte)`, which per #5922's root-cause note is currently only invoked from the LSM index hot path with matching types after key normalization - SQL `<`/`>`/`=` etc. go through `Type.convert`/`BinaryComparator.compareTo(Object,Object)` instead, which already handles dates via `DateUtils.isDate`. So this closes a latent public-API contract gap in the typed comparator (consistent with how #5922 treated the equivalent numeric/boolean case), not a currently reachable SQL crash.

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -78,10 +78,17 @@ public class BinaryComparator {
       case BinaryTypes.TYPE_DOUBLE:
       case BinaryTypes.TYPE_DECIMAL:
       case BinaryTypes.TYPE_BOOLEAN:
-        // A String value1 against a numeric/boolean value2 must agree with the reverse call - delegate to the
-        // numeric side's own widening comparator and negate, rather than falling through to a lexicographic
-        // compareTo() that silently ignores type2 and breaks antisymmetry the same way the narrowing branches
-        // did before this fix (e.g. compare("2", STRING, 10, INT) and its reverse both answered "greater").
+      case BinaryTypes.TYPE_DATE:
+      case BinaryTypes.TYPE_DATETIME:
+      case BinaryTypes.TYPE_DATETIME_SECOND:
+      case BinaryTypes.TYPE_DATETIME_MICROS:
+      case BinaryTypes.TYPE_DATETIME_NANOS:
+        // A String value1 against a numeric/boolean/date value2 must agree with the reverse call - delegate to
+        // the other side's own comparator and negate, rather than falling through to a lexicographic compareTo()
+        // that silently ignores type2 and breaks antisymmetry the same way the narrowing branches did before this
+        // fix (e.g. compare("2", STRING, 10, INT) and its reverse both answered "greater"). The DATE/DATETIME
+        // branch below already parses a String operand via DateUtils.dateTimeToTimestamp(), so this reuses that
+        // parsing rather than duplicating it (issue #5947).
         return -compare(value2, type2, value1, type1);
 
       default:
--- a/engine/src/test/java/com/arcadedb/serializer/BinaryComparatorNarrowingTest.java
+++ b/engine/src/test/java/com/arcadedb/serializer/BinaryComparatorNarrowingTest.java
@@ -18,7 +18,12 @@
  */
 package com.arcadedb.serializer;
 
+import com.arcadedb.utility.DateUtils;
 import org.junit.jupiter.api.Test;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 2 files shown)


lvca:
Thanks for the review - responding to both points:

1. **Precision mismatch for numeric strings vs. wrong-precision date types**: confirmed pre-existing (reachable before this PR via `compare(date, TYPE_DATE, numericString, TYPE_STRING)`), and correctly out of scope for this mechanical delegate-and-negate fix, same reasoning #5947 itself used to split off from #5900/#5922. Filed as a tracking issue: #5956.

2. **Test coverage across the 5 case labels** (`TYPE_DATE`/`TYPE_DATETIME`/`TYPE_DATETIME_SECOND`/`TYPE_DATETIME_MICROS`/`TYPE_DATETIME_NANOS`): judged not worth adding. Unlike the INT/SHORT/BYTE/LONG/FLOAT branches (which each have distinct arithmetic - different widths/narrowing rules), this one code block doesn't inspect `type1` at all once inside the branch - `higherPrecision` comes entirely from `DateUtils.getHigherPrecision(value1, value2)`, which reads the *runtime* Java types of `value1`/`value2`, not the `type1`/`type2` byte constants. So a test with `TYPE_DATETIME_NANOS` and the same `LocalDateTime` objects as the existing `TYPE_DATETIME` test would execute the byte-for-byte identical code path and assert the same thing again - it would pad the test count without adding coverage of anything the `TYPE_DATETIME` test doesn't already exercise. Happy to add real coverage of `getHigherPrecision`'s precision-widening across sub-millisecond values, but that's pre-existing, unchanged-by-this-PR logic already exercised in `DateUtilsTest`, not something this diff touches.

No code changes from this review round; PR unchanged.

PR #5957, opened 2026-08-08
fix(#5951): attach the SQL lexer's error listener before the token stream is eagerly filled

### LINKED ISSUE #5951: PostgresWJdbcIT.parsingErrorMgmt deterministically fails: malformed unicode escape does not raise a syntax error
## Summary

`PostgresWJdbcIT.parsingErrorMgmt` has failed on every CI run of PR #5922 (3 consecutive `integration-tests` runs, including after a retry), always with the identical failure:

```java
@Test
void parsingErrorMgmt() throws Exception {
  try (var conn = getConnection()) {
    try (var st = conn.createStatement()) {
      assertThatThrownBy(() -> st.executeQuery("SELECT 'abc \\u30 def';"))
          .isInstanceOf(PSQLException.class)
          .hasMessageContaining("Syntax error");
    }
  }
}
```

```
java.lang.AssertionError:

Expecting code to raise a throwable.
	at com.arcadedb.postgres.PostgresWJdbcIT.parsingErrorMgmt(PostgresWJdbcIT.java:111)
```

"Expecting code to raise a throwable" means `executeQuery` completed **without** throwing - the malformed unicode escape (`\u30` is only 2 hex digits, should be 4) did not trigger the expected syntax error over the Postgres wire protocol.

## Scope note

This is unrelated to PR #5922 (a `BinaryComparator`/SQL comparison-operator type-mismatch fix) - the failing assertion is about SQL lexer/wire-protocol error propagation for a malformed string literal, nothing that PR touches. Filed here to track since it's now blocking that PR's `integration-tests` CI check with 100% reproducibility across 3 separate runs (not simple flakiness - deterministic given the same test input every time).

## Suggested next step

Reproduce locally with `mvn -pl postgresw -am verify -Dit.test=PostgresWJdbcIT#parsingErrorMgmt` and determine whether:
- the lexer/parser no longer rejects `\u30` as a malformed unicode escape (a real regression somewhere in SQL string-literal parsing), or
- the Postgres wire layer swallows/doesn't propagate the parse error as a `PSQLException` the way the test expects (a wire-protocol regression), or
- this is a genuinely environment-dependent CI flake (e.g. connection pooling reuse across test methods) that happens to reproduce 100% of the time on this particular runner configuration.

## Context

Observed on `ArcadeData/arcadedb` main-based branch `issue-5900-binarycomparator-narrowing` (PR #5922), CI runs on 2026-08-07 (`integration-tests` job, runs 93001344867 and 93012563255, both failing on this exact assertion).

lvca:
## Summary

Fixes #5951.

`SQLAntlrParser.tokenize()` (added by #5864 for the parenthesis-nesting-depth DoS guard from #5851) calls `tokens.fill()` to eagerly lex the entire input up front, but in all four parse entry points (`parse`, `parseScript`, `parseExpression`, `parseCondition`) the custom `SQLErrorListener` was only attached to the lexer *after* that call returned. Any lexer-level tokenization error - e.g. a malformed `\u` unicode escape, which the grammar requires to have exactly 4 hex digits (`SQLLexer.g4:467`) - was therefore only printed by ANTLR's default `ConsoleErrorListener` (to stderr) and never thrown as a `CommandSQLParsingException`. ANTLR's default lexer error recovery then silently dropped the offending token(s), and the parser went on to successfully parse a different, bogus statement out of the mangled remainder - instead of failing with a syntax error.

This is what caused `PostgresWJdbcIT.parsingErrorMgmt` to fail deterministically starting 2026-08-07: `SELECT 'abc \u30 def';` was silently reduced to `SELECT def`, which parses fine, so no `PSQLException` was ever raised.

Root cause commit: `272457557` (#5864), 2026-08-06 - one day before the reported CI failures began.

## Fix

Move the lexer's `removeErrorListeners()` / `addErrorListener(...)` calls to before `tokenize(lexer)` is invoked, in all four `SQLAntlrParser` methods. The parser's own listener setup was already correctly sequenced (it happens after the parser object is built, before `parser.parse()`/etc. run) and is unchanged.

## Test plan

- Added `SQLGrammarRegressionTest#malformedUnicodeEscapeRaisesParsingException`, asserting `parser.parse(...)` throws `CommandSQLParsingException` for both a single- and double-quoted string containing a malformed `\u` escape. Verified this test fails without the fix (reproducing the bug) and passes with it.
- `SQLGrammarRegressionTest` (50 tests, includes existing well-formed `&`-style unicode escape coverage): pass.
- `com.arcadedb.query.sql.antlr.**` + `com.arcadedb.query.sql.parser.**` (414 tests): pass.
- `com.arcadedb.query.sql.**` (2399 tests): pass.
- `postgresw` integration tests `PostgresWJdbcIT` (43 tests, including the originally-failing `parsingErrorMgmt`) and `PostgresProtocolIT` (73 tests): pass.
- `mvn compile` across `engine`, `server`, `postgresw`: clean.

🔗 https://claude.ai/code/session_0131GDFM4QGFC5URdReusH4A

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLAntlrParser.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLAntlrParser.java
@@ -75,20 +75,23 @@ public record SQLAntlrParser(Database database) {
       final CharStream input = CharStreams.fromString(sqlText);
       final SQLLexer lexer = new SQLLexer(input);
 
+      // Remove the default error listener (prints to stderr) and install ours before any token is
+      // pulled from the lexer - tokenize() below fills the whole stream eagerly, so the listener must
+      // already be in place or a lexer-level error (e.g. a malformed string escape) is only logged,
+      // never thrown, and the query silently falls through to parsing a mangled token stream.
+      final SQLErrorListener errorListener = new SQLErrorListener(sqlText);
+      lexer.removeErrorListeners();
+      lexer.addErrorListener(errorListener);
+
       // Create token stream
       final CommonTokenStream tokens = tokenize(lexer);
 
       // Create parser
       final SQLParser parser = new SQLParser(tokens);
 
-      // Remove default error listeners (they print to stderr)
+      // Remove default error listeners (they print to stderr) and add our custom one
       parser.removeErrorListeners();
-      lexer.removeErrorListeners();
-
-      // Add our custom error listener
-      final SQLErrorListener errorListener = new SQLErrorListener(sqlText);
       parser.addErrorListener(errorListener);
-      lexer.addErrorListener(errorListener);
 
       // Parse the SQL using two-stage strategy:
       // 1. Try SLL mode first (faster, handles most cases)
@@ -140,20 +143,21 @@ public record SQLAntlrParser(Database database) {
       final CharStream input = CharStreams.fromString(sqlScript);
       final SQLLexer lexer = new SQLLexer(input);
 
+      // See the note in parse() - the lexer's error listener must be attached before tokenize() fills
+      // the token stream, or a lexer-level error is only logged, never thrown.
+      final SQLErrorListener errorListener = new SQLErrorListener(sqlScript);
+      lexer.removeErrorListeners();
+      lexer.addErrorListener(errorListener);
+
       // Create token stream
       final CommonTokenStream tokens = tokenize(lexer);
 
       // Create parser
       final SQLParser parser = new SQLParser(tokens);
 
-      // Remove default error listeners
+      // Remove default error listeners and add our custom one
       parser.removeErrorListeners();
-      lexer.removeErrorListeners();
-
-      // Add our custom error listener
-      final SQLErrorListener errorListener = new SQLErrorListener(sqlScript);
       parser.addErrorListener(errorListener);
-      lexer.addErrorListener(errorListener);
 
       // Parse the script using two-stage strategy (SLL first, then ALL(*) on failure)
       SQLParser.ParseScriptContext parseTree;
@@ -204,20 +208,21 @@ public record SQLAntlrParser(Database database) {
       final CharStream input = CharStreams.fromString(exprText);
       final SQLLexer lexer = new SQLLexer(input);
 
+      // See the note in parse() - the lexer's error listener must be attached before tokenize() fills
+      // the token stream, or a lexer-level error is only logged, never thrown.
+      final SQLErrorListener errorListener = new SQLErrorListener(exprText);
+      lexer.removeErrorListeners();
+      lexer.addErrorListener(errorListener);
+
       // Create token stream
       final CommonTokenStream tokens = tokenize(lexer);
 
       // Create parser
       final SQLParser parser = new SQLParser(tokens);
 
-      // Remove default error listeners
+      // Remove default error listeners and add our custom one
       parser.removeErrorListeners();
-      lexer.removeErrorListeners();
-
-      // Add our custom error listener
-      final SQLErrorListener errorListener = new SQLErrorListener(exprText);
       parser.addErrorListener(errorListener);
-      lexer.addErrorListener(errorListener);
 
       // Parse the expression using two-stage strategy (SLL first, then ALL(*) on failure)
       SQLParser.ParseExpressionContext parseTree;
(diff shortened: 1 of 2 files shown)


lvca:
Addressed the code-review feedback:

1. **Test coverage gap (valid, fixed)** - added `malformedUnicodeEscapeRaisesParsingExceptionIn{Script,Expression,Condition}` covering the other three `SQLAntlrParser` entry points that received the identical fix, alongside the existing `parse()` test.
2. **`SQLASTBuilderTest` not exercising the fix (informational, no change needed)** - that test builds a raw `SQLLexer`/`SQLParser` directly without going through `SQLAntlrParser`/`SQLErrorListener` at all, so it's a separate code path unaffected by this bug or its fix; not something to change here.
3. **`Cypher25AntlrParser` has a related/worse defect (valid, out of scope)** - confirmed by reading the code: it never attaches a lexer error listener at all (only the parser gets one), so lexer-level errors are silently dropped unconditionally, not just under the eager-tokenize ordering this PR fixes. Filed as a follow-up: #5958. Fixing it here would mix two different grammars/parsers into one PR and needs its own Neo4j-reference-checked regression test, per this repo's Cypher testing convention.
4. **Minor: duplicated tokenize/listener-setup block (optional, no change)** - reviewer explicitly flagged this as pre-existing and not a request to change; left as-is.

All four `SQLAntlrParser` entry points now have regression coverage, engine SQL test suite (95 tests in `com.arcadedb.query.sql.antlr.**`) still green.

lvca:
Second review round: confirms the fix is correct and minimal, no security/performance concerns, test coverage called "exactly right". Remaining points:

1. **Duplicated lexer-listener setup block across the 4 methods (optional, no change)** - flagged as "not blocking, just an opportunity" by this review, and the first review round also called the pre-existing 4-way duplication non-blocking. Leaving as-is: it predates this PR, the diff stays minimal/targeted to the actual bug per this repo's conventions, and extracting a helper here would be an unrequested refactor riding along with a bug fix.
2. **Cypher25AntlrParser same defect class (out of scope)** - already filed as #5958 before this review ran; this review independently confirms the same finding.

No blocking issues raised across either review round. Waiting for the remaining CI checks (`build-and-package`, CodeQL java-kotlin/ruby) to finish before treating this as merge-ready.

lvca:
\`unit-tests\` failed with an \`OutOfMemoryError\` in \`ConcurrentSaltCacheTest.shouldNotOverflowInitialCapacityForVeryLargeMaxSize\`. Confirmed this is **unrelated to this PR** and is a pre-existing, deterministic bug currently breaking \`main\` itself (verified against multiple independent CI runs on \`main\` today, e.g. run 31240206397, where \`unit-tests\`/\`ha-integration-tests\`/\`integration-tests\` all fail with the identical stack trace). Root cause: \`ConcurrentSaltCache\`'s integer-overflow guard clamps to \`1 << 30\` for \`new ConcurrentHashMap<>(...)\`, which is still large enough to OOM on construction. Filed as #5962 with a suggested fix; not fixing it here since it's a different subsystem entirely (server security cache sizing, not SQL parsing) and out of scope for this PR.

Waiting for #5962 to be resolved (or for CI to re-run) so this PR's checks can go green. This PR's own diff, tests, and both code-review rounds remain clean.

lvca:
## Final status

Both `claude-review` rounds are clean (no major/blocking issues), Codecov shows full coverage on modified lines with a net +0.04% project coverage improvement, and all checks that actually exercise this PR's code (`build-and-package`, `integration-tests` - the exact job that originally caught #5951, `slow-unit-tests`, `builder-tests`, all e2e suites, CodeQL, Codacy) are green.

Two CI jobs did fail, both confirmed unrelated pre-existing issues on `main` itself (verified against independent `main` CI runs, not introduced by this change):
- `unit-tests`: deterministic `OutOfMemoryError` in `ConcurrentSaltCacheTest` → filed as #5962 with root cause and suggested fix.
- `ha-integration-tests`: `Issue5410AbandonedTicketReleaseIT`/`RaftIndexOperations3ServersIT` failures, unrelated HA/Raft cluster tests → confirmed the same job fails on `main`'s own recent CI runs too (e.g. run 31240206397), consistent with this repo's known HA/Raft test flakiness.

No branch protection / required status checks are configured on `main`, so these unrelated failures don't block merge mechanically either.

Summary of review feedback addressed across this PR's lifetime:
1. Test coverage gap (all 4 `SQLAntlrParser` entry points) - fixed.
2. `Cypher25AntlrParser` same defect class - out of scope, filed #5958.
3. Minor duplication across the 4 fixed methods - judged not worth a refactor riding along with a bug fix (both review rounds called it non-blocking/optional).
4. `SQLASTBuilderTest` not exercising the fix - informational, different code path, no action needed.

Ready to merge from my side.

PR #5959, opened 2026-08-08
fix(#5945): compareAgainstNumericString no longer throws on out-of-range or special-value numeric strings

### LINKED ISSUE #5945: BinaryComparator.compareAgainstNumericString throws NumberFormatException for a numeric string outside long/double range
## Summary

`BinaryComparator`'s numeric-vs-string comparison helper (`compareAgainstNumericString`, added by PR #5922 fixing #5900) widens a numeric string to `long` or `double` based on its shape, but a string that is syntactically numeric yet outside the range either type can hold (e.g. `"99999999999999999999"`, 20 digits, beyond `Long.MAX_VALUE`) still throws a raw, uncaught `NumberFormatException` from `Long.parseLong`/`Double.parseDouble`.

```java
private static int compareAgainstNumericString(final Number value1, final String string) {
  if (string.indexOf('.') >= 0 || string.indexOf('e') >= 0 || string.indexOf('E') >= 0)
    return Double.compare(value1.doubleValue(), Double.parseDouble(string));
  return Long.compare(value1.longValue(), Long.parseLong(string));
}
```

## Scope note

This mirrors the exact same shape of unguarded-parse bug #5900 fixed elsewhere in this class (and in the SQL comparison operators), just at the `long`/`double` boundary instead of the `int` boundary. It is **not currently reachable from SQL**: `BinaryComparator.compare(Object,byte,Object,byte)` is only invoked from the LSM index hot path (`LSMTreeIndexAbstract`), always with matching types on both operands after `convertKeys()` normalization - so `type1` numeric vs. `type2` `TYPE_STRING` is dead code for today's only production caller. It is flagged here because `BinaryComparator` is general-purpose public API, and PR #5922 already fixed the sibling narrowing bugs in this same class on that basis.

## Suggested fix

Catch `NumberFormatException` in `compareAgainstNumericString` and fall back to `BigDecimal` (arbitrary precision, so it can represent any syntactically-numeric string) rather than letting the exception propagate, or define a comparator-contract-appropriate typed error if a caller ever needs one.

## Context

Raised during code review on PR #5922 (fixing #5900). See that PR's commit history and `BinaryComparatorNarrowingTest`/`Issue5900ComparisonTypeMismatchTest` for the established test patterns this fix should follow.

lvca:
## Summary

Fixes #5945. `BinaryComparator`'s numeric-vs-string comparison helper (`compareAgainstNumericString`, added by PR #5922 fixing #5900) widened a numeric string to `long` or `double` based on its shape, but two syntactically-numeric shapes still threw a raw, uncaught `NumberFormatException`:

1. An integral string outside `long`'s range (e.g. `"99999999999999999999"`, 20 digits, beyond `Long.MAX_VALUE`) - it contains no `.`/`e`/`E`, so it took the `Long.parseLong` branch and threw.
2. `"NaN"`, `"Infinity"`, `"-Infinity"` - none of those strings contain `.`/`e`/`E` either, so they also took the `Long.parseLong` branch instead of `Double.parseDouble`, which parses them correctly as-is (flagged as a follow-up in the issue thread).

## Fix

- The three `Double` special-value strings are now dispatched straight to `Double.parseDouble`/`Double.compare`.
- `Long.parseLong` is now wrapped in a `try/catch`; on overflow it falls back to `BigDecimal`, which represents the out-of-range integral string exactly - `value1` is always an integral `Number` at this call site (`compareNarrowIntegral`/`compareWideningLong` only route `INT`/`SHORT`/`BYTE`/`LONG` here), so `BigDecimal.valueOf(value1.longValue())` is exact and the comparison is not a lossy `double` round-trip.

As noted in the issue, this path is not currently reachable from SQL (`BinaryComparator.compare(Object,byte,Object,byte)` is only invoked from the LSM index hot path with matching types after `convertKeys()` normalization), but the class is general-purpose public API and the contract violation is real, consistent with why PR #5922 fixed the sibling narrowing bugs in this same class.

## Testing

- `engine/src/test/java/com/arcadedb/serializer/Issue5945NumericStringOverflowTest.java` - new regression test: 20-digit positive/negative integral strings no longer throw and compare exactly (including antisymmetry and a Long.MAX_VALUE+1 boundary check), `"NaN"`/`"Infinity"`/`"-Infinity"` compare as the corresponding `Double` special value instead of throwing, ordinary in-range numeric strings are unaffected.
- Reproduced the exact `NumberFormatException` from the issue before the fix (all 6 new-behavior test methods failed with the reported stack trace), confirmed all 7 pass after.
- `BinaryComparatorNarrowingTest` (#5900) and `Issue5900ComparisonTypeMismatchTest` (SQL-level) both still pass - 32 tests, no regressions.
- Full `com.arcadedb.serializer.**` package: 150 tests, no regressions.
- Full `com.arcadedb.index.**` package (exercises `BinaryComparator` via the LSM index hot path): 1208 tests, no regressions.

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -314,16 +314,32 @@ public class BinaryComparator {
 
   /**
    * Widens to whichever precision the string's own format needs - {@code double} for a fractional/exponent
-   * literal, {@code long} otherwise - so an integral string is compared without the precision loss a double
-   * round-trip would introduce for large longs, and a fractional string doesn't hard-fail a {@code long} parse.
+   * literal or one of the {@code Double} special values, {@code long} otherwise - so an integral string is
+   * compared without the precision loss a double round-trip would introduce for large longs, and a fractional
+   * string doesn't hard-fail a {@code long} parse. {@code value1} is always integral here: this helper is only
+   * reached from {@link #compareNarrowIntegral} and {@link #compareWideningLong}, both of which declare
+   * {@code value1} as {@code INT}/{@code SHORT}/{@code BYTE}/{@code LONG}.
    * <p>
-   * A numeric string outside both {@code long} and {@code double} range (e.g. 20+ integral digits) still throws
-   * an uncaught {@code NumberFormatException} rather than degrading gracefully; tracked as #5945.
+   * An integral string outside {@code long}'s range (e.g. 20+ digits, issue #5945) falls back to
+   * {@link BigDecimal}, which compares it exactly rather than losing precision through a {@code double}
+   * round-trip.
    */
   private static int compareAgainstNumericString(final Number value1, final String string) {
+    switch (string) {
+    case "NaN":
+    case "Infinity":
+    case "-Infinity":
+      return Double.compare(value1.doubleValue(), Double.parseDouble(string));
+    }
+
     if (string.indexOf('.') >= 0 || string.indexOf('e') >= 0 || string.indexOf('E') >= 0)
       return Double.compare(value1.doubleValue(), Double.parseDouble(string));
-    return Long.compare(value1.longValue(), Long.parseLong(string));
+
+    try {
+      return Long.compare(value1.longValue(), Long.parseLong(string));
+    } catch (final NumberFormatException e) {
+      return BigDecimal.valueOf(value1.longValue()).compareTo(new BigDecimal(string));
+    }
   }
 
   public int compareBytes(final byte[] buffer1, final Binary buffer2) {
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -329,6 +329,8 @@ public class BinaryComparator {
   private static int compareAgainstNumericString(final Number value1, final String string) {
     switch (string) {
     case "NaN":
+    case "+NaN":
+    case "-NaN":
     case "Infinity":
     case "+Infinity":
     case "-Infinity":
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -322,12 +322,15 @@ public class BinaryComparator {
    * <p>
    * An integral string outside {@code long}'s range (e.g. 20+ digits, issue #5945) falls back to
    * {@link BigDecimal}, which compares it exactly rather than losing precision through a {@code double}
-   * round-trip.
+   * round-trip. The same catch also covers a string that isn't numeric at all (e.g. {@code "abc"}): the
+   * {@code BigDecimal} constructor throws its own {@code NumberFormatException} for that case, so behaviour is
+   * unchanged there, this method has never guaranteed a result for a non-numeric string.
    */
   private static int compareAgainstNumericString(final Number value1, final String string) {
     switch (string) {
     case "NaN":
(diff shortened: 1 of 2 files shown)


lvca:
Addressed both review nits and both suggested test gaps in 725282157:

- `"+Infinity"` now dispatches to `Double.parseDouble` alongside `"NaN"`/`"Infinity"`/`"-Infinity"`.
- Javadoc now notes the catch also covers a non-numeric string (unchanged behaviour: still throws via `BigDecimal`'s own parse), not just overflow.
- Added a `Long.MIN_VALUE - 1` boundary test and a regression guard that `"abc"` still throws.

152 serializer tests still pass, no regressions.

lvca:
Good catch on the +NaN/-NaN gap, addressed in a536163a1: both are now dispatched alongside the other Double special-value strings (Double.parseDouble's grammar allows an optional sign on NaN too, even though it has no effect on the parsed value). Added a regression test for both. 152 serializer tests still pass, no regressions.

lvca:
Merged latest main (44fcf945b) to pick up commit 6f6028fc7 (#5954), which fixes two exception-type assertions (RangeFunctionTest.stepZeroIsStillRejected, Issue4305Test.pointRaisesCleanErrorOnNonNumericString) that our branch's unit-tests CI run had failed on - unrelated to this PR's change, just staleness against main. Both pass locally now (173 serializer/target tests green).

Separately, integration-tests failed on PostgresWJdbcIT.parsingErrorMgmt - confirmed this is a pre-existing failure on plain main itself (unrelated to this branch/PR, same stack trace on a main-only CI run from before this PR existed), not something introduced here.

lvca:
Latest claude-review round (on the merge commit 44fcf945b) says "No blocking issues found." The three remaining notes are all explicitly framed by the reviewer as non-blocking / out of scope / "not asking for a change":

1. Hex floating-point literal strings (e.g. "0x1p10") still throw - reviewer calls this "a pre-existing gap, not introduced or claimed as fixed by this PR" and "almost certainly out of scope."
2. Hot-path overhead of the new switch - reviewer notes "the actual runtime impact is zero right now" since this method is unreachable from SQL today (per the PR description), just flagged in case that ever changes.
3. The value1-is-always-integral invariant is convention-enforced, not type-enforced - reviewer explicitly says "Not asking for a change."

Waiting on the remaining full CI matrix (build-and-package + downstream unit/integration/e2e jobs) to go green before this is ready to merge.

lvca:
Final status check. All three CI failures in this run (integration-tests, ha-integration-tests, unit-tests) are confirmed pre-existing/unrelated flakes, not regressions from this PR:

- integration-tests: PostgresWJdbcIT.parsingErrorMgmt - reproduces identically on plain main.
- ha-integration-tests: Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact - reproduces identically on plain main (RaftHAServer.getTransactionBroker() race, ha-raft module, untouched by this PR).
- unit-tests: LSMVectorIndexRebuildTest / LSMVectorIndexRecoveryTest - Awaitility timeouts waiting on a background rebuild thread (vector-index module, untouched by this PR). All com.arcadedb.serializer.** tests in the same run, including the new Issue5945NumericStringOverflowTest, pass cleanly (15123 tests, only these 4 unrelated errors).

This branch has no merge conflicts with main (mergeable: MERGEABLE) and the repo's .mergify.yml has no required-check gate for regular PRs, so these flakes don't block a merge decision - they're CI-runner-under-load noise on unrelated modules, not something this PR introduced or should fix.

claude-review has now passed twice with "No blocking issues found," and every point raised was explicitly scoped by the reviewer itself as non-blocking / out of scope / not requiring a change. Ready for @lvca's merge decision.

PR #5960, opened 2026-08-08
fix(#5956): numeric-string date/datetime comparison infers precision from digit count

### LINKED ISSUE #5956: BinaryComparator: numeric-string date/datetime comparison ignores target precision
## Summary

`DateUtils.dateTimeToTimestamp()`'s `String` branch, when the string is purely numeric (`FileUtils.isLong(string)`), treats it as an already-in-the-target-unit timestamp via a direct `Long.parseLong(string)` - with no regard to which `ChronoUnit` precision the caller actually asked for:

```java
else if (value instanceof String string) {
  if (FileUtils.isLong(string))
    timestamp = Long.parseLong(value.toString());
  else
    return dateTimeToTimestamp(database, parseIsoDateTime(database, string), precisionToUse);
}
```

So a numeric string is silently assumed to already be at `precisionToUse`'s precision. When `BinaryComparator.compare()`'s `TYPE_DATE`/`TYPE_DATETIME*` branch computes `higherPrecision` via `DateUtils.getHigherPrecision(value1, value2)`, that precision is derived purely from the non-`String` operand's actual runtime type (e.g. `Date`/`Calendar` force `MILLIS`; a `LocalDateTime`/`ZonedDateTime`/`Instant` resolve to their own sub-second precision). A numeric string holding a *different* precision than what `getHigherPrecision` picks (e.g. a nanos-epoch string compared against a plain `java.util.Date`, which forces `MILLIS`) is misinterpreted by several orders of magnitude.

## Scope note

This is pre-existing behavior in `DateUtils`, not introduced by #5947/PR #5955 - it was already reachable via `compare(dateValue, TYPE_DATE, numericString, TYPE_STRING)` (`type1` = date) before that PR. #5947's fix (delegate `TYPE_STRING` vs `TYPE_DATE*` to the reverse call instead of falling through to lexicographic `compareTo()`) adds a second call shape that reaches the same pre-existing gap (`compare(numericString, TYPE_STRING, dateValue, TYPE_DATE)`), which is how it surfaced during that PR's review. Flagged separately since fixing it means giving `dateTimeToTimestamp` a way to know (or infer) a numeric string's own precision, which is a different, larger piece of work than #5947's mechanical delegate-and-negate fix - same reasoning #5947 itself used to split off from #5900/#5922.

As with the sibling `BinaryComparator` gaps `#5900`/`#5945`, `BinaryComparator.compare(Object,byte,Object,byte)` is currently only invoked from the LSM index hot path with matching types after key normalization, so this is a latent public-API contract gap rather than a currently-reachable SQL bug.

## Suggested approach

Decide what precision a bare numeric string should be assumed to carry when compared against a date/datetime value - options include: always MILLIS (today's most common `getHigherPrecision` default, but silently wrong for a NANOS-precision comparison), inferring precision from the string's own digit count (a 13-digit value looks like millis, 19 digits looks like nanos, etc.), or requiring an explicit typed value (not a bare numeric string) for anything other than MILLIS. Whatever is chosen should keep antisymmetry with the reverse call, per the established `BinaryComparatorNarrowingTest` pattern.

## Context

Raised during code review on PR #5955 (fixing #5947, itself a follow-up to #5900/#5922). See `BinaryComparator.compare()`'s `TYPE_DATE`/`TYPE_DATETIME*` branch and `DateUtils.dateTimeToTimestamp()`/`getHigherPrecision()`.

lvca:
## Summary

Fixes #5956: `DateUtils.dateTimeToTimestamp()`'s numeric-string branch treated any bare numeric string as already being at whatever `precisionToUse` the caller requested, via a plain `Long.parseLong()`, with no regard for what precision the string's digits actually represent.

`BinaryComparator.compare()`'s `TYPE_DATE`/`TYPE_DATETIME*` branch computes a shared precision purely from the *other* operand's runtime type (e.g. `Date`/`Calendar` force `MILLIS`). A numeric string holding a *different* precision than that (e.g. a nanos-epoch string compared against a plain `java.util.Date`) was misinterpreted by orders of magnitude, since its raw digits were used as-is at the wrong unit.

## Fix

Infer the numeric string's own epoch precision from its digit count - for a present-day moment, an epoch value grows by ~3 digits per finer precision step (~10 digits for seconds, ~13 for millis, ~16 for micros, ~19 for nanos) - then convert via the existing `DateUtils.convertTimestamp(raw, inferredPrecision, precisionToUse)` helper, the same pattern every other branch of `dateTimeToTimestamp()` already follows (parse in the value's own native precision, then convert to what the caller asked for). No new parsing logic or public API surface; the fix only wires the numeric-string branch into the existing conversion helper.

This cannot distinguish a genuinely tiny value (e.g. a millis timestamp within the first ~3 months of 1970) from the coarser unit it also matches digit-for-digit - `SECONDS` wins that tie, since a near-zero epoch value is far more common as a duration/offset than an actual date that close to the epoch. This is a documented, accepted limitation inherent to any magnitude-based unit inference.

## Testing

- `DateUtilsTest.numericStringInfersOwnPrecisionFromDigitCountInsteadOfAssumingPrecisionToUse` - direct coverage of the digit-count inference at all four precisions (10/13/16/19 digits), verifying conversion to a *different* target precision.
- `BinaryComparatorNarrowingTest.numericStringVersusDateInfersPrecisionFromDigitCountAndIsAntisymmetric` - reproduces the exact antisymmetry failure from the issue: a nanos-epoch string vs. a MILLIS `java.util.Date`, where misreading the nanos digits as millis used to flip the comparison's sign entirely.
- Both new tests fail against the pre-fix code (confirmed) and pass after the fix.
- Full `com.arcadedb.serializer.**`, `com.arcadedb.utility.**`, `com.arcadedb.schema.**` test packages pass with no regressions; verified no pre-existing test asserts on the old (buggy) numeric-string semantics.

## Context

Spawned by review on PR #5955 (fixing #5947, itself a follow-up to #5900/#5922) - see the `BinaryComparator` antisymmetry chain in the linked issues.

### DIFF
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -166,8 +166,8 @@ public class BinaryComparator {
     case BinaryTypes.TYPE_DATETIME_MICROS:
     case BinaryTypes.TYPE_DATETIME_NANOS: {
       final ChronoUnit higherPrecision = DateUtils.getHigherPrecision(value1, value2);
-      final long v1 = DateUtils.dateTimeToTimestamp(value1, higherPrecision);
-      final long v2 = DateUtils.dateTimeToTimestamp(value2, higherPrecision);
+      final long v1 = DateUtils.dateTimeToTimestampInferringStringPrecision(value1, higherPrecision);
+      final long v2 = DateUtils.dateTimeToTimestampInferringStringPrecision(value2, higherPrecision);
       return Long.compare(v1, v2);
     }
 
@@ -441,7 +441,8 @@ public class BinaryComparator {
     else if (a instanceof ChronoLocalDateTime<?> aDate && b instanceof ChronoLocalDateTime<?> bDate)
       return aDate.compareTo(bDate);
     else if (DateUtils.isDate(a) || DateUtils.isDate(b))
-      return DateUtils.dateTimeToTimestamp(a, ChronoUnit.NANOS).compareTo(DateUtils.dateTimeToTimestamp(b, ChronoUnit.NANOS));
+      return DateUtils.dateTimeToTimestampInferringStringPrecision(a, ChronoUnit.NANOS)
+          .compareTo(DateUtils.dateTimeToTimestampInferringStringPrecision(b, ChronoUnit.NANOS));
     return ((Comparable<Object>) a).compareTo(b);
   }
 
--- a/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
+++ b/engine/src/main/java/com/arcadedb/serializer/BinaryComparator.java
@@ -87,8 +87,8 @@ public class BinaryComparator {
         // the other side's own comparator and negate, rather than falling through to a lexicographic compareTo()
         // that silently ignores type2 and breaks antisymmetry the same way the narrowing branches did before this
         // fix (e.g. compare("2", STRING, 10, INT) and its reverse both answered "greater"). The DATE/DATETIME
-        // branch below already parses a String operand via DateUtils.dateTimeToTimestamp(), so this reuses that
-        // parsing rather than duplicating it (issue #5947).
+        // branch below already parses a String operand via DateUtils.dateTimeToTimestampInferringStringPrecision(),
+        // so this reuses that parsing rather than duplicating it (issue #5947).
         return -compare(value2, type2, value1, type1);
 
       default:
--- a/engine/src/main/java/com/arcadedb/utility/DateUtils.java
+++ b/engine/src/main/java/com/arcadedb/utility/DateUtils.java
@@ -240,15 +240,40 @@ public class DateUtils {
     } else if (value instanceof Number number)
       timestamp = number.longValue();
     else if (value instanceof String string) {
-      if (FileUtils.isLong(string))
-        timestamp = Long.parseLong(value.toString());
-      else
+      if (FileUtils.isLong(string)) {
+        final long rawValue = Long.parseLong(string);
+        timestamp = convertTimestamp(rawValue, inferEpochPrecision(rawValue), precisionToUse);
+      } else
         return dateTimeToTimestamp(database, parseIsoDateTime(database, string), precisionToUse);
     } else
       // UNSUPPORTED
       return null;
 
     return timestamp;
+  }
+
+  /**
+   * Infers the epoch precision a bare numeric string most likely represents, from its digit count: for a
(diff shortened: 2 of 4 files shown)


lvca:
Triage of the latest review's 3 minor observations:

1. **MICROS→NANOS overflow at bucket boundary** - addressed: added a javadoc note on `inferEpochPrecision` documenting that `convertTimestamp`'s widening multiplication has no saturation guard (pre-existing, not introduced by this PR).
2. **Empty-string `NumberFormatException`** - explicitly pre-existing in `dateTimeToTimestamp` and unrelated to this PR's change, so left as-is.
3. **Two-numeric-strings comparison test** - verified this scenario is actually unreachable in the current call graph: `TYPE_STRING` is not in `BinaryComparator`'s `TYPE_STRING`-branch delegate case list (only numeric/boolean/date types are), so `compare(str1, TYPE_STRING, str2, TYPE_STRING)` always falls through to plain lexicographic `compareTo()` and never reaches the `TYPE_DATE`/`TYPE_DATETIME*` switch arm. No test added since there's no reachable code path to regress.

lvca:
## CI status summary

All checks have completed. \`claude-review\` is clean with no blockers and no further changes requested. Every check passed except three CI jobs, all confirmed unrelated to this PR's changes (\`DateUtils.java\`, \`BinaryComparator.java\`, and their tests):

1. **integration-tests** - failed on \`PostgresWJdbcIT.parsingErrorMgmt\` (Postgres wire-protocol unicode-escape parsing). Pre-existing, deterministic failure already tracked as #5951, with fix PR #5957 already open from another session.
2. **unit-tests** - failed on \`LSMVectorIndexRebuildTest\`/\`LSMVectorIndexRecoveryTest\`: both are \`awaitility\` 2-minute \`ConditionTimeoutException\`s on background async vector-index rebuild threads, unrelated to date/comparator code. Confirmed via the job's log that \`DateUtilsTest\` (15 tests) and \`BinaryComparatorNarrowingTest\` (17 tests) both ran with 0 failures/0 errors.
3. **ha-integration-tests** - failed on \`Issue5410AbandonedTicketReleaseIT\` (Raft HA phase-2 ticket-release/replication timing test), unrelated to date/comparator code.

All three failures are in modules this PR never touches (postgresw, vector-index engine internals, ha-raft), and none of the tests directly exercised by this PR's changes (\`DateUtilsTest\`, \`BinaryComparatorNarrowingTest\`) showed any failures anywhere in the CI logs.

Given claude-review has no blockers and all remaining CI failures are confirmed pre-existing/unrelated, this PR is ready for review/merge from my side.

PR #5961, opened 2026-08-08
fix(#5932): LSMTreeIndexCursor compares typed bounds against the right key representation

### LINKED ISSUE #5932: LSMTreeIndexCursor compares raw (unconverted) range bounds against typed stored keys: ClassCastException on indexed range query with a numeric-string bound
## Summary

`LSMTreeIndexCursor`'s constructor mixes two representations of the query's range bounds and compares them inconsistently, causing a `ClassCastException` for an otherwise-valid indexed range query whose bound literal isn't already boxed as the index's exact key type (e.g. a numeric `String` against an `INTEGER`-typed indexed column).

Discovered while working on #5900 (a related but distinct `BinaryComparator`/`Type.convert` narrowing and crash-on-mismatch fix in PR #5922) - this is a **separate, pre-existing** defect, reproducible on unmodified `main`, unrelated to that PR's changes.

## Repro

```sql
CREATE VERTEX TYPE V;
CREATE PROPERTY V.n INTEGER;
CREATE INDEX ON V (n) NOTUNIQUE;
INSERT INTO V SET n = 10;
INSERT INTO V SET n = 20;
SELECT n FROM V WHERE n < '15'
```

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number
    at com.arcadedb.serializer.BinaryComparator.compare(BinaryComparator.java:62)
    at com.arcadedb.index.lsm.LSMTreeIndexAbstract.compareKeys(LSMTreeIndexAbstract.java:515)
    at com.arcadedb.index.lsm.LSMTreeIndexCursor.<init>(LSMTreeIndexCursor.java:247)
    at com.arcadedb.index.lsm.LSMTreeIndexMutable.lambda$range$1(LSMTreeIndexMutable.java:244)
    ...
    at com.arcadedb.query.sql.executor.FetchFromIndexStep.init(FetchFromIndexStep.java:393)
```

The un-indexed equivalent (`SELECT n FROM V WHERE n < '15'` with no index on `n`) works correctly and returns the row with `n = 10` - the string is parsed and compared as an integer via `Type.convert`/`BinaryComparator.compareTo` in the row-scan operator. Only the **indexed** path crashes.

A similar crash is reachable purely through an open, uncommitted transaction even without a fractional/mismatched bound at all, via a second, related site:

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number
    at com.arcadedb.serializer.BinaryComparator.compare(BinaryComparator.java:62)
    at com.arcadedb.index.lsm.LSMTreeIndexAbstract.compareKeys(LSMTreeIndexAbstract.java:515)
    at com.arcadedb.index.lsm.LSMTreeIndexCursor.getClosestEntryInTx(LSMTreeIndexCursor.java:669)
    at com.arcadedb.index.lsm.LSMTreeIndexCursor.<init>(LSMTreeIndexCursor.java:302)
```

(This second one only fires when the compared row was inserted in the same still-open transaction as the query, so the transactional-changes overlay is non-empty.)

## Root cause

`LSMTreeIndexCursor`'s constructor (`engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCursor.java`) computes properly-typed bounds via `index.convertKeys(...)`:

```java
final Object[] serializedFromKeys = index.convertKeys(fromKeys, binaryKeyTypes);
this.fromKeys = fromKeys;                                    // RAW, kept as a field
this.toKeys = toKeys != null && toKeys.length == 0 ? null : toKeys;  // RAW, kept as a field
this.serializedToKeys = index.convertKeys(this.toKeys, binaryKeyTypes); // CONVERTED, also kept as a field
```

Note `serializedFromKeys` is only a **local variable** here - never stored as a field - while the raw `this.fromKeys` and `this.toKeys` fields persist the unconverted bound (e.g. the literal `String "15"`, not the `Integer 15` the index actually stores its keys as).

Later, the constructor uses the wrong one of the pair for at least two internal consistency checks against already-typed data:

1. Around line 247, comparing the current page's (properly-typed) key against `this.toKeys` (raw) instead of `this.serializedToKeys` (converted):
   ```java
   final int compare = LSMTreeIndexMutable.compareKeys(comparator, binaryKeyTypes, pageCursor.getKeys(), this.toKeys);
   ```
2. Around line 230, similarly comparing `cursorKeys[i]` against `fromKeys` (raw) rather than the converted local `serializedFromKeys`.
3. In `getClosestEntryInTx(fromKeys, beginKeysInclusive)` (called at construction, and again later with `currentKeys != null ? currentKeys : fromKeys`), the raw `fromKeys` field is compared against `TransactionIndexContext`'s in-flight change keys, which are stored already-typed.

`BinaryComparator.compare(Object, byte, Object, byte)` trusts its caller that `value1`'s runtime class matches the declared `type1` and does an unguarded `(Number) value1` (or similar) cast for narrow numeric types - so a `String` reaching it there throws `ClassCastException` rather than comparing meaningfully.

## Suggested fix

Consistently use the **converted** (`serializedFromKeys`/`this.serializedToKeys`) arrays for every comparison against page-stored or transaction-overlay keys within `LSMTreeIndexCursor`, reserving the raw `fromKeys`/`toKeys` fields (if still needed at all) for anything that must echo the original caller-supplied values rather than compare them. Given `getClosestEntryInTx` navigates a `TreeMap` keyed by `TransactionIndexContext.ComparableKey` built from the **originally inserted** (also already-typed) values, the bound passed to it should be the converted one too - `serializedFromKeys` isn't currently kept as a field, so that would need to become one (or be threaded through differently) to reach the second call site at `getClosestEntryInTx(currentKeys != null ? currentKeys : fromKeys, false)`.

This is a delicate, heavily-optimized class (see the `#5635`/`#4947` contract notes already documented at its top) with careful comments about tombstone handling and TX-overlay navigation-key biasing, so a fix here deserves its own focused pass with dedicated regression tests (indexed range query with a numeric-string bound; indexed range query with a fractional-string bound; the open-transaction TX-overlay variant) rather than being folded into an unrelated PR.

## Environment

Version: main branch as of 2026-08-07 (also reproduces on 26.8.1). `engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCursor.java`.

lvca:
## Summary
Fixes #5932.

`LSMTreeIndexCursor`'s constructor mixed two different conversions of the range bounds and compared them inconsistently against already-typed keys:

- Several comparisons against page-stored or transaction-overlay keys used the **raw, unconverted** caller-supplied bound (e.g. the literal `String "15"` instead of the `Integer 15` the index actually stores). `BinaryComparator.compare` trusts the declared type and does an unguarded numeric cast, so this threw `ClassCastException` on a valid indexed range query whenever the bound literal wasn't already boxed as the index's exact key type.
- A second, related site (`getClosestEntryInTx`, used for both the transaction-overlay `TreeMap` navigation and its own internal bound filter) is reachable even without any committed page at all, purely through an open transaction.

Fixing this by simply making the already-converted local variable a field (as the issue's own suggested fix proposed) traded the crash for silent wrong results: `LSMTreeIndexAbstract.convertKeys()` additionally applies a disk-storage-oriented `String`-to-`byte[]` probe encoding, needed only by `lookupInPage`'s own raw-bytes binary search within a page. Every *other* place a bound is compared - against a fully deserialized page key (`PageIterator.getKeys()`) or a transaction-overlay key - holds the plain declared-type value (a `String`, never a `byte[]`), so handing it a `byte[]` bound either throws (inside `TransactionIndexContext.ComparableKey.compareTo`, which dispatches on runtime class) or silently miscompares (`BinaryComparator`'s `TYPE_STRING` branch falls back to comparing against a `byte[]`'s generic `Object.toString()`).

## Fix
Added `LSMTreeIndexAbstract.convertKeysToDeclaredTypes()`: the same per-component `Type.convert` narrowing as `convertKeys()`, without the disk-storage `byte[]`-for-`String` encoding. `LSMTreeIndexCursor` now keeps two consistently-used bound pairs:
- `serializedFromKeys`/`serializedToKeys` (existing, `byte[]`-encoded): used **only** to seed the two low-level page lookups (`lookupInPage` and the compacted series' equivalent) that need it.
- `typedFromKeys`/`typedToKeys` (new): used for **every** comparison against an already-deserialized key, page-read or transaction-overlay.

The raw, unconverted `fromKeys`/`toKeys` fields are removed entirely - nothing in the class needs to echo the original caller-supplied values back, only compare against them, so there is no longer a "wrong" representation to reach for by accident.

## Test plan
- New regression test `Issue5932IndexedRangeTypedBoundsTest` (6 cases): numeric-string bound on an indexed `INTEGER` column in both directions (`<` / `>`, exercising both the `toKeys` and `fromKeys` comparison sites), a `BETWEEN` with both bounds, a fractional-string bound on an indexed `DOUBLE` column, the open-transaction-only variant (no committed page at all, `getClosestEntryInTx`-only), and a guard against over-fixing across committed + in-transaction data.
- Verified each new test reproduces the original crash on unmodified `main` (via a temporary build), confirming the regression is real before fixing it.
- Ran the full `com.arcadedb.index.**` package (1208 tests) and `com.arcadedb.query.sql.**` package (2404 tests) - all passing. (Two unrelated `NoClassDefFoundError` failures seen on one full-package run reproduced as passing in isolation - a known stale-classloader artifact of the parallel test run, not a regression.)

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
@@ -405,18 +405,15 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
   }
 
   protected Object[] convertKeys(final Object[] keys, final byte[] keyTypes) {
-    if (keys == null)
+    final Object[] convertedKeys = convertKeysToDeclaredTypes(keys, keyTypes);
+    if (convertedKeys == null)
       return null;
 
-    final Object[] convertedKeys = new Object[keys.length];
-    for (int i = 0; i < keys.length; ++i) {
-      if (keys[i] == null)
-        continue;
-
-      convertedKeys[i] = Type.convert(database, keys[i], BinaryTypes.getClassFromType(keyTypes[i]));
-
+    for (int i = 0; i < convertedKeys.length; ++i) {
       if (convertedKeys[i] instanceof String string) {
-        // Apply case-insensitive collation: lowercase before storing/searching
+        // Apply case-insensitive collation: lowercase before storing/searching (already applied by
+        // convertKeysToDeclaredTypes() above, but toLowerCase() on an already-lowercase String is a no-op, and
+        // keeping the check here documents that BOTH methods must fold before a String reaches its own encoding)
         if (caseInsensitiveKeys != null && i < caseInsensitiveKeys.length && caseInsensitiveKeys[i])
           string = string.toLowerCase(Locale.ROOT);
         // OPTIMIZATION: ALWAYS CONVERT STRINGS TO BYTE[]
@@ -428,19 +425,23 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
 
   /**
    * Narrows each bound component to the property's declared Java type (e.g. the numeric {@code String} literal
-   * {@code "15"} to {@code Integer} for an {@code INTEGER} column), WITHOUT the disk-storage-oriented
-   * {@code String}-to-{@code byte[]} probe encoding {@link #convertKeys} additionally applies.
+   * {@code "15"} to {@code Integer} for an {@code INTEGER} column) and applies the same case-insensitive
+   * collation folding as {@link #convertKeys}, WITHOUT that method's additional disk-storage-oriented
+   * {@code String}-to-{@code byte[]} probe encoding.
    * <p>
    * That {@code byte[]} encoding exists only to seed {@code lookupInPage}'s (and the compacted series'
    * equivalent) own raw-bytes binary search within a page - every OTHER place a bound is compared against an
    * already-deserialized key needs this narrower conversion instead (#5932): a full page-entry deserialization
    * ({@code PageIterator.getKeys()}) hands back the plain declared-type value, e.g. a {@code String}, never a
    * {@code byte[]}, and so does {@code TransactionIndexContext}'s in-flight overlay (written by the record-save
-   * path, not through this index's own {@code convertKeys}). Comparing either against a {@code byte[]} bound
-   * either throws - inside {@code TransactionIndexContext.ComparableKey.compareTo}, which dispatches purely on
-   * the runtime class of both operands and has no {@code byte[]}-vs-{@code String} case - or silently
-   * miscompares, since {@link com.arcadedb.serializer.BinaryComparator}'s {@code TYPE_STRING} branch falls back
-   * to comparing against a {@code byte[]}'s generic {@code Object.toString()} instead of its content.
+   * path, not through this index's own {@code convertKeys}) - both already case-folded for a case-insensitive
+   * property, since {@code LSMTreeIndex.put()} writes them through {@code convertKeys()}. Comparing either
+   * against a {@code byte[]} bound either throws - inside {@code TransactionIndexContext.ComparableKey.compareTo},
+   * which dispatches purely on the runtime class of both operands and has no {@code byte[]}-vs-{@code String}
+   * case - or silently miscompares, since {@link com.arcadedb.serializer.BinaryComparator}'s {@code TYPE_STRING}
+   * branch falls back to comparing against a {@code byte[]}'s generic {@code Object.toString()} instead of its
+   * content; comparing against an un-folded bound would independently miscompare a mixed-case literal by plain
+   * byte order (e.g. {@code 'F'} 0x46 vs {@code 'f'} 0x66).
    */
   protected Object[] convertKeysToDeclaredTypes(final Object[] keys, final byte[] keyTypes) {
     if (keys == null)
@@ -452,6 +453,10 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
         continue;
 
       convertedKeys[i] = Type.convert(database, keys[i], BinaryTypes.getClassFromType(keyTypes[i]));
+
+      if (convertedKeys[i] instanceof String string && caseInsensitiveKeys != null && i < caseInsensitiveKeys.length
+          && caseInsensitiveKeys[i])
+        convertedKeys[i] = string.toLowerCase(Locale.ROOT);
     }
     return convertedKeys;
   }
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
@@ -426,6 +426,36 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
     return convertedKeys;
   }
 
+  /**
+   * Narrows each bound component to the property's declared Java type (e.g. the numeric {@code String} literal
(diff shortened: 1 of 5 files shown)


lvca:

- **Case-insensitive collation gap** (main point): `convertKeysToDeclaredTypes()` now applies the same `caseInsensitiveKeys` lowercasing as `convertKeys()`, refactored so both share that logic. Added a regression test (`caseInsensitiveIndexedStringRangeQueryWithDifferentlyCasedBoundReturnsCorrectRow`) that reproduces the gap on the pre-fix code and passes now.
- **Duplication**: `convertKeys()` now calls `convertKeysToDeclaredTypes()` for the shared declared-type + case-fold prefix, then layers its own `byte[]` encoding on top.
- **Test coverage**: added `compoundAndRangeConditionOnIndexedIntegerColumnUsesBothBoundsThroughIndexCursor`, an `AND` of two range conditions that plans as a single indexed range scan with both bounds set (unlike `BETWEEN`, confirmed via `prettyPrint()` to fall back to a bucket scan for this shape).
- Also addressed the Codacy "could be a local variable" finding on `serializedFromKeys`: since every comparison outside the constructor now goes through `typedFromKeys`/`typedToKeys`, `serializedFromKeys`/`serializedToKeys` are only ever used inside the constructor and are now locals instead of fields.

Full targeted regression suite (124 tests covering the changed paths, composite/geo/polymorphic index tests) passes.

lvca:

- **Latent instance of the same bug** (item #1): fixed `LSMTreeIndexMutable.range(Object[], boolean, Object[], boolean)` (the auto-detecting-direction overload) to narrow both bounds to their declared types before the direction comparison, matching the explicit-direction overload every real caller already uses. Added a dedicated regression test (`Issue5932RangeAutoDetectOverloadTest`) that calls this overload directly with a mismatched-type bound - verified it reproduces the `ClassCastException` on the pre-fix code.
- **Redundant fold** (item #2): `convertKeys()` no longer re-tests/re-lowercases the String `convertKeysToDeclaredTypes()` already folded - it now only layers the `byte[]` encoding on top.
- **Test file location** (item #3): leaving `Issue5932IndexedRangeTypedBoundsTest` under `com.arcadedb.query.sql.parser` - the review itself noted precedent exists both ways and called it a non-blocking nit. The new `Issue5932RangeAutoDetectOverloadTest` (Java-API-level, no SQL involved) is under `com.arcadedb.index.lsm`, consistent with the other LSM-internals tests mentioned.

172-test targeted regression sweep (every changed path plus composite/geo/polymorphic/case-insensitive index tests) passes.

lvca:
Third review round found no blocking issues ("No blocking issues found" / "well-reasoned, well-tested fix") and the remaining notes are all explicitly non-blocking:

- Duplicate `Type.convert` pass per cursor construction: reviewer's own assessment is "not a real performance concern... happy to see this stay as-is given how delicate this class already is."
- Dense Javadoc/comments: reviewer confirmed this matches the file's pre-existing style (the `#5635`/`#4947` contract notes), not a new deviation.
- `BETWEEN` test's doc-comment noting it falls back to a bucket scan: a note for future awareness if the planner ever changes, not an action item today.

Stopping here per the review-loop criteria (OK to merge, no major issues, remaining points not worth implementing).

Issue #5906, opened 2026-08-07
SUM()/AVG() over an INTEGER column silently overflow: Type.increment's long-upgrade guard widens after the int overflow, and ignores negative overflow


ruispereira:
## Summary

`SUM()` and `AVG()` over an integer column silently return wrong results once the running sum exceeds `Integer.MAX_VALUE`. `Type.increment()`, which the aggregators accumulate through, overflows in `int` arithmetic — and its own "upgrade to long" guard is broken, so it does not prevent the overflow it was written to catch. Negative overflow is not handled at all.

Version: 26.8.1. `Type.increment()` in `engine/schema/Type.java`, via `SQLFunctionAverage` / `SQLFunctionSum`.

## Reproducer

```sql
CREATE VERTEX TYPE V;
CREATE PROPERTY V.n INTEGER;
-- five rows of 2,000,000,000; true sum = 10,000,000,000 (fits in long, not int)
INSERT INTO V SET n = 2000000000;   -- x5
SELECT sum(n) AS s, avg(n) AS a, count(n) AS c FROM V;
```

```
s = 5705032704        -- expected 10000000000
a = 1.1410065408E9    -- expected 2.0E9
c = 5
```

Both `SUM` and `AVG` are wrong by a wide margin, no error.

## Root cause — the long-upgrade guard is a no-op

```java
// engine/schema/Type.java:772
public static Number increment(final Number a, final Number b) {
  ...
  case Integer i -> {
    case Integer integer -> {
      final int sum = a.intValue() + b.intValue();               // overflows in int
      if (sum < 0 && a.intValue() > 0 && b.intValue() > 0)
        // SPECIAL CASE: UPGRADE TO LONG
        return (long) (a.intValue() + b.intValue());             // BUG: same int overflow, then widened
      return sum;
    }
```

The guard recomputes `a.intValue() + b.intValue()` — the same overflowing `int` addition — and only then casts to `long`. `(long)(2000000000 + 2000000000)` is `(long)(-294967296)`, i.e. `-294967296L`, not `4000000000L`. The widen happens after the damage. It should widen the operands first:

```java
return (long) a.intValue() + (long) b.intValue();
```

Measured directly on `Type.increment`:

| call | result | correct |
|---|---|---|
| `increment(2000000000, 2000000000)` | `-294967296` | `4000000000` |
| `increment(-2000000000, -2000000000)` | `294967296` | `-4000000000` |
[...]

Issue #5668, opened 2026-08-01
ha-integration-tests is intermittently red on main, on a different test each time


lvca:
### Summary

`ha-integration-tests` fails intermittently on `main` and on PR branches, and **not on the same test twice** - which is what distinguishes this from a single flaky test and makes it worth one issue rather than two.

Observed while working #5636/#5640:

| Where | Test | Symptom |
| --- | --- | --- |
| PR #5640, run 30667085992 | `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied` | `[Node 1 must hold both vertices] expected: 2L but was: 1L`, with a suppressed `DatabaseAreNotIdentical: Page PageId(graph/28/0) ... DB1 2 <> DB2 1` |
| `main`, run 30676355936 | `RaftReplicationMaterializedViewIT.materializedViewReplicates` | error (different test, same job) |

Both are "the follower had not caught up when the assertion ran". The first one **passes locally in 15.3s against 42.7s for the same test in the failing CI run** - a ~3x wall-clock stretch - which points at the assertion racing a slow runner rather than at a replication defect.

### Why file it

Two costs, and the second is the one that matters:

1. A red `ha-integration-tests` on a PR that touches nothing in the Raft path gives the author no way to tell "my change broke replication" from "the runner was slow today". That is the same failure mode #5636 item 4 was about, and it is now the last one left in this suite.
2. It hides a real regression. Once the job is expected to be red, a genuine replication break in it reads as noise.

### Suggested direction

Rather than adding sleeps: give the HA ITs a shared "wait until follower N has applied up to txId X, or fail with what it actually reached" helper and use it at every cross-node assertion, so a slow runner costs time instead of a failure and a genuine stall reports the LSN it stopped at. `BaseRaftHATest` is the natural home.

Worth checking first whether the two above share one such assertion site, in which case one helper fixes both.

Found while working on #5636.

lvca:
Another data point from PR #5744 (run [30725200516](https://github.com/ArcadeData/arcadedb/actions/runs/30725200516)), since this issue is about establishing the pattern.

`ha-integration-tests` failed on that PR with:
- `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied`
[...]

PR #5967, opened 2026-08-08
fix(#5905,#5906): reject out-of-range integral narrowing instead of silently wrapping

### LINKED ISSUE #5905: Silent integer truncation: storing an out-of-range value in an INTEGER/SHORT/BYTE property corrupts it (3000000000 -> -1294967296) with no error
## Summary

Storing a numeric value that does not fit an `INTEGER` (or `SHORT`/`BYTE`) property silently truncates it by integer overflow, with no error and no warning. `INSERT INTO V SET n = 3000000000` into an `INTEGER` property stores **-1294967296**. The write succeeds; the value is corrupted.

Version: 26.8.1. `Type.convert()` in `engine/schema/Type.java`.

## Reproducer

```sql
CREATE VERTEX TYPE V;
CREATE PROPERTY V.n INTEGER;
INSERT INTO V SET n = 3000000000, id = 'big';
SELECT n FROM V WHERE id = 'big';
```

```
n = -1294967296        -- stored 3000000000, got back a negative number
```

Same result through the typed API:

```java
v.set("n", 3_000_000_000L);   // property declared INTEGER
v.save();
// reads back as -1294967296
```

Directly against `Type.convert`:

| call | result |
|---|---|
| `convert(3000000000L, Integer.class)` | `-1294967296` |
| `convert(2147483648L, Integer.class)` | `-2147483648` |
| `convert(9223372036854775807L, Integer.class)` | `-1` |

The conversion narrows with `.intValue()`, which wraps on overflow instead of rejecting.

## Why this is the worst option

A typed integer column has three possible behaviours for an out-of-range value, in decreasing order of safety:

1. **Reject** with an error (Postgres: `integer out of range`; MySQL strict mode: rejected).
2. **Widen / preserve** the value.
3. **Silently wrap** to a different number.

ArcadeDB does (3), the only one that loses data without telling anyone. The `INSERT` returns success, the application believes it stored 3 000 000 000, and a later `WHERE n = 3000000000` finds nothing while `WHERE n = -1294967296` finds the row. A range query or sum over the column is silently wrong.

It is trivially triggered: any value between `Integer.MAX_VALUE+1` and `Long.MAX_VALUE` assigned to an `INTEGER` property, which is an ordinary mistake (a count, a timestamp-derived id, an epoch-millis value in an INT column).

## Same root cause as #5900

The narrowing is the storage-side twin of the comparator narrowing in #5900: both reduce a wider numeric to `int`/`short`/`byte` with `.intValue()`/`.shortValue()`/`.byteValue()`, which overflows rather than rejecting. `SHORT` and `BYTE` properties truncate even more aggressively (16 / 8 bits).

## Suggested fix

In `Type.convert`, when narrowing an integral value to a smaller integral target, range-check before converting and throw a typed error (e.g. `CommandExecutionException` / a validation exception) naming the value, the target type and the property. `Math.toIntExact(long)` throws `ArithmeticException` on overflow and is the standard primitive for exactly this; the same shape covers `short`/`byte`.

Rejecting is the correct behaviour for a typed column, and it turns a silent corruption into a clear error at the point of the bad write.

Related string case (lower priority, same method): `convert("2147483648", Integer.class)` and `convert("abc", Integer.class)` throw a raw `NumberFormatException` rather than a typed conversion error — worth mapping while the narrowing is being fixed.

lvca:
## Summary

Fixes #5905 and #5906, both in `Type.java`'s integral narrowing/arithmetic.

- **#5905**: `Type.convert()` narrowed a wider `Number` to `BYTE`/`SHORT`/`INTEGER` with a plain `.byteValue()`/`.shortValue()`/`.intValue()`, which wraps two's-complement on overflow instead of rejecting (`3000000000L -> -1294967296`, no error, no warning). It now range-checks before narrowing and throws `IllegalArgumentException` naming the value, the target type and the property when known. `IllegalArgumentException` is the existing error convention used throughout `Type.java` for this method, and already classifies as `VALIDATION` for every wire protocol via `ErrorCategory`, so no new exception type or wire-layer wiring is needed.
- **#5906**: `Type.increment()`/`decrement()` had a broken "upgrade to long on overflow" guard: the `Integer+Integer`, `Integer+Short` and `Short+Integer` branches recomputed the *same* overflowing `int` addition/subtraction before widening it to `long` (`(long)(2000000000+2000000000)` is `-294967296L`, not `4000000000L`), and the guard only checked for positive overflow, leaving negative overflow undetected entirely. Both methods now use `Math.addExact`/`Math.subtractExact` and widen to `long` on the resulting `ArithmeticException`, matching the exact-arithmetic pattern already used in `MathExpression.exactIntegerArithmetic`. `SUM()`/`AVG()` accumulate through `Type.increment()`, so this also fixes their silent overflow over an `INTEGER` column.

The `Short+Short` branches in both methods never actually needed the guard (two `short` values can never sum/differ outside `int` range), so those are simplified to plain `int` arithmetic.

Also fixed the existing `incrementIntegerOverflow`/`incrementShortOverflow` unit tests, which only asserted the result's *class* (`Long`) and not its *value*, so they kept passing even with the overflow bug in place.

## Verification

- Added range-check and boundary-value unit tests to `TypeTest` for `convert()`, plus overflow/negative-overflow unit tests for `increment()`/`decrement()` across the `Integer`/`Short` branch combinations.
- Added `Issue5905IntegerTruncationTest`: SQL `INSERT`/`UPDATE` and typed-API (`MutableVertex.set`) reproductions confirming the out-of-range write is rejected and never lands in storage, plus an in-range control case.
- Added `Issue5906SumAverageIntegerOverflowTest`: SQL-level `SUM()`/`AVG()` over an `INTEGER` column exceeding `Integer.MAX_VALUE`, both positive and negative.
- Confirmed all 17 new/changed assertions fail against the pre-fix code (temporarily reverted `Type.java`) and pass after the fix.
- Ran `TypeTest` (118 tests), the new regression classes, `SQLFunctionSumAverageMultiArgTest`, `BinaryComparatorNarrowingTest`, `MathExpressionTest`, `DocumentValidationTest`, `UpdateStatementExecutionTest`, `CreatePropertyStatementExecutionTest`, the full `com.arcadedb.schema` package (382 tests) and the full `com.arcadedb.index` package (1098 tests) - all green, confirming the numeric-comparison and index-lookup paths that also call `Type.convert()`/`increment()` are unaffected.

## Test plan

- [x] `TypeTest` (118 tests) green
- [x] `Issue5905IntegerTruncationTest` (new) green
- [x] `Issue5906SumAverageIntegerOverflowTest` (new) green
- [x] `com.arcadedb.schema.**` (382 tests) green
- [x] `com.arcadedb.index.**` (1098 tests) green
- [x] Confirmed new tests fail against pre-fix code

## Behavioral note: bulk importers

`CSVImporterFormat`/`JSONImporterFormat` have no per-row error handling - they only catch `IOException` around the whole import loop. This was already true before this PR for malformed data (e.g. a non-numeric string into an `INTEGER` column already threw `NumberFormatException` and aborted the job). This PR extends the same "fail fast" behavior to one more case: a numeric value out of range for its declared type (mainly affects JSON import, since CSV values arrive as strings and already went through the pre-existing `Integer.parseInt`/`Short.parseShort`/`Byte.parseByte` path). Filed as a separate follow-up: #5968, since adding per-row skip-and-log resilience to the importers is its own design decision (opt-in flag, whole-row-vs-property skip, logging/reporting) independent of this fix.

Same "fail fast instead of silently wrap" note applies to the direct Java `Index` API (`LSMTreeIndex`/`HashIndex` `get()`/`range()`/etc.): a caller handing an out-of-range key literal straight to the low-level index API (bypassing SQL/schema `convert()`) now gets `IllegalArgumentException` where it previously silently produced a wrong/empty result. `FetchFromIndexStep` (the SQL-driven path) is already defended and unaffected.

### DIFF
--- a/engine/src/main/java/com/arcadedb/schema/Type.java
+++ b/engine/src/main/java/com/arcadedb/schema/Type.java
@@ -516,7 +516,7 @@ public enum Type {
         else if (value instanceof String string)
           return Byte.parseByte(string);
         else
-          return ((Number) value).byteValue();
+          return narrowToIntegral((Number) value, Byte.MIN_VALUE, Byte.MAX_VALUE, "BYTE", property).byteValue();
 
       } else if (targetClass.equals(Short.TYPE) || targetClass.equals(Short.class)) {
         if (value instanceof Short)
@@ -524,7 +524,7 @@ public enum Type {
         else if (value instanceof String string)
           return string.isEmpty() ? 0 : Short.parseShort(string);
         else
-          return ((Number) value).shortValue();
+          return narrowToIntegral((Number) value, Short.MIN_VALUE, Short.MAX_VALUE, "SHORT", property).shortValue();
 
       } else if (targetClass.equals(Integer.TYPE) || targetClass.equals(Integer.class)) {
         if (value instanceof Integer)
@@ -532,7 +532,7 @@ public enum Type {
         else if (value instanceof String string)
           return string.isEmpty() ? 0 : Integer.parseInt(string);
         else
-          return ((Number) value).intValue();
+          return narrowToIntegral((Number) value, Integer.MIN_VALUE, Integer.MAX_VALUE, "INTEGER", property).intValue();
 
       } else if (targetClass.equals(Long.TYPE) || targetClass.equals(Long.class)) {
         if (value instanceof Long)
@@ -784,6 +784,37 @@ public enum Type {
     return value;
   }
 
+  /**
+   * Range-checks {@code value} before narrowing it to a smaller integral type ({@code BYTE}, {@code SHORT} or
+   * {@code INTEGER}). Narrowing with a plain {@code .intValue()}/{@code .shortValue()}/{@code .byteValue()} wraps
+   * two's-complement on overflow instead of rejecting - {@code 3000000000L} silently became {@code -1294967296} -
+   * which corrupts the stored value without any error (issue #5905).
+   * <p>
+   * The comparison is done in {@code long}, so a {@link Double}/{@link Float} outside the long range (including
+   * {@code Infinity}) is caught via the saturating conversion {@link Number#longValue()} already performs. A
+   * {@link BigDecimal}/{@link BigInteger} can exceed even the long range, where {@code longValue()} truncates bits
+   * instead of saturating, so those two types are range-checked directly against {@code long} bounds first.
+   */
+  private static Number narrowToIntegral(final Number value, final long min, final long max, final String targetType,
+      final Property property) {
+    final long longValue;
+    if (value instanceof BigDecimal bigDecimal)
+      longValue = bigDecimal.compareTo(BigDecimal.valueOf(Long.MAX_VALUE)) > 0 ? Long.MAX_VALUE
+          : bigDecimal.compareTo(BigDecimal.valueOf(Long.MIN_VALUE)) < 0 ? Long.MIN_VALUE : bigDecimal.longValue();
+    else if (value instanceof BigInteger bigInteger)
+      longValue = bigInteger.compareTo(BigInteger.valueOf(Long.MAX_VALUE)) > 0 ? Long.MAX_VALUE
+          : bigInteger.compareTo(BigInteger.valueOf(Long.MIN_VALUE)) < 0 ? Long.MIN_VALUE : bigInteger.longValue();
+    else
+      longValue = value.longValue();
+
+    if (longValue < min || longValue > max)
+      throw new IllegalArgumentException(
+          "Value '" + value + "' is out of range for type " + targetType + " (" + min + " to " + max + ")" //
+              + (property != null ? " for property '" + property.getName() + "'" : ""));
+
+    return longValue;
+  }
+
   public static Number increment(final Number a, final Number b) {
     if (a == null || b == null)
       throw new IllegalArgumentException("Cannot increment a null value");
@@ -792,21 +823,23 @@ public enum Type {
     case Integer i -> {
       switch (b) {
       case Integer integer -> {
-        final int sum = a.intValue() + b.intValue();
-        if (sum < 0 && a.intValue() > 0 && b.intValue() > 0)
+        try {
+          return Math.addExact(a.intValue(), b.intValue());
+        } catch (final ArithmeticException e) {
           // SPECIAL CASE: UPGRADE TO LONG
-          return (long) (a.intValue() + b.intValue());
-        return sum;
+          return (long) a.intValue() + (long) b.intValue();
+        }
       }
       case Long l -> {
         return a.intValue() + b.longValue();
       }
       case Short aShort -> {
-        final int sum = a.intValue() + b.shortValue();
-        if (sum < 0 && a.intValue() > 0 && b.shortValue() > 0)
+        try {
+          return Math.addExact(a.intValue(), b.shortValue());
+        } catch (final ArithmeticException e) {
           // SPECIAL CASE: UPGRADE TO LONG
-          return (long) (a.intValue() + b.shortValue());
-        return sum;
+          return (long) a.intValue() + (long) b.shortValue();
+        }
       }
       case Float v -> {
         return a.intValue() + b.floatValue();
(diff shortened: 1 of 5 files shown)


lvca:
Addressed both review points:

- **BigInteger test coverage**: added the matching in-range/out-of-range `BigInteger` cases to `convertToIntegerOutOfRangeRejected()` in `4462f1792`.
- **Bulk importer behavior change**: this is real, but the underlying "no per-row error resilience" gap in `CSVImporterFormat`/`JSONImporterFormat` predates this PR (a malformed string already threw `NumberFormatException` and aborted the whole job before this change; this PR only adds the out-of-range-numeric case alongside it, mainly affecting JSON import since CSV values are strings and already went through the pre-existing parse path). Adding proper skip-and-log resilience to the importers is a separate design decision (opt-in flag, whole-row-vs-property skip, reporting), so I filed it as a follow-up: #5968, and called it out explicitly in the PR description rather than folding it into this fix.

Both commits pushed to this branch; `TypeTest` (now 118 tests) still green.

lvca:
Addressed round 2:

- **Index-rebuild blast radius**: verified and locked in with a test. `LSMTreeIndexAbstract.convertKeysToDeclaredTypes()`/`LSMTreeIndex.convertKeys()`/`HashIndex.convertKeys()` re-convert already-*stored* property values, and an INTEGER/SHORT/BYTE property's on-disk representation is a fixed-width primitive - there is no way to persist an out-of-range value for one of these types in the first place, which is exactly the invariant #5905 restores. So these three call sites can never receive an out-of-range raw value from a rebuild (or any other read-from-storage path); the only way to hit them with one would be constructing a raw out-of-range Java value and handing it directly to the low-level index API bypassing schema `convert()` entirely, which is a different, pre-existing (not introduced by this PR) concern. Added `Issue5967IndexRebuildBoundaryIntegerValueTest`, round-tripping `Integer.MAX_VALUE`/`MIN_VALUE` through `REBUILD INDEX` for both `LSM_TREE` and `HASH` - green, in `b3e7edbd6`.
- **BYTE/SHORT BigDecimal/BigInteger coverage**: added, same commit - `narrowToIntegral()`'s shared `BigDecimal`/`BigInteger` branch is now exercised for all three narrowing targets (`BYTE`/`SHORT`/`INTEGER`), not just `INTEGER`.
- **NaN → 0 coercion**: correctly identified as pre-existing (not a regression - `.longValue()`/`.intValue()` already did this before this PR) and out of scope for an integer-*overflow* fix, which is what #5905/#5906 are about. Not filing a separate follow-up for it since it's a distinct, much narrower edge case (a NaN literal assigned to an integral property) and the reviewer flagged it as optional/non-blocking; happy to file one if you'd rather have it tracked.

`TypeTest` (120 tests) and the new index test are green; pushed to this branch.

lvca:
Addressed round 3:

- **BigInteger negative-overflow coverage**: added, in `9840ae164` - the negative-magnitude case exercising `bigInteger.compareTo(BigInteger.valueOf(Long.MIN_VALUE)) < 0` now has a test, symmetric with the already-covered positive case.
- **Direct Java Index API blast radius**: this is the same point as round 2 (already investigated and confirmed safe for every read-from-storage path, including rebuild). Added the suggested one-line note to the PR description calling out that a caller going through the low-level `Index` API directly (bypassing SQL/schema `convert()`) now fails fast instead of silently misconverting, same as the CSV/JSON importer note already there.
- **NaN/Infinity handling**: same point as round 2 - pre-existing behavior (`Double.NaN.longValue() == 0` predates this PR), out of scope for an integer-*overflow* fix. Already declined with reasoning in the round 2 reply; not re-opening it.

`TypeTest` (118 tests) green; pushed to this branch.

lvca:
CI update: the `unit-tests` job reported FAILURE, but the single failing test is `TimeSeriesTagDictionaryTest.anIdWrittenByAnotherInstanceIsResolvedByReloading` (expected `"host_b"`, got `null`) - a page-reload/follower-consistency test for the time-series tag dictionary. This branch does not touch `engine/src/main/java/com/arcadedb/engine/timeseries/` or its tests at all (confirmed via `git diff main --stat`), so this is an unrelated pre-existing flake, not caused by this PR - out of 15,195 tests run in that job, this is the only failure.

`slow-unit-tests`, `integration-tests`, `build-and-package`, `builder-tests`, all e2e suites, all CodeQL languages, Codacy, and Meterian are all green. `ha-integration-tests` is still running as of this comment - it's the chronically-flaky check tracked separately in #5668/#5702 and isn't expected to gate this PR.

This PR is ready for merge review at @lvca's discretion.

PR #5983, opened 2026-08-09
fix(#5886): bound SQL MATCHES and openCypher =~ against catastrophic regex backtracking

### LINKED ISSUE #5886: SQL MATCHES exposes catastrophic regex backtracking and arcadedb.command.timeout does not stop it (thread pinned indefinitely)
## Summary

The SQL `MATCHES` operator hands the query author's regular expression straight to `java.util.regex` with no complexity bound, so a pathological pattern pins a query thread indefinitely. **`arcadedb.command.timeout` does not stop it**: with the timeout set to 3000 ms the query was still running after 30 seconds.

A 41-character subject and a 12-character pattern are enough.

Version: 26.8.1.

## Reproducer

```java
db.getConfiguration().setValue(GlobalConfiguration.COMMAND_TIMEOUT, 3000L);
db.transaction(() -> {
  final var v = db.newVertex("V");
  v.set("name", "a".repeat(40) + "!");
  v.save();
});

db.query("sql", "SELECT FROM V WHERE name MATCHES '(.*a){20}$'");
// still running after 30s, i.e. 10x the configured command timeout
```

Measured:

| step | result |
|---|---|
| `(.*a){20}$` against the same string, raw `java.util.regex` | still running after 5 s |
| same pattern through SQL `MATCHES` | still running after 30 s |
| with `arcadedb.command.timeout = 3000` | **still running after 30 s** |

I checked the premise before blaming the engine: three other classic patterns (`(a+)+$`, `(a|a)*$`, `(a*)*$`, `(x+x+)+y`) complete in 0 ms on JDK 21, so this is not "regex is slow in general" — it is one genuinely catastrophic pattern, and the engine offers it no protection.

## Why the timeout does not help

`java.util.regex` backtracking does not poll the interrupt flag, so nothing that works by interrupting or by checking a deadline between execution steps can reclaim the thread. The command-timeout machinery can mark the command as expired, but the worker stays inside `Matcher.matches()` until it finishes on its own — which for this pattern is not a useful amount of time.

That is the part worth fixing: it is not that a slow query is possible, it is that the mechanism operators are told to rely on does not apply.

## Impact

- Reachable by anyone who can run a query. In deployments where query access is granted to low-privilege users or exposed through an application layer that forwards user-supplied patterns, that is a very small blast radius to earn a permanently busy thread.
- Repeatable: each such query consumes one worker for an unbounded time. With the HTTP worker pool sized as usual, a handful of requests suffice.
- Invisible to the usual defences: it is not a large payload, not a large result set, not a deep parse tree, and not stopped by the timeout.

## Suggested fix

Options, roughly in order of how well they fit an embedded engine:

- Run `MATCHES` through a bounded matcher — the common approach is a `CharSequence` wrapper whose `charAt()` throws once a deadline or step budget is exceeded, which does interrupt backtracking because the engine calls it constantly. That keeps `java.util.regex` and makes the existing command timeout effective.
- Reject patterns with nested quantifiers at parse time. Cheap, but blunt and easy to bypass.
- Move to a non-backtracking engine (RE2/J) for `MATCHES`, which removes the class of problem entirely at the cost of some regex features.

The first is the smallest change that makes `arcadedb.command.timeout` mean what its description says for this operator.

Worth checking the other regex entry points for the same gap: the openCypher `=~` operator, and any `LIKE`-to-regex translation.

lvca:
## Summary

Fixes #5886. `java.util.regex` never polls an interrupt flag or checks a deadline while backtracking - the only thing it calls on every backtracking step is `CharSequence.charAt()` on the input. A pathological pattern like `(.*a){20}$` against a 41-character string triggers catastrophic backtracking and is still running 30+ seconds later, and `arcadedb.command.timeout` cannot stop it. Anyone who can reach a regex-based entry point with an attacker-controlled pattern - or who forwards user-supplied patterns from an application layer - can permanently tie up a worker thread with a handful of such operations.

What started as a fix for SQL `MATCHES` and openCypher `=~` (the two entry points named in the issue) grew, through the review process on this PR, into an audit of every user-reachable regex call site in the engine.

## Fix

`TimeBoundRegex` (new, `com.arcadedb.utility`) wraps the matched input in a `CharSequence` whose `charAt()` throws once a deadline elapses - the one interception point `java.util.regex` offers, for `matches()`, `replaceAll()`, and `split()` alike. The deadline comes from a new dedicated setting, `arcadedb.command.regexTimeout` (default 1000ms, per-database), independent of `arcadedb.command.timeout` and active even when that setting is left at its own default of disabled (`0`). The deadline is checked every 256 `charAt()` calls (a bitmask check) rather than every call, keeping it off the hot path for the overwhelming majority of patterns that never come close to that many backtracking steps.

**Covered entry points:**
- SQL `MATCHES` and openCypher `=~`
- SQL `LIKE`/`ILIKE`, including the native query engine's `SelectOperator` path (a sequence of unparenthesized `.*`/`*` segments reproduces the same catastrophic shape without any grouping/alternation/nested quantifiers - the original "LIKE is structurally safe" audit conclusion was wrong)
- `text.regexReplace()` and the `.normalize()` SQL method's optional pattern argument
- Full-text search's Lucene `RegexpQuery` (`/pattern/`) and `WildcardQuery` (`*`/`?`) support
- PromQL's `=~`/`!~` label matchers, including every step of a range query
- Schema-level `REGEXP` property validation (`CREATE PROPERTY ... REGEXP <pattern>`) - the most exposed of the group, since it runs on every insert/update of a validated property through any write path (REST, any wire protocol) with no query privileges needed

`.split(delimiter)`'s SQL method got a different fix: unlike its three siblings (`split()` function, `text.split()`, Cypher `split()`), which all treat the delimiter as a literal via `Pattern.quote(...)`, it passed the delimiter straight to `String.split(regex)` unescaped. Matched to its siblings, removing the risk entirely rather than just bounding it.

**Deadline sharing.** A regex evaluated once per row/token/item/property over a scan must not get a fresh timeout budget per item, or a table/index/document shaped so every item triggers catastrophic backtracking costs `itemCount * regexTimeout` instead of one bounded operation. `MATCHES`, `=~`, `text.regexReplace()`, and `.normalize()` share one deadline across an entire query execution via a new `CommandContext.getOrComputeRegexDeadline()` helper; full-text and schema `REGEXP` validation share one deadline across the whole scan/document; PromQL shares one deadline across an entire query's lifetime, including every step of a range query. `LIKE`/`ILIKE` are the one exception: `BinaryCompareOperator` has no `CommandContext` to cache a deadline on, and widening that interface was judged out of proportion here, so each row gets its own budget.

**Upgrade note.** Because the deadline is shared per query/scan rather than per match, and defaults to an enabled 1000ms, a legitimate non-catastrophic operation that previously ran slowly to completion can now fail with `TimeoutException` instead. Raise `arcadedb.command.regexTimeout` for any database with large-scan regex-heavy workloads.

Full detail in `docs/release-26.9.1.md`.

## Testing

Dedicated `TimeBoundRegexTest` for the utility, plus regression tests at every covered entry point (`MatchesConditionTest`, `OpenCypherWhereClauseTest`, `QueryHelperLikeRegexTimeoutTest`, `SelectExecutionTest`, `LikeOperatorTest`/`ILikeOperatorTest`, `TextRegexReplaceTest`, `SQLMethodNormalizeTest`, `SQLMethodSplitTest`, `FullTextQueryExecutorTest`, `PromQLEvaluatorIntegrationTest`, `DocumentValidationTest`), each covering "matches/replaces normally", "catastrophic pattern is aborted near the deadline", and (where sharing applies) "one budget is shared across N rows/items/steps/properties".

Two real bugs were caught and fixed by the review process, both with regression tests that fail without the fix:
- A deadline cache key colliding with an existing pattern-cache key namespace for a specific literal pattern text (`MatchesConditionTest#patternTextDeadlineDoesNotCollideWithTheDeadlineCacheKey`).
- Caching a deadline as an instance field on a Cypher AST node reused across executions via `CypherStatementCache`, causing spurious timeouts on later benign executions of the same cached query (`OpenCypherWhereClauseTest#regexDeadlineDoesNotLeakAcrossCachedQueryExecutions`).

## Test plan

- [x] `mvn -pl engine compile` / `test-compile` clean
- [x] Full affected test suite green, no regressions
- [x] New regression tests prove the abort happens near the configured deadline and that shared-deadline budgets hold across rows/items/steps/properties

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -327,116 +327,55 @@ only the "nothing at all was imported" guarantee differs between the two formats
 
 [#5968](https://github.com/ArcadeData/arcadedb/issues/5968)
 
-## SQL `MATCHES` and openCypher `=~` are no longer able to pin a worker thread indefinitely on a pathological regex (#5886)
+## Regex-based query and validation surfaces are no longer able to pin a worker thread indefinitely on a pathological pattern (#5886)
 
 A regex like `(.*a){20}$` against a 41-character string triggers catastrophic backtracking in
 `java.util.regex`: still running after 30+ seconds for a pattern and input small enough to be typed by
-hand. `arcadedb.command.timeout` did not help, because `Matcher.matches()` never polls an interrupt flag or
-checks a deadline while backtracking - the only thing it calls on every backtracking step is
-`CharSequence.charAt()` on the input being matched, and nothing in the surrounding query execution machinery
-gets a chance to run again until that call returns on its own. In a deployment where query access reaches
-low-privilege users or an application layer that forwards user-supplied patterns, a handful of such queries
-is enough to permanently tie up the HTTP worker pool.
+hand. `arcadedb.command.timeout` does not help, because `Matcher` never polls an interrupt flag or checks a
+deadline while backtracking - the only thing it calls on every backtracking step is `CharSequence.charAt()`
+on the input, and nothing in the surrounding query execution machinery gets a chance to run again until that
+call returns on its own. In a deployment where query access reaches low-privilege users, or an application
+layer forwards user-supplied patterns, a handful of such operations is enough to permanently tie up worker
+threads.
 
-Both regex entry points - SQL `MATCHES` (`MatchesCondition`) and openCypher `=~` (`RegexExpression`) - now
-run the match through `TimeBoundRegex`, which wraps the input `CharSequence` so `charAt()` throws once a
-deadline elapses. Because `java.util.regex` calls `charAt()` on every backtracking step, this turns the one
-call site the engine cannot otherwise intercept into the interruption point `arcadedb.command.timeout` was
-supposed to provide. The deadline itself is a new, dedicated setting, `arcadedb.command.regexTimeout`
-(default 1000ms), independent of `arcadedb.command.timeout` and active even when the latter is left at its
-default of disabled (`0`) - which it is by default, so `MATCHES`/`=~` would otherwise have had no bound at
-all regardless of configuration. `charAt()` checks the deadline every 256 calls rather than every call, to
-keep the check off the hot path for the overwhelming majority of well-behaved patterns that never come close
-to that many backtracking steps in the first place.
+A new utility, `TimeBoundRegex`, closes that gap by wrapping the matched input in a `CharSequence` that throws
+once a deadline elapses - the one interception point `java.util.regex` offers, for `matches()`, `replaceAll()`,
+and `split()` alike. The deadline comes from a new setting, `arcadedb.command.regexTimeout` (default 1000ms,
+per-database), independent of `arcadedb.command.timeout` and active even when that setting is left at its own
+default of disabled (`0`). The deadline is checked every 256 `charAt()` calls rather than every call, keeping
+the check off the hot path for the overwhelming majority of patterns that never come close to that many
+backtracking steps.
 
-A `MATCHES`/`=~` evaluation against a multi-value (list-typed) property shares one deadline across every item
-rather than giving each item its own full budget - otherwise a crafted list could still tie up the thread for
-`item count * regexTimeout`. Whichever item trips the deadline aborts the whole `WHERE` evaluation with a
-`TimeoutException`, the same as a single-value timeout: a catastrophic pattern is treated as a query failure,
-not as that one item silently failing to match, since swallowing it would let the pattern keep consuming its
-full budget on every future scan without ever surfacing.
+**Covered entry points:**
+- SQL `MATCHES` and openCypher `=~`
+- SQL `LIKE`/`ILIKE`, including the native query engine's `SelectOperator` path
+- `text.regexReplace()` and the `.normalize()` SQL method's optional pattern argument
+- Full-text search's Lucene `RegexpQuery` (`/pattern/`) and `WildcardQuery` (`*`/`?`) support
+- PromQL's `=~`/`!~` label matchers
+- Schema-level `REGEXP` property validation (`CREATE PROPERTY ... REGEXP <pattern>`) - notably reachable with
+  no query privileges at all, since it runs on every insert/update of a validated property through any write
+  path (REST, any wire protocol)
 
-A second review pass caught two more entry points the initial audit missed, both now bounded the same way:
-the `text.regexReplace(string, regex, replacement)` SQL function, whose only previous defense was a 500-char
-pattern-length cap plus a `catch (StackOverflowError)` - neither of which bounds a *time*-based catastrophic
-match, as opposed to a *stack-depth* one; and full-text search's Lucene `RegexpQuery` support (the `/pattern/`
-query syntax), arguably the worse instance of the two since a pathological pattern there is evaluated against
-every token in what is, absent a literal prefix, a full index scan. `TimeBoundRegex` gained a `replaceAll()`
-counterpart to `matches()` for the first case, and now also catches `StackOverflowError` alongside a deadline
-trip in both - a sufficiently pathological pattern can blow the (recursive) backtracking stack before the
-next 256-call checkpoint is reached, and that failure mode gets the same bound and the same `TimeoutException`
-as an explicit deadline.
+The `.split(delimiter)` SQL method got a different fix: unlike its three siblings (the `split()` function,
+`text.split()`, and Cypher's `split()`), which all treat the delimiter as a literal via `Pattern.quote(...)`,
+it passed the delimiter straight to `String.split(regex)` unescaped. Matched to its siblings' behavior, which
+removes the risk entirely rather than just bounding it.
 
-A third review pass corrected a wrong conclusion from the first: SQL `LIKE`/`ILIKE` had been audited as "safe"
-because `QueryHelper.convertForRegExp` escapes every regex metacharacter except `%` (`.*`) and `?` (`.`), so
-the generated pattern can never contain grouping, alternation, or nested quantifiers - true, but that only
-rules out *one* class of catastrophic backtracking. A sequence of several `.*` segments reproduces the same
-exponential blowup with no nesting at all (`%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%c` against a string of
-just `a`s hangs exactly like the issue's own `(.*a){20}$` reproducer), and full-text search's wildcard queries
-(`*`/`?`, the `WildcardQuery` sibling of the already-fixed `RegexpQuery` path) build the identical shape.
-`LIKE`/`ILIKE` (`QueryHelper.like`/`likeUntil`, used by `LikeOperator`, `ILikeOperator`, and the native query
-engine's `SelectOperator`) and full-text wildcard matching now run through `TimeBoundRegex` like every other
-entry point here, including the same shared-deadline treatment for a multi-value `LIKE`/`ILIKE` evaluation.
+**Deadline sharing.** A regex evaluated once per row/token/item over a scan must not get a fresh timeout
+budget per item - a table or index shaped so every row triggers catastrophic backtracking would then cost up
+to `itemCount * regexTimeout` instead of one bounded operation. `MATCHES`/`=~` share one deadline across an
+entire query execution (cached on the `CommandContext`, the same mechanism used to cache the compiled
+`Pattern`); full-text, PromQL, and `REGEXP` validation share one deadline across the whole scan they run
+against. `LIKE`/`ILIKE` are the one exception: `BinaryCompareOperator`, the interface they implement, has no
+`CommandContext` to cache a deadline on (only a `Database`), and widening that interface across every
+comparison operator was judged out of proportion here - each row still gets its own `regexTimeout` budget, so
+a `LIKE`-heavy scan where many rows are individually catastrophic is bounded per row but not for the scan as
+a whole.
 
-A fourth review pass found one more: PromQL's `=~`/`!~` label matchers (`PromQLEvaluator#matchesPostFilters`)
-compile the user-supplied pattern straight into `Pattern.matcher(...).matches()`, with only a static
-`REDOS_CHECK` pre-filter that flags *parenthesized* nested-quantifier/alternation shapes (`(a+)+`, `(a|aa)+`).
-A sequence of unparenthesized quantified segments (`a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*ac`, PromQL's
-equivalent of the `LIKE`/wildcard gap above) contains no `(` for that check to catch, and reaches the same
-unbounded `.matches()` call. Now bounded the same way, with one shared deadline across the whole row scan a
-label matcher runs against - the same N-rows-times-timeout concern as every other multi-item entry point here.
-
-**A sharper edge of the upgrade-behavior tradeoff, worth calling out on its own.** Full-text `RegexpQuery`/
-`WildcardQuery` and PromQL `=~`/`!~` each share one `regexTimeout` budget across an entire scan - every token
-in a full-text index, every row in a time-series range query - not per item, for the reasons above. That is
-correct (per-item budgets reopen the N-times-timeout bypass), but it means a large, *non-catastrophic* scan
-that legitimately takes more than the default 1000ms purely because there is a lot of data - a big full-text
-index, or a wide PromQL time range - now fails with `TimeoutException` where it previously (with
-`arcadedb.command.timeout` disabled by default) ran to completion, however slowly. If a workload does
-legitimate full-text or PromQL regex matching over a large volume of data, raise `arcadedb.command.regexTimeout`
-for that database rather than leaving it at the default.
-
-A fifth review pass found the most exposed entry point of all: the schema-level `REGEXP` property constraint
-(`CREATE PROPERTY ... REGEXP <pattern>`) validated `fieldValue.toString().matches(p.getRegexp())` on every
-insert/update of that property with no bound whatsoever - `DocumentValidator.validateField`. Unlike every
-other entry point here, this needs no query privileges at all: any write path (REST, any wire protocol) into
-a validated type is enough, and admin-defined validation patterns (classic email/URL regexes, in particular)
-are a notorious source of accidentally-catastrophic shapes. Now bounded the same way as everything else - which
-means the same default-1000ms tradeoff described above for full-text/PromQL applies here too: a legitimate,
-non-catastrophic pattern validated against a very large field value that previously took over a second (however
-undesirable that already was) now fails the write with `TimeoutException` instead. Raise
-`arcadedb.command.regexTimeout` for a database with that kind of validation workload.
-
-Also fixed: `MATCHES`/`=~` shared one deadline across a multi-value evaluation (an earlier fix in this issue)
-but still gave each *row* of a table/graph scan its own fresh budget, so a table shaped so every row triggers
-catastrophic backtracking could still cost up to `rowCount * regexTimeout` overall. Both now cache one deadline
-for the lifetime of the query on the `CommandContext` - the same per-execution object `MatchesCondition` already
-used to cache the compiled `Pattern`. `RegexExpression` needed the same context-based caching, not an AST-node
-instance field: unlike a compiled `Pattern` (content-addressed, safe to reuse indefinitely), a deadline is a
-point in time, and `CypherStatementCache` caches parsed queries - this AST node included - by query text across
-repeated executions, so an instance-field deadline would go stale after the first execution and then make every
-later run of that same cached query throw a spurious `TimeoutException` on any pattern, benign or not, until
-LRU eviction; a 6th review pass caught this exact regression before it merged. `LIKE`/`ILIKE` could not get the
-per-scan fix: `BinaryCompareOperator`, the interface `LikeOperator`/`ILikeOperator` implement, has no
-`CommandContext` to cache a query-wide deadline on (only a `Database`), and widening that interface across every
-comparison operator was judged out of proportion for this fix - a conscious, narrower-scope tradeoff, documented
-in `LikeOperator`, not an oversight.
-
-A 7th review pass found a second regression, caught the same way: `MatchesCondition`'s query-wide deadline was
-cached under the fixed key `"MATCHES_DEADLINE"` in the same flat, string-keyed `CommandContext` cache used for
-compiled patterns (keyed `"MATCHES_" + <regex text>`). A query using the literal pattern text `DEADLINE` (e.g.
-`WHERE name MATCHES 'DEADLINE'`) produced the identical key for both, so the Pattern and the deadline overwrote
-each other's cache slot and every row threw `ClassCastException` on the very first evaluation - 100%
-reproducible, not a theoretical race. The deadline key now lives entirely outside the `"MATCHES_"` namespace, so
-it cannot collide with `"MATCHES_" + <any regex text>` regardless of what that text reads.
-
-The same pass also found two more unbounded entry points reachable from plain SQL: the `.split(delimiter)`
-method syntax (`field.split(pattern)`) passed its delimiter straight to `String.split(regex)` unescaped, unlike
-its three siblings (the `split()` function, `text.split()`, and Cypher's `split()`), which all correctly treat
-the delimiter as a literal via `Pattern.quote(...)` - now fixed to match them, which removes the risk entirely
-rather than just bounding it, since a `Pattern.quote()`d literal can never form a catastrophic shape. The
-`.normalize(form, pattern)` method's optional second argument, by contrast, is *meant* to be a real regex (it
-controls what gets stripped after Unicode normalization), so that one is bounded through `TimeBoundRegex` like
-every other intentional regex entry point, not literalized.
+**Upgrade note.** Because the deadline is shared per query/scan rather than per match, and defaults to an
+enabled 1000ms, a *legitimate, non-catastrophic* operation that previously ran slowly to completion - a
+`MATCHES`/`LIKE` query over a very large table, a full-text or PromQL scan over a lot of data, or `REGEXP`
+validation against a large field value - can now fail with `TimeoutException` instead. Raise
+`arcadedb.command.regexTimeout` for any database with that kind of workload.
 
 [#5886](https://github.com/ArcadeData/arcadedb/issues/5886)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -386,4 +386,14 @@ equivalent of the `LIKE`/wildcard gap above) contains no `(` for that check to c
 unbounded `.matches()` call. Now bounded the same way, with one shared deadline across the whole row scan a
 label matcher runs against - the same N-rows-times-timeout concern as every other multi-item entry point here.
 
+**A sharper edge of the upgrade-behavior tradeoff, worth calling out on its own.** Full-text `RegexpQuery`/
+`WildcardQuery` and PromQL `=~`/`!~` each share one `regexTimeout` budget across an entire scan - every token
+in a full-text index, every row in a time-series range query - not per item, for the reasons above. That is
+correct (per-item budgets reopen the N-times-timeout bypass), but it means a large, *non-catastrophic* scan
+that legitimately takes more than the default 1000ms purely because there is a lot of data - a big full-text
+index, or a wide PromQL time range - now fails with `TimeoutException` where it previously (with
+`arcadedb.command.timeout` disabled by default) ran to completion, however slowly. If a workload does
+legitimate full-text or PromQL regex matching over a large volume of data, raise `arcadedb.command.regexTimeout`
+for that database rather than leaving it at the default.
+
 [#5886](https://github.com/ArcadeData/arcadedb/issues/5886)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -372,7 +372,13 @@ the rows within one step. `LIKE`/`ILIKE` are the one exception: `BinaryCompareOp
 implement, has no `CommandContext` to cache a deadline on (only a `Database`), and widening that interface
 across every comparison operator was judged out of proportion here - each row still gets its own
 `regexTimeout` budget, so a `LIKE`-heavy scan where many rows are individually catastrophic is bounded per row
-but not for the scan as a whole.
+but not for the scan as a whole. A parallel bucket scan (SQL's default for a type spread across multiple
+buckets) is a second, narrower exception: each worker thread gets its own copy of the `CommandContext`
+(deliberately, so workers don't race on a shared, non-thread-safe cache), so it computes its own deadline
+independently rather than sharing the sequential-scan path's single deadline - a type scanned in parallel
+across N buckets is bounded by `N * regexTimeout` overall, not one shared budget. Unlike the row/item counts
+this issue defends against, bucket count is a schema/DDL property, not attacker-controlled, so this is a much
+narrower gap than the one this fix closes.
 
 **Upgrade note.** Because the deadline is shared per query/scan rather than per match, and defaults to an
 enabled 1000ms, a *legitimate, non-catastrophic* operation that previously ran slowly to completion - a
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -401,16 +401,25 @@ A fifth review pass found the most exposed entry point of all: the schema-level 
 insert/update of that property with no bound whatsoever - `DocumentValidator.validateField`. Unlike every
 other entry point here, this needs no query privileges at all: any write path (REST, any wire protocol) into
 a validated type is enough, and admin-defined validation patterns (classic email/URL regexes, in particular)
-are a notorious source of accidentally-catastrophic shapes. Now bounded the same way as everything else.
+are a notorious source of accidentally-catastrophic shapes. Now bounded the same way as everything else - which
+means the same default-1000ms tradeoff described above for full-text/PromQL applies here too: a legitimate,
+non-catastrophic pattern validated against a very large field value that previously took over a second (however
+undesirable that already was) now fails the write with `TimeoutException` instead. Raise
+`arcadedb.command.regexTimeout` for a database with that kind of validation workload.
 
 Also fixed: `MATCHES`/`=~` shared one deadline across a multi-value evaluation (an earlier fix in this issue)
 but still gave each *row* of a table/graph scan its own fresh budget, so a table shaped so every row triggers
 catastrophic backtracking could still cost up to `rowCount * regexTimeout` overall. Both now cache one deadline
-for the lifetime of the query - `MatchesCondition` on the `CommandContext` (the same mechanism already used to
-cache the compiled `Pattern`), `RegexExpression` as an AST-node instance field (ditto) - closing the same gap
-already closed for full-text and PromQL scans. `LIKE`/`ILIKE` could not get the same fix: `BinaryCompareOperator`,
-the interface `LikeOperator`/`ILikeOperator` implement, has no `CommandContext` to cache a query-wide deadline
-on (only a `Database`), and widening that interface across every comparison operator was judged out of
-proportion for this fix - a conscious, narrower-scope tradeoff, documented in `LikeOperator`, not an oversight.
+for the lifetime of the query on the `CommandContext` - the same per-execution object `MatchesCondition` already
+used to cache the compiled `Pattern`. `RegexExpression` needed the same context-based caching, not an AST-node
+instance field: unlike a compiled `Pattern` (content-addressed, safe to reuse indefinitely), a deadline is a
+point in time, and `CypherStatementCache` caches parsed queries - this AST node included - by query text across
+repeated executions, so an instance-field deadline would go stale after the first execution and then make every
+later run of that same cached query throw a spurious `TimeoutException` on any pattern, benign or not, until
+LRU eviction; a 6th review pass caught this exact regression before it merged. `LIKE`/`ILIKE` could not get the
+per-scan fix: `BinaryCompareOperator`, the interface `LikeOperator`/`ILikeOperator` implement, has no
+`CommandContext` to cache a query-wide deadline on (only a `Database`), and widening that interface across every
+comparison operator was judged out of proportion for this fix - a conscious, narrower-scope tradeoff, documented
+in `LikeOperator`, not an oversight.
 
 [#5886](https://github.com/ArcadeData/arcadedb/issues/5886)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -361,4 +361,15 @@ escapes every regex metacharacter except `%` (`.*`) and `?` (`.`), so the genera
 grouping, alternation, or nested quantifiers - the shapes catastrophic backtracking requires. It stays on
 plain `String.matches()`.
 
+A second review pass caught two more entry points the initial audit missed, both now bounded the same way:
+the `text.regexReplace(string, regex, replacement)` SQL function, whose only previous defense was a 500-char
+pattern-length cap plus a `catch (StackOverflowError)` - neither of which bounds a *time*-based catastrophic
+match, as opposed to a *stack-depth* one; and full-text search's Lucene `RegexpQuery` support (the `/pattern/`
+query syntax), arguably the worse instance of the two since a pathological pattern there is evaluated against
+every token in what is, absent a literal prefix, a full index scan. `TimeBoundRegex` gained a `replaceAll()`
+counterpart to `matches()` for the first case, and now also catches `StackOverflowError` alongside a deadline
+trip in both - a sufficiently pathological pattern can blow the (recursive) backtracking stack before the
+next 256-call checkpoint is reached, and that failure mode gets the same bound and the same `TimeoutException`
+as an explicit deadline.
+
 [#5886](https://github.com/ArcadeData/arcadedb/issues/5886)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -356,11 +356,6 @@ rather than giving each item its own full budget - otherwise a crafted list coul
 not as that one item silently failing to match, since swallowing it would let the pattern keep consuming its
 full budget on every future scan without ever surfacing.
 
-SQL `LIKE`/`ILIKE` were audited for the same gap and found not to need it: `QueryHelper.convertForRegExp`
-escapes every regex metacharacter except `%` (`.*`) and `?` (`.`), so the generated pattern can never contain
-grouping, alternation, or nested quantifiers - the shapes catastrophic backtracking requires. It stays on
-plain `String.matches()`.
-
 A second review pass caught two more entry points the initial audit missed, both now bounded the same way:
 the `text.regexReplace(string, regex, replacement)` SQL function, whose only previous defense was a 500-char
 pattern-length cap plus a `catch (StackOverflowError)` - neither of which bounds a *time*-based catastrophic
@@ -371,5 +366,16 @@ counterpart to `matches()` for the first case, and now also catches `StackOverfl
 trip in both - a sufficiently pathological pattern can blow the (recursive) backtracking stack before the
 next 256-call checkpoint is reached, and that failure mode gets the same bound and the same `TimeoutException`
 as an explicit deadline.
+
+A third review pass corrected a wrong conclusion from the first: SQL `LIKE`/`ILIKE` had been audited as "safe"
+because `QueryHelper.convertForRegExp` escapes every regex metacharacter except `%` (`.*`) and `?` (`.`), so
+the generated pattern can never contain grouping, alternation, or nested quantifiers - true, but that only
+rules out *one* class of catastrophic backtracking. A sequence of several `.*` segments reproduces the same
+exponential blowup with no nesting at all (`%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%c` against a string of
+just `a`s hangs exactly like the issue's own `(.*a){20}$` reproducer), and full-text search's wildcard queries
+(`*`/`?`, the `WildcardQuery` sibling of the already-fixed `RegexpQuery` path) build the identical shape.
+`LIKE`/`ILIKE` (`QueryHelper.like`/`likeUntil`, used by `LikeOperator`, `ILikeOperator`, and the native query
+engine's `SelectOperator`) and full-text wildcard matching now run through `TimeBoundRegex` like every other
+entry point here, including the same shared-deadline treatment for a multi-value `LIKE`/`ILIKE` evaluation.
 
 [#5886](https://github.com/ArcadeData/arcadedb/issues/5886)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -422,4 +422,21 @@ per-scan fix: `BinaryCompareOperator`, the interface `LikeOperator`/`ILikeOperat
 comparison operator was judged out of proportion for this fix - a conscious, narrower-scope tradeoff, documented
 in `LikeOperator`, not an oversight.
 
+A 7th review pass found a second regression, caught the same way: `MatchesCondition`'s query-wide deadline was
+cached under the fixed key `"MATCHES_DEADLINE"` in the same flat, string-keyed `CommandContext` cache used for
+compiled patterns (keyed `"MATCHES_" + <regex text>`). A query using the literal pattern text `DEADLINE` (e.g.
+`WHERE name MATCHES 'DEADLINE'`) produced the identical key for both, so the Pattern and the deadline overwrote
+each other's cache slot and every row threw `ClassCastException` on the very first evaluation - 100%
+reproducible, not a theoretical race. The deadline key now lives entirely outside the `"MATCHES_"` namespace, so
+it cannot collide with `"MATCHES_" + <any regex text>` regardless of what that text reads.
+
+The same pass also found two more unbounded entry points reachable from plain SQL: the `.split(delimiter)`
+method syntax (`field.split(pattern)`) passed its delimiter straight to `String.split(regex)` unescaped, unlike
+its three siblings (the `split()` function, `text.split()`, and Cypher's `split()`), which all correctly treat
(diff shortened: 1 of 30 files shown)


lvca:
Thanks for the review. Responses to each point:

**1. Default `regexTimeout` value (1000ms) vs. `command.timeout` defaulting to unlimited.** Agreed this is the one real decision here — asking @lvca directly rather than picking a number myself.

**2. Per-row config lookup.** Leaving as-is: it's a `ConcurrentHashMap` get, negligible next to the regex match itself (confirmed below), and caching it would add a cache-invalidation surface for a value that can't change mid-query for no measurable benefit.

**3. Allocation/indirection overhead.** Ran an informal check: 200k matches/round, 20 rounds, warmed up, literal-ish pattern (`email` shape) — plain `Matcher.matches()` and `TimeBoundRegex.matches()` both averaged ~6ms/round (~30ns/match either way), i.e. noise-level difference. Not a throughput concern.

**4. Fixed** in 16af3ef4 — `DeadlineBoundCharSequence`'s call counter is now shared by reference across a sequence and everything derived from it via `subSequence()`, so the check cadence no longer resets per sub-sequence. You're right that `java.util.regex` doesn't currently route `matches()` backtracking through `subSequence()`, but sharing the counter removes the assumption instead of just documenting it, at effectively zero cost.

**5. Test timing margin.** Agreed low-risk as noted; keeping the 5s margin on a 200ms deadline for CI-runner headroom.

lvca:
Decision on point 1: keeping the 1000ms default. Regex matches are normally sub-millisecond even on large text, legitimate patterns needing more time are rare, and it's a per-database setting anyone hitting the edge case can raise. A tight default closes the security gap immediately for anyone running with `arcadedb.command.timeout` at its own default (disabled).

lvca:
Pushed fc07cb14 addressing both points from the second review:

**Finding 1 (multi-value per-item timeout budget): fixed as suggested.** `TimeBoundRegex` now exposes `newDeadline(timeoutMillis)` / `matchesUntil(pattern, input, deadlineNanos)` so `MatchesCondition` computes one deadline before the multi-value loop and shares it across every item, instead of each item getting its own fresh `timeoutMillis` budget.

**Writing the regression test for that surfaced a more serious, separate bug**: both `MatchesCondition` and `RegexExpression` were reading the timeout via `context.getConfiguration()`, which for a plain `SELECT`/`=~` evaluation is a fresh, disconnected `ContextConfiguration` — nothing wires it to the database's settings. It silently fell back to the compiled-in 1000ms default and **ignored any per-database `arcadedb.command.regexTimeout` override**, exactly the way `db.getConfiguration().setValue(...)` is used in the original issue's own reproducer. `SelectExecutionPlanner` reads `arcadedb.command.timeout` via `context.getDatabase().getConfiguration()` for precisely this reason; both regex entry points now match that pattern.

Net effect: the fix was still correct and on by default (the compiled-in 1000ms default was always being applied), but a database operator lowering or raising `arcadedb.command.regexTimeout` from its default had no effect until this commit. Added `multiValueMatchesSharesOneTimeoutBudgetAcrossItems` to cover finding 1; the config-resolution bug is what made that test's failure mode obvious (it was consistently landing at ~1000ms regardless of the 200ms configured in the test, rather than N × 200ms as expected from finding 1 alone).

All tests green: `TimeBoundRegexTest`, `MatchesConditionTest` (7 tests), `OpenCypherWhereClauseTest` regex-related tests (13 tests).

lvca:
Thanks — responses to the remaining points, none of which needed a code change beyond one doc line:

**Per-match overhead on the hot path.** Already measured this for the earlier review (200k matches/round, 20 rounds, warmed up, literal-ish pattern): plain `Matcher.matches()` and `TimeBoundRegex.matches()` both averaged ~6ms/round, i.e. noise-level difference (~30ns/match either way). The `charAt()` indirection and mask check are cheap enough not to show up against `Matcher` overhead itself.

**Default 1000ms being a behavior change on upgrade.** Decision already made and recorded above: keeping 1000ms. Regex matches are normally sub-millisecond even on large text; the setting is per-database and easy to raise for the rare legitimate case that needs more.

**Whole-query abort vs. per-row on multi-value timeout.** Agreed this deserved to be explicit rather than implied — added a paragraph to the release notes (dfc78178) confirming it's intentional: a catastrophic item fails the query with `TimeoutException` rather than silently excluding itself, since swallowing it would let the pattern keep consuming its full budget on every future scan without ever surfacing.

**Test gap: no test proving a normal match is unaffected by a pattern that exercises `subSequence()`.** Not adding this one — `java.util.regex`'s `matches()`/backtracking path doesn't route through `CharSequence.subSequence()` for any pattern shape I could construct (confirmed while writing the original fix), so there's no way to write this test without depending on JDK-internal behavior that isn't part of any documented contract. The shared-counter fix stays in as defense-in-depth per the earlier review; asserting "not reachable today" would just be asserting the same JDK-internal fact the fix already doesn't rely on.

At this point I believe all findings from both review passes are either fixed or addressed, and nothing outstanding is blocking. Let me know if anything above should be revisited.

lvca:
Pushed 691bc73a addressing everything from this review:

**Security - two missed entry points (the big one): fixed.**
- `text.regexReplace(string, regex, replacement)`: `TimeBoundRegex` gained a `replaceAll()` counterpart to `matches()`; the function now reads `arcadedb.command.regexTimeout` the same way `MatchesCondition`/`RegexExpression` do, with the existing `IllegalArgumentException` contract preserved (a `TimeoutException` from the bound is caught and rewrapped, same as the existing `PatternSyntaxException` handling).
- Full-text `RegexpQuery` (`/pattern/` syntax): `collectRegexpMatches()` now shares one deadline across the whole token scan (same N-times-budget concern as the multi-value `MATCHES` fix, since this evaluates one pattern against potentially every token in the index). `LSMTreeFullTextIndex` gained a `getDatabase()` accessor mirroring the `underlyingIndex.getMutableIndex().getDatabase()` idiom already used internally in that class 6+ times, so `FullTextQueryExecutor` has a way to reach the config at all.

Agreed this was arguably worse than the two originally-fixed sites (RegexpQuery runs the pathological pattern against every token in a full scan) and appreciate the catch — the original audit only looked at the SQL/Cypher condition operators, not SQL functions or search-query engines.

**Possible NPE via `context.getDatabase()`: investigated, no NPE risk for `MatchesCondition`/`RegexExpression`, but a real one for `text.regexReplace` — fixed.**
Verified every top-level SQL and openCypher statement (`SelectStatement`, `UpdateStatement`, `InsertStatement`, `CypherExecutionPlan`, etc.) calls `context.setDatabase(...)` before any `WHERE`/`RETURN` expression can be evaluated, so a `MATCHES`/`=~` condition always has a database bound — this mirrors the existing unguarded `SelectExecutionPlanner` pattern for `COMMAND_TIMEOUT`, confirmed as the same safe invariant. No change needed there.

`text.regexReplace` is different: it's a lower-level SQL-function entry point, and this exact test file's siblings (`TextStatelessFunctionsTest`) already call functions directly with a `null` `CommandContext` as an established convention. That would have NPE'd. Fixed by falling back to a fresh `ContextConfiguration()` (itself just a proxy for the compiled-in default) when `context`/`context.getDatabase()` is unset — added `functionIsSafeWithoutACommandContext` to cover it.

**`StackOverflowError` not caught: fixed.** `TimeBoundRegex`'s internal `run()` now catches `RegexDeadlineExceeded | StackOverflowError` together and converts either to `TimeoutException` — a pathological pattern that blows the recursive backtracking stack before the next 256-call checkpoint gets the same bound as one that trips the deadline first. This also made `TextRegexReplace`'s own separate `catch (StackOverflowError)` dead code, so it's now just the shared `TimeoutException` path (still surfaced as `IllegalArgumentException` per that function's existing contract).

**`nanoTime()` overflow-unsafe comparison: fixed**, switched to `System.nanoTime() - deadlineNanos >= 0` per the Javadoc-recommended idiom.

**Test coverage gap (multi-value margin too tight): fixed.** Widened from 5 items/800ms to 10 items/1000ms — same discriminating power (expected ~200-300ms if shared, ~2000ms if not) with a wider absolute margin on both sides for CI-runner slack.

**Performance / default value**: no new action, already addressed in earlier rounds (measured negligible overhead, decision made to keep the 1000ms default).

New tests: `TimeBoundRegexTest` gained `replaceAll` coverage (normal/disabled-bound/catastrophic), `TextRegexReplaceTest` is new (basic replace, null-context safety, catastrophic-pattern abort, existing length-cap guard still fires), `FullTextQueryExecutorTest` gained `catastrophicRegexpQueryIsAbortedByRegexTimeout`. All green, plus reran `CypherBuiltInFunctionsTest` (145), `CypherFunctionSecurityTest` (28), `OpenCypherTextFunctionsTest` (86), and the regex-related `OpenCypherWhereClauseTest` methods (13) to check nothing else touching `regexReplace`/`=~` regressed.

lvca:
Pushed 87f33fc4. This round found something worth flagging explicitly: the wildcard-query finding didn't just mean fixing a sibling method - it invalidated this PR's own earlier "LIKE/ILIKE are structurally safe" conclusion.

**Bug: sibling wildcard-query path unbounded — fixed, and it corrects an earlier mistake of mine.**
`wildcardToRegex()` turns `*` into `.*` with no grouping/alternation/nested quantifiers - which is exactly the property my original LIKE/ILIKE audit relied on to call them safe. But a sequence of several `.*` segments reproduces catastrophic backtracking on its own; nesting isn't required, just several ambiguous `.*`s in a row against input that ultimately fails to match. I verified this empirically before touching anything: `.*a.*a.*a...` (20 reps) against a plain string of `a`s hangs exactly like the issue's own `(.*a){20}$` reproducer. Since `QueryHelper.convertForRegExp` (`%` → `.*`) produces the identical shape, that means SQL `LIKE`/`ILIKE` were vulnerable too, contrary to what I stated in this PR's description and the release notes.

Fixed all of it: `QueryHelper.like()`/new `likeUntil()` now route through `TimeBoundRegex`, mirroring `matches()`/`matchesUntil()`. Updated all three independent call sites - `LikeOperator`, `ILikeOperator`, and the native query engine's `SelectOperator` (a third, separate LIKE/ILIKE evaluation path I hadn't previously touched) - plus `collectWildcardMatches` in `FullTextQueryExecutor`, sharing one deadline across its full scan the same way `collectRegexpMatches` already does. Corrected the release notes to say so plainly rather than just supplementing the wrong claim.

**Caught my own regression before it shipped**: `LikeOperator.execute(DatabaseInternal database, ...)` and `ILikeOperator`'s equivalent are called with `database == null` directly in the existing `LikeOperatorTest`/`ILikeOperatorTest` (an established unit-test convention for these operators). My first pass at wiring in the config lookup would have NPE'd on exactly those tests. Applied the same null-safe `ContextConfiguration` fallback `TextRegexReplace` already needed.

New tests: `QueryHelperLikeRegexTimeoutTest` (utility-level `like()`/`likeUntil()` plus SQL-level `LIKE` and `ILIKE` regression tests through a real query), `FullTextQueryExecutorTest#catastrophicWildcardQueryIsAbortedByRegexTimeout` (careful to avoid a leading wildcard, since `allowLeadingWildcard` defaults to `false` and I didn't want the test coupled to that config). Reran `LikeOperatorTest`, `ILikeOperatorTest`, `QueryTest` (66), `SelectStatementExecutionTest` (159), `SelectOperatorsExtendedTest` (12), `SelectExecutionTest` (24) - all green, including the null-database convention.

At this point every entry point named across four review rounds - `MATCHES`, `=~`, `text.regexReplace`, full-text `RegexpQuery`, and now `LIKE`/`ILIKE`/full-text `WildcardQuery` - is bounded the same way, and the audit conclusion that was wrong is corrected rather than just left standing next to a patch. Open to another pass if there's more, but I believe this closes the class of bug the issue was about.

lvca:
Pushed 9adab87c addressing all four points:

**1. Inconsistent null-database fallback: investigated (already done in an earlier round) and documented rather than guarded.** Confirmed `MatchesCondition`/`RegexExpression` are only ever constructed by the SQL/openCypher parsers and evaluated by the query execution engine, which always binds a database to the context before any `WHERE`/`RETURN` expression runs — there's no direct-unit-test-style entry point into these classes the way there is for a standalone operator (`LikeOperatorTest` et al. construct and call `LikeOperator` directly) or SQL function. Added a comment at each call site explaining this instead of an unneeded guard, per your own suggested resolution.

**2. Deadline overflow: fixed, and it's a real one.** `newDeadline()` now uses `Math.addExact`/`multiplyExact`, falling back to the disabled (`Long.MAX_VALUE`) sentinel on overflow — the safe direction, since an admin fat-fingering a huge `arcadedb.command.regexTimeout` clearly meant "a lot of time," not "abort every match immediately," which is what the unguarded arithmetic would have silently produced. Added `newDeadlineDoesNotOverflowOnAnOversizedTimeout`.

**3. GC pressure**: no new action — already measured at ~30ns/match (noise-level) in an earlier round, consistent with your own assessment that it's "probably negligible relative to `Pattern.compile()`."

**4. Boyer-Moore/fast-path concern: documented.** You're right that this PR's own tests already provide decent evidence (the literal-heavy `%a%a%a...` and `a*a*a*...` patterns correctly abort, so `charAt()` is definitely still on the path for those shapes) — added that to `TimeBoundRegex`'s class Javadoc explicitly, so a future reader doesn't have to re-derive the confidence from test behavior.

All tests green: `TimeBoundRegexTest` (8, incl. the new overflow test), `MatchesConditionTest`, `TextRegexReplaceTest`, `FullTextQueryExecutorTest`, `QueryHelperLikeRegexTimeoutTest`, `LikeOperatorTest`, `ILikeOperatorTest`.

lvca:
Pushed c9d816b3 addressing everything from this round:

**PromQL `=~`/`!~` label matchers: fixed, and it's real.** Verified empirically before touching anything, same discipline as the LIKE finding: `a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*a*ac` (no parens, so `REDOS_CHECK` — which only flags parenthesized nested-quantifier/alternation shapes — never sees it) against a 40-char string of `a`s hangs exactly like every other reproducer in this issue. `matchesPostFilters()` now routes both `RE`/`NRE` cases through `TimeBoundRegex.matchesUntil()`, sharing one deadline across the whole row scan (it runs once per row, so this needed the same shared-deadline treatment as the multi-value `MATCHES`/full-text-scan cases, not a fresh budget per row). Kept `REDOS_CHECK` in place as a cheap static pre-filter alongside the new runtime bound — same layering `text.regexReplace`'s length cap got.

**Misleading message on disabled timeout + stack overflow: fixed.** `TimeBoundRegex.run()` now only converts `StackOverflowError` to `TimeoutException` when the bound is actually active; when `arcadedb.command.regexTimeout <= 0` it propagates as itself, since blaming a setting the caller deliberately disabled would be actively confusing. Didn't add an automated test for this specific branch — reliably triggering a genuine `StackOverflowError` from `java.util.regex` in a fast, portable, CI-safe way is its own small research project (JVM stack size varies, and a pattern deep enough to guarantee it risks being slow/flaky), and the change is a message-clarity fix, not a behavior change in whether the operation eventually completes or fails.

**Possible silent fallback via `BinaryCondition`'s pre-existing null-context tolerance: investigated, not reachable today.** Traced every call site in the engine that invokes `.evaluate(Result, CommandContext)` with a literal `null` context. Found exactly two, both openCypher-only and both narrow constant-folding cases (`ExpressionEvaluator#evaluatePropertyAccess` for plain property reads, `CypherASTBuilder`'s static-list-literal folding) — neither can reach `LikeOperator`/`ILikeOperator` (SQL-only, never touched by openCypher) or `RegexExpression`/`MatchesCondition`. No live SQL-engine caller passes `BinaryCondition.evaluate(Result, CommandContext)` a null context today. Even if that changes, the null-safe fallback already in place resolves to the compiled-in default rather than crashing, which is the same safe direction taken everywhere else in this PR.

**Pattern compilation per call in `QueryHelper.like`**: no action — pre-existing behavior (`String.matches()` never cached either), not a regression from this PR, and out of scope for a security fix.

New test: `PromQLEvaluatorIntegrationTest#unparenthesizedCatastrophicPatternIsAbortedByRegexTimeout`. All tests green: that one (17 total in the class now), `TimeBoundRegexTest` (8), `MatchesConditionTest`, `TextRegexReplaceTest`, `FullTextQueryExecutorTest`, `QueryHelperLikeRegexTimeoutTest`, `LikeOperatorTest`, `ILikeOperatorTest`.

At this point every regex-based query surface I can find in the codebase — `MATCHES`, `=~` (Cypher), `LIKE`/`ILIKE`, `text.regexReplace`, full-text `RegexpQuery`/`WildcardQuery`, and PromQL `=~`/`!~` — routes through the same `TimeBoundRegex` bound.

lvca:
Pushed bf3bfdbe addressing all four points:

**1. Inconsistent null-safety: fixed.** Agreed on the second ask — `MatchesCondition`/`RegexExpression` now fall back to a plain `ContextConfiguration()` when `context.getDatabase()` is null, matching the other three modified entry points. Investigated and confirmed not reachable twice now, but the guard costs nothing and removes the asymmetry, so not worth relying on the audit holding forever against every future refactor.

**2. Shared-deadline-over-full-scan + 1000ms default is aggressive for bulk paths: documented explicitly.** Added a standalone paragraph to the release notes calling this out as its own tradeoff (distinct from the earlier "any single match >1s now fails" note) — a legitimate, non-catastrophic full-text or PromQL scan over a lot of data can now fail on `TimeoutException` where it previously just ran slowly, and recommends raising `arcadedb.command.regexTimeout` for such workloads. This is a natural, unavoidable consequence of the shared-deadline fix being correct (per-item budgets reopen the bypass), not something further code can resolve — it's a real behavior change worth surfacing to whoever reads the release notes, not something to silently absorb.

**3. Test coverage gaps: both closed.**
- `SelectExecutionTest#catastrophicLikeIsAbortedByRegexTimeout` — the native query engine's `SelectOperator.like`/`ilike` (a third, independent LIKE/ILIKE path, distinct from the SQL parser's operators) now has its own regression test, matching every other modified entry point.
- `QueryHelperLikeRegexTimeoutTest#multiValueLikeSharesOneTimeoutBudgetAcrossItems` — mirrors `MatchesConditionTest`'s multi-value test for `LikeOperator`/`ILikeOperator`'s `BY ITEM` branch, which had the shared-deadline logic already but no test proving it.

**4. `QueryHelper.like`/`likeUntil` duplication: fixed.** `like()` now just delegates to `likeUntil(currentValue, value, TimeBoundRegex.newDeadline(timeoutMillis))`.

All tests green: `MatchesConditionTest`, `TimeBoundRegexTest`, `QueryHelperLikeRegexTimeoutTest` (6, incl. the new multi-value test), `SelectExecutionTest` (25, incl. the new catastrophic-LIKE test), `LikeOperatorTest`, `ILikeOperatorTest`, and the regex-related `OpenCypherWhereClauseTest` methods.

lvca:
Pushed 163fa729. The security-critical finding this round:

**`DocumentValidator`'s schema-level `REGEXP` property constraint: fixed, and it's the most exposed entry point yet.** `fieldValue.toString().matches(p.getRegexp())` ran on every insert/update of a validated property with zero bound. Unlike everything else in this issue, this needs **no query privileges at all** — any write path (REST, any wire protocol) into a validated type is enough, and admin-defined validation patterns (classic email/URL regexes especially) are a well-known source of accidentally-catastrophic shapes. Now routed through `TimeBoundRegex`. Added `DocumentValidationTest#catastrophicRegExpValidationIsAbortedByRegexTimeout`, and reran the full CRUD/transaction sweep (`CRUDTest`, `ACIDTransactionTest`, plus the config tests — 59 tests) given this sits on every document write; all green.

**Design question on per-row vs per-scan deadline: fixed for `MATCHES`/`=~`, documented as a conscious trade-off for `LIKE`/`ILIKE`.** You're right that the multi-value fix didn't close the row-by-row gap — a table shaped so every row triggers catastrophic backtracking could still cost `rowCount * regexTimeout`. `MatchesCondition` now caches one deadline on the `CommandContext` for the query's lifetime, reusing the exact mechanism already proven (by `MatchesConditionTest`'s own per-context cache tests) to cache the compiled `Pattern` across rows. `RegexExpression` caches it as an AST-node instance field, mirroring its existing `compiledPattern` caching. `LikeOperator`/`ILikeOperator` couldn't get the same treatment: `BinaryCompareOperator.execute()` only receives a `Database`, no `CommandContext` to cache on, and widening that interface across every comparison operator (`EqualsCompareOperator`, `GtOperator`, `InOperator`, ...) for this one operator pair felt out of proportion — documented explicitly in `LikeOperator` as a conscious, narrower-scope choice rather than left as an unexplained inconsistency. Added `MatchesConditionTest#matchesSharesOneTimeoutBudgetAcrossAllRowsInTheScan` and `OpenCypherWhereClauseTest#regexSharesOneTimeoutBudgetAcrossAllRowsInTheScan` (10 separate pathological rows each, not the multi-value-in-one-row case already covered).

**Repeated boilerplate: fixed.** Added `GlobalConfiguration.getValueAsLong(Database)` — resolves a `SCOPE.DATABASE` setting against a possibly-null `Database`, falling back to the compiled-in default — and switched all nine call sites (the five from before, `FullTextQueryExecutor` x2, `PromQLEvaluator` x2, and the new `DocumentValidator` site) onto it, replacing the duplicated null-check-then-`getConfiguration()`-then-`getValueAsLong()` chain everywhere.

**1000ms default being aggressive for full-scan paths**: no new action, already documented explicitly in the prior round's release-notes addition — this comment is restating that concern, not raising a new one.

All tests green: `MatchesConditionTest` (8), `DocumentValidationTest` (23), `TimeBoundRegexTest`, `TextRegexReplaceTest`, `FullTextQueryExecutorTest`, `QueryHelperLikeRegexTimeoutTest`, `LikeOperatorTest`, `ILikeOperatorTest`, `PromQLEvaluatorIntegrationTest` (17), `SelectExecutionTest` (25), the regex-related `OpenCypherWhereClauseTest` methods (14), plus the broader CRUD/config/transaction sweep above.

lvca:
Pushed 825596b0. This was a genuine catch that would have shipped a real regression — thank you for the thorough investigation into `CypherStatementCache`.

**🔴 Critical: fixed, and confirmed the bug was real before fixing it.** You're completely right: I mirrored `compiledPattern`'s existing instance-field caching for `deadline` without checking whether that caching pattern's safety assumption (content-addressed, never stale) actually transfers to a deadline (a point in time, goes stale by definition). It doesn't. I reverted just that change, ran the exact regression test I was about to add, and confirmed it fails with precisely the predicted spurious `TimeoutException` on a completely benign match — then restored the fix (cache on `context.getCachedValue`/`setCachedValue`, same mechanism `MatchesCondition` already used correctly) and confirmed the same test passes. Added `regexDeadlineDoesNotLeakAcrossCachedQueryExecutions` using your suggested repro shape almost verbatim (execute once, sleep past the deadline, execute the identical query text again, assert the second execution still succeeds).

**🟠 Moderate (`TextRegexReplace` losing `StackOverflowError` handling when disabled): fixed.** Added a direct `catch (StackOverflowError)` back in `TextRegexReplace` itself, restoring the original `IllegalArgumentException` contract for that specific combination (bound disabled + stack-overflow-inducing pattern) regardless of what `TimeBoundRegex` does internally.

**🟡 `DocumentValidator`'s default-1000ms tradeoff: documented explicitly**, appended to the same release-notes paragraph that introduced the fix, called out as its own point given it's the one entry point needing no query privileges at all.

**🟢 GC/performance note**: no new action — already measured negligible overhead in earlier rounds (~30ns/match either way), consistent with your own read that `CHECK_INTERVAL` keeps `nanoTime()` calls rare enough not to matter.

All tests green: the full regex-related `OpenCypherWhereClauseTest` suite (15, incl. the new critical-regression test), `TextRegexReplaceTest`, `DocumentValidationTest`, `MatchesConditionTest`, plus a broader Cypher sweep (`CypherBuiltInFunctionsTest` 145, `CypherFunctionSecurityTest` 28, `CypherUncorrelatedSubqueryCountPushDownIssue5686Test` 14) given `RegexExpression` is a core, widely-cached AST node touched by this fix.

lvca:
Pushed 33426c1e. Another genuine catch — thank you for the careful cache-key analysis.

**Bug (100% reproducible cache-key collision): fixed, and verified the same rigorous way as the last one.** `MatchesCondition`'s deadline key `"MATCHES_DEADLINE"` collided exactly with the pattern-cache key for literal pattern text `DEADLINE` (`"MATCHES_" + "DEADLINE"`). I reverted just the key change, ran the new regression test, and confirmed it reproduces your exact predicted `ClassCastException` (`Pattern cannot be cast to Long`) on the first row — then restored the fix. The new key (`"5886_MATCHES_CONTEXT_DEADLINE"`) lives entirely outside the `"MATCHES_"` prefix namespace the pattern cache uses, so it's not just unlikely to collide with `"MATCHES_" + <any regex text>` — it's structurally impossible, since string equality requires matching that literal 8-character prefix first. Added `patternTextDeadlineDoesNotCollideWithTheDeadlineCacheKey`.

**`SQLMethodSplit` (`.split()` method syntax): fixed.** You correctly identified it as the outlier — its three siblings (`split()` function, `text.split()`, Cypher `split()`) all already wrap the delimiter in `Pattern.quote(...)`, treating it as a literal by design. Matched that, which is a stronger fix than bounding: a quoted literal can never form a catastrophic shape at all, not just a bounded one. Added `delimiterIsTreatedLiterallyNotAsRegex`, which also confirms this wasn't just a security gap but a correctness one (a `.` delimiter was splitting on every character instead of the literal dot).

**`SQLMethodNormalize` (`.normalize()` method's pattern argument): fixed via `TimeBoundRegex`.** Unlike `.split()`, this one is *supposed* to accept a real regex (it controls what gets stripped post-normalization), so literalizing it would have broken intended functionality — bounded it the same way as every other genuine regex entry point instead. Added `catastrophicPatternArgumentIsAbortedByRegexTimeout`.

**Stale config description: fixed.** `COMMAND_REGEX_TIMEOUT`'s description now lists everything it actually covers (it had drifted since the original two-entry-point description from round 1) and explains the per-query vs. per-scan sharing distinction.

**LIKE/ILIKE residual per-row budget & broad `StackOverflowError` catch**: no new action — both already reviewed/documented as conscious, acceptable trade-offs in earlier rounds, and this review's framing ("worth confirming... probably fine in practice") reads as re-confirming rather than raising something new.

All tests green: `MatchesConditionTest` (9), `SQLMethodSplitTest`, `SQLMethodNormalizeTest`, plus a broader sweep (`SQLMethodAdditionalCoverageTest` 64, `SQLFunctionsTest` 32, `QueryTest` 66) given SQL method dispatch is a widely-shared path.

At this point two independent review rounds have each caught a real, verifiable bug in the fix itself (not just missed entry points) — I've verified both against their actual failure modes before and after the fix, not just read the code and assumed correctness. I believe the fix is now solid, but given the track record I'd rather have one more look than assume it's done.

lvca:
Pushed 4cab15f9. This round came back clean — "No correctness bugs or security gaps found" — with three minor, non-blocking points, all addressed:

**1. `SelectOperator` bypassing the new `getValueAsLong(Database)` helper: fixed.** Switched both `like`/`ilike` cases to it. Functionally identical either way (`record.getDatabase()` is never null there), but the helper exists specifically to keep this lookup from drifting across call sites, so it should be used everywhere, not everywhere-but-one.

**2. LIKE/ILIKE residual per-row budget & broad `StackOverflowError` catch**: no new action — this round's framing reads as re-confirming both as acceptable, already-documented trade-offs from earlier rounds rather than raising something new.

**3. Release notes reading as an internal changelog: rewritten.** Fair nit — ten review rounds had left the notes narrating "a 6th review pass found...", "a 7th review pass caught...", which is exactly the kind of process detail that belongs in this PR thread, not in what ships as user-facing documentation. Rewrote the whole section as a single, direct description of the final state: what's covered, how the deadline-sharing design works (including the one documented exception for `LIKE`/`ILIKE`), and the upgrade-behavior tradeoff — with none of the round-by-round narration. All the substantive technical content survived the rewrite, just not the "which review pass found it" framing.

All tests green: `SelectExecutionTest` (25), `SelectOperatorsExtendedTest` (12).

Ten review rounds, two of which caught genuine, independently-verified bugs in the fix itself (not just missed entry points), and this one found none. I think this is a solid stopping point unless something new surfaces.

lvca:
Pushed 989392fb. Thanks for the thorough final pass — this is a great summary to close on.

**1. Inconsistent exception contract (`TextRegexReplace` rewraps, `.normalize()` doesn't): documented, not unified.** Added a code comment explaining the reasoning rather than changing behavior: `TextRegexReplace` rewraps to `IllegalArgumentException` because it already had that contract *before* this issue (it already converted `PatternSyntaxException` the same way, for invalid patterns). `.normalize()`'s pattern argument had no such prior contract — letting `TimeoutException` propagate as-is is actually the *more* consistent choice across this PR as a whole, since that's the shape `MatchesCondition`, `RegexExpression`, `LIKE`/`ILIKE`, full-text, and PromQL all surface. Unifying in the other direction (making `TextRegexReplace` also throw raw `TimeoutException`) would be a breaking change to an existing, pre-issue contract for a function that predates this PR, which felt like the wrong tradeoff for a security fix.

**2. `LIKE`/`ILIKE` per-row budget as a tracked follow-up issue**: agreed this is worth tracking separately rather than living only as a code comment, given `LIKE` is likely the more commonly used operator. Deliberately not filing a new GitHub issue myself here — that's a call for @lvca as maintainer, and it's clearly framed as a follow-up rather than something blocking this PR.

**3. Upgrade-behavior callout prominence**: looked at the document structure — there's no separate top-level "highlights" list this section reports into; each `##` heading is its own self-contained unit, and the "Upgrade note" is already a clearly-labeled, bolded paragraph within it (same pattern the rest of this release-notes file uses for callouts). Didn't see anywhere else in the doc to duplicate it into without inventing a mechanism that doesn't otherwise exist here.

**4. `DocumentValidator` not caching the compiled `Pattern` per-`Property`**: agreed this is a good opportunistic follow-up, but declined for this PR specifically — it's explicitly pre-existing behavior (not a regression from this fix), and expanding scope into a performance optimization unrelated to the security fix risks the kind of scope creep this project's own guidelines caution against. Worth its own follow-up if it matters in practice.

All tests green (`SQLMethodNormalizeTest` unaffected by the doc-only change).

This round confirmed no bugs and no blockers — just refinements. Across 11 rounds: 2 real, independently-verified bugs in the fix itself (a `RegexExpression` stale-deadline regression and a `MatchesCondition` cache-key collision, both caught and fixed before merge), and the regex-entry-point audit grew from the 2 named in the original issue to 11 (`MATCHES`, `=~`, `LIKE`/`ILIKE` across 3 independent evaluation paths, `text.regexReplace()`, `.normalize()`, `.split()`, full-text `RegexpQuery`/`WildcardQuery`, PromQL `=~`/`!~`, and schema `REGEXP` validation). I believe this is ready to merge.

PR: **https://github.com/ArcadeData/arcadedb/pull/5983**

lvca:
Pushed 8e3a0f5c. All three points were real, valid instances of the exact principle this PR establishes elsewhere — fixed all of them.

**1. `DocumentValidator` not sharing a deadline across a document's properties: fixed.** This is the entry point the PR itself calls "most exposed" (no query privileges needed), so this gap mattered more here than anywhere else. `validate()` now computes one deadline and threads it through `validateField()` via a new 3-arg overload; the existing 2-arg `validateField()` (no evidence of external callers, but it's public API) keeps its own per-call deadline for standalone single-field use. Corrected the release note's claim that REGEXP validation already shared a deadline — it didn't until this commit, now it does.

**2. PromQL `evaluateRange()` recomputing the deadline per step: fixed.** You're right this was a real gap, not just theoretical — at the 1000ms default, `MAX_RANGE_STEPS` (1,000,000) times over is genuinely ~11.5 days worst case. Fixed by caching the deadline as a `PromQLEvaluator` instance field, computed once lazily. Confirmed this is safe (unlike the earlier `RegexExpression` AST-instance-field mistake) by checking `PromQLEvaluator`'s only construction site (`SQLFunctionPromQL`) — a fresh evaluator is created per top-level query, never cached/reused across executions the way `RegexExpression`'s AST nodes are by `CypherStatementCache`. That distinction is exactly what made an instance field wrong there and right here.

**3. `TextRegexReplace`/`.normalize()` having a `CommandContext` but not using it: fixed.** Both now share one deadline across a query's rows via `context.getCachedValue`/`setCachedValue`, the same pattern `MatchesCondition`/`RegexExpression` already use. Added `TimeBoundRegex.replaceAllUntil()` (the deadline-based counterpart to `replaceAll()`, mirroring `matches()`/`matchesUntil()`) to support it.

**Minor cache-key naming: fixed too**, since I was touching that file anyway — `MatchesCondition`'s deadline key is now `"__MATCHES_DEADLINE__"` instead of embedding the issue number, per your earlier suggestion.

New tests for all four: `DocumentValidationTest#regExpValidationSharesOneTimeoutBudgetAcrossAllPropertiesOnOneDocument`, `PromQLEvaluatorIntegrationTest#rangeQuerySharesOneTimeoutBudgetAcrossAllSteps`, `TextRegexReplaceTest#multiRowReplaceSharesOneTimeoutBudgetAcrossRows`, `SQLMethodAdditionalCoverageTest#normalizeMultiRowSharesOneTimeoutBudgetAcrossRows`, `TimeBoundRegexTest#replaceAllUntilSharesOneDeadlineAcrossASeries`. All follow the same "N independent budgets would exceed X, a shared deadline keeps it under X" proof pattern as the earlier multi-row/multi-value tests. Full affected suite (135 tests) plus a broader sweep (`CRUDTest`, `SQLFunctionsTest`, `QueryTest`) all green.

This was a genuinely valuable round — three real architectural gaps in the core principle this PR is built around, all now closed consistently across every entry point.

lvca:
Addressed the duplication point from the last review round: `MatchesCondition`, `RegexExpression`, `TextRegexReplace`, and `SQLMethodNormalize` all hand-rolled the same get-cached-or-compute-and-cache-deadline pattern with their own literal cache key - exactly the shape that produced the earlier cache-key-collision bug. Extracted it into a shared `CommandContext.getOrComputeRegexDeadline(String cacheKey)` default method (and a `(cacheKey, timeoutMillis)` overload for the no-CommandContext-available null-safe fallback paths), and migrated all four call sites onto it. Each call site still owns its own collision-free key, but no longer hand-writes the get-or-compute logic.

Compiled clean and re-ran the full affected suite (`TimeBoundRegexTest`, `MatchesConditionTest`, `OpenCypherWhereClauseTest` - including the stale-deadline regression test - `TextRegexReplaceTest`, `SQLMethodNormalizeTest`, `SQLMethodAdditionalCoverageTest`, `SQLMethodSplitTest`, `DocumentValidationTest`, `PromQLEvaluatorIntegrationTest`, `FullTextQueryExecutorTest`, `QueryHelperLikeRegexTimeoutTest`, `SelectExecutionTest`, `LikeOperatorTest`, `ILikeOperatorTest`): all green, no regressions.

On the other reconfirmed-but-not-actioned points from the last round: no code change intended.
- Default `arcadedb.command.regexTimeout` of 1000ms was an explicit decision, not an oversight.
- The per-context allocation overhead of the shared-deadline caching was already measured negligible relative to a regex evaluation itself.
- `TimeBoundRegex` converting a `StackOverflowError` into a `TimeoutException` (only when the bound is active) is a deliberate, narrowly-scoped choice: that class exists specifically to bound catastrophic-backtracking cost, and a stack overflow under an active deadline is itself evidence of the same catastrophic-backtracking class of pattern, so surfacing it as the same `TimeoutException` the caller already handles is more useful than a raw `StackOverflowError` escaping.

lvca:
Addressed the two easy, actionable points from the last round; the other two are genuinely non-blocking nits with no further action needed.

**1. `DocumentValidator.validate()` unconditionally computing a deadline: fixed.** Now computed lazily, only on the first REGEXP-constrained property actually encountered, mirroring the lazy `PromQLEvaluator.regexDeadline()` field pattern already used elsewhere in this PR. Document types with no REGEXP-constrained properties (the common case) no longer pay `System.nanoTime()` + the overflow-safe arithmetic on every insert/update. `DocumentValidationTest` (24 tests) still green.

**4. Stale PR description: fixed.** Rewrote it to describe the actual final scope (all covered entry points, deadline sharing, the two bugs caught during review) instead of the original narrow MATCHES/`=~`-only description that predated the audit growing to 11 entry points.

**2. `StackOverflowError` catching risk**: no change, per the review's own assessment - "narrowly scoped... I'd call this an acceptable, pre-existing-pattern tradeoff rather than a new problem." Agreed, and this exact tradeoff was reconfirmed in the prior round's response too.

**3. Upgrade-note double-check**: no code change - this is a request to verify behavior in practice before release, not a code or doc gap. The upgrade note itself is already accurate and prominent (bolded paragraph in `docs/release-26.9.1.md`).

Compiled clean, `DocumentValidationTest` re-run green, pushed as 9ed06084e.

lvca:
Verified the parallel-bucket-scan nuance from the last round rigorously before deciding how to respond - traced it end to end:

- `FetchFromTypeExecutionStep.syncPullParallel` calls `context.copy()` once per bucket/worker (`engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromTypeExecutionStep.java:282`).
- `BasicCommandContext.copy()` does **not** carry `cachedValues` over to the copy - it's left `null` on the new instance (`engine/src/main/java/com/arcadedb/query/sql/executor/BasicCommandContext.java:401-415`).

So the claim is correct: each parallel worker's `getOrComputeRegexDeadline()` computes its own deadline from scratch, independent of every other worker. A type scanned across N buckets in parallel is bounded by `N * regexTimeout` overall, not one shared budget the way the sequential scan path is.

**Not fixing the sharing itself.** I looked at actually sharing one deadline across workers and concluded it would trade a narrow, non-attacker-controlled gap for a real bug: `cachedValues` is a plain `HashMap`, not thread-safe, and `getOrComputeRegexDeadline`'s check-then-compute-then-set has no synchronization. Aliasing that same map across N worker threads (the way `inputParameters` is deliberately aliased, since it's read-only during execution) would let two workers race on populating the same key on first use - a real concurrency bug, for a gap the review itself already assessed as "not a security hole" and "much smaller... bucket count is bounded and admin-controlled, not attacker-controlled." Trading a documented, bounded gap for an actual race condition isn't a good trade.

**Fixed instead**: corrected the "shares one deadline across an entire query execution" claim - both the `CommandContext.getOrComputeRegexDeadline()` Javadoc and the release notes' "Deadline sharing" section now explicitly caveat the parallel-scan case, with the reasoning above for why it's accepted as-is rather than fixed.

Compiled clean; this is a doc/Javadoc-only change with no test impact. Pushed as 96d3376c5.

lvca:
Verified the `promql()` finding rigorously before deciding how to respond, and want to be transparent about what I actually found (it's more nuanced than the review framing).

**The code characteristic is real**: `SQLFunctionPromQL.execute()` does construct `new PromQLEvaluator(database)` on every call, with no wiring to the `CommandContext`-cached deadline. Confirmed.

**The specific claimed consequence didn't hold up empirically.** I reproduced the exact scenario described (`SELECT promql(patternColumn) FROM BigTypeWithManyRows`, 10 rows, all with the same catastrophic label matcher) and measured: **220ms elapsed, not the ~2000ms** ten independent 200ms budgets would take. Reason: the SQL engine aborts the whole query on the *first* row's `TimeoutException` rather than continuing to evaluate subsequent rows - it doesn't catch-and-continue past it. So the real-world cost is bounded to ~1 `regexTimeout` (plus whatever fast, non-catastrophic work preceded it) regardless of row count, the same "abort on first throw" characteristic that already applies to `WHERE ... MATCHES` scans and `DocumentValidator`'s per-property loop.

**Fixed anyway, for a narrower, real reason.** Rather than leave this as a documented non-issue, I added `PromQLEvaluator.setRegexDeadline()` and wired `SQLFunctionPromQL` to resolve the deadline from `context.getOrComputeRegexDeadline()`, same as `text.regexReplace()`/`.normalize()`. This closes the actual inconsistency (every other per-row SQL function with a `CommandContext` shares the deadline; `promql()` was the one exception) and removes any dependency on the specific abort-on-first-exception behavior continuing to hold in future refactors, at effectively zero cost.

**Test added honestly.** I didn't reuse the "many identical rows" pattern from other tests in this PR, since I'd just shown it doesn't discriminate a regression here. Instead: `setRegexDeadlineOverridesTheLazilyComputedDefault` proves the injected deadline is actually honored (using an already-expired deadline to force an immediate abort on an otherwise-instant, non-catastrophic match), and `promQLSqlFunctionAbortsOnCatastrophicPattern` adds missing end-to-end coverage for the `promql()` SQL-function path specifically (previously only `PromQLEvaluator.evaluateInstant()` called directly was covered for the catastrophic-pattern case).

Full `PromQLEvaluatorIntegrationTest` suite (20 tests, including the 2 new ones) green. Pushed as 912485509.

lvca:
This round re-confirms three trade-offs already settled in earlier rounds rather than raising new ones - no code changes needed, restating why for the record:

**1. Hot-path allocation from `TimeBoundRegex.run()`**: reconfirmed as an accepted trade-off (round 13's response: "the per-context allocation overhead of the shared-deadline caching was already measured negligible relative to a regex evaluation itself"). A `DeadlineBoundCharSequence` + `int[1]` per call is a small, bounded cost against closing an unbounded-hang vulnerability.

**2. `LIKE`/`ILIKE` per-row `System.nanoTime()`**: this is the documented, deliberate exception to deadline-sharing (`BinaryCompareOperator` has no `CommandContext` to cache a scan-wide deadline on - widening that interface was judged out of proportion). Already called out explicitly in `docs/release-26.9.1.md`'s "Deadline sharing" section.

**3. `TextRegexReplace` vs `SQLMethodNormalize` exception-type inconsistency**: addressed directly in round 15's response - `TextRegexReplace` rewraps to `IllegalArgumentException` because it already had that contract *before* this issue; `.normalize()` had no such prior contract, so letting `TimeoutException` propagate is actually the more consistent choice relative to the rest of this PR (`MatchesCondition`, `RegexExpression`, `LIKE`/`ILIKE`, full-text, PromQL all surface `TimeoutException` directly). Unifying the other direction would be a breaking change to `TextRegexReplace`'s pre-existing, documented contract.

The review's own conclusion: "solid and mergeable... not blockers." Agreed - no further action this round.

lvca:
\`unit-tests\` failed on the last CI run (2 failures: \`Issue4960LSMCloseDuringCompactionTest#closeCancelsScheduledCompaction\` and \`DictionaryMultiPageTest#aDictionaryPageThatIsClaimedButMissingFailsLoudly\`). Investigated rigorously before treating as a real regression:

- Neither test touches any file this PR modifies (both are in unrelated LSM index / dictionary-page subsystems - nothing to do with regex, \`CommandContext\`, PromQL, or \`DocumentValidator\`).
- Both failure symptoms are classic CI-resource-contention artifacts: \`Issue4960LSMCloseDuringCompactionTest\` failed a post-test database-integrity check with WAL-preservation warnings during a close/compaction race; \`DictionaryMultiPageTest\` failed to observe an expected exception during concurrent page access - both timing-sensitive under a loaded runner.
- Ran both locally on this exact branch (with all of this PR's changes applied): both pass cleanly, 15/15 tests green.

Conclusion: pre-existing CI flakes unrelated to this PR, not a regression. Re-running the failed job once the run finishes (can't re-run failed jobs while the workflow run is still in progress).

lvca:
\`ha-integration-tests\` failed a second consecutive time on retry, same two tests as before: \`Issue5410AbandonedTicketReleaseIT#abandonedEntryReleasesItsPhase2TicketOnceApplied\` and \`Issue5569SlotMergeDeleteRaftIT#mergedDeletesReplicateIntact\`. Per the instruction to treat a repeat failure more seriously than a first one, compared both runs in detail rather than just retrying again blind:

| | Run 1 | Run 2 |
|---|---|---|
| \`Issue5410...\` assertion | "Node **0** must hold both vertices" | "Node **2** must hold both vertices" |
| \`Issue5569...\` assertion | "Update returned null: record=**4** seq=**25**" | "Update returned null: record=**7** seq=**40**" |

Same test identities fail both times, but the specific node index and record/seq numbers differ between runs. That pattern is what a genuine timing/replication race looks like (non-deterministic which node/record loses the race), not what a deterministic bug from this PR's diff would produce (which would fail the same way every time, or not fail at all if truly unrelated). Both tests are in \`arcadedb-ha-raft\` - Raft leader-election, ticket-release, and slot-merge-delete replication internals - with zero file overlap with this PR's changes (regex/\`CommandContext\`/PromQL/\`DocumentValidator\`). This also matches project history: \`ha-integration-tests\` is a documented chronic flake independent of PR content (tracked separately, running red a large fraction of the time regardless of what's being merged).

Retrying once more. If it fails a third time with the same non-deterministic pattern, I'll treat that as sufficient confirmation of the known pre-existing chronic flake rather than continuing to retry indefinitely, and will flag that explicitly rather than silently blocking on it.

lvca:
\`ha-integration-tests\` has now failed 3 consecutive times. Summary of the investigation across all 3 attempts before stopping the retry loop, per plan:

**Attempts 1 & 2**: same two tests failed both times - \`Issue5410AbandonedTicketReleaseIT#abandonedEntryReleasesItsPhase2TicketOnceApplied\` and \`Issue5569SlotMergeDeleteRaftIT#mergedDeletesReplicateIntact\` - with different specific node indices and record/seq numbers each run (Node 0/record=4/seq=25, then Node 2/record=7/seq=40), the signature of a non-deterministic timing/replication race, not a deterministic bug.

**Attempt 3**: the same two tests failed again (Node 2/record=8/seq=17 this time - again different data), **plus a third, different test** failed: \`Issue4743LargeRecordReplicationIT#recordLargerThanTheOldCeilingReplicatesToEveryNode\`, with \`Schema Cannot create type 'Blob' because already exists\` - a test-isolation/schema-state-pollution error, a different failure class entirely from the other two.

All three failing tests are in \`arcadedb-ha-raft\` (Raft leader-election, ticket-release, slot-merge-delete, and large-record replication internals) - **zero file overlap** with this PR's changes (regex/\`CommandContext\`/PromQL/\`DocumentValidator\`). The variety of failure types and non-deterministic data across 3 independent runs points to general instability of this test suite/CI runner rather than a regression introduced by this PR. This also matches this suite's documented status as a chronic flake independent of PR content.

Not retrying a 4th time. Every other check on this PR is green, including \`unit-tests\` (which failed once and passed clean on retry) and the code review itself (round 18: "solid and mergeable... not blockers"). Bringing the merge decision to @lvca given \`ha-integration-tests\` is a required check that hasn't gone green across 3 attempts despite being unrelated to this PR's diff.

PR #5992, opened 2026-08-09
fix(#5781): cover and correct the TypeIndexBuilder half of the guarded-create METADATA comparison

### LINKED ISSUE #5781: The TypeIndexBuilder half of the guarded-create METADATA comparison (#5765) has no test coverage
Coverage gap in my own change on PR #5775 (#5765, item 3), found while writing up the follow-ups rather than during review. Verified against `main` at 0b2ccfb72.

## The comparison has two entry points; the tests only reach one

`CREATE INDEX IF NOT EXISTS ... METADATA {...}` now compares the settings the clause NAMED against the existing index. That question is asked in two places, both routed through the shared `IndexBuilder.findUnsatisfiedSettings`:

1. the `existsIndex(name)` shortcut in `CreateIndexStatement.executeDDL`, which answers before any builder exists;
2. `TypeIndexBuilder.create()`, for a request that gets past the shortcut.

`Issue5765IndexSettingsIfNotExistsTest` covers **only the first**. Every case in it is written in the auto-derived-name form:

```sql
CREATE INDEX IF NOT EXISTS ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 768}
```

With no explicit name, `CreateIndexStatement` derives `name` as `Doc[embedding]`, `existsIndex` is true, and the statement returns from the shortcut. It never reaches `TypeIndexBuilder.create()`.

## The uncovered path is reachable

The builder half fires when the requested NAME does not exist but an index on the same properties does - i.e. a **manually named** guarded create over an auto-named existing index:

```sql
CREATE INDEX ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 384};
-- creates Doc[embedding]

CREATE INDEX myVectorIdx IF NOT EXISTS ON Doc (embedding) LSM_VECTOR METADATA {"dimensions": 768};
-- existsIndex("myVectorIdx") is false -> falls through to TypeIndexBuilder.create(),
-- which finds the existing index by properties and compares the named settings
```

That second statement should be refused naming `dimensions`. I believe it is - the code is shared with the covered path and the surrounding branch is exercised by `Issue5675IndexIfNotExistsConflictTest` - but "I believe it is" is exactly the state this comparison exists to remove, and it is not asserted anywhere.

## What to add

A case in `Issue5765IndexSettingsIfNotExistsTest` in the named form, asserting both directions:

- named guarded create with a **differing** named setting over an existing auto-named index -> refused, message names the setting;
- named guarded create with **matching** named settings -> no-op, `created: false`, and the existing index untouched.

Mutation-check it: removing the `findUnsatisfiedSettings` call from `TypeIndexBuilder.create()` must fail the new test and leave the existing eight passing. If it does not fail, the path is unreachable and the call there is dead code, which is worth knowing on its own.

Low severity - the shipped behaviour is covered on the path users reach by default, and the shared helper means both call sites run the same comparison. It is the assertion that is missing, not (as far as I can tell) the behaviour.

robfrank:
Closes #5781

## The gap

`CREATE INDEX IF NOT EXISTS ... METADATA {...}` compares the settings the clause *named* against the existing index, and that question is asked from two places, both routed through the shared `IndexBuilder.findUnsatisfiedSettings`:

1. the `existsIndex(name)` shortcut in `CreateIndexStatement.executeDDL`, which answers before any builder exists;
2. `TypeIndexBuilder.create()`, for a request that gets past the shortcut.

Every case in `Issue5765IndexSettingsIfNotExistsTest` was written in the auto-derived-name form (`CREATE INDEX IF NOT EXISTS ON Doc (embedding) ...`), so `name` resolved to `Doc[embedding]`, `existsIndex` was true, and the statement returned from the shortcut. The builder half had no assertion on it.

The way past the shortcut is a **manually named** guarded create over an auto-named existing index: `myVectorIdx` does not exist, so the statement builds a builder, and the builder finds the existing index *by properties* and compares the named settings.

## What this adds

Four cases in `Issue5765IndexSettingsIfNotExistsTest`, all in the named form, asserting both directions on both a vector index (`dimensions`) and a full-text one (`analyzer`), so the coverage is not vector-specific:

- `namedGuardedVectorDimensionsMismatchIsReportedByTheBuilder`
- `namedGuardedFullTextAnalyzerMismatchIsReportedByTheBuilder`
- `namedGuardedMatchingSettingsStayANoOp`
- `namedGuardedMatchingFullTextSettingsStayANoOp`

**Mutation check** (requested in the issue): removing the `findUnsatisfiedSettings` call from `TypeIndexBuilder.create()` fails **exactly** the two new mismatch tests (`namedGuardedVectorDimensionsMismatchIsReportedByTheBuilder`, `namedGuardedFullTextAnalyzerMismatchIsReportedByTheBuilder`) and leaves the original eight passing. The builder path is reachable and that call is not dead code, as the issue predicted.

## The defect the matching direction found

The issue expected "it is the assertion that is missing, not the behaviour". That holds for the *mismatch* direction, which passed unchanged. The *matching* direction did not.

When `TypeIndexBuilder.create()` answers a guarded request with the index already on those properties, it returns that existing index. `CreateIndexStatement` then fell through to its unconditional tail and reported:

```
name: myVectorIdx, created: true, totalIndexed: 0
```

Nothing was created, and nothing named `myVectorIdx` exists. A caller reading `created: true` as "my name is now on an index" gets a later `DROP INDEX myVectorIdx` failing on a name that never existed. Same shape as #5675 one level down: a guarded statement reporting success for something it did not do.

The vector branch additionally ran `buildVectorGraphNow()` over the returned index, rebuilding the HNSW graph of an index the statement was told to leave alone.

Both are now keyed off whether `create()` answered with a new index. `isNewlyCreated` compares the returned index's name against the requested one: an index this statement created always carries the name asked for (the manual one when written, otherwise the auto-derived `typeName[properties]` form that `LocalDocumentType.addIndexInternal` derives), while the existing index the builder hands back carries its own. The no-op now reports `created: false` under the name of the index that actually satisfied the request.

Each branch of the create chain now captures its `create()` return value, which is what makes the single decision at the tail possible; the `else` fallthrough and the sparse/geo/full-text branches previously discarded it.

Note this also corrects the inherited-index case reached through the auto-derived name: `CREATE INDEX IF NOT EXISTS ON Child (embedding)` where only `Parent` carries the index used to report `created: true` under `Child[embedding]`, and now reports `created: false` under `Parent[embedding]`.

## Test plan

- [x] `mvn -pl engine test -Dtest=Issue5765IndexSettingsIfNotExistsTest` - 12 pass (8 pre-existing + 4 new)
- [x] Mutation check: `findUnsatisfiedSettings` removed from `TypeIndexBuilder.create()` fails exactly the 2 new mismatch tests, the original 8 stay green; call restored afterwards
- [x] Targeted regression run, 264 tests, all green: `Issue5675IndexIfNotExistsConflictTest`, `PolymorphicIndexConflictTest`, `IndexSyntaxTest`, `LSMTreeIndexTest` (incl. `Issue4139ManualIndexName`), `MapIndexByKeyValueTest`, `CreateIndexStatementTestParserTest`, `IndexMetadataUserSettingComparisonTest`, `Issue5723CopyTypeIndexConfigurationTest`, `Issue5713HashIndexPageSizeTest`, `Issue5639IndexMetadataKeysTest`, `Issue5607VectorIndexMetadataTest`, `RebuildIndexPreservesNameTest`, `LSMVectorIndexInheritanceTest`, `VectorIndexBuildGraphOnCreateTest`, `OpenCypherIndexTest`, `OpenCypherConstraintTest`, `LSMVectorIndexTest`, `LSMTreeGeoIndexSchemaTest`, `FullTextAnalyzerConfigTest`, `LSMSparseVectorIndexTest`, `SchemaTest`, `DDLTest`
- [x] Server module, the HTTP consumers of the `created` field: `Issue5675CreateIndexIfNotExistsHttpTest`, `ExecutionResponseTest` - 12 pass

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/CreateIndexStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/CreateIndexStatement.java
@@ -254,6 +254,11 @@ public class CreateIndexStatement extends DDLStatement {
 
     final AtomicLong total = new AtomicLong();
 
+    // What create() answered with: the index this statement built, or - for a guarded request whose NAME did not
+    // exist but whose properties are already indexed - the existing index the builder decided satisfies it. The two
+    // are told apart by the name, see {@link #isNewlyCreated}.
+    final TypeIndex resultingIndex;
+
     // Use unified buildTypeIndex() API for all index types
     TypeIndexBuilder builder = database.getSchema().buildTypeIndex(typeName.getStringValue(), fields);
     builder = builder.withType(indexType);  // This may return LSMVectorIndexBuilder for LSM_VECTOR
@@ -324,9 +329,12 @@ public class CreateIndexStatement extends DDLStatement {
       }
 
       final TypeIndex typeIndex = vectorBuilder.create();
+      resultingIndex = typeIndex;
 
-      // Build the HNSW graph immediately unless explicitly disabled
-      if (buildGraphNow)
+      // Build the HNSW graph immediately unless explicitly disabled - and only for an index this statement actually
+      // created. When create() answered a guarded request with the index already on those properties (issue #5781),
+      // rebuilding its graph is work nobody asked for on an index the statement was told to leave alone.
+      if (buildGraphNow && isNewlyCreated(typeIndex, name.getValue()))
         for (final Index idx : typeIndex.getIndexesOnBuckets())
           if (idx instanceof LSMVectorIndex)
             ((LSMVectorIndex) idx).buildVectorGraphNow();
@@ -345,7 +353,7 @@ public class CreateIndexStatement extends DDLStatement {
         // statement, so an unknown key or an out-of-range BM25 parameter is a client mistake and must answer 400.
         throw new CommandSQLParsingException("Invalid METADATA for FULL_TEXT index: " + e.getMessage(), e);
       }
-      ftBuilder.create();
+      resultingIndex = ftBuilder.create();
 
     } else if (indexType == Schema.INDEX_TYPE.LSM_SPARSE_VECTOR) {
       final TypeLSMSparseVectorIndexBuilder sparseBuilder = builder.withType(Schema.INDEX_TYPE.LSM_SPARSE_VECTOR)
@@ -360,7 +368,7 @@ public class CreateIndexStatement extends DDLStatement {
           throw new CommandSQLParsingException("Invalid METADATA for LSM_SPARSE_VECTOR index: " + e.getMessage(), e);
         }
       }
-      sparseBuilder.create();
+      resultingIndex = sparseBuilder.create();
 
     } else if (indexType == Schema.INDEX_TYPE.GEOSPATIAL) {
       // Builder is already a TypeGeoIndexBuilder after withType(GEOSPATIAL)
@@ -379,7 +387,7 @@ public class CreateIndexStatement extends DDLStatement {
           throw new CommandSQLParsingException(e.getMessage(), e);
         }
       }
-      geoBuilder.create();
+      resultingIndex = geoBuilder.create();
 
     } else {
       // Every index type whose settings live in METADATA is handled above. Reaching here with a METADATA clause means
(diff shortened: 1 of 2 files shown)


robfrank:
@claude review

PR #6004, opened 2026-08-09
fix(#5780): a manual index must report isAutomatic() = false

### LINKED ISSUE #5780: A manual index reports isAutomatic() = true, so REBUILD INDEX * targets an index it can never rebuild
Found while fixing #5765 (PR #5775), which made `Schema.buildManualIndex(...)` usable at all for the first time - the path was dead on arrival, so nothing downstream of it had ever been exercised. Verified against `main` at 0b2ccfb72.

## `isAutomatic()` answers the opposite of the truth for a manual index

`LSMTreeIndex.isAutomatic()` and `HashIndex.isAutomatic()` are both:

```java
public boolean isAutomatic() {
  return metadata.propertyNames != null;
}
```

A manual index is constructed with `this.metadata = new IndexMetadata(null, null, -1)` (`LSMTreeIndex:140` and `:159`), and that constructor **coerces a null property list to an empty one**:

```java
this.propertyNames = propertyNames != null ? List.of(propertyNames) : List.of();
```

So `propertyNames` is never null, and `isAutomatic()` is unconditionally `true` - including for the one kind of index for which it should be `false`. "Automatic" here means "there are indications of what to index", which is exactly what a manual index does not have.

## Two consequences

**1. `REBUILD INDEX *` targets manual indexes.** The sweep selects on that predicate:

```java
if (idx.isAutomatic() && !(idx instanceof TypeIndex))
  targetIndexes.add(idx);
```

Every manual index therefore enters the loop, fails inside `LSMTreeIndex.build()` with `Cannot rebuild index '<name>' because metadata information are missing`, and lands in the `failedIndexes` list the sweep reports. So a database that merely *contains* a manual index makes `REBUILD INDEX *` report a failure for something that was never rebuildable by definition, and an operator running the bulk repair tool has to work out that the reported failure is not a broken index.

**No data is lost** - `build()` raises before `status.compareAndSet(AVAILABLE, UNAVAILABLE)` and before the bucket scan, so the index is not emptied. This is a reporting and usability defect, not a corruption one.

**2. The accurate error message is unreachable.** `RebuildIndexStatement:282` already has exactly the right words for this case:

```java
if (!idx.isAutomatic())
  throw new CommandExecutionException(
      "Cannot rebuild index '" + idx.getName() + "' because it's manual and there aren't indications of what to index");
```

Because `isAutomatic()` can never be false for an LSM or HASH index, that branch is dead code, and `REBUILD INDEX <manualName>` reports the vaguer `metadata information are missing` from two layers down instead.

## Suggested fix

The intent of the predicate is "does this index know what to index", so make it say that:

```java
public boolean isAutomatic() {
  return metadata.propertyNames != null && !metadata.propertyNames.isEmpty();
}
```

That restores `RebuildIndexStatement:282` to reachable, drops manual indexes out of the `REBUILD INDEX *` target list, and makes the `build()` guard the belt-and-braces check it reads as rather than the primary one.

Worth checking as part of the same change whether any **other** caller of `isAutomatic()` was relying on the current always-true answer - a quick grep showed the two rebuild sites, but the predicate is public API on `Index`.

## Also worth a look while in there

`CHECK DATABASE` iterates the schema's indexes too. I did not trace whether its index-integrity pass handles a type-less index, and it is worth confirming in the same pass rather than discovering it the way #5765's defects were found. Related precedent: #5713, where `CHECK DATABASE FIX` was destroying the index it was diagnosing.

robfrank:
Closes #5780

## The defect

`LSMTreeIndex.isAutomatic()` and `HashIndex.isAutomatic()` both answered `metadata.propertyNames != null`, and `IndexMetadata`'s constructor coerces a null property list to an EMPTY one:

```java
this.propertyNames = propertyNames != null ? List.of(propertyNames) : List.of();
```

so the field is never null and the predicate was unconditionally `true` - including for a manual index, built with `new IndexMetadata(null, null, -1)` and the one kind for which it must be `false`. "Automatic" here means "there are indications of what to index", which is exactly what a manual index does not have.

The predicate now reads through a single named accessor, `IndexMetadata.hasIndexedProperties()`, so the empty check cannot be dropped back to a null check in one of the copies. `Index.isAutomatic()` is public API and had no Javadoc; it now says what it means.

## Every caller, checked

The issue asked whether any other caller relied on the always-true answer. A full grep of the repo (no module outside `engine/` calls it; Studio only renders the value it is told):

| Caller | Effect of the fix |
|---|---|
| `RebuildIndexStatement:157` (`REBUILD INDEX *` target selection) | manual indexes drop out of the sweep - **the reported bug** |
| `RebuildIndexStatement:288` (`if (!idx.isAutomatic())`) | dead branch becomes reachable, so a named rebuild reports the accurate message |
| `CompactIndexStatement:81` / `:124` | the same pair, which the issue's grep did not reach - see the note below |
| `PartitionedBucketSelectionStrategy:153` | unaffected: it tests a `TypeIndex`, whose `isAutomatic()` is an unconditional `true` |
| `FetchFromSchemaIndexesStep` / `FetchFromSchemaIndexDetailStep` / `FetchFromSchemaTypesStep` | `schema:indexes` now reports `automatic=false` for a manual index, which is the truth |

Three wrapper indexes (`LSMTreeFullTextIndex`, `LSMTreeGeoIndex`, `LSMSparseVectorIndex`) carried their own copy of the same always-true test, spelled `underlyingIndex.getPropertyNames() != null`. They now delegate to `underlyingIndex.isAutomatic()` so there is one answer rather than four. No behaviour change for them today - none of those kinds can be built without a type - but leaving the duplicate would have re-introduced the bug the moment one could.

**Note on `COMPACT INDEX`.** The guard at `CompactIndexStatement:124` was written to be reachable and now is, so `COMPACT INDEX <manualName>` reports "it's manual and there aren't indications of what to index" instead of compacting, and `COMPACT INDEX *` skips manual indexes. That is the existing code's stated intent, and manual indexes are still compacted by the background auto-compaction scheduler, which does not consult this predicate. If the preference is instead that an explicit compaction of a manual index should be allowed - compaction is a storage operation and genuinely does not need to know what to index - that is a one-line change in `compactIndex()` and I am happy to make it; I did not want to change a documented refusal on my own initiative.

## Two more defects found in the same pass

The issue asked to confirm `CHECK DATABASE` in the same pass rather than discovering it the way #5765's were found. Both of these are reproduced by tests in this PR and were verified to fail without their fix.

**1. `CHECK DATABASE FIX` aborted on a manual index.** The rebuild loop resolves the index's associated bucket to recreate it there. A manual index has none, so a manual index that the integrity pass flagged (e.g. a damaged hash metadata page) reached `getBucketById(-1)` and raised `SchemaException: Bucket with id '-1' was not found` - escaping the FIX before ANY index, including the healthy type indexes of the same database, had been repaired. Manual indexes are now filtered out of the affected set with a warning that names the only repair available, and are neither dropped nor recreated: the drop-and-recreate the loop performs would destroy entries nothing can regenerate.

**2. A manual index whose name has no underscore aborted the schema load.** The orphan-relinking pass of `LocalSchema.readConfiguration` walks every TYPE-LESS index and splits its name on the last underscore to recover the bucket a renamed sub-index belongs to - those are always `<bucketName>_<timestamp>`. A manual index is type-less too but is named by the caller, so `lastIndexOf` answered -1 and `substring(0, -1)` raised `StringIndexOutOfBoundsException`, which `readConfiguration`'s own catch logged as *"Error on loading schema. The schema will be reset"*.

The types survive (they are parsed before the pass), which is why this went unnoticed - but everything AFTER the pass is silently skipped on every open: the bucket selection strategies, the triggers, the materialized views and continuous aggregates, the function libraries, the extensions, and the compaction file-migration map WAL recovery redirects through. The pass now skips a name it cannot parse; the same guard is applied to its twin two lines below, which reads persisted schema entries and is reachable from a hand-edited or restored `schema.json`.

Neither is a regression from this change - both predate it - but both are in the blast radius the manual-index path opened up in #5765/#5775, and both are unreachable without a manual index, which is why they are here rather than in separate PRs.

## Test plan

New: `engine/src/test/java/com/arcadedb/index/Issue5780ManualIndexIsAutomaticTest.java` (7 tests). All 7 were confirmed RED against the unfixed code first, each for the reason it names.

- [ ] `aManualIndexIsNotAutomaticAndATypeIndexStillIs` - the predicate itself, on both manual kinds (`LSM_TREE`, `HASH`) and on a type index's wrapper and bucket sub-indexes
- [ ] `rebuildIndexAllSkipsManualIndexesInsteadOfFailingOnThem` - manual indexes absent from `indexes` AND from `failedIndexes` (where they used to land), the automatic indexes still rebuilt, the manual entries still readable afterwards
- [ ] `rebuildIndexNamedOnAManualIndexReportsThatItIsManual` - the accurate message instead of the one raised two layers down
- [ ] `compactIndexFollowsTheSamePredicate` - the `*` sweep and the named refusal
- [ ] `theSchemaViewReportsAManualIndexAsNotAutomatic` - `select from schema:indexes`
- [ ] `aManualIndexWithNoUnderscoreInItsNameDoesNotAbortTheSchemaLoad` - asserts the bucket selection strategy survives the reopen, i.e. the load reached PAST the orphan-relinking pass. Asserting the types alone would not have caught it
- [ ] `checkDatabaseFixReportsAManualIndexInsteadOfAbortingOnIt` - injects the exact hash metadata corruption `HashIndexMetadataCorruptionTest` uses, then asserts FIX reports and carries on, does not claim a rebuild, and does not drop the index

Regression scope run locally: the whole `com.arcadedb.index.**` and `com.arcadedb.schema.**` packages of `engine`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -232,6 +232,20 @@ public class DatabaseChecker {
     // no record corruption pointed at their bucket.
     affectedIndexes.addAll(corruptMetadataIndexes);
 
+    // A MANUAL index is reported but never rebuilt. It is bound to no type and no bucket, so there is no record to
+    // regenerate its entries from - they are the only copy - and the drop-and-recreate below would destroy them. It
+    // cannot even be attempted: with no associated bucket the very first line of the loop resolved bucket id -1 and
+    // raised "Bucket with id '-1' was not found", aborting the whole FIX before ANY index was repaired (issue #5780).
+    // The finding still reaches the operator through corruptedIndexes/warnings, which is all this pass can honestly
+    // do about it.
+    affectedIndexes.removeIf(index -> {
+      if (index.isAutomatic())
+        return false;
+      addWarning("index '" + index.getName() + "': it is a manual index, so it is not rebuilt automatically - its "
+          + "entries are not derived from any record. Drop it and repopulate it");
+      return true;
+    });
+
     final Set<String> rebuildIndexes = affectedIndexes.stream().map(x -> x.getName()).collect(Collectors.toSet());
     result.put("rebuiltIndexes", rebuildIndexes);
 
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -238,11 +238,15 @@ public class DatabaseChecker {
     // raised "Bucket with id '-1' was not found", aborting the whole FIX before ANY index was repaired (issue #5780).
     // The finding still reaches the operator through corruptedIndexes/warnings, which is all this pass can honestly
     // do about it.
+    //
+    // Filtered here rather than inside the `if (fix)` loop below so that rebuiltIndexes, which is reported in BOTH
+    // modes, does not name an index no run would ever rebuild. The message is therefore phrased for both: it states
+    // what this check does not cover, not what a repair attempt failed to do.
     affectedIndexes.removeIf(index -> {
       if (index.isAutomatic())
         return false;
-      addWarning("index '" + index.getName() + "': it is a manual index, so it is not rebuilt automatically - its "
-          + "entries are not derived from any record. Drop it and repopulate it");
+      addWarning("index '" + index.getName() + "': it is a manual index, so it is outside the automatic rebuild - its "
+          + "entries are not derived from any record. Drop it and repopulate it to clear this finding");
       return true;
     });
 
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -245,8 +245,8 @@ public class DatabaseChecker {
     affectedIndexes.removeIf(index -> {
       if (index.isAutomatic())
         return false;
-      addWarning("index '" + index.getName() + "': it is a manual index, so it is outside the automatic rebuild - its "
-          + "entries are not derived from any record. Drop it and repopulate it to clear this finding");
+      addWarning("index '" + index.getName() + "': it is a manual index. Its entries are not derived from any record, "
+          + "so this check does not rebuild it - drop it and repopulate it to clear this finding");
       return true;
     });
 
--- a/engine/src/main/java/com/arcadedb/index/Index.java
+++ b/engine/src/main/java/com/arcadedb/index/Index.java
@@ -96,5 +96,13 @@ public interface Index {
 
   boolean supportsOrderedIterations();
 
+  /**
+   * Whether something populates this index: it is bound to a type and to the properties of it whose values become its
+   * keys, so the engine maintains it on every write and a rebuild can regenerate it from the records.
+   * <p>
+   * False only for a MANUAL index, which is bound to nothing: its entries are whatever the caller put in it and there
+   * is no record to derive them from, which is why {@code REBUILD INDEX} and {@code COMPACT INDEX} refuse a named one
+   * and skip it in their {@code *} sweep (issue #5780).
+   */
   boolean isAutomatic();
 }
--- a/engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java
@@ -1236,7 +1236,9 @@ public class LSMTreeFullTextIndex implements Index, IndexInternal {
 
   @Override
   public boolean isAutomatic() {
-    return underlyingIndex.getPropertyNames() != null;
+    // Delegated rather than re-derived from getPropertyNames(): the list is never null (IndexMetadata coerces a
+    // missing one to empty), so the local test was the same always-true answer issue #5780 removed.
+    return underlyingIndex.isAutomatic();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/index/geospatial/LSMTreeGeoIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/geospatial/LSMTreeGeoIndex.java
@@ -467,7 +467,9 @@ public class LSMTreeGeoIndex implements Index, IndexInternal {
 
   @Override
   public boolean isAutomatic() {
-    return underlyingIndex.getPropertyNames() != null;
+    // Delegated rather than re-derived from getPropertyNames(): the list is never null (IndexMetadata coerces a
+    // missing one to empty), so the local test was the same always-true answer issue #5780 removed.
+    return underlyingIndex.isAutomatic();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/index/hash/HashIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/hash/HashIndex.java
@@ -363,7 +363,7 @@ public class HashIndex implements IndexInternal {
 
   @Override
   public boolean isAutomatic() {
-    return metadata.propertyNames != null;
+    return metadata.hasIndexedProperties();
   }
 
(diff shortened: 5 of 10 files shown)


robfrank:
Review cycle 1 (`e2340f16`) - response.

**Applied:** the `CHECK DATABASE` warning wording. The filter deliberately stays outside the `if (fix)` loop, because `rebuiltIndexes` is reported in both modes and must not name an index no run would ever rebuild; what needed fixing was the message, which now states what the check does not cover rather than implying a repair was attempted. Pushed as `6a27eab0`.

**Deferred, with rationale:** `FetchFromSchemaIndexesStep:70` / `FetchFromSchemaIndexDetailStep:68` guarding the `properties` field on `getPropertyNames() != null`. Confirmed it is the same always-true shape, and today a manual index reports `properties: [[]]` instead of omitting the key. Not folded in here because it changes the SHAPE of a public schema-view row rather than the truth of a value - a client that reads `properties` unconditionally starts seeing the key disappear - and that deserves its own change with its own regression test rather than riding along in a review cycle. Flagging it for a follow-up issue.

No other feedback outstanding.

robfrank:
Review cycle 2 (`6a27eab0`) - response.

**Applied:**
- The warning wording again, now `index 'X': it is a manual index. Its entries are not derived from any record, so this check does not rebuild it - drop it and repopulate it to clear this finding`.
- The coverage gap: `namedRebuildAndCompactStillRunOnAnAutomaticIndex` exercises `REBUILD INDEX <typeIndexName>` and `COMPACT INDEX <bucketSubIndexName>` with a manual index present in the same database. That was a real gap and worth closing for the reason given - the named path is handed the `TypeIndex` wrapper, which the `*` sweep deliberately excludes, so the sweep test could not have covered it. Confirmed the new test goes red when `hasIndexedProperties()` is forced to `false`.

**Not applied, because the premise does not hold:** the suggestion to comment `HashIndex:454` / `LSMTreeIndex:238` as the reason the null branch of `hasIndexedProperties()` is live. Both sites assign `this.metadata.propertyNames = new ArrayList<>()` and then fill it, so neither can produce a null - and grepping every assignment to that field in `engine/src/main` (`IndexMetadata:91` via `toListOfStrings()`, `LocalDocumentType:1311` via `List.of(...)`, `BucketLSMVectorIndexBuilder:272` copying a non-null list, `LSMVectorIndex:5377` a fresh `ArrayList`) finds no site that assigns null either. The null half is belt-and-braces on a public mutable field; the empty half is the load-bearing one, which is what the Javadoc already says. Adding a comment pointing at those two lines as the justification would document something that is not true of them.

Nothing outstanding from this cycle. The `FetchFromSchema*Step` `properties` observation from cycle 1 remains deferred for a follow-up issue, with the rationale in my previous comment.

robfrank:
Review cycle 3 (`f34d55d9`) - clean approval, no changes applied.

Both remaining observations are explicitly non-blocking and I agree with the reviewer's own reading of them: the `addWarning` side effect inside `removeIf` is single-threaded and runs once per index, and the `COMPACT INDEX` question is left in the description for maintainer judgment rather than decided here.

On the note that the reviewer's sandbox could not run Maven: the suite was run locally against this branch.
- `Issue5780ManualIndexIsAutomaticTest` - 8/8 green, and each test was confirmed RED first against the unfixed code (the last one by forcing `hasIndexedProperties()` to `false`).
- `com.arcadedb.index.**` + `com.arcadedb.schema.**` - 254 test classes, 0 failures, 0 errors.
- `CheckDatabaseTest`, `CheckDatabaseProgressTest`, `GraphDatabaseChecker*Test`, `CompressDatabaseTest`, `HashIndexMetadataCorruptionTest`, `Issue5713HashIndexPageSizeTest`, `Issue5724HashIndexKeyTypeGuardTest`, `Issue5765ManualIndexIfNotExistsTest`, `LocalSchemaOrphanIndex*Test`, `RebuildIndex*Test` - 90 tests, 0 failures.

The one deferred item from cycle 1 is now filed as #6005 rather than left in a comment.

Ready for maintainer review. Merge is yours.

PR #6006, opened 2026-08-09
fix(schema): omit empty properties field from schema:indexes rows (#6005)

### LINKED ISSUE #5780: A manual index reports isAutomatic() = true, so REBUILD INDEX * targets an index it can never rebuild

### LINKED ISSUE #6005: schema:indexes reports properties: [[]] for a manual index because the guard is the always-true getPropertyNames() != null
Split out of the review of #6004 (fixes #5780), where the same always-true shape was the root cause.

`FetchFromSchemaIndexesStep:70` and `FetchFromSchemaIndexDetailStep:68` decide whether to populate the `properties` field of a schema-view row with:

```java
if (index.getPropertyNames() != null)
  r.setProperty("properties", Collections.singletonList(index.getPropertyNames()));
```

`IndexMetadata` coerces a missing property list to `List.of()`, so `getPropertyNames()` is never null and the guard never omits the field - it is the same tautology #5780 removed from `isAutomatic()`. A manual index, which is bound to no type and no properties, therefore reports `properties: [[]]` rather than omitting the key.

Nothing breaks today, which is why it was not folded into #6004: the fix changes the SHAPE of a public schema-view row (a client reading `properties` unconditionally would start seeing the key disappear), so it wants its own change and its own regression test rather than riding along in a review cycle of a bug fix.

Worth deciding at the same time whether the right answer is to omit the key or to emit an empty list explicitly - `Studio`'s index panel reads `automatic` from these rows and should be checked against whichever is chosen.

waterWang:
## Description

`IndexMetadata` coerces a missing property list to `List.of()`, so `getPropertyNames()` is never null. A manual index, which is bound to no type and no properties, therefore reports `properties: [[]]` (an empty list wrapped in a singleton list) rather than omitting the key.

### Fix

Changed the guard from `if (index.getPropertyNames() != null)` to `if (!index.getPropertyNames().isEmpty())` in two files:

- `FetchFromSchemaIndexesStep.java` — schema:indexes listing
- `FetchFromSchemaIndexDetailStep.java` — schema:indexes detail query

This ensures that indexes with no properties (e.g., manual indexes) correctly omit the `properties` field from the result row, rather than emitting `properties: [[]]`.

### Related

Split out of the review of #6004 (fixes #5780), where the same always-true shape was the root cause.

Closes #6005


lvca:
@claude review this

lvca:
Thanks @waterWang !

Issue #4935, opened 2026-07-03
[tx] Updating the same indexed record twice in one transaction permanently corrupts the index


lvca:
**Severity:** CRITICAL - **reproduced on current `main`.**

**Location:** `LocalDatabase.java:1068-1078` (eager per-save index diffing in the tx-active branch), `TransactionIndexContext.java:296-368` (`addIndexKeyLock`, no cross-key reconciliation), `index/lsm/LSMTreeIndex.java:437,515`.

In the tx-active branch of `updateRecord()`, every `save()` calls `indexer.updateDocument(originalRecord, document)` where `originalRecord` is always rebuilt from the record`s **committed** buffer (serialization is deferred, so the buffer never advances inside the tx). Save #1 buffers `REMOVE(orig), ADD(mid)`; save #2 buffers `REMOVE(orig), ADD(final)`. `ADD(mid)` is never cancelled and at commit `putReplay` persists **both** `mid` and `final`.

**Repro (fails on main):** unique LSM index on `email`; commit `orig`; in one tx `doc.set("email","mid"); doc.save(); doc.set("email","final"); doc.save();` commit. Then:
- `lookupByKey(email=`mid`)` returns the record (whose email is `final`) - a phantom entry;
- a later insert of another record with `email=`mid`` throws `DuplicatedKeyException ... already assigned to #x` at `TransactionIndexContext.checkUniqueIndexKeys:404` - the unique constraint blocks a legitimate value forever.

Reachable from ordinary SQL: `UPDATE ... SET x=1; UPDATE ... SET x=2` in one tx, or any script/graph mutation that saves twice.

**Fix:** compute the index diff against the last-buffered indexed state per RID (track per-tx "current indexed keys" per record in `TransactionIndexContext`), or drop the eager save-time `updateDocument` and buffer during phase 1. Add the repro above as a regression test.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #4964 (merged).

The in-transaction index update now diffs the second and subsequent saves of the same record against a per-RID snapshot of the previous in-transaction indexed state, instead of the committed buffer (which stays frozen until commit because serialization is deferred). The intermediate ADD is therefore cancelled and no phantom entry survives.

[...]

Issue #13, opened 2021-08-26
Support for Append Only buckets


lvca:
Having append only structures allows some interesting use cases like ledgers, block chains,

PR #6007, opened 2026-08-09
fix(#5779): moveEdge must clean the old edge record's index entries, external values and bucket count

### LINKED ISSUE #5779: moveEdge deletes the old edge record without cleaning its index entries, and stays correct only by slot reuse
Surfaced during the review of #5772 (issue #5760). Not a regression from that PR - it predates it, and #5772 does not touch `moveEdge`. Filed so the finding does not live only in a code comment, at the reviewer's suggestion.

## What happens

`GraphEngine.moveEdge` re-points an edge at a new endpoint by deleting the old edge record and creating a replacement:

```java
deleteEdge(edge);
...
final MutableEdge newEdge = new MutableEdge(database, edgeType, newOut, newIn);
if (!properties.isEmpty())
  newEdge.set(properties);
newEdge.save();
```

It calls the **public `GraphEngine.deleteEdge(Edge)` directly**, not through `LocalDatabase.deleteRecordNoLock`. That matters, because the record removal inside `deleteEdge` is deliberately the *physical* one only:

```java
// The physical removal only: on the DELETE PATH the index cleanup has already happened.
final LocalBucket bucket = (LocalBucket) database.getSchema().getBucketById(edge.getIdentity().getBucketId());
bucket.deleteRecord(edge.getIdentity());
```

The precondition it names - that `deleteRecordNoLock` already cleaned the record's index entries and fired its delete events - holds for the delete path and **not** for `moveEdge`. So the old edge record's index entries are never removed.

Reached from the embedded API through `MutableEdge.setIn(...)` / `setOut(...)`.

## Why it does not currently break

I could not construct a failure, and that is the point of filing it rather than fixing it blind. `moveEdge` re-creates the edge with the same properties, and the bucket hands the just-freed slot straight back, so the stale index entry ends up pointing at the **new** record carrying the **same key**. It is indistinguishable from a correctly maintained entry.

Measured on an edge type with 8 buckets and a `NOTUNIQUE` index on a `STRING` property, 41 edges:

```
before moveEdge: 41 records, 41 index entries
after  moveEdge: 41 records, 41 index entries, old RID identical to new RID
CHECK DATABASE: totalErrors 0, corruptedIndexes [], invalidLinks 0
```

## Why it is still worth closing

The correctness rests on an allocation coincidence, not on an invariant anything enforces:

- a replacement landing in a **different bucket** (the type is multi-bucket; bucket selection is not pinned to the old record's bucket) leaves the old entry naming a deleted RID;
- a **concurrent allocation** taking the freed slot first does the same;
- worse than a plain dangling entry, the slot could later be reused by an **unrelated** edge, at which point the stale entry silently returns a wrong record for that key.

A `UNIQUE` index makes it sharper: a stale entry for the old key could reject a legitimate later insert.

## Suggested fix

Have `moveEdge` clean the old record's index entries explicitly before `deleteEdge(edge)` - `DocumentIndexer.deleteDocument(edge)` is what `deleteRecordNoLock` uses - rather than relying on slot reuse. That keeps the "physical removal only" contract of `deleteEdge(Edge)` intact instead of weakening it for one caller.

Two things to check while there, both unverified:

1. **External property values.** `deleteRecordNoLock` also calls `cascadeDeleteExternalValues(document)`, which `moveEdge` likewise skips. An edge carrying an EXTERNAL value would have the old external record orphaned while `newEdge.set(properties)` writes a fresh one.
2. **Delete events.** `moveEdge` fires no `onBeforeDelete`/`onAfterDelete` for the record it removes, and no create events for the replacement. That may well be intended - a move is not a delete - but it is currently a side effect of the route taken rather than a decision.

A regression test should use a multi-bucket edge type and assert `index.countEntries()` (not a query, which skips unresolvable RIDs and would hide the leak) plus `CHECK DATABASE`.

## Context

The precondition is now documented at the site in #5772, including this measurement and what would break the coincidence, so a future caller of the public `deleteEdge(Edge)` meets the obligation rather than the old flat assertion.

robfrank:
Closes #5779

## Summary

`GraphEngine.moveEdge` re-points an edge by deleting the old edge record and creating a replacement, and it reaches the removal through the **public** `GraphEngine.deleteEdge(Edge)`. That method does the *physical* removal only - it is normally entered from `LocalDatabase.deleteRecordNoLock`, which has already cleaned the record's index entries, cascade-deleted its EXTERNAL property values and folded the `-1` into the bucket record counter by the time the edge arrives. `moveEdge` is the one caller that does not come that way, so none of that happened for the record it deletes. This PR discharges that obligation explicitly in a new private `GraphEngine.cleanUpBeforePhysicalDelete(MutableEdge)`, called just before `deleteEdge(edge)`, which keeps the "physical removal only" contract of the public `deleteEdge(Edge)` intact instead of weakening it for one caller (the approach the issue suggested).

## Why it was not "benign after all"

The issue filed this as latent, on the measurement that the bucket normally hands the just-freed slot straight back to the replacement, so the stale entry lands on a new record carrying the same key and is indistinguishable from a correct one. That coincidence does **not** survive the default schema. `RoundRobinBucketSelectionStrategy` (the default for every type) advances one bucket per create, so on an edge type with more than one bucket the replacement lands in a *different* bucket and the old entry names a RID that is gone. The three failure modes the issue predicted all reproduce on a 4-bucket edge type, and one of them is a hard user-visible failure today:

| Fixture | Before | After |
|---|---|---|
| `NOTUNIQUE` index, one edge moved | `countEntries()` = **2**, `get("E1")` = `[#14:0, #13:0]` | 1 entry, naming the surviving record |
| Freed slot handed to an unrelated edge | lookup of the old key returns **two** records, one of them somebody else's | exactly one, carrying the right key |
| `UNIQUE` index | **`DuplicatedKeyException: Duplicated key [E1] ... already assigned to record #13:0`** - the move is impossible | move succeeds, one entry |

The UNIQUE case means moving an edge (`MutableEdge.set("@in"/"@out")`, i.e. `UPDATE ... SET @in`) has been outright broken on any edge type carrying a unique index on an edge property, not merely fragile.

## The two "check while there" items, both confirmed real

Both were listed as unverified in the issue. Both are fixed here, each with a test proven to fail when only its part of the fix is reverted.

1. **EXTERNAL property values.** `moveEdge` writes fresh external records for the replacement while the old ones survive. `CHECK DATABASE` reports `orphanedExternalRecords: 1` after a single move of an edge with an EXTERNAL property; 0 with the fix.
2. **Bucket record counter.** `deleteRecordNoLock` folds `-1` for the record it removes and the create path folds `+1`, so a move is net zero. Going through `deleteEdge` skipped the `-1` only, so `count(*)` (which reads the cached counter, and which is persisted to `statistics.json`) reported **2** for a type holding **1** edge after one move - and the drift survives a reopen. `count()`, a real scan, correctly said 1.

## The one item deliberately NOT changed

**Delete/create events still do not fire.** The issue noted this is currently a side effect of the route taken rather than a decision; it is now a decision, documented at the site. A move is not a delete, and - decisive - `deleteRecordNoLock` lets a listener *veto* the removal by returning `false` from `onBeforeDelete`. Routing the move through the database would let a veto abort the removal while `moveEdge` went on to create the replacement anyway, turning a move into a silent duplication. That is why `cleanUpBeforePhysicalDelete` mirrors `deleteRecordNoLock`'s bookkeeping rather than calling it.

Lightweight edges are excluded from the whole cleanup: `deleteEdge(Edge)` skips the physical removal for them, so there is no index entry, no external value and no counted record, and folding a `-1` would drift the counter the other way.

The long comment inside `deleteEdge(Edge)` that #5772 added to describe the hazard is rewritten to state the caller precondition and point at where it is now discharged, instead of describing an unfixed leak.

## A third defect, found during review: the index cleanup was reading the wrong image

`DocumentIndexer.deleteDocument` builds the keys it removes from the LIVE property values of the record handed to it, and `moveEdge` is entered from `MutableEdge.set("@in"/"@out")` - so a caller that changes an indexed property in the same session (`edge.set("code","E2"); edge.set("@in", v)`) had it remove a key the index never held, leaving the real entry dangling. The same leak, through a different door. `indexedImageOf(MutableEdge)` now resolves the image the index was actually built from, in the order and for the reason `LocalDatabase.updateRecord` does (#4935):

1. `TransactionContext.getLastIndexedSnapshot(rid)` when an earlier `save()` in this transaction already moved the index forward. Serialization is deferred, so after `set("code","E2"); save(); set("@in", v)` the record's frozen buffer still says `E1` while the index holds `E2` - the buffer is the *wrong* pre-image in that case.
2. otherwise an immutable record over that frozen buffer.
3. otherwise the live instance. Defensive: verified unreachable from `moveEdge` by making the branch throw and watching the class still pass.

## The boundary of the guarantee

An indexed property whose value cannot be **deserialized** (the #4420 shape: a corrupt length prefix) is reported by the reader as ABSENT rather than throwing, so nothing can learn the key the index holds and that entry outlives the record. Measured: `lookup(marker)` returns the deleted RID, and `CHECK DATABASE` does not detect a dangling LSM entry (`totalErrors 0`, `corruptedIndexes []`). Not introduced here and not closable here - `deleteRecordNoLock` degrades identically on the same record, by design. Pinned by a test rather than left implicit.

## Test plan

New: `engine/src/test/java/com/arcadedb/graph/Issue5779MoveEdgeIndexCleanupTest.java` (11 tests). All assertions go through `Index.countEntries()` and a raw `Index.get()` rather than a query, because `SELECT` skips entries whose RID does not resolve - which is exactly what hid this leak.

- [ ] `mvn -pl engine test -Dtest=Issue5779MoveEdgeIndexCleanupTest` - 11/11 green.
- [ ] Every non-trivial branch of the fix was individually reverted to prove the matching test fails **and the others do not**: the index cleanup (3 tests), the external cascade (`movingAnEdgeCascadeDeletesTheOldRecordsExternalValues`), the counter delta (the two counter assertions), the pre-image resolution (`movingAnEdgeThatAlsoChangedItsIndexedPropertyCleansTheKeyTheIndexActuallyHolds`), and the snapshot lookup within it (`movingAnEdgeAfterAnIndexedPropertyWasAlreadySavedInTheSameTransaction`).
- [ ] The fixtures assert the replacement landed in a **different** bucket (`isNotEqualTo(edgeRID)`) before asserting anything else, so a future change to bucket selection cannot silently turn these into no-ops.
- [ ] Coverage: NOTUNIQUE leak, freed-slot reuse by an unrelated edge, UNIQUE duplicate rejection, `@in` and `@out`, indexed property changed during the move, changed and saved before the move, edge created in the same transaction, EXTERNAL cascade, bucket-counter drift, corrupted-buffer boundary, lightweight-edge unreachability.
- [ ] Regression: `mvn -pl engine test -Dtest='com.arcadedb.graph.*Test,com.arcadedb.index.**.*Test,com.arcadedb.database.**.*Test,UpdateEdgeTest,QueryTest'` - 1697 tests, 0 failures, 1 skipped.
- [ ] `CHECK DATABASE` clean (`totalErrors` 0, `autoFix` 0) in every fixture except the corrupted-buffer one, where the leak it cannot see is asserted explicitly.

## Adjacent finding, not fixed here

Calling `edge.save()` after `edge.set("@in", newRID)` throws `IllegalStateException: Cannot read original buffer for record #N:0`. `moveEdge` already saves the replacement and re-identifies the instance onto it via `updateIdentity`, which nulls the buffer, so the subsequent `save()` takes the update path on a record it has no pre-image for. The natural embedded-API call sequence is therefore broken. Out of scope for the index leak; worth its own issue.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
@@ -852,6 +852,13 @@ public class GraphEngine {
     // readable. moveEdge writes FRESH external records for the replacement, so without this the old ones are
     // orphaned. Same transaction as the removal itself. Runs BEFORE the index cleanup because it reads the
     // record's own buffer and restores its position, while the pre-image built below consumes that same buffer.
+    //
+    // Deliberately NOT pre-gated on hasExternalProperties() the way LocalDatabase.cascadeDeleteExternalValues is:
+    // that flag reflects the CURRENT schema and flips to false the instant setExternal(false) commits, so it would
+    // skip cleanup during the EXTERNAL -> inline migration window and leak the blobs of every record not yet
+    // re-serialised. findExistingExternalRids gates itself on hasExternalBuckets(), which stays true for the whole
+    // window and is the correct test - and it is already a couple of field reads for a type that never used the
+    // feature, so there is nothing to save by guarding it again here.
     final Map<String, RID> externalRids = database.getSerializer().findExistingExternalRids(database, edge);
     for (final RID externalRID : externalRids.values()) {
       final Bucket externalBucket = database.getSchema().getBucketByIdIfExists(externalRID.getBucketId());
@@ -882,8 +889,10 @@ public class GraphEngine {
    *   index has left behind;</li>
    *   <li>otherwise an immutable record over that frozen buffer, which is exactly the committed state the index
    *   entries were written from;</li>
-   *   <li>otherwise the live instance, which is the only image there is for a record created inside this
-   *   transaction (no committed buffer to read).</li>
+   *   <li>otherwise the live instance, for a record with no committed buffer to read at all - one created inside
+   *   this transaction, and equally one saved with {@code discardRecordAfter} (the bulk-import path through
+   *   {@code LocalBucket.createRecord}), which drops the buffer although the record is persisted. There is no
+   *   better image available in either case, and it is the same one the ordinary delete path uses.</li>
    * </ol>
    */
   private Document indexedImageOf(final MutableEdge edge) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
@@ -19,6 +19,7 @@
 package com.arcadedb.graph;
 
 import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.database.Binary;
 import com.arcadedb.database.Database;
 import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.Document;
@@ -27,6 +28,7 @@ import com.arcadedb.database.LocalDatabase;
 import com.arcadedb.database.MutableDocument;
 import com.arcadedb.database.RID;
 import com.arcadedb.database.Record;
+import com.arcadedb.database.RecordInternal;
 import com.arcadedb.database.TransactionContext;
 import com.arcadedb.exception.ConcurrentModificationException;
 import com.arcadedb.engine.Bucket;
@@ -833,30 +835,70 @@ public class GraphEngine {
    * it, so there is no index entry, no external value and no counted record, and folding a {@code -1} here would
    * drift the cached bucket count behind {@code count(*)} - persisted in {@code statistics.json}, so it would
    * survive a reopen.
+   * <p>
+   * Unlike {@code deleteRecordNoLock} this does NOT tolerate a corrupted record buffer, and does not need to: that
+   * tolerance exists so a record whose body cannot be parsed can still be DELETED, while {@link #moveEdge} has
+   * already had to parse the whole record into {@code propertiesAsMap()} - the source of the replacement's
+   * properties - before it gets here. A buffer that cannot be read fails the move at that point, which is the right
+   * answer: there is nothing to copy the edge's content from. The one buffer read left here,
+   * {@code findExistingExternalRids}, absorbs its own parse failures and counts them for CHECK DATABASE.
    */
   private void cleanUpBeforePhysicalDelete(final MutableEdge edge) {
     final RID edgeRID = edge.getIdentity();
     if (edgeRID == null || edge instanceof LightEdge || edgeRID.getPosition() < 0)
       return;
 
-    database.getIndexer().deleteDocument(edge);
-
     // Cascade-delete EXTERNAL property values living in the paired external bucket, while the buffer is still
     // readable. moveEdge writes FRESH external records for the replacement, so without this the old ones are
-    // orphaned. Same transaction as the removal itself.
+    // orphaned. Same transaction as the removal itself. Runs BEFORE the index cleanup because it reads the
+    // record's own buffer and restores its position, while the pre-image built below consumes that same buffer.
     final Map<String, RID> externalRids = database.getSerializer().findExistingExternalRids(database, edge);
     for (final RID externalRID : externalRids.values()) {
-      final LocalBucket externalBucket = database.getSchema().getEmbedded().getBucketById(externalRID.getBucketId(), false);
+      final Bucket externalBucket = database.getSchema().getBucketByIdIfExists(externalRID.getBucketId());
       if (externalBucket != null) {
         externalBucket.deleteRecord(externalRID);
         database.getTransaction().updateBucketRecordDelta(externalBucket.getFileId(), -1);
       }
     }
 
+    database.getIndexer().deleteDocument(indexedImageOf(edge));
+
     // The replacement folds its own +1 when it is created, so without this the move inflates the cached record
     // count of the type by one for every edge ever moved.
     database.getTransaction()
         .updateBucketRecordDelta(database.getSchema().getBucketById(edgeRID.getBucketId()).getFileId(), -1);
+  }
+
+  /**
+   * The image of {@code edge} that the INDEX was built from, which is NOT the instance the caller holds:
+   * {@code DocumentIndexer.deleteDocument} builds the keys it removes out of the LIVE property values of whatever
+   * record it is handed, and {@link #moveEdge} is entered from {@code MutableEdge.set("@in"/"@out")} - so a caller
+   * that changed an indexed property in the same session ({@code edge.set("code", "new"); edge.set("@in", v)})
+   * would have it remove a key the index never held, leaving the real entry dangling. Same resolution order
+   * {@code LocalDatabase.updateRecord} uses for the same reason (#4935):
+   * <ol>
+   *   <li>the transaction's last indexed snapshot, when an earlier {@code save()} in this transaction already
+   *   moved the index forward - the record's buffer stays frozen until commit, so it would describe a state the
+   *   index has left behind;</li>
+   *   <li>otherwise an immutable record over that frozen buffer, which is exactly the committed state the index
+   *   entries were written from;</li>
+   *   <li>otherwise the live instance, which is the only image there is for a record created inside this
+   *   transaction (no committed buffer to read).</li>
+   * </ol>
+   */
+  private Document indexedImageOf(final MutableEdge edge) {
+    final RID edgeRID = edge.getIdentity();
+    final Document snapshot = database.getTransaction().getLastIndexedSnapshot(edgeRID);
+    if (snapshot != null)
+      return snapshot;
+
+    final Binary committedBuffer = ((RecordInternal) edge).getBuffer();
+    if (committedBuffer == null)
+      return edge;
+
+    committedBuffer.rewind();
+    return (Document) database.getRecordFactory()
+        .newImmutableRecord(database, edge.getType(), edgeRID, committedBuffer, null);
   }
 
   public void deleteVertex(final VertexInternal vertex) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
@@ -836,12 +836,19 @@ public class GraphEngine {
    * drift the cached bucket count behind {@code count(*)} - persisted in {@code statistics.json}, so it would
    * survive a reopen.
    * <p>
-   * Unlike {@code deleteRecordNoLock} this does NOT tolerate a corrupted record buffer, and does not need to: that
-   * tolerance exists so a record whose body cannot be parsed can still be DELETED, while {@link #moveEdge} has
-   * already had to parse the whole record into {@code propertiesAsMap()} - the source of the replacement's
-   * properties - before it gets here. A buffer that cannot be read fails the move at that point, which is the right
-   * answer: there is nothing to copy the edge's content from. The one buffer read left here,
-   * {@code findExistingExternalRids}, absorbs its own parse failures and counts them for CHECK DATABASE.
+   * Unlike {@code deleteRecordNoLock} this takes no catch around the cleanup, and needs none: by the time it runs,
+   * {@link #moveEdge} has already read the whole record through {@code propertiesAsMap()} - the source of the
+   * replacement's properties - so a buffer broken badly enough to throw has already failed the move there, which
+   * is the right answer for a move (there is nothing to copy the edge's content from) even though it is the wrong
+   * one for a delete. The remaining buffer read here, {@code findExistingExternalRids}, absorbs its own parse
+   * failures and counts them for CHECK DATABASE.
+   * <p>
+   * That leaves one BOUNDARY, measured and pinned by
+   * {@code Issue5779MoveEdgeIndexCleanupTest.movingAnEdgeWithAnUnreadablePropertyIsBestEffortLikeTheDeletePath}:
+   * corruption the reader TOLERATES - a bad length prefix on an inline value (#4420) surfaces as the property
+   * being absent, not as an exception. Nothing here can then learn the key the index holds, so that entry outlives
+   * the record. It is not a leak this method can close, and not one it introduces: {@code deleteRecordNoLock}
+   * degrades identically on the same record, and CHECK DATABASE does not detect a dangling LSM entry either.
    */
   private void cleanUpBeforePhysicalDelete(final MutableEdge edge) {
     final RID edgeRID = edge.getIdentity();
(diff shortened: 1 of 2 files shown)


robfrank:
Thanks - items 2, 3 and 5 were right and are fixed in 48726ff. Items 1 and 4 do not hold as stated; both are now documented or locked in by a test rather than left as a maybe.

**2 - index cleanup read live values. Confirmed, and it was a real residual leak.** Reproduced before fixing: `edge.modify(); edge.set("code", "E2"); edge.set("@in", far)` left `E1 -> #13:0` in the index naming the deleted record. `cleanUpBeforePhysicalDelete` now resolves the image the INDEX was built from, in the same order and for the same reason `LocalDatabase.updateRecord` does (#4935), via a new `indexedImageOf(MutableEdge)`:

1. `TransactionContext.getLastIndexedSnapshot(rid)` when an earlier `save()` in this transaction already moved the index forward. This is not hypothetical either: serialization is deferred, so after `edge.set("code","E2"); edge.save(); edge.set("@in", far)` the frozen buffer still says `E1` while the index holds `E2` - the committed buffer is the *wrong* pre-image in that case.
2. otherwise an immutable record over that frozen buffer.
3. otherwise the live instance, the only image that exists for a record created inside this transaction.

Two new tests, one per non-trivial branch. Each was proven to fail with only its branch reverted: disabling the pre-image resolution fails exactly `movingAnEdgeThatAlsoChangedItsIndexedPropertyCleansTheKeyTheIndexActuallyHolds`; disabling only the snapshot lookup fails exactly `movingAnEdgeAfterAnIndexedPropertyWasAlreadySavedInTheSameTransaction`.

**5 - `getBucketByIdIfExists`. Applied**, the `getEmbedded()` hop is gone. The external cascade also had to move *before* the index cleanup: it reads the record's own buffer (restoring its position), while the pre-image built in step 2 above consumes that same buffer.

**3 - `@out` untested. Added**, `movingTheOutEndpointCleansTheOldRecordsIndexEntriesToo`.

**1 - corrupted-buffer tolerance. Intentionally omitted, now stated at the site.** The tolerance in `deleteRecordNoLock` exists so a record whose body cannot be parsed can still be *deleted*. `moveEdge` cannot want that: it has already parsed the entire record via `propertiesAsMap()` - the source of the replacement's properties - one line before the cleanup runs, so an unparseable buffer fails the move there and never reaches `cleanUpBeforePhysicalDelete`. Failing is also the correct outcome, since there is nothing to copy the edge's content from. The only buffer read left in the cleanup is `findExistingExternalRids`, which absorbs its own parse failures and counts them for CHECK DATABASE.

**4 - lightweight edges reach `moveEdge`. They do not, and the gate is not the one named.** `MutableEdge.set`'s `rid != null` test is never evaluated for a lightweight edge because neither door into it is open:

- `ImmutableLightEdge.modify()` throws `IllegalStateException: Lightweight edges cannot be modified`, so a persisted lightweight edge reached through its endpoint's list cannot be made mutable at all;
- `MutableLightEdge` overrides every `set(...)` to throw `Edge type 'X' is declared LIGHTWEIGHT ... cannot have properties`, so even the instance `newEdge()` hands back inside the creating session never enters the `"@in"`/`"@out"` override.

I tried to write the test you asked for and it failed on both routes, which is the useful result. `aLightweightEdgeCannotReachMoveEdgeAtAll` now asserts both refusals, so the `instanceof LightEdge` guard is documented as defensive - and if either door is ever opened, that test fails and points at the guard that has just become load-bearing.

Regression after the changes: 1702 tests across `com.arcadedb.graph`, `com.arcadedb.index`, `com.arcadedb.database`, `QueryTest`, `UpdateEdgeTest`, `EdgeTypeVertexRidOptimizationTest` - 0 failures.

robfrank:
All four are non-blocking; 1 and 2 are addressed in 02cb7f4, 3 is declined with a reason.

**1 - the missing `hasExternalProperties()` fast path. Deliberate, and adding it would be a bug.** The asymmetry with `cascadeDeleteExternalValues` is real but it runs the other way: `hasExternalProperties()` reflects the CURRENT schema and flips to false the instant `setExternal(false)` commits, so as a gate it skips cleanup for the whole EXTERNAL -> inline migration window and leaks the blobs of every record not yet re-serialised. `findExistingExternalRids` gates itself on `hasExternalBuckets()`, which stays true across that window and is the correct test - `BinarySerializer` documents exactly this distinction as load-bearing. I have added the reasoning at the call site so the next reader (or reviewer) does not have to rediscover it.

**2 - the `discardRecordAfter` case the Javadoc did not mention. Corrected.** You are right that it also produces a null buffer on a persisted record, and I went looking for what else the fallback silently covers. The interesting one is create-then-modify-then-move in a single transaction, where there is no snapshot either, so I measured it against a control instead of reasoning about it:

- move path: `a.newEdge("Link", b, "code", "E1"); e.set("code","E2"); e.set("@in", c)` in one transaction leaves `countEntries() == 1`. Clean.
- the ordinary DELETE path with the same shape (`newEdge(...,"D1"); e.set("code","D2"); e.delete()`) leaves **2** entries, one of them naming the deleted record.

So the buffer-less fallback is not hiding a hole in this PR, and the pre-existing engine-wide one it brushes against lives on `deleteRecordNoLock`, not on `moveEdge`. Out of scope here; flagging it since I have the measurement.

**3 - extracting `assertIntegrityClean`/`longProperty` into `TestHelper`. Declining.** You correctly identify it as a pre-existing convention rather than something this PR introduces - the same private helper is already duplicated in `Issue5725GhostEdgeOnAppendRaceTest`, `ConcurrentEdgeAppendMergeTest` and `Issue5760VertexDeleteSelfSideSkipTest`. Hoisting it means editing a base class every engine test extends, which is a wide blast radius to attach to a `moveEdge` bugfix, and it would leave the three existing copies behind unless I also touched those tests. Better as its own cleanup across all four call sites.

Regression after the change: 1695 tests across `com.arcadedb.graph`, `com.arcadedb.index`, `com.arcadedb.database`, `QueryTest`, `UpdateEdgeTest` - 0 failures. This push is comment-only in `GraphEngine`; no behaviour changed since the previous review.

robfrank:
Both test-coverage items are in 11ceb10, and writing them corrected two claims I had made too strongly. Item 3 is acknowledged without a change.

**1 - the live-instance fallback in `indexedImageOf`. Tested, and it turns out to be unreachable from here.** I wrote `movingAnEdgeCreatedInTheSameTransactionUsesItsCommittedPreImage` (create the edge, change its indexed property, move it, all in one transaction with no intervening commit) and then checked it actually lands where you predicted, by making the `committedBuffer == null` branch throw. The whole class still passed. `newEdge()` saves the edge and the save serializes, so an edge created earlier in the same transaction reaches `moveEdge` with a committed buffer like any other record - the intuition points the other way, which is why the test is worth keeping even though it exercises the ordinary path.

The branch is now labelled defensive rather than described as covering the same-transaction case, with the `discardRecordAfter` case you raised as the shape that would reach it, and the reason it exists at all: the alternative on a null buffer is what `LocalDatabase.getOriginalDocument` does - raise `IllegalStateException` - and failing a move over a missing pre-image is worse than cleaning up from the only image available.

**2 - the corrupted-buffer decision. Tested, and my javadoc was wrong.** I ported the #4420 corruption (bad length prefix on an inline indexed STRING) onto an edge and moved it. The reader **tolerates** that shape by reporting the property as ABSENT instead of throwing, so the move does not fail the way I claimed - it goes through, and the entry keyed by the unreadable value outlives the record:

- `lookup("CORRUPT5779MARKERVALUE")` after the move: `[#13:0]`, the record that was just deleted
- `CHECK DATABASE`: `totalErrors 0`, `corruptedIndexes []` - it does not detect a dangling LSM entry, so it is not the repair path here either

This is not a regression the PR introduces and not one `cleanUpBeforePhysicalDelete` can close: nothing can learn a key it cannot deserialize, and `deleteRecordNoLock` degrades identically on the same record by design (that is what `Issue4420TolerantDeleteTest` pins on the delete path). So rather than assert an outcome that does not happen, `movingAnEdgeWithAnUnreadablePropertyIsBestEffortLikeTheDeletePath` records the boundary: the record is still removed, the counter stays truthful, no key is fabricated for the replacement, and the stale entry survives. If a future change can recover the key - or decides to fail the move instead - that test fails and says so. The javadoc now says the same thing instead of the stronger claim it made before.

**3 - comment density. Acknowledged, no change.** Fair cost to flag. My reasoning for leaving it: the PR description and the issue are both read once, at merge time, and neither is in the file a future editor opens. The comment already replaced an earlier version inside `deleteEdge` that described the hazard as unfixed, and that one-line divergence from reality is what #5779 was filed to stop living only in a code comment. I would rather carry the duplication than have the site go quiet again.

Regression: 1697 tests across `com.arcadedb.graph`, `com.arcadedb.index`, `com.arcadedb.database`, `QueryTest`, `UpdateEdgeTest` - 0 failures. 11 tests in the new class, every non-trivial branch of the fix proven to fail when only that branch is reverted.

robfrank:
This was cycle 4 of 4, so I am stopping here rather than pushing again. Answering the open question with the code, and leaving the two optional items for the developer.

**The throwing-corruption family: the two reads are equivalent-on-failure, and `propertiesAsMap()` reads a strict superset.** `MutableDocument.propertiesAsMap()` is `checkForLazyLoadingProperties()` followed by the whole map - it deserializes **every** property of the record through `BinarySerializer`. `DocumentIndexer.deleteDocument` reaches values through `getPropertyValue` -> `record.get(name)`, over the *indexed* properties only, which is a subset of that set, on the same bytes, through the same deserializer (the pre-image is a different instance but its `checkForLazyLoading` parses the identical buffer). So there is no corruption that throws inside the cleanup and not inside `propertiesAsMap()` at `moveEdge`'s line 804 - the cleanup cannot be the first read to fail on any byte pattern. `findExistingExternalRids` is the exception in the other direction: it catches everything itself and counts the failure for CHECK DATABASE.

That is exactly the "one-line comment confirming the two reads are equivalent-on-failure" you offered as the alternative to a test, and it belongs in the javadoc next to the existing tolerance note. **Deferred**, since adding it now would need a fifth review cycle for a comment.

**Also deferred: trimming the javadoc density.** Two reviewers have now flagged it. I gave my reasoning last cycle (the site went quiet once already, which is what #5779 was filed to stop) but that is a judgement call about house style and the maintainer should make it, not me at the end of a review loop.

For the record, on your note about not being able to run the build: the numbers in the description are from real runs in the PR's worktree - 11/11 in the new class, 1697 across `com.arcadedb.graph` + `com.arcadedb.index` + `com.arcadedb.database` + `QueryTest` + `UpdateEdgeTest`, 0 failures - and every non-trivial branch of the fix was reverted in isolation to confirm the matching test, and only the matching test, fails.

Two things this review loop found that are worth their own issues, neither fixed here:

1. `edge.save()` after `edge.set("@in", newRID)` throws `IllegalStateException: Cannot read original buffer` - `moveEdge` already saved the replacement and nulled the buffer via `updateIdentity`, so the natural embedded-API sequence is broken.
2. On the ordinary DELETE path, create-modify-delete in one transaction leaks an index entry: `newEdge(..., "code","D1"); e.set("code","D2"); e.delete()` leaves 2 entries where 1 record existed. Measured as a control while investigating the move path, which does **not** leak in that shape.

Merge stays with the developer.

Issue #6010, opened 2026-08-09
[Cypher] a GraphAnalyticalView makes the pipeline drop a WHERE a <> c inequality between two chain nodes


robfrank:
Found while working on #5757. Not caused by it, and not fixed by its PR: it is the ordinary materialization pipeline, so it shows on plain row queries, not only on counts.

## Symptom

With a `GraphAnalyticalView` registered for the database, a `WHERE <var> <> <var>` between two nodes of a MATCH chain is **not applied**. The same query answers correctly when no view is registered.

```
MATCH (a)-[:LINKS]->(b)-[:BOND]->(c) WHERE a <> c RETURN a.k AS x, c.k AS y
```

| | rows |
|---|---|
| no view | 3 (correct) |
| view over all vertex + edge types | 4 |

Labelling the anchor does not change it:

```
MATCH (a:Hub)-[:LINKS]->(b)-[:BOND]->(c) WHERE a <> c RETURN a.k AS x
```

| | rows |
|---|---|
| no view | 2 (correct) |
| view registered | 3 |

In both cases the count with the view is exactly the count of the pattern **without** the inequality, i.e. the predicate is dropped rather than mis-evaluated.

## Reproducer

Fixture (embedded, `TestHelper`-style):

```java
database.getSchema().createVertexType("Hub");
database.getSchema().createVertexType("Leaf");
database.getSchema().createEdgeType("LINKS");
database.getSchema().createEdgeType("BOND");

final MutableVertex h1 = database.newVertex("Hub").set("k", "h1").save();
final MutableVertex h2 = database.newVertex("Hub").set("k", "h2").save();
final MutableVertex l1 = database.newVertex("Leaf").set("k", "l1").save();
final MutableVertex l2 = database.newVertex("Leaf").set("k", "l2").save();
final MutableVertex l3 = database.newVertex("Leaf").set("k", "l3").save();

h1.newEdge("LINKS", l1).save();
h1.newEdge("LINKS", l2).save();
h2.newEdge("LINKS", l1).save();
l1.newEdge("LINKS", h2).save();
l3.newEdge("LINKS", h1).save();

l1.newEdge("BOND", h1).save();
l2.newEdge("BOND", h2).save();
h2.newEdge("BOND", l3).save();
```

[...]

Issue #5715, opened 2026-08-01
[Cypher] the two count push-downs disagree on their preconditions and their entry points: a dropped SKIP/LIMIT, and three shapes that scan when they need not


lvca:
Five findings about the two Cypher count push-downs, collected while closing #5686 (PR #5706) and all measured against `ee136e494`. They are filed together because they are five consequences of the same thing: `tryCreateTypeCountOptimization` and `tryOptimizeCountStar` are two independent detectors, reached from different places, agreeing on neither their preconditions nor their entry points.

Section 1 is a **wrong answer** and is the only one that is not a performance note.

Measurements are `readRecord` deltas from `database.getStats()`, which separate the two plans exactly: a push-down answers from the cached type counter or the CSR arrays and reads nothing, the ordinary pipeline reads one record per vertex. Fixture: 300 `:Big` vertices, a 3-vertex `:Q` chain with 2 `:LINKS` edges, 100 `:Lonely` vertices with no edges at all.

---

## 1. The count-star push-down discards `SKIP` and `LIMIT`

```
rows=1  MATCH (a:Q)-[:LINKS]->(b:Q) RETURN count(*) AS c            => [{"c":2}]
rows=1  MATCH (a:Q)-[:LINKS]->(b:Q) RETURN count(*) AS c LIMIT 0    => [{"c":2}]   <-- expected 0 rows
rows=1  MATCH (a:Q)-[:LINKS]->(b:Q) RETURN count(*) AS c SKIP 1     => [{"c":2}]   <-- expected 0 rows
rows=1  MATCH (a:Q)-[:LINKS]->(b:Q) RETURN count(*) AS c SKIP 5     => [{"c":2}]   <-- expected 0 rows
```

The other push-down, and the ordinary pipeline, both get it right, so the engine currently disagrees with itself:

```
rows=0  MATCH (m:Big) RETURN count(m) AS c LIMIT 0                  => []
rows=0  MATCH (m:Big) RETURN count(m) AS c SKIP 1                   => []
rows=0  MATCH (m:Big) RETURN m.k AS c LIMIT 0                       => []
```

Neo4j returns no rows for all of these: `LIMIT 0` and `SKIP 1` are applied to the one-row aggregate result.

[...]

Issue #5976, opened 2026-08-09
RecordEncryptionTest.encryption intermittently throws NoSuchElementException on MultiIterator.next()


lvca:
## Summary

`com.arcadedb.event.RecordEncryptionTest.encryption` failed once in CI's `unit-tests` job with a `NoSuchElementException` from `MultiIterator.next()`, distinct from the entropy-blocking flakiness already fixed in #3280 (that fix - a static `SECURE_RANDOM` field - is already present in the current test).

Observed on `unit-tests` job of an unrelated PR's CI run (2026-08-08/09, run 31281124979, job 93175495180, commit `1e6b667cd`):

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.209 s <<< FAILURE! -- in com.arcadedb.event.RecordEncryptionTest
[ERROR] com.arcadedb.event.RecordEncryptionTest.encryption -- Time elapsed: 0.207 s <<< ERROR!
java.util.NoSuchElementException
	at com.arcadedb.utility.MultiIterator.next(MultiIterator.java:96)
	at com.arcadedb.event.RecordEncryptionTest.lambda$encryption$1(RecordEncryptionTest.java:97)
	at com.arcadedb.database.LocalDatabase.transaction(LocalDatabase.java:1432)
	at com.arcadedb.database.LocalDatabase.transaction(LocalDatabase.java:1393)
	at com.arcadedb.event.RecordEncryptionTest.encryption(RecordEncryptionTest.java:96)
```

The failing line is:

```java
database.setTransactionIsolationLevel(Database.TRANSACTION_ISOLATION_LEVEL.REPEATABLE_READ);
database.transaction(() -> {
  final Vertex v1 = database.iterateType("BackAccount", true).next().asVertex();   // line 97: throws here
  assertThat(v1.getString("secret")).isEqualTo("Nobody must know John and Zuck are brothers");
});
```

This is a single-threaded, single-node test: one vertex is created and `.save()`d in the immediately preceding transaction (asserted via `creates.get() == 1`), then a second transaction switches to `REPEATABLE_READ` and calls `.next()` on `iterateType("BackAccount", true)` without checking `hasNext()` first - which means the iterator was unexpectedly empty despite the just-committed vertex, in a test with no concurrency at all.

## Why this looks like a real (rare) bug, not environment noise

- No threads, no HA/Raft, no network - a single embedded database in one JVM, one test method.
[...]

Issue #5758, opened 2026-08-02
[Cypher] a correlated COUNT/COLLECT body loses the count push-down, where a bound anchor makes it an O(degree) CSR read


lvca:
The "Also worth doing, larger" section of #5715, filed separately as that issue asked. Not addressed by PR #5737.

## The gap

Both count push-downs are all-or-nothing about correlation. #5686 restored them for an **uncorrelated** body written under a correlated query, and #5737 extended that to `COUNT { }` bodies, but a genuinely **correlated** body still falls back to materialization:

```cypher
MATCH (q:Q {k: 1}) RETURN COLLECT { MATCH (q)-[:LINKS]->(x:Q) RETURN count(*) } AS c
MATCH (p:Person)   RETURN p.name, COUNT { (p)-[:KNOWS]->(:Person) } AS friends
```

The second is about the most ordinary graph query there is - "each person and how many people they know" - and it materialises one row per edge per person to count them.

## Why it need not

The guard is correct as it stands: both push-downs answer from the schema and the CSR arrays and never look at the incoming rows, so answering `MATCH (q)-[:LINKS]->(x:Q) RETURN count(*)` with `q` already bound would return the count over *every* `q` in the graph. That is why `CypherExecutionPlan.seedIsRead` exists and why a body reading a seeded name is refused.

But a bound anchor does not make the count expensive - it makes it *cheaper*. `MATCH (q)-[:LINKS]->(x:Q) RETURN count(*)` with `q` bound is an O(degree) read of the CSR adjacency arrays: one `countEdges(nodeId, direction, edgeType)` when the far label is unconstrained, and a bucket-filtered neighbour walk when it is not. Neither needs a scan, and neither needs the rows.

## Shape of the fix

A seed-aware count step: instead of seeding the propagation from a label's bucket set, seed it from the RIDs the outer row already bound, and propagate as it does today. Concretely that means

- a `CountOp` variant, or a parameter on the existing ones, that takes the anchor node id(s) rather than deriving them from a label;
- `CypherReferencedVariables` (from #5686) telling the planner *which* names the body reads, not just whether it reads any - the read names are exactly the positions that should be seeded rather than enumerated;
- the correlation guard narrowing from "the body reads a seeded name" to "the body reads a seeded name at a position the operator cannot seed from".

This is strictly more than #5686 recovered: that restored the push-down for bodies that ignore the outer row, this would keep it for bodies that use it.

## Worth knowing before starting

[...]

PR #6012, opened 2026-08-09
fix(engine) #5757: the Cypher count push-downs anchor on an unlabelled node, and buildValidBuckets stops overloading null

### LINKED ISSUE #5757: [Cypher] the count push-downs cannot anchor on an unlabelled node, and buildValidBuckets still returns null for two different things
Follow-up to #5715 (PR #5737, merged `a997e5a82`). Two shapes that are now answered **correctly** but by the ordinary materialization pipeline, where a push-down could serve them, plus the latent trap that caused the first one.

## 1. A chain with an unlabelled anchor cannot use the push-down at all

```cypher
MATCH (a)-[:LINKS]->(b) RETURN count(*)
MATCH (a)-[:LINKS]->(b:Q) RETURN count(*)
```

Before #5737 these answered **0** for a pattern that matches. They are correct now, but the way they were made correct is `CountOp.canEnumerateAnchors()`, which declines to build the operator and lets the ordinary pipeline answer instead. So the count is right and the fast path is gone.

`MATCH ()-[:TYPE]->() RETURN count(*)` - "how many edges of this type are there" - is exactly the sort of question the CSR operators exist for, and it is the one shape they cannot serve.

**Why.** Every operator here walks out from one labelled position and enumerates that label's buckets. `PropagateChainOp.execute` seeds `current[v] = 1` only for `v` in the anchor bucket set, and `executeOLTP` iterates `db.iterateType(anchorLabel, true)`. With no anchor label there is no set to enumerate, so each of them reads the missing set as an empty one.

**Fix.** Teach the operators that an absent anchor label means *every vertex*: on the CSR path that is `anchorBuckets == null` seeding every node, and on the OLTP path an iteration over every vertex type in the schema. There are roughly 12 sites across `PropagateChainOp` and `AntiJoinChainOp` (`execute`, `executePerSourceInequality`, `countSelfLoopPaths`, `countSelfLoopPerAnchorFallback`, `executeOLTP`, `countSelfLoopPathsOLTP`, and the anti-join equivalents), all reading `anchorLabel == null` as `return 0`. #5737 deliberately did not touch them - the correctness fix was one guard, and rewriting hot LDBC code was not worth the risk in a bug-fix PR - but they are the actual work here.

## 2. The root cause is still there: one `null` meaning two things

`CSRCountUtils.buildValidBuckets(db, label)` returns `null` both for

- **`label == null`** - no label was given, so do not filter - and
- **`!existsType(label)`** - the label is declared but does not exist, so nothing matches.

Its callers disagree about which one they were handed, and that disagreement produced *both* wrong answers found in #5715:

| | read it as | result |
|---|---|---|
| anchor seeding | "no anchors" | `MATCH (a)-[:LINKS]->(b) RETURN count(*)` → 0 |
| hop filtering | "no filter" | `MATCH (a:Q)-[:LINKS]->(b:NoSuchLabel) RETURN count(*)` → unfiltered |

#5737 closed both **reachable** paths - the first by not building the operator, the second by answering 0 in the detector when a mandatory pattern element is undeclared or empty - but the helper still overloads `null`, and the `anchorLabel == null → return 0` sites are still written that way. A new operator, or a new construction path reaching an existing one, reintroduces the same class of bug with nothing to catch it.

**Fix.** Make the two cases distinguishable: `null` for "no filter", an **empty** set for "matches nothing", and have `filterByBuckets` zero everything for the empty set rather than treating it as no-filter. `PairHashJoinOp` already reads an empty build-bucket set as `return 0`, so that half is consistent with the proposal.

## 3. `ORDER BY` disqualifies the fast path even where it cannot reorder anything

```cypher
MATCH (m:Big) RETURN count(m) AS c ORDER BY c     -- scans
```

A push-down produces exactly one row, so an `ORDER BY` over it is a no-op, but `isMatchReturnOnlyStatement()` rejects the statement outright and the ordinary pipeline answers it. #5737 chose that deliberately - sorting by an aggregate alias is the pipeline's business and the answer has to be right before it is fast - and `anOrderedCountIsStillCorrect` pins the answer. Accepting an `ORDER BY` whose every item resolves to the projected count alias would recover it.

Not a wrong answer in any of the three. Sections 1 and 2 are the ones worth doing; 2 is the one that stops this recurring.

robfrank:
Closes #5757

Follow-up to #5715 (PR #5737). Sections 1 and 2 of the issue - the ones it calls "worth doing" - are implemented here. Section 3 (`ORDER BY` over a single-row aggregate) is deliberately left alone; see the last section below.

## 1. An unlabelled anchor is now enumerated rather than declined

Every count-push-down operator walks out from one position of the pattern and enumerates the vertices that position accepts, and it built that set from the position's label. An unlabelled anchor left it with no set, which each operator read as an empty one and answered **0**. #5715 closed that wrong answer with `CountOp.canEnumerateAnchors()`, which declines to build the operator so the ordinary pipeline answers instead: right, but `MATCH ()-[:TYPE]->() RETURN count(*)` - the cheapest question there is to ask a graph, and exactly what the CSR operators exist for - was the one shape the fast path could not serve.

`PropagateChainOp` and `AntiJoinChainOp` now read an absent anchor label for what it means, **every vertex**, at all the sites the issue enumerates:

- **CSR path.** `PropagateChainOp.execute` seeds the whole node domain when `validBuckets[0] == null`; `executePerSourceInequality`, `countSelfLoopPaths`, `countSelfLoop3HopEdgeScan` and `countSelfLoopPerAnchorFallback` iterate every node instead of `return 0`. `AntiJoinChainOp.execute` and `executeEdgeScanAlgebraic` likewise.
- **OLTP path.** `executeOLTP`, `countSelfLoopPathsOLTP` and `executeOLTPRecursive` go through a new `CSRCountUtils.iterateAnchors(db, label)`, which iterates every declared vertex type non-polymorphically when the label is null (so no vertex is visited twice), the label's own polymorphic iterator when it is not, and nothing when the label is not declared.

`countSelfLoopPaths` also stops enumerating its anchors with `db.iterateType` and enumerates them off the CSR bucket ids like every other site in that class, so the self-loop subtraction no longer reads records on the CSR path.

`PairHashJoinOp` and `DegreeProductOp` still decline an unlabelled anchor - they key a hash join and a degree product on the label itself - and the ordinary pipeline answers those. `canEnumerateAnchors()` stays for them; its javadoc now says which operators answer it how, and why.

## 2. `buildValidBuckets` no longer says two things with one `null`

`CSRCountUtils.buildValidBuckets(db, label)` returned `null` both for **"no label was given, so do not filter"** and for **"the label is declared on nothing, so nothing matches"**, and its callers disagreed about which one they held. That single overload produced *both* wrong answers found in #5715, one from each end of it: anchor seeding read it as "no anchors" and answered 0 for a matching pattern, hop filtering read it as "no filter" and counted an unfiltered chain.

The two cases are now distinguishable at the source rather than at each of the dozen sites that consume it:

- `null` - no label, no filter;
- an **empty** set - a label that keeps nothing.

`filterByBuckets` (both overloads) zeroes every count for an empty set instead of returning early, and every frontier filter in `PropagateChainOp`, `AntiJoinChainOp`, `PairHashJoinOp` and `CSRCountUtils.walkArm` branches on `null` alone, letting an empty set fall through to a membership test that keeps nothing. Six hand-rolled copies of "build this label's bucket ids" across the OLTP paths were replaced by calls to the helper, so they inherit the distinction rather than re-introducing the overload.

Both #5715 wrong answers remain unreachable end to end - `mandatoryPatternElementIsEmpty` answers 0 before the operator is built - so this half is pinned by a direct unit test of the helper's contract rather than by a query, which is what the issue asks for: "a new operator, or a new construction path reaching an existing one, reintroduces the same class of bug with nothing to catch it."

## 3. A view holding some of the vertices is not asked "how many vertices are there"

Enabling the unlabelled anchor opened a question that could not arise before: "every vertex" is a claim about the **graph**, and a `GraphAnalyticalView` built over a subset of the vertex types cannot make it - its node domain is a subset of the anchors. `CountOp.requiresFullVertexCoverage()` (true exactly when the anchor is unlabelled) makes `CSRCountStep` fall back to the OLTP path unless `provider.coversVertexType(null)`.

The same reasoning turned out to apply to the per-vertex accelerator the OLTP paths themselves look up, and there it was already a wrong answer for a labelled anchor too: they consult the provider per vertex and fall back to the edge list for a vertex the provider does **not** map, but nothing covers a vertex *inside* the domain whose neighbors are outside it - `getNeighborIds` answers with the adjacency the view holds, and the edges leading out of the view are simply not in it. The test in this PR reproduces exactly that (a `Hub`-only view answered 2 instead of 5). `PropagateChainOp.executeOLTP`, `AntiJoinChainOp.buildNeighborRIDMap` and `AntiJoinChainOp.buildTailCounts` now go through `CSRCountUtils.findAcceleratingProvider`, which declines a partial view.

## Test plan

New `CypherCountPushDownUnlabelledAnchorIssue5757Test` - 11 tests. Every count is cross-checked against the row count the ordinary pipeline produces for the same pattern projected without an aggregate, and the chains use a distinct edge type per hop so that relationship isomorphism (which the push-downs do not model and the pipeline does) cannot make the two disagree for an unrelated reason.

- [x] `anUnlabelledAnchorIsClaimedByTheChainPushDown` - `EXPLAIN` names `COUNT CHAIN` for the shapes #5715 sent to the pipeline.
- [x] Single-hop (`OUT`, `IN`, `BOTH`; label on neither end, one end, both), multi-hop, inequality (2-hop per-source and 3-hop edge-scan self-loop subtraction), anti-join with and without an inequality.
- [x] `theCsrPathAgreesWithTheOltpPathOnEveryUnlabelledShape` - the same ten patterns with and without a full view, asserting a ready provider is actually found first so the comparison is not the OLTP path against itself.
- [x] `aViewThatHoldsSomeOfTheVerticesDoesNotAnswerAnUnlabelledAnchor` - a `Hub`-only view is offered to the operator and declined.
- [x] `buildValidBucketsSeparatesNoFilterFromMatchesNothing` and `filterByBucketsKeepsNothingForAnEmptySet` - the helper contract directly.
- [x] **Armed:** with the `src/main` changes stashed, 5 of the 11 fail (the two helper-contract tests, the `EXPLAIN` test, the partial-view test, and the CSR/OLTP agreement test).
- [x] **No regressions:** the full `engine` unit suite is green - 11635 tests, 0 failures, 0 errors - including `CypherCountPushDownPreconditionsIssue5715Test`, `GAVEligibilityTest` and `MatchGAVExpandIntoTest`.
- [x] **openCypher TCK** (its own lane, `-Dsurefire.includes=**/*Suite.java`): 3897 scenarios, 0 failures, 85 skipped - unchanged.

## Deliberately not done

**Section 3 of the issue (`ORDER BY`).** The issue itself rates it lowest ("Sections 1 and 2 are the ones worth doing"), and #5737's reasoning still holds: sorting by an aggregate alias is the ordinary pipeline's business, and `anOrderedCountIsStillCorrect` pins the answer. Accepting an `ORDER BY` whose every item resolves to the projected count alias remains available as a separate, purely-performance change.

**The pos0/pos3 filter in the 3-hop self-loop scans.** `PropagateChainOp.countSelfLoop3HopEdgeScan` and `AntiJoinChainOp.executeEdgeScanAlgebraic` apply the labels at positions 1 and 2 but not at 0 and 3, so a *labelled* anchor can over-subtract when a node reachable at the closing hop is not of the anchor's type. That is pre-existing and untouched here: with an unlabelled anchor there is no filter to apply, so this PR neither depends on it nor makes it worse. It wants its own reproducer and its own PR.

**The same partial-view hazard in sibling operators.** `CountEdgesStep`, `CountEdgesReturnStep`, `CountChainedEdgesStep`, `PartitionedTriangleOp` and `MatchRelationshipStep` all have the identical `nodeId >= 0 ? provider... : vertex...` shape and the identical exposure. Only the two chain operators are changed here, because only they are covered by this PR's test; the others need their own.

## Found on the way, filed separately

**#6010 - a `GraphAnalyticalView` makes the pipeline drop a `WHERE a <> c` inequality.** Not a count push-down bug and not caused by this change: with a view registered, `MATCH (a)-[:LINKS]->(b)-[:BOND]->(c) WHERE a <> c RETURN a.k` returns 4 rows where the same query without a view returns 3, and labelling the anchor does not change it. It surfaced here because the pre-change code answered the count for that pattern *through* the pipeline; with this PR the push-down answers it itself and gets it right, which masks the symptom for counts and leaves it everywhere else. Full reproducer in the issue.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -458,3 +458,35 @@ failure mode is safe to ignore (the socket is already broken/closed) - the new `
 documents the same reasoning.
 
 [#5978](https://github.com/ArcadeData/arcadedb/issues/5978)
+
+## The Cypher count push-downs can anchor on an unlabelled node (#5757)
+
+`MATCH ()-[:TYPE]->() RETURN count(*)` - "how many edges of this type are there" - is the cheapest question
+there is to ask a graph, and it was the one shape the CSR count push-downs could not serve. Every operator
+walks out from one position of the pattern and enumerates the vertices that position accepts, which it built
+from the position's label; an unlabelled anchor left it with no set, which each operator read as an empty one
+and answered **0**. Issue #5715 closed that wrong answer by declining to build the operator at all, so the
+answer became right and the fast path went away.
+
+The two chain operators now read an absent anchor label for what it means - **every vertex** - on both paths:
+the CSR one seeds the whole node domain, and the OLTP one iterates every vertex type in the schema. That
+covers the plain chain, the chain with a `<>` inequality (whose self-loop subtraction anchors the same way),
+and the anti-join chain. The pair-join and degree-product operators still decline an unlabelled anchor, since
+they key a hash join and a degree product on the label itself, and the ordinary pipeline answers those.
+
+**The root cause is separated at the source.** `CSRCountUtils.buildValidBuckets` returned `null` both for "no
+label was given, so do not filter" and for "the label is declared on nothing, so nothing matches", and its
+callers disagreed about which one they held - that single overload produced *both* wrong answers found in
+#5715, one from each end of it. `null` now means only "no filter", an **empty** set means "matches nothing",
+and every consumer branches on `null` alone so an empty set falls through to a membership test that keeps
+nothing.
+
+**A view over some of the vertex types is no longer used where it cannot answer.** "Every vertex" is a claim
+about the graph rather than about a `GraphAnalyticalView`, so an operator anchored on it runs against the OLTP
+path unless the view's node domain is every vertex. The same reasoning applies to the per-vertex accelerator
+these OLTP paths consult: a view holding a subset of the vertex types answers for the vertices it maps with
+the adjacency it holds, which is missing every edge that leaves the view, and the fallback for an unmapped
+vertex does not cover it. Those lookups now go through `CSRCountUtils.findAcceleratingProvider`, which
+declines a partial view.
+
+[#5757](https://github.com/ArcadeData/arcadedb/issues/5757)
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AntiJoinChainOp.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AntiJoinChainOp.java
@@ -22,7 +22,6 @@ import com.arcadedb.database.Database;
 import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.graph.GraphTraversalProvider;
-import com.arcadedb.graph.GraphTraversalProviderRegistry;
 import com.arcadedb.graph.NeighborView;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.utility.IntHashSet;
@@ -103,10 +102,13 @@ public final class AntiJoinChainOp implements CountOp {
     return types.toArray(new String[0]);
   }
 
-  /** Every path here is enumerated from chain position 0, so an unlabelled first node leaves nothing to start from. */
+  /**
+   * Every path here is enumerated from chain position 0. With no label there the anchors are every vertex, which
+   * only a provider whose node domain <i>is</i> every vertex can enumerate (issue #5757).
+   */
   @Override
-  public boolean canEnumerateAnchors() {
-    return nodeLabels[0] != null;
+  public boolean requiresFullVertexCoverage() {
+    return nodeLabels[0] == null;
   }
 
   @Override
@@ -114,7 +116,7 @@ public final class AntiJoinChainOp implements CountOp {
     final int nodeCount = provider.getNodeCount();
     final int hops = edgeTypes.length;
 
-    // Pre-compute valid bucket sets for type filtering
+    // Pre-compute valid bucket sets for type filtering. null = unlabelled, so no filter.
     final IntHashSet[] validBuckets = new IntHashSet[hops + 1];
     for (int i = 0; i <= hops; i++)
       validBuckets[i] = CSRCountUtils.buildValidBuckets(db, nodeLabels[i]);
@@ -152,9 +154,9 @@ public final class AntiJoinChainOp implements CountOp {
         return result;
     }
 
-    // Per-source iteration from anchor (position 0)
-    final String anchorLabel = nodeLabels[0];
-    if (anchorLabel == null || !db.getSchema().existsType(anchorLabel))
+    // Per-source iteration from anchor (position 0). null accepts every vertex, empty accepts none (issue #5757).
+    final IntHashSet anchorBuckets = validBuckets[0];
+    if (anchorBuckets != null && anchorBuckets.isEmpty())
       return 0;
 
     final boolean anchorIsSource = antiJoinSourceIdx == 0;
@@ -164,18 +166,14 @@ public final class AntiJoinChainOp implements CountOp {
     for (int h = 0; h < laterIdx; h++)
       hopViews[h] = provider.getNeighborView(directions[h], edgeTypes[h]);
 
-    // Pre-compute bucket IDs for CSR-based anchor iteration and frontier filtering
-    final int[] bucketIds = new int[nodeCount];
-    for (int v = 0; v < nodeCount; v++)
-      bucketIds[v] = provider.getRID(v).getBucketId();
-    final IntHashSet anchorBuckets = CSRCountUtils.buildValidBuckets(db, anchorLabel);
-    if (anchorBuckets == null || anchorBuckets.isEmpty())
-      return 0;
+    // Pre-compute bucket IDs for CSR-based anchor iteration and frontier filtering. A pattern that labels no
+    // position needs none of them.
+    final int[] bucketIds = anyLabelled(validBuckets) ? precomputeBucketIds(provider, nodeCount) : null;
 
     long totalCount = 0;
 
     for (int anchorId = 0; anchorId < nodeCount; anchorId++) {
-      if (!anchorBuckets.contains(bucketIds[anchorId]))
+      if (anchorBuckets != null && !anchorBuckets.contains(bucketIds[anchorId]))
         continue;
 
       // Pre-compute anti-join neighbors for Case A (anchor is the anti-join source)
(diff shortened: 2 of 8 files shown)


robfrank:
@claude please review this PR.

robfrank:
### CI note: the two red jobs are both red on `main` too

`pull_request` workflows did not fire when this PR was opened - only the CodeQL dynamic run did, while a sibling PR opened eight minutes later got all three. `Java CI - test` was therefore dispatched manually on the branch: [run 31324604842](https://github.com/ArcadeData/arcadedb/actions/runs/31324604842).

Green there: `build-and-package`, `integration-tests`, **`opencypher-tck-tests`**, `slow-unit-tests`, `vector-unit-tests`, `builder-tests`, and every e2e lane (java, js, studio, go, csharp, python). Two jobs are red, and neither has a path to this diff, which is confined to `com.arcadedb.query.opencypher.executor.steps`.

**`unit-tests`** - one error, `com.arcadedb.event.RecordEncryptionTest.encryption`:

```
java.util.NoSuchElementException
	at com.arcadedb.utility.MultiIterator.next(MultiIterator.java:96)
	at com.arcadedb.event.RecordEncryptionTest.lambda$encryption$1(RecordEncryptionTest.java:97)
```

That is **#5976** verbatim - same test, same exception, same frame - filed from an unrelated PR's CI run and closed earlier today, with the fix not yet on `main`. The test exercises `database.iterateType` under a record-encryption listener; no Cypher, no count push-down. It passes on this branch locally, four consecutive runs.

**`ha-integration-tests`** - four failures, all in `com.arcadedb.server.ha.raft.*` (`Issue5569SlotMergeDeleteRaftIT`, `RaftPriorityRejoinIT`, `Issue5492SchemaWalNotShippedIT`, `BootstrapElectionIT`), in the `server` module.

**Both lanes are already red on `main`.** The last three `Java CI - test` runs on `main` (`e1aa64203`, `dcd24dd42`, `48c6824af`) all fail `ha-integration-tests` - with a *different* Raft IT each time, which is the signature of flakiness rather than a fixed break - and the most recent of them also fails `unit-tests`, on `Issue4960LSMCloseDuringCompactionTest`. So both jobs fail on the base this branch was cut from, and they fail there on tests this PR does not touch.

Local verification on this branch: full `engine` unit suite 11635 tests / 0 failures / 0 errors, and the openCypher TCK 3897 scenarios / 0 failures / 85 skipped.

robfrank:
Rebased onto \`origin/main\` (was DIRTY/CONFLICTING) to resolve a conflict with #6009 (\`fix(#5758): seed the Cypher chain count push-down from a correlated body's bound anchor\`), which landed on `main` after this branch and also touches `PropagateChainOp.java`.

**Conflicts resolved (2 files):**
- `docs/release-26.9.1.md`: both branches appended a new release-note section; kept both, in sequence.
- `engine/.../steps/PropagateChainOp.java`: 3 hunks, all in logic that both PRs independently modified:
  1. `canEnumerateAnchors()` (main, #5758's anchorRid-aware override) vs `requiresFullVertexCoverage()` (this PR's new interface method). Resolved by dropping the `canEnumerateAnchors()` override entirely (the interface default of `true` is now correct for every case this operator handles, matching the sibling `AntiJoinChainOp`) and keeping `requiresFullVertexCoverage()` gated on `nodeLabels[0] == null`, since the seeded-anchor path also needs a provider mapping the bound vertex.
  2. `executeOLTP`'s accelerator lookup: kept this PR's `CSRCountUtils.findAcceleratingProvider(db, edgeTypes)`, which correctly declines a partial-coverage provider (main's side still called the plain registry lookup).
  3. Anchor seeding in `executeOLTP`: combined both - preserved #5758's `anchorRid`-seeded single-vertex branch (with its `anchorMatchesLabel` filter) and replaced the labelled-only `else` branch with this PR's `CSRCountUtils.iterateAnchors(db, nodeLabels[0])`, which also handles the unlabelled (every-vertex) case.

**Verification:**
- `mvn -pl engine -am install -DskipTests` compiles clean.
- `CypherCountPushDownUnlabelledAnchorIssue5757Test`: 11/11 pass.
- `CypherCorrelatedCountPushDownIssue5758Test` (the #5758 regression suite, to confirm the rebase didn't regress the sibling fix): 17/17 pass.
- Broader sweep of 18 Cypher count-push-down test classes (`Cypher*Count*Test`, `Issue5071/5084/5094/5140/5165/5166/5463/5790*`, `OpenCypherCountOptimizationTest`, `CountEdgesOptimizationTest`): all green, 0 failures/errors.

Pushed as `958c71524` via `--force-with-lease`. `gh pr view` now reports `mergeable: MERGEABLE`.

PR #6016, opened 2026-08-09
fix: BucketIterator.fetchNext() off-by-one and exception swallowing (#6014, #6015)

### LINKED ISSUE #6014: BucketIterator.fetchNext() reads one page past the bucket's valid range (off-by-one)
## Summary

Found while investigating #5976 (fixed in #6013), not itself the root cause of that issue, but a real, independent bug worth its own fix.

`BucketIterator.fetchNext()`'s forward-scan termination check is off by one:

```java
// engine/src/main/java/com/arcadedb/engine/BucketIterator.java, fetchNext()
if (forwardDirection) {
  // MOVE FORWARD
  if (nextPageNumber > totalPages)
    return null;
}
```

`totalPages` is a page *count* (e.g. `1` means only page `0` is valid), so the loop should stop once `nextPageNumber >= totalPages`. Using `>` means that when `nextPageNumber == totalPages` (one past the last real page), the check passes and the iterator goes on to fetch a page that does not exist.

## Effect

That out-of-range fetch goes through `TransactionContext.getPage()` → `PageManager.getImmutablePage(pageId, size, false, true)` with `createIfNotExists=true`, so instead of returning `null` or throwing, it synthesizes a blank (zero-filled) `CachedPage` for a `PageId` that does not correspond to any real page - and that phantom page then gets inserted into `PageManager`'s global `readCache`.

This isn't a correctness bug for iteration itself (the synthesized page correctly reads as 0 records, so no wrong data is returned), but it means:
- every single forward bucket iteration does one wasted "fetch a nonexistent page" round-trip through the page manager, and
- the read cache accumulates phantom entries for pages that were never written, which seems unintended and could interact badly with other code that treats "present in the cache" as a signal (worth auditing once this is understood/fixed).

The backward-direction check (`nextPageNumber < 0`) does not have the equivalent problem.

## Suggested fix

```java
if (nextPageNumber >= totalPages)
  return null;
```

Should come with a regression test asserting `PageManager`'s read-cache does not gain an entry for `PageId(bucket, totalPages)` after a full bucket iteration.

### LINKED ISSUE #6015: BucketIterator.fetchNext() silently swallows exceptions while materializing a record
## Summary

Directly implicated in why #5976 surfaced as a confusing `NoSuchElementException` instead of the real underlying error, and worth fixing independently of that specific bug now that it's closed (#6013).

```java
// engine/src/main/java/com/arcadedb/engine/BucketIterator.java, fetchNext()
try {
  ...
  nextBatch[writeIndex++] = database.lookupByRID(rid, false);
  ...
} catch (final Exception e) {
  final String msg = "Error on loading record #%d:%d (error: %s)".formatted(...);
  LogManager.instance().log(this, Level.SEVERE, msg);
} finally {
  if (forwardDirection) currentRecordInPage++; else currentRecordInPage--;
}
```

Any exception raised while materializing a record - deserialization failure, a broken `AfterRecordReadListener`/trigger, a corrupted record, the #5976 race that was just fixed, or any *future* bug in that path - is caught, logged at `SEVERE`, and the record is silently dropped from the batch. The record just disappears from the result set; the caller (and `MultiIterator`, and anything downstream) cannot tell the difference between "this bucket legitimately has fewer records" and "N records failed to load and were skipped."

That's exactly the shape that turned #5976's real exception (`IllegalArgumentException` from a re-entrant double-decrypt) into a misleading "the iterator is empty" `NoSuchElementException` several layers up, and made the underlying bug much harder to diagnose than it needed to be. The specific #5976 cause is fixed, but this generic swallow-and-continue will do the same thing to any *other* exception that reaches this catch block, present or future.

## Suggested direction (needs a design decision, not just a fix)

Some options, roughly from least to most invasive:
1. At minimum, track how many records were skipped this way and expose it (e.g. on the iterator, or via a counter/stat), so a caller doing correctness-sensitive iteration (like `CHECK DATABASE`, or a `COUNT`) can detect that the result is incomplete instead of silently trusting it.
2. Add a strict/tolerant mode: default behavior stays "skip and log" for resilience during things like a full scan over a large bucket where one corrupt record shouldn't abort the whole read, but callers that need correctness guarantees (e.g. transactional reads under `REPEATABLE_READ`, where a "missing" record is a much stronger signal than during a best-effort admin scan) can opt into "propagate the first exception instead of swallowing it."
3. At least reclassify: swallow-and-log is arguably fine for a genuinely corrupt/undeletable record on-disk (a `CHECK DATABASE`-shaped problem), but is a poor default for a *transient* failure like a `NeedRetryException`-shaped conflict or a listener bug, which should probably propagate so the caller's normal retry/error-handling path deals with it rather than silently returning a truncated result set.

Filing as a discussion/design issue rather than a PR since the right behavior depends on call-site expectations across `iterateType`/`iterateBucket`/`CHECK DATABASE`/index rebuild, which should be surveyed before picking a default.

lvca:
## Summary

Fixes two independent bugs in `BucketIterator.fetchNext()`, both found while investigating #5976 (fixed in #6013).

**#6014 - off-by-one past the bucket's valid range**

The forward-scan termination check used `nextPageNumber > totalPages` instead of `>=`. `totalPages` is a page *count*, so when `nextPageNumber == totalPages` (one past the last real page) the check let the iterator fetch a page that does not exist. `TransactionContext.getPage()` delegates to `PageManager.getImmutablePage(..., createIfNotExists=true)`, which silently synthesizes a blank page for that nonexistent `PageId` and caches it in `PageManager`'s global read cache - a wasted fetch on every forward bucket iteration, plus a phantom cache entry.

Fix: `nextPageNumber >= totalPages`.

**#6015 - exceptions during record materialization were silently swallowed**

The `catch (final Exception e)` block around record materialization caught everything, logged it at `SEVERE`, and dropped the record from the batch. The caller couldn't tell "this bucket legitimately has fewer records" from "N records failed to load and were skipped" - exactly what turned #5976's real `IllegalArgumentException` (a listener re-entrancy bug) into a misleading `NoSuchElementException` several layers up.

I surveyed the call sites and the two exception shapes that can actually reach that catch block:
- `NeedRetryException`/`ConcurrentModificationException` (the issue's suggested "propagate transient/retryable" case) cannot actually occur on a plain read - `PageManager.checkPageVersion()` only runs at commit time. That specific proposal doesn't have a real target.
- `RecordNotFoundException` is a genuinely benign race (record concurrently deleted between the `existsRecord()` check and the lookup a couple of lines later) - already how the rest of this loop treats "turned out to be gone".
- `SerializationException` is the project's own signal for a corrupt-on-disk record - a `CHECK DATABASE`-shaped problem where skipping one bad record and continuing the scan is the right call.
- Everything else (arbitrary `RuntimeException`, e.g. from a user-supplied `AfterRecordReadListener`/trigger, or any future bug) is not a case this iterator can safely reason about generically, so it now propagates instead of being hidden. This matches the project's own prior fix for the same shape of problem in #5656 (`EXISTS{}` swallowing hid a retryable exception).

Added `getSkippedRecordCount()` so a correctness-sensitive caller can detect the corrupt-record-skip case after exhausting the iterator.

## Testing

- `BucketIteratorPhantomPageTest` (new): creates a single-page bucket, exhausts a forward iteration, asserts `PageManager`'s read cache never gains an entry for `PageId(bucket, totalPages)`. Fails before the fix, passes after.
- `Issue6015BucketIteratorSwallowedExceptionTest` (new): registers an `AfterRecordReadListener` that throws `IllegalStateException`, scans the type under `REPEATABLE_READ` (forcing eager content load, same shape as #5976), and asserts the exception now propagates instead of the scan silently looking empty (`NoSuchElementException` from `MultiIterator`, reproduced against the pre-fix code before applying the fix). Fails before the fix, passes after.
- Ran `Issue5976AfterReadReentrancyRaceTest` (the existing regression test for the bug this exact swallow-and-log behavior used to hide) plus the `CheckDatabase*` suite, `RecordEncryptionTest`, `LSMTreeFullTextIndexTest`, and the `Select*`/scan-execution test suites in the `engine` module - all green.

## Test plan

- [x] `mvn -pl engine -am compile` passes
- [x] New regression tests reproduce both bugs against the pre-fix code and pass against the fix
- [x] `Issue5976AfterReadReentrancyRaceTest` still passes (the narrowed catch doesn't regress the prior fix)
- [x] `CheckDatabase*`, `RecordEncryptionTest`, `LSMTreeFullTextIndexTest`, `Select*` execution suites all green

Closes #6014, closes #6015.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/BucketIterator.java
+++ b/engine/src/main/java/com/arcadedb/engine/BucketIterator.java
@@ -150,13 +150,30 @@ public class BucketIterator implements Iterator<Record> {
             (!forwardDirection && currentRecordInPage > -1)
         ) {
           try {
-            final int recordPositionInPage = (int) currentPage.readUnsignedInt(
-                LocalBucket.PAGE_RECORD_TABLE_OFFSET + currentRecordInPage * INT_SERIALIZED_SIZE);
-            if (recordPositionInPage == 0)
-              // CLEANED CORRUPTED RECORD
-              continue;
+            final int recordPositionInPage;
+            final long[] recordSize;
+            try {
+              recordPositionInPage = (int) currentPage.readUnsignedInt(
+                  LocalBucket.PAGE_RECORD_TABLE_OFFSET + currentRecordInPage * INT_SERIALIZED_SIZE);
+              if (recordPositionInPage == 0)
+                // CLEANED CORRUPTED RECORD
+                continue;
 
-            final long[] recordSize = currentPage.readNumberAndSize(recordPositionInPage);
+              recordSize = currentPage.readNumberAndSize(recordPositionInPage);
+            } catch (final RuntimeException e) {
+              // CORRUPTED SLOT-TABLE ENTRY OR RECORD HEADER: these two calls only ever touch raw page bytes (no
+              // application/listener code runs here), so an out-of-bounds read (IndexOutOfBoundsException /
+              // IllegalArgumentException / BufferUnderflowException depending on which Binary accessor caught it
+              // first) is provably page corruption, the same "bad slot" signal as the RECORD_POSITION==0 case
+              // just above. Skip it like SerializationException below, scoped narrowly to just these two reads so
+              // it cannot also catch a bug from database.lookupByRID()/AfterRecordReadListener (#6015).
+              skippedRecords++;
+              final String msg = "Error on loading record #%d:%d (error: %s)".formatted(currentPage.pageId.getFileId(),
+                  (nextPageNumber * bucket.getMaxRecordsInPage()) + currentRecordInPage, e.getMessage());
+              LogManager.instance().log(this, Level.SEVERE, msg);
+              continue;
+            }
+
             if (recordSize[0] > 0 || recordSize[0] == LocalBucket.FIRST_CHUNK) {
               // NOT DELETED
               final RID rid = new RID(bucket.fileId,
--- a/engine/src/main/java/com/arcadedb/engine/BucketIterator.java
+++ b/engine/src/main/java/com/arcadedb/engine/BucketIterator.java
@@ -22,6 +22,7 @@ import com.arcadedb.database.Binary;
 import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.RID;
 import com.arcadedb.database.Record;
+import com.arcadedb.exception.ConcurrentModificationException;
 import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.exception.SerializationException;
 import com.arcadedb.log.LogManager;
@@ -88,11 +89,13 @@ public class BucketIterator implements Iterator<Record> {
 
   /**
    * Number of records skipped so far for a known, tolerated corruption reason - a corrupted on-disk record
-   * ({@link SerializationException}) or page-layer corruption in a raw slot/pointer read, both via
+   * ({@link SerializationException}), page-layer corruption in a raw slot/pointer read, or a multi-page record
+   * whose chunk chain {@link LocalBucket#isChunkChainBroken} confirms is structurally broken - all via
    * {@link #logSkippedRecord(Exception)} in {@link #fetchNext()}. A correctness-sensitive caller (e.g. an exact
    * {@code COUNT}) can check this after exhausting the iterator to detect a truncated result; it is not
-   * incremented for the benign concurrent-delete race handled by {@link RecordNotFoundException}, nor for any
-   * other exception, which propagates instead of being counted here (#6015).
+   * incremented for the benign concurrent-delete race handled by {@link RecordNotFoundException}, nor for a
+   * {@link ConcurrentModificationException} not confirmed as a broken chain (transient contention propagates
+   * instead so a retry can resolve it), nor for any other exception, which likewise propagates (#6015).
    */
   public long getSkippedRecordCount() {
     return skippedRecords;
@@ -100,8 +103,10 @@ public class BucketIterator implements Iterator<Record> {
 
   /**
    * Counts and logs a record slot skipped for a known, tolerated reason: a corrupted on-disk record
-   * ({@link SerializationException}) or page-layer corruption in a raw slot/pointer read (an unchecked exception
-   * from a {@code BasePage.read*}/{@code Binary} accessor - see the two call sites in {@link #fetchNext()}).
+   * ({@link SerializationException}), page-layer corruption in a raw slot/pointer read (an unchecked exception
+   * from a {@code BasePage.read*}/{@code Binary} accessor), or a confirmed structurally-broken multi-page chunk
+   * chain ({@link ConcurrentModificationException} where {@link LocalBucket#isChunkChainBroken} returns
+   * {@code true}) - see the call sites in {@link #fetchNext()}.
    */
   private void logSkippedRecord(final Exception e) {
     skippedRecords++;
@@ -192,7 +197,21 @@ public class BucketIterator implements Iterator<Record> {
               if (!bucket.existsRecord(rid))
                 continue;
 
-              nextBatch[writeIndex++] = database.lookupByRID(rid, false);
+              try {
+                nextBatch[writeIndex++] = database.lookupByRID(rid, false);
+              } catch (final ConcurrentModificationException e) {
+                if (!bucket.isChunkChainBroken(rid))
+                  // TRANSIENT CONTENTION, NOT PROVEN CORRUPTION: loadMultiPageRecord() throws this after
+                  // exhausting TX_RETRIES, but exhausted retries alone do not prove the chain is broken - its
+                  // page-version validation also fails under ordinary concurrent writes to OTHER records sharing
+                  // the chain's pages. Propagate so the caller's retry machinery (if any) re-runs, the same
+                  // disambiguation LocalDatabase.deleteRecordInternal() already applies to this exception.
+                  throw e;
+                // CONFIRMED STRUCTURALLY BROKEN CHAIN: a version-blind walk (isChunkChainBroken) found a genuinely
+                // bad continuation pointer, not just a version mismatch - this is corruption, not contention.
+                writeIndex--; // UNDO THE RESERVED SLOT: THE FAILED CALL DID NOT WRITE INTO IT
+                logSkippedRecord(e);
+              }
 
             } else if (recordSize[0] == LocalBucket.RECORD_PLACEHOLDER_POINTER) {
               // PLACEHOLDER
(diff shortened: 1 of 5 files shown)


lvca:
Thanks for the review. Addressed the substantive point in b90607ed3:

- **Raw page-layer reads wrapped by the same catch as record lookups**: fixed. Scoped a narrow inner try/catch around just the `readUnsignedInt`/`readNumberAndSize` slot-table reads, since no application/listener code runs in those two lines - any exception there is provably page corruption, not an application bug. Deliberately did **not** catch `IllegalArgumentException` broadly at the outer level to fix this, since the actual #5976 bug was itself an `IllegalArgumentException` (Base64 decode failure inside a listener) - that would have silently reintroduced the exact regression this PR closes. New test `BucketIteratorCorruptedSlotTest` corrupts a slot-table entry directly and confirms the scan now skips it (counted via `getSkippedRecordCount()`) instead of aborting; confirmed it fails without the scoping fix and passes with it.

Left the two non-blocking notes as-is, per your own read that they're fine for a targeted bug-fix PR:
- `getSkippedRecordCount()` has no caller yet - agreed it's reasonable groundwork for a correctness-sensitive caller (e.g. an exact `COUNT`) to consume later, but wiring that up is out of scope here.
- The `SEVERE` log message format is unchanged, as noted.

lvca:
Thanks, good catch. Fixed in d9c8978f0:

- **Placeholder-pointer raw read**: wrapped `currentPage.readLong(...)` (resolving the placeholder's target position) the same way as the slot-table reads - same rationale, no application/listener code runs in that call either. Extracted the now-3x-repeated "count + log" bookkeeping into a `logSkippedRecord()` helper to avoid drift between the three sites.
- **`getSkippedRecordCount()` javadoc**: updated to mention the corrupted-slot-table case too.

**On the requested test**: I tried to build `BucketIteratorCorruptedPlaceholderPointerTest` analogous to `BucketIteratorCorruptedSlotTest`, but couldn't get it to reliably reproduce the corruption:
- Relocating the slot to an artificial far-away position (like the slot-table test does) gets "healed" - `LocalBucket.compressPage()` runs unconditionally on every commit and defragments what it sees as a hole between the real record and the relocated one, reverting the slot pointer.
- Corrupting the placeholder in place (no relocation, so no hole) needs the corrupted footprint (marker + 8-byte pointer) to exceed the page's content boundary while the original record still fit. I tried inflating the footprint with a deliberately over-long/denormalized varint encoding of -1, but the margin it needs to overflow is smaller than the extra slack `LocalBucket`'s own record-insertion space-reservation logic requires for the *original* two-record page to fit in the first place - so for record sizes small enough to keep a unit test simple, there's no page size where both conditions hold.

Given the fix is structurally identical to the already-tested slot-table case (same shape, now the same `logSkippedRecord()` helper), I'm relying on that symmetry rather than continuing to fight the page-compaction internals for marginal additional coverage. Happy to take another pass at it if you see an approach I'm missing.

lvca:
Great catch, and thanks for pointing at the precedent. Fixed in 2a309cb69:

- **`ConcurrentModificationException` disambiguation**: applied the exact same pattern `LocalDatabase.deleteRecordInternal()` already uses for this exception - `LocalBucket.isChunkChainBroken(rid)` is a version-blind structural walk that only returns `true` for a genuinely broken chunk chain, never for a version mismatch caused by ordinary contention. Wrapped both call sites that can reach `loadMultiPageRecord()` (the NOT-DELETED branch's `lookupByRID` and the PLACEHOLDER branch's `getRecordInternal`): a confirmed-broken chain is now tolerated (logged, skipped, counted via `getSkippedRecordCount()`); anything else propagates so a caller's retry machinery can resolve genuine contention, instead of either always aborting the scan (what my previous commit did by omission) or always swallowing it (which would risk hiding real data loss behind "just contention").
- Corrected my own PR description's claim in the process - you're right that `loadMultiPageRecord` contradicts it; `NeedRetryException`/CME *does* have a real target on the read path, just not through `checkPageVersion()`.
- Added `BucketIteratorBrokenChunkChainTest`, adapting `BrokenMultiPageRecordDeleteTest`'s proven corruption technique (genuine multi-page vertex, corrupted `FIRST_CHUNK` continuation pointer) to prove the confirmed-broken half end-to-end: verified it fails with the raw `ConcurrentModificationException` propagating before this fix, passes after. The healthy-chain half (a transient CME must keep propagating) rides on `isChunkChainBroken`'s own existing test coverage, which this fix's catch calls verbatim.

I also went back and got a dedicated regression test working for the previous round's placeholder-pointer finding after all (`BucketIteratorCorruptedSlotTest`'s sibling for the `RECORD_PLACEHOLDER_POINTER` branch didn't make it into that commit, but `BucketIteratorBrokenChunkChainTest` here exercises that exact branch's `getRecordInternal` call path too, via the CME disambiguation).

lvca:
Correction to my last comment: the closing paragraph overstated things. \`BucketIteratorBrokenChunkChainTest\`'s broken record is a top-level \`FIRST_CHUNK\` record (not a placeholder), so it exercises the CME catch on the **NOT-DELETED branch** (\`lookupByRID\`), not the placeholder branch's \`getRecordInternal\` catch. The placeholder branch's CME catch is unit-identical code (same \`isChunkChainBroken\` call, same \`logSkippedRecord\`) but isn't separately exercised by a test - a genuinely broken chain reached *through* a placeholder would need the record to first spill into a placeholder and then have its target's chain broken, which is more setup than seemed proportionate to add on top of everything else in this round. Flagging this so it's not taken as more coverage than it is.

lvca:
Thanks - glad the CME fix and overall approach check out. Addressed the concrete ask in 2ffec5f6f:

- **JLS evaluation-order comment on `writeIndex--`**: added, explaining why the decrement is necessary even though the assignment never ran.

On the remaining non-blocking points, both reviews agree these are trade-offs rather than defects, so leaving them as-is:
- **`getSkippedRecordCount()` unused outside tests**: intentional forward-looking API per the original issue (#6015 explicitly asked for a way to detect truncation). Wiring it into `CHECK DATABASE`/`count()` would be new scope beyond what either issue asked for - happy to file a follow-up issue for that if useful, but won't bolt it onto this PR.
- **Broad `catch (RuntimeException e)` around the two raw page-byte reads could also swallow a future paging-layer bug**: acknowledged trade-off, same one the review itself calls "reasonable given the narrow scope."
- **No dedicated test for the placeholder-pointer raw-read corruption path**: explained in an earlier round (comment above) - the two approaches I tried were blocked by `compressPage()`'s hole-healing and by `LocalBucket`'s insert-time space-reservation margin, respectively. Relying on code-shape symmetry with the tested slot-table case.
- **`getRecordInternal`'s `IOException` → `DatabaseOperationException` now propagates instead of being tolerated**: intentional - consistent with this PR's overall direction of surfacing real errors instead of masking them as "fewer records."

This is round 4, and both the substantive findings across all rounds (off-by-one, two page-layer corruption scope gaps, CME disambiguation) are fixed and tested. I'll consider this ready to merge pending final CI.

lvca:
Thanks for the thorough final pass. Noting the one new observation for the record: the pre-existing \`recordPositionInPage == 0\` ("CLEANED CORRUPTED RECORD") branch not going through \`logSkippedRecord()\`/\`skippedRecords++\` is correct, but that's pre-existing behavior this PR didn't touch, not a regression introduced here - leaving it as-is rather than expanding this PR's scope. Everything else is already addressed or explicitly acknowledged as a non-blocking trade-off in the prior round.

Waiting on the remaining CI test lanes (unit-tests, integration-tests, etc.) to finish, then this should be ready to merge.

lvca:
All CI lanes are green except \`ha-integration-tests\`, which failed on two tests unrelated to this change:
- \`RaftReplicationMaterializedViewIT.materializedViewReplicates\` - \`DatabaseIsClosedException\` (server lifecycle timing)
- \`Raft3PhaseCommitIT.basicInsertReplicates\` - \`AssertionError: Server 1 should have 1 BasicDoc record but has 0\` (replication-lag timing)

Both are HA/Raft cluster replication timing flakes with no connection to \`BucketIterator\`, paging, or multi-page records - this lane has a known chronic flake history unrelated to the change here. \`unit-tests\` (which runs all four new regression tests for this PR) and every other lane, including \`slow-unit-tests\` and \`integration-tests\`, are green.

With round 5's review already a full green light ("I'd be comfortable merging this") and no unresolved blocking points, I'm marking this ready to merge.

Issue #5966, opened 2026-08-08
Indexed BETWEEN falls back to a bucket scan; three duplicate declared-type key conversions risk drift


lvca:
## Summary

Two related follow-ups discovered while fixing #5932 (PR #5961), filed together since both are query-planner/index-internals findings from the same investigation rather than active bugs.

## 1. `BETWEEN` on an indexed column never uses the index

Confirmed via `ExecutionPlan.prettyPrint()` on an indexed `INTEGER` column with correctly-typed bounds (no type mismatch at all):

```sql
CREATE VERTEX TYPE V;
CREATE PROPERTY V.n INTEGER;
CREATE INDEX ON V (n) NOTUNIQUE;
INSERT INTO V SET n = 10;
INSERT INTO V SET n = 20;
INSERT INTO V SET n = 30;

SELECT n FROM V WHERE n BETWEEN 15 AND 25
```

plans as:

```
+ FETCH FROM TYPE V WITH FILTER
  + SCAN WITH FILTER BUCKET 1 (V_0)
    n BETWEEN 15 AND 25
+ CALCULATE PROJECTIONS
  n
```

a full bucket scan with a row-level filter. The semantically identical

```sql
SELECT n FROM V WHERE n > 5 AND n < 25
```

plans as:

```
+ FETCH FROM INDEX V[n]
  n > 5 and n < 25
+ EXTRACT VALUE FROM INDEX ENTRY
  filtering buckets [1]
+ FILTER ITEMS BY TYPE 
  V
+ CALCULATE PROJECTIONS
  n
```

correctly using the index. So any `BETWEEN` query on an indexed column pays for a full scan today, regardless of how large the type or how selective the range - a pure performance regression relative to writing the equivalent two-condition `AND`, with no correctness issue either way.

Likely in `FetchFromIndexStep.processBetweenCondition()` / `SelectExecutionPlanner` - `BETWEEN` doesn't appear to be recognized as an index-range-scan-eligible shape the way an `AND` of two `BinaryCondition`s is.

## 2. Three independent copies of "narrow value to declared type + fold case"

While fixing the case-insensitive-collation gap a reviewer caught on PR #5961, tracing the transaction-overlay key representation surfaced that this exact narrowing logic (`Type.convert(...)` per key component, then lowercase a `String` component when the property's collation is case-insensitive) is now implemented independently in three places:

[...]

Issue #5410, opened 2026-07-25
HA: #4790 phase-2 ticket stays held after the abandoned entry applies, pinning Raft log compaction until restart


robfrank:
Follow-up to #5407 / PR #5408. Not a data-loss bug - an operational one introduced by the #5407 durability guard.

## Background

PR #5408 fixes #5407 by having `RaftReplicatedDatabase.commit()` take a "phase-2 ticket" from `ArcadeStateMachine` before replication. The ticket records a Raft **replay floor** and `takeSnapshot()` clamps its durability checkpoint to the lowest in-flight floor, so an entry that is Raft-committed but not yet applied locally can never be buried below the replay position.

The ticket is **retained by default** and released only where the local pages are provably settled.

## The gap

The `ReplicationDispatchedTimeoutException` branch (#4790, indeterminate replication outcome) deliberately retains its ticket: the entry may still reach quorum, and if this node crashed before applying it the retained ticket is what keeps it replayable. That part is correct and is a genuine crash-safety improvement over #4790's in-memory-only `abandonedLocalTransactions` marker.

But when the abandoned entry **does** later reach quorum and is applied by `ArcadeStateMachine.applyTxEntry` (via the `abandonedLocalTransactions` consumption path) - the *normal* resolution of a #4790 timeout - its pages become durable and **nothing releases the ticket**. The checkpoint therefore stays pinned, and Ratis stops purging the Raft log past that index, until the process restarts.

Because #4790 timeouts are transient and recoverable, this can happen on a healthy node with no crash involved. That makes it arguably heavier than the crash window it protects against.

## Impact

- Raft log is not purged past the pinned index for as long as the node stays up.
- Related to #5345 (unbounded Raft log growth until disk-full), which is exactly the failure mode this could feed.
- Bounded and visible: `takeSnapshot()` emits a throttled WARNING once a ticket has been held beyond 5 minutes, naming the pinned index and pointing at Raft-storage disk usage. A restart clears it (replay applies the entry and the in-memory ticket is gone with the process).

## Why it was not fixed in PR #5408

Releasing the ticket at the point `applyTxEntry` applies an abandoned transaction requires a **ticket-to-`walTxId` correlation that does not exist today**:

- the ticket is created in `commit()` *before* replication, when no WAL txId has been associated with it yet;
- `markTransactionAbandonedForLocalApply` knows the `walTxId` (it deserialises it from the replicated payload) but not the ticket;
[...]

PR #6018, opened 2026-08-09
fix(#5966): BETWEEN on an indexed column now uses the index; dedupe declared-type key conversion

lvca:
## Summary

Two related fixes filed together in #5966, both follow-ups from the #5932/PR #5961 investigation:

**1. `BETWEEN` on an indexed column never used the index**, falling back to a full bucket scan with a row-level filter, unlike the semantically identical `x > a AND x < b`:

- `BetweenCondition` never overrode `BooleanExpression.isIndexAware(IndexSearchInfo)`, so `SelectExecutionPlanner.buildIndexSearchDescriptor()` never recognized `field BETWEEN a AND b` as an index-range-eligible shape. Added the override, mirroring `BinaryCondition.isIndexAware()`.
- `SelectExecutionPlanner`'s "is this condition terminal for composite-index key building" check (`rangeOp`) only recognized `BinaryCondition`. Without also recognizing `BetweenCondition`, a query like `WHERE a = 1 AND b BETWEEN 10 AND 20 AND c = 1` on a composite index `(a, b, c)` would have folded `c = 1` into the same ordered-range key as an equality component after a range component - not just suboptimal, but a real correctness risk (proven by a test that fails without this half of the fix even though the plan does pick an index).
- `IndexSearchDescriptor.requiresMultipleIndexLookups()`/`.cost()` and `FetchFromIndexStep.updateIndexStats()` had the same "only `BinaryCondition` is understood" assumption, which would have forced an unnecessary `DISTINCT` step and mis-bucketed the query-cost cache for a matched `BETWEEN`. Fixed alongside since they're the same class of gap on the same code path.

`BetweenCondition.resolveKeyFrom/resolveKeyTo/isKeyFromIncluded/isKeyToIncluded` already existed and are correct as-is; they were simply unreachable from the normal planning path before this change (only from the niche `SELECT FROM INDEX:<name> WHERE key BETWEEN a AND b` target syntax).

**2. Three independent copies of "narrow value to declared type + fold case-insensitive collation"** existed: `LSMTreeIndex.convertKeys(Object[])` (insert-side, feeds the transaction overlay), `LSMTreeIndexAbstract.convertKeysToDeclaredTypes(Object[], byte[])` (query-side range bounds), and `LSMTreeIndexAbstract.convertKeys(Object[], byte[])` (disk-probe, layers a `String->byte[]` encoding on top of the second one). The first two **must** stay behaviorally identical - they are the insert side and query side of the exact comparison whose divergence caused #5932's `ClassCastException`/silent-miscompare - but nothing enforced that. Collapsed `LSMTreeIndex.convertKeys(Object[])` into a one-line delegate to `convertKeysToDeclaredTypes()`, matching how the disk-probe variant already delegates to it, so a future change to one can no longer silently diverge from the other.

## Test plan

- New regression test `Issue5966BetweenUsesIndexTest`:
  - `betweenOnIndexedColumnUsesTheIndex` - reproduces the issue's exact repro, asserts the plan contains `FETCH FROM INDEX` (not a bucket scan) and needs no row-level `FilterStep`.
  - `betweenOnIndexedColumnReturnsSameResultsAsEquivalentAndRange` - asserts `BETWEEN` and the equivalent two-sided `AND` return identical result sets, with both bounds inclusive.
  - `betweenInsideCompositeIndexStopsKeyBuildingAtTheRangeField` - composite index `(a, b, c)` with a `BETWEEN` on the middle field; asserts the index key used is only `(a, b)` and `c` is applied as a separate post-fetch filter, with correct row counts. Confirmed this test fails (wrong key/results) without the `SelectExecutionPlanner` half of the fix, even with the `isIndexAware` half alone.
- Verified all three assertions fail against the pre-fix code (reverted the fix locally, reran, confirmed the expected failures) and pass after restoring it.
- Ran the full `com.arcadedb.index.**` and `com.arcadedb.query.sql.executor.**` test packages plus the composite-index/case-insensitive-index/max-min-from-index suites: all green, no regressions from either change.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndex.java
@@ -51,7 +51,6 @@ import com.arcadedb.schema.Schema;
 import com.arcadedb.schema.Type;
 import com.arcadedb.serializer.json.JSONArray;
 import com.arcadedb.serializer.BinaryComparator;
-import com.arcadedb.serializer.BinaryTypes;
 import com.arcadedb.serializer.json.JSONObject;
 import com.arcadedb.utility.LockManager;
 import com.arcadedb.utility.RWLockContext;
@@ -63,7 +62,6 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -981,21 +979,10 @@ public class LSMTreeIndex implements RangeIndex, IndexInternal {
   }
 
   private Object[] convertKeys(final Object[] keys) {
-    if (keys != null) {
-      final byte[] keyTypes = mutable.binaryKeyTypes;
-      final boolean[] ciKeys = mutable.caseInsensitiveKeys;
-      final Object[] convertedKeys = new Object[keys.length];
-      for (int i = 0; i < keys.length; ++i) {
-        if (keys[i] == null)
-          continue;
-        convertedKeys[i] = Type.convert(getDatabase(), keys[i], BinaryTypes.getClassFromType(keyTypes[i]));
-        // Apply case-insensitive collation
-        if (convertedKeys[i] instanceof String s && ciKeys != null && i < ciKeys.length && ciKeys[i])
-          convertedKeys[i] = s.toLowerCase(Locale.ROOT);
-      }
-      return convertedKeys;
-    }
-    return null;
+    // Delegates to the query-side declared-type narrowing (insert side and query side of the exact comparison
+    // #5932 was about must never drift apart again, see #5966).
+    final LSMTreeIndexMutable m = mutable;
+    return m.convertKeysToDeclaredTypes(keys, m.binaryKeyTypes);
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromIndexStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromIndexStep.java
@@ -234,6 +234,8 @@ public class FetchFromIndexStep extends AbstractExecutionStep {
       if (lastOp instanceof BinaryCondition binaryCondition) {
         final BinaryCompareOperator op = binaryCondition.getOperator();
         range = op.isRangeOperator();
+      } else if (lastOp instanceof BetweenCondition) {
+        range = true;
       }
     } else if (condition instanceof InCondition) {
       size = 1;
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/IndexSearchDescriptor.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/IndexSearchDescriptor.java
@@ -21,6 +21,7 @@ package com.arcadedb.query.sql.executor;
 import com.arcadedb.index.Index;
 import com.arcadedb.index.RangeIndex;
 import com.arcadedb.query.sql.parser.AndBlock;
+import com.arcadedb.query.sql.parser.BetweenCondition;
 import com.arcadedb.query.sql.parser.BinaryCompareOperator;
 import com.arcadedb.query.sql.parser.BinaryCondition;
 import com.arcadedb.query.sql.parser.BooleanExpression;
@@ -66,7 +67,7 @@ public class IndexSearchDescriptor {
    */
   public boolean requiresMultipleIndexLookups() {
     for (BooleanExpression oBooleanExpression : getSubBlocks()) {
-      if (!(oBooleanExpression instanceof BinaryCondition))
+      if (!(oBooleanExpression instanceof BinaryCondition) && !(oBooleanExpression instanceof BetweenCondition))
         return true;
     }
     return false;
@@ -82,6 +83,8 @@ public class IndexSearchDescriptor {
     if (lastOp instanceof BinaryCondition condition) {
       final BinaryCompareOperator op = condition.getOperator();
       range = op.isRangeOperator();
+    } else if (lastOp instanceof BetweenCondition) {
+      range = true;
     }
 
     final long val = stats.getIndexStats(indexName, size, range, additionalRangeCondition != null);
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -3849,7 +3849,11 @@ public class SelectExecutionPlanner {
           indexFieldFound = true;
           indexKeyValue.getSubBlocks().add(singleExp.copy());
           blockIterator.remove();
-          if (singleExp instanceof BinaryCondition condition && condition.getOperator().isRangeOperator()) {
+          if (singleExp instanceof BetweenCondition
+              || (singleExp instanceof BinaryCondition condition && condition.getOperator().isRangeOperator())) {
+            // a range-shaped condition (BETWEEN, or a single-sided comparison like >/</>=/<=) is terminal for
+            // composite-index key building: no further field can be appended as if it were still an ordered
+            // equality prefix once the key contains a range component
             rangeOp = true;
           }
           if (rangeOp && info.allowsRange()) {
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -3458,7 +3458,7 @@ public class SelectExecutionPlanner {
             .mapToInt(i -> i).boxed().toList();
 
       subPlan.chain(new GetValueFromIndexEntryStep(context, filterClusterIds));
-      if (requiresMultipleIndexLookups((AndBlock) desc.keyCondition)) {
+      if (desc.requiresMultipleIndexLookups()) {
         subPlan.chain(new DistinctExecutionStep(context));
       }
       if (desc.remainingCondition != null && !desc.remainingCondition.isEmpty()) {
@@ -3467,22 +3467,6 @@ public class SelectExecutionPlanner {
       subPlans.add(subPlan);
     }
     return new ParallelExecStep(subPlans, context);
-  }
-
-  /**
-   * checks whether the condition has CONTAINSANY or similar expressions, that require multiple index evaluations
-   *
-   * @param keyCondition
-   *
-   * @return
-   */
-  private boolean requiresMultipleIndexLookups(final AndBlock keyCondition) {
-    for (final BooleanExpression oBooleanExpression : keyCondition.getSubBlocks()) {
-      if (!(oBooleanExpression instanceof BinaryCondition)) {
-        return true;
-      }
-    }
-    return false;
   }
 
   private WhereClause createWhereFrom(final BooleanExpression remainingCondition) {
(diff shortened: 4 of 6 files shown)


lvca:
Addressed the test-coverage suggestions from the review in 356279a82:
- Added `notBetweenOnIndexedColumnStillReturnsCorrectResults` - confirms `NOT BETWEEN` still returns correct results after this change.
- Added `betweenOnCaseInsensitiveIndexedColumnUsesIndexAndFolds` - `BETWEEN` on a `COLLATE CI` indexed STRING column uses the index and folds case correctly.
- Added `betweenAsFirstFieldOfCompositeUniqueIndex` - `BETWEEN` as the first field of a composite `UNIQUE` index, which also exercises `LSMTreeIndex.convertKeys()` on the insert side (not just the query side the other tests cover).

On the `NOT BETWEEN` question: confirmed intentional - `NotBlock` correctly has no `isIndexAware()` override, so it keeps falling back to the pre-existing bucket-scan/row-filter path. This PR only adds index-range support for a bare `BETWEEN`; pushing `NOT BETWEEN` into an index range is a separate, unrelated enhancement (would need a "everything outside this range" index scan shape, not a simple range lookup) and out of scope for #5966.

All 6 tests in `Issue5966BetweenUsesIndexTest` pass, plus the composite-index/case-insensitive-index/max-min-from-index suites - no regressions.

lvca:
Addressed the code-review finding in 8f7c46510:

**Third duplicate copy of "only BinaryCondition is understood" - fixed by removing the duplication, not patching it.** `SelectExecutionPlanner` had its own private `requiresMultipleIndexLookups(AndBlock)` (used only by `createParallelIndexFetch()`), separate from the one on `IndexSearchDescriptor` that the first commit already fixed. Rather than add the same `BetweenCondition` check a third time, removed `SelectExecutionPlanner`'s copy entirely and pointed the call site at `desc.requiresMultipleIndexLookups()` - one implementation instead of two, consistent with the PR's own stated goal for the key-conversion half of this fix.

Also added `orAcrossTwoIndexesWithBetweenBranchMergesWithoutRedundantDistinct`, covering the OR-across-two-indexes-with-a-BETWEEN-branch scenario the review pointed at: asserts both indexes are used, exactly one top-level `DISTINCT` merges the branches (not a redundant one per branch), and the returned row set is correct.

**On the reviewer's "minor/non-blocking" note about `createParallelIndexFetch` itself:** while tracing this I confirmed that method is reached only through the polymorphic-subtype recursion (`handleTypeAsTargetWithIndexRecursive`, used when a base type has zero records of its own and dispatches per-subtype) - a plain OR query on a single concrete type is intercepted earlier by `handleTypeAsTargetWithIndexedFunction`, which already called the correct `bestIndex.requiresDistinctStep()` before this PR. I didn't add a test that reaches `createParallelIndexFetch` specifically (constructing a reliable, non-contrived repro requires a base type with zero own records plus per-subtype-resolved indexes, and the reviewer's own writeup already establishes this was never a correctness bug - just a redundant step). The fix itself removes the risk regardless of which path exercises it, since there's now only one implementation to be either right or wrong.

All 7 tests in `Issue5966BetweenUsesIndexTest` pass, plus the composite-index/case-insensitive-index/max-min-from-index suites and the full `com.arcadedb.query.sql.executor.**` package - no regressions.

lvca:
Latest review pass found no correctness issues and confirmed the two fixes from the prior rounds. On the one minor, explicitly-optional observation - `WHERE LOWER(name) BETWEEN 'a' AND 'c'` on a CI index not using the `isFieldWithLowerCaseMethod` optimization branch `BinaryCondition.isIndexAware()` has: agreed this is out of scope for #5966 (whose repro is a plain indexed column, not the `LOWER()`-wrapped CI-index optimization), and the reviewer's own text already reaches that conclusion ("a reasonable place to stop... might be worth a follow-up issue if that pattern shows up in practice"). Not implementing it here; happy to file a follow-up issue if it comes up in practice.

Waiting on `build-and-package` (runs the full `mvn verify`) and the last `Analyze` job to finish before this is mergeable.

lvca:
\`ha-integration-tests\` failed on this run with two test failures, both in the \`ha-raft\` module:
- \`Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied\` - \`AssertionFailedError: [Node 0 must hold both vertices] expected: 2L but was: 1L\` (replication-convergence timing)
- \`Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact\` - \`ExecutionException: AssertionError: Update returned null: record=3 seq=8\` (slot-merge delete replication timing)

Both are Raft cluster convergence/timing assertions in \`ha-raft\`, a module this PR's diff never touches - the change here is scoped to \`engine\`'s SQL planner (\`SelectExecutionPlanner\`, \`FetchFromIndexStep\`, \`IndexSearchDescriptor\`, \`BetweenCondition\`) and \`LSMTreeIndex\`'s key-conversion dedupe, none of which is on the Raft commit/replication path these tests exercise. This matches the profile of this lane's known chronic flakiness under CI resource contention, unrelated to any specific diff.

Triggered a rerun of the failed jobs (\`gh run rerun --failed\`) to confirm non-determinism rather than asserting it without evidence.

lvca:
Rerun of \`ha-integration-tests\` failed again, but with a **different** specific failure profile than the first run:
- \`Issue5410AbandonedTicketReleaseIT\` (again - same class as run 1)
- \`RemoteStickyConnectionStrategyIT\` (new this run, run 1 instead had \`Issue5569SlotMergeDeleteRaftIT\` fail)

Both new failures are, again, Raft cluster timing/connection-convergence assertions in \`ha-raft\` (gRPC connection errors, \`APPEND_ENTRIES\` cancellations, ticket-release timing) - unrelated to this PR's diff. Different specific tests failing across independent runs is the signature of environment/timing non-determinism rather than a reproducible regression tied to a code change; a real regression from this PR's diff (SQL planner + LSM key-conversion, nothing in \`ha-raft\`) would fail the *same* assertion deterministically every run.

Triggering another rerun.

lvca:
Third \`ha-integration-tests\` rerun failed again, this time with a single failure: \`Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied\` (also seen in run 1; run 2 had it plus a different test).

Checked whether this is specific to this branch by looking at recent CI runs on \`main\` itself (\`gh run list --branch main --workflow=mvn-test.yml\`): the last 4 completed runs on \`main\` **all show \`conclusion: failure\`**, and the most recent one (commit 8ea92f03, unrelated to #5966) failed both \`ha-integration-tests\` **and** \`unit-tests\` - a lane that passed cleanly on every attempt of this PR's run. This confirms the CI environment itself is currently red independent of any branch's diff; this PR isn't introducing or uniquely triggering the flake.

Triggering one more rerun. Given main itself is failing the same lane right now, this may need to resolve on the infrastructure side rather than something fixable in this PR - will keep documenting outcomes either way.

lvca:
Fourth \`ha-integration-tests\` rerun failed with yet another different test: \`RaftPriorityRejoinIT\` (an \`Error\`, not an assertion failure this time).

Summary across all 4 attempts on this PR:
| Run | Failure(s) |
|---|---|
| 1 | \`Issue5410AbandonedTicketReleaseIT\` + \`Issue5569SlotMergeDeleteRaftIT\` |
| 2 | \`Issue5410AbandonedTicketReleaseIT\` + \`RemoteStickyConnectionStrategyIT\` |
| 3 | \`Issue5410AbandonedTicketReleaseIT\` (alone) |
| 4 | \`RaftPriorityRejoinIT\` (alone) |

Four different failure combinations across four Raft cluster integration test classes, none overlapping with this PR's diff (SQL planner + LSM index key-conversion in `engine`, nothing in `ha-raft`). Combined with the earlier finding that `main` itself failed its last 4 CI runs (one of which also failed the unrelated `unit-tests` lane), this is conclusive: the CI environment is currently exhibiting broad timing-sensitive flakiness in the Raft integration suite, independent of any specific branch's changes.

Triggering a 5th rerun. I'll keep documenting outcomes, but at this point the evidence strongly supports that this PR is correctness-complete and blocked only by pre-existing CI infrastructure flakiness that also affects `main` - worth flagging to whoever owns CI/infra if it keeps happening across other PRs too.

lvca:
Fifth \`ha-integration-tests\` rerun: failed again, same test as runs 1-3: \`Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied\` (4 of 5 runs now). This test was added in PR #5472 (fixing #5410) - it's an established test, not something new or touched by this PR.

**Status summary:** the code review (\`claude-review\`) passed cleanly across two rounds with all valid points addressed (BETWEEN index-awareness, the composite-key rangeOp fix, the third duplicate-check removal, and the follow-up test coverage it asked for). Every other CI lane passes consistently on every attempt - \`unit-tests\`, \`slow-unit-tests\`, \`integration-tests\`, \`vector-unit-tests\`, \`build-and-package\`, all e2e suites, \`opencypher-tck-tests\`, static analysis. Only \`ha-integration-tests\` is unstable, and `main` itself is independently red on this same lane right now (confirmed via its last 4 CI runs). This PR's diff never touches \`ha-raft\`.

Continuing to rerun per the babysitting loop, but at this point further reruns are unlikely to add new information - the diagnosis is solid. This may need attention from whoever owns CI reliability for the \`ha-raft\` suite, separate from this PR.

lvca:
Sixth rerun: failed with two more different tests - \`RaftReplicationMaterializedViewIT\` and \`Raft3PhaseCommitIT.basicInsertReplicates\` - neither seen in any of the previous 5 runs.

Running tally across 6 attempts, 8 distinct failures, 7 different test classes: \`Issue5410AbandonedTicketReleaseIT\` (×4), \`Issue5569SlotMergeDeleteRaftIT\`, \`RemoteStickyConnectionStrategyIT\`, \`RaftPriorityRejoinIT\`, \`RaftReplicationMaterializedViewIT\`, \`Raft3PhaseCommitIT\`. This is no longer "one flaky test" - it's broad instability across the whole \`ha-raft\` integration suite, consistent with the CI runner being under resource pressure during these 3-node Raft cluster tests (which are inherently timing-sensitive). Combined with \`main\` independently failing its last several runs on the same lane, I'm confident this is a CI infrastructure issue, not a defect in this PR.

Triggering rerun #7.

lvca:
Seventh rerun: failed again, and worse than before - **three** distinct failures this time (\`RemoteStickyConnectionStrategyIT\`, \`RaftPriorityRejoinIT\`, \`RaftRemoteReadYourWritesIT\`) and the run took 1h6m instead of the usual ~56m.

Final tally across 7 attempts: 9 distinct \`ha-raft\` test classes have failed at least once (\`Issue5410AbandonedTicketReleaseIT\` ×4, \`RemoteStickyConnectionStrategyIT\` ×2, \`RaftPriorityRejoinIT\` ×2, \`Issue5569SlotMergeDeleteRaftIT\`, \`RaftReplicationMaterializedViewIT\`, \`Raft3PhaseCommitIT\`, \`RaftRemoteReadYourWritesIT\`). The trend is getting worse, not better, and \`main\` itself was independently failing this same lane before this PR ever existed. This is CI infrastructure/resource contention affecting the whole \`ha-raft\` suite, not a defect this PR introduces or can fix - the diff here never touches that module.

**This PR is code-complete:** the code review passed cleanly across two rounds with every valid point addressed, and every other CI lane (\`unit-tests\`, \`slow-unit-tests\`, \`integration-tests\`, \`vector-unit-tests\`, \`build-and-package\`, all e2e suites, static analysis) passes consistently. I'll keep retrying \`ha-integration-tests\` per the babysitting loop, but further identical reruns are unlikely to add new information at this point - this may need either a CI infra fix, an admin merge override, or the branch protection rule adjusted to not require this specific chronically-unstable lane.

Issue #5977, opened 2026-08-09
HA integration tests repeatedly trip a DatabaseIsClosedException race right after a snapshot-reinstall resync


lvca:
## Summary

Three independent `ha-raft` integration tests have now failed with the identical shape - a `DatabaseIsClosedException` thrown from a query/command issued on a database handle obtained just before a snapshot-reinstall resync completes:

1. `Issue5492TruncateBatchNotReplicatedIT` - already documented as a known trap in `ha-raft/CLAUDE.md`:
   > "the resync reinstalls the follower's database and a stale handle then throws `DatabaseIsClosed`, which reads like infrastructure noise rather than divergence."
2. `RaftReadConsistencyBookmarkIT.readConsistencyContextAppliesOnFollowerQueries` - observed on PR #5975's CI (2026-08-08, run 31281124979, job 93163224266):
   ```
   com.arcadedb.exception.DatabaseIsClosedException: graph
   	at com.arcadedb.database.LocalDatabase.checkDatabaseIsOpen(LocalDatabase.java:2637)
   	at com.arcadedb.database.LocalDatabase.query(LocalDatabase.java:1813)
   	at com.arcadedb.server.ha.raft.RaftReplicatedDatabase.query(RaftReplicatedDatabase.java:1287)
   	at com.arcadedb.server.ServerDatabase.query(ServerDatabase.java:567)
   	at com.arcadedb.server.ha.raft.RaftReadConsistencyBookmarkIT.readConsistencyContextAppliesOnFollowerQueries(RaftReadConsistencyBookmarkIT.java:86)
   ```
   Preceded by: `Database 'graph' bootstrap mismatch (...); reinstalling from leader-shipped full snapshot` / `Database 'graph' reinstalled after bootstrap mismatch`.
3. `Issue5492SchemaWalNotShippedIT.walBufferedInsideRecordFileChangesReachesFollowers` - observed on the same PR's CI, a later rerun (job 93163224266's rerun, commit `1e6b667cd`):
   ```
   com.arcadedb.exception.DatabaseIsClosedException: graph
   	at com.arcadedb.database.LocalDatabase.checkDatabaseIsOpen(LocalDatabase.java:2637)
   	at com.arcadedb.database.LocalDatabase.command(LocalDatabase.java:1764)
   	at com.arcadedb.server.ha.raft.RaftReplicatedDatabase.command(RaftReplicatedDatabase.java:779)
[...]

Issue #5971, opened 2026-08-08
Issue4960LSMCloseDuringCompactionTest flaky: post-test integrity check finds torn index page after close-during-compaction


lvca:
## Summary

`com.arcadedb.index.lsm.Issue4960LSMCloseDuringCompactionTest.closeCancelsScheduledCompaction` failed on CI's `unit-tests` job. Not caused by the PR it ran on (#5969, SQL function/method argument validation - completely unrelated code paths); flagging as a standalone flake since it wasn't previously tracked.

## What happened

The test's own assertions passed, but `TestHelper.afterTest`'s post-test `checkDatabaseIntegrity` check found the index left in a torn state:

```
Issue4960LSMCloseDuringCompactionTest>TestHelper.afterTest:175->TestHelper.checkDatabaseIntegrity:224
Warnings[
  index 'CloseIdx_0_1909563823331': the physical key order does not match the current comparator, the index must be rebuilt,
  index 'CloseIdx_0_1909563823331': page 0: keys cannot be read (Invalid position 67109887 (size=262136))
]
```

Preceding log line:
```
SEVER [TransactionManager] Preserving the WAL files of database 'Issue4960LSMCloseDuringCompactionTest': not all pages reached
the next open will recover them
SEVER [LocalDatabase] Database 'Issue4960LSMCloseDuringCompactionTest' closed with unflushed pages: the lock file and the WAL
were preserved so the next open will recover them
```

## Where

`engine/src/test/java/com/arcadedb/index/lsm/Issue4960LSMCloseDuringCompactionTest.java`, method `closeCancelsScheduledCompaction`.

## Why it looks like a real race, not test flakiness

The test name states the intent: closing the database should cancel a scheduled compaction cleanly. The failure surfaces as a corrupted index page (bad key order + an unreadable page position) picked up by the *post*-test integrity checker rather than by the test's own assertions - which is the shape of a compaction write landing partially on a page while close is tearing down the database concurrently, rather than a simple test-logic bug.

## Reproduction

Not yet reproduced locally - only observed once on CI (`unit-tests` job, run https://github.com/ArcadeData/arcadedb/actions/runs/31264257741/job/93127417276, commit aac25a4f8). Given the WAL "unflushed pages" log line, this is likely timing-dependent and may need looping the test or running under load to reproduce reliably.

## Suggested next step

[...]

Issue #5418, opened 2026-07-25
Background threads vs database lifecycle: non-daemon AsyncFlush keeps embedder JVM alive after a leaked Database; vector inactivity timer fires on closed/deleted databases


tae898:
Two related lifecycle behaviors, both observed while chasing stochastic stalls in the Python-bindings CI. Reproduced on 26.8.1 dev13 through current main; neither is a recent regression, they just get exposed by test timing.

**1. A Database that is never closed keeps the JVM alive forever**

Minimal repro, no vector index involved:

```python
db = create_database(path)     # embedded, in-process JVM
db.command("sql", "CREATE DOCUMENT TYPE Doc")
# exit WITHOUT db.close()
```

The process never exits: `DestroyJavaVM` waits for non-daemon threads, and the engine's "ArcadeDB AsyncFlush" thread (plus friends) is non-daemon and runs until `Database.close()`. Thread dump from a hung process shows the main thread inside JVM shutdown and AsyncFlush parked in TIMED_WAITING.

In an embedded host this turns any leaked handle into an interpreter that cannot exit. In CI it is a stochastic full-run stall: a test failure path that skips a close (for example, a crashed graph rebuild leaving a test bailing early) wedges the entire pytest session at teardown. On Windows the same end state sometimes surfaces as an access violation inside `shutdownJVM` instead of a hang.

We have shipped a bindings-side mitigation (an atexit hook that walks `DatabaseFactory.getActiveDatabaseInstances()` and closes whatever is still open before JVM shutdown), so this is not blocking us. Filing because every embedder faces it: engine-side options would be daemonizing the background threads and/or a JVM shutdown hook that closes active instances.

**2. The vector inactivity-rebuild timer fires on closed and even deleted databases**

From a green CI run, seconds after the suite finished and its temp dirs were removed:

```
WARNI [TransactionManager] Could not write last-tx-id.bin; ... Cause: /tmp/arcadedb_test_db_9kwgi1yx/test_db/last-tx-id.bin.tmp (No such file or directory)
INFO  [LSMVectorIndex$2] Inactivity timeout expired (15000 ms), triggering graph rebuild for 3 pending mutations (index: Doc_0_...)
WARNI [LSMVectorIndexPageParser] Error parsing page 0 in file 4: test_lsm_vector_db
SEVER [LSMVectorIndex] Error building graph from scratch
com.arcadedb.exception.DatabaseIsClosedException: test_vector_sql_db
[...]

PR #6019, opened 2026-08-09
fix: reject NaN narrowing (#5970), fix LSM close-during-compaction test race (#5971), dedupe HA resync-retry helper (#5977)

### LINKED ISSUE #5970: Type.convert(): NaN silently becomes 0 when narrowed to BYTE/SHORT/INTEGER instead of being rejected
## Summary

Follow-up from code review on #5967 (fix for #5905/#5906: `Type.convert()` now range-checks before narrowing a `Number` to `BYTE`/`SHORT`/`INTEGER` instead of silently wrapping on overflow).

`narrowToIntegral()`'s primitive path falls through to `value.longValue()` for `Double`/`Float`. Per the JLS narrowing-conversion rules, `Double.NaN.longValue()`/`Float.NaN.longValue()` return `0`, which is in-range for every target type - so `Type.convert(db, Double.NaN, Integer.class)` silently returns `0` instead of raising an error. `Infinity` is already handled correctly (it saturates to `Long.MAX_VALUE`/`MIN_VALUE`, which is caught by the new range check), so this is specifically a `NaN` gap.

## Why this matters

This is the same class of bug #5905 fixed: a value is silently coerced into a plausible-looking but wrong number instead of the write being rejected. `NaN` reaching an `INTEGER`/`SHORT`/`BYTE` property (e.g. from a division-by-zero-as-double result stored directly, or a JSON `NaN` literal parsed upstream) becomes an unremarkable `0` with no error and no warning.

This is **not a regression** from #5967 - `.intValue()`/`.shortValue()` had the identical behavior before that fix - it just wasn't in scope there (the fix was about overflow rejection, not `NaN` rejection specifically). Flagged independently in two review rounds on #5967 as worth tracking separately.

## Reproducer

```java
Type.convert(database, Double.NaN, Integer.class);   // returns 0, no error
Type.convert(database, Float.NaN, Short.class);       // returns (short) 0, no error
```

## Suggested fix

In `Type.narrowToIntegral()` (`engine/src/main/java/com/arcadedb/schema/Type.java`), add a `Double.isNaN`/`Float.isNaN` guard alongside the existing range check and reject with the same `IllegalArgumentException` shape used for out-of-range values.

cc @lvca

lvca:
## Summary

Three independent issue triages bundled into one PR since each is small and none touch overlapping code.

### #5970: `Type.convert()` NaN silently becomes 0

`Type.narrowToIntegral()` range-checks overflow before narrowing to `BYTE`/`SHORT`/`INTEGER`, but `Double.NaN.longValue()`/`Float.NaN.longValue()` return `0` per the JLS narrowing-conversion rules - in-range for every target type, so a NaN slipped through the range check as a silent `0` instead of being rejected. `Infinity` was already caught (it saturates to a `Long` boundary, which is out of range).

**Fix:** explicit `isNaN()` guard in `narrowToIntegral()`, same `IllegalArgumentException` shape as the existing range-check rejection.

### #5971: `Issue4960LSMCloseDuringCompactionTest` flaky - torn index page after close-during-compaction

Reported as a single CI observation, not yet reproduced locally. Root-caused by inspection + a reproduction:

The test calls `LSMTreeIndex.close()` directly on a live, open database, immediately after an insert transaction. That transaction's pages are still queued for **asynchronous** flush (`TransactionContext.asyncFlush` defaults to `true`, `TX_WAL_FLUSH` defaults to no-flush-at-commit) - closing the index's files early races `PageManagerFlushThread`, which later finds the file already closed (`PageManager.flushPage()` throws `DatabaseMetadataException`), leaves the page unflushed, and turns the WAL-preserving recovery that follows into a torn index page instead of a clean replay.

This is not a production bug: `IndexInternal.close()` has **zero** production callers today (grepped the whole repo) - `LocalDatabase`'s own close path deliberately avoids calling it directly (see the `#5418` comment in `closeDurableParts()`), closing index files only via `FileManager.close()` once `PageManager.waitAllPagesOfDatabaseAreFlushed()` has run. The regression test violated that undocumented precondition.

**Reproduction:** amplified the test's insert to 200k records (with auto-compaction disabled via config) so the flush thread could not possibly keep up with an immediate `close()` - reliably reproduced the exact CI log lines (`"Preserving the WAL files... not all pages reached the disk"` / `"closed with unflushed pages"`) with zero code changes, confirming the race.

**Fix:** the test now calls `PageManager.INSTANCE.waitAllPagesOfDatabaseAreFlushed(database)` before `close()`, matching the real close path's guarantee. Also documented the precondition directly on `IndexInternal.close()`'s Javadoc (previously only implied via `releaseBackgroundResources()`'s doc) so the next direct caller doesn't rediscover this by triggering corruption.

Verified: 5 consecutive runs clean before the fix reliably showed the WAL-preservation warning at amplified scale; 5 consecutive runs clean after, and the actual regression test at its original (small) scale runs clean 5/5 with no warnings.

### #5977: HA integration tests repeatedly trip `DatabaseIsClosedException` across a resync boundary

Three HA ITs hit `DatabaseIsClosedException` from querying/commanding a database handle obtained just before a snapshot-reinstall resync reinstalls the target database out from under it. Two ITs (`Issue5492TruncateBatchNotReplicatedIT`, `Issue5655CypherCommitsOnInnerDatabaseIT`) had already independently reinvented a resolve-fresh-and-retry fix for this; a third (`Issue5492SchemaWalNotShippedIT`) and `RaftReadConsistencyBookmarkIT` had no defense at all and are the ones the issue's CI runs actually caught failing.

The issue suggested a "wait for any in-flight resync to finish, then re-resolve" helper. That shape is still racy - a resync can start in the gap between the wait check and the use that follows it. The proven, already-working pattern in this codebase (independently arrived at twice) is **resolve fresh, retry on the actual failure** (`DatabaseIsClosedException`), which is what's implemented here.

**Fix:** moved the pattern into `BaseRaftHATest` as `withResyncRetry()` (general one-shot retry, for `RaftReadConsistencyBookmarkIT`'s single query), `countOn()`/`awaitValue()`/`awaitCountOn()` (the polling variant, for the count-convergence tests). All four HA IT classes now use the shared version instead of three divergent ad hoc copies (one of which was simply absent).

## Test plan

- [x] `TypeTest` (full class, including new `convertNaNToIntegralRejected` regression test) - pass
- [x] `Issue4960LSMCloseDuringCompactionTest` - pass, 5 consecutive runs
- [x] Reproduction test (amplified insert, no fix) - reliably reproduced the exact CI log lines; removed after confirming root cause and fix
- [x] `RaftReadConsistencyBookmarkIT` (both methods, including the previously-failing one) - pass
- [x] `Issue5492SchemaWalNotShippedIT` - pass
- [x] `Issue5492TruncateBatchNotReplicatedIT` (dedup regression check) - pass
- [x] `Issue5655CypherCommitsOnInnerDatabaseIT` (dedup regression check) - pass
- [x] Full `mvn compile` + `test-compile` across `engine`, `server`, `ha-raft` - clean
- [x] Full `com.arcadedb.schema.**` + `com.arcadedb.index.**` suite in `engine` (254 test classes) - 0 failures

Fixes #5970, #5971, #5977

Claude-Session: https://claude.ai/code/session_01YEu5VG1ATFVFtxDWsRhdpj

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/IndexInternal.java
+++ b/engine/src/main/java/com/arcadedb/index/IndexInternal.java
@@ -88,6 +88,15 @@ public interface IndexInternal extends Index {
   default void releaseBackgroundResources() {
   }
 
+  /**
+   * Closes the index files. The caller must guarantee that every page committed against this index has already
+   * reached disk (e.g. {@code PageManager.INSTANCE.waitAllPagesOfDatabaseAreFlushed(database)}) before calling
+   * this: a page still queued for asynchronous flush (the default, see {@code TransactionContext.asyncFlush})
+   * finds its file closed underneath it, is left unflushed, and turns the WAL-preserving recovery that follows
+   * into a torn page instead of a clean replay (issue #5971). {@code LocalDatabase}'s own close path satisfies
+   * this by construction - it never calls this method directly, closing index files only via
+   * {@code FileManager.close()} once the database-wide flush has completed.
+   */
   void close();
 
   void drop();
--- a/engine/src/main/java/com/arcadedb/schema/Type.java
+++ b/engine/src/main/java/com/arcadedb/schema/Type.java
@@ -541,6 +541,12 @@ public enum Type {
           return string.isEmpty() ? 0L : Long.parseLong(string);
         else if (DateUtils.isDate(value))
           return DateUtils.dateTimeToTimestamp(value, ChronoUnit.MILLIS);
+        else if (isNaN((Number) value))
+          // LONG never goes through narrowToIntegral() - there is no narrower range to check, it IS the widest
+          // integral type - so it needs its own NaN guard (issue #5970).
+          throw new IllegalArgumentException(
+              "Value '" + value + "' is NaN and cannot be converted to type LONG" //
+                  + (property != null ? " for property '" + property.getName() + "'" : ""));
         else
           return ((Number) value).longValue();
 
@@ -785,6 +791,16 @@ public enum Type {
   }
 
   /**
+   * True when {@code value} is a {@link Double} or {@link Float} holding {@code NaN}. Per the JLS narrowing-
+   * conversion rules {@code Double.NaN.longValue()}/{@code Float.NaN.longValue()} return {@code 0}, which is
+   * in-range for every integral target type, so every narrowing path (not just {@link #narrowToIntegral}) must
+   * check this explicitly instead of trusting the range check to catch it (issue #5970).
+   */
+  private static boolean isNaN(final Number value) {
+    return (value instanceof Double doubleValue && doubleValue.isNaN()) || (value instanceof Float floatValue && floatValue.isNaN());
+  }
+
+  /**
    * Range-checks {@code value} before narrowing it to a smaller integral type ({@code BYTE}, {@code SHORT} or
    * {@code INTEGER}). Narrowing with a plain {@code .intValue()}/{@code .shortValue()}/{@code .byteValue()} wraps
    * two's-complement on overflow instead of rejecting - {@code 3000000000L} silently became {@code -1294967296} -
@@ -801,7 +817,7 @@ public enum Type {
    */
   private static Number narrowToIntegral(final Number value, final long min, final long max, final String targetType,
       final Property property) {
-    if (value instanceof Double doubleValue && doubleValue.isNaN() || value instanceof Float floatValue && floatValue.isNaN())
+    if (isNaN(value))
       throw new IllegalArgumentException(
           "Value '" + value + "' is NaN and cannot be converted to type " + targetType //
               + (property != null ? " for property '" + property.getName() + "'" : ""));
--- a/engine/src/main/java/com/arcadedb/schema/Type.java
+++ b/engine/src/main/java/com/arcadedb/schema/Type.java
@@ -794,9 +794,18 @@ public enum Type {
    * {@code Infinity}) is caught via the saturating conversion {@link Number#longValue()} already performs. A
    * {@link BigDecimal}/{@link BigInteger} can exceed even the long range, where {@code longValue()} truncates bits
    * instead of saturating, so those two types are range-checked directly against {@code long} bounds first.
+   * <p>
+   * {@code NaN} needs its own guard: per the JLS narrowing-conversion rules {@code Double.NaN.longValue()}/
+   * {@code Float.NaN.longValue()} return {@code 0}, which is in-range for every target type and would otherwise
+   * slip through the range check as a silent {@code 0} (issue #5970).
    */
   private static Number narrowToIntegral(final Number value, final long min, final long max, final String targetType,
       final Property property) {
+    if (value instanceof Double doubleValue && doubleValue.isNaN() || value instanceof Float floatValue && floatValue.isNaN())
+      throw new IllegalArgumentException(
+          "Value '" + value + "' is NaN and cannot be converted to type " + targetType //
+              + (property != null ? " for property '" + property.getName() + "'" : ""));
+
     final long longValue;
     if (value instanceof BigDecimal bigDecimal)
       longValue = bigDecimal.compareTo(BigDecimal.valueOf(Long.MAX_VALUE)) > 0 ? Long.MAX_VALUE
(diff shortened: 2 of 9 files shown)


lvca:
Thanks for the review. Addressed:

- **LONG NaN gap (confirmed valid, fixed)**: extracted the NaN check into a shared `isNaN()` helper and applied it to the `LONG` branch too, which had the identical `longValue()`-returns-0 defect. Added `Long.class` cases to `convertNaNToIntegralRejected`. This also resolves the operator-precedence nit, since `isNaN()` now has its own explicit-parens implementation instead of the inline condition.

- **`awaitValue()`'s broad `RuntimeException` catch vs `withResyncRetry`'s narrow `DatabaseIsClosedException`**: leaving this as-is. It's pre-existing behavior carried over unchanged from both prior ad hoc implementations (`Issue5492TruncateBatchNotReplicatedIT`'s old `awaitCountOn`, `Issue5655CypherCommitsOnInnerDatabaseIT`'s old `await`), not something introduced by this PR. This test suite is a documented chronic-flake area, and the broad catch was a deliberate choice to tolerate whatever transient shape a mid-resync failure takes, not just `DatabaseIsClosedException` specifically (e.g. a wrapped Raft/gRPC error during the same window). Narrowing it risks trading "the polling helper is slow to fail" for "an unrelated transient error now hard-fails the test instead of being retried," which would work against #5977's actual goal. `withResyncRetry()` (the new one-shot primitive, used where there's no numeric convergence to poll toward) stays narrow deliberately, since there the single-attempt semantics make a broad catch swallow real bugs.

Pushed as b119b29a2.

lvca:
Second review round addressed:

- **NaN blast radius via typed getters (checked, no change needed)**: scanned the vector/embedding read paths the review flagged. Embeddings are stored as array-typed properties (`ARRAY_OF_FLOATS`/`ARRAY_OF_DOUBLES`/vector types), never as scalar `BYTE`/`SHORT`/`INTEGER`/`LONG`, so they never route through the narrowing branches this fix touches - confirmed by `SQLFunctionVectorHasNaN`/`IsNaNFunction` existing specifically to let vector code *check for* NaN rather than anything relying on it silently becoming 0 through a scalar narrowing conversion. A property that legitimately needs NaN-tolerant narrowing to an integral type isn't a pattern that exists in this codebase, and if one ever does, throwing loudly is exactly the outcome #5970 wants.
- **Stale Javadoc cross-reference**: fixed - `{@link #countOn}` now points at `{@link BaseRaftHATest#countOn}` with a bit more context, since `countOn` moved out of the local class.
- **`RESYNC_RETRY_TIMEOUT_MS` placement**: moved next to `BASE_RAFT_PORT` at the top of the class.

Pushed as 2fb8b30cf.

lvca:
Third review round: no blocking findings, confirmed all three fixes correct. It raised one new point:

- **Array-narrowing branches (`double[]`/`float[]` -> `int[]`/`long[]`/`short[]`) have the same NaN/overflow gap** - confirmed valid, but explicitly out of scope for this PR (different code path, different call sites, deserves its own regression test rather than expanding this bundle). Filed as #6020 for tracking.

This PR is scoped to #5970 (rejected), #5971 (rejected), #5977 (rejected) plus the LONG-branch gap the first review round caught within #5970's own fix. Everything else across three review rounds has come back clean.

lvca:
## Full CI matrix result

All pre-checks, `build-and-package`, and 10 of 12 test-matrix lanes are green, including `unit-tests`, `slow-unit-tests`, `integration-tests`, and every e2e lane.

Two lanes show red, both confirmed unrelated to this PR:

- **`vector-unit-tests`**: `LSMVectorIndexRebuildTest`/`LSMVectorIndexRecoveryTest` timed out after 5 minutes waiting on an async rebuild counter (`ConditionTimeoutException`). This is the documented pre-existing `REBUILD_SEMAPHORE` chronic flake on `main`, independent of any diff - this PR touches zero vector-index code.
- **`ha-integration-tests`**: 208/210 tests passed, including all four HA IT files this PR actually touches (`Issue5492SchemaWalNotShippedIT`, `Issue5492TruncateBatchNotReplicatedIT`, `Issue5655CypherCommitsOnInnerDatabaseIT`, `RaftReadConsistencyBookmarkIT` - the ones #5977 fixed/deduped - all green). The 2 failures are in `Issue5410AbandonedTicketReleaseIT` (a page-version divergence assertion in Raft ticket-release logic) and `Issue5569SlotMergeDeleteRaftIT` (a `TX_PAGE_SLOT_MERGE` delete-replication edge case) - neither file is touched by this PR, and neither failure involves `Type.convert()`, `IndexInternal.close()`, or the new `BaseRaftHATest` helpers. This lane is a documented chronic-flake area independent of this diff.

Summary across three review rounds: one real gap found and fixed (LONG branch NaN guard), one confirmed-valid follow-up filed separately as it's out of scope (#6020, array-narrowing branches), and everything else explained or already correct. Ready to merge from my side.

Issue #5677, opened 2026-08-01
HASH index cannot index @out/@in (LINK) keys, and fails at insert time claiming page corruption


lvca:
## What happens

A `UNIQUE` HASH index on an edge type's endpoints is accepted at creation and then fails on the first
insert:

```java
database.getSchema().createVertexType("Account");
database.getSchema().createEdgeType("INITIATED");
database.getSchema().buildTypeIndex("INITIATED", new String[] { "@out", "@in" })
    .withType(Schema.INDEX_TYPE.HASH).withUnique(true).create();   // succeeds

hub.newEdge("INITIATED", leaf).save();                              // throws on commit
```

```
com.arcadedb.index.IndexException: Unsupported key type for hash index
  'INITIATED_0_1380184008737625' (fileId=5): 14 (0x0E) while parsing entry at content offset 11.
  Loaded key types=[14(0x0E)=INVALID, 14(0x0E)=INVALID]. This means the index metadata or a bucket
  page is corrupted; rebuild the index (DROP and recreate it).
    at com.arcadedb.index.hash.HashIndexBucket.unsupportedKeyType(HashIndexBucket.java:914)
    at com.arcadedb.index.hash.HashIndexBucket.getSerializedValueSize(HashIndexBucket.java:1737)
    at com.arcadedb.index.hash.HashIndexBucket.computeKeyLengthFromPage(HashIndexBucket.java:1681)
    at com.arcadedb.index.hash.HashIndexBucket.compareKeyBytes(HashIndexBucket.java:1780)
    at com.arcadedb.index.hash.HashIndexBucket.findInsertionPoint(HashIndexBucket.java:1579)
```

## Cause

`TypeIndexBuilder:257` maps `@out`/`@in` on a `LocalEdgeType` to `Type.LINK`. A LINK whose document has
an identity serializes as `BinaryTypes.TYPE_RID` (14). `HashIndexBucket` handles
`TYPE_COMPRESSED_RID` (13) in its size, compare and serialize paths but has no case for `TYPE_RID`
(14), so every key-length computation on such an entry falls through to `unsupportedKeyType`.

The same index built as `LSM_TREE` works, so this is specific to the hash implementation.

## Two separate defects

1. **The key type is unsupported.** Either `HashIndexBucket` should handle `TYPE_RID` alongside
   `TYPE_COMPRESSED_RID`, or the builder should refuse the combination.
[...]

Issue #150, opened 2021-10-19
Realtimey features, aka change streams


lvca:
### Discussed in https://github.com/ArcadeData/arcadedb/discussions/149

<div type='discussions-op-text'>

<sup>Originally posted by **tetious** October 18, 2021</sup>
I'm wondering if there are plans around or interest in implementing something like Mongo's [Change Streams](https://docs.mongodb.com/manual/changeStreams/)? 

It seems like it might not be too bad to add, given it is essentially a special-case of replication. This might be an oversimplification based on optimistic naivety, though. 😄 

Being able to subscribe to changes at the "Type" level, as well as database level, would be very interesting I think. It would also allow for fleshing out some of the more flashy bits of the Redis (pub/sub) and Mongo feature-sets.

Thoughts?</div>

PR #6021, opened 2026-08-09
fix(#5703): store an LSM_TREE LINK key as a compressed RID

### LINKED ISSUE #5703: LSM_TREE stores a LINK key as a fixed 12 bytes per column; evaluate the compressed RID used by HASH
## Background

#5677 made a `HASH` index able to key on a `LINK` (a `LINK` property, or an edge type's `@out`/`@in`). The fix stored
those keys as `TYPE_COMPRESSED_RID` - the varint form the hash bucket already uses for every entry *value* - rather
than the `TYPE_RID` the schema declares:

- `TYPE_RID` is `putInt(bucketId)` + `putLong(position)` = **12 bytes, fixed**
- `TYPE_COMPRESSED_RID` is `putNumber(bucketId)` + `putNumber(position)` = **~2-7 bytes** in practice

On the composite `(@out, @in)` key that motivates an endpoint-keyed unique index that is roughly half the key bytes,
and therefore roughly half the pages the index occupies.

`LSM_TREE` still uses the declared type. `LSMTreeIndexAbstract` (~line 138):

```java
this.binaryKeyTypes = new byte[keyTypes.length];
for (...) this.binaryKeyTypes[i] = keyTypes[i].getBinaryType();   // LINK -> TYPE_RID -> 12 bytes
```

So an LSM index on `(@out, @in)` spends 24 bytes per entry on the key where ~8-14 would do. Edge de-duplication
indexes are exactly the case that gets large, so this is worth measuring.

## Why this is not a straight copy of the HASH fix

A hash index only needs the encoding to be **deterministic and injective** - it compares serialized key bytes for
equality and the ordering within a page is an internal detail. An LSM index is ordered: range scans, `MIN`/`MAX` and
every ordered cursor depend on key order being meaningful.

The encouraging part is that LSM does **not** compare raw bytes. `LSMTreeIndexAbstract.compareKey` deserializes and
uses the typed comparator:

```java
final Object keyValue = serializer.deserializeValue(database, currentPageBuffer, binaryKeyTypes[keyIndex], ...);
result = comparator.compare(key, binaryKeyTypes[keyIndex], keyValue, binaryKeyTypes[keyIndex]);
```

so ordering comes from comparing RIDs semantically, not from the byte layout - which suggests a narrower encoding
would not by itself reorder anything. **That needs proving, not assuming**, before anyone touches it. In particular:

- `BinaryComparator` must order `TYPE_COMPRESSED_RID` identically to `TYPE_RID` (and handle a mixed pair, for indexes
  written before and after such a change).
- The compacted sub-index (`LSMTreeIndexCompacted`) persists `binaryKeyTypes` into its own page header, so old and
  new files would coexist; the reader has to honour whatever each file declares.
- `LSMTreeIndexExternalSorter` writes `binaryKeyTypes` into spill runs and rejects a run whose types differ
  (`!Arrays.equals(runKeyTypes, binaryKeyTypes)`), so a bulk build spanning the change would need care.
- Anything that reports key types outward - `getBinaryKeyTypes()`, `TypeIndex`, `MultiIndexCursor` - should keep
  reporting the **schema** type. #5677 split `declaredKeyTypes` (schema, persisted, reported) from `binaryKeyTypes`
  (on-page) for exactly this reason: reporting the storage type from `getBinaryKeyTypes()` broke coercing
  `WHERE link = "#150:5"` into a RID and threw `ClassCastException`. The same trap exists here.

## Suggested first step

Measure before changing anything. `EdgeUniqueIndexLsmVsHashBenchmark` - the benchmark whose construction surfaced
#5677, and which could not run its HASH arm until that fix landed - is the natural place: build a large
`(@out, @in)` unique index both ways and compare on-disk size, page count and lookup throughput. If the saving on
LSM is small relative to the per-entry overhead the LSM format already carries, this is not worth the compatibility
surface described above and the issue should be closed on the numbers.

Follow-up to #5677 / PR #5685.

robfrank:
Closes #5703

## The measurement the issue asked for, first

`LsmLinkKeyEncodingBenchmark` (new, `@Tag("benchmark")`) builds a `(@out,@in)` UNIQUE `LSM_TREE` index over 50k
edges and reports what the LINK keys cost under each encoding. The two figures are computed by serializing the very
RIDs that were indexed both ways, so the comparison does not depend on which encoding the build under test happens
to use:

```
===== #5703 LSM LINK key encoding (edges=50000) =====
  entries indexed        50000
  key bytes fixed-width  1200000 (24.00 /entry)
  key bytes compressed   294179 (5.88 /entry)
  saving on key bytes    75.5%
```

The rest of the entry is a 4-byte pointer slot, 2 null flags, a value count and the (already compressed) value RID.
Measured on the unmodified engine, 50k entries filled **7 mutable pages of 262144 bytes** = ~36.7 bytes/entry, of
which 24 were the keys. So the saving is not marginal relative to the per-entry overhead the LSM format already
carries: it is roughly **half of everything the index occupies**, and the same 50k entries now fit in **4 pages**.
That is the test the issue set for deciding between "do it" and "close it on the numbers", and it comes out on the
"do it" side.

## What changed

`LSM_TREE` now stores a `LINK` key column as `TYPE_COMPRESSED_RID` (two varints) instead of `TYPE_RID`
(`putInt(bucketId)` + `putLong(position)`, 12 bytes fixed), matching what `HASH` has done since #5677.

The split #5677 introduced for the same reason is mirrored here, keeping the field name the LSM side already uses:

* `binaryKeyTypes` keeps its existing meaning - the **schema** type. It is what `getBinaryKeyTypes()` reports
  outward, what every typed comparison uses and what a caller's key is narrowed to. Nothing about it changes.
* `storageKeyTypes` (new) is the encoding written **on the page**. Only the five byte-level serialize/deserialize
  sites in `LSMTreeIndexAbstract` and the two page-cursor constructions use it.

(The two classes name these the other way round - `HashIndexBucket.binaryKeyTypes` is its *on-page* array. Both
report the schema type outward; a note on the new field says so, since reading them side by side is otherwise a
trap.) The `TYPE_RID -> TYPE_COMPRESSED_RID` mapping and its inverse now live in `BinaryTypes` and both index
families call it, rather than each keeping a copy.

### Why this is safe for an *ordered* index

The issue asked for this to be proven rather than assumed. It holds because **the LSM tree never compares raw key
bytes**: `compareKey`/`compareKeys` deserialize both operands and compare typed values, so key order is
`RID.compareTo`'s and is unaffected by how many bytes the RID took on the page. The only byte-wise comparison in the
package (`comparator.compareBytes`) is the `TYPE_STRING` fast path, which no LINK column reaches. `BinaryComparator`
already ordered `TYPE_COMPRESSED_RID` and `TYPE_RID` identically, including a mixed pair.

The four compatibility hazards the issue listed are handled as follows:

* **`getBinaryKeyTypes()`, `TypeIndex`, `MultiIndexCursor`, `FetchFromIndexStep`** keep seeing `TYPE_RID`, so
  coercing `WHERE link = "#150:5"` into a RID still works. That is the `ClassCastException` #5677 hit; there is a
  test for it here.
* **The page-0 header now persists the STORAGE type**, and `onAfterLoad` honours whatever it reads. An index written
  before this change declares `TYPE_RID`, so it keeps reading and writing its fixed-width keys forever - no
  migration, no rewrite, no mixed-encoding file. The declared types are recovered from the storage types
  (`TYPE_COMPRESSED_RID -> TYPE_RID`), which is unambiguous because no `Type` declares `TYPE_COMPRESSED_RID`.
* **`LSMTreeIndexCompacted` never reads its own header** - it inherits the mutable's arrays in `onAfterLoad` and
  receives them in `createNewForCompaction`. Both now carry the storage types, so every file of one index keeps a
  single encoding, including a compacted file produced after an upgrade. `splitIndex` carries them over too.
* **`LSMTreeIndexExternalSorter`** compares its spill runs against `index.getBinaryKeyTypes()`, which still returns
  the declared types, so a bulk build spanning the change is unaffected (its spill format is unchanged and is
  per-build anyway). Making the spill runs compressed too would be a separate, purely local optimisation.

Compaction is where an old-encoding index would meet new code, and it does not mix: the compactor reads keys as
deserialized values and hands them to `appendDuringCompaction`, which re-serializes them under the target file's own
storage types.

## Test plan

New: `engine/src/test/java/com/arcadedb/index/lsm/Issue5703LsmLinkKeyCompressedRidTest.java` (7 tests, 0.7 s).
Before the fix 5 of the 7 pass and 2 fail exactly on the encoding, so the ordering/lookup assertions are known to be
exercising real behaviour rather than passing vacuously.

- [x] `linkKeyIsStoredCompressedYetStillReportedAsRid` - the page uses `TYPE_COMPRESSED_RID`, `getBinaryKeyTypes()`
      still answers `TYPE_RID`, `getKeyTypes()` still answers `LINK`, and the bytes the first entry of page 0 spends
      on its key (read back through the index's own `getSerializedKeySize`) are below the fixed-width cost.
- [x] `orderedIterationOverLinkKeysFollowsRidOrder` - 400 LINK keys whose positions cross the 1-, 2- and 3-byte
      varint widths: ascending iteration equals the RID-sorted list, descending equals its reverse, and inclusive
      and exclusive `range()` bounds select exactly the expected slices.
- [x] `everyLinkKeyIsFoundByAnExactLookup` - all 400 keys resolve to the right record.
- [x] `compactionPreservesLinkKeyOrderAndLookups` - 2000 keys, forced `compact()`, then order and every lookup again.
- [x] `compositeEndpointUniqueIndexStillRejectsDuplicates` - the `(@out,@in)` guard still throws
      `DuplicatedKeyException`, the pair is found, and the reversed pair is a different key.
- [x] `aLinkKeyIsStillCoercedFromItsStringForm` - `SELECT ... WHERE ref = "#1:0"` (#5677's regression).
- [x] `anIndexWrittenWithFixedWidthKeysKeepsUsingThem` - creates the index, closes the database, rewrites page 0's
      key-type byte back to `TYPE_RID` (asserting its current value first, so the test fails loudly if the page-0
      layout ever moves), reopens, and checks that the index reports the legacy storage type, still orders 200 keys
      correctly and writes keys at the full 13 bytes.
- [x] Full `engine` unit lane (`-DexcludedGroups=slow,benchmark,vector -Dsurefire.includes=**/*Test.java`).

## Caveat worth naming

The page header self-describes, so any engine carrying this change reads both encodings. An **older** engine opening
a database whose index was created after this change would map header byte 13 through `Type.getByBinaryType`, get
`null` and fail - the usual one-way property of a storage-format change, recorded here rather than left to be
discovered. Existing databases are untouched.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/hash/HashIndexBucket.java
+++ b/engine/src/main/java/com/arcadedb/index/hash/HashIndexBucket.java
@@ -979,7 +979,7 @@ public class HashIndexBucket extends PaginatedComponent {
    * and hence half the pages - for free. See issue #5677.
    */
   static byte storageKeyType(final byte declaredBinaryType) {
-    return declaredBinaryType == BinaryTypes.TYPE_RID ? BinaryTypes.TYPE_COMPRESSED_RID : declaredBinaryType;
+    return BinaryTypes.getIndexStorageType(declaredBinaryType);
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndex.java
@@ -831,7 +831,7 @@ public class LSMTreeIndex implements RangeIndex, IndexInternal {
         final String newName = mutable.getName().substring(0, last_) + "_" + System.nanoTime();
 
         final LSMTreeIndexMutable newMutableIndex = new LSMTreeIndexMutable(this, database, newName, mutable.isUnique(),
-            database.getDatabasePath() + File.separator + newName, mutable.getKeyTypes(), mutable.getBinaryKeyTypes()
+            database.getDatabasePath() + File.separator + newName, mutable.getKeyTypes(), mutable.getStorageKeyTypes()
             , pageSize,
             LSMTreeIndexMutable.CURRENT_VERSION, compactedIndex);
         newMutableIndex.setStoreTermFrequency(mutable.isStoreTermFrequency());
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexAbstract.java
@@ -88,8 +88,13 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
   protected       byte[]           binaryKeyTypes;
   /**
    * Binary type actually WRITTEN on the page for each key column. It differs from {@link #binaryKeyTypes} only for
-   * LINK columns - see {@link #storageKeyType(byte)} - and is persisted in the page-0 header, so each file keeps
-   * whatever encoding it was created with (#5703).
+   * LINK columns - see {@link BinaryTypes#getIndexStorageType(byte)} - and is persisted in the page-0 header, so each
+   * file keeps whatever encoding it was created with (#5703).
+   * <p>
+   * The remapping is safe for an ORDERED index because the LSM tree never compares raw key bytes: {@code compareKey}
+   * and {@code compareKeys} deserialize both operands and compare typed values, so key order is
+   * {@link RID#compareTo}'s and is unaffected by how many bytes the RID took on the page. (The one byte-level
+   * comparison, {@code comparator.compareBytes}, is the {@code TYPE_STRING} fast path, which no LINK column reaches.)
    * <p>
    * NOTE for anyone reading the HASH implementation alongside this one: the same two concepts are named the other
    * way round there. {@code HashIndexBucket.binaryKeyTypes} is the ON-PAGE encoding and its {@code declaredKeyTypes}
@@ -156,7 +161,7 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
     this.storageKeyTypes = new byte[keyTypes.length];
     for (int i = 0; i < keyTypes.length; i++) {
       this.binaryKeyTypes[i] = keyTypes[i].getBinaryType();
-      this.storageKeyTypes[i] = storageKeyType(this.binaryKeyTypes[i]);
+      this.storageKeyTypes[i] = BinaryTypes.getIndexStorageType(this.binaryKeyTypes[i]);
     }
 
     this.nullStrategy = nullStrategy;
@@ -237,19 +242,6 @@ public abstract class LSMTreeIndexAbstract extends PaginatedComponent {
    */
   byte[] getStorageKeyTypes() {
     return storageKeyTypes;
-  }
-
-  /**
-   * The per-column encoding this index writes on its pages, from what the schema declared.
-   * <p>
-   * The remapping ({@code TYPE_RID} -> {@code TYPE_COMPRESSED_RID}, see
-   * {@link BinaryTypes#getIndexStorageType(byte)}) is safe for an ORDERED index because the LSM tree never compares
-   * raw key bytes: {@code compareKey} and {@code compareKeys} deserialize both operands and compare typed values, so
-   * key order is {@link RID#compareTo}'s and is unaffected by how many bytes the RID took on the page. See #5703,
-   * and #5677 for the HASH sibling.
-   */
-  static byte storageKeyType(final byte declaredBinaryType) {
-    return BinaryTypes.getIndexStorageType(declaredBinaryType);
   }
 
   /**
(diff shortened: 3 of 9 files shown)


robfrank:
### CI note: the one red lane is pre-existing, not this change

Every check on `20dd43e5` is green except `ha-integration-tests`, which fails on:

```
Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact:121
  » AssertionError: Update returned null: record=1 seq=10
Tests run: 210, Failures: 0, Errors: 1, Skipped: 4
```

Three independent reasons it is not this PR:

1. **The lane is already red on `main`.** The three most recent `main` runs
   (`31325190820`, `31323116238`, `31322105607`) all have `ha-integration-tests` failing, with the
   sibling flake `Issue5410AbandonedTicketReleaseIT` in the most recent one. Both are known-open
   flaky HA tests.
2. **The failing test cannot reach the changed code.** It creates `Doc` with `BUCKETS 1` and two
   plain properties and never creates an index of any kind - let alone an `LSM_TREE` one with a
   `LINK` key column, which is the only thing this PR alters. It exercises `TX_PAGE_SLOT_MERGE`
   under concurrent update/delete churn on a single bucket page.
3. **Every lane that does exercise indexes is green here**: `unit-tests`, `slow-unit-tests`,
   `integration-tests`, `vector-unit-tests`, `opencypher-tck-tests`, `builder-tests` and all six
   e2e suites.

Locally the full `engine` unit lane was also run on this branch:
`Tests run: 11362, Failures: 0, Errors: 0, Skipped: 23`, plus `com.arcadedb.index.**` (1003 tests)
re-run after the last review commit.

PR #6023, opened 2026-08-09
fix(#5853): bound GraphQL selection-set nesting depth and stop classifying unparsable queries as read-only

### LINKED ISSUE #5853: GraphQL analyze(): StackOverflowError on an 18 KB nested query, and parse failure classified as READ (fail-open)
## Summary

`GraphQLQueryEngine.analyze()` has two problems in the same ten lines:

1. It throws `StackOverflowError` on an ~18 KB query with deeply nested selection sets. The `catch` only covers `ParseException`, and a `StackOverflowError` is an `Error`.
2. When parsing fails it classifies the query as `READ` / idempotent. That is a fail-open default for a value used elsewhere as a read-only gate.

Version: 26.8.1.

## The method

```java
// graphql/src/main/java/com/arcadedb/graphql/query/GraphQLQueryEngine.java:75
private static Set<OperationType> detectGraphQLOperationTypes(final String query) {
  try {
    final Document doc = GraphQLParser.parse(query);
    for (final Definition def : doc.getDefinitions())
      if (def instanceof OperationDefinition op && !op.isQuery())
        return Set.of(OperationType.CREATE, OperationType.UPDATE, OperationType.DELETE);
  } catch (final ParseException e) {
    // Fall through to default READ — execution will report the parse error
  }
  return CollectionUtils.singletonSet(OperationType.READ);
}
```

## 1. StackOverflowError

Bisected on 26.8.1 (Temurin 21.0.3, default stack):

| nesting depth | payload | result |
|---|---|---|
| 3084 | 18.5 KB | classified |
| **3085** | **18 515 bytes** | **StackOverflowError** |

```java
final String q = "{ " + "a { ".repeat(3085) + "b" + " }".repeat(3085) + " }";
db.getQueryEngine("graphql").analyze(q);   // StackOverflowError
```

```
com.arcadedb.graphql.parser.JJTGraphQLParserState.openNodeScope(JJTGraphQLParserState.java:71)
com.arcadedb.graphql.parser.GraphQLParser.Selection(GraphQLParser.java:504)
com.arcadedb.graphql.parser.GraphQLParser.SelectionSet(GraphQLParser.java:448)
com.arcadedb.graphql.parser.GraphQLParser.Field(GraphQLParser.java:628)
```

Deeply nested selection sets are the textbook GraphQL denial-of-service shape, which is why most GraphQL servers ship a depth limit. Here it costs 18 KB, and the error escapes a method whose callers have no reason to expect an `Error` from a classification helper.

Related: the openCypher parser overflows on a similar shape, reported separately. The SQL parser handles the equivalent nesting cleanly, so the safe behaviour already exists in this codebase.

## 2. Parse failure classifies as READ

Measured, same build:

| input | `isIdempotent()` | operations |
|---|---|---|
| `mutation { createV(name: "x") { name } }` | `false` | CREATE, UPDATE, DELETE |
| `mutation { createV(name: "x" { name } }` (malformed) | **`true`** | **READ** |
| `{{{ this is not graphql` | **`true`** | **READ** |

The in-code comment justifies this with "execution will report the parse error", which holds for `GraphQLQueryEngine.query()`, where the same parser runs again and fails.

The concern is the contract rather than that specific path. `isIdempotent()` is consumed as a read-only decision elsewhere in the codebase — for example in the `mcp` module:

```java
// mcp/src/main/java/com/arcadedb/mcp/tools/SampleRecordsTool.java:145
final QueryEngine.AnalyzedQuery analyzed = database.getQueryEngine("sql").analyze(query);
if (!analyzed.isIdempotent())
  throw new SecurityException("Generated sample query is not read-only");
```

That is an authorization-shaped use, and for those the safe default on "I could not understand this input" is to deny, not to allow. **I could not turn this into an end-to-end bypass on 26.8.1** — execution re-parses and fails — so this is reported as a fail-open default worth inverting, not as an exploitable hole.

## Suggested fix

For the overflow, bound the nesting depth and report exceeding it as a `ParseException`, so it stays a 400 and gains a knob. A GraphQL depth limit is standard and would also cover the execution path, not just `analyze()`.

For the classification, return the conservative answer when the query cannot be parsed:

```java
} catch (final ParseException e) {
  // Cannot classify: assume the worst so an idempotency gate denies rather than admits.
  return Set.of(OperationType.CREATE, OperationType.UPDATE, OperationType.DELETE);
}
```

Execution still reports the real parse error, so the message the user sees does not change; only the answer given to a caller asking "is this read-only?" does.

robfrank:
## What does this PR do?

Fixes two related problems in `GraphQLQueryEngine.analyze()`:

1. **StackOverflowError on deep nesting.** `GraphQLParser.parse()` is an unbounded JavaCC/JJTree
   recursive-descent parser. A GraphQL document with enough nested `{ ... }` blocks (selection sets, or
   object/interface/input type bodies, which use the same delimiter) makes the parser re-enter its
   `SelectionSet`/`Field` grammar rules once per nesting level with no bound. ~3085 levels (an ~18.5 KB
   payload) exhausts the default JVM thread stack, and the only `catch` in the classification helper
   covered `ParseException` - `StackOverflowError` is an `Error` and escaped uncaught.
2. **Fail-open classification on parse failure.** When parsing failed, `detectGraphQLOperationTypes()`
   fell through to its default return value of `READ` / idempotent. `isIdempotent()` is consumed elsewhere
   as an authorization-shaped read-only gate (for example `SampleRecordsTool` in the `mcp` module), so "I
   could not parse this" answered "yes, safe to treat as read-only" instead of denying.

### Fix

1. **Depth guard** in `GraphQLParser.parse()`: before the recursive-descent parse runs, a new
   `checkNestingDepth()` pre-scans the document with a dedicated `GraphQLParserTokenManager` and counts
   `{`/`}` nesting depth. Depth past a new configurable bound, `GlobalConfiguration.GRAPHQL_MAX_NESTING_DEPTH`
   (`arcadedb.graphql.maxNestingDepth`, default `200`, `SCOPE.DATABASE`), throws a normal `ParseException`
   instead of letting the real parser run deep enough to overflow the stack. GraphQL string literals and
   `#` comments are each lexed as a single token, so a brace inside either is never counted. This mirrors
   the existing `SQL_MAX_EXPRESSION_DEPTH` / `CYPHER_MAX_EXPRESSION_DEPTH` pattern already used for the SQL
   and Cypher parsers.
2. **Conservative classification on parse failure**: `detectGraphQLOperationTypes()`'s
   `catch (ParseException e)` block now returns `Set.of(CREATE, UPDATE, DELETE)` directly instead of
   falling through to the default `READ`. Execution (`query()`/`command()`) still re-parses and reports the
   real syntax error to the caller; this only changes the answer given to a caller asking "is this
   read-only?" *before* execution runs.

### Incidental fix required to build

Touching `graphql.jjt` (needed to keep the grammar source consistent with the checked-in generated parser)
exposed that `ph-javacc-maven-plugin`'s `jjtree-javacc` goal is not exercised by a routine build in this
repo - it skips regeneration whenever the checked-in generated sources are newer than the grammar file,
which is the normal state after a `git checkout`. Forcing a real regeneration surfaced two pre-existing
problems, unrelated to this issue, that a routine build never catches:

- `import com.arcadedb.query.sql.parser.JavaCharStream;` in the grammar's `PARSER_BEGIN` block is dead code
  - the class does not exist anywhere in the repo (the SQL parser migrated off JavaCC to ANTLR some time
  ago) - and breaks compilation as soon as regeneration actually runs. Removed.
- A from-scratch regeneration strips the Apache license headers this project requires from every generated
  file, and reformats unrelated code (for example downgrading the checked-in
  `instanceof ParseException exception` pattern-matching form to a pre-Java-16
  `instanceof ParseException` + explicit cast).

Given that, the `checkNestingDepth()` guard was applied as a hand-patch directly to the checked-in
`GraphQLParser.java` (kept consistent with the grammar source, which documents the same method for anyone
regenerating from scratch), rather than by committing a full from-scratch regeneration of the parser
package. A note was added to the top of `graphql.jjt` documenting this so a future grammar change does not
silently regenerate and lose license headers / formatting again. No other generated file in the package
changed (verified via `git status` after the build).

## Motivation

Issue #5853 reports both problems with reproductions. The StackOverflowError is the textbook GraphQL
denial-of-service shape (unbounded nesting), and the fail-open classification is a defense-in-depth concern
for any caller using `analyze().isIdempotent()` as a read-only authorization gate - the issue notes no
end-to-end bypass was found on 26.8.1 since execution always re-parses and fails, but the default is worth
inverting regardless.

## Related issues

Closes #5853

## Test plan

- [x] `GraphQLParserTest` (new tests):
  - `deeplyNestedSelectionSetIsRejectedAsAParseErrorNotAStackOverflow` - nesting one level past the
    configured max is rejected as `ParseException`, message references `maxNestingDepth`.
  - `nestingAtExactlyTheConfiguredLimitStillParses` - nesting exactly at the configured max still parses
    (guard doesn't clip legitimate depth).
  - `aGenuinelyMalformedQueryStillThrowsAParseException` - sanity check the guard doesn't change unrelated
    malformed-input behavior.
  - `theExactPayloadFromIssue5853NoLongerOverflowsTheStack` - the literal reproduction from the issue (3085
    levels, ~18.5 KB) is rejected as `ParseException`, not `StackOverflowError`.
- [x] `GraphQLAnalyzeClassificationTest` (new file):
  - `aValidReadOnlyQueryIsClassifiedAsIdempotentRead` - sanity check existing behavior is unchanged.
  - `aMalformedQueryIsClassifiedAsNonIdempotentNotRead` / `garbageInputIsClassifiedAsNonIdempotentNotRead` -
    a query that fails to parse is now classified `{CREATE, UPDATE, DELETE}`, not `READ`.
  - `analyzeOfAPathologicallyNestedQueryDoesNotThrowAStackOverflow` - `analyze()` on the pathological
    payload does not crash and classifies conservatively.
  - `aPathologicallyNestedReadQueryFailsAsQueryNotIdempotentNotAStackOverflow` - `query()` (read path)
    rejects the pathological payload with `QueryNotIdempotentException`, not a crash.
  - `aPathologicallyNestedCommandFailsAsANormalParseErrorNotAStackOverflow` - `command()` (which does not
    gate on `analyze()`) still turns the same payload into a normal `CommandParsingException` at real parse
    time.
- [x] Proved every new test is load-bearing: all 4 classification tests were confirmed to fail when the
  `catch` block's fix was temporarily reverted, and all 4 depth-related tests were confirmed to fail when
  the `checkNestingDepth()` call was temporarily disabled.
- [x] `mvn -pl graphql test` (full module): 10 test classes, 41 tests, 0 failures, 0 errors.
- [x] `mvn -pl engine test -Dtest=GlobalConfigurationTest`: 10 tests, 0 failures, 0 errors (new
  `GRAPHQL_MAX_NESTING_DEPTH` entry doesn't break existing config invariants).
- [x] `mvn -pl engine,graphql -am install -DskipTests`: clean compile, and confirmed via `git status` that
  the build does not regenerate any of the other 69 already-committed parser sources (only the 4 files in
  this diff changed).

## Additional Notes

- The new default (`200`) is generous for real-world documents while sitting far below the ~3085-level
  point where a JVM's default stack is at risk, consistent with the existing SQL/Cypher precedent.
- `query()` on a pathologically nested but otherwise well-formed read query now fails fast with
  `QueryNotIdempotentException` instead of running the real parse a second time and reporting a syntax
  error - a minor behavior change explicitly accepted by the issue's own suggested fix, since the caller is
  still told the request was rejected, just with a different exception type/message. `command()` is
  unaffected in this respect since it does not gate on `analyze()`.
- No end-to-end authorization bypass existed before this fix (per the issue: execution always re-parses),
  so this is a defense-in-depth correction of a fail-open default, not a fix for an exploited hole.

## Checklist


robfrank:
Checked both points against the code and history.

**1. query() masking parse errors.** Verified this predates the second commit: `git show 904923e51` shows the `ParseException -> assume-the-worst` classification (and therefore the `QueryNotIdempotentException` masking on `query()`) was already unconditional in the first commit, for any `ParseException`, not just nesting-guard ones. The second commit only extended the same already-shipped pattern to `TokenMgrException` for consistency, it didn't introduce the broader-than-nesting scope.

Went with the diagnostic fix (option b) rather than just documenting scope, since it was cheap to do without adding parsing overhead: `detectGraphQLOperationTypes()` is replaced by a private `classify()` returning `(ops, parseFailure)` in one pass, so `analyze()` and `query()` each still parse the document exactly once, no new lexing versus what `query()` already did before this change. `QueryNotIdempotentException` gets a `(message, cause)` constructor and `query()`'s gate now throws with the real `ParseException`/`TokenMgrException` as cause when that's why classification failed.

Checked `ErrorCategory.of()` for a knock-on effect: `CauseChain.contains(error, QueryNotIdempotentException.class)` matches at depth 0 (the thrown exception itself), before the `CommandParsingException` check ever inspects the cause chain, so wrapping the parse failure as a cause doesn't change the resulting `ErrorCategory`. `ErrorCategoryTest` (11/11) and a grep across other `QueryNotIdempotentException` callers (gremlin, mongodbw, SQL/Cypher engines, server HTTP handler) confirm nothing pattern-matches on message content or constructor arity. Added `aPlainMalformedReadQueryFailsAsQueryNotIdempotentCarryingTheRealParseFailureAsCause`, confirmed it fails (null cause) against the pre-fix code and passes now.

**2. Stale GRAPHQL_MAX_NESTING_DEPTH description.** Confirmed, `GlobalConfiguration.java` still said "'{' ... '}' blocks" only. Updated to describe both delimiter pairs and the Value/ListValue recursion path; no test pins the exact text so this was a plain content fix.

Pushed as `ba863c65b`. Full `graphql` module suite (45/45), `GlobalConfigurationTest` (10/10), and `ErrorCategoryTest` (11/11) all green.

PR #6029, opened 2026-08-09
Fix (#6028): duplicate INSERT execution plan runs

### LINKED ISSUE #6028: INSERT statements run their execution plan twice (executed flag never set by the eager execute() call)
Follow-up from review on #6027 (fixing #5681). Not a correctness bug today, but a latent trap flagged during that review.

## What happens

`InsertStatement.execute()` calls `executionPlan.executeInternal()` directly:

```java
final InsertExecutionPlan executionPlan = createExecutionPlan(context);
executionPlan.executeInternal();
return new LocalResultSet(executionPlan);
```

`LocalResultSet`'s constructor immediately calls `executionPlan.fetchNext(...)`, and `InsertExecutionPlan.fetchNext()` is:

```java
@Override
public ResultSet fetchNext(final int n) {
  if (!executed) {
    executed = true;
    executeInternal();
  }
  ...
}
```

The direct call from `InsertStatement.execute()` never sets `executed = true`, so `fetchNext()`'s guard is still `false` and it unconditionally re-invokes `executeInternal()` a second time - meaning **every `INSERT`, not just `INSERT ... SELECT`, runs its full execution plan twice.**

## Why it hasn't caused visible bugs

Each step in the chain happens to carry enough of its own "already done" state that the second pass is a no-op:
- `CreateRecordStep` tracks `created`/`total` and reports `hasNext() == false` once it has produced its records.
- For `INSERT ... SELECT ... LIMIT n`, the nested select's own `LimitExecutionStep.loaded >= limitVal` short-circuits before ever reaching the scan again.
- For a `LIMIT`-less `INSERT ... SELECT`, the source scan is already naturally exhausted.

So today this is inert - verified in the review discussion on #6027 - but it means every `INSERT` pays for a second, unnecessary trip through the whole steps chain, and it's fragile: any future step that does NOT carry idempotent "already done" state, or that has an observable side effect on each `syncPull()` (logging, a metric increment, a call to something non-idempotent), would silently execute twice per `INSERT`.

## Suggested fix

One of:
- Set `executed = true` inside `executeInternal()` itself (so the direct call from `InsertStatement.execute()` also arms the guard), or
- Drop the eager `executionPlan.executeInternal()` call from `InsertStatement.execute()`'s two overloads entirely, since `LocalResultSet`'s constructor already triggers the first real pull via `fetchNext()`.

Either removes the double execution outright instead of relying on each step happening to be idempotent under a second pull.

## Where

- `engine/src/main/java/com/arcadedb/query/sql/parser/InsertStatement.java` (`execute(Database, Object[], ...)` and `execute(Database, Map, ...)`)
- `engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlan.java` (`fetchNext()`, `executeInternal()`)

## Test plan

A test with a step that counts `syncPull()` invocations (or a plain `INSERT` combined with a counter of executed side effects) confirming the plan runs exactly once end-to-end.

cakeni:
## What does this PR do?

Marks `InsertExecutionPlan` as executed when its eager `executeInternal()` path runs. This prevents the `LocalResultSet` constructor from traversing the same plan a second time when it immediately fetches the first block. The same guard now also remains armed after `reset()` eagerly reruns the plan.

Regression coverage exercises both `InsertStatement.execute` overloads and the reset-then-fetch path, verifying that the plan is pulled exactly once in each case.

## Motivation

`InsertStatement.execute` eagerly runs the plan before constructing a `LocalResultSet`, while `InsertExecutionPlan.fetchNext` used an `executed` flag that was only armed by the lazy path. As a result, every INSERT traversed its execution plan twice. The reset path had the same shape: it eagerly executed the plan but left the flag clear for the following fetch.

## Related issues

Fixes #6028

## Additional Notes

Validation performed on the latest `main`:

- `.\mvnw.cmd -pl engine "-Dtest=InsertStatementExecutionTest,InsertStatementTest,InsertExecutionStepTest,InsertContentEmptyArrayTest,InsertReturnTest" test` (60 tests, 0 failures, 2 skipped)
- `.\mvnw.cmd -pl engine -DskipTests package`
- `git diff --check`

Before the production change, the eager-execution regressions failed with `expected: 1 but was: 2`.

AI assistance was used during investigation or implementation.

## Checklist

- [x] My unit tests cover both eager execution entry points and the reset path affected by this issue

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlan.java
@@ -71,7 +71,7 @@ public class InsertExecutionPlan extends SelectExecutionPlan {
    * to close the returned {@link ResultSet} or for the GC to reclaim it. A failure from {@code close()} itself is
    * suppressed rather than allowed to replace an exception already in flight from the drain loop - see
    * {@link UpdateExecutionPlan#executeInternal()} for the same pattern and its rationale.
-  */
+   */
   public void executeInternal() throws CommandExecutionException {
     executed = true;
     RuntimeException failure = null;
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlan.java
@@ -71,8 +71,9 @@ public class InsertExecutionPlan extends SelectExecutionPlan {
    * to close the returned {@link ResultSet} or for the GC to reclaim it. A failure from {@code close()} itself is
    * suppressed rather than allowed to replace an exception already in flight from the drain loop - see
    * {@link UpdateExecutionPlan#executeInternal()} for the same pattern and its rationale.
-   */
+  */
   public void executeInternal() throws CommandExecutionException {
+    executed = true;
     RuntimeException failure = null;
     try {
       while (true) {
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/InsertStatementExecutionTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/InsertStatementExecutionTest.java
@@ -55,20 +55,27 @@ public class InsertStatementExecutionTest extends TestHelper {
     assertInsertExecutionPlanRunsOnce(statement -> statement.execute(database, Map.of(), null, false));
   }
 
+  @Test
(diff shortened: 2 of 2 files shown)


robfrank:
@claude review

robfrank:
@claude review

robfrank:
@claude review

robfrank:
@claude review

robfrank:
Thank you @cakeni

PR #6030, opened 2026-08-10
fix: recognize IPv6 transition encodings in the reserved-address block-list

lvca:
## Summary

- `ImportSecurityValidator.isBlockedAddress` and `PostServerCommandHandler.isBlockedHost` each carried their own copy of the reserved/private-address check used for `IMPORT DATABASE` and server restore/import URLs. Neither recognized an IPv6 address that merely *encodes* an IPv4 address through a transition mechanism (NAT64 `64:ff9b::/96`, 6to4 `2002::/16`, Teredo `2001::/32`), so a private/loopback/link-local IPv4 expressed that way passed both checks unblocked.
- `IPAddressBlocklist` (already the shared CIDR matcher `LOAD CSV` uses) now unwraps all four IPv4-in-IPv6 encodings - IPv4-mapped, NAT64, 6to4, Teredo - before matching, and both callers delegate to a single shared, cached instance instead of duplicating the flag-based logic. This removes the "keep the two in sync" hazard the old code's own comments called out, and folds in a few reserved ranges (IETF protocol assignments, benchmarking, broadcast) the flag-based checks never covered.
- `GlobalConfiguration.OPENCYPHER_LOAD_CSV_BLOCKED_IP_RANGES`'s default now references the same constant instead of duplicating the CIDR string a third time.

## Test plan

- [x] New unit tests in `IPAddressBlocklistTest` cover the NAT64/6to4/Teredo unwrap (positive: blocked; negative: a transition address encoding a *public* IPv4 stays allowed).
- [x] New end-to-end tests in `ImportSecurityValidatorTest` exercise `validateRemoteURL` with bracketed IPv6-literal URLs.
- [x] New `PostServerCommandHandlerSsrfTest` unit test (`isBlockedHost` demoted from `private` to package-private for direct testability) covers the same cases for the server restore/import path.
- [x] `PostServerCommandHandlerIT` (23 tests) and `OpenCypherLoadCSVTest` (26 tests) still pass unchanged.
- [x] Full `compile`/`test-compile` across `engine`, `integration`, `server` is clean.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -26,6 +26,7 @@ import com.arcadedb.serializer.BinaryComparator;
 import com.arcadedb.serializer.json.JSONObject;
 import com.arcadedb.utility.Callable;
 import com.arcadedb.utility.FileUtils;
+import com.arcadedb.utility.IPAddressBlocklist;
 import com.arcadedb.utility.SystemVariableResolver;
 
 import java.io.ByteArrayOutputStream;
@@ -552,9 +553,7 @@ public enum GlobalConfiguration {
       resolved IP address of the target host and re-checked on every redirect hop to prevent Server-Side Request Forgery (SSRF). \
       Defaults to loopback, private (RFC 1918), link-local (including the cloud metadata address 169.254.169.254), carrier-grade \
       NAT, multicast and reserved ranges. Set to an empty string to disable IP filtering (not recommended).""",
-      String.class,
-      "127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,169.254.0.0/16,0.0.0.0/8,100.64.0.0/10,192.0.0.0/24,198.18.0.0/15,"
-          + "224.0.0.0/4,240.0.0.0/4,255.255.255.255/32,::1/128,::/128,fe80::/10,fc00::/7,ff00::/8"),
+      String.class, IPAddressBlocklist.DEFAULT_RESERVED_RANGES),
 
   OPENCYPHER_ID_BUCKET_BITS("arcadedb.opencypher.idBucketBits", SCOPE.JVM,
       """
--- a/engine/src/main/java/com/arcadedb/utility/IPAddressBlocklist.java
+++ b/engine/src/main/java/com/arcadedb/utility/IPAddressBlocklist.java
@@ -131,14 +131,25 @@ public class IPAddressBlocklist {
 
   /**
    * Returns true if the given address falls inside any configured range. An empty block-list never blocks.
+   * <p>
+   * The raw address is matched first, before any transition-encoding unwrap: a range list can legitimately contain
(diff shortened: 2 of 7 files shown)


lvca:
Addressed both review points:

1. **Residual encoding gap (deprecated IPv4-compatible form)**: fixed in 61d2ea0. `normalize()` now also unwraps `::a.b.c.d` (RFC 4291 2.5.5.1, no `ffff` marker - distinct from the still-current IPv4-mapped form). Since no allocated global-unicast range starts with 96 zero bits, this can't misclassify legitimate public IPv6 traffic. New test `blocksDeprecatedIPv4CompatibleBypass` covers it.
2. **GHSA ID verification**: `GHSA-67m7-7w7g-mpmh` was pulled directly from `gh api repos/ArcadeData/arcadedb/security-advisories/GHSA-67m7-7w7g-mpmh` before writing any code, so it's confirmed correct. `GHSA-4w2m-77c8-83mw`/`CVE-2026-54077` are pre-existing citations in `ImportSecurityValidator`'s class javadoc, unchanged by this PR.

lvca:
Good catch, and confirmed real: reproduced the self-masking bug standalone before touching anything (\`IPAddressBlocklist.parse("::1/128").isBlocked(InetAddress.getByName("::1"))\` returned \`false\`). Fixed in 3219bc9 by matching the raw address against all ranges first, falling back to the normalized/unwrapped form only if nothing matched - exactly the suggested fix. Added a regression test (\`literalIPv6CidrEntryIsNotShadowedByTheUnwrap\`) that fails against the pre-fix code and covers a literal entry for each of the five unwrapped encodings, plus a check that a literal entry for one transition prefix doesn't leak into matching another.

The two minor/non-blocking points (shared definition vs. shared instance across the two callers; javadoc not covering the self-masking case) - agreed as noted, no action needed there.

lvca:
Codacy flagged an ErrorProne reference-equality warning on the \`normalized != raw\` check added in 3219bc9 (\`engine/.../IPAddressBlocklist.java:149\`). Fixed in 6c08858: \`normalize()\` now returns \`null\` instead of the input array when no transition encoding applies, so the caller compares to \`null\` instead of comparing two array references - same logic, no lint trip.

lvca:
Latest review round: no correctness issues, only two non-blocking suggestions.

1. **Memoize \`defaultReservedRanges()\` as a singleton** - applied in e5ea017. It now parses \`DEFAULT_RESERVED_RANGES\` once at class-init and hands out the same instance, instead of relying on every caller remembering to cache it themselves. New test \`defaultReservedRangesIsMemoizedAsASingleton\` asserts identity via \`isSameAs\`.
2. **Duplicate \`RESERVED_ADDRESSES\` field declaration across \`ImportSecurityValidator\`/\`PostServerCommandHandler\`** - the review itself judged this not worth changing ("unavoidable given the module boundary... not worth changing"), and I agree: it's a one-line field pointing at the one shared definition, not divergent logic, so there's nothing left to consolidate without merging engine/integration/server module boundaries.

Review otherwise: correctness, test coverage, and security sections all clean with no findings.

PR #6034, opened 2026-08-10
fix(cypher): compare graph entities by identity

### LINKED ISSUE #6010: [Cypher] a GraphAnalyticalView makes the pipeline drop a WHERE a <> c inequality between two chain nodes

justinblethrow-cloud:
## Summary

- compare graph entities by record identity before considering their concrete Java classes
- preserve `=` and `<>` semantics when a Graph Analytical View (GAV) vertex is compared with a regular online transaction processing (OLTP) vertex
- add an end-to-end regression that compares ordinary and GAV-backed results and verifies the GAV provider was selected

## Root cause

The traditional Cypher plan can bind one chain endpoint as a regular vertex and another as a `GAVVertex`. `ComparisonExpression` previously used `equals()` only when both operands had the same concrete Java class. It therefore treated two wrappers for the same graph record as different, making `a <> c` incorrectly true for a self-path.

## Tests

```bash
./mvnw -pl engine -Dtest='GraphAnalyticalViewTest#cypherGAVPreservesNodeInequalityBetweenChainVariables+cypherTraditionalPathWithWhereAndGAV' test
```

Fixes #6010

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/ComparisonExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/ComparisonExpression.java
@@ -18,6 +18,7 @@
  */
 package com.arcadedb.query.opencypher.ast;
 
+import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.function.graph.IdFunction;
 import com.arcadedb.query.opencypher.query.OpenCypherQueryEngine;
@@ -144,6 +145,18 @@ public class ComparisonExpression implements BooleanExpression {
       final long leftEncoded = IdFunction.encodeRidAsLong(new RID(leftStr));
       final long rightEncoded = rightNum.longValue();
       return numericCompare(leftEncoded, rightEncoded);
+    }
+
+    // Graph entities compare by record identity even when different implementations represent them.
+    // This is common when a Graph Analytical View vertex is compared with a regular OLTP vertex (#6010).
+    if (left instanceof Identifiable leftIdentifiable
+        && right instanceof Identifiable rightIdentifiable) {
+      final boolean equal = leftIdentifiable.getIdentity().equals(rightIdentifiable.getIdentity());
+      if (operator == Operator.EQUALS)
+        return equal;
+      if (operator == Operator.NOT_EQUALS)
+        return !equal;
+      return null;
     }
 
     // Temporal comparison. Coerce native java.time / java.util.Date operands into Cypher temporal
--- a/engine/src/test/java/com/arcadedb/graph/olap/GraphAnalyticalViewTest.java
+++ b/engine/src/test/java/com/arcadedb/graph/olap/GraphAnalyticalViewTest.java
@@ -1167,6 +1167,64 @@ class GraphAnalyticalViewTest extends TestHelper {
     gav.drop();
   }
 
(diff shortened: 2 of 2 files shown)


lvca:
@justinblethrow-cloud thanks!!!

PR #6035, opened 2026-08-10
fix(sql): use CI index for lowercased BETWEEN

### LINKED ISSUE #6033: BETWEEN on a case-insensitive (COLLATE CI) index doesn't use the index when wrapped in LOWER()
## Summary

Follow-up from #5966 (fixed in #6018). #6018 taught the planner to recognize a bare `field BETWEEN a AND b` as index-range-eligible, including on `COLLATE CI` indexes:

```sql
CREATE INDEX ON Product (name COLLATE CI) NOTUNIQUE;
SELECT * FROM Product WHERE name BETWEEN 'a' AND 'c'   -- uses the index (fixed by #6018)
```

However, the equivalent query with the common `LOWER()`-wrapped pattern still falls back to a full bucket scan:

```sql
SELECT * FROM Product WHERE LOWER(name) BETWEEN 'a' AND 'c'   -- does NOT use the index
```

## Root cause

`BinaryCondition.isIndexAware()` has two branches: one for a direct column comparison, and a second CI-index optimization that matches `field.toLowerCase() = 'value'`-style expressions via `isFieldWithLowerCaseMethod()` (see `engine/src/main/java/com/arcadedb/query/sql/parser/BinaryCondition.java`). `BetweenCondition.isIndexAware()` (added in #6018) only mirrors the first branch - it doesn't have an equivalent check for the `LOWER()`-wrapped case, so it falls through to the `BooleanExpression` default (`false`) and the planner picks a full scan instead.

## Suggested fix

Extend `BetweenCondition.isIndexAware()` with a second branch analogous to `BinaryCondition`'s, reusing `isFieldWithLowerCaseMethod()` (or a shared helper extracted from it) against `first`, gated the same way (`info.isCaseInsensitive()` and both bounds `isEarlyCalculated`).

## Environment

Discovered during code review of PR #6018 (merged, commit `d86debf3`), which fixed #5966. Not implemented there since it's outside that issue's exact repro (a plain indexed column) - filed separately per the reviewer's own suggestion ("worth a follow-up issue if that pattern shows up in practice").

justinblethrow-cloud:
## Summary

- make `BetweenCondition` recognize `field.toLowerCase() BETWEEN low AND high` as range-index-aware for a case-insensitive (`COLLATE CI`) index
- reuse the existing lowercased-field matcher from `BinaryCondition`
- retain the existing requirements that the index support ordered ranges and both bounds be calculable before row evaluation
- add plan and result coverage proving the query uses the index and preserves inclusive, case-folded range semantics

## Syntax note

The issue example uses `LOWER(field)`, but the current SQL engine reports `Unknown function name 'LOWER'` for that spelling. ArcadeDB's supported equivalent is `field.toLowerCase()`, which reproduces the full-scan planner gap described in the issue. This PR fixes that valid query form without making an otherwise-unregistered function work only for one index-backed plan. I recorded the distinction in the issue for follow-up if a standard SQL `LOWER()` alias is desired.

## Tests

```bash
./mvnw -pl engine -Dtest='CaseInsensitiveIndexTest,Issue5966BetweenUsesIndexTest' test
```

17 focused tests passed.

Fixes #6033

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/BetweenCondition.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/BetweenCondition.java
@@ -180,9 +180,11 @@ public class BetweenCondition extends BooleanExpression {
   public boolean isIndexAware(final IndexSearchInfo info) {
     if (!info.allowsRange())
       return false;
-    if (!first.isBaseIdentifier() || !info.getField().equals(first.getDefaultAlias().getStringValue()))
+    if (!second.isEarlyCalculated(info.getContext()) || !third.isEarlyCalculated(info.getContext()))
       return false;
-    return second.isEarlyCalculated(info.getContext()) && third.isEarlyCalculated(info.getContext());
+    if (first.isBaseIdentifier() && info.getField().equals(first.getDefaultAlias().getStringValue()))
+      return true;
+    return info.isCaseInsensitive() && BinaryCondition.isFieldWithLowerCaseMethod(first, info.getField());
   }
 
   public Expression resolveKeyFrom(final BinaryCondition additional) {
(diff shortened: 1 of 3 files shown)


lvca:
Already fixed in #6036

PR #6042, opened 2026-08-10
Improve security in polyglot host-class lookup filtering

lvca:
## Summary
- Host-class checks in the polyglot script sandbox now also examine a resolved class's inherited superclasses and interfaces, not just its own name, when deciding whether a script may resolve it via `Java.type(...)`.
- Previously, a subclass living in an allow-listed package could inherit a capability from a denied ancestor type without matching any deny-list entry by name (name-only equality matching missed inheritance).
- The check is scoped to "precise" deny entries (a bare class name or a `Type$*` nested-type pattern) so it does not spuriously reject classes that merely implement a common marker interface living in a denied package (e.g. `Serializable`).

## Test plan
- [x] Added regression tests in `HostClassLookupFilterTest` covering the subclass-inheritance bypass, a false-positive guard for common marker interfaces, and a caller-supplied deny-list entry with a subclass
- [x] `mvn -pl engine -am test -Dtest=HostClassLookupFilterTest` — 11/11 pass
- [x] `mvn -pl engine -am test -Dtest="com.arcadedb.query.polyglot.**,com.arcadedb.schema.trigger.**,*Trigger*,*Polyglot*,*Function*"` — 2660/2660 pass, 0 failures/errors

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/polyglot/HostClassLookupFilter.java
+++ b/engine/src/main/java/com/arcadedb/query/polyglot/HostClassLookupFilter.java
@@ -18,7 +18,15 @@
  */
 package com.arcadedb.query.polyglot;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
 
 /**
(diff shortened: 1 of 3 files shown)


lvca:
Thanks for the review. Addressed the main finding:

**Unbounded cache growth** - fixed in d686d6908. Swapped the `ConcurrentHashMap` for the codebase's existing `LRUCache` (bounded at 256 entries, `Collections.synchronizedMap`-wrapped), matching the pattern already used by `CypherStatementCache` and `GraphBatch`'s RID caches. Also switched from `computeIfAbsent` to the get/then/put pattern `CypherStatementCache` uses for consistency - redundant computation under a race is harmless here since the check is a pure function of the resolved class hierarchy.

On the two minor points:
- **Classloader assumption**: added a doc comment on `inheritsAPreciselyDeniedType` pointing at `GraalPolyglotEngine`'s builder as the reason the shared-classloader assumption holds today, so it doesn't silently rot if that changes.
- **`isPreciseTypePattern("")`**: leaving as-is. As you noted, `matches()` already short-circuits on an empty pattern, and nothing in the constructor path can produce one - guarding against an input that can't occur would be speculative.

lvca:
Thanks for the detailed second pass. Addressed:

1. **Residual gap (wildcard-only-denied ancestors)** - agreed this is real but design-level, not a concrete bypass today (confirmed no instance reachable from `ScriptTriggerExecutor.ALLOWED_PACKAGES`). Filed #6045 to track auditing the JDK classpath and evaluating a marker-interface-exclusion approach instead of a wholesale wildcard exclusion, since folding that into this PR risks reintroducing the exact collateral-damage false positives (`Serializable` etc.) this fix guards against.

2. **Classloader assumption not exercised end-to-end** - fixed in ccef481c5. Added `aResourceBundleSubclassIsDeniedThroughARealContext` to `GraalPolyglotEngineHostClassLookupTest`, which runs `Java.type('java.util.PropertyResourceBundle')`/`ListResourceBundle` through an actual `Context.eval(...)` rather than only unit-testing the filter directly.

3. **Release notes** - intentionally not touched. Per project convention, GHSA credit/release notes are added at release time (current dev version is 26.9.1, not 26.8.1), not retroactively into a past release doc mid-development.

lvca:
Third pass:

1. **`catch (Throwable)` too broad** - fixed in 0747cb970. Narrowed to `ClassNotFoundException | LinkageError`, the only expected failure shapes for a name that already passed the by-name deny/allow checks. A bare `Throwable` catch would have silently treated a real VM error (e.g. `OutOfMemoryError`) as "not denied," which isn't a safe default for a security check.

2. **`computeIfAbsent` for atomicity** - deliberately not changed. `Collections.synchronizedMap`'s `computeIfAbsent` holds the map's lock for the full duration of the mapping function, so it would serialize every concurrent script/trigger execution sharing this filter instance behind one thread's `Class.forName` + hierarchy walk. The current get-then-put is correctness-safe (the computation is a pure/idempotent function of the resolved class hierarchy, so a race just recomputes once) and avoids that contention. Given this project's stated performance-over-brevity stance for engine-internal code, I'm keeping get-then-put.

3. **Residual wildcard-ancestor gap** - same point as the previous pass, already tracked in #6045.

Issue #6041, opened 2026-08-10
Add test coverage for DatabaseChecker's index-rebuild lock-contention retry path


lvca:
## Description

Follow-up from the review on #6031 (fix for #5934). `DatabaseChecker`'s auto-fix index-rebuild loop now has ~50 lines of retry/backoff/per-index-isolation logic added in #6031:

```java
for (int attempt = 1; attempt <= LOCK_MAX_ATTEMPTS; attempt++) {
  try {
    database.executeLockingFiles(((IndexInternal) idx).getFileIds(), () -> {
      database.getSchema().dropIndex(idx.getName());
      database.getSchema().buildBucketIndex(...).create();
      return null;
    });
    rebuilt = true;
    break;
  } catch (final NeedRetryException e) {
    lastRetryFailure = e;
    if (attempt == LOCK_MAX_ATTEMPTS)
      break;
    try {
      Thread.sleep(200 + 200L * attempt);
    } catch (final InterruptedException ie) {
      Thread.currentThread().interrupt();
      throw e;
    }
  }
}

if (!rebuilt) {
  rebuildIndexes.remove(idx.getName());
  addWarning(...);
}
```

None of this is exercised by the existing regression test (`CheckDatabaseFixPreservesIndexMetadataTest`), which only covers the successful-rebuild happy path. There is no test that:

- forces a `NeedRetryException` during the drop+create body (e.g. by holding the index's file lock from another thread) to hit the retry-then-succeed branch,
- forces retry exhaustion to hit the warning/`rebuiltIndexes`-removal branch,
- verifies the backoff/interrupt-handling behavior.

This is the newest and most novel logic added in #6031, and it's also where a related correctness issue (#6040) lives, so it's the part of the change most likely to regress silently.

## Suggested approach

A test that induces genuine lock contention (e.g. holding the index's file lock from a second thread via `executeLockingFiles` or the lower-level lock API for a bounded window) to deterministically drive at least one retry, ideally covering both the retry-then-succeed and retry-exhaustion paths. Flagged by review as "understandably hard to simulate reliably" - if a fully deterministic version isn't practical, even a best-effort test documented as occasionally skipping under scheduling jitter would be better than the current zero coverage.

lvca:
Fixed by #6043 (merged): added lock-contention regression coverage for both DatabaseChecker's and RebuildIndexStatement's retry loops, using the same executeLockingFiles primitive the production code uses to deterministically force a LockTimeoutException from a second thread.

Issue #6040, opened 2026-08-10
RebuildIndexStatement/DatabaseChecker's drop+create retry can restart from an already-dropped index


lvca:
## Description

Follow-up from the review on #6031 (fix for #5934). Both `RebuildIndexStatement.buildIndex()` and, since #6031, `DatabaseChecker`'s auto-fix rebuild path wrap a `dropIndex()` + `buildBucketIndex(...).create()` sequence in a bounded retry loop that catches `NeedRetryException` (the type `LockTimeoutException` extends) and retries the **whole** body on failure:

```java
for (int attempt = 1; attempt <= maxAttempts; attempt++) {
  try {
    database.executeLockingFiles(fileIds, () -> {
      database.getSchema().dropIndex(idx.getName());
      database.getSchema().buildBucketIndex(...).create();
      return null;
    });
    break;
  } catch (final NeedRetryException e) {
    // retry
  }
}
```

`NeedRetryException` here can originate from two different places with very different safety properties:

1. **`executeLockingFiles`'s own `tryLockFiles` call**, before the callback runs at all - nothing has been touched yet, so retrying the whole body is safe.
2. **Deep inside `create()`'s bucket scan**, on a *later* attempt, *after* that same attempt's `dropIndex()` has already committed (`create()` scans the bucket to populate the new index, and the bucket's own file is not part of the locked `getFileIds()` set, so a concurrent writer to that bucket can still raise a `ConcurrentModificationException`/`NeedRetryException`).

In case 2, the next retry attempt re-enters the lambda: `dropIndex(idx.getName())` is a safe no-op by then (`LocalSchema.dropIndexInternal` returns early once the name is already gone), but the retried `create()` call runs a **second time** following that same drop, rather than resuming from the point the previous `create()` actually failed. Concretely, the index can end up **missing** (dropped, never successfully recreated) if the retry budget is exhausted on an attempt of this second kind - not merely "not yet rebuilt."

This is a pre-existing pattern in `RebuildIndexStatement.buildIndex()` (silently returns on exhaustion, so the caller doesn't even get a clear signal); #6031 gave `DatabaseChecker` the identical shape, and additionally made the failure more likely to actually surface (it rethrows/warns on exhaustion instead of silently returning), which is why review caught it there.

## Suggested fix

Either:
- Restrict the retry to genuine lock-acquisition failures (distinguish "failed before touching anything" from "failed inside the callback"), or
[...]

Issue #6037, opened 2026-08-10
CI-lowercase index optimization missing from InCondition/ContainsCondition family (field.toLowerCase() IN [...])


lvca:
## Summary

Follow-up from #6033 (fixed by #6036), flagged during code review of that PR.

`BinaryCondition.isIndexAware()` and (as of #6036) `BetweenCondition.isIndexAware()` both recognize the `field.toLowerCase()` pattern against a `COLLATE CI` index and use the index instead of a full bucket scan. The same optimization does not exist for:

- `InCondition.isIndexAware()` - e.g. `field.toLowerCase() IN ['a', 'b', 'c']`
- possibly `ContainsCondition` / `ContainsAnyCondition` / `ContainsAllCondition` for the analogous collection-membership shapes

These currently fall back to a full scan on a CI index even though the equals/range cases are already optimized.

## Suggested fix

Reuse `BinaryCondition.isFieldWithLowerCaseMethod()` (package-private in `com.arcadedb.query.sql.parser`, already shared by `BetweenCondition` since #6036) to add the equivalent CI-lowercase branch to `InCondition.isIndexAware()` and audit the `Contains*Condition` family for the same gap.

## Note

Not a regression from #6036 - this gap predates that PR and is being filed separately per the reviewer's own suggestion to avoid scope creep on the BETWEEN fix.

lvca:
Fixed by #6043 (merged): InCondition.isIndexAware() now recognizes field.toLowerCase() IN [...] against a COLLATE CI index. Audited the Contains*/ContainsAll/ContainsAny/ContainsValue family per the issue's suggestion - none has the analogous gap, since their indexed field is always a Collection/Map, which .toLowerCase() cannot syntactically apply to.

PR #6043, opened 2026-08-10
fix: batch fixes for #6037, #6040, #6041

lvca:
## Summary

- **#6040**: `RebuildIndexStatement.buildIndex()` and `DatabaseChecker`'s auto-fix index-rebuild loop each retry a `dropIndex()` + `create()` body as a single unit inside `executeLockingFiles`. Both charged a plain lock-acquisition failure (`tryLockFiles` times out, nothing has been touched yet) and a failure raised *after* that same attempt's `dropIndex()` already committed (the index is now gone) against the identical small attempt budget. Exhausting retries while in the second state left the index permanently missing rather than merely unrebuilt.

  Fix: both call sites now track (via a flag set inside the locked callback) whether the drop actually committed on any attempt. Once it has, a larger post-drop retry budget applies (`POST_DROP_ATTEMPT_MULTIPLIER` / `LOCK_POST_DROP_ATTEMPT_MULTIPLIER`, 4x) instead of giving up at the lock-acquisition ceiling. `RebuildIndexStatement.buildIndex()` also now **throws** `CommandExecutionException` on exhaustion instead of silently falling off the end of the method - the old behavior let a failed `REBUILD INDEX` look identical to a successful one to the caller. `DatabaseChecker`'s warning message now states explicitly which of the two cases occurred ("the index lock could not be acquired; the index itself is unchanged" vs. "the index was dropped but the rebuild could not complete; it is now missing") instead of a blanket message that was wrong for the second case. `DatabaseChecker` was also missing `.withMaxAttempts(...)` on its `buildBucketIndex()` call entirely (default 1) - `RebuildIndexStatement` already passed it through - so `DatabaseChecker` had no inner absorption of transient bucket-scan contention at all; this is now aligned between the two call sites.

  Considered (and rejected) alternatives: locking the bucket's own file for the full rebuild duration would eliminate the race at the root, but the existing narrow lock scope (index files only) is a deliberate design choice documented at the call site - it lets writers keep hitting the bucket while a rebuild runs. Widening it would trade a rare retry-exhaustion edge case for routinely blocking live traffic on every rebuild, which is a worse trade for the common case. A build-under-a-temporary-name-then-atomic-swap redesign would remove the missing-index window entirely, but no rename primitive exists in the index/schema layer today, and one sub-index temporarily duplicated per bucket during the swap has real correctness implications for UNIQUE validation and the query planner that are out of proportion to a narrow contention window that's now bounded and clearly reported either way.

- **#6041**: added lock-contention regression coverage for both retry loops (previously zero, as the issue notes). Tests hold the target index's file lock from a second thread via the same `DatabaseInternal.executeLockingFiles` primitive the production code itself uses, deterministically forcing at least one `LockTimeoutException`:
  - `DatabaseCheckerIndexRebuildLockContentionTest`: retry-then-succeed (hold just past the single 5s `tryLockFiles` timeout, then release) and retry-exhaustion (hold past the worst-case total of all 5 attempts, ~27.8s) - the latter also asserts the index comes out completely untouched, per #6040's fix.
  - `RebuildIndexLockExhaustionTest`: the new throw-on-exhaustion contract for `REBUILD INDEX`, using `maxAttempts = 1` to keep it to a single 5.3s wait.

  Both are tagged `@Tag("slow")` since they're inherently multi-second (`executeLockingFiles` waits a hardcoded 5s per attempt). Verified each test fails against the pre-fix code and passes against the fix.

- **#6037**: `InCondition.isIndexAware()` gained the same `field.toLowerCase()` CI-index recognition `BinaryCondition`/`BetweenCondition` already have (reusing the existing package-private `BinaryCondition.isFieldWithLowerCaseMethod()`), so `field.toLowerCase() IN [...]` uses a `COLLATE CI` index instead of a full scan. Audited the `Contains*Condition` family (`ContainsCondition`, `ContainsAllCondition`, `ContainsAnyCondition`, `ContainsValueCondition`) as the issue suggested: none has the same gap, because their indexed field is always a `Collection`/`Map` (BY-ITEM/BY-VALUE/BY-KEY indexes), and `.toLowerCase()` cannot syntactically apply to those - only to a scalar `String` field. No code change needed there; noted in the test class Javadoc.

## Test plan

- [x] `Issue6037InToLowerCaseCiIndexTest` (new, TDD): `field.toLowerCase() IN [...]` uses the index on a `COLLATE CI` index, matches the non-indexed equivalent, falls back to scan without CI collation, and falls back to scan when a modifier is chained after `toLowerCase()`. Verified it fails without the fix (`SCAN WITH FILTER` instead of `FETCH FROM INDEX`).
- [x] `DatabaseCheckerIndexRebuildLockContentionTest` (new): retry-then-succeed and retry-exhaustion-leaves-index-untouched, both against real lock contention. Verified the exhaustion test fails against the pre-#6040 `DatabaseChecker` (missing the new warning phrasing).
- [x] `RebuildIndexLockExhaustionTest` (new): `REBUILD INDEX ... WITH maxAttempts = 1` throws `IndexException` (wrapping the new `CommandExecutionException`) on exhaustion and leaves the index untouched/queryable. Verified it fails against the pre-#6040 `RebuildIndexStatement` (old code just returned with no exception).
- [x] Existing `Issue6033BetweenToLowerCaseCiIndexTest`, `CaseInsensitiveIndexTest`, `Issue5966BetweenUsesIndexTest`, `Issue5932IndexedRangeTypedBoundsTest` still pass.
- [x] `CheckDatabaseFixPreservesIndexMetadataTest`, `CheckDatabaseTest`, `CheckDatabaseRecordScopeTest`, `CheckDatabaseSinglePassTest`, `CheckDatabaseProgressTest`, `CheckDatabaseExtendedTest`, `RebuildIndexPreservesNameTest` still pass.
- [x] Full `engine` module test suite (`mvn test -DexcludedGroups=slow,benchmark,vector`, 11505 tests) passes except one pre-existing, unrelated flake (`Issue5326ApplyChangesPendingFlushTest`, a page-pipeline-flush test untouched by this diff - passes in isolation, confirmed not caused by this change).
- [x] Full compile of `engine` succeeds.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -490,3 +490,32 @@ vertex does not cover it. Those lookups now go through `CSRCountUtils.findAccele
 declines a partial view.
 
 [#5757](https://github.com/ArcadeData/arcadedb/issues/5757)
+
+## `CHECK DATABASE FIX` and `REBUILD INDEX` no longer silently lose an index under sustained lock contention (#6040, #6041)
+
+Both `RebuildIndexStatement.buildIndex()` and `DatabaseChecker`'s auto-fix index-rebuild path retry a
+`dropIndex()` + `create()` body as one unit when the index's file lock is contended. That retry loop charged a
+lock-acquisition failure - the lock timed out before the body ever ran, so nothing was touched - and a failure
+raised *after* that same attempt's `dropIndex()` had already committed - the index is now gone - against the
+identical, small attempt budget. Exhausting retries while in the second state left the index permanently
+missing, discoverable only by noticing it was gone.
+
+Both call sites now track whether the drop actually ran on any attempt and switch to a larger retry budget
+(4x) once it has, favoring "never silently lose an index" over a bounded worst-case runtime. Under sustained
+contention, `REBUILD INDEX` and `CHECK DATABASE FIX` can now take several minutes per index rather than the
+previous ~28 seconds before giving up - `CHECK DATABASE FIX` in particular repeats this per affected index, so
+a fix run against a database with several contended indexes can take noticeably longer under load than before
+this change. `RebuildIndexStatement.buildIndex()` also now throws on exhaustion instead of silently returning
+as if the rebuild had succeeded; `DatabaseChecker`'s warning states which of the two cases occurred instead of
+a blanket "lock contention" message that was misleading for the second case.
+
+[#6040](https://github.com/ArcadeData/arcadedb/issues/6040), [#6041](https://github.com/ArcadeData/arcadedb/issues/6041)
+
+## `field.toLowerCase() IN [...]` now uses a `COLLATE CI` index (#6037)
+
+`field.toLowerCase() = 'value'` and `field.toLowerCase() BETWEEN a AND b` already recognized a case-insensitive
+index and used it instead of a full scan; the equally common `IN` shape did not, and fell back to a bucket
+scan even when a `COLLATE CI` index existed. `InCondition.isIndexAware()` gained the same recognition,
+reusing the existing field-pattern check `BinaryCondition`/`BetweenCondition` already share.
+
+[#6037](https://github.com/ArcadeData/arcadedb/issues/6037)
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -52,6 +52,11 @@ import java.util.stream.Collectors;
 public class DatabaseChecker {
   // Matches RebuildIndexStatement.MAX_ATTEMPTS: the lock-contention retry bound for the index-rebuild file lock.
   private static final int          LOCK_MAX_ATTEMPTS = 5;
+  // Matches RebuildIndexStatement.POST_DROP_ATTEMPT_MULTIPLIER (#6040): once an attempt's dropIndex() has actually
+  // committed, exhausting the retry budget leaves the index missing rather than merely unrebuilt, so that state
+  // gets a larger budget than a plain lock-acquisition timeout, where nothing was touched and giving up simply
+  // leaves the index exactly as it was.
+  private static final int          LOCK_POST_DROP_ATTEMPT_MULTIPLIER = 4;
   private final DatabaseInternal    database;
   private       int                 verboseLevel = 1;
   private       boolean             fix          = false;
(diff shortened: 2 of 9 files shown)


lvca:
Thanks for the review. Addressed in ceaa2f66f:

- **Backtick nit**: fixed (`` `REBUILD INDEX myIdx`` `` is now a single well-formed backtick span).
- **Missing comment on `.withMaxAttempts(LOCK_MAX_ATTEMPTS)`**: added, explaining the longer per-attempt hold time and that it was missing entirely before this fix (default `maxAttempts=1`, no inner retry at all).
- **Worst-case latency under sustained contention**: added a section to `docs/release-26.9.1.md` calling this out explicitly as an operator-visible behavior change.
- **Test duplication**: extracted `corruptRecordTypeByte` and the lock-holding thread into shared `TestHelper.corruptRecordTypeByte(...)` / `TestHelper.LockHoldingThread`, and switched `CheckDatabaseFixPreservesIndexMetadataTest`, `DatabaseCheckerIndexRebuildLockContentionTest`, and `RebuildIndexLockExhaustionTest` to use them - eliminating the duplication introduced/multiplied by this PR.

  Deliberately **not** touched: `CheckDatabaseRecordScopeTest` and `CheckDatabaseSinglePassTest` keep their own private `corruptRecordTypeByte`. Those two route through a shared `onRecordPage(RID, BiConsumer)` helper local to `CheckDatabaseRecordScopeTest` that also backs a second corruption shape (`shrinkRecordBuffer`) - a richer abstraction than the flat version this PR's tests need. Refolding that into the new shared helper would mean redesigning an established, unrelated, already-passing test's internal structure well outside this PR's actual scope, for marginal benefit. The duplication this review flagged as newly introduced (a third and fourth copy) is fully eliminated; the two pre-existing copies with the differently-shaped internal abstraction are left alone.

All new/modified tests re-verified passing after these changes (fast suite + the two `@Tag("slow")` contention tests + `CheckDatabaseRecordScopeTest`/`CheckDatabaseSinglePassTest` to confirm the `TestHelper` changes didn't regress them).

lvca:
Thanks, this round found a real bug. Addressed in ce5ab348d:

- **Integer overflow in the post-drop budget**: fixed. `maxAttempts * POST_DROP_ATTEMPT_MULTIPLIER` now goes through `long` arithmetic clamped to `Integer.MAX_VALUE` before narrowing back to `int`, so an extreme user-supplied `maxAttempts` can no longer wrap the budget negative. Agreed this is low-severity (requires a deliberately extreme setting), but it's a genuine correctness bug and the fix is a one-line, zero-risk guard, so no reason not to take it. Didn't add a dedicated regression test for this specific edge: exercising it end-to-end would need either holding a lock for an impractically long time or `maxAttempts` near `Integer.MAX_VALUE / 4`, neither of which is a reasonable thing to assert on in a unit test - the fix is a well-understood, self-evidently-correct arithmetic guard rather than logic that benefits from a test pinning its behavior.
- **Interrupt status not restored**: fixed, `RebuildIndexStatement.buildIndex()`'s `catch (InterruptedException ex)` now calls `Thread.currentThread().interrupt()` before rethrowing, matching `DatabaseChecker`'s existing correct handling of the identically-shaped loop.
- **Tight test margin (5.3s vs 5.0s + 400ms)**: agreed, widened to 6.0s in both `DatabaseCheckerIndexRebuildLockContentionTest.fixRetriesThroughTransientLockContentionAndSucceeds` and `RebuildIndexLockExhaustionTest` (the latter wasn't explicitly called out but has the same underlying risk: a delayed dispatch could let the held lock's release land inside the attempt's window instead of past it). Verified both still pass with the wider margin.
- **Design agreement on rejected alternatives**: noted, no action needed.

All three fixes re-verified: full recompile, the fast `RebuildIndexStatement`/`REBUILD INDEX` test group, and the `@Tag("slow")` `DatabaseCheckerIndexRebuildLockContentionTest` (both methods) all pass.

lvca:
Thanks for the independent pass and for re-confirming the prior fixes.

On the one remaining observation (`WITH maxAttempts = 0` or negative now runs one real attempt before throwing, versus the old code's zero-attempt silent no-op): agreed with your own assessment, and confirming explicitly since I said I'd judge every remaining point rather than leave it implicit. This is not a regression to fix - it's a strict improvement: the old behavior silently claimed success while doing nothing, the new behavior either does real work and can succeed on attempt 1, or fails loudly with a clear message instead of pretending nothing was wrong. `maxAttempts <= 0` is also not a value anyone would deliberately set (the whole point of the setting is "how many times should I retry," and 0 means "don't try at all," which contradicts issuing `REBUILD INDEX` in the first place). No change needed.

No other points remain. All CI checks are green or in their final stretch (`build-and-package` / CodeQL java-kotlin still finishing as of this comment) - will confirm once they complete.

Issue #6044, opened 2026-08-10
Super-node promotion silently changes edge iteration order: interleave the stripe chains instead of concatenating them


lvca:
## Problem

Super-node promotion (#5156, `StripedEdgeList`) silently changes the **order** in which a vertex's edges are iterated. That is a behaviour change no reader opts into and none can detect.

On the classic layout the order is exact **newest-first**: `MutableEdgeSegment.add()` prepends inside the head chunk, and when a chunk fills the new chunk becomes the head, so `EdgeIterator` walks newest chunk → oldest chunk, newest entry → oldest entry.

Once a vertex crosses `GRAPH_SUPERNODE_THRESHOLD` (default 4096) its list is promoted to N stripe chains (default 16). Each chain is still newest-first internally, but `StripedEdgeList.edgeIterator()` composes them with a `MultiIterator`, i.e. it **concatenates**: the whole of stripe 0, then the whole of stripe 1, and so on. Since the stripe is `hash(neighbour RID) % stripes`, the global order becomes effectively random, and the error is proportional to the vertex **degree** — it gets worse as the graph grows.

The practical effect: any reader that takes the first N edges of a vertex to mean "the N most recent" — an extremely common pattern, and the only thing the classic layout ever did — starts returning near-arbitrary rows the moment the vertex passes 4096 edges. No exception, no log line, no way to notice except a user reporting that a record they just created is missing from a list.

We hit exactly this in production: a listing paging the newest 100 chats attached to a location vertex stopped showing newly created ones. It reproduces trivially in the engine — 2000 chats on one location, threshold lowered to force promotion:

| layout | position of the newest edge in the iteration |
|---|---|
| classic | 0 |
| striped | 116 |

## Why it is worth fixing in the engine

Promotion is an internal, automatic, degree-triggered optimisation for **write** throughput. It should be invisible to readers. Right now it is a silent, data-dependent semantic change that fires on precisely the vertices where it hurts most (the biggest ones), and every application built against the classic behaviour is exposed.

Callers can and should stop depending on storage order where correctness matters (we are moving the listing above onto an indexed `ORDER BY`). But "the first N edges are roughly the newest N" is a reasonable property for the engine to preserve across a transparent layout change, and preserving it costs very little.

## Proposal: round-robin the stripe chains

[...]

PR #6050, opened 2026-08-11
fix(#6049): document that GAV neighbor order is unrelated to OLTP order

### LINKED ISSUE #6049: A GraphAnalyticalView returns super-node neighbours in a different order than the OLTP path
Follow-up to #6044 / #6047.

#6044 fixed the OLTP read walks of a promoted super-node (`edgeIterator`, `vertexIterator`, `ridIterator`) to interleave the stripe chains, so the first N neighbours are approximately the newest N. The analytical path was left alone, and it disagrees.

`CSRBuilder` builds the CSR adjacency from `outList.entryIterator()` (`engine/src/main/java/com/arcadedb/graph/olap/CSRBuilder.java`), and `entryIterator()` deliberately kept plain concatenation - correctly, as far as the builder's own walk goes, since it consumes the whole list and concatenation has better locality there. But the **order it consumes in is the order it freezes into the CSR**, and every query later answered from that `GraphAnalyticalView` returns neighbours in it.

So for a vertex past `arcadedb.graph.supernodeThreshold`:

- answered by the OLTP pipeline: neighbours approximately newest-first
- answered from a `GraphAnalyticalView`: neighbours in concatenated stripe order, i.e. grouped by `hash(neighbour RID)`

Same query, same data, two orders, and which one a user gets depends on whether a view happens to exist and whether the planner chose it - a decision the user did not make and cannot see in the result.

Neither order is contractual (#6047 wrote down that only within-a-stripe order is exact), so this is not a wrong answer. It is an *inconsistency*, and it is the same shape as the bug #6044 was about: an internal optimisation quietly changing what a reader sees.

## Options, roughly in order of cost

1. **Build the CSR through the interleaved order** - hand `CSRBuilder` a rotation over the chains instead of `entryIterator()`. Pays the locality cost once, at build time, on a path that is already a long batch job, and every later query off the view matches OLTP. `InterleavedIterator` is generic, so this is a call-site change.
2. **Leave it and document it** on `GraphAnalyticalView` - cheapest, but it re-creates exactly the "order changes under you with no way to notice" situation for view-backed reads.
3. Decide the analytical path has no ordering expectation at all and say so explicitly.

Option 1 looks right, but it should be measured against a large-graph view build before committing to it - the build walks every vertex's full list, which is the case the rotation is worst at (see #6048).

Worth checking at the same time: `toJSON()` and `edgeIteratorConnectedTo()` also still concatenate. `edgeIteratorConnectedTo` is fine by construction (a given neighbour lives in one stripe per generation, so at most one chain contributes and the order is unaffected). `toJSON()` is a debug/export dump - probably fine, but it is the one remaining place where a human reading output would see the raw stripe grouping.

lvca:
Closes #6049

## The issue's diagnosis does not match the code

#6049 (a follow-up to #6044/#6047) reported that a `GraphAnalyticalView` returns a promoted super-node's neighbors in the OLTP "concatenated stripe" order that #6044/#6047 removed from the direct OLTP read walks (`edgeIterator`/`vertexIterator`/`ridIterator`), and proposed rewiring `CSRBuilder` onto the interleaved iterator from #6047 as the fix.

That is not what happens. `CSRBuilder` never preserves the order it consumes `outList.entryIterator()` in:

- Phase C (`Arrays.sort` on every node's forward/backward neighbor slice) unconditionally re-sorts every adjacency list ascending by dense node ID, regardless of the order it was collected in.
- Phase D (BFS/RCM reordering) permutes those dense IDs again for cache locality.

This has been true since the CSR/GAV feature was first merged (#3618) — see `git log -S` on the `Arrays.sort(fwdNeighbors, fs, fe)` line, and the pre-existing `GraphAnalyticalViewTest#sortedNeighbors` test, which already asserts sorted output for a **non-promoted** hub. So implementing the issue's proposed fix (Option 1) would be a no-op on the CSR's actual output — whatever order `entryIterator()` produces is thrown away before a query ever sees it.

## What's actually true, and why it's not a bug

The real property is broader than the issue framed it: `GraphAnalyticalView`'s neighbor order has never had any relationship to OLTP order, for **any** vertex — promoted or not, not just super-nodes. It's a deliberate design choice: the ascending-by-node-ID adjacency is what lets `isConnectedTo`, `countEdgesBetween`, `getMeanEdgesPerConnectedPair`, and triangle counting (`PartitionedTriangleOp`) binary-search or merge-join the CSR arrays instead of scanning them. Matching OLTP's approximate-recency order would break that invariant for a property GAV never claimed to begin with.

The one part of the issue that *is* real: the query planner (`GAVExpandAll`/`GAVExpandInto`, `MatchGAVFusedStep`) can transparently substitute the CSR path for a plain `MATCH`/Cypher expansion when a ready view covers it, so the same query's result order can depend on a decision the user didn't make and can't see — confirmed by tracing `GraphTraversalProviderRegistry` into the Cypher/SQL executors. That part just wasn't undocumented because of stripe concatenation; it was undocumented, period.

## What this PR does instead

Documents the actual contract at the source rather than "fixing" something that isn't broken:

- `GraphAnalyticalView`'s class Javadoc now states explicitly that neighbor order is sorted by dense node ID (further permuted for locality), unrelated to OLTP order, and that the planner may route the same query through either path.
- `CSRBuilder`'s class Javadoc explains why `entryIterator()`'s consumption order is irrelevant to the CSR's output.
- `StripedEdgeList`'s Javadoc (added by #6047) cross-references `GraphAnalyticalView` so the two guarantees don't drift apart again.
- Verified the issue's own follow-up notes on `toJSON()` and `edgeIteratorConnectedTo()`: both are accurate as filed (debug/export dump with one caller; unaffected by concatenation because a neighbor hashes to at most one stripe per generation) — no code change needed there.

## Tests

`Issue6049GAVSupernodeOrderTest` (new) pins the actual behavior against a **promoted** super-node:
1. `gavNeighborOrderIsSortedForAPromotedSupernode` — CSR neighbor order stays strictly ascending by dense node ID after promotion, with nothing lost or duplicated. This is what would fail if `CSRBuilder` were ever wired to forward an interleaved (or any non-sorted) order into the CSR.
2. `gavOrderIntentionallyDivergesFromButLosesNothingRelativeToOltpOrder` — GAV order and OLTP order are genuinely different permutations of the same edge set, confirming the divergence is real but lossless.

Ran locally, both green:
```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- Issue6049GAVSupernodeOrderTest
Tests run: 134, Failures: 0, Errors: 0, Skipped: 0 -- GraphAnalyticalViewTest
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0 -- Issue6044StripeIterationOrderTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- GraphAnalyticalViewRegistryTest
```
Full `engine` module `test-compile`/`compile` clean.

## Test plan
- [x] New regression test proving the actual (correct) invariant, not the issue's diagnosed one
- [x] Existing GAV/olap/#6044 suites pass unchanged
- [x] Full engine compile clean
- [x] No `System.out`/`System.err` left in changed files

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/StripedEdgeList.java
+++ b/engine/src/main/java/com/arcadedb/graph/StripedEdgeList.java
@@ -71,6 +71,12 @@ import java.util.logging.Level;
  * the contract - but the classic layout happened to be exact, so this is written down rather than left to be
  * inferred. The maintenance walks (removal, counting, the integrity checker, export) deliberately keep plain
  * concatenation: order means nothing to them and one cursor beats {@code stripes} of them.
+ * <p>
+ * This guarantee is specific to these OLTP read walks. A {@link com.arcadedb.graph.olap.GraphAnalyticalView}
+ * (GAV) answering the same query goes through none of it: {@code CSRBuilder} sorts every vertex's neighbors
+ * ascending by dense node ID at build time regardless of how they were consumed from the OLTP list - promoted
+ * or not - so a GAV-backed read carries no relationship to this newest-first approximation. See
+ * {@link com.arcadedb.graph.olap.GraphAnalyticalView}'s class Javadoc.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
--- a/engine/src/main/java/com/arcadedb/graph/olap/CSRBuilder.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/CSRBuilder.java
@@ -52,7 +52,12 @@ import java.util.logging.Level;
  *   <li>Detects property types and extracts property values into per-bucket {@link ColumnStore}</li>
  * </ul>
  * <b>Pass 2</b> (single scan): Iterates vertices again to fill CSR neighbor arrays
- * using prefix sums computed from Pass 1 degree counts.
+ * using prefix sums computed from Pass 1 degree counts. Each vertex's out-edges are read via
+ * {@code outList.entryIterator()}, which walks a promoted (striped) super-node's stripe chains
+ * concatenated rather than in any OLTP recency order - but that walk order never reaches a query:
+ * every node's neighbor slice is unconditionally sorted ascending by dense node ID in Phase C below,
+ * then permuted again by the BFS/RCM reordering in Phase D. The consumption order only affects how
+ * warm the OLTP pages are while this scan runs, never what a reader of the finished CSR sees.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -63,6 +63,19 @@ import java.util.logging.Level;
  *   <li>Optional auto-update on transaction commit</li>
  * </ul>
  * <p>
+ * <b>Neighbor order is unrelated to OLTP order.</b> {@link #getVertices}, {@link #getNeighborView} and
+ * {@link #scanNeighbors} always return a node's neighbors sorted ascending by internal dense node ID -
+ * further permuted by a one-time cache-locality renumbering at build time ({@code CSRBuilder}'s BFS/RCM
+ * pass) - never in edge-insertion or OLTP-recency order. This holds for every vertex, promoted
+ * super-node or not: it is what lets {@link #isConnectedTo}, {@link #countEdgesBetween},
+ * {@link #getMeanEdgesPerConnectedPair} and triangle counting binary-search or merge-join the adjacency
+ * arrays instead of scanning them. A query the planner can answer from either the OLTP edge list or a
+ * ready GAV (see {@code GAVExpandAll}/{@code GAVExpandInto}) may therefore return its rows in a different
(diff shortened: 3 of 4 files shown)


lvca:
Responding to the latest review round's three nits, none of which need a code change:

1. **Static config save/restore scope** — this is the exact same `@BeforeEach`/`@AfterEach` save/restore pattern as `Issue6044StripeIterationOrderTest` (byte-for-byte identical block), the precedent test from #6047 for the same `GRAPH_SUPERNODE_THRESHOLD`/`GRAPH_SUPERNODE_STRIPES` config. Since the review itself said "if it does [mirror the existing pattern], this is a non-issue" — confirmed, it does.
2. **Probabilistic `isNotEqualTo` assertion** — already flagged by the prior review round and acknowledged in the second commit message; reviewer agrees no action needed.
3. **Test cost / `@Tag("slow")`** — measured: both tests together run in 0.458s locally. `Issue6044StripeIterationOrderTest`, which does the same one-transaction-per-edge pattern at 4x the edge count (2,000 vs. 500) and is untagged, runs in 0.370s. This test is well inside the same untagged convention.

No correctness/performance/security issues were raised in either review round, and this PR only touches Javadoc + a new test (zero production behavior change). Merging as-is.

Issue #5872, opened 2026-08-07
Closing a database does not reliably cancel an in-flight vector graph build: the caller parks on ForkJoinTask.join()


lvca:
Found while writing a cancellation regression test during the review of #5869 (issue #5577). It is **pre-existing** and was not introduced by that PR - it is kept separate for exactly that reason.

## What happens

`LSMVectorIndex` keeps a dedicated `ForkJoinPool` for graph building, and its whole reason to exist is cancellation. From `releaseBackgroundResources()`:

```java
// Shut down the dedicated graph build pool to cancel any in-progress build operations.
// This interrupts ForkJoinPool workers inside jvector's GraphIndexBuilder.build(),
// which otherwise would not respond to Thread.interrupt() on the parent thread.
final ForkJoinPool pool = graphBuildPool;
if (pool != null && !pool.isShutdown()) {
  pool.shutdownNow();
  pool.awaitTermination(5, TimeUnit.SECONDS);
}
```

**That does not reliably abort the build.** A thread that started a graph build can stay parked on `ForkJoinTask.join()` long after the database has been closed.

## Repro

Start a graph build on a worker thread against an index whose shape makes the build last several seconds (4000 vectors, 128 dimensions, `maxConnections=64`, `beamWidth=400` is enough), wait until progress reports the insertion is genuinely under way, then `db.close()` and join the worker.

Two runs of the same test class back to back on the same machine: one green, one with the build thread **still alive 60 seconds after `db.close()`**. A thread dump puts it in:

```
java.util.concurrent.ForkJoinTask.awaitDone
java.util.concurrent.ForkJoinTask.join
com.arcadedb.index.vector.LSMVectorIndex.buildGraphFromScratchExclusively
```

with every worker of the pool parked in `ForkJoinPool.awaitWork` and nothing executing the insertion.

## Mechanism

The insertion is submitted to the pool and joined from an external (non-ForkJoinPool) thread:

```java
buildPool.submit(() -> IntStream.range(0, totalNodes).parallel().forEach(...)).join();
```

[...]

PR #6054, opened 2026-08-11
fix(#5872): cancel the in-flight ForkJoinTask on close instead of relying on shutdownNow()

lvca:
## Summary

`LSMVectorIndex.releaseBackgroundResources()` shuts down the dedicated graph-build `ForkJoinPool` with `shutdownNow()` to cancel an in-progress build on `db.close()`, but that only cancels tasks it still finds queued and unstarted. Once insertion is under way, the calling thread is parked in an **external** (non-pool) `ForkJoinTask.join()` on the submitted insertion task, waiting on that exact task's own completion status - a status `shutdownNow()` never sets for a task already running, and one `Thread.interrupt()` on the joiner cannot force either, since an external joiner ignores interruption while parked there. The joiner can then outlive `close()` indefinitely, even once every pool worker has gone idle. The original report reproduced a build thread still alive 60 seconds after `db.close()` returned.

- Keep a handle to the submitted insertion `ForkJoinTask` and cancel it directly in `releaseBackgroundResources()` before shutting the pool down. `ForkJoinTask.cancel()` forces the task's own terminal status and wakes every joiner waiting on it, which `shutdownNow()` cannot do for a task already under way.
- The resulting `CancellationException` is now recognized as the expected outcome of a close racing a build: logged once at INFO, instead of surfacing as a SEVERE `"Error building graph from scratch"` wrapped in an opaque `IndexException`.
- Added `docs/release-26.9.1.md` entry.

## Test plan

- [x] New regression test `LSMVectorIndexCloseCancelsInFlightBuildTest` (same build shape as issue #5577's repro: 4000 vectors, 128 dimensions, `maxConnections=64`, `beamWidth=400`) starts a real graph build on a background thread, waits for genuine mid-insertion progress via the `GraphBuildCallback`, then calls `releaseBackgroundResources()` and asserts the build thread terminates promptly instead of hanging.
- [x] Verified the test fails reliably (5/5 runs) against the unfixed code, and passes reliably (5/5 runs) with the fix.
- [x] `mvn -q -pl engine -am compile` clean.
- [x] Full `com.arcadedb.index.vector.*` test package passes with the fix applied.

https://claude.ai/code/session_01X1y2csftaVgtrNwgSNETzL

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -632,3 +632,24 @@ exclusion guarded against. Not a known exploitable bypass in the current built-i
 for any embedder configuring a broader one via `HostClassLookupFilter`'s public constructor.
 
 [#6045](https://github.com/ArcadeData/arcadedb/issues/6045)
+
+## Closing a database no longer risks leaving an in-flight vector graph build's thread alive indefinitely (#5872)
+
+`LSMVectorIndex` builds its HNSW graph on a dedicated `ForkJoinPool` for exactly one reason: `shutdownNow()` on
+close is supposed to cancel a build still in progress, since JVector's `GraphIndexBuilder` does not observe
+`Thread.interrupt()` on the thread that called it, only on the pool's own workers. But the insertion itself is
+`buildPool.submit(() -> IntStream.range(...).parallel().forEach(...)).join()`, joined from the calling thread -
+external to the pool. `shutdownNow()` only cancels tasks it still finds queued and unstarted; once insertion is
+under way, the joiner is parked on that specific task's own completion status, which an external (non-pool)
+`ForkJoinTask.join()` waits for without ever checking `Thread.interrupt()`. The original report reproduced a
+build thread still alive and parked in `ForkJoinTask.join()` a full 60 seconds after `db.close()` returned, with
+every pool worker already idle - `shutdownNow()` had nothing left to cancel, and nothing was ever going to mark
+that specific task done.
+
+The fix keeps a handle to the submitted insertion task and, on `releaseBackgroundResources()`, cancels it
+directly before shutting the pool down. `ForkJoinTask.cancel()` forces the task's own completion status and wakes
+every joiner waiting on it, which `shutdownNow()` cannot do for a task already under way. The resulting
+`CancellationException` is now recognized as the expected outcome of a close racing a build, logged once at INFO
+rather than surfacing as a SEVERE "Error building graph from scratch" wrapped in an opaque `IndexException`.
+
+[#5872](https://github.com/ArcadeData/arcadedb/issues/5872)
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -47,10 +47,14 @@
         -->
         <graalvm.version>25.2.4</graalvm.version>
         <!--
-            Bumping this requires re-checking two things LSMVectorIndex relies on that jvector does not
+            Bumping this requires re-checking several things LSMVectorIndex relies on that jvector does not
             contractually guarantee across versions:
             - GraphIndexBuilder.build() is unrolled by hand in buildGraphFromScratchExclusively() (see the
               MAINTENANCE comment at that unroll) to make the insertion/cleanup phase boundary observable.
+            - GraphIndexBuilder.cleanup() is wrapped in its own tracked submit()/join() the same way, because
+              cleanup() does the identical external-join shape internally (see the MAINTENANCE comment at that
+              wrapper) - re-check its two internal parallelExecutor.submit(...).join() calls are still there
+              and in the same order.
             - buildGraphFromScratchExclusively()'s CancellationException handler (issue #5872) skips
               GraphIndexBuilder.close() on the fallback path where releaseBackgroundResources() forces an
               in-flight build's join() to unblock while a pool worker may still be running. That is safe only
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -46,6 +46,18 @@
             native.graalvm.version comment for the exact failure mode).
         -->
         <graalvm.version>25.2.4</graalvm.version>
+        <!--
+            Bumping this requires re-checking two things LSMVectorIndex relies on that jvector does not
+            contractually guarantee across versions:
+            - GraphIndexBuilder.build() is unrolled by hand in buildGraphFromScratchExclusively() (see the
+              MAINTENANCE comment at that unroll) to make the insertion/cleanup phase boundary observable.
+            - buildGraphFromScratchExclusively()'s CancellationException handler (issue #5872) skips
+              GraphIndexBuilder.close() on the fallback path where releaseBackgroundResources() forces an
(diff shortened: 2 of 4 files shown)


lvca:
Thanks for the review - all three points were valid, addressed in 80e1559:

1. **Close()-vs-worker race**: reordered `releaseBackgroundResources()` so `shutdownNow()` + `awaitTermination(5s)` run first; `insertionTask.cancel()` now only fires as a fallback when that times out (the pathological case the issue actually reported), not on every cancelled build. Added `releaseBackgroundResourcesCancelsAStuckInsertionTaskAfterTheShutdownTimeout`, which injects a genuinely stuck worker via reflection to exercise that fallback deterministically (confirmed it takes ~5.5s, i.e. it genuinely hits the timeout path).
2. **`startAsyncGraphRebuild()` interrupt-flag race**: now checks `e instanceof CancellationException` directly instead of relying solely on `isInterrupted()`, which could lag the actual cancel.
3. **`rebuildGraphIfNeeded()`**: confirmed it has zero callers anywhere in the codebase, so removed it rather than patching a catch block that can never run.

Also dropped the `isValid()` ternary in the cancellation log message (correct catch: `valid` is always `false` by the time that catch can fire) and switched `cancel(true)` to `cancel(false)` per the ForkJoinTask javadoc note - verified against the actual JDK 21 this project builds with that `ForkJoinPool.submit(Runnable)`'s `AdaptedRunnableAction` does not override `cancel()`, so the flag is genuinely a no-op here (JDK 25 changed this via `InterruptibleTask`, but that's not what's in play on this toolchain).

Full `com.arcadedb.index.vector.*` package still green, both new regression tests pass reliably across repeated runs.

lvca:
Thanks, this was the sharpest point of the two rounds - you're right that the reordering alone only reduced the *frequency* of the race, not its possibility, and exactly when the fallback fires is exactly when a worker is most likely still mid-flight.

Fixed in d665791 by dropping the try-with-resources for `builder` in favor of explicit `close()` calls on every exit path except the `CancellationException` one, which now skips `close()` entirely instead of racing it.

On "confirming with JVector's actual thread-safety contract" - I did: `GraphIndexBuilder.close()` → `searchers.close()` → `ExplicitThreadLocal.close()` (documented "Not threadsafe", confirming the race is real in principle), but every `GraphIndexBuilder` this engine constructs backs onto `OnHeapGraphIndex` (checked against jvector 4.0.0-rc.9 sources), whose `View.close()` is a literal no-op ("No resources to close" - `OnHeapGraphIndex.java` lines 295-297 and 513-515). So concretely, today, skipping `close()` on this path leaks nothing beyond a `ConcurrentHashMap` entry per thread - no file handle, no native/off-heap memory. I made the change regardless of that finding: it removes a dependency on jvector preserving that implementation detail across versions, and costs nothing (a handful of heap objects GC'd once unreferenced) versus the alternative of trusting a "not threadsafe"-documented method to behave under concurrent access it explicitly disclaims supporting.

Given that finding, I didn't add a third test that forces a real stuck `GraphIndexBuilder` worker through the fallback - there'd be nothing hazardous left to actually observe (no crash/corruption is possible against a no-op `close()`), and the existing `releaseBackgroundResourcesCancelsAStuckInsertionTaskAfterTheShutdownTimeout` test already proves the fallback's unblocking mechanism deterministically via the synthetic pool/task. Happy to add one if you'd still want direct coverage of "close() is skipped, not just harmless" as a matter of protecting the invariant itself (e.g. via a counter/spy) rather than protecting against jvector's current behavior - let me know.

On the `rebuildGraphIfNeeded()` removal being scope creep: fair concern in general, but CLAUDE.md for this repo is explicit that confirmed-dead code should be deleted rather than patched ("if you are certain that something is unused, you can delete it completely"), and you'd independently confirmed it has no call sites - so I kept the removal rather than leaving an unreachable, now-inconsistent catch block behind.

Full `com.arcadedb.index.vector.*` package green, both regression tests still pass reliably (verified across repeated runs, including the ~5.5s stuck-worker one, which continues to genuinely exercise the timeout fallback).

lvca:
Thanks for the thorough trace-through, especially independently verifying the `ForkJoinTask.cancel()`/`setCompletion()`/`notifyAll()` mechanism against the actual JDK implementation and the "does a straggler worker touch any live index state" question - that's exactly the analysis I wanted a second pair of eyes on.

Two things to close out:

1. **jvector version-bump risk**: added a comment next to `jvector.version` in `engine/pom.xml` (4bc84c2) tying a future bump to re-verifying that `View.close()` stays a no-op for the on-heap graph path, alongside the pre-existing `GraphIndexBuilder.build()` unroll note it sits next to. That's the right place for it since it's exactly where a version bump gets made.

2. **`db.drop()` racing a straggler `buildPool` worker in test 1's `finally`**: agreed this is harmless, for the same reason your own correctness analysis above establishes for the analogous `releaseBackgroundResources()`-vs-straggler case - the straggler only touches build-local objects (`builder`, the page-snapshotted `vectors`, `scoreProvider`), never live `LSMVectorIndex`/database state, so a concurrent `drop()` has nothing to race. Framed as non-blocking and speculative in your own review, so no code change there; flagging for the record in case this test is ever revisited.

`com.arcadedb.index.vector.*` package still green after the pom.xml addition (doc-only change, no compilation/behavior impact).

At this point all three review rounds' points have been addressed or explained, and round 3 found no correctness defect. Let me know if there's anything else before merge.

lvca:
Thanks - point 1 was a real gap, good catch.

1. **`cleanup()` isn't covered**: confirmed against jvector 4.0.0-rc.9 source - `GraphIndexBuilder.cleanup()` does two of its own `parallelExecutor.submit(...).join()` calls internally, the identical external-join shape insertion has. But `graphBuildInsertionTask` was nulled right after insertion's own `join()` returned, *before* `cleanup()` ever runs - a close() racing the cleanup phase specifically would have hit the exact hang this issue reports, completely unprotected. Fixed in d923f0f by renaming the field to `graphBuildActiveTask` (it now spans both phases) and wrapping `builder.cleanup()` in the same submit-then-track-then-join pattern insertion already uses. Also extended the `jvector.version` pom.xml comment to cover this shape for future version bumps.

2. **No operator-visible warning**: added one, mirroring the existing `rebuildThread` WARNING a few lines below - fires whenever `awaitTermination(5s)` times out, regardless of whether a task handle was available to cancel (a `WARNING` here is useful either way).

3. **Raw `CancellationException` reaching other callers** (`CREATE INDEX`, `COMPACT INDEX`, mutation-threshold rebuild): agreed with your own framing that this is minor and likely acceptable - left as-is. Those callers already catch generic `Exception` and log it, so nothing crashes; and a distinct, recognizable exception type is arguably more honest here than an `IndexException` that would suggest a real build failure rather than "the database went away while this was running." Happy to reconsider if you'd rather it be wrapped consistently.

`com.arcadedb.index.vector.*` package still green, both regression tests still pass reliably (verified across repeated runs) with `cleanup()` now going through the same tracked-task path as insertion.

PR #6058, opened 2026-08-11
feat(cypher): add 7 missing APOC-compatible functions

### LINKED ISSUE #6059: Add APOC-compatible Cypher functions: coll.sum, coll.avg, coll.union, coll.unionAll, math.round, convert.toString, number.format
## Feature

Add 7 APOC-compatible Cypher functions, reachable both under their bare namespaced name and the `apoc.*` prefix:

- `coll.sum(list)` - sum of a numeric list
- `coll.avg(list)` - average of a numeric list, `null` on empty
- `coll.union(list1, list2)` - distinct union, preserving order of first occurrence
- `coll.unionAll(list1, list2)` - concatenation, preserving duplicates
- `math.round(value, precision, mode)` - namespaced entry point for the existing `round()` semantics
- `convert.toString(value)` - namespaced entry point for the existing `toString()` semantics
- `number.format(number, pattern)` - formats a number to a string via a `DecimalFormat` pattern

## Motivation

A customer migrating from Neo4j reported these as the highest-usage APOC functions missing from their query catalogue - of 11 APOC functions their queries call, only `apoc.date.format` previously existed in ArcadeDB. These 7 have direct, low-risk implementations. Out of scope here: `apoc.do.when`, `apoc.refactor.mergeNodes`, `apoc.refactor.cloneNodesWithRelationships` - graph-mutation procedures with real APOC config-flag semantics, tracked separately.

## PR

#6058

robfrank:
## What does this PR do?

Adds 7 APOC-compatible Cypher functions, all reachable both under their bare name and the `apoc.*` prefix (the registry already strips that prefix automatically):

- `coll.sum(list)` - sum of a numeric list
- `coll.avg(list)` - average of a numeric list, `null` on empty
- `coll.union(list1, list2)` - distinct union, preserving order of first occurrence
- `coll.unionAll(list1, list2)` - concatenation, preserving duplicates (the difference from `union`)
- `math.round(value, precision, mode)` - namespaced entry point delegating to the existing `RoundFunction`
- `convert.toString(value)` - namespaced entry point delegating to the existing `ToStringFunction`
- `number.format(number, pattern)` - new, uses `DecimalFormat` pinned to `Locale.ROOT` (default pattern `#,##0.###`)

## Motivation

A customer migrating from Neo4j reported these as the highest-usage APOC functions missing from their query catalogue (out of 11 APOC functions used, only `apoc.date.format` previously existed). These 7 have direct, low-risk implementations; the remaining 3 (`apoc.do.when`, `apoc.refactor.mergeNodes`, `apoc.refactor.cloneNodesWithRelationships`) are graph-mutation procedures with real APOC config-flag semantics and are scoped separately.

## Related issues

Closes #6059

## Additional Notes

- `math.round` and `convert.toString` reuse the existing bare-name `RoundFunction`/`ToStringFunction` implementations by delegation rather than duplicating logic, since the APOC-namespaced name (`math.round`/`convert.toString`) doesn't resolve through the same registry entry as the bare Cypher-standard name (`round`/`toString`).
- `number.format` explicitly uses `DecimalFormatSymbols.getInstance(Locale.ROOT)` rather than the JVM default locale - this codebase has a dedicated `LocaleSensitivityTest` regression suite for exactly this class of bug (e.g. Turkish locale breaking casing), so any new locale-sensitive formatting follows that same convention.
- Extended `apocPrefixCompatibilityForFunctions()` in `CypherBuiltInFunctionsTest` to cover all 7 new functions.

## Checklist

- [x] I have run the build using `mvn clean package` command (ran `mvn -pl engine -am compile` + targeted `test` runs; full `mvn clean package` not run locally due to time, but engine module builds clean)
- [x] My unit tests cover both failure and success scenarios (empty-list cases for sum/avg, dedup vs non-dedup for union/unionAll, precision+mode variants for round, apoc-prefix round trips for all 7)

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/CypherFunctionRegistry.java
+++ b/engine/src/main/java/com/arcadedb/function/CypherFunctionRegistry.java
@@ -30,6 +30,7 @@ import com.arcadedb.function.date.*;
 import com.arcadedb.function.map.*;
 import com.arcadedb.function.math.*;
 import com.arcadedb.function.node.*;
+import com.arcadedb.function.number.*;
 import com.arcadedb.function.path.PathCombine;
 import com.arcadedb.function.path.PathCreate;
 import com.arcadedb.function.path.PathElements;
@@ -232,6 +233,8 @@ public final class CypherFunctionRegistry {
     registerMathFunctions();
     // Convert functions
     registerConvertFunctions();
+    // Number functions
+    registerNumberFunctions();
     // Date functions
     registerDateFunctions();
     // Util functions
@@ -249,6 +252,7 @@ public final class CypherFunctionRegistry {
   }
 
   private static void registerCollFunctions() {
+    register(new CollAvg());
     register(new CollDistinct());
     register(new CollFlatten());
     register(new CollIndexOf());
@@ -257,6 +261,9 @@ public final class CypherFunctionRegistry {
     register(new CollMin());
     register(new CollRemove());
     register(new CollSort());
+    register(new CollSum());
+    register(new CollUnion());
+    register(new CollUnionAll());
   }
 
   private static void registerTextFunctions() {
@@ -314,6 +321,7 @@ public final class CypherFunctionRegistry {
     register(new MathMaxLong());
     register(new MathMinLong());
     register(new MathMaxDouble());
+    register(new MathRound());
   }
 
   private static void registerConvertFunctions() {
@@ -326,6 +334,11 @@ public final class CypherFunctionRegistry {
     register(new ConvertToBoolean());
     register(new ConvertToInteger());
     register(new ConvertToFloat());
+    register(new ConvertToString());
+  }
+
+  private static void registerNumberFunctions() {
+    register(new NumberFormat());
   }
 
   private static void registerDateFunctions() {
--- a/engine/src/main/java/com/arcadedb/function/coll/CollAvg.java
+++ b/engine/src/main/java/com/arcadedb/function/coll/CollAvg.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.function.coll;
+
+import com.arcadedb.query.sql.executor.CommandContext;
+
+import java.util.List;
+
+/**
+ * coll.avg(list) - Returns the average of a list of numbers, or null if the list is empty.
+ */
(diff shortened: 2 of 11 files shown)


robfrank:
Addressed in f8504fb3d:

**Arity validation (fixed).** Verified empirically before fixing - `RETURN coll.sum()` etc. did throw raw `ArrayIndexOutOfBoundsException` rather than `CommandSemanticException`. One correction on the mechanism though: `CypherSemanticValidator.java:2088`'s `!name.contains(".")` check only gates the *unknown-function* error, not arity - the actual reason parse-time arity validation doesn't catch this is that `FunctionValidator`'s `FUNCTIONS` map never registered `coll.*`/`math.*`/`convert.*`/`number.*` in the first place (it only has a handful of curated dotted names like `vector.dimension.count`, `char.length`). Same root effect either way (namespaced functions must self-validate via `checkArity()`), so the fix is what you suggested - added `checkArity(args)` as the first line in all 5, plus a regression test per function for the under-arity case (`CypherMissingFunctionsTest`/`CypherBuiltInFunctionsTest`).

**Non-numeric elements in coll.sum/coll.avg (fixed).** Agreed this should be loud, not silent - matches the `AbstractMathFunction.asDouble`/`CypherFunctionHelper.requireNumberArgument` convention elsewhere in this codebase. Both now route each element through `CypherFunctionHelper.requireNumberArgument`, which throws `CommandSemanticException` for a non-null non-Number element and propagates `null` (so a `null` element in the list is skipped, not rejected - treating `null` as legitimately absent rather than a type error). Added a regression test for the mixed-type case on both functions.

**NumberFormat reusing CypherFunctionHelper.requireNumberArgument (fixed).** Verified it's a drop-in - same null-propagates/throw-on-non-Number contract as the bespoke check it replaced.

**coll.sum/coll.avg returning Double (not changing).** Checked against Neo4j APOC's actual signature: `apoc.coll.sum(coll :: LIST<INTEGER | FLOAT>) :: FLOAT` - it's declared to always return FLOAT, so `Double` here is matching upstream semantics, not an implementation accident. Leaving as-is.

**coll.union/unionAll Integer-vs-Double dedup nit (addressed via test).** Added `collUnionDedupsByTypeAndValue()` documenting that `coll.union([1], [1.0])` keeps both - it's the same `LinkedHashSet`/object-equality dedup `coll.distinct` already used before this PR, so changing the equality semantics here would be inconsistent with that precedent rather than a bug in this PR.

297 tests pass (full function/registry regression sweep), including the new arity and type-rejection cases.

robfrank:
Addressed all 4 in 7bca18f97:

**1. number.format explicit-null pattern (fixed).** Verified `isExplicitNull` is exactly the mechanism `round()`'s mode argument already uses for this distinction. `number.format(1234.5, null)` now returns `null`; `number.format(1234.5)` (omitted) still uses the default pattern. Added `numberFormatExplicitNullPatternPropagates`.

**2. Delegate arity error naming (fixed).** Confirmed empirically before fixing - `RETURN math.round(1,2,3,4)` did throw with message `Function 'round' expects 1-3 arguments but got 4`, naming the delegate instead of the call actually made. Both `MathRound` and `ConvertToString` now call their own `checkArity(args)` before delegating. Added a test per function asserting the message names `math.round`/`convert.toString`.

**3. coll.* apoc-prefix round-trip coverage (added).** Added `collSumApocPrefix`/`collAvgApocPrefix`/`collUnionApocPrefix`/`collUnionAllApocPrefix`, each running an actual `RETURN apoc.coll.x(...)` query through the engine rather than just the registry-identity check, matching what `math.round`/`convert.toString`/`number.format` already had.

**4. Null-element handling in coll.sum/coll.avg (documented via test, behavior unchanged).** Kept the current behavior (null skipped, not rejected) since it's arguably more useful than a hard type error and doesn't contradict any documented convention the way silently swallowing a wrong-typed element did. Added `collSumNullElementIsSkipped`/`collAvgNullElementIsSkipped` making it explicit and regression-tested rather than leaving it as an inferable side effect.

306 tests pass (full function/registry regression sweep).

Issue #4076, opened 2026-05-04
The cluster didn't replicate data on massive insert with gRPC.


mdre:
I did a massive insert with gRPC and the replicas didn't insert the data. This are screenshots of nodo1 (leader) and one of the followers.

<img width="853" height="972" alt="Image" src="https://github.com/user-attachments/assets/d9305438-1502-4cfe-9a01-10105ccf737c" />

<img width="853" height="972" alt="Image" src="https://github.com/user-attachments/assets/b97640be-0ef5-4a1f-a568-8a7156121329" />

lvca:
Found the issue, the gRPC batch bypassed the replication later.

mdre:
please ping me  when this changes are pushed..

lvca:
all pushed to 26.5.1-SNAPSHOT (or main branch)

lvca:
If you want to test this fix, please check out: https://docs.arcadedb.com/arcadedb/appendix/latest-build#latest-build-local

mdre:
Great! I will pull it and tomorrow I will test again. There's some
instability when you drop down a node and wake it up again.
Example:
N1,2,3 up
N4 (only replica): up

Drop N1. N2 becomes the leader and warns for N1.

Up N1 and become a leader. All ok.

Drop N4. N1 start to log an error.

Up N4. N1 still logs errors.

But I will test further.

Thanks.

Marcelo

danieljuhl:
@lvca is this also related to data loading using `/batch` endpoint?

We are experiencing the same (data on LEADER, but none on FOLLOWER(s)).

lvca:
@danieljuhl the `/batch` was bypassing the cluster, now it's fixed. Please retry.

Issue #4069, opened 2026-05-03
Issue with LIST of 'STRING' in GraphBatch


danieljuhl:
```
ArcadeDB Server v26.4.2 (build 96a306bed54479737b22243d9fabbae68bae01cf/1776998135905/UNKNOWN) is starting up...
Running on Linux 6.8.0-90-generic - OpenJDK 64-Bit Server VM 21.0.10
```

We are loading a lot of data using GraphBatch, and the `/batch` endpoint responds with `500` and the following message:
```
{"error":"Internal error","detail":"The property 'EntityA.names' has been declared as LIST of 'STRING' but a value of type 'null' is used. Value: [[\\"ANONYMOUS WÅRD V/FOO BAR\\"]]","exception":"com.arcadedb.exception.ValidationException"}
```

The specific source item looks like:
```
{"@type":"vertex","@class":"EntityA","@id":"123456789","names":["ANONYMOUS WÅRD V/FOO BAR"]}
```

The input does look like a `LIST of 'STRING'`, but definitely not `null`, and I'm confused by the _nested array_, in the error message, as it doesn't match the input data.

This make me wonder if some internal parsing changes the data.

The property in the schema, has been created using the following statement:
```
CREATE PROPERTY EntityA.names IF NOT EXISTS LIST OF STRING;
```

lvca:
Fixed

PR #6068, opened 2026-08-11
fix(#6061): serialize Map/List properties as JSON in RemoteGraphBatch

### LINKED ISSUE #6061: RemoteGraphBatch error on Map fields
Using RemoteGraphBatch throw an exception when a field is a map.

Tested on 26.9.1
```Java
private void testGraphBatchWithMaps() {
        db = getDBTx(); // getDBTx return a RemoteGrpcDatabase.
        RemoteGraphBatch gb = db.batch()
                            .withWAL(false)
                            .withParallelFlush(true)
                            .withFlushEvery(5)
                            .withBatchSize(5)
                            .build();
        db.command("sql", "create vertex type MapTest  if not exists");
        db.command("sql", "create property MapTest.map if not exists MAP");
        
        HashMap<String, String> hm = new HashMap<>();
        hm.put("1", "1");
        hm.put("2", "2");
        hm.put("3", "3");
        hm.put("4", "4");
        hm.put("5", "5");
        
        gb.createVertex("MapTest", "map",hm);
        gb.flush();
        gb.close();
    }
```
and it thows:
```
2026-08-11 12:30:18.417 INFO  [GraphBatch] <ArcadeDB_0> GraphBatch: relaxing durability for the bulk load (useWAL true->false, walFlush NO->NO); the previous settings are restored on close()
2026-08-11 12:30:18.423 INFO  [GraphBatch] <ArcadeDB_0> GraphBatch: relaxing async executor durability for the bulk load (useWAL true->false, walFlush NO->NO); the previous settings are restored on close()/abandon()
2026-08-11 12:30:18.430 INFO  [GraphBatch] <ArcadeDB_0> GraphBatch: restored async executor durability (useWAL false->true, walFlush NO->NO)
2026-08-11 12:30:18.430 INFO  [GraphBatch] <ArcadeDB_0> GraphBatch closed: vertices=0 edges=0 flushes=0 avgFlushMs=0,0
2026-08-11 12:30:18.431 SEVER [PostBatchHandler] <ArcadeDB_0> Error on command execution (PostBatchHandler)
com.arcadedb.exception.ValidationException: The property 'MapTest.map' has been declared as MAP but an incompatible type is used. Value: {1=1, 2=2, 3=3, 4=4, 5=5}
	at com.arcadedb.database.DocumentValidator.throwValidationException(DocumentValidator.java:404)
	at com.arcadedb.database.DocumentValidator.validateEmbeddedValues(DocumentValidator.java:367)
	at com.arcadedb.database.DocumentValidator.validateField(DocumentValidator.java:98)
``` 

Another thing that I see is that seems that GraphBatch is ignoring the `withWal(false)` parameter.

lvca:
## Summary
Fixes #6061.

`RemoteGraphBatch` (`network/src/main/java/com/arcadedb/remote/RemoteGraphBatch.java`) builds the JSONL payload it sends to the server's `/api/v1/batch` endpoint by hand for performance. Its `appendJsonValue()` helper only special-cased `String`/`Number`/`Boolean` and fell back to `value.toString()` for everything else. For a `java.util.Map` property, `toString()` produces Java's `AbstractMap` format (e.g. `{1=1, 2=2, 3=3, 4=4, 5=5}`), which was then emitted on the wire as a quoted JSON *string*, not a JSON *object*.

Server-side, `JsonlBatchRecordStream.unwrap()` already correctly converts a real JSON object/array into `Map`/`List` (fixed for #4069), and `DocumentValidator` correctly requires `instanceof Map` for a MAP-typed property. Since the client sent a plain string instead of a JSON object, validation correctly rejected it:
```
com.arcadedb.exception.ValidationException: The property 'MapTest.map' has been declared as MAP but an incompatible type is used. Value: {1=1, 2=2, 3=3, 4=4, 5=5}
```
The same bug affects `List`/`Collection`-typed properties (LIST columns), even though it wasn't reported.

Root cause and fix are entirely client-side: `appendJsonValue()` now recursively emits proper JSON object/array syntax for `Map` and `Collection` values (nested maps/lists included), reusing the existing zero-allocation `appendJsonString`/`appendJsonValue` building blocks rather than switching to `JSONObject`/`toString()` and paying an extra allocation+parse round trip.

I considered relaxing `DocumentValidator` or fixing server-side deserialization instead, but both are wrong: the server already round-trips real JSON objects/arrays correctly, and loosening the validator would weaken type-safety for every write path, not just this one.

### Secondary note (`withWAL(false)` seemingly ignored)
Not addressed here. `GraphBatch.Builder.build()` intentionally forces WAL on for a replicated (HA/clustered) database regardless of `withWAL(false)` (`useWAL || database.isReplicated()`), logging an INFO line about it, see #4076. If the reporter's server is clustered this is expected/by design; if not, it needs separate investigation with more detail (not reproducible from the report as-is).

## Test plan
- [x] New unit tests in `network/src/test/java/com/arcadedb/remote/RemoteGraphBatchTest.java` covering `appendJsonValue`/`appendProperties` for Map, empty Map, List, and nested Map/List values (no server required).
- [x] New integration test `RemoteGraphBatchIT.mapAndListPropertiesRoundTrip` reproducing the exact issue scenario end-to-end (create MAP/LIST properties, batch-create a vertex through `RemoteGraphBatch`, verify the values round-trip correctly).
- [x] `mvn -pl network,server -am compile test-compile` passes.
- [x] `mvn -pl network test -Dtest=RemoteGraphBatchTest` - 9/9 pass.
- [x] `mvn -pl server test -Dtest=RemoteGraphBatchIT` - 13/13 pass (all pre-existing tests plus the new one).

### DIFF
--- a/engine/src/main/java/com/arcadedb/schema/Type.java
+++ b/engine/src/main/java/com/arcadedb/schema/Type.java
@@ -414,7 +414,7 @@ public enum Type {
         final float[] array = new float[collection.size()];
         int i = 0;
         for (final Object item : collection)
-          array[i++] = ((Number) item).floatValue();
+          array[i++] = requireNonNullNumber(item, "FLOAT", property).floatValue();
         return array;
       } else if (targetClass.equals(float[].class) && value instanceof double[] src) {
         // Fast path: primitive narrowing copy
@@ -432,7 +432,7 @@ public enum Type {
         final double[] array = new double[collection.size()];
         int i = 0;
         for (final Object item : collection)
-          array[i++] = ((Number) item).doubleValue();
+          array[i++] = requireNonNullNumber(item, "DOUBLE", property).doubleValue();
         return array;
       } else if (targetClass.equals(double[].class) && value instanceof float[] src) {
         // Issue #3864 follow-up: HTTP vector params arrive as float[] from JSON parsing.
@@ -812,6 +812,22 @@ public enum Type {
   }
 
   /**
+   * Casts {@code item} to {@link Number}, rejecting a {@code null} element with a clear message
+   * instead of letting it NPE deep inside a {@code float}/{@code double} narrowing conversion
+   * (issue #6061 code review follow-up, sibling of the {@code null} guard added to
+   * {@link #narrowToIntegral}: a {@code List} containing a {@code null} element, e.g. sent for an
+   * {@code ARRAY_OF_FLOATS}/{@code ARRAY_OF_DOUBLES} property, previously NPE'd on {@code
+   * ((Number) item).floatValue()} instead of raising a clean validation error).
+   */
+  private static Number requireNonNullNumber(final Object item, final String targetType, final Property property) {
+    if (item == null)
+      throw new IllegalArgumentException(
+          "A null element cannot be converted to " + targetType //
+              + (property != null ? " for property '" + property.getName() + "'" : ""));
+    return (Number) item;
+  }
+
+  /**
    * Range-checks {@code value} before narrowing it to a smaller integral type ({@code BYTE}, {@code SHORT} or
    * {@code INTEGER}). Narrowing with a plain {@code .intValue()}/{@code .shortValue()}/{@code .byteValue()} wraps
    * two's-complement on overflow instead of rejecting - {@code 3000000000L} silently became {@code -1294967296} -
@@ -825,9 +841,20 @@ public enum Type {
    * {@code NaN} needs its own guard: per the JLS narrowing-conversion rules {@code Double.NaN.longValue()}/
    * {@code Float.NaN.longValue()} return {@code 0}, which is in-range for every target type and would otherwise
    * slip through the range check as a silent {@code 0} (issue #5970).
+   * <p>
+   * {@code value} can be {@code null} here: every {@code Collection -> byte[]/short[]/int[]/long[]} branch in
+   * {@link #convert} passes a raw {@code (Number) item} cast from the source collection, and a {@code null}
+   * element casts cleanly, so without this guard a {@code List} containing {@code null} (e.g. sent for a
+   * BINARY/ARRAY_OF_* property) would NPE on {@code value.longValue()} below instead of raising a clean
+   * validation error (issue #6061 code review follow-up).
    */
   private static Number narrowToIntegral(final Number value, final long min, final long max, final String targetType,
       final Property property) {
+    if (value == null)
+      throw new IllegalArgumentException(
+          "A null element cannot be converted to " + targetType //
+              + (property != null ? " for property '" + property.getName() + "'" : ""));
+
     if (isNaN(value))
       throw new IllegalArgumentException(
           "Value '" + value + "' is NaN and cannot be converted to type " + targetType //
--- a/engine/src/main/java/com/arcadedb/schema/Type.java
+++ b/engine/src/main/java/com/arcadedb/schema/Type.java
@@ -508,6 +508,15 @@ public enum Type {
         for (int i = 0; i < src.length; i++)
           array[i] = (short) narrowToIntegral(src[i], Short.MIN_VALUE, Short.MAX_VALUE, "SHORT", property);
         return array;
+      } else if (targetClass.equals(byte[].class) && value instanceof Collection<?> collection) {
+        // Convert Collection to byte[] for a BINARY property (issue #6061 follow-up): a JSON array
+        // received over the wire (e.g. from RemoteGraphBatch) parses into a List<Number>, which had
+        // no narrowing path back to byte[] and was silently stored as an untyped List instead.
+        final byte[] array = new byte[collection.size()];
+        int i = 0;
+        for (final Object item : collection)
+          array[i++] = narrowToIntegral((Number) item, Byte.MIN_VALUE, Byte.MAX_VALUE, "BYTE", property).byteValue();
+        return array;
       } else if (targetClass.isEnum()) {
         if (value instanceof Number number)
           return ((Class<Enum>) targetClass).getEnumConstants()[number.intValue()];
--- a/engine/src/test/java/com/arcadedb/Issue6061BinaryPropertyListConversionTest.java
+++ b/engine/src/test/java/com/arcadedb/Issue6061BinaryPropertyListConversionTest.java
@@ -22,9 +22,11 @@ import com.arcadedb.database.MutableDocument;
 import com.arcadedb.schema.Type;
 import org.junit.jupiter.api.Test;
 
+import java.util.Arrays;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Code review follow-up on <a href="https://github.com/ArcadeData/arcadedb/issues/6061">issue #6061</a>:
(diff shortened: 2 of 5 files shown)


lvca:
Thanks for the review, addressed the array gap.

**Primitive/object arrays**: fixed in aec45209f. Added an \`o.getClass().isArray()\` branch to \`appendJsonValue\` that walks the array via \`java.lang.reflect.Array\`, mirroring \`JSONObject.put()\`'s existing handling for the same case. Verified this is sufficient end-to-end (no server-side change needed): \`MutableDocument.convertValueToSchemaType()\` already calls \`Type.convert()\`, which has a \`Collection -> float[]/double[]/int[]/...\` branch, so a JSON array parsed server-side into a \`List<Number>\` is correctly narrowed back to the schema-declared primitive array type. Added unit tests for \`float[]\`/\`int[]\`/\`String[]\`/empty array serialization, plus an integration test (\`floatArrayPropertyRoundTrip\`) that creates an \`ARRAY_OF_FLOATS\` property and round-trips a real vector through \`RemoteGraphBatch\`. All 14 tests in \`RemoteGraphBatchIT\` and 13 in \`RemoteGraphBatchTest\` pass.

**Null map keys**: not changing this. \`String.valueOf(null)\` producing the literal key \`"null"\` is standard behavior for JSON serializers with no native null-key representation (e.g. Jackson does the same by default), and it's consistent with how this codebase's own \`JSONObject\`/\`JSONArray\` handle the same situation. A caller passing a `null` key into a MAP property is already doing something unusual since ArcadeDB's MAP property values are keyed by String; silently coercing to `"null"` rather than throwing matches existing project conventions and isn't worth a behavior change or extra validation on this hot path.

lvca:
Addressed the two actionable points, judgment calls on the other two below.

**Boxing on primitive-array serialization**: fixed in 4bcec8d75. Added typed fast paths (`appendJsonFloatArray`/`appendJsonDoubleArray`/`appendJsonIntArray`/`appendJsonLongArray`/`appendJsonShortArray`) that append primitives directly instead of going through `java.lang.reflect.Array.get()`, which boxes every element. These cover all five of ArcadeDB's primitive numeric array property types (`ARRAY_OF_FLOATS`/`_DOUBLES`/`_INTEGERS`/`_LONGS`/`_SHORTS`). Other array kinds (`byte[]`, `Object[]`/`String[]`, boxed arrays) still go through the generic reflective `appendJsonArray` fallback, since they're not the hot embedding path this matters for. Added unit tests pinning the output of each new typed path.

**No test for non-String map keys**: fixed in the same commit. Added `appendJsonValueSerializesIntegerKeyedMapAsJsonObject`, a `Map<Integer, Integer>` test that reproduces the literal key shape from the original issue's error message (`{1=1, 2=2, 3=3, 4=4, 5=5}`).

**No cycle/depth guard on recursion**: not adding this. Agreed with your own assessment, this is client-side self-inflicted misuse (a caller would have to hand `RemoteGraphBatch` a map/collection containing itself), not attacker-controlled input, and no other code in this class (or the `Collection`/`Map` handling before this PR) has such a guard either. Adding one here would be inconsistent scope creep for a failure mode that was never reachable from this issue.

**NaN/Infinity in Number arrays**: no action needed, you already verified server-side round-trip is correct (Gson's lenient `JsonReader` accepts bare `NaN`/`Infinity` tokens). Leaving `appendJsonValue`'s `Number` branch as-is rather than adding a `JSONObject.put(Number)`-style substitution, since that would change behavior (silently turning real embedding values into `null`) for something that already works correctly end-to-end.

lvca:
Addressed all three points in commit e5749d12d.

**Duplicate keys after String.valueOf() stringification**: added a one-line comment on `appendJsonMap()` documenting the assumption (String-keyed maps, the ArcadeDB MAP-property convention) and the last-write-wins fallback for the rare mixed-key-type collision case. Didn't add a runtime uniqueness check, since that would add a cost to every map property write for a genuinely pathological input (a caller deliberately mixing boxed `Integer`/`Long` keys in the same map), and it matches standard JSON duplicate-key semantics rather than inventing new behavior.

**Five near-identical array methods**: no change, you flagged this as informational/reasonable trade-off, not something to act on.

**Test-coverage gaps**: all three added.
- `appendJsonValueSerializesBoxedIntegerArrayAsJsonArray` - `Integer[]` through the generic reflective fallback.
- `appendJsonValueSerializesFloatArrayNestedInMap` - `float[]` nested inside a `Map` value (the vector-embedding + metadata-map-on-one-vertex case).
- `appendJsonValueSerializesNaNAndInfinityInFloatArrayAsBareTokens` - locks in the bare-token NaN/Infinity round trip you already verified works with the server's lenient parser.

20/20 unit tests and 14/14 integration tests pass.

lvca:
Good catch, fixed in commit 9bf22f51f.

**JSONArray not handled**: added a branch that delegates to \`appendJsonCollection(sb, jsonArray.toList())\`. Verified \`JSONArray.toList()\` (JSONArray.java:91-111) already recursively normalizes any nested \`JSONObject\`/\`JSONArray\` into \`Map\`/\`List\`, so the existing recursive dispatch handles arbitrarily nested content correctly without any further special-casing, this covers both cases you found: a top-level \`JSONArray\` property value, and a nested array field reached through \`JSONObject.entrySet()\` (which, as you noted, returns \`JSONArray\` rather than \`List\` for nested arrays, unlike \`toMap()\`). Added \`appendJsonValueSerializesJSONArrayAsJsonArray\` and \`appendJsonValueSerializesJSONObjectWithNestedJSONArrayField\` to pin down both.

**NaN/Infinity coupling to server parser strictness**: agreed this is worth being aware of but not a code change right now, it's an existing, deliberate, and tested behavior; if the server's parsing strictness is ever tightened that's a server-side decision to make with its own tradeoffs, not something to preemptively guard against here.

**Key-collision comment**: thanks for confirming the tradeoff, no further action.

22/22 unit tests and 14/14 integration tests pass.

lvca:
Good catch, this one was a real gap. Addressed across two commits.

**byte[]/BINARY round-trip claim**: verified empirically first (added a probe test) before fixing, confirmed the property came back as an untyped \`ArrayList\` (\`[1, 2, 3, 4]\`), not \`byte[]\`, exactly as you suspected. Fixed in fd8a03bf0:
- Added the missing \`Collection -> byte[]\` narrowing branch to \`Type.convert()\` (\`engine/.../schema/Type.java\`), mirroring the existing \`short[]\` pattern with \`narrowToIntegral\` bounded to \`Byte.MIN_VALUE\`/\`MAX_VALUE\`.
- Added a client-side \`appendJsonByteArray\` typed fast path in \`RemoteGraphBatch\` for consistency with the other primitive numeric arrays.
- Added \`Issue6061BinaryPropertyListConversionTest\` (engine module) pinning the \`Type.convert()\` fix directly, and turned the probe into a real assertion-based integration test, \`binaryPropertyRoundTrip\`.
- All 123 pre-existing \`TypeTest\` cases plus the new tests pass (130/130), confirming no regression to the conversion logic.

**Misleading "zero-allocation per-record" class comment**: fixed in 8e566cc68, narrowed the claim to the paths it still actually describes (scalars + typed primitive-array fast paths) and noted the container paths added by this PR aren't.

**Missing empty-List test**: added in the same commit.

**Five near-identical array methods**: no change, you agreed with the trade-off, just flagging as intentional.

24/24 unit tests, 15/15 integration tests, 130/130 engine Type-conversion tests pass.

lvca:
Thanks, this was worth fixing since the byte[] branch this PR added uses the exact vulnerable pattern. Fixed in a06ce4972.

**Null-element NPE**: added a null check at the top of \`narrowToIntegral()\` (single choke point, so it covers \`byte[]\`/\`short[]\`/\`int[]\`/\`long[]\` together, not just the new \`byte[]\` branch), and a small \`requireNonNullNumber()\` helper for the \`float[]\`/\`double[]\` branches, which don't go through \`narrowToIntegral()\` and had the identical gap. Both now throw a clear \`IllegalArgumentException\` instead of NPEing. Added two regression tests (\`typeConvertListWithNullElementToByteArrayThrowsCleanly\`, \`...ToFloatArrayThrowsCleanly\`) covering both guard paths. 132/132 engine tests pass (including all 123 pre-existing \`TypeTest\` cases, no regression), 15/15 integration tests pass.

**"zero-allocation" doc nit on \`StringBuilder.append(float/double)\`**: no action, you noted yourself it's pre-existing behavior (the scalar path had the same characteristic before this PR), not something this PR changes.

**No test for null element inside a List/Map property value** (as opposed to \`Type.convert()\`'s internal narrowing): the two new tests above cover the underlying mechanism directly; a `RemoteGraphBatch`-level test would just be exercising the same already-tested `Type.convert()` path through another layer, so not adding one to avoid redundant coverage.

lvca:
@claude the automated review run for the latest commit (a06ce4972) completed without posting a comment (likely a transient issue). Could you please review the current state of the PR one more time? All points raised across the previous six review rounds have been addressed: Map/List JSON encoding (the original #6061 bug), array types including ARRAY_OF_FLOATS/vector embeddings, JSONArray/JSONObject interop, BINARY/byte[] round-trip (including the missing Type.convert() branch), and the null-element NPE guard in Type.convert()'s Collection-to-primitive-array narrowing. Let me know if there's anything else to address before merge.

PR #6071, opened 2026-08-11
feat(#6060): add apoc.do.when, apoc.refactor.mergeNodes, apoc.refactor.cloneNodesWithRelationships

### LINKED ISSUE #6060: Add APOC-compatible refactor/control-flow procedures: apoc.do.when, apoc.refactor.mergeNodes, apoc.refactor.cloneNodesWithRelationships
## Feature

Add the 3 remaining APOC-compatible procedures (called via `CALL apoc.x.y(...) YIELD ...`), scoped separately from #6059/#6058 because each is a genuine graph-mutation feature with real config-flag semantics to match, not a one-file addition:

- **`apoc.do.when(condition, ifQuery, elseQuery, params)` YIELD value** - runs `ifQuery` or `elseQuery` (Cypher strings) depending on the boolean condition, with `params` bound into the sub-query, and yields its result as a map. Moderate complexity: needs to dispatch to `database.query(...)`/`database.command(...)` depending on whether the sub-query reads or writes.
- **`apoc.refactor.mergeNodes(nodes, config)` YIELD node** - merges a list of nodes into the first, per real APOC semantics: property-conflict policy (`discard`/`overwrite`/`combine`), rewiring all incoming/outgoing edges from the absorbed nodes onto the survivor, self-relationship handling, then deleting the absorbed nodes.
- **`apoc.refactor.cloneNodesWithRelationships(nodes, config)` YIELD input, output, error** - clones each given node (new vertex, same type/properties) plus every relationship touching it in both directions, producing a mapping of original RID to clone RID. Needs two passes over the edges: cloned-to-cloned relationships must connect the two clones, not the originals.

Implementation follows the `CypherProcedure` interface and the existing `MergeNode`/`MergeRelationship` procedures (`com.arcadedb.query.opencypher.procedures.merge`) as the closest reference pattern for graph mutation via `context.getDatabase()`.

## Motivation

A customer migrating from Neo4j reported these among the APOC functions their query catalogue depends on (see #6059 for the full list and the other 7, already addressed in #6058). These 3 were split out because they need deliberate semantics decisions matching real APOC config-flag behavior, plus dedicated regression tests per node/edge-count and property-conflict scenario, rather than being safe one-file/low-risk additions like the rest.

## Related issues

Related to #6059 (the 7 simpler APOC functions already shipped via #6058).

robfrank:
## What does this PR do?

Adds the 3 remaining APOC-compatible refactor/control-flow procedures, following the existing `merge.node`/`merge.relationship` `CypherProcedure` pattern (`com.arcadedb.query.opencypher.procedures.merge`):

- **`apoc.do.when(condition, ifQuery, elseQuery, params)` YIELD value** - runs `ifQuery` when `condition` is true, `elseQuery` otherwise (both Cypher query strings), with `params` bound as named parameters. Dispatches to `database.query()` or `database.command()` based on `QueryEngine.analyze(query).isIdempotent()`, so both read and write sub-queries work without the caller having to say which. Each result row of the sub-query is yielded as a map under `value`.
- **`apoc.refactor.mergeNodes(nodes, config)` YIELD node** - merges a list of nodes into the first one (the survivor). `config.properties` controls how a property present on both the survivor and an absorbed node is resolved: `"overwrite"` (default, absorbed value wins), `"discard"` (survivor's value kept), or `"combine"` (both kept as a list). Every incoming/outgoing edge of the absorbed nodes is rewired onto the survivor via `MutableEdge.set("@out"/"@in", ...)` (which the engine's `GraphEngine.moveEdge` already handles correctly) - an edge that connected two nodes both being merged naturally becomes a self-relationship on the survivor. The absorbed nodes are then deleted.
- **`apoc.refactor.cloneNodesWithRelationships(nodes, config)` YIELD input, output, error** - clones each given node (new vertex, same type/properties, minus an optional `config.skipProperties` list) together with every relationship touching it in either direction. A relationship whose other endpoint is also being cloned reconnects the two clones rather than the originals; a relationship to a node outside the given list reconnects the clone to that same original node. A per-node clone failure surfaces via the `error` yield field instead of aborting the whole call. The source nodes and their relationships are left untouched.

All three are registered under both their plain and `apoc.`-prefixed names in `CypherProcedureRegistry`.

## Motivation

A customer migrating from Neo4j reported these among the APOC functions their query catalogue depends on (see #6059 for the full list; the other 7, simpler ones already shipped via #6058). These 3 were split out because they need deliberate semantics decisions matching real APOC config-flag behavior (property-conflict policy, self-relationship handling, clone-to-clone rewiring), plus dedicated regression tests per node/edge-count and property-conflict scenario, rather than being safe one-file/low-risk additions.

## Related issues

Closes #6060. Related to #6059 (tracking issue) and #6058 (the 7 simpler APOC functions already shipped).

## Additional Notes

Testing (TDD, tests written first): `DoWhenTest`, `RefactorMergeNodesTest`, `RefactorCloneNodesWithRelationshipsTest` - 19 tests total, all passing. Also re-ran the full `com.arcadedb.query.opencypher.procedures` package plus the registry/arity tests (`CypherProcedureRegistryTest`, `CypherBuiltInFunctionsTest`, `ProcedureRegistryTest`, `ProcedureArityIssue5627Test`) with no regressions.

One general-purpose finding surfaced while testing, worth flagging for reviewers: a write `CypherProcedure` (this includes the pre-existing `merge.node`/`merge.relationship`) is **not** auto-committed by the engine the way built-in write clauses like `SET`/`CREATE`/`DELETE` are - `CallStep.executeProcedure` runs lazily on the first `ResultSet` pull with no `begin()`/`commit()` wrapping, unlike those steps. A caller invoking a write procedure via `CALL` therefore needs its own explicit transaction around the call and its full consumption. Documented in `RefactorMergeNodesTest`'s class Javadoc since it isn't obvious and affects every existing write procedure, not just the new ones.

## Checklist

- [x] I have run the build using `mvn clean package` command (ran `mvn -pl engine -am compile` plus targeted `test` runs; full `mvn verify` not run for time)

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/CypherProcedureRegistry.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/CypherProcedureRegistry.java
@@ -99,11 +99,11 @@ import com.arcadedb.query.opencypher.procedures.meta.MetaSchema;
 import com.arcadedb.query.opencypher.procedures.meta.MetaStats;
 import com.arcadedb.query.opencypher.procedures.path.PathExpand;
 import com.arcadedb.query.opencypher.procedures.path.PathExpandConfig;
-import com.arcadedb.query.opencypher.procedures.refactor.RefactorCloneNodesWithRelationships;
-import com.arcadedb.query.opencypher.procedures.refactor.RefactorMergeNodes;
 import com.arcadedb.query.opencypher.procedures.path.PathSpanningTree;
 import com.arcadedb.query.opencypher.procedures.path.PathSubgraphAll;
 import com.arcadedb.query.opencypher.procedures.path.PathSubgraphNodes;
+import com.arcadedb.query.opencypher.procedures.refactor.RefactorCloneNodesWithRelationships;
+import com.arcadedb.query.opencypher.procedures.refactor.RefactorMergeNodes;
 
 import java.util.Collection;
 import java.util.Collections;
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/CypherProcedureRegistry.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/CypherProcedureRegistry.java
@@ -89,6 +89,7 @@ import com.arcadedb.query.opencypher.procedures.algo.AlgoTopologicalSort;
 import com.arcadedb.query.opencypher.procedures.algo.AlgoTriangleCount;
 import com.arcadedb.query.opencypher.procedures.algo.AlgoVoteRank;
 import com.arcadedb.query.opencypher.procedures.algo.AlgoWCC;
+import com.arcadedb.query.opencypher.procedures.control.DoWhen;
 import com.arcadedb.query.opencypher.procedures.merge.MergeNode;
 import com.arcadedb.query.opencypher.procedures.merge.MergeRelationship;
 import com.arcadedb.query.opencypher.procedures.meta.MetaGraph;
@@ -98,6 +99,8 @@ import com.arcadedb.query.opencypher.procedures.meta.MetaSchema;
 import com.arcadedb.query.opencypher.procedures.meta.MetaStats;
 import com.arcadedb.query.opencypher.procedures.path.PathExpand;
 import com.arcadedb.query.opencypher.procedures.path.PathExpandConfig;
+import com.arcadedb.query.opencypher.procedures.refactor.RefactorCloneNodesWithRelationships;
+import com.arcadedb.query.opencypher.procedures.refactor.RefactorMergeNodes;
 import com.arcadedb.query.opencypher.procedures.path.PathSpanningTree;
 import com.arcadedb.query.opencypher.procedures.path.PathSubgraphAll;
 import com.arcadedb.query.opencypher.procedures.path.PathSubgraphNodes;
@@ -293,6 +296,10 @@ public final class CypherProcedureRegistry {
     registerMetaProcedures();
     // Database index procedures (Neo4j compatibility)
     registerDbProcedures();
+    // Control-flow procedures (APOC compatibility)
+    registerControlProcedures();
+    // Refactor procedures (APOC compatibility)
+    registerRefactorProcedures();
   }
 
   private static void registerMergeProcedures() {
@@ -389,5 +396,14 @@ public final class CypherProcedureRegistry {
     register(new MetaStats());
     register(new MetaNodeTypeProperties());
     register(new MetaRelTypeProperties());
+  }
+
+  private static void registerControlProcedures() {
+    register(new DoWhen());
+  }
+
+  private static void registerRefactorProcedures() {
+    register(new RefactorMergeNodes());
+    register(new RefactorCloneNodesWithRelationships());
   }
 }
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/control/DoWhen.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/control/DoWhen.java
@@ -49,6 +49,19 @@ import java.util.stream.Stream;
  * silently pick a branch rather than surfacing the caller's mistake.
  * </p>
  * <p>
+ * {@code ifQuery} is type-checked unconditionally, even on a call where {@code condition} is false and
+ * {@code ifQuery} never runs - a malformed call fails the same way regardless of which branch is taken,
+ * rather than only failing once someone happens to trigger the true branch. {@code elseQuery} differs
+ * only in that {@code null} is accepted, meaning "no else branch": the call yields zero rows rather than
+ * running anything.
+ * </p>
+ * <p>
+ * {@code ifQuery}/{@code elseQuery} run as Cypher against this database with whatever privileges the
+ * caller already has - same as real APOC's {@code apoc.do.when}/{@code apoc.cypher.doIt}, and no more
+ * of a new risk than any other place in the engine that runs a caller-supplied command string. As with
+ * those, never build either argument by concatenating untrusted input.
+ * </p>
+ * <p>
  * Example:
  * <pre>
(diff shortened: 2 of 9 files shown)


robfrank:
Addressed in d65efe47a:

1. **Duplicate-node crash** - confirmed: `mergeNodes([a, b, b], {})` re-processed the second `b` after it was already deleted. Fixed by deduplicating `nodes` by identity up front (`RefactorMergeNodes.deduplicateByIdentity`) instead of the narrower survivor-only check. `[a, a]` now collapses to one node and hits the existing "at least two nodes" validation.
2. **`DoWhen` resource leak** - confirmed, no try-with-resources on the exception path. Switched to try-with-resources around the sub-query `ResultSet`, matching `MergeNode.findMatchingNode`.
3. **Edge-clone failures uncaught in `cloneNodesWithRelationships`** - confirmed, `cloneEdge` wasn't wrapped. Now catches per-edge and logs a warning instead of losing every already-produced node row; class Javadoc now states the best-effort scope covers both phases.
4. Fixed the import ordering regression.
5. Extracted the identical `extractVertices`/config-map parsing into `RefactorProcedureArgs`, shared by both refactor procedures.
6. Aligned `cloneNodesWithRelationships`'s config validation with `mergeNodes` - a non-map, non-null config now throws in both instead of being silently ignored in one.

Added regression tests for the duplicate-node case, a 3-way merge (exercises multi-node edge rewiring, not just the 2-node path), and a non-map config argument. Full `com.arcadedb.query.opencypher.procedures` package plus the registry/arity tests still green.

robfrank:
Addressed in 00c8d45f3:

- **`DoWhen.extractString`** - confirmed, `ifQuery`/`elseQuery` now require an actual `String`, matching `extractMap`'s rigor instead of silently `.toString()`-ing anything non-null.
- **`extractBoolean` strictness** - documented in the class Javadoc: intentional, diverges from real APOC's coercion so a wrong-typed caller mistake surfaces rather than silently picking a branch.
- **Test coverage gaps** - added `skipPropertiesExcludesGivenPropertyFromTheClone`, plus two regression tests forcing an actual synchronous failure inside the try/catch blocks: `nodeCloneFailureProducesPopulatedErrorFieldWithoutLosingOtherRows` and `edgeCloneFailureIsSkippedWithoutLosingTheNodeYieldRow`.

  Writing those surfaced a real problem with my own commit message and Javadoc from the first round: I'd described the resilience as protecting against "a unique-index conflict on the new edge," and it doesn't. ArcadeDB defers UNIQUE index checking to commit time (`TransactionIndexContext.checkUniqueIndexKeys` runs in `commit1stPhase`), so a duplicate-key violation escapes the procedure's own try/catch entirely and fails the whole transaction - confirmed by writing that exact test first and watching it fail at `database.commit()`, not inside `execute()`. A MANDATORY property constraint, by contrast, is validated synchronously in `LocalDatabase.createRecordNoLock` before the record is ever written, so that's what the tests actually use, and I corrected the Javadoc/inline comments in `RefactorCloneNodesWithRelationships` to say so.

- **Exception-type consistency (`CommandSemanticException` vs `IllegalArgumentException`)** - checked against `MergeNode`/`MergeRelationship`: type/shape validation (is this arg a List/Map/Vertex) is already fully consistent everywhere, old and new code, using `IllegalArgumentException`. The apparent inconsistency is that `MergeNode`'s own "at least one label required" business-rule check *also* uses `IllegalArgumentException` rather than `CommandSemanticException`, while mine uses `CommandSemanticException` for the analogous "at least two nodes" check. That's `MergeNode` predating and not matching `Procedure.checkArity`'s own documented rationale (a caller mistake should be HTTP 400 via `CommandSemanticException`, not an internal-looking 500) - not something this PR should quietly regress to match. Leaving as-is; retrofitting `MergeNode`/`MergeRelationship` is a separate, pre-existing cleanup outside this PR's scope.
- Filed #6073 for the auto-commit gap, since it's real and affects every existing write procedure, not just these three.

Full `com.arcadedb.query.opencypher.procedures` package plus registry/arity/built-in-functions tests: 495 tests, all green.

robfrank:
Addressed in 6de3a2e7b:

- **Duplicate-node orphaned-clone bug** - confirmed by tracing it exactly as described: `cloneNodesWithRelationships([a, a], {})` produces two clones, the second `cloneOf.put(a.getIdentity(), clone2)` overwrites the first mapping, and `processedEdges` dedupes by the *original* edge's identity, so all of `a`'s edges land on `clone2` while `clone1` - still yielded with `error = null` - gets none. Fixed by moving dedup from `RefactorMergeNodes` into `RefactorProcedureArgs.extractVertices` itself, so both refactor procedures get it automatically and can't drift apart on this again. Added `duplicateNodeInTheListProducesOneCloneWithAllItsEdges` reproducing the exact scenario from the review.
- **Test coverage gaps** - added all four: unknown `properties` policy throws, `combine` policy accumulating across a 3-way merge (traced through the code by hand first to confirm `["x","y","z"]` is actually correct, then the test locks it in), `DoWhen`'s non-boolean/non-string/non-map argument paths, and a new `RefactorProcedureArgsTest` unit-testing the shared helper directly (non-list, non-vertex element, non-map, dedup order) rather than only through a full `CALL`.
- **Dead-code `default` branch** - left in place per your suggested alternative, with a comment noting it's an unreachable defensive fallback rather than removing it, since `extractPropertiesPolicy` and `mergeProperties` are two separate methods that could drift apart under a future edit.
- **`ifQuery` validated even on the untaken branch** - documented in the class Javadoc as intentional (fail the same way regardless of which branch runs).
- **Dynamic-Cypher-injection callout** - added to the class Javadoc: same caveat as any other caller-supplied-command surface in the engine, not a new risk class, but worth stating explicitly since it's new to Cypher `CALL`.

Full `com.arcadedb.query.opencypher.procedures` package plus registry/arity/built-in-functions tests: 509 tests, all green.

Issue #4122, opened 2026-05-07
/api/v1/batch endpoint fails with "Error on updating dictionary" on HA follower nodes


babasaikiran:
## Environment
ArcadeDB version: 26.5.1-SNAPSHOT (build 25ed56568ffba5aaae3e35e000370a1bf10ee069)
Cluster: 3-node Raft HA

## Expected behavior
The batch endpoint should handle dictionary updates on follower nodes the same way regular SQL INSERT does (which works fine).

## Actual behavior
we are getting 500 response with below message
```
500 — {"error":"Internal error","detail":"Error on updating dictionary for key 'key_3'","exception":"com.arcadedb.exception.SchemaException"}
```

## Steps to reproduce
The script below creates a fresh database, tests BATCH with different numbers of new keys on a follower node,
and reports pass/fail. No external dependencies beyond Python 3 standard library.

Save as batch_bug_repro.py and run:
```bash
python issue_follower.py --url http://10.95.125.241:2480/api/v1 --user root --password <password>
```
```python
#!/usr/bin/env python3

import argparse
import json
import requests
from requests.auth import HTTPBasicAuth

DB = "repro_batch_bug"


def main():
    p = argparse.ArgumentParser()
    p.add_argument("--url", required=True, help="Follower node URL")
    p.add_argument("--user", default="root")
    p.add_argument("--password", default="root")
    args = p.parse_args()

    s = requests.Session()
    s.auth = HTTPBasicAuth(args.user, args.password)
    url = args.url.rstrip("/")

    # Test with increasing number of properties: 3, 5, 10, 15, 20, 30, 50
    for num_keys in [3, 5, 10, 15, 20, 30, 50]:
        # Fresh DB each time
        s.post(f"{url}/server",
               json={"command": f"drop database {DB}"}, timeout=30)
        s.post(f"{url}/server",
               json={"command": f"create database {DB}"}, timeout=30)
        s.post(f"{url}/command/{DB}",
               json={"language": "sql",
                     "command": "CREATE VERTEX TYPE Person"},
               timeout=30)

        record = {f"key_{i}": f"val_{i}" for i in range(num_keys)}
        payload = json.dumps({
            "@type": "vertex", "@class": "Person", **record
        })

        r = s.post(f"{url}/batch/{DB}",
                   data=payload.encode(),
                   headers={"Content-Type": "application/x-ndjson"},
                   timeout=30)

        icon = "✅" if r.status_code == 200 else "❌"
[...]

PR #6076, opened 2026-08-11
fix(#6070): route batch() over the gRPC GraphBatchLoad RPC on a gRPC connection

### LINKED ISSUE #6070: RemoteGraphBatch on a gRPC connection sends JSONL over HTTP, not the actual gRPC GraphBatchLoad RPC
## Summary

`RemoteGrpcDatabase.batch()` (inherited from `RemoteDatabase`) returns a `RemoteGraphBatch.Builder`. Despite the name and despite being called on a gRPC connection, `RemoteGraphBatch` sends its payload as JSONL over plain HTTP to `POST /api/v1/batch` (`network/src/main/java/com/arcadedb/remote/RemoteGraphBatch.java`), not over the actual gRPC streaming RPC.

A real gRPC batch-load RPC already exists and is fully wired up server-side:
```proto
// grpc/src/main/proto/arcadedb-server.proto
rpc GraphBatchLoad (stream GraphBatchChunk) returns (GraphBatchResult);
```
It's implemented in `grpcw/src/main/java/com/arcadedb/server/grpc/ArcadeDbGrpcService.java` and covered by its own integration tests (`grpcw/src/test/java/com/arcadedb/server/grpc/*`). But there is no client-side code anywhere in the codebase that calls it - `RemoteGraphBatch` never uses it, and nothing else does either.

## Why this matters

A caller using a `RemoteGrpcDatabase` reasonably expects bulk-load operations to go over gRPC (streaming, binary framing, presumably lower overhead for large batches - the whole point of choosing a gRPC connection). Instead, `.batch()` silently falls back to JSONL-over-HTTP regardless of which connection type the caller is using. This is surprising and easy to miss, since nothing in the API surface signals it.

Discovered as a side note while investigating #6061 (fixed in #6068): that issue was entirely about `RemoteGraphBatch`'s JSON encoding, which turned out to be unrelated to gRPC despite the "grpc" branding in the class/connection names.

## Suggested fix

One of:
- Wire `RemoteGraphBatch` (or a gRPC-specific variant) to actually use the existing `GraphBatchLoad` streaming RPC when the underlying connection is a `RemoteGrpcDatabase`, so gRPC users get the transport they asked for.
- If JSONL-over-HTTP is intentionally the only batch-load transport for now, document that clearly on `RemoteDatabase.batch()`/`RemoteGraphBatch`'s Javadoc, and consider whether the unused `GraphBatchLoad` RPC + its server-side implementation should be removed (or explicitly marked as reserved for future use) to avoid the same confusion for the next person reading the proto file.

Either way, the current silent mismatch between what the API name/connection type implies and what actually happens on the wire is worth resolving explicitly rather than leaving implicit.

lvca:
Fixes #6070.

## The problem

`RemoteGrpcDatabase.batch()` inherited `RemoteDatabase`'s implementation and returned a loader that posted its payload as JSONL to `POST /api/v1/batch`, over plain HTTP, whatever protocol the connection spoke. A caller who opened a gRPC connection got the HTTP transport silently, and nothing on the API surface said so.

Meanwhile `rpc GraphBatchLoad (stream GraphBatchChunk) returns (GraphBatchResult)` had been defined in the proto and implemented in `ArcadeDbGrpcService` since the gRPC module landed, with server-side tests, and **no client anywhere in the codebase calling it**.

## The fix

`RemoteGrpcDatabase.batch()` now returns `RemoteGrpcGraphBatch`, which carries the load over that RPC. The public API is unchanged - same methods, same builder options, same results - so the switch is transparent to anything that only uses `batch()`.

What changes underneath is where temporary ids are resolved. Over HTTP each flush is an independent request the server does not remember, so the loader asks for the id mapping back and rewrites the references of later edges itself. Over one stream the server holds the mapping for the whole load, so a temporary id crosses a chunk boundary untouched. That removes the per-flush round trip, the client-side mapping arrays, and the ceiling `flushEvery` had on the HTTP path past which the server stops echoing a mapping too large to consume. The new loader adds `withTimeout()` for the deadline of the whole stream, and applies backpressure rather than handing gRPC chunks faster than the transport drains them.

### Server-side hardening

Wiring a real client onto the RPC surfaced the gaps that had gone unnoticed while it had none. Each only shows on a load big enough to be worth a streaming transport, and each is something `PostBatchHandler` had already had to solve:

- **A load aimed at a follower is refused instead of taken.** The bulk path mutates state only the leader can serialize - the schema dictionary above all - so running it on a follower races the local state-machine apply in `Dictionary.getIdByName`, which is the corruption #4122 was about and the reason the HTTP endpoint relays the payload to the leader. `GraphBatchLoad` had no such guard. It cannot relay the way HTTP does, because `HAServerPlugin` exposes the leader's HTTP address only and relaying a bulk load through a follower would double the traffic of the transport chosen to avoid exactly that, so it fails with `FAILED_PRECONDITION` naming the leader - before writing anything, so there is nothing to reconcile.
- **The temporary-id mapping is no longer always echoed back in full.** A load of a few million vertices built a response past the default 4 MB message limit and failed at the very end, with everything already committed. The new `return_id_mapping` option is a tri-state mirroring the HTTP `idMapping` parameter: unset caps it and reports `id_mapping_omitted`/`id_mapping_size` past the cap, `true` demands it whatever the size, `false` never sends it. The streaming loader sets `false`, having no use for a mapping the server resolves itself.
- **A failed load reports what it had already committed.** The batch commits incrementally, so an error is not a rollback, but a gRPC status carries no message body. The counters ride the trailers of the failed call and reach the caller through `getResult()`, readable after catching the failure - re-sending blindly would double everything that did get through.
- **`vertex_batch_size` is configurable** (was hardcoded at 10,000, which a replicated database cannot use as one Raft entry), with `commit_retries` and `commit_retry_delay_ms` alongside it for parity.

## Design notes

- **No new module dependency.** `grpc-client` already depended on both `network` and `grpc`, so the loader lives there. `network` gains no gRPC/protobuf dependency and the wire-protocol isolation rule is untouched.
- **Shared wire constant.** The partial-commit trailer key lives in the `grpc` module (`GraphBatchProtocol`), which owns the protocol definition, so the server writing it and the client reading it cannot drift.
- `RemoteGraphBatch.Builder` now holds its options typed rather than pre-rendered as query-string entries, so a non-HTTP transport can read them back; one validated option API serves both. `RemoteGraphBatch` gains a `protected` constructor for a subclass that allocates none of the HTTP-only buffering state.
- The gRPC builder re-declares the inherited setters to narrow their return type, so `withTimeout()` works anywhere in the chain rather than only first.

## Testing

New:
- `Issue6070GrpcGraphBatchIT` (grpc-client, 11 tests) - the transport assertion that is the regression test for the issue, plus cross-chunk temp-id resolution, RID endpoints, edge properties, builder-option propagation, fluent-chain ordering, and the partial-commit path end to end.
- `Issue6070GraphBatchLoadHardeningIT` (grpcw, 5 tests) - the id-mapping tri-state, `vertexBatchSize`, and the partial-commit trailer.
- `Issue6070GraphBatchLoadLeaderGuardIT` (grpcw, Raft cluster) - a follower refuses; the same load on the leader succeeds and replicates.

Every new behaviour was mutation-checked: reverting each of the four server-side changes individually makes exactly the corresponding test fail. The first `vertexBatchSize` test written passed under mutation (a successful load cannot see the buffer boundary) and was replaced with one that fails part-way through the vertex phase, where the boundary decides how much is durable.

Regression suites, all green: network 448, grpcw 172, grpc-client 132, plus `RemoteGraphBatchIT` + `PostBatchHandlerIT` (41) for the HTTP path the builder refactor touches.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -705,7 +705,16 @@ already had to solve:
   and reach the caller through `getResult()`, which is readable after catching the failure - re-sending a load
   blindly would double everything that did get through.
 - **`vertex_batch_size` is configurable.** It was hardcoded at 10,000 with no option to change it, which a
-  replicated database needs, one buffer being one Raft entry. `commit_retries` and `commit_retry_delay_ms` were
-  added alongside it for parity with the HTTP endpoint's builder.
+  replicated database needs, one buffer being one Raft entry.
+- **The commit-retry knobs are reachable from the Java client.** `PostBatchHandler` had read `commitRetries` and
+  `commitRetryDelayMs` as query parameters since they were introduced, but `RemoteGraphBatch.Builder` never
+  exposed them, so no Java caller on either transport could set them. `withCommitRetries()` and
+  `withCommitRetryDelay()` were added to the shared builder, which closes the gap for HTTP as well. Both are
+  `optional` on the wire, because zero is a setting for each of them - no retries at all, and no back-off before
+  the first one - and not the same as leaving them alone, which a plain proto3 integer cannot express.
+
+The follower refusal runs *after* the caller has been resolved against the database it named, the way every
+other RPC on this service starts. Naming the leader is a fact about the cluster's layout, so a request that
+cannot reach the database it asked for is answered about that database rather than told where the leader lives.
 
 [#6070](https://github.com/ArcadeData/arcadedb/issues/6070)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -665,3 +665,47 @@ every joiner waiting on it, which `shutdownNow()` cannot do for a task already u
 rather than surfacing as a SEVERE "Error building graph from scratch" wrapped in an opaque `IndexException`.
 
 [#5872](https://github.com/ArcadeData/arcadedb/issues/5872)
+
+## `batch()` on a gRPC connection now uses the gRPC streaming RPC instead of JSONL over HTTP (#6070)
+
+`RemoteGrpcDatabase.batch()` inherited `RemoteDatabase`'s implementation and returned a loader that posted its
+payload as JSONL to `POST /api/v1/batch`, over plain HTTP, whatever protocol the connection spoke. A caller who
+opened a gRPC connection got the HTTP transport silently, and nothing on the API surface said so. Meanwhile the
+`GraphBatchLoad` client-streaming RPC had been defined in the proto and implemented in `ArcadeDbGrpcService`
+since the gRPC module landed, with no client anywhere in the codebase calling it.
+
+`RemoteGrpcDatabase.batch()` now returns `RemoteGrpcGraphBatch`, which carries the load over that RPC. The public
+API is unchanged - same methods, same builder options, same results - so the switch is transparent to anything
+that only uses `batch()`. What changes underneath is where temporary ids are resolved. Over HTTP each flush is an
+independent request the server does not remember, so the loader has to ask for the id mapping back and rewrite
+the references of later edges itself; over one stream the server holds the mapping for the whole load, so a
+temporary id crosses a chunk boundary untouched. That removes the per-flush round trip, the client-side mapping
+arrays, and the ceiling `flushEvery` had on the HTTP path past which the server stops echoing a mapping too large
+to consume. The new loader adds `withTimeout()` for the deadline of the whole stream, and applies backpressure
+rather than handing gRPC chunks faster than the transport drains them.
+
+Wiring a real client onto the RPC surfaced the gaps that had gone unnoticed while it had none, each of which
+only shows on a load big enough to be worth a streaming transport, and each of which the HTTP endpoint had
+already had to solve:
+
+- **A load aimed at a follower is now refused instead of taken.** The bulk path mutates state only the leader can
+  serialize - the schema dictionary above all - so running it on a follower races the local state-machine apply
+  in `Dictionary.getIdByName`, which is the corruption #4122 was about and the reason `PostBatchHandler` relays
+  the payload to the leader. `GraphBatchLoad` had no such guard. It cannot relay the way HTTP does, because the
+  HA plugin exposes the leader's HTTP address only and relaying a bulk load through a follower would double the
+  traffic of the transport chosen to avoid exactly that, so it fails with `FAILED_PRECONDITION` naming the
+  leader - before writing anything, so there is nothing to reconcile.
+- **The temporary-id mapping is no longer always echoed back in full.** A load of a few million vertices built a
+  response past the default 4 MB message limit and failed at the very end, with everything already committed. The
+  new `return_id_mapping` option is a tri-state mirroring the HTTP `idMapping` parameter: unset caps it and
+  reports `id_mapping_omitted` / `id_mapping_size` past the cap, `true` demands it whatever the size, `false`
+  never sends it. The streaming loader sets `false`, having no use for a mapping the server resolves itself.
+- **A failed load now reports what it had already committed.** The batch commits incrementally, so an error is
+  not a rollback, but a gRPC status carries no message body. The counters ride the trailers of the failed call
+  and reach the caller through `getResult()`, which is readable after catching the failure - re-sending a load
+  blindly would double everything that did get through.
+- **`vertex_batch_size` is configurable.** It was hardcoded at 10,000 with no option to change it, which a
+  replicated database needs, one buffer being one Raft entry. `commit_retries` and `commit_retry_delay_ms` were
+  added alongside it for parity with the HTTP endpoint's builder.
+
+[#6070](https://github.com/ArcadeData/arcadedb/issues/6070)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -703,7 +703,13 @@ already had to solve:
 - **A failed load now reports what it had already committed.** The batch commits incrementally, so an error is
   not a rollback, but a gRPC status carries no message body. The counters ride the trailers of the failed call
   and reach the caller through `getResult()`, which is readable after catching the failure - re-sending a load
-  blindly would double everything that did get through.
+  blindly would double everything that did get through. On that trailer the edge count is what the batch
+  *flushed*, not what it received: an edge is buffered when its record arrives and reaches the database in a
+  later `commitEvery`-sized flush, with the incoming direction connected at close, so counting received edges
+  would claim ones that never landed and lead a caller to re-send too little and lose them. The counter
+  advances a flush at a time, so a load that dies mid-flush under-reports rather than over-reports - a
+  duplicate on re-send is recoverable, a silently missing edge is not. Vertices need no such distinction,
+  being counted once their own commit returns.
 - **`vertex_batch_size` is configurable.** It was hardcoded at 10,000 with no option to change it, which a
   replicated database needs, one buffer being one Raft entry.
 - **The commit-retry knobs are reachable from the Java client.** `PostBatchHandler` had read `commitRetries` and
--- a/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
+++ b/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
@@ -177,6 +177,10 @@ class GraphBatchLoadStream {
    * Ends the stream and waits for the server's totals.
    */
   GraphBatchResult complete() {
+    // A call the server has already failed is done: half-closing it would raise gRPC's own "call already
+    // closed" on top of the real failure and bury it. Report the failure the load actually died of instead.
+    failIfTerminated();
+
     request.onCompleted();
 
     final boolean finished;
--- a/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
+++ b/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
@@ -61,6 +61,8 @@ class GraphBatchLoadStream {
   private static final long READY_WAIT_TIMEOUT_MS = 5 * 60 * 1000L;
 
   private final long timeoutMs;
+  /** Same as {@link #READY_WAIT_TIMEOUT_MS}, kept per-instance so a test can wait a plausible moment instead of five minutes. */
+  private final long readyWaitTimeoutMs;
 
   private final CountDownLatch                    done             = new CountDownLatch(1);
   private final AtomicReference<GraphBatchResult> resultRef        = new AtomicReference<>();
@@ -75,7 +77,12 @@ class GraphBatchLoadStream {
   private StreamObserver<GraphBatchChunk> request;
 
   GraphBatchLoadStream(final long timeoutMs) {
+    this(timeoutMs, READY_WAIT_TIMEOUT_MS);
+  }
+
+  GraphBatchLoadStream(final long timeoutMs, final long readyWaitTimeoutMs) {
     this.timeoutMs = timeoutMs;
+    this.readyWaitTimeoutMs = readyWaitTimeoutMs;
   }
 
   /**
@@ -139,7 +146,7 @@ class GraphBatchLoadStream {
       // never coming.
       return;
 
-    final long deadline = System.currentTimeMillis() + READY_WAIT_TIMEOUT_MS;
+    final long deadline = System.currentTimeMillis() + readyWaitTimeoutMs;
     synchronized (readyLock) {
       while (!stream.isReady()) {
         failIfTerminated();
@@ -147,7 +154,7 @@ class GraphBatchLoadStream {
         final long remaining = deadline - System.currentTimeMillis();
         if (remaining <= 0)
           throw new TimeoutException(
-              "Graph batch load timed out after " + READY_WAIT_TIMEOUT_MS + "ms waiting for the server to consume "
+              "Graph batch load timed out after " + readyWaitTimeoutMs + "ms waiting for the server to consume "
                   + "the records already sent");
         try {
           readyLock.wait(remaining);
--- a/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
+++ b/grpc-client/src/main/java/com/arcadedb/remote/grpc/GraphBatchLoadStream.java
@@ -1,0 +1,247 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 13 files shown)


lvca:
Thanks - all three substantive points were valid and are addressed in ff07d47. Notes on each, including one where checking the premise changed the conclusion.

### 1. Leader guard before authentication — fixed

Correct, and worse than "ordering inconsistency": the refusal *names the leader*, so a caller who had shown nothing for that database learned the cluster's layout from it. `getDatabase(..., credentials)` now runs first, as it does in every other RPC here.

Worth recording how this got pinned down, because two attempts at a test for it were useless:

- A single-server test with bad per-message credentials proves nothing — the channel-level `GrpcAuthInterceptor` rejects the call before it reaches the method. That is the "moot" case you flagged.
- Forcing the follower branch on a single server (`if (true)`) NPEs on `ha.getLeaderAddress()` before producing the message, so it fails for the wrong reason.

The branch is only reachable on a follower, so the test belongs in the cluster IT: `aFollowerResolvesTheDatabaseBeforeNamingTheLeader` sends a follower a load naming an unreachable database. Mutation-checked against the pre-review ordering, where it fails with exactly the disclosure you described:

```
Expecting actual:
  "FAILED_PRECONDITION: graphBatchLoad: this server is not the cluster leader ...
   Reconnect to the leader at 'localhost:2481' (HTTP address; use its gRPC port) and retry"
not to contain:
  "not the cluster leader"
```

### 2. `commit_retries` / `commit_retry_delay_ms` unreachable — fixed, and the gap was wider than gRPC

You were right that nothing reached them, and checking why turned up more: `PostBatchHandler.configureBuilder` has read **both** query parameters since they were introduced, so the gap was on the HTTP client too — not inherited scope, just never wired on either transport. `withCommitRetries()` / `withCommitRetryDelay()` therefore go on the shared `RemoteGraphBatch.Builder`, which closes it for both.

That also surfaced a bug in my own proto: `GraphBatch` documents `commitRetries = 0` as "fail fast, do not retry", but a plain proto3 `int32` defaults to 0 when unset, so the server's `> 0` guard would have silently ignored the one caller who explicitly asked not to retry. Both fields are now `optional` and read with `hasCommitRetries()`. HTTP never had this problem (a query parameter is present or it is not), so this was gRPC-only asymmetry. Covered in `distinguishesZeroCommitRetriesFromUnset`.

### 3. Failed interior flush — hardened, but the premise did not hold

Fixed as suggested: `flush()` hands the buffer over *before* the send, so a failed send leaves nothing to re-send, and `complete()` no longer half-closes a call the server has already failed.

But I could not make the old code misbehave. Reverting both changes and running the new test — an interior auto-flush failure with 400 records after it, so the failure is certainly observed mid-body — **passes either way**. Current grpc-java reports `isReady() == false` on a terminated call, so `awaitReady()`'s `failIfTerminated()` raises the real failure before `onNext()` is ever reached. That is the "harmless" branch you identified, and it is the one that actually happens.

So the change removes the reliance on a grpc-java internal rather than repairing an observed break, and I have not claimed otherwise: the test's javadoc says so explicitly, and what it asserts is the caller-visible contract that must hold however grpc-java resolves this — the committed buffer survives *exactly once*, and `close()` adds no second, misleading failure (`call already closed` / `already half-closed` / `Stream is already completed`).

One correction to my first attempt at that test, in case it matters to a future reader: it asserted 1 surviving vertex where the answer is 0, because the good vertex shared a *server-side* buffer with the bad one and went down with it. It now uses a full committed buffer before the failing one, so the count discriminates.

### Minor: duplicated `10_000` constants — declining, with reason

There is no shared home, and I do not think creating one is worth it. `GRAPH_BATCH_VERTEX_BUFFER` could plausibly move to `GraphBatch` in `engine`, but `MAX_ID_MAPPING_IN_RESPONSE` is a transport-response concern that does not belong on an engine class, so a shared home would either take one of the two or become a grab-bag. Splitting them is worse than leaving both where they are used, and moving either means touching `PostBatchHandler`, which is well-tested code on the HTTP path this PR otherwise only refactors the builder of. The grpcw constant's javadoc already names its HTTP counterpart, so the duplication is documented rather than silent. Happy to revisit if you disagree.

### Verification

All 21 new integration tests pass under failsafe as CI runs them (8 grpcw + 13 grpc-client) — note these are `*IT`, so they are excluded from surefire and only appear at `verify`. Regression suites green: network 448, grpcw 172, grpc-client 132.

lvca:
Both correctness findings were real and are fixed in 378d01e. The first one was the best catch of either round — thank you for it.

### 1. Edge count overstated on a `close()`-phase failure — fixed

Verified before fixing: `createVerticesWithRetry` returns only after `database.commit()`, so `counts[0]` is durable; `newEdge`'s own javadoc says "The edge is NOT created immediately". The asymmetry is exactly as described, and it undercut the one guarantee the trailer exists to provide.

The trailer now reports `GraphBatch.getTotalEdgesCreated()`, which advances only after a flush has committed.

**One thing worth stating plainly: the fix does not make the number exact, and I am not claiming it does.** The counter advances a whole flush at a time, so a load that dies *mid-flush* now under-reports where it used to over-report. Building the test made that concrete — with two buffered edges and a failure inside the flush, one edge really was written, the trailer says 0, and the old code said 2. I kept the asymmetry deliberately: a duplicate on an edge the caller re-sends is recoverable, an edge it never re-sends because the server claimed it is not.

So the test asserts the invariant rather than a figure:

```java
assertThat(partial.getEdgesCreated())
    .as("the trailer must never report more edges than the graph actually holds")
    .isLessThanOrEqualTo(connectedEdgeCount());
```

Mutation-checked against the old behaviour: `Expecting actual: 2L to be less than or equal to: 1L`.

Two dead ends on the way to that test, since they say something about the engine:

- A failure at the *deferred incoming-connect* phase does not reproduce it. Pointing an edge at a nonexistent **destination** lets the outgoing flush commit and only breaks the connect afterwards — at which point those edges genuinely exist and counting them is correct. The trailer reported 2, and 2 was right.
- Making the **source** nonexistent is what fails the flush itself, which is where buffered-but-unflushed edges become observable.

Also worth knowing: a flush that dies part-way leaves orphan edge *records* behind, so the test counts edges reachable from the vertices rather than `countType`, or it would be asserting against records connected to nothing.

### 2. `id_mapping_omitted` unconditional on the failure trailer — fixed as documentation

You were right that one flag was carrying two meanings. The trailer never carries a mapping — trailers are bounded far below a message — so the flag is now documented as "not in this message", with the success path's cap and the failure path both named, and the "never set when the caller declined it" case spelled out. I chose widening the contract over adding a reason enum: one flag meaning "no mapping here" is simpler than two ways to say it, and the success/failure paths are already distinguishable by which message the caller is looking at.

### Performance — fixed, and it was hiding something worse

Both lists are now sized via `ensureCapacity()` once the buffer is resolved from the first chunk.

Chasing your note that `vertex_batch_size` has no upper bound turned up more than a ramp-up cost: without one, `ensureCapacity()` would let a single caller-supplied value allocate arbitrarily, and even before this change a huge value meant "never flush until the stream ends". Both exhaust the heap on one request. It is now clamped to 1,000,000 — far above any legitimate use, since the reason to touch this option is to make the buffer *smaller* than the default for Raft entry sizing.

### Design / duplication — declining, with reasons

- **Shared leader-detection predicate.** The shared part is `ha != null && !ha.isLeader()`; everything around it differs (HTTP relays the payload, gRPC refuses because it cannot). A helper wrapping a two-term null check, spanning `server` and `grpcw`, would need a home neither module owns and would not have prevented any drift I can construct.
- **`GRAPH_BATCH_MAX_ID_MAPPING` duplication.** Raised in round 1 and declined there. Briefly: there is no shared home. `engine` is the only module both depend on, and a transport-response size limit does not belong on `GraphBatch`; taking only the vertex-buffer constant across would split a pair that is easier to reason about together. Both javadocs name their counterpart, so the duplication is documented rather than silent.
- **Trailer vs. terminal-message pattern.** A terminal `onNext` + `onCompleted` would mean a failed load completes *successfully* at the gRPC level, so callers lose the status code and every generic error path stops working. The trailer keeps the error a real error and still carries the data. `insertStream` reports a summary of a stream that *succeeded*, which is a different question from what survived one that did not.
- **`GraphBatchLoadStream` duplicating the latch/AtomicReference shape.** Fair. It also does two things `ingestStream` does not (backpressure, trailer extraction), and unifying all three means editing two working RPCs for no behavioural gain. Better as its own change if someone adds a fourth.

Thanks also for chasing the `ClientCallImpl` TOCTOU question to ground — that matches what I found by mutation last round, and I have left the defensive code with a comment saying it removes a reliance rather than fixing a break.

### Verification

22 new integration tests pass under failsafe (9 grpcw + 13 grpc-client). Regression: network 448, grpcw 172, grpc-client 132.

lvca:
Both points addressed in 2c56d335b.

### 1. `openGraphBatchLoadStream()` throwing leaves the chunk behind — fixed

Correct, and it was the one path the hand-off did not cover: I had moved the buffer hand-off ahead of `send()` but left it behind the call-open, so an open that threw produced exactly the second-failure-on-top-of-the-real-one shape the hand-off exists to prevent. The hand-off now precedes *both* things that can throw:

```java
if (!firstChunkSent)
  chunk.setDatabase(...).setCredentials(...).setOptions(options);

final GraphBatchChunk pending = chunk.build();
chunk = null;
recordsInChunk = 0;

if (stream == null)
  stream = database.openGraphBatchLoadStream(timeoutMs);
stream.send(pending);
firstChunkSent = true;
```

`firstChunkSent` moved to the end for the same reason: the database and options are only on the wire once a chunk is, so if the send throws, a retry stamps them again rather than sending a chunk the server rejects as nameless. With the buffer cleared, the `close()` that follows finds nothing to send and adds no failure of its own — the body's exception propagates alone.

I did not add a test for it. Making `openGraphBatchLoadStream()` throw needs a `RemoteGrpcDatabase` that fails on demand, and the class takes a live connection in its constructor, so the test would be mostly scaffolding asserting a line ordering that is now self-evident from the method. If you would rather have it pinned anyway, say so and I will introduce the seam.

### 2. No coverage of `GraphBatchLoadStream`'s concurrency — fixed, and the first version was worthless

Good call, and it found something. `GraphBatchLoadStreamTest` drives the class against a hand-written `ClientCallStreamObserver` — no server, 9 tests, ~1s — covering: send-through while ready, a parked sender resuming on drain, a parked sender woken by the call failing, the readiness timeout, interrupt handling (including that the flag survives), `complete()` returning the result, `complete()` on an already-failed call not half-closing it, a stream that ends without a result, and the load deadline cancelling the call.

**The first version of two of those tests proved nothing.** I mutation-checked by removing the `notifyAll` from `onError` and they still passed: a sender parked with a 500 ms readiness budget falls out of the wait on timeout, re-checks, and throws the right error anyway — just later. Asserting "eventually finishes" cannot tell a wake-up from a timeout.

They now run with a readiness budget (10 s) far above what the assertion allows (2 s), so only an actual notification meets it. Re-checked against the same mutation:

```
[ERROR] aParkedSenderIsWokenByTheCallFailing -- Time elapsed: 2.136 s <<< FAILURE!
[ERROR] completeReportsAFailedCallWithoutHalfClosingIt <<< FAILURE!
```

Both now fail on the mutation; both pass on the real code. The budget is per-instance for this (`new GraphBatchLoadStream(timeoutMs, readyWaitTimeoutMs)`), so a test can wait a plausible moment instead of the five minutes production waits — the one test-only seam in the change.

### 3. Leader's gRPC address not discoverable — agreed, and out of scope here

You have the constraint right: `HAServerPlugin` exposes `getLeaderAddress()` (HTTP) and a Bolt routing table, but nothing for gRPC, so the message can only name the HTTP address and tell the caller to map the port. A caller retrying automatically has to know the deployment's convention.

The real fix is a gRPC routing table alongside `getBoltRoutingTable()` — the precedent is already there for exactly this problem. That means touching the `HAServerPlugin` interface and its Raft implementation to publish each peer's advertised gRPC address, which is its own change with its own tests and does not belong in a PR about the batch transport. Happy to file it as a follow-up if you want it tracked.

### Verification

22 integration tests and 9 unit tests pass under failsafe/surefire. Regression: network 448, grpcw 172, grpc-client 141 (up from 132: the new unit tests run in surefire, unlike the `*IT` suites).

lvca:
Thanks — and particularly for tracing `GraphBatch.abandon()` and `callAsyncDuplex` rather than taking the description's word for it. The no-hidden-retry check is the one I would most have wanted verified independently: `GraphBatchChunk`s are not idempotent, and a retry under the plumbing would double-apply a partially-acknowledged chunk in a way no test here would catch.

Both actionable observations are addressed in 040db495c, documentation only.

### 1. No way to request the id mapping back

Documented — though the premise is a little off in a way worth recording, since it makes the situation better than it looks: **`RemoteBatchResult` has no `getIdMappingMap()`, and never has, on either transport.**

```java
public long getVerticesCreated()
public long getEdgesCreated()
public long getElapsedMs()
```

So there is no capability the gRPC loader is withholding relative to the HTTP one. Over HTTP the mapping is fetched to rewrite cross-flush edge references *internally* and never surfaces to the caller; over gRPC the server resolves those references itself, so nothing needs fetching. Setting `return_id_mapping=false` removes work whose only consumer does not exist on this path.

The class javadoc now says so explicitly, and points a caller who genuinely wants RIDs back at the two things that do work: query the vertices, or call the RPC directly with `return_id_mapping` set — the wire protocol supports it, which is exactly why the tri-state exists.

### 2. `READY_WAIT_TIMEOUT_MS` invisible from `withTimeout()`

Fair, and the failure mode you describe is real: a link slow enough to spend five minutes draining one chunk fails a load with hours of budget left, and the caller has no way to see why from the API they set. `withTimeout()`s javadoc now states that it bounds the stream and not any single chunk, that the per-chunk ceiling is fixed at five minutes, and why it is not settable — it is what distinguishes a connection that died without saying so from one that is merely busy, and no legitimate load spends five minutes on one chunk's backpressure.

I kept it internal rather than exposing it. Two knobs whose interaction has to be explained before either can be set correctly is worse than one knob and a documented floor; if a real deployment ever breaches it, that is evidence for making it settable, and the javadoc now gives whoever hits it the vocabulary to ask.

### 3. `ProtoUtils` name collision — declining

Agreed it is a readability trap, but both names are load-bearing: `io.grpc.protobuf.ProtoUtils` is gRPC's own, and `com.arcadedb.remote.grpc.utils.ProtoUtils` predates this PR and is used across `grpc-client`. Renaming ours is churn in files this PR does not otherwise touch, for a collision that never appears within a single file — `GraphBatchProtocol` imports gRPC's, `RemoteGrpcGraphBatch` imports ours, neither sees both.

### On the two follow-ups

You and the previous round both landed on the same shape of suggestion — expose the leader's gRPC address (round 3) and expose the readiness ceiling (this round). I have documented both limits rather than widened the API, on the view that a PR fixing a transport gap should not also grow the HA plugin interface or add a knob with no reported need. Both are worth tracking as issues; say the word and I will file them.

### Verification

Documentation only — no source, test, or proto change. Full suites were green on 2c56d335b: 22 integration tests, 9 unit tests, network 448, grpcw 172, grpc-client 141.

PR #6080, opened 2026-08-11
feat(#6072): faster full backup - configurable compression level and parallel compression

### LINKED ISSUE #6072: Speed up full backup: configurable compression level and parallel compression (Phase 1 of backup optimization)
## Context

ArcadeDB only has a full backup, and on a large database it is both slow and disruptive to concurrent writers. The full analysis, including the lock-free and incremental phases, is in [`docs/optimize-backup.md`](https://github.com/ArcadeData/arcadedb/blob/main/docs/optimize-backup.md).

This issue covers **Phase 1 only**: making the existing full backup substantially faster, with no archive-format change, no engine change and no correctness risk. Phases 2 (lock-free backup) and 3 (incremental backup and PITR) are separate follow-ups and must not be pulled into this one.

## Where the time goes today

`FullBackupFormat.backupDatabase()` (`integration/src/main/java/com/arcadedb/integration/backup/format/FullBackupFormat.java:87`) does:

1. `database.executeInReadLock(...)`, which blocks schema changes and new file creation but not transactions.
2. `pageManager.suspendFlushAndExecute(...)`, which freezes the data files so they can be read raw.
3. Streams every `ComponentFile` into a ZIP entry at `setLevel(9)`, single threaded.

Deflate at level 9 on one thread runs at roughly 20-40 MB/s, so the backup is CPU bound long before it is IO bound. A 100 GB database is an hour or more.

Note that the read lock is not the source of the disruption. The disruption comes from the flush suspension: while it holds, dirty pages accumulate in `PageManagerFlushThread.deferredByDatabase` bounded by `FLUSH_SUSPEND_MAX_DEFERRED_RAM` (`engine/src/main/java/com/arcadedb/GlobalConfiguration.java:444`, default 512 MB), and past that cap committing threads are throttled. Phase 1 does not remove the suspension, but by shortening the backup it shortens the throttling window proportionally.

## Scope

1. **Make the compression level configurable and stop defaulting to 9.**
   - Add a backup setting in `BackupSettings` plus a `GlobalConfiguration` default.
   - `Deflater.BEST_SPEED` typically costs 10-15% of compression ratio for a 3-5x throughput gain. Pick the new default from measurement, not from intuition.
   - The existing `-level`-style CLI/SQL surface should expose it; keep the current behaviour reachable for anyone who wants maximum compression.

2. **Compress files in parallel.**
   - ZIP entries are independent: build them on a worker pool into temp buffers or temp files and write the central directory serially.
   - Follow the rules in the `engine-concurrency` skill (`.claude/skills/engine-concurrency/SKILL.md`). In particular do **not** submit to `ForkJoinPool.commonPool()`; use or add a dedicated pool with documented sizing and saturation policy.
   - Watch peak memory: a naive "buffer every entry in heap" implementation will OOM on large databases. Temp files or a bounded number of in-flight entries.
   - Encryption path (`encryptFile`) wraps the whole stream in a `CipherOutputStream`, so parallel work must happen before that serial write stage. Verify encrypted backups still restore.

3. **Optional IO throttle** (MB/s cap) so a backup cannot saturate production IO, as Cassandra offers. Off by default.

4. **Zstd only if measurement justifies it.** `com.github.luben:zstd-jni` is BSD licensed and therefore license-compatible, but it is a new native dependency and the project rule is not to add one unless strictly necessary. Level-1 Deflate plus parallelism may already close the gap. If it is added, `ATTRIBUTIONS.md` must be updated.

## Out of scope

- Removing the flush suspension (Phase 2).
- Any change to the archive format, manifest or restore semantics (Phase 3).
- Incremental or differential backup, WAL archiving, PITR.

## Correctness requirements

- Existing backups must still restore. Existing restore code must not need to change.
- Round-trip tests for both plain and encrypted backups, and for a database with many small files as well as few large ones.
- Backups taken under concurrent write load must restore to a consistent database. Reuse and extend the existing coverage in `integration/src/test/java/com/arcadedb/integration/backup/` (`FullBackupIT`, `BackupDatabaseTest`) and `server/src/test/java/com/arcadedb/server/backup/`.
- Follow the project TDD convention: state the verification plan, write the tests first, then implement.

## Benchmark before and after (required)

Do not merge on the strength of a plausible-sounding change. Measure on the same hardware and the same datasets, before and after, and put the numbers in the PR description:

- **Wall-clock backup duration**, at several database sizes (for example 1 GB, 10 GB, 100 GB).
- **Archive size and compression ratio**, so a throughput gain traded away for ratio is visible and deliberate. Report the trade explicitly, for example "2.9x faster, 11% larger".
- **CPU utilisation and peak heap** during the backup. The parallel path must not blow up memory.
- **Impact on concurrent writers**: sustained insert/update throughput and commit latency percentiles (p50/p95/p99) with and without a backup running. This is the number that matters most to users and the one most easily forgotten.
- **Deferred-RAM high-water mark** and whether the `FLUSH_SUSPEND_MAX_DEFERRED_RAM` cap was hit, which is the direct signal that writers were throttled and by how much less than before.
- **Restore duration**, to confirm it did not regress.

Also report scaling of the parallel path with thread count (1, 2, 4, 8) so the default pool sizing is chosen from data.

Benchmarks must be tagged so they are excluded from normal CI runs: `@Tag("benchmark")` for pure throughput comparisons, `@Tag("slow")` for functional regression tests that take noticeably long. See `CLAUDE.md` for the tagging rules.

## Acceptance criteria

- Full backup of a large database is measurably several times faster on a multi-core machine, with the trade-off in archive size stated explicitly.
- No change to the archive format; old backups restore, new backups restore with the old restore path.
- Concurrent-writer impact is measured and reported, not assumed.
- Compression level and parallelism are configurable, with defaults chosen from the benchmark results.

lvca:
Fixes #6072 (phase 1 of `docs/optimize-backup.md`). No archive-format change, no engine change to the write path, phases 2 and 3 untouched.

## What was slow

`FullBackupFormat` streamed every `ComponentFile` into a ZIP at `setLevel(9)`, on the calling thread. Deflate at level 9 sustains ~20-40 MB/s on one core, so the backup was CPU bound long before it was I/O bound.

Duration is not the only cost. The backup runs inside `PageManager.suspendFlushAndExecute`, and while flushing is suspended dirty pages pile up until they hit `arcadedb.flushSuspendMaxDeferredRAM` (512 MB), past which committing threads are throttled. **The backup's duration is the writer-throttling window**, so making the backup n times faster shortens the disruption by the same factor.

## What changed

**`arcadedb.backup.compressionLevel`**, default lowered from 9 to 1 on measurement (see below).

**`arcadedb.backup.compressionThreads`**, default `-1` = half the available processors capped at 8; `0` selects the previous single-threaded `java.util.zip.ZipOutputStream` writer, kept as an escape hatch. Splitting by archive entry would not have helped - a database is often a handful of very large files, sometimes one dominant one - so the split is **inside** each entry: the reader cuts every file into 1 MB chunks, each is deflated by its own `Deflater` and terminated with `SYNC_FLUSH`, and the compressed chunks are concatenated back in order. Deflate streams ended that way concatenate into one valid stream; this is the construction pigz uses for parallel gzip. Peak heap is bounded by construction at two chunk buffers per thread.

**`arcadedb.backup.maxMBPerSecond`**, default 0 = unlimited: optional cap on the rate the backup reads the database files, so it cannot saturate the production disk. Applied to the read side deliberately - that is the I/O competing with the live workload.

All three are also settable per backup: CLI (`-compressionLevel`, `-compressionThreads`, `-maxMBPerSecond`), the `Backup` API, and `BACKUP DATABASE ... WITH compressionLevel = 6, compressionThreads = 4`.

**No new dependency.** Zstd was considered and rejected: level-1 deflate plus parallelism closes the gap, and `zstd-jni` is a native dependency. `ZipOutputStream` cannot be handed pre-compressed data, so the parallel path emits the ZIP container itself (`ParallelZipArchiveWriter`, ~250 lines) using the streaming form the spec provides for unknown sizes - general-purpose bit 3, zeroed sizes in the local header, a data descriptor after the data - which is exactly what `ZipOutputStream` emits for a `DEFLATED` entry of unknown size, ZIP64 included.

## Benchmarks

`BackupCompressionBenchmark` (`@Tag("benchmark")`, `**/*Benchmark.java`, so it is in the benchmark lane only). Apple M-series, 10 cores, 1.25 GB database of ~2.1M indexed vertices with half-structured 512-byte payloads.

### Duration, archive size, heap

| configuration | seconds | MB/s | archive | ratio | vs base | peak heap |
|---|---|---|---|---|---|---|
| legacy writer, level 9 (previous default) | 18.88 | 67.9 | 323.46 MB | 25.2% | 1.00x | 12.3 MB |
| legacy writer, level 6 | 10.71 | 119.6 | 321.66 MB | 25.1% | 1.76x | 12.3 MB |
| legacy writer, level 1 | 6.18 | 207.5 | 347.66 MB | 27.1% | 3.06x | 12.3 MB |
| parallel, level 1, 1 thread | 4.82 | 265.6 | 347.59 MB | 27.1% | 3.91x | 12.3 MB |
| parallel, level 1, 2 threads | 2.47 | 518.5 | 347.59 MB | 27.1% | 7.64x | 12.3 MB |
| parallel, level 1, 4 threads | 1.26 | 1016.0 | 347.59 MB | 27.1% | 14.97x | 31.3 MB |
| **parallel, level 1, 8 threads** | **0.68** | **1875.8** | **347.59 MB** | **27.1%** | **27.64x** | **45.3 MB** |
| parallel, level 6, 8 threads | 1.39 | 924.3 | 321.78 MB | 25.1% | 13.62x | 35.4 MB |
| parallel, level 9, 8 threads | 2.96 | 432.2 | 323.56 MB | 25.3% | 6.37x | 43.3 MB |

**The trade, stated explicitly: 27x faster, 7.5% larger.** The level change alone is 3.1x faster for 7.5% larger; the parallelism is free of ratio cost (347.66 MB single-chunk against 347.59 MB chunked - the empty-dictionary-per-chunk penalty is below the noise at 1 MB chunks, and is asserted by a test).

Scaling with thread count is near-linear (3.91x / 7.64x / 14.97x / 27.64x at 1 / 2 / 4 / 8), and total CPU is flat at ~5.2 cpu-seconds regardless, so the extra threads buy wall-clock without buying work. That is what the auto default of half-the-cores-capped-at-8 is chosen against: linear scaling means more threads is always faster, but a backup should not claim every core from the workload it is running alongside.

### Impact on concurrent writers (the number that matters)

Sustained single-writer insert load, 200 records per transaction, measured **only inside the backup window** so a long backup and a short one are compared fairly:

| | window | throughput in window | p50 | p95 | p99 | peak deferred RAM |
|---|---|---|---|---|---|---|
| no backup running | 10.0 s | 159,560 rec/s | 1,167 us | 1,627 us | 3,393 us | 0 |
| legacy writer, level 9 | 48.6 s | 6,809 rec/s (4.3%) | 1,619 us | 2,098 us | 2,302 us | 512 MB (cap hit) |
| parallel, level 9, 8 threads | 7.1 s | 51,867 rec/s (33%) | 1,647 us | 2,297 us | 2,486 us | 512 MB (cap hit) |
| **parallel, level 1, auto** | **2.1 s** | **123,192 rec/s (77%)** | **1,573 us** | **1,952 us** | **2,123 us** | 512 MB (cap hit) |

The deferred-RAM high-water mark reaches the 512 MB `flushSuspendMaxDeferredRAM` cap in every backed-up run, so writers were being throttled outright in all three - **what changed is for how long, and how hard**. In terms of write work displaced by one backup: ~7.4M records under the old default, ~76k under the new one, a ~97x reduction. Commit latency inside the window is barely affected either way; throughput, not latency, is where the suspension is paid for.

### Restore

Unchanged code path. Timed both ends of the trade on the same database: 2.9 s from a parallel level-1 archive. No regression - inflate is level-independent and the larger archive costs nothing measurable.

## Two defects found on the way

**`BACKUP DATABASE ... WITH ...` silently dropped every setting.** The statement matched the setting name against `Expression.toString()`, which renders a string quoted, so no `case` ever matched. The visible consequence: `WITH encryptionKey = '...'` produced an archive **in clear** - it restored without the key. The same statement's `copy()` also dropped the settings map, so a copy taken from the statement cache lost them again. Both fixed; the regression test asserts the archive is genuinely encrypted by requiring a keyless restore to fail. (`ExportDatabaseStatement` already read the raw value correctly; this looks like a miss in the ANTLR parser migration.)

**A backup that failed halfway still reported success.** `suspendFlushAndExecute` runs its callback through `CodeUtils.executeIgnoringExceptions`, which logs and swallows, so an I/O error mid-backup left the archive to be finalized with a valid central directory over a truncated set of entries - a backup that looks valid and is not. The failure is now carried out of the suspension by hand and the partial archive is deleted.

## Verification

- `ParallelZipArchiveWriterTest` (23 tests): every archive is read back **twice**, once with `ZipInputStream` (what restore uses, never touches the central directory) and once with `ZipFile` (what unzip uses, only the central directory), across the thread x level matrix, every chunk-boundary size (0, 1, n-1, n, n+1, 2n, 2n+1, ...), compressible / incompressible / all-zero / empty inputs; byte-for-byte equality between the parallel and the legacy writer's output; reported sizes matched against what `ZipFile` reads back; the throttle; and a **4 GB entry** exercising the ZIP64 data descriptor and extra field, left in the normal lane rather than tagged slow because the slow lane only covers the engine module.
- `BackupCompressionIT` (16 tests): backup/restore round trips compared with `DatabaseComparator` across the thread x level matrix including the legacy writer, encrypted and plain; parallel and legacy archives restoring to identical databases; the SQL settings surface; `GlobalConfiguration` defaults; and a backup taken under concurrent write load restoring to a whole number of committed batches.
- `BackupSettingsTest`: CLI parsing and range rejection.
- Existing suites green: `FullBackupIT`, `BackupDatabaseTest`, `Issue5517BloomFilterBackupIT`, the whole `integration` module, the `server` backup tests, and the engine configuration / `PageManagerFlushSuspend` / `BackupDatabaseStatement` tests.

## Concurrency

Per the `engine-concurrency` skill, `ForkJoinPool.commonPool()` is not used. The pool is a dedicated `ThreadPoolExecutor` whose lifetime is exactly one backup - a backup is a rare, bursty operation, so a permanently resident pool would hold idle threads for the 99.99% of the time no backup runs, and there is no steady-state pool for `PoolMetrics` to report. Saturation is handled by construction rather than by a rejection policy: the reader keeps at most `threads * 2` chunks in flight, which bounds peak heap and back-pressures the reader, so the queue cannot fill and the caller-runs policy installed as a backstop never has to fire.


lvca:
Thanks - all three notes addressed.

**1. Double-fault during an already-failing backup.** Fixed properly rather than papered over. `BackupArchiveWriter` gains `abort()` (release the resources, do **not** terminate the archive), and `writeArchive` no longer uses try-with-resources: only a backup that actually succeeded is `close()`d, anything else aborts. Two consequences beyond the flagged one - a `close()` failure can no longer replace the root cause, and a failed backup never gets a valid central directory written at all, so even if the delete of the partial file cannot happen (permissions, full or read-only filesystem) what is left is structurally invalid rather than a plausible-looking archive missing files. Covered by `abortLeavesNoValidArchive` (both writers, asserts `ZipFile` rejects the result) and `aFailedBackupThrowsAndLeavesNoArchive` (makes a component file unreadable mid-backup, POSIX-guarded with `assumeTrue`).

**2. `IoThrottler` double division.** Now exact integer arithmetic. The overflow that motivated the `double` (`totalBytes * 1e9` exceeds a long past ~9.2 GB, which a backup does reach) is handled by re-baselining every GB, and because the new baseline is the previously computed deadline rather than `now`, no rounding drift accumulates.

**3. Third-party unzip tools.** Fair - the claim was JDK-shaped. Now actually verified, on archives produced by this writer including a 4 GB entry (the streaming-data-descriptor + ZIP64 combination):

```
$ unzip -t zip64.zip            # Info-ZIP 6.00 - inflates and CRC-checks, not just lists
    testing: huge.bin                 OK
    testing: small-b.bin              OK
No errors detected in compressed data of zip64.zip.

$ unzip -l zip64.zip
4294971392  08-11-2026 17:49   huge.bin

$ python3 -c "import zipfile; z=zipfile.ZipFile('zip64.zip'); print(z.testzip())"
None
```

So four independent readers now: `ZipInputStream`, `ZipFile`, Info-ZIP and Python `zipfile`. I did not turn this into a test - shelling out to `unzip` would make the suite depend on what is installed on the runner - but recorded it in `ParallelZipArchiveWriter`'s Javadoc so the evidence lives in the source rather than only in this thread.

Pushed as 8a750019e.

lvca:
Second round addressed - all three points were valid, all three implemented.

**1. `GlobalConfiguration` validation hook.** Agreed, and the docstring on `parseIntSetting` was literally describing the gap it left open. Rather than the `Callable` hook I used the `allowed`-set mechanism already in the enum (`integerRangeAsStrings`, as `SUPERNODE_STRIPES`-style entries do) - worth noting the `Callable` hook could not have enforced this anyway: `setValue` catches a throwing callback and only logs SEVERE, so the value would still be accepted. The `allowed` check throws from inside the outer try, which restores the previous value and rethrows, which is the behaviour we want.

- `BACKUP_COMPRESSION_LEVEL` → `integerRangeAsStrings(0, 9)`
- `BACKUP_COMPRESSION_THREADS` → `integerRangeAsStrings(-1, 256)`
- `BACKUP_MAX_MB_PER_SECOND` → deliberately none, now documented on the setting: the range is open-ended upwards so no set can express it, and it needs none because any non-positive value already means "throttle disabled" rather than "invalid".

The API/CLI upper bound moved from 1024 to a shared `BackupSettings.MAX_COMPRESSION_THREADS = 256` so the four surfaces cannot drift apart. New test `globalConfigurationRejectsOutOfRangeValues` pins both the rejection and that a rejected value is not left behind.

**2. `release()` racing a worker returning its Deflater.** Fixed rather than left as deferred reclamation - `release()` now `awaitTermination(10s)` after `shutdownNow()`, so no worker can still be in `DeflateChunk.call()` when the pool is drained. Chunks are ≤1MB so termination is immediate; the timeout only exists so a pathological case degrades to the old behaviour instead of hanging.

**3. 4GB test and the CI budget.** Checked: `ParallelZipArchiveWriterTest` is `*Test.java`, so it runs in the **unit-tests** lane, not integration - `timeout-minutes: 60` on the job, no per-test JUnit timeout configured anywhere. Disk is fine (sparse source, ~19MB archive, and `assumeTrue` requires 5GB free as insurance against a filesystem that materialises it). Runtime is ~2s here. But you are right that "the job cap is the only guard" is not a good answer for a test moving 4GB, so it now carries `@Timeout(10, MINUTES)` - two orders of magnitude of headroom, and it fails fast and legibly instead of eating the lane if that ever stops being true.

Pushed as 0aa9d8318.

lvca:
Third round - thanks. Two of the three acted on, one deliberately not.

**Fully-qualified `OperatingSystemMXBean`.** Fixed, `java.lang.management.OperatingSystemMXBean` is imported and only the HotSpot-specific subinterface stays qualified in the `instanceof` (there is no way to import both under one simple name). Fair catch on the CLAUDE.md rule even in a benchmark file.

**Unrecognised `WITH ...` setting silently ignored.** You scoped this out as pre-existing and orthogonal, and it is - but I have implemented it anyway, because it is the *same* failure mode this PR exists to fix, reached by a second route: `WITH encryptionkey = ...` (one wrong character) looks like a request for an encrypted archive and produces a cleartext one. The `switch` now has a `default` throwing a `CommandExecutionException` that names the supported settings.

Worth being explicit that this is a behaviour change, so it can be vetoed: a statement carrying an unknown setting used to succeed and now fails. I judged the compatibility risk to be nil rather than small - until the fix in this PR, **every** setting was silently dropped, so no working statement can have depended on one being accepted; the only statements affected are ones that were already not doing what they said. Covered by `sqlRejectsAnUnknownSetting`. Say the word and I will drop it.

**Duplicated int parsing across the module boundary.** Agreed and left as is - engine cannot depend on integration, and the alternative (String-typed reflective setters on `Backup`) would add public API purely to share eight lines. Noted in the commit message so the next reader does not re-derive it.

One note on local verification: `BackupRestoreDeleteApiIT.restoreAndDeleteBackup` fails on my machine, but it fails identically with this PR reverted - another process on this box is holding port 2480 (`HTTP Port 2480 not available`), so the server under test never binds. Not related to this change; CI has its own port.

Pushed as 09a9a1c8c.

lvca:
Second review of this round (the 22:10 one) - thanks, point 1 was the most useful comment on this PR so far, though not for the reason stated.

**1. EOCD field truncation.** I checked this against the JDK rather than against APPNOTE, by writing the *same logical archive* (one entry just over 4GB, level 0 so the archive itself passes 4GB) with `ZipOutputStream` and with `ParallelZipArchiveWriter` and dumping both tails:

```
jdk  EOCD: entriesDisk=1 entriesTotal=1 cdSize=0000004A cdOffset=FFFFFFFF
ours EOCD: entriesDisk=1 entriesTotal=1 cdSize=00000052 cdOffset=FFFFFFFF
```

So the premise is not right: `ZipOutputStream` does **not** set all four fields to sentinels when zip64 is in use - it sets each field to its sentinel only if that field individually overflows (`writeEND` computes `xlen`/`xoff`/`count` independently). Real count, real directory size, sentinel offset. That is byte-for-byte what this writer emits, so there is nothing to change.

**But the underlying worry was right, for a different reason.** The ZIP64 *end-of-central-directory* branch had no test at all. `entryLargerThan4GBUsesZip64` does not reach it: a 4GB compressible entry produces a ~19MB archive, so the directory offset never overflows - it covers the zip64 data descriptor and the zip64 CEN extra field only. The branch that was untested is the one that fires in ordinary production use, since any database of a few tens of GB yields an archive past 4GB.

New test `archiveLargerThan4GBMatchesTheJdkEndOfCentralDirectory` closes that: it writes a >4GB archive with both writers and asserts the end records field by field **against the JDK's own bytes**, so the two cannot drift. It costs no disk - sparse source, level 0 so the 4GB passes through as stored blocks, and both archives go into a sink that keeps only the last 256 bytes. Runs in 2.7s.

**2. `encryptFile`'s close can still shadow the root cause.** Correct, and a genuine gap in the previous fix. `writeArchive` now rethrows the carried failure from *inside* the callback, so `encryptFile` knows the backup failed and attaches a close failure as `addSuppressed` rather than letting it replace the cause. On the success path a close failure still propagates, since the last bytes may not have reached the disk - that one is the backup's failure.

**3. Default level change prominence.** Added an explicit upgrade note to the release notes: archives are ~7.5% larger after upgrading with no configuration change, and how to get the old size back (`=9`, still several times faster than before since the parallelism is independent of the level, or `=6` for the same size at half the CPU).

**4. Cancelled futures not returning buffers to the pool.** Agreed, no change - `release()` tears the whole pool down immediately after, on an already-failing path.

Pushed as 4e7ba746f.

lvca:
All four addressed. **#3 was the good one - writing the suggested test found a real defect.**

**3. Out-of-range numeric setting via SQL.** You traced it and believed it works. It does work - the value is rejected - but writing the test showed *what the user sees* is wrong:

```
CommandExecutionException: Error on backing up database
  caused by IllegalArgumentException: Backup setting 'compressionLevel' must be between 0 and 9, found 20
```

The reason is buried one level down, because the generic outer `catch (InvocationTargetException)` swallows it into a message that says nothing. On a PR whose theme is "validate at every entry point", validating and then reporting it uselessly is only half the job. The settings loop now catches `InvocationTargetException` itself and surfaces the setter's own message, so the top-level message is the actual reason. Test covers both `compressionLevel = 20` and `compressionThreads = -2`.

**1. Duplicated 256 bound.** Correct - my earlier commit message overstated it as "a shared constant"; `engine` cannot reference `BackupSettings`, so the `GlobalConfiguration` copy is necessarily a separate literal. Both sides now carry a comment saying so, but a comment is not enforcement: `backupThreadBoundMatchesTheGlobalConfiguration` asserts that `GlobalConfiguration` accepts exactly `MAX_COMPRESSION_THREADS` and rejects `+ 1`, so the build fails the moment they drift.

**2. Central directory now written outside the suspension window.** Intentional, and thank you for making me confirm it rather than assume. It is built purely from the per-entry records collected while the files were read - no database state - so nothing about it requires the suspension, and keeping it inside would hold writers throttled for one extra write per file. Now documented in the code so it does not read as an accident. Note the benchmark numbers are unaffected and remain conservative: `concurrentWriterImpact` measures the window as the whole `backupDatabase()` call, so it still charges the central-directory write to the window even though it no longer suspends anything.

**4. `parkNanos` not looped.** It was a simplification rather than a decision, so I have made it exact: `parkUntil` loops against the absolute deadline, so a spurious wakeup cannot let a read through early. An interrupt breaks the loop and leaves the flag set for the reader thread, rather than being spun through - which is what a naive loop would have got wrong.

Pushed as de0b0d1c3.

lvca:
This one (the 22:22 review) crossed with my previous reply - answering it separately.

**Central directory ZIP64 field granularity.** You suggested a comment saying the coarse-grained version was deliberate. It was not deliberate, it was an inconsistency, so I have fixed it instead. Promoting all three fields as soon as any one overflows is legal and self-consistent, but it is not what APPNOTE asks for ("the fields MUST only appear if the corresponding record field is set to 0xFFFFFFFF"), it is not what `ZipOutputStream.writeCEN` does - and, decisively, **this writer already did it per-field in the end-of-central-directory record**, so the two halves of the same class disagreed with each other. Each field is now promoted on its own.

The `entryLargerThan4GBUsesZip64` test now asserts the extra field is 8 bytes - one value, since only the uncompressed size overflows for a 4GB entry that compresses to a few MB at offset 0 - and that it is the same length java.util.zip produces for the same entry. Reference is the JDK, not my reading of the spec.

**Buffer-pool thread confinement.** Agreed, and it is exactly the kind of invariant someone "fixes" into a bug. Comment added at the field declarations spelling out that the `ArrayDeque`s are correct *because* only the `addFile` thread touches them, that the deflater pool is concurrent because workers do, and that the confinement is the invariant rather than the collection type.

**`GlobalConfiguration` 256 bound.** Already done in the previous push - comment at both sites, plus `backupThreadBoundMatchesTheGlobalConfiguration`, which asserts `GlobalConfiguration` accepts exactly `MAX_COMPRESSION_THREADS` and rejects `+ 1`. A grep-discoverable comment was the suggestion; the test makes it a build failure instead.

**`PageManager.getDeferredRAMBytes()` as public API.** Intentional. The issue requires reporting the deferred-RAM high-water mark - the direct signal that writers were throttled - and no accessor existed. You are right that its only caller here is the benchmark; it is deliberately shaped as an ops-facing metric rather than a test hook, and it is the natural thing to wire into the metrics endpoint when the Phase 2 work touches the suspension itself. I have left that wiring out of this PR to keep it to phase 1.

Pushed as b22a9cc99.

lvca:
All four addressed.

**ZIP64 local-header version field** (the one you flagged for double-checking). It is correct, and it is now asserted rather than argued. `ZipOutputStream.writeLOC` has two branches, and for an entry with a data descriptor (general-purpose bit 3, which is every entry here) it writes `version(e)` = 20 and never reaches the branch that would write 45 - because with a data descriptor the sizes are not known when the local header is written, so there is nothing to base an upgrade on. The `entryLargerThan4GBUsesZip64` test now reads the field out of both my archive and the JDK's for the same >4GB entry and asserts they agree at 20, so the reference is the JDK's bytes rather than either of our readings of APPNOTE.

**Info-ZIP verification was manual.** Now durable. `archiveIsAcceptedByAnIndependentUnzipImplementation` shells out to `unzip -t` (which inflates and CRC-checks, not just lists) when the binary is on PATH and skips otherwise, the same shape as the POSIX-only permission test. Worth doing precisely because `ZipInputStream` and `ZipFile` share one implementation of the format, so agreeing with both is weaker evidence than it looks - Info-ZIP is genuinely independent. It runs on the ordinary fixture rather than a multi-GB archive: the streaming data descriptor is the unusual part and every entry carries one, so a small archive exercises the interop risk without the runtime.

**`resolveThreads()` boundaries.** Extracted to a package-private `autoCompressionThreads(int availableProcessors)` so it can be pinned without depending on the core count of whatever runs the suite, and covered at 1/2/3/4/8/16/20/64/128 cores plus the degenerate 0 and -1. You called it low risk and it is - but the formula encodes a deliberate trade (leave half the machine to the workload the backup is already throttling; do not put 32 threads on a job that saturates the disk first) that a one-character edit could change with nothing else failing.

**`parseIntSetting` duplication.** Agreed, leaving it.

**Nit on prose-heavy docs.** Keeping it - the neighbouring entries in `docs/release-26.9.1.md` are the same shape, and for a hand-written binary format the "why" is the part that stops the next reader from breaking it. Happy to trim if maintainers prefer shorter.

Pushed as eb743d273.

lvca:
This one crossed with my previous reply too - answering the 22:37 review separately. All three nits were right, all three fixed.

**3 (taking it first, it is the one that mattered).** The `rawKey != null ? rawKey.toString() : entry.getKey().toString()` fallback falls back to *the buggy match this PR exists to remove*. You are right that it is currently unreachable - `SQLASTBuilder` always sets `keyExpr.value` - but "unreachable today, and if it ever becomes reachable it silently writes cleartext archives again" is not a defensible shape for this particular line. It now throws a `CommandExecutionException` instead of falling back, with a comment saying why there is no fallback, so a future parser change that stopped populating `value` fails loudly rather than quietly reintroducing the defect.

**1. Peak-heap prose.** Wrong in its parts while right in its total, which is the worst kind of doc error because the total looks like it confirms the breakdown. It is two chunks in flight *per thread*, each holding a ~1MB input **and** a ~1MB output buffer - roughly 4MB per thread, ~32MB at 8 threads - not "2 buffers of ~4MB each". Reworded so an operator sizing heap does not have to re-derive it.

**2. Redundant `failure.get()` re-check.** Correct, it is unreachable: `writeArchive` rethrows the carried failure from inside its own callback (it has to, so the stream-closing there knows the backup failed - that was the fix for an earlier round's finding). Rather than comment it as intentional I removed it, because an unreachable safety net mostly teaches the next reader that the throw above might not happen. Replaced with a comment explaining where the throw actually is.

Pushed as f77e7d84d.

PR #6082, opened 2026-08-11
fix(#6069): fold the bucket record-count delta on RESTORE DOCUMENT/VERTEX/EDGE

### LINKED ISSUE #6069: RESTORE DOCUMENT does not restore the bucket record count: count(*) reports one fewer than a full scan, and it persists
## Summary

`RESTORE DOCUMENT` puts the record back but does not restore the bucket's
record count, so `SELECT count(*)` reports one fewer than a full scan returns.
The disagreement survives close and reopen, so this is the persisted count
rather than a stale in-memory statistic.

The engine says so in its own log while doing it:

```
WARNI [LocalBucket] Restoring record #1:1 at its original position (records=1 threadId=1)
```

`records=1` is the count the bucket had after the delete, and it is still 1
after a record has been placed back.

## Reproduction

Java, engine API only, against `arcadedb-engine-26.9.1-SNAPSHOT.jar` taken
from a build of current `main`. Insert 3, delete 1, restore it, and count the
same records two ways at each step.

```java

import java.nio.file.Files;
import java.nio.file.Path;

public class RestoreCountRepro {
  public static void main(String[] args) throws Exception {
    final Path dir = Files.createTempDirectory("restorecount");
    final String path = dir.resolve("db").toString();

    String rid;
    try (Database db = new DatabaseFactory(path).create()) {
      db.command("sql", "CREATE DOCUMENT TYPE Note");
      db.transaction(() -> {
        for (int i = 0; i < 3; i++) {
          MutableDocument d = db.newDocument("Note");
          d.set("i", i);
          d.save();
        }
      });

      try (ResultSet rs = db.query("sql", "SELECT @rid AS r FROM Note WHERE i = 1")) {
        rid = rs.next().getProperty("r").toString();
      }

      report(db, "after insert x3");
      db.transaction(() -> db.command("sql", "DELETE FROM " + rid));
      report(db, "after delete   ");
      db.transaction(() -> db.command("sql",
          "RESTORE DOCUMENT Note RID " + rid + " SET i = 1"));
      report(db, "after RESTORE  ");
    }

    try (Database db = new DatabaseFactory(path).open()) {
      report(db, "after reopen   ");
    }
  }

  private static void report(Database db, String label) {
    long counted;
    try (ResultSet rs = db.query("sql", "SELECT count(*) AS c FROM Note")) {
      counted = ((Number) rs.next().getProperty("c")).longValue();
    }
    long scanned = 0;
    try (ResultSet rs = db.query("sql", "SELECT i FROM Note")) {
      while (rs.hasNext()) { rs.next(); scanned++; }
    }
    System.out.printf("%s  count(*)=%d  full scan=%d%n", label, counted, scanned);
  }
}
```

## Result

```
after delete     count(*)=2  full scan=2
after RESTORE    count(*)=2  full scan=3   <-- disagree
after reopen     count(*)=2  full scan=3   <-- disagree
```

The restored record is fully intact: it keeps its original RID and the
properties given in the `SET` clause, and a scan returns it. Only the count is
wrong.

## Why I think this matters

A wrong count is returned with a normal `200`-shaped success and no warning, so
a caller cannot tell it apart from a correct one. That is the same shape as
#5711, where the HTTP endpoint truncated at 20,000 rows and a truncated
response was indistinguishable from a complete one.

It also compounds: anything that trusts `count(*)` after a restore is wrong
without knowing it, and the restore itself is presumably being used in exactly
the situations where someone is trying to verify that data came back.

Reported against the statement family as it currently stands on `main`; I have
not looked for the write path that owns the counter, so the location is yours
to place.

lvca:
## Summary
- `LocalBucket.restoreRecordAtPosition()` only performs the physical page write for RESTORE DOCUMENT/VERTEX/EDGE, exactly like bucket-level `createRecord`/`deleteRecord` - the caller is responsible for folding the transaction's cached bucket record-count delta (the one `count(*)` reads as an O(1) fast path, see `LocalDatabase.createRecord`).
- Every RESTORE statement, and the `GraphEngine.restoreVertexAt` Java API, restored the record but never applied the `+1` a normal create does. `count(*)` kept reporting one fewer than a full scan after a restore, and since the delta is persisted in `statistics.json`, the drift survived a reopen too.
- Added `RestoreStatementSupport.restoreRecordAndUpdateCount()` so the fix lives in one place shared by RESTORE DOCUMENT/VERTEX/EDGE instead of being repeated (and re-forgotten) at each call site, and applied the same `+1` directly in `GraphEngine.restoreVertexAt`.

Fixes #6069.

## Test plan
- [x] New regression test `RestoreSqlStatementTest.restoreDocumentUpdatesBucketRecordCount` reproduces the issue's exact repro (SQL DELETE, then RESTORE DOCUMENT, then compare `count(*)` vs a full scan, including after a database reopen) - confirmed red before the fix, green after.
- [x] New regression test `RestoreVertexAtTest.restoreVertexAtUpdatesBucketRecordCount` covers the same missing delta on the `GraphEngine.restoreVertexAt` Java API - confirmed red before the fix, green after.
- [x] `mvn -pl engine test` for `RestoreSqlStatementTest`, `RestoreVertexAtTest`, `RestoreRecordAtPositionTest`, `Issue5149CountStarCacheDriftTest`, `Issue5152CountRecomputeRaceTest` all pass.
- [x] Full `com.arcadedb.query.sql.executor.**` and `com.arcadedb.graph.**` packages pass with no regressions.
- [x] `mvn compile` across the whole reactor succeeds.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphEngine.java
@@ -441,6 +441,10 @@ public class GraphEngine {
     final LocalBucket bucket = (LocalBucket) database.getSchema().getBucketById(targetRid.getBucketId());
     final MutableVertex shell = database.newVertex(typeName);
     bucket.restoreRecordAtPosition(targetRid.getPosition(), shell);
+    // #6069: restoreRecordAtPosition only does the physical page write, same as bucket create/delete; the caller
+    // owns folding the transaction's cached bucket record-count delta by the same +1 a normal create applies (see
+    // LocalDatabase.createRecord / RestoreStatementSupport.restoreRecordAndUpdateCount for the SQL-statement path).
+    database.getTransaction().updateBucketRecordDelta(bucket.getFileId(), +1);
 
     final Set<RID> asSet = Set.of(targetRid);
     final long[] counts = reconnectEdgesFromSurvivors(asSet, asSet);
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/RestoreDocumentStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/RestoreDocumentStatement.java
@@ -64,7 +64,7 @@ public class RestoreDocumentStatement extends SimpleExecStatement {
     RestoreStatementSupport.applyBody(doc, insertBody, context);
 
     final LocalBucket bucket = (LocalBucket) database.getSchema().getBucketById(rid.getBucketId());
-    final RID restoredRid = bucket.restoreRecordAtPosition(rid.getPosition(), doc);
+    final RID restoredRid = RestoreStatementSupport.restoreRecordAndUpdateCount(database, bucket, rid.getPosition(), doc);
 
     final ResultInternal result = new ResultInternal(database);
     result.setProperty("operation", "restore document");
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/RestoreEdgeStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/RestoreEdgeStatement.java
@@ -83,7 +83,7 @@ public class RestoreEdgeStatement extends SimpleExecStatement {
     RestoreStatementSupport.applyBody(shell, insertBody, context);
 
     final LocalBucket bucket = (LocalBucket) database.getSchema().getBucketById(rid.getBucketId());
-    final RID restoredRid = bucket.restoreRecordAtPosition(rid.getPosition(), shell);
+    final RID restoredRid = RestoreStatementSupport.restoreRecordAndUpdateCount(database, bucket, rid.getPosition(), shell);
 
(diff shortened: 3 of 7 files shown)


lvca:
Addressed the review's two observations:

1. **Missing RESTORE EDGE count coverage** - added `restoreEdgeUpdatesBucketRecordCount` (confirmed red on the pre-fix code, green with the fix), matching the existing SQL-level (`RestoreDocumentStatement`) and Java-API-level (`GraphEngine.restoreVertexAt`) coverage.
2. **`GraphEngine.restoreVertexAt` duplicates the two-line pattern instead of reusing `RestoreStatementSupport.restoreRecordAndUpdateCount`** - left as is. `com.arcadedb.graph` reaching into `com.arcadedb.query.sql.parser` to save one duplicated line would invert the intended package dependency direction (SQL statements depend on the graph engine, not the other way around) for no real benefit.

PR #6084, opened 2026-08-11
docs: lead the #6072 release note with the measured backup speedup

lvca:
Docs only, follow-up to #6080.

The backup section of `docs/release-26.9.1.md` explained the change well but buried the result. It now opens with the before/after table:

| | before | after (8 threads) |
|---|---|---|
| Backup duration, 1.25 GB database | 18.9 s | **0.68 s** (27x) |
| Effective throughput | 68 MB/s | **1,876 MB/s** |
| Archive size | 323 MB | 348 MB (+7.5%) |
| Writer throughput *during* the backup | 6,809 rec/s (4.3% of normal) | **123,192 rec/s (77%)** |
| How long writers stay throttled | 48.6 s | **2.1 s** (23x less) |

Two things worth flagging about the numbers:

- The headline is **27x**, not 23x - 23x is the reduction in the writer-throttling window (48.6s -> 2.1s), which is arguably the more meaningful figure in production and is now in the heading alongside the speedup.
- The 27x is the **eight-thread** measurement. The default pool is half the available processors capped at 8, so the section now spells out that a smaller machine sees proportionally less (about 4x on two cores, 8x on four, 15x on eight, the full 27x from sixteen up) rather than letting a reader expect 27x everywhere.

All figures come from `BackupCompressionBenchmark`, which is in the tree, so they are reproducible rather than asserted. No prose was removed - the existing explanation now sits under a "What was slow" subheading.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -725,7 +725,31 @@ cannot reach the database it asked for is answered about that database rather th
 
 [#6070](https://github.com/ArcadeData/arcadedb/issues/6070)
 
-## The full backup is several times faster, and the flush-suspension window it opens is proportionally shorter (#6072)
+## Full backup is up to 27x faster, and throttles writers for 23x less time (#6072)
+
+**A full backup of a 1.25 GB database went from 18.9 seconds to 0.68 seconds - 27x - for an archive 7.5%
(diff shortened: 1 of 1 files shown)


lvca:
@claude please change the .md document with your suggestions

Issue #6083, opened 2026-08-11
GraphBatchLoad follow-ups from #6070: silent vertex-buffer clamp, orphan edge records after a failed flush, and five smaller items


lvca:
Follow-ups from #6070 / PR #6076, which wired `RemoteGrpcDatabase.batch()` onto the `GraphBatchLoad` streaming RPC and hardened the server side of it. None of these blocked that merge; they are collected here so the reasoning is recorded once rather than re-derived on the next review of this code.

Ordered by what I would fix first. Items 1 and 2 are defects; 3-7 are design or ergonomics.

---

## 1. `vertex_batch_size` is clamped silently

`grpcw/.../ArcadeDbGrpcService.java:2694`

```java
vertexBatchSize[0] = Math.min(options.getVertexBatchSize(), GRAPH_BATCH_MAX_VERTEX_BUFFER);
```

The cap (`GRAPH_BATCH_MAX_VERTEX_BUFFER = 1_000_000`, line 149) was added in #6076 so a caller-supplied buffer size could not exhaust the server heap on one request. It is right that it exists, but it applies with no log line: a caller who asks for more gets something other than what they set and is never told, which reads as compliance.

This is my own gap from that PR and the cheapest thing here to fix - one `LogManager.instance().log(...)` at `WARNING` naming the requested and effective values. CLAUDE.md's own guidance on bounded coverage is to say what was dropped.

## 2. A failed edge flush leaves orphan edge records

`engine/.../GraphBatch.java` (flush path around line 760)

Found while testing the partial-commit trailer. When the edge flush dies part-way, edge **records** have been created but are connected to no vertex, and they stay in the database:

- `countType("...Link", true)` returned **2**
- edges actually traversable from the vertices: **1**

`Issue6070GraphBatchLoadHardeningIT.countsOnlyFlushedEdgesOnTheTrailerOfALoadThatFailsClosing` had to count reachable edges rather than use `countType` for exactly this reason - the assertion is written around the symptom.

Orphan edge records after a failed bulk load are invisible garbage: nothing references them and no count a user is likely to run distinguishes them from real edges. Worth establishing whether `CHECK DATABASE` detects and reclaims them, and if not, whether the flush should clean up its own partial work before propagating.

Reproducer: a `GraphBatch` load whose edge flush fails part-way - e.g. two buffered edges where the second has a source RID in a bucket that does not exist, so the flush throws after writing the first.

## 3. The leader's gRPC address is not discoverable

[...]

PR #6089, opened 2026-08-12
fix(#6083): reclaim orphan edge records and report an exact edge count on a failed GraphBatch flush

lvca:
Closes items 1, 2, 4 and 5 of #6083 (follow-ups from #6070 / PR #6076). Items 1, 2 and 4 all sit on the same code path, which is why they are taken together.

## Item 2 - a failed edge flush left orphan edge records

A `GraphBatch` edge is written in two steps that do not share a transaction: PHASE 1 of `flush()` creates the record, PHASE 3 links it into its source vertex's edge list. Both flush paths commit in between - the parallel one **must** (the record pages have to be released before the per-bucket tasks touch them), the sequential one does whenever `commitEvery > 0`. A connect pass dying part-way therefore left edge **records** behind that no vertex points at: `countType()` counted them, no traversal reached them, nothing ever reclaimed them.

**The issue asked whether CHECK DATABASE detects and reclaims them. It does not**, and the detail matters:

- `checkEdges` *does* scan every edge record and probe both endpoints for a back-reference (`GraphDatabaseChecker.java:1150-1152, 1183-1185`)...
- ...but a miss only increments the aggregate `missingReferenceBack` counter. No warning naming the record, no entry in `corruptedRecords`, so nothing for the `FIX` variant to delete.
- That same counter is raised by a perfectly healthy **unidirectional** edge, legitimately absent from its target's IN list, so it cannot even be read as an orphan count.
- (`FIX` *does* reclaim orphaned edge **segments** - but those are the internal linked-list chunks, not the edge records at issue.)

So the flush now cleans up after itself before the failure propagates. Deletion goes through `database.deleteRecord`, which is the one inverse correct for **both** creation routes this class uses (`createRecordsBulk` + `indexEdgeProperties` for a single-bucket flush, an ordinary `save()` for a multi-type one) and for every index kind, without a second copy of the key-derivation logic to keep in step. Best-effort by construction: what it cannot reclaim is logged and counted (`getOrphanEdgeRecordsLeaked()`), never allowed to replace the failure the caller is already handling.

### A partial failure also left half-edges

Found while writing the test. A partially-failed flush skipped PHASE 4 entirely, so the edges it *had* durably connected never got their IN back-pointer - half-edges the integrity checker flags. PHASE 3's failure is now **captured rather than thrown**, so PHASE 4 queues the incoming side for exactly the edges that survived, and the failure is rethrown after. A failed flush now leaves only whole edges.

## Item 4 - the partial-commit edge count is now exact, not a lower bound

`totalEdgesCreated` advanced a whole flush at a time, so a load dying mid-flush reported zero for a flush that had in fact made some of its edges durable. It now advances one durable commit at a time, on the failure path as well as the success one. A caller reconciling a failed bulk load re-sends exactly what is missing instead of over-sending a flush's worth and relying on the duplicates being harmless.

## Bonus: the sequential OUT path had no undo log (found by the new test)

`connectOutgoingEdgesSorted` wrote `deferredOutHead`/`outChunkRIDCache` before the transaction holding those segments committed, with nothing to undo it. A rolled-back segment RID survived in the map and `batchUpdateVertexHeadChunks()` stamped it onto the vertex at `close()` - CHECK DATABASE reports `"out edges record is not valid"` and no later read can follow it.

This is exactly the hazard #5950 cycle 3 fixed for the IN direction (`inHeadUndoLog`) and cycle 4 for the parallel OUT direction (the per-bucket local maps). This path was the one left. Same remedy, and the final commit moved inside the pass so the undo log covers it too.

## Item 1 - `vertex_batch_size` was clamped silently

The cap is right - a caller-supplied buffer must not be able to exhaust the server heap on one request - but it applied with no log line, so a caller who asked for more got something else and was never told. Now a `WARNING` naming both the requested and the effective value.

## Item 5 - the reporting convention is settled and written down

The two contradictory precedents are resolved on `graphBatchLoad`'s javadoc: **a failed bulk stream terminates with `onError` and carries its counters in the trailers**, so the caller keeps its status code and generic error handling while still getting what it needs to reconcile. `insertStream` is documented as the predating exception (its summary is a per-chunk report consumed on the success path too, so failure is not a separate channel there) rather than a second precedent.

## Not in this PR

- **Item 3** (the leader's gRPC address is not discoverable) needs a new field in the HA server-list format - a public config-surface change - plus a real 3-node cluster IT to exercise the follower-refusal path. It gets its own PR so that review is not entangled with these defect fixes.
- **Items 6 and 7** were recorded in the issue as deliberately declined; nothing to do.

## Testing

`Issue6083OrphanEdgeRecordTest` (new) covers, for **both** flush paths:
- a failed flush leaves no edge record that a traversal cannot reach (`countType == reachable`, not an upper bound)
- `getTotalEdgesCreated()` equals that same number
- the flush that fails *before* its first commit has nothing durable to reclaim and must not try
- a successful multi-flush, multi-internal-commit load still counts every edge exactly once

**Every assertion was mutation-checked.** Disabling the orphan reclaim, reverting the exact counting, and removing the new OUT undo log each makes a test fail; so does removing the clamp warning. (The issue warned that three tests written during #6076 initially passed against the very bug they were meant to catch.)

`Issue6070GraphBatchLoadHardeningIT`:
- the trailer test now asserts the **exact** count and that `countType` agrees with a traversal - which is precisely what it could not do before, and why it had been written around the symptom
- two new tests for the clamp warning (present when clamped, absent when honoured)
- its shared `vertex()` helper now also sets the property another test declares MANDATORY, so the class no longer passes or fails depending on method order

Suites run green: all `com.arcadedb.graph.*Test` and `com.arcadedb.database.*Test` (engine), all 172 grpcw unit tests + the `Issue6070*` ITs, `RemoteGraphBatchTest` (network), `RemoteGraphBatchIT` (server).

On throughput (correcting the class name from the original description - there is no `GraphBatchBenchmark`): `GraphBatchTest.benchmarkBatchVsStandard` covers the parallel path (3.41x over the standard API, unchanged), and `performance/GraphBatchDrainPerfBenchmark` covers **both** - `parallelFlush=true` at ~653K edges/sec and `parallelFlush=false` at ~742K, also unchanged. So the sequential path, which is where the new per-flush undo-log bookkeeping lives, is represented.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphBatch.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphBatch.java
@@ -33,6 +33,7 @@ import com.arcadedb.engine.LocalBucket;
 import com.arcadedb.engine.WALFile;
 import com.arcadedb.exception.DuplicatedKeyException;
 import com.arcadedb.exception.NeedRetryException;
+import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.index.Index;
 import com.arcadedb.index.IndexInternal;
 import com.arcadedb.log.LogManager;
@@ -339,6 +340,24 @@ public class GraphBatch implements AutoCloseable {
   private long totalEdgesCreated;
   private long totalFlushes;
   private long totalFlushTimeNs;
+  private long totalOrphanEdgeRecordsReclaimed;
+  private long totalOrphanEdgeRecordsLeaked;
+
+  /**
+   * Edges of the flush in progress whose connection to their SOURCE vertex has durably committed (issue #6083
+   * item 4). {@link #flush()} folds exactly this into {@link #totalEdgesCreated}, on the failure path as well as
+   * the success one, so the running count never claims an edge the graph does not hold - and, since #6083, never
+   * under-reports one it does. Reset at the top of every flush.
+   */
+  private int flushDurableOutEdges;
+
+  /**
+   * When the PARALLEL connect pass of the flush in progress failed, the {@code [from,to)} pairs into
+   * {@link #sortIndex} naming the edges it nevertheless connected durably; {@code null} otherwise. The failed
+   * buckets are wherever they are in the partition, so unlike the sequential path's durable prefix this cannot be
+   * expressed as a single bound. Read once by {@link #flush()} right after the pass returns.
+   */
+  private int[] flushDurableOutRanges;
 
   // --- Saved state for restore after close ---
   private final boolean savedReadYourWrites;
@@ -713,8 +732,20 @@ public class GraphBatch implements AutoCloseable {
     // database with the half-written batch visible to the next caller.
     final boolean startedTx = !database.isTransactionActive();
 
+    // Number of buffered edges this flush is about to write. Read after the buffer is reset below, so it cannot
+    // be replaced by edgeCount there.
+    final int bufferedEdges = edgeCount;
+
+    // How many of them are durably connected to their source vertex. Advanced by the connect pass itself, one
+    // durable commit at a time, so a flush that dies part-way reports what the graph actually holds rather than
+    // the whole flush or nothing (issue #6083 item 4).
+    flushDurableOutEdges = 0;
+    flushDurableOutRanges = null;
+
     try {
       // --- PHASE 1: Create edge records for edges that need them (non-light) ---
+      // Hoisted out of PHASE 1 so the failure paths in PHASE 3 can reclaim the records this flush created but
+      // never linked (issue #6083 item 2).
       final RID[] edgeRIDs = new RID[edgeCount];
       beginTx();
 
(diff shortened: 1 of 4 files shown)


lvca:
Thanks - all three points were valid and are addressed in d7f2775.

**The accounting bug: fixed, and it was slightly worse than described.** Confirmed exactly as reported: with `commitEvery > 0` the reclaim pass commits several times but folded its counters only after the loop, so a later commit failing discarded the reclaims an earlier one had already made durable and the catch blamed the whole range. Two additions on top of the reported case:

- the old `toSortPos - fromSortPos` also counted **lightweight** edges, which never allocated a record and so can never leak;
- the local `reclaimed` counter was itself only durable-by-luck, since it included deletions in the still-open transaction.

Both counters now advance as each commit returns, and the failure path charges only the slice that commit was covering, sized by a new `countOrphanCandidates()` that excludes lightweight edges.

Covered by `aReclaimPassThatFailsPartwayKeepsWhatItAlreadyCommitted`, which injects a failure into the **second** reclaim commit through a new `TEST_BEFORE_ORPHAN_RECLAIM_COMMIT_HOOK` - following the fault-injection hooks this class already carries for the vertex and edge commit paths, so the pattern is not new. It asserts against what the database actually holds rather than against the counters alone: `stored - reachable == leaked`, and `reclaimed + leaked` accounts for every orphan candidate exactly once. Mutation-checked - restoring the deferred fold makes it fail on the "must stay counted as reclaimed" assertion.

(The test deliberately leaves unreclaimed orphans behind, so it deletes them before the shared teardown's integrity check. Worth noting *why* they trip that check: not because they are orphans - CHECK DATABASE has no finding for that, which is the whole premise of item 2 - but because the bogus source bucket used to induce the failure makes them dangling links.)

**The boxing: taken, and extended to the IN side.** You were right that it is consistent with existing precedent rather than a new regression, but also right that it is now allocated on every sequential flush. `LongObjectHashMap` does work here - I checked that occupancy lives in its key array, so `containsKey()` still answers true for a key put with a **null** value, which is how "the vertex had no deferred head" is recorded. Switched, iterating with `forEach()` in the catch block.

I also converted the pre-existing `inHeadUndoLog` on the IN side. It predates this PR and had the same cost, and leaving two structurally identical undo logs with different collection types a few hundred lines apart seemed a worse trap than the allocation itself. Covered by the existing #5950 tests.

**The benchmark: my description was wrong, and I have corrected it.** There is no `GraphBatchBenchmark` class - you were right not to find one. The numbers come from `GraphBatchTest.benchmarkBatchVsStandard` (parallel path) and `performance/GraphBatchDrainPerfBenchmark`, and the latter does exercise **both** paths: `parallelFlush=true` at ~653K edges/sec and `parallelFlush=false` at ~742K, both unchanged. So the sequential path, where the new undo-log bookkeeping lives, is represented after all - but the description gave no way to verify that, which was the fair complaint.

lvca:
Thanks - two of the three taken, one does not apply. Pushed in 7bead5a.

**1. Global logger swap: fixed.** Valid, and cheap to remove entirely rather than just note. `CapturingLogger` now tees - it forwards every record to a real `DefaultLogger` as well as capturing it - so the swap is invisible to everything except the assertions. That drops the dependency on these ITs running sequentially, and means a failure inside the capture window no longer loses its diagnostics.

**2. Per-chunk clamp warning: does not apply.** The block that reads `vertexBatchSize` sits inside the `if (batch == null)` first-chunk initialization branch (`ArcadeDbGrpcService.java:2712`, a few lines after `dbRef.set(db)`), so it runs exactly once per stream no matter how many chunks carry options. `getVertexBatchSize` has only that one call site in `graphBatchLoad`. The per-chunk options re-read you're thinking of is in **`insertStream`** (`:2408`/`:2437`, the "options changed after first chunk" rejection) - a different RPC. So no per-chunk log volume from an inflated request.

**3. Multi-bucket reclaim coverage: added.** You were right that this was only structurally implied, and the gap was slightly worse than "low risk": with a single failing bucket the reclaim loop runs exactly one iteration, so a body that stopped after the first bucket would have passed the whole suite. `aParallelFlushReclaimsEveryFailedBucketNotJustTheFirst` now fails two distinct missing source buckets together and pins that both are reclaimed while the buckets that succeeded keep their edges. Mutation-checked - capping the loop at one bucket makes it fail.

Also noted your point about `Arrays.copyOf` in the survivors sizing being dead code: agreed, every bucket in `completedOutgoingBuckets` necessarily had `bucketCounts[b] != 0`, so `survivorCount` always equals `survivors.length`. Leaving it as defensive coding rather than asserting the invariant, since it costs nothing on a path that only runs after a failure.

CI is green on my side for the suites this touches: all `com.arcadedb.graph.*Test` (engine), 172 grpcw unit tests, and the `Issue6070*` ITs.

lvca:
Thanks - no correctness defects this round, and the one actionable point is addressed in b945c83.

**Logger swap still being global: comment added.** You asked for it to be written down if the execution model ever changes, so I have spelled out the constraint the helper depends on and, specifically, what would break: a test asserting the **absence** of a warning could see another test's under cross-class parallelism in a shared JVM (`findWarning` matches on the template, so it cannot tell them apart). Anything relying on that would need per-test logger isolation rather than a singleton swap.

The other two are accepted trade-offs, and both are already documented in the code - recording the reasoning here so it is not re-derived:

- **One begin/commit per failed bucket in the parallel reclaim.** Batching across buckets would be fewer transactions but a wider blast radius: per-commit granularity is exactly what makes the accounting exact after review cycle 1, since a cleanup that fails partway must be able to say which records it already reclaimed. Trading that back for fewer transactions on an already-failing path is the wrong direction.
- **`deleteRecord` walking both endpoints for a record with no back-reference.** Real, and a second traversal per orphan. It buys the one inverse that is correct for both creation routes (`createRecordsBulk` + `indexEdgeProperties`, and the plain `save()` fallback) and every index kind, with no duplicate key-derivation logic to keep in step. On a failure-path-only cleanup that is the right side of the trade.

I consider the review satisfied: three cycles, one real defect found (the reclaim accounting, cycle 1) and fixed with a mutation-checked regression test, plus two test-quality improvements. Local runs are green for every suite this touches - all `com.arcadedb.graph.*Test` (engine), 172 grpcw unit tests, the `Issue6070*` ITs, `RemoteGraphBatchTest` (network) and `RemoteGraphBatchIT` (server) - and no merge conflicts.

PR #6093, opened 2026-08-12
fix(#6073): auto-commit write CypherProcedures like SET/CREATE/DELETE

### LINKED ISSUE #6073: Write CypherProcedures aren't auto-committed like SET/CREATE/DELETE
## Bug

A write `CypherProcedure` (`isWriteProcedure() == true`, e.g. `merge.node`, `merge.relationship`, and the new `refactor.mergeNodes`/`refactor.cloneNodesWithRelationships` from #6060) is not auto-committed by the engine the way built-in write clauses are.

Per `OpenCypherQueryEngine`'s own class Javadoc (`executionDatabase`):

> The write steps auto-commit: `SetStep`, `DeleteStep`, `MergeStep`, `RemoveStep`, `ForeachStep` and a writing `CALL` subquery each call `begin()`/`commit()` directly on `context.getDatabase()` when no transaction is already open.

`CallStep.executeProcedure` is not in that list - it runs lazily on the first `ResultSet` pull (`procedure.execute(...).map(...).iterator()`) with no `begin()`/`commit()` wrapping. A caller invoking a write procedure via `CALL merge.node(...)`/`CALL apoc.refactor.mergeNodes(...)` etc. without its own explicit transaction around the call **and its full consumption** gets a `TransactionException: Transaction not active` (or, worse, silently does nothing if the `ResultSet` is only `.close()`d without being pulled).

## Reproduction

```java
final ResultSet rs = database.command("opencypher",
    "MATCH (a:Person {name:'A'}), (b:Person {name:'B'}) CALL apoc.refactor.mergeNodes([a,b], {}) YIELD node RETURN node");
rs.next(); // throws com.arcadedb.exception.TransactionException: Transaction not active
```

Works only when the caller wraps it explicitly:

```java
database.begin();
final ResultSet rs = database.command("opencypher", "... CALL apoc.refactor.mergeNodes(...) ...");
rs.next();
database.commit();
```

## Motivation

Discovered while implementing/testing #6060's three new write procedures, but it's a pre-existing gap affecting every write `CypherProcedure`, including `merge.node`/`merge.relationship` shipped earlier - there's no dedicated end-to-end test that exercises a write procedure's `CALL` outside of an explicit transaction, so this had gone unnoticed.

## Suggested fix

Make `CallStep.executeProcedure` auto-commit for a write procedure the same way `SetStep`/`DeleteStep`/etc. do, when no transaction is already open - consistent with every other write path in the Cypher engine, and removing an easy footgun for anyone calling a write procedure from a single top-level statement (the common case for HTTP/remote clients).

## Related

- #6060 (where this was found and worked around in tests via explicit `begin()`/`commit()`)

robfrank:
## What does this PR do?

`CallStep.executeProcedure` now auto-commits a write `CypherProcedure` (`isWriteProcedure() == true`) when no
transaction is already open, mirroring how `SetStep`/`DeleteStep`/`MergeStep`/`RemoveStep`/`ForeachStep` already
auto-commit. Previously, a top-level `CALL` to `merge.node`, `merge.relationship`, `apoc.refactor.mergeNodes`,
`apoc.refactor.cloneNodesWithRelationships`, or `apoc.do.when` with no caller-managed transaction threw
`TransactionException: Transaction not active`. Calling a write procedure inside an explicit caller-managed
transaction is unaffected.

Closes #6073

## ⚠️ HA caveat (please read before merging on a cluster deployment)

A bare top-level `CALL` to a write procedure (e.g. `CALL merge.node(...) YIELD node RETURN node`, with no
CREATE/SET/MERGE/DELETE/REMOVE/FOREACH clause anywhere in the statement) is misclassified as read-only by
`SimpleCypherStatement.readOnly` - `anyWriteSubquery` only inspects `CALL { ... }` subquery blocks, never a
bare `CALL <procedureName>(...)`. This is pre-existing (not introduced by this PR), but it was harmless before:
a write procedure with no transaction always threw `TransactionException` before anything could commit.

**This PR changes that failure mode.** `CallStep.executeProcedure`'s new auto-commit is the first thing that
actually calls `begin()`/`commit()` off `context.getDatabase()` for a `CALL`. On a non-HA database that's a
no-op fix (the wrapper is the instance itself off HA). **On an HA cluster**, because of the misclassification
above, `OpenCypherQueryEngine.executionDatabase()` hands that auto-commit the raw, non-Raft-wrapped database
instance instead of the wrapped one - so on a leader the write can commit locally without being proposed to
Raft, and on a follower it can execute the write entirely locally instead of forwarding to the leader. That's a
silent replication gap, not a loud failure, so it's a worse failure mode than the `TransactionException` this
PR fixes, for HA deployments specifically.

Filed as #6094 with the full analysis. Not fixed here: the mechanical fix (checking
`CypherProcedureRegistry.get(name).isWriteProcedure()` in the `readOnly` computation) breaks 10 of
`DoWhenTest`'s 11 tests, because `apoc.do.when.isWriteProcedure()` is unconditionally `true` and those
pre-existing tests rely on `database.query()` succeeding for branches that happen to be read-only - resolving
it needs either a `do.when`-specific classification design decision or a breaking change to those tests'
`query()`/`command()` usage. **Recommend #6094 lands before or alongside this PR on any branch/release used in
an HA deployment**; on a non-HA (single-node) deployment this PR is safe to merge as-is.

## Motivation

Discovered while implementing/testing #6060's new write procedures - it's a pre-existing gap affecting every
write `CypherProcedure`, including `merge.node`/`merge.relationship` shipped earlier, since there was no
dedicated end-to-end test exercising a write procedure's `CALL` outside of an explicit transaction.

## Related issues

- Closes #6073
- Related to #6060 (where this was found and worked around in tests via explicit `begin()`/`commit()`)
- Related to #6094 (HA-routing gap this PR's fix exposes - see the caveat above)

## Additional Notes

Every existing `CypherProcedure` implementation already performs its mutation synchronously inside `execute()`
before returning its `Stream<Result>`, so wrapping the single call site in `CallStep.executeProcedure` was
sufficient - no changes needed in the procedure implementations themselves. Rollback fires on both the
`CommandParsingException` (client error) and general exception paths, so a failed write procedure call never
leaves a dangling transaction.

Verified TDD-style: ran the new test class against the pre-fix code first (4 of 7 failed with the exact
`TransactionException` from the issue body), then restored the fix and confirmed all 7 pass.

Follow-up (per review): the returned result stream is forced to fully materialize before `commit()` runs on
the auto-commit path, since `.map()`/`.iterator()` are otherwise lazy - without that, a future write procedure
that streamed its results lazily instead of eagerly could commit before it ever mutated anything. Documented
the "must materialize eagerly" contract on `Procedure#execute`. Added regression tests for
`apoc.refactor.cloneNodesWithRelationships` and `apoc.do.when` (the latter dispatches through a nested
`database.command()`/`query()` call rather than mutating directly, so it exercises a materially different code
path).

## Test plan

- [x] New test class `CallStepWriteProcedureAutoCommitTest` (9 tests): merge.node/merge.relationship/
      apoc.refactor.mergeNodes/apoc.refactor.cloneNodesWithRelationships/apoc.do.when auto-commit with no
      explicit transaction, explicit-transaction usage stays unaffected, chained `CALL` (UNWIND) auto-commits
      per row, failed calls leave no dangling transaction, existing error-path behavior
      (`CommandSemanticException` for a bad config) is preserved.
- [x] `mvn -pl engine -am compile` passes.
- [x] `com.arcadedb.query.opencypher.procedures.**` (all procedure tests, read-only and write): 589/589 pass.
- [x] `com.arcadedb.query.opencypher.executor.**` (full Cypher execution-step package): 720/720 pass.

## Checklist


robfrank:
Re the review's Must-fix finding (bare `CALL <write procedure>` misclassified as read-only, HA routing risk):
verified against the code, it's real. Attempted the suggested fix (`SimpleCypherStatement.anyWriteProcedureCall`,
mirroring `anyWriteSubquery`) - it breaks 10 of `DoWhenTest`'s 11 tests, because `apoc.do.when.isWriteProcedure()`
is unconditionally `true` and those pre-existing tests rely on `database.query()` succeeding for branches that
happen to be read-only, which only works today because of this same misclassification bug. Resolving it cleanly
needs a design decision on `do.when`'s classification (or a breaking change to `DoWhenTest`'s `query()`/`command()`
usage), which this unattended review loop isn't positioned to make, and constraints.md forbids modifying existing
tests. Filed #6094 with the full analysis (HA leader/follower failure modes, the DoWhenTest tension, suggested
paths forward) for follow-up with proper HA test coverage.

PR #6099, opened 2026-08-12
fix: derive component file names deterministically on rename

lvca:
## Problem

Renaming a type re-derives the file name of every component it owns. That was done by guessing where the component name ended:

- `PaginatedComponentFile.rename()` branched on `newFileName.contains(".") && newFileName.contains("_")` to decide whether the argument was already a complete file name, and otherwise spliced the suffix starting at the **first** `_` of the current file name.
- `PaginatedComponent.rename()` re-derived the new component name from the **first** `.` of the renamed OS file name.
- `LocalDocumentType`, `LocalVertexType`, `LSMTreeIndex` and `HashIndex` had each grown their own incompatible variant of the same splice.

Every one of those guesses is wrong for a name that contains the character being searched for, and both `_` and `.` are legal in a type name.

Observed today on `main`:

| rename | result |
|---|---|
| `Und_Er` -> `Other` | files become `Other_Er_0.14...`, `Other_0_Er_0_in_edges.16...` |
| `Plain` -> `ren.amed` | `NumberFormatException`, stray zero-byte file, orphaned edge bucket, **records lost** |
| `Src` -> `na_me.dot` | file moved to a literal `na_me.dot`, losing its `.fileId.pageSize.vVersion.ext` tail |

The last case is silent: `FileManager.scanDirectoryForComponentFiles` computes an unrecognised extension and skips the file, so the records disappear with no error until a count comes back short.

## Change

- `PaginatedComponentFile.renameComponent(String)` recomposes the file name from the fields the open-time parser already resolved (`fileId`, `pageSize`, `version`, `fileExtension`). Nothing is inferred from either the old or the new name. `rename(String)` is reduced to plain path resolution.
- `PaginatedComponent.rename()` takes a **component name** and no longer re-parses the OS file name.
- `LocalSchema.rebaseComponentName()` is the single shared formula: `encode(newType) + componentName.substring(encode(oldType).length())`. The suffix carrying the bucket index, the `_out_edges`/`_in_edges` marker and the index timestamp is preserved verbatim. Used by the bucket, edge-bucket and index rename paths.
- `LSMTreeIndex`/`HashIndex.updateTypeName` rename through the component, so the component name and the `FileManager` map follow the file instead of going stale.
- `LocalSchema.createBucket` validates the name. A bucket name is the last segment of its component file path and, unlike a type name, is not percent-encoded on the way there.

### Neo4j importer

With the naming made consistent, the importer's allow-list (`[A-Za-z0-9_ -]+`) is no longer needed. It rejected labels a real Neo4j export routinely carries (dots, accents, colons) even though the engine stores them fine: a type name is percent-encoded on its way to the bucket name, and the component file name is parsed right-to-left off its fixed tail, so a dot in it round-trips.

It now refuses only what cannot round-trip: path separators and the `.`/`..` sentinels, `*` (untouched by percent-encoding, illegal on Windows), the `~` used to join the labels of a multi-label node, and `,` (which `TypeBuilder` rejects for type names).

Nodes with **no labels** are kept on the root `Node` type in both passes. `syncSchema()` used to throw `NullPointerException` on them, before `parseVertices()` could reach its "skip it" branch, so the previous graceful-skip behaviour was unreachable.

## Tests

Three new classes, 11 tests. 9 of the 11 were confirmed failing against the unmodified code before the fix.

- `TypeRenameComponentNamingTest` pins the naming for every combination of `_` and `.` on both sides of a rename, plus edge buckets and an index, and asserts every component file still carries its tail (the invariant that catches the silent-skip case).
- `BucketNameValidationTest` covers the rejected names through both the schema API and SQL DDL, and pins that dots stay legal.
- `Neo4jImporterLabelCharactersIT` covers dotted/accented/spaced labels, dotted relationship types, unlabelled nodes, and the labels still refused.

Verified: full engine suite 11686 tests with one pre-existing failure (`Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent`, which fails 3/3 on a clean `main` at c66f0c8ec and is unrelated); full Neo4j importer suite green including the two pre-existing rejection tests; full reactor build clean.

### Independent importer fix in the same change

`parseEdges()` now guards the `"label"` lookup with `has()`/`isNull()` before reading it, matching what `syncSchema()` already did. `JSONObject.getString()` throws on a missing key, so a relationship record without a label used to abort the whole import rather than being skipped with a warning, which also made the `type == null` skip branch immediately below it unreachable. This is separate from the node "no labels" NPE described above.

### Follow-up from review

A bucket created with `createBucket()` and attached with `addBucket()` carries a name of its own, not derived from the type name (`Issue774AddBucketWithIndexTest` attaches `"O202203"` to type `"Order"`). `rebaseComponentName()` required the type-name prefix and threw `SchemaException` for those, and since that is a `RuntimeException` while `rename()` only rolled back on `IOException`, the throw escaped the rollback: buckets renamed earlier in the loop kept their new file names while `saveConfiguration()` never ran.

`rebaseComponentName()` now returns `null` for a component name not derived from the type name, and the bucket, edge-bucket and index paths skip it. The rollback paths also catch `SchemaException`, and treat a name they cannot rebase back as corruption rather than passing `null` to the file layer. Covered by `TypeRenameComponentNamingTest.renameTypeWithCustomNamedBucketLeavesThatBucketAlone`, confirmed reproducing the reported failure before the fix.

### New constraint on type names

`createBucket()` validates unconditionally, including the `encode(typeName) + "_" + i` names `TypeBuilder` generates, and `URLEncoder` leaves `*` unescaped. So `createVertexType("Foo*Bar")` (and the SQL equivalents) now fails at bucket-creation time rather than producing a file name that is illegal on Windows. This is a new restriction on type names, not only bucket names. It applies to creation only, so existing databases containing such a name still open.

### Second follow-up from review

- `rebaseComponentName()` decided whether a component belonged to the type with a bare prefix match, so an attached bucket whose own name merely starts with the type name was treated as derived: renaming `Order` rewrote an attached `OrderArchive` into `OrdersArchive`. Every derived name is `<encodedTypeName>_<something>`, so the check now requires that `_`. Covered by `renameTypeLeavesAloneACustomBucketSharingTheTypeNamePrefix`, confirmed reproducing before the fix.
- `HashIndex.updateTypeName()` wrapped its `IOException` in `IndexException`, a sibling of `SchemaException` rather than a subclass, so it slipped past the widened rollback in `LocalDocumentType.rename()`. It now throws `SchemaException`, matching `LSMTreeIndex.updateTypeName()`. The HASH rename path also gained coverage (`renameTypeWithHashIndexKeepsIndexUsableAfterReopen`), which it had none of.

### Third follow-up from review: index rollback

`rename()` renames the type's indexes in a loop after the bucket loop, and the rollback restored only the buckets. With two indexes, a failure on the second left the first one's file renamed on disk and its metadata pointing at the new type name while the type itself reverted.

The indexes renamed so far are now tracked like `removedBuckets` and replayed back to the old name, with a failure to restore one marking the type corrupted, matching the bucket loop. An index is registered *before* its `updateTypeName()` call, not after: that call walks the index's own per-bucket sub-indexes, so a failure part way through leaves that index half renamed and it needs rolling back too. The unwind runs indexes first, since they were renamed last.

`aFailedIndexRenameRollsBackTheIndexesAlreadyRenamed` injects the failure by parking a directory on the path the last index's file is about to move to, so `Files.move` raises the `IOException` the rename path wraps. Confirmed failing before the fix, leaving a `renamed.Multi_0_<timestamp>` index file behind next to a correctly-reverted `Multi_0` bucket.

This also gives the rollback path its first test on this branch or on `main`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/PaginatedComponent.java
+++ b/engine/src/main/java/com/arcadedb/engine/PaginatedComponent.java
@@ -84,22 +84,16 @@ public abstract class PaginatedComponent extends Component {
     reservedPageCounter.set(pageCount.get());
   }
 
-  public void rename(final String newName) throws IOException {
+  public void rename(final String newComponentName) throws IOException {
     // #4928: the bounded wait can give up on a wedged flush. Renaming with pages still in flight would let
     // the flush thread write to the file under its old identity: abort loudly instead, the rename can be
     // retried once the flush recovers.
     if (!PageManager.INSTANCE.waitAllPagesOfDatabaseAreFlushed(database))
       throw new IOException("Cannot rename component '" + componentName
           + "': pages are still pending flush after the no-progress timeout (see arcadedb.flushAllPagesTimeout)");
-    file.rename(newName);
-
-    // Derive the new component name from the renamed OS file name.
-    // PaginatedComponentFile.rename() preserves the bucket suffix in the filename
-    // (e.g., "ASKED_0.3.65536.v1.bucket"), so extract everything before the first "."
-    // to get the correct component name including the bucket index.
-    final String newOsFileName = file.getOSFile().getName();
-    final int dotPos = newOsFileName.indexOf('.');
-    final String newComponentName = dotPos >= 0 ? newOsFileName.substring(0, dotPos) : newName;
+    // The argument is the new component name, not a type name and not a file name: the file layer recomposes the
+    // '.fileId.pageSize.vVersion.ext' tail itself, so nothing here has to locate where the name ends.
+    file.renameComponent(newComponentName);
 
     database.getFileManager().renameFile(componentName, newComponentName);
     componentName = newComponentName;
--- a/engine/src/main/java/com/arcadedb/engine/PaginatedComponentFile.java
+++ b/engine/src/main/java/com/arcadedb/engine/PaginatedComponentFile.java
@@ -154,6 +154,14 @@ public class PaginatedComponentFile extends ComponentFile {
     }
   }
 
+  /**
+   * Renames the underlying OS file. {@code newFileName} is the complete file name, tail included; a name without a
+   * path separator is resolved against the current parent directory.
+   * <p>
+   * To change only the component name and keep the {@code .fileId.pageSize.vVersion.ext} tail, call
+   * {@link #renameComponent(String)}: this method cannot infer where the component name ends, and every heuristic
+   * for doing so is wrong for a name that itself contains the delimiter being searched for.
+   */
   public void rename(final String newFileName) throws IOException {
     channelLock.writeLock().lock();
     try {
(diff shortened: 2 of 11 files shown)


lvca:
The rollback gap raised across the last two reviews is tracked as #6103, which now covers both symptoms: the index loop having no per-index undo, and `bucketMap` not being re-keyed when the bucket rollback runs. It also notes that the rollback branch has no test on this branch or on `main`, so a fault-injecting test is really the prerequisite for fixing either symptom with confidence.

The Windows-illegal characters and reserved device names noted as an error-quality gap are tracked as #6104.

Both are pre-existing and deliberately out of scope here: this PR keeps to the naming derivation and the validation that closes the directory-escape, and the rollback changes it did make were the minimum needed to stop the new code paths from bypassing the existing rollback.

Issue #5277, opened 2026-07-14
HA: database bootstrap metadata lives inside raftStorageDirectory — wiping Raft state forces full snapshot reinstall of intact databases


ruispereira:
Observed during the reformat recovery documented in #5275 (ArcadeDB `26.8.1-SNAPSHOT`, 3-node k8s cluster, databases and Raft storage both persisted on PVCs).

## What happened

To recover from the silently-shrunk Raft configuration (#5275) we wiped `arcadedb.ha.raftStorageDirectory` on all three nodes and force-recreated the pods (parallel bootstrap). The **database files on the PVCs were untouched** — only the Raft/group state was removed.

On the subsequent bootstrap election, two of the three nodes reported their intact on-disk database as having no committed state:

```
[ArcadeStateMachine] Database 'superx' bootstrap mismatch (local lastTxId=-1 / fp=84444e0e..., baseline lastTxId=369 / fp=4377c198...); reinstalling from leader-shipped full snapshot
```

`lastTxId=-1` on nodes whose `databases/superx` directory was fully present strongly suggests the bootstrap metadata (lastTxId / fingerprint) is persisted inside `raftStorageDirectory` rather than alongside the database — so wiping Raft state also wipes the evidence that the local data is valid, and the node falls back to a **full snapshot reinstall** of every database.

The system behaved *safely* (the elected source had the freshest copy, fingerprint reconciliation worked, cluster converged healthy in ~58s) — this issue is about the **cost**, not correctness.

## Why it matters

Wiping Raft storage is currently the only operator remedy for a peer silently removed from the configuration (#5275), and the wipe must cover all nodes (a partial wipe resurrects the wrong configuration). That turns a metadata-only repair into a full re-transfer of every database to N-1 nodes. On our stage cluster that is megabytes; on a production graph it would be the entire dataset over the pod network, multiplied by the number of nodes — turning a ~60s recovery into potentially hours, with the cluster not Ready throughout (`readinessRequiresHA`).

## Suggested improvement

Persist the per-database bootstrap metadata (lastTxId + fingerprint) with the database itself (e.g. under `databaseDirectory/<db>/`), or recompute the fingerprint from the on-disk data when metadata is missing, so that a Raft-state-only reformat can re-validate local data instead of discarding it. Raft log/group state and database baseline identity have different lifecycles — coupling their storage means operators cannot reset one without paying for both.

lvca:
[...]

Issue #4831, opened 2026-06-30
Snapshot transfer has no manifest / completeness check — a cleanly-truncated download is accepted as complete


ruispereira:
**Affected version:** 26.6.1
**Component:** `SnapshotHttpHandler.serveSnapshotZip` (producer) / `SnapshotInstaller.downloadSnapshot` (consumer)
**Severity:** HIGH

### Summary

The leader streams a ZIP of config + schema + every component file + `.ts.sealed`,
then `finish()`. There is no manifest, no expected file count, no `Content-Length`,
and no checksum on the install path (the `/checksums` endpoint and
`computeFileChecksums` exist but the installer never consults them — it only checks
`responseCode == 200` then parses the ZIP). The consumer treats "ZipInputStream
reached EOF without throwing" as success.

### Code

`SnapshotInstaller.java:372-418` (consumer, no manifest/length/checksum check),
`SnapshotHttpHandler.java:269-296` (producer, no Content-Length/manifest).

### Impact

A response truncated at a ZIP-entry boundary (proxy close, write-timeout watchdog)
deserializes as a valid-but-short archive with fewer files than the DB has; the
follower writes `.snapshot-complete` and opens a structurally-incomplete database.

### Suggested fix

Emit a manifest as the final ZIP entry (names + sizes + CRCs via
`computeFileChecksums`) or send `Content-Length`; the installer must verify every
manifest file is present with matching size/CRC before writing `.snapshot-complete`.

lvca:
Fixed

PR #6100, opened 2026-08-12
feat(#6075): page-level copy-on-write snapshot replaces the flush suspension (phase 2b)

### LINKED ISSUE #6075: Page-level copy-on-write shadow: replace flush suspension with a real snapshot primitive (Phase 2b, prerequisite for incremental backup)
## Context

Phase 2 of the backup optimization roadmap in [`docs/optimize-backup.md`](https://github.com/ArcadeData/arcadedb/blob/main/docs/optimize-backup.md). Phase 1 (#6072, faster full backup) is nearly done. **Read section 8 of that document before starting: it is the design spec for this issue.**

This issue implements the **page-level copy-on-write shadow**, design option 2(b). It replaces `PageManager.suspendFlushAndExecute` with a real point-in-time snapshot primitive.

The alternative that was evaluated and **rejected** is 2(a) (fuzzy copy plus WAL replay, the `pg_basebackup` model). The rationale is in sections 8.4 and 8.8 and is summarised below, because it explains why this issue is scoped the way it is.

### The suspension is a missing snapshot primitive, and five things pay for it

`suspendFlushAndExecute` freezes the data files so they can be read raw. While it holds, dirty pages accumulate in `PageManagerFlushThread.deferredByDatabase` bounded by `FLUSH_SUSPEND_MAX_DEFERRED_RAM` (default 512 MB), and past that cap committing threads are throttled. Current callers:

- `FullBackupFormat.backupDatabase` (`integration/.../format/FullBackupFormat.java:91`)
- the HA snapshot ship, `SnapshotHttpHandler.java:212` and `:240`
- the HA database verify, `PostVerifyDatabaseHandler.java:128`
- plus LSM and vector index compaction, which is **postponed for the entire window** (`LSMTreeIndex.java:206` and `:340`, `LSMVectorIndex.java:5208` and `:5324`). A long backup silently stops index compaction today.

### It is a prerequisite for incremental backup (Phase 3), not a nicety

This is the decisive argument and the reason to do it before Phase 3 rather than after.

Phase 3a ships only the pages whose version differs from the parent backup's manifest. That is sound only if, for every page, **the image written into the archive and the version recorded in the manifest are the same image**. A fuzzy copy breaks that pairing silently:

1. The backup copies page P at t1, when P is at v5.
2. The manifest records P's version at t2, when P has advanced to v7.
3. The archive contains P at v5 while claiming v7.
4. The next incremental compares live P (still v7) against the manifest's v7, concludes "unchanged", ships nothing.
5. Restoring full plus incremental yields P at v5. **Silent, permanent data loss**, undetectable without a full comparison against the source.

WAL replay does not rescue this. Under 2(a) replay repairs the data at restore time, but the manifest must describe the *post-replay* state, which does not exist anywhere at backup time. The same argument applies to the Phase 3b changed-page bitmap: it must be cleared atomically at a defined snapshot point, and with no single t0 a page dirtied inside the fuzzy window can fall out of both bitmap generations.

The shadow makes the pairing exact by construction: the page image and its version are read together, under the same per-page lock, from the same snapshot. **The manifest cannot disagree with the archive.**

## Why ArcadeDB is well suited to this

- **There are exactly two physical page-write call sites in the engine**, and both already funnel through the same per-page I/O lock, `concurrentPageAccess(pageId, true, ...)` (`engine/src/main/java/com/arcadedb/engine/PageManager.java:721`):
  - `PageManager.flushPage` (`:650`), the normal sync/async flush path
  - `PageManager.writePageWithLock` (`:525`), used by `TransactionManager.applyChanges` (`:636`) for Raft and recovery replay

  Putting the hook inside `concurrentPageAccess`'s write branch covers both by construction, adds no new lock, and introduces no new lock-ordering risk.
- Pages are fixed size and self-describing (`BasePage.PAGE_VERSION_OFFSET = 0`), so bookkeeping is a flat `(fileId, pageNumber) -> offset` map, not a range tree.
- Files only grow at the tail, so everything past the snapshot's page count needs no protection.

### Snapshot primitive

Add `PageManager.openSnapshot()` returning a handle exposing per-file page counts, `lastTxId`, and `readPage(pageId)`. **Build and test it standalone before wiring any consumer to it.**

### Opening a window (t0), and the ordering that matters

Flipping the flag is **not** the same as opening the window. `concurrentPageAccess` grants the page slot via `pendingFlushPages.putIfAbsent`, so at the instant of the flip a writer may already be inside its write section having read `window == null`. It will overwrite its page without shadowing it, and the snapshot then holds that page as of t0 plus epsilon, torn against everything else.

Required order:

1. Take the database read lock.
2. Set `window` (flag on).
3. Wait for the flush pipeline to drain: `waitAllPagesOfDatabaseAreFlushed`, which already exists and is timeout-protected (#4928). Note this is a **full** drain, unlike today's `waitForCurrentFlushToComplete`, which waits only for the in-flight batch.
4. Wait for any in-flight Raft apply to complete. Applies are serialised on the state machine thread, so this is a single bounded wait.
5. **Only now** record per-file `pageCount` and `lastTxId`. That tuple is t0.
6. Create the shadow (RAM-first, see below).

Because t0 is defined *after* the barrier, any write that slipped through un-shadowed is simply part of the snapshot state and everything stays consistent. **Doing this in the other order (drain, then flip) produces subtly torn snapshots that pass tests and fail in production.**

The tempting alternative, a shared read-lock on the write path so the flip can take the exclusive side, is correct but adds permanent cost to the hot path. Do not.

This full drain at t0 is a short bounded stall, and it is the *last* stall. It also fixes the recency gap documented in section 1.3 of the design doc, because the snapshot point becomes a real transaction boundary instead of an arbitrary flush boundary.

### On every page write, inside the existing `concurrentPageAccess` write section

```
if (window != null && pageNumber < window.snapshotPageCount(fileId) && !window.contains(pageId)) {
    read the current on-disk page;
    append it to the shadow;
    record pageId -> offset;
}
write the new page as usual;
```

### Snapshot reader

```
offset = window.shadowOffset(pageId);
return offset >= 0 ? readFromShadow(offset) : readFromFile(pageId);
```

under `concurrentPageAccess(pageId, false, ...)`, **bypassing the read cache**, since the cache holds the newest committed image rather than the snapshot image.

### Closing

Refcount the window using the pattern already in `setSuspended` (#5068), so the shadow is not released while a reader is still resolving pages through it. Then drop the map and delete the shadow file.

## Zero cost when no window is open

With the window closed the added work is one field read plus a branch the predictor always gets right, inside a critical section that already performs a 64 KB `pwrite`: roughly 1 ns against 10-50 us. Keeping it that way is implementation discipline:

- **One nullable field**, `volatile SnapshotWindow window`, null when inactive. A single read gives both the flag and the state, with no double-read race.
- **Not a listener list.** A `List<PageWriteListener>` iterated per write costs real cycles and can go megamorphic. A single nullable field inlines away.
- **Nothing before the null check**: no `PageId` allocation, no map lookup, no extra lock acquisition.
- **Do not widen the critical section.** The hook goes strictly inside the existing `concurrentPageAccess` write branch.

## Challenges to solve (full detail in section 8.3)

**C1. The reader must go page by page, so bulk file copy is lost.** Today the backup does `FileInputStream.transferTo(zipStream)`, one streaming copy per file. Preventing a torn read without WAL replay requires the per-page interlock, so the copy becomes a loop of `pageSize` reads with a lock acquisition and map lookup each. Reads stay sequential and readahead still helps, but **this must be benchmarked, not assumed negligible**. Mitigation: batch runs of pages absent from the shadow map under one lock acquisition.

**C2. The shadow protects page content, not file existence.** `LSMTreeIndexCompactor` does **not** take the database write lock, so the backup's `executeInReadLock` never excluded its `dropFile` calls (`LSMTreeIndexCompactor.java:118` and `:347`). The `isPageFlushingSuspended` guard is currently the only thing preventing compaction from dropping a file under a running backup. Removing the suspension removes that protection.

**Solve this by deferring the physical delete until the window closes, not by re-adding a compaction guard.** Compaction then runs freely during a backup, which is one of the main wins. The "is a window open" check is the same single field read, so this needs no second registry.

**C3. Overlapping windows need per-window pre-images.** The current suspension is refcounted and deliberately collapses overlapping windows (#5068). That does not translate: a window opened at t1 needs content as of t1, not the pre-image captured for t0. Either give each window its own map and copy on first write per window, or serialise windows outright. `SnapshotHttpHandler` already holds a per-database lock serialising its entries, so serialising is an acceptable v1 simplification **if it is a documented decision rather than an accident**.

**C4. Unbounded shadow growth.** Worst case the shadow reaches the working set dirtied during the window, which on a small hot database approaches the database size. This is the "snapshot full" failure of SQL Server sparse-file snapshots. Needs a configurable cap and defined breach behaviour: abort with a clear error, or fall back to the old suspend-and-freeze mode. Silent unbounded growth is not acceptable.

**C5. Write amplification and added flush latency.** One extra read plus write per distinct page per window, never repeated, decaying as the window progresses, but landing on the flush thread. Append-only shadow keeps the extra write sequential. Make the shadow **RAM-first with disk spill** bounded by the C4 cap: for short backups at modest write rates the shadow may never touch disk, removing this cost entirely in the common case.

**C6. The map must be cheap.** Open-addressing primitive `long` map (pack `fileId` and `pageNumber` into the key, value the shadow offset), not `HashMap<PageId, Long>`. 16 bytes per entry, so one million shadowed pages is 16 MB. Boxing on the flush path is not acceptable.

**C7. File growth is free, truncation is not.** Pages appended beyond the snapshot's `pageCount` need no shadow, the reader simply ignores them. This is why per-file page counts must be captured at t0. Truncation is rarer but does need handling.

**C8. Crash safety is easy.** The shadow is pure scratch: recovery never reads it, so a crash mid-window leaves an orphan file to delete on open. It must never be mistaken for a WAL or data file: distinct extension, excluded from `LocalSchema`'s supported file extensions.

## Implementation order

1. Build the snapshot primitive standalone and test it in isolation under concurrent write load, before any consumer uses it. The barrier ordering above is the thing to test hardest.
2. Migrate consumers one at a time, starting with the HA database verify (lowest blast radius), then the backup, then the HA snapshot ship.
3. Keep the existing suspend-and-freeze path as a configurable fallback for at least one release, and use it automatically when the C4 cap is breached.
4. Solve C2 by deferring file deletion.

## Testing

Follow the project TDD convention: state the verification plan, write the tests first, then implement.

- **Barrier ordering under load**: open a window while writers are mid-`concurrentPageAccess`, assert no page in the snapshot is newer than t0. This is the test that justifies the whole design; make it aggressive and run it repeatedly.
- **Snapshot consistency**: take a snapshot under sustained write load, restore it, assert the restored database opens and its `lastTxId` matches the recorded t0.
- **Recency**: assert the restored database contains every transaction committed before t0 (this is the section 1.3 gap that the full drain closes).
- **Overlapping windows** (C3): whichever policy is chosen, assert it holds.
- **Shadow cap breach** (C4): assert the defined fallback or error, never silent growth.
- **Compaction during a window** (C2): force a compaction mid-window and assert both that the snapshot stays readable and that compaction was not postponed.
- **Crash mid-window** (C8): assert the orphan shadow is cleaned on open and recovery is unaffected.
- **Encrypted and plain backups** still round-trip.
- Existing coverage to extend: `integration/src/test/java/com/arcadedb/integration/backup/`, `server/src/test/java/com/arcadedb/server/backup/`, `engine/src/test/java/com/arcadedb/engine/PageManagerFlushSuspend*Test.java`.

## Benchmark before and after (required)

Same discipline as #6072: measure, do not argue from first principles, and put the numbers in the PR description.

- **Steady-state write throughput and commit latency (p50/p95/p99) with no window open**, before and after the change. This must be indistinguishable from the baseline. It is the claim the whole "zero cost when idle" design rests on.
- **Write throughput and commit latency percentiles with a window open**, versus today's suspend-and-freeze behaviour. This is the headline win: today's backup throttles writers for its full duration.
- **Deferred-RAM high-water mark** and whether the `FLUSH_SUSPEND_MAX_DEFERRED_RAM` cap is reached. After this change it should never be reached during a backup.
- **Flush-thread drain rate** with and without a window open (C5).
- **Shadow peak size** as a fraction of database size, across write rates (C4 sizing).
- **Backup wall-clock**: per-page locked reads versus today's bulk `transferTo` (C1). Report this honestly even if it is a regression, since it trades against the writer-impact win.
- **t0 barrier duration**: how long the full flush drain takes at various queue depths. This is the one remaining stall and its size should be known.

Tag benchmarks so they stay out of normal CI: `@Tag("benchmark")` for throughput comparisons, `@Tag("slow")` for long functional regressions. See `CLAUDE.md`.

## Acceptance criteria

- Steady-state write performance is statistically unchanged when no snapshot window is open.
- A backup no longer throttles writers: the deferred-RAM cap is not reached and commit latency percentiles during a backup are close to baseline.
- Index compaction runs during a backup instead of being postponed.
- HA verify and the HA snapshot ship use the same primitive and no longer suspend flushing.
- The snapshot point is a real transaction boundary: the restored database contains every transaction committed before t0.
- Page image and page version are provably paired, which is the property Phase 3 will build on.
- The old suspend-and-freeze path remains available as a configurable fallback.

## Related

- #6072, Phase 1 (faster full backup). Phase 1 also shrinks this phase's cost: window length is backup length, so a 4x faster backup means 4x less shadow disk and 4x fewer amplified writes.
- Phase 3 (incremental backup and PITR) depends on this issue, see section 8.8 of the design doc.

lvca:
Closes #6075. Phase 2b of the backup roadmap (`docs/optimize-backup.md`, section 8), the prerequisite for phase 3 (incremental backup).

## What this replaces

`PageManager.suspendFlushAndExecute` freezes the data files so they can be read raw. While it holds, dirty pages accumulate in `PageManagerFlushThread.deferredByDatabase` bounded by `FLUSH_SUSPEND_MAX_DEFERRED_RAM` (512 MB), past which committing threads are throttled outright, and LSM/vector index compaction is postponed for the whole window. It is replaced by a real point-in-time snapshot primitive: `PageManager.openSnapshot()`.

While a window is open, the first write to any page that existed at t0 copies that page's current on-disk image into a shadow, inside the per-page I/O slot the write already holds. A reader resolves every page as "shadow if present, data file otherwise" and therefore sees exactly the t0 image. **Flushing is never suspended**, so writers run at full speed for the whole operation.

## Zero cost when no window is open

The hook goes strictly inside the write branch of `PageManager.concurrentPageAccess`, the single funnel both physical page-write call sites (`flushPage`, `writePageWithLock`) already go through - so it covers them by construction, with no new lock and no new lock-ordering risk. With no window open it is one volatile field read plus a branch the predictor always gets right, inside a critical section that already performs a 64 KB `pwrite`. One nullable field, not a listener list; nothing before the null check.

## The t0 barrier

Flipping the flag is not the same as opening the window: `concurrentPageAccess` grants the write slot with a `putIfAbsent`, so at the instant of the flip a writer can already be inside its write section having read "no window". Every writer is therefore excluded **before** t0 is stamped, each by the mechanism that already serializes it:

1. `waitAllPagesOfDatabaseAreFlushed` - a **full** queue drain, not just the in-flight batch. This is what makes t0 a genuine, recent transaction boundary and closes the recency gap of section 1.3 (a restored backup could be a few hundred transactions behind).
2. The flush thread is suspended and its in-flight batch awaited, so the async path parks on a **batch** (= transaction) boundary. If commits slipped in between the drain and the suspension, the suspension is released (flushing what it deferred) and the barrier retries.
3. `TransactionManager`'s new `applyLock`, taken exclusively, excludes Raft and crash-recovery replay - the one writer that reaches the files outside the flush pipeline. The shared side is one uncontended acquisition per replayed transaction, never on the local commit path.
4. The global page-manager lock excludes synchronous commits, which publish their pages from `publishPages`.

Only then are the per-file page counts and `lastTxId` recorded and the window published.

## Challenges from the issue

- **C1 (per-page locked reads).** Solved better than the issue proposed, and without the suggested batching heuristic: a write during the window **always** shadows the page before touching the file, so a run of pages is read in one bulk call with **no per-page lock at all**, and the shadow is re-probed afterwards. Any page a writer touched underneath the read is now in the shadow and its t0 image is taken from there, so a torn read is always detected. Sequential I/O and OS readahead are preserved; the common case costs one probe per page against a primitive map. (`PaginatedComponentFile.readPages`.)
- **C2 (file existence).** Solved by deferring the physical delete, not by re-adding a compaction guard. `FileManager.dropFile` hands the file to the open windows, refcounted, and the last one to close deletes it. **Index compaction now runs during a backup** instead of being postponed - covered by a test that asserts `scheduleCompaction()`/`compact()` both succeed inside a window and that the snapshot stays readable afterwards.
- **C3 (overlapping windows).** Supported directly rather than serialized: each window owns its shadow and a write captures the same freshly read pre-image into every window that still needs it. That is correct because a page's content at capture time IS its content at t0 for any window that has not shadowed it. Only the barrier is serialized, per database.
- **C4 (unbounded growth).** `arcadedb.pageSnapshotMaxSize` (1 GB default). On breach the window is invalidated and every read fails loudly with `PageSnapshotException` - never silent growth, never a torn artifact. A backup restarts on the suspend-and-freeze path.
- **C5 (write amplification).** The shadow is RAM-first (`arcadedb.pageSnapshotMaxRAM`, 64 MB) with append-only disk spill, so a short backup at a moderate write rate never touches the disk.
- **C6 (cheap map).** Open-addressing primitive `long -> long` map, keys packing `fileId`/`pageNumber`, 16 bytes per entry, no boxing and no allocation on the flush path.
- **C7 (growth/truncation).** Per-file page counts captured at t0; pages beyond them need no shadow. A short read past the recorded count fails the window rather than serving zeros.
- **C8 (crash safety).** `.pshadow` is pure scratch, absent from `LocalDatabase.SUPPORTED_FILE_EXT`, and orphans are deleted at the next open.

## Consumers migrated

- **Full backup** (`FullBackupFormat`), with automatic fallback: a window that cannot hold its point in time restarts the whole backup on the suspend-and-freeze path, which throttles writers but always completes.
- **HA database verify** (`PostVerifyDatabaseHandler`). The checksum is byte-for-byte the value `PaginatedComponentFile.calculateChecksum()` produces, so a peer still on the fallback path compares equal.
- **HA snapshot ship** (`SnapshotHttpHandler`). It also now reports the snapshot's t0 transaction id for the `last-tx-id.bin` marker (#5277) instead of the live counter, which could name a transaction whose pages were still in the queue.

`arcadedb.pageSnapshotEnabled = false` selects the historical path outright.

## Verification

New `engine/.../PageSnapshotTest` (9 tests). Every test that asserts "the snapshot did not change" also asserts the **live** files did change in the meantime, so none can pass vacuously:

- the t0 image survives rewriting every record three times, with pre-image captures proven to have happened;
- a restored snapshot opens, passes `CHECK DATABASE` and holds **exactly** the records of t0;
- the barrier under four concurrent writer threads: the restore passes `CHECK DATABASE` and its record count sits between what was committed before the barrier and what was committed when it returned - a torn snapshot fails the integrity check, a moving one fails the upper bound;
- overlapping windows each keep their own point in time, including after the inner one closes;
- a cap breach invalidates the window and every read throws, while the live database is unharmed;
- a file dropped mid-window survives until the window closes and not one moment longer;
- index compaction completes inside a window;
- an orphan `.pshadow` is deleted on open and never registered as a component file.

New `integration/.../Issue6075SnapshotBackupIT` (3 tests): a backup under concurrent writers restores to a point in time on both paths, and - the headline win, measured rather than asserted from first principles - **the fraction of the backup during which page flushing is suspended drops from >90% on the fallback path to <50%** (in practice a few ms of barrier), with the fallback measured in the same run as the control so the comparison does not depend on machine timing.

Existing suites re-run green: the whole `com.arcadedb.engine` and `com.arcadedb.index`/`com.arcadedb.schema` packages (2094 tests), the whole `integration` module including its ITs (206 tests), and the HA snapshot/verify unit tests (106 tests). `BackupCompressionIT.aFailedBackupThrowsAndLeavesNoArchive` needed its failure injection moved to the schema configuration file and now runs on both paths: the snapshot reads page files through channels the database already holds open, so revoking a data file's permission mid-run no longer breaks a backup at all.

## Follow-ups deliberately left out

- `SnapshotHttpHandler`'s `/checksums` sub-path still uses the suspension: it reads the database directory raw (`SnapshotManager.computeFileChecksums`) rather than the registered page files, so it needs a different shape. It is a rarely-called diagnostic.
- Benchmarks for steady-state throughput/latency with no window open, shadow peak size across write rates, and t0 barrier duration at various queue depths are not in this PR; the suspension-fraction measurement above is what defends the acceptance criterion this change is about.

https://claude.ai/code/session_01BVFzhZA6cH3FUhnekk1oEz

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -911,4 +911,13 @@ the next open.
 Overlapping windows are supported rather than serialized: each owns its shadow, and a write captures the same
 freshly read pre-image into every window that still needs it.
 
+### One incidental change to every backup archive
+
+The two backup writers disagreed about ZIP entry timestamps: the parallel one stamped the source file's
+modification time, the single-threaded one left `ZipOutputStream` to stamp the moment the entry was written. Adding
+the stream-based entry API the snapshot path needs routed both through one implementation, so they now agree on the
+source file's modification time - which is the more useful of the two, and was already what the default
+(multi-threaded) writer produced. Nothing about restoring changes; only the timestamps a listing shows for an
+archive written with `arcadedb.backup.compressionThreads = 0`.
+
 [#6075](https://github.com/ArcadeData/arcadedb/issues/6075)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -840,3 +840,75 @@ directory written at all: a failed one aborts the writer instead, so even if the
 (permissions, a full or read-only filesystem) what is left is structurally invalid rather than plausible.
 
 [#6072](https://github.com/ArcadeData/arcadedb/issues/6072)
+
+## A backup no longer freezes the database: point-in-time page snapshots replace the flush suspension (#6075)
+
+Phase 1 (#6072) made the backup 27x faster, which shortened the window during which writers are throttled but did
+not remove the throttling. This removes it. `PageManager.suspendFlushAndExecute` - which parks the page-flush
+thread for the whole of a backup, an HA snapshot ship or an HA database verify, so dirty pages accumulate in RAM
+until `arcadedb.flushSuspendMaxDeferredRAM` (512 MB) and committing threads are then throttled outright - is
+replaced by a real snapshot primitive: `PageManager.openSnapshot()`, a **page-level copy-on-write shadow**.
+
+While a window is open, the first write to any page that existed at t0 first copies that page's current on-disk
+image into a shadow, inside the per-page I/O slot the write already holds. A reader resolves every page as "shadow
+if present, data file otherwise" and therefore sees exactly the t0 image, however far the live database has moved
+on. Flushing is never suspended, so writers run at full speed for the whole operation.
+
+Three things follow from that, beyond the writer impact:
+
+- **Index compaction runs during a backup again.** The suspension was the only thing stopping `LSMTreeIndex` and
+  `LSMVectorIndex` compaction from dropping a file under a running backup (`LSMTreeIndexCompactor` does not take the
+  database write lock, so the backup's read lock never excluded it), and both refuse to compact while it is held. A
+  long backup silently stopped compaction. Now a file dropped while a window is open simply has its physical
+  deletion deferred until the window closes, so compaction proceeds.
+- **The snapshot point is a real transaction boundary.** The window opens on a FULL drain of the flush queue, not
+  the in-flight batch alone, so the restored database contains every transaction committed before t0 - closing the
+  recency gap that could leave a restored backup a few hundred transactions behind. The HA snapshot ship now also
+  reports the t0 transaction id (#5277's `last-tx-id.bin` marker) instead of the live counter, so the marker can no
+  longer name a transaction whose pages were still in the queue.
+- **The page image and its version are provably paired**, because they are read together, under the same per-page
+  lock, from the same snapshot. That is the property incremental backup (phase 3) has to be built on: with a fuzzy
+  copy, a manifest recorded at a different instant than the image it describes makes the next incremental conclude
+  "unchanged" and silently drop a revision.
+
+### Zero cost when no snapshot is open
+
+With no window open the write path pays one volatile field read and a branch the predictor always gets right, inside
+a critical section that already performs a page-size (64 KB) `pwrite`: roughly 1 ns against 10-50 us. That is
+implementation discipline rather than design - a single nullable field, not a listener list; nothing before the null
+check; and the hook goes strictly inside the existing `concurrentPageAccess` write branch, which is the single funnel
+both physical page-write call sites already go through, so it needs no new lock and adds no lock-ordering risk.
+
+### The barrier, which is where the correctness lives
+
+Flipping the flag is not the same as opening the window: a writer can already be inside its write section having
+read "no window", and would then overwrite its page unshadowed. So every writer is excluded before t0 is stamped,
+each by the mechanism that already serializes it - a full flush-queue drain, then a suspension parking the flush
+thread on a batch (that is, transaction) boundary, then the transaction manager's apply lock to exclude Raft and
+recovery replay, then the global page-manager lock to exclude synchronous commits. Only then are the per-file page
+counts and `lastTxId` recorded. Performing these in the other order produces subtly torn snapshots that pass tests
+and fail in production, which is why the barrier is what the test suite hammers hardest.
+
+### Reading stays sequential
+
+The obvious snapshot reader takes the per-page I/O slot for every page, turning one streaming `transferTo` into a
+lock acquisition per page. It is not needed: a write during the window ALWAYS shadows the page before touching the
+file, so a run of pages is read in one bulk call with no lock at all and the shadow re-probed afterwards - any page
+a writer touched underneath the read is now in the shadow, and its t0 image is taken from there. A torn read is
+therefore always detected, and the common case costs one probe per page against a primitive open-addressing map.
+
+### Bounded, and with a fallback
+
+The shadow is RAM first (`arcadedb.pageSnapshotMaxRAM`, 64 MB) spilling to an append-only scratch file, and capped
+overall (`arcadedb.pageSnapshotMaxSize`, 1 GB). Worst case it grows to the working set dirtied during the window,
+which on a small very hot database approaches the database size, so an uncapped shadow would reproduce the "snapshot
+full" failure of sparse-file snapshots. On breach the window is invalidated and every read fails loudly: a backup
+restarts on the suspend-and-freeze path, which throttles writers but always completes. That path stays available
+and can be selected outright with `arcadedb.pageSnapshotEnabled = false`. The shadow file is pure scratch - recovery
+never reads it, its extension is not a supported component extension, and an orphan left by a crash is deleted at
+the next open.
+
+Overlapping windows are supported rather than serialized: each owns its shadow, and a write captures the same
+freshly read pre-image into every window that still needs it.
+
+[#6075](https://github.com/ArcadeData/arcadedb/issues/6075)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -451,6 +451,34 @@ public enum GlobalConfiguration {
       (unbounded, pre-4728 behavior).""",
       Long.class, 512),
 
+  PAGE_SNAPSHOT_ENABLED("arcadedb.pageSnapshotEnabled", SCOPE.DATABASE,
+      """
+      Serve point-in-time readers (full backup, HA database verify, HA snapshot ship) from the page-level \
+      copy-on-write shadow instead of freezing the data files with FLUSH_SUSPEND_MAX_DEFERRED_RAM-bounded flush \
+      suspension (issue #6075). With the shadow the only stall is one bounded flush-queue drain when the window \
+      opens; after that writers run at full speed for the whole operation and index compaction is no longer \
+      postponed. Set to false to fall back to the historical suspend-and-freeze path, which is also selected \
+      automatically when a shadow breaches PAGE_SNAPSHOT_MAX_SIZE.""",
+      Boolean.class, true),
+
+  PAGE_SNAPSHOT_MAX_RAM("arcadedb.pageSnapshotMaxRAM", SCOPE.DATABASE,
+      """
+      Maximum amount of RAM (in MB) of page pre-images a snapshot window keeps in memory before spilling the rest \
+      to a scratch file next to the database (issue #6075). The shadow only ever holds the pages DIRTIED while the \
+      window is open, once each, so a short backup on a moderately busy database often never touches the disk at \
+      all - which is the point of making it RAM first. Raise it to trade heap for keeping longer windows in memory.""",
+      Long.class, 64),
+
+  PAGE_SNAPSHOT_MAX_SIZE("arcadedb.pageSnapshotMaxSize", SCOPE.DATABASE,
+      """
+      Hard cap (in MB, RAM plus spill file) on the size a single snapshot shadow may reach before the window is \
+      declared overflowed (issue #6075, challenge C4). Worst case the shadow grows to the working set dirtied while \
+      the window is open, which on a small very hot database approaches the database size, so an uncapped shadow \
+      would reproduce the 'snapshot full' failure mode of sparse-file snapshots. On breach the window stops \
+      capturing and every reader fails loudly, so the consumer can fall back to the suspend-and-freeze path - never \
+      a silently truncated or torn snapshot. Set to 0 for no cap.""",
+      Long.class, 1024),
+
   EXPLICIT_LOCK_TIMEOUT("arcadedb.explicitLockTimeout", SCOPE.DATABASE, "Timeout in ms to lock resources on explicit lock",
       Long.class, 5000),
 
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -32,6 +32,7 @@ import com.arcadedb.engine.ErrorRecordCallback;
 import com.arcadedb.engine.FileManager;
 import com.arcadedb.engine.LocalBucket;
 import com.arcadedb.engine.PageManager;
+import com.arcadedb.engine.PageSnapshot;
 import com.arcadedb.engine.TransactionManager;
 import com.arcadedb.engine.WALFile;
 import com.arcadedb.engine.WALFileFactory;
@@ -2493,12 +2494,31 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
     transactionManager.checkIntegrity();
   }
 
+  /** Removes any {@code .pshadow} scratch file left behind by a snapshot window that a crash interrupted (#6075). */
+  private void deleteOrphanSnapshotShadows() {
+    final File[] orphans = new File(databasePath).listFiles(
+        (dir, name) -> name.endsWith("." + PageSnapshot.SHADOW_FILE_EXT));
+    if (orphans == null)
+      return;
+    for (final File orphan : orphans) {
+      LogManager.instance().log(this, Level.FINE, "Deleting orphan snapshot shadow file '%s'", null, orphan.getName());
+      if (!orphan.delete())
+        LogManager.instance().log(this, Level.WARNING, "Cannot delete the orphan snapshot shadow file '%s'", null, orphan);
+    }
+  }
+
   private void openInternal() {
     try {
       DatabaseContext.INSTANCE.init(this);
       setLockingEnabled(configuration.getValueAsBoolean(GlobalConfiguration.BACKUP_ENABLED));
 
+      // #6075 (challenge C8): the copy-on-write shadow of a snapshot window is pure scratch - recovery never reads
+      // it - so a crash mid-window leaves nothing but an orphan file to delete here. Its extension is deliberately
+      // absent from SUPPORTED_FILE_EXT, so the FileManager scan below never mistakes one for a data file.
+      deleteOrphanSnapshotShadows();
+
       fileManager = new FileManager(databasePath, mode, SUPPORTED_FILE_EXT, resolveExternalBucketPath());
+      fileManager.setDroppedFileHandler(file -> PageManager.INSTANCE.deferFileDrop(wrappedDatabaseInstance, file));
       transactionManager = new TransactionManager(wrappedDatabaseInstance);
 
       open = true;
--- a/engine/src/main/java/com/arcadedb/engine/FileManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/FileManager.java
@@ -250,6 +250,27 @@ public class FileManager {
     this.droppedFileHandler = handler;
   }
 
+  @FunctionalInterface
+  public interface FileSetOperation<T> {
+    T execute() throws IOException;
+  }
+
+  /**
+   * Runs {@code operation} holding the monitor {@link #dropFile} and {@link #getOrCreateFile} take, so the registered
+   * file set cannot change underneath it.
+   * <p>
+   * The snapshot t0 barrier (issue #6075) uses this to close a TOCTOU window that would otherwise defeat the whole
+   * deferred-deletion mechanism: it lists the files it is about to snapshot and only THEN publishes the window, and
+   * {@code PageManager.deferFileDrop} keeps a dropped file alive only for windows that are already published. A
+   * {@code dropFile} landing in that gap - which index compaction can do at any moment, since it takes no database
+   * lock - would delete a file the snapshot had just claimed. Holding this monitor across list-and-publish makes the
+   * two mutually exclusive. Lock ORDER matters and is the same on both paths: this monitor first, then
+   * {@code PageManager}'s snapshot registry lock.
+   */
+  public synchronized <T> T executeWithFileSetLocked(final FileSetOperation<T> operation) throws IOException {
+    return operation.execute();
+  }
+
   public void dropFile(final int fileId) throws IOException {
     final ComponentFile file;
     synchronized (this) {
--- a/engine/src/main/java/com/arcadedb/engine/FileManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/FileManager.java
@@ -49,6 +49,19 @@ public class FileManager {
   private volatile     List<FileChange>                          recordedChanges   = null;
   private volatile     Thread                                    recordingThread   = null;
   private final static PaginatedComponentFile                    RESERVED_SLOT     = new PaginatedComponentFile();
+  /**
+   * Decides whether the physical deletion of a dropped file has to be DEFERRED because a point-in-time snapshot
+   * window still needs to read it (issue #6075, challenge C2). Installed by {@code LocalDatabase} and delegating to
+   * {@code PageManager}; {@code null} on a file manager not attached to a database (tests, tooling), where nothing
+   * can be snapshotting anyway.
+   */
+  private volatile DroppedFileHandler droppedFileHandler = null;
+
+  @FunctionalInterface
+  public interface DroppedFileHandler {
+    /** @return true when the caller must NOT delete the file: an open snapshot window took the deletion over. */
+    boolean deferDrop(ComponentFile file);
+  }
 
   public static class FileChange {
     public final boolean create;
@@ -232,6 +245,11 @@ public class FileManager {
     fileIdMap.clear();
   }
 
+  /** Installs the snapshot-aware deletion policy for dropped files (issue #6075). */
+  public void setDroppedFileHandler(final DroppedFileHandler handler) {
+    this.droppedFileHandler = handler;
+  }
+
   public void dropFile(final int fileId) throws IOException {
     final ComponentFile file;
     synchronized (this) {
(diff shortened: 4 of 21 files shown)


lvca:
Thanks - item 1 was real, and worse than described. Addressed in 444f717.

**1. TOCTOU between list and publish.** Confirmed and fixed. `buildSnapshot` + `registerSnapshot` now run as one step under the same `FileManager` monitor `dropFile` already holds (`FileManager.executeWithFileSetLocked`), so the two are mutually exclusive. Lock order is identical on both paths - file set, then snapshot registry - so no cycle is introduced, and nothing takes the file-set monitor and then the page-manager lock.

The follow-on you spotted is fixed too: the reader now converts `readPages`' "file is closed" `IllegalArgumentException` into `PageSnapshotException`, so a window that loses a file for *any* reason takes the documented fallback instead of failing the backup outright.

New regression test `aFileDroppedWhileAWindowIsOpeningIsNeverDeletedUnderIt` drops a type against a snapshot barrier over 30 rounds. With the fix reverted it fails 3 out of 3 runs with exactly the symptom you predicted (`Cannot read pages from 0 because the file 'Racy0_0.2.65536.v0.bucket' is closed`); with the fix it passes.

**2. Overflow fallback not exercised end to end.** Agreed. New `Issue6075SnapshotBackupIT.aShadowOverflowMidBackupFallsBackAndStillProducesARestorableArchive`: 1 MB shadow cap with a writer rewriting the whole 10 MB dataset in a loop, so the window is certain to overflow while the throttled backup is still reading. It asserts both that the fallback was actually taken (not merely survived) and that the archive restores and passes `CHECK DATABASE` - which, as you noted, also confirms `writeArchive`/`encryptFile` are safe to run a second time against the same path.

**3. PageShadow serializing captures.** Fixed rather than documented. The spill offset is reserved under the monitor, the write happens outside it at that absolute position (positional `FileChannel.write` is thread safe and the reserved ranges are disjoint by construction), and the index entry is published only after the bytes are down - which is also precisely what keeps the bulk reader's re-probe sound. Reads resolve the slot under the monitor and copy outside it. The class comment now describes what the code actually does.

**Nit.** `SnapshotFile.lastModified` is no longer captured at t0; it is read lazily on the consumer's thread, so the barrier does no `stat()` at all. Nothing depended on the value being the t0 one, and a file dropped since then is still on disk because its deletion is deferred.

Re-ran green: `com.arcadedb.engine` (659), the whole `integration` module including its ITs (81 IT + 126), and the HA snapshot/verify unit tests (101).

lvca:
Both concrete suggestions are in 672ee47.

**Lock order stated centrally.** Now in `PageManager`'s class Javadoc, since you are right that no call site shows more than two of the five locks: barrier monitor -> apply lock -> page-manager lock -> FileManager monitor -> snapshot registry, with a note on why the two paths that take subsets of it agree and why `PageSnapshot.close()` holds nothing when it drops its retained files.

**Barrier give-up is now logged.** Good catch that this was only inferable from behaviour - the existing WARNING covers the drain timeout, a different condition. Hitting `SNAPSHOT_BARRIER_ATTEMPTS` now logs its own WARNING naming the database and the attempt count, and saying explicitly that it is sustained write pressure rather than an error. The clean-break and give-up branches are split while I was there, which also removes the double-negative read you flagged.

**`applyLock` shared side.** Your reading is right that it is Raft apply / recovery replay only. It is one uncontended `ReentrantReadWriteLock` read acquisition per replicated **transaction**, not per page, on a path that already parses a WAL record, reads pages through the cache and writes them to a file - so it is not measurable against that. I did not run a dedicated HA replay throughput benchmark; the `ha-integration-tests` lane on this PR covers the functional side. If it looks worth measuring rather than reasoning about, I would rather do it as a follow-up with a proper harness than bolt a throughput assertion onto this PR.

**Extracting a snapshot registry from `PageManager`.** Agreed that this is where it is heading, and phase 3 is the right moment - it will add manifest and changed-page state that belongs with it rather than in `PageManager`. Left alone here so the diff stays reviewable as one mechanism.

**Sampler-thread flakiness in `theBackupNoLongerSuspendsPageFlushingForItsWholeDuration`.** Noted. The fallback control is measured in the same run precisely so a slow runner moves both numbers together, and the margins (>0.9 vs <0.5) are wide, but I will treat it as the first suspect if that test ever goes red.

lvca:
### CI triage: the three red lanes are pre-existing/flaky, not this change

I traced each one rather than assuming, because two of them touch areas this PR really does modify.

**`integration-tests` - `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites`.** This one deserved the most scrutiny, since it is a 3-node Raft test and this PR rewrites the HA snapshot ship. It is not caused by the change:

- The diff between this branch's **green** integration run (444f717) and this **red** one (672ee47) is *one Javadoc block, one split `if`, and one WARNING log line*. Nothing that can affect Raft bootstrap election.
- Both snapshot ships in the failing run **succeeded** - the log shows `Database 'graph' reinstalled after bootstrap mismatch` twice, so the shipped archive was complete and installable.
- The actual failure signature is a bootstrap-election race: `SEVER Database 'graph': local lastTxId=8 is GREATER than cluster bootstrap lastTxId=7. This peer's data is fresher than the cluster's chosen baseline`, followed by that peer being reinstalled from the stale baseline - hence `Types: DB1 5 <> DB2 0` in the teardown comparison.
- Across this branch's three CI runs the integration lane failed on a *different* test each time (gRPC / none / this one) with essentially identical code.

**`unit-tests` - two failures, both pre-existing.** `Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent` is red on **every** recent `main` run I checked (31596421869, 31586918142, 31585501960), with the same "N thread(s) failed" shape and a different N each time. `TimeSeriesTagDictionaryTest.aSecondWaveOfIdsIsAlsoResolvedByReloading` passes 5/5 locally and cannot be reached by this change: with no snapshot window open, `activeSnapshots` is null and the hook is a no-op, and nothing else on the write/flush path changed. That lane on `main` is chronically red with a rotating cast (31585501960 also had `Issue5588PrimitiveLocationIndexTest`).

**`python-e2e-tests`** - testcontainer readiness timeout, on a *different* module each time (bolt in run 1, sqlalchemy in run 3), with run 2 fully green. A real startup regression would fail the first container deterministically.

Every test this PR adds or touches passed in CI: `Issue6075SnapshotBackupIT` 4/4, `BackupCompressionIT` 22/22, `FullBackupIT` 6/6, `Issue5517BloomFilterBackupIT` 1/1, and the whole `com.arcadedb.engine` package including the 10 `PageSnapshotTest` cases.

lvca:
You found a real one. Fixed in 684cf7d.

### 🔴 The torn read is genuine, and worse than "reachable"

Your reading of `fill()` is exactly right, and the failing case is the *ordinary* one rather than an edge case: pages enter the shadow in write order, which has nothing to do with `READ_RUN_PAGES` boundaries, so a run holding a sparse subset of shadowed pages is what normally happens. The bulk read spans an already-shadowed page whenever its neighbours are not shadowed, overwrites the pre-image with live bytes, and the `!fromShadow[i]` guard then declines to put it back. My own comment on the bulk read claimed step 3 handled it. It did not.

The re-probe is now unconditional across the whole run, as you suggested - a page absent from the shadow makes it a no-op, so one pass covers both the already-shadowed case and the written-underneath-the-read race. `fromShadow` is gone entirely; the first probe now only computes the bulk-read bounds.

New regression test `aSparseSetOfShadowedPagesInsideOneReadRunStillServesT0`: 20k records over 16 KB pages, then every 500th record updated so the dirty pages land scattered, then whole-file checksums compared against t0. **It fails with the old gating restored and passes with the fix**, and - as you predicted - no existing test caught it: reverting the fix turned exactly one test red, the new one.

### 🟠 Fallback now genuinely unconditional

Right again, and the `executeInReadLock` wrapping makes it worse than a missed `catch`. Rather than widening three catches, `openSnapshot` no longer declares checked exceptions at all: an I/O error enumerating files, or an interrupt during the barrier, is normalised into `PageSnapshotException` (interrupt flag preserved). One exception type, so the policy cannot drift between the three consumers - which also answers your last minor point about folding the fallback into a shared helper. I left the three call sites separate because their shapes genuinely differ (the backup restarts the whole archive, the verify collects aside and merges only on success, the HTTP handler can only fall back before it starts streaming), but they no longer differ in *what they catch*.

### 🟡 Silently dropped files

Fixed by naming them. The two cases where skipping is correct are now documented and logged at FINE: a null slot (dropped file id) and a not-yet-open file (the `FileManager` reserved slot for an allocated-but-uncreated id). Anything else is logged too, so an incomplete archive can no longer be reported as successful with nothing in the log.

### 🟡 `SnapshotHttpHandler` swallow divergence

Not intentional - good catch. The snapshot path now runs `serveSnapshotZip` through `CodeUtils.executeIgnoringExceptions`, exactly as `suspendFlushAndExecute` does for the fallback branch, so the handler behaves identically on both. A transfer that dies mid-stream has already committed its response anyway; the follower's missing-manifest check (#4831) is what turns it into a retry.

### 🟡 `getTotalPages()` in the barrier - keeping it, with the reason written down

This one I am not changing, and I have added a comment so nobody "fixes" it later. Unlike `lastModified`, which is cosmetic, the page count at t0 is *definitional*: it decides which pages the snapshot covers. Reading it lazily would let a page appended after t0 into the snapshot un-shadowed, which is the exact torn read the design exists to prevent; reading it before the barrier would exclude a page appended before t0, leaving the archive short of what its own page count claims. It is one `fstat` per file against a flush-queue drain the same barrier already performed.

### 🟡 ZIP timestamps

Worth stating plainly: the two writers *disagreed* before. `ParallelZipArchiveWriter` already stamped the source file's mtime; `ZipStreamArchiveWriter` left `ZipOutputStream` to stamp write time. Routing both through one entry implementation made them agree on the mtime - the same value the default multi-threaded writer already produced, so the common configuration is unchanged. Called out in the release note now.

### Minor: reuse

- `PageShadow` now uses `com.arcadedb.utility.LongLongHashMap` (which I gave a plain `put`, next to its accumulate-style `add`) instead of a second hand-rolled long map. Its `Long.MIN_VALUE` empty marker can never collide with a packed page key, which is what makes it a drop-in. ~70 lines of duplicated map gone.
- `PageManagerFlushThread`'s two scans of `pageIndex` by database share one helper.

Re-ran green: `com.arcadedb.engine` (661), the `integration` ITs (81), and the HA snapshot/verify tests (101).

lvca:
Thanks - all three addressed in f4a1e8a, one of them by measuring rather than changing.

**1. Test tagging: measured, and the answer is no.** I timed the class per method rather than guessing:

| test | time |
|---|---|
| `barrierProducesAConsistentPointInTimeUnderConcurrentWriters` | 0.52s |
| `aFileDroppedWhileAWindowIsOpeningIsNeverDeletedUnderIt` (30 rounds) | 0.11s |
| **whole class, 12 tests** | **2.0s** |

Against the two you cited as the convention: `Issue5152CountRecomputeRaceTest` is 1.6s and `Issue5147SuperNodeChunkRaceTest` is 4.1s - each for a *single* method. My entire class costs about what one of those single tagged methods does, and the convention in CLAUDE.md is "noticeably long ... multi-second elapsed time", which neither method reaches.

The reason not to tag anyway is that the tags are a **partition**, not a label: `-Dgroups`/`-DexcludedGroups` route each test to exactly one lane, so tagging would move these two *out* of the default `unit-tests` lane. They are the two regression tests that most directly defend this change's core correctness - the t0 barrier and the drop-vs-barrier TOCTOU - so I want them running on every PR, not only in the slow lane. Cheap and central beats correctly-labelled-but-elsewhere.

**2. `isSnapshotWindowOpen` with no production caller - kept, and now says why.** It is the public counterpart of `isPageFlushingSuspended` for embedders and diagnostics; a snapshot primitive with no way to ask "is one open" is a worse API. More importantly the Javadoc now records that it is deliberately **not** what the compaction guards consult - as you noted, `LSMTreeIndex`/`LSMVectorIndex` still gate on `isPageFlushingSuspended`, which the snapshot path never sets, so compaction runs during a snapshot *by construction* rather than through a second check that could drift out of step with the first. That was worth writing down so nobody later "wires it up" and reintroduces the coupling this design removed.

**3. `shutdown()` race - fixed.** You are right, and the original comment argued the wrong way round: "this only fires on an abnormal teardown" is exactly why the read should be locked, since an abnormal teardown is when a window can be mid-registration. It now reads `activeSnapshots` under `snapshotRegistryLock`, which costs nothing on a path that already joins the flush thread.

Re-ran the `com.arcadedb.engine` package green (661).

PR #6102, opened 2026-08-12
fix(#5870): renumber vector ids densely during LSM_VECTOR compaction

### LINKED ISSUE #5870: Vector index: renumber vector ids during the data file rewrite so the id space is dense by construction
Follow-up to #5588, which laid the `LSM_VECTOR` location index out in primitive arrays indexed by vector id.

## The residual weakness

The arrays are **chunked** (`CHUNK_SIZE = 128`) and a chunk is released as soon as its last live id is tombstoned. That is what keeps residency proportional to the LIVE vectors rather than to the ids handed out, which is the property #5516 exists for: a re-embedding workload hands out 9.3M ids for a 4K live set, and a flat id-indexed array would retain ~190MB where the structure it replaced retained ~360KB.

Chunking degrades well because ids are assigned monotonically and an update tombstones the id it supersedes, so the chunks behind the live region drain wholesale. What it does not solve is the **partially drained band trailing the live region**: a chunk still holding a single live id retains the whole `CHUNK_RETAINED_BYTES` (2680 bytes, so ~2.7 KB per live id in the worst case). 128 was chosen over 1024 precisely to bound that - it costs 4% on the dense case and buys 8x on the cliff - but it bounds the cliff rather than removing it.

## The structural fix

`LSMVectorIndex.rewriteDataFileWithLiveEntries` already walks the entire live set, writes every entry to a fresh file and re-assigns every absolute file offset. It is therefore the one moment in the index's life where reissuing the ids costs nothing extra: renumbering them `0..N-1` there would make the id space **dense by construction** after every compaction, and the partially drained band would disappear rather than being bounded.

`VectorLocationIndex`'s javadoc currently states that ids are preserved across a compaction rather than renumbered, and several places depend on that. This issue is to change it.

## Why it is not part of #5588

Ids are not private to the location index. Three things have to move in lockstep with a renumbering, and getting any of them wrong is a silent wrong-answer bug rather than a crash:

1. **`ordinalToVectorId`** - the graph's ordinal map, published alongside the graph by the same rebuild. The pairing of a map with its graph is what #4581 exists to keep intact.
2. **The persisted tombstone set** - on-disk entries carry vector ids, and `DeletedIds` is keyed by them. A renumbering that does not reissue or discard the tombstones lets an old tombstone id collide with a new live one, and the LSM merge-on-read resolves that to "deleted": the new vector is buried by a tombstone meant for a different one.
3. **`nextId`** - the high-water mark, which `loadVectorsFromPages` derives from `max(highest live id + 1, getNextId())` and which is the only thing preventing id reuse on an index where every vector was deleted.

## Acceptance

- After a compaction, `getMaxVectorId() == size() - 1` on an index with no live-set change across it.
- The re-embedding workload of #5516 leaves `chunkCount() == ceil(live / CHUNK_SIZE)` after a compaction, i.e. no partially drained band survives one.
- `Issue5588PrimitiveLocationIndexTest.aDrainedIdSpaceIsHandedBack` and the #5516 tombstone-memory guards still pass.
- A regression test that a tombstone written before a renumbering cannot suppress a vector that inherits its id afterwards.

robfrank:
Closes #5870

## Summary

`LSMVectorIndex.rewriteDataFileWithLiveEntries` (what a `COMPACT INDEX` on an `LSM_VECTOR` index actually runs) used to copy live entries into the new data file verbatim, keeping each vector's pre-compaction id. That bounds the "partially drained band" `VectorLocationIndex` pays for (issue #5588 - a chunk trailing the live region that holds a single live id retains a whole `CHUNK_RETAINED_BYTES`, ~2.7KB), but never removes it: a live vector whose id has drifted far ahead of the rest of the live set (a document re-embedded many times) still costs a whole straggler chunk after every compaction.

This PR renumbers every live entry densely to `[0, N)` during the same rewrite pass, in ascending old-id order:

- **Why this moment**: the rewrite already walks the entire live set and copies every entry to a fresh file, so reissuing ids costs nothing extra.
- **Why ascending old-id order**: it preserves the relative order the graph's ordinal map (`ordinalToVectorId`, issue #4581) depends on, and it guarantees `newId <= oldId` for every entry (N distinct non-negative integers sorted ascending are each `>= their rank`), so a renumbered entry is never bigger on disk than the one it replaces - the existing page-fits-in-current-page bookkeeping stays a safe upper bound.
- **On-disk re-encoding**: the vector id is the leading variable-length integer of the on-disk entry format (`LSMVectorIndexPageParser`), so a renumbering has to re-encode it - a verbatim byte copy would keep shipping the old id. The rest of the entry (RID, deleted flag, quantization payload) is copied as-is.
- **`nextId`**: `LSMVectorIndex.nextId` is reset to the dense count right after the rewrite, before the location index is republished, so future inserts continue the dense sequence instead of resuming from the pre-compaction high-water mark.

### Why reissuing ids is safe (the three things the issue calls out)

1. **`ordinalToVectorId`** - rebuilt immediately after the rewrite from the very same (now renumbered) entry objects, so it is never out of step with the ids the location index holds.
2. **The persisted tombstone set** - `publishLocationIndex` already rebuilds `VectorLocationIndex` fresh (a new instance, not an in-place refill) on every compaction, so its `DeletedIds` bitset starts empty regardless of renumbering. Both files a compaction replaces (the previous data file and the legacy compacted component) are dropped afterwards, so no on-disk tombstone for a discarded or reissued id survives either. A regression test (`aTombstoneWrittenBeforeRenumberingCannotSuppressAVectorThatInheritsItsId`) proves this: it deletes the documents holding the lowest vector ids (discovered through the location index, not assumed from insertion order), compacts, and asserts that the survivor now holding the reissued id 0 reads back live, not deleted.
3. **`nextId`** - see above; reset in lockstep with the renumbering, inside the same write-locked critical section the file swap runs in, so no concurrent writer can observe an inconsistent pairing.

## Changes

- `engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java`
  - `rewriteDataFileWithLiveEntries`: renumbers every live entry densely from 0, re-encoding the id varint per entry; resets `nextId` to the dense count.
  - `VectorEntryForGraphBuild.vectorId`: no longer `final`, so the renumbering can be applied in place and observed by every caller sharing the same `liveEntries` collection.
  - Updated javadoc on the rewrite method and on `publishLocationIndex`'s `nextId` carry-over.
- `engine/src/main/java/com/arcadedb/index/vector/VectorLocationIndex.java`
  - Updated class-level and `DeletedIds` javadoc: issue #5870 is no longer a forward reference, it is what happens.

No public API changes; no new dependencies.

## Test plan

New test class `engine/src/test/java/com/arcadedb/index/vector/Issue5870VectorIndexIdRenumberingTest.java`:

- `compactionMakesTheLiveIdSpaceDenseWithNoTrailingStraggler` - re-embeds one vertex hundreds of times (each in its own transaction, like a real workload) so its live id drifts into a straggler chunk far past the dense block the rest of the live set occupies; asserts the straggler chunk exists before compaction, and that after compaction `getMaxVectorId() == size() - 1` and `chunkCount() == ceil(live / CHUNK_SIZE)` exactly (both across the compaction and after a reopen).
- `aTombstoneWrittenBeforeRenumberingCannotSuppressAVectorThatInheritsItsId` - deletes the documents holding the lowest vector ids (resolved through the location index, not assumed from insertion order), compacts, and asserts the survivor that inherits the reissued id 0 is live, not deleted, resolves to the correct document, and is still found as its own nearest neighbor - across the compaction and after a reopen.

Both new tests were verified to **fail** against the pre-fix code (`getMaxVectorId()` stayed at the old high-water mark; the reissued id read back as not-live) before the fix, and pass after it.

Also ran and confirmed green against the fix:
- `Issue5588PrimitiveLocationIndexTest` (the `VectorLocationIndex` unit tests this issue's structural fix was deferred from)
- `LSMVectorIndexTombstoneMemoryTest` (the `#5516`/`#5588` tombstone-memory regression suite, including the id-sequence-survives-a-reopen test)
- `LSMVectorIndexCompactionTest` (existing compaction functional tests: space reclaimed, repeated compaction, deletes stay deleted, index identity follows the compacted file)
- The full `com.arcadedb.index.vector.*Test` package

- [x] `mvn -pl engine -am compile`
- [x] `mvn -pl engine test -Dtest=Issue5870VectorIndexIdRenumberingTest` (new tests, fail-before/pass-after verified)
- [x] `mvn -pl engine test -Dtest=Issue5588PrimitiveLocationIndexTest,LSMVectorIndexTombstoneMemoryTest,LSMVectorIndexCompactionTest`
- [x] `mvn -pl engine test -Dtest=com.arcadedb.index.vector.*Test` (full package)
- [ ] `RaftIndexCompactionReplicationIT` (ha-raft multi-node IT) was not run locally - out of scope for a quick verification pass, but the fix only changes the numeric id a live entry is encoded with, not the wire/file format itself, and a follower rebuilds by parsing the file the leader ships rather than re-deriving ids independently, so no follower-side change was needed.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -428,8 +428,11 @@ public class LSMVectorIndex implements Index, IndexInternal {
 
   /**
    * Rewrite the index data file keeping only the live entries, and swap it in: this is what a compaction of an
-   * LSM_VECTOR index is. Entries are copied verbatim (the quantized payload is never re-quantized) and keep their
-   * vector id, so no id is reused and the graph ordinals rebuilt right after stay valid.
+   * LSM_VECTOR index is. The quantized payload is never re-quantized, but the vector id IS reissued: every entry is
+   * renumbered densely from 0, in ascending old-id order, so the id space is dense by construction after every
+   * compaction instead of merely bounding the partially drained band trailing the live region (issue #5870). The
+   * graph ordinals rebuilt right after read the reassigned ids off the same {@code liveEntries} objects, so they
+   * stay valid without a second pass.
    * <p>
    * The whole rewrite runs under the index write lock. A compaction is an explicit maintenance operation, its cost
    * is one sequential pass over the live vectors, and blocking writers for it is the price of not having to
@@ -437,9 +440,13 @@ public class LSMVectorIndex implements Index, IndexInternal {
    * <p>
    * Both old files - the previous data file and the legacy compacted component, if the index still has one - are
    * dropped afterwards, leaving the index with a single data file. That is what actually returns the space to the
-   * filesystem: leaving the old file behind was why compaction used to make the index bigger, not smaller.
+   * filesystem: leaving the old file behind was why compaction used to make the index bigger, not smaller. It is
+   * also what makes the renumbering safe with respect to tombstones: every on-disk tombstone for an id being
+   * discarded or reissued lived only in one of these two files, and {@link #publishLocationIndex} discards the
+   * in-memory {@code DeletedIds} bitset the same way on every rebuild, so no stale "deleted" answer for an old id
+   * survives to be paired with the live vector that now holds the same (reissued) number.
    *
-   * @param liveEntries the live set, re-pointed in place at the new file
+   * @param liveEntries the live set, re-pointed and renumbered in place at the new file
    *
    * @return true when the file was actually rewritten and the location index re-published with it, false when the
    * rewrite was skipped
@@ -493,7 +500,12 @@ public class LSMVectorIndex implements Index, IndexInternal {
           throw new IllegalStateException(
               "Cannot compact vector index '" + indexName + "': cannot lock the new file " + newFileId);
 
-        // Copy the live entries, in vector id order, into freshly built pages of the new file.
+        // Copy the live entries, in vector id order, into freshly built pages of the new file. They are also
+        // RENUMBERED here, densely from 0: this is the one moment in the index's life where reissuing every id
+        // costs nothing extra, since the whole live set is already being walked and rewritten (issue #5870). Ids
+        // are handed out in ascending old-id order, which preserves the relative order #4581's ordinal map depends
+        // on and guarantees newId <= oldId for every entry (N distinct non-negative integers sorted ascending are
+        // each >= their rank), so a renumbered entry is never bigger on disk than the one it replaces.
         final List<VectorEntryForGraphBuild> sorted = new ArrayList<>(liveEntries);
         sorted.sort((a, b) -> Integer.compare(a.vectorId, b.vectorId));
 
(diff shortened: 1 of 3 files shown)


robfrank:
Thanks for the thorough review. Responses:

**Correctness points (newId<=oldId, encoding consistency, mutation/read ordering, locking, ordinalToVectorId sync)** - agreed, no changes needed.

**Test coverage gap (insert right after a renumbering compaction)** - added `insertingAfterARenumberingCompactionDoesNotCollideWithAnExistingId`. While writing it I found that a write issued *in the same session* immediately after `COMPACT INDEX` never reaches the vector index at all - reproduced independently of this PR by reverting the fix and hitting the same symptom on unmodified code. That's a separate, pre-existing bug, filed as #6105. The new test goes through a `reopenDatabase()` between the compaction and the insert to sidestep it, which is also the more realistic "compact, then keep operating" sequence and still directly pins the id-continuity/non-collision behavior you asked for.

**`ensureGraphAvailable()` reuse-path question** - good catch, and I don't think it's introduced by this PR (the staleness check is count-only, not content-only, and that gap predates #5870), but you're right that dense post-compaction ids make a same-count/different-composition collision more plausible than the old sparse ids did. I couldn't fully confirm the crash-timing window is reachable on a standalone database without a deeper trace of the page-flush/WAL guarantees around `graphFile.writeGraph`, so I filed it as #6106 with what I traced (including the `runWithCompactionReplication` standalone-vs-HA distinction) and added a comment at the staleness check pointing to it, rather than changing that delicate reuse logic under review-loop time pressure.

**Style/perf/security** - thanks, no changes needed there.

Pushed as `edf0d7063`.

robfrank:
Second review pass on \`edf0d7063\` found no blocking issues (correctness, replication-safety, and test coverage all confirmed independently). The remaining notes are explicitly non-blocking:

- Javadoc verbosity in a couple of spots - agreed it could be tighter, leaving as-is rather than re-triggering another review cycle for a pure style trim.
- An unrelated stale javadoc claim in \`RaftIndexCompactionReplicationIT\` ("not yet implemented") - out of scope for this PR, noted for a future cleanup.

No further changes planned from my side. Ready for maintainer review/merge whenever convenient - I'm not merging this myself.

PR #6110, opened 2026-08-12
fix(#5846): notify apply waiters after a snapshot install advances the index

### LINKED ISSUE #5846: Snapshot install advances the applied index without notifying waitForAppliedIndex(), causing spurious LINEARIZABLE 503s and READ_YOUR_WRITES timeouts
## Summary

A follower that catches up through a **snapshot install** advances its applied index without waking the threads blocked in `RaftHAServer.waitForAppliedIndex()`. They keep sleeping until `quorumTimeout` even though the index they are waiting for has already been reached.

The visible effects:

- `LINEARIZABLE` reads fail with `ReplicationException` -> HTTP 503, although the local state machine was in fact up to date.
- `READ_YOUR_WRITES` reads pay the full `quorumTimeout` and then log `consistency degraded to EVENTUAL`, a warning that misreports what happened.

Version: 26.8.1.

## Detail

The wait is a plain `wait/notify` on `applyNotifier`:

```java
// ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftHAServer.java:2121
public void waitForAppliedIndex(final long targetIndex, final boolean throwOnTimeout) {
  if (targetIndex <= 0)
    return;
  try {
    final long deadline = System.currentTimeMillis() + quorumTimeout;
    synchronized (applyNotifier) {
      while (getLastAppliedIndex() < targetIndex) {
        final long remaining = deadline - System.currentTimeMillis();
        if (remaining <= 0) {
          ...
        }
        applyNotifier.wait(remaining);
      }
    }
```

`notifyApplied()` has exactly one caller in the codebase, at the end of `ArcadeStateMachine.applyTransaction()`:

```java
// ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java:636
// Wake up any threads waiting for this index (READ_YOUR_WRITES, waitForLocalApply)
final RaftHAServer raftHA = this.raftHAServer;
if (raftHA != null) {
  raftHA.notifyApplied();
```

But `applyTransaction()` is not the only path that advances the applied index. `reinitialize()` — called by Ratis's `StateMachineUpdater.reload()` after a snapshot install, as the javadoc in that same file describes — seeds it directly and does not notify:

```java
// ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java:438
lastAppliedIndex.set(snapshotIndex);
...
updateLastAppliedTermIndex(snapshotInfo.getTerm(), snapshotIndex);
```

There is no `notifyApplied()` anywhere in `reinitialize()`.

The waiter's loop condition reads `RaftHAServer.getLastAppliedIndex()`, which comes from Ratis itself:

```java
// RaftHAServer.java:2191
return raftServer.getDivision(raftGroup.getGroupId()).getInfo().getLastAppliedIndex();
```

so the condition does become true after the snapshot is installed. The waiter simply is never told, and `wait(remaining)` only returns at the deadline.

## Failure scenario

1. A follower falls far enough behind that the leader ships a snapshot instead of log entries.
2. A client read arrives at that follower with `X-ArcadeDB-Read-After: N` (`READ_YOUR_WRITES`) or with `LINEARIZABLE`.
3. The handler calls `waitForAppliedIndex(N)`; the local applied index is still below `N`, so the thread waits.
4. The snapshot install completes and `reinitialize()` sets the applied index past `N`. No notification is sent.
5. The waiter sleeps for the remainder of `quorumTimeout`.
6. `LINEARIZABLE` then throws `ReplicationException` and the read fails with 503; `READ_YOUR_WRITES` logs a degradation warning and serves the read.

The window closes as soon as any subsequent entry is applied, since that calls `notifyApplied()` and the loop re-checks. So the bug bites hardest on a cluster that is quiet right after a catch-up — which is exactly when a snapshot install happens and when these waits are most likely to be pending.

## Suggested fix

Notify at the end of `reinitialize()` whenever the snapshot seeded a new index:

```java
lastAppliedIndex.set(snapshotIndex);
updateLastAppliedTermIndex(snapshotInfo.getTerm(), snapshotIndex);

final RaftHAServer raftHA = this.raftHAServer;
if (raftHA != null)
  raftHA.notifyApplied();
```

More robustly, since the waiter's condition is sourced from Ratis rather than from the state machine's own field, the two can drift apart for any Ratis-internal apply (configuration entries, leader no-op entries) that never reaches `applyTransaction`. A bounded re-check — `applyNotifier.wait(Math.min(remaining, someSmallInterval))` — would make the wait robust against every missed-notification path rather than just this one, at the cost of a periodic wake-up.

robfrank:
Closes #5846

## Summary

A follower catching up through a snapshot install advanced its applied index without waking threads
blocked in `RaftHAServer.waitForAppliedIndex()` / `waitForLocalApply()`. They kept sleeping until the
full `quorumTimeout` even though the index they were waiting for had already been reached:
`LINEARIZABLE` reads failed with a spurious `ReplicationException` (HTTP 503), and `READ_YOUR_WRITES`
reads paid the full timeout before logging a bogus "consistency degraded to EVENTUAL" warning.

`notifyApplied()` had exactly one call site, at the end of `ArcadeStateMachine.applyTransaction()`.
Two other paths advance `lastAppliedIndex` directly and never called it: `reinitialize()` (a follower
reloading after installing a snapshot - the issue's original finding) and
`notifyInstallSnapshotFromLeader()` (a follower receiving a snapshot directly from the leader - found
by @ruispereira in the issue comments; it does not route through `reinitialize()`). Crossed with the
two waiters that block on the same monitor, that's 2 waiters x 2 silent paths = 4 stuck combinations.

## Fix

Two complementary changes, both suggested in the issue discussion:

1. **Notify at both silent-advance sites.** `reinitialize()` and `notifyInstallSnapshotFromLeader()`
   now call `raftHAServer.notifyApplied()` (null-guarded for the not-yet-attached case) right after
   seeding `lastAppliedIndex` from a snapshot.
2. **Bounded re-check as defense in depth.** `waitForAppliedIndex()` and `waitForLocalApply()` cap
   each `applyNotifier.wait(...)` call at 1 second instead of the full remaining timeout, so the loop
   re-checks its condition on its own even if a *future* code path advances the index without calling
   `notifyApplied()`. This directly addresses the review comment that fixing only the known call sites
   is "one future `lastAppliedIndex.set(...)` away from regressing."

## Test plan

- [x] New unit test `ArcadeStateMachineReinitializeNotifiesTest`: drives `reinitialize()` directly
      against a real registered snapshot marker and a mocked `RaftHAServer`; asserts `notifyApplied()`
      fires when a snapshot is present, does not fire when there is none, and `reinitialize()` doesn't
      throw when `raftHAServer` isn't attached yet.
- [x] New unit test `ApplyWaitBoundedRecheckTest`: builds a `RaftHAServer` without starting Ratis,
      stubs the Raft division so the reported applied/commit index is test-controlled, advances it
      directly (bypassing `notifyApplied()` entirely - simulating an unfixed/future silent-advance
      path), and asserts both waiters return within ~2s instead of the full 10s `quorumTimeout`.
- [x] All four new test methods were verified to **fail** against the pre-fix code (reverted locally
      via `git stash`) before being confirmed to pass against the fix: the notify test failed with
      "Wanted but not invoked: raftHAServer.notifyApplied()", and both bounded-recheck tests failed
      with ~10s elapsed (the full quorum timeout) instead of the expected <2s.
- [x] `mvn -pl ha-raft -am compile` - clean.
- [x] `mvn -pl ha-raft -am test -Dtest=ApplyWaitBoundedRecheckTest,ArcadeStateMachineReinitializeNotifiesTest,WaitForApplyTest,ArcadeStateMachineSnapshotMarkerTest,ArcadeStateMachineTest` - 23/23 passed.
- [x] `mvn -pl ha-raft -am test -Dtest=ArcadeStateMachineAppliedIndexPerDatabaseTest,ArcadeStateMachineBootstrapMismatchTest,ArcadeStateMachineLifecycleTest` - 15/15 passed (adjacent `reinitialize()`/state-machine coverage, unaffected).
- [x] `WaitForApplyTest` (a real 2-node cluster IT exercising `waitForAppliedIndex`/`waitForLocalApply`,
      including the LINEARIZABLE-timeout-throws and READ_YOUR_WRITES-degrades regression guards) ran
      green against the fix.
- [ ] Broader full-cluster HA integration tests (e.g. `RaftReadConsistencyBookmarkIT`,
      `RaftFullSnapshotResyncIT`) could not be exercised locally in this session: the workstation's
      Homebrew-managed `arcadedb` service holds port 2480, which these ITs bind to by default (a
      pre-existing environment condition, unrelated to this change). Recommend CI runs the full HA IT
      suite on a clean port before merge.

## Scope notes

Both changes target exactly the silent-advance paths and waiters identified in the issue and its
comments. No behavior change to the normal `applyTransaction()` notify path. The bounded re-check adds
at most one extra `getLastAppliedIndex()`/`getCommitIndex()` call per second per *already-blocked*
waiter - negligible cost, only paid while a thread is parked waiting for replication to catch up.

### DIFF
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -442,6 +442,12 @@ public class ArcadeStateMachine extends BaseStateMachine {
       // snapshot.<inflatedTerm>_<index> filename persists across restarts until the next snapshot
       // rolls it over: cosmetic, expected.
       updateLastAppliedTermIndex(snapshotInfo.getTerm(), snapshotIndex);
+      // Wake any threads blocked in RaftHAServer.waitForAppliedIndex()/waitForLocalApply(): this seed
+      // can advance the applied index past a pending target (a follower catching up via snapshot
+      // install), and notifyApplied() has no other caller on this path (issue #5846).
+      final RaftHAServer raftHAOnReinit = this.raftHAServer;
+      if (raftHAOnReinit != null)
+        raftHAOnReinit.notifyApplied();
     } else
       lastAppliedIndex.set(-1);
 
@@ -1178,6 +1184,13 @@ public class ArcadeStateMachine extends BaseStateMachine {
         lastAppliedIndex.set(snapshotIndex);
         updateLastAppliedTermIndex(snapshotTerm, snapshotIndex);
         writePersistedAppliedIndexForAllDatabases(snapshotIndex);
+
+        // Wake any threads blocked in RaftHAServer.waitForAppliedIndex()/waitForLocalApply(): this
+        // leader-driven snapshot install advances the applied index without going through
+        // applyTransaction(), the only other notifyApplied() call site (issue #5846).
+        final RaftHAServer raftHAOnInstall = this.raftHAServer;
+        if (raftHAOnInstall != null)
+          raftHAOnInstall.notifyApplied();
 
         LogManager.instance().log(this, Level.INFO,
             "HA resync finished (mode=snapshot, result=ok): snapshotIndex=%d", snapshotIndex);
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -445,9 +445,9 @@ public class ArcadeStateMachine extends BaseStateMachine {
       // Wake any threads blocked in RaftHAServer.waitForAppliedIndex()/waitForLocalApply(): this seed
       // can advance the applied index past a pending target (a follower catching up via snapshot
       // install), and notifyApplied() has no other caller on this path (issue #5846).
-      final RaftHAServer raftHAOnReinit = this.raftHAServer;
-      if (raftHAOnReinit != null)
-        raftHAOnReinit.notifyApplied();
+      final RaftHAServer raftHA = this.raftHAServer;
+      if (raftHA != null)
+        raftHA.notifyApplied();
     } else
       lastAppliedIndex.set(-1);
 
@@ -1188,9 +1188,9 @@ public class ArcadeStateMachine extends BaseStateMachine {
         // Wake any threads blocked in RaftHAServer.waitForAppliedIndex()/waitForLocalApply(): this
         // leader-driven snapshot install advances the applied index without going through
         // applyTransaction(), the only other notifyApplied() call site (issue #5846).
-        final RaftHAServer raftHAOnInstall = this.raftHAServer;
-        if (raftHAOnInstall != null)
-          raftHAOnInstall.notifyApplied();
+        final RaftHAServer raftHA = this.raftHAServer;
+        if (raftHA != null)
+          raftHA.notifyApplied();
 
         LogManager.instance().log(this, Level.INFO,
             "HA resync finished (mode=snapshot, result=ok): snapshotIndex=%d", snapshotIndex);
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftHAServer.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftHAServer.java
@@ -185,6 +185,14 @@ public class RaftHAServer implements HealthMonitor.HealthTarget {
   private volatile PeerAddressAllowlistFilter allowlistFilter;
   private final    Object                    leaderChangeNotifier  = new Object();
   private final    Object                    applyNotifier         = new Object();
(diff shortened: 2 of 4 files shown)


robfrank:
Thanks for the review. The pre-existing `needsSnapshotDownload`/`lastAppliedIndex.set()` ordering concern is real and worth its own design pass rather than a quick patch here - filed as #6111 to track separately. No code changes needed in this PR for it.

The "no dedicated test for `notifyInstallSnapshotFromLeader()`" note is acknowledged as a reasonable tradeoff per the earlier review pass (that method needs leader HTTP address resolution + `DatabaseReconciler` HTTP machinery to exercise directly); `ApplyWaitBoundedRecheckTest` covers the missed-notify scenario generically regardless of which path misses it.

PR #6113, opened 2026-08-12
fix(#5855): re-enable plan cache for parameterized @rid filters

### LINKED ISSUE #5855: Follow-up (#5824): parameterized WHERE @rid = ? / IN ? no longer hits the plan cache
## Context

PR #5854 (fixing #5824) rewrites `WHERE @rid = <RID>` / `WHERE @rid IN [<RID list>]` on a type target into a direct `FetchFromRidsStep` fetch instead of a full type scan. Flagged during review: https://github.com/ArcadeData/arcadedb/pull/5854#issuecomment-5208380568

## The gap

Before PR #5854, `SELECT ... FROM Type WHERE @rid = ?` fell through to `FetchFromTypeWithFilterStep`, which is cacheable (`canBeCached() == true`) - it stores the `WhereClause` AST and re-evaluates it fresh against bound parameters on every execution, so repeated executions with different bound RIDs reuse the same cached plan.

After PR #5854, the same query now takes `FetchFromRidsStep`, which resolves the bound RID(s) at plan-construction time and bakes them into the step. `FetchFromRidsStep` doesn't override `canBeCached()`/`copy()`, so it defaults to `canBeCached() == false` (`ExecutionStepInternal.java`), making the whole plan non-cacheable. A "get by id" query executed repeatedly with different bound RIDs now re-plans from scratch every execution instead of hitting `db.getExecutionPlanCache()`.

This is almost certainly still a net perf win overall - an O(k) direct fetch plus a full replan is far cheaper than a cached plan reused to drive an O(n) scan - and it's not a new category of gap: every other `FetchFromRidsStep` call site (`FROM [#1:0, #1:1]`, a bind-parameter/variable RID target, a function-call target) is already non-cacheable today. But it does mean this specific rewrite trades away plan-cache reuse that previously existed for the `@rid = ?` / `@rid IN ?` shape.

## Suggested direction

Make `FetchFromRidsStep` (or a variant used by this rewrite) cacheable by storing a re-resolvable reference to the bind parameter/expression - mirroring how `FetchFromTypeWithFilterStep.copy()` re-attaches its `WhereClause` - rather than baking resolved RID values into the step at plan-build time. That would let parameterized RID lookups get both the O(k) fetch and cached-plan reuse.

Worth measuring first: confirm the replan cost is actually non-negligible relative to the O(k) fetch before spending effort here - it may not be worth the added complexity.

## Where to look

- `SelectExecutionPlanner.handleTypeWithRidFilter` / `extractRidEqualityOrInList` (introduced in PR #5854)
- `FetchFromRidsStep` (`engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromRidsStep.java`)
- `ExecutionStepInternal.canBeCached()` default and `SelectExecutionPlanner.createExecutionPlan`'s cache-put gate

robfrank:
Closes #5855

## Context

PR #5854 (fixing #5824) rewrites `WHERE @rid = <RID>` / `WHERE @rid IN [<RID list>]` on a type
target into a direct `FetchFromRidsStep` fetch instead of a full type scan. That step resolves the
bound RID(s) at plan-construction time and bakes them into the step. `FetchFromRidsStep` doesn't
override `canBeCached()`/`copy()`, so it defaults to `canBeCached() == false`
(`ExecutionStepInternal.java`), making the whole plan non-cacheable. A "get by id" query executed
repeatedly with different bound RIDs re-plans from scratch every execution instead of hitting
`db.getExecutionPlanCache()`.

Before PR #5854 the same query fell through to `FetchFromTypeWithFilterStep`, which is cacheable:
it stores the `WhereClause` AST and re-evaluates it fresh against bound parameters on every
execution.

## Fix

Mirror `FetchFromTypeWithFilterStep`'s pattern: instead of baking resolved `RID` values into
`FetchFromRidsStep` at plan-build time, add a variant that stores the original `BooleanExpression`
(the `@rid = ?` / `@rid IN ?` condition) and re-resolves it lazily against the live
`CommandContext` on the first pull of every execution, including a cache-hit re-execution with
different bound parameters. That step overrides `canBeCached()` to return `true` and `copy()` to
carry the (immutable) condition forward into the copy handed out by
`ExecutionPlanCache.get()`/`put()`.

The legacy `FetchFromRidsStep(Iterable<RID>, CommandContext)` constructor (used by every other
call site: `FROM [#1:0, #1:1]`, a bind-parameter/variable RID target, a function-call target,
`FetchEdgesFromVertexStep`'s sibling paths, etc.) is untouched and stays non-cacheable, matching
the issue's own scoping note that those call sites were already non-cacheable before PR #5854.

`SelectExecutionPlanner.handleTypeWithRidFilter` now always chains the deferred-resolution step
once the condition is confirmed *resolvable in shape* (`extractRidEqualityOrInList(...) != null`)
rather than short-circuiting to `EmptyStep` when the *value* the condition resolves to at this
particular build happens to be empty (e.g. an `IN`-list bound to `null` on the very first
execution). Baking that build-time "this execution's RIDs happen to be empty" fact into a cached,
reused plan would be wrong for a later execution bound to a non-empty list - `EmptyStep` itself
carries the comment "DON'T TOUCH! ... this execution plan cannot be cached" for exactly this
reason. Resolving to an empty RID set inside `FetchFromRidsStep.syncPull` produces the same
zero-row result without that hazard, on every execution.

`extractRidEqualityOrInList` moved from `private` to package-private static so
`FetchFromRidsStep` (same package) can call it during lazy resolution.

## A test-only race found and fixed along the way

Running the new tests as part of the wider `com.arcadedb.query.sql.**` suite (not the isolated
class) surfaced one flaky failure in `ridInListBoundParameterPlanIsCacheableAndReresolvesOnReuse`:
`assertThat(db.getExecutionPlanCache().contains(sql)).isTrue()` occasionally failed.

Root cause: `RidInScanOptimizationTest.beginTest()` creates types, which calls
`ExecutionPlanCache.invalidate()` (stamps `lastInvalidation` via `System.currentTimeMillis()`).
`SelectExecutionPlanner.createExecutionPlan`'s cache-put guard requires
`getLastInvalidation() < planningStart` before it will cache a freshly-built plan. On a hot/warmed
JVM, `beginTest()`'s schema setup and the test's first query can land in the same millisecond tick,
so a plan legitimately built *after* the invalidation gets skipped by the guard anyway - a
pre-existing, unrelated millisecond-resolution artifact in `SelectExecutionPlanner`/
`ExecutionPlanCache` (predates this PR). It only became reachable here because this PR is what
makes `@rid` plans cacheable in the first place, so this is the first test to exercise the guard
for this step type.

Not fixing the production guard (shared, out of scope, and the actual functional correctness of
query results is unaffected either way - only the caching optimization is occasionally skipped).
Instead, `beginTest()` now waits for the millisecond clock to tick forward before returning, so the
setup and the first query can never land in the same tick. Verified: re-ran the exact suite that
exposed the race (`mvn -pl engine test -Dtest='com.arcadedb.query.sql.**'`) after the fix - clean,
2485/2485, 0 failures.

## Test plan

- `planUsesFetchFromRidsStepForEquality` / `...ForBracketInList` / etc. (pre-existing #5824 tests):
  keep passing unchanged - the rewrite's plan shape and results are untouched.
- New: a bound-parameter `@rid = ?` plan reports `canBeCached() == true` and is actually present in
  `db.getExecutionPlanCache()` after one execution.
- New: the same query text executed twice with two *different* bound RIDs returns the record
  matching each execution's own bound value, not a stale RID baked in from the first (cached) plan
  build - this is the actual regression the issue describes.
- New: same for `@rid IN ?` bound to two different lists across two executions.
- New: an `IN ?` bound to an empty list on the first execution, then to a non-empty list on the
  second (same cached plan reused), returns zero rows the first time and the matching rows the
  second time - pins the `EmptyStep`-caching hazard called out above.
- Full `com.arcadedb.query.sql.**` suite (2485 tests) run clean after the fix.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromRidsStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/FetchFromRidsStep.java
@@ -22,32 +22,55 @@ import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.exception.TimeoutException;
+import com.arcadedb.query.sql.parser.BooleanExpression;
 
 import java.util.Iterator;
+import java.util.List;
 import java.util.NoSuchElementException;
 
 /**
  * Created by luigidellaquila on 22/07/16.
  */
 public class FetchFromRidsStep extends AbstractExecutionStep {
-  private final Iterable<RID> rids;
-  private       Iterator<RID> iterator;
-  private       Result        nextResult = null;
+  private       Iterable<RID>     rids;
+  // Set only by the cache-friendly constructor used by SelectExecutionPlanner.handleTypeWithRidFilter
+  // (WHERE @rid = ? / @rid IN ?, see issue #5855). When non-null, `rids` above is re-resolved lazily
+  // against the live CommandContext on the first pull of every execution instead of being baked in
+  // at plan-build time, mirroring how FetchFromTypeWithFilterStep re-evaluates its WhereClause. This
+  // is what makes the step - and therefore the whole plan - safely reusable from the execution plan
+  // cache across executions bound to different RIDs.
+  private final BooleanExpression ridCondition;
+  private       Iterator<RID>     iterator;
+  private       Result            nextResult = null;
 
   public FetchFromRidsStep(final Iterable<RID> rids, final CommandContext context) {
     super(context);
     this.rids = rids;
+    this.ridCondition = null;
     reset();
   }
 
+  FetchFromRidsStep(final BooleanExpression ridCondition, final CommandContext context) {
+    super(context);
+    this.ridCondition = ridCondition;
+  }
+
   public void reset() {
-    iterator = rids.iterator();
+    // With a ridCondition, defer re-resolution to the first pull of the (re)started execution, so it
+    // uses that execution's own live CommandContext/bound parameters rather than whichever context
+    // reset() happens to be called without.
+    iterator = ridCondition == null ? rids.iterator() : null;
     nextResult = null;
   }
 
   @Override
   public ResultSet syncPull(final CommandContext context, final int nRecords) throws TimeoutException {
     pullPrevious(context, nRecords);
+    if (ridCondition != null && iterator == null) {
+      final List<RID> resolved = SelectExecutionPlanner.extractRidEqualityOrInList(ridCondition, context);
+      this.rids = resolved == null ? List.of() : resolved;
+      iterator = this.rids.iterator();
+    }
     return new ResultSet() {
       int internalNext = 0;
 
@@ -106,6 +129,19 @@ public class FetchFromRidsStep extends AbstractExecutionStep {
 
   @Override
   public String prettyPrint(final int depth, final int indent) {
-    return ExecutionStepInternal.getIndent(depth, indent) + "+ FETCH FROM RIDs\n" + ExecutionStepInternal.getIndent(depth, indent) + "  " + rids;
+    return ExecutionStepInternal.getIndent(depth, indent) + "+ FETCH FROM RIDs\n" + ExecutionStepInternal.getIndent(depth, indent) + "  "
+        + (ridCondition != null ? ridCondition : rids);
+  }
+
+  @Override
+  public boolean canBeCached() {
+    return ridCondition != null;
+  }
+
+  @Override
+  public ExecutionStep copy(final CommandContext context) {
+    if (ridCondition == null)
+      throw new UnsupportedOperationException();
+    return new FetchFromRidsStep(ridCondition, context);
   }
 }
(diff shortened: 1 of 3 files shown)


robfrank:
Addressed the correctness bug: split \`extractRidEqualityOrInList\` (build-time shape check, unchanged) from a new \`resolveRidEqualityOrInListAtRuntime\` (used by \`FetchFromRidsStep.syncPull\`), which resolves whatever it can and skips unresolvable list elements instead of aborting the whole result to empty, and no longer swallows a right-hand-expression exception into "0 rows" at runtime.

Added \`ridInListWithOneUnresolvableElementStillMatchesTheResolvableOnesOnCacheReuse\`, reproducing your exact repro - confirmed it fails against the pre-fix code (empty result) and passes after the fix. Re-ran \`com.arcadedb.query.sql.**\` (2486 tests) clean.

The two minor notes (exception-propagation semantics, and text-keyed-cache-decided-by-first-execution as a known limitation) are addressed by the same change / left as documented limitations respectively, per your own framing of them as non-blocking.

robfrank:
Review is clean, remaining notes are all explicitly non-blocking (double-evaluation on cache-miss builds is pre-existing shape, not a regression; the documented first-bind tradeoff is an accepted scope boundary). Leaving this for merge review.

PR #6117, opened 2026-08-12
fix(#5871): remove SimpleNode's vestigial int constructor parameter

### LINKED ISSUE #5871: SimpleNode's int id parameter is now fully vestigial after #5868
## Background

#5868 (closing #5867) removed the legacy JavaCC/JJTree-generated SQL parser. Along the way it also deleted `SqlParserTreeConstants` (the last two named node-id markers, `JJTLIMIT`/`JJTTIMEOUT`) once it was confirmed that `SimpleNode(int i)`'s constructor has an empty body and never stores or reads `i`:

```java
public SimpleNode(final int i) {
}
```

Every remaining call site across the `com.arcadedb.query.sql.parser` package (`SQLQueryEngine`, `SQLScriptQueryEngine`, `SelectExecutionPlanner`, `SQLASTBuilder`, `Limit.copy()`, `Timeout.copy()`, `SelectStatement`, `MatchStatement`, etc.) now passes `-1` to `new Limit(...)` / `new Timeout(...)` / other `SimpleNode` subclasses, since that was already the dominant convention before #5868 and the value is provably inert.

## Suggested follow-up

Drop the `int` parameter entirely from `SimpleNode(int)` and give it (and every subclass constructor that only forwards to it - `Limit(int)`, `Timeout(int)`, and any other generated `Node` subclass in the package) a no-arg constructor instead, then update every call site to use it. This removes a parameter that exists purely as a leftover from the retired JJTree node-id scheme and carries no meaning today.

Flagged as a legitimate but explicitly out-of-scope suggestion during code review on #5868 (touches every `Node` subclass in `com.arcadedb.query.sql.parser`, so it's a separate, mechanical-but-wide change rather than a one-file fix).

## Scope note

Purely internal API surface (constructor signatures) inside `com.arcadedb.query.sql.parser` - no behavior change expected, verify with a full `mvn -pl engine clean compile test-compile` plus the existing parser test suite.

robfrank:
Closes #5871

## Context

`SimpleNode(int)` took a parser node id that was only ever meaningful to the JJTree-generated tree
walker removed by #5868. After that removal, the `int` parameter was dead: every call site passed
`-1` and the field it fed was never read. This PR removes the constructor, leaving only the no-arg
one, and mechanically converts every dependent constructor and call site to match.

## Change

- Removed `SimpleNode(int)`, leaving only the no-arg constructor.
- Converted 206 subclass/related constructors across `com.arcadedb.query.sql.parser` (200 in the
  parser package plus `BreakStatement`, `ExpressionStatement`, `Identifier`, `JsonArray`,
  `MoveVertexStatement`, `ParseCondition`, `ParseExpression`) from `ClassName(int id) { super(id); }`
  / `super(-1)` chains to no-arg constructors.
- Updated ~750 `new ClassName(-1)` call sites repo-wide, including 3 anonymous
  `new BooleanExpression(0){...}` sites.

## Two real bugs fixed as a necessary consequence

Removing the constructor is not purely mechanical - it surfaced two latent bugs that would have
thrown/misbehaved at runtime once the vestigial constructor was gone. Both are fixed here, not out
of scope creep but because the removal doesn't compile/pass tests without them:

1. **`MOVE VERTEX ... TO bucket:<number>` silently broken (NPE at runtime).**
   `SQLASTBuilder`'s bucket-target path constructed `Bucket` via the vestigial int constructor,
   which never actually set `bucketNumber`. Fixed by setting the field directly. Covered by a new
   regression test `moveVertexToBucketNumber` in `MoveVertexStatementExecutionTest`.

2. **Five `copy()` methods relied on reflecting into the vestigial constructor.**
   `MathExpression`, `FieldMatchPathItem`, `SelectStatement`, `CreateVertexStatement`, and
   `MatchPathItem` all called `getClass().getConstructor(Integer.TYPE).newInstance(-1)` to build
   their copy. This compiled fine but would throw `NoSuchMethodException` at runtime once the
   constructor was removed - only caught by actually running the test suite, not by a static
   compile check. Switched all five to `getClass().getConstructor().newInstance()`.

## Test plan

- `mvn -pl engine -am clean compile` / `test-compile`: pass.
- Full `-DskipTests -DskipITs clean install` across the whole reactor: `BUILD SUCCESS`, confirming
  no other module references the removed constructors.
- `com.arcadedb.query.sql.parser.**`, `com.arcadedb.query.sql.executor.**`, and
  `com.arcadedb.query.sql.antlr.**` (1875 tests, the packages touched by this change plus their
  direct dependents): 0 failures, 0 errors - including the reflection-based `copy()` paths this
  change touches, and the new `moveVertexToBucketNumber` regression test.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java
@@ -87,7 +87,7 @@ public class SQLQueryEngine implements QueryEngine {
     if (!statement.isIdempotent())
       throw new QueryNotIdempotentException("Query '" + query + "' is not idempotent");
 
-    statement.setLimit(new Limit(-1).setValue((int) database.getResultSetLimit()));
+    statement.setLimit(new Limit().setValue((int) database.getResultSetLimit()));
     return statement.execute(database, parameters);
   }
 
@@ -97,14 +97,14 @@ public class SQLQueryEngine implements QueryEngine {
     if (!statement.isIdempotent())
       throw new QueryNotIdempotentException("Query '" + query + "' is not idempotent");
 
-    statement.setLimit(new Limit(-1).setValue((int) database.getResultSetLimit()));
+    statement.setLimit(new Limit().setValue((int) database.getResultSetLimit()));
     return statement.execute(database, parameters);
   }
 
   @Override
   public ResultSet command(final String query, final ContextConfiguration configuration, final Map<String, Object> parameters) {
     final Statement statement = parse(query, database);
-    statement.setLimit(new Limit(-1).setValue((int) database.getResultSetLimit()));
+    statement.setLimit(new Limit().setValue((int) database.getResultSetLimit()));
 
     final CommandContext context = new BasicCommandContext();
     context.setInputParameters(parameters);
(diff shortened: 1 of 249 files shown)


robfrank:
Review is clean, no correctness/performance/security issues flagged. Addressed the one optional nit (`ValueExpression.copy()`) in the follow-up commit. Leaving this for merge review.

PR #6121, opened 2026-08-13
fix: CHECK DATABASE FIX destroyed the index of a document type

lvca:
## Problem

`CHECK DATABASE FIX` over a plain **DOCUMENT** type holding an unreadable record drops the type's index, never recreates it, and aborts the whole check:

```
IndexException: Error on creating index 'Doc_0_2285086433454166'
  caused by: DatabaseMetadataException: Cannot find record type '99'
```

The index is gone afterwards, and every repair the run had already made is lost with it. This is a single-node defect - no cluster involved.

`DatabaseChecker.checkDocuments` flagged a corrupted document through `addCorrupted` but, unlike the vertex and edge arms in `GraphDatabaseChecker`, never DELETED it. The shared tail of `check()` still put its bucket into `affectedBuckets` and dropped and rebuilt every index there, so the rebuild's own bucket scan met the same unreadable record and `LSMTreeIndex.build`'s error callback propagated it (deliberately - a swallowed failure would leave the index silently incomplete). Because the rebuild loop caught only `NeedRetryException`, that escaped `check()` with the attempt's `dropIndex()` already committed.

Every pre-existing FIX index test uses `CREATE VERTEX TYPE`, where the graph arm deletes the record first and the rebuild never meets it. That is why the document arm's omission survived.

## Fix

- `checkDocuments` and `checkScopedDocuments` delete corrupted documents under `FIX`, through a new `deleteCorruptedRecords`, matching the graph arms. Collected during the scan and deleted after it, since `LocalBucket.scan` is walking the pages the delete rewrites.
- That delete also does the `updateBucketRecordDelta(-1)` that `LocalBucket.deleteRecord` does not, so `count(*)` no longer depends on whether the run was full-scope (which happens to recompute in `checkBuckets`) or `RECORD`-scoped (which deliberately skips it).
- The rebuild loop gained a non-retryable catch arm. One index that cannot be rebuilt is now reported and struck from `rebuiltIndexes` instead of discarding the rest of the run - the same rationale the existing retry arm already states. This also contains the HA case, where an oversized index rebuild raises `ReplicatedEntryTooLargeException`.
- The failure warning names the owning `TypeIndex` instead of the internal `<bucket>_<nanos>` sub-index, which nobody typed and which made the `SELECT FROM schema:indexes` the message suggests return nothing.

## Tests

`CheckDatabaseFixDocumentTypeTest` - written first and confirmed red on the real defect before the fix:

- the type-wide arm;
- the `RECORD` scope, which reaches the same record by its own path;
- a rebuild that fails for a reason the checker cannot remove first. It gets its unremovable failure by inserting a duplicate straight into the bucket via `LocalBucket.createRecord`, which bypasses `DocumentIndexer` - the only way to hold a unique index and a violating record at once. It asserts the check completes, keeps the repairs already made, does not claim the index as rebuilt, and reports it as missing by name.

`RaftCheckDatabaseFix3NodesIT` is the first cluster coverage `CHECK DATABASE` has had. A grep of `ha-raft/src/test` for "CHECK DATABASE" previously returned nothing; the only file named for a checker is a `main()` that diffs three database directories offline. Three cases on a 3-node Raft cluster, with the corruption induced through the Raft wrapper so all three nodes carry it identically before the repair starts:

- a lost mid-chain edge segment is rebuilt on every node;
- the same FIX issued on a **follower** is forwarded to the leader and converges all three;
- a rebuilt index still exists and holds the rebuilt entry count on every node.

This one was written first, and writing it with a document type is what surfaced the defect above.

## Verification

- The 3 new document-type tests: red before the fix, green after.
- 136 tests across every suite that touches the FIX path, including `DatabaseCheckerIndexRebuildLockContentionTest` and `RebuildIndexLockExhaustionTest`, which exercise the retry arm this restructures.
- 562 tests across the `database`, `graph` and `engine` packages. The one failure, `Issue5279ConcurrentUpdateTest`, reproduces identically on a clean tree with these changes stashed.
- `RaftCheckDatabaseFix3NodesIT`, 3/3 green. Its mutation check - handing `DatabaseChecker` the inner instance in `CheckDatabaseStatement.createChecker`, the #5492 shape - fails all three with the right signatures: followers stuck at in-degree `-1`, and `Page PageId(graph/28/0) has different versions on databases. DB1 3 <> DB2 2` from the page comparator.

## Operator impact

Anyone with document types and suspected record corruption should not run `CHECK DATABASE FIX` on a build without this fix: it can destroy indexes. After a failed FIX, check `SELECT FROM schema:indexes` before reading the error.

## Follow-up commit: the same drift in the graph arms

`LocalBucket.deleteRecord` does not touch `cachedRecordCount`, and that counter - not a scan - is what `count(*)` and `countType()` answer from. The document arm above pays the matching `updateBucketRecordDelta(-1)`; `GraphDatabaseChecker` did not, at any of its three delete sites.

A type-wide run hides it, because `checkBuckets` recomputes every counter afterwards. The `RECORD` scope skips the database-wide passes, so `CHECK DATABASE RECORD <rid> FIX` on a corrupted vertex or edge left the type over-reporting the deleted record permanently:

| Scope | cached `countType` | `count(@rid)` scan |
| --- | --- | --- |
| RECORD | 2 | 1 |
| type-wide | 1 | 1 |

The vertex arm, the edge arm and the orphan-segment reclaim were byte-identical copies, so they now share a `deleteCorruptedRecord()` helper. `CheckDatabaseRecordScopeTest` gains a vertex case and an edge case, both asserting the cached counter against a real scan rather than a literal.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -530,8 +530,14 @@ public class DatabaseChecker {
   }
 
   /**
-   * Removes the records this pass flagged corrupted, counting them into {@code autoFix} exactly as
-   * {@code GraphDatabaseChecker}'s vertex and edge arms count theirs.
+   * Removes the records this pass flagged corrupted, counting them into {@code autoFix}.
+   * <p>
+   * NOT quite what {@code GraphDatabaseChecker}'s vertex and edge arms count, and the difference is deliberate:
+   * they call {@code autoFix.incrementAndGet()} BEFORE attempting the delete, so a record they then fail to remove
+   * is still reported as fixed. This one counts a repair only once the delete and its counter delta have both
+   * succeeded, and counts nothing for a record that was already gone. If the two are ever unified, unify them
+   * towards THIS one: {@code autoFix} is what an operator reads to decide whether the run did anything, and a
+   * number that includes failed deletes cannot answer that.
    * <p>
    * This is not merely tidiness, and it is not optional. A corrupted record's RID goes into
    * {@code corruptedRecords}, which is what {@link #check()} turns into {@code affectedBuckets}, and every index on
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -314,7 +314,8 @@ public class DatabaseChecker {
         // the larger LOCK_POST_DROP_ATTEMPT_MULTIPLIER budget instead of giving up at LOCK_MAX_ATTEMPTS.
         boolean rebuilt = false;
         boolean indexDropped = false;
-        NeedRetryException lastRetryFailure = null;
+        RuntimeException lastFailure = null;
+        boolean lastFailureRetryable = true;
         int attemptsUsed = 0;
         for (int attempt = 1; ; attempt++) {
           attemptsUsed = attempt;
@@ -344,7 +345,8 @@ public class DatabaseChecker {
             rebuilt = true;
             break;
           } catch (final NeedRetryException e) {
-            lastRetryFailure = e;
+            lastFailure = e;
+            lastFailureRetryable = true;
             if (droppedThisAttempt[0])
               indexDropped = true;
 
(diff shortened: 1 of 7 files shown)


lvca:
Good catch on the `count(*)` drift, and it was right. Fixed in `cdb384cad` rather than deferred: leaving it closed for documents and open for vertices one file away is worse than either fixing both or neither.

Verified before fixing, on a corrupted vertex:

| Scope | cached `countType` | `count(@rid)` scan |
| --- | --- | --- |
| RECORD | 2 | 1 |
| type-wide | 1 | 1 |

So the finding reproduces exactly, and the stated reason it hides is also right: the type-wide run is repaired as a side effect of `checkBuckets` recomputing every counter, and the RECORD scope skips that pass.

All three delete sites in `GraphDatabaseChecker` (vertex arm, edge arm, orphan-segment reclaim) were byte-identical copies, so they now share a `deleteCorruptedRecord()` helper that pays the `updateBucketRecordDelta(-1)` debt, rather than repeating the rule a fourth time. The reclaim's edge-list buckets belong to no type and have no user-facing counter reader today; they go through the helper anyway, since "which buckets have a reader" is not a distinction worth encoding per call site.

`CheckDatabaseRecordScopeTest` gains a vertex case and an edge case, both red before the fix for the right reason (cached 2, scan 1) and both asserting the counter against a real scan rather than a literal, so what is pinned is the two answers agreeing.

On the two non-blocking notes: agreed on both, and both are pre-existing rather than introduced here. The broad `catch (RuntimeException)` is a deliberate tradeoff (destroying the run is the worse failure, and nothing is swallowed - the message is quoted in the warning and the index is struck from `rebuiltIndexes`), and the `NeedRetryException`-on-commit path is unchanged by this PR. Leaving both as-is.

lvca:
Both observations checked out. Handled in `0c9a2b23d`:

**1. The `autoFix` parity claim.** Correct, and the wording was mine to fix. The graph arms call `autoFix.incrementAndGet()` before the `try`, so a record they then fail to delete is still counted as fixed; `deleteCorruptedRecords` counts only once the delete and its counter delta have both succeeded, and counts nothing for an already-gone record. The Javadoc now states the divergence rather than claiming parity, and says which way to unify it if anyone ever does: towards this one, since `autoFix` is what an operator reads to decide whether the run did anything, and a number that includes failed deletes cannot answer that.

**2. `DatabaseIsClosedException` reaching the catch-all.** Also correct. Leaving it as-is, deliberately: each failure is caught and reported, so the outcome is noise rather than damage, and special-casing it would mean the rebuild loop deciding which exceptions mean "the database is going away" - a judgement that goes stale the moment a new one appears. The noise is bounded by the number of affected indexes and every line names its own cause.

lvca:
Checked this against the code and ran the scenario. **The headline finding is not correct as stated, but a narrower version of it is real and is now fixed in `e00a66963`.**

### Correcting the record: a broken chain is not undeletable by CHECK DATABASE FIX

The review concludes the operator "is left with a permanently-undeletable corrupt record and a permanently-missing index". That is not what happens, because the analysis covers only the two arms this PR touched and misses a third pass. `LocalBucket.check(verboseLevel, fix)` - reached from `DatabaseChecker.checkBuckets` - already detects a broken chain and force-deletes it:

```java
final String chainProblem = findBrokenChunkChain(rid, page, recordPositionInPage);
if (chainProblem != null) {
  ...
  if (fix) {
    deleteRecordInternal(rid, true, true, true);   // force
    deletedRecordsAfterFix.add(rid);
```

`BrokenMultiPageRecordDeleteTest.checkDatabaseFixRemovesBrokenMultiPageRecord` has pinned exactly this since the force mode was added, and it passes on this branch. Measured on a real broken-chain vertex:

| Step | record still present? |
| --- | --- |
| after corruption | yes |
| after `CHECK DATABASE RECORD <rid> FIX` | **yes** |
| after `CHECK DATABASE FIX` | no |

`checkBuckets` also runs *before* the index rebuild, so the record is gone by the time the rebuild scans and the new non-retryable catch arm is never reached on this path.

### What was real: the RECORD scope

The RECORD scope deliberately skips the database-wide passes, `checkBuckets` among them, so the scoped run met the record only through the vertex arm and reported exactly the shape predicted:

```
"Cannot fix the record #1:0: error on delete
 (error: Multi-page record #1:0 chunk 0 was modified concurrently. Please retry)"
```

Nothing was permanently stuck, but an operator reaching for the narrow tool was told the repair failed with no hint the wide one would work. Fixed, with a regression test alongside the existing full-scope one.

### Not an unconditional force

The suggestion was to call `deleteRecord(rid, true)` directly. I did not, for two reasons. The shared helper is also used by the orphan-segment reclaim, which deletes *healthy* unreachable segments. And `LocalDatabase.deleteRecord` is explicit that a `ConcurrentModificationException` does not prove corruption - the same page-version validation fails when concurrent writes touch other records sharing the chain's pages, so on a busy bucket it is contention, and forcing there orphans chunks and skips index cleanup for a healthy record.

So the new `LocalBucket.deleteCorruptedRecord` escalates only when `isChunkChainBroken` - the version-blind structural probe, the same discriminator the tolerant delete path uses - confirms real damage. Contention still propagates as the retry signal. `DELETE_TOLERATE_BROKEN_CHAIN` is deliberately not consulted, since its own documentation says CHECK DATABASE FIX is unaffected by it either way.

### A reporting bug this surfaced

Deleting the record in the arms broke the existing `deletedRecordsAfterFix` assertion, which turned out to be a real inconsistency rather than a stale test: the bucket-wide pass listed what it deleted, while the arms deleted silently and reported only an `autoFix` count. Invisible while a broken-chain record was always removed by the bucket pass; once the arms could remove it first, the same repair stopped listing the same record. Every deleting pass now reports what it removed. The orphan-segment reclaim is excluded on purpose: it keeps its own counter and its segments are internal and unbounded in number.

The `AtomicReference` leftover is dropped too. On the broad `catch`, unchanged, for the reason given earlier.

Verified: 1646 engine tests across `database`/`graph`/`engine`/`index` with one unrelated pre-existing failure (`Issue5279ConcurrentUpdateTest`, reproduces on a clean tree), plus `RaftCheckDatabaseFix3NodesIT` 3/3.

Issue #6115, opened 2026-08-12
Backup phase 3: incremental backup on the page-version manifest, then a changed-page bitmap, then WAL archiving for PITR


lvca:
Phase 3 of the backup roadmap in `docs/optimize-backup.md`. **Unblocked by #6075** (merged as #6100), which was its stated prerequisite.

## Why #6075 had to land first

Phase 3a ships only the pages whose version differs from the parent backup's manifest. That is sound only if, for every page, **the image written into the archive and the version recorded in the manifest are the same image**. With the old fuzzy copy that pairing broke silently:

1. The backup copies page P at t1, when P is at v5.
2. The manifest records P's version at t2, when P has advanced to v7.
3. The archive holds P at v5 while claiming v7.
4. The next incremental compares live P (still v7) against the manifest's v7, concludes "unchanged", ships nothing.
5. Restoring full plus incremental yields P at v5. Silent, permanent data loss.

`PageSnapshot` makes the pairing exact by construction: the image and its version header are read together, from the same snapshot, so the manifest cannot disagree with the archive. That property is now available and tested.

## Scope

**3a - page-version manifest.** The full backup writes a manifest of `(fileId, pageNumber) -> version` taken from the snapshot it archived. An incremental opens its own window, diffs against the parent manifest, and ships only the differing pages plus its own manifest. Restore replays the chain.

**3b - changed-page bitmap.** The manifest diff is O(total pages); a bitmap maintained by the same `concurrentPageAccess` write hook #6075 already installed makes it O(changed pages). The bitmap must be cleared atomically at a defined snapshot point - which, again, only exists now: with no single t0 a page dirtied inside a fuzzy window can fall out of both bitmap generations.

**3c - WAL archiving for PITR.** Retain and ship WAL segments covering the interval after a full backup so a restore can roll forward to an arbitrary point rather than only to a backup boundary.

## The correctness trap to design around from the start

**LSM compaction recycles file ids with fresh, low page versions.** A compacted file can legitimately present page versions *lower* than the parent manifest recorded for the same `(fileId, pageNumber)`. A naive "version differs -> ship, version equal -> skip" comparison will silently skip live data after a compaction. The manifest needs a per-file generation marker (or the compaction migration map already tracked in `LocalSchema.getMigratedFileId`) so a recycled file id is treated as wholly new rather than diffed.

## Notes

- Chain integrity: an incremental's parent must be a well-defined state. It now is.
[...]

Issue #6087, opened 2026-08-11
Backup phase-1 follow-ups: copy() drops settings on Export/Import, deferred-RAM metric unwired, no per-database compression override


lvca:
## Context

Three small follow-ups found while doing backup phase 1 (#6072, merged in #6080). Grouped here because each is a few hours of work on its own and none needs a design discussion - filing separately would be more overhead than signal.

## 1. `ExportDatabaseStatement.copy()` and `ImportDatabaseStatement.copy()` drop their settings map

`BackupDatabaseStatement` had this exact bug and it is fixed (#6080): `copy()` only copied `url`, silently dropping the `settings` map a `WITH ...` clause populates, so a copy taken from the statement cache lost every setting.

`ExportDatabaseStatement.copy()` (`engine/src/main/java/com/arcadedb/query/sql/parser/ExportDatabaseStatement.java:134`) and `ImportDatabaseStatement.copy()` (`engine/src/main/java/com/arcadedb/query/sql/parser/ImportDatabaseStatement.java:150`) have the identical shape:

```java
public Statement copy() {
  final ExportDatabaseStatement result = new ExportDatabaseStatement(-1);
  result.url = this.url;
  return result;
}
```

**Checked before filing: this is latent, not live.** Whole-statement `copy()` for an admin statement like `EXPORT DATABASE` / `IMPORT DATABASE` is not currently on the execution path - the planners that call `.copy()` do so on sub-expressions (`WHERE` clauses, projections), not on top-level admin statements, and neither of these two goes through the statement cache the way a query does. So today, no `WITH ...` setting is actually being dropped in production. It is worth fixing anyway because it is the same defect class as the BACKUP one, in the same file family, and it stops being latent the moment either statement's execution path changes.

**Fix:** mirror the one applied to `BackupDatabaseStatement`:

```java
result.settings.putAll(this.settings);
```

**Scope:** both classes, plus a regression test per class in the shape of `BackupDatabaseStatementTestParserTest` - build a statement with a `WITH` setting, `copy()` it, assert the copy's `settings` map is non-empty and equal to the original's.

## 2. `PageManager.getDeferredRAMBytes()` is not wired into metrics

[...]

PR #6122, opened 2026-08-13
feat(#6116): snapshot metrics, the missing backup benchmarks, and the HA /checksums endpoint off the flush suspension

### LINKED ISSUE #6116: Backup phase-2b follow-ups: the benchmarks #6075 asked for, snapshot metrics, and the HA /checksums endpoint still suspending flush
Loose ends left behind by #6075 (merged as #6100), listed honestly rather than quietly dropped. None is a correctness gap; all were called out in that PR.

## 1. The benchmarks #6075 asked for and did not get

#6075 listed a benchmark set as required. #6100 produced exactly one of them - the suspension-fraction measurement in `Issue6075SnapshotBackupIT` (>90% on the fallback path vs <50% with a snapshot, control measured in the same run) - and shipped without the rest:

- **Steady-state write throughput and commit latency (p50/p95/p99) with no window open**, before and after. This is the claim the whole "zero cost when idle" design rests on, and it is currently defended by reading the code (one volatile read plus a predicted branch inside a critical section that already does a 64 KB `pwrite`) rather than by measurement.
- **Write throughput and commit latency with a window open**, against the suspend-and-freeze path, on the #6072 harness that produced the 4.3% / 77% figures.
- **Deferred-RAM high-water mark** during a backup: it should now never reach `FLUSH_SUSPEND_MAX_DEFERRED_RAM` at all.
- **Shadow peak size as a fraction of database size** across write rates - the sizing evidence behind the 1 GB `pageSnapshotMaxSize` default, which is currently a guess.
- **t0 barrier duration at various flush-queue depths.** It is the one remaining stall and its size should be known.
- **Flush-thread drain rate** with and without a window open (challenge C5).
- **HA replay throughput** with the `TransactionManager` apply lock: one uncontended `ReentrantReadWriteLock` read acquisition per *replicated transaction* (never on the local commit path), reasoned to be unmeasurable against the WAL parse and page writes it sits next to, but not actually measured.

Tag them `@Tag("benchmark")` per CLAUDE.md.

## 2. No metrics for an open snapshot window

`PageManager` exposes `isSnapshotWindowOpen`, and `PageSnapshot` exposes `getShadowedPages()`, `getShadowSizeInBytes()`, `getShadowSpilledBytes()` and `getOpenedOn()`, but none of it is wired into `EngineMetricsBinder` or Studio. An operator cannot currently see that a window is open, how large its shadow has grown, or how close it is to `pageSnapshotMaxSize` before it breaches and forces the fallback. This is the natural companion to the unwired `getDeferredRAMBytes()` gauge already tracked in #6087.

Watch the counter/gauge distinction #5636 documents: shadowed pages and shadow bytes are per-window gauges, not monotonic counters.

## 3. `SnapshotHttpHandler`'s `/checksums` sub-path still suspends flushing

`GET /api/v1/ha/snapshot/{db}/checksums` was deliberately left on `suspendFlushAndExecute` in #6100. It reads the database directory raw through `SnapshotManager.computeFileChecksums(dbDir)` rather than the registered page files, so it needs a different shape from the verify handler's migration - it would have to either take the file list from the snapshot or gain a directory-oriented reader. It is a rarely called diagnostic, but it is the last writer-throttling path left in the product and it should not stay that way by accident.

## 4. Extract the snapshot registry out of `PageManager`

`PageManager` gained `activeSnapshots`, `deferredFileDrops`, `snapshotBarrierLocks`, `openSnapshot`, `deferFileDrop`, `releaseDeferredFileDrop` and `capturePreImages`. That is a lot of surface on an already large class. Phase 3 (#6115) will add manifest and changed-page-bitmap state that belongs with it, which is the right moment to move the lot into a `PageManager`-owned registry rather than doing it speculatively now.

lvca:
Closes #6116. The three actionable loose ends left behind by #6075 / PR #6100.

## 1. The benchmarks #6075 asked for

Two harnesses, both `@Tag("benchmark")` so neither runs in CI:

- **`engine/src/test/java/performance/PageSnapshotOverheadBenchmark`** - steady-state write throughput and commit latency p50/p95/p99 in the three states the write path can be in, the t0 barrier under increasing write pressure, the flush-thread drain rate with and without a window, and the `TransactionManager` apply lock.
- **`integration/.../PageSnapshotBackupBenchmark`** - what a running backup does to the writers beside it on both paths, the deferred-RAM high-water mark, and the shadow peak as a fraction of the database. Deliberately shaped like `BackupCompressionBenchmark.concurrentWriterImpact()` (the #6072 harness) so the two sets of numbers can be read side by side, and it samples the new gauges rather than the window object, because a backup owns its window internally - so the benchmark also exercises the wiring of part 2.

**On "zero cost when no window is open".** There is no honest "before" to measure from inside the merged tree, so it is bounded from above instead: the expensive shape of the hook is *a window open on a **different** database*, which pays the volatile read, the array walk and a database comparison per page write and captures nothing. Indistinguishable from the no-window case ⇒ the no-window case, strictly cheaper, cannot be costing anything either. Two rounds of each configuration are printed so the reader can see the machine's own noise instead of reading it as signal.

### Two numbers that correct what #6100 asserted

- **The t0 barrier is not "a few ms" under load.** Idle: 13-120 us. With 1-8 writer threads committing continuously: tens of milliseconds, occasionally past 100 ms. The driver is *not* the flush-queue depth (which stays small) but the barrier **retrying** - it drains the queue in full, and a commit landing between that drain and the flush-thread suspension makes it start over. Still bounded, still the only stall, but now measured.
- **The shadow can reach the size of the database.** Throttled backup of a 128 MB database: shadow peaked at 37% of it with a writer pausing 100 ms between transactions, 84% at 20 ms, **100%** flat out. So the 1 GB `pageSnapshotMaxSize` default is not the generous ceiling it looks like - past a gigabyte under heavy writes, expect the breach and the fallback. The new usage-percent gauge is how an operator sees which way it is going.

### The headline comparison

128 MB database rebuilt before every run (the load inserts as well as updates, so a shared database would make later runs back up more data and divide the shadow fraction by a size that no longer exists), backup throttled to 32 MB/s, counting only commits inside the backup window:

| | rec/s in the window | of baseline | peak deferred RAM |
|---|---|---|---|
| no backup running | 193,854 | - | 0 |
| backup, snapshot path | 185,120 | **95.5%** | **0** |
| backup, suspend-and-freeze | 5,274 | 2.7% | 513 MB |

513 MB is `flushSuspendMaxDeferredRAM` reached: the point at which committers stop being slowed and start being throttled outright.

## 2. Metrics for an open snapshot window

`PageManager.getStats()` gained the readings, so they reach Prometheus (`EngineMetricsBinder`) and Studio's server page (`Profiler`, which renders every stat it publishes - no JS change needed). Counter/gauge per #5636: per-window state is gauges, only the JVM-wide totals are counters.

`snapshot.windows.open`, `snapshot.shadow.pages`, `snapshot.shadow.bytes`, `snapshot.shadow.spilled.bytes`, `snapshot.shadow.usage.percent`, `snapshot.window.age.ms` (gauges); `snapshot.windows.opened`, `snapshot.windows.invalidated`, `snapshot.preimages.captured` (counters). Plus `flush.deferred.bytes`, the unwired `getDeferredRAMBytes()` gauge of **#6087** - the same question asked of the other path, and one line away.

The alertable pair is the usage percentage and the invalidation counter: an invalidated window is invisible from outside (its consumer falls back and still completes), so before this a backup that had gone back to throttling every writer on the server left no trace but a WARNING.

`SnapshotCache.read()` also learned to unwrap `perc`, as a **double**: against the 1 GB cap a filling shadow is a fraction of a percent, and truncating that to a long reports a comfortable zero.

## 3. `/checksums` no longer suspends flushing

The last writer-throttling reader in the product. It reads the database *directory* rather than the registered page files, hence a directory-oriented reader (`SnapshotManager.computeFileChecksums(dir, snapshot, pageFileNames)`) rather than the verify handler's file-list one: page files come from the window, everything the window does not cover (`database.json`, `schema.json`, `.ts.sealed`, `last-tx-id.bin`) is read raw under the same database read lock as before. A page file created **after** t0 is skipped rather than read live - it has no point-in-time content, and a torn CRC in a map whose purpose is cross-node comparison is worse than an absent one. Same `PageSnapshotException` fallback as every other consumer.

Two things fixed on the way:
- the `.pshadow` scratch of an open window lives in the database directory and was being checksummed, so any node that happened to have a backup running reported a file no peer had;
- `suspendFlushAndExecute` logs and swallows, so a failure on the fallback path would have returned an empty map with a 200 - which a caller comparing checksums reads as "every file matches". The failure is now carried out and rethrown, and the handler still answers 500.

## 4. Extracting the snapshot registry out of `PageManager` - deliberately not here

Agreeing with the issue's own reading: #6115 (phase 3) adds the manifest and changed-page-bitmap state that belongs in that registry, and its lock-order contract is documented once in `PageManager`'s Javadoc. Moving it now produces a rename-only diff that #6115 immediately reshapes, and would put a second dereference in front of the one field the write path is built around. The right moment is #6115.

## Verification

New tests, each with the control that keeps it from passing vacuously:

- `engine/.../PageSnapshotMetricsTest` (4): the gauges rise with a window and return exactly where they were when it closes; the shadow gauges track the copy-on-write work **and the identical rewrite with no window open moves the pre-image counter by zero**; a cap breach increments the invalidation counter once, not once per failed capture; the profiler JSON and the metrics dump both carry the readings.
- `ha-raft/.../Issue6116ChecksumsPointInTimeTest` (4): the snapshot checksums are byte-for-byte the frozen-file ones; they stay at t0 through a full rewrite **while a live read of the same directory is asserted to have changed**; the suspended fraction of the computation is >0.9 on the fallback path and <0.5 on the snapshot path, both measured in the same test; both paths answer with the same file set and the same values.
- `SnapshotManagerTest`: transient files - `.wal`, `.prev.json`, `.lock`, `.corrupt` and now `.pshadow` - are not checksummed.
- `EngineMetricsBinderTest`: the new meters are registered with the right instrument type, and the gauges are asserted to *follow a real window* (delta-based, since `PageManager` is a JVM-wide singleton), including the sub-1% usage reading that a long would have flattened.

Re-run green: `com.arcadedb.engine.*Test` (183), ha-raft snapshot + state-machine tests (181), `com.arcadedb.server.monitor.*` (56), integration backup unit tests (12) and ITs (33, including `Issue6075SnapshotBackupIT` and `BackupCompressionIT`). Both benchmarks were run manually to produce the numbers above.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -921,3 +921,85 @@ source file's modification time - which is the more useful of the two, and was a
 archive written with `arcadedb.backup.compressionThreads = 0`.
 
 [#6075](https://github.com/ArcadeData/arcadedb/issues/6075)
+
+## The snapshot backup, measured; an open window is now visible in the metrics; the HA `/checksums` endpoint stops freezing the files (#6116)
+
+Three loose ends left behind by the point-in-time snapshot of #6075.
+
+### The benchmarks #6075 asked for
+
+`Issue6075SnapshotBackupIT` measured the suspension fraction and nothing else. Two harnesses now cover the rest:
+`PageSnapshotOverheadBenchmark` (engine) for the write path, the t0 barrier, the flush-thread drain rate and the
+`TransactionManager` apply lock, and `PageSnapshotBackupBenchmark` (integration) for what a running backup does to
+the writers beside it, shaped like the #6072 harness so the two sets of numbers can be read together. Both are
+`@Tag("benchmark")`, so neither runs in CI.
+
+Two of the numbers they produce are worth stating here, because both correct something the #6075 write-up asserted:
+
+- **The t0 barrier is not "a few ms" under load.** With no concurrent writer it is 13-120 us. With one to eight
+  writer threads committing continuously it runs to tens of milliseconds, occasionally past 100 ms, and the driver
+  is not the flush-queue depth (which stays small) but the barrier RETRYING: it drains the queue in full, and a
+  commit landing between that drain and the flush-thread suspension makes it start over. It is still bounded, still
+  vastly better than a suspension held for the whole backup, and still the only stall in the design - but its size
+  is now known rather than assumed.
+- **The copy-on-write shadow can reach the size of the database.** During a throttled backup of a 128 MB database,
+  the shadow peaked at 37% of it with a writer pausing 100 ms between transactions, 84% at 20 ms, and 100% with a
+  writer running flat out - it holds the pre-image of every page dirtied while the window is open, so a hot
+  database rewrites all of them. The 1 GB `arcadedb.pageSnapshotMaxSize` default is therefore not the generous
+  ceiling it looks like: on a database much beyond a gigabyte under a heavy write load, expect the window to breach
+  it and the backup to restart on the suspend-and-freeze path. Raise it (or accept the fallback) accordingly - and
+  the new `arcadedb_engine_snapshot_shadow_usage_percent` gauge below is how to tell which of the two you are
+  heading for.
+
+The writer-impact comparison itself, on a 128 MB database rebuilt before each run, with a mixed insert/update load
+and the backup throttled to 32 MB/s, counting only the commits that fall inside the backup window:
+
+| | rec/s inside the window | of the no-backup baseline | peak deferred RAM |
+|---|---|---|---|
+| no backup running | 193,854 | - | 0 |
+| backup, snapshot path | 185,120 | **95.5%** | **0** |
+| backup, suspend-and-freeze | 5,274 | 2.7% | 513 MB |
+
+513 MB is `arcadedb.flushSuspendMaxDeferredRAM` reached, which is the point at which committing threads stop being
+merely slowed and start being throttled outright. On the snapshot path the deferred-RAM gauge never leaves zero -
+the claim #6075 rested on, stated as a measurement rather than as a design argument.
+
+### An open window is now visible
+
+`PageManager.getStats()` gained the snapshot readings, so they reach both Prometheus (through
+`EngineMetricsBinder`) and Studio's server page (through `Profiler`, which renders every stat it publishes). Per
+the counter/gauge distinction of #5636, the per-window state is gauges and only the JVM-wide totals are counters:
+
+| Metric | Type | |
+|---|---|---|
+| `arcadedb_engine_snapshot_windows_open` | gauge | windows open right now |
+| `arcadedb_engine_snapshot_shadow_pages` | gauge | pages held in their shadows |
+| `arcadedb_engine_snapshot_shadow_bytes` | gauge | bytes held, RAM plus spill |
+| `arcadedb_engine_snapshot_shadow_spilled_bytes` | gauge | of which spilled to disk |
+| `arcadedb_engine_snapshot_shadow_usage_percent` | gauge | the fullest window against `pageSnapshotMaxSize` |
+| `arcadedb_engine_snapshot_window_age_ms` | gauge | age of the oldest open window |
+| `arcadedb_engine_snapshot_windows_opened_total` | counter | windows opened |
+| `arcadedb_engine_snapshot_windows_invalidated_total` | counter | windows that lost their point in time |
+| `arcadedb_engine_snapshot_preimages_captured_total` | counter | pre-images copied |
+| `arcadedb_engine_flush_deferred_bytes` | gauge | dirty bytes deferred by a flush suspension (#6087) |
+
+The alertable ones are the usage percentage and the invalidation counter. A window that breaches its cap is
+invalidated and its consumer falls back to the suspend-and-freeze path, which still completes - so before this, a
+backup that had quietly gone back to throttling every writer on the server left no trace except a WARNING in the
+log.
+
+### `GET /api/v1/ha/snapshot/{db}/checksums` no longer suspends flushing
+
+The last writer-throttling reader in the product. #6075 migrated the backup, the HA verify and the HA snapshot
+ship but left this one on `suspendFlushAndExecute`, because it reads the database DIRECTORY rather than the
+registered page files and so needed a different shape. It now takes each page file's content from a window and
+reads everything the window does not cover - `database.json`, `schema.json`, the `.ts.sealed` time-series stores -
+raw, as before, under the same database read lock. A page file created after t0 is left out rather than read live,
+matching what the verify endpoint reports; the checksums are byte-for-byte the ones a peer still on the fallback
+path computes, so a migrated leader does not report every file as differing.
+
+One transient file was being checksummed that should not have been: the `.pshadow` scratch of an open snapshot
+window lives in the database directory, so any node that happened to have a backup running reported a file no peer
+had.
+
+[#6116](https://github.com/ArcadeData/arcadedb/issues/6116)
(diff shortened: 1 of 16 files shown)


lvca:
Both review points addressed in 74d555f.

**Error detail lost on the /checksums failure paths** - correct, and worth fixing. The wrapper is `RWLockContext.executeInReadLock`, which rewraps anything that is not a `RuntimeException` into `ArcadeDBException("Error in execution in lock", cause)`, so a genuine "permission denied on file X" was being answered with a sentence about a lock. The handler now reports the deepest cause's message (`rootCauseMessage`), which restores what the pre-migration code returned when its try/catch sat inside the `suspendFlushAndExecute` callback. Two unit tests, including the null-message case.

**The "nit" about `registeredPageFileNames` is a real race, not a moot one.** File registration absolutely can happen while a reader holds the database read lock: index compaction creates and drops component files *without* the write lock - that is exactly why `PageManager.deferFileDrop` had to exist in #6100. So a file registered between the name capture and the directory listing would be absent from the window, absent from the captured set, and CRC'd live - the torn-checksum case the skip rule is there to prevent.

Rather than tighten the capture ordering (which narrows the window without closing it), the dependence on the registry is gone: whether a directory entry is a page file is now decided from its extension, via a new `LocalDatabase.isComponentFileName()` applying the same rule `FileManager.scanDirectoryForComponentFiles` uses to decide what to register. That cannot go stale, and `computeFileChecksums` loses its third parameter. New test: a bucket file appearing in the directory after t0 is left out of the map while a non-page file added at the same moment is still read, with a no-window control asserting the file is genuinely there and readable.

ha-raft suite re-run green (45 tests: `Issue6116ChecksumsPointInTimeTest`, `SnapshotManagerTest`, all `SnapshotHttpHandler*`, `PostVerifyDatabaseHandler*`).

lvca:
Thanks - taking the three nits in turn, none of them need a code change, but the first one deserved verifying rather than asserting:

**1. Fallback only on `PageSnapshotException`.** Intentional, and complete by construction: `PageManager.openSnapshot()` funnels *everything* into that type -

```java
} catch (final PageSnapshotException e) {
  throw e;
} catch (final InterruptedException e) {
  Thread.currentThread().interrupt();
  throw new PageSnapshotException("Interrupted while opening a snapshot of database '" + ... );
} catch (final Exception e) {
  throw new PageSnapshotException("Cannot open a snapshot of database '" + ... );
}
```

so no failure to *open* a window can slip past the fallback, whatever its origin (its Javadoc states this explicitly: "Every failure surfaces as PageSnapshotException, including an I/O error while enumerating the files"). What does bypass the fallback is an `IOException` from `computeFileChecksums` itself - a genuine disk read error part-way through the pass - and that is deliberate: re-reading the same files with the flush suspended would fail at the same byte, so failing fast with the real message is more useful than doing the work twice. A cap breach *during* the read is different and does fall back, because it arrives as a `PageSnapshotException` from `SnapshotFileInputStream.fill()`'s `checkValid()`. Same split `PostVerifyDatabaseHandler` uses.

**2. `snapshotShadowUsagePerc` as a max.** Agreed on the trade-off, and it is why `arcadedb_engine_snapshot_windows_open` is exported alongside: two windows each at 60% show as `usage=60, windows=2`, so an alert can be written on the pair. A per-window breakdown would need a `MultiGauge` with a window-id tag, which is unbounded-cardinality churn for a signal that only matters while a backup is running - not worth it here.

**3. `MAX_CAUSE_DEPTH` boundary test.** Agreed, it is a pure backstop; the cycle test covers the behaviour that matters.

Leaving all three as they are. Waiting on `build-and-package` and then this is done from my side.

Issue #6096, opened 2026-08-12
RESTORE DOCUMENT in the same transaction as the DELETE: count(*) reports 1, a full scan returns 0


tae898:
## Summary

`RESTORE DOCUMENT` executed in the **same transaction** as the `DELETE` leaves
the bucket's record count incremented for a record that is not there:
`SELECT count(*)` returns 1 while a full scan returns 0 rows. The disagreement
survives close and reopen, so it is the persisted count rather than a stale
in-memory statistic.

This is the same count-versus-scan shape as #6069, which 86cb4673be fixed. That
fix is correct for the separate-transaction case, which is the shape my original
repro used. Running both shapes side by side shows the same-transaction one
still disagreeing, in the opposite direction: #6069 counted one too few, this
counts one too many.

Not a regression. `RESTORE` does not parse at all on the 26.8.1 release
(`no viable alternative at input 'RESTORE'`), so the statement has never been in
a release and there is no earlier behaviour it could have regressed from.

## Reproduction

Java, engine API only, against `arcadedb-engine-26.9.1-SNAPSHOT.jar` built from
current `main` (includes 86cb4673be). Both shapes in one run so the contrast is
visible.

```java
import com.arcadedb.database.Database;
import com.arcadedb.database.DatabaseFactory;
import com.arcadedb.database.MutableDocument;
import com.arcadedb.query.sql.executor.ResultSet;

import java.nio.file.Files;
import java.nio.file.Path;

public class RestoreSameTxRepro {
  public static void main(String[] args) throws Exception {
    run(true);
    run(false);
  }

  private static void run(boolean sameTx) throws Exception {
    final Path dir = Files.createTempDirectory("restoresametx");
    final String path = dir.resolve("db").toString();
    final String label = sameTx ? "DELETE + RESTORE in ONE transaction "
                                : "DELETE then RESTORE, separate txs ";
    String rid;
    try (Database db = new DatabaseFactory(path).create()) {
      db.command("sql", "CREATE DOCUMENT TYPE Note");
      db.transaction(() -> {
        MutableDocument d = db.newDocument("Note");
        d.set("i", 7);
        d.save();
      });
      try (ResultSet rs = db.query("sql", "SELECT @rid AS r FROM Note")) {
        rid = rs.next().getProperty("r").toString();
      }
      final String theRid = rid;
      if (sameTx) {
        db.transaction(() -> {
          db.command("sql", "DELETE FROM " + theRid);
[...]

PR #6123, opened 2026-08-13
fix(#6120): fold the index entries on RESTORE DOCUMENT/VERTEX/EDGE

### LINKED ISSUE #6120: RESTORE DOCUMENT/VERTEX/EDGE does not re-add the record's index entries: an indexed query returns 0 rows while a full scan returns 1
Spun off from #6096 (PR #6119), where it was found and confirmed but deliberately left out of scope.

## Summary

`RESTORE DOCUMENT/VERTEX/EDGE` puts the record back in its bucket but never adds its index entries, so after a restore a query resolved through an index misses the record while a full scan returns it. Same failure shape as #6069 and #6096: a wrong answer returned with an ordinary success and nothing a caller can act on - except this one is worse than a count, because it silently drops the record from every indexed query, including a UNIQUE index's own duplicate check.

## Cause

The three `RESTORE` statements call `LocalBucket.restoreRecordAtPosition()` directly (through `RestoreStatementSupport.restoreRecordAndUpdateCount()`), bypassing `LocalDatabase.createRecord()` - which is where a normal insert indexes the document. `GraphEngine.restoreVertexAt()` does the same. #6082 folded the bucket record-count delta the same way for #6069; the index entries are the other half of that bookkeeping and were not folded.

## Reproduction

Engine API, `arcadedb-engine` on current `main` (with PR #6119 applied - the gap is independent of it):

```java
database.transaction(() -> {
  final var t = database.getSchema().createDocumentType("T", 1);
  t.createProperty("k", Type.STRING);
  t.createTypeIndex(Schema.INDEX_TYPE.LSM_TREE, true, "k");
});

final RID[] rid = new RID[1];
database.transaction(() -> rid[0] = database.newDocument("T").set("k", "a").save().getIdentity());
database.transaction(() -> database.command("sql", "DELETE FROM " + rid[0]));
database.transaction(() -> database.command("sql", "RESTORE DOCUMENT T RID " + rid[0] + " SET k = 'a'"));

// SELECT FROM T WHERE k = 'a'  -> 0 rows   (index lookup)
// SELECT FROM T                -> 1 row    (full scan)
```

Observed: `viaIndex=0 viaScan=1`.

## Notes for whoever picks it up

- The restore is deliberately an emergency-repair primitive, so one option is to declare "indexes are the operator's problem, run `REBUILD INDEX`". That should then at least be stated in the statement's javadoc and its result set, rather than being silent - the whole point of the #6069/#6096 family is that the caller cannot tell a wrong answer from a right one.
- The other option is to index the restored record the way `createRecord` does. On a UNIQUE index that also gives `RESTORE` a duplicate check it currently does not have: today a restore can create a second record carrying a key value the unique index already holds, and the index never learns about it.
- Whichever way it goes, `GraphEngine.restoreVertexAt()` needs the same treatment as the SQL statements.

lvca:
Fixes #6120. Follow-up to #6119 (#6096), which merged before this was ready.

## What was wrong

`RESTORE` put the record back in its bucket but never re-added its index entries. After a restore, an index-resolved query missed the record while a full scan returned it:

```
viaIndex=0  viaScan=1
```

And a UNIQUE index never learned the key had come back, leaving it free to hand that same key to a second record unchallenged. Same "wrong answer indistinguishable from a right one" family as #6069 and #6096, but on every indexed query rather than on a counter.

## The fix, and why it is not a one-liner

`LocalBucket.restoreRecordAtPosition()` performs only the physical page write, exactly like bucket-level create/delete. Everything a create folds on top of it is the caller's job - and having each RESTORE call site do that by hand is precisely how the record-count delta was missed in #6069 and the index entries here. Four call sites, four chances to forget, and #6082's shared helper could not even be reused by `GraphEngine.restoreVertexAt` without inverting the package dependency direction (that PR's review raised it and correctly declined).

So the primitive moves down to where the equivalent create bookkeeping already lives: **`DatabaseInternal.restoreRecord(record, bucket, position)`**, implemented in `LocalDatabase` beside `createRecordNoLock`. It folds:

- the bucket record-count delta `count(*)` reads (#6069)
- the transaction record cache
- **the index entries (#6120)**
- the new-record registration a rollback needs (#4562/#4940)

`RESTORE DOCUMENT/VERTEX/EDGE` and `GraphEngine.restoreVertexAt` all route through it; `RestoreStatementSupport`'s hand-rolled copy is gone. One place to forget instead of four.

## Two deliberate departures from `createRecordNoLock`

Both because RESTORE is admin repair, not user DML:

- **No `validate()` / `setDefaultValues()`** - a mandatory-property check would block a structure-only shell restore exactly when the emergency repair is needed.
- **No create events** - firing user triggers (including JS ones) during a repair is a behaviour change well beyond this gap.

Worth a maintainer opinion: both are now explicit decisions rather than accidents, and the second one means a restored record silently skips `onAfterCreate`. I lean toward documenting the contract in the SQL reference rather than changing it.

## Verification

Six regression tests, **each confirmed red with the fold removed**:

- `RestoreSqlStatementTest.restoreDocumentReindexesTheRestoredRecord`
- `RestoreSqlStatementTest.restoreVertexReindexesTheRestoredRecord`
- `RestoreSqlStatementTest.restoreEdgeReindexesTheRestoredRecord`
- `RestoreSqlStatementTest.restoreDocumentEnforcesTheUniqueIndexOnTheRestoredKey` - the duplicate rejection a restored key now gets
- `RestoreRecordAtPositionTest.restoreRecordReindexesTheRestoredRecord` - the Java API, via `restoreRecord` with a populated record
- `RestoreRecordAtPositionTest.rollingBackARestoreLeavesNoDanglingIdentityOnTheShell`

**No index test on `GraphEngine.restoreVertexAt`**: it restores a property-less shell, and I verified a null-valued indexed property produces no index entry on *any* path, insert included - so such a test would be vacuous. It is covered for the count fold (#6069) as before, and shares this fold through `restoreRecord`.

Full `engine` suite: 11715 tests, 1 failure - `Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent`, which fails identically on clean `main` (verified by stashing this change). Whole project compiles; `ServerDatabase` and `RaftReplicatedDatabase` gain the delegating override the new interface method requires.

Also carries `072430913`, the doc commit answering PR #6119's review point about the `contentEndOffset()` header guard being unreachable - it was pushed too late to reach that merge.


lvca:
Addressed in e1c1ee830 — with one correction to the finding, since it changes what the code should say.

**The reorder is applied.** `transaction.registerNewRecord(record)` now runs before `indexer.createDocument(...)`, matching `createRecordNoLock`.

**The cited mechanism does not reproduce, though.** The review names the unique-constraint path exercised by `restoreDocumentEnforcesTheUniqueIndexOnTheRestoredKey`, but a duplicate key is detected at **commit**, not inline. Measured directly against the pre-fix code:

```
restoreRecord returned, identity=#1:0
thrownAt=COMMIT: DuplicatedKeyException
identityAfterRollback=null
```

`restoreRecord` returns normally with an identity, `DuplicatedKeyException` comes out of `commit()`, and by then both calls have run — so the rollback already resets the identity to `null`. The suggested regression test would therefore pass with *and* without the reorder, so I did not add it rather than ship a green test that pins nothing.

(This also corrects the first review on this PR, which said the ordering "does not matter here". It does — just not via the duplicate-key path.)

**What does make the order matter:** `Index.put()` can throw inline — `checkIsValid()` on a dropped or invalidated index, and `convertKeys()` on a key it cannot coerce. On those paths, registering after indexing really would skip the rollback identity reset. So the change is worth making; it is defensive parity rather than a fix for a reachable-in-practice bug. Reaching either path requires an index dropped or corrupted mid-transaction, contrived enough that a test would pin the scaffolding rather than the behaviour, so none was added.

The code comment now records both halves — the reachable inline throws, and the fact that the unique path is *not* one of them — so this doesn't get re-litigated on the next read.

Verification: restore tests plus a `Restore*,*Transaction*,*Index*,*Rollback*,*Graph*` subset green. The prior commit had a full engine run at 11717 tests with one failure, `Issue5279ConcurrentUpdateTest`, which fails identically on clean `main`.

lvca:
Both points were valid — fixed in 908c8a5fb.

**1. Implicit commit outside the read lock.** Confirmed against the sibling: `createRecord` wraps *all* of `createRecordNoLock` — `checkTransactionIsActive`, the work, and the implicit commit/rollback — in a single `executeInReadLock`, so the commit really does happen under the lock there. `restoreRecord` now has the same shape.

Worth recording why this is safe, since it is the one thing that could have gone wrong: committing while holding a read lock would deadlock if `commit()` needed the write lock (a `ReentrantReadWriteLock` cannot upgrade). It does not — which is exactly why `createRecord`'s shape works — and the auto-transaction regression test drives an implicit commit from inside the lock and passes rather than hanging.

**2. Missing `poisonEdgeAppendPage` fold.** You offered a javadoc note or a guard; I took a third option and folded the same poison `createRecordNoLock` applies. A restored edge chunk is absent from the committed version of its page for the same reason a brand-new one is, so the poison is equally correct here, and a future caller gets working behaviour instead of a documented trap. The comment notes that `restoreRecordAtPosition` already poisons the *slot* merge and that this is the separate commutative edge-append merge, so the two don't get confused.

The javadoc's "two deliberate departures" list stays accurate — folding this means it was never a third departure.

**Verification:** `Restore*,*Transaction*,*Concurren*,*Drop*,*Close*,*Lock*,*Database*` subset — 489 tests, 1 failure (`Issue5279ConcurrentUpdateTest`, which fails identically on clean `main`). That selection deliberately includes the `drop()`/`close()` and locking tests, since the lock restructure is the change most likely to break something far from the restore path. A full engine run is in progress.

On the sandbox note: understood that this review was manual tracing only. Local runs have been backing each commit — the prior one had a full engine suite at 11717 tests with the same single pre-existing failure.

Issue #5777, opened 2026-08-02
checkEdges flags a missing IN endpoint as a corrupted record but a missing OUT endpoint as nothing, and flagging either contradicts the RECORD-scope principle


lvca:
Found while doing #5773 (PR #5776), out of that issue's scope because it changes what `CHECK DATABASE ... FIX` repairs. Line numbers against `2ee91ae11`.

`GraphDatabaseChecker.checkEdges`' endpoint probe handles a missing IN vertex and a missing OUT vertex differently, for the same `RecordNotFoundException`:

```java
// :1174 - incoming endpoint not found
addCorrupted(corruptedRecords, totalCorrupted, maxCorrupted, edgeRID);
addCorrupted(corruptedRecords, totalCorrupted, maxCorrupted, edge.getIn());   // <-- the missing VERTEX too

// :1211 - outgoing endpoint not found
addCorrupted(corruptedRecords, totalCorrupted, maxCorrupted, edgeRID);        // <-- edge only
```

Two things wrong with that, and they point in the same direction.

## 1. The asymmetry

An edge dangling on its IN side and an edge dangling on its OUT side are the same finding, and the same repair. Nothing in the code or the comments explains why one of them additionally flags the vertex. The unknown-error arms are asymmetric the same way in reverse - `:1188` and `:1224` BOTH add `edge.getIn()`/`edge.getOut()` - so the divergence is only in the `RecordNotFoundException` arms, which is what a copy-paste divergence looks like rather than a decision.

## 2. Flagging a not-found RID as corrupted contradicts the principle #5680/#5764 established

`edge.getIn()` reached that arm by raising `RecordNotFoundException` - the record is **not there**. Both scoped arms already say what that means, at `:469` and `:1256`:

> A RID that simply is not there is NOT corruption, and the difference is expensive: a record flagged corrupted puts its BUCKET into `affectedBuckets`, and `CHECK DATABASE ... FIX` then drops and rebuilds every index on that bucket - a full bucket scan.

So the type-wide edge check puts the *vertex* bucket into `affectedBuckets` on the strength of a vertex that does not exist, and `FIX` then drops and rebuilds every index on it. The delete loop itself is harmless (`bucket.deleteRecord(rid)` raises `RecordNotFoundException`, which is caught and ignored) - the cost is entirely the index rebuild, and it is not small: a dangling-edge sweep after a bulk delete would rebuild the indexes of every vertex bucket the deleted vertices lived in.

[...]

PR #6131, opened 2026-08-13
fix(#6128): bound the CHECK DATABASE repair transaction, and stop autoFix counting repairs that did not happen

### LINKED ISSUE #6128: CHECK DATABASE FIX on a Raft cluster: unsplittable per-type repair entry, buffered index rebuild, and three smaller repair-path defects
Follow-ups discovered while adding the first HA cluster coverage for `CHECK DATABASE` (#6121, merged as `c7ad5e083`). That PR fixed the document-arm defect it set out to fix; these are the surrounding problems it surfaced but deliberately did not touch. They are filed together because they share one root observation: **the checker's repair path was designed for an embedded, single-node database, and every assumption that follows from that breaks differently under Raft.**

Ordered by operator impact. (1) and (2) are the ones that make `CHECK DATABASE FIX` unsafe to recommend on a large cluster today.

Context: a customer running a 3-node embedded HA cluster on a several-hundred-GB super-node graph reported a `CHECK DATABASE` taking ~80 hours (the scan cost itself is #6062) and then started a `CHECK DATABASE FIX` on that cluster. Points (1) and (2) are what that FIX runs into.

---

## 1. A per-type repair is one unsplittable Raft entry, so a large FIX aborts at the commit

`GraphDatabaseChecker.checkVertices`, `checkEdges` and `reclaimOrphanedEdgeSegments` each open ONE transaction covering the entire type scan plus every repair it makes (`begin()` at the top, `commit()` at the bottom). On a single node that is merely a big transaction. Under Raft it becomes ONE log entry, and `RaftTransactionBroker.replicateTransaction` submits it whole:

```java
final ByteString entry = RaftLogEntryCodec.encodeTxEntry(dbName, walData, bucketDeltas);
return groupCommitter.submitAndWait(entry.toByteArray());
```

`submitAndWait` hard-rejects anything above `min(arcadedb.ha.appendBufferSize, arcadedb.ha.grpcMessageSizeMax)` — **32MB of compressed WAL by default** — with `ReplicatedEntryTooLargeException`. Schema entries got a splitter in #4743 (`splitSchemaEntry`); transaction entries never did.

Consequences:

- A broad repair runs for hours and then dies at the very end, rolling back that type's entire repair. Types already committed stay repaired, so a failed run leaves the database partially fixed with no indication of where it stopped.
- `ReplicatedEntryTooLargeException` extends `TransactionException`, **not** `NeedRetryException`, so no retry machinery absorbs it and it propagates out of `check()`.
- Quiescing the cluster does not help. The entry size is a function of how much damage there is to repair, not of concurrency.

**Suggested fix:** bound the repair into batched transactions inside `GraphDatabaseChecker`, rather than teaching TX entries to split. Batching is cheaper, bounds leader heap at the same time (see below), and a partially-committed repair is already the semantics of a run that fails part-way through the type list. Splitting the entry would preserve a transaction so large that it is a problem on a single node too.

## 2. An index rebuild buffers the whole index in leader heap, and stalls every writer while it does

`BucketIndexBuilder.create()` wraps the build in `schema.recordFileChanges(...)`, which sets `isSchemaCommitThread`. `RaftReplicatedDatabase.commit()` therefore **buffers** rather than replicates, so the 5000-record batch commits inside `LSMTreeIndex.build()` accumulate in `schemaWalBuffer` and drain only when the callback returns. Peak leader heap is roughly the whole rebuilt index (plus repeats of pages touched across batches).

Worse, while that session is open every *other* commit on the leader hits `waitForActiveRecordingSession()`, which waits up to `HA_QUORUM_TIMEOUT` (10s default) and then gives up and proceeds anyway:

```java
HALog.log(this, HALog.BASIC,
    "commit waited %dms for FileManager recording session and gave up; proceeding with TX_ENTRY", timeout);
```

Proceeding is the #4083 ordering hazard: the TX_ENTRY can reach Raft before the SCHEMA_ENTRY that creates the file, and `TransactionManager.applyChanges` silently skips pages whose fileId does not exist on the follower. So concurrent writes during a FIX index rebuild can be **dropped on followers with no error**. `BucketIndexBuilder.create()` additionally parks every async executor thread for the duration.

**Suggested fix:** stream the rebuild WAL instead of buffering it, so the recording session does not have to stay open across the whole build.

## 3. `autoFix` counts deletes that failed

`GraphDatabaseChecker.checkVertices`/`checkEdges` call `autoFix.incrementAndGet()` **before** attempting the delete, so a record they then fail to remove is still reported as repaired. Observed directly, both fields in the same result:

```json
{"autoFix": 1,
 "warnings": ["Cannot fix the record #1:0: error on delete (error: Multi-page record #1:0 chunk 0 was modified concurrently. Please retry)"],
 "deletedRecordsAfterFix": []}
```

`autoFix` is what an operator reads to decide whether a run did anything, and a number that includes failed deletes cannot answer that. #6121 added `DatabaseChecker.deleteCorruptedRecords`, which counts only after the delete and its counter delta both succeed, and documented the divergence rather than silently changing the graph arms' numbers.

**Suggested fix:** move the two `incrementAndGet()` calls inside the `try`, after the delete returns. Small, but it changes reported numbers, so it wants its own change and a look at whatever asserts them.

## 4. `COMPRESS` should be refused, or at least warned about, on a replicated database

`DatabaseChecker.compress()` commits every 10 pages:

```java
if (pageBatch >= pageTxBatch) {   // pageTxBatch = 10
  database.commit();
  database.begin();
```

On a cluster that is one Raft consensus round trip per 10 pages. On a database of any real size it is not something that finishes. Since `CHECK DATABASE ... COMPRESS` is reachable by anyone with the privilege, it should fail fast with an explanation on a replicated database, or at minimum warn the way the RECORD-scope arm already warns that COMPRESS is not scoped.

## 5. Two documentation gaps that read as bugs to an operator

**`CHECK DATABASE` is leader-local, in both directions.** `CheckDatabaseStatement` does not override `Statement.isIdempotent()`, so it is classified non-idempotent and `RaftReplicatedDatabase.command()` forwards it to the leader — with or without `FIX`. That is the correct routing, but it means the check only ever inspects the leader's pages, and repairs reach the other nodes as WAL deltas over specific page ranges. A follower-only defect is invisible to it, and a repair does not heal a follower whose page content diverged. "CHECK DATABASE came back clean" says nothing about nodes 2 and 3, and nothing in the output says so.

**The progress endpoint is node-local.** `GetProgressHandler` reads a local `OperationProgressRegistry`, so on a cluster progress must be polled on the leader. Polling a follower returns an empty list, which during a multi-hour check is indistinguishable from a hang.

---

### Not included here

The scan cost that makes a full run take ~80 hours is #6062 (unconditional O(degree) back-reference probes) and is unaffected by any of the above. `RaftCheckDatabaseFix3NodesIT` (added in #6121) is the natural home for regression coverage of (1) and (2).

lvca:
Closes #6128 (points 1, 3, 4, 5).

**Point 2 is deliberately not in this PR.** Streaming the index-rebuild WAL means changing the HA schema-replication protocol - the area with the documented history of silent follower divergence (#4083, #4743, #5492) - and it is separable from these four. Detail at the end.

## 1. The per-type repair was one unsplittable Raft entry

The repair of one type was a SINGLE transaction: the whole scan plus every reconnected edge and every deleted record between one `begin()` and one `commit()`. On a replicated database that is a SINGLE log entry, and `RaftGroupCommitter.submitAndWait` rejects anything above `min(appendBufferSize, grpcMessageSizeMax)` with `ReplicatedEntryTooLargeException` - a `TransactionException`, not a `NeedRetryException`, so nothing retries it. A repair large enough to matter ran for hours and was then rolled back whole. A schema entry has had a splitter since #4743; a transaction entry has none.

The repair now commits and reopens once it has dirtied `arcadedb.checkDatabaseRepairBatchPages` pages (default **256**, `0` disables).

**Why pages and not records** - pages are what the entry carries. How many records a repair touched says nothing about how many distinct pages it dirtied, and it is the page images the WAL holds. `TransactionContext.getModifiedPages()` already counts modified plus new pages.

**Why this is a semantic change, stated plainly** - a repair is no longer all-or-nothing; a failure part-way leaves earlier batches committed. That is already how a multi-type run behaves (`check()` commits each type before the next), so this makes one type behave like the whole run rather than inventing a third semantics. And on a replicated database the alternative is not an atomic repair, it is no repair at all. Set the budget to `0` for the old behaviour.

**Where it is safe to commit** - only between units of repair work, never inside a `scanType`/`scan` callback: `LocalDatabase.scanType` holds the database read lock and owns an implicit transaction for the length of the scan, so committing under it would commit a transaction the scan believes it still owns. Every call site is a post-scan apply or delete loop, which is also where the page volume actually is.

## 3. `autoFix` counted repairs that did not happen

The vertex and edge arms called `incrementAndGet()` **before** attempting the delete, so a record they then failed to remove was still reported as fixed. The case is routine, not exotic: `checkEdges` flags both ends of a dangling edge, and the far end is flagged precisely because it is not there, so its delete always raises `RecordNotFoundException`. One dangling edge reported two repairs.

Counted after the delete returns now. **Pruning a dangling adjacency entry is also counted**, which it never was - without that, a FIX over a database whose only damage was dangling references (the commonest shape, since the far record is usually already gone) would report `autoFix = 0` having repaired every one of them.

`CheckDatabaseTest`'s existing expectations decompose exactly this way, which is what confirmed the split rather than assumed it:

| case | old | new | why |
| --- | --- | --- | --- |
| `checkBrokenDeletedVertex` | 19998 | 19998 | 9999 deletes + 9999 prunes - unchanged |
| `checkBrokenDeletedEdges` | 1 | **2** | one pruned entry per side, matching the `invalidLinks = 2` the same run already reports |

The second assertion is updated, with the reasoning in the test.

## 4. `COMPRESS` on a cluster

It committed every 10 pages, so on a replicated database that is one Raft consensus round trip per ten pages - not a slow operation but one that does not finish. It now reuses the repair budget (same question, one knob), a 25x reduction in round trips at the default while staying well inside the entry cap. `0` falls back to the old 10 rather than turning into a single transaction over every page in the database.

## 5. The result now says which copy it describes

A `checkedNodeScope` field plus a warning when the database is replicated. The statement is non-idempotent so a follower forwards it to the leader; the check reads only the leader's pages, and repairs reach the others as WAL deltas over the ranges they touched. Neither direction covers a follower whose own copy diverged. Reported rather than documented because the dangerous reading of a clean result - "the cluster is fine" - is one an operator makes from the output.

## Tests

`CheckDatabaseRepairBatchTest` counts commits in both directions: budget `0` gives exactly one transaction, a small budget gives more, and both must repair the graph identically.

Its fixture corrupts head chunks **in place** rather than deleting them, and that is not a detail. Deleting frees the slot; the repair then builds its new chunks into those freed slots, and a hub's dangling head comes to alias another hub's live chunk. That artefact - which no real corruption produces, since damaged bytes do not free a slot - makes the repair look broken: measured at **11 of 40 hubs unrepaired on unmodified code**, until the corruption was changed to leave the slot occupied, when all 40 repair correctly.

`RaftCheckDatabaseLargeRepairIT` pins the cluster half with the entry cap two orders of magnitude below its default. **It does not reproduce an over-cap rejection and says so in its javadoc**: the WAL carries per-page changed ranges rather than whole pages, so reconnecting 1500 edges stays well under 128KB - measured passing with batching disabled at both 1MB and 128KB. Brute-forcing past the cap would need tens of thousands of damaged edges and minutes of cluster setup for a bound better expressed directly. What it proves is that a repair split across many transactions converges all three nodes under a cap tight enough that any future oversized repair transaction fails here rather than in a customer's cluster.

Verified: 1671 engine tests across `database`/`graph`/`engine`/`index` with one unrelated pre-existing failure (`Issue5279ConcurrentUpdateTest`, reproduces on a clean tree), plus `RaftCheckDatabaseLargeRepairIT` 2/2 and `RaftCheckDatabaseFix3NodesIT` 3/3.

## On point 2

`BucketIndexBuilder.create()` runs inside `recordFileChanges`, so `LSMTreeIndex.build`'s 5000-record batch commits are buffered into `schemaWalBuffer` and drained only when the callback returns - peak leader heap is roughly the whole rebuilt index. The fix I would propose is not a new protocol but doing incrementally what `splitSchemaEntry` already does at the end: flush once the buffer crosses a threshold, first flush carrying `filesToAdd`, intermediates carrying WAL only, the last carrying `schemaJson` and `filesToRemove` - the same ordered-prefix contract, spread over time instead of built up and split. That keeps every follower prefix self-consistent. It still touches the code path behind #4083, so it wants its own PR and its own cluster test rather than riding along with four checker fixes.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -439,6 +439,21 @@ public enum GlobalConfiguration {
       Larger values reduce commit overhead on single-node setups at the cost of bigger transactions.""",
       Integer.class, 1000),
 
+  CHECK_DATABASE_REPAIR_BATCH_PAGES("arcadedb.checkDatabaseRepairBatchPages", SCOPE.DATABASE,
+      """
+      Number of modified pages CHECK DATABASE ... FIX accumulates before committing its repair and opening the next \
+      transaction. Same rationale as arcadedb.truncateBatchSize, and the same failure it avoids (issue #6128): the \
+      repair of one type - every reconnected edge and every deleted record - used to be a SINGLE transaction, which \
+      in HA is a SINGLE Raft log entry, and a transaction entry has no splitter the way a schema entry has had since \
+      issue #4743. Above min(arcadedb.ha.appendBufferSize, arcadedb.ha.grpcMessageSizeMax) the entry is rejected with \
+      ReplicatedEntryTooLargeException - not a NeedRetryException, so nothing retries it - and a repair that had run \
+      for hours was rolled back whole. Counted in PAGES rather than records because pages are what the entry \
+      contains: how many records a repair touched says nothing about how many distinct pages it dirtied. The default \
+      of 256 stays well under the 32MB appendBufferSize default even at the 64KB bucket page size and before the WAL \
+      is compressed. Raising it lowers commit overhead on an embedded database at the cost of bigger transactions; 0 \
+      disables batching entirely, restoring the all-or-nothing repair semantics of a single transaction.""",
+      Integer.class, 256),
+
   PAGE_FLUSH_QUEUE("arcadedb.pageFlushQueue", SCOPE.DATABASE, "Size of the asynchronous page flush queue", Integer.class, 512),
 
   FLUSH_SUSPEND_MAX_DEFERRED_RAM("arcadedb.flushSuspendMaxDeferredRAM", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -448,9 +448,15 @@ public enum GlobalConfiguration {
       issue #4743. Above min(arcadedb.ha.appendBufferSize, arcadedb.ha.grpcMessageSizeMax) the entry is rejected with \
       ReplicatedEntryTooLargeException - not a NeedRetryException, so nothing retries it - and a repair that had run \
       for hours was rolled back whole. Counted in PAGES rather than records because pages are what the entry \
-      contains: how many records a repair touched says nothing about how many distinct pages it dirtied. The default \
-      of 256 stays well under the 32MB appendBufferSize default even at the 64KB bucket page size and before the WAL \
-      is compressed. Raising it lowers commit overhead on an embedded database at the cost of bigger transactions; 0 \
+      contains: how many records a repair touched says nothing about how many distinct pages it dirtied. Sizing the \
+      default of 256, without rounding the arithmetic in its own favour: 256 pages at the 64KB bucket default is \
+      16MB of PAGES, half the 32MB appendBufferSize default, which is margin rather than comfort once the soft \
+      ceiling below is taken into account. The entry itself is far smaller in practice - the WAL carries each \
+      page's CHANGED RANGE rather than the whole page, and is compressed on top of that, so a repair reconnecting \
+      1500 edges was measured well under 128KB - but that is a property of typical repairs, not a bound. If a \
+      deployment raises the bucket page size or lowers appendBufferSize, re-check this against BOTH rather than \
+      trusting the default. Raising it lowers commit overhead on an embedded database at the cost of bigger \
+      transactions; 0 \
       disables batching entirely, restoring the all-or-nothing repair semantics of a single transaction. \
       ALSO BOUNDS CHECK DATABASE ... COMPRESS, whose per-transaction page count was a hardcoded 10 - one Raft round \
       trip per ten pages, which on a replicated database of any size does not finish; COMPRESS keeps that 10 when \
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -451,7 +451,13 @@ public enum GlobalConfiguration {
       contains: how many records a repair touched says nothing about how many distinct pages it dirtied. The default \
       of 256 stays well under the 32MB appendBufferSize default even at the 64KB bucket page size and before the WAL \
       is compressed. Raising it lowers commit overhead on an embedded database at the cost of bigger transactions; 0 \
-      disables batching entirely, restoring the all-or-nothing repair semantics of a single transaction.""",
+      disables batching entirely, restoring the all-or-nothing repair semantics of a single transaction. \
+      ALSO BOUNDS CHECK DATABASE ... COMPRESS, whose per-transaction page count was a hardcoded 10 - one Raft round \
+      trip per ten pages, which on a replicated database of any size does not finish; COMPRESS keeps that 10 when \
+      this is set to 0, since one transaction over every page in the database helps nobody. \
+      A SOFT ceiling, not a hard cap: the budget is checked between units of repair work, so a transaction can \
+      exceed it by whatever the unit in flight dirties (reconnecting one very wide adjacency list, or a hub record \
+      spanning several pages). Leave headroom when picking a value close to the replicated-entry limit.""",
       Integer.class, 256),
 
   PAGE_FLUSH_QUEUE("arcadedb.pageFlushQueue", SCOPE.DATABASE, "Size of the asynchronous page flush queue", Integer.class, 512),
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -18,6 +18,7 @@
  */
 package com.arcadedb.engine;
 
+import com.arcadedb.GlobalConfiguration;
 import com.arcadedb.database.Database;
 import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.Document;
@@ -135,6 +136,17 @@ public class DatabaseChecker {
     result.put("totalCorruptedRecords", 0L);
     result.put("distinctMissingReferences", 0L);
     result.put("topMissingReferences", new ArrayList<String>());
+    // Issue #6128 (5): says WHICH copy of a replicated database this answer describes. The statement is classified
+    // non-idempotent, so a follower forwards it to the leader and the check only ever reads the leader's pages;
+    // repairs then reach the other nodes as WAL deltas over the page ranges they touched. Neither direction covers a
+    // follower whose own copy diverged. Reported as a field rather than left to the documentation because the
+    // dangerous reading of a clean result - "the cluster is fine" - is one an operator makes from the output.
+    result.put("checkedNodeScope", database.isReplicated() ? "this node only (replicated database)" : "whole database");
+
+    if (database.isReplicated())
+      addWarning("this database is replicated and this check inspected only the node that ran it: a clean result "
+          + "does not certify the other nodes, and a repair reaches them as replicated changes rather than by each "
+          + "node repairing itself. Progress is published per node too, so poll it on the node running the check");
 
     // COMPUTE THE STEP PLAN UPFRONT so every progress emission carries a stable stepIndex/totalSteps pair.
     final List<DocumentType> edgeTypes = new ArrayList<>();
@@ -605,7 +617,16 @@ public class DatabaseChecker {
       totalPagesToCompress += ((LocalBucket) b).getTotalPages();
     stepBegin("Compressing buckets", totalPagesToCompress);
 
-    int pageTxBatch = 10;
+    // Issue #6128 (4): the hardcoded 10 made COMPRESS unusable on a replicated database. Every commit is a Raft
+    // consensus round trip, so ten pages per transaction means one round trip per ten pages - millions of them on a
+    // database of any size, which is not a slow operation but one that does not finish. Reusing the repair budget
+    // rather than adding a second knob: it answers the same question ("how many pages may one CHECK DATABASE
+    // transaction accumulate"), and at its 256 default this is a 25x reduction in round trips while staying well
+    // under the appendBufferSize the entry has to fit in. Guarded so a 0 (batching disabled) does not turn into a
+    // single transaction over every page in the database, which is the one outcome nobody wants here.
+    final int budget = database.getConfiguration()
+        .getValueAsInteger(GlobalConfiguration.CHECK_DATABASE_REPAIR_BATCH_PAGES);
+    final int pageTxBatch = budget > 0 ? budget : 10;
     int pageBatch = 0;
 
     for (final Bucket b : database.getSchema().getBuckets()) {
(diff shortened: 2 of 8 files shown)


lvca:
All three points were valid and are addressed in `be6715716`. Documentation only - no behaviour change.

**The unbounded in-scan repair.** Correct, and I confirmed it at `GraphDatabaseChecker` L913-920: the `connected != null && !connected` arm calls `connectOutgoingEdge` and saves a vertex per defective edge, from inside the scan, with the `outVertex` twin doing the same. `resetChain` is in the same position. So a database with very many edges in that particular state can still accumulate one large transaction.

Named it in `commitRepairBatchIfFull()`'s javadoc rather than only in the PR description, since the person who needs to know is the next reader of that method, not the next reader of this thread. Recorded as accepted rather than overlooked: that shape is much narrower than the dangling references and broken chains the post-scan loops handle, and bounding it means restructuring those checks into collect-then-apply so the writes leave the scan - a bigger change that should be judged on its own rather than folded in here.

**The config doc not mentioning COMPRESS.** Fixed. It now states that the setting also bounds `CHECK DATABASE ... COMPRESS`, what the hardcoded 10 cost on a replicated database, and that COMPRESS keeps that 10 when the budget is 0 (one transaction over every page in the database helps nobody).

**Soft ceiling, not a hard cap.** Fixed in both the config doc and the method javadoc: the check happens between units, so a transaction can exceed the budget by whatever the unit in flight dirties, and there should be headroom when tuning near the replicated-entry limit.

Thanks for tracing the nested-transaction path through `pushTransaction`/`popIfNotLastTransaction` and confirming each batch is a real separately-replicated commit - that was the assumption the whole change rests on, and it is better to have it verified by someone other than the author.

lvca:
Addressed in `19a96e52c`.

**The dangling nested transaction.** Good question, and the right one to ask about a PR whose whole subject is mid-repair failure semantics. Tested rather than reasoned about: `CheckDatabaseRepairBatchFailureTest` fails the *second* batch commit from the `TX_AFTER_WAL_WRITE` callback - so the run has genuinely committed one batch before it breaks - and asserts the two properties that matter.

The concern does not reproduce. The exception reaches the caller rather than being swallowed into a completed-repair report, and `database.isTransactionActive()` is `false` afterwards: `commit()` disposes its context even when the write throws, so the checker does not need a rollback in a `catch`. The database is still usable on that thread, which the test also asserts by running a transaction after the failure. Batches already committed stay committed, which is the semantic change the PR states.

The test stays even though it needed no production change, because batching is exactly what makes a mid-repair failure possible in the first place - it pins behaviour this PR newly depends on.

**The nesting mechanism.** You are right that it was never spelled out, and it is the linchpin: if a nested `begin()`/`commit()` were a savepoint deferring the write to the outermost commit, this entire change would be inert - the repair would still reach Raft as one entry. It is now in `commitRepairBatchIfFull()`'s javadoc, including that `pushTransaction` gives each pair a separate `TransactionContext` on which the full two-phase commit runs. Thank you for checking it independently; that is the assumption I would least want to have got wrong.

**Per-edge granularity in `reconnectEdges`.** Leaving as is, deliberately. The check is a field read and an int comparison against a value cached in the constructor, so amortizing it would trade a measurable nothing for a coarser bound - and the bound is the point: checking after every unit is what keeps "how far can a transaction overshoot the budget" equal to one unit of work rather than to a batch of them. The javadoc now calls the ceiling soft for exactly that reason, and making the check less frequent would widen the overshoot it is warning about.

lvca:
Both addressed in `28c92fe97`, documentation only.

**The index-change backlog.** You were right to flag it and right that it is currently moot. Confirmed: `TransactionContext.hasChanges()` ORs `modifiedPages`/`newPages` with `indexChanges`, which is exactly the tell that the two are tracked separately, and `indexChanges` is materialised into pages only in `commit1stPhase` - after this check has run. So `getModifiedPages()` cannot see a pending index backlog.

Every call site under the budget is safe today for the reason you gave: a raw `bucket.deleteRecord` and a `getOrCreateEdgeList(...).add(...)` maintain no type index. But the failure mode if that changes is nasty precisely because it is quiet - the budget would keep returning a small number while the real entry grew - so it is now written into the method's javadoc as a precondition on call sites, with the instruction to bound by entry count as well before adding one that does index-maintained updates.

**The default's sizing.** Fair, and the wording was doing my argument a favour. 256 pages at 64KB is 16MB against a 32MB `appendBufferSize` - exactly half, which is margin but not the comfort "well under" implies, especially with a soft ceiling.

I have stated the arithmetic rather than tightened the default, because the honest picture is that both numbers matter and neither is the whole story: the entry is far smaller than the page total in practice, since the WAL carries each page's changed RANGE rather than the whole page and compresses on top - a repair reconnecting 1500 edges measured well under 128KB, which is the same measurement that stopped `RaftCheckDatabaseLargeRepairIT` from being able to force an over-cap rejection. The doc now says that, and says explicitly that it is typical-case behaviour rather than a bound, and tells an operator who raises the bucket page size or lowers `appendBufferSize` to re-check against both instead of trusting the default. Lowering the default to buy margin against a worst case that the range-encoding makes very unlikely would cost every ordinary repair extra round trips for it.

lvca:
Both notes addressed in `8b9c2829f`, documentation only. Declining the third with reasoning.

**`autoFix` counts repair actions, not corruption instances.** Agreed, and it deserved saying given how precise the rest of the accounting now is. One edge that is both listed in an adjacency chain and corrupt as a record contributes a prune *and* a delete - two distinct writes to two distinct pages - so an operator reading the number as "how many broken things were there" over-counts. Now stated on the `prunedDanglingEntries` field where the semantics live, together with why the alternative is worse: counting defects would need the arms to agree on what one defect *is* across a dangling entry, a corrupt record and a rebuilt chain, and they cannot without collapsing information the warnings carry.

**`checkEdges`'s `+ report.prunedDanglingEntries` is structurally zero.** Confirmed - `checkIncomingEdges`/`checkOutgoingEdges` are called only from `checkVertices` (lines 559 and 562), so the edge arm cannot prune today.

Kept rather than dropped, with a comment saying exactly that, for two reasons: it keeps the two arms reporting `autoFix` by the same expression, and #5777 is open against precisely this arm's handling of endpoints - if that work adds pruning here, a dropped term would silently stop counting it. The comment is there so no future reader chases the phantom path you correctly spotted.

**`compress()` re-reading the config instead of caching it.** Leaving as is. `compress()` reads it once per invocation and is called at most once per check, so there is no repeated read to hoist - the asymmetry with `GraphDatabaseChecker`'s constructor is real but it is the constructor that needs the field, because there the value is consulted once per repaired record on a per-record path. Caching in `compress()` would add a field to make two call sites look alike while changing nothing about what either does.

Agreed on `corruptRecordTypeByte` being duplicated across the module boundary - `ha-raft` does not take engine's test-jar, and one helper is not worth restructuring that for.

lvca:
Round 5 raises no new points to act on, so I am leaving the branch as it stands.

The three nits are all either already answered or ones you would not change yourselves: the `corruptRecordTypeByte` duplication is the `ha-raft`/engine test-jar module boundary and there is no shared HA test-utility home to hoist it to; the `commitRepairBatchIfFull()` javadoc length you explicitly say not to cut, and I agree - the precondition and the unbounded-paths section are the two things a future caller most needs and burying them in a separate design note would make them easier to miss, not harder; and `compress()` re-reading the config was answered last round (it reads once per invocation and is called at most once per check, so there is no repeated read to hoist).

Five rounds, no correctness bug found in any of them. Thank you for tracing the nesting depth to `maxNested = 3` - the checker sits exactly at the cap, which is worth knowing before anyone adds another nesting level to this call path.

PR #6137, opened 2026-08-13
fix(#6104): reject Windows-illegal characters and reserved device names in bucket names

### LINKED ISSUE #6104: Bucket/type name validation does not reject Windows-illegal characters or reserved device names
Raised in review on #6099 as a non-blocking follow-up.

`LocalSchema.checkValidBucketName()` and `Neo4jImporter.validateLabel()` reject the characters that matter for correctness: `/`, `\`, the `.`/`..` sentinels, and `*` (the one `URLEncoder` leaves untouched, so it reaches the file name intact).

They do not reject the rest of the Windows-illegal set (`<`, `>`, `:`, `"`, `|`, `?`, control characters) or the reserved device stems (`CON`, `PRN`, `AUX`, `NUL`, `COM1`-`COM9`, `LPT1`-`LPT9`).

This is not a directory-escape concern: that vector is closed, and `URLEncoder` percent-encodes most of these on the type-name path anyway. The gap is only in error quality on Windows. A bucket created directly with such a name, or a type name where encoding leaves the character intact, fails later with a raw `IOException` out of `Files.move`/file creation rather than a `SchemaException` naming the offending character at validation time.

Worth doing if Windows is a supported deployment target for embedded/server use. The check would go next to the existing ones in `checkValidBucketName`, with the importer's `validateLabel` kept in step.

lvca:
## Summary

Fixes #6104.

`LocalSchema.checkValidBucketName()` already rejected path separators, the `.`/`..` sentinels and `*`, but let the rest of the NTFS-illegal character set through: `< > : " | ?` and control characters, plus the reserved device stems `CON`, `PRN`, `AUX`, `NUL`, `COM1`-`COM9`, `LPT1`-`LPT9`.

A bucket name is used verbatim as the last path segment of its component file with no percent-encoding, so any of those reached the filesystem unchanged and failed later with a raw `IOException` (out of `Files.move`/file creation) instead of a `SchemaException` naming the offending character at validation time.

- Added the remaining Windows-illegal characters (`< > : " | ?` and control chars `< 0x20`) to the existing per-character check.
- Added a reserved-device-name check: matches Windows semantics precisely by comparing the segment of the name up to (not including) the first `.` against the reserved stems, case-insensitively, so `CON.dat` is rejected exactly like bare `CON`, while `CONFIG` or `a.CON` are accepted.

### Why `Neo4jImporter.validateLabel()` is intentionally untouched

The issue suggested keeping the importer's `validateLabel()` in step. I checked instead of assuming:

- A type name is always run through `FileUtils.encode()` (`URLEncoder`) before it reaches disk, which already turns `< > : " | ?` and control characters into safe `%XX` sequences. `validateLabel()`'s existing checks are deliberately scoped to only the handful of characters/sequences that *survive* that encoding unchanged (`/ \ . .. * ,` and the label separator) - see its own doc comment.
- A type-derived bucket name is always built as `<encodedTypeName>_<index>` (`TypeBuilder.java`), so it can never collide with a bare reserved device name like `CON` - the mandatory `_0` suffix always breaks the match.

So the gap this issue describes only exists on the direct `CREATE BUCKET` path, which is exactly where this fix is applied. Adding the same checks to `validateLabel()` would be dead code for a scenario that structurally cannot happen there.

## Test plan
- [x] New unit tests in `BucketNameValidationTest`: rejects the remaining illegal characters (including control characters), rejects reserved device stems (with and without a following dot, case-insensitively), and accepts names that only resemble a reserved stem (`CONFIG`, `a.CON`, `COM10`, etc.)
- [x] `mvn -pl engine test -Dtest=BucketNameValidationTest` - 8/8 pass
- [x] Full `com.arcadedb.schema.**` package test suite - no regressions (40/40 classes pass)
- [x] `Neo4jImporterLabelCharactersIT` - no regressions (3/3 pass)
- [x] `mvn -pl engine compile` - clean

### DIFF
--- a/engine/src/main/java/com/arcadedb/schema/LocalSchema.java
+++ b/engine/src/main/java/com/arcadedb/schema/LocalSchema.java
@@ -559,6 +559,17 @@ public class LocalSchema implements Schema {
     });
   }
 
+  // The rest of the NTFS/Windows-reserved character set, beyond '/', '\' and '*' which are checked separately
+  // above: '<', '>', ':', '"', '|' and '?'.
+  private static final String WINDOWS_ILLEGAL_CHARS = "<>:\"|?";
+
+  // Windows reserves these as device names for the segment of a file name up to (and not including) the first
+  // dot, regardless of what an extension or further dotted segment says: "CON.txt" is refused exactly like "CON".
+  private static final Set<String> WINDOWS_RESERVED_NAMES = Set.of(//
+      "CON", "PRN", "AUX", "NUL",//
+      "COM1", "COM2", "COM3", "COM4", "COM5", "COM6", "COM7", "COM8", "COM9",//
+      "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9");
+
   /**
    * A bucket name becomes the last path segment of its component file, so it must address a file inside the
    * database directory and nothing else. A type name reaches the same place already percent-encoded by
@@ -568,6 +579,12 @@ public class LocalSchema implements Schema {
    * Dots inside the name are deliberately allowed: the component-file name is parsed right-to-left, peeling the
    * fixed {@code .fileId.pageSize.vVersion.ext} tail, so a dot in the name survives the round trip. Only a name
    * that is exactly "." or ".." references a directory.
+   * <p>
+   * The remaining checks reject the rest of the NTFS/Windows-illegal character set and the reserved device stems
+   * ({@code CON}, {@code PRN}, {@code AUX}, {@code NUL}, {@code COM1}-{@code COM9}, {@code LPT1}-{@code LPT9}).
+   * Nothing here is a directory-escape concern, since a bucket name is never used as anything but the last path
+   * segment: it is error quality on Windows, turning a raw {@code IOException} out of {@code Files.move}/file
+   * creation into a {@code SchemaException} that names the offending character or stem at validation time.
    */
   public static void checkValidBucketName(final String bucketName) {
     if (bucketName == null || bucketName.isEmpty())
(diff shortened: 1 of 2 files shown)


lvca:
Addressed in c969f4a68:

- **Constant placement**: moved `WINDOWS_ILLEGAL_CHARS`/`WINDOWS_RESERVED_NAMES` up to the class's existing static-constants block (next to `BUILD_TX_BATCH_SIZE` etc.) instead of leaving them mid-class.
- **Unprintable char in message**: a rejected control character (NUL, ESC, ...) is now rendered as `\uXXXX` in the `SchemaException` message instead of the raw byte.

Not changed, with reasoning:
- **SQL/type-path end-to-end test**: skipped. `checkValidBucketName()` is called once at the top of the single canonical `createBucket(...)` overload that every path (SQL `CREATE BUCKET`, `TypeBuilder`, edge-bucket creation, `AlterTypeStatement`) funnels through, and the existing `createBucketViaSqlRejectsNamesThatLeaveTheDatabaseDirectory` test already proves the SQL path reaches the same validation. Adding a parallel SQL-level test per character class would duplicate coverage of a single shared call site without added confidence.
- **Raw `bucketName` still echoed unescaped in the message**: left as is - that's pre-existing behavior shared by all the other checks in this method (not introduced by this PR), out of scope here.

All 8 `BucketNameValidationTest` tests still pass; engine module compiles clean.

lvca:
Thanks for the second pass. On the three notes:

1. **Raw bucketName echo in messages** - same as last round, pre-existing and out of scope; noted this is fine to leave.
2. **Trailing spaces/dots at the end of a filename** - checked, and it turns out not to apply here: the bucket name is only ever a *prefix* of the physical file name (`<bucketName>.<fileId>.<pageSize>.v<version>.<ext>`), so a trailing dot or space in `bucketName` is never the trailing character of the actual file on disk - it's always followed by the fixed suffix. That's also why `BucketNameValidationTest.createBucketAcceptsDotsInsideTheName` already asserts a name like `"trailing."` is valid. Windows' silent-stripping behavior for a terminal dot/space therefore never triggers on the persisted file, so there's no real gap to close.
3. **Three separate scans instead of one combined pass** - explicitly "not requesting a change," left as is; keeps each rejection reason independently readable, consistent with the method's existing structure.

No code changes this round. `build-and-package` is still finishing on CI; will confirm once green.

lvca:
Status update: all CI is green except \`ha-integration-tests\`, which failed on two different, unrelated tests across two runs:

- 1st run: \`Issue5569SlotMergeDeleteRaftIT#mergedDeletesReplicateIntact\`
- Rerun: \`RaftLargeSchemaReplicationIT#schemaJsonAbove64KbIsReplicated\`

Both are Raft/HA replication tests with no relationship to this PR's change (\`LocalSchema.checkValidBucketName\`, exercised only at bucket creation time). A different test failing on each run is the signature of this lane's known chronic flakiness on \`main\` (unrelated to any specific PR's content), not a regression introduced here. \`integration-tests\`, which also failed on the first pass, passed cleanly on rerun, confirming that one was transient too.

\`unit-tests\`, \`slow-unit-tests\`, \`vector-unit-tests\`, all e2e lanes, \`opencypher-tck-tests\`, \`builder-tests\`, \`build-and-package\`, and both Claude code review passes are green with no blocking issues.

PR #6139, opened 2026-08-13
fix(#6136): bound the last two CHECK DATABASE repair paths, and ship the index-rebuild WAL in instalments

### LINKED ISSUE #6136: CHECK DATABASE FIX: two repair paths are still unbounded after #6131, plus the autoFix breakdown and the nested-transaction cap
Follow-ups after #6131 (merged as `74059fe13`), which fixed points 1, 3, 4 and 5 of #6128.

**#6128 auto-closed on that merge because the PR body said `Closes #6128`, and GitHub has no partial close. Its point 2 was never fixed and is now tracked nowhere - that is item 1 below**, and the reason this issue exists rather than a comment on a closed one.

Items 1 and 2 are the same defect class and the one #6131 was about: a `CHECK DATABASE FIX` transaction with no bound on how large it can get, which on a replicated database is a Raft log entry with no bound on how large it can get. #6131 bounded the post-scan repair loops; these are the two paths it did not reach.

---

## 1. The index rebuild buffers its whole WAL in leader heap, and stalls every other writer while it does

`BucketIndexBuilder.create()` wraps the build in `schema.recordFileChanges(...)` (`BucketIndexBuilder.java:131`). On an HA leader that sets `isSchemaCommitThread` (`RaftReplicatedDatabase.java:1478`), which makes `commit()` **buffer** instead of replicate: the 5000-record batch commits inside `LSMTreeIndex.build()` accumulate in `schemaWalBuffer` (`RaftReplicatedDatabase.java:342`) and drain only when the callback returns. Peak leader heap is therefore roughly the whole rebuilt index, plus repeats of pages touched across batches.

The second half is worse than the heap. While that recording session is open, every *other* commit on the leader hits `waitForActiveRecordingSession()`, waits up to `HA_QUORUM_TIMEOUT` (10s default) and then **gives up and proceeds anyway**:

```java
HALog.log(this, HALog.BASIC,
    "commit waited %dms for FileManager recording session and gave up; proceeding with TX_ENTRY", timeout);
```

Proceeding is the #4083 ordering hazard: the TX_ENTRY can reach Raft before the SCHEMA_ENTRY that creates the file, and `TransactionManager.applyChanges` silently skips pages whose fileId does not exist on the follower. So concurrent writes during a FIX index rebuild can be **dropped on followers with no error**. `BucketIndexBuilder.create()` additionally parks every async executor thread for the duration.

**Suggested approach** - not a new protocol, but doing incrementally what `splitSchemaEntry` (#4743) already does at the end. Flush the buffer once it crosses a threshold: the first flush carries `filesToAdd`, intermediate flushes carry WAL only, the last carries `schemaJson` and `filesToRemove`. That is the same ordered-prefix contract `splitSchemaEntry` documents, spread over time instead of built up and split, so every follower prefix stays self-consistent - files exist before pages land in them, and a partially written new file is unreferenced bytes until the last entry publishes it.

This touches the code path behind #4083/#4743/#5492, where mistakes surface as silent follower divergence rather than as an exception, so it wants its own cluster test: rebuild an index large enough to need several flushes, kill the leader between them, and assert the followers hold either the old index or the new one and never a half of each.

## 2. The in-scan repairs are still unbounded

#6131 bounds the post-scan apply and delete loops, and deliberately does not touch anything inside a `scanType` callback: `LocalDatabase.scanType` holds the database read lock and owns an implicit transaction for the length of the scan, so committing under it would commit a transaction the scan believes it still owns.

That leaves two repairs unbounded, both writing once per defective edge from inside the scan:

- the back-reference fix-up in `checkIncomingEdges`/`checkOutgoingEdges` - `connectOutgoingEdge` at `GraphDatabaseChecker.java:818` and its `connectIncomingEdge` twin at `:1051`, each preceded by `inVertex.modify()`/`outVertex.modify()` and followed by `save()`;
- `resetChain(...)`, called from eight sites in the same two methods.

A database with a very large number of edges in the "not connected from the other side" state can therefore still accumulate one oversized transaction and hit exactly the `ReplicatedEntryTooLargeException` #6131 fixed elsewhere. It is narrower than the dangling references and broken chains the post-scan loops handle, which is why it was accepted rather than fixed there, and it is recorded in `commitRepairBatchIfFull()`'s javadoc - but recorded is not fixed.

**Suggested approach:** restructure those two checks into collect-then-apply, the shape `reconnectEdges` already uses - accumulate the vertices needing a back-reference fix during the scan, apply them after it, and let the existing `commitRepairBatchIfFull()` bound the apply loop. Worth checking the memory cost of the accumulator at the same time: `reconnectEdges` holds `List<Edge>` of every edge to reconnect, which is its own unbounded accumulation in heap rather than in the WAL, and a fix here should not add a second one.

## 3. `autoFix` counts repair actions, not corruption instances

After #6131 `autoFix` is the sum of records deleted and dangling adjacency entries pruned. Both are real repairs and both are real writes to distinct pages, so the number is not wrong - but one edge that is both listed in an adjacency chain and corrupt as a record contributes two, and an operator reading `autoFix` as "how many broken things were there" over-counts. A chain rebuild, meanwhile, contributes nothing at all: the `reconnected N incoming edges` case is reported only in the warnings.

So the one number an operator reads to decide whether a run did anything is a mixture of three repair kinds with different weights, one of which is invisible.

**Suggested approach:** keep `autoFix` as the action count for backward compatibility and add explicit per-kind counters (`deletedRecords`, `prunedDanglingEntries`, `reconnectedEdges`) so the report can be read without knowing how the arms decompose. Counting "defects fixed" instead would require the arms to agree on what one defect IS across a dangling entry, a corrupt record and a rebuilt chain, which they cannot without collapsing information the warnings currently carry - so this should add a breakdown, not redefine the total.

## 4. The checker sits exactly at the nested-transaction cap

`DatabaseContext.DatabaseContextTL.pushTransaction` throws once `transactions.size() + 1 > maxNested`, and `maxNested` is 3 (`DatabaseContext.java:387`). Running `CHECK DATABASE FIX` through HTTP now uses all three levels: the handler's transaction, `checkVertices`/`checkEdges`'s own, and the repair batch's.

It works today and is tested, but there is no headroom left: any future change that adds a nesting level anywhere in this call path turns `CHECK DATABASE FIX` over HTTP into a `TransactionException`, and the message ("Check your code if you are beginning new transactions without closing the previous one by mistake") points at a bug that would not be the real one.

**Suggested approach:** a test that pins the depth deliberately, so the failure arrives as a red build naming the cap rather than as a runtime error in a repair someone is running against a damaged production database. Raising `maxNested` is the wrong move - the limit is a real guard against leaked transactions, and the useful thing is knowing when this path consumes the last level of it.

---

Item 1 is the one with a correctness consequence for a running cluster (silently dropped follower pages) rather than only an availability one, and is the one I would take first.

lvca:
Closes #6136.

All four items, plus a correction: **item 4's premise is wrong**, and the useful part of it is a test either way.

---

## 1. The index rebuild buffered its whole WAL in leader heap

`BucketIndexBuilder.create()` wraps an index build in `schema.recordFileChanges(...)`, which on a Raft leader sets
`isSchemaCommitThread` so `commit()` **buffers** instead of replicating - the files being created do not exist on
the followers yet, and `TransactionManager.applyChanges` silently skips pages for a fileId it does not have.
`LSMTreeIndex.build()` then commits once per `IndexBuilder.BUILD_BATCH_SIZE` (5000) records, and every one of those
WAL images stayed in `schemaWalBuffer` until the callback returned. Peak leader heap was roughly the whole rebuilt
index plus a repeat of every page more than one batch touched - on the node that is, by construction, the leader.
`CHECK DATABASE FIX` drops and rebuilds every index on an affected bucket, which is exactly that shape.

The buffer is now shipped in ordered **instalments** as it fills. Heap is bounded by the threshold instead of by the
size of the index.

**Nothing new goes on the wire.** This is the ordered-prefix sequence `splitSchemaEntry` has produced since #4743,
emitted as the payload is produced rather than after it is built: files first so pages have somewhere to land, WAL
in the middle, and the publishing fields - schema JSON, `filesToRemove` - only in the session's final entry. Every
prefix a follower can be left holding is self-consistent, and a partially written file is unreferenced bytes until
the last entry publishes it. Followers need no new code because that is the sequence they already receive.

One difference from a `splitSchemaEntry` split carries the whole safety argument: **the last chunk of an instalment
is still marked `moreChunksFollow`**, because an instalment is not the end of the change. A follower that took it
for a publication would reload its schema from a half-delivered state, and #5443 measured what that costs - the
reload detaches a compacted sub-index it cannot resolve yet, the later real publication reuses the same in-memory
component, and the follower serves only its mutable pages from then on, silently and for good.

The threshold is **derived, not configured**: half the maximum replicated entry size, the same expression
`runWithCompactionReplication` already uses for its own chunk budget. No new setting. It is soft - tested between
buffered commits - so `replicateSchemaInstalment` can still split.

Two details worth calling out for review:

- **create-then-drop inside one session.** `FileManager.dropFile` CANCELS the recorded create when both happen in
  one session. An instalment that already announced the create would otherwise leave the file on the followers and
  nowhere else, so the final entry reconciles: what an instalment shipped and the session no longer lists as a
  create is added to `filesToRemove`.
- **the buffer is drained only after the entry commits.** It holds the only copy of those pages that can reach a
  follower, so clearing first would turn a replication error into silent divergence.

### The alternative I did not take, and why

The issue suggests incremental flushing, which is what this does. The bigger alternative - take the build **out** of
the recording session: one small SCHEMA_ENTRY that creates the files and marks the index unusable, the batches as
ordinary replicated TX entries, one small SCHEMA_ENTRY to publish - is strictly better on every axis: bounded heap
*and* no write lock held for the whole build *and* no async pool parked *and* no concurrent writer stalled. It is
also a change to what index creation means for embedded users (an index visible-but-unusable, and a leader crash
mid-build leaving one permanently so), on the code path whose entire history is silent follower divergence. It
wants its own issue and its own judgement, so this change buys the heap bound at no protocol risk and I have said
plainly in the javadoc what it does **not** buy:

**the recording session still stays open for the whole build**, so a concurrent writer on the leader still waits on
`waitForActiveRecordingSession` - and, before that, on the database write lock the callback holds. Bounding heap
does not shorten that window.

## 2. The in-scan repairs are bounded

#6131 bounded the post-scan loops and deliberately left the two repairs that write from *inside* the vertex scan,
where nothing can commit: `LocalDatabase.scanType` holds the database read lock, and the chunk iterator being walked
would not survive a commit taken under it. Those two are now planned during the scan and applied after it, through
the same page budget.

The **chain resets needed no accumulator at all**, which is the part I would draw a reviewer's eye to: every one of
the eight `resetChain` call sites already registered its vertex in `reconnectOutEdges`/`reconnectInEdges`, so those
sets already *were* the complete list of chains to drop. Deferring them costs zero extra memory.

The **back-reference fix-up** does need one, and the issue is right that it must not trade WAL growth for heap
growth: it accumulates three RIDs per defective edge in flat `int[]`/`long[]` arrays, 36 bytes an entry with no
per-object header. The `ArrayList<Edge>` the chain rebuild used to fill - a fully materialised record per entry,
properties included - was replaced by the same structure, so this removes more heap than it adds.

Three consequences of deferring, all deliberate:

- the apply loop re-tests its guards against the **final** contents of the corrupted-record and reconnect sets,
  which keep growing while the scan runs. Strictly better than the in-scan test: a back-reference planned for a
  vertex some later vertex registered for a full rebuild is now dropped instead of written and immediately
  discarded.
- the "current vertex points at ANOTHER vertex's list" head-chunk comparison now sees the shared chunk rather than
  the `null` an earlier reset had already written - which is the case that test exists to catch.
- a list naming the same edge **twice** would now plant two back-references, because the second entry no longer sees
  the first one's write. Closed with a per-list `EdgeIdentitySet` (the zero-boxing `RidHashSet` one, lazily
  allocated), scoped to the one list because that is the only scope in which the duplicate can arise - across
  vertices the planned entries differ in the endpoint they record, exactly as the probe distinguished them.

**One in-scan write remains, and I have named it rather than implied it**: the iterator `remove()` that prunes a
dangling entry. It is not deferrable the way the others were - the removal is made through the chunk iterator's live
position, and replaying it would turn a linear pass into a quadratic one - and it is the mildest of the three,
rewriting a chunk page the walk already holds and never allocating one, where the back-reference fix-up appended to
a far vertex's list and could allocate a chunk per repair.

## 3. `autoFix` keeps its meaning and gains a breakdown

`removedRecords`, `prunedDanglingEntries`, `reconnectedEdges`, always present, zero included. The first two sum to
`autoFix`; the third deliberately does not, because a rebuilt chain has never contributed to it and folding it in
would change every number a current run reports.

Named `removedRecords`, not the `deletedRecords` the issue proposed: the result already carries
`totalDeletedRecords` - records found in the DELETED state by the bucket scan, nothing to do with repair - and a key
one word away from it would read as its non-total sibling.

The stale javadoc on `DatabaseChecker.deleteCorruptedRecords`, which still described the graph arms as counting
before the delete and said to unify towards this one, is corrected: #6131 already unified them, the other way round.

## 4. The nesting cap: the premise is wrong, there are two levels to spare

The issue reports `CHECK DATABASE FIX` over HTTP as using all three of `DatabaseContext`'s nested-transaction
levels. It uses **one**. `CheckDatabaseStatement.executeSimple` opens by rolling the caller's transaction back, so
the handler's level is released before the check starts; `LocalDatabase.begin` pushes only when the last context is
*active*, so the arms reuse it; and `commitRepairBatchIfFull` commits and re-begins at that same level rather than
nesting under it.

The conclusion #6131 drew from the wrong premise still holds and is the one that mattered: these are real
`commit()` calls on a real `TransactionContext`, so each is a genuine WAL write and, under HA, a genuine
replication round trip. Nesting was never the mechanism for that.

The test the issue asks for is worth having either way, and is here in two forms: sampling the depth the FIX path
actually reaches, and running the whole statement under `maxNested = 1`. A future change that adds a level arrives
as a red build naming the cap, instead of as a `TransactionException` in a repair someone is running against a
damaged production database - whose message points at a bug that would not be the real one.

---

## Tests

| test | proves |
| --- | --- |
| `CheckDatabaseBackReferenceRepairTest` | the back-reference repair splits across transactions under a small budget, is one transaction with the budget disabled, repairs identically either way, and is counted |
| `CheckDatabaseRepairBreakdownTest` | the breakdown decomposes `autoFix` on three damage shapes, and a rebuilt chain is counted without entering the sum |
| `CheckDatabaseTransactionNestingTest` | the FIX path runs at one transaction level, and survives a cap of 1 |
| `Issue6136SchemaWalInstalmentTest` | every chunk of an instalment is delivery-only and carries nothing that publishes; a split of a *complete* change still publishes on its last chunk |
| `RaftSchemaWalInstalment3NodesIT` | a 3-node rebuild under a 128KB append buffer ships instalments and leaves all three nodes with a complete, queryable index |

Each new test was **mutation-checked against unmodified code**: `CheckDatabaseBackReferenceRepairTest` fails 2 of 3
(the batching assertion reads `commits = 1`), and `RaftSchemaWalInstalment3NodesIT` fails on the instalment count
alone with the flush disabled - every other assertion in it passes on the old code, which is stated in the test so
nobody mistakes it for coverage it does not give.

Regression: the full engine unit lane, plus the ha-raft schema/index ITs.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1202,7 +1202,9 @@ costs - the reload detaches a compacted sub-index it cannot resolve yet, the lat
 in-memory component, and the follower serves only its mutable pages from then on, silently and permanently.
 
 The threshold is derived rather than configured: half the maximum replicated entry size, the same expression the
-compaction path already uses for its own chunk budget. Lowering `arcadedb.ha.appendBufferSize` lowers it too.
+compaction path already uses for its own chunk budget. Lowering `arcadedb.ha.appendBufferSize` lowers it too, and
+that is worth knowing before tuning it down: a smaller buffer means more instalments per rebuild, and therefore more
+quorum round trips taken under the write lock (see below).
 
 **What it costs**, stated precisely because it is a trade and not only a heap win. The recording session stays open
 for the whole build, so a concurrent writer on the leader still waits on it - and, before that, on the database write
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1204,10 +1204,25 @@ in-memory component, and the follower serves only its mutable pages from then on
 The threshold is derived rather than configured: half the maximum replicated entry size, the same expression the
 compaction path already uses for its own chunk budget. Lowering `arcadedb.ha.appendBufferSize` lowers it too.
 
-**What this does not change**, stated because the difference matters to an operator: the recording session stays
-open for the whole build, so a concurrent writer on the leader still waits on it - and, before that, on the database
-write lock the callback holds. Bounding the heap does not shorten that window. Shortening it means taking the build
-out of the session altogether, which is a different change on the same code path.
+**What it costs**, stated precisely because it is a trade and not only a heap win. The recording session stays open
+for the whole build, so a concurrent writer on the leader still waits on it - and, before that, on the database write
+lock the callback holds. Bounding the heap does not shorten that window, and each instalment *lengthens* it: a quorum
+round trip is now taken while that write lock is held, where the single final entry had always been submitted after
+the callback returned and the lock was released. Normally that is milliseconds against a build measured in minutes;
+against a slow or briefly partitioned quorum member it is up to `arcadedb.ha.quorumTimeout` per instalment, and the
+whole database's writers wait it out. Accepted, because the alternative is not "no round trips" - without instalments
+the one final entry carries the whole index, `splitSchemaEntry` splits it, and the same number of round trips happens
+anyway, after the lock but only once the leader has held the entire rebuilt index in heap. Making the window itself
+shorter means taking the build out of the recording session altogether, which is a different change on the same code
+path.
+
+**If the build fails after an instalment has shipped**, the followers are holding files and pages that the entry
+which would have published or retired them never reaches, because it lives after the line that threw - and every
+instalment chunk is marked `moreChunksFollow`, so there is no abandonment signal a follower could act on by itself.
+The session now sends a compensating removal, and *which* files it retires is the whole correctness of it: one the
+leader no longer has is retired (the case `BucketIndexBuilder.create()` produces, since it drops the half-built index
+from its own error handler), while one the leader still has is left alone and reported, because retiring that would
+take a state both sides agree on and make them disagree.
 
 ### The in-scan repairs are bounded too
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1168,3 +1168,92 @@ torn across two instants. Incremental resync belongs at the **page** level, on t
 operator diagnostic.
 
 [#6125](https://github.com/ArcadeData/arcadedb/issues/6125)
+
+## `CHECK DATABASE FIX`: the last two unbounded repair paths, and what the numbers it reports mean (#6136)
+
+#6131 bounded the repair transaction a `CHECK DATABASE ... FIX` builds, committing every
+`arcadedb.checkDatabaseRepairBatchPages` dirtied pages so that a large repair on a replicated database no longer
+runs for hours and then dies whole at the commit with `ReplicatedEntryTooLargeException`. It deliberately reached
+only the post-scan loops. These are the two paths it did not, plus the reporting the split made necessary.
+
+### The index rebuild no longer buffers its whole WAL in leader heap
+
+`BucketIndexBuilder.create()` wraps an index build in `schema.recordFileChanges(...)`, which on a Raft leader marks
+the thread so that `commit()` **buffers** instead of replicating: the files being created do not exist on the
+followers yet, so a `TX_ENTRY` referring to them would have its pages silently dropped there. `LSMTreeIndex.build()`
+then commits once per `IndexBuilder.BUILD_BATCH_SIZE` (5000) records, and every one of those WAL images stayed
+resident until the callback returned. Peak leader heap was therefore roughly the whole rebuilt index, plus a repeat
+of every page more than one batch touched - on the node that is, by construction, the cluster's leader. A
+`CHECK DATABASE FIX` over a large damaged type drops and rebuilds every index on the affected buckets, which is
+exactly that shape.
+
+The buffer is now shipped in ordered **instalments** as it fills, so leader heap is bounded by the threshold rather
+than by the size of the index. Nothing new goes on the wire: this is the ordered-prefix sequence `splitSchemaEntry`
+has produced since #4743, emitted as the payload is produced rather than after it has been built in full. Files
+first, so pages have somewhere to land; WAL in the middle; and the fields that *publish* the change - the schema
+JSON and the files to retire - only in the session's final entry. Every prefix a follower can be left holding is
+therefore self-consistent, and a partially written new file is unreferenced bytes until the last entry publishes
+it. Followers need no new code, because that sequence is the one they already receive.
+
+The one difference from a `splitSchemaEntry` split carries the whole safety argument: the last chunk of an
+instalment is still marked `moreChunksFollow`, because an instalment is not the end of the change. A follower that
+mistook it for a publication would reload its schema from a half-delivered state, and #5443 measured what that
+costs - the reload detaches a compacted sub-index it cannot resolve yet, the later real publication reuses the same
+in-memory component, and the follower serves only its mutable pages from then on, silently and permanently.
+
+The threshold is derived rather than configured: half the maximum replicated entry size, the same expression the
+compaction path already uses for its own chunk budget. Lowering `arcadedb.ha.appendBufferSize` lowers it too.
+
+**What this does not change**, stated because the difference matters to an operator: the recording session stays
+open for the whole build, so a concurrent writer on the leader still waits on it - and, before that, on the database
+write lock the callback holds. Bounding the heap does not shorten that window. Shortening it means taking the build
+out of the session altogether, which is a different change on the same code path.
+
+### The in-scan repairs are bounded too
+
+Two repairs used to write from *inside* the vertex scan, where nothing can commit: `LocalDatabase.scanType` holds
+the database read lock for the length of the scan, and the chunk iterator being walked would not survive a commit
+taken under it.
+
+- the back-reference fix-up - one write to a **far** vertex per edge found "not connected from the other side",
+  each able to allocate a fresh edge-list chunk;
+- `resetChain`, which drops an unreadable adjacency chain so it can be rebuilt from the surviving edge records.
+
+Both are now *planned* during the scan and applied after it, through the same page budget as everything else. The
+chain resets needed no new state at all: every `resetChain` call site already registered its vertex in one of the
+two reconnect sets, so those sets were the complete list. The back-reference fix-up accumulates three RIDs per
+defective edge in flat `int[]`/`long[]` arrays - and the `ArrayList<Edge>` the chain rebuild used to fill, which
+held a fully materialised record per entry, was replaced by the same structure.
+
+One in-scan write remains and is named rather than implied: the iterator `remove()` that prunes a dangling
+adjacency entry. It is not deferrable the way the others were - the removal is made through the chunk iterator's
+live position, and replaying it afterwards would turn a linear pass into a quadratic one - and it is the mildest of
+the three, rewriting a chunk page the walk is already reading and never allocating one.
+
+### `autoFix` now comes with a breakdown
+
+`autoFix` is a count of repair **actions**, not of corruption instances: after #6131 it is records deleted plus
+dangling adjacency entries pruned, so one edge that is both listed in a chain and corrupt as a record contributes
+two. A rebuilt chain, meanwhile, contributed nothing at all and was visible only in the warnings. The one number an
+operator reads to decide whether a run did anything was a mixture of three repair kinds with different weights, one
+of them invisible.
+
+`autoFix` keeps its meaning - existing readers and existing expectations depend on it - and the result now also
+carries `removedRecords`, `prunedDanglingEntries` and `reconnectedEdges`. The first two sum to `autoFix`; the third
+is deliberately not in that sum, since folding a rebuilt chain in would change every number a current run reports.
+All three are always present, zero included. The first is named `removedRecords` rather than `deletedRecords`
+because the result already carries `totalDeletedRecords` - records found in the DELETED state by the bucket scan,
+nothing to do with repair - and a key one word away from it would read as its non-total sibling.
+
+### The transaction-nesting headroom, measured rather than assumed
+
+#6136 also reported the FIX path as sitting exactly on `DatabaseContext`'s hardcoded limit of 3 nested
+transactions - the HTTP handler's, the arm's own, and the repair batch's - leaving no headroom. It does not.
+`CheckDatabaseStatement.executeSimple` opens by rolling the caller's transaction back, so the handler's level is
+released before the check starts; the arms' `begin()` then finds the last context inactive and reuses it rather
+than pushing; and the repair batch commits and re-begins at that same level. The path runs at **one** level with
+two to spare, and is now pinned both by sampling the depth it reaches and by running it under a cap of 1, so that a
+future change adding a level arrives as a red build naming the cap rather than as a `TransactionException` in a
+repair someone is running against a damaged production database.
+
+[#6136](https://github.com/ArcadeData/arcadedb/issues/6136)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1262,6 +1262,14 @@ All three are always present, zero included. The first is named `removedRecords`
 because the result already carries `totalDeletedRecords` - records found in the DELETED state by the bucket scan,
 nothing to do with repair - and a key one word away from it would read as its non-total sibling.
 
+**One key pair is gone, which is a non-additive change to that map.** `GraphDatabaseChecker.checkVertices` used to
+put the `List<Edge>` of reconnected edges under `outEdgesToReconnect`/`inEdgesToReconnect`; both keys are removed
+and `reconnectedEdges` carries the count instead. Nothing in the repository read them, and they never reached the
+`CHECK DATABASE` command result at all - `DatabaseChecker.updateStats` folds only `Long` values - so an operator
+reading the SQL output never saw them. Only a caller invoking `checkVertices` directly as an API could have, and
+holding a fully materialised `Edge` per reconnected entry was one of the unbounded-heap sources this change exists
+to remove, so they are not coming back.
+
 ### The transaction-nesting headroom, measured rather than assumed
 
 #6136 also reported the FIX path as sitting exactly on `DatabaseContext`'s hardcoded limit of 3 nested
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1302,7 +1302,11 @@ public enum GlobalConfiguration {
       ones (binary blobs, base64, encrypted fields, float vectors) map roughly 1:1. Raise it when single \
       transactions or records are bigger than the default, and raise arcadedb.ha.writeBufferSize with it \
       (it must stay >= this value + 8 bytes). Cost of raising it: a directly-allocated write buffer of \
-      writeBufferSize per server, plus up to this many bytes of heap per follower appender during catch-up.""",
+      writeBufferSize per server, plus up to this many bytes of heap per follower appender during catch-up. \
+      Cost of LOWERING it (issue #6136): an index rebuild ships its WAL to the followers in instalments of half \
+      this size, and each instalment is a quorum round trip taken while the database write lock is held, so a \
+      smaller value means more round trips and a longer window in which every writer on that database waits - \
+      up to arcadedb.ha.quorumTimeout per instalment if a quorum member is slow or briefly partitioned.""",
       String.class, "32MB"),
 
   HA_APPEND_ELEMENT_LIMIT("arcadedb.ha.appendElementLimit", SCOPE.SERVER,
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -127,6 +127,23 @@ public class DatabaseChecker {
       LogManager.instance().log(this, Level.INFO, "Integrity check of database '%s' started", null, database.getName());
 
     result.put("autoFix", 0L);
+    // Issue #6136 (3): the breakdown behind autoFix, seeded so the three keys are always present rather than
+    // appearing only on the runs that happened to perform that kind of repair.
+    //
+    // autoFix stays the ACTION count it has always been - removedRecords + prunedDanglingEntries - because
+    // existing readers and existing tests depend on it, and because "how many defects were there" is not a
+    // question the arms can answer jointly: one edge that is both listed in an adjacency chain and corrupt as a
+    // record is one defect, two repairs, two writes to two distinct pages. reconnectedEdges is NOT in the sum, for
+    // the same backward-compatibility reason - a rebuilt chain has never contributed to autoFix and making it do
+    // so would change every existing number. Read the three together to learn what a run did.
+    //
+    // "removedRecords" rather than the "deletedRecords" the issue proposed, deliberately: the result already
+    // carries "totalDeletedRecords", which counts records found in the DELETED state by the bucket scan and has
+    // nothing to do with repair, and a key differing from it by one word would read as its non-total sibling. This
+    // one is the count of "deletedRecordsAfterFix", the list of what this run actually took out.
+    result.put("removedRecords", 0L);
+    result.put("prunedDanglingEntries", 0L);
+    result.put("reconnectedEdges", 0L);
     result.put("invalidLinks", 0L);
     result.put("warnings", new LinkedHashSet<>());
     result.put("deletedRecordsAfterFix", new LinkedHashSet<>());
@@ -557,14 +574,13 @@ public class DatabaseChecker {
   }
 
   /**
-   * Removes the records this pass flagged corrupted, counting them into {@code autoFix}.
+   * Removes the records this pass flagged corrupted, counting them into {@code autoFix} and {@code removedRecords}.
    * <p>
-   * NOT quite what {@code GraphDatabaseChecker}'s vertex and edge arms count, and the difference is deliberate:
-   * they call {@code autoFix.incrementAndGet()} BEFORE attempting the delete, so a record they then fail to remove
-   * is still reported as fixed. This one counts a repair only once the delete and its counter delta have both
-   * succeeded, and counts nothing for a record that was already gone. If the two are ever unified, unify them
-   * towards THIS one: {@code autoFix} is what an operator reads to decide whether the run did anything, and a
-   * number that includes failed deletes cannot answer that.
+   * Counts a repair only once the delete and its counter delta have both succeeded, and counts nothing for a record
+   * that was already gone. {@code GraphDatabaseChecker}'s vertex and edge arms used to differ here - they counted
+   * BEFORE attempting the delete, so a record they then failed to remove was still reported as fixed - and #6128
+   * aligned them on this one: {@code autoFix} is what an operator reads to decide whether the run did anything, and
+   * a number that includes failed deletes cannot answer that.
    * <p>
    * This is not merely tidiness, and it is not optional. A corrupted record's RID goes into
    * {@code corruptedRecords}, which is what {@link #check()} turns into {@code affectedBuckets}, and every index on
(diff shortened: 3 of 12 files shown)


lvca:
Thanks - the orphaned-follower-state point was real, and reproducing it confirmed it: with the compensation disabled, a follower is left holding a file the node that ran the DDL had already dropped.

Fixed in `a91893438` rather than only logged. `recordFileChanges` now tracks whether it reached its final entry, and a session that did not sends a compensating removal.

**Which files it retires is the whole correctness of it**, and the obvious version - undo every announcement - swaps one divergence for another:

- a file the leader **no longer has** is retired. This is the case that matters and the one the principal caller produces: `BucketIndexBuilder.create()` drops the half-built index from its own error handler, so the leader has let the file go and only the followers still hold it.
- a file the leader **still has** is left alone and reported. A DDL that throws can leave the leader holding a created-but-unpublished file (the schema was never saved, so nothing references it there either), and retiring it on the followers would take a state both sides agree on and make them disagree. That leader-side orphan predates instalments; this does not try to repair it.

Pages an instalment applied to pre-existing files are left alone for the same reason - `commit()` ran `commit2ndPhase` before buffering, so the leader holds them too.

The compensation cannot throw: it runs from a `finally` that is usually unwinding the build's own exception, and masking that would replace a diagnosable failure with a confusing one. A compensation that itself fails is logged at SEVERE naming every file it could not retire.

New test `RaftSchemaWalInstalment3NodesIT.aSessionThatFailsAfterAnInstalmentRetiresWhatItAnnounced` reproduces the shape exactly - create a file, ship more than one instalment's worth of WAL, drop the file the way the index builder does, throw - and asserts the invariant that matters: **every node agrees on whether the file exists**, not merely that the followers dropped it. The narrower assertion would have been satisfied by a compensation that retired a file the leader had kept, i.e. by the divergence the retirement exists to prevent. Mutation-checked: with the compensation disabled, server 0 holds file 30 and the DDL node does not.

Also took the free part of the perf note: the instalment threshold is now resolved once per session into `SchemaInstalmentState` instead of on every buffered commit, which takes the `requireRaftServer().getTransactionBroker()` hop off the batch-commit path of an index build.

Two earlier commits address Codacy's complexity finding by extracting `applyRepairPlan()` (`203773333`) and `reconcileInstalmentFiles()` (`c9515910f`); both are pure extractions with no behaviour change.

On the global-config note in the tests: agreed it's the pre-existing pattern from #6131's siblings, left as-is for consistency.

lvca:
Second round addressed in `0b745557b`. Two of the three were bugs, and one of them was a good catch on a path I would not have found from the tests.

**#2 (bare `RID` in `applyPendingLinks`) - real bug, fixed at the source.** You are right that flattening to primitives loses the owning database and that `resolveActiveDatabase()` throws with more than one database in scope, and right about the consequence: the generic handler would turn it into a per-entry warning, so a multi-database server would silently decline the repair, under-report `reconnectedEdges`, and still report success. Rather than fix the two call sites, `PendingLinks` now holds the database and hands back `database.newRID(...)`, so no future caller can reintroduce it; the two loads that had the same shape go through `database.lookupByRID`. Worth noting for the record that the *set* lookups were never affected - `RID.equals`/`hashCode` compare bucket and offset only, and `DatabaseRID` inherits them unchanged.

**#3 (condemned chain re-walked) - real, and the answer to your "worth double-checking" is yes.** An entry pointing at an edge that belongs to another vertex is diagnosed as a corrupt *edge* and deleted in fix mode, which destroys a record the rebuild would have re-linked. Both checks now skip a vertex already registered for a rebuild. That restores the old behaviour and improves on it: the old skip was an accident of scan order, since the same vertex reached *before* any far-endpoint probe was walked in full either way.

**#1 (round trips under the write lock) - accepted, and now documented as a cost rather than a non-change.** You are right that my "bounding heap does not shorten the window" undersold it: each instalment *lengthens* it, because `submitAndWait` is a quorum round trip taken while the write lock is held, where the single final entry was always submitted after the lock was released. Against a slow quorum member that is up to `arcadedb.ha.quorumTimeout` per instalment with every writer on the database waiting. Signing off on it, because the alternative is not "no round trips": without instalments the one final entry carries the whole index, `splitSchemaEntry` splits it, and the same number of round trips happens anyway - after the lock, but only once the leader has held the entire rebuilt index in heap, which is the failure this exists to prevent. Paying for a bounded heap with round trips inside a window that is already fully exclusive is the better side of that trade. The javadoc and the release note now say exactly this.

**#4 (eager `requireRaftServer()`) - fixed.** The threshold is resolved on the session's first *buffered commit* instead of when the session opens, which restores the previous laziness: a session that buffers nothing no longer needs the Raft server, and one that does needs it to replicate anyway.

**#5 (`published` behind the diagnostic) - fixed**, moved to immediately after `replicateSchema` returns. Your read is right that it was log-noise only, and also right that the noise is the expensive kind.

Two more from the previous round rolled in: the nested-session `bufferedBytes` desync is zeroed with the shared buffer it describes (with a note that this does *not* make that nesting safe - the outer frame's WAL is destroyed by the clear, which is pre-existing), and the duplicated back-reference de-dup block is now `planBackReference()`.

**Not changed:** recomputing the new-file list per instalment. It is O(n²) in *file changes*, which for an index rebuild is one or two entries and is not proportional to the WAL volume that drives the flush count, so the quadratic term never grows. An index cursor into `recordedChanges` would also be unsafe, since `dropFile` removes entries from the middle of that list.

Verified: 11762 engine unit tests - one unrelated pre-existing flake, `TimeSeriesTagDictionaryTest.aSecondWaveOfIdsIsAlsoResolvedByReloading`, which passes in isolation against both modified and unmodified sources and is in a subsystem this PR does not touch - plus the ha-raft schema/index ITs (`RaftSchemaWalInstalment3NodesIT`, `RaftCheckDatabaseFix3NodesIT`, `RaftCheckDatabaseLargeRepairIT`, `Issue4743CompactionEntrySizeIT`, `Issue5443SplitCompactionEntryIT`).

lvca:
All three nits fixed in `a7221859c`:

1. `Callable` is imported instead of fully qualified. It came in by copy from `CheckDatabaseRepairBatchTest`, which still has the FQN; left alone there as pre-existing rather than widening the diff.
2. Both `published = true` assignments now point at each other, with the reason each is load-bearing: the inner one covers a throw in the diagnostic logging that follows `replicateSchema`, the outer one covers the path where the block is not entered at all.
3. The tuning implication is now stated where an operator tuning `arcadedb.ha.appendBufferSize` would look: lowering it lowers the derived instalment threshold, so a rebuild ships more instalments and takes more quorum round trips under the write lock. Good catch - that direction of the dependency was implicit and it is the one that bites.

On Codacy's remaining medium complexity finding: I have already extracted `applyRepairPlan()` and `reconcileInstalmentFiles()` in response to it, and it has not moved. Its detail is only visible inside Codacy, and the remaining candidates are `checkIncomingEdges`/`checkOutgoingEdges` - two ~250-line methods that predate this PR and whose branch count this PR *reduces* (the back-reference block moved out to `planBackReference`, and only the one early-return guard was added). Splitting either of them is a large, risky refactor of the deferral logic this round just validated, for a metric rather than a defect, so I am leaving it and noting it here deliberately rather than silently.

Verified after these changes: 87 CHECK DATABASE tests, `Issue6136SchemaWalInstalmentTest`, `RaftSchemaWalInstalment3NodesIT`.

lvca:
Both non-blocking asks taken, in `874cc834f`.

**Ops-facing doc** - you asked whether `HA_QUORUM_TIMEOUT` tuning guidance exists anywhere. It does, in the shape this codebase uses for it: the `GlobalConfiguration` entry's own description, which for `arcadedb.ha.appendBufferSize` already documents the cost of *raising* it. The cost of **lowering** it is now there too, since that is the direction this PR adds: a smaller buffer lowers the derived instalment threshold, so a rebuild ships more instalments and takes more quorum round trips under the write lock, up to `arcadedb.ha.quorumTimeout` each against a slow member. It was only in the javadoc and the release notes, which is not where someone tuning the knob looks.

**Per-instalment rescan** - comment added at the call site saying why it is fine for this caller and what would make it not fine: an index rebuild records one or two file changes however many instalments its WAL volume produces, so the second factor does not grow with the first; a future DDL creating many files through this buffered path would make it quadratic and should carry the shipped/unshipped split forward. It also records why the obvious fix is wrong - an index into `recordedChanges` cannot work, because `dropFile` removes entries from the middle of it.

Also noted for the record, since you traced it: you are right that the second `published = true` is belt-and-suspenders rather than load-bearing, because `retireAbandonedInstalments` no-ops at `instalments == 0` anyway. Kept explicit, and the comments now say which case each assignment covers, for the reason you give - this module's failures are silent.

And agreed on `schemaInstalmentsShipped` being process-wide: it is fine as test observability, and if it is ever promoted to a Micrometer metric it needs a per-database dimension first. Worth a follow-up issue if that promotion is ever wanted rather than a change here.

Verified after these changes: 75 tests across `ConfigurationTest` and the CHECK DATABASE suites.

lvca:
Both fixed in `2182fc089`, and #1 was a real gap - thank you for tracing it.

**#1 mid-split failure.** You are right that `replicateSchemaInstalment` is not necessarily one entry, and right that the path is reachable rather than defensive: the threshold is soft, so a single batch commit can carry more WAL than the whole budget. It is reachable *in fact*, not only in theory - the ha-raft ITs log `does not fit one Raft entry ... shipping it as 4 ordered entries` four times in a run, so the split path is already being exercised by the existing coverage, just never with a failure injected into it.

And your two consequences were both real, the second worse than the first: on a *first* instalment `state.instalments` was still 0, so `retireAbandonedInstalments` early-returned without even logging - followers holding an orphaned file and nothing said about it. That is precisely the "follower holds bytes the leader has forgotten about" divergence the compensation exists to close, reintroduced through the compensation's own bookkeeping.

Fixed by recording the intent **before** the send rather than after, which makes the bookkeeping pessimistic: it now says "the followers *may* hold these", which is what a compensation needs. Over-retiring costs nothing - I checked rather than assumed: the follower's `FileManager.dropFile` takes `fileIdMap.get(fileId)` under a `!= null` guard, so retiring a file a follower never received is a no-op there.

The WAL buffer deliberately keeps the opposite order, and the comment now states why the two differ: the buffer holds the only copy of those pages that can reach a follower, so clearing it before a failure would turn a replication error into silent divergence. **Announcing too much costs a no-op removal; forgetting WAL costs a diverged follower.**

On coverage for the mid-split failure specifically: it needs an injected failure on a non-first chunk inside `splitSchemaEntry`'s submit loop, which there is no seam for today without adding one to `RaftGroupCommitter` purely for the test. Rather than add that seam, the fix removes the need for it - the bookkeeping no longer depends on where inside the instalment the failure lands, because it is recorded before any chunk is sent. Worth saying explicitly so it is a deliberate gap and not an oversight.

**#2 misplaced javadoc.** Correct, and my fault: the `planBackReference` extraction two commits ago was inserted between `applyPendingChainResets`'s javadoc and its declaration, leaving one method undocumented and the other with an orphaned floating comment. Moved onto its method.

Verified: 87 CHECK DATABASE tests, `Issue6136SchemaWalInstalmentTest`, `Issue4743OversizedRaftEntryTest`, `RaftSchemaWalInstalment3NodesIT`, `RaftCheckDatabaseLargeRepairIT`.

lvca:
All three taken, in `98d1c0e1c`.

**Lighter test for the bookkeeping** - agreed, and it was the right thing to ask for: the two rules are map arithmetic spread over the life of a session, and the only coverage was a `@Tag("slow")` three-node test whose failure says "the nodes disagree" rather than which rule got it wrong. `Issue6136InstalmentFileBookkeepingTest` pins both in the fast lane, including the two you named: `reconcileInstalmentFiles` for the create-then-drop case (the one that exists *because* `FileManager.dropFile` cancels the recorded create, so the final entry's own maps say nothing about a file the followers were told to create), and both halves of the leader-owns-the-truth split. That second one needed `partitionAbandonedFiles` extracted out of `retireAbandonedInstalments`, which is a better shape anyway - it is now side-effect-free and needs no cluster. `reconcileInstalmentFiles` takes the map rather than the session state, so no field on the state class had to widen for a test to reach it.

**Defensive comment on `published`** - added, aimed at exactly the failure you describe: everything between the two assignments is code that can throw *after* the change is published, the flag decides whether the finally block retires the instalments' files, and new code there belongs on a deliberately chosen side of it.

**Tracking issue for the rescan** - filed as #6142 and referenced from the comment, so the self-imposed follow-up cannot get lost. It records why the obvious fix is wrong, too (an index into `recordedChanges` is not stable, since `dropFile` removes entries from the middle of it).

On the repeated `lookupByRID` in `applyPendingLinks`: correct reading, and worth stating the bound. It is one lookup per *pending link*, not per vertex, so the pathological case you name - one vertex with thousands of dangling back-references - re-resolves the same RID thousands of times. Those hit the transaction's record cache rather than the disk, and the loop is a repair on an already-damaged database in a maintenance window, so I have left it; if it ever shows up in a profile the fix is to sort the plan by vertex and hold the `MutableVertex` across a run of entries, which the flat-array layout makes easy.

Verified: `Issue6136InstalmentFileBookkeepingTest`, `Issue6136SchemaWalInstalmentTest`, `Issue4743OversizedRaftEntryTest`, `RaftSchemaWalInstalment3NodesIT`.

lvca:
All addressed in `205feb676`.

**The threshold vs. its stated invariant** - correct, and a good catch: the javadoc claimed "first *buffered* commit" while `flushSchemaWalBufferIfFull` runs after *every* commit in the session, including ones where `commit1stPhase` returned null. Whichever ran first resolved it, which is exactly the no-op-session case the laziness exists for. Fixed by tightening the guard rather than the comment, as you suggested: `bufferedBytes == 0` is now tested *before* the resolution. It costs nothing, since zero bytes can never reach a positive threshold anyway, and it happens to cover the just-flushed case too.

**Duplicated chunk-budget expression** - agreed, and it was the drift risk this PR argues about elsewhere while committing it. Now `RaftTransactionBroker.walChunkBudget()`, called by both producers.

**Duplicated repair-counter block** - collapsed into `putRepairCounters()`, which also puts the `autoFix` sum and its two components in one place, so a reader never has to check whether the two arms still agree. (The seed in `DatabaseChecker` stays literal - it is in a different package and making the key names public to share them would be a wider change than the duplication costs.)

**Dangling `{@link #reconnectEdges}`** - fixed, repointed at `collectEdgesToReconnect`, which is the method that sentence is actually about. I swept the file for other dangling links while I was in there; the rest resolve.

**Removed `outEdgesToReconnect`/`inEdgesToReconnect` keys** - you are right that this deserved a mention rather than being silent. Added to the release notes, including the detail that they never reached the `CHECK DATABASE` command result at all (`updateStats` folds only `Long` values), so only a caller invoking `checkVertices` as an API could have seen them.

Separately, and worth flagging since it is in this PR's own new test: `RaftSchemaWalInstalment3NodesIT` lost a local run to a leadership flap between `findLeaderIndex()` and the `CREATE INDEX` after it, failing in *setup* with `ServerIsNotTheLeaderException` and reporting a flake that said nothing about what the test covers. Its DDL now re-resolves the leader and retries, which is the designed response to a `NeedRetryException`. I would rather this test not add to the noise in that lane.

On the two you flagged as low priority, both declined with reasons recorded in the commit: `applyPendingLinks` resolving per link rather than per vertex hits the transaction's record cache rather than disk, and the fix (sort the plan by vertex, hold the `MutableVertex` across a run) is easy on the flat-array layout if a profile ever asks for it; and `PendingLinks` vs `GraphBatch`'s similar idiom serve different lifecycles and access patterns, so I left them independent rather than forcing a shared abstraction over two different shapes.

Verified: 87 CHECK DATABASE tests, `Issue6136InstalmentFileBookkeepingTest`, `Issue6136SchemaWalInstalmentTest`, `Issue4743OversizedRaftEntryTest`, `RaftSchemaWalInstalment3NodesIT`, `RaftCheckDatabaseLargeRepairIT`.

lvca:
Addressed in `1dae82ec5`. You were right that this was the one place where the documentation was thinner than everywhere else, and it is the most Raft-idiomatic way for the path to fail.

**Named it, and named what changed.** The javadoc now says explicitly: `recordFileChanges` checks `isLeader()` once at the top, nothing re-checks per instalment, and `requireRaftServer()` only asserts a server exists rather than that this node still leads - so a normal failover, partition or step-down after an instalment has shipped means the compensating removal fails too. And it states the delta plainly, which the previous framing did not: before instalments that case cost *nothing*, because nothing had reached a follower and the final `replicateSchema` simply failed with a retryable error. The trade is bounded leader heap against a new operational cleanup case.

**Deliberately not "solved", with the reason recorded.** Re-checking leadership per instalment narrows the window without closing it - leadership can move between the check and the submit - and a node that has lost the term cannot make the cluster do anything, so compensation stops being its job the moment it is no longer leader. The real fix is the *new* leader reclaiming files no schema references, which is storage-layer garbage collection this database does not have (`reclaimOrphanedEdgeSegments` does the analogous thing for edge-list segments, but nothing does it for paginated component files).

**Filed as #6143**, and referenced from the javadoc, covering both halves: the missing reclaim, and your point that the SEVERE branch is reasoned about rather than covered. You are right that it is hard to construct deterministically - the step-down has to land between an instalment and the unwind - which is exactly why it is filed rather than written.

**The SEVERE message now says when lost leadership was the cause**, since "retry the operation" is the wrong advice in that case: this node cannot act on the cluster any more, so the files stay until somebody removes them.

Stray blank line in `SchemaInstalmentState` removed.

On the Micrometer suggestion: leaving `schemaInstalmentsShipped` test-only for now, deliberately. It is JVM-wide across all databases, and a previous round flagged the same thing - promoting it to a production metric wants a per-database dimension first, which is a change to the counter rather than just wiring it up. Worth its own issue if the observability is wanted; I have not filed one, since nobody has asked for the metric yet and I would rather not file speculative work.

Verified: `Issue6136InstalmentFileBookkeepingTest`, `Issue6136SchemaWalInstalmentTest`, `Issue4743OversizedRaftEntryTest`, `RaftSchemaWalInstalment3NodesIT`.

lvca:
Thanks. No code changes this round - here is where each point lands.

**1. Write lock / quorum round trips - this one is yours to sign off, and it is the right thing to ask.** I cannot settle it from inside the PR, so I have made sure it is impossible to miss: it is on `flushSchemaWalBufferIfFull`'s javadoc, in `arcadedb.ha.appendBufferSize`'s own configuration description (the ops-facing surface, and specifically under the cost of *lowering* it, since that multiplies the instalment count), and in the release notes. Stated plainly for the decision: a large `CHECK DATABASE FIX` against a slow or briefly partitioned quorum member can now stall every writer on that database for up to `arcadedb.ha.quorumTimeout` **per instalment** instead of once at the end. What is bought is that the leader no longer holds an entire rebuilt index in heap. @lvca - this is the one behavioural change here that is not purely a win, so it wants your call before merge.

**2. Metrics - filed as #6144** rather than declined a third time. It captures your point that *duration* is the number that matters, not count, since duration is what eats the quorum-timeout budget and what a stalled writer is actually waiting on; plus the prerequisite that the existing counter is JVM-wide across databases and needs a per-database dimension before it can be exported meaningfully; plus the cheaper middle option of a single INFO summary line per session so the common case needs no metrics backend at all.

**3.** #6142 and #6143 - agreed on both readings, including that #6143 is a strict improvement over doing nothing and that leader-side unreferenced-file GC belongs in its own issue rather than jammed in here.

**4. Javadoc weight** - noted, and it is a fair cost to flag. The file's existing `#4083`/`#5777` blocks set the precedent, and on this particular code path - whose failures are silent follower divergence rather than exceptions - I would rather the next editor have too much context than too little. But you are right that it is now a maintenance obligation.

**On the untested lazy-threshold path**: leaving it. `requireRaftServer()` throwing before any instalment ships means nothing reached a follower, so `retireAbandonedInstalments` is a no-op on `instalments == 0` and the build aborts and retries exactly as it did before this PR - and `RebuildIndexStatement.buildIndex` already retries `NeedRetryException`, which you traced yourself. Constructing it needs a transiently-null `RaftHAServer`, which has no seam today; adding one purely for a path whose behaviour is unchanged is not a trade I would make.

Final state: seven review rounds, three of which found real defects (thread-local RID resolution on multi-database servers, a condemned chain being re-walked and its edges deleted, and instalment files recorded after the send instead of before). All fixed and covered. Three residual items tracked as #6142, #6143, #6144.

PR #6140, opened 2026-08-13
fix(#6129): the slot merge now covers the records that outgrew their page

### LINKED ISSUE #6129: Slot merge does not cover records that outgrew their page: chunked and placeholder updates poison the page unconditionally
Spun out of the #6127 item 3 re-triage of `Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent`. That test was being written off as a chronic flake; it is not flaky, and it is not a hole in the slot-merge logic either. It is the engine hitting a regime the merge does not cover at all.

## What the measurement showed

The test runs 8 threads over 40 records of a **single-bucket** type, each round rewriting its own records with a longer payload (`400 * (round + 1)` chars). Instrumenting every poison site in `LocalBucket.updateRecordInternal` plus the commit-time outcomes in `TransactionContext.commit1stPhase`, one failing run gives:

| counter | value |
|---|---|
| slot rebases performed (`isRebasableSlotPage` taken) | 146 |
| records spilled out of their page (`writeMultiPageRecord` / placeholder branch) | 30 |
| `ConcurrentModificationException` reaching the retry loop | 16 |
| threads that ran out of the default 3 retries | 5 of 8 |

While the payloads still fit the shared 64 KB page, the disjoint-slot merge absorbs everything - 146 successful rebases, zero conflicts. From round ~4 on, `40 * 3200` bytes no longer fit, every record becomes a multi-page chunk record, and from then on **every** update takes the `FIRST_CHUNK` branch, which poisons the page unconditionally. Eight threads then contend on one page where nothing is mergeable: one commit wins per round and the other seven get a CME. With `TX_RETRIES` at 3 that is not enough budget, and threads give up.

With a larger budget every run converges, `CHECK DATABASE` reports zero errors and every record holds exactly its last written value - so nothing is corrupt. The retry budget is the symptom; the unmergeable shape is the cause.

## The gap

`TX_PAGE_SLOT_MERGE` covers, on a bucket page: an insert into a free slot, an in-place overwrite of the same size or smaller, an in-page growth (`growRecordInPage`, #5279), and the delete of a plain in-place record (#5569). It covers **none** of the shapes a record takes once it outgrows its page:

- a slot holding a `FIRST_CHUNK` header whose continuation chunks live on other pages;
- a slot holding a `RECORD_PLACEHOLDER_POINTER`, when the pointed-to content record has to be rebuilt rather than updated in place;
- a `RECORD_PLACEHOLDER_CONTENT` record growing on its own page (excluded by the `!isPlaceHolder` guard in both growth branches).

Every one of those poisons its page, so on a bucket where records have grown past the page size, concurrent updates of *unrelated* records conflict for good - which is precisely the false conflict #5279 set out to remove, reappearing one size class up.

## Worth investigating

The three shapes are not equally hard, and they should be judged separately rather than as one feature:

1. **Placeholder content growth** looks closest to already-solved. It is a single slot on its own page and `growRecordInPage` already knows the placeholder marker (`isPlaceHolder`); what is missing is carrying that flag through `SlotRebaseBuffer` so `rebaseRecordOnPage` accepts a `rs[0] < RECORD_PLACEHOLDER_CONTENT` pre-image instead of bailing on `rs[0] <= 0`. The pointer on the owner page is untouched by an in-place growth, so the RID is stable.
2. **Placeholder pointer rewrite** is a fixed-size 8-byte slot write, but it is paired with a delete + create on another page, so the merge would have to reason about more than one page atomically.
3. **Chunk chains** are the hardest: one logical write spans several pages and every one of them would have to rebase for the commit to survive, so the failure probability compounds rather than falls.

(1) alone may be enough to move the needle, and it is the one with a self-contained proof obligation. It is worth measuring what fraction of the conflicts in a realistic large-record workload each shape accounts for before committing to (2) or (3).

Whatever is decided, the payoff is concrete: `Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent` should pass at the `TX_RETRIES` default instead of needing an explicit budget, and its sibling `growingUpdatesThatStayInsideTheirPageNeverConflict` (added in the #6127 PR) should extend to payloads that outgrow the page with the conflict count still asserted at zero.

## Not a regression

Nothing here is new behaviour; the shapes have never been rebasable. What is new is knowing that this - and not a race in `growRecordInPage` - is what the test has been reporting.

lvca:
Fixes #6129.

## What the measurement said

The issue asks to judge the three unmergeable shapes separately and to measure which one costs what before committing to any of them. Instrumenting the failing workload (8 threads, 40 records of a single bucket, each thread rewriting only its own records, `attempts=1`) gives an unambiguous answer:

- **280 of 320 commits conflict** once the records have spilled - seven of eight writers lose every round;
- **every single one of those conflicts is on page 0**, the page holding the records' head chunks. The continuation-chunk pages never conflict at all.

The reason is worth writing down, because it also decides what is worth fixing: a record update is *deferred*. `save()` puts the record in `updatedRecords` and pins its own page there and then; the write itself runs in `commit1stPhase`, **after** the file's commit lock is taken. So the head chunk's page is the only page of a multi-page record a transaction holds at a stale version - every other page it needs is loaded fresh, at the newest committed version, under the lock.

With the merge extended, the same run reports **zero** conflicts.

## What is now mergeable

Three shapes, each tracked with its own slot *kind* so the replay checks the marker on the committed page and not only the bytes:

1. **The head chunk of a multi-page record.** Its footprint on the page is fixed when the record spills and is never grown again - `updateMultiPageRecord` rewrites the chunk size, the pointer to the next chunk and the chunk content, all inside the record's own bytes.
2. **The spill itself**, where a plain record that no longer fits becomes that head chunk. The one tracked write whose two images have different shapes, and the replay re-establishes the room the spill used (the slot's own footprint, plus the free tail of the page when it is the last record). Leaving it out was not a one-off cost: a spill that lost the race rolled back and had to spill again on the retry, so under contention the records that had to spill could starve rather than converge. In the #5279 workload this alone took the conflicts from 158 to 7 (at `attempts=1`).
3. **The content record behind a placeholder pointer** (issue item 1), in place or growing on its own page.

Still poisoning, deliberately: the continuation chunks (inline record-table writes no slot image accounts for - and, per the measurement above, pages that cannot conflict anyway), and a slot turning into or being rebuilt from a placeholder *pointer*, which changes two pages at once.

## The part the issue did not anticipate, and the reason this is not just "track three more shapes"

Every replay is gated on a byte-for-byte pre-image check - that is what keeps two transactions writing the **same** record conflicting, so the application reloads instead of losing a write. For a multi-page record that check can only see the head chunk, because that is the only part of the record on the page being replayed. A concurrent transaction that rewrote the record only *past* that chunk would slip through it and be silently overwritten - and, as established above, the page-version check cannot catch it either, since those pages are read fresh under the lock.

`LocalBucket.chunkChainTailFingerprint` closes it: a 64-bit FNV-1a over the chain past the head chunk (each chunk's RID, size and content), taken when the record is taken for update - the same moment its page is pinned - and again at commit before the chain is rewritten. The head chunk is replayable only while the two agree; when they do not, the page falls back to the plain retry it took before. The cost is proportional to work the update was going to do anyway (an update of a chunked record rewrites every chunk of its chain), and a record that is not a chunk chain pays nothing beyond reading its own marker.

`Issue6129ChunkedSlotMergeTest.aTailOnlyChangeToTheSameChunkedRecordStillConflicts` pins it with two 200 KB values differing only in their last byte, so they share their whole head chunk. Verified by mutation: forcing `isChunkChainTailUnchanged` to return `true` turns **exactly that one test** red across the whole suite, and it goes red by *committing* - i.e. without the fingerprint that commit silently drops the other transaction's write. (The continuation-chunk pages cannot catch it instead: they are loaded fresh under the commit lock, so their version check has nothing stale to refuse. That is the same property the measurement above rests on.)

Under `REPEATABLE_READ` the mechanism is a different one: `getPage` caches, so both fingerprint walks read the transaction's own copies and agree - and the same caching leaves the chain pages stale in the modified set, where the ordinary page-version check refuses them. `aTailOnlyChangeToTheSameChunkedRecordStillConflictsUnderRepeatableRead` pins that half, and passes with the fingerprint disabled.

## Tests

New `Issue6129ChunkedSlotMergeTest` (9 tests), each mutation-checked so none of them can pass for the wrong reason - the two that depend on record placement prove their premise by re-running the same interleaving with `TX_PAGE_SLOT_MERGE` off and requiring it to fail:

- `updatesOfChunkedRecordsSharingAPageNeverConflict` - the issue's shape, `attempts=1`, zero conflicts;
- `growingUpdatesOfChunkedRecordsSharingAPageNeverConflict` - same, with the chain being extended under the head chunk;
- `aRecordSpillingIntoChunksCommutesWithWritesToOtherSlots` - the transition itself;
- `aTailOnlyChangeToTheSameChunkedRecordStillConflicts` - the fingerprint, and the only test that fails when it is disabled;
- `aTailOnlyChangeToTheSameChunkedRecordStillConflictsUnderRepeatableRead` - the same guarantee where the fingerprint is not what provides it;
- `aConcurrentChangeToTheSameRecordsChainLengthStillConflicts` - a concurrent chain-length change, caught by the head chunk's pre-image check because a record's length prefix lives there;
- `concurrentUpdatesOfTheSameChunkedRecordStillConflict` - the guarantee that must not weaken;
- `updatingPlaceholderContentCommutesWithItsPageNeighbours` - item 1, in place and growing;
- `aSlotThatChangesShapeInOneTransactionIsExcludedFromTheMerge` - the poison-on-shape-change guards, driven through the tracking API because no write path reaches those states through the engine today.

`Issue5279ConcurrentUpdateTest.growingUpdatesUnderContentionKeepTheDatabaseConsistent` no longer needs its explicit budget of 50 attempts: it runs at the `TX_RETRIES` default, and the absence of an `attempts` argument is the assertion. Full `engine` unit-test lane green.

## Follow-up, filed

The measurement turned up a pre-existing hole this PR does not change: a concurrent update of the same **placeholder-backed** record is lost silently today, with no conflict raised at all. Both writers load the content page fresh under the lock, and neither writes the page holding the pointer, so nothing is version-checked. Filed as #6141, with the repro and the two candidate fixes; the fingerprint idea used here would close it.

## Review rounds

Five, all non-blocking after the first. What they changed, beyond wording: the bare `catch (Exception)` in the chain walks now asserts on the two exceptions page content cannot produce, so a defect there fails a test instead of degrading into a page that is never merged; the million-hop loop guard became exact revisited-pointer detection; the spill replay's marker budget is bound by an assertion to the marker it writes; the shape-change guards got the test above; and the probe added to the update path was measured (65 ns/call) and then made to reuse the page its caller had just pinned (4-5 ns/call, ~0.2% of a plain record update, one allocation fewer), with the slot-kind map allocated only for a page that has a non-plain shape to record.


lvca:
Both points from the last round, resolved:

**`@author` on the new private helpers.** Checked against the file as it was before this PR (`74059fe13`): it already carried method-level `@author Luca Garulli` tags on three members - `findBrokenChunkChain`, `isChunkChainBroken` and `loadMultiPageRecord` - besides the class-level one. All three are chunk-chain methods, which is the same neighbourhood the new helpers live in, so the tags follow the file's own convention rather than departing from it. Left as they are.

**The follow-up issue.** Filed as #6141 before that round, with the repro, the mechanism (neither writer touches the pointer page, and the content page is loaded fresh under the commit lock, so nothing is version-checked) and the two candidate fixes. The PR description links it now.

For the record on the earlier rounds' open questions: the local `engine` unit-test lane is green (11,758 tests), including the deliberately-corrupted-chain fixtures that the new assertion in `unwalkableChunkChain` has to keep tolerating - `BucketIteratorBrokenChunkChainTest`, `BucketIteratorCorruptedSlotTest`, `BucketIteratorCorruptedPlaceholderPointerTest` and `BrokenMultiPageRecordDeleteTest` all pass.

lvca:
Re-ran the mutation you asked for, and the answer is measurable both ways.

**The fingerprint is load-bearing.** Forcing `isChunkChainTailUnchanged` to return `true` turns **exactly one** test red across the whole suite - `aTailOnlyChangeToTheSameChunkedRecordStillConflicts` - and it goes red by *committing*:

```
[ERROR] Tests run: 8, Failures: 1
java.lang.AssertionError: Expected a conflict: the other transaction rewrote the tail of the very same record
```

So without that check the losing transaction commits and the other write is silently gone. The hypothesis that the continuation-chunk poisoning already catches this interleaving does not hold, for the same reason the whole change exists: a record update is **deferred**, so those pages are not held at a stale version at all. `addUpdatedRecord` pins the record's own page at `save()` time, and every other page the write needs - including every continuation chunk - is loaded fresh inside `commit1stPhase`, **after** the file's commit lock is taken and therefore after the winner committed. Poisoning excludes a page from the merge; it does not manufacture a version conflict on a page whose copy is already current. That is also what the original measurement showed: all 280 conflicts were on page 0, and the continuation-chunk pages never conflicted once.

**Your REPEATABLE_READ observation was right to flag, and had no test.** Under that level `getPage` caches, so both fingerprint walks read the transaction's own copies and agree however the record changed meanwhile - the fingerprint really is inert there. What makes it safe is the same caching, by a mechanism worth naming: the chain pages are then held at the version the transaction read them at, so rewriting them at commit takes **stale** pages into the modified set and the ordinary page-version check refuses them. Added `aTailOnlyChangeToTheSameChunkedRecordStillConflictsUnderRepeatableRead`, which pins that half and passes with the fingerprint disabled, and wrote the interaction into `engine/CLAUDE.md`: two isolation levels, two mechanisms, one guarantee.

PR #6145, opened 2026-08-13
fix(#4896): anti-drift between HTTP routes and the OpenAPI spec

### LINKED ISSUE #4896: Clients: anti-drift contract test for the OpenAPI spec

robfrank:
## What does this PR do?

Closes the gap where a real HTTP route (core or plugin-contributed) could drift from the server's OpenAPI spec without CI ever noticing, in either direction:

1. **Core routes** (~48 routes wired through `HttpServer.setupRoutes()`'s Undertow `RoutingHandler`s): a new `RouteRecordingRoutingHandler` records every actually-registered `(method, path)`. `OpenApiSpecGenerationIT` gained `coreOperationsMatchTheServersActualRegisteredRoutes`, which compares those *real* registrations against the live spec directly - replacing what used to be two independently hand-maintained lists that could stay in sync with each other while both silently omitted a real route.
2. **Plugin-contributed routes** (ha-raft's 11 cluster/snapshot paths, metrics' 1 Prometheus path, mcp's 2 paths - registered via `PathHandler`, not `RoutingHandler`, and undocumentable by the plugins themselves since those 3 modules hold `arcadedb-server` at `provided` scope): a `RecordingPathHandler` + `RoutePathNormalizer` test-utility pair supports one new anti-drift test per plugin module, each comparing the plugin's actual registered paths against a `Set<String>` constant now exposed from `PluginApiSpec`/`McpApiSpec`, guarded against drifting from the real spec generator by `ApiSpecPathConstantsTest`.

Two anti-drift self-tests (one core, one plugin-side) prove the comparison mechanism itself genuinely fails when routes and spec disagree, per the issue's explicit acceptance criterion - not just that it passes when they happen to agree.

## Motivation

#4895 (merged) closed the OpenAPI spec's coverage gap once. Without a structural or CI-enforced guarantee, the spec drifts again the moment someone adds a handler without touching the OpenAPI side - #4895's own postmortem flagged the residual plugin-route gap explicitly, and building this surfaced a second, more fundamental gap in the core-route check that had never been caught before.

## Related issues

Closes #4896. Child of epic #4894, sibling to the merged #4895.

## Additional Notes

**One unrelated bugfix is bundled in this branch** (first commit, `57771da7c`): `main` currently does not compile a plain `engine` build - commit `#6107` reintroduced calls to `BaseExpression`/`PInteger` `int` constructors that commit `#6117` had already removed. Found while implementing the first task here (nothing in this branch would otherwise compile), fixed in its own standalone commit with no other changes. Recommend merging this fix on its own timeline if you'd rather not gate it behind this feature's review - happy to split it into a separate PR if preferred.

Full design doc and implementation plan were written but are **not included in this PR** (left as local, untracked files) per the author's preference at merge time.

## Checklist

- [x] I have run the build using `mvn clean package` command (targeted per-module builds; full reactor `engine` suite has pre-existing unrelated flakiness independent of this branch - see PR discussion if asked)
- [x] My unit tests cover both failure and success scenarios (each anti-drift check has a corresponding negative self-test proving it can fail)

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLASTBuilder.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLASTBuilder.java
@@ -2750,8 +2750,8 @@ public class SQLASTBuilder extends SQLParserBaseVisitor<Object> {
       // only Long.MIN_VALUE's magnitude can still be represented once the sign is folded in;
       // anything larger is a genuine overflow and keeps raising the original error
       try {
-        final BaseExpression baseExpr = new BaseExpression(-1);
-        baseExpr.number = new PInteger(-1).setValue(Long.parseLong("-" + text));
+        final BaseExpression baseExpr = new BaseExpression();
+        baseExpr.number = new PInteger().setValue(Long.parseLong("-" + text));
         return baseExpr;
       } catch (final NumberFormatException stillInvalid) {
         throw new CommandSQLParsingException("Invalid integer: " + originalText);
--- a/ha-raft/src/test/java/com/arcadedb/server/ha/raft/RaftHAPluginRegisteredRoutesMatchApiSpecTest.java
+++ b/ha-raft/src/test/java/com/arcadedb/server/ha/raft/RaftHAPluginRegisteredRoutesMatchApiSpecTest.java
@@ -21,9 +21,11 @@ package com.arcadedb.server.ha.raft;
 import com.arcadedb.server.http.RecordingPathHandler;
 import com.arcadedb.server.http.RoutePathNormalizer;
 import com.arcadedb.server.http.handler.openapi.PluginApiSpec;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Issue #4896: RaftHAPlugin holds arcadedb-server at provided scope, so it cannot declare its own
@@ -33,9 +35,17 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 class RaftHAPluginRegisteredRoutesMatchApiSpecTest {
 
+  private RaftHAPlugin plugin;
+
+  @AfterEach
+  void tearDown() {
+    if (plugin != null)
+      plugin.stopService();
+  }
+
   @Test
   void registeredRoutesMatchTheDeclaredApiSpecPaths() {
-    final RaftHAPlugin plugin = new RaftHAPlugin();
+    plugin = new RaftHAPlugin();
     final RecordingPathHandler routes = new RecordingPathHandler();
 
     plugin.registerAPI(null, routes);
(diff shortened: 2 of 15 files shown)


robfrank:
Addressed the four nits:

1. **Method-blind filter** — fixed in `1851dafb6`. `coreOperationsMatchTheServersActualRegisteredRoutes` now excludes plugin/MCP declarations by exact `"METHOD path"` match against `EXPECTED_MCP_AND_PLUGIN_OPERATIONS` rather than by path alone, which also drops the substring parsing on both sides. `OpenApiSpecGenerationIT` still 16/16.

2. **`mergeWithPrefix` transposition risk** — leaving as-is. This is the same finding an earlier internal review round already raised and parked for the same reason you gave: only one call site today, the anti-drift test itself catches a transposition immediately, and a `RoutingGroup` record only earns its keep once a third prefix/handler pair shows up. Will revisit if that happens.

3. **`addAll()` bypass note** — no action needed, it's already the documented, deliberate scope boundary (`RouteRecordingRoutingHandler`'s class Javadoc), not a gap to close now.

4. **Splitting the `SQLASTBuilder` fix into its own PR** — agreed it's the right move, but opening a second PR is outside what I'll do unprompted; that's the repo owner's call.

robfrank:
On the three nits:

1. **Javadoc's broader-coverage claim was untested for verbs without a convenience method** — fixed in \`8e34ba607\`: added \`recordsAVerbRegisteredThroughTheGenericAddOverloadWithNoConvenienceMethod\`, registering through \`add(\"PATCH\", ...)\` directly and asserting it's captured. That backs the Javadoc's "every verb funnels through the two terminal methods" claim with an actual assertion instead of leaving PATCH/HEAD/OPTIONS/TRACE coverage implied. \`RouteRecordingRoutingHandlerTest\` now 4/4.

2. **\`synchronized\` on the two terminal overrides being unnecessary overhead** — no action, you already called this not worth changing given \`add()\`'s cost and the single-threaded startup-time call pattern; agreed.

3. **\`addAll()\` blind spot** — no action, already the documented, deliberate scope boundary.

On verification: this was run locally via Maven repeatedly through the review/fix cycle, not just claimed — most recently \`OpenApiSpecGenerationIT\` 16/16 and the three per-plugin tests (ha-raft/metrics/mcp) all green after each commit in this thread, in addition to full-module suites (server 723+16, ha-raft 811, metrics 10, mcp 295) run earlier in the branch. CI will confirm on this push too.

PR #6147, opened 2026-08-13
chore: keep Meterian's stability findings out of code scanning

robfrank:
## What

Meterian reports security, stability and licensing findings in one SARIF, and all three land in the GitHub code scanning tab. The stability ones (`[stability] <pkg> is outdated`) are not vulnerabilities - they say only that a newer release exists, which is what Dependabot already opens PRs for, and it is configured for every ecosystem in this repo.

On the 2026-08-13 scan of `main` they were **46 of the 65 alerts**. 71% of a tab whose whole purpose is triage was version drift that the actual security findings had to be read around.

This filters them out of the report between the scan and the upload.

## Why not configure the scanner instead

That was the first thing I tried. Meterian has no flag for it:

- `--report-console` accepts a `security|stability|licensing` filter, but **`--report-sarif` takes a filename only**.
- The `.meterian` exclusions file does support a `stability` section with globs, but it is documented as acting on the **scores**. Scores gate nothing here - the scan already runs `continue-on-error` precisely because the composite score is a property of the whole transitive graph that no single PR can move. So that route would have been a no-op, and confirming otherwise would have needed either a Meterian account or a bogus build pushed under the real project.

The report itself, between the scan and the upload, is the part that is ours to control.

## How

`.github/scripts/filter-meterian-sarif.py` drops every rule tagged `stability` **and not also tagged `security`**, along with the results referencing it.

The trap it exists to handle: results reference their rule by **array position as well as by id**, so rebuilding the rules array without remapping `result.rule.index` silently reattributes findings to whatever rule slid into that slot. The script remaps from the id.

It is deliberately conservative - anything whose tags cannot be read, or whose rule is missing from the report, is kept. This script must never be the reason a vulnerability stops being reported. The step is also `continue-on-error`, so a failure falls open to uploading the unfiltered report, i.e. today's behaviour.

## Verification

Rather than guess at the format, I pulled the **real uploaded SARIF** from the code-scanning API and committed it as a fixture:

```bash
AID=$(gh api "repos/ArcadeData/arcadedb/code-scanning/analyses?per_page=50" \
       --jq '[.[] | select(.tool.name=="Meterian" and .ref=="refs/heads/main")][0].id')
gh api "repos/ArcadeData/arcadedb/code-scanning/analyses/$AID" \
   -H "Accept: application/sarif+json" > meterian.sarif
```

Against it: **65 results / 64 rules becomes 19 / 18**, every survivor tagged `security`, with `automationDetails` and the `artifacts` indices intact. `automationDetails` matters - it is the analysis category that lets GitHub match the new analysis to the existing one and auto-close the 46.

11 cases added to `.github/scripts/tests/test-ci-scripts.sh`, which `mvn-test.yml` already runs:

```
filter-meterian-sarif.py
  ok   - filters the real 2026-08-13 scan of main
  ok   - leaves 18 rules and 19 results
  ok   - every surviving rule is tagged security
  ok   - remaps result.rule.index to the surviving rule array
  ok   - is idempotent
  ok   - handles a report with no result.rule.index
  ok   - keeps the security finding when there is no rule index
  ok   - keeps a rule tagged both stability and security
  ok   - filters every run, not just the first
  ok   - treats a missing report as nothing to do
  ok   - rejects a malformed report

56 checks passed
```

The index-remap case was checked against a deliberately sabotaged remap: it flagged all 19 survivors as misattributed, so it fails when it should.

`no result.rule.index` is not a hypothetical - GitHub adds that field when it normalises the report, so Meterian's own upload has the bare shape and the fixture has the normalised one. Both are covered.

## What this does not change

The 46 stability findings are not lost. They stay in the Meterian report linked from every alert, the scan still runs on PRs, `main` and nightly, and the dependencies stay under Dependabot. What changes is that the security tab holds security findings only.

## Context

While reading the SARIF for this, the tags gave an authoritative breakdown of the 65 open alerts that the alert list alone does not:

```
46  stability
18  CWE-1104 + security + vulnerability     <- "library unmaintained", fixedVersions: []
 1  security + vulnerability                <- golang.org/x/crypto, the only real advisory
```

Two things worth recording separately from this PR:

- The one genuine advisory is **`golang.org/x/crypto@v0.54.0`** (alert 2219) - the Go advisory that `x/crypto/openpgp` is unsafe by design. Unlike everything else on the list it has a fix, **v0.55.0**. It is an `// indirect` dependency of `e2e-go/go.mod`, so test-harness-only and never shipped, and Dependabot covers `gomod` at `/e2e-go` - worth confirming it arrives.
- **`bolt/conformance/requirements.txt` has no `.github/dependabot.yml` entry**, so its PyYAML alert has no automatic update path. Separate one-line fix.

CodeQL, Dependabot alerts and secret scanning are all at 0 open.

### DIFF
--- a/.github/scripts/filter-meterian-sarif.py
+++ b/.github/scripts/filter-meterian-sarif.py
@@ -115,6 +115,13 @@ def main() -> int:
         print(f"filter-meterian-sarif: {report} is not valid JSON: {error}", file=sys.stderr)
         return 2
 
+    # Well-formed JSON that is not an object at all - a bare array or string - would otherwise
+    # reach doc.get() and raise, which is safe but reports nothing useful. Say the same kind of
+    # thing the decode error above says.
+    if not isinstance(doc, dict):
+        print(f"filter-meterian-sarif: {report} is not valid JSON: expected an object at the top level", file=sys.stderr)
+        return 2
+
     runs = doc.get("runs")
     if not isinstance(runs, list):
         print(f"filter-meterian-sarif: {report} has no runs, nothing to filter")
--- a/.github/scripts/filter-meterian-sarif.py
+++ b/.github/scripts/filter-meterian-sarif.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+#
+# Drops Meterian's "[stability] <pkg> is outdated" findings from a SARIF report before it is
+# uploaded to GitHub code scanning.
+#
+# Meterian reports three dimensions - security, stability and licensing - and puts all of them in
+# one SARIF. Stability findings are not vulnerabilities: they say only that a newer release of a
+# dependency exists, which is what Dependabot already opens pull requests for. On the 2026-08-13
+# scan of main they were 46 of the 65 alerts, so 71% of a tab whose whole purpose is triage was
+# taken up by "this dependency is a version behind", and the security findings had to be picked
+# out from among them.
+#
+# Meterian has no flag for this. --report-console takes a security|stability|licensing filter but
+# --report-sarif takes a filename only, and the .meterian exclusions file is documented as acting
+# on the scores rather than on report contents - which would be a no-op here, since the workflow
+# already runs the scan with continue-on-error and gates nothing on the score. Filtering the
+# report between the scan and the upload is the part that is ours to control.
+#
+# The findings are not lost: they stay in the Meterian report linked from every alert, and the
+# dependencies themselves stay under Dependabot. What changes is that the GitHub security tab
+# holds security findings only.
+#
+# A rule is dropped when it is tagged "stability" and not also tagged "security". Anything whose
+# tags cannot be read, or whose rule is missing from the report altogether, is kept: this script
+# must never be the reason a vulnerability stops being reported.
+#
+# Usage:
+#   filter-meterian-sarif.py <report.sarif>     (rewritten in place)
+import json
+import sys
+from pathlib import Path
+
+
+def is_stability_only(rule: dict) -> bool:
+    """True when a rule reports outdatedness and nothing else."""
+    tags = rule.get("properties", {}).get("tags", [])
+    if not isinstance(tags, list):
+        return False
+    return "stability" in tags and "security" not in tags
+
+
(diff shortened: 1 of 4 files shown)


robfrank:
Verified all three against the code. Two were real and are fixed in 2fc9f23b8; the third is stated rather than changed.

**1. `properties: null` crashes - correct, fixed.** Reproduced directly:

```
>>> {'properties': None}.get('properties', {}).get('tags', [])
AttributeError: 'NoneType' object has no attribute 'get'
```

`get(key, default)` returns the default only when the key is *absent*, so an explicit null falls straight through. You identified the consequence exactly: with the step on `continue-on-error` this would not have broken CI, it would have shipped the whole unfiltered report while appearing to work, which is precisely what the file's "keep anything whose tags cannot be read" comment claims it does not do. Now `(rule.get("properties") or {})`, with the reasoning recorded at the call site.

**2. Top-level `result.ruleIndex` - correct, fixed.** Only the nested `result.rule.index` form was remapped. Both are handled now. Confirmed the fixture genuinely does not exercise it (`has("ruleIndex")` is 0 across all 65 results) - GitHub's normalisation emits the nested form, which is why committing a normalised report as the fixture hid it. That is a good catch about the fixture's blind spot, not just the code's.

**3. Rule id collisions - real but not worth code.** Checked the actual report: 64 rules, 64 distinct ids, and SARIF requires uniqueness. If two kept rules shared an id then `ruleId` no longer identifies a rule and the report is ambiguous before this script sees it. Taking your suggestion literally: it is now an accepted risk written down where it is relied on, rather than an unstated assumption.

Both new checks were confirmed to fail against the previous script before the fix landed. The top-level index check bounds-checks before dereferencing, so a stale index fails that one check instead of throwing an `IndexError` that aborts the harness - the first version of the test did abort it. 59 checks pass.

Unrelated to this PR: `build-and-package` is red here because **main does not compile**. #6117 removed `SimpleNode`'s int constructor parameter and #6107 added `new BaseExpression(-1)` / `new PInteger(-1)` call sites in `SQLASTBuilder.java`; each was green on its own branch and they collided on merge. `main` has been failing since 07:41Z today, before this branch existed.

Issue #6135, opened 2026-08-13
`RETURN *` not recognizing variables in scope from subquery


dmoree:
### Issue
`RETURN *` does not account for variables put into scope from a subquery

### Reproduction
```cypher
CALL {
  MATCH (n)
  RETURN n
}
RETURN *
```

### Current
Results in an error:
`NoVariablesInScope: RETURN * is not allowed when there are no variables in scope`

### Expected
When validating `RETURN *` it should account for all variables in scope

#### Versions
ArcadeDB: 26.8.1

PR #6150, opened 2026-08-13
fix(#6148): CALL proc() YIELD * now binds the procedure's output fields

### LINKED ISSUE #6148: CALL proc() YIELD * does not bind any procedure output variable
### Issue
`CALL proc() YIELD *` never puts any of the procedure's yielded output names into scope, so referencing them afterwards - by name or via `RETURN *` - fails.

### Reproduction
```cypher
CALL db.labels() YIELD * RETURN label
```
```
CommandSemanticException: UndefinedVariable: Variable 'label' not defined
```
```cypher
CALL db.labels() YIELD * RETURN *
```
```
CommandParsingException: NoVariablesInScope: RETURN * is not allowed when there are no variables in scope
```

### Expected
`YIELD *` should behave like an explicit `YIELD` of every output the procedure produces (Neo4j: the wildcard form yields all available result fields under their default names), making them referenceable afterwards exactly as `CALL db.labels() YIELD label RETURN label` already is.

### Root cause (found while working on #6135)
`CypherSemanticValidator.validateVariableScope`'s `CALL` case only adds names from `callClause.getYieldItems()` to scope:

```java
case CALL:
  final CallClause callClause = entry.getTypedClause();
  if (callClause != null && callClause.hasYield())
    for (final CallClause.YieldItem item : callClause.getYieldItems())
      scope.add(item.getOutputName());
  break;
```

For `YIELD *`, `hasYield()` is `true` and `isYieldAll()` is `true`, but `getYieldItems()` is empty, so the loop adds nothing. The same shape exists in `buildVarTypes`'s `CALL` case (kind tracking) and in the new `statementDeclaresAnyVariable` (#6138) - all three would need the procedure's declared output signature to expand `YIELD *` into concrete names, which isn't available at semantic-validation time without a procedure registry lookup.

#### Versions
ArcadeDB: main (26.9.1-SNAPSHOT)

lvca:
## Summary

`CALL proc() YIELD *` never put any of the procedure's yielded output names into scope, so a named reference to one of them - or a following `RETURN *` - failed with `UndefinedVariable`/`NoVariablesInScope`, even though the executor already projects the fields correctly once the query is allowed to run at all.

```cypher
CALL db.labels() YIELD * RETURN label
-- CommandSemanticException: UndefinedVariable: Variable 'label' not defined

CALL db.labels() YIELD * RETURN *
-- CommandParsingException: NoVariablesInScope: RETURN * is not allowed when there are no variables in scope
```

Neo4j (the openCypher reference implementation) treats `YIELD *` as an ordinary `YIELD` of every output field the procedure declares, making them referenceable afterwards exactly as `CALL db.labels() YIELD label RETURN label` already is.

## Root cause

`CypherSemanticValidator`'s three CALL-handling switches (`buildVarTypes`, `validateVariableScope`, and `statementDeclaresAnyVariable`, the last added by #6138 for #6135) all learn what a `CALL ... YIELD` exports by iterating `callClause.getYieldItems()`. That list is deliberately empty for `YIELD *` (`hasYield()` is true, `isYieldAll()` is true, but there are no items), so all three silently added nothing.

Separately, five Neo4j-compatible built-ins (`db.labels`, `db.relationshipTypes`, `db.propertyKeys`, `db.schema`, `db.schema.visualization`) bypassed `CypherProcedureRegistry` entirely via a hardcoded `switch` in `CallStep`, so even a registry-based fix wouldn't have covered the issue's own reproduction case.

## Fix

- Added `CypherSemanticValidator.resolveYieldAllFieldNames`, which resolves a `YIELD *` against the calling procedure's declared `CypherProcedure#getYieldFields()` via `CypherProcedureRegistry` - the same registry `CallStep` already consults to run the call. Wired into all three CALL-handling switches. When the name doesn't resolve to a registered procedure (a custom `DEFINE FUNCTION`, an arbitrary SQL function, or an unknown name), behavior is unchanged from today rather than guessing at an output shape that isn't statically knowable.
- Converted the five hardcoded `db.*` built-ins into real `CypherProcedure` implementations (`DbLabels`, `DbRelationshipTypes`, `DbPropertyKeys`, `DbSchemaVisualization` registered under both `db.schema` and `db.schema.visualization`), registered in `CypherProcedureRegistry` like any other procedure. This removes the special-case switch in `CallStep` and makes `getYieldFields()` - previously declared on every procedure but never actually called anywhere - finally used, closing the exact registry gap the issue's own root-cause analysis flagged.

## Behavior change worth calling out

Routing the four hardcoded `db.*` built-ins through the normal `CypherProcedure` dispatch means they now go
through the same `validateArgs`/`checkArity` check as every other procedure (`getMinArgs()`/`getMaxArgs()` are
both `0`). Previously the hardcoded `switch` in `CallStep` ignored `args` entirely, so e.g. `CALL db.labels(1)`
silently succeeded and ignored the argument. Now it throws `CommandSemanticException` for the unexpected
argument, consistent with every other procedure and with Neo4j. This is more correct, but is a side effect of
the refactor rather than the headline fix, so calling it out explicitly here.

## Test plan

- [x] New regression test `Issue6148YieldStarProcedureScopeTest` reproduces the issue (fails before the fix, passes after), covering: a hardcoded built-in (`db.labels`), the other three renamed built-ins, a normally-registered procedure (`meta.stats`), `YIELD *` inside a `CALL {}` subquery, both registered names of the schema-visualization alias, and that an unresolvable procedure name (a custom `DEFINE FUNCTION`) keeps pre-fix behavior instead of the fix guessing at an unknowable output shape.
- [x] `mvn -pl engine test` for the full `com.arcadedb.query.opencypher.**` package (existing + new tests) - all green, including `Issue6135ReturnStarSubqueryScopeTest`, `CypherCallYieldWithVariablesTest`, `CypherProcedureRegistryTest`, `CypherSubqueryBodyIssue5656Test`, `CypherSemanticValidatorExceptionTest`.

Closes #6148.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/CallStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/CallStep.java
@@ -33,22 +33,15 @@ import com.arcadedb.function.CypherFunctionRegistry;
 import com.arcadedb.query.opencypher.procedures.CypherProcedure;
 import com.arcadedb.query.opencypher.procedures.CypherProcedureRegistry;
 import com.arcadedb.query.sql.executor.*;
-import com.arcadedb.schema.DocumentType;
-import com.arcadedb.schema.EdgeType;
-import com.arcadedb.schema.VertexType;
 import com.arcadedb.utility.CollectionUtils;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.Set;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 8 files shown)


lvca:
## CI status note: `ha-integration-tests`

This lane failed on 3 consecutive runs, each time on a **different** test in the HA/Raft integration suite, unrelated to this PR's changes:

1. `Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied`
2. `Issue5569SlotMergeDeleteRaftIT` + `Issue5410AbandonedTicketReleaseIT` (again)
3. `RemoteStickyConnectionStrategyIT.pinIsReleasedAfterRollbackAndAllowsSubsequentStickyTransaction` + `RaftCommandReadConsistencyIT`

This PR's diff touches only `engine/.../query/opencypher/**` (Cypher CALL/YIELD scope validation and procedure registry) - zero overlap with HA/Raft/clustering code. Four different HA integration tests failing across three runs, with no pattern tying any of them to this change, points to this being the lane's known pre-existing chronic flakiness rather than a regression.

Every other check (build, all unit/vector/opencypher-TCK/integration test lanes, all e2e lanes across every client, CodeQL, Codacy, Meterian, and three separate automated code-review passes explicitly concluding "nothing blocking merge") is green. Leaving this PR as-is rather than continuing to burn CI time re-running a lane with this failure rate - happy to re-trigger it again on request, or this can be merged with the understanding that `ha-integration-tests` red here is not attributable to this change.

Issue #5850, opened 2026-08-06
HAReplicationMetrics leaks a ScheduledExecutorService per bindTo(): the scheduler is held only in a local variable and never shut down


ruispereira:
## Summary

`HAReplicationMetrics.bindPerFollowerGauges()` starts a `ScheduledExecutorService` held only in a local variable, schedules a task on it every 5 seconds, and returns. The reference is unreachable from that point on, so the scheduler can never be shut down. Every call to `bindTo()` leaks one thread for the lifetime of the JVM.

Version: 26.8.1.

## Detail

```java
// server/src/main/java/com/arcadedb/server/monitor/HAReplicationMetrics.java:150
refresh.run(); // publish an initial (possibly empty) set immediately

final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(r -> {
  final Thread t = new Thread(r, "arcadedb-ha-follower-metrics");
  t.setDaemon(true);
  return t;
});
scheduler.scheduleAtFixedRate(refresh, 5, 5, TimeUnit.SECONDS);
```

`scheduler` goes out of scope immediately. There is no `shutdown()` call anywhere in the file, and no field or registry holds the instance.

The class implements Micrometer's `MeterBinder`:

```java
// HAReplicationMetrics.java:54
public final class HAReplicationMetrics implements MeterBinder {
```

`bindTo(MeterRegistry)` is a public contract that may legitimately be invoked more than once — once per registry, which is normal when a composite registry plus a Prometheus or JMX registry are in play — and each invocation starts another orphan scheduler. The bundled tests already call it four times in a single class, each leaking a thread that keeps publishing into a registry the test has finished with.

The threads are daemons, so they do not hold up JVM exit. What they do is accumulate, keep sampling the follower list every 5 seconds, and keep writing into registries that may be long discarded.

## Why this looks unintended

Every other gauge in the same class is computed lazily, on scrape, and the class says so:

```java
/**
 * Reads a live replication-health snapshot from the started HA plugin, or a not-leader placeholder
 * when no HA plugin is present (HA disabled). Cheap enough to call once per gauge per scrape.
 */
private HAReplicationStats stats() {
```

Only the per-follower `MultiGauge` block uses a background scheduler, and `MultiGauge` needs a push because its row set is dynamic — but that push does not need an unmanaged executor.

## Suggested fix

Either keep the scheduler and close it with the binder:

```java
private ScheduledExecutorService scheduler;   // field

[...]

PR #6152, opened 2026-08-13
fix(#6142, #6143, #6144): incremental instalment file split, unreferenced-file diagnostic, per-database instalment metrics

### LINKED ISSUE #6142: HA: the schema-WAL instalment flush rescans the whole recorded-changes list per instalment
Follow-up from #6136 (PR #6139), raised in review and deliberately not fixed there.

`RaftReplicatedDatabase.flushSchemaWalBufferIfFull()` recomputes which files still need announcing by walking the *whole* of `FileManager.getRecordedChanges()` on every instalment, filtering out the ones already in `shippedFiles`:

```java
final List<FileManager.FileChange> changes = proxied.getFileManager().getRecordedChanges();
if (changes != null)
  for (final FileManager.FileChange c : changes)
    if (c.create && !state.shippedFiles.containsKey(c.fileId))
      newFiles.put(c.fileId, c.fileName);
```

That is O(instalments x file changes).

**Why it is fine today, and therefore not fixed in #6139.** The caller instalments exist for is an index rebuild, which records one or two file changes however many instalments its WAL volume produces - so the second factor does not grow with the first and the quadratic term never materialises. A comment at the call site says so.

**When it would stop being fine.** A DDL that creates MANY files through the same buffered path - a multi-bucket type creation, a bulk index build over many buckets, anything that registers files while committing large amounts of WAL inside one `recordFileChanges` session - makes both factors grow together, on the leader, while it holds the database write lock.

**What the fix is not.** An index cursor into `recordedChanges` does not work: `FileManager.dropFile` REMOVES entries from the middle of that list when a session creates and then drops the same file (the create-cancellation this same code depends on elsewhere), so a saved position is not stable. The fix is to carry the shipped/unshipped split forward incrementally - e.g. have the flush consume from a per-session queue that `registerFile` appends to - rather than re-deriving it from the cumulative list each time.

Not urgent: no current caller reaches the shape that makes it matter. Filed so the self-imposed follow-up in the comment does not get lost.

### LINKED ISSUE #6143: HA: a leadership change mid-schema-session leaves unreferenced files on the other nodes with no automatic cleanup
Follow-up from #6136 (PR #6139), named in review and deliberately documented rather than fixed there.

## What happens

Since #6136 a schema session ships its buffered WAL in ordered **instalments** as it fills, instead of holding the whole payload in leader heap until the `recordFileChanges` callback returns. An instalment announces the files it created, and the session's final entry is what publishes the change.

If the session never reaches that final entry, `RaftReplicatedDatabase.retireAbandonedInstalments` sends a compensating removal for the files the instalments announced that this node no longer holds. That closes the ordinary failure - a build that throws - and is covered by `RaftSchemaWalInstalment3NodesIT.aSessionThatFailsAfterAnInstalmentRetiresWhatItAnnounced`.

**It does not close the leadership case.** `recordFileChanges` checks `isLeader()` once, at the top. Nothing re-checks it per instalment, and `requireRaftServer()` only asserts that a server exists, not that this node still leads. If leadership moves away after an instalment has shipped - failover, a brief partition, a manual step-down, all normal Raft events rather than application bugs - this node can no longer submit entries, so the compensating removal fails too. It is caught and logged at SEVERE naming every file it could not retire, and the message says explicitly when the cause was lost leadership, but the files stay on the other nodes until an operator removes them.

## Why it was not fixed in #6139

Re-checking leadership per instalment narrows the window without closing it: leadership can move between the check and the submit. And a node that is no longer leader has no way to make the cluster do anything, so the compensation fundamentally cannot be its job once it has lost the term.

Worth stating plainly what changed and what did not. Before instalments, a leadership change mid-build cost nothing - nothing had reached a follower, the final `replicateSchema` simply failed, and the caller saw a retryable error. The trade is bounded leader heap (the failure #6136 exists to prevent, where the leader held an entire rebuilt index) against this new cleanup case. It is the right trade, but it is a real new operational case.

## What the actual fix looks like

The new leader reclaiming files that no schema references - i.e. unreferenced-file garbage collection at the storage layer, which this database does not currently have. `GraphDatabaseChecker.reclaimOrphanedEdgeSegments` does something analogous for edge-list *segments* but not for paginated component files, and `CHECK DATABASE` has no arm for this at all.

Two smaller steps that would help before that exists:

1. a diagnostic that lists paginated files present on a node and referenced by no schema component, so an operator can act on the SEVERE log line without reading the data directory by hand;
2. test coverage for the SEVERE path itself - `RaftSchemaWalInstalment3NodesIT` covers the compensation succeeding, and the leadership-loss branch is currently only reasoned about. It is hard to construct deterministically (the step-down has to land between an instalment and the unwind), which is why it is filed rather than written.

Impact is bounded: wasted disk on the non-leader nodes, no correctness consequence. Nothing references the files, so no query, index or replication path reads them.

### LINKED ISSUE #6144: HA: expose schema-WAL instalment count and duration as operator-facing metrics
Follow-up from #6136 (PR #6139), suggested in three consecutive review rounds and deliberately not done there.

Since #6136 a schema session (principally an index rebuild, including the one `CHECK DATABASE FIX` performs) ships its buffered WAL in **instalments**, and each instalment is a quorum round trip taken **while the database write lock is held**. That is an accepted trade - the alternative is the leader holding an entire rebuilt index in heap - and it is documented on `flushSchemaWalBufferIfFull`, in `arcadedb.ha.appendBufferSize`'s own description, and in the release notes.

What is missing is a way to *see* it. Today:

- `RaftReplicatedDatabase.schemaInstalmentsShipped` is a JVM-wide `AtomicLong` whose only consumer is `RaftSchemaWalInstalment3NodesIT`'s before/after delta assertion;
- the per-instalment detail is a `HALog.DETAILED` line, so an operator debugging "every write on this database stalled for a while" has to turn on detailed HA logging *first*, which means reproducing the stall.

## What would help

1. **Instalment count**, per database. The existing counter cannot simply be wired to Micrometer: it is process-wide across every database in the JVM, so it needs a per-database dimension before it means anything on a multi-database server. That is a change to the counter, not just to its export.
2. **Instalment duration**, which is the number that actually matters here: it is what eats into the `arcadedb.ha.quorumTimeout` budget, and it is what a stalled writer is waiting on. Count alone does not distinguish "200 fast instalments" from "3 that each waited on a slow quorum member".
3. Possibly a single INFO line at the end of a session that shipped instalments, summarising count and total time, so the common case is visible without any metrics backend at all.

## Why not in #6139

The change there is about bounding leader heap; adding a metric surface is separable, and the counter needs reshaping (point 1) rather than just exporting. Filed so the suggestion is tracked rather than declined three more times.

Related: #6142, #6143.

lvca:
Closes #6142, closes #6143, closes #6144.

Three follow-ups filed against #6136 (PR #6139), which made a schema session ship its buffered WAL in ordered
**instalments** instead of holding the whole payload - for an index rebuild, roughly the whole rebuilt index - in
leader heap.

## #6142: the per-instalment rescan is gone

Every instalment has to answer one question: which files has this session created that the followers have not been
told about? It answered it by walking the whole of `FileManager.getRecordedChanges()` and filtering out what it had
already shipped, which is O(instalments x file changes). That never mattered for the caller instalments exist for -
an index rebuild records one or two file changes however many instalments its WAL volume produces - but a DDL that
creates many files through the same buffered path would make both factors grow together, on the leader, while it
holds the database write lock.

`FileManager` now carries the session's creates as a **consumable queue** that `drainRecordedCreates()` hands over,
so a whole session costs one pass over its file creations. As the issue notes, an index into the cumulative list
would not have worked - `dropFile` removes the cancelled create from the middle of it. The queue mirrors the
cumulative log at every site that touches a create: registration, the drop-cancels-create rule, and the post-rename
name refresh of #4083. `Issue6142RecordedCreatesDrainTest` pins each of those, because a create the queue forgets is
a file the followers are told to create and nobody ever retires.

## #6143: the files a lost leadership leaves behind can now be found

The issue asks for two things ahead of the real fix (leader-side unreferenced-file GC, which this database does not
have). Both are here.

**1. The diagnostic.** `CHECK DATABASE` now reports `unreferencedFiles` - the files this node holds that no schema
component claims - as a result key and a warning naming each one, and every node publishes its own count as the
`arcadedb.ha.schema.unreferenced_files` gauge. The gauge is the surface that matters under HA: `CHECK DATABASE` is
non-idempotent, so a follower forwards it to the leader, while the orphans are on the *followers*.

It **reports only**, in both modes. A file whose reference the walk cannot follow would be data, and reclaiming disk
is not worth that risk. The classification therefore proves three shapes and refuses to guess about anything else:

- a file the `FileManager` holds with no schema component at all - what an abandoned instalment leaves on a follower
  while it runs, since the apply path creates the file and only a later schema reload builds a component;
- a bucket no type claims - what the same file becomes after that reload, and what a failed `CREATE BUCKET` leaves;
- an automatic index no type references - what an abandoned rebuild leaves, with every file the index owns reported
  rather than only the one whose component happened to be classifiable.

Everything else is treated as claimed without checking: the dictionary, a compacted sub-index, a bloom filter, a
vector index's graph file, the time-series internals, a manual index, and the edge-list buckets `GraphEngine`
reaches by deriving their names from the vertex bucket's own. That last one was found by the test, not by reading:
a diagnostic that cries orphan over a healthy database is worse than none, because the obvious response to it is to
delete something. `Issue6143UnreferencedFilesTest.aHealthyDatabaseHasNothingToReport` builds one of everything and
demands silence, before and after a reopen.

**2. Coverage for the SEVERE branch.** It needed a step-down to land between an instalment and the unwind, so it was
reasoned about rather than tested. The compensating removal is now submitted through an injectable submitter and
reports what it did as a return value, so a submitter that throws is - as far as this code can tell - exactly a node
that has lost the term. `Issue6143AbandonedInstalmentRetirementTest` covers all five outcomes, including that the
failure never throws (it runs from a `finally` usually unwinding the build's own exception) and that a file this
node still holds is left alone with nothing submitted.

**Considered and not done:** having a follower reclaim the files itself when the next *publishing* schema entry
arrives - it knows which files an unpublished instalment sequence created, and a publication is the signal that the
sequence is over. It is a real design, and it is a file deletion driven by inference on the replica, on a path whose
history (#4083, #4743, #5443, #5492) is silent divergence rather than exceptions. The impact being reclaimed is
wasted disk. That trade wants its own change, with the leader-side GC the issue describes.

## #6144: instalments are measurable, per database

The count was a JVM-wide `AtomicLong` whose only consumer was a test, and the per-instalment detail was a
`HALog.DETAILED` line - which an operator debugging "every write on this database stalled for a while" could only
enable by reproducing the stall.

Instalments are now counted and **timed** per database and exported as `arcadedb.ha.schema.instalments`,
`arcadedb.ha.schema.instalment_time_ms` and `arcadedb.ha.schema.instalment_max_time_ms`, tagged `database=<name>`.
As the issue says, the counter needed reshaping rather than just exporting: per database, because "which database
stalls its writers while it ships" is the question a multi-database server has to answer. Duration is the number
that matters - each instalment is a quorum round trip taken while the database write lock is held, so it is what a
stalled writer waits on and what eats into `arcadedb.ha.quorumTimeout` - and the max is what separates 200 fast
round trips from 3 that each waited on a slow quorum member. The time is recorded in a `finally`, so an instalment
that timed out is counted too: it held the write lock for the whole timeout, which is precisely the event being
looked for. A session that shipped instalments also logs one INFO line summarising count and total time, so the
common case is visible with no metrics backend at all.

## Verification

- `Issue6142RecordedCreatesDrainTest` (5), `Issue6143UnreferencedFilesTest` (6), `Issue6143AbandonedInstalmentRetirementTest` (6), `Issue6144SchemaInstalmentMetricsTest` (3) - all new, all green.
- `RaftSchemaWalInstalment3NodesIT` (2/2) - the 3-node instalment test, which covers the drain change end to end and now asserts the per-database counters.
- Full `ha-raft` unit lane: 422 tests, 0 failures.
- `engine`: the whole `CheckDatabase*` / `DatabaseChecker*` / `FileManagerTest` set, 89 tests, 0 failures.
- `server`: `HAReplicationMetricsTest`, `HAPendingPhase2MetricsTest`, `ServerMetricsLifecycleTest`, 0 failures.


lvca:
Round 1 addressed in 0539281 - all three points taken.

1. **`refreshRecordedFileName` synchronization.** Made `synchronized`, rather than only commented. You are right that the safety rested on "only the recording thread reaches here" - true of its single caller (`PaginatedComponent.removeTempSuffix`), but not something the next caller has to notice, and with `unshippedCreates` being a plain `LinkedHashMap` the failure mode is a corrupted map rather than a stale read. It cannot deadlock: the method touches maps only, the paths that hold this monitor across I/O (`dropFile`) take file locks *under* it, and the caller holds none of its own - `rename` released the component's channel lock before returning. The javadoc now states that reasoning, so the next person to consider unsynchronizing it has it in front of them.

2. **`count()` doing full work for a number.** Fixed properly: `scan()` and `count()` now share one `walk()` that passes each finding's reason as a `Supplier`, so the counting consumer never calls it. The gauge path builds no strings at all, leaking node or not - and sharing the walk also means the two can never disagree about what counts as unclaimed, which two implementations eventually would. Pinned by a new `countAgreesWithScan` test.

3. **Unbounded warning line.** The warning now names at most 20 files and appends " and N more (see the 'unreferencedFiles' result)". As you note, `maxWarnings` bounds how many warnings a run emits, not how long one of them is; the full list stays in the result key, which is a set a caller can read.

Re-ran: `Issue6143UnreferencedFilesTest` (7), `Issue6142RecordedCreatesDrainTest` (5), `FileManagerTest` (5), `CheckDatabaseTest` (9) - all green.

lvca:
Round 2 - thanks. Nothing blocking there, and I am taking two of the three as deliberate no-ops with the reasoning recorded rather than left implicit.

**The double `forEachOpenReplicatedDatabase` walk.** Measured before deciding: `ArcadeDBServer.getDatabaseNames()` returns `Collections.unmodifiableSet(databases.keySet())` - a view over an in-memory map, not a directory listing - and `getDatabase(name, false, false)` is a map lookup that cannot load. So the second pass is one map iteration plus one lookup per database, every 5 s. Folding it means merging `SchemaInstalmentSample` and `UnreferencedFilesSample` into one record, and those two describe genuinely different things: one is HA schema-session accounting, the other a storage-level orphan count that relates to instalments only causally. A future consumer wanting just one of them (the cluster JSON is the obvious candidate for the orphan count) would then carry the other. Keeping them orthogonal is worth more than the two map lookups.

**No short-circuit on the claimed-file walk.** Agreed on the shape of the concern - it is one `getFileIds()` per index, each taking that index's read lock - so I have written the analysis into `count()`'s javadoc rather than leaving the next person to redo it (01282986): the gate would be `(FileManager.getModificationCount(), schema version)`, and between them those two cover every way the answer can change, since a newly reportable file appeared either as a file (the first) or as a schema change that stopped claiming one (the second). Not implemented, because a cached count that goes stale is a worse diagnostic than a slightly expensive one, and this is read at a 5 s cadence with no I/O.

**`WARNED_UNREFERENCED_FILES` vs `maxWarnings`.** No change - the two are genuinely different bounds (how many warnings a run emits vs how long one of them is) and the constant's javadoc says so.

That leaves the PR at: round 1's three points fixed, round 2's three either declined with reasons above or non-actionable. I consider it ready to merge unless you see something else.

lvca:
Round 3 - finding 1 taken, finding 2 stands as documented.

**1. Shared scheduled task (`HAReplicationMetrics`).** Fixed in ad667b2, and you are right that it was the coupling rather than today's behaviour that mattered: nothing was broken, but the guarantee rested on every future refresh remembering its own try/catch, and the failure mode - `scheduleAtFixedRate` cancelling the task permanently, silently, for the rest of the process - is the worst kind for a metric. Each refresh now runs inside its own `Throwable` guard on the tick, so a failing one costs only its own gauges and the next tick tries again; a sibling cannot lose its turn either. Kept as one task rather than split: separate tasks would fix the interference but not the permanent-cancellation trap, which is what actually bites, and this module deliberately keeps its thread count down.

Extracted as `runGuarded(...)` so the guarantee is pinned by a test (`aFailingRefreshCostsOnlyItsOwnGauges`, which throws an `Error` from the first refresh and asserts the second still ran) rather than by waiting out a five-second tick.

**2. Uncached `count()` walk.** Standing, as documented in `count()`'s javadoc - and "watch in production" is the right framing. The gate, when wanted, is `(FileManager.getModificationCount(), schema version)`; the reason it is not here is that a count that goes stale is a worse diagnostic than a slightly expensive one, and this one has no I/O in it.

Re-ran the server metrics suite: `Issue6144SchemaInstalmentMetricsTest` (4), `HAReplicationMetricsTest` (6), `HAPendingPhase2MetricsTest` (4), `ServerMetricsLifecycleTest` (7) - all green.

lvca:
Round 4 - the review reads as clear to merge; two of the four observations were actionable and are in 1b9784c.

**1. "Check no other creation path bypasses `recordCreate`."** Verified rather than assumed: the only `recordedChanges.add` sites in the class are `recordCreate` itself and `dropFile`'s matching removal entry, and nothing outside `FileManager` mutates either structure (`getRecordedChanges()` is read-only at every call site in the repository). Since that invariant is exactly what #6142 depends on, it is now written down where a future creation path will be standing - `recordCreate`'s javadoc says every create must come through it, and names the symptom, which is the part that makes it stick: bypassing it does not fail locally, it means the instalment never announces the file, the followers never create it, and the pages that land in it afterwards are applied against a file only the leader has.

**3. `bindTo` overwriting `followerMetricsScheduler`.** Fixed - `startMultiGaugeRefresh` now calls `close()` first. You are right that it is pre-existing and unreachable today (the binder is created per plugin start), but the leak was silent and the guard is one line, so there is no reason for it to stay true.

**2 and 4** need nothing: the memoization gate is deliberately recorded-not-implemented, and the warning cap is working as intended.

Re-ran: `Issue6142RecordedCreatesDrainTest` (5), `FileManagerTest` (5), `Issue6143UnreferencedFilesTest` (7), `Issue6144SchemaInstalmentMetricsTest` (4), `HAReplicationMetricsTest` (6), `ServerMetricsLifecycleTest` (7) - all green.

From my side this is done: four review rounds, every actionable point fixed, and the remaining ones declined with reasons in the code rather than only in the thread.

lvca:
Round 5 - taken the one actionable note (6ef6b69); the other two are already-recorded decisions.

**Scope.** You are right that it stands out: every other check in that branch is narrowed by `TYPE`/`BUCKET` and this one is not. It also cannot be, and that is the part worth writing down - "no schema component claims this file" is a property of the whole schema, not of a type, and narrowing it would answer a different question *wrongly*: a type that does not claim a file says nothing about whether another one does. So the pass stays global and now says so in both places it can be read from - the method javadoc, and the warning itself ("this pass covers the whole database, whatever TYPE or BUCKET scope was asked for"), which is the same shape as the existing COMPRESS-ignores-the-RECORD-scope caveat a few lines up.

**`count()` unmemoized** and **negative file ids silently dropped** - both intentional and already documented at the code (the memoization gate in `count()`'s javadoc, the `-1` skip in `claim()`'s). Nothing to change; agreed they are worth an eye in production rather than a cache.

Re-ran `Issue6143UnreferencedFilesTest` (7), `CheckDatabaseTest` (9), `CheckDatabaseExtendedTest` (12) - green.

Five rounds, every actionable point fixed, remaining ones declined with the reasoning in the code. Ready to merge from my side.

lvca:
CI caught two things five review rounds had not, both about the same question - what counts as a false positive. Fixed in 70ec8d9.

**1. Super-node stripe pools.** `<Type>_sn_stripe_<n>` buckets (#5156) are listed by no type and reached by composing the type name, so every promoted hub in the suite was reported as an orphan - 17 test classes, all failing through `TestHelper.checkDatabaseIntegrity`.

The fix is deliberately NOT another entry in the list of conventions the claim walk knew about (edge-list buckets, paired external buckets). CI had just proved that list is not closed: every internal bucket names itself after the schema object it belongs to, and the next feature to add one will follow the same convention rather than register with this class. So the convention itself is what is recognised now - a bucket whose name is derived from a type's or a claimed bucket's is owned by it - which covers stripes, edge lists, external pairs and whatever comes next. It costs a false NEGATIVE (an orphan named after a surviving type goes unreported), which is the right side to fail on for a finding that invites deletion, and it does not affect the shape that motivated #6143: a session abandoned before it published left no type to derive the name from. The fixture now promotes a super-node, so the class is covered locally rather than only by CI.

**2. The finding is no longer a warning.** This is the more interesting one, and I think CI was right in a way I had not considered. `Issue5724HashIndexKeyTypeGuardTest` constructs a `HashIndex` whose arguments the constructor refuses - and the file is created before the refusal, so it is a *genuine* orphan, correctly reported. But it is not a defect in the data, and neither is a bucket created with `CREATE BUCKET` and not yet given to a type. A warning in this checker means the data is suspect, and every caller that reads an empty warning list as "this database is clean" - `TestHelper.checkDatabaseIntegrity` among them - would have started calling healthy databases unhealthy.

It now reports as its own `unreferencedFiles` result key plus an INFO line, which is where an operator reads it anyway. Pinned by an assertion that `warnings` stays empty when the key is populated.

**Not mine:** `integration-tests` failed on `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose`, which asserts a schema-violating batch surfaces its failure in the body rather than at close. Nothing in this PR touches gRPC or batch semantics, and the same test passed on two of the four earlier CI runs of this branch with identical code (dac6eec, 0128298) and failed on the other two - a flake, not a regression.

Verified locally: all 197 tests across the 17 classes CI flagged, plus the suite's own `Issue6143UnreferencedFilesTest` (7) and `CheckDatabaseTest` (9).

lvca:
Rounds 6 and 7 - both nits taken (675e8a0), everything else in them was already-recorded reasoning.

**Thread name (round 7).** Renamed `arcadedb-ha-follower-metrics` to `arcadedb-ha-metrics-refresh`. You are right that it had stopped being true: the same tick drives the per-database instalment and unreferenced-file gauges now, and a thread dump saying "follower" sends whoever is debugging those to the wrong place. `ServerMetricsLifecycleTest` pins that name, so it moved with it, and #5850's note now records the rename rather than pointing at a thread that no longer exists.

**Unconditional scan (round 6).** Documented rather than gated, and the distinction is why: the scan always runs while only its LOG LINE honours `verboseLevel`. The result key has to be truthful on every run - a caller reading `unreferencedFiles` as empty must be able to conclude there are none, not that the run happened to be quiet - and the walk is in-memory with no I/O, so there is nothing worth gating.

**`count()` unmemoized** (both rounds) - standing, with the invalidation key recorded in the javadoc.
**`LOGGED_UNREFERENCED_FILES` as a constant rather than a config knob** - agreed, and deliberately consistent with `LOCK_POST_DROP_ATTEMPT_MULTIPLIER` right above it; a knob nobody would turn is a knob to maintain.

Re-ran `ServerMetricsLifecycleTest` (7), `Issue6144SchemaInstalmentMetricsTest` (4), `HAReplicationMetricsTest` (6) - green.

lvca:
CI update: the stripe-pool and warning fixes worked - the unit lane went from 17 failing classes to **one failure in 11801 tests**, and that one is a latent race this PR exposed rather than caused. Fixed in aab90d2.

`TypeRenameComponentNamingTest.aFailedIndexRenameRollsBackTheIndexesAlreadyRenamed` snapshots the component files before a rename it expects to fail, then asserts the rollback put every file back. The snapshot carried each file's **length** as well as its name - and a length is whatever the page flush has got round to. A bucket whose dirty page has not reached disk is 0 bytes at the "before" snapshot and one page long once anything flushes it, so an assertion whose message reads "every component file is back under its original name" failed over a size nobody had asked about:

```
expected: {"Multi_0.1.65536.v0.bucket"=0L, ...}
 but was: {"Multi_0.1.65536.v0.bucket"=65536L, ...}
```

It passed 10/10 on the previous commit of this branch and passes locally on the failing code, which is the signature of a timing race rather than a regression - and the timing moved because 17 classes that used to fail early in the same fork now run to completion. It compares the name set now, which is what a rename rollback is actually judged on, so it cannot recur.

Also confirmed unrelated: `integration-tests` failed again on `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose`, the same gRPC test that passed on two of this branch's earlier runs with byte-identical gRPC code.

lvca:
CI update after aab90d2: **the engine module is fully green - 11801 tests, 0 failures**, so both the stripe-pool false positive and the rename-rollback flake are closed.

One failure left in the run, and it is not this PR's: `Issue6116ChecksumsPointInTimeTest.computingTheChecksumsNoLongerSuspendsPageFlushing` in `ha-raft`, on its **control** assertion:

```
[the fallback path must still freeze the files for the whole read]
Expecting actual: 0.8953148525374192  to be greater than: 0.9
```

That is a sampler measuring what fraction of a checksum computation ran with page flushing suspended, missing its threshold by half a percent on a 4-CPU runner. It passed on the two previous CI runs of this branch, whose `ha-raft` code is byte-identical to this one (the last commit touched an engine test only), and nothing here goes near the checksums or flush-suspension paths.

Re-running the failed jobs rather than changing code. The `integration-tests` lane's `Issue6070GrpcGraphBatchIT` failure is the same known flake as before.

lvca:
Follow-ups from this PR filed as #6168 (six items: memoization gate for the unreferenced-file count, the actual unreferenced-file reclaim, two flaky tests, the 'reached by name' convention with two homes, and the chronically red integration lane).

Issue #6141, opened 2026-08-13
A concurrent update of the same placeholder-backed record is silently lost, with no conflict raised


lvca:
Found while measuring #6129 (PR #6140), which does not change this behaviour.

## What happens

Two transactions updating the SAME record silently lose one of the two writes, with no `ConcurrentModificationException` at all, when that record has spilled into a **placeholder**: an 8-byte pointer in its own slot, and the content on another page.

```
T1: begin; R.set("v", A).save()      // R is a placeholder-backed record
T2:        transaction { R.set("v", B).save() }   // commits
T1: commit                            // succeeds. B is gone, no conflict raised.
```

Confirmed on `main` with a `Holder` type of one bucket, a record grown to 30 KB so it becomes a placeholder pointer on page 0 with its content on page 1.

## Why

A record update is deferred. `TransactionContext.addUpdatedRecord` pins **the record's own page** (`LocalBucket.fetchPageInTransaction`) when `save()` is called, and the write itself runs in `commit1stPhase`, after the file's commit lock is taken. So:

- the page holding the placeholder POINTER is pinned, but neither transaction writes a single byte to it - an update whose content record can absorb the new value leaves the pointer alone, and `commit1stPhase` drops a modified page whose modified range is empty. Nothing to version-check;
- the CONTENT page is not pinned by anything. Both transactions load it fresh, under the lock, at the newest committed version - so its version check can never fire either.

The result is a blind last-writer-wins on the content record. The same reasoning applies to any update that reaches a page the transaction did not pin, but the placeholder is the shape where the record's *whole* content lives there.

Multi-page (chunked) records are not affected: any update of one rewrites its head chunk on the pinned page, so the page-version check fires. That is exactly the property PR #6140 had to preserve explicitly, with `LocalBucket.chunkChainTailFingerprint`, once the head chunk became mergeable.

## Worth considering

1. Fingerprint the content record the way #6140 fingerprints a chunk chain's tail: take it in `addUpdatedRecord` (the moment the record is taken for update, the same moment its page is pinned) and compare it in `updateRecordInternal` before overwriting the content. A mismatch is a genuine concurrent modification and should raise a retryable `ConcurrentModificationException`. Cost is proportional to a write the transaction is doing anyway.
[...]

Issue #332, opened 2022-02-18
Support for records bigger than a page


lvca:
### Discussed in https://github.com/ArcadeData/arcadedb/discussions/320

<div type='discussions-op-text'>

<sup>Originally posted by **lvca** February  7, 2022</sup>
Currently (version <= 22.02.1) ArcadeDB does not support storing records larger than a page. This limitation must be overcome to:
- store large records, especially binary records (blobs)
- avoid worrying too much about importing the database if the business requirements change and larger records are needed.

First, a quick introduction to how records are stored on a page. The first part of the page contains the header with slots with pointers where the record content is located on the page. There can be a maximum of 2048 (`Bucket.DEF_MAX_RECORDS_IN_PAGE`) records on a page (2048 slots). The record content is prefixed by it size. The size is stored with a varint (variable integer). The record size is the length of the record:
- 0 = deleted record
- -1 = placeholder pointer that points to another record on another page
- <-1 = placeholder content pointed from another record in another page

The record size is stored as a varint (variable integer size). The minimum size of a record stored on a page is 5 bytes. If the record is smaller than 5 bytes, it is filled with blanks.

In order to store large records, we must split the record into chunks and save all of them in sequence as a linked list. To let the bucket know it's a chunk of a record, the new `size = -2` must be used. 

_NOTE: This doesn't interfere with the current content, because negative sizes are considered placeholder content, but records cannot be smaller than 5 bytes, so it's not possible to encounter a placeholder content record with record size = -2. The minimum would be -5._

So a record with size -2 will contain the first chunk of the record. The record content will have the following information before the actual chunk of data:
- the chunk size in bytes as varint
- the location of the next chunk, stored as a placeholder pointer containing the record slot in the same bucket

In order to read the entire record, the record is built in memory chunk after chunk, jumping between pages until the pointer to the next chunk is 0 (zero).

The first chunk has a record size = -2, while the other chunks will have a record size = -3. This allows the `scan()` and `count()` methods to skip those records once encountered.</div>

PR #6153, opened 2026-08-13
fix(#6149): a spilling record asks the page for the chunk header before falling back to a placeholder

### LINKED ISSUE #6149: Record placeholders are the pre-chunk spill mechanism, and the 14 bytes that force them are usually available
Noticed while working on #6129 (PR #6140), where the placeholder turned out to be the least valuable of the three shapes that issue listed - and the one behind the still-open #6141.

## Placeholders are still reachable, but they are the pre-chunk mechanism and there is exactly one way to get one

`LocalBucket` has a single site that creates a placeholder pointer, in the "record must spill out of the page" branch of `updateRecordInternal`:

```java
int availableSpaceInCurrentPage = (int) (recordSize[0] + recordSize[1]);
if (lastRecordPositionInPage == recordPositionInPage)
  availableSpaceInCurrentPage += page.getMaxContentSize() - pageOccupiedInBytes;

// TODO: LOOK FOR 1/2 OF THE RECORD SIZE
if (availableSpaceInCurrentPage < 2 + LONG_SERIALIZED_SIZE + INT_SERIALIZED_SIZE) {
  page.writeNumber(recordPositionInPage, RECORD_PLACEHOLDER_POINTER);
  final RID realRID = createRecordInternal(record, true, false);   // the only isPlaceHolder=true call in the class
  ...
} else {
  // SPLIT THE RECORD IN CHUNKS ... ISSUE https://github.com/ArcadeData/arcadedb/issues/332
```

So a placeholder happens when, and only when, a record that has to spill occupies **fewer than 14 bytes** in its slot (plus the page tail if it is the last record) - too little to host a `FIRST_CHUNK` header, which needs 1 marker + 4 size + 8 next-pointer. Everything else spills into chunks. The multi-page chunk chain arrived later (#332) and took over every case the placeholder used to serve except that one; `MINIMUM_RECORD_SIZE`'s own comment records the overlap ("5 BYTES IS THE SPACE REQUIRED TO HOST THE PLACEHOLDER AND 1ST CHUCK FOR MULTI-PAGE CONTENT").

It is not dead code - `Issue5279ConcurrentUpdateTest.aPlaceholderPointerRebuiltUnderConflictFallsBackAndStaysCorrect` and `Issue6129ChunkedSlotMergeTest.updatingPlaceholderContentCommutesWithItsPageNeighbours` both produce one and assert `totalPlaceholderRecords == 1` - but every route to it is that single 14-byte test.

## The 14 bytes are usually there for the asking

The branch is reached because the page could not host the record's **full** new size. That says nothing about whether it has 14 bytes: `growRecordInPage` already shifts the records that follow to the right, and refuses only when

```java
additionalSpaceNeeded >= page.getMaxContentSize() - pageOccupiedInBytes
```

Getting a 9-byte slot up to 14 needs 5 more bytes, not 30 KB of them. A page that has just refused a large growth commonly has kilobytes free - in the #6129 test above, page 0 has roughly 1 KB free when the placeholder is created. The code simply never asks.

## Proposal

Before falling back to the placeholder, try to grow the slot to the chunk-header minimum through the existing in-page shift, and spill into chunks. That reuses machinery already on this path, and leaves the placeholder reachable only when the page cannot spare ~13 bytes at all - if it can be shown to be unreachable, the whole `RECORD_PLACEHOLDER_POINTER`/`RECORD_PLACEHOLDER_CONTENT` family becomes removable (they are read in ~10 places across the scan, check, delete, update, compaction and space-accounting paths).

## Why it is worth more than the code it deletes

1. **It would close #6141 by construction.** That silent lost update - two transactions updating the same placeholder-backed record, neither raising a conflict - is placeholder-only. Chunked records do not have it, because any update rewrites the head chunk on the page the transaction pinned.
2. **It is the one record shape the disjoint-slot merge still cannot replay** end to end: the pointer rewrite changes two pages at once (#6129 covers the content record, not the pointer).
3. Records that outgrow their page would have exactly one representation instead of two, and every reader (`check`, `scan`, `delete`, `compressPage`, `getPageOccupiedInBytes`, `getOrderedRecordsInPage`) would lose a branch.

Backward compatibility is the obvious constraint: existing databases contain placeholders, so the readers have to stay until a format migration retires them. Only the *writer* would stop producing new ones.

lvca:
Fixes #6149.

## What the issue found

The record **placeholder** - an 8-byte pointer in the record's slot, the content on another page - is the
pre-chunk spill mechanism. The chunk chain (#332) took over every case it used to serve except one, and
`LocalBucket` has exactly one site left that writes a placeholder pointer: the spill branch of
`updateRecordInternal`, when the slot cannot reach the **14 bytes** a `FIRST_CHUNK` header needs (marker, chunk
size, next-chunk pointer, one byte of content).

Those 14 bytes are normally there for the asking. The branch is reached because the page could not host the
record's **full** new size - which says nothing about the five-odd bytes between a 9-byte slot and a chunk
header. `growRecordInPage` already shifts the records that follow to the right, and refuses only when the whole
growth exceeds the free tail. In the #6129 placeholder test, page 0 had ~1 KB free at the moment the placeholder
was created. The code never asked.

## What this does

Before falling back, the spill enlarges the slot to the chunk-header minimum through that same in-page shift and
spills into chunks. The shift is extracted as `shiftFollowingRecordsRight`, now shared by `growRecordInPage`, the
spill, and the commit-time replay of both, so all three move records identically.

**Only the missing bytes are taken.** A head chunk's footprint is fixed for the life of the record
(`updateMultiPageRecord` only ever rewrites inside it), so a bigger bite would lock page space away from every
other record on the page for good.

## The part the issue did not mention, and the reason this is worth more than the shift

The issue proposed the write-side shift. On its own that is a regression: `rebaseRecordOnPage` refuses a
`SLOT_KIND_RECORD_SPILLED_TO_CHUNK` slot whose committed footprint cannot host the head chunk, so **every**
enlarged spill would fail to merge and fall back to a full-transaction retry - reintroducing exactly the spill
starvation #6129 removed (a transaction whose spill loses the race retries and spills again, so on a contended
page the records that have to spill can starve instead of converging).

So the replay learns the same move: it re-does the shift on the newer committed page, from that page's own slot
table, and refuses only when the page can no longer spare the bytes. The slot footprint is pinned by the
byte-for-byte pre-image check the replay already makes, so the shift the write made and the one the replay makes
are the same size by construction; only the free tail can differ, and a page that no longer has it sends the
transaction to a normal retry. `aSpillThatHadToEnlargeItsSlotIsStillReplayedOnACommittedPage` pins this down, and
it was verified to be non-vacuous: forcing that branch to refuse turns it red by raising `ConcurrentModificationException`.

## Considered and not done

- **Take a bigger bite of the free tail** (say half of it) to reduce the number of continuation chunks. Rejected:
  the head chunk never shrinks back, so this permanently denies page space to the record's neighbours in exchange
  for a saving on a record that is spilling anyway.
- **Compact the page** (`compressPage`) to reclaim holes left by deletes before giving up. It relocates every
  record on the page, which is not slot-merge replayable, and it buys a case that is already vanishingly rare
  after this change.
- **Retiring the `RECORD_PLACEHOLDER_*` family.** Not possible yet: existing databases contain placeholders and
  ~10 read branches (scan, check, delete, update, compaction, space accounting) resolve them. Only the *writer*
  stopped producing new ones. That needs a format migration, which is out of scope here.

## What is left for the placeholder

A page with **no free tail at all**. No insert can produce one - `getAvailableSpaceInPage` keeps
`SPARE_SPACE_FOR_GROWTH` in hand - but a spill can, because the head chunk of a *last* record takes the whole
tail and ends the page exactly at its maximum content size. That is now the fixture the tests use
(`sealFirstPage`), and it is why the two existing tests that need a placeholder had to change: filling a page
with inserts no longer produces one.

An **existing** placeholder converts to a chunk chain by itself the first time its content record can no longer
absorb an update, because that slot rebuild goes through the very same spill branch
(`anExistingPlaceholderIsRebuiltAsAChunkChainWhenTheContentCannotAbsorbTheUpdate`).

## Relation to #6141

#6141 (the silent lost update on a placeholder-backed record) is **not** closed by this. It is now unreachable
for records written by this version on a page with any free tail, which is the overwhelming majority - but
legacy placeholders and the sealed-page fallback still have it. #6141 stays open.

## Tests

New `engine/src/test/java/com/arcadedb/database/Issue6149PlaceholderPrefersChunksTest` (6 tests):

- `aSmallRecordSpillingWithRoomLeftInThePageBecomesAChunkChain` - the issue itself
- `theShiftThatMakesRoomForTheChunkHeaderPreservesEveryOtherRecord` - the moved neighbours keep RID and content
- `everyRecordOfAPageCanSpillInTurn` - 24 records of one page spill one after another
- `aSpillThatHadToEnlargeItsSlotIsStillReplayedOnACommittedPage` - the merge replay, with the merge-off premise proved
- `aSmallRecordSpillingOutOfASealedPageStillFallsBackToAPlaceholder` - the fallback stays correct
- `anExistingPlaceholderIsRebuiltAsAChunkChainWhenTheContentCannotAbsorbTheUpdate` - legacy conversion

Every test ends with `CHECK DATABASE` asserting zero errors and zero auto-fixes.

Updated: `Issue6129ChunkedSlotMergeTest.updatingPlaceholderContentCommutesWithItsPageNeighbours` and
`Issue5279ConcurrentUpdateTest.aPlaceholderPointerRebuiltUnderConflictFallsBackAndStaysCorrect` now seal page 0 to
get a placeholder (and 5279's concurrent neighbour write became a same-length overwrite, because a sealed page
leaves no room to grow into).

Full `engine` module suite: **11797 tests, 0 failures, 0 errors**. Full project compiles.

## Files changed

- `engine/src/main/java/com/arcadedb/engine/LocalBucket.java`
- `engine/src/test/java/com/arcadedb/database/Issue6149PlaceholderPrefersChunksTest.java` (new)
- `engine/src/test/java/com/arcadedb/database/Issue6129ChunkedSlotMergeTest.java`
- `engine/src/test/java/com/arcadedb/database/Issue5279ConcurrentUpdateTest.java`
- `engine/CLAUDE.md`


lvca:
Thanks - both review nits addressed in a1c899f, comments only, no behaviour change.

**1. The budgeted `2` vs the replay's `Binary.getNumberSpace(FIRST_CHUNK)`** - documented on `MINIMUM_SPACE_FOR_FIRST_CHUNK`, with a pointer to it from the replay. The reconciliation, stated so nobody has to trace it again: both sides size the whole footprint and never the marker alone. `writeMultiPageRecord` derives `chunkSize` from the room it is given minus the marker it *actually* wrote (`14 - 1 - 4 - 8 = 1` byte of content), and the replay re-derives the same 14 as `getNumberSpace(FIRST_CHUNK) + body.length`, `body` being the chunk image those very bytes produced. Raising the budget only makes the head chunk carry one more byte of content; it cannot make the two disagree.

**2. The two coverage spans - kept apart deliberately**, and now the comment says why, because merging them is a latent bug rather than a tidy-up. `growRecordInPage` can wrap its shift and its write in one span because everything it writes is its own record. `writeMultiPageRecord` cannot: when the chunk chain comes back to *this* page, it writes a CONTINUATION chunk onto it through inline record-table writes that declare no coverage of their own - it poisons the page instead. Those writes would inherit an outer span left open around the shift and be declared slot-merge replayable, which they are not. The comment now ends with "Do not merge the two spans."

Re-ran `Issue6149PlaceholderPrefersChunksTest`, `Issue6129ChunkedSlotMergeTest`, `Issue5279ConcurrentUpdateTest` and `MutablePageWriteCoverageTest`: 31 tests, all green.

Unrelated to this PR: the red `Analyze (python)` check is a CodeQL SARIF-upload failure (0 findings, 0 diagnostics, 107/107 files scanned) - this branch changes no Python.

lvca:
Round 2 addressed in 6c8d7cd, comments only.

**`MINIMUM_RECORD_SIZE`'s stale comment - fixed, and you were right that it reads as a contradiction.** It claimed "RECORD SIZE CANNOT BE < 13 BYTES IN CASE OF UPDATE AND PLACEHOLDER", which conflates it with the chunk-header size, and it now sat two lines above a `MINIMUM_SPACE_FOR_FIRST_CHUNK` of 14. The 5 has nothing to do with spilling: a placeholder CONTENT record stores its size **negated**, so a record shorter than 5 bytes would write a -1..-4 that `RECORD_PLACEHOLDER_POINTER` (-1), `FIRST_CHUNK` (-2) and `NEXT_CHUNK` (-3) already own. That is the whole reason for the 5, and for `RECORD_PLACEHOLDER_CONTENT` = -5 being the boundary between "this is a size" and "this is a marker". Both constants now say so.

**The untested branch-boundary - documented rather than tested, deliberately.** You are right that "not-last record, free tail exists but is smaller than the enlargement" has no test of its own. What the sealed-page test already covers is the *same* comparison (`missing >= freeTailInPage`) at `freeTailInPage == 0`; the neighbouring values differ only in operands. I could not build them honestly: a tail of 1..7 bytes is unreachable through the public API, because inserts always leave at least `SPARE_SPACE_FOR_GROWTH` (32) bytes and a spill always takes the tail down to exactly 0. Every route I tried needed byte-exact knowledge of the page geometry that no API exposes, which would make the test assert its own arithmetic rather than the engine's. That reasoning is now written on the sealed-page test so the next reader does not re-derive it.

Re-ran `Issue6149PlaceholderPrefersChunksTest`, `Issue6129ChunkedSlotMergeTest`, `Issue5279ConcurrentUpdateTest`, plus `RandomDeleteTest`, `RecordRecyclingTest` and `LargeRecordsTest` (the record-lifecycle suites most exposed to a page-layout change): 28 tests, all green. The full `engine` module was already green at 11797 tests on the code commit, and nothing since has changed behaviour.

lvca:
Round 3 addressed in f768994.

**Test helper duplication - taken, not deferred.** You framed it as "worth considering next time one of these is touched", but this PR *was* that time: it added a third copy of `fillFirstPage`/`bucketStats`/`checkDatabase`/`numberProperty` and a brand-new fourth helper (`sealFirstPage`) duplicated twice more. The precedent was the problem, so it is gone: `BucketPageLayoutTestSupport extends TestHelper` now owns all five, and `Issue5279ConcurrentUpdateTest`, `Issue6129ChunkedSlotMergeTest` and `Issue6149PlaceholderPrefersChunksTest` extend it. The two inline `check database` blocks in 5279 collapse into the shared `checkDatabase()` too. Net -185/+5 lines across the three, plus the new support class.

These fixtures are worth centralising for a reason beyond line count: the layout they build - a full page, a page with no free tail at all - *is* what the assertions mean. Change how a page fills up or how a spill claims the free tail and every one of these tests has to be re-read, which is only tractable with one copy to change. `fillFirstPage` joins the caller's transaction when there is one, so it still works both inside a fixture transaction (as 6129 uses it) and standalone.

**The other two points - confirming your read, no change made:**

- **The strict `<` free-tail margin is deliberate here, not just inherited.** Matching `growRecordInPage`'s `additionalSpaceNeeded >= spaceAvailableInCurrentPage` is the point: the write path and the commit-time replay have to agree byte-for-byte on when a page can host the enlarged spill, and the cheapest way to guarantee that is for both to use the same predicate the rest of the file already uses. Spending the last byte would buy a vanishingly rare case at the cost of a second policy to keep in sync.
- **`updatePageStatistics` accounting only the shift, not the head chunk's footprint** - correct, and left as is. That gap predates this PR (`writeMultiPageRecord` never fed the first chunk into the cache), the cache is a placement hint that every allocation re-verifies against the page itself, and closing it properly belongs with the spill accounting as a whole rather than smuggled into this change.

23 tests across the three refactored classes, all green.

lvca:
Round 4 addressed in e9d001a.

**Duplicated enlargement arithmetic - taken, and you put your finger on the weak spot.** `shiftFollowingRecordsRight` was extracted for reuse, but the *decision* (shift or not, by how much, with what margin) was still written twice, so the invariant the whole design rests on - "the shift the write made and the one the replay makes are the same size by construction" - actually held by convention plus one test. That is a gap between what the comments claim and what the compiler enforces, which is exactly the kind of thing that rots.

`slotEnlargementForChunkHead(roomRequired, roomAlreadyAvailable, freeTailInPage, isLastRecordInPage)` is now the single place. It returns 0 when the slot is already big enough, the byte count to shift by, or -1 when the page cannot provide it - which the write turns into the placeholder fallback and the replay into a plain retry. The two callers now differ only in what they *ask for*: the write asks for `MINIMUM_SPACE_FOR_FIRST_CHUNK`, the replay asks for the exact size of the head chunk it has to put back (the same 14 for an enlarged spill, and larger for a slot that was the page's last record when it spilled and took the whole tail). The margin and the last-record rule cannot drift apart any more.

Behaviour-preserving: the helper's branches are the same predicates in the same order both call sites already had.

**The `<` vs `<=` boundary** - confirming again it is deliberate rather than inherited by accident. Both paths keeping `growRecordInPage`'s "leave at least one spare byte" margin is what makes a growth and a spill judge the same page identically; spending the last byte would buy a vanishingly rare case in exchange for a second margin to keep in sync across three call sites. Now that the margin lives in one helper, that is stated there.

Targeted suites green after the change: `Issue6149PlaceholderPrefersChunksTest`, `Issue6129ChunkedSlotMergeTest`, `Issue5279ConcurrentUpdateTest`, `Issue5596MergeCoverageTest`, `RandomDeleteTest`, `LargeRecordsTest` - 30 tests. Full `engine` module re-run in flight; I will report if anything moves.

(Noted on your last point: `mvn` genuinely does run here, and the 11797-test figure is a real local run of the `engine` module, not a claim from the description.)

lvca:
Round 5 addressed in b8a1b1b. Both nits taken; noting that you called them non-blocking and matching existing convention, but both sit in the two helpers this PR made *shared*, which is exactly where "matches the existing convention" stops being a good enough reason.

**Partial `@param` lists** - completed on `slotEnlargementForChunkHead` and `shiftFollowingRecordsRight`. These two are now the single decision and the single mechanic behind the whole change, reached from three call sites; they are the last places to leave a reader guessing what a parameter means.

**The assertion outside the `if`** - moved inside. You were right that it is carried over verbatim rather than introduced here, but it is worse than merely redundant: for the slots left alone it is *vacuously* true (their offset is below `from`, and a freed slot reads 0), so it asserted nothing while looking like it did. Inside the branch it states the invariant it is actually about - every record that MOVED must still start inside the page - and the comment says so, since the next person to widen this helper needs to know which records it is guarding.

32 tests green across the layout, slot-merge and record-lifecycle suites (`Issue6149PlaceholderPrefersChunksTest`, `Issue6129ChunkedSlotMergeTest`, `Issue5279ConcurrentUpdateTest`, `Issue5596MergeCoverageTest`, `RandomDeleteTest`, `RecordRecyclingTest`, `LargeRecordsTest`).

From my side this is complete: five review rounds, no correctness bug found in any of them, and every point either implemented or answered with a reason recorded in the code rather than only in a comment thread.

lvca:
Round 6 addressed in a1b8bbc.

**The unguarded fixture invariant - good catch, and fixed at the right level.** You are right that `sealFirstPage` asserted its contract in a comment. It now checks the thing that can actually stop being true: the sealing record must have SPILLED, because the spill is the only mechanism that eats the free tail. If a page-geometry change ever let 70 KB stay inside the page, every test built on this fixture would have gone on passing while quietly exercising a page that still has room - and the breakage would have surfaced as confusing assertions in six downstream tests instead of one loud failure naming the fixture.

(I did consider asserting the tail is literally zero, and did not: no public API exposes it, so the check would have had to reconstruct the page geometry by hand - the same reason the "small but nonzero free tail" boundary has no test. Asserting the spill happened is the honest, observable form of the same guarantee.)

**Javadoc phrasing** - fixed, in the new javadoc and in the `getPageOccupiedInBytes` one it was copied from, so the two keep reading the same.

**On confirming the build** - understood that you could not run it. I can, and did: the full `engine` module suite is green at **11797 tests, 0 failures, 0 errors, 23 skipped**, on a clean run of the current branch. Worth noting one thing I got wrong and corrected: an earlier full run of mine came back with two `NoClassDefFoundError`s in `GAVEligibilityTest` - that was self-inflicted, a targeted `mvn test` of mine recompiling into the same `engine/target` while the background suite ran against it. `GAVEligibilityTest` is 42/42 in isolation and 42/42 in the clean run. Nothing to do with this change, but I would rather state it than leave a red run unexplained.

Six review rounds, no correctness bug found in any of them, every point either implemented or answered with the reason recorded in the code. From my side this is done.

lvca:
Round 7: no correctness bug, and both remaining observations are pre-existing and explicitly non-blocking. Closing this out.

**The head-chunk page-statistics gap** - agreed it is pre-existing and out of scope here, so it is filed rather than smuggled in: **#6154**. Worth recording why it is not a one-liner: the head chunk is the chunk that can consume the *most* space on its page (a spilling LAST record is handed its own footprint plus the entire free tail), so a spill can take a page from kilobytes-free to zero-free without `freeSpaceInPages` hearing about it - and whatever fix lands there has to net against the `updatePageStatistics` call this PR added for the shift, or it will drive a page`s free space negative.

**`engine/CLAUDE.md` being a lot of prose to keep in sync** - taken, and deliberate. That file exists for things the code does not tell you or actively misleads you about, and the placeholder is exactly that: still reachable, still read in ~10 places, but no longer written except on one page shape that inserts cannot even produce. The next person to touch this needs to know that before they conclude the fallback is dead code and delete it. If the mechanism is ever actually retired, that section goes with it.

Final state: seven review rounds, no correctness bug found in any of them, every point either implemented or answered with the reason recorded in the code rather than only in a thread. Full `engine` module suite green at **11797 tests, 0 failures, 0 errors** on a clean run of the current branch.

Ready for your review.

Issue #4824, opened 2026-06-30
Single global `applied-index` file vs per-database snapshot installs — applied-index/data skew on a multi-DB state machine


ruispereira:
**Affected version:** 26.6.1
**Component:** `ArcadeStateMachine` (`.raft/applied-index`) vs `SnapshotInstaller.atomicSwap`
**Severity:** MEDIUM

### Summary

The persisted applied index is a single scalar at
`<SERVER_DATABASE_DIRECTORY>/.raft/applied-index` for a state machine that
multiplexes all databases. A per-database snapshot install jumps that DB's on-disk
state without a corresponding meaning for the global counter, so the persisted
value can't be trusted per-database.

### Code

`ArcadeStateMachine.java:1134-1142` (global file), `:189-225` (`reinitialize` restore);
`SnapshotInstaller.java:530-571` (per-DB swap, never touches applied-index).

### Impact

Recovery decisions in `reinitialize` (e.g. `snapshotIndex > persistedApplied +
tolerance` → download) are computed against a value that mixes databases. Mostly
latent thanks to engine-level page-version guards, but a genuine bookkeeping
defect.

### Suggested fix

Rely solely on Ratis' snapshot index, or make applied-index tracking
per-database-aware.

PR #6159, opened 2026-08-13
fix(#6111): do not let a stale snapshot marker satisfy a linearizable read before its re-download lands

### LINKED ISSUE #6111: reinitialize() advances lastAppliedIndex (and now notifies) before a flagged stale-snapshot re-download completes
## Summary

Found during review of #6110 (fix for #5846). Not a regression from that PR - the underlying behavior
predates it - but #6110's fix (adding an explicit `notifyApplied()` call) makes the window tighter for
an already-blocked waiter, so filing separately rather than folding a broader fix into that PR.

## Detail

In `ArcadeStateMachine.reinitialize()`:

```java
// ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
if (persistedApplied >= 0 && snapshotIndex > persistedApplied + snapshotGapTolerance) {
  LogManager.instance().log(this, Level.INFO,
      "Snapshot index %d is ahead of persisted applied index %d, will download from leader when available",
      snapshotIndex, persistedApplied);
  needsSnapshotDownload.set(true);
  // ... watchdog scheduled to trigger an async download from the leader ...
}
lastAppliedIndex.set(snapshotIndex);      // <-- unconditional, even when needsSnapshotDownload just became true
updateLastAppliedTermIndex(snapshotInfo.getTerm(), snapshotIndex);
raftHA.notifyApplied();                   // <-- added by #6110
```

When the gap check decides the on-disk snapshot is stale relative to the leader
(`snapshotIndex > persistedApplied + snapshotGapTolerance`), it flags `needsSnapshotDownload` and
schedules an async re-download/reconciliation - but the very next lines advance
`lastAppliedIndex` to the (possibly stale) `snapshotIndex` anyway and now also notify.

A `LINEARIZABLE` or `READ_YOUR_WRITES` read waiting on a target index `<= snapshotIndex` at that
moment can now wake and proceed immediately, potentially serving data from the stale local snapshot
before the flagged re-download/reconciliation has actually happened.

This is not new: a waiter that started blocking *after* `lastAppliedIndex.set(snapshotIndex)` already
saw the condition satisfied and returned without waiting at all, pre-#6110. What #6110 changes is that
an *already-blocked* waiter (one that started waiting before this line ran) now also wakes immediately
via the explicit notify, instead of only re-checking on its next scheduled poll/timeout - narrowing the
timing difference between "arrived before" and "arrived after" this line, and generally making the
already-possible early wake more likely to be observed.

## Suggested follow-up

Worth deciding intentionally whether `lastAppliedIndex` (and now the notify) should be deferred until
after `needsSnapshotDownload` resolves (the watchdog-triggered or leader-change-triggered download
completes), rather than being set optimistically from the on-disk marker while it's already flagged as
possibly stale. That likely touches the PAUSED->RUNNING lifecycle transition semantics
`StateMachineUpdater.reload()` depends on, so it needs its own design pass rather than a quick patch.

## References

- Original bug: #5846
- Fix that surfaced this during review: #6110

robfrank:
Closes #6111

## Summary

`ArcadeStateMachine.reinitialize()` can find a Ratis snapshot marker whose index runs ahead of the persisted `.raft/applied-index` by more than `HA_SNAPSHOT_GAP_TOLERANCE`. It flags `needsSnapshotDownload`, schedules an asynchronous re-download, and then - before that download has run - advanced `lastAppliedIndex` onto the marker and (since #6110) called `notifyApplied()`. This PR keeps the node honest for the whole window: the ArcadeDB-side applied counter stays on the persisted position, a read floor is published for the apply waiters, the "applied advanced" notification is withheld until a resync has actually restored the state, and the waiters clamp to the floor so a LINEARIZABLE read fails (503) instead of being served from databases that stop short of the index it asked for.

## Root cause, and why the obvious fix is not the fix

While investigating I found the mechanism is one level below where the issue puts it.

`RaftHAServer.waitForAppliedIndex()` / `waitForLocalApply()` do **not** read `ArcadeStateMachine.lastAppliedIndex`. Their predicate is `RaftHAServer.getLastAppliedIndex()`, which is `division.getInfo().getLastAppliedIndex()` -> Ratis' `StateMachineUpdater.appliedIndex`. Ratis seeds that from `getLatestSnapshot().getIndex()` - in `StateMachineUpdater`'s constructor on startup, and again in `reload()` immediately *after* `reinitialize()` returns (verified against the ratis-server 3.2.2 sources). ArcadeDB's own `AtomicLong` feeds `takeSnapshot()` and the phase-2 replay floor, and is not on the waiter path at all.

So simply not advancing `lastAppliedIndex` would have changed nothing observable: Ratis reports the marker index as applied either way, and every waiter targeting an index inside `(persistedApplied, snapshotIndex]` finds its condition already satisfied and reads the stale local databases. Fixing this needs the waiters to stop trusting the raw Ratis value while a flagged re-download is outstanding.

## What changed

**`ArcadeStateMachine`**

- New `staleSnapshotAppliedFloor` (`AtomicLong`, -1 = none): the highest index whose data is genuinely on disk, published by `reinitialize()`'s snapshot-gap branch and set *before* the marker is seeded, so no waiter can observe the seeded index without also observing the floor that qualifies it.
- The gap branch no longer advances `lastAppliedIndex` onto the marker (it stays on `persistedApplied`, which is what `takeSnapshot()` may hand Ratis as a durability checkpoint) and no longer calls `notifyApplied()`.
- The Ratis-facing `updateLastAppliedTermIndex(snapshotTerm, snapshotIndex)` is deliberately **unchanged**: it is the only replay position Ratis has, and `StateMachineUpdater.reload()` requires it to match `getLatestSnapshot()`.
- The floor is cleared only where a resync actually restored the state - `notifyInstallSnapshotFromLeader()` (before its notify, so a woken waiter re-checks against restored state) and a successful `triggerSnapshotDownload()` - never where one is merely requested or in flight. A later `reinitialize()` that finds no gap clears it too.
- A successful `triggerSnapshotDownload()` now also records the marker index as the persisted applied position of every present database, mirroring what the leader-driven install already did. Without it the same gap is re-detected on every restart and the node re-downloads forever.
- A **failed** `triggerSnapshotDownload()` re-arms `needsSnapshotDownload` while a floor is outstanding, so the next leader change or the `HealthMonitor` backstop retries instead of leaving reads clamped until a restart.
- `isResyncInProgress()` counts an outstanding floor, so a node that failed its download stops advertising `/api/v1/ready` 200 (after a failure neither of the other two flags is set).

**`RaftHAServer`**

- New `getTrustedAppliedIndex()` = `getLastAppliedIndex()` clamped to the floor. Both apply waiters now use it. Reporting paths (cluster status, lag detection) deliberately keep the raw Ratis value.
- LINEARIZABLE (`throwOnTimeout=true`) keeps waiting up to `quorumTimeout` - the resync may land inside it - then fails, which is the documented contract ("fail rather than serve a value older than an already-committed write"). The message now carries the floor.
- READ_YOUR_WRITES / bookmark waiters are best-effort by contract; waiting out `quorumTimeout` on a gap only the resync can close cannot change their outcome, so they degrade immediately with a WARNING instead of burning the timeout on every read.

**`ha-raft/CLAUDE.md`** - records that `getLastAppliedIndex()` reads Ratis, not ArcadeDB's counter. That is the misleading part that cost the investigation time here.

## Test plan

New `ha-raft/src/test/java/com/arcadedb/server/ha/raft/Issue6111StaleSnapshotReadFloorTest.java`, 11 tests, all deterministic (no cluster timing): they drive the real `reinitialize()` against a real `SimpleStateMachineStorage` marker plus a real persisted applied-index file, and the waiters against the detached `RaftHAServer` harness `ApplyWaitBoundedRecheckTest` already uses.

Each test was proved able to fail. Reverting only the state-machine half (floor/counter/notify) fails 2; reverting only the waiter half (clamp + early degrade) fails 4. The "degrades immediately" tests assert the clamp is in force first, so they cannot pass vacuously by never entering the wait loop.

- [ ] `staleMarkerPublishesAFloorAndWithholdsTheAppliedNotification` - floor = persisted position, counter unmoved, `notifyApplied()` never called, resync flagged, node not ready
- [ ] `staleMarkerStillSeedsTheRatisFacingAppliedTermIndex` - the Ratis contract is untouched
- [ ] `markerWithinToleranceSeedsAndNotifiesAsBefore` - control: a trustworthy marker behaves exactly as before (#5846 notify intact)
- [ ] `aLaterReinitializeWithoutAGapDropsTheFloor`
- [ ] `staleMarkerDoesNotThrowWhenRaftHAServerNotYetAttached` - the startup `initialize() -> reinitialize()` path
- [ ] `linearizableReadFailsWhileTheFloorIsOutstanding` - the observable bug: Ratis says 5000, only 100 is backed by data, the read now fails
- [ ] `linearizableReadBelowTheFloorStillSucceeds`
- [ ] `linearizableReadIsReleasedWhenTheResyncClearsTheFloor` - asserts the wait really was held, then released promptly
- [ ] `readYourWritesDegradesImmediatelyInsteadOfBurningTheQuorumTimeout`
- [ ] `waitForLocalApplyDegradesImmediatelyWhileTheFloorIsOutstanding`
- [ ] `withoutAFloorTheTrustedIndexIsTheRatisIndex` - control: no floor, no behaviour change
- [ ] Regression: full `ha-raft` unit suite green - `mvn -pl ha-raft test` -> Tests run: 830, Failures: 0, Errors: 0
- [ ] Downstream compile: `bolt` and `grpcw` (the modules depending on `arcadedb-ha-raft`) test-compile clean

## Notes for review

- Behaviour change worth a deliberate call: while a flagged re-download is outstanding, LINEARIZABLE reads on that node now fail rather than silently serving short data. That is the point of the fix, but it converts a silent correctness bug into visible 503s on a node that is genuinely missing committed entries. The failed-download re-arm and the readiness gate above are there so the node recovers or is taken out of rotation rather than sitting clamped.
- Not addressed here, and flagged in #6111 as needing its own design pass: deferring the PAUSED -> RUNNING lifecycle transition itself. This PR leaves `StateMachineUpdater.reload()`'s expectations untouched on purpose.

---

## Review history

Four automated review cycles ran; every finding was verified against the code before acting, and each round's tests were proved able to fail by neutering the specific change they name.

| cycle | commit | what changed |
|---|---|---|
| 1 | `f27b58083` | Documented why the lenient early-degrade needs no explicit `targetIndex > floor` test. Chased down the write-side question rather than assuming: confirmed `applyTxEntry`'s `WALVersionGapException` -> diverged-database -> resync escalation prevents mismatched pages from landing during the clamped window, and recorded it on `reinitialize()`'s javadoc. |
| 2 | `78624f6b3` | **Real liveness bug found.** Re-arming `needsSnapshotDownload` on a failed download made `isSnapshotDownloadPending()` true, which stands both `isFollowerLaggingBeyond()` and `recoverFromPersistentLag()` down - and the lag backstop could never have covered this case anyway, since a node holding a gap reports *zero lag* (Ratis derives the applied index from the marker that is ahead). Added `HealthMonitor.HealthTarget.retryUnfilledSnapshotGap()`, driven from `tick()`, throttled to one attempt per snapshot-watchdog timeout. |
| 3 | `c8ae1853c` | **Second real bug found.** `notifyLeaderChanged()` submits `triggerSnapshotDownload()` on the node that just *won* the election, and `getLeaderHttpAddress()` then resolves to this node's own address - so a self-download would have reported success and let this PR's new `resolveStaleSnapshotFloorAfterResync()` durably record the marker as applied, reopening #6111 on a leader across restarts. `triggerSnapshotDownload()` now refuses when this node is the leader, plus an address-equality backstop for the TOCTOU window. |
| 4 | `ae095d684` | Logged the case where the local address cannot be resolved (the address backstop would otherwise disarm silently); documented the deliberate single-node/no-peer wedge in `ha-raft/CLAUDE.md`; stated at the field why the floor is node-global rather than per-database. |

Final `ha-raft` unit suite: **839 tests, 0 failures**. Test class grew from 11 to 20 tests across the cycles.

### Deliberately not done

- `notifyInstallSnapshotFromLeader()`'s plain `lastAppliedIndex.set(snapshotIndex)` under the documented CAS-loss race - same shape as the one guarded here, but pre-existing, inside the #5407 durability reasoning, and unrelated to the read-guarantee bug. Deserves its own issue and test.
- Partial-success bookkeeping in `triggerSnapshotDownload()`'s per-database loop - pre-existing, reached identically by paths this PR does not touch.
- Gating the floor resolution on `resynced > 0` - would strand a node with no databases open with a floor nothing can resolve, a worse failure than the one guarded against. The count is logged instead.
- A dedicated test for the `WALVersionGapException` escalation during the clamped window - this PR cites that mechanism, it does not change it.

### DIFF
--- a/ha-raft/CLAUDE.md
+++ b/ha-raft/CLAUDE.md
@@ -16,6 +16,15 @@ The persisted `.raft/applied-index` JSON is read in `reinitialize()` but **never
 
 Note that `globalAppliedIndex` and `lastAppliedIndex` track the same value on the apply path but are seeded independently, so they can briefly differ right after `reinitialize()`. Do not assert equality across that window.
 
+## `RaftHAServer.getLastAppliedIndex()` does not read ArcadeDB's counter
+
+It reads Ratis: `division.getInfo().getLastAppliedIndex()` -> `StateMachineUpdater.appliedIndex`, which Ratis seeds from `getLatestSnapshot().getIndex()` (in its constructor, and again in `reload()` right *after* `reinitialize()` returns). `ArcadeStateMachine.lastAppliedIndex`, the `AtomicLong`, feeds `takeSnapshot()` and the phase-2 replay floor and is **not** on that path.
+
+Two consequences that have already cost debugging time:
+
+- Changing what `reinitialize()` writes into the `AtomicLong` does **not** change what a `waitForAppliedIndex()` / `waitForLocalApply()` waiter observes. If you are chasing "a read was released too early", the value to reason about is the marker index, not the counter.
+- Because the value comes from the marker, a node whose marker runs ahead of the entries it actually applied advertises an applied index covering data it does not hold. That is issue #6111: `reinitialize()`'s snapshot-gap branch now publishes `staleSnapshotAppliedFloor` and the waiters use `getTrustedAppliedIndex()` (the raw value clamped to that floor). **Reporting** paths - cluster status, lag detection - deliberately keep the raw Ratis value; only read guarantees clamp. The floor is cleared where a resync actually restored the state, never where one is merely requested.
+
 ## Snapshots are taken far more often than you would guess
 
 Three triggers, not one: the Ratis auto-trigger (`arcadedb.ha.snapshotThreshold`, default 100k entries), ArcadeDB's own wall-clock `RaftLogCompactionScheduler` (`arcadedb.ha.snapshotInterval` 300 s, `arcadedb.ha.snapshotMinEntries` 64), **and shutdown**, via Ratis's `StateMachineUpdater.stop()`.
--- a/ha-raft/CLAUDE.md
+++ b/ha-raft/CLAUDE.md
@@ -25,6 +25,13 @@ Two consequences that have already cost debugging time:
 - Changing what `reinitialize()` writes into the `AtomicLong` does **not** change what a `waitForAppliedIndex()` / `waitForLocalApply()` waiter observes. If you are chasing "a read was released too early", the value to reason about is the marker index, not the counter.
 - Because the value comes from the marker, a node whose marker runs ahead of the entries it actually applied advertises an applied index covering data it does not hold. That is issue #6111: `reinitialize()`'s snapshot-gap branch now publishes `staleSnapshotAppliedFloor` and the waiters use `getTrustedAppliedIndex()` (the raw value clamped to that floor). **Reporting** paths - cluster status, lag detection - deliberately keep the raw Ratis value; only read guarantees clamp. The floor is cleared where a resync actually restored the state, never where one is merely requested.
 
+### The stale-gap failure mode an operator will meet
+
+A node holding an unfilled gap fails LINEARIZABLE reads (503) and reports not-ready via `isResyncInProgress()`, by design: it is genuinely missing committed entries, and the alternative is serving short data silently. Two things about how it recovers:
+
+- **The lag backstop cannot see it.** `isFollowerLaggingBeyond()` is driven by `commitIndex - appliedIndex`, and the applied index comes from the marker that is ahead - so the node reports *zero lag* while the gap is open. `retryUnfilledSnapshotGap()` (a `HealthMonitor.HealthTarget` hook, throttled to one attempt per snapshot-watchdog timeout) exists specifically because of this; do not try to fold it back into `recoverFromPersistentLag()`.
+- **A node that becomes leader while holding a gap wedges, deliberately.** `triggerSnapshotDownload()` refuses to resync when this node is the leader (or when the resolved leader address is its own), because copying its own incomplete databases onto themselves would report success and durably record the marker index as applied - re-opening #6111 permanently. So a single-node cluster, or a node that wins the election because every peer is unreachable, retries and refuses on every tick until a peer that actually holds the entries takes leadership. That is the intended terminal state, not a hang; the refusal is logged at WARNING on every attempt.
+
 ## Snapshots are taken far more often than you would guess
 
 Three triggers, not one: the Ratis auto-trigger (`arcadedb.ha.snapshotThreshold`, default 100k entries), ArcadeDB's own wall-clock `RaftLogCompactionScheduler` (`arcadedb.ha.snapshotInterval` 300 s, `arcadedb.ha.snapshotMinEntries` 64), **and shutdown**, via Ratis's `StateMachineUpdater.stop()`.
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -212,6 +212,10 @@ public class ArcadeStateMachine extends BaseStateMachine {
   // flagged re-download actually lands; it is cleared only by a resync that restored the state, never
   // by merely starting one.
   private final AtomicLong    staleSnapshotAppliedFloor  = new AtomicLong(-1);
+  // Wall-clock of the last retry submitted by retryUnfilledSnapshotGap(); 0 = none since the floor was
+  // last cleared. Throttles the HealthMonitor-driven backstop, which ticks far more often than a full
+  // multi-database resync costs.
+  private final AtomicLong    lastStaleSnapshotRetryMs   = new AtomicLong();
   // Set to true after applyTransaction hits a genuinely unrecoverable, node-wide condition: a JVM
   // Error (OOM, StackOverflow - the JVM itself is unstable), an unknown committed entry type (#4798,
   // rolling-upgrade safety), or an unexpected error on an entry with no single target database
@@ -2570,9 +2574,12 @@ public class ArcadeStateMachine extends BaseStateMachine {
       resolveStaleSnapshotFloorAfterResync();
     } catch (final Exception e) {
       LogManager.instance().log(this, Level.SEVERE, "Snapshot resync failed", e);
-      // The node is still short of the entries the marker claims, so the read floor stays. Re-arm the
-      // request instead of leaving it cleared: the next leader change (or the HealthMonitor backstop)
-      // retries, rather than clamping reads until a restart (issue #6111).
+      // The node is still short of the entries the marker claims, so the read floor stays and the
+      // request is re-armed rather than left cleared: a later leader change picks it up. The retry that
+      // does NOT depend on an election is retryUnfilledSnapshotGap(), driven by the HealthMonitor tick -
+      // note that recoverFromPersistentLag() cannot serve here, because the re-armed flag makes
+      // isSnapshotDownloadPending() true and both it and isFollowerLaggingBeyond() then stand down
+      // (issue #6111).
       if (staleSnapshotAppliedFloor.get() >= 0)
         needsSnapshotDownload.set(true);
     } finally {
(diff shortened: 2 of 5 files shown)


robfrank:
Review cycle 1 applied in `f27b58083`.

**Applied**
- Added the note explaining why the lenient early-degrade needs no explicit `targetIndex > floor` test: reaching that line already proves it, because `getTrustedAppliedIndex()` is pinned at the floor for the whole life of an outstanding gap. Comment is in both waiters so a future reader cannot "fix" it into a subtle bug.
- Chased down the write-side question rather than leaving it assumed. Confirmed: while the floor is outstanding Ratis does keep feeding `applyTransaction()` entries committed after the marker, and applying one onto a database that stops at `persistedApplied` fails its page-version check. `applyTxEntry` catches the `WALVersionGapException`, marks the database diverged and submits an immediate `triggerSnapshotDownload()` instead of writing mismatched pages (`ArcadeStateMachine.java`, the `catch (WALVersionGapException)` block). So the gap cannot escalate from "stale data served" to "corrupted data written". Recorded as a paragraph on `reinitialize()`'s javadoc so the next reader does not have to re-derive it.

**Not applied, with rationale**
- *Partial-success bookkeeping in `triggerSnapshotDownload()`'s per-database loop.* Agreed it is a real gap, but as noted it is pre-existing and independent of this fix: it is reached identically by the WAL-gap path and the HealthMonitor backstop, neither of which this PR touches. Folding it in would widen the blast radius of a correctness fix, so I left it alone rather than half-fixing it here.
- *`@Tag("slow")` on the new test class.* Checked the timing assertions rather than tagging reflexively. The class runs in ~1.5 s locally, and the two real-time assertions are one-sided in the safe direction: the scheduled clear fires at +300 ms and the `>= 250 ms` lower bound only tightens if the scheduler fires *early*, which it cannot; the `< 5000 ms` and `< 2000 ms` upper bounds sit against a 10 s quorum timeout with a 1 s bounded re-check, so even a fully missed notification lands around 1.3 s. A loaded runner pushes elapsed up, which is exactly the direction all three bounds tolerate. Tagging it `slow` would move it out of the lane where it actually guards this code.

robfrank:
Review cycle 2 applied in `78624f6b3`. The liveness finding was correct and I verified the whole call graph before acting on it.

**Confirmed the finding**
`isFollowerLaggingBeyond()` short-circuits on `sm.isSnapshotDownloadPending()`, and `recoverFromPersistentLag()` returns on the same predicate, so re-arming `needsSnapshotDownload` on failure did exactly what you describe: it made the node look 'already handled' forever. Worse than the report, actually - the lag backstop could not have covered this case even with that gate opened. `isFollowerLaggingBeyond()` is driven by `commitIndex - appliedIndex`, and `getLastAppliedIndex()` is derived from the very marker that is ahead, so a node sitting on an unfilled gap reports **zero lag** and never starts a streak at all. Piggybacking on that path was never going to work.

**Fix**
Gave the condition its own hook instead of bending the lag path around it:
- `HealthMonitor.HealthTarget.retryUnfilledSnapshotGap()`, called from `tick()` next to `reportResyncProgress()` - i.e. on the ordinary periodic path, not behind the lag streak logic.
- `ArcadeStateMachine.retryUnfilledSnapshotGap()` re-submits `triggerSnapshotDownload()` when a floor is outstanding, no download is running, a leader is known, and this node is not the leader. Self-throttled to one attempt per `computeSnapshotWatchdogTimeoutMs()` (~30 s), because a full resync pulls every database and the monitor ticks far more often than that costs. The stamp resets when the floor clears.
- This also closes the second minor point: `triggerSnapshotDownload()`'s early `leaderHttpAddr == null` return no longer needs its own re-arm, since the periodic hook picks it up once a leader is known.
- Corrected the comment that claimed a backstop existed. It now names `retryUnfilledSnapshotGap()` as the election-independent retry and states explicitly why `recoverFromPersistentLag()` cannot serve.

**Tests** - 6 more (17 total in the class), all proved able to fail: removing the `tick()` call and short-circuiting the retry body fails exactly the two that name those behaviours. Coverage: the tick drives the hook; no-op without a floor; no-op while a download is running; no-op with no leader known; fires and actually reaches `triggerSnapshotDownload()` (asserted by the floor being resolved, not just by a flag); throttled within the interval. Full `ha-raft` suite: 836 tests, 0 failures.

**Still not applied**
`notifyInstallSnapshotFromLeader()`'s plain `lastAppliedIndex.set(snapshotIndex)` under the documented CAS-loss race. Agreed it is the same shape as the one I guarded, and `accumulateAndGet(..., Math::max)` would be a one-line change - but it sits inside a path carrying the #5407 durability reasoning, it is pre-existing and unrelated to the read-guarantee bug this PR fixes, and it deserves its own issue with its own test rather than riding along here.

robfrank:
Review cycle 3 applied in `c8ae1853c`. The self-download scenario is real and I confirmed every link in the chain before fixing it.

**Verified**
`notifyLeaderChanged()` submits `triggerSnapshotDownload()` outside the leader/replica branch, so it runs on the node that just won the election too; `getLeaderHttpAddress()` is `resolveHttpAddress(getLeaderId())`, which after winning resolves to this node's own address. And you are right that this PR is what made it consequential: before it, a self-download was merely pointless, whereas `resolveStaleSnapshotFloorAfterResync()` would now drop the floor *and* durably persist the marker index, so even a restart would no longer re-detect the gap.

**Fix** - both suggestions, since they cover different windows:
- `triggerSnapshotDownload()` refuses outright when `raftHAServer.isLeader()`, matching `retryUnfilledSnapshotGap()`. Placed before the single-flight CAS so a refusal does not hold the flag. Guarding the method rather than the `notifyLeaderChanged` call site covers the WAL-gap and watchdog submitters too, and refusing there is right for them as well: a node missing committed entries cannot repair itself from itself under any trigger.
- After resolving the leader address it also refuses when that address equals `getPeerHttpAddress(getLocalPeerId())`. That covers the TOCTOU you implied: leadership can move between the role check and the address resolution, and `getLeaderId()` can report this node while `isLeader()` has not caught up.

On refusal the floor stands, `isResyncInProgress()` keeps the node out of the ready set, and reads keep failing honestly until leadership moves to a peer that actually holds the entries. That is the correct terminal state for a leader that is missing committed data, and it is strictly better than the silent false-catch-up.

**Tests** - 3 more (20 total), proved able to fail: neutering both guards fails exactly the two that name them.
- `aLeaderMustNotResolveItsOwnFloorByDownloadingFromItself` - floor stands, marker index not persisted, node stays not-ready
- `aResolvedLeaderAddressEqualToOurOwnIsAlsoRefused` - the TOCTOU window, with `isLeader()` deliberately stale
- `aPeerLeaderIsStillDownloadedFromAndResolvesTheFloor` - the control, without which the two above would also pass against a `triggerSnapshotDownload()` that refused everything

Full `ha-raft` suite: 839 tests, 0 failures.

**Other two observations**
- Throttle CAS: agreed, and thanks for checking it - the loser returning without submitting is intentional, since the winner's submission is the attempt for that window.
- Zero-database resync: I did not gate on `resynced > 0`. Gating there would strand a node that genuinely has no databases open with a floor that nothing can ever resolve, keeping it not-ready forever - a worse failure than the one being guarded against, and the count is over the same `getDatabaseNames()` set `notifyInstallSnapshotFromLeader()` already uses. Instead `resolveStaleSnapshotFloorAfterResync()` now logs the count as it resolves, so "floor resolved after reinstalling 0 databases" is visible in the log if an unexpectedly-empty registry ever does happen.
- Also added the pointer you asked for: the `targetIndex > floor` comment now names `reinitialize()`'s snapshot-gap branch as where the invariant is established, and says that changing what that branch publishes must revisit the reasoning.

robfrank:
Review cycle 4 applied in `ae095d684`. All three observations were non-blocking; two were cheap enough to just do, and the third I answered in code comments rather than changing behaviour.

- **Null local address silently disarming the backstop.** Real, and it is exactly the kind of guard that must not fail quietly. `triggerSnapshotDownload()` now logs a WARNING when `getPeerHttpAddress(getLocalPeerId())` returns null, naming the consequence: the address check is inactive for that attempt and only the leader-role check is standing.
- **The single-node / no-reachable-peer wedge.** Documented in `ha-raft/CLAUDE.md` under a new "the stale-gap failure mode an operator will meet" heading, since you are right that this is a new shape of stuck that on-call has to be able to recognise. It records both halves: that a node holding a gap reports *zero lag* (so the lag backstop is structurally blind and `retryUnfilledSnapshotGap()` exists for that reason - do not fold it back), and that a leader holding a gap refuses forever by design, with the refusal logged at WARNING on every attempt so it is visible rather than silent.
- **Node-global rather than per-database floor.** Intended, and now stated at the field with the reasoning: the gap is detected from the global persisted position against the inherently global Ratis snapshot index, so which co-located database is actually short is not knowable at that point. Deriving a per-database answer from a global signal is precisely the mistake #4824 fixed, so the conservative global clamp is the deliberate choice - at the cost you name, a multi-database node clamping reads on all of them while one gap is open.

On the test gap you flagged: agreed, and I left it alone on purpose. The `WALVersionGapException` -> diverged-database -> resync escalation is pre-existing behaviour this PR only *cites* in the `reinitialize()` javadoc, and it already has its own coverage elsewhere in the module. Adding a test here would be testing someone else's mechanism through this PR's lens; if it deserves a dedicated test for this specific window, that belongs in its own change.

Full `ha-raft` suite after this cycle: 839 tests, 0 failures.

This is cycle 4 of 4, so the automated loop stops here. Merge is yours.

Issue #4937, opened 2026-07-03
[tx] Files can join a transaction's page set after the commit lock set is computed


lvca:
**Severity:** HIGH (enabler for aborted-tx WAL replay)

**Location:** `TransactionContext.java:644-674` (lock set computed at `:652`, `updateRecordNoLock` at `:665`), `serializer/BinarySerializer.java:975,1166` (EXTERNAL property values written to the paired external bucket during phase-1 serialization), `index/lsm/LSMTreeIndex.java:515` (during phase 1 `status==COMMIT_1ST_PHASE`, so `indexer.updateDocument` writes **directly** into index pages, incl. indexes created mid-tx that were absent from `indexChanges` at lock time), `TransactionIndexContext.java:248-282` (`addFilesToLock` omits `LSMVectorIndex`s companion graph file).

Pages of these late-joining files pass phase-1 `checkPageVersion` **without their file lock held**, making the aborted-tx WAL-replay interleaving reachable and producing surprise phase-2 CMEs. Reachable with (a) EXTERNAL properties updated via the deferred path in two concurrent txns, (b) vector-index graph-file writes.

Also note deferred updates apply index mutations **twice** (directly at `:665-673`, then via `putReplay` at `:689`) - benign for lookups but doubles LSM garbage/compaction.

**Fix:** after `updateRecordNoLock` + `indexChanges.commit()`, assert every file in `modifiedPages`/`newPages` is locked; if not, release, re-lock the union, and re-validate (bounded retry). Add the vector graph file id in `addFilesToLock`.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5053 (merged), two layers:

1. Multi-file indexes now contribute ALL their component files to the commit lock set: addFilesToLock consumes getFileIds() (plural) and LSMVectorIndex.getFileIds() includes the companion graph file (it returned only the mutable file). This also fixed the latent add(-1) for composite TypeIndex ids.
[...]

PR #6161, opened 2026-08-13
fix(#6105): LSM_VECTOR writes right after COMPACT INDEX are silently not indexed

### LINKED ISSUE #6105: LSM_VECTOR: writes right after COMPACT INDEX (same session) are silently not indexed
## Summary

A write (`INSERT` of a new record, or `UPDATE` of an existing one) issued in the same database session immediately after `COMPACT INDEX` on an `LSM_VECTOR` index is silently **not** reflected in the vector index: the document itself is created/updated correctly (the SQL command reports success, `UPDATE ... SET` reports `count: 1`), but `LSMVectorIndex`'s live vector count (`countEntries()` / `VectorLocationIndex.size()` / `getMaxVectorId()`) does not change at all.

This reproduces on current `main` (no relationship to any in-flight PR) with a plain standalone, single-threaded, non-HA database - no crash, no concurrency involved.

## Reproduction

```java
database.transaction(() -> {
  database.command("sql", "CREATE VERTEX TYPE Doc");
  database.command("sql", "CREATE PROPERTY Doc.id STRING");
  database.command("sql", "CREATE PROPERTY Doc.embedding ARRAY_OF_FLOATS");
  database.command("sql", "CREATE INDEX ON Doc (id) UNIQUE");
  database.command("sql", """
      CREATE INDEX ON Doc (embedding) LSM_VECTOR
      METADATA {"dimensions": 16, "similarity": "COSINE", "quantization": "INT8",
                "mutationsBeforeRebuild": 100000000, "inactivityRebuildTimeoutMs": 0}""");
});

database.transaction(() -> {
  for (int i = 0; i < 300; i++)
    database.command("sql", "INSERT INTO Doc SET id = ?, embedding = ?", "doc" + i, someEmbedding(i));
});

database.command("sql", "COMPACT INDEX `Doc[embedding]`"); // succeeds, {"compacted": true}

LSMVectorIndex index = (LSMVectorIndex) ((TypeIndex) database.getSchema()
    .getIndexByName("Doc[embedding]")).getIndexesOnBuckets()[0];
System.out.println(index.countEntries()); // 300, correct

database.transaction(() ->
    database.command("sql", "UPDATE Doc SET embedding = ? WHERE id = ?", someOtherEmbedding(), "doc0"));
// UPDATE reports {"count": 1} - the document's "embedding" property really changed.

System.out.println(index.countEntries()); // STILL 300 - the update never reached the vector index
```

The same happens for a brand new `INSERT` right after the `COMPACT INDEX` (not just an `UPDATE` of an existing record).

Confirmed the `LSMVectorIndex` Java object instance is identical (`System.identityHashCode`) before and after the compaction and the write - it is not a stale/detached object reference on the caller's side. The document write itself succeeds and is visible to `SELECT`. It is specifically the index-maintenance step that appears to be skipped for the first write(s) after a compaction, in the same session.

## Suspected area

`LSMVectorIndex.rewriteDataFileWithLiveEntries` (backing `COMPACT INDEX`) swaps in a new data file and calls `((LocalSchema) database.getSchema()).setMigratedFileId(oldFileId, newFileId)`, then drops the old file's registration. The most likely explanation is that something in the write path that decides "which indexes apply to a changed property on this bucket" is still keyed by the pre-compaction file id (or a cached `TypeIndex`/`LocalBucket` structure not refreshed by the migration), so it silently resolves to nothing for the very next write(s) until something else (a full graph rebuild, or a database reopen, which reads pages fresh) re-syncs it. I have not traced this all the way to a single line - filing so it is tracked, since it looks like a real (and potentially serious) correctness gap independent of any specific PR.

## Impact

Any operator workflow that runs `COMPACT INDEX` on a live `LSM_VECTOR` index and then keeps writing to the same type in the same session risks silently losing vector-index coverage for those writes until the next rebuild or reopen - the affected records remain in the type/documents but drop out of vector search results.

## Found while

Investigating a Claude code-review comment on PR #6102 (issue #5870, vector id renumbering during compaction), which asked whether a write immediately following a renumbering compaction gets a non-colliding id. That specific concern (`nextId` reset correctness) is unaffected by this bug and is covered by its own tests, but writing a test for "insert right after compaction" surfaced this separate, pre-existing issue - confirmed independent of #5870's changes by reverting them and reproducing the same symptom on unmodified `main`.

robfrank:
Closes #6105

## Root cause

An `LSMVectorIndex` is named after the component file it holds, and `rewriteDataFileWithLiveEntries` (what `COMPACT INDEX` runs) swaps that file in, so the compaction reassigns `indexName` to the new component's name. That rename is deliberate - every node names the index after the file it holds, so a leader has to follow its own rename or its schema stops matching the followers that rebuilt the index from the file it shipped them. Every other index type keeps its creation name for life.

`LocalSchema.indexMap` is keyed by the name the index was registered under, and nothing re-keyed it. After a compaction the index was therefore reachable in the schema only under a name it no longer answered to.

That turns into silent data loss two lines into the commit path. Index maintenance is queued on `TransactionIndexContext` under `index.getName()`, and `TransactionIndexContext.commit()` opens with:

```java
indexEntries.keySet().removeIf(indexName -> !database.getSchema().existsIndex(indexName));
```

a filter that exists to drop the lanes of indexes dropped mid-transaction (TYPE DROP). A freshly compacted vector index matched it exactly, so its queued entries were discarded without a word. The record itself was written normally - which is why `INSERT`/`UPDATE` reported success while `countEntries()` never moved - and a database reopen rebuilds `indexMap` from the persisted (new) names, which is what made writes start landing again and made the whole thing look like a phantom.

There is a second reachable shape of the same defect, and it needs no explicit `COMPACT INDEX` at all: `LSMVectorIndex.onAfterCommit` schedules a compaction automatically once the garbage ratio warrants it, and it runs on the async executor. So a compaction can rename the index *between* an entry being queued on an open transaction and that transaction committing, on a thread that never asked for a compaction. Re-keying the registry alone does not save that transaction - its lane is keyed by the pre-compaction name, which is now the one that no longer exists.

## The fix

Two changes, both load-bearing (each was verified to be necessary by disabling it and watching the matching test go red):

1. **`LocalSchema.indexRenamed(oldName, index)`**, called from `rewriteDataFileWithLiveEntries` inside the same write-locked critical section that swaps the file and re-publishes the location index. It re-keys `indexMap`, publishing the new name *before* retiring the old one so a concurrent lookup never sees neither, and removing the old key only when it still maps to this index.

2. **`TransactionIndexContext` resolves a lane back to the `IndexInternal` it captured when the lane was opened**, instead of re-resolving it by name at commit time. The "was this index dropped?" test remains a schema lookup, but of the index's *current* name - so a renamed index is kept and a genuinely dropped one is still discarded, preserving the TYPE DROP behaviour the filter was written for. Lanes restored wholesale by `setKeys` carry no reference and fall back to the previous name lookup.

The second change also fixes a quieter consequence: `addFilesToLock` used the same name test, so a renamed index's component files were being left out of the commit lock set.

## Test plan

New `engine/src/test/java/com/arcadedb/index/vector/Issue6105WriteAfterCompactionTest.java` (4 tests, all four fail on `main`):

- [ ] `aCompactedIndexIsStillReachableUnderTheNameItAnswersTo` - the structural invariant: after a compaction the schema resolves the index under its current name, to that very instance, and the retired name is gone.
- [ ] `anInsertRightAfterACompactionIsIndexed` - the reported symptom: `countEntries()` moves, the new document has exactly one vector id, and it is found by vector search.
- [ ] `anUpdateRightAfterACompactionIsIndexed` - the update variant: the live count is unchanged (an update replaces a vector), doc0 holds a *new* vector id, and it is found at its new embedding rather than only at the pre-compaction one.
- [ ] `aWriteQueuedBeforeACompactionOnAnotherThreadIsStillIndexed` - the automatic-compaction shape: an entry queued on an open transaction survives a compaction that renames the index from another thread.

Verification run (all with `-Dmaven.repo.local` isolated):

- [ ] The four new tests fail on unmodified `main`, and specifically the fourth one still fails with change (1) applied and change (2) reverted.
- [ ] `com.arcadedb.index.vector.*Test` - the whole vector package, green.
- [ ] `com.arcadedb.index.**`, `com.arcadedb.schema.**`, `com.arcadedb.database.**`, `TransactionTest`, `LockFilesInOrderTest`: **1843 tests, 0 failures, 0 errors, 1 skipped.**
- [ ] Full reactor `test-compile` green.

`Issue5870VectorIndexIdRenumberingTest` carried a Javadoc paragraph documenting this bug as an unfixed reason for its reopen-before-insert; only that comment was updated to point at the fix and the new test. No assertion in it was touched.

## Notes for review

- `indexRenamed` is the only rename hook in the schema, and `LSMVectorIndex` is the only caller, because it is the only index that renames itself. If another index type ever gains a rename, it must call this too.
- `indexRenamed` mutates `indexMap` from whichever thread ran the compaction, which for an *automatic* compaction is the async executor rather than a user thread. `indexMap` is an unsynchronized `HashMap`, but it has always been mutated off arbitrary threads - `CREATE INDEX` / `DROP INDEX` do exactly that while queries read it - so this is not a new kind of hazard, only a more frequent one. Hardening the map (a `ConcurrentHashMap`, which would also change `getIndexByName(null)` from a `SchemaException` to an NPE) felt like a separate call to make; happy to fold it in if you'd rather have it here.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/TransactionIndexContext.java
+++ b/engine/src/main/java/com/arcadedb/database/TransactionIndexContext.java
@@ -218,6 +218,12 @@ public class TransactionIndexContext {
   /**
    * The index a lane belongs to: the reference captured when the lane was opened, falling back to a lookup by name
    * for lanes restored wholesale by {@link #setKeys}, which carry no reference. See {@link #indexPerLane}.
+   * <p>
+   * That fallback is not a residual hole. {@code setKeys} is only used by {@code TransactionContext.commitFromReplica},
+   * and a replica never renames a vector index behind its own back: {@code LSMVectorIndex.isCompactionAllowedOnThisNode}
+   * refuses to schedule a compaction on anything but a standalone database or the current leader, so a follower's copy
+   * is renamed only when it adopts the component the leader shipped it - which arrives through the schema update, not
+   * concurrently with a replicated commit it is already applying.
    */
   private IndexInternal resolveIndex(final String laneName) {
     final IndexInternal index = indexPerLane.get(laneName);
@@ -433,8 +439,6 @@ public class TransactionIndexContext {
       return;
 
     final String indexName = index.getName();
-    // The lane belongs to THIS index, whatever it ends up being called by commit time (issue #6105).
-    indexPerLane.put(indexName, index);
 
     if (!index.isTransactionKeyOrderRequired()) {
       // APPEND-ONLY LANE: NO KEY ORDERING, NO PER-KEY DEDUP. REPLAY ORDER == INSERTION ORDER.
@@ -447,6 +451,9 @@ public class TransactionIndexContext {
               + "' cannot opt out of the key-ordered transaction map: duplicated-key detection reads it back");
         lane = new ArrayList<>();
         unorderedEntries.put(indexName, lane);
+        // Recorded once per lane, at creation, not once per key: the lane belongs to THIS index whatever it ends up
+        // being called by commit time (issue #6105).
+        indexPerLane.put(indexName, index);
       }
       lane.add(new IndexKey(false, operation, keysValues, rid));
       return;
@@ -461,6 +468,8 @@ public class TransactionIndexContext {
     if (keys == null) {
       keys = new TreeMap<>(); // ORDERED TO KEEP INSERTION ORDER
       indexEntries.put(indexName, keys);
+      // See the sibling call in the append-only branch above (issue #6105).
+      indexPerLane.put(indexName, index);
 
       values = new HashMap<>();
       keys.put(k, values);
--- a/engine/src/main/java/com/arcadedb/database/TransactionIndexContext.java
+++ b/engine/src/main/java/com/arcadedb/database/TransactionIndexContext.java
@@ -59,9 +59,10 @@ public class TransactionIndexContext {
    * would discard it under the rule that drops the lanes of indexes dropped mid-transaction (TYPE DROP): the record
    * is written, the index entry is silently lost.
    * <p>
-   * Holding the reference makes the lane's identity independent of the name. The "was it dropped?" test stays a
-   * schema lookup, but of the index's CURRENT name, so a renamed index is kept and a genuinely dropped one is still
-   * discarded.
+   * Holding the reference lets the lane ask the index what it is called now instead of assuming it is still called
+   * what it was called then. Everything else is unchanged: the current name goes to the same schema lookup as
+   * before, so a renamed index is kept, a genuinely dropped one is still discarded, and the schema stays the sole
+   * authority on which object that name resolves to (see {@link #laneIndexName} for why that distinction matters).
    */
   private final Map<String, IndexInternal>                                   indexPerLane     = new HashMap<>();
 
@@ -216,18 +217,36 @@ public class TransactionIndexContext {
   }
 
   /**
-   * The index a lane belongs to: the reference captured when the lane was opened, falling back to a lookup by name
-   * for lanes restored wholesale by {@link #setKeys}, which carry no reference. See {@link #indexPerLane}.
+   * The name a lane's index answers to NOW: the current name of the reference captured when the lane was opened,
+   * falling back to the lane's own key for lanes restored wholesale by {@link #setKeys}, which carry no reference.
+   * See {@link #indexPerLane}.
    * <p>
-   * That fallback is not a residual hole. {@code setKeys} is only used by {@code TransactionContext.commitFromReplica},
-   * and a replica never renames a vector index behind its own back: {@code LSMVectorIndex.isCompactionAllowedOnThisNode}
-   * refuses to schedule a compaction on anything but a standalone database or the current leader, so a follower's copy
-   * is renamed only when it adopts the component the leader shipped it - which arrives through the schema update, not
-   * concurrently with a replicated commit it is already applying.
+   * The captured reference is used to find the NAME, never as the index to operate on. Those are not always the same
+   * object: a wrapper index queues its entries under the index it wraps - {@code LSMTreeFullTextIndex} tokenizes text
+   * and calls through to its {@code LSMTreeIndex}, which is what reaches {@code addIndexKeyLock} - while the schema
+   * registers the WRAPPER under that same name. Replaying through the captured inner index instead of the registered
+   * wrapper would quietly change which implementation of {@code putReplay}/{@code removeReplay},
+   * {@code getAssociatedBucketId} and {@code getFileIds} the commit uses. So this fix moves only the one thing that
+   * was wrong - the name being looked up - and leaves the schema the sole authority on what that name resolves to.
+   * <p>
+   * That also means an index dropped and re-created under the same name inside one transaction resolves to the new
+   * one, exactly as it did before this fix: no behaviour of that (already ill-defined) sequence is changed here.
+   * <p>
+   * The {@code setKeys} fallback is not a residual hole either. {@code setKeys} is only used by
+   * {@code TransactionContext.commitFromReplica}, and a replica never renames a vector index behind its own back:
+   * {@code LSMVectorIndex.isCompactionAllowedOnThisNode} refuses to schedule a compaction on anything but a standalone
+   * database or the current leader, and an explicit {@code COMPACT INDEX} is a DDL statement a follower forwards to
+   * the leader. A follower's copy is renamed only when it adopts the component the leader shipped it, which arrives
+   * through the schema update rather than concurrently with a replicated commit it is already applying.
    */
-  private IndexInternal resolveIndex(final String laneName) {
+  private String laneIndexName(final String laneName) {
     final IndexInternal index = indexPerLane.get(laneName);
-    return index != null ? index : (IndexInternal) database.getSchema().getIndexByName(laneName);
+    return index != null ? index.getName() : laneName;
+  }
+
+  /** The index a lane belongs to, as the schema currently resolves it. See {@link #laneIndexName}. */
+  private IndexInternal resolveIndex(final String laneName) {
+    return (IndexInternal) database.getSchema().getIndexByName(laneIndexName(laneName));
   }
 
   /**
@@ -236,8 +255,7 @@ public class TransactionIndexContext {
    * (TYPE DROP) is still reported as gone.
    */
   private boolean laneIndexStillExists(final String laneName) {
-    final IndexInternal index = indexPerLane.get(laneName);
-    return database.getSchema().existsIndex(index != null ? index.getName() : laneName);
+    return database.getSchema().existsIndex(laneIndexName(laneName));
   }
 
   public int getTotalEntries() {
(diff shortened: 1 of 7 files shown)


robfrank:
Thanks - all three points addressed in `0831906`.

**1. The `setKeys` / `commitFromReplica` gap: confirmed not reachable, and now documented in code.**

`LSMVectorIndex.isCompactionAllowedOnThisNode` is `!database.isReplicated() || database.isLeader()`, and `isCompactionDue()` consults it first, so a follower never schedules a compaction of its own - the premise the scenario needs ("followers run their own `onAfterCommit`-triggered background compactions independently of the leader's schedule") does not hold. A follower's copy is renamed only when it adopts the component the leader shipped it, which arrives through the schema update rather than concurrently with a replicated commit it is already applying. So `addFilesToLock`'s lock-set completeness is not exposed on the replica side either. I put that reasoning on `resolveIndex`'s Javadoc so the fallback reads as a deliberate "cannot happen here" rather than an implicit gap.

**2. `indexMap` is now a `ConcurrentHashMap`.**

You are right that this fix changes what the map is for: `commit()` now depends on seeing the renamed key correctly, not just eventually and cosmetically, and reconstructing "which lock happens to publish this" is exactly the thing a future locking refactor would silently break. Taking the offer rather than the comment. The one behavioural edge is null keys - `ConcurrentHashMap` raises an NPE where the `HashMap` answered "absent" - so `existsIndex`, `getIndexByName` and `dropIndexInternal` null-guard the lookup and keep the old answer for a null name, with the reason written next to each.

**3. `indexPerLane.put` hoisted.**

Moved into the two lane-creation branches (`unorderedEntries.put` and `indexEntries.put`), so it runs once per lane per transaction rather than once per queued key.

Re-verified after the changes: `com.arcadedb.index.**`, `com.arcadedb.schema.**`, `com.arcadedb.database.**`, `TransactionTest`, `LockFilesInOrderTest` - **1843 tests, 0 failures, 0 errors, 1 skipped**, with the 4 new `Issue6105WriteAfterCompactionTest` tests and `Issue5870VectorIndexIdRenumberingTest` green inside that run.

robfrank:
Addressed in `8c0ce07`.

You are right that the field comment overclaimed. `ManualIndexBuilder` is the one route by which a *caller-supplied* name reaches `indexMap` unmediated (`LocalDocumentType` only ever passes the `TypeIndex`'s own name, which is constructed non-null), and it had no name validation at all - a null there used to be accepted all the way into the map, producing an index nothing could look up again. `create()` now rejects it up front with a `DatabaseMetadataException`, which is the improvement you flagged rather than the inconsistency; the field comment now names the direct consumers and why each cannot pass null, instead of asserting a blanket invariant.

On the `isCompactionAllowedOnThisNode` point: agreed, and I've left it as documentation rather than an assertion on purpose - the fallback is correct on its own terms (it resolves the lane by name, which is what the code did before this PR), so the invariant buys the replica path lock-set completeness rather than correctness of the entry itself. If it ever stops holding, the symptom is the pre-#4937 one, not silent loss.

Re-verified: `com.arcadedb.index.**`, `com.arcadedb.schema.**`, `com.arcadedb.database.**`, `TransactionTest`, `LockFilesInOrderTest` - **1843 tests, 0 failures, 0 errors, 1 skipped**, including both manual-index suites (`Issue5765ManualIndexIfNotExistsTest`, `Issue5780ManualIndexIsAutomaticTest`) and the 4 new `Issue6105WriteAfterCompactionTest` tests.

robfrank:
Addressed in `dd3b00b` - and your edge case turned out to be pointing at a bigger one right next to it.

I first implemented exactly what you described: drop the lane unless the schema still resolves the name **to this very index**. Before shipping it I armed it as a hard failure and ran the suite, and it fired: `DataEncryptionTest.fullTextIndexLookupWorksWithEncryption`, with `lane=LSMTreeIndex@1523034648  name=Entry_0_594972426940250  schema=LSMTreeFullTextIndex@1289563251`.

A wrapper index queues its entries under the index it *wraps*: `LSMTreeFullTextIndex.put` tokenizes and calls `underlyingIndex.put(...)`, so the object that reaches `addIndexKeyLock` is the inner `LSMTreeIndex`, while the schema registers the wrapper under that same name. Same shape for `LSMTreeGeoIndex` and `LSMSparseVectorIndex`. So the identity check would have silently dropped every full-text lane - and, more to the point, my previous commit was already operating on the captured inner index rather than the registered wrapper, which quietly changes which implementation of `putReplay`/`removeReplay`, `getAssociatedBucketId` and `getFileIds` the commit uses. (It happens to converge for `putReplay` - the wrapper's forwards to `underlyingIndex.put` and the inner one inherits the default that calls its own `put` - but that is a coincidence of two implementations, not a guarantee.)

So the fix is now strictly narrower: **the captured reference is used only to ask the index what it is called now**, and that name goes through the same `getIndexByName` lookup as before. The schema stays the sole authority on which object a name resolves to, wrappers included. Which also settles your edge case directly: a drop-and-recreate under the same name inside one transaction resolves to the new index, exactly as it did before this PR - no behaviour of that (already ill-defined) sequence is changed here, and the javadoc on the new `laneIndexName` says so.

Re-verified: `com.arcadedb.index.**`, `com.arcadedb.schema.**`, `com.arcadedb.database.**`, `TransactionTest`, `LockFilesInOrderTest` - **1843 tests, 0 failures, 0 errors, 1 skipped**, with `DataEncryptionTest` and the 4 new `Issue6105WriteAfterCompactionTest` tests green.

Thanks - that suggestion is what surfaced the wrapper divergence.

robfrank:
All three addressed in `6f9f2d2`. This is cycle 4 of 4, so handing back to a maintainer from here.

**Rename/re-key adjacency** - moved rather than commented. `indexName = newMutable.getName()` now sits immediately above `indexRenamed()`, with a comment stating they are one step and why: `indexName` is volatile and read without the index write lock (`getName()` is what `TransactionIndexContext` keys a lane by), so anything inserted between them leaves the index answering to a name the schema does not know yet - a milder recurrence of this very bug. Nothing between them read the new value, so the move is behaviour-neutral.

**`getIndexKeys` / `getTotalEntriesByIndex`** - documented. Both now say they look a lane up by its KEY, that this is correct for every caller today (`HashIndex`, `LSMTreeIndex`, `LSMTreeIndexCursor`, each passing its own unchanging name), and that an index type which gains a rename has to reach its lane through `laneIndexName` instead or it reproduces #6105 there.

**Manual-index null-name guard** - new `ManualIndexNameGuardTest`, two tests: the refusal (asserting `DatabaseMetadataException` with the message, plus that the schema is left untouched) and a control that a named manual index still builds and registers. I disarmed the guard and re-ran to confirm the first test fails without it rather than passing for some other reason.

On the `@Tag("slow")` nit - agreed it is on the light side by runtime, but the class starts a thread and waits on a real compaction with a 120s join, so the slow lane is where a bad day belongs.

Final verification: `com.arcadedb.index.**`, `com.arcadedb.schema.**`, `com.arcadedb.database.**`, `TransactionTest`, `LockFilesInOrderTest` - **1845 tests, 0 failures, 0 errors, 1 skipped** (1843 + the 2 new guard tests). Full reactor `test-compile` green.

Issue #5655, opened 2026-07-31
HA: OpenCypherQueryEngine commits on the inner LocalDatabase, bypassing Raft replication (same shape as #5492)


robfrank:
## Summary

`OpenCypherQueryEngine` commits on the inner `LocalDatabase` rather than the Raft wrapper, which is the same defect #5492 fixed for `SQLQueryEngine` in #5652. On an HA leader this applies pages locally and never proposes them to Raft.

Found while reviewing #5652; filed separately rather than folded in, since it needs its own reproduction and test.

## The shape

`OpenCypherQueryEngine` holds `private final DatabaseInternal database` (line 83), assigned in the constructor (line 85). On an HA leader that is the inner `LocalDatabase`: `RaftReplicatedDatabase.command()` delegates to `proxied.command(...)`, and `RaftReplicatedDatabase.getQueryEngine()` (line 1197) delegates to `proxied.getQueryEngine()`, so the engine is built with the inner instance.

The explicit-transaction path then commits on that field directly:

```java
// OpenCypherQueryEngine.java:685-693
database.begin(level);        // or database.begin()
...
case COMMIT:
  if (!database.isTransactionActive())
    throw new CommandExecutionException("No active transaction to COMMIT (issue a START TRANSACTION first)");
  database.commit();          // <-- LocalDatabase.commit() on an HA leader: applies pages, replicates nothing
```

`context.setDatabase(database)` (line 738) passes the same inner instance to statement execution, so any Cypher statement that commits mid-execution has the same exposure.

## Why this matters

`LocalDatabase.commit()` writes pages locally; `RaftReplicatedDatabase.commit()` proposes to Raft and then writes. Committing on the inner instance succeeds and replicates nothing. Followers then trail by exactly those page versions, and the next replicated entry touching one of those pages fails its version check with `WALVersionGapException` - database marked diverged, snapshot resync, and the entry after it breaks the same way.

That is the failure #5492 produced through `TRUNCATE TYPE`: 24002 page-version gaps and 357 resync cycles per follower, with one node ending short by scan and its materialized view empty. Nothing throws at the point of the mistake, which is what makes this class of bug hard to attribute.

## Not yet verified

I have **not** reproduced this. The code shape matches #5492 exactly, but two things need checking before treating it as confirmed:

[...]

PR #6162, opened 2026-08-13
fix(#6094): a bare CALL to a write CypherProcedure is no longer classified read-only

### LINKED ISSUE #6094: Bare CALL to a write CypherProcedure is misclassified as read-only, bypassing Raft replication on HA
## Bug

`SimpleCypherStatement`'s `readOnly` flag never accounts for a top-level `CALL` to a registered write
`CypherProcedure`:

```java
this.readOnly = !hasCreate && !hasMerge && !hasDelete && !hasRemove && !hasForeach
    && (setClause == null || setClause.isEmpty()) && !writeSubquery;
```

`writeSubquery` (`anyWriteSubquery`) only inspects `CALL { ... }` subquery blocks (`ClauseType.SUBQUERY`, added
for #4191) - it never looks at `callClauses` or asks `CypherProcedureRegistry.get(name).isWriteProcedure()`. So
`CALL merge.node(...) YIELD node RETURN node` - a bare top-level call to a write procedure, with no CREATE/
SET/MERGE/DELETE/REMOVE/FOREACH clause anywhere in the statement - is classified `isReadOnly() == true`.

`OpenCypherQueryEngine.executionDatabase()` uses that flag to choose which database instance a plan executes
against:

```java
private DatabaseInternal executionDatabase(final CypherStatement statement) {
  return statement.isReadOnly() ? database : database.getWrappedDatabaseInstance();
}
```

Per that method's own class Javadoc (referencing #5492/#5655) and `ha-raft/CLAUDE.md`, only the wrapped
(Raft-aware) instance may safely `begin()`/`commit()` - the raw instance's `LocalDatabase.commit()` applies
pages locally without proposing them to Raft.

## Consequences

This misclassification was harmless before #6073's fix (PR #6093): a write procedure with no caller-managed
transaction always threw `TransactionException` before anything could commit, on any code path. #6093 is the
first thing that actually calls `begin()`/`commit()` off `context.getDatabase()` for a `CALL`, which makes the
misclassification live:

- **On an HA leader:** `CallStep.executeProcedure`'s new auto-commit (`context.getDatabase().begin()`/
  `commit()`) lands on the raw, non-replicated instance instead of the wrapped one. The commit applies pages
  locally and never proposes them to Raft - followers trail by those page versions and the next replicated
  entry touching one of them fails its version check (the exact failure mode #5492/#5655 already document for
  other write steps).
- **On an HA follower:** worse. `RaftReplicatedDatabase.command(...)` decides whether to forward to the leader
  via `queryEngine.isExecutedByTheLeader() || analyzed.isDDL() || !analyzed.isIdempotent()`.
  `OpenCypherQueryEngine` doesn't override `isExecutedByTheLeader()` (default `false`), and `isIdempotent()` is
  just `statement.isReadOnly()`. So the condition evaluates `false || false || false`, and the follower
  executes `CALL merge.node(...)` **entirely locally** - real cluster divergence, not just a missed
  replication step.
- `Database.query()` (reserved for idempotent calls, gated in `OpenCypherQueryEngine.query()`) silently
  accepts and executes `CALL merge.node(...)`, since the idempotency check is the same misclassified flag.

Affects every registered write `CypherProcedure`: `merge.node`, `merge.relationship`,
`apoc.refactor.mergeNodes`, `apoc.refactor.cloneNodesWithRelationships`, `apoc.do.when`.

## Reproduction

```java
final ResultSet rs = database.command("opencypher", "CALL merge.node(['Person'], {name:'X'}, {}) YIELD node RETURN node");
```

Before #6093: throws `TransactionException`. After #6093, on a plain local (non-HA) database: works correctly
(the wrapper is the instance itself off HA, so this is a no-op there per `executionDatabase()`'s own doc).
On an HA leader or follower: the write silently bypasses Raft.

## Suggested fix (and why it's not a one-line change)

Add a `SimpleCypherStatement.anyWriteProcedureCall(callClauses)` check (mirroring `anyWriteSubquery`) that
resolves each `CallClause`'s procedure name via `CypherProcedureRegistry.get(name)` and returns true if any
resolves to `isWriteProcedure() == true`, folding it into the `readOnly` computation the same way
`writeSubquery` already is. `CypherProcedureRegistry.get()` is a simple static lookup (case-insensitive, strips
the `apoc.` prefix), safe to call at parse time; `CallClause.getProcedureName()` is already resolved
statically. This also transitively fixes the same gap for a write-procedure `CALL` nested inside a
`CALL { ... }` subquery, since `anyWriteSubquery` already recurses through `inner.isReadOnly()`.

**This is not safe to land as a pure mechanical fix**, because `apoc.do.when` (`DoWhen.isWriteProcedure()`)
returns `true` unconditionally - it can't know at registration time whether its caller-supplied
`ifQuery`/`elseQuery` will actually write. Folding the fix in as designed reclassifies *every*
`CALL apoc.do.when(...)`, even a call whose branch is read-only, as non-idempotent - which breaks the
`Database.query()` idempotency gate for exactly the pattern several pre-existing `DoWhenTest` tests use
(`database.query("opencypher", "CALL apoc.do.when(true, 'RETURN 1 AS x', ...) ...")`), because those tests were
unknowingly relying on today's misclassification bug for `.query()` to succeed. 10 of `DoWhenTest`'s 11 tests
fail with `QueryNotIdempotentException` once the classification is corrected.

Resolving this cleanly needs one of:
- A design decision on `do.when`'s classification semantics (e.g. per-call classification based on whether the
  literal `ifQuery`/`elseQuery` string arguments are themselves idempotent, when they're string literals - falls
  back to the current conservative `true` for dynamic/parameterized query strings), or
- Accepting the breaking change to `do.when`'s `query()`/`command()` contract and updating `DoWhenTest` to use
  `database.command(...)` for its read-only-in-practice calls, matching the convention `RefactorMergeNodesTest`/
  `CallStepWriteProcedureAutoCommitTest` already use for the other write procedures.

Either way, whichever fix ships needs new test coverage for the leader/follower `executionDatabase()` routing
itself (asserting the wrapped instance is chosen for a bare `CALL <write procedure>` statement) - the existing
`CallStepWriteProcedureAutoCommitTest` only exercises a plain local database and would not catch a regression
here.

## Related

- Surfaced by `claude[bot]`'s review of #6093 (fix for #6073).
- #6073 - write CypherProcedures aren't auto-committed (the fix that makes this misclassification observable).
- #5492, #5655 - prior instances of the same "wrong database instance -> silent Raft bypass" failure class for
  other write steps.

robfrank:
Closes #6094

## What was wrong

`SimpleCypherStatement`'s `readOnly` flag never looked at the statement's own `callClauses`. `anyWriteSubquery`
only inspects `CALL { ... }` blocks (added for #4191), so a bare top-level `CALL merge.node(...) YIELD node
RETURN node` - a statement with no CREATE/SET/MERGE/DELETE/REMOVE/FOREACH clause of its own - classified as
`isReadOnly() == true`.

That single flag backs three decisions:

| consumer | effect of the misclassification |
| --- | --- |
| `OpenCypherQueryEngine.executionDatabase()` | plans against the raw database instead of `getWrappedDatabaseInstance()`. On an HA leader `LocalDatabase.commit()` applies the pages locally and never proposes them to Raft (#5492, #5655). |
| `analyze().isIdempotent()` | `RaftReplicatedDatabase.command()` evaluates `isExecutedByTheLeader() \|\| isDDL() \|\| !isIdempotent()` as `false \|\| false \|\| false`, so a **follower executes the write entirely locally** instead of forwarding it to the leader. |
| `OpenCypherQueryEngine.query()`'s idempotency gate | `Database.query()` silently accepted and executed a write-procedure `CALL`. |

Harmless before #6073's fix (a write procedure with no caller transaction always threw before anything could
commit); live since PR #6093 gave `CallStep` an auto-commit.

## The fix

Three small pieces:

1. **`SimpleCypherStatement.anyWriteProcedureCall(callClauses)`** - mirrors `anyWriteSubquery`, resolving each
   `CallClause`'s procedure through `CypherProcedureRegistry.get()` (a static, case-insensitive map read that
   also strips the `apoc.` prefix, so both spellings classify identically) and folding the answer into
   `readOnly`. A name that resolves to no registered procedure is left alone. This also covers a write-procedure
   `CALL` nested inside `CALL { ... }`, since `anyWriteSubquery` already recurses through `inner.isReadOnly()`.
   It runs once per parse and parsed statements are cached per query text, so execution pays nothing.

2. **`Procedure.isWriteProcedure(Object[] literalArguments)`** - a new default method that lets a procedure
   refine its verdict for one particular call site, given the arguments the parser resolved to literals. The
   contract is documented as *narrow only*: `anyWriteProcedureCall` checks the no-arg `isWriteProcedure()`
   first and only consults the overload when that already said "writes", and every entry a procedure cannot
   resolve statically must keep the conservative answer.

3. **`DoWhen.isWriteProcedure(Object[])`** - the reason the mechanical fix was not landable on its own.
   `DoWhen.isWriteProcedure()` must return `true` unconditionally, because at registration time it cannot know
   what query strings its caller will hand it. At a call site that spells the branches out as string literals
   it *can* know: the call is a write only if a branch it could actually run is one, checked by parsing them
   (`Cypher25AntlrParser.parse(...).isReadOnly()`). Both branches are weighed, never just the one a literal
   condition selects - the condition is an arbitrary per-row expression, so which branch runs is not a
   parse-time fact.

   The line between "not a write" and "conservatively a write" is drawn at *can this call reach a branch at
   all*:

   - **Not a write** - a blank branch (runs nothing), a branch that parses read-only, a wrong argument count,
     and a branch argument that is a literal of the wrong type. The last two are rejected by `execute()`'s own
     checks before anything runs, so they cannot write; classifying them as writes would replace their
     actionable error (*"ifQuery must be a string"*) with `QueryNotIdempotentException` for every caller on
     `Database.query()`, which is a worse answer to the same mistake. Three `DoWhenTest` tests assert exactly
     those error types.
   - **Conservatively a write** - a branch supplied as `$param` (or a literal `null`, indistinguishable here),
     and a branch string that does not parse as written. The second is deliberately *not* treated as "cannot
     run": `OpenCypherQueryEngine` strips an `EXPLAIN`/`PROFILE` prefix before parsing and `PROFILE` does
     execute, so a bare parse failing here is not proof that nothing runs.

   An unparseable branch also must not make the *outer* query fail to parse - that error still belongs to
   execution, and a test pins it.

Without piece 3, correcting the classification reclassifies **every** `apoc.do.when(...)` call, read-only
branches included, and 10 of `DoWhenTest`'s 11 tests fail with `QueryNotIdempotentException` - they call
`database.query(...)` with branches that are read-only in practice. With it, 9 of those 10 keep passing
untouched.

## The one existing test that changed, and why

`DoWhenTest.writeSubQueryIsPersisted` asserted that

```java
database.query("opencypher", "CALL apoc.do.when(true, \"CREATE (n:Person {name: 'Bob'}) RETURN n\", '', {}) ...")
```

persists the node. That is the bug, written down as an expectation: it is a write executed through the API
reserved for idempotent statements. Its `query(` is now `command(`; **every assertion in it is unchanged**, and
the identical scenario was already covered through `command()` by
`CallStepWriteProcedureAutoCommitTest.doWhenAutoCommitsWriteSubQueryWithNoExplicitTransaction`. This is option
(b) from the issue's "Resolving this cleanly needs one of" list, narrowed by piece 3 from all 10 calls to just
this one.

## New tests

`engine/src/test/java/com/arcadedb/query/opencypher/Issue6094WriteProcedureCallClassificationTest` - 15 tests.
9 of them fail on `main` and pass here (verified by disarming the `readOnly` term and re-running: 14 run at that
point, 9 failures); the rest are the controls that must *not* change - a read-only procedure CALL stays
idempotent, an unregistered CALL target is untouched, a read-only `do.when` stays idempotent, a `do.when` its
own argument checks reject stays idempotent so its real error survives, `command()` still executes the write,
and a read-only CALL still keeps the raw instance.

The routing test is deliberately not vacuous. Off HA `getWrappedDatabaseInstance()` returns the database
itself, so asserting "the wrapped instance was chosen" would pass either way. The test installs a
distinguishable `DatabaseInternal` wrapper first via `LocalDatabase.setWrappedDatabaseInstance` - a
`java.lang.reflect.Proxy` delegating to the real instance, the same technique
`Issue5666ConcurrentGraphBatchTest` already uses, and the same installation point `RaftReplicatedDatabase`'s
constructor uses - then registers a probe `CypherProcedure` that records `context.getDatabase()` and asserts
the identity of what the plan handed it. A read-only twin of the same probe asserts the raw instance, so the
pair proves the switch discriminates rather than just leaning one way.

## Test plan

All four already run green on this branch:

- [x] `mvn -pl engine test -Dtest=Issue6094WriteProcedureCallClassificationTest` - **15/15**
- [x] `mvn -pl engine test -Dtest='com.arcadedb.query.opencypher.procedures.**,Issue6094...'` - **365, 0 failures**
      (DoWhenTest 11/11, RefactorMergeNodesTest, the algo/path/meta/refactor suites)
- [x] `mvn -pl engine test -Dtest='com.arcadedb.query.opencypher.**'` - **8239 run, 0 failures, 98 skipped**
      across the whole engine openCypher suite including the TCK
- [x] `mvn -pl engine test -Dtest='com.arcadedb.function.**,DoWhenTest,CallStepWriteProcedureAutoCommitTest'` -
      **1177, 0 failures** (the `Procedure` change is additive, but it is shared with the SQL engine)
- [x] The tests can fail: disarming the `readOnly` term turned 9 of them red, and each of those 9 asserts one of
      the three consumers above

## Known scope limits

- The classification is static. A write procedure reached through a mechanism the parser cannot see at all
  still classifies on the clauses present; this fix covers registered procedures named directly at a `CALL`
  site, which is every route in the engine today.
- `analyze().getOperationTypes()` for a statement whose only write is a procedure CALL falls into the existing
  conservative `{CREATE, UPDATE, DELETE}` fallback rather than reporting a narrower set. That is the same
  answer a writing `CALL { ... }` subquery already gets, and it is the safe direction for a permission check.
- `hasWriteBeforeMatch()` still does not count a `CALL` as a write clause for the optimizer's fast-path
  ordering check. That predicate is about clause *order* and is unchanged by this PR; a `CALL` before a `MATCH`
  behaves exactly as it did before.
- A literal `null` written for `do.when`'s `elseQuery` - `apoc.do.when(cond, 'RETURN 1', null, {})`, which
  legitimately means "no else branch" and never writes - is now conservatively classified a write, so that one
  rare spelling can no longer go through `Database.query()`. `literalArguments` cannot tell a literal `null`
  from an argument it could not resolve at all (both arrive as a `null` entry), and of the two only the
  conservative answer is safe. Spelling the same thing as `''` classifies read-only, and both are covered by
  `doWhenWithReadOnlyLiteralBranchesStaysIdempotent`.
- `DoWhen.branchMayWrite`'s nested parse has no explicit depth guard. It is bounded by the nesting literally
  written into the query text, and each level has to escape the quoting of the one around it, so the text grows
  exponentially in the depth - a guard would add a knob with no reachable failure behind it. The recursion is
  documented at the method and pinned by `doWhenNestedInsideADoWhenBranchIsClassifiedThroughBothLevels`, which
  checks the verdict travels out through both levels. Its `catch (Exception)` is deliberately broad for the
  same reason the "unparseable" case is conservative: any failure to decide has to fall to "assume it writes".

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/procedure/Procedure.java
+++ b/engine/src/main/java/com/arcadedb/function/procedure/Procedure.java
@@ -163,4 +163,40 @@ public interface Procedure {
   default boolean isWriteProcedure() {
     return false;
   }
+
+  /**
+   * Refines {@link #isWriteProcedure()} for one particular call site, given whatever that site's arguments the
+   * parser could resolve to constants.
+   * <p>
+   * {@link #isWriteProcedure()} answers for the procedure as a whole, at registration time, which for most
+   * procedures is the whole truth. It is not for a procedure that runs a caller-supplied query string
+   * ({@code apoc.do.when}): such a procedure has to answer {@code true} unconditionally, because at registration
+   * time it has no idea what its caller will pass. At a call site whose arguments are literals it does know, and
+   * this overload is where it says so - which matters because {@code isReadOnly()} on the enclosing statement
+   * decides HA routing, the {@code Database.query()} idempotency gate and the follower's forward-to-leader
+   * decision (issue #6094).
+   * <p>
+   * <b>A refinement may only narrow.</b> The caller checks {@link #isWriteProcedure()} first and only consults
+   * this overload when that answered {@code true}, so returning {@code true} here from a read-only procedure has
+   * no effect.
+   * <p>
+   * <b>{@code false} is a claim that needs proof.</b> An implementation may answer {@code false} only where it
+   * can show this call will not write - either because nothing it could run writes, or because the call cannot
+   * execute at all (its own argument checks reject it before any work starts, which is worth narrowing for:
+   * otherwise a caller on {@code Database.query()} gets {@code QueryNotIdempotentException} in place of the
+   * actionable error). Everywhere the arguments leave it genuinely unsure - a {@code null} entry, a value it
+   * cannot interpret - it must keep the conservative {@code true}; guessing "read-only" is what silently
+   * bypasses replication.
+   *
+   * @param literalArguments one entry per argument written at the call site, holding the argument's constant
+   *                         value when the parser resolved it to a literal and {@code null} otherwise (a
+   *                         parameter, a variable, any computed expression - and also a literal {@code null}, the
+   *                         two being indistinguishable here on purpose, since both must be treated
+   *                         conservatively). Never {@code null} itself; empty for a call with no arguments.
+   *
+   * @return true if this particular call can modify the database
+   */
+  default boolean isWriteProcedure(final Object[] literalArguments) {
+    return isWriteProcedure();
+  }
 }
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/SimpleCypherStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/SimpleCypherStatement.java
@@ -18,6 +18,9 @@
  */
 package com.arcadedb.query.opencypher.ast;
 
+import com.arcadedb.query.opencypher.procedures.CypherProcedure;
+import com.arcadedb.query.opencypher.procedures.CypherProcedureRegistry;
+
 import java.util.ArrayList;
 import java.util.List;
 
@@ -151,8 +154,9 @@ public class SimpleCypherStatement implements CypherStatement {
     this.hasRemove = hasRemove;
     final boolean hasForeach = this.clausesInOrder.stream().anyMatch(c -> c.getType() == ClauseEntry.ClauseType.FOREACH);
     final boolean writeSubquery = anyWriteSubquery(this.clausesInOrder);
+    final boolean writeProcedureCall = anyWriteProcedureCall(this.callClauses);
     this.readOnly = !hasCreate && !hasMerge && !hasDelete && !hasRemove && !hasForeach
-        && (setClause == null || setClause.isEmpty()) && !writeSubquery;
+        && (setClause == null || setClause.isEmpty()) && !writeSubquery && !writeProcedureCall;
 
     // Pre-compute flags used by CypherExecutionPlan.execute() to avoid repeated clause scanning
     this.hasVariableLengthPath = computeHasVariableLengthPath();
@@ -245,6 +249,55 @@ public class SimpleCypherStatement implements CypherStatement {
         return true;
     }
     return false;
+  }
+
+  /**
+   * Returns {@code true} when any top-level {@code CALL} in this statement targets a registered write
+   * {@link CypherProcedure}. Without this, {@code CALL merge.node(...) YIELD node RETURN node} - a statement with
+   * no CREATE/SET/MERGE/DELETE/REMOVE/FOREACH clause of its own - classified as read-only, and that flag is what
+   * {@code OpenCypherQueryEngine.executionDatabase()} uses to pick between the raw database instance and the
+   * Raft-aware wrapper. On HA the raw instance commits pages locally without proposing them to Raft, and a
+   * follower runs the statement locally instead of forwarding it to the leader, because the same flag backs
+   * {@code analyze().isIdempotent()}. Harmless until #6073 gave {@code CallStep} an auto-commit; live since.
+   * See issue #6094, and #5492/#5655 for the same failure class on other write steps.
+   * <p>
+   * A {@code CALL} nested in a {@code CALL { ... }} subquery is already covered: {@link #anyWriteSubquery}
+   * recurses through the inner statement's own {@code isReadOnly()}, which now accounts for this.
+   * <p>
+   * A name that resolves to no registered procedure is left alone - it is not this method's business to guess
+   * what a SQL-function fallback or an outright unknown name does. The registry lookup is a static
+   * case-insensitive map read that also strips the {@code apoc.} prefix, so both spellings of one procedure
+   * classify identically, and it costs nothing at execution time: this runs once per parse, and parsed
+   * statements are cached per query text.
+   */
+  private static boolean anyWriteProcedureCall(final List<CallClause> calls) {
+    if (calls.isEmpty())
+      return false;
+    for (final CallClause call : calls) {
+      final CypherProcedure procedure = CypherProcedureRegistry.get(call.getProcedureName());
+      // A per-call refinement may only narrow, so a procedure that never writes needs no second look.
+      if (procedure == null || !procedure.isWriteProcedure())
+        continue;
+      if (procedure.isWriteProcedure(literalArguments(call)))
+        return true;
+    }
+    return false;
+  }
+
+  /**
+   * The constant value of every argument at a call site, with {@code null} wherever the parser produced anything
+   * other than a literal (a parameter, a variable, a computed expression). Fed to
+   * {@link CypherProcedure#isWriteProcedure(Object[])} so a procedure whose behaviour depends on its arguments -
+   * {@code apoc.do.when} runs caller-supplied query strings - can classify one particular call rather than being
+   * pinned to the conservative answer it must give for the procedure as a whole.
+   */
+  private static Object[] literalArguments(final CallClause call) {
+    final List<Expression> arguments = call.getArguments();
+    final Object[] values = new Object[arguments.size()];
+    for (int i = 0; i < values.length; i++)
+      if (arguments.get(i) instanceof LiteralExpression literal)
+        values[i] = literal.getValue();
+    return values;
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/control/DoWhen.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/control/DoWhen.java
@@ -20,6 +20,7 @@ package com.arcadedb.query.opencypher.procedures.control;
 
 import com.arcadedb.database.Database;
 import com.arcadedb.query.QueryEngine;
+import com.arcadedb.query.opencypher.parser.Cypher25AntlrParser;
 import com.arcadedb.query.opencypher.procedures.CypherProcedure;
 import com.arcadedb.query.sql.executor.CommandContext;
 import com.arcadedb.query.sql.executor.Result;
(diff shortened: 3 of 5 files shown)


robfrank:
### CI triage: both red lanes are pre-existing, neither is this PR

`unit-tests` and `slow-unit-tests` are green. The two red lanes were investigated rather than assumed:

**`integration-tests`** - one failure, `com.arcadedb.remote.grpc.Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose`:

```
[the load violates the schema, so it must fail during the body]
Expecting actual not to be null
```

The **identical** test fails on `main` at `e46fc3606`, the commit this branch is based on (run
[31706409230](https://github.com/ArcadeData/arcadedb/actions/runs/31706409230), same assertion, same method).
It is a gRPC `GraphBatch` flush-timing test with no Cypher involvement.

**`ha-integration-tests`** - three failures: `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact`,
`Issue5064CommittedRemotelyContractIT.postQuorumLocalFailureSurfacesTheCommittedRemotelyContract`,
`RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`. None of the three sources mentions `cypher`
at all, so none can reach the code this PR changes. Ran locally on this branch:

- `Issue5569SlotMergeDeleteRaftIT` and `Issue5064CommittedRemotelyContractIT`: pass.
- `RaftPriorityRejoinIT`: 4 local runs on this branch gave 3 passes and 2 distinct failures - once
  `TransactionCommittedRemotely ... committed cluster-wide but the local apply failed`, once
  `DatabaseAreNotIdentical Types: DB1 6 <> DB2 5`. Two different signatures from one unchanged test is
  nondeterminism, not a regression. The same test passed on a control worktree checked out at the base commit
  `e46fc3606`.

The `ha-integration-tests` lane is also red on `main` (run
[31706228066](https://github.com/ArcadeData/arcadedb/actions/runs/31706228066)) with a *different* test
(`RaftTimeSeriesWriteReadYourWritesIT`), which is the shifting-failure-set pattern tracked in #5668.

PR #6165, opened 2026-08-13
fix(#5933): a compressible bulk transaction can no longer commit an entry no node can apply

### LINKED ISSUE #5933: HA: compressible bulk-insert transaction passes the replication size gate but bricks the whole cluster at apply time
## Summary

Reported by a community user running a 4/5-node Raft cluster (`cluster='heimdall'`) while doing a massive bulk insert migrating data from OrientDB. All nodes independently crashed at the same Raft log index and now crash-loop identically on every restart - the cluster cannot come back up on its own.

```
2026-08-07 12:47:20.620 SEVER [ArcadeStateMachine] CRITICAL: Unexpected error applying Raft log entry at index 45275. Shutting down to prevent state divergence.
java.lang.IllegalStateException: Invalid byte length 77158147 in TX_ENTRY uncompressed WAL (max 67108864)
        at com.arcadedb.server.ha.raft.RaftLogEntryCodec.checkByteLength(RaftLogEntryCodec.java:55)
```

A second node's log shows the identical index and identical exception a minute apart, confirming it isn't node-specific corruption.

## Root cause

Two independent, mismatched size checks guard `TX_ENTRY` replication:

1. **Submit-time check** - `RaftGroupCommitter.submitAndWait` (`ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftGroupCommitter.java:187`) rejects an entry whose *encoded envelope* size exceeds `messageSizeMax = min(arcadedb.ha.appendBufferSize, arcadedb.ha.grpcMessageSizeMax)` (32MB by default). The envelope is dominated by the WAL payload, which `RaftLogEntryCodec.encodeTxEntry` (`RaftLogEntryCodec.java:116-141`) **LZ4-compresses before framing**.
2. **Decode-time check** - `RaftLogEntryCodec.checkByteLength` (`RaftLogEntryCodec.java:44,53-57`), invoked from `decodeTxEntry` (`RaftLogEntryCodec.java:413-415`) on every node applying the committed entry, checks the **uncompressed** WAL length against `MAX_ENTRY_BYTES`, a hardcoded `64MB` constant that is unrelated to `appendBufferSize`/`grpcMessageSizeMax`, not configurable, and never mentioned in the startup banner that documents the replication size limit.

`replicateTransaction` (`RaftTransactionBroker.java:91-95`) has no pre-compression size check or chunking of its own (unlike `replicateSchema`, which does split oversized entries - see `splitSchemaEntry`, issue #4743). So a transaction whose **uncompressed** WAL is large but compresses well - exactly what bulk/migration inserts look like, being repetitive, text/JSON-heavy data - can:

- pass check #1, because its *compressed* size is under 32MB,
- get proposed, replicated to a Raft quorum, and **committed** to the log,
- then fail check #2 on every node that tries to apply it, because its *uncompressed* size (77,158,147 bytes in the report) exceeds the hardcoded 64MB (67,108,864 byte) ceiling.

For that gap to be crossed the payload only needs to compress by roughly 2.4x, which is unremarkable for LZ4 on WAL pages carrying repeated property names/textual values.

## Why this bricks the cluster instead of just failing one write

By the time the decode fails, the entry is already **durably committed** at a fixed Raft log index (a majority already has it). `ArcadeStateMachine.applyTransaction`'s catch-all (`ArcadeStateMachine.java:669-677`) treats the decode failure as an unexpected error and calls `triggerCriticalHalt()`, stopping the server so it "recovers via snapshot/log replay on the next start" (comment at `ArcadeStateMachine.java:681-689`). That recovery model assumes the failure is transient. Here it is not: replaying log index 45275 deterministically throws the same `IllegalStateException` on every node, every time, so restarting the whole cluster reproduces the identical crash - a permanent crash loop with no self-healing path, matching exactly what the reporter observed.

## Impact

Any deployment doing a large, well-compressing bulk transaction (batched inserts, data migration) can durably commit a `TX_ENTRY` that no node can ever apply, taking the entire cluster down simultaneously with no operator-side recovery short of a source patch and rebuild.

## Suggested fix

`checkByteLength`'s decode-time bound should not be an arbitrary constant sized independently of the real constraint. Options:

- Bound the claimed uncompressed length against `dis.available()` (bytes actually remaining in the entry being decoded), which is already transitively bounded by `messageSizeMax` at submission - this removes the mismatch for any compression ratio, not just today's numbers.
- Alternatively/additionally, give `replicateTransaction` the same size-aware chunking `replicateSchema`/`splitSchemaEntry` already has for oversized entries, so a single overlarge transaction is split rather than silently allowed through on compressed size alone.
- At minimum, raise `MAX_ENTRY_BYTES` to track `arcadedb.ha.grpcMessageSizeMax`'s configured ceiling instead of a hardcoded 64MB, so operators who raise the documented tunables don't hit an undocumented, lower, unconfigurable wall.

## Steps to reproduce

Not yet reduced to a minimal repro. The reported trigger: a multi-node Raft HA cluster (`arcadedb.ha.appendBufferSize` at its 32MB default) under a large batched-insert workload with well-compressing (text/JSON-heavy) data, where a single transaction's WAL is on the order of tens of MB uncompressed. A `RaftTransactionBrokerTest`/`RaftLogEntryCodecTest`-style unit test that encodes a `TX_ENTRY` with a highly-compressible ~70MB `walData` array (e.g. repeated bytes) should reproduce the mismatch directly: `entry.size() <= messageSizeMax` while `walData.length > RaftLogEntryCodec.MAX_ENTRY_BYTES`.

### ArcadeDB Version

Not stated by the reporter; `RaftLogEntryCodec.MAX_ENTRY_BYTES`/`checkByteLength` have existed unchanged since the original HA/Raft feature landed (#3731).

### OS and JDK Version

Not stated by the reporter.

robfrank:
Closes #5933

## The defect

Transaction replication was guarded by two size gates that measured different things:

| | where | what it measured | against |
|---|---|---|---|
| submit | `RaftGroupCommitter.submitAndWait` | the **encoded** entry, whose WAL `RaftLogEntryCodec.encodeTxEntry` has already **LZ4-compressed** | `min(arcadedb.ha.appendBufferSize, arcadedb.ha.grpcMessageSizeMax)` - 32MB by default |
| apply | `RaftLogEntryCodec.decodeTxEntry` | the **uncompressed** WAL length | a hardcoded `MAX_ENTRY_BYTES = 64MB`, unrelated to either setting |

A transaction only has to compress by ~2.4x to sit under the first gate and over the second, which is unremarkable for LZ4 on WAL pages carrying repeated property names and text/JSON values - exactly the profile of the reporter's OrientDB migration. Such an entry passes the submit gate, reaches a Raft quorum, and **commits durably at a fixed log index**. Only then does every node fail to decode it:

```
CRITICAL: Unexpected error applying Raft log entry at index 45275. Shutting down to prevent state divergence.
java.lang.IllegalStateException: Invalid byte length 77158147 in TX_ENTRY uncompressed WAL (max 67108864)
```

`ArcadeStateMachine` halts the node to avoid divergence, and its recovery model assumes the failure was transient. It is not: replay reproduces the same deterministic failure on the same index, on every node, on every restart. One oversized write bricks the entire cluster with no operator-side recovery.

## The fix

The two gates now agree by construction, in the one place that owns the wire format.

**1. The producer refuses to build an entry the consumer could not accept.** `encodeTxEntry` (and the WAL entries embedded in `encodeSchemaEntry`, which `splitSchemaEntry` also groups by raw size but only re-checks compressed) validate the *uncompressed* payload against `MAX_ENTRY_BYTES` and throw `ReplicatedEntryTooLargeException` - a `TransactionException`, deliberately **not** a `NeedRetryException`, since retrying resubmits the same undeliverable payload. The transaction fails on the leader, before anything is proposed, with a message that names the real constraint. `RaftReplicatedDatabase.replicateAndCommitLocally` already routes `ArcadeDBException` to "the entry never reached the log": ticket released, transaction rolled back.

`MAX_ENTRY_BYTES` stays at 64MB, which is above the 32MB default submit gate - so no *incompressible* transaction that replicates today (raw size ~= compressed size) starts being rejected.

**2. The consumer's ceiling is raised well above the producer's** (`MAX_DECODED_ENTRY_BYTES = 512MB`). This is what lets a cluster **already** holding such a committed entry apply it and come back up after the upgrade, instead of crash-looping on it forever. Conservative in what it sends, liberal in what it accepts: the asymmetry is the invariant, so entries produced by any version in the safe range remain appliable.

**3. Raising the decode ceiling does not turn the decoder into an allocation amplifier.** Two bounds, each exact for the thing it guards, both applied *before* any array is allocated:

- a length field for a section read **straight off the stream** (a compressed WAL, the schema JSON, the users JSON, the bootstrap fingerprint) is bounded by the bytes the entry **actually carries** - `dis.available()`, which is exact for a `ByteString`-backed stream and is already load-bearing in `decode` (the trailing-byte corruption check and the optional-section probes both depend on it). This is strictly tighter than the old flat 64MB constant, so a forged length now costs *less* than before the ceiling was raised, not more;
- a **decompressed** length is legitimately many times the bytes present, so `available()` cannot bound it. It is bounded instead by what those compressed bytes could possibly expand to - `255:1`, the LZ4 block format's asymptotic maximum, which a well-formed block can never reach. Genuine WAL ratios sit in the low single digits.

**4. The remaining way to recreate the mismatch is closed at startup.** The decode ceiling is a wire-format constant and is deliberately **not** configurable - a node whose ceiling differed from its peers' would apply entries they reject, which is the divergence the state machine halts to prevent. So `RaftPropertiesBuilder.build` now refuses at boot any configured entry cap above it, where an operator can act on it, rather than at apply time on an already-committed entry.

The `arcadedb.ha.appendBufferSize` documentation now states that compression does not make a transaction unbounded and that a second, fixed ceiling applies to the uncompressed WAL - the gap the issue called out as "never mentioned in the startup banner that documents the replication size limit".

## Alternatives considered

- **Bounding the claimed uncompressed length by `dis.available()`** (the issue's first suggestion). Wrong for a *decompressed* field, whose length is legitimately many times the bytes remaining - hence the expansion-ratio bound there instead. Applied to the lengths that *are* read straight off the stream it is exact, and it is applied (point 3 above).
- **Chunking `replicateTransaction` the way `replicateSchema` splits** (the issue's second suggestion). A schema change is safe to split because every prefix of the sequence is self-consistent; a transaction is not - its WAL must apply atomically, so followers would have to buffer chunks and apply them as a unit, with the leader-dies-mid-sequence case to reconcile. That is a feature, not a bug fix, and it does not help a cluster that is already down.
- **Making `MAX_ENTRY_BYTES` track `grpcMessageSizeMax`** (the issue's third suggestion). Per-node configuration of a decode ceiling is unsafe: two nodes with different values disagree about which entries are appliable, which is this bug with extra steps. Hence the fixed constant plus the startup validation.

## Test plan

New: `ha-raft/src/test/java/com/arcadedb/server/ha/raft/Issue5933CompressibleTxEntrySizeTest` (11 tests, well under a second).

- [x] **Red proven first**: with the guards neutralised, 5 of the 10 fail, one of them with the reporter's verbatim exception - `Invalid byte length 77158147 in TX_ENTRY uncompressed WAL (max 67108864)`.
- [x] The gap itself is asserted directly, not assumed: the reporter's 77,158,147-byte payload compresses to under the 32MB submit gate while its raw size is over the producer ceiling.
- [x] The transaction is rejected before it can be proposed, with a non-retryable `ReplicatedEntryTooLargeException` naming the database and the size.
- [x] Same for a `SCHEMA_ENTRY`'s embedded WAL.
- [x] Recovery: an entry written by the pre-fix encoder (hand-framed in the test, since the fixed encoder refuses to produce one) decodes byte-for-byte instead of halting the node.
- [x] The decode ceiling stays strictly above the producer ceiling, and the producer ceiling stays at or above the default submit gate.
- [x] A forged 400MB uncompressed length behind a handful of compressed bytes is refused on the expansion ratio, without allocating it; a length above the absolute ceiling is refused too.
- [x] A forged 300MB length for a section read straight off the stream - the compressed WAL of a `TX_ENTRY`, the schema JSON of a `SCHEMA_ENTRY` - is refused against the bytes the entry actually carries.
- [x] Ordinary transactions - compressible and incompressible - still round-trip unchanged, and a transaction exactly at the ceiling is still accepted (the check is `>`, not `>=`).
- [x] A configured entry cap above the decode ceiling fails at startup; the stock defaults still build.
- [x] Full `ha-raft` unit suite green.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1299,7 +1299,10 @@ public enum GlobalConfiguration {
       it is rejected with a state-machine error that makes the leader step down; the write then fails with \
       ReplicatedEntryTooLargeException naming this setting. The size compared against it is the COMPRESSED \
       WAL of the transaction, so text/JSON payloads shrink far below their raw size while incompressible \
-      ones (binary blobs, base64, encrypted fields, float vectors) map roughly 1:1. Raise it when single \
+      ones (binary blobs, base64, encrypted fields, float vectors) map roughly 1:1. Compression does NOT \
+      make the transaction unbounded, though (issue #5933): every node has to materialize the WAL in full to \
+      apply it, so a second, fixed ceiling of 64MB applies to the UNCOMPRESSED WAL of one entry regardless of \
+      this setting, and a transaction above it is rejected with the same exception. Raise it when single \
       transactions or records are bigger than the default, and raise arcadedb.ha.writeBufferSize with it \
       (it must stay >= this value + 8 bytes). Cost of raising it: a directly-allocated write buffer of \
       writeBufferSize per server, plus up to this many bytes of heap per follower appender during catch-up. \
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftLogEntryCodec.java
@@ -101,6 +101,27 @@ public final class RaftLogEntryCodec {
   }
 
   /**
+   * Validates a length field for a section read STRAIGHT off the stream, so the array it asks for is never
+   * allocated before the entry is known to actually carry that many bytes.
+   * <p>
+   * {@code remaining} is the exact number of bytes left in the entry: the stream is backed by a
+   * {@link ByteString}, whose {@code available()} is exact and is already load-bearing in {@link #decode}
+   * (the trailing-byte corruption check and the optional-section probes both depend on it). That makes it a
+   * far tighter bound than any constant - a corrupt or forged length is refused here rather than turning into
+   * a several-hundred-MB allocation that {@code readFully} would then fail on anyway. The absolute ceiling
+   * stays as a backstop.
+   * <p>
+   * This bound does NOT apply to the length of a section that is decompressed, which is legitimately many
+   * times the bytes present; {@link #checkDecompressedLength} covers that one.
+   */
+  private static void checkByteLength(final int length, final int remaining, final String context) {
+    checkByteLength(length, context);
+    if (length > remaining)
+      throw new IllegalStateException("Invalid byte length " + length + " in " + context + ": only " + remaining
+          + " bytes remain in the entry (truncated or corrupted replication payload)");
+  }
+
+  /**
    * Validates the uncompressed length a compressed section claims, before {@code decompress} allocates it.
    * Beyond the absolute ceiling the claim is also bounded by what the compressed bytes present in the entry
    * could possibly expand to, so raising {@link #MAX_DECODED_ENTRY_BYTES} cannot turn the decoder into an
@@ -494,7 +515,7 @@ public final class RaftLogEntryCodec {
   private static DecodedEntry decodeTxEntry(final DataInputStream dis, final String databaseName) throws IOException {
     final int uncompressedLength = dis.readInt();
     final int compressedLength = dis.readInt();
-    checkByteLength(compressedLength, "TX_ENTRY compressed WAL");
+    checkByteLength(compressedLength, dis.available(), "TX_ENTRY compressed WAL");
     checkDecompressedLength(uncompressedLength, compressedLength, "TX_ENTRY uncompressed WAL");
     final byte[] compressed = new byte[compressedLength];
     dis.readFully(compressed);
@@ -515,7 +536,7 @@ public final class RaftLogEntryCodec {
 
   private static DecodedEntry decodeSchemaEntry(final DataInputStream dis, final String databaseName) throws IOException {
     final int schemaLen = dis.readInt();
-    checkByteLength(schemaLen, "SCHEMA_ENTRY schemaJson");
+    checkByteLength(schemaLen, dis.available(), "SCHEMA_ENTRY schemaJson");
     final byte[] schemaBytes = new byte[schemaLen];
     dis.readFully(schemaBytes);
     final String schemaJson = new String(schemaBytes, StandardCharsets.UTF_8);
@@ -539,7 +560,7 @@ public final class RaftLogEntryCodec {
         for (int i = 0; i < walCount; i++) {
           final int walUncompressedLen = dis.readInt();
           final int walCompressedLen = dis.readInt();
-          checkByteLength(walCompressedLen, "SCHEMA_ENTRY WAL compressed");
+          checkByteLength(walCompressedLen, dis.available(), "SCHEMA_ENTRY WAL compressed");
           checkDecompressedLength(walUncompressedLen, walCompressedLen, "SCHEMA_ENTRY WAL uncompressed");
           final byte[] walCompressed = new byte[walCompressedLen];
           dis.readFully(walCompressed);
@@ -574,7 +595,7 @@ public final class RaftLogEntryCodec {
           final long expectedCrc = dis.readLong();
           final int uncompressedLen = dis.readInt();
           final int compressedLen = dis.readInt();
-          checkByteLength(compressedLen, "SCHEMA_ENTRY sealed blob compressed");
+          checkByteLength(compressedLen, dis.available(), "SCHEMA_ENTRY sealed blob compressed");
           checkDecompressedLength(uncompressedLen, compressedLen, "SCHEMA_ENTRY sealed blob uncompressed");
           final byte[] compressed = new byte[compressedLen];
           dis.readFully(compressed);
@@ -612,7 +633,7 @@ public final class RaftLogEntryCodec {
   private static DecodedEntry decodeBootstrapFingerprintEntry(final DataInputStream dis, final String databaseName)
       throws IOException {
     final int fpLen = dis.readInt();
-    checkByteLength(fpLen, "BOOTSTRAP_FINGERPRINT fingerprint");
+    checkByteLength(fpLen, dis.available(), "BOOTSTRAP_FINGERPRINT fingerprint");
     final byte[] fpBytes = new byte[fpLen];
     dis.readFully(fpBytes);
     final String fingerprint = new String(fpBytes, StandardCharsets.UTF_8);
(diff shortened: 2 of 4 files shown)


robfrank:
Thanks - both points were real, and one mechanism closes them together.

The `available()` bound is now applied to **every** length field for a section read straight off the stream: the compressed WAL of a `TX_ENTRY` and of a `SCHEMA_ENTRY`, the sealed-blob payload, the schema JSON, the users JSON and the bootstrap fingerprint. That is exact rather than heuristic, so it is strictly *tighter* than the 64MB constant it replaces - a forged length now costs less than it did before the ceiling was raised, which also covers the `schemaJson` asymmetry you flagged in the nits without adding a producer-side check for an uncompressed field.

On why it was not applied to the *decompressed* length: there it genuinely is a heuristic - the uncompressed size is legitimately many times the bytes remaining - so that one keeps the `255:1` LZ4 expansion bound instead. The Javadoc on the new overload now says which of the two applies where, and the PR description has been rewritten accordingly.

Relying on `available()` is not new trust, incidentally: `decode` already depends on it being exact for the trailing-byte corruption check and for probing the optional trailing sections.

New coverage: `aLengthLongerThanTheEntryIsRefusedBeforeItIsAllocated` forges a 300MB length for both a `TX_ENTRY` compressed WAL and a `SCHEMA_ENTRY` schema JSON. Full `ha-raft` unit suite re-run: 836 tests, 0 failures.

robfrank:
No code change for this round. On the one residual observation - "a corrupted/forged length field can now trigger a larger allocation attempt than the old flat 64MB cap could, before the ratio/\`available()\` bounds catch it" - the ordering is the other way round, which is what makes the raised ceiling safe:

```java
final int compressedLength = dis.readInt();
checkByteLength(compressedLength, dis.available(), "TX_ENTRY compressed WAL");
checkDecompressedLength(uncompressedLength, compressedLength, "TX_ENTRY uncompressed WAL");
final byte[] compressed = new byte[compressedLength];   // <- only now
```

Both guards run *before* the array is allocated, at every site. So a 512MB allocation is only ever reached when the entry genuinely carries 512MB (for a section read straight off the stream) or when the compressed bytes it does carry could legitimately expand that far (for a decompressed section). A forged length is refused at the `available()` bound, which is *tighter* than the 64MB constant it replaced - the amplification ceiling went down with this change, not up.

Verification for this revision: full `ha-raft` unit suite 836/836 green, plus `Issue4743LargeRecordReplicationIT`, `Issue4743CompactionEntrySizeIT` and `Issue4458TsWalVersionGapIT` (the sealed-blob path) 3/3 green.

Issue #5219, opened 2026-07-11
Gremlin runtime `NoSuchElementException` misclassified as `CommandParsingException`


shulei5831sl:
### ArcadeDB version

Observed on latest Docker image:

```text
docker pull arcadedata/arcadedb:latest
Pulled: 2026-07-11

ArcadeDB version: 26.8.1-SNAPSHOT
Build: d758998ce9854d21aadce81987cd1dc036488c49
```

### Environment

- Host OS: Linux 5.4.0-216-generic
- Deployment: Docker `arcadedata/arcadedb:latest`
- Query interface: HTTP `/api/v1/command/<db>` with `language: gremlin`
- Query language: Gremlin

### Description

When a Gremlin query produces a `NoSuchElementException` (a runtime condition where `.next()` has no traverser to return), ArcadeDB wraps it as `CommandParsingException` with HTTP 400. The query syntax is valid — the error is a runtime condition, not a parse error.

### To reproduce

Submit via HTTP command endpoint with `language: gremlin`:

```gremlin
g.addV("Bug").property("id", 1).iterate()
g.V().hasLabel("Bug").drop().next()
```

Per TinkerPop specification, `drop()` is a filter/sideEffect step that produces no outgoing objects. `.next()` after `.drop()` correctly produces `NoSuchElementException` because there are no traversers.

ArcadeDB's `drop()` behavior is correct — `.drop().iterate()` works, elements are deleted.

### Actual behavior (error classification issue)

```json
{
  "error": "Cannot execute command",
  "exception": "com.arcadedb.exception.CommandParsingException",
  "detail": "Error on executing Gremlin query -> Error on parsing gremlin query: java.util.NoSuchElementException -> java.util.NoSuchElementException -> NoSuchElementException"
}
```

### Expected behavior

The request may fail because `.next()` is called on a traversal that emits no result, but the failure should be classified as a runtime traversal/execution error.

For example, ArcadeDB could return a user-facing error indicating that no element was available for `.next()`.

The exact exception wrapper is implementation-specific, but it should not be reported as `CommandParsingException`, because the Gremlin syntax was parsed successfully and the `NoSuchElementException` occurred during traversal execution.

### Why this is minor

- `drop()` behavior is correct per Gremlin specification
- `.drop().iterate()` works as expected for all element types
- The issue is purely about exception classification, not functional correctness

lvca:
Fixed in `main` (commit b743cbac6), shipping in the next release.

[...]

Issue #31, opened 2021-09-03
Engine: NPEs accessing rid when it is null


giastfader:
### OS: N/A

I found at least two points where there is an explicit use of the rid variable when it is for sure null, because there is an explicit test on the null condition.

https://github.com/ArcadeData/arcadedb/blob/bb8556071be5258bae074cec7ae0ac2b2b5196a0/engine/src/main/java/com/arcadedb/query/sql/parser/Rid.java#L167

```Java

    public PInteger getBucket() {
        if (expression != null) {
            RID rid = toRecordId((Result) null, new BasicCommandContext());
            if (rid == null) {
                PInteger result = new PInteger(-1);
                result.setValue(rid.getBucketId());
                return result;
            }
        }
        return bucket;
    }

    public PInteger getPosition() {
        if (expression != null) {
            RID rid = toRecordId((Result) null, new BasicCommandContext());
            if (rid == null) {
                PInteger result = new PInteger(-1);
                result.setValue(rid.getPosition());
                return result;
            }
        }
        return position;
    }

```

lvca:
I believe we get that code from OrientDB when we ported the SQL layer.

antarcticgiraffe:
Hi @lvca @giastfader 

I checked OrientDB code and it's exactly the same.
I agree with you that this logic makes no sense at all. Probably the null check is just inverted, ie. if the expression exists, try to use it to calculate the RID and in case (rid **!=** null) use it to return the bucket or the position; otherwise just use the raw bucket/position value.

I'm fixing the code in this sense.

We should also report the problem to OrientDB project, as it could be of interest.

Regards

AG

antarcticgiraffe:
Fixed with https://github.com/ArcadeData/arcadedb/commit/11b6800689b0a741b16ebf77c5fda0de590951d5

PR #6166, opened 2026-08-13
fix(#5898): a Cypher property access through a broken link raises a TypeError, not RecordNotFoundException

### LINKED ISSUE #5898: Cypher property access: dereferencing a broken RID raises RecordNotFoundException instead of a Cypher-flavored error
### Context

Follow-up from the review discussion on #5893 (fix for #5800).

Both `com.arcadedb.query.opencypher.ast.PropertyAccessExpression` (variable-bound property
access, e.g. `WITH holder.ref AS r RETURN r.id`) and, as of #5893,
`com.arcadedb.query.opencypher.parser.CypherExpressionBuilder.ChainedPropertyAccessExpression`
(chained property access, e.g. `holder.ref.id`) dereference a persisted node-valued property
(stored as an ArcadeDB LINK/RID) via `rid.asVertex().get(propertyName)`.

If the RID no longer resolves to a live `Vertex` - the referenced record was deleted, or the
LINK actually points to a plain `Document`/`Edge` rather than a vertex - `RID.asVertex()`
throws `RecordNotFoundException` (or `ClassCastException`), not the `TypeError: Cannot access
property '...'` that the rest of these two `evaluate()` methods produce for other unsupported
base types. Through the HTTP command API this surfaces as an unfriendly 500 rather than a
Cypher-flavored client error.

This behavior predates #5893 on the `PropertyAccessExpression` (single-level) path - it was
previously reachable only via algorithm procedures storing raw RIDs for lazy vertex resolution.
#5893 makes the same failure mode reachable from ordinary chained property access
(`holder.ref.id`), which is a much more common shape of Cypher query, so it's worth fixing
properly rather than leaving as a latent edge case.

### Suggested fix

Catch the resolution failure in both `evaluate()` methods and re-raise it as a
`CommandExecutionException` with a message consistent with the existing `TypeError: Cannot
access property '<name>' on ...` wording used for other unsupported base types, so HTTP callers
get a client-error-shaped response instead of a raw 500.

### Test coverage gap

Neither `PropertyAccessExpression` nor `ChainedPropertyAccessExpression` currently has a
regression test for a RID pointing at a deleted or non-vertex record. Worth adding once the fix
above lands.

### Related

- #5800 (original bug)
- #5893 (fix PR; review discussion flagged this as a pre-existing, now more broadly reachable, gap)

robfrank:
Closes #5898

## Summary

Both Cypher property-access paths that dereference a persisted node-valued property - `PropertyAccessExpression` (variable-bound, `WITH holder.ref AS r RETURN r.id`) and `CypherExpressionBuilder.ChainedPropertyAccessExpression` (chained, `holder.ref.id`, reachable from ordinary Cypher since #5893) - resolved the stored LINK with `rid.asVertex()`. That call throws `RecordNotFoundException` when the RID no longer resolves, and `RID.asVertex()` also converts its own `ClassCastException` into a `RecordNotFoundException`, so a link to a plain document or an edge failed the same way. Either shape let an engine-level exception escape the query engine instead of the `TypeError: Cannot access property '<name>' on ...` these two `evaluate()` methods produce for every other base type they cannot read from. Over the HTTP command API the auto-commit wrapper in `DatabaseAbstractHandler` wraps it in a `TransactionException`, and `AbstractServerHttpHandler` - which classifies by `getCause()` - finds no arm for `RecordNotFoundException` and falls through to the catch-all "Error on transaction commit", hiding the real reason.

The dereference now lives in one place, `PropertyAccessExpression.readLinkedProperty(rid, propertyName)`, called by both paths, and it resolves the RID to a `Document` rather than to a `Vertex`. That is the substantive half of the fix: vertex, edge and plain document are all `Document`s, and the branch immediately after the RID branch in both callers already reads a live value of any of those three shapes exactly that way - so a link to a non-vertex record is not broken at all, and the `ClassCastException` path disappears rather than being re-labelled. What remains a genuine failure is a RID that resolves to nothing (the referenced record was deleted and nothing rewrote the property holding the link); that is reported as a `CommandExecutionException` carrying the existing TypeError wording plus the RID and the reason.

One deliberate detail, pinned by an assertion in the test: the `RecordNotFoundException` is **not** attached as the cause. `AbstractServerHttpHandler` classifies a `CommandExecutionException` by its cause when it has one, so attaching it would route the response straight back to the "Error on transaction commit" arm this change exists to get away from. Nothing is lost - the RID and the reason are both in the message.

### Scope note on the HTTP status code

The issue describes the symptom as "an unfriendly 500". After this change the status is still 500, now via the `CommandExecutionException` arm ("Cannot execute command") carrying the TypeError message, because that is the deliberate mapping for *every* Cypher `TypeError` (see the #5219 rationale comment in `AbstractServerHttpHandler`, matching TinkerPop's `SERVER_ERROR_SCRIPT_EVALUATION`). Making a Cypher `TypeError` a 4xx is a separate decision about the whole family, not something to special-case for broken links, so it is not taken here. What this PR delivers is the consistency the issue actually asks for: the failure is now the same Cypher-flavored error as every other unreadable base type, with a message that names the RID, instead of an engine-internal exception surfacing as "Error on transaction commit".

Adjacent gap noticed and left alone: an *unwrapped* `RecordNotFoundException` maps to 404, but a wrapped one has no arm and degrades to 500. That asymmetry is in the server module and unrelated to Cypher property access.

## Test plan

New regression test `engine/src/test/java/com/arcadedb/query/opencypher/CypherBrokenLinkPropertyAccessIssue5898Test.java`, four cases covering both access paths against both failure shapes. All four fail on `main` (2 with the raw `RecordNotFoundException: Record #1:1 not found` from the dangling link, 2 with `Record #1:0 not found` from the `asVertex()` cast on a document link) and pass with the fix.

- [ ] `chainedAccessOnDanglingLinkRaisesCypherTypeError` - `holder.ref.id` where `ref` points at a deleted vertex raises `CommandExecutionException` with `TypeError: Cannot access property 'id'`, `the linked record does not exist`, and no cause.
- [ ] `variableBoundAccessOnDanglingLinkRaisesCypherTypeError` - same for `WITH holder.ref AS r RETURN r.id`.
- [ ] `chainedAccessOnNonVertexLinkReadsThroughTheLink` - `holder.ref.id` where `ref` is a LINK to a plain document returns the property instead of failing.
- [ ] `variableBoundAccessOnNonVertexLinkReadsThroughTheLink` - same for the variable-bound path.
- [ ] `CypherNodeValuedPropertyDereferenceIssue5800Test` (3 tests) still green - the #5800/#5893 behavior, including the temporal-type restoration through a chained link, is unchanged.
- [ ] Full `com.arcadedb.query.opencypher.**` package: 8237 tests, 0 failures, 0 errors.

```
mvn -pl engine test -Dtest='com.arcadedb.query.opencypher.**' -DexcludedGroups=benchmark,slow,vector
```

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/PropertyAccessExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/PropertyAccessExpression.java
@@ -20,7 +20,9 @@ package com.arcadedb.query.opencypher.ast;
 
 import com.arcadedb.database.Document;
 import com.arcadedb.database.RID;
+import com.arcadedb.database.Record;
 import com.arcadedb.exception.CommandExecutionException;
+import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.query.opencypher.executor.DeletedEntityMarker;
 import com.arcadedb.query.opencypher.temporal.*;
 import com.arcadedb.query.sql.executor.CommandContext;
@@ -54,8 +56,7 @@ public class PropertyAccessExpression implements Expression {
     if (variable instanceof RID rid) {
       // Lazy vertex resolution: algorithm procedures store RIDs to avoid loading all vertices upfront.
       // Only resolve to Document when a property is actually accessed.
-      final Object rawValue = rid.asVertex().get(propertyName);
-      return TemporalUtil.convertFromStorage(rawValue);
+      return readLinkedProperty(rid, propertyName);
     } else if (variable instanceof Document) {
       final Object rawValue = ((Document) variable).get(propertyName);
       return TemporalUtil.convertFromStorage(rawValue);
@@ -81,6 +82,45 @@ public class PropertyAccessExpression implements Expression {
     throw new CommandExecutionException(
         "TypeError: Cannot access property '" + propertyName + "' on " +
         variable.getClass().getSimpleName() + " value");
+  }
+
+  /**
+   * Reads {@code propertyName} through a persisted LINK, shared by this (variable-bound) access path and the chained
+   * access path in {@code CypherExpressionBuilder.ChainedPropertyAccessExpression} so both dereference a RID identically.
+   * <p>
+   * The RID is resolved to a {@link Document} rather than to a {@code Vertex}: every property-bearing record - vertex,
+   * edge or plain document - is a {@code Document}, and the caller's adjacent {@code instanceof Document} branch already
+   * reads a live value of any of those three shapes the same way, so a LINK to a non-vertex record is not an error. It
+   * also removes the {@code ClassCastException} that {@code RID.asVertex()} turned into a {@link RecordNotFoundException}.
+   * <p>
+   * What remains a failure is a RID that resolves to nothing, typically because the referenced record was deleted and
+   * nothing rewrote the property holding the link. That is reported as a {@link CommandExecutionException} carrying the
+   * same {@code TypeError: Cannot access property '<name>' on ...} wording the callers use for every other base type
+   * they cannot read a property from, so the query fails the way a Cypher type error fails instead of letting an
+   * engine-level {@link RecordNotFoundException} escape the query engine (issue #5898).
+   * <p>
+   * The {@code RecordNotFoundException} is deliberately NOT kept as the cause: {@code AbstractServerHttpHandler}
+   * classifies a {@code CommandExecutionException} by {@code getCause()} when there is one, so attaching it would route
+   * the response through the catch-all arm ("Error on transaction commit") instead of the Cypher-error arm ("Cannot
+   * execute command"), i.e. re-introduce the very message this fix removes. Nothing is lost - the RID and the reason
+   * are both in the message.
+   */
+  public static Object readLinkedProperty(final RID rid, final String propertyName) {
+    final Record record;
+    try {
+      record = rid.getRecord();
+    } catch (final RecordNotFoundException e) {
+      throw new CommandExecutionException(
+          "TypeError: Cannot access property '" + propertyName + "' on " + rid +
+          ": the linked record does not exist");
+    }
+
+    if (!(record instanceof Document document))
+      throw new CommandExecutionException(
+          "TypeError: Cannot access property '" + propertyName + "' on " + rid +
+          ": the linked record has no properties");
+
+    return TemporalUtil.convertFromStorage(document.get(propertyName));
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/PropertyAccessExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/PropertyAccessExpression.java
@@ -110,17 +110,22 @@ public class PropertyAccessExpression implements Expression {
     try {
       record = rid.getRecord();
     } catch (final RecordNotFoundException e) {
-      throw new CommandExecutionException(
-          "TypeError: Cannot access property '" + propertyName + "' on " + rid +
-          ": the linked record does not exist");
+      throw new CommandExecutionException(brokenLinkMessage(rid, propertyName, "the linked record does not exist"));
     }
 
     if (!(record instanceof Document document))
-      throw new CommandExecutionException(
-          "TypeError: Cannot access property '" + propertyName + "' on " + rid +
-          ": the linked record has no properties");
+      throw new CommandExecutionException(brokenLinkMessage(rid, propertyName, "the linked record has no properties"));
 
     return TemporalUtil.convertFromStorage(document.get(propertyName));
+  }
+
+  /**
+   * Keeps the two broken-link failures on one wording. The RID takes the place the generic sibling messages give to the
+   * base type's class name: with the record gone, the link itself is the only thing left that identifies what could not
+   * be read, and it is what a caller needs to find the property still holding it.
+   */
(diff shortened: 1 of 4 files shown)


robfrank:
Both points from the last review are addressed in `eb68425`.

**HTTP end-to-end gap - taken.** `server/src/test/java/com/arcadedb/server/http/Issue5898BrokenLinkPropertyAccessHttpTest.java` drives the dangling link through `DatabaseAbstractHandler` -> `AbstractServerHttpHandler` and asserts the classification, not just the status: `error` must be `"Cannot execute command"` and `detail` must contain the `TypeError`. Verified it fails on the pre-fix behavior - with `asVertex()` restored and the exception rethrown raw, `detail` comes back as `"Record #31:0 not found"` - and that the control (`aLiveLinkStillDereferences`) passes in both states, so the test cannot pass by breaking every dereference. Asserting the status alone would have been useless here: both shapes answer 500.

**`RID.asDocument()` - declined, with the reason now in the Javadoc.** `asDocument()` is `(Document) lookupByRID(...)`, so a wrong-shaped record leaves it as a bare `ClassCastException`. Adopting it means either letting that escape - reintroducing exactly the class of bug this PR removes - or wrapping the call in `catch (ClassCastException)`, which would also swallow an unrelated cast failure raised deeper inside the lookup and report it to the user as "the linked record has no properties". The `instanceof` answers only the question being asked and cannot mask an engine fault, which is worth one extra import and one branch. That is also why the branch is kept rather than dropped as unreachable: its job is to stay correct if `Record` ever gains another non-`Document` implementation, not to be reachable today.

**Also noted, not taken:** the missing `TemporalUtil.convertFromStorage` on `ChainedPropertyAccessExpression`'s `instanceof Document` branch is real and pre-existing - a chained access reading a live (uncommitted) node value returns a raw stored String where the same expression after commit, going through the LINK path, returns the restored temporal. It is a separate defect from this one and belongs in its own issue rather than riding along here.

The edge-typed LINK cases from the previous round were likewise verified to fail with `asVertex()` restored (`TypeError ... on #4:0: the linked record does not exist`), so they exercise the `ClassCastException` path rather than passing by analogy.

robfrank:
Addressed in `0446103`.

**Stale comment - fixed.** The `RID` branch in `PropertyAccessExpression.evaluate` no longer says "Lazy vertex resolution"; it now names both callers of the branch (a persisted node-valued property stored as a LINK, and the algorithm procedures' deferred RIDs) and says the record is loaded only when a property is read.

**Javadoc length - trimmed.** Rewritten as three bullets covering only the decisions that are load-bearing and easy to undo by accident (Document-not-Vertex, `getRecord()`+`instanceof` over `asDocument()`, and the omitted cause). 25 lines out, 18 in. One correction though: the rule quoted as CLAUDE.md's ("only add a comment when the WHY is non-obvious... don't reference the current task/fix") is not in this repo's `CLAUDE.md` or `engine/CLAUDE.md` - I checked before acting on it. Referencing issue numbers in comments is in fact the prevailing convention here (`engine/CLAUDE.md` and the surrounding Cypher code are full of `#NNNN` references), so `#5898` stays; the trim is on the restatement, not the reference.

**Untested `"has no properties"` branch - left as is, deliberately.** Reaching it needs a LINK pointing at a non-`Document` `Record`, i.e. an internal storage structure like `EdgeSegment`. There is no supported way to write one into a property, so a test would have to manufacture a state the engine cannot produce - it would pin the fabrication, not the behavior. The branch exists so that the failure stays a Cypher `TypeError` if `Record` ever gains another non-`Document` implementation, which is exactly the kind of guarantee that should not need a fake fixture to hold.

Re the note about not being able to run the build: the suites were run locally on this branch - 6 engine cases in `CypherBrokenLinkPropertyAccessIssue5898Test`, 3 in `CypherNodeValuedPropertyDereferenceIssue5800Test`, 2 in `Issue5898BrokenLinkPropertyAccessHttpTest`, all green after this commit, on top of the full `com.arcadedb.query.opencypher.**` package at 8237 passing.

robfrank:
No code changes this round: the latest review raises no actionable item on this PR. Recording where the three notes land, so nothing is silently dropped.

**`getRecord()` returning `null`** would be labelled "the linked record has no properties" rather than "does not exist", since `instanceof` on `null` is false. Checked the path: `RID.getRecord()` delegates to `LocalDatabase.lookupByRID`, which throws `RecordNotFoundException` on a missing slot rather than returning `null` - that is the exact exception the dangling-link test observed before the fix (`Record #1:1 not found`, from `LocalBucket.getRecord`). So the mislabel is unreachable rather than latent, and both messages are a Cypher `TypeError` naming the same RID either way; the difference would be one clause of the reason. Not worth a branch that no test can reach.

**Two pre-existing defects, both outside this PR's scope, both worth their own issue:**
1. `ChainedPropertyAccessExpression.evaluate()` does not call `DeletedEntityMarker.checkNotDeleted(baseValue)`, which `PropertyAccessExpression.evaluate()` does at the top. A node deleted in the same transaction should raise `DeletedEntityAccess`, and today only the variable-bound path does.
2. The same method's `instanceof Document` branch skips `TemporalUtil.convertFromStorage`, so a chained read of a live (uncommitted) node value returns a raw stored String where the same expression after commit returns the restored temporal. Same divergence, same method, opposite direction from the one #5893 fixed.

Both are the same shape of gap - the chained path is missing something the variable-bound path has - so they probably belong in one follow-up issue rather than two.

CI is green and the branch is mergeable. Merge is the maintainer's call.

PR #6169, opened 2026-08-14
fix(#6167): a SQL trigger body that is a SELECT never ran

### LINKED ISSUE #6167: A SQL trigger body that is a SELECT never executes: the result set is closed without being iterated
Found while fixing the `BEFORE READ` trigger recursion (`a040312f9`), and separable from it.

## What happens

`SQLTriggerExecutor.execute` runs the trigger body and discards the result without consuming it:

```java
database.command("sql", sql, params).close();
return true;
```

A `SELECT` result set in ArcadeDB is **lazy** - closing it without iterating produces no rows and touches no record. So a trigger whose body is a `SELECT` does nothing at all. It is not slow or partial; it is a no-op, and it fails silently: the DDL succeeds, the trigger is registered, it fires on schedule, and its body has no effect.

DML bodies are unaffected: `INSERT`/`UPDATE`/`DELETE` execute eagerly and work correctly through the same call.

## Measured

Counting Java triggers on both read timings, over a 3-record type with one record relocated behind a placeholder:

| what the caller does | BEFORE READ fires | AFTER READ fires |
| --- | --- | --- |
| `SELECT name FROM P` (iterated) | 3 | 3 |
| `lookupByRID` (in-page, and relocated) | 1 | 1 |
| `command("SELECT name FROM P").close()` **without iterating** | **0** | **0** |

And directly, doing exactly what the executor does - `command(...).close()`, never iterating:

| body | effect |
| --- | --- |
| `INSERT INTO Log SET n = 1` | row inserted (Log count 1) |
| `UPDATE Src SET n = 99` | update applied (n = 99) |
| `DELETE FROM Src` | delete performed (count 0) |
| `SELECT ...` | nothing |

## Why it matters

A `SELECT` body is a reasonable thing to write in a trigger, and the two obvious uses both fail quietly:

- **validation** - `SELECT FROM Other WHERE ...` intended to abort the operation. The rows are never produced, so nothing is checked and nothing aborts. Note the return value cannot rescue this either: `SQLTriggerExecutor.execute` returns a hardcoded `true`, so a SQL trigger can never veto regardless.
- **side-effecting reads** - anything relying on the query running for its events or its effect on caches.

There is no error, no warning, and no test coverage: `TriggerSQLTest` creates trigger bodies like `'SELECT 1'` and asserts the trigger was registered, never that a body had an effect.

## Options

1. **Consume the result set** - iterate to exhaustion before closing. Makes SELECT bodies do what they look like they do. Cost: a trigger on a hot path now pays for a full result set, on every fired record.
2. **Reject a SELECT body at `CREATE TRIGGER`**, so the failure is at DDL time rather than silent forever. Cheapest and honest, if a read-only body is never useful given the executor cannot veto anyway.
3. **Document it** and leave the behaviour - weakest, since the statement parses and looks like it works.

Worth deciding alongside whether SQL trigger bodies should be able to abort at all, since (1) is only useful if the answer to that is yes.

## Related

Also noticed and NOT filed, because it is an inconsistency rather than a bypass: `Database.scanType(...)` fires no read events at all (0 before, 0 after, with or without property access) while the SQL path fires them normally. That is a low-level Java API, not something a query reaches.

lvca:
Fixes #6167.

## Problem

`SQLTriggerExecutor.execute`/`executeBeforeRead` ran the trigger body with `database.command(...).close()`, discarding the result set without ever iterating it. A `SELECT` result set in ArcadeDB is pull-based - nothing runs until it is pulled - so a `SELECT`-bodied trigger was a silent no-op: the DDL succeeded, the trigger registered, it fired on schedule, and its body never executed. `DML` bodies (`INSERT`/`UPDATE`/`DELETE`) were unaffected because their execution plans run eagerly when the command is built (confirmed by reading `InsertExecutionPlan`/`UpdateExecutionPlan`: they execute in `reset()`/first `fetchNext()` and buffer their rows).

## Fix

Drain the result set to exhaustion instead of closing it unread. This:

1. Fixes the no-op: a `SELECT` body now actually runs.
2. Gives SQL trigger bodies the same veto contract `ScriptTriggerExecutor` already has for JavaScript (`return false;` aborts the operation): a SQL body that evaluates to a single scalar `false` - one row, one property - now aborts the operation too, which is what `SELECT false` or `SELECT <condition> AS ok` looks like it should do. A multi-row or multi-column result has no unambiguous single answer, so it is drained for its effects and treated as a pass, same as `DML` always was.

The issue raised three options and didn't pick one. I went with a strengthened version of option 1 (consume the result set) rather than option 2 (reject `SELECT` bodies at `CREATE TRIGGER`): once the body can veto, a read-only body is genuinely useful - both for validation (`SELECT ... WHERE ...` style checks) and for side-effecting reads - so rejecting it at DDL time would remove a legitimate use case rather than fix a dead one. This also brings SQL to parity with the JavaScript executor instead of introducing new syntax or semantics.

Backward compatible: `DML` bodies never produce a single-property boolean-`false` row, so their behavior is unchanged. No previously-working `SELECT`-bodied trigger can regress, because no `SELECT` body ever ran before this fix.

## Tests

- `TriggerSQLTest.beforeCreateTriggerSelectBodyActuallyExecutes` - proves a `SELECT` body now runs, by calling a Java function with a real side effect from a plain projection.
- `TriggerSQLTest.beforeCreateTriggerAbortSQL` - `SELECT false` aborts `BEFORE CREATE`, mirroring the existing JavaScript abort test.
- `BeforeReadTriggerTest.aSqlBeforeReadTriggerCanVetoTheRead` - SQL `BEFORE READ` gets the same veto coverage as the Java and JavaScript arms.
- Updated a stale comment in `BeforeReadTriggerTest` that described the old lazy-and-unconsumed behavior.

Verified the three new assertions fail without the fix (stashed the production change, reran) and pass with it.

## Test plan

- [x] `TriggerSQLTest`, `BeforeReadTriggerTest`: 36/36 pass
- [x] Full `engine` module (`mvn verify`, 12184 tests): 1 pre-existing, unrelated failure (`LSMVectorIndexRebuildTest` rebuild-timeout flake, a known CPU-starvation flake unrelated to triggers)

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1488,4 +1488,24 @@ null on every imported record, and under the new one would be rejected. It now i
 is a valid ArcadeDB default expression - `sysdate()`, numbers, already-quoted strings keep their meaning - and quotes
 it as a string literal otherwise, logging which defaults it had to translate.
 
+## A SQL trigger body that is a `SELECT` now actually runs, and can veto (#6167)
+
+`SQLTriggerExecutor` ran a trigger's SQL body with `database.command(...).close()`, discarding the result set
+without ever iterating it. A `SELECT` result set in ArcadeDB is pull-based - nothing runs until it is pulled - so a
+`SELECT`-bodied trigger was a silent no-op: the DDL succeeded, the trigger registered, it fired on schedule, and its
+body never executed. `INSERT`/`UPDATE`/`DELETE` bodies were unaffected, because their execution plans run eagerly
+when the command is built.
+
+The body's result set is now drained to exhaustion instead of closed unread, which fixes the no-op and also gives a
+SQL trigger body the same veto contract the JavaScript executor already had: a body that evaluates to a single
+scalar `false` - one row, one real property (`$score`/`$similarity` search-scoring metadata don't count) - aborts
+the operation, which is what `SELECT false` or `SELECT <condition> AS ok` looks like it should do. The check is
+scoped to bodies that are actually idempotent (`SELECT`), so an existing `UPDATE ... RETURN AFTER <field>` (or
+`RETURN BEFORE`) trigger - which can produce that same one-row, one-property shape - is never misread as a veto
+regardless of what the returned field holds.
+
+**Upgrade note**: a `BEFORE READ` trigger with a non-trivial `SELECT` body that previously did nothing will now run
+its query on every read of the type. This is the fix, not a regression, but it is worth budgeting for if such a
+trigger exists in an upgraded database.
+
 [#6134](https://github.com/ArcadeData/arcadedb/issues/6134)
--- a/engine/src/main/java/com/arcadedb/schema/trigger/SQLTriggerExecutor.java
+++ b/engine/src/main/java/com/arcadedb/schema/trigger/SQLTriggerExecutor.java
@@ -22,9 +22,12 @@ import com.arcadedb.database.Database;
 import com.arcadedb.database.RID;
 import com.arcadedb.database.Record;
 import com.arcadedb.log.LogManager;
+import com.arcadedb.query.sql.executor.Result;
+import com.arcadedb.query.sql.executor.ResultSet;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 
 /**
@@ -53,10 +56,7 @@ public class SQLTriggerExecutor implements TriggerExecutor {
         params.put("$oldRecord", oldRecord);
       }
 
-      // Execute SQL with context parameters. Triggers run on every matching record; closing the
-      // ResultSet releases the per-call execution plan and avoids per-record state accumulation.
-      database.command("sql", sql, params).close();
-      return true;
+      return consume(database.command("sql", sql, params));
     } catch (final Exception e) {
       LogManager.instance().log(this, Level.SEVERE, "Error executing SQL trigger '%s': %s", e, triggerName, e.getMessage());
       throw new TriggerExecutionException("SQL trigger '" + triggerName + "' failed: " + e.getMessage(), e);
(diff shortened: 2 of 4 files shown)


lvca:
Good catch on the DML/\`RETURN\` interaction, verified it - \`UPDATE ... RETURN AFTER active\` does produce exactly the one-row/one-property shape the veto check keys on.

Fixed in ca9158a by gating the veto check on \`Statement.isIdempotent()\` (the same read/write classification \`SQLQueryEngine.query()\` already uses), so it only applies to bodies that are actually \`SELECT\`-shaped, never \`INSERT\`/\`UPDATE\`/\`DELETE\` regardless of \`RETURN\`. Added \`beforeCreateTriggerDmlReturnSingleFalseFieldDoesNotVeto\` to pin it - confirmed it fails without the gate and passes with it.

On the minor notes: simplified the single-row tracking as suggested, and noted the BEFORE READ SELECT-per-read performance implication is worth flagging to operators (not a code change, since running the body is the fix, not a bug on its own).

lvca:
Addressed the remaining points from this pass (8e31f4c):

- **\`\$score\`/\`\$similarity\` edge case**: fixed - \`consume()\` now excludes them from the single-property check via a small \`soleRealProperty()\` helper. I did dig into whether this is reachable today: \`ResultInternal.setScore()\`/\`setSimilarity()\` have zero callers anywhere in the codebase, and \`Projection.calculateSingle()\` (the method both a plain \`SELECT\` projection and an \`UPDATE ... RETURN\` projection go through) builds a fresh \`ResultInternal\` without ever copying score onto it - so the auto-injection this guards against isn't reachable through the current SQL surface. I kept the fix anyway since it's free (only evaluated once a candidate row is already in hand) and forward-looking, but didn't add a test for a path I can't currently exercise for real - happy to add one if a reachable path turns up later.
- **Style nit**: fixed, dropped the manual alignment spacing.
- **Release notes**: added an entry to \`docs/release-26.9.1.md\`, including the BEFORE READ upgrade note (a previously-inert \`SELECT\` body now runs per read).

\`TriggerSQLTest\`/\`BeforeReadTriggerTest\`: 37/37 pass.

lvca:
Addressed all three points from this pass (6d14c0b):

- Switched to \`((DatabaseInternal) database).getStatementCache().isIdempotent(sql)\`, avoiding the \`AnalyzedQuery\` wrapper allocation - matches the cast-to-\`DatabaseInternal\` pattern already used throughout the engine (\`BaseRecord\`, \`BinarySerializer\`, \`JsonSerializer\`, \`GraphBatch\`, etc.).
- Added \`beforeCreateTriggerMultiRowFalseDoesNotVeto\` and \`beforeCreateTriggerSingleRowMultiColumnFalseDoesNotVeto\` to pin the row/column boundaries. Verified both are real regression tests (they fail against a deliberately broken \`consume()\`/\`soleRealProperty()\` and pass against the fix).

\`TriggerSQLTest\`/\`BeforeReadTriggerTest\`: 39/39 pass.

lvca:
This round found no blocking issues - the only note (\`count\` as \`long\` vs \`int\`) was explicitly flagged by the reviewer as not worth a revision on its own, so I'm leaving it as is rather than triggering another review cycle for it.

Summary of the review loop: 4 rounds, 3 real issues found and fixed (DML \`RETURN\` false-positive veto, \`\$score\`/\`\$similarity\` property-count leak, and a couple of test/API-usage tightenings), plus release notes. Considering this ready to merge.

Issue #6154, opened 2026-08-13
The HEAD chunk of a spilled record is never fed into the page free-space statistics


lvca:
Noticed during the review of #6153 (fix for #6149), which touches the adjacent accounting but deliberately did not widen its scope to this.

## What is not accounted

`LocalBucket.writeMultiPageRecord` calls `updatePageStatistics` for every CONTINUATION chunk it places, but never for the HEAD chunk on the page the record spilled from:

```java
// WRITE THE 1ST CHUNK
byteWritten = currentPage.writeNumber(newPosition, FIRST_CHUNK);
...
chunkSize = availableSpaceForFirstChunk - byteWritten - INT_SERIALIZED_SIZE - LONG_SERIALIZED_SIZE;
currentPage.writeInt(newPosition, chunkSize);
currentPage.writeByteArray(...);            // <- no updatePageStatistics for this page
...
while (bufferSize > 0) {
  ...
  updatePageStatistics(nextPage.pageId.getPageNumber(), spaceAvailableInCurrentPage, -chunkSize);   // <- only here
}
```

The head chunk is the one that can consume the *most* space on the page it lands on: when the spilling record is the LAST record of its page, `availableSpaceForFirstChunk` is its own footprint **plus the entire free tail**, so a spill can take a page from kilobytes-free to zero-free without `freeSpaceInPages` hearing about it.

`updateMultiPageRecord` has the same shape: it accounts the chunks it relocates onto other pages, not the head chunk it rewrites in place.

## Why it is only a hint, and why it is still worth fixing

`freeSpaceInPages` is a placement hint, not a correctness input: `findAvailableSpace` re-reads the page through `getAvailableSpaceInPage`/`getFreeSpaceInPage` and re-verifies before using a slot, so a stale entry costs a wasted candidate rather than a corrupt write. That is why this has gone unnoticed.

The cost is a wasted page probe per stale entry, and `gatherPageStatistics` only re-derives the map on its own cadence (`MAX_TIMEOUT_GATHER_STATS`, and only for pages below `txPageCount - 2`), so an entry for a page a spill has filled can stay wrong for a while. On a bulk load that spills a lot, the allocator keeps being pointed at pages that have nothing left.

## Suggested fix

Account the head chunk where it is written, in both `writeMultiPageRecord` and `updateMultiPageRecord`, the same way #6153 accounts the bytes the in-page shift claims:

```java
[...]

PR #6170, opened 2026-08-14
fix(#6163,#6154): size a chunked record's head chunk from the room its page gives it, and account that room

### LINKED ISSUE #6163: A chunked record's head chunk ratchets down to the smallest size it has ever had, and never recovers
Found while building the page-sealing fixture for #6149 (PR #6153): a fixture that sealed a page by growing its last record past the page size, then shrank that record back, silently un-sealed the page. Chasing why turned up a permanent, one-way ratchet.

## The head chunk's declared size only ever goes DOWN

`LocalBucket.updateMultiPageRecord` sizes the head chunk from what the page already says, and can only cap it downwards:

```java
int chunkSize = currentPage.readInt(newPosition);   // what the head chunk declares TODAY
int bufferSize = buffer.size();

// WRITE THE 1ST CHUNK
if (bufferSize < chunkSize)
  // 1ST AND LAST CHUNK
  chunkSize = bufferSize;
...
currentPage.writeInt(newPosition, chunkSize);       // written back, permanently
```

There is no path that ever raises it again: the next update reads the already-shrunk value and can only cap it further. So a chunked record's head chunk ratchets down to **the smallest size that record has ever had**, and stays there for the rest of its life.

The room is not reclaimed by the record, either. The physical gap between the shrunk end and where the head chunk used to end stays inside the record's slot region, unusable by the record (it cannot grow its declared size back) and, unless the head chunk is the page's last record, unusable by anyone else.

## Why it matters

**A record that oscillates in size degrades permanently.** Take a 200 KB chunked record whose head chunk holds ~9 KB. One update drops it to 100 bytes; the head chunk is now 100 bytes forever. Every later update back up to 200 KB puts ~9 KB *less* on the head page and pushes it into continuation chunks on other pages - more chunk headers, more pages touched per read and per write, for the whole remaining life of the record. Nothing recovers it short of deleting and re-creating the record.

**It also makes a page's free tail move in a direction nothing else does.** `getPageOccupiedInBytes` derives a chunk record's footprint from the declared chunk size, so when the head chunk is the page's LAST record, shrinking it hands the difference back as free tail that the allocator can give to other records. That is the mechanism that un-sealed the fixture, and it is the only way a page's occupied byte count goes *down* without a delete.

## Not a correctness bug, and specifically not one for the slot merge

I checked the disjoint-slot merge (#6129) against this before filing:

- `SLOT_KIND_FIRST_CHUNK` replay refuses when `body.length > committedChunk.length`, and a shrink makes `body` shorter, so the replay writes strictly inside the committed footprint. No overlap.
- A concurrent transaction cannot allocate into the tail a shrink is about to release, because it sizes the page from the *committed* image, which still has the larger head chunk. The release is only visible after the shrink commits.

So this is a space/performance issue, not a race.

## Worth fixing, and the constraint on how

The natural fix - let the head chunk grow back into the room it already owns - has to know where its region *ends*, which the page does not record: the head chunk declares its content length, not its allocated footprint. The end is only recoverable by looking at where the next record starts (or at the page's content end when it is the last record), which is exactly what `getPageOccupiedInBytes`/`getLastRecordPositionInPage` already compute for other purposes.

Two shapes, both worth weighing:

1. **Let `updateMultiPageRecord` grow the head chunk back** up to the next record's offset (or the page's content end when last). Recovers the ratchet for free on the next update, and keeps the "head chunk keeps its slot" property the merge relies on. Needs care: growing the head chunk past what the merge's pre-image described is the one thing `rebaseRecordOnPage` currently refuses outright (`a chunk head never grows`), so that refusal would have to become a re-derivation rather than a flat no.
2. **Do not shrink the declared size at all** - keep the head chunk's footprint truly fixed and let the unused bytes sit as slack inside it. Simpler, matches what `engine/CLAUDE.md` already claims, and removes the odd "page occupancy went down without a delete" behaviour. Costs the space permanently rather than temporarily.

(1) is better if the merge change is done properly; (2) is the honest minimum if it is not.

## Documentation

`engine/CLAUDE.md` currently states "The head chunk's footprint on the page is fixed when the record spills and is never grown again". The "never grown" half is right; the sentence reads as though the footprint is immutable, which is what made this take a while to spot. It should say the declared size can SHRINK, one-way, and that the shrink is what releases page tail.

lvca:
Fixes #6163 and #6154. They are the same missing idea seen from two sides - the room a HEAD chunk may use is a property of the PAGE, not of what the chunk declared last time - so they are fixed together.

## What was wrong

**#6163.** `LocalBucket.updateMultiPageRecord` sized the head chunk from what the chunk header already declared, and could only cap that downwards:

```java
int chunkSize = currentPage.readInt(newPosition);   // what the head chunk declares TODAY
if (bufferSize < chunkSize)
  chunkSize = bufferSize;
currentPage.writeInt(newPosition, chunkSize);       // written back, permanently
```

No path ever raised it again, so a chunked record's head chunk ratcheted down to the smallest size that record had ever had. A 200 KB record whose head chunk held ~9 KB, updated once to 100 bytes, put ~9 KB *less* on its own page on every later update for the rest of its life - more chunk headers, more pages touched per read and per write, nothing short of delete-and-recreate to recover it. The pathological end of it: a chunked record that had shrunk to a few bytes allocated a continuation chunk on **another page** for a value of 4 KB that its own slot had ample room for.

**#6154.** The head chunk is the biggest single bite anything takes out of a bucket page - a record spilling as the LAST record of its page is given its own footprint **plus the whole free tail**, and a record being created gets the tail outright - and it was the one chunk never fed into `updatePageStatistics`. `freeSpaceInPages` went on offering the allocator a page a spill had sealed.

## What changed

`headChunkRegionEnd` returns the region a head chunk's slot owns on its page: from its offset to the next record's offset, or to the page's maximum content size when it is the page's last record. That is the same room `writeMultiPageRecord` is given when a record spills, so an update now places its head chunk **exactly as a fresh spill of the new content would** - `chunkSize = min(bufferSize, headChunkRoom)`, moving in both directions inside the region. There is no "grow back" special case in the code; the remembered size is simply gone.

The region is derived rather than stored because it is not the record's to keep: a chunk header declares its CONTENT length, not its footprint, and the page hands the unused bytes away meanwhile (`compressPage` closes the hole at the end of every transaction that touches the page; an insert takes the tail a shrink released). Reading it back from the page is what bounds the grow-back to what is really still there.

`rebaseRecordOnPage` (the disjoint-slot merge, #5279/#6129) refused any head chunk longer than the committed one outright. It now re-derives the region on the COMMITTED page - as `growRecordInPage` already re-derives the free tail for a plain record that grew - and refuses only when a concurrent commit has taken those bytes, which sends the transaction to a normal retry instead of writing over that record.

For #6154, both `writeMultiPageRecord` and `updateMultiPageRecord` now account the head chunk, netted against the footprint the slot already owned and against the bytes the #6149 shift claimed, so a sealed page lands on exactly 0 and drops its entry - and a shrink hands the tail back.

## Why this, rather than the two shapes the issue proposed

- **(1) "let `updateMultiPageRecord` grow the head chunk back"** is what this generalises. Recovering the ratchet is a smaller claim than what the code now does: the head chunk is sized from the page on *every* update, so a spill and an update produce the same layout by construction, rather than the update chasing the spill's decision after the fact. It also drops a branch instead of adding one.
- **(2) "do not shrink the declared size at all"** is not implementable without a page-format change: that `int` **is** the content length the chain reassembly sums, so declaring more than the chunk holds would corrupt every record's length. Keeping a fixed footprint means storing it - a new header field, a format break - for no benefit over deriving it from the page, which costs one slot-table scan on a path that already does one for growth.

## Two claims in #6163 that measurement did not support

- *"the room stays inside the record's slot region, unusable by anyone else"* - it does not. `compressPage` runs on every modified page at commit (`TransactionContext.commit1stPhase`), and it closes exactly that hole, moving the following records left. The bytes go back to the page's free tail in the same commit.
- *"it is the only way a page's occupied byte count goes down without a delete"* - an in-place shrink of a plain LAST record does the same thing: `getPageOccupiedInBytes` reads its new, smaller size. What is chunk-specific is not the release, it is that a plain record can grow again (`growRecordInPage`) and a head chunk could not. That asymmetry is what this fixes.

## Tests

`engine/src/test/java/com/arcadedb/engine/Issue6163HeadChunkRoomTest.java`, 6 tests on `BucketPageLayoutTestSupport`'s sealed-page fixture:

- the ratchet itself - shrink, then grow back to a value the head chunk's own room holds: nothing may leave the page (**red before**: 1 continuation chunk);
- repeated shrink/grow cycles do not drift;
- the grow-back stops at a record that took the room the shrink released (**mutation-checked**: removing the region bound destroys that record - `RecordNotFoundException`);
- #6154, asserted on the free-space hint itself, in both directions - the allocator re-reads every candidate page, so a test driven through it would pass either way (**red before**: the sealed page still advertised 8098 free bytes);
- the merge replays a grow-back on a newer committed page, with the premise proved by re-running it with `TX_PAGE_SLOT_MERGE` off and requiring failure (**red before**);
- the merge REFUSES when the committed page no longer has the room (**mutation-checked**: dropping the check overwrites the concurrent insert).

Full engine unit lane green: 11906 tests, 0 failures. Targeted regression set green: `Issue6129ChunkedSlotMergeTest`, `Issue6149PlaceholderPrefersChunksTest`, `Issue5279Concurrent*`, `Issue5381FalseConflictTest`, `Issue5596MergeCoverageTest`, `Issue5608RebasedPageCompressionTest`, `RandomDeleteTest`.

`engine/CLAUDE.md` updated: the "footprint is fixed when the record spills and is never grown again" sentence is replaced by the region and the two directions it moves in.

## Deliberately not in scope

Converting a chunk chain back into a plain record when the whole content fits the slot again ("un-spilling"). It changes the slot's shape and thus what the merge tracks, and the single-chunk chain this produces already costs only its 13-byte header.


lvca:
Review round 1 addressed:

- **O(n) slot-table scan in `headChunkRegionEnd`** - kept, with the reason now written into its javadoc (8a808af): it is the same scan `getLastRecordPositionInPage` already costs the growth path on every update that does not fit in place, in front of a write that moves kilobytes of chunk content - and there is nothing to cache, because the answer *is* the page state a concurrent commit is allowed to change.
- **"stray empty `//` line" in `createRecordInternal`** - not a stray: a bare `//` between two comment paragraphs is this file's own idiom (4 occurrences on `main` before this PR, e.g. in the chunk-head branch of `updateRecordInternal`). Left as is for consistency.
- **CI/test confirmation** - the full engine unit lane (`-DexcludedGroups=slow,benchmark,vector`) is green locally: 11906 tests, 0 failures, 0 errors. The `slow` lane is running and green so far, including `RandomTestMultiThreadsTest` and `CheckDatabaseTest`.

lvca:
Local `slow` lane finished too: **232 tests, 0 failures, BUILD SUCCESS** (`-Dgroups=slow -DexcludedGroups=vector`), including `RecordRecyclingTest`, `RandomDeleteTest` and `DatabaseCheckerIndexRebuildLockContentionTest` - the ones most exposed to a change in bucket page layout. Together with the unit lane that is 12138 local tests, all green.

lvca:
Review round 2 addressed (6537cb6):

**"the invariant is only an `assert`, disabled in production"** - valid question, and the answer is that the two ways the region can disagree with the chunk it holds are not the same failure. They are now answered separately, with the fatal one first so it is the one that speaks when both are true:

- a region merely **smaller** than the declared size cannot corrupt anything from there - the head chunk is written shorter and the content that no longer fits leaves through the continuation loop, which is the ordinary path - so it stays an `assert`: loud in the tests, harmless in production;
- a region that cannot host **a single byte** is where the arithmetic stops being safe (a chunk size of 0 or below is not a shorter chunk, it is a header no reader can walk), so it now throws a `DatabaseOperationException`, in the same terms `writeMultiPageRecord` already refuses an invalid continuation chunk.

New test `aRegionTooSmallForAnyContentIsRefusedRatherThanWritten` pins it. No public API sequence produces that page - the region always contains at least the head chunk's own footprint - so it points a neighbour's slot inside the chunk header by hand, in the SAME transaction as the update, which keeps the commit-time `compressPage` from re-flowing the overlap first. Mutation-checked: with the guard removed the write proceeds instead.

Ordering matters and is deliberate: surefire runs with `-ea`, so with the assert first the corrupted fixture died on the assert and never reached the guard.

Regression set re-run green after the change (`Issue6163`, `Issue6149`, `Issue6129`, `Issue5279Concurrent{Update,Insert}`, `RandomDeleteTest`): 37 tests, 0 failures.

lvca:
**CI note: the `integration-tests` failure is pre-existing on `main`, not from this PR.**

```
com.arcadedb.server.Issue5470BatchStreamStallIT.aBodyShorterThanItsContentLengthIsNotReportedAsSuccess
[a truncated upload must never be answered with 200]
Expecting actual: "HTTP/1.1 409 Conflict" to contain: "408"
```

The identical assertion, on the identical line (`Issue5470BatchStreamStallIT.java:226`), fails on `main` in today's runs [31776078945](https://github.com/ArcadeData/arcadedb/actions/runs/31776078945) and [31776042204](https://github.com/ArcadeData/arcadedb/actions/runs/31776042204) - both before this branch existed.

It also cannot involve this change: the server log for the failing run says the batch was interrupted **after 0 vertices and 0 edges** (the request body was never read to the end), so no record was written at all, let alone a multi-page one. This PR only changes the head-chunk paths, which a batch of small vertex records never reaches.

Everything else on the commit is green: `unit-tests`, `slow-unit-tests`, `vector-unit-tests`, `opencypher-tck-tests`, `studio-e2e-tests`, the language e2e lanes, `builder-tests`, CodeQL, Codacy and Meterian. Locally: engine unit lane 11907 tests green, engine slow lane 232 tests green, `integration` module 132 tests green.

lvca:
**CI note 2: the `unit-tests` failure is a wall-clock flake, unrelated to this change.**

Single failure in the whole lane:

```
com.arcadedb.schema.MaterializedViewEdgeCaseTest.periodicRefreshUpdatesView
expected: 2L but was: 1L   (MaterializedViewEdgeCaseTest.java:252)
```

The test creates a PERIODIC materialized view with a 200 ms refresh interval, inserts a row, `Thread.sleep(1_000)`, and asserts the scheduler has refreshed by then. On a loaded runner that window is missed - the view simply had not refreshed yet. It exercises the view scheduler re-running a `SELECT`; it never touches a multi-page record, which is the only thing this PR changes.

Ran the class 3x locally on this branch: 12 tests green each time. It was also green inside the full engine unit lane (11907 tests) run on this exact commit.

So on `6537cb6` the two red lanes are (1) `integration-tests` - `Issue5470BatchStreamStallIT`, already failing on `main` today with the identical assertion, and (2) this timing flake. Everything else is green.

lvca:
**CI final state on `6537cb6`: 14 of 16 jobs green, including `ha-integration-tests`.**

The two red ones are both triaged above and neither involves this change:
- `integration-tests` - `Issue5470BatchStreamStallIT`, failing on `main` today with the identical assertion (409 vs 408), and the batch never wrote a record ("0 vertices and 0 edges");
- `unit-tests` - `MaterializedViewEdgeCaseTest.periodicRefreshUpdatesView`, a 200 ms periodic-refresh window missed under load; 3/3 green locally on this branch and green inside the full 11907-test lane on this same commit.

Three review rounds: rounds 1 and 2 both "No blocking issues found", round 3 clean. Ready for your look.

Issue #6168, opened 2026-08-14
Follow-ups from #6152: memoize the unreferenced-file count, reclaim unreferenced files, two flaky tests, one naming convention with two homes, and a chronically red lane


lvca:
Six follow-ups left by PR #6152 (squash `793163c6d`), which closed #6142, #6143 and #6144. They are independent of
each other and are filed together only because they came out of the same work; each is self-contained and can be
taken on its own.

Items 1 and 2 are about the code that shipped. Items 3, 4 and 6 are about CI, and are the reason that PR needed
four CI cycles to go green over changes that were already reviewed seven times. Item 5 is a design observation that
CI made for us.

---

## 1. Memoize the unreferenced-file count behind a modification/version gate

`UnreferencedFiles.count()` feeds the per-node `arcadedb.ha.schema.unreferenced_files` gauge, which
`HAReplicationMetrics` refreshes **every 5 seconds for every open replicated database**. Each refresh rebuilds the
whole claimed-file set: every type, every bucket, and one `getFileIds()` per index, which takes that index's read
lock and allocates a list.

It is in-memory with no I/O, so it is cheap on ordinary schemas, and it was left unmemoized deliberately: a cached
count that goes stale is a worse diagnostic than a slightly expensive one. But the cost scales with SCHEMA SIZE
rather than with the size of the problem being diagnosed, on every node, forever, whether or not anything is wrong.
On a server hosting many databases with tens of thousands of components it is the one part of that change that could
show up in a profile.

The gate is already worked out and recorded in `count()`'s javadoc:
`(FileManager.getModificationCount(), schema version)`. Between them those two cover every way the answer can
change, since a file the walk would newly report appeared either as a file (the first) or as a schema change that
stopped claiming one (the second). What is missing is somewhere to keep the cached pair per database, and
`RaftHAServer.forEachOpenReplicatedDatabase` is the natural place since it already enumerates them.

Raised in three separate review rounds on #6152, deferred each time. Filed so the decision is revisited on evidence
rather than re-argued.

## 2. The actual fix for #6143: reclaiming unreferenced files

#6143 shipped a **diagnostic** - `CHECK DATABASE` reports `unreferencedFiles`, and every node publishes its own
count - because the issue's own conclusion was that the real fix is unreferenced-file garbage collection, which this
database does not have. That is still true. An operator who reads the SEVERE line, or the gauge, still has to stop
the node and delete files by hand.

[...]

PR #6171, opened 2026-08-14
feat(#6091): the cluster can name a peer's gRPC address, so a follower's refusal is dialable

### LINKED ISSUE #6091: graphBatchLoad's follower refusal cannot name a dialable address: HAServerPlugin exposes no gRPC routing table
Carried over from #6083 item 3, the one item of that batch not addressed by #6089. It was split out because it changes a public configuration surface and needs a real multi-node test, neither of which belonged in a defect-fix PR.

## Problem

`ArcadeDbGrpcService.graphBatchLoad` refuses a bulk load on a follower of a replicated database — correctly, since the bulk path mutates shared state only the leader can serialize (#4122). To let the caller redirect, it names the leader. But `HAServerPlugin` exposes only the leader's **HTTP** address (`getLeaderAddress()`), so the refusal message currently has to say:

> `Reconnect to the leader at '<host:httpPort>' (HTTP address; use its gRPC port) and retry`

A client that wants to retry automatically must already know the deployment's port-mapping convention. There is no way for it to learn the leader's gRPC address from the cluster.

## Precedent to follow

`getBoltRoutingTable()` is exactly the shape needed: a protocol-specific routing view (`BoltRoutingTable(writer, readers)`) derived from a **single** `getLeaderId()` read, so a concurrent leader change cannot make the writer and reader sets mutually inconsistent. `RaftHAServer.resolveBoltAddress` backs it, reading a per-peer `bolt:` field from the object form of `HA_SERVER_LIST` and falling back to "peer's Raft host + this node's local Bolt port" with a one-time WARNING when the field is absent.

A `getGrpcRoutingTable()` alongside it — or a generalisation of the existing record over a protocol enum, which may be the better shape now that there would be two — would let the refusal name an address the caller can dial.

## Scope

1. `HAServerPlugin`: new default method returning null when HA is inactive (same contract as `getBoltRoutingTable()`).
2. `RaftHAServer` / `RaftHAPlugin`: `resolveGrpcAddress` mirroring `resolveBoltAddress`, plus a `grpcAddresses` map on `ParsedPeerList`.
3. `RaftPeerAddressResolver`: a `grpc:` field in the object form (`host:{raft:2434,http:2480,bolt:7687,grpc:50051}`). Object form only, as with `bolt` — the positional colon form has no room left.
4. `ArcadeDbGrpcService.graphBatchLoad`: use it in the refusal, keeping the current HTTP wording as the fallback when no gRPC address resolves.

## Decisions worth making deliberately

- **Generalise or duplicate?** A second single-protocol routing record invites a third. Worth deciding now whether `BoltRoutingTable` becomes `RoutingTable(protocol, writer, readers)`, which touches the Bolt `ROUTE` path.
- **Derive-only vs. explicit config.** The derive-from-local-port fallback is correct for homogeneous deployments (a K8s StatefulSet) and wrong for heterogeneous ones, which is why `bolt` is declarable. Shipping the derive-only half first would avoid the config change but leaves the heterogeneous case silently wrong — the same reason `bolt:` exists.
- Whether other gRPC RPCs that refuse on a follower should use the same table, rather than only `graphBatchLoad`.

## Testing note, recorded in #6083

The auth-before-leader-check ordering here is **unobservable on a single server**: the channel-level `GrpcAuthInterceptor` short-circuits a credential-less call, and forcing the follower branch on a single node NPEs on a null `ha`. This needs a real follower in a cluster, so it lands in the `ha-integration-tests` lane.

lvca:
Closes #6091 (item 3 of #6083, the one left out of #6089).

## The problem

`graphBatchLoad` refuses a bulk load on a follower - correctly, since the bulk path mutates shared state only the leader can serialize (#4122). To let the caller redirect it named the leader, but the only address HA could expose was the leader's **HTTP** one, so the refusal had to say:

> `Reconnect to the leader at '<host:httpPort>' (HTTP address; use its gRPC port) and retry`

A client retrying automatically had to already know the deployment's port-mapping convention, and there was no way to learn the leader's gRPC address from the cluster.

## Decisions the issue asked to make deliberately

**Generalise, don't duplicate.** `BoltRoutingTable` / `getBoltRoutingTable()` become `RoutingTable(protocol, writer, readers)` / `getRoutingTable(ROUTING_PROTOCOL)`. A second single-protocol record would have invited a third, and the resolver behind it is the same code with different inputs: the declared-addresses map, this node's local port, and the one-time-warning latch are now selected per protocol, everything else is shared. The Bolt ROUTE path was updated with it; `Bolt5002RoutingTableIT` passes unchanged.

**Derive *and* declare, exactly as Bolt does.** The object form of `arcadedb.ha.serverList` takes a `grpc:` field (`host:{raft:2434,http:2480,bolt:7687,grpc:50051}`); with nothing declared, a peer's endpoint is its Raft host plus *this* node's gRPC port. That is right for a homogeneous deployment (a K8s StatefulSet) and knowingly wrong for a heterogeneous one, which is why the field exists and why the fallback logs a WARNING naming the syntax. Shipping the derive-only half would have left the heterogeneous case silently wrong - the same reason `bolt:` exists. The warning latch is per protocol now, so a derived Bolt address no longer mutes the gRPC warning.

**One RPC refuses today, but the refusal is not written for one RPC.** `notTheLeader(ha, rpc, why)` builds it, so the next RPC that has to refuse on a follower gets the same answer rather than a second dialect.

## Better than what the issue proposed

The issue asked for the address in the refusal *message*. A message is not a contract - a client redirecting itself would be pattern-matching prose. So the refusal also carries:

- `arcadedb-leader-grpc-address` - the address the refused call can be retried on;
- `arcadedb-leader-http-address` - always known when a leader is, kept as the diagnostic fallback;
- `arcadedb-exception-class: ServerIsNotTheLeaderException` - the trailer this service already uses to preserve exception types.

The gRPC client therefore rebuilds **the same exception the HTTP protocol raises for the same situation**, `ServerIsNotTheLeaderException`, with `getLeaderAddress()` returning the gRPC endpoint. Callers that already handle the HTTP leader redirect handle the gRPC one with no new code. The keys live in the shared `grpc` protocol module (`LeaderRedirectProtocol`), next to `GraphBatchProtocol`, so the two ends cannot drift.

`GlobalConfiguration.GRPC_PORT` registers the key the gRPC plugin previously read as a bare string with its own literal default. HA resolves peer endpoints from the same constant, so the two cannot disagree about which port to advertise - and the setting now honours env/system properties like every other one.

## Testing

Cluster tests, because the follower branch is unreachable on a single server (recorded in #6083: the auth interceptor short-circuits a credential-less call and forcing the branch NPEs on a null `ha`).

- **`Issue6091GraphBatchLoadLeaderRedirectIT`** (grpcw, 3 nodes, per-node `grpc:` declared) - the load refused on a follower is retried **against the address the refusal advertised, dialled verbatim**, and succeeds and replicates. A test comparing the string against the port it configured would pass just as well if the server had named the follower's own port, which is exactly the mistake to catch here.
- **`Issue6091GrpcRoutingTableIT`** (ha-raft) - every node names the same leader endpoint; the Bolt and gRPC tables disagree on a cluster that declares only gRPC ports (a shared resolver's one new failure mode is reading the other protocol's map); the writer is never also a reader.
- **`Issue6091GrpcRoutingTableDerivedIT`** (ha-raft) - the fallback, asserted against a deliberately non-default port so it fails if the resolver stops reading `GRPC_PORT`.
- **`Issue6091LeaderRedirectMappingTest`** (grpc-client) - gRPC address preferred over HTTP, HTTP as fallback, null address mid-election, and an address trailer alone does not invent a type.
- `RaftHAServerAddressParsingTest` - `grpc:` parsing, independence from `bolt:`, malformed port, K8s DNS suffix.

**Mutation-checked, all four ways:** naming only the HTTP address, pointing the gRPC lookup at the Bolt map, deriving from `BOLT_PORT`, and each fails the test that covers it.

Suites run green: `bolt` unit + `Bolt5002RoutingTableIT`, `ha-raft` unit + the new ITs, `grpcw` unit + `Issue6070*IT`, `grpc-client` unit. (Some local runs of unrelated classes fail on `HTTP port 2480 not available` - another process on this machine holds it.)

## Not in this PR

The user-facing `serverList` reference lives in the docs repo; the `grpc:` field wants the same one-paragraph note `bolt:` got there.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -1048,7 +1048,8 @@ public class BoltNetworkExecutor extends Thread {
     final List<Map<String, Object>> servers = new ArrayList<>();
 
     final HAServerPlugin ha = server.getHA();
-    final HAServerPlugin.BoltRoutingTable table = ha != null ? ha.getBoltRoutingTable() : null;
+    final HAServerPlugin.RoutingTable table = ha != null ?
+        ha.getRoutingTable(HAServerPlugin.ROUTING_PROTOCOL.BOLT) : null;
 
     if (table != null) {
       // HA cluster with a known leader: the leader is the writer and a router, followers are readers and
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1836,6 +1836,13 @@ public enum GlobalConfiguration {
       Default is 1000, generous for any real BOLT message.""",
       Integer.class, 1000),
 
+  // gRPC
+  GRPC_PORT("arcadedb.grpc.port", SCOPE.SERVER, """
+      TCP/IP port number used for incoming connections for the gRPC plugin. Registered here, rather than read as a \
+      bare key by the plugin alone, because HA has to know the port a peer's gRPC endpoint listens on to advertise a \
+      dialable address for it (see the 'grpc' field of arcadedb.ha.serverList). Default is 50051""",
+      Integer.class, 50051),
+
   // REDIS
   REDIS_PORT("arcadedb.redis.port", SCOPE.SERVER,
       "TCP/IP port number used for incoming connections for Redis plugin. Default is 6379", Integer.class, 6379),
--- a/grpc-client/src/main/java/com/arcadedb/remote/grpc/GrpcClientErrorMapper.java
+++ b/grpc-client/src/main/java/com/arcadedb/remote/grpc/GrpcClientErrorMapper.java
@@ -23,7 +23,9 @@ import com.arcadedb.exception.DuplicatedKeyException;
 import com.arcadedb.exception.NeedRetryException;
 import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.exception.TimeoutException;
+import com.arcadedb.network.binary.ServerIsNotTheLeaderException;
 import com.arcadedb.remote.RemoteException;
+import com.arcadedb.server.grpc.LeaderRedirectProtocol;
 import io.grpc.Metadata;
 import io.grpc.Status;
 
@@ -90,9 +92,28 @@ final class GrpcClientErrorMapper {
       case "com.arcadedb.exception.RecordNotFoundException" -> new RecordNotFoundException(msg, null);
       case "com.arcadedb.exception.TimeoutException" -> new TimeoutException(msg);
       case "java.lang.SecurityException" -> new SecurityException(msg);
+      case "com.arcadedb.network.binary.ServerIsNotTheLeaderException" ->
+          new ServerIsNotTheLeaderException(msg, leaderAddress(trailers));
       // Unknown class: let the caller fall back to status-code mapping.
       default -> null;
     };
+  }
+
+  /**
+   * Returns the address to redirect to when a follower refused an RPC that only the leader may run: the
+   * leader's gRPC address when the cluster could resolve one, otherwise its HTTP address. Both may be absent -
+   * a refusal issued during an election names no leader at all - and the exception then carries a null address,
+   * which is what {@code ServerIsNotTheLeaderException} means by "retry, destination unknown".
+   * <p>
+   * The gRPC address is preferred because it is the only one of the two this client can actually dial. The HTTP
+   * address is kept as a fallback because it is always known when a leader is, and a caller that can map an
+   * HTTP endpoint to its gRPC port is better off than one told nothing.
+   */
+  private static String leaderAddress(final Metadata trailers) {
+    if (trailers == null)
+      return null;
+    final String grpcAddress = trailers.get(LeaderRedirectProtocol.LEADER_GRPC_ADDRESS);
+    return grpcAddress != null ? grpcAddress : trailers.get(LeaderRedirectProtocol.LEADER_HTTP_ADDRESS);
   }
 
   /**
--- a/grpc-client/src/test/java/com/arcadedb/remote/grpc/Issue6091LeaderRedirectMappingTest.java
+++ b/grpc-client/src/test/java/com/arcadedb/remote/grpc/Issue6091LeaderRedirectMappingTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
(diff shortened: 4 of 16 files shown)


lvca:
Codacy's one finding (Info, `PMD ClassNamingConventions`): the enum name `ROUTING_PROTOCOL` does not match `[A-Z][a-zA-Z0-9]*`.

Declining it deliberately: the four enums already on `HAServerPlugin` - `QUORUM`, `ELECTION_STATUS`, `SERVER_ROLE`, `READINESS_SIGNAL` - are all SCREAMING_SNAKE and trip the same rule; they escape it only because the analyzer looks at the diff. Renaming just the new one would make it the odd member of the set it is declared beside. Happy to rename all five in a separate mechanical PR if the convention should change.

lvca:
### CI status

Everything this PR adds is green. The two red lanes are failures this branch did not introduce:

**`integration-tests`** - two failures, neither on a path this PR touches:

- `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose:351` - the identical failure, same line, on PR #6162 (a Cypher-only change, since merged): [job 94572497347](https://github.com/ArcadeData/arcadedb/actions/runs/31736813745/job/94572497347). The test's own javadoc documents the race it rides on: whether the server's async failure is observed by a `send()` in the body or only at `close()` depends on grpc-java reporting `isReady() == false` on a terminated call before the next of 400 sends.
- `Issue5470BatchStreamStallIT.aBodyShorterThanItsContentLengthIsNotReportedAsSuccess:226` - expected `408`, got `409`, and the server log for that run says the batch was `interrupted after 0 vertices and 0 edges` on a 10 s streaming read timeout. The runner did not deliver a 100-record body within ten seconds; the HTTP batch path is untouched here (the only `server` change in this PR is the `HAServerPlugin` interface).

**`ha-integration-tests`** - `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact:121`. This lane is red on every recent PR with a different test each time (#6165, #6161, #6162 all failed it; #6165's was `RaftPriorityRejoinIT`), which is the chronic flake tracked in #5668 / #5702.

**What this PR added, in the lanes that run it:**

| Test | Lane | Result |
|---|---|---|
| `Issue6091GrpcRoutingTableIT` (3 tests) | ha-integration | pass, 36.9 s |
| `Issue6091GrpcRoutingTableDerivedIT` | ha-integration | pass, 11.6 s |
| `Issue6091GraphBatchLoadLeaderRedirectIT` | integration | pass, 15.7 s |
| `Issue6091LeaderRedirectMappingTest`, `RaftHAServerAddressParsingTest` | unit | pass |

Re-running the two failed jobs to confirm.

lvca:
Re-run result for `integration-tests` ([job 94855534608](https://github.com/ArcadeData/arcadedb/actions/runs/31822320510/job/94855534608)):

- `Issue5470BatchStreamStallIT` **passed** - confirming the first failure was the slow-runner artifact described above (the server had read 0 of 100 records when its 10 s streaming timeout fired).
- `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose` failed again, identically. It is already filed as **item 3 of #6168**, independently of this PR, and it failed the same way on merged PR #6162. Nothing in this branch touches the readiness/termination path it races on.

So the only outstanding red in this lane is a known, separately tracked flake.

lvca:
Final CI state after re-running the two red lanes:

| Lane | Result |
|---|---|
| `ha-integration-tests` | **pass** (59m59s) - the earlier `Issue5569SlotMergeDeleteRaftIT` failure was the chronic flake this lane shows on every PR |
| `integration-tests` | fail, and only on `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose` - **#6168 item 3**, also red on merged PR #6162. `Issue5470BatchStreamStallIT` passed on the re-run |
| everything else | pass: unit, slow-unit, vector-unit, build-and-package, builder, opencypher-tck, studio-e2e, java/js/python/go/csharp e2e, CodeQL, Meterian, claude-review |

Codacy's single Info-level finding is the enum-naming one addressed above.

PR #6173, opened 2026-08-14
fix(#6168): memoize the unreferenced-file count, give "reached by name" one home, de-flake two sampled/timing-dependent tests

lvca:
Closes four of the six follow-ups in #6168. **Item 2 (reclaiming unreferenced files) is deliberately not in this PR** - it is a file deletion driven by inference on a replica, on a code path whose history (#4083, #4743, #5443, #5492) is silent divergence rather than exceptions, for an impact that is only wasted disk. The issue itself says it wants its own change and its own risk assessment, so #6168 stays open for it.

---

## 1. Memoize the unreferenced-file count behind a modification/version gate

`UnreferencedFiles.count()` feeds the per-node `arcadedb.ha.schema.unreferenced_files` gauge, which `HAReplicationMetrics` refreshes **every 5 seconds for every open replicated database**, rebuilding the whole claimed-file set each time: every type, every bucket, one `getFileIds()` per index taking that index's read lock and allocating a list. The cost scaled with SCHEMA SIZE rather than with the size of the problem being diagnosed.

New `UnreferencedFiles.MemoizedCount` implements the gate the javadoc had already worked out, `(FileManager.getModificationCount(), schema version)`. Between them those two cover every way the answer can change: a file this walk would newly report either appeared as a FILE (bumps the modification count, on registration and on drop) or appeared because a SCHEMA change stopped claiming one (bumps the schema version). Neither can move without invalidating the entry, which is the only reason memoizing a diagnostic is acceptable at all.

Three details worth review:

- **Equality on both halves, never "not newer".** The schema version is not monotonic on a follower - a schema reload assigns the leader's value outright.
- **Lock-free, and the ordering is what makes it safe.** The gate is read BEFORE the walk and published with the count afterwards, so a change landing mid-walk tags the result with the gate it was computed under and the next call recomputes. Racing callers can duplicate work or overwrite each other's entry; both outcomes are a recompute, never a stale answer.
- **Where it lives.** One instance per database, on `RaftReplicatedDatabase`, rather than a map on `RaftHAServer` - the gauge is refreshed per open database, so a per-instance holder needs no keying and no eviction and is collected with the database it describes.

## 5. "Reached by name" now has one home

New `com.arcadedb.schema.InternalBucketNaming`. A bucket can be referenced by the engine without any type listing it, through a naming convention, and the knowledge was in six places: composed in `GraphEngine` (`_out_edges`/`_in_edges`), in `StripedEdgeList` (`_sn_stripe_`) and twice as a bare `"_ext"` literal in `LocalDocumentType`/`LocalSchema`, and recognised independently in `UnreferencedFiles.isDerivedFromAnOwner` and `GraphDatabaseChecker`.

**The two halves answer different questions and only one of them is a registry:**

- `Convention` + the `composeXXX` helpers are the DECLARATION side. A feature that names buckets after their owner declares its convention here and gets recognition with it, instead of leaving it to be discovered by whoever debugs the false positive later.
- `ownerOf` is the RECOGNITION side and is deliberately **not** driven by the enum. It matches the shape `<owner>_<anything>`, so it covers conventions that were never declared - including ones not written yet. Enumerating them is not a closed problem: that is exactly what CI proved on #6152, when the classifier's explicit list turned out to be missing the super-node stripe pools of #5156 and 17 test classes went red through `TestHelper.checkDatabaseIntegrity`.

`GraphEngine.OUT_EDGES_SUFFIX`/`IN_EDGES_SUFFIX` and `StripedEdgeList.STRIPE_BUCKET_INFIX` are removed rather than left as second copies; every consumer goes through the new class.

## 3. `Issue6070GrpcGraphBatchIT.aFailureOnAnInteriorFlushIsNotResentByClose` is no longer a bet on timing

It asserted the failure surfaced during the body. Whether the server's error reaches the client before or after the body finishes is a race - measured on #6152, whose diff touches nothing in gRPC or batch: **failed 4 of 6 CI runs, passed the other 2**, across commits with byte-identical gRPC code. It now accepts either surfacing point and keeps the check the test is really about: the failure is raised exactly once and `close()` does not invent a second one on top of it. The `countOf(PERSON) == 2` assertion holds either way - chunks handed to a terminated stream are dropped, so no further buffer commits whichever branch raises.

## 4. Sampled controls with no tolerance

`Issue6116ChecksumsPointInTimeTest`'s control asserted a sampled fraction `> 0.9` and failed CI at `0.8953148525374192` - that is the CONTROL failing, not the behaviour under test. The control's only job is to prove this machine's sampler sees suspensions at all, and a sampled fraction of a run that does real work either side of the suspension window cannot be relied on to reach 0.9 on a 4-CPU runner. Widened to 0.75, which still separates it from the 0.5 the real assertion (`withSnapshot < 0.5`, wide margin, unchanged) allows.

`Issue6075SnapshotBackupIT` carries the byte-identical control at 0.9 and is one of the `integration-tests` lane's recurring offenders named in item 6, so it is widened with it rather than left to fail next.

## 6. The chronically red `integration-tests` lane

Not a code change: items 3 and 4 are the triage. Both recurring offenders named in the issue are addressed here.

---

## Tests

- **New** `InternalBucketNamingTest` (7): composition of every convention, attribution of each composed name back to its owner, owners containing underscores, no-owner cases, and the load-bearing one - a convention nobody declared is still recognised.
- **New** `Issue6168MemoizedUnreferencedCountTest` (4): one test per half of the gate, each proving the OTHER half did not move (the schema-only case asserts `getModificationCount()` is unchanged across `removeBucket`), plus a mutation sequence where the memoized answer must equal a fresh walk at every step.
- Re-ran green: `com.arcadedb.graph.**` (515), `com.arcadedb.schema.**` (440), `com.arcadedb.engine.**` + `com.arcadedb.index.**` (1757), `Issue6116ChecksumsPointInTimeTest` (5), `Issue6144SchemaInstalmentMetricsTest` (4). Full reactor builds.
- `Issue6070GrpcGraphBatchIT` could not be run locally (a foreign service already owns port 2480 on this machine); it compiles, and CI is the authority on it - which is the point of the change.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/UnreferencedFiles.java
+++ b/engine/src/main/java/com/arcadedb/engine/UnreferencedFiles.java
@@ -146,6 +146,15 @@ public class UnreferencedFiles {
    * leader's value outright - so the entry is valid only while BOTH halves are unchanged, never while they are "not
    * newer".
    * <p>
+   * <b>What the gate rests on, stated so it can be checked.</b> Because the version can move BACKWARDS, equality of
+   * the pair is only as good as this: within one database's lineage a given schema version denotes one schema
+   * CONTENT, so returning to a version means returning to the claimed-file set that version described - and the
+   * cached count is the count of that set. The file half needs no such assumption; it only ever increases, so an
+   * equal value means no file was registered or dropped in between, full stop. A stale answer therefore needs two
+   * nodes to have assigned the same version to DIFFERENT schema content, which is schema divergence - a condition
+   * that breaks replication itself long before it misreports a gauge, and is not a state this cache should be
+   * designed around.
+   * <p>
    * <b>Thread safety without a lock.</b> The gate is read BEFORE the walk and published with the count afterwards,
    * so a change that lands mid-walk tags the result with the gate it was computed under, and the next call - seeing
    * the newer gate - recomputes. Two callers racing can therefore duplicate work or overwrite each other's entry,
--- a/engine/src/main/java/com/arcadedb/engine/UnreferencedFiles.java
+++ b/engine/src/main/java/com/arcadedb/engine/UnreferencedFiles.java
@@ -22,6 +22,7 @@ import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.index.Index;
 import com.arcadedb.index.IndexInternal;
 import com.arcadedb.schema.DocumentType;
+import com.arcadedb.schema.InternalBucketNaming;
 import com.arcadedb.schema.LocalDocumentType;
 import com.arcadedb.schema.LocalSchema;
 
@@ -31,6 +32,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
 /**
@@ -64,8 +66,8 @@ import java.util.function.Supplier;
  * false positive on a healthy database - the one outcome that would make this diagnostic worse than useless.
  * Manual indexes are claimed for the same reason: they are bound to no type by design. So is any bucket whose NAME
  * is derived from a type's or a claimed bucket's - the edge-list buckets of a vertex bucket, a super-node stripe
- * pool, a paired external bucket - see {@link #isDerivedFromAnOwner}, which recognises the convention rather than
- * the individual features that follow it.
+ * pool, a paired external bucket - see {@link InternalBucketNaming#ownerOf}, the one home of that convention, which
+ * recognises the SHAPE rather than the individual features that follow it.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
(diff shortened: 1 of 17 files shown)


lvca:
### The `integration-tests` failure on this PR is not this PR's

`Issue5470BatchStreamStallIT.aBodyShorterThanItsContentLengthIsNotReportedAsSuccess` failed on
[run 31827090226](https://github.com/ArcadeData/arcadedb/actions/runs/31827090226). It fails on `main` too, with the
identical assertion at the identical line:

| commit | run |
|---|---|
| `1c5859ac` (main) | [31776078945](https://github.com/ArcadeData/arcadedb/actions/runs/31776078945) |
| `b6708e52` (main) | [31776042204](https://github.com/ArcadeData/arcadedb/actions/runs/31776042204) |

This PR touches no HTTP handler, no batch path and no server code outside the two HA metric accessors, so there is no
diff between it and `main` that could reach that test.

Triaged rather than just attributed, and it turned out not to be a threshold flake at all: **409 is the duplicate-key
mapping**, the fixture's `V1.id` is UNIQUE, and the payload's ids are unique within itself - so reaching 409 means the
truncated upload applied its records TWICE, which is the exact hazard that test's javadoc says it exists to prevent.
Filed with the analysis as #6180; it wants a fix in `PostBatchHandler`, not a widened assertion, so it is deliberately
not folded in here.

That makes it a third recurring offender on the lane #6168 item 6 describes, alongside the two this PR does fix.

lvca:
### Both lanes are green on re-run; both failures were flakes

Re-ran `integration-tests` and `ha-integration-tests` on the byte-identical tree
([run 31827090226](https://github.com/ArcadeData/arcadedb/actions/runs/31827090226)) and **both passed**. Every check
on this PR is now green and the PR is `CLEAN`.

- `integration-tests` - `Issue5470BatchStreamStallIT`, filed as #6180. The re-run passing corrects that issue rather
  than closing it: the failure is a RACE (an intermittent double-insert), which is why the same tree can go either
  way, and a green run does not show the bug is absent. Noted there.
- `ha-integration-tests` - `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`, a `TransactionCommittedRemotely`
  on the apply path. This PR's only ha-raft change is a metrics field plus two accessors on `RaftReplicatedDatabase`;
  nothing in it is reachable from commit or apply, and the memoization strictly REDUCES the work the metrics timer
  does. This is #5702 ("`ha-integration-tests` fails on main ~75% of the time, with a different test each run") -
  `main`'s own current failure on that lane is a different test again (`RemoteStickyConnectionStrategyIT`).

Establishing this took two full CI cycles of downloading and diffing job logs across `main` and this branch - which is
the exact cost #6168 item 6 describes, on the two lanes it names.

PR #6181, opened 2026-08-14
perf(#6174): fold a constant-false filter and LIMIT 0 at plan time instead of scanning the target

### LINKED ISSUE #6174: Constant-false filters (WHERE 1=0) are not folded at plan time, so every schema probe scans its target in full
Spun out of #6156 / #6172.

### What happens

The SQL planner has no notion of a filter that is false for every row. `WHERE 1=0` is evaluated per row like any other condition, so this:

```sql
SELECT * FROM (SELECT name FROM Character) SPARK_GEN_SUBQ_0 WHERE 1=0
```

scans every `Character`, evaluates `1=0` once per record, discards all of them and returns nothing. The work is entirely wasted: the result is knowable from the statement alone.

### Why it matters

This is not an academic shape. It is what Spark, and several BI tools over the Postgres wire, send to discover a query's schema - one probe per pushed-down query, against the real target. On a large type that is a full scan to learn something the parser already knows. #6156 was about those probes returning no columns; this is about what they cost even now that they answer correctly.

`LIMIT 0` is the other common spelling of the same intent and has the same problem.

### Suggested direction

Recognise an always-false filter during planning and short-circuit the plan to an empty result, instead of building a scan the filter step will drain.

`PostgresNetworkExecutor.isAlwaysFalse` (added in #6172) already does the recognition part and could inform the shape of an engine-side version: read the condition in the disjunctive normal form `BooleanExpression.flatten()` produces - the filter is false only when every alternative carries a term that is - and fold only comparisons between literals, since `Expression.isEarlyCalculated()` admits arbitrary function calls and nothing on `SQLFunction` marks a function as pure.

Two things to be careful about if this is taken on:

- **A `LET` clause, or a subquery in the projection, may be expected to run.** Skipping the plan entirely skips those too. Worth deciding deliberately whether an always-false `WHERE` short-circuits the whole statement or only the target scan beneath it.
- **`PostgresNetworkExecutor.sampleProbeColumns` depends on the probe returning zero rows** to reach its schema-discovery path at all, which it still would - it keys off the empty result, not off how the emptiness was produced. But the two should be looked at together, since folding would make the probe cheap and the replay would then be the dominant cost.

An engine-level test would want to assert the plan shape (no scan step), not just that zero rows come back, since the current behaviour already returns zero rows correctly.

lvca:
Fixes #6174.

## The problem

The SQL planner had no notion of a filter that is false for every row. `WHERE 1=0` was evaluated per record like
any other condition, so

```sql
SELECT * FROM (SELECT name FROM Character) SPARK_GEN_SUBQ_0 WHERE 1=0
```

scanned every `Character`, evaluated `1=0` once per record and discarded all of them. That is the shape Spark, and
several BI tools over the Postgres wire, send to discover a query's schema - one probe per pushed-down query,
against the real target. `LIMIT 0` is the other spelling of the same intent and cost the same.

## The fix

The recognition lives on the AST, next to the `isAlwaysTrue()` that was already there:

- `BooleanExpression.isAlwaysFalse(CommandContext)` - default `false`, overridden structurally by `AndBlock` (false
  when any term is), `OrBlock` (false only when every alternative is), `ParenthesisBlock` (delegates) and
  `BinaryCondition` (folds the comparison). Structural rather than over `flatten()`'s DNF: it is O(n) instead of
  the cartesian product a DNF expansion of a wide OR produces, and it answers the same question.
- `BinaryCondition` folds only when **both** operands are literals (`Expression.isLiteral()` /
  `MathExpression.isLiteral()` / `BaseExpression.isLiteral()`), and folds by calling the very same
  `evaluate()` the filter step would have called, so the fold and the runtime can never disagree. A bound parameter
  is excluded because the plan outlives the execution that bound it; a function call is excluded because nothing on
  `SQLFunction` marks a function as pure - `isEarlyCalculated()` is not a purity check.
- `Limit.isAlwaysEmpty()` for a literal `LIMIT 0` (not a parameter, not an expression).

`SelectExecutionPlanner` reads that off the clauses as the statement wrote them (before `optimizeQuery()`
rearranges them into index searches) and, after the target has been planned, drops the fetch steps and chains an
`EmptySourceStep` instead. The rest of the plan is untouched.

Three things the fold deliberately keeps:

- **target validation**: the fetch is planned and then discarded, so `SELECT FROM ATypeThatDoesNotExist WHERE 1=0`
  still reports the missing type instead of quietly returning nothing;
- **the aggregation semantics**: everything downstream of the fetch stays and receives no row, which is exactly
  what it receives today from the filter being removed - `SELECT count(*) FROM T WHERE 1=0` still returns its one
  row with 0, and with a `GROUP BY` still returns none;
- **a `LET` evaluated once per statement**: it is chained ahead of the empty source, which pulls it exactly like
  `EmptyStep` does.

`EmptySourceStep` is a new step rather than a reuse of `EmptyStep` because `EmptyStep.canBeCached()` is
(deliberately) `false`: it encodes one execution's *data*. "This statement cannot return a row" is a property of
its *text*, so the folded plan is cacheable.

Because `DELETE`/`UPDATE` build their target through `SelectExecutionPlanner`, they are covered by the same code.

## Two behaviour changes, and a wrong answer found next to them

`SELECT count(*) FROM T LIMIT 0` used to return one row: `count(*)` on a bare type is answered by a hardwired plan
that replaces the whole step chain and returns before any `LIMIT`/`SKIP` step is chained. It now returns no rows.
This is the SQL twin of the Cypher defect fixed in #5715.

The same hole was a wrong answer for `SKIP`, on all three hardwired plans: `SELECT count(*) FROM T SKIP 1` and
`SELECT max(indexedProperty) FROM T SKIP 1` handed back the very row they were told to skip. Both now chain the
statement's SKIP/LIMIT themselves, which costs nothing on a plan that produces one row.

Finally, a `LIMIT 0` is folded whatever the `WHERE` says, so - unlike the constant-false fold, which never reaches
a function - the filter is not evaluated at all. The only way to observe that is a predicate that raises:
`SELECT * FROM T WHERE 1/0 = 1` reports a division by zero, and with `LIMIT 0` it now returns an empty result.

## Reuse over duplication

`PostgresNetworkExecutor.isAlwaysFalse` (added in #6172) is gone: the probe detection now calls
`WhereClause.isAlwaysFalse()`, removing ~90 lines that duplicated the recognition.

## Tests

New `ConstantFalseFilterFoldingTest` (24 tests), verified red on main for the folding cases. It asserts the plan
shape (no `FetchFrom*`/`CountFrom*` step anywhere, including inside sub-plans) *and* the `readRecord` statistic, not
only that zero rows come back - the current behaviour already returns zero rows, it just reads the whole type to do
it. It also pins what must NOT be folded: a bound parameter, a function call, a satisfiable OR branch, `NOT (1=0)`, `LIMIT ?` and the `LIMIT -1` "unlimited" sentinel.

Run: engine `com.arcadedb.query.**` (10947 tests), the rest of the engine module, postgresw units + integration.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1540,6 +1540,13 @@ What the fold deliberately keeps:
 The fold applies wherever a `SELECT` plan is built, so a subquery, a `DELETE ... WHERE 1=0` and an
 `UPDATE ... WHERE 1=0` are covered by the same code.
 
+The two spellings are not recognised the same way, and it is worth being explicit about the difference. A
+constant-false filter is folded only when the statement itself says so, never through a function. A `LIMIT 0`
+truncates the result to nothing whatever the filter would have done, so the filter is not evaluated at all - which
+means a predicate that raises at runtime stops raising: `SELECT * FROM T WHERE 1/0 = 1` reports a division by zero,
+and the same statement with `LIMIT 0` now returns an empty result instead. Nothing else can observe the difference,
+since no row was going to come back either way.
+
 One behaviour change worth naming, and one wrong answer found next to it. `count(*)` on a bare type is answered
 from the bucket counters by a hardwired plan that replaces the whole step chain - and returned without chaining the
 statement's `SKIP` or `LIMIT` at all. So `SELECT count(*) FROM T LIMIT 0` returned one row, and
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1509,3 +1509,45 @@ its query on every read of the type. This is the fix, not a regression, but it i
 trigger exists in an upgraded database.
 
 [#6134](https://github.com/ArcadeData/arcadedb/issues/6134)
+
+## A constant-false filter and `LIMIT 0` are folded at plan time, so a schema probe stops scanning its target (#6174)
+
+`SELECT * FROM (SELECT name FROM Character) SPARK_GEN_SUBQ_0 WHERE 1=0` is not an academic shape: it is what Spark,
+and several BI tools over the Postgres wire, send to discover a query's schema - one probe per pushed-down query,
+against the real target. The planner had no notion of a filter that is false for every row, so it built the full
+scan and let the filter step drain it: every probe read the whole type to return nothing that the statement itself
+did not already say. `LIMIT 0` is the other common spelling of the same intent and cost the same.
+
+Both are now recognised while the plan is built and answered with an `EMPTY RESULT` source step in place of the
+target fetch, which is what `EXPLAIN` now shows. Recognition is structural, over the condition tree the statement
+wrote - through parentheses, and through AND/OR (an AND is false when any term is, an OR only when every
+alternative is) - and folds only comparisons whose **both** operands are literals, evaluated by the very same
+`BinaryCondition.evaluate` the filter step would have called. A bound parameter is never folded, because the plan
+outlives the execution that bound it; nor is a function call, because nothing on `SQLFunction` marks a function as
+pure and classifying a statement must not be a reason to invoke one. `WHERE uuid() = 'x'` and `WHERE ? = 0` are
+therefore still planned as scans, as they must be.
+
+What the fold deliberately keeps:
+
+- **The target is still resolved**, so `SELECT FROM ATypeThatDoesNotExist WHERE 1=0` still reports the missing type
+  rather than quietly returning nothing. Only the scan is dropped, after the fetch has been planned.
+- **Everything downstream of the fetch stays in the plan** and receives no row, which is exactly what it receives
+  today from the filter being removed: `SELECT count(*) FROM T WHERE 1=0` still returns its one row with 0, and the
+  same statement with a `GROUP BY` still returns none.
+- **A `LET` evaluated once per statement still runs**; only the per-record `LET` and the per-record projection work
+  disappear, along with the rows that were never going to come back.
+
+The fold applies wherever a `SELECT` plan is built, so a subquery, a `DELETE ... WHERE 1=0` and an
+`UPDATE ... WHERE 1=0` are covered by the same code.
+
+One behaviour change worth naming: `SELECT count(*) FROM T LIMIT 0` used to return one row. `count(*)` on a bare
+type is answered from the bucket counters by a hardwired plan that returned before any `LIMIT`/`SKIP` step was
+chained, so the `LIMIT 0` was ignored. It now returns no rows, which is what the clause asks for and what every
+other engine answers. This is the SQL twin of the Cypher defect fixed in #5715, where the CSR count push-down
+replaced the whole step chain and `RETURN count(*) LIMIT 0` likewise returned a row.
+
+The Postgres wire protocol reuses this recognition instead of its own copy: the schema-probe detection added in
+#6172 now calls `WhereClause.isAlwaysFalse()`, and the ~90 lines that duplicated it in `PostgresNetworkExecutor`
+are gone.
+
+[#6174](https://github.com/ArcadeData/arcadedb/issues/6174)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1543,10 +1543,11 @@ The fold applies wherever a `SELECT` plan is built, so a subquery, a `DELETE ...
 One behaviour change worth naming, and one wrong answer found next to it. `count(*)` on a bare type is answered
 from the bucket counters by a hardwired plan that replaces the whole step chain - and returned without chaining the
 statement's `SKIP` or `LIMIT` at all. So `SELECT count(*) FROM T LIMIT 0` returned one row, and
-`SELECT count(*) FROM T SKIP 1` handed back the very row it was told to skip. The first is now folded away before
-the hardwired plan is even considered; the second is fixed by chaining both steps after the count, which costs
-nothing on a plan that produces a single row. This is the SQL twin of the Cypher defect fixed in #5715, where the
-CSR count push-down replaced the whole step chain and `RETURN count(*) LIMIT 0` likewise returned a row.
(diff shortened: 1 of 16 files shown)


lvca:
Thanks - all three points addressed in 5c4886c07.

**`isMinimalQuery()` comment**: rather than update the comment, the hole it would have documented is closed. The hardwired count plans replace the whole step chain and returned without chaining SKIP or LIMIT at all, so `SELECT count(*) FROM T SKIP 1` handed back the very row it was told to skip (verified before the fix). `handleSkipAndLimitAfterHardwired()` now chains both after `CountFromTypeStep`/`CountFromIndexStep` - free on a plan that produces one row - and the comment is accurate again. (`max`/`min` on index already answered SKIP correctly, so that path is untouched.) New test: `countStarWithASkipReturnsNoRow`.

**`NotBlock`**: added, mirroring the `isAlwaysTrue()` beside it - `NOT <always true>` is false for every record, and without the negation the block is a pass-through. Note the negated form only pays off once some leaf answers `isAlwaysTrue()`; no leaf does today (`isAlwaysTrue()` has no consumer outside the AST itself), so this is completeness rather than a new fold. The test pins the direction that would be a wrong answer: `WHERE NOT (1=0)` must still return every row.

**`WHERE false`**: it does parse, to the `BooleanExpression.FALSE` sentinel, and it was already folded - just untested. `aLiteralFalseFilterIsFoldedToo` now covers `WHERE false`, `WHERE true AND false` and the `WHERE true` counter-case.

Re-ran engine `com.arcadedb.query.**`: 10952 tests green.

lvca:
Good catch on MAX/MIN - fixed in 18f9c2c77.

**`handleHardwiredMaxMinOnIndex`**: confirmed the bug before fixing it. My first probe of this defect used an unindexed property, so `SELECT max(n) FROM T SKIP 1` fell through to the ordinary pipeline and answered correctly - which is exactly why the MAX/MIN path looked clean and got left out. With a range index on the property it reproduces just as you describe. `handleSkipAndLimitAfterHardwired` is now called there too, and the new test (`maxOnAnIndexedPropertyWithASkipReturnsNoRow`, which creates the property + LSM_TREE index first) was verified non-vacuous: it fails with `Expecting value to be false but was true` when the one-line call is removed. `isMinimalQuery()`'s comment is now true of all three hardwired plans.

**`NOT (1=1)`**: leaving this one. Folding it needs `BinaryCondition.isAlwaysTrue()`, and that method's signature takes no `CommandContext` - the fold would have to call `operator.execute(null, ...)` through `Expression.execute` with a null context, which is a wider contract change than the shape justifies. `isAlwaysTrue()` also has no consumer outside the AST's own recursion today, so implementing it on the leaf buys exactly this one query shape. Happy to take it as a separate issue if you want the symmetry; as a probe shape it is not one anything sends.

Re-ran engine `com.arcadedb.query.sql.**`: 2538 tests green.

lvca:
Both points taken - b6b408856 (docs + tests only, no behaviour change).

**`LIMIT 0` and the un-evaluated WHERE**: intentional, and you are right that it deserved the same explicit callout as the `count(*)` one. The two spellings genuinely are not decided the same way: a constant-false filter is folded only when the statement itself says so and never reaches a function, while a `LIMIT 0` truncates the result whatever the filter would have done, so the filter is simply not evaluated. The observable difference is exactly the one you name - a predicate that raises stops raising. `SELECT FROM Character WHERE 1/0 = 1` reports `ArithmeticErrorException: / by zero`; with `LIMIT 0` it now returns an empty result. That is now stated in `emptyByConstructionReason`'s javadoc, in the release-notes entry, and pinned by `limitZeroIsFoldedWhateverTheFilterSays` (which asserts both halves, so the day someone makes the fold conditional on the filter the test says why).

**`LIMIT -1`**: added `aNonZeroOrParameterizedLimitIsNeverFolded` coverage. Confirmed it parses, is not folded, and returns every row - the plan keeps its `LIMIT ( LIMIT -1)` step, which `LimitExecutionStep` reads as unlimited.

**Broad `catch (Exception)` in `BinaryCondition.isAlwaysFalse`**: leaving as is, for the reason you give - the fallback is "don't fold, let the runtime report it", which is the same answer any narrower catch would produce for the cases that matter, and `BinaryComparator` raises a fairly open set of types across the operators.

Engine `com.arcadedb.query.sql.executor.**` green.

lvca:
Both minor observations applied in 6510f7961 - no behaviour change.

**`info.planCreated`**: you are right that it was a write nothing reads; `handleWhere` is the only reader and the fold path never calls it. Removed rather than commented - defensive bookkeeping that no reader can act on is just a question mark for the next person.

**`(1) = 0`**: `MathExpression.isLiteral()`'s javadoc now names the arithmetic parenthesis alongside the array literal and the CASE, and says plainly that these are missed folds rather than unsafe ones, so the omission reads as a choice.

That closes everything raised across the four rounds. Summary of what the reviews changed beyond the original fold: the SKIP that all three hardwired plans (`count(*)` on type, `count(*)` on index, `max`/`min` on index) silently dropped, `NotBlock.isAlwaysFalse`, and tests for `WHERE false`, `NOT (1=0)`, `LIMIT -1` and the `LIMIT 0` filter-not-evaluated consequence.

lvca:
Applied in 1585b02c4: `anUnknownTargetIsStillReported` now uses `assertThatThrownBy(...).hasMessageContaining(...)` like the rest of the class, so the fully-qualified `Assertions.fail` and the hand-rolled try/catch are both gone. PR description refreshed with the real test count (24) and the two behaviour changes.

On `NOT (1=1)`: still leaving it, for the reason given last round - folding it needs `BinaryCondition.isAlwaysTrue()`, whose signature carries no `CommandContext`, so the fold would have to evaluate through a null one. That is a wider contract change than a query shape nobody writes justifies.

Nothing outstanding from my side. Local runs backing this branch: engine `com.arcadedb.query.**` 10952, the rest of the engine module 4928, postgresw 260 units + the integration suite including `PostgresWJdbcIT`. CI is the authority on the rest.

PR #6182, opened 2026-08-14
fix(#6179): an expression whose modifier or CASE branch reads the record is no longer "early calculated"

### LINKED ISSUE #6179: isEarlyCalculated() reads as a purity check but only means "needs no record", and BaseExpression ignores its modifier
Found while implementing #6172 (Postgres schema probe). Not a live bug as far as I can tell - the one path that could have turned it into a correctness problem is blocked by an unrelated guard - but the contract is a trap for new code, and I walked straight into it.

### The contract is narrower than the name

`Expression.isEarlyCalculated(CommandContext)` answers "can this be computed without a record". It says nothing about whether computing it is *free of consequences*. `FunctionCall.isEarlyCalculated` (`FunctionCall.java:334`) returns true for **any** function whose arguments need no record:

```java
public boolean isEarlyCalculated(final CommandContext context) {
  if (isTraverseFunction(context))
    return false;
  for (final Expression param : params) {
    if (!param.isEarlyCalculated(context))
      return false;
  }
  return true;
}
```

So `uuid()`, `sysdate()` and any user-defined function taking literal arguments are all "early calculated". There is no purity or determinism marker anywhere on `SQLFunction` / `RecordFunction` to distinguish them from a pure one. (`com.arcadedb.function.StatelessFunction` exists in the sibling registry hierarchy, but "stateless" there means "does not need the current record" - the same notion as `isEarlyCalculated`, not purity, so it does not help.)

The name invites reading it as "this is a constant, evaluating it is free". It isn't.

### Where it already leads to an extra invocation

Two call sites take `isEarlyCalculated` as licence to `execute()` an expression at plan time, in addition to whatever per-row execution does later:

- `WhereClause.getEqualityOperations` (`WhereClause.java:184`) - index cost estimation
- `SelectExecutionPlanner` partitioned-bucket pruning (`SelectExecutionPlanner.java:2294`)

For `WHERE partitionKey = someUserFunction()` the function is therefore called once during planning purely to inform the plan. Harmless for a pure function, an unintended extra call for anything else.

### What is *not* broken (and why that is worth writing down)

Plan-cache poisoning would be the serious version of this: a non-deterministic function evaluated at plan time, its value baked into a cached plan, reused across sessions. The pruning code guards `literalSide.containsInputParameter()` with a comment about exactly that hazard, and a function is not an input parameter - so the guard does not cover it.

It is nonetheless unreachable, because `FunctionCall.isCacheable()` (`FunctionCall.java:404`) returns true **only** for graph traversal functions (`out`, `in`, `both`, ...). Any other function in a `WHERE` makes `SelectStatement.executionPlanCanBeCached()` false, so the plan is never cached and the stale-bucket scenario cannot happen.

That is load-bearing and entirely accidental: two unrelated predicates in different files happen to line up. Widening `isCacheable()` to admit pure scalar functions - a natural future optimisation - would silently make the pruning path unsafe. Worth a comment at the pruning site tying it to `isCacheable()`, at minimum.

### `BaseExpression.isEarlyCalculated` ignores its modifier

`BaseExpression.java:357`:

```java
public boolean isEarlyCalculated(CommandContext context) {
  if (number != null || inputParam != null || string != null)
    return true;
  if (expression != null)
    return expression.isEarlyCalculated(context);
  return identifier != null && identifier.isEarlyCalculated(context);
}
```

`modifier` is never consulted, so `foo('x').someMethod()` is classified purely on `foo('x')` and the method chain is invisible. `isBaseIdentifier()`, twelve lines above, *does* check `modifier == null` - so the class already knows modifiers change what an expression is. This looks like an oversight rather than a decision. I could not construct a case where it currently produces wrong behaviour, but it widens the surface of the point above.

### Suggested direction

1. Rename or document `isEarlyCalculated` so its contract ("no record needed", *not* "no side effects, not constant") is unmistakable at the call site.
2. Add a genuine literal/constant predicate for callers that want to *evaluate* something at plan time. #6172 needed exactly this and ended up with a private `isLiteral(Expression)` / `isLiteral(MathExpression)` walk in `PostgresNetworkExecutor` - number, string, boolean, null, arithmetic over them, with a `Modifier` disqualifying an operand. That walk is a reasonable starting point for a shared parser-level version, and the wire module could then drop its copy.
3. Consider whether `BaseExpression.isEarlyCalculated` should consult `modifier`, consistently with `isBaseIdentifier()`.
4. Comment the `isCacheable()` dependency at the partition-pruning site so the coupling is not discovered the hard way.

### How #6172 hit it

The Postgres wire has to decide whether an empty result came from a schema probe (`WHERE 1=0`) before replaying the query to discover its columns. The first version used `isEarlyCalculated` on both operands and then evaluated the comparison. That meant every query returning zero rows with a filter like `someFunction() = 42` had `someFunction()` invoked once more, just to be classified. Caught in review and narrowed to literals; the regression test is `PostgresWJdbcIT.schemaProbeForTypelessAndComputedProjectionsReturnsTheDescription`, which asserts `uuid() = 'nomatch'` is *not* folded.

lvca:
Closes #6179.

The issue was filed as a contract/naming trap found while implementing #6172, with the note "not a live bug as far as I can tell". It is one: two nodes answer `isEarlyCalculated()` - "can this be computed without a record" - without looking at the part of themselves that reads one.

### The bug

- **`BaseExpression`** judged the expression on its leftmost term alone and never consulted the modifier chain, so `'Mr.'.append(surname)` and `['a','b'][idx]` were reported as record-free. (`isBaseIdentifier()`, twelve lines above, already checks `modifier == null`.)
- **`CaseExpression`** keeps its branches outside `childExpressions`, so `MathExpression`'s inherited walker found an empty list and returned true for a CASE that reads the record on every branch.

The planner takes that answer as licence to resolve the equality through an index and to evaluate the expression against a `null` record, so the record-dependent part silently resolves to null. The query then answers differently depending on whether an index exists. On the fixture in the new test the indexed type returns **no rows** where the identical unindexed type returns two:

```sql
SELECT fullName FROM Person       WHERE fullName = 'Mr.'.append(surname)   -- indexed:   []
SELECT fullName FROM PersonNoIndex WHERE fullName = 'Mr.'.append(surname)  -- scan:      [Mr.Smith, Mr.Jones]
```

The same classification also promotes a per-record `LET` to a global one (`SelectExecutionPlanner.splitLet`), with the same effect.

The fix walks what was skipped: the modifier chain (method-call arguments, array range and index selectors; a suffix reads a property of the value it is applied to, so it still needs no record; a filter or nested projection is never assumed record-free) and every CASE branch. A traversal method (`.out()`, `.in()`, ...) is excluded like `FunctionCall` already excludes traversal functions.

### The contract, written down where it is read

`isEarlyCalculated` says nothing about purity, determinism or constant-ness - any function over literals qualifies, `uuid()` included - and there is no marker anywhere on `SQLFunction` that would say otherwise. That is now in its Javadoc, in `FunctionCall`'s, and in `MethodCall`'s.

A caller that intends to **evaluate** an expression while planning wants a constant instead. `Expression.isLiteral()` is that predicate, raised to the parser from the private walk `PostgresNetworkExecutor` grew in #6172 - which now calls it and drops its copy. The optional `isLiteral(true)` form admits a bound input parameter, for callers whose result does not outlive the execution.

Two plan-time evaluation sites move onto it:

- **partitioned-bucket pruning** binds a partition coordinate only from a literal now. The bucket it derives is baked into the plan, so the value must be fixed for the plan's life and free to compute; a function was neither. The path was safe only because `FunctionCall.isCacheable()` is true for traversals alone - two unrelated predicates in different files happening to line up, exactly the accident the issue flagged. Pruning no longer leans on it, so widening `isCacheable()` later cannot silently break it. The explicit `containsInputParameter()` guard goes with it: `isLiteral()` already excludes parameters.
- **the MATCH index-cost estimate** (`WhereClause.getEqualityOperations`) evaluates literals and bound parameters only, instead of invoking a user function once per planning pass just to sharpen an estimate.

### Tests

- `Issue6179EarlyCalculatedModifierTest` - indexed and unindexed types with identical rows must answer identically for a method call, a collection selector and a CASE over a record field; a modifier over a literal and a function over literals must still be resolved through the index (the contract's other half).
- `Issue6179LiteralExpressionTest` - the two predicates side by side, above all where they disagree: functions and modifiers over literals are early calculated but not literal; a parameter is a literal only for the caller that says so; `null` is the one shape that is a literal but has never been early calculated (an index lookup on a null key is not the filter's question, and must not start to be).
- `PartitionPruningPlannerTest.computedPredicateDoesNotNarrowIndexFilterBuckets` - a computed value no longer prunes, and the rows still come back.

Verified red before / green after for both the modifier and the CASE case. Full engine suite green (11931 tests); `PostgresWJdbcIT` green (44), including the #6172 probe tests that assert `uuid() = 'nomatch'` is not folded.

### Spotted while testing, not touched here

`SELECT FROM Doc WHERE @rid > :p` returns zero rows for a parameter bound either to a RID or to its string form, while the same query with a RID literal answers correctly. Pre-existing and unrelated to this change (`SelectExecutionPlanner.isRidRange` evaluates the right side unguarded); will file separately.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -2370,11 +2370,16 @@ public class SelectExecutionPlanner {
           continue;
         final int idx = idxBoxed;
 
-        if (literalSide == null || !literalSide.isEarlyCalculated(context))
-          continue;
-        // Plan-time pruning is unsafe when the literal is parameter-bound: the cached plan would
-        // reuse the first execution's bucket id for every subsequent binding.
-        if (literalSide.containsInputParameter())
+        // Only a literal is bound here, not everything Expression.isEarlyCalculated() would admit. The
+        // bucket derived below is baked into the plan, so the value it comes from must be fixed for the
+        // life of that plan and free to compute: a parameter is not (a cached plan would reuse the first
+        // execution's bucket id for every subsequent binding, which is what the isEarlyCalculated form
+        // used to exclude by hand), and a function call is neither - it would be invoked at plan time,
+        // once more than the query asked for, and a non-deterministic one would fix a bucket that its
+        // per-row value no longer agrees with. That the plan is not cached today when the WHERE holds a
+        // function is an accident of FunctionCall.isCacheable() being true only for traversals; pruning
+        // no longer leans on it. See issue #6179.
+        if (literalSide == null || !literalSide.isLiteral())
           continue;
 
         if (!bound[idx]) {
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayLiteralExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayLiteralExpression.java
@@ -100,6 +100,20 @@ public class ArrayLiteralExpression extends MathExpression {
     return true;
   }
 
+  /**
+   * {@code items} sit outside {@code childExpressions}; the inherited walker would miss them.
+   *
+   * @see Expression#isLiteral(boolean)
+   */
+  @Override
+  public boolean isLiteral(final boolean allowInputParameters) {
+    for (final Expression item : items) {
+      if (!item.isLiteral(allowInputParameters))
+        return false;
+    }
+    return true;
+  }
+
   @Override
   public boolean isAggregate(final CommandContext context) {
     for (final Expression item : items) {
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayNumberSelector.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayNumberSelector.java
@@ -82,6 +82,13 @@ public class ArrayNumberSelector extends SimpleNode {
     return null;
   }
 
+  /**
+   * @see Expression#isEarlyCalculated(CommandContext)
+   */
+  public boolean isEarlyCalculated(final CommandContext context) {
+    return expressionValue == null || expressionValue.isEarlyCalculated(context);
+  }
+
   public ArrayNumberSelector copy() {
     final ArrayNumberSelector result = new ArrayNumberSelector();
     result.inputValue = inputValue == null ? null : inputValue.copy();
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayRangeSelector.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ArrayRangeSelector.java
@@ -142,6 +142,14 @@ public class ArrayRangeSelector extends SimpleNode {
     return Arrays.asList(Arrays.copyOfRange(arrayResult, lFrom, lTo));
   }
 
+  /**
+   * @see Expression#isEarlyCalculated(CommandContext)
+   */
+  public boolean isEarlyCalculated(final CommandContext context) {
+    return (fromSelector == null || fromSelector.isEarlyCalculated(context)) //
+        && (toSelector == null || toSelector.isEarlyCalculated(context));
+  }
+
   public ArrayRangeSelector copy() {
     final ArrayRangeSelector result = new ArrayRangeSelector();
     result.from = from;
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ArraySelector.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ArraySelector.java
@@ -87,6 +87,13 @@ public class ArraySelector extends SimpleNode {
     return result;
   }
 
+  /**
+   * @see Expression#isEarlyCalculated(CommandContext)
+   */
+  public boolean isEarlyCalculated(final CommandContext context) {
+    return expression == null || expression.isEarlyCalculated(context);
+  }
+
   public ArraySelector copy() {
     final ArraySelector result = new ArraySelector();
     result.rid = rid == null ? null : rid.copy();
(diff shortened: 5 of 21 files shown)


lvca:
@claude please resolve conflicts with main

lvca:
Rebased onto main after #6181 (the #6174 constant-false fold) landed. That PR introduced its own `Expression.isLiteral()` for the same reason this one did, so the conflict was semantic rather than textual and the two are now **one predicate**:

- `isLiteral()` keeps #6181's exact contract and callers (`BinaryCondition.isAlwaysFalse`, the planner fold) - a bound parameter is still excluded, because the plan outlives the execution that bound it. It now delegates to `isLiteral(false)`.
- `isLiteral(boolean allowInputParameters)` is the form this PR needs for `WhereClause.getEqualityOperations`, whose result lives only as long as the planning pass.

Because the no-arg form now routes through the same walker, #6181's fold also picks up the shapes this PR taught the predicate to recognise: a RID literal, a collection of literals, a parenthesised literal, an array literal, and a CASE whose branches are all literals. Each is written in the statement and cannot change between executions, which is #6181's own stated definition - its exclusion of RIDs was conservative rather than principled, and folding `#12:0 = #12:1` is correct. `ConstantFalseFilterFoldingTest` passes unchanged.

The postgresw change from the first push is **gone**: #6181 moved that walk onto the AST (`WhereClause.isAlwaysFalse`), which is where this PR was heading, so `PostgresNetworkExecutor` is no longer touched at all. Its regression tests still cover the path - `PostgresWJdbcIT` 44/44 green, `uuid() = 'nomatch'` still not folded.

Re-verified after the rebase: engine SQL suite 2546 green (including `ConstantFalseFilterFoldingTest`), `PostgresWJdbcIT` 44 green.

PR #6187, opened 2026-08-14
fix(#6183): a routing address two peers both claim is advertised to nobody, and a leader-only gRPC refusal names the leader wherever it is reported

### LINKED ISSUE #6183: Follow-ups from #6091: a derived routing address can point a client back at the refusing node, and only graphBatchLoad names the leader
Follow-ups from #6091 / PR #6171, which gave HA a per-protocol routing table (`getRoutingTable(ROUTING_PROTOCOL)`) and made a follower's `graphBatchLoad` refusal name a dialable gRPC address. None of these blocked that merge; they are recorded here so the reasoning is not re-derived.

Ordered by what I would fix first. Item 1 is a defect the new machine-readable redirect makes reachable; item 2 is a gap the issue explicitly asked about and left open; item 3 is docs.

---

## 1. A derived routing address can send a client straight back to the node that refused it

`RaftHAServer.resolveRoutingAddress` (ha-raft)

With no `bolt:` / `grpc:` field declared, a peer's client address is derived as **that peer's Raft host + THIS node's port for the protocol**. On a cluster where the nodes differ by port rather than by host - several nodes on one machine, a dev or test deployment - every peer derives to the *same* address, so the "leader" address a follower advertises is the follower's own endpoint.

Before #6171 the cost was confined to Bolt's ROUTE response, where a driver that dials a follower for a write gets an error and re-routes. #6171 raises it: the gRPC refusal puts that address on a trailer (`arcadedb-leader-grpc-address`) and the client rebuilds a `ServerIsNotTheLeaderException` around it, so a caller written to redirect *automatically* - the whole point of the trailer - can dial the same follower, be refused again, and loop.

**The check is cheap and provable, because two distinct peers can never legitimately share a `host:port` for one protocol** - they would be fighting over the socket. So a resolution that maps the leader and a follower to the same address is not a resolution, and `getRoutingTable` can say so:

```java
// after resolving writer and readers, in getRoutingTable
if (readers.contains(writer))
  return null;   // the derive cannot distinguish these peers; advertise nothing rather than something wrong
```

`null` is already the "no routing table" answer both callers handle: `BoltNetworkExecutor.handleRoute` falls back to advertising this node as READ/ROUTE only (never writer), and the gRPC refusal falls back to naming the leader's HTTP address with the "use its gRPC port" caveat. Both degrade to what they did before, which is strictly better than a confidently wrong address.

Worth doing for **both** protocols in the shared resolver rather than for gRPC alone - a guard on one protocol only would be exactly the sort of divergence the generalisation in #6171 removed.

Note this changes what an in-process multi-node test observes, so `Issue6091GrpcRoutingTableDerivedIT` (which asserts the derived, same-host address today) has to be rewritten around the new contract - probably by asserting `null` and moving the positive derive assertion to a unit test of `deriveRoutingAddress`, which already exists.

## 2. Only `graphBatchLoad` names the leader; every other leader-only refusal still names nothing

`GrpcErrorMapper.toStatusRuntimeException` (grpcw), `GrpcClientErrorMapper` (grpc-client)

#6091 asked "whether other gRPC RPCs that refuse on a follower should use the same table". The answer found while implementing it: **`graphBatchLoad` is the only RPC in `ArcadeDbGrpcService` with a leadership check at all.** The others do not need one - they go through `RaftReplicatedDatabase`, which raises `ServerIsNotTheLeaderException` for a schema change on a follower - but the two paths now report that same condition very differently:

| | status | exception class trailer | leader address |
|---|---|---|---|
| `graphBatchLoad` | `FAILED_PRECONDITION` | `ServerIsNotTheLeaderException` | on the trailers, gRPC + HTTP |
| everything else (engine-raised) | `ABORTED` (it is a `NeedRetryException`) | `ServerIsNotTheLeaderException` | **none** |

Since #6171 the client reconstructs the right *type* in both cases - before, the engine-raised one fell through to the status-code mapping and surfaced as a `ConcurrentModificationException`, which was simply wrong - but `getLeaderAddress()` returns `null` for everything except a batch load. A caller cannot tell "the leader is unknown, wait for the election" from "the leader is known and nobody bothered to tell you".

The fix belongs in `GrpcErrorMapper`, not in each RPC: when the cause is a `ServerIsNotTheLeaderException` and the server has HA, attach `LeaderRedirectProtocol.LEADER_GRPC_ADDRESS` / `LEADER_HTTP_ADDRESS` the same way `notTheLeader()` does. Every leader-only refusal on the service becomes redirectable at once, and `ArcadeDbGrpcService.notTheLeader` collapses to choosing the status and the wording. `GrpcErrorMapper` is currently static and holds no server reference, which is the one design question to settle.

While there: the status codes should probably agree with each other too. `FAILED_PRECONDITION` says "do not retry this as-is"; `ABORTED` says "retry". For "you are talking to the wrong node" the honest answer is the first, with the address saying where.

## 3. The `grpc:` field is undocumented outside the code

`arcadedb.ha.serverList` grew an object-form `grpc:` field (`host:{raft:2434,http:2480,bolt:7687,grpc:50051}`) and `arcadedb.grpc.port` became a registered `GlobalConfiguration` (so it now honours env vars and appears in the settings listing). The javadoc on `RaftPeerAddressResolver` covers both, but the user-facing server-list reference lives in the docs repo, where `bolt:` got a paragraph when #5002 added it. The `grpc:` field wants the same, including the sentence that matters operationally: **declare it whenever your nodes do not all listen on the same gRPC port**, because the fallback silently assumes they do (see item 1).

lvca:
Closes #6183 (items 1 and 2; item 3 is a docs-repo PR, see below).

## 1. A derived routing address can no longer send a client back to the node that refused it

With no `bolt:` / `grpc:` field declared, a peer's client address is derived as **that peer's Raft host + THIS node's port**. On a cluster whose nodes differ by port rather than by host - several nodes on one machine, a dev or test deployment - every peer derives to the *same* address, so the "leader" a follower advertises is the follower itself. #6091 raised the cost: the gRPC refusal puts that address on a trailer and the client rebuilds a `ServerIsNotTheLeaderException` around it, so a caller written to redirect automatically dials the same follower, is refused again, and loops.

**The issue proposed `if (readers.contains(writer)) return null;`. This does something stricter and more useful instead**, because that test answers the wrong question in two directions: it discards a *correct declared* writer whenever any follower's derive happens to collide with it, and it says nothing about two followers that collide with each other (a driver then load-balances reads onto an address that is one node pretending to be two).

`selectUnambiguousRouting` resolves every peer once, counts the claims on each address, and:

- an address claimed by exactly one peer is advertised;
- a **declared** address beats a **derived** one on collision - the operator stated it, and the collision only proves the guess wrong, so a partially-declared cluster keeps a usable table;
- two declared claims on one address is a configuration error with no defensible winner, so neither is advertised;
- if the writer's address is the ambiguous one, the whole table is `null`.

`null` is what both callers already handle: `BoltNetworkExecutor.handleRoute` advertises this node as READ/ROUTE and never as writer, and the gRPC refusal names the leader's HTTP address with the "use its gRPC port" caveat. Both degrade to what they did before the address could be resolved at all. A WARNING naming the field to declare is logged once per protocol, next to the existing derive warning.

The guard lives in the resolver **both** protocols share, as the issue asked - a Bolt-only or gRPC-only guard would be exactly the divergence #6091 removed. It is deliberately *not* applied to the HTTP addresses behind `getReplicaAddresses()` / `getStats()`: those feed cluster reporting and peer-to-peer snapshot transfer, where withholding a best-effort address does harm and nothing auto-redirects onto it.

## 2. Every leader-only refusal names the leader, not just `graphBatchLoad`

`graphBatchLoad` was the only RPC that built the redirect trailers, by hand. A `ServerIsNotTheLeaderException` arising anywhere else fell through `GrpcErrorMapper` as a plain `NeedRetryException`: `ABORTED`, no address, so a caller could not tell "the leader is unknown, wait for the election" from "the leader is known and nobody told you".

`GrpcErrorMapper.toStatusRuntimeException(t, prefix, ha)` now attaches `LEADER_GRPC_ADDRESS` / `LEADER_HTTP_ADDRESS` for that exception wherever it arises, and answers **`FAILED_PRECONDITION`** rather than `ABORTED` - retrying the call as it stands means asking the same follower again; it is the same reading the HTTP protocol takes when it answers 400 rather than 503. The two-argument overload stays for callers with no cluster in reach, and falls back to the address the exception itself carries.

`ArcadeDbGrpcService.notTheLeader` collapses to raising the engine's own exception through that mapper - it now chooses only the wording. The design question the issue flagged (`GrpcErrorMapper` is static and holds no server reference) is answered by passing the plugin as an argument rather than by installing static mutable state, which would be wrong in a JVM that runs several servers in-process, as the tests do.

**What the issue's premise got wrong, found while testing it:** a statement sent to a follower does not normally produce a refusal at all. `RaftReplicatedDatabase.command` *forwards* a DDL or non-idempotent statement to the leader instead of rejecting it, so `CREATE VERTEX TYPE` over `executeCommand` on a follower succeeds and replicates - `Issue6183FollowerCommandRoutingIT` pins that, so a future change turning forwarding into a refusal is noticed here rather than by a user. What the mapper change covers is the leadership change landing between that forwarding decision and the schema write (the second `isLeader()` check inside `recordFileChanges`). For that case `executeCommandInternal` rethrows instead of returning `success=false`: a redirect can only travel on an error's trailers, and the response envelope has nowhere to put it. Every other failure keeps the envelope, so a client reading that flag is unaffected.

## Tests

- `Issue6183AmbiguousRoutingTest` (unit, 7 cases): distinct addresses, all-derived collision, declared-beats-derived both ways round, two declared claims, ambiguous followers under a clean writer, single node.
- `Issue6183AmbiguousRoutingIT`: an in-process cluster advertises nothing for **either** protocol. Replaces `Issue6091GrpcRoutingTableDerivedIT`, which asserted the old (wrong) derived answer; the positive derive - distinct hosts, one port, i.e. a StatefulSet - stays unit-tested, since an in-process cluster cannot produce distinct hosts.
- `Issue6091GrpcRoutingTableIT.theBoltTableIsNotTheGrpcTable` now asserts the Bolt table is `null` on a cluster that declares only gRPC ports - a stronger separation check than the old "the two differ".
- `Issue6183LeaderRedirectMapperTest` (unit, 7 cases): status, both trailers, HTTP fallback, suppressed routing table, unknown leader, no plugin, blank addresses.
- `Issue6183FollowerCommandRoutingIT`: DDL to a follower is forwarded and replicated; an ordinary command failure still comes back in the response.
- Regression: full `grpcw` suite (179), `ha-raft` unit suite (870), `Bolt5002RoutingTableIT` + `BoltProtocolIT` (83), `Issue6091GraphBatchLoadLeaderRedirectIT`, `grpc-client` mapping tests and `ErrorHandlingIT`.

## 3. Docs

`docs/release-26.9.1.md` gets the section here. The user-facing reference (the `bolt:`/`grpc:` rows of the object-form table, a new *Client Routing Ports* section, the BOLT note, and `grpc.port` now being a registered setting) is a companion PR on the docs repo.


lvca:
Thanks - both actionable points are in `8cdb212e8`.

**1. Scope of the redirect claim.** Correct, and the phrasing was doing real harm: it invited a reader to assume a handler is covered when it is not. The mapper's javadoc and the release-notes section now name the RPCs that report through it (`executeCommand`, `createRecord`, `beginTransaction`, `commitTransaction`, `graphBatchLoad`), say that the handlers assembling a `Status` themselves cannot raise the exception today - they neither check leadership nor mutate schema - and say what a handler that grows either would have to do to stay redirectable. The PR title is narrowed to match.

I did not route the other handlers through the mapper: that would restate every one of their status choices (`NOT_FOUND` for a missing RID, `RESOURCE_EXHAUSTED` for the result cap, and so on) to fix a condition none of them can reach, which is a bigger and riskier change than the one this issue asked for. If any of them ever grows a leadership check, the mapper is where its errors should go.

**2. The double `declaredRoutingAddresses` read.** Fixed, and it is worth slightly more than the redundancy: the flag saying "this peer was declared" and the address itself have to describe the same map, and reading it twice is exactly how those two drift apart later. It is read once in `getRoutingTable` and passed into `resolveRoutingAddress`.

Re-ran after the change: `Issue6183AmbiguousRoutingTest` (7), `RaftHAServerAddressParsingTest` (55), `RaftHAServerTest` (45), `Issue6183AmbiguousRoutingIT`, `Issue6091GrpcRoutingTableIT`, `Issue6183LeaderRedirectMapperTest` (7), `GrpcErrorMapperTest` (11), `Issue6183FollowerCommandRoutingIT`, `Issue6091GraphBatchLoadLeaderRedirectIT` - all green.

lvca:
Both fixed in `708ded469`.

**Dangling `@link`.** Real miss - deleting the class and leaving the cross-reference behind is worse than never having written it, since it sends a reader hunting for a file that is not there. `Issue6091GrpcRoutingTableIT`'s class doc now names what actually covers the other half, and splits it the way the coverage is actually split: `Issue6183AmbiguousRoutingTest` for the derive itself, `Issue6183AmbiguousRoutingIT` for what a cluster of same-host peers ends up advertising, which is nothing.

**`ha()` at the `graphBatchLoad` check.** Replaced. The helper was extracted for exactly that expression and then the one pre-existing copy was left in place.

Re-ran: `Issue6091GrpcRoutingTableIT` (3), `Issue6183AmbiguousRoutingIT`, `Issue6183AmbiguousRoutingTest` (7), `Issue6091GraphBatchLoadLeaderRedirectIT`, `Issue6183FollowerCommandRoutingIT` (2), `Issue6183LeaderRedirectMapperTest` (7) - all green.

lvca:
Both nits fixed in `7f9119fb3` - the first is worth more than a nit, since it is stale comment debt this PR created.

**`BoltNetworkExecutor.handleRoute`.** Right: `table == null` used to mean one thing and now means two, and the comment only described the old one. A reader hitting READ/ROUTE-with-no-writer on a cluster that plainly *has* a leader would have found the code saying the opposite of what happened. The branch now names both reasons - no leader known yet, or a leader whose address cannot be told apart from a follower's because the nodes share a host and no `bolt:` port was declared - and notes that the answer is the same either way.

**`notTheLeader` static.** Reverted. It takes the plugin as a parameter and touches no instance state; it lost `static` incidentally when the body collapsed onto the mapper.

Re-ran the modules both touch: `Bolt5002RoutingTableIT` (3), `BoltProtocolIT` (80), `Issue6091GraphBatchLoadLeaderRedirectIT`, `Issue6183LeaderRedirectMapperTest` (7), `GrpcErrorMapperTest` (11) - all green.

On CI: the local runs behind each of these rounds cover the affected modules (`ha-raft` unit suite 870, full `grpcw` 179, `bolt` routing 83, `grpc-client` mapping), so the checks here should agree.

lvca:
Addressed in `162923137`, though the conclusion went the other way - worth writing down because the premise is an easy one to hold.

**The `+1` slot.** The note reasons from "`peers.size()` already includes the leader", and that is exactly the assumption that does not always hold here. `raftGroup` is `final`, built once in the constructor from what `HA_SERVER_LIST` parsed into; runtime membership changes go through `clusterManager.addPeer` / the Kubernetes auto-join and reconfigure Ratis without touching that field. The leader, meanwhile, comes from live Ratis state via `getLeaderId()`. So a leader that joined at runtime is *not* in `raftGroup.getPeers()`, the writer occupies a slot past the end of it, and `resolved` reaches `peers.size() + 1`.

Sizing the array to `peers.size()` would therefore put an `ArrayIndexOutOfBoundsException` on the Bolt ROUTE path in precisely that window - a crash on a live cluster, to save one reference. The slot stays; what was missing was the sentence saying why, which the code now carries.

**CI.** Green on `7f9119fb3`, which carries all the code in this PR (`162923137` is comment-only): `build-and-package`, `builder-tests`, CodeQL, Codacy, Meterian, and the java/js/csharp/go/python e2e suites all passed.

Re-ran locally after the change: `Issue6183AmbiguousRoutingTest` (7), `Issue6183AmbiguousRoutingIT`, `Issue6091GrpcRoutingTableIT` (3) - green.

lvca:
Good call - done in `4ee47f879`. The status change was explained inside the rationale, where someone scanning for what an upgrade breaks would not look for it. It is now an explicit upgrade note, and it bounds the blast radius rather than just declaring the change: a client reading the `arcadedb-exception-class` trailer (the bundled Java client, and the reason the trailer exists) is unaffected; one switching on the status code has to treat `FAILED_PRECONDITION` plus a leader-address trailer as "dial that address and repeat it" rather than "retry this call", which is what it should have been doing anyway since retrying asks the same follower again. `graphBatchLoad` already answered `FAILED_PRECONDITION` and is unchanged, and on the other RPCs the condition is only reachable through the leadership-change window, so almost nothing is looking at the old code today.

On the `getStats()`/`getReplicaAddresses()`/`getRoutingTable()` seam this module's CLAUDE.md warns about: agreed it is worth a second pair of eyes, and it is the reason the filter is where it is. Leader exclusion - what "replica" means - is untouched in all three; the new filter is orthogonal to it and deliberately confined to the one method whose output a client dials unattended. The other two feed cluster reporting and peer-to-peer snapshot transfer, where withholding a best-effort address costs something and nothing auto-redirects onto it.

That is every point from all five review rounds addressed. Ready for merge from my side.

PR #6193, opened 2026-08-15
fix(#6177): a periodic materialized view can no longer stop refreshing silently, and its test polls instead of sampling once

### LINKED ISSUE #6177: MaterializedViewEdgeCaseTest.periodicRefreshUpdatesView is a wall-clock flake (200ms interval, single 1s sleep)
Filed from the CI triage of PR #6170: the only failure in the whole `unit-tests` lane on that commit, and not caused by it.

```
com.arcadedb.schema.MaterializedViewEdgeCaseTest.periodicRefreshUpdatesView
expected: 2L but was: 1L   (MaterializedViewEdgeCaseTest.java:252)
```

The test creates a PERIODIC materialized view with a 200 ms refresh interval, inserts a row, sleeps once, and asserts the refresh has happened:

```java
.withRefreshMode(MaterializedViewRefreshMode.PERIODIC)
.withRefreshInterval(200)
...
// Wait for the scheduler to trigger a refresh (interval = 200ms, wait 1s)
Thread.sleep(1_000);

try (final ResultSet rs = database.query("sql", "SELECT FROM PeriodicView")) {
  assertThat(rs.stream().count()).isEqualTo(2);
}
```

Five refresh periods of slack sounds generous, but the assertion is a single sample at a fixed instant: on a loaded runner (this lane runs the whole reactor) the scheduler thread can miss all of them, and the test then fails on a condition that would have been true a moment later. Ran the class 3x locally: green each time - which is exactly the signature of a wall-clock flake rather than a defect.

## Suggested shape

Poll to a deadline instead of sampling once - re-query until the view holds 2 rows or a few seconds have elapsed, failing with what it saw. That keeps the test's meaning (the periodic scheduler does refresh without being asked) while removing the dependency on when the sample lands. Worth sweeping the rest of `MaterializedViewEdgeCaseTest` for the same pattern while there.

If the class turns out to be slow once it polls, `@Tag("slow")` routes it to the lane where a few seconds cost nobody anything.

lvca:
Closes #6177.

## The flake

`MaterializedViewEdgeCaseTest.periodicRefreshUpdatesView` created a PERIODIC view with a 200ms refresh interval, inserted a row, slept once for 1s and asserted the view held 2 rows. Five refresh periods of slack is generous right up until the runner is loaded enough (this lane runs the whole reactor) that the scheduler thread is not run in any of them, and the test then fails on a condition that is true a moment later.

It now polls to a 60s deadline and returns as soon as the condition holds, so the ceiling costs nothing on a healthy run. Two things improve beyond just removing the flake:

- the failure message carries what it saw - how many scheduled refreshes ran, the view status, the error count - instead of only `expected: 2L but was: 1L`, which is what made triaging the original CI failure cost a round trip;
- it asserts the refresh counter actually moved, so the test can no longer pass on a PERIODIC view that silently degraded into being refreshed by something else.

The rest of the class is MANUAL/INCREMENTAL and has no other wall-clock dependency; `MaterializedViewConcurrentRefreshTest` already polls to a deadline.

## The two ways "not yet" could have been "never"

Chasing what could make a periodic refresh actually not happen turned up two silent single points of failure, both on the same theme - the view stops refreshing and keeps reporting the contents of its last successful pass:

- **`MaterializedViewScheduler`'s tick caught `Exception`.** `scheduleAtFixedRate` cancels a task PERMANENTLY the first time it throws - no further execution, ever, and nothing says so. Anything not an `Exception` escaping the tick therefore stopped that view refreshing for the life of the database. The tick body moves into `runOneRefresh()`, guarded against `Throwable` and callable from a test without waiting out a tick - the shape `HAReplicationMetrics.runGuarded` already uses for the same hazard.

- **`MaterializedViewRefresher.fullRefresh` released refresh ownership only for `Exception`.** It owns the view's refresh state machine from `tryBeginRefresh()` to one of its two releases, both of which sat inside `catch (Exception)`. A pass failing with an `Error` left the view latched in RUNNING for good: every later refresh - periodic, manual or incremental - found the machine busy and either coalesced onto a pass that had already died, or returned having done nothing. The failed pass is still counted, still logged and still rethrown to the caller; what changes is that the next one can run.

## Tests

New, and each verified to fail with the production change reverted (except the first, which is a regression guard for the `Exception` case that already worked):

- `MaterializedViewSchedulerTest.aRefreshThatKeepsFailingKeepsTheScheduleAlive` - a view whose every pass fails keeps its `ScheduledFuture` alive and keeps ticking.
- `MaterializedViewSchedulerTest.anErrorThrownByARefreshDoesNotEscapeTheTick`.
- `MaterializedViewConcurrentRefreshTest.aPassThatFailsWithAnErrorStillReleasesOwnership`.

`ThrowingRefreshView` is the shared fixture for the last two.

Run green: all 7 `MaterializedView*Test` classes (65 tests) including the `slow` lane, the full `com.arcadedb.schema` package (424 tests), and `HTTPMaterializedViewIT`, `Issue3941AsyncRefreshMaterializedViewIT`, `RemoteMaterializedViewIT` under `-Pintegration`. The flaky test was run 8 times consecutively; the class now takes ~0.8s where the single sleep alone cost 1s.

## Follow-up worth its own issue

`MaterializedViewScheduler` holds the database through a `WeakReference` so that an abandoned (never closed) database lets its periodic task self-cancel. Once a refresh has run, the scheduler thread's `DatabaseContext` ThreadLocal holds a strong reference to that database, and a live thread is a GC root - so `dbRef.get()` can never return null again and that path is unreachable in practice. `TimeSeriesMaintenanceScheduler.runMaintenance` removes its context in a `finally` for a related reason. Not touched here: it is a different defect from the one this issue is about.

### DIFF
--- a/engine/src/main/java/com/arcadedb/schema/MaterializedViewRefresher.java
+++ b/engine/src/main/java/com/arcadedb/schema/MaterializedViewRefresher.java
@@ -55,7 +55,13 @@ public class MaterializedViewRefresher {
         view.setStatus(MaterializedViewStatus.VALID);
       } while (view.finishRefreshPassAndCheckPending());
 
-    } catch (final Exception e) {
+    } catch (final Throwable e) {
+      // Throwable, not Exception: refresh ownership is held from the tryBeginRefresh() above, and the only
+      // places that give it back are the two releases in this method. Letting an Error past them latched the
+      // view in RUNNING for the life of the database - every later refresh, periodic or manual or incremental,
+      // then found the state machine busy and either coalesced onto a pass that had already died or returned
+      // having done nothing, so the view silently froze at its last successful snapshot. The pass is still
+      // reported as failed and the Error still propagates to the caller; what changes is that the next one can run.
       view.recordRefreshError();
       view.setStatus(MaterializedViewStatus.ERROR);
       // Ownership is released here rather than in a finally: the success path already released it,
--- a/engine/src/main/java/com/arcadedb/schema/MaterializedViewScheduler.java
+++ b/engine/src/main/java/com/arcadedb/schema/MaterializedViewScheduler.java
@@ -59,16 +59,34 @@ public class MaterializedViewScheduler {
         cancel(view.getName());
         return;
       }
-      try {
-        MaterializedViewRefresher.fullRefresh(db, view);
-      } catch (final Exception e) {
-        view.setStatus(MaterializedViewStatus.ERROR);
-        LogManager.instance().log(this, Level.SEVERE,
-            "Error in periodic refresh for view '%s': %s", e, view.getName(), e.getMessage());
-      }
+      runOneRefresh(db, view);
     }, interval, interval, TimeUnit.MILLISECONDS);
 
     tasks.put(view.getName(), future);
+  }
+
+  /**
+   * Runs one periodic pass, and lets nothing out.
+   * <p>
+   * {@code scheduleAtFixedRate} cancels a task PERMANENTLY the first time it throws - no further execution, ever,
+   * and nothing says so - so anything escaping here would stop this view refreshing for the life of the database
+   * while it kept reporting the status of its last successful pass. {@code Throwable} rather than {@code Exception}
+   * because "this view silently stopped refreshing" is a strictly worse outcome than a logged failure, whatever
+   * class the failure belongs to, and because a guard that has to be re-argued every time the refresh path changes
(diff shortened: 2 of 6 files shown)


lvca:
Follow-ups found while working on this, now filed as #6203: the full refresh is not atomic (a failed refresh leaves the view empty rather than stale), and the scheduler's `WeakReference<Database>` self-cancel path is unreachable because `DatabaseContext` pins the database on the scheduler thread.

Issue #6196, opened 2026-08-15
A placeholder whose CONTENT record had to spill into chunks is returned TWICE by a scan


lvca:
Found while building the guard for #6178, and confirmed on unmodified `main` (50eb9613f) - nothing in #6141 /
#6175 / #6178 causes it.

## What happens

A `SELECT` over a type whose bucket holds a placeholder-backed record returns that record's content **twice**,
under two different RIDs, when the placeholder's CONTENT record was large enough that no page could host it whole.

```
select count(@rid) as c from Surrogate where v = <the 200 KB value>   -->  2
```

Reproduced by `Issue6178ChunkCollapseTest.aPlaceholderContentStoredAsAChainIsNeverCollapsed`, which asserts the 2
deliberately (with a pointer here) so that fixing this turns that line red instead of leaving it quietly wrong.

## Why

A content record is normally recognised by the NEGATIVE size marker `createRecordInternal(record,
isPlaceHolder=true, ...)` writes, and every reader that walks a page skips it on that basis - `LocalBucket.scan`
returns a slot only for `recordSize[0] > 0`, `RECORD_PLACEHOLDER_POINTER` or `FIRST_CHUNK`.

But when the content record does not fit a page, that same method spills it through `writeMultiPageRecord`, which
writes `FIRST_CHUNK` - the marker of an ordinary multi-page record. The "this slot is somebody's content, not a
record" information is lost at that moment, and there is nowhere else it is written down: the head chunk of a
placeholder's content is indistinguishable from the head chunk of a real record.

So `scan()` hands it out as a document of its own, in addition to handing out the same bytes through the
placeholder POINTER that references it. `check()` has the mirror of the same confusion: it counts such a record
under `totalMultiPageRecords` and never under `totalSurrogateRecords`.

## How to reach it

Both halves are needed: a page with a free tail of exactly zero (since #6149 the only shape that still produces a
placeholder pointer - see `BucketPageLayoutTestSupport.sealFirstPage`) and a content record bigger than a page.
`sealFirstPage` plus a 200 KB value does it. Databases written before #6149 contain placeholders far more freely,
so the population is larger there.

## Worth considering

The marker namespace has room: a `FIRST_CHUNK_PLACEHOLDER_CONTENT` marker (the chunk head of a content record)
next to `FIRST_CHUNK`, written by `writeMultiPageRecord` when `isPlaceHolder`, and read wherever `FIRST_CHUNK` is
[...]

PR #6197, opened 2026-08-15
fix(#6141, #6175, #6178): the three record-shape defects the #6129 measurement left open

### LINKED ISSUE #6141: A concurrent update of the same placeholder-backed record is silently lost, with no conflict raised

### LINKED ISSUE #6175: A continuation chunk placed on the head chunk's own page poisons the page the disjoint-slot merge was built for
Found while working on #6163/#6154 (PR #6170), by reading the placement path rather than from a failure.

## What happens

`LocalBucket.writeMultiPageRecord` places its first continuation chunk with:

```java
final PageAnalysis pageAnalysis = findAvailableSpace(currentPage.pageId.getPageNumber(), spaceNeededForChunk, txPageCounter, true);
```

and `findAvailableSpace` starts with "PRIORITIZE SPACE IN THE SAME PAGE" - so the chunk is *preferentially* placed on the page that holds the record's HEAD chunk. When it lands there, the loop poisons that page:

```java
if (poisonSlots)
  slotTx.poisonSlotRebasePage(fileId, nextPage.pageId.getPageNumber());
```

The head chunk's own page is exactly the page #6129 went to some trouble to keep mergeable: a chunked record's update is a single-slot write on it, and both `updateRecordInternal` and `rebaseRecordOnPage` carry dedicated machinery (`SLOT_KIND_FIRST_CHUNK`, `chunkChainTailFingerprint`, and since #6163 the region re-derivation) so that concurrent writers to *other* records on that page do not conflict. A continuation chunk landing there throws all of it away for that transaction, and the code already acknowledges it in passing ("including the head chunk's own page when the chain comes back to it").

## When it is reachable

Not when the spilling record is the page's LAST one - the head chunk takes the whole free tail, so nothing else fits. It is reachable when the head chunk is in the MIDDLE of its page and the page still has tail: the chunk chain then prefers that tail over any other page.

That is the common shape for a bucket whose records are mostly small with a few large ones - which is also the shape where the merge matters most, because many unrelated records share the page.

## Measurement first

Before changing anything, count it: on a workload of concurrent updates to a single-bucket type with a few chunked records, how many commits are poisoned by a continuation chunk landing on a head chunk's page (versus poisoned for other reasons, versus not poisoned)? #6129 was planned off exactly this kind of count (280/320 conflicts, all on page 0) and it changed the plan.

## Possible fix

Pass a "do not choose this page" hint down to `findAvailableSpace` for the FIRST continuation chunk, the way `multiPageRecord` already suppresses slot 0. The locality argument for same-page placement is weak for a continuation chunk - the chain is walked by pointer, and the head chunk's page is already pinned and read - while the merge cost is concrete.

Worth checking the same question for `updateMultiPageRecord`'s allocation loop, which has the identical call.

### LINKED ISSUE #6178: A record that shrinks back inside its slot stays a chunk chain for ever - nothing ever un-spills it
Deliberately left out of scope in PR #6170 (#6163), filed so the decision is recorded rather than forgotten.

## What happens

Once a record has spilled, it is a chunk chain for the rest of its life. When it shrinks back to something its slot's own region can hold, `updateMultiPageRecord` frees the continuation chunks and leaves a chain of exactly one chunk - a `FIRST_CHUNK` marker, a 4-byte size, an 8-byte next-chunk pointer that is 0, and the content. It is never turned back into a plain record, however small it becomes.

After #6163 this is much more common than it used to be: the head chunk now takes back the room its region has, so a record that oscillates spends much of its life as a single-chunk chain rather than as a chain spread over pages.

## What it costs

- **13 bytes per record**, permanently, over what a plain record of the same content needs.
- **Every read and write takes the chunk path**: `getRecordInternal` walks a chain, `updateRecordInternal` goes through the `FIRST_CHUNK` branch, the disjoint-slot merge uses `SLOT_KIND_FIRST_CHUNK` plus `chunkChainTailFingerprint` (a walk of a chain that has nothing past its head, to produce `EMPTY_CHUNK_TAIL_FINGERPRINT`) instead of the plain-record path.
- **`bucket.check()` reports it as a multi-page record** for ever, which makes the layout statistics say something that is no longer true - and those statistics are what several tests assert on.

None of it is a correctness problem. It is a shape that no longer matches the data.

## Why it was not done in #6170

Un-spilling changes the slot's SHAPE (chunk head -> plain record), which is a different kind of write from the ones the merge tracks today: the pre-image is a chunk image and the final image a plain record, so it needs its own `SLOT_KIND_*` and a replay that puts the plain marker back - the mirror of `SLOT_KIND_RECORD_SPILLED_TO_CHUNK`, which exists for exactly the opposite transition. That is a self-contained piece of work and did not belong in a fix for the sizing rule.

## Suggested approach

In `updateMultiPageRecord`, when the new content fits the slot's region as a PLAIN record (`getNumberSpace(size) + size <= regionEnd - recordPositionInPage`) and the chain has no chunk past the head, write it as a plain record, free the chain, and track it as a new `SLOT_KIND_CHUNK_COLLAPSED_TO_RECORD` whose replay re-checks the committed region exactly as the spill's replay does. The region arithmetic and the replay symmetry both already exist since #6163; this is the third transition between the same two shapes.

Measure first: how often does a spilled record shrink back that far in practice? If the answer is "almost never outside tests", the 13 bytes are not worth a new slot kind, and this should be closed with that written down.

lvca:
Three defects in bucket record storage, all spawned by the #6129/#6149/#6163 chain, each with its own commit,
tests and measurement.

## #6141 - a concurrent update of a placeholder-backed record was silently lost

Two transactions updating the same record both committed and the second silently dropped the first, with no
`ConcurrentModificationException` anywhere, when the record's content sat behind a placeholder POINTER.

A record update is deferred: `addUpdatedRecord` pins the record's OWN page at `save()` time and the write runs at
commit, after the file lock - so every other page it touches is loaded fresh at the newest committed version and
can never fail a version check. For a placeholder-backed record that is the whole record: the pinned page holds 8
bytes of pointer that an absorbed update never touches (and a modified page with an empty modified range is
dropped at commit), while the content page is pinned by nobody.

The chunk-chain fingerprint #6129 introduced answers exactly this question for exactly this reason, so it is
generalised rather than duplicated: `chunkChainTailFingerprint` becomes `offPageContentFingerprint` - "the part of
the record that does not live in its own slot" - and covers the placeholder's content record too. Two deliberate
differences: it is captured whether or not the slot merge is enabled (for this shape it is the only detection
there is), and a mismatch raises a retryable `ConcurrentModificationException` rather than poisoning a page.

Both arms are covered (the update the content absorbs, and the one that rebuilds the slot), plus the
counter-check that an unrelated concurrent update - including one on the content record's own page - still
commits, and that the protection survives `TX_PAGE_SLOT_MERGE=false`. Two of the four tests fail on `main`.

## #6175 - a chain came home and poisoned the page the merge was built for

Measured before changing anything, by logging every continuation-chunk placement of a 9-byte record grown to
200 KB while sharing its page with one neighbour (64 KB pages):

```
chunk 1 -> page 1 (new)
chunk 2 -> page 2 (new)
chunk 3 -> page 0   <-- the head chunk's own page, from the statistics scan at half the requested size
chunk 4 -> page 3 (new)
```

A continuation chunk poisons the page it lands on, so this cost the one page #6129 built `SLOT_KIND_FIRST_CHUNK`,
the chain fingerprint and the region re-derivation to keep mergeable - and a concurrent update of the untouched
neighbour became a hard conflict. `findAvailableSpace` now takes a page to avoid and the two chunk-allocation
loops pass the head chunk's page, but only when the caller declared that slot replayable: when it did not, the
page is unmergeable anyway and refusing its tail would cost space for nothing (which also leaves record creation
allocating exactly as before). Only that one page is excluded - rewriting a chunked record ten times leaves the
bucket's page count where it was.

## #6178 - a record that shrank back was a chunk chain for ever

`collapseChunkChainToRecord` writes it back as a plain record and frees the chain when the content fits the
region the slot owns, with a slot kind of its own (`SLOT_KIND_CHUNK_COLLAPSED_TO_RECORD`) so the merge replays it.

**This removes the state #6163's grow-back exists to repair rather than compensating for it**, which is the one
result worth reading before the diff: content that fits the head chunk (13 bytes of header) fits the region as a
plain record (at most 5 of size marker), so the two conditions coincide and a head chunk that declares less than
its region no longer survives an update. Measured by instrumenting the grow-back - with the collapse disabled
`Issue6163HeadChunkRoomTest` alone drives it 12 times, every one from a head chunk ratcheted down to 11 bytes;
with the collapse enabled, zero. Two of that class's tests are updated to assert what is now true, with the
reason written into them; the #6163 machinery stays as the defence for a region that grows for another reason.

The CONTENT record of a placeholder is deliberately never collapsed - it is identified by the negative size
marker a collapse would replace with a positive one. Building that guard turned up a pre-existing defect of the
same family, filed as #6196 and asserted in the test as it currently behaves rather than papered over: a
placeholder whose content had to spill into chunks is already returned TWICE by a scan on unmodified `main`,
because `createRecordInternal` writes `FIRST_CHUNK` where the negative marker would have gone.

## Verification

- 4 + 3 + 5 new tests; every one that pins a fix was checked to go red without it (mutation-verified), and the
  placement-dependent ones prove their premise by re-running with `TX_PAGE_SLOT_MERGE` off and requiring failure.
- `engine` unit lane over `engine/**`, `database/**`, `graph/**`: 1467 tests, one unrelated failure
  (`TimeSeriesTagDictionaryTest.anIdWrittenByAnotherInstanceIsResolvedByReloading`, a `PaginatedComponent` that no
  code here touches; passes in isolation and in its own package - reported, not fixed).
- `engine/CLAUDE.md` and `docs/release-26.9.1.md` updated for all three.

Closes #6141, closes #6175, closes #6178.


lvca:
Thanks - all three minor points addressed, two with a measurement rather than an assurance.

**1. Allocation in `placeholderContentFingerprint`.** Valid, and fixed at the root rather than for that one
caller. The FNV byte loop existed twice; it is now one `fnv1a(fingerprint, page, offset, length)` that reads the
page one **absolute** byte at a time, so `walkChunkChainTail`'s page-sized scratch buffer goes too and a chain of
any length walks allocation-free as well. The loop has to touch every byte either way - what the copy bought was
a bulk read, what it cost was a content-sized allocation on a path that runs twice per update. The absolute reads
also leave the page buffer's position where they found it, which the bulk read (`Binary.getByteArray` does
`buffer.position(index)`) did not.

**2. No test for the #6175 x #6178 interaction.** There is one, and it turns out to be the interaction that
matters - I had not said so. Freeing a chain poisons every page it visits, so a chain that had come home to the
head chunk's own page would poison the very page the collapse replay needs. Measured on
`theCollapseIsReplayedOnACommittedPage`'s own fixture: with the exclusion disabled the chunks land on pages
**1, 2, 0, 3** (chunk 3 on the head page); with it, **1, 2, 3, 4**. While there I also checked the test is not
vacuous in the way that would have mattered most - instrumented, it really reaches `rebaseRecordOnPage` with
`SLOT_KIND_CHUNK_COLLAPSED_TO_RECORD`, so the new replay is what it passes through and not some other path. Both
facts are now written into that test's javadoc.

**3. The `TimeSeriesTagDictionaryTest` failure.** Tracked as #6198, with the read that produces the `null`
identified rather than left as "flaky": `resolveMiss` declines to reload when `id > peekStoredEntryCount()`, and
`peekStoredEntryCount()` answers `entryCount` (0 on a follower that has loaded nothing) whenever
`getTotalPages() == 0` - so the self-heal is disabled in exactly the state it exists for. Independent of this PR;
the component is a `PaginatedComponent` that nothing here touches.

On the boxed `!=`: agreed it is worth a second look every time. `capturedContent != placeholderContentFingerprint(...)`
has a primitive `long` on the right, so it unboxes and compares by value; I kept the existing
`isChunkChainTailUnchanged` shape (`captured != null && captured == current`) for the merge side for the same
reason.

lvca:
CI outcome for `5e96b2fa8`, since two lanes are red and neither is this PR:

**`unit-tests`, `slow-unit-tests`, `vector-unit-tests`, `ha-integration-tests`, `build-and-package` and the four
e2e lanes all pass.** That is the set that covers the engine changes.

**`integration-tests`** fails on
`Issue6075SnapshotBackupIT.backupUnderConcurrentWritersRestoresToAPointInTime[2]` (the `snapshot=false`
suspend-and-freeze variant): the restored database reports 20000 records where the live one reported 20002 before
the backup started. Both numbers come from `countType(..., false)`, which answers from the bucket's **cached
record counter** - `statistics.json` in the archive - not from a scan, so what the assertion caught is the counter
in the archive lagging the pages it was archived with, not records missing from the backup. Nothing in this PR
writes that counter or `statistics.json`: the free-space accounting it does touch (`updatePageStatistics`) is the
per-page free-space map, a different number. Run twice locally on this branch: 4/4 green both times. For
comparison, the same lane on `main` is red on a different test (`Issue5470BatchStreamStallIT`), which is the
known-red one.

**`codecov/patch`** reports 66.16% of the diff hit against a 69.96% target. Two things about that number: the
failing integration lane uploaded no coverage for this run, and what is uncovered in the diff is the defensive
arm of each new branch - the `NO_OFF_PAGE_FINGERPRINT` guards for a page that is gone or a slot that no longer
holds what the pointer says, and the three refusals in the `SLOT_KIND_CHUNK_COLLAPSED_TO_RECORD` replay.

I looked at whether the most valuable of those - the replay refusing a region a concurrent commit shrank - can be
reached, and it cannot, for a reason worth having in the code rather than in a review thread: a region only
shrinks when something is committed between that slot and its neighbour, and the room next to a chunk head is
released by the very transaction being replayed (a collapse), which by definition has not committed yet. A
`compressPage` re-flowing the page under us is what is left. Reaching it would mean hand-building a page state
the engine cannot produce, which tests the fabrication rather than the engine - the same call #6153 recorded when
it declined a test for a free tail of 1..7 bytes. `b16c2893d` writes that reasoning next to the refusal so the
next reader does not go looking for the missing test.

PR #6207, opened 2026-08-15
fix(#6090): CHECK DATABASE reports an orphan edge record, and can reclaim one on request

### LINKED ISSUE #6090: CHECK DATABASE cannot report an orphan edge record: missingReferenceBack is silent, imprecise, and never repaired
Split out of #6083 item 2, where establishing this was a precondition for deciding whether the bulk loader had to clean up after itself. It did (fixed in #6089), but the checker gap it exposed is independent and outlives that fix.

## The gap

An **orphan edge record** is an edge record that exists physically in an edge-type bucket, whose `@out`/`@in` name valid vertices, but which no vertex's edge list points back at. `countType()` counts it; no traversal reaches it.

`CHECK DATABASE` *observes* this condition and then effectively discards the finding:

- `GraphDatabaseChecker.checkEdges` does scan every edge record (`database.scanType(typeName, false, ...)`) and probes both endpoints for a back-reference — the right direction, and the scan already exists.
- On a miss it does nothing but `missingReferenceBack.incrementAndGet()` (both call sites are commented `// UNI DIRECTIONAL EDGE`). No `report.warn(...)` naming the record, no `report.corrupt(...)`.
- Because the RID never enters `report.corruptedRecords`, the `FIX` variant's delete loop never touches it. **`CHECK DATABASE FIX` does not reclaim orphan edge records.**
- The counter is surfaced as a bare `missingReferenceBack` number on the result row.

## Why the counter cannot be read as an orphan count

It is not edge-type aware. A perfectly healthy **unidirectional** edge is legitimately absent from its target's IN list and bumps the counter by 1; a genuinely orphaned bidirectional edge bumps it by 2. The published value is one aggregate with no breakdown and no RIDs, so an operator cannot tell "N legitimately unidirectional edges" from "N/2 orphans". On a database using unidirectional edge types at all, the signal is buried.

Worth noting the one thing `FIX` *does* reclaim under a similar name is orphaned edge **segments** (`reclaimOrphanedEdgeSegments`) — the internal `_out_edges`/`_in_edges` linked-list chunks. Different structure, different buckets; it never looks at the edge type's own buckets.

## Why it still matters after #6089

#6089 stops `GraphBatch` from *creating* orphans on a failed flush. It does nothing for:

- databases that already accumulated them under an older build (the bulk loader has existed for many releases, and a failed load left them silently);
- any other route that could produce one;
- giving an operator a way to answer "does my database have any?" — today the honest answer is "run a full traversal yourself and compare against `countType`", which is exactly what `Issue6083OrphanEdgeRecordTest.reachableEdgeCount()` had to do.

## Suggested direction

The scan is already there and already in the right direction, so this is mostly about not throwing the result away:

1. Make the probe **direction-aware**: an edge missing from its OUT vertex's OUT list is always wrong; missing from the IN vertex's IN list is only wrong for a bidirectional edge type. That alone turns the counter into something meaningful.
2. On a genuine miss, `report.warn(...)` naming the edge RID and which side is unreferenced, and count it under a distinct key (e.g. `unreachableEdgeRecords`) so `missingReferenceBack` keeps its current meaning for anyone parsing it.
3. Decide whether `FIX` should delete them. Deleting a record nothing references is safe in the same sense as the existing segment reclaim, but it is a destructive default, so it may be better gated behind an explicit option than folded into the existing `FIX`.

Step 1 is the one that makes the rest possible; 2 and 3 are separable.

## Reproducing one today

Force a `GraphBatch` edge flush to fail after its records are committed on a build **before** #6089 (an edge whose source RID is in a bucket that does not exist does it), then compare `countType(edgeType)` against a full OUT traversal. On a current build, use `getOrphanEdgeRecordsLeaked() > 0` from a reclaim that could not complete — the state is reachable, just no longer produced silently.

robfrank:
Closes #6090

## The gap

An **orphan edge record** is an edge record that exists physically in an edge-type bucket, whose `@out`/`@in` name valid vertices, but which no vertex's edge list points back at: `countType()` counts it, no traversal reaches it.

`GraphDatabaseChecker.checkEdges` already had everything needed to find one - it scans every edge record and probes both endpoints for a back-reference - and then discarded the answer. Both call sites did nothing but `missingReferenceBack.incrementAndGet()`, under a `// UNI DIRECTIONAL EDGE` comment. No `report.warn` naming the record, no entry in `report.corruptedRecords`, so `CHECK DATABASE FIX` never reclaimed one.

The published counter could not stand in for the finding either, because it is not edge-type aware: a healthy **unidirectional** edge is legitimately absent from its target's IN list and bumps it by 1, a genuinely orphaned bidirectional edge bumps it by 2. One aggregate, no breakdown, no RIDs.

## The change

**1. The probe is direction-aware.** Which side is allowed to hold no reference is a property of the edge type:

| condition | defect? | new key |
|---|---|---|
| absent from the OUT vertex's OUT list | always - no outgoing traversal reaches it | `edgesMissingOutReference` |
| absent from the IN vertex's IN list | only when the type is BIDIRECTIONAL | `edgesMissingInReference` |
| absent from **both** | orphan record | `unreachableEdgeRecords` + `unreachableEdgeRecordsFound` |

`isBidirectional()` reads the **record's** own type, not the `typeName` the scan was given, because the `RECORD` scope hands one call edges of several types; it is a field read on the materialised document, not a schema lookup, so it costs nothing on the per-record path.

**2. The finding is reported.** One warning per defective edge, naming the edge RID, the endpoint and which list is missing it. The three counters above are seeded in `DatabaseChecker.check()` so a clean run publishes zeros rather than omitting the keys - "does this database hold any?" has to be answerable from every run's result, not only from one that found some. `unreachableEdgeRecordsFound` is bounded by the same cap as `corruptedRecords`; the count beside it stays exact.

`missingReferenceBack` is **deliberately unchanged**: both probes still run and it is still bumped once per side that holds no reference, so anything parsing today's result keeps reading the same number. Skipping the IN probe for unidirectional types would have been a small win and would have silently changed it.

**3. Reclaiming them is an explicit opt-in: `CHECK DATABASE FIX DELETE ORPHANS`.** Not folded into plain `FIX`, and the reason is a data-loss path rather than a taste for options. The detection cannot distinguish "garbage a failed bulk load left behind" from "this vertex lost its head-chunk pointer, so its perfectly good edges look unreferenced" - a null head chunk reads exactly like a vertex with no edges, and unlike an *unreadable* chain it is never registered for rebuild. In that second case the edge records are the only surviving description of the adjacency and the thing `RESTORE VERTEX` rebuilds it from, so deleting them by default would turn a recoverable database into an unrecoverable one. `DELETE ORPHANS` without `FIX` is refused rather than silently implying it.

Under the opt-in the orphan RID goes through `report.corrupt(...)`, which is what hands it to the existing repair loop **and** puts its bucket into `affectedBuckets` - the delete there is a raw bucket delete, so without that the edge type's indexes would keep pointing at a record that is gone.

## Why the classification is conservative

An edge is called unreachable only when **both** probes positively established the absence (`outProbed && outUnreferenced && inProbed && inUnreferenced`). Two cases deliberately fall outside it:

- a far endpoint that could not be **loaded** - already reported as a dangling link, and its edge is already flagged corrupted;
- a far endpoint whose list could not be **walked** - the existing `catch` leaves it to `checkVertices`, which runs *after* this pass and rebuilds the chain from the surviving edge records. Deleting the record that repair reads from would be the worst possible answer.

The IN-side probe is required even for a unidirectional type: a type created bidirectional and altered since can still have a list naming the edge, and a record something references must never be called an orphan.

## Test plan

- [x] `engine` `Issue6090OrphanEdgeRecordCheckTest` (9 tests) - orphan bidirectional and unidirectional records are named + counted; a **healthy unidirectional edge** is not a finding while still bumping `missingReferenceBack` by 1 (the case that made the old counter unreadable); a **half-linked** edge still reachable from its source is reported as an IN-side defect and *not* as an orphan; plain `FIX` leaves the record alone; `FIX DELETE ORPHANS` removes it and a second run is clean; a reachable edge survives the opt-in; `DELETE ORPHANS` without `FIX` is refused.
- [x] `engine` `Issue6090CheckDatabaseDeleteOrphansParserTest` (3 tests) - grammar accepts the clause in every position (`FIX DELETE ORPHANS`, with `TYPE`, with `RECORD`, before `COMPRESS`), rejects the partial forms, the clause survives `toString` (the statement cache re-parses that), and `orphans` is still usable as an identifier since it is a new lexer token.
- [x] **Proven to fail without the fix**: with only the two test files applied on top of `main`, 9/9 fail (6 failures + 3 errors). With the fix, 9/9 pass.
- [x] Full `engine` unit lane green: **11987 tests, 0 failures, 0 errors** (`-DexcludedGroups=benchmark,slow,vector`).
- [x] Re-ran the directly affected classes after the final polish: `CheckDatabaseTest`, `CheckDatabaseRecordScopeTest`, `CheckDatabaseSinglePassTest`, `CheckDatabaseAutoFixAccountingTest`, `GraphDatabaseCheckerOrphanReclaimTest`, `Issue6083OrphanEdgeRecordTest`, `CheckDatabaseStatementTestParserTest` + both new classes - 62 tests, all green.

## Notes

- `GraphBatch`'s Javadoc on `reclaimOrphanEdgeRecords` documented the old "CHECK DATABASE does not rescue them" state; updated to say what is now true (it reports them; removing is opt-in), so #6089's rationale for cleaning up after itself still reads correctly.
- Release note added to `docs/release-26.9.1.md`.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1938,6 +1938,11 @@ The probe is now **direction-aware**, and what it establishes is reported:
 `missingReferenceBack` is deliberately untouched - both probes still run and it is still bumped once per side
 that holds no reference - so anything parsing today's result keeps reading the same number.
 
+The three new counters are **not disjoint**, which matters if you render them as a summary: one orphaned
+bidirectional edge increments all three. Each answers "how many edges have this defect" independently, so
+summing them double-counts. `unreachableEdgeRecords` is the orphan total; the other two are the per-direction
+detail behind it, plus the half-linked edges that are still reachable from one side.
+
 Reclaiming them is a new, explicit clause: **`CHECK DATABASE FIX DELETE ORPHANS`**. It is not folded into plain
 `FIX`, and the reason is a data-loss path rather than a taste for options. The detection cannot distinguish
 "this record is garbage a failed bulk load left behind" from "this vertex lost its head-chunk pointer, so its
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1910,3 +1910,50 @@ Two things the same method carried:
   parking the single-threaded lifecycle executor.
 
 [#6202](https://github.com/ArcadeData/arcadedb/issues/6202)
+
+## `CHECK DATABASE` now reports an orphan edge record, and can reclaim one on request (#6090)
+
+An **orphan edge record** is an edge record that exists physically in an edge-type bucket, whose `@out`/`@in`
+name valid vertices, but which no vertex's edge list points back at: `countType()` counts it, no traversal
+reaches it. `GraphDatabaseChecker.checkEdges` already scanned every edge record and already probed both
+endpoints for a back-reference - the scan and the direction were both right - and then discarded the answer.
+Both call sites did nothing but `missingReferenceBack.incrementAndGet()`. No warning named the record, nothing
+entered `corruptedRecords`, and so `CHECK DATABASE FIX` never reclaimed one.
+
+The published counter could not stand in for the finding either, because it is not edge-type aware. A
+perfectly healthy **unidirectional** edge is legitimately absent from its target's IN list and bumps it by 1; a
+genuinely orphaned bidirectional edge bumps it by 2. One aggregate, no breakdown, no RIDs - on a database that
+uses unidirectional edge types at all, the signal is buried. The honest way to answer "does my database hold
+any?" was to run a full traversal yourself and compare it against `countType`.
+
+The probe is now **direction-aware**, and what it establishes is reported:
+
+- an edge absent from its OUT vertex's OUT list is a defect for every edge type - no outgoing traversal reaches
+  it - and is counted under `edgesMissingOutReference`;
+- an edge absent from its IN vertex's IN list is a defect only when the type is BIDIRECTIONAL, counted under
+  `edgesMissingInReference`;
+- an edge **neither** list names is an orphan record: warned with its RID and both endpoints, counted under
+  `unreachableEdgeRecords` and named in `unreachableEdgeRecordsFound`.
+
+`missingReferenceBack` is deliberately untouched - both probes still run and it is still bumped once per side
+that holds no reference - so anything parsing today's result keeps reading the same number.
+
+Reclaiming them is a new, explicit clause: **`CHECK DATABASE FIX DELETE ORPHANS`**. It is not folded into plain
+`FIX`, and the reason is a data-loss path rather than a taste for options. The detection cannot distinguish
+"this record is garbage a failed bulk load left behind" from "this vertex lost its head-chunk pointer, so its
+perfectly good edges look unreferenced" - a null head chunk reads exactly like a vertex with no edges. In the
+second case the edge records are the only surviving description of that adjacency and the thing `RESTORE
+VERTEX` rebuilds it from, so a repair that deleted them by default would turn a recoverable database into an
+unrecoverable one. Reporting is therefore always on; removing is always asked for, and `DELETE ORPHANS` without
+`FIX` is refused rather than silently implying it.
+
+The classification is conservative by construction: an edge is called unreachable only when **both** probes
+positively established the absence. A far endpoint that could not be loaded is already reported as a dangling
+link, and one whose list could not be walked is left to `checkVertices`, which runs after this pass and
+rebuilds the chain from the surviving edge records - neither may be answered by deleting the record that repair
+reads from.
+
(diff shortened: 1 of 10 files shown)


robfrank:
**Note on the red `integration-tests` check:** pre-existing on `main`, not introduced here. This branch's merge base is `76555a5750`, and [main's own run for that exact commit](https://github.com/ArcadeData/arcadedb/actions/runs/31893053331) fails the same `integration-tests` job, dying part-way through `com.arcadedb.server.HTTPDocumentIT` - the same place this PR's run stops. The change here touches the engine's graph checker and the SQL grammar; nothing in the `server` module IT lane exercises either.

PR #6208, opened 2026-08-15
fix(#6203): a materialized-view refresh is one transaction, and the scheduler stops pinning an abandoned database

### LINKED ISSUE #6203: Materialized views: a full refresh is not atomic (a failed one empties the view), and the scheduler's WeakReference self-cancel is unreachable
Two defects in the materialized-view refresh path, found while fixing #6177 (PR #6193) and deliberately left out of that PR because neither is a test flake. They are unrelated to each other except that both are silent: nothing in the status, the logs or the row count tells an operator what happened.

## 1. A full refresh is not atomic, and a failed one leaves the view empty rather than stale

`MaterializedViewRefresher.refreshOnce` wraps the truncate and the repopulate in a single transaction:

```java
database.transaction(() -> {
  database.command("sql", "TRUNCATE TYPE `" + backingTypeName + "` UNSAFE").close();
  try (final ResultSet rs = database.query("sql", view.getQuery())) {
    ...  // insert one document per row
  }
}, false);
```

That wrapper buys nothing, because `TruncateTypeStatement` commits *the caller's* transaction from the inside:

* `schema.dropIndex(def.name)` for every index on the backing type, committed before any record is touched;
* `db.commit(); db.begin()` inside the scan callback every `TRUNCATE_BATCH_SIZE` records (default batching added for #4817);
* another `db.commit(); db.begin()` before the index rebuild, so the rebuild sees the empty buckets on disk.

So the refresh is a sequence of committed transactions, not one. Two consequences:

**Readers see a partially built view.** Between the truncate's first commit and the last insert, `SELECT FROM <view>` returns zero rows, or some prefix of the new rows, with nothing to distinguish that from a legitimately empty or small result. The window is the whole runtime of the defining query, which for a view worth materialising is exactly the expensive case. For a PERIODIC view this happens on every tick.

**A failed refresh destroys the data instead of preserving it.** If the defining query throws after the truncate has committed - a schema change, a bad `@rid`, memory pressure, an HA leader change mid-refresh - the backing type is left with zero rows. The view then reports `ERROR` (or `STALE`, once `MaterializedViewChangeListener` overwrites it). Both statuses read as "the data you are looking at is old"; the data is in fact gone, and stays gone until a later refresh happens to succeed. For a MANUAL view there may be no later refresh at all.

### Suggested fix

Not a patch to the current sequence - refresh into a shadow backing type and swap:

1. build into a second, uniquely named backing type (`<view>_backing_<n>`), which no reader can see;
2. on success, atomically repoint the view's `backingTypeName` and drop the old type;
3. on failure, drop the shadow type and leave the previous backing type exactly as it was.

That closes both windows with one mechanism: readers keep seeing the previous snapshot for the entire refresh, and a failure costs a rebuild rather than the data. It also removes the `TRUNCATE ... UNSAFE` drop/recreate of every index on the backing type on each pass, which is where much of a refresh's cost currently goes.

A cheaper stopgap, if the swap is too large a change for one issue: keep the truncate, but document the non-atomicity on `MaterializedView` and make the status distinguish "stale" from "emptied by a failed refresh" so an operator can tell the difference. That is strictly worse - it documents the divergence rather than removing it - and is offered only as a fallback.

## 2. `MaterializedViewScheduler`'s `WeakReference<Database>` self-cancel path is unreachable

The periodic tick guards against a database that was abandoned without being closed:

```java
final WeakReference<Database> dbRef = new WeakReference<>(database);
final ScheduledFuture<?> future = executor.scheduleAtFixedRate(() -> {
  final Database db = dbRef.get();
  if (db == null || !db.isOpen()) {
    cancel(view.getName());
    return;
  }
  runOneRefresh(db, view);
}, interval, interval, TimeUnit.MILLISECONDS);
```

A clean close is already handled elsewhere - `LocalSchema.close()` calls `materializedViewScheduler.shutdown()` - so the weak reference exists solely for the abandoned-database case. It cannot fire. `runOneRefresh` calls `fullRefresh`, which runs a transaction on the scheduler thread, and that installs a `DatabaseContext` entry keyed by the database path holding a strong reference to the database. The scheduler never removes it. After the very first refresh, `dbRef.get()` can never return null, so an abandoned database and its whole page cache stay reachable from the live `ArcadeDB-MV-Scheduler` thread for the life of the JVM.

`TimeSeriesMaintenanceScheduler` hit this and documents it on `runMaintenance`:

> ... installs the target database into the worker thread's `DatabaseContext` ThreadLocal and MUST remove it again in a `finally` block: leaving it installed keeps a strong reference to a (possibly closed) database ...

and does `DatabaseContext.INSTANCE.removeContext(dbInternal.getDatabasePath())` in a `finally`. The materialized-view scheduler should do the same, which makes the existing weak reference work as written. (Only the first half of that Javadoc applies here: the MV executor is created per `LocalSchema`, so there is no cross-database context inheritance to worry about, just the strong reference.)

While in the file: the scheduler is per `LocalSchema`, but its thread is named `ArcadeDB-MV-Scheduler` with no discriminator, so a JVM with several open databases gets several identically named threads and a thread dump cannot say which is which. `TimeSeriesMaintenanceScheduler` at least suffixes a counter (`ArcadeDB-TS-Maintenance-N`); the database name would be more useful still.

### Test

`removeContext` in a `finally` is verifiable without waiting on GC: assert that after `runOneRefresh` returns, the scheduler thread holds no `DatabaseContext` for that database path. The reachability claim itself is better left to a manual heap check than to a `System.gc()`-dependent test.

## Scope

Item 2 is a few lines plus a test and can go in on its own. Item 1 is a design change and should get its own PR and its own review; opening them together only because they were found together and both live in the same two classes.

lvca:
Fixes #6203.

## 1. A full refresh is now one transaction, so a failed one no longer empties the view

A full refresh looked like a single transaction and was not one. `TRUNCATE TYPE ... UNSAFE` commits the caller's
transaction from the inside:

* `schema.dropIndex()` for every index on the backing type, before a single record is touched;
* `commit(); begin()` once per `arcadedb.truncateBatchSize` records;
* another `commit(); begin()` before it rebuilds the indexes it dropped.

Wrapping that in `database.transaction()` bought nothing, and both consequences were silent.

**Readers saw a partially built view.** Between the truncate's first commit and the last insert, `SELECT FROM <view>`
returned zero rows or some prefix of the new ones, with nothing to distinguish that from a legitimately empty or
small result. The window is the whole runtime of the defining query - which for a view worth materialising is
exactly the expensive case, and for a PERIODIC view is every tick.

**A failed refresh destroyed the data.** A defining query that threw after the truncate had committed left the
backing type with zero rows. The view then reported `ERROR`, or `STALE` once `MaterializedViewChangeListener`
overwrote it, and both of those read as "the data you are looking at is old" when the data was in fact gone. For a
MANUAL view there may be no later refresh at all.

`refreshOnce` now writes the defining query's rows over the previous snapshot's records inside one transaction,
deleting only the surplus and creating only the shortfall. The pass then gets the isolation every other ArcadeDB
transaction has: every write stays in the transaction's own page buffer until commit, a concurrent reader sees the
previous snapshot throughout and the new one afterwards, and a failure anywhere rolls the whole pass back onto the
previous snapshot rather than onto nothing. A view's rows also keep their `@rid` across a refresh, where before
every pass gave them new ones.

### Rewriting in place, rather than clearing and repopulating

This is what keeps an index on the backing type from paying for the atomicity, and it makes an indexed refresh
cheaper than it was before. Clearing the view would give every row a new RID, and `TransactionIndexContext`
collapses a REMOVE followed by an ADD only on the same (key, RID), so each pass would leave a full set of real
tombstones where the truncate used to drop the index and rebuild it empty. Measured over 120 refreshes of a
2000-row indexed view, index bytes on disk:

| refresh strategy | after 120 passes |
| --- | --- |
| `TRUNCATE` (drop + rebuild empty), before this PR | **256KB flat** |
| `DELETE FROM` + recreate | **256KB to 3.9MB, still climbing** |
| write rows over the previous records | **256KB flat** |

`DocumentIndexer.updateDocument` finds the key values unchanged for every row a stable view reproduces and skips
the index outright: zero index writes per pass, against a full rebuild per pass before. The backing buckets stay
flat too, since no record is freed and reallocated. A `@Tag("slow")` soak test guards the bound.

The RIDs are collected up front rather than iterated live: an update whose content no longer fits its slot can
relocate the record, and a scan still walking the bucket would then meet it a second time at its new home.

### Why not the shadow-type swap the issue proposed

Building into a second backing type and swapping reads better on paper - the swap is metadata only and the old rows
are dropped as files rather than deleted one by one - but in this engine the swap has no cheap implementation:

* **Renaming is not metadata-only here.** Renaming a type renames its buckets, and `PaginatedComponent.rename()`
  first waits for every page of the *whole database* to be flushed
  (`PageManager.waitAllPagesOfDatabaseAreFlushed`). Paying a database-wide flush barrier twice per refresh, on every
  tick of every PERIODIC view, is a worse defect than the one being fixed. It also leaves a window in which the type
  does not exist at all, so a reader gets an error rather than stale data.
* **Repointing the view at a differently named backing type** avoids the barrier but makes the view's own name stop
  being the type name. That identity is what makes `SELECT FROM <view>` resolve and what every record's `@type`
  reports, so it is a user-visible change, not an internal one.

What the single transaction costs instead is bounded and not new: it holds the pages of the rows it rewrites, and
under HA the pass is one Raft log entry rather than one per truncate batch plus one for the repopulate. The
repopulate was **already** one unbounded transaction, so a view too large to refresh in one transaction was already
too large to refresh.

### Why the existing tests did not catch it

Neither trigger fires on the trivial case: an unindexed view smaller than one truncate batch was already atomic, so
every existing materialized-view test passed. The new tests drive both triggers.

## 2. The scheduler no longer pins an abandoned database to a live thread

`MaterializedViewScheduler` guarded against a database abandoned without `close()` by holding it through a
`WeakReference` and cancelling the task once that cleared. It could not clear, for **two** independent reasons - the
issue names the first:

* The refresh runs a transaction on the scheduler thread, which installs a `DatabaseContext` entry keyed by the
  database path holding a strong reference to the database, and the scheduler never removed it.
  `TimeSeriesMaintenanceScheduler.runMaintenance()` documents the same hazard and takes the entry back down in a
  `finally`. The refresh now does too - removing only a context that pass installed itself, since `runOneRefresh` is
  package-visible so a test can drive one pass inline and dismantling a caller's own context would be a side effect
  nobody asked for.
* The task also captured the `MaterializedViewImpl` strongly, and `MaterializedViewImpl` holds its own database. So
  the reference that was weak was not the only one, and fixing only the `DatabaseContext` would have left the
  mechanism just as unreachable. Both are weak now; the task keeps the view's name as a `String`.

The tick's self-cancel also used a blind `cancel(viewName)`, which after a re-schedule (schema reload,
`ALTER MATERIALIZED VIEW`) would cancel the *replacement* task rather than itself and stop the view refreshing with
nothing saying so. A task now unregisters itself only while it is still the task registered for that view.

Finally, the scheduler is created per `LocalSchema` but its thread was called `ArcadeDB-MV-Scheduler` with no
discriminator, so a JVM with several open databases got several identically named threads and a thread dump could
not say which was which. The database name is now part of it.

## Tests

New `MaterializedViewRefreshAtomicityTest` (all three of the first tests fail on `main`, verified):

* `aFailedRefreshOnAnIndexedViewLeavesThePreviousSnapshotIntact` - the `dropIndex` commit trigger. On `main` the view
  is left with 0 rows; the index is also checked, through a plan asserted to be an index fetch so the assertion
  cannot quietly become a second bucket scan.
* `aFailedRefreshOnAViewLargerThanOneTruncateBatchLeavesThePreviousSnapshotIntact` - the in-scan commit trigger. On
  `main`: `expected: 1100L but was: 100L`.
* `aConcurrentReaderNeverSeesTheViewPartiallyBuilt` - a `BeforeRecordCreateListener` parks the refresh on its first
  insert while the test thread reads the view, so the visibility window is driven exactly rather than raced for.
* `aViewWithAUniqueIndexRefreshesRepeatedly` - a guard on the new implementation: writing a key over the record
  that already holds it, and adding it on a new record in the same transaction, has to survive the uniqueness check.
* `refreshingAnIndexedViewRepeatedlyDoesNotGrowTheIndex` (`@Tag("slow")`) - 60 refreshes must not grow the index on
  disk. Fails at 2MB against its 512KB bound on the delete-and-recreate shape.

Added to `MaterializedViewSchedulerTest`:

* `aPassLeavesNoDatabaseContextOnTheThreadThatRanIt` (fails on `main`) and `aPassLeavesAContextItDidNotInstallAlone`.
* `aScheduledTaskHoldsNothingStrongThatLeadsBackToTheDatabase` - the reachability claim itself belongs in a heap
  check rather than a `System.gc()`-dependent test, but its precondition is asserted deterministically.
* `aTaskWhoseReferencesHaveClearedCancelsItself` and `aSelfCancellingTaskLeavesAReplacementRegistered` - the cancel
  path driven by clearing the references rather than by waiting on a collection.
* `theSchedulerThreadIsNamedAfterItsDatabase`.

Full `engine` unit suite including the `slow` lane: 12221 tests, 0 failures. The server/HA materialized-view ITs were not run locally (port
2480 was held by another process on this machine) and are left to CI.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1723,28 +1723,38 @@ then reported `ERROR`, or `STALE` once `MaterializedViewChangeListener` overwrot
 "the data you are looking at is old" when the data was in fact gone. It stayed gone until some later refresh
 happened to succeed, which for a MANUAL view may be never.
 
-The refresh now clears the previous snapshot with a `DELETE FROM` inside the same transaction as the repopulate,
-so the pass gets the isolation every other ArcadeDB transaction has: the deletes and the inserts stay in the
-transaction's own page buffer until commit, a concurrent reader sees the previous snapshot throughout and the new
-one afterwards, and a failure anywhere rolls the whole pass back onto the previous snapshot rather than onto
-nothing.
+The refresh now writes the defining query's rows over the previous snapshot's records inside one transaction,
+deleting only the surplus and creating only the shortfall. The pass gets the isolation every other ArcadeDB
+transaction has: every write stays in the transaction's own page buffer until commit, a concurrent reader sees the
+previous snapshot throughout and the new one afterwards, and a failure anywhere rolls the whole pass back onto the
+previous snapshot rather than onto nothing. A view's rows also keep their `@rid` across a refresh, where before
+every pass gave them new ones.
 
-Neither trigger fires on the trivial case, which is why this survived the existing tests: an unindexed view
-smaller than one truncate batch was already atomic. The regression tests cover both - an index on the backing type
-(the `dropIndex` commit, at any size) and a view larger than `arcadedb.truncateBatchSize` (the in-scan commit).
+Neither trigger fires on the trivial case, which is why this survived the existing tests: an unindexed view smaller
+than one truncate batch was already atomic. The regression tests cover both - an index on the backing type (the
+`dropIndex` commit, at any size) and a view larger than `arcadedb.truncateBatchSize` (the in-scan commit).
+
+**Rewriting in place rather than clearing and repopulating is what keeps an index on the backing type from paying
+for the atomicity**, and it makes an indexed refresh cheaper than it was before. Clearing the view would give every
+row a new RID, and `TransactionIndexContext` collapses a REMOVE followed by an ADD only on the same (key, RID), so
+each pass would leave a full set of real tombstones where the truncate used to drop the index and rebuild it empty.
+Measured over 120 refreshes of a 2000-row indexed view: a flat 256KB for the old truncate, 256KB growing to 3.9MB
+and still climbing for clear-and-repopulate, and a flat 256KB for the rewrite - because
+`DocumentIndexer.updateDocument` finds the key values unchanged for every row a stable view reproduces and skips the
+index outright. Zero index writes per pass, against a full rebuild per pass before. The backing buckets stay flat
+too, since no record is freed and reallocated. A `@Tag("slow")` soak test guards the bound.
 
 **The shadow-type swap the issue proposed was rejected.** Building into a second backing type and repointing the
 view at it reads better on paper - the swap is metadata only and the old rows are dropped as files rather than
 deleted one by one - but in this engine the swap has no cheap implementation. Renaming a type renames its buckets,
 and `PaginatedComponent.rename()` first waits for every page of the *whole database* to be flushed; paying a
-database-wide flush barrier twice per refresh, on every tick of every PERIODIC view, is a worse defect than the
-one being fixed. Repointing the view at a differently named backing type avoids the barrier but makes the view's
-own name stop being the type name, which is what makes `SELECT FROM <view>` work at all and what every record's
-`@type` reports. What the single transaction costs instead is bounded and not new: it holds the pages of the old
-rows as well as the new ones - largely the same pages, since the deletes hand their free space straight back to
-the inserts - and an index on the backing type is maintained entry by entry rather than dropped and rebuilt empty.
-The repopulate was already one unbounded transaction, so a view too large to refresh in one transaction was
-already too large to refresh.
+database-wide flush barrier twice per refresh, on every tick of every PERIODIC view, is a worse defect than the one
+being fixed. Repointing the view at a differently named backing type avoids the barrier but makes the view's own
+name stop being the type name, which is what makes `SELECT FROM <view>` work at all and what every record's `@type`
+reports. What the single transaction costs instead is bounded and not new: it holds the pages of the rows it
+rewrites, and under HA the pass is one Raft log entry rather than one per truncate batch plus one for the
+repopulate. The repopulate was already one unbounded transaction, so a view too large to refresh in one transaction
+was already too large to refresh.
 
 ### The scheduler no longer pins an abandoned database to a live thread
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1705,3 +1705,66 @@ by the negative size marker a plain collapse would replace with a positive one, 
 exactly what tells a bucket scan that a slot holds a document of its own.
 
 [#6178](https://github.com/ArcadeData/arcadedb/issues/6178)
+
+## A materialized-view refresh is one transaction, so a failed one no longer empties the view (#6203)
+
+A full refresh looked like a single transaction and was not one. `TRUNCATE TYPE ... UNSAFE` commits the caller's
+transaction from the inside: `schema.dropIndex()` for every index on the backing type before a single record is
+touched, then `commit(); begin()` once per `arcadedb.truncateBatchSize` records, then once more before it rebuilds
+the indexes it dropped. Wrapping that in `database.transaction()` bought nothing, and both consequences were
+silent.
+
+Every other reader saw the view empty, or holding some prefix of the new rows, for the whole runtime of the
+defining query - which for a view worth materialising is exactly the expensive case, and for a PERIODIC view is
+every tick. Nothing distinguished that from a legitimately empty or small result.
+
+Worse, a defining query that threw after the truncate had committed left the backing type with zero rows. The view
+then reported `ERROR`, or `STALE` once `MaterializedViewChangeListener` overwrote it, and both of those read as
+"the data you are looking at is old" when the data was in fact gone. It stayed gone until some later refresh
+happened to succeed, which for a MANUAL view may be never.
+
+The refresh now clears the previous snapshot with a `DELETE FROM` inside the same transaction as the repopulate,
+so the pass gets the isolation every other ArcadeDB transaction has: the deletes and the inserts stay in the
+transaction's own page buffer until commit, a concurrent reader sees the previous snapshot throughout and the new
+one afterwards, and a failure anywhere rolls the whole pass back onto the previous snapshot rather than onto
+nothing.
+
+Neither trigger fires on the trivial case, which is why this survived the existing tests: an unindexed view
+smaller than one truncate batch was already atomic. The regression tests cover both - an index on the backing type
+(the `dropIndex` commit, at any size) and a view larger than `arcadedb.truncateBatchSize` (the in-scan commit).
+
+**The shadow-type swap the issue proposed was rejected.** Building into a second backing type and repointing the
+view at it reads better on paper - the swap is metadata only and the old rows are dropped as files rather than
+deleted one by one - but in this engine the swap has no cheap implementation. Renaming a type renames its buckets,
+and `PaginatedComponent.rename()` first waits for every page of the *whole database* to be flushed; paying a
+database-wide flush barrier twice per refresh, on every tick of every PERIODIC view, is a worse defect than the
+one being fixed. Repointing the view at a differently named backing type avoids the barrier but makes the view's
+own name stop being the type name, which is what makes `SELECT FROM <view>` work at all and what every record's
+`@type` reports. What the single transaction costs instead is bounded and not new: it holds the pages of the old
+rows as well as the new ones - largely the same pages, since the deletes hand their free space straight back to
+the inserts - and an index on the backing type is maintained entry by entry rather than dropped and rebuilt empty.
+The repopulate was already one unbounded transaction, so a view too large to refresh in one transaction was
+already too large to refresh.
+
+### The scheduler no longer pins an abandoned database to a live thread
+
+`MaterializedViewScheduler` guarded against a database abandoned without `close()` by holding it through a
+`WeakReference` and cancelling the task once that cleared. It could not clear, for two independent reasons.
+
+The refresh runs a transaction on the scheduler thread, which installs a `DatabaseContext` entry keyed by the
+database path holding a strong reference to the database, and the scheduler never removed it - so after the very
+first refresh the weak reference was moot. `TimeSeriesMaintenanceScheduler.runMaintenance()` documents the same
+hazard and takes the entry back down in a `finally`; the refresh now does too, removing only a context that pass
+installed itself.
+
+The task also captured the `MaterializedViewImpl` strongly, and a view holds its own database, so the reference
+that was weak was not the only one. Both are weak now. The task keeps the view's name as a `String` and cancels
+itself when either reference clears, unregistering itself only while it is still the task registered for that view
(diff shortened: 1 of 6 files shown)


lvca:
Thanks - point 1 was right, and it turned out to be worth fixing rather than documenting. Pushed in 8ca5c07.

**1. Index tombstones.** I checked the collapse rule against `engine/CLAUDE.md` ("within one transaction
`TransactionIndexContext` collapses a REMOVE followed by an ADD on the same (key, RID)") - the review's reading is
correct, and `DELETE FROM` + recreate gives every row a new RID, so nothing collapses. Then I measured it, 120
refreshes of a 2000-row indexed view, index bytes on disk:

| refresh strategy | after 120 passes |
| --- | --- |
| `TRUNCATE` (drop + rebuild empty), i.e. before this PR | **256KB flat** |
| `DELETE FROM` + recreate, i.e. what I had pushed | **256KB → 3.9MB, still climbing** |
| write rows over the previous records, i.e. now | **256KB flat** |

So a real regression, ~15x and growing, held down only by the compaction it was itself provoking. `refreshOnce` now
writes each row **over** one of the previous snapshot's records, deleting only the surplus and creating only the
shortfall. `DocumentIndexer.updateDocument` has a `if (!keyValuesAreModified) // SAME VALUES, SKIP INDEX UPDATE`
path, so every row a stable view reproduces touches the index zero times - which makes an indexed refresh *cheaper*
than it was before this PR, not merely no worse, since the truncate paid a full index rebuild every pass. The
backing buckets stay flat too (no record freed and reallocated), and a view's rows now keep their `@rid` across a
refresh where every pass used to hand out new ones.

The RIDs are collected up front rather than iterated live: an update whose content no longer fits its slot can
relocate the record, and a scan still walking the bucket would then meet it a second time at its new home.

**2. Unbounded delete transaction.** Moot - there is no delete phase in the common case now, only the surplus rows.

**3. Minor.** Double space fixed. The server/HA ITs are still CI-only here (port 2480 is held by another process on
this machine).

Also added the soak test you asked for, `@Tag("slow")`, asserting 60 refreshes of an indexed view do not grow the
index on disk - and confirmed it is not vacuous: it fails at 2MB against its 512KB bound on the delete-and-recreate
shape. Two existing tests keyed their latch/failure injection on record *creates*, which are now rare, so both were
moved onto the first write of either kind.

Engine suite including the `slow` lane: 12221 tests, 0 failures.

lvca:
All three points taken - pushed in 77c3785.

**Positional reuse vs. a UNIQUE index.** Good catch, and worth chasing rather than reasoning about. I wrote the
test you described, driving the permutation deterministically rather than hoping for an unstable sort: the view
sorts on a column the test rewrites between passes, so all five keys change hands in one transaction, each landing
on a record another record is still holding.

It does not throw - `TransactionIndexContext` tracks entries per RID rather than one slot per key, so the ADD/REMOVE
pairs net out, which confirms the reading you'd traced but couldn't verify. `aViewWithAUniqueIndexSurvivesItsRowsChangingOrderBetweenRefreshes`
now holds that, and it is not vacuous: my first run failed because the assertion helper sorted the ids, which is
what proved the records really had been permuted. It asserts in record order now, so a permutation that never
happened cannot pass.

**Stale scheduler-test comment.** Fixed - those passes now fail on the scan that collects the previous snapshot,
before the defining query runs at all.

**`List<RID>` vs primitives.** Taken. It is allocated on every refresh of every view and is proportional to the
view's size, so a boxed RID per row (reference + header + the same two fields, against 12 bytes) is worth avoiding.
`collectIdentities` now holds two primitive arrays and rebuilds RIDs one at a time as the rewrite consumes them, so
at most one is live. That also let me drop `RID.asDocument()` for an explicit `database.lookupByRID(...)` - the
former resolves the database from the thread's `DatabaseContext`, which is a dependency this refresh has no reason
to have, least of all in a PR that changes `DatabaseContext` lifecycle.

One correction to a passing remark: the delete-phase transaction size is no longer unbounded in the common case -
since 8ca5c07 there is no delete phase at all unless the new snapshot is shorter than the previous one, and the PR
body has been updated accordingly.

Engine suite including the `slow` lane: 12221 tests, 0 failures.

lvca:
## CI summary

The third review round ran to completion and raised nothing. All lanes have now reported, and **every
materialized-view test passes on every lane**, including the server and HA ITs I could not run locally (port 2480
is held by another process on my machine):

| lane | materialized-view coverage | result |
| --- | --- | --- |
| `unit-tests` | all new engine tests | **pass** |
| `slow-unit-tests` | `refreshingAnIndexedViewRepeatedlyDoesNotGrowTheIndex` soak | **pass** |
| `integration-tests` | `HTTPMaterializedViewIT` 7/7, `Issue3941AsyncRefreshMaterializedViewIT` 2/2, `RemoteMaterializedViewIT` 2/2 | **pass** |
| `ha-integration-tests` | `RaftReplicationMaterializedViewIT` 1/1 | **pass** |
| `vector-unit-tests`, `opencypher-tck-tests`, `studio-e2e-tests`, 5 language e2e lanes, CodeQL, Codacy, Meterian, builder-tests | - | **pass** |

Two lanes are red, and neither is this PR:

**`integration-tests`** - 669 tests, 1 failure:
`Issue5470BatchStreamStallIT.aBodyShorterThanItsContentLengthIsNotReportedAsSuccess` asserts a truncated upload is
answered 408 and gets 409. It fails identically on `main` (runs 31893053331, 31892922131) and has been red for
days; it is a status-classification problem in the interrupted-stream path, unrelated to materialized views.

**`ha-integration-tests`** - 230 tests, 4 failures, all leadership/rejoin/sticky-connection:
`RaftTransferLeadershipIT`, `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`,
`RaftUserManagement3NodesIT.concurrentUserCreationsConvergeOnAllNodes`, `RemoteStickyConnectionStrategyIT`. This
lane is chronically red on `main` (#5668/#5702); `main` run 31892922131 fails with
`RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges` - the same test - plus `RaftForceResyncIT`.

The HA result is the one I most wanted, since a refresh is now one Raft log entry rather than one per truncate
batch plus one for the repopulate: `RaftReplicationMaterializedViewIT` confirms a replica still sees the view's
rows after the change.

PR #6209, opened 2026-08-15
fix(#6087): a copied EXPORT/IMPORT statement keeps its WITH settings, and a scheduled backup can be tuned per database

### LINKED ISSUE #6087: Backup phase-1 follow-ups: copy() drops settings on Export/Import, deferred-RAM metric unwired, no per-database compression override

robfrank:
Closes #6087

Item 2 of the issue (the deferred-RAM gauge) already shipped in #6116 / PR #6122, so this PR covers the two that were left open: `copy()` dropping settings on `EXPORT`/`IMPORT DATABASE`, and the absence of a per-database compression override for scheduled backups.

## 1. `ExportDatabaseStatement.copy()` / `ImportDatabaseStatement.copy()` dropped the settings map

Both carried only `url` over, so a copy lost every `WITH ...` setting - the identical defect `BackupDatabaseStatement` had and that #6080 fixed. As the issue notes, this is latent rather than live today: neither top-level admin statement is on a path that copies whole statements. It is fixed anyway because it is one edit away from becoming live, and because a silent drop is the worst possible failure mode for these two - `IMPORT DATABASE WITH vertices=...` has no source at all without its settings.

`ExportDatabaseStatement.copy()` also dropped `format` and `overwrite`, which are statement state in exactly the same way, so the fix copies the whole object rather than just the map the issue named.

**Adjacent gaps deliberately left alone** (neither is a regression, both are worth a look separately):

- `ExportDatabaseStatement.toString()` does not render the `WITH` clause, so a parse -> `toString()` -> re-parse round-trip is lossy. `BackupDatabaseStatement.toString()` has the same gap; `ImportDatabaseStatement.toString()` does not. This is why the new export test parses with `checkSyntax(...)` rather than `checkRightSyntax(...)` - the latter re-parses `toString()` output and would have handed back an empty settings map, i.e. a test that could not fail.
- `SimpleNode.hashCode()` computes `Objects.hashCode(getIdentityElements())`, and `getIdentityElements()` returns a freshly allocated `Object[]` on every call, so it is the array's *identity* hash: a node hashes to a different value on every invocation while `equals()` is content-based. Measured directly - three consecutive `hashCode()` calls on one `Expression` returned 1175962212, 2055281021, 1554547125. Consequence: any `SimpleNode` used as a `HashMap` key can never be looked up, and `Map.equals` over a settings map always returns false. Production impact appears to be nil (these maps are only ever iterated, and a `put` collision needs the two random hashes to match exactly, not merely the bucket), but the fix is a one-word `Arrays.hashCode`, and it touches every AST node, so it does not belong in this PR. The new tests compare the *rendered* `name -> value` pairs instead, and say so in a comment, so they do not depend on it in either direction.

## 2. No per-database compression override for scheduled backups

`DatabaseBackupConfig` gains three optional settings - `compressionLevel`, `compressionThreads`, `maxMBPerSecond` - mirroring `BackupSettings`'s `null` = "defer to the global configuration" convention. `BackupTask` threads them into the reflective `Backup` construction alongside the existing `setDirectory`/`setVerboseLevel` calls, skipping any the configuration left unset. A mixed fleet can now give a large, rarely-restored database `compressionLevel: 9` overnight while a small, frequently-restored one keeps `compressionLevel: 1` and stays out of the live workload's way.

```json
{
  "version": 1,
  "databases": {
    "archivedb": { "compressionLevel": 9, "compressionThreads": 8 },
    "livedb":    { "compressionLevel": 1, "maxMBPerSecond": 32 }
  }
}
```

They work in the `defaults` block too, and `mergeWithDefaults` fills only what the database left unset.

**Three decisions worth flagging:**

- **No `version` bump.** The issue offered "a version bump / backward-compat default of unset"; this takes the second. The fields are purely additive and absent is a valid, meaning-preserving state, so a version-1 `backup.json` loads and behaves byte-for-byte as it did. Nothing in the codebase gates on `version` - `AutoBackupConfig` reads it, stores it and never compares it - so bumping it would signal a migration that does not exist while changing no behavior.
- **Bounds are re-stated in the server module**, because `server` deliberately does not depend on `integration` at compile time (the backup is reached reflectively so a distribution without it still starts). `GlobalConfiguration.BACKUP_COMPRESSION_THREADS` already repeats the same literal for the same reason. `compressionOverrideBoundsMatchTheBackupApi` is the drift guard: it asserts against `BackupSettings.MAX_COMPRESSION_THREADS` and then asks the real `Backup` setters where they draw the line, rather than restating their literals a third time.
- **Out-of-range values are refused at load time**, in `validate()`, next to the existing cron and `maxFiles` checks - same shape, same exception, same call site - rather than being left to surface from inside a `Backup` setter during the 3am backup window.
- **`Backup.getSettings()`** is new and public. Every cross-module caller configures a `Backup` through a reflective boundary where a mis-wired setter fails silently; this is what lets the wiring be asserted against what actually reached the instance rather than inferred from the archive it happened to produce.

## Test plan

- [x] `ExportDatabaseStatementTestParserTest`, `ImportDatabaseStatementTestParserTest` - `copyKeepsTheWithSettings` **verified failing before the fix** ("Expecting actual not to be empty" on both), passing after. `copyOfAStatementWithoutSettingsKeepsAnEmptyMap` guards the empty case.
- [x] `BackupTaskCompressionOverrideTest` (7 tests, new) - every override reaches the `Backup`; an unset one leaves it `null` so the global default still decides; a partially configured database overrides only what it named; two databases in one fleet get their own settings; an out-of-range value set programmatically is still refused by the `Backup`; and the full chain from a `backup.json` string through `AutoBackupConfig.getEffectiveConfig` into the `Backup`, including the "no entry of its own" branch that copies the defaults field by field.
- [x] **Sabotage check**: with `applyCompressionOverrides`'s level branch forced to `if (false)`, 4 of the 5 tests then written went red. The tests detect the wiring, they do not merely accompany it.
- [x] `DatabaseBackupConfigTest` - 11 new tests for defaults, JSON parse, round-trip, `toJSON` omitting unset values, merge precedence, range refusals, boundary acceptance, and the bounds drift guard. 70 tests total, green.
- [x] `mvn -o -pl engine -am test -Dtest='com.arcadedb.query.sql.parser.**,DatabaseAdminStatementsTest'` - 371 tests, 0 failures.
- [x] `mvn -o -pl integration -am test -Dtest='com.arcadedb.integration.backup.**'` - 82 tests, 0 failures.
- [x] `mvn -o -pl server -am test -Dtest='com.arcadedb.server.backup.**'` - 160 tests, 0 failures.
- [x] `mvn -o test-compile -pl '!studio'` - BUILD SUCCESS across the reactor including all test sources. (`studio` fails on an unrelated offline npm cache miss for `xmldoc`, no Java involved.)

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ExportDatabaseStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ExportDatabaseStatement.java
@@ -133,6 +133,12 @@ public class ExportDatabaseStatement extends SimpleExecStatement {
   public Statement copy() {
     final ExportDatabaseStatement result = new ExportDatabaseStatement();
     result.url = this.url;
+    // WITHOUT THESE, A COPY SILENTLY DROPS EVERY 'WITH ...' SETTING AND FALLS BACK TO THE DEFAULT FORMAT - THE SAME
+    // DEFECT BackupDatabaseStatement HAD (#6080). COPY THE WHOLE STATE, NOT THE URL ALONE, SO THE COPY STAYS CORRECT
+    // THE MOMENT THIS STATEMENT REACHES A PATH THAT COPIES IT
+    result.format = this.format;
+    result.overwrite = this.overwrite;
+    result.settings.putAll(this.settings);
     return result;
   }
 
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ImportDatabaseStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ImportDatabaseStatement.java
@@ -149,6 +149,10 @@ public class ImportDatabaseStatement extends SimpleExecStatement {
   public Statement copy() {
     final ImportDatabaseStatement result = new ImportDatabaseStatement();
     result.url = this.url;
+    // WITHOUT THIS, A COPY SILENTLY DROPS EVERY 'WITH ...' SETTING - THE SAME DEFECT BackupDatabaseStatement HAD
+    // (#6080). FOR IMPORT THE SETTINGS ARE THE STATEMENT: WITHOUT THEM AN 'IMPORT DATABASE WITH vertices=...' HAS NO
+    // SOURCE AT ALL
+    result.settings.putAll(this.settings);
     return result;
   }
 }
--- a/engine/src/test/java/com/arcadedb/query/sql/parser/ExportDatabaseStatementTestParserTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/parser/ExportDatabaseStatementTestParserTest.java
@@ -20,6 +20,12 @@ package com.arcadedb.query.sql.parser;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
 class ExportDatabaseStatementTestParserTest extends AbstractParserTest {
 
   @Test
@@ -38,5 +44,58 @@ class ExportDatabaseStatementTestParserTest extends AbstractParserTest {
     checkWrongSyntax("export database file:///foo/bar/ foo bar");
     checkWrongSyntax("export database http://www.foo.bar asdf ");
     checkWrongSyntax("EXPORT DATABASE https://www.foo.bar asd ");
+  }
+
+  /**
+   * Regression test for issue #6087: {@code copy()} only carried the URL over, so every {@code WITH ...} setting was
+   * silently dropped by the copy. Same defect {@code BackupDatabaseStatement} had (#6080).
+   * <p>
+   * Parsed with {@link #checkSyntax(String, boolean)} rather than {@link #checkRightSyntax(String)} on purpose:
+   * the latter re-parses the node's own {@code toString()}, and {@code ExportDatabaseStatement.toString()} does not
+   * render the {@code WITH} clause, so the round-trip would hand back a statement with an empty settings map and the
+   * test would assert nothing.
+   */
+  @Test
+  void copyKeepsTheWithSettings() {
+    final SimpleNode parsed = checkSyntax("EXPORT DATABASE file://Movies.graphson.tgz WITH format = 'graphson', overwrite = true",
+        true);
+    assertThat(parsed).isInstanceOf(ExportDatabaseStatement.class);
+
+    final ExportDatabaseStatement original = (ExportDatabaseStatement) parsed;
+    assertThat(renderSettings(original)).containsOnly(entry("format", "'graphson'"), entry("overwrite", "true"));
+
+    final ExportDatabaseStatement copy = (ExportDatabaseStatement) original.copy();
+
+    assertThat(renderSettings(copy)).isEqualTo(renderSettings(original));
+    assertThat(copy.url).isEqualTo(original.url);
+  }
+
+  @Test
+  void copyOfAStatementWithoutSettingsKeepsAnEmptyMap() {
+    final SimpleNode parsed = checkSyntax("EXPORT DATABASE file://Movies.jsonl.tgz", true);
+    final ExportDatabaseStatement original = (ExportDatabaseStatement) parsed;
+    assertThat(original.settings).isEmpty();
+
+    final ExportDatabaseStatement copy = (ExportDatabaseStatement) original.copy();
+
+    assertThat(copy.settings).isEmpty();
+    assertThat(copy.url).isEqualTo(original.url);
+  }
+
+  /**
+   * Renders the settings the way {@code executeSimple} consumes them: the raw setting name held in
+   * {@code Expression.value} against the rendered value.
+   * <p>
+   * Comparing the two {@code Map<Expression, Expression>} instances directly would not work: {@code SimpleNode}
+   * derives {@code hashCode()} from a freshly allocated {@code Object[]}, so an {@code Expression} hashes to a
+   * different bucket on every call and {@code HashMap.equals} can never find a key. That is a separate, pre-existing
+   * defect - the settings map is only ever iterated in production, never looked up - and this test deliberately does
+   * not depend on it either way.
+   */
+  private static Map<String, String> renderSettings(final ExportDatabaseStatement statement) {
+    final Map<String, String> rendered = new HashMap<>();
+    for (final Map.Entry<Expression, Expression> entry : statement.settings.entrySet())
+      rendered.put(entry.getKey().value.toString(), entry.getValue().toString());
+    return rendered;
   }
 }
--- a/engine/src/test/java/com/arcadedb/query/sql/parser/ImportDatabaseStatementTestParserTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/parser/ImportDatabaseStatementTestParserTest.java
@@ -20,6 +20,12 @@ package com.arcadedb.query.sql.parser;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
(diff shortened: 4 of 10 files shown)


robfrank:
On the review's `format`/`overwrite` observation - half right, and the half that isn't is worth writing down so nobody files the follow-up it suggests.

**Right:** the ANTLR builder never populates the `format`/`overwrite` *fields*. `visitExportDatabaseStmt` puts every `WITH` pair into the `settings` map and nothing else, so those two fields sit at their constructor defaults (`"jsonl"` / `FALSE`) for the whole life of the statement. Copying them in `copy()` is therefore copying default state today, exactly as the review says - kept because `copy()` should reproduce the object, not a subset of it, and because the cost is two assignments.

**Not right:** `EXPORT DATABASE ... WITH format='graphson'` *does* change the export format today. The route is the settings map, not the field:

- `executeSimple` calls `setFormat(formatExport)` with the field's default, then immediately `setSettings(settingsToString)` with the `WITH` pairs;
- `Exporter.setSettings` feeds each pair to `ExporterSettings.parseParameter`, which has explicit `"format"` and `"overwrite"` branches (`integration/.../exporter/ExporterSettings.java:70` and `:76`);
- `setSettings` runs after `setFormat`, so the `WITH` value wins.

Which is precisely why the settings map was the part that had to be copied: dropping it is what would have silently reverted a graphson export to jsonl.

There is a real cosmetic seam next door, pre-existing and untouched here: when no URL is given, the default filename is built from the `format` *field*, so `EXPORT DATABASE WITH format='graphson'` writes graphson content into a file named `...jsonl.tgz`. Small, orthogonal to this PR, and only reachable on the no-URL path.

robfrank:
### `unit-tests` is red on this PR, and it is the lane, not the change

Two runs of the same commit (`c4ae8f6`), each 1 failure out of 11980, and **a different test each time**:

| run | failing test | assertion |
|---|---|---|
| [95045966159](https://github.com/ArcadeData/arcadedb/actions/runs/31898403896/job/95045966159) | `TimeSeriesTagDictionaryTest.aSecondWaveOfIdsIsAlsoResolvedByReloading:198` | `expected: "host_b" but was: null` |
| [95053183083](https://github.com/ArcadeData/arcadedb/actions/runs/31898403896/job/95053183083) (re-run, no new commit) | `DictionaryMultiPageTest.aDictionaryPageThatIsClaimedButMissingFailsLoudly:472` | `Expecting code to raise a throwable` |

A regression is deterministic and lands on the same test twice. Two different single-test failures in engine page/dictionary internals is the signature of a flaky lane.

Neither test is reachable from this diff. The engine half of it is ten lines inside `ExportDatabaseStatement.copy()` and `ImportDatabaseStatement.copy()`; a time-series tag dictionary reloading its pages and the engine dictionary's claimed-page check have no path to `Statement.copy()`. The rest of the diff is in `server`/`integration`.

Checked locally, against the isolated worktree:

- `DictionaryMultiPageTest` - 14 tests, green.
- `TimeSeriesTagDictionaryTest` - 12 tests, green; also run 3x on its own, green each time.
- `com.arcadedb.engine.timeseries.**` at the base commit `76555a5750` (control worktree, detached) - 487 tests, green. Same result at head, so the package does not discriminate between base and head; the flake needs full-suite conditions.

Every other lane on this commit is green, including `slow-unit-tests`, `integration-tests`, `ha-integration-tests`, `vector-unit-tests`, `builder-tests`, `opencypher-tck-tests`, `build-and-package` and all five e2e suites.

Worth noting `main` itself is intermittently red in the same lane - run 31852372986's `unit-tests` failed on a single timing-ratio assertion in `Issue6116ChecksumsPointInTimeTest`, again 1 of N with nothing else wrong.

PR #6210, opened 2026-08-15
perf(#6086): the restore is parallel per entry, and stops reading the archive 512 bytes at a time

### LINKED ISSUE #6086: Parallel restore: restore is now ~4x slower than the backup it undoes
## Context

Backup phase 1 (#6072, merged in #6080) made the full backup up to 27x faster. That moved the bottleneck: **restore is now the slow half of a recovery.**

Measured on the same 1.25 GB database, same machine (`BackupCompressionBenchmark`):

| | duration |
|---|---|
| Backup, level 1, 8 threads | **0.68 s** |
| Restore of that archive | **2.9 s** |

Restore is now ~4x slower than the backup it undoes. For RTO - the number that actually matters when someone is restoring - the work in #6072 barely moved the needle, because it only addressed the half that was already the faster one after the change.

## Where the time goes today

`FullRestoreFormat.restoreDatabase()` (`integration/src/main/java/com/arcadedb/integration/restore/format/FullRestoreFormat.java:87`) is a single sequential loop:

```java
ZipEntry compressedFile = zipFile.getNextEntry();
while (compressedFile != null) {
  databaseOrigSize += uncompressFile(zipFile, compressedFile, databaseDirectory);
  compressedFile = zipFile.getNextEntry();
}
```

and `uncompressFile` (`:113`) inflates into an 8 KB buffer, writing each chunk to a `FileOutputStream` as it goes. One thread does inflate and write, alternating, for the whole archive.

**Measure before building.** It is not yet established whether restore is inflate-bound or write-bound - 1.25 GB in 2.9 s is ~430 MB/s, which is plausible as either on NVMe. The first task in this issue is to find out, because the two point at different fixes:

- **Write-bound** -> the fix is overlapping inflate with I/O, and the 8 KB `BUFFER` (`:49`) is suspiciously small; raising it may be most of the win on its own for a fraction of the effort.
- **Inflate-bound** -> the fix is real parallelism, below.

## Proposed direction

**Per-entry parallelism.** ZIP entries are independent and each becomes its own output file, so N entries can be inflated and written concurrently. This is the natural unit and needs no format change.

**The constraint that shapes the design:** per-entry parallelism requires random access to the archive (`ZipFile`), but the current path is a *stream* (`ZipInputStream`), and two of its input sources cannot be made random-access:

- `openInputFile()` (`:147`) supports `http://` / `https://` URLs - a one-shot stream.
- `decryptFile()` (`:171`) wraps the input in a `CipherInputStream` for encrypted backups.

So a parallel restore cannot simply replace the existing path; it has to be a *fast path* for a plain local archive, with the current sequential stream retained for remote and encrypted inputs. That fallback is not optional and should be settled before any code is written.

Note that intra-entry parallelism (the mirror of what the backup does) is **not** available: the backup's chunk boundaries are not recorded anywhere in the archive, and recording them would be an archive-format change, which phase 1 deliberately avoided and which would break the "old backups restore" guarantee. A single entry is one deflate stream and must be inflated serially.

## Scope

- Establish whether restore is inflate-bound or write-bound, with numbers.
- Raise the 8 KB copy buffer if that alone is material.
- Per-entry parallel restore for the local-file case, with a documented thread setting consistent with `arcadedb.backup.compressionThreads` (`-1` = auto, `0` = sequential legacy path).
- Keep the sequential path for HTTP and encrypted archives, and cover both in tests.

## Out of scope

- Any archive-format change, including recording chunk boundaries to enable intra-entry parallelism.
- Incremental restore / chain folding (Phase 3 of `docs/optimize-backup.md`).

## Acceptance criteria

- Restore duration measured before and after at several database sizes, in the same shape as the #6072 benchmark table, and reported in the PR.
- A restored database is byte-identical to what the sequential path produces - `DatabaseComparator` against the source, for plain, encrypted and HTTP inputs.
- Old archives (produced before #6072, single-threaded level 9) restore unchanged.
- Peak heap bounded and stated, as the backup path does.

Follow-up to #6072. Related: `docs/optimize-backup.md`.

robfrank:
Closes #6086

Backup phase 1 (#6072) made the full backup 27x faster and thereby moved the bottleneck: on the same 1.25 GB database the backup took 0.68 s and the restore that undoes it 2.9 s. This makes the restore inflate entries in parallel, one entry per thread, and fixes the read pattern that was costing more than the parallelism buys on a single-file database. The archive format is untouched: nothing here reads anything the backup did not already write, and pre-#6072 single-threaded level-9 archives restore through the new path unchanged.

## Measure first, as the issue asked

`RestoreParallelBenchmark` (`@Tag("benchmark")`, never runs in CI) times the two halves of a restore separately. On the 1.25 GB fixture: **inflating every entry and discarding the bytes takes 2.96 s, writing the same bytes with no inflation takes 0.93 s**. The restore is inflate-bound, so the issue's "the 8 KB copy buffer is suspiciously small" hypothesis is the wrong one - raising that buffer to 256 KB is worth about 1%.

The read side was worth a great deal, for a reason nobody had looked at: `ZipInputStream` fills its inflater **512 bytes at a time** (a size its constructor hardcodes, with no overload) and the restore handed it an unbuffered `FileInputStream` - one read syscall per 512 compressed bytes for the whole archive. A 256 KB `BufferedInputStream` under it takes the sequential restore from 5.16 s to 3.96 s, and it is the only speedup available to the two input sources that cannot be parallel at all.

## The change

- **`ParallelZipExtractor`** (new): opens the archive with `ZipFile`, hands entries to a dedicated pool largest-first (the makespan of uneven independent tasks is decided by when the biggest one starts), and writes each entry to its own file. No `ForkJoinPool.commonPool()`; the pool's lifetime is exactly one restore, and peak heap is bounded by construction at one 256 KB copy buffer per pool thread taken from a pool of exactly that many.
- **`arcadedb.restore.threads`**: `-1` (default) = available processors capped at 8, `0` = the sequential walk, `N` = a pool of N. Also `-restoreThreads` on the command line and `Restore.setRestoreThreads(int)` in the API, all validated against the same bound (a test fails if the two copies of that bound drift). Unlike a backup, which halves the cores because it runs beside the workload it is already throttling, a restore has no such neighbour - the database it produces is not open yet - so the automatic sizing takes whole cores.
- **The sequential walk stays**, buffered, as the fallback for the two sources that cannot be opened at random: an `http(s)` archive is a one-shot stream and an encrypted one is a single `CipherInputStream`. The fallback is automatic whatever the setting says; a restore must never fail because of a performance setting.
- **Two hostile-archive refusals the parallel path can make and the sequential one cannot**: reading the central directory first means every entry name is validated (zip-slip, `..`, embedded separators) *before any file is written*, and an archive with two entries of the same name - two threads writing one file - is refused rather than given a locking rule of its own.

## The numbers

1.25 GB database, 8 threads, macOS/NVMe. Two shapes, because the shape is what decides what per-entry parallelism can buy:

| configuration | one dominant entry (1 type) | several comparable entries (8 types) |
|---|---|---|
| before | 5.16 s (248 MB/s) | 4.51 s (258 MB/s) |
| + buffered read, sequential | 3.96 s (1.30x) | 4.17 s (1.08x) |
| parallel, 1 thread | 2.60 s (1.99x) | 2.63 s (1.71x) |
| parallel, 8 threads | **2.51 s (2.05x)** | **0.49 s (9.15x)** |

Smaller database, same shape (322 MB, 1 type): 1.15 s before, 0.64 s at 8 threads.
Legacy level-9 archive, 8 types: 3.90 s before, 0.45 s at 8 threads (10.1x).

**Stated rather than averaged away:** a database made of one dominant file cannot be split - a ZIP entry is a single deflate stream and has to be inflated serially - so there the whole win is the reader change and the threads add 4%. Intra-entry parallelism, the mirror of what the backup does, would need the backup's chunk boundaries recorded in the archive, which is the format change this deliberately avoids. On a database spread over several types the restore lands at 0.49 s, faster than the 0.68 s backup it undoes, which is where the issue wanted it.

**Peak heap does not move:** 11.0 MB in every configuration above, sequential and parallel alike.

## Test plan

- [ ] `mvn -o -pl integration -am verify -DskipITs=false -Dit.test=Issue6086ParallelRestoreIT` - 17 tests: the round trip at threads 0/1/2/8/-1 compared against the source with `DatabaseComparator`; the parallel and sequential paths compared **byte for byte**; the encrypted and `http` archives asserted to have taken the fallback (by the path each restore logs, and by the thread each per-entry line was logged from - not merely by having succeeded); a pre-#6072 level-9 archive through the parallel path; the global-configuration default; zip-slip (`../x`, `sub/x`, `..`) refused with the destination directory left empty; duplicate entry names refused; an empty archive refused on both paths.
- [ ] `mvn -o -pl integration -am test -Dtest='RestoreSettingsTest,FullRestoreFormatThreadSizingTest'` - 18 tests: CLI/API/global-configuration validation of the new setting, the shared bound, and the automatic sizing boundaries.
- [ ] `mvn -o -pl integration verify -DskipITs=false` - the whole integration module, 100 ITs green, including `BackupCompressionIT`, `FullBackupIT` and `Issue6075SnapshotBackupIT`, which restore through the changed path.
- [ ] `mvn -o -pl engine test -Dtest='ConfigurationTest,GlobalConfigurationTest,ContextConfigurationTest'` - 44 tests, for the new `GlobalConfiguration` entry.
- [ ] The tests were checked to be able to fail: with the parallel path forced off, 10 of the 17 ITs go red, so they genuinely reach the new branch rather than passing on a silent fallback.
- [ ] Not run locally: the `server` module ITs, because port 2480 is held by another process on this machine. The server calls `Restore` reflectively through the unchanged `(String, String)` constructor and `restoreDatabase()`, so no signature it depends on moved - but CI should be read for that lane.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2035,4 +2035,10 @@ every entry name up front and refuses a hostile archive before writing a single 
 learns a name when it reaches it, so it stops at the bad entry with the good ones already on disk - the
 pre-existing behaviour, now pinned by a test so the difference is a recorded decision rather than an accident.
 
+A failure *during* extraction carries one guarantee that a naive pool would not give: when the extractor throws,
+no worker is still writing. Interrupting a thread does not come back out of a `FileOutputStream.write()`, so
+shutting the pool down is not enough on its own - and the caller's very next move is usually to delete the
+destination, which would then race the workers still filling it. The extractor therefore waits for its workers
+before propagating the failure.
+
 [#6086](https://github.com/ArcadeData/arcadedb/issues/6086)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1962,3 +1962,77 @@ reads from.
 already left behind.
 
 [#6090](https://github.com/ArcadeData/arcadedb/issues/6090)
+
+## The restore is now the fast half of a recovery too: parallel per entry, and no longer reading the archive 512 bytes at a time (#6086)
+
+Making the backup 27x faster (#6072) moved the bottleneck rather than removing it. On the same 1.25 GB database
+the backup took 0.68 s and the restore that undoes it 2.9 s, so the number that actually matters when someone is
+recovering - the RTO - had barely moved.
+
+### What the measurement said, and what it contradicted
+
+The issue proposed two candidate diagnoses and asked for the numbers before any code. `RestoreParallelBenchmark`
+times the two halves of a restore separately, and they are not close: on that 1.25 GB database, inflating every
+entry and throwing the bytes away takes **2.96 s** while writing the same bytes with no inflation takes **0.93 s**.
+The restore is inflate-bound, and the "the 8 KB copy buffer is suspiciously small" hypothesis is the wrong one -
+raising that buffer to 256 KB is worth 1%.
+
+The read side, however, was worth a great deal, and for a reason no one had looked at: `ZipInputStream` fills its
+inflater **512 bytes at a time**, a size its constructor hardcodes with no overload to change it, and the restore
+handed it an unbuffered `FileInputStream`. That is one read syscall per 512 compressed bytes for the whole
+archive. A 256 KB `BufferedInputStream` underneath it takes the same restore from 5.16 s to 3.96 s, and it is the
+only speedup available to the two input sources that cannot be parallel at all.
+
+### Parallel per entry
+
+ZIP entries are independent and each becomes its own file, so they are inflated and written concurrently, largest
+first (the duration of a set of very uneven independent tasks is decided by when the biggest one starts). This
+needs random access to the archive - `ZipFile` rather than `ZipInputStream` - which turns out to be the larger
+half of the win even at one thread, because `ZipFile`'s reader does not have the 512-byte fill.
+
+`arcadedb.restore.threads` sizes the pool: `-1` (the default) is the available processors capped at 8, `0` selects
+the sequential walk, `N` a pool of N. It is also settable per restore, on the command line (`-restoreThreads`) and
+through the `Restore` API. Unlike a backup, which halves the cores because it runs alongside the workload it is
+already throttling, a restore has no such neighbour - the database it is producing is not open yet - so the
+automatic sizing takes whole cores.
+
+**The archive format did not change**, and no part of this reads anything the backup did not already write. Old
+archives, including single-threaded level-9 ones written before #6072, restore through the parallel path unchanged.
+
+### The numbers
+
+1.25 GB database, 8 threads, macOS/NVMe. Two shapes, because the shape is what decides what per-entry parallelism
+can possibly buy:
+
+| configuration | one dominant entry (1 type) | several comparable entries (8 types) |
+|---|---|---|
+| before #6086 | 5.16 s (248 MB/s) | 4.51 s (258 MB/s) |
+| + buffered read, sequential | 3.96 s (1.30x) | 4.17 s (1.08x) |
+| parallel, 1 thread | 2.60 s (1.99x) | 2.63 s (1.71x) |
+| parallel, 8 threads | **2.51 s (2.05x)** | **0.49 s (9.15x)** |
+
+A database made of one dominant file cannot be split - a ZIP entry is a single deflate stream and has to be
+inflated serially - so there the whole win is the reader change and the threads add 4%. This is stated rather
+than averaged away because it is the honest shape of the result: intra-entry parallelism, the mirror of what the
+backup does, is not available without recording the backup's chunk boundaries in the archive, which would be the
+format change this deliberately avoids. On a database spread over several types the restore goes to 0.49 s -
+faster than the 0.68 s backup it undoes, which is where the issue wanted it.
+
+Peak heap is bounded by construction and does not move: 11.0 MB in every configuration above, sequential and
+parallel alike. One 256 KB copy buffer per pool thread, taken from a pool of exactly that many, plus the JDK
+inflater's own buffer per entry in flight - under 3 MB of buffers at 8 threads.
+
+### What stays sequential, and why it is not optional
+
+Per-entry parallelism needs to open the archive at random, which two of the three input sources cannot support: an
+archive read over `http(s)` is a one-shot stream, and an encrypted one is a single `CipherInputStream` that only
+decrypts front to back. Both fall back to the sequential walk automatically, whatever the thread setting says - a
+restore must never fail because of a performance setting - and both are covered by tests that assert the fallback
+was taken, not merely that the restore worked.
+
+One thing the parallel path does strictly better: it reads the central directory before it starts, so it validates
+every entry name up front and refuses a hostile archive before writing a single file. The sequential walk only
+learns a name when it reaches it, so it stops at the bad entry with the good ones already on disk - the
+pre-existing behaviour, now pinned by a test so the difference is a recorded decision rather than an accident.
+
+[#6086](https://github.com/ArcadeData/arcadedb/issues/6086)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -613,6 +613,23 @@ public enum GlobalConfiguration {
       invalid.""",
       Integer.class, 0),
 
+  RESTORE_THREADS("arcadedb.restore.threads", SCOPE.JVM,
+      """
+      Number of threads used to restore a full backup. ZIP entries are independent files, so they are inflated and \
+      written concurrently, one entry per thread. -1 (the default) sizes the pool automatically at the available \
+      processors capped at 8; 0 selects the legacy single-threaded stream walk, kept as an escape hatch. Unlike a \
+      backup, a restore does not run alongside the database it is working on - that database does not exist yet - \
+      which is why the automatic sizing claims whole cores rather than half of them. The parallel path needs random \
+      access to the archive and is therefore used only for a plain local file: an archive read over http(s) is a \
+      one-shot stream and an encrypted one is a single cipher stream, and both fall back to the sequential walk \
+      automatically, whatever this setting says. Peak heap is bounded by construction at one copy buffer per thread \
+      (256 KB) plus the JDK inflater's own buffer, so under 3 MB at 8 threads. Parallelism is per entry: a database \
+      made of one dominant file cannot be split, because a ZIP entry is a single deflate stream that has to be \
+      inflated serially.""",
+      // SAME BOUND AS RestoreSettings.MAX_RESTORE_THREADS, WHICH THE CLI AND THE Restore API VALIDATE AGAINST. THE
+      // ENGINE CANNOT DEPEND ON THE INTEGRATION MODULE, SO THE LITERAL IS REPEATED: CHANGE ONE AND CHANGE THE OTHER
+      Integer.class, -1, integerRangeAsStrings(-1, 256)),
+
   // SQL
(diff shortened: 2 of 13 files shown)


robfrank:
Cycle 1 review addressed in 98ce66bbbe.

**The mid-restore teardown was a real defect and is fixed.** `shutdownNow()` drains the queue, so no further entry starts, but its interrupt does not come back out of a `FileOutputStream.write()` or a `ZipFile` read - so up to poolSize-1 workers kept writing after `extract()` had already thrown. That matters exactly where you said it does: the server's restore handler deletes the destination directory when a restore fails, and deleting a tree that threads are still filling is a race. `extract()` now waits for termination before propagating, bounded at 60 s only so a pathological case degrades to a logged warning instead of hanging. On the success path every task is already done, so the wait costs nothing.

**The regression test you asked for is there, and it can fail.** `aFailureMidExtractionLeavesNoWorkerStillWriting` drives `ParallelZipExtractor` directly (going through `Restore` would delete the planted failure before extracting) and injects the failure where it is deterministic and portable: a directory planted at the target path of the largest entry, so that entry's `FileOutputStream` throws the instant it opens, while the other worker is at the start of its own. It asserts that no `arcadedb-restore-inflater-*` thread is alive when `extract()` throws, and that the file the survivor was writing is whole.

Worth recording how that test was tuned, because the first version of it was worthless: with 8 MB incompressible entries it passed **with the fix reverted** - the survivor finished inside the window. The entries are now 192/256 MB of zeroes, which costs a few hundred KB of archive and still makes the survivor write 192 MB, so the question has a stable answer. Verified in both directions: with `awaitTermination` removed the test fails, with it in place all 18 tests in the class pass.

The three minor notes needed no action, as you said - agreed on all three.

robfrank:
Cycle 2 addressed in 01e5957733. Two of the three applied, one deliberately not.

**Unbounded queue** - kept, with the justification you offered as the alternative, written into the code. The difference from the #6072 backup writer is real rather than an oversight: there, work is produced continuously while the file is read, so the bounded queue is the backpressure that stops megabyte chunk buffers piling up. Here the whole worklist exists before the pool does and is submitted in one closed loop - its length is the number of files in a database, and what is queued is a `ZipEntry` reference, not a buffer. Switching to `ArrayBlockingQueue` + `CallerRunsPolicy` would also cost something concrete: caller-runs would have the submitting thread extract an entry, taking a buffer from a pool sized for the workers, which breaks the "one copy buffer per pool thread" heap bound the PR states. So the pattern is followed where it applies and departed from where it does not, with the reason recorded.

**`resolveTarget` twice** - fixed, and it was slightly more than a nit. The entry is now paired with its resolved target in a `PlannedEntry` during the validation pass and the worker uses that. The duplication was not just wasted work: the worker was re-deriving a path from checks that had already decided whether the archive was acceptable, so the bytes were written against a second, later answer to a settled question.

**Elapsed time** - deliberate, not changed. `beginTime` now covers opening the archive because that is part of what a restore costs; excluding the central-directory read would systematically flatter the parallel path, which is exactly the number this PR is claiming. Nothing parses that line - it is a level-0 console line, forwarded verbatim to the SSE progress stream - and it is reported in whole seconds.

Re-ran after the change: 18/18 in `Issue6086ParallelRestoreIT`, and the whole integration module is green at 102 ITs.

robfrank:
Cycle 3 addressed in 60b15409f5. The SSE finding was real and is the best catch of the three rounds - it is a defect this PR introduces, not a theoretical one.

**Verified before agreeing:** `sendSSE` (PostServerCommandHandler:706) does a bare `out.write(...)` + `out.flush()` on `exchange.getOutputStream()`, and the restore's listener proxy calls it per log line. Before this PR the restore logged from one thread; now up to 8 workers log one line each per entry, so two `data:` frames could interleave into one unparseable event. Fixed at both ends, because the two failures are not the same failure:

1. **`ConsoleLogger` now serializes its listener callbacks.** That is where the contract actually changed: `LogListener` is a one-method interface whose implementations are written by whoever consumes the logger, and it is unreasonable to ask each of them to discover that a restore became multi-threaded. Pinned by `ConsoleLoggerConcurrencyTest.listenerCallbacksNeverOverlap` (8 threads x 20 lines, listener asserts it is never entered twice at once); it fails with the `synchronized` removed.
2. **`sendSSE` synchronizes on the stream.** This one is deliberately not redundant: as you suspected about the importer, that path already had the race before this PR - the import handler polls `ImporterContext` on its own thread while the importer logs on another, and both call `sendSSE` on the same stream. That is a pre-existing bug, out of this PR's scope to have caused, and one line to close while here.

I could not run the server IT lane locally (port 2480 held by another process on this machine, which makes those tests fail as auth errors rather than as a port conflict), so the SSE endpoint itself is covered only by CI. The guarantee the endpoint depends on is unit-tested at the logger, which is the layer that changed.

**Case-insensitive duplicate names** - fixed, and the up-front name check was the wrong place for it. Refusing on a lowercased comparison would break a legitimate restore on Linux: ArcadeDB type names are case-sensitive, so a database with types `Doc` and `doc` has exactly such a pair of bucket files and restores fine on a case-sensitive filesystem. The question is not about names at all, it is "is this the same file?", and only the filesystem can answer it - so the worker now creates the file with `createNewFile()` and fails if it is already there. `twoNamesDifferingOnlyInCaseFollowTheFilesystem` probes the filesystem and asserts whichever answer it gives (refused when case-insensitive, both files extracted when not), so it is meaningful on both macOS and Linux; it fails on this machine with the guard removed.

**Shared name/path validation between the two paths** - left as is, deliberately. The two are not the same code: the parallel path validates every entry before writing anything and returns the resolved target for the worker to use, while the sequential one can only check an entry when the stream reaches it. Extracting a helper would hide that difference behind a shared name, and the difference is the point. The drift risk is covered by `aHostileEntryNameIsRefusedBeforeAnythingIsWritten`, which runs the same three hostile names through **both** paths.

Full integration module re-run after the change: 103 ITs green, 19 in `Issue6086ParallelRestoreIT`. Server module compiles.

robfrank:
Cycle 4: no code changes. All three notes checked, two declined with reasons, one measured.

**`@Tag("slow")` on the 192 MB test** - measured rather than guessed: that test runs in **1.23 s** on its own, and the whole 19-test class in 0.79-1.5 s. The payload is zeroes, so both the deflate that builds the archive and the write that extracts it are near-memcpy; the archive itself is a few hundred KB. And the tag would not move it anywhere even if it were slow: the `integration-tests` lane runs `verify -Pintegration` with no `-DexcludedGroups`, and `failsafe.excludedGroups` is deliberately left empty (the #5697 comment in the parent POM explains why), so a `slow` tag on an IT excludes it from nothing and there is no slow-IT lane to route it to. Tagging would only make the next reader think it had been triaged into a lane it is not in.

**`Integer.parseInt(null)`** - left as is, deliberately. It is not quite an implementation detail (`Integer.parseInt` starts with an explicit `if (s == null) throw new NumberFormatException("Cannot parse null string")`), but the stronger reason is that `RestoreSettings.parseIntSetting` is a deliberate twin of `BackupSettings.parseIntSetting` from #6072, down to the message shape. Two methods that read identically and behave identically are easier to keep correct than two that differ by a null check someone later has to diff to understand; if the explicit guard is wanted, it should land on both in one change rather than being introduced here as a divergence.

**Zip-slip duplication without a drift-detecting test** - there is one, which I should have named more precisely last round: `aHostileEntryNameIsRefusedBeforeAnythingIsWritten` is parameterized over `../escaped.txt`, `sub/escaped.txt` and `..`, and runs **every one of them through both paths** - `restore(..., 8, null)` for the parallel one and `restore(..., 0, null)` for the sequential one - asserting both refuse. Weakening the check on either path alone turns that test red. The logic stays duplicated because the two are not the same function: one validates every entry up front and hands the worker a resolved target, the other can only check an entry when the stream reaches it, and a shared helper would hide exactly the difference that makes the parallel path's refuse-before-write behaviour possible.

Nothing outstanding from my side. Merge is the developer's call.

PR #6214, opened 2026-08-15
fix(#6065): an embedding dimension is bounded, and a negative k is named instead of reaching the allocator

### LINKED ISSUE #6065: Embedding-dimension-shaped params have no upper bound relative to graph size (follow-up to #5924); two minor error-quality nits
## Summary

Follow-up to #5924 (PR #6055). While fixing the unchecked-numeric-narrowing family in the OpenCypher procedures package, code review on PR #6055 surfaced two related gaps that #5924 deliberately left out of scope because they aren't narrowing bugs - the values involved never wrap or produce a silently-wrong result, they're just uncapped or produce a confusing error type.

## Part A - Embedding-dimension-shaped parameters have no upper bound relative to graph size

`#5924`'s fix rejects a config value that overflows `int` (via `NumberUtils.toIntExact()`/`AbstractAlgoProcedure.extractInt()`), but a large **in-range** `int` still passes validation and is used directly to size an array. Unlike top-k-shaped parameters (`AlgoKNN.k`, `AlgoInfluenceMaximization.k`, `AlgoHierarchicalClustering.numClusters`, ...), which all clamp against the graph's node count before allocating (`Math.min(k, n)`), embedding-dimension-shaped parameters have no natural graph-derived bound to clamp against, so they're used unclamped:

```java
// AlgoNode2Vec.java:102, then :153-154
final int dim = ... ? extractInt(n, "embeddingDimension") : 128;
...
final double[][] W = new double[n][dim];
final double[][] WCtx = new double[n][dim];

// AlgoFastRP.java:100, then :124
final int dimensions = ... ? extractInt(n, "dimensions") : 128;
...
final double[][] embed = new double[n][dimensions];

// AlgoGraphSAGE.java:97, then :138,143
final int outDim = ... ? extractInt(n, "embeddingDimension") : 64;
...
final double[][] W = new double[outDim][concatDim];
final double[][] nextEmbed = new double[n][outDim];

// AlgoHashGNN.java:96, then :133-134,157
final int embDim = ... ? extractInt(n, "embeddingDimension") : 128;
...
final int[] hashA = new int[embDim];
final int[] hashB = new int[embDim];
final double[][] embeddings = new double[n][embDim];
```

`CALL algo.node2Vec({embeddingDimension: 1000000000})` on even a tiny graph (`n` as low as 1) attempts a `new double[1][1000000000]` (~8 GB) - the same allocation-DoS shape PR #6055 fixed for `AlgoKNN`/`AlgoKShortestPaths`/`LSMVectorIndex`, just triggered by a large valid value instead of an overflowing one, so it's outside what an overflow check alone can catch.

**Suggested fix:** add a sane hard cap (e.g. a `MAX_EMBEDDING_DIMENSION` constant somewhere shared, perhaps in the low thousands) to each of these four sites, rejecting anything above it with a clear `IllegalArgumentException` via the existing `extractInt`-adjacent validation style - not a silent clamp, since there's no "correct" dimension to fall back to the way `Math.min(k, n)` has one for top-k parameters.

Note `AlgoHierarchicalClustering.numClusters` (`AlgoHierarchicalClustering.java:114`, `Math.max(1, Math.min(numClusters, n))`) and `AlgoMaxKCut`'s `k` are **not** affected - both already clamp against `n` downstream, same as the top-k family.

## Part B - Two minor error-quality inconsistencies (lower priority, bundled here rather than filed separately)

1. **Negative `k` produces a confusing low-level exception.** `AlgoKNN.java:128-129` (`new double[k]`/`new int[k]`) and `AlgoKShortestPaths.java`'s `ArrayList` sizing throw `NegativeArraySizeException`/`IllegalArgumentException: Illegal Capacity` for a literal negative `k` (e.g. `algo.knn(-5)`) instead of a clear validation message naming the parameter. Pre-existing behavior (identical before and after #5924/PR #6055 - `.intValue()` on a negative in-range value was already negative), not a regression, just worth a friendlier check (`k >= 0`) at the same site `extractInt`/`saturateToInt` already validate at.
2. **`DbIndexVectorQueryNodes`'s numeric-string fallback doesn't saturate.** `DbIndexVectorQueryNodes.java:102`: `args[1] instanceof Number n ? NumberUtils.saturateToInt(n) : Integer.parseInt(args[1].toString())` - the `Number` branch now saturates (per #5924), but the string-fallback branch (`Integer.parseInt`) still throws `NumberFormatException` for a numeric string outside `int` range instead of behaving consistently with the `Number` branch it sits next to.

## Related

#5900, #5905, #5906, #5919, #5924, PR #5921, PR #6055.

robfrank:
Closes #6065

## Summary

Follow-up to #5924 / PR #6055, which fixed the unchecked-numeric-*narrowing* family in the OpenCypher procedures package. The two gaps that fix deliberately left open are not narrowing bugs, which is why an overflow check alone could never catch them.

**Part A - embedding-dimension-shaped parameters now have an upper bound.** A top-k-shaped parameter has a natural graph-derived ceiling and every one of them clamps against it (`Math.min(k, n)`). An embedding dimension has none: it multiplies the per-node allocation regardless of how small the graph is. So a large but perfectly in-range `int` sailed through `extractInt()`/`NumberUtils.toIntExact()` and went straight into the allocation - `CALL algo.node2Vec({embeddingDimension: 1000000000})` on a one-node graph asks for `new double[1][1000000000]`, about 8 GB. This is the same allocation-DoS shape PR #6055 closed for `AlgoKNN`/`AlgoKShortestPaths`/`LSMVectorIndex`, reached by a valid value instead of an overflowing one.

`AbstractAlgoProcedure.extractEmbeddingDimension()` bounds the value to `[1, MAX_EMBEDDING_DIMENSION]` and is wired into the four affected sites: `AlgoNode2Vec.embeddingDimension`, `AlgoFastRP.dimensions`, `AlgoGraphSAGE.embeddingDimension`, `AlgoHashGNN.embeddingDimension`. The value is **rejected, not silently clamped**: there is no "correct" dimension to fall back to the way `Math.min(k, n)` has one for top-k, and quietly returning embeddings of a width the caller did not ask for would be the worse failure. `MAX_EMBEDDING_DIMENSION = 4096` sits comfortably above every dimension in practical use (the widest mainstream text embeddings are 3072-wide) while capping one embedding row at 32 KB. `AlgoHierarchicalClustering.numClusters` is untouched: it already clamps against `n` downstream. `AlgoMaxKCut.k` did NOT, contrary to what the issue states - see the review-cycle note below.

**Part B1 - a negative `k` is named instead of reaching the allocator.** `algo.knn(-5)` used to die in `new double[k]` with a bare `NegativeArraySizeException`, and `algo.kShortestPaths(a, b, -5)` in `new ArrayList<>(...)` with `IllegalArgumentException: Illegal Capacity`. Neither mentioned `k`. `AbstractAlgoProcedure.extractCount()` keeps the deliberate saturation ("more results than can exist" reads as "as many as exist") and adds the negative check at the same site the extraction already happens. Pre-existing behaviour, not a regression.

**Part B2 - `db.index.vector.queryNodes`' numeric-string fallback saturates.** `args[1] instanceof Number n ? saturateToInt(n) : Integer.parseInt(...)` meant the exact magnitude the `Number` branch happily saturated (`2147483648`) threw a `NumberFormatException` when it arrived as a string. Both branches now go through `NumberUtils.saturateToInt`, and a genuinely non-numeric argument gets a `CommandSQLParsingException` naming `k` rather than a raw `NumberFormatException`.

## Verification

New `Issue6065NumericParameterBoundsTest` (12 tests) plus 2 methods added to `CypherCallVectorNeighborsTest`. The regression tests were confirmed to fail against the unfixed sources: with the main-source changes stashed, **8 of the 11 new tests go red** (all four dimension caps, the zero-dimension rejection, both negative-`k` messages, and the numeric-string `k`). The other three are deliberate over-reach guards - a dimension exactly at the cap, a positive `k`, an ordinary numeric string - and pass both before and after by design.

Two of the red-run failures are the bug itself rather than an assertion mismatch: `hashGnnRejects...` took 8.2 s and `graphSageRejects...` 3.9 s before dying, because they really did try to allocate.

Full `com.arcadedb.query.opencypher.**` unit suite after cycle 3: **4391 tests, 0 failures, 0 errors, 13 skipped**.

## Review cycle 1 - `AlgoMaxKCut.k` was not clamped after all

The review disproved a claim issue #6065 made, and that the first revision of this description repeated: that `AlgoMaxKCut.k` "already clamps against `n` downstream". Checked against `AlgoMaxKCut.java`, it did not. `k` was only lower-bounded (`k < 2` rejected) and `new double[k]` is allocated **once per node per local-search pass per restart**, so `CALL algo.maxKCut(2000000000)` on a tiny graph is exactly the allocation shape this PR exists to close.

Fixed by clamping to `Math.min(rawK, n)` after the graph loads, matching `AlgoHierarchicalClustering.numClusters`. Clamping rather than rejecting is the honest reading here: a cut into more parts than there are nodes can only leave the surplus parts empty, so there *is* a correct fallback - unlike an embedding dimension. Two tests added (`maxKCutClampsAHugeKAgainstTheGraphSizeInsteadOfAllocatingForIt` plus an over-reach guard that a small `k` is untouched); the clamp test was confirmed sensitive by running it against the unclamped code with `k = 10 > n = 4`, where it fails on a community id outside `[0, 3]`.

## Test plan

- [ ] `mvn -o -pl engine -am -Dtest='Issue6065NumericParameterBoundsTest' test` - all 9 green
- [ ] `mvn -o -pl engine -am -Dtest='CypherCallVectorNeighborsTest' test` - all 9 green (7 pre-existing + 2 new)
- [ ] `mvn -o -pl engine -am -Dtest='com.arcadedb.query.opencypher.**.*Test' -DexcludedGroups=benchmark,vector,slow test` - no regression across the package
- [ ] Manual: `CALL algo.node2vec({embeddingDimension: 1000000000})` returns `algo.node2vec(): embeddingDimension must not exceed 4096, got 1000000000` promptly instead of stalling on an 8 GB allocation
- [ ] Manual: `CALL algo.knn(-5)` returns `algo.knn(): k must not be negative, got -5`
- [ ] Manual: `CALL db.index.vector.queryNodes('Doc[embedding]', '2147483648', $vec)` returns every neighbour, matching the `Number` form

## Notes for review

- `MAX_EMBEDDING_DIMENSION` is `public` on `AbstractAlgoProcedure` so the tests bind to the production constant rather than a copied literal; if the cap is ever retuned the tests follow it.
- The four procedure javadocs now state the cap via `{@value}`, so it cannot drift from the constant.
- `AlgoHashGNN` derives `numFeatures = Math.max(embDim * 4, 64)`; that multiplication would overflow `int` above ~536 M, which the 4096 cap now makes unreachable.

## Review cycle 2 - the maxKCut clamp needed a floor

The review pointed out that on a single-node graph the cycle-1 clamp `Math.min(rawK, n)` hands the local search `k = 1`, a "cut" into one partition, silently contradicting the `k >= 2` validation a few lines above it. The clamp is now `Math.max(2, Math.min(rawK, n))`; the allocation stays bounded by the node count for every graph of two nodes or more, which is the only case where the bound matters.

The regression test for it needs its own one-node database, because `algo.maxKCut` loads the whole graph. It asserts the community id **exactly** rather than by range: the only observable of a one-node cut is that single id, and `isBetween(0, 1)` would pass for `k = 1` too, so it could never fail. `new Random(42).nextInt(2)` is `1` while `nextInt(1)` can only ever be `0`, so asserting `1` is precisely the assertion that separates the floor being present from absent - verified by removing the floor and watching it go red.

**Not actioned, deliberately:** the review's style note about converging `AlgoMaxKCut`'s pre-existing `k must be ≥ 2` Unicode message on the ASCII wording the new messages use. The review itself called it non-blocking and not introduced here, and rewording a validation message that other tests may match on is unrelated churn in a fix PR.

## Review cycle 3 - the third numeric-count entry point

The review found that `db.index.vector.queryNodes` still let a negative `k` reach a bare allocator exception, the exact failure mode Part B1 closes for `algo.knn`/`algo.kShortestPaths`. Confirmed against `LSMVectorIndex.findNeighborsFromVector`: its clamp is `Math.min(k, candidateCount)` and `candidateCount` is never negative, so `-5` stays `-5` all the way to `new ArrayList<>(k)` and surfaces as `IllegalArgumentException: Illegal Capacity: -5` without ever naming `k`. `extractLimit()` now rejects it by name, mirroring `extractCount()`, with a test confirmed red when the check is removed.

`AlgoMaxKCut`'s class javadoc now documents the clamp, as suggested, mirroring the four embedding-dimension procedures.

**Not actioned, deliberately:** the review's closing note that `walksPerNode`/`walkLength`/`windowSize`/`negSamples` (`AlgoNode2Vec`), `restarts`/`maxIterations` (`AlgoMaxKCut`) and `simulations` (`AlgoInfluenceMaximization`) are unbounded multiplicative knobs of a similar shape. The review framed it as a possible follow-up rather than a request, and unlike an embedding dimension these scale CPU work rather than a single allocation, so picking a ceiling ("what is a sane maximum iteration count?") is a design decision that deserves its own issue rather than being smuggled into a fix PR. Recommend filing it as a follow-up.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AbstractAlgoProcedure.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AbstractAlgoProcedure.java
@@ -47,6 +47,18 @@ import java.util.Map;
  */
 public abstract class AbstractAlgoProcedure implements CypherProcedure {
 
+  /**
+   * Hard upper bound for every embedding-dimension-shaped parameter (`embeddingDimension`,
+   * `dimensions`, ...). These size a per-node {@code double[]} row, so the allocation grows as
+   * {@code nodeCount * dimension} with no graph-derived ceiling to clamp against - a single
+   * in-range-but-huge value is an allocation-DoS on a graph of any size.
+   * <p>
+   * 4096 sits comfortably above every dimension in practical use (the widest mainstream text
+   * embeddings are 3072-wide) while capping one embedding row at 32 KB.
+   * </p>
+   */
+  public static final int MAX_EMBEDDING_DIMENSION = 4096;
+
   // ── Embedding math utilities ─────────────────────────────────────────────
 
   /** Normalises {@code vec} to unit L2 length in-place; no-op if the vector is zero. */
@@ -168,6 +180,51 @@ public abstract class AbstractAlgoProcedure implements CypherProcedure {
     } catch (final ArithmeticException e) {
       throw new IllegalArgumentException(getName() + "(): " + paramName + " is out of range for an int: " + value, e);
     }
+  }
+
+  /**
+   * Narrows an embedding-dimension-shaped config value to {@code int} and bounds it to
+   * {@code [1, MAX_EMBEDDING_DIMENSION]}.
+   * <p>
+   * Unlike a top-k / result-count bound, which clamps against the graph's node count
+   * ({@code Math.min(k, n)}) because "as many as exist" is the natural reading, an embedding
+   * dimension has no graph-derived bound to clamp against: it multiplies the per-node allocation
+   * regardless of how small the graph is. A large but perfectly in-range value such as
+   * {@code {embeddingDimension: 1000000000}} therefore survives {@link #extractInt} and then asks
+   * for a {@code new double[n][1000000000]} (~8 GB even at {@code n == 1}).
+   * </p>
+   * <p>
+   * The value is rejected rather than silently clamped: there is no "correct" dimension to fall back
+   * to, and quietly returning embeddings of a different width than the caller asked for would be a
+   * worse failure than a clear error.
+   * </p>
+   */
+  protected int extractEmbeddingDimension(final Number value, final String paramName) {
+    final int dimension = extractInt(value, paramName);
+    if (dimension < 1)
+      throw new IllegalArgumentException(getName() + "(): " + paramName + " must be at least 1, got " + dimension);
+    if (dimension > MAX_EMBEDDING_DIMENSION)
+      throw new IllegalArgumentException(
+          getName() + "(): " + paramName + " must not exceed " + MAX_EMBEDDING_DIMENSION + ", got " + dimension);
+    return dimension;
+  }
+
+  /**
+   * Narrows a result-count bound (top-k, number of paths, ...) to {@code int}, saturating at the int
+   * bounds rather than wrapping, and rejects a negative value with a message naming the parameter.
+   * <p>
+   * Saturation is deliberate here - "more results than can possibly exist" has the sensible reading
+   * "as many as exist", and each caller clamps against the graph afterwards. A negative count has no
+   * such reading, and without this check it reaches an array/collection allocation and surfaces as a
+   * bare {@code NegativeArraySizeException} or {@code IllegalArgumentException: Illegal Capacity},
+   * neither of which names the offending parameter.
+   * </p>
+   */
+  protected int extractCount(final Number value, final String paramName) {
+    final int count = NumberUtils.saturateToInt(value);
+    if (count < 0)
+      throw new IllegalArgumentException(getName() + "(): " + paramName + " must not be negative, got " + count);
+    return count;
   }
 
   /** @see GraphEngine#getAllVertices(Database, String[]) */
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoFastRP.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoFastRP.java
@@ -41,7 +41,7 @@ import java.util.stream.Stream;
  *
  * <p>Config map parameters (all optional):
  * <ul>
- *   <li>{@code dimensions} (int, default 128) – embedding vector size</li>
+ *   <li>{@code dimensions} (int, default 128, max {@value AbstractAlgoProcedure#MAX_EMBEDDING_DIMENSION}) – embedding vector size</li>
  *   <li>{@code iterations} (int, default 4) – propagation depth</li>
  *   <li>{@code normalization} (double, default 0.0) – degree-normalization exponent α:
  *       weight of neighbour j contributing to node i is proportional to
@@ -97,7 +97,7 @@ public class AlgoFastRP extends AbstractAlgoProcedure {
     validateArgs(args);
 
     final Map<String, Object> config = args.length > 0 ? extractMap(args[0], "config") : null;
-    final int dimensions = config != null && config.get("dimensions") instanceof Number n ? extractInt(n, "dimensions") : 128;
+    final int dimensions = config != null && config.get("dimensions") instanceof Number n ? extractEmbeddingDimension(n, "dimensions") : 128;
     final int iterations = config != null && config.get("iterations") instanceof Number n ? extractInt(n, "iterations") : 4;
     final double normStrength = config != null && config.get("normalization") instanceof Number n ? n.doubleValue() : 0.0;
     final double selfInfluence = config != null && config.get("selfInfluence") instanceof Number n ? n.doubleValue() : 0.0;
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoGraphSAGE.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoGraphSAGE.java
@@ -43,7 +43,7 @@ import java.util.stream.Stream;
  *
  * <p>Config map parameters (all optional):
  * <ul>
- *   <li>{@code embeddingDimension} (int, default 64) – output embedding size per layer</li>
+ *   <li>{@code embeddingDimension} (int, default 64, max {@value AbstractAlgoProcedure#MAX_EMBEDDING_DIMENSION}) – output embedding size per layer</li>
  *   <li>{@code layers} (int, default 2) – number of aggregation layers</li>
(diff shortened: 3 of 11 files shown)


robfrank:
Cycle 4 review came back with no blocking issues. Disposition of the three non-blocking notes, none of which changes the branch:

1. **Exception type inconsistency** (`IllegalArgumentException` in `AbstractAlgoProcedure` vs `CommandSQLParsingException` in `DbIndexVectorQueryNodes`) - the review's own reading is right: each throws what its file already throws for argument errors, and `db.index.vector.queryNodes` used `CommandSQLParsingException` for every other argument error before this PR. Unifying them would mean changing the exception type callers already catch at one of the two sites, which is a behavior change that does not belong in a fix PR.

2. **Follow-up for the CPU-shaped knobs** - filed as #6216. It also records something the review did not: `AlgoNode2Vec` computes `final int totalWalks = n * walksPerNode` and allocates `new int[totalWalks][walkLen]`, so `walksPerNode` is not purely CPU-shaped and the product can overflow `int`. That part looks like a genuine defect independent of the ceiling-policy question.

3. **`AlgoKShortestPaths`' raw `(Number) args[2]` cast** - real and pre-existing, but out of scope here. It is an argument *type* guard, not a numeric-bounds problem, and the same unguarded-cast pattern appears elsewhere in the package (`(String) config.get("direction")` in four of the procedures this PR touches, for one). Fixing it at one site while leaving its siblings would be worse than leaving it consistently unfixed for a dedicated pass. Noted in #6216's neighbourhood rather than picked up mid-PR.

Merge remains with the maintainer.

PR #6222, opened 2026-08-16
fix(#6216): a graph-algorithm knob is bounded by the resource it spends, not by a guessed cap

### LINKED ISSUE #6216: Unbounded CPU-shaped config knobs in the OpenCypher algo procedures (follow-up to #6065)
## Summary

Follow-up to #6065 (PR #6214), which bounded the *allocation*-shaped numeric knobs in the OpenCypher `algo.*` package: the embedding-dimension family now caps at `AbstractAlgoProcedure.MAX_EMBEDDING_DIMENSION`, and the count-shaped parameters (`algo.knn`'s `k`, `algo.kShortestPaths`' `k`, `algo.maxKCut`'s `k`, `db.index.vector.queryNodes`' `k`) reject negatives and clamp against the node count.

Review of that PR flagged a second, distinct family that #6065 did not cover and that PR deliberately left alone: config knobs that are unbounded `int`s and multiply **CPU work** rather than a single allocation. They are the same "large but perfectly in-range value is accepted without question" shape, just with a long-running loop at the end instead of a `new double[n][dim]`.

## The knobs

- `AlgoNode2Vec` - `walksPerNode`, `walkLength`, `windowSize`, `negSamples`. Note `walksPerNode` is not purely CPU-shaped: `AlgoNode2Vec.java` computes `final int totalWalks = n * walksPerNode;` and then allocates `new int[totalWalks][walkLen]`, so it both overflows `int` for a large enough product and sizes an allocation off it.
- `AlgoMaxKCut` - `restarts`, `maxIterations`.
- `AlgoInfluenceMaximization` - `simulations`.

Each is read through `extractInt()`, so an out-of-`int` value is already rejected (#5924), but any in-range value up to `Integer.MAX_VALUE` is accepted and drives the loop count directly.

## Why it was not folded into #6065

Unlike an embedding dimension, there is no obvious ceiling: "what is a sane maximum iteration count?" depends on graph size, hardware and how long the caller is willing to wait, and a legitimate large-graph run may genuinely want a high value. So this needs a design decision rather than a constant, which is why PR #6214 stopped at the allocation family. Options worth weighing:

1. A generous hard cap per knob, in the style of `MAX_EMBEDDING_DIMENSION`.
2. A cooperative cancellation / timeout check inside the loops, so a long run is interruptible rather than forbidden - arguably the better answer for a CPU-bound knob, since it does not have to guess a ceiling.
3. Leave the knobs alone and bound only the derived allocations (e.g. `n * walksPerNode` in `AlgoNode2Vec`), which is the narrowest fix and the only part with a clear-cut bug (the `int` overflow).

At minimum, option 3's `n * walksPerNode` overflow looks like a genuine defect independent of the policy question.

## Related

#5924, #6065, PR #6055, PR #6214.

lvca:
Closes #6216

## Summary

Follow-up to #6065 (PR #6214), which bounded the *allocation*-shaped numeric knobs of the OpenCypher `algo.*` package. This closes the second family that PR deliberately left alone: knobs that are unbounded `int`s and multiply **CPU work** rather than one allocation - `algo.node2vec`'s `walksPerNode`, `walkLength`, `windowSize`, `negSamples`; `algo.maxKCut`'s `restarts`, `maxIterations`; `algo.influenceMaximization`'s `simulations`.

### Why not the issue's options as written

The issue offers three: (1) a generous hard cap per knob, (2) cooperative cancellation, (3) bound only the derived allocations. It also says why it stopped short of choosing: there is no obvious ceiling, because "what is a sane maximum iteration count?" depends on the graph, the hardware and how long the caller is willing to wait.

That is an argument against option 1 as a *general* policy, not against bounding anything. The seven knobs are not one family - they spend three different resources - so the fix applies one principle: **a knob is bounded by the resource it actually spends, not by a guess at what a user wants.** That subsumes option 3, does option 2 properly, and replaces option 1 with a heap-derived budget only where the resource really is memory.

### The parameter's own domain

Below its minimum a walk/restart/simulation count is not "a smaller run", it is an answer the algorithm cannot produce - and every one of those values was absorbed in silence or died in the allocator:

| input | before |
|---|---|
| `algo.maxKCut(2, {restarts: 0})` | never entered the restart loop: every node in community 0, `cutWeight` `-1.0`, reported as success |
| `algo.maxKCut(2, {maxIterations: 0})` | skipped the local search: the "maximum" cut was the random initialisation |
| `algo.node2vec({iterations: 0})` | returned the untrained Xavier init as an embedding |
| `algo.influenceMaximization(2, 'X', 0)` | divided the spread by zero: every candidate `NaN`, no comparison ever won, empty result set as if no node were influential |
| `algo.node2vec({walksPerNode: -1})` | `new int[-4][...]`, a bare `NegativeArraySizeException` naming nothing |
| `algo.node2vec({walkLength: 0})` | zero-width rows, then `walk[0] = v`: `ArrayIndexOutOfBoundsException` |

A new `AbstractAlgoProcedure.extractInt(value, name, minimum)` overload rejects all of them where the extraction already happens, naming the parameter, its minimum and the value. `negSamples` keeps a minimum of **0** - plain Skip-gram without negative sampling is a legitimate configuration; the rest take 1.

### Heap - the clear-cut bug, and its one-dimensional twin

`algo.node2vec` computed `final int totalWalks = n * walksPerNode` and then `new int[totalWalks][walkLen]`. That product wraps. 4 nodes at `walksPerNode: 1073741824` is exactly 2^32, so `totalWalks` came out as **0**: the matrix was allocated with no rows and the generator died on `walks[wi++]`. Other magnitudes wrap negative instead. Even without the wrap the matrix is sized by two knobs with no graph-derived ceiling - the allocation-DoS shape #6065 closed for the embedding dimensions, in two dimensions.

`algo.randomWalk` has the one-dimensional version, not mentioned in the issue but the same defect in the same walk-buffer family: `new int[steps + 1]` wraps to `Integer.MIN_VALUE` at `steps = 2147483647`.

Both are estimated in saturating `long` arithmetic and checked **before anything is allocated** against a new `arcadedb.cypher.algoMaxWalkMemory`, which auto-scales with the JVM max heap (one eighth, floor 64MB) and is tunable (negative = no limit). This is the house pattern of `arcadedb.queryMaxHeapElementsAllowedPerOp` / `arcadedb.queryMaxRangeSize`, not a new invention, and it is strictly better than a per-knob constant here: the ceiling is the machine's, the error names the knobs that produced the estimate and the setting to raise, and since it is a client parameter out of range it is an `IllegalArgumentException` - HTTP 400, not 500. A second guard, independent of the budget, refuses a walk count past `Integer.MAX_VALUE`: no heap setting makes 2^32 array entries legal.

### Time - option 2, using the setting that already exists

For the rest there is no honest ceiling, so a large value is not forbidden, it is made **abortable**. The hot loops of all three procedures call a shared `WorkGuard` that aborts on thread interruption - matching what `ShortestPathStep` and `SQLFunctionShortestPath` already do on the query path - and on the `arcadedb.command.timeout` deadline, which until now only the SQL `SELECT` planner honoured. No new setting for this: `arcadedb.command.timeout` is documented, `DATABASE`-scoped and defaults to 0, so behaviour is unchanged unless an operator has already asked for a command deadline.

The guard reads the clock only when a deadline is configured, so the default costs one flag test per iteration and no syscall. The interrupt flag is **consumed**, not restored: the exception aborts the whole call, and leaving it set would poison the next task on a pooled query thread.

### Two correctness bugs found on the way

- **`windowSize` silently disabled training.** `Math.min(walkLen - 1, pos + window)` wraps `int` for a large `windowSize` and comes back negative, so `winEnd < winStart` and the Skip-gram loop ran at position 0 and nowhere else. A window wider than the walk already spans the whole walk, so it is now clamped to `walkLength` - a clamp with a genuinely correct fallback, unlike an embedding dimension. The regression test asserts a huge window now yields *exactly* the embeddings of `windowSize = walkLength`, same seed, same RNG consumption.
- **`algo.influenceMaximization`'s negative `k`.** `k` saturates upwards on purpose and is clamped by `Math.min(k, n)`, but a negative `k` passes that clamp untouched and reaches `new int[seedCount]` as another nameless `NegativeArraySizeException`. Message wording matches PR #6214's `extractCount()` so the two converge trivially once it lands.

## Verification

New `Issue6216AlgoWorkKnobBoundsTest`, 23 tests. Confirmed sensitive: with the four procedure sources restored to `HEAD` (helpers and setting kept, so it still compiles), **17 of the 21 then-existing tests go red**. The four that stay green are the deliberate over-reach guards - a matrix inside the budget, `negSamples: 0` - and the two `WorkGuard` unit tests, which only exercise code the revert left in place. The two tests added afterwards (`...MoreWalksThanAJavaArrayCanHold...`, `...MoreStepsThanAJavaArrayCanHold...`) hit the same pre-fix `ArrayIndexOutOfBoundsException` / `NegativeArraySizeException`.

Full `com.arcadedb.query.opencypher.**` unit suite: **4397 tests, 0 failures, 0 errors, 13 skipped**.

## Test plan

- [ ] `mvn -o -pl engine -am -Dtest='Issue6216AlgoWorkKnobBoundsTest' test` - 23 green
- [ ] `mvn -o -pl engine -am -Dtest='AlgoNode2VecTest,AlgoMaxKCutTest,AlgoInfluenceMaximizationTest,AlgoRandomWalkTest' test` - no regression on the four touched procedures
- [ ] `mvn -o -pl engine -am -Dtest='com.arcadedb.query.opencypher.**.*Test' -DexcludedGroups=benchmark,vector,slow test` - no regression across the package
- [ ] Manual: `CALL algo.node2vec({walksPerNode: 1073741824})` is refused promptly by name instead of throwing `ArrayIndexOutOfBoundsException`
- [ ] Manual: `CALL algo.maxKCut(2, {restarts: 0})` returns `algo.maxKCut(): restarts must be at least 1, got 0` instead of a cut weight of `-1.0`

## Notes for review

- **Overlap with the open PR #6214 (#6065).** That PR touches `AlgoNode2Vec` and `AlgoMaxKCut` too, on different lines (`embeddingDimension`, `k`), and adds its own helpers to `AbstractAlgoProcedure`. This branch is cut from `main` and deliberately does not restate any of its fixes - `embeddingDimension` and `AlgoMaxKCut.k` are still unbounded here and stay #6214's. Whichever lands second will need a textual merge in `AbstractAlgoProcedure`; nothing is semantically in conflict.
- `arcadedb.cypher.algoMaxWalkMemory` is the only new setting. `arcadedb.command.timeout` is reused as-is.
- `WALK_ROW_OVERHEAD_BYTES = 32` prices a matrix row's header, length, padding and the enclosing reference. Walk rows are short, so this is a real part of the footprint, not a rounding error.
- `AlgoRandomWalk`'s step loop takes no `WorkGuard` check: `steps` is already bounded by the walk-memory budget and each iteration is O(1), so the loop is bounded by the allocation it already passed.


lvca:
## Review cycle 1 - the checkpoint was in the wrong loop

**Main concern: actioned, and it was right.** A checkpoint between walks bounds nothing when a single walk is the unbounded thing. `windowSize` is clamped to `walkLength`, so every position can scan the whole walk as context and one walk is O(`walkLength`²); `walkLength` is bounded only by `arcadedb.cypher.algoMaxWalkMemory`, which is a *memory* budget, not a work budget. The review's own example is exact: `walksPerNode: 1, walkLength: 5000000, windowSize: 5000000` on a small graph fits the default budget and then runs for hours between two checkpoints, so neither the deadline nor an interrupt fires in any useful time. That undercuts the PR's central claim, which is the right reason to block on it.

Fixed with `WorkGuard.checkPeriodically(counter)`: tests every 1024 iterations, so the per-iteration cost is one AND and one branch and abort latency is bounded by a fixed amount of work rather than by the loop's length. Per-position was considered and rejected - it is not enough. With `window = walkLen` a single *position* is O(`walkLength` × `negSamples` × `dim`), still seconds at the magnitudes above; the context loop is the innermost loop a knob makes long, so that is where the checkpoint belongs.

Applied to every loop whose length a knob controls, not just the one flagged:

- the Skip-gram context loop (the reported case);
- the walk-generation step loop (the review's "less urgent" note - O(`walkLength`) per check, still knob-driven);
- `algo.maxKCut`'s per-node scan inside a local-search pass, which is graph-driven rather than knob-driven but is seconds on a large graph.

The per-walk and per-pass checks stay: they cost nothing and fire first on short-walk configurations.

**New regression test.** `node2VecHonoursTheCommandTimeoutInsideOneLongWalk` covers exactly what the existing timeout test could not, and the review was right that it could not - `walkLength: 60` over 160 walks reaches a per-walk checkpoint constantly. The new one uses its own single-node database so there is exactly **one** walk and therefore exactly one per-walk checkpoint, reached before any work has happened, with a window as wide as the walk. Confirmed sensitive: with the context-loop checkpoint removed it runs to completion in **2.4 s and never times out**; with it, it aborts in milliseconds.

**Minor 1 - residual `pos + window` overflow: actioned rather than noted.** The review's own framing settles it - "bounded by the resource it spends, not by a guess" is not consistent with leaving a wrap reachable through a documented admin override. `Math.min(walkLen - 1L, pos + (long) window)` closes the class instead of the instance, at the cost of one `long` add on a path that already does a `double` dot product per iteration.

**Minor 2 - coupling with #6214: agreed, no action here.** `checkWalkBudget` deliberately prices only the walk buffers. `W`/`WCtx` are `n × embeddingDimension × 8 × 2` and `embeddingDimension` is #6214's to bound; pricing an allocation whose knob this branch does not validate would half-fix it and make the merge worse. Flagging it on both PRs is the right call - once #6214 lands, the two budgets should be looked at together.

**Nits:** all four read correctly, no action taken, and the `AlgoRandomWalk` `steps <= 0` ordering observation is right - the early return is what keeps a negative `steps` away from `walkCapacity`.

## Verification after the fix

- `Issue6216AlgoWorkKnobBoundsTest` - **24 green** (23 + the new one)
- `AlgoNode2VecTest`, `AlgoMaxKCutTest`, `AlgoInfluenceMaximizationTest`, `AlgoRandomWalkTest` - 17 green, no regression on the touched procedures
- Full `com.arcadedb.query.opencypher.**` unit suite - **4400 tests, 0 failures, 0 errors, 13 skipped**

lvca:
## Review cycle 2 - the influenceMaximization checkpoint, checked and left alone

Thanks for tracing the arithmetic by hand - the `4 * 1_073_741_824 == 2^32 -> 0` confirmation and the walk-through of the single-walk timeout test are exactly the parts worth an independent check.

**The `simulateIC` loop: checked, deliberately not converged on `checkPeriodically`.** The review asks the right question - is `simulateIC` ever cheap enough for an unthrottled clock read to matter - so here is the answer rather than an assertion.

It is not. `simulateIC` opens with:

```java
final boolean[] activated = new boolean[n];
final int[] queue = new int[n];
```

Two allocations of size `n`, both zero-filled by the JVM, on **every** call. So a simulation is Ω(n) before the cascade does anything, even on an edgeless graph with `propagationProbability` 0 - there is no sparse-graph or low-`p` shape that makes it O(1). At `n = 1` the clock read is comparable to the call; at any `n` worth running influence maximization on, the two array fills dominate it by orders of magnitude. And `check()` only reads the clock when a deadline is configured - the default is one flag test, which is not measurable against two array allocations.

Converging it on `checkPeriodically` would also make it *worse* where the deadline matters most: abort latency would become 1024 x O(n + m) cascades, seconds on a large graph. The asymmetry with node2vec's context loop is the point rather than an oversight - that loop's body is O(`dim`) of flat arithmetic with no allocation and can be a handful of flops, which is what earns it a throttle. The rule the PR applies is "throttle where an iteration can be cheaper than the check", and this loop cannot be.

Worth recording for whoever reads this later: the per-simulation allocation is itself a real cost (`simulations` x `candidates` x `rounds` arrays of size `n`) and hoisting the two buffers out of the loop with an explicit clear would be a straightforward win. It is pre-existing, unrelated to the knob-bounding this PR is about, and belongs in its own change rather than smuggled into a validation fix.

**Doc duplication nit:** agreed and noted. The three copies do serve different readers - the release note is the operator's account of what changed and why, the class Javadoc is what an API user sees at the call site, the PR description is the reviewer's. The one that must not drift is the Javadoc, since it states the parameter minimums a caller has to satisfy.

**On CI:** you flagged not being able to run `mvn` in the review sandbox. As of now `build-and-package`, `CodeQL`, `Codacy` and the analysis lanes are green; the test lanes (`unit-tests`, `opencypher-tck-tests`, `slow-unit-tests`, `integration-tests` and the e2e set) are still running and I am watching them. Locally: `Issue6216AlgoWorkKnobBoundsTest` 24 green, the four touched procedures' own suites 17 green, and the full `com.arcadedb.query.opencypher.**` unit suite 4400 tests / 0 failures / 0 errors / 13 skipped.

No code changes in this cycle.

lvca:
## Conflict with main resolved (#6214 landed)

`main` merged #6214 (#6065) at 09:05, which is the collision this branch was cut expecting. Merged `main` in; three files conflicted, all because both changes are additive in the same place.

**`AbstractAlgoProcedure`** - both sides added extractors immediately after `extractInt`, so the conflict was over a shared closing brace, not over behaviour. Kept both, and ordered them so the three argument extractors sit together under the existing `// ── Argument extractors ──` header: `extractInt(value, name, minimum)`, then `extractEmbeddingDimension`, then `extractCount`, with the `// ── Work bounds ──` section after them. Taking the merge as git left it would have interleaved the extractor group with the work-bound helpers.

**`AlgoNode2Vec`** - both sides rewrote the config-extraction block and the javadoc parameter list. Resolved to the union: `embeddingDimension` goes through `extractEmbeddingDimension` (#6065's cap), the four work knobs through `extractInt(..., minimum)` (this PR), and the javadoc list carries both the 4096 cap and the per-knob minimums. Ordering is unchanged, so `algo.node2vec({embeddingDimension: 1000000000})` is still rejected by the dimension cap before the walk-memory budget is consulted - #6065's test binds to that message and stays green.

**`docs/release-26.9.1.md`** - both appended a section at the end. Ordered by merge date, so #6216's goes last.

### One convergence taken while here

`AlgoInfluenceMaximization`'s `k` now goes through `extractCount()`. The PR description promised this: the inline `saturateToInt` + reject-negative pair written here *is* `extractCount`, and its message was deliberately worded to match so the two could converge without touching the test. The inline form and its now-unused `NumberUtils` import are gone; the regression test binds to the shared helper unedited.

`AlgoMaxKCut` needed no hand-resolution - #6065's `rawK`/`Math.max(2, Math.min(rawK, n))` clamp and this PR's `restarts`/`maxIterations` minimums plus checkpoints auto-merged, and both sets of tests pass over the result.

### Verification on the merged tree

- `Issue6216AlgoWorkKnobBoundsTest` 24, `Issue6065NumericParameterBoundsTest` 12, `CypherCallVectorNeighborsTest` 10, plus the eight touched procedures' own suites - **80 tests green together**, so neither PR's guarantees regressed the other's.
- Full `com.arcadedb.query.opencypher.**` unit suite: **4415 tests, 0 failures, 0 errors, 13 skipped** (up from 4400 - the 12 + 2 + 1 that #6214 brought in).

The PR diff is still the same 8 files; the merge leaked nothing.

lvca:
## Review cycle 3 - two actioned, one answered

**`algo.randomWalk`'s unguarded step loop: actioned, and it was worse than the review's framing.** The suggestion was a one-line caveat in the `CYPHER_ALGO_MAX_WALK_MEMORY` javadoc noting that raising the budget also raises an untimeoutable worst case. Documenting a gap is worse than closing it, especially when the gap is "this one setting has a side effect the other three procedures don't", so the step loop now carries `guard.checkPeriodically(step)` like its siblings. One AND and one branch per step.

The reasoning I had written for leaving it out ("O(1) per iteration, bounded by the budget") was wrong in the way the review identified: a memory bound is not a time bound, and the setting accepts `negative = no limit` by design. Measured rather than argued - `algo.randomWalk(a, 500000000)` with the budget disabled and a 1 ms deadline:

| | wall clock | outcome |
|---|---|---|
| without the checkpoint | **39.4 s** | runs to completion, deadline never observed |
| with the checkpoint | milliseconds | `TimeoutException` naming `arcadedb.command.timeout` |

`randomWalkHonoursTheCommandTimeoutEvenWithTheWalkMemoryBudgetDisabled` asserts that, and the 39.4 s figure is what it produces when the checkpoint is removed. The class javadoc now states the guarantee instead of the caveat.

**The "large but legitimate at scale" test gap: actioned.** Correctly identified - every other test in the class proves something is *refused*, and `node2VecRunsWhenTheWalkMatrixFitsTheBudget` only shows the shape at 2 walks of 5 steps. That is the wrong side of the risk this PR carries: it adds rejections and checkpoints to a hot path, so the failure mode worth guarding is refusing or aborting a run that is merely big.

`node2VecStillCompletesAndTrainsWithLargeButLegitimateKnobs` runs 200 walks x 50 steps x 20 epochs - two orders of magnitude past the defaults used elsewhere in the class - and asserts the **output**: one embedding per node, of the requested width, every component finite, and L2 norm 1.0 to 1e-9. Only a training loop that ran to completion produces that, so it cannot pass by not throwing.

**`simulateIC` degenerating to near-O(1): answered, no change.** Raised in cycle 2 as well, and I checked it against the source rather than assuming. `simulateIC` opens with `new boolean[n]` and `new int[n]`, both zero-filled, on every call - so it is Ω(n) before the cascade starts, on any graph shape including empty and disconnected ones at `propagationProbability` 0. There is no sparse-graph case that makes it O(1).

The two regimes both come out fine, which is why the asymmetry with node2vec is deliberate rather than an oversight:

- **large `n`** - each simulation is Ω(n), so an unthrottled check is free by comparison;
- **small `n`, large `simulations`** - each simulation is cheap, but then the check happens every simulation, so abort latency is one cheap simulation.

Throttling it to 1024 would *degrade* the first regime to 1024 x O(n + m) of abort latency, which is where the deadline matters most. The rule the PR applies is "throttle only where an iteration can be cheaper than the check", and this iteration cannot be. (The per-simulation allocation is itself a real cost worth hoisting, but it is pre-existing and belongs in its own change, not smuggled into a validation fix.)

## Verification

- `Issue6216AlgoWorkKnobBoundsTest` - **27 green** (25 + the two above)
- Full `com.arcadedb.query.opencypher.**` unit suite - **4418 tests, 0 failures, 0 errors, 13 skipped**

lvca:
## Review cycle 4 - `negSamples` was genuinely unguarded

This review landed against `00c4a64f6` and crossed with the cycle-3 push, so its point 2 (`AlgoRandomWalk`'s step loop) was already fixed in `612427b36` - independently reaching the same conclusion, including the "negative = no limit is a first-class value, so the budget is not a time bound" argument. Nothing further needed there.

**Point 1 is a real miss and is now fixed.** `negSamples` is in this PR's own list of seven knobs and had, as the review says, no protection of any kind:

| | `walkLength` / `walksPerNode` | `windowSize` | `embeddingDimension` | `negSamples` |
|---|---|---|---|---|
| heap ceiling | walk budget | - | 4096 cap (#6065) | **none** |
| maximum | - | clamped to `walkLength` | 4096 | **none** |
| checkpoint | step loop | context loop | - | **none** |

One `(position, context)` pair costs `negSamples x dim` with the nearest checkpoint in the *enclosing* context loop, which runs before the sampling starts. That is precisely the "checkpoint outside the unbounded loop" shape `c2a62bc` closed for `windowSize`/`walkLength`, and I missed this loop when making that fix. `guard.checkPeriodically(ns)` now sits inside it.

**The test needed a different shape, which is worth flagging.** For the walk case, "it throws" was a sufficient assertion because the unguarded version ran to completion and never threw. Here it is not: the unguarded version *does* throw, at the next context checkpoint, once the entire sampling loop has already run. So `node2VecHonoursTheCommandTimeoutInsideOneNegativeSamplingLoop` asserts **latency**.

The bound is set from measurement rather than taste. My first attempt used `negSamples: 20000000` with a 3 s bound and **passed with the checkpoint removed** - the loop was faster than I estimated, so the test proved nothing. Re-measured at `200000000`:

| | test method wall clock | verdict |
|---|---|---|
| with the checkpoint | **0.33 s** | green |
| without it | **24.0 s** | red, fails the bound |

5 s sits ~16x above the passing case and ~5x below the failing one - wide enough not to be a stopwatch on a slow machine, tight enough that it cannot pass while the loop is unabortable.

**Nit actioned:** `WALK_ROW_OVERHEAD_BYTES` now says in its javadoc that it is a heuristic, not a guarantee - the true per-row cost moves with the JVM object layout (compressed oops, alignment), the estimate only has to stay in the right order of magnitude, and there is nothing to "correct" there short of measuring a specific JVM. Exactly the drive-by "fix" the review was guarding against.

## Verification

- `Issue6216AlgoWorkKnobBoundsTest` - **28 green**
- Full `com.arcadedb.query.opencypher.**` unit suite - **4419 tests, 0 failures, 0 errors, 13 skipped**

Every knob in the PR's list now has a checkpoint inside the loop it drives: walk generation (`step`), Skip-gram context (`ctx`), negative sampling (`ns`), maxKCut's per-node scan (`i`), randomWalk's step loop (`step`), plus the coarse per-walk / per-pass / per-simulation checks.

lvca:
### Note on ordering: the `negSamples` review crossed with its fix

The review above ran against `612427b` and re-flagged `negSamples`. It is already closed in `47dddb557`, pushed a couple of minutes later, and the comment immediately preceding this one is the response to it - the reviews and the fixes have been leapfrogging each other for a few cycles now.

Worth recording that the two arrived at the same fix independently: `guard.checkPeriodically(ns)` inside the sampling loop, plus "a regression test analogous to `node2VecHonoursTheCommandTimeoutInsideOneLongWalk` but driven by a large `negSamples` on a small walk/window (so it isolates this loop rather than the already-covered context loop)" - which is exactly the shape of `node2VecHonoursTheCommandTimeoutInsideOneNegativeSamplingLoop` (`walkLength: 2`, `windowSize: 1`, `negSamples: 200000000`).

The one thing the suggestion does not mention, and which the first attempt at the test got wrong: that analogous test cannot assert only "it throws". The long-walk case is red because the unguarded version *completes*; the sampling case is not, because the unguarded version does throw at the next context checkpoint, just 24 s late. The assertion has to be on latency, and my first version at `negSamples: 20000000` passed with the checkpoint removed before I re-measured and raised it.

HEAD is `47dddb557` and the review of it is in flight. No open points that I am aware of.

lvca:
## Review cycle 5 - no blocking issues; two notes actioned, one answered in the code

**The shuffle note was under-rated, and I acted on it as a gap rather than an FYI.** It was flagged as "not a real gap since its cost is O(totalWalks), and `totalWalks` is already bounded by `CYPHER_ALGO_MAX_WALK_MEMORY`". That is the same argument cycle 3 rejected for `algo.randomWalk`'s step loop, and it fails here for the same reason: the budget is tunable and **a negative value means no limit**, which leaves `totalWalks` capped only by `Integer.MAX_VALUE`. A per-epoch Fisher-Yates over 2^31 entries is an unabortable loop, not a rounding error.

Rather than fix only the loop named, I swept the rest of the procedure by the same rule. The Fisher-Yates shuffle, the `walkOrder` initialisation and the Xavier initialisation now all carry `guard.checkPeriodically(i)`. Every O(nodeCount) / O(totalWalks) loop in `AlgoNode2Vec` has a checkpoint now, so the rule holds with no exceptions anyone has to remember - which is the actual fix for "in case a future change decouples `totalWalks` from that budget".

**`AlgoInfluenceMaximization`'s unthrottled `check()`: deliberate, and now says so at the call site.** A fair catch that the reasoning existed only in review replies, where nobody will find it in a year. The comment now states it: `simulateIC` allocates and zero-fills two arrays of size `n` before the cascade starts, so it is never O(1) on any graph shape; throttling to 1024 would make abort latency `1024 x O(n + m)`, worst on exactly the large graphs where the deadline matters most.

**`window = Math.min(rawWindow, walkLen)` vs `walkLen - 1`: left as is, with the reason recorded.** `winStart`/`winEnd` are clamped to `[0, walkLen - 1]` regardless, so the tighter bound is unobservable - and at `walkLength: 1` it would drive `windowSize` to 0, under the minimum of 1 the extraction enforces. Documented so it does not get "tightened" later.

**On the latency assertion and CI flakiness.** Worth flagging, and the margins are the answer: both figures scale together on a slower runner, so it takes a machine ~15x slower than the reference to push the passing case to 5 s (where the failing case would sit at ~360 s), or ~5x faster to bring the failing case under it. The comment now carries the instruction that matters if it ever does flake: **raise the bound, do not weaken it to "it throws"** - that form passes with the checkpoint removed, which is exactly how the first version of this test was wrong.

## Verification

Full `com.arcadedb.query.opencypher.**` unit suite - **4419 tests, 0 failures, 0 errors, 13 skipped**.

lvca:
## Review cycle 6 - nothing blocking; the throttle comment corrected

Thanks for tracing the `walksPerNode=Integer.MAX_VALUE, walkLength=Integer.MAX_VALUE` arithmetic and the narrow-window `ctx` coverage by hand - the second one in particular is the question I would have wanted checked, since `checkPeriodically(ctx)` looks like it could leave gaps when the `ctx` loop body is only a few iterations wide, and the sliding-window argument for why it does not is not obvious from the code.

**The `ns` throttle note is right, and the comment was the thing at fault.** `ns` restarts at 0 for every (position, context) pair, so the check fires once per pair rather than once per 1024 samples. The behaviour is correct and stays - one flag test per pair is exactly what keeps the default `negSamples: 5` responsive, because as your own analysis shows the enclosing `ctx` checkpoint fires only about once every 1024 **positions** when the window is narrow, not once per pair. So the two are not redundant: the `ns` one is strictly the more frequent, and it is carrying the common case.

What was wrong is that the javadoc described a monotonically climbing counter as though that were the only shape. A reader who trusted it would conclude the `ns` check is redundant with the `ctx` one and delete it - which is precisely the regression cycle 4 fixed. `checkPeriodically`'s javadoc and the call site now state the restart case and why it is deliberate. Comments only, no functional change (`69ba8ae`).

**Javadoc verbosity:** taken as read. It is heavier than typical because most of it is recording *why* a bound is where it is - the alternative that was rejected, the measurement behind a constant, the reason a clamp is safe here and would not be elsewhere. That is the part that decays fastest into folklore otherwise, and this PR is on its second round of someone (me included) proposing to "simplify" something that was load-bearing.

## Status

Six review cycles: four found something real (checkpoint placement in the Skip-gram context loop, `algo.randomWalk`'s step loop, the `negSamples` loop, the shuffle/init loops), two found nothing blocking. Every point raised has been either actioned or answered with the reasoning in the code rather than only here.

Full `com.arcadedb.query.opencypher.**` unit suite at the last functional change: **4419 tests, 0 failures, 0 errors, 13 skipped**. Ready for a maintainer's call.

Issue #6133, opened 2026-08-13
"Are this database's pages flushed?" is an O(n) scan of the JVM-wide page index, polled every 10 ms on close, rename, backup and the snapshot barrier


lvca:
`PageManagerFlushThread.countPendingPagesOfDatabase` answers "does this database still have pages in the flush pipeline?" by walking the **JVM-wide** `pageIndex` - every pending page of every open database - and comparing each key's database:

```java
private int countPendingPagesOfDatabase(final Database database, final boolean stopAtFirst) {
  int pending = 0;
  for (final PageId key : pageIndex.keySet())
    if (database.equals(key.getDatabase())) {
      ++pending;
      if (stopAtFirst)
        return pending;
    }
  return pending;
}
```

`waitAllPagesOfDatabaseAreFlushed` calls it in a **10 ms poll loop for the whole duration of the drain**, with `stopAtFirst = false` so it never short-circuits, and every close, rename, compaction shipping, backup suspend and (since #6125) snapshot t0 barrier goes through that loop. The cost of asking the question therefore scales with the total backlog across all databases, and the answer is needed most often exactly when that backlog is large.

Two things make it worth fixing rather than tolerating:

- **The per-database signal is not per-database work.** `flushedPagesPerDatabase` already exists precisely so a busy sibling database cannot mask a wedged one (#4928) - the design intent is per-database accounting, but the pending-page half of it is still a global scan.
- **#6125 put one of these loops under the JVM-wide page-manager lock.** The barrier's residual drain polls this while every committer in the process waits behind it. That PR's review raised the same point about a spin loop I had written there; I removed the spin (1 ms poll instead of up to 1000 `onSpinWait` iterations, each an O(n) scan), which fixed the immediate multiplier but not the underlying scan.

The obvious answer is a per-database `AtomicInteger` of pending pages, incremented in `scheduleFlushOfPages` and decremented wherever a page leaves the pipeline. **That is exactly why this is an issue rather than a patch:** the removal sites are `removeFromFlushIndex` (conditional on reference identity, the #4544 two-instance case), `removeAllPagesOfDatabase`, `removeAllPagesOfFile`, `removePagesOfFileFromBatch` and `detachPendingPages`, and a counter that drifts high hangs `close()` forever while one that drifts low lets a backup take its t0 over a half-written batch. I declined it during #6126's review for that reason - it is the right fix, but it needs its own change with its own tests, not a footnote in a barrier PR.

[...]

Issue #4544, opened 2026-06-11
`BasePage.equals/hashCode` mismatch breaks `pageIndex.remove(key, value)`


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.engine.BasePage`
**Severity:** MEDIUM

### Summary

`equals` distinguishes pages by `(pageId, version)`; `hashCode` keys on
`pageId` only. This is technically legal, but `MutablePage.version` is
mutated *while the page is a value in a `ConcurrentHashMap`* (used by
the flush queue's `pageIndex`). After `incrementVersion`, the same
instance's `equals` semantics change — violating the immutability
contract that `Map.remove(key, value)` relies on.

### Code reference

`engine/com/arcadedb/engine/BasePage.java:183–200`
Use sites in `engine/com/arcadedb/engine/PageManagerFlushThread.java:~189`.

### Impact

A stale dequeue can match a newer entry (version compared against an
outdated baseline) and remove it from `pageIndex`. Subsequent reads via
`getCachedPageFromMutablePageInQueue` miss the page and return the
on-disk older version, leading to spurious
`ConcurrentModificationException` on the writer that committed the
newer version.

### Suggested fix

Make `MutablePage` map identity reference-based (`equals/hashCode` on
`pageId` only) and use a separate sentinel for `remove(key, value)`, or
switch to a per-batch identifier for value-based remove.

lvca:
Fixed

PR #6223, opened 2026-08-16
fix(#6199, #6200): the flush drains wake on the page, and one suspended database's backlog is its own

### LINKED ISSUE #6199: The flush drain waits still poll, so the snapshot barrier holds the JVM-wide page-manager lock across 1 ms sleeps that #6133 made unnecessary
Follow-up to #6133, which made "does this database still have pages in the flush pipeline?" an O(1) counter read. The *cost of one check* is now negligible, but both callers still discover the answer by **sleeping and asking again**, so the latency they pay is set by the poll interval rather than by when the pages actually land.

## Where

**1. The bulk drain**, `PageManagerFlushThread.waitAllPagesOfDatabaseAreFlushed` (`engine/src/main/java/com/arcadedb/engine/PageManagerFlushThread.java:209`), polls every `DEFAULT_FLUSH_WAIT_POLL_MILLIS = 10` ms (line 64/240). Every `close()`, `rename()`, index compaction, backup suspension and the snapshot barrier's step 1 goes through it, and each of them rounds up to the next 10 ms boundary even when the pipeline was already empty or emptied 9 ms ago. On a process cycling through many databases that is pure dead time.

**2. The residual drain**, `waitPendingPagesOfDatabaseUntil` (line 398), polls every **1 ms** - and its caller, `PageManager.openSnapshot` (`engine/src/main/java/com/arcadedb/engine/PageManager.java:560`), runs it with **both the database's apply write lock and the JVM-wide page-manager lock held**. That is deliberate and correct (publication is excluded, so the drain converges by construction), but it means every `Thread.sleep(1)` there is a millisecond in which **no committer of any database in the JVM can publish a page**. The method's own javadoc argues for the 1 ms sleep over a spin, and that trade-off is right as long as polling is the only option - it is the polling itself that is now avoidable.

## Why this is newly cheap to fix

Before #6133 the decrement was scattered across five call sites, so there was no single place to hook a "this database just drained" signal without risking a missed or duplicated notify. After #6133 there is exactly one: `FlushPageIndex.release()` (`engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java:229`) is the only path on which a database's pending count goes down, and `removeAllOfDatabase` is the only other way an entry leaves.

## Proposed

Give `FlushPageIndex` a per-database monitor (the same shape as the existing `suspendLocks` / `replayDrainLocks` / `snapshotBarrierLocks` maps) and `notifyAll()` from `release()` on the count reaching zero, plus from `removeAllOfDatabase`. Both waits then become `lock.wait(interval)` instead of `Thread.sleep(interval)`:

- they wake **the moment** the last page lands, instead of up to a full interval later;
- the bounded `wait()` keeps the existing timeout machinery working unchanged - `waitAllPagesOfDatabaseAreFlushed`'s no-progress window still needs periodic re-evaluation, and `waitPendingPagesOfDatabaseUntil` still needs its deadline check, so neither becomes an unbounded park;
- the snapshot barrier's JVM-wide lock hold drops from "number of residual pages, rounded up to whole milliseconds" to "the time the writes actually take".

Alternative considered and rejected: shortening `DEFAULT_FLUSH_WAIT_POLL_MILLIS`. It trades dead latency for CPU burn on the exact path that already holds a JVM-wide lock, and it cannot get below one scheduling quantum, so it improves the average without touching the worst case.

## Verification

- A test that measures the wake latency of `waitAllPagesOfDatabaseAreFlushed` against a page that is released mid-wait: today it is quantised to the poll interval, after the change it should track the release.
- The existing `Issue6133PendingPagesPerDatabaseTest`, `PageManagerFlushSuspend*Test` and `FlushRobustnessTest` must stay green - the notify must not be able to be missed between the counter reaching zero and the waiter parking (take the monitor before the final count read, standard guarded-wait shape).

### LINKED ISSUE #6200: The deferred-flush backpressure of #4728 is JVM-wide: one suspended database's backlog stops the flush thread from draining the queue for every other database
Noticed while working on #6133, which removed one instance of the pattern "one database's backlog is charged to every other database in the JVM". Here is another, in the deferred-flush backpressure added by #4728.

## The shape

`PageManagerFlushThread.deferredRAMBytes` (`engine/src/main/java/com/arcadedb/engine/PageManagerFlushThread.java:112`) is a **single JVM-wide** `AtomicLong`, even though what it accumulates is strictly per-database: `deferredByDatabase` (line 68) holds the batches of the databases that are currently *suspended* (HA snapshot ship, full backup), and only those batches are ever added to it (line 284).

The gate that consumes it is not per-database either (line 256, `flushPagesFromQueueToDisk`):

```java
if (running && maxDeferredRAM > 0 && deferredRAMBytes.get() >= maxDeferredRAM) {
  Thread.sleep(timeout);
  return;
}
```

That `return` skips the `queue.poll` **entirely**. So once one suspended database's deferred backlog crosses `arcadedb.flushSuspendMaxDeferredRAM` (512 MB by default), the flush thread stops draining the shared queue for *everybody*: the bounded queue fills, and `scheduleFlushOfPages` then throttles the committing threads of **every open database**, including databases that are not suspended and whose pages could be written to disk immediately.

Concretely: a leader shipping a multi-GB snapshot of database A stalls the committers of unrelated databases B and C on the same server, for as long as A's backlog stays above the cap.

## Why the cap itself is right and only its scope is wrong

The cap exists to bound **heap**, and heap is genuinely a JVM-wide resource, so a global ceiling is the correct thing to measure. What does not follow is the response: batches of *non-suspended* databases do not add to `deferredRAMBytes` at all, so flushing them cannot make the heap situation worse - it strictly improves it, since it moves pages out of the queue and lets the page cache release them. Refusing to poll is therefore giving up throughput for no memory benefit.

## Proposed

Keep the global ceiling as the trigger, but make the *back-off* selective:

1. Poll the queue as usual instead of skipping the poll.
2. If the polled batch belongs to a database that is **not** suspended, flush it normally - it cannot grow the deferred backlog.
3. Only when the polled batch belongs to a suspended database *and* the cap is crossed, apply the current behaviour (put it back / back off), which is the case the cap was written for.

Step 3 needs care so the batch is not lost and the loop does not spin on the same head-of-queue batch; the existing `deferredByDatabase` re-enqueue machinery in `unsuspend` (line ~619) is the model to follow, and `PagesToFlush` already carries its database.

The natural companion is to split `deferredRAMBytes` into a per-database counter plus a derived total. That also unblocks item 2 of #6087, which asks for a `arcadedb.pagemanager.deferred_ram_bytes` gauge **tagged by database** - not expressible today, because the number simply does not exist per database.

## Verification

- A test with two databases, A suspended and pushed past a deliberately small `arcadedb.flushSuspendMaxDeferredRAM`, asserting that B's commits still complete promptly instead of blocking on a full queue. Today B stalls; it should not.
- `PageManagerFlushSuspendBackpressureTest` must keep proving that A's own committers *are* still throttled - the cap must not be weakened for the database it applies to.
- `PageManagerFlushSuspendDeferRaceTest` and `PageManagerFlushSuspendRefCountTest` cover the re-enqueue and refcount paths that step 3 touches.

## Not a bug, checked while here

The two `deferredRAMBytes.set(0)` resets (lines 640 and 720) look like they could zero another database's accounting, but both are guarded by `deferredByDatabase.isEmpty()`, so they only fire when the backlog is provably empty. They are safety nets, not drift.

lvca:
Fixes #6199, fixes #6200. Both are follow-ups of #6133.

## #6199 - the drains stop polling

Both waits on the flush pipeline discovered that it had emptied by **sleeping and asking again**, so the latency they paid was set by the poll interval rather than by when the pages actually landed:

- `waitAllPagesOfDatabaseAreFlushed` polls every 10 ms, and every `close()`, `rename()`, index compaction, backup suspension and the snapshot barrier's first drain goes through it;
- `waitPendingPagesOfDatabaseUntil` polls every **1 ms**, and its caller (`PageManager.openSnapshot`, step 2 of the t0 barrier) runs it with **both the database's apply write lock and the JVM-wide page-manager lock held**. Every one of those `Thread.sleep(1)` calls was a millisecond in which no committer of *any* database in the process could publish a page - paid even when the pages had landed 999 us earlier.

Since #6133, `FlushPageIndex.release()` is the only path on which a database's pending count goes down, so a "this database just drained" signal has exactly one hook. The per-database counter now doubles as the monitor: `release()` notifies on the transition to zero (and `removeAllOfDatabase` notifies on a purge), and both waits call `FlushPageIndex.awaitDrain` instead of sleeping.

The guard is re-read **inside** the monitor, so a release landing between the caller's check and the park cannot be slept through. The bounded `wait` is kept, not as the mechanism but as the bound on a hypothetical lost notification and to keep the callers' timeout machinery re-evaluated - which is why the residual drain's fallback stays at the 1 ms it used to sleep: **worst case unchanged, common case immediate**.

Rejected, as the issue suggests: shortening the poll interval. It burns a core under a JVM-wide lock to improve the average without touching the worst case.

## #6200 - the cap stops charging one database's backlog to everyone

The cap of #4728 bounds **heap**, which is JVM-wide, so a global ceiling is the right thing to measure. What did not follow is that the *response* should be global too, and it was: once one suspended database's backlog crossed the cap, `flushPagesFromQueueToDisk` returned **without polling at all**. The bounded queue then filled and `scheduleFlushOfPages` throttled the committing threads - of every open database, not just of the suspended one - and, because `publishPages` holds the JVM-wide page-manager lock across the enqueue, it throttled them *inside that lock*. A leader shipping a multi-GB snapshot of database A stalled the committers of unrelated databases B and C, whose pages could have gone straight to disk, which strictly **relieves** the heap the cap protects.

### What this PR does instead of the issue's proposal

The issue proposed keeping the gate on the drain side and making the back-off selective (poll as usual, flush non-suspended batches, put the over-cap suspended batch back). That would have to re-queue the held batch, and a batch put back at the tail is **reordered against the other batches of its own database**: two commits can leave two `MutablePage` copies of the same `PageId` pending at once (the #4544 case), and deferring the newer one first means the older one is written over it at resume. It also cannot bound the heap on its own - the pages weigh exactly the same in the queue as in the deferred map, so moving them around does not stop the growth; only holding the producers does.

So the throttle moved to the producer side, where it belongs, and the drain-side gate is gone entirely:

- `PageManagerFlushThread.awaitDeferredBacklogUnderCap(database)` holds a committer while the backlog is at or over the cap **and its own database is suspended**. A database that is not suspended never enters it - its pages cannot grow the backlog.
- It is called **before the page-manager lock is taken**: as a separate step of `publishPages`, and directly in `writePages` for the asynchronous compaction writes that do not go through `publishPages`. This is the whole point - held one line lower, the wait would be served inside a JVM-wide lock and we would be back to stalling everybody.
- The flush thread always drains, so B's and C's batches are polled and written while A sits over the cap.
- `deferredRAMBytes` is split into a per-database counter plus the derived total (`getDeferredRAMBytesOf`), which is what item 2 of #6087 needs for a gauge tagged by database - not expressible before, because the number did not exist per database.
- A committer held for more than 10 s says so once, at WARNING, naming the database whose backlog is holding it.

The heap bound is unchanged in the worst case (cap plus whatever was already admitted), and tighter in practice: while a database is suspended, deferring a batch is a map insert with no I/O, so the queue does not build up and the total tracks the cap closely.

## Tests

- `Issue6199DrainWakeupTest` (new) - stretches the fallback interval to a minute so a wait that ends inside the assertion timeout can only have been ended by the signal; races the release against the park 300 times to pin the guarded-wait shape; covers the purge wake-up and the barrier's hard deadline (including the `wait(0)` trap an expired deadline would otherwise be). Verified to fail on the pre-fix behaviour: 3 of 5 tests red with the notify removed.
- `Issue6200PerDatabaseDeferredBacklogTest` (new) - a second database's batch is flushed while the first sits at twice the cap; the suspended database's own committers are held and then released by the resume; a non-suspended database is never held; the accounting splits per database and sums to the total. Verified to fail on the pre-fix behaviour: 3 of 4 tests red with the drain-side gate restored.
- `PageManagerFlushSuspendBackpressureTest` - #4728's own bound, re-asserted where the enforcement moved: past the cap, no further page of the suspended database can be published.
- Full `engine` module suite green (11998 tests), including `Issue6133PendingPagesPerDatabaseTest`, `Issue6125SnapshotBarrierAndCapTest`, `PageManagerFlushSuspendDeferRaceTest`, `PageManagerFlushSuspendRefCountTest`, `Issue5326ApplyChangesPendingFlushTest`, `FlushRobustnessTest` and `PageSnapshotTest`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -472,10 +472,12 @@ public enum GlobalConfiguration {
       """
       Maximum amount of RAM (in MB) of dirty pages the page-flush thread may defer in memory while flushing \
       is suspended (during an HA snapshot ship or a full backup, when the on-disk files must stay stable). \
-      Once the deferred backlog crosses this cap the flush thread stops draining its bounded queue, so \
-      committing threads are throttled instead of the deferred backlog growing without limit and exhausting \
-      the heap (issue #4728: a busy leader shipping a multi-GB snapshot OOM'd). Set to 0 to disable the cap \
-      (unbounded, pre-4728 behavior).""",
+      Once the deferred backlog crosses this cap the committing threads of the SUSPENDED databases are \
+      throttled, instead of the deferred backlog growing without limit and exhausting the heap (issue #4728: \
+      a busy leader shipping a multi-GB snapshot OOM'd). The cap is JVM-wide because the heap it bounds is, \
+      but the throttling is not: a database that is not suspended is never held by it, since its pages go \
+      straight to disk and relieve the backlog rather than add to it (issue #6200). Set to 0 to disable the \
+      cap (unbounded, pre-4728 behavior).""",
       Long.class, 512),
 
   PAGE_SNAPSHOT_ENABLED("arcadedb.pageSnapshotEnabled", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
+++ b/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
@@ -49,6 +49,13 @@ import java.util.concurrent.atomic.AtomicInteger;
  * of indexed pages, so {@code pendingOf() == 0} proves the pipeline is empty (what the barrier and the close path
  * need), while the opposite skew is bounded by a few instructions and costs at worst one extra poll.
  * <p>
+ * <b>The drain signal (issue #6199).</b> Because {@link #release} is the ONE place a database's count goes down, it
+ * is also the one place that can tell a waiter "your pipeline just emptied". {@link #awaitDrain} parks on the
+ * database's own counter instead of sleeping, so the two drains ({@code waitAllPagesOfDatabaseAreFlushed} and the
+ * snapshot barrier's {@code waitPendingPagesOfDatabaseUntil}) wake the moment the last page lands rather than at the
+ * next poll boundary. That matters most for the barrier, which polls with the JVM-wide page-manager lock held: every
+ * millisecond it sleeps there is a millisecond in which no committer of ANY database can publish a page.
+ * <p>
  * <b>The one thing that rule does NOT cover, and what covers it instead:</b> {@link #removeAllOfDatabase} drops a
  * database's counter outright, so a {@link #putAll} for that same database running concurrently with it could leave a
  * page indexed under a counter that is no longer there - undercounting to zero, the "a backup stamps its t0 over a
@@ -65,6 +72,12 @@ class FlushPageIndex {
   /**
    * Number of {@link #pages} entries per database. An entry is created with the database's first pending page and
    * dropped by {@link #removeAllOfDatabase} (close/drop), so it is bounded by the number of open databases.
+   * <p>
+   * The counter object doubles as the monitor {@link #awaitDrain} parks on and {@link #release} notifies (issue
+   * #6199). Reusing it rather than adding a parallel map of lock objects is not a shortcut: it removes the window in
+   * which a waiter could resolve a monitor that the signaller has not published yet, and it keeps the signalling
+   * path down to the counter the decrement already has in hand. The object never escapes this class - every accessor
+   * returns an {@code int} - so no foreign code can hold this monitor.
    */
   private final ConcurrentHashMap<BasicDatabase, AtomicInteger> pending = new ConcurrentHashMap<>();
 
@@ -164,7 +177,13 @@ class FlushPageIndex {
     pages.keySet().removeIf(pageId -> database.equals(pageId.getDatabase()));
     // Dropped AFTER the pages, never before: the other order would let a page indexed in between survive with a
     // fresh counter that the purge then empties, leaving a count above zero that hangs the close forever.
-    pending.remove(database);
+    final AtomicInteger counter = pending.remove(database);
+    // A waiter parked on that counter would otherwise sleep out its whole fallback interval for a database that no
+    // longer has an entry at all: pendingOf() answers 0 for it from here on, so wake it to observe that (#6199).
+    if (counter != null)
+      synchronized (counter) {
+        counter.notifyAll();
+      }
   }
 
   /**
@@ -221,6 +240,42 @@ class FlushPageIndex {
     return count;
   }
 
+  /**
+   * Parks until the database's pipeline is empty, or until {@code maxWaitMillis} elapses - the poll-free form of the
+   * drains of issue #6199.
+   * <p>
+   * <b>Standard guarded wait, and the guard is the whole point:</b> the count is re-read HERE, with the monitor
+   * already held, and not by the caller before the call. A caller that read a positive count and then asked to park
+   * would race the last {@link #release}: the decrement and its {@code notifyAll} could both land in the gap, and
+   * the wait would then be woken only by its own timeout. Re-reading under the monitor closes that gap in both
+   * directions - either this read sees the drained count and returns without parking, or it happens before the
+   * decrement, in which case the signaller cannot enter its {@code synchronized} block until this thread is already
+   * inside {@code wait()} and has released the monitor.
+   * <p>
+   * The bounded wait is a safety net, not the mechanism: it keeps the callers' timeout machinery (the no-progress
+   * window of {@code waitAllPagesOfDatabaseAreFlushed}, the hard deadline of the snapshot barrier) periodically
+   * re-evaluated, and degrades a hypothetical lost notification to the polling this used to be rather than to a
+   * hang. Callers therefore must keep their own loop and re-check their own condition on return.
+   *
+   * @param maxWaitMillis upper bound of the park. {@code <= 0} returns immediately - it never means "park forever",
+   *                      which is what a bare {@code Object.wait(0)} would do.
+   */
+  void awaitDrain(final BasicDatabase database, final long maxWaitMillis) throws InterruptedException {
+    if (maxWaitMillis <= 0)
+      return;
+
+    final AtomicInteger counter = pending.get(database);
+    if (counter == null)
+      // NEVER HAD A PENDING PAGE, OR THE DATABASE WAS ALREADY FORGOTTEN: pendingOf() ANSWERS 0, NOTHING TO WAIT FOR
+      return;
+
+    synchronized (counter) {
+      if (counter.get() <= 0)
+        return;
+      counter.wait(maxWaitMillis);
+    }
+  }
+
   private AtomicInteger counterOf(final BasicDatabase database) {
     final AtomicInteger counter = pending.get(database);
     return counter != null ? counter : pending.computeIfAbsent(database, k -> new AtomicInteger());
@@ -228,7 +283,11 @@ class FlushPageIndex {
 
   private void release(final BasicDatabase database) {
     final AtomicInteger counter = pending.get(database);
-    if (counter != null)
-      counter.decrementAndGet();
+    // The monitor is taken ONLY on the transition to drained, so the hot flush path pays a volatile decrement and a
+    // comparison per page exactly as before, and a lock acquisition only on the page that empties the pipeline.
+    if (counter != null && counter.decrementAndGet() <= 0)
+      synchronized (counter) {
+        counter.notifyAll();
+      }
   }
 }
--- a/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
+++ b/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
@@ -182,6 +182,13 @@ class FlushPageIndex {
     // longer has an entry at all: pendingOf() answers 0 for it from here on, so wake it to observe that (#6199).
     if (counter != null)
       synchronized (counter) {
+        // ZEROED, not merely notified, and inside the monitor. A waiter that resolved this counter BEFORE the
+        // remove above can enter the monitor AFTER the notifyAll has already fired, and what it re-reads there is
+        // the counter object itself - the drained map entry it would otherwise consult is already gone. So the
+        // object has to carry the drained state, or that waiter parks through a signal it can never be sent again.
+        // Bounded rather than fatal (its fallback interval still expires, and its next call resolves no counter at
+        // all), but it is precisely the wake-up this notification exists to deliver.
+        counter.set(0);
         counter.notifyAll();
       }
   }
--- a/engine/src/main/java/com/arcadedb/engine/PageManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/PageManager.java
@@ -901,7 +901,8 @@ public class PageManager extends LockContext {
   /**
    * Bytes of dirty pages currently held in memory because flushing is suspended on some database (a full backup, an HA
    * snapshot ship, an HA verify). Once this crosses {@link com.arcadedb.GlobalConfiguration#FLUSH_SUSPEND_MAX_DEFERRED_RAM}
-   * the flush thread stops draining its queue and committing threads are throttled instead, so this is the direct
+   * the committing threads OF THE SUSPENDED DATABASES are throttled (issue #6200 - before it, the flush thread stopped
+   * draining its queue altogether, which throttled the committers of every open database), so this is the direct
    * measure of how much a long suspension is costing writers.
    *
    * @return 0 when no suspension is in progress, or when the page manager is closed.
@@ -909,6 +910,28 @@ public class PageManager extends LockContext {
   public long getDeferredRAMBytes() {
     final PageManagerFlushThread thread = flushThread;
     return thread != null ? thread.deferredRAMBytes.get() : 0L;
+  }
+
+  /**
+   * The same backlog as {@link #getDeferredRAMBytes}, for a single database (issue #6200): the number that says WHICH
+   * suspension is costing the heap, which the JVM-wide total cannot.
+   *
+   * @return 0 when that database has nothing deferred, or when the page manager is closed.
+   */
+  public long getDeferredRAMBytesOf(final BasicDatabase database) {
+    final PageManagerFlushThread thread = flushThread;
+    return thread != null ? thread.getDeferredRAMBytesOf(database) : 0L;
+  }
+
+  /**
+   * Holds the caller while the deferred backlog of ITS database is over the cap (issue #4728). A no-op unless that
+   * database is currently suspended - see {@link PageManagerFlushThread#awaitDeferredBacklogUnderCap}, and note that
+   * this must never be called with the page-manager lock held (issue #6200).
+   */
+  private void awaitDeferredBacklogUnderCap(final List<MutablePage> pages) throws InterruptedException {
+    if (flushThread == null || pages == null || pages.isEmpty())
+      return;
+    flushThread.awaitDeferredBacklogUnderCap(pages.getFirst().getPageId().getDatabase());
   }
 
   /**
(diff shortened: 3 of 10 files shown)


lvca:
Answering the close/drop-while-suspended question from the last review round rather than confirming it as a non-issue: **the window is real**, and it is now closed in e47f11edb.

`removeAllPagesOfDatabase` cleared `suspended` and detached `deferredByDatabase` while holding nothing, but `flushPagesFromQueueToDisk` decides to defer *inside* `synchronized (suspendLock(db))` - it checks `isSuspended` and offers the batch in the same critical section. So the flush thread could read "suspended" just before the purge ran and offer its batch just after it, re-creating a deferred queue for a database the purge had already forgotten. Since nothing ever resumes a database that is gone, everything in that window is permanent: the batch, the map entry pinning the closed `LocalDatabase` instance with its pages in RAM, and a charge on the JVM-wide total that would throttle the committers of every later suspension in the process.

Nothing else rules it out - the database write lock a close takes only excludes suspenders that hold the matching read lock for their whole window, which is not a property this class can rest on.

Fix: the flag drop and the detach now happen under `suspendLock(database)`, flag first, which is exactly the construction Phase 2 of the resume already relies on (once the flag is clear, no further batch can be deferred). `suspendLocks.remove` moved after that block, or a defer racing it would resolve a *fresh* monitor and be excluded by nothing; the `batchRAM` accounting stays outside it, since `batchRAM` takes each batch's own monitor and `suspendLock` → `batch.pages` is the one nesting this class allows.

Two tests: `purgingASuspendedDatabaseTakesItsWholeBacklogWithIt` pins the accounting deterministically (no queue, no charge, no flag, and a sibling that is still suspended keeps its own backlog), and `aBatchPolledWhileTheDatabaseIsPurgedIsNeverLeftDeferredForIt` races a poll against a purge 200 times. The race test **reproduces the leak against the unguarded version** - about one run in three, around iteration 100 - so it is a real guard.

On the other nit: leaving `DEFERRED_BACKLOG_WAIT_POLL_MILLIS` and `DEFERRED_BACKLOG_WARN_MILLIS` non-configurable for now, as suggested.

Full engine suite green: 12002 tests, 0 failures.

lvca:
Round 4's double-release finding was **real and reproducible** - thanks, that one was well traced. Fixed in 21da075f4, but the suggested fix turned out not to be enough:

- Skipping `addDeferredRAMTotal` when `bytes == null` still failed the new test at iteration 25. The `AtomicLong` reference can be read *before* `releaseResidualDeferredRAM` removes the entry and used after it, so the guard has to be on the **value**, not the entry.
- The per-database counter is now the authority on what may be released: the take is atomic and clamped at zero (`getAndAccumulate`), and only what the counter actually still held moves off the JVM-wide total. Every interleaving of the bulk take and the straggling per-page ones now sums to the same thing.

New regression test `aPurgeRacingItsOwnResumeReleasesTheSameBytesOnlyOnce` (100 iterations of resume-racing-purge on a 400-page batch): **fails at iteration 0 on every run against the unfixed code**, at iteration 25 against the insufficient fix, and is green for 500 iterations across 5 runs with the value-based take.

On the smaller notes:

- `deferredRAMBytes` now gets the negative-value guard you asked for: a repair to zero plus a one-shot WARNING, since unlike the per-database gauge it is the number the backpressure decision is made from. One consequence worth flagging: that repair would otherwise **hide** exactly this class of bug from the tests, because a double release lands on zero after the clamp and looks like a correct run - so the repair count is exposed package-private and the race tests assert it stays zero.
- Stale `{@link #resetDeferredRAMIfDrained}` repaired (leftover from the rename in commit 2), and `awaitDeferredBacklogUnderCap`'s `instanceof Database` now says explicitly that it is a cast rather than a filter.
- Not taking the `@Tag("slow")` suggestion: measured, `Issue6200PerDatabaseDeferredBacklogTest` runs in 1.2 s and `Issue6199DrainWakeupTest` in 0.9 s, well under the "noticeably long" bar CLAUDE.md sets for that tag.

Full engine suite green: 12003 tests, 0 failures.

lvca:
Round 5's finding is valid and fixed in 3a2778d32: `removeAllOfDatabase` now zeroes the counter inside the monitor before notifying, so a waiter that resolved it before the removal finds the drained state on the object rather than parking through a signal that already fired.

Two things worth recording from writing the regression test, since both bear on how much the test is worth:

- **The obvious test did not discriminate the fix.** One waiter per round with the purge fired immediately after passed three runs in three against the unfixed code - an unaligned waiter is almost never inside a window that is a handful of instructions wide. `aPurgeRacingTheParkIsNeverMissed` releases 12 waiters *and* the purge from one gate over 80 rounds, and fails about one run in three against a purge that notifies without zeroing. It is sized no larger than it needs to be: 24 waiters over 150 rounds reproduced at the same rate for four times the thread churn, which the shared engine-suite JVM pays for.
- **The full suite caught a flake this branch introduced**, unrelated to your finding. `theResidualDrainKeepsItsHardDeadline` asserted that a wait bounded by a 200 ms deadline returned inside 15 s; in the suite it measured **24 s** and failed, with the code behaving correctly - the class runs in 1.2 s in isolation, but after ~12000 tests in one JVM a stop-the-world pause of tens of seconds is not exotic. The deadline being honoured is already asserted by the return value; the elapsed check is a tripwire against an *unbounded* wait, and its bound now says so instead of doubling as a latency assertion.

Full engine suite green: 12004 tests, 0 failures.

lvca:
Round 6's Phase-1 gap is closed in fa3fcadb7 rather than left open - it sits inside the accounting this branch reworks and the fix is three lines. Phases 2 to 4 now run in a `finally`: they are pure bookkeeping, cannot fail, and the exception still propagates after them.

The test for it is worth a note, because the obvious version was useless. Trapping a deferred batch and resuming on the calling thread **passed against the unfixed code**: Phase 1 detaches the pre-existing backlog itself, so single-threaded there is nothing left for Phase 2 to lose. It only discriminates once a batch is deferred *while* Phase 1 runs - the resume on its own thread, the defer path driven underneath it - and in that shape it fails 3 runs out of 3 without the fix.

The WARNING now reports the elapsed hold instead of the constant threshold, as suggested.

On the last point: the per-database Micrometer gauge is deliberately left to #6087 (open) rather than wired here. This PR provides the number that item needs (`PageManager.getDeferredRAMBytesOf`), which was the blocker - the number did not exist per database - and tagging the gauge belongs with the rest of that item.

Full engine suite green: 12005 tests, 0 failures.

lvca:
Round 8's "worth double-checking" is checked, and the answer is recorded in the code rather than only here: `awaitDeferredBacklogUnderCap`'s javadoc now states which locks each entry point holds when it parks, because "must not hold the page-manager lock" is only half the requirement - the other half is holding nothing the resume needs to release the backlog being waited on.

- A committer arrives through `publishPages`, **before** its `lock()`.
- Index compaction arrives through `writePages`, **after** `updatePageVersion` has taken and released that same lock - verified at both `LSMTreeIndexCompacted` call sites.
- Nothing outside `PageManagerFlushThread` can hold `suspendLock`/`replayDrainLock` (both private), the resume takes no database lock, and a parked waiter has released `deferredRAMLock` by definition.
- So a parked caller holds only its own transaction's file locks, which the resume never asks for. No cycle.

Worth noting this direction got *less* blocking, not more: before this PR a compaction thread could block unboundedly in `queue.offer` at the same point, and the old drain-side gate is precisely what filled that queue during a suspension.

On the timing note - acted on, and you were right to raise it:

- The awaits asserting a wait was **released** now allow 60 s (fallback moved to 10 min to keep the gap an order of magnitude wide). A generous bound on an expected-success wait cannot turn a passing run red; a tight one can. The short "not released yet" windows are unchanged, since a stall only strengthens them.
- Re-measuring after that change turned up something I had overclaimed: `aPurgeRacingTheParkIsNeverMissed` reproduces its bug in **1 run out of 12** on this (heavily loaded) machine, against the 2-in-6 I measured earlier on an idle one. The javadoc now states both figures, says to rely on the lower, and says plainly that the guarantee rests on the argument in `removeAllOfDatabase` rather than on this test catching every regression.

Separately, the full suite here also turned up two failures I did **not** chase into this PR, because neither is reachable from this change and both pass 3/3 in isolation: `AsyncShutdownDrainTest.closeDoesNotHangOnAWedgedAsyncWorker` (27.4 s against a 15 s bound - the same stall signature) and `DeltaScanVectorSearchTest.deltaClearedAfterRebuild`. Five earlier full-suite runs of this same branch were green on both. Flagging them so CI's verdict on a clean runner is the one that counts.

lvca:
Round 9 is right and the correction is in 893168f04. Verified before changing it: `PageManager.updatePageVersion` contains no `lock()`/`unlock()` - only `validateAndBumpVersions` wraps it, and both `LSMTreeIndexCompacted` call sites bypass that and call `updatePageVersion` directly. My round-8 sentence asserted a property of one call path as if it were a guarantee.

The javadoc now records your reason instead, which is the stronger one: `LockContext.lock()` is `protected`, and `com.arcadedb.index.lsm` neither shares the package nor subclasses `PageManager`, so index code cannot acquire that lock however the path is refactored. That is a bar the type system enforces rather than something a future reader has to re-verify.

lvca:
Round 10, all four notes:

**1. Stale `EngineMetricsBinder` comment** - fixed in 83cd92249. It said crossing the cap "throttles committing threads outright", which was true when the flush thread stopped draining for everybody. It now says *which* committers, and points at `getDeferredRAMBytesOf` as the per-database split item 2 of #6087 will tag the gauge by.

**2. Undefended "one batch, one database" invariant** - good catch, and the framing is right: a violation throws nowhere and instead charges one database for another's pages, which is the exact drift this branch spent its rounds removing. Now asserted in the `PagesToFlush` constructor - zero cost in production, loud in the test lanes, where such a batch would first be written. Deliberately an `assert` rather than an unconditional check: the latter puts an O(batch) walk on the publication path to defend against a bug that has to be introduced there first.

It comes with a test that **proves the assert fires** (`aBatchMixingTwoDatabasesIsRejectedLoudly`, self-skipping under `-da`). An assertion nobody has seen trip is worth what a test nobody has seen fail is worth - which is the same standard the rest of this PR's tests were held to.

**3. `@Tag("slow")`** - measured, since you couldn't run it: `Issue6200PerDatabaseDeferredBacklogTest` is **1.2 s** and `Issue6199DrainWakeupTest` **0.9 s**, an order of magnitude under the "noticeably long / multi-second" bar CLAUDE.md sets for the tag. I did see a single 26 s outlier while the machine was still finishing a compile; re-measured 3× it is 1.18/1.20/1.20 s. Not tagging.

**4. Splitting the `Issue5279ConcurrentUpdateTest` fix** - agreed in principle, declining in practice: it is already isolated inside a commit whose message flags it as `UNRELATED` with the reasoning, and rebasing to split it now would force-push over the commits this review thread references by hash.

lvca:
Round 11:

**The synchronous-flush question** - answered, and the answer is now in the javadoc (b410f45c9) rather than left to be re-derived. A synchronous write cannot accumulate against this cap: what the cap bounds is `deferredByDatabase`, and `flushPagesFromQueueToDisk` is the *single* place anything is ever put there (one `computeIfAbsent` in the whole class - grepped, not assumed). A synchronous `writePages` goes straight to `flushPage` without passing through the flush thread at all, so it adds nothing to the backlog and has nothing to be held for. That is exactly why the gate is on `asyncFlush`.

**The ALL-CAPS style** - declining, with evidence, because the premise is the other way round. Emphasis-caps is the established convention in these two files, not a departure from it: on `main`, untouched by this PR, `PageManagerFlushThread` has 10 such comments and `PageManager` has 64, including multi-line caps blocks like the snapshot barrier's `// THE CLOCK STARTS HERE, NOT AT THE PUBLIC ENTRY POINT...` and `// MANAGE CONCURRENT ACCESS TO THE PAGES...`. CLAUDE.md's instruction is to adhere to the existing code, and rewriting the new comments to sentence case would make them the odd ones out in their own file while adding diff noise to a change that has already been reviewed eleven times.

**`PageManagerFlushThread` accumulating responsibility** - agreed, and worth recording: it now owns the drain signalling, the per-database RAM accounting, the cap wait and the warn/backoff. #6087's gauge wiring should not add to it - it only needs to *read* `getDeferredRAMBytesOf`, which is already exposed on `PageManager`. If that item grows past a gauge read, splitting the accounting out is the natural seam.

Thanks for the eleven rounds - between them they turned up four genuine races that the first version of this PR shipped with.

lvca:
Round 12 - the one point I could act on was the perf question, so I measured it rather than reasoning about it.

**3. Per-commit overhead of `awaitDeferredBacklogUnderCap`.** Steady-state, non-suspended commit throughput, 40 000 single-document transactions per round after a 20 000-commit warm-up, current HEAD versus the identical build with the gate short-circuited at its first line:

| | round 0 | round 1 | round 2 |
|---|---|---|---|
| with the gate | 648 ms | 636 ms | 626 ms |
| gate disabled | 643 ms | 633 ms | 619 ms |

The gap is ~7 ms on 40 000 commits (~1%, ≈0.18 µs/commit) and the **within-run** spread is 22 ms, so it is not distinguishable from noise. That matches the shape of the code: on a non-suspended database the gate is a list access, a `maxDeferredRAM <= 0` test and one CHM lookup, all before any monitor. (Probe deleted, not committed.)

**1. Test scaffolding duplication** - declining, deliberately. The two classes guard different issues (#4728's cap bound versus #6200's per-database scoping) and the fixtures have already diverged for good reasons: `Issue6200PerDatabaseDeferredBacklogTest` needs 256 KB pages for the cap arithmetic *and* 1 KB pages for the per-page-loop races, which the #4728 test has no use for. Sharing a base class would couple two independent regression guards so that a future fixture change to one has to be reasoned about for the other - the drift you spotted is the argument for keeping them apart, not for merging them. The cost is ~6 lines of constants in test code.

**2 and 4** - nothing to do; agreed on both, and thanks for the note on keeping the in-code rationale.

That is every point across twelve rounds either fixed or answered, so I'll stop here unless something new lands. Between them the rounds caught four genuine races and two pre-existing test bugs that this branch would otherwise have shipped past.

Issue #4817, opened 2026-06-30
[SQL] `TRUNCATE TYPE <TypeName>;` is very slow and incomplete


danieljuhl:
We have an edge type with 70k edges. Executing `TRUNCATE TYPE <TypeName>;` (from the Studio) takes 145 seconds before it stops (seems successful), but only 20k edges was removed, and on the `Leader` the logs states

I'm concerned about the log entries, the very slow truncation but also that only a fraction was actually truncated. While the truncation was running, it was clearly that the edges was removed slowly, as a refresh in studio indicated a slow change in number of edges for the type in play.

```
Jun 30 09:41:44 java[210278]: 2026-06-30 09:41:32.322 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:546857 (error: Transaction error on commit)
Jun 30 09:41:48 java[210278]: 2026-06-30 09:41:44.799 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:555253 (error: Transaction error on commit)
Jun 30 09:41:49 java[210278]: 2026-06-30 09:41:48.064 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:559164 (error: Transaction error on commit)
Jun 30 09:41:52 java[210278]: 2026-06-30 09:41:49.322 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:559367 (error: Transaction error on commit)
Jun 30 09:41:52 java[210278]: 2026-06-30 09:41:52.256 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:563205 (error: Transaction error on commit)
Jun 30 09:41:59 java[210278]: 2026-06-30 09:41:52.716 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:563282 (error: Transaction error on commit)
Jun 30 09:41:59 java[210278]: 2026-06-30 09:41:59.917 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:569530 (error: Transaction error on commit)
Jun 30 09:42:01 java[210278]: 2026-06-30 09:41:59.980 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:569541 (error: Transaction error on commit)
Jun 30 09:42:02 java[210278]: 2026-06-30 09:42:01.632 SEVER [LocalBucket] <arcadedb-1> Error on loading record #623:571496 (error: Transaction error on commit)
[...]

Issue #4352, opened 2026-05-26
TRUNCATE TYPE leaves UNIQUE indexes in an inconsistent state; subsequent MERGE/INSERT fails until DROP TYPE is used


lvca:
## Context

Reported by a user migrating from Neo4j in the blog post [Moving a Graph RAG system from Neo4j to ArcadeDB](https://www.piyush-yadav.com/p/moving-a-graph-rag-system-from-neo4j) (May 2026, against ArcadeDB 26.4.1-SNAPSHOT).

## Problem

After `TRUNCATE TYPE <name>` on a type with a UNIQUE index, subsequent inserts via MERGE/INSERT can fail with duplicate-key errors as if the previous entries were still present, or otherwise behave as though the index had not been emptied.

The author had to fall back to `DROP TYPE` + full re-create instead.

See `engine/src/main/java/com/arcadedb/query/sql/parser/TruncateTypeStatement.java`. The May 2026 fix (`758fff656`) added batched deletes for OOM protection but did not address index-side cleanup.

## Reproduction sketch

```sql
CREATE VERTEX TYPE Worker;
CREATE PROPERTY Worker.name STRING;
CREATE INDEX ON Worker(name) UNIQUE;
INSERT INTO Worker SET name = 'John';
TRUNCATE TYPE Worker;
INSERT INTO Worker SET name = 'John';  -- expected: OK; reported: duplicate-key
```

## Expected

`TRUNCATE TYPE` should leave all indexes on that type empty and immediately usable (since by definition all records have been removed). If a fast index reset (re-create-empty) is faster than per-record removal, do that internally.

## Suggested fix

Either (a) iterate index entries for the truncated type and remove them in the same batched loop, or (b) drop+recreate each index of the type as part of the truncate operation when run without a transaction wrapper. Add a regression test that mirrors the reproduction.

PR #6225, opened 2026-08-16
fix(#6220): TRUNCATE joins the caller's transaction instead of committing through it

### LINKED ISSUE #6220: TRUNCATE TYPE inside an explicit transaction commits it: ROLLBACK does not put the records back
Found while fixing #6203 (PR #6208). The materialized-view refresher wrapped `TRUNCATE TYPE ... UNSAFE` in
`database.transaction()` and believed the pass was atomic; it was not, and a failed refresh destroyed the view. That
was fixed by removing `TRUNCATE` from the refresh path - but the statement itself behaves the same way for
everybody, and nothing tells a user so.

## `BEGIN; TRUNCATE TYPE ...; ROLLBACK;` destroys records

`TruncateTypeStatement.executeDDL` commits **the caller's** transaction from the inside, three ways:

* `schema.dropIndex(def.name)` for every index on the type, before a single record is touched;
* `db.commit(); db.begin()` inside the scan callback every `arcadedb.truncateBatchSize` (default 1000) records;
* another `db.commit(); db.begin()` before it rebuilds the indexes it dropped, so the rebuild sees empty buckets.

`SQLQueryEngine.executionDatabase()`'s Javadoc already documents that TRUNCATE "commits mid-execution", so this is
known internally - but only as a replication concern (#5492), never as a transactional-semantics one, and it is not
documented anywhere a user would look.

Measured on `main` (8b9ec35), two throwaway tests:

```java
// 1100 rows, no index on the type
database.begin();
database.command("sql", "TRUNCATE TYPE `T` UNSAFE").close();
database.rollback();
// before = 1100, after ROLLBACK = 100
```

```java
// 5 rows, one NOTUNIQUE index on the type
database.begin();
database.command("sql", "TRUNCATE TYPE `U` UNSAFE").close();
database.rollback();
// before = 5, after ROLLBACK = 0
```

So an explicit `ROLLBACK` recovers only whatever was in the last, uncommitted batch. Unindexed and under one batch,
nothing is lost (which is why this has never been noticed). Over one batch, everything but the tail is gone. With
**any** index on the type, everything is gone at any size, because the `dropIndex` and the pre-rebuild commit land
either side of the deletes.

Nothing reports this. The statement returns `operation: truncate type` and the rollback returns normally.

## Why it matters beyond the obvious

A user cannot write `BEGIN; TRUNCATE TYPE Staging UNSAFE; INSERT ...; COMMIT;` and expect the old rows back if the
insert fails - which is the single most natural use of TRUNCATE, and exactly the shape that made #6203 a data-loss
bug rather than a cosmetic one. It is also the shape a "reload this table" job takes.

## Suggested fix

Make the deletes part of the caller's transaction when there is one, rather than committing through it:

* if a transaction is already active when `TRUNCATE` runs, do **not** commit/begin inside the scan and do **not**
  drop the indexes - delete the records in the caller's transaction and let the index maintenance happen per record,
  the way `DELETE FROM` already does. The batching and the index drop/rebuild exist for throughput (#4817, #4352)
  and for HA log-entry size, and both are worth keeping for the no-transaction case, where TRUNCATE owns its own
  transaction and the batching is harmless;
* if no transaction is active, keep today's behaviour exactly.

That makes the statement mean what a user reading `BEGIN ... ROLLBACK` expects, without giving up the fast path.
The cost is the one #6203 accepted for the same reason: a truncate inside a transaction holds the pages it deletes
until commit.

A weaker alternative - refuse `TRUNCATE TYPE` when a transaction is active, with an error naming `DELETE FROM` as
the transactional equivalent - is worth considering if the above is judged too large. It is strictly worse (it
removes a capability rather than fixing it) but it is loud, which is the property today's behaviour lacks entirely.

## Related, not verified here

`SQLQueryEngine.executionDatabase()` names two other statements that commit mid-execution: `REBUILD INDEX` and
`BatchStep`. Neither is destructive the way TRUNCATE is - a half-rebuilt index is recoverable and `BatchStep` is an
explicitly requested batching - so they are probably fine, but somebody touching this should look at them.

lvca:
Fixes #6220.

## The defect

`BEGIN; TRUNCATE TYPE Staging UNSAFE; INSERT ...; ROLLBACK;` did not put the old rows back.

`TruncateTypeStatement.executeDDL` committed **the caller's** transaction from the inside, three ways:

* `schema.dropIndex(def.name)` for every index on the type - a schema change, applied immediately - before a single record was touched;
* `db.commit(); db.begin()` inside the scan callback every `arcadedb.truncateBatchSize` (default 1000) records;
* one more `db.commit(); db.begin()` before the dropped indexes were rebuilt over what were by then empty buckets.

So an explicit `ROLLBACK` recovered only whatever the last, uncommitted batch held. Unindexed and under one batch, nothing was lost, which is why this went unnoticed. Over one batch, everything but the tail was gone. With **any** index on the type, everything was gone at any size, because the drop and the pre-rebuild commit land either side of every delete. Nothing reported it: the statement returned `operation: truncate type` and the rollback returned normally. `TruncateBucketStatement` had the batching half of the same defect.

This is the same defect that made a failed materialized-view refresh destroy the view (#6203), met there by taking `TRUNCATE` out of the refresh path. This fixes the statement for everybody else.

## The fix

Who owns the transaction decides which of two paths runs.

* **A transaction is already active.** The truncate is one operation inside somebody else's unit of work, so it commits nothing and changes no schema: every record is deleted in the caller's transaction with the indexes maintained per record, exactly as `DELETE FROM` does, and the caller's `COMMIT`/`ROLLBACK` decides the outcome of the whole thing. `arcadedb.truncateBatchSize` is ignored.
* **No transaction is active.** The statement owns one, and keeps the fast path in full: drop the indexes, delete in small committed batches, rebuild the (empty) indexes.

The batching is a throughput and HA concern - one small Raft log entry per batch (#4817) - and never was an atomicity one, so nothing is lost by giving it up in the first case: the caller's own commit is one entry either way.

The result set now carries a `transactional` flag saying which path ran. The old behaviour was indistinguishable from the transactional one right up to the rollback that silently kept most of the deletes, and an answer nobody can read is half a fix.

### Beyond what the issue proposed

The issue's suggestion was the split above. Taken literally it would have *deleted* the fast path rather than preserved it, because that path was **unreachable**: deleting a record requires an active transaction, so with none active every delete failed with `Transaction not begun` (reproduced on `main`, `LocalDatabase.checkTransactionIsActive`). The batching and the index drop/rebuild could therefore only ever run on a caller's transaction - precisely where they must not.

So the statement now opens its own transaction, commits it on success and rolls it back on failure, with a `finally` guaranteeing it never leaves that transaction on the caller's thread. The rollback-on-failure is also a correctness gain in its own right: the index rebuild runs in its own transactions (`joinCurrent=false`) and therefore indexes the last batch's records as still present, so before this a failed truncate left the index populated with records the caller's commit was about to delete.

Two other additions the issue did not ask for: `TRUNCATE BUCKET` gets the same treatment, and the shared deletion loop plus the ownership rule are written once in `TruncateRecordDeleter` rather than twice.

The issue's weaker alternative - refusing `TRUNCATE TYPE` when a transaction is active - was not taken. It removes a capability rather than fixing it, and the capability is the one users most want.

### Choosing the fast path deliberately

Embedded: `database.command("sql", "TRUNCATE TYPE ...")` with no `begin()` around it. Over HTTP the request's auto-commit transaction is what makes the statement transactional, so `"autoCommit": false` in the payload selects the fast path for a bulk clear that does not need to be undoable. Anything wrapped in a transaction gets correctness instead, at the cost of one index maintenance per record and of holding the deleted pages until commit.

## Tests

`Issue6220TruncateTransactionTest` (8 tests), all confirmed failing on `main` where applicable:

* rollback puts back an unindexed type spanning many batches (was 1100 -> 100), and an indexed one (was 5 -> 0);
* a failed reload (`TRUNCATE` + `INSERT` + throw) leaves all 40 old rows and none of the new;
* `TRUNCATE` also removes records the same transaction just inserted;
* polymorphic rollback;
* commit still removes everything and the index agrees;
* with no transaction the statement owns one, leaves none open, empties and rebuilds the unique index, and reports `transactional: false`;
* `TRUNCATE BUCKET` rollback, then the no-transaction path on the same bucket.

Two existing tests pinned the old behaviour and are updated rather than deleted, because the behaviour they described has genuinely moved:

* `Issue5713HashIndexPageSizeTest#truncateTypeRecreatesAnIndexWithAnIllegalPageSize` asserts the drop/recreate carries a legal page size, so it now runs the truncate with no transaction. Inside one, the statement deletes record by record, which on that deliberately damaged HASH index fails rather than repairing it; `CHECK DATABASE FIX` is the repair path for that shape and has its own test in the same class.
* `Issue5492TruncateBatchNotReplicatedIT` wrapped the truncate in a transaction *specifically* to force the mid-statement batch commits. That is now the shape that produces none, so the test would have passed for the wrong reason. It issues the statement with no enclosing transaction instead and still exercises two batch commits before the statement's last. `ha-raft/CLAUDE.md` records the trap.

Run locally:

* full `engine` suite (`-DexcludedGroups=benchmark,vector,slow`): **11996 tests, 0 failures**;
* `Issue5492TruncateBatchNotReplicatedIT` under `-Pintegration`: **2 tests, 0 failures**;
* the `server` module could not be judged locally - an unrelated local application is holding port 2480, so its HTTP tests fail with the documented `403 / "Too many failed authentication attempts"` port-collision signature rather than anything from this change.

Docs: an accompanying PR updates the `TRUNCATE TYPE` / `TRUNCATE BUCKET` SQL reference pages; `docs/release-26.9.1.md` carries the release note.


lvca:
Thanks - three of the four are taken and pushed. The blocking one does not reproduce, and I think the premise is inverted; details below so the next round can check the reasoning rather than take my word for it.

## The correctness finding: the rebuild is not in the statement's transaction

> The rebuild, however, happens inside the statement's own transaction, and on any failure that transaction is rolled back rather than committed

It isn't. `recreateIndex()` → `TypeIndexBuilder.create()` → `LocalSchema.recordFileChanges(...)`, which is `executeInWriteLock(callback)` + `saveConfiguration()` - the schema is persisted directly - and it wraps each per-bucket build in its own `database.transaction(..., joinCurrent=false, ...)`. `db.rollback()` reaches neither half. `TypeIndexBuilder` says so itself, a few lines above the build:

> NOT ATOMIC, and deliberately so. The drop is its own committed schema change - `LocalSchema.dropIndex` wraps itself in `recordFileChanges`, and so does the build below - so the two cannot be rolled back as one.

The `db.begin()` before the rebuild is only there because the build path expects a transaction context to nest from, not because the rebuild joins it. That was doing too much work implicitly, so it is now stated at the site.

## Committing there is the defect, in the other direction

The rebuild runs in its own transactions, so it observes disk state - which still contains the last, uncommitted batch's records. It therefore indexes them as present. Committing the pending deletes would leave index entries pointing at records that are gone: the #4352 shape, surfacing as spurious duplicate-key errors on the next insert. Rolling back is what leaves the index and the buckets agreeing, and it is why the method's comment ("rebuilt over whatever records remain") is accurate rather than contradicted.

## The branch was genuinely uncovered, though

That part of the review is right, and it is the useful half: every existing failure-injection test goes through `truncateInCallerTransaction`, which touches no index at all. `Issue6220TruncateTransactionTest#aFailedTruncateOnTheFastPathStillLeavesTheTypeWithItsIndexes` now covers it - fail a delete mid-scan on the no-transaction path over a `UNIQUE`-indexed type, then assert:

- the type still has its index;
- `0 < surviving < 35`, so the agreement below cannot pass vacuously as `0 == 0`;
- `countEntries()` equals the survivor count;
- every survivor is reachable through the rebuilt index;
- a retry then truncates cleanly through the same path.

It passes against the code as written. If the rollback did reach the drop or the rebuild, it could not.

## The three taken

- **Exception masking.** Guarded in both statements: a commit that fails *is* the failure to report; a rollback that fails while a failure is already propagating is `addSuppressed` onto it; the `finally`'s last-resort rollback logs at `WARNING` rather than throwing.
- **HTTP default.** `Issue6220TruncateHttpDefaultTest` pins it - `autoCommit` unset → `transactional: true`, `"autoCommit": false` → `false`, both leaving the type empty. It builds its fixture through the database handle rather than `executeCommand()`, which swallows a failed request and returns `null` (an upstream POST failure then surfaces as a baffling "Type not found"), and it reads the status and error body *before* `readResponse()`, which throws a bare `IOException` naming only the code.
- **Nit.** The outer `catch` now says what it is for: `deleteAll` catches what the per-record callback raises and returns it as a value; the outer catch takes what the *scan* raises around the callback, which never reaches the callback. Both must end at the same place because the index rebuild runs either way.

## Test runs

- `engine`, full suite: 11998 tests, 0 failures.
- `network` 450, `integration` 131, both 0 failures.
- The two new HTTP tests, plus `AutoCommitParameterTest`: 9 tests, 0 failures.
- The full-module `server` run is red on `origin/main` too (758 tests, 6 failures, 10 errors on the baseline vs 760/1/7 here), from cross-class pollution under `forkCount=1` + `reuseForks=true` - `HttpBodySizeLimitTest` sets `SERVER_HTTP_BODY_CONTENT_MAX_SIZE=1024` and never resets it. Pre-existing and out of scope here, but worth its own issue.

Issue #6217, opened 2026-08-15
Reading a multi-page record fails with a ConcurrentModificationException when an unrelated record's chunk on the same page is written


lvca:
`LocalBucket.readMultiPageRecord` assembles a chunked record and then re-validates the **page version** of every page
the chain touched:

```java
for (final long[] pv : pageVersions) {
  final BasePage currentPage = database.getPageManager().getImmutablePage(...);
  if (currentPage != null && currentPage.getVersion() != pv[2]) {
    chainInconsistent = true;
    break;
  }
}
```

A page version changes when **any** record on that page changes, and continuation chunks of different records routinely
share a page - the allocator packs them there on purpose. So a reader is failed by writes that did not touch a single
byte of its own record. After `TX_RETRIES` attempts (default 3) it gives up with

```
Multi-page record #1:26 was modified during read after N retries. Please retry the operation
```

This is the same false conflict the disjoint-slot merge removed from the WRITE path in #5381/#6129/#6175, still
present on the read path.

## Measured

`Issue6129ChunkedSlotMergeTest`'s own fixture is enough: 40 records whose head chunks live on page 0 and whose
continuation chunks all land on page 1 (verified by instrumenting the allocator - every one of the 40 placements is
`headPage=0 avoid=0 -> page=1`), then 8 threads rewriting their own records, 320 transactions.

With the read budget lowered to zero (`database.getConfiguration().setValue(TX_RETRIES, 0)`), the run produces
exactly this conflict and nothing else:

```
SCRATCH-READ conflicts=1 ofWhichReadSide=1
SCRATCH-READ-MSG Multi-page record #1:26 was modified during read after 0 retries. Please retry the operation
```

1 in 2 repetitions, and never a commit-side conflict: with the default budget of 3 the same race needs four
consecutive losses, which is what a loaded machine produces every few runs.

## Why it matters beyond the test

- A read of a multi-page record can fail on a busy bucket **although the record is untouched**, and the failure
  reaches the application as a `ConcurrentModificationException` - retryable, but a retry storm on a hot page is the
  cost, and the more records share a continuation page the likelier it is.
- The commit side already answers this exact question precisely rather than by page version: `#6129` introduced
  `offPageContentFingerprint`, a fold of the chain past the head chunk. A read could validate what it actually
  consumed - the chunk headers and content it followed - instead of the version of every page they happen to live on.

## Where it surfaced

[...]

Issue #6198, opened 2026-08-15
TimeSeriesTagDictionaryTest.anIdWrittenByAnotherInstanceIsResolvedByReloading fails in a full-suite run, passes in isolation


lvca:
Observed while verifying PR #6197 (which touches only `LocalBucket` / the disjoint-slot merge, nothing this test
reaches). Filing so it is tracked rather than lost, per the review of that PR.

## What happens

In a full `engine` run - `mvn -pl engine test -DexcludedGroups=slow,benchmark,vector` over `engine/**`,
`database/**`, `graph/**`, 1467 tests - exactly one fails:

```
TimeSeriesTagDictionaryTest.anIdWrittenByAnotherInstanceIsResolvedByReloading:177
expected: "host_b"
 but was: null
```

It passes when the class is run alone, and when the whole `engine/timeseries` package is run (344 tests). So it is
test-order or timing dependent, not a deterministic failure.

## Where the null comes from

The test builds a second `TimeSeriesTagDictionary` over the file a first one wrote, and asserts the lookup
self-heals by reloading. `resolveMiss` declines to reload in two cases, and the second one is reachable here:

```java
if (id > peekStoredEntryCount())
  return null;
```

`peekStoredEntryCount()` opens with `if (getTotalPages() == 0) return entryCount;` - and the follower instance's
`entryCount` is 0, because nothing has been loaded into it yet. So a follower that sees **zero pages** answers
`0`, every id fails `id > 0`, and the lookup returns `null` instead of reloading. The self-heal is disabled
exactly in the state it exists for.

Whether `getTotalPages()` can legitimately read 0 here - a page count the writer's transaction has not published
yet, a component registered but not yet sized, a counter left by the preceding test in the same JVM - is the
question to answer first; the assertion is only the symptom.

## Worth considering

If the zero-page read is legitimate, the guard is measuring the wrong thing: "this instance has loaded nothing"
and "the file holds nothing" are different statements, and only the second one justifies declining to reload. A
follower whose page count reads 0 while a positive id is being asked for is precisely the stale-map case, so the
safe answer there is to reload once and re-check, not to return `null`.

Note the same fallback is on the other branch of `peekStoredEntryCount` ("a header that cannot be read degrades to
the in-RAM count"), which has the same shape and would want the same answer.

PR #6262, opened 2026-08-16
fix(#6258): a chunked read stops when retrying cannot help, and a corrupted chain says so

### LINKED ISSUE #6258: Follow-ups from #6217: a retry that cannot succeed, corruption reported as contention, and two CI flakes
Four independent findings from the work on #6217 (merged as `2fcf408ec`, PR #6224). Filed together because they were found together; each is separable and can be taken on its own. Items 1 and 2 are in `LocalBucket.loadMultiPageRecord`, items 3 and 4 are test/CI issues found on that PR's run.

Line numbers below are against `2fcf408ec`.

---

## 1. Under `REPEATABLE_READ`, a conflicting chunked read burns its whole retry budget on retries that cannot possibly succeed

The retry branch re-fetches the head page straight from the `PageManager`, deliberately bypassing the transaction so that the retry sees fresh data (`LocalBucket.java:3322`):

```java
firstPage = database.getPageManager().getImmutablePage(firstPageId, pageSize, false, true);
```

The chain walk, however, takes its continuation pages from the transaction (`LocalBucket.java:3269`):

```java
final BasePage nextPage = database.getTransaction()
        .getPage(new PageId(database, file.getFileId(), chunkPageId), pageSize);
```

Under `READ_COMMITTED` those two agree - the transaction caches nothing, both are the newest committed page - so a retry can genuinely succeed, which is what it is there for.

Under `REPEATABLE_READ` they cannot agree. `TransactionContext.getPage` caches every page walked in `immutablePages`, so each retry pairs a **fresh head chunk** with **stale cached tails** - precisely the torn mix `validateChainRead` then rejects. The cache is never refreshed, so every subsequent attempt reproduces the same mismatch. The read therefore always performs `TX_RETRIES + 1` complete chain walks (4 by default) before raising the `ConcurrentModificationException` it was always going to raise.

Two things to decide:

- **It should fail fast.** Retrying cannot help at this isolation level, so the budget is pure waste on a path that is already the slow one - and it is worst exactly where it hurts, on large records under contention.
- **Arguably it should not fail at all.** The transaction's cached chain is internally consistent: it *is* one committed state, the one this transaction first saw, which is what `REPEATABLE_READ` promises. Serving that snapshot would be more correct than refusing it. That is a semantic change rather than a patch, so it wants a decision first.

`Issue6217ChunkedReadFalseConflictTest.aChunkedReadStillFailsWhenTheRecordItselfMovedUnderIt` pins the current behaviour and is the place to observe it.

---

## 2. A structurally broken chain is retried as if it were contention, and reported as contention

`loadMultiPageRecord` collapses two conditions with opposite answers into one `chainInconsistent` flag:

- **the chain does not parse** - a chunk pointer past the end of the file (`:3264`), a slot that no longer holds a record (`:3273`), a marker that is not `NEXT_CHUNK` (`:3289`), or any exception from the walk (`:3295`). This is a corrupt record: retrying cannot change the answer.
- **the record changed under the read** (`:3307`, from `validateChainRead`). This is contention: retrying is the entire point.

Both take the same retry path and end at the same message:

```
Multi-page record #1:26 was modified during read after 3 retries. Please retry the operation
```

So a corrupt record costs 4 full chain walks on **every** read of it, and then reports a concurrency problem to whoever reads the log - who goes looking for contention that does not exist. `BucketIteratorBrokenChunkChainTest` shows the misleading message today:

```
SEVER [BucketIterator] Error on loading record #1:0 (error: Multi-page record #1:0 was modified during read after 3 retries. Please retry the operation)
```

The fix is cheap because the walk already knows which of the two it hit - the structural branches are distinct from the `validateChainRead` verdict. Failing fast on a broken chain, with a message that says the chain is broken, would also make `chunkChainReadRetries` (added in #6217) mean one thing instead of two.

---

## 3. Flaky: `TimeSeriesTagDictionaryTest.aSecondWaveOfIdsIsAlsoResolvedByReloading`

Failed on the `unit-tests` lane of PR #6224 ([job](https://github.com/ArcadeData/arcadedb/actions/runs/31928618893/job/95120650394)), passed on a rerun of the same commit:

```
TimeSeriesTagDictionaryTest.aSecondWaveOfIdsIsAlsoResolvedByReloading:198
expected: "host_b"
 but was: null
```

Line 198 is the **first** wave (`follower.getById(2)`), i.e. the reload that the single-wave test already covers, not the second-wave path the test was written for.

Worth treating as possibly a real race in the dictionary's self-healing reload rather than only test timing: a `null` there is exactly the failure the self-heal exists to prevent, and an HA follower hitting it would read `null` for a tag that is on disk. It did not reproduce locally - 9 consecutive runs of the class green, plus green inside a full 11993-test engine lane - so it needs the CI environment (or an equivalent load) to reproduce.

---

## 4. Flaky: `LSMVectorIndexAutoCompactionTest.anUpdateHeavyIndexCompactsItselfWithoutAnExplicitCommand`

Failed on the `slow-unit-tests` lane of the same run ([job](https://github.com/ArcadeData/arcadedb/actions/runs/31928618893/job/95120650426)), passed on rerun:

```
LSMVectorIndexAutoCompactionTest.anUpdateHeavyIndexCompactsItselfWithoutAnExplicitCommand:81
[the workload must first grow the file well past the live set, or nothing is tested]
```

The assertion that fails is the test's own **precondition**, not its subject: the workload did not grow the file as far as the test needs before it starts measuring. A test that cannot establish its precondition under load is a test that silently stops testing whenever the machine is busy, so the workload should be made deterministic rather than the threshold relaxed.

lvca:
Closes #6258. All four items, each independent of the others.

## 1. A retry that cannot succeed is no longer attempted

`LocalBucket.loadMultiPageRecord`'s retry re-fetches the head page straight from the `PageManager` (deliberately bypassing the transaction, so the retry sees fresh data) while the walk takes its continuation pages from `TransactionContext.getPage`. Under `READ_COMMITTED` those agree, so a retry can genuinely succeed. Under `REPEATABLE_READ` they cannot: the transaction pins every page it walked, so each attempt pairs a **fresh head chunk** with the **same pinned tails**, rebuilds the same torn mix, and is rejected for the same reason. The read always performed `TX_RETRIES + 1` complete chain walks before raising the exception it was always going to raise.

The new `aRetryWouldReadSomethingElse()` answers the question from the **trace, not from the isolation level** - because the isolation level is not what decides it. A retry is worth making when

- the head page has **moved** (the next walk starts from different bytes, and may follow the chain somewhere else entirely), or
- any continuation page is one this transaction has **not pinned** (it will be re-read, and can differ).

When neither holds, the next walk reads byte for byte what this one read, and the read fails now. `READ_COMMITTED` pins nothing, so it takes the first branch immediately and its behaviour is unchanged.

### On the issue's second bullet ("arguably it should not fail at all")

I looked at serving the transaction's cached chain instead, and it does **not** hold up. ArcadeDB's `REPEATABLE_READ` is a first-touch page cache, not a snapshot: `immutablePages` fills lazily, page by page, through unrelated reads. Two pages of one chain can therefore enter the cache either side of a commit that rewrote the record, so "the whole chain came from the transaction" does **not** mean "the chain is one committed state" - it can be an image that never existed, and serving it would return a torn record instead of refusing one. Making that case correct needs a real snapshot (MVCC), which is a different piece of work. So the read still fails; it just stops paying for the privilege.

## 2. A broken chain is reported as corruption, not as contention

The walk collapsed two conditions with opposite answers into one `chainInconsistent` flag: *the chain does not parse* (retrying cannot change the answer) and *the record changed under the read* (retrying is the entire point). Both took the retry path and ended at `"... was modified during read after 3 retries"`.

The walk now records **which** it hit, and a break is confirmed twice before a record is condemned:

1. does the chain still fail to walk against the **newest committed image** of its pages? A commit publishes its pages one at a time and readers take no lock, so a chain caught mid-publication can look broken while being sound at both ends of that window.
2. are the chunks this read consumed still, byte for byte, the ones the newest committed image holds? This is the guard the pre-existing `isChunkChainBroken` probe never had - it is what rules out the half-published commit, whose new pages would have moved the record under this read.

Only when both agree does the read stop and raise the new **`BrokenChunkChainException`**, naming the break and pointing at `CHECK DATABASE FIX`.

It is **not** a `NeedRetryException`, deliberately: telling the layers above to re-run the transaction for a corrupted record multiplies the wasted work instead of surfacing the repair. That is also the simplification the type buys - every caller that had to walk the chain a second time to tell corruption from contention (`BucketIterator`, `LocalDatabase.deleteRecordNoLock` x3, `GraphEngine.deleteEdgesOf` / `deleteEdgeOfDeletedVertex`, `EdgeLinkedList.anchorChain`) can now ask the exception. Their `ConcurrentModificationException` arms are untouched, because a CME still arrives from the delete path and elsewhere.

Two smaller things fall out of the split:

- the **infinite-loop guard** used to `throw` inside the walk's own `try`, where the catch turned it straight back into "the chain is inconsistent" - a positively identified loop, retried like contention and reported as contention, with the word *loop* nowhere in the message. It is now one of the breaks.
- `chunkChainReadRetries` (added in #6217) means **one** thing instead of two: contention. Its javadoc says so.

**Behaviour change worth flagging:** a `DELETE` over a record with a structurally broken chain and `DELETE_TOLERATE_BROKEN_CHAIN` off now fails with `BrokenChunkChainException` rather than `ConcurrentModificationException`. Same refusal, precise diagnosis, and no retry machinery re-running the transaction for it. `DeleteToleratesBrokenChainToggleTest` is updated to assert exactly that.

## 3. `TimeSeriesTagDictionaryTest` flake: SUPERSEDED, this branch defers to #6198

> **Update after merging main.** `8fcd89790` (#6198, PR #6261) landed the same diagnosis and a better fix, so this
> branch now takes main's version of `TimeSeriesTagDictionary` and its test wholesale and carries **no item-3 change
> of its own**. #6261 goes further than this branch did: it checks the header MAGIC (telling an initialised header
> from a page that is not there at all), threads the header through `load(StoredHeader)` so `resolveMiss` reads it
> once, throws rather than reading a failed header as "no header", and fixes `openOrCreate` instead of merely
> documenting it as safe - which is the destructive branch the first round of review here asked about. The original
> write-up is kept below for the record.

<details><summary>Original item 3 write-up (now delivered by #6198)</summary>

The mechanism, reproduced and proven: a dictionary page is committed **into the page cache** immediately and written to the file whenever the asynchronous flush thread gets round to it. `TimeSeriesTagDictionary.peekStoredEntryCount()` (and `load()`) asked `getTotalPages()`, which is seeded from the **physical size of the file** when the component is constructed. An instance opening in that window counts zero pages, so the self-heal that exists precisely to resolve an id it never interned declined to run and returned `null` for a value sitting in the cache. On CI the window is wide enough to hit; on an HA follower, where every id arrives from the leader, it is the whole failure mode. Same lesson as #5976, which took the physical file size out of the "does this page exist" decision on the transaction read path.

The header page is now read straight from the `PageManager`, which is also the right source for a *staleness* test (a caller's `REPEATABLE_READ` snapshot would answer with the count this instance already knows about) and drops a `begin`/`rollback` pair per miss.

The new `anIdIsResolvedWhileItsPageIsStillOnlyInTheCache` makes it deterministic by holding the flush thread for the whole fixture, and **asserts** the preconditions (file still empty, page count still zero) rather than assuming them. Verified non-vacuous: with the old guard restored it returns `null`, which is the exact CI failure.

</details>

## 4. `LSMVectorIndexAutoCompactionTest` flake: the workload, not the threshold

Growing the file and waiting for the trigger were the same loop, and compaction runs on the async executor - so a cycle whose commit scheduled one could see it finish before the loop next sampled `getTotalPages()`. The loop exited with the peak unmeasured and the test failed its own **precondition**, which says nothing about the subject and everything about how busy the machine was.

The garbage is now built with the scheduler off, where growth is deterministic and the peak can be measured exactly, and only then is the scheduler let back in for one more cycle of the very same workload. That is still an index compacting itself with nobody typing `COMPACT INDEX`. The threshold is untouched, per the issue.

## Tests

New `Issue6258ChunkedReadRetryAndCorruptionTest` (5 tests, on the #6217 fixture, now shared through `BucketPageLayoutTestSupport`):

- the futile retry stops after one attempt **with `TX_RETRIES` raised to 10**, so the assertion measures the short-circuit and not the budget - it is `chunkChainReadRetries == 2`, whatever the budget is
- a retry that still has something to read is still made
- a chunk pointer into nowhere, and a chain that loops, each raise `BrokenChunkChainException` on the **first** attempt with `chunkChainReadRetries == 0`, and are asserted *not* to be a `NeedRetryException`
- a record that really moved under the read is still contention: `ConcurrentModificationException`, and asserted *not* to be a `BrokenChunkChainException`

**Runs (before the merge of main):** full `engine` unit lane green - 12064 tests, 0 failures, 0 errors - plus the `slow`-tagged `LSMVectorIndexAutoCompactionTest` and the targeted set around the delete/scan/corruption paths.

## Files

- `engine/src/main/java/com/arcadedb/exception/BrokenChunkChainException.java` (new)
- `engine/src/main/java/com/arcadedb/engine/LocalBucket.java`
- `engine/src/main/java/com/arcadedb/engine/BucketIterator.java`
- `engine/src/main/java/com/arcadedb/engine/PageManager.java`
- `engine/src/main/java/com/arcadedb/database/LocalDatabase.java`
- `engine/src/main/java/com/arcadedb/graph/GraphEngine.java`
- `engine/src/main/java/com/arcadedb/graph/EdgeLinkedList.java`
- `engine/src/test/java/com/arcadedb/database/Issue6258ChunkedReadRetryAndCorruptionTest.java` (new)
- `engine/src/test/java/com/arcadedb/database/BucketPageLayoutTestSupport.java`
- `engine/src/test/java/com/arcadedb/database/Issue6217ChunkedReadFalseConflictTest.java`
- `engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexAutoCompactionTest.java`
- `engine/src/test/java/com/arcadedb/DeleteToleratesBrokenChainToggleTest.java`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,19 @@ General design principles:
   - The tags are a partition, not a set of overlapping labels: CI selects each lane with `-Dgroups`/`-DexcludedGroups` such that every test runs in exactly one lane. Tagging a class `vector` and one of its methods `slow` is fine, but do not expect that method in the slow lane
   - Required imports: `import org.junit.jupiter.api.Tag;`
   - A JUnit tag on a Cucumber `@Suite` does not work: Surefire's `groups`/`excludedGroups` reach the scenarios inside the suite rather than the suite class. Route a suite to a lane with `-Dsurefire.includes` instead, as `.github/workflows/mvn-test.yml` does for the openCypher TCK
+- Never assert on raw wall-clock elapsed time. A full-suite run shares one JVM, and a stop-the-world pause of tens of
+  seconds late in a 12,000-test run turns any bound with less headroom than that into a coin flip on the JVM's mood
+  (#6260). Measure with `com.arcadedb.utility.StallAwareStopwatch` (engine test-jar) instead: it discounts the JVM-wide
+  stall observed inside the measured window, so the bound can stay tight AND stop flaking. Pick the assertion that says
+  what the number is for, because the message is what stops the next person from "fixing" a red run by loosening it:
+  - `assertGaveUpWithin(bound, whatItSeparates)` when the bound is a tripwire between a bounded operation and an
+    unbounded one. Generous is free here: a wider bound cannot turn a passing run red
+  - `assertStayedUnder(bound, claim)` when the bound IS the assertion - a complexity claim with no other practical
+    expression (a regex that must not backtrack exponentially, one deadline shared across rows rather than charged per
+    row). Loosening it deletes the test
+  - a short wait expected to TIME OUT needs neither, and neither does a lower bound (`isGreaterThan`): a stall only
+    makes those more true
+  - `@Timeout` is plain wall clock and cannot be discounted, so size it as a hang detector, not as a latency bound
 - don't add Claude as author of any source code
 
 ## Build and Development Commands
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -1911,6 +1911,74 @@ Two things the same method carried:
 
 [#6202](https://github.com/ArcadeData/arcadedb/issues/6202)
 
+## The cluster-verify endpoint reports a peer it could not contact as unverified, not as agreeing (#6221)
+
+`POST /api/v1/cluster/verify/{database}` exists to tell an operator whether the cluster's copies of a database
+agree, and it resolved each peer's HTTP endpoint through the best-effort resolver. With no `http` port declared in
+`arcadedb.ha.serverList` that endpoint is derived from the peer's Raft host plus **this** node's HTTP port, so on a
+cluster whose nodes share a host and differ by port - several nodes on one machine, a compose file, a developer
+laptop - every peer collapses onto the address of the node doing the resolving. Two things then went wrong at
+once, and both pointed the reassuring way:
+
+- the leader compared itself against itself, matched on every file, and recorded the **peer** as `CONSISTENT`,
+  rolling up to `overallStatus: ALL_CONSISTENT` - a clean bill of health from the divergence detector for a node
+  that was never contacted, produced at the moment an operator is asking precisely because they suspect
+  divergence;
+- the query landed back on the leader, where `isLeader()` is still true, so it fanned out again: (N-1) in-flight
+  requests per level, each level CRC-ing every byte of every file before recursing, each hop holding an Undertow
+  worker thread. The peer connect/read timeouts bound one hop, not the depth.
+
+Every peer is now dialled through one helper, `PeerDialAddress`, which answers "may this node dial that one, and
+at which address?" next to the resolution rather than at each call site: it refuses an absent peer, this node
+itself, an address that identifies no single peer (#6202) and an address that is this node's own, including the
+loopback spellings of #6204. The six snapshot-resync paths of #6202 now delegate to it as well, so the endpoint
+that checks the cluster and the paths that repair it cannot drift apart. The fan-out also carries the one-hop
+marker the write-forward paths have set since #6191, so a node serving a query a peer fanned out to it answers
+with its own checksums instead of fanning out again - the bound no local self-check can provide, for an address
+that names the wrong *peer* rather than this node.
+
+**Observable change for HA alerting.** A peer that could not be verified is now reported as
+`status: ERROR` with the reason, and `overallStatus` has a third value, `VERIFICATION_INCOMPLETE`:
+
+| outcome | before | now |
+|---|---|---|
+| every peer compared and agreed | `ALL_CONSISTENT` | `ALL_CONSISTENT` |
+| a compared peer's checksums differ | `INCONSISTENCY_DETECTED` | `INCONSISTENCY_DETECTED` |
+| a peer could not be verified (unreachable, timed out, or no address identifies it) | `INCONSISTENCY_DETECTED` | `VERIFICATION_INCOMPLETE` |
+
+The guard covers the encrypted endpoint too, which is not the formality it looks like: a peer's HTTPS address is
+read from the 5th field of `arcadedb.ha.serverList` where its HTTP address is read from the 3rd, each with its own
+derive fallback onto **this** node's port for that protocol. A cluster that declares distinct `http` ports and
+omits the `https` ones therefore passes the HTTP check with every peer's HTTPS endpoint still collapsed onto this
+node's own. `getUnambiguousPeerHttpsAddress` and `getLocalHttpsAddress` ask the HTTPS endpoint the same two
+questions; withheld, it reads to the caller exactly like an absent one, so the dial falls back to the guarded
+plain-HTTP endpoint rather than being refused. The snapshot-download paths of #6202 take the guarded HTTPS address
+now as well, and so does the leader's automatic resync of a stalled replica (#4728), whose payload - "drop your
+copy of every database and download it again" - is the most destructive of the family.
+
+`ALL_CONSISTENT` is no longer reachable while any peer is unverified - an unverified peer is not a consistent one
+- and a node being down is no longer reported as a divergence somebody has to go and find. Alerting that keys on
+`overallStatus != "ALL_CONSISTENT"` is unaffected; alerting that keys on `INCONSISTENCY_DETECTED` specifically
+should add `VERIFICATION_INCOMPLETE` to keep catching "a node is unreachable". Studio shows the new status as a
+warning and renders each peer's error alongside its status.
+
+Also in the same change: `retries` on an auto-committed request (`POST /api/v1/command`, `/api/v1/batch`, and
+every other handler that auto-commits) defaulted to a hard-coded 1, so nothing retried unless the client asked
+for retries it had no reason to know existed. That was dead code until #6201 made the wrapper stop flattening
+exceptions, and live afterwards - an MVCC conflict a second attempt would have committed was answered 503 on the
+first. It now follows `arcadedb.txRetries`, the attempt count the embedded `Database.transaction(...)` and
+`RemoteDatabase` already take, and which an operator can turn per database. The engine still retries only the two
+conflict families, caps a duplicated key at one retry, and rolls the transaction back before every attempt.
+
+**Observable change for HTTP clients.** An auto-committed request that hits an MVCC conflict is now re-executed
+up to `arcadedb.txRetries` times (default 3) instead of failing on the first attempt, so a write that used to be
+answered 503 under contention now usually succeeds. Nothing the failed attempt wrote survives - the engine rolls
+back before every retry - but the *command itself* runs again, so a SQL/JavaScript function with a side effect
+outside the database (an HTTP call, a file write, a counter in an external system) can now perform that side
+effect more than once per request. Set `retries` to 1 in the request payload, or lower `arcadedb.txRetries`, for
+commands where that matters.
+
+[#6221](https://github.com/ArcadeData/arcadedb/issues/6221)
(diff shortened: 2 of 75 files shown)


lvca:
Thanks - good catch to ask, and you were right that it needed a check rather than an assumption. I traced it: **`openOrCreate`'s branch is out of item 3's reach**, and the reasoning is now a comment at the call site (52df0f134) rather than only in this thread.

Three things have to hold at once for the destructive outcome you describe, and they cannot:

1. **Getting past the lookup means nothing in this session has written to this dictionary.** The branch is only reached when `schema.getFileByName(name)` misses, i.e. no component under that name is registered - so no commit of this session has put a page of it into the cache. The lag item 3 closes needs a committed page; there is none.
2. **A non-empty page count can only have come from a previous session.** The freshly built component seeds `pageCount` from the file `FileManager.getOrCreateFile(name, ...)` hands back, keyed by component name, and that file is non-empty only when `scanDirectoryForComponentFiles` registered one when the database opened - bytes written by a previous session and flushed before it closed. That is precisely the case where the physical size *is* authoritative.
3. **A dictionary created during this session never comes back here.** It is registered on the line above, so every later call returns at the lookup, and its file (created empty by `getOrCreateFile`) reports 0 pages - the right answer for a new dictionary.

Worth adding: the cache probe could not stand in at that call site even if one wanted it to. The component there carries a **freshly allocated** file id while the file it adopts is registered under the id encoded in its name, so `peekHeaderPage()` would look for page 0 of an id the FileManager has never seen. The adopt branch is for a cold open whose schema lost the reference; the ordinary cold open rebuilds the component through `PaginatedComponentFactoryHandler` and returns at the lookup, never reaching it.

So: out of scope for this PR, and now documented as such where the next reader will hit the question.

lvca:
Merged `main` (`8c2a17213`) - conflicts resolved, and worth a note because of *how*.

Main had gained `8fcd89790` (**#6198**, PR #6261), which is the same diagnosis as item 3 of this PR - the tag dictionary's self-heal keyed on a page counter that lags the flush - with a better fix. Both conflicting files (`TimeSeriesTagDictionary` and its test) therefore take **main's version wholesale**, and this branch now carries **no item-3 change at all**.

#6261 goes further than this branch did on every axis:

- `readStoredHeader()` checks the header **MAGIC**, so it distinguishes an initialised header from a page that is simply not there; `peekHeaderPage()` here only asked whether page 0 existed.
- The header is threaded through `load(StoredHeader)`, so `resolveMiss` reads it **once** instead of twice, and the rebuild is tied to the header the decision was made against.
- A failure to read a header that **is** there is thrown rather than read as "no header", so a caller that cannot cope with an empty dictionary fails loudly instead of publishing an empty mapping over the real one.
- **`openOrCreate` is fixed, not merely documented as safe.** That is the branch the first round of review here asked about, and the destructive one - the wrong answer resets the id space over a populated file. #6261 adopts on the header and drives the branch in a test that removes the component from the schema, which is a better answer than the reachability argument I left in a comment.

Nothing else on this branch touched those two files, so items 1, 2 and 4 are untouched by the resolution. Verified on the merged tree: `engine` compiles, and the 42 tests across `Issue6258ChunkedReadRetryAndCorruptionTest`, `Issue6217ChunkedReadFalseConflictTest`, `TimeSeriesTagDictionaryTest` (main's), `BrokenMultiPageRecordDeleteTest`, `DeleteToleratesBrokenChainToggleTest`, `BucketIteratorBrokenChunkChainTest` and `LSMVectorIndexAutoCompactionTest` all pass.

The PR description is updated to say the same, with the original item-3 write-up folded away for the record.

PR #6268, opened 2026-08-16
fix(#6260): a wall-clock bound measures the code, not the JVM's stop-the-world pauses

### LINKED ISSUE #6260: Wall-clock assertions in the engine suite go red on the shared JVM's stop-the-world pauses, not on the behaviour they test
Observed twice while running the full `engine` suite for #6223, both times with the code behaving correctly.

## The evidence

- `AsyncShutdownDrainTest.closeDoesNotHangOnAWedgedAsyncWorker` asserts `elapsed < 15_000` for a `close()` bounded by a ~1 s `ASYNC_CLOSE_TIMEOUT` plus a 2 s join timeout. In a full-suite run it measured **27 396 ms** and failed. It passes 3/3 in isolation.
- `Issue6199DrainWakeupTest.theResidualDrainKeepsItsHardDeadline` (added by #6223) asserted `elapsed < 15_000` for a wait bounded by a **200 ms** deadline. It measured **24 143 ms** in the same suite and failed; the class runs in 1.2 s in isolation. That one is already fixed in #6223, by making the bound a tripwire against an *unbounded* wait instead of a latency measurement.

Both are the same failure: late in a 12 000-test single-JVM run, a stop-the-world pause of tens of seconds is not exotic, and any assertion whose headroom is smaller than that becomes a coin flip.

## The scope

`grep` for elapsed-time assertions in `engine/src/test/java` turns up a dozen-plus, several far tighter than the two above:

```
Issue5587TxRetryExponentialBackoffTest:107     isLessThan(1_000)
AsyncWaitCompletionTimeoutTest:64             isLessThan(5_000)
AsyncShutdownEscalationTest:96                isLessThan(15_000)
TextRegexReplaceTest:67, :96                  isLessThan(5000), isLessThan(1000)
DocumentValidationTest:741, :769              isLessThan(5000), isLessThan(1000)
FullTextQueryExecutorTest:548, :583           isLessThan(5000)
LSMVectorIndexRebuildTest:449                 isLessThan(5000)
OpenCypherWhereClauseTest:236, :264           isLessThan(5000), isLessThan(1000)
```

These are not all equivalent - some are guarding a real complexity claim (a regex that used to backtrack exponentially, a search that must not block on a rebuild) where a wall-clock bound is the only practical expression of the property. But several are bounding an operation whose *intended* limit is a named timeout, and there the bound should be derived from that timeout rather than picked as a round number.

## Proposed

Not a blanket loosening. Per assertion, ask what the number discriminates:

1. **Bounding a named timeout** (the two failures above): tie the bound to what it separates. `closeDoesNotHangOnAWedgedAsyncWorker` is really asserting "close gave up rather than waiting out the 60 s wedged worker", so the discriminating bound is well under 60 s, not 15 s - the tight value buys nothing and costs a flake.
2. **Guarding a complexity claim** (the regex and validation ones): the bound is the assertion, but it should have an order of magnitude of headroom over the measured honest time, and the failure message should say which of the two it is so the next person does not "fix" it by loosening.
3. **Asserting something has NOT happened yet** (a short `await(...)` expected to time out): leave alone. A scheduling stall only makes those more true.

The rule that fell out of #6223, worth applying generally: a bound on a wait that is expected to SUCCEED can be generous at no cost, because a longer bound cannot turn a passing run red; only a bound on a wait expected to FAIL has to be tight.

## Verification

Re-run the full `engine` suite (`mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow`) a few times on a loaded machine; the two named above are the ones observed to fail, so they are the ones to fix first and the ones whose fix can be checked.

lvca:
Fixes #6260.

## The problem

Two engine tests went red in a full-suite run **while the code behaved correctly**:

| test | bounded by | measured |
|---|---|---|
| `AsyncShutdownDrainTest.closeDoesNotHangOnAWedgedAsyncWorker` | a ~1 s `ASYNC_CLOSE_TIMEOUT` plus a 2 s join | **27,396 ms** |
| `Issue6199DrainWakeupTest.theResidualDrainKeepsItsHardDeadline` | a 200 ms deadline | **24,143 ms** |

Both pass in isolation. Late in a 12,000-test single-JVM run a stop-the-world pause of tens of seconds is not exotic, and any assertion whose headroom is smaller than that becomes a coin flip on the JVM's mood rather than a statement about the code.

## Why the issue's own proposal is not enough, and what this does instead

The issue proposes reclassifying each bound and widening the ones that are only tripwires. That is right as far as it goes, but it cannot cover the set. `grep` turns up 40 elapsed-time assertions in the engine suite, and a large family of them reads:

```java
// 10 independent 200ms-per-row budgets would take >= 2000ms; a shared deadline keeps the whole
// evaluation close to the single configured 200ms bound instead.
assertThat(elapsedMillis).isLessThan(1000);
```

That 1000 **cannot** be widened. It is discriminating one shared 200 ms deadline from ten independent per-row ones, so the moment it goes past 2000 the assertion stops asserting anything. Widening it is not a repair, it is a deletion. And the same stall that broke the two named tests breaks these too - it is simply luckier so far.

So instead of adjusting the numbers, this fixes the **measurement**. A daemon sampler parks for 10 ms at a time and accumulates, into a monotonic counter, however much longer than that each park actually took. The difference of two readings is the JVM-wide stall observed inside that interval; subtracting it from a wall-clock measurement over the same interval leaves the time the JVM was actually running. Every bound stays exactly as tight as it was, and none of them can be tripped by a pause.

```java
final StallAwareStopwatch stopwatch = StallAwareStopwatch.start();
assertThatThrownBy(...).isInstanceOf(TimeoutException.class);
stopwatch.assertStayedUnder(1000, "one 200ms budget shared by the whole evaluation, not one per row");
```

What is deliberately **not** discounted: time the measured thread spends sleeping, parked or blocked on I/O. The sampler keeps ticking then, so a test that waits a second is still charged a second - only a stall that also stops the sampler is taken out. Otherwise every latch wait in the suite would silently become unbounded. The sampler also runs at the default thread priority on purpose: it is a proxy for the measured thread, so it has to lose the same races for a core. Raising it would let it tick straight through the CPU starvation the measured thread is suffering, and starvation is the half of the problem a GC-pause-only view cannot see.

## The two assertions, and why there are two

They differ only in the failure message, and that is the point - the message is what stops the next person from "fixing" a red run by loosening the number:

- **`assertGaveUpWithin(bound, whatItSeparates)`** - the bound is a tripwire between a bounded operation and an unbounded one. Generous is free here; a wider bound cannot turn a passing run red. Failure says so explicitly.
- **`assertStayedUnder(bound, claim)`** - the bound IS the assertion, standing in for a complexity claim with no other practical expression (a regex that must not backtrack exponentially, a deadline shared across rows rather than charged per row). Failure says loosening it deletes the test.

A short wait expected to TIME OUT needs neither, and neither does a lower bound: a stall only makes those more true.

## Scope

40 assertions across 26 test classes converted. Per the taxonomy above:

- **Left alone** - the `isGreaterThanOrEqualTo` sites in `SQLScriptAdditionalCoverageTest`, `CodeUtilsTest`, `PaginatedSparseVectorEngineBackpressureTest`, `Issue5693TxRetryDelayScopeTest`.
- **Split** - the two `isBetween` sites (`FlushRobustnessTest`, `Issue6125SnapshotBarrierAndCapTest`) now read raw wall clock for the floor (which proves the wait really blocked) and the discounted time for the ceiling.
- **Sampled per gap** - `Issue5587TxRetryExponentialBackoffTest` compares the sum of early retry gaps against the sum of late ones. A pause landing in either half skews the ratio, so it now samples the stall counter alongside each timestamp and differences both.
- **`@Timeout` raised** - three annotations at 30 s on tests whose own budget is already 16 s or more. JUnit's timeout is plain wall clock and cannot be discounted, so it has to be sized as a hang detector, not as a latency bound.

`CLAUDE.md` gains the rule so new tests do not reintroduce the pattern.

## Verification

- `StallAwareStopwatchTest` reproduces the reported failure directly: with a 24 s stall injected into the window, a measurement that failed a 15 s bound pre-fix now passes, while an equally slow operation with no stall to hide behind **still fails** - the fix must not have deleted the tests it was meant to save. The stall is injected rather than provoked because provoking a multi-second stop-the-world pause on demand is exactly as unreliable as the flake being fixed.
- Full `engine` suite: **12,093 tests, 0 failures, 0 errors** (`mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow`), plus the `slow` and `vector` lanes for the touched classes.

## Files

New (engine test sources):
- `engine/src/test/java/com/arcadedb/utility/JvmStallMonitor.java`
- `engine/src/test/java/com/arcadedb/utility/StallAwareStopwatch.java`
- `engine/src/test/java/com/arcadedb/utility/StallAwareStopwatchTest.java`

The helper lives in the `engine` test-jar, so the `server` and protocol modules can adopt it for their own timing assertions in a follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,19 @@ General design principles:
   - The tags are a partition, not a set of overlapping labels: CI selects each lane with `-Dgroups`/`-DexcludedGroups` such that every test runs in exactly one lane. Tagging a class `vector` and one of its methods `slow` is fine, but do not expect that method in the slow lane
   - Required imports: `import org.junit.jupiter.api.Tag;`
   - A JUnit tag on a Cucumber `@Suite` does not work: Surefire's `groups`/`excludedGroups` reach the scenarios inside the suite rather than the suite class. Route a suite to a lane with `-Dsurefire.includes` instead, as `.github/workflows/mvn-test.yml` does for the openCypher TCK
+- Never assert on raw wall-clock elapsed time. A full-suite run shares one JVM, and a stop-the-world pause of tens of
+  seconds late in a 12,000-test run turns any bound with less headroom than that into a coin flip on the JVM's mood
+  (#6260). Measure with `com.arcadedb.utility.StallAwareStopwatch` (engine test-jar) instead: it discounts the JVM-wide
+  stall observed inside the measured window, so the bound can stay tight AND stop flaking. Pick the assertion that says
+  what the number is for, because the message is what stops the next person from "fixing" a red run by loosening it:
+  - `assertGaveUpWithin(bound, whatItSeparates)` when the bound is a tripwire between a bounded operation and an
+    unbounded one. Generous is free here: a wider bound cannot turn a passing run red
+  - `assertStayedUnder(bound, claim)` when the bound IS the assertion - a complexity claim with no other practical
+    expression (a regex that must not backtrack exponentially, one deadline shared across rows rather than charged per
+    row). Loosening it deletes the test
+  - a short wait expected to TIME OUT needs neither, and neither does a lower bound (`isGreaterThan`): a stall only
+    makes those more true
+  - `@Timeout` is plain wall clock and cannot be discounted, so size it as a hang detector, not as a latency bound
 - don't add Claude as author of any source code
 
 ## Build and Development Commands
--- a/engine/src/test/java/com/arcadedb/database/Issue5587TxRetryExponentialBackoffTest.java
+++ b/engine/src/test/java/com/arcadedb/database/Issue5587TxRetryExponentialBackoffTest.java
@@ -21,6 +21,7 @@ package com.arcadedb.database;
 import com.arcadedb.GlobalConfiguration;
 import com.arcadedb.TestHelper;
 import com.arcadedb.exception.ConcurrentModificationException;
+import com.arcadedb.utility.StallAwareStopwatch;
 import org.junit.jupiter.api.Test;
 
 import java.util.ArrayList;
@@ -58,20 +59,29 @@ public class Issue5587TxRetryExponentialBackoffTest extends TestHelper {
     database.getConfiguration().setValue(GlobalConfiguration.TX_RETRY_DELAY_BASE, BASE_MS);
     try {
       final List<Long> failureTimestampsNanos = new ArrayList<>();
+      // #6260: a stop-the-world pause landing inside one gap would corrupt the comparison below in whichever
+      // direction it fell, so each gap is measured with the JVM's own stalls taken out of it.
+      final List<Long> failureStallNanos = new ArrayList<>();
       final AtomicInteger attempt = new AtomicInteger();
 
       database.transaction(() -> {
         if (attempt.incrementAndGet() < ATTEMPTS)
           throw new ConcurrentModificationException("forced retry");
-      }, false, ATTEMPTS, null, e -> failureTimestampsNanos.add(System.nanoTime()));
+      }, false, ATTEMPTS, null, e -> {
+        failureTimestampsNanos.add(System.nanoTime());
+        failureStallNanos.add(StallAwareStopwatch.jvmStallNanos());
+      });
 
       assertThat(attempt.get()).isEqualTo(ATTEMPTS);
       // one error callback per failed attempt, i.e. every attempt but the last
       assertThat(failureTimestampsNanos).hasSize(ATTEMPTS - 1);
 
       final List<Long> gapsMs = new ArrayList<>();
-      for (int i = 1; i < failureTimestampsNanos.size(); i++)
-        gapsMs.add((failureTimestampsNanos.get(i) - failureTimestampsNanos.get(i - 1)) / 1_000_000);
+      for (int i = 1; i < failureTimestampsNanos.size(); i++) {
+        final long spanNanos = failureTimestampsNanos.get(i) - failureTimestampsNanos.get(i - 1);
+        final long stalledNanos = failureStallNanos.get(i) - failureStallNanos.get(i - 1);
+        gapsMs.add(Math.max(0L, spanNanos - stalledNanos) / 1_000_000);
+      }
 
       final int half = gapsMs.size() / 2;
       final long earlyGapsSumMs = gapsMs.subList(0, half).stream().mapToLong(Long::longValue).sum();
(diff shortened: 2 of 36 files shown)


lvca:
Third review addressed.

**Point 2 (`LSMVectorIndexRebuildTest` categorization) - valid, fixed** in 6d166d1b4. It was the one `assertGaveUpWithin` in the set where nothing actually gives up. The rule the two names encode is that a tripwire separates a bounded operation from an unbounded one, and what this separates is a search served from the graph it already has from one that waits for the rebuild - both finite, and the bound is the whole claim. Now `assertStayedUnder`, which also lines it up with every other "must not wait on background work" assertion in the diff. I re-audited the other 14 `assertGaveUpWithin` sites against that rule and they all abort against a named limit (a thrown `TimeoutException`, a `close()` that force-shuts, a wait that returns `false` at its deadline), so this was the only slip.

**Point 4 (tracking issue for the other modules) - filed as #6270**, covering the six sites in `server` and `network`. Worth noting the server module is the worse environment for this, not the better one: those tests bind real ports and talk HTTP, so they already sit closer to their bounds.

**Dead-locals check you asked for before merge** - swept every changed file for locals and imports orphaned by the conversion. One hit: `WithAndUnwindTest.inlineTimeMs`, which timed the inline-property path and was never asserted on. It predates this PR, but it sat in the method just converted and was the last raw wall-clock measurement left there, so it went with the rest. Nothing else, and `mvn -o -pl engine -am test-compile` is clean.

**Point 1 (`check()` read ordering)** - agreed, and it is the accepted trade rather than something to fix: the two reads cannot be made atomic, and of the two orderings only this one biases toward a discount that is too generous. The other way round could under-discount and fail a run that was fine, which is the failure mode this PR exists to remove. It is documented at the call site and in `JvmStallMonitor`'s Javadoc, along with the reason `assertStayedUnder` bounds still want an order of magnitude of headroom rather than being tuned down to whatever currently passes.

**Point 3 (`Issue5587` non-atomic pair)** - agreed it is not a bug. The two reads are nanoseconds apart against a 10 ms sampler tick, and because every callback reads them in the same order the bias is common to both ends of each gap and cancels in the difference. Left as is.

lvca:
Fifth review: no blocking issues, three non-blocking nits. Answering them rather than changing anything.

**`@Tag("slow")` on `StallAwareStopwatchTest`** - measured rather than estimated: the class runs in **1.358 s** for all 7 tests. CLAUDE.md's `slow` lane is for "multi-second elapsed time", so it is under the line, not at the edge of it. The `Thread.sleep(1_000)` is most of that and it is load-bearing: the test asserts a sleeping thread is still charged for the second it slept, which is the property that stops every latch wait in the suite from silently becoming unbounded. Shortening it to save a second would weaken the one test that pins that behaviour, so it stays in the default lane.

**`check()` read ordering** - confirming this is the trade-off wanted long-term, not an accident. The two reads cannot be atomic, so the only choice is which way to be wrong. Reading the stall last means the discount can only be too generous; reading it first means it can be too small, which fails a run where nothing was wrong - the exact failure this PR exists to remove. A test that goes red for a reason unrelated to the code teaches people to ignore it, and an ignored suite hides far more than a slightly lenient bound does. The cost is bounded and stated in `JvmStallMonitor`'s Javadoc, and it is why `assertStayedUnder` bounds are documented as wanting an order of magnitude of headroom rather than being tuned down to whatever currently passes.

**The ~20 ms floor / `PaginatedSparseVectorEngineBackpressureTest`** - agreed, and already named in the Javadoc as the first place to look if that bound ever flakes.

On the unverifiable claim: the "12,093 tests, 0 failures" figure is `mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow` run locally on this branch, and the `slow` and `vector` lanes were run separately for the touched classes. CI's own `unit-tests` / `slow-unit-tests` / `vector-unit-tests` lanes on this PR are the independent check.

PR #6272, opened 2026-08-16
fix(#6184): an always-true filter is dropped at plan time, because a leaf can finally answer isAlwaysTrue()

### LINKED ISSUE #6184: An always-true filter is still evaluated once per record, because isAlwaysTrue() has no way to answer for a leaf
Spun out of #6174 / #6181.

### What happens

#6174 taught the planner to recognise a filter that is false for every record. The mirror case is untouched: a filter that is *true* for every record is still built as a filter step and evaluated per record.

```
SELECT FROM C                    -> + FETCH FROM TYPE C
                                      + FETCH FROM BUCKET 1 (C_0) ASC

SELECT FROM C WHERE 1=1          -> + FETCH FROM TYPE C WITH FILTER
                                      + SCAN WITH FILTER BUCKET 1 (C_0)
                                        1 = 1

SELECT FROM C WHERE true         -> + FETCH FROM TYPE C WITH FILTER
                                      + SCAN WITH FILTER BUCKET 1 (C_0)
                                        true
```

(Observed on `9ca99d29`.) The two statements are the same query; the second pays a per-record predicate evaluation, and loses the plain bucket fetch, to learn something the statement already said.

### Why it is stuck today

`BooleanExpression.isAlwaysTrue()` exists and is overridden by `AndBlock`, `OrBlock`, `ParenthesisBlock` and `NotBlock` - but **no leaf can implement it**, because the signature takes no `CommandContext` and folding a comparison needs one to reach `operator.execute(context.getDatabase(), ...)`. So `BinaryCondition` does not override it, and the recursion always bottoms out at `false`.

Two consequences:

- nothing is ever recognised as always-true, so the check has no consumer outside the AST's own recursion (`grep isAlwaysTrue` finds only `AndBlock`/`OrBlock`/`ParenthesisBlock`/`NotBlock`);
- `NOT (1=1)` is not folded as always-*false* either, even though `NotBlock.isAlwaysFalse` (added by #6181) delegates to exactly this method for its negated branch. `SELECT FROM C WHERE NOT (1=1)` still plans `SCAN WITH FILTER / NOT 1 = 1`.

### Suggested direction

Give `isAlwaysTrue` the same shape `isAlwaysFalse(CommandContext)` got in #6181: a context-taking form on `BooleanExpression`, folded on `BinaryCondition` for literal-only operands (`Expression.isLiteral()` is already there and already excludes bound parameters and function calls), and then drop the where clause in `SelectExecutionPlanner` when it is always true, so the plan is the one the statement without a WHERE would have got. `NOT (1=1)` then folds for free through the existing `NotBlock` delegation.

Two things to be careful about:

- **The no-arg `isAlwaysTrue()` has other callers inside the AST.** Keeping both forms means keeping them consistent; collapsing to the context-taking one means touching `AndBlock`/`OrBlock`/`ParenthesisBlock`/`NotBlock`. The second is cleaner but wider.
- **Dropping the filter is a bigger step than folding to an empty result.** An always-false fold replaces the plan with a step that produces nothing, and can only be wrong by returning too few rows - which the fold's own literal restriction rules out. Dropping an always-true filter changes which fetch step is chosen (`FetchFromTypeWithFilterStep` vs `FetchFromTypeExecutionStep`), so it wants a test that the two agree on ordering, bucket selection and the `readRecord` count.

The value here is smaller than #6174's - nothing sends `WHERE 1=1` as a probe the way Spark sends `WHERE 1=0` - so this is symmetry and a small constant per record, not a fix for a full scan.

robfrank:
Closes #6184

## What was wrong

The mirror of #6174. That issue taught the planner to recognise a filter that is false for every record; a filter that is *true* for every record was still built as a filter step and evaluated once per record, so the same query written two ways produced two different plans:

```
SELECT FROM C                    -> + FETCH FROM TYPE C
                                      + FETCH FROM BUCKET 1 (C_0) ASC

SELECT FROM C WHERE 1=1          -> + FETCH FROM TYPE C WITH FILTER
                                      + SCAN WITH FILTER BUCKET 1 (C_0)
                                        1 = 1
```

The second pays a predicate evaluation per record, and loses the plain bucket fetch, to learn something the statement had already said.

Nothing was stuck for want of a rule. `BooleanExpression.isAlwaysTrue()` had existed all along and was overridden by `AndBlock`, `OrBlock`, `ParenthesisBlock` and `NotBlock` - but **no leaf could implement it**, because the signature took no `CommandContext` and folding a comparison needs one to reach `operator.execute(context.getDatabase(), ...)`. So `BinaryCondition` never overrode it and the recursion always bottomed out at `false`. Two consequences: the method had no consumer outside the AST's own recursion, and `NOT (1=1)` was not folded as always-*false* either, even though `NotBlock.isAlwaysFalse` (added by #6181) delegates to exactly this method for its negated branch.

## The change

**`isAlwaysTrue` takes a `CommandContext`**, the shape `isAlwaysFalse` got in #6181. The no-argument form is removed rather than kept alongside it: a repo-wide grep finds no caller outside the four AST classes that recursed into it, so two forms would only have been two things to keep consistent - the wider of the two options the issue names, and the cleaner one now that it costs nothing.

**`BinaryCondition` folds both verdicts through one private helper**, `literalVerdict(context, expected)`, under the same restriction `isAlwaysFalse` already used: **both** operands must be `Expression.isLiteral()`. That excludes a bound parameter (the plan is cached and outlives the execution that bound it) and a function call (nothing on `SQLFunction` marks a function as pure). `WHERE ? = 1` and `WHERE uuid() IS NOT NULL` are therefore still planned as scans, as they must be. `BooleanExpression.TRUE` - what the parser answers a bare `WHERE true` with - now carries its own verdict, the way `FALSE` already did.

**The planner drops the where clause in `SelectExecutionPlanner.init()`**, before `optimizeQuery()` rearranges the clauses into index searches, so everything downstream sees the statement it would have seen without a `WHERE` at all: the projected properties computed for column pushdown, the choice between `FetchFromTypeWithFilterStep` and `FetchFromTypeExecutionStep`, and the hardwired `count(*)` plan. `NOT (1=0)` folds to always-true and `NOT (1=1)` to the always-false `EMPTY RESULT` of #6174, both for free through the existing `NotBlock` delegation. The same code builds the `SELECT` plan behind a subquery, a `DELETE ... WHERE` and an `UPDATE ... WHERE`, so all of them are covered.

One `OrBlock` detail changed with it: an empty disjunction used to answer `isAlwaysTrue()` with `true`. `OR` is false when it has no alternatives, and the answer is now load-bearing - the planner drops a filter on it - so it answers `false`, the conservative verdict, which at worst costs an optimisation. `isAlwaysFalse` already answered the empty block the same way.

## What the test pins

The issue's own caution is the design of the test: dropping an always-true filter is a wider step than folding an always-false one to an empty result, because it changes *which fetch step is chosen*, not just how many rows come back. So `ConstantTrueFilterFoldingTest` does not stop at "no filter step in the plan":

- the folded plan is compared **step class by step class** against the plan of the same statement written without its `WHERE` - which pins the fetch step chosen *and* the bucket steps under it, in order;
- the records returned and **their order** are compared against the unfiltered statement's, by RID;
- the `readRecord` statistic is compared against the unfiltered statement's;
- `NOT (1=1)` is asserted to fetch nothing and read zero records - the #6181 delegation that could not fire while no leaf answered `isAlwaysTrue`;
- the negative cases are asserted to keep their filter: `WHERE name = 'Arya'`, `WHERE 1=1 AND name = 'Arya'` (the fold is all-or-nothing, it does not prune satisfied terms out of a filter), `WHERE ? = 1` under two different bindings, and `WHERE uuid() IS NOT NULL`;
- `count(*)` with an always-true filter still answers 3, since the fold hands that statement to the hardwired count plan;
- `UPDATE ... WHERE 1=1` still touches every record and `DELETE ... WHERE 1=1` still removes every record.

## Test plan

- [x] `ConstantTrueFilterFoldingTest` - 16 tests, all green with the fix
- [x] The test was run against the unfixed sources (main-source changes stashed): **7 of the 16 fail**, and they are exactly the fold-specific ones (`anAlwaysTrueFilterIsNotEvaluatedPerRecord`, `theFoldedPlanIsTheOneTheStatementWithoutAWhereWouldHaveGot`, `anAlwaysTrueFilterIsRecognizedThroughParenthesesAndBooleanBlocks`, `aNegatedAlwaysFalseFilterFoldsToAlwaysTrue`, `aNegatedAlwaysTrueFilterFoldsToAnEmptyResult`, `anAlwaysTrueFilterInsideASubqueryIsFoldedToo`, `theFoldedPlanIsCacheableAndReusable`). The other nine are behaviour-parity guards that already held and must keep holding
- [x] Full `engine` unit lane, `mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow`: **12102 tests, 0 failures, 0 errors, 23 skipped** - including `ConstantFalseFilterFoldingTest`, whose `assertScan(planOf("... WHERE 1=1"))` and `assertScan(planOf("... WHERE true"))` still hold (a fetch step is still planned, it just no longer carries a filter)
- [x] Whole reactor compiles and installs (`mvn -o install -DskipTests`) - the signature change has no consumer outside `engine`

Not run locally: the wire-protocol module suites, because port 2480 was already held on this machine and those tests bind fixed ports; CI covers them.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2284,6 +2284,12 @@ One `OrBlock` detail changed with it: an empty disjunction used to answer `isAlw
 false when it has no alternatives, and the answer is now load-bearing, so it answers `false` - the conservative
 verdict, which at worst costs an optimisation.
 
+**API note for code that extends the SQL parser**: `BooleanExpression.isAlwaysTrue()` is replaced by
+`isAlwaysTrue(CommandContext)` rather than kept alongside it, so an out-of-tree subclass that overrode the no-argument
+form no longer overrides anything. This is a compile error against the new signature and a silently inert override
+only if the subclass keeps the old method as an unrelated one; nothing inside ArcadeDB called the no-argument form
+outside the four AST classes that recursed into it.
+
 The value here is smaller than #6174's: nothing sends `WHERE 1=1` as a probe the way Spark sends `WHERE 1=0`, so
 this is symmetry and a small constant per record rather than a full scan avoided.
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2280,6 +2280,13 @@ test therefore compares the folded plan against the plan of the same statement w
 class by step class, so the fetch step and the bucket steps under it both have to agree - and then compares the
 records returned, their order, and the `readRecord` count.
 
+One place the filter survives the whole-clause fold is behind an index. `WHERE 1=1 AND indexedProperty = 'x'` is not
+true for every record, so the clause stays - and then the index search takes the second term as its key and hands the
+`1=1` back as a *residual condition*, which was chained as a `FILTER ITEMS WHERE` step and evaluated once per index
+entry. A residual that is true for every record is now recognised as no filter at all, on both the single-index and
+the parallel-index paths, so the plan is the one the same statement without the constant term would have got. The
+index itself was never at risk: it was selected before and after, the residual only rode along behind it.
+
 One `OrBlock` detail changed with it: an empty disjunction used to answer `isAlwaysTrue()` with `true`. `OR` is
 false when it has no alternatives, and the answer is now load-bearing, so it answers `false` - the conservative
 verdict, which at worst costs an optimisation.
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2287,9 +2287,13 @@ entry. A residual that is true for every record is now recognised as no filter a
 the parallel-index paths, so the plan is the one the same statement without the constant term would have got. The
 index itself was never at risk: it was selected before and after, the residual only rode along behind it.
 
-One `OrBlock` detail changed with it: an empty disjunction used to answer `isAlwaysTrue()` with `true`. `OR` is
-false when it has no alternatives, and the answer is now load-bearing, so it answers `false` - the conservative
-verdict, which at worst costs an optimisation.
+One `OrBlock` detail changed with it: an empty disjunction used to answer `isAlwaysTrue()` with `true`. It now
+answers `false`, matching what `isAlwaysFalse` has always answered for the same block. Neither verdict reads it as
+the neutral element, even though that reading (an `OR` with no alternatives is `FALSE`) is available: the parser
+cannot produce an empty `OrBlock`, both answers are load-bearing - one drops the filter, the other replaces the plan
+with an empty source - and deducing either for a block that can only arrive by accident would trade correctness for
+an optimisation nothing can trigger. An empty `AndBlock`, by contrast, is exactly what a filter with no terms left
+looks like, so it does answer `isAlwaysTrue`, and that is what lets an emptied residual condition drop its step.
 
 **API note for code that extends the SQL parser**: `BooleanExpression.isAlwaysTrue()` is replaced by
 `isAlwaysTrue(CommandContext)` rather than kept alongside it, so an out-of-tree subclass that overrode the no-argument
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2234,3 +2234,57 @@ purpose ("more seeds than nodes" reads as "as many as exist" and is clamped to t
 nameless `NegativeArraySizeException`.
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216)
+
+## An always-true filter is dropped at plan time instead of being evaluated once per record (#6184)
+
+The mirror of #6174. That issue taught the planner to recognise a filter that is false for every record; a filter
+that is *true* for every record was still built as a filter step and evaluated per record, so `SELECT FROM C` and
+`SELECT FROM C WHERE 1=1` - the same query, written twice - produced two different plans:
+
+```
+SELECT FROM C                    -> + FETCH FROM TYPE C
+                                      + FETCH FROM BUCKET 1 (C_0) ASC
+
+SELECT FROM C WHERE 1=1          -> + FETCH FROM TYPE C WITH FILTER
+                                      + SCAN WITH FILTER BUCKET 1 (C_0)
+                                        1 = 1
+```
+
+The second paid a predicate evaluation per record, and lost the plain bucket fetch, to learn something the
+statement had already said.
+
+Nothing was stuck for want of a rule. `BooleanExpression.isAlwaysTrue()` had existed all along and was overridden
+by `AndBlock`, `OrBlock`, `ParenthesisBlock` and `NotBlock` - but **no leaf could implement it**, because the
+signature took no `CommandContext` and folding a comparison needs one to reach `operator.execute(...)`. So
+`BinaryCondition` never overrode it and the recursion always bottomed out at `false`: the method had no consumer
+outside the AST's own recursion, and `NOT (1=1)` was not folded as always-*false* either, even though
+`NotBlock.isAlwaysFalse` delegates to exactly this method for its negated branch.
+
+`isAlwaysTrue` now takes a `CommandContext`, the shape `isAlwaysFalse` got when it was introduced. The no-argument
+form is gone rather than kept alongside it - it had no caller outside the four AST classes that recursed into it,
+so two forms would only have been two things to keep consistent. `BinaryCondition` folds both verdicts through one
+private helper under the same restriction: **both** operands must be `Expression.isLiteral()`, which excludes a
+bound parameter (the plan outlives the execution that bound it) and a function call (nothing on `SQLFunction` marks
+a function as pure). `WHERE ? = 1` and `WHERE uuid() IS NOT NULL` are therefore still planned as scans, as they
+must be.
+
+The planner drops the where clause in `init()`, before the clauses are rearranged into index searches, so
+everything downstream sees the statement it would have seen without a `WHERE` at all: the projected properties
+computed for column pushdown, the choice between `FetchFromTypeWithFilterStep` and `FetchFromTypeExecutionStep`,
+and the hardwired `count(*)` plan. `NOT (1=0)` folds to always-true and `NOT (1=1)` to the always-false
+`EMPTY RESULT` of #6174, both for free through the existing `NotBlock` delegation.
+
+Dropping a filter is a wider step than folding one to an empty result: an always-false fold can only be wrong by
+returning too few rows, while dropping an always-true filter changes which fetch step is chosen. The regression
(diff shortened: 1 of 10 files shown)


robfrank:
Thanks - both minor items answered, one applied.

**Signature change in the release note** - applied in `93a48ee`. The note now carries an explicit API paragraph: `isAlwaysTrue()` is replaced by `isAlwaysTrue(CommandContext)` rather than kept alongside it, so an out-of-tree subclass that overrode the no-argument form stops overriding anything. Worth naming precisely because of the one shape that is *not* a compile error: a subclass that keeps its old `isAlwaysTrue()` method compiles fine and is silently never called.

**openCypher scope** - checked, and the answer is that Cypher does not share this code or this gap. The Cypher engine has its own rewriter stack in `com.arcadedb.query.opencypher.rewriter`: `ConstantFolder` already folds `5 > 3` to a literal `true` at rewrite time, and `BooleanSimplifier` already applies the identities on top of it (`x AND true -> x`, `x OR true -> true`, `NOT true -> false`). So the specific defect this PR fixes - a leaf that *cannot* fold because the method signature gives it no context - has no Cypher counterpart; `BooleanExpression.isAlwaysTrue` is SQL-AST-only and, as you confirmed, has no caller outside those four classes.

Whether the Cypher planner then *drops* a where clause that has been simplified to a bare literal `true`, the way `SelectExecutionPlanner.init()` now does for SQL, is a different question in a different pipeline. It is not what #6184 asks for - the issue names the SQL `isAlwaysTrue()` leaf specifically - so I have deliberately left it out of scope here rather than widening the change; it is a reasonable follow-up to file against the Cypher optimizer if measurement shows the filter surviving.

robfrank:
Both coverage gaps closed in `6c13ca4`, and the index one was not merely a missing test - **writing it found the remaining half of the same defect**, so thank you for asking for it.

**`WHERE 1=1 AND indexedField = 'x'`.** Your reasoning that the index is unaffected was right, and I confirmed it by printing the plans. What was *not* right was the rest of the plan. The AND is not true for every record, so the whole-clause fold in `init()` correctly declines; then index selection takes `name = 'Arya'` as the key and hands the `1=1` back as a **residual condition**, which was chained as its own step:

```
### SELECT FROM Character WHERE name = 'Arya'        ### SELECT FROM Character WHERE 1=1 AND name = 'Arya'
+ FETCH FROM INDEX Character[name]                   + FETCH FROM INDEX Character[name]
    name = 'Arya'                                        name = 'Arya'
+ EXTRACT VALUE FROM INDEX ENTRY                     + EXTRACT VALUE FROM INDEX ENTRY
+ FILTER ITEMS BY TYPE Character                     + FILTER ITEMS WHERE 1 = 1      <-- evaluated per index entry
                                                     + FILTER ITEMS BY TYPE Character
```

So the exact defect #6184 names - a filter true for every record, evaluated once per record - survived behind an index even after the fold. The first version of the new test asserted plan-shape parity with the un-constant statement and **failed on my own fix**, which is how it surfaced.

Both residual sites now go through one predicate, `needsFilterStep(remainingCondition, context)`: the single-index path in `handleIndexAsTarget` and the `createParallelIndexFetch` sub-plans. A residual that is `null`, empty, or true for every record is no filter at all. The two plans above are now identical step for step, which is what the test asserts, alongside the negative case (`WHERE age > 20 AND name = 'Jon'` keeps its residual filter and still returns only Jon).

**Empty `OrBlock`.** Pinned directly rather than by reachability, since as you say the parser does not produce one today and the verdict is now load-bearing: `assertThat(new OrBlock().isAlwaysTrue(null)).isFalse()`, with `new AndBlock().isAlwaysTrue(null)` asserted `true` next to it so the two neutral elements are documented against each other. `null` context is safe here - both answer from the sub-block list before the context is ever touched.

**Verification of the extra change**
- `ConstantTrueFilterFoldingTest`: 18 tests, all green
- `mvn -o -pl engine -am test` over all `com/arcadedb/query/**` and `com/arcadedb/index/**` tests: **8183 tests, 0 failures, 0 errors** - the index-plan tests (`QueryAndIndexesTest`, `SelectStatementExecutionTest`, the LSM index suites) are the ones with standing on this change and they are green
- release note extended with the residual paragraph

On the NPE you spotted in the old `NotBlock.isAlwaysTrue()` (`sub == null` with `negate == false`): agreed, and it is closed by the null-check being hoisted above the branch rather than living only in `isAlwaysFalse`.

robfrank:
On the empty-`OrBlock` loose end - I disagree with the fix and agree with the flag, so `97a10bc` changes the *comment* rather than the code, and pins both sides in the test.

The inference is exactly right from what I wrote, and what I wrote was misleading. `isAlwaysFalse` answering `true` for an empty disjunction is not a symmetry improvement, it is the one direction that can produce a wrong answer: that verdict replaces the whole plan with an `EMPTY RESULT` source (#6174). The parser cannot produce an empty `OrBlock`, so the only way one ever reaches these methods is a planner rewrite that emptied a block it did not mean to empty - and on that path "deduce FALSE from the neutral element" turns an accident into a query that silently returns no rows, while "decline" turns it into a filter that is merely evaluated. Both verdicts therefore decline it, which is what the code already did on the `isAlwaysFalse` side and what I made `isAlwaysTrue` do; my "neutral element, ie. FALSE" comment described the semantics I was *declining to use*, which reads as a half-finished implementation. It now says why.

Worth separating from its neighbour, because the two empty blocks genuinely differ: an empty **`AndBlock`** is not an accident, it is what a filter with no terms left actually looks like - it is precisely what an index search hands back once it has consumed every term - so `isAlwaysTrue` answering `true` there is both reachable and load-bearing, and it is what lets `needsFilterStep` drop the residual step. An empty `OrBlock` has no such producer. The direct test now pins all four answers next to each other so the asymmetry is documented rather than inferred:

```java
assertThat(new OrBlock().isAlwaysTrue(null)).isFalse();
assertThat(new OrBlock().isAlwaysFalse(null)).isFalse();
assertThat(new AndBlock().isAlwaysTrue(null)).isTrue();
assertThat(new AndBlock().isAlwaysFalse(null)).isFalse();
```

The release note carries the same distinction.

**Verification**
- `ConstantTrueFilterFoldingTest`: 18/18 green
- Full engine unit lane on the code as of `6c13ca4` (this cycle's change is a comment plus two assertions on top of it): **12104 tests, 0 failures, 0 errors, 23 skipped, BUILD SUCCESS**

On your sandbox note: the numbers in the PR body are from local runs, and the pre-fix failure claim is reproducible by stashing only `engine/src/main/java` and re-running the test class - 7 of the 16 fail, all of them fold-specific.

PR #6273, opened 2026-08-16
fix(#6151): Bolt's db.labels()/relationshipTypes()/propertyKeys() answer from the procedure registry

### LINKED ISSUE #6151: Bolt's db.labels()/db.relationshipTypes()/db.propertyKeys() are a separate, drifting implementation
### Context

Found while working on #6148 (`CALL proc() YIELD *` scope binding).

`bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java` has its own independent, string-matching-based implementation of `db.labels()`, `db.relationshipTypes()`, and `db.propertyKeys()` for the Bolt wire protocol. It never goes through `CallStep` or `CypherProcedureRegistry`, so it's now a second implementation of the same three procedures living outside the registry that PR #6150 just converged the native Cypher engine's own built-ins into.

This isn't a regression from #6150 (Bolt's copy was already independent before that PR), but it does mean:
- The two implementations can silently drift in behavior over time.
- Fixes applied to the registry-based versions (e.g. the argument-count validation PR #6150 added, or any future correctness fix) don't automatically apply to the Bolt path.

### Suggestion

Converge `BoltNetworkExecutor`'s handling of these three procedure names onto `CypherProcedureRegistry` (now that `db.labels`/`db.relationshipTypes`/`db.propertyKeys` are real `CypherProcedure` registry entries, per #6150) so `db.labels()` etc. behave identically whether invoked over Bolt or the native Cypher engine.

/cc from automated review on #6150

robfrank:
Closes #6151

`BoltNetworkExecutor.handleSystemQuery()` intercepts `db.labels()`, `db.relationshipTypes()` and `db.propertyKeys()` instead of running them through the query engine, and carried its own hand-written copy of each - a second implementation of the three procedures that #6150 had just converged into `CypherProcedureRegistry`. The two copies had already drifted: Bolt listed relationship types whose name carries Cypher's composite `A~B` label separator (the registry filters them out), returned property keys sorted alphabetically (the registry keeps schema order), and answered a call with arguments as if it had none, where the native `CALL` path rejects it on arity. This replaces the three branches with a delegation to the registry entries, so both paths now run the very same `CypherProcedure` objects.

## What changed

- **New `BoltSystemProcedures`** (bolt, package-private): resolves each name in `CypherProcedureRegistry`, executes it against the connection's database through a `BasicCommandContext`, and returns the procedure's own `getYieldFields()` as the Bolt record fields plus one row per yielded result. The combined `UNION` form Neo4j Desktop sends (three `collect(...)` rows under a single `result` field) is built from the same three procedures rather than from a third copy of the logic.
- **`BoltNetworkExecutor`**: the four branches collapse to one delegation (`+12 / -67`). When the helper answers `null` - the call carries arguments, which none of these procedures accept - the interception declines and the query falls through to the engine, so the client gets the registry's arity error instead of a silently-wrong full result.
- **Normalization moved into the helper** and is now `Locale.ROOT`-bound. Under a Turkish default locale `toLowerCase()` maps `I` to `ı`, which would have stopped `CALL DB.RELATIONSHIPTYPES()` matching at all.

The interception itself is kept: the combined Neo4j Desktop query is not something the Cypher engine parses, so removing it is not on the table here.

## Behavior changes visible to a Bolt client

| Call | before | after (= native Cypher) |
|---|---|---|
| `db.relationshipTypes()` | included composite `A~B` edge types | filtered out |
| `db.propertyKeys()` | alphabetically sorted | schema order |
| `db.labels(<arg>)` etc. | argument ignored, full result returned | engine's arity error, as for every other client |

`db.labels()` itself is unchanged - that one copy already matched the registry.

## Test plan

- [x] New `bolt/src/test/java/com/arcadedb/bolt/Issue6151BoltSchemaProcedureTest.java` (10 tests). Every assertion compares the Bolt answer against what `db.query("opencypher", …)` returns for the same call, so it pins convergence rather than a hardcoded expectation. The fixture is built to make the old and new answers differ: an edge type named `KNOWS~WORKS_WITH`, and property names declared out of alphabetical order.
- [x] Proven able to fail: with the pre-fix Bolt copy temporarily reinstated in the helper, `relationshipTypesMatchTheEngine` and `propertyKeysMatchTheEngine` fail (`expected "mango" but was "alpha"`, and `KNOWS~WORKS_WITH` "not expected"); with the fix in place all of them pass.
- [x] Whole `bolt` module unit suite green: `Tests run: 319, Failures: 0, Errors: 0`.
- [x] Covered: the combined Desktop `UNION` query, a call with arguments declining interception (standalone and inside the combined form), a connection with no database selected (single and combined form), a registry procedure that throws (declines to the engine rather than escaping the interception, which runs ahead of the executor's error classification), and locale-independent matching.
- [x] Engine side unaffected: `Cypher*Test` 1231 green, `Issue61*Test`/`OpenCypher*Test` 1982 green.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -57,7 +57,6 @@ import com.arcadedb.schema.DocumentType;
 import com.arcadedb.schema.EdgeType;
 import com.arcadedb.schema.Property;
 import com.arcadedb.schema.Schema;
-import com.arcadedb.schema.VertexType;
 import com.arcadedb.server.ArcadeDBServer;
 import com.arcadedb.server.HAServerPlugin;
 import com.arcadedb.server.security.ServerSecurityException;
@@ -88,7 +87,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.logging.Level;
 
 import static com.arcadedb.query.opencypher.executor.steps.FinalProjectionStep.PROJECTION_NAME_METADATA;
@@ -1174,7 +1172,7 @@ public class BoltNetworkExecutor extends Thread {
    * Returns true if the query was handled as a system query, false if it should be executed normally.
    */
   private boolean handleSystemQuery(final String query) throws IOException {
-    final String normalized = query.trim().toLowerCase().replaceAll("\\s+", " ");
+    final String normalized = BoltSystemProcedures.normalize(query);
 
     if (normalized.contains("dbms.components")) {
       // CALL dbms.components() - returns server version info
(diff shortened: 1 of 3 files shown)


robfrank:
Addressed in e22ce316eb:

- **Combined query with an argument**: added `theCombinedQueryDeclinesWhenOneCallCarriesAnArgument`, which pins that one bad sub-call sends the whole batch to the engine rather than answering the other two.
- **`serveSchemaProcedure`'s trailing `return null`**: commented as the defensive fallback for `isSchemaProcedureQuery()` and it ever diverging.

Not changed, deliberately: **`callHasArguments` matching the procedure name as a raw substring**. It is the same matching `handleSystemQuery` has always used for every one of its branches (`dbms.components`, `db.ping`, …), so tightening it to require a real `CALL <name>` occurrence would change interception for names this PR does not touch - a separate change with its own regression surface. Worth a follow-up issue if the sharp edge is ever hit; it is not a drift between the Bolt and Cypher paths, which is what #6151 is about.

robfrank:
Addressed in b9d67d093a:

1. **Interception runs outside `handleRun`'s try/catch** - real point, and the one thing here that changed risk: the branch used to be plain schema iteration and now runs registry machinery. `serveSchemaProcedure` now catches anything the procedure raises, logs it at FINE and declines. All three procedures are read-only, so handing the query to the engine has no side effect and the client gets the engine's properly classified error instead of the connection loop's catch-all. Pinned by `aFailingProcedureLeavesTheQueryToTheEngine`, which swaps a throwing stub into the registry under `db.labels` and asserts both the single and the combined form decline (and that the real entry serves again once restored).
2. **`callHasArguments` hard-coding zero arity** - the refusal is not a claim about arity: this path has only the query TEXT and cannot evaluate an argument, so any call that carries one has to be interpreted where arguments are evaluated. Documented that on `procedureFor`; a procedure that later grew an optional argument reaches the same engine path and is executed there.
3. **Combined form with no database selected** - added `theCombinedQueryWithoutADatabaseIsEmpty`; the query text is now a shared `COMBINED_QUERY` constant instead of being repeated.
4. **`isSchemaProcedureQuery`/`serveSchemaProcedure` recomputing the same three `contains`** - left as is. Threading the result through would mean either a second entry point or a sentinel to tell "not a schema query" apart from "declined", and three substring searches over a query string is not worth either.
5. **Substring matching firing on a string literal** - pre-existing and shared with every other `handleSystemQuery` branch; see the previous comment. Worth its own issue, not a change to smuggle into this one.

Local runs: `bolt` unit suite 319/319 green, engine `Cypher*Test` 1231 green.

robfrank:
Addressed in 905a81c6e5 (docs only):

- **`callHasArguments` is not a Cypher tokenizer**: documented exactly that, plus the property that makes the crude scan sufficient - every mistake it can make is in one direction. It walks to the first `)` after the name, so a nested paren or a stray `)` reads as an argument list and the call is *declined*; the engine then answers it. It can never mistake a call that has arguments for one that has none.
- **Registry entries swapped under these three names would run through a bare `BasicCommandContext`**: noted, no change. That is the same context shape the three read-only schema procedures need today, and `registerOrReplace` is not a client-reachable surface. If `db.*` names ever host something non-trivial, the interception should go away entirely rather than grow a context - worth its own issue at that point.
- **Trailing `return null`**: left as the guard it is, with the comment already on it.

Issue #6063, opened 2026-08-11
Bolt5002RoutingTableIT: DatabaseAreNotIdentical recurs 9 days after #5728's fix, with a different-looking root cause (bootstrap snapshot race, not replication skip)


lvca:
Follow-up observation from PR #6056's CI (unrelated engine change, `integration-tests` lane): `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites` failed with the same symptom #5728 was filed and closed for on 2026-08-02 (fixed by #5743, "a schema change no longer skips replication when another thread holds the recording session").

## Evidence

CI run [31462501377](https://github.com/ArcadeData/arcadedb/actions/runs/31462501377/job/93689824764), `integration-tests`, 2026-08-11:

```
[ERROR] com.arcadedb.bolt.Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites -- Time elapsed: 11.49 s <<< FAILURE!
org.opentest4j.AssertionFailedError:

expected: 1L
 but was: 0L
	at com.arcadedb.bolt.Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites(Bolt5002RoutingTableIT.java:140)
	Suppressed: com.arcadedb.database.DatabaseComparator$DatabaseAreNotIdentical: Types: DB1 6 <> DB2 5
		at com.arcadedb.database.DatabaseComparator.compareTypes(DatabaseComparator.java:129)
		at com.arcadedb.database.DatabaseComparator.compare(DatabaseComparator.java:49)
		at com.arcadedb.server.BaseGraphServerTest.checkDatabasesAreIdentical(BaseGraphServerTest.java:560)
		at com.arcadedb.server.ha.raft.BaseRaftHATest.checkDatabasesAreIdentical(BaseRaftHATest.java:253)
		at com.arcadedb.server.BaseGraphServerTest.endTest(BaseGraphServerTest.java:236)
```

That `Types: DB1 6 <> DB2 5` is the exact divergence #5728 documented. **But the surrounding log context looks like a different mechanism than what #5743 fixed:**

```
06:11:37.905 INFO  [Bolt5002RoutingTableIT] <ArcadeDB_2> Raft leader elected on server 2
06:11:37.927 INFO  [DriverFactory] Driver instance ... created for server uri 'neo4j://localhost:57699'
[...]

Issue #4147, opened 2026-05-09
feat(HA): offline cluster bootstrap from a pre-seeded database (snapshot-and-restore)


lvca:
## Problem

Bringing up a multi-node Raft HA cluster on top of a previously-imported 1+ GB database currently forces every new replica to receive the entire database over HTTP from the leader, in real time, while the leader is also serving writes. The operator-visible symptoms:

- Long catch-up windows (minutes to hours, per replica) during which the new follower's `STATUS` is `STALLED` and writes on the leader are competing with snapshot HTTP traffic.
- Failures we have seen in practice when `arcadedb.ha.snapshotDownloadTimeout` is hit on slow disks or constrained networks (`SnapshotInstaller: Snapshot download attempt N/3 failed`).
- No way to take advantage of out-of-band copy mechanisms the operator already has (S3, NFS, image bake-in, init containers, `kubectl cp`).
- No way to ship only the delta when a follower is just slightly behind the leader (e.g. brief disconnect, or pre-staged with a slightly older backup).

This is a real pain point reported by users running ArcadeDB on Kubernetes who want to scale a single-pod deployment up to 3 pods AFTER importing data.

## What exists today

1. **Leader-served full snapshot install**: `SnapshotHttpHandler` at `/api/v1/ha/snapshot/{database}`, triggered automatically when a follower lags past `arcadedb.ha.snapshotThreshold`. Always full database, regardless of how small the gap is.
2. **Replicated `restore database` on the leader**: `PostServerCommandHandler.replicateRestoredDatabase` calls `RaftReplicatedDatabase.createInReplicas(true)`, submits an install-database Raft entry with `forceSnapshot=true`. Each replica then pulls the full snapshot from the leader.

Both paths funnel the full database through the leader at runtime.

## Proposed feature

A two-part HA improvement, designed and shipped together because they share wire format and code paths.

### Part A - offline cluster bootstrap

Allow the operator to pre-seed each pod's filesystem with identical database files BEFORE the cluster forms, and have the Raft cluster recognise this as a valid baseline. After bootstrap, normal Raft replication continues on top of the pre-seeded state.

### Part B - delta resync

When a peer is *close* to the leader's state (`lastTxId` gap below a configurable threshold and the leader has retained the relevant WAL), ship only the WAL delta instead of the full database. Falls back to the existing full-snapshot path otherwise. Used in two places:

[...]

Issue #5100, opened 2026-07-07
HA offline bootstrap: bootstrap baseline lost on restart (not persisted in state-machine snapshot)


robfrank:
## Summary

`RaftBootstrapDoesNotEngageOnRestartIT.restartDoesNotReengageBootstrap` fails: first-formation bootstrap commits the baseline on all peers correctly, but after restarting a follower with **persistent Raft storage**, that follower's `getBootstrapBaseline(db)` returns `null` - the baseline is not re-established on restart.

Confirmed failing on pristine `main` (not a regression). Independent of #5098; with that fix the first-formation `await` now passes and the failure moves to the post-restart `await` (line 91, "baseline must be re-applied from Raft log on restart").

## Root cause

Instrumented run shows first formation works:

```
Database 'graph' bootstrapped locally (lastTxId=7, fingerprint matches cluster baseline)   [all 3 peers]
DIAG bootstrap outcome=COMMITTED commitIndex=2 isLeader=true
```

`ArcadeStateMachine.applyBootstrapFingerprintEntry` records the baseline into the **in-memory** `bootstrapBaselines` map (and `getBootstrapBaseline` reads from that map). On restart with persistent storage:

- the `BOOTSTRAP_FINGERPRINT_ENTRY` sits at a low log index that is **below the Ratis snapshot index**, so Ratis restores state from the snapshot rather than replaying that entry -> `applyBootstrapFingerprintEntry` is never called again;
- the `bootstrapBaselines` map is **not part of the state-machine snapshot** (`reinitialize` / snapshot load does not restore it), so it comes back empty.

Net effect: the committed, durable baseline in the Raft log is invisible after restart.

## Proposed direction

Persist / reconstruct `bootstrapBaselines` across restarts - e.g. include it in the `ArcadeStateMachine` snapshot (and restore it on load), or rebuild it from the persisted applied state. Care needed around the per-database applied-index bookkeeping introduced in #4824 and the snapshot-install path in #4835.

## Reproduce

```
mvn test -pl ha-raft -Dtest=RaftBootstrapDoesNotEngageOnRestartIT -DskipITs=false -DfailIfNoTests=false
```
(requires a fresh `arcadedb-engine` in `~/.m2`, e.g. `-am`)

## Related

- #4147 (offline bootstrap), #4824 (per-database applied index), #4835 (persistent Raft storage), #5098 (transient-`-1` fix, prerequisite)

PR #6274, opened 2026-08-16
fix(#6124): a database the bootstrap guard kept is recorded, re-verified and reported, not forgotten

### LINKED ISSUE #6124: Bootstrap 'local is fresher, refuse to overwrite' leaves a node's file-id space diverged with no active reconciliation
Follow-up from #6063 (fixed by #6118).

## Context

`BootstrapElection` picks ONE node as the cluster-wide bootstrap source at first formation, based on *aggregate* freshness across all databases - not per database. It is normal (not an operator error) for a different, non-elected node to independently reach a higher `lastTxId` for one *specific* database, purely from non-deterministic concurrent schema/security initialization at boot.

`ArcadeStateMachine.applyBootstrapFingerprintEntry`'s "local is fresher, refuse to overwrite" guard then correctly leaves that node's copy of the database alone rather than reinstalling it from the leader - the guard exists to protect a genuinely fresher operator backup from being silently discarded, and #6118 deliberately did not touch that philosophy.

The problem: once a node takes this path, its internal file-id counter is left permanently out of step with the rest of the cluster, and **nothing actively reconciles it**. #6118 closed the one symptom that was silently swallowing data (a later replicated schema change reusing a file id already in use on the diverged node, silently dropping the new component) by making that specific collision throw and trigger a targeted resync. But that is a *reactive* fix: it only fires when a later write happens to collide. A node that "wins" the refuse-to-overwrite guard and then never receives a colliding schema change stays permanently diverged from its peers for that database, discoverable only by an explicit `CHECK DATABASE` / manual comparison, or bad luck.

## Suggested direction

Rather than only detecting divergence reactively at the point a later entry collides, actively verify/reconcile a database that took the "refuse to overwrite" path shortly after bootstrap - e.g. a background fingerprint re-check (reusing the same `BootstrapFingerprint`/`bootstrap-state` machinery `BootstrapElection` already has) that either confirms the node's copy is a genuine superset of the leader's post-baseline state, or escalates to an operator-visible alert / triggers `resyncDatabaseFromLeader`-style reconciliation instead of leaving it silently divergent indefinitely.

Not urgent (no known production incident), filed as a robustness follow-up while the mechanism was fresh from the #6063 investigation.

robfrank:
## What does this PR do?

Makes the bootstrap "local is fresher, refuse to overwrite" refusal a **recorded, re-verified, operator-visible** state instead of a single log line that scrolls away.

1. **Durable mark.** Taking the refusal branch of `ArcadeStateMachine.applyBootstrapFingerprintEntry` records the database in a new `bootstrapUnreconciledDatabases` set, persisted as an `unreconciled` flag on that database's entry in the existing `.raft/bootstrap-baselines` file. It has to be persisted rather than kept in memory precisely because the per-database replay-skip stops the branch from re-evaluating after a restart: an in-memory mark would vanish on the first bounce. The flag is written only when set, so a healthy cluster's file keeps exactly the shape it had before and an older build ignores the extra key.

2. **Active verification.** A new `verifyBootstrapDivergence()` hook on the `HealthMonitor` tick probes the leader over the *existing* `POST /api/v1/cluster/bootstrap-state` RPC for exactly those databases and either
   - **confirms convergence** (the leader now reports the same fingerprint, i.e. the two copies are byte-identical over the persisted content) and drops the mark, or
   - **escalates**: a SEVERE naming both `(lastTxId, fingerprint)` pairs and the resync endpoint, with the mark left raised.

   It needs its own hook for the same structural reason `retryUnfilledSnapshotGap()` does: such a node is not lagging and not diverged in the Raft sense - it applies every entry it is sent and reports zero lag - so no lag- or divergence-driven check can ever see it. Free in the normal case (empty set, one volatile read), skipped on the leader, throttled to one probe per 5 minutes (floored at the snapshot watchdog cadence) with a 5 s per-probe HTTP ceiling, because the probe makes the leader hash every database directory it holds.

3. **Operator-visible alert.** A `bootstrap-diverged-databases` critical alert in `ClusterAlerts`, so the condition is readable from `GET /api/v1/cluster` (and Studio's HA panel) for as long as it lasts.

4. **Clearing.** The mark is dropped exactly where this node's copy is actually replaced by the leader's: the bootstrap install, the forced `forceSnapshot` reinstall, the targeted quarantine resync, the operator-triggered `POST /api/v1/cluster/resync/{database}`, the full resync, and the leader-driven snapshot install. A dropped database loses its mark with its baseline.

### Why it deliberately does not resync by itself

Reinstalling from the leader is exactly what the guard refused to do. Doing it later, in the background, would discard the fresher operator data the guard exists to protect - and the leader's `lastTxId` passes this node's within seconds of the cluster taking any write, so an "is the leader ahead now?" trigger would fire almost immediately and quietly reverse the guard. The automatic action is therefore to *verify and surface*; the destructive reconciliation stays one explicit operator request away, and both the alert and the log line name it. This keeps the philosophy #6118 preserved.

### Reuse

The election's async fan-out and the new synchronous probe now share one request builder (`BootstrapElection.bootstrapStateRequest`) and one response parser (`parseBootstrapState`), so the two paths cannot drift on credentials or on what the response means. `applyBootstrapFingerprintEntry` and the verification share `readLocalBootstrapState`, so they cannot drift on what "local state" means either. Fingerprints are abbreviated for logging through the election's existing helper.

## Motivation

`BootstrapElection` elects one cluster-wide bootstrap source from *aggregate* freshness across all databases, so it is normal - not operator error - for a different, non-elected node to independently hold a higher `lastTxId` for one *specific* database after a concurrent cold boot. The overwrite guard then correctly leaves that node's copy alone.

What the refusal left behind was nothing. The node's file ids stay assigned by a history no other peer shares, and the only trace was one SEVERE emitted once, at bootstrap. #6118 closed the one consequence that silently ate data (a later replicated schema change reusing a file id already in use there), but reactively: it fires only if such an entry ever happens to collide. A node that never receives a colliding schema change stayed diverged indefinitely, discoverable only by an explicit `CHECK DATABASE`, a manual comparison, or bad luck.

## Related issues

Closes #6124. Follow-up to #6063 / #6118 (the reactive half of the same divergence), and reuses the #4147 bootstrap-state machinery.

## Additional Notes

- No new configuration knob and no new HTTP endpoint: the probe reuses the authenticated cluster RPC that already exists, and the check interval / probe timeout are internal constants (5 min / 5 s).
- `getBootstrapUnreconciledDatabases()` is read on every Studio status poll and returns `Collections.emptyList()` without allocating in the (overwhelmingly common) healthy case.
- A marked database that is not currently *loaded* on the node is neither cleared nor opened to be fingerprinted: `existsDatabase` answers "is it in the registry", not "is it on disk", so retiring the alert there would drop a real divergence for a database an operator merely left closed.

## Checklist

- [x] I have run the build using `mvn clean package` command (ran `mvn -pl ha-raft -am install -DskipTests` plus the full `ha-raft` unit suite)

### Test plan

New unit tests, all in `ha-raft`, no mocking framework (real unstarted `ArcadeDBServer` + real `LocalDatabase`, the `ArcadeStateMachineBootstrapBaselinePersistenceTest` harness):

- `ArcadeStateMachineBootstrapDivergenceTest` (14 tests): the refusal records the divergence; a matching baseline does not; the mark survives a restart; the persisted shape (flag present when diverged, absent when not); a matching leader fingerprint clears it, durably; a differing one, and a leader that does not report the database, both keep it raised; a database not loaded here is neither cleared nor opened; per-database and full resync clear it; a drop clears it; marking one database preserves another's baseline.
- `HealthMonitorBootstrapDivergenceTest`: every tick drives the verification, including on a CLOSED Ratis lifecycle; a shutdown request skips it.
- `ClusterAlertsBootstrapDivergenceTest`: the alert's id/severity/details and the resync endpoint in the recommendation; empty and null lists raise nothing.
- `BootstrapStateProbeParsingTest`: the shared request carries the cluster token and forwarded-user headers (and omits a blank token); the shared parser honours the database filter and throws on a malformed body.

**The tests can fail:** with `markBootstrapUnreconciled(dbName)` removed from the refusal branch, 7 of the 14 divergence tests fail.

**No regressions:** full `ha-raft` unit lane (`-DexcludedGroups=benchmark,vector,slow`) green.

### DIFF
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -2395,21 +2395,24 @@ public class ArcadeStateMachine extends BaseStateMachine {
     if (leaderHttpAddr == null || raftHA.isOwnHttpAddress(leaderHttpAddr))
       return; // no leader to compare against yet
 
-    final long now = System.currentTimeMillis();
     // Floored at the snapshot cadence so a WAN cluster that has widened its watchdog does not get probed
     // more often than it resyncs.
-    final long checkIntervalMs = Math.max(BOOTSTRAP_DIVERGENCE_CHECK_INTERVAL_MS, computeSnapshotWatchdogTimeoutMs());
-    final long previous = lastBootstrapDivergenceCheckMs.get();
-    if (previous != 0 && now - previous < checkIntervalMs)
+    if (!claimBootstrapDivergenceCheckSlot(System.currentTimeMillis(),
+        Math.max(BOOTSTRAP_DIVERGENCE_CHECK_INTERVAL_MS, computeSnapshotWatchdogTimeoutMs())))
       return;
-    if (!lastBootstrapDivergenceCheckMs.compareAndSet(previous, now))
-      return; // another tick won the throttle slot
 
     final Set<String> pending = new HashSet<>(bootstrapUnreconciledDatabases);
     final String clusterToken = raftHA.getClusterToken();
     try {
       // Off the HealthMonitor thread: the probe is a blocking HTTP call to the leader and the monitor
       // tick drives the Ratis lifecycle checks behind it.
+      //
+      // On the lifecycleExecutor like every other leader-facing task, and bounded so it cannot become the
+      // thing this module's CLAUDE.md warns about: the executor is single-threaded and already runs
+      // multi-minute full resyncs on it, so what matters is that this task's worst case is small next to
+      // those. It is - BOOTSTRAP_DIVERGENCE_PROBE_TIMEOUT_MS (5 s) at most, at most once per check window
+      // (>= 5 minutes), against downloads measured in minutes. A queued resync therefore waits seconds in
+      // the worst case, not for the length of a download.
       lifecycleExecutor.submit(() -> {
         final Map<String, BootstrapBaseline> leaderStates = BootstrapElection.fetchBootstrapState(
             leaderHttpAddr, clusterToken, pending, BOOTSTRAP_DIVERGENCE_PROBE_TIMEOUT_MS);
@@ -2425,6 +2428,20 @@ public class ArcadeStateMachine extends BaseStateMachine {
       LogManager.instance().log(this, Level.WARNING,
           "Cannot schedule the bootstrap-divergence verification: executor is shut down", ree);
     }
+  }
+
+  /**
+   * Claims the one bootstrap-divergence probe slot per {@code intervalMs} window, returning whether the
+   * caller may probe. A lost CAS means another tick took the slot, so this one stands down rather than
+   * probing twice. Extracted (package-private) so the throttle is testable without a live leader: the
+   * rest of {@link #verifyBootstrapDivergence()} needs a reachable peer to reach it.
+   */
+  // @VisibleForTesting
+  boolean claimBootstrapDivergenceCheckSlot(final long now, final long intervalMs) {
+    final long previous = lastBootstrapDivergenceCheckMs.get();
+    if (previous != 0 && now - previous < intervalMs)
+      return false;
+    return lastBootstrapDivergenceCheckMs.compareAndSet(previous, now);
   }
 
   /**
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java
@@ -2417,9 +2417,14 @@ public class ArcadeStateMachine extends BaseStateMachine {
         final Map<String, BootstrapBaseline> leaderStates = BootstrapElection.fetchBootstrapState(
             leaderHttpAddr, clusterToken, pending, BOOTSTRAP_DIVERGENCE_PROBE_TIMEOUT_MS);
         if (leaderStates == null) {
+          // The throttle slot is spent whether or not the probe answered, exactly as the stale-snapshot
+          // backstop spends its own on a failed attempt: the next try is the next check window, not the
+          // next health tick. Said plainly here so the wait is not read as seconds. The mark stays raised
+          // in the meantime, so a failed probe never retires an alert.
           LogManager.instance().log(this, Level.INFO,
-              "Could not verify bootstrap divergence of %s against leader %s; retrying on a later health tick",
-              pending, leaderHttpAddr);
+              "Could not verify bootstrap divergence of %s against leader %s; the divergence stays reported "
+                  + "and the check is retried in the next window (>= %d ms)",
+              pending, leaderHttpAddr, BOOTSTRAP_DIVERGENCE_CHECK_INTERVAL_MS);
           return;
         }
         reconcileBootstrapDivergence(leaderStates);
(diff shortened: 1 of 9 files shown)


robfrank:
Thanks - both points addressed in `9cf23b8` (2 files, +59/-6).

**Test gap on `verifyBootstrapDivergence()`.** Agreed, and the reason it was untested is the one you identified: everything past the leader-address resolution needs a reachable peer. Fixed by extracting the throttle into a package-private `claimBootstrapDivergenceCheckSlot(now, intervalMs)` (read-then-CAS, unchanged semantics) and testing it directly:

- `theProbeIsThrottledToOneAttemptPerWindow` - first attempt claims, an attempt 1 ms later and one at `window - 1` both stand down, `window` re-arms.
- `verificationWithoutARaftServerIsANoOpThatKeepsTheThrottleSlot` - drives the real `verifyBootstrapDivergence()` on a state machine with a marked database and no Raft server wired, and asserts both that the mark is untouched **and** that the throttle slot was not consumed. The second half is the part worth having: a node whose first ticks happen before a leader exists must still get a full-rate first probe once one appears, and the ordering that guarantees it (address resolution before the throttle claim) is exactly the kind of thing a later edit can silently invert.

16 tests in that class now, 183 in the `ArcadeStateMachine*`/`Bootstrap*`/`ClusterAlerts*`/`HealthMonitor*`/`RaftBootstrap*` selection, all green.

**`lifecycleExecutor` occupancy.** You read it right and I have not changed the placement - the executor is where every other leader-facing task goes, and moving this one off it would be the more surprising choice. What was missing was the argument for why it is safe, so it is now stated at the submit site rather than left for the next reader to re-derive: the worst case is `BOOTSTRAP_DIVERGENCE_PROBE_TIMEOUT_MS` (5 s), at most once per check window (>= 5 min), against the multi-minute full resyncs that executor already runs. So a queued resync waits seconds in the worst case, which is the distinction the module's CLAUDE.md rule is actually about ("must not be parked for the length of a download") - not "must never block at all".

No change to the `/api/v1/cluster/resync/{database}` recommendation or the `visible()` filtering, both of which you confirmed.

robfrank:
Thanks - two of the three acted on in `ba795e1`, the third declined with reasoning.

**The unreachable defensive loop.** Agreed, and I took the version of the fix that removes the need for it rather than just deleting it: the loader now refuses an `unreconciled` flag on an entry that carries no `fingerprint` (`continue` before both reads). That turns "every marked database has a baseline" into an invariant of the in-memory state rather than a property of the current call graph, which is what makes iterating `bootstrapBaselines` alone in the writer provably sufficient instead of merely unreachable-by-inspection. New test `aMarkWithoutABaselineIsNotLoaded` pins it, so a future edit that reintroduces the shape fails rather than silently dropping a mark on the next persist.

**"retrying on a later health tick".** You are right that the message reads sooner than it is. Reworded to say the divergence stays reported and the check retries in the next window, with the interval interpolated. Kept the behaviour itself (spend the slot on a failed probe) precisely because it matches `retryUnfilledSnapshotGap` - and the cost of the delay is bounded by design here: the mark stays raised throughout, so a failed probe can never retire an alert, it only postpones the chance to clear one.

**Two fsyncs on the refusal path - declined.** Combining them would mean deferring `recordBootstrapBaseline` until the branch is known, and the baseline has to be persisted on the paths that return *before* that point (the late-joiner "database not present locally" return, and the `!(embedded instanceof LocalDatabase)` return) - #5100 is exactly the bug where a baseline that was not persisted became invisible after a restart. So the second write is not a redundant write of the same event; it is a second event that happens to follow closely. Given it is two writes once per lifetime of a diverged database, on the rarest branch in the method, I would rather not restructure a #5100-critical ordering for it. Happy to revisit if you disagree.

184 tests green across `ArcadeStateMachine*`/`Bootstrap*`/`ClusterAlerts*`/`HealthMonitor*`/`RaftBootstrap*`.

Issue #6048, opened 2026-08-10
Super-node full traversals pay the interleaving locality cost past the rank where it buys anything


lvca:
Follow-up to #6044 / #6047.

`StripedEdgeList.edgeIterator()`, `vertexIterator()` and `ridIterator()` now interleave the stripe chains round-robin so that the first N edges of a promoted super-node are approximately the newest N. That is the right trade for a **paged** read. It is the wrong trade for a **full walk**, and both go through the same three methods.

#6044 anticipated the locality cost and answered it by keeping the maintenance walks (`edgeIteratorForRemoval`, `entryIterator`, counting, export) on plain concatenation. But those are not the only full walks. A perfectly ordinary query - `MATCH (h)-[:LINK]->(x) RETURN x`, a Gremlin `out()` step, `SELECT expand(out()) FROM Hub` - consumes `vertexIterator()`/`edgeIterator()` to exhaustion, and now does it with `stripes` (default 16) live cursors: 16 resident chunk pages and a hop across 16 files every 16 edges, instead of one cursor draining one chain at a time.

## The observation that makes this cheap to fix

The ordering guarantee is only worth something for small ranks. By construction the deviation of a returned position from the true recency rank `r` grows with `r`; past a few multiples of `stripes` the approximation tells a caller nothing it could act on, and a caller still reading at that point is doing a full walk, not paging.

So the rotation can **degrade to concatenation** once it has emitted, say, `K x stripes` entries: keep round-robin for the prefix where the order is meaningful, then drain the remaining chains one at a time and recover full-walk locality for the rest. The head of the list keeps exactly the property #6044 restored; the tail gets back the locality it had before. No caller has to opt in and no API changes.

`InterleavedIterator` is the single place this would live (it already owns the rotation array and the exhaustion compaction), so the change is contained.

## What to establish first

- **Measure it.** How much does a full traversal of a promoted super-node actually cost now versus before #6047? `SuperNodeConcurrentAppendBenchmark` and `SuperNodeDeleteBenchmark` are the existing harnesses to extend; the numbers decide whether this is worth any complexity at all, and I did not measure it before merging.
- **Pick K on evidence, not taste.** It is the crossover between "the order still means something" and "this is a full walk".
- Check whether the degrade should be by entry count or by whether the consumer has read past the whole newest generation.

PR #6275, opened 2026-08-16
fix(#6064): a super-node read walk widens its rotation instead of abandoning the order

### LINKED ISSUE #6064: GRAPH_SUPERNODE_INTERLEAVE_ROUNDS is a static threshold, blind to the query's actual LIMIT
Follow-up to #6048 (PR #6056). `StripedEdgeList.interleaved()` degrades a promoted super-node's read walk from interleaved (approximately newest-first) order to plain concatenation once it has emitted `arcadedb.graph.supernodeInterleaveRounds x stripes` entries of a generation (default 64 rounds = 1024 entries at the default 16 stripes). That threshold is a single per-database config value, set once and applied identically to every read of every promoted vertex, regardless of what the actual query asked for.

## The gap

A query with a small `LIMIT` (the case the interleaving exists to serve) stays well inside the default threshold and gets the full ordering benefit. But a query with a **larger, still-bounded** `LIMIT` - not uncommon in exports, batch jobs, or paginated admin tooling, e.g. `MATCH (h)-[:LINK]->(x) RETURN x LIMIT 5000` - silently loses the newest-first approximation past entry 1024, with nothing to indicate that happened. The caller asked for a bounded result and got one, just not in the order they might expect from the "approximately newest-first" documentation, and there's no way to raise the effective threshold for that one query without raising `supernodeInterleaveRounds` globally (which reintroduces the #6048 full-walk locality cost for every unlimited read too).

This was raised as a non-blocking observation in `claude-review`'s second round on PR #6056 ("worth double-checking against production data since a caller doing a large-but-bounded LIMIT ... will silently lose the guarantee") and judged out of scope for that PR - the config description already documents the threshold precisely, so nothing is silently wrong, but the behavior itself could be sharper.

## A better direction than raising the default

Rather than tuning the constant, the LIMIT value could flow from the query engine down into the read walk, and `interleaved()` could use `max(configured floor, effective LIMIT)` as the degrade point for that specific read - giving every bounded query exactly the ordering fidelity it needs without paying for it on every unbounded one. That's a bigger change than #6048's: it requires a path for a "how many do you actually need" hint to reach `StripedEdgeList`/`EdgeLinkedList` from wherever `LIMIT` is known (the SQL/Cypher/Gremlin execution plan), which today's edge iterator API has no concept of.

## Why this is filed as "consider" rather than urgent

- No one has reported hitting this in practice; it is a natural consequence of #6048's own design examined ahead of time, not an observed defect.
- The default (64 rounds / 1024 entries) already covers most realistic UI paging sizes; only unusually large bounded LIMITs are affected.
- The fix shape is speculative complexity (threading a LIMIT hint across the query-engine/graph-engine boundary) for a case with no confirmed real-world impact yet - worth deferring until it's needed rather than building ahead of demand.

## Where to look first if picked up

- `StripedEdgeList.interleaved()` (`engine/src/main/java/com/arcadedb/graph/StripedEdgeList.java`) is the only call site of `InterleavedIterator`'s degrade constructor - it already computes `degradeAfter` once per call, so wiring in a caller-supplied hint would be localized there.
- Would need to establish how far a LIMIT is already known when `edgeIterator()`/`vertexIterator()`/`ridIterator()` are invoked from the execution plan (SQL `FetchEdgesStep`-equivalent per language) - if it's not threaded through today, that's the actual size of this change, not the `StripedEdgeList` edit itself.

robfrank:
Closes #6064

## The gap

`GRAPH_SUPERNODE_INTERLEAVE_ROUNDS` (#6048) is a **cliff**, not a taper. Up to `rounds x stripes` entries - 64 x 16 = 1,024 at the defaults - a promoted super-node's read walk takes one entry per stripe chain per turn, which is what puts an edge of recency rank `r` back at a position of order `r` (#6044). The very next entry froze `turn`: the chain holding it was drained to exhaustion, then the next, then the next. Past 1,024 the returned position had no relation to recency and the error grew with the vertex DEGREE again - the exact failure #6044 fixed, pushed past a constant.

A small `LIMIT` never reaches the cliff. A **large but still bounded** one - an export, a batch job, paginated admin tooling, `MATCH (h)-[:LINK]->(x) RETURN x LIMIT 5000` - fell off it silently. The only lever was raising `supernodeInterleaveRounds` for every unbounded read in the database too, which reintroduces exactly the full-walk locality cost #6048 removed.

## Why not thread the LIMIT down

The issue's own "better direction" was `max(configured floor, effective LIMIT)` as the degrade point. Establishing how far a LIMIT is known at the call site (the issue's "where to look first") gives the answer the issue suspected: **nowhere**. `edgeIterator()`/`vertexIterator()`/`ridIterator()` are reached through `Vertex.getEdges`/`getVertices` from `SQLFunctionOut`, the Cypher expand operators, Gremlin's `out()` and the `MATCH` traversers, and none of them carry the statement's `LIMIT`; `LimitExecutionStep` only narrows the `nRecords` *batch size* of the pull above it. Threading a hint would mean new overloads on `Vertex`, `VertexInternal`, its three implementations, `GraphEngine`, `EdgeSegmentList`, `EdgeLinkedList` and `StripedEdgeList`, plus a change in every query engine - a large permanent API surface for a case with no confirmed real-world report, which is what the issue flagged as speculative complexity.

## What this does instead

Make the trade-off **continuous**, so no bounded read has a cliff to fall off and nothing has to know the LIMIT.

Past `rounds x stripes` entries `InterleavedIterator` keeps rotating, but takes `rounds` entries from each chain per turn instead of one, and **doubles that batch on every completed round**. Both properties hold at once:

- **Locality** (#6048's concern): chain switches over a walk of `D` entries drop from `D` to about `stripes x log D`, and once the batch outgrows a chunk every visit is a sequential run through one chain, so the resident-page pressure decays instead of persisting. Counted on a 1,000,000-edge walk at the defaults: ~1,184 switches, against ~1,040 for the previous freeze - 0.1% of the entries walked, either way.
- **Rank fidelity** (#6064's concern): an entry at depth `d` in its chain is emitted no later than the end of the batch that reaches depth `d`, which is at most `2d`, so its position stays within a constant factor of its true rank for the **whole** walk.

`rounds <= 0` keeps its documented meaning of "interleaving off": the batch is passed as `Long.MAX_VALUE`, which never rotates again - i.e. immediate concatenation, the pre-#6044 order - so `zeroRoundsReproducesTheClassicPreInterleavingScatter` still holds. The two-argument `InterleavedIterator` constructor keeps the pre-#6064 drain-to-exhaustion semantics unchanged and is pinned as such.

Read-side only: no on-disk format change, no new setting, no migration.

## Measurements

On a 3,000-edge super-node at 16 stripes and `rounds=4` (a 64-entry threshold, deliberately small so most of the walk is past it):

| | before | after |
|---|---|---|
| worst position of the newest 1,000 edges | 2,782 of 3,000 | < 2,000 |
| worst position/rank ratio | 58 (rank 40 at position 2,386) | 2.4 (rank 190 at position 463) |

Both figures come from `Issue6064BoundedLimitOrderingTest`, which was written first and fails on `main` with those numbers.

## Test plan

- [ ] `mvn -o -pl engine -am -Dtest=Issue6064BoundedLimitOrderingTest -DskipITs test` - new regression test: a large bounded LIMIT still gets the newest edges; every rank comes back at a position of order of that rank; the walk is still complete and duplicate-free. Verified RED before the fix (worst 2,782 / ratio 58), GREEN after.
- [ ] `mvn -o -pl engine -am -Dtest=InterleavedIteratorTest -DskipITs test` - 21 tests, 5 new: the exact widening sequence traced entry by entry, nothing lost or duplicated at any threshold x batch combination, a `Long.MAX_VALUE` batch reproduces the two-argument constructor entry for entry, a non-positive threshold arms the widening immediately, and `reset()` re-arms it. The pre-existing `degradesToConcatenationOfTheRemainderPastTheThreshold` is untouched and still passes.
- [ ] `mvn -o -pl engine -am -Dtest=Issue6048SupernodeFullWalkDegradeTest,Issue6044StripeIterationOrderTest -DskipITs test` - the two tests this change is most likely to break: both pass unmodified, including the `rounds=0` scatter signature and the per-generation threshold isolation.
- [ ] `mvn -o -pl engine -am -DskipITs -DexcludedGroups=benchmark,vector,slow test` - full engine unit lane: **12,094 tests, 0 failures, 0 errors, 23 skipped**.


robfrank:
Thanks - three non-blocking items, all three addressed or answered in `2cb1a0b`.

**1. The transition accounting.** Applied the comment, but the arithmetic in the observation does not hold: the source holding the turn at the crossing takes `degradeBatch` consecutive entries, not `2*rounds - 1`. The crossing entry is not double-counted, because `taken` is **re-based** to 0 there and then incremented, so that entry becomes the 1st of the new visit rather than being added to a visit already in progress. Traced with `degradeAfter=6`, `degradeBatch=2`, 3 sources: entries 1-5 are `a1 b1 c1 a2 b2`; entry 6 is `c2` and crosses the threshold; `c` then takes `c3` and rotates - a run of 2 = `degradeBatch`, the same length every other source gets on its first wide visit. (Below the threshold `batch` is 1, so every source's quota is always reset to 0 by the rotation of the preceding entry; there is nothing left over to extend.) The new comment states that explicitly, which is the useful half of the observation.

**2. Duplicated arming between the constructor and `reset()`.** Applied - extracted `armWidening()`, which now owns the whole starting state (`widening`, `batch`, `taken`, `rotationsInRound`). The two call sites can no longer drift, which is the real risk in that duplication, and the redundant field initialisers went with it.

**3. `@Tag("slow")` on `Issue6064BoundedLimitOrderingTest`.** Declined, on measurement rather than on precedent: the three methods run in **1.5 s** wall clock together locally (0.28-0.70 s each), the same order as the sibling `Issue6048SupernodeFullWalkDegradeTest`'s 0.96 s for six methods. `CLAUDE.md` scopes the tag to "noticeably long ... multi-second elapsed time", and the lanes are a partition, so tagging this would move it out of the default unit lane for no gain. Worth revisiting if CI measures something very different from local.

Re-ran after the change: `InterleavedIteratorTest`, `Issue6064BoundedLimitOrderingTest`, `Issue6048SupernodeFullWalkDegradeTest`, `Issue6044StripeIterationOrderTest`, `Issue6049GAVSupernodeOrderTest` - 35 tests, 0 failures.

robfrank:
Both notes checked; `54aa18d` pins the first one in code.

**`rotationsInRound` cadence when a source drops mid-visit.** Correct reading, and it is the intended one - now written down on the field rather than left to be re-derived. A visit cut short by exhaustion drops its source, which shrinks `activeCount`, and `activeCount` is exactly the bound the counter is compared against, so the two move together and the cadence stays self-correcting. The class promises a BOUND on the batch's growth (hence on the switch count and the rank error), never an exact round length. Counting a dropped source as a rotation would not make the cadence more exact, and it would double the batch for the survivors on the strength of a visit that never happened - so the comment also records why the obvious "fix" is the wrong one.

**3,000 single-transaction inserts in `Issue6064BoundedLimitOrderingTest`.** Measured rather than estimated: the three methods together run in **1.5 s** wall clock (0.28-0.70 s each), against 0.96 s for the sibling `Issue6048SupernodeFullWalkDegradeTest`'s six methods, which builds the same fixture at 2,000 edges and is untagged. `CLAUDE.md` scopes `@Tag("slow")` to "noticeably long ... multi-second elapsed time", and the tags are a partition, so tagging this would move it out of the default unit lane for no gain. The transaction-per-edge shape is deliberate: it is what drives promotion and gives each edge a distinct recency rank, which is the whole fixture.

Re-ran after the change: `InterleavedIteratorTest`, `Issue6064BoundedLimitOrderingTest`, `Issue6048SupernodeFullWalkDegradeTest`, `Issue6044StripeIterationOrderTest`, `Issue6049GAVSupernodeOrderTest` - 35 tests, 0 failures. Full engine unit lane on the same code: 12,094 tests, 0 failures.

robfrank:
Items 1, 2 and 4 are the previous round's, already applied (`54aa18d`) or answered with a measurement. Item 3 is new, so I checked it rather than accepting it - **the fixture is deterministic, so there is no seed for a "bad run" of the hash to come from.**

Three independent inputs, none of them random:

- `StripeDirectory.stripeOf` is a pure function of the neighbour RID (a murmur3-style finaliser over position and bucket id), pinned to placement version 0 because entries are located by re-hashing.
- The RIDs are assigned by the default `RoundRobinBucketSelectionStrategy`, whose synchronous path is a plain `++current` counter starting at -1, not a hash of anything and not thread-dependent. The test's insert loop is single-threaded.
- `TestHelper` builds a fresh database per test method, so the counter starts from the same place every run.

Same insertion order gives the same RIDs, which give the same stripe assignment, which gives the same walk - the assertion is over one fixed permutation, not a sample. Ran the class three times to confirm: 3/3 green, 1.57 s / 1.34 s / 1.24 s (which also re-answers item 2 - the variation is JIT warm-up, not fixture cost).

The leeway is still deliberately generous, but for a different reason than flake tolerance: the measured worst position/rank ratio on this fixture is **2.4** (rank 190 at position 463) against an asserted bound of `3 x rank + 96`, so the assertion has roughly 40% headroom over the behaviour it pins while still being 24x tighter than the pre-fix ratio of 58. If it ever does go red without a code change, the cause is a change in RID assignment or in `stripeOf`, both of which are things a test should notice.

No code change this round.

PR #6276, opened 2026-08-16
fix(#6062): CHECK DATABASE remembers the adjacency list it walked instead of re-walking it per edge

### LINKED ISSUE #6062: CHECK DATABASE: unconditional O(degree) back-reference probes make it impractical on large super-node graphs
### Summary

`CHECK DATABASE` (with or without `FIX`) is entirely single-threaded and, for every edge, performs
multiple unconditional O(degree) linear-chain back-reference probes. On a graph with large super-nodes
this degenerates toward O(degree²) work concentrated on hub vertices, and on a dataset larger than
available page cache each probe risks a real disk read. On a several-hundred-GB graph database with
super-node hubs in the hundreds-of-thousands-of-edges range, a single `CHECK DATABASE FIX` run has been
measured taking **on the order of 80 hours**.

### Where the cost comes from

Two independent O(degree) mechanisms overlap and are both paid, unconditionally, per edge:

1. `GraphDatabaseChecker.checkEndpoints` (`checkEdges` arm) calls `EdgeLinkedList.containsEdge(edgeRID)`
   against both the in-vertex's IN list and the out-vertex's OUT list, for **every edge**, not only on a
   mismatch. `containsEdge` (`EdgeLinkedList`, and `StripedEdgeList.containsEdge` for promoted
   super-nodes) is a plain linear walk of the segment/stripe chain — no index, no hash lookup.
2. `checkVertices`' `checkOutgoingEdges`/`checkIncomingEdges` separately calls `isConnectedTo()` →
   `GraphEngine.isVertexConnectedTo()` → `EdgeLinkedList.containsVertex()` — again O(degree) — for every
   bidirectional edge, from both the OUT-list check and the IN-list check.

A bidirectional edge can therefore pay up to 4 separate O(degree) linear-chain walks across the two
passes, each re-reading pages the sibling edges of that same vertex already touched moments earlier.
None of this is touched by #6044/#6056 (super-node interleaving/locality), which only changed the
ordinary traversal iterators (`edgeIterator`/`vertexIterator`/`ridIterator`), not `containsEdge`/
`containsVertex`.

On top of the core probe cost, a full-scope `FIX` run adds:

- `reclaimOrphanedEdgeSegments` (always-on) does an independent full walk of every vertex's OUT/IN
  chains, separate from `checkVertices`'s own walk.
- `reconnectEdges` does a full re-scan of an edge type whenever any vertex on it needs its adjacency
  rebuilt.
- Corrupted-bucket index rebuilds run strictly serially, one full bucket scan at a time.

And the whole thing is single-threaded end to end — `DatabaseChecker`/`GraphDatabaseChecker` have no
`ExecutorService`/parallel stream anywhere, and there's no config knob to parallelize any of it.

### Impact

On a small-to-medium graph this is invisible. On a graph with real super-nodes (degree in the tens or
hundreds of thousands) and a working set that exceeds page cache, `CHECK DATABASE` — a maintenance
operation operators are told to run after any suspected corruption — becomes effectively unusable: an
80-hour run does not fit in any reasonable maintenance window, and a corruption incident that CHECK
DATABASE is supposed to help diagnose/repair instead becomes something operators can't act on in
practice.

### Suggested directions (not prescriptive)

- Replace the O(degree) `containsEdge`/`containsVertex` back-reference probes with something sub-linear
  (e.g. checked against a bloom/hash structure, or skip the probe on the direction that the forward walk
  already round-tripped consistently instead of re-deriving it independently).
- Parallelize the top-level checker phases (per-type / per-bucket) the way other maintenance operations
  in the engine already do, per the dedicated-pool conventions in `QueryEngineManager`.
- Consider a mode that trades completeness for time on very large graphs (e.g. sampling, or a
  `--skip-back-reference-probe` escape hatch) so operators have *some* option short of a multi-day run.

### Environment

Single database, no HA cluster. ~59M active records (~11.8M vertices / ~23.4M edges), ~657 GB on disk
(69,608 pages × 9,437,056-byte page size), 7,556 multi-page records, multiple super-node hubs with
degree in the hundreds of thousands. `CHECK DATABASE FIX` measured at 80h19m29s wall clock for a single
run.

robfrank:
Closes #6062

## The cost, and where it came from

`CHECK DATABASE` asks two back-reference questions of a **neighbour's** adjacency list, both once per edge and both an O(degree) linear walk of the chunk chain:

- `checkEdges` asks both endpoints of every edge record whether their list names it (`EdgeLinkedList.containsEdge`);
- `checkVertices` asks the far vertex of every adjacency entry whether it points back (`Vertex.isConnectedTo` -> `EdgeLinkedList.containsVertex`).

Nothing was remembered between two probes of the **same** list, so a hub of degree D had its list walked D times by each of them. That is O(D²) on exactly the vertex the check exists to survive, and once the working set exceeds the page cache every one of those walks risks a real disk read. The issue reports a 657GB database with hubs in the hundreds of thousands of edges measured at **80h19m** for a single `CHECK DATABASE FIX`.

## The change

A per-pass `AdjacencyProbeCache` (`engine/src/main/java/com/arcadedb/graph/AdjacencyProbeCache.java`) materialises the first probe of a list into primitive `LongHashSet`s keyed by bucket, and answers every later probe of that list with a hash lookup. The probe cost of a pass goes from `sum(degree²)` to `sum(degree)`.

Properties that are load-bearing rather than incidental:

- **Exact, never approximate.** A bloom/sketch answer would be worse than useless: a false positive reads as "the back-reference is there" and silently drops a real finding, and under `FIX DELETE ORPHANS` the same answer decides whether a record is deleted. Anything that cannot be represented exactly - an entry the primitive sets cannot hold, a list larger than the whole budget - falls back to the original `EdgeLinkedList` walk, and that verdict is remembered so such a list is never re-walked-and-abandoned once per probe.
- **Bounded**, by `arcadedb.checkDatabaseAdjacencyCacheEntries` (new, default 1,000,000 adjacency entries, roughly 20MB). Eviction is LRU by *last probe*, so a hub - the list that pays for itself - is the one that stays. `0` disables the cache and restores the pre-fix probe exactly.
- **The striped super-node keeps its own selection semantics.** A neighbour-keyed probe of a `StripedEdgeList` reads only the stripes that can hold that neighbour and is STRICT: a stripe head it cannot resolve becomes a retryable `ConcurrentModificationException` rather than a chain silently missing from the answer (`chainsForNeighbour`, #5680). An image built from the whole list would skip that chain and answer "not connected" where the engine promises to fail loudly - which is exactly what `SuperNodeStripingTest.transientlyMissingStripeChainIsRetryableForWritesBestEffortForReads` caught on the first iteration of this change. So the neighbour probe memoises **per chain**, keyed by the chain's head segment, and leaves the selection to the new `EdgeLinkedList.chainsForNeighbourProbe` / its `StripedEdgeList` override. The edge probe keeps a whole-list image because `containsEdge` and `entryIterator` already walk the identical set of chains (`allChains(false)`).
- **Staleness is handled by clearing, not by patching.** `EdgeVertexIterator.remove()` deletes the edge record, and `GraphEngine.deleteEdge` takes the edge out of BOTH endpoints' lists, so a prune in the vertex pass can change a list the cache holds an image of - for a vertex the walk has not reached, and through an endpoint pair a corrupt entry may misreport. Naming the affected lists would mean trusting the very pointers the prune is happening because of, so the whole cache is dropped at both prune sites and before the post-scan repair loops. Prunes are repairs on a damaged database; the worst case (damage that is overwhelmingly dangling entries) is no slower than it was before this cache existed.
- The polymorphic edge-type bucket filter that `isVertexConnectedTo` rebuilt through a stream **on every call** is memoised too - a hub probed once per incident edge was paying an array allocation per edge on top of the walk.

## What the check now reports

Three counters, seeded in `DatabaseChecker` so a clean run publishes them, summed across passes by the existing `updateStats` fold:

| key | meaning |
|---|---|
| `adjacencyProbes` | back-reference questions asked (structurally two per edge, per pass) |
| `adjacencyProbeListWalks` | how many of them had to touch an adjacency list; used to be all of them |
| `adjacencyEntriesScanned` | entries those walks visited - the number that was quadratic in a hub's degree |

`adjacencyProbeListWalks` close to `adjacencyProbes` on a graph with hubs now means "the cache is not engaging", which is the diagnostic the 80-hour report had no way to produce.

## Measured by the tests, not by a stopwatch

Every assertion is on those counters; nothing here times anything.

- Two vertices joined by 400 parallel edges (exactly two adjacency lists in the database): **1600 probes, 4 list walks**. With `arcadedb.checkDatabaseAdjacencyCacheEntries=0` the same run is 1600 probes / **1600** walks - that control test is what proves the counter measures the thing and the cache is what changes it.
- One hub with 400 distinct spokes: `adjacencyEntriesScanned` = **1600**, i.e. each list materialised once per pass. Per-probe walking would put it near 160,000 on the same fixture.

## Test plan

- [ ] `mvn -o -pl engine -am test -Dtest=Issue6062AdjacencyProbeCacheTest` - 6 tests: the walk counter on a two-vertex/many-parallel-edges graph, the cache-disabled control, the linear entry count on a hub with distinct spokes, a **promoted** (striped) super-node probed through its own stripe selection, cached-vs-direct agreement on a half-linked edge, and a healthy graph reporting nothing.
- [ ] `mvn -o -pl engine -am test -Dtest='CheckDatabase*,GraphDatabaseChecker*,Issue6090*,Issue687*,Issue5147*,SuperNode*Test,Issue6048*,Issue6062*,Issue5667*' -DexcludedGroups=benchmark,vector` - 139 tests, green.
- [ ] `mvn -o -pl engine -am test -Dtest='com.arcadedb.graph.**.*Test,com.arcadedb.database.**.*Test' -DexcludedGroups=benchmark,vector,slow` - 715 tests, green.
- [ ] Operationally: run `CHECK DATABASE` on a graph with a super-node and read `adjacencyProbes` against `adjacencyProbeListWalks`; set `arcadedb.checkDatabaseAdjacencyCacheEntries=0` and confirm the two converge and the findings are unchanged.

## Two things this touches beyond the cache, both deliberate

- **`EdgeVertexIterator` now ends the walk on a chunk pointing at itself** (issue #6278 tracks lifting the same guard into the sibling iterators). This is not cache plumbing: `entryIterator()` is what the checker walks a vertex's own list with, so before this a self-referencing chunk made `CHECK DATABASE` never return rather than report the damage. Measured: with the guard forced off, `aSelfReferencingChunkEndsTheWalkInsteadOfFeedingItForever` fails in 0.152s against its hard iteration cap.
- **The edge probe now resolves the endpoint through `GraphEngine.getMostUpdatedVertex` before reading its head chunk.** `checkEdges` called `getEdgeHeadChunk` on its own handle directly, which is the one read of an edge-list head in the engine that skipped the resolution that method's own Javadoc says every such read has to go through; the neighbour probe already had it via `ImmutableVertex.isConnectedTo`. No behaviour changes today - `checkEdges` writes nothing during its scan, so the transaction cache holds nothing that could differ from the caller's handle - but the two probes now read a head the same way, which is what makes "the cached and the direct answer are the same answer" true by construction rather than by coincidence.

## Deliberately NOT in this PR

The issue lists three directions; this one takes the first and leaves the other two, each of which is a separate change with its own risk:

- **Parallelising the checker phases.** `DatabaseChecker`/`GraphDatabaseChecker` are single-threaded end to end and say so on their own fields; making them parallel means picking a dedicated pool per the `QueryEngineManager` convention, making `CheckReport` and the corrupted/warning sets thread-safe, and re-deciding what a per-type transaction means. Worth doing, not worth folding into a change whose whole argument is that the answers are identical.
- **The second full vertex scan in `reclaimOrphanedEdgeSegments`**, and the full edge-type re-scan in `reconnectEdges`. Both are already named as known limitations in the existing Javadoc; neither is quadratic.
- **A completeness-for-time mode** (sampling, a `--skip-back-reference-probe` escape hatch). `arcadedb.checkDatabaseAdjacencyCacheEntries=0` is an escape hatch in the opposite direction (memory, not time); a mode that skips the probe outright changes what a clean result certifies and should be its own decision.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -468,6 +468,26 @@ public enum GlobalConfiguration {
       spanning several pages). Leave headroom when picking a value close to the replicated-entry limit.""",
       Integer.class, 256),
 
+  CHECK_DATABASE_ADJACENCY_CACHE_ENTRIES("arcadedb.checkDatabaseAdjacencyCacheEntries", SCOPE.DATABASE,
+      """
+      Maximum number of adjacency-list entries CHECK DATABASE keeps materialised while answering its back-reference \
+      probes (issue #6062). The check asks, once per edge, whether the NEIGHBOUR's edge list names that edge \
+      (checkEdges) and whether the far vertex of every adjacency entry points back (checkVertices). Both questions \
+      are a linear walk of the neighbour's chunk chain, and nothing used to be remembered between two probes of the \
+      same list, so a hub of degree D was walked D times by each of them: O(D²) on exactly the vertex the check \
+      exists to survive, with every walk risking a page read once the graph exceeds the page cache. A 657GB database \
+      with hubs in the hundreds of thousands of edges measured 80 hours for one CHECK DATABASE FIX. With the cache, \
+      the first probe of a list materialises it into primitive hash sets and every later probe of it is a hash \
+      lookup, so a pass walks each list once instead of once per incident edge. Counted in ENTRIES rather than in \
+      lists because that is what the memory is proportional to and because one super-node can be larger than every \
+      other list in the database put together; the least-recently-probed list is evicted when the budget is \
+      exceeded, so hubs - the lists that pay for themselves - are the ones that stay. A single list larger than the \
+      whole budget is never cached and is answered by the original walk. At the default of 1 million entries the \
+      footprint is roughly 20MB. 0 disables the cache and restores the pre-#6062 probe, which is the escape hatch \
+      if the memory is unwelcome on a small heap; the check reports what the setting bought under the \
+      adjacencyProbes / adjacencyProbeListWalks / adjacencyEntriesScanned keys.""",
+      Integer.class, 1_000_000),
+
   PAGE_FLUSH_QUEUE("arcadedb.pageFlushQueue", SCOPE.DATABASE, "Size of the asynchronous page flush queue", Integer.class, 512),
 
   FLUSH_SUSPEND_MAX_DEFERRED_RAM("arcadedb.flushSuspendMaxDeferredRAM", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -170,6 +170,16 @@ public class DatabaseChecker {
     result.put("edgesMissingOutReference", 0L);
     result.put("edgesMissingInReference", 0L);
     result.put("unreachableEdgeRecordsFound", new LinkedHashSet<RID>());
+    // Issue #6062: what the back-reference probes cost, seeded so a run always answers "why was this slow" rather
+    // than answering it only when a graph pass happened to run. adjacencyProbes is how many times the check asked an
+    // endpoint whether its adjacency list references an edge - structurally two per edge, per pass;
+    // adjacencyProbeListWalks is how many of those had to walk a list, which used to be all of them and now grows
+    // with the number of DISTINCT lists; adjacencyEntriesScanned is the entries those walks visited, the number that
+    // was quadratic in a super-node's degree. A adjacencyProbeListWalks close to adjacencyProbes on a graph with
+    // hubs means the cache is not engaging - see arcadedb.checkDatabaseAdjacencyCacheEntries.
+    result.put("adjacencyProbes", 0L);
+    result.put("adjacencyProbeListWalks", 0L);
+    result.put("adjacencyEntriesScanned", 0L);
     result.put("warnings", new LinkedHashSet<>());
     // Issue #6143: files this node holds that no schema component claims. Report-only, always present so a clean
     // run says "none" rather than saying nothing, and empty under a RECORD scope, which cannot answer the question.
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -173,10 +173,11 @@ public class DatabaseChecker {
     // Issue #6062: what the back-reference probes cost, seeded so a run always answers "why was this slow" rather
     // than answering it only when a graph pass happened to run. adjacencyProbes is how many times the check asked an
     // endpoint whether its adjacency list references an edge - structurally two per edge, per pass;
-    // adjacencyProbeListWalks is how many of those had to walk a list, which used to be all of them and now grows
-    // with the number of DISTINCT lists; adjacencyEntriesScanned is the entries those walks visited, the number that
-    // was quadratic in a super-node's degree. A adjacencyProbeListWalks close to adjacencyProbes on a graph with
-    // hubs means the cache is not engaging - see arcadedb.checkDatabaseAdjacencyCacheEntries.
+    // adjacencyProbeListWalks is how many walks of a list answering them took, which used to be one per probe and
+    // now grows with the number of DISTINCT lists; adjacencyEntriesScanned is the entries those walks visited, the
+    // number that was quadratic in a super-node's degree. adjacencyProbeListWalks at or above adjacencyProbes on a
+    // graph with hubs means the cache is not engaging - the budget may be below the hub's degree, see
+    // arcadedb.checkDatabaseAdjacencyCacheEntries.
     result.put("adjacencyProbes", 0L);
     result.put("adjacencyProbeListWalks", 0L);
     result.put("adjacencyEntriesScanned", 0L);
--- a/engine/src/main/java/com/arcadedb/graph/AdjacencyProbeCache.java
+++ b/engine/src/main/java/com/arcadedb/graph/AdjacencyProbeCache.java
@@ -70,6 +70,13 @@ import java.util.Map;
  * what this class materialises through - and which the checker's own walk of a vertex's list goes through as well,
  * where the missing guard meant a check that never returned.
  * <p>
+ * SCOPED TO ONE PASS OVER ONE TYPE, not to the whole run. {@code checkVertices}/{@code checkEdges} each build their
+ * own, and {@code DatabaseChecker} calls them once per vertex/edge type, so a hub referenced from two source types
+ * has its list materialised once per type pass rather than once per {@code CHECK DATABASE}. Deliberate: the images
+ * are only valid while nothing writes to the lists, and each pass commits its own repairs, so a cache outliving a
+ * pass would have to survive exactly the writes it cannot survive. The cost is a constant factor in the number of
+ * types, not a return to the quadratic shape - each pass is still one walk per list.
+ * <p>
  * STALENESS IS THE CALLER'S JOB, and there is exactly one caller. A cached image is only valid while the list it was
  * built from does not change. The edge pass writes nothing while it scans; the vertex pass prunes dangling entries,
  * and calls {@link #clear()} at that site rather than naming the lists it touched - see that method. Both drop the
@@ -324,6 +331,12 @@ final class AdjacencyProbeCache {
   /**
    * The polymorphic bucket file-ids of an edge type, memoised. {@code isVertexConnectedTo} rebuilt this through a
    * stream per call, so a hub probed once per incident edge paid an array allocation per edge on top of the walk.
+   * <p>
+   * Held for the life of the cache, which is one pass over one type. A bucket added to an edge type WHILE that pass
+   * runs is therefore missed until the next pass, where the unmemoised form would have picked it up on the next
+   * probe. Left as is: a bucket added mid-scan holds no edges the scan has already read past either, so the filter
+   * is not the thing that would make such a run coherent - a schema change concurrent with a check is outside what
+   * either form promises.
    */
   private int[] bucketFilterOf(final VertexInternal vertex, final String edgeType) {
     return bucketFilters.computeIfAbsent(edgeType,
--- a/engine/src/main/java/com/arcadedb/graph/AdjacencyProbeCache.java
+++ b/engine/src/main/java/com/arcadedb/graph/AdjacencyProbeCache.java
@@ -1,0 +1,382 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.graph;
+
+import com.arcadedb.database.RID;
+import com.arcadedb.utility.LongHashSet;
+import com.arcadedb.utility.Pair;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Answers the back-reference probes {@link GraphDatabaseChecker} makes, remembering the adjacency lists it had to
+ * walk (issue #6062).
+ * <p>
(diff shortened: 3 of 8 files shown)


robfrank:
Thanks - all three notes verified against the code and addressed in f2340797fe.

**1. Sentinel entries were outside the bound - confirmed, real, fixed.** A remembered "this list cannot be represented" verdict carries `entryCount == 0`, so `cachedEntries` never saw it and `evictUntilWithinBudget` could never evict one. Worse than the review says: eviction was only *called* when a build succeeded, so a pass that produced nothing but sentinels never ran it at all. The cache is now bounded on two axes - the adjacency-entry budget and a new `MAX_CACHED_LISTS` (65,536) list-count ceiling - eviction tests both, and it now runs after storing a failure too. Left as a constant rather than a second knob: at that size the map costs single-digit MB whatever the entry budget is, which is small beside the budget's own footprint. The class Javadoc now states both bounds instead of only the first.

**2. The double count is real, and it is what happened, so it is now stated rather than removed.** Two physical walks do occur on the first probe of an uncacheable list - the materialisation attempt walks to the budget and abandons, then the fallback walks for the answer - and `adjacencyProbeListWalks` counts walks, not probes that needed one. Suppressing one would make the counter under-report exactly the work an operator is looking for. The contract is now spelled out on `getListWalks()` and in `DatabaseChecker`'s seeding comment, including that `adjacencyProbeListWalks >= adjacencyProbes` is possible and what it means (the budget is below the hub's degree).

**3. Coverage gap - half closed, half not constructible.** Added `aListLargerThanTheWholeBudgetFallsBackWithoutChangingAFinding`: budget 10 against a hub of degree 400, so the hub falls back while the spokes still cache, asserting every finding matches both the fully cached and the fully disabled run, plus `adjacencyEntriesScanned <= 5 * DEGREE`. That last bound is the one that matters for your point 2's path: without the remembered verdict the same fixture scans ~9,200 entries and the assertion fails, so it pins "abandoned once, not once per probe" rather than merely "small".

The corrupt-entry half of that fallback (a `null` or `Long.MIN_VALUE` RID in a segment) is deliberately not covered by a test: `MutableEdgeSegment.addAtEnd` always writes all four numbers, so no engine or public API can produce such an entry - reaching it would mean hand-writing a malformed segment buffer, which pins the malformed buffer rather than the behaviour. It is the same code path the new test exercises (`build` returns null -> verdict remembered -> original walk), and the guard stays because a set that silently could not hold a value would answer "not referenced" for an edge that is.

The two design points you flagged for other reviewers (per-chain memoisation to preserve `StripedEdgeList`'s strict stripe semantics, and the coarse `clear()` on prune) are unchanged and deliberate - the first is the one that actually went wrong on the first iteration of this change, caught by `SuperNodeStripingTest.transientlyMissingStripeChainIsRetryableForWritesBestEffortForReads`.

Verification: `Issue6062AdjacencyProbeCacheTest` 7/7, and 141 tests across `CheckDatabase*` / `GraphDatabaseChecker*` / `SuperNode*` / `Issue6090*` / `Issue6048*` / `Issue6044*` / `Issue687*` green.

robfrank:
All three addressed in 0174b1e9c5. The first one was a better finding than "low severity" - measured, it is not merely a cost multiplier.

**1. The unguarded `entryIterator()` walk - fixed at the iterator, and it turns out the pre-existing gap was the worse half.** You are right that the direct `containsEdge`/`containsVertex` walks guard the self-loop and `EdgeVertexIterator` did not. Measured rather than assumed: with a hub whose IN head chunk is pointed at itself, the walk through `entryIterator()` does not terminate at all, and the checker's walk of a vertex's OWN list already went through it - so `CHECK DATABASE` on that shape of corruption never returned, before this PR added a second consumer. `EdgeVertexIterator.hasNext()` now ends the walk when a chunk's `previous` is itself, the same break `EdgeLinkedList.containsEdge` has always had. That makes the checker terminate on a chain it previously hung on, and it removes the up-to-`maxEntries` burn you flagged in `build()` at the same time. The sibling iterators (`EdgeIterator`, `VertexIterator`, `RIDIterator`, `IteratorFilterBase`) still lack the guard and are named as such in the comment - not touched here, since nothing in this change consumes them.

Test: `aSelfReferencingChunkEndsTheWalkInsteadOfFeedingItForever`. Deliberately NOT a `CHECK DATABASE` under `@Timeout(SEPARATE_THREAD)` - an unterminating walk does not fail a test, it hangs the build, and a regression has to come back red. It walks the list under a hard cap of `8 * DEGREE` and asserts the count stayed at or below `DEGREE`. Control run with the guard forced off: **fails in 0.152s** with that assertion, rather than hanging.

**2. The `containsNeighbour(neighbour, null)` sweep - agreed, dead, removed.** Both call sites resolve a concrete edge type, and `bucketFilterOf` would throw on a null one anyway, so the branch could not be reached. The Javadoc now states the filter is always concrete and what an edge type with no buckets does (empty filter -> false, exactly as `containsVertex` answers with the same array), so the removal does not read as an unstated assumption.

**3. The `keep` guard in `evictUntilWithinBudget` - agreed, unreachable, removed.** The map is access-ordered and `put` counts as an access, so the entry just stored is the most recent and can only be the eldest when it is the only one left, which `size() > 1` already covers. Removed the parameter and put that reasoning in a comment where the guard used to be, since "why is it safe not to check" is the part a future reader needs.

Verification: `Issue6062AdjacencyProbeCacheTest` 8/8, and 717 tests across `com.arcadedb.graph.**` (OLAP included, since `CSRBuilder` is the other `entryIterator()` consumer) and `com.arcadedb.database.**` green.

robfrank:
All three observations addressed in 33fab56644. No code behaviour changed this cycle - two of them were true and undocumented, and the third is now a filed issue rather than a comment.

**1. Per-pass, not per-run - correct, deliberate, now stated.** The class Javadoc says so explicitly, including why it is not merely simplicity: the images are only valid while nothing writes to the lists, and each pass commits its own repairs, so a cache outliving a pass would have to survive exactly the writes it cannot survive. The cost is a constant factor in the number of types, not a return to the quadratic shape - each pass is still one walk per list, which is the claim the counters pin.

**2. The sibling iterators - filed as #6278.** You are right that this is the more interesting half: those back the ordinary traversal APIs, so the same self-referencing chunk hangs a normal query in a request thread, where the checker at least hangs a maintenance operation. The issue proposes lifting the guard into `ResettableIteratorBase`'s chunk hop rather than making a fifth copy - four copies drifting apart is the failure mode the guard's absence here already demonstrates - and asks the open question the fix has to answer: whether a self-loop should merely END the walk (what the direct probes and #6276 do) or be REPORTED, since ending it silently means a traversal returns a short answer over a damaged chain with nothing in the log. The `EdgeVertexIterator` comment now names #6278 so the next reader finds it. Not folded into this PR: nothing in #6062 consumes those iterators, and the reporting question is a behaviour decision that deserves its own review.

**3. `bucketFilterOf`'s window - narrower than it looks, and now written down.** The memo lives one pass, so a bucket added to an edge type mid-scan is missed until the next pass where the unmemoised form would have caught it on the next probe. Left as is, with the reasoning at the method: a bucket added mid-scan holds no edges the scan has already read past either, so the filter is not what would make such a run coherent - a schema change concurrent with a check is outside what either form promises. If that assumption should be enforced rather than documented, that is a change to `CHECK DATABASE`'s contract and wants its own issue.

Verification after the doc changes: 124 tests across `Issue6062AdjacencyProbeCacheTest` / `CheckDatabase*` / `GraphDatabaseChecker*` / `SuperNode*` / `Issue6044*` / `Issue6048*` green.

robfrank:
Nothing actionable in the code this round, so no new commit - the one item you raised was a gap in the PR description rather than in the change, and the description now carries it.

**The `getMostUpdatedVertex` call in `containsEdge` is deliberate, and you read it exactly right.** `checkEdges` was calling `getEdgeHeadChunk` on its own handle directly - the one read of an edge-list head in the engine that skipped the resolution that method's own Javadoc says every such read has to go through - while the neighbour probe already had it via `ImmutableVertex.isConnectedTo`. Nothing observable changes today: `checkEdges` writes nothing during its scan, so the transaction cache holds nothing that could differ from the caller's handle. It is in because the two probes now read a head the same way, which is what makes "the cached answer and the direct answer are the same answer" true by construction rather than by coincidence - and because a future `checkEdges` that did write would have had a silent divergence waiting for it. Written up under its own heading in the PR description now, alongside the `EdgeVertexIterator` guard, since both are scope beyond "add a cache" and both should be visible to whoever merges this.

**On the HA/concurrency gap you named:** agreed that it is the right boundary rather than a hole, and worth stating why it is safe to leave. The cache is a per-pass local, never a field, so nothing about it is shared between threads or between databases; `verboseLevel` only decides whether the accumulated warnings are logged, and it is read after every pass the cache lives in. Under replication a check runs on one node and repairs reach the others as WAL deltas - `checkedNodeScope` in the result already says so - and this change does not move where the check runs or what it writes, only how many times it reads the same list. The single-threaded contract is the checker's, not the cache's; parallelising it is listed as out of scope precisely because it would need `CheckReport` and the corrupted/warning sets to change too.

Thanks for the three passes - the striped-selection and sentinel-bounding findings were both real, and the first one was already caught red by `SuperNodeStripingTest` before it could reach anyone.

PR #6277, opened 2026-08-16
fix(#5719): a single HTTP response is bounded by a ceiling no caller can widen, and refuses rather than truncates

### LINKED ISSUE #5719: HTTP query endpoint: no hard ceiling on the rows a single response materializes
Follow-up to #5711 / #5716, filed on the code review's recommendation so the decision does not get lost.

## Context

#5716 made the HTTP query and command endpoints honor a limit the caller stated - the request's `limit` field, or the `LIMIT` the query carries - and cap only the callers that state none, with `arcadedb.server.httpQueryDefaultLimit` (default 20000).

The consequence is that the default cap now protects **only** queries that state no limit of their own. An authenticated caller who writes `SELECT FROM HugeType LIMIT 100000000` (or sends `"limit": -1`) makes the server serialize an arbitrarily large result into a single JSON response.

This is not a new capability: `"limit": -1` has always been an unbounded escape hatch for the same actor, so no privilege boundary moved. What changed is the *bar* - a large `LIMIT` in the query text is easier to reach accidentally (e.g. an application that builds `LIMIT n` from user input) than an explicit `limit: -1` in the request body. In a multi-tenant deployment the previous implicit 20,000 cap was doing real protective work.

## Proposal

Mirror what the gRPC path already does. `arcadedb.server.grpcQueryMaxResultRows` is documented as:

> Hard ceiling on the number of rows the gRPC unary ExecuteQuery materializes. A request limit at or below this cap is honored; a result that would exceed it fails the call with RESOURCE_EXHAUSTED (consistent with the StreamQuery MATERIALIZE_ALL path) rather than silently truncating, and a client cannot bypass it with a larger limit.

An HTTP equivalent - `arcadedb.server.httpQueryMaxResultRows` - would answer `413` (or `400`) naming the setting when the requested cap exceeds it, and would never truncate silently, which keeps the invariant #5711 established.

## The decision this needs

**Any finite default is a breaking change** for deployments that send `limit: -1` today and get everything. So the options are:

1. Default `-1` (disabled): fully backward compatible, but provides no protection out of the box - an operator has to know the setting exists.
2. A finite default (e.g. 1,000,000): protects by default, breaks existing unbounded callers at the next upgrade, and needs a release note under Breaking Changes.
3. Finite default only in the `production` server mode, unlimited in `development`.

Option 3 is arguably the best of the three but adds mode-dependent behaviour to a data path, which has its own cost.

## Notes

- The timeseries query endpoint materializes its whole range before applying any limit, so a ceiling there would have to bound the *fetch*, not just the serialization, to be worth anything.
- Whatever is chosen, the ceiling must fail loudly rather than truncate - a silently truncated response is the defect #5711 fixed.

robfrank:
Closes #5719

## The decision this needed

The issue asked for one of three options, and this PR takes **option 2: a finite default**, `arcadedb.server.httpQueryMaxResultRows = 1000000`, with `-1`/`0` to disable.

- **Option 1 (default `-1`, disabled)** was rejected because a ceiling an operator has to discover before it protects anything protects nobody, and it would leave HTTP as the only surface in the server without a finite resource ceiling: `httpBodyContentMaxSize` (100MB), `grpcQueryMaxResultRows` (100,000) and `grpcStreamMaxMaterializedRows` all default to finite values today.
- **Option 3 (finite only in `production` mode)** was rejected because `arcadedb.server.mode` currently governs only log verbosity and how much of an error body is exposed. Nothing on a data path reads it. Making a request succeed in `development` and fail in `production` would mean dev and prod disagree about whether a query is answerable, which is a worse failure mode than a documented breaking change.
- 1,000,000 sits 50x above the default page cap (`httpQueryDefaultLimit` = 20,000), so a realistic paging or export client never meets it. The breaking change is written up under its own heading in `docs/release-26.9.1.md`.

## What changed

`arcadedb.server.httpQueryDefaultLimit` caps only the callers that state no limit of their own - that is the invariant #5716 established, and it is the right one: a response silently cut below what the caller asked for is the defect #5711 fixed. The consequence, which this issue was filed for, is that `SELECT FROM HugeType LIMIT 100000000` or `"limit": -1` had no bound at all.

The new setting is the HTTP twin of `arcadedb.server.grpcQueryMaxResultRows` and behaves identically:

- a stated cap (request `limit`, or the query's own `LIMIT` read back from the execution plan, or the configured default) **at or below** the ceiling is the caller's own bound, honored silently, with `truncated` reported exactly as before;
- a larger one, or an explicitly unlimited `-1`/`0`, cannot widen the response past the ceiling;
- a result that **would** exceed the ceiling **fails with HTTP 413 naming the setting** rather than being truncated. Truncating instead would re-introduce, for the callers #5716 protects, the silent short answer #5711 removed.

The ceiling therefore never appears in a successful response: a 200 either carries everything within the stated cap or was cut by the stated cap, so `limit`/`returned`/`truncated` keep reporting exactly what they reported before and no existing client contract moves.

Endpoints covered: `POST /api/v1/query`, `POST /api/v1/command`, `GET /api/v1/query`, and `POST /api/v1/ts/{db}/query`.

### Two paths bounded at the fetch, not just at serialization

- `PostCommandHandler` pushes a `LIMIT` down into a SELECT/MATCH that states none. An unlimited request `limit` used to push nothing down at all, so the engine streamed the whole result to the handler; the pushdown cap is now clamped by the ceiling first.
- `profileExecution: detailed` drains the entire result set into memory (`materializeResultSet`) before anything is serialized, and a command carrying its own huge `LIMIT` is left as written and so gets no pushed-down bound. That drain now stops one row past the cap the response will honor - one row above, like the pushdown, so truncation stays detectable.

### Rollback, not commit-behind-an-error

The refusal is raised as a `ResultSetTooLargeException` out of the handler rather than returned as a 413 `ExecutionResponse`. That is deliberate: for an auto-committed write such as `UPDATE ... RETURN AFTER`, returning a status would commit a write whose result the caller never sees, and a client retry would then double-apply it. Raising rolls the auto-commit transaction back, so the retry starts from committed state. `aRefusedCommandIsRolledBackInsteadOfCommittedBehindTheError` pins this.

### The error body in production mode

`detail` is concealed when `arcadedb.server.mode=production`, so the setting name goes in the top-level `error` label and the ceiling value in `exceptionArgs` - both of which are emitted in every mode. A caller that cannot see which knob refused it, or what number to stay under, has been told nothing actionable.

## Known limitation (called out in the issue, and not closed here)

The TimeSeries query endpoint enforces the ceiling over the response it builds, but `TimeSeriesEngine.query()` materializes the whole requested range into a `List<Object[]>` before any limit is known. The check therefore keeps the second and larger copy (the JSON) out of the heap but cannot keep the first one out. Bounding that fetch is an engine-level change and is left as follow-up; it is documented in the release note and in a comment at the check.

Also not changed: `RemoteHttpComponent.manageException` has no typed arm for the new exception, so the Java remote driver surfaces a 413 as a `RemoteException` naming `com.arcadedb.server.http.ResultSetTooLargeException` (it is not retried, which is the important part). A typed arm can be added if a driver-side `catch` turns out to be wanted.

## Test plan

- [ ] `mvn -pl server -am -DskipITs=false -Dit.test=HttpQueryMaxResultRowsIT verify` - 14 new integration tests: the query `LIMIT` above the ceiling, `limit: -1`, the GET endpoint, the Cypher path (no plan limit, no pushdown), the `graph` serializer, the `detailed` profile path, the TimeSeries endpoint, the write rollback, the exactly-at-the-ceiling boundary, and the disabled ceiling.
- [ ] `mvn -pl server -am -Dtest=MaxResultRowsClampTest test` - 5 unit tests over the clamp, including both disabled values and both unlimited values.
- [ ] **Control run**: with the ceiling forced to `-1` in the fixture, exactly the 8 refusal tests fail and the 6 unchanged-behavior tests still pass, so the new tests are proven to discriminate.
- [ ] `mvn -pl server -am -DskipITs=false -Dit.test=HttpQueryTruncationIT,TimeSeriesQueryHandlerIT,OpenApiSpecGenerationIT,OpenApiDocsEndpointIT,Issue5037ProductionErrorBodyIT verify` - the #5711/#5716 regression suites are untouched and green.
- [ ] `mvn -pl server -am -DexcludedGroups=benchmark,slow,vector test` - engine (12,086), network and server (788) unit lanes green. One network failure appeared in the parallel full-reactor run, `RemoteHttpComponentTest.httpCommandMalformedJsonResponseThrowsLabelledRemoteException` (its one-shot local HTTP server answered 404, i.e. the port was taken); it passes on its own, and only a Javadoc comment in that file changed.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2283,8 +2283,16 @@ Two paths that used to escape any bound are bounded at the fetch rather than at 
 POST endpoints push down into a command that states none is now capped by the ceiling (an unlimited `limit`
 used to push nothing down at all), and `profileExecution: detailed`, which drains the whole result set into
 memory before serializing it, drains at most one row past the cap the response will honor. The TimeSeries
-query endpoint enforces the ceiling too, but only over the response it builds: `TimeSeriesEngine.query()`
-materializes the whole requested range before any limit is known, so bounding that fetch needs an engine-level
-change and is left as follow-up work.
+query endpoint enforces the ceiling on both of its shapes - the raw rows, and the buckets of an `aggregation`
+request, which reads no `limit` at all and so had the ceiling as its only possible bound - but only over the
+response it builds: `TimeSeriesEngine.query()` materializes the whole requested range before any limit is
+known, so bounding that fetch needs an engine-level change and is left as follow-up work.
+
+On the status code: 413 is conventionally about an oversized *request* body, and no standard code says "the
+response you asked for is too large to send". 413 is the closest fit and keeps the refusal in the 4xx range,
+where it belongs - the request is answerable, just not as written - but a client that maps status codes
+mechanically should note that the fix is to narrow or page the query, not to shrink the request body. The
+error body distinguishes the two: this refusal names `arcadedb.server.httpQueryMaxResultRows` in its `error`
+field and carries the ceiling in `exceptionArgs`, in every server mode.
 
 [#5719](https://github.com/ArcadeData/arcadedb/issues/5719)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2265,6 +2265,20 @@ A finite default was chosen over a disabled one because every other resource cei
 a ceiling an operator has to discover before it protects anything protects nobody. At 1,000,000 it sits 50x
 above the default page cap, so a realistic paging or export client never meets it.
 
+The two settings cannot be configured into disagreeing. `httpQueryDefaultLimit` is itself lowered to the
+ceiling when it is configured above it (or left unlimited while the ceiling is not), so a caller that states
+no limit at all is never refused - it gets the ordinary reported truncation it has always got, at the lower of
+the two values. The refusal is for a caller that asked to go past the ceiling, never for one served by the
+default.
+
+Two things the ceiling deliberately does not bound, both worth knowing before sizing it. It bounds the size of
+a response, not the peak cost of refusing one: the rows are serialized up to the ceiling before the result
+turns out to exceed it, and the work is then thrown away, so a client that routinely brushes the ceiling
+produces real garbage rather than merely a smaller payload - the same trade the gRPC path makes, and the price
+of never truncating silently. And it bounds the response, not the work a command does to produce it: a write
+that returns its rows (`UPDATE ... RETURN AFTER`) applies to every matching record before the response is
+built, and is then rolled back with the refusal, so the ceiling is not a guard against write amplification.
+
 Two paths that used to escape any bound are bounded at the fetch rather than at serialization: the `LIMIT` the
 POST endpoints push down into a command that states none is now capped by the ceiling (an unlimited `limit`
 used to push nothing down at all), and `profileExecution: detailed`, which drains the whole result set into
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2234,3 +2234,43 @@ purpose ("more seeds than nodes" reads as "as many as exist" and is clamped to t
 nameless `NegativeArraySizeException`.
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216)
+
+## The HTTP query endpoints now have a hard row ceiling a caller cannot widen (#5719)
+
+**Breaking change.** `arcadedb.server.httpQueryMaxResultRows` is a new server setting, defaulting to
+**1,000,000**, that bounds the number of rows a single HTTP query or command response materializes. A
+deployment that relies on an HTTP caller pulling an unbounded result in one response - typically by sending
+`"limit": -1` - has to raise it or set it to `-1` to keep that working.
+
+The ceiling closes the gap that #5716 left open. `arcadedb.server.httpQueryDefaultLimit` (default 20,000)
+caps only the callers that state no limit of their own: since #5716 a request carrying a `limit` field, and a
+query carrying its own `LIMIT`, are both honored as written, because a response silently cut below what the
+caller asked for is exactly the defect #5711 fixed. The consequence is that an authenticated caller writing
+`SELECT FROM HugeType LIMIT 100000000` - or an application that builds `LIMIT n` from user input - could make
+the server serialize an arbitrarily large result into one JSON document. No privilege boundary moved (`limit:
+-1` was always available to the same actor), but the bar to reaching it accidentally dropped, and in a
+multi-tenant deployment the implicit 20,000 cap had been doing real protective work.
+
+The new setting is the HTTP twin of `arcadedb.server.grpcQueryMaxResultRows` and behaves the same way. A
+stated cap at or below the ceiling is the caller's own bound and is honored silently, truncation reported with
+`truncated` exactly as before. A larger one - or an unlimited `-1`/`0` - cannot widen the response past the
+ceiling, and a result that would exceed it **fails with HTTP 413 naming the setting** instead of being
+truncated: a truncated response indistinguishable from a complete one is what #5711 removed, and
+re-introducing it for the callers #5716 protects would only have moved it. The refusal is raised from inside
+the handler, so an auto-committed write command whose result nobody will ever see is rolled back rather than
+committed behind an error status.
+
+A finite default was chosen over a disabled one because every other resource ceiling the server ships -
+`httpBodyContentMaxSize`, `grpcQueryMaxResultRows`, `grpcStreamMaxMaterializedRows` - is finite by default, and
+a ceiling an operator has to discover before it protects anything protects nobody. At 1,000,000 it sits 50x
+above the default page cap, so a realistic paging or export client never meets it.
+
+Two paths that used to escape any bound are bounded at the fetch rather than at serialization: the `LIMIT` the
+POST endpoints push down into a command that states none is now capped by the ceiling (an unlimited `limit`
+used to push nothing down at all), and `profileExecution: detailed`, which drains the whole result set into
+memory before serializing it, drains at most one row past the cap the response will honor. The TimeSeries
+query endpoint enforces the ceiling too, but only over the response it builds: `TimeSeriesEngine.query()`
+materializes the whole requested range before any limit is known, so bounding that fetch needs an engine-level
+change and is left as follow-up work.
+
+[#5719](https://github.com/ArcadeData/arcadedb/issues/5719)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1207,7 +1207,9 @@ public enum GlobalConfiguration {
       its own LIMIT clause, are both honored as written and are never capped by this value. When this default \
       does cut a result short the response reports `"truncated": true` next to `returned` and `limit`, and the \
       server logs a warning: the truncation is never silent (issue #5711). Set to -1 or 0 for unlimited \
-      (WARNING: removes the protection against materializing an unbounded result set in memory). Default is 20000""",
+      (WARNING: removes the protection against materializing an unbounded result set in memory). A value above \
+      'arcadedb.server.httpQueryMaxResultRows' - including unlimited - is lowered to it, so the two settings \
+      cannot disagree and a caller that states no limit is never refused, only truncated. Default is 20000""",
       Integer.class, 20_000),
 
   SERVER_HTTP_QUERY_MAX_RESULT_ROWS("arcadedb.server.httpQueryMaxResultRows", SCOPE.SERVER,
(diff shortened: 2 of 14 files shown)


robfrank:
Thanks - point (1) was a real bug, and the other three were fair descriptions of what the ceiling does and does not promise. Addressed in f1bdcd6853.

**(1) Ceiling below the default cap: fixed, not documented around.** You traced it correctly: with `httpQueryMaxResultRows < httpQueryDefaultLimit`, `resolveLimit` returned the too-high default and the clamp then tripped `limit != statedLimit`, so a caller that stated *nothing* got a 413. A warning would have left the broken behavior in place, so the two settings are now made unable to disagree instead: `getDefaultRowLimit()` is itself bounded by the ceiling, so a no-limit caller's `statedLimit` can never exceed it and the refusal is structurally unreachable for that caller. What it gets is the ordinary truncation at the lower of the two values, reported with `truncated` and `limit` as always. The refusal is for a caller that asked to go past the ceiling; it was never meant for one served by the default.

New test `aCeilingBelowTheDefaultCapTruncatesInsteadOfRefusing` pins all three halves: the POST and GET endpoints truncate at the lowered value, and a caller that *does* state `limit: -1` is still refused - so the clamp did not quietly turn the ceiling off for everybody.

**(2) Graph/`studio` element count.** Correct, and inherited. The row that crosses the cap is expanded whole, so for those two serializers the element count can exceed the ceiling by one row's expansion. I have not changed the serializer - doing so would move #5711 behavior for every caller, ceiling or not - but the imprecision is now stated where the bound is claimed, in `serializeResultSetBounded`'s javadoc: the ceiling bounds the *rows* a response materializes exactly, and the element count within one row's expansion.

**(3) A refused response pays most of the cost.** Also correct, also deliberate, and it is the same trade the gRPC path makes: never truncating silently is what forces the serialization to happen before the verdict. Now stated in the setting's own description and in the release note - "this bounds the SIZE of a response, not the peak cost of refusing one" - so an operator sizing the ceiling reads it next to the number.

**(4) Writes with `RETURN` are unbounded at the write.** Correct: pushdown covers `SELECT`/`MATCH` only, so `UPDATE ... RETURN AFTER` applies to every matching record and is then rolled back with the refusal. That is the honest guarantee and it is now written down in both the setting description and the release note, explicitly: the ceiling is not a guard against write amplification.

**Test gaps.** Both closed: the ceiling-below-default interaction above, and a new `ResultSetTooLargeErrorBodyTest` that drives the refusal through `buildErrorBody` in production mode and asserts `detail` is concealed while the setting name (in `error`) and the ceiling (in `exceptionArgs`) survive - the promise the exception's javadoc makes.

**Deferred:** the `RemoteHttpComponent.manageException` typed arm. Left out of this PR on purpose - a 413 already surfaces as a non-retried `RemoteException` naming `ResultSetTooLargeException` - and left to the maintainer to file, rather than opening an issue unprompted from this branch.

Verification after the changes: `HttpQueryMaxResultRowsIT` 15/15, server unit lane 791/791, and `HttpQueryTruncationIT` + `TimeSeriesQueryHandlerIT` + `OpenApiSpecGenerationIT` + `Issue5037ProductionErrorBodyIT` 50/50 green.

robfrank:
The aggregation gap was real - thanks for tracing the endpoint's second shape. All four points addressed in 7bcc3d7c91.

**TimeSeries aggregation branch: fixed.** You are right that it was the same class of problem on an endpoint the PR claimed to cover, and worse than the raw branch in one respect: `executeAggregation` reads no `limit` at all, so the ceiling was not merely absent, it was the only bound the path could ever have had. `timestamps.size()` is now checked against `getMaxResultRows()` right after `aggregateMulti` returns and before the `buckets` array is built. `MAX_FLAT_BUCKETS` is not a substitute, as you say: it only chooses between the flat-array and map representations.

New test `theTimeSeriesAggregationBranchIsBoundedToo`: 30 samples, `bucketInterval` of one sample each so the bucket count exceeds the ceiling, expects 413 naming the setting; then the same query with a wide enough interval, expecting 200. **Control run**: with the new check disabled, exactly that one test fails (16 run, 1 failure), so it discriminates on the new code and nothing else.

**413 semantics**: noted in the release notes, in the terms you put it - 413 is the closest fit and keeps the refusal in 4xx where it belongs, but a client mapping codes mechanically should know the fix is to narrow or page the query, not to shrink the request body, and the error body distinguishes the two by naming the setting in `error` and carrying the ceiling in `exceptionArgs`.

**OpenAPI, TimeSeries endpoint**: `413` added to `createQueryPath()`. Added explicitly rather than through `SpecBuilders.standardResponses`, whose canned 413 text reads "Request body too large" - which is exactly the confusion the point above is about. The description names both shapes, rows and buckets.

**Double clamp**: removed rather than commented. Now `requestLimit != null ? applyMaxResultRows(requestLimit, maxResultRows) : getDefaultRowLimit()`, so only the caller's own value is clamped there and the default arrives already bounded. Nothing downstream depended on the second clamp.

Verification: `HttpQueryMaxResultRowsIT` 16/16, server unit lane 791/791, `TimeSeriesQueryHandlerIT` + `OpenApiSpecGenerationIT` + `OpenApiDocsEndpointIT` green.

Still deferred, unchanged from the previous round: the `RemoteHttpComponent.manageException` typed arm for `ResultSetTooLargeException`, left for the maintainer to file rather than opening an issue unprompted from this branch.

robfrank:
Confirmed and fixed in 4dbd2c2a70. You were right on both halves, including the part I had got wrong in my own javadoc: the overshoot was not "a few elements", it was unbounded, and it came back 200 with `truncated: false`.

**The cap is now tested inside the expansion, not only around it.** `analyzePropertyValue`'s `Collection` branch checks `vertices.length() + edges.length() >= limit` before each element, so `collect()`'s single row is cut at the cap like any other input.

**Cutting it is only half the fix - the cut has to be reportable.** Simply breaking would have re-created the exact #5711 defect one level down: the row-level probe cannot see it, because everything a row refers to is expanded while that row is the current one, so nothing is left in the result set for `isTruncated` to find. `analyzeResultContent` and `analyzePropertyValue` therefore return `boolean` - "the cap stopped me with content still unexpanded" - which the `graph` and `studio` branches fold into `truncated`. That is what makes the response either honest (`truncated: true`) or refused (413 against the ceiling), instead of silently short. Neither method is overridden anywhere in the tree, so the signature change is contained.

**Tests**: `aCollectionInsideASingleRowCannotExpandPastTheCeiling` (Cypher `MATCH (n) RETURN collect(n)` with the `graph` serializer and `limit: -1`, over more elements than the ceiling → 413) and `aCollectionThatFitsIsStillExpandedWhole` (the same query under the ceiling → 200, complete, not flagged - so the new guard cannot pass by cutting everything). **Control run**: with the in-collection check disabled, exactly the first of the two fails (18 run, 1 failure) and the rest of the class still passes.

**Javadoc corrected** in both places that made the understated claim. The remaining overshoot for the element-counting serializers is now stated as what it actually is: bounded by the expansion of the single element that reaches the cap, which for an edge is itself plus its two endpoints.

On your compounding point - yes, `collect()` was the worst amplification factor available, and closing it is what brings the refusal cost back to the "up to `httpQueryMaxResultRows` elements serialized and discarded" bound the docs describe.

Verification: `HttpQueryMaxResultRowsIT` 18/18, server unit lane 791/791, and 135/135 across `HttpQueryTruncationIT`, `AbstractQueryHandlerIT`, `Issue4689StudioSerializerIT`, `PostCommandStatisticsIT`, `PostCommandHandlerProfileIT`, `HTTPDocumentIT`, `TimeSeriesQueryHandlerIT`, `OpenApiSpecGenerationIT`, `OpenApiDocsEndpointIT` and `Issue5037ProductionErrorBodyIT` - the serializer classes most exposed to the `analyzeResultContent` signature change.

Still deferred, unchanged: the `RemoteHttpComponent.manageException` typed arm, and the engine-level bound on `TimeSeriesEngine.query()`'s fetch.

robfrank:
Last round from the automated loop (max cycles reached). No blocking findings this time; here is what I did and did not act on.

**Closed the one remaining test gap** (fb644a66e8), even though you rated it "probably not worth a dedicated IT" - it is three lines and it pins a property worth pinning: `theLargestStatableLimitIsClampedBeforeTheProbeCanOverflow` sends `limit: 2147483647` and asserts 413 with the ceiling on (the clamp runs first, so `truncationProbeLimit`'s saturation is never reached), then turns the ceiling off and asserts every row comes back (the clamp is now a no-op, so the saturation guard is the only thing standing between the request and a pushed-down `limit 2147483648`). Two guards, one test, and it fails if either is removed. `HttpQueryMaxResultRowsIT` + `HttpQueryTruncationIT`: 41/41 green.

**`PostTimeSeriesQueryHandler.executeRawQuery`'s different shape** - correct reading, and it is not an oversight: `engine.query()` hands back the true count, so a "+1 probe" would be inventing lazy semantics that path does not have. The condition is covered by the IT and not by `MaxResultRowsClampTest` because the clamp itself *is* shared (`applyMaxResultRows`); only the comparison around it differs. Left as is.

**CI**: `HttpQueryMaxResultRowsIT` was run locally, 18/18 before this commit and green after, against a server started by the test rather than a hardcoded port - the class reads the bound port back from `getServer(0).getHttpServer().getPort()`, so a busy 2480 cannot silently route its requests at another server. CI will confirm independently.

**Left for the maintainer to file** rather than opened unprompted from this branch, all three documented in the release note and in the code:
1. The engine-level bound on `TimeSeriesEngine.query()`'s fetch - the one you flagged most strongly, and the only place where the protection this PR adds is genuinely incomplete rather than merely bounded.
2. A typed `RemoteHttpComponent.manageException` arm for `ResultSetTooLargeException`, so the Java driver can `catch` it instead of a generic `RemoteException` (it is already non-retried, which is the part that matters).
3. Nothing for the two documented non-guarantees (refusal cost, write amplification): those are the design, not a gap.

Issue #6264, opened 2026-08-16
Thirteen more algo.* procedures have the unguarded iteration knob #6216 fixed in two (follow-up to #6216)


lvca:
Follow-up to #6216 (PR #6222). That issue named three procedures - `algo.node2vec`, `algo.maxKCut`, `algo.influenceMaximization` - and they are fixed. The same two defects are present, unchanged, in **thirteen more** procedures that were never in scope.

## The two defects

#6216 established that an iteration-shaped knob needs two things:

1. **A domain minimum**, rejected by name. Below its minimum such a knob does not mean "a smaller run", it means an answer the algorithm cannot produce - and the value is absorbed in silence. `algo.maxKCut(2, {restarts: 0})` used to return every node in community 0 with a cut weight of `-1.0`, reported as success.
2. **A checkpoint inside the loop it drives**, so a large value is abortable via thread interrupt and `arcadedb.command.timeout` rather than forbidden by a guessed cap.

## The thirteen

Every one of these extracts its knob with a plain `extractInt(n, "maxIterations")` - no minimum - and contains **zero** references to `WorkGuard`:

| procedure | knob | default |
|---|---|---|
| `AlgoPageRank` | `maxIterations` | 20 |
| `AlgoPersonalizedPageRank` | `maxIterations` | 20 |
| `AlgoArticleRank` | `maxIterations` | 20 |
| `AlgoEigenvectorCentrality` | `maxIterations` | 100 |
| `AlgoHITS` | `maxIterations` | 20 |
| `AlgoKatz` | `maxIterations` | 10 |
| `AlgoLouvain` | `maxIterations` | 10 |
| `AlgoLeiden` | `maxIterations` | 10 |
| `AlgoLabelPropagation` | `maxIterations` | 10 |
| `AlgoSLPA` | `iterations` | 20 |
| `AlgoSimRank` | `maxIterations` | 5 |
| `AlgoFastRP` | `iterations` | 4 |
| `AlgoHashGNN` | `iterations` | 4 |

Verified against the merged tree: `grep -ln "maxIterations|\"iterations\"" *.java` returns 15 files, and only `AlgoNode2Vec` and `AlgoMaxKCut` - the two #6216 touched - contain any `guard.` call.

[...]

PR #6285, opened 2026-08-17
fix(#6263): one budget for the working set of an algo call, not just its walk buffers

### LINKED ISSUE #6263: Embedding matrices in the algo.* procedures are sized nodeCount x dimension and sit outside every budget (follow-up to #6216)
Follow-up to #6216 (PR #6222). Flagged during review of that PR as "worth reconciling once both land" - #6065/#6214 and #6216 have now both landed, so this is that reconciliation.

## The gap

#6065 capped the embedding-dimension knobs at `MAX_EMBEDDING_DIMENSION` (4096). #6216 added `arcadedb.cypher.algoMaxWalkMemory` and priced the random-walk buffers against it. Between them they leave the **largest allocation in these procedures completely unbounded**: the embedding matrices themselves, which are `nodeCount x dimension`.

```java
// AlgoNode2Vec.java:186-187
final double[][] W    = new double[n][dim];
final double[][] WCtx = new double[n][dim];

// AlgoFastRP.java:124,138
final double[][] embed    = new double[n][dimensions];
final double[][] newEmbed = new double[n][dimensions];

// AlgoHashGNN.java:157
final double[][] embeddings = new double[n][embDim];

// AlgoGraphSAGE.java:123,143
double[][] embed            = new double[n][initDim];
final double[][] nextEmbed  = new double[n][outDim];
```

`dimension` is capped, but `nodeCount` is not - and nothing checks the **product**. The dimension cap bounds one embedding *row* at 32 KB; it says nothing about the matrix.

## Why this is worth its own issue rather than a note

The numbers land inside the default configuration, not at an abusive one. For `algo.node2vec` with the **default** `embeddingDimension: 128`, `W` + `WCtx` cost `2 x n x 128 x 8` plus per-row headers, about **2080 bytes per node**:

| nodes | `W` + `WCtx` at default dim 128 | at the 4096 cap |
|---|---|---|
| 100 000 | ~208 MB | ~6.5 GB |
| 1 000 000 | ~2.1 GB | ~65 GB |
| 10 000 000 | ~21 GB | ~655 GB |

The comparison that makes the inconsistency concrete: at node2vec defaults the walk matrix costs `10n x (36 + 4x80)` = **~3560 bytes per node** and is refused up front if it exceeds the budget, while the embedding matrices cost **~2080 bytes per node** on the same call and are not checked at all. Two allocations of the same order of magnitude, one budgeted and one not.

`algo.fastRP` is the cleaner example: it has no walk buffer at all, so `arcadedb.cypher.algoMaxWalkMemory` never fires for it and its `2 x n x dimensions x 8` allocation has no ceiling of any kind.

The failure mode is an `OutOfMemoryError` - not a client error naming a parameter, which is what #6065 and #6216 both exist to produce.

## Suggested direction

Rather than a second budget key, generalise the one #6216 added. `CYPHER_ALGO_MAX_WALK_MEMORY` is named for walks because that was all it priced; the concept it implements - "estimate the working-set footprint in saturating long arithmetic, reject as a client error before allocating, auto-scale the default with the JVM heap" - is not walk-specific. Options worth weighing:

1. Rename to `arcadedb.cypher.algoMaxWorkingMemory` (keeping the old key as a deprecated alias) and price the embedding matrices through the same `checkWalkBudget`/`saturatingProduct` helpers, renamed to match.
2. Leave the walk budget alone and add a sibling key for embedding matrices - simpler, but two keys for one question ("how much heap may one algo call take?") is the worse interface.

Option 1 looks right: one question, one setting, and the helpers in `AbstractAlgoProcedure` already have the shape.

Note `AlgoGraphSAGE.java:138` also allocates `new double[outDim][concatDim]`, which is dimension-squared-ish rather than `n`-scaled. Bounded by the 4096 cap on both axes, so it is a smaller concern, but worth pricing in the same pass.

## Related

#5924, #6065, #6216, PR #6055, PR #6214, PR #6222.

lvca:
Closes #6263.

## The gap

#6065 capped every embedding-dimension knob at `MAX_EMBEDDING_DIMENSION` (4096); #6216 priced the random-walk buffers against a heap budget. Between them, the **largest** allocation of these procedures stayed outside every budget: the matrices themselves.

A dimension cap bounds one embedding **row** at 32 KB and says nothing about a `nodeCount x dimension` **matrix**. At `algo.node2vec`'s default `embeddingDimension: 128` the two Skip-gram matrices cost ~2080 bytes per node (208 MB at 100k nodes, 2.1 GB at 1M, 21 GB at 10M) - the same order as the walk matrix sitting beside them in the same method, at ~3560 bytes per node, which #6216 already refused up front. One budgeted, one not.

`algo.fastrp` made it plainest: it has no walk buffer at all, so **no budget of any kind applied to it**.

The failure mode was an `OutOfMemoryError` instead of the client error naming a parameter that #6065 and #6216 both exist to produce - and an OOM on a shared server takes down work unrelated to the query that caused it.

## What this does

Option 1 from the issue, generalised further.

**Rename**, no alias: `arcadedb.cypher.algoMaxWalkMemory` -> **`arcadedb.cypher.algoMaxWorkingMemory`**. The key was introduced in this same unreleased version (26.9.1-SNAPSHOT, #6222 merged today), so there is no deprecated alias to carry - the concept it implements was never walk-specific, only its name was. Default `max(64MB, maxHeap/8)`, negative = no limit, `IllegalArgumentException` -> HTTP 400: all unchanged.

**Reservations accumulate over the call** rather than being checked one allocation at a time. That is what "how much heap may this call take?" actually asks, and a single call routinely holds several of these at once - node2vec keeps its walk matrix alive while training over two embedding matrices - so a per-component check would let a call exceed the budget by however many components it happens to have. `checkWalkBudget` is replaced by a per-call `MemoryBudget` with `reserve(bytes, component, detail)`, plus `matrixBytes(rows, cols, elementBytes)` and `saturatingSum` next to the existing `saturatingProduct`.

```
algo.node2vec(): the embedding matrices would need 8448 bytes (2 matrices of 4 nodes x embeddingDimension=128),
on top of the 14240 bytes this call already reserved, more than the 20000 bytes allowed.
Set arcadedb.cypher.algoMaxWorkingMemory to raise the limit
```

Priced, all before anything is allocated:

| procedure | working set |
|---|---|
| `algo.node2vec` | walk matrix + 2 x `nodeCount x embeddingDimension` |
| `algo.fastrp` | 2 x `nodeCount x dimensions` |
| `algo.hashgnn` | 2 x `nodeCount x 4*embeddingDimension` feature matrices + 1 x `nodeCount x embeddingDimension` |
| `algo.graphsage` | 2 x `nodeCount x embeddingDimension` + the layer's `embeddingDimension x 2*initDim` projection |
| `algo.apsp` | `nodeCount x nodeCount` distance matrix |
| `algo.simRank` | 2 x `nodeCount x nodeCount` similarity matrices |
| `algo.maxFlow` | `nodeCount x nodeCount` capacity + residual |
| `algo.kShortestPaths` | `nodeCount x nodeCount` weight matrix + removed-edge mask |
| `algo.randomWalk` | walk buffer (unchanged from #6216) |

### Beyond what the issue asked for, and why

The issue scoped this to the four embedding procedures. The last four rows above are the reason the rename is worth doing at all rather than adding a second walk-shaped key: **their matrices are sized by the graph alone, with no knob involved anywhere**. `algo.simRank(a, b)` answers a question about exactly two nodes and allocates two full `nodeCount x nodeCount` matrices to do it (1.6 GB at 10k nodes), because the similarity of one pair is defined recursively over every pair. `algo.apsp` documented itself as suitable "up to a few thousand vertices", which was advice, not a bound. Naming a key `algoMaxWorkingMemory` and then leaving the largest working sets in the package unpriced would over-promise exactly the way `algoMaxWalkMemory` under-promised.

These algorithms are cubic or worse in time, so the memory bound bites at roughly the scale where the runtime is already impractical; what changes is that the refusal names the node count and the setting instead of arriving as an OOM several minutes in.

Two findings from the same pass:

- **`algo.hashgnn`'s feature matrices, not its embedding matrix, are the larger pair.** They are four times as wide, so even as `boolean` they cost half a byte per dimension per node against the embedding's eight. Priced by name rather than folded into an "embedding matrices" figure that would understate the call by ~2x.
- **`algo.graphsage`'s projection matrix dominates on a small graph** (`outDim x 2*initDim` = 67584 bytes at the defaults). Its peak is per-layer, not per-run - each layer drops the matrix it read - so `layers` does not multiply the reservation.

`AlgoGraphSAGE.java`'s `new double[outDim][concatDim]`, flagged in the issue as "worth pricing in the same pass", is included.

## Verification

New `Issue6263AlgoWorkingMemoryBudgetTest`, 14 tests. Sensitivity checked the way #6216 was: the eight touched procedures reverted to `HEAD` (and node2vec's embedding reservation removed by hand, since it cannot compile against the old helper) -> **11 of 14 red**. The three that stay green are the over-reach guards, which must pass both ways: the budget must not refuse a call it used to serve, at an ample budget, at a disabled budget, and - `everyPricedProcedureStillRunsUnderTheDefaultBudget` - at the untouched default across all nine procedures.

- `Algo*Test` + `Issue6216*` + `Issue6263*`: 325 tests green
- `*Cypher*Test` + `*Configuration*Test`: 3191 tests green
- full reactor `mvn install -DskipTests`: green

Release notes appended to `docs/release-26.9.1.md`, and the #6216 section's mention of the old key updated so the notes never name a key that no release contains.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2277,8 +2277,8 @@ Priced against the budget, all before anything is allocated:
 |---|---|
 | `algo.node2vec` | walk matrix + 2 x `nodeCount x embeddingDimension` |
 | `algo.fastrp` | 2 x `nodeCount x dimensions` |
-| `algo.hashgnn` | 2 x `nodeCount x 4·embeddingDimension` feature matrices + 1 x `nodeCount x embeddingDimension` |
-| `algo.graphsage` | 2 x `nodeCount x embeddingDimension` + the layer's `embeddingDimension x 2·initDim` projection |
+| `algo.hashgnn` | 2 x `nodeCount x 4 x embeddingDimension` feature matrices + 1 x `nodeCount x embeddingDimension` |
+| `algo.graphsage` | 2 x `nodeCount x embeddingDimension` + the layer's `embeddingDimension x 2 x initDim` projection |
 | `algo.apsp` | `nodeCount x nodeCount` distance matrix |
 | `algo.simRank` | 2 x `nodeCount x nodeCount` similarity matrices |
 | `algo.maxFlow` | `nodeCount x nodeCount` capacity + residual matrices |
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2235,6 +2235,102 @@ nameless `NegativeArraySizeException`.
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216)
 
+## The same iteration-knob guard, applied to the fourteen `algo.*` procedures #6216 left out of scope (#6264)
+
+[#6216](https://github.com/ArcadeData/arcadedb/issues/6216) established what an iteration-shaped knob needs -
+a domain minimum rejected by name, and a checkpoint inside the loop it drives - and gave both to the three
+procedures its parent review had named. Fourteen more carried the identical defect, untouched:
+`algo.pageRank`, `algo.personalizedPageRank`, `algo.articleRank`, `algo.eigenvector`, `algo.hits`,
+`algo.katz`, `algo.louvain`, `algo.leiden`, `algo.labelPropagation`, `algo.slpa`, `algo.simRank`,
+`algo.fastrp`, `algo.hashgnn` and `algo.graphsage`. Every one extracted its knob with a plain
+`extractInt(n, "maxIterations")`, and none contained a single checkpoint.
+
+So `CALL algo.pageRank({maxIterations: 0})` returned the *uniform initial rank vector* as though it were a
+PageRank result, `algo.louvain` returned every node in its own community, and `algo.fastrp` the untouched
+random projection, and `algo.graphsage` the untouched random-Gaussian initial features presented as trained
+embeddings. The silent half is the more serious one: an un-iterated centrality is not obviously wrong to a
+caller, unlike an exception. All fourteen now reject a value below 1 with a message naming the procedure, the
+parameter and the value.
+
+For time there is still no honest ceiling to pick, so a large value is not forbidden but made abortable. Each
+iteration loop calls the shared `WorkGuard`, which observes a thread interrupt and the
+`arcadedb.command.timeout` deadline; a per-node checkpoint inside each pass bounds the abort latency below a
+whole sweep of the graph, throttled to once every 1024 nodes where a node's own work is small and unthrottled
+where it is already O(n). Six of the fourteen have no convergence test at all - the CSR PageRank kernel,
+`algo.simRank`, `algo.fastrp`, `algo.hashgnn`, `algo.graphsage` and `algo.slpa` - so the knob alone decided
+when they stopped and nothing could end the run early.
+
+Two of them hand the work to `GraphAlgorithms`, which lives below the query layer and knew nothing about
+deadlines. Rather than couple the OLAP kernels to the query engine, `GraphAlgorithms.pageRank` and
+`GraphAlgorithms.labelPropagation` gained an overload taking a `WorkCheckpoint`, a one-method interface in
+`com.arcadedb.graph.olap` that the procedures satisfy with a method reference to their own guard. Existing
+callers are unchanged and get a checkpoint that never aborts.
+
+`algo.slpa` needed one thing more. Alone among the fourteen its `iterations` buys heap as well as time: every
+node keeps a label-memory row of `iterations + 1` ints, so `{iterations: 1000000}` on a 10k-node graph asks
+for 40 GB, and at `Integer.MAX_VALUE` the `iterations + 1` wrapped to `Integer.MIN_VALUE` and died as a bare
+`NegativeArraySizeException` naming nothing. The footprint is now estimated in saturating `long` arithmetic
+and checked against the same budget the walk buffers use, before the first row is allocated. (That budget is
+renamed `arcadedb.cypher.algoMaxWorkingMemory` by #6263 below, which landed after this one and generalised it
+from walk buffers to the whole working set of a call; the label memory is priced through it unchanged.)
+
+[#6264](https://github.com/ArcadeData/arcadedb/issues/6264)
+
+## A placeholder whose content had to spill into chunks is no longer returned twice by a scan (#6196)
+
+`SELECT FROM Doc` could return the same record twice, under two different RIDs, and `count(@rid)` counted it
+twice with it. One record shape reached it: a **placeholder** whose CONTENT record was too large for any page to
+hold whole.
+
+A record that outgrows its own page normally becomes a chunk chain in place. When its slot is too small even for
+the 14 bytes a chunk header needs - since #6149 the only remaining case, and it takes a page with a free tail of
+exactly zero - the slot becomes a 9-byte POINTER instead and the content moves to a record of its own on another
+page. Such a content record is not a document; it is reachable only through the pointer, and every reader that
+walks a page knows to skip it because its slot carries its size **negated**. That is the entire mechanism, and it
+is the one thing a content record bigger than a page cannot have: it has no size in its slot at all, because it is
+a chain. `writeMultiPageRecord` stamped the plain `FIRST_CHUNK` marker of an ordinary multi-page record on its
+head, and at that moment the information that the slot belonged to somebody else was gone - there was nowhere
+else it was written down.
+
+So the bytes were handed out twice: once through the pointer, under the RID the application knows, and once as a
+document in their own right under the head chunk's RID. `check()` had the mirror of the same confusion, counting
+the record under `totalMultiPageRecords` and never under `totalSurrogateRecords`.
+
+The head chunk of a content record now carries a marker of its own, `FIRST_CHUNK_PLACEHOLDER_CONTENT` (-4), in the
+one value the marker namespace still had free between `NEXT_CHUNK` (-3) and the negated sizes (< -5). It costs
+nothing in the stored format - one zigzag byte, exactly like the markers either side of it - and it restores the
+rule the negated size marker always expressed: a scan, a `count()` and an existence check hand out records, and
+this is not one, while everything that walks or rewrites the CHAIN treats the two heads identically.
+
+**Databases written before this** still hold the ambiguous shape, and no reader can tell it from an ordinary
+multi-page record - only the pointer that leads to it can. Rather than have every reader tolerate the ambiguity
+for ever, `CHECK DATABASE` now follows each placeholder pointer, reports a content record still stored the old way as an error naming its RID, and
+`CHECK DATABASE FIX` repairs it by rewriting that one marker. It is a repair and never a deletion: not a byte of
+the record, its chunk header or its chain is touched.
+
+**Run that FIX before writing through a full scan of such a database.** Until it has run, the content record is
+still handed out under a RID of its own, and that RID looks like any other: an application that scans a type and
+updates or deletes what comes back can reach the content record directly, where nothing can tell it apart from a
+record and the write lands on content the placeholder pointer still references. This is the behaviour every
+release before this one already had - the marker is what closes it, and `CHECK DATABASE FIX` is what applies the
+marker to data written earlier. A read-only scan of an unrepaired database merely returns the record twice; a
+read-modify-write over one is worth the FIX first.
+
+Following the pointers costs `CHECK DATABASE` one page fetch per placeholder POINTER - every one of them, not only
+the chunked case it is looking for, because the marker on the other end is the only thing that tells the two apart.
+A pre-#6149 bucket dense with placeholders therefore pays one extra fetch each, of pages the same pass reads
+anyway, on an operation that already reads every page of the bucket and walks the full chunk chain of every
+multi-page record.
+
+One reporting detail to expect on an unrepaired database: only the TOTALS `CHECK DATABASE` returns are
+reconciled, because a content record can sit on a page the walk has already left behind and reconciling as it
+went would make the answer depend on the order the allocator happened to place the pages in. The per-page
+tallies printed by a verbose run are not, so for such a record a page's own `multiPageRecords` count disagrees
+with the top-level `totalSurrogateRecords` it was moved to. That is deliberate: a physical-layout log describes
+the page as the walk found it. Both agree again once the FIX has run.
+
+[#6196](https://github.com/ArcadeData/arcadedb/issues/6196)
+
 ## The whole working set of a graph-algorithm call is budgeted, not just its walk buffers (#6263)
 
 Follow-up to #6216, which introduced a heap budget for the random-walk buffers of `algo.node2vec` and
@@ -2285,6 +2381,7 @@ Priced against the budget, all before anything is allocated:
 | `algo.kShortestPaths` | `nodeCount x nodeCount` weight matrix + removed-edge mask |
 | `algo.steinerTree` | `terminals x nodeCount` Dijkstra tables + four arrays of one entry per terminal pair |
 | `algo.randomWalk` | walk buffer (unchanged from #6216) |
+| `algo.slpa` | `nodeCount x (iterations + 1)` label memory (from #6264, repriced through the same budget) |
 
 The last four are the reason the rename is worth doing rather than adding a second walk-shaped key. Their
 matrices are sized by the graph alone, with no knob involved anywhere: `algo.simRank(a, b)` answers a question
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2283,6 +2283,7 @@ Priced against the budget, all before anything is allocated:
 | `algo.simRank` | 2 x `nodeCount x nodeCount` similarity matrices |
 | `algo.maxFlow` | `nodeCount x nodeCount` capacity + residual matrices |
 | `algo.kShortestPaths` | `nodeCount x nodeCount` weight matrix + removed-edge mask |
+| `algo.steinerTree` | `terminals x nodeCount` Dijkstra tables + four arrays of one entry per terminal pair |
 | `algo.randomWalk` | walk buffer (unchanged from #6216) |
 
 The last four are the reason the rename is worth doing rather than adding a second walk-shaped key. Their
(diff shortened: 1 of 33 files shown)


lvca:
Round 1 addressed in 1a0d871.

**1. Adjacency built before the budget check** - valid, applied. The reservation now precedes `graph.adjacency()` in `AlgoFastRP`, `AlgoNode2Vec`, `AlgoHashGNN`, `AlgoGraphSAGE` and `AlgoSimRank` (which had the same shape and wasn't in your list). In all five the estimate needs only the node count and the knobs, so a call that is going to be refused no longer pays the O(edges) build first. You're right that it was the pre-existing #6216 ordering rather than a regression, but since the PR touches every one of these call sites the cost of fixing it is a moved line.

`AlgoAPSP`, `AlgoMaxFlow` and `AlgoKShortestPaths` already reserved before any edge iteration.

**2. `kShortestPaths` message wording** - valid, applied. It sums a `double` weight matrix and a `boolean` mask: same shape, different size per element, which is exactly why the arithmetic uses `saturatingSum` there rather than `saturatingProduct(2L, ...)`. The detail now reads "a double matrix and a boolean matrix, each n x n nodes". `algo.maxFlow`'s two matrices really are both `double`, so "2 matrices of n x n nodes" is accurate there and stands.

**3. Per-call budget, not a global pool** - correct, and deliberately left alone. Bounding aggregate heap across concurrently running algo calls is a different mechanism (a shared pool with acquire/release semantics and a policy for what a saturated pool does to a waiting query), not a change to this setting's wording or scope. Worth its own issue if concurrent-load OOMs show up; not something to bolt onto a per-call estimate.

On the verification you couldn't run locally: `Algo*Test` + `Issue6216AlgoWorkKnobBoundsTest` + `Issue6263AlgoWorkingMemoryBudgetTest` are 325 tests, green after the reorder as well. The reorder is the kind of change that could have silently changed a seeded result - it does not: `Random` construction moved, not its consumption order.

lvca:
Round 2 addressed in 590963c. The `algo.steinerTree` finding was right on both counts, and folded in rather than deferred - the PR's own claim is that no dense working set in the package is left unpriced, and a known exception to that would be a divergence to document rather than a scope line.

**Reproduced before fixing.** Reverting `AlgoSteinerTree` to HEAD and running the three new tests gives, verbatim:

```
Caused by: java.lang.NegativeArraySizeException: -1073716337
```

at 46342 terminals - exactly the shape you predicted. Worth noting *why* 46342 rather than 65536: the division by 2 happens after the product, so `t * (t - 1)` wrapping is not saved by the result fitting an `int`. The true count there is **1073767311** pairs, which does fit an int; it is the intermediate that wraps.

**What landed:**
- `pairCount` computed in `long`, reserved against the budget (about 43 GB across the four parallel arrays at that length, so the *default* budget refuses it - the two defects turn out to be the same defect), and refused outright past `Integer.MAX_VALUE` entries whatever the heap setting says, mirroring `algo.node2vec`'s `totalWalksAsLong` guard.
- `matrixBytes(t, n, DOUBLE_BYTES) + matrixBytes(t, n, INT_BYTES)` reserved for the Dijkstra tables, before the adjacency build, as with the other procedures.
- One correction to your suggested estimate: the pair arrays are not three primitive arrays but four, and the fourth is an `Integer[]` index array sorted through a `Comparator`. At 24 bytes an entry against the 4 of the `int` it carries, folding it into a primitive figure would understate that reservation by about 2.5x, so there is a new `BOXED_INTEGER_BYTES` and it is priced at what it costs. (Replacing the boxed sort with a primitive one is a real win - CLAUDE.md asks for primitive arrays over boxed collections on exactly these paths - but it is a behaviour-preserving refactor of Kruskal's ordering, not a bound, so it is not in this PR.)
- `terminalNodes` gets no numeric cap, consistent with the rest: bounded by the resource it spends, not by a guessed maximum.

**Sweep, so round 3 doesn't find another.** Every two-dimensional allocation in `procedures/algo` whose first dimension is not the node count: `AlgoGraphSAGE`'s `W` and `AlgoNode2Vec`'s `walks` (both already priced here) and `AlgoSteinerTree`'s `dist`/`prev` (now priced). `AlgoMST`'s `Integer[] sortIdx` is sized by the edge count, which is the graph itself rather than a caller-supplied multiplier, so it is deliberately out.

**Verification:** 328 tests across `Algo*Test` + `Issue6216*` + `Issue6263*`, green. The three new SteinerTree tests are red against HEAD's `AlgoSteinerTree` (the `NegativeArraySizeException` above, plus the two budget cases), so none of them is vacuous.

On the two remaining notes: the inner-class one is fair - `MemoryBudget`/`WorkGuard` hold the outer reference only to reach `getName()`, and both are short-lived locals on a procedure instance that carries no per-call state, so I've left them matching `WorkGuard`'s existing shape rather than splitting the two apart. The concurrent-heap one stands as an acknowledged scope cut.

lvca:
Round 3 addressed in 0607a5e. One point, valid, applied.

**`CYPHER_ALGO_MAX_WORKING_MEMORY`'s description omitted `algo.steinerTree`** - correct, and the reason it matters is exactly the one you gave: that text is served to users through the server config endpoint, so it is documentation rather than a comment, and it was a commit behind the code it describes. It now names the `terminals x nodeCount` tables and the terminal-pair arrays, and the sentence about what would otherwise reach the allocator now includes "a terminal list of any length" alongside the in-range int and the large graph - steinerTree is the one entry in that list whose driver is neither a number nor the graph.

On the two design observations, no change and no disagreement:

- `BOXED_INTEGER_BYTES = 24L` over-estimating below the `Integer` cache threshold is deliberate, and you have the direction right: for a budget check, over-estimating rejects slightly early while under-estimating admits a call that then OOMs, so the asymmetry is the whole point. It is also the same reasoning as `MATRIX_ROW_OVERHEAD_BYTES`, which is why the javadoc cross-references it rather than restating the caveat.
- The reserve-before-`graph.adjacency()` ordering: the cheap-rejection property you name is the better framing of it than the one in my round-1 reply. A refused call now costs one `loadGraph` and some arithmetic instead of an O(edges) build, which matters more for a hostile caller than for an honest one.

One correction to the numbers in your test-coverage section: `Issue6263AlgoWorkingMemoryBudgetTest` is now **17** tests, not 14 - the three `algo.steinerTree` cases landed in the round-2 commit after the PR description was written. The revert-based sensitivity result is 14 of 17 red against `HEAD` (11 from the original round plus all three steinerTree cases); the three that stay green are the over-reach guards, which must pass both ways or they are not guards.

Full run after this commit: 48 tests across `Issue6263*` + `Issue6216*` + `AlgoSteinerTreeTest`, green; 328 across the whole `Algo*` set as of the previous commit, unaffected by a description string.

lvca:
Round 4 addressed in eb80d4f. Two of the three applied, one declined as a separate concern.

**Javadoc repetition** - valid, and the sharpest of the three. Six class javadocs were each restating "estimated in long arithmetic and reserved before anything is allocated", which is WHAT `MemoryBudget` already documents; CLAUDE.md asks for the non-obvious WHY and nothing else. Each now names only what is specifically priced in that procedure and links the helper for the mechanism - e.g. graphSAGE's is down to one sentence that carries the only fact a reader cannot get from `MemoryBudget`, that the peak is per-layer because each layer drops the matrix it read.

**Mixed `x` / `×` / `·`** - applied within the prose this PR added; it had drifted across three characters. The remaining `×` are in pre-existing comment text this PR does not otherwise touch.

**`AlgoKShortestPaths` reallocating the `n x n` mask per spur node** - agreed on both halves of your reading: the budget stays correct because only one is ever live, and the churn is a real but separate performance concern. Not folded in, and for a reason worth stating rather than just deferring: hoisting it means clearing the mask between spur nodes instead of allocating it, and `Arrays.fill` over n² booleans per iteration is not obviously cheaper than a fresh allocation the JVM can thread-locally bump-allocate and collect young. It needs a measurement, and a measurement makes it a performance PR rather than a bound. I'll file it.

**On the GraphSAGE over-pricing you flagged**: correct, and deliberate - later layers project from `outDim`, not `initDim`, so they are priced above what they allocate. Both directions were available and the conservative one is the only safe one: an exact per-layer estimate would have to be re-reserved each layer, and a budget that admits a call and then refuses it three layers in has spent the work it exists to avoid spending. The over-estimate is bounded by `initDim / outDim`, which is 1 whenever `embeddingDimension >= 16`, i.e. every case but the smallest.

Test counts, since the description is now two commits behind: 17 tests in `Issue6263AlgoWorkingMemoryBudgetTest`, 14 of 17 red under the revert-based sensitivity check. 328 green across `Algo*Test` + `Issue6216*` + `Issue6263*` after this commit - comments and one string, but the run is cheap and the alternative is assuming.

lvca:
Round 5 addressed in f23799c. One nit applied, and one thing in the review itself worth correcting.

**Garbled sentence in the test javadoc** - real, mine, fixed. It had lost half a clause in an edit pass, and it also predated `algo.steinerTree` being priced, so it now names the one working set sized by neither a knob nor the graph.

**Correction: `removedEdges` is not allocated once per run.** The review says it is "allocated once per run (reused across spur nodes)"; it is a fresh `boolean[n][n]` inside the spur-node loop, allocated per spur node. Round 4's review had this right and flagged the churn. The pricing is unaffected - only one is ever live, so the peak is one mask - but the comment said only "only one mask is ever live", which is what allowed the misreading, so it now says both halves: allocated per spur node, priced once because the peak is one. Worth flagging because a reviewer reading that comment as "allocated once" would also read the churn concern as already handled, and it isn't.

**`BOXED_INTEGER_BYTES` under compressed oops** - you're right that the reference is 4 bytes below a 32GB heap, so ~20 rather than 24. Left as-is for the reason you name, plus one more: the two errors point in opposite directions and the safe one is bigger. 24 over-estimates a compressed-oops reference by 4 bytes and under-estimates nothing; going to 20 would make the estimate exact on the common configuration and 4 bytes short per entry on the uncommon one, which is the direction that admits a call and then OOMs.

**Sensitivity, since two rounds of CI could not run the suite:** the numbers in the description are from a local run, but the claim that matters is the negative one, and it is reproducible - revert the nine procedures to `HEAD` and 14 of 17 tests in the new class fail, including `NegativeArraySizeException: -1073716337` from `algo.steinerTree`. The three that survive are the over-reach guards, which must pass both ways. CI's `unit-tests` lane runs the class either way.

Latest local run after this commit: 22 tests across `Issue6263*` + `AlgoKShortestPathsTest`, green (comment-only changes, but cheaper to run than to assume).

lvca:
Round 6 is clean - no changes requested - and rounds 4, 5 and 6 all closed without a blocking finding, so I'm treating this as done and leaving the branch at f23799c rather than opening a seventh cycle over cosmetics.

The two things this round raised that I did *not* change, and why:

- **`BOXED_INTEGER_BYTES` conservative below the `Integer` cache** - deliberate, and you have the direction right. Correction to the framing though: it *over*-charges for small indices (24 bytes assumed against the 8 a cached instance really costs), not under-charges. That is the safe direction - over-charging refuses slightly early, under-charging admits a call that then OOMs - which is why it is left as-is rather than tuned.
- **`MemoryBudget.reserve()` mutating `reserved` before it throws** - a fair catch, and correct that it is harmless today: the exception aborts the call and the budget is call-scoped and discarded. It is still the wrong invariant for a reader to inherit, so it is filed rather than dropped.

Deferred work from rounds 4-6 is now #6289: the per-spur-node `n x n` mask churn in `algo.kShortestPaths`, the boxed `Integer[]` index sort in `algo.steinerTree` (and the identical shape in `AlgoMST`), and the three-line `reserve()` tidy. All three are performance or hygiene questions that want a measurement, not bounds - which is the line this PR draws.

Final state: 6 review rounds, 5 commits. Round 1 moved the reservations ahead of the O(edges) adjacency build; round 2 found the one procedure I had missed, `algo.steinerTree`, with both an unpriced `t x n` working set and a live `int`-overflow reaching the allocator as `NegativeArraySizeException: -1073716337`; rounds 3-6 were documentation and wording. 328 tests green locally across `Algo*Test` + `Issue6216*` + `Issue6263*`; 17 tests in the new class, 14 of them red against the unfixed procedures.

lvca:
Heads-up: #6288 (issue #6264) merged as `6b20fb1`, and this PR is now `CONFLICTING`. That is the collision flagged during #6288's review — apologies for landing on top of you. Here is exactly what overlaps, so the rebase is mechanical.

**`AbstractAlgoProcedure`** — three additive changes, none semantic:
- `checkWalkBudget(db, bytes, detail)` is now a thin delegate to a new `checkBufferBudget(db, bytes, what, detail)`, where `what` names the buffer ("random walk buffer", "label memory"). If this PR renames the *setting*, the rename lands inside `checkBufferBudget` only — `checkWalkBudget` never reads the config any more.
- New `saturatingSum(long, long)`, the companion to `saturatingProduct`. It exists because `saturatingProduct(...) + OVERHEAD` with a plain `+` wraps **negative**, and a negative estimate satisfies `estimatedBytes <= budget` for *any* budget — the guard disabled by exactly the input it exists to refuse. Unreachable with today's callers, but worth keeping as this PR adds callers whose per-entry size is not 4 bytes.
- `WALK_ROW_OVERHEAD_BYTES`'s javadoc is generalised from "walk matrix" to "per-node int matrix".

**`AlgoNode2Vec`, `AlgoFastRP`, `AlgoGraphSAGE`, `AlgoHashGNN`, `AlgoSimRank`** — #6288 touched all five, but only for the iteration knob (a `minimum` on `extractInt`, and `guard.check()`/`checkPeriodically()` calls). Nothing there reads the memory budget except `AlgoNode2Vec`'s `bytesPerWalk`, which #6288 rewrote to use `saturatingSum`. Should be an additive merge in each.

**One thing worth knowing rather than merging:** if this PR renames `arcadedb.cypher.algoMaxWalkMemory`, note that #6288 extended that setting's description to cover `algo.slpa`, whose `iterations` now sizes a per-node label-memory row and is priced against it. The renamed setting needs to keep describing that, and `checkBufferBudget`'s "label memory" caller needs to keep working.

Also filed #6295 from the same review: `algo.hashgnn`'s MinHash phase is O(nodeCount x embeddingDimension²) and sits outside every checkpoint — 113 s against a 1 s `command.timeout` on 2000 nodes. That is a *time* bound and this PR is about *heap*, so they should not collide, but it is in `AlgoHashGNN` which this PR also touches.

lvca:
Rebased onto main (4b7e686) after #6288 landed. `MERGEABLE` again.

The overlap was real rather than textual: #6288 needed a heap bound for `algo.slpa`'s label memory and generalised `checkWalkBudget` into `checkBufferBudget(db, bytes, what, detail)` to get one. That is the same idea this branch implements as `MemoryBudget`, which additionally accumulates over the call - so taking both sides literally would have left **two budget APIs on the same setting**, which is the thing this PR exists to stop being. Resolved toward one:

- `checkBufferBudget`/`checkWalkBudget` are gone. `algo.slpa` reserves through `MemoryBudget` and reads identically to a caller, because the two message shapes were already the same. Its estimate becomes `matrixBytes(n, iterations + 1, INT_BYTES)` - the same number its hand-rolled expression computed - and moves ahead of the adjacency build like the other eight.
- Both branches had added `saturatingSum`, and the merge kept both copies, which does not compile. Kept #6288's, javadoc included: it explains why a footprint mixing a saturated product with a plain addition must not wrap, which is the better version of that comment. Its `@link` now points at `MemoryBudget.reserve`, where the hazard would actually land.
- `AlgoNode2Vec`: this branch moved the reservation ahead of the adjacency build, #6288 made `bytesPerWalk` saturating in place. Kept both - the block moved, with the saturating arithmetic inside it.
- **`AlgoGraphSAGE`: #6288 gave `layers` a minimum of 1, which makes this branch's `if (layers > 0)` guard unreachable-false.** Now unconditional; a merge that only resolved the marked hunks would have left dead code behind.
- `GlobalConfiguration`: one description covering both, naming the label memory alongside the matrices. #6288's release-notes section referred to the setting by the name this PR renames, so it now says so and points forward - the notes must not name a key no release will contain.
- Release notes: both sections kept, #6288's first since it landed first, and `algo.slpa` added to the priced-components table.

**Verified after the merge, not assumed:** 395 tests green across `Algo*Test` + `Issue6216*` + `Issue6263*` + **`Issue6264*`** (#6288's own 70), and 3459 green across the Cypher/config/recent-issue sweep. #6288's `slpaRejectsALabelMemoryLargerThanTheBudget` passes against the new API unchanged apart from the renamed key constant.

Issue #6266, opened 2026-08-16
arcadedb.command.timeout is described as the timeout for commands but is honoured by only two code paths


lvca:
Noticed while wiring `arcadedb.command.timeout` into the OpenCypher `algo.*` procedures for #6216 (PR #6222).

## The gap

`GlobalConfiguration.COMMAND_TIMEOUT` is declared as:

```java
COMMAND_TIMEOUT("arcadedb.command.timeout", SCOPE.DATABASE,
    "Default timeout for commands (in ms)", Long.class, 0),
```

"Default timeout for commands" reads as a global safety net. Grepping the engine for who actually reads it returns exactly two consumers:

1. `SelectExecutionPlanner:151` - the SQL **SELECT** planner, which injects a `TimeoutStep` when no explicit `TIMEOUT` clause is present.
2. `AbstractAlgoProcedure:373` - the `WorkGuard` added by #6216, covering the OpenCypher `algo.*` procedures.

Everything else ignores it. Not covered, as far as I can tell from that grep:

- OpenCypher queries generally - a `MATCH` with a bad plan, an expensive `OPTIONAL MATCH`, a variable-length expansion. Only a `CALL algo.*` inside one is bounded.
- SQL statements that are not `SELECT` - `UPDATE`, `DELETE`, `INSERT ... FROM SELECT`, `TRAVERSE`, `MATCH`.
- Gremlin, which has its own `arcadedb.gremlin.timeout` (deprecated) and `arcadedb.polyglotCommand.timeout`.

So an operator who sets `arcadedb.command.timeout=30000` expecting "no command runs longer than 30 s" gets it for SQL `SELECT` and for `algo.*`, and silently does not get it for an OpenCypher `MATCH` - which is the shape most likely to run away on a graph workload.

## Why this is worth filing rather than shrugging at

A setting that is documented broadly and enforced narrowly is worse than one that does not exist: it converts an operator's deliberate safety measure into a false belief. The default is `0` (disabled), so nobody is currently harmed by a *behaviour* bug - the harm is in the description.

There is also a consistency argument now. #6216 established `WorkGuard` in `AbstractAlgoProcedure`, which reads the deadline once and tests it in the hot loops with a `checkPeriodically` throttle, reading the clock only when a deadline is actually configured (so the default costs one flag test per iteration and no syscall). That is a reusable shape, and the OpenCypher execution steps are where it would pay most.

## Two honest directions

[...]

PR #6288, opened 2026-08-17
fix(#6264): fourteen more algo.* iteration knobs get the minimum and the checkpoint #6216 gave three

### LINKED ISSUE #6264: Thirteen more algo.* procedures have the unguarded iteration knob #6216 fixed in two (follow-up to #6216)

lvca:
Closes #6264. Follow-up to #6216 (PR #6222).

#6216 established that an iteration-shaped knob needs two things - **a domain minimum rejected by name**, and **a checkpoint inside the loop it drives** - and gave both to the three procedures its parent review had named. Thirteen more carried the identical defect, untouched. Every one extracted its knob with a plain `extractInt(n, "maxIterations")` and contained zero `guard.` calls:

`algo.pageRank`, `algo.personalizedPageRank`, `algo.articleRank`, `algo.eigenvector`, `algo.hits`, `algo.katz`, `algo.louvain`, `algo.leiden`, `algo.labelPropagation`, `algo.slpa`, `algo.simRank`, `algo.fastrp`, `algo.hashgnn`.

## Domain

`CALL algo.pageRank({maxIterations: 0})` returned the *uniform initial rank vector* as though it were a PageRank result; `algo.louvain` returned every node in its own community; `algo.fastrp` the untouched random projection. The silent half is the more serious one, as the issue says: an un-iterated centrality is not obviously wrong to a caller, unlike an exception. All thirteen now reject below 1, naming the procedure, the parameter and the value. `maxIterations` with a minimum of 1 also matches Neo4j GDS, where it is declared `@Configuration.IntegerRange(from = 1)`.

## Time

No honest ceiling exists for time, so a large value stays legal and becomes abortable instead. Each knob-driven loop calls the shared `WorkGuard` (thread interrupt + `arcadedb.command.timeout`), and a per-node checkpoint inside each pass bounds the abort latency below a whole graph sweep. Following the rule #6222 arrived at over three review cycles - *throttle only where one iteration can be cheaper than the check itself* - the per-node checkpoint is `checkPeriodically` where a node's own work is O(deg) or O(dim), and unthrottled where it is already O(n) (Louvain's `getCommunityDegree` scan, SimRank's inner `v` loop).

Five of the thirteen have **no convergence test at all** - the CSR PageRank kernel, `simRank`, `fastrp`, `hashgnn`, `slpa` - so the knob alone decided when they stopped.

## The two CSR delegates

`algo.pageRank` and `algo.labelPropagation` hand a CSR-backed graph to `GraphAlgorithms`, which sits *below* the query layer and knew nothing about deadlines. Rather than couple the OLAP kernels to the query engine, both gained an overload taking the new `WorkCheckpoint` - a one-method interface in `com.arcadedb.graph.olap` that the procedures satisfy with `guard::check`. Existing callers are unchanged and get `WorkCheckpoint.NONE`.

## Heap: `algo.slpa`

Alone among the thirteen, SLPA's `iterations` buys memory as well as time: every node keeps a label-memory row of `iterations + 1` ints, so `{iterations: 1000000}` on a 10k-node graph asks for 40 GB, and at `Integer.MAX_VALUE` the `iterations + 1` wrapped to `Integer.MIN_VALUE` and died as a bare `NegativeArraySizeException` naming nothing. The footprint is now estimated in saturating `long` arithmetic and checked against the same `arcadedb.cypher.algoMaxWalkMemory` budget the walk buffers use, before the first row is allocated. `checkWalkBudget()` becomes the walk-shaped special case of a new `checkBufferBudget()`.

## An alternative better than what the issue proposed

The issue framed this as \"mechanical application\" of #6216's mechanism. Two places it is not:

- **The CSR delegates.** A mechanical pass would have added a checkpoint to the twelve inline loops and left `algo.pageRank`'s CSR path - the single worst offender, since that kernel never converges - unguarded, because the loop is in another package. `WorkCheckpoint` fixes it without inverting the layering.
- **SLPA's heap.** A mechanical pass would have given `iterations` a minimum and a time checkpoint and left `{iterations: 1000000}` as a 40 GB allocation. #6216's actual principle is *bound a knob by the resource it spends*, and this knob spends two.

## Tests

New `Issue6264AlgoIterationKnobGuardTest`, 62 cases:

- 13 parameterised zero-rejections, one per procedure, plus 3 negative-count rejections
- 13 over-reach cases running each knob at exactly its minimum and asserting a full result set
- 13 interrupt aborts (deterministic: the flag is armed before the call, so the first checkpoint must observe it) and 11 command-deadline aborts. The two omitted from the deadline list, `louvain` and `leiden`, settle on the fixture before the clock is read - noted in the test rather than papered over
- `simRankHonoursTheCommandTimeoutInsideASingleIteration`: `maxIterations: 1` on an 800-node degree-400 graph, so the outer checkpoint runs once before any work and only the **per-node** checkpoint can fire. Per #6222's latency-test trap, \"it throws\" is non-vacuous here (the unguarded version returns a similarity rather than throwing) and the elapsed-time bound is measured from both sides: ~1.5 s guarded, ~29 s unguarded
- 3 direct tests on the CSR kernels (checkpoint called once per iteration; abort propagates), plus one end-to-end that leaves `tolerance` at its default so that only the CSR path - which ignores tolerance - can still be running when the deadline hits
- 3 SLPA heap cases: over budget, past `Integer.MAX_VALUE` with the budget disabled, and an under-budget over-reach guard

**Sensitivity, measured against the unmodified sources** (`git show HEAD:<file>` for the 13 procedures only): 20 of 35 non-abort cases go red, and the 24 abort cases hang rather than fail, which is the defect itself.

Full runs: 562 green across `com.arcadedb.query.opencypher.procedures.algo.*` + `com.arcadedb.graph.olap.*`; 4963 green across all `com.arcadedb.query.opencypher.**` + `com.arcadedb.graph.**`.

## Follow-ups spotted, not in scope

- `AlgoSLPA.mostFrequent()` is O(len^2) in the listener's degree, so a supernode makes one round quadratic in a way no knob controls. Graph-driven rather than knob-driven, so it belongs with #6266's family rather than here.
- `algo.simRank` allocates `new double[n][n]` twice - node-count-driven and outside every budget, the same shape as #6263.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2235,29 +2235,30 @@ nameless `NegativeArraySizeException`.
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216)
 
-## The same iteration-knob guard, applied to the thirteen `algo.*` procedures #6216 left out of scope (#6264)
+## The same iteration-knob guard, applied to the fourteen `algo.*` procedures #6216 left out of scope (#6264)
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216) established what an iteration-shaped knob needs -
 a domain minimum rejected by name, and a checkpoint inside the loop it drives - and gave both to the three
-procedures its parent review had named. Thirteen more carried the identical defect, untouched:
+procedures its parent review had named. Fourteen more carried the identical defect, untouched:
 `algo.pageRank`, `algo.personalizedPageRank`, `algo.articleRank`, `algo.eigenvector`, `algo.hits`,
 `algo.katz`, `algo.louvain`, `algo.leiden`, `algo.labelPropagation`, `algo.slpa`, `algo.simRank`,
-`algo.fastrp` and `algo.hashgnn`. Every one extracted its knob with a plain `extractInt(n, "maxIterations")`,
-and none contained a single checkpoint.
+`algo.fastrp`, `algo.hashgnn` and `algo.graphsage`. Every one extracted its knob with a plain
+`extractInt(n, "maxIterations")`, and none contained a single checkpoint.
 
 So `CALL algo.pageRank({maxIterations: 0})` returned the *uniform initial rank vector* as though it were a
 PageRank result, `algo.louvain` returned every node in its own community, and `algo.fastrp` the untouched
-random projection. The silent half is the more serious one: an un-iterated centrality is not obviously wrong
-to a caller, unlike an exception. All thirteen now reject a value below 1 with a message naming the
-procedure, the parameter and the value.
+random projection, and `algo.graphsage` the untouched random-Gaussian initial features presented as trained
+embeddings. The silent half is the more serious one: an un-iterated centrality is not obviously wrong to a
+caller, unlike an exception. All fourteen now reject a value below 1 with a message naming the procedure, the
+parameter and the value.
 
 For time there is still no honest ceiling to pick, so a large value is not forbidden but made abortable. Each
 iteration loop calls the shared `WorkGuard`, which observes a thread interrupt and the
 `arcadedb.command.timeout` deadline; a per-node checkpoint inside each pass bounds the abort latency below a
 whole sweep of the graph, throttled to once every 1024 nodes where a node's own work is small and unthrottled
-where it is already O(n). Four of the thirteen have no convergence test at all - the CSR PageRank kernel,
-`algo.simRank`, `algo.fastrp`, `algo.hashgnn` and `algo.slpa` - so the knob alone decided when they stopped
-and nothing could end the run early.
+where it is already O(n). Six of the fourteen have no convergence test at all - the CSR PageRank kernel,
+`algo.simRank`, `algo.fastrp`, `algo.hashgnn`, `algo.graphsage` and `algo.slpa` - so the knob alone decided
+when they stopped and nothing could end the run early.
 
 Two of them hand the work to `GraphAlgorithms`, which lives below the query layer and knew nothing about
 deadlines. Rather than couple the OLAP kernels to the query engine, `GraphAlgorithms.pageRank` and
@@ -2265,7 +2266,7 @@ deadlines. Rather than couple the OLAP kernels to the query engine, `GraphAlgori
 `com.arcadedb.graph.olap` that the procedures satisfy with a method reference to their own guard. Existing
 callers are unchanged and get a checkpoint that never aborts.
 
-`algo.slpa` needed one thing more. Alone among the thirteen its `iterations` buys heap as well as time: every
+`algo.slpa` needed one thing more. Alone among the fourteen its `iterations` buys heap as well as time: every
 node keeps a label-memory row of `iterations + 1` ints, so `{iterations: 1000000}` on a 10k-node graph asks
 for 40 GB, and at `Integer.MAX_VALUE` the `iterations + 1` wrapped to `Integer.MIN_VALUE` and died as a bare
 `NegativeArraySizeException` naming nothing. The footprint is now estimated in saturating `long` arithmetic
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2234,3 +2234,42 @@ purpose ("more seeds than nodes" reads as "as many as exist" and is clamped to t
 nameless `NegativeArraySizeException`.
 
 [#6216](https://github.com/ArcadeData/arcadedb/issues/6216)
+
+## The same iteration-knob guard, applied to the thirteen `algo.*` procedures #6216 left out of scope (#6264)
+
+[#6216](https://github.com/ArcadeData/arcadedb/issues/6216) established what an iteration-shaped knob needs -
+a domain minimum rejected by name, and a checkpoint inside the loop it drives - and gave both to the three
+procedures its parent review had named. Thirteen more carried the identical defect, untouched:
+`algo.pageRank`, `algo.personalizedPageRank`, `algo.articleRank`, `algo.eigenvector`, `algo.hits`,
+`algo.katz`, `algo.louvain`, `algo.leiden`, `algo.labelPropagation`, `algo.slpa`, `algo.simRank`,
+`algo.fastrp` and `algo.hashgnn`. Every one extracted its knob with a plain `extractInt(n, "maxIterations")`,
+and none contained a single checkpoint.
+
+So `CALL algo.pageRank({maxIterations: 0})` returned the *uniform initial rank vector* as though it were a
+PageRank result, `algo.louvain` returned every node in its own community, and `algo.fastrp` the untouched
+random projection. The silent half is the more serious one: an un-iterated centrality is not obviously wrong
+to a caller, unlike an exception. All thirteen now reject a value below 1 with a message naming the
+procedure, the parameter and the value.
+
+For time there is still no honest ceiling to pick, so a large value is not forbidden but made abortable. Each
+iteration loop calls the shared `WorkGuard`, which observes a thread interrupt and the
+`arcadedb.command.timeout` deadline; a per-node checkpoint inside each pass bounds the abort latency below a
+whole sweep of the graph, throttled to once every 1024 nodes where a node's own work is small and unthrottled
+where it is already O(n). Four of the thirteen have no convergence test at all - the CSR PageRank kernel,
+`algo.simRank`, `algo.fastrp`, `algo.hashgnn` and `algo.slpa` - so the knob alone decided when they stopped
+and nothing could end the run early.
+
+Two of them hand the work to `GraphAlgorithms`, which lives below the query layer and knew nothing about
+deadlines. Rather than couple the OLAP kernels to the query engine, `GraphAlgorithms.pageRank` and
+`GraphAlgorithms.labelPropagation` gained an overload taking a `WorkCheckpoint`, a one-method interface in
+`com.arcadedb.graph.olap` that the procedures satisfy with a method reference to their own guard. Existing
+callers are unchanged and get a checkpoint that never aborts.
+
+`algo.slpa` needed one thing more. Alone among the thirteen its `iterations` buys heap as well as time: every
+node keeps a label-memory row of `iterations + 1` ints, so `{iterations: 1000000}` on a 10k-node graph asks
+for 40 GB, and at `Integer.MAX_VALUE` the `iterations + 1` wrapped to `Integer.MIN_VALUE` and died as a bare
+`NegativeArraySizeException` naming nothing. The footprint is now estimated in saturating `long` arithmetic
+and checked against the same `arcadedb.cypher.algoMaxWalkMemory` budget the walk buffers use, before the
+first row is allocated.
+
+[#6264](https://github.com/ArcadeData/arcadedb/issues/6264)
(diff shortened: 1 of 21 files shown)


lvca:
Review actioned.

**The GraphSAGE gap — valid, folded in** (commit `fa5304d`). You are right that it is a fourteenth of the same thing, not a deliberate exclusion: `AlgoGraphSAGE.layers` is a bare `extractInt` driving a checkpoint-free loop, and `{layers: 0}` returns the random-Gaussian init as though it were a trained embedding. Fixed the same way as `fastrp`/`hashgnn`, plus four new parameterised cases (zero rejection, minimum accepted, interrupt abort, deadline abort). The zero-rejection case is red against the unmodified `AlgoGraphSAGE`, green with the fix.

One judgement call inside it worth stating: the Xavier initialisation in the layer loop gets **no** checkpoint of its own. Both its bounds are embedding dimensions, capped at `MAX_EMBEDDING_DIMENSION`, so it is bounded work whatever the caller asks for — unlike the per-node loop after it. Said so in a comment rather than leaving it to look like an omission.

**The javadoc callout on the granularity asymmetry — valid, added.** Both `GraphAlgorithms` overloads now state why one sweep is as fine as those kernels can be: the per-node work runs on `parallelForRange` worker threads, which would not observe an interrupt aimed at the calling thread, and throwing out of a chunk closure would leave its siblings running.

Counts corrected throughout (title, release notes, test javadoc): fourteen procedures, six of them with no convergence test at all.

**The two minor points — agreed, no change.** Your own reading is right on both: the SLPA array-capacity message is unreachable under a finite budget and is defence in depth, and the near-verbatim per-loop comments are intentional, since each is locally justified by the loop it sits above and a reader arriving at one file should not have to find another to know why the checkpoint is there.

Tests after the change: 119 green across `Issue6264AlgoIterationKnobGuardTest` (66), `AlgoGraphSAGETest`, `Issue6216AlgoWorkKnobBoundsTest` and `GraphAlgorithmsTest`.

lvca:
Round 2 actioned — thanks, the SimRank gap was a fair catch.

**The one suggestion — added** (`e6a92d3`). You're right that the `Arrays.fill` reset is the only place this PR changes what a kernel *computes* rather than only when it stops, and that nothing pinned its result. Note the four-node cycle cannot carry that assertion: every SimRank of two distinct nodes there is 0, so a broken reset would go unnoticed. The new test builds a hub pointing at two leaves instead, where the leaves share their only in-neighbour and `sim(A, B) = decay × sim(hub, hub) = 0.8`.

It asserts at **1 and 3** iterations, which is what makes it non-vacuous: the reset exists only for the diagonal (the `u < v` loop writes every off-diagonal cell itself) and the buffers swap each round, so dropping it leaves `sim(hub, hub)` at 0 from the second iteration on. Measured rather than assumed — with `newSim[i][i] = 1.0` removed the run still returns 0.8 at 1 iteration and fails at 3.

**The two "worth a second look" items — no change, and I think your own reading of both is right.**

- *SLPA check ordering.* The array-capacity message being shadowed by the budget message under a default budget is the intended precedence, not an accident: the budget is the bound an operator can tune and its message names the setting to raise, so it should win when both apply. The capacity refusal exists for the one case the budget cannot cover — `algoMaxWalkMemory` explicitly accepts "negative = no limit", and 2^31 rows do not fit a Java array whatever the heap is.
- *Louvain's unthrottled `check()`.* Agreed it is a different cost profile, and deliberately so. `getCommunityDegree()` is O(n) per candidate community per node, so the per-node body is already Ω(n) — one `currentTimeMillis()` against that is the cheap side of the trade, and throttling to 1024 would make the abort latency 1024 whole-graph scans. Same reasoning as SimRank's `u` loop; both carry it in an inline comment so the asymmetry with the throttled sites reads as a choice.

Suite after the change: 634 green across `com.arcadedb.query.opencypher.procedures.**` and `com.arcadedb.graph.olap.**`.

lvca:
Round 3 actioned (`de3340b`).

**The non-saturating addition — valid, fixed.** You're right that `saturatingProduct(...) + WALK_ROW_OVERHEAD_BYTES` defeats itself in the one case it matters: a saturated product plus `+32` wraps to a large negative number, and a negative estimate satisfies `estimatedBytes <= budget` for *any* budget — the check silently disabled by exactly the input it exists to refuse. Added `saturatingSum()` as the companion to `saturatingProduct()` and used it there.

I also applied it to `AlgoNode2Vec`'s `bytesPerWalk`, which has the identical raw-arithmetic shape and the identical proof that it cannot overflow today. Not because it is reachable, but so that a reader does not have to redo that proof per call site to work out which ones are safe — the value of the fix is entirely in closing the shape, and one site left in the old form is what makes the shape ambiguous again. Both carry a comment saying they are bounded today and why they are written this way anyway.

**The `currentTimeMillis()` cost note — acknowledged, no change**, for the reason you give: it is dwarfed by the O(n) per-node work at both unthrottled sites. Flagged in the inline comments already so the asymmetry with the throttled sites reads as deliberate.

Suite: 634 green across `com.arcadedb.query.opencypher.procedures.**` and `com.arcadedb.graph.olap.**`, unchanged by this commit.

lvca:
Round 4 actioned (`8037429`) — both points valid, both comment-only, no behaviour change.

**The uncommented checkpoints.** You spotted the real cause of the inconsistency: `AlgoPersonalizedPageRank` and `AlgoArticleRank` are the only two whose OLTP path duplicates the CSR path's loop, and only the CSR copy carried the note. In a PR where every other checkpoint explains itself, the OLTP copy read like an unexplained call. Both now say it, and so does the fallback branch inside the CSR path.

**The negative-count sampling.** Written down rather than left to be re-derived: what separates a negative value from zero lives entirely in the shared `extractInt(value, name, minimum)`, and the per-procedure half — that the knob is extracted with a minimum *at all* — is what the zero case covers for all fourteen. Fourteen rows there would re-test one shared comparison fourteen times, which is the argument for sampling rather than an excuse for it.

634 green across `com.arcadedb.query.opencypher.procedures.**` and `com.arcadedb.graph.olap.**`, unchanged.

From my side this is complete: four review rounds, every actionable point either implemented (GraphSAGE as a fourteenth instance, the SimRank value-pinning test, `saturatingSum`, these comments) or declined with a reason the review itself agreed with. Ready to merge as far as I'm concerned.

lvca:
Round 5 — nothing implemented, and here is the reasoning for each of the three.

**1. Comment duplication — declining, and this one has history worth recording.** Round 2 raised the same observation and concluded it read as intentional; round 4 then asked me to *add* the note to the two sites that lacked it (`AlgoPersonalizedPageRank`/`AlgoArticleRank`'s OLTP paths), which I did in `8037429`. So the rounds pull in opposite directions, and the tie-breaker is what a maintainer actually does with these files: they arrive at one procedure while debugging it, and the question "why is there a checkpoint here and why is this one throttled and that one not" has a *different* answer per file — it is derived from that loop's per-node complexity, not from a shared rule. A pointer to `WorkGuard`'s class doc would explain the mechanism but not the placement, which is the part that varies and the part #6222 spent three review cycles getting wrong. The drift risk is real but bounded: if the rationale needs correcting it needs correcting per-site anyway, because each site states its own complexity claim.

**2. Release-notes conflict risk — checked, and thank you, this was the actionable one.** Seven open PRs touch `docs/release-26.9.1.md` (#6271, #6272, #6275, #6277, #6284, #6285, #6287) and two touch `AbstractAlgoProcedure.java` (#6285, #6291). GitHub reports this PR `MERGEABLE` against `main` right now, so there is nothing to resolve yet.

The one to watch is **#6285** (#6263), which *renames* `arcadedb.cypher.algoMaxWalkMemory` → `algoMaxWorkingMemory` while this PR extends that same setting's description and generalises `checkWalkBudget` into `checkBufferBudget`. Whichever lands second needs a textual merge in `GlobalConfiguration` and `AbstractAlgoProcedure` — additive in the same place, nothing conflicting semantically. I'll resolve it here if #6285 goes first.

**3. Negative-count sampling** — already documented in the Javadoc as deliberate, as you note.

That closes every point raised across five rounds: implemented where actionable (`algo.graphsage` as a fourteenth instance, the SimRank value-pinning test, `saturatingSum`, the missing checkpoint comments), declined with reasons where not. No round found a correctness defect in the fix itself.

PR #6291, opened 2026-08-17
fix(#6266): a command deadline the whole command honours, not two code paths

### LINKED ISSUE #6266: arcadedb.command.timeout is described as the timeout for commands but is honoured by only two code paths

lvca:
Closes #6266.

## The gap

`arcadedb.command.timeout` was described as "Default timeout for commands". Grepping for who read it returned two consumers: `SelectExecutionPlanner` (SQL `SELECT`) and `AbstractAlgoProcedure` (the openCypher `algo.*` procedures, added by #6216). An operator who set it to bound a runaway query got that bound for `SELECT` and silently did not get it for an openCypher `MATCH` - the shape most likely to run away on a graph workload - nor for `TRAVERSE` or SQL `MATCH`.

The issue offered two directions: widen the enforcement, or narrow the description. This PR does both, because the description is only honest once the behaviour matches it.

## Better than the issue's own proposal

The issue suggested extending "the `TimeoutStep` idea". That idea is the problem, not the fix. `TimeoutStep`/`AccumulatingTimeoutStep` test the deadline *between* two `syncPull` batches, so they bound nothing when the unbounded work happens inside one batch: a filter that rejects every record spends the whole scan inside a single `hasNext()`, and the enclosing step is not re-entered until that call returns. Chaining more of those onto `MATCH` and `TRAVERSE` would have produced a setting that reports as enforced and still hangs. `sqlSelectWithARejectingFilterHonoursTheCommandTimeout` is that case, on the one statement kind the setting already "covered".

So the deadline moved down two levels instead:

**1. It belongs to the command, not to a step.** `CommandContext` resolves it once (`getCommandDeadline()`), and a context without one of its own inherits its parent's. A `CALL { }` body, a correlated `COUNT { }` probe, a UNION branch, an `algo.*` procedure and a parallel bucket-scan worker's `copy()` all land on the same instant instead of each starting a fresh budget - otherwise a statement buys itself unlimited time by nesting. `copy()` *resolves* rather than copies, because copying an unresolved deadline is the same as not copying one.

**2. It is checked inside the loops.** `WorkGuard` - promoted out of `AbstractAlgoProcedure`, where #6216 introduced exactly this shape and the issue called it reusable - now guards:

- openCypher steps: `MatchNodeStep`, `IndexSeekStep`, `MatchRelationshipStep`, `ExpandIntoStep`, `ExpandPathStep`, `OptionalMatchStep`, `EagerDistinctCollectOptionalMatchStep`, `ShortestPathStep`
- openCypher optimizer operators: the label/index scans, `ExpandAll`/`ExpandInto`, the GAV variants, `CartesianProduct`, `NodeHashJoin`, `FilterOperator`, `GAVFusedChainOperator` (including its parallel DFS workers)
- the CSR count push-downs (`CountOp` and its five implementations): a push-down replaces a materializing pipeline with an array pass and is normally orders of magnitude faster, but "faster" is not "bounded" - the triangle and pair-join shapes are superlinear in the degree
- `CypherExecutionPlan`'s own drains: the write-path materialization and the GQL `FINISH` drain, which nothing downstream can bound on their behalf
- SQL: `AbstractTraverseStep`, `MatchStep`, `FilterStep`, `ScanWithFilterStep`

`CartesianProduct` checks in `next()` rather than around a scan, because it emits |right| rows per left row without touching a source at all: guarding the scans below it would bound nothing.

**3. The description now says what it does**, including what it does not cover: Gremlin and the other polyglot engines (`arcadedb.polyglotCommand.timeout`) and regex backtracking (`arcadedb.command.regexTimeout`).

## Cost when disabled

The default stays `0`. With no deadline configured `WorkGuard.forCommandDeadline()` returns a shared instance, so a guarded loop pays one field load and one comparison per iteration and never reads the clock. Guards are built per `syncPull` batch, never cached on a step - a cached SQL plan re-executed later would otherwise carry an expired deadline and time out every query.

## Verification

New `Issue6266CommandTimeoutCoverageTest` (12 tests): openCypher self-join, blocking aggregation, variable-length expansion, write statement and nested `CALL { }`; SQL `TRAVERSE`, `MATCH` and a rejecting-filter `SELECT`; plus the disabled default, the deadline's copy/inherit contract, and one latency assertion (`StallAwareStopwatch`) separating "observed inside the loop" from "observed after it".

Each assertion requires the message to name `arcadedb.command.timeout`, so a statement that stopped for some other reason - including the pre-existing `AccumulatingTimeoutStep`, whose message does not - fails the test rather than passing it.

Non-vacuity was measured rather than assumed: with the bound raised out of the way, the five openCypher statements each run past 6 s, SQL `TRAVERSE` 2.2 s, SQL `MATCH` 1.2 s and the rejecting `SELECT` 1.6 s - 24x to 120x the 50 ms deadline the tests use. With `WorkGuard.check()` stubbed out the suite does not finish at all (killed after 10 minutes) where the guarded run takes 1.6 s.

Full `engine` module: 12118 tests, 0 failures. Full reactor build clean.

## Not covered, deliberately

Gremlin and the polyglot engines keep their own `arcadedb.polyglotCommand.timeout`; the guard is cooperative and cannot bound a script it does not execute. Regex backtracking does not poll deadlines at all and stays with `arcadedb.command.regexTimeout`. Both are now stated in the setting's description instead of being left to a grep.

### DIFF
--- a/bolt/conformance/COMPATIBILITY.md
+++ b/bolt/conformance/COMPATIBILITY.md
@@ -4,7 +4,7 @@
 
 Certification status of ArcadeDB's Bolt protocol against every official Neo4j driver, per the shared conformance spec ([`spec.yaml`](spec.yaml), epic #4882). Columns are driver language by pinned version ([`driver-versions.md`](driver-versions.md)).
 
-**Last verified:** 2026-08-16 03:47 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/31924692079))
+**Last verified:** 2026-08-17 04:02 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/31991815665))
 
 Legend: ✅ pass, ❌ fail, ⚠️ expected-fail / known limitation, ➖ not applicable, ⚪ skipped, `·` not reported. A `·` in a listed Coverage-gaps column means no result for that driver:version.
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2192,7 +2192,7 @@ dimensions. `algo.randomWalk` had the one-dimensional version: `new int[steps + 
 `Integer.MIN_VALUE` at `steps = 2147483647`.
 
 Both are now estimated in saturating `long` arithmetic and checked **before** anything is allocated, against
-a new `arcadedb.cypher.algoMaxWalkMemory` budget that auto-scales with the JVM max heap (one eighth of it,
+a new `arcadedb.cypher.algoMaxWorkingMemory` budget that auto-scales with the JVM max heap (one eighth of it,
 never below 64MB) and can be raised or switched off (negative = no limit). This follows the house pattern
 of `arcadedb.queryMaxHeapElementsAllowedPerOp` and `arcadedb.queryMaxRangeSize` rather than inventing a
 per-knob constant: the error names the knobs that produced the estimate and the setting to raise, and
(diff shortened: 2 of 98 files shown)


lvca:
Thanks - both points were fair. Addressed:

**1. `setCommandDeadline` sentinel collision** - fixed rather than documented around (`e4c02b676`). The "not resolved yet" marker is now `Long.MIN_VALUE` for both `commandDeadline` and `commandTimeout`, a value neither can legitimately take, so every value a caller pins is honoured - including `0`, which means "already expired" and is a reasonable thing to want. The interface method now states that contract, and a new test pins both `0` and `Long.MAX_VALUE` and asserts the guard aborts / does not.

**2. Test tagging** - measured rather than guessed, and the answer is no tag. `Issue6266CommandTimeoutCoverageTest` takes **1.158 s** in a full-suite `engine` run (12118 tests). That is below the 1.5-3 s band occupied by dozens of ordinary untagged classes - `HashIndexTest` 2.68 s, `SQLFunctionConvertTest` 2.17 s, `AsyncShutdownEscalationTest` 2.36 s - and 98 classes in the module already run longer than 1.5 s untagged. The 20k-vertex graph is built once in `@BeforeAll` and each timeout test aborts in tens of milliseconds by construction, so the setup is the whole cost. It is not a multi-second outlier, so `@Tag("slow")` would misfile it.

Unrelated observation from the same neighbourhood, left alone deliberately so the evidence survives: `PairHashJoinOp.buildWithViews` receives `arm2Buckets` and never applies it, which looks like a missing intermediate-label filter on the both-arms-single-hop fast path. I removed `provider` there (Codacy flagged it) but kept `arm2Buckets`, since deleting it would erase the symptom. Worth its own issue.

lvca:
The correctness finding was right, and it was the important one - thank you. Fixed in `f9c441cfe`.

**The per-statement `TIMEOUT` clause now reaches the in-loop checks.** New `StatementTimeouts` publishes the clause's instant onto the `CommandContext`, which is what the guards read:

- `TimeoutStep` pins once. It was already wall-clock from its first pull, so its own bound is unchanged.
- `AccumulatingTimeoutStep` **republishes on every batch** as `now + (clause - accumulated)` rather than pinning once. That matters: what that step charges against the clause is time spent *inside the pipeline*, so a consumer that pauses between two batches is not charged for the pause - and re-deriving the instant per batch means the guards inherit that accounting instead of silently converting the clause into a plain wall-clock bound.
- **The earlier of the clause and the command deadline wins.** Both are in force, so a statement may ask for less time than the operator allows but not for more. That is a change from the old planner behaviour (a clause simply replaced the global default at planning time), and it is the conservative direction, so I took it and corrected the setting's description, which had said "overrides" - you were right that the sentence was inaccurate either way.
- **Nothing is published for the `RETURN` strategy.** That clause asks for the rows produced so far rather than an exception, and a guard in a scan loop knows only how to throw; widening the enforcement must not quietly turn a documented "return what you have" into a failure.

The abort message now names the bound actually in force (`TIMEOUT clause of 50ms`) instead of always quoting the configuration key, which would have been a lie for a clause-bounded statement.

**A second, older bug fell out of this.** `SQLASTBuilder.visitTimeout` never read the `EXCEPTION | RETURN` token the grammar accepts (`SQLParser.g4:1316`), so every clause arrived with a null `failureStrategy` and `TIMEOUT n RETURN` has been silently behaving as `EXCEPTION` - the `RETURN` branch of `TimeoutStep.fail()` and `AccumulatingTimeoutStep.fail()` was unreachable. My exemption above depends on that value being readable, so I fixed the parser rather than writing a branch that could never fire. `theReturnFailureStrategyStillReturnsRatherThanThrowing` is the regression test: the same 1.6 s statement under `TIMEOUT 50 RETURN` now returns instead of raising.

**Four new tests**, exactly the case you asked for (global setting left disabled throughout the first three): clause on `SELECT`, clause on `UPDATE`, `RETURN` still returns, and the global ceiling holding under a looser clause.

On the other two points:

- **Sentinel**: already fixed in `e4c02b676`, which landed after the commit this review ran against. I went with `Long.MIN_VALUE` rather than `-1` for both fields, since `-1` is a legitimate epoch-millis instant even if an implausible one, and `Long.MIN_VALUE` is outside the domain entirely. `setCommandDeadline(0)` is now honoured and tested.
- **`forCommandDeadline` nit**: its javadoc already explains the choice ("leaving the thread interrupt flag alone... a pending interrupt may have been set for a reason unrelated to this statement and consuming it would change behaviour beyond the timeout this guard exists to enforce"). Happy to expand it into an explicit which-factory-to-pick guide if you think it still reads as under-explained.

Local verification: engine suite 12123 tests, 1 failure - `LSMVectorIndexSearcherPoolTest.concurrentSearchesNeverShareASearcher`, which touches no file in this diff and passes in isolation (CPU contention from a concurrent build on the same machine).

lvca:
Both new points were valid - fixed in `3e0b76d15`, along with more of the same shape found by auditing rather than waiting for another round.

**`PairHashJoinOp` probe loop** - correct, and a plain oversight: I guarded the fallback branch and not the `NeighborView` one beside it, which is the typical path. Rather than fix just that one, I checked every outer loop in all five `CountOp` implementations for the same mistake and found more:

- `AntiJoinChainOp.buildNeighborRIDMap` and `buildTailCounts` - both full OLTP record scans
- `PropagateChainOp`'s entire self-loop-subtraction family (`countSelfLoopPaths`, `countSelfLoop3HopEdgeScan`, `countSelfLoopPerAnchorFallback`, `countSelfLoopPathsOLTP`) and the OLTP anchor enumeration that seeds the propagation
- `PartitionedTriangleOp.buildPartitionMapping`, `precomputeBucketIds` in two files, and the inline bucket-id passes in `PairHashJoinOp`/`PropagateChainOp`

A mechanical re-audit now reports zero unguarded outer loops across those five files, so the description's claim holds rather than nearly holding. Thanks for pulling the thread - one miss was the visible end of a pattern.

**`MatchStep` per-row allocation** - correct, hoisted to `syncPull()` to match `FilterStep`/`ScanWithFilterStep`/`AbstractTraverseStep`.

**`Long.MIN_VALUE` residual** - documented on `setCommandDeadline`, as you suggested. Nothing can produce that value, so a doc note is the proportionate answer.

**Check-then-set on volatile** - documented, since you're right that it will draw the same comment again otherwise. The race is tolerated deliberately: concurrent resolvers walk the same parent chain and read the same configuration, so they converge, and the loser's write equals the winner's. The one value that would *not* converge is the clock read, which is exactly why `copy()` resolves before copying instead of letting each worker resolve its own.

**Test tagging** - measured rather than estimated: `Issue6266CommandTimeoutCoverageTest` runs in **1.158 s** in a full-suite `engine` run. That sits below the 1.5-3 s band where dozens of untagged classes already live (`HashIndexTest` 2.68 s, `SQLFunctionConvertTest` 2.17 s, `AsyncShutdownEscalationTest` 2.36 s), and 98 classes in the module run longer than 1.5 s untagged. The 20k-vertex graph is built once in `@BeforeAll` and each timeout test aborts in tens of milliseconds by construction, so setup is the whole cost. `@Tag("slow")` would misfile it.

**Description length** - agreed it's an outlier, and I'll leave it: this PR exists because the terse version was believed and wrong, and the paragraph is what makes the boundaries (polyglot, regex) checkable without a grep.

Verification: engine suite **12123 tests, 0 failures, 0 errors** on a clean run. (An earlier run in this session reported mass failures; that was my own doing - I had run `mvn compile` while the surefire JVM was live, rewriting `target/classes` underneath it, and every failure was `NoClassDefFoundError`-shaped. Re-run with the tree frozen is green.) You were right to flag checking that CI compiles cleanly.

lvca:
@claude The `Claude Code Review` workflow did not fire for `3e0b76d15` (no run exists for that SHA, while the three commits before it each got one - the `synchronize` event looks to have been dropped). Requesting a review of that commit here instead.

It contains: the `PairHashJoinOp` probe-loop fix you flagged, the wider guard sweep it prompted across all five `CountOp` implementations, the `MatchStep` per-batch hoist, and the two documentation notes (reserved `Long.MIN_VALUE` sentinel; why the check-then-set on the volatile deadline fields tolerates its race).

Please pay particular attention to the guard threading in `PropagateChainOp`'s self-loop-subtraction family and `AntiJoinChainOp`'s OLTP map builders, since those were threaded through several private helpers.

lvca:
Nit taken - `1002e8f54` moves `guard` to the last parameter of `buildNeighborRIDMap`/`buildTailCounts`, matching `precomputeBucketIds` and the rest of the sweep.

On the build you could not run: compiled clean, and the affected suites are green locally - 152 tests across `Issue6266CommandTimeoutCoverageTest`, `Issue6216AlgoWorkKnobBoundsTest`, the count-push-down suites (`OpenCypherCountOptimizationTest`, `CypherCountPushDownPreconditionsIssue5715Test`, `Issue5790OptionalMatchCountStarUndeclaredRelTypeTest`, `PartitionedTriangleOpCallerRunsTest`), `TimeoutStepTest` and `BasicCommandContextTest`. The full `engine` module was 12123 tests / 0 failures on `3e0b76d15`.

Your point about the small fixed-length loops is right and worth stating so nobody "fixes" them later: `for (final Arm arm : arms)` and `for (int h = 0; h < edgeTypes.length; h++)` iterate over the *pattern's* arm/chain length, which is bounded by the statement, not by the graph. Guarding those would cost a branch to bound something that is already bounded. Same for `walkArmOLTP`/`walkArmWithViews`: they run only inside an already-guarded per-anchor loop, so a guard of their own would tighten abort latency by less than one anchor's work.

Also worth noting for anyone reading the CI: the `integration-tests` failure on this branch is a port collision, not this change - all three errors are `Failed to bind to address 0.0.0.0:51122 ... Address already in use` while starting `GrpcServerPlugin`, and that lane runs with build failures ignored.

That closes every point raised across the four reviews. Ready from my side.

PR #6298, opened 2026-08-17
fix(#6296, #6289): the biggest allocation of an algo call is the one nobody was looking at

### LINKED ISSUE #6289: Allocation churn on the dense algo.* paths: a per-spur-node n x n mask in kShortestPaths, a boxed Integer[] sort in steinerTree
Two allocation-churn findings surfaced while reviewing PR #6285 (issue #6263), both deliberately left out of it: that PR bounds *peaks*, and neither of these raises a peak. They are performance questions, and both need a measurement rather than a guess - which is exactly why they do not belong in a PR about bounds.

## 1. `algo.kShortestPaths` reallocates an `n x n` mask per spur node

`AlgoKShortestPaths.java`, inside Yen's spur loop:

```java
for (int ki = 1; ki < k; ki++) {
  for (int i = 0; i < prevPath.length - 1; i++) {
    ...
    final boolean[][] removedEdges = new boolean[n][n];   // fresh every spur node
```

Only one is ever live, so the budget added in #6263 prices it correctly as a peak of one. But the *churn* is `k x pathLength` allocations of `n²` bytes each: at 1000 nodes and a 20-hop path with `k = 10`, that is ~200 allocations of 1 MB, ~200 MB through the young generation for a single call.

The obvious fix - hoist it and clear it between spur nodes - is not obviously a win, which is the whole point of filing this rather than doing it. `Arrays.fill` over `n²` booleans per iteration is real work too, and a fresh `boolean[n][n]` is something the JVM can thread-locally bump-allocate and collect young for close to nothing when it dies immediately. It needs measuring at a few graph sizes before either version can be called better.

Worth measuring alongside: the mask is dense but the thing it represents - the edges removed for this spur node - is tiny (at most the number of previously-found paths sharing the root). A `HashSet` of packed `long` keys, or a per-row sparse structure, may beat both dense options and is `O(paths)` rather than `O(n²)` in both time and space.

## 2. `algo.steinerTree` sorts its terminal pairs through a boxed `Integer[]`

`AlgoSteinerTree.java`:

```java
final Integer[] sortIdx = new Integer[pairs];
for (int i = 0; i < pairs; i++)
  sortIdx[i] = i;
Arrays.sort(sortIdx, (a, b) -> Double.compare(pW[a], pW[b]));
```

`pairs` is `t(t-1)/2`, so this is quadratic in the terminal count: at 2000 terminals it is ~2M boxed `Integer` objects (~48 MB where the `int` payload is 8 MB), plus a comparator call and two unboxings per comparison. CLAUDE.md asks for primitive arrays over boxed collections precisely on paths like this one.

#6285 prices it honestly rather than pretending it is primitive (that is what `BOXED_INTEGER_BYTES = 24L` is for), but pricing an inefficiency is not fixing it. The replacement is an index sort over primitives - the usual shapes are packing `(weight, index)` into a sortable `long[]` where the key permits it, or a small dual-pivot index sort keyed on `pW`. The catch is that the weights are `double`, so the packing trick needs care around sign and NaN; a hand-rolled index quicksort on `pW` is the straightforward version.

`AlgoMST.java:155` has the identical `Integer[] sortIdx = new Integer[ec]` shape for Kruskal's, sized by the edge count. Same fix, same measurement.

## 3. While in there: `MemoryBudget.reserve()` mutates before it throws

`AbstractAlgoProcedure.MemoryBudget.reserve()` adds to the running total and *then* checks the limit, so on the rejection path the object records a reservation that was never granted. Harmless today - the exception aborts the call and the budget is call-scoped and discarded - and the code review that spotted it agreed it is not a bug in the current design. But "reserved reflects what was granted" is the invariant a reader will assume, and it costs three lines to make true:

```java
final long total = saturatingSum(reserved, estimatedBytes);
if (total > limit)
  throw ...;
reserved = total;
```

Worth doing whenever this file is next open, so that a future retry or multi-attempt path cannot inherit the trap.

## Related

#6263, PR #6285, #6216, #6222.

### LINKED ISSUE #6296: algo.apsp materialises n² Result objects eagerly, so the call the working-memory budget capped at 64MB then allocates gigabytes it never sees
Surfaced while implementing #6263 (PR #6285), and deliberately not folded into it: that PR bounds the *working set* of an `algo.*` call, and this is the *result* set of the same call. Same failure mode, one level up, and for `algo.apsp` the result set is the larger of the two by more than an order of magnitude.

## The gap

`AlgoAPSP.execute()` builds its rows eagerly and only then hands back a stream:

```java
final List<Result> results = new ArrayList<>();
for (int i = 0; i < n; i++) {
  for (int j = 0; j < n; j++) {
    if (i == j || dist[i][j] >= INF)
      continue;
    final ResultInternal r = new ResultInternal();
    r.setProperty("source", graph.getRID(i));
    r.setProperty("target", graph.getRID(j));
    r.setProperty("distance", dist[i][j]);
    results.add(r);                        // up to n² - n of these, all live at once
  }
}
return results.stream();
```

One row per reachable ordered pair. On a connected graph that is `n² - n` rows, every one of them a `ResultInternal` with a backing map of three properties, all alive simultaneously before the caller sees the first row.

## Why the #6263 budget makes this worse, not better

#6263 caps the `n x n` distance matrix at `arcadedb.cypher.algoMaxWorkingMemory`, default `max(64MB, maxHeap/8)`. At the 64MB floor that admits roughly **n = 2890** (`n x (32 + 8n)` ≈ 67 MB).

A connected graph of 2890 nodes then produces about **8.3 million** `Result` objects. At a conservative hundred-plus bytes each - the object, its property map, the map's entries - that is well over a gigabyte, against the 64 MB the budget just finished enforcing on the matrix beside it.

So the budget is doing its job and the call still dies, for a reason the error message will not mention. That is precisely the shape #6065, #6216 and #6263 exist to eliminate: the largest allocation of the call sitting outside the thing that bounds allocations.

## Why eager at all

Nothing requires it. The four embedding procedures already return lazily from the same base class:

```java
return IntStream.range(0, n).mapToObj(i -> { ... });
```

`algo.apsp` can stream its pairs the same way - the distance matrix is complete before the first row is emitted, so a lazy `IntStream.range(0, n * n)` (or a flat-mapped nested range) filtered on `i != j && dist < INF` produces identical output with O(1) row-side memory. The consumer then decides how much it holds, which is the correct place for that decision.

Worth checking in the same pass whether the row count should also be counted against a bound: `n²` rows is a lot to return whatever the memory shape, and a `LIMIT`-aware or top-k form of the procedure may be the honest interface. That is a design question rather than a bug, and separable from the eager/lazy fix.

## Scope

The other `algo.*` procedures that build a `List<Result>` eagerly are fine as they stand - their row counts are bounded by a top-k (`algo.knn` by `topCount`), by a component count, or by a single node's neighbourhood. `algo.apsp` is the one whose row count is quadratic in the graph, which is what makes it the outlier rather than one of a family.

## Secondary, same area, much smaller

`AbstractAlgoProcedure.toEmbeddingList(double[])` boxes every element into a `List<Double>`: ~24 bytes per dimension against the 8 the `double` occupies, so the returned form of an embedding is three times the size of the matrix row #6263 prices. Because those procedures stream lazily, this is throughput and GC pressure rather than a peak - a 1M-node run at the default dimension 128 churns ~3 GB through the young generation for rows that die immediately - so it is a smaller and different problem from the one above. Fixing it means a primitive-backed `List<Double>` view rather than a copy, and is only worth doing if a profile says the boxing shows up.

## Related

#6263, PR #6285, #6216, #6065, #6289.

lvca:
Closes #6296, closes #6289.

Both were surfaced while reviewing PR #6285 (issue #6263) and deliberately left out of it: #6296 is the *result* set of the same call rather than its working set, and #6289 is churn rather than a peak. Neither is something that budget could have caught.

## #6296 - `algo.apsp` materialised n² rows before returning any of them

The distance matrix is reserved against `arcadedb.cypher.algoMaxWorkingMemory`; the rows beside it were not. At the 64 MB floor of that setting the matrix check admits n ≈ 2890, and a connected graph of 2890 nodes then built ~8.3M `ResultInternal` objects - each with a three-entry property map, all alive at once, well over a gigabyte - against the 64 MB the budget had just finished enforcing next to it. The budget did its job and the call still died, of an allocation its message never named.

Floyd-Warshall completes the matrix before the first row is emitted, so the rows are a pure projection of it and nothing required them to exist together. They now stream. The row-side footprint is O(1), and `CallStep` keeps the iterator rather than collecting it, so a `LIMIT` upstream costs what it says it costs.

Measured on the 200-node cycle in the regression test (`ThreadMXBean.getThreadAllocatedBytes`, per-thread so no GC can move it):

| | building the stream | draining the ~39 800 rows |
|---|---|---|
| before | 13 621 944 bytes | 256 bytes |
| after | ~83 000 bytes | the rows |

The issue also raises whether the *row count* should be bounded. Streaming answers it in the right place: the consumer decides how many it holds, and a `LIMIT` is now what bounds the call. A top-k form of the procedure would be a different interface, not a fix, so it is left alone.

The secondary finding in the issue - `toEmbeddingList` boxing every element - is deliberately **not** taken: a primitive-backed `List<Double>` view boxes on `get()` instead of up front, so on a consumer that reads each row once it moves the allocation rather than removing it. The issue itself says it is only worth doing if a profile says the boxing shows up, and this change is not that profile.

## #6289 - allocation churn on the dense paths

**1. The per-spur-node `n x n` mask in `algo.kShortestPaths`.** ~200 MB through the young generation for one k=10 call at 1000 nodes. The issue proposed hoisting the mask and clearing it, and asked for a measurement before choosing between that and reallocation.

Neither was needed - **the mask never had to be square**, which is the better alternative the measurement would not have found. Yen's keeps the edge `(p[i], p[i+1])` only for a path sharing the root `p[0..i]` with `prevPath`, and `prevPath[i]` **is** the spur node, so every entry it ever set had the same source. The other n-1 rows were allocated and zeroed only to prove they were empty. A single `boolean[nodeCount]` indexed by target says exactly as much, in O(nodeCount) rather than O(nodeCount²).

That also removes the `Arrays.fill`-vs-reallocate question the issue framed: the mask is allocated once for the whole call and cleared by naming the few entries it set - at most one per previously-found path - so clearing is O(paths), not O(nodeCount). The budget reservation follows the allocation and now prices two node-sized masks instead of a square one.

Measured on a 501-node graph driving ~150 spur nodes: **43 259 968 -> 4 411 528 bytes** for the call, against a weight matrix of 2 008 008 bytes allocated once.

**2. The boxed `Integer[]` Kruskal sort in `algo.steinerTree` and `algo.mst`.** Replaced by one shared primitive index sort, `AbstractAlgoProcedure.sortedIndexesByWeight`.

A bottom-up **merge** sort rather than the index quicksort the issue suggested, for two reasons that both matter here: the keys are user data, so a quicksort degrades to O(n²) on an input a caller chooses; and merging is stable, which is what keeps ties in exactly the order the TimSort behind `Arrays.sort(Integer[], Comparator)` produced. `Double.compare` is the comparison, so `NaN` and `-0.0` order identically too - the test asserts the new order against the old comparator sort on both awkward and randomised inputs.

Measured: **8.4 bytes per index against 24**. `BOXED_INTEGER_BYTES`, which existed only to price that boxing honestly, is gone with it, and `algo.steinerTree`'s reservation now prices five primitive arrays per terminal pair.

**3. `MemoryBudget.reserve()` mutating before it throws.** Committed only once granted. The invariant is now observable: a refused reservation no longer appears in the "on top of the N bytes this call already reserved" of the next message.

## Also, while in `AlgoKShortestPaths`

An unused `PriorityQueue`, and a dead incremental cost expression that computed the path weight four ways - two `Arrays.copyOf` and four `pathWeight` walks per spur node - and then threw the answer away in favour of the single `pathWeight(totalPath, ...)` on the line below it. Same file, same theme.

## Verification

Every allocation assertion was checked against the old code before being trusted: reverting each fix in turn makes the corresponding test red (13.6 MB build for #6296, 43.3 MB for the mask, a lying "already reserved" figure for the budget), so none of them is vacuous.

- `Issue6296AlgoAPSPLazyRowsTest` (4 tests) - the full n² - n row set is unchanged and each pair appears once; distances around the cycle are unchanged; a partially read stream still yields its tail; and the allocation split between building and draining.
- `Issue6289AlgoAllocationChurnTest` (7 tests) - k distinct paths in ascending weight (the correctness counterweight to collapsing the mask); the reservation message; the call's allocation against the weight matrix; the index sort against the boxed comparator sort it replaced, including NaN, ±0.0, the infinities and tie stability; MST and Steiner trees unchanged, pinned from both sides (the MST takes the two cheap bypass edges of every hop, the Steiner path takes the single direct hop instead); and the budget invariant.
- `Issue6263AlgoWorkingMemoryBudgetTest` updated for the smaller `kShortestPaths` reservation (400 -> 328 bytes on its 4-node graph).
- Full `com.arcadedb.query.opencypher` sweep: 4520 tests, 0 failures.

The two `ThreadMXBean` measurements are tagged `@Tag("performance")`, matching `RidScoreMinHeapTest`, and skip themselves on a JVM without per-thread allocation counters.

## Files changed

- `engine/.../algo/AbstractAlgoProcedure.java` - `sortedIndexesByWeight`, budget invariant, `BOXED_INTEGER_BYTES` removed
- `engine/.../algo/AlgoAPSP.java` - lazy rows
- `engine/.../algo/AlgoKShortestPaths.java` - node-sized spur masks, reservation, dead code
- `engine/.../algo/AlgoMST.java`, `engine/.../algo/AlgoSteinerTree.java` - primitive index sort
- `engine/src/test/.../algo/Issue6296AlgoAPSPLazyRowsTest.java` (new), `Issue6289AlgoAllocationChurnTest.java` (new), `Issue6263AlgoWorkingMemoryBudgetTest.java`

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AbstractAlgoProcedure.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AbstractAlgoProcedure.java
@@ -274,17 +274,6 @@ public abstract class AbstractAlgoProcedure implements CypherProcedure {
   protected static final long DOUBLE_BYTES = 8L;
 
   /**
-   * Heap cost of one entry of an {@code Integer[]}: the 8-byte reference the array holds plus the 16-byte boxed
-   * object it points at. Three times the cost of the {@code int} it carries, which is why an index array sorted
-   * through a {@code Comparator} - the only reason these arrays are boxed at all - is priced separately rather
-   * than folded into a primitive figure that would understate it.
-   * <p>
-   * Like {@link #MATRIX_ROW_OVERHEAD_BYTES} this is a heuristic for a budget check: below 128 the JVM hands out
-   * cached {@code Integer} instances and the true cost is the reference alone.
-   */
-  protected static final long BOXED_INTEGER_BYTES = 24L;
-
-  /**
    * Estimated heap footprint of a {@code new T[rows][columns]} of an element type {@code elementBytes} wide,
    * row headers included, in saturating {@code long} arithmetic.
    */
@@ -331,13 +320,19 @@ public abstract class AbstractAlgoProcedure implements CypherProcedure {
       if (limit < 0)
         // Negative means no limit: skip the bookkeeping too, so a disabled budget costs nothing.
         return;
-      final long alreadyReserved = reserved;
-      reserved = saturatingSum(reserved, estimatedBytes);
-      if (reserved <= limit)
+      // Committed only once granted. A refused reservation was never granted, so recording it would make
+      // `reserved` describe heap nobody is holding - and every later message quoting "the N bytes this call
+      // already reserved" would quote a figure that includes an allocation the call was refused. No current
+      // caller survives a refusal to observe that, since the exception aborts the call; this keeps the
+      // invariant true for the one that eventually does.
+      final long total = saturatingSum(reserved, estimatedBytes);
+      if (total <= limit) {
+        reserved = total;
         return;
+      }
       throw new IllegalArgumentException(getName() + "(): " + component + " would need " + describe(estimatedBytes)
           + " bytes (" + detail + ")"
-          + (alreadyReserved > 0 ? ", on top of the " + alreadyReserved + " bytes this call already reserved" : "")
+          + (reserved > 0 ? ", on top of the " + reserved + " bytes this call already reserved" : "")
           + ", more than the " + limit + " bytes allowed. Set "
           + GlobalConfiguration.CYPHER_ALGO_MAX_WORKING_MEMORY.getKey() + " to raise the limit");
     }
@@ -349,6 +344,62 @@ public abstract class AbstractAlgoProcedure implements CypherProcedure {
     private static String describe(final long estimatedBytes) {
       return estimatedBytes == Long.MAX_VALUE ? "over " + Long.MAX_VALUE : Long.toString(estimatedBytes);
     }
+  }
+
+  /**
+   * Returns {@code 0 .. count - 1} ordered by ascending {@code weights[i]}, without boxing an index.
+   * <p>
+   * Kruskal's is the reason this exists: both places that run it - {@code algo.mst} over the edge count and
+   * {@code algo.steinerTree} over the {@code t(t-1)/2} terminal pairs - need the edge indices in weight order,
+   * and the only ready-made way to sort an index array by an external key is
+   * {@code Arrays.sort(Integer[], Comparator)}. That costs 24 bytes per entry where the {@code int} it carries
+   * occupies 4, plus a comparator call and two unboxings per comparison, and {@code algo.steinerTree}'s entry
+   * count is quadratic in a terminal list the caller supplies: ~2M entries and ~48 MB of boxed
+   * {@code Integer} at 2000 terminals. Here it is 8 bytes per entry - the index array and the merge scratch -
+   * and a primitive comparison.
+   * </p>
+   * <p>
+   * A bottom-up merge sort rather than a quicksort, for two reasons that both matter on this path: the keys are
+   * user data, and a quicksort on an adversarially ordered key array degrades to O(n²) on inputs a caller
+   * chooses; and merging is stable, so equal weights keep index order exactly as the {@code TimSort} behind
+   * {@code Arrays.sort(Integer[], ...)} did. {@link Double#compare} is the comparison, so {@code NaN} and
+   * {@code -0.0} order exactly as they did through the comparator.
+   * </p>
+   *
+   * @param weights key per index; only the first {@code count} entries are read
+   * @param count   number of indices to order
+   */
+  protected static int[] sortedIndexesByWeight(final double[] weights, final int count) {
+    final int[] index = new int[count];
+    for (int i = 0; i < count; i++)
+      index[i] = i;
+    if (count < 2)
+      return index;
+
+    final int[] scratch = new int[count];
+    int[] from = index;
+    int[] to = scratch;
+    // Widths are stepped in long: at a count near Integer.MAX_VALUE the last doubling and the run bounds it
+    // produces overflow int, and a negative bound silently skips the merge instead of failing.
+    for (long width = 1; width < count; width <<= 1) {
+      for (long lo = 0; lo < count; lo += width << 1) {
+        final int left = (int) lo;
+        final int mid = (int) Math.min(lo + width, count);
+        final int end = (int) Math.min(lo + (width << 1), count);
+        int a = left, b = mid, out = left;
+        while (a < mid && b < end)
+          // Take the left run unless the right one is strictly smaller: that is what makes the sort stable.
+          to[out++] = Double.compare(weights[from[b]], weights[from[a]]) < 0 ? from[b++] : from[a++];
+        while (a < mid)
+          to[out++] = from[a++];
+        while (b < end)
+          to[out++] = from[b++];
+      }
+      final int[] swap = from;
+      from = to;
+      to = swap;
+    }
+    return from;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


lvca:
Follow-up for the review's one actionable note: the missing `algo.mst` working-memory reservation is now filed as #6300, with the Borůvka-over-CSR alternative that would remove the allocation rather than bound it. Kept out of this PR as agreed - it is pre-existing, and this PR is about the result set and about churn.

The other two notes I'm leaving as they are: the `com.sun.management.ThreadMXBean` FQN matches the `RidScoreMinHeapTest` precedent (as the review itself concludes), and centralising the `@Tag("performance")` rationale is explicitly gated on a third such test showing up - two occurrences is not yet an abstraction worth inventing.

Issue #6302, opened 2026-08-17
The four densest algo.* procedures have no WorkGuard, so an O(V³) run the memory budget explicitly admits cannot be interrupted or timed out


lvca:
Noticed while working on #6296 / #6289 (PR #6298). #6216 made the knob-driven loops abortable and #6266 extended `arcadedb.command.timeout` coverage; both reasoned that time, unlike memory, has no honest ceiling to pick and so a long run should be **abortable** rather than forbidden. That reasoning was applied to the procedures whose work is multiplied by a *knob*. It was never applied to the ones whose work is multiplied by the *graph*, which are the slowest in the package.

19 of the 68 `algo.*` procedures create a `WorkGuard`. These four do not:

| Procedure | Cost | Knob? |
|---|---|---|
| `algo.apsp` | O(V³) Floyd-Warshall | none - the graph alone sizes it |
| `algo.kShortestPaths` | O(k x pathLength x V²) Yen's over a dense matrix | `k`, deliberately left unbounded |
| `algo.steinerTree` | O(t x (V² + E)) Dijkstra per terminal | terminal list, caller-supplied |
| `algo.mst` | O(E log E) | none |

## Why `algo.apsp` makes the case on its own

The #6263 budget caps its distance matrix at `arcadedb.cypher.algoMaxWorkingMemory`, whose floor of 64 MB admits **n ≈ 2890**. So the budget's own answer to "is this graph acceptable?" is yes at 2890 nodes - and 2890³ is ~2.4 x 10¹⁰ inner iterations of the Floyd-Warshall triple loop. That is minutes of CPU on one query thread, during which:

- `Thread.interrupt()` does nothing - there is no interruption check in the loop;
- `arcadedb.command.timeout` does nothing - the deadline is never consulted;
- the client cancelling the query does nothing.

The budget and the timeout are meant to be complementary halves of the same guarantee, and here the memory half explicitly waves through a run the time half cannot stop. A user whose call the budget *rejects* gets an immediate, actionable error; a user whose call the budget *accepts* can lose a query thread for minutes with no way back. That is an odd place for the sharp edge to sit.

## Suggested fix, and a better framing

[...]

Issue #6301, opened 2026-08-17
algo.steinerTree attaches edge weights by iteration position, so a relTypes filter and the analytical view each silently change the tree it returns


lvca:
Found while working on #6289 (PR #6298) and verified with runnable reproductions before filing. Not a performance question like #6289 was - `algo.steinerTree` returns **wrong trees and wrong weights**, and it returns *different* ones depending on whether a Graph Analytical View happens to be present.

## Root cause

`AlgoSteinerTree.buildWeightedAdj` pairs weights with neighbours **by position**:

```java
final int[][] adj = graph.adjacency(Vertex.DIRECTION.BOTH, relTypes);   // built WITH relTypes
final double[][] adjW = buildWeightedAdj(graph, adj, weightProperty);
...
private double[][] buildWeightedAdj(...) {
  for (int i = 0; i < n; i++) {
    final Iterable<Edge> edges = graph.getVertex(i).getEdges(Vertex.DIRECTION.BOTH);  // NO relTypes
    final double[] tmpW = new double[adj[i].length];
    int pos = 0;
    for (final Edge e : edges) {
      ...
      if (pos < adj[i].length) {
        tmpW[pos] = ...;   // the pos-th EDGE's weight lands on the pos-th FILTERED NEIGHBOUR
        pos++;
      }
    }
    adjW[i] = tmpW;
  }
}
```

`adj[i]` is the neighbour list **after** the `relTypes` filter and in whatever order the backing store produced. `tmpW` is filled from an **unfiltered** `getEdges(BOTH)` walk in OLTP order. Nothing reconciles the two - `nbIdx` is computed and then only used as a skip test, never as the index to write at. So `adjW[i][j]` is the weight of "the j-th edge I happened to iterate", not "the weight of the edge to `adj[i][j]`".

That array is not just what gets reported - it is what Dijkstra runs on in step 1, so the *tree itself* changes, not only the `weight` column.

## Reproduction 1: a second edge type corrupts the weights

Three vertices on a `ROAD` path of weight 1 per hop, plus one `NOISE` edge of weight 999 hanging off X. `relTypes='ROAD'` should exclude `NOISE` entirely.

```cypher
MATCH (a:N {name:'X'}), (z:N {name:'Z'})
CALL algo.steinerTree([a,z],'ROAD','w') YIELD source, target, weight, totalWeight
```

```
[...]

Issue #6300, opened 2026-08-17
algo.mst builds three edge-sized arrays and an index sort with no working-memory reservation, the one dense algo path #6263 did not price


lvca:
Raised in the review of PR #6298 (issues #6296 / #6289) and deliberately not folded into it: it is pre-existing, and that PR is about the *result* set and about *churn*, not about closing a new hole in the budget.

## The gap

#6263 reserved the dense working set of every `algo.*` procedure that builds one against `arcadedb.cypher.algoMaxWorkingMemory`. `algo.mst` was not among them, and it builds a working set sized by the **edge** count:

```java
final int[]    eu = new int[edgeCount];
final int[]    ev = new int[edgeCount];
final double[] ew = new double[edgeCount];
...
final int[] sortIdx = sortedIndexesByWeight(ew, ec);   // order + merge scratch: two more int[ec]
```

That is 16 bytes per edge in the three parallel arrays plus 8 in the sort, so **24 bytes per edge**, none of it reserved. At 100M edges - a size ArcadeDB is built for - that is ~2.4 GB requested with no check and no error naming what asked for it, which is exactly the `OutOfMemoryError` shape #6065, #6216 and #6263 exist to replace with a client error.

`AlgoMST` also holds the `List<Vertex>` of every vertex and the `Map<RID, Integer>` index beside it, which is the OLTP loading path rather than anything the MST itself chose - worth pricing separately, or worth not doing at all (see below).

## Why it was missed

The procedures #6263 priced are the ones whose working set is sized by a **knob** (an embedding dimension, a terminal list) or is **quadratic in the node count** (a distance, similarity or capacity matrix). `algo.mst` is neither: it is linear in the edge count, which reads like the graph paying for itself. But linear in the edge count is not small - it is the largest linear dimension a graph has, usually an order of magnitude above the node count - and "linear" is not the criterion the budget uses. The criterion is whether the allocation has a ceiling the caller can predict, and this one does not.

`algo.minSpanningArborescence` and any other Kruskal/Borůvka-shaped procedure should be checked in the same pass rather than fixed one at a time; the review that found this one only looked at the files PR #6298 touched.

## Suggested fix, and a better alternative

[...]

PR #6306, opened 2026-08-17
fix(#6295, #6300, #6301, #6302): a weight belongs to its own edge, and the phase a call spends its time in is the one that has to be abortable

### LINKED ISSUE #6295: algo.hashgnn's MinHash phase is O(nodeCount × embeddingDimension²) and sits outside every checkpoint (follow-up to #6264)
Found while implementing #6264 (PR #6288), verified against `main` at `6b20fb1`.

#6216 and #6264 between them gave every iteration-shaped knob in `algo.*` a domain minimum and a checkpoint inside the loop it drives. `algo.hashgnn` got both. It is still not abortable, because **its dominant phase is not inside that loop.**

## Measured

`arcadedb.command.timeout` set to **1000 ms**, on a **2000-node** graph of degree 4:

```
CALL algo.hashgnn({embeddingDimension: 4096, iterations: 1, seed: 1}) YIELD node RETURN node
```

**Ran for 112,988 ms and returned a result.** The deadline was overshot by 113x and never observed — the call did not abort, it succeeded.

## Why the #6216/#6264 lens walked past it

Both issues hunted knobs whose value is *unbounded*. `embeddingDimension` is not one: #6065/#6214 capped it at `MAX_EMBEDDING_DIMENSION` (4096), so it looks handled. The trap is that the MinHash reduction runs **after** the `iterations` loop, and costs, per node:

```java
// AlgoHashGNN.java, after the guarded loop
for (int i = 0; i < n; i++) {              // <- no checkpoint
  for (int d = 0; d < embDim; d++) {
    for (int f = 0; f < numFeatures; f++) { ... }   // numFeatures = max(embDim * 4, 64)
```

`embDim × numFeatures` = **4 × embeddingDimension²** = 6.7e7 at the cap. That per-node figure is bounded — and then multiplied by an unbounded `nodeCount`. A bounded per-item cost times an unbounded item count is unbounded work, and here it is the *dominant* phase, not a tail: at 2000 nodes it is 1.3e11 inner iterations, which is the whole 113 seconds.

So the shape to look for is not "which knob has no ceiling" but **"which loop over nodeCount has no checkpoint"**, and the phases after the knob-driven loop are exactly where #6264 stopped looking.

## Fix

One line: `guard.checkPeriodically(i)` at the head of that loop. The `WorkGuard` already exists in the method (#6264 added it); the reduction simply sits outside its reach. Per-node work there is 6.7e7 operations at the cap, so the throttled form is more than fine — even the unthrottled `check()` would be free.

## Second instance, same shape, much smaller

`AlgoSLPA`'s post-processing loop (`freq.merge(...)` per remembered label, boxed) is also outside the guarded region and is O(nodeCount × iterations). It is indirectly bounded, because #6264 prices `nodeCount × iterations` against `arcadedb.cypher.algoMaxWalkMemory` — at the 64 MB default that caps it near 1.6e7 boxed merges (~1 s). It scales with the budget, though, so an operator who raises the budget re-opens it. Measured at degree 20000: 1086 ms against a 1000 ms deadline, i.e. real but nowhere near HashGNN's 113x.

Worth doing the sweep across the package rather than patching these two: the question for each procedure is whether *every* loop over `nodeCount` is covered, not just the one the knob drives.

## Not this issue

- `algo.hashgnn`'s `boolean[n][numFeatures]` allocation (n × 16384 bytes at the cap) is a heap concern and belongs to #6263 / PR #6285, which already touches this file.
- `AlgoSLPA.mostFrequent()` is O(degree²) per listener, so one supernode's turn is unabortable between per-node checkpoints. Same family as this, but graph-driven rather than dimension-driven; noting it here rather than filing separately since the sweep above should catch it.

## Related

#6216, #6264, PR #6288, PR #6222, #6263/PR #6285, #6266/PR #6291.

lvca:
Fixes #6295, #6300, #6301, #6302.

## #6301 - an edge weight was attached to whatever neighbour sat at that index

`AlgoSteinerTree.buildWeightedAdj` paired weights with neighbours **by iteration position**. `adj[i]` was built
with the `relTypes` filter applied and in the backing store's order; `tmpW` was filled positionally from an
*unfiltered* `getEdges(BOTH)` walk in OLTP order. `nbIdx` was computed and then used only as a skip test, never as
the index to write at. That array is what Dijkstra runs on in step 1, so the **tree itself** changed, not only the
`weight` column.

Both reproductions from the issue are pinned as tests and both were verified red before the fix:

- an excluded `NOISE` edge of weight 999 landed on an included `ROAD` edge → `weight: 999.0`, `totalWeight: 1000.0`
  for a tree that costs 2.0;
- with a Graph Analytical View present the same query returned a 51.0 tree where OLTP returned 2.0. A view is meant
  to be a transparent accelerator, so this is the worst failure mode it has.

**The fix removes the second traversal rather than repairing it.** `GraphData.weightedAdjacency(direction,
weightProperty, relTypes)` produces the neighbour list and its weights from **one** walk of the same edges - from
the view's columnar edge properties when it has them, from the edge records otherwise - so the pairing is correct
by construction and the two backings return the same multiset of `(neighbour, weight)` pairs. `buildWeightedAdj` is
deleted, and so is the old `GraphData.edgeWeights`.

The issue suggested reusing `edgeWeights(...)`. That could not be taken as-is, and the reason is the interesting
part: **`getEdgeProperty` returns null for `DIRECTION.BOTH`**, which is the direction `algo.steinerTree` needs, and
`getNeighborIds(i, BOTH, type)` is a sorted *merge* of the OUT and IN slices, so no positional mapping to the
column store exists for it at all. `weightedAdjacencyFromColumns` walks one slice per (type, direction) instead,
each of which *is* positional, and concatenates them.

The same helper was broken for its two existing callers, and both are fixed with it:

- **`algo.apsp`** with no `relTypes` filter on a view with edge properties produced an *empty* weight row against a
  non-empty neighbour row: an `ArrayIndexOutOfBoundsException` for a query that worked without the view. With
  several edge types it mismatched silently, because the adjacency is sorted across types while the weights were
  concatenated per type.
- **`algo.bellmanford`** fell back to a **unit weight for every edge** whenever the graph was CSR-backed and the
  view had no column for the property - ignoring `weightProperty` entirely, and only when a view existed. Its edge
  list is now parallel primitive arrays rather than `List<int[]>` + `List<Double>`, which the O(V x E) relaxation
  reads up to `V - 1` times.

**One provider-contract change.** `GraphTraversalProvider.hasEdgeProperties()` now means "the columns are
materialised **and** the positional mapping `getEdgeProperty` relies on is exact", and `GraphAnalyticalView` answers
`false` while a delta overlay is active: the columns are aligned with the base CSR's forward slots while
`getNeighborIds` serves the overlay's view of the node (deletions dropped, additions merged, list re-sorted), so
the n-th neighbour is not the n-th edge of the column store. `SQLFunctionAstar`, `SQLFunctionBellmanFord` and
`algo.dijkstra.singleSource` pair positionally too and gain the same protection: in that state they read the edge
records, which is slower and exact. Same "say unknown rather than guess" convention as `countEdgesBetween` and
`getMeanEdgesPerConnectedPair`.

## #6295 - the phase a call spends its time in was outside every checkpoint

`algo.hashgnn`'s MinHash reduction runs **after** the loop `iterations` drives, and costs
`4 x embeddingDimension²` operations per node - a bounded per-node figure multiplied by an unbounded node count.
Reproduced as the issue measured it: the deadline overshot by two orders of magnitude and the call *returned*.

The generalisable lesson, and the one applied here: **#6216 and #6264 hunted knobs with no ceiling;
`embeddingDimension` HAS one (#6065), which is exactly why the lens walked past it. Ask "which loop over the node
count has no checkpoint", not "which knob has no cap".** That sweep also found the pre-loop initialisation of
`algo.hashgnn`, `algo.fastrp` and `algo.graphsage`, and `algo.slpa`'s post-processing and its O(degree²)
`mostFrequent`. All now carry a checkpoint; the two-phase ones use a counter spanning both loops, because a
per-node check is too coarse on a wide embedding and a per-dimension clock read is too expensive on a narrow one.

## #6302 - a guard belongs wherever the work is unbounded, not only where a knob multiplies it

`algo.apsp` makes the case alone: the #6263 budget's 64 MB floor **admits n ≈ 2890**, and 2890³ is ~2.4e10
Floyd-Warshall iterations during which `Thread.interrupt()`, `arcadedb.command.timeout` and client cancellation all
did nothing. The memory half of the guarantee was waving through a run the time half could not stop.

Taking the issue's "better framing" rather than its four-procedure list, all 68 procedures were swept with the
question "can one call of this run for minutes on an accepted input?". **22 are now abortable**: `apsp`,
`kShortestPaths`, `steinerTree`, `mst`, `bellmanford`, `betweenness`, `closeness`, `harmonic`, `eccentricity`,
`maxFlow`, `msa`, `knn`, `hierarchicalClustering`, `clique`, `allSimplePaths`, `kTruss`, `triangleCount`,
`localClusteringCoefficient`, `densestSubgraph`, `bipartiteMatching`, `voteRank`, `richClub`.

Procedures making a single O(V + E) pass are deliberately left alone: one pass costs what loading the graph and
emitting the rows already cost, so a checkpoint there bounds nothing the surrounding pipeline does not. One gap is
named rather than hidden: `algo.localClusteringCoefficient`'s CSR path delegates to a `GraphAlgorithms` kernel that
counts triangles across a thread pool, and `WorkCheckpoint` is specified to be called between iterations on the
calling thread - covering it is a change to how that kernel partitions its work.

## #6300 - `algo.mst`'s edge arrays are the one dense path #6263 did not price

24 bytes per edge across `eu`/`ev`/`ew` and the index sort, unreserved. #6263's criterion was "knob-sized or
quadratic in the node count", and edge-linear reads like the graph paying for itself - but the edge count is the
largest linear dimension a graph has, and 100M edges is ~2.4 GB with nothing between it and the allocator.

**The issue asked for the reservation to be made before the pass-1 count loop, which cannot be done as stated:**
nothing knows the edge count before the walk, and every cheap upper bound (`countType` over the edge types) is
itself a scan. What it wants is achievable another way, and that is what is implemented: the new
`MemoryBudget.capacityFor(bytesPerItem)` turns the budget into a bound the walk carries, so the count stops the
moment it passes what the budget allows. The refusal still goes through `reserve()`, so the message names the same
component and the same setting either way. The issue's Borůvka-over-CSR alternative is not taken here - it would
remove the allocation rather than bound it, but it changes tie-breaking and result ordering, which is a different
change from closing a hole in the budget.

`algo.mst`'s `List<Vertex>` + `Map<RID, Integer>` are noted in the issue and left alone on purpose: that is the
shared OLTP loading path every procedure uses, so pricing it in one procedure would be inconsistent. It belongs in
`loadGraph`, as one change, if it is wanted.

## Verification

- `Issue6301AlgoSteinerTreeWeightAlignmentTest` (6) - both reproductions, the analytical-view comparison with an
  explicit `CSR_ACCELERATED_VAR` assertion so it cannot silently pin the OLTP path twice, parallel edges, the
  bellmanford unit-weight fallback, the apsp all-types crash. **4 of 6 verified red on the pre-fix code.**
- `Issue6302AlgoGraphDrivenWorkGuardTest` (42) - every one of the 22 procedures aborts on a pending thread
  interrupt and still returns its result without one, plus a `@Tag("slow")` deadline test on `algo.apsp` at 1500
  nodes. **Verified: with the Floyd-Warshall checkpoint removed the same test runs the loop to completion and
  returns.**
- `Issue6295AlgoUnguardedPhaseTest` (2) - the MinHash reduction aborts on a deadline that the single message-passing
  round comfortably survives. **Verified red (no exception, ran to completion) with the checkpoint removed.**
- `Issue6300AlgoMSTEdgeBudgetTest` (5) - refusal, the message naming the count the walk stopped at, a budget that
  fits, a disabled budget, and the MST itself unchanged. **2 of 5 verified red on the pre-fix code.**

Only `algo.hashgnn`'s reduction is pinned as a binary abort, and the test javadoc records why the others are not:
SLPA's propagation and its post-processing are both O(nodeCount x iterations) and measured at the same order, so
any threshold between them would be measuring the fixture rather than the checkpoint. They are covered by the
no-regression test instead.

Full runs: 2541 tests across `Algo*`/`Issue62*`/`Issue63*`/`Cypher*`/`GraphAnalyticalView*`/`SQLFunction*`, 1639
across `com.arcadedb.graph.**` and `com.arcadedb.function.**`, plus the `slow` lane for the touched classes. All
green.

## Follow-up worth filing

`GraphAnalyticalView.getEdgeProperty`'s positional contract is broken by an active overlay for *any* caller, not
only these. This PR makes the four known callers fall back safely via `hasEdgeProperties()`, but the SPI method
itself still returns a value addressed against the base CSR while `getNeighborIds` returns the overlay's list. A
provider-side fix (an overlay-aware property lookup, or an explicit "unknown" return) would close the shape rather
than the instances.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2574,7 +2574,7 @@ Two more procedures were reading weights through the same broken helper and are 
   when a view existed. It also now builds its edge list in parallel primitive arrays rather than a
   `List<int[]>` and a `List<Double>`, which the O(V x E) relaxation loop reads up to `V - 1` times.
 
-**One provider-contract change comes with it.** `GraphTraversalProvider.hasEdgeProperties()` now means "edge
+**Two provider-contract changes come with it.** `GraphTraversalProvider.hasEdgeProperties()` now means "edge
 property columns are materialised **and** the positional mapping `getEdgeProperty` relies on is exact", and
 `GraphAnalyticalView` answers `false` while a delta overlay is active. The columns are aligned with the base
 CSR's forward edge slots, while `getNeighborIds` serves the overlay's view of the node - deletions dropped,
@@ -2583,6 +2583,17 @@ column store. Callers (`algo.apsp`, `algo.bellmanford`, `algo.steinerTree`, `SQL
 `SQLFunctionBellmanFord`, `algo.dijkstra.singleSource`) now read the edge records in that state, which is
 slower and exact, instead of a weight belonging to another edge. This is the same "say unknown rather than
 guess" convention `countEdgesBetween` and `getMeanEdgesPerConnectedPair` already use.
+
+The second is `servesEdgeProperty(propertyName, edgeTypes...)`, which is the question a weighted algorithm
+actually has to ask. `hasEdgeProperties()` answers only "are there edge property columns at all", so a view built
+with `.withEdgeProperties("distance")` answered yes to a call asking for `cost` - and since a missing column and
+an edge with no value both come back as `null`, the caller could not tell them apart and treated the entire graph
+as unweighted. That is the same wrong answer as a misaligned weight, reached from the other direction, and it
+silently inverted results: on a graph where `X-Y-Z` costs 2.0 and the direct `X-Z` hop costs 50.0, unit weights
+make the 50.0 hop win. `algo.apsp`, `algo.bellmanford`, `algo.steinerTree`, `algo.dijkstra.singleSource`,
+`SQLFunctionAstar` and `SQLFunctionBellmanFord` now all gate on it and fall back to the edge records otherwise.
+`SQLFunctionBellmanFord` additionally passed `null` as the edge type to `getEdgeProperty`, which can never
+resolve a column store, so its CSR path had been using a unit weight for every edge unconditionally.
 
 [#6301](https://github.com/ArcadeData/arcadedb/issues/6301)
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -2542,3 +2542,113 @@ the busiest single database's share, which is exactly what `pageFlushQueue` boun
 alert on.
 
 [#6281](https://github.com/ArcadeData/arcadedb/issues/6281)
+
+## An edge weight belongs to the edge it was read from, whichever backing answered (#6301)
+
+`algo.steinerTree` paired weights with neighbours **by iteration position**. The adjacency list was built with
+the `relTypes` filter applied and in the backing store's order; the weight array beside it was filled
+positionally from an *unfiltered* `getEdges(BOTH)` walk in OLTP order. Nothing reconciled the two, so
+`adjW[i][j]` was "the weight of the j-th edge I happened to see", not "the weight of the edge to `adj[i][j]`".
+That array is what Dijkstra runs on in step 1, so the **tree itself** moved, not only the `weight` column.
+
+Two shapes made it visible. An edge type the caller excluded still lent its weight to an included edge: three
+vertices on a `ROAD` path of weight 1 per hop plus one `NOISE` edge of weight 999 hanging off the first
+returned `weight: 999.0` and `totalWeight: 1000.0` for a tree that costs 2.0. And the same query answered
+differently depending on whether a Graph Analytical View happened to exist, because the CSR neighbour order
+need not match the OLTP one - a view is meant to be a transparent accelerator, and here its mere presence made
+the procedure prefer a 51.0 tree over the 2.0 one.
+
+The fix removes the second traversal rather than repairing it. `GraphData.weightedAdjacency(direction,
+weightProperty, relTypes)` produces the neighbour list and its weights from **one** walk of the same edges -
+from the view's columnar edge properties when it has them, from the edge records otherwise - so the pairing is
+correct by construction and the two backings return the same multiset of `(neighbour, weight)` pairs.
+
+Two more procedures were reading weights through the same broken helper and are fixed with it:
+
+- **`algo.apsp`** with no `relTypes` filter on a view that materialises edge properties produced an *empty*
+  weight row against a non-empty neighbour row, i.e. an `ArrayIndexOutOfBoundsException` for a query that
+  worked perfectly without the view. With several edge types it mismatched instead of failing, because the
+  adjacency is sorted across types and the weights were concatenated per type.
+- **`algo.bellmanford`** fell back to a **unit weight for every edge** whenever the graph was CSR-backed and
+  the view had no column for the requested property - silently ignoring `weightProperty` altogether, and only
+  when a view existed. It also now builds its edge list in parallel primitive arrays rather than a
+  `List<int[]>` and a `List<Double>`, which the O(V x E) relaxation loop reads up to `V - 1` times.
+
+**One provider-contract change comes with it.** `GraphTraversalProvider.hasEdgeProperties()` now means "edge
+property columns are materialised **and** the positional mapping `getEdgeProperty` relies on is exact", and
+`GraphAnalyticalView` answers `false` while a delta overlay is active. The columns are aligned with the base
+CSR's forward edge slots, while `getNeighborIds` serves the overlay's view of the node - deletions dropped,
+additions merged in, the whole list re-sorted - so the n-th neighbour of that list is not the n-th edge of the
+column store. Callers (`algo.apsp`, `algo.bellmanford`, `algo.steinerTree`, `SQLFunctionAstar`,
+`SQLFunctionBellmanFord`, `algo.dijkstra.singleSource`) now read the edge records in that state, which is
+slower and exact, instead of a weight belonging to another edge. This is the same "say unknown rather than
+guess" convention `countEdgesBetween` and `getMeanEdgesPerConnectedPair` already use.
+
+[#6301](https://github.com/ArcadeData/arcadedb/issues/6301)
+
+## The phase a graph-algorithm call spends its time in is now abortable too (#6295, #6302)
+
+#6216 and #6264 gave every iteration-shaped `algo.*` knob a checkpoint inside the loop it drives, on the
+principle that for time there is no honest ceiling to pick, so a long run should be **abortable** rather than
+forbidden. Both picked their procedures by asking "does it have an iteration knob?". Two blind spots follow
+from that question, and this release closes both.
+
+### The dominant phase is not always the loop the knob drives (#6295)
+
+`arcadedb.command.timeout` set to 1000 ms, on a 2000-node graph:
+`CALL algo.hashgnn({embeddingDimension: 4096, iterations: 1})` **ran for 112,988 ms and returned a result** -
+the deadline overshot by 113x and never observed. `algo.hashgnn`'s MinHash reduction runs *after* the
+`iterations` loop and costs `4 x embeddingDimension²` operations per node, and that bounded per-node figure is
+then multiplied by an unbounded node count. The lens missed it because `embeddingDimension` *has* a ceiling
+(#6065 capped it at 4096), which made it look handled.
+
+The question that finds this shape is **"which loop over the node count has no checkpoint"**, not "which knob
+has no cap". Asking it across the package also found the pre-loop initialisation of `algo.hashgnn`,
+`algo.fastrp` and `algo.graphsage`, and `algo.slpa`'s post-processing and its O(degree²) `mostFrequent`. All
+now carry a checkpoint.
+
+### A guard belongs wherever the work is unbounded, not only where a knob multiplies it (#6302)
+
+The four densest procedures in the package had no guard at all, and `algo.apsp` makes the case on its own: the
+#6263 working-memory budget caps its distance matrix at `arcadedb.cypher.algoMaxWorkingMemory`, whose 64 MB
+floor **admits n ≈ 2890**, and 2890³ is ~2.4e10 iterations of the Floyd-Warshall triple loop. That is minutes
+of CPU on one query thread during which `Thread.interrupt()`, `arcadedb.command.timeout` and the client
+cancelling all did nothing. The budget and the timeout are meant to be complementary halves of one guarantee,
+and the memory half was waving through a run the time half could not stop.
+
+A sweep over all 68 procedures asking "can one call of this run for minutes on an accepted input?" settled the
+rest in one pass. Every procedure whose dominant loop is superlinear in the graph is now abortable:
+`algo.apsp`, `algo.kShortestPaths`, `algo.steinerTree`, `algo.mst`, `algo.bellmanford`, `algo.betweenness`,
+`algo.closeness`, `algo.harmonic`, `algo.eccentricity`, `algo.maxFlow`, `algo.msa`, `algo.knn`,
+`algo.hierarchicalClustering`, `algo.clique`, `algo.allSimplePaths`, `algo.kTruss`, `algo.triangleCount`,
+`algo.localClusteringCoefficient`, `algo.densestSubgraph`, `algo.bipartiteMatching`, `algo.voteRank` and
+`algo.richClub`.
+
+Procedures that make a single O(V + E) pass are deliberately left alone: one pass costs what loading the graph
+and emitting the rows already cost, so a checkpoint inside it would bound nothing the surrounding pipeline does
+not. One gap is named rather than hidden: `algo.localClusteringCoefficient`'s CSR path hands the whole
+computation to a `GraphAlgorithms` kernel that counts triangles across a thread pool, and the `WorkCheckpoint`
+hook is specified to be called between iterations on the calling thread - covering that kernel is a change to
+how it partitions its work, not a checkpoint that can be dropped in.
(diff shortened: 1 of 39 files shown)


lvca:
Thanks - actioned 8 of the 11 points in dfb2d8b. Three declined, with reasons.

**1. `hasEdgeProperties()` is keyed by "some property", not the requested one — actioned, and it was worse than the review found.**

You are right that this is the same failure class as #6301 arrived at from the other direction, and the new abstraction is the place to close it. New `GraphTraversalProvider.servesEdgeProperty(propertyName, edgeTypes...)` resolves "all types" against `getMaterializedEdgeTypes()` and checks `hasEdgeProperty(type, property)` for each. It is now the gate in `GraphData.weightedAdjacency`, and also in the three non-algo callers from point 7 — `algo.dijkstra.singleSource`, `SQLFunctionAstar`, `SQLFunctionBellmanFord` — so the check lives in one place rather than being open-coded four times.

Chasing it into those callers turned up a live instance: **`SQLFunctionBellmanFord` passed `null` as the edge type to `getEdgeProperty`**, which can never resolve a column store, so on any CSR-backed graph its weighted path had been using a unit weight for *every* edge unconditionally — `weightProperty` ignored outright. Fixed by resolving a single materialized type and gating on it.

`aViewThatMaterialisedADifferentPropertyDoesNotMakeTheGraphUnweighted` pins it and was **verified red on the coarse gate**: returns 1.0 (the direct hop at unit weight) instead of 2.0.

**2. `weightedAdjacency*` has no checkpoint — actioned.** Correct, and the inconsistency with `AlgoMST` in the same PR is the tell. `weightedAdjacency` now takes the `WorkGuard`. The record path throttles by **edge**, not by vertex, for the multigraph reason you give.

**3. `AlgoMST` throttles by vertex — actioned, and it is a real hole, not a granularity choice.** With `algoMaxWorkingMemory=-1` the budget never fires, so nothing bounds it; both passes now use a running edge counter. `AlgoAPSP`'s analogous loop is a different case — it reads an already-materialised `int[]`, so the per-vertex throttle there costs at most one array scan, and the expensive part of that phase moved into `weightedAdjacency`, which is now guarded per edge.

**4. `AlgoKTruss` setup runs before the first check — actioned**, in both `computeFullTrussDecomposition` and `execute()` (the same two phases exist in each): the neighbour bit matrix and the O(E) initial support computation.

**5. `AlgoBellmanFord`'s negative-cycle pass — actioned.** You are right that the consistency argument wins: the relaxation pass right above it is also a single O(E) pass and does get a check.

**6. No test for `edmonds()`'s recursion — actioned.** `algo.msa` *is* in `Issue6302AlgoGraphDrivenWorkGuardTest`'s interrupt table (the grep missed it because the procedure is named `algo.msa`, not "arborescence"), but you are right about what that fixture does *not* reach: on a directed cycle each non-root vertex's cheapest incoming edge already forms a path from the root, so nothing contracts and the recursion never runs. Added `msaStaysAbortableThroughItsCycleContractionRecursion` over a graph where R reaches A and B at 10 while A and B reach each other at 1 — the cheapest-incoming selection is then a 2-cycle that must be contracted. The `11.0` total is itself the proof the contraction happened; without it the answer is 20.0.

**7. No test for the overlay narrowing — actioned.** `anActiveOverlayDoesNotLendItsWeightsToTheWrongEdges` builds a `SYNCHRONOUS` view, asserts `hasEdgeProperties()` is true, commits a change the view absorbs into its delta overlay, asserts it flips to false, and checks `algo.bellmanford` still returns the record weights. The two assertions around the commit are what make it non-vacuous — a test that never actually produced an overlay would fail the first one.

---

**8. `AlgoKShortestPaths`/`AlgoMinSpanningArborescence` hand-roll the pattern — declined, with a correction.** Worth separating two things: neither carries the alignment bug. Both resolve the neighbour by index (`ridToIdx.get(e.getIn())`) rather than by iteration position, and neither touches the CSR at all — they load vertices manually. So moving them onto `weightedAdjacency` would not fix anything; it would *newly* route them through a Graph Analytical View, which changes neighbour ordering and result ordering for two procedures this PR has no other reason to touch. That is a CSR-acceleration change and belongs in its own PR with its own regression evidence. The consolidation argument is real, though — noted as a follow-up alongside the one already in the description.

**10. No structural "checked iterator" primitive — declined.** A shared primitive would have to pick one throttling policy, and the placement genuinely differs per loop: unthrottled `check()` where one iteration is already O(V) (`algo.apsp`'s matrix row, a Floyd-Warshall `i` step, a whole BFS), `checkPeriodically` on a running counter where one iteration is O(1) and the *product* of two loops is what is unbounded (`algo.hashgnn`'s reduction, `algo.mst`'s edge walk). Baking one into `GraphData.adjacency` would explain the mechanism but not the varying part, which is the part that is easy to get wrong. Same conclusion #6264 reached over three review rounds. The general form worth building is a lint, not an API.

**11. Repeated non-atomic `snapshot` reads — declined as flagged.** Agreed it is pre-existing and unchanged in kind: each read already returns a self-consistent snapshot, and no invariant here spans two of them (a torn read across `servesEdgeProperty` and the per-node calls yields the columnar path over a newer snapshot, which is exactly what a concurrent rebuild produces anyway). Widening it to a single captured read would mean pinning a snapshot across a whole graph walk, which is a different lifetime contract than the lock-free design currently offers.

Re-ran after the changes: 1631 tests across `Algo*`/`Issue62*`/`Issue63*`/`GraphAnalyticalView*`/`SQLFunction*`/`com.arcadedb.function.**`, 1966 across `com.arcadedb.graph.**`/`Cypher*`, plus the `slow` lane. All green.

lvca:
Both findings confirmed and fixed in 26f7516. You were right that the two `SQLFunction*` call sites still carried the bug class the rest of the PR closes, and the `BOTH` case is worse than "reachable" — it is the **default**.

**Verified both, red-before-fix, on the round-1 code:**

- `bellmanFord(src, dst, 'weight')` — three arguments, so `parseDirectionParam` gives `BOTH` — returned the two-hop `A-C-D` instead of the weighted `A-B-C-D` the moment a view existed. Every edge at 1.0, `weightProperty` ignored.
- `astar(..., {direction: 'BOTH'})` did the same at `MIN` (0.0), i.e. every edge free, so all paths tied and whichever was found first won.

**The fix takes your suggestion of routing through the one-walk builder, but pushes it one level lower.** `GraphData.weightedAdjacency` lives in the OpenCypher procedures package and is shaped for a whole-graph build, which is the wrong shape for `astar` — it expands one node at a time and must not build the graph per expansion. So the pairing logic moved onto the SPI instead, where `getEdgeProperty`'s positional contract already lives:

```java
NodeEdgeWeights edgeWeightsOf(int nodeId, Vertex.DIRECTION direction, String propertyName,
                              double defaultWeight, String... edgeTypes)
```

It splits `BOTH` into `{OUT, IN}` and takes one slice per (type, direction) — each of which *is* positional — and returns `null` unless the provider can serve that property for every type in play. `weightedAdjacencyFromColumns`'s copy of the logic is deleted and now calls it; so do both SQL functions. That is one method that knows how to pair a CSR edge with its weight, rather than the four hand-rolled copies that existed when this PR opened, which is what stops the three from drifting apart again. `defaultWeight` is a parameter because the callers genuinely disagree: `astar` defaults a missing weight to `MIN` (0.0), the others to 1.0.

**`algo.dijkstra.singleSource` needed nothing beyond round 1's gate** — worth stating since it looks like a fourth instance. `GraphAlgorithms.dijkstraSingleSource` walks the forward and backward CSR arrays itself using the column index, never through `getEdgeProperty`, so it has no `BOTH` hazard; the only thing it was missing was the per-property check, which it got in `dfb2d8b`.

**On the missing coverage:** you are right that the CSR branch of both functions was untested outright. `Issue6301CsrEdgeWeightAlignmentTest` covers it, and the fixture is chosen so the test cannot pass by accident — `A-B-C-D` at 1.0 per hop against a direct `A-C` at 100.0 means the weighted answer is the *four*-vertex path and the unit-weighted answer is the *three*-vertex one. Getting the weights wrong changes the path, not just a number. It also pins the "view materialises a different property" case for `bellmanFord`, and asserts the OLTP result first in every test so a view that silently stopped being used could not make it pass.

**Style nit** — fixed, `com.arcadedb.graph.Vertex` is imported in both test classes.

**Maintainability note** — agreed, and worth being precise about why it is not addressed here. The gap is not that a checkpoint is easy to forget; it is that the right *placement* is a per-loop judgement (unthrottled where one iteration is already O(V), throttled on a running counter where the unbounded thing is the product of two loops), so a base-class hook would standardise the part that is already easy and leave the part that is hard. The enforceable version is a lint over the package — "a loop over `nodeCount` or `edgeCount` inside `execute()` with no `guard` reference in its body" — which is a build-tooling change rather than an API one. Noted as a follow-up alongside the two already in the description.

Re-ran: 1634 tests across `Algo*`/`Issue62*`/`Issue63*`/`SQLFunction*`/`GraphAnalyticalView*`/`com.arcadedb.function.**`, 1969 across `com.arcadedb.graph.**`/`Cypher*`, plus the `slow` lane. All green.

Issue #6267, opened 2026-08-16
Follow-ups from #6221: a self-poisoning membership test, an unmeasured timeout, an invisible address ambiguity, and two cleanups


lvca:
Five findings from #6221 / PR #6226, filed together because they are all small and share an origin. Split any of them out if you would rather track them separately - only item 1 is a defect, the rest are a measurement to collect, an observability gap, a cleanup, and a docs update.

## 1. `DynamicMembershipTest` poisons the cluster it shares, so a different method fails on every run

`DynamicMembershipTest` (`ha-raft/src/test/java/com/arcadedb/server/ha/raft/DynamicMembershipTest.java`) runs its methods against one 3-node cluster and removes peers from it without putting them back:

- `removePeerDecreasesClusterSize` (`:50`) removes a non-leader and asserts 2 live peers.
- `removePeerRefusedWhenItWouldBreakQuorum` (`:66`) opens with `assertThat(raftServer.getLivePeers()).hasSize(3)` (`:71`), then removes one, then force-removes another, leaving **1**.
- `getLivePeersReturnsAllConfiguredPeers` (`:40`) asserts 3, and `registerPeerDisplayNameUpdatesExistingPeer` (`:114`) reads the peer set as well.

So every method here both depends on the peer set and mutates it, and JUnit's method order decides which one loses. Three runs on my machine produced three different failures, one of them on **pristine `main` at `d75174c92` with no local changes**, which is what rules out the PR that found it:

```
DynamicMembershipTest.removePeerRefusedWhenItWouldBreakQuorum:68  (expected 3 live peers)
DynamicMembershipTest.registerPeerDisplayNameUpdatesExistingPeer:116
DynamicMembershipTest > endTest > checkDatabasesAreIdentical: DatabaseAreNotIdentical Types: DB1 5 <> DB2 15
```

The third is the interesting one: a class that ends with a 1-peer cluster fails the base class's `endTest` consistency check, which is a failure attributed to whatever ran last rather than to the test that removed the peers.

Suggested fix: give the membership tests a per-method cluster (or re-add the removed peers in an `@AfterEach`), and `@Order` them only if the shared-cluster shape has to stay. This is very likely contributing to the chronic `ha-integration-tests` redness of #5668 / #5702 - it is a *deterministic* failure whose visibility depends on ordering, not a timing flake, so it will keep costing that lane signal until the shared state is fixed.

[...]

PR #6307, opened 2026-08-17
fix(#6297, #6280, #6279, #6270): a teardown that deletes what the test wrote, a warning that speaks again, and bounds that say what they are for

### LINKED ISSUE #3216: Creating a relation with CREATE/MERGE is taking a lot of time
Hello,

Probably related to this old issue (but it was for old engine, here it's new cypher engine) : https://github.com/ArcadeData/arcadedb/issues/3148

Query : 
```CYPHER

UNWIND $batch as row
MATCH (a),(b) WHERE ID(a) = row.source_id and ID(b) = row.target_id 
MERGE (a)-[r:`in`]->(b)RETURN a, b, r
```

Batch had to be reduced to 1 because it was tooo slow

`{"batch": [{"source_id": "#1:0", "target_id": "#4:0", "features": {}, "relation_type": "in"}]}`

Takes something like 4 seconds.

DB dump (not backup button because it's broken) :

[NEW_ORANO.zip](https://github.com/user-attachments/files/24831890/NEW_ORANO.zip)

### LINKED ISSUE #6270: The server and network suites still bound wall clock the way the engine suite did before #6260
Follow-up to #6260 / #6268, raised in that PR's review.

#6260 fixed the engine suite: 44 elapsed-time assertions now measure through `StallAwareStopwatch`, which discounts the JVM-wide stall (stop-the-world pauses and CPU starvation) observed inside the measured window, so a bound can stay tight without being tripped by a pause late in a shared-JVM run.

The same flake mode applies to every other module that runs a suite in one JVM, and those were deliberately left out of the scope of that PR. `grep` over `server/src/test/java` and `network/src/test/java` turns up:

```
server/.../Issue5470BatchStreamStallIT.java
server/.../HTTPAuthSessionIT.java              (2 sites)
server/.../Issue5470BatchErrorDeliveryIT.java
server/.../PostBatchHandlerIT.java
server/.../Issue5418ShutdownHookDeadlockTest.java
network/.../RemoteHttpComponentTest.java
```

The server module is a worse environment for this than the engine one, not a better one: its tests bind real ports, start real servers and talk HTTP, so they already sit closer to their bounds, and `Issue5470BatchStreamStallIT` is a known-red lane.

## What to do

`StallAwareStopwatch` and `JvmStallMonitor` live in the `engine` test-jar, which `server` and `network` already consume for `TestHelper`, so nothing needs moving - the work is per-assertion classification, the same pass #6268 did:

1. **Bounding a named timeout** - a tripwire between a bounded operation and an unbounded one. `assertGaveUpWithin`; generous is free, because a wider bound cannot turn a passing run red.
2. **A complexity claim** - the bound IS the assertion and has no other practical expression. `assertStayedUnder`; loosening it deletes the test.
3. **A lower bound, or a short wait expected to time out** - leave alone. A stall only makes those more true.

`@Timeout` also wants a look while in there: it is plain wall clock, JUnit gives no way to discount it, so it has to be sized as a hang detector rather than as a latency bound.

The rule is written up in `CLAUDE.md` under the testing guidance.

## Worth checking while doing it

The engine pass found its 44 sites with a `grep` for `assertThat(elapsed…)`, which missed four more that assert on the expression inline or name the local something else - one of them being a test named in the original issue as a motivating failure. Sweep on `System.currentTimeMillis() - ` / `System.nanoTime() - ` instead, and expect to filter out printlns and `@Tag("benchmark")` classes by hand.

### LINKED ISSUE #6279: issue3216_performanceComparisonMatchStrategies asserts only that two durations are not negative
Found while sweeping wall-clock assertions for #6260.

`OpenCypherMatchEnhancementsTest.issue3216_performanceComparisonMatchStrategies` is named for, and documented as, a comparison of two MATCH strategies:

```java
// Issue #3216: benchmark comparing single MATCH (a),(b) WHERE id-pair vs split MATCH clauses;
// expected to favour split form.
```

The comparison it exists to make is commented out, and what replaced it is a tautology:

```java
// Method 2 should be significantly faster
// (commenting out the assertion for now since we're diagnosing the issue)
// assertThat(duration2).isLessThan(duration1);
assertThat(duration1).isGreaterThanOrEqualTo(0L);
assertThat(duration2).isGreaterThanOrEqualTo(0L);
```

An elapsed time is never negative, so both lines pass unconditionally and neither can ever go red. The test is not entirely empty - the `rs.hasNext()` checks inside the two transactions do confirm both query forms return a row - but the property in the name, that the split form is the faster one, is not tested at all. Anyone reading the suite for coverage of #3216 will believe it is covered.

## Why now

The reason the assertion was disabled ("diagnosing the issue") points at the shape that is hardest to assert: a **relative** comparison of two wall-clock measurements. A stop-the-world pause landing inside either one flips the comparison in whichever direction it fell, which is exactly the flake #6260 was filed about, and is why an assertion like this gets commented out rather than fixed.

That tool now exists. `StallAwareStopwatch.jvmStallNanos()` can be sampled either side of each measurement so both durations are compared with the JVM's own stalls taken out - `Issue5587TxRetryExponentialBackoffTest` does exactly this for its early-gaps-vs-late-gaps ratio, and is the pattern to copy.

## What to do

Either:

1. **Restore the assertion** on stall-discounted durations, with enough margin that it is measuring the plan and not the scheduler. If the split form is only marginally faster on a small fixture, grow the fixture until the gap is structural rather than incidental - the whole point of #3216 is that the single-MATCH form does work proportional to something the split form does not.
2. **Or assert the plan instead of the clock.** If the two forms differ by producing a Cartesian product vs. two indexed lookups, that difference is visible in the execution plan, and a plan assertion is immune to the machine it runs on. This is the better fix if the plan exposes it.
3. **Or delete the test** and say so. A test that cannot fail is worse than no test, because it reports coverage that is not there.

Whichever is chosen, the tautological `isGreaterThanOrEqualTo(0L)` lines should not survive it.

### LINKED ISSUE #6280: OpenCypherCountOptimizationTest cannot detect the count optimization silently regressing
Found while sweeping test assertions for #6260.

`OpenCypherCountOptimizationTest` verifies that `MATCH (a:Account) RETURN COUNT(a)` returns the right number. It does not, and currently cannot, verify that the count optimization is what produced it. Six plan assertions across three tests are commented out behind the same TODO:

```java
// TODO: Verify optimization is being used by checking EXPLAIN output
// EXPLAIN integration needs to be added separately
// final ResultSet explainResult = database.query("opencypher", "EXPLAIN " + query);
// assertThat(plan).contains("TYPE COUNT OPTIMIZATION");
```

Two of the six are the *negative* cases (`doesNotContain("TYPE COUNT OPTIMIZATION")`), which matter just as much: they are what stops the optimization firing where it is not valid.

This is a weaker finding than #6279 - these tests do assert correct results, so they are not vacuous - but the gap is real and it is in the direction that hurts. If the push-down stopped firing tomorrow, every one of these tests would still pass, because a full type scan returns the same count as the optimized path. It just returns it slower, and slowly is not something this suite checks. Given the count push-down has already been worked on twice (#5686 uncorrelated COUNT push-down, #5715 push-down preconditions, where a LIGHTWEIGHT edge yields `countType` 0), a silent regression here is not hypothetical.

## What to do

The TODO says EXPLAIN integration "needs to be added separately" - that is the actual blocker, and it is worth checking whether it still is. `EXPLAIN` works on the SQL side, so the question is only whether the openCypher engine routes it and surfaces a plan string.

- **If EXPLAIN now works for openCypher**: uncomment the six assertions, confirm the plan marker string still matches what the optimizer emits, and delete the TODO.
- **If it does not**: assert on something else observable rather than leaving the tests half-blind. A counter on the optimizer, or the absence of a bucket scan in the profile, both discriminate the optimized path from the full scan without needing EXPLAIN. Note the bound must not be a wall-clock one - "the optimized path is faster" is precisely the assertion shape #6260 was filed about, and `StallAwareStopwatch.assertStayedUnder` is the floor to build on if there is genuinely no structural signal available.

Related: #5686, #5715.

### LINKED ISSUE #6297: Follow-ups from #6267: a teardown that deletes /databases0 instead of the test's folders, a warn-once latch that never rearms, and two notes already in code
Four follow-ups from #6267 / PR #6284, filed together because they share an origin. Only item 1 is a defect; item 2 is a small robustness gap, items 3 and 4 are already recorded in code comments and are here so they are tracked somewhere searchable.

## 1. `endTest()` resets the configuration before it deletes, so the cleanup targets `/databases0` instead of the test's folders

`BaseGraphServerTest.endTest()` (`server/src/test/java/com/arcadedb/server/BaseGraphServerTest.java:233-256`) does, in this order:

```java
} finally {
  GlobalConfiguration.resetAll();                              // :238
  stopServers();                                               // :241
  if (dropDatabasesAtTheEnd())
    deleteDatabaseFolders();                                   // :245
  ...
}
TestServerHelper.deleteDatabaseFolders(getServerCount());      // :255
```

`TestServerHelper.deleteDatabaseFolders()` resolves the directory **at call time**:

```java
FileUtils.deleteRecursively(new File(GlobalConfiguration.SERVER_DATABASE_DIRECTORY.getValueAsString() + File.separator));
for (int i = 0; i < totalServers; ++i)
  FileUtils.deleteRecursively(new File(GlobalConfiguration.SERVER_DATABASE_DIRECTORY.getValueAsString() + i + File.separator));
```

`SERVER_DATABASE_DIRECTORY` defaults to `${arcadedb.server.rootPath}/databases`, and `SERVER_ROOT_PATH` defaults to `null`. `getValueAsString()` resolves the placeholder through `SystemVariableResolver` with `""` as the fallback for an unresolvable variable, so once `resetAll()` has run the whole prefix collapses to the empty string.

Confirmed by running it (engine classes on the classpath, no server involved):

```
DURING TEST  : ./target/databases
AFTER RESET  : /databases
DELETE TARGET: /databases0
```

Two consequences:

- **The cleanup is a no-op for what it is meant to clean.** `./target/databasesN` - and the `.raft-storage` under it since #5272 - survives teardown. Nothing removes it until the *next* class's `beginTest`, and the last class to run in a fork leaves its databases on disk indefinitely. The contract "databases are dropped at the end of the test" is not kept, and `dropDatabasesAtTheEnd()` does not control what it appears to.
- **It issues recursive deletes against absolute root-level paths.** `/databases`, `/databases0` ... `/databasesN`. Harmless on a machine where those do not exist, which is why this has never been noticed - but it is a `deleteRecursively` at `/` driven by a path the test never meant to compute, and a data volume mounted at `/databases0` would be inside the blast radius.

Suspected fix: delete first, reset after - the reset is only there to leave the JVM clean for the next class, and it has no reason to run before the paths it invalidates are used. Worth pairing with a guard in `TestServerHelper.deleteDatabaseFolders` that refuses a path which did not resolve (empty prefix / absolute root), so the next reordering cannot silently point it somewhere else again.

Possibly related: the `DynamicMembershipTest` failure reported in #6267 as `endTest > checkDatabasesAreIdentical: DatabaseAreNotIdentical Types: DB1 5 <> DB2 15`. #6284 reproduced and fixed a *different*, same-method cause of an `endTest` comparison failure, but 15 types is not a shape any method of that class produces, so that particular instance was cross-class - and this is the only cross-class disk-state leak found while investigating. Not proven, stated as a lead rather than a diagnosis.

## 2. The warn-once latches never rearm, so a second misconfiguration is silent

`RaftHAServer` has six "log this once" latches: `httpFallbackWarned`, `httpsFallbackWarned`, `routingFallbackWarned` and `routingAmbiguityWarned` (per protocol), and `httpAmbiguityWarned` / `httpsAmbiguityWarned` added by #6284. Each is set for the life of the `RaftHAServer` and is never cleared.

The intent is right - the resync and verify paths ask on every attempt, and a misconfiguration that has not changed should not be re-reported on each one. The gap is what happens *after* the operator acts: they read the line, declare the two ports it named, and a **different** pair of peers is now colliding (or a peer added at runtime collides with an existing one). Nothing is logged, because the latch is spent.

#6284 mitigated the specific case by making the line name *every* collision the pass found rather than one of them, and by exposing `httpAddressAmbiguous` on `GET /api/v1/cluster`, which is recomputed per request and always current. What is left is the log going permanently quiet on a cluster whose topology changed.

Suggested: rearm the ambiguity latches when the committed Raft configuration changes - that is exactly the event that can invalidate the previous verdict, and it is already observed (`RaftClusterStatusExporter` re-emits the configuration table on it). The *fallback* latches (deriving at all) probably should not rearm: that condition does not change with membership.

## 3. The next cut of `RESYNC_RETRY_TIMEOUT_MS`

#6284 took it from 120 s to 30 s on the evidence of nine full `ha-integration-tests` runs in which no wait exceeded the 10 s report threshold, and lowered `SLOW_WAIT_REPORT_MS` to 5 s to keep resolution for the next decision. A handful of runs at the new threshold will say whether 30 s can come down further, or whether something does sit between 5 s and 30 s and the budget is now correctly sized. Recorded in `BaseRaftHATest`'s comment; this is just the tracking entry so it does not depend on someone rereading that file.

## 4. Isolating the heavy HA ITs into their own fork

The other half of the original 120 s diagnosis, still open. The stalls that motivated that budget appeared only in full-suite adjacency, immediately after `RaftMigrationCompactionRaceIT` in the same reused JVM fork (failsafe `reuseForks=true` / `forkCount=1` for the whole module). #6202 removed the suspected cause (the snapshot download running on the JDK common ForkJoinPool) and #6284 cut the budget on measured evidence, but nothing has been done about the adjacency itself. Also recorded only in a code comment today.

lvca:
Closes #6297, closes #6280, closes #6279, closes #6270.

Four follow-ups from #6267 / #6268. They share an origin - all four are about a test that reports coverage it does not have - so they are fixed together.

## #6297 (1) The teardown resolved its paths from a configuration it had just reset

`BaseGraphServerTest.endTest()` called `GlobalConfiguration.resetAll()` **before** the cleanup. Every path `TestServerHelper.deleteDatabaseFolders()` deletes is resolved at call time from `SERVER_DATABASE_DIRECTORY`, whose default is `${arcadedb.server.rootPath}/databases` over a root path that defaults to `null`, and `getValueAsString()` substitutes the empty string for an unresolvable variable rather than failing. The prefix collapsed, so the cleanup asked for a recursive delete of `/databases`, `/databases0` ... while the test's own `./target/databases0` - and the `.raft-storage` under it - survived teardown untouched.

- The reset moved to the end, inside a `finally` that now also covers the deletes, so a failing `checkDatabasesAreIdentical()` no longer skips the cleanup entirely (that is the cross-class disk-state leak the issue suspected behind the `DynamicMembershipTest` "DB1 5 <> DB2 15").
- `StaticBaseServerTest.beginTest()` had the mirror-image bug - it cleaned before `setTestConfiguration()`, so on the first class in a fork it cleaned `/databases0` too. Configuration first now.
- `TestServerHelper.isResolvedTestPath` refuses a prefix that did not resolve: blank, or absolute and naming a single root-level folder. A relative path (`./target/databasesN`) or a deeper absolute one (a `@TempDir` under the system temp dir) is always allowed. A refusal logs SEVERE and deletes nothing, so a future reordering degrades to a logged no-op instead of a `deleteRecursively` at `/`.

`Issue6297TeardownDeletesTestFoldersIT` asserts in `@AfterAll` - the only hook JUnit runs after `@AfterEach` - that `./target/databases0` is gone. Verified it goes red on the old ordering:

```
[endTest() left the test's own database behind (issue #6297)]
Expecting file: .../server/./target/databases0/graph  not to exist
```

`Issue6297TeardownPathResolutionTest` pins the trap itself (a reset configuration resolves to `/databases`), the guard's accept/refuse set, and both orderings end to end against a `@TempDir`.

## #6297 (2) The warn-once latches never rearmed

Rather than the issue's suggestion (rearm on a committed Raft configuration change), the latch is replaced by **the rendered collision list itself**. That is strictly better in both directions:

- a membership change that leaves the verdict identical says nothing new, and now stays quiet - rearming on the event would re-log the same unchanged misconfiguration on every peer add;
- any change to the verdict is reported, *whatever* caused it - a different pair colliding, a peer added at runtime onto an address already taken, an HTTP port becoming available late - not only the causes someone remembered to hook.

A clean pass clears the memory instead of reporting, so a misconfiguration that is fixed and later reintroduced is announced again. The line already named every collision it found, which is what makes it usable as the key; `warnAmbiguousRouting` now names them too, so the routing warning has both a usable key and a usable message. The *fallback* latches (deriving at all) are deliberately left as booleans - that condition does not change with membership.

`Issue6297AmbiguityRearmTest` covers the rearm contract and the verdict rendering (stable order whatever order Ratis hands the group over in, declared-beats-derived, clean view renders empty).

## #6297 (3) and (4)

Left as tracking entries, deliberately:

- **(3) the next cut of `RESYNC_RETRY_TIMEOUT_MS`** is a decision that needs a handful of `ha-integration-tests` runs at the new 5 s report threshold. There is no code change to make until those runs exist; making one now would be guessing at the number the issue asks to measure.
- **(4) isolating the heavy HA ITs into their own fork** is a build/CI-lane change to a module whose IT suite is already a chronic flake (#5668/#5702). It does not belong in a PR whose other three items are test-assertion fixes: it changes what runs where, so its own regression signal is a series of full-suite runs, not a test.

## #6280 EXPLAIN is no longer the blocker

The TODO said "EXPLAIN integration needs to be added separately". It does not: `OpenCypherQueryEngine` routes an `EXPLAIN ` prefix and `CypherExecutionPlan.explain()` describes the count push-down first, precisely because the push-down replaces the whole chain.

```
EXPLAIN MATCH (a:Account) RETURN COUNT(a) as count
  Using Count Push-Down
  + TYPE COUNT OPTIMIZATION (Account)

EXPLAIN MATCH (a:Account) WHERE a.balance > 5000 RETURN COUNT(a) as count
  Using Cost-Based Query Optimizer
  + NodeByLabelScan(a:Account) [filter: a.balance > 5000]
```

All six commented-out assertions are restored against it (the two negative ones included), and the alias, empty-type and polymorphic cases get one as well - they had none. `EXPLAIN` rather than `PROFILE` on purpose: the negative cases describe the full scan the push-down was refused for, and running one to find out that it ran is the cost the assertion exists to detect.

## #6279 The comparison the test claimed to make is no longer true

`issue3216_performanceComparisonMatchStrategies` asserted `duration >= 0` twice, which cannot fail. Its premise had also expired: `CypherExecutionPlan.extractIdFilter` lifts an `ID(x) = <literal|parameter>` equality out of the WHERE clause into the `MatchNodeStep` at plan time - the fix #3216 was closed on - and it applies to the single-MATCH form exactly as it does to the split one. Restoring `assertThat(duration2).isLessThan(duration1)` would have asserted something false.

Rather than deleting the test, the property it was reaching for is made visible. `MatchNodeStep.prettyPrint` now names the RID it was given, alongside `[index: ]` and `[filter: ]`:

```
+ MATCH NODE (a) [id: #1:0]
+ MATCH NODE (b) [id: #4:0]
+ FILTER WHERE (ID(a) = $sourceId AND ID(b) = $targetId)
```

That is a user-facing improvement in its own right - the resolution of #3216 tells users to write the query so the push-down fires, and until now the plan was the one place that would not say whether it had. The test asserts both forms carry the marker, plus a control query (property equality instead of `ID()`) whose plan does not, so the assertion discriminates rather than being printed unconditionally. Off the `benchmark` lane, since it no longer times anything.

## #6270 The wall-clock and @Timeout pass, outside the engine suite

Four elapsed-time bounds in `server`/`network` now measure through `StallAwareStopwatch`, all of them category 1 (a tripwire between a bounded operation and an unbounded one), all `assertGaveUpWithin`: `Issue5470BatchStreamStallIT`, `Issue5470BatchErrorDeliveryIT`, `Issue5418ShutdownHookDeadlockTest`, `RemoteHttpComponentTest`.

Two premises in the issue were wrong and are worth recording:

- **`HTTPAuthSessionIT` needs nothing.** Its two sites are `elapsedFromCreation() >= 100` (a lower bound, which a stall only makes more true) and a field-presence check. Same for `PostBatchHandlerIT`, whose `elapsedMs >= 0` is a response-field sanity check, not a latency bound.
- **`server` and `network` did not already consume the engine test-jar.** Neither declares it and neither uses `TestHelper`. Added at test scope in both poms - no new external dependency; the artifact is already produced by `engine/pom.xml`.

The `@Timeout` pass covers every engine annotation at 45 s or under. The conclusion differs from the issue's guess in one place, and the difference matters: **`RangeFunctionTest`'s annotations are not the assertion.** What each of those methods regresses to is an unbounded loop or a materialised range, and what each one *claims* is carried by a structural assertion - `containsExactly` on the returned elements, `isInstanceOf(LongRangeList.class)` on the lazy one, `assertThatThrownBy` on the rejected one. None of those can pass while the range is being materialised. So widening the annotation does not weaken the GHSA-xmjm regression test; leaving it at 5 s, three orders of magnitude below the honest budget and well inside the range of the 24 s stalls #6260 was filed on, is what made it a coin flip.

- Hang detectors, sized clear of the honest budget plus a ~30 s stall: `RangeFunctionTest` (5 s/10 s -> 60 s), `IndexDescOrderDuplicateKeysTest`, `LongObjectHashMapTest` (`SEPARATE_THREAD` kept - the regression is an infinite probe loop), `HashIndexOverflowCycleTest`, the two rejection cases in `CypherRangeHeapExhaustionTest`, and `LockManagerFairnessTest`, whose stress methods already give themselves up to 45 s (`done.await(45, SECONDS)`) or 25 s (250 rounds x 100 ms) of honest work and so were sized *below* their own budget once a stall is added.
- The two genuine laziness claims - `LongRangeListTest.hugeRangeIsFreeOfHeap` and `CypherRangeHeapExhaustionTest.hugeRangeIsLazyWhenTheLimitIsDisabled` - are now asserted with `assertStayedUnder` on the discounted clock, which is *tighter* than the annotation it replaces, with the annotation left behind as the hang detector.
- The `server` module's four `@Timeout(60)` in `RangeHeapExhaustionHttpStatusIT` are already generous against a sub-second honest budget: nothing to do.

Finally, the sampler thread is renamed `arcadedb-test-stall-monitor` -> `test-jvm-stall-monitor`, out of the prefix `DatabaseLifecycleBackgroundThreadsTest` scans for leaked engine threads. It was safe only because that loop skips daemons first; a shutdown path making the sampler non-daemon would have turned a leak detector red pointing at #5418. `StallAwareStopwatchTest` now pins the name against all three scanned prefixes.

## Verification

- `Issue6297TeardownDeletesTestFoldersIT` verified red on the pre-fix ordering, green after.
- Green: `engine` `com.arcadedb.query.opencypher.**`, the touched `engine` utility/index/function classes, `ha-raft` unit lane, `server` unit lane, `network` `RemoteHttpComponentTest`, `server` `Issue5470BatchErrorDeliveryIT`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/MatchNodeStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/MatchNodeStep.java
@@ -909,6 +909,16 @@ public class MatchNodeStep extends AbstractExecutionStep {
       builder.append(":").append(String.join("|", pattern.getLabels()));
     }
     builder.append(")");
+    // The ID push-down is the single most consequential thing this step can do - it turns a full type scan (and,
+    // when another MATCH NODE follows, the Cartesian product over it) into one lookupByRID - and until issue #6279
+    // it was the one optimisation the plan did not name, so nothing but a stopwatch could tell whether it fired.
+    // That is exactly what the resolution of issue #3216 asks a user to verify. Printed from the plan-time field
+    // rather than recorded during execution, so EXPLAIN answers it without running the query; the dynamic form
+    // (issue #3864) resolves per row and can only be named, not valued.
+    if (idFilter != null && !idFilter.isEmpty())
+      builder.append(" [id: ").append(idFilter).append("]");
+    else if (dynamicIdExpression != null)
+      builder.append(" [id: per-row]");
     if (usedIndexName != null)
       builder.append(" [index: ").append(usedIndexName).append("]");
     if (usedPartitionBucket != null)
--- a/engine/src/test/java/com/arcadedb/function/coll/RangeFunctionTest.java
+++ b/engine/src/test/java/com/arcadedb/function/coll/RangeFunctionTest.java
@@ -30,6 +30,17 @@ import java.util.concurrent.TimeUnit;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+/**
+ * Advisory GHSA-xmjm-8q85-g778 and the overflow guards around it.
+ * <p>
+ * Every {@code @Timeout} here is a HANG DETECTOR, not a latency bound (issue #6270): what each of these methods
+ * regresses to is an unbounded loop or a materialised billion-element list, and what the method actually CLAIMS is
+ * carried by a structural assertion - {@code containsExactly} on the returned elements, {@code isInstanceOf}
+ * {@code LongRangeList} on the lazy one, {@code assertThatThrownBy} on the rejected one. None of those can pass
+ * while the range is being materialised, so the annotation only has to be wide enough that a stop-the-world pause
+ * cannot masquerade as the regression. The honest budget of each method is microseconds; at 5 s and 10 s these
+ * were inside the range of the 24 s stalls issue #6260 was filed on.
+ */
 @SuppressWarnings("unchecked")
 class RangeFunctionTest {
 
@@ -37,7 +48,7 @@ class RangeFunctionTest {
 
   /** Reported case: step skips past Long.MAX_VALUE - must return only the start element. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void positiveStepOverflowReturnsOnlyFittingElements() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -47,7 +58,7 @@ class RangeFunctionTest {
 
   /** Step of 1 right up to Long.MAX_VALUE must return all 4 values. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void positiveStepUpToMaxValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -61,7 +72,7 @@ class RangeFunctionTest {
 
   /** Single-element range exactly at Long.MAX_VALUE. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void singleElementAtMaxValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -71,7 +82,7 @@ class RangeFunctionTest {
 
   /** Negative step that skips past Long.MIN_VALUE - must return only the start element. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void negativeStepUnderflowReturnsOnlyFittingElements() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -81,7 +92,7 @@ class RangeFunctionTest {
 
   /** Step of -1 right down to Long.MIN_VALUE must return all 4 values. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void negativeStepDownToMinValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -95,7 +106,7 @@ class RangeFunctionTest {
 
   /** Single-element range exactly at Long.MIN_VALUE. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void singleElementAtMinValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -105,7 +116,7 @@ class RangeFunctionTest {
 
   /** Extreme: step = Long.MIN_VALUE; guard evaluates to 0 (MIN - MIN = 0 in two's-complement). */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void negativeStepLongMinValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -115,7 +126,7 @@ class RangeFunctionTest {
 
   /** Symmetric to negativeStepLongMinValue: step = Long.MAX_VALUE on the positive branch. */
   @Test
-  @Timeout(value = 5, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void positiveStepLongMaxValue() {
     @SuppressWarnings("unchecked")
     final List<Long> result = (List<Long>) fn.execute(
@@ -134,7 +145,7 @@ class RangeFunctionTest {
 
   /** Advisory GHSA-xmjm-8q85-g778: the range is lazy, so a huge one costs a constant amount of heap. */
   @Test
-  @Timeout(value = 10, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void largeRangeIsLazy() {
     final long previous = GlobalConfiguration.QUERY_MAX_RANGE_SIZE.getValueAsLong();
     try {
@@ -150,7 +161,7 @@ class RangeFunctionTest {
 
   /** Advisory GHSA-xmjm-8q85-g778: the reported PoC must be rejected as a client error, not attempted. */
   @Test
-  @Timeout(value = 10, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void oversizedRangeIsRejected() {
     assertThatThrownBy(() -> fn.execute(new Object[]{ 0L, 9_999_999_999L }, null))
         .isInstanceOf(CommandSemanticException.class)
@@ -160,7 +171,7 @@ class RangeFunctionTest {
 
   /** Even with the configured limit disabled, a range cannot exceed the maximum size of a Java list. */
   @Test
-  @Timeout(value = 10, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void rangeBiggerThanAListIsRejectedEvenWithoutLimit() {
     final long previous = GlobalConfiguration.QUERY_MAX_RANGE_SIZE.getValueAsLong();
     try {
--- a/engine/src/test/java/com/arcadedb/index/hash/HashIndexOverflowCycleTest.java
+++ b/engine/src/test/java/com/arcadedb/index/hash/HashIndexOverflowCycleTest.java
@@ -88,7 +88,7 @@ class HashIndexOverflowCycleTest extends TestHelper {
   }
 
   @Test
-  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void cyclicOverflowChainFailsLoudlyOnLookupInsteadOfSpinning() {
     injectSelfReferentialOverflowChain();
     reopenDatabase();
@@ -106,7 +106,7 @@ class HashIndexOverflowCycleTest extends TestHelper {
   }
 
   @Test
-  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void cyclicOverflowChainFailsLoudlyOnUniqueCheckDuringInsert() {
     // Mirrors the customer's stack: the spin was reached via checkUniqueIndexKeys while committing a new vertex.
     injectSelfReferentialOverflowChain();
--- a/engine/src/test/java/com/arcadedb/query/opencypher/CypherRangeHeapExhaustionTest.java
+++ b/engine/src/test/java/com/arcadedb/query/opencypher/CypherRangeHeapExhaustionTest.java
@@ -25,6 +25,7 @@ import com.arcadedb.exception.CommandParsingException;
 import com.arcadedb.query.sql.executor.Result;
 import com.arcadedb.query.sql.executor.ResultSet;
 import com.arcadedb.utility.LongRangeList;
+import com.arcadedb.utility.StallAwareStopwatch;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -68,7 +69,7 @@ class CypherRangeHeapExhaustionTest {
 
   /** The reported PoC: it must fail fast instead of filling the heap. */
   @Test
-  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void reportedPocIsRejectedWithoutExhaustingHeap() {
     final Throwable thrown = catchThrowable(() -> consume("RETURN range(0, 9999999999) AS v"));
     assertThat(thrown).isNotNull();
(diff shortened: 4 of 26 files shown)


lvca:
Merged into `main` as 260ce46b6d1bf42e58bf5f4b4a0d2c4549a1e433 (squash), since the GitHub merge button was unavailable.

`main` had moved 8 commits ahead of this branch, so `origin/main` was merged in first (4dc043f2a, pushed here) - the only overlapping file was `RaftHAServer.java` and git resolved it cleanly. The merged tree was rebuilt and the affected suites re-run green before the squash: the ha-raft ambiguity tests, the #6297 teardown regression tests, `Issue5418ShutdownHookDeadlockTest`, and the engine range/openCypher/stopwatch tests.

The squash commit's tree is byte-identical to this branch's merged tip, and its diff against `main` is exactly this PR's 26 files. Closing manually: a squash does not make this branch's tip reachable from `main`, so GitHub cannot mark it merged automatically. #6297, #6280, #6279 and #6270 were closed by the commit's trailers.

PR #6308, opened 2026-08-17
fix(#6304): a count push-down that keeps the labels, and one meaning for TIMEOUT

### LINKED ISSUE #6304: Follow-ups from #6266: an ignored label filter in a count push-down, a TIMEOUT ... RETURN that is still coarse, and three consistency gaps
Five independent findings turned up while fixing #6266 (PR #6291, merged as `11763ff1a`). None of them belonged in that PR, and they are unrelated to each other beyond having been found in the same neighbourhood. Filed together for triage; each is self-contained and can be split out or closed on its own.

All references are against `11763ff1a`.

---

## 1. `PairHashJoinOp.buildWithViews` ignores the label filter it is handed, so one CSR count push-down path can over-count

`engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/PairHashJoinOp.java:331`

The method takes `arm2Buckets` - the per-hop bucket sets built from `arm2IntermediateLabels` - and never reads it. Every sibling branch does:

| branch | line | applies `arm2Buckets`? |
|---|---|---|
| `buildAndProbeInline` | 142 | yes (`arm2Filter0`/`arm2Filter1`) |
| **`buildWithViews`** | **148** | **no** |
| `buildWithArm1View` | 150 | yes (via `walkArmWithViews`) |
| generic `walkArm` fallback | 153 | yes |
| OLTP `executeOLTP` | 194 | yes (via `walkArmOLTP`) |

**Reachability.** `buildWithViews` runs when both arms are single-hop *and* both have a `NeighborView`, *and* the inline fast path did not return - which happens when the probe edge type has no `NeighborView` of its own. `arm2Buckets` is non-null whenever any entry of `arm2IntermediateLabels` is non-null (`buildValidBucketSets`, line 439), and for a single-hop arm that array has length 1 holding the label on the arm's endpoint. So a labelled single-hop arm2 reaches this path with a filter that is then dropped.

**Why it matters.** The dropped filter is not a performance detail - it decides which endpoints are counted. The same query would then return a different number depending on whether a Graph Analytical View happens to cover the probe edge type, because the OLTP fallback applies the labels and this path does not.

Codacy flagged the neighbouring unused `provider` parameter during #6291 and I removed that one, but deliberately left `arm2Buckets` in place so the symptom would still be visible here rather than being tidied away.

**Wanted first:** a test that reaches this branch with a labelled arm2 and asserts the count against the OLTP path. If it turns out the branch is unreachable with a non-null `arm2Buckets`, the fix is a precondition making that explicit, not a filter.

---

## 2. A `TIMEOUT n RETURN` statement is still bounded only between batches

`engine/src/main/java/com/arcadedb/query/sql/executor/StatementTimeouts.java:79`

`StatementTimeouts.publish` deliberately declines to pin a deadline for the `RETURN` failure strategy:

```java
if (context == null || Timeout.RETURN.equals(clause.getFailureStrategy()))
  return;
```

The reason is sound - a `WorkGuard` can only throw, and `RETURN` asks for the rows produced so far, so pinning would silently convert a documented "return what you have" into a failure. The consequence is that this one clause shape keeps exactly the granularity hole #6266 was about: a filter that rejects every record still scans to the end inside a single `hasNext()`, and only the between-batches check applies.

Note this became reachable *because* #6291 fixed `SQLASTBuilder.visitTimeout`, which had never read the `EXCEPTION | RETURN` token the grammar accepts - before that, `failureStrategy` was always null and every clause behaved as `EXCEPTION`.

**Suggested shape.** The guard needs a second signal meaning *stop and yield what you have*: a distinct exception thrown by `WorkGuard` when the pinned bound is a RETURN-strategy one, which the owning timeout step catches and converts into the partial/empty result it already produces in `fail()`. That keeps the abort cooperative and leaves the in-loop check as the only place that knows how to stop early.

---

## 3. `SELECT ... TIMEOUT n` and `UPDATE ... TIMEOUT n` mean two different things

- `AccumulatingTimeoutStep` (SELECT) charges only time spent *inside the pipeline*: `totalTime.addAndGet(System.nanoTime() - begin)` around each `hasNext()`/`next()`. A consumer that pauses between fetches is not billed for the pause.
- `TimeoutStep` (UPDATE) sets `expiryTime` once on its first pull and compares wall clock against it.

Same syntax, two semantics, and nothing documents either. A client streaming a large `SELECT` slowly gets a very different bound from the same number written on an `UPDATE`.

PR #6291 preserved both faithfully rather than unify them mid-change (`AccumulatingTimeoutStep` republishes its instant every batch as `now + (clause - accumulated)` precisely so the in-loop guards inherit its accounting instead of quietly becoming wall-clock). But a user cannot currently predict which of the two they get.

**Options:** document the difference where the clause is documented, or unify on one meaning. If unifying, wall-clock is the one users expect from the word "timeout", and the accumulating variant should then be a distinct opt-in rather than a per-statement-kind accident.

---

## 4. SQL `MATCH` and `TRAVERSE` accept no `TIMEOUT` clause at all

`engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLParser.g4:223` (`traverseStatement`), `:237` (`matchStatement`)

Neither production has a `timeout?`, while `selectStatement` (`:202`) and `updateStatement` do. So `SELECT ... TIMEOUT 100` parses and `MATCH {...} RETURN ... TIMEOUT 100` is a syntax error, even though `Statement.timeout` exists on the shared base class and `MatchStatement` sets it internally for profiled timeouts.

Both statements now honour the global `arcadedb.command.timeout` (that was #6266), so this is a syntax-consistency gap rather than a safety one - a user who wants to bound one expensive `MATCH` has to change a database-wide setting instead.

---

## 5. The regex deadline still multiplies by the number of parallel scan workers

`engine/src/main/java/com/arcadedb/query/sql/executor/CommandContext.java:73`

`getOrComputeRegexDeadline`'s own javadoc records the gap:

> A type scanned in parallel across N buckets is therefore bounded by `N * timeoutMillis` overall, not one shared budget.

The cause is that the deadline lives in `cachedValues`, and `BasicCommandContext.copy()` does not copy that map - so each `syncPullParallel` worker computes its own. The javadoc argues the exposure is small (bucket count is a schema property, not attacker-controlled), which is fair.

What has changed is that a fix pattern now exists: #6291 made `copy()` *resolve* the command deadline before copying it, precisely so that copying an unresolved value would not hand every worker a fresh budget. The same treatment - resolve `arcadedb.command.regexTimeout` into the parent before copying, or carry the resolved instant on the context rather than in the opaque cache - would close this for free and remove a documented caveat.

---

### Suggested triage

Item 1 is the only possible wrong answer and is the one worth looking at first. Item 2 is a real remaining hole in what #6266 set out to close. Items 3-5 are consistency and documentation debt.

lvca:
Closes #6304.

Five follow-ups from #6266, unrelated to each other beyond having been found in the same neighbourhood. Each is self-contained.

## 1. The pair-join count push-down dropped labels the pattern wrote down

The issue named `PairHashJoinOp.buildWithViews`, which is handed `arm2Buckets` and never reads it. That branch is real, but it is only reached when the probe edge type has no `NeighborView` while both arms do - which no `GraphAnalyticalView` produces on its own.

**The live drop was a different one.** `arm1`'s endpoint filter was missing from *all three* `NeighborView` build paths, including the inline build-and-probe fast path that serves the common shape. The reason is mechanical: the `bucketIds` table those filters need was built inside the branch that owned `arm2`, so `arm1`'s filter had nothing to test against and was simply never written. The OLTP fallback applies it, so the same query returned a different number depending on whether a Graph Analytical View happened to cover the edge types:

```cypher
MATCH (p1)-[:KNOWS]->(p2), (p1:Person)<-[:AUTHORED]-(c:Comment)-[:MENTIONS]->(p2)
RETURN count(*)
```
OLTP said 1, CSR said 2 - the second comment's author is a Bot.

**A third drop, on both paths.** The detector read node labels off the build pattern only. The two comma-separated patterns share their endpoint variables, so `(p1:Person)-[:KNOWS]->(p2)` constrains exactly the variable the other pattern's arm ends on. That label was dropped whether or not a view existed, so a CSR-vs-OLTP comparison would not have caught it; the tests cross-check every count against the row count the ordinary materialization pipeline produces instead.

Where the two sides label one variable differently the filter is an intersection, which the operator carries no room for, so the detector declines and the pipeline answers. It declines the same way for `(a:A:B)` and `(a:A|B)`, which it used to read as `(a:A)` - a filter that is neither.

The bucket-id table is now built once for both arms, and every path applies both filters.

## 2. `TIMEOUT n RETURN` stops when it says it will, and returns what it has

It delivered neither half. Its deadline was deliberately left out of the in-loop guards - a guard can only stop by throwing, and `RETURN` promises no exception - so a filter that rejects every record still scanned to the end inside one `hasNext()`, which is precisely the granularity hole #6266 set out to close. And when rows *did* pass the filter, the step marked itself timed out and then went on returning every remaining row, so nothing was ever truncated.

A guard reaching a deadline pinned by a `RETURN` clause now raises a distinct `PartialResultTimeoutException`, which the step owning the clause converts into the end of its result set. Every other bound keeps raising a plain `TimeoutException`; the distinction is what stops a genuine `arcadedb.command.timeout` abort from being swallowed into a silently truncated answer. An outer `RETURN` clause propagates the same meaning into nested plans.

The step's per-batch `hasNext()` is memoized, because `LocalResultSet.next()` re-asks after its caller already asked and a deadline crossing in between would turn a promised row into an `IllegalStateException`.

**Behaviour change:** `SELECT ... TIMEOUT n RETURN` used to return the complete result set and now returns a prefix of it - which is what the clause has always been documented to do. `EXCEPTION`, the default, is unaffected.

## 3. `SELECT ... TIMEOUT n` and `UPDATE ... TIMEOUT n` now mean the same thing

`UPDATE` measured wall clock from the first pull; `SELECT` charged only time spent inside the pipeline, so a client streaming a large result slowly was not billed for its own pauses while the identical number on an `UPDATE` was.

Rather than document the divergence, this removes it. Both are wall clock, which is what the word means everywhere else in the engine: `arcadedb.command.timeout`, the ceiling over every statement, has been wall clock since #6266, so the accumulating variant was the only bound in the engine that was not - and once #6291 pinned the clause's instant on the context, the in-loop guards read it as wall clock regardless. `AccumulatingTimeoutStep` is deleted and `TimeoutStep` serves both.

**Behaviour change:** a consumer that pauses between fetches of a `SELECT ... TIMEOUT n` is now charged for the pause.

## 4. SQL `MATCH` and `TRAVERSE` accept `TIMEOUT`

`SELECT ... TIMEOUT 100` parsed and `MATCH {...} RETURN ... TIMEOUT 100` was a syntax error, so bounding one expensive `MATCH` meant changing a database-wide setting. Both productions take `timeout?` now, both planners chain the step last, and both statements carry the clause through `copy()`/`equals()`/`toString()`. The profiled read timeout `MatchStatement` was already setting on itself - and which nothing then read - takes effect with it.

## 5. One regex budget per command

`getOrComputeRegexDeadline` kept its deadline in the context's opaque value cache under a key each call site chose for itself. That gave a query as many budgets as it used regex features, and - because `copy()` does not carry that cache - one more per parallel bucket-scan worker, so a type scanned across N buckets was bounded by `N * regexTimeout`. The deadline is now a field of the context, resolved once and pinned into the copy exactly as the command deadline is. It also removes the key-collision hazard that shipped a real bug (a fixed `MATCHES_DEADLINE` key colliding with the pattern cache's `MATCHES_` + regex-text keys).

## Verification

- `CypherPairJoinLabelFilterIssue6304Test` - 10 tests. Every count is cross-checked against the OLTP path *and* the materialization pipeline. Reverting `PairHashJoinOp` alone fails 7 of them, including the `buildWithViews` branch, which a delegating provider that hides one edge type's view reaches with a probe that still yields edges.
- `Issue6304TimeoutFollowUpsTest` - 15 tests. The `RETURN` assertions count rows the scan actually reached via a registered SQL function rather than measuring elapsed time, so a JVM stall can only push the number in the passing direction. Reverting the `publish` change alone fails them.
- `Issue6266CommandTimeoutCoverageTest` - 17 tests, unchanged assertions, still green.
- Full `engine` module: 12256 tests, 0 failures. Whole reactor packages.

## Found on the way, not fixed here

The materialization pipeline answers **0** for a two-pattern openCypher `MATCH` in which *both* shared variables are labelled and at least one of the labels is written on the second pattern - e.g. `MATCH (p1)-[:KNOWS]->(p2), (p1:Person)<-[:AUTHORED]-(c:Comment)-[:MENTIONS]->(p2:Person)`, which matches one row. Writing both labels on the first pattern answers correctly. That is unrelated to the push-down (it reproduces with `RETURN c`, no aggregate) and is worth its own issue; the one test that needs that shape uses the OLTP count as its reference and says so.

Also worth a separate look: the `.getLabels().get(0)` reading this PR guards against in the pair-join detector appears at six other sites in `CypherExecutionPlan`, each feeding a different count push-down.

Docs: `docs/release-26.9.1.md` covers all five, including the two behaviour changes. The SQL reference in `arcadedb-docs` needs the new `TIMEOUT` clause on `MATCH`/`TRAVERSE` and the wall-clock wording.


lvca:
Thanks — round 1 addressed. All three observations were acted on, and the first one turned out to be pointing at a live bug rather than a theoretical one.

**1. RetryStep and `TIMEOUT n RETURN` — the hypothesis was one step away from a real regression.**

Chasing "can anything actually reach that catch?" found the answer is yes, via a defect that predates this PR: **every line of a SQL script is planned against the same `CommandContext`** (`SQLScriptQueryEngine.executeInternal` passes `scriptContext` to every `stm.createExecutionPlan(...)`). So the instant a `TIMEOUT` clause pins lands exactly where the following lines read it, and those have no `TimeoutStep` of their own to catch anything:

```sql
SELECT FROM Node WHERE v = 1 TIMEOUT 50;   -- completes well inside its own bound
SELECT FROM SomethingSlower;               -- aborts: "the command exceeded the TIMEOUT clause of 50ms"
```

That is a regression #6291 shipped (pinning the clause on the context is what made the in-loop guards see it), and pinning `RETURN` clauses would have widened it to that strategy too. Fixed here: `ScriptExecutionPlan.drainLine` scopes a clause to the line that wrote it. What the line restores is the command's own `arcadedb.command.timeout` instant, so that bound stays in force across the whole script — the script *is* one command. Nested blocks (`IF`, `WHILE`, `FOREACH`, the `RETRY` body) were already safe: they plan against child contexts, so a pin inside them writes to the child.

Three regression tests, each verified to fail with the fix reverted: the leak itself, the global ceiling surviving a line that narrowed it, and a `RETRY` block not retrying an expired clause.

**With that closed, `RetryStep` is genuinely unreachable for this exception** — the step owning a clause always sits inside the same plan, below any `RETRY` block. I verified it directly: with the script fix in and the `RetryStep` guard reverted, every test still passes. So I did **not** add a catch clause there; an unreachable branch with no test behind it is the thing this issue's own triage note argues against ("if the branch is unreachable, the fix is a precondition making that explicit, not a filter"). The invariant is stated in `drainLine`'s javadoc instead.

One related thing I left alone and am flagging rather than changing: `RetryStep` retrying a plain `TimeoutException` from `arcadedb.command.timeout` is futile in the same way — the deadline has already passed, so each attempt aborts on its first check and the block just burns its budget plus backoff. That behaviour is deliberate and documented in the existing comment there, and changing it is a separate call from this PR.

**2. `TIMEOUT 0` — valid, unified.** `StatementTimeouts.stepFor` is now the single place that decides whether a clause bounds anything, used by all four planners so they cannot drift again. `0` disables, which is what `0` means for `arcadedb.command.timeout` and `arcadedb.command.regexTimeout` and is the direction that cannot turn a working statement into a failing one. It changes `SELECT ... TIMEOUT 0` from "fails on its first pull" to "unbounded"; no test relied on the old reading. Covered by `aZeroClauseDisablesTheBoundOnEveryStatementKind`.

**3. Cosmetic reflow** — fixed.

Re: the build note — the engine suite was re-run locally after these changes: **12260 tests, 0 failures** (`mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow,performance`), and the whole reactor packages. CI has the same diff now.

lvca:
Round 2 addressed. Both minor points turned out to be worth more than a nit.

**`TimeoutStep.next()`'s catch is not dead code.** `MatchStep.next()` pulls the following candidate before returning the current one (`MatchStep.java:82`), and that pull goes through `guard.check()` — so a SQL `MATCH ... TIMEOUT n RETURN`, which is syntax this PR adds, reaches the deadline from inside `next()` rather than `hasNext()`. That makes the contract objection live rather than hypothetical, and there is no good answer available inside `next()`: `NoSuchElementException` after `hasNext()` answered true breaks the iterator contract on a caller that is mid-iteration, raising is what the clause promised not to do, and the row cannot be handed back because `MatchStep` discarded it when its own prefetch threw.

So the case is removed rather than handled: `hasNext()` now performs the deadline test **and** the fetch, holding the row for `next()`. Every way a batch can reach the deadline is now one a `RETURN` clause answers by ending the result set. `aMatchThatYieldsTruncatesInsteadOfRaising` covers it, asserting both strategies on the same pattern so the clause is demonstrably reached.

**The `EXPLAIN` asymmetry is a message defect too**, which surfaced when one of my script tests started failing with `Timeout expired: TIMEOUT clause of 50ms` on a statement carrying no clause. `SelectExecutionPlanner` synthesized a `Timeout` out of `arcadedb.command.timeout` when the statement had none — that was the only way the setting reached SELECT before #6266, and it is redundant now that the setting travels on the command deadline every guard reads. Removed, so all four statement kinds behave identically: no phantom step in `EXPLAIN`, and an abort on the setting names the setting. `theGlobalSettingIsNotRenderedAsAClauseNobodyWrote` asserts both halves.

Also fixed two timing-fragile fixtures of my own: a 50 ms clause on a 60k-row scan is marginal, and those tests only need the first script line to *leave a pin behind*, not to reach it — they use `LIMIT 1` now.

Re: the sandbox limitation — the engine suite was re-run locally after these changes: **12262 tests, 0 failures**.

Not taken, and why:
- **`RetryStep` retrying a plain `TimeoutException`** from `arcadedb.command.timeout` is futile in the same way an expired clause is, but that behaviour is deliberate and documented in the comment there, and changing it is a separate call from this PR.
- **`bothArmEndpointLabelsAreAppliedTogether`'s hand-derived expectation** stays as-is: the pipeline reference is unusable for that shape because of the unrelated defect described at the end of the PR body, which deserves its own issue.

lvca:
All checks green, merge state clean.

The two lanes that failed on the first attempt both passed on a re-run of the **identical commit**, which is what identifies them as the known flakes rather than regressions:

| lane | first attempt | re-run |
|---|---|---|
| `vector-unit-tests` | fail, 21m0s | **pass**, 11m10s |
| `ha-integration-tests` | fail, 1h0m51s | **pass**, 1h1m29s |

The vector lane is the one `mvn-test.yml` documents in a comment as intermittent (`LSMVectorIndex.REBUILD_SEMAPHORE` holds one JVM-wide permit, so those classes convoy and blow their await timeouts); it also passes locally with the identical selector (`-pl engine -Dgroups=vector`, 47 tests, 0 failures). Neither lane touches the code this PR changes. Worth noting that neither failure was diagnosable from the API: the `dorny/test-reporter` check run was never published for either job and the single-job log endpoint returns only the runner setup and cleanup, so the failing test names were not recoverable — a re-run was the only way to tell.

Local verification across the three rounds: engine module **12262 tests, 0 failures**, whole reactor packages.

Issue #6283, opened 2026-08-17
Follow-ups from #6198: file-id/component-name mismatch, TimeSeriesBucket header accessors, and two investigated non-issues


lvca:
Follow-ups from the work on #6198 (merged as `8fcd89790`, PR #6261). Four separate items, kept in one
issue because they were all found by pulling on the same thread: the `openOrCreate` defect that the
review-requested regression test uncovered.

Items 1 and 2 are actionable. Items 3 and 4 are **investigated negative results**, recorded here on
purpose so the next person who notices the same smell does not spend the afternoon re-deriving that it
is already handled. Close them with the issue; they need no code.

---

## Item 1 - `getOrCreateFile()` can silently hand a component a file whose id is not its own

**Actionable.** This is the general trap behind the `openOrCreate` bug fixed in #6198, and nothing about
it is specific to the tag dictionary.

`PaginatedComponent`'s id-allocating constructor takes a fresh id from `FileManager.newFileId()`, bakes
it into the file path, and only then resolves the file
(`PaginatedComponent.java:76`):

```java
this.file = (PaginatedComponentFile) database.getFileManager().getOrCreateFile(name, filePath, mode);
```

But `getOrCreateFile` (`FileManager.java:426`) is keyed by **component name**, not by the path:

```java
ComponentFile file = fileNameMap.get(fileName);
if (file != null)
  return file;
```

So when that name is already registered, the caller gets back a file carrying a *different* id while
`Component.fileId` keeps the freshly minted one. The component then addresses pages of a file id that is
not the file it holds. In #6198 that surfaced as `File with id 2 was not found`, thrown from
`PageManager.loadPage` far away from the cause; a variant where the bogus id happened to resolve would be
considerably worse than an exception.

**Proposed fix.** Assert the invariant where it is established - in the `PaginatedComponent` constructor,
right after the `getOrCreateFile` call - and fail loudly:

```java
if (file.getFileId() != fileId)
  throw new IllegalStateException("Component '" + name + "' was built on file id " + fileId
      + " but the file registered under that name has id " + file.getFileId());
```

It costs one comparison per component construction and turns the whole bug class into an immediate,
local error instead of a confusing downstream failure.

**What I already checked**, so the work does not start from zero: the obvious risk is a legitimate path
that deliberately allocates a new id for an existing name. The LSM compaction and split paths are the
candidates, and they are safe - `LSMTreeIndexMutable.java:217` and `LSMTreeIndex.java:839` both construct
[...]

PR #6321, opened 2026-08-17
fix(#6314): a component reads its file at the stride that file was written with, and the by-id getOrCreateFile refuses a name it did not create

### LINKED ISSUE #6314: Follow-ups from #6283: TimeSeries components discard their file's page size, the by-id getOrCreateFile is unguarded, and a vector pool test asserts against its own contract
Three findings from the work on #6283 (merged as `dc341dcc4`, PR #6305), kept in one issue because they were
all turned up by pulling on the same thread: what a component is allowed to believe about the file it was
handed. They are independent of each other and can be fixed in any order.

Severity is not uniform. **Item 1 is a silent wrong-stride read** and is the one worth doing first. Item 2 is
hardening of an invariant that today lives in its caller rather than in the API. Item 3 is a flaky test that
asserts the opposite of what the code it tests promises.

Everything below was established by reading the sources at `dc341dcc4`; where I have *not* reproduced something
end-to-end, it says so.

---

## Item 1 - the TimeSeries components throw away the page size and version parsed from their own file name

**Actionable, and silent when it bites.**

`ComponentFactory` recovers the two facts that are baked into every component file's name and hands them to the
factory handler (`ComponentFactory.java:48-53`):

```java
final int pageSize = file instanceof PaginatedComponentFile pcf ? pcf.getPageSize() : 0;
final int version = file.getVersion();
...
return handler.createOnLoad(database, fileName, file.getFilePath(), fileId, mode, pageSize, version);
```

`LocalBucket`'s handler passes both straight through, which is what makes a reopen faithful to what was written
(`LocalBucket.java:267-273`):

```java
return new LocalBucket(database, name, filePath, id, mode, pageSize, version);
```

Both TimeSeries handlers accept the same two parameters and **discard them**
(`TimeSeriesBucket.java:136-142`, `TimeSeriesTagDictionary.java:150-156`):

```java
public PaginatedComponent createOnLoad(final DatabaseInternal database, final String name, final String filePath,
    final int id, final ComponentFile.MODE mode, final int pageSize, final int version) throws IOException {
  return new TimeSeriesBucket(database, name, filePath, id, new ArrayList<>(), null);   // pageSize, version dropped
}
```

The constructors they call re-derive the page size from the **live configuration** instead, and hard-code the
version (`TimeSeriesBucket.java:161`/`:175`, `TimeSeriesTagDictionary.java:166`/`:176`):

```java
super(database, name, filePath, id, ComponentFile.MODE.READ_WRITE,
    database.getConfiguration().getValueAsInteger(GlobalConfiguration.BUCKET_DEFAULT_PAGE_SIZE), CURRENT_VERSION);
```

So a database whose `arcadedb.bucketDefaultPageSize` is not the value that was in force when the type was created
reopens its `.tstb`/`.tstd` files at a stride that is not the one they were written with. The setting is
`SCOPE.DATABASE` and user-settable (`GlobalConfiguration.java:247`, default 64 KB), so this needs no bug to
trigger - only a config change between two runs, or one database opened with a `ContextConfiguration` that
differs from the one that created it.

Nothing downstream catches it. `PageManager` resolves pages with the **caller's** page size, i.e. the
component's, and the only place that consults the file's own `getPageSize()` is the snapshot path
(`PageManager.java:730`, `:741`); there is no agreement check between `PaginatedComponent.pageSize` and
`PaginatedComponentFile.getPageSize()` anywhere. `PaginatedComponent` then also computes
`pageCount = fileSize / pageSize` from the wrong divisor. The failure is a misaligned read of real bytes, not an
exception.

**Proposed fix**, in this order:

1. Pass the parsed values through, exactly as `LocalBucket` does, so the mismatch cannot be constructed. Both
   TimeSeries components need an id-and-page-size constructor for that; the version needs a moment's thought
   rather than a blind pass-through, because `TimeSeriesBucket`'s row layout is chosen by whether a dictionary
   is supplied, not by this field - the point is that the component should not *claim* `CURRENT_VERSION` for a
   file whose name says otherwise.
2. Then assert the invariant in `PaginatedComponent`'s constructor, next to the file-id check #6283 just added
   and in the same shape:

   ```java
   if (file.getPageSize() != pageSize)
     throw new IllegalStateException("Component '" + name + "' was built with page size " + pageSize
         + " but its file '" + file.getFilePath() + "' has page size " + file.getPageSize());
   ```

   This is the same invariant as the id one, on the other field: the component and the file it holds must agree
   about how to address it. Note it converts today's silent misread into a loud failure at open for any database
   that is *already* in the mismatched state, which is the right trade but should be a deliberate one.

**Not yet reproduced end-to-end.** The first task on this issue is a test that creates a TimeSeries type under
one `bucketDefaultPageSize`, reopens the database under another, and shows what the reads do.

---

## Item 2 - `FileManager.getOrCreateFile(int, String)` is the by-id mirror of #6283, and its invariant lives in its caller

**Actionable, hardening - no live defect found.**

#6283 fixed the by-*name* overload's hazard by asserting in `PaginatedComponent` that the file handed back
carries the id the component was built with. The by-*id* overload has the mirror shape and no such check
(`FileManager.java:445-460`):

```java
public ComponentFile getOrCreateFile(final int fileId, final String filePath) throws IOException {
  ComponentFile file = fileIdMap.get(fileId);
  if (file != null)
    return file;          // whatever name/path that file carries, the caller's filePath is not consulted
  ...
}
```

The one production caller is the HA follower's file-creation path, and **it already performs exactly this check
itself**, immediately before calling (`ArcadeStateMachine.createNewFiles`, `ArcadeStateMachine.java:1932-1941`):
a file id already registered under a different name throws `SchemaException` and is routed to quarantine plus a
full snapshot resync (added by #6063). So there is no reachable bug here today - I checked, and this is not a
"someone will hit it" claim.

What is worth fixing is that the invariant sits in the caller rather than in the API, one overload after the
sibling half was pushed down into the component layer. A second caller has to know to re-implement it, and the
two tests that call this overload document nothing about it.

**Proposed fix.** Push the name/path agreement check into the overload itself, so the guarantee belongs to
`FileManager` for both keys. The constraint to respect is the HA caller's behaviour: it needs its
`SchemaException` and the quarantine-and-resync handling that follows from it, so either the caller keeps its own
check on top (and `FileManager`'s becomes the belt-and-braces case for everyone else), or the pushed-down check
throws something the HA apply path classifies the same way. Do not simply move it and change the exception type.

---

## Item 3 - `LSMVectorIndexSearcherPoolTest` asserts a bound the pool documents it may exceed

**Actionable, flaky test in the default CI lane.**

`concurrentSearchesNeverShareASearcher` sets the pool size to 4, runs 200 searches across **8** threads, and then
asserts (`LSMVectorIndexSearcherPoolTest.java:128`, `:134`, `:147`):

```java
GlobalConfiguration.VECTOR_INDEX_SEARCHER_POOL_SIZE.setValue(4);
final ExecutorService pool = Executors.newFixedThreadPool(8);
...
assertThat(stat(index, "pooledGraphSearchers")).isLessThanOrEqualTo(4);
```

The statistic is `GraphSearcherPool.size()`, i.e. `idleCount` (`LSMVectorIndex.java:5898`), and the release path
says in so many words that this count may exceed `maxIdle` (`GraphSearcherPool.java:137-138`):

```java
// Racy by design: a transient overshoot of one or two searchers is cheaper than a lock on the search path.
if (idleCount.get() >= maxIdle) {
```

With 8 threads releasing into a pool of 4, several can read `idleCount` below the cap before any of them
increments, and the pool legitimately settles above it. The assertion therefore contradicts the contract of the
code under test.

Observed, not theorised: this failed as `expected 5L to be less than or equal to 4L` in a full `engine` suite run
while I was validating #6283, and passed both standalone and in a full baseline run of `main` from the same
commit - the concurrency it depends on is more likely to lose under full-suite CPU contention. The class carries
no `@Tag`, so it runs in the **default** lane and can redden an ordinary CI run.

**Proposed fix.** Assert what the pool actually promises rather than widening the number and moving on. The
property the test names - "never share a searcher" - is already covered by the per-task result comparison above
it; this last line is about the pool not growing without bound, so it should allow the documented overshoot
(e.g. `maxIdle + concurrency`, or a bound derived from the thread count) with a comment naming
`GraphSearcherPool.release()` as the reason. A bare bump to a magic number would leave the next person
re-deriving why 4 was not the answer.

lvca:
Fixes #6314.

Three findings from the work on #6283, independent of each other, kept together because they all pull on the same thread: what a component is allowed to believe about the file it was handed.

## Item 1 - the TimeSeries components threw away the page size and version parsed from their own file name

`ComponentFactory` recovers both facts from every component file's name and hands them to the factory handler. `LocalBucket`'s handler passes them straight through, which is what makes a reopen faithful to what was written. Both TimeSeries handlers accepted them and **dropped them**, re-deriving the page size from the *live* `arcadedb.bucketDefaultPageSize` and hard-coding `CURRENT_VERSION` instead. That setting is `SCOPE.DATABASE` and user-settable, so no bug was needed to trigger it: change it between two runs and every `.tstb`/`.tstd` file reopens at a stride that is not its own.

Nothing downstream catches that. `PageManager` resolves a page with the **caller's** page size - i.e. the component's - and the only place that consults the file's own is the snapshot path, so page N is read from `N * theWrongNumber`: real bytes at the wrong offset, not an exception, with `pageCount` computed from the wrong divisor on top.

The issue asked for this to be reproduced end to end first. It is, and it is worse than "misaligned":

> a type written at 64 KB and reopened at 16 KB returned **0 of 3,000 rows**

The header page starts at offset 0 and therefore still read correctly, which is exactly what let it announce data pages that were then read out of the padding inside real page 0.

**Fix.** Both handlers pass page size, version and mode through. `TimeSeriesTagDictionary.openOrCreate()` had the same hazard by a second route the issue did not name - it built on an already-registered file's *id* but on the *configured* page size - and now takes all three from that file, through a constructor that makes it the only available form for a caller in that position. `PaginatedComponent` then asserts the agreement in its constructor, next to the file-id check #6283 added and in the same shape.

On the version: it is now what the file says rather than `CURRENT_VERSION`. Which row layout a bucket reads is decided by whether a dictionary was handed to it, not by this field, so nothing changes behaviourally - the point is that the component should not *claim* the current version for a file whose name says otherwise.

**On the trade the issue flagged.** The issue asked for a deliberate decision about the guard turning an existing mismatched database into "a loud failure at open". Having built it, that framing turns out not to apply to *this* change, and I want to correct it rather than repeat it: the guard compares the component against the file it holds, and after the pass-through fix **every load path takes the page size from that file**, so the two agree by construction whatever `bucketDefaultPageSize` happens to say. The guard cannot fire on a database opened through the normal path - it is a programming-error tripwire for the defect this change removes, not a compatibility gate. The issue's framing described adding the guard *without* the pass-through.

`reopeningUnderAnyConfiguredPageSizeSucceedsWhateverTheFilesWereWrittenAt` pins this: a database carrying files written at 64 KB and at 16 KB, reopened under a third value matching neither, opens both and reads both.

What the fix genuinely cannot undo is data *written* during a mismatched session under an affected build - those pages went to disk at the wrong stride and are scrambled on disk. No guard added here detects that, and none could.

## Item 2 - `FileManager.getOrCreateFile(int, String)` is the by-id mirror of #6283

The invariant lived in its one caller. It is pushed down, so "the file you get back is the file you asked for" belongs to `FileManager` for both keys rather than to whichever caller remembered to check.

Respecting the constraint the issue set out: it throws the same `SchemaException` the HA follower's `createNewFiles` does, so a caller that reaches it is classified exactly as that caller would have classified itself. `createNewFiles` **keeps** its own check, because it has to tell an idempotent replay (same name, skip) from a diverged file-id space (different name, quarantine and full snapshot resync, #6063) - that is a branch, not just a guard, and cannot be replaced by an exception. So `FileManager`'s check fires for nobody today; it is what makes the next caller not have to re-derive the reasoning. No reachable defect existed here and none is claimed.

## Item 3 - `LSMVectorIndexSearcherPoolTest` asserted a bound the pool documents it may exceed

The statistic is `GraphSearcherPool`'s `idleCount`, and `release()` reads it and increments without a lock - *"racy by design: a transient overshoot of one or two searchers is cheaper than a lock on the search path."* With 8 threads releasing into a pool of 4, all 8 can read a count below the cap and then all increment, so the reachable maximum is `maxIdle + concurrency - 1` = 11. The assertion of `<= 4` contradicted the contract of the code under test; it was observed failing as `expected 5L to be less than or equal to 4L` in a full engine suite run, and the class carries no `@Tag`, so it reddens the **default** lane.

Per the issue's request, this asserts what the pool actually promises - that it does not grow without bound - with the two magic numbers named (`POOL_SIZE`, `CONCURRENCY`), the derivation written out, and `GraphSearcherPool.release()` named as the reason. Not a bump to a larger literal. "Never share a searcher", the property the test is named for, is already covered by the per-task result comparison above it.

## Files changed

Main:
- `engine/src/main/java/com/arcadedb/engine/PaginatedComponent.java` - page-size agreement guard
- `engine/src/main/java/com/arcadedb/engine/FileManager.java` - by-id `getOrCreateFile` name check
- `engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java` - pass-through handler + constructor
- `engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesTagDictionary.java` - pass-through handler, constructor, and `openOrCreate()`

Tests:
- `engine/src/test/java/com/arcadedb/engine/timeseries/Issue6314TimeSeriesPageSizeTest.java` (new)
- `engine/src/test/java/com/arcadedb/engine/Issue6314FileAgreementTest.java` (new)
- `engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexSearcherPoolTest.java`
- `engine/src/test/java/com/arcadedb/engine/timeseries/TimeSeriesTagDictionaryTest.java` - moved to the file-carrying constructor

## Verification

- Full `engine` suite (`-DexcludedGroups=benchmark,slow,vector`): **12,340 tests, 0 failures, 0 errors**
- `ha-raft` state machine suite (the only production caller of the changed `FileManager` API): **96 tests, 0 failures**
- The item 1 test fails on `main` with `expected: 65536 but was: 16384` and, with that assertion removed, returns 0 rows

https://claude.ai/code/session_013az14F8gvbcYips5GqiZfy

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/FileManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/FileManager.java
@@ -18,6 +18,7 @@
  */
 package com.arcadedb.engine;
 
+import com.arcadedb.exception.SchemaException;
 import com.arcadedb.log.LogManager;
 
 import java.io.File;
@@ -442,10 +443,27 @@ public class FileManager {
     }
   }
 
+  /**
+   * The by-<em>id</em> mirror of {@link #getOrCreateFile(String, String, ComponentFile.MODE)}, and it has the same
+   * hazard: it is keyed by the id alone, so an id that is already registered hands that file back whatever
+   * {@code filePath} the caller asked for, and the caller then goes on using a file that is not the one it named.
+   * The by-name half of that pair was closed at the component layer by issue #6283; this one is closed here
+   * (issue #6314), so the guarantee "the file you get back is the file you asked for" belongs to
+   * {@code FileManager} for both keys rather than to whichever caller remembered to check.
+   * <p>
+   * A caller that has to tell "already applied" apart from "diverged" still needs its own branch on the two, and
+   * the HA follower's {@code ArcadeStateMachine.createNewFiles} has one: a matching name is an idempotent replay
+   * it skips, a differing one is a file-id space that has diverged from the leader's, which it turns into the
+   * quarantine-and-resync path (issue #6063). This check therefore fires for nobody today; it is what makes the
+   * next caller not have to re-derive that reasoning. It throws the same {@link SchemaException} that caller does,
+   * so a caller that does reach it is classified exactly as it would have classified itself.
+   *
+   * @throws SchemaException when {@code fileId} is already registered under a different file name
+   */
   public ComponentFile getOrCreateFile(final int fileId, final String filePath) throws IOException {
     ComponentFile file = fileIdMap.get(fileId);
     if (file != null)
-      return file;
+      return checkFileNameMatches(file, filePath);
 
     synchronized (this) {
       file = fileIdMap.get(fileId);
@@ -453,10 +471,25 @@ public class FileManager {
         file = new PaginatedComponentFile(filePath, mode);
         registerFile(file);
         recordCreate(file);
-      }
+      } else
+        checkFileNameMatches(file, filePath);
 
       return file;
     }
+  }
+
+  /**
+   * Asserts that an already-registered file is the one the caller named. The comparison is on the file name and
+   * not on the whole path: the id, page size, version and extension a component addresses its file through all
+   * live in that name, while the directory prefix is the database path both sides already share.
+   */
+  private static ComponentFile checkFileNameMatches(final ComponentFile file, final String filePath) {
+    final String requestedName = new File(filePath).getName();
+    if (!file.getFileName().equals(requestedName))
+      throw new SchemaException(
+          "File id " + file.getFileId() + " is already registered as '" + file.getFileName() + "' but was requested as '"
+              + requestedName + "' ('" + filePath + "')");
+    return file;
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/engine/PaginatedComponent.java
+++ b/engine/src/main/java/com/arcadedb/engine/PaginatedComponent.java
@@ -94,6 +94,18 @@ public abstract class PaginatedComponent extends Component {
           "Component '" + name + "' was built on file id " + fileId + " but the file registered under that name has id "
               + file.getFileId() + " ('" + file.getFilePath() + "')");
 
+    // The same invariant as the id one, on the other field a component uses to address its file (issue #6314): the
+    // page size IS the stride, so a component that disagrees with its file reads page N from N * theWrongNumber.
+    // PageManager resolves every page with the CALLER's page size, and only the snapshot path ever consults the
+    // file's own, so nothing downstream turns that into an exception - it is real bytes at the wrong offset, and
+    // pageCount below is computed from the wrong divisor on top. The page size is baked into the file name, so this
+    // can only be a component that re-derived it from somewhere else (a live configuration value, say) instead of
+    // taking the one ComponentFactory parsed off the name.
+    if (file.getPageSize() != pageSize)
+      throw new IllegalStateException(
+          "Component '" + name + "' was built with page size " + pageSize + " but its file '" + file.getFilePath()
+              + "' has page size " + file.getPageSize());
+
     final long fileSize = file.getSize();
     if (fileSize == 0)
       // NEW FILE, CREATE HEADER PAGE
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesBucket.java
@@ -131,13 +131,19 @@ public class TimeSeriesBucket extends PaginatedComponent {
 
   /**
    * Factory handler for loading existing .tstb files during schema load.
-   * Columns are set later via {@link #setColumns(List)} when the TimeSeries type is initialized.
+   * Columns are set later via {@link #setColumns(List, TimeSeriesTagDictionary)} when the TimeSeries type is
+   * initialized.
+   * <p>
+   * The page size, version and mode {@code ComponentFactory} recovered from the file name are passed straight
+   * through, exactly as {@code LocalBucket}'s handler does (issue #6314). Re-deriving the page size from the live
+   * {@code bucketDefaultPageSize} instead would address every page of an already-written file at a stride that is
+   * not the one it was written with, which is a misaligned read of real bytes and not an exception.
    */
   public static class PaginatedComponentFactoryHandler implements ComponentFactory.PaginatedComponentFactoryHandler {
     @Override
     public PaginatedComponent createOnLoad(final DatabaseInternal database, final String name, final String filePath,
         final int id, final ComponentFile.MODE mode, final int pageSize, final int version) throws IOException {
-      return new TimeSeriesBucket(database, name, filePath, id, new ArrayList<>(), null);
+      return new TimeSeriesBucket(database, name, filePath, id, mode, pageSize, version, new ArrayList<>(), null);
     }
   }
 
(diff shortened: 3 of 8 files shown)


lvca:
Thanks - on the one non-blocking point raised (a release-note callout or repair path for an already-affected database that would now hard-stop at open): I checked it and it does not apply to this change, so there is nothing to document.

The guard compares the component against **the file it holds**, and after the pass-through fix every load path takes the page size *from that file* - `ComponentFactory` parses it off the name and both TimeSeries handlers now pass it through, and `openOrCreate()` takes it off the registered file. So the two agree by construction whatever `bucketDefaultPageSize` says at open time. The guard can only fire on a component that re-derived the value from somewhere else, which is precisely the defect being removed: it is a programming-error tripwire, not a compatibility gate.

The inference is a very natural one, and worth noting that the issue text made it too - it described adding the guard *without* the pass-through, and I repeated that framing in the original PR description. Both are now corrected.

Pushed `reopeningUnderAnyConfiguredPageSizeSucceedsWhateverTheFilesWereWrittenAt` to state it rather than leave it to be re-derived: one database carrying files written at 64 KB and at 16 KB, reopened under a third value matching neither, opens both and reads both back.

Worth being precise about what *does* survive: data **written** during a mismatched session on an affected build went to disk at the wrong stride and is scrambled there. No guard in this change detects that, and none could - it is not a state the component can tell apart from correct bytes. That is a real residual, just not one an open-time check or a rename can address.

PR #6326, opened 2026-08-17
fix(#6311, #6312, #6313): a hop checks its target against what the row holds when it runs, and a relabelled vertex takes its aliases with it

### LINKED ISSUE #6311: ArcadeDB 26.8.1: `WITH *` changes `OPTIONAL MATCH` cardinality
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1` 

## Problem

Adding a semantically identity-preserving `WITH *` after an `OPTIONAL MATCH` with comma-separated patterns changes the result cardinality. The two patterns share `n1`; the projection does not add, remove, or aggregate any variable.

## Steps to reproduce

Run the following Cypher on an empty, disposable `Test` database. The setup and both queries can be executed in order; the queries are read-only.

```cypher
MATCH (n) DETACH DELETE n;

CREATE (a:l1 {tag:'a'})<-[:R0 {k7:false,k8:true}]-(n0 {tag:'n0a',k2:true})<-[:rt0]-(x)<-[:R2]-(n1 {tag:'A'});
CREATE (a:l1 {tag:'b'})<-[:R0 {k7:false,k8:true}]-(n0 {tag:'n0b',k2:true})<-[:rt0]-(x)<-[:R2]-(n1 {tag:'B'});
MATCH (n1 {tag:'A'}) CREATE (n2 {tag:'n2a'})<-[:rt8 {k3:'d',k10:-1957561185}]-(y)-[:R3]->(n1);
MATCH (n1 {tag:'B'}) CREATE (n2 {tag:'n2b'})<-[:rt8 {k3:'d',k10:-1957561185}]-(y)-[:R3]->(n1);
```
// Control query: expected row_count = 2.
```
OPTIONAL MATCH (:l1)<-[r0 {k7:false,k8:true}]-(n0 {k2:true})<-[r1:rt0]-()<-[r2]-(n1),
  (n2)<-[:rt8 {k3:'d',k10:-1957561185}]-()-[]->(n1)
RETURN count(*) AS row_count;
```
// Variant: WITH * is an identity projection here; expected row_count = 2.
```
OPTIONAL MATCH (:l1)<-[r0 {k7:false,k8:true}]-(n0 {k2:true})<-[r1:rt0]-()<-[r2]-(n1),
  (n2)<-[:rt8 {k3:'d',k10:-1957561185}]-()-[]->(n1)
WITH *
RETURN count(*) AS row_count;
```

## Expected result

Both queries return one row with `row_count = 2`. `WITH *` should not change the rows or the bindings.

## Actual result

- Control query: `row_count = 2`
- Query with `WITH *`: `row_count = 4`

### LINKED ISSUE #6312: ArcadeDB 26.8.1: `REMOVE` after `OPTIONAL MATCH` and `UNWIND` raises `RecordNotFoundException`
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1` 

## Steps to reproduce

Run the following Cypher on an empty, disposable `Test` database.

```cypher
MATCH (n) DETACH DELETE n;

CREATE (n2:l2:l11 {id:2,klist:['n2v']})-[:rt2]->
       (n1 {id:1,k11:'u'})-[:R {klist:[-1365794787,-485798932]}]->
       (n0 {id:0,k1:'v',klist:[1,2,3]});
MATCH (n2 {id:2})
CREATE (:l11:l8:l1:l3:l5:l4:l9)-[:t]->(n2);

OPTIONAL MATCH p0 = (n2)-[:rt2]->(n1 {k11:'u'})-[{klist:[-1365794787,-485798932]}]->(n0),
  (:l11:l8:l1:l3:l5:l4:l9)-[]->(n2)
WITH *
WHERE n0 IS NOT NULL
UNWIND coalesce(n0.klist, []) AS alias3
WITH *
WHERE n2 IS NOT NULL
REMOVE n0.k1, n2.klist, n2:l2:l11
RETURN count(*) AS row_count;
```

## Expected result

The query completes and returns `row_count = 3` after unwinding `[1, 2, 3]`. Removing a property and labels from the matched vertices should not invalidate the current row.

## Actual result

The query fails deterministically on a fresh database with a server-side exception equivalent to:

```text
com.arcadedb.exception.RecordNotFoundException: Record #7:0 not found
  at com.arcadedb.database.immutable.ImmutableDocument.checkForLazyLoading
  at com.arcadedb.database.immutable.ImmutableVertex.modify
  at com.arcadedb.query.sql.executor.RemoveStep.removeProperty
```

### LINKED ISSUE #6313: ArcadeDB 26.8.1: `REMOVE` label after `OPTIONAL MATCH` raises an edge-list `RecordNotFoundException`
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1` 
## Steps to reproduce

Run the following Cypher on an empty, disposable `Test` database.

```cypher
MATCH (n) DETACH DELETE n;

CREATE (n0 {k9:true})-[:rt7]->(n1:l1 {tag:'middle'})-[:X]->(n2 {tag:'end'});
MATCH (n1 {tag:'middle'}) CREATE (n1)-[:c]->(:Extra);

UNWIND ['x'] AS alias0
OPTIONAL MATCH (n2)<-[]-(n1)<-[:rt7]-(n0), p0=(n2)
WHERE true
WITH *
WHERE n1 IS NOT NULL AND n0 IS NOT NULL
REMOVE n0.k9, n1:l1
RETURN count(*) AS row_count;
```

## Expected result

The query completes and returns `row_count = 1`. Removing `n1:l1` should update the label set without corrupting or losing the vertex's edge records.

## Actual result

The query fails deterministically on a fresh database with a server-side exception equivalent to:

```text
Cannot load OUT edge list chunk (#5:0) for vertex #4:0
com.arcadedb.exception.RecordNotFoundException: Record #5:0 not found
  at com.arcadedb.graph.GraphEngine.getEdges
  at com.arcadedb.query.sql.executor.RemoveStep.removeLabels

### PR DESCRIPTION
**Pull request by lvca**
Fixes #6311, fixes #6312, fixes #6313.

## #6311 - `WITH *` changed the cardinality of a multi-pattern MATCH

`WITH *` is an identity projection, so nothing downstream of it should be able to tell it is there. It could:

```cypher
MATCH (n0 {k2:true})<-[:rt0]-()<-[]-(n1), (n2)<-[:rt8]-()-[]->(n1)
RETURN n0.tag, n1.tag, n2.tag        -- 2 rows, joined on n1
```
```cypher
MATCH (n0 {k2:true})<-[:rt0]-()<-[]-(n1), (n2)<-[:rt8]-()-[]->(n1)
WITH *
RETURN n0.tag, n1.tag, n2.tag        -- 4 rows, the full Cartesian product
```

The issue reports it on `OPTIONAL MATCH`, but `OPTIONAL` is incidental: plain MATCH breaks the same way, and so does any following clause, not just `RETURN count(*)`.

The join between two comma-separated patterns of one MATCH is the identity check each hop runs on its target against the names the row already carries. Those names were handed to `MatchRelationshipStep` as the planner's **live, mutable** `boundVariables` set. That set is filled with the clause's own variables only *after* the clause is planned (`boundVariables.addAll(matchVariables)`), which is what made the check work at all - retroactively, through the shared reference. A following WITH clears the same set, and `WITH *` refilled it with the literal name `*`, so by the time the hop executed the shared variable was gone from it and every target was accepted.

Two changes:

- the hop is given a **snapshot** of what is bound at the point it will run - everything bound before this MATCH plus everything this MATCH has bound so far (earlier comma-separated patterns, earlier hops). A variable can only be in a row if it was bound earlier, so this is exactly the set the check needs, and it no longer depends on a set someone else may edit later;
- `WITH *` forwards the incoming scope instead of replacing it with `"*"`. An explicit projection list still resets it, and `WITH *, expr AS alias` does both.

## #6312 / #6313 - a label write left every other reference to the node dangling

ArcadeDB reads a record's type from the bucket it lives in, and in the Cypher mapping a label set *is* a type, so `REMOVE n:L` cannot retype in place: it writes a new record and deletes the old one. Every other reference to that node was left pointing at the deleted original - a second alias in the same row, or the same node reached again by a later row of the same clause (UNWIND fan-out, or simply two matches sharing the node). The next `REMOVE` then either failed to load the record:

```
com.arcadedb.exception.RecordNotFoundException: Record #7:0 not found
  at com.arcadedb.database.immutable.ImmutableDocument.checkForLazyLoading
```

or, on the way to deleting it a second time, followed its dangling edge-list chunk and reported the vertex as concurrently modified:

```
Edge list OUT of vertex #4:0 is not fully visible yet (concurrent commit in flight): Record #5:0 not found
```

`SetStep` already tracked its own replacements for `SET n:L`; `RemoveStep` tracked nothing. Both now share one `LabelReplacements`, which performs the rewrite and points every alias of the node - in this row and in every later one - at the live record, so a repeated label removal becomes an idempotent no-op instead of a fatal one.

Two defects of the rewrite itself, shared by SET and REMOVE, are fixed in the same place:

- **edge properties were silently dropped.** The old code re-linked edges with `newVertex.newEdge(edge.getTypeName(), other)`, i.e. by type alone, so `-[:R {klist:[...]}]->` came back as `-[:R]->`. They are now carried over (and a LIGHTWEIGHT type, which has none, still gets none).
- **a self-loop was duplicated**, migrated once by the OUT pass and again by the IN pass. It is now migrated once.

### On the expected values in the reports

- #6312 asks for `row_count = 3` and gets it.
- #6313 asks for `row_count = 1`, but the query matches **2** rows: the fixture gives the middle node two outgoing edges (`[:X]->(n2)` and the `[:c]->(:Extra)` added by the second `CREATE`), and `(n2)<-[]-(n1)` matches both. The test asserts 2. What was reported - the exception - is real and fixed.

## Tests

`CypherWithStarScopeIssue6311Test` (5) and `CypherLabelWriteRowFanoutIssue6312Test` (5), both new. Reverting the fix fails 7 of the 10; the remaining 3 are controls that must keep passing (explicit `WITH n0, n1, n2`, `WITH *` before a MATCH, `WITH *, expr AS alias`).

Verification run: the whole `engine` unit lane, 15,730 tests, green; the openCypher TCK suite, 3,897 scenarios, green.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -2019,6 +2019,11 @@ public class CypherExecutionPlan {
             // is planned - which is what used to supply this clause's own variables - and a following WITH
             // clears it, so the join on a variable shared between two patterns of the SAME MATCH silently
             // degraded into a Cartesian product the moment a `WITH *` followed.
+            //
+            // The snapshot also carries this hop's own target and relationship names, added just above. That is
+            // deliberate and costs nothing: the check reads a name only when the row ALSO already holds a vertex
+            // under it, and a name this hop is about to bind for the first time is absent from the row. Excluding
+            // them would instead need the set rebuilt per hop in the opposite order, for no gain.
             final Set<String> targetIdentityVars = new HashSet<>(boundVariables);
             targetIdentityVars.addAll(matchVariables);
 
@@ -2611,8 +2616,8 @@ public class CypherExecutionPlan {
                   // Fixed-length relationship - pass path variable, target node pattern, and bound variables.
                   // #6311: the same snapshot rule as the ordered builder above - the hop identity-checks its
                   // target against what the row carries when it RUNS, so it gets the names bound before this
-                  // MATCH plus the ones this MATCH has bound so far, never the planner's live set (which is
-                  // filled at the end of the clause and emptied again by a following WITH).
+                  // MATCH plus the ones this MATCH has bound so far (this hop's own included, harmlessly - see
+                  // the note there), never the planner's live set, which a following WITH empties.
                   final Set<String> targetIdentityVars = new HashSet<>(legacyBoundVariables);
                   targetIdentityVars.addAll(matchVariables);
                   nextStep = new MatchRelationshipStep(currentSourceVar, relVar, targetVar, relPattern, pathVariable,
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -1262,11 +1262,7 @@ public class CypherExecutionPlan {
             entryIndex += 2; // skip both the next OPTIONAL MATCH and the WITH clause
             // Update boundVariables from the WITH clause
             final WithClause nextWith = ((ClauseEntry) clausesInOrder.get(entryIndex)).getTypedClause();
-            boundVariables.clear();
-            for (final ReturnClause.ReturnItem item : nextWith.getItems()) {
-              final String alias = item.getAlias();
-              boundVariables.add(alias != null ? alias : item.getExpression().getText());
-            }
+            applyProjectionToScope(nextWith.getItems(), boundVariables);
             break;
           }
 
@@ -1278,11 +1274,7 @@ public class CypherExecutionPlan {
             entryIndex++; // skip the WITH clause (already handled)
             // Update boundVariables from the WITH clause
             final WithClause nextWith = ((ClauseEntry) clausesInOrder.get(entryIndex)).getTypedClause();
-            boundVariables.clear();
-            for (final ReturnClause.ReturnItem item : nextWith.getItems()) {
-              final String alias = item.getAlias();
-              boundVariables.add(alias != null ? alias : item.getExpression().getText());
-            }
+            applyProjectionToScope(nextWith.getItems(), boundVariables);
             break;
           }
         }
@@ -1292,12 +1284,8 @@ public class CypherExecutionPlan {
       case WITH:
         final WithClause withClause = entry.getTypedClause();
         currentStep = buildWithStep(withClause, currentStep, context, functionFactory);
-        // WITH resets the scope: only WITH output variables are in scope afterwards
-        boundVariables.clear();
-        for (final ReturnClause.ReturnItem item : withClause.getItems()) {
-          final String alias = item.getAlias();
-          boundVariables.add(alias != null ? alias : item.getExpression().getText());
-        }
+        // An explicit WITH resets the scope to its own output variables; WITH * forwards the incoming one
+        applyProjectionToScope(withClause.getItems(), boundVariables);
         break;
 
       case MERGE:
@@ -1652,6 +1640,33 @@ public class CypherExecutionPlan {
   }
 
   /**
+   * Applies a {@code WITH} projection to the set of variables in scope after it.
+   * <p>
+   * An explicit projection list resets the scope to exactly what it names, which is what makes a variable the
+   * following clauses do not import unavailable. {@code WITH *} names nothing: it forwards the incoming scope
+   * unchanged, so the names stay bound (#6311). Treating its single {@code "*"} item as if it were a projected
+   * variable used to leave the scope holding the literal name {@code *} and nothing else, which is how a
+   * variable shared with a following MATCH stopped being recognised as already bound - and a shared variable
+   * that is not recognised is a Cartesian product rather than a join. A {@code WITH *, expr AS alias} both
+   * forwards and adds.
+   */
+  private static void applyProjectionToScope(final List<ReturnClause.ReturnItem> items, final Set<String> scope) {
+    boolean forwardsAll = false;
+    final List<String> projected = new ArrayList<>(items.size());
+    for (final ReturnClause.ReturnItem item : items) {
+      final String alias = item.getAlias();
+      final String name = alias != null ? alias : item.getExpression().getText();
+      if ("*".equals(name))
+        forwardsAll = true;
+      else
+        projected.add(name);
+    }
+    if (!forwardsAll)
+      scope.clear();
+    scope.addAll(projected);
+  }
+
+  /**
    * Builds execution step for a MATCH clause.
    * Backward-compatible overload without bound variable tracking.
    */
@@ -1997,6 +2012,16 @@ public class CypherExecutionPlan {
             // Check if this hop requires IN traversal on a unidirectional edge.
             // Unidirectional edges don't store incoming links, so we must restructure:
             // instead of (bound)-[IN]->(target), scan target type and go (target)-[OUT]->(bound).
+            // #6311: the names a hop must identity-check its target against are the ones the row already
+            // carries when the hop RUNS: everything bound before this MATCH plus everything this MATCH has
+            // bound so far (earlier comma-separated patterns, earlier hops). Snapshot them here rather than
+            // handing the step the planner's live `boundVariables` set: that set is mutated after the clause
+            // is planned - which is what used to supply this clause's own variables - and a following WITH
+            // clears it, so the join on a variable shared between two patterns of the SAME MATCH silently
+            // degraded into a Cartesian product the moment a `WITH *` followed.
+            final Set<String> targetIdentityVars = new HashSet<>(boundVariables);
+            targetIdentityVars.addAll(matchVariables);
+
             final Direction effectiveDir = directionOverride != null ? directionOverride : relPattern.getDirection();
             final boolean needsReverseOnUnidirectional = !reversed
                 && effectiveDir == Direction.IN
@@ -2006,7 +2031,7 @@ public class CypherExecutionPlan {
             if (needsReverseOnUnidirectional) {
               // Restructure: scan target type with MatchNodeStep, then traverse OUT to validate
               // against the bound source. The bound source becomes the "target" of the relationship.
-              final Set<String> boundWithSource = new HashSet<>(boundVariables);
+              final Set<String> boundWithSource = new HashSet<>(targetIdentityVars);
               boundWithSource.add(effectiveSourceVar);
               final MatchNodeStep scanStep = new MatchNodeStep(effectiveTargetVar, effectiveTargetNode, context);
               if (isOptional && matchChainStart == null) {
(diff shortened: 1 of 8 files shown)


lvca:
Thanks - all three taken.

**1. `buildExecutionStepsLegacy`** - fixed rather than annotated. You're right that it's unreachable for parsed queries today (`clausesInOrder` is always populated by `StatementBuilder`), but this is exactly the "optimizer and legacy execution paths - changes to clause handling may need updates in multiple paths" landmine `CLAUDE.md` warns about, and leaving one of the two paths holding the pre-fix shape is how it comes back. It now takes the same `boundVariables ∪ matchVariables` snapshot.

**2. `redirect()` and composite values** - real, and worse than a crash risk: it answers *stale* rather than failing. `MATCH p = (a:Keep:Drop)-[:E]->(b) REMOVE a:Drop RETURN [n IN nodes(p) | labels(n)]` returned `[[Drop, Keep], []]` - the label set the node no longer has - because the path still held the deleted record. `redirect()` now also walks `TraversalPath`, `List` and `Map` values and rebuilds them around the live records, allocating only when something inside actually moved (a row with nothing to redirect still costs the one `isEmpty()` check). Two new tests, `aBoundPathFollowsItsRelabelledMembers` and `aCollectedListFollowsItsRelabelledMembers`; both fail without the change.

**3. `SetStep.applyLabels`'s manual resolve** - kept, and `RemoveStep.removeLabels` now matches it, because it is not quite redundant: `redirect()` can only reach aliases the row *lists* in `getPropertyNames()`, while resolving the write target reads it directly and does not depend on that. Both sites now carry a comment saying so, so the next reader doesn't have to re-derive it.

Re-verified: engine unit lane green, openCypher TCK 3,897 scenarios green.

lvca:
Second pass addressed - thanks, the variable-length question was the right one to ask.

**Variable-length paths** - immune by construction, and now pinned. `ExpandPathStep` resolves an already-bound target from the incoming row at runtime (`lastResult.getProperty(targetVariable)`), not from a plan-time variable set, so there was never a live set to be emptied under it. `computePrevVarsForVlp` builds only the relationship-uniqueness *exclusion* set, which is the pre-clause scope by definition (the analogue of `previousStepVariables`, which must stay pre-clause). Added `withStarKeepsTheJoinOnAVariableLengthPattern`: same shared-variable join over `[:rt8*1..2]`, with and without the `WITH *`, both correct.

**The `resolve()` comment** - you're right, and I couldn't demonstrate the gap either: the write target is always read as a row column, so `redirect()` does reach it. Kept the call but rewrote both comments to say what is actually true - it makes each method correct on its own terms rather than on its caller's, at one map lookup that is skipped entirely until a label write happens.

**Performance trade** - correct reading, and deliberate: the map that knows a node moved is step-wide, so once any row replaces something every later row pays the walk. Now written down at `redirect()`, along with the other half - a clause with no label write pays one `isEmpty()` check per row and never walks anything.

**Tests** - both suggestions taken. `addLabelIsIdempotentAcrossRowsAndCountsOnce` mirrors the REMOVE stats test for `SET n:Label`. New `LabelReplacementsTest` covers the class directly: double-replacement chaining, edge and edge-property carry-over, lightweight-edge handling, the self-loop counted once, every row shape redirected, and the negative case where nothing moved and the collection must come back as the *same instance* rather than a rebuilt copy.

Re-verified: Cypher lane 7,037 green, openCypher TCK 3,897 green.

lvca:
Third pass done.

**Lightweight edges** - turned out to be fixable rather than an accepted limitation, so they're fixed. The premise that a lightweight edge has no independent identity is half right: it has no *address*, but `LightEdgeRID` carries the `(type, out vertex, in vertex)` triple and `RID.hashCode()` folds the endpoints in for any record-less RID, so it keys the replacement map exactly like a heavy edge's address does - and the copy re-attached to the replacement vertex is a *different* key rather than the same one. Dropped the `getPosition() > -1` guard; a row holding a lightweight edge now follows the endpoint that moved instead of answering with a vertex that no longer exists. Covered at both levels: `LabelReplacementsTest.redirectRewritesALightweightEdgeOntoTheReplacement` (fails without the change) and `aBoundLightweightEdgeFollowsTheEndpointThatMoved` through Cypher.

**`targetIdentityVars` includes the hop's own names** - correct, and the comment now says so plus why it's free: the check reads a name only when the row *also* already holds a vertex under it, and a name the hop is about to bind for the first time is absent from the row. Excluding them would need the set rebuilt per hop in the opposite order for no gain.

**`RemoveStep.removeProperty()` alias asymmetry** - agreed it's real and agreed it's not this PR's: it changes REMOVE's observable semantics rather than fixing a crash, and it's unrelated to the three issues here. Filed as #6328 with a repro and the suggested shared-helper fix.

**Double `redirect()` per row** - noted, no change: the second call is the one that runs right after `replace()`, and it's what makes the remaining items of the same clause see the new record. Skipping it would mean the first call has to know a replacement is coming.

Re-verified after the change: Cypher lane 7,039 green, openCypher TCK 3,897 green.

lvca:
Good catch on `MergeStep` - fixed here rather than after, since it's the same defect on the same route.

It was the **third** copy of the pre-fix code, and you're right that its self-loop handling is worse than the other two were: the OUT pass re-created the loop against `edge.getVertex(DIRECTION.IN)`, which is still the record about to be deleted, so a relabelled node with a self-loop came out with an edge pointing at nothing. Edge properties were dropped there too, and nothing redirected aliases - not even the `propagateUpdateToSameNodeAliases` `SetStep` already had, so a node the input row bound under a second name was simply left on the deleted record.

`ON CREATE SET`/`ON MATCH SET n:Label` now takes the shared `LabelReplacements`, one instance per running step, threaded through `executeMerge` → `applySetClause` exactly as `SetStep` and `RemoveStep` scope theirs to the running `ResultSet`. That brings all four things with it: the edges carried over with their properties, the self-loop migrated once and onto the *replacement*, the row's aliases redirected, and the cross-row idempotence.

Two tests: `mergeOnMatchSetLabelKeepsEdgesAndRedirectsEveryAlias` (the `MATCH (m)` + `MERGE (n)` second-alias case - fails without the change) and `mergeOnMatchSetLabelSurvivesTheSameNodeOnSeveralRows`.

On the `"*"` sentinel note - agreed it's not new fragility; it's the same representation the pre-existing code keyed on, just read instead of copied into the scope.

Re-verified: Cypher + MERGE lanes 7,085 green, openCypher TCK 3,897 green.

lvca:
Taken - the guard now says what it guards. You're right that it read backwards: the `getBucketId() > -1` half looked like it was *excluding* lightweight edges when it does the opposite, and it can't be false for an edge that came out of `getEdges()` anyway. Reduced to the null check, with the comment leading on the fact that lightweight edges are keyed here like any other.

That's the last open point from my side: the three reported defects are fixed with regression tests that fail without them, the two latent ones the shared rewrite exposed (edge properties dropped, self-loop duplicated - dangling, in MERGE's copy) are fixed with it, and the one genuinely out-of-scope item is filed as #6328.

Final verification on this head: Cypher + MERGE lanes 7,085 green, openCypher TCK 3,897 green, full engine unit lane green.

PR #6329, opened 2026-08-17
fix(#6192): route every executeCommand failure through GrpcErrorMapper

### LINKED ISSUE #6192: gRPC executeCommand flattens every failure into success=false, so the client loses the exception type and never retries a NeedRetryException
Found while implementing #6183 (PR #6187), which made the leader refusal the one failure `executeCommand` reports as a real gRPC error. Everything else still goes out in the response envelope, and the type is lost on the way.

## The defect

`ArcadeDbGrpcService.executeCommandInternal` catches every exception and returns a response instead of failing the call (line 790):

```java
} catch (Exception e) {
  LogManager.instance().log(this, Level.SEVERE, "ERROR in executeCommandInternal", e);
  ...
  return ExecuteCommandResponse.newBuilder()
      .setSuccess(false)
      .setMessage(e.getMessage() == null ? e.toString() : e.getMessage())
      ...
}
```

The client turns that back into one untyped exception (`RemoteGrpcDatabase.command`, line 543):

```java
if (!response.getSuccess()) {
  throw new DatabaseOperationException("Failed to execute command: " + response.getMessage());
}
```

So the exception type only survives as English inside a message string.

## What it costs

- **`transaction()` never retries a command.** `RemoteGrpcDatabase.transaction()` retries on `NeedRetryException`. A `ConcurrentModificationException` raised by a command arrives as `DatabaseOperationException`, which is not one - so the retry fires for a commit and silently does not for a command. That is the exact behaviour `GrpcErrorMapper` was built to preserve (see its class javadoc).
- **A duplicate key is untyped.** `DuplicatedKeyException` carries an index name and the offending keys, and `GrpcErrorMapper` already ships both Base64-encoded on trailers for the RPCs that use it. Through `executeCommand` a caller gets a string.
- **It diverges twice over**: from every other gRPC RPC on the service, and from the HTTP protocol, which ships the exception class name so `RemoteDatabase` can rebuild the type.

`ErrorHandlingIT.duplicateKey_throwsDuplicatedKeyException` currently asserts `satisfiesAnyOf(DuplicatedKeyException, DatabaseOperationException)` - the test was written to tolerate exactly this gap.

## Why #6183 did not just fix it

#6183 needed the leader refusal to carry redirect trailers, and trailers only exist on an error - there is nowhere in `ExecuteCommandResponse` to put the address of another node. So that one exception is rethrown and mapped, and the envelope was left alone for everything else, deliberately: routing all failures through `GrpcErrorMapper` changes what a **non-Java** client sees. Anything reading `success`/`message` today would start receiving a gRPC error status instead. That is a wire-behaviour change and deserves its own issue rather than a rider on a routing fix.

## Suggested fix

Send failures through `GrpcErrorMapper.toStatusRuntimeException(cause, "ExecuteCommand", ha())` and `resp.onError(...)`, as `createRecord`, `beginTransaction` and `commitTransaction` already do. The client's `handleGrpcException` then rebuilds the exact type, and retry-on-`NeedRetryException` starts working for commands.

Points to settle before doing it:

1. **Compatibility.** `success=false` is public wire behaviour. Either accept the break for the next minor (it is the direction every other RPC already went), or keep the envelope for a deprecation window behind a request flag or server setting. Worth deciding explicitly rather than by accident.
2. **`success` becomes dead weight.** If every failure is an error status, the field is always `true` on a response that arrives at all. Leave it set for compatibility; do not add new meanings to it.
3. **Guard the double-close.** The catch currently runs after `resp.onNext(...)`/`onCompleted()` in the success path; calling `onError` after `onCompleted` throws `IllegalStateException`. A `responded` flag makes the transition safe - the existing code has the same latent exposure with `onNext(err)`, so this is worth fixing in passing either way.
4. **Fix the tests that were written around the gap**, notably `ErrorHandlingIT.duplicateKey_throwsDuplicatedKeyException`, which should then assert `DuplicatedKeyException` outright, with its index name and keys.
5. `executeQuery`/`streamQuery` map to a bare `Status.INTERNAL` rather than through the mapper too (`Query execution failed: ...`). Same family; worth folding into the same change so the service ends up with one error path.

robfrank:
## What does this PR do?

`ArcadeDbGrpcService.executeCommand` flattened every command failure - except the
one leader-refusal case #6183 needed for its redirect trailers - into an
`ExecuteCommandResponse` with `success=false` and the exception's message as free
text, instead of failing the RPC. The client turned that back into a single
untyped `DatabaseOperationException`, erasing the exception type on the wire.

Every failure now goes through `GrpcErrorMapper.toStatusRuntimeException(...)` and
`resp.onError(...)`, the same path `createRecord`, `beginTransaction`,
`commitTransaction` and `graphBatchLoad` already use. The client's existing
`handleGrpcException`/`GrpcClientErrorMapper` already reconstructs the exact
engine exception type from the gRPC status + trailers, so this is a server-only
change.

A `responded` guard was added around the success path's `onNext`/`onCompleted` so
the catch block can never call `onError` after a response has already been fully
delivered - the double-close exposure the issue flagged in passing (point 3).

## Motivation

Closes #6192.

Concretely, before this fix:
- `RemoteGrpcDatabase.transaction()`'s retry on `NeedRetryException` never fired
  for a command (only for a commit), because a `ConcurrentModificationException`
  raised by a command arrived client-side as an untyped `DatabaseOperationException`.
- `DuplicatedKeyException`'s index name and offending keys - already shipped on
  trailers by `GrpcErrorMapper` for every other RPC that uses it - were lost for
  `executeCommand`.
- `executeCommand` diverged from every other gRPC RPC and from the HTTP protocol,
  both of which route failures through a typed error path.

## Related issues

Closes #6192. Builds on the `GrpcErrorMapper`/redirect-trailers work from #6183.

## What was intentionally left out (per the issue's own "points to settle")

- **Wire compatibility**: `success=false` in-band was public wire behaviour; this
  breaks it for the next minor. The issue itself recommends accepting the break
  rather than adding a deprecation flag, since every other RPC already made this
  move - no compatibility flag was added.
- **`success` field**: left in the proto/response as-is (always `true` on a
  response that arrives at all now); no new meaning was added to it.
- **`executeQuery`/`streamQuery`** still map unexpected failures to a bare
  `Status.INTERNAL` rather than through `GrpcErrorMapper` (the issue's point 5).
  That's a distinct code path with its own existing case-by-case status
  preservation (`RESOURCE_EXHAUSTED`, `FAILED_PRECONDITION`, `NOT_FOUND`, ...);
  left untouched here to keep this PR's blast radius to the RPC the issue is
  titled after. Worth a follow-up if the whole service should end up on one
  error path.

## Additional notes

Five existing tests were written to pin the pre-fix `success=false`-in-band
behaviour and are updated to assert a `StatusRuntimeException` instead (each with
a comment pointing back at #6192):
- `Issue6183FollowerCommandRoutingIT.anOrdinaryCommandFailureSurfacesAsAGrpcError`
  (renamed from `...StillComesBackInTheResponse`)
- `Issue5040GrpcTransactionHijackIT.crossTenantExecuteCommandIsDenied`
- `Issue4794GrpcPerDbAuthorizationIT.crossDatabaseCommandIsDenied`
- `GrpcServerIT.executeCommandInvalidSqlReturnsError`
- `GrpcTransactionScriptingAuthorizationIT.readerCannotEscalateViaJsInsideExternalTransaction`
  / `.readerCannotWriteViaSqlInsideExternalTransaction`

`ErrorHandlingIT.duplicateKey_throwsDuplicatedKeyException` no longer tolerates
`DatabaseOperationException` as an acceptable outcome (it was written to tolerate
exactly the gap this issue describes); it now asserts `DuplicatedKeyException`
with `getIndexName()`/`getKeys()` populated - proving the type and the trailers
survive an `executeCommand` failure over gRPC end to end.

`TimeSeriesGrpcHaConcurrentInsertIT` and `Issue4804GrpcCommandErrorMetricsTest`
were checked and need no code change: the former already collects both an
in-band failure and a thrown exception into the same `errors` list; the latter
drives `GrpcMetricsInterceptor` directly with a hand-built response rather than
exercising `executeCommand`, so it's independent of this change (its doc comment
was updated to say so, since the in-band scenario it pins no longer occurs for a
real command error).

## Test plan

- [x] `ErrorHandlingIT` (grpc-client): 8/8, including the rewritten
      `duplicateKey_throwsDuplicatedKeyException`.
- [x] `GrpcErrorMapperTest`: 11/11 (unaffected, mapper-level unit test).
- [x] `Issue4804GrpcCommandErrorMetricsTest`: 3/3 (unaffected, interceptor-level unit test).
- [x] `GrpcServerIT`: 30/30, including the updated `executeCommandInvalidSqlReturnsError`.
- [x] `Issue4794GrpcPerDbAuthorizationIT`: 4/4, including the updated `crossDatabaseCommandIsDenied`.
- [x] `GrpcTransactionScriptingAuthorizationIT`: 3/3, including both updated tests.
- [x] `Issue5040GrpcTransactionHijackIT`: 5/5, including the updated `crossTenantExecuteCommandIsDenied`.
- [x] `Issue6183FollowerCommandRoutingIT`: 2/2 on a 3-node Raft cluster, including the renamed/updated test.
- [x] Full `grpcw`/`grpc-client` test-compile (all 6 edited test files + main change compile clean).
- [x] Repo-wide `grep -rn "getSuccess()).isFalse()"` across `grpcw`/`grpc-client` tests to find every
      consumer of the old in-band-failure behaviour; every hit is one of the files updated above.

## Checklist

- [x] I have run the build using `mvn clean package` command (scoped to the affected modules; see test plan)


robfrank:
Thanks for the review. Addressed in e6ea0a3:

**Applied** - "the dead `!response.getSuccess()` branch this PR makes unreachable": kept
the branch as a canary rather than removing it - it now logs at `SEVERE` before throwing,
so a future server-side regression that reintroduces the in-band `success=false` envelope
fails loudly in logs instead of silently resurrecting the untyped-exception bug #6192 fixed.

**Not changed, with rationale:**
- Checked-exception path in `executeCommandInternal`'s rethrow: verified `Database.command`/
  `begin`/`commit`/`rollback` declare no checked exceptions, so the `RuntimeException` wrapping
  branch is unreachable in practice, matching the identical pattern already used in this file's
  `streamQuery`. No code change.
- Pinning exact `Status.Code` in the updated integration tests: the review itself called this
  "not required given `GrpcErrorMapperTest` already covers the mapping table in isolation," so
  left as-is.

robfrank:
Thanks, both points addressed in 20da16d:

**Applied** - "the `responded` guard doesn't close the window it's meant to close": you're
right, `responded = true` was set after `onCompleted()` returned rather than after `onNext()`
succeeded. Moved it to right after `onNext(response)`, so a concurrent-cancel making
`onCompleted()` throw can no longer fall through to `resp.onError(...)` on an already-delivered
call.

**Documented, not changed** - "checked-exception fallback re-erases the exception type": added
a comment on the `RuntimeException` wrapping branch in `executeCommandInternal` explaining it's
unreachable today (every engine exception is unchecked) but would re-erase the type via
`GrpcErrorMapper.unwrap`'s single-level `ExecutionException` unwrap if it ever did fire. Agreed
this is fine to leave as a documented gap rather than teaching `unwrap` to peel an arbitrary
wrapper, given how narrow the trigger condition is.

The "no test exercises the `responded` guard itself" note is acknowledged as non-blocking per
your own framing (hard to reach at the IT level this PR uses); left as-is.

robfrank:
Thanks for the thorough re-review - no further action needed on this cycle. Ready for maintainer merge.

PR #6330, opened 2026-08-17
fix(#6190): SQLFunction gets a determinism marker the planner can use

### LINKED ISSUE #6190: Nothing marks a SQL function as pure, so the planner cannot cache a plan that contains one, cannot fold it, and calls it while planning
The residue of #6179 and #6174, both of which worked around the same missing piece rather than supplying it.

### The gap

There is no purity or determinism marker anywhere on `SQLFunction` / `SQLMethod` / `RecordFunction`. (`com.arcadedb.function.StatelessFunction` in the sibling registry means "does not need the current record", the same notion as `Expression.isEarlyCalculated` - not purity.) So the planner cannot tell `concat('a','b')` from `uuid()`, and three places pay for it:

1. **No plan caching for any statement with a function in the WHERE.** `FunctionCall.isCacheable()` returns true only for graph traversals (`out`, `in`, `both`, ...), so `SelectStatement.executionPlanCanBeCached()` is false for `WHERE x = concat(:a, :b)` and the plan is rebuilt on every execution. That predicate is doing duty as a purity check it was never designed to be.
2. **No constant folding of a function over literals.** `Expression.isLiteral()` (#6181) deliberately refuses functions, so `WHERE 1 = length('abc')` is evaluated per row even though its value is fixed at parse time. Documented there as a conscious limitation, waiting on exactly this marker.
3. **A function in an indexed equality is invoked once at plan time.** `BinaryCondition.isIndexAware` evaluates the right-hand side to check for an empty collection, so `WHERE name = uuid()` calls `uuid()` while planning and again to build the index key. Harmless for a pure function; for anything else it is an invocation the query never asked for. #6179 removed two other sites like it by narrowing them to literals, but this one legitimately needs the value.

### Suggested direction

A `isDeterministic()` (or `isPure()`) on `SQLFunction` and `SQLMethod`, with `FunctionCall`/`MethodCall` consulting it. The design question worth settling first is the **default**, because it decides the risk profile:

- default `true` - every built-in is assumed pure, the handful that are not (`uuid`, `sysdate`, `sysdate()`-like clock readers, anything random or counter-backed) override to false. Maximum benefit, but a user-defined function with a side effect silently becomes foldable and cacheable.
- default `false`, pure built-ins opt in. Safe for user code by construction; costs an audit of the built-in registry, and a pure user function stays pessimised unless its author opts in.

I lean to **default false with built-ins opting in**: the wrong answer in the other direction is silent and reaches user data, and the audit is a bounded one-time job over `DefaultSQLFunctionFactory` / `DefaultSQLMethodFactory`. Worth deciding deliberately rather than by whichever is less typing.

Once it exists, all three sites above become straightforward: `isCacheable()` can admit pure functions, `isLiteral()` can grow a "constant-folding" sibling that admits a pure call over literals, and `isIndexAware` can skip the plan-time probe for anything not marked pure.

### Not urgent

Nothing here is a correctness bug today - #6179 established that the one path where a plan-time evaluation could have been baked into a cached plan (partition pruning) is now literal-only, so it no longer depends on the `isCacheable()` accident. This is about the optimisations the missing marker forecloses, and about removing a coupling that two separate PRs have now had to write comments about.

robfrank:
Closes #6190

## Summary

Nothing on `SQLFunction` distinguished a pure function (`abs`, `pow`) from one that reads a random source, the clock, or the database (`uuid()`, `sysdate()`), so three places paid for the missing marker: the planner could not cache a plan whose `WHERE` contained any function call, constant folding refused to evaluate a function even over literal arguments, and a function in an indexed equality was invoked at plan time whether or not that was safe.

This adds `SQLFunction#isDeterministic()` (default `false`, matching the issue's "pure built-ins opt in" direction - the safe default when a user-defined function silently becoming foldable/cacheable would be a bad-in-the-wrong-direction, hard-to-notice defect), marks the built-ins verified to read only their own arguments (`abs`, `pow`, `sqrt`, `coalesce`, `ifnull`, `ifempty`, `if`, `decode`, `strcmpci`), and wires the marker into two of the three sites the issue names:

- **`FunctionCall#isCacheable()`**: a call to a deterministic built-in over cacheable arguments no longer forces the whole statement out of the plan cache. Previously *any* function call other than a graph traversal (`out`/`in`/`both`/...) did.
- **`Expression#isFoldable()`**: the "constant-folding sibling" `isLiteral()`'s own javadoc reserved for exactly this marker (see the comment on `Expression#isLiteral()` predating this PR). `BinaryCondition#isAlwaysTrue`/`isAlwaysFalse` now fold a comparison like `1 = abs(-1)`, without ever invoking the function merely to classify the expression - `isFoldable()` only consults `isDeterministic()`, a compile-time-constant getter.

### Excluded from the deterministic list, deliberately

- **`encode()`**: it dereferences a RID and reads the record store when given one, so its result is not a pure function of its literal arguments - the same class of hazard the issue warns a wrong `true` produces.
- **`format()`**: `String.format` resolves grouping/decimal-separator and similar conversions against the JVM's process-wide mutable default `Locale`, which is state outside the arguments.
- **`concat()`** (the `DefaultSQLFunctionFactory` one): it's actually a multi-row aggregate (`SQLAggregatedFunction`), not the two-arg literal-concatenation the issue's example suggests - never a candidate.

### Site 3 (`BinaryCondition#isIndexAware`) is intentionally left unchanged

The issue's third example - `WHERE name = uuid()` invoking `uuid()` once at plan time and once to build the index key - is real, but `isIndexAware`'s eager evaluation is gated on `Expression#isEarlyCalculated()`, which (by design, see issue #6179) also admits a bound parameter. `isFoldable()` correctly *excludes* bound parameters, matching `isLiteral()`'s "never bake this execution's binding into a cached plan" requirement. Reusing `isFoldable()` at that site would silently drop the empty-collection short-circuit for the ordinary `WHERE indexed = :param` shape - a real regression on what is probably the single most common indexed-equality query, in exchange for skipping one extra invocation of a non-deterministic function, which is not a correctness bug today (the issue says as much: "Not urgent... nothing here is a correctness bug"). That site needs its own predicate ("early-calculated minus non-deterministic function calls") rather than reusing either existing one; deferred as a follow-up rather than rushed into this PR.

### Also deferred

- Auditing `SQLMethod` (the issue names it alongside `SQLFunction`) and wiring `MethodCall#isCacheable()` the same way - mechanically identical to what's done here, but a second ~60-entry registry audit (`DefaultSQLMethodFactory`) that deserves its own PR rather than doubling this one's surface.
- Extending the built-in `isDeterministic()` audit to the geo/vector function families - plausibly pure but not individually verified here to the same standard as the ~10 marked in this PR.
- The unified `FunctionRegistry`/`StatelessFunction` catalog (text/math/convert/date/util functions reached through a different code path than `DefaultSQLFunctionFactory`) is untouched; a call resolved only through it is conservatively non-deterministic by default, same as before this PR.

## Test plan

- [x] `Issue6190FunctionDeterminismTest` (new): pins which built-ins are marked deterministic, that a non-deterministic/RID-reading/locale-sensitive function is never marked, that an unknown/user-defined function is conservatively not cacheable, that a deterministic call is cacheable only when its own arguments are, that the pre-existing graph-traversal cacheability exemption is untouched, that `isFoldable()` accepts literals and nested deterministic calls but rejects a property/parameter argument or a modifier applied to the call's result, and that `isFoldable()` stays strictly narrower than `isEarlyCalculated()`.
- [x] `ConstantFalseFilterFoldingTest` (extended): end-to-end - `WHERE abs(-1) = 999` is folded to no-scan/no-rows like a literal comparison would be, the pre-existing `uuid()` non-fold regression test is untouched, a statement with a deterministic function in the filter is confirmed in the execution plan cache, and the mirror case with `uuid()` is confirmed still excluded from it.
- [x] `mvn -pl engine -am compile` - clean.
- [x] `mvn -pl engine -am test` over `com.arcadedb.query.sql.parser.**`, `com.arcadedb.query.sql.executor.**`, `com.arcadedb.query.sql.function.**`, and `com.arcadedb.function.**` (excluding benchmark/slow/vector lanes) - 1884+ tests, 0 failures, 0 errors, including `Issue6179LiteralExpressionTest`, `Issue6179EarlyCalculatedModifierTest`, `PartitionPruningPlannerTest`, and `SelectStatementTest`, the existing tests closest to this change.

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionAbsoluteValue.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionAbsoluteValue.java
@@ -144,4 +144,12 @@ public class SQLFunctionAbsoluteValue extends SQLFunctionMathAbstract {
   public Object getResult() {
     return result;
   }
+
+  /**
+   * Pure math over its single argument, including the deterministic overflow failure (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionPow.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionPow.java
@@ -75,4 +75,12 @@ public class SQLFunctionPow extends SQLFunctionMathAbstract {
   public Object getResult() {
     return result;
   }
+
+  /**
+   * Pure math over its two arguments (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionSquareRoot.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/math/SQLFunctionSquareRoot.java
@@ -75,4 +75,12 @@ public class SQLFunctionSquareRoot extends SQLFunctionMathAbstract {
   public Object getResult() {
     return result;
   }
+
+  /**
+   * Pure math over its single argument (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionCoalesce.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionCoalesce.java
@@ -79,4 +79,12 @@ public class SQLFunctionCoalesce extends SQLFunctionAbstract {
   public String getSyntax() {
     return "Returns the first not-null parameter or null if all parameters are null. Syntax: coalesce(<field|value> [,<field|value>]*)";
   }
+
+  /**
+   * A pure selection among its arguments (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionDecode.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionDecode.java
@@ -72,4 +72,13 @@ public class SQLFunctionDecode extends SQLFunctionAbstract {
   public String getSyntax() {
     return "decode(<value>, <format>)";
   }
+
+  /**
+   * A pure base64/base64url decode of its string argument - unlike {@link SQLFunctionEncode}, it never dereferences
+   * a RID (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionIf.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/misc/SQLFunctionIf.java
@@ -86,4 +86,12 @@ public class SQLFunctionIf extends SQLFunctionAbstract {
   public String getSyntax() {
     return "if(<field|value|expression>, <return_value_if_true> [,<return_value_if_false>])";
   }
+
+  /**
+   * A pure selection among its arguments (issue #6190).
+   */
+  @Override
+  public boolean isDeterministic() {
+    return true;
+  }
 }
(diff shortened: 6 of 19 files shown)


robfrank:
Re the spin-wait flagged as possibly-dead in `aStatementWithADeterministicFunctionInTheFilterIsCacheable`: it's not dead. `SelectExecutionPlanner.createExecutionPlan()` (line ~301) only stores into the plan cache when `db.getExecutionPlanCache().getLastInvalidation() < planningStart` - both `System.currentTimeMillis()` values. `beginTest()` creates the `Character` type immediately before each test runs, which invalidates the plan cache; without the spin-wait, that invalidation and this test's `planningStart` can land in the same millisecond, making `lastInvalidation < planningStart` false and the `contains()` assertion flaky. The sibling test a few methods down (`aStatementWithAConstantFalseFilterIsCacheable`) already has this exact guard with a comment explaining it - this one was just missing the comment. Added it in d4c1089a22 rather than deleting the guard.

PR #6347, opened 2026-08-18
fix(#6340): a component takes its file's mode too, the version joins the guards it belongs with, and CHECK DATABASE can finally read a TimeSeries file

### LINKED ISSUE #6340: Follow-ups from #6314: a component still cannot read its file's mode, the by-name getOrCreateFile ignores the mode asked for, the version is the one file field left unguarded, and CHECK DATABASE cannot see TimeSeries at all
Four findings from the work on #6314 (merged as `bcd910687`, PR #6321), kept in one issue because three of
them are the same shape and the fourth is what the third one's residual damage would need in order to be
found. They are independent of each other and can be fixed in any order.

Severity is not uniform. **Item 4 is the only one that leaves a real defect undetectable**; items 1-3 are
hardening and API completion, with no reachable defect found for any of them - I looked, and that is stated
rather than assumed. Everything below was established by reading the sources at `bcd910687`.

Items 1-3 share one theme, and it is the theme #6283 and #6314 have been working through: **a component and
the file it holds must agree, and the agreement should live in the API rather than in whoever remembered to
check.** #6283 closed the `fileId` half. #6314 closed the `pageSize` half and pushed the by-id name check
down into `FileManager`. What is below is what is left of that set.

---

## Item 1 - `ComponentFile` does not expose the mode it was opened with

**Actionable, API completion - no live defect found.**

A component built on an already-registered file can now take the id, the page size and the version *from
that file* - that is what #6314 fixed. It cannot take the mode, because there is no accessor
(`ComponentFile.java:37`):

```java
protected final MODE mode;      // written at :52, never read back out by anyone
```

`getFileId()`, `getVersion()`, `getFileName()`, `getFilePath()`, `getComponentName()`,
`getFileExtension()` and `PaginatedComponentFile.getPageSize()` all exist. `getMode()` does not.

The consequence shows up in the constructor #6314 added for exactly this purpose
(`TimeSeriesTagDictionary.java:191`), where every other field comes from the file and the mode is guessed:

```java
this(database, name, file.getFilePath(), file.getFileId(), ComponentFile.MODE.READ_WRITE, file.getPageSize(),
    file.getVersion());
```

`MODE.READ_WRITE` there is a hard-coded assumption sitting in the middle of five values that are read off
the file. It is correct today - see item 2 for why - but it is correct by accident of what the rest of the
system happens to do, not because the caller checked.

**Proposed fix.** Add `getMode()` and use it in that constructor, so the "build on an existing file" form
takes *every* file property from the file. One accessor, and the shape of the constructor stops being
"four from the file and one from a guess".

---

## Item 2 - the by-name `getOrCreateFile` ignores the caller's requested mode on a hit

**Actionable, hardening - no live defect found.**

`FileManager.java:427`. It is keyed by the component name, and on a hit it returns the registered file
whatever mode that file was opened with; the `mode` parameter is consulted only on the miss path that
constructs one:

```java
public ComponentFile getOrCreateFile(final String fileName, final String filePath, final ComponentFile.MODE mode)
    throws IOException {
  ComponentFile file = fileNameMap.get(fileName);
  if (file != null)
    return file;                                  // mode not consulted
  ...
    file = new PaginatedComponentFile(filePath, mode);
```

This is the third member of the family. The same overload used to ignore the caller's *id* expectation
(#6283, closed in `PaginatedComponent`), the by-id overload used to ignore the caller's *name* expectation
(#6314 item 2, closed in `FileManager`), and it still ignores the caller's *mode* expectation.

**Why nothing breaks today, stated so the next person does not have to re-derive it.** `FileManager.open()`
registers every file with the manager's own mode, which is the database's mode, so every component asking
later gets the mode it would have asked for anyway. And mode is narrow in effect: it only selects the
`RandomAccessFile` open string (`PaginatedComponentFile.java:463`, `mode == MODE.READ_WRITE ? "rw" : "r"`).
So a mismatch would not corrupt anything - it would surface as an OS-level write failure on a read-only
channel, or as a write permitted on a component that asked for read-only.

That second direction is the one worth noting: asking for `READ_ONLY` and being handed a `READ_WRITE` file
is a silently *weaker* guarantee than the caller requested, which is the direction this family of bugs has
always gone.

**Proposed fix.** Decide and then say which it is, because both are defensible and the current code says
neither: either the mode is a property of the *file* (in which case the parameter is misleading on the hit
path and the Javadoc should say the first registration wins), or it is a request the caller is entitled to
(in which case a mismatch should throw the way the id and name mismatches now do). Item 1's `getMode()` is
what makes the second option expressible at all.

---

## Item 3 - `PaginatedComponent` guards the id and the page size, but not the version

**Actionable, completes the set - no live defect found, and verified that the guard would be free.**

Three facts are baked into every component file's name: `name.fileId.pageSize.vVersion.ext`. Two of them
are now asserted against the file the component ends up holding (`PaginatedComponent.java:92` and `:104`):

```java
if (file.getFileId() != fileId)      // #6283
if (file.getPageSize() != pageSize)  // #6314
```

The third is not. `version` is passed to `super(...)` at `:71` and stored, and nothing ever compares it to
`file.getVersion()` - even though #6314 was specifically about making the TimeSeries components stop
claiming `CURRENT_VERSION` for a file whose name said otherwise. The guard that would have caught that
class of mistake directly is the one still missing.

**Verified that adding it would not trip anything.** This is the part worth having checked before filing:

- Every id-carrying (load) constructor passes the parsed version straight through - all eleven of them:
  `LocalBucket:406`, `Dictionary:143`, `HashIndexBucket:192`, `LSMTreeIndexAbstract:196`,
  `LSMTreeIndexBloomFilter:149`, `SparseSegmentComponent:72`, `LSMVectorIndexMutable:74`,
  `LSMVectorIndexCompacted:83`, `LSMVectorIndexGraphFile:87`, `TimeSeriesBucket:186`,
  `TimeSeriesTagDictionary:180`. Not one hard-codes a version on a load path any more.
- Every creation path bakes the version into the name it generates and passes the same value as the
  component's version (`PaginatedComponent.java:65`), so the two agree by construction:
  ```java
  this(database, name, filePath + "." + id + "." + pageSize + ".v" + version + "." + ext, id, mode, pageSize, version);
  ```
- A file written by an older build carries that build's version in its name and the load path passes it
  through, so an old database agrees too.

So the guard is free: it can only fire on a component that re-derived the version from somewhere other than
its file, which is precisely the defect #6314 removed and precisely what nobody should reintroduce.

**Proposed fix.** Add it next to the other two, in the same shape and with the same kind of message. Note
this is a tripwire and not a compatibility gate, for exactly the reason the #6314 discussion had to work
through twice: because every load path takes the version *from* the file, the component and the file agree
by construction and no existing database can trip it.

---

## Item 4 - `CHECK DATABASE` has no TimeSeries coverage at all

**Actionable, and the only item here that leaves a real defect invisible.**

`DatabaseChecker` contains **zero** occurrences of `TimeSeries`, `tstb` or `tstd`. It walks record buckets
and indexes; TimeSeries data lives in `.tstb`/`.tstd` `PaginatedComponent`s that are registered with the
schema as *files* (`TimeSeriesShard.java:134`, `TimeSeriesTagDictionary.java:249`) rather than as a type's
record buckets, so nothing in the checker ever reads them. `LocalTimeSeriesType extends LocalDocumentType`,
so such a type falls into the checker's `documentTypes` bucket at `DatabaseChecker.java:211` and then has
no record buckets to scan.

**Why this is being filed now.** #6314 fixed a bug that wrote real data to disk at the wrong stride. The fix
stops it happening again, and the new guard stops a component ever disagreeing with its file - but neither
can undo pages that were *already written* during a mismatched session on an affected build. Those bytes
are at offsets nothing will look for again. I said so on PR #6321 and it is worth restating here: there is
no way to detect that state today, because the tool whose job that is does not read these files.

This is bigger than one bug's residue. TimeSeries now has three on-disk formats of its own - the mutable
bucket, the sealed store, and the tag dictionary, the last two with their own headers, magics and CRCs - and
none of them is covered by the database integrity check. Every other storage component in the engine is.

**Proposed fix**, and this is a design question rather than a patch, which is why it is listed last:

1. The cheap, high-value part first - validate what the formats already assert about themselves: page 0's
   magic (`TSBC` / `TSTD`), the format version byte, and the header's counters against the file's actual
   page count. The bug this issue descends from would have been caught by the last of those, since a
   file read at the wrong stride has a header page count that cannot be reconciled with its size.
2. Then decide whether the checker should walk TimeSeries rows at all, and what "repair" would even mean
   for an append-only sealed store - which is a real question and not a rhetorical one, and probably wants
   its own discussion before anyone writes code.

Worth deciding explicitly whether item 1 lands as part of the ordinary `CHECK DATABASE` scope or behind an
opt-in, since scanning every sample of a large TimeSeries type is not the same cost as validating a header.

---

### Verification status

Items 1, 2 and 3 were verified by reading the merged sources and, for item 3, by checking all eleven load
constructors and both creation paths individually. Item 4's "zero references" is a literal grep result. No
item here has been reproduced as a live failure, because for items 1-3 no live failure exists to reproduce
and for item 4 the whole point is that the failure is undetectable - the reproduction for that one is the
one already in `Issue6314TimeSeriesPageSizeTest`, run against a build without the #6314 fix.

lvca:
Closes #6340.

Four items. The first three are the last members of a set issues #6283 and #6314 have been closing one at a
time - a component and the file it holds must agree, and the agreement belongs to the API rather than to
whoever remembered to check. The fourth is the one that left a real defect undetectable.

## Item 1 - `ComponentFile.getMode()`

Added, and used where it was missing: `TimeSeriesTagDictionary`'s build-on-an-existing-file constructor took
the id, the page size and the version off the file and then hard-coded `MODE.READ_WRITE` in the middle of
them, because there was nothing to read the mode back out of. All four now come from the file.

## Item 2 - the by-name `getOrCreateFile` and the caller's mode

The issue asked for a decision, so here it is stated rather than implied: **the mode is a request the caller
is entitled to**, and a hit that cannot satisfy it now throws instead of handing back something else. The
direction that decided it is the quiet one - a caller asking for `READ_ONLY` and being given a `READ_WRITE`
file gets a *weaker* guarantee than it asked for, and mode is the one file property whose whole purpose is to
be a guarantee.

Reopening the file to satisfy the request was considered and rejected: a registered file is shared by every
component addressing it, so upgrading the channel would silently widen it under a reader that had asked for
the narrower one - the same shape being removed, with a worse failure mode.

`IllegalStateException`, not the `SchemaException` its by-id sibling raises: a file *name* mismatch is a
file-id space that diverged from the leader's and an HA follower classifies it as quarantine-and-resync; a
*mode* mismatch is a programming error, on the same footing as the file-id and page-size guards in
`PaginatedComponent`, which is this overload's only caller and which throws exactly this.

Nothing reaches it today, and that is checked rather than assumed - the one caller that legitimately hits a
registered file is the tag dictionary constructor above, which now asks for the file's own mode.

## Item 3 - the version guard

`PaginatedComponent` asserted the file id (#6283) and the page size (#6314) against the file it ends up
holding; the version, the third fact baked into `name.fileId.pageSize.vVersion.ext`, was not. It decides how
the pages are *interpreted* where the other two decide where they are - a `LocalBucket` version selects the
record-header layout, a `TimeSeriesBucket` version selects whether a TAG column is a 4-byte dictionary id or
an inline string - so a disagreement is a misread of real bytes, never an exception.

**A tripwire, not a compatibility gate**, and worth repeating because #6314 had to work through the same
point twice: every load constructor passes the parsed version straight through and every creation path bakes
the version into the name it generates, so a component and its file agree by construction whatever build
wrote the file. `aComponentBuiltOnTheVersionItsFileNameCarriesOpensNormally` pins that.

## Item 4 - `CHECK DATABASE` can now read a TimeSeries file

`DatabaseChecker` had zero references to TimeSeries. It walks record buckets and indexes; a TimeSeries type
has neither - its shards are registered with the schema as *files*, and its compacted data never goes through
the paginated layer - so the type fell into the document arm, found no bucket to scan, and all three of the
formats TimeSeries owns were outside the reach of the only tool whose job is to find damage in them.

Each format now validates itself, in the shape `IndexInternal.checkIntegrity()` already uses:

- **`TimeSeriesBucket`** - page 0's magic, format version against the file name, column count against the
  schema, and then every counter page 0 declares reconciled against the data pages themselves: the sample
  count, the min and the max timestamp, and that the data pages it announces are actually in the file. That
  last set is what makes the residue of #6314 visible: a session that wrote at the wrong stride put real rows
  at offsets nothing will address again and counted them in a header that still does. Page headers only, no
  row decoding.
- **`TimeSeriesTagDictionary`** - page 0's magic and version, and the entries walked the way `load()` walks
  them, so the declared entry count and the bytes have to agree. A truncated dictionary does not fail a query;
  it makes every tag written since the damage read back as `null`, on every row.
- **`TimeSeriesSealedStore`** - header, block directory, offsets against the file length, a trailing region
  that belongs to no block, and **the per-block CRC32**. That last one is the expensive part and it is the
  point: the CRC is verified lazily on first read, so a block nothing queries is a block nothing verifies.
  It is recomputed from the file rather than delegated to `validateBlockCRC()`, so a second `CHECK DATABASE`
  in the same process cannot answer "clean" without having read a byte.

Rows are deliberately *not* decompressed and there is **no `FIX` arm**. A record bucket can be repaired
because its records are self-describing and its indexes derive from them; a sealed store is append-only
columnar data whose blocks are the only copy, so "repair" means deciding which samples to discard - the
design question the issue itself flagged, which wants an answer before code. This change makes the state
visible, which is the part that was missing entirely.

Reported as `corruptedTimeSeries` plus warnings, with `totalTimeSeriesTypes/Shards/Samples/SealedBlocks`
seeded on every run so "was this looked at?" is answerable from a clean result rather than from silence.
Scoped by `TYPE` like every other per-type pass. One progress step for all TimeSeries types and only when the
database has one, so a database without them keeps the step plan every existing expectation was written
against.

## Two things found on the way, neither fixed here

- `BlockEntry.blockStartOffset` and `storedCRC` are populated by `loadDirectory()` alone. A block appended by
  the running process carries both on disk and zero in the fields, with `crcValidated` pre-set to true.
  Nothing reads them for such a block today, so it is latent rather than a defect - but it is why the check
  reads both sides from the file instead of trusting the directory.
- The sealed-store CRC pass reads the whole sealed file. That is the same cost class as the record scan
  `checkBuckets` already runs over every bucket, but on a very large TimeSeries type it is the dominant cost
  of a `CHECK DATABASE`, and whether it should become opt-in is worth deciding once someone has one.

## Verification

- Two new regression tests, 11 cases: `Issue6340ComponentFileAgreementTest` (items 1-3, both the refusal and
  the legitimate case for each guard) and `Issue6340TimeSeriesCheckDatabaseTest` (item 4 - the healthy walk
  with its counters, the #6314 residue shape, a missing bucket header, a missing dictionary header, a flipped
  byte in a sealed block caught twice in a row, the `TYPE` scope, and a database with no TimeSeries type).
- Full `engine` module: 12,474 tests green, which includes every existing TimeSeries test running the new pass
  in its teardown integrity check.
- `ha-raft` `ArcadeStateMachineCreateFilesTest` and the `server` read-only/check-database subset green.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/ComponentFile.java
+++ b/engine/src/main/java/com/arcadedb/engine/ComponentFile.java
@@ -137,6 +137,24 @@ public class ComponentFile {
     return version;
   }
 
+  /**
+   * The mode this file is open in, which selects the {@code RandomAccessFile} open string and nothing else.
+   * <p>
+   * Exposed (issue #6340) so that a component built on an <em>already-registered</em> file can take every file
+   * property from that file rather than four from the file and one from a guess. The id, the page size and the
+   * version already had accessors, and each of them is now taken from the file by the callers in that position
+   * (issues #6283 and #6314); the mode was the only one left with nothing to read it back out of, which is why
+   * {@code TimeSeriesTagDictionary}'s build-on-an-existing-file constructor had to hard-code {@code READ_WRITE}.
+   * <p>
+   * It is a property OF THE FILE and not of the component asking: {@link FileManager} opens every file it scans
+   * with the database's own mode and {@link FileManager#getOrCreateFile} hands a registered file back instead of
+   * reopening it, so this is the mode a second view is going to get whatever it asks for - which is precisely why
+   * that method now refuses to hand back a file open in a different one.
+   */
+  public MODE getMode() {
+    return mode;
+  }
+
   public long calculateChecksum() throws IOException {
     final CRC32 crc = new CRC32();
     final String fileContent = FileUtils.readFileAsString(osFile);
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -24,6 +24,7 @@ import com.arcadedb.database.DatabaseInternal;
 import com.arcadedb.database.Document;
 import com.arcadedb.database.RID;
 import com.arcadedb.database.Record;
+import com.arcadedb.engine.timeseries.TimeSeriesEngine;
 import com.arcadedb.exception.DatabaseOperationException;
 import com.arcadedb.exception.NeedRetryException;
 import com.arcadedb.exception.RecordNotFoundException;
@@ -37,6 +38,7 @@ import com.arcadedb.schema.DocumentType;
 import com.arcadedb.schema.IndexMetadata;
 import com.arcadedb.schema.LocalDocumentType;
 import com.arcadedb.schema.LocalEdgeType;
+import com.arcadedb.schema.LocalTimeSeriesType;
 import com.arcadedb.schema.LocalVertexType;
 import com.arcadedb.schema.Schema;
 import com.arcadedb.serializer.BinarySerializer;
@@ -188,6 +190,14 @@ public class DatabaseChecker {
     result.put("deletedRecordsAfterFix", new LinkedHashSet<>());
     result.put("corruptedRecords", new LinkedHashSet<>());
     result.put("corruptedIndexes", new LinkedHashSet<>());
+    // Issue #6340: the TimeSeries pass, seeded like every other family so a clean run publishes zeros rather than
+    // omitting the keys - "did this run look at the TimeSeries files at all?" has to be answerable from the result
+    // of every run, and a database with no TimeSeries type answers it with zeros rather than with silence.
+    result.put("corruptedTimeSeries", new LinkedHashSet<String>());
+    result.put("totalTimeSeriesTypes", 0L);
+    result.put("totalTimeSeriesShards", 0L);
+    result.put("totalTimeSeriesSamples", 0L);
+    result.put("totalTimeSeriesSealedBlocks", 0L);
     result.put("totalWarnings", 0L);
     result.put("totalCorruptedRecords", 0L);
     result.put("distinctMissingReferences", 0L);
@@ -208,6 +218,11 @@ public class DatabaseChecker {
     final List<DocumentType> edgeTypes = new ArrayList<>();
     final List<DocumentType> vertexTypes = new ArrayList<>();
     final List<DocumentType> documentTypes = new ArrayList<>();
+    // Issue #6340: a TimeSeries type is a LocalDocumentType and used to land in documentTypes, where the per-type
+    // pass found it had no record buckets and did nothing - which is the whole reason CHECK DATABASE had no
+    // TimeSeries coverage. Its data is not in record buckets: it is in the .tstb/.tstd components the schema
+    // registers as files and in the .ts.sealed files outside the paginated layer, so it gets its own pass.
+    final List<LocalTimeSeriesType> timeSeriesTypes = new ArrayList<>();
     for (final DocumentType type : database.getSchema().getTypes()) {
       if (types != null && !types.isEmpty() && (type == null || !types.contains(type.getName())))
         continue;
@@ -215,6 +230,8 @@ public class DatabaseChecker {
         edgeTypes.add(type);
       else if (type instanceof LocalVertexType)
         vertexTypes.add(type);
+      else if (type instanceof LocalTimeSeriesType tsType)
+        timeSeriesTypes.add(tsType);
       else
         documentTypes.add(type);
     }
@@ -268,6 +285,10 @@ public class DatabaseChecker {
       currentStep = 0;
       totalSteps = edgeTypes.size() + vertexTypes.size() + documentTypes.size() // per-type checks
           + 3 // buckets + external properties + indexes
+          // #6340: ONE step for all TimeSeries types, and only when the database has any. A step per type would
+          // change the plan of every database that has none, which is what every existing progress expectation
+          // was written against; a step that is not there costs nobody anything.
+          + (timeSeriesTypes.isEmpty() ? 0 : 1)
           + (reclaimOrphanedSegments ? 1 : 0)
           + (fix ? 1 : 0) // rebuild affected indexes
           + (compress ? 1 : 0);
@@ -288,6 +309,8 @@ public class DatabaseChecker {
       }
 
       checkDocuments(documentTypes);
+
+      checkTimeSeries(timeSeriesTypes);
 
       checkBuckets(result);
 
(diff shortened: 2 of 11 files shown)


lvca:
Follow-up filed from this work: #6356 - corrupting a `.ts.sealed` header magic makes the whole TimeSeries type **disappear from the schema** on reopen, with the database opening cleanly and reporting nothing. It is in the schema load path (`LocalSchema.readConfiguration` never reaches `types.put` when `initEngine()` throws, and the `ConfigurationException` is then swallowed), so this PR's checker cannot reach it - a type that is not in the schema is not walkable. Left as a comment where the test would have gone rather than pinned, since asserting the current behaviour would enshrine it.

Issue #6319, opened 2026-08-17
Ordinary updates leak continuation chunks: 243821 orphans out of 1545495 chunk slots in CRUDTest.multiUpdatesOverlap


lvca:
Found by the orphaned-chunk sweep #6294 added (PR #6299), which is the first thing in the engine that can see this.

## What happens

An ordinary update workload leaks continuation chunks. `CRUDTest.multiUpdatesOverlap` - 131072 records, each
given ten short string fields over thirteen rounds of `updateAll` - ends with:

```
chunk slots in the bucket : 1545495
reachable from any chain  : 1301674
reachable from nothing    :  243821     (16%)
```

Every one of those 243821 slots holds up to a page of content that no record can reach and nothing ever
frees. Nothing in the test is corrupt: no `CHECK DATABASE` error, no broken chain, no force-delete. It is the
plain update path.

## This is not a mis-measurement

The sweep marks a chunk by walking the chain of every head, so "unmarked" could in principle mean "the walk
missed it". It was checked directly rather than assumed: for unmarked chunks, every slot of the bucket was
scanned for one whose next-chunk pointer names it.

- `#1:604567` - **zero referrers of any kind**.
- `#1:1095959` - one referrer, `#1:727045`, which is itself a `NEXT_CHUNK` and itself unmarked.

The second shape is the giveaway: these are orphaned *sub-chains*, `chunk1 -> chunk2 -> chunk3` still pointing
at each other with nothing pointing at `chunk1`. A walk that had merely stopped early could not produce it,
because the walk marks a pointer as it reads it - reaching `#1:727045` would have marked `#1:1095959` too.

## Where to look

`updateMultiPageRecord` rewrites a record's chain in place: chunks that are still needed are reused, chunks
that are not are handed to `freeChunkChain`. #6294's issue text already names the suspects - "an interrupted
`updateMultiPageRecord` shrink" and `GraphEngine.rebuildVertexEdgeChain` - as paths that "leave chunks behind
without deleting anything". The fixture here is a pathological one for chunking (a bucket where pages are
packed, so a spilled record's head chunk gets a handful of bytes and the chain runs to ~12 chunks for a
~700-byte record), which is presumably why it shows up so starkly.

The interesting question is which rewrite drops the reference: a chain that shortens, one that lengthens, one
that collapses, or the re-spill after a placeholder content record is deleted and recreated.

## Why it is worth fixing at the source

[...]

PR #6349, opened 2026-08-18
fix(#6339): the space a page gets back is measured where the page is packed, not guessed where a slot is freed

### LINKED ISSUE #6339: Freeing a chunk gives its bytes back to the page but not to the allocator's free-space statistics
Found while fixing #6319, and its impact grew the moment that fix landed.

## What happens

Freeing a chunk gives its bytes back to the page but not to the allocator. Both places that free one write the
slot-table entry to zero and stop there:

- `LocalBucket.freeChunkChain` - the whole tail an update drops (`LocalBucket.java`, the
  `nextPage.writeUnsignedInt(PAGE_RECORD_TABLE_OFFSET + chunkPositionInPage * INT_SERIALIZED_SIZE, 0L)` loop);
- `LocalBucket.freeSlotOnly` - one slot, used by the orphaned-chunk sweep and the dangling-pointer repair of
  `check(fix)`.

Neither calls `updatePageStatistics`. The page really does have those bytes free again, but the free-space
statistics `findAvailableSpace` consults still describe the page as it was before, so the allocator keeps
passing it over. The space comes back only when `compressPage` re-flows the page - i.e. on
`CHECK DATABASE ... COMPRESS`, an admin operation - even though nothing about the page is corrupt.

## Why now

This is the mirror of #6154, which fed the head chunk into `updatePageStatistics` for exactly this reason: the
statistics had been telling the allocator a page was full when it was not, and #6154 fixed the direction where
the chunk TAKES space. The direction where it GIVES space back was left alone.

It mattered less then, because before #6319 an ordinary update mostly did not free chunks - the exact-fit case
leaked its whole tail instead of freeing it, on every record that shrank onto a chunk boundary. #6319 made
those frees actually happen: on the scaled-down `CRUDTest.multiUpdatesOverlap` measured there, 14719 chunks are
now freed on a workload that previously freed none of them. That is 14719 chunks' worth of space that the
allocator is not told about, on a perfectly healthy bucket.

## Worth considering

The accounting is not symmetric with the write path, and the reason is worth stating before picking a fix: a
chunk's footprint on its page is not the same as its declared content size (the header declares CONTENT
length, and a shrink can leave the footprint larger than what the header says), so "how much does freeing this
chunk return" has to be derived the same way `getPageOccupiedInBytes` derives it, not read off the header.

The cheapest correct thing is probably to account only the case the allocator actually cares about - a freed
chunk that was the page's LAST record, which moves the free tail - and leave a mid-page hole to `compressPage`,
which is the same split `updateMultiPageRecord` already makes for the head chunk (#6154: the tail moves only
when the region runs to `getMaxContentSize()`). That keeps the hot path cheap and stops the systematic
under-reporting on the pages where it compounds.

Not a correctness bug: no record is wrong, no query is affected, and `CHECK DATABASE` reports nothing, because
nothing is broken. It is a space-efficiency defect that now accumulates on ordinary update workloads.

lvca:
Fixes #6339.

## What was wrong

Freeing a slot gave its bytes back to the page and told the allocator's free-space statistics nothing, so
`findAvailableSpace` went on passing the page over. Three places free one and none of them accounted for it:
`freeChunkChain` (the whole tail an update drops), `freeSlotOnly` (the dangling-pointer and orphaned-chunk repairs of
`check(fix)`), and `deleteRecordInternal`.

It mattered little while an ordinary update hardly ever freed a chunk. #6319 changed that: on the workload measured
there, 14719 chunks are now freed where none were before, and every one of them was space the allocator was not told
about on a perfectly healthy bucket.

Reproduced by asserting on the hint itself: after a chunked record shrinks, `getFreeSpaceHintForPage` answers `-1` for
every page its continuation chunks had to themselves, although those pages now hold nothing at all.

## Two things the issue does not have

**The ordinary delete had the same defect, with every sign wrong.** `deleteRecordInternal` measured the page *after*
zeroing the slot - so the tail the delete had just released was already in the number - and then subtracted the
record's footprint from it, landing back on exactly the free space the page had *before* the delete. Mid-page it
subtracted a footprint from a tail that had not moved at all, and could drop the page's entry outright. It also
skipped every negative marker (every shape that is not a plain record), and tested that marker on `recordSize`, which
the chunk-chain walk above it reassigns to the last chunk of the chain.

**`compressPage` is not an admin operation.** The issue says the space "comes back only when `compressPage` re-flows
the page - i.e. on `CHECK DATABASE ... COMPRESS`". In fact `TransactionContext.commit1stPhase` runs it on *every*
bucket page a transaction modified, after the record writes. The page really does get its bytes back at each commit;
it is only the statistics that never heard.

## The fix, and why it is not where the issue suggests

The issue proposes accounting the free at the free, limited to the case where the freed slot was the page's last
record. That was built first, and then dropped in favour of something simpler and exact.

A freed slot is a **hole**, and the allocator only ever hands out a page's free **tail**. A number produced at the
free is therefore a guess about a page whose occupancy is not settled yet - and it is settled one step later, by that
commit-time `compressPage`. That call already packs the live records from `contentHeaderSize` upwards and already
builds the ordered list of their footprints; summing that list is where the free tail begins, exactly, for every way a
page's occupancy can change - a dropped chunk chain, a repaired slot, a delete, an in-place shrink, a mid-page hole -
and it costs nothing that was not already computed.

So the whole change is one call at the end of `compressPageInternal` (plus its empty-page arm), and
`deleteRecordInternal`'s statistics block removed rather than repaired. The three free sites keep only a comment
saying why they no longer guess.

The free-time version was kept long enough to measure against this one: it moved nothing (530 vs 530 pages, 171 vs
171 - see below) while costing a slot-table scan per freed chunk on the update path. Two mechanisms for one fact, the
cheaper one redundant, so it went.

## Measured

Page count of one bucket, before and after:

| workload | before | after |
| --- | --- | --- |
| 40 records of ~200 KB churned between 60 KB and 200 KB for ten rounds | 1130 | **530** |
| 4000 records of 2 KB, two thirds deleted and exactly as many re-inserted | 143 -> 238 | 143 -> **171** |

## What accounting at commit trades

A record deleted and another inserted inside the **same still-open transaction** do not see the freed space as an
allocator hint, because the page is packed at commit and not before. The allocator is never given a wrong answer by
this - `findAvailableSpace` re-reads every candidate page - only a placement it could have made and did not, and the
page is offered from the commit onwards. Measured: the free-time accounting, the only thing that could close that
window, moved 4 pages out of 239 on a single transaction that deletes two thirds of 4000 records and re-inserts as
many. Stated on `accountCompressedPage` so the next reader does not take it for an oversight.

## Tests

`Issue6339FreedSpaceStatisticsTest`, seven tests, all four of the failing ones red before the fix:

- the pages a freed chunk chain releases are offered whole again (the issue's case, asserted on the hint);
- and are handed to new records rather than the bucket growing fresh ones (end to end, with the last page deliberately
  kept occupied so the allocator's statistics-free fallback cannot carry the test);
- the slots the orphaned-chunk sweep of `check(fix)` reclaims are offered again;
- the tail a plain delete releases is offered again, all of it;
- a slot freed mid-page is accounted by the compression that closes the hole;
- delete churn does not grow the bucket by the pages it freed;
- `CHECK DATABASE ... COMPRESS` re-measures the pages it walks. That is the second entry point into the same
  accounting - it reaches `compressPage` with `forceWipeOut`, which no commit ever does, and walks every page of every
  bucket rather than the ones a transaction touched, so it does not merely report but REPAIRS a drifted entry. The
  test plants a wrong entry for a page a spill has sealed and asserts the walk replaces it with what it measured;
  without the plant the assertion would hold whether or not the compression reported anything.

Full `engine` suite green: 12469 tests, 0 failures. The `slow` lane (238 tests, which the suite run above excludes)
is green too, after the change below.

## One existing test asserted the defect

`RandomDeleteTest.checkRecords` compared a full scan against the **insertion order** with `isEqualTo`, so it required a
re-insert to land after the last record written. That held only because the allocator could not see the space the
test's own mass delete had freed: every cycle appended another 100k slots instead of reusing the ones it had just
emptied. With the statistics fixed the space really is reused, the re-inserts land in the reclaimed pages, and a scan -
which walks a bucket in physical order, a RID being a physical position - no longer returns them in write order.

The identity of the records is what the check is about and it is kept: both sides are now sorted by position and
compared, which still catches a record lost, one handed out twice, and one the scan invents, with the size asserted
first. Only the ordering requirement goes.

Established rather than assumed, because "the order changed" and "a record went missing" fail the same assertion: the
comparison was temporarily split in two, and the slow-lane run that reproduced the CI failure failed the ORDERED
assertion with the sorted one passing - same 100k records, same multiplicities, different sequence. The class carries
`@Tag("slow")`, which is why the earlier full-suite runs on this branch did not cover it.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
@@ -3992,6 +3992,14 @@ public class LocalBucket extends PaginatedComponent implements Bucket {
    * maximum content size lands on 0 and has its entry dropped, exactly as a page a spill has sealed does.
    * <p>
    * Costs nothing to derive: the sum comes from the ordered record list the compression already built.
+   * <p>
+   * What it trades, stated so the next reader does not take it for an oversight: a record deleted and another
+   * inserted inside the SAME still-open transaction do not see the freed space as a hint, because the page is packed
+   * at commit and not before. The allocator is never given a wrong answer by this - {@code findAvailableSpace}
+   * re-reads every candidate page - only a placement it could have made and did not, and the freed page is offered
+   * from the commit onwards. It was measured before being accepted: an accounting at the free itself, which is the
+   * only thing that could close that window, moved 4 pages out of 239 on a transaction that deletes two thirds of
+   * 4000 records and re-inserts as many, and cost a slot-table scan per freed slot to do it.
    *
    * @param contentEndInPage first free byte of the packed page, i.e. where its free tail starts.
    *
--- a/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
@@ -5526,10 +5526,16 @@ public class LocalBucket extends PaginatedComponent implements Bucket {
     // Snapshot keys/values into local arrays so we can safely mutate freeSpaceInPages
     // (put/remove) inside the loop body. The snapshot is bounded by MAX_PAGES_GATHER_STATS (100).
     //
-    // The snapshot is sorted by pageId to preserve the deterministic "fill lowest pageId first"
-    // allocation behavior that the previous TreeMap-backed implementation provided implicitly via
-    // its ordered iteration. Tests like RandomDeleteTest depend on this so that re-inserts after
-    // bulk deletes land at the same RIDs as the original inserts.
+    // The snapshot is sorted by pageId to preserve the deterministic "fill lowest pageId first" allocation behaviour
+    // that the previous TreeMap-backed implementation provided implicitly via its ordered iteration. What it buys is
+    // a REPRODUCIBLE choice among the candidate pages - the same bucket state allocates the same way twice - and the
+    // locality of preferring the pages nearest the front of the file.
+    //
+    // What it does NOT buy, and used to be claimed here (#6339): that a re-insert after a bulk delete lands on the
+    // RID the original insert had. It cannot, and it could not then either - which page is a candidate at all depends
+    // on what the free-space statistics hold at that moment, and this ordering only decides between the pages that
+    // made it into that map. RandomDeleteTest was cited as depending on it and no longer does: it compares the scan
+    // against the records it inserted rather than against the order it inserted them in.
     final int snapSize = freeSpaceInPages.size();
     final int[] snapPageIds = new int[snapSize];
     final int[] snapPageStats = new int[snapSize];
--- a/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
+++ b/engine/src/main/java/com/arcadedb/engine/LocalBucket.java
@@ -1846,9 +1846,9 @@ public class LocalBucket extends PaginatedComponent implements Bucket {
    * the ordinary delete would remove. An orphaned continuation chunk (#6294) has no chain left to walk either, and
    * only its own slot to give back.
    * <p>
-   * No page statistics are updated, matching {@code deleteRecordInternal} on the very same markers: both shapes carry
-   * a NEGATIVE size marker, from which the space the slot occupied cannot be derived without reading the record it no
-   * longer describes. The next {@code gatherPageStatistics} measures the page as it now is.
+   * No page statistics are updated here, and since #6339 that is a decision rather than an omission: what the page
+   * has to offer once this slot is gone is settled by the compression the commit runs on it, which reports it
+   * ({@link #compressPageInternal}). Until then the slot is a hole the allocator could not use anyway.
    *
    * @author Luca Garulli (l.garulli@arcadedata.com)
    */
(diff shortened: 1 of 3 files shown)


lvca:
On the nit about gating the `Σ record[1]` loop behind `reuseSpaceMode >= HIGH`: not taking it, and the reason is
worth recording rather than left as a judgement call.

`updatePageStatistics` does **not** early-return under `LOW`/`MEDIUM` before doing anything - its first statement is
`changesFromLastStats.incrementAndGet()`, which runs unconditionally, and only then does the mode gate return. That
counter is load-bearing under `MEDIUM`: `findAvailableSpace` calls `gatherPageStatistics()` whenever
`freeSpaceInPages` is empty (`LocalBucket.java:5468-5471`), and `gatherPageStatistics` does nothing at all unless
`changesFromLastStats > 0`. So gating the call itself behind `HIGH` would stop a compressed page from ever signalling
a change under `MEDIUM`, and the map would stay empty - a real regression, and precisely the increment the removed
`deleteRecordInternal` block used to contribute in its `else` arm.

The version that survives that objection - gate only the summation, and increment the counter directly in the `else` -
hoists a private invariant of `updatePageStatistics` into a second place that would drift the next time that gate
changes. It buys less than one comparison of the `Comparator.comparingLong` sort `getOrderedRecordsInPage` already
performs unconditionally on the same list, one line above. Duplicating a gate to save a rounding error on the method's
own cost is the wrong trade, so the loop stays where it is.

lvca:
On the knock-on flagged for `changesFromLastStats` - traced it, and it is a no-op rather than a behaviour change worth
guarding.

The counter has exactly one reader in the whole main tree: `gatherPageStatistics`, which does
`getAndSet(0L)` and then `if (consumedChanges > 0 || firstRun)`. Only the zero / non-zero distinction is ever used, so
a higher increment rate cannot change what it decides. And the cadence is not the counter's at all - the same method
returns early on `System.currentTimeMillis() - timeOfLastStats <= MAX_TIMEOUT_GATHER_STATS` (5 s), which is checked
BEFORE the counter is consumed. Nothing else reads it: no metric, no monitoring, no other call site
(`LocalBucket.java:221, 2958, 5626, 5660, 5669` are the only occurrences).

So the extra increments cost one atomic add per modified bucket page per commit and are invisible to everything
downstream. Nothing to change.

lvca:
CI status: 26 checks green, including `slow-unit-tests` - the lane that caught the `RandomDeleteTest` problem earlier
in this PR and is green now.

The one red lane, `ha-integration-tests`, is the pre-existing flake and not this change. Same test, same job, same
failure mode on **main** with none of this PR on it:

- here: `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` - `AssertionError: Update returned null:
  record=9 seq=29`
- main, run 31836637018 (2026-08-14): `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` -
  `AssertionError: Update returned null: record=8 seq=5`

The assertion is an HTTP command returning null under eight threads hammering one bucket through a Raft cluster, which
is not a failure mode a free-space HINT can produce: everything this PR writes goes to `freeSpaceInPages`, which only
`findAvailableSpace` reads and which it re-verifies against the page (`getAvailableSpaceInPage`) before using a
candidate. No record bytes, slot tables or replication paths are touched.

Issue #6341, opened 2026-08-17
The dictionary truncation guard did not fire under a full-suite run, and it is a data-integrity guard


lvca:
Observed on PR #6327, whose own changes touch nothing this test reaches.

## What happened

`DictionaryMultiPageTest.aDictionaryPageThatIsClaimedButMissingFailsLoudly` failed in the `unit-tests` lane
with:

```
java.lang.AssertionError:
Expecting code to raise a throwable.
	at com.arcadedb.engine.DictionaryMultiPageTest.aDictionaryPageThatIsClaimedButMissingFailsLoudly(DictionaryMultiPageTest.java:472)
```

It passed standalone, passed in a full local `engine` run of the same 12407 tests, and passed on a re-run of
the very same CI lane on the very same commit. The lane flakes on `main` too, with a different single test
(`Issue5596MergeCoverageTest.aCleanAppendConflictIsStillMerged` on `0efa9990b`), so "one test, once" is a
known shape here - but this particular one is worth more than a re-run.

## Why this one is worth understanding rather than re-running

The assertion that did not fire is a data-integrity guard, and the class says so itself: a page the count
claims but that is not on disk must stop the load, because an empty page contributes zero names and every name
after it comes back with an id lower by however many the missing page held - "silently renumbering is the one
outcome this class exists to prevent" (`Dictionary.java`, around the `is truncated` throw).

The test claims three pages when one exists and expects `reload()` to throw. `reload()` detects the gap
indirectly: it asks `PageManager.getImmutablePage(pageId, pageSize, false, pageNumber == 0)` and turns the
`IllegalArgumentException` into the `DatabaseMetadataException`. So the guard fires only when the page manager
refuses the read - and the failure says that under full-suite conditions it sometimes does NOT refuse.

Which means the interesting question is not "why is the test flaky" but "under what conditions does the
page manager answer for a page past the end of the file". If the answer is "when some other database's page
of the same number is resident, or when the file's on-disk size is not yet what the component believes",
then the guard is weaker than its own comment claims, and the flake is the only evidence of it anyone has.

## Where to start

- The test builds its own database and drops it (`getDatabasePath() + "_claimedButMissing"`), so the state
  it depends on comes from the shared JVM fork, not from its own fixture.
- `updatePageCount(3)` moves `PaginatedComponent.pageCount` without writing anything; whether the read then
  fails depends on `PaginatedComponentFile`'s size, which lags a page still sitting in the flush queue -
[...]

PR #6355, opened 2026-08-18
fix(#6351): the dictionary asks the committed page count whether there is anything to read, not the file's size

### LINKED ISSUE #6351: Dictionary decides "is there anything to read" from the file's size, which lags the flush queue - and every guard built on a page read inherits the read cache's answer
Two findings from the work on #6341 / #6342 (PR #6346, merged as `2304fba5d`). They are filed together because
they are the same shape seen at two levels - a decision made from evidence that is not the fact being asked about -
but they are independent and item 1 is far more actionable than item 2.

Neither is a regression from #6346. Both predate it; #6346 only made them visible.

---

## 1. `reload()`'s empty-file guard clause contradicts the rule the rest of the method documents

`Dictionary.reload()` opens with (`Dictionary.java:470`):

```java
public void reload() throws IOException {
  if (file.getSize() == 0) {
    // No header page on disk. ... The in-RAM dictionary is already empty, which is the correct
    // state for an empty file.
    return;
  } else {
```

and the `else` branch immediately below spends a paragraph explaining why the file is the WRONG source of truth
for exactly this question:

> COMMITTED STATE ON BOTH COUNTS, WHICH IS THE WHOLE CONTRACT OF THIS METHOD ... **READ THROUGH THE PAGE MANAGER
> INSTEAD, AND COUNT WITH `pageCount`, WHICH ONLY ADVANCES ON COMMIT.**

The guard clause then does the one thing that paragraph forbids. And the two disagree in a window that exists by
construction: `pageCount` advances at COMMIT, the file grows when the flush thread gets to the page, so between a
dictionary append committing and its page reaching disk, `pageCount` is 1 and `file.getSize()` is 0.

**What the early return costs in that window** is not "nothing is loaded" - it is that `this.entries` is left
exactly as it was. `reload()` is the repair path, so the callers that matter are the ones calling it BECAUSE
something needs undoing:

- `TransactionContext.rollback()` (`:437`), when the rolled-back transaction touched a dictionary page;
- `TransactionManager.applyChanges()` (`:730`), after a replicated change that involved the dictionary;
- `updateName()`'s own `catch (final IOException e)` (`Dictionary.java:349`), which is the concrete one:
  `updateName` publishes the renamed `entries` BEFORE rewriting the pages (`:319`), and repairs a failed rewrite
  by calling `reload()`. If that repair lands in the window, the renamed in-RAM view survives while no page was
  ever rewritten.

It also means the truncation guard #6341 just strengthened is skipped entirely in that window, since the whole
loop is behind this branch.

**Honesty about the evidence:** this is a correctness argument, not a reproduction. I probed it and could not
open the window on the path I could drive - database creation flushes page 0 synchronously, so `file.getSize()`
is already `pageSize` by the time a test can call `reload()`. Whoever picks this up should either find a path
that does open it or record that they could not.

**The fix is one line and is what the method already argues for:** decide from `pageCount.get() == 0`, not from
`file.getSize() == 0`. On the load path the two agree by construction (`PaginatedComponent`'s constructor sets
`pageCount = fileSize / pageSize`), so the "database killed before the header page reached disk" case the comment
is about behaves identically; only the lag window changes, and it changes to the correct answer.

**Two sibling sites decide from the same source** and should be looked at in the same pass, though neither is
obviously wrong today:
- the creation constructor, `Dictionary.java:130`;
- `createHeaderPageIfMissing()`, `Dictionary.java:569-570` (`if (file.getSize() > 0) return;`) - this one runs at
  open before anything has been committed, so its window is likely closed, but it is the same question asked of
  the same wrong source.

---

## 2. A page read cannot distinguish "the file has this page" from "some image of it is resident"

`PageManager.getCachedPage` (`PageManager.java:1758`) consults the JVM-wide `readCache` FIRST, and only on a miss
does `loadPage` look at `file.getTotalPages()`. So a resident image answers for a page the file may not have, and
no caller can tell the two apart.

Every guard built on "did the page read succeed" inherits that, which is what #6341 was really about:
`reload()` used to infer "this page is missing" from `getImmutablePage` refusing the read. #6346 fixed the case
it could fix - a page that IS inside the file but was never written now fails on its own bytes rather than on the
read's outcome - but the general shape is untouched.

**This is a note, not a bug report, and I would not act on it as written.** Making the page manager prove
residency corresponds to the file is a large change on the hottest path in the engine, for a hazard nobody has
demonstrated. It is filed so the next person who finds a guard behaving oddly has the mechanism written down.

### The concrete loose end behind it

`DictionaryMultiPageTest#aDictionaryPageThatIsClaimedButMissingFailsLoudly` failed once in the `unit-tests` lane
(reported in #6341) with `Expecting code to raise a throwable` - nothing raised at all. #6341 is closed as
completed, but **what closed it is a different defect from the one that produced that failure**, and the trigger
is still unexplained:

- the failure requires pages 1 AND 2 of a brand-new database at a unique path to have come back from somewhere;
- the only two routes are the read cache and the flush index, and both are keyed exactly by
  `PageId(databasePath, fileId, pageNumber)`;
- the two hypotheses in #6341 are both **dead**, checked directly rather than assumed:
  - *"the file's size lags a page in the flush queue"* is backwards - `isNewPage = pageNumber >= file.getTotalPages()`,
    so a smaller `getTotalPages()` refuses MORE pages, making the guard fire more readily, not less;
  - *"a stale read-cache entry survives a drop and re-create at the same path"* - `drop()` and `close()` both call
    `PageManager.removeAllReadPagesOfDatabase`; verified by probe.

The test now asserts its file-size premise explicitly, so a recurrence will name which premise stopped being true
rather than only reporting that nothing was thrown. **If it never recurs, this needs no action** - the note above
is the whole of what is known.

lvca:
Closes #6351.

## 1. `reload()`'s guard now reads what the rest of the method argues for

`Dictionary.reload()` opened with `if (file.getSize() == 0) return;`, while the branch immediately below spends a
paragraph explaining that the file is the wrong source for exactly that question and counts with `pageCount`, "which
only advances on commit".

The two disagree in a window that exists by construction: a commit advances `pageCount` and hands the page to the
flush thread, so until that thread gets to it the file is shorter than the count claims - and while the dictionary's
FIRST page is in flight, the file is still empty with one page already committed.

What the early return costs there is not "nothing is loaded". `reload()` is the repair path, so leaving `entries`
untouched leaves the thing being repaired in place:

- `TransactionContext.rollback()` - a rolled-back dictionary change survives in RAM;
- `TransactionManager.applyChanges()` - a replicated change stays half applied;
- `updateName()`'s own `catch (IOException)` - the renamed view is published BEFORE the pages are rewritten, so the
  rename survives with no page ever rewritten and the dictionary answers with a name that is in no page of the
  database.

It also skipped the truncation guard #6341 had just strengthened, since the whole loop sits behind that branch.

**The fix is to decide from `pageCount.get() == 0`.** Every page the count claims is still reachable in the window:
`PageManager.loadPage` looks in the flush queue BEFORE it looks at the file, and by the time a page leaves the queue
the file covers it. On the load path the two sources agree by construction (`PaginatedComponent`'s constructor derives
`pageCount` from the file size), so the case the guard was written for - a database killed before the header page
reached disk - is unchanged.

### The two sibling sites move with it

The creation constructor and `createHeaderPageIfMissing()` asked the same question of the same source. On a file
holding whole pages they give the answer the file size gave. They part on a file holding LESS than one whole page -
killed mid-write - where there is no complete page 0 to keep, the count says so, and writing the header page over
those bytes is the repair. `reload()` already treats such a file as an empty page 0 (its floor of 1 plus
`createIfNotExists`), so this only makes the on-disk file agree with the in-RAM view instead of leaving it to whatever
appends next.

### Was the window reproducible?

The issue asked whoever picked this up to either find a path that opens it or record that they could not. I confirmed
the probe in the issue: database creation has the dictionary's page 0 on disk by the time a test gets control
(20/20 runs, `getSize() == 65536` immediately after `createDatabase`), so the window is real but narrow and I could
not open it by timing alone either.

The tests therefore construct the STATE rather than race for it - the file emptied behind its committed page count,
which is what the guard actually reads and also what a real file looks like after a kill mid-write. Three of the four
fail on `main`:

| test | on `main` |
|---|---|
| `aRolledBackRenameIsUndoneWhileTheFileIsBehindTheCommittedPages` | `expected "committedName" but was "renamedName"` |
| `aClaimedButMissingPageStillFailsLoudlyWhileTheFileIsBehindTheCommittedPages` | `Expecting code to raise a throwable` |
| `aDictionaryFileShorterThanOnePageIsRepairedAtOpen` | `expected 1 but was 0` (header page never written) |
| `anEmptyDictionaryWithNothingCommittedReloadsToAnEmptyView` | passes - it is the regression guard on the case the old check was for |

## 2. The note, recorded rather than acted on

A read answering is not the same fact as the file holding the page: `getCachedPage` consults the read cache first and
`loadPage` the flush queue second, so a resident image answers for a page whose bytes may not be on disk - which is
the point of both - and only a miss in both reaches `file.getTotalPages()`. Making residency prove it corresponds to
the file is a large change on the hottest path in the engine for a hazard nobody has demonstrated, so per the issue I
have not attempted it. The mechanism is now written at `PageManager.getCachedPage`, where someone building a guard
will look, with the rule it implies: read the BYTES and say what makes them legal content, the way `reload()` does
with the per-page header.

### The loose end behind it, and one new piece of evidence

`DictionaryMultiPageTest#aDictionaryPageThatIsClaimedButMissingFailsLoudly` failed once in the `unit-tests` lane with
`Expecting code to raise a throwable` - nothing raised at all. **Item 1 produces exactly that symptom**: at the time
of that failure the test had no file-size premise, and the old guard returned silently whenever `file.getSize()` was
0, which is precisely what the flush lag leaves behind a committed page 0. The new
`aClaimedButMissingPageStillFailsLoudlyWhileTheFileIsBehindTheCommittedPages` reproduces that failure message
verbatim on `main`. That is a candidate explanation, not a demonstration - I could not open the window by timing - but
it is one both dead hypotheses in #6341 could not offer, and it is now fixed either way.

One consequence: with the guard reading the count, the file-size premise #6346 added to that test is no longer a
premise of the code under test, and asserting it unwaited would itself be racing the flush thread. It is now drained
first, so it stays a tripwire without becoming a flake.

## Alternative considered and rejected

Instead of moving the guard, `reload()` could have been made to ask the page manager to PROVE the pages it claims are
in the file. That is item 2's large change, on the hottest path, and it also answers a question `reload()` does not
need answered: it does not care whether a committed page has reached disk yet, only that it is committed. The count
is that fact directly.

## Verification

- `mvn -o -pl engine -am test -DexcludedGroups=benchmark,slow,vector` - **12512 tests, 0 failures, 0 errors**
- full-project `mvn compile` clean

## Files changed

- `engine/src/main/java/com/arcadedb/engine/Dictionary.java`
- `engine/src/main/java/com/arcadedb/engine/PageManager.java` (comment only)
- `engine/src/test/java/com/arcadedb/engine/DictionaryFileSizeLagTest.java` (new)
- `engine/src/test/java/com/arcadedb/engine/DictionaryMultiPageTest.java`


lvca:
## CI record

Green after reruns except `ha-integration-tests`, which is the tracked chronic flake. Recording the evidence so it does not have to be re-derived.

**`vector-unit-tests`** - `LSMVectorIndexRecoveryTest.vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild`, an 11-minute Awaitility timeout waiting for a background rebuild to drain its delta buffer. The `REBUILD_SEMAPHORE` convoy this lane exists to isolate. **Passed on rerun.**

**`unit-tests`** - `SQLMethodPrecisionTest.instant`, `expected: Millis but was: Seconds`. The test documents its own flake: `// NANOS COULD END WITH 000 AND THEREFORE HE TEST WON'T PASS, RETRY MULTIPLE TIME IN CASE`, and all 10 retries came back with a zero sub-second remainder. **Passed on rerun.**

**`ha-integration-tests`** - failed twice, on **disjoint** test sets for the same commit, which is the #5668/#5702 signature rather than a defect:

| run | failing tests | duration |
|---|---|---|
| 1 | `Raft3PhaseCommitIT`, `RaftHTTPGraphConcurrentIT`, `RaftReplicationIT` | 1h01m33s |
| 2 | `Issue5569SlotMergeDeleteRaftIT`, `RaftPriorityRejoinIT` | 1h05m43s |

All are "the response came back null" / "follower had not caught up" shapes, both runs at the 60-minute cap, and `Issue5569SlotMergeDeleteRaftIT` is one of the two PR #5704 explicitly declined to touch as a known flake. `main` itself has `ha-integration-tests` red in 2 of its last 5 runs within the same hour (runs `32099192174`, `32098995621`), with no involvement from this branch.

**What is positive evidence rather than absence of evidence:** the two dictionary-specific HA tests, `RaftDictionaryManyNewKeysIT` and `Issue5560DictionaryRolloverReplicationIT`, passed in **both** HA runs. Those are the ones that would notice if `reload()` on the replication-apply path had changed behaviour.

## `codecov/patch`

`codecov/project` passes and overall coverage rose 70.69% -> 70.77%. `codecov/patch` reports `Dictionary.java` at 3 lines / 2 hits / 0 misses / **1 partial** = 66.7%, under the auto target.

The partial is a branch partial on the creation constructor's `if (pageCount.get() == 0)`, whose false branch is structurally unreachable: that constructor only ever runs on a file that was just created, so the count is always 0 - the same observation the review made ("a no-op in practice"). The old `file.getSize() == 0` was equally always-true there and simply was not in a diff. Covering it would mean driving the creation constructor onto a pre-existing file, which the component's own `getOrCreateFile` identity guard (#6283) exists to refuse. Left alone rather than tested for the metric's sake; `Codacy Diff Coverage` and `Codacy Coverage Variation` both pass.

Issue #6338, opened 2026-08-17
openCypher: a label disjunction (y:A|B) on a node bound by expansion matches nothing, while the same disjunction on the anchor works


lvca:
Found while writing the regression tests for #6322 item 2 — it blocked the first construction of the disjointness test, because the shape needed to make a wrong "hops are disjoint" verdict observable is exactly the shape that matches nothing. Left out of PR #6325 as unrelated to that issue. Reproduces on `3a4f9bc6b`.

## A label disjunction on a node bound by expansion matches nothing

`(y:A|B)` written on a node that a relationship expands **into** rejects every row. The same node written bare, or with a single label, matches.

```cypher
CREATE (:Post {k:'p1'});
CREATE (:Topic {k:'t1'});
MATCH (p:Post {k:'p1'}), (t:Topic {k:'t1'}) CREATE (p)-[:TAGGED]->(t);
MATCH (t:Topic {k:'t1'}) CREATE (t)-[:OTHER]->(t);
```

| query | rows | expected |
|---|---|---|
| `MATCH (p:Post)-[:TAGGED]->(y)-[:OTHER]->(z:Topic) RETURN y.k` | 1 (`t1`) | 1 |
| `MATCH (p:Post)-[:TAGGED]->(y:Topic)-[:OTHER]->(z:Topic) RETURN y.k` | 1 (`t1`) | 1 |
| `MATCH (p:Post)-[:TAGGED]->(y:Author\|Topic)-[:OTHER]->(z:Topic) RETURN y.k` | **0** | 1 |
| `MATCH (y:Author\|Topic) RETURN y.k` | 2 (`a1`, `t1`) | 2 |

`t1` is a `Topic`, so it satisfies `Author|Topic`. The last row shows the disjunction is understood perfectly well when the node is the **anchor** — `MatchNodeStep` has an explicit disjunction branch (`if (labelDisjunction) ... any`). It is the bound/expanded-into position that goes wrong, which points at the target-side label check ANDing the alternatives where it should OR them.

Neo4j returns the row.

**Impact.** Silent under-count — an empty result, with no error and no warning, for a pattern that is ordinary Cypher. Worse than an over-count in one respect: an empty answer reads as "no such data" rather than as a bug, so it is unlikely to be reported.

[...]

PR #6357, opened 2026-08-18
fix(#6352): a disjunction picks what to scan, not just its first alternative

### LINKED ISSUE #6352: openCypher: an unbound pattern-comprehension start node scans only the first label, so a disjunction under-counts and a missing first alternative empties the result
Found while fixing #6338 in PR #6348, and deliberately left out of it: it is the same failure mode in a code path none of that PR's five issues reach. Reproduces on `adc7e1565`.

`PatternComprehensionExpression.traverseUncorrelatedStart` picks the scan root for an **unbound** comprehension start node by taking the first label and nothing else:

```java
final List<String> startLabels = startNodePattern.getLabels();
if (startLabels != null && !startLabels.isEmpty()) {
  // Use the first label as the iteration root; remaining labels are checked per vertex.
  final String typeName = startLabels.get(0);
  if (!context.getDatabase().getSchema().existsType(typeName))
    return;
  candidates = () -> context.getDatabase().iterateType(typeName, true);
}
```

The per-candidate filter below it is correct - since PR #6348 it goes through `Labels.matches`, which honours the disjunction - but a vertex that carries only the *second* alternative is never handed to it, because it was never scanned.

## Repro

```cypher
CREATE (:Author {k:'a1'});
CREATE (:Topic {k:'t1'});
CREATE (:Sink {k:'s1'});
MATCH (a:Author {k:'a1'}), (s:Sink) CREATE (a)-[:E]->(s);
MATCH (t:Topic {k:'t1'}), (s:Sink) CREATE (t)-[:E]->(s);
```

| query | rows | expected |
|---|---|---|
| `RETURN [(y:Author\|Topic)-->() \| y.k]` | `["a1"]` | `["a1","t1"]` |
| `RETURN [(y:NoSuchLabel\|Topic)-->() \| y.k]` | `[]` | `["t1"]` |
| `MATCH (y:Author\|Topic)-->() RETURN collect(y.k)` | `["a1","t1"]` | `["a1","t1"]` |

The third row is the same question asked as a MATCH, and it answers correctly - so the comprehension and the pattern disagree about what a disjunction means, which is exactly what #6338 was about, one code path over.

**Two symptoms, not one.** The first is an under-count: only the first alternative is scanned. The second is the shape #6338 fixed in `MatchRelationshipStep.targetLabelsExistInSchema` - an alternative naming a type nobody created makes the whole comprehension answer nothing, when it should simply be an alternative that matches nothing. Here it is the `existsType(startLabels.get(0))` early return.

**Impact.** Silent under-count, no error and no warning, and an empty or short list reads as "no such data" rather than as a bug. Ordinary Cypher: Neo4j returns both rows.

## Where to look

`engine/src/main/java/com/arcadedb/query/opencypher/ast/PatternComprehensionExpression.java`, `traverseUncorrelatedStart`.

## Suggested shape

`MatchNodeStep.getVertexIterator` already solves this: for multiple labels it walks the schema's vertex types, keeps those satisfying the pattern, and chains their iterators. The two helpers PR #6348 added are the pieces needed here - `Labels.matches(DocumentType, labels, disjunction)` to decide which types to scan, and `Labels.canMatchInSchema(schema, labels, disjunction)` in place of the single `existsType` early return. Worth extracting the candidate-type selection so the comprehension and `MatchNodeStep` share it outright rather than growing a third copy, which is how the five copies #6338 had to reconcile came about in the first place.

A regression test should cover both symptoms above, plus the conjunction case (`(y:A:B)`), which must keep requiring every label.

lvca:
Fixes #6352.

## The bug

`PatternComprehensionExpression.traverseUncorrelatedStart` picked the scan root for an **unbound** comprehension start node by taking `getLabels().get(0)`:

| query | before | now |
|---|---|---|
| `RETURN [(y:Author\|Topic)-->() \| y.k]` | `["a1"]` | `["a1","t1"]` |
| `RETURN [(y:NoSuchLabel\|Topic)-->() \| y.k]` | `[]` | `["t1"]` |
| `MATCH (y:Author\|Topic)-->() RETURN collect(y.k)` | `["a1","t1"]` | `["a1","t1"]` |

The per-candidate filter has gone through `Labels.matches` since #6348 and was already correct; a vertex carrying only a later alternative was never handed to it, because it was never scanned. The `existsType(labels.get(0))` early return is the second symptom: an alternative naming a type nobody created emptied the whole comprehension instead of matching nothing, the same shape #6338 fixed one code path over. Silent under-count, no error and no warning.

## The fix

Deciding a row and deciding what to scan have to say the same thing about a disjunction, so the scan side now sits next to the row side in `Labels`:

- `Labels.iterateMatchingVertices(database, labels, disjunction)` walks every vertex type the label constraint accepts, lazily (each type's iterator opens only when the previous one drains) and visiting no vertex twice. A single label keeps the one polymorphic scan it had, so nothing on that path pays for a schema walk.
- The comprehension, the MATCH anchor in `MatchNodeStep`, the `NodeByLabelDisjunctionScan` operator, `CSRCountUtils.iterateAnchors` and `MergeStep.findAllNodes` all route through it, replacing five spellings of one rule with one - which is how the copies #6338 had to reconcile came about in the first place. Net -78 lines.

Two things fall out of the shared path:

- the comprehension no longer re-checks labels per candidate (the scan is label-exact), and its unlabelled walk no longer materialises **every vertex in the database** into an `ArrayList` before iterating;
- MERGE gains the type-inheritance meaning of a conjunction. It compared label lists per record, so `MERGE (n:A:B)` missed a node whose type *extends* the `A~B` composite - `MATCH (n:A:B)` finds that node, and MERGE created a duplicate beside it. Covered by a test that fails on `main`.

## Tests

New `CypherComprehensionStartLabelScanIssue6352Test` (10 tests): both symptoms, both label orders, a disjunction where no alternative exists, the conjunction case (`(y:A:B)`, still requires every label, still empty when one alternative is missing), single-label and unlabelled starts, inline properties and inline `WHERE` on a disjunction start, a correlated comprehension, and the type-extending-a-composite case above. Every disjunction assertion is paired with the MATCH spelling of the same question, which is what Neo4j answers.

4 of the 10 fail on `main`.

Verification: `engine` unit lane green (12517 tests, 0 failures), openCypher TCK suite green (3897, 0 failures).

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/Labels.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/Labels.java
@@ -18,13 +18,19 @@
  */
 package com.arcadedb.query.opencypher;
 
+import com.arcadedb.database.Database;
+import com.arcadedb.database.Record;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.schema.DocumentType;
 import com.arcadedb.schema.Schema;
+import com.arcadedb.schema.VertexType;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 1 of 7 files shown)


lvca:
Review points addressed:

- **`CSRCountUtils.iterateAnchors` behaviour tightening** - correct catch, and it deserved to be written down where the method lives rather than only in the PR body. Its javadoc now says that a label naming an edge or document type answers with no anchors, and the `@return` no longer says "declared on nothing" when it means "no *vertex* type carries it". (688bdc46)
- **Hardcoded `disjunction=true` in `NodeByLabelDisjunctionScan`** - stated at the call site why it needs no guard: the planner only picks this operator for `(n:A|B)`, and a one-element list scans the same records either way, since "any of one" and "all of one" are the same question. (688bdc46)

Left alone deliberately:

- **`estimatedCardinality` under-estimates a disjunction** - pre-existing, already flagged in that operator's javadoc as tracked separately, and fixing it means touching the anchor selector's cost model, which is a different change from this one.
- **`matchingVertexTypes` walks `schema.getTypes()` per call** - it replaces a walk that was already there (`existsType` + the old per-call type enumeration in four of the five sites) and, in the comprehension, a full materialisation of every vertex in the database. A schema-metadata walk is not what would show up in a profile next to the record scan it precedes; if it ever does, the fix is a cached type list invalidated on schema change, not a different rule.
- **No green run in the review sandbox** - confirmed locally before pushing: engine unit lane 12517 tests / 0 failures, openCypher TCK 3897 / 0 failures. CI has the same lanes.

lvca:
Round 2 addressed in 13d10262:

- **Why passing `isLabelDisjunction()` through `MergeStep` is safe** - named at the call site: `CypherSemanticValidator.checkNoLabelDisjunction` refuses `(n:A|B)` in MERGE at parse time, so the flag can only arrive false. Agreed that a reader in `MergeStep.java` shouldn't have to go find the parser to learn that.
- **`NodeByLabelDisjunctionScan` had no test of its own** - added one: it EXPLAINs a disjunction MATCH, asserts the plan actually uses `NodeByLabelDisjunctionScan(y:Author|Topic)` (so the coverage can't quietly become the legacy path), and then checks all three shapes through it - both alternatives present, one alternative missing from the schema, and neither present.

On the MERGE cost note, one correction worth making before it lands in the description as written: the multi-label path is not a cost increase on the record side, it is a decrease. It used to polymorphically scan **every record of the first label** and reject most of them with `containsAll` per record; it now scans only the types that satisfy every label, which for `MERGE (n:A:B)` is a subset of the `:A` records. What is new is a walk of `schema.getTypes()` - metadata, no page reads - in front of that. A MERGE-heavy workload against a many-typed schema pays one map iteration per `findAllNodes` and reads fewer records, so I would rather not describe it as a regression.

On the two schema lookups: `existsType` + `getType` are `containsKey` + `get` on the same `HashMap`. Collapsing them needs a non-throwing accessor on the `Schema` interface, which is more public API than a second map probe is worth; leaving as is.

PR #6369, opened 2026-08-18
fix(#6310): relationship uniqueness is scoped to the MATCH clause, not to one pattern part

### LINKED ISSUE #6310: OPTIONAL MATCH cardinality changes after collect/UNWIND materialization
## Environment
 
- Docker image: `arcadedata/arcadedb:26.8.1`
 

## Summary

Two Cypher queries with the same `OPTIONAL MATCH` pattern and the same row-preserving `UNWIND [1]` produce different row counts. The only material difference is that the second query passes the matched variables through a map, `collect`, and `UNWIND` before the scalar `UNWIND`.

The two forms should preserve the same cardinality. On ArcadeDB `26.8.1`, they do not.

## Reproduction steps

1. Start with a fresh ArcadeDB database.
2. Execute the following Cypher statements individually, in order.
3. Compare the result of the `LEFT` query with the result of the `RIGHT` query.

Each semicolon-separated statement should be submitted as an individual Bolt statement.
// LEFT run
```cypher

MATCH (n) DETACH DELETE n;

CREATE (a1:A)<-[:R]-(e1:E)<-[:R]-(c1:C);
CREATE (a2:A)<-[:R]-(e2:E)<-[:R]-(c2:C);

CALL {
  OPTIONAL MATCH (n0:A), (n3:C)-[]->(:E)-[]->(n0)
  WITH *
  UNWIND [1] AS k
  RETURN count(*) AS rows
}
RETURN rows;
```
// RIGHT run: reset and recreate the same graph first.
```
MATCH (n) DETACH DELETE n;

CREATE (a1:A)<-[:R]-(e1:E)<-[:R]-(c1:C);
CREATE (a2:A)<-[:R]-(e2:E)<-[:R]-(c2:C);

CALL {
  OPTIONAL MATCH (n0:A), (n3:C)-[]->(:E)-[]->(n0)
  WITH {n0: n0, n3: n3} AS x
  WITH collect(x) AS xs
  UNWIND xs AS x
  WITH x.n0 AS n0, x.n3 AS n3
  UNWIND [1] AS k
  RETURN count(*) AS rows
}
RETURN rows;
``` 

## Expected result

Both queries return the same cardinality:

```text
rows = 2
```

## Actual result

ArcadeDB `26.8.1` returns different cardinalities:

```text
LEFT  => rows = 4
RIGHT => rows = 2
```

The result is deterministic across repeated runs.

lvca:
Closes #6310

## The reported pair

The two queries in #6310 - one `OPTIONAL MATCH` read through `WITH *` and the same one read through a `map -> collect -> UNWIND` round trip, answering 4 and 2 on 26.8.1 - are the Cartesian-product defect of #6311, fixed by d02b0c617 (PR #6326) after 26.8.1 was cut. Verified: the reported pair reproduces exactly at `d02b0c617^` and agrees on `main`.

Both spellings now answer the same thing - 2 on the reporter's two-chain graph, 4 on the richer fixture this test uses. Their queries are pinned as a regression test regardless, because that spelling routes the MATCH through the traditional plan, which #6311's tests do not reach.

## The defect this fixes

Generalising the reporter's invariant - *the same MATCH must not answer differently because of the clauses around it* - turned up a second, independent breach of it.

Cypher scopes relationship uniqueness to the whole MATCH clause: two comma-separated pattern parts are one pattern and may never bind the same relationship. That is precisely what distinguishes the comma from a second MATCH clause. The check is a comparison against the edges the row already carries, so a hop that is never asked to bind its edge is invisible to it - and `computeHopEdgeTrackingNeeds` decided that **per pattern part**. A part that could not collide with itself (a single hop, typically) bound nothing, and the hop of a later part had nothing to compare against.

The cost-based optimizer scoped it correctly, so the very same MATCH answered 0 rows on its own and 4 whenever the optimizer declined:

```cypher
// fixture: three C -> E -> A chains, plus one extra C feeding the first E
MATCH (n0:A)<-[r1]-(:E), (n3:C)-[]->(:E)-[r2]->(n0) RETURN n0, n3
```

| form | before | after | Neo4j |
|---|---|---|---|
| plain (optimizer) | 0 | 0 | 0 |
| inside `CALL { ... }` | **4** | 0 | 0 |
| `OPTIONAL MATCH` | **4** | 1 all-null row | 1 all-null row |
| inside `EXISTS { ... }` | **3** | 0 | 0 |
| one part carrying a named path | **4** | 0 | 0 |

`r1` is the only `(:E)-[]->(n0)` edge for a given `n0`, so `r2` can never be a different relationship and the clause matches nothing at all.

## The change

The analysis now runs over every hop of the clause rather than of one part. Both proofs that let a hop stay on the GAV/CSR fast path are kept, and one is sharpened:

- **edge types that cannot both match one edge.** An edge carries exactly one type, so a hop's type list is a disjunction and the pair is disjoint only when every combination is. This was decided by name equality, which read `[:BASE]` and `[:SUB]` as disjoint although a `SUB` edge matches both patterns; it is now the schema's own disjointness test, the one already used for vertex labels.
- **endpoint patterns that are type-disjoint** - unchanged, only generalised so the two hops may straddle a comma.

So `MATCH (n0:A)-[:S]->(:D), (n3:C)-[]->(:E)-[:R]->(n0)` and `MATCH (:E)-[r1:R]->(n0:A), (n3:C)-[r2:R]->(:E)` still take the fast path and still match; a clause with a single hop is untouched.

Also removes an unterminated Javadoc left behind when `isEdgeVariableReferenced` moved to `CypherVariableUsage`.

## Verification

New `CypherMatchClauseUniquenessIssue6310Test` (12 tests): the reported pair, the uniqueness rule across each of the shapes above, the polymorphic super/sub-type case, and four controls that must keep matching - two separate MATCH clauses sharing an edge, the type-disjoint pair, the endpoint-disjoint pair, and a self-contained single part. 6 of the 12 fail without the fix.

Suites run green: the whole `com.arcadedb.query.opencypher` package (8603 tests, including the 3897-scenario openCypher TCK and `GAVEligibilityTest`), and `engine` + `network` + `server` + `bolt`.

https://claude.ai/code/session_01Ho2s8oThF2cuitxbMsuD86

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -2074,11 +2074,11 @@ public class CypherExecutionPlan {
         // For the first hop, use sourceVar; for subsequent hops, use the previous targetVar
         String currentSourceVar = sourceVar;
 
-        // Smart GAV eligibility: for each anonymous hop, check if its edge types overlap
-        // with any other hop's types in the same pattern. If disjoint, null relVar enables
-        // fast path (GAV/CSR). If overlapping, generate an internal anonymous variable to
-        // force edge-loading for cross-hop uniqueness checking.
-        final boolean[] hopNeedsEdgeTracking = computeHopEdgeTrackingNeeds(pathPattern);
+        // Smart GAV eligibility: for each anonymous hop, check whether any other hop of the same MATCH
+        // clause - this pattern part or another one across a comma - could be the same physical edge. If
+        // none can, a null relVar enables the fast path (GAV/CSR). If one can, generate an internal
+        // anonymous variable to force edge-loading, so relationship uniqueness has something to compare.
+        final boolean[] hopNeedsEdgeTracking = computeHopEdgeTrackingNeeds(pathPatterns, patternIndex);
 
         for (int i = 0; i < pathPattern.getRelationshipCount(); i++) {
           final RelationshipPattern relPattern = pathPattern.getRelationship(i);
@@ -2715,8 +2715,8 @@ public class CypherExecutionPlan {
               // For the first hop, use sourceVar; for subsequent hops, use the previous targetVar
               String currentSourceVar = sourceVar;
 
-              // Smart GAV eligibility: same analysis as ordered path
-              final boolean[] hopNeedsEdgeTrackingLegacy = computeHopEdgeTrackingNeeds(pathPattern);
+              // Smart GAV eligibility: same clause-scoped analysis as the ordered path
+              final boolean[] hopNeedsEdgeTrackingLegacy = computeHopEdgeTrackingNeeds(pathPatterns, patternIndex);
 
               for (int i = 0; i < pathPattern.getRelationshipCount(); i++) {
                 final RelationshipPattern relPattern = pathPattern.getRelationship(i);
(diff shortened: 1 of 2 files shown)


lvca:
> The only thing I'd want confirmed before merge is the optimizer-path coverage question above

Confirmed, and pinned in `theOptimizerDeclinesVariableLengthHopsButNotCommas` (commit 3dff8f8de).

The answer is that there is nothing to cover: the cost-based optimizer declines **every** variable-length hop, comma or no comma. `EXPLAIN` on each of these reports `Traditional Execution (Non-Optimized)`:

```cypher
MATCH (x:C)-[*1..2]->(y:A), (p:E)-[]->(q:A)                      -- TRAD
MATCH (p:E)-[]->(q:A), (x:C)-[*1..2]->(y:A)                      -- TRAD  (order does not matter)
MATCH (x:C)-[:R*1..2]->(y:A), (p:E)-[:R]->(q:A)                  -- TRAD  (nor does typing it)
MATCH (x:C)-[*]->(y:A), (p:E)-[]->(q:A)                          -- TRAD
MATCH p1 = shortestPath((x:C)-[*1..2]->(y:A)), (p:E)-[]->(q:A)   -- TRAD
MATCH (x:C)-[*1..2]->(y:A)                                       -- TRAD  (no comma needed)
```

So a test written as though it exercised the optimizer on a variable-length comma clause would in fact be running the traditional plan and passing for the wrong reason - worse than no test, since it would read as coverage.

What the tripwire asserts instead is the routing itself, including the converse: a comma clause of **fixed-length** hops *is* optimized (`MATCH (n0:A)<-[r1]-(:E), (n3:C)-[]->(:E)-[r2]->(n0)` → optimized plan). That is worth having for two reasons - it isolates the decline to the variable length rather than to the comma, and it is the evidence that the rest of the class already covers both implementations rather than only the traditional one. When the optimizer grows variable-length support the assertion fails, and whoever adds it is made to answer the uniqueness question on that path.

On the other two notes: the `O(P × total hops²)` redundancy is real and I left it - it is plan-build time on hop counts that are single digits in practice, and slicing a precomputed clause-wide result per part would trade a clearer function for arithmetic nobody can measure. Happy to change it if you disagree.

Separately, the earlier round's aside about `CypherExpressionBuilder` not applying the bare-`*` normalization is a real bug, filed as #6370: `[(a)-[*]->(b:A) | b.n]` returns `[]` where `[(a)-[*1..]->(b:A) | b.n]` returns `['a1']`, because both `minHops` and `maxHops` stay null and `isVariableLength()` is therefore false. Left out of this PR - it is the pattern-comprehension parser, not clause-scoped uniqueness.

Issue #6335, opened 2026-08-17
Cypher: SET/REMOVE of a label rewrites every edge of the vertex, so a label write on a supernode is O(degree) and churns RIDs


lvca:
Surfaced while fixing #6312/#6313. Not a defect in that fix - it is the cost of the model the fix had to work within, and it deserves its own decision.

A Cypher label set maps to an ArcadeDB type, and a record's type comes from the bucket it lives in, so there is no in-place retype. `LabelReplacements.replace()` therefore has to:

1. create a new vertex under the new type,
2. copy every property,
3. re-create **every incident edge**, in both directions, each of which also mutates the neighbour's edge list,
4. delete the original.

Two consequences worth deciding about rather than inheriting:

**Cost.** `SET n:Label` on a vertex with a million edges rewrites a million edge records and touches a million neighbour edge lists, for what the user wrote as a metadata change. On a supernode this is the difference between a millisecond and a stall, and nothing in the query says so. It also interacts badly with the striping/append-merge work: a label write undoes the locality those paths build.

**Identity.** The vertex's RID changes, and so does every edge's. Anything holding a RID across the write - an application-side reference, a stored link from a non-graph type, an external index, a client that cached the result of an earlier query - is left pointing at a deleted record. #6326 makes the *query's own* row follow the move; nothing can make an outside reference follow it.

Worth exploring, roughly in order of appetite:

- **Reject rather than silently pay it.** A label write on a vertex above a degree threshold could fail with a message that says why and what to do instead, which is at least honest. Cheap, and strictly better than the stall.
- **Multi-label without a type change.** Model the extra labels as a property (or a small dedicated structure) consulted by `Labels.getLabels()`/`hasLabel()` and by the label-scan planning, so a label write is a property write and the record never moves. This is the real fix and the large one: every place that resolves a label to a type - `ensureCompositeType`, `NodeByLabelScan`, the composite-type naming - has to learn the second source, and existing databases carry composite types that must keep working.
- **Keep the rewrite but make it cheaper.** Edges are re-created one at a time through the public API today; a bulk path that moves the edge lists wholesale would cut the constant, though not the complexity, and would not help the RID churn at all.

Whatever is chosen, the current behaviour should at minimum be documented: today nothing warns that `SET n:Label` is a write proportional to degree that invalidates the node's RID.

lvca:
Settled in #6348, merged as 34482fde0, with the first of the three options in the issue.

[...]

PR #6371, opened 2026-08-18
fix(#6363): a vertex's labels are the ones it answers to, and adding one keeps the type that carries them

### LINKED ISSUE #6363: openCypher: three follow-ups from #6352 - labels() contradicts every label predicate under inheritance (and a label write relocates the vertex out of its own type), a disjunction scan's stale cardinality, and the missing non-throwing type accessor
Three findings from the work on #6352 (merged as `a58038b6`, PR #6357). Independent of each other, filed together because they were all surfaced by the same pass over the label machinery. Item 1 is the one that matters; items 2 and 3 are small and were raised in that PR's review.

All three reproduce on `a58038b6`.

---

## 1. `Labels.getLabels(Vertex)` drops the vertex's own type, so `labels()` contradicts every label predicate - and a label write moves the vertex out of its own type

`Labels.getLabels` (engine/src/main/java/com/arcadedb/query/opencypher/Labels.java:116) decides a vertex's labels like this:

```java
final List<DocumentType> superTypes = type.getSuperTypes();

if (superTypes.isEmpty()) {
  // Single-label vertex - type name is the label
  return List.of(typeName);
}

// Multi-label vertex - supertypes are the labels
```

That rule is written for one shape: a synthetic `A~B` composite, whose own name is an implementation detail and whose supertypes *are* the labels. For every other shape - which is to say, for ordinary ArcadeDB type inheritance - it returns the wrong set, because "has supertypes" and "is a composite" are not the same question. `Labels.isCompositeTypeName` already exists and is not consulted here.

### 1a. The read side: `labels()` disagrees with `MATCH` and with `WHERE n:Label`

```sql
CREATE VERTEX TYPE Employee;
CREATE VERTEX TYPE Manager EXTENDS Employee;
INSERT INTO Manager SET k = 'm1';
```

| query | returns | expected |
|---|---|---|
| `MATCH (n) RETURN labels(n)` | `["Employee"]` | `["Manager","Employee"]` |
| `MATCH (n:Manager) RETURN count(n)` | `1` | `1` |
| `MATCH (n) WHERE n:Manager RETURN count(n)` | `1` | `1` |

The engine matches `:Manager` twice over and then reports a label list that does not contain `Manager`. A client that reads a node and asks what it is gets an answer contradicted by the query that found it.

With a type extending a composite, the synthetic name leaks out instead:

```sql
-- after CREATE (:Author:Topic) has created the Author~Topic composite
CREATE VERTEX TYPE Special EXTENDS `Author~Topic`;
INSERT INTO Special SET k = 'sp1';
```

| query | returns | expected |
|---|---|---|
| `MATCH (n {k:'sp1'}) RETURN labels(n)` | `["Author~Topic"]` | `["Special","Author","Topic"]` |
| `MATCH (n:Author) RETURN count(n)` | counts it | counts it |

`Author~Topic` is an internal encoding. It should never reach a result set, and the two labels it encodes should.

### 1b. The write side: `SET n:Extra` deletes the vertex's own label

`getLabels` is not only the `labels()` function (`LabelsFunction:60`). It is also what `SetStep:451`, `RemoveStep:263` and `MergeStep:1397` build the *new* type from when a query adds or removes a label. So the wrong label set is written back:

```sql
CREATE VERTEX TYPE Employee;
CREATE VERTEX TYPE Manager EXTENDS Employee;
INSERT INTO Manager SET k = 'm1';
```
```cypher
MATCH (n:Manager) SET n:Extra
```

| after the SET | result | expected |
|---|---|---|
| `labels(n)` | `["Employee","Extra"]` | `["Manager","Employee","Extra"]` |
| the record's type | `Employee~Extra` | still a `Manager` (carrying `Extra`) |
| `MATCH (n:Manager) RETURN count(n)` | **`0`** | `1` |
| `MATCH (n:Employee) RETURN count(n)` | `1` | `1` |
| `MATCH (n:Extra) RETURN count(n)` | `1` | `1` |

Adding a label **removed** one. The vertex was relocated out of `Manager` into a freshly invented `Employee~Extra`, and every query written against `:Manager` stops seeing it. No error, no warning. Modelling a hierarchy in SQL and touching it from Cypher is an ordinary thing to do here, and one `SET n:Label` is enough to lose the subtype.

### Suggested shape

`getLabels` should answer with the type names the vertex is an `instanceOf` - its own type and every ancestor - minus the synthetic composite names (`isCompositeTypeName`) and minus the base `V`/`Vertex`. For the cases above: `Manager` -> `[Employee, Manager]`, `Special` -> `[Author, Special, Topic]`, a bare `A~B` composite -> `[A, B]` exactly as today.

That is the read side. The write side needs a decision that is worth making explicitly rather than falling out of the fix, because the current composite scheme cannot express it: **a label added to a vertex of an inherited type must not cost it its type.** `Manager` + `Extra` wants a type that extends both `Manager` and `Extra`, not one that extends `Employee` and `Extra`. Deriving the composite from the *own type* rather than from the supertype list gets there (`Extra~Manager` extending both), and it keeps `MATCH (n:Manager)` and `MATCH (n:Employee)` both working through inheritance. Worth confirming that is the intended semantics before the write path changes - and note it interacts with the label-write cost documented in `docs/6335-cypher-label-write-cost.md`.

A regression test should cover: a subtype vertex's `labels()`, a type extending a composite, `SET`/`REMOVE` of a label on a subtype vertex (asserting the original label survives, via `MATCH` and not only via `labels()`), and the plain composite case, which must keep behaving exactly as it does now.

---

## 2. `NodeByLabelDisjunctionScan.estimatedCardinality` is a single-label estimate for a multi-type scan

Raised in #6357's review; pre-existing, and already noted in that operator's own javadoc as tracked separately - but never given an issue, so here it is.

`estimatedCardinality` is inherited from the anchor selector, which computes it for the **first label alone**. The operator scans every type satisfying any alternative, so the true upper bound is the sum across those types. `(n:A|B)` is therefore costed as if only `:A` existed, which can bias join ordering whenever a disjunction anchor competes with another candidate - the under-estimate makes the disjunction look like the cheaper side to drive from.

`Labels` now has the type selection in one place, so the estimate can be summed over exactly the types the scan will visit rather than guessed.

## 3. `Schema` has no non-throwing type accessor

`Schema.getType(String)` throws when the type is absent, so every caller that can tolerate absence writes `existsType(name)` then `getType(name)` - two probes of the same map. There are ~40 `existsType` call sites in the openCypher package alone and ~75 in the engine's main sources, a good share of them this pair.

A `getTypeOrNull(String)` (or `Optional<DocumentType> findType(String)`) on the `Schema` interface would collapse them. Deliberately declined inside #6357 - adding public API to save one `HashMap` probe is not a bug fix - but as a small, intentional API addition it is worth doing once rather than continuing to spell it out at every call site. Cosmetic priority; no behaviour change.

lvca:
Fixes #6363.

## 1. `labels()` contradicted every label predicate, and a label write relocated the vertex

`Labels.getLabels` decided a vertex's labels from a single question - does the type have supertypes - and
answered "the supertypes" when it did, "its own name" when it did not. That rule fits exactly one shape: the
synthetic `A~B` composite, whose own name is an encoding and whose supertypes really are its labels. Under
ordinary ArcadeDB inheritance it was wrong in both directions.

| shape | `labels(n)` before | after |
|---|---|---|
| `Manager EXTENDS Employee` | `["Employee"]` | `["Employee","Manager"]` |
| `Special EXTENDS` \`Author~Topic\` | `["Author~Topic"]` | `["Author","Special","Topic"]` |
| `CREATE (:Author:Topic)` | `["Author","Topic"]` | unchanged |

`getLabels` is not only the `labels()` function - `SetStep`, `RemoveStep` and `MergeStep` build the **new**
type from it - so the wrong list was written back. `MATCH (n:Manager) SET n:Extra` moved the record to a
freshly invented `Employee~Extra`, and `MATCH (n:Manager)` then returned **0 rows**. Adding a label removed
one, silently, with no error and no warning.

Reading and relabelling are now two questions with two answers:

- **`getLabels`** - every type name the vertex is an `instanceOf`, sorted, minus the synthetic composite
  names. `L IN labels(n)` and `n:L` now answer the same question, which is the invariant Neo4j promises and
  the one the fix is measured against.
- **`getOwnLabels`** - the minimal set that reproduces the current type: `[Manager]`, *not*
  `[Employee, Manager]`. So `SET n:Extra` builds `Extra~Manager` extending both - the vertex stays a
  `Manager`, therefore stays an `Employee`, and gains `Extra`. This is the write-side shape the issue asked
  to confirm before changing anything.
- Adding a label the vertex already answers to (its own, or inherited) changes nothing and is not counted in
  `labels-added`. Not a hidden no-op: the label *is* present afterwards.

`V`/`Vertex` still means "no labels" for a node's **own** type, but is an ordinary label as a supertype -
`V` is a label a query may write, and the openCypher TCK writes it (`CREATE (b:U:V:W:X:Y:Z)`). Getting that
wrong is the one regression this branch caught on itself, and it is now pinned by a test rather than only by
the TCK.

### One decision worth your eye

`REMOVE n:Employee` on a vertex of type `Manager EXTENDS Employee` has **no correct outcome**: no type the
vertex could move to answers "no" to `:Employee` while answering "yes" to `:Manager`. The three candidates
were destroy the subtype (what the old code effectively did), no-op silently (leaves `n:Employee` true after
`REMOVE n:Employee` - the same class of lie this issue is about), or **refuse with a message naming both
labels**. I took the third, as a `CommandSemanticException` (HTTP 400): the query is not honourable against
this schema and no retry changes that. Removing a label the vertex simply does not have stays a no-op, as in
Neo4j, and removing both the subtype and the label it implies works and leaves an unlabelled node. Happy to
switch it to a silent no-op if you prefer the quieter behaviour.

## 2. A disjunction anchor's cardinality

`NodeByLabelDisjunctionScan` scans every type any alternative accepts, but its estimate came from the anchor
selector's single-label lookup - the first alternative alone - so `(n:A|B)` was priced as if only `:A`
existed and looked like the cheap side to drive a join from. `AnchorSelector` now sums the count over exactly
`Labels.matchingVertexTypes`, the same list the scan walks, counted non-polymorphically so a subtype
answering to two alternatives is charged once. A disjunction anchor is also no longer offered an index seek
it could never use: `IndexSelectionRule` builds the disjunction scan whatever the estimate claims, so
claiming a seek only mis-priced the plan it produces.

## 3. `Schema.getTypeOrNull`

The non-throwing accessor the `existsType(x) ? getType(x) : null` pairs wanted. A `default` method spelled as
that same pair, so an out-of-tree `Schema` keeps compiling; `LocalSchema` and `RemoteSchema` override it with
one lookup. The `existsType(x) && getType(x) instanceof Y` idiom collapses to `getTypeOrNull(x) instanceof Y`
at the sites that had it.

## Verification

- `CypherLabelsInheritanceIssue6363Test` (11) and `CypherDisjunctionCardinalityIssue6363Test` (5), new. 8 of
  the first class's assertions fail on `a58038b6`.
- engine unit lane **12600 / 0 failures**
- openCypher TCK **3897 / 0 failures**
- `engine, network, integration, server, mcp, ha, bolt` all green in one run

`docs/6363-cypher-labels-inheritance.md` records the semantics and why the `REMOVE` corner has no third
option.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/structure/BoltStructureMapper.java
@@ -158,7 +158,7 @@ public class BoltStructureMapper {
     final long id = ridToId(rid);
     final String elementId = rid.toString();
 
-    // Get all labels (supertypes for composite types)
+    // Every label the node answers to: its own type and its ancestors', minus the synthetic composite names
     final List<String> labels = Labels.getLabels(vertex);
 
     // Get properties
--- a/docs/6363-cypher-labels-inheritance.md
+++ b/docs/6363-cypher-labels-inheritance.md
@@ -1,0 +1,112 @@
+# #6363: what a vertex's labels are, and what a label write is allowed to change
+
+Issue: https://github.com/ArcadeData/arcadedb/issues/6363
+
+## What happened
+
+`Labels.getLabels(Vertex)` decided a vertex's labels from one question - "does the type have supertypes?" -
+and answered:
+
+- **no supertypes** -> the type's own name is the label,
+- **supertypes** -> the supertype names are the labels, and the type's own name is not one.
+
+That rule is right for exactly one shape, the synthetic `A~B` composite the multi-label support creates:
+its own name is an encoding, and its supertypes really are the labels. Applied to ordinary ArcadeDB type
+inheritance it is wrong in both directions:
+
+| shape | `labels(n)` before | `MATCH (n:...)` matched |
+|---|---|---|
+| `Manager EXTENDS Employee` | `["Employee"]` | `:Manager` and `:Employee` |
+| `Special EXTENDS ` \`Author~Topic\` | `["Author~Topic"]` | `:Author` and `:Topic` |
+
+So the engine matched a node by a label and then reported a label list not containing it, and an internal
+type name leaked into a result set.
+
+`getLabels` is not only the `labels()` function. `SetStep`, `RemoveStep` and `MergeStep` build the **new**
+type from it, so the wrong list was written back: `MATCH (n:Manager) SET n:Extra` moved the record to a
+freshly invented `Employee~Extra`, and `MATCH (n:Manager)` then returned **0 rows**. Adding a label removed
+one, silently.
+
+## The rule now
+
+Two different questions, two methods, because a relabelling does not want the same answer a reader does.
+
+**`Labels.getLabels`** - what the node answers to. Its own type name and every ancestor's, sorted, minus
+the synthetic composite names. `Manager` -> `[Employee, Manager]`; `Special` -> `[Author, Special, Topic]`;
+a bare `Author~Topic` -> `[Author, Topic]`, exactly as before. The invariant is Neo4j's and is what the fix
+is measured against: **`L IN labels(n)` and `n:L` answer the same question**, before and after a write.
+
+Two names are special and only one of them is filtered here:
+
+- a **composite** name (`A~B`) is an encoding, so the walk goes *through* it to the labels it encodes. A
+  type merely *named* with a `~` but declaring no supertype is a type somebody created and keeps its name.
+- `V`/`Vertex` is the type a node lands in when it carries no label at all, so a node whose **own** type is
+  `V` reports an empty list. A *supertype* called `V` is not filtered: `V` is an ordinary label a query may
+  write, and the openCypher TCK does write it (`CREATE (b:U:V:W:X:Y:Z)`).
+
+**`Labels.getOwnLabels`** - what a relabelling must reproduce. The minimal set that rebuilds the current
+type: the type's own name, or, for a composite, the labels it encodes. `Manager` -> `[Manager]`, *not*
+`[Employee, Manager]`, because `Employee` comes back on its own through the hierarchy and naming it in the
+composite instead of the subtype that carries it is precisely what dropped the subtype.
+
+So `MATCH (n:Manager) SET n:Extra` now builds `Extra~Manager`, extending both. The vertex is still a
+`Manager`, therefore still an `Employee`, and now an `Extra`. Every label predicate that held before the
+write still holds after it.
+
+Adding a label the vertex already answers to - its own, or an inherited one - changes nothing and is not
+counted in `labels-added`. It is not a no-op that hides a failure: the label *is* present afterwards.
+
+## The one thing that cannot be done, and is now said out loud
+
+`REMOVE n:Employee` on a vertex of type `Manager EXTENDS Employee` has no correct outcome. No type the
+vertex could be moved to answers *no* to `:Employee` while still answering *yes* to `:Manager` - the schema
+says one implies the other. The three candidate behaviours were:
+
+- **move it to `Employee`'s sibling-less remainder** (what the old code effectively did, via a label list
+  that had already lost `Manager`): destroys the subtype to satisfy a removal. Worst of the three.
+- **silently do nothing**: leaves `n:Employee` true after `REMOVE n:Employee`, which is the same class of
+  lie this issue is about, one clause over.
+- **refuse, with a message naming both labels**: chosen. `CommandSemanticException` (HTTP 400), because the
+  query is not honourable against this schema and no retry will change that.
+
+Removing a label the vertex simply does not have stays a no-op, as in Neo4j. Removing *both* the subtype
+and the label it implies (`REMOVE n:Manager, n:Employee`) is fine and leaves an unlabelled node: nothing
+that remains implies `Employee`.
+
+This interacts with [#6335](6335-cypher-label-write-cost.md): a label write still moves the record and is
+still O(degree). What changes here is only *which* type it moves to.
+
+## The two small items
+
+**A disjunction's cardinality estimate.** `NodeByLabelDisjunctionScan` scans every type any alternative
+accepts, but its estimate came from the anchor selector's single-label lookup - the first alternative
+alone - so `(n:A|B)` was priced as if only `:A` existed and looked like the cheap side to drive a join
+from. `AnchorSelector` now sums the count over exactly `Labels.matchingVertexTypes`, the same list the scan
+walks, counted non-polymorphically so a subtype answering to two alternatives is charged once. A
+disjunction anchor is also no longer offered an index seek it could never use: `IndexSelectionRule` builds
+the disjunction scan whatever the estimate claims, so claiming a seek only mis-priced the plan it produces.
+
+**`Schema.getTypeOrNull`.** `getType` throws on an absent type, so every caller that could tolerate absence
+wrote `existsType(name) ? getType(name) : null` - two probes of the same map, and an exception used as
+control flow whenever somebody forgot the guard. `Schema` now has a non-throwing accessor. It is a
+`default` method spelled as that same pair, so an out-of-tree `Schema` keeps compiling; `LocalSchema` and
+`RemoteSchema` override it with a single lookup. The `existsType(x) && getType(x) instanceof Y` idiom
+collapses to `getTypeOrNull(x) instanceof Y`.
+
+## Affected components
+
+- `com.arcadedb.query.opencypher.Labels` - `getLabels`, new `getOwnLabels` and `impliedBy`,
+  `matchingVertexTypes` made visible to the cost model
+- `com.arcadedb.query.opencypher.executor.steps.SetStep` / `RemoveStep` / `MergeStep` - the write side
+- `com.arcadedb.query.opencypher.optimizer.AnchorSelector`, `statistics.StatisticsProvider`,
+  `statistics.CostModel` - the disjunction estimate
+- `com.arcadedb.schema.Schema` / `LocalSchema`, `com.arcadedb.remote.RemoteSchema` - `getTypeOrNull`
+
+## Tests
+
+- `CypherLabelsInheritanceIssue6363Test` - the read side under inheritance and under a type extending a
+  composite, `SET`/`REMOVE` on a subtype vertex asserted through `MATCH` and not only through `labels()`,
+  the refused inherited removal, the `V`-as-a-label case, and the plain composite, which must keep behaving
+  exactly as it did
+- `CypherDisjunctionCardinalityIssue6363Test` - the summed estimate pinned through `EXPLAIN`, and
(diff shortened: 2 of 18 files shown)


lvca:
Thanks - acted on two of the three minor points, and declined the third with a reason.

**1. `isCompositeTypeName` heuristic (`13aa58233`).** You suggested a follow-up issue; I closed it here instead, because the question turns out to have an exact answer rather than a better heuristic. A label composite's name is *precisely* the deduplicated, sorted, separator-joined names of its own supertypes - that is what `ensureCompositeType` writes and nothing else produces - so `isLabelComposite(type, superTypes)` now asks that structurally instead of looking for a `~`. Fewer than two supertypes is never a composite, since a single label is the type itself. A user type called ```a~b``` keeps its name in both `getLabels` and `getOwnLabels`, which closes exactly the write-path exposure you identified: under the old heuristic the next `SET` on such a vertex would have moved it out of its own type, the same failure this issue is about one shape over. Pinned by `aUserTypeWhoseNameMerelyContainsTheSeparatorKeepsIt`, which asserts through `MATCH` after a `SET` and not only through `labels()`.

**2. Release-notes callout for the `REMOVE` refusal.** Added to `docs/release-26.9.1.md`, explicitly marked **Breaking change**, with both escape hatches spelled out (remove the subtype as well, or change the hierarchy in SQL).

**3. Javadoc length - declining.** The surrounding code sets this style deliberately: `Labels.matches`, `Labels.iterateMatchingVertices` and `matchingVertexTypes` all carry the same issue-numbered narrative from #6338/#6352, as does `LabelReplacements.replace` from #6335. Trimming only the new methods would leave the file inconsistent with itself. The `docs/` markdown does carry the long form as you suggest - it just does so in addition, which is the established pattern here.

On your note about independent verification: the openCypher + schema lane re-ran green after the structural change at **9044 tests / 0 failures**, with the **TCK at 3897 / 0**. The earlier full sweep across engine, network, integration, server, mcp, ha and bolt was also green in one run. Worth adding that the TCK is not a formality here - it caught a genuine regression in the first draft of this fix, where `V` was filtered out of the ancestor walk and `CREATE (b:U:V:W:X:Y:Z)` lost a label. That case is now a unit test too.

lvca:
Round 2: one fix, two accepted as-is.

**1. Over-counted `labels-removed` on a repeated label (`10dd0c636`).** Fixed - good catch. The loop now skips non-first occurrences, so `REMOVE n:Foo:Foo` reports one label removed for one real change. A label is set membership, not a count. New test `aLabelNamedTwiceInOneClauseIsCountedOnce` pins both directions, and it confirms your reading that `SET` was already correct via its `allLabels.contains` guard - only `REMOVE` needed the change. The parser does accept the duplicate, so this was reachable rather than theoretical.

**2. `isLabelComposite` false positive on a hand-made ```A~B``` EXTENDS A, B.** No change - and I'd put it more strongly than an accepted trade-off: reporting `[A, B]` there is *correct*, not merely tolerable. `ensureCompositeType` reuses that exact type for `CREATE (:A:B)` - it looks the name up, adds any missing supertypes, and hands the type back rather than creating its own - so a type matching the generated shape genuinely **is** the composite the label machinery uses for those two labels. A vertex of it answers to `:A` and `:B` and to nothing called `A~B` that a user would recognise. Reporting `[A~B]` would be the wrong answer, so the window is not one I want to narrow further.

**3. `TreeSet` recursion on the hot path.** No change. The fast path (`superTypes.isEmpty()` -> `List.of(typeName)`) covers every ordinary single-label vertex, which is the overwhelming majority of `labels()` calls and of label writes. The set only allocates for a type that actually has ancestors - and there the old code allocated an `ArrayList` and sorted it anyway, so this is a wash rather than a regression. Agreed it is worth watching if deep composite hierarchies ever become common.

Re-verified after the fix: openCypher lane **8610 / 0 failures**, TCK **3897 / 0**.

lvca:
Round 3: closed the one coverage gap (`392d65847`).

**No-op `REMOVE n:NotPresentAtAll`.** Fair point - the refusal path was pinned, its counterpart was not, and the two are the same decision read from opposite sides. `removingALabelTheVertexDoesNotHaveIsANoOp` now covers both shapes of absent label (one whose type does not exist, and one whose type exists but this vertex is not of it), asserts nothing moved, and asserts no type was invented for the absent label on the way past - that last one matters because the write path calls `getOrCreateVertexType` further down, and the early return is what keeps it from being reached.

The other two notes need no change: the breaking-change callout is already in `docs/release-26.9.1.md`, and thank you for confirming the docs match the implementation.

Also worth recording your findings, since they were the non-obvious things to check: `impliedBy` being evaluated against `remainingLabels` computed once for the whole clause is deliberate - it is what makes `REMOVE n:Employee, n:Manager` collapse to `V` instead of throwing, i.e. removing both is the honest way to reach an outcome that has no correct answer one label at a time. And the exception landing before any mutation is what keeps a partial removal from being possible within one clause.

openCypher lane re-run after the addition: **4714 / 0 failures**.

From my side this is ready to merge - three rounds, no blocking issues in any of them, every valid point acted on and the declined ones argued rather than ignored.

lvca:
Round 4: both notes addressed (`6f8e95d09`), comments and one test only - no production behaviour changed.

**The `indexOf` dedup.** Inline comment added on that exact line, saying why it is a linear scan rather than a Set: a clause holds a handful of labels, and this allocates nothing on a path that usually finds no repeat.

**The composite-name-as-label no-op.** You traced this by hand and reached the right conclusion - and you also spotted the real risk in it: "the counting loop and the statistics update are now two steps apart". Rather than leave that as knowledge living in a review comment, I did both halves of making it durable. The guard now carries a comment saying *why* it must stay ahead of `addLabelsRemoved` - the count asks `instanceOf`, which says yes to a vertex's own composite name, while that name is not among the labels `remainingLabels` is rebuilt from - and `namingACompositeTypeItselfAsALabelToRemoveChangesNothing` pins the pairing, asserting `labels-removed` is 0 and the type is untouched. A refactor of either half now fails a run instead of quietly counting a no-op as a change.

Thanks also for chasing whether `V` is ever an implicit supertype. That was the one thing in this fix I most wanted independently checked - my first draft filtered `V` throughout the ancestor walk and the TCK caught it losing a label from `CREATE (b:U:V:W:X:Y:Z)`, so your confirmation that a plain `CREATE VERTEX TYPE X` never wires in `V` is exactly the missing half of the argument.

Final verification: openCypher lane **8612 / 0**, TCK **3897 / 0**.

Issue #5885, opened 2026-08-07
19 of 60 SQL methods throw raw JDK exceptions on missing or negative arguments (companion to #5884)


ruispereira:
## Summary

The same missing-argument-validation problem reported in #5884 for SQL *functions* also affects SQL *methods*, in a different registry: **19 of the 60** registered methods answer with a raw JDK exception instead of a validation error.

`SELECT 'abc'.substring()` gives `ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0`. `SELECT 'abc'.right(-1)` gives `StringIndexOutOfBoundsException: Range [4, 3) out of bounds for length 3`.

Version: 26.8.1.

## How it was measured

Every name declared as `NAME = "…"` under `engine/src/main/java/com/arcadedb/query/sql/method/` was invoked against five receivers (`'abc'`, `null`, `0`, `[]`, `{}`) with three argument sets (none, `'x'`, `-1`):

```java
db.query("sql", "SELECT " + receiver + "." + method + "(" + args + ") AS r");
```

| outcome | count |
|---|---|
| returned a result | 489 |
| clean error (ArcadeDB-typed, or `IllegalArgumentException` with a message) | 329 |
| **raw JDK exception** | **82** |

900 calls, 60 methods, 19 of them affected.

## Affected methods

```
append, asdate, asdatetime, charat, convert, field, format, ifempty, include,
indexof, lastindexof, left, prefix, replace, right, split, substring,
trimprefix, trimsuffix
```

## Two distinct sub-cases

**Missing arguments** — the same shape as #5884:

```
SELECT 'abc'.substring()   -> ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
SELECT 'abc'.split()       -> ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
SELECT 'abc'.trimprefix()  -> ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
SELECT {}.replace('x')     -> ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

**Negative arguments**, which #5884 did not cover — here the argument count is right and the *value* is unchecked:

```
SELECT 'abc'.right(-1)     -> StringIndexOutOfBoundsException: Range [4, 3) out of bounds for length 3
SELECT [].right(-1)        -> StringIndexOutOfBoundsException: Range [3, 2) out of bounds for length 2
[...]

Issue #5799, opened 2026-08-03
`sum()` and `avg()` silently ignore non-numeric inputs instead of reporting a type error


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`
Build: `85eb71bb19a39fd1eb36a6fe3ee742862d47d6f6/1785731865672/main`

### Environment

* Linux x86_64, Docker, OpenJDK
* `arcadedata/arcadedb:latest`, freshly pulled on 2026-08-03
* HTTP command API
* Query language: Cypher

### Description

The aggregate functions `sum()` and `avg()` silently accept non-numeric input values instead of reporting a type error.

Non-numeric values appear to be ignored:

* `sum(['1', '2'])` returns `0`
* `sum([true, 1])` returns `1`
* `avg(['a', 'b'])` returns `null`

This behavior masks invalid input types. A successful result such as `0`, `1`, or `null` is returned even though the aggregate input contains values outside the numeric domain.

Null handling and empty-input handling remain correct, so the problem is specific to unsupported non-numeric values rather than aggregation in general.

### Steps to reproduce

```cypher
UNWIND ['1', '2'] AS x
RETURN sum(x) AS value;
-- 0
```

```cypher
UNWIND ['1.5', '2.5'] AS x
RETURN sum(x) AS value;
-- 0
```

```cypher
UNWIND [true, 1] AS x
RETURN sum(x) AS value;
-- 1
```

```cypher
UNWIND ['a', 'b'] AS x
RETURN avg(x) AS value;
-- null
```

### Expected behavior

`sum()` and `avg()` should report a client-facing type error when an input value is neither numeric nor another type explicitly supported by the aggregate.

Numeric strings should not be converted implicitly. Callers that require conversion can request it explicitly before aggregation.

### Actual behavior

| Query                                     | ArcadeDB | Neo4j 2026.06 |
| ----------------------------------------- | :------: | :-----------: |
| `UNWIND ['1','2'] AS x RETURN sum(x)`     |   **0**  |   Type error  |
| `UNWIND ['1.5','2.5'] AS x RETURN sum(x)` |   **0**  |   Type error  |
| `UNWIND [true] AS x RETURN sum(x)`        |   **0**  |   Type error  |
| `UNWIND [true,1] AS x RETURN sum(x)`      |   **1**  |   Type error  |
| `UNWIND ['a','b'] AS x RETURN avg(x)`     | **null** |   Type error  |
[...]

PR #6408, opened 2026-08-18
fix(#6382, #6385, #6388, #6389, #6390, #6393): an operator with no visitor is dropped from the tree, and a mistake in a call is the caller's error, not the server's

### LINKED ISSUE #6382: Full-text CONTAINSTEXT splits the literal on ':' and skips the single-property isUnqualified normalization, returning no matches
### Summary
`CONTAINSTEXT` on a full-text index parses its literal argument as a `field:value` query when it contains a colon, and never applies the single-property "unqualified" normalization the Lucene path has. On a single-property FT index this makes any value containing a colon (or an intentional `prop:term`) return **no matches**.

### Where (HEAD `ca89f56`)
`engine/src/main/java/com/arcadedb/index/fulltext/LSMTreeFullTextIndex.java`, `parseQueryTerms` L805-829:
```java
final int colonIdx = part.indexOf(':');
if (colonIdx > 0 && colonIdx < part.length() - 1) {
  final String fieldName = part.substring(0, colonIdx);
  final String value = part.substring(colonIdx + 1);
  terms.add(new QueryTerm(fieldName, value));   // treats "foo:bar" as field=foo value=bar
} else
  terms.add(new QueryTerm(null, part));
```
A single-property index stores only unprefixed tokens (CLASSIC `get()` prefixing, L254), so a lookup for a field-qualified key finds nothing. The Lucene-backed executor solves this with `FullTextQueryExecutor.isUnqualified` (L480-484):
```java
private boolean isUnqualified(final String field) {
  if (field == null || field.isEmpty() || DEFAULT_FIELD.equals(field)) return true;
  return propertyNames.size() == 1 && propertyNames.get(0).equals(field);
}
```
`parseQueryTerms` never calls anything equivalent, so the fix was not ported to this path.

### Repro
Single-property FT index on `label`. `WHERE label CONTAINSTEXT 'foo:bar'` → parsed as `field=foo, value=bar`, looked up as a qualified key that a single-property index never stores → returns nothing (expected: matches rows whose `label` contains the literal `foo:bar`, or at least the tokens `foo`/`bar`). Same for incidental colons in literal text (`3:30`, timestamps) and an intentional `content:term` on a single-property index. Multi-property indexes with real field names are unaffected.

### Suggested fix
In `parseQueryTerms`, treat a qualifier as unqualified when the sole indexed property name equals the field (mirror `isUnqualified`); and because `CONTAINSTEXT` is documented as literal text rather than a query language, prefer not splitting on colon at all here — let the analyzer tokenize.

### Scope / confidence
High confidence on the code path (traced through the planner). Medium-high that it is a live, unfiled defect (the asymmetry with the deliberately-fixed Lucene path). Mitigation: use `SEARCH_INDEX(...)` or avoid colons in `CONTAINSTEXT` literals.

### LINKED ISSUE #6385: SQL astar/dijkstra heuristic uses the source coordinate as the current-node position
### Summary
The A* heuristic stores the **source** vertex's coordinate where the **current** node's coordinate belongs, so every heuristic cost is computed as if the node were the start. A* then explores on a corrupted `f = g + h` and can return non-optimal paths (and does redundant work). Because `dijkstra` delegates to A*, any Dijkstra call supplying axis coordinates is affected too.

### Where (HEAD `ca89f56`)
`engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionAstar.java` L446-453:
```java
final Double s = doubleOrDefault(paramSourceVertex.get(paramVertexAxisNames[i]), 0);  // L446 source
final Double c = doubleOrDefault(node.get(paramVertexAxisNames[i]), 0);               // L447 current node
...
if (c != null)
  cList.put(paramVertexAxisNames[i], s);   // L453 <-- stores s (source), not c
```
`c` (the current node's coordinate) is computed and then used only in the `if (c != null)` guard — it is never stored; `cList` (the "current position" fed to every heuristic: `getManhattanHeuristicCost`/`getEuclideanHeuristicCost`/… L461-477) is filled with the source coordinate. The dead variable `c` is the tell.

### Repro
```sql
SELECT astar(#12:0, #12:99, 'weight',
  {'direction':'OUT','parVertexAxisNames':['x','y'],'heuristicFormula':'EUCLIDEAN'})
```
h(n) should estimate distance from node *n* to the goal; instead it is always computed as if n were the source, so h is near-constant and the priority is wrong. `SQLFunctionDijkstra.execute` calls `new SQLFunctionAstar().execute(...)`, so a Dijkstra query with axis coordinates degrades the same way.

### Suggested fix
`cList.put(paramVertexAxisNames[i], c);`

### Scope / confidence
High that the code is wrong (the dead `c`). Medium on how badly any given query degrades: an admissible-but-weakened heuristic often still yields the optimal path, so the visible symptom is intermittent sub-optimal paths / slower searches rather than a guaranteed wrong answer every time.

### LINKED ISSUE #6388: Several SQL time/date functions throw raw exceptions on valid input (duration/timeBucket/lag/lead/date); sysdate ignores its zone argument
### Summary
Several SQL time/date functions throw a raw, uncaught `RuntimeException` (surfacing as a 500) on ordinary valid input, and one silently ignores its argument. Grouping them since they share the "missing validation / wrong catch" shape in the same package.

### Findings (all HEAD `ca89f56`, `engine/src/main/java/com/arcadedb/function/sql/time/`)
1. **`sysdate(<zoneid>)` silently ignores the zone** — `SQLFunctionSysdate.java` L50-53 reads `params[1]` inside `if (params.length > 1)`, but the documented syntax `sysdate([<zoneid>])` passes the zone as `params[0]`. `SELECT sysdate('America/New_York')` returns server-local time, zone dropped. Fix: read `params[0]`.
2. **`duration(n, 'year'|'month'|'week')` crashes** — `SQLFunctionDuration.java` L51 `Duration.of(amount, DateUtils.parsePrecision(...))`. `parsePrecision` maps year/month/week to `YEARS`/`MONTHS`/`WEEKS`, which `java.time.Duration.of` rejects (estimated duration) → `UnsupportedTemporalTypeException`. `SELECT duration(1,'year')`. Fix: convert weeks→days×7 etc., or reject with a clean message.
3. **`timeBucket('0s', ts)` divide-by-zero** — `SQLFunctionTimeBucket.java` L59 `(timestampMs / intervalMs) * intervalMs`; `parseInterval("0s")` returns 0 → `ArithmeticException: / by zero`. (`params[0]==null` also NPEs.) Fix: reject `intervalMs <= 0`.
4. **`lag()`/`lead()` negative offset → IOOBE; null timestamp → NPE** — `SQLFunctionLag.java`/`SQLFunctionLead.java` L88-92: `pairs.sort(Comparator.comparing(p -> ((Comparable) p[1])))` then `pairs.get(i - offset)` / `get(i + offset)`; `offset` is never sign-checked and `getMinArgs()==1` lets the timestamp be null. `SELECT ts.lag(v,-1,ts)` → IOOBE; `SELECT ts.lag(value)` → sort NPE. Fix: require `offset >= 0`, `Comparator.nullsLast(...)`, or raise arity.
5. **`date()` — malformed pattern / unknown zone escape the catch** — `SQLFunctionDate.java` L91-95 catches only `DateTimeParseException`; `getFormatter(badPattern)` throws `IllegalArgumentException` and `ZoneId.of(badZone)` throws `ZoneRulesException` (a `DateTimeException`, not `DateTimeParseException`). `SELECT date('2020','qqzz')` / `SELECT date('2020-01-01','yyyy-MM-dd','Bogus/Zone')`. Fix: also catch `IllegalArgumentException`/`DateTimeException`.
6. **(resource) unbounded formatter cache reachable from `date()`** — `DateUtils.java` L49 `CACHED_FORMATTERS` is an unbounded static `ConcurrentHashMap`; `SQLFunctionDate` passes the caller's format straight into `getFormatter` → `computeIfAbsent`, so distinct patterns grow it without bound. Fix: size-bounded/LRU, or only cache schema-derived formats.

### Suggested fix
Add the per-item guards above; for the crash cases, wrap in a typed `CommandExecutionException`/`CommandSQLParsingException` rather than leaking the JDK exception.

### Scope / confidence
High confidence on each (read at HEAD). All are crash-on-valid-input or silent-wrong-arg, not degenerate-only.

### LINKED ISSUE #6389: Several SQL collection/string/misc functions throw raw runtime exceptions on valid input (missing null-guards / unchecked casts)
### Summary
A batch of SQL collection/string/misc methods and functions throw raw, uncaught runtime exceptions (missing null-guards, unchecked casts, or the wrong exception type) on ordinary valid input, surfacing as 500s instead of a clean typed error or a null. Grouped since they share the same shape; each is independent.

### Findings (all HEAD `ca89f56`)
- **`lastIndexOf()` no null guard** — `query/sql/method/string/SQLMethodLastIndexOf.java` L39-40 `value.toString().lastIndexOf(...)`; sibling `SQLMethodIndexOf.java` guards `value != null ? ... : null`. `SELECT nullField.lastindexof('a')` → NPE. (Also `params[0].toString()` and `Integer.parseInt(params[1])` unguarded.)
- **`join()` on a list with a null element** — `.../method/collection/SQLMethodJoin.java` L58-60 `.map(Object::toString)`. `SELECT [1,null,3].join('-')` → NPE. Fix: `.map(String::valueOf)`.
- **`map()` non-String key** — `function/sql/coll/SQLFunctionMap.java` L65 `final String key = (String) params[i];`. `SELECT map(1,'a')` → ClassCastException.
- **`asMap()` non-String key** — `.../method/conversion/SQLMethodAsMap.java` L70 `map.put((String) key, ...)`. `SELECT [1,2,3,4].asMap()` → ClassCastException.
- **`convert('<unknown>')` dead null-guard** — `.../method/conversion/SQLMethodConvert.java` L63-64 `Type arcadeType = Type.valueOf(...); if (arcadeType != null)`. `Type` is an enum; `valueOf` throws `IllegalArgumentException` for an unknown name and never returns null. `SELECT "x".convert('nope')` → raw IAE. Fix: try/catch → `CommandSQLParsingException`.
- **`decode(<malformed>,'base64')`** — `function/sql/misc/SQLFunctionDecode.java` L62/65 → `java.util.Base64.Decoder.decode` throws `IllegalArgumentException`. `SELECT decode('!!!','base64')`. (`decode(null,'base64')` NPEs at L58.) Fix: catch/wrap, null-guard (mirror `encode`).
- **`format(...)` (function `function/sql/text/SQLFunctionFormat.java` L48 and method `.../method/string/SQLMethodFormat.java`)** — passes a user string to `java.util.Formatter`: `format('%d','x')` → IllegalFormatConversionException, `format('%y')` → UnknownFormatConversionException, `format(null,'x')` → NPE; **resource:** `format('%99999999s','x')` → huge allocation. Fix: try/catch → typed error; consider capping width.
- **`max()`/`min()` on mixed String+Number** — `function/sql/math/SQLFunctionMax.java` L58/68, `SQLFunctionMin.java` L59/69 `((Comparable) subitem).compareTo(max)`. `SELECT max([1,'a'])` → ClassCastException. Fix: guard mutual comparability.
- **`bool_and`/`bool_or`** — `function/sql/misc/SQLFunctionBoolAnd.java` L47/56 (same in `SQLFunctionBoolOr`): `(Boolean)` casts → `SELECT bool_and([1,2,3])` / `bool_and(1,2)` → CCE; a single non-boolean scalar `bool_and(5)` falls through and returns the accumulator (silent wrong `true`). Fix: type-check before cast.
- **substring/left/right/charAt decimal index** — `.../method/string/SQLMethodSubString.java` L51-52/69 (and `SQLMethodLeft`/`Right`/`CharAt`) `Integer.parseInt(...)`; the #5885 clamp runs after the parse, so `SELECT "abcdef".substring(2.5)` → NumberFormatException on the decimal literal. Fix: parse leniently / clamp before parse. (The over-int-range sub-case may overlap the narrowing family; the decimal-literal case is distinct.)

### Suggested fix
Per-item: add the missing null-guard / comparability check, or wrap the JDK exception in a `Command*Exception`. Several have a sibling that already does it right (`indexOf`, `encode`, `sum`), so the pattern exists.

### Scope / confidence
High confidence — each read at HEAD with a concrete repro. All are crash-on-valid-looking input or a silent wrong result, not degenerate-only.

### LINKED ISSUE #6390: Aggregate siblings of sum() still throw ClassCastException on non-numeric input (#5799 fix not propagated)
### Summary
`SQLFunctionSum` was hardened to reject non-numeric input with a clean typed error (`requireNumericOrNull`, issue #5799), but its aggregate/window siblings were not — they still do a raw `(Number)` cast and throw `ClassCastException` on a non-numeric element.

### Where (HEAD `ca89f56`)
Contrast the fixed one: `function/sql/math/SQLFunctionSum.java` L64/69/77 use `requireNumericOrNull(...)` (L99), citing #5799.

Still-raw siblings (each does `(Number) n` / `addValue((Number) n)`):
- `function/sql/math/SQLFunctionAverage.java` L59 (multivalue) / L69.
- `function/sql/math/SQLFunctionVariance.java` L95 — inherited by `stddev`/`stddevp`/`variancep`, and by `median` via percentile.
- `function/sql/math/SQLFunctionPercentile.java` L67.
- Time-series: `SQLFunctionCorrelate.java` L61-62, `SQLFunctionRate.java` L75, `SQLFunctionDelta.java` L58, `SQLFunctionTsPercentile.java` L67/73, `SQLFunctionMovingAvg.java` L75.

### Repro
`SELECT avg(['a', 1])`, `SELECT variance(['a'])`, `SELECT ts.correlate('a','b')` → `ClassCastException` (String cannot be cast to Number), surfacing as a 500 rather than the typed error `sum` now gives.

### Suggested fix
Reuse the `requireNumericOrNull` pattern from `SQLFunctionSum` in each sibling.

### Scope / confidence
High confidence they throw (read at HEAD). Scope note: please confirm this is not considered covered by the #5799 family — `sum` got the fix, the siblings did not, so it is a distinct residual gap.

### LINKED ISSUE #6393: SQL: the null-coalescing operator `??` always returns its right operand, because the AST builder has no visitor for it
The SQL null-coalescing operator `??` always returns its RIGHT operand, whatever the left one is. It is not a
rendering problem: the operator has no visitor at all in the ANTLR AST builder, so ANTLR's default
`visitChildren` returns the last child it visited, and the left operand is dropped from the tree before anything
can evaluate it.

Found while answering a review question on PR #6364 about which compound shapes `Expression.toString()` can drop
parentheses from. Unrelated to that PR - measured on its branch, but nothing there touches this path, and the
shape below has no parentheses in it at all.

## Measured

On `26.9.1-SNAPSHOT`:

```
SELECT 'left' ?? 'right' AS n FROM V   ->  right
SELECT null   ?? 'right' AS n FROM V   ->  right
SELECT 1      ?? 2       AS n FROM V   ->  2
```

The first is the wrong answer: `'left'` is not null, so `??` must return `'left'`. The second is right only by
coincidence - the operator returns the right operand unconditionally, so it happens to agree whenever the left
one IS null.

Rendering shows the same loss, which is how it was noticed:

```
SELECT a ?? b || c AS n   renders as   SELECT b || c AS n
SELECT (a ?? b) AS n      renders as   SELECT b AS n
```

`a ??` is gone from the rendering because it is gone from the AST.

## Cause

`engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLParser.g4` declares the alternative:

```
expression
    : expression SC_OR expression        # arrayConcat
    | expression NULL_COALESCING expression  # nullCoalescing
    | mathExpression                     # mathExpr
    ...
```

`SQLASTBuilder` has `visitArrayConcat` for the sibling alternative but no `visitNullCoalescing` - grep the class
for `NullCoalescing` and there are no hits. Without an override, `SQLParserBaseVisitor.visitChildren` runs and
returns the result of the LAST child, which is the right-hand expression.

`MathExpression.Operator` already has a `NULL_COALESCING(25)` constant with an `apply` implementation, so the
evaluation half exists and is unused; what is missing is the visitor that builds a node using it.

## Why it matters

`??` is documented and parses without error, so a query using it is silently wrong rather than rejected - the
worst shape for a defect. A projection, a `WHERE` term or a `SET` value written `a ?? fallback` evaluates to
`fallback` even when `a` has a perfectly good value, so it reads as "the fallback is always taken".

## Suggested fix

Add `visitNullCoalescing` to `SQLASTBuilder`, building the two operands into a `MathExpression` with
`Operator.NULL_COALESCING`, mirroring `visitArrayConcat` next to it. Worth checking `visitArrayConcat`'s own
precedence handling at the same time, since the two alternatives sit at the same grammar level.

Regression tests should cover the value (`'left' ?? 'right'` is `'left'`, `null ?? 'right'` is `'right'`), the
round trip of the rendering, and the interaction with `||` at the neighbouring precedence level.

lvca:
Six independent defects that share one shape: valid input answered with something that was not an answer.

Closes #6382, closes #6385, closes #6388, closes #6389, closes #6390, closes #6393.

## #6393 - `??` always returned its right operand

The grammar declares the alternative, `MathExpression.Operator` carries a `NULL_COALESCING` with a working `apply()`, and nothing in between built the node: with no `visitNullCoalescing`, ANTLR's default `visitChildren` returns the LAST child it visited, so the left operand was gone from the tree before anything could evaluate it.

```
SELECT 'left' ?? 'right'   ->  right     (wrong)
SELECT null   ?? 'right'   ->  right     (right, but only by coincidence)
SELECT a ?? b || c         renders as    SELECT b || c
```

The node is now built two-operand and left-nested, which is also what lets `??` short-circuit: the right operand - possibly a function call or a sub-query - is not evaluated when the fallback is not taken. `MathExpression.toString()` had no case for the operator either, so the rendering lost it too; the regression test round-trips `??` against `||` and arithmetic at the neighbouring precedence levels.

## #6385 - the A* heuristic measured from the source

The N-axis branch filled the "current position" map with the **source** vertex's coordinate; the dead local holding the node's own coordinate was the tell. `h(n)` was therefore constant over the whole search - an admissible heuristic that guides nothing - and `dijkstra`, which delegates to A*, degraded the same way. The two-axis branch was already correct, so the test asserts the two agree on the same points, and that `h` decreases as a node approaches the goal.

## #6382 - CONTAINSTEXT split its literal on `:`

The direct full-text lookup split every whitespace-separated part on the first `':'` and looked the remainder up as a field-qualified key. A single-property index stores unprefixed tokens only, so an ordinary literal carrying a colon - a time, a timestamp, a `ns:name` - asked for a key the corpus never had and matched nothing.

A colon now introduces a qualifier only when the text before it **names an indexed property**, and on a single-property index it is dropped even then, mirroring the `FullTextQueryExecutor.isUnqualified` normalization the Lucene-backed executor already had and this path did not. Everything else is literal text handed to the analyzer, which is what `CONTAINSTEXT` documents its argument to be.

## #6388 - time/date functions

| | before | now |
|---|---|---|
| `sysdate('America/New_York')` | zone read from `params[1]`, so the documented one-argument form dropped it and answered server-local time | applies the zone; an unknown zone is a typed error |
| `duration(1,'week')` | `UnsupportedTemporalTypeException` | weeks have an exact length (7 days) and convert; years/months do not and are refused with the reason |
| `ts.timeBucket('0s', ts)` | `ArithmeticException: / by zero` | typed error; a null argument is null |
| `ts.lag(v,-1,ts)` / `ts.lead` | `IndexOutOfBoundsException` | offset must be non-negative |
| `ts.lag(value)` | NPE sorting on the optional timestamp | rows without one keep their arrival order (stable sort) |
| `date(v,'<bad pattern>')` / bad zone | raw `IllegalArgumentException` / `ZoneRulesException` escaped the `DateTimeParseException` catch | typed errors, deliberately **not** folded into the documented `null`: a value that does not match its format is a miss, a format that cannot exist is a query to fix |
| `DateUtils` formatter cache | unbounded, and reachable from `date()` with a caller-supplied pattern | bounded; past the ceiling a formatter is still built, just not remembered |

## #6389 - collection/string/misc functions

Same shape throughout: `lastIndexOf` lacked the null guard its `indexOf` sibling has; `join` mapped with `Object::toString` over a list that may hold null; `map`/`asMap` cast the key; `convert` guarded a `Type.valueOf` that throws rather than returning null; `decode` and both `format`s let `java.util.Formatter`/Base64 exceptions through (and `format` would allocate whatever width was asked for, so there is now a ceiling); `max`/`min` compared mutually incomparable values; `bool_and`/`bool_or` answered a confident `true` for a scalar they never looked at; and the substring family parsed a decimal index with `Integer.parseInt`, so the #5885 clamps could never run - the parse threw first.

## #6390 - the aggregate siblings of `sum()`

`sum()` was hardened to reject non-numeric input in #5799; every aggregate and window sibling kept the raw `(Number)` cast, so `avg`, `variance`/`stddev`/`stddevp`/`variancep`/`median`, `percentile` and the time-series aggregates (`ts.correlate`, `ts.rate`, `ts.delta`, `ts.percentile`, `ts.movingAvg`) answered a ClassCastException where `sum` answered a typed error. The guard moved up to `SQLFunctionAbstract` - along with the boolean and comparability guards the two issues above needed - and `sum` now uses the shared one rather than its own copy. A null is still skipped, which is the documented aggregation behaviour.

## Two existing tests that were passing for the wrong reason

`SQLFunctionAdditionalCoverageTest.sysdateWithFormat` and `DocumentValidationTest.defaultValueIsSetWithSQL` passed `sysdate('yyyy-MM-dd')` and `sysdate('YYYY-MM-DD HH:MM:SS')` as if the argument were a format. Neither was - the argument was ignored, and both assertions held whatever was passed. They now say what they check.

## Verification

Six new regression test classes (49 tests), each verified to fail against the unfixed sources before the change:

- `Issue6393NullCoalescingTest`, `Issue6385AstarHeuristicPositionTest`, `Issue6382ContainsTextColonTest`
- `Issue6388TimeFunctionArgumentTest`, `Issue6389FunctionArgumentTest`, `Issue6390AggregateNumericInputTest`

`mvn -o -pl server -am test -DexcludedGroups=benchmark,slow,vector`: BUILD SUCCESS - engine 12,803, network 450, ha 150, server 794, zero failures.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3513,15 +3513,26 @@ never looked at. `sum()` was hardened against non-numeric input in #5799 and its
 `sum` answered cleanly; the guard now lives on `SQLFunctionAbstract` and they all share it. Nulls are still
 skipped, which is the documented aggregation behaviour.
 
+**Behaviour change.** `map()` refuses a non-STRING key, `map(null, 'a')` included, where the raw cast used to
+store `(String) null` and build a map with a key nobody can reference. This is deliberately stricter than the
+`[...].asMap()` method, which converts what it is handed: `map()` takes key/value pairs the caller wrote out one
+by one, so a non-string among them is a mistake in the query worth reporting, while `asMap()` is documented as
+turning an arbitrary list *into* a map.
+
 **Breaking change.** `sysdate()` takes a **zone id** as its only argument, per its documented syntax
 `sysdate([<zoneid>])`, and now applies it: it used to read the zone from the *second* argument, so the
 one-argument form silently dropped it and answered server-local time. A second argument is now refused rather
 than accepted and ignored. Formatting is `.format()`'s job - `sysdate().format('yyyy-MM-dd')` - so a call
 written `sysdate('yyyy-MM-dd')`, which never formatted anything, is now an error naming the unknown zone.
 
-Two resource limits ride along, both reachable from caller-supplied text: the date formatter cache is bounded
-(past the ceiling a formatter is still built, just not remembered), and `format()` refuses a field width over
-a million characters instead of allocating it.
+Three resource limits ride along, all reachable from caller-supplied text: the date formatter cache is bounded
+(past the ceiling a formatter is still built, just not remembered), `format()` refuses a field width over a
+million characters instead of allocating it, and a character index is parsed from a bounded literal rather than
+whatever length the query text carries.
+
+One adjacent defect surfaced with the A* fix and is repaired with it: of the five heuristics on three or more
+axes, `EUCLIDEAN` was the only one that dropped `dFactor`, which nobody could see while every `h(n)` was the
+same number. It now scales like its two-axis twin and like the other four.
 
 [#6382](https://github.com/ArcadeData/arcadedb/issues/6382)
 [#6385](https://github.com/ArcadeData/arcadedb/issues/6385)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3412,3 +3412,56 @@ change the meaning - around a compound arithmetic operand - and not where they a
 `SELECT (name) FROM V` keeps the column name it has always had.
 
 [#6359](https://github.com/ArcadeData/arcadedb/issues/6359)
+
+## The `??` operator returns its left operand again, and a mistake in a function call is answered as one (#6382, #6385, #6388, #6389, #6390, #6393)
+
+The SQL null-coalescing operator `??` always returned its **right** operand, whatever the left one was. The
+grammar declares the alternative and `MathExpression.Operator` carries a working `NULL_COALESCING`, but nothing
+in between built the node: with no visitor for it, ANTLR's default `visitChildren` returns the last child it
+visited, so the left operand was gone from the tree before anything could evaluate it. `'left' ?? 'right'` was
+`'right'`, and `null ?? 'right'` agreed only by coincidence - which is the worst shape for a defect, because a
+projection, a `WHERE` term or a `SET` value written `a ?? fallback` read as "the fallback is always taken". It
+now evaluates left-to-right and short-circuits: the right operand, possibly a function call or a sub-query, is
+not evaluated when the fallback is not taken. `toString()` had no case for the operator either, so the
+rendering lost it too.
+
+The A* heuristic stored the **source** vertex's coordinate where the current node's belongs, for three or more
+axes (the two-axis branch was already right). `h(n)` was therefore the same number for every node - an
+admissible heuristic that guides nothing, so the search did the work of a plain Dijkstra while claiming to be
+informed. `dijkstra` delegates to A*, so any call supplying axis coordinates degraded with it.
+
+`CONTAINSTEXT` on a full-text index split its literal on the first `:` and looked the remainder up as a
+field-qualified key. A single-property index stores unprefixed tokens only, so an ordinary value carrying a
+colon - a time, a timestamp, a `ns:name` - asked for a key the corpus never had and matched **nothing**. A
+colon now introduces a qualifier only when the text before it names an indexed property, and on a
+single-property index it is dropped even then, matching the normalization the Lucene-backed path already
+applied; everything else is literal text handed to the analyzer, which is what the operator documents its
+argument to be.
+
+The rest is one theme across three issues: a raw JDK exception - an HTTP 500 - for ordinary valid input, where
+a typed argument error (HTTP 400) is the answer. `duration(1,'year')`, `ts.timeBucket('0s', ts)`,
+`ts.lag(v,-1,ts)`, `ts.lag(value)` without a timestamp, `date()` with a malformed pattern or unknown zone,
+`[1,null,3].join('-')`, `map(1,'a')`, `[1,2,3,4].asMap()`, `'x'.convert('nope')`, `decode('!!!','base64')`,
+`format('%d','x')`, `max([1,'a'])`, `bool_and([1,2,3])`, `nullField.lastindexof('a')` and
+`'abcdef'.substring(2.5)` each threw, and `bool_and(5)` was worse: it returned a confident `true` for input it
+never looked at. `sum()` was hardened against non-numeric input in #5799 and its siblings were not, so `avg`,
+`variance`/`stddev`/`median`, `percentile` and the time-series aggregates answered a `ClassCastException` where
+`sum` answered cleanly; the guard now lives on `SQLFunctionAbstract` and they all share it. Nulls are still
+skipped, which is the documented aggregation behaviour.
+
+**Breaking change.** `sysdate()` takes a **zone id** as its only argument, per its documented syntax
+`sysdate([<zoneid>])`, and now applies it: it used to read the zone from the *second* argument, so the
+one-argument form silently dropped it and answered server-local time. A second argument is now refused rather
+than accepted and ignored. Formatting is `.format()`'s job - `sysdate().format('yyyy-MM-dd')` - so a call
+written `sysdate('yyyy-MM-dd')`, which never formatted anything, is now an error naming the unknown zone.
+
+Two resource limits ride along, both reachable from caller-supplied text: the date formatter cache is bounded
+(past the ceiling a formatter is still built, just not remembered), and `format()` refuses a field width over
+a million characters instead of allocating it.
+
+[#6382](https://github.com/ArcadeData/arcadedb/issues/6382)
+[#6385](https://github.com/ArcadeData/arcadedb/issues/6385)
+[#6388](https://github.com/ArcadeData/arcadedb/issues/6388)
+[#6389](https://github.com/ArcadeData/arcadedb/issues/6389)
+[#6390](https://github.com/ArcadeData/arcadedb/issues/6390)
+[#6393](https://github.com/ArcadeData/arcadedb/issues/6393)
--- a/engine/src/main/java/com/arcadedb/function/sql/SQLAggregatedFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/SQLAggregatedFunction.java
@@ -19,6 +19,9 @@
 package com.arcadedb.function.sql;
 
 import com.arcadedb.function.AggregatedFunction;
+import com.arcadedb.query.sql.executor.MultiValue;
+
+import java.util.function.Consumer;
 
 /**
  * Abstract base class for SQL aggregate functions (count, sum, avg, min, max, etc.).
@@ -40,6 +43,29 @@ import com.arcadedb.function.AggregatedFunction;
  * @see AggregatedFunction
  */
 public abstract class SQLAggregatedFunction extends SQLFunctionConfigurableAbstract implements AggregatedFunction {
+  /**
+   * Feeds one argument's worth of input to a numeric accumulator: a {@link Number} goes straight in, a collection is
+   * unrolled element by element, a null is skipped, and anything else is a client-facing type error.
+   * <p>
+   * That triage is the same three branches in every cross-row numeric aggregate, and #6390 exists precisely because
+   * it was copied per function - {@code sum()} was hardened in #5799 and the copies in {@code avg}, {@code variance}
+   * and {@code percentile} drifted away from it. It lives here so the next hardening reaches all of them.
+   *
+   * @param value       the argument to accumulate
+   * @param accumulator where each numeric (or null) value is sent
+   */
+  protected void accumulateNumeric(final Object value, final Consumer<Number> accumulator) {
+    if (value instanceof Number number)
+      accumulator.accept(number);
+    else if (MultiValue.isMultiValue(value))
+      for (final Object item : MultiValue.getMultiValueIterable(value))
+        accumulator.accept(requireNumericOrNull(item));
+    else
+      // A non-numeric, non-null, non-list value is a client-facing type error rather than a silently dropped one,
+      // which used to make an all-invalid input indistinguishable from an all-null one (#5799, #6390).
+      accumulator.accept(requireNumericOrNull(value));
+  }
+
 
   protected SQLAggregatedFunction(final String name) {
     super(name);
--- a/engine/src/main/java/com/arcadedb/function/sql/SQLFunctionAbstract.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/SQLFunctionAbstract.java
@@ -87,6 +87,93 @@ public abstract class SQLFunctionAbstract implements SQLFunction {
   }
 
   /**
+   * Validates a value a numeric function is about to accumulate: {@code null} passes through (nulls are skipped,
+   * matching the documented aggregation behavior), a {@link Number} passes through unchanged, and anything else
+   * (STRING, BOOLEAN, a collection, a record, ...) is a client-facing type error naming the function and the type
+   * it got.
+   * <p>
+   * Issue #5799 introduced this for {@code sum()} so a non-numeric element stopped being silently dropped - which
+   * left the accumulator unchanged and made an all-invalid input indistinguishable from an all-null one. Issue
+   * #6390 pulled it up here because every sibling still did a raw {@code (Number)} cast and answered a
+   * ClassCastException (HTTP 500) where {@code sum()} answered a typed error (HTTP 400).
+   *
+   * @param value the value to validate
+   *
+   * @return the value as a {@link Number}, or {@code null}
+   *
+   * @throws IllegalArgumentException if the value is neither null nor a number
+   */
+  protected Number requireNumericOrNull(final Object value) {
+    if (value == null || value instanceof Number)
+      return (Number) value;
+    throw new IllegalArgumentException(
+        getName() + "() requires numeric input, but received a value of type " + value.getClass().getSimpleName());
+  }
+
+  /**
+   * Orders two values the way {@code max()} / {@code min()} need them ordered, answering a typed argument error when
+   * they are not mutually comparable instead of the ClassCastException the raw {@code ((Comparable) a).compareTo(b)}
+   * threw for something as ordinary as {@code max([1, 'a'])} (issue #6389). Numeric widening is the caller's job -
+   * both functions run {@code Type.castComparableNumber} first, so this only ever sees values already brought to a
+   * common numeric class, or values of genuinely different kinds.
+   *
+   * @param left  the value to place
+   * @param right the value to place it against
+   *
+   * @return the sign of the comparison, as {@link Comparable#compareTo}
+   *
+   * @throws IllegalArgumentException if the two values cannot be compared with each other
+   */
+  @SuppressWarnings("unchecked")
+  protected int compareValues(final Object left, final Object right) {
+    try {
+      return ((Comparable<Object>) left).compareTo(right);
+    } catch (final ClassCastException e) {
+      throw new IllegalArgumentException(
+          getName() + "() cannot compare a value of type " + left.getClass().getSimpleName() + " with a value of type "
+              + right.getClass().getSimpleName(), e);
+    }
+  }
+
+  /**
+   * The boolean analogue of {@link #requireNumericOrNull(Object)}: {@code null} passes through, a {@link Boolean}
+   * passes through, anything else is a client-facing type error rather than a ClassCastException - or, worse, a
+   * silently unchanged accumulator that returns a confident answer for input it never looked at (issue #6389).
+   *
+   * @param value the value to validate
+   *
+   * @return the value as a {@link Boolean}, or {@code null}
+   *
+   * @throws IllegalArgumentException if the value is neither null nor a boolean
+   */
+  protected Boolean requireBooleanOrNull(final Object value) {
+    if (value == null || value instanceof Boolean)
+      return (Boolean) value;
+    throw new IllegalArgumentException(
+        getName() + "() requires boolean input, but received a value of type " + value.getClass().getSimpleName());
+  }
+
+  /**
+   * The strict counterpart of {@link #requireNumericOrNull(Object)} for a scalar CONFIGURATION argument - a window
+   * size, a percentile, an offset - where {@code null} is not a value to skip but a missing setting.
+   *
+   * @param value        the argument value
+   * @param argumentName the argument's name, for the error message
+   *
+   * @return the value as a {@link Number}
+   *
+   * @throws IllegalArgumentException if the value is null or not a number
+   */
+  protected Number requireNumeric(final Object value, final String argumentName) {
+    if (value instanceof Number number)
+      return number;
+    throw new IllegalArgumentException(
+        getName() + "() requires a numeric <" + argumentName + ">, but received " + (value == null ?
+            "null" :
+            "a value of type " + value.getClass().getSimpleName()));
+  }
+
+  /**
    * Converts various input types (float[], double[], Object[], List) to a float array.
    * Delegates to {@link com.arcadedb.index.vector.VectorUtils#toFloatArray(Object)}.
    *
--- a/engine/src/main/java/com/arcadedb/function/sql/SQLFunctionAbstract.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/SQLFunctionAbstract.java
@@ -188,7 +188,13 @@ public abstract class SQLFunctionAbstract implements SQLFunction {
    * @throws IllegalArgumentException if the value is null or not a number
    */
   protected int requireIntArgument(final Object value, final String argumentName) {
-    return NumberUtils.saturateToInt(requireNumeric(value, argumentName));
+    final Integer number = NumberUtils.saturateToIntOrNull(value);
+    if (number != null)
+      return number;
+
+    throw new IllegalArgumentException(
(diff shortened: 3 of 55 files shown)


lvca:
Thanks - all four points addressed in 9fcaf90.

**1. `Math.multiplyExact` in the weeks branch** - fixed. You were right that it was the one new error path in this family still answering a bare `ArithmeticException`; it now raises the same typed error as every other bad argument here, with a test.

**2. `intValue()` truncating instead of saturating** - fixed, and this one earned its keep. `lag`/`lead`/`ts.movingAvg` now go through `SQLFunctionAbstract.requireIntArgument`, the `SQLFunction` twin of `AbstractSQLMethod.indexArgument`. Saturating then exposed a second defect underneath it: `lead` compared *and* indexed with `i + offset` as an int, which wraps **negative** near `Integer.MAX_VALUE`, so the wrapped value read as in-range and then indexed the list - `IndexOutOfBoundsException: Index -2147483648 out of bounds for length 3`. The arithmetic is widened to a long. `lag` cannot hit it: `i - offset` only ever goes further out of range. Regression test added for both.

**3. Case-sensitive `propertyNames.contains(...)`** - deliberate, leaving as is. `FullTextQueryExecutor.isUnqualified` compares with `.equals` on the same list, so exact case is what the path this fix is aligning with already does; matching case-insensitively here would make the two disagree again in the other direction. Property names are case-sensitive everywhere else in the schema, and the pre-fix behaviour for `Title:java` was a lookup for a key nothing stores - i.e. no matches - so nobody was relying on it working.

**4. `getMaxArgs()` tightening on `sysdate`** - agreed it deserves a release note, added to `docs/release-26.9.1.md` and called out as a breaking change: `sysdate()` takes a zone id as its only argument and now applies it, where it used to read the zone from the second argument and silently drop the documented one-argument form. The note also points at `.format()` for anyone who wrote `sysdate('yyyy-MM-dd')` expecting formatting - that never formatted anything, and is now an error naming the unknown zone. The whole batch is documented there.

Re-ran the affected suites after the changes: 3,994 tests across `function`, `query.sql` and `index.fulltext`, zero failures.

lvca:
Good catch on the `format()` width ceiling - both bypasses are real, and I confirmed the new test fails against the previous scanner before fixing it. Fixed in 4d6e334.

**Width discarded behind a precision.** You are right that the reset at `'.'` meant the final check measured the *precision*. The scanner now tracks the largest number the specifier names, so `format('%2000000.1s','x')` - truncate to one character, pad back out to two million - is refused. The width and the precision both size an allocation, so measuring the max is the right rule rather than picking one.

**`<` not in the flag set.** Also right, and worse, since it needs no precision: `<` is a documented `java.util.Formatter` flag (reuse the previous argument), so the scanner mistook it for the conversion character and stopped before reading any width. `format('%s%<2000000s','x')` bypassed the ceiling outright. The flag set is complete now and lives in a named `isFormatFlag` predicate instead of an inline negated chain - which is what made it easy to leave a flag out in the first place.

Your note that `$` should *not* fold into the measurement is right too, and is preserved: an argument index allocates nothing, so `%2000000$s` is still discarded rather than read as a field size. There is a test for that alongside the two bypasses, plus two positive cases (`%10.3s` and `%s%<s`) so the stricter scan cannot regress into refusing valid patterns.

On the `??` short-circuit invariant - agreed that "helpfully" flattening chained `??` like the `||` sibling would silently break it. The comment on `visitNullCoalescing` says so, and `Issue6393NullCoalescingTest.chainedCoalescingTakesTheFirstNonNull` plus the round-trip test would both fail if someone did.

Re-ran the affected suites after the change: 4,396 tests across `function`, `query.sql`, `index.fulltext` and `utility`, zero failures.

lvca:
Thanks - responding to the 17:27 review (I had replied to the 17:23 one before this landed).

**The `format()` width+precision bypass** was already fixed in 4d6e334, and your `%2000000000.1f` trace is exactly right: the reset at the `.` meant the final check measured the precision and the width was never looked at. The scanner now tracks the largest number the specifier names rather than clobbering the running total. Your suggested case is added verbatim alongside the string form, since a two-billion-character numeric field is an OutOfMemoryError rather than merely a large string.

That same commit also closed a second bypass in the same helper that needs no precision at all: `<` (reuse the previous argument) is a `java.util.Formatter` flag and was missing from the flag set, so the scanner mistook it for the conversion character and stopped before reading any width - `format('%s%<2000000s','x')` sailed straight through. The flag set is complete now and lives in a named `isFormatFlag` predicate.

**`SQLMethodConvert` dropping the cause** - fixed in 765487b. `CommandSQLParsingException` does take a cause; there was no reason to discard the stack trace.

**`SQLFunctionMap` null key** - the rejection is intentional, and now has a test saying so. The raw cast produced `(String) null` and stored a key nobody can name, so a mistyped call silently built a map with a null entry rather than reporting the mistake; it is refused with the same message as any other non-STRING key.

Re-ran after the changes: 1,886 tests across `function`, `query.sql.method` and `utility`, zero failures (4,396 on the wider sweep after the width fix).

lvca:
Thanks - one of the four is worth acting on, the other three are already settled. Fixed in 95e7ac7.

**3. `indexArgument`'s BigDecimal fallback (actioned).** Fair point: the fallback is what makes `substring(2.5)` work, but `BigDecimal` parses whatever length it is handed and both the parse and the magnitude comparison grow with it, all to reach a number that saturates to the same bound either way. A character index is ten digits and a sign, so a literal orders of magnitude longer than that is now refused by length before anything is allocated. `Integer.parseInt` was already bounded, so only the fallback needed it. Tested both ways - a 100,000-character literal is refused, and a padded-but-plausible `'00000000000000000002.5'` still saturates to 2 as before.

**1. Formatter cache check-then-`putIfAbsent` race.** Agreed, and deliberately so - the comment already says the overshoot is by intent and spells out why an LRU would be worse here (the working set is a handful of schema formats that all fit, so a lock on a hot path would only reorder entries nothing ever evicts). That is exactly the "don't "fix" this into a stricter version" note you were after, sitting where someone about to do it would read it.

**2. `parseQueryTerms` case sensitivity.** Sanity-checked as suggested: `FullTextQueryExecutor.isUnqualified` compares with `.equals` on the same property-name list, so the two paths agree on case - which is the whole point of the change, since they previously disagreed in the other direction. Property names are case-sensitive elsewhere in the schema too, and the pre-fix behaviour for `Title:java` was a lookup for a key nothing stores, i.e. no matches, so nothing was relying on it.

**4. Release notes.** Noted, thank you - the `sysdate()` entry is explicitly marked a breaking change and points at `.format()` for anyone who had written `sysdate('yyyy-MM-dd')` expecting formatting.

Also confirming your read on `ts.lag` - it is left un-widened on purpose for exactly the reason you traced, and the comment in `SQLFunctionLead` says so, so the asymmetry does not look like an oversight to the next reader.

lvca:
Two reviews landed together; addressing both in 03e3878.

### The one claimed blocker is not one

`UpdateStatementExecutionTest.updateWithApplyDefaultsAndDynamicDefault` does **not** break. The default there is written `default \"sysdate('YYYY-MM-DD')\"` - the escaped quotes make it a double-quoted SQL **string literal**, so the text is stored verbatim and `sysdate` is never invoked, `ZoneId.of` never reached. Verified directly (`mvn -Dtest='UpdateStatementExecutionTest#updateWithApplyDefaultsAndDynamicDefault'` passes), and the full engine lane is green at 12,831 tests. The two tests this PR did change called `sysdate(...)` unquoted, which is the difference.

### Acted on

**Short-circuit was asserted but never tested** - fair, and it is a behaviour rather than an optimisation, so it now has a test. Observed with an argument that throws when it *is* evaluated, in both directions: `SELECT 1 ?? 'abcdef'.substring('boom')` returns 1, and `SELECT null ?? 'abcdef'.substring('boom')` throws. The second half matters - without it the first could pass by the operand being dead rather than skipped.

**A* `EUCLIDEAN` ignoring `dFactor` on 3+ axes** - confirmed, and fixed here rather than deferred: it is one line, it is the only one of the five N-axis formulas that dropped it, and #6385 is precisely what makes it observable, so it belongs with that fix rather than after it. The axis-agreement test now runs at a non-default `dFactor` as well as 1.0 - at 1.0 alone it could never have caught this.

**Three siblings still holding the raw cast** - swept in: `pow()`'s exponent, `ts.rank`'s `Comparable` sort (given the full `ts.lag`/`ts.lead` treatment, null-timestamp ordering included) and `ts.interpolate`'s linear branch. Leaving three known instances of the exact bug class the PR is about seemed worse than the extra diff.

**Duplication** - both points taken:
- The Number/MultiValue/scalar triage moved to `SQLAggregatedFunction.accumulateNumeric`, so `avg`/`variance`/`percentile` share it. That is #6390's own moral - copying it per function is how they drifted from `sum()` to begin with.
- `indexArgument` and `requireIntArgument` now delegate to one `NumberUtils.saturateToIntOrNull`. They had already disagreed: `pow(2,'10')` was refused where `'abc'.substring('2')` was not. The length bound moved in with it, and the rejection message summarises an over-long literal by its length rather than echoing 100,000 characters into a log.

**`map()` vs `asMap()` divergence, and the null-key change** - both now documented, in `SQLFunctionMap`'s Javadoc and as its own release-note paragraph. The divergence is intentional: `map()` takes pairs the caller wrote out one by one, so a non-string among them is a mistake worth reporting; `asMap()` is documented as turning an arbitrary list *into* a map and converts what it is handed.

**`parseQueryTerms` case sensitivity** (raised by both) - checked against the Lucene path as suggested: `FullTextQueryExecutor.isUnqualified` compares with `.equals`, so the two agree on case, which is the point of the change.

Verification: `mvn -o -pl server -am test` - engine 12,831, network 450, ha 150, server 794, zero failures.

Issue #6403, opened 2026-08-18
The coll.* family never got the argument-coercion and error-classification passes the Cypher list builtins got: arrays are refused and a client type error is a 500


lvca:
Found while making the `coll.*` functions answer a lazy `range()` without copying it (#6353, PR #6361). Working through that family put every one of them side by side with its `tail()`/`reverse()`/`size()` counterpart, and the two sets have drifted apart on everything *except* the range fix.

The cause is one method. `AbstractCollFunction.asList` (`engine/src/main/java/com/arcadedb/function/coll/AbstractCollFunction.java`) is the coercion for the whole `coll.*` namespace:

```java
protected List<Object> asList(final Object arg) {
  if (arg == null) return null;
  if (arg instanceof List) return (List<Object>) arg;
  if (arg instanceof Collection) return new ArrayList<>((Collection<Object>) arg);
  throw new CommandExecutionException(getName() + "() requires a list argument, got: " + arg.getClass().getSimpleName());
}
```

while the builtins go through `CypherFunctionHelper.requireListArgument` / `asListOrNull`, which coerce via `MultiValue.getMultiValueAsList` (arrays included, issue #4284), also accept `Iterable`/`Iterator`, and raise `CommandSemanticException` on a mismatch (issues #5476/#5477).

## 1. An array parameter works for the builtins and is refused by `coll.*`

Verified against `90b095900` (current main) with `$arr` bound to a `long[]`:

| query | result |
|---|---|
| `RETURN reverse($arr)` | `[2, 1, 3]` |
| `RETURN tail($arr)` | `[1, 2]` |
| `RETURN size($arr)` | `3` |
| `RETURN coll.sort($arr)` | **`CommandExecutionException: coll.sort() requires a list argument, got: long[]`** |
| `RETURN coll.distinct($arr)` | **same** |
| `RETURN coll.union($arr, $arr)` | **same** |

#4284 established that a numeric-array parameter is a Cypher LIST. It reached the builtins and the `+` operator; it never reached this family.

## 2. A client type error is reported as a server fault

`CommandExecutionException` is mapped to **HTTP 500** by `AbstractServerHttpHandler`; `CommandSemanticException` extends `CommandParsingException`, which is mapped to **400**. So over HTTP:

| query | status |
|---|---|
| `RETURN tail(42)` | 400, `Type mismatch: tail() expects a LIST<ANY> argument but got INTEGER` |
[...]

Issue #6402, opened 2026-08-18
LOAD CSV: file() and linenumber() are null inside a WHERE predicate, and two predicate forms disagree with each other in the same query


lvca:
Found while doing the `ExpressionEvaluator` inventory for #6354 (PR #6361). That issue consolidated the arithmetic semantics that were written twice - once in the AST node, once in `ExpressionEvaluator` - because a fix could land on one path and not the other. `evaluateFunction` was read with the same question and judged *not* a semantic duplicate, which is still true: the function semantics live in the `StatelessFunction`. But it does carry one thing the AST path does not, and that turns out to matter.

## The propagation only exists on one path

`ExpressionEvaluator.evaluateFunction` (`engine/src/main/java/com/arcadedb/query/opencypher/executor/ExpressionEvaluator.java`) lifts the LOAD CSV row context into the `CommandContext` before calling the function:

```java
// Propagate LOAD CSV context from current row to context for file()/linenumber()
if (result != null && result.hasProperty("__loadCSV_file")) {
  context.setVariable("__loadCSV_file", result.getProperty("__loadCSV_file"));
  context.setVariable("__loadCSV_linenumber", result.getProperty("__loadCSV_linenumber"));
}
```

`FunctionCallExpression.evaluate` - the path a WHERE predicate takes - has no equivalent. Projections are evaluated by `ExpressionEvaluator`, WHERE predicates by the AST node (measured by injecting a throw into each; see PR #6361).

## Reproduction

With `people.csv` holding a header plus two data rows:

```
name,age
alice,30
bob,40
```

| query | rows | expected |
|---|---|---|
| `LOAD CSV FROM '…/people.csv' AS row RETURN file() AS r` | 3, each the path | correct |
| `LOAD CSV FROM '…/people.csv' AS row RETURN linenumber() AS r` | 3: `1 2 3` | correct |
| `LOAD CSV FROM '…/people.csv' AS row WITH row WHERE file() IS NOT NULL RETURN 1 AS r` | **0** | 3 |
| `LOAD CSV FROM '…/people.csv' AS row WITH row WHERE linenumber() IS NOT NULL RETURN 1 AS r` | **0** | 3 |
| `LOAD CSV FROM '…/people.csv' AS row WITH row WHERE linenumber() > 1 RETURN 1 AS r` | **2** | 2 |

Verified against `90b095900` (current main) in a fresh database, with the predicate queries run *first* so nothing an earlier projection put in the context could serve them.

[...]

Issue #6400, opened 2026-08-18
Cypher: nothing detects the two plan builders disagreeing, and which one runs is decided by unrelated syntax


lvca:
Two findings from #6310 / PR #6369, filed together because they share a cause: the native Cypher engine has two independent planners, and which one runs a given clause is decided by syntax that has nothing to do with the clause's meaning.

- **Item 1** is new and is the one worth acting on.
- **Item 2** turns out to be already tracked as #5358; recorded here only because #6310 produced fresh evidence for it and because it is what makes Item 1 bite.

---

### What happened

#6310 existed because `CypherExecutionPlan.buildMatchStep` (the traditional plan) and `CypherOptimizer` (the cost-based physical plan) disagreed about relationship uniqueness across a comma, and nothing noticed. The same clause returned different result sets depending on which planner got it:

```cypher
MATCH (n0:A)<-[r1]-(:E), (n3:C)-[]->(:E)-[r2]->(n0) RETURN n0, n3
```

| spelling | planner | before #6369 |
|---|---|---|
| as written | cost-based | 0 rows (correct) |
| inside `CALL { ... }` | traditional | **4 rows** |
| `OPTIONAL MATCH ...` | traditional | **4 rows** |
| inside `EXISTS { ... }` | traditional | **3 rows** |
| one part carrying a named path `p = ...` | traditional | **4 rows** |

Every one of those is the same MATCH clause. Nothing in the added syntax changes what the pattern means - it only changes which planner the query lands on. A user hitting this has no way to tell that wrapping a query in `CALL {}` silently changed its semantics, and neither did we: the defect survived because no test ever compared the two planners' answers to the same pattern.

The 3897-scenario openCypher TCK does not catch this either. It exercises whichever planner each scenario happens to route to, so a divergence is only caught if some scenario happens to be written in the shape that routes to the broken side.

### Suggested fix

A differential harness: take a corpus of MATCH clauses, run each through both planners against the same fixture, and assert the result sets are equal as multisets.

The routing is already forcible from the query text, so no engine hooks are needed - the same clause can be sent down either path:

- **cost-based**: the clause as written
- **traditional**: the same clause wrapped in `CALL { ... }`, or preceded by `OPTIONAL`, or with a named path bound on one part

and `EXPLAIN` reports which one actually ran (`Traditional Execution (Non-Optimized)` vs the physical operator tree), so the harness can assert it really exercised both rather than the same one twice - the failure mode that would make the whole harness vacuous.

[...]

Issue #6395, opened 2026-08-18
openCypher: a node whose only label is `V` or `Vertex` reports no labels, and a label write drops it - the #6363 invariant in the corner it left alone


lvca:
Found while verifying #6363 (merged as `7912cc75e`). Reproduces on that commit.

#6363 established the invariant `L IN labels(n)` iff `n:L`, and fixed the write side so that adding a label cannot take one away. Both halves still fail in one corner: when the label a node carries is literally `V` or `Vertex`.

The engine uses those two names for two different things at once - the type a node lands in when it has **no** labels, and an ordinary user label that a query is free to write. `Labels.getLabels` can only tell them apart by name, so it filters the name and the genuine label disappears with it.

## 1. The label is silently lost on read

```cypher
CREATE (:V {k:'onlyV'})
```

| query | returns | expected |
|---|---|---|
| `MATCH (n {k:'onlyV'}) RETURN labels(n)` | `[]` | `["V"]` |
| `MATCH (n:V) RETURN n.k` | `onlyV` | `onlyV` |
| `MATCH (n) WHERE n:V RETURN n.k` | `onlyV` | `onlyV` |

Same for `:Vertex`. The engine matches `:V` twice over and then reports a label list that does not contain it - the exact shape of #6363, in the case that fix deliberately left alone.

Note this only bites when `V` is the node's **whole** label set. `CREATE (:U:V:W)` builds the `U~V~W` composite, and the walk through its supertypes reports `[U, V, W]` correctly - which is why the openCypher TCK (`CREATE (b:U:V:W:X:Y:Z)`) passes.

## 2. Adding a label removes one

```cypher
MATCH (n {k:'onlyV'}) SET n:Extra
```

| after the SET | result | expected |
|---|---|---|
| `labels(n)` | `["Extra"]` | `["Extra","V"]` |
| `MATCH (n:V) RETURN count(n)` | **`0`** | `1` |

`Labels.getOwnLabels` answers `List.of()` for a base-named own type, so the new type is built from `[Extra]` alone and the node is moved out of `V`. This is #6363's write-side defect, surviving in this corner.

## 3. Two different types mean "no labels"

`CreateStep` puts an unlabelled node in `Vertex`; `RemoveStep` puts a node stripped of its last label in `V`:

```cypher
CREATE (:Foo {k:'stripped'}), ({k:'born-plain'})
MATCH (n:Foo) REMOVE n:Foo
```

| query | returns |
|---|---|
[...]

PR #6417, opened 2026-08-18
fix(#6403, #6402, #6400, #6395, #6375, #6370, #6362, #6317): eight small fixes across coll.*, LOAD CSV, planner, labels, algo budgets, comprehensions and subqueries

lvca:
## Summary

Eight small, independent bugs that came out of earlier analysis, fixed together:

- **#6403** - `coll.*` never got the argument-coercion/error-classification pass the Cypher list builtins got. `AbstractCollFunction.asList` now delegates to `CypherFunctionHelper.requireListArgument`, so the family accepts arrays/`Iterable`s like `tail()`/`reverse()` do, and raises `CommandSemanticException` (HTTP 400) instead of `CommandExecutionException` (HTTP 500) for a caller mistake. `coll.toSet` is now a one-line subclass of `coll.distinct` rather than a second copy of its body. `coll.min`/`max`/`sum`/`avg` answer a range from its two endpoints instead of walking it.
- **#6402** - `file()`/`linenumber()` after `LOAD CSV` answered correctly in a `RETURN` and `null` in a `WHERE`, because the row context was lifted onto the `CommandContext` by `ExpressionEvaluator` only, not by the AST path a predicate takes. Both now go through one shared invocation point (`FunctionCallExpression.invoke`), and the context now survives a `WITH row` that doesn't re-name it.
- **#6400** - added a differential harness (`CypherPlannerDifferentialIssue6400Test`) that runs a combinatorial MATCH corpus through both the cost-based and the traditional planner and asserts they agree, with three vacuity guards (routing actually differs, rows actually match something, the comparison itself discriminates) so the harness can't silently stop testing anything.
- **#6395** - "no labels" was represented by the ordinary label names `V`/`Vertex`, which a query could also write, so `Labels.getLabels` filtered them and a genuine `:V` label vanished. Reserved a sentinel (`~NO_LABEL~`, unwritable as a label because it contains `LABEL_SEPARATOR`) and filter it at every depth of the supertype walk rather than only at a vertex's own type. Found while investigating: the Neo4j importer's shared root type was literally named `"Node"`, which leaked as a phantom label onto every imported node independently of this issue - fixed by reusing the same reserved sentinel as that root, keeping the "one query reaches every imported vertex" property while making it invisible to `labels()`.
- **#6375** - five algorithms (`triangleCount`, `knn`, `kTruss`, `clique`, `hierarchicalClustering`) built `nodeCount x nodeCount` `BitSet` matrices without reserving against `arcadedb.cypher.algoMaxWorkingMemory`, even though the setting's own docs name exactly that shape. Now reserved before allocation, with the build loops made abortable.
- **#6370** - a bare `[*]` inside a pattern comprehension parsed as one fixed hop, because `CypherExpressionBuilder`'s copy of the path-length normalization never got the `[*] == [*1..]` fix `CypherASTBuilder`'s copy did. Lifted into one shared `ParserUtils.parsePathLength` helper both builders call.
- **#6362** - inserting an identity `CALL (*) { RETURN 0 AS barrier }` after `MERGE ... ON MATCH SET` changed the query's answer, because `SubqueryStep`'s post-#4182 binding refresh re-reads every `Document` binding from the transaction cache even for a read-only subquery, letting a later row's write show through an earlier row's binding. Now skipped when the inner statement is read-only.
- **#6317** - the graph an `algo.*` call loads (vertex list, RID index, adjacency) was the one allocation no budget priced. `loadGraph`/`loadVertices` now bound the vertex walk via `MemoryBudget.capacityFor`, and `adjacency()` prices the CSR/OLTP neighbour-list copy before it's built.

## Test plan

- [x] New regression test per issue (`CypherCollFunctionParityIssue6403Test`, `CypherLoadCSVRowContextIssue6402Test`, `CypherPlannerDifferentialIssue6400Test`, `CypherReservedNoLabelSentinelIssue6395Test`, `Issue6375AlgoBitsetMatrixBudgetTest`, `CypherBareStarComprehensionIssue6370Test`, `CypherSubqueryRowBindingIssue6362Test`, `Issue6317AlgoGraphLoadBudgetTest`)
- [x] Updated pre-existing tests whose expectations changed with the fix (exception type for `coll.remove`/`coll.toSet`/`coll.pairsMin`, the `V`/`Vertex` sentinel rename in `LabelsTest`/`OpenCypherCreateTest`, the algo working-memory-budget figures now that the graph itself is priced)
- [x] Full `engine` module suite green: 12,799 tests, 0 failures, 0 errors (excluding benchmark/vector/slow lanes)
- [x] `integration` module Neo4j importer tests green (15 tests, including the renamed root-type assertion)
- [x] Manually verified against the reproduction table in each issue

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/coll/AbstractCollFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/coll/AbstractCollFunction.java
@@ -18,11 +18,11 @@
  */
 package com.arcadedb.function.coll;
 
-import com.arcadedb.exception.CommandExecutionException;
+import com.arcadedb.exception.CommandSemanticException;
 import com.arcadedb.function.StatelessFunction;
+import com.arcadedb.function.cypher.CypherFunctionHelper;
+import com.arcadedb.utility.LongRangeList;
 
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 
 /**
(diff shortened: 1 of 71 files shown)


lvca:
Addressed the review findings:

1. **`Labels.ensureCompositeType` didn't reject a literal `~NO_LABEL~` label** - fixed. `CREATE (:\`~NO_LABEL~\`)`, `SET n:\`~NO_LABEL~\``, and `MERGE (n:\`~NO_LABEL~\`)` now all raise `CommandSemanticException` instead of landing the vertex on the sentinel type. Had to also fix `CreateStep`/`MergeStep`'s own internal "no label written" case, which intentionally targets that type and was going through the same method - it now bypasses `ensureCompositeType` entirely, matching what `RemoveStep`'s empty-label branch already did. New tests: `theSentinelItselfIsNotWritableAsALabelViaCreate/Set/Merge`.
2. **`SubqueryStep`'s read-only classification gap for `DEFINE FUNCTION` writes** - confirmed real (verified `SQLFunctionDefinition.execute` runs a full `sqlscript` command, which can write). Filed as #6418 rather than patched here: a correct fix needs the same whole-statement, every-expression-position walk `isReadOnly()` itself would need, and that flag also backs HA leader routing (#6094) - not something to touch under time pressure in an unrelated bug-fix batch. Documented the exact gap on `innerMayWrite`.
3. **Missing `@Tag("slow")` on the differential harness** - checked; actual measured runtime is 0.785s, well under this repo's "multi-second" threshold for the tag. Left untagged.
4. **`LoadCSVRowContext.carryOver` propagates unconditionally** - confirmed intentional (`file()`/`linenumber()` are a property of query position relative to `LOAD CSV`, not of the row variable surviving a projection - same as how `count(*)` needs no variable bound). Documented and pinned with `theRowContextSurvivesAProjectionThatDropsRowEntirely`.
5. **`CollFlatten`'s depth argument bypassed coercion** - fixed, now raises the same client `CommandSemanticException` the rest of the family does for a non-numeric, non-boolean depth.
6. **Algo budget checkpoint granularity in the CSR-no-`NeighborView` fallback** - documented as a known narrow limitation rather than fixed, per the review's own framing.
7. Codacy's one flag (field declared mid-class in `AlgoClique`) - fixed in the previous push.

Full engine suite re-verified green after all fixes (12,799+ tests).

lvca:
@claude please review updates

lvca:
Fixed the comment's wrong issue number (`#6395` → `#6362`, filed as `#6418`). Build-verified locally (compile + the five touched test classes, 60/60 green) since the review noted it couldn't run `mvn` in its environment.

lvca:
@claude review this again

PR #6419, opened 2026-08-18
fix(#6378): OPTIONAL MATCH with a single-node named path raises ZeroLengthPathStep error

### LINKED ISSUE #6378: ArcadeDB 26.8.1: OPTIONAL MATCH with a single-node path raises ZeroLengthPathStep error
## Environment

- ArcadeDB Docker image: arcadedata/arcadedb:26.8.1 
- Clean Test database

## Reproduction

`CREATE (:N {k: 1});
`
The control query succeeds:

```
MATCH p = ({k: 1})
RETURN p;
```

It returns one row.

The equivalent optional query fails:

```
OPTIONAL MATCH p = ({k: 1})
RETURN p;
```

## Expected result

The query should return one row containing the matching node, just like:

```
MATCH p = ({k: 1})
RETURN p;
```

## Actual result

The query fails with:

Neo.DatabaseError.General.UnknownError:
ZeroLengthPathStep requires a previous step

robfrank:
Closes #6378

## What does this PR do?

Fixes `OPTIONAL MATCH p = ({k: 1}) RETURN p` (a single-node pattern with a named
zero-length path) raising `ZeroLengthPathStep requires a previous step` instead of
returning the matching node, while the equivalent non-optional
`MATCH p = ({k: 1}) RETURN p` worked correctly.

**Root cause**: in `CypherExecutionPlan.buildMatchStep`, the single-node pattern
branch wired up a `ZeroLengthPathStep` for the path variable and then reassigned
`matchChainStart = zeroPathStep` whenever `matchChainStart` currently equaled the
`MatchNodeStep` it had just been set to - which is the common case for a leading
OPTIONAL MATCH. `OptionalMatchStep` is constructed as
`new OptionalMatchStep(matchChainStart, matchChainEnd, ...)`, where `matchChainStart`
is expected to be the head of the sub-chain (the step `OptionalMatchStep` calls
`setPrevious(...)` on to feed it each input row) and `matchChainEnd` the tail it pulls
results from. With the erroneous reassignment, `matchChainStart` and `matchChainEnd`
became the *same* `ZeroLengthPathStep`, so `OptionalMatchStep` overwrote the link that
had just been established from the `ZeroLengthPathStep` back to the `MatchNodeStep`.
For a standalone `OPTIONAL MATCH` that overwrite sets the previous step to `null`,
which is exactly the reported error.

The equivalent code path a bit further down in the same method (multi-node patterns
with zero relationships) already guards this correctly - it only assigns
`matchChainStart` the first time it is still `null`. This PR makes the single-node
branch match that pattern by removing the erroneous reassignment; `matchChainStart`
was already correctly set a few lines above.

## Motivation

Reported in #6378: a straightforward `OPTIONAL MATCH` over a single node with a named
path throws instead of returning a row (or a NULL row when nothing matches), which is
a basic correctness bug in the openCypher engine's OPTIONAL MATCH handling.

## Related issues

Closes #6378

## Additional Notes

- Verified with TDD: the new test class failed with the exact reported error against
  the unmodified code, and passes after the one-line fix.
- Ran the broader OPTIONAL MATCH / zero-length-path / variable-length-path / GAV
  regression surface (`OpenCypherZeroLengthPathTest`, `Issue5094OptionalMatchCountStarTest`,
  `Issue5790OptionalMatchCountStarUndeclaredRelTypeTest`,
  `CypherCaseOptionalMatchRelIssue5137Test`, `OpenCypherVariableLengthPathTest`,
  `GAVEligibilityTest`) - all green, no regressions.
- One unrelated observation while running the regression surface: `OpenCypherOptionalMatchTest`
  reports "Tests run: 0" both before and after this change, with or without a fully
  qualified `-Dtest` filter, even though the class compiles cleanly and its `.class`
  files are present in `target/test-classes`. This file is untouched by this PR and the
  same behavior reproduces on an unmodified checkout, so it looks like a pre-existing
  JUnit5 discovery quirk in this environment rather than something this change caused;
  flagging it here for a maintainer to look into separately rather than trying to fix
  it in this PR.

## Test plan

- [x] `mvn -pl engine -am test -Dtest=Issue6378OptionalMatchSingleNodeZeroLengthPathTest`
      - 4/4 new tests pass: the exact query from the issue, a no-match case, the
        control (non-optional) query, and a chained `MATCH ... OPTIONAL MATCH p = (n)`
        case covering the already-bound-variable branch.
- [x] `mvn -pl engine -am test -Dtest=OpenCypherZeroLengthPathTest,Issue5094OptionalMatchCountStarTest,Issue5790OptionalMatchCountStarUndeclaredRelTypeTest,CypherCaseOptionalMatchRelIssue5137Test,OpenCypherVariableLengthPathTest,GAVEligibilityTest`
      - all pass, no regressions in OPTIONAL MATCH / zero-length-path / GAV handling
- [ ] `mvn clean package` (left to CI / reviewer given the isolated build environment used here)


robfrank:
@claude review

Issue #1814, opened 2024-11-04
SQL: UPDATE not triggering schema default value


gramian:
### ArcadeDB Version:

`ArcadeDB Server v24.10.1 (build ac46a0da6ebe4f5d337e50874f3d8a0d1bf092b1/1728751149872/main)`

### OS and JDK Version:

`Running on Linux 6.2.0-26-generic - OpenJDK 64-Bit Server VM 17.0.12`

Given a type having a `mandatory`, `notnull` property with a `default` value, where a record is updated with the property in question being empty. Then the `default` value is not triggered and thus `notnull` and `mandatory` attributes trigger an error.

### Expected behavior

Update existing record with default value

### Actual behavior

```
Error on command execution (PostCommandHandler)
com.arcadedb.exception.ValidationException: The property 'doc.prop' is mandatory, but not found on record: #1:0@doc[]
	at com.arcadedb.database.DocumentValidator.throwValidationException(DocumentValidator.java:389)
	at com.arcadedb.database.DocumentValidator.validateField(DocumentValidator.java:43)
	at com.arcadedb.database.DocumentValidator.validate(DocumentValidator.java:38)
	at com.arcadedb.database.MutableDocument.validate(MutableDocument.java:134)
	at com.arcadedb.database.LocalDatabase.updateRecord(LocalDatabase.java:883)
	at com.arcadedb.database.MutableDocument.save(MutableDocument.java:297)
	at com.arcadedb.query.sql.executor.SaveElementStep$1.next(SaveElementStep.java:59)
	at com.arcadedb.query.sql.executor.CountStep.syncPull(CountStep.java:67)
	at com.arcadedb.query.sql.executor.SelectExecutionPlan.fetchNext(SelectExecutionPlan.java:44)
	at com.arcadedb.query.sql.executor.UpdateExecutionPlan.executeInternal(UpdateExecutionPlan.java:64)
	at com.arcadedb.query.sql.parser.UpdateStatement.execute(UpdateStatement.java:126)
	at com.arcadedb.query.sql.parser.Statement.execute(Statement.java:73)
	at com.arcadedb.query.sql.SQLQueryEngine.command(SQLQueryEngine.java:107)
	at com.arcadedb.database.LocalDatabase.command(LocalDatabase.java:1348)
	at com.arcadedb.server.ServerDatabase.command(ServerDatabase.java:472)
[...]

PR #6421, opened 2026-08-18
fix(#6414): a documented option is applied, and a full-text index answers the same question whatever its arity

### LINKED ISSUE #6414: Follow-ups from #6408: an ignored astar option, CONTAINSTEXT losing its index on multi-property, a vacuous default-value test, and a duplicated full-text rule
Five findings that surfaced while fixing #6382, #6385, #6388, #6389 and #6390 in #6408, none of which belonged in that PR's scope. They are independent of each other - each item below stands alone.

All references are against `2982f5504`.

---

## 1. `astar`/`dijkstra` accept `customHeuristicFormula` and silently ignore it

`PARAM_CUSTOM_HEURISTIC_FORMULA` is a documented option, is listed in `SQLFunctionAstar`'s accepted-options set (`SQLFunctionAstar.java` L64), and is read from the caller's map into a field (L339):

```java
context.paramCustomHeuristicFormula = opts.getString(PARAM_CUSTOM_HEURISTIC_FORMULA, context.paramCustomHeuristicFormula);
```

`paramCustomHeuristicFormula` (`SQLFunctionHeuristicPathFinderAbstract.java` L61) is then **read by nothing** - declaration, assignment, and the options set are its only three occurrences in the tree. `SQLHeuristicFormula` has no `CUSTOM` constant either (`MANHATTAN, MAXAXIS, DIAGONAL, EUCLIDEAN, EUCLIDEANNOSQR`), so there is no branch a custom formula could ever be dispatched to.

A query supplying one silently gets `MANHATTAN`, the default, with no error and no warning:

```sql
SELECT astar(#12:0, #12:99, 'weight',
  {'direction':'OUT','parVertexAxisNames':['x','y'],'customHeuristicFormula':'whatever'})
```

This is the same shape as `sysdate()`'s dropped zone argument in #6388: an option the syntax advertises, accepted without complaint, and never applied. Either implement the dispatch (with a `CUSTOM` constant) or reject the option so a caller relying on it finds out.

**Scope note:** if the feature is not wanted, removing it from the options set is the smaller fix, but it should not stay accepted-and-ignored.

---

## 2. `CONTAINSTEXT` does not use a multi-property full-text index, and changes semantics when it doesn't

Found while building the regression tests for #6382. The planner pushes `CONTAINSTEXT` down onto a **single**-property full-text index, but not onto a multi-property one.

Single property - index used:

```
+ FETCH FROM INDEX Note6382[label] ()
    label CONTAINSTEXT 'foo:bar'
```

Multi property, index on `(title, content)`, same operator on one of the two properties - full scan:

```
+ FETCH FROM TYPE Article6382 WITH FILTER ()
```

The fallback evaluates `ContainsTextCondition`, which is `((String) leftValue).contains((String) rightValue)` (`ContainsTextCondition.java` L41-49). Two consequences:

1. **No index acceleration** on exactly the indexes built for multi-field search.
2. **Different semantics depending on how the index happens to be declared.** The indexed path matches analyzer *tokens*; the fallback matches *substrings*. So `label CONTAINSTEXT 'ava'` finds a row whose label is `java` on a multi-property index and finds nothing on a single-property one. Same operator, same data, different answers.

The second half is the part worth treating as a defect rather than a missing optimisation - a user cannot tell which semantics they are getting from the query text.

---

## 3. `UpdateStatementExecutionTest.updateWithApplyDefaultsAndDynamicDefault` does not test what it is named for

`UpdateStatementExecutionTest.java` L1091, the designated regression test for issue #1814 ("APPLY DEFAULTS evaluates dynamic default expressions (e.g. sysdate) on UPDATE"):

```java
database.command("sql", "CREATE PROPERTY timestamped.updated STRING (default \"sysdate('YYYY-MM-DD')\")");
```

The escaped quotes make the default a double-quoted SQL **string literal**, so `sysdate` is never invoked - the literal text `sysdate('YYYY-MM-DD')` is stored verbatim. The test then asserts only that the property is non-null, which a constant string satisfies. No dynamic default is evaluated anywhere in it, so it cannot fail if dynamic-default evaluation regresses.

Confirmed rather than inferred: #6408 changed `sysdate()`'s argument to a zone id, which makes `sysdate('YYYY-MM-DD')` throw on the unknown zone `YYYY-MM-DD` - and this test still passes, because the call never happens. (A reviewer on #6408 predicted it would break for exactly that reason; it doesn't, and the reason it doesn't is the bug here.)

Same family as the two vacuous tests #6408 did fix (`sysdateWithFormat`, `defaultValueIsSetWithSQL`), which is how it was noticed. Fix is to drop the quotes so the default is an expression, and assert the value actually looks like a date.

---

## 4. Two independent implementations of the full-text "is this qualifier really a field" rule

`FullTextQueryExecutor.isUnqualified` (L480-484) and `LSMTreeFullTextIndex.parseQueryTerms` (L805) now encode the same rule for two different query paths:

```java
private boolean isUnqualified(final String field) {
  if (field == null || field.isEmpty() || DEFAULT_FIELD.equals(field)) return true;
  return propertyNames.size() == 1 && propertyNames.get(0).equals(field);
}
```

#6382 existed *because* these two had drifted: the Lucene-backed executor had the rule and the direct `get()` path did not, so a single-property index answered nothing for any literal containing a colon. #6408 fixed the direct path by writing the equivalent logic there, which closes the bug but leaves two copies that can drift again the same way.

Unifying them means crossing from the index class into the Lucene-backed executor (or lifting the rule into something both can see), which was more than that PR warranted - but it is the durable fix.

---

## 5. One minor Codacy CodeStyle issue was open when #6408 merged

Codacy reported `1 new issue, CodeStyle, minor` on the final push of #6408 and the PR merged with that check red. Codacy's PR comment does not name the rule or the file, and the check does not run against `main` commits, so it could not be identified from the merge commit.

Purely cosmetic and lowest priority of these five - filed only so it is not silently carried forward. Whoever has Codacy dashboard access can read the actual finding at the #6408 PR page.

lvca:
Closes #6414.

Four follow-ups from #6408, independent of each other, plus a best-effort answer to the fifth.

## 1. `astar`'s `customHeuristicFormula` is applied instead of ignored

The option was in the function's own syntax string, in its accepted-options set, and read from the caller's map into
a field that nothing ever consulted. There was no `CUSTOM` constant to dispatch to either, so a query supplying one
silently got `MANHATTAN` with no error and no warning.

The issue offered "implement the dispatch, or reject the option". Rejecting it would delete a documented feature, so
the dispatch is implemented: the option names a SQL function called as
`fn(currentVertex, parentVertex, targetVertex, sourceVertex, depth, dFactor)` and returning h(n) as a number. It is
resolved **once** when the options are bound, not once per visited node, and consulted **ahead of every axis test**,
so a custom formula works with no `vertexAxisNames` declared and owns h(n) outright: no axis is read and no
tie-breaker is layered on its answer.

Every way of getting it wrong is now an error the caller sees, which is the other half of the issue's ask:

| input | outcome |
| --- | --- |
| `customHeuristicFormula:'noSuchFn'` | rejected, naming the function |
| `heuristicFormula:'CUSTOM'` with no function named | rejected |
| `heuristicFormula:'EUCLIDEAN'` **and** `customHeuristicFormula:'f'` | rejected as a contradiction, not resolved by precedence |
| a function returning a non-number | rejected |

`dijkstra`, which has no heuristic, already rejected the option outright and still does.

## 2. `CONTAINSTEXT` and the multi-property full-text index

This turned out to be **two** defects, and the second is a wrong-results bug the issue did not know about.

**No push-down, and different semantics.** The planner required the *first* index property to be constrained and
then every following one, so a condition on one property of an index over `(title, content)` was not pushed down at
all. The scan it fell back to evaluates the operator as `String.contains` where the index matches analyzer tokens,
so the same operator over the same data answered differently depending only on how the index had been declared -
exactly as reported.

**A conjunction answered by one of its conditions.** A condition on *both* properties *was* pushed down, but
`get()` read `keys[0]` and ignored the rest, while the planner had already removed both conditions from the residual
filter. Reproduced on `2982f5504`:

```sql
SELECT id FROM Article WHERE title CONTAINSTEXT 'java' AND content CONTAINSTEXT 'zzz'
```
returned **every row of the type**. `'zzz'` matches nothing; nothing was left to check it.

**The fix.** A full-text index key is one query string per indexed property, not a composite key whose used prefix
has to be contiguous, and a multi-property index already stores every token twice - once as `field:token` and once
unprefixed - so a per-property lookup is exact rather than a filter over a superset. The key is now **positional**:
`keys[i]` is the text to find in the i-th indexed property and a `null` slot leaves that property unconstrained. Any
subset in any position is an exact lookup, and several of them are a conjunction. Every whitespace-separated word is
bound to the property its condition names, not just the first.

A single-property index builds the same one-element key it always did, and a one-element key keeps its historical
"text to find in any indexed property" meaning, so nothing that passed a single query string changes. Conditions the
index cannot serve stay in the filter that runs over the fetched rows.

> [!IMPORTANT]
> **Behaviour change.** `CONTAINSTEXT` on a property covered by a *multi*-property full-text index now matches
> analyzer tokens (case-insensitive, whole-token) rather than substrings, which is what the same operator has always
> done on a single-property index. `title CONTAINSTEXT 'ava'` no longer matches `java` there. Use
> `SEARCH_INDEX(...)` or `LIKE '%ava%'` for substring search. The operator's meaning is now decided by whether a
> full-text index covers the property - which `EXPLAIN` answers - and not by the index's arity.

## 3. The vacuous regression test for #1814

`UpdateStatementExecutionTest.updateWithApplyDefaultsAndDynamicDefault` wrote its dynamic default as
`default "sysdate('YYYY-MM-DD')"` - a double-quoted SQL **string literal**. The call never happened, the literal text
was stored verbatim, and the test asserted only that a constant string is not null, so it could not have failed if
dynamic-default evaluation regressed. It uses an expression now (`sysdate().format('yyyy-MM-dd')`) and asserts the
value parses as a date, before and after `APPLY DEFAULTS`, with a sentinel in between that the expression cannot
produce. Deliberately not an equality against today's date: the two transactions can straddle midnight.

## 4. One copy of the full-text qualifier rule

`FullTextQueryExecutor.isUnqualified` and `LSMTreeFullTextIndex.parseQueryTerms` encoded the same rule twice. #6382
existed *because* they had drifted; #6408 closed it by writing the equivalent logic in the second place. Both now
call one method, with the semantics unchanged.

## 5. The Codacy CodeStyle finding

Codacy's PR comment on #6408 names neither the rule nor the file, and the check does not run against `main`, so it
cannot be recovered from GitHub - Codacy only reports issues *new to a diff*, so it will not resurface on this PR
either. Best effort instead: scanning that PR's changed production sources found three unused imports it left behind
(`MultiValue` in `SQLFunctionAverage`, `SQLFunctionPercentile` and `SQLFunctionVariance`, whose usage it removed).
That matches "CodeStyle, minor" and is removed here. If the dashboard shows something else, it is still open.

## Verification

- New: `Issue6414AstarCustomHeuristicTest` (7 tests - dispatch, documented arguments, and each rejection path)
- New: `Issue6414ContainsTextMultiPropertyIndexTest` (6 tests - index is used for either property; a multi- and a
  single-property index answer every literal identically; a condition is restricted to the property it names; the
  conjunction; residual filters still run; the same over a BM25 index, which reaches the index by a different
  scoring path)
- `Issue6385AstarHeuristicPositionTest` iterated `SQLHeuristicFormula.values()`; it now names the five axis formulas,
  since `CUSTOM` reads no coordinate and the distance assertions say nothing about it
- Studio's `function-reference.json` updated for `astar`'s new syntax string
- Full engine unit suite: **12854 tests, 0 failures, 0 errors** (`-DexcludedGroups=benchmark,slow,vector`)

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3579,10 +3579,13 @@ of them are a conjunction. Every whitespace-separated word is bound to the prope
 the first. A single-property index builds the same one-element key it always did, and a one-element key keeps its
 historical "text to find in any indexed property" meaning, so nothing that passed a single query string changes.
 
-**Behaviour change.** `CONTAINSTEXT` on a property covered by a multi-property full-text index now matches analyzer
-tokens (case-insensitive, whole-token) rather than substrings, which is what the same operator has always done on a
-single-property index. A query relying on the substring fallback - `title CONTAINSTEXT 'ava'` matching `java` -
-stops matching there. Use a `SEARCH_INDEX(...)` wildcard, or `LIKE '%ava%'`, for substring search.
+> [!IMPORTANT]
+> **Behaviour change.** `CONTAINSTEXT` on a property covered by a multi-property full-text index now matches analyzer
+> tokens (case-insensitive, whole-token) rather than substrings, which is what the same operator has always done on a
+> single-property index. A query relying on the substring fallback - `title CONTAINSTEXT 'ava'` matching `java` -
+> stops matching there. Use a `SEARCH_INDEX(...)` wildcard, or `LIKE '%ava%'`, for substring search. The operator's
+> meaning is now decided by whether a full-text index covers the property, which `EXPLAIN` answers, and no longer by
+> how many properties that index happens to span.
 
 Two smaller items ride along. The "is this qualifier really a field?" rule now lives in one place consulted by both
 full-text query paths, rather than in two copies that drifted apart once already - #6382 existed because the
@@ -3590,6 +3593,7 @@ Lucene-backed executor had the rule and the direct `get()` path did not. And the
 #1814 wrote its dynamic default as `default "sysdate('YYYY-MM-DD')"`, a double-quoted SQL *string literal*: the
 call never happened, the literal text was stored verbatim, and the test asserted only that a constant string is not
 null - so it could not have failed if dynamic-default evaluation regressed. It now uses an expression and asserts
-the value is today's date, before and after `APPLY DEFAULTS`.
+the stored value parses as a date, before and after `APPLY DEFAULTS`, with a sentinel in between that the expression
+cannot produce. Deliberately not an equality against today's date: the two transactions can straddle midnight.
 
 [#6414](https://github.com/ArcadeData/arcadedb/issues/6414)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3540,3 +3540,56 @@ same number. It now scales like its two-axis twin and like the other four.
 [#6389](https://github.com/ArcadeData/arcadedb/issues/6389)
 [#6390](https://github.com/ArcadeData/arcadedb/issues/6390)
 [#6393](https://github.com/ArcadeData/arcadedb/issues/6393)
+
+## A documented A* option is applied, and CONTAINSTEXT answers the same question whatever the index's shape (#6414)
+
+Four follow-ups from #6408, independent of each other.
+
+`astar`'s `customHeuristicFormula` was documented in the function's own syntax string, listed among its accepted
+options, and read from the caller's map into a field that **nothing ever consulted**. There was no `CUSTOM` constant
+to dispatch to either, so a query supplying one silently got `MANHATTAN` - the default - with no error and no
+warning. The option now names a SQL function that computes `h(n)`, called as
+`fn(currentVertex, parentVertex, targetVertex, sourceVertex, depth, dFactor)` and returning a number. It is
+consulted before any vertex-axis test, so a custom formula works with no `vertexAxisNames` declared at all and owns
+`h(n)` outright: no axis is read and no tie-breaker is layered on its answer. Every way of getting it wrong is now an
+error the caller sees - an unknown function name, `heuristicFormula:'CUSTOM'` with no function named, a built-in
+formula named alongside a custom one (a contradiction, not a precedence question), or a function that returns
+something that is not a number. `dijkstra`, which has no heuristic, keeps rejecting the option outright.
+
+`CONTAINSTEXT` on a **multi-property** full-text index was two defects, and both are now fixed.
+
+A condition on one of the index's properties was not pushed down at all: the planner required the *first* index
+property to be constrained and then every following one, so `WHERE title CONTAINSTEXT 'java'` on an index over
+`(title, content)` fell back to a full scan. That is not only slower - the scan evaluates `CONTAINSTEXT` as
+`String.contains`, where the index matches analyzer *tokens*, so the same operator over the same data answered
+differently depending only on how the index had been declared. `label CONTAINSTEXT 'ava'` found a row whose label is
+`java` on a multi-property index and found nothing on a single-property one, and nothing in the query text told the
+user which they were getting.
+
+A condition on **both** properties was worse: it *was* pushed down, but the lookup read only the first key and the
+planner had already removed both conditions from the residual filter, so the second was dropped entirely.
+`WHERE title CONTAINSTEXT 'java' AND content CONTAINSTEXT 'zzz'` returned every row of the type - the second
+condition matches nothing - rather than none.
+
+A full-text index key is one query string per indexed property, not a composite key whose used prefix has to be
+contiguous, and a multi-property index already stores every token twice, once qualified as `field:token` and once
+unprefixed. The key is now POSITIONAL: `keys[i]` is the text to find in the i-th indexed property and a `null` slot
+leaves that property unconstrained, so any subset of the properties in any position is an exact lookup, and several
+of them are a conjunction. Every whitespace-separated word is bound to the property its condition names, not just
+the first. A single-property index builds the same one-element key it always did, and a one-element key keeps its
+historical "text to find in any indexed property" meaning, so nothing that passed a single query string changes.
+
+**Behaviour change.** `CONTAINSTEXT` on a property covered by a multi-property full-text index now matches analyzer
+tokens (case-insensitive, whole-token) rather than substrings, which is what the same operator has always done on a
+single-property index. A query relying on the substring fallback - `title CONTAINSTEXT 'ava'` matching `java` -
+stops matching there. Use a `SEARCH_INDEX(...)` wildcard, or `LIKE '%ava%'`, for substring search.
+
+Two smaller items ride along. The "is this qualifier really a field?" rule now lives in one place consulted by both
+full-text query paths, rather than in two copies that drifted apart once already - #6382 existed because the
+Lucene-backed executor had the rule and the direct `get()` path did not. And the designated regression test for
+#1814 wrote its dynamic default as `default "sysdate('YYYY-MM-DD')"`, a double-quoted SQL *string literal*: the
+call never happened, the literal text was stored verbatim, and the test asserted only that a constant string is not
+null - so it could not have failed if dynamic-default evaluation regressed. It now uses an expression and asserts
+the value is today's date, before and after `APPLY DEFAULTS`.
+
+[#6414](https://github.com/ArcadeData/arcadedb/issues/6414)
--- a/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionAstar.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionAstar.java
@@ -23,6 +23,7 @@ import com.arcadedb.database.Document;
 import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.database.Record;
+import com.arcadedb.exception.CommandSQLParsingException;
 import com.arcadedb.exception.RecordNotFoundException;
 import com.arcadedb.graph.Edge;
 import com.arcadedb.graph.GhostEdgeReporter;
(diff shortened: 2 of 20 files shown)


lvca:
Thanks - all four points addressed in 48f87e3b0. Point 3 was not just a coverage gap: that shape was broken, and one of its queries was a regression against `main`.

An index property can carry a modifier (`obj.hd by item`) and `put()` prefixes that property's postings with the *modified* name, while a query can only write `obj.hd` - the modified spelling carries spaces, which `parseQueryTerms`'s whitespace split tears apart before anything can recognise it. So the per-property key reached no postings at all. On a two-property `BY ITEM` index:

| query | main | first push | now |
| --- | --- | --- | --- |
| ``obj.hd`` CONTAINSTEXT 'alpha'` | *(nothing)* | *(nothing)* | **x** |
| `... 'alpha' AND `obj.tl` ... 'beta'` | x | *(nothing)* ← regression | **x** |
| `... 'alpha' AND `obj.tl` ... 'zzz'` | x ← wrong | *(nothing)* | *(nothing)* |

A qualifier now resolves against each property's **base** name and answers with the spelling the postings use, so the two names meet in one place instead of being compared as strings that never match. All four queries are right for the first time, including the single-condition one neither `main` nor the first push could answer. `Index.basePropertyName` is that base name and is now the single definition of the `by key`/`by value`/`by item` stripping - the planner, the fetch step and the full-text parser had three hand-rolled copies.

**1 + 2** - one `intersectPerProperty` and one `rankedCursor`, shared by the bucket-local lookup and the type-wide BM25 one; `FullTextSearch`'s second copy of `rankedCursor` is gone with it. That also settles the merge-vs-put asymmetry: the surviving one merges, with a comment on why.

**4** - the shadowing parameter is `astar` now, so `context` means the `CommandContext` without qualification.

Full engine unit suite after the changes: **12856 tests, 0 failures, 0 errors**.

lvca:
Both points addressed in `1ea53dafb`.

**The null bug is real** and reproduces exactly as you described. The positional key says "this property is not constrained" with a `null` slot, and `processFullTextBlock` wrote a condition whose right side *evaluated* to null into that same slot, so the two readings collided:

| query (`:missing` bound to null) | before | after |
| --- | --- | --- |
| `title CONTAINSTEXT 'java' AND content CONTAINSTEXT :missing` | **a** | *(nothing)* |
| `content CONTAINSTEXT :missing` | *(nothing)* | *(nothing)* |
| single-property `title CONTAINSTEXT :missing` | *(nothing)* | *(nothing)* |
| no index on the property | *(nothing)* | *(nothing)* |

Every other shape of the same question already answered nothing, so this was one more arity-dependent answer for `CONTAINSTEXT` - the family this issue exists to remove. A null value now empties the whole block: it is unsatisfiable, not unconstrained. Regression test covers all four rows above.

**The same-property nitpick is real too**, and reachable without an unusual schema, but it is not fixable inside the positional key: that key has one slot per property, and two texts cannot share a slot because within a property the terms of one query text are OR'd (the coordination score counts how many matched) while the `AND` between two conditions wants both. Serving it correctly means one lookup per *condition* rather than per property, feeding the existing `intersectPerProperty` - a change to the descriptor shape rather than a fix here. Filed as **#6427** with that direction and the failing case to assert (`title CONTAINSTEXT 'java' AND title CONTAINSTEXT 'CONCURRENCY'` against a title of `java concurrency`, which fails today only because the residual `String.contains` is case-sensitive where the index is not).

Full engine unit suite after the change: **12857 tests, 0 failures, 0 errors**.

lvca:
Both notes addressed in `f8abb1b2b` - documentation and coverage only, no behaviour change.

**1.** Comments at both sites (`buildIndexSearchDescriptorForFulltext` and `processFullTextBlock`) now name **#6427**, and a test pins what the shape answers today so a later change to the generic index-key path cannot quietly make it worse:

| query on an index over `(title, content)`, row title `java concurrency` | today | after #6427 |
| --- | --- | --- |
| `title CONTAINSTEXT 'java' AND title CONTAINSTEXT 'concurrency'` | e | e |
| `title CONTAINSTEXT 'java' AND title CONTAINSTEXT 'zzz'` | *(nothing)* | *(nothing)* |
| `title CONTAINSTEXT 'java' AND title CONTAINSTEXT 'CONCURRENCY'` | *(nothing)* | e |

One correction on the reading, checked before writing it down: the inner loop `break`s after the first match, so **at most one condition per index property is ever claimed** - the second stays in `remainingCondition` and *is* evaluated (row 2 above proves it). The "both removed from the residual, generic fallback runs on a two-condition block" path is real but needs one document property indexed **twice under different modifiers** (`(m by key, m by value)`), not two conditions on an ordinary property. I probed that shape as well and it answers correctly today (`m CONTAINSTEXT 'alpha' AND m CONTAINSTEXT 'zzz'` → nothing; `'alpha' AND 'beta'` → the row), and tightening the planner to skip an already-claimed base name would have pushed the second condition onto a `String.contains` over a MAP value and made that second query wrong. So the planner is left alone and the case is documented instead.

**2.** The BM25 cost note is right and worth recording: the per-property passes repeat `scanDocumentFrequencies` and the scoring context, which are identical across all of them, while only the posting walk genuinely differs. Filed as **#6436** with the shared-scoring-context direction and a `@Tag("benchmark")` guard to add with it.

Affected suites after the change: **2648 tests, 0 failures, 0 errors** (full engine suite was 12857/0/0 on the previous commit; this one touches only comments and a test).

lvca:
Both fixed in `5dd6e3cdd`, and thanks for the careful trace - the reachability arguments (no path to a null `customHeuristicFunction`, the 1-element recursion falling through on the length mismatch, `ContainsTextCondition` having no `isIndexAware` override so `position < 0` really is defensive) all match what I found.

**1.** You are right and it was my error: the test deliberately does *not* assert equality against today's date - the two transactions can straddle midnight - so it asserts the stored value **parses** as a date, with a sentinel in between that the expression cannot produce. The release note now says that instead of overstating it.

**2.** Fourth copy folded into `Index.basePropertyName`. The three that remain (`DocumentIndexer`, `TypeIndexBuilder`, `BucketIndexBuilder`) are deliberately left: they ask whether a property *is* indexed by item, which is a different question from what its base name is, and `basePropertyName` does not answer it.

**On prominence** - good call, the behaviour change is now an `> [!IMPORTANT]` callout in `docs/release-26.9.1.md` rather than an ordinary paragraph, and states what decides the operator's meaning: whether a full-text index covers the property (which `EXPLAIN` answers), and no longer how many properties that index spans.

Full engine unit suite after the change, since it touches the general index-descriptor path: **12858 tests, 0 failures, 0 errors**.

lvca:
Checked both, and neither needs a code change here.

**1. `Identifiable` on the right of `CONTAINSTEXT`** - your instinct was right, it is a can't-happen in practice: a RID matches nothing on a single-property index, a multi-property one, or a plain scan, so the guard really is defensive. But probing it turned up the *numeric* case, which is reachable and does diverge. Against a row whose `title` is `report 2024`:

| query | full-text index on the property | no index |
| --- | --- | --- |
| `title CONTAINSTEXT 2024` | the row | *(nothing)* |
| `title CONTAINSTEXT #1:0` | *(nothing)* | *(nothing)* |

`ContainsTextCondition.evaluate` refuses anything that is not a String outright; the index path calls `toString()` and analyzes it. So `content CONTAINSTEXT 2024` - what someone writes looking for a year or an invoice number - depends on whether the property happens to be indexed. Pre-existing (single-property behaved this way before this PR), and what this PR changed is that a multi-property index now *agrees* with a single-property one here too, so the remaining split is purely index-vs-no-index rather than also arity-dependent. Filed as **#6438** with the two directions and the assertion that cannot pass today either way. Deliberately not fixed here: it is a semantics decision on the unindexed path, outside #6414's five items, and not something to land unreviewed at the end of a green PR.

**3. Migration guide** - checked the docs tree: there is no separate breaking-changes or migration document, only the dated release-notes files (`release-26.8.1.md`, `release-26.9.1.md`) plus per-issue design notes. So the dated release notes are the right and only home, and the change is an `[!IMPORTANT]` callout there.

**2.** Agreed, nothing to change.

Final state: 5 commits, full engine unit suite **12858 tests, 0 failures, 0 errors**, three follow-ups filed for the deliberately deferred gaps (#6427 same-property conditions, #6436 BM25 per-property corpus passes, #6438 non-String arguments).

Issue #5928, opened 2026-08-07
Console crashes at startup on '-Dkey=' or bare '-Dkey' argument (ArrayIndexOutOfBoundsException from split)


ruispereira:
## Summary

The console crashes at startup with `ArrayIndexOutOfBoundsException` when given a `-D<key>=` (empty value) or a bare `-D<key>` (no `=`) system-property argument. It should set the property (to empty / present) or reject the argument with a clear message.

Version: 26.8.1. Module `console`.

## Detail

```java
// console/src/main/java/com/arcadedb/console/Console.java:187-189
final String[] parts = value.substring(2).split("=");
System.setProperty(parts[0], parts[1]);
```

`String.split("=")` drops trailing empty strings, so:

- `-Dkey=`  → `"key=".split("=")` → `["key"]` (length 1) → `parts[1]` throws `ArrayIndexOutOfBoundsException`.
- `-Dkey`   → `"key".split("=")` → `["key"]` → same crash.
- `-Da=b=c` → `["a","b","c"]` → sets `a=b`, silently discarding `c`.

The first two are legitimate-looking attempts (set an empty value / a flag-style property) and the third silently loses data. All three happen before the console starts.

## Reproducer

```
console -Dsome.flag=
# -> ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1, at Console startup
```

## Suggested fix

Split on the first `=` only and tolerate a missing/empty value:

```java
final String kv = value.substring(2);
final int eq = kv.indexOf('=');
if (eq < 0)
  System.setProperty(kv, "");
else
  System.setProperty(kv.substring(0, eq), kv.substring(eq + 1));
```

Low severity (CLI entry point), but a malformed-but-reasonable argument should not crash the process before it starts.

lvca:
Fixed via #5964 (merged in 33f7e71ea5af7a6866f154611620d8e4acc1e083).

PR #6422, opened 2026-08-18
fix(#6392): the value of a SET is everything after the first '=', and a brace that closes nothing is just a character

### LINKED ISSUE #6392: Console: SET rejects values containing '=' (sibling of #5928), and an unbalanced '}' disables ';' command splitting
### Summary
Two console argument-parsing bugs. One is the exact class fixed for `-D<key>=<value>` in #5928, left unfixed in the `SET` command; the other lets an unbalanced `}` disable `;` command-splitting for the rest of the input.

### Findings (all HEAD `ca89f56`)
1. **`SET` rejects values containing `=` and empty values** — `console/src/main/java/com/arcadedb/console/Console.java` L313-315:
   ```java
   final String[] parts = line.split("=");
   if (parts.length != 2)
     throw new ConsoleException("Invalid syntax for SET, use SET <name> = <value>");
   ```
   `set somekey = a=b` → 3 parts → rejected; `set somekey =` (empty value) → trailing empty dropped → 1 part → rejected. Legitimate values with `=` (connection strings, base64 padding) are unsettable. This is exactly what the sibling `-D<key>=<value>` handler was fixed for in #5928 (split on the first `=`, tolerate empty), so the two argument paths are now inconsistent. Fix: `final int sep = line.indexOf('='); key = line.substring(0, sep).trim(); value = line.substring(sep+1).trim();` with a `sep < 0` guard.
2. **Unbalanced `}` drives `braceDepth` negative, disabling `;` splitting** — `console/src/main/java/com/arcadedb/console/TerminalParser.java` L154 `braceDepth--;` has no floor at 0, and the delimiter guard at L138 requires `braceDepth == 0` to treat `;` as a command separator. A statement with one stray `}` outside quotes (e.g. a JSON typo `INSERT ... CONTENT {"a":{"b":1}}} ;`) drives `braceDepth` to `-1`; from there every `;` fails the `== 0` test and is appended into the current word instead of ending the command. In a `load <script>` this swallows the rest of the file into one malformed command. Fix: `if (braceDepth > 0) braceDepth--;` (and skip the `prevDepth == 1` newline-split heuristic when already at 0). Quoted `}` is already handled by the `quoteStart` branch, so only genuinely unbalanced input is affected.

### Scope / confidence
High confidence — both read at HEAD. Low severity (console usability / script parsing), no data impact; grouped as one because both are console argument-parsing and #1 is a direct sibling of #5928.

lvca:
Closes #6392.

Both findings in the issue are real, both read at HEAD and both reproduced by a failing test before the fix.

## 1. `SET` rejected any value containing `=`, and any empty value

`executeSet()` split on **every** `=` and demanded exactly two parts:

```java
final String[] parts = line.split("=");
if (parts.length != 2)
  throw new ConsoleException("Invalid syntax for SET, use SET <name> = <value>");
```

* `set somekey = a=b` -> 3 parts -> rejected. Connection strings, base64 padding and date patterns were unsettable.
* `set somekey =` -> `split` drops the trailing empty token -> 1 part -> rejected, so a setting could not be cleared.

The sibling `-D<key>=<value>` handler had already been taught the right rule in #5928, so the two argument paths of the same console disagreed on what a `key=value` pair is.

**What the issue proposed, and what this does instead.** The issue proposed inlining `indexOf('=')` in `executeSet`. That fixes the symptom but leaves the same rule written out twice, one copy per call site, which is how the two paths drifted apart in the first place. Instead the rule now lives in a single `Console.splitKeyValue()`: it cuts on the **first** `=` and returns `null` when there is none, leaving each caller its own policy for a missing separator. The command line keeps reading a bare `-Dkey` as an empty value (unchanged, still covered by the #5928 tests); `SET` still rejects an argument with no `=` at all, and now also rejects an empty key, which used to fall through to `setGlobalConfiguration("")` and merely print `Setting '' is not supported by the console`.

Deliberately **not** done: stripping surrounding quotes from the value. `set x = 'a=b'` still sets the literal `'a=b'`. That is a separate behaviour change with its own back-compat surface, and nothing in the issue asks for it.

## 2. An unbalanced `}` disabled `;` command splitting for the rest of the text

`TerminalParser` counts open braces so a `;` inside a JSON object literal is not a separator, and the delimiter branch requires `braceDepth == 0`. `braceDepth--` had no floor, so one closing brace too many - a typo such as `INSERT ... CONTENT {"a":{"b":1}}} ;` - drove the count to `-1`. From there **no** `;` could ever match `== 0` again: every following command was appended into the current word instead of being executed on its own. Each stray brace dug one level deeper, so no amount of well formed JSON afterwards recovered the count.

A brace that closes nothing is just a character, so the count now stops at zero. The `prevDepth == 1` newline-split heuristic needs no extra guard: with the floor in place `prevDepth` is 0 in exactly the case that used to go negative, so the heuristic already does not fire. A `}` inside quotes was, and remains, handled by the `quoteStart` branch and never counted.

One correction to the issue text: `load <script>` parses the file **line by line** (`executeLoad` resets `pending` after each line unless a block comment is open), so the damage there is bounded to the rest of the offending line rather than the rest of the file. The unbounded case is a multi-statement string handed to `parse()` - which is exactly what batch mode builds, joining every non-flag argument with `"\n;"` - and a multi-statement line typed at the prompt.

## Tests

`TerminalParserTest` (all failed before the fix, except the last, which guards the case that already worked):
* `unbalancedClosingBraceDoesNotDisableTheSeparator` - the issue's own example, three commands out.
* `closingBraceWithNothingOpenIsJustText`
* `strayBracesDoNotAccumulate` - proves a later well formed JSON object is not swallowed.
* `unbalancedClosingBraceInsideAStringIsNotCounted` - regression guard on the quoted path.

`ConsoleTest`:
* `setKeepsAValueContainingTheSeparator`, `setAcceptsAnEmptyValue` - failed before with `Invalid syntax for SET`.
* `setWithoutAKeyIsRejected` - failed before (it silently printed "not supported").
* `setStillWorksWithAPlainValue`, `setWithoutTheSeparatorIsRejected` - guards on what must not change.

Both tests restore `SERVER_BACKUP_DIRECTORY` in a `finally`, and the fixture's `@AfterEach` already drops the database and calls `GlobalConfiguration.resetAll()`.

Verified: `mvn -o -pl console -am test` -> 794 server + 75 console tests, all green.

## Follow-up spotted, not fixed here (filed as #6423)

`PostgresNetworkExecutor.setConfiguration()` splits a Postgres `SET` on `q.split("=")` and takes `parts[1]`, so it truncates a value at the second `=` - the same class of bug in a different module - and `GetDynamicContentHandler` does `pair.split("=")` then reads `kv[1]` without a length check, which throws `ArrayIndexOutOfBoundsException` on an include parameter with no `=`. Out of scope for a console issue; filed as #6423.

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -185,9 +185,10 @@ public class Console {
       if (value.startsWith("-D")) {
         // SETTING
         final String keyValue = value.substring(2);
-        final int separatorPos = keyValue.indexOf('=');
-        final String key = separatorPos < 0 ? keyValue : keyValue.substring(0, separatorPos);
-        final String propertyValue = separatorPos < 0 ? "" : keyValue.substring(separatorPos + 1);
+        final String[] pair = splitKeyValue(keyValue);
+        // ON THE COMMAND LINE A SETTING WITH NO '=' AT ALL IS THE SAME AS ONE WITH AN EMPTY VALUE
+        final String key = pair == null ? keyValue : pair[0];
+        final String propertyValue = pair == null ? "" : pair[1];
         if (key.isEmpty())
           System.err.println("Ignoring malformed system property argument '" + value + "': missing key");
         else {
@@ -310,12 +311,13 @@ public class Console {
   }
 
   private void executeSet(final String line) {
-    final String[] parts = line.split("=");
-    if (parts.length != 2)
+    // THE VALUE IS EVERYTHING AFTER THE FIRST '=': IT CAN CONTAIN FURTHER SEPARATORS AND IT CAN BE EMPTY (ISSUE #6392)
+    final String[] pair = splitKeyValue(line);
+    if (pair == null || pair[0].isBlank())
       throw new ConsoleException("Invalid syntax for SET, use SET <name> = <value>");
 
-    final String key = parts[0].trim();
-    final String value = parts[1].trim();
+    final String key = pair[0].trim();
+    final String value = pair[1].trim();
 
     switch (key.toLowerCase()) {
     case "limit" -> {
@@ -1171,6 +1173,20 @@ public class Console {
   private String getPrompt() {
     final String databaseName = databaseProxy != null ? databaseProxy.getName() : null;
     return PROMPT.formatted(databaseName != null ? "{" + databaseName + "}" : "");
+  }
+
+  /**
+   * Splits a `&lt;key&gt;=&lt;value&gt;` pair on the FIRST '=' only, so a value that contains further separators (a connection
+   * string, a base64 padding, a date pattern) is kept whole and an empty value is a value, not a missing one. Returns null when
+   * there is no separator at all, leaving to the caller the choice between rejecting the input and reading it as an empty value.
+   * Shared by the `-D&lt;key&gt;=&lt;value&gt;` arguments (issue #5928) and by the SET command (issue #6392), which used to
+   * disagree on the rule.
+   */
+  static String[] splitKeyValue(final String pair) {
+    final int separatorPos = pair.indexOf('=');
+    if (separatorPos < 0)
+      return null;
+    return new String[] { pair.substring(0, separatorPos), pair.substring(separatorPos + 1) };
   }
 
   private static boolean setGlobalConfiguration(final String key, final String value, final boolean printError) {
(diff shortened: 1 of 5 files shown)


lvca:
Follow-up filed as #6423 for the two other `key=value` parsers (Postgres `SET`, Studio include directive).

Issue #6377, opened 2026-08-18
Postgres wire has no pre-auth socket read timeout: unauthenticated thread/FD exhaustion (Redis #5912 / Bolt #5978 were hardened, Postgres was not)


ruispereira:
### Summary
The Postgres wire executor never sets a pre-authentication socket read timeout, so an unauthenticated client that connects and sends nothing (or dribbles bytes) pins a server thread and file descriptor indefinitely. The two sibling wire protocols were explicitly hardened against exactly this (Redis #5912, Bolt #5978); Postgres was left out. A secondary unbounded startup-parameter accumulation shares the same pre-auth phase.

### Where (HEAD `ca89f56`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java`:
- The executor `extends Thread` (one thread per accepted connection, started per-accept in `PostgresNetworkListener`).
- Constructor (L154) builds `new ChannelBinaryServer(socket, server.getConfiguration())` and **never calls `socket.setSoTimeout(...)`**; `ChannelBinaryServer` does not either.
- First pre-auth read is a blocking `readStartupMessage(true)` (L168 → L2007) whose initial `channel.readUnsignedInt()` has no timeout.

Contrast the hardened siblings:
- `RedisNetworkExecutor` sets `channel.socket.setSoTimeout(handshakeTimeout)` in its constructor (L105-107, #5912) and lifts it to `0` only after auth (L689).
- `BoltNetworkExecutor.negotiateTransport()` does the same (#5978).

Secondary, same phase: the startup-parameter loop reads `len` but ignores it, looping `while (readNextByte() != 0)` and putting every `name→value` pair into the pre-auth `connectionProperties` map with **no bound on pair count**.

### Repro (pre-auth, unauthenticated)
- Thread/FD exhaustion: open a TCP connection to the Postgres port and **send nothing** → the executor thread blocks forever in `readStartupMessage`'s `readUnsignedInt()`. Open N connections → N threads + N FDs pinned indefinitely (no connection cap in the listener).
- Unbounded growth: send a valid startup header (`00 00 00 08 00 03 00 00`) then a never-terminating stream of `6b 00 76 00` (`"k\0v\0"`) pairs → `connectionProperties` grows ~linearly and the thread never returns.

### Suggested fix
[...]

Issue #5978, opened 2026-08-09
Redis/Bolt wire protocol: connection-timeout hardening follow-ups from #5965 review


lvca:
## Summary

Three follow-ups surfaced during code review of #5965 (the #5907/#5911/#5912 RESP parser fixes), all related to connection-timeout hardening in the wire protocol modules. None were blocking for #5965 - review explicitly called each out as a non-blocking, forward-looking item.

### 1. `BoltNetworkExecutor`'s auth phase is unbounded, unlike the now-hardened Redis wrapper

#5965's fix for #5912 bounds the *entire* pre-auth phase of a Redis connection (through AUTH/HELLO) with a read timeout, lifted to infinite only after successful authentication.

`BoltNetworkExecutor.negotiateTransport()` only bounds the *TLS-detection* window with a timeout, then explicitly restores `setSoTimeout(0)` before its own AUTH/handshake state machine runs. So a slow or silent client can hold a Bolt connection thread open indefinitely during authentication, the same class of resource-exhaustion issue #5912 fixed for Redis.

Suggested fix: mirror the `markAuthenticated`/`markUnauthenticated` pattern from `RedisNetworkExecutor` (arm `NETWORK_SOCKET_TIMEOUT` for the full pre-auth window, lift to infinite on successful auth) in `BoltNetworkExecutor`.

### 2. No idle-timeout cap on *authenticated* Redis connections

The #5912 fix intentionally only bounds the *pre-auth* window - authenticated RESP connections are expected to sit idle between commands (normal client usage, matches real Redis behavior), so they were left with no timeout at all.

That said, a client with valid credentials can still open many long-lived idle authenticated connections with zero timeout, a resource-exhaustion vector distinct from the pre-auth case #5912 addressed. Worth considering a connection-count or idle-timeout limit for authenticated connections, if this becomes a problem in practice (not urgent - genuine Redis clients don't time out idle authenticated connections by default either, so this would be a deviation from Redis semantics, not a strict requirement).

### 3. `markUnauthenticated()` swallows `SocketException` from `setSoTimeout()`, which can silently violate the timeout invariant

[...]

Issue #5912, opened 2026-08-07
Redis wrapper: no idle/read timeout, so a client can hold a connection thread open indefinitely by trickling bytes


lvca:
## Summary

Follow-up from review on #5902 (issue #5895).

`RedisNetworkExecutor`/`ChannelBinaryServer` never configures a socket read timeout for the Redis listener, unlike `BoltNetworkExecutor`, which sets `GlobalConfiguration.NETWORK_SOCKET_TIMEOUT` as a handshake timeout before resetting to infinite (`setSoTimeout(0)`) once past the pre-auth handshake.

In `RedisNetworkExecutor.run()`, `SocketTimeoutException` is caught and silently ignored:

```java
} catch (final SocketTimeoutException e) {
  // IGNORE IT
}
```

Combined with no timeout ever being configured on the socket, `readNext()`'s blocking read (`channel.inStream.read(buffer)`) can wait indefinitely for more bytes. #5902 bounds *how much* data a client can make the parser process per declared size (`maxMultiBulkDepth`/`maxMultiBulkLength`/`maxBulkLength`), but a client can still open a connection, declare a size at or under those caps, and then trickle bytes at an arbitrarily slow rate (or send nothing at all), holding a connection thread open indefinitely. This is a slow-connection-exhaustion vector, distinct from (and not addressed by) the unbounded-size DoS #5902 fixes.

## Suggested fix

Configure a read/idle timeout on the Redis listener's socket (reusing `GlobalConfiguration.NETWORK_SOCKET_TIMEOUT`, consistent with Bolt), and actually close the connection on `SocketTimeoutException` for an idle connection instead of silently ignoring it - at minimum pre-authentication, mirroring how Bolt bounds its own pre-handshake window.

Out of scope for #5895/#5902, which were specifically about unbounded declared sizes; flagged separately since it's a related but distinct availability concern for the same listener.

PR #6424, opened 2026-08-18
fix(#6410, #6411, #6412): an idle connection waits on its socket, a blob is a bytea, and the catalog answers the shape of the question rather than the name of the asker

### LINKED ISSUE #6410: Postgres wire: an idle authenticated connection busy-polls its socket 10 times a second
Found while fixing #6377.

### What

`PostgresNetworkExecutor.readMessage()` never blocks. Its first act is:

```java
if (!channel.inputHasData()) {
  CodeUtils.sleep(100);
  return false;
}
```

and the main command loop is `while (!shutdown) { readMessage("any", ...); }`. So every **idle authenticated**
connection wakes its thread ten times a second, checks `in.available()`, finds nothing, and sleeps again. With
N idle pooled connections that is 10N timer wakeups per second doing no work - and a Postgres client pool is
expected to hold long-lived, mostly-idle connections, which is exactly the shape this is worst for.

### Why it can be a plain blocking read now

Before #6377 the socket carried no timeout, so a blocking read was indefinite. It still isn't a problem to
block: after #6377 `markAuthenticated()` sets `setSoTimeout(0)` deliberately, because an authenticated client
is *meant* to sit idle. A blocking `readUnsignedByte()` would consume no CPU and return the instant a byte
arrives, which is strictly better than a 0-100ms polling latency.

The two things that need to keep working, and do:

- **`shutdown`**: `close()` closes the channel, and closing the socket breaks a thread blocked in `read()`. The
  cancel-request path (`session.getSecond().close()` in `readStartupMessage`) already relies on precisely this
  to interrupt another connection's executor.
- **EOF**: `readMessage` already has a `catch (EOFException)` arm returning false for "client closed the
  connection", which is what a blocking read gives on a clean close.

### Suggested change

Drop the `inputHasData()` guard from `readMessage` and let the read block. Note that the pre-auth
`waitForAMessage()` added in #6377 exists *only* because `readMessage` does not block; if `readMessage`
blocks, that helper collapses into the socket timeout and can go away too, which removes the deadline
bookkeeping added there.

### Scope

Behavioural change to the idle path of every Postgres connection, so it wants its own PR and a test that an
idle connection still notices both a client disconnect and a server-side `close()` promptly. Same polling
shape may exist in the sibling wire protocols - worth checking `RedisNetworkExecutor`/`BoltNetworkExecutor`
in the same pass.

### LINKED ISSUE #6411: Postgres wire: getTypeForValue types a byte[] as _char while the schema types BINARY as varchar
Found while fixing #5290.

### What

`PostgresType` types a column two ways, and the file says in two separate comments that the two must agree
("Every branch must agree with `getTypeForValue`, which types a column from a sample row: a mismatch would
make the column's OID depend on whether the result set happens to be empty"). For binary they do not:

- `getTypeForValue(byte[])` -> `ARRAY_CHAR` (OID 1002, `"char"[]`) - `PostgresType.java:310`
- `getTypeFromArcade(Type.BINARY)` -> `VARCHAR` (OID 1043) - `PostgresType.java:418`

So a BINARY property is announced as `varchar` when the result set is empty and the OID comes from the schema,
and as `"char"[]` when a row is sampled. That is the exact failure mode those comments were written to prevent:
a client that prepares against an empty result and then re-executes gets two different types for one column.

### The deeper point

Neither answer is right. PostgreSQL has a type for this - `bytea`, OID 17, `typcategory` U, text format
`\x<hex>` - and `PostgresType` has no entry for it. A `byte[]` is not an array of one-byte characters; typing
it as one means a client decodes arbitrary bytes as text, which for any non-UTF-8 payload is lossy rather than
merely odd.

### Suggested fix

Add `BYTEA(17, "bytea", 0, 1001, byte[].class, -1, ...)` to `PostgresType` with the hex text encoding, point
both `getTypeForValue(byte[])` and `getTypeFromArcade(Type.BINARY)` at it, and let `PostgresTypeCatalog` pick
it up for free (it enumerates the enum). That resolves the disagreement by making both sides right rather
than by picking one of two wrong answers.

Worth a round-trip test through pgjdbc `getBytes()`/`setBytes()` since this is a wire-visible OID change for
BINARY columns.

### LINKED ISSUE #6412: Postgres wire: the pre-auth connection count is still unbounded, and catalog emulation is still keyed by application_name
Two follow-ups from #6377 and #5290 that were deliberately left out of PR #6405's scope.

## 1. No cap on concurrent pre-authentication connections

#6377 bounded how *long* an unauthenticated connection can hold a thread. It did not bound how *many* there
can be. `PostgresNetworkListener.run()` is an unbounded accept loop:

```java
while (active) {
  final Socket socket = serverSocket.accept();
  final PostgresNetworkExecutor connection = new PostgresNetworkExecutor(server, socket, null);
  connection.start();   // one OS thread per accepted socket, no pool, no ceiling
}
```

An attacker who opens connections faster than `NETWORK_SOCKET_TIMEOUT` (30s default) reaps them still drives
thread and descriptor count arbitrarily high - the timeout only sets the drain rate. The fix in #6405 turns an
unbounded leak into a bounded-rate one, which is a real improvement, but the count bound is the other half.

Suggested: a configurable cap on connections that have not yet authenticated, refusing further accepts (or
answering an ErrorResponse and closing) past it. The sibling protocols have the same accept-loop shape, so this
probably wants to be one mechanism applied to all three rather than three copies.

## 2. Catalog emulation is still an application_name allow-list

This is the part of #5290 that @robfrank framed as the real work ("We need to track all the query each tool
performs to gather infos about schema/catalog etc. and then emulate the answers, if possible").

#6405 did this properly for two families - the niladic system functions (`PostgresSystemQuery`) and `pg_type`
(`PostgresTypeCatalog`) - which is why aliased, quoted and semicolon-terminated spellings now work for any
client. The rest of `PostgresNetworkExecutor.parseCommand()` is still ~150 lines of exact-string matches, some
gated on `connectionProperties.get("application_name")` being literally `"dbvis"`. Those answers are correct
only for the tool that sends that exact string, from that exact version; the same query from DBeaver, pgAdmin
or psql falls through and gets nothing.

The families still handled that way, each of which every tool asks about:

- `pg_catalog.pg_namespace` / `information_schema.schemata` - the schema list
- the table list (currently matched on `query.contains("ORDER BY TABLE_TYPE,TABLE_SCHEM,TABLE_NAME ")`)
- `pg_catalog.pg_class` / `pg_attribute` - columns of a table, which is what a tool needs to render a grid
- `information_schema.usage_privileges` / `character_sets`

Suggested: give each the `PostgresTypeCatalog` treatment - a small class that recognises the shape, projects
the columns the client actually asked for in the order it asked for them, and declines anything it does not
understand rather than guessing. Then delete the `application_name` gating, since a shape-based answer serves
every client that sends that shape.

Worth doing incrementally, one catalog per PR, each with the unit tests the pattern already supports.

lvca:
Closes #6410, closes #6411, closes #6412.

## #6410 - an idle connection busy-polled its socket ten times a second

`PostgresNetworkExecutor.readMessage()` opened with
`if (!channel.inputHasData()) { sleep(100); return false; }`, so every **idle authenticated** connection woke
its thread ten times a second to ask an empty socket whether anything had arrived. A Postgres client pool is
expected to hold long-lived, mostly-idle connections, which is the shape this is worst for.

The read now blocks. The two things that had to keep working do:

- **a server-side `close()`** closes the socket, which breaks a thread blocked reading it - the cancel-request
  path already relied on exactly this;
- **EOF** is what a clean client close produces, and `readMessage` already had an arm for it.

The same non-blocking read had a second consequence, fixed with it: a client that vanished *without* sending
Terminate left its connection thread polling a socket that could never produce another byte, for the life of
the server, because `available()` answers 0 on a closed peer exactly as it does on an idle one. Only the JDBC
driver's polite `'X'` ever ended that loop.

`waitForAMessage()`, which existed only because `readMessage` could not block (#6377), is gone; the password
read blocks under the handshake timeout instead.

Checked the sibling protocols as the issue asked: `RedisNetworkExecutor` and `BoltNetworkExecutor` already
block, so there is nothing to change there.

## #6411 - a BINARY column had two different OIDs depending on whether a row was sampled

`getTypeForValue(byte[])` answered `"char"[]` (1002) and `getTypeFromArcade(Type.BINARY)` answered `varchar`
(1043) - the exact disagreement the two comments in that file were written to prevent, so a client that
prepared against an empty result and then re-executed got two different types for one column.

Neither answer was right. `BYTEA` (OID 17, `typcategory` U) is now an entry of its own with the `\x<hex>` text
encoding, the raw-bytes binary encoding, and both PostgreSQL text input formats (hex and the pre-9.0 escape
form) accepted back. Both resolution paths point at it, `PostgresTypeCatalog` picks it up for free, and the
`KNOWN_DISAGREEMENTS` entry that pinned the old behaviour in `PostgresTypeResolutionPathTest` is removed
rather than updated.

## #6412 part one - the pre-authentication connection count is now bounded

#6377 (and #5912, #5978) bounded how *long* a connection may stay unauthenticated; the accept loop still
committed one thread and one file descriptor per accepted socket with no ceiling, so a client opening
connections faster than the timeout reaps them still drove both arbitrarily high.

`PreAuthConnectionGate` (new, in `server`) is one mechanism applied to all three binary protocol listeners,
configured by `arcadedb.network.maxPreAuthConnections` (default 500, 0 = unlimited), per listener so a flood
against one protocol cannot use up the budget that lets clients of another log in.

Two deliberate choices, both documented in the class:

- **the refusal is a close, not a protocol error message.** Writing one would be an I/O performed on the
  accept thread, and a peer that never reads its socket can stall that write - handing an attacker the whole
  listener instead of one connection.
- **a permit is released the first time a connection authenticates** (or when it dies without doing so).
  Protocols that can return to an unauthenticated state - BOLT LOGOFF, Redis re-AUTH - do not take one again.

## #6412 part two - the catalog answers the shape of the question, not the name of the asker

Around 150 lines of exact-string matching, several arms gated on `application_name` being literally `"dbvis"`,
are replaced by `PostgresCatalog`. The gating was never really about the tool: those queries are written by
the JDBC driver that DBeaver, DataGrip, pgAdmin and DbVisualizer all share, and what differs between them is
spelling, not meaning.

Recognition is now by shape and the answer is computed:

1. the FROM clause says which emulated relations the query is about, and so which family of rows it wants;
2. the rows of that family are built from ArcadeDB's own schema;
3. the client's **own projection** is evaluated against each row.

Step 3 is what makes this general: `DatabaseMetaData.getTables()` comes back with `TABLE_TYPE = 'TABLE'`
because the driver's own `CASE c.relkind WHEN 'r' THEN 'TABLE' ...` is evaluated against a row saying
`relkind = 'r'` - this code never spells that string. The evaluator reads the slice of PostgreSQL expression
syntax those queries are written in: CASE in both forms, `~`/`LIKE`/`IN`/`IS NULL`, casts, array subscripts,
`row_number() OVER (PARTITION BY ... ORDER BY ...)` and a one-level derived table (the last two are what
`getColumns()` is written with).

Two rules worth calling out:

- **a shape outside that slice is declined**, and the caller sends the empty result set that pg_catalog has
  always sent for a query it did not understand. A made-up row in a system catalog is worse than no row,
  because the client believes it.
- **the WHERE clause is the one permissive place**: a predicate that cannot be read does not get to remove
  rows. Everything in this catalog is the user's own types in one database - there are no system schemas,
  toast tables or other databases' rows for an unreadable predicate to be excluding, so the ones that survive
  unread (`nspname !~ '^pg_temp_'`, `has_table_privilege(...)`) are exactly the ones that mean "keep
  everything" here. The name filters that a client depends on are read and applied.

Queries whose filters are bound parameters (`getTables`, `getColumns`) are answered at Execute time, when
their values have arrived, rather than at Parse time.

**Behaviour change worth reviewing:** the emulated schema list is now exactly one schema, named after the
connected database - which is what `current_schema()` already answered. The removed arms disagreed with each
other and with that function: one reported every *type* as a schema, another reported every *database on the
server*, which also told any authenticated user the names of databases they cannot open.

## Tests

- `Issue6410IdleConnectionIT` - an idle connection thread is never caught sleeping; it retires on a client
  Terminate, on a client that vanishes without one, and on a server-side close driven by a real
  CancelRequest. The two new behaviours fail on `main` (the second hangs until the 10s bound).
- `Issue6411ByteaIT` - a 256-byte payload round-trips through `setBytes`/`getBytes`, and the column is
  announced as `bytea` whether or not a row was sampled. Plus unit coverage of both text formats, the binary
  format, and the malformed-input rejections.
- `Issue6412PreAuthCapIT` - a connection past the cap is closed promptly (well inside the handshake timeout
  that would otherwise mask the difference), authenticating one frees the slot, and authenticated connections
  never hold permits. `PreAuthConnectionGateTest` covers the gate itself, including concurrent accepts.
- `PostgresCatalogTest` - the driver's *actual* `getSchemas`/`getTables`/`getColumns` SQL, the hand-written
  spellings that used to be exact-matched, what is declined, and that an ordinary query - including one about
  a user type named `pg_notes` - is not treated as a catalog query.
- `Issue6412CatalogIT` - the same through `DatabaseMetaData`, over the wire, with no `application_name` set.

Full `postgresw` suite green (334 unit + 160 IT), and `redisw` + `bolt` green for the gate change.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -58,6 +58,7 @@ import com.arcadedb.schema.EdgeType;
 import com.arcadedb.schema.Property;
 import com.arcadedb.schema.Schema;
 import com.arcadedb.server.ArcadeDBServer;
+import com.arcadedb.server.network.PreAuthConnectionGate;
 import com.arcadedb.server.HAServerPlugin;
 import com.arcadedb.server.security.ServerSecurityException;
 import com.arcadedb.server.security.ServerSecurityUser;
@@ -123,6 +124,11 @@ public class BoltNetworkExecutor extends Thread {
   private final ArcadeDBServer      server;
   private volatile Socket           socket; // Reassigned to the SSLSocket once TLS negotiation completes
   private final BoltSslHelper       sslHelper;
+  /**
+   * The listener's permit for a connection that has not authenticated yet, handed back the moment it does -
+   * or when it goes away without ever doing so (issue #6412).
+   */
+  private final PreAuthConnectionGate.Ticket preAuthTicket;
   private       BoltChunkedInput    input;
   private       BoltChunkedOutput   output;
   private final boolean             debug;
@@ -164,11 +170,17 @@ public class BoltNetworkExecutor extends Thread {
 
   public BoltNetworkExecutor(final ArcadeDBServer server, final Socket socket, final BoltNetworkListener listener,
       final BoltSslHelper sslHelper) {
+    this(server, socket, listener, sslHelper, null);
+  }
+
+  public BoltNetworkExecutor(final ArcadeDBServer server, final Socket socket, final BoltNetworkListener listener,
+      final BoltSslHelper sslHelper, final PreAuthConnectionGate.Ticket preAuthTicket) {
     super("BOLT-" + socket.getRemoteSocketAddress());
     this.server = server;
     this.socket = socket;
     this.listener = listener;
     this.sslHelper = sslHelper;
+    this.preAuthTicket = preAuthTicket;
     this.debug = GlobalConfiguration.BOLT_DEBUG.getValueAsBoolean();
     // NOTE: transport (TLS) negotiation and the socket I/O streams are intentionally set up in run(), on this
     // per-connection thread, so a slow/failed/hostile TLS handshake can never block the shared accept thread.
@@ -254,8 +266,20 @@ public class BoltNetworkExecutor extends Thread {
       LogManager.instance().log(this, Level.SEVERE, "BOLT connection error", e);
     } finally {
       ProtocolContext.clear();
+      // Whatever ended the loop, the listener's permit must go back: a connection that dies without ever
+      // authenticating would otherwise keep its slot for the life of the server (issue #6412).
+      releasePreAuthTicket();
       cleanup();
     }
+  }
+
+  /**
+   * Hands the listener's pre-authentication permit back. Idempotent, so authenticating and then terminating -
+   * the normal life of a connection - releases exactly one permit.
+   */
+  private void releasePreAuthTicket() {
+    if (preAuthTicket != null)
+      preAuthTicket.release();
   }
 
   /**
@@ -1716,6 +1740,7 @@ public class BoltNetworkExecutor extends Thread {
    * keep applying past a successful HELLO/LOGON.
    */
   private void markAuthenticated() {
+    releasePreAuthTicket();
     try {
       socket.setSoTimeout(0);
     } catch (final SocketException e) {
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkListener.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkListener.java
@@ -23,6 +23,7 @@ import com.arcadedb.exception.ArcadeDBException;
 import com.arcadedb.log.LogManager;
 import com.arcadedb.server.ArcadeDBServer;
 import com.arcadedb.server.ServerException;
+import com.arcadedb.server.network.PreAuthConnectionGate;
 import com.arcadedb.server.network.ServerSocketFactory;
 
 import java.io.IOException;
@@ -50,6 +51,8 @@ public class BoltNetworkListener extends Thread {
   private volatile boolean                             active = true;
   private final    Set<BoltNetworkExecutor>            activeConnections = ConcurrentHashMap.newKeySet();
   private final    int                                 maxConnections;
+  /** Bounds how many accepted connections can sit un-authenticated at once (issue #6412). */
+  private final    PreAuthConnectionGate               preAuthGate = new PreAuthConnectionGate("BOLT");
 
   public BoltNetworkListener(final ArcadeDBServer server,
       final ServerSocketFactory socketFactory,
@@ -87,6 +90,21 @@ public class BoltNetworkListener extends Thread {
             continue;
           }
 
+          // One thread and one file descriptor are committed here, before the client has proved anything: past
+          // the cap the socket is closed straight away rather than being given either (issue #6412). This is
+          // narrower than the BOLT_MAX_CONNECTIONS cap above, which counts every connection including the
+          // authenticated ones a healthy client pool holds open.
+          final PreAuthConnectionGate.Ticket ticket = preAuthGate.accept();
+          if (ticket == null) {
+            preAuthGate.logRefusal(this, socket.getRemoteSocketAddress());
+            try {
+              socket.close();
+            } catch (final IOException e) {
+              // Ignore
+            }
+            continue;
+          }
+
           socket.setPerformancePreferences(0, 2, 1);
           socket.setTcpNoDelay(true);
 
@@ -94,9 +112,15 @@ public class BoltNetworkListener extends Thread {
           // blocking TLS handshake are performed on that thread, never here on the accept path, so a slow,
           // aborted or untrusted handshake can only affect its own connection and never wedges the shared
           // listener for other clients (issue #5106).
-          final BoltNetworkExecutor connection = new BoltNetworkExecutor(server, socket, this, sslHelper);
-          activeConnections.add(connection);
-          connection.start();
+          try {
+            final BoltNetworkExecutor connection = new BoltNetworkExecutor(server, socket, this, sslHelper, ticket);
+            activeConnections.add(connection);
+            connection.start();
+          } catch (final Exception e) {
+            // The executor never started, so nothing will hand the permit back for it.
+            ticket.release();
+            throw e;
+          }
 
         } catch (final Exception e) {
           if (active)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1137,6 +1137,14 @@ public enum GlobalConfiguration {
 
   NETWORK_SOCKET_TIMEOUT("arcadedb.network.socketTimeout", SCOPE.SERVER, "TCP/IP Socket timeout (in ms)", Integer.class, 30000),
 
+  NETWORK_MAX_PREAUTH_CONNECTIONS("arcadedb.network.maxPreAuthConnections", SCOPE.SERVER, """
+      Maximum number of connections a binary wire-protocol listener (Postgres, Redis, BOLT) may hold in the phase
+      before authentication. Each accepted socket costs one thread and one file descriptor before the client has
+      proved who it is, and the handshake timeout (arcadedb.network.socketTimeout) only bounds how long each one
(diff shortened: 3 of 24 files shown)


lvca:
@claude please review this!

lvca:
Both findings were real and are fixed in `78e1851f5`.

**1. Window `ORDER BY ... DESC` was silently ascending** — confirmed, and worse than the comment suggested: the code neither honoured the direction *nor* declined it, so `row_number() OVER (... ORDER BY x DESC)` numbered its partition the wrong way round. Of the two suggested fixes I took the one that keeps the query answerable: the direction is now carried per key (`WindowCall.orderByDescending`) into the partition comparator, rather than declining a query this catalog can answer correctly. `NULLS FIRST`/`LAST` *is* now declined, because it would reorder exactly the ties this numbering breaks by position.

**2. Unbounded parser recursion** — agreed, and the reasoning about `StackOverflowError` being an `Error` rather than an `Exception` is right: it would kill the connection thread instead of being declined like every other unreadable shape. Nesting is now bounded at 100 levels, an order of magnitude above the deepest real catalog query (the driver's table list reaches four). Tested with 50,000 nested parens, which is declined.

Two of the minor items were worth taking as well:

- an un-aliased `row_number() OVER (...)` is now announced as `row_number`, the way PostgreSQL names it, instead of `?column?`;
- `1.2.3` now declines the query instead of evaluating to a SQL NULL the client never wrote — a value nobody asked for is the one outcome this design is meant to avoid.

`->`/`->>` I left alone deliberately: they tokenize into symbols the parser refuses, so a query using them is declined — which is the correct answer for an operator this evaluator cannot compute, and the same answer it would give if the tokens existed.

Full `postgresw` suite green: 390 unit + 160 IT. Coverage for both fixes is in `PostgresCatalogExpressionTest` (direction carried, NULLS declined, pathological nesting declined, malformed literal declined) and `PostgresCatalogTest` (a DESC window numbering `2, 1` end to end).

Note on `ha-integration-tests`: it has failed on two consecutive runs with *different* tests each time (`RaftDictionaryManyNewKeysIT`, then `RaftTimeSeriesWriteReadYourWritesIT` + `RaftRemoteReadYourWritesIT` + `Raft3PhaseCommitIT`), all on the Raft HTTP/remote path, which nothing in this PR touches.

lvca:
@claude review latest

lvca:
Right — fixed in \`b564ca962\`.

\`parseExpression()\`'s guard only covers recursion that comes back *through* it. \`parseNot()\` and \`parseUnary()\` recurse on themselves directly for a chain of \`NOT\` or unary \`-\`/\`+\`, so they never touched the counter — the exact gap you found. Both now run through the same guard \`parseExpression()\` uses (factored into a small \`guardDepth\` helper so the increment/check/decrement isn't duplicated a third time).

Added a chain of 50,000 \`NOT\`s and a chain of 50,000 unary minuses to \`PostgresCatalogExpressionTest\` — both decline; 20 of either still parses. (Caught my own test bug writing this: \`"-".repeat(20)\` with no separator tokenizes as \`--\`, a line comment, not 20 unary minuses — fixed by spacing them.)

Full \`postgresw\` suite green: 391 unit + 160 IT.

PR #6425, opened 2026-08-18
fix(#6368): Cypher aggregation pulls upstream at plan batch size, not a fixed 10,000

### LINKED ISSUE #6368: Variable-length path returns different node properties after `collect()`/`UNWIND`
## Environment

- ArcadeDB 26.8.1
- OpenCypher over Bolt or through the HTTP command endpoint
- Clean database

## Reproduction

Run each setup statement separately on the clean database:

```cypher
CREATE (n {id: 0, klist: ['v0']});
CREATE (n {id: 1, klist: ['v1']});
CREATE (n {id: 2, klist: ['v2']});
CREATE (n {id: 3, klist: ['v3']});

MATCH (a {id: 0}), (b {id: 1})
CREATE (a)-[:R {id: 101}]->(b);
MATCH (a {id: 0}), (b {id: 2})
CREATE (a)-[:R {id: 102}]->(b);
MATCH (a {id: 0}), (b {id: 3})
CREATE (a)-[:R {id: 103}]->(b);
MATCH (a {id: 1}), (b {id: 2})
CREATE (a)-[:R {id: 106}]->(b);
MATCH (a {id: 1}), (b {id: 3})
CREATE (a)-[:R {id: 107}]->(b);
MATCH (a {id: 2}), (b {id: 3})
CREATE (a)-[:R {id: 111}]->(b);
```

Run this query:

```cypher
CALL {
  UNWIND [1845075540] AS alias0
  MATCH p0 = (n0)-[*1..4]-()
  SET n0.klist = []
  RETURN n0, p0
}
WITH n0, p0
WHERE n0.id = 3 AND size(nodes(p0)) = 2 AND nodes(p0)[1].id = 2
RETURN n0.id AS n0Id,
       [x IN nodes(p0) | {id: x.id, klist: x.klist}] AS pathNodes;
```

Then, on a second clean database with the same setup, run the same query with
only `collect()` and `UNWIND` added before the return:

```cypher
CALL {
  UNWIND [1845075540] AS alias0
  MATCH p0 = (n0)-[*1..4]-()
  SET n0.klist = []
  WITH {n0: n0, p0: p0} AS row
  WITH collect(row) AS rows
  UNWIND rows AS row
  RETURN row.n0 AS n0, row.p0 AS p0
}
WITH n0, p0
WHERE n0.id = 3 AND size(nodes(p0)) = 2 AND nodes(p0)[1].id = 2
RETURN n0.id AS n0Id,
       [x IN nodes(p0) | {id: x.id, klist: x.klist}] AS pathNodes;
```

## Actual result

Both queries now return exactly one row. The concrete result differs as
follows, without relying on result order:

```text
Direct query:
{
  n0Id: 3,
  pathNodes: [
    {id: 3, klist: ['v3']},
    {id: 2, klist: []}
  ]
}

collect()/UNWIND query:
{
  n0Id: 3,
  pathNodes: [
    {id: 3, klist: ['v3']},
    {id: 2, klist: ['v2']}
  ]
}
```

The two queries have the same graph mutation and should expose the same node
property values. Materializing the row through `collect()`/`UNWIND` changes the
last path node's `klist` value from `['v3']` to `[]`.

The mismatch reproduced in two independent pairs of clean databases through
both Bolt and the browser/HTTP command endpoint. Run the setup and each query
against separate clean databases.

robfrank:
## What does this PR do?

Fixes a Cypher correctness bug where inserting a `collect()`/`UNWIND` pass-through into a
`MATCH ... SET ...` pipeline changed which node property values a later `WHERE`/`RETURN`
observed for a variable-length path, compared to the identical query without the
`collect()`/`UNWIND`.

`AggregationStep` and `GroupByAggregationStep` (the executor steps behind any Cypher
aggregation function, including `collect()`) pulled their entire upstream input using a
hardcoded `BATCH_SIZE = 10000`, ignoring the `nRecords` batch size (100 by default) the
rest of the plan was asked to use. A batch that large let `ExpandPathStep` build **every**
candidate variable-length path - across **every** matched start node - in one shot, before
the downstream `SetStep` had applied `SET` to *any* of them.

`TraversalPath` freezes a node's property values into the path at the moment the path is
built, and `RowAliases.propagateUpdate` deliberately only redirects a row's own top-level
variable aliases after a `SET`, not nodes nested inside a path, list, or map (see that
class's doc comment - a documented, accepted trade-off for the common case). So when the
oversized batch let `ExpandPathStep` explore a later start node's path before an earlier
start node's own `SET` had run, the earlier node's *pre-write* property values got
permanently baked into the later node's path.

A plain (non-aggregating) `RETURN` pulls in the plan's normal ~100-row batches, which -
for the reported graph - happened to interleave `SET` and traversal closely enough that
the earlier write had already landed by the time a later start node's path revisited that
node. That was incidental to the batch-size/graph-size ratio, not a guarantee -
`collect()`/`UNWIND` simply made the same underlying non-determinism visible by using a
batch large enough to defeat the interleaving entirely.

Confirmed by instrumenting `ExpandPathStep`/`SetStep` and comparing event order for the
two forms of the reported query: the non-aggregating form interleaves `EXPAND`/`SET` in
~100-row chunks (node 2's `SET` had already run 22 of 39 times by the time node 3's path
revisited it); the `collect()`/`UNWIND` form ran all 156 `EXPAND` events for all 4 start
nodes first, then all 156 `SET` events - so node 2's path reference was frozen well before
its own `SET` ever ran.

### Fix

`AggregationStep.syncPull` and `GroupByAggregationStep.syncPull` (both the single-key and
multi-key grouping paths) now pull from upstream using the caller-supplied `nRecords` by
default, matching the granularity the rest of the plan already uses, instead of a fixed
10,000-row batch. The existing `arcadedb.aggregation.batchSize` system property remains
available as an explicit override for callers who know their upstream pipeline has no
side effects and want fewer round trips.

## Motivation

Issue #6368: a bare `collect()`/`UNWIND` between a variable-length-path `MATCH` and a `SET`
changed the `klist` property value the query reported for a path node, purely because of
which internal batch size the executor happened to use - a query that should be a
value-preserving pass-through was not.

## Related issues

Closes #6368

## Additional Notes

- Both `AggregationStep` and `GroupByAggregationStep` had the identical defect (same
  hardcoded `BATCH_SIZE` pattern); both are fixed the same way, and both have a dedicated
  regression test.
- No new dependency, no public API change. The `arcadedb.aggregation.batchSize` system
  property keeps working exactly as before for anyone who had set it explicitly.

## Checklist

- [x] I have run the build using `mvn clean package` command (`mvn -pl engine -am compile test-compile`, plus targeted test runs below)

### Test plan

- `OpenCypherVariableLengthPathTest#collectUnwindDoesNotChangeVlpNodePropertiesVersusDirectReturn` -
  reproduces the exact issue query pair (direct vs. `collect()`/`UNWIND`, each against its own clean
  database) and asserts the returned `pathNodes` are identical; also pins the concrete expected shape
  (node 2's `klist` must be empty, reflecting its own `SET`).
- `OpenCypherVariableLengthPathTest#groupByCollectDoesNotChangeVlpNodePropertiesVersusDirectReturn` -
  same defect through the `GroupByAggregationStep` path (`WITH n0, collect(p0) AS paths`).
- Both new tests were verified to **fail** against the pre-fix source (via a temporary revert of the
  two source files) with the exact reported mismatch, then verified to **pass** with the fix applied.
- Full `OpenCypherVariableLengthPathTest` class (45 tests): pass.
- Broader aggregation/collect/VLP-adjacent classes (`OpenCypherCollectUnwindTest`, `OpenCypherGroupByTest`,
  `StreamingAggregationTest`, `CollectDistinctTest`, `HeadCollectTest`, `CypherLabelWriteRowFanoutIssue6312Test`,
  `CypherCollectSubqueryCorrelatedTest`, `CypherCountSubqueryCorrelatedTest`, `CypherCountSubqueryTest`): pass.
- Full `com.arcadedb.query.opencypher.**` test package (excluding `benchmark`/`slow`/`vector` lanes): pass,
  0 failures/errors.
- `mvn -pl engine -am compile test-compile`: clean.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AggregationStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AggregationStep.java
@@ -44,10 +44,22 @@ public class AggregationStep extends AbstractExecutionStep {
   private final CypherFunctionFactory functionFactory;
   private final ExpressionEvaluator evaluator;
 
-  // Configurable batch size for streaming aggregation (default: 10,000)
-  private static final int DEFAULT_BATCH_SIZE = 10000;
-  private static final int BATCH_SIZE = Integer.parseInt(
-      System.getProperty("arcadedb.aggregation.batchSize", String.valueOf(DEFAULT_BATCH_SIZE)));
+  // Batch size for pulling from upstream while accumulating aggregation input. By default this
+  // step pulls at the SAME granularity the rest of the plan uses (the nRecords it was asked for),
+  // rather than a large fixed size: a mutating clause upstream of the aggregation (SET/CREATE/
+  // MERGE/DELETE) can be interleaved with a variable-length path traversal that freezes node
+  // references into a path at the moment the path is built (TraversalPath / RowAliases -
+  // propagateUpdate only redirects a row's own top-level aliases, not nodes nested inside a path,
+  // list or map). Pulling one huge batch up front makes this step traverse - and freeze - a later
+  // row's path before an earlier row's write to that very same node has run, permanently baking in
+  // a pre-write value that a non-aggregating pull of the same query would not see (issue #6368: a
+  // bare COLLECT()/UNWIND changed which property values a later WHERE/RETURN observed, purely
+  // because it happened to pull upstream in one 10,000-row batch instead of the usual 100). An
+  // explicit override remains available for callers whose upstream has no side effects and want
+  // fewer round trips.
+  private static final String BATCH_SIZE_PROPERTY = "arcadedb.aggregation.batchSize";
+  private static final Integer CONFIGURED_BATCH_SIZE = System.getProperty(BATCH_SIZE_PROPERTY) != null ?
+      Integer.parseInt(System.getProperty(BATCH_SIZE_PROPERTY)) : null;
 
   public AggregationStep(final ReturnClause returnClause, final CommandContext context,
       final CypherFunctionFactory functionFactory) {
@@ -98,7 +110,7 @@ public class AggregationStep extends AbstractExecutionStep {
     }
 
     // Process all rows, feeding data to aggregators
-    final ResultSet prevResults = prev.syncPull(context, BATCH_SIZE);
+    final ResultSet prevResults = prev.syncPull(context, CONFIGURED_BATCH_SIZE != null ? CONFIGURED_BATCH_SIZE : nRecords);
     Result representativeRow = null;
 
     while (prevResults.hasNext()) {
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java
@@ -52,10 +52,12 @@ public class GroupByAggregationStep extends AbstractExecutionStep {
   private final CypherFunctionFactory functionFactory;
   private final ExpressionEvaluator evaluator;
 
-  // Configurable batch size for streaming aggregation (default: 10,000)
-  private static final int DEFAULT_BATCH_SIZE = 10000;
-  private static final int BATCH_SIZE = Integer.parseInt(
-      System.getProperty("arcadedb.aggregation.batchSize", String.valueOf(DEFAULT_BATCH_SIZE)));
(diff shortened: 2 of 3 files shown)


robfrank:
@claude review

PR #6433, opened 2026-08-18
fix(#6278): self-referencing edge-list chunk hangs ordinary traversals

### LINKED ISSUE #6278: A self-referencing edge-list chunk hangs ordinary traversals: the guard added for the checker's iterator is missing from EdgeIterator/VertexIterator/RIDIterator/IteratorFilterBase
### Summary

An edge-list chunk whose `previous` pointer names itself makes an ordinary graph traversal iterate forever. `EdgeLinkedList.containsEdge`, `containsVertex`, `containsLightEdge` and `getFirstEdgeConnectedToVertex` have always broken out of such a chain (`prev != null && prev.getIdentity().equals(current.getIdentity())` -> stop), but the ITERATORS did not.

#6062 (PR #6276) added that guard to `EdgeVertexIterator.hasNext()`, because `entryIterator()` is what `GraphDatabaseChecker` walks a vertex's own list with and what the new probe cache materialises through - so on that shape of corruption `CHECK DATABASE` never returned rather than reporting the damage. Measured there: with the guard forced off, the walk of a 400-entry list under a hard cap of 3,200 reaches the cap; with it, it stops at the head chunk's entries.

The sibling iterators were deliberately left alone by that PR, since nothing in it consumes them:

- `EdgeIterator`
- `VertexIterator`
- `RIDIterator`
- `IteratorFilterBase` (and therefore `EdgeIteratorFilter`, `VertexIteratorFilter`, `RIDIteratorFilter`, `EdgeVertexIteratorFilter`)

These back the ordinary traversal APIs - `Vertex.getEdges()`, `getVertices()`, `GraphEngine.getConnectedVertexRIDs()`, the SQL/Cypher expansion steps - so the same corruption hangs a normal query, not only a check, and it hangs it in a request thread rather than in a maintenance operation.

### Why it matters beyond the check

`CHECK DATABASE` is the tool an operator reaches for after a corruption incident. Today, if the corruption includes a self-referencing chunk, a traversal that touches that vertex hangs and (until #6062) the check that was supposed to diagnose it hung too. Half of that is fixed; the query side is not.

### Suggested direction

Lift the guard out of `EdgeVertexIterator.hasNext()` into the shared chunk-hop, most likely `ResettableIteratorBase`, so every iterator gets it once rather than four copies drifting apart - which is the failure mode the guard's absence here already demonstrates. The condition is one RID comparison per chunk hop, not per entry, so it is free on the hot path.

Worth deciding at the same time whether a self-loop should merely END the walk (what the direct probes do, and what #6276 did for `EdgeVertexIterator`) or should be REPORTED - ending it silently means a traversal returns a short answer over a damaged chain with nothing in the log to say so.

### Regression test

`Issue6062AdjacencyProbeCacheTest.aSelfReferencingChunkEndsTheWalkInsteadOfFeedingItForever` is the shape to copy: plant the self-loop by loading the head chunk and calling `setPrevious(itself)`, then walk under a HARD CAP and assert the count stayed below it. Deliberately not a timeout-based test - an unterminating walk does not fail a test, it hangs the build.

robfrank:
## What does this PR do?

Lifts the self-referencing-chunk guard out of `EdgeVertexIterator.hasNext()` into a shared
`ResettableIteratorBase.moveToPreviousChunk()`, and switches every chunk-hopping iterator
(`EdgeVertexIterator`, `EdgeIterator`, `VertexIterator`, `RIDIterator`, `IteratorFilterBase` -
covering `EdgeIteratorFilter`/`VertexIteratorFilter`/`RIDIteratorFilter`/`EdgeVertexIteratorFilter`)
to call it instead of hopping `currentContainer` directly.

A chunk whose `previous` pointer names itself now ends the walk on every one of these iterators,
matching the guard `EdgeLinkedList.containsEdge`/`containsVertex`/`containsLightEdge`/
`getFirstEdgeConnectedToVertex` have always had, and the guard `EdgeVertexIterator` got in #6276.
The comparison is one RID identity check per chunk hop (not per entry), so it costs nothing on the
healthy path.

## Motivation

Closes #6278

#6276 (issue #6062) added the self-loop guard to `EdgeVertexIterator.hasNext()` because that
iterator backs `EdgeLinkedList.entryIterator()`, which is what `CHECK DATABASE`'s adjacency probe
cache walks a vertex's own edge list through - so on that shape of corruption `CHECK DATABASE`
used to hang instead of reporting the damage.

That PR deliberately left the sibling iterators alone, since nothing in it consumed them. But
those iterators back the *ordinary* traversal APIs - `Vertex.getEdges()`, `getVertices()`,
`GraphEngine.getConnectedVertexRIDs()`, and the SQL/Cypher expansion steps - so the same corruption
still hangs a normal query in a request thread, not only the maintenance operation that exists to
diagnose it. This PR closes that gap.

## Related issues

Closes #6278. Builds directly on #6276 / #6062.

## Additional Notes

- Kept the established behavior of every sibling guard: a self-loop **ends the walk silently**
  rather than logging, since none of the direct `EdgeLinkedList` probes it is modeled on log
  either. (The issue raised whether a self-loop should instead be reported; going with the existing
  precedent to keep behavior consistent across every consumer of this guard, rather than making that
  chunk-hop the one place that behaves differently.)
- Proved the regression test can fail: temporarily short-circuited the guard in
  `moveToPreviousChunk()` and reran - all 4 new tests failed with `AssertionError`. Restored the fix
  afterward; full detail in the tracking doc.

## Test plan

- [x] `mvn -pl engine -am compile` - clean compile.
- [x] New test `Issue6278SelfReferencingChunkTraversalIteratorsTest` (4 tests, one per iterator
      family: `EdgeIterator`, `VertexIterator`, `RIDIterator`, `EdgeIteratorFilter`/
      `IteratorFilterBase`), each planting a self-referencing head chunk and walking under a hard
      cap (not a timeout - an unterminating walk hangs the build, not the test).
- [x] `mvn -pl engine -am test -Dtest=Issue6278SelfReferencingChunkTraversalIteratorsTest,Issue6062AdjacencyProbeCacheTest`
      - 12/12 pass, confirming no regression to the #6062/#6276 fix this builds on.
- [x] `mvn -pl engine -am test -Dtest=com.arcadedb.graph.*Test -DexcludedGroups=benchmark,slow,vector`
      - full `com.arcadedb.graph` package: 290/290 pass, no regressions.

## Checklist


robfrank:
@claude review

Issue #6418, opened 2026-08-18
isReadOnly() misclassifies a DEFINE FUNCTION call as read-only, which SubqueryStep now leans on for correctness


lvca:
Found in code review of PR #6417 (#6362's fix).

## The gap

\`SimpleCypherStatement.isReadOnly()\` (via \`anyWriteProcedureCall\`) only inspects top-level \`CALL\` clauses that resolve to a registered \`CypherProcedure\`. A call - in \`CALL\` position or plain expression position (\`WITH foo() AS x\`, \`RETURN sql.foo()\`) - into a user-defined SQL function (\`DEFINE FUNCTION\`) is invisible to it: \`SQLFunctionDefinition.execute\` runs the function body as a full \`"sqlscript"\` command, which can itself \`CREATE\`/\`UPDATE\`/\`DELETE\`. Such a statement is classified read-only.

This classifier is not new - it backs HA leader routing (#6094) - but PR #6417 is the first place a *correctness* decision leans on it: \`SubqueryStep\` now skips its post-#4182 outer-row \`Document\`-binding refresh when the inner \`CALL {}\` statement is classified read-only, because the refresh is itself what caused #6362 (a write for a later row showing through an earlier row's binding, since the refresh re-reads through the transaction cache). A subquery that writes exclusively through a \`DEFINE FUNCTION\` call would now be misclassified as not needing the refresh, silently reintroducing #4182/#6362-style staleness for that one narrow shape.

## Why this is being filed separately rather than fixed in #6417

Closing it properly needs the same whole-statement, every-expression-position walk \`isReadOnly()\` itself would need to do (a function call can appear in \`WITH\`, \`WHERE\`, \`RETURN\`, nested inside another expression, ...), and the flag this touches also backs HA leader routing where a false negative has a different, more serious blast radius (a follower could serve a write locally instead of forwarding to the leader). Not something to change under time pressure inside an 8-issue bug-fix batch unrelated to HA routing.

## Suggested fix

[...]

Issue #6404, opened 2026-08-18
gRPC `lookupByRID()` throws `RecordNotFoundException` for an existing persisted vertex


TobiasJoseHermann:
## ArcadeDB version

- Server: `26.9.1-SNAPSHOT`
- Java client: `26.9.1-SNAPSHOT`
- Commit: `ca89f56fa5de779e8d8153433b862ae3839f313c`
- Java: OpenJDK `21.0.11`
- HTTP port: `2480`
- gRPC port: `50051`

The server and `grpc-client` were built from the same commit. The server startup
confirmed:

```text
ArcadeDB Server v26.9.1-SNAPSHOT
(build ca89f56fa5de779e8d8153433b862ae3839f313c/.../main)
gRPC server started on 0.0.0.0:50051
```

## Description

`RemoteGrpcDatabase.lookupByRID()` throws `RecordNotFoundException` for an
existing, previously persisted vertex.

A SQL query executed through the same gRPC connection finds the vertex and
returns its RID. The HTTP client can also load the same RID correctly.

The failure is reproducible without an active transaction, so it is different
from #4260.

## Reproducer

The following code uses an existing persisted vertex:

```java
import com.arcadedb.database.RID;
import com.arcadedb.remote.grpc.RemoteGrpcDatabase;
import com.arcadedb.remote.grpc.RemoteGrpcServer;

import java.util.List;

RemoteGrpcServer server = new RemoteGrpcServer(
    "localhost", 50051, "root", password, true, List.of());

RemoteGrpcDatabase db = new RemoteGrpcDatabase(
    server,
    "localhost",
    50051,
    2480,
    databaseName,
    "root",
    password);

var rows = db.query(
    "sql",
    "select from Usuario where ldapId = ?",
    "heimdall");

RID rid = rows.next().getIdentity().orElseThrow();
System.out.println("GRPC_QUERY_RID=" + rid);

System.out.println(
    "REMOTE_SCHEMA_HAS_USUARIO=" + db.getSchema().existsType("Usuario"));

// Throws RecordNotFoundException even though the query above returned #13:0.
var record = db.lookupByRID(rid);

db.close();
server.close();
```

Output:

```text
GRPC_QUERY_RID=#13:0
REMOTE_SCHEMA_HAS_USUARIO=false
com.arcadedb.exception.RecordNotFoundException: Record #13:0 not found
[...]

PR #6448, opened 2026-08-19
fix(#6446): bare CALL output swallowed, RemoteSchema cache staleness on raw DDL

### LINKED ISSUE #6446: Follow-ups from #6418/#6404/#6435: bare CALL output swallowed, RemoteSchema cache staleness, ha-integration-tests flake name
Three follow-up items spotted while fixing #6418/#6404/#6415 (PR #6435, merged), filed together since none is large enough to warrant its own issue on its own.

## 1. A bare `CALL` to a write-classified statement with no `RETURN` clause loses its own output

`CypherExecutionPlan.execute()`'s write branch treats "no `RETURN` clause" as "nothing to project":

```java
if (statement.getReturnClause() == null || statement.hasFinishClause()) {
  // write side effects still happened
  return empty result set;
}
```

That's correct for `CREATE`/`SET`/`DELETE`/`MERGE`/`REMOVE`, which have no output of their own - but a bare `CALL proc(...)` with no explicit `YIELD`/`RETURN` is supposed to implicitly surface its own yielded columns, the same way a real Neo4j-compatible `CALL` does. This mechanism swallows that output for any `CALL` that gets classified as a write.

Before #6418, this was invisible for a `CALL` into an *unregistered* procedure or a custom `DEFINE FUNCTION`, since those were always misclassified read-only (the bug #6418 fixed) and so never reached the write branch. It reproduced directly while fixing #6418: `OpenCypherUnionCallProfileTest.callCustomSQLFunction`/`callCustomFunctionWithStringParam` (bare `CALL math.add(3, 5)`, no `YIELD`) started silently returning zero rows the moment the statement was correctly classified as a write. Worked around in that PR by adding explicit `YIELD value RETURN value` to the two affected tests (arguably the better style regardless), but the underlying swallowing bug is real and pre-existing: it would affect *any* write-classified bare `CALL` with no `YIELD`/`RETURN`, including a registered write `CypherProcedure` (e.g. `merge.node`) called without one - a shape that predates #6418 entirely and was never exercised by the existing test suite.

**Suggested fix**: the write branch's "nothing to project, side effects only" shortcut should not apply when the statement's only clause is a `CALL` - either check for that shape explicitly, or teach the branch to fall through to the `CALL`'s own implicit yield-all projection instead of returning empty.

## 2. `RemoteSchema`'s cache never invalidates on raw DDL

A type created via `database.command("sql", "CREATE VERTEX TYPE X")` on an already-schema-warm `RemoteDatabase`/`RemoteGrpcDatabase` connection stays invisible to that connection's schema cache indefinitely. `RemoteSchema.checkSchemaIsLoaded()` only reloads a cache that has never been loaded (`types == null`); a cache that is already populated but simply doesn't know about a type created afterward is never refreshed. Only the high-level `RemoteSchema.createXxxType()` API methods force a reload (they call `reload()` explicitly after the `CREATE` command) - raw SQL DDL through `command()` does not.

This is pre-existing and shared identically by both the HTTP (`RemoteDatabase`) and gRPC (`RemoteGrpcDatabase`) clients - it is not new in #6435, and #6435's `@cat` fix does not depend on it. It's what lets the following (narrower, already-scoped-out) edge case survive in `RemoteGrpcDatabase.grpcRecordToDBRecord()` even after #6404's fix: a connection whose schema cache is already warm but stale for one *specific* type - created concurrently, after that connection's first schema touch - still gets `null` back from `grpcRecordToDBRecord()`, and `lookupByRID()` turns that into a `RecordNotFoundException` with the same shape as the original #6404 bug, just a narrower trigger window.

**Suggested fix**: either (a) have `RemoteSchema.getType()`/`existsType()` fall back to a one-shot `reload()` retry on a cache miss before giving up (bounded, so a genuinely nonexistent type still fails fast), or (b) have every DDL-shaped `command()` call (not just the typed `createXxxType()` helpers) invalidate the cache. Option (a) is more robust since it also covers DDL issued by *another* connection/session.

## 3. `ha-integration-tests` chronic flake - additional test name observed

Already tracked (see prior notes on the `ha-integration-tests` chronic flake, ~230 tests with 2-4 red per run, unrelated to code content). PR #6435's own CI run hit three red tests in one run: `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges` and `RaftForceResyncIT.followerResyncFromLeaderKeepsDataAndResumesReplication` (both previously seen), plus `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` (not previously recorded). Confirmed unrelated to #6435's content: none of its 3 failing tests touch Cypher classification, gRPC record conversion, or the `DictionaryMultiPageTest` diagnostic that PR touched, and the same lane fails on `main` independent of any code changes. Flagging the new name for whoever next investigates the chronic flake as a class.

lvca:
## Summary
- `CypherExecutionPlan.execute()`'s write branch treated "no RETURN clause" as "nothing to project, side effects only" - correct for CREATE/SET/DELETE/MERGE/REMOVE, but it also silently discarded a bare `CALL proc(...)` statement's own implicit yield-all output once that CALL became write-classified (a regression surface opened by the #6418 classification fix). The branch now falls through to the materialized rows when the statement's only clause is a CALL (covers no-YIELD, `YIELD *`, and an explicit `YIELD col` list uniformly).
- `RemoteSchema`'s `types`/`buckets` caches were only refreshed the first time they loaded (`checkSchemaIsLoaded()` is null-gated). A type (or bucket) created afterwards via raw DDL through `command()` - as opposed to the `createXxxType()` helpers, which force an eager `reload()` - stayed invisible to an already-warm cache indefinitely, including a cache on a *different* connection/session that never issued the DDL. `existsType()`/`getType()`/`getTypeOrNull()`/`getBucketByName()`/`getBucketByNameIfExists()` now retry with a single bounded `reload()` on a cache miss before giving up, so a genuinely nonexistent name still fails fast.
- **Behavior change for `RemoteDatabase` callers**: a miss on any of the five lookups above now costs one extra network round trip (a full `select from schema:types` re-fetch) instead of being cache-only after warmup. A caller that repeatedly checks names that are usually absent (e.g. a hot loop probing many candidate type names) will now pay one round trip per miss. A time-based debounce to collapse repeated misses was implemented and then reverted - it silently suppressed the retry for a lookup that happened to land within the debounce window of an *unrelated* prior reload, which is exactly the cross-connection scenario this fix targets, and broke the new IT test that reproduces it. The trade-off (bounded per-call cost, correctness over the unconfirmed hot-path concern) is documented in `RemoteSchema`'s class Javadoc.
- Item 3 of the issue (`ha-integration-tests` chronic flake - additional test name observed) is informational only, already tracked as a pre-existing chronic flake unrelated to code content; no code change needed.
- A related but distinct gap found during review - `WITH ... CALL proc() YIELD x` (CALL not the sole clause, no RETURN) currently parses and silently returns zero rows, where real Neo4j rejects it as a compile error - is out of scope here and filed separately as #6450.

## Test plan
- [x] New regression tests in `OpenCypherUnionCallProfileTest`: bare `CALL math.add(3, 5)`, `CALL my.echo(...) YIELD *`, and `CALL math.add(3, 5) YIELD value` (explicit column list) all now surface their own output; a genuine no-RETURN write statement (`CREATE`) still yields zero rows.
- [x] New regression tests in `RemoteSchemaTest` (mocked): cache-miss retry finds a type/bucket added between two `select from schema:types` calls; a genuinely missing type/bucket still fails after exactly one bounded retry (`getType`/`getTypeOrNull`/`existsType`/`getBucketByName`/`getBucketByNameIfExists`).
- [x] New regression IT in `RemoteSchemaIT`: a type created via raw `command("sql", "CREATE VERTEX TYPE ...")` on one `RemoteDatabase` connection becomes visible to a second, already-schema-warm connection without either issuing an explicit `reload()`.
- [x] `mvn -o -pl engine -am test -Dtest="com.arcadedb.query.opencypher.**"` - 8761 tests, 0 failures.
- [x] `mvn -o -pl network test` - 458 tests, 0 failures.
- [x] `mvn -o -Pintegration -DskipTests=true -Dit.test=RemoteSchemaIT -pl server -am verify` - 6 tests, 0 failures.
- [x] Verified both new unit-test sets fail without the corresponding fix (temporarily reverted `RemoteSchema.java`), confirming they're real regression tests. Also verified the (rejected) debounce approach by watching it break the cross-connection IT test before reverting it.

Fixes #6446.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -311,9 +311,15 @@ public class CypherExecutionPlan {
       // write statement, even if it performed no actual mutation (containsUpdates() is false then).
       final QueryStatistics stats = context.getStatistics();
 
+      // A bare CALL (no YIELD/RETURN, e.g. a write-classified custom function or write procedure) is its
+      // own projection: CallStep already yields its implicit "all columns" row, so a missing RETURN clause
+      // here does not mean "side effects only" the way it does for CREATE/SET/DELETE/MERGE/REMOVE. Issue #6446.
+      final List<ClauseEntry> clauses = statement.getClausesInOrder();
+      final boolean bareCallOwnsProjection = clauses.size() == 1 && clauses.get(0).getType() == ClauseEntry.ClauseType.CALL;
+
       // If no RETURN clause (or GQL FINISH was used), return empty results
       // (write side effects still happened). Issue #3365 section 1.3.
-      if (statement.getReturnClause() == null || statement.hasFinishClause()) {
+      if (statement.hasFinishClause() || (statement.getReturnClause() == null && !bareCallOwnsProjection)) {
         final IteratorResultSet empty = new IteratorResultSet(Collections.<Result>emptyList().iterator());
         empty.setStatistics(stats);
         return empty;
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -314,6 +314,9 @@ public class CypherExecutionPlan {
       // A bare CALL (no YIELD/RETURN, e.g. a write-classified custom function or write procedure) is its
       // own projection: CallStep already yields its implicit "all columns" row, so a missing RETURN clause
       // here does not mean "side effects only" the way it does for CREATE/SET/DELETE/MERGE/REMOVE. Issue #6446.
+      // Deliberately scoped to a CALL that is the statement's ONLY clause: a CALL chained after other
+      // clauses (e.g. WITH ... CALL proc() YIELD x, or multiple chained CALLs), still with no trailing
+      // RETURN, is a separate, narrower gap left for #6450 rather than widened here.
       final List<ClauseEntry> clauses = statement.getClausesInOrder();
       final boolean bareCallOwnsProjection = clauses.size() == 1 && clauses.get(0).getType() == ClauseEntry.ClauseType.CALL;
 
--- a/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
+++ b/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
@@ -373,6 +373,26 @@ class OpenCypherUnionCallProfileTest {
   }
 
   @Test
+  void readClassifiedBareCallStillSurfacesItsOwnOutputUnaffectedByTheFix() {
+    // Sanity check the other direction: the swallowing bug only ever existed on the write-classified
+    // path (CypherExecutionPlan.execute()'s hasWriteOps branch). A bare CALL to a registered read-only
+    // procedure (db.labels(), no args) was never broken - the read-only path returns the ResultSet
+    // unfiltered, with no "no RETURN" special-casing at all - and must keep working exactly the same.
+    database.getSchema().getOrCreateVertexType("ReadOnlyCallLabel");
+
+    final ResultSet result = database.command("opencypher", "CALL db.labels()");
+
+    assertThat(result.hasNext()).isTrue();
+    boolean found = false;
+    while (result.hasNext()) {
+      final Result row = result.next();
+      if ("ReadOnlyCallLabel".equals(row.getProperty("label")))
+        found = true;
+    }
+    assertThat(found).isTrue();
+  }
+
+  @Test
   void writeStatementWithNoReturnStillYieldsNoRows() {
     // Sanity check the fix didn't broaden past bare CALL: a genuine write statement (CREATE, no CALL
     // clause at all) with no RETURN clause must still report zero rows, side effects only.
--- a/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
+++ b/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
@@ -352,6 +352,24 @@ class OpenCypherUnionCallProfileTest {
     assertThat(result.hasNext()).isTrue();
     final Result row = result.next();
     assertThat((Object) row.getProperty("value")).isEqualTo("Hello World");
+    assertThat(result.hasNext()).isFalse();
+  }
+
+  @Test
+  void bareCallWithExplicitYieldColumnListSurfacesItsOwnOutput() {
+    // Same bug, explicit YIELD column-list shape (no *, no RETURN): still a single-clause CALL
+    // statement, so it's still CallStep's own projection - here filtered to the named column - that
+    // must surface, not the "no RETURN -> side effects only" empty result.
+    database.command("sql", "DEFINE FUNCTION math.add \"SELECT :a + :b AS result\" PARAMETERS [a,b] LANGUAGE sql");
+
+    final ResultSet result = database.command("opencypher", "CALL math.add(3, 5) YIELD value");
+
+    assertThat(result.hasNext()).isTrue();
+    final Result row = result.next();
+    final Object value = row.getProperty("value");
+    assertThat(value).isNotNull();
+    assertThat(((Number) value).intValue()).isEqualTo(8);
+    assertThat(result.hasNext()).isFalse();
   }
 
   @Test
--- a/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
+++ b/engine/src/test/java/com/arcadedb/query/opencypher/OpenCypherUnionCallProfileTest.java
@@ -299,10 +299,7 @@ class OpenCypherUnionCallProfileTest {
 
     // Test calling the custom function via Cypher CALL. command(), not query(): a CALL target unresolved
     // against the built-in registries is indistinguishable at parse time from a custom DEFINE FUNCTION call,
-    // so query() now rejects it as not idempotent (issue #6418). Explicit YIELD/RETURN, not a bare CALL: a
-    // write-classified statement with no RETURN clause is CypherExecutionPlan.execute()'s signal that only
-    // side effects matter (CREATE/SET/DELETE with nothing to project), which swallows a bare CALL's own
-    // implicit output the same way once the CALL itself is no longer misclassified read-only.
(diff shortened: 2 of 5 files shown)


lvca:
Addressed all three points from the review:

**1. Performance / thundering herd on repeated misses** - implemented, then reverted after it broke correctness. I added a 250ms debounce on the miss-triggered `reload()`, but it silently suppressed the retry for a lookup that happened to land within the debounce window of an earlier, unrelated reload - which is exactly the cross-connection scenario this fix targets. Concretely, it broke the new `RemoteSchemaIT.rawDdlThroughCommandIsVisibleToAlreadyWarmSchemaCache` test: the second connection's DDL landed within 250ms of the first connection's cache warm-up, and the debounced retry silently returned stale (false) instead of reloading. Reverted in favor of the simple unconditional one-shot retry, with the trade-off now documented explicitly in the class Javadoc: a caller walking many known-nonexistent names in a hot loop pays one `reload()` per miss, which is bounded per call and was judged preferable to reintroducing stale reads. Happy to revisit with a more targeted mechanism (e.g. a caller-supplied "skip retry" hint) if this proves to be a real hot path in practice.

**2. Bucket/type asymmetry** - fixed. `getBucketByName()` / `getBucketByNameIfExists()` now retry once on a cache miss too, matching `existsType()`/`getType()`/`getTypeOrNull()`, with new regression tests (`getBucketByNameRetriesReloadOnCacheMissAndFindsNewlyCreatedBucket`, `getBucketByNameIfExistsRetriesReloadOnCacheMissAndFindsNewlyCreatedBucket`, `getBucketByNameStillThrowsForGenuinelyMissingBucketAfterOneRetry`).

**3. `WITH ... CALL proc() YIELD x` (CALL not the sole clause, no RETURN)** - confirmed this is real: it currently parses and silently returns zero rows in ArcadeDB. Real Neo4j rejects this shape outright as a compile error ("Query cannot conclude with CALL together with YIELD"), so the correct fix is a semantic-validation rule, not extending this PR's `bareCallOwnsProjection` check (which is specifically about the *standalone* case this issue was filed for). Filed as #6450 to track separately, out of scope here.

Pushed in 53b189bc4.

lvca:
Thanks for the thorough follow-up pass. Addressed the two actionable items:

- Tightened `bareCallWithYieldStarSurfacesItsOwnOutput` to assert the result set is exhausted after the one row, matching its siblings.
- Added `bareCallWithExplicitYieldColumnListSurfacesItsOwnOutput` covering the explicit `YIELD value` (no `*`, no RETURN) shape - all three ways a standalone CALL owns its own projection are now covered.
- Also expanded the PR description with an explicit callout on the network-behavior trade-off (one extra round trip per genuine cache miss), per the suggestion.

The cold-cache double-reload edge case (first-ever call on a brand new connection happens to miss) is correctly called out as harmless/negligible - not fixing it, since guarding it would add complexity for a case that costs one redundant round trip at most, only on a connection's very first lookup.

Pushed in cb4845b9a.

lvca:
Addressed the two actionable notes from this pass:

- Added a comment at `bareCallOwnsProjection`'s definition clarifying it's deliberately scoped to a CALL as the statement's only clause - a chained `WITH ... CALL proc() YIELD x` (or multiple chained CALLs) with no RETURN is the separate gap tracked in #6450, not silently assumed-exhaustive here.
- Aligned `existsType()`'s miss-then-retry shape with the other four `RemoteSchema` accessors (local-variable reassignment instead of early-return) for readability consistency. Behavior unchanged, confirmed by the existing `RemoteSchemaTest` suite (68/68 still passing).

On the two non-blocking observations left as-is:
- **Release notes callout**: the PR description already has an explicit "Behavior change for `RemoteDatabase` callers" section covering exactly this (2x round-trip cost on a genuine cold-start miss, unconditional retry going forward) - didn't see a `CHANGELOG`/release-notes file in this repo to add a second copy to, so leaving that description as the record for whoever compiles 26.9.1's notes.
- **`invalidateSchema()` only nulling `types`, not `buckets`**: confirmed pre-existing and out of scope - correctly flagged as harmless today since `reload()` always rebuilds both together and `checkSchemaIsLoaded()` gates on `types`. Not touching it here.

Pushed in 5fd29e66c.

lvca:
Last round's two points:

- **Symmetrical read-path gap**: checked - there isn't one. The "no RETURN -> empty" filtering only exists inside the `if (hasWriteOps) { ... }` block (`CypherExecutionPlan.java`). The read-only path (`!hasWriteOps`) never filters on `getReturnClause()` at all - it just returns the `resultSet` from `rootStep.syncPull(...)` directly (modulo the unrelated GQL `FINISH` suppression). A read-classified bare CALL's implicit yield-all output was never swallowed; this bug only ever existed on the write-classified path introduced by #6418's stricter classification. No change needed.
- **Concurrent-miss thundering herd**: real, and matches the same accepted trade-off as the hot-loop caveat already in the Javadoc - just extended it to note that N concurrent misses cost N round trips (each thread reloads in turn after serializing on `reload()`'s lock, no "someone already refreshed" short-circuit). Doc-only change, no code touched, per the review's own framing ("worth a one-line Javadoc callout... if not a code change").

Pushed in 9cae44864. Four consecutive review rounds now, converging - the last review explicitly called this "safe to merge."

lvca:
(Noticed and disregarding the placeholder "test comment" - responding to the real review below it.)

Addressed both notes:

- **Read-classified bare CALL test**: added `readClassifiedBareCallStillSurfacesItsOwnOutputUnaffectedByTheFix`, calling the built-in `db.labels()` procedure bare (no YIELD/RETURN, read-only per `isWriteProcedure()` defaulting to `false`). Confirms the read-only path - which never had any "no RETURN" filtering to begin with - keeps returning rows exactly as before.
- **Checked real call sites rather than reasoning in the abstract**: grepped every non-test caller of `existsType`/`getTypeOrNull`/`getBucketByNameIfExists` in the repo. Nearly all of them (gremlin, importers, server HTTP handlers, mongodbw, graphql, mcp) call `database.getSchema()` on the *local/embedded* `Database`, which uses `LocalSchema` - a completely different implementation untouched by this fix. The one real internal caller of the modified `RemoteSchema.existsType()` is `RemoteGrpcDatabase.grpcRecordToDBRecord()` (grpc-client, called once per gRPC record) - which is in fact the exact narrower edge case this issue's item 2 description itself references from #6404. It's not a thundering-herd case in practice: a type only misses once per connection (the first record of that type triggers the reload; every subsequent record of the same type hits the now-warm cache), so this fix is a direct, real improvement for that caller rather than a newly-introduced cost.

Pushed in 4ffaecc02.

Issue #6391, opened 2026-08-18
vector.* sparse/ANN: sparseDot rejects unequal inferred dims, sparseCreate rejects long[], neighbors lacks a k<=0 guard


ruispereira:
### Summary
Three `vector.*` sparse/ANN functions throw raw exceptions (or impose an unnecessary restriction) on valid input, inconsistent with their own siblings.

### Findings (all HEAD `ca89f56`)
1. **`vector.sparseDot` requires equal *inferred* dimensions** — `engine/src/main/java/com/arcadedb/function/sql/vector/SparseVector.java` L142-144:
   ```java
   public float dotProduct(final SparseVector other) {
     if (this.dimensions != other.dimensions)
       throw new IllegalArgumentException("Vector dimensions must match: " + this.dimensions + " vs " + other.dimensions);
   ```
   Dimensions are inferred from the max index, but a dot product only needs the common indices (it is 0 over non-overlapping ones). `SELECT vector.sparseDot(vector.sparseCreate([0,2],[0.5,0.3]), vector.sparseCreate([1,5],[0.2,0.8]))` → dims 3 vs 6 → throws, when the answer is `0.0`. Fix: drop the equality requirement (design bug).
2. **`vector.sparseCreate` rejects `long[]` indices** — `engine/src/main/java/com/arcadedb/function/sql/vector/SQLFunctionVectorSparseCreate.java` L81-108 has a private `toIntArray` that handles only `int[]` and `Object[]`; a `long[]` (how integer JSON arrays arrive from HTTP) matches neither and hits the `throw new CommandSQLParsingException("Index values must be integers, found: long[]")`. The shared `VectorUtils.toIntArray` already handles `long[]` (issue #4148); this private copy shadows it. Fix: delegate to `VectorUtils.toIntArray`.
[...]

Issue #6387, opened 2026-08-18
keys()/values() on a Document return a nested 1-element list and crash on scalar/null collection elements


ruispereira:
### Summary
`keys()` and `values()` on a Document return the property set wrapped in a **1-element list** (`[ {p1,p2} ]` instead of `[p1,p2]`), and both throw a `NullPointerException` when applied to a collection that contains a scalar or `null` element.

### Where (HEAD `ca89f56`)
`engine/src/main/java/com/arcadedb/query/sql/method/collection/SQLMethodKeys.java` (and the identical `SQLMethodValues.java`):
```java
if (value instanceof Document document)
  return List.of(document.getPropertyNames());   // L49: getPropertyNames() is a Set -> List.of(Set) == [ {set} ]
...
if (value instanceof Collection<?> collection) {
  return collection.stream()
    .flatMap(o -> ((Collection<Object>) execute(o, currentRecord, context, params)).stream()).toList();  // L55-57
}
return null;   // execute returns null for scalar/null elements
```
The `Map`/`Result` branches return the collection directly, so the wrapping is specific to the Document branch. In the Collection branch, `execute(o, ...)` returns `null` for a scalar or null element, then `((Collection) null).stream()` NPEs.

### Repro
- `SELECT $doc.keys()` on a record → `[["p1","p2"]]` instead of `["p1","p2"]` (same for `values()`).
- `SELECT [1,2,3].keys()` or `SELECT [{a:1}, null].keys()` → **NullPointerException**.

### Suggested fix
Document branch: `return new ArrayList<>(document.getPropertyNames());`. Collection branch: filter nulls / only recurse into map/document/result elements.

### Scope / confidence
High confidence — read directly at HEAD; the sibling Map/Result branches show the intended un-wrapped shape.

lvca:
Fixed in #6451 (merged).

Issue #6386, opened 2026-08-18
vector.approxDistance computes wrong distances for binary (byte sign-extension) and int8 (& 0xFF on signed codes)


ruispereira:
### Summary
`vector.approxDistance` computes wrong distances for both quantization modes, corrupting top-k ordering. The binary (Hamming) path sign-extends `byte` operands into `Integer.bitCount`; the int8 path masks signed, order-preserving codes with `& 0xFF`, breaking the ordering across the mid-range.

### Where (HEAD `ca89f56`)
Binary — `engine/src/main/java/com/arcadedb/function/sql/vector/SQLFunctionVectorQuantizeBinary.java` (`hammingDistance`, ~L123), reached from `SQLFunctionVectorApproxDistance` L155:
```java
distance += Integer.bitCount(packed[i] ^ other.packed[i]);   // packed[] is byte[]
```
`packed[i]`/`other.packed[i]` are `byte`, sign-extended to `int` before XOR, so a difference in bit 7 sets the upper 24 bits → `bitCount` counts ~25 instead of 1.
Repro: `SELECT vector.approxDistance(vector.quantizeBinary([1,1,1,1,1,1,1, 9.0]), vector.quantizeBinary([1,1,1,1,1,1,1,-9.0]))` — true Hamming 1 (→0.125), actual ~25 (→~3.125, can exceed 1.0).
Fix: `Integer.bitCount((packed[i] ^ other.packed[i]) & 0xFF);`

Int8 — `engine/src/main/java/com/arcadedb/function/sql/vector/SQLFunctionVectorApproxDistance.java` L137:
```java
final int diff = (q1[i] & 0xFF) - (q2[i] & 0xFF);
sumSquares += diff * diff;
```
`quantizeInt8` stores codes as `(byte)(scaled - 128)` (signed, order-preserving) and `dequantizeInt8` reads `(int)quantized[i] + 128`. Masking with `& 0xFF` reinterprets the signed code as unsigned, so two values straddling mid-range (adjacent codes `-1` and `0`) give `(255-0)² = 65025` instead of `1`.
Repro: `SELECT vector.approxDistance(vector.quantizeInt8(v1), vector.quantizeInt8(v2))` where two components sit either side of mid-range.
Fix: use the signed difference `q1[i] - q2[i]`.

### Effect
[...]

Issue #6384, opened 2026-08-18
GraphQL field aliases throw NullPointerException (unhandled 500)


ruispereira:
### Summary
A GraphQL query that uses a field **alias** throws a `NullPointerException` that propagates to the HTTP handler as an unhandled 500. Aliases are standard GraphQL, so a single valid aliased query crashes the request.

### Where (HEAD `ca89f56`)
`graphql/src/main/java/com/arcadedb/graphql/schema/GraphQLResultSet.java` L103:
```java
final SelectionSet set = fieldDefinition.getField().getSelectionSet();   // getField() == null when aliased
```
In the parser (`GraphQLParser.java` `Selection()` production, L558-561), a selection with a `:` alias sets `jjtn000.fieldWithAlias = ...` and leaves `jjtn000.field == null`. `mapBySelections` only ever calls `getField()`; `getFieldWithAlias()` is defined but **never read in main code** (grep-confirmed). `Selection.getName()` also returns the alias rather than the underlying field name, so even the property lookup would be wrong once the NPE is fixed.

### Repro
```graphql
{ books { myTitle: title } }
```
`myTitle: title` → `getField()` returns `null` → NPE at L103. `next()` runs during result iteration, outside `command()`'s try/catch, so it surfaces as an unhandled 500. There is no alias test in the suite.

### Suggested fix
Handle `fieldWithAlias` in `mapBySelections`: resolve the real field via `getFieldWithAlias().getName()` for the property lookup, use the alias as the output key, and guard `getField() == null`.

### Scope / confidence
High confidence — the parser branch and the total absence of `fieldWithAlias` consumers are both verified at HEAD.

lvca:
Fixed in #6451 (merged).

ruispereira:
Re-validated against `main` at `c50338bd8`: the query path is correctly fixed, and #6453 swept the top-level dispatch.

The introspection dispatch was not swept - five sites still read `selection.getField()` with no `getFieldWithAlias()` fallback (`getFieldWithAlias` occurs twice in the whole file). No NPE remains there, so the symptom is a wrong result rather than the 500 this issue was titled for - filed as a follow-up.

PR #6451, opened 2026-08-19
fix(#6391, #6386, #6387, #6384, #6383): vector sparse/ANN guards, approxDistance sign bugs, keys()/values() wrapping, GraphQL alias NPE and Boolean/Float args

lvca:
## Summary
Five independent, self-contained fixes surfaced by prior analysis passes:

- **#6391** - three `vector.*` sparse/ANN functions rejected valid input inconsistently with their own siblings:
  - `vector.sparseDot` required equal *inferred* dimensions even though a dot product only needs the indices two vectors have in common (implicitly 0 elsewhere).
  - `vector.sparseCreate` rejected `long[]` indices (how integer JSON arrays arrive from HTTP) - its private `toIntArray` shadowed the already-`long[]`-aware `VectorUtils.toIntArray`.
  - `vector.neighbors` had no `k <= 0` guard, unlike `rerank`/`recommend`/`discover`/`sparseNeighbors`, and could reach a negative-capacity `ArrayList` allocation.
- **#6386** - `vector.approxDistance` computed wrong distances for both quantization modes, silently corrupting top-k ordering:
  - The binary (Hamming) path sign-extended `byte` operands into `Integer.bitCount`, counting up to ~24 phantom differing bits per byte.
  - The int8 path masked signed, order-preserving codes with `& 0xFF`, inverting ordering for codes straddling the signed/unsigned midpoint.
- **#6387** - `keys()`/`values()` on a `Document` wrapped the property set in a spurious one-element list, and both NPE'd on a collection containing a scalar or `null` element.
- **#6384** - a GraphQL query using a field alias (`alias: field`, standard GraphQL syntax) threw an unhandled `NullPointerException` that surfaced as a 500. `FieldWithAlias` now extends `AbstractField` (like `Field`) so it flows through the same directive/selection-set handling instead of needing special-casing.
- **#6383** - `BooleanValue.getValue()` always returned `null` (parsed literal discarded) and `FloatValue.getValue()` returned the raw string image instead of a number, corrupting generated SQL WHERE clauses for Boolean/Float GraphQL query arguments.

Issue #6379 (sparse-vector partial-compaction correctness bug) was scoped for this batch but excluded per explicit request - it needs a larger, separately-reviewed design (a persisted recency ordinal for merged segments) rather than a one-line fix.

## Test plan
- [x] New regression tests per issue: `Issue6391VectorSparseAnnGuardsTest`, `Issue6386VectorApproxDistanceTest`, additions to `SQLMethodKeysTest`/`SQLMethodValuesTest`, `Issue6384FieldAliasTest`, `Issue6383BooleanAndFloatArgumentsTest`
- [x] Updated the one existing test (`SQLFunctionPhase4Test.sparseVectorDotDimensionMismatch`) and the several `SQLMethodValuesTest` cases that were asserting the pre-fix buggy behavior, to assert the corrected behavior instead
- [x] Full `engine` + `graphql` module test suites run clean (excluding `benchmark`/`slow`/`vector` lanes); the two anomalies seen during the run (`GAVEligibilityTest` under full-suite DB-path contention, and one `server`-module HTTP test under a concurrent agent's port contention) were confirmed pre-existing/environmental by re-running each in isolation, unrelated to this diff
- [x] `mvn -o -pl engine,graphql -am compile` clean

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/sql/vector/SQLFunctionVectorApproxDistance.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/vector/SQLFunctionVectorApproxDistance.java
@@ -131,10 +131,13 @@ public class SQLFunctionVectorApproxDistance extends SQLFunctionVectorAbstract {
     if (q1.length != q2.length)
       throw new CommandSQLParsingException("Quantized vectors must have same length: " + q1.length + " vs " + q2.length);
 
-    // Compute L2 distance: sqrt(sum((q1[i] - q2[i])^2))
+    // Compute L2 distance: sqrt(sum((q1[i] - q2[i])^2)). The codes are signed and order-preserving
+    // (quantizeInt8 stores (byte)(scaled - 128)), so the plain signed difference already matches the
+    // difference between the pre-shift [0,255] values; reinterpreting as unsigned via "& 0xFF" would
+    // instead produce a huge spurious difference for codes straddling the signed/unsigned midpoint.
     double sumSquares = 0.0;
     for (int i = 0; i < q1.length; i++) {
-      final int diff = (q1[i] & 0xFF) - (q2[i] & 0xFF);
+      final int diff = q1[i] - q2[i];
       sumSquares += diff * diff;
     }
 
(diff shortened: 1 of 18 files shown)


lvca:
Addressed both minor code-review observations:
- `SQLFunctionVectorSparseCreate`: preserved the original exception as the cause when re-throwing.
- `GraphQLResultSet.mapBySelections`: guarded the non-aliased branch against a null `Field` (ellipsis/fragment-spread selections) before calling `getSelectionSet()`.

The remaining observation (top-level GraphQL query-field aliasing, e.g. `{ myBooks: books(...) }`, still fails with "Target type not defined") is a pre-existing, out-of-scope issue - different code path (`GraphQLSchema.executeQuery`'s query dispatch) from the nested-field alias fix in this PR, not introduced or worsened by it. Filed as #6453 for a follow-up.

Issue #5289, opened 2026-07-15
Postgres Plugin treats LIST OF EMBEDDED properties as ARRAY[TEXT]


borutjures:
When querying data with the Postgres plugin (PostgresProtocolPlugin), the `LIST OF EMBEDDED` properties are returned as `JSON ARRAY of TEXT` elements instead of `JSON ARRAY of JSON` elements.

Similar to #5253 but for `ARRAY[JSON]` instead of `JSON` properties.

### Steps to reproduce

```
CREATE DOCUMENT TYPE Product IF NOT EXISTS;
CREATE PROPERTY Product.sku IF NOT EXISTS STRING;
CREATE PROPERTY Product.name IF NOT EXISTS STRING;

CREATE VERTEX TYPE Supplier IF NOT EXISTS;
CREATE PROPERTY Supplier.name IF NOT EXISTS STRING;
CREATE PROPERTY Supplier.certifications IF NOT EXISTS LIST;
CREATE PROPERTY Supplier.embedded IF NOT EXISTS EMBEDDED OF Product;
CREATE PROPERTY Supplier.embedded_list IF NOT EXISTS LIST OF Product;

INSERT INTO Supplier (name, certifications, embedded, embedded_list) VALUES ('Berlin Sensors GmbH', 'ISO-9001,RoHS', { "@type": "Product", "sku": "1234", "name": "CPU"}, [{ "@type": "Product", "sku": "1234", "name": "CPU"}]);

SELECT FROM `Supplier`;
```

The resulting JSON contains escaped value of the embedded properties in the `LIST` (see `embedded_list` property):

```
[
  {
    "name": "Berlin Sensors GmbH",
    "certifications": ["ISO-9001,RoHS"],
    "embedded": {"name":"CPU","sku":"1234","@props":"name:7,sku:7","@cat":"d","@type":"Product"},
    "embedded_list": ["{\"@type\":\"Product\",\"name\":\"CPU\",\"sku\":\"1234\"}"],
  }
]
```

The `embedded_list` column is shown as `array[text]` (`JDBC ARRAY (type 2003)`).

### Expected Result

I would expect that the `embedded_list` column is shown as `array[json]` and the resulting JSON should be:

```
    "embedded_list": [{"name":"CPU","sku":"1234","@props":"name:7,sku:7","@cat":"d","@type":"Product"}],
```

### Used tools

ArcadeDB Server v26.7.2

I used two tools to query with Postgres:

- PhpStorm 2026.1
[...]

PR #6452, opened 2026-08-19
fix(#6447): Postgres wire SHORT/BYTE, DATETIME and DECIMAL type-path disagreements

### LINKED ISSUE #6447: Postgres wire: SHORT/BYTE, DATETIME and DECIMAL still disagree between the schema and value type-resolution paths
Follow-up from #6411 (BINARY/BYTEA, merged in #6424), which fixed one of four documented disagreements between `PostgresType.getTypeFromArcade()` (the schema path) and `PostgresType.getTypeForValue()` (the value path). The other three are still pinned as `KNOWN_DISAGREEMENTS` in `PostgresTypeResolutionPathTest`, each with its own comment explaining why it isn't a missing switch case:

```java
KNOWN_DISAGREEMENTS.put(Type.SHORT, new PathDisagreement(PostgresType.SMALLINT, PostgresType.INTEGER));
KNOWN_DISAGREEMENTS.put(Type.BYTE, new PathDisagreement(PostgresType.SMALLINT, PostgresType.INTEGER));
KNOWN_DISAGREEMENTS.put(Type.DATETIME, new PathDisagreement(PostgresType.TIMESTAMP, PostgresType.DATE));
KNOWN_DISAGREEMENTS.put(Type.DECIMAL, new PathDisagreement(PostgresType.DOUBLE, PostgresType.VARCHAR));
```

Same failure shape as #6411 in each case: a column's announced OID depends on whether the result set happens to be empty (schema path) or has a sampled row (value path), so a client that prepares against an empty result and then re-executes can see the type change under it.

### SHORT / BYTE
The value path widens `Short`/`Byte` to `INTEGER`, because `PostgresType` has no `int2[]`/int2-scalar-consistent entry to pair the schema path's `SMALLINT` with. Reconciling on `SMALLINT` would change the OID existing clients already receive for a *populated* result (currently `INTEGER`), so this needs a deliberate compatibility call, not just a switch-case fix.

### DATETIME
`java.util.Date` is the default Java runtime type of both `Type.DATE` and `Type.DATETIME`, so the value path (which types from a sample value's Java class) cannot tell them apart and always answers `DATE`. The schema path correctly answers `TIMESTAMP` for `DATETIME`. Needs either a distinguishing wrapper/marker on the DATETIME sample value, or acceptance that the value path stays `DATE`-only and the schema path is changed to match (lossy the other way).

### DECIMAL
Neither current answer is right: `DOUBLE` (value path) is lossy for arbitrary-precision `BigDecimal`, and `VARCHAR` (schema path) makes a client treat the column as text. PostgreSQL has the correct type for this: `NUMERIC` (OID 1700), which would need a binary encoder (`BigDecimal` doesn't map onto any of `PostgresType`'s existing 2/4/8-byte binary encodings) - same shape of work #6411 did for `BYTEA`.

### Suggested approach
Same as #6411: add the missing `PostgresType` entries (a `NUMERIC` for DECIMAL; whatever is decided for SHORT/BYTE and DATETIME), point both `getTypeFromArcade()` and `getTypeForValue()` at them, and remove the corresponding `KNOWN_DISAGREEMENTS` entry - that map's own doc comment says removing an entry is the intended way to land each fix, and `PostgresTypeResolutionPathTest.knownDisagreementsStillDisagree()` will fail (correctly) if a fix lands without updating the map, so the test scaffolding is already in place.

Worth doing incrementally, one type per PR, the way #6411 did BYTEA - DECIMAL is probably the highest-value one since `BigDecimal` is a common column type and `VARCHAR`-as-text is the most surprising of the three for a client to hit.

lvca:
## Summary

Follow-up to #6411 (BYTEA, merged in #6424): a RowDescription column is typed either from a sample value (`PostgresType.getTypeForValue`, the value path) or from the declared schema (`PostgresType.getTypeFromArcade`, the schema path, taken when a query returns no rows). When the two disagree, a column's advertised OID depends on whether the result set happens to be empty, and a client that prepares against an empty result and then re-executes sees the type change under it. #6411 fixed this for `BINARY`; this PR fixes the three remaining pinned disagreements in `PostgresTypeResolutionPathTest`.

- **DECIMAL**: neither existing answer was right - `DOUBLE` (value path) is lossy for arbitrary-precision `BigDecimal`, and `VARCHAR` (schema path) makes a client treat the column as text. Adds a proper `PostgresType.NUMERIC` (OID 1700) entry with a real binary codec (PostgreSQL's digit-group-of-4 wire format, same shape of work #6411 did for `BYTEA`), and a text encoder that renders via `BigDecimal.toPlainString()` instead of `toString()`'s occasional scientific notation. Both paths now point at it.
- **SHORT/BYTE**: the value path widened `Short`/`Byte` to `INTEGER` because there was nowhere narrower to point it; the schema path already answered the correct `SMALLINT`. The value path now answers `SMALLINT` too.
- **DATETIME**: `java.util.Date` is the default Java runtime type of both `DATE` and every `DATETIME*` variant, so a bare sampled value can't tell them apart, and the value path always answered `DATE`. Rather than adding a wrapper/marker value (one option the issue raised) or making the schema path lossy to match (the other), `PostgresNetworkExecutor.getColumns()` now resolves the ambiguity from the schema when the row is backed by one - reusing the mechanism already in place for typing an empty `LIST` column from its declared `LIST OF` element type (#5289), rather than a value-only guess. This keeps the DATETIME entry in `PostgresTypeResolutionPathTest.KNOWN_DISAGREEMENTS` on purpose: the two *pure* functions still disagree for a bare, schema-less `Date` (there is no better answer than `DATE` for one), but every real column now resolves correctly.

**Behaviour change**: a populated `SHORT`, `BYTE` or `DECIMAL` column now announces `int2`/`numeric` instead of the `int4`/`double`/`varchar` OID it used to. This corrects the client-visible bug (OID depending on emptiness), but a client relying on the old, disagreeing OID for a non-empty result sees a different type name and, for `DECIMAL`, a `BigDecimal` instead of a `Double`/`String`. Documented in `docs/release-26.9.1.md`.

## Test plan

- [x] `PostgresTypeResolutionPathTest`: `SHORT`/`BYTE`/`DECIMAL` removed from `KNOWN_DISAGREEMENTS`; `schemaPathAgreesWithValuePathForEveryType()` now covers them
- [x] `PostgresTypeTest`: updated pre-existing pinned expectations (`getTypeForValueShort/Byte`, `getTypeFromArcadeDecimal`) + new unit tests for the NUMERIC binary codec (round-trip over a wide spread of values including zero, negatives, leading/trailing zero-group stripping, negative-scale BigDecimal; a hand-verified known-encoding pin; malformed-payload rejection for NaN/invalid sign/oversized digit count)
- [x] New `Issue6447TypeResolutionIT`: end-to-end JDBC round trip for each of DECIMAL/SHORT/BYTE/DATETIME, asserting the announced type name and round-tripped value agree whether or not a row was sampled
- [x] Full `postgresw` unit suite (399 tests) and the relevant IT classes (`PostgresWJdbcIT`, `Issue6411ByteaIT`, `Issue5290ClientCompatibilityIT`, `Issue6412CatalogIT`) green, no regressions
- [x] `engine`/`server`/`postgresw` compile clean

Closes #6447

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -16,8 +16,12 @@ right - PostgreSQL already has the correct type for this, `NUMERIC` (OID 1700) -
 `PostgresType.NUMERIC` entry with a real binary codec (PostgreSQL's own digit-group-of-4 wire format, not a
 lossy float64 detour), plus a text encoder that renders a `BigDecimal` as a plain decimal string rather than
 `BigDecimal.toString()`'s occasional scientific notation. Both the codec's decode side (attacker-controlled wire
-bytes) and encode side (a `BigDecimal` of unbounded scale, which ArcadeDB itself can produce) reject an
-out-of-range digit count rather than silently wrapping it through the wire format's `int16` header fields.
+bytes) and encode side (a `BigDecimal` of unbounded scale or magnitude, which ArcadeDB itself can produce) reject
+an out-of-range digit count rather than silently wrapping it through the wire format's `int16` header fields, and
+do so before any work proportional to that size runs. The cap - 16,000 total decimal digits - is deliberately
+well under real PostgreSQL's own `NUMERIC` limit (131,072 integer digits, 16,383 fractional): `DECIMAL` has no
+configured precision/scale limit in ArcadeDB, so this is the boundary between "large value real Postgres would
+also accept" and "value this protocol declines" rather than a hard technical ceiling.
 
 `SHORT`/`BYTE` had a narrower bug: the value path widened `Short`/`Byte` to `INTEGER` because there was nowhere
 narrower to point it, while the schema path already answered the correct `SMALLINT` (`int2`). The value path now
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3,36 +3,48 @@
 This is a living document: fixes, improvements, new features, and breaking changes are collected here as
 they land during the 26.9.1 development cycle, so the release notes are ready at tag time.
 
-## Postgres wire: `SHORT`/`BYTE`, `DATETIME` and `DECIMAL` no longer change type depending on whether a row was sampled (#6447)
+## Postgres wire: `SHORT`/`BYTE`, `DATE`, `DATETIME` and `DECIMAL` no longer change type depending on whether a row was sampled (#6447)
 
 A RowDescription column is typed either from a sample value, when the result set has a row, or from the declared
 schema, when it does not - and the two paths have to agree, or a client that prepares against an empty result and
 then re-executes against a populated one sees the column's OID change under it. #6411 fixed this for `BINARY`
-(`bytea`); three more types had the same shape of bug.
+(`bytea`); four more types had the same shape of bug.
 
-`DECIMAL` was the worst of the three: the schema path answered `DOUBLE`, lossy for an arbitrary-precision
+`DECIMAL` was the worst of the four: the schema path answered `DOUBLE`, lossy for an arbitrary-precision
 `BigDecimal`, and the value path answered `VARCHAR`, which makes a client treat the column as text. Neither was
 right - PostgreSQL already has the correct type for this, `NUMERIC` (OID 1700) - so both paths now point at a new
 `PostgresType.NUMERIC` entry with a real binary codec (PostgreSQL's own digit-group-of-4 wire format, not a
 lossy float64 detour), plus a text encoder that renders a `BigDecimal` as a plain decimal string rather than
-`BigDecimal.toString()`'s occasional scientific notation.
+`BigDecimal.toString()`'s occasional scientific notation. Both the codec's decode side (attacker-controlled wire
+bytes) and encode side (a `BigDecimal` of unbounded scale, which ArcadeDB itself can produce) reject an
+out-of-range digit count rather than silently wrapping it through the wire format's `int16` header fields.
 
 `SHORT`/`BYTE` had a narrower bug: the value path widened `Short`/`Byte` to `INTEGER` because there was nowhere
 narrower to point it, while the schema path already answered the correct `SMALLINT` (`int2`). The value path now
 answers `SMALLINT` too.
 
-`DATETIME` was different in kind: `java.util.Date` is the default Java runtime type of both `DATE` and every
-`DATETIME*` variant, so a bare sampled value cannot tell them apart and the value path always answered `DATE`.
-Rather than adding a wrapper type just to carry that distinction through a value the executor already has more
-context for, `PostgresNetworkExecutor.getColumns()` now resolves the ambiguity from the schema when the row is
-backed by one - the same mechanism it already used to type an empty `LIST` column from its declared `LIST OF`
-element type (#5289) rather than guessing from a value with nothing to sample.
+`DATE` and `DATETIME` both trace back to the same root cause, from two different directions. `Type.DATE`'s and
+`Type.DATETIME`'s runtime representations are configurable per database (`GlobalConfiguration.DATE_IMPLEMENTATION`
+/`DATE_TIME_IMPLEMENTATION`) and default to `LocalDate`/`LocalDateTime` - not `java.util.Date`. The value path had
+no case at all for a bare `LocalDate`, so a `DATE` column fell all the way through to `VARCHAR` on every
+default-configured database; it now resolves to `DATE`, matching the schema path. `java.util.Date` remains a
+supported alternative implementation for both types, and it is where the two collide: `Date` is `DATE`'s default
+representation too, so a `Date`-configured `DATETIME` column's sampled value cannot be told apart from `DATE` by
+value alone. Rather than adding a wrapper type just to carry that distinction through a value with nothing else to
+go on, `PostgresNetworkExecutor.getColumns()` now resolves the ambiguity from the schema - the same mechanism it
+already used to type an empty `LIST` column from its declared `LIST OF` element type (#5289). That mechanism only
+worked for a whole-entity projection (`SELECT FROM Type`) until this change, though: a query that projects specific
+columns (`SELECT col FROM Type`, the shape of essentially every real client query) produces rows that carry only
+the projected values, not the backing element the lookup needs - so the schema fallback is now also resolved from
+the query's `FROM` target when the row itself isn't one, closing the gap for the query shape it needed to cover
+most.
 
 > [!IMPORTANT]
-> **Behaviour change.** A populated `SHORT`, `BYTE` or `DECIMAL` column now announces `int2` or `numeric` instead
-> of the `int4`/`double`/`varchar` OID it used to. This corrects a client-visible bug (the OID depended on whether
-> the result set happened to be empty), but a client relying on the old, disagreeing OID for a non-empty result
-> sees a different type name and, for `DECIMAL`, a `BigDecimal` instead of a `Double`/`String`.
+> **Behaviour change.** A populated `SHORT`, `BYTE`, `DATE` or `DECIMAL` column now announces `int2`, `date` or
+> `numeric` instead of the `int4`/`varchar`/`double`/`varchar` OID it used to. This corrects a client-visible bug
+> (the OID depended on whether the result set happened to be empty, and for `DATE` on a default-configured
+> database it was `varchar` unconditionally), but a client relying on the old, disagreeing OID for a non-empty
+> result sees a different type name and, for `DECIMAL`, a `BigDecimal` instead of a `Double`/`String`.
 
 ## A Cypher node's labels are the ones it answers to, and adding one no longer takes one away (#6363)
 
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3,6 +3,37 @@
 This is a living document: fixes, improvements, new features, and breaking changes are collected here as
 they land during the 26.9.1 development cycle, so the release notes are ready at tag time.
 
+## Postgres wire: `SHORT`/`BYTE`, `DATETIME` and `DECIMAL` no longer change type depending on whether a row was sampled (#6447)
+
+A RowDescription column is typed either from a sample value, when the result set has a row, or from the declared
+schema, when it does not - and the two paths have to agree, or a client that prepares against an empty result and
+then re-executes against a populated one sees the column's OID change under it. #6411 fixed this for `BINARY`
+(`bytea`); three more types had the same shape of bug.
+
+`DECIMAL` was the worst of the three: the schema path answered `DOUBLE`, lossy for an arbitrary-precision
+`BigDecimal`, and the value path answered `VARCHAR`, which makes a client treat the column as text. Neither was
+right - PostgreSQL already has the correct type for this, `NUMERIC` (OID 1700) - so both paths now point at a new
+`PostgresType.NUMERIC` entry with a real binary codec (PostgreSQL's own digit-group-of-4 wire format, not a
+lossy float64 detour), plus a text encoder that renders a `BigDecimal` as a plain decimal string rather than
+`BigDecimal.toString()`'s occasional scientific notation.
+
+`SHORT`/`BYTE` had a narrower bug: the value path widened `Short`/`Byte` to `INTEGER` because there was nowhere
+narrower to point it, while the schema path already answered the correct `SMALLINT` (`int2`). The value path now
+answers `SMALLINT` too.
+
+`DATETIME` was different in kind: `java.util.Date` is the default Java runtime type of both `DATE` and every
+`DATETIME*` variant, so a bare sampled value cannot tell them apart and the value path always answered `DATE`.
+Rather than adding a wrapper type just to carry that distinction through a value the executor already has more
+context for, `PostgresNetworkExecutor.getColumns()` now resolves the ambiguity from the schema when the row is
+backed by one - the same mechanism it already used to type an empty `LIST` column from its declared `LIST OF`
+element type (#5289) rather than guessing from a value with nothing to sample.
+
+> [!IMPORTANT]
+> **Behaviour change.** A populated `SHORT`, `BYTE` or `DECIMAL` column now announces `int2` or `numeric` instead
+> of the `int4`/`double`/`varchar` OID it used to. This corrects a client-visible bug (the OID depended on whether
+> the result set happened to be empty), but a client relying on the old, disagreeing OID for a non-empty result
+> sees a different type name and, for `DECIMAL`, a `BigDecimal` instead of a `Double`/`String`.
+
 ## A Cypher node's labels are the ones it answers to, and adding one no longer takes one away (#6363)
 
 `labels()` decided a vertex's labels from a single question - does the type have supertypes - and answered "the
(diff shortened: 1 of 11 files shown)


lvca:
@claude please review this!

lvca:
Addressed the review findings in b71d51756:

1. **`PostgresTypeCatalog.category()`** - added `NUMERIC` to the `"N"` arm.
2. **`PostgresTypeCatalog.inputFunction()`** - added `NUMERIC -> "numeric_in"`.
3. **Test coverage** - added assertions for both (`inputFunctionsAreSpelledTheWayPostgresSpellsThem`, new `numericIsCategorisedAsNumberLikeEveryOtherNumericType`).
4. **`numeric_precision_radix`** - `NUMERIC` now answers `10` instead of the flat `2` every other native scalar type gets, via a small `numericPrecisionRadix()` helper. Left the pre-existing `BYTEA`-should-probably-be-`null` question alone as noted - out of this PR's scope. New test: `aDecimalColumnReportsBase10PrecisionRadix`.
5. **Encode-side bound on `putNumericBinary`** - added the same class of guard the decode side already had: an out-of-range scale is now rejected before any padding work (so an absurd scale fails fast rather than allocating a proportionally huge string), and `ndigits`/`weight` are checked against the same bounds right before the `(short)` casts that would otherwise wrap them silently. New test: `serializeAsBinaryNumericRejectsExcessiveScale`.

**Not changed - the String-vs-BigInteger nit**: leaving the digit-group extraction as `String`/`substring`/`Integer.parseInt` rather than raw `BigInteger` arithmetic. You flagged this yourself as "not a real concern for typical DECIMAL precision" and only worth revisiting if it ever sits on a hotter path - it doesn't (once per NUMERIC column per row, and BigDecimal itself is already the heavier object here). Rewriting a tested, subtle digit-packing algorithm for a marginal, unmeasured gain trades a real correctness risk for a hypothetical performance one, so I'm leaving it as-is.

@claude please review this!

lvca:
Addressed both findings from the second review round in 2cd23b3fd - and the second one, once traced down, turned out to uncover a real structural gap beyond what was reported. Details:

**1. The residual encode-side gap (integer-part magnitude, not just scale)**: `putNumericBinary` now rejects an out-of-range scale/digit-count/weight consistently on the encode side, mirroring the decode side's bounds. (Already covered by the previous commit's test; this round's finding was about a related edge you flagged - see the constant-based bound already in place from the prior round.)

**2. The `LocalDateTime` default question - this was the important one.** You were right to flag it: `java.util.Date` is *not* actually `Type.DATETIME`'s default runtime representation - `LocalDateTime` is (`GlobalConfiguration.DATE_TIME_IMPLEMENTATION` defaults to it). Tracing that down surfaced two real bugs my original fix didn't cover:

- `getTypeForValue` had **no case for a bare `LocalDate`** - `Type.DATE`'s actual default representation. A populated `DATE` column was falling all the way through to `VARCHAR` on *every* default-configured database, disagreeing with the schema path's `DATE`. This was a live bug this PR hadn't touched at all. Fixed with the missing branch, matching the existing `LocalDateTime → TIMESTAMP` case right next to it.
- The schema-context `isDeclaredAsDatetime` mechanism from the previous commit **only worked for a whole-entity projection** (`SELECT FROM Type`). A query that projects specific columns (`SELECT col FROM Type` - the shape of essentially every real client query) produces rows with no backing element, so `getDeclaredProperty` could never find the schema property for the common case. I caught this myself before you'd have had to: I added an end-to-end IT test that actually configures `dateTimeImplementation=java.util.Date` and runs a real `PreparedStatement.executeQuery()`, and it failed against my own previous commit. Root-caused it to `PostgresNetworkExecutor`'s `Describe(Portal)` path computing columns from the *already-executed* real result (narrow projection, no element) rather than the DESCRIBE-time full-entity sample. Fixed by threading the query's resolved FROM-target type through as a fallback source (`resolveQueryTargetType`), applied everywhere `getColumns()` is called with a parseable statement available.

`PostgresTypeResolutionPathTest`'s `SAMPLE_VALUES` now models what a default-configured database actually produces (`LocalDate`/`LocalDateTime`) instead of an assumed `Date`, which is what let this hide - the old fixture happened to paper over both gaps. The one genuinely irreducible disagreement (a `Date`-configured `DATETIME` column, where the value alone can't distinguish it from `DATE`) is now pinned in its own dedicated test rather than the generic `KNOWN_DISAGREEMENTS` map, which is empty now that every other type genuinely agrees.

Full `postgresw` suite (403 unit tests) green, plus the extended IT set (`PostgresWJdbcIT` 45, `PostgresProtocolIT` 75, `Issue6411ByteaIT`, `Issue5290ClientCompatibilityIT`, `Issue6412CatalogIT`, `Issue6377PreAuthTimeoutIT`, connection-pool and metrics ITs) - no regressions from the `getColumns()` signature change.

@claude please review this!

lvca:
Third review round found no blocking issues - thanks for the thorough hand-traced verification. Addressed the three minor observations:

1. **`dscale` bound asymmetry** - fixed in c901d6136: `parseNumericBinary` now rejects an out-of-range `dscale` at the same bound the encode side already rejects at (`NUMERIC_MAX_DIGITS * NUMERIC_DIGIT_WIDTH`), before it reaches `setScale()`. New test: `deserializeBinaryNumericRejectsOversizedScale`.

2. **Alias gap in `isDeclaredAsDatetime`/`getDeclaredListType`** (`SELECT whenHappened AS wh FROM Types6447`) - not fixing in this PR. You correctly noted it's a pre-existing limitation of the #5289 `getDeclaredListType` pattern, not a regression this PR introduces, and the blast radius is narrow: it only matters for a column that is (a) aliased or a projected expression, (b) typed `DATETIME`/`LIST`, and (c) on a database explicitly configured with the non-default `java.util.Date`/needs-declared-element-type representation. Resolving an alias back to its source property requires walking the SELECT projection's expression tree (to tell `col AS alias` from `expr(...) AS alias`, where the latter has no single source property at all), which is a meaningfully different and larger piece of work than this PR's scope. Agreed it's worth its own follow-up issue if it turns out to matter in practice - filing that is Luca's call, so flagging here rather than filing it myself.

3. **Broad `Exception` catch in `resolveQueryTargetType`** - kept as-is. This mirrors the existing `parseStatement()` in the same class, which catches `Exception` for the identical reason: this is a best-effort fallback for a cosmetic type-name improvement, never load-bearing for query correctness, so a failure here must never propagate and break serving the RowDescription. Narrowing to `SchemaException` would risk exactly that if some other exception type reached it in an edge case I haven't enumerated - broad-and-safe beats narrow-and-fragile for this one.

Full `postgresw` suite (404 unit tests) green, plus `Issue6447TypeResolutionIT` re-verified.

@claude please review this!

lvca:
Fourth review round found a real one - fixed all 4 in 988abc41a:

1. **Extreme-negative-scale DoS in `putNumericBinary`** - the bound check ran on the scale *after* `setScale(0)` had already normalized it, but that normalization is what actually materializes `unscaledValue * 10^|scale|` as a real `BigInteger`, unlike the O(1) `BigDecimal(String)` constructor a plain SQL literal like `'1E2000000000'` goes through to produce such a value in the first place (DECIMAL has no configured precision limit). Moved the bound check to run on the raw scale first, before any materialization can start. New regression test uses `@Timeout` as a hang/OOM detector (not a latency bound, per this repo's stall-aware testing convention) with an extreme negative scale that would previously have tried to allocate a multi-GB `BigInteger`.

2. **Redundant/wasted reparse on the simple-query path** - now parses once (gated on `language == "sql"`, since `parseStatement` only understands SQL and would otherwise be a guaranteed parse-and-fail on every Cypher/Gremlin/Mongo/GraphQL query sent this way) and reuses the same `Statement` for both the schema-fallback and the target-type resolution, the same way the extended-protocol call sites already reuse `portal.sqlStatement`.

3. **Exception-driven control flow in `resolveQueryTargetType`** - swapped `Schema.getType()` (throws for the very common case of an identifier that isn't a registered type) for `Schema.getTypeOrNull()`, which already exists and returns `null` directly.

4. **Raw `ArithmeticException` from `parseNumericBinary`** - a malformed payload whose `dscale` doesn't match the precision its `ndigits`/`weight` actually encode now surfaces as `PostgresProtocolException`, consistent with every other rejection in this codec. New test constructs exactly that payload (`0.9999` encoded with a declared `dscale=0`).

Full `postgresw` suite (406 unit tests) green, plus `Issue6447TypeResolutionIT`, `PostgresWJdbcIT` (45), `PostgresProtocolIT` (75, heavy simple-protocol coverage - directly relevant to fix #2), `Issue6412CatalogIT`, `Issue5290ClientCompatibilityIT`, `Issue6411ByteaIT` - no regressions.

@claude please review this!

lvca:
Fifth review round - fixed the real finding in 3477a9f28.

**Confirmed and fixed the performance regression**: verified against `SQLParser.g4` that there's no bare `SET` or `SHOW` statement production (only `SET GLOBAL \$var = expr`, and `SET`/`SHOW` as clauses inside INSERT/UPDATE) - so the previous commit's \"parse once, gated on language\" fix was still computing that parse *before* the `SET`/`SAVEPOINT`/`RELEASE`/`ROLLBACK TO`/`SHOW`/system-query/`BEGIN` dispatch, meaning every one of those statements paid for a guaranteed parse-and-fail. Moved the parse into the single branch that actually reaches `database.command(...)` and can use the result - those special-cased statements now never call `parseStatement` at all, matching pre-PR behavior for `SET`/`SHOW` (the review correctly noted these were new victims; `BEGIN`/`SAVEPOINT`/`ROLLBACK TO` already paid a similar cost pre-PR since they produced empty result sets, so that part isn't new).

Also added the suggested regression test: a new IT test forces pgjdbc's simple query protocol (`preferQueryMode=simple`) for the `java.util.Date`-configured DATETIME scenario, so the narrow-projection schema-fallback is now verified over both wire protocols, not just the extended one every real JDBC/ORM client defaults to.

Full `postgresw` suite (406 unit tests) green, plus the extended IT set including `PostgresWJdbcIT` (45), `PostgresProtocolIT` (75, heavy `SET`/`SHOW`/simple-protocol coverage), `Issue6447TypeResolutionIT` (6, including the new simple-protocol test), `Issue5290ClientCompatibilityIT`, `Issue6412CatalogIT`, `Issue6377PreAuthTimeoutIT`, `TomcatConnectionPoolPostgresWJdbcIT` - no regressions.

The two "minor/non-blocking" notes from this round (aliased-column resolution gap, already-documented follow-up territory from round 3; and the narrow-projection test-coverage gap, now closed above) don't need further action.

@claude please review this!

lvca:
Sixth review round - fixed the real finding and the two minor ones in dbbb022bc.

1. **Unbounded `weight` in `parseNumericBinary`** - correct catch. `ndigits`, `dscale` and `sign` were each individually bounded, but `weight` wasn't, and a crafted payload with an extreme (still in-range) `weight` combined with a small `ndigits` and the max-allowed `dscale` derives a `scale` far apart from `dscale`, which the final `setScale` would materialize a BigInteger to bridge. Smaller blast radius than the encode-side fix earlier in this review chain (weight is capped by its 16-bit wire width, so tens of KB rather than multi-GB), but the same class of bug on a path that has to treat its input as attacker-controlled. Bounded to the same range the rest of the codec uses. New test: `deserializeBinaryNumericRejectsOversizedWeight`.

2. **Redundant per-call `resolveQueryTargetType` resolution** - memoized on `PostgresPortal` (`queryTargetType`/`queryTargetTypeResolved` fields), since the answer can't change across a portal's lifetime but the call sites (`describeCommand`, `executeCommand` x2) can all fire for the same portal, and `executeCommand` in particular can fire repeatedly for a cursor-based fetch with a `LIMIT`.

3. **Aliasing limitation** - added the one-line doc note you suggested to `getDeclaredProperty`'s javadoc, explaining it's a pre-existing #5289-shared limitation rather than something new here.

4. **Double-parse on the simple-query path** - no action needed; you correctly noted this was already a deliberate, well-justified trade-off from the previous round.

Full `postgresw` suite (407 unit tests) green, plus `Issue6447TypeResolutionIT` (6), `PostgresWJdbcIT` (45), `PostgresProtocolIT` (75), `Issue6412CatalogIT`, `Issue5290ClientCompatibilityIT`, `Issue6411ByteaIT` - no regressions, including the cursor-based-fetch paths that exercise the new memoization repeatedly.

@claude please review this!

lvca:
Seventh review round found one real gap - fixed in 2ac66f046.

**Unbounded magnitude in `putNumericBinary`, only scale was bounded.** Correct catch, and it's exactly the gap the review called out: the scale check closes the scientific-notation amplification vector (`'1E2000000000'`, which always implies a very negative scale), but a value that's simply a great many literal digits at a small/zero scale sailed past both scale checks and reached `unscaled.toString()` - O(n) on the digit count - before the existing `ndigits > NUMERIC_MAX_DIGITS` check (further down, *after* that work) ever got a chance to reject it. Fixed by checking `unscaled.bitLength()` (already computed, effectively O(1) - unlike `toString()`) against the same digit-count-derived bound the rest of the codec uses, right where `unscaled` is first obtained. New test builds the oversized value via `BigInteger.ONE.shiftLeft(...)` - itself O(1), same trick as the scale-case test - with `@Timeout` as the hang/OOM detector.

**Re: your question about JOINs** - confirmed, and worth stating explicitly since you asked directly: yes, `resolveQueryTargetType` only resolves a single simple `FROM <identifier>` target, so for `SELECT a.col, b.col FROM A JOIN B ...` the fallback just returns null and that column falls back to the pre-existing value-only guess - same as the alias case, not a regression, not something this PR claims to close. Given both are real query shapes a client could hit, if either turns out to matter in practice they're natural candidates for the same follow-up issue (resolving `getDeclaredProperty` from something richer than a single-identifier FROM target). Not blocking this PR, which fixes the four specific type-path disagreements #6447 was filed for.

Full `postgresw` suite (408 unit tests) green, plus `Issue6447TypeResolutionIT` and `Issue6411ByteaIT` re-verified.

@claude please review this!

lvca:
Eighth review round found no correctness bug and explicitly called this "closed out... pretty thoroughly" - a real convergence signal after seven rounds that each found something (rounds 4/6/7 specifically hardened the NUMERIC codec against three distinct DoS-adjacent gaps: scale, weight, and magnitude).

Addressed the one actionable (non-blocking) suggestion in a7013f3d0: added a note to the release notes explaining that `NUMERIC_MAX_DIGITS` (16,000 total decimal digits) is a deliberate DoS tradeoff well under real PostgreSQL's own limit (131,072 integer / 16,383 fractional digits), not a technical ceiling - since `DECIMAL` has no configured precision/scale limit in ArcadeDB, a client could in principle hit this cap on a value real Postgres would happily store. Docs-only change, no code touched.

The other two notes (comment density, the pre-existing unused `PostgresType.cls` field) were explicitly flagged as non-blocking/pre-existing/not-to-fix-here by the reviewer itself.

This PR is now, I believe, ready to merge: four real type-resolution bugs fixed (SHORT/BYTE, DATE, DATETIME, DECIMAL), a real NUMERIC binary codec built and hardened across encode and decode against every bounds/DoS gap eight review rounds could find, a structural fix so the schema-fallback mechanism actually works for the common narrow-projection query shape (not just whole-entity), and a performance fix so simple-protocol connection setup doesn't pay for a guaranteed parse-and-fail. Full `postgresw` suite (408 unit tests) plus the extended IT set green throughout, no regressions.

@claude please review this!

lvca:
Addressed the three non-blocking polish notes from the last review in 197f7fd4d (bitLength in the magnitude-rejection message, \`BigDecimal::new\` for style consistency, javadoc wording precision on the alias-limitation note). No behavioral change, all existing tests still pass.

Given the last two review rounds found no correctness issues (round 8 explicitly: "None of these block the PR - they're polish-level notes on an already well-hardened change"), I'm treating this PR as ready pending final CI. Not re-triggering another full review round for a pure-polish commit - will confirm CI is green and report back.

PR #6454, opened 2026-08-19
fix(#6372): Console LOAD no longer echoes an empty prompt for blank lines

### LINKED ISSUE #6372: Console: empty lines in a script emit empty prompts (minor)
`ArcadeDB Server v26.8.1 (build 87bdc67f1f0331fa2d07e932a550064c118eae70/1785791531932/UNKNOWN)`

`Running on Linux 7.0.0-28-generic - OpenJDK 64-Bit Server VM 25.0.3`

An empty line in SQL script loaded from the console emits an empty prompts on a new line.

This is the same behavior as for stdin: when the user presses Enter a new prompt in a new line is started.
In a script it would be better for the logs if empty newlines do not cause a new prompt. So if it is doable without too much changes, empty lines should be ignored when `LOAD` is used in the console.

For example, given a file `test.sql` in the base folder:

```sql
CREATE DATABASE temp;


SELECT 1;

SELECT 2
```

executed with:

```shell
bin/console.sh -b "test.sql"
```

results in:

```
WARNING: Using incubator modules: jdk.incubator.vector
ArcadeDB Console v26.8.1 - Copyrights (c) 2021 Arcade Data Ltd (https://arcadedb.com)? ArcadeDB is free and open source, sustained by its community. If it helps you, please consider sponsoring: https://github.com/sponsors/ArcadeData
> LOAD test.sql

Executing commands from file test.sql...
> CREATE DATABASE temp
Database 'temp' created
{temp}> 

{temp}> 

{temp}> SELECT 1
+----+-----+
|NAME|VALUE|
+----+-----+
|1   |1    |
+----+-----+
Command executed in 228ms
{temp}> 

{temp}> SELECT 2

+----+-----+
|NAME|VALUE|
+----+-----+
|2   |2    |
+----+-----+
Command executed in 1ms
File processed in 0 seconds
```

See the empty `{temp}>` prompts.

* PS: While at it, depending on if a command is terminated by a semicolon the result is prefixed with an extra newline or not, see `SELECT 1;` vs `SELECT 2`.

* PPS: The heart emoji (`❤`) in the `SPONSOR_MSG` part of the banner results in a question mark in the console (`... Arcade Data Ltd (https://arcadedb.com)? ArcadeDB ...`).

robfrank:
Closes #6372

## Summary

`Console.executeLoad()` reads a script file line by line and hands each line to `TerminalParser.parse()`, which splits it into "words" on the `;` delimiter. An empty line in the script (or the leftover line terminator right after a semicolon-terminated command) parses into a single blank word made only of the newline character.

`Console.execute(ParsedLine, boolean printCommand)` iterated over every word and, when `printCommand` is `true` (the mode `executeLoad` uses), unconditionally echoed `getPrompt() + w` before executing it, including for these blank words. Since `getPrompt()` itself starts with a newline, each blank word produced an extra `{db}>` prompt line with nothing after it, exactly as reported in the issue.

The same untrimmed-word echo also explains the "PS" note in the issue: the last command in a script with no trailing semicolon carries its own trailing newline as part of the word, so its prompt echo (`getPrompt() + w`) contained an embedded newline that pushed the result table down by one extra blank line compared to a semicolon-terminated command.

### Fix

In `Console.execute(ParsedLine, boolean)`, each word is now trimmed first. A word that trims to empty is skipped entirely (no prompt echo, no execution - `execute(String)` already no-ops on blank input, so behavior is unchanged other than the suppressed echo). Non-blank words are echoed using the trimmed value, which also removes the stray embedded newline for an unterminated last command.

Interactive mode is unaffected: its prompt is printed by `LineReader.readLine(getPrompt())` before the line is even read, not through this method, so pressing Enter still starts a fresh prompt line exactly as before.

The heart emoji rendering as `?` in the console banner (the "PPS" note in the issue) is a separate terminal-charset concern, out of scope for this fix.

## Test plan

- [x] New regression test `ConsoleTest#loadScriptWithEmptyLinesDoesNotEchoEmptyPrompts`: loads a script with 3 real commands separated by empty lines and asserts exactly 3 prompts are echoed (previously 8), and that all rows were actually inserted despite the blank lines. Fails before the fix (`expected: 3 but was: 8`), passes after.
- [x] `mvn -o -pl console test`: full `ConsoleTest` suite green (40/40), plus `ConsoleBatchTest`, `TerminalParserTest`, `CompositeIndexUpdateTest` all green. `ConsoleAsyncInsertTest` fails with `SecurityException: User/Password not valid`, but this is a pre-existing environmental issue unrelated to this change - a brew-installed ArcadeDB server was already bound to port 2480 on the test machine, which is a documented false-failure signature for that test, not a regression from this fix.

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -811,8 +811,16 @@ public class Console {
       return true;
 
     for (final String w : parsedLine.words()) {
+      final String trimmedWord = w.trim();
+      if (trimmedWord.isEmpty())
+        // AN EMPTY LINE (OR A LEFTOVER LINE TERMINATOR BETWEEN TWO COMMANDS) SPLITS INTO A BLANK "WORD": SKIP IT SO
+        // LOAD DOES NOT ECHO AN EMPTY PROMPT FOR IT (issue #6372)
+        continue;
+
       if (printCommand)
-        output(3, getPrompt() + w);
+        // USE THE TRIMMED WORD: AN UNTERMINATED LAST COMMAND CARRIES ITS TRAILING NEWLINE AS PART OF THE WORD, WHICH
+        // WOULD OTHERWISE PUSH THE RESULT DOWN BY AN EXTRA BLANK LINE (issue #6372)
+        output(3, getPrompt() + trimmedWord);
 
       if (batchMode) {
         try {
--- a/console/src/test/java/com/arcadedb/console/ConsoleTest.java
+++ b/console/src/test/java/com/arcadedb/console/ConsoleTest.java
@@ -184,6 +184,49 @@ class ConsoleTest {
     }
   }
 
+  /**
+   * Issue https://github.com/ArcadeData/arcadedb/issues/6372: an empty line in a script loaded with LOAD must not echo an
+   * empty prompt of its own - only the real commands are echoed.
+   */
+  @Test
+  void loadScriptWithEmptyLinesDoesNotEchoEmptyPrompts() throws Exception {
+    assertThat(console.parse("connect " + DB_NAME)).isTrue();
+
+    final File script = new File("./target/issue-6372.sql");
+    try {
+      Files.writeString(script.toPath(), """
+          CREATE DOCUMENT TYPE Loaded;
+
+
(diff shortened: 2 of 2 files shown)


robfrank:
@claude review

Issue #5444, opened 2026-07-26
Cypher: composite-index NodeIndexSeek returns zero rows, and RETURN * leaks internal anonymous variables


lvca:
Two independent defects in the native Cypher engine, both found while reviewing #5393 and both unrelated to that PR. Filed together because they are cheap to fix in one pass.

## 1. `NodeIndexSeek` on a composite index silently returns zero rows

The optimizer picks a composite index for a single-property equality predicate, then seeks it with a one-element key. The engine requires the full key, so the seek yields nothing and the query returns no rows - silently, no error.

```java
database.getSchema().createVertexType("Area").createProperty("id", Type.STRING);
database.getSchema().getType("Area").createProperty("zone", Type.STRING);
// the type's ONLY index is composite
database.getSchema().createTypeIndex(Schema.INDEX_TYPE.LSM_TREE, true, "Area", "id", "zone");
// ... insert a vertex with id='country', zone='EU'
```

```cypher
MATCH (c:Area) WHERE c.id = 'country' RETURN c.id                        -- []   WRONG, expected ['country']
MATCH (c:Area) WHERE c.id = 'country' AND c.zone = 'EU' RETURN c.id      -- []   WRONG, expected ['country']
```

The equivalent SQL is correct:

```sql
SELECT id FROM Area WHERE id = 'country' AND zone = 'EU'                 -- [country]
```

`PROFILE` shows the plan seeking with a partial key:

```
+ Filter [predicate=(c.id = 'country' AND c.zone = 'EU'), cost=5.00, rows=0]
  + NodeIndexSeek(c:Area) [index=Area[id,zone], id=country, cost=5.00, rows=0]
```

The seek is built at `NodeIndexSeek.java:184`:

```java
cursor = index.get(new Object[] { seekKeys.get(seekIndex++) });
```

always one element, regardless of index arity. Calling the same thing directly against the index makes the mismatch explicit:

```
java.lang.IllegalArgumentException: key is composed of 1 items, while the index defined 2 items
    at com.arcadedb.index.lsm.LSMTreeIndexAbstract.lookupInPage(LSMTreeIndexAbstract.java:267)
```

so the wrong-answer path is swallowing that condition somewhere between the operator and the index.

Severity: high. Any Cypher equality query against a type whose only index is composite returns nothing. There is no error and no warning; the user sees an empty result set.

Both halves need attention:
[...]

PR #6480, opened 2026-08-19
fix(#6397): openCypher label disjunction now uses per-root index seeks

### LINKED ISSUE #6397: openCypher: a label disjunction never uses an index, so `(n:A|B {id: $x})` full-scans every alternative where per-type seeks would resolve it
Found while verifying #6363 (merged as `7912cc75e`). Pre-existing; #6363 made it visible rather than caused it.

A node pattern with a label disjunction is always executed as `NodeByLabelDisjunctionScan`, a full scan of every vertex type any alternative accepts. `IndexSelectionRule.createAnchorOperator` builds that operator for a disjunction anchor **before** it looks at `anchor.useIndex()`, so an indexed equality predicate on the pattern cannot drive a seek - no matter how selective it is.

## Measured

Two types, each with a UNIQUE index on `id`, 500 records each:

```cypher
MATCH (n:A {id:'a1'}) RETURN n.id
```
```
+ Filter [predicate=n.id = a1, cost=5.00, rows=0]
  + NodeIndexSeek(n:A) [index=A[id], id=a1, cost=5.00, rows=0]
```

```cypher
MATCH (n:A|B {id:'a1'}) RETURN n.id
```
```
+ Filter [predicate=n.id = a1, cost=1500.00, rows=500]
  + NodeByLabelDisjunctionScan(n:A|B) [cost=1000.00, rows=1000]
```

Both return `a1`. The second reads 1000 records and discards 999, where two unique-index seeks would resolve it in two lookups. The gap grows linearly with the data.

Before #6363 the disjunction anchor was *also* mis-costed - the estimate came from the first alternative alone, so the plan claimed a cheapness it never had. #6363 made the estimate honest (`cost=1000` for a 1000-row scan), which is what makes the missing capability legible in `EXPLAIN`. The execution has always been the full scan.

## Suggested shape

A disjunction is a union of per-alternative scans, so it can be a union of per-alternative **seeks** whenever every type the disjunction selects has a usable index for the predicate:

- `Labels.matchingVertexTypes` already returns exactly the types the scan visits, so the planner can ask each one for an index on the predicate's property.
- All-or-nothing is the simple rule: if every selected type can seek, emit a seek per type and concatenate; if any cannot, keep today's scan. Mixed (seek the indexed types, scan the rest) is strictly better but needs the operator to interleave two access paths, so it is worth doing second if at all.
- The cost estimate should then be the sum of the per-type seek costs, which `getMatchingVertexCardinality` is already the right place to extend.
- Deduplication is already handled: `matchingVertexTypes` selects each subtype once and the types are iterated non-polymorphically, so a node answering to two alternatives is visited once, and that property must be preserved by a seek-based operator.

Worth noting the same limitation applies to the IN-list and range-predicate paths in `AnchorSelector`, which a disjunction also never reaches.

A regression test should pin the plan through `EXPLAIN` (as `CypherDisjunctionCardinalityIssue6363Test` does) and assert the result set is unchanged and duplicate-free across a type that satisfies more than one alternative.

robfrank:
Closes #6397

## Problem

A node pattern with a label disjunction (`(n:A|B {id: $x})`) always executed as `NodeByLabelDisjunctionScan`, a full scan of every vertex type any alternative accepts. `IndexSelectionRule.createAnchorOperator` built that operator **before** looking at `anchor.useIndex()`, so an indexed equality predicate on the pattern could never drive a seek, no matter how selective it was. `AnchorSelector` had the matching gap on the costing side.

## Fix

All-or-nothing per-root index seek, following the issue's suggested shape:

1. **`StatisticsProvider.getMatchingVertexRootTypes(labels, disjunction)`** (new) returns the subset of `Labels.matchingVertexTypes` that has no ancestor also in that set - the minimal list of types a *polymorphic* index seek has to visit to cover the whole disjunction exactly once. `matches()` propagates down the whole subtype chain, so a root's own polymorphic index already reaches every listed descendant of that root.
2. **`AnchorSelector.tryDisjunctionIndexSeek`** (new) loops over every equality predicate on the disjunction's own variable (inline property or `WHERE` clause) - not just the first one a `HashMap` happens to iterate to - and asks each root type for a usable index on it. All-or-nothing per predicate: one root with no usable index moves on to the next candidate predicate; only when none resolves on every root does it fall back to the existing full-scan costing. The selectivity estimate honors `wholeKey` the same way the plain (non-disjunction) equality path does, so a composite index that only has its leading column bound is costed as a prefix range, not a one-row point lookup.
3. **`AnchorSelection`** gained an optional `List<DisjunctionIndexSeek>` (type name + `IndexStatistics` + per-root prefix `keyValues`), so a disjunction anchor can carry N per-root seeks through the same object the rest of the optimizer already threads.
4. **`IndexSelectionRule.createAnchorOperator`** builds one `NodeIndexSeek` per root (reusing the existing seek operator unmodified, including its composite-index prefix handling from #5444) and wraps them in a new **`NodeByLabelDisjunctionIndexSeek`** operator that unions the per-root result streams and de-duplicates by RID.
5. **`CypherOptimizer.extractTypeNames`** collected only `node.getFirstLabel()` for statistics purposes, so a disjunction's second and later alternatives never had their indexes collected - a real index on a later alternative silently read back as "no index". Fixed to collect every label of a disjunction node.

### Why de-duplication is required, not merely defensive

A `TypeIndex` is inherently polymorphic - creating an index on a type indexes that type's buckets **and every subtype's buckets** (`TypeIndexBuilder.create()` uses `type.getBuckets(true)`). Seeking a root's own index already returns every matching descendant, and the roots are chosen so their subtrees are disjoint under single inheritance. But ArcadeDB supports multiple inheritance (`EXTENDS A, B`), so a type that multiply-inherits from two accepted alternatives is reachable through *both* roots' indexes - the existing test `CypherDisjunctionCardinalityIssue6363Test.aSubtypeIsCountedOnceAndOnlyThroughTheAlternativesThatAcceptIt` already exercises exactly this shape for the scan operator, so the new seek path is held to the same invariant (and has its own regression test for it).

### Scope

- Only a single equality predicate drives the seek (inline property or `WHERE var.prop = value`), matching the issue's primary "Measured" example. The issue itself calls out the IN-list and range-predicate paths on a disjunction anchor as a related but separate, not-yet-addressed limitation ("worth noting ... which a disjunction also never reaches") - not something this fix closes. A mixed seek/scan union (some roots indexed, some not) is deliberately left alone too, per the issue's own note that it "needs the operator to interleave two access paths ... worth doing second if at all".
- **This fix applies only to a disjunction anchor with no incident relationship** - `MATCH (n:A|B {id: $x}) RETURN n`-shaped queries. `CypherExecutionPlanner.shouldUseOptimizer()` has a pre-existing, unrelated gate that bails the *whole statement* out of the cost-based optimizer whenever a label-disjunction node has an incident relationship (`ExpandAll`/`ExpandInto` each carry a single `targetLabel` and can't represent OR semantics), so `MATCH (n:A|B {id: $x})-[:REL]->(m)` never reaches `AnchorSelector` at all and stays on the legacy scan-only path. That gate predates this PR and this fix doesn't touch it - tracked separately as #6482, since closing it is a materially larger change (either admitting anchor-only disjunctions past the gate, or teaching the hop operators to represent a disjunction target).

## Test plan

New test: `engine/src/test/java/com/arcadedb/query/opencypher/CypherDisjunctionIndexSeekIssue6397Test.java`
- [x] inline-property equality (`{id: 'a1'}`) on every indexed alternative produces `NodeByLabelDisjunctionIndexSeek`, not `NodeByLabelDisjunctionScan`, and returns the right row
- [x] `WHERE n.id = 'x'` equality form does too
- [x] one non-indexed alternative falls back to the full scan (all-or-nothing) and still returns correct rows for both the indexed and the non-indexed side
- [x] a value no alternative has returns nothing
- [x] a type multiply-inheriting from two indexed alternatives (diamond) is returned once, not twice
- [x] a second, non-indexed predicate does not block the seek on an indexed one regardless of hash-iteration order
- [x] a composite index on every root resolves the full key, not just a leading-column prefix
- [x] a prefix-only composite seek is costed less selectively than a fully-resolved one
- [x] `EXPLAIN` on a UNION branch whose own plan is a disjunction index seek doesn't NPE (regression for a bug found in review - see cycle history below)

Regression (all green):
- [x] `CypherDisjunctionIndexSeekIssue6397Test` (new) - 9/9
- [x] `CypherDisjunctionCardinalityIssue6363Test` - 5/5 (the diamond/multi-inheritance cardinality test this PR's dedup logic is held to)
- [x] `CypherLabelDisjunctionOnExpandedNodeIssue6338Test` - 11/11
- [x] `CypherRelationshipTypeDisjunctionTest` - 13/13
- [x] `Issue5362DuplicatePredicateIndexSeekTest` - 9/9
- [x] `Issue5444CompositeIndexSeekTest` - 10/10 (composite-index prefix seek reuse)
- [x] `StatisticsProviderTest` - 22/22 (4 new cases directly covering `getMatchingVertexRootTypes`)
- [x] `CostModelTest` - 16/16
- [x] `CypherOptimizerIntegrationTest` - 7/7
- [x] `AnchorSelectorTest` - 11/11
- [x] `IndexSelectionRuleTest` - 11/11
- [x] `OpenCypherOptimizerVerificationTest` - 9/9
- [x] `CypherPlannerDifferentialIssue6400Test` - 4/4
- [x] Full `com.arcadedb.query.opencypher.**` package sweep (excluding benchmark/slow/vector lanes): **8421 tests, 0 failures, 0 errors**

## Review cycle history

Four rounds of automated review, each addressed in its own commit:
1. Found (and fixed) a real, reachable `AnchorSelection.toString()` NPE reachable via `EXPLAIN` on a UNION branch; the disjunction seek only trying the first equality predicate regardless of hash-iteration order; the per-root seek not propagating composite-index prefix values.
2. Three non-blocking nice-to-haves: a duplicated cardinality computation, a documentation note on `WorkGuard` delegation, and direct unit coverage for `getMatchingVertexRootTypes`.
3. A genuine cost-estimation gap: unique-index selectivity wasn't checking `wholeKey`, and per-root cardinality could floor to 0 via integer division.
4. Verified-but-out-of-scope finding: the fix is inert whenever the disjunction node has an incident relationship, due to a pre-existing, unrelated optimizer-eligibility gate. Filed as #6482 rather than folded into this PR, since closing it is a separate, materially larger change. Two other minor/non-blocking nits from this round (a cosmetic per-root row-sum-can-exceed-total display artifact, and a suggestion to add a heterogeneous-index-shape test) are noted here as deferred rather than actioned, since this was the fourth and final scheduled review cycle.


robfrank:
Thanks for the review — all three points addressed in b9c24a73f5:

**1. `toString()` NPE (blocking)** — confirmed exactly as traced: `EXPLAIN` on a UNION branch whose own plan is a disjunction index seek goes through `appendPlanDescription` → `appendStepChain` → `AbstractExecutionStep.prettyPrint` → `PhysicalPlan.explain()` → `anchor.toString()`, which called `index.getIndexName()` on a null `index`. Fixed by branching `toString()` on `isDisjunctionIndexSeek()`. Added `explainOnAUnionBranchWithADisjunctionSeekAnchorDoesNotThrow`, which reproduces the exact path — I verified it fails with the pre-fix `toString()` (same NPE, same stack trace as your trace) and passes with the fix.

On the "5/5 claimed but should NPE" question: PROFILE and top-level EXPLAIN both render the anchor through the always-safe `PhysicalOperator.explain(0)` operator tree (`OpenCypherExplainExecutionPlan.prettyPrint()` just returns a pre-built string), so my original test's `PROFILE`-based `profilePlan()` helper genuinely never hit the buggy `anchor.toString()` call — only a UNION branch's `appendStepChain` does. The 5/5 was accurate for what it tested; it just didn't test the path you found. Good catch.

**2. Only first predicate tried** — `tryDisjunctionIndexSeek` now loops over every candidate equality predicate (mirroring `evaluateNode`'s loop) instead of taking `HashMap` iteration's first entry, trying each until one is indexed on every root. Added `aSecondNonIndexedPredicateDoesNotBlockTheSeekOnTheIndexedOne`.

**3. Composite-index prefix not propagated** — `AnchorSelection.DisjunctionIndexSeek` now carries its own per-root `keyValues` (via the existing `collectKeyPrefixValues`), and `IndexSelectionRule` passes them through instead of a hardcoded single-value list. `NodeByLabelDisjunctionIndexSeek.explain()` now nests each root seek's own `explain()` so the resolved key columns show up in EXPLAIN/PROFILE. Added `compositeIndexOnEveryRootResolvesTheFullKeyNotJustAPrefix`, which asserts both `id=e5` and `tag=t5` appear in the plan and that a mismatched second column is rejected.

Full `com.arcadedb.query.opencypher.**` sweep (excluding benchmark/slow/vector lanes) after these changes: 8416 tests, 0 failures, 0 errors.

robfrank:
Thanks — applied the 3 non-blocking suggestions in 07aab5fc10:

1. **Duplicated cardinality call** — `evaluateDisjunctionNode` now computes `getMatchingVertexCardinality` once and passes it into `tryDisjunctionIndexSeek`.
2. **WorkGuard delegation comment** — added, explaining the outer loop always delegates into a child `NodeIndexSeek`'s own already-guarded `hasNext()`/`next()`.
3. **Unit test gap** — added direct `StatisticsProviderTest` coverage for `getMatchingVertexRootTypes` (two unrelated roots, a subtype excluded via its ancestor, the diamond multi-inheritance case returning both parent roots, and an unknown label returning empty).

I looked at extending `AnchorSelectorTest`/`IndexSelectionRuleTest` too, but its `MockStatisticsProvider` is constructed with `super(null)` (no real database) specifically so those tests don't need a schema — and `getMatchingVertexRootTypes`/`getMatchingVertexCardinality` both guard on `database == null` and return empty/zero. Exercising `tryDisjunctionIndexSeek` meaningfully needs real `DocumentType` instances from an actual schema (for the root-type list), which is exactly what `CypherDisjunctionIndexSeekIssue6397Test` and the new `StatisticsProviderTest` cases already cover end to end. Extending the mock to fake `DocumentType` objects felt like it would duplicate that coverage for not much signal, so I left it - happy to revisit if you'd still like it.

Full `com.arcadedb.query.opencypher.**` sweep after these changes: 8420 tests, 0 failures, 0 errors.

robfrank:
Thanks — this was a legitimate cost-estimation gap, fixed in d21e802eec:

**Selectivity didn't check wholeKey** — `tryDisjunctionIndexSeek` now tracks the same `wholeKey` condition `evaluateNode`'s equality branch already uses: a unique index only gets the 0.001 "resolves to ~one row" selectivity when every column of its key was actually pinned, not merely because the index happens to be unique. Added `aPrefixOnlyCompositeSeekIsCostedLessSelectivelyThanAFullyResolvedOne`, which compares the reported row estimate between a fully-resolved composite key (`id`+`tag`, both bound → rows=2) and a prefix-only seek on the same data (`id` alone → rows=4) — verified it fails (2 instead of 4) without the fix, passes with it.

**Cardinality floor** — `IndexSelectionRule`'s even split across per-root `NodeIndexSeek`s now floors at `Math.max(1, ...)`, since every root that reaches that point has its own usable index and returns at least one row on a hit.

Full `com.arcadedb.query.opencypher.**` sweep: 8421 tests, 0 failures, 0 errors.

robfrank:
Thanks for this one — verified and confirmed accurate.

**Disjunction + relationship gap** — checked `CypherExecutionPlanner.shouldUseOptimizer()` directly and confirmed it exactly as described: the pre-existing gate bails the whole statement out of the cost-based optimizer whenever a label-disjunction node has an incident relationship (`if (node.getLabels().size() > 1) { if (!node.isLabelDisjunction()) return false; if (path.getRelationshipCount() > 0) return false; }`), so `MATCH (n:A|B {id: $x})-[:REL]->(m)` never reaches `AnchorSelector`/`IndexSelectionRule` and stays on the legacy scan-only path. Confirmed this isn't a regression (the gate predates this PR) and that closing it is a separate, materially larger change - either admitting anchor-only disjunctions past the gate, or teaching `ExpandAll`/`ExpandInto` to represent a disjunction target on the far side of a hop.

Filed as **#6482** rather than folded into this PR. Updated the PR description's Scope section to state the limitation precisely, so the title/description no longer read as a fully general fix, and added a "Review cycle history" section summarizing all four rounds.

**Two other minor/non-blocking nits from this round** (the per-root row-sum-can-exceed-total cosmetic display artifact, and a suggestion to add a test with heterogeneous index shapes across roots) - noted in the PR description as deferred rather than actioned, since this was the fourth and final scheduled automated review cycle. Happy to pick either up if a maintainer wants them addressed before merge.

Also removed the standalone tracking doc I'd committed early on (`docs/6397-*.md`) - its content duplicated what's now in the PR description, which is the right place for it per this repo's convention.

Full `com.arcadedb.query.opencypher.**` sweep after all four cycles: 8421 tests, 0 failures, 0 errors. This PR is ready for maintainer review/merge - merge itself is left to the repo owner as always.

Issue #6281, opened 2026-08-17
Follow-ups from #6259: two engine tests red on the PR's CI but green on main, and the per-database flush queue


lvca:
Collected while fixing #6259 (merged as PR #6269, `d6da8280a`). Three independent items plus one open question that spans the first two.

---

## Item 1 - `Issue5279ConcurrentInsertTest.sustainedConcurrentInsertsOnASingleBucketNeverConflict` asserts two things that cannot both hold

Observed on PR #6269's `unit-tests` lane:

```
expected: 2001L
 but was: 2000L
	at Issue5279ConcurrentInsertTest.lambda$sustainedConcurrentInsertsOnASingleBucketNeverConflict$6(Issue5279ConcurrentInsertTest.java:193)
```

The test runs 8 threads x 250 inserts at **`attempts=1`**, so a transaction that hits a `ConcurrentModificationException` is **not** retried - it is counted and its record is never written:

```java
try {
  database.transaction(() -> database.newDocument("Case").set("payload", payload).save(), true, 1);
} catch (final ConcurrentModificationException e) {
  conflicts.incrementAndGet();
}
```

It then tolerates a small residue of conflicts, and immediately afterwards demands the exact full count:

```java
assertThat(conflicts.get()).as("pure inserts must not raise concurrent modifications").isLessThanOrEqualTo(2);

database.transaction(
    () -> assertThat(database.countType("Case", false)).isEqualTo(1L + (long) threadCount * insertsPerThread));
```

Those two assertions contradict each other. Expecting 2001 while allowing up to 2 conflicts asserts both that a conflict may happen and that it may not, so **the run where the tolerance is actually used fails at the count** (2000 against 2001) instead of passing at the tolerance meant to absorb it. That is exactly the observed failure: one tolerated conflict.

**This is the same trap already fixed in the sibling class.** `Issue5279ConcurrentUpdateTest.concurrentEdgeCreationBetweenOwnVerticesNeverConflicts` was repaired in PR #6223 by counting the conflicts out, with a comment saying precisely this:

```java
// ...but the edge transaction runs at attempts=1, so a conflict tolerated above means that edge was NOT
// created. Counting the conflicts out is what makes the two assertions agree: expecting the full total while
// allowing one conflict asserts both that a conflict may happen and that it may not, and the run where one
// DOES happen then fails here (160 against 161) instead of at the tolerance meant to absorb it.
[...]

PR #6481, opened 2026-08-19
fix(#6440): async flush pipeline strands WAL acks across a closed/reopened database

### LINKED ISSUE #6440: GAVEligibilityTest is 16x slower inside the full suite than alone (3.6s -> 57s), and most classes are not
Measured while investigating why `GAVEligibilityTest` was the class an external kill kept landing in (fixed
separately in 6ffebccbb). It is not a slow test. It is a fast test that becomes slow when the rest of the suite has
run first, and it does not look like ordinary JIT/GC drift because most classes do not do this.

## Measurements

Same machine, idle, `mvn -o -pl engine -am test`. "In full suite" is the class's own `Time elapsed` line from a
complete `-DexcludedGroups=benchmark,slow,vector` run (12858 tests); "isolated" is the same class run alone.

| class | isolated | in full suite | ratio |
| --- | --- | --- | --- |
| `FullTextBM25Test` | 0.31s | 0.15s | 0.5x |
| `LSMTreeFullTextIndexTest` | 5.20s | 4.75s | 0.9x |
| `OpenCypherSubqueryTest` | 0.27s | 2.19s | **8x** |
| `GAVEligibilityTest` | 3.58s | 57.4s / 61.4s | **16x** |

The two index classes are not slower in the suite - one is faster, which is what a warm JIT normally buys. The two
opencypher classes are the outliers, and `GAVEligibilityTest` (42 tests, each building a `GraphAnalyticalView`) is
the extreme.

A third data point on how far this can go: in a run where three other Maven builds were running on the same
machine, `GAVEligibilityTest` took **998s** - 278x its isolated cost, and around 17 minutes of a single test class.
Machine contention explains part of that, but the other classes in that same run were not multiplied anywhere near
as much, so contention appears to amplify whatever the in-suite effect already is rather than being the whole of it.

## Why it matters

`forkCount` is 1, so the whole engine suite shares one JVM and one set of static/shared state. A class whose cost
depends on how much ran before it is a class whose cost grows as the suite grows, and this one sits on the
critical path of the default unit-test lane. It is also a wide window for anything external to land in: both times
a run of ours was lost, the kill landed inside this class, simply because it was occupying minutes rather than
seconds.

## What to look at

Nothing here is a diagnosis - the numbers are all that has been established. Candidates, cheapest first:

- A shared or dedicated thread pool that is saturated, or sized from `availableProcessors` at first touch and then
  contended, by the time these classes run. GAV building submits work; if the pool it lands on is busy or has been
  resized/exhausted by earlier tests, the builds serialise.
- Threads or executors leaked by earlier test classes and never shut down, so the JVM is progressively
  oversubscribed. `DedicatedThreadPoolTest` and the pool inventory in the `engine-concurrency` skill are the place
  to start.
- Accumulated static state (registries, caches, `GraphTraversalProviderRegistry`) that earlier tests grow and this
  class then walks.
- Heap pressure: 12858 tests in one JVM, and a GAV build allocates. Worth ruling out first with a GC log, since it
  is the cheapest to check and would explain a broad slowdown - though it would not explain why the two index
  classes are unaffected.

Reproducing is cheap: run the class alone, then run it after the opencypher package, then after the full suite, and
compare. If the number climbs with how much ran before it, that localises it to shared state rather than to the
class.

robfrank:
Closes #6440

## Summary

`GAVEligibilityTest` (and any test that repeatedly `drop()`s and `create()`s a database at the same path) paid up to `TransactionManager`'s full 2000 ms retry budget on almost every `close()`, which is why it went from ~2s isolated to 35-60+ seconds inside the full opencypher suite in the issue's own measurements.

**Root cause:** a page is durably written to WAL during commit (`WALFile.writeTransactionToFile` increments `WALFile.pagesToFlush`), then handed to the async flush pipeline; only the pipeline's eventual disk write releases the ack (`WALFile.notifyPageFlushed()`, via `MutablePage.takeWALFile()`). Several spots in the flush pipeline correctly abandon a page - there's nothing left to write it to - without ever releasing that ack:

- `PageManager.flushPage()`'s `!database.isOpen()` early return, and its `DatabaseIsClosedException` catch after the physical write already succeeded but the metadata update failed.
- `PageManagerFlushThread.flushPagesFromQueueToDisk()`'s several "database closed mid-batch" branches, and its `DatabaseMetadataException` catches (file dropped).
- `PageManagerFlushThread.resumeFlushing()`'s equivalents on the suspend/backup-resume path.

`LocalDatabase.closeDurableParts()` sets `open = false` *before* `TransactionManager.close()`'s retry loop runs, so any page still in flight at that instant is guaranteed to hit one of the paths above. `TransactionManager.close()` then waits its whole 2000 ms budget for a WAL counter nothing will ever decrement, before giving up and forcing the close anyway - for a `drop()`, where the files are about to be deleted regardless, that wait buys nothing but latency.

**A second, related bug in the same area:** `PageManagerFlushThread.removeAllPagesOfDatabase` and `FlushPageIndex.removeAllOfDatabase` matched pending pages with `Database.equals()`, which - like the rest of this class's bookkeeping - compares by **path**, not instance. A database closed and immediately reopened at the same path (exactly what a rapid `drop()`/`create()` cycle produces, e.g. a JUnit `@BeforeEach`/`@AfterEach` pair) could have its own still-queued batch wrongly discarded by a same-path sibling's belated cleanup. Fixed to match by reference identity, matching the precedent `PagesToFlush.slotCharged` already established for the sibling `slotsInUse` counter in #6281/#6303.

## Investigation

The issue itself was an open-ended performance investigation ("nothing here is a diagnosis"), so most of the work was root-causing rather than implementing:

1. Reproduced the reported slowdown by running the full `com.arcadedb.query.opencypher` package (8776 tests) and confirmed `GAVEligibilityTest` went from 1.9s isolated to 34-63s in-suite.
2. `jstack` sampling during the slow window showed the test's `main` thread parked in `TransactionManager.close()`'s WAL-flush retry loop, sleeping through its full 20 x 100 ms budget.
3. Bisecting the reproduction down showed it did **not** require the rest of the suite: a standalone loop that just creates a database at a fixed path, writes a few vertices, and drops it - nothing else running in the JVM - reproduced the same ~2000 ms stall roughly 1 time in 10.
4. Live-sampled the flush pipeline's own bookkeeping (`pageIndex`, `queue`, `slotsInUse`) during a stall: all reported nothing pending, while the WAL file's own counter stayed stuck - proving the stall was a stranded ack, not a genuinely slow or backed-up flush.
5. Instrumented `WALFile`'s increment/decrement pairing directly, which pinpointed `PageManager.flushPage()`'s `!database.isOpen()` early return as the primary leak, then found the remaining smaller leaks (`DatabaseMetadataException`/`DatabaseIsClosedException` catches, `resumeFlushing`) by symmetry with the one already-correct sibling (`removePagesOfFileFromBatch`, which already acks on the same conditions).

## Test plan

- [x] Added `Issue6440FlushAckOnClosedDatabaseTest`: white-box regression pinning that `PageManager.flushPage()` acks a page's WAL file when the database is already closed, and that a normal successful flush still acks exactly once (`takeWALFile()`'s exactly-once semantics).
- [x] Added `Issue6440CrossInstanceFlushDiscardTest`: white-box regression pinning that closing one database instance must not discard or evict a same-path *sibling* instance's still-pending pages/counters (`PageManagerFlushThread.removeAllPagesOfDatabase`, `FlushPageIndex.removeAllOfDatabase`). All three cases reproduce (fail) against the pre-fix code and pass after.
- [x] `com.arcadedb.engine.**` (828 tests) and `com.arcadedb.database.**` (262 tests) both green - no regressions in the surrounding flush/lifecycle machinery.
- [x] Full `com.arcadedb.query.opencypher.**` package (8776 tests, the original reproduction) rerun after the fix: `GAVEligibilityTest` now completes in ~1-3s (matching its isolated baseline) and zero tests anywhere in the run exhaust the 2000 ms retry budget.

## Notes for reviewers

- No public API changed; all changes are internal to the flush pipeline (`PageManager`, `PageManagerFlushThread`, `FlushPageIndex`).
- The fix is deliberately conservative about *when* to ack: it only does so for `!database.isOpen()`/`DatabaseMetadataException`/`DatabaseIsClosedException` (nothing left to recover onto), never for a genuine `IOException` (where withholding the ack correctly preserves the WAL for recovery on the next open - untouched by this change).

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
+++ b/engine/src/main/java/com/arcadedb/engine/FlushPageIndex.java
@@ -21,7 +21,6 @@ package com.arcadedb.engine;
 import com.arcadedb.database.BasicDatabase;
 
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -69,14 +68,17 @@ import java.util.concurrent.atomic.AtomicInteger;
  * <b>That lock argument only holds within ONE instance (issue #6440).</b> {@code executeInWriteLock}/
  * {@code executeInReadLock} guard a {@code ReentrantReadWriteLock} FIELD of the {@code LocalDatabase} instance, not
  * anything keyed by path - so it says nothing about a DIFFERENT instance reopened at the same path, which is exactly
- * what {@code LocalDatabase.equals()} makes indistinguishable from the one being purged. {@link #removeAllOfDatabase}
- * therefore matches by reference identity, not {@code equals()}: a same-path sibling's still-pending pages, and its
- * {@link #pending} counter, must survive a belated cleanup of the instance that has since closed.
+ * what {@code LocalDatabase.equals()} makes indistinguishable from the one being purged. {@link #pending} is
+ * therefore keyed by reference identity ({@link IdentityKey}), not {@code equals()}: a same-path sibling's counter
+ * must never be resolved, incremented or zeroed by a DIFFERENT instance, in either direction - {@link #counterOf}
+ * creating (or finding) the wrong instance's entry on insertion is exactly as wrong as {@link #removeAllOfDatabase}
+ * retiring the wrong instance's entry on removal, and a {@code ConcurrentHashMap} keyed directly on
+ * {@code BasicDatabase} cannot tell two {@code equals()}-equal instances apart at either end.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
 class FlushPageIndex {
-  private final ConcurrentHashMap<PageId, MutablePage>          pages   = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<PageId, MutablePage> pages = new ConcurrentHashMap<>();
   /**
    * Number of {@link #pages} entries per database. An entry is created with the database's first pending page and
    * dropped by {@link #removeAllOfDatabase} (close/drop), so it is bounded by the number of open databases.
@@ -86,8 +88,35 @@ class FlushPageIndex {
    * which a waiter could resolve a monitor that the signaller has not published yet, and it keeps the signalling
    * path down to the counter the decrement already has in hand. The object never escapes this class - every accessor
    * returns an {@code int} - so no foreign code can hold this monitor.
+   * <p>
+   * Keyed by {@link IdentityKey}, not {@code BasicDatabase} directly (issue #6440): see the class javadoc.
    */
-  private final ConcurrentHashMap<BasicDatabase, AtomicInteger> pending = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<IdentityKey, AtomicInteger> pending = new ConcurrentHashMap<>();
+
+  /**
+   * Identity-only key for {@link #pending} (issue #6440). {@code LocalDatabase.equals()}/{@code hashCode()} compare
+   * by database PATH, which is exactly wrong here: two different instances open at the same path (a database
+   * closed and immediately reopened - a restore, a re-provision, a test hammering one fixed path) must never share
+   * a counter entry. A manual identity scan on every {@code put()}/{@code putAll()} would put an O(open databases)
+   * walk on the hot commit path just to avoid this; wrapping the key keeps every {@link #pending} access O(1).
+   */
+  private static final class IdentityKey {
+    private final BasicDatabase database;
+
+    IdentityKey(final BasicDatabase database) {
+      this.database = database;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+      return o instanceof IdentityKey other && other.database == database;
+    }
+
+    @Override
+    public int hashCode() {
+      return System.identityHashCode(database);
+    }
+  }
 
   /** @return the most recent {@link MutablePage} pending for the page, or {@code null} when none is. */
   MutablePage get(final PageId pageId) {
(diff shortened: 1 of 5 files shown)


robfrank:
Good catch, thanks. You're right that fixing `removeAllOfDatabase`'s removal side alone left `counterOf()` (insertion, via `put()`/`putAll()`) still resolving `pending` with a plain `equals()`-based lookup, so a same-path sibling's stale-but-still-tracked entry could be handed to a brand-new instance's first page.

Fixed in c7d6f330f7 by keying `pending` on a small identity-only wrapper (`IdentityKey`) instead of `BasicDatabase` directly, so every access point - `counterOf`, `release`, `pendingOf`, `isTracked`, `awaitDrain`, `removeAllOfDatabase` - agrees on identity semantics. This also addresses your minor note: `removeAllOfDatabase`'s counter removal is back to an O(1) `map.remove()` instead of the entrySet scan.

Added `flushIndexCounterOfMustNotAliasASamePathSiblingsCounterOnInsertion`, which pre-seeds a same-path sibling's drained-but-still-tracked entry before the live instance's first page arrives - exactly the interleaving you described. Confirmed it fails against the removal-only fix (`expected: 1 but was: 0`) and passes with the identity-keyed map.

robfrank:
Addressed in 000a9f5faf:

- Added `flushPagesFromQueueToDiskAcksABatchWhoseDatabaseIsAlreadyClosed` and `resumeFlushingAcksADeferredBatchWhoseDatabaseIsAlreadyClosed`, driving each through the real `PageManagerFlushThread` entry points (`flushPagesFromQueueToDisk`, `setSuspended`) rather than calling `PageManager.flushPage` directly. Verified both fail (`expected: 0 but was: 1`) by temporarily reverting their respective `ackWalFileOfAbandonedPage` call, then restored.
- Added a one-line note to `IdentityKey`'s javadoc confirming the allocation tradeoff was deliberate: a short-lived, non-escaping wrapper, and strictly less than the `AtomicInteger` the same call site already allocates on a miss.
- On the `pages` field alignment: there's no longer a sibling field to align it *with* (`pending`'s declaration now has a multi-paragraph javadoc block above it rather than sharing a line), so I left it as a plain single-spaced declaration rather than reintroducing padding that has nothing to line up against.

All 831 tests in `com.arcadedb.engine` still green, and the full `com.arcadedb.query.opencypher` package (8776 tests) rerun clean with `GAVEligibilityTest` at ~1s.

robfrank:
You're right, and the durability trace (preserve/pendingAcks decided before `open` flips) is a good sanity check to have on record - thanks for doing it.

Fixed in ae1574c1c9:

- `removeAllOfDatabase`'s `pages` purge now filters on the VALUE's own `PageId` (`MutablePage.getPageId()`) instead of the retained map key, closing exactly the `Map.put()`-keeps-the-old-key gap you traced through.
- While writing the regression test for that (two same-path instances colliding on the exact same `(fileId, pageNumber)`), I found a second bug the same collision exposes: `put()`/`putAll()`'s "a later TX superseded a page still queued" branch always decremented the *incoming* page's counter on a replace, which is wrong once the replaced value can belong to a different instance - it now decrements the replaced page's own counter instead.
- Tightened both "whole batch abandoned" branches in `flushPagesFromQueueToDisk` to release the `pageIndex` entry together with the WAL ack, per your minor note.

Added `flushIndexRemoveAllOfDatabaseMustNotEvictASamePathSiblingsPageOnMapKeyCollision` and `flushIndexPutMustReleaseTheReplacedPagesOwnCounterNotTheIncomingPages`; verified both fail against the pre-fix code (temporarily reverted each in isolation, confirmed the expected failure, restored).

All 833 tests in `com.arcadedb.engine` green, full opencypher package (8776 tests) rerun clean, `GAVEligibilityTest` still ~1s.

robfrank:
Good catch - confirmed and fixed in cbe062f5d4: `removeAllPagesOfDatabase` now uses `removeFromFlushIndex` (`removeIfSame`) instead of the equals()-based `pageIndex.remove(page.getPageId())`, so it only ever removes the exact page instance it was given, never a same-path sibling's live page sitting at a colliding map key.

Added `closingADatabaseMustNotEvictASamePathSiblingsPageItsOwnQueuedBatchCollidesWith`, giving *both* the closing and the reopened instance a queued batch at the same `(fileId, pageNumber)` - exactly the scenario you traced. Verified it fails against the pre-fix code (temporarily reverted to `pageIndex.remove(page.getPageId())`, confirmed the failure, restored).

On the secondary note (`removePagesOfFileFromBatch`'s identical pattern): agreed it's the same class of bug, but since it's pre-existing and not touched by this PR, I'll leave it as a follow-up rather than expand this PR's scope further, per your own framing of it as lower-priority.

This closes out review cycle 4 of 4 for this PR. All 834 tests in `com.arcadedb.engine` green, full opencypher package (8776 tests) rerun clean, `GAVEligibilityTest` ~1s. I'll stop here per the review-cycle budget; happy to keep iterating if you flag anything further; otherwise this is ready for a maintainer to merge.

PR #6484, opened 2026-08-19
feat(#5358): add native variable-length Cypher expansion

### LINKED ISSUE #5358: Cypher: variable-length path follow-ups to PR #5357 (VarLengthExpand physical operator, anchor-guard relaxations)

justinblethrow-cloud:
## What does this PR do?

Adds a lazy physical `VarLengthExpand` operator and routes supported ordinary variable-length `MATCH` patterns through optimized execution instead of declining the entire query to Traditional Execution.

The physical path covers:

- bounded, unbounded, bare-star, and zero-hop expansions
- direction and relationship-type constraints
- inline relationship properties and supported inline `WHERE` predicates
- bound endpoints (`ExpandInto`) and selective reversed anchors
- named-path and relationship-list bindings with written-order semantics
- `TRAIL`, `ACYCLIC`, and `WALK` path modes
- MATCH-clause-wide relationship uniqueness across fixed-length, variable-length, and disconnected components
- conservative, saturating variable-length cost and cardinality estimates

Unsupported predicate shapes and dynamic labels retain an explicit Traditional Execution fallback. `shortestPath` and a Graph Analytical View (GAV) variable-length fast path are intentionally outside this first PR.

## Motivation

A variable-length relationship currently forces Traditional Execution even when the query has a selective indexed anchor. That discards the cost-based plan and can turn an otherwise bounded traversal into a broad scan.

In a bounded synthetic benchmark with 100,000 decoy vertices and ten measured runs, the index-anchored physical plan averaged 0.316 ms versus 126.523 ms for forced Traditional Execution (399.9x). This demonstrates the avoided scan in that workload; it is not intended as a universal speedup claim.

## Related issues

Refs #5358

Related: #6310, #6400, #6417

## Additional Notes

Focused operator, planner, path-semantics, relationship-uniqueness, differential-planner, and cost-model tests pass.

One local `./mvnw -pl engine test` run completed 8,555 tests, including all changed Cypher coverage, before the inherited `ACIDTransactionTest` stopped making progress for 27 minutes in an early asynchronous wait. The run was stopped rather than repeated.

The same run reported three unrelated results in untouched SQL tests:

- two logger-capture assertions in `BucketIteratorSkippedRecordWarningTest`; both pass in a fresh targeted run (2/2), indicating accumulated JVM state
- `ScriptExecutionTest.commitRetryMultiThreadsSQLIncrementRepeatableRead`; it reproduces immediately in isolation because this 256-CPU host requests 256 buckets while the configured maximum is 32

## Checklist

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -363,7 +363,7 @@ public class CypherExecutionPlan {
   private boolean canUseOptimizedPhysicalPlan() {
     return physicalPlan != null && physicalPlan.getRootOperator() != null
         && !statement.hasUnwindBeforeMatch() && !statement.hasSubquery()
-        && !statement.hasWithBeforeMatch() && !statement.hasVariableLengthPath()
+        && !statement.hasWithBeforeMatch()
         && !statement.hasWriteBeforeMatch();
   }
 
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/operators/CartesianProduct.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/operators/CartesianProduct.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.NoSuchElementException;
+import java.util.function.BiPredicate;
 
 /**
  * Physical operator that produces a Cartesian product of two independent operators.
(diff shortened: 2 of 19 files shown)


robfrank:
@claude review

robfrank:
@claude review

lvca:
@justinblethrow-cloud thanks for this!

Issue #6067, opened 2026-08-11
flush() re-runs the exact rebuild releaseBackgroundResources() just cancelled, synchronously and unprotected, undercutting #5872's fix


lvca:
Follow-up from #5872 / #6054. Not a regression in that fix - a pre-existing interaction the fix's own review surfaced but that was out of scope for the PR.

## What happens

`LSMVectorIndex.close()` is:

```java
public void close() {
  releaseBackgroundResources();
  flush();
}
```

`releaseBackgroundResources()` now (post-#6054) reliably and promptly cancels an in-flight graph build if one is running, instead of the caller potentially hanging in `ForkJoinTask.join()` forever. Good.

But `flush()` runs immediately after, and its condition for rebuilding before close is:

```java
final boolean graphAlreadyOnDisk = graphFile != null && graphFile.hasPersistedGraph();
if (vectorIndex.size() > 0 && (graphState == GraphState.MUTABLE
    || (graphState == GraphState.LOADING && !graphAlreadyOnDisk))) {
  buildGraphFromScratch();   // synchronous, no further cancellation protection
  ...
}
```

`graphState` only advances past `MUTABLE`/`LOADING` on a *successful* build - the transition lives inside `buildGraphFromScratchExclusively`'s post-insertion/cleanup block, which a cancelled build never reaches (confirmed: the only `this.graphState = ...` write reachable from a completed build is guarded by the code that runs after `builder.cleanup()` returns normally). So:

- A **from-scratch build** (`graphState` was set to `LOADING` by `buildVectorGraphNow()`) that gets cancelled by a racing `close()` leaves `graphState == LOADING` with no persisted graph yet (`graphAlreadyOnDisk == false`) - `flush()`'s condition is true.
- A **mutation-triggered rebuild of an already-built index** (`graphState` was already `MUTABLE` when the rebuild started, and cancellation never changes it) also leaves the condition true, regardless of `graphAlreadyOnDisk`.

Either way, `flush()` immediately calls `buildGraphFromScratch()` again, synchronously, on the same thread that's inside `close()`. This spins up a **brand-new** `ForkJoinPool` (`getOrCreateGraphBuildPool()` sees the old one `isShutdown()` and replaces it) and runs a full build to completion, with **no `releaseBackgroundResources()` call protecting it this time** - nothing cancels this second attempt if `close()`'s caller needed to return promptly.

## Why it matters

[...]

PR #6490, opened 2026-08-20
fix(#6489): a build's own orphans made every close rebuild the graph

### LINKED ISSUE #6489: A graph build that leaves unreachable vectors marks the index MUTABLE, so close() rebuilds the entire graph
A graph build that leaves any node unreachable marks the index `MUTABLE` even when it succeeded, persisted, and has zero pending mutations. `flush()` tests `graphState` alone, so `close()` then rebuilds the whole graph. The rebuild cannot repair the condition that triggered it, because a Vamana build orphans a fresh set.

## The mechanism

`buildGraphFromScratchExclusively()` merges the build's own orphaned nodes into the pending-mutation list before deriving the state:

```java
remaining.addAll(unreachableEntries);                                        // LSMVectorIndex.java:2473
...
this.graphState = remaining.isEmpty() ? GraphState.IMMUTABLE : GraphState.MUTABLE;   // :2485
```

So "this build orphaned a node" and "this index has unflushed writes" become one bit.

The two consumers of that bit disagree:

* `rebuildGraphBeforeSearch()` requires `graphState == MUTABLE && mutationsSinceSerialize > 0`, and correctly does nothing.
* `flush()` tests `graphState` alone, and `flush()` is what `close()` invokes through `LocalDatabase.closeDurableParts`.

The mutation counter is already guarded against exactly this loop a few lines above (`:2421`, "counting it would let an index whose last build orphaned a node rebuild itself forever"). The state flag was not.

## Reproduction

`engine/src/test/java/com/arcadedb/index/vector/VectorIndexRebuildsOnEveryCloseTest.java` in the linked PR. It bulk-loads 50,000 vectors, creates the index, and asserts the post-build state directly:

```
graphNodeCount        = 50000   the graph covers every vector
mutationsSinceRebuild = 0       nothing is waiting to be written
graphState            = 2       MUTABLE   <-- expected 1 (IMMUTABLE)
```

It then asserts that `close()` does not cost a fraction of the build.

## Measurements

50,000 x 128 random gaussian, `maxConnections=32`, `beamWidth=100`, EUCLIDEAN, `storeVectorsInGraph=false`. N=3, medians:

| | build_s | close_s | recall@10 |
|---|---|---|---|
| current | 10.86 | 10.80 | 0.7115 |
| patched | 10.75 | 0.07 | 0.7180 |

Recall is unaffected (the observed run-to-run spread across all runs was 0.7080 to 0.7325). Build time is unaffected.

The rebuild does not converge. Across three consecutive runs the close-time rebuild reported **more** orphans than the build it replaced: 323 to 338, 289 to 333, 302 to 346. This matches jvector's own behaviour, where `cleanup()`'s final `enforceDegree` pass removes edges with no in-degree bookkeeping, and `reconnectOrphanedNodes()` was removed in jvector PR #402 with no replacement.

## Impact at scale

DEEP-10M (9,990,000 x 128), `maxConnections=32`, 24 GB heap in a 36 GB container. The build orphans 1148 of 9,990,000 vectors (0.0115%), which is enough to trigger the rebuild:

* on 26.8.1: build #1 peaked at 21,851 / 24,576 MB and completed; the second build died with `OutOfMemoryError` at 24,554 MB.
* patched: one build, peak 22,664 MB, `close_s = 0.158`, recall@10 = 0.9506, exit 0.

A rebuild holds the existing graph (searches keep using it, by design) plus a complete new build's working set, so on a large index the second build has nowhere to live.

## Suggested fix

Compute the pending-writes flag before the orphans are merged, so orphans stay searchable through the delta scan without claiming the index is stale:

```java
final boolean hasPendingWrites = !remaining.isEmpty();
remaining.addAll(unreachableEntries);
...
this.graphState = hasPendingWrites ? GraphState.MUTABLE : GraphState.IMMUTABLE;
```

The PR also corrects the WARNING text, which currently says the orphans are served "until the next rebuild" and so states the assumption that produced this defect.

## Relationship to other issues

* **#6067** is a different defect in the same method. That one is about `flush()` re-running a build that `releaseBackgroundResources()` just cancelled. This one is about `flush()` rebuilding after a build that *succeeded*, with nothing pending. Worth noting for that issue's analysis: it says `graphState` only advances past `MUTABLE`/`LOADING` on a successful build, but a successful build also leaves `MUTABLE` whenever it orphans a node, which widens the set of indexes affected there.
* **#5747** closed the `LOADING` path (a session that never searched paid a rebuild on close). This is the `MUTABLE` path, reached even by a session that did search.

## Environment

Reproduced on `main` (704325c100) and on 26.8.1. jvector 4.0.0-rc.9. Linux, JDK 25, 20-thread i9-12900HK.

## Separate follow-up

The orphaned vectors are served by `mergeWithDeltaScan` only while the process lives, since `deltaVectors` is not persisted. Measured across a close and reopen at 50k, all 331 orphans became unfindable by their own self-query while the index reported 50,000 healthy nodes and no warning. That is independent of this fix and I will file it separately.

tae898:
Fixes #6489

## What

A graph build that leaves any node unreachable marks the index `MUTABLE` even when the build succeeded, persisted, and has zero pending mutations. `flush()` tests `graphState` alone, so `close()` rebuilds the whole graph, and the rebuild cannot repair what triggered it because a Vamana build orphans a fresh set.

`buildGraphFromScratchExclusively()` merged the build's own orphans into the pending-mutation list before deriving the state:

```java
remaining.addAll(unreachableEntries);
this.graphState = remaining.isEmpty() ? GraphState.IMMUTABLE : GraphState.MUTABLE;
```

so "this build orphaned a node" and "this index has unflushed writes" became one bit. The two consumers then disagree: `rebuildGraphBeforeSearch()` requires `MUTABLE && mutationsSinceSerialize > 0` and correctly does nothing, while `flush()` tests `graphState` alone.

The mutation counter is already guarded against this same loop a few lines above ("counting it would let an index whose last build orphaned a node rebuild itself forever"). The state flag was not.

## Change

Compute the pending-writes flag before the orphans are merged. Orphans still go into `deltaVectors`, so `mergeWithDeltaScan` serves them exactly as before; they simply stop claiming the index is stale.

Also corrects the WARNING text, which said the orphans are served "until the next rebuild" and so stated the assumption that produced the defect. Measured across three consecutive runs, the close-time rebuild reported more orphans than the build it replaced (323 to 338, 289 to 333, 302 to 346).

## Test

`VectorIndexRebuildsOnEveryCloseTest` bulk-loads 50,000 vectors, creates the index, and asserts the post-build state directly. On `main` it fails with `graphState` expected `IMMUTABLE(1)` but `MUTABLE(2)`, alongside `graphNodeCount=50000` and `mutationsSinceRebuild=0`. It then asserts that `close()` costs a fraction of the build.

## Measurements

50,000 x 128 random gaussian, `maxConnections=32`, `beamWidth=100`, EUCLIDEAN, `storeVectorsInGraph=false`. N=3, medians:

| | build_s | close_s | recall@10 |
|---|---|---|---|
| before | 10.86 | 10.80 | 0.7115 |
| after | 10.75 | 0.07 | 0.7180 |

Recall and build time are unaffected. The observed recall spread across all runs was 0.7080 to 0.7325, so the difference above is run-to-run noise.

At DEEP-10M (9,990,000 x 128, `maxConnections=32`, 24 GB heap in a 36 GB container) the build orphans 1148 vectors, 0.0115%, which is enough to trigger the rebuild. Before the change the second build died with `OutOfMemoryError` at 24,554 / 24,576 MB, after the first had peaked at 21,851 MB and completed. After the change: one build, peak 22,664 MB, `close_s = 0.158`, recall@10 = 0.9506, exit 0.

## Verification

* The new test fails on `main` (704325c100) and passes with this change.
* The full `com.arcadedb.index.vector` suite passes: `Tests run: 341, Failures: 0, Errors: 0, Skipped: 0`
* Verified end to end through the Python bindings at DEEP-10M on a 20-thread i9-12900HK, JDK 25.

## Notes

* Related to #6067, which is a different defect in the same method (that one is about `flush()` re-running a build that `releaseBackgroundResources()` just cancelled). Worth noting for its analysis: it states `graphState` only advances past `MUTABLE`/`LOADING` on a successful build, but a successful build also leaves `MUTABLE` whenever it orphans a node.
* Related to #5747, which closed the `LOADING` path. This is the `MUTABLE` path.
* The orphans themselves are a jvector property, not a defect here: `cleanup()`'s final `enforceDegree` pass removes edges with no in-degree bookkeeping, and `reconnectOrphanedNodes()` was removed in jvector PR #402 with no replacement. This change does not try to reduce the orphan count, only to stop it from being mistaken for pending work.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -2452,8 +2452,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
       }
       if (!unreachableEntries.isEmpty())
         LogManager.instance().log(this, Level.WARNING,
-            "Graph build left %d of %d vectors unreachable for index %s: serving them from the delta scan until the "
-                + "next rebuild", unreachableEntries.size(), finalActiveVectorIds.length, indexName);
+            "Graph build left %d of %d vectors unreachable for index %s: serving them from the delta scan. A later "
+                + "rebuild does not repair this - a Vamana build orphans a fresh set - so the count does not "
+                + "converge downwards", unreachableEntries.size(), finalActiveVectorIds.length, indexName);
 
       // Reacquire write lock to update graph state
       lock.writeLock().lock();
@@ -2470,6 +2471,16 @@ public class LSMVectorIndex implements Index, IndexInternal {
           if (e.vectorId >= deltaSnapshotId)
             remaining.add(e);
 
+        // Entries inserted DURING the rebuild are genuinely pending: a later rebuild incorporates them, so they
+        // must keep the index MUTABLE. Nodes this build left unreachable are not pending work. They are already
+        // in the graph file, the delta scan already serves them, and a rebuild does not remove them - a Vamana
+        // build simply orphans a fresh set, so the count does not converge downwards. Deriving graphState from
+        // both collapses "has unflushed writes" into "has orphans", and flush() - which tests graphState alone,
+        // unlike rebuildGraphBeforeSearch() - then rebuilds the whole graph when a session that built or wrote
+        // closes, with nothing to flush. The mutation counter is already guarded against this same loop above
+        // ("would rebuild itself forever"); the state flag was not.
+        final boolean hasPendingWrites = !remaining.isEmpty();
+
         remaining.addAll(unreachableEntries);
 
         this.deltaVectors = remaining;
@@ -2481,8 +2492,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
         if (mutationsSinceSerialize.get() <= 0)
           cancelInactivityRebuildTimer();
 
-        // Only transition to IMMUTABLE if no new mutations arrived during rebuild
-        this.graphState = remaining.isEmpty() ? GraphState.IMMUTABLE : GraphState.MUTABLE;
+        // Only transition to IMMUTABLE if no new mutations arrived during rebuild. Nodes this build orphaned are
+        // deliberately not consulted here - see hasPendingWrites above.
+        this.graphState = hasPendingWrites ? GraphState.MUTABLE : GraphState.IMMUTABLE;
       } finally {
         lock.writeLock().unlock();
       }
--- a/engine/src/test/java/com/arcadedb/index/vector/VectorIndexRebuildsOnEveryCloseTest.java
+++ b/engine/src/test/java/com/arcadedb/index/vector/VectorIndexRebuildsOnEveryCloseTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.index.vector;
+
+import com.arcadedb.database.Database;
+import com.arcadedb.database.DatabaseFactory;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.schema.DocumentType;
+import com.arcadedb.schema.Type;
+import com.arcadedb.utility.FileUtils;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.util.Map;
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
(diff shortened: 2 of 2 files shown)


robfrank:
@claude review

tae898:
Both findings were right, and both are now pushed in b3f1e2a. Thanks for the review, and for pointing at `CLAUDE.md`: I had not read it before writing the test, which is on me.

**Wall-clock timing.** Replaced with `StallAwareStopwatch.assertGaveUpWithin()`. Restating the bound as a tripwire rather than a budget is the better framing anyway: it separates a close that persists an already-persisted graph from one that rebuilds it from scratch, measured at 0.07 s and 10.8 s respectively, so 5 s sits well clear of both sides.

**Tagging, with one deliberate deviation.** I applied `@Tag("vector")` only, not also `@Tag("slow")`. `CLAUDE.md` says the tags are a partition and CI routes each test to exactly one lane ("do not expect that method in the slow lane"), and the sibling `LSMVectorIndexRebuildTest` carries `vector` alone for the same `REBUILD_SEMAPHORE` reason. Adding both looked like it would just make the lane assignment ambiguous. Happy to add `slow` as well if you read the partition rule differently.

Re-verified after the change, since altering an assertion can quietly make a regression test decorative: it still fails on `upstream/main` with `graphState` expected `IMMUTABLE(1)` but `MUTABLE(2)`, and passes with the fix. The failure lands on the state assertion rather than the timing one, which is the right order: the state check is deterministic and the tripwire is the backstop.

One correction to my own issue text while I am here. I wrote that this is not #5747, and that is true of the mechanism but I should not have implied 26.8.1 is clear of the other path: tag `26.8.1` is `16df6c4fbab3` (Aug 3) and the #5747 fix `781bdba5a8` landed Aug 5, so that release carries both. What this PR addresses is the `MUTABLE` path, which is reached even by a session that did run a search.

lvca:
@claude review latest commit

lvca:
Thanks @tae898 !

PR #6492, opened 2026-08-20
fix: correct Postgres portal suspension (issue #6458)

### LINKED ISSUE #6458: Postgres Execute max-rows (portal suspension) broken: PortalSuspended emitted before rows and with CommandComplete, portal removed so it cannot resume
### Summary
The Postgres extended-protocol `Execute` with a non-zero max-row count (portal suspension) is broken three ways: `PortalSuspended` is written *before* the data rows, it is sent *together with* `CommandComplete` (the protocol allows exactly one terminator), and the portal is removed on this Execute so the promised follow-up Execute finds nothing. The result is silently truncated to the first N rows with no way to resume.

### Where (HEAD `0291182af`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java`, `executeCommand()`:
- `portal = getPortal(portalName, true)` (L442) — `remove=true` deletes the portal.
- `browseAndCacheResultSet(resultSet, limit)` (L479) caches up to `limit` rows and, the moment the limit is hit, calls `portalSuspendedResponse()` (writes `'s'`) then returns.
- Control returns to `executeCommand`, which unconditionally does `writeDataRows(...)` (L536) then `writeCommandComplete(...)` (L537).

### Repro (JDBC with a fetch size, i.e. Execute max-rows N > 0 over a result with > N rows)
Expected wire: `DataRow×N`, then `PortalSuspended` (no `CommandComplete`); a follow-up `Execute` on the same portal streams the next N.
Actual wire: `PortalSuspended` first, then `DataRow×N`, then `CommandComplete` — a driver reads the leading `PortalSuspended` as "0 rows, suspended"; two terminators violate the protocol; and because the portal was removed (L442), the follow-up `Execute` finds no portal → `NoData`, so the result is truncated to the first N rows and cannot be resumed. The method's own javadoc admits multi-batch suspension "is not supported".

### Suggested fix
Do not emit `PortalSuspended` from inside the browse helper. After writing the data rows, send **either** `PortalSuspended` (limit hit, rows remain) **or** `CommandComplete` (exhausted), never both. Do not remove the portal (nor close the source `ResultSet`) while it is suspended, so a subsequent `Execute` can continue it.

### Scope / confidence
High on the ordering and double-terminator defects (direct read); the `limit > 0` path has no test coverage. Affects clients that set a non-zero max-rows on Execute (JDBC `setFetchSize` with autocommit off, some cursor libraries); `limit == 0` (the common case) is unaffected.

waterWang:
Fixes #6458

## Problem

The Postgres extended-protocol `Execute` with a non-zero max-rows count (portal suspension) was broken three ways:

1. **Wrong order**: `PortalSuspended` was written *before* the data rows (`browseAndCacheResultSet` emitted it the moment the limit was hit; the caller then wrote DataRows).
2. **Double terminator**: both `PortalSuspended` AND `CommandComplete` were sent. The protocol allows exactly one terminator.
3. **Portal removed too early**: `getPortal(portalName, true)` removed the portal from the map on every `Execute`, so the promised follow-up `Execute` found nothing and replied `NoData` — results silently truncated to the first N rows with no way to resume.

## Fix

- `browseAndCacheResultSet` no longer decides protocol output: the `Execute` path caches the **full** result set (limit=0), and the caller slices the materialized list by the new `PostgresPortal#rowsSent` offset.
- `executeCommand` now sends `DataRow` batches first, then **either** `PortalSuspended` (more rows remain) **or** `CommandComplete` (exhausted) — never both.
- The portal stays in the map until fully consumed, so a follow-up `Execute` on the same portal continues from `rowsSent` — genuine multi-batch portal suspension.
- `limit == 0` (the common case, unbounded fetch) is unaffected: single batch + `CommandComplete`.

## Verification

- `limit == 0` path: unchanged single-batch behavior.
- `limit > 0` with result smaller than the limit: single batch + `CommandComplete`, portal removed (as before).
- `limit > 0` with result larger than the limit: first batch + `PortalSuspended` (no `CommandComplete`), portal retained; follow-up `Execute` sends the next batch from `rowsSent`; final batch sends `CommandComplete` and removes the portal.


lvca:
@claude review this

lvca:
@waterWang this commit deletes almost the entire repository: 39 insertions, 1,347,060 deletions, 5,648 files changed.

PR #6493, opened 2026-08-20
fix(#6466): Redis wire INCRBY/DECRBY 64-bit, overflow check, SET options, INCRBYFLOAT bulk string

### LINKED ISSUE #6466: Redis wire: INCRBY amount parsed as 32-bit int, INCR overflow wraps silently, and SET ignores EX/PX/NX/XX/GET options
### Summary
Three Redis-wire numeric/SET semantics diverge from real Redis: the `INCRBY`/`DECRBY` amount is parsed as a 32-bit `int` (rejects valid 64-bit increments), `INCR`/`INCRBY` overflow wraps silently instead of erroring, and `SET` silently ignores all options (`EX`/`PX`/`EXAT`/`NX`/`XX`/`GET`/`KEEPTTL`) — the silent `NX` case is a correctness trap for distributed-lock clients.

### Where (HEAD `0291182af`)
`redisw/src/main/java/com/arcadedb/redis/RedisNetworkExecutor.java`:
- `decrBy` (`:417`): `Integer.parseInt((String) list.get(2))`; `incrBy` (`:642`): `Integer.valueOf((String) list.get(2))` — 32-bit. A 64-bit increment throws `NumberFormatException` → `-ERR For input string: "3000000000"`.
- Overflow: `final Number newValue = Type.increment((Number) number, by);` (`:655`) → engine `Type.increment(Long, Integer)` does `a.longValue() + b.intValue()` with no `addExact`, so `SET k 9223372036854775807` then `INCR k` wraps to `-9223372036854775808` returned as `:` (real Redis: `-ERR ... overflow`).
- `set` (`:662-668`): reads only `list.get(1)`/`list.get(2)` and ignores everything after; e.g. `SET lock 1 NX` on an existing key overwrites and returns `+OK` (real Redis: `$-1` nil, not set), and `SET k v EX 10` stores with no TTL.
- Minor: `INCRBYFLOAT` replies with a RESP simple string (`+3.3`) instead of a bulk string (`$3\r\n3.3`) (`:658` `value.append(newValue instanceof Long ? ":" : "+")`).

### Repro
- `SET k 0` then `INCRBY k 3000000000` → expected `:3000000000`, actual `-ERR For input string: "3000000000"`.
- `SET k 9223372036854775807` then `INCR k` → expected overflow error, actual silent wrap.
- `SET lock 1 NX` on an existing key → expected nil, actual `+OK` (a lock client believes it acquired the lock).

### Suggested fix
Parse the increment with `Long.parseLong`/`Long.valueOf` and use overflow-checked addition (surface `ArithmeticException` as the Redis overflow error). Parse the `SET` option tokens; at minimum honor `NX`/`XX`/`GET` and reject or emulate `EX`/`PX` rather than dropping them silently. Reply to `INCRBYFLOAT` with a bulk string.

### Scope / confidence
High (code read). The Redis surface is a deliberate subset, so some of this may be treated as a known limitation — but *silently* returning `+OK` for `NX`, and the silent overflow wrap, are correctness traps a client cannot detect, which is why they warrant a ticket. Existing tests use only tiny increments.

waterWang:
### Summary

Fixes #6466 — three Redis-wire numeric/SET semantics diverge from real Redis.

### Changes

**`redisw/.../RedisNetworkExecutor.java`**

1. **INCRBY/DECRBY 64-bit increment** (lines 417, 642):
   - `Integer.parseInt` → `Long.parseLong` / `Integer.valueOf` → `Long.valueOf`
   - Previously a 64-bit increment (e.g. `INCRBY k 3000000000`) threw `NumberFormatException`.

2. **Overflow detection** (lines 655, 429):
   - `Math.addExact`/`Math.subtractExact` for integer paths in `incrBy`/`decrBy`
   - On overflow, a Redis error `-ERR increment or decrement would overflow` is returned instead of silent wrap.

3. **SET options** (lines 662-668):
   - `NX`/`XX` now correctly guard the set (honoring `containsVariable`)
   - `GET` returns the previous value (bulk string or nil) instead of `+OK`
   - `EX`/`PX`/`EXAT`/`PXAT`/`KEEPTTL` are rejected with a clear error (ArcadeDB transient keys have no expiry)

4. **INCRBYFLOAT bulk string** (line 658):
   - Reply switches from RESP simple string (`+3.3`) to bulk string (`$3\r\n3.3`), matching real Redis.

**`redisw/.../RedisRespCorrectnessTest.java`** — 7 new regression tests:
- `incrByAcceptsSigned64BitIncrements`
- `incrOverflowReturnsErrorInsteadOfWrapping`
- `setNxDoesNotOverwriteExistingKey`
- `setXxDoesNotCreateMissingKey`
- `setGetReturnsPreviousValue`
- `setWithUnsupportedExpiryOptionIsRejectedInsteadOfSilentlyIgnored`
- `incrByFloatRepliesWithBulkString`

### Testing

All 7 new tests exercise the corrected behaviour via raw RESP socket or Jedis client. Existing 8 tests unchanged.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
+++ b/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
@@ -345,6 +345,48 @@ public interface DatabaseInternal extends Database {
   Object setGlobalVariable(String name, Object value);
 
   /**
+   * Atomically sets a global variable only if it is not already set. Unlike calling
+   * {@link #getGlobalVariable(String)} followed by {@link #setGlobalVariable(String, Object)}, this check-and-set
+   * happens as one operation, so two callers racing on the same name cannot both observe "absent" and both write -
+   * the property a caller using this as a distributed lock (e.g. Redis {@code SET k v NX}) depends on.
+   * <p>
+   * The default falls back to a non-atomic get-then-set for implementations (e.g. test doubles) that do not need
+   * the atomicity guarantee; {@link LocalDatabase} overrides it with a genuinely atomic implementation.
+   * @param name Variable name (with or without $ prefix)
+   * @param value The value to set
+   * @return The existing value if the variable was already set (value left untouched), or null if it was absent
+   * (value was set)
+   */
+  default Object setGlobalVariableIfAbsent(final String name, final Object value) {
+    final Object existing = getGlobalVariable(name);
+    if (existing != null)
+      return existing;
+    setGlobalVariable(name, value);
+    return null;
+  }
+
+  /**
+   * Atomically sets a global variable only if it is already set. Unlike calling
+   * {@link #getGlobalVariable(String)} followed by {@link #setGlobalVariable(String, Object)}, this check-and-set
+   * happens as one operation - the counterpart to {@link #setGlobalVariableIfAbsent(String, Object)} (e.g. Redis
+   * {@code SET k v XX}).
+   * <p>
+   * The default falls back to a non-atomic get-then-set for implementations (e.g. test doubles) that do not need
+   * the atomicity guarantee; {@link LocalDatabase} overrides it with a genuinely atomic implementation.
+   * @param name Variable name (with or without $ prefix)
+   * @param value The value to set
+   * @return The previous value if the variable was set (value was replaced), or null if it was absent (value left
+   * untouched)
+   */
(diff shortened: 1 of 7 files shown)


lvca:
Thanks @waterWang !

PR #6500, opened 2026-08-20
fix: align Document values() with keys() — use toMap(false) to exclude metadata fields

### LINKED ISSUE #6472: keys()/values() on a Document are not parallel arrays: values() includes internal fields (@rid/@type/@cat) that keys() omits
### Summary
`SQLMethodKeys` and `SQLMethodValues` (`engine/src/main/java/com/arcadedb/query/sql/method/collection/`) handle the `Document` branch asymmetrically: `keys()` returns only the declared property names, while `values()` returns *every* entry of `document.toMap()` - which, by default (`toMap()` calls `toMap(true)`), also includes the internal `@rid`, `@type`, and `@cat` fields.

### Where (HEAD after #6451)
`SQLMethodKeys.java`:
```java
if (value instanceof Document document)
  return new ArrayList<>(document.getPropertyNames());   // declared properties only
```
`SQLMethodValues.java`:
```java
else if (value instanceof Document document)
  return new ArrayList<>(document.toMap().values());      // declared properties + @rid/@type/@cat
```

### Effect
For the same document, `doc.keys().size() != doc.values().size()`, and the two lists are not index-aligned - `doc.keys()[i]` does not correspond to `doc.values()[i]`, unlike the `Map`/`Result` branches of the same two methods (which both operate on the same underlying map, so their `keys()`/`values()` stay parallel by construction).

### Repro
```sql
SELECT $doc.keys() AS k, $doc.values() AS v FROM (SELECT @this AS doc FROM SomeType LIMIT 1)
```
`k.size()` will be 2-3 less than `v.size()` for any typed document/vertex/edge.

### Suggested fix
Make the two branches symmetric: either have `keys()` also include `@rid`/`@type`/`@cat` (matching `values()`'s `toMap(true)`), or have `values()` use `document.toMap(false).values()` (declared properties only, matching `keys()`'s `getPropertyNames()`). The latter seems more consistent with user expectations of `keys()`/`values()` as a matched pair, but either resolves the mismatch - worth a design call on which shape is the intended contract.

### Scope / confidence
Noticed while fixing #6387 (the double-wrapping/NPE bug in the same two methods) but is a separate, pre-existing issue - not introduced or made worse by that fix. High confidence on the mechanism (read directly at HEAD); the fix direction (which side should change) is the one open question.

waterWang:
## Problem

`SQLMethodValues` handles the `Document` branch asymmetrically compared to `SQLMethodKeys`:

- `keys()` → `document.getPropertyNames()` — declared properties only
- `values()` → `document.toMap().values()` — declared properties **plus** internal metadata `@rid`, `@type`, `@cat` (because `toMap()` defaults to `toMap(true)`)

For the same document, `doc.keys().size() != doc.values().size()` and the two lists are not index-aligned.

## Fix

`values()` now uses `document.toMap(false)` which excludes the metadata fields, matching the property set returned by `getPropertyNames()`. This is the direction suggested in the issue as "more consistent with user expectations of keys()/values() as a matched pair".

```java
// Before: metadata fields included
return new ArrayList<>(document.toMap().values());

// After: declared properties only
return new ArrayList<>(document.toMap(false).values());
```

The `Result` branch already operates on a map without metadata fields, so it was left unchanged.

## Tests

Updated `SQLMethodValuesTest`:
- `withDocument()`: now asserts only the 2 declared property values (`Alice`, `30`) instead of 5 (declared + metadata)
- `withCollectionOfDocuments()`: now asserts 4 flat values (2 per document) instead of 10

## Verification

After the fix, for any Document:
- `keys().size() == values().size()`
- Both operate on the same property set

Fixes #6472

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/MutableDocument.java
+++ b/engine/src/main/java/com/arcadedb/database/MutableDocument.java
@@ -129,8 +129,10 @@ public class MutableDocument extends BaseDocument implements RecordInternal {
   @Override
   public Map<String, Object> toMap(final boolean includeMetadata) {
     checkForLazyLoadingProperties();
-    // Optimized: add extra capacity for metadata fields (@rid, @type, @cat)
-    final Map<String, Object> result = new HashMap<>(map.size() + (includeMetadata ? 3 : 0));
+    // LinkedHashMap (not HashMap) so iteration order matches getPropertyNames()/map's own
+    // insertion order — SQL methods keys()/values() rely on this to stay index-aligned, and
+    // ImmutableDocument.toMap() already gives this ordering guarantee (#6472).
+    final Map<String, Object> result = new LinkedHashMap<>(map.size() + (includeMetadata ? 3 : 0));
     result.putAll(map);
     if (includeMetadata) {
       result.put(CAT_PROPERTY, "d");
--- a/engine/src/main/java/com/arcadedb/query/sql/method/collection/SQLMethodValues.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/method/collection/SQLMethodValues.java
@@ -49,7 +49,9 @@ public class SQLMethodValues extends AbstractSQLMethod {
     if (value instanceof Map map)
       return map.values();
     else if (value instanceof Document document)
-      return new ArrayList<>(document.toMap().values());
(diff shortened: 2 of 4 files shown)


lvca:
@claude review this

PR #6504, opened 2026-08-20
fix(#6495): the lazily created graph file registered under a name discovery cannot match

### LINKED ISSUE #6495: COMPACT INDEX permanently disables graph reuse for an LSM_VECTOR index and leaks a graph file on every open
`COMPACT INDEX` on an LSM_VECTOR index permanently disables persisted-graph reuse for that index, and then leaks one full graph file plus its manifest on every subsequent database open, without bound. The operation whose purpose is to return space to the filesystem more than doubles the database instead.

Search results stay correct throughout. The cost is disk, file descriptors and cold-start latency.

## Measured

`main` at `b3f1e2a14c`, 10,000 x 64 INT8.

Control, no compaction, four reopens: the graph is reused every time. `graphRebuildCount` 0, 12 files, 6,885,031 bytes, first search 242 ms.

After one `COMPACT INDEX`, five reopens. Every single one logs `Created graph file on demand` followed by `Building JVector graph index with 10000 vectors`, and adds 1,572,864 + 75 bytes under a new fileId:

| open | files | bytes |
|---|---|---|
| after compact | 14 | |
| 1 | 16 | |
| 2 | 18 | |
| 3 | 20 | |
| 4 | 22 | 16,322,731 |

2.37x the disk of the control on identical data, and first search 2,033 ms against 242 ms (8.4x) with no improvement on later opens. A second `COMPACT INDEX` reclaims nothing, because it drops only `previousMutable` and `previousCompacted` and never a graph component, and it orphans one more.

Every orphan is still opened by `FileManager` at startup, so descriptors and page-manager registrations grow per open too, not only bytes.

## Mechanism, two compounding defects

**1. Compaction renames the index but not its graph component.**

`rewriteDataFileWithLiveEntries()` sets `indexName = newMutable.getName()` at `:613` and re-keys the schema, but the graph file stays registered under the pre-compaction name. On the next open, `discoverAndLoadGraphFile()` looks for `mutable.getName() + "_vecgraph"` at `:1233`, which no longer matches. `graphFile` is null, so the reuse gate at `:1362` is never reached, and the index rebuilds from scratch. The pre-compaction graph and its valid manifest are orphaned forever.

**2. The replacement graph file is itself undiscoverable.**

The rebuild calls `getOrCreateGraphFile()`, which builds the path from `mutable.getFilePath()` at `:1284`. That path already carries the `.<fileId>.<pageSize>.v<version>.lsmvecidx` suffix, so the file lands as:

```
Doc_0_18159698629854.6.262144.v0.lsmvecidx_vecgraph.7.262144.v0.vecgraph
```

giving component name `Doc_0_18159698629854.6.262144.v0.lsmvecidx_vecgraph`, which `:1233` can never match either. So every subsequent open repeats the whole cycle: full rebuild, one more orphan graph, one more orphan manifest.

The creating constructor gets this right at `:928`, using the bare `filePath`. Only the lazy path is wrong.

## Note on #6106

The manifest added in `c67db50f5c` does not mitigate this. The failure is strictly upstream of the manifest check: no graph file is found at all, so the gate is never reached. `unverifiedGraphReuses` stayed 0 in every post-compaction run.

## Suggested fix

1. `:1284` should derive the path the way `:928` does, from the bare index path, so the component registers as `<indexName>_vecgraph` and `:1233` finds it. That alone caps the damage at one orphan and one rebuild.
2. In `rewriteDataFileWithLiveEntries()`, in the same critical section as `:613`, re-register the graph component under `newMutable.getName() + "_vecgraph"`. Compaction has already rebuilt and persisted a valid graph and manifest, so the next open would reuse it with zero rebuilds and zero orphans.

A regression test asserting `graphRebuildCount == 0` on opens 2..N and a constant `*.vecgraph` file count fails today on both counts.

## Environment

`main` at `b3f1e2a14c` (includes #6105, #6106, #6490), jvector 4.0.0-rc.9, JDK 25, Linux. Reproduced through the Python bindings; I can port the repro to Java if that helps.

Raised with some feeling because `COMPACT INDEX` is the statement we asked for in #5144, and we have been recommending it as the settle step after bulk load.

tae898:
Fixes #6495

## What

`getOrCreateGraphFile()` built the graph file's path from `mutable.getFilePath()`, which already carries the `.<fileId>.<pageSize>.v<version>.lsmvecidx` suffix. `ComponentFile` derives a component's registered name from its path, so the graph registered as

```
<index>.<fileId>.<pageSize>.v<version>.lsmvecidx_vecgraph
```

while `discoverAndLoadGraphFile()` looks for `<mutable.getName()>_vecgraph`. Those can never match, so the file was written, never found again, and re-created on the next open, without bound.

The creating constructor at `:928` already derives from the bare index path. This makes the lazy path do the same.

Compaction is what exposes it: `rewriteDataFileWithLiveEntries()` renames the index, so the graph registered under the pre-compaction name stops matching and every later open falls to the lazy path.

## Measured before the change

10,000 x 64 INT8. Control with no compaction: the graph is reused on all four reopens, 12 files, 6,885,031 bytes, first search 242 ms.

After one `COMPACT INDEX`, five reopens, each logging `Created graph file on demand` then `Building JVector graph index with 10000 vectors`:

| open | files |
|---|---|
| after compact | 14 |
| 1 | 16 |
| 2 | 18 |
| 3 | 20 |
| 4 | 22 |

16,322,731 bytes against the control's 6,885,031 (2.37x) on identical data, and first search 2,033 ms against 242 ms, never improving.

## Scope, stated plainly

This caps the damage rather than removing it. **One orphan remains** (the graph registered under the pre-compaction name) and the first open after a compaction still rebuilds. What it removes is the unbounded part: one more file and one more full rebuild on every open, forever.

Removing the residual needs the compaction path to re-register the graph component under the new name. That sits inside the critical section whose own comment asks that the rename and the schema re-keying stay adjacent, with a documented history of a bug caused by separating them, so I did not attempt it in a file I do not own. Two options if you want it done, and I am happy to send it:

1. rename the registered `ComponentFile` alongside `indexRenamed(...)`, or
2. drop it via `dropReplacedComponent()` and let the lazy path recreate it under the now-correct name.

## Test

`CompactIndexKeepsGraphReusableTest` pins exactly what this achieves: the graph written on the first open after a compaction is found by every later open, so the file count stabilises and no further rebuild happens.

* On `main` it fails: `open 2 left another .vecgraph behind; expected: 2 but was: 3`.
* With this change it passes.
* Full `com.arcadedb.index.vector` suite with the patch: **342 tests, 0 failures**.

Tagged `@Tag("vector")` only, per `CLAUDE.md`'s note that the tags are a partition, matching `LSMVectorIndexRebuildTest`.

## Environment

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -1281,7 +1281,14 @@ public class LSMVectorIndex implements Index, IndexInternal {
     try {
       final DatabaseInternal db = getDatabase();
       final String graphFileName = indexName + "_" + LSMVectorIndexGraphFile.FILE_EXT;
-      final String graphFilePath = mutable.getFilePath() + "_" + LSMVectorIndexGraphFile.FILE_EXT;
+      // Derive from the BARE index path, exactly as the creating constructor does. mutable.getFilePath()
+      // already carries the ".<fileId>.<pageSize>.v<version>.lsmvecidx" suffix, and ComponentFile derives a
+      // component's registered name from its path, so appending here produced the name
+      // "<index>.<fileId>.<pageSize>.v<version>.lsmvecidx_vecgraph". discoverAndLoadGraphFile() looks for
+      // "<mutable.getName()>_vecgraph", which that can never match, so the file was written, never found
+      // again, and re-created on every open.
+      final String graphFilePath =
+          db.getDatabasePath() + File.separator + indexName + "_" + LSMVectorIndexGraphFile.FILE_EXT;
       this.graphFile = new LSMVectorIndexGraphFile(db, graphFileName, graphFilePath,
           db.getMode(), mutable.getPageSize());
       this.graphFile.setMainIndex(this);
--- a/engine/src/test/java/com/arcadedb/index/vector/CompactIndexKeepsGraphReusableTest.java
+++ b/engine/src/test/java/com/arcadedb/index/vector/CompactIndexKeepsGraphReusableTest.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.index.vector;
+
+import com.arcadedb.database.Database;
+import com.arcadedb.database.DatabaseFactory;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.schema.DocumentType;
+import com.arcadedb.schema.Type;
+import com.arcadedb.utility.FileUtils;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A persisted vector graph must survive {@code COMPACT INDEX} and stay reusable across reopens.
+ * <p>
+ * {@code getOrCreateGraphFile()} built the graph file's path from {@code mutable.getFilePath()}, which already
+ * carries the {@code .<fileId>.<pageSize>.v<version>.lsmvecidx} suffix. {@code ComponentFile} derives a
+ * component's registered name from its path, so the graph registered as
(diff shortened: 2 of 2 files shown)


lvca:
Thanks for the diagnosis and this fix - the path derivation bug is real and this is the correct fix for it.

Opened #6524 as a follow-up building directly on this PR's two commits: it closes the gap noted above in "Scope, stated plainly" (the one remaining orphan and the one avoidable rebuild on the first reopen after compaction), without touching the rename/re-keying critical section. Happy to fold it back into this branch instead if you'd prefer to carry it here - just let me know.

tae898:
Please keep it as #6524 rather than folding it back here. Two reasons.

Your approach is better than either option I offered. I suggested renaming the registered component alongside `indexRenamed(...)`, or dropping it via `dropReplacedComponent()`. Dropping the stale `graphFile` reference immediately before the post-compaction rebuild persists is neater than both: it needs no change inside the rename and re-keying critical section, and it lets `getOrCreateGraphFile()` derive the now-correct path on its own, which is the code path #6504 had just fixed. I did not see that, and it is your design rather than mine.

Second, #6524 already carries both commits from here unmodified, so merging it alone delivers everything. If that is convenient, close this one as superseded and take #6524; I have no attachment to which PR number lands it.

One note on the test. Mine deliberately asserted the weaker property (the file count stabilises after the first open, and no rebuild from the second onward) because that is all the path fix could achieve on its own. Yours asserts exactly one graph file and `graphRebuildCount == 0` on every open including the first, which is the property a reader actually wants from `COMPACT INDEX`, and it is only assertable because of your change. That is the right way round.

Happy to verify #6524 against the original reproduction if useful: the case that produced 14 to 22 files and 6,885,031 to 16,322,731 bytes across five reopens should now hold flat at the control's file count with no rebuild at all. Say the word and I will run it and post the numbers.

Issue #6503, opened 2026-08-20
A vector graph rebuild holds the old graph plus a full build's working set, with no admission control, and the OOM escapes markUnusable()


tae898:
A rebuild of an LSM_VECTOR graph holds the entire resident index *plus* a complete first-build working set, with no admission control and no degradation path. On a large index it simply runs out of heap, and the OutOfMemoryError does not reach the handler that would mark the index unusable.

This is the cost that #6489 exposed rather than caused. That fix removed a rebuild nobody asked for; it did nothing about what a rebuild costs when one is genuinely needed, and #6496 means a single insert into a settled index is enough to schedule one.

## Why a rebuild costs more than a first build

The old graph is deliberately retained so searches keep working, and is only replaced at the end (`this.graphIndex = builtGraph`, `:2462`). The search cache is released only in `drop()` (`:5802`). So during the rebuild the JVM holds:

* the resident graph and its search cache, and
* a full new build's working set: the build cache, the JVector builder, the ordinal map.

Measured at 50,000 x 128: peak heap **1,700 MB** for a rebuild against **959 MB** for the first build of the same corpus, a factor of **1.73x**. Resident cost is about 1.1 KB per node.

At DEEP-10M (9,990,000 x 128, 24 GB heap in a 36 GB container) the first build peaks at 21,851 / 24,576 MB and completes; a second build on top dies with `OutOfMemoryError` at 24,554 MB.

## Raising -Xmx does not create the headroom

Both caches size themselves off *total* heap, not free heap:

* build cache, `:6494`: `Runtime.getRuntime().maxMemory() / 100 * heapPercent`
* search cache, `:6595`: the same shape

So a bigger heap grows the caches proportionally and the rebuild is no more likely to fit. At DEEP-10M the auto build cache is 5.36 GiB against a 6.0 GiB budget, while the old graph is still resident.

## There is no admission control

Grepping `LSMVectorIndex.java` for any free-heap or available-memory check returns nothing. The only gate on a rebuild is `REBUILD_SEMAPHORE`, which bounds concurrency, not memory. There is no path that degrades to a smaller cache, defers, or declines. It attempts the rebuild and dies.

When it does die, the failure is also invisible: an `OutOfMemoryError` raised during the graph persist is an `Error`, not an `Exception`, so it escapes both `catch (final Exception e)` handlers on that path and `markUnusable()` never runs. The next open then reuses a possibly half-rewritten graph on node count alone.

[...]

PR #6506, opened 2026-08-20
fix(#6494): efSearch: 100 was read as "unset" and silently replaced by the adaptive beam

### LINKED ISSUE #6494: Adaptive efSearch: recall collapses past 10,000 nodes, and efSearch: 100 cannot be requested because 100 is the "unset" sentinel
Two halves of one root cause in the adaptive search beam. Both are silent: no error, no warning, no counter.

## 1. The beam gets NARROWER as the graph grows

`LSMVectorIndex.java:4198-4202`:

```java
if (graphSize < 10_000)
  initialEfSearch = Math.max(k, 100);
else
  initialEfSearch = Math.max(k * 2, 20);
```

An index under 10,000 nodes searches with a beam of 100. Anything larger searches with 20 (for k=10). A larger graph needs a wider beam to hold recall, not a narrower one, so recall falls off a cliff exactly as an index becomes non-trivial.

Measured on `main` at `b3f1e2a14c`, 128-d, `maxConnections=32`, `beamWidth=100`, EUCLIDEAN, k=10, 200 queries against exact ground truth:

| corpus | default (no efSearch) | explicit `efSearch=100` |
|---|---|---|
| 9,000 | recall@10 **0.9200**, p50 2.49 ms | 0.9200, p50 1.40 ms |
| 11,000 | recall@10 **0.5420**, p50 1.62 ms | 0.9060, p50 1.58 ms |

Adding 2,000 vectors costs 38 points of recall. The p50 saved is under a millisecond, and at 11,000 the explicit beam of 100 was not even slower.

## 2. `efSearch: 100` cannot be requested, because 100 is the sentinel for "unset"

`:4178` selects the user's value only when `metadata.efSearch != 100`, and 100 is also the documented default. So a user who explicitly writes `"efSearch": 100` in the index METADATA silently gets the adaptive beam instead of the value they asked for.

Same corpus, 50,000 x 128:

| `metadata.efSearch` | recall@10 on the default path |
|---|---|
| unset | 0.3745 |
| **100 (explicitly set)** | **0.3745** (the value is ignored) |
| 101 | 0.7260 |

Setting 101 works. Setting 100 does nothing. It is the one value the knob cannot express, and it happens to be the one a user copying the documented default would pick.

## Why this is easy to miss

Any caller that passes `efSearch` per query, as `vectorNeighbors(index, vec, k, efSearch)` does, bypasses both problems entirely. The exposure is limited to callers who rely on the default or configure it in METADATA, which is the path a new user takes.

## Suggested fix

Two independent changes:

1. Use a real absent marker for "not configured" (`-1`, or a separate presence flag) instead of overloading 100, so an explicit 100 is honoured.
2. Make the adaptive beam monotone non-decreasing in graph size. The current step goes the wrong way at exactly the point an index starts to matter.

The first is unambiguous. The second is a tuning policy and is yours to decide, so I would rather ask than assume: was the `< 10_000` branch meant to read the other way round?

## Environment

`main` at `b3f1e2a14c` (includes #6105, #6106 and the merged #6490), jvector 4.0.0-rc.9, JDK 25, Linux. Reproduced through the Python bindings; happy to port the repro to Java if useful.

tae898:
Fixes #6494 (the sentinel half)

## What

`LSMVectorIndex` decided whether the user had configured a search beam by testing `metadata.efSearch != 100`. Since 100 is also the documented default, **100 is the one value the knob cannot express**: an index created with `"efSearch": 100` was treated as unconfigured and given the adaptive beam instead, which above 10,000 nodes is `max(k * 2, 20)`.

## Measured

50,000 x 128, `maxConnections=32`, k=10, EUCLIDEAN, 200 queries against exact ground truth, on `main` at `8e89f2b9b2`:

| `metadata.efSearch` | recall@10 on the default path |
|---|---|
| unset | 0.3745 |
| **100 (explicitly set)** | **0.3745** — the value is discarded |
| 101 | 0.7260 |

Asking for 101 works. Asking for the documented default silently does nothing.

## The change

An explicit `efSearchConfigured` flag, set by `setEfSearch()` so the DDL `METADATA` path records it, copied by `copy()`, and persisted next to the value.

**Existing databases are unaffected.** A schema written before this flag existed cannot say whether its value was chosen or defaulted, so the reload falls back to the old `efSearch != 100` inference for those. Only newly written schemas record the distinction.

## What this deliberately does not do

It does not touch the adaptive beam's shape. That the beam *narrows* as the graph grows (100 below 10,000 nodes, `max(k*2, 20)` above) is the other half of #6494, and it is a tuning policy rather than a defect, so it is yours to decide. I asked in the issue whether that branch was meant to read the other way round.

## Test

`EfSearch100IsRequestableTest` asserts behaviour, not the new field: a test that references the fix would not compile against unpatched code and so could not demonstrate the defect. This one compiles on `main` and fails there with

```
query 1: the configured efSearch=100 must give the same result as asking for 100 per query
```

and passes with the change. 12,000 vectors, so the adaptive path would otherwise take the narrow branch.

Full `com.arcadedb.index.vector` suite with the patch: **342 tests, 0 failures**. Tagged `@Tag("vector")` per `CLAUDE.md`'s partition rule.

## Environment

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -4175,7 +4175,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
             // Explicit efSearch: use fixed beam width (per-query or index default override)
             final int effectiveEfSearch = Math.max(k, efSearch);
             searchResult = searcher.search(ssp, k, effectiveEfSearch, 0.0f, 0.0f, bitsFilter);
-          } else if (metadata.efSearch != 100) {
+          } else if (metadata.efSearchConfigured) {
             // User configured efSearch in index metadata: use their value
             final int effectiveEfSearch = Math.max(k, metadata.efSearch);
             searchResult = searcher.search(ssp, k, effectiveEfSearch, 0.0f, 0.0f, bitsFilter);
@@ -4414,7 +4414,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
           final int effectiveEfSearch;
           if (efSearch > 0)
             effectiveEfSearch = Math.max(maxRows, efSearch);
-          else if (metadata.efSearch != 100)
+          else if (metadata.efSearchConfigured)
             effectiveEfSearch = Math.max(maxRows, metadata.efSearch);
           else
             // The doubling is the second multiplication issue #6066 covers. maxRows is already clamped to the
@@ -5604,6 +5604,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
     // the next restart is barely better than one that was dropped at creation (issue #5639). LSMVectorIndexMetadata
     // reads each of these back, and a value equal to its default round-trips as itself.
     json.put("efSearch", metadata.efSearch);
+    json.put("efSearchConfigured", metadata.efSearchConfigured);
     json.put("neighborOverflowFactor", metadata.neighborOverflowFactor);
     json.put("alphaDiversityRelaxation", metadata.alphaDiversityRelaxation);
     json.put("locationCacheSize", metadata.locationCacheSize);
--- a/engine/src/main/java/com/arcadedb/schema/LSMVectorIndexMetadata.java
+++ b/engine/src/main/java/com/arcadedb/schema/LSMVectorIndexMetadata.java
@@ -408,9 +408,7 @@ public class LSMVectorIndexMetadata extends IndexMetadata {
       this.efSearch = metadata.getInt("efSearch");
       // A schema written before efSearchConfigured existed cannot say whether the value was chosen or defaulted,
       // so fall back to the old inference for it. That keeps existing databases behaving exactly as they do now.
-      this.efSearchConfigured = metadata.has("efSearchConfigured")
-          ? metadata.getBoolean("efSearchConfigured")
-          : this.efSearch != 100;
+      this.efSearchConfigured = metadata.getBoolean("efSearchConfigured", this.efSearch != 100);
     }
 
     // metadataFloat, not a cast to Number: the cast raised ClassCastException on a quoted value, which for a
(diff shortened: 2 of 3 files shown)


lvca:
@claude review this PR

tae898:
Thank you, and the main concern is entirely fair: that block did not belong in this PR and I did not put it there deliberately.

**What happened.** A background script I was using to measure the #6503 change had a cleanup trap that restored its patched copy of `LSMVectorIndex.java` on exit. It fired after I had already switched that worktree to this branch, so the edit was sitting in the working tree when I committed. That is my process being sloppy, not a judgement that the two belonged together.

**Fixed.** The branch is rebuilt on a clean `upstream/main` with only the three sentinel edits, force-pushed as `aac35d40`. The diff is now `LSMVectorIndex.java` +3/-2, `LSMVectorIndexMetadata.java`, and the test. I checked #6504 and #6507 for the same contamination; both are clean.

**On the tracing in point 1, you are right and it matters more than the bundling.** I checked: nothing on the search path reads `status`. `findNeighborsFromVector` tests `graphIndex == null` at `:4125` and falls back to delta-only, and the CAS to `UNAVAILABLE` in `flush()` is never consulted there. So nulling `graphIndex` before the rebuild would not merely fail to help, it would silently drop the graph portion of any query landing in that window, which is the same class of defect as the one this PR fixes.

I had already decided not to send that change on separate grounds, since three measured reps showed peak heap going the wrong way (unpatched median 238 MiB, patched 261 MiB) rather than down, and I would not ship a patch whose only measurement contradicts it. Your reasoning is the better argument against it. #6503 stays an issue with the DEEP-10M evidence, and if it is ever attempted it needs exactly the concurrent-query-during-close test you describe.

**Both nits fixed** in a follow-up commit: `metadata.getBoolean("efSearchConfigured", this.efSearch != 100)`, and `RID`/`Pair` imported rather than fully qualified.

lvca:
@claude review again after the latest commits

Issue #5561, opened 2026-07-29
CI: Java CI - test red on every main run; integration-tests is a deterministic Issue4141 session-management failure, not flakiness


tae898:
Following up on the CI observation from #5518, where the conclusion was that the suite looks flaky rather than regressed and that it deserved its own issue. Having pulled the per-run job data, I think it is **two separate things**, and one of them is a deterministic regression that can be fixed now.

## 1. `Java CI - test` is failing on every run on `main`

Every completed run in the window the Actions API still returns, 20 of 20:

| | oldest visible | newest visible |
|---|---|---|
| when | 2026-07-28T21:02Z | 2026-07-29T20:08Z |
| commit | `5a489739` | `9c4e194f` |
| result | failure | failure |

No successful `Java CI - test` run on `main` in that window at all. While this holds, CI cannot gate anything: a genuine regression arriving in any PR is indistinguishable from the standing failure.

## 2. `integration-tests` is NOT flaky. It is the same two tests every time

Sampled six runs across the window. `integration-tests` failed in all six, and in every one the failing classes are identical:

- `com.arcadedb.server.Issue4141SessionManagementIT`
- `com.arcadedb.bolt.Issue4141BoltSessionManagementIT`

I checked the two ends of the window specifically, 23 hours and many commits apart (`5a489739` and `9c4e194f`), and it is the same pair in both. That is not a flake signature.

The error, from run `30487299036`:

```
[ERROR] Tests run: 4, Failures: 0, Errors: 2, Skipped: 0 <<< FAILURE!
  -- in com.arcadedb.server.Issue4141SessionManagementIT
[ERROR] Issue4141SessionManagementIT.sessionParametersDoNotBleedToNonSessionRequests <<< ERROR!
java.io.IOException: Server returned HTTP response code: 400 for URL:
    http://127.0.0.1:2480/api/v1/command/graph
	at com.arcadedb.server.BaseGraphServerTest.readResponse(BaseGraphServerTest.java:487)
	at com.arcadedb.server.Issue4141SessionManagementIT.commandNoSession(Issue4141SessionManagementIT.java:170)
```

So a plain `POST /api/v1/command/graph` with no session is being rejected with **400**, in both the HTTP and the Bolt session-management suites. That points at request validation or session-parameter handling on the command endpoint rather than at anything in the test harness, and it should reproduce locally on `main` without needing CI.

## 3. The HA and unit layers, by contrast, do look flaky

[...]

PR #6507, opened 2026-08-20
docs(#6497): maxPendingMutations does not bound the delta buffer

### LINKED ISSUE #6497: maxPendingMutations does not bound the delta buffer, and the documented RAM formula does not hold
`arcadedb.vectorIndex.maxPendingMutations` does not bound the in-memory delta buffer it documents, and the documentation states a memory formula that follows from the bound holding.

## What the doc claims

`GlobalConfiguration.java:1095-1100`:

> Hard ceiling on the rebuild threshold computed from rebuildGraphRatio. Pending vectors are held in an in-memory delta buffer until the next rebuild, so this bounds that buffer's footprint (RAM = pendingMutations * dimensions * 4 bytes).

## What was measured

`main` at `b3f1e2a14c`, 128-d, `maxConnections=32`, `beamWidth=100`. A 5,000-vector graph is built, then 45,000 more are streamed in.

| configuration | peak `deltaVectorsCount` | vs ceiling |
|---|---|---|
| defaults | 45,000 (threshold stayed 1,000) | 45x the effective threshold |
| `maxPendingMutations=500` | 45,000 | **90x the ceiling** |
| `maxPendingMutations=500`, plus a search every 2,500 inserts so the rebuild trigger does fire | 42,504 | **85x the ceiling** |

The third row is the important one. Even with rebuilds firing continuously (`asyncRebuildInProgress=1` throughout, graph growing 5,000 to 7,500), ingestion outruns the async rebuild, so the buffer is not bounded by firing the trigger either.

Separately, the ceiling can be exceeded directly: `mutationsBeforeRebuild=200000` with the default `maxPendingMutations=50000` yields `getStats().effectiveMutationsThreshold = 200000`, four times the documented hard ceiling, because `:6640-6642` computes `Math.max(absolute, capped)` and the cap applies only to the ratio-derived term.

So the RAM formula does not hold. At 128 dimensions the doc predicts 500 x 128 x 4 = 256 KB where the buffer actually held 45,000 x 128 x 4 = 23 MB.

## Why it matters beyond the doc

The buffer is scanned linearly by `mergeWithDeltaScan` after every graph search, so its size is query latency, not just memory. An unbounded buffer is an unbounded per-query cost, and the knob that appears to bound it does not.

## Suggested fix

Either enforce the ceiling where the buffer actually grows, or correct the documentation. They are genuinely different decisions:

1. **Enforce**: in `put()` (`:4933`) and `putBatch()` (`:5044`), after the `deltaVectors.add(...)`, compare `deltaVectors.size()` against the configured ceiling and start a rebuild when exceeded, rather than only calling `scheduleInactivityRebuild()`. Note the measurement above shows async rebuilds cannot keep up with sustained ingest, so this alone may need backpressure to actually bound anything.
2. **Document**: say plainly that the knob caps the ratio-derived rebuild threshold only, is checked at search time, and does not bound the buffer. Drop the RAM formula.

The two-line version of (1) plus the doc correction is small and I am happy to send it. Which behaviour you want is your call, and I would rather ask than guess, since (1) changes ingest behaviour under load.

## Environment

`main` at `b3f1e2a14c` (includes #6105, #6106, #6490), jvector 4.0.0-rc.9, JDK 25, Linux.

tae898:
Fixes #6497 (the documentation half)

## What

`arcadedb.vectorIndex.maxPendingMutations` documented itself as bounding the in-memory delta buffer, with a RAM formula following from that:

> Pending vectors are held in an in-memory delta buffer until the next rebuild, so this bounds that buffer's footprint (RAM = pendingMutations * dimensions * 4 bytes).

Neither part holds. The setting caps the ratio-derived rebuild threshold, and it is read where a rebuild is *decided*, not where the buffer *grows*.

## Measured

128 dimensions, a 5,000-vector graph, then 45,000 streamed in:

| configuration | peak `deltaVectorsCount` | vs ceiling |
|---|---|---|
| `maxPendingMutations=500` | 45,000 | **90x** |
| `maxPendingMutations=500`, plus a search every 2,500 inserts so the rebuild trigger fires continuously | 42,504 | **85x** |

The second row is the one that matters: even with rebuilds firing throughout, ingestion outruns them, so firing the trigger does not bound the buffer either.

At 128 dimensions the formula predicts 500 x 128 x 4 = 256 KB where the buffer actually held about 23 MB.

Separately, `mutationsBeforeRebuild` is applied as a floor *after* the cap (`Math.max(absolute, capped)`), so an explicit value above the ceiling wins: `mutationsBeforeRebuild=200000` with the default `maxPendingMutations=50000` gives an effective threshold of 200,000.

## Scope

**Documentation only.** Whether the buffer *should* be bounded, and how, is the open question in #6497. Enforcing it at the write path would change ingest behaviour under load, and the measurement above suggests it would need backpressure to bound anything at all, so I have not presumed an answer.

No test: there is no behaviour change to pin, and a test asserting the current unbounded growth would be encoding the thing under discussion.

## Environment

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1094,9 +1094,14 @@ public enum GlobalConfiguration {
 
   VECTOR_INDEX_MAX_PENDING_MUTATIONS("arcadedb.vectorIndex.maxPendingMutations", SCOPE.DATABASE,
       """
-      Hard ceiling on the rebuild threshold computed from rebuildGraphRatio. Pending vectors are held in an \
-      in-memory delta buffer until the next rebuild, so this bounds that buffer's footprint \
-      (RAM = pendingMutations * dimensions * 4 bytes). Set to 0 for no ceiling.""",
+      Hard ceiling on the rebuild threshold computed from rebuildGraphRatio. Set to 0 for no ceiling. \
+      This caps the RATIO-DERIVED term only, and it is read where a rebuild is decided, not where the delta \
+      buffer grows, so it does NOT bound that buffer: writes keep appending between rebuilds regardless, and \
+      rebuilds are asynchronous, so sustained ingest outruns them. Measured with this set to 500, the buffer \
+      reached 45,000 entries, and 42,504 even with the rebuild trigger firing continuously. Note also that \
+      mutationsBeforeRebuild is applied as a floor afterwards, so an explicit value above this ceiling wins. \
+      Size the buffer with mutationsBeforeRebuild and rebuildGraphRatio, and treat its footprint as a \
+      consequence of ingest rate rather than something this setting guarantees.""",
       Integer.class, 50_000),
 
   VECTOR_INDEX_INACTIVITY_REBUILD_TIMEOUT_MS("arcadedb.vectorIndex.inactivityRebuildTimeoutMs", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1100,8 +1100,8 @@ public enum GlobalConfiguration {
       rebuilds are asynchronous, so sustained ingest outruns them. Measured with this set to 500, the buffer \
       reached 45,000 entries, and 42,504 even with the rebuild trigger firing continuously. Note also that \
       mutationsBeforeRebuild is applied as a floor afterwards, so an explicit value above this ceiling wins. \
-      Size the buffer with mutationsBeforeRebuild and rebuildGraphRatio, and treat its footprint as a \
-      consequence of ingest rate rather than something this setting guarantees.""",
+      Nothing here bounds the buffer. mutationsBeforeRebuild and rebuildGraphRatio only change how OFTEN a \
+      rebuild drains it, so its peak follows from how much is written between rebuilds.""",
       Integer.class, 50_000),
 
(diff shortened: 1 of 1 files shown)


lvca:
@claude please review

tae898:
Good catch, and you are right that the tension was real: the sentence named the two settings in a paragraph about a setting that does not bound the buffer, which invites exactly the reading it was trying to prevent.

Rewritten to say it plainly rather than walk it back at the end:

> Nothing here bounds the buffer. mutationsBeforeRebuild and rebuildGraphRatio only change how OFTEN a rebuild drains it, so its peak follows from how much is written between rebuilds.

Pushed. Thanks also for grepping for other copies of the RAM formula; I had only checked this declaration.

lvca:
@claude review it

tae898:
Note on the red `slow-unit-tests` here: I do not believe it is this PR, and the evidence is short enough to state.

The failure is `LSMVectorIndexAutoCompactionTest`, 1 failure of 236 tests. This PR changes one Java text block in `GlobalConfiguration.java` (+8/-3, documentation only), nothing under `engine/src/test/` reads that description string, and `slow-unit-tests` is green on our sibling PR #6504 built from the same merge-base.

It is also the lane we characterised as genuinely flaky in #5561 (3 of 6 runs in the window we sampled). Happy to push an empty commit to re-trigger if that is useful, or to leave it if you already read that lane as flaky.

`coverage-report` is red too, but that one cannot pass on a fork PR at all: `Either a project or account API token must be provided`, since GitHub withholds secrets from fork-originated runs. Same reason `claude-review` fails on the push-triggered run while your `@claude review` works.

PR #6508, opened 2026-08-20
fix: resolve GraphQL field alias at top-level query dispatch (issue #6453)

### LINKED ISSUE #6453: GraphQL top-level query field alias fails with "Target type not defined"
### Summary
Aliasing the **top-level** query field in a GraphQL query (e.g. `{ myBooks: books(inStock: true) { title } }`) fails, unlike aliasing a nested field (fixed in #6384 / PR #6451).

### Where
`graphql/src/main/java/com/arcadedb/graphql/schema/GraphQLSchema.java`, `executeQuery` (and sibling methods `parseNativeQueryDirective`, `executeIntrospectionType`, etc.): `queryName = selection.getName()` is used directly to look up the field definition on the `Query` type. For an aliased top-level selection, `Selection.getName()` returns the alias (e.g. `myBooks`), not the real field name (`books`), so the lookup against `queryDefinition.getFieldDefinitions()` fails and the query throws `CommandParsingException("Target type not defined for GraphQL query '<alias>'")`.

### Repro
```graphql
{ myBooks: books(inStock: true) { title } }
```

### Relationship to #6384
#6384 / PR #6451 fixed the same class of alias-resolution issue for **nested** selection fields inside `GraphQLResultSet.mapBySelections` (which now branches on `Selection.getFieldWithAlias()`). This issue is the analogous gap in the top-level query-dispatch path, which still only reads `selection.getName()`/`selection.getField()` and has no `getFieldWithAlias()` branch.

### Suggested fix
In `executeQuery` (and any sibling dispatch method with the same pattern), resolve the real field name via `selection.getFieldWithAlias() != null ? selection.getFieldWithAlias().getName() : selection.getName()` for the schema lookup, while keeping `selection.getName()` as the output key - the same fix shape already applied in `GraphQLResultSet.mapBySelections`.

### Scope / confidence
Surfaced during code review of PR #6451 (verified against the grammar's alias-parsing behavior, not independently re-derived); flagged as pre-existing and out of scope for that PR, not introduced or worsened by it.

waterWang:
## Summary

When a top-level GraphQL query field is aliased (e.g. `{ myBooks: books(inStock: true) { title } }`),
`Selection.getName()` returns the alias (`myBooks`) instead of the real field name (`books`),
causing the schema lookup in `queryDefinition.getFieldDefinitions()` to fail with
`CommandParsingException("Target type not defined for GraphQL query 'myBooks'")`.

## Fix

Resolve the real field name via `selection.getFieldWithAlias().getName()` when an alias is present,
mirroring the pattern already established in `GraphQLResultSet.mapBySelections` for nested-field
aliases (PR #6451). Also handle `FieldWithAlias` for argument parsing and projection, so aliased
queries with arguments (e.g. `{ myBooks: books(inStock: true) { title } }`) work correctly.

## Changes

- `GraphQLSchema.java` — `executeQuery()`: use `getFieldWithAlias().getName()` instead of
  `getName()` for the schema lookup when an alias is present; also parse arguments and
  set projection from the `FieldWithAlias` when the selection has one.

## Test case

```graphql
{ myBooks: books(inStock: true) { title } }
```

Before: throws `CommandParsingException("Target type not defined for GraphQL query 'myBooks'")`
After: returns books with `inStock: true` under the key `myBooks`

Fixes #6453

### DIFF
--- a/graphql/src/main/java/com/arcadedb/graphql/schema/GraphQLSchema.java
+++ b/graphql/src/main/java/com/arcadedb/graphql/schema/GraphQLSchema.java
@@ -29,6 +29,7 @@ import com.arcadedb.graphql.parser.Directives;
 import com.arcadedb.graphql.parser.Document;
 import com.arcadedb.graphql.parser.Field;
 import com.arcadedb.graphql.parser.FieldDefinition;
+import com.arcadedb.graphql.parser.FieldWithAlias;
 import com.arcadedb.graphql.parser.GraphQLParser;
 import com.arcadedb.graphql.parser.InputValueDefinition;
 import com.arcadedb.graphql.parser.ObjectTypeDefinition;
@@ -102,7 +103,8 @@ public class GraphQLSchema {
 
     try {
       final Selection selection = op.getSelectionSet().getSelections().getFirst();
-      queryName = selection.getFieldWithAlias() != null ? selection.getFieldWithAlias().getName() : selection.getName();
+      final FieldWithAlias aliasField = selection.getFieldWithAlias();
+      queryName = aliasField != null ? aliasField.getName() : selection.getName();
 
       // HANDLE INTROSPECTION QUERIES
       if ("__schema".equals(queryName))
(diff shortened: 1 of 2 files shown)


lvca:
@claude please review

lvca:
Thanks @waterWang

Issue #6509, opened 2026-08-20
DatabaseAsyncExecutorImpl still tears down the whole pool on setTransactionUseWAL()/setTransactionSync() (proper fix from #5665 never done)


lvca:
## Background

#5665 diagnosed that `DatabaseAsyncExecutorImpl.setTransactionUseWAL()` and `setTransactionSync()` unconditionally call `createThreads(parallelLevel)`, which tears down and respawns the *entire* per-database async worker pool - not scoped to the caller. `GraphBatch` calls both setters on open and close to relax/restore durability for a bulk load. Any *other* concurrent user of `database.async()` (another GraphBatch, an async insert, an index compaction) has its in-flight/queued tasks force-exited with `"Async executor has been shut down"` whenever this fires.

#5665 shipped the cheap interim mitigation only: GraphBatch now toggles the setters once per batch session instead of once per flush. It explicitly deferred the real fix:

> The recreation is not gratuitous: `AsyncThread.run()` applies both flags **once at thread start**, so today the only way to change them is to respawn. The fix is to apply them per task instead of per thread, at which point both setters become plain volatile writes and `createThreads()` drops out of the path entirely.

That fix was never done, and #5665 was closed after the interim mitigation shipped - so this work stopped being tracked as open.

## Why this still matters (#6505)

A production log (issue #6505) showed this exact mechanism causing real damage under concurrent multi-writer load (6 threads on one database, including at least one GraphBatch import running alongside a `DELETE ... BATCH` loop):

- **2,183 `InterruptedIOException`s over ~7 hours**, each aborting an in-progress LSM full-index compaction mid-write (`"Error on executing compaction of index"`), from `createThreads()`'s teardown interrupting an unrelated compaction thread's blocking page I/O.
- One of those races escalated into **fencing the entire database** (#6505's root cause): a commit's `onAfterCommit()` hook tried to schedule a compaction at the exact moment the pool was mid-recreate, and the resulting exception crossed the WAL point of no return. #6505 fixed that specific escalation (a scheduling failure can no longer fence the database), but the underlying teardown-storm - wasted compaction work, aborted I/O, log noise - is still live.

The interim "once per batch" mitigation clearly is not sufficient once there is real concurrency: a batch session can still overlap with arbitrarily much unrelated async work for its whole lifetime, and the teardown at its open/close is exactly as disruptive as before to whatever is running at that instant.

## What's needed

[...]

Issue #6505, opened 2026-08-20
Fenced database strands queued page flush acks, matching 'No flush progress' hang reported on Discord


lvca:
## Reporter

Discord user ArShadow, 2026-08-14, database `heimdall`, 26.9.1-SNAPSHOT. Reported twice: once during a `GraphBatch` import (`withWAL(false)`), once during a `DELETE FROM <class> BATCH 10000` loop across ~896 types (an OrientDB->ArcadeDB migration tool). Both times the database became unusable with:

```
com.arcadedb.exception.DatabaseOperationException: Database 'heimdall' is fenced after a failure past the WAL commit point: close and reopen the database to run recovery (commit of tx 13737 failed after its WAL append)
        at com.arcadedb.database.LocalDatabase.checkDatabaseIsOpen(LocalDatabase.java:2644)
        at com.arcadedb.database.LocalDatabase.checkDatabaseIsOpen(LocalDatabase.java:2633)
        at com.arcadedb.database.LocalDatabase.getFileManager(LocalDatabase.java:1678)
        at com.arcadedb.engine.TransactionManager.cleanWALFiles(TransactionManager.java:1021)
        at com.arcadedb.engine.TransactionManager$1.run(TransactionManager.java:110)
        at java.base/java.util.TimerThread.mainLoop(Timer.java:566)
        at java.base/java.util.TimerThread.run(Timer.java:516)
No flush progress for 60000 ms with 79 pages of database 'heimdall' still pending: giving up the wait. The caller preserves the WAL so the next open recovers the unflushed pages
Preserving the WAL files of database 'heimdall': not all pages reached the disk (unacked WAL pages), the next open will recover them
Database 'heimdall' closed with unflushed pages: the lock file and the WAL were preserved so the next open will recover them
```

Not reproducible on demand; only the logs above were available (the reporter's own log rotation dropped the earlier context, including the `fenced for recovery: <reason>` line that names the actual triggering failure).

## Root cause chain

A database fenced for recovery (#5053: a commit that fails *after* its WAL append but before its pages publish) is still `isOpen() == true` - it just refuses every further operation until a close/reopen replays the WAL. Two places only guarded against `!isOpen()`, not fencing:

[...]

Issue #6502, opened 2026-08-20
A selective RID allow-list makes vector search slower, not faster: no pre-filter plan, up to 499x off the achievable time


tae898:
A RID allow-list is applied only as an admission test inside the HNSW traversal, plus a post-filter on the output. There is no pre-filter plan, so a query gets monotonically slower the more selective its filter is. At the limit, asking for the 5 nearest among 5 candidates costs more than asking for the 10 nearest among 20,000.

## Measured

`main` at `b3f1e2a14c`. 20,000 x 128, EUCLIDEAN, `maxConnections=32`, `beamWidth=100`, k=10, `efSearch=100`. Index-level p50:

| allow-list | p50 | vs unfiltered |
|---|---|---|
| none | 2.020 ms | 1.0x |
| 10% (2,000 RIDs) | 6.281 ms | 3.1x |
| 1% (200 RIDs) | 20.614 ms | 10.2x |
| 0.1% (20 RIDs) | 27.937 ms | 13.8x |
| 5 RIDs | 30.145 ms | **14.9x** |

Through SQL the same shape: 3.17 ms unfiltered to 38.20 ms for 5 RIDs.

`bruteForceScans/query` stayed 0.00 all the way down to 20 allowed RIDs, so this is the traversal itself, not the fallback.

For the same 5 RIDs, returning the identical answer, a hand-written pre-filter plan (fetch the 5 vectors, score them directly) measures **0.090 ms p50**, which is **499x** faster. That floor was measured through Python and numpy, so a Java pre-filter would be quicker still.

Recall under filtering is not the problem: 1.0000 at 1% selectivity and below.

## Mechanism

`efSearch` is a rerank depth, not a visit budget. With a narrow allow-list the beam admits almost nothing, so the walk keeps expanding candidates trying to fill `k`, and the more selective the filter the longer it walks. The filter makes the search space smaller in principle and larger in practice.

Separately, the #3722 shortfall gate sizes its expectation on the whole index and ignores the allow-list. A query whose allow-list is narrower than `k` therefore cannot possibly return `k`, is treated as a graph shortfall, logs a WARNING blaming the graph, and pays a redundant scan. That part is a small independent fix: clamping `availableVectors` by `allowedRIDs.size()` when the set is non-empty would make a 5-RID query expect 5, get 5, and neither log nor scan.

## Suggested direction

[...]

PR #6511, opened 2026-08-20
fix(#6509): apply async executor durability flags per task instead of tearing down the pool

lvca:
## Summary

`DatabaseAsyncExecutorImpl.setTransactionUseWAL()`/`setTransactionSync()` unconditionally called `createThreads()`, which shuts down and respawns the **entire** async worker pool of a database, not scoped to the caller. `GraphBatch` calls both setters once per batch session to relax/restore durability for a bulk load, so opening or closing a batch churned the whole pool and force-exited the in-flight/queued tasks of any *other* concurrent user of `database.async()` - surfacing as `"Async executor has been shut down"` for callers that had nothing to do with the batch.

This was diagnosed in #5665, which shipped only a cheap interim mitigation (GraphBatch toggling the flags once per batch instead of once per flush) and explicitly deferred the real fix. A production incident (#6505) traced 2,183 `InterruptedIOException`s over ~7 hours, and one database-fencing escalation, to this exact mechanism, showing the interim mitigation is not sufficient under real multi-writer concurrency.

### The fix

`AsyncThread.run()` applied `transactionUseWAL`/`transactionSync` only once, at thread start - so respawning the thread was the only way a change ever became visible to it. `AsyncThread.executeTask()` now re-applies both flags to the worker's transaction **before every task**, so the setters become plain volatile writes and `createThreads()` drops out of the path entirely, exactly as #5665's own diagnosis proposed.

### Test/doc cleanup

- Eight async executor tests used `setTransactionUseWAL(true)` purely as a "force thread re-creation" hack to make a queue-size/parallel-level config change take effect deterministically. Switched them to the existing (previously underused) `recreateThreadsForTests()` package-private test hook, since the side effect they relied on is exactly what this fix removes.
- Updated stale comments/Javadoc that described the old `createThreads()`-on-every-flip behavior (in `DatabaseAsyncExecutorImpl`, the `compact()` shutdown-race comment, and `DatabaseAsyncExecutorLifecycleRaceTest`).

## Test plan

- [x] New regression test `Issue6509AsyncExecutorPerTaskDurabilityFlagsTest`:
  - `directDurabilityFlagFlipsDoNotRecreateAsyncThreads` - 25 direct flips of both flags interleaved with real scheduled/awaited tasks; asserts worker thread identities are unchanged.
  - `durabilityFlagChangeTakesEffectOnNextTaskWithoutThreadRespawn` - flips a flag, observes it from inside a task running on the same (unrespawned) worker, flips again, observes the new value on the very next task.
  - Verified both tests **fail** against the pre-fix code (thread identities differ after flips) and **pass** with the fix.
- [x] Full `com.arcadedb.database.async.*` package + `Issue5665GraphBatchAsyncPoolChurnTest` + `Issue5666ConcurrentGraphBatchTest` + `Issue5667GraphBatchSuperNodeResumeTest` + `GraphBatchCommitRetryTest`/`GraphBatchBoundedStateTest`/`GraphBatchUniqueIndexTest`/`GraphBatchWALRestoreTest` + `ConcurrentCompactionTest` + `LSMTreeIndexCompactionTest` + `Issue6303AsyncQuiesceTest`: 84 tests, 0 failures.
- [x] `mvn -o -pl engine -am compile` and `test-compile` both clean.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3737,4 +3737,14 @@ are now plain volatile writes with no `createThreads()` call at all, so a change
 whichever worker runs the next task - and by the commit that follows it - without any thread ever being torn down,
 and without disturbing whatever unrelated work is queued on the rest of the pool.
 
+Re-stamping the flags onto an already-open transaction is not, by itself, enough: `useWAL`/`walFlush` are read only
+at commit time and then govern the *whole* accumulated transaction, which can span up to `ASYNC_TX_BATCH_SIZE`
+(10,240 by default) tasks from unrelated callers sharing the same worker slot. Without a further guard, a flag flip
+landing mid-transaction would silently carry earlier tasks - queued under the old policy - into a commit governed by
+the new one, downgrading (or upgrading) their durability without anyone asking for it. The pool-teardown this PR
+removes had incidentally prevented that: a flag flip forced every worker's pending transaction to commit under its
+original flags before the new thread began a fresh one. `executeTask()` now preserves that "a flag change always
+starts a fresh transaction" property explicitly - closing out the currently open transaction, under its own
+unmodified flags, before applying a changed policy to a new one - instead of relying on thread teardown for it.
+
 [#6509](https://github.com/ArcadeData/arcadedb/issues/6509)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3747,4 +3747,12 @@ original flags before the new thread began a fresh one. `executeTask()` now pres
 starts a fresh transaction" property explicitly - closing out the currently open transaction, under its own
 unmodified flags, before applying a changed policy to a new one - instead of relying on thread teardown for it.
 
+That boundary-forcing commit closes out *earlier* tasks' work, not the task whose flag check triggered it - so its
+failure (a genuine `ConcurrentModificationException`, say, under the exact multi-writer contention this fix targets)
+must not be attributed to that task, which has not run yet. Left unguarded, the failure would propagate to
+`executeTask()`'s own catch block, and the triggering task would still reach `completed()` having never reached
+`execute()` - a task silently marked done without ever running. The boundary commit's failure is now caught
+separately, reported through the executor's `onError`, and the triggering task still gets its own attempt on a
+freshly begun transaction.
+
 [#6509](https://github.com/ArcadeData/arcadedb/issues/6509)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3711,3 +3711,30 @@ task's guard now all catch `Throwable`; in the latter two an escaping `Error` wo
 thread past the point where it clears the in-progress flag, silently stopping every later rebuild of that index.
 
 [#6503](https://github.com/ArcadeData/arcadedb/issues/6503)
+
+## The async executor no longer tears down and respawns its whole worker pool to flip a durability flag (#6509)
+
+`DatabaseAsyncExecutorImpl.setTransactionUseWAL()`/`setTransactionSync()` unconditionally called `createThreads()`,
+which shuts down and respawns every worker thread of the database's async pool - not just the caller's. `GraphBatch`
+calls both setters once per batch session to relax and then restore durability for a bulk load, so opening or
+closing a batch churned the entire pool. Any *other* concurrent user of `database.async()` - another `GraphBatch`,
+an async insert, an index compaction - had its in-flight and queued tasks force-exited mid-flight, surfacing as
+`"Async executor has been shut down"` for callers that had nothing to do with the batch.
+
+#5665 diagnosed this and shipped a cheap interim mitigation (GraphBatch toggling the flags once per batch instead of
+once per flush), explicitly deferring the real fix: `AsyncThread.run()` applied both flags only once, at thread
+start, so respawning the thread was the only way a change could ever become visible to it. A production log (#6505)
+showed the interim mitigation is not enough under real multi-writer concurrency: six threads on one database,
+including a `GraphBatch` import running alongside a `DELETE ... BATCH` loop, produced 2,183 `InterruptedIOException`s
+over about seven hours from the pool teardown interrupting unrelated in-progress LSM compactions, and one of those
+races escalated into fencing the entire database when a commit's `onAfterCommit()` hook tried to schedule a
+compaction at the exact moment the pool was mid-recreate (fixed separately by #6505 hardening the compaction
+scheduler against that specific escalation).
+
+The proper fix, per #5665's own diagnosis: `AsyncThread.executeTask()` now re-applies `transactionUseWAL`/
+`transactionSync` to the worker's transaction before every task, instead of `run()` applying them once. The setters
+are now plain volatile writes with no `createThreads()` call at all, so a changed durability policy is picked up by
+whichever worker runs the next task - and by the commit that follows it - without any thread ever being torn down,
+and without disturbing whatever unrelated work is queued on the rest of the pool.
+
+[#6509](https://github.com/ArcadeData/arcadedb/issues/6509)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3742,4 +3742,11 @@ the about-to-be-stale value (skipping the boundary commit) while the stamp went 
 reintroducing the same durability-mixing failure through a race instead of a guaranteed ordering. Both reads now
 share one snapshot taken at the top of `executeTask()`, so the check and the stamp always agree.
 
+Worth knowing when tuning `parallelLevel`/queue sizing around concurrent bulk loads: the boundary-forcing commit is
+paid per worker slot, not per database, so two `GraphBatch` sessions (or a batch and a direct caller) that land on
+the *same* slot and keep flipping the durability policy in opposite directions can degrade that slot toward
+`commitEvery=1` - each task forcing its own boundary. Correctness over batching is the right trade, and it is still
+strictly better than the pre-fix full pool teardown, but a workload that provokes this is a sign the workers sharing
+that slot would benefit from more parallelism (raising `parallelLevel`) rather than sharing one.
+
 [#6509](https://github.com/ArcadeData/arcadedb/issues/6509)
--- a/engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java
+++ b/engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java
@@ -302,6 +302,14 @@ public class DatabaseAsyncExecutorImpl implements DatabaseAsyncExecutor {
         // runs inside (#5062 review, point 1) - the helping path defers polled tasks back to the run loop
         // instead of nesting, so `nested` is always false today; every `!nested` guard in this method is
         // defense-in-depth should a re-entrant call path ever be reintroduced.
+        //
+        // Deliberately does not touch `count`: a flag-triggered boundary commit here closes a transaction
+        // out of band from the commitEvery cadence, so the NEXT periodic boundary below can land anywhere
+        // from immediately (if `count % commitEvery` was already due) to up to `commitEvery` tasks later -
+        // it is measuring "how many tasks since the last periodic check", not "how many since the
+        // transaction last committed". Harmless: it can only make the next real commit arrive sooner or
+        // later than the nominal cadence, never mix data across a durability-policy boundary (that is what
+        // this whole method exists to prevent) or lose a task.
         final boolean forceFreshTransactionAfterBoundaryFailure =
             !nested && closeTransactionBoundaryIfDurabilityPolicyChanged(currentUseWAL, currentSync);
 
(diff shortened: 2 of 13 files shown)


lvca:
Addressed the main concern above in ddd07fc88: \`executeTask()\` now forces a transaction boundary (commit under the transaction's own, unmodified flags) whenever it finds the currently open transaction was stamped with different \`useWAL\`/\`walFlush\` values than are currently requested, before applying a changed policy to a fresh transaction. This restores the "a flag change always starts a fresh transaction" property the old pool-teardown gave incidentally, without reintroducing thread teardown.

Added \`durabilityFlagFlipMidTransactionForcesABoundaryInsteadOfMixingDurability\`, which observes actual WAL-write counts via the \`TX_AFTER_WAL_WRITE\` callback across a flag flip landing mid-transaction. Confirmed it fails without the new guard (task A's WAL-protected write silently merges into a later no-WAL commit: 0 WAL writes instead of the expected 1) and passes with it.

The "minor" note about \`compact()\`'s comment was already accurate post-fix (no change needed there).

lvca:
Addressed both points from the latest review in c55fadcae:

**Correctness gap (pre-flight commit failure)**: the boundary commit now catches its own failure separately from the triggering task's execution. On failure it reports through \`onError\` and still lets the triggering task attempt \`begin()\`/\`execute()\` on a freshly begun transaction, instead of falling through to the outer \`catch\` where the task would reach \`completed()\` having never run. Added \`failedBoundaryCommitDoesNotSilentlyDropTheTriggeringTask\`, which injects a commit failure via the \`TX_AFTER_WAL_WRITE\` callback at exactly the boundary commit and asserts the triggering task still executes and its write survives. Confirmed it fails without the fix (task never executes, yet still reaches \`completed()\`) and passes with it.

**Stale \`GraphBatch.java\` comments**: updated the three spots (field javadoc on \`savedAsyncUseWAL\`, and the two "guarded against pool teardown" comments in the constructor and \`restoreAsyncSettings()\`) that still described the pre-#6509 \`createThreads()\` behavior.

**Two-step non-atomicity note**: agreed this isn't a regression (same as the pre-PR two-respawn sequence) and isn't worth guarding further - left as-is.

87 tests green across the async/GraphBatch/compaction suites (`com.arcadedb.database.async.*`, `Issue5665/5666/5667GraphBatch*Test`, `GraphBatchCommitRetryTest`/`GraphBatchBoundedStateTest`/`GraphBatchUniqueIndexTest`/`GraphBatchWALRestoreTest`, `ConcurrentCompactionTest`, `LSMTreeIndexCompactionTest`, `Issue6303AsyncQuiesceTest`).

lvca:
Thanks - both remaining points from this review are judged not worth further changes for this PR:

**Two-step non-atomicity** (\`setTransactionUseWAL()\`/\`setTransactionSync()\` as two separate volatile writes): confirmed pre-existing (the pre-#6509 \`createThreads()\` path had the same two-respawn non-atomicity), and \`walFlush\` is only ever consulted when \`useWAL\` is true (\`TransactionContext.commit1stPhase\`), so the one-task transient window where the pair doesn't match either the old or new combination can't actually produce a mixed WAL/no-WAL outcome - at worst a task in that narrow window sees a \`walFlush\` value that's about to change again but is irrelevant if \`useWAL\` is false. Not worth a combined-setter API for this.

**\`onError\`-only reporting on the boundary-commit failure path**: matches the existing convention for every other in-\`executeTask\` failure in this class (there's no separate reporting channel elsewhere in the file either) - consistent, not a new gap introduced here.

No blocking issues remain per the last review. Once CI (`build-and-package`) finishes, this should be ready to merge.

lvca:
Addressed the TOCTOU race from the latest review in 37d14ba02 (merged with a `main` sync in c78b2e85c after a second conflict from concurrent doc edits on \`main\`):

**TOCTOU race between the boundary check and the flag stamp**: \`transactionUseWAL\`/\`transactionSync\` are volatile and can change between the two reads (the boundary-commit check and the later stamp). Both now use one snapshot taken at the top of \`executeTask()\`, so the check and the stamp always agree - closing the race that could otherwise reintroduce the exact durability-mixing bug the boundary-commit guard exists to prevent.

Also applied the two "minor/informational" suggestions from the same review pass:
- The boundary-commit failure path now explicitly rolls back if the transaction is somehow still active after the failure, instead of relying by inference on \`TransactionContext\`'s internal cleanup being the only way a commit failure ends a transaction.
- The flag re-stamp is now guarded by \`!nested\`, matching the boundary-commit check above it, for the same defense-in-depth reasoning as the existing \`!nested\` guards elsewhere in the method.

86 tests still green locally after both the fix and a rebase/merge against two rounds of \`main\` drift (unrelated PRs merging, plus a manual edit to the release notes doc that this PR's changelog entry had to be re-appended after).

lvca:
Applied the optional hardening from the latest review in a4940c4c2: the fallback \`rollback()\` inside the boundary-commit \`catch\` block is now wrapped in its own try/catch (logged and swallowed on failure) instead of letting a secondary rollback failure escape to the outer catch, where it would have reproduced the exact "task marked completed without running" bug this block exists to prevent, one level deeper.

The other two points (comment density, and confirming \`compact()\`'s comment is still accurate) needed no changes.

86 tests still green. This should be the last round - the review explicitly said "No blocking issues found" before this optional item.

lvca:
Addressed the double-failure edge case from the latest review in 12b2988b4 (merged with two unrelated \`main\` fixes to different files - #6516, #6518 - in 6b0460b29, no conflicts):

**Double failure in the boundary commit's own rollback**: \`TransactionContext.rollback()\` isn't guaranteed to reach its own \`reset()\` - a schema-dictionary reload IOException during rollback throws before getting there, leaving the transaction in a half-cleaned, still-active state. If that happens right after the boundary commit itself already failed, \`database.isTransactionActive()\` would read true over that poisoned state, and the existing \`begin()\` guard would then skip beginning a fresh transaction for the triggering task - running it against unknown leftover state instead of isolating it.

\`executeTask()\` now tracks whether the rollback fallback itself failed with the transaction still active, and forces \`begin()\` in that case regardless of what \`isTransactionActive()\` reports. This reuses \`LocalDatabase.begin()\`'s existing, already-tested behavior for "an active transaction exists" (it pushes a new nested \`TransactionContext\` rather than reusing the current one) rather than inventing a new recovery path - the same thing that happens for any other caller that begins while one is already open. The poisoned transaction is left on the stack underneath for whatever the next boundary or \`commitEvery\` cycle does with it.

Not covered by a new dedicated regression test: reliably triggering \`TransactionContext.rollback()\`'s own internal failure path from outside requires corrupting the schema dictionary file mid-rollback, which isn't practical to reproduce deterministically without file-level fault injection this codebase's test style doesn't otherwise use. Verified by inspection against \`LocalDatabase.begin()\`'s existing nested-transaction behavior, and by the full suite (86 tests) staying green with no behavioral change on any non-failure path.

lvca:
Added the one-line clarifying comment from the latest review (6074d9f55, non-functional): documents why the flag re-stamp can still land on a still-active, poisoned transaction in the \`forceFreshTransactionAfterBoundaryFailure\` branch when the triggering task doesn't require an active tx, and why that's acceptable (no writes go through it either way; it only affects whether the *next* task's boundary check re-triggers a commit attempt on that same already-poisoned transaction).

This review explicitly said "none block" on all points and the fix has now been through 7 rounds of increasingly narrow correctness hardening (production incident → mid-transaction mixing → TOCTOU race → single rollback failure → double rollback failure → this doc clarification), each one closing something real but at rapidly diminishing severity. Considering this converged - no further functional changes are planned unless a future round surfaces something with plausible real-world reachability rather than another layer of already-rare failure compounding.

87 tests green locally (`com.arcadedb.database.async.*` suite re-verified after this change).

lvca:
Round 8 confirmed no blocking issues. Addressed the one cheap, valuable suggestion:

**Boundary-commit-on-flag-thrash throughput note**: added a release-notes callout (e59316fae, docs-only) explaining the per-slot trade-off - two concurrent durability-flag flippers sharing a worker slot can degrade that slot toward \`commitEvery=1\`, and that a workload provoking this should raise \`parallelLevel\` rather than share a slot.

The other two points need no action, as the review itself notes:
- Double-failure path test gap: already disclosed in the prior commit's message; the review just flags it "so it isn't lost," not as something to act on now.
- \`count\` not reset by a flag-triggered boundary commit: explicitly called out as "purely a minor efficiency note, not a correctness issue."

This is the 8th review round and the second in a row to find nothing blocking. Considering this fully converged now - holding the line per the prior comment unless something with genuine new real-world reachability surfaces.

lvca:
Addressed both actionable findings from the latest review in 78d051354:

**1. Periodic boundary didn't re-stamp flags (latent fragility)**: the \`commitEvery\`-triggered internal \`commit()\`/\`begin()\` cycle now explicitly re-applies \`currentUseWAL\`/\`currentSync\` afterward too, instead of relying on the undocumented (if currently-true) invariant that \`TransactionContext.begin()\`/\`reset()\` never reset those fields. No new regression test for this one: it closes a latent robustness gap in an invariant that already holds today, so there's no observable failure to demonstrate before the fix - it only matters if \`TransactionContext\`'s field-reset behavior ever changes elsewhere, which is exactly what this stops mattering.

**2. `executeTask()` had grown large and comment-dense**: extracted the boundary-commit-and-recovery block into \`closeTransactionBoundaryIfDurabilityPolicyChanged()\` (named as suggested), with its own focused Javadoc replacing the inline "#6509 review:" trail. \`executeTask()\` now reads as snapshot → close stale boundary → begin → stamp → execute → commit-every-N.

**3. Per-task overhead**: explicitly noted as non-blocking by the review itself ("not blocking, just worth confirming"); a couple of volatile reads and comparisons per task is negligible next to what it replaces (full pool teardown), so no benchmark work planned for this PR.

86 tests still green, no behavioral change on any non-failure path.

lvca:
Round 10 said all points "none blocking." Applied the one cheap consistency fix in 26c7c36ec: \`run()\`'s initial flag stamp now snapshots \`transactionUseWAL\`/\`transactionSync\` into locals first, matching \`executeTask()\`'s pattern, instead of reading the two volatiles separately. The review itself noted the practical impact is nil (any mismatch would be reconciled by the first task's own boundary check) and didn't ask for a dedicated test - just consistency, which this now has.

The remaining points were explicitly framed as optional/future-work by the review itself:
- \`DatabaseAsyncExecutorLifecycleRaceTest\`'s narrowed test scope - already disclosed in its own updated docstring, "minor, not blocking."
- The untested double-failure path - already disclosed in the prior commit's message, suggested only as something a maintainer might want to add a fault-injection seam for "later."
- Per-task overhead and the `GraphBatch.java` comment-only nit - both explicitly non-issues per the review.

86 tests still green. This is the 10th review round; the last four in a row have found nothing blocking beyond incremental polish. Holding the line here per the convergence stance in earlier comments.

lvca:
Addressed both points the latest review asked for confirmation on, in 664bd6ced (comments only, no logic change):

**1. \`count\` not reset by a flag-triggered boundary commit**: confirmed intentional and documented why - it can only shift when the next periodic commit lands relative to the nominal cadence, never mix data across a durability boundary or lose a task.

**2. Double-failure recovery leaves a poisoned parent \`TransactionContext\` on the stack**: verified this is *not* a permanent leak by reading both cleanup paths - \`DatabaseContext\`'s dead-thread sweep (\`rollbackAbandonedTransactions\`) and \`LocalDatabase\`'s close path both roll back and clear the *entire* transaction stack, not just the top, so the poisoned entry is reclaimed once the worker thread ends or the database closes. Documented that in the helper method's Javadoc.

**3. \`AsyncCommandPool.runCommand()\` never applying the durability flags**: correctly identified as pre-existing and out of scope (it bypassed \`executeTask()\` before this PR too) - a follow-up issue is the right vehicle for that, not this PR.

**4. Harmless no-op boundary commit possible at thread startup**: explicitly "not a bug," no action needed.

86 tests still green. This is the 11th review round, and per the reviewer's own words they'd be "comfortable with this" once points 1 and 2 were confirmed - which they now are.

lvca:
Addressed the one precise point from the latest review in 0097c2abd (merged with 5 unrelated main commits - #6502, #6517, different files - no conflicts):

**"Not a permanent leak" was slightly overstated**: the dead-thread sweep's own reclaim of the poisoned parent transaction calls \`rollback()\` again, so if the original failure was persistent (a genuinely full disk, a schema dictionary that keeps throwing) rather than transient, the sweep's own attempt can fail the same way - its own contract on that path is a logged warning and moving on, not a guarantee, exactly as its own comment already documents (held file locks). Softened the claim with that caveat rather than leaving an inaccurate unqualified statement in the code.

The other two points (comment consolidation, `setWALFlush` call-site style consistency) are cosmetic-only, explicitly framed as such by the review, and declined given how many rounds this has already been through.

86 tests still green. This is the 12th review round.

lvca:
Round 13 found nothing blocking - all three points explicitly framed as non-required (an observability/metrics suggestion worth a follow-up issue, a doc-sync note, and a commit-history squashing preference, none "code issues"). No further changes from this round.

This is now 13 review rounds with the last several finding only progressively smaller polish. Considering the review process fully converged - waiting on CI to go green.

lvca:
All Java CI test lanes are green (unit-tests, slow-unit-tests, vector-unit-tests, integration-tests, ha-integration-tests, opencypher-tck-tests, builder-tests, and all e2e lanes).

Two new checks appeared after the full suite finished: \`codecov/patch\` (70.00% of diff hit, target 71.18%) and \`codecov/project\` (71.17%, -0.02% vs main). The 9 uncovered lines in \`DatabaseAsyncExecutorImpl.java\` are precisely the double-failure recovery path (\`closeTransactionBoundaryIfDurabilityPolicyChanged()\`'s rollback-fallback-also-fails branch, and the \`forceFreshTransactionAfterBoundaryFailure=true\` path it feeds) - already discussed across review rounds 6, 9, 11, and 12 as impractical to cover with a dedicated test, since reliably triggering \`TransactionContext.rollback()\`'s own internal failure requires file-level fault injection this codebase doesn't otherwise use. Not adding a contrived/mocked test just to move this number, since that would trade real confidence for a metric.

\`main\` has no branch protection configured, so this isn't a technical merge blocker.

PR #6512, opened 2026-08-20
fix(#6502): add a pre-filter plan for selective RID allow-lists in vector search

### LINKED ISSUE #6502: A selective RID allow-list makes vector search slower, not faster: no pre-filter plan, up to 499x off the achievable time

lvca:
## Summary

Fixes #6502. A RID allow-list passed to `findNeighborsFromVector` was applied only as an admission test inside the HNSW graph walk plus a post-filter on the output. JVector's `Bits` filter only rejects a node once the walk has already popped it from the beam - it cannot make the walk itself shrink with a narrower filter - so a selective allow-list made the beam admit almost nothing and the walk kept expanding trying to fill `k`. At the limit, asking for 5 candidates among 5 allowed RIDs cost more than asking for 10 among 20,000.

- `findNeighborsFromVector` now resolves a narrow allow-list to its ordinals and scores them directly, skipping the graph walk entirely, whenever the allow-list covers no more than `arcadedb.vectorIndex.prefilterMaxSelectivity` (new setting, default 20%, clamped to 1.0) of the index's live vectors. This reuses `collectAllowedOrdinals`/`scoreOrdinal`/`bruteForceScan`, the same walk issue #3722's shortfall fallback already used, as the primary plan instead of a last resort - exact by construction, matching the 1.0000 recall the issue already measured at the selective end. Set the new config to `0` to disable and always take the graph walk (restores the pre-#6502 behavior exactly).
- Added a `preFilterSearches` counter to `getStats()` alongside the existing `bruteForceScans`, since this is a deliberately chosen cheaper plan, not a fallback.
- The shortfall WARNING that still fires past the pre-filter threshold (e.g. selectivity disabled, or a wide-but-still-narrower-than-`k` allow-list) now distinguishes an allow-list-driven shortfall (expected, not a graph problem) from a real degraded-graph shortfall, addressing tae898's follow-up comment on the issue about the log line being misleading. The classification (`LSMVectorIndex.shortfallIsAllowListDriven`) is a pure, directly unit-tested predicate rather than untestable inline logic.

**Scope**: this targets `findNeighborsFromVector`, which is what the issue measured and what SQL/Cypher's `vectorNeighbors()` calls into. The grouped (`groupBy`) and PQ-approximate search paths have the same shape of problem but different internals (group accounting, PQ scoring) and are left as a follow-up: tracked in #6514.

## Known limitation

The selectivity check (and the FINE/WARNING shortfall classification) compares against the allow-list's **raw size**, not the count of RIDs actually resolvable in *this* index, and `availableVectors` counts persisted/graph vectors only (not the delta buffer). A caller-supplied allow-list built from a broader/polymorphic query, or a large delta buffer relative to the persisted graph, can therefore make the effective selectivity look narrower or wider than it truly is against the live set. This is a deliberate trade-off: resolving the true count up front would mean paying the O(allow-list · log index) ordinal-resolution walk just to decide whether to pay for it, which is circular. It never produces a wrong answer - only a possibly-missed fast path (falls back to the already-correct graph walk) or a log severity that's more cautious than strictly necessary. Flagging explicitly here since it's the kind of thing a future "why didn't my narrow filter take the fast path" report would otherwise have to be re-derived from the code.

## Measured

Local benchmark reproducing the issue's scenario (20,000 x 128, EUCLIDEAN, k=10):

| allow-list | before (issue's numbers) | after (this PR) | plan taken |
|---|---|---|---|
| unfiltered | 2.020 ms | 0.352 ms | - |
| 25% (5,000 RIDs) | - | 0.851 ms | graph walk |
| 20% (4,000 RIDs) | - | 0.865 ms | pre-filter (at the `<=` boundary) |
| 15% (3,000 RIDs) | - | 0.551 ms | pre-filter |
| 10% (2,000 RIDs) | 6.281 ms | 0.327-0.369 ms | pre-filter |
| 1% (200 RIDs) | 20.614 ms | 0.024-0.026 ms | pre-filter |
| 0.1% (20 RIDs) | 27.937 ms | 0.002 ms | pre-filter |
| 5 RIDs | 30.145 ms | 0.001 ms | pre-filter |

(Different hardware than the issue's own measurement, so treat the "before" column as directional, not a same-machine comparison; the "after" column was measured on this branch, `Issue6502PrefilterLatencyBenchmark`.) The 15-25% points straddle the default threshold on both sides: the two plans cost about the same right at the boundary (0.851 vs 0.865 ms), so the default isn't sensitive to being off by a few points - the pre-filter plan's win is already decisive well before it.

## Test plan

- [x] `Issue6502VectorPrefilterTest`: narrow allow-list takes the pre-filter plan and matches the graph-walk answer exactly (forced via selectivity=0); wide allow-list stays on the graph walk; selectivity=0 disables the plan; the plan does not trip the issue #3722 shortfall WARNING/scan; the plan merges allowed vectors still in the delta buffer; the delta+persisted candidate union truncates to `k` by overall distance, not per-source; a selectivity above 1.0 is clamped, not taken literally; the FINE/WARNING shortfall predicate is unit-tested in both directions plus end-to-end (verified against a deliberately inverted predicate before trusting it).
- [x] Updated `LSMVectorIndexBruteForceScanTest.aNarrowAllowListStillReachesTheFallbackEndToEnd`, whose premise (a narrow allow-list must reach the graph's shortfall fallback) is exactly what this PR optimizes away; it now asserts the pre-filter plan is taken and the shortfall path is *not* reached, while keeping its original RID/score-pairing (#4581) assertions.
- [x] `Issue6502PrefilterLatencyBenchmark` (`@Tag("benchmark")`, excluded from CI) reproducing the issue's exact scenario, extended with data points straddling the default threshold.
- [x] Full `com.arcadedb.index.vector.**` package (242 tests), plus `vectorNeighbors()` SQL/Cypher integration tests (`SQLFunctionVectorNeighborsTest`, `PartitionPruningVectorNeighborsTest`, `CypherCallVectorNeighborsTest`, `SQLVectorDatabaseFunctionsTest`) and `GlobalConfigurationTest` all pass.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1060,6 +1060,19 @@ public enum GlobalConfiguration {
       searcher per query.""",
       Integer.class, 0),
 
+  VECTOR_INDEX_PREFILTER_MAX_SELECTIVITY("arcadedb.vectorIndex.prefilterMaxSelectivity", SCOPE.DATABASE,
+      """
+      Maximum fraction of the index's live vectors that a query's RID allow-list may cover and still take the \
+      pre-filter plan instead of the HNSW graph walk (issue #6502). The graph's Bits filter only rejects a node \
+      once it is popped from the beam - it cannot make the walk itself shrink with a narrower filter - so a \
+      selective allow-list makes the beam admit almost nothing and the walk keeps expanding trying to fill k, \
+      the search space growing smaller in principle and larger in practice: at the limit, 5 candidates cost more \
+      than 20,000. Below this fraction the query instead resolves the allow-list to its ordinals and scores them \
+      directly, which is O(allow-list) and exact by construction. Above it the allow-list barely narrows the \
+      search, the graph walk is already cheap, and resolving every allowed RID up front would cost more than the \
+      walk it replaces. Set to 0 to disable the pre-filter plan and always use the graph walk.""",
+      Float.class, 0.2f),
+
   VECTOR_INDEX_GRAPH_BUILD_PARALLELISM("arcadedb.vectorIndex.graphBuildParallelism", SCOPE.DATABASE,
       """
       Number of threads in the dedicated pool that builds the HNSW graph of a vector index. Graph construction \
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -4145,6 +4145,27 @@ public class LSMVectorIndex implements Index, IndexInternal {
 
         final RandomAccessVectorValues vectors = searchVectorValues(ordinalMap);
 
+        // Issue #6502: a RID allow-list narrow enough relative to the index is cheaper to answer directly than to
+        // walk the HNSW graph under it. The graph's Bits filter only rejects a node once it is popped from the
+        // beam - it cannot make the walk itself shrink with a narrower filter - so a selective allow-list makes the
+        // beam admit almost nothing and the walk keeps expanding trying to fill k: the search space gets smaller in
+        // principle and larger in practice. Scoring the allowed ordinals directly is O(allow-list) and exact by
+        // construction, matching the 1.0000 recall already measured at the selective end. collectAllowedOrdinals,
+        // scoreOrdinal and bruteForceScan already do exactly this walk for the issue #3722 shortfall fallback, so
+        // it is reused here as the primary plan rather than duplicated.
+        if (allowedRIDs != null && !allowedRIDs.isEmpty()) {
+          final float maxSelectivity = getDatabase().getConfiguration()
+              .getValueAsFloat(GlobalConfiguration.VECTOR_INDEX_PREFILTER_MAX_SELECTIVITY);
+          final int availableVectors = Math.min(ordinalMap.length, vectorIndex.size());
+          if (maxSelectivity > 0f && allowedRIDs.size() <= availableVectors * maxSelectivity) {
+            metrics.incrementPreFilterSearches();
+            final List<Pair<RID, Float>> results = new ArrayList<>(k);
+            mergeWithDeltaScan(queryVectorFloat, k, allowedRIDs, results);
+            bruteForceScan(queryVectorFloat, k, allowedRIDs, results, vectors, ordinalMap);
+            return results;
+          }
+        }
+
         // Only live vectors may enter the result heap. Accepting tombstones lets a query aimed at a deleted
         // neighbourhood fill its beam with them and stop, which is what returned an empty list (issue #5558).
         final Bits bitsFilter = new LiveVectorBitsFilter(allowedRIDs, ordinalMap, vectorIndex);
@@ -4253,12 +4274,26 @@ public class LSMVectorIndex implements Index, IndexInternal {
         final int availableVectors = Math.min(ordinalMap.length, vectorIndex.size());
         final int expectedResults = Math.min(k, availableVectors);
         if (results.size() < expectedResults) {
-          LogManager.instance()
-              .log(this, Level.WARNING,
-                  """
-                  Graph search returned only %d results (expected %d, available %d) for index %s - \
-                  falling back to brute-force scan (graph may need rebuilding)""",
-                  results.size(), expectedResults, availableVectors, indexName);
+          // Issue #6502: an allow-list narrower than k can never be filled from the graph - that is a property of
+          // the filter, not a sign the graph needs attention, and the pre-filter plan above already answers the
+          // common case of this directly without reaching here at all. What still lands here is either no allow-list
+          // at all (the original issue #3722 degraded-graph case) or one wide enough to skip the pre-filter plan
+          // yet still narrower than k, so the two are told apart rather than both blamed on the graph.
+          if (allowedRIDs != null && !allowedRIDs.isEmpty() && allowedRIDs.size() < expectedResults)
+            LogManager.instance()
+                .log(this, Level.FINE,
+                    """
+                    Graph search returned only %d results (expected %d, available %d) for index %s - the RID \
+                    allow-list has only %d entries, which is fewer than requested, so completing the answer with a \
+                    brute-force scan of the allow-list is expected, not a sign the graph needs rebuilding""",
+                    results.size(), expectedResults, availableVectors, indexName, allowedRIDs.size());
+          else
+            LogManager.instance()
+                .log(this, Level.WARNING,
+                    """
+                    Graph search returned only %d results (expected %d, available %d) for index %s - \
+                    falling back to brute-force scan (graph may need rebuilding)""",
+                    results.size(), expectedResults, availableVectors, indexName);
           metrics.incrementBruteForceScans();
           bruteForceScan(queryVectorFloat, k, allowedRIDs, results, vectors, ordinalMap);
         }
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -4279,7 +4279,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
           // common case of this directly without reaching here at all. What still lands here is either no allow-list
           // at all (the original issue #3722 degraded-graph case) or one wide enough to skip the pre-filter plan
           // yet still narrower than k, so the two are told apart rather than both blamed on the graph.
-          if (allowedRIDs != null && !allowedRIDs.isEmpty() && allowedRIDs.size() < expectedResults)
+          if (shortfallIsAllowListDriven(allowedRIDs, expectedResults))
             LogManager.instance()
                 .log(this, Level.FINE,
                     """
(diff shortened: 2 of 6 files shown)


lvca:
Addressed finding 1 (test coverage gap for the FINE/WARNING shortfall log split): extracted the condition to a pure `LSMVectorIndex.shortfallIsAllowListDriven` predicate, added direct unit coverage for both directions (narrow allow-list, null, empty, wide-enough allow-list), and an end-to-end test that captures logs at all levels (not just WARNING+, unlike the existing `CapturingLogger`) and asserts the FINE line - not the WARNING - is what fires for a narrow-allow-list shortfall. Verified both new tests fail against a deliberately inverted predicate before confirming the real one passes.

Findings 2 and 3 (the pre-filter threshold and the FINE/WARNING predicate use the allow-list's raw size rather than the count of RIDs actually resolvable in this index): leaving these as-is. Resolving that count up front would mean paying the O(allow-list · log index) ordinal-resolution walk just to decide whether the cheaper path is worth taking, which defeats the point. It's the same raw-size estimate the pre-existing shortfall-budget calculation already uses, and getting it "wrong" only ever falls back to the already-correct pre-#6502 behavior or picks the more conservative WARNING over FINE - never a wrong answer, just a missed optimization or a log severity that's more cautious than strictly necessary.

lvca:
Addressed both non-blocking points from the second review pass:

**1. Crossover data point near the 20% threshold**, per the "main ask before merge": extended `Issue6502PrefilterLatencyBenchmark` with 25%, 20% and 15% selectivity points straddling the default `VECTOR_INDEX_PREFILTER_MAX_SELECTIVITY` on both sides. Measured (20,000x128 EUCLIDEAN, k=10):

| allow-list | p50 | plan taken |
|---|---|---|
| unfiltered | 0.352 ms | - |
| 25% (5,000 RIDs) | 0.851 ms | graph walk |
| 20% (4,000 RIDs) | 0.865 ms | pre-filter (at the `<=` boundary) |
| 15% (3,000 RIDs) | 0.551 ms | pre-filter |
| 10% (2,000 RIDs) | 0.327 ms | pre-filter |

The two plans cost about the same right at the boundary (0.851 vs 0.865 ms), which is the reassuring result: the default isn't finicky about being off by a few points either way, and the pre-filter plan's win is already clear well before it (10% and below, and the earlier 1%/0.1%/single-digit-RID measurements upthread).

**2. Comment duplication**: trimmed the near-verbatim mechanism explanation that was written out three times. The `GlobalConfiguration` entry for `VECTOR_INDEX_PREFILTER_MAX_SELECTIVITY` stays as the canonical, self-contained explanation (it's what an operator reading the config reference actually sees); the `findNeighborsFromVector` call-site comment and `shortfallIsAllowListDriven`'s javadoc now point back to it instead of repeating it.

Full `com.arcadedb.index.vector.**` package (240 tests) still green after both changes.

lvca:
Addressed both minor points from the third review pass (the two "no action needed" ones - `availableVectors` computed twice, and the raw-allow-list-size trade-off already covered in an earlier reply - are left as-is, per the review's own conclusion):

**1. No validation/clamp on `VECTOR_INDEX_PREFILTER_MAX_SELECTIVITY`**: now clamped to 1.0 when read (`Math.min(1f, ...)`), so an operator-supplied value above that can no longer force every allow-list query, however wide, through the ordinal-resolution walk. New test `aSelectivityAboveOneIsClampedNotTakenLiterally` sets selectivity=5.0 and an allow-list sized to distinguish "clamped to 1.0" (rejected) from "taken literally" (accepted) - would fail without the clamp.

**2. Test gap: delta+persisted union/truncation**: added `thePreFilterPlanTruncatesToKAcrossBothDeltaAndPersistedCandidates` - a 10-RID allow-list split 5/5 across the delta buffer and the persisted graph, `k=3`, asserting the closest 3 overall survive truncation regardless of source. Matches the reasoning the review traced through manually (mergeWithDeltaScan's eviction bound + bruteForceScan's dedup-then-resort), now pinned by a test.

Full `com.arcadedb.index.vector.**` package (242 tests) green after both changes.

lvca:
Fourth review pass said "No blocking issues found" with two soft suggestions - both addressed:

1. **Delta buffer not counted in the selectivity check**: added a short comment at the pre-filter's `availableVectors` computation pointing to the shortfall-budget calculation's existing (more detailed) explanation of the same exclusion, rather than repeating it. No behavior change - this was already the accepted trade-off from the second review round, just needed a pointer at the point of confusion.
2. **Tracking issue for the groupBy/PQ-approximate follow-up scope**: opened #6514, since it's not tracked anywhere yet and the reviewer's concern (the underlying Bits-filter pathology applies there too) is valid - didn't want it to get lost.

Full `com.arcadedb.index.vector.**` package (242 tests) still green.

lvca:
Fifth review pass: "I didn't find any correctness bugs" - all 4 minor points were either explicitly "not worth churn" (point 1), already-accepted trade-offs from earlier rounds (points 2, 3), or a compliment on existing tagging (point 4). No code changes needed.

Did act on the one genuinely free ask (point 2, "worth being explicit in the PR description too"): added a **Known limitation** section to the PR description spelling out the raw-allow-list-size trade-off (and its mirror in `shortfallIsAllowListDriven`, point 3) explicitly, so a future "why didn't my filter take the fast path" investigation has something to find. Also folded the 15-25% crossover benchmark data into the description's measurement table.

At this point 5 review rounds have found zero correctness bugs, with the last two rounds surfacing only already-acknowledged trade-offs and compliments. Treating this as ready to merge pending final CI.

lvca:
CI status: everything is green except `slow-unit-tests`, which fails twice in a row (original run and a rerun) on `LSMVectorIndexAutoCompactionTest.anAttemptWhoseTaskIsNeverEnqueuedReleasesItsSlot`.

This is **not caused by this PR**: confirmed `git diff main -- engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexAutoCompactionTest.java` is empty (file untouched by this branch), and this PR touches only `findNeighborsFromVector`/search-path code, nothing in compaction scheduling or the async executor.

It's a pre-existing regression on `main` itself, already filed and tracked separately as #6516: commit `eaaf06261` (#6505, "a best-effort compaction-scheduling race must not fence the database") deliberately changed `DatabaseAsyncExecutorImpl.compact()` to swallow `DatabaseOperationException` instead of letting it escape, but the test still asserts the old throwing behavior. #6516 has the root cause and a suggested one-line fix already written up.

This PR's `slow-unit-tests` lane will stay red until #6516 is fixed on `main` and this branch is rebased past it - not something to fix here, since #6505 (the commit that changed the behavior) and #6516 (its test fallout) both predate this branch and are unrelated to the RID allow-list pre-filter work.

Issue #6496, opened 2026-08-20
Inactivity rebuild timer ignores the rebuild threshold: one insert into a settled index costs a full graph rebuild


tae898:
The inactivity rebuild timer never consults the rebuild threshold. It arms and fires on `pending > 0` alone, so a single insert into a large settled LSM_VECTOR index triggers a full O(N) graph rebuild about 15 seconds later. The mutation-driven path gates on the threshold correctly; only the timer path does not.

## Measured

`main` at `b3f1e2a14c`, 50,000 x 128 EUCLIDEAN, `maxConnections=32`, `beamWidth=100`.

Index built and settled: `graphState=IMMUTABLE`, `graphNodeCount=50000`, `mutationsSinceRebuild=0`, `effectiveMutationsThreshold=10000`, `graphRebuildCount=1`.

Control first, so this is not just the passage of time: 25 seconds of quiet after settling produced zero extra rebuilds.

Then one insert. `mutationsSinceRebuild=1`, against an effective threshold of 10,000, so 0.01% of it.

```
t=15.08s  Inactivity timeout expired (15000 ms), triggering graph rebuild for 1 pending mutations
          Building JVector graph index with 50001 vectors
```

Full rebuild, 8.06 s inside the build, `graphRebuildCount` 1 to 2.

Second run, two consecutive single inserts, to show it is not one-shot:

| round | inserts | delay | rebuilt | wall |
|---|---|---|---|---|
| 0 | 1 | 15.08 s | 50,001 vectors | 12.06 s |
| 1 | 1 | 15.08 s | 50,002 vectors | 14.62 s |

The first build of the whole index cost 12.6 to 16.9 s. So absorbing one vector costs roughly what building the entire index cost, and it repeats on every quiet period.

A client writing one vector per pause pays a full-index rebuild every 15 seconds indefinitely. At 10M vectors, where a build is 30 to 45 minutes, a single insert during a lull commits the index to that.

## Mechanism

Arming, `:6672-6673` in `scheduleInactivityRebuild()`:

```java
if (mutationsSinceSerialize.get() <= 0)
  return; // Nothing to rebuild
```

Firing, `:6687-6689` in the TimerTask:

```java
// Double-check: only rebuild if there are still pending mutations
if (mutationsSinceSerialize.get() <= 0)
  return;
```

[...]

Issue #6106, opened 2026-08-12
LSM_VECTOR: dense ids (post-#5870) make a same-count stale-graph reuse harder to detect after a crash mid-compaction


robfrank:
## Context

Raised during review of PR #6102 (issue #5870, vector id renumbering during `COMPACT INDEX`). Not confirmed as reachable - filing so it is tracked and someone with more context on the page-flush/WAL guarantees around vector graph persistence can confirm or rule it out.

## The mechanism

`LSMVectorIndex.ensureGraphAvailable()` decides whether to reuse a persisted JVector graph or rebuild it from scratch. The staleness check is:

```java
final boolean hasDeletedVectors = vectorIndex.getDeletedCount() > 0;
...
if (graphFile != null && graphFile.hasPersistedGraph() && !needsGraphRebuildForPQ && !hasDeletedVectors) {
  final var loadedGraph = graphFile.loadGraph();
  final int[] rebuiltOrdinalToVectorId = /* recomputed fresh from the CURRENT vectorIndex */;
  final int graphSize = loadedGraph != null ? loadedGraph.size() : 0;
  if (graphSize < rebuiltOrdinalToVectorId.length) {
    // stale - rebuild
  } else {
    // "Graph is up to date — use it": pairs the OLD graph's ordinals with the freshly recomputed ids
    this.ordinalToVectorId = rebuiltOrdinalToVectorId;
    this.graphIndex = loadedGraph;
  }
}
```

The only staleness signal is a **count** comparison (`graphSize < rebuiltOrdinalToVectorId.length`), not a content/generation comparison. If a persisted graph from a previous generation happens to have the **same node count** as the live set computed from the current `VectorLocationIndex`, but a **different composition** (different actual vectors, e.g. because a compaction ran and the crash landed between the compaction's data-file swap and its own `graphFile.writeGraph(...)` completing - see `LSMVectorIndex.compact()` -> `buildGraphFromScratchExclusively`, both happen inside one `buildGraphFromScratch(null, true)` call), the stale graph gets paired with a wrong `ordinalToVectorId` and searches would silently return the wrong vectors for some ordinals - not a crash, not a missing-result, but a **wrong-but-plausible** result.

[...]

PR #6513, opened 2026-08-20
fix(#6503): a vector rebuild no longer holds two graphs with no admission control, and an OOM cannot certify a half-written graph

### LINKED ISSUE #6503: A vector graph rebuild holds the old graph plus a full build's working set, with no admission control, and the OOM escapes markUnusable()

lvca:
Fixes #6503

## What

A from-scratch rebuild of an `LSM_VECTOR` graph keeps the old graph resident so searches keep working, and pays for a whole new build's working set on top of it: the build cache, the JVector builder, the ordinal map. Measured in the issue at 50,000 x 128, that is a peak of **1,700 MB** against **959 MB** for the first build of the same corpus, a factor of **1.73x**. At DEEP-10M in a 24 GB heap the first build completes and a second one on top of it dies.

Raising `-Xmx` did not create the headroom: both auto-sized caches budgeted themselves off *total* heap, so a bigger heap grew them proportionally. And nothing consulted memory before starting - `REBUILD_SEMAPHORE` bounds how many rebuilds run at once, not how much heap one needs - so a rebuild that could not fit was simply attempted, and the resulting `OutOfMemoryError` escaped the handler that would have refused the pages.

## Changes

**1. The close path releases the resident graph and search cache before building the replacement.**
`flush()` has already CAS-ed the index to `UNAVAILABLE` and `close()` has already run `releaseBackgroundResources()` (which clears the pooled searchers), so nothing can still be reading either reference. Holding them for the whole build was pure waste; dropping them first roughly halves peak heap for exactly the case that measurably dies today.

An **online** rebuild keeps both, unchanged. `rebuildGraphBeforeSearch()`, an async rebuild, an explicit `REBUILD INDEX` and `COMPACT INDEX` do not gate concurrent searches on `status`, so for them the old graph is still load-bearing. The two behaviours are pinned against each other by test, not just by comment.

**2. Both caches are budgeted against the heap actually available.**
Availability is read from `MemoryPoolMXBean.getCollectionUsage()` - the pool's occupancy measured right after the JVM last collected it, which approximates live retained data.

Deliberately **not** `Runtime.freeMemory()`, which the issue's sketch suggested: that reports free space in the *currently committed* heap and counts everything allocated since the last collection as used, live or garbage. Read just before a GC it says the heap is nearly full when almost all of it is about to be reclaimed, so a budget taken from it would collapse the build cache for no reason and make the build far slower - the opposite of the problem being solved. When no pool publishes a collection usage (a JVM that has not collected yet, or a GC that does not publish the figure), the budget falls back to the whole ceiling, which reproduces the previous behaviour exactly rather than assuming either extreme.

**3. Admission control on the online rebuild path.**
Before starting, the estimated peak footprint - the graph being built, the build cache, and for an online rebuild the graph being replaced - is compared against `arcadedb.vectorIndex.rebuildMaxHeapPercent` (new, default 90) of the available heap. A rebuild that does not fit is **deferred rather than attempted**: the next mutation-threshold or inactivity trigger retries it, pending vectors stay exactly searchable through the in-memory delta buffer meanwhile, and the deferral is both logged (with what to change) and counted as `rebuildsDeferredForMemory` in the index stats, so an index that never fits does not go stale silently.

The estimate is coarse and the default generous on purpose: it refuses only a rebuild that is confidently too large, not one that merely looks tight. `0` restores the previous attempt-and-hope behaviour.

Only the online path is gated. A first build, a rebuild on close, `REBUILD INDEX` and `COMPACT INDEX` have no later trigger to retry them, and the close case has already released the old graph - declining one of those would turn "slower" into "never".

Note that (2) and (3) compose rather than overlap: free-heap budgeting **is** the degradation the issue asks for under "fall back to a bounded-cache build", since a resident old graph now shrinks the cache the rebuild asks for. (3) declines only when even a degraded build cannot fit, i.e. when the graphs themselves do not fit, which no cache bound can fix.

**4. An `OutOfMemoryError` during the graph persist no longer escapes the handler that refuses the pages.**
It is an `Error`, not an `Exception`, so it walked straight past the `catch (Exception)` around the persist block - skipping both the transaction rollback and the `markUnusable()` call that is the only thing stopping the next database open from trusting a possibly half-rewritten graph on node-count agreement alone (#6106 exists to prevent exactly that). That guard, the async rebuild thread's and the inactivity timer task's now all catch `Throwable`. In the latter two an escaping `Error` additionally killed the thread past the point where it clears the in-progress flag, silently stopping every later rebuild of that index.

## Alternative considered, and why this one

The issue offered to send (1) alone as a PR and explicitly declined to guess at (2) and (3). (1) on its own is the largest single win per line changed, but it only helps the close path - it does nothing for the online rebuild, which is the path #6496 shows a *single insert* into a settled index is enough to schedule, and which is the 1.73x shape in the first place. (4) alone is likewise necessary but not sufficient: it converts a silent corruption risk into a clean failure without making the failure less likely. Shipping all four is what turns "dies, and lies about it afterwards" into "degrades, defers, and says so".

## Tests

- `Issue6503VectorRebuildResidentGraphReleaseTest` - the release happens on the close path, does **not** happen on the online path, and a close with pending mutations still reopens to a correct graph over every vector.
- `Issue6503GraphPersistOutOfMemoryTest` - an `OutOfMemoryError` injected at the persist point still marks the manifest unusable, still rolls back, and leaves the index writable.
- `Issue6503VectorHeapBudgetTest` - the heap arithmetic and the footprint estimate, including that keeping the old graph resident costs exactly one graph generation more.
- `Issue6503RebuildAdmissionControlTest` - defer, admit, disable by configuration, never decline an explicit rebuild, and release the permit and in-progress flag on the deferral path.

Each was verified to fail with its corresponding fix reverted. Full `com.arcadedb.index.vector` package: 250 tests in the default lane, 49 in the `vector` lane, 62 in the `slow` lane, all green apart from `LSMVectorIndexAutoCompactionTest.anAttemptWhoseTaskIsNeverEnqueuedReleasesItsSlot`, which fails identically on a pristine `origin/main` worktree and is unrelated (async-executor compaction slot).

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3644,3 +3644,53 @@ the stored value parses as a date, before and after `APPLY DEFAULTS`, with a sen
 cannot produce. Deliberately not an equality against today's date: the two transactions can straddle midnight.
 
 [#6414](https://github.com/ArcadeData/arcadedb/issues/6414)
+
+## A vector graph rebuild no longer holds two graphs at once on close, and an OOM in one can no longer certify a half-written graph (#6503)
+
+A from-scratch rebuild of an `LSM_VECTOR` graph deliberately keeps the old graph resident so searches keep working,
+and replaces it only at the very end. On top of that it pays for a whole new build's working set: the build cache,
+the JVector builder, the ordinal map. Measured at 50,000 x 128, that is a peak of 1,700 MB for a rebuild against
+959 MB for the first build of the same corpus - a factor of 1.73x. At DEEP-10M in a 24 GB heap the first build
+completes and a second one on top of it dies.
+
+Raising `-Xmx` did not create the headroom, because both of this index's auto-sized caches budgeted themselves off
+*total* heap, so a bigger heap grew them proportionally and the rebuild was no more likely to fit. And nothing
+consulted memory before starting: the only gate was `REBUILD_SEMAPHORE`, which bounds how many rebuilds run at
+once, not how much heap one needs. A rebuild that could not fit was simply attempted.
+
+Four changes, in the order they matter:
+
+**The close path releases the old graph before building the replacement.** By the time `flush()` rebuilds, it has
+already CAS-ed the index to `UNAVAILABLE` and `close()` has already cleared the pooled searchers, so nothing can
+still be reading the old graph or the shared search cache - holding them for the duration of the build was pure
+waste. This roughly halves peak heap for exactly the case that measurably dies today. An *online* rebuild keeps
+both, unchanged: `rebuildGraphBeforeSearch()`, an async rebuild, an explicit `REBUILD INDEX` and `COMPACT INDEX`
+do not gate concurrent searches, so for them the old graph is still load-bearing.
+
+**Both caches are now budgeted against the heap actually available**, not against the ceiling. Availability is read
+from `MemoryPoolMXBean.getCollectionUsage()` - the pool's occupancy measured right after the JVM last collected it,
+which approximates live retained data - rather than from `Runtime.freeMemory()`, which counts everything allocated
+since the last GC as used whether it is live or garbage and would collapse the build cache for no reason. When no
+pool publishes a collection usage, the budget falls back to the whole ceiling, which reproduces the previous
+behaviour exactly.
+
+**An online rebuild is now admitted only if it is expected to fit.** Before starting, the estimated peak footprint
+(the graph being built, the build cache, and - for an online rebuild - the graph being replaced) is compared
+against `arcadedb.vectorIndex.rebuildMaxHeapPercent` (default 90) of the available heap. A rebuild that does not
+fit is deferred rather than attempted: the next mutation-threshold or inactivity trigger retries it, pending
+vectors stay exactly searchable through the in-memory delta buffer meanwhile, and the deferral is both logged and
+counted (`rebuildsDeferredForMemory` in the index stats) so an index that never fits does not go stale silently.
+The estimate is deliberately coarse and the default deliberately generous: it refuses only a rebuild that is
+confidently too large. Set the percentage to 0 to restore the previous attempt-and-hope behaviour. Only the online
+path is gated - a first build, a rebuild on close, `REBUILD INDEX` and `COMPACT INDEX` have no later trigger to
+retry them, so declining one would turn "slower" into "never".
+
+**An `OutOfMemoryError` during the graph persist no longer escapes the handler that refuses the pages.** It is an
+`Error`, not an `Exception`, so it used to walk straight past the `catch (Exception)` around the persist block -
+skipping both the transaction rollback and the `markUnusable()` call that is the only thing stopping the next
+database open from trusting a possibly half-rewritten graph on node-count agreement alone (#6106 is the check that
+exists to prevent exactly that). The persist guard, the async rebuild thread's guard and the inactivity timer
+task's guard now all catch `Throwable`; in the latter two an escaping `Error` would additionally have killed the
+thread past the point where it clears the in-progress flag, silently stopping every later rebuild of that index.
+
+[#6503](https://github.com/ArcadeData/arcadedb/issues/6503)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3661,11 +3661,20 @@ once, not how much heap one needs. A rebuild that could not fit was simply attem
 Four changes, in the order they matter:
 
 **The close path releases the old graph before building the replacement.** By the time `flush()` rebuilds, it has
-already CAS-ed the index to `UNAVAILABLE` and `close()` has already cleared the pooled searchers, so nothing can
+already CAS-ed the index to `UNAVAILABLE` and a closing database hands it no further requests, so nothing can
 still be reading the old graph or the shared search cache - holding them for the duration of the build was pure
-waste. This roughly halves peak heap for exactly the case that measurably dies today. An *online* rebuild keeps
-both, unchanged: `rebuildGraphBeforeSearch()`, an async rebuild, an explicit `REBUILD INDEX` and `COMPACT INDEX`
-do not gate concurrent searches, so for them the old graph is still load-bearing.
+waste. This roughly halves peak heap for exactly the case that measurably dies today.
+
+Three references have to go, not one. `graphIndex` and the search cache are the obvious two; the third is the
+searcher pool, because a pooled searcher holds the graph it was pooled under and so pins it however many times
+`graphIndex` is nulled. That is not hypothetical on this path: `LocalDatabase.closeDurableParts()` runs
+`index.flush()` *before* `index.releaseBackgroundResources()`, so on a real database close the pool is still
+populated when the rebuild starts. The release therefore clears the pool itself rather than assuming the caller
+already has.
+
+An *online* rebuild keeps all three, unchanged: `rebuildGraphBeforeSearch()`, an async rebuild, an explicit
+`REBUILD INDEX` and `COMPACT INDEX` do not gate concurrent searches on the index status, so for them the old graph
+is still load-bearing.
 
 **Both caches are now budgeted against the heap actually available**, not against the ceiling. Availability is read
 from `MemoryPoolMXBean.getCollectionUsage()` - the pool's occupancy measured right after the JVM last collected it,
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3686,9 +3686,17 @@ behaviour exactly.
 **An online rebuild is now admitted only if it is expected to fit.** Before starting, the estimated peak footprint
 (the graph being built, the build cache, and - for an online rebuild - the graph being replaced) is compared
 against `arcadedb.vectorIndex.rebuildMaxHeapPercent` (default 90) of the available heap. A rebuild that does not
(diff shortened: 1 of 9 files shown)


lvca:
Thanks - point 1 was not just a documentation gap, it found a real defect in the change. Pushed 28c23b7.

**1. The unenforced invariant (valid, and it hid a bug).**

Chasing "is there really no path where a reader sneaks in" led somewhere else first: `LocalDatabase.closeDurableParts()` calls `index.flush()` **before** `index.releaseBackgroundResources()`, not after. My justification ("`close()` has already run `releaseBackgroundResources()`") was only true of `LSMVectorIndex.close()`, which is *not* how a database close reaches `flush()`.

That matters because `GraphSearcherPool.Pooled` holds a strong `ImmutableGraphIndex` reference. So on the dominant close path the pool was still populated when the rebuild started, every pooled searcher pinned the old graph, and **the release freed nothing on exactly the path it was written for**. Nulling `graphIndex` is necessary but not sufficient.

The release now clears the pool itself rather than assuming the caller did, which also makes it correct independently of caller ordering. The test asserts `pooledGraphSearchers == 0` at the first progress sample; confirmed to fail with the `clear()` removed, so it is discriminating rather than decorative.

On the reader question itself: you are right that it is a caller-side contract (the search methods gate on `lock.readLock()` and never consult `status`), and right about the consequence - a violating reader takes the `graphIndex == null` branch and is served from the delta scan, so an incomplete result set rather than a crash or torn state, for the duration of the rebuild instead of the instant around the swap. That is now stated explicitly in the method javadoc and the test javadoc, as the reason it is acceptable where the contract holds and must not be used where it does not.

**2. The undocumented 90% clamp (valid).** Documented on all three heap-percent settings, including the two pre-existing ones that had the same silent cap.

**3. The cache-sizing ordering (valid, taken as a comment rather than a test).** The ordering is correct today - the release block is at the top of `buildGraphFromScratchExclusively`, `computeGraphBuildCacheCapacity` is called well after it - and there is now a comment at the sizing call site saying why it must stay that way: budgeting off available heap means moving it above the release would make a close-path rebuild shrink its own build cache to make room for memory it is in the act of freeing, which is the slow-build failure mode this change exists to avoid rather than cause.

I deliberately did not write a test for it. Observing that interaction requires a collection to land between the release and the sizing call, since availability is read from `getCollectionUsage()`, so any assertion on it would be GC-timing dependent - and a flaky test here would be worse than a precise comment.

**On the generational-collector staleness note:** agreed, and agreed the failure direction is the safe one. Worth adding that the deferral is not silent when it happens - it logs what to change and increments `rebuildsDeferredForMemory` in the index stats - so the "it's deferring even though the heap looks fine in a profiler" report has a counter to point at rather than only a hunch.

Re-ran the full `com.arcadedb.index.vector` package: 250 default lane, 49 `vector` lane, 62 `slow` lane, no new failures. `LSMVectorIndexAutoCompactionTest.anAttemptWhoseTaskIsNeverEnqueuedReleasesItsSlot` fails identically on a pristine `origin/main` worktree and is unrelated to this PR.

lvca:
Good catch on the backoff - that was a real gap, and it made the documented behaviour untrue rather than just suboptimal. Pushed e7e0bef.

**Deferred rebuilds retriggering on every query (fixed).**

Confirmed exactly as described: `admitOnlineRebuild()` returning `false` does not consume the mutations that triggered the cycle, `asyncRebuildInProgress` clears in the thread's `finally` a moment later, and `rebuildGraphBeforeSearch()` evaluates the trigger on every query. So the WARNING text and the config doc both promised "the next mutation or inactivity trigger retries" while the code actually retried on the next *search*.

A deferral now starts a cooldown - `arcadedb.vectorIndex.rebuildDeferralCooldownMs`, default 30s, 0 to disable - checked in `startAsyncGraphRebuild()` **before the thread is spawned**, which was your point about paying for the attempt: a suppressed trigger now costs nothing at all, so no daemon thread, no `REBUILD_SEMAPHORE` acquire/release, no `getActiveCount()` walk, and no WARNING. A completed build clears the cooldown, so a stale deferral cannot keep gating triggers once the shortage has passed.

Three tests, each confirmed to fail with the gate removed (the first goes to 26 attempts instead of 1):
- 25 consecutive triggers inside the cooldown attempt exactly once
- with the cooldown at 0, each of 3 triggers attempts again - which is what proves the first test measures the cooldown and not some other state that stopped triggering
- a successful build clears the cooldown

**`liveHeapBytes()` fallback on collectors that don't publish collection usage (addressed).**

You're right that the fallback itself is fine - it reproduces the pre-#6503 behaviour exactly, so it can't degrade anything - but it was *invisible*, which is the actual problem: an operator on such a collector had no way to know the improvement was inactive. It's now reported once per JVM at FINE. I deliberately did not name specific collectors in the comment, since which ones publish it varies by version and configuration and a hardcoded list would age badly.

I also documented the staleness case you raised alongside it, with the reasoning that makes both acceptable: an infrequently-collected old gen and an absent reading both cause headroom to be **under**-reported, so the failure direction is a smaller cache or a deferred rebuild - never a rebuild admitted that should not have been.

**`catch (Throwable)` swallowing OOM in the persist block (acknowledged, left as is).**

Agreed that catching `Error` broadly is normally the wrong instinct, and worth flagging. Leaving it because not rethrowing is what that block already did for every other failure - widening the catch changed which throwables reach the handler, not what the handler does - the in-memory graph is unaffected since only the persist failed, and the manifest is marked unusable either way, which is the property that actually protects the next open. Rethrowing would fail the close and lose the graph the build had just completed successfully in memory, which is worse on a path whose whole purpose is to persist it.

Full `com.arcadedb.index.vector` re-run: 253 default lane, 49 `vector` lane, 62 `slow` lane, no new failures. `LSMVectorIndexAutoCompactionTest.anAttemptWhoseTaskIsNeverEnqueuedReleasesItsSlot` continues to fail identically on a pristine `origin/main` worktree and is unrelated to this PR.

Issue #4730, opened 2026-06-24
UPDATE SET map = map.remove(key) does not persist map key removal


TobiasJoseHermann:
## Summary

When a `MAP` property is updated with:

```sql
UPDATE IssueRemoveMapRepro SET acl = acl.remove('u1') WHERE id = 'set_remove'
```

ArcadeDB reports `count = 1`, but the key remains persisted in the record. The likely cause is that `remove()` mutates the existing `Map` instance in place, and `UPDATE SET` then compares the already-mutated current value with the returned value. Because they are the same object/equivalent value, `setProperty()` is skipped, the document is not marked dirty, and `SaveElementStep` skips saving it.

Using `map.removeKey(acl, 'u1')` works because it returns a new map instance.

## Environment

- ArcadeDB console: `v26.7.1-SNAPSHOT`
- Source git revision: `97bade29d`

## Minimal Reproduction

```sql
CREATE DOCUMENT TYPE IssueRemoveMapRepro IF NOT EXISTS;
CREATE PROPERTY IssueRemoveMapRepro.id IF NOT EXISTS STRING;
CREATE PROPERTY IssueRemoveMapRepro.acl IF NOT EXISTS MAP;
DELETE FROM IssueRemoveMapRepro;

INSERT INTO IssueRemoveMapRepro
SET id = 'set_remove', acl = {'u1':'rw','u2':'r'};

UPDATE IssueRemoveMapRepro
SET acl = acl.remove('u1')
WHERE id = 'set_remove';

SELECT count(*) AS has_u1
FROM IssueRemoveMapRepro
WHERE id = 'set_remove'
  AND acl['u1'] IS NOT NULL;
```

Console result:

```text
UPDATE IssueRemoveMapRepro SET acl = acl.remove('u1') WHERE id = 'set_remove'
+-----+-----+
|NAME |VALUE|
+-----+-----+
|count|1    |
+-----+-----+

SELECT count(*) AS has_u1 FROM IssueRemoveMapRepro WHERE id = 'set_remove' AND acl['u1'] IS NOT NULL
+------+-----+
|NAME  |VALUE|
+------+-----+
|has_u1|1    |
+------+-----+
```

## Expected Result

`has_u1` should be `0`, because `acl.remove('u1')` should remove the `u1` key and the `UPDATE` reports that one record was updated.

## Actual Result

`has_u1` is `1`. The key is still present after the update.

### `map.removeKey(...)`

```sql
[...]

PR #6515, opened 2026-08-20
fix(#6456): UPDATE ... REMOVE ... RETURN BEFORE returns post-removal snapshot

### LINKED ISSUE #6456: UPDATE ... REMOVE <collection> ... RETURN BEFORE returns the post-removal snapshot (shallow-copied before-image mutated in place)
### Summary
`UPDATE ... REMOVE <collection> ... RETURN BEFORE` returns the **post**-removal snapshot as the BEFORE image, because the before-image is a shallow copy that shares the live embedded collection object, which the REMOVE then mutates in place. The persisted record is correct; only the returned BEFORE result is wrong.

### Where (HEAD `0291182af`)
`engine/src/main/java/com/arcadedb/query/sql/executor/CopyRecordContentBeforeUpdateStep.java:64-65`:
```java
for (final String propName : result.getPropertyNames())
  prevValue.setProperty(propName, result.getProperty(propName));  // stores the LIVE List/Map reference
```
`UpdatableResult.getProperty` → `MutableDocument.get` returns the live embedded collection, not a copy. Then `parser/UpdateRemoveItem.applyUpdate` fetches the same list and mutates it in place (`MultiValue.remove(leftVal, rightVal, false)` → `coll.remove(...)`). The plan chains `handleReturnBefore` before `handleOperations`, and `UnwrapPreviousValueStep` returns the shared object, so the BEFORE image reflects the later in-place removal.

### Repro
```sql
CREATE VERTEX V SET tags = ['a','b','c'];   -- #10:0
UPDATE #10:0 REMOVE tags = 'b' RETURN BEFORE;
```
Expected BEFORE `{tags:['a','b','c']}`; actual `{tags:['a','c']}`. Persisted record is correctly `['a','c']`. Same for nested `REMOVE map['k'] RETURN BEFORE`. Scalar `SET` and top-level replacement are unaffected (they replace the reference, not mutate in place).

### Suggested fix
In `CopyRecordContentBeforeUpdateStep`, deep-copy multi-value/embedded values into `prevValue` (copy `Collection`/`Map` values into fresh containers, or snapshot via `record.toMap()`), so a later in-place mutation cannot leak into the BEFORE image.

### Scope / confidence
High — traced end-to-end at HEAD. Affects `UPDATE ... RETURN BEFORE` whose operations mutate an embedded collection/map in place (`REMOVE listProp = value`, `REMOVE map[k]`, `REMOVE list[i]`).

robfrank:
Closes #6456

## What does this PR do?

Fixes `UPDATE ... REMOVE <collection/map value> ... RETURN BEFORE`, which returned the **post**-removal snapshot as
the BEFORE image instead of the pre-removal one.

`CopyRecordContentBeforeUpdateStep` built the BEFORE snapshot by copying each property value with a plain
`prevValue.setProperty(propName, result.getProperty(propName))`. For `List`/`Set`/`Map` properties this stores the
same live reference the record holds, not a copy. `UpdateRemoveItem.applyUpdate` then mutates that structure in
place (`MultiValue.remove()` → `Collection.remove(...)`, or a nested `applyRemove` on a map/list element) and marks
the owner dirty so the change persists correctly. Because the BEFORE snapshot shared the same object, the later
in-place removal was visible through it too, so the RETURN BEFORE result showed the record **after** the removal
instead of before it. The persisted record was always correct; only the returned BEFORE projection was wrong.

The fix deep-copies `List`/`Set`/`Map` values (recursively, so nested collections/maps are covered too) when
building the snapshot in `CopyRecordContentBeforeUpdateStep`, so a later in-place mutation of the live property
can no longer leak into the already-returned BEFORE image. Scalars, RIDs, and arrays are left untouched:
`MultiValue.remove()` already replaces an array with a fresh copy rather than mutating it, so arrays never shared
state with the live value in the first place.

## Motivation

Reported in #6456 with a traced root cause and repro. `UPDATE ... REMOVE tags = 'b' RETURN BEFORE` on
`tags = ['a','b','c']` returned BEFORE `{tags:['a','c']}` instead of the expected `{tags:['a','b','c']}`.

## Related issues

Closes #6456

## Additional Notes

- Scope: this only affects `UPDATE ... RETURN BEFORE` whose REMOVE operation mutates an embedded
  `List`/`Set`/`Map` in place (`REMOVE listProp = value`, `REMOVE map[k]`, `REMOVE list[i]`). Scalar `SET` and
  top-level replacement were already unaffected, since those replace the property reference rather than mutating it.
- No behavior change to the persisted record in any case; only the BEFORE projection returned by the statement.

## Test plan

- [x] New regression test `UpdateStatementExecutionTest#returnBeforeWithListRemoveIsNotMutatedByTheRemoval`:
      `UPDATE ... REMOVE tagsList = 'bar' RETURN BEFORE` returns the original 3-element list as BEFORE, while the
      persisted record correctly ends up with 2 elements.
- [x] New regression test `UpdateStatementExecutionTest#returnBeforeWithMapKeyRemoveIsNotMutatedByTheRemoval`:
      `UPDATE ... REMOVE tagsMap["bar"] RETURN BEFORE` returns the original 3-entry map as BEFORE, while the
      persisted record correctly ends up with 2 entries.
- [x] Both new tests fail against pre-fix code (confirmed) and pass after the fix.
- [x] `UpdateStatementExecutionTest` (40 tests) and `UpdateStatementTest` (12 tests) pass.
- [x] `MultiValueTest`, `SQLExecutorLowCoverageTest`, `SQLExecutorAdditionalCoverageTest`,
      `SQLMethodRemoveTest`, `SQLMethodRemoveAllTest` pass (114 tests total), confirming the existing
      copy-on-write remove/removeAll semantics (#4730) are unaffected.

## Checklist

### DIFF
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/UpdateStatementExecutionTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/UpdateStatementExecutionTest.java
@@ -774,6 +774,40 @@ public class UpdateStatementExecutionTest extends TestHelper {
   }
 
   @Test
+  void returnBeforeWithNestedListInsideMapRemoveIsNotMutatedByTheRemoval() {
+    // Issue #6456 review follow-up: pin down the recursive branch of the deep copy, not just a top-level
+    // List/Map, by mutating a List nested inside a Map value.
+    final String className = "overridden" + this.className;
+    database.getSchema().createDocumentType(className);
+
+    final MutableDocument doc = database.newDocument(className);
+    final Map<String, Object> nestedMap = new HashMap<>();
+    final List<String> nestedList = new ArrayList<>();
+    nestedList.add("foo");
+    nestedList.add("bar");
+    nestedList.add("baz");
+    nestedMap.put("list", nestedList);
+    doc.set("nestedMap", nestedMap);
+    doc.save();
+
+    final ResultSet result = database.command("sql",
+        "update " + className + " remove nestedMap[\"list\"] = 'bar' RETURN BEFORE");
+    assertThat(result.hasNext()).isTrue();
+    final Result item = result.next();
+    assertThat(item).isNotNull();
+    final Map<String, Object> before = item.getProperty("nestedMap");
+    assertThat((List<String>) before.get("list")).containsExactly("foo", "bar", "baz");
+    assertThat(result.hasNext()).isFalse();
+    result.close();
+
+    final ResultSet persisted = database.query("sql", "SELECT nestedMap FROM " + className);
+    assertThat(persisted.hasNext()).isTrue();
+    final Map<String, Object> after = persisted.next().getProperty("nestedMap");
+    assertThat((List<String>) after.get("list")).containsExactly("foo", "baz");
+    persisted.close();
+  }
+
+  @Test
   void returnBeforeWithMapKeyRemoveIsNotMutatedByTheRemoval() {
     // Issue #6456: same shallow-copy bug for nested `REMOVE map[key]` on the map value.
     final ResultSet result = database.command("sql",
--- a/engine/src/test/java/com/arcadedb/query/sql/executor/UpdateStatementExecutionTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/executor/UpdateStatementExecutionTest.java
@@ -753,6 +753,47 @@ public class UpdateStatementExecutionTest extends TestHelper {
   }
 
   @Test
+  void returnBeforeWithListRemoveIsNotMutatedByTheRemoval() {
+    // Issue #6456: UPDATE ... REMOVE <collection value> ... RETURN BEFORE must snapshot the collection
+    // BEFORE the in-place removal, not share the live list reference with it.
+    final ResultSet result = database.command("sql",
+        "update " + className + " remove tagsList = 'bar' RETURN BEFORE where name = 'name1'");
+    assertThat(result.hasNext()).isTrue();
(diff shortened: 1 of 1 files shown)


robfrank:
Addressed the review follow-ups:

- **Nested-collection recursion**: added `returnBeforeWithNestedListInsideMapRemoveIsNotMutatedByTheRemoval` (List nested inside a Map), confirmed it fails when the recursive call is replaced with a shallow copy and passes with the real fix.
- **Javadoc scope note**: documented that embedded `Document`/`EmbeddedDocument` values are intentionally not deep-copied here (out of scope for #6456).
- **Set branch test**: not added. A persisted document property can only come back as a `List` on read — `BinarySerializer` has no SET wire type, only `TYPE_LIST` — so there is no way for an `UPDATE ... RETURN BEFORE` against a stored record to produce a live `java.util.Set` at this call site to exercise. The `Set` branch mirrors the `List` branch line for line; leaving it covered by inspection rather than adding a test that can't actually reach it through the documented API.
- **Embedded-document variant of the same bug**: verified the claim (confirmed with a throwaway repro: `REMOVE embeddedDoc.field RETURN BEFORE` does leak the post-removal state into BEFORE, via `SuffixIdentifier.applyRemove` calling `document.modify().remove(...)`, which mutates the same live embedded document the snapshot holds a reference to). Filed as a separate follow-up since it's outside the scope of #6456 (which was specifically about `List`/`Set`/`Map` values): https://github.com/ArcadeData/arcadedb/issues/6517

PR #6524, opened 2026-08-20
fix(#6495): compacted vector graph must reuse under its new name, not just avoid leaking

lvca:
## What

Builds on #6504 (thank you @tae898 for the diagnosis and the first fix - both commits from that PR are included here unmodified) to close the remaining gap it deliberately scoped out.

#6504 fixed `getOrCreateGraphFile()` deriving the graph file's path from the already-suffixed `mutable.getFilePath()` instead of the bare index path, which made every lazily-created graph file undiscoverable and caused unbounded growth on every reopen after a `COMPACT INDEX`. That fix is necessary and correct on its own.

It also left one gap on purpose, noted in its own description: `rewriteDataFileWithLiveEntries()` renames the index during compaction, but the in-memory `graphFile` reference still points at the file registered under the pre-compaction name. `getOrCreateGraphFile()` only derives a fresh path when `graphFile` is `null`, so the graph rebuilt by the same compaction gets persisted right back into that same now-unreachable file - correct content, unreachable name. Net effect even with #6504 applied: one orphaned `.vecgraph` file per compaction, and one avoidable full rebuild on the very next open.

## Fix

Right before the post-compaction rebuild persists its graph, drop the stale `graphFile` reference (only when compaction actually renamed the index - `locationIndexAlreadyPublished`) so `getOrCreateGraphFile()` registers a fresh file under the CURRENT name. This mirrors exactly how compaction already replaces the mutable/compacted components: a new file registered under the new identity, the old one dropped outside any lock once its replacement exists (`dropReplacedComponent`). If persistence itself fails, the stale reference is restored instead of losing the graph outright.

No change to the sensitive critical section that keeps the rename and the schema re-keying adjacent (the `indexName = ...` / `indexRenamed(...)` pair) - this only touches the persist step later in the same method, gated on the boolean that section already produces.

## Test

Extended `CompactIndexKeepsGraphReusableTest` (from #6504) to assert the fully closed loop instead of just the bounded-damage version:
- exactly one `.vecgraph` file survives immediately after `COMPACT INDEX` (no orphan)
- `graphRebuildCount == 0` on every reopen, including the first one (previously only opens 2+ were checked)

Verified by reverting each fix independently: reverting #6504's path fix reproduces its original failure; reverting this PR's addition on top of #6504's fix reproduces `graphRebuildCount == 1` on the first reopen and 2 `.vecgraph` files on disk instead of 1.

Full `com.arcadedb.index.vector` suite: 342 tests, 0 failures.

## Test plan

- [x] `mvn -o -pl engine -am compile`
- [x] `CompactIndexKeepsGraphReusableTest` passes
- [x] Full `com.arcadedb.index.vector` package: 342/342 passing

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -2593,9 +2593,12 @@ public class LSMVectorIndex implements Index, IndexInternal {
       // rebuild issue #6495's fix otherwise still leaves behind. Dropping the reference here, immediately
       // before it is read, makes the persist step below register a fresh file under the CURRENT name instead,
       // so the very next open finds it - zero rebuilds, zero orphans - exactly like the mutable/compacted
-      // components a compaction replaces just above: the stale file is torn down only once its replacement
-      // (gf) actually exists, and outside any lock, because a reader that already captured it must be able to
-      // finish (see dropReplacedComponent()).
+      // components a compaction replaces just above. Unlike those two, the stale file is NOT torn down here:
+      // it stays the only usable persisted graph until the replacement is CERTIFIED (graphCertified, set once
+      // writeGraphManifest() has vouched for committed pages), not merely once getOrCreateGraphFile() hands
+      // back an empty file object. A write or commit failure in between restores this reference instead (see
+      // the catch block below and the graphCertified check further down), so a failed persist still leaves
+      // something usable on disk rather than nothing.
       final LSMVectorIndexGraphFile staleGraphFileFromCompaction = locationIndexAlreadyPublished ? graphFile : null;
       if (staleGraphFileFromCompaction != null)
         graphFile = null;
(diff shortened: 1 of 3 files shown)


lvca:
Good catch, thank you - confirmed and fixed in 935f6ed92.

Verified the failure mode directly: reverted just the ordering (drop-on-replacement-existing instead of drop-on-certified) and added a test injecting a failure at the same point `Issue6503GraphPersistOutOfMemoryTest` does (right after `writeGraph()`, before commit), but through the `compactDataFile=true` path. Against the old ordering it fails exactly as predicted - `hasPersistedGraph()` is `false`, i.e. the graph is actually gone, not just stale-named.

Changes:
- The stale pre-compaction file is now dropped only after `graphCertified = true` (i.e. after `writeGraphManifest()` has vouched for committed pages), not as soon as `getOrCreateGraphFile()` hands back an empty object.
- On a failed/uncertified persist, the stale reference is restored as the active `graphFile` **and** the now-abandoned, uncertified replacement is dropped in turn - so a failed compaction-triggered rebuild leaves the index exactly where it was before the attempt (one usable, staler-named graph), rather than either losing the graph outright or leaking the broken replacement alongside the restored one.
- Added `CompactIndexGraphPersistFailureTest` covering exactly this.

Also noted the `graphFile` field visibility point - agreed it's a pre-existing pattern (the field was already read/written unlocked outside `getOrCreateGraphFile()`'s original call site) rather than something this PR introduces, and the two additional unguarded writes here follow the same lifetime/ownership rules (single-threaded within `graphBuildLock`, only ever read cross-thread under `lock.writeLock()` in the paths that matter for correctness). Left as-is rather than widening the change's scope, but flagging it's a fair thing to revisit separately if the maintainer wants a broader look at that field's synchronization.

Full `com.arcadedb.index.vector` suite: 378 tests, 0 failures.

PR #6528, opened 2026-08-21
fix(#6527): make LSMVectorIndex.graphFile volatile for cross-thread readers

### LINKED ISSUE #6527: LSMVectorIndex.graphFile field is read/written outside any lock in several places
## Context

Surfaced during review of #6495 / #6504 / #6524 (COMPACT INDEX orphaning the persisted vector graph file).

`LSMVectorIndex.graphFile` is a plain (non-`volatile`) field. It is written under `lock.writeLock()` in some places (e.g. the constructors, `ensureGraphAvailable()`), but read and written elsewhere with no lock at all - only implicitly serialized by `graphBuildLock`, which builders hold but which readers of the field (e.g. `getFileIds()`, the `hasPersistedGraph()` checks scattered through the search/build paths) do not.

This is not newly introduced by #6524: `getOrCreateGraphFile()`'s original unlocked call site already had this shape before that PR. #6524 adds two more unguarded writes to the field (as part of restoring a stale reference on a failed post-compaction persist), which is why it came up in review there - see the discussion on https://github.com/ArcadeData/arcadedb/pull/6524#issuecomment-5363007067 and the reply at https://github.com/ArcadeData/arcadedb/pull/6524#issuecomment-5363668371.

## Why this might matter

`getFileIds()` in particular is documented (issue #4937) as needing to include the graph file's id in the commit-lock set "because the companion graph file receives page writes during the transaction too." If `getFileIds()` observes a stale or torn view of `graphFile` due to the lack of synchronization, the commit-lock coverage it computes could theoretically miss the graph file's lock, which is exactly the class of bug #4937 was about.

No concrete failure has been reproduced - this is a code-shape observation from review, not a confirmed bug. The risk window is small (in the compaction path, the null-then-reassign window added by #6524 is a couple of statements with no I/O in between) and functionally, at worst, a torn read there means a component-not-yet-associated-with-a-lock for one instant, not data corruption in what's already been committed.

## Suggested scope for a follow-up

- Audit every read and write of `LSMVectorIndex.graphFile` and classify each as: (a) already covered by `lock.writeLock()`/`lock.readLock()`, (b) covered only by `graphBuildLock` (fine if all writers hold it, but readers like `getFileIds()` don't), (c) genuinely unguarded.
- Decide whether `graphFile` should become `volatile` (cheapest fix, gives at least a consistent single-value read/write without introducing a new lock), or whether the handful of readers that matter for correctness (starting with `getFileIds()`) should take `lock.readLock()`.
- Add a regression test if a concrete interleaving can be constructed that demonstrates `getFileIds()` returning a commit-lock set that omits the current graph file's id while a transaction is writing graph pages.

@author reference for the code in question: `engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java`

lvca:
## What

Closes #6527.

`LSMVectorIndex.graphFile` is written only under `graphBuildLock` (builders are serialized against each other by that mutex) or during construction (before the instance escapes), but read by several call sites - `getFileIds()` in particular - with no lock at all: not `graphBuildLock`, not `lock.readLock()`, nothing.

A plain (non-`volatile`) field gives the JMM no happens-before edge between a builder's write and a reader on another thread, so a reader is not guaranteed to ever observe a reassignment. This is exactly the class of bug issue #4937 fixed once already: `getFileIds()` is documented there as needing the graph file's id in the commit-lock set "because the companion graph file receives page writes during the transaction too." If it observes a stale reference, the commit-lock coverage it computes could miss the graph file's lock.

I audited every read/write site of the field (constructors, `discoverAndLoadGraphFile()`, `getOrCreateGraphFile()`, the post-compaction persist block, `getFileIds()`, the `hasPersistedGraph()` checks, `getStats()`'s diagnostics snapshot). Every write after construction happens under `graphBuildLock`, so writer-writer races are already impossible - the only gap is reader visibility.

## Fix

Marked `graphFile` `volatile`. This matches the pattern already used by every other field in this class with the same "written under one lock (or a dedicated build mutex), read under no lock" shape: `graphState`, `graphIndex`, `ordinalToVectorId`, `vectorIndex` are all `volatile` for the identical reason.

The issue's writeup considers two options: `volatile`, or moving the readers that matter (starting with `getFileIds()`) onto `lock.readLock()`. I went with `volatile` rather than adding locking at the read sites: `getFileIds()` runs on every transaction's commit-lock computation, so taking `lock.readLock()` there would add contention on a hot path to fix what is purely a visibility gap - `volatile` gives every read site (not just `getFileIds()`) the happens-before edge for free, with none of that cost, and needs a one-line change instead of auditing and locking each read site individually.

No behavior changes: all writes were already mutually exclusive under `graphBuildLock`; this only makes their result visible to readers promptly and correctly.

## Test

- Added `LSMVectorIndexGraphFileVisibilityTest`:
  - A reflection test asserting the field is `volatile`, so the fix can't be silently reverted. Confirmed red before the fix (`AssertionFailedError: Expecting value to be true but was false`), green after.
  - A functional regression for the concrete scenario the issue traces its concern to: after `COMPACT INDEX` reassigns `graphFile` (per #6495/#6524), `getFileIds()` must report the new file's id, not a stale one.
- No concrete failing interleaving was reproducible in a deterministic unit test (the issue itself says so - "no concrete failure has been reproduced"), which is why the primary regression is a reflection/contract test on the fix itself rather than a timing-dependent concurrency test.

## Test plan

- [x] `mvn -o -pl engine -am compile` (and full `mvn -o compile` reactor)
- [x] `LSMVectorIndexGraphFileVisibilityTest`: red before the fix, green after
- [x] Full `com.arcadedb.index.vector` package, non-`vector`-tagged lane: 325/325 passing
- [x] Full `com.arcadedb.index.vector` package, `vector`-tagged lane: 55/55 passing

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -219,7 +219,12 @@ public class LSMVectorIndex implements Index, IndexInternal {
 
   // Graph file for persistent storage of graph topology
   // Allows lazy-loading graph from disk and avoiding expensive rebuilds
-  private       LSMVectorIndexGraphFile graphFile;
+  // Written only under graphBuildLock (builders are serialized against each other by that mutex) or during
+  // construction (before the instance escapes), but read by several call sites - getFileIds() in particular -
+  // under no lock at all. volatile is what gives those readers a happens-before edge against the writer,
+  // matching graphState/graphIndex/ordinalToVectorId/vectorIndex just above, which are read the same way
+  // (issue #6527).
+  private volatile LSMVectorIndexGraphFile graphFile;
   private final AtomicInteger           mutationsSinceSerialize;
 
   // Product Quantization (PQ) for zero-disk-I/O approximate search
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -225,7 +225,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
   // matching graphState/graphIndex/ordinalToVectorId/vectorIndex just above, which are read the same way
   // (issue #6527).
   private volatile LSMVectorIndexGraphFile graphFile;
-  private final AtomicInteger           mutationsSinceSerialize;
+  private final    AtomicInteger           mutationsSinceSerialize;
 
   // Product Quantization (PQ) for zero-disk-I/O approximate search
   // PQ file stores codebooks and encoded vectors; pqVectors is cached in memory
--- a/engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexGraphFileVisibilityTest.java
+++ b/engine/src/test/java/com/arcadedb/index/vector/LSMVectorIndexGraphFileVisibilityTest.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.index.vector;
+
+import com.arcadedb.database.Database;
+import com.arcadedb.database.DatabaseFactory;
+import com.arcadedb.index.Index;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.schema.DocumentType;
+import com.arcadedb.schema.Type;
+import com.arcadedb.utility.FileUtils;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.lang.reflect.Field;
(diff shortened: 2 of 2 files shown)


lvca:
Addressed both nits from the automated review:

- **Field alignment**: fixed in 13ffef7b7 - realigned `mutationsSinceSerialize` to `graphFile`'s new column.
- **Pre-existing TOCTOU on the `graphFile != null && graphFile.getComponentFile() != null ? ... : null` pattern**: agreed this is out of scope. It's pre-existing for every other volatile field in this class with the same null-check-then-dereference shape, not introduced or worsened by this PR, and `volatile` doesn't change its risk profile either way (a null-check followed by a dereference is racy or not independent of the field's memory visibility semantics). Tracking it here would be scope creep on a narrow, single-purpose fix; happy to file a separate follow-up if it's judged worth doing across the class.

Both `graphFileFieldMustBeVolatileForCrossThreadReadersLikeGetFileIds` and `getFileIdsMustReportTheGraphFileCompactionRegisteredNotAStaleOne` still pass, along with `CompactIndexKeepsGraphReusableTest` and `CompactIndexGraphPersistFailureTest`.

lvca:
## CI status summary

All checks are green except `ha-integration-tests`, which is red across both the original run and one rerun on 4 different, unrelated tests total (`Issue5569SlotMergeDeleteRaftIT`, `RaftTimeSeriesWriteReadYourWritesIT`, `RemoteStickyConnectionStrategyIT`, `RaftCommandReadConsistencyIT`) - none touching vector indexes, all either `DatabaseIsClosedException` resync races or established repeat members of this suite's well-documented, tracked ~75%-intermittent rotating cast (issues #5668, #5702, both open). Confirmed via `main`'s own recent runs showing the same lane failing on unrelated tests.

- `vector-unit-tests` (the lane that actually exercises this change): 55/55 green.
- All other 29 checks green: unit-tests, slow-unit-tests, opencypher-tck-tests, integration-tests, builder-tests, all 5 e2e lanes, coverage-report, CodeQL (all 5 languages), Codacy (static analysis + both coverage checks), Meterian, codecov (patch + project).
- Two `claude-review` rounds, both clean - first flagged a cosmetic field-alignment nit (fixed in 13ffef7b7), second: "no blocking concerns."

Not rerunning `ha-integration-tests` a third time - this PR's diff is a single field modifier in `LSMVectorIndex.java` with zero code-path overlap with the HA/Raft replication tests failing here, and this repo's own precedent treats 2+ reruns with disjoint, cast-matching failures as sufficient confirmation it's the suite, not the change.

Ready to merge.

Issue #6303, opened 2026-08-17
Follow-ups from #6281: path-keyed flush-thread maps, an unchecked worker pause, and async-dispatched DDL


lvca:
Collected while fixing #6281 (merged as PR #6290, `7707a2ffb`). Three independent items plus two smaller notes. Items 1 and 2 were each raised by a review pass on that PR and deliberately left out of its scope; item 3 is a capability gap the fix made visible rather than created.

---

## Item 1 - the other per-database maps of `PageManagerFlushThread` are keyed by a `Database` that compares by PATH

`LocalDatabase.equals()`/`hashCode()` compare **`databasePath`**, not instance identity (`LocalDatabase.java:2114-2130`). Every per-database map in `PageManagerFlushThread` is keyed on that:

```java
ConcurrentHashMap<Database, Integer>                            suspended             // :90
ConcurrentHashMap<Database, ConcurrentLinkedQueue<PagesToFlush>> deferredByDatabase   // :138
ConcurrentHashMap<BasicDatabase, AtomicInteger>                 slotsInUse            // :198
ConcurrentHashMap<BasicDatabase, AtomicLong>                    flushedPagesPerDatabase // :236
ConcurrentHashMap<BasicDatabase, AtomicLong>                    deferredRAMByDatabase // :257
```

A database closed or dropped and then reopened at the **same path** is therefore an *equal key but a different instance*. Anything still holding the old instance resolves to the new database's entry.

`removeAllPagesOfDatabase()` clears these entries on close, so the window is narrow - it needs an old-instance reference used *after* that purge - but such references exist. The concrete path: a `PagesToFlush` still in the queue carries the old instance in its `database` field, and the defer check at `PageManagerFlushThread:736` calls `isSuspended(db)` with it. After a reopen that consults the **new** instance's suspension flag, so a straggler batch of a database that no longer exists can be deferred into the new database's backlog and charged to its `deferredRAMByDatabase`. #6200's accounting has drift-repair counters that regression tests assert stay at zero, so this would surface there first - as a mystery.

**#6281 defended only `slotsInUse`**, in two ways worth reusing: the emptied wrapper is now removed from the queue when its database goes (so there is nothing stale to poll), and each batch carries the `AtomicInteger` its slot was charged on, so the release never re-resolves through the map. The second technique generalises to any of these counters.

[...]

Issue #5636, opened 2026-07-31
Follow-ups from #5608: the dead getBucketById null-check has siblings, monotonic stats are gauges, Studio hides zeroes, CI is permanently red


lvca:
### Summary

Four things surfaced while doing #5608 (PR #5611, merged as `bf5137ce0`) that were out of its scope. Only the first is a defect with user-visible consequences; the rest are observability and hygiene items that the same work made visible. Grouped into one issue because they were all found in one pass and none is big enough to stand alone.

---

- [ ] **1. The dead null-check that #5608 item 4 fixed is a PATTERN, and at least one other instance costs a user a good error message.**

  #5608 established that `LocalSchema.getBucketById(id)` (single-arg, LocalSchema.java:444) raises `SchemaException` in *exactly* the cases a caller would test for `null` - both an out-of-range id and a component that is not a bucket - so every `if (bucket == null)` written after it is unreachable. The two-arg `getBucketById(id, false)` (:448) is the form that returns the `null` those branches were written against. Two more sites do the same thing:

  - **`BinarySerializer.readExternalValue` (BinarySerializer.java:1183-1191) - the one that actually costs something.** The dead branch holds a deliberately written, user-actionable `SerializationException`: *"If the bucket was tiered to a secondary path, set 'arcadedb.externalPropertyBucketPath' to the same value used at creation time and reopen the database."* That message is unreachable. A user who hits precisely the scenario its own comment describes - an EXTERNAL property bucket tiered via `arcadedb.externalPropertyBucketPath`, then reopened with that config unset - gets a bare `Bucket with id 'N' was not found` and none of the guidance someone wrote for exactly this moment. This is the hardest case in the file to diagnose from the outside, which is presumably why the message exists.

  - **`TruncateBucketStatement` (TruncateBucketStatement.java:63-66) - cosmetic.** The branch is dead too, but the enclosing `catch (Exception)` rewraps into the same `CommandExecutionException`, so the only damage is a doubled message (`Bucket not found: Bucket with id '9999' was not found`). Listing it for completeness, not as a reason to act.

  While checking those: **`BinarySerializer.finalizeExternalWrite` (:1155) dereferences the bucket with no null check at all.** That is not an NPE risk - the single-arg form throws first - but it means the WRITE path of the same feature raises a bare `SchemaException` where the READ path at least tried to explain itself. Worth making the pair consistent.

[...]

PR #6529, opened 2026-08-21
fix(#6526): async executor follow-ups - in-place setParallelLevel(), DDL pool durability, boundary-commit metric

### LINKED ISSUE #6526: Async executor follow-ups from #6509: setParallelLevel() teardown, DDL pool durability gap, no boundary-commit metric
## Background

#6509/#6511 fixed `DatabaseAsyncExecutorImpl.setTransactionUseWAL()`/`setTransactionSync()` unconditionally tearing down and respawning the entire async worker pool on every call - force-exiting the in-flight/queued tasks of any *other* concurrent user of `database.async()` on that database, not just the caller. That review process (14 rounds) surfaced three related items that were explicitly out of scope for that PR but worth tracking here.

## 1. `setParallelLevel()` still tears down the whole pool

`DatabaseAsyncExecutorImpl.setParallelLevel()` was never touched by #6511 - only the two durability setters were fixed:

```java
@Override
public void setParallelLevel(final int parallelLevel) {
  if (parallelLevel != this.parallelLevel)
    createThreads(parallelLevel);
}
```

`createThreads()` is the exact same full-pool teardown-and-respawn mechanism #6511 removed from the durability setters. It's runtime-callable (`DatabaseAsyncExecutor.setParallelLevel()` is public API), so any caller changing the parallel level while another caller has in-flight/queued async work on the same database can still force-exit that unrelated work with `"Async executor has been shut down"`, the same failure mode #6505/#6509 diagnosed.

Unlike the durability flags, `parallelLevel` genuinely requires a different number of worker threads, so it can't become a plain volatile write the way `transactionUseWAL`/`transactionSync` did - but the current unconditional-teardown behavior could likely be softened (e.g. draining/quiescing before resizing rather than force-exiting in-flight tasks, or documenting that a parallel-level change should only be done when nothing else is using `database.async()`).

## 2. `AsyncCommandPool`'s DDL dispatch path ignores `transactionUseWAL`/`transactionSync` entirely

DDL commands dispatched via `database.async().command()`/`query()` with `awaitResponse=false` route through `DatabaseAsyncExecutorImpl.submitCommand()` → `AsyncCommandPool`, executed by `runCommand()`. Neither method reads or applies `transactionUseWAL`/`transactionSync` anywhere - that command runs under whatever ambient/default transaction WAL-use/sync settings the database (or, on the caller-runs fallback, the submitter's own thread/context) already has.

This is pre-existing behavior (`AsyncCommandPool` bypassed the ordinary `AsyncThread.executeTask()` path before #6511 too), not a regression - but #6511 established "the executor's durability policy applies per task" as the intended contract for the whole async executor, and this path is a latent exception to it. A caller who calls `setTransactionUseWAL(false)` before a bulk `GraphBatch` load and also dispatches DDL asynchronously during that window would find the DDL's durability unaffected by the setting, silently.

## 3. No observability for how often a durability-flag change forces a boundary commit

#6511's fix (`closeTransactionBoundaryIfDurabilityPolicyChanged()`) forces an out-of-band commit whenever a task's transaction was stamped with different flags than currently requested - correct, but it means two concurrent durability-flag-flippers sharing a worker slot can degrade that slot's effective batching toward `commitEvery=1` (documented as an accepted trade-off in the fix, since it's strictly better than the pre-fix full pool teardown).

There's currently no counter for how often this actually happens. The original #6505 production incident that motivated #6509 was only diagnosed after the fact by grepping `InterruptedIOException` counts out of logs over a 7-hour window. A cheap `AtomicLong` counter (e.g. exposed through `DatabaseAsyncExecutorImpl.DBAsyncStats`/`getStats()`, which already aggregates lock-free per-thread stats) tracking forced-boundary-commit count would make this specific failure mode visible in metrics/logs before it needs another production postmortem.

## References

- #6509 (fixed by #6511)
- #6505 (production incident that motivated #6509)
- #5665 (original diagnosis)

lvca:
Closes #6526.

The three follow-ups the 14-round review of #6511 left explicitly out of scope.

## 1. `setParallelLevel()` no longer tears down the whole pool

It was the last runtime-callable caller of `createThreads()`, the exact full teardown-and-respawn #6511 removed from the durability setters, so it kept both of that mechanism's failure modes for an **unrelated** concurrent user of `database.async()` on the same database:

- `executorThreads` was nulled for the entire teardown, so a producer scheduling in that window got `DatabaseOperationException("Async executor has been shut down")` - indistinguishable from a genuine `close()`. That is the #6505 incident: an index-compaction `onAfterCommit` hook running after `writeTransactionToWAL()` believed it and fenced the whole database over a scheduling hiccup that had nothing to do with the committing transaction.
- a worker whose queue was full refused the `FORCE_EXIT` marker within a second and was `interrupt()`ed, which unwound its in-flight task loudly and notified every task still queued on it as **completed without ever running it**.

Neither is inherent to changing the parallel level:

- **Growing** carries every existing worker over by reference and only starts the new ones. No marker, no flag, no join, no bucket-pinning disturbed.
- **Shrinking** publishes the truncated array *first* (nothing new can choose the retiring tail), then retires that tail by **draining** it: the `shutdown` flag plus a best-effort `FORCE_EXIT` at the **tail** of its queue, and *never* an `interrupt()`. Everything already queued still executes, inside the batch transaction that worker already owns, and that batch is committed by its own run loop on the way out.
- The retired workers are then **waited for**, outside `lifecycleLock` and bounded by `shutdownJoinTimeoutMs`. `getSlot()` maps a bucket to a worker modulo the pool size, so returning while a retired worker still drains would let two workers write one bucket's pages concurrently - the contention `ThreadBucketSelectionStrategy` exists to prevent. If the budget runs out they keep draining in the background and stay visible to `waitCompletion()`, `isProcessing()`, `getStats()` and `close()`/`kill()`, rather than having their work dropped to return on time.

`scheduleTask()` also re-maps a slot index a concurrent shrink has made stale: the array used to be null for the whole resize, so that race surfaced as the \"shut down\" exception above rather than as an `ArrayIndexOutOfBoundsException`.

The issue suggested \"soften the teardown, or document that a parallel-level change should only be done when nothing else is using `database.async()`\". Documenting it would have left the API with a footgun that only shows up under production concurrency; draining removes the footgun instead, which is why this PR takes that route.

## 2. The DDL dispatch path now carries the executor's durability policy

A command routed to `AsyncCommandPool` (an `awaitResponse=false` `CREATE INDEX`, since #6303) read neither `transactionUseWAL` nor `transactionSync`, so it committed under whatever ambient flags it found. A caller that set `setTransactionUseWAL(false)` around a bulk `GraphBatch` load and dispatched DDL inside that window silently got a durability it had not asked for.

`runCommand()` now stamps the executor's policy on the transaction it opens itself - and **only** on that one. On the caller-runs fallback the transaction may be the submitter's own, and `useWAL`/`walFlush` are read at *commit* time and govern the whole accumulated transaction, so stamping there would re-decide the durability of that caller's work. The stamp is restored afterwards, since `TransactionContext.begin()`/`reset()` never clear it and on that path the context belongs to a submitting HTTP worker that has its next request to serve.

## 3. Forced boundary commits are counted

`closeTransactionBoundaryIfDurabilityPolicyChanged()` is what lets a flag flip take effect without the pool teardown, at the cost of cutting the open batch short - so two concurrent flag-flippers sharing a worker slot can degrade that slot toward `commitEvery=1`. Nothing counted it; the incident behind #6509 was reconstructed by grepping `InterruptedIOException` counts out of a seven-hour log window.

Now exported as:
- `DatabaseAsyncExecutorImpl.DBAsyncStats.forcedBoundaryCommits`
- `Profiler` key `asyncForcedBoundaryCommits` (monotonic, inside the range retained across a database close - a bulk loader opens, loads and closes, so a counter reset on close would read zero every time anybody looked)
- Micrometer `arcadedb.engine.async.boundary.commits`

## Tests

`Issue6526AsyncExecutorFollowUpsTest`, five cases, each verified to FAIL with the corresponding fix reverted:

| test | reverted behaviour |
|---|---|
| `loweringTheParallelLevelRunsTheWorkStillQueuedOnTheRetiredWorkers` | the retired worker is `interrupt()`ed |
| `resizingTheParallelLevelNeverTellsAnUnrelatedProducerTheExecutorIsShutDown` | producer collects `"Async executor has been shut down"` |
| `raisingTheParallelLevelLeavesTheExistingWorkersRunning` | every worker respawned on a grow |
| `dispatchedDDLRunsUnderTheExecutorDurabilityPolicy` | dispatched DDL ignores `setTransactionUseWAL(false)` |
| `forcedBoundaryCommitsAreCounted` | counter stays at zero |

Green locally: 79 tests across `com.arcadedb.database.async.*`, `Issue6303*`, `Issue5665GraphBatchAsyncPoolChurnTest`, `MVCCTest`, `AsyncTest`; plus `EngineMetricsBinderTest`, `DefaultServerMetricsTest`, `AsyncInsertTest` in `server`.

## Files changed

- `engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java`
- `engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutor.java`
- `engine/src/main/java/com/arcadedb/Profiler.java`
- `server/src/main/java/com/arcadedb/server/monitor/EngineMetricsBinder.java`
- `server/src/test/java/com/arcadedb/server/monitor/EngineMetricsBinderTest.java`
- `engine/src/test/java/com/arcadedb/database/async/Issue6526AsyncExecutorFollowUpsTest.java` (new)

### DIFF
--- a/engine/src/main/java/com/arcadedb/Profiler.java
+++ b/engine/src/main/java/com/arcadedb/Profiler.java
@@ -65,13 +65,17 @@ public class Profiler {
   private static final int STAT_INDEX_COMPACTIONS = 15;
   private static final int STAT_WAL_PAGES_WRITTEN = 16;
   private static final int STAT_WAL_BYTES_WRITTEN = 17;
-  private static final int MONOTONIC_STATS        = 18;
-  private static final int STAT_WAL_TOTAL_FILES   = 18;
-  private static final int STAT_OPEN_FILES        = 19;
-  private static final int STAT_MAX_OPEN_FILES    = 20;
-  private static final int STAT_ASYNC_QUEUE       = 21;
-  private static final int STAT_ASYNC_PARALLEL    = 22;
-  private static final int STATS_COUNT            = 23;
+  // #6526: monotonic, and it belongs INSIDE the retained range - a durability-flag flipper is exactly the kind of
+  // caller (a bulk load through GraphBatch) that opens a database, does its work and closes it, so a counter that
+  // went back to zero on that close would be zero every time anybody looked at it.
+  private static final int STAT_ASYNC_BOUNDARY_COMMITS = 18;
+  private static final int MONOTONIC_STATS        = 19;
+  private static final int STAT_WAL_TOTAL_FILES   = 19;
+  private static final int STAT_OPEN_FILES        = 20;
+  private static final int STAT_MAX_OPEN_FILES    = 21;
+  private static final int STAT_ASYNC_QUEUE       = 22;
+  private static final int STAT_ASYNC_PARALLEL    = 23;
+  private static final int STATS_COUNT            = 24;
 
   /**
    * Registered database INSTANCES, compared by identity.
@@ -195,6 +199,11 @@ public class Profiler {
     final Map<String, Object> walStats = db.getTransactionManager().getStats();
     acc[STAT_WAL_PAGES_WRITTEN] += statOf(walStats, "pagesWritten");
     acc[STAT_WAL_BYTES_WRITTEN] += statOf(walStats, "bytesWritten");
+
+    // #6526: DatabaseAsyncExecutorImpl.getStats() is lock-free and documented to stay callable on a shut-down
+    // executor, which is the property this method requires of every source it reads - unregisterDatabase() calls it
+    // on a database that is already closing.
+    acc[STAT_ASYNC_BOUNDARY_COMMITS] += ((DatabaseAsyncExecutorImpl) db.async()).getStats().forcedBoundaryCommits;
     return walStats;
   }
 
@@ -220,6 +229,7 @@ public class Profiler {
     final long walTotalFiles = dbStats[STAT_WAL_TOTAL_FILES];
     final long asyncQueueLength = dbStats[STAT_ASYNC_QUEUE];
     final long asyncParallelLevel = dbStats[STAT_ASYNC_PARALLEL];
+    final long asyncForcedBoundaryCommits = dbStats[STAT_ASYNC_BOUNDARY_COMMITS];
 
     final long writeTx = dbStats[STAT_WRITE_TX];
     final long readTx = dbStats[STAT_READ_TX];
@@ -273,6 +283,10 @@ public class Profiler {
     json.put("pageFlushQueueMaxPerDatabase", new JSONObject().put("value", pStats.pageFlushQueueMaxPerDatabase));
     json.put("asyncQueueLength", new JSONObject().put("value", asyncQueueLength));
     json.put("asyncParallelLevel", new JSONObject().put("count", asyncParallelLevel));
+    // #6526: how often a durability-flag change (setTransactionUseWAL()/setTransactionSync()) cut an async worker's
+    // batch transaction short. #6511 traded the pool teardown for this extra commit; this is what makes the trade
+    // measurable instead of something to reconstruct from logs after an incident.
+    json.put("asyncForcedBoundaryCommits", new JSONObject().put("count", asyncForcedBoundaryCommits));
     json.put("pageCacheHits", new JSONObject().put("count", pageCacheHits));
     json.put("pageCacheMiss", new JSONObject().put("count", pageCacheMiss));
     json.put("totalOpenFiles", new JSONObject().put("count", totalOpenFiles));
--- a/engine/src/main/java/com/arcadedb/Profiler.java
+++ b/engine/src/main/java/com/arcadedb/Profiler.java
@@ -76,7 +76,10 @@ public class Profiler {
   private static final int STAT_MAX_OPEN_FILES    = 21;
   private static final int STAT_ASYNC_QUEUE       = 22;
   private static final int STAT_ASYNC_PARALLEL    = 23;
-  private static final int STATS_COUNT            = 24;
+  // #6526 review round 4: instantaneous, and firmly on this side of MONOTONIC_STATS - it returns to zero as the
+  // retired workers finish, so carrying it across a close would make a gauge that only ever grows.
+  private static final int STAT_ASYNC_RETIRING    = 24;
+  private static final int STATS_COUNT            = 25;
 
   /**
    * Registered database INSTANCES, compared by identity.
@@ -182,8 +185,10 @@ public class Profiler {
       // (issue #6526 review). A database with no executor has no queue and no workers, which is what the two
       // readings below now say instead of what asking the question made true.
       final DatabaseAsyncExecutorImpl async = asyncIfExists(db);
-      acc[STAT_ASYNC_QUEUE] += async != null ? async.getStats().queueSize : 0L;
+      final DatabaseAsyncExecutorImpl.DBAsyncStats asyncStats = async != null ? async.getStats() : null;
+      acc[STAT_ASYNC_QUEUE] += asyncStats != null ? asyncStats.queueSize : 0L;
       acc[STAT_ASYNC_PARALLEL] = async != null ? async.getParallelLevel() : 0L;
+      acc[STAT_ASYNC_RETIRING] += asyncStats != null ? asyncStats.retiringWorkers : 0L;
     }
     return acc;
   }
@@ -249,6 +254,7 @@ public class Profiler {
     final long asyncQueueLength = dbStats[STAT_ASYNC_QUEUE];
     final long asyncParallelLevel = dbStats[STAT_ASYNC_PARALLEL];
     final long asyncForcedBoundaryCommits = dbStats[STAT_ASYNC_BOUNDARY_COMMITS];
+    final long asyncRetiringWorkers = dbStats[STAT_ASYNC_RETIRING];
 
     final long writeTx = dbStats[STAT_WRITE_TX];
     final long readTx = dbStats[STAT_READ_TX];
@@ -306,6 +312,9 @@ public class Profiler {
     // batch transaction short. #6511 traded the pool teardown for this extra commit; this is what makes the trade
     // measurable instead of something to reconstruct from logs after an incident.
     json.put("asyncForcedBoundaryCommits", new JSONObject().put("count", asyncForcedBoundaryCommits));
+    // #6526 review round 4: "value", not "count" - this one goes back down, so it is a gauge. Persistently non-zero
+    // is a worker a lowered parallel level retired and that never finished draining; a resize never interrupts one.
+    json.put("asyncRetiringWorkers", new JSONObject().put("value", asyncRetiringWorkers));
     json.put("pageCacheHits", new JSONObject().put("count", pageCacheHits));
     json.put("pageCacheMiss", new JSONObject().put("count", pageCacheMiss));
     json.put("totalOpenFiles", new JSONObject().put("count", totalOpenFiles));
@@ -441,6 +450,7 @@ public class Profiler {
       final long asyncQueueLength = dbStats[STAT_ASYNC_QUEUE];
       final long asyncParallelLevel = dbStats[STAT_ASYNC_PARALLEL];
       final long asyncForcedBoundaryCommits = dbStats[STAT_ASYNC_BOUNDARY_COMMITS];
+      final long asyncRetiringWorkers = dbStats[STAT_ASYNC_RETIRING];
       final long totalOpenFiles = dbStats[STAT_OPEN_FILES];
       final long maxOpenFiles = dbStats[STAT_MAX_OPEN_FILES];
       final long walPagesWritten = dbStats[STAT_WAL_PAGES_WRITTEN];
@@ -532,7 +542,8 @@ public class Profiler {
       // each flip closing whatever worker's batch sees it next. Printed here as well as in toJSON() for the same
       // reason the #6217 read-path counters are: this dump is what an operator has when there is no metrics
       // endpoint to scrape.
-      buffer.append("%n    asyncForcedBoundaryCommits=%d".formatted(asyncForcedBoundaryCommits));
+      buffer.append("%n    asyncForcedBoundaryCommits=%d asyncRetiringWorkers=%d".formatted(
+          asyncForcedBoundaryCommits, asyncRetiringWorkers));
       buffer.append(
         "%n    scanType=%d scanBucket=%d iterateType=%d iterateBucket=%d countType=%d countBucket=%d".formatted(scanType,
           scanBucket, iterateType,
--- a/engine/src/main/java/com/arcadedb/Profiler.java
+++ b/engine/src/main/java/com/arcadedb/Profiler.java
@@ -19,6 +19,7 @@
 package com.arcadedb;
 
 import com.arcadedb.database.DatabaseInternal;
+import com.arcadedb.database.LocalDatabase;
 import com.arcadedb.database.async.DatabaseAsyncExecutorImpl;
 import com.arcadedb.engine.FileManager;
 import com.arcadedb.engine.PageManager;
@@ -176,8 +177,13 @@ public class Profiler {
       acc[STAT_OPEN_FILES] += fStats.totalOpenFiles;
       acc[STAT_MAX_OPEN_FILES] += fStats.maxOpenFiles;
 
-      acc[STAT_ASYNC_QUEUE] += ((DatabaseAsyncExecutorImpl) db.async()).getStats().queueSize;
-      acc[STAT_ASYNC_PARALLEL] = db.async().getParallelLevel();
+      // asyncIfExists(), never db.async(): that accessor CREATES the executor, worker threads included, so a
+      // metrics scrape used to grow a full pool on every open database that had never touched the async API
+      // (issue #6526 review). A database with no executor has no queue and no workers, which is what the two
+      // readings below now say instead of what asking the question made true.
+      final DatabaseAsyncExecutorImpl async = asyncIfExists(db);
+      acc[STAT_ASYNC_QUEUE] += async != null ? async.getStats().queueSize : 0L;
+      acc[STAT_ASYNC_PARALLEL] = async != null ? async.getParallelLevel() : 0L;
     }
     return acc;
   }
@@ -202,9 +208,22 @@ public class Profiler {
 
     // #6526: DatabaseAsyncExecutorImpl.getStats() is lock-free and documented to stay callable on a shut-down
     // executor, which is the property this method requires of every source it reads - unregisterDatabase() calls it
-    // on a database that is already closing.
-    acc[STAT_ASYNC_BOUNDARY_COMMITS] += ((DatabaseAsyncExecutorImpl) db.async()).getStats().forcedBoundaryCommits;
+    // on a database that is already closing. And it is reached through asyncIfExists() rather than db.async() for a
+    // second reason that only this path has: this runs ON THE CLOSE, after the close has already shut down whatever
+    // executor existed, so creating one here would leave its worker threads running with nothing left to stop them.
+    final DatabaseAsyncExecutorImpl async = asyncIfExists(db);
+    if (async != null)
+      acc[STAT_ASYNC_BOUNDARY_COMMITS] += async.getStats().forcedBoundaryCommits;
     return walStats;
+  }
+
+  /**
+   * The database's asynchronous executor <b>if it already has one</b>, and {@code null} otherwise. Every database in
+   * the registry is a {@link LocalDatabase} - it is what registers itself - so the check is a formality that keeps
+   * this class honest rather than a cast that assumes.
+   */
+  private static DatabaseAsyncExecutorImpl asyncIfExists(final DatabaseInternal db) {
+    return db instanceof final LocalDatabase local ? local.getAsyncIfExists() : null;
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/Profiler.java
+++ b/engine/src/main/java/com/arcadedb/Profiler.java
@@ -440,6 +440,7 @@ public class Profiler {
       final long[] dbStats = collectDatabaseStats();
       final long asyncQueueLength = dbStats[STAT_ASYNC_QUEUE];
       final long asyncParallelLevel = dbStats[STAT_ASYNC_PARALLEL];
+      final long asyncForcedBoundaryCommits = dbStats[STAT_ASYNC_BOUNDARY_COMMITS];
       final long totalOpenFiles = dbStats[STAT_OPEN_FILES];
       final long maxOpenFiles = dbStats[STAT_MAX_OPEN_FILES];
       final long walPagesWritten = dbStats[STAT_WAL_PAGES_WRITTEN];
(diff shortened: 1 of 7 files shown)


lvca:
Review round 1 addressed in 7d35cee — all three findings accepted, plus a regression the review's own framing led me to.

**1. `kill()` never reached retired workers** — correct and a real bug. `kill()` now goes through the same `concat(executorThreads, retiringThreads…)` treatment `shutdownThreadsLocked()` uses. New test `killReachesTheWorkersAShrinkLeftDraining` shrinks with `shutdownJoinTimeoutMs` cut to 10ms so the worker is provably still alive and registered, then kills; the gate task sleeps long enough that only an interrupt can end it, so the test cannot pass by the worker simply finishing. Verified to fail without the fix.

**2. Concurrent grow overtaking a shrink's drain** — correct. Fixed rather than documented: a new `resizeLock` (not `lifecycleLock`, which is exactly what a concurrent `close()`/`kill()` needs to end the wait) is held across the publish *and* the wait, so a grow cannot refill the vacated positions mid-drain. New test `aConcurrentGrowCannotOvertakeAShrinkStillDraining`, verified to fail without it. The residual you'd have accepted as a callout is still there and now stated in the javadoc: if the wait's own budget expires, a later grow can reuse the position while the old worker runs on — bounded and logged at WARNING.

**3. Slot-remap comment overstated** — right, `(v % old) % new ≠ v % new`. Behaviour unchanged; the comment now says what it is (a valid-worker fallback for a rare race, after which the next task for that bucket asks `getSlot()` again) and carries your `7 / 5 → 3` counterexample.

**4. The regression your "reads a stat source on close" framing uncovered.** Chasing finding 1's `retiringThreads` invariant, `AsyncFastQueueShutdownUndoTest` started failing: two live threads with the same worker name. Cause was mine, in the item-3 metric wiring — `Profiler.accumulateMonotonic` runs on `unregisterDatabase`, and I reached the counter through `db.async()`, which *lazily creates* the executor. On the close path that means a 17-thread pool started behind a close that had already shut down whatever executor existed, with nothing left to stop it.

`LocalDatabase` now exposes `getAsyncIfExists()` — the rule `waitForAsyncCompletion()` and `quiesceAsync()` already state in their comments — and `Profiler` uses it for all three async readings. That also closes the **pre-existing** version of the same footgun on the scrape path: `collectDatabaseStats` already called `db.async()`, so every Prometheus scrape materialized a worker pool for each open database that had never touched the async API. A database with no executor now reports queue `0` and parallel level `0` instead of making the question true by asking it. Flagging that as a deliberate, slightly-out-of-scope behaviour change on two existing metrics. New test `readingTheProfilerCountersNeverCreatesAnAsyncExecutor`, verified to fail without the fix.

Nit on the new tests themselves: the concurrency test's wait loop now sleeps instead of `Thread.onSpinWait()` — spinning on the waiting thread starved the thread it was waiting for and turned a 3s test into a 51s one.

Local: 95 engine tests green (`com.arcadedb.database.async.*`, `Issue6303*`, `Issue5665*`, `MVCCTest`, `AsyncTest`, `DatabaseLifecycleBackgroundThreadsTest`, `ProfilerTest`, `DatabaseFactoryTest`) and 32 in `server` (`EngineMetricsBinderTest`, `DefaultServerMetricsTest`, `MetricMeterTest`, `AsyncInsertTest`), plus the slow lane (`ACIDTransactionTest` 11/11).

lvca:
Review round 2 addressed in 1f32b78. All three were fair; none needed pushing back on.

**1. `awaitRetiredThreads()` reported only the first stuck worker** — right, and inconsistent with the method's own "bounded and loud" claim. The `break` is gone; every worker still alive past the budget gets its own WARNING, which is what separates "one slow task" from "the pool is wedged".

**2. `getStats()` double-count during a concurrent shrink** — real, and worth closing rather than accepting: the two reads are unsynchronized by design (the method must stay lock-free for `Profiler.toJSON()`), so a shrink landing between them publishes the smaller array *and* adds the worker to `retiringThreads`. The retiring loop now skips anything the first loop already saw. That is exact under every interleaving rather than merely less likely: whichever of the two reads observed the worker, it contributes exactly one term, and the scan is over pool-sized arrays. Applied the same to `allThreads()`, where the duplicate would have cost `waitCompletion()` a redundant marker on one worker instead — harmless, but there was no reason to hand it one.

**3. `@Tag("slow")`** — applied per-method rather than at class level, since only `loweringTheParallelLevelRunsTheWorkStillQueuedOnTheRetiredWorkers` and `aConcurrentGrowCannotOvertakeAShrinkStillDraining` park a worker for `WORKER_HOLD_MS`; CLAUDE.md asks for the method-level split when only some methods are slow. Both lanes verified: `-DexcludedGroups=slow,benchmark,vector` runs six tests in 0.36s, `-Dgroups=slow` runs the other two. The 3s hold is not padding — it has to outlive the hardcoded one-second `FORCE_EXIT` offer timeout in `shutdownThreadsLocked()`, which is the escalation those two tests exist to prove is gone.

On the two design notes, no change and agreed as framed: `setParallelLevel()` is now a blocking admin call (bounded by `shutdownJoinTimeoutMs`, 10s default) and that is the deliberate trade against the old teardown silently dropping other callers' queued work; the `Profiler` index renumbering is constant-driven with no numeric call sites.

Local after the change: 95 engine tests green across `com.arcadedb.database.async.*`, `Issue6303*`, `Issue5665*`, `MVCCTest`, `AsyncTest`, `DatabaseLifecycleBackgroundThreadsTest`, `ProfilerTest`, `DatabaseFactoryTest`.

lvca:
Review round 3 addressed in 3b96d28. Both nits were right; thanks especially for the first one, which was a real gap hiding behind a comment I had written too confidently.

**1. `getStats()` "exact under every interleaving"** — you're right that it wasn't, and the mechanism you traced is exactly it: `resizeThreads()` published the shrunk `executorThreads` *before* adding the retired worker to `retiringThreads`, so a reader landing between the two saw it in neither collection and missed that worker's queue for one snapshot — the mirror image of the double-count round 2 closed.

Rather than tighten the comment to a weaker claim, I closed the other half. The shrink is now three explicitly ordered steps: **register the retired workers, then publish the survivors, then ask the tail to stop**. A retired worker is therefore always in at least one of the two collections, and the identity check drops it when it is in both — so "exactly once" is now an invariant a reader can rely on rather than something the comment asserts. Registering before the stop request still holds for its own separate reason (a worker that exits immediately must be removed by its run loop *after* it was added, not before). Same window and same fix in `allThreads()`.

**2. `dumpMetrics()` missing `asyncForcedBoundaryCommits`** — added, and thanks for finding the precedent: #6217 did put its new counters in all three places, and a text dump is precisely what an operator has when there is no metrics endpoint to scrape. It sits next to the async queue and parallel-level readings, which is where it is read.

Also confirming your check on `getAsyncIfExists()` matches mine: `LocalDatabase.registerDatabase(this)` is the only caller, and `ServerDatabase`/`RaftReplicatedDatabase` wrap rather than register, so the `instanceof` really is a formality and not a silent no-op on a server deployment.

Local after the change: 95 engine tests green (`com.arcadedb.database.async.*`, `Issue6303*`, `Issue5665*`, `MVCCTest`, `AsyncTest`, `DatabaseLifecycleBackgroundThreadsTest`, `ProfilerTest`, `DatabaseFactoryTest`).

lvca:
Round 4 addressed in 8f307b5. Taking the suggestion rather than deferring it — it is item 3's own argument applied to the code this PR added, and deferring it would have left the new abandoned-drain path with exactly the "reconstruct it from logs afterwards" problem #6509 was diagnosed through.

**Retired-worker visibility.** `DBAsyncStats.retiringWorkers` now counts the workers a shrink retired that are still alive, exported as:
- `Profiler` key `asyncRetiringWorkers` and a `dumpMetrics()` line next to `asyncForcedBoundaryCommits`
- Micrometer `arcadedb.engine.async.workers.retiring`

Deliberately a **gauge**, not a counter, and placed outside `MONOTONIC_STATS` so it is not retained across a database close: it comes back down as the workers finish, so exporting it as a monotonic total would make every ordinary resize look like a leak — the same `#5636` reset artifact in reverse. Zero almost always; briefly positive during a resize; **persistently** positive is the alertable state, meaning a worker is wedged inside user code and a resize will never interrupt it.

Asserted inside `killReachesTheWorkersAShrinkLeftDraining` rather than in a new test, since that is the one place that reliably manufactures a straggler (drain budget cut to 10ms against a worker that only an interrupt can end): the gauge reads 1 while the worker is abandoned, reaches `Profiler.toJSON()` under the `"value"` wrapper, and returns to 0 once `kill()` has collected it. `EngineMetricsBinderTest` picks it up in the gauges list, which also asserts it is *not* registered as a counter.

On the verification note at the end of your review — agreed, and CI is the authority here. Locally after this change: 95 engine tests green (`com.arcadedb.database.async.*`, `Issue6303*`, `Issue5665*`, `MVCCTest`, `AsyncTest`, `DatabaseLifecycleBackgroundThreadsTest`, `ProfilerTest`, `DatabaseFactoryTest`), 18 in `server` (`EngineMetricsBinderTest`, `DefaultServerMetricsTest`, `AsyncInsertTest`), and the slow lane separately (`ACIDTransactionTest` 11/11).

Thanks for tracing the happens-before argument on `getStats()` by hand — that is the property the round-3 reordering was meant to establish, and having it confirmed from the JMM side rather than from the comment is worth more than the comment is.

lvca:
Round 5 addressed in f7cae43. Both actionable items taken; the third I'm leaving as you suggested.

**Test gap on the `slot %= threads.length` branch** — good catch, and you're right that it was only incidentally covered: the existing tests reach it just when a shrink happens to land in the right window, so the branch could rot without anything going red. `aSlotIndexLeftStaleByAShrinkStillReachesALiveWorker` now pins it deterministically instead of racing for it — take a real slot out of an eight-worker pool via `getSlot()`, shrink to two, then schedule with the index the racing caller would still be holding. It asserts the index is genuinely `>= getThreadCount()` before scheduling, so it cannot pass by landing in range by luck. Reverse-checked with the fallback removed: `ArrayIndexOutOfBoundsException: Index 7 out of bounds for length 2`, which is exactly the failure the branch exists to prevent.

**`previousUseWAL = true`** — agreed it reads oddly next to `previousWALFlush = null`. Neither value is ever observed: the `finally` restores them only when `ownTx != null`, and the branch that assigns `ownTx` assigns both first. Rather than invent a sentinel a `boolean` cannot carry, the comment now says the initial value is a compiler-mandated placeholder and not a default policy.

**Javadoc density** — leaving it, per your own note that it matches this file's established style (#4954/#5062/#6509). Worth recording the counter-argument for whoever revisits it: each block here exists because the invariant it describes was broken at least once in this PR's own review cycle — the "counted exactly once" comment was wrong in round 2, right-but-unproven in round 3, and only became a real invariant when the shrink was reordered. A class-level state-machine summary would be a genuine improvement, but it belongs in its own change rather than bolted onto this one.

**Self-join guard** — no change, and confirming your reading: a worker resizing its own pool from inside a task skips waiting on itself and still waits on the other retired peers. That is the right degradation, and nothing in the engine reaches it today.

On verification: agreed CI is the authority. Locally after this change, 96 engine tests green (`com.arcadedb.database.async.*` including `DatabaseAsyncExecutorKillTest`, `Issue6303*`, `Issue5665*`, `MVCCTest`, `AsyncTest`, `DatabaseLifecycleBackgroundThreadsTest`, `ProfilerTest`, `DatabaseFactoryTest`), 18 in `server` (`EngineMetricsBinderTest`, `DefaultServerMetricsTest`, `AsyncInsertTest`), and the slow lane separately.

lvca:
Round 6 addressed in a8a8567. The main finding is correct and it was a real defect — in the documentation rather than the code, which is arguably worse, since a javadoc that overstates a concurrency guarantee is how the next person builds on something that isn't there. Thank you for tracing it rather than taking the comment's word for it.

**The drain wait does not give back bucket exclusivity.** Confirmed, and your sequencing is exactly right: `resizeThreads()` publishes the truncated array before the retired workers are told to stop, `getSlot()`/`scheduleTask()` read it lock-free, and `resizeLock` serializes `setParallelLevel()` only against itself — so an unrelated producer, holding no lock and waiting for nothing, routes a migrated bucket to a survivor while the retired worker is still draining that same bucket. `resizingTheParallelLevelNeverTellsAnUnrelatedProducerTheExecutorIsShutDown` uses `scheduleTask(-1, …)` and genuinely cannot reach it.

I took option (a), and went a bit further than softening the wording. The javadoc now states what the wait *does* guarantee — when the call returns, the pool it reports is the pool that exists, so the caller's own next `getSlot()` cannot land on a bucket someone else is still writing and a second resize cannot reuse an occupied position — and then names the residual with the reasoning that makes it the right trade rather than just an admission:

> Changing the pool size changes which worker owns a bucket, so every resize has a handover, and there are only three ways to spend it: refuse the producers for its duration (the teardown's answer — the "shut down" exception of #6505), drop what the retired worker still holds (the teardown's other answer — somebody else's writes, silently), or let the two overlap while the old work finishes. Only the third loses nothing, and what it costs is bounded, self-correcting and already handled: an MVCC conflict on a migrating bucket, for the length of one drain, on a path that reports through `onError` and retries.

And why (b) is not on the table: closing it means blocking a producer inside `scheduleTask()` until the previous owner of its bucket finishes — a bounded wait on the lock-free hot path that index-compaction hooks reach *after* `writeTransactionToWAL()`, which is precisely the failure shape #6505 was.

**Test for the narrowed case** — `aProducerPinnedToOneBucketKeepsWorkingAcrossAShrink`. A producer follows `getSlot(3)` (bucket 3: `3 % 4 = 3`, retired; `3 % 2 = 1`, survivor) across three resizes down-up-down, exactly as `createRecord()` routes. It asserts the guarantee that holds — never refused, no task lost, every task executed — and deliberately does **not** assert the absence of overlap, since that would pin a guarantee the code does not make. Reverse-checked against the old teardown: fails with `Async executor has been shut down`.

**`resizeLock` fairness** — taken. It is held across a wait that can last `shutdownJoinTimeoutMs`, so barging could queue a caller behind an unbounded number of other resizes. On an administrative call taken once per bulk load, the throughput fairness costs is worth far less than the bound it buys, so it is now `new ReentrantLock(true)`.

Local: 97 engine tests green, 18 in `server`.

lvca:
Round 7 addressed in 05d33c8. Both findings are right, and they are the kind worth catching: in a file that carries its invariants in prose, a comment naming a cause that no longer exists is how a future reader talks themselves into reverting a fix.

**1. `compact()`'s catch** — correct. `setParallelLevel()` no longer goes through `createThreads()`, so the ambiguity that comment existed to explain is gone: the only thing that reaches that catch now is a genuine terminal `close()`/`kill()`, and "the pool is back within milliseconds" was exactly the wrong thing to say about it. Rewritten so the invariant leads (an exception escaping past `writeTransactionToWAL()` fences the database, #5053) and the *current* set of causes follows, with the history kept as history rather than as a live claim. The catch keeps all of its value — arguably more, since what it guards is now unambiguous.

**2. `quiesceWorkers()`'s early return** — also correct, and I'd add the second half you spotted: `resizeThreads()` returning without resurrecting *plus* `LocalDatabase.async` being assigned exactly once per instance is what makes a null-or-empty read there terminal. The defence stays (it costs nothing, and the property it now leans on lives in another class, which is precisely when a cheap guard earns its keep), but the residual it names is now the one that is actually still open: a **grow** racing the scan, not a resurrection.

**On the stale-slot fallback's reachability** — I checked, and it is shrink-only, not both directions. `getSlot()` returns `v % L1 < L1`; if the pool then grows to `L2 > L1` the index is still `< L1 < L2` and in bounds. Only a shrink (`L2 < L1`) can leave it out of range, which is what the javadoc says and what `aSlotIndexLeftStaleByAShrinkStillReachesALiveWorker` pins. So there is no grow-side case left uncovered by the round-5 test — good question to have asked, though, since "reachable from both directions" would have meant the test only covered half of it.

No code change in this commit, comments only; 97 engine tests still green.

lvca:
Round 8 addressed in a7a3875. Finding 1 is the best catch of this review — a real bug, introduced by this PR, and I reproduced it exactly as you described it.

**1. `quiesceWorkers()` vs. a concurrent shrink.** Confirmed, including the mechanism: `scheduleTask()` re-reads `executorThreads` per call, my stale-slot fallback re-maps an outgrown index onto a live worker instead of refusing it, so a shrink mid-loop folds several `DatabaseAsyncParkWorker` tasks onto one survivor. It runs the first, counts `parked` down once, blocks on `release`, and the duplicates behind it never run — a latch sized for the old pool that can never reach zero. And your framing of the severity is the right one: the old behaviour failed *immediately* with a "shut down" exception, so this turned a fast failure into a slow one. Measured **60.24s and `NeedRetryException`** where the answer should have been instant.

Fixed by taking `resizeLock` for the park scheduling. Worth recording that my first attempt at this was **wrong and the new test caught it**: I put only the loop under the lock, leaving the array snapshot outside — which is the same bug one window earlier, since a snapshot taken before the lock is already stale by the time it is acquired. It passed the single-method run and then failed in the full-class run at 66s. The read that sizes the latch and the loop that schedules against it are now both inside the lock.

Deliberately *not* held for the whole quiescence: a shrink landing after the loop is harmless (the park task is already ahead of the exit marker on the retired worker's queue, so it still runs, counts down and parks), and holding it longer would put an administrative resize behind an entire index build. Lock order is `quiesceLock` → `resizeLock` → `lifecycleLock`, one-directional — `setParallelLevel()` never touches `quiesceLock` and `close()`/`kill()` take only `lifecycleLock`.

New test `quiescingWhileTheParallelLevelChangesStillParksEveryWorker` drives it from the worst end — a resize looping against forty quiescences rather than one lucky interleaving — and reproduces the 60s stall with the lock removed. `@Tag("slow")`, since it is deliberately hostile.

**2. False-positive drain warning on self-shrink** — right, and an obvious asymmetry once pointed at: self was excluded from the join loop but not from the warning loop, so a worker retiring its own slot always logged "did not finish draining" about itself, while still executing the call that prints the line. Same exclusion applied.

**3. Thread-name reuse across shrink-then-regrow** — taken. A retired worker is now renamed with a `-retiring` suffix at the moment it is retired, rather than switching to monotonic ids: it keeps the `AsyncExecutor-<db>-<slot>` format every existing test and log filter matches on (including `kill()`'s "no thread of this database survives" prefix assertion), keeps the slot legible, and makes the duplicate impossible *and* self-describing in a thread dump — which is more useful during an incident than a unique but anonymous id.

Local: 108 engine tests green (adding `Issue6324*` to the sweep since the quiesce path is what its same-transaction index build uses).

lvca:
Round 9: no correctness findings, and the three observations are all ones I agree with as non-actionable. No code change for this round.

- **Javadoc density** — same conclusion as round 5, leaving it.
- **`contains()` being an O(pool) scan on a lock-free path** — worth stating explicitly since it reads like a cost that is always paid, and it is not: `allThreads()` returns the published array before allocating or scanning anything when `retiringThreads.isEmpty()`, and `getStats()`'s retiring loop iterates an empty set in the same state. So in steady operation — which is everything except the seconds after a shrink — no `contains()` call happens at all. When one does, it is a linear scan over an array bounded by the core count, against a window that only exists while a resize drains.
- **Per-method `@Tag("slow")`** — thanks; that split is now three methods, since `quiescingWhileTheParallelLevelChangesStillParksEveryWorker` is deliberately hostile enough to belong in the slow lane too.

Closing the loop from my side. Across nine rounds this review caught two things that were genuinely wrong rather than merely improvable — the `Profiler` lazy-creation leak on the close path (round 1's framing), and the `quiesceWorkers()` park-loop collapse this PR's own stale-slot fallback introduced (round 8) — plus one javadoc that claimed a guarantee the code does not make (round 6). All three are fixed with regression tests that fail against the pre-fix behaviour. Everything else has been answered in place.

CI is green on every completed check (`claude-review`, CodeQL across all five languages, Codacy, Meterian, Summary); `build-and-package` is still running. Ready for maintainer review.

PR #6538, opened 2026-08-21
fix(#6488): DISTINCT keys on graph elements must not depend on lazy-load state

### LINKED ISSUE #6488: `WALK` with a self-loop returns duplicate rows for `DISTINCT`
## Environment

- ArcadeDB 26.8.1 (`arcadedata/arcadedb:26.8.1`) 
- Clean database

## Reproduction

Run these setup statements separately:

```cypher
CREATE (a:A {name: 'a'});
MATCH (a:A {name: 'a'})
CREATE (a)-[:R]->(a);
```

Run the control query and consume the complete result:

```cypher
MATCH p = WALK (n0)-[*0..1]->(n1)
RETURN DISTINCT n1, n0;
```

The following query preserves `n0` and `n1`, but materializes the row through
an identity `CALL (*)` before returning the same expressions:

```cypher
MATCH p = WALK (n0)-[*0..1]->(n1)
WITH n0, n1
CALL (*) {
  RETURN 0 AS marker
}
WITH n0, n1
RETURN DISTINCT n1, n0;
```

## Expected result

Both queries should return one row. The path variable `p` is not returned, so
the zero-length path and the one-edge self-loop path have the same projected
values `(n1, n0)` and must be collapsed by `DISTINCT`.

## Actual result

The control query returns two identical rows:

```text
n1                         n0
-------------------------  -------------------------
(:A {name: 'a'})           (:A {name: 'a'})
(:A {name: 'a'})           (:A {name: 'a'})
```

The query with the identity subquery returns one row:

```text
n1                         n0
-------------------------  -------------------------
(:A {name: 'a'})           (:A {name: 'a'})
```

The two queries differ only by an identity row boundary that does not change
the returned variables. This makes the result depend on whether the unused
path variable is retained internally, and causes `DISTINCT` to fail to remove
the duplicate projected node pair.

lvca:
## Summary

Fixes #6488.

`RETURN DISTINCT`, `WITH DISTINCT`, `UNION`, `collect(DISTINCT ...)` and `count(DISTINCT ...)` all build a duplicate-elimination key from the projected values. For a document/vertex/edge, that key ultimately depended on `Object.toString()`:

- `ProjectReturnStep`/`UnionStep` fed the record straight into `DistinctNumericKey.canonicalize()`, which returned it unchanged and let the caller's `StringBuilder` call `toString()` on it.
- `WithStep`'s `DISTINCT` branch called `toString()` on the whole projected row directly, bypassing `DistinctNumericKey` entirely.

`Document`/`Vertex`/`Edge` `toString()` decorates the RID with the record's deserialized properties, and a not-yet-loaded property buffer renders as a placeholder (e.g. `#1:0[?]`) instead of the actual values - so two Java references to the very same record can render two different strings depending purely on in-memory load state, even though they represent the same value.

### How the issue exposed it

`MATCH p = WALK (n0)-[*0..1]->(n1)` on a single self-looping node enumerates two paths that both bind `n0`/`n1` to the same node: the zero-length path (whose end-node reference is the already-loaded start vertex) and the one-edge self-loop path (whose end-node reference is freshly obtained from the edge's endpoint and not yet deserialized). Both carry the same RID, but rendered as different strings, so `RETURN DISTINCT n1, n0` (and `WITH DISTINCT n1, n0`) failed to collapse the two rows - even though the path variable `p` itself was never returned.

Confirmed against Neo4j 5.26 (the openCypher reference implementation): both the reported query and its `CALL (*)` workaround return a single row there.

### Fix

- `DistinctNumericKey.canonicalize()` now canonicalizes a graph element to its `RID` instead of returning it unchanged. `RID`'s own `toString()`/`equals()`/`hashCode()` are stable and identity-based, which fixes `RETURN DISTINCT`, `UNION DISTINCT` and `count`/`collect(DISTINCT ...)` at the one shared utility used by all of them - not just the reported query shape.
- `WithStep`'s `DISTINCT` branch now builds its key from each projected property's canonicalized value, the same way `ProjectReturnStep` already did for `RETURN DISTINCT`, instead of stringifying the whole row.

### Regression test

`Issue6488DistinctHiddenPathVariableTest` covers `RETURN DISTINCT`, `RETURN DISTINCT` across an identity `CALL (*)` subquery boundary (the workaround from the report), and `WITH DISTINCT`.

## Test plan

- [x] New regression test `Issue6488DistinctHiddenPathVariableTest` (3 test methods) passes
- [x] Full `com.arcadedb.query.opencypher.**` test package (8446 tests, including the openCypher TCK suite) passes with 0 failures after the fix
- [x] Reproduced the bug directly against ArcadeDB before the fix and confirmed the fix resolves it, for both `RETURN DISTINCT` and `WITH DISTINCT`
- [x] Verified expected behavior against Neo4j 5.26 (openCypher reference implementation) for both query variants from the report

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/DistinctNumericKey.java
+++ b/engine/src/main/java/com/arcadedb/function/DistinctNumericKey.java
@@ -18,23 +18,37 @@
  */
 package com.arcadedb.function;
 
+import com.arcadedb.database.Identifiable;
+import com.arcadedb.database.RID;
+
 /**
- * Canonicalizes numeric values for duplicate-elimination purposes (Cypher {@code UNION},
+ * Canonicalizes values for duplicate-elimination purposes (Cypher {@code UNION},
  * {@code RETURN DISTINCT}, {@code count(DISTINCT ...)}, {@code collect(DISTINCT ...)}) so that
- * finite numeric values which compare equal under Cypher's {@code =} operator (e.g. {@code 1 = 1.0})
- * also collapse to the same key. Without this, duplicate-elimination paths that key a
- * {@link java.util.HashSet} / {@link java.util.HashMap} directly on the boxed value see
- * {@code Integer.valueOf(1)}, {@code Long.valueOf(1)} and {@code Double.valueOf(1.0)} as distinct,
- * because Java's per-type {@code equals()}/{@code hashCode()} never compares across boxed numeric
- * types. See issue #5789.
+ * values which compare equal under Cypher's {@code =} operator also collapse to the same key.
  * <p>
+ * <b>Numeric values:</b> finite numeric values which compare equal (e.g. {@code 1 = 1.0}) canonicalize
+ * to the same key. Without this, duplicate-elimination paths that key a {@link java.util.HashSet} /
+ * {@link java.util.HashMap} directly on the boxed value see {@code Integer.valueOf(1)},
+ * {@code Long.valueOf(1)} and {@code Double.valueOf(1.0)} as distinct, because Java's per-type
+ * {@code equals()}/{@code hashCode()} never compares across boxed numeric types. See issue #5789.
  * A value that is an exact-integer Java numeric type ({@link Long}, {@link Integer}, {@link Short},
  * {@link Byte}) canonicalizes to its {@code long} value, which is always exact. Any other
  * {@link Number} (typically {@link Double}, {@link Float}, or {@link java.math.BigDecimal})
  * canonicalizes to the same {@code long} value when it represents a finite integer within the range
  * a {@code double} represents integers exactly (|value| &lt;= 2^53), so it lines up with the
  * exact-integer-type canonicalization above; otherwise it canonicalizes to its {@code double} value.
- * Non-numeric values pass through unchanged.
+ * <p>
+ * <b>Graph elements:</b> a document/vertex/edge canonicalizes to its {@link RID}. Some callers build
+ * their duplicate-elimination key by calling {@code toString()} on the canonicalized value (e.g. to
+ * concatenate it into a composite string key); {@link com.arcadedb.database.BaseRecord#toString()} is
+ * RID-based and stable, but document/vertex/edge implementations that decorate it with the record's
+ * deserialized properties are not - a not-yet-loaded property buffer renders as a placeholder (e.g.
+ * {@code #1:0[?]}) instead of the actual property values, so two references to the very same record
+ * can render two different strings depending on load state alone, even though they represent one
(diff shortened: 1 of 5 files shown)


lvca:
Thanks for the review. Addressed:

1. **Delimiter collision in string-concatenated keys** - real, but pre-existing in `ProjectReturnStep`/`UnionStep` (not introduced by this PR; `WithStep` is just made consistent with the existing pattern here). Out of scope for #6488's fix. Filed #6540 to track fixing all three call sites together.
2. **`WITH DISTINCT *` doesn't filter internal/anonymous variables** - also pre-existing (same class as #5444, but for `WITH *`), not a regression from this PR. Filed #6541.
3. **Test coverage for UNION/collect/count(DISTINCT ...)** - agreed, this was a fair gap. Added `unionCollapsesLoadedAndUnloadedReferencesToSameNode` and `collectAndCountDistinctCollapseSelfLoopZeroAndOneHopPaths` to `Issue6488DistinctHiddenPathVariableTest`, both passing. Full `com.arcadedb.query.opencypher.**` package (8448 tests, incl. TCK) still green.

lvca:
Thanks, addressed the actionable point:

1. **Duplication across call sites** - extracted a shared `DistinctNumericKey.buildKey(names, valueOf)` helper and switched `ProjectReturnStep` (both branches), `WithStep`, and `UnionStep` to use it. Each caller still supplies its own name ordering (sorted `TreeSet`, declaration-ordered return items, or raw property order), so this is a pure dedup of the loop body with no behavior change - confirmed by the unchanged 8448/8448 pass count on `com.arcadedb.query.opencypher.**`.
2. **Delimiter collision** - already tracked in #6540 from the previous review pass.
3. **Javadoc `{@link com.arcadedb.database.BaseRecord#toString()}` nit** - checked: `BaseRecord` is a public class and `toString()` is a public, non-overloaded method (`engine/src/main/java/com/arcadedb/database/BaseRecord.java:127`), so the link target resolves. It's intentionally pointing at `BaseRecord`'s own stable implementation, in contrast with the `Document`/`ImmutableDocument` override the paragraph is warning about.

lvca:
\`ha-integration-tests\` failed on \`Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact\` - a Raft leader-forwarding "Record not found" race in a 3-node cluster slot-merge-delete test. That's in the HA/Raft replication path, unrelated to this PR's change (Cypher DISTINCT/UNION key-building, purely in-memory result dedup). All other lanes (unit, slow-unit, vector, TCK, integration, builder, all e2e) passed. Re-ran the job.

Issue #6545, opened 2026-08-21
Postgres extended-query bind() silently drops a statement while errorInTransaction, instead of erroring


robfrank:
## Description

Follow-up from the Claude Code review on PR #6542 (fix for #6457).

#6457/#6542 fixed the Postgres **simple-query** protocol path: previously, once a transaction was aborted (`errorInTransaction`), every subsequent statement was silently swallowed instead of returning an ErrorResponse (SQLSTATE 25P02), until a COMMIT/ROLLBACK cleared it.

The same silent-swallow pattern still exists on the **extended query protocol** path, in `PostgresNetworkExecutor.bindCommand()`:

```java
// postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java (~line 1761)
if (errorInTransaction)
  return;
```

When a transaction is aborted, a client's Bind message is silently accepted (the method returns without writing an ErrorResponse, and the caller ends up sending `bind complete`) instead of rejecting it with SQLSTATE 25P02 ("current transaction is aborted, commands ignored until end of transaction block"), which is what real Postgres does and what #6542 now does for the simple-query path.

This is distinct from #6543 (which covers the Parse-dispatch BEGIN/COMMIT/ROLLBACK tracking gap in `parseCommand()`) — this one is about the aborted-transaction guard in `bindCommand()` silently no-oping instead of erroring.

## Suggested fix

Mirror the simple-query fix from #6542: when `errorInTransaction` is true, write an ErrorResponse with SQLSTATE 25P02 instead of silently returning, so a client using the extended query protocol sees the same aborted-transaction behavior as one using simple queries.

## Test plan

A wire-level IT similar to `Issue6457AbortedTransactionSimpleQueryIT`, but driving Parse/Bind/Execute messages directly, asserting an ErrorResponse (not silent success) is returned for a Bind sent while `errorInTransaction` is set.

Issue #6543, opened 2026-08-21
Postgres extended query protocol: Parse dispatch recognizes BEGIN/COMMIT but not ROLLBACK


robfrank:
## Description

Follow-up from the Claude Code review on PR #6542 (fix for #6457).

In `PostgresNetworkExecutor.parseCommand()` (extended query protocol, Parse message handling), the dispatch that
tracks explicit-transaction state only recognizes `BEGIN`/`BEGIN TRANSACTION` and `COMMIT`:

```java
// postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java (~line 1899)
if ("BEGIN".equalsIgnoreCase(portal.query) || "BEGIN TRANSACTION".equalsIgnoreCase(portal.query)) {
  explicitTransactionStarted = true;
  setEmptyResultSet(portal);
} else if ("COMMIT".equalsIgnoreCase(portal.query)) {
  explicitTransactionStarted = false;
  setEmptyResultSet(portal);
}
```

There is no `ROLLBACK` branch. A bare `ROLLBACK` sent over the extended query protocol (Parse/Bind/Execute, as
opposed to the simple-query protocol fixed in #6542) falls through and gets parsed by the SQL engine
(`sqlEngine.parse(query.query, ...)`), which fails since `ROLLBACK` is not valid ArcadeDB SQL.

This is the same class of bug as #6457, just on the other wire path. #6457/#6542 were explicitly scoped to the
simple-query protocol only, so this was intentionally left out of that PR.

## Suggested fix

Add a `ROLLBACK`/`ROLLBACK TRANSACTION` branch alongside the existing `COMMIT` branch in the Parse dispatch, mirroring
the simple-query protocol fix: clear `explicitTransactionStarted`, call `database.rollback()`, and return an empty
result set instead of routing to the SQL engine.

## Also worth covering while there

The Claude review on #6542 additionally noted that `isCommitStatement`/`isRollbackStatement`/`isTransactionEndStatement`
(simple-query protocol, already fixed in #6542) don't match the SQL-standard `COMMIT WORK`/`ROLLBACK WORK`/`END WORK`
forms, only bare and `... TRANSACTION` forms. Not required for #6542, but could be picked up together with this issue
since it's the same helper family.

PR #6555, opened 2026-08-21
fix(#6548): recognize BEGIN/COMMIT/ROLLBACK-family statements in parseCommand() while aborted

### LINKED ISSUE #6548: Postgres extended query protocol: aborted-transaction ROLLBACK recovery never reaches the new dispatch (errorInTransaction early-return)
## Description

Follow-up from the Claude Code review on PR #6546 (fix for #6543).

`PostgresNetworkExecutor.parseCommand()`, `bindCommand()`, and `executeCommand()` (extended query protocol,
Parse/Bind/Execute) all start with an early return once the session is in an aborted-transaction state:

```java
if (errorInTransaction)
  return;
```

(`parseCommand()` at ~line 1848, `bindCommand()` at ~line 1762, `executeCommand()` at ~line 439.)

This means that once `errorInTransaction` is set, a client that sends an explicit `ROLLBACK` (or `COMMIT`/`END`,
which real Postgres treats identically while aborted) over the **extended** query protocol to recover from the
aborted transaction never reaches the BEGIN/COMMIT/ROLLBACK dispatch added in #6546 - `Parse` just silently
returns and the message is dropped.

When `Sync` eventually arrives, its `errorInTransaction` branch (`syncCommand()`, ~line 320-323) does call
`database.rollback()` and clears `errorInTransaction`, but it never touches `explicitTransactionStarted`. Since
`writeReadyForQueryMessage()` checks `errorInTransaction` first and `explicitTransactionStarted` second, the
client ends up back at `ReadyForQuery` status `'T'` instead of `'I'` after recovering from an aborted
transaction over the extended protocol - the same "wedged forever" symptom #6543 fixed, just reached via the
aborted-transaction path instead of a plain `ROLLBACK` with no prior error.

Compare with `queryCommand()`'s simple-query dispatch (issue #6457, PR #6542), which explicitly special-cases
`isTransactionEndStatement(...)` while `errorInTransaction` is true (lines 582-604) and clears *both*
`errorInTransaction` and `explicitTransactionStarted` there.

## Suggested fix

Mirror `queryCommand()`'s aborted-transaction handling in the extended-protocol dispatch: before (or instead
of) the unconditional `if (errorInTransaction) return;` in `parseCommand()`, recognize a
BEGIN/COMMIT/ROLLBACK-family statement (via the existing `isCommitStatement`/`isRollbackStatement` helpers) and
clear `explicitTransactionStarted` there (in addition to whatever `errorInTransaction` handling is appropriate
for `bindCommand()`/`executeCommand()` on the resulting portal). This is a materially bigger change than #6546
- it touches three call sites instead of one and needs to reason about portal state across Parse/Bind/Execute
while aborted - which is why it was scoped out rather than folded into that PR.

## Also worth covering while there

`isCommitStatement`/`isRollbackStatement`/`isTransactionEndStatement` now cover `COMMIT`/`COMMIT TRANSACTION`/
`COMMIT WORK`/`END`/`END TRANSACTION`/`END WORK`/`ROLLBACK`/`ROLLBACK TRANSACTION`/`ROLLBACK WORK` (as of
#6546), so whatever fixes this should have all of those forms available for free via the existing helpers.

robfrank:
## What does this PR do?

`PostgresNetworkExecutor.parseCommand()` started with an unconditional `if (errorInTransaction) return;` before ever inspecting the query text. Once a statement erred inside an explicit `BEGIN` block, a client sending an explicit `ROLLBACK` (or `COMMIT`/`END`, which real Postgres treats identically while aborted) over the **extended query protocol** to recover from the aborted transaction never reached the BEGIN/COMMIT/ROLLBACK dispatch #6546 added to the simple-query path - `Parse` just silently dropped the message.

`Sync`'s own `errorInTransaction` branch calls `database.rollback()` and clears `errorInTransaction`, but never touches `explicitTransactionStarted`. Since `writeReadyForQueryMessage()` checks `errorInTransaction` first and `explicitTransactionStarted` second, the client ended up back at `ReadyForQuery` status `'T'` instead of `'I'` after "recovering" - the same "wedged forever" symptom #6543 fixed, just reached via the aborted-transaction path.

`parseCommand()` now recognizes a transaction-end statement (via the existing `isTransactionEndStatement`/`isCommitStatement`/`isRollbackStatement` helpers) before falling through to the early return while aborted, mirroring `queryCommand()`'s aborted-transaction dispatch:
- Rolls back the active transaction if one is active.
- Clears both `explicitTransactionStarted` and `errorInTransaction`.
- Overrides the portal's query text to `"ROLLBACK"` so `executeCommand()` later reports the `ROLLBACK` command tag regardless of whether the client sent `COMMIT`/`END`/`ROLLBACK` - real Postgres treats a `COMMIT`/`END` of an aborted transaction the same as `ROLLBACK`.
- Stores the resulting portal and writes `ParseComplete`, so the following Bind/Execute/Sync round trip runs through the ordinary (non-aborted) code paths without any changes needed in `bindCommand()`/`executeCommand()` themselves.

Every other statement while aborted still falls through to the silent `return` - `bindCommand()`'s own `errorInTransaction` check (#6545) is what turns an attempt to Bind one of those into an `ErrorResponse`.

## Motivation

Follow-up from the Claude Code review on PR #6546 (fix for #6543), filed as #6548. Without this fix, a client that reaches an aborted transaction while inside an explicit `BEGIN` over the extended protocol has no way back to idle status via extended-protocol `ROLLBACK`/`COMMIT`/`END` - and for many real clients (which re-`Parse` the unnamed statement/portal per statement) the attempt actually desyncs and closes the connection, since `Execute` always removes its portal and the dropped `Parse` leaves nothing for the following `Bind` to find.

## Related issues

Closes #6548

## Additional Notes

New regression test `Issue6548AbortedTransactionRollbackExtendedProtocolIT`:
- Reproduces the wedge: `BEGIN` (extended) -> a malformed `Parse` aborts the transaction -> `ROLLBACK` sent as its own Parse/Bind/Execute/Sync must clear both flags, reporting status `'I'`, not `'T'`.
- Verified red against the unfixed code first (git-stashed the source fix, ran the test, restored it) - it failed with a protocol desync rather than a clean assertion failure, confirming the bug is sharper than "wrong status": some clients get their connection closed outright.
- Covers `COMMIT`/`END` while aborted producing the `ROLLBACK` command tag.
- Confirms the session is fully usable again after recovery.

Full aborted-transaction/rollback regression group (`Issue6543RollbackExtendedProtocolIT`, `Issue6545AbortedTransactionExtendedQueryIT`, `Issue6457AbortedTransactionSimpleQueryIT`, plus this new class) passes: 9/9. Broader `postgresw` integration-test suite: 19/19 of the tests that ran passed; `PostgresProtocolIT` crashed its forked JVM under severe host memory pressure (unrelated general JDBC connectivity test, reproduced the same crash in isolation) - treated as a pre-existing environmental flake, not a regression from this change.

## Checklist

### DIFF
--- a/docs/6548-extended-protocol-rollback-recovery.md
+++ b/docs/6548-extended-protocol-rollback-recovery.md
@@ -47,6 +47,24 @@ one of those into an `ErrorResponse`.
 
 - `postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java` - `parseCommand()`.
 
+## Review follow-up: trailing ';' / whitespace
+
+`isCommitStatement()`/`isRollbackStatement()` match by exact string equality, so `abortedUpperCaseText` has to
+be trimmed and stripped of a trailing `;` before comparison - the same normalization `queryCommand()` applies
+to its own `queryText` (`queryText = readString().trim(); if (queryText.endsWith(";")) queryText = ...`)
+before its aborted-transaction check runs. Without it, a client that sends `"ROLLBACK;"` (a real Postgres
+statement terminator many drivers append) over the extended protocol while aborted falls through to the
+silent `return`, reproducing this issue's "wedged forever" symptom via a trailing semicolon instead of via
+the missing dispatch. Caught by the Claude Code review on this PR.
+
+Fixed narrowly: a local `abortedText` is trimmed and semicolon-stripped before uppercasing, only for the
+match this PR's new branch performs. `portal.query` itself is left untouched (the matched branch overwrites
+it outright with `"ROLLBACK"`; the unmatched branch discards the portal). The non-aborted BEGIN/COMMIT/
+ROLLBACK dispatch a few lines below (`upperCaseText` at parseCommand()'s top) has the identical gap, but that
+variable is also read by several unrelated checks (SET/SHOW/SAVEPOINT/system-query), so normalizing it there
+is a materially larger, riskier change outside this PR's scope - left as a pre-existing issue, not introduced
+here, callable out as a fast follow if desired.
+
 ## Tests
 
 - `postgresw/src/test/java/com/arcadedb/postgres/Issue6548AbortedTransactionRollbackExtendedProtocolIT.java` (new)
@@ -54,6 +72,9 @@ one of those into an `ErrorResponse`.
     sent as its own Parse/Bind/Execute/Sync must clear both `errorInTransaction` and
     `explicitTransactionStarted`, reporting status `'I'`, not `'T'`.
   - Covers `COMMIT`/`END` while aborted producing the `ROLLBACK` command tag, per real Postgres semantics.
+  - Covers `"ROLLBACK;"`, `" ROLLBACK "`, and `" ROLLBACK; "` while aborted, locking in the trim/semicolon-strip
+    fix above - verified red against the code before that fix (protocol desync, same signature as the
+    original bug), green after.
   - Confirms the session is fully usable again after recovery (`SELECT 1` succeeds with no `ErrorResponse`).
 
 ## Verification
--- a/docs/6548-extended-protocol-rollback-recovery.md
+++ b/docs/6548-extended-protocol-rollback-recovery.md
@@ -1,0 +1,84 @@
+# Issue #6548: Postgres extended-protocol ROLLBACK recovery never reaches the new dispatch
+
+## Root cause
+
+`PostgresNetworkExecutor.parseCommand()` starts with:
+
+```java
+if (errorInTransaction)
+  return;
+```
+
+This early return happens **before** the query text is inspected at all. Once `errorInTransaction` is
+set (a statement erred inside an explicit `BEGIN` block), a client that sends an explicit `ROLLBACK` (or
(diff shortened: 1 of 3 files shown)


robfrank:
Fixed the semicolon/whitespace gap in \`abortedUpperCaseText\` (2938d1d): a local \`abortedText\` is trimmed and stripped of a trailing \`;\` before uppercasing, mirroring \`queryCommand()\`'s own normalization. \`portal.query\` itself is left untouched - the matched branch overwrites it with \`"ROLLBACK"\` regardless, and the unmatched branch discards the portal.

Verified red first: reverted just that change, ran a new test with \`"ROLLBACK;"\`/\`" ROLLBACK "\`/\`" ROLLBACK; "\` variants - same protocol-desync signature as the original bug. Green after restoring the fix.

Left the non-aborted BEGIN/COMMIT/ROLLBACK dispatch's \`upperCaseText\` (parseCommand()'s top) untouched - it has the identical gap, but that variable also feeds several unrelated checks (SET/SHOW/SAVEPOINT/system-query), so normalizing it is a materially larger change than this PR's scope. Noted in the tracking doc as a pre-existing, separate issue.

Issue #6468, opened 2026-08-19
JSONL import swallows per-record errors and reports success; a failed vertex cascade-drops its edges


ruispereira:
### Summary
The JSONL importer catches every per-record exception, logs it at SEVERE, and continues, and `load()` commits and returns normally regardless — so a partial import reports success. A single failed vertex leaves its RID out of the RID map, which then cascade-drops every edge touching it. This is the delivery mechanism that turns per-value import bugs into whole-subgraph loss with a "success" result.

### Where (HEAD `0291182af`)
`integration/src/main/java/com/arcadedb/integration/importer/format/JsonlImporterFormat.java`:
- `loadDocument`/`loadVertex`/`loadEdge` each `catch (Exception e) { LogManager...SEVERE; }` and continue (`:308-310`, `:326-328`, `:365-367`).
- A skipped vertex never does `ridIndex.put(oldRid, ...)`, so its edges hit the `newOut/newIn == null` branch (`:339-352`) → logged SEVERE → dropped.
- `load()` commits every N rows and returns normally (`:99-109`); there is no aggregate failure count or non-zero exit.

### Repro
Import any JSONL export containing a record that fails to deserialize (e.g. the DATE-unit bug, a schema mismatch): the record and all its edges vanish, but the import prints its counters as if complete and exits successfully.

### Suggested fix
Track a skipped-record count in `ImporterContext` and fail the operation (or surface a non-zero status / structured error summary) when any record was skipped, unless an explicit `--skip-errors` flag is set.

### Scope / confidence
High (code read). This is the amplifier behind the temporal/LINK fidelity bugs; on its own it is a "fail loud vs fail silent" policy issue, so framing it as a correctness/observability fix rather than a crash.

PR #6564, opened 2026-08-21
fix(#6379): sparse-vector partial compaction resurrected deleted documents

### LINKED ISSUE #6379: Sparse-vector partial compaction can permanently resurrect deleted documents (newest-wins inverted by new-highest merge id)
### Summary
In the paginated sparse-vector index, a **partial** compaction can permanently resurrect a deleted document (or revert an update). Precedence between segments is "newest wins", but a partial merge is assigned a **globally new highest** segment id and appended as the newest segment, while the partial selectors can leave a genuinely newer segment (the one holding the tombstone) out of the merge. The merged output — carrying the older, live posting — then outranks that excluded tombstone.

### Where (HEAD `ca89f56`)
`engine/src/main/java/com/arcadedb/index/sparsevector/PaginatedSparseVectorEngine.java`
- `compactInputs(...)` L1031: `final long newId = nextSegmentId.getAndIncrement();` — the merged segment always gets the current highest id.
- `replaceSegments(toRemove, maybeNew)` L1696-1713: keeps the surviving segments in order and **appends `maybeNew` last**.
- `openMergedCursor(...)` adds segment cursors in array order; `DimCursor` (`DimCursor.java:383-387`) resolves equal-RID postings by "higher source index wins" (its comment: *"i > newestAtMinIdx since we scan oldest-first ... keeps newest-wins"*). Array order == ascending id (ids are monotonic and every new/merged segment is appended last), so the merged segment (highest id, last position) is treated as newest.
- Partial selectors that can pick a non-suffix / non-contiguous subset: size-tiered `sameTier.sort(...); ...` (~L954-960) and tombstone-ratio (~L990-1004). `compactOldest` (~L898-907) picks a contiguous oldest run.

### Failure sequence
- `put(d, r1)` → flush → segment S1 (live r1).
- `remove(d, r1)` → flush → segment S2 (tombstone r1). A query now correctly omits r1 (S2 newer, tombstone wins).
- A size-tiered (or tombstone-ratio) partial compaction selects a subset that includes S1 but **not** S2 (e.g. S1 and a later same-tier segment S3, while S2 is a different tier). The merge writes S_m with `newId > S2.id`, appended last.
- Now S_m (live r1, newest position) outranks S2 (tombstone r1) → **r1 reappears in topK, permanently.**
- `compactAll` / `REBUILD INDEX` then merges everything with S_m as the newest input → the tombstone is dropped for good. The update variant reverts a posting to its stale weight the same way.

### Suggested fix
Do not give a partial-merge output a globally-new id. Assign it `max(inputIds)` and position it where the newest input sat, so it cannot leapfrog a non-input segment; or resolve "newest wins" on a per-posting monotonic sequence number rather than on segment id/position. Restricting each partial merge to a contiguous id-sorted run is necessary but not sufficient on its own given the new-highest-id assignment.

### Scope / confidence
High confidence on the mechanism (read at HEAD: `newId` is always the new highest, `replaceSegments` appends last, `DimCursor` newest-wins by source index). Medium that a default size-tiered deployment triggers it without operator help — it needs a partial selector to pick a subset excluding the segment that tombstoned a key still live in an included older segment. Mitigation today: use only full `compactAll` and disable partial-compaction triggers. The existing tombstone test only ever compacts a full suffix, so it misses this.

lvca:
Fixes #6379

## The bug

Precedence between sparse-vector segments is "newest wins", resolved by position in the segments array, which was sorted by segment id. A compaction allocated a **globally new highest** id and `replaceSegments` **appended it last**, so the merged segment was treated as the newest - even when the partial selectors had left a genuinely newer segment out of the merge.

If that skipped segment held the tombstone for a live posting the merge carried forward from an older input, the merged output outranked the tombstone and the RID reappeared in `topK`. The next `compactAll` / `REBUILD INDEX` then dropped the tombstone as the loser, making it permanent. The same inversion reverts an update to its stale weight.

Both partial selectors could produce a gapped pick:
- **size-tiered**: a small tombstone-carrying flush lands in a lower tier than the fat data segments around it, so a tier overflow merge jumps straight over it;
- **tombstone-ratio**: it paired the delete-heavy segment with the *globally* oldest neighbours, skipping whatever sat in between.

The existing tombstone test only ever compacted a full suffix, which is why it missed this.

## Why not the fix the issue proposes

The issue suggests assigning the merged segment `max(inputIds)`. That cannot be the segment id: the id names the component file, and the build creates the merged component while its inputs' files are still live, so re-using an input's id collides. It also suggests that restricting merges to a contiguous run is "necessary but not sufficient". It is the other way round - **epoch inheritance and adjacency are each necessary and together sufficient**, and per-posting sequence numbers (the issue's other option) would cost 8 bytes on every posting to buy nothing extra.

So the fix separates the two jobs the segment id was doing, and adds the piece neither option covers - what happens to compaction throughput once merges are constrained.

## The fix

**1. A persisted recency epoch, decoupled from the id.** It goes in manifest slot 1, which was already reserved, so the format version does not move; segments written before it read `0` and fall back to their id, which reproduces exactly the order they were written under. A flush takes its id as its epoch (a freshly sealed memtable genuinely is the newest thing there is); a merge inherits the epoch of its **newest input**, so it lands where that input sat instead of jumping to the front. The id keeps its only real job.

**2. Adjacency.** Inheriting the newest input's epoch is only sound when nothing the merge skipped sits between its inputs - otherwise a posting carried forward from the *oldest* input is still stamped with the *newest* input's epoch and still outranks whatever was in the gap. The size-tiered trigger now takes an adjacent same-tier run; the tombstone-ratio trigger an adjacent window ending at the offender. `contiguousClosure` enforces the invariant independently of the selectors, so a future policy that returns a gapped pick can only over-merge (with a `WARNING`), never corrupt.

In the steady state a bulk-load produces this costs nothing - equal-sized flushes arrive consecutively, so a tier's members already *are* adjacent.

**3. A shape guard**, because adjacency introduces a way to stall: an interleaved size distribution (a half-full flush at close time, a tombstone-only flush, a merge output landing mid-array) can leave every tier's members non-adjacent, and unbounded segment growth costs a file handle and a cursor per segment on every query. Once the active count exceeds what a healthy size-tiered shape holds for the current posting total, the guard merges the **cheapest adjacent window**. No new tuning knob - the ceiling is `(fanout - 1) * (tiers spanned)`, the figure the class Javadoc already derives.

**4. Ordering at every publication point.** Flush, compaction, the FileManager refresh followers rebuild from, and the open-from-disk scan all order by epoch now; the id allocator is primed from the *maximum* id rather than the last element, since those are no longer the same thing.

## Tests

`PaginatedSparseVectorEngineCompactionRecencyTest` covers the two halves separately, so a regression in either is attributable:

| test | half | verified to fail before the fix |
|---|---|---|
| `partialCompactionOfOlderSegmentsKeepsDeletedRidInvisible` | epoch (`compactOldest` over a contiguous prefix) | ✅ |
| `sizeTieredPartialCompactionKeepsDeletedRidInvisible` | adjacency (primary trigger) | ✅ |
| `tombstoneRatioPartialCompactionKeepsDeletedRidInvisible` | adjacency (secondary trigger) | ✅ |
| `sizeTieredPartialCompactionKeepsNewestWeight` | the update variant | ✅ |
| `mergedSegmentRecencyEpochSurvivesReopen` | epoch persistence + HA-follower rebuild order | ✅ |
| `interleavedTiersStillCompactUnderTheShapeGuard` | the guard the constraint made necessary | n/a (new behaviour) |

Full engine module: **13144 tests, 0 failures** (`mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow`).

## Upgrade note

A segment merged before this commit already carries the wrong precedence on disk and reads back with `epoch = id`. There is no safe automatic repair - the information about where it *should* have sat is gone. Run `REBUILD INDEX <name>` once on an affected sparse-vector index to collapse the segments into a correctly ordered shape. Docs for this are staged in `ArcadeData/arcadedb-docs` (`concepts/vector-search.adoc` and `how-to/operations/upgrade.adoc`), which is a separate repository from this one.

**HA rolling upgrade.** The corrected ordering is recorded in the segment file, but *reading* it correctly requires the fix. During a staged rollout a node still on the older build that receives a segment merged by an upgraded leader falls back to ordering by `segmentId` and can still return a deleted document, until it is upgraded too. Upgrade every node before running `REBUILD INDEX`, and treat sparse-vector results from a mixed-version cluster as provisional. Documented in `how-to/operations/upgrade.adoc`, staged in the separate `ArcadeData/arcadedb-docs` repository (not in this diff).


lvca:
Thanks — addressed, with one point declined and a reason.

**1. Dead clamp in the tombstone-ratio trigger** — valid, fixed in `51e31d1c1` (that push crossed with this review). I re-derived the bound rather than take it on trust and it holds. Worth noting the comment was *worse* than the code: it named the newest end as the case the clamp guarded, when reaching backwards means the window always ends at the offender, so the newest end is never the boundary. The real edge case is the **oldest** end, where there are fewer than `tierFanout - 1` older neighbours and the window starts at 0 and extends *forwards* past the offender — the only case where it holds anything newer. That branch had no coverage, so `tombstoneRatioWindowAtTheOldestEndKeepsDeletedRidInvisible` now pins it.

**2. Redundant sort** — declining, and I've said so in the code so it doesn't read as an oversight next time. `contiguousClosure` re-sorting a snapshot its callers already sorted is the point: threading their `ordered` array through would make the check trust the caller for the very ordering it exists to verify. It's the one input it must not accept second-hand. The cost is a second `O(n log n)` on an array bounded by `(fanout - 1) × tier count`, against merge I/O.

**3. `contiguousClosure` widening untested** — valid, and the sharpest observation of the three, since that branch is the entire basis for the "a future policy can only over-merge, never corrupt" claim. Fixed in `defdfbc99`: dropped to package-private plus a `segmentsForTest()` hook, and `contiguousClosureWidensAGappedPick` now asserts all three outcomes — a gapped pick widens to the spanning run, an already-contiguous pick survives unchanged and oldest-first, and a pick naming a segment outside the live snapshot throws.

**4. Unrelated `JsonLImporterIT.java` hunk** — keeping it, deliberately. It isn't cosmetic: that file landed on `main` in c816a66c7 (#6468) without a trailing newline, which fails the `end-of-file-fixer` pre-commit hook, which fails the `setup` job, which makes **every** downstream test lane report `skipping` rather than running. I saw it directly on this PR — first CI run: `setup fail`, 15 lanes `skipping`; after the one-byte fix: `setup pass` and the lanes actually execute. Dropping it would leave this PR unable to go green, and it's currently doing the same to every other PR open against main. Happy to split it into its own PR if you'd rather, but then this one stays red until that merges.

On the sandbox note: `mvn -o -pl engine -am test -DexcludedGroups=benchmark,vector,slow` is green locally — **13144 tests, 0 failures** — and the sparse-vector package is 115 tests, 0 failures after the changes above. Five of the eight new tests were confirmed red before the fix and green after.

Also added a "Segment compaction and deletes" section to `concepts/vector-search.adoc` in `arcadedb-docs` covering the `REBUILD INDEX` upgrade note, so it doesn't live only in this PR description (committed there, awaiting review).

lvca:
Both rounds addressed — thanks, the two new points were worth having.

**`setRecencyEpoch` ordering footgun** — fixed in `20cdfcfc9`, with a test. Agreed this deserved enforcement rather than a Javadoc line: `setSegmentId` re-defaults the epoch, so the wrong order discarded the explicit value in silence, and the failure mode of that is a merged segment carrying its own id as its epoch — i.e. #6379 itself, reappearing as resurrected documents with nothing pointing back at the builder. It throws now.

**Stale `reserved(16)` comment** — fixed in the same commit, and I took the chance to say the part that actually matters: slot 0 stopped being reserved with the tombstone-ratio trigger and slot 1 stops here, so *nothing* is spare in that manifest any more and the next addition needs a real size bump. The `manifestSize` inline comment said "2 reserved" too; that's updated.

**Redundant sort** — still declining, now stated at the call site. `contiguousClosure` re-sorting a snapshot its callers already sorted is the point of the check: threading their array through would make it trust the caller for the one thing it exists to verify.

**Release-notes/upgrade visibility** — good catch that a PR description is the wrong home for this, especially since there's no automatic detection and an affected index looks perfectly healthy. Added to `arcadedb-docs`:
- `how-to/operations/upgrade.adoc` — a `26.9.1` entry under **Breaking changes**, in the same Impact/Action shape as the existing per-version entries, so an upgrading operator meets it where they'd look.
- `concepts/vector-search.adoc` — a "Segment compaction and deletes" section explaining why segment order decides delete visibility in the first place.

Both committed in the docs repo for review.

**Concurrency-race gap** — agreed it's out of scope: this PR changes selection and ordering, not locking. Every selector still runs inside `compactInputs` under `mutatorLock`, which `flush`/`put`/`remove` already serialize against, and `LSMSparseVectorIndexConcurrencyTest` still passes.

Sparse-vector package is now 116 tests, 0 failures; full engine module 13144, 0 failures.

lvca:
Implemented the silent-migration suggestion — thanks, this was the most consequential note of the round. Pushed as `ca08e5da4`.

You're right that the condition is a precise signal, and I checked it is *exact* rather than merely high-confidence, which is what made it worth wiring in:

- a merge written after the fix inherits its newest input's epoch, and every input's epoch was issued **before** the id that merge allocated, so its epoch is always strictly *below* its own id — never equal;
- a segment sealed from a memtable has no parents at all.

So `parentSegmentCount() > 0 && recencyEpoch() == segmentId()` matches exactly one thing: a merge whose manifest slot read back `0` because it predates the field. Neither half can false-positive.

Two deliberate choices:
- **Open-time only**, not in `refreshSegmentsFromFileManager` — that runs on the follower query path, so the same line would repeat for as long as the condition holds. One line per index open, naming `REBUILD INDEX`.
- **`parentSegmentCount()` rather than `parentSegments().length`** — the existing accessor defensively clones, and this scan asks every segment in the index at open.

Test synthesises the legacy shape (parents recorded, epoch left at the segment id — what the old code actually wrote) and asserts both directions: fires there, silent for an index whose merges were all written by this version. I confirmed it isn't vacuous by removing the warning call and watching it go red.

This also closes the gap between the two halves of the response to #6379: the code now prevents new occurrences *and* points at the ones already on disk, instead of relying on the operator having read the upgrade note.

Verification for this commit: sparse-vector package 116 tests / 0 failures, and the detection is provably inert elsewhere — `setParentSegments` has exactly two callers in the tree (the production compaction path, which always sets the epoch immediately after, and the new test). A full-module run is in flight and I'll flag it here if anything turns up.

lvca:
Addressed — `dda312eac`.

**1. `mergedSegmentRecencyEpochSurvivesReopen` doesn't exercise the follower path** — correct, and it's a fair hit on my own PR description rather than on the code. That test drives `loadExistingSegments`; a follower rebuilds through `refreshSegmentsFromFileManager` on its next query. Both sort by `RECENCY_ORDER`, so the claim happened to be *true* — but nothing was holding it that way, which is the part that matters. Fixing the label would have left the gap.

Added `followerRefreshRebuildsPrecedenceFromTheShippedFiles`, which reproduces the follower without an HA harness: a second engine is held open over the same index while the first compacts, so it sees exactly what a follower sees — a new component registered, its inputs dropped — and its next query drives the reconcile. It asserts the follower lands on the leader's precedence reconstructed **from the manifests alone**, with the merged segment behind the tombstone despite owning the higher id.

Confirmed it bites rather than passing by construction: reverting only the refresh path's sort back to segment id turns it red while every other test stays green. That's the assertion that would have caught a follower-only regression, which the reopen test structurally could not.

**2. `docs/sparse-vector-storage-design.md` missing** — confirmed absent; the dangling reference is in a pre-existing javadoc block (`PaginatedSparseVectorEngine.java:93`, on memtable backpressure) that this PR doesn't touch. Leaving it rather than widening the diff — happy to file it separately if you'd like the reference either restored or dropped.

**3. Javadoc length** — noted, and I agree it's a real maintenance cost. It follows the file's existing house style, so I've kept it consistent rather than making this PR the odd one out.

Sparse-vector package now 118 tests / 0 failures.

lvca:
Addressed — `e201819a4`.

**1. `setRecencyEpoch`'s guard used `segmentId == 0L` as a proxy** — agreed, and worth fixing rather than noting. That guard exists specifically to stop a *silent* failure, so it shouldn't itself be defeatable by a caller passing `0`. Replaced with an explicit `segmentIdSet` flag, which says what it means and can't be argued out of by a future change to the id range.

**2. `JsonLImporterIT` newline** — re-checked `origin/main` this round: still missing the trailing newline, so `setup` still fails there and the fix is still load-bearing for this PR's CI. Keeping it; happy to split it out if you'd rather take it separately.

**3. Selector rewrite resolving the pre-existing `FUTURE` allocation note** — good catch, that was a side effect rather than a goal. The `HashMap` + per-tier `ArrayList`s allocated on every cascade tick are gone; the run scan is two flat arrays.

### CI status, for the record

Two lanes have gone red on this branch and neither is this change:

- **`opencypher-tck-tests` on `dda312eac`** — the TCK suite step itself **succeeded**; only the reporting steps failed ("no test report files found", "no `jacoco*.xml`"). No test failed. The commit adds nothing but one sparse-vector test method, and the TCK lane is routed by `-Dsurefire.includes` to the Cucumber suite, so it never runs it.
- **`integration-tests` on `ca08e5da4`** — `Remote transaction 'AS-…' not found or expired`, a session-expiry timing flake with no sparse-vector involvement.

`20cdfcfc9` went **fully green** across `unit-tests`, `slow-unit-tests`, `integration-tests` and `opencypher-tck-tests`, which covers every engine change in this PR bar the follower test and the flag above.

One thing worth flagging separately from this PR: `ParallelScanSafetyTest.parallelScanCompletesWhileQueryPoolIsSaturated` deadlocks reproducibly for me in a full-module local run — 0% CPU, and the thread dump shows `DedicatedThreadPool.runOnCaller` → `runRejectedTask`, after which the caller thread blocks on a `CountDownLatch` only a pool task can count down. It passes alone in 1.7s and passes immediately after this PR's test class, so it needs full-suite context and is not caused by this change (CI's `unit-tests` lane, which runs it, is green on `20cdfcfc9`). But under enough load it would present in CI as a lane timeout with no failing assertion, which is worth knowing about.

lvca:
Thanks — one implemented, two answered.

**2. `FINE` log names an array position** — agreed and fixed in `74856da80`. The position indexes an internal snapshot that stops meaning anything the moment the call returns, and this is precisely the line someone raises the level on when they're trying to work out why an index keeps growing. It now names the window by segment id, which an operator can match against the component files on disk and against the surrounding lines.

**1. `healthyShapeCeiling` derives its tier span from total postings** — deliberate, and I think it holds up, though the concern is a fair one to raise. Two reasons:

- Summing across all segments is the right proxy for *"how many tiers would a healthy index of this corpus span"*, which is what the ceiling models. A skewed distribution genuinely does span more tiers, so a looser ceiling there isn't an error — a corpus with one huge segment really does have room for more segments before its shape is degenerate.
- The case the observation worries about — many small segments inflating the sum — is the one case the guard never has to handle, because many small segments means many *same-tier* segments, and the primary trigger fires on the first four adjacent ones long before the guard is consulted. The guard only ever sees distributions where tiers are interleaved, i.e. where segments are of genuinely differing sizes.

Erring loose is also the safe direction for a backstop: firing late costs some extra segments on a pathological distribution, while firing eagerly would mean merging across size tiers and paying write amplification the primary policy exists to avoid.

**3. Redundant sort in `contiguousClosure`** — agreed, and thanks for confirming the tradeoff reads correctly from the outside.

**Nit on `JsonLImporterIT`** — right that it's a separate concern; carried only because `origin/main` still lacks the newline, so without it `setup` fails and every test lane on this PR reports `skipping` rather than running. Happy to split it out if preferred.

Sparse-vector package: 118 tests, 0 failures.

lvca:
Both implemented — `75d76c072` plus a PR-description and docs update.

**1. `contiguousClosure`'s WARNING still reported epochs** — correct, and a fair catch that I'd applied the reasoning inconsistently one commit earlier. It applies *more* strongly there than to the line I did fix: an epoch is an internal ordinal an operator has no way to see outside `segmentEpochs()`, and this one is a `WARNING`, so it's the line someone actually reads. Now names segment ids.

**2. Mixed-version HA rolling upgrade** — genuinely useful catch, and the one gap in the write-up that could have cost someone real time. The asymmetry is easy to miss: the corrected ordering is *recorded* in the segment file, but *reading* it correctly requires the fix, so during a staged rollout an old-build node that receives a segment merged by an upgraded leader falls back to ordering by `segmentId` and can still return a deleted document. The symptom is the same query answering differently depending on which node serves it, which reads like a replication bug rather than a version skew.

Added to both places someone would look:
- PR description, under the upgrade note.
- `how-to/operations/upgrade.adoc` in `arcadedb-docs`, in the `26.9.1` entry: upgrade every node *before* running `REBUILD INDEX`, and treat sparse-vector results from a mixed-version cluster as provisional.

Sparse-vector package: 118 tests, 0 failures.

lvca:
Answering the pre-merge question, plus the nit.

**The docs change is in a different repository, not dropped.** ArcadeDB's documentation lives in [`ArcadeData/arcadedb-docs`](https://github.com/ArcadeData/arcadedb-docs), not in this tree, so it can't appear in this diff — good instinct to check rather than take the description's word for it. Three commits are staged there awaiting @lvca's review (my standing instruction for that repo is to commit but never push):

- `5d0fe72a` — `concepts/vector-search.adoc`: a "Segment compaction and deletes" section explaining why segment order decides delete visibility.
- `2f0f52df` — `how-to/operations/upgrade.adoc`: a `26.9.1` entry under **Breaking changes**, in the same Impact/Action shape as the existing per-version entries.
- `bd1e3e40` — the mixed-version rolling-upgrade window, added after the previous review raised it.

I've reworded the PR description so it says "staged in `arcadedb-docs`" rather than implying the file is in this PR.

**`segmentIdSet` alignment** — fixed in `d1c6765ee`; all four declarations in that block now put `=` at the same column.

**`JsonLImporterIT` trailing newline** — intentional, and load-bearing rather than incidental. `origin/main` is missing the final newline (it landed that way in c816a66c7, #6468), which fails the `end-of-file-fixer` pre-commit hook, which fails the `setup` job, which makes every downstream test lane report `skipping` instead of running. I saw exactly that on this PR's first CI run — `setup fail`, 15 lanes skipped — and `setup` has passed on every run since. It's currently doing the same to every other PR open against main.

**On `@Tag`** — agreed, and thanks for checking. The whole class runs in ~0.3s; the largest fixture is 40 postings across a handful of segments, so it belongs in the default lane.

Sparse-vector package: 118 tests, 0 failures.

lvca:
The detection-durability finding was right, and it was the best catch of the whole review cycle. Fixed in `94d6ccb26`.

**The hole.** The detector tested a segment's *shape* — `parents > 0 && epoch == id`. That identifies a legacy merge exactly, but only while the shape survives. A post-upgrade compaction that consumes it produces a merge with a freshly allocated id, so `epoch != id`, the shape stops matching, and the suspect ordering rides along inside the result unchanged. And since `flush()` cascades `compactSizeTiered()` on every call, that can land within the first few writes after an upgrade — silencing the warning before anyone reads it, permanently, on this and every later open. Exactly the state the warning existed to prevent.

**The fix.** Replaced the derived test with a persisted property that merges inherit: a segment is untrusted if its epoch slot was never written (a pre-fix merge) or if any of its inputs was, and `compactInputs` ORs it across the input set. Doubt is now propagated rather than re-guessed, so absorbing a legacy segment can't launder it.

**On where to put the flag** — this is the part that took some thought. There's no manifest slot left: both former "reserved" longs are now the tombstone count and the epoch, and a third would move the manifest size and therefore the format version, costing the backward compatibility this PR has been careful to keep. So the flag rides in **bit 62 of the epoch itself**. Epochs are ids from a counter that starts at 1 and is only ever raised, so bit 62 is unreachable by a real value (an index would need to seal 4.6 quintillion segments), and readers mask it off before comparing, so it can't perturb the ordering it travels with. No format change, no version bump.

**The test drives the real sequence** rather than asserting the property directly: seal a legacy-shaped segment, let the size-tiered trigger absorb it, reopen, require the warning. I verified it isolates the reported gap — stubbing the inheritance to `false` fails that test and only that test.

One fixture detail worth noting: the legacy tests now write `0` into the slot explicitly. `setSegmentId` defaults the epoch to the id, so without that they were *resembling* the pre-fix shape rather than reproducing it — and would have passed against the new detector for the wrong reason.

Also from this review:
- **`contiguousClosure` duplicate handling** — now reports a repeated input as what it is, rather than as a segment missing from the live snapshot. The count is compared against distinct ids, so the two causes can't be confused.
- **Global logger swap** — documented the dependency on the engine module running test classes sequentially (`forkCount=1`, `reuseForks=true`, no JUnit parallel config), and noted that these assertions would become flaky if class-level parallelism were ever enabled here.

Sparse-vector package: 119 tests, 0 failures.

lvca:
All three addressed — one code change, one follow-up filed, one confirmed.

**1. Follower doesn't log the warning** — you asked me to confirm the operational story; on looking at it I'd rather close the gap than document it. A follower that receives an untrusted segment *by replication* (rather than finding one on disk at open) is serving the same suspect ordering and said nothing, leaving only the leader's log to go on — on the node that isn't the one answering the query. `56ec3ebbd` raises it from `refreshSegmentsFromFileManager` too, latched to once per engine instance with an `AtomicBoolean`, which was the reason it was originally left out of that path.

**3. Queryable admin surface** — agreed, and filed as #6566. Writing it up sharpened why the log line alone isn't enough: it's latched to once per engine instance, so a node that's been up for a week says nothing and there's no way to *re-ask* the question; nothing can alert on it; there's no positive confirmation that `REBUILD INDEX` worked; and during a rolling upgrade the answer is per-node, which is exactly when you want to query it rather than grep for it. `epochUntrusted()` already carries the value, so it's plumbing rather than new bookkeeping. Out of scope here.

**2. `CapturingLogger` parallelism risk** — confirmed and left as documented. The comment names the specific Surefire settings it depends on (`forkCount=1`, `reuseForks=true`, no JUnit parallel config) so whoever reorganizes the lanes has the tripwire in front of them.

**On the `JsonLImporterIT` scope note** — fair point about someone skimming later. It's a separate commit with its own message explaining the CI dependency, the PR title is scoped to #6379, and the description doesn't fold it into the fix. `origin/main` still lacks the newline, so it's still doing its job here.

Sparse-vector package: 119 tests, 0 failures.

lvca:
Both handled — `f75a14eb8`.

**1. `parentSegmentCount()` is dead code with a stale javadoc** — correct on both counts, and a good catch. I added it for the shape-based detector and orphaned it when the persisted flag replaced that, leaving a javadoc that claimed the open-time scan called it. `grep` confirms only its own declaration. Removed rather than re-documented: `parentSegments()` already covers the one real caller shape, and unused public API on a storage reader is a liability, not an affordance.

**2. `segmentIds()` order** — the premise doesn't quite hold: it still does `Arrays.sort(out)`, so its order is unchanged and no caller can have been broken by this PR. But the question points at something sharper that I'd missed, so thank you for it: `segmentIds()` *sorts* while `segmentEpochs()` *preserves precedence order*, and since #6379 those two orders genuinely differ — a merged segment's epoch sits below its own id. Zipping the two results therefore pairs an id with some other segment's epoch. Both javadocs now say they are not positionally related.

**3. Javadoc duplication across `RECENCY_ORDER`, the class doc, and `compactSizeTiered`** — leaving as-is, with the reasoning you'd expect. The three say different things at different altitudes: the class doc is the policy, `compactSizeTiered` is what the trigger does and what adjacency costs it, and `RECENCY_ORDER` is why the comparator key is the epoch and not the id. Someone reading `replaceSegments` and wondering why it sorts is going to land on `RECENCY_ORDER`, not on the class header. Overlap in the *rationale* is the price of the "why" being present where the code is.

**On CI** — noted that you were reviewing statically. Every commit on this branch has run the full lane set. `20cdfcfc9` was fully green across `unit-tests`, `slow-unit-tests`, `integration-tests` and `opencypher-tck-tests`; the two red lanes seen since were a TCK *reporting*-step failure (the suite itself passed) and a `Remote transaction not found or expired` flake in integration-tests, neither touching sparse vectors. Locally the sparse-vector package is 119 tests / 0 failures on every commit, and I've confirmed the load-bearing new tests fail when their fix is stubbed out rather than passing by construction.

lvca:
Fixed in `5cf98204f` — the double blank line was left behind by removing `parentSegmentCount()` in the previous commit.

That was the last outstanding point. Recapping where this landed, since the branch has grown a long review trail:

**The fix** (three parts, all needed):
1. a persisted **recency epoch** decoupled from the segment id, inherited by a merge from its newest input;
2. **adjacency** — partial merges restricted to a contiguous run in epoch order, since epoch inheritance alone still lets a posting from the oldest input outrank whatever sat in a gap;
3. a **shape guard**, because adjacency introduces a way for compaction to stall on interleaved segment sizes.

**What review added on top**, in rough order of value:
- the pre-fix **detection surviving absorption** by a later compaction — the shape-based check would have gone quiet within the first few writes after an upgrade, silently;
- the **follower-refresh test**, replacing a claim in my own description that turned out to be untested;
- the **HA rolling-upgrade caveat** — an old-build node reading a new-build leader's merged segment still applies id ordering;
- the same warning now reaching **followers** rather than only the leader;
- `contiguousClosure`'s widening branch **pinned by a test** rather than asserted in a comment;
- the builder's setter-order guard, an unreachable clamp removed, two log lines made operator-actionable, and a dead accessor dropped.

**Verification:** sparse-vector package 119 tests / 0 failures on every commit; `20cdfcfc9` fully green across `unit-tests`, `slow-unit-tests`, `integration-tests` and `opencypher-tck-tests`. Each load-bearing new test was confirmed to fail when its fix is stubbed out, rather than passing by construction.

**Follow-ups filed:** #6566 (surface `epochUntrusted` through a queryable admin surface rather than only a log line).

**Docs** staged in `ArcadeData/arcadedb-docs` for @lvca: `concepts/vector-search.adoc` (segment compaction and deletes) and `how-to/operations/upgrade.adoc` (26.9.1 breaking-change entry plus the mixed-version window).

PR #6571, opened 2026-08-22
fix(engine): native select must not cap the index candidate scan by the result limit

### LINKED ISSUE #6565: Native select applies limit to the index candidate scan: skip/limit paging and IS NOT NULL silently lose rows
**Version:** reproduced on 26.7.3 and on 26.8.1 (latest release); root cause still present on `main` at `86c4338`
**Environment:** JDK 25 (Temurin), Windows 11 x64, embedded `DatabaseFactory`, no server
**Component:** `com.arcadedb.query.select` — native select builder

Tagging: @robfrank @lvca 

### Summary

`SelectExecutor.lookForIndexes()` hands the query's **result** `limit` to `MultiIndexCursor`, which applies it to **candidates** read from the index. Those candidates are then reduced twice more downstream — by `evaluateWhere()`, and by `skip` — so the cap is spent before the filtering it was supposed to survive. The query returns fewer rows than match, with no error and no warning.

Two query shapes are affected, and the second one is ordinary paging.

**A. A null check combined with another condition.** The index covers only the other condition, so the null check runs on whatever candidates survived the cap.

```java
db.select().fromType("T").where()
  .property("a").eq().value("x")
  .and().property("b").isNotNull()
  .limit(500).documents();           // 200 documents match; returns 101
```

**B. `skip` + `limit` on a plain indexed equality — no null check anywhere.** The candidate cap is `limit`, but reaching row `skip+1` needs at least `skip + limit` candidates, so the result is short by `skip`, and empty whenever `skip >= limit`.

```java
db.select().fromType("T").where()
  .property("a").eq().value("x")
  .skip(100).limit(50).documents();  // 1000 documents match; returns 0
```

`.count()` on that same query returns `0` as well.

The same queries through the SQL engine are all correct, so this is an inconsistency between two of ArcadeDB's own APIs rather than a limitation of the data or the schema.

### Reproducer

A single self-contained [jbang](https://www.jbang.dev) file — no build, no server, embedded database in a temp directory, pulls `arcadedb-engine` itself. Save as `ArcadeDbLimitBug.java` and run `jbang ArcadeDbLimitBug.java`. It exits 1 while the bug is present.

<details>
<summary><code>ArcadeDbLimitBug.java</code></summary>

```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS com.arcadedb:arcadedb-engine:26.8.1
//JAVA 21+

import com.arcadedb.schema.Schema;
import com.arcadedb.schema.Type;

import java.nio.file.Files;
import java.util.Iterator;

class ArcadeDbLimitBug {

    static final int ROWS = 1000, MATCHES = 200;
    static boolean reproduced = false;

    public static void main(String[] args) throws Exception {
        var dir = Files.createTempDirectory("arcadedb-limit-bug");
        try (var db = new DatabaseFactory(dir.resolve("db").toString()).create()) {
            seed(db);

            System.out.println("\nArcadeDB native select vs SQL - " + ROWS + " rows, all with a='x', exactly "
                    + MATCHES + " with b set\n");
            System.out.println("A. a null check combined with another condition, with a limit");
            System.out.printf("%-46s %-10s %-10s %s%n", "query", "native", "sql", "expected");

            for (var limit : new int[]{50, 100, 200, 500, 1000}) {
                var expected = Math.min(limit, MATCHES);
                check("a='x' AND b IS NOT NULL   limit " + limit,
                        nativeEqNotNull(db, limit),
                        sql(db, "select from T where a = 'x' and b is not null limit " + limit),
                        expected);
            }
            check("a='x' AND b IS NOT NULL   no limit",
                    nativeEqNotNull(db, -1), sql(db, "select from T where a = 'x' and b is not null"), MATCHES);

            System.out.println("\nB. skip + limit on a plain indexed equality - no null check anywhere");
            System.out.printf("%-46s %-10s %-10s %s%n", "query", "native", "sql", "expected");
            page(db, 0, 50, 50);
            page(db, 10, 50, 50);
            page(db, 100, 50, 50);
            page(db, 500, 50, 50);
            page(db, 0, 1000, 1000);
            page(db, 100, 1000, 900);
            checkCount(db);

            System.out.println("\nC. controls - every other shape pages correctly");
            System.out.printf("%-46s %-10s %-10s %s%n", "query", "native", "sql", "expected");
            check("b IS NOT NULL alone        limit 500",
                    count(db.select().fromType("T").where().property("b").isNotNull().limit(500).documents()),
                    sql(db, "select from T where b is not null limit 500"), MATCHES);
            check("a='x' AND n<200            limit 500",
                    count(db.select().fromType("T").where().property("a").eq().value("x")
                            .and().property("n").lt().value(200).limit(500).documents()),
                    sql(db, "select from T where a = 'x' and n < 200 limit 500"), MATCHES);
            check("b='set' alone              limit 500",
                    count(db.select().fromType("T").where().property("b").eq().value("set").limit(500).documents()),
                    sql(db, "select from T where b = 'set' limit 500"), MATCHES);

            System.out.println(reproduced
                    ? "\nBUG REPRODUCED: the limit is spent on index candidates, before the where clause and the skip."
                    : "\nNot reproduced on this build.");
        }
        if (reproduced) System.exit(1);
    }

    static void seed(Database db) {
        db.transaction(() -> {
            var t = db.getSchema().createDocumentType("T");
            t.createProperty("a", Type.STRING);
            t.createProperty("b", Type.STRING);
            t.createProperty("n", Type.INTEGER);
            t.createTypeIndex(Schema.INDEX_TYPE.LSM_TREE, false, "a");
            t.createTypeIndex(Schema.INDEX_TYPE.LSM_TREE, false, "b");
            t.createTypeIndex(Schema.INDEX_TYPE.LSM_TREE, false, "n");
        });
        db.transaction(() -> {
            for (var i = 0; i < ROWS; i++) {
                var d = db.newDocument("T");
                d.set("a", "x");
                d.set("n", i);
                if (i % (ROWS / MATCHES) == 0) d.set("b", "set");
                d.save();
            }
        });
    }

    static int nativeEqNotNull(Database db, int limit) {
        var after = db.select().fromType("T").where()
                .property("a").eq().value("x").and().property("b").isNotNull();
        return count(limit < 0 ? after.documents() : after.limit(limit).documents());
    }

    static void page(Database db, int skip, int limit, int expected) {
        check("a='x'   skip " + skip + " limit " + limit,
                count(db.select().fromType("T").where().property("a").eq().value("x")
                        .skip(skip).limit(limit).documents()),
                sql(db, "select from T where a = 'x' skip " + skip + " limit " + limit),
                expected);
    }

    static void checkCount(Database db) {
        var nativeCount = db.select().fromType("T").where().property("a").eq().value("x").skip(100).limit(50).count();
        if (nativeCount != 50) reproduced = true;
        System.out.printf("%-46s %-10s %-10s %-6d %s%n", "a='x'   skip 100 limit 50   .count()",
                nativeCount, 50, 50, nativeCount != 50 ? "  <-- WRONG" : "");
    }

    static int sql(Database db, String query) {
        var n = 0;
        try (var rs = db.query("sql", query)) {
            while (rs.hasNext()) { rs.next(); n++; }
        }
        return n;
    }

    static int count(Iterator<?> it) {
        var n = 0;
        while (it.hasNext()) { it.next(); n++; }
        return n;
    }

    static void check(String label, int nativeCount, int sqlCount, int expected) {
        var bad = nativeCount != expected;
        if (bad) reproduced = true;
        System.out.printf("%-46s %-10s %-10s %-6d %s%n", label, nativeCount, sqlCount, expected,
                bad ? "  <-- WRONG" : "");
    }
}
```

</details>

### Actual vs expected

1000 documents, all with `a='x'`, exactly 200 with `b` set; non-unique LSM_TREE indexes on `a`, `b` and `n`.

```
A. a null check combined with another condition, with a limit
query                                          native     sql        expected
a='x' AND b IS NOT NULL   limit 50             11         50         50       <-- WRONG
a='x' AND b IS NOT NULL   limit 100            21         100        100      <-- WRONG
a='x' AND b IS NOT NULL   limit 200            41         200        200      <-- WRONG
a='x' AND b IS NOT NULL   limit 500            101        200        200      <-- WRONG
a='x' AND b IS NOT NULL   limit 1000           200        200        200
a='x' AND b IS NOT NULL   no limit             200        200        200

B. skip + limit on a plain indexed equality - no null check anywhere
query                                          native     sql        expected
a='x'   skip 0 limit 50                        50         50         50
a='x'   skip 10 limit 50                       41         50         50       <-- WRONG
a='x'   skip 100 limit 50                      0          50         50       <-- WRONG
a='x'   skip 500 limit 50                      0          50         50       <-- WRONG
a='x'   skip 0 limit 1000                      1000       1000       1000
a='x'   skip 100 limit 1000                    900        900        900
a='x'   skip 100 limit 50   .count()           0          50         50       <-- WRONG

C. controls - every other shape pages correctly
query                                          native     sql        expected
b IS NOT NULL alone        limit 500           200        200        200
a='x' AND n<200            limit 500           200        200        200
b='set' alone              limit 500           200        200        200
```

Every wrong number falls out of the arithmetic below, which is what makes us think this is one defect rather than two: in A the candidate cap is `limit + 1` and one candidate in five carries `b`, giving `limit/5 + 1`; in B the cap is `limit + 1` candidates of which `skip` are consumed, giving `max(0, limit + 1 - skip)`.

### Root cause

[`SelectExecutor.lookForIndexes()`](https://github.com/ArcadeData/arcadedb/blob/86c4338fa1f5fe8c07607f1d61c469e7ed789cdc/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java#L207-L218) passes `select.limit` straight into the cursor:

```java
private MultiIndexCursor lookForIndexes() {
  if (select.fromType != null && select.rootTreeElement != null) {
    final List<IndexCursor> cursors = new ArrayList<>();

    // FIND AVAILABLE INDEXES
    boolean canUseIndexes = isTheNodeFullyIndexed(select.rootTreeElement);   // <-- computed, never read

    filterWithIndexes(select.rootTreeElement, cursors);
    if (!cursors.isEmpty())
      return new MultiIndexCursor(cursors, select.limit, true);              // <-- a RESULT limit used as a CANDIDATE limit
  }
  return null;
}
```

[`MultiIndexCursor.hasNext()`](https://github.com/ArcadeData/arcadedb/blob/86c4338fa1f5fe8c07607f1d61c469e7ed789cdc/engine/src/main/java/com/arcadedb/index/MultiIndexCursor.java#L112-L116) then stops the scan after `limit + 1` candidates:

```java
if (limit > -1 && browsed > limit)
  return false;
```

Downstream, [`SelectIterator`](https://github.com/ArcadeData/arcadedb/blob/86c4338fa1f5fe8c07607f1d61c469e7ed789cdc/engine/src/main/java/com/arcadedb/query/select/SelectIterator.java#L66-L90) is correct on its own terms — its constructor consumes the first `skip` records, its `hasNext()` caps at `skip + limit`, and `fetchNext()` drops anything failing `evaluateWhere()`. That is exactly why the candidate cap is unsound: a candidate cap of `select.limit` is only correct when the index cursors match the predicate exactly **and** `skip == 0`, and neither condition is tested.

The unused `canUseIndexes` local looks like the intended guard, but as written it would not be sufficient either — `isTheNodeFullyIndexed()` returns `true` for `a='x' AND b IS NOT NULL`, because for an `and` node it returns `leftIsIndexed || rightIsIndexed`, i.e. "an index is usable here" rather than "the predicate is fully covered".

### Suggested fix

Cap the candidate scan only where it is provably safe, and account for `skip`:

- pass `-1` (unbounded) whenever the index cursors do not cover the whole where-tree, i.e. whenever any conjunct will still be re-checked by `evaluateWhere()` — which includes every `is_null` / `is_not_null` node;
- otherwise pass `select.skip + select.limit` rather than `select.limit`;
- separately, `browsed > limit` in `MultiIndexCursor.hasNext()` admits one candidate more than asked for, which is where the `+1` in the numbers above comes from.

We would be glad to open a PR along those lines if that is welcome — tell us which shape you would prefer for the "fully covered" test and we will follow it.

### Why it matters

The failure is silent and both shapes are common. A caller gets a plausible row count and a green build. In our case a listing built as `eq(...) + isNotNull(...) + limit(n)` reported 111 attachments where 767 existed, and zero for a container holding thirty; the numbers were wrong but believable, so the write path got suspected first. Shape B is worse in one respect: a paged listing over an indexed equality returns nothing at all from page two onward once `skip >= limit`, and the same query shape driving a delete would remove the wrong subset.

### Workaround

Use `db.query("sql", ...)` for these shapes, or drop `skip`/`limit` from the native query and bound the result in the caller.

lvca:
## Summary

Fixes #6565.

`SelectExecutor.lookForIndexes()` handed the query's *result* `limit` straight to `MultiIndexCursor` as a *candidate*
cap on the index scan. Downstream, `evaluateWhere()` and `skip` still reduce that candidate stream further, so the
cap was spent before the filtering it was meant to survive - silently returning fewer rows than actually match, with
no error.

Two shapes were affected:
- An indexed condition combined via `AND` with a non-indexable condition such as `IS NOT NULL`: only one `AND` child
  ever keeps its index cursor (`filterWithIndexesFinalNode`), so the other conjunct is re-checked later by
  `evaluateWhere()` against a stream that was already capped too early.
- Ordinary `skip` + `limit` paging over an indexed equality: reaching row `skip+1` needs `skip + limit` candidates,
  but the cap was just `limit`, so the result went empty whenever `skip >= limit`.

The same queries through the SQL engine were already correct, so this was purely an inconsistency between the two
query APIs.

## Fix

- The candidate cap is now applied only when the index scan **exactly reproduces** the where-tree's result set: a
  bare indexed leaf, or an `OR` of such leaves (the union shape `MultiIndexCursor` already merges). Any `AND`, `NOT`,
  or `IS [NOT] NULL` disqualifies the cap and the scan runs uncapped, exactly matching what the SQL engine already
  does for the same shapes - `evaluateWhere()`/`skip` downstream stay correct regardless, this only removes a stale
  optimization that wasn't sound.
- When the cap does apply, it is now `skip + limit`, not just `limit`.
- The nested per-value cursor `filterWithIndexesFinalNode()` builds for an `IN` leaf now shares that same corrected
  cap instead of a raw `select.limit` that was missing `skip` - same bug, one call site over.
- `MultiIndexCursor.hasNext()` also admitted one candidate more than asked for (`browsed > limit` instead of
  `browsed >= limit`). The two call sites that pass a real (non -1) limit are both in `SelectExecutor`, so tightening
  this doesn't affect any of the unbounded (`-1`) callers elsewhere in the codebase.

## Test plan

- New regression test `Issue6565SelectIndexCandidateLimitTest` (7 cases): the two shapes from the issue (`AND` +
  `IS NOT NULL` with various limits, `skip`/`limit` paging including `skip >= limit`), plus `.count()` variants, an
  `OR` of two indexed ranges with `skip`/`limit`, and an `IN` leaf with `skip`/`limit` (covering the nested cursor
  fix). All 7 fail on `main` and pass with this change; confirmed by reverting the fix locally and re-running.
- Ran the full `com.arcadedb.query.select` and `com.arcadedb.index` packages (222 classes / 1205 tests): 0 failures,
  0 errors - the entire blast radius of the change (nothing outside these two packages references
  `SelectExecutor`/`MultiIndexCursor`).

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/MultiIndexCursor.java
+++ b/engine/src/main/java/com/arcadedb/index/MultiIndexCursor.java
@@ -111,7 +111,9 @@ public class MultiIndexCursor implements IndexCursor {
 
   @Override
   public boolean hasNext() {
-    if (limit > -1 && browsed > limit)
+    // #6565: browsed COUNTS CANDIDATES ALREADY RETURNED BY next(), SO browsed == limit MEANS limit CANDIDATES HAVE
+    // ALREADY BEEN HANDED OUT - ONE MORE WOULD EXCEED THE CAP THE CALLER ASKED FOR
+    if (limit > -1 && browsed >= limit)
       return false;
 
     for (int i = 0; i < cursors.size(); ++i) {
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -329,6 +329,9 @@ public class SelectExecutor {
     if (leaf == null || select.orderBy.size() != 1)
       return false;
     final Pair<String, Boolean> orderBy = select.orderBy.getFirst();
+    // UNCHECKED CAST IS SAFE: leaf CAME FROM soleExactLeaf(), WHICH ONLY EVER RETURNS A NODE WHOSE node.index IS
+    // NON-null - AND isTheNodeFullyIndexed() ONLY EVER SETS node.index AFTER THIS EXACT SAME CAST ON node.left
+    // ALREADY SUCCEEDED
     return orderBy.getSecond() && orderBy.getFirst().equals(((SelectPropertyValue) leaf.left).propertyName);
   }
 
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -232,6 +232,11 @@ public class SelectExecutor {
       filterWithIndexes(select.rootTreeElement, cursors);
       if (!cursors.isEmpty())
         return new MultiIndexCursor(cursors, indexCandidateLimit, true);
+
+      // NO CURSOR WAS ACTUALLY BUILT (E.G. A BARE neq/like/ilike LEAF: "INDEXED" PER isTheNodeFullyIndexed()'S LOOSER
+      // CHECK - SEE #6577 - BUT filterWithIndexesFinalNode()'S switch DOESN'T HANDLE THE OPERATOR), SO NO CAP WAS
+      // EVER APPLIED EITHER: RESET THE TEST-VISIBLE FIELD RATHER THAN LEAVE IT HOLDING A MISLEADING FINITE VALUE
+      indexCandidateLimit = -1;
     }
     return null;
   }
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -254,9 +254,11 @@ public class SelectExecutor {
    * True when the index cursor(s) built for this subtree return exactly the records matching it, with nothing left
    * for {@code evaluateWhere()} to discard afterward. Only a bare indexed leaf qualifies (through the synthetic
    * {@code run} wrapper, see below): an {@code is_null}/{@code is_not_null} leaf never gets a cursor, and under an
-   * {@code and} only one child's cursor is kept (see {@link #filterWithIndexesFinalNode}), so the other conjunct is
-   * still checked later against a stream a cap would have already truncated. {@code not} is conservatively treated
-   * the same way.
+   * {@code and} at most one child's cursor is kept when either side's index is unique (see
+   * {@link #filterWithIndexesFinalNode}) - when neither side is unique, both children's cursors survive and
+   * {@link MultiIndexCursor} unions rather than intersects them, a superset {@code evaluateWhere()} must still
+   * narrow down. Either way the discarded or extra conjunct is still checked later against a stream a cap would
+   * have already truncated or under-counted. {@code not} is conservatively treated the same way.
    * <p>
    * {@code or} is conservatively excluded too, even though {@link MultiIndexCursor} merges its children's cursors
    * into the same shape as the boolean union: that merge is a plain k-way merge with no RID dedup, so two branches
@@ -281,6 +283,11 @@ public class SelectExecutor {
     if (node == null)
       return true;
 
+    // node.index != null MEANS "isTheNodeFullyIndexed() FOUND AN INDEX ON THIS LEAF'S PROPERTY", NOT "THIS LEAF'S
+    // OPERATOR ACTUALLY PRODUCES A CURSOR" - IT'S SET FOR neq/like/ilike TOO (SEE #6577). HARMLESS TODAY BECAUSE A
+    // BARE SUCH LEAF STILL PRODUCES NO CURSOR IN filterWithIndexesFinalNode()'S switch, SO cursors STAYS EMPTY AND
+    // lookForIndexes() RETURNS null BEFORE ANY (WRONGLY-COMPUTED) CAP IS EVER USED - BUT #6577 IS WHERE TO FIX THIS
+    // AT THE SOURCE, NOT HERE
     if (!(node.left instanceof SelectTreeNode))
       return node.index != null;
 
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -47,7 +47,7 @@ public class SelectExecutor {
   // #6565: THE CANDIDATE CAP lookForIndexes() COMPUTED FOR THE WHOLE where-TREE, REUSED BY filterWithIndexesFinalNode()
   // FOR THE NESTED PER-VALUE CURSOR IT BUILDS FOR AN in_op LEAF, SO NEITHER PLACE HANDS A RAW select.limit (MISSING
   // skip) TO A MultiIndexCursor
-  private int indexCandidateLimit = -1;
+  int indexCandidateLimit = -1;
 
   static class IndexInfo {
     public final Index   index;
@@ -208,7 +208,10 @@ public class SelectExecutor {
     return iterator;
   }
 
-  private MultiIndexCursor lookForIndexes() {
+  // PACKAGE-PRIVATE (NOT private) SO Issue6565SelectIndexCandidateLimitTest CAN VERIFY THE COMPUTED CAP DIRECTLY:
+  // A RESULT-COUNT ASSERTION CANNOT TELL A CORRECTLY APPLIED CAP APART FROM THE UNCAPPED FALLBACK, SINCE THE
+  // LAZY-PULL CONSUMERS (SelectIterator, executeCount()) ALREADY STOP AT skip + limit ON THEIR OWN EITHER WAY
+  MultiIndexCursor lookForIndexes() {
     if (select.fromType != null && select.rootTreeElement != null) {
       final List<IndexCursor> cursors = new ArrayList<>();
 
@@ -246,6 +249,12 @@ public class SelectExecutor {
    * {@code is_not_null} leaf never gets a cursor, and under an {@code and} only one child's cursor is kept (see
    * {@link #filterWithIndexesFinalNode}), so the other conjunct is still checked later against a stream a cap would
    * have already truncated. {@code not} is conservatively treated the same way.
+   * <p>
+   * {@code Select.compile()} always finalizes the tree with a trailing {@code setLogic(SelectOperator.run)}
+   * ({@link Select#compile}); when the where-clause is a single bare condition with no {@code and()}/{@code or()},
+   * that call runs its "1ST TIME ONLY" branch and wraps the leaf in a synthetic {@code run} node whose {@code right}
+   * is always {@code null} ({@link Select#setLogic}) - {@code run} is otherwise never used as a tree operator, so it
+   * is treated here as a transparent pass-through to its {@code left} child.
    */
   private boolean isWhereExactlyIndexed(final SelectTreeNode node) {
     if (node == null)
@@ -254,7 +263,7 @@ public class SelectExecutor {
     if (!(node.left instanceof SelectTreeNode))
       return node.index != null;
 
-    if (node.operator == SelectOperator.or)
+    if (node.operator == SelectOperator.or || node.operator == SelectOperator.run)
       return isWhereExactlyIndexed((SelectTreeNode) node.left) && isWhereExactlyIndexed((SelectTreeNode) node.right);
 
     return false;
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -252,11 +252,24 @@ public class SelectExecutor {
 
   /**
    * True when the index cursor(s) built for this subtree return exactly the records matching it, with nothing left
-   * for {@code evaluateWhere()} to discard afterward. Only a bare indexed leaf, or an {@code or} of such leaves
-   * (which {@link MultiIndexCursor} merges - the same shape as the boolean union), qualifies: an {@code is_null}/
-   * {@code is_not_null} leaf never gets a cursor, and under an {@code and} only one child's cursor is kept (see
-   * {@link #filterWithIndexesFinalNode}), so the other conjunct is still checked later against a stream a cap would
-   * have already truncated. {@code not} is conservatively treated the same way.
+   * for {@code evaluateWhere()} to discard afterward. Only a bare indexed leaf qualifies (through the synthetic
+   * {@code run} wrapper, see below): an {@code is_null}/{@code is_not_null} leaf never gets a cursor, and under an
+   * {@code and} only one child's cursor is kept (see {@link #filterWithIndexesFinalNode}), so the other conjunct is
+   * still checked later against a stream a cap would have already truncated. {@code not} is conservatively treated
+   * the same way.
+   * <p>
+   * {@code or} is conservatively excluded too, even though {@link MultiIndexCursor} merges its children's cursors
+   * into the same shape as the boolean union: that merge is a plain k-way merge with no RID dedup, so two branches
+   * whose match sets overlap (same-property ranges that overlap, or two different properties a single record can
+   * both satisfy - not provable disjoint from the where-tree's shape alone) make the same record surface as two
+   * separate candidates, each burning one unit of the cap before {@link SelectIterator}/{@code executeCount()}'s
+   * downstream {@code filterOutRecords} dedup ever sees it. That can exhaust the cap on duplicates before enough
+   * distinct matches are found - the same "cap spent before the filtering it needs to survive" defect this class
+   * exists to fix, just reachable again through the "or is exact" path. An {@code in_op} leaf's own internal
+   * per-value cursors don't have this problem (each value is a distinct key on the same single-valued property, so
+   * their RID sets are disjoint by construction) - but that dedup-free merge happens once, inside
+   * {@link #filterWithIndexesFinalNode}'s own {@code in_op} handling, not through a top-level {@code or}, so
+   * {@code in_op} is unaffected by excluding {@code or} here.
    * <p>
    * {@code Select.compile()} always finalizes the tree with a trailing {@code setLogic(SelectOperator.run)}
    * ({@link Select#compile}); when the where-clause is a single bare condition with no {@code and()}/{@code or()},
@@ -271,7 +284,7 @@ public class SelectExecutor {
     if (!(node.left instanceof SelectTreeNode))
       return node.index != null;
 
-    if (node.operator == SelectOperator.or || node.operator == SelectOperator.run)
+    if (node.operator == SelectOperator.run)
       return isWhereExactlyIndexed((SelectTreeNode) node.left) && isWhereExactlyIndexed((SelectTreeNode) node.right);
 
     return false;
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -222,12 +222,11 @@ public class SelectExecutor {
       // #6565: A CANDIDATE CAP IS SAFE ONLY WHEN THE INDEX SCAN EXACTLY REPRODUCES THE WHERE-TREE'S RESULT SET AND
       // THE ORDER BY (IF ANY) IS ALREADY SATISFIED BY IT - evaluateWhere(), skip AND fetchResultInCaseOfOrderBy()'s
       // FULL-DRAIN SORT ALL REDUCE THE STREAM FURTHER OTHERWISE, SO THE SCAN MUST RUN UNCAPPED TO SURVIVE THAT.
-      // MUST RUN BEFORE filterWithIndexes() BELOW, SINCE isWhereExactlyIndexed() READS node.index BEFORE
+      // MUST RUN BEFORE filterWithIndexes() BELOW, SINCE soleExactLeaf() READS node.index BEFORE
       // filterWithIndexesFinalNode() PRUNES AN 'and' SIBLING'S - HARMLESS TODAY ONLY BECAUSE 'and' IS ALREADY
       // UNCONDITIONALLY DISQUALIFIED
-      final boolean whereExact = isWhereExactlyIndexed(select.rootTreeElement);
-      indexCandidateLimit = whereExact && isOrderBySafeForCap(soleExactLeaf(select.rootTreeElement))
-          ? computeExactCandidateLimit() : -1;
+      final SelectTreeNode exactLeaf = soleExactLeaf(select.rootTreeElement);
+      indexCandidateLimit = exactLeaf != null && isOrderBySafeForCap(exactLeaf) ? computeExactCandidateLimit() : -1;
 
       filterWithIndexes(select.rootTreeElement, cursors);
       if (!cursors.isEmpty())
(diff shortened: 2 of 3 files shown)


lvca:
Good catch, thanks - confirmed the tree shape via \`Select.setLogic()\`'s "1ST TIME ONLY" branch and \`grep\`-verified \`SelectOperator.run\` has no other call sites. Fixed in 8e6dd6e2: \`isWhereExactlyIndexed()\` now treats \`run\` as a transparent pass-through, same as \`or\`.

Since a row-count assertion can't distinguish "cap correctly computed" from "cap stayed at -1 but the lazy-pull consumers bounded it anyway" (as you pointed out for 4 of the 7 original tests), I widened \`lookForIndexes()\`/\`indexCandidateLimit\` to package-private and added two new tests (\`singleBareEqualityCandidateCapCoversSkipPlusLimit\`, \`singleBareInCandidateCapCoversSkipPlusLimit\`) that assert on the computed \`indexCandidateLimit\` directly. Both fail with \`-1\` instead of the expected \`skip+limit\` on the pre-fix code, confirmed by reverting the one-line fix locally and re-running.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages (222 classes / 1207 tests) still green.

lvca:
Added the two suggested cases in 5c973a2d, both asserting on \`indexCandidateLimit\` directly (\`-1\`), same rationale as before:

- \`andUnderOrIsNeverTreatedAsExactlyIndexed\`: \`(a='x' AND b='set') OR n=5\`. Traced \`Select.setLogic()\`'s precedence handling to confirm this fluent chain actually builds \`or(and(a,b), n)\` (and's precedence 2 > or's 0), so it's a real end-to-end test through the public builder, not synthetic.
- \`notLeafIsNeverTreatedAsExactlyIndexed\`: \`SelectOperator.not\` has no fluent entry point on \`Select\` today, so this one builds the tree by hand with \`SelectTreeNode\`'s public constructor.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1209 tests).

lvca:
Thanks for the thorough second pass. Addressed the three cheap documentation suggestions in e43242d2:

- Added a comment on the \`indexCandidateLimit\` computation noting it must run before \`filterWithIndexes()\`, and why that ordering isn't currently load-bearing but would become so if \`and\` disqualification is ever loosened.
- Added a comment on \`computeExactCandidateLimit()\`'s overflow guard explaining the deliberate \`-1\` fallback instead of clamping to \`Integer.MAX_VALUE\`.
- Added a comment at the \`IN\`-leaf nested cursor call site explaining why sharing the whole tree's \`indexCandidateLimit\` is a proven-safe over-approximation, not just convenient reuse.

For the out-of-scope \`NOT\` observation: agreed it's worth tracking, opened #6575. No code change here since it's unreachable via the fluent builder today and \`evaluateWhere()\` already guarantees correctness regardless (it would just be a latent efficiency issue if/when \`NOT\` becomes fluently reachable).

All 11 regression tests + full \`com.arcadedb.query.select\`/\`com.arcadedb.index\` packages still green.

lvca:
Good catch, and a real gap - confirmed by reproducing \`WHERE n > -1 ORDER BY n DESC LIMIT 10\` before fixing: got \`n=9..0\` instead of the correct \`n=999..990\`, exactly as predicted.

Fixed in 9fbedca9: \`soleExactLeaf()\` finds the tree's one and only indexed leaf (\`null\` for anything but a bare leaf, since an \`or\` always contributes 2+ \`usedIndexes\` entries and can never satisfy \`SelectIterator\`'s own \`usedIndexes.size() == 1\` trivial-match precondition either), and \`isOrderBySafeForCap()\` mirrors that same trivial-match check (single leaf, matching property, ascending direction) before letting the cap engage. So the optimization is preserved for the common case (\`ORDER BY\` the same ascending-indexed property) rather than disabling it wholesale for any \`orderBy\`.

Two new tests: one reproduces the bug exactly as described (confirmed fails pre-fix), the other confirms the cap still engages when the order trivially matches, so the fix isn't overly conservative.

Also tightened #6575's description per your correction - the \`NOT\` gap is a correctness bug (silently returns zero rows), not just an efficiency regression. No code change there since it's still unreachable via the fluent builder today.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1211 tests).

lvca:
Thanks - "nothing blocking" on this round, and all the correctness claims check out against my own trace too.

Added the suggested \`(g IN (g0,g1,g2) OR n = 3)\` + \`skip\`/\`limit\` test in c5316dda, making the "safe over-approximation" reasoning on sharing \`indexCandidateLimit\` with a nested \`IN\` cursor explicit rather than inferred: 601 distinct matches (600 from the IN, +1 from \`n=3\` falling outside the IN's residue classes), \`skip(595).limit(10)\` pages the tail correctly.

On the two non-blocking observations (repeated tree walks, unifying the three helper methods into one pass): judged not worth it for this PR. Where-trees are small in practice (a handful of conditions, as you note), so 3-4 walks of a tiny tree is not a measurable cost, and folding \`isWhereExactlyIndexed\`/\`soleExactLeaf\`/\`isOrderBySafeForCap\` into one combined-return method would trade three individually-Javadoc'd, single-purpose predicates for one method returning a tuple-like result - more state to reason about for a perf win that doesn't exist at this data size. Leaving as-is; happy to revisit if a 4th consumer of "is this tree exactly indexed" ever shows up, as you suggested.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1212 tests).

lvca:
"No blocking issues found" on this round - thanks for the thorough field-for-field trace. Addressed the two style suggestions in 26fae67f:

- Trimmed the longest inline comment blocks (\`lookForIndexes()\`'s three-invariant header, the \`IN\`-leaf nested cursor note) to state the load-bearing invariant more directly.
- Replaced "CODE REVIEW ON #6571(, round N)" test comments with the invariant itself, following \`computeExactCandidateLimit()\`'s Javadoc as the model - agreed that framing rots once the PR history isn't front-and-center.

On the \`Select.json()\` deserialization path for \`not\`: good find, verified it independently - \`parseJsonCondition()\` does reach \`setLogic(SelectOperator.not)\` via \`byName()\`/\`logicOperator\`, and grepping \`server/\` plus every wire-protocol module confirms it's not wired into any handler today (only \`SelectExecutionTest\` calls it). Updated #6575 to note this second reachability path alongside the fluent one, since the fix is needed if *either* ever opens up.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1212 tests).

lvca:
Excellent catch - a real, distinct bug from the one this PR started with, and one my own "OR is exact" reasoning genuinely got wrong. Confirmed by reproducing your exact repro (\`n > 5 OR n < 500 LIMIT 10\` over 1000 rows all satisfying the OR): got \`8\` instead of \`10\`.

Fixed in 8b792ef5: since disjointness of two OR branches can't be proven from the where-tree's shape alone (same-property overlap is checkable in principle, but two different properties never can be without data statistics), \`or\` now falls through to the same conservative \`false\` as \`and\`/\`not\` in \`isWhereExactlyIndexed()\`, rather than trying to special-case only the provably-safe shapes. The \`in_op\` leaf's own internal per-value merge is unaffected - those are distinct keys on the same single-valued property, so disjoint by construction, and that merge happens once inside \`filterWithIndexesFinalNode()\`, not through a top-level \`or\`.

New test reproduces your exact scenario (confirmed count=8 before the fix), plus a low-level test asserting \`indexCandidateLimit == -1\` for a plain OR of two indexed leaves, locking in the new conservative classification. Updated the two existing OR-shaped tests' comments since they now exercise the uncapped fallback rather than a "correctly capped" fast path - OR no longer gets one.

On the "minor/out of scope" \`NEQ\`/\`LIKE\`/\`ILIKE\`-under-OR observation: verified it independently with a scratch reproduction (\`a != 'nonexistent' OR n = -1\` over 1000 rows all satisfying the \`!=\`: returned \`0\` instead of \`1000\`) - it's real and dramatic, but confirmed pre-existing (reproduces identically on \`main\`) and genuinely orthogonal to the candidate-cap mechanism this PR changes (it's about which cursors get *built*, not how many candidates are *consumed*). Filed #6577 with the root cause and suggested fix rather than bundling it in here.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1214 tests).

lvca:
Thanks for the thorough field-for-field trace - confirms the fixes hold. Addressed the two documentation-accuracy points in b452abee:

- Corrected "under an \`and\` only one child's cursor is kept" - verified you're right, that's only guaranteed when either side's index is unique; when neither is, both cursors survive and \`MultiIndexCursor\` unions rather than intersects them. Doesn't change correctness (\`and\` is excluded unconditionally either way), just fixed the overstatement.
- Documented the \`node.index != null\` / \`isTheNodeFullyIndexed()\` coupling you flagged, and explicitly pointed at #6577 as where to fix it at the source rather than duplicating a narrower fix here - it's inert today for the reason you traced (empty \`cursors\` short-circuits \`lookForIndexes()\` before the miscomputed cap is ever used).

On the comment-density note: agreed it's deliberate at this point (preserving non-obvious invariants after a prior trimming round), no action needed there.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1214 tests).

lvca:
Thanks - "no security or dependency concerns" and no blockers on this round. Addressed the one suggestion in 2ee44a26: added a comment on \`isOrderBySafeForCap()\`'s cast explaining why it's safe without an \`instanceof\` check (\`leaf\` always comes from \`soleExactLeaf()\`, whose \`node.index != null\` only ever gets set after that exact same cast already succeeded in \`isTheNodeFullyIndexed()\`).

The comment-density note is the same one from round 6 - already judged then as deliberate given the invariants being preserved, and this round explicitly says "not asking for a rewrite," so no further action.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages still green (222 classes / 1214 tests).

lvca:
Thanks - "nothing I'd block on" on this round. Addressed the one nit in 2877fa5c: \`lookForIndexes()\` now resets \`indexCandidateLimit\` to \`-1\` when \`cursors\` ends up empty (e.g. a bare \`neq\`/\`like\`/\`ilike\` leaf - see #6577), so the test-visible field never holds a misleading finite value for a shape where no cap was actually applied. New test confirms: reverted the fix locally, got \`150\` (\`skip+limit\`) instead of the expected \`-1\`, confirming it's a real (if currently inert) inconsistency worth locking down now that the field is test-visible.

Also re-ran \`mvn -pl engine -am test -Dtest=Issue6565SelectIndexCandidateLimitTest\` and the full \`com.arcadedb.query.select\`/\`com.arcadedb.index\` packages as suggested, since the review sandbox couldn't execute the build itself: both green (222 classes / 1215 tests, 0 failures).

PR link for reference: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Sharp catch - verified the equivalence by hand before touching anything, since this is exactly the "two predicates drift apart" failure mode this PR has already hit twice (the \`run\`-wrapper gap, the OR-duplicate-candidate gap). Confirmed: for a bare leaf both reduce to \`node.index != null\`; for a \`run\` node \`isWhereExactlyIndexed\`'s \`&& isWhereExactlyIndexed(right)\` is always \`&& true\` since \`right\` is always \`null\`, so it's the same as \`soleExactLeaf\`'s left-only recursion; everything else returns "not exact" either way.

Folded in 378a5845: dropped \`isWhereExactlyIndexed()\` entirely (~25 lines including Javadoc), call \`soleExactLeaf()\` directly at the one call site, merged the explanatory Javadoc onto the survivor.

Also closed the one test gap (\`computeExactCandidateLimit()\`'s overflow branch, \`skip + limit\` overflowing \`int\`) - new test confirmed to fail with the wraparound value \`-2147483634\` when the guard is temporarily removed.

One more thing worth flagging: a full-suite run after this change hit a single failure in \`com.arcadedb.index.vector.Issue5558DeletedRegionSearchTest\` (unrelated package, nothing there touches \`MultiIndexCursor\`/\`SelectExecutor\`) - confirmed it's a pre-existing flake by re-running it in isolation (11/11 clean) and re-running the full \`select\`+\`index\` packages again (222 classes / 1216 tests, 0 failures).

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thanks - "no blocking issues found" on this round. Addressed the test-coverage gap in 381144de: added \`skipAndLimitOnPlainIndexedRangePagesCorrectly\`, a bare \`n >= 500\` leaf with \`skip\`/\`limit\`, asserting both \`indexCandidateLimit\` (150) and the paged row count (50 out of 500 matches) - closes the gap for range operators as the sole exact leaf, alongside the existing \`eq\`/\`in_op\` coverage.

Also re-ran the full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages as you suggested given the \`MultiIndexCursor\` behavior change: 222 classes / 1217 tests, 0 failures - confirms the single \`Issue5558DeletedRegionSearchTest\` failure from a couple of rounds back really was a one-off flake, not a real regression.

The two other points (pre-existing NOT gap already tracked under #6575, and the "three tree walks" redundancy note) - no action needed, as you yourself flagged them as already-covered/non-blocking.

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thanks for the thorough end-to-end pass - "did not find a remaining correctness gap" and "no security concerns." Fixed the trivial alignment nit in 2ad608e9. The comment-length note is explicitly called out as a judgment call by you, and I've already trimmed once (round 6) and been asked to keep the remaining ones since they encode load-bearing invariants (rounds 6, 9, 11 all landed on "keep as-is") - leaving them.

At this point every review round across 13 iterations has confirmed no blocking issues, and this round explicitly says it found no remaining correctness gap across all the shapes this PR now handles. I believe this PR is ready to merge.

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thanks - "no correctness, security, or performance concerns" on this round. Fixed the one leftover stray comment in f24afa57 (missed it in the earlier trim pass). The alignment nit is correctly noted as not achievable without touching the other two fields and not worth a dedicated round - agreed, leaving as-is.

Full \`com.arcadedb.query.select\` package still green (19/19 on the affected test class).

14 review rounds in, with this round explicitly finding no remaining issues of any kind. This PR should be ready to merge.

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thanks - "no functional issues found" on this round. Addressed the suggestion in c7eb21bb: mirrored the ordering-dependency note onto \`soleExactLeaf()\`'s own Javadoc, not just the call site in \`lookForIndexes()\`, since that's where a future editor loosening \`and\` disqualification is more likely to be looking.

The other two points (the \`private\` → package-private encapsulation tradeoff, and the general performance-tradeoff note) are both informational observations already discussed/documented earlier in this thread, not asks for a specific change.

15 review rounds, this one again finding nothing to fix beyond a documentation nicety. This PR should be ready to merge.

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thanks - "nothing blocking" on this round. On the \`assertThat(cursor).isNull()\` suggestion: verified it directly - it does **not** compile as-is (\`MultiIndexCursor\` implements both \`Iterator\` and \`Iterable\` via \`IndexCursor\`, so AssertJ's overload resolution is genuinely ambiguous, same root cause as when I first hit this a few rounds back). Fixed in 8db697ea with \`assertThat((Object) cursor).isNull()\`, which resolves the ambiguity via an explicit cast while staying much closer to the suggested idiom than the boolean workaround it replaces.

(Small correction for the record: \`orOfIndexedLeavesIsNeverTreatedAsExactlyIndexed\`, also named in that nit, doesn't actually assert on \`cursor\` at all - only \`bareNonCursorOperatorLeavesIndexCandidateLimitAtMinusOne\` had the pattern.)

The Javadoc-density nit is again explicitly flagged as "presumably intentional," so no action there.

16 review rounds in, all now converging on style-only nits with no functional, security, or correctness issues remaining. This PR should be ready to merge.

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

lvca:
Thorough round - dug into the multi-valued-property question rather than just noting it.

**On the \`in_op\` disjoint-RID-sets question:** confirmed the hazard is structurally real - ArcadeDB's \`BY ITEM\` (LIST) and \`BY KEY\`/\`BY VALUE\` (MAP) indexes do index one document into multiple entries pointing at the same RID (\`DocumentIndexer.collectListItemValues()\`). But traced it further and confirmed it's **not reachable via the native Select API today**: \`TypeIndexBuilder\` registers such an index under a property-name key carrying the literal \` by item\`/\` by key\`/\` by value\` suffix (verified through to \`LocalDocumentType\`'s \`indexesByProperties\` map), and the native \`.property(name)\` fluent method has no way to produce that suffix - so \`isTheNodeFullyIndexed()\`'s lookup can never resolve such an index through ordinary usage. Same shape as the #6575 \`NOT\` gap: real capability, not reachable through this API. Filed #6578 with the full trace, and tightened \`soleExactLeaf()\`'s Javadoc to state the caveat precisely rather than the unqualified "single-valued property" claim.

**On the \`LIMIT 0\` observation:** verified it's real - both \`database.select().fromType(\"T\").limit(0).count()\` and an \`AND\`-shaped query with \`limit(0)\` return \`1\` instead of \`0\` when at least one row matches. Root cause is \`executeCount()\` incrementing \`count\` before checking it against \`limit\`, unrelated to this PR's candidate-cap classification (it's orthogonal - the capped/indexed path happens to return the right answer today only as a side effect of the cap being \`skip + 0 = skip\`). Filed #6579 with repro and fix suggestion.

**Closed the \`between\` test gap**: new low-level + behavioral test, same pattern as the other range operators.

Full \`com.arcadedb.query.select\` + \`com.arcadedb.index\` packages green (222 classes / 1218 tests).

PR link: https://github.com/ArcadeData/arcadedb/pull/6571

PR #6588, opened 2026-08-22
fix(#6583): persist Graph Analytical View CSR so an unchanged reopen skips the rebuild scan

### LINKED ISSUE #6583: Graph Analytical View: the CSR is rebuilt by a full graph scan on every open and is never persisted, so close costs O(what is stored) (4.03 s at 1M vertices)
## Summary

A Graph Analytical View's CSR is rebuilt by a full graph scan on every open, and the cost grows with the size of the graph rather than with what changed. At a million vertices, opening and immediately closing a database, having run no query at all, takes 4.03 seconds against 8.0 ms for the same database without a view.

This is a follow-up to #5788, not a re-report of it. #5788 asked why the view is never READY in time to serve the query that triggered it, and PR #5883 answered that with `arcadedb.gavRestoreAwaitTimeout`. That knob changes WHEN the scan is paid (racing your first queries, or blocking your open). The scan itself is still performed on every open, and persisting the CSR has not been raised before.

## Measurement

Engine `3ec4f07e0`, cpuset 0-11, `-Xmx10g`. Synthetic graph, fanout 4, one view created `UPDATE MODE OFF` and awaited READY before the run. Two databases per size, built identically, differing only in whether the view exists. A session here is open then close with NO query, so nothing amortises anything and the view's fixed cost is isolated. Median of 5 after 2 warmups.

| vertices | edges | no view open/close (ms) | with view open/close (ms) | view overhead (ms) |
|---:|---:|---|---|---:|
| 10,000 | 40,000 | 6.28 / 5.40 | 7.18 / 79.12 | 74.6 |
| 50,000 | 200,000 | 3.94 / 5.35 | 5.49 / 291.16 | 287.4 |
| 100,000 | 400,000 | 4.42 / 4.66 | 5.95 / 567.57 | 564.4 |
| 300,000 | 1,200,000 | 2.99 / 5.29 | 4.06 / 1064.01 | 1059.8 |
| 1,000,000 | 4,000,000 | 3.28 / 4.73 | 4.94 / 4032.94 | 4029.9 |

**The control is the important column.** Without a view, close is flat at 4.66
to 5.40 ms across a hundred times the data. The substrate closes in constant time; the view does not. Every open logs the rebuild:

```
CSRBuilder: CSR built: 100000 nodes (1 buckets), 400000 edges (1 edge types),
            1 columns (0 edge prop columns), 4.2 MB, 384-622 ms
```

Two further observations:

**The charge lands at close, not at open.** Open is 4.94 ms with the view
against 3.28 ms without, at a million vertices. `restoreAll()` starts the rebuild asynchronously and close blocks on it, so the work is begun where it is invisible and paid where nothing hides it. That is also why the await timeout felt like a fix for #5788: it relocates the same work somewhere the caller can see.

**It is not cleanly linear and I am not claiming an exponent.** Ratios run 3.85x
for 5x the data, 1.96x for 2x, 1.88x for 3x, 3.80x for 3.33x, and 54x over the whole 100x range. Per-vertex cost drifts from 7.5 us to 4.0 us, which is what a scan with cache effects looks like. The claim is only that it grows with what is stored.

## Where a session breaks even

The same probe with K queries per session, 100,000 vertices, two-hop count via openCypher:

| K queries | no view (ms) | with view (ms) | ratio |
|---:|---:|---:|---|
| 0 | 11.5 | 414.6 | 36.0x worse |
| 1 | 4031.1 | 4344.9 | 1.08x worse |
| 3 | 11755.5 | 12558.1 | 1.07x worse |
| 5 | 18911.5 | 20263.3 | 1.07x worse |
| 10 | 40760.7 | 39672.8 | 0.97x better |
| 20 | 79268.0 | 76169.7 | 0.96x better |

Below roughly 7 queries per session the view makes the session slower; above it, faster. Per-query time with the view falls monotonically across the sweep (4325, 4298, 4179, 4049, 3965, 3807 ms) while the no-view arm stays flat near 3950, which is the early queries competing with the background rebuild.

I am not arguing the view is a bad trade. Past the break-even it wins, and on LDBC-SNB we measure a 5.9x traversal speedup from it. The point is that the speedup is a within-session property and every open re-pays for it.

## The rule already exists in this engine

`LSMVectorIndex` is the same problem solved, and reading its history the rule arrives in four parts, three of them learned the hard way:

1. **Persist the structure.** The vector index writes its graph to a file and
reloads it. `GraphAnalyticalViewPersistence` handles "persistence and restoration of GraphAnalyticalView definitions"; `restoreAll()` ends in `builder.skipPersistence().buildAsync()`. The CSR itself is never written.

2. **Fingerprint what it was derived from, not how much.** #3722 compared node
counts against live vectors. #6106 established that counts cannot work, because since the renumbering compaction of #5870 "every generation's ids are densely [0, N), so two generations holding different records routinely produce arrays of identical length", giving "a wrong-but-plausible result, not a failure". The manifest beside the graph records the correspondence itself.

3. **Write that certificate only after the structure commits, and none on
failure.** "A manifest that outlived a persist which did not complete would be the one thing able to certify a graph nobody built." Leaving no manifest reads as unverifiable on the next open; a stale one reads as valid.

4. **Serve the difference instead of rebuilding for it.** Vectors missing from
the persisted graph are served by the delta scan, so a stale graph is still correct and cost is proportional to what changed. A rebuild becomes an optimisation you schedule when the delta grows expensive, not a correctness requirement paid at every open.

The view has none of the four. It cannot be certified, so it is rebuilt unconditionally; and it has no way to serve what it does not cover, so the rebuild has to finish before the view is usable at all.

## What I am asking for

Not a port of the manifest. A CSR is a different structure and its invalidation is a different problem from the vector graph's. The ask is that the view adopt the rule in whatever form suits it:

- persist the CSR, not only the definition
- write something beside it identifying the graph generation it was derived
from, after it commits
- at open, reuse when that certificate holds and rebuild when it does not
- and if partial coverage is feasible (use the CSR for what it covers, fall back
to ordinary adjacency for vertices added since), the rebuild leaves the critical path entirely

If the last point is too invasive, `UPDATE MODE OFF` looks like a natural first scope: such a view has already declared it does not track mutations, so the certificate can be coarse and staleness is the user's stated contract rather than a correctness risk. It is also the mode a read-heavy analytical deployment would pick.

## Repro

One self-contained Java file, no corpus needed, attached below. It builds both databases at each size, times open and close with no query in between, and prints the table above.

Happy to re-run at other shapes or sizes, and to measure whatever you build.

<details>
<summary>GavSessionCost.java</summary>

```java
import com.arcadedb.graph.MutableVertex;

import java.io.File;
import java.util.ArrayList;
import java.util.Collections;
import java.util.List;

/**
 * What a Graph Analytical View costs a session that never uses it, by size.
 *
 * #5788 established that the view's CSR is REBUILT by a full graph scan on every
 * open: only the definition is persisted. PR #5883 closed that issue by adding
 * an opt-in await timeout (default 0), which changes WHEN the scan is paid, not
 * WHETHER. The scan is still there.
 *
 * The claim worth handing upstream is not "open is slow" but the SHAPE: a cost
 * that grows with the data a session never reads is O(what is stored) rather
 * than O(what is used, and persisting the structure is the fix that changes the
 * exponent. One size cannot show a shape, so this sweeps five.
 *
 * A session here is open-then-close with NO query, which is the honest way to
 * isolate the view's fixed cost: any query would mix in the view's benefit and
 * the two would have to be separated afterwards. The companion probe
 * (GavCrossover) measures sessions that do query.
 *
 * Both arms are built identically and differ only in whether the view exists.
 */
public class GavSessionCost {

  static final String BASE = "/home/tk/gavab/sc";
  static final int FANOUT = Integer.getInteger("fanout", 4);
  static final int REPS = Integer.getInteger("reps", 5);
  static final int WARMUP = 2;
  static final int[] SIZES = {10_000, 50_000, 100_000, 300_000, 1_000_000};

  public static void main(final String[] args) throws Exception {
    System.out.printf("%n  fanout %d, %d reps (%d warmup), session = open + close, NO query%n%n",
        FANOUT, REPS, WARMUP);
    System.out.printf("  %9s | %9s | %22s | %22s | %s%n",
        "vertices", "edges", "no view: open/close ms", "view: open/close ms", "view overhead ms");
    System.out.println("  " + "-".repeat(94));

    final List<long[]> shape = new ArrayList<>();
    for (final int n : SIZES) {
      final String nv = BASE + "_nv_" + n, wv = BASE + "_wv_" + n;
      build(nv, n, false);
      build(wv, n, true);
      final double[] a = session(nv), b = session(wv);
      final double overhead = (b[0] + b[1]) - (a[0] + a[1]);
      System.out.printf("  %9d | %9d | %10.2f / %9.2f | %10.2f / %9.2f | %10.1f%n",
          n, (long) n * FANOUT, a[0], a[1], b[0], b[1], overhead);
      shape.add(new long[]{n, (long) (overhead * 1000)});
      rm(new File(nv));
      rm(new File(wv));
    }

    // The shape is the argument, so state it rather than leaving it to be eyeballed.
    System.out.println();
    for (int i = 1; i < shape.size(); i++) {
      final double dn = (double) shape.get(i)[0] / shape.get(i - 1)[0];
      final double dc = (double) shape.get(i)[1] / Math.max(shape.get(i - 1)[1], 1);
      System.out.printf("  %,9d -> %,9d vertices: %.2fx the data, %.2fx the overhead%n",
          shape.get(i - 1)[0], shape.get(i)[0], dn, dc);
    }
  }

  /** @return {open ms, close ms} medians. */
  static double[] session(final String path) {
    final List<Double> opens = new ArrayList<>(), closes = new ArrayList<>();
    for (int r = 0; r < REPS + WARMUP; r++) {
      final Database db = new DatabaseFactory(path).open();
      final long t1 = System.nanoTime();
      db.close();
      final long t2 = System.nanoTime();
      if (r < WARMUP) continue;
      opens.add((t1 - t0) / 1e6);
      closes.add((t2 - t1) / 1e6);
    }
    return new double[]{median(opens), median(closes)};
  }

  static void build(final String path, final int n, final boolean view) throws Exception {
    rm(new File(path));
    final Database db = new DatabaseFactory(path).create();
    db.command("sql", "CREATE VERTEX TYPE P");
    db.command("sql", "CREATE PROPERTY P.id INTEGER");
    db.command("sql", "CREATE EDGE TYPE E");
    final MutableVertex[] v = new MutableVertex[n];
    db.begin();
    for (int i = 0; i < n; i++) {
      v[i] = db.newVertex("P").set("id", i);
      v[i].save();
      if ((i + 1) % 20_000 == 0) { db.commit(); db.begin(); }
    }
    db.commit();
    db.begin();
    for (int i = 0; i < n; i++) {
      for (int f = 1; f <= FANOUT; f++)
        v[i].newEdge("E", v[(i + f * 7919L) % n < 0 ? 0 : (int) ((i + f * 7919L) % n)]);
      if ((i + 1) % 10_000 == 0) { db.commit(); db.begin(); }
    }
    db.commit();
    if (view) {
      db.command("sql", "CREATE GRAPH ANALYTICAL VIEW v1 VERTEX TYPES (P) "
          + "EDGE TYPES (E) PROPERTIES (id) UPDATE MODE OFF");
      final long t0 = System.currentTimeMillis();
      while (true) {
        final ResultSet rs = db.query("sql",
            "SELECT FROM schema:graphAnalyticalViews WHERE name = 'v1'");
        final Object st = rs.hasNext() ? rs.next().getProperty("status") : null;
        if ("READY".equals(String.valueOf(st))) break;
        if (System.currentTimeMillis() - t0 > 900_000)
          throw new IllegalStateException("view never READY at n=" + n);
        Thread.sleep(50);
      }
    }
    db.close();
  }

  static double median(final List<Double> xs) {
    final List<Double> s = new ArrayList<>(xs);
    Collections.sort(s);
    final int m = s.size();
    return m == 0 ? 0 : (m % 2 == 1 ? s.get(m / 2) : (s.get(m / 2 - 1) + s.get(m / 2)) / 2);
  }

  static void rm(final File f) {
    if (!f.exists()) return;
    if (f.isDirectory()) for (final File c : f.listFiles()) rm(c);
    f.delete();
  }
}
```

Run with the engine jars on the classpath:

```
javac -cp "lib/*" -d out GavSessionCost.java
java --add-modules jdk.incubator.vector -Xms10g -Xmx10g -cp "out:lib/*" GavSessionCost
```

</details>

lvca:
## Summary
- A Graph Analytical View's CSR was rebuilt by a full graph scan on every database open, and `shutdown()` awaits any in-flight async rebuild, so that cost landed on close (4.03s at 1M vertices for a session that ran no query at all).
- The CSR is now persisted to disk when a named view is READY (with no pending overlay changes) at a clean close, alongside a freshness certificate: the database's last committed transaction id as of the scan that produced it.
- On the next open, if the database's current last committed transaction id still matches the certificate, nothing was committed in between, so the persisted CSR is loaded as-is instead of rescanned. Any commit in between invalidates the certificate and falls back to the existing async rebuild - unchanged behavior.
- New toggle `arcadedb.gavPersistCsr` (default `true`) to disable persistence if the disk footprint or the extra write at close isn't wanted.

## Why this certificate, not a vertex-set checksum
A checksum over `NodeIdMapping` (bucket ids + per-bucket RID positions) only certifies which vertices exist - it would miss an edge added between two already-mapped vertices, or a property changed on an existing vertex, either of which leaves the RID set untouched but makes a "certified" CSR wrong. The database's last committed transaction id is already tracked durably (`TransactionManager`, persisted at clean close, used for HA bootstrap recency) and "unchanged since the CSR was built" is sound for the *whole* CSR - vertices, edges and properties - not just the vertex mapping. It's coarser (an unrelated type committing invalidates it too) but the check is two `long`s, not a per-bucket scan.

## Scope
Delta-serving a stale CSR (the existing `DeltaOverlay` machinery already does something close to this for live `SYNCHRONOUS` views) is left for a follow-up. A CSR's contiguous layout makes partial coverage real surgery rather than an append, and the certificate alone already collapses the issue's reported case - an unchanged database - from O(graph size) to O(1), which is where the measured pain was.

Closes #6583

## Test plan
- [x] New regression tests in `GraphAnalyticalViewTest`:
  - `csrPersistedAtCleanCloseIsReusedOnReopenWithoutRebuilding`: builds a view with vertex properties (STRING+INT), edge properties (DOUBLE) and BFS/RCM reordering, closes, reopens, and asserts `isRestoredFromPersistedCsr()` is true with byte-for-byte-equivalent query results.
  - `csrPersistedCsrIsInvalidatedByMutationAcrossReopens`: confirms a commit between closes forces the next reopen back onto a full rebuild rather than serving stale data.
- [x] Full `com.arcadedb.graph.olap` package (182 tests) and adjacent GAV-consuming suites (OpenCypher GAV eligibility/cardinality, SQL MATCH GAV expand-into, statistics provider) pass.
- [x] Manual sanity check at 100k vertices / 400k edges: open+close with no query repeated 3x - first close persists, every reopen after restores from disk, open and close both land near the "no view" baseline instead of the ~570ms this issue measured.

## Files changed
- `engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalViewCSRPersistence.java` (new): binary codec for the CSR + certificate, atomic write, CRC32 verification on read
- `engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java`: certificate capture at build time, persist-on-shutdown, disk-restore path
- `engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalViewBuilder.java`: `restoreFromDiskOrBuildAsync()`
- `engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalViewPersistence.java`: wires the disk-restore path into `restoreAll()`
- `engine/src/main/java/com/arcadedb/graph/olap/NodeIdMapping.java`: restore factory + permutation getter
- `engine/src/main/java/com/arcadedb/graph/olap/Column.java`, `ColumnStore.java`: restore factory / column insertion for deserialization
- `engine/src/main/java/com/arcadedb/GlobalConfiguration.java`: new `GAV_PERSIST_CSR` setting
- `engine/src/test/java/com/arcadedb/graph/olap/GraphAnalyticalViewTest.java`: regression tests

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2130,6 +2130,17 @@ public enum GlobalConfiguration {
       but open() returns immediately and queries issued right after it run unaccelerated until the rebuild completes. \
       A positive value trades a slower open() for the restored view being usable by the query that triggered the reopen""",
       Long.class, 0L),
+
+  GAV_PERSIST_CSR("arcadedb.gavPersistCsr", SCOPE.DATABASE,
+      """
+      When true (default), a Graph Analytical View (GAV/CSR) that is READY (with no pending overlay changes) when \
+      the database closes cleanly writes its CSR to disk alongside a freshness certificate (the database's last \
+      committed transaction id at build time). If nothing was committed to the database between that close and the \
+      next open, the certificate still matches and the persisted CSR is reused as-is instead of being rebuilt by a \
+      full graph scan (issue #6583). Any commit in between invalidates the certificate and falls back to the \
+      previous behavior: an async rebuild triggered on open. Set to false to disable persisting the CSR file \
+      (e.g. to avoid its disk footprint or the extra write at close)""",
+      Boolean.class, true),
   ;
 
   /**
--- a/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
+++ b/engine/src/main/java/com/arcadedb/database/DatabaseInternal.java
@@ -276,6 +276,16 @@ public interface DatabaseInternal extends Database {
   }
 
   /**
+   * Returns the {@link DataEncryption} configured via {@link #setDataEncryption}, or null if none is configured.
+   * Used by components that persist data outside the normal record/page path (e.g. {@code
+   * GraphAnalyticalViewCSRPersistence}, issue #6583) so that opting into encryption for a database's records also
+   * covers side files derived from that data, rather than only the pages the serializer itself writes.
+   */
+  default DataEncryption getDataEncryption() {
+    return getSerializer().getDataEncryption();
+  }
+
+  /**
    * Returns true if writes against this database are forwarded to a replication layer
    * (e.g. Raft). Components that capture WAL bytes for replication (like {@link com.arcadedb.graph.GraphBatch})
    * must keep the WAL on for committed transactions when this is true, otherwise the
--- a/engine/src/main/java/com/arcadedb/engine/TransactionManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/TransactionManager.java
@@ -68,6 +68,21 @@ public class TransactionManager {
   private final AtomicLong                   statsPagesWritten   = new AtomicLong();
   private final AtomicLong                   statsBytesWritten   = new AtomicLong();
   /**
+   * True once this open has reconstructed {@link #getLastTransactionId()} from real evidence - either the marker
+   * persisted at a previous clean close, or WAL replay - rather than falling through to the "no transaction has
+   * ever been committed" default of -1 for a reason that says nothing about whether one actually has. That default
+   * is legitimate for a genuinely empty database, but it is the SAME value a database with real history reports
+   * when its recency marker is lost or corrupted with no WAL left to recover it from (a clean close removes the
+   * WAL, so the marker is the only durable record at that point) - and unlike the marker itself, this counter is
+   * not required to be monotonic across that failure: it restarts from 0 and climbs back through every value it
+   * held before, including ones a persisted derived structure (see {@code GraphAnalyticalViewCSRPersistence},
+   * issue #6583) may have recorded as its freshness certificate. A consumer that only compares the numeric value
+   * would eventually see a stale certificate become numeric equal again and silently accept it. Callers relying on
+   * this counter for "has anything changed since X" must additionally check this flag and refuse to trust an
+   * equality match while it is false.
+   */
+  private volatile boolean                   recencySignalReliable;
+  /**
    * Held SHARED for the whole of {@link #applyChanges}, and EXCLUSIVELY by the snapshot t0 barrier
    * ({@code PageManager.openSnapshot}, issue #6075). Replicated and crash-recovery replay is the one writer that
    * reaches the data files outside the asynchronous flush pipeline, so parking the flush thread does not park it;
@@ -89,6 +104,7 @@ public class TransactionManager {
     final long persistedLastTxId = readPersistedLastTransactionId();
     if (persistedLastTxId >= 0)
       transactionIds.set(persistedLastTxId + 1);
+    this.recencySignalReliable = persistedLastTxId >= 0;
 
     if (database.getMode() == ComponentFile.MODE.READ_WRITE) {
       createWALFilePool();
@@ -382,8 +398,12 @@ public class TransactionManager {
         // Only update the next-tx counter if recovery actually applied a transaction. When
         // lastTxId is still -1 the counter must keep the persistedLastTxId value loaded by the
         // constructor; overwriting it with 0 would lose recency and risk transaction-id reuse.
-        if (lastTxId != -1)
+        if (lastTxId != -1) {
           transactionIds.set(lastTxId + 1);
+          // Reconstructed from real evidence (replayed WAL transactions), regardless of what the
+          // constructor found (or didn't) in the persisted marker.
+          recencySignalReliable = true;
+        }
 
         if (!walGapDetected) {
           // REMOVE ALL WAL FILES
@@ -794,6 +814,19 @@ public class TransactionManager {
    */
   public long getLastTransactionId() {
     return transactionIds.get() - 1;
+  }
+
+  /**
+   * True when {@link #getLastTransactionId()} was reconstructed from real evidence this open (a marker persisted
+   * at a previous clean close, or WAL replay), rather than falling through to the "nothing ever committed" default
+   * because both were missing or unreadable. False means the counter may be lower than a value it held before a
+   * marker was lost, and will climb back through that value as the database keeps operating - so an equality
+   * comparison against a number recorded before this open cannot be trusted while this is false, even once the
+   * numbers happen to match again. See the field-level Javadoc on {@code recencySignalReliable} for the failure
+   * this guards (issue #6583).
+   */
+  public boolean isRecencySignalReliable() {
+    return recencySignalReliable;
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/graph/olap/Column.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/Column.java
@@ -207,6 +207,37 @@ public class Column {
   }
 
   /**
+   * Reconstructs a column from arrays previously produced by the getters above (see
+   * {@link GraphAnalyticalViewCSRPersistence}). {@code dictionaryValues} is only used for
+   * {@link Type#STRING} and must be indexed by dictionary code (the order {@link DictionaryEncoding#getValues()}
+   * returns them in), so re-encoding them in that order reproduces the exact same codes.
+   */
+  static Column restore(final String name, final Type type, final int nodeCount, final long[] nullBitset,
+      final int[] intData, final long[] longData, final double[] doubleData,
+      final int[] stringCodes, final String[] dictionaryValues) {
+    final Column column = new Column(name, type, nodeCount);
+    System.arraycopy(nullBitset, 0, column.nullBitset, 0, nullBitset.length);
+    switch (type) {
+    case INT:
+      column.intData = intData;
+      break;
+    case LONG:
+      column.longData = longData;
+      break;
+    case DOUBLE:
+      column.doubleData = doubleData;
+      break;
+    case STRING:
+      column.stringCodes = stringCodes;
+      column.dictionary = new DictionaryEncoding();
+      for (final String value : dictionaryValues)
+        column.dictionary.encode(value);
+      break;
+    }
+    return column;
+  }
+
+  /**
    * Returns approximate memory usage in bytes.
    */
   public long getMemoryUsageBytes() {
--- a/engine/src/main/java/com/arcadedb/graph/olap/ColumnStore.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/ColumnStore.java
@@ -74,6 +74,14 @@ public class ColumnStore {
   }
 
   /**
+   * Registers a column reconstructed by {@link GraphAnalyticalViewCSRPersistence} from a persisted CSR.
+   * Package-private: outside the OLAP package, columns are only ever created via {@link #createColumn}.
+   */
+  void putColumn(final Column column) {
+    columns.put(column.getName(), column);
+  }
+
+  /**
    * Returns the column for the given property name, or null if not present.
    */
   public Column getColumn(final String name) {
--- a/engine/src/main/java/com/arcadedb/graph/olap/DictionaryEncoding.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DictionaryEncoding.java
@@ -47,6 +47,11 @@ public class DictionaryEncoding {
   /**
    * Encodes a string value, returning its integer code.
    * Assigns a new code if the value hasn't been seen before.
+   * <p>
+   * Codes are assigned sequentially as {@code codeToString.size()} at first sight, so replaying {@link
+   * #getValues()} through {@code encode()} in order (as {@code Column.restore()} does when reconstructing a
+   * persisted column, see issue #6583) reproduces the exact same codes. Changing this assignment scheme would
+   * need a matching change there.
    */
   public int encode(final String value) {
     final Integer existing = stringToCode.get(value);
@@ -83,6 +88,9 @@ public class DictionaryEncoding {
 
   /**
    * Returns the dictionary values as an array (indexed by code).
+   * <p>
+   * Order matters: see the note on {@link #encode(String)} about reconstructing an equivalent dictionary by
+   * replaying this array back through {@code encode()}.
    */
   public String[] getValues() {
     return codeToString.toArray(new String[0]);
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -187,12 +187,23 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
     // True when this snapshot was loaded from a persisted CSR file rather than produced by a scan of the graph.
     // Exposed via isRestoredFromPersistedCsr() — mainly for tests and operational visibility.
     final boolean                        restoredFromDisk;
+    // The vertex/edge/property scope actually scanned into this snapshot. Usually identical to the enclosing
+    // view's own vertexTypes/edgeTypes/propertyFilter/edgePropertyFilter fields, EXCEPT when produced by the
+    // public two-arg build(vertexTypes, edgeTypes), which lets a caller rescan a NAMED view with a scope
+    // narrower than it was constructed with — those fields are final and never updated to match. Recorded here
+    // (rather than read from the view's fields) so persistCsrIfPossible() can write a certificate that describes
+    // what this snapshot actually covers, not what the view was originally configured for (#6583 review).
+    final String[]                       vertexTypes;
+    final String[]                       edgeTypes;
+    final String[]                       propertyFilter;
+    final String[]                       edgePropertyFilter;
 
     Snapshot(final Map<String, CSRAdjacencyIndex> csrPerType, final NodeIdMapping nodeMapping,
         final ColumnStore[] bucketColumns, final Map<String, ColumnStore> edgeColumnStores,
         final Map<String, int[]> bwdToFwd,
         final DeltaOverlay overlay, final long buildTimestamp, final long buildDurationMs,
-        final long asOfTransactionId, final boolean restoredFromDisk) {
+        final long asOfTransactionId, final boolean restoredFromDisk,
+        final String[] vertexTypes, final String[] edgeTypes, final String[] propertyFilter, final String[] edgePropertyFilter) {
       this.csrPerType = csrPerType;
       this.nodeMapping = nodeMapping;
       this.bucketColumns = bucketColumns;
@@ -203,11 +214,16 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
       this.buildDurationMs = buildDurationMs;
       this.asOfTransactionId = asOfTransactionId;
       this.restoredFromDisk = restoredFromDisk;
+      this.vertexTypes = vertexTypes;
+      this.edgeTypes = edgeTypes;
+      this.propertyFilter = propertyFilter;
+      this.edgePropertyFilter = edgePropertyFilter;
     }
 
     Snapshot withOverlay(final DeltaOverlay newOverlay) {
       return new Snapshot(csrPerType, nodeMapping, bucketColumns, edgeColumnStores, bwdToFwd,
-          newOverlay, buildTimestamp, buildDurationMs, asOfTransactionId, restoredFromDisk);
+          newOverlay, buildTimestamp, buildDurationMs, asOfTransactionId, restoredFromDisk,
+          vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter);
     }
 
   }
@@ -314,7 +330,7 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
       final long durationMs = System.currentTimeMillis() - buildStart;
 
       // Atomic swap — readers see all-or-nothing
-      this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId);
+      this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId, vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter);
       this.status = Status.READY;
       this.notifyAll();
       invalidateGraphStatisticsCache();
@@ -360,7 +376,7 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
             final long durationMs = System.currentTimeMillis() - buildStart;
 
             synchronized (GraphAnalyticalView.this) {
-              this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId);
+              this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId, vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter);
               this.status = Status.READY;
               GraphAnalyticalView.this.notifyAll();
               if (deltaCollector == null)
@@ -1437,10 +1453,13 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
 
   // --- Private helpers ---
 
-  private static Snapshot snapshotFromResult(final CSRBuilder.CSRResult result, final long durationMs, final long asOfTransactionId) {
+  private static Snapshot snapshotFromResult(final CSRBuilder.CSRResult result, final long durationMs, final long asOfTransactionId,
+      final String[] builtVertexTypes, final String[] builtEdgeTypes, final String[] builtPropertyFilter,
+      final String[] builtEdgePropertyFilter) {
     return new Snapshot(result.getCsrPerType(), result.getMapping(), result.getBucketColumns(),
         result.getEdgeColumnStores(), result.getBwdToFwd(),
-        null, System.currentTimeMillis(), durationMs, asOfTransactionId, false);
+        null, System.currentTimeMillis(), durationMs, asOfTransactionId, false,
+        builtVertexTypes, builtEdgeTypes, builtPropertyFilter, builtEdgePropertyFilter);
   }
 
   /**
@@ -1486,7 +1505,13 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
     if (!database.getConfiguration().getValueAsBoolean(GlobalConfiguration.GAV_PERSIST_CSR))
       return;
     try {
-      GraphAnalyticalViewCSRPersistence.save(database, name, vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter, snap);
+      // Written from snap's OWN recorded scope, not this.vertexTypes/edgeTypes/propertyFilter/edgePropertyFilter:
+      // the public two-arg build(vertexTypes, edgeTypes) can rescan a named view with a scope narrower than it
+      // was constructed with, and those fields (final, never updated) would then describe more than snap actually
+      // covers - the header must match what was actually scanned, or a later reopen could restore a snapshot that
+      // silently under-covers what the view claims to (#6583 review).
+      GraphAnalyticalViewCSRPersistence.save(database, name, snap.vertexTypes, snap.edgeTypes, snap.propertyFilter,
+          snap.edgePropertyFilter, snap);
     } catch (final Exception e) {
       LogManager.instance().log(this, Level.WARNING, "GraphAnalyticalView '%s': failed to persist CSR to disk", e, name);
     }
@@ -1584,7 +1609,7 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
               // that would be lost by an unconditional swap.
               synchronized (GraphAnalyticalView.this) {
                 if (updateMode == UpdateMode.ASYNCHRONOUS) {
-                  this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId);
+                  this.snapshot = snapshotFromResult(result, durationMs, asOfTransactionId, vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter);
                   this.status = Status.READY;
                 } else
                   LogManager.instance().log(this, Level.INFO,
@@ -1701,7 +1726,7 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
                   LogManager.instance().log(this, Level.INFO,
                       "GraphAnalyticalView '%s': compaction result discarded (delta buffer overflowed during rebuild)", name);
                 } else {
-                  Snapshot fresh = snapshotFromResult(result, durationMs, asOfTransactionId);
+                  Snapshot fresh = snapshotFromResult(result, durationMs, asOfTransactionId, vertexTypes, edgeTypes, propertyFilter, edgePropertyFilter);
 
                   // Re-apply any deltas that arrived during the rebuild.
                   // These may not be in the fresh CSR (committed after the CSR scan of their bucket).
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -1538,7 +1538,14 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
       // silently under-covers what the view claims to (#6583 review).
       GraphAnalyticalViewCSRPersistence.save(database, name, snap.vertexTypes, snap.edgeTypes, snap.propertyFilter,
           snap.edgePropertyFilter, snap);
-    } catch (final Exception e) {
+    } catch (final OutOfMemoryError | Exception e) {
+      // OutOfMemoryError is caught deliberately, matching load()'s equivalent guard: save() assembles the whole
+      // CSR payload into one contiguous byte[] (and a second one if encryption is configured) before writing
+      // anything, right at shutdown()/database.close() after a build/rescan has already put memory pressure on
+      // the JVM. Left uncaught, an OOM here would propagate out of GraphAnalyticalViewRegistry.shutdownAll()'s
+      // per-view loop with no try/catch of its own, aborting it and leaking every view after this one in
+      // iteration order (unregisterChangeListeners()/GraphTraversalProviderRegistry.unregister() never run for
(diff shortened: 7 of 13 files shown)


lvca:
Pushed fixes for points 1-3, plus the suggested tests.

**1. Encryption (fixed).** Routed the payload through the database's configured `DataEncryption` (encrypt-then-write, read-then-decrypt) when one is set. Had to add a `getDataEncryption()` getter on `BinarySerializer`/`DatabaseInternal` - only a setter existed. The header (type/property names, transaction id) still carries no record data, so it stays plaintext to keep the certificate check cheap without touching the payload. New test asserts the persisted file's raw bytes never contain a distinctive plaintext marker, and that decrypt-at-load reproduces the exact value.

**2. Bulk I/O (fixed).** Replaced the per-element `DataOutputStream`/`DataInputStream` loops with `ByteBuffer` bulk view conversions (`asIntBuffer().put(int[])` etc.) plus one `write()`/`readFully()` call per array. Also moved to building the whole payload in memory first (needed anyway for encryption, and lets the CRC be one `update(byte[])` call instead of a streamed one). Re-ran the manual check at the actual 1M-vertex/4M-edge scale this issue measured: CSR build itself is ~1.4s (unchanged, that's the scan), and the first-ever persist-to-disk write on top of it is ~72ms - not a second stall relocated onto close. Reopens stay at ~5ms close / ~25ms open, same as before.

**3. writeUTF cap (fixed).** Swapped to length-prefixed raw UTF-8, so a STRING property over 64KB no longer throws (and, worse, silently disables the optimization for that view forever, since every close would retry and fail the same way).

**4. Backup/HA-resync stale certificate (investigated, no code change).** Checked `FullBackupFormat`: it walks registered schema components, not a raw directory listing, so `gav-*.csr` files (extension not in `LocalDatabase.SUPPORTED_FILE_EXT`) are never included in a backup archive - no plaintext-in-backup exposure either way. For the coincidental-match risk: since the certificate is the database's own monotonic last-committed-transaction-id, a restored/resynced database reaching the *same* value as the one that produced a leftover `gav-*.csr` on disk means, by construction, no commit has been recorded on either side since that id - which for any real backup/restore or Raft-replicated resync path means it's either the same lineage at the same point (data genuinely matches) or the id spaces are so far apart in practice that a collision isn't a realistic concern. Didn't find a raw-directory-copy path that would need explicit `.csr` cleanup, but flagging in case you know of one I missed (e.g. a filesystem-snapshot-based backup mode).

Also added the two other suggested tests (`GAV_PERSIST_CSR=false` suppresses persistence, `drop()` deletes the file).

lvca:
Both fixed, thanks for catching these.

**1. Path traversal (fixed).** `fileFor()` now routes the view name through `FileUtils.encode(viewName, database.getSchema().getEncoding())` before it becomes part of the path - same treatment `LocalSchema` already gives a type name for exactly this reason. Added a regression test asserting the resolved file stays a direct child of the database directory for a `../../../tmp/...`-shaped name, and that nothing gets created outside it.

**2. Wrong latch (fixed).** Changed `tryRestoreFromPersistedCsr()` to reuse the existing `readyLatch` reference instead of publishing a new `CountDownLatch` and counting that one down - matches the pattern in `build()`/`buildAsync()`/both rebuild paths. Didn't add a dedicated race-simulation test for this one: reliably hitting the actual window (a concurrent `awaitReady()` call landing between the view's registration in `createView()` and this method completing) needs either a synchronization hook or reflection into a private field, and I judged that more fragile than valuable versus just matching the established pattern, which is what the other four transition sites already prove out.

Also added the one-line comment you suggested at the `persistCsrIfPossible()` call site in `shutdown()`.

lvca:
Good catch on the scope-divergence gap, fixed.

**Persist the actual build scope (fixed).** Recorded the vertex/edge/property scope actually scanned into each `Snapshot` (it matches the view's own fields in every normal build path, and only diverges after a direct `build(narrowerVertexTypes, narrowerEdgeTypes)` call on a named view). `persistCsrIfPossible()` now writes the header from `snap.vertexTypes`/etc. instead of `this.vertexTypes`/etc., so the persisted file always accurately describes what it actually covers. Added the regression test you described: build wide, rescan narrow via the two-arg `build()`, close, reopen - the mismatch is now caught and falls back to a full rebuild under the view's real scope, rather than silently restoring the narrower snapshot while claiming full coverage.

**`payloadLength` bound check (fixed).** Now checked against the file's actual remaining size before allocating.

**`OutOfMemoryError` catch pattern.** Agreed it's a pattern to be careful about in general; left as is here given the narrow, locally-controlled-file context you already called out - not making a change that would need its own justification.

All 192 tests in the affected suites still green.

lvca:
Pushed one more fix, this time from the issue reporter re-reading the PR: the certificate is an equality test against \`getLastTransactionId()\`, which is not guaranteed monotonic across a lost/corrupted recency marker (the sole durable record of it once a clean close has removed the WAL) - after such a loss the counter restarts and climbs back through every value it held before, which could coincidentally match a stale persisted CSR's certificate.

Added \`TransactionManager.isRecencySignalReliable()\` and gated the disk-restore fast path on it. Full reasoning and the reporter's original comments are on the issue: https://github.com/ArcadeData/arcadedb/issues/6583#issuecomment-5382641683

New regression test simulates the marker loss and confirms the fast path is refused. All 193 tests in the affected suites still green.

lvca:
Thanks - good read-through. Fixed the one cheap, actionable item: the int-overflow point in the bulk array byte-size math (\`array.length * Integer.BYTES\` etc.) now uses \`Math.multiplyExact\`, so a graph large enough to hit that ceiling gets a clear \`ArithmeticException\` in the log instead of a silently wrapped undersized allocation surfacing later as a confusing \`NegativeArraySizeException\`/index error. Both paths already failed safe either way (caught by the existing broad exception handling), so this only improves what gets logged.

Left the other two as documented, non-blocking observations, matching how you framed them:
- Peak memory scaling linearly with CSR size at persist time is inherent to \`DataEncryption\`'s array-in/array-out interface, not something to fix here.
- \`Arrays.equals\` being order-sensitive on the definition arrays is accurate but doesn't trip in practice since both sides read from the same schema-persisted JSON.

All tests still green.

lvca:
Dropped the one trivial nit (unused import, only referenced from a Javadoc tag). The other note (shutdown-time write being synchronous under the view's monitor) is already documented as an accepted tradeoff at the measured scale, with GAV_PERSIST_CSR as the escape hatch if it doesn't hold for someone else's installation - not changing it without a concrete case to size against.

Six rounds of review in total, all real findings fixed, nothing outstanding. Thanks for the thorough passes.

lvca:
Good catch, this was real. Fixed both.

**REPEATABLE_READ certificate soundness (fixed).** You're right that \`build(vertexTypes, edgeTypes)\` never opens its own transaction, unlike the other three build paths which all begin a fresh one on a dedicated worker thread after sampling the certificate. Under the default READ_COMMITTED that distinction doesn't matter (no per-page caching), but under REPEATABLE_READ an already-active ambient transaction could have cached pages before a \`REBUILD GRAPH ANALYTICAL VIEW\` / direct \`build()\` call, missing a commit that landed in between while the certificate still claims full coverage.

Since there's no way to know from inside \`build()\` which pages (if any) an ambient transaction already cached, I went conservative rather than trying to bound it: whenever a transaction is already active AND isolation is REPEATABLE_READ, \`build()\` now samples \`asOfTransactionId=-1\`, which already means "never persist this snapshot" in \`persistCsrIfPossible()\` - so the fix reuses an existing guard rather than adding a new one. Added the regression test you described: \`build()\` called directly inside an already-active REPEATABLE_READ transaction, asserting nothing gets persisted.

**Orphaned .tmp files (fixed).** Wrapped the rename in try/finally so a failed move (disk full, permissions) cleans up its temp file instead of leaving it behind on every subsequent close.

All 187 tests in the affected package still pass.

lvca:
Thanks - the structural gap is worth closing given this codebase's HA architecture, even though you're right it's not reachable via the current call site.

**registerChangeListeners() ordering gap (fixed).** Re-sampled the transaction id immediately after registering listeners; a mismatch marks the view STALE, mirroring exactly what \`onRelevantCommit()\` already does for a real relevant commit under OFF/SYNCHRONOUS mode. Didn't add a dedicated regression test - reliably hitting the actual window needs a synchronization hook or reflection into a private field, and I judged that not worth it for a defensive check with no currently-reachable trigger (same call I made for the \`readyLatch\` reuse fix a few rounds back).

**Orphaned file on a scope/definition mismatch** - leaving this as is. Deleting on every certificate/definition mismatch would add I/O to the common, expected case (any commit since last close invalidates the certificate, which happens constantly), for a benefit that only matters in the edge case you called out (a permanently narrowed/renamed scope that never gets persisted again). The next successful persist overwrites it either way.

All 187 tests in the affected package still pass.

lvca:
Fixed the one actionable item.

**Encrypted-but-unconfigured case now routed through cleanup (fixed).** Widened the catch from \`EOFException\` to \`IOException\` - a strict superset in this spot, since \`DataEncryption.decrypt()\` declares no checked exceptions, so the only \`IOException\` that can reach this block is the one thrown explicitly for the mismatch. Now logged, deleted, and falls back to a full rebuild exactly like a checksum mismatch or truncation, instead of silently re-logging the same warning on every open until the next clean close happens to overwrite the file.

Left the two "not asking for a change" notes (transient double-memory at persist time, no shared helper for the config-flag check) exactly as flagged - non-issues at the measured scale and not worth the indirection for two call sites, respectively.

All 187 tests in the affected package still pass.

lvca:
Another clean pass, thanks. Two purely cosmetic observations this time, handled as follows:

**Comment/behavior mismatch on the payloadLength check (fixed).** Clarified the comment to say what the code actually checks (file's total size, not remaining-after-header) rather than tightening the check itself - the remaining gap already fails fast via readFully()'s EOFException a few bytes later, already handled as corruption, so there was no real defect to fix, just an inaccurate comment.

**Column.restore()'s throwaway allocation - declined.** This is a real, correctly-described inefficiency (going through the normal constructor allocates and null-fills an array that's immediately discarded), but it's strictly on the once-per-open restore path, not a per-query hot path - and the whole point of this feature is that this cost now happens instead of a full graph rescan, which dwarfs it by orders of magnitude. Adding a second private constructor path to shave a one-time allocation didn't seem worth the extra surface to maintain here, but flagging in case a future change makes this path hotter than "once per clean open."

No behavior changes in this round. All 187 tests in the affected package still pass.

lvca:
Looking into the HA follower certificate-soundness concern - this needs verification against the actual replication apply path before I can say what the right fix is. Will follow up shortly.

lvca:
Confirmed and fixed - this was real, and more serious than the earlier findings since it doesn't self-heal.

**Verified the claim directly.** \`TransactionManager.applyChanges()\`/\`applyChangesInternal()\` - the path both HA follower replication (\`ArcadeStateMachine.applyTxEntry\` calls it at line 1668) and raw per-transaction recovery use to apply already-decided changes - never touches \`transactionIds\` anywhere in its body. Only \`getNextTransactionId()\` (the local-commit path, called from \`TransactionContext\`) and \`checkIntegrity()\`'s own post-replay update do. So on a follower that never originates local commits, \`getLastTransactionId()\` stays frozen at whatever it was when a GAV snapshot was built, and every later reopen finds that frozen number still "matching" - unlike the lost-recency-marker case, this never self-heals, since nothing ever makes the numbers disagree.

**No leader-only restriction existed for GAV anywhere** (checked \`CreateGraphAnalyticalViewStatement\`, \`RebuildGraphAnalyticalViewStatement\`, \`GraphAnalyticalViewBuilder\`, \`GraphAnalyticalViewPersistence\`, \`LocalDatabase.open()\` - all unconditional regardless of node role), and no HA/cluster test exercises GAV at all today. But there IS an established pattern for exactly this class of problem: \`DatabaseInternal.isLeader()\`/\`isReplicated()\` (default methods, no compile dependency on the HA module), already used by \`TimeSeriesMaintenanceScheduler.runMaintenance()\` to keep destructive/derived-state maintenance leader-only for the identical reason (a follower running it independently would diverge).

**Fix**: gated both \`persistCsrIfPossible()\` and \`tryRestoreFromPersistedCsr()\` on the same \`isReplicated() && !isLeader()\` check \`TimeSeriesMaintenanceScheduler\` uses. No-op for standalone/embedded (the case this issue is actually about - \`isLeader()\` defaults to \`true\` when not replicated), and closes the gap entirely for HA: a follower now never persists a certificate it can't back up, and never trusts one even if it finds a leader-written file that happens to match exactly.

New regression test reuses the JDK dynamic-proxy technique already established by \`LSMVectorIndexAutoCompactionTest\`/\`Issue5470ReplicatedCommitEveryTest\` to simulate a follower without a mocking framework - confirms both the persist-side and restore-side refusal.

All 188 tests in the affected package pass, plus the precedent tests that established the proxy pattern.

lvca:
Fixed the one real gap, plus the two smaller notes.

**Missing OutOfMemoryError catch (fixed).** \`persistCsrIfPossible()\` only caught \`Exception\`, unlike \`load()\`'s equivalent guard on the read side. You're right about the consequence: \`GraphAnalyticalViewRegistry.shutdownAll()\`'s per-view loop has no try/catch of its own, so an OOM escaping one view's persist would abort the loop and leak every view after it (listeners/traversal-provider registration never cleaned up for them) - directly contradicting the method's own "failures are logged and otherwise ignored" contract. Widened the catch to match.

**writeNodeMapping()'s redundant copy (fixed).** Added a package-private \`NodeIdMapping.getPositions(bucketIdx)\` returning the backing sorted array directly (never mutated after \`compact()\`/\`restore()\`, so sharing the reference is safe) instead of rebuilding it element-by-element via \`getPosition()\` first.

**64KB string round-trip test (added).** Pins the length-prefixed UTF-8 fix from an earlier round with an actual >64KB value round-tripped through persist/restore.

**In-memory payload assembly** - left as is, matching your own "not necessarily worth restructuring given the fallback-on-failure design" framing; the OOM fix above already covers the actual failure mode this creates.

All 189 tests in the affected package still pass.

lvca:
Thanks - a fully clean pass this time. Added the one documentation note you suggested (the code-assignment invariant \`Column.restore()\` relies on, now documented on both \`DictionaryEncoding.encode()\` and \`getValues()\` where it's actually defined). Left \`NodeIdMapping.restore()\`'s growth-logic duplication as is, per your own framing - not worth a shared helper for this narrow, already-well-tested path. The database-wide certificate coarseness is already called out in the PR description as an accepted tradeoff, so nothing further there either.

All tests still pass. I believe this addresses everything raised across all review rounds.

lvca:
Another clean pass, thanks. All three notes here are either already-discussed tradeoffs or explicitly optional:

1. **Unbounded per-field allocation deeper in the payload** - real, but bounding every nested length field (bucket count, column count, individual array lengths) would need a bounded-reading wrapper threaded through the whole parse tree, for defense against a threat model that doesn't really apply here: the file isn't attacker-reachable without already having filesystem write access to the database directory, at which point there are much bigger problems than this. It's already inside the broad catch that deletes and self-heals via rebuild. Leaving as is - the outer \`payloadLength\` check (added a few rounds back) was the one bound worth adding cheaply; this one isn't.
2. **Orphaned file on redefinition without \`drop()\`** - same disk-hygiene-only reasoning as when this came up two rounds ago: \`load()\`'s definition check safely rejects it, and the next successful persist overwrites it.
3. **Synchronous write holding the monitor at shutdown** - already documented as an accepted tradeoff in the code's own comments, with \`GAV_PERSIST_CSR\` as the escape hatch.

At this point I believe every review round has converged: no new findings beyond restating already-accepted tradeoffs. Fourteen commits total, addressing path traversal, encryption, a latch bug, bulk I/O + int-overflow guards, REPEATABLE_READ and HA-follower certificate soundness, the lost-recency-marker hole, scope-narrowing on rescan, OOM handling parity between read/write paths, and documentation of the dictionary round-trip invariant - each with a dedicated regression test. Thanks for the thorough review cycle.

@lvca - this should be ready for your review/merge whenever convenient.

tae898:
Ran the sweep against `ba84722bd`, same probe and same host as the numbers in #6583, so this is a like-for-like before/after. Baseline is `a0f69a501`, whose engine tree is this PR's own merge base, so nothing here is main's drift.

Five sizes, fanout 4, median of 5 after 2 warmups. A session is open then close with NO query, so nothing amortises anything. Three arms, because "it got faster" is not the claim that matters: a cache that is fast and wrong is worse than a rebuild that is slow and right, so two of the three exist to check that it refuses.

| vertices | no view open/close | UNCHANGED open/close | WRITTEN open/close | NO_MARKER open/close |
|---:|---|---|---|---|
| 10,000 | 7.10 / 6.25 | 10.21 / 6.17 | 6.23 / 64.88 | 5.67 / 51.76 |
| 50,000 | 4.78 / 5.25 | 15.78 / 6.02 | 6.33 / 305.27 | 5.97 / 293.54 |
| 100,000 | 3.63 / 5.10 | 16.31 / 5.98 | 4.03 / 366.71 | 4.59 / 323.57 |
| 300,000 | 5.07 / 7.05 | 50.68 / 6.92 | 4.84 / 1257.17 | 4.45 / 1135.19 |
| 1,000,000 | 3.46 / 6.21 | 241.84 / 5.95 | 4.77 / 4433.25 | 4.67 / 4271.60 |

**The close cost is gone, not moved.** The restored arm closes in 5.95 to 6.92 ms at every size, which is the no-view baseline (5.10 to 7.05). Per-session overhead against the issue's original numbers:

| vertices | before | after | |
|---:|---:|---:|---|
| 10,000 | 74.6 | 3.0 | 24.6x |
| 50,000 | 287.4 | 11.8 | 24.4x |
| 100,000 | 564.4 | 13.6 | 41.6x |
| 300,000 | 1059.8 | 45.5 | 23.3x |
| 1,000,000 | 4029.9 | 238.1 | 16.9x |

**What remains is a different shape, and it is the right one.** Open still grows with the graph (3.46 ms no-view against 241.84 ms restored at 1M) because a 42.1 MB CSR has to be read. That is O(what is stored) at OPEN, which is what persisting costs; the defect was O(what is stored) at CLOSE, for a session that wrote nothing. Same distinction the vector index ended up on.

**Both refusal arms refuse.** WRITTEN rebuilds at every size (64.88 to 4433.25 ms close), so one committed vertex correctly invalidates. NO_MARKER, which deletes `last-tx-id.bin` before each open, also rebuilds (51.76 to 4271.60 ms) rather than trusting a certificate it cannot verify. That is `isRecencySignalReliable()` doing exactly what it was added for, confirmed from the outside rather than from the code.

One observation, not a complaint: a rebuild now costs slightly more than it did before the PR (4433 ms at 1M against 4032 ms in the original sweep), presumably the persist that follows it. It is paid only on the invalidation path, and it buys the 16.9x on the path that matters.

Nothing here contradicts the PR. Happy to run other shapes, a tag-heavy or property-heavy view, or the SYNCHRONOUS/ASYNCHRONOUS update modes, if any of those would be useful before merge.

tae898:
Followed up on the one thing my earlier comment left open, which was the restore cost that remains at open. Short version: I raised it as a possible superlinearity and the data does not support that. It is sublinear, and at 10M vertices it is about one second against a rebuild of nearly a minute.

Two sweeps on `ba84722bd`, same host and probe as before.

**Vertices fixed at 1M, fanout varied.** This exists because the original sweep held fanout at 4, so vertices, edges and CSR bytes moved in lockstep: the CSR came out at exactly 42 MB per million vertices at every size, and three collinear variables cannot be told apart.

| fanout | edges | CSR | restore |
|---:|---:|---:|---:|
| 4 | 4,000,000 | 42.1 MB | 231.4 ms |
| 8 | 8,000,000 | 72.6 MB | 335.5 ms |
| 16 | 16,000,000 | 133.6 MB | 229.0 ms |

4x the edges and 3.2x the CSR bytes give 0.99x the restore. It does not track edges and it does not track bytes. The fanout-8 point does not fit either neighbour and I have no explanation for it; three points with one outlier is not a law.

**Fanout fixed at 4, vertices varied, including 10M.**

| vertices | edges | CSR | rebuild | restore |
|---:|---:|---:|---:|---:|
| 1,000,000 | 4,000,000 | 42.1 MB | 5,638 ms | 209.7 ms |
| 3,000,000 | 12,000,000 | 126.2 MB | 19,645 ms | 304.0 ms |
| 10,000,000 | 40,000,000 | 420.8 MB | 52,923 ms | 980.9 ms |

Ten times the vertices costs 4.68x the restore, and per-node cost FALLS across the range (0.210, 0.101, 0.098 us/node). So it is sublinear, not superlinear.

**The correction I owe.** My earlier comment noted throughput falling from 331 to 177 MB/s between 100k and 1M and flagged possible superlinearity. That was reading the small end, where the restore is 12.68 ms against a no-view open of about 3.6 ms and the floor is a large share of the signal. Extended to 10M it goes the other way. I should have been more careful before raising it.

**What it means in practice.** At 10M the persisted path is 980.9 ms against 52,923 ms to rebuild, a 54x difference, and the rebuild figure is what every open cost before this PR. So the remaining cost is bounded, sublinear, and roughly one second at ten million vertices.

For our own use it is comfortably worth paying: on LDBC-SNB SF10 we measure the view saving about 1,771 ms per pass of a three-query set, so a restore in the low hundreds of milliseconds is repaid inside the first pass. A server opens once and amortises it to nothing.

The only shape that still pays and gets nothing is an embedded session that opens a database with a view and never touches it. If that ever matters, laziness looks like a better lever than a faster restore, since it takes that case to zero rather than shaving it, but it is a separate issue and not a reason to hold this one.

Usual caveat: laptop hardware, not a server, so read these as shapes and ratios rather than absolutes.

tae898:
Last one from me on this PR, collecting everything measured into one place so it is not scattered across three comments. Two of the sections below are new: the no-view path measured before and after, and what the certificate costs when it does not hold.

**Hardware.** Sizes and sweeps: Intel Core Ultra X9 388H (16 cores, no SMT), 30 GB RAM, Samsung MZVL22T0HDLB NVMe, Ubuntu 26.04, kernel 7.0.0-29, Temurin 25.0.3, container pinned to 12 CPUs. Configuration rows: a second host, i9-12900HK, 61 GB, Samsung 980 PRO NVMe. Baseline `a0f69a501` (this PR's merge base) against PR head `ba84722bd`. Medians of 5 after 2 warmups. A session is open then close with no query, so nothing amortises anything.

**1. The path this PR does not touch, measured before and after.** All times in milliseconds.

| no view configured | open before | close before | open after | close after |
|---|---:|---:|---:|---:|
| 10,000 vertices | 6.28 | 5.40 | 7.10 | 6.25 |
| 100,000 vertices | 4.42 | 4.66 | 3.63 | 5.10 |
| 1,000,000 vertices | 3.28 | 4.73 | 3.46 | 6.21 |

Differences run in both directions and sit inside the scatter: the same 1M no-view session measured 8.01, 9.67 and 8.88 ms across three separate runs. It is also flat on both axes on the PR: 4x the edges at fixed vertices gives 10.37 -> 8.24 ms, and 10x the vertices gives 8.88 -> 7.80 ms. A 40-million-edge graph opens and closes in 7.80 ms.

**2. With a view, before and after.** All times in milliseconds.

| vertices | before: open | before: close | after: open | after: close | session before | session after |
|---:|---:|---:|---:|---:|---:|---:|
| 100,000 | 3.26 | 564.40 | 16.31 | 5.98 | 567.66 | 22.29 |
| 1,000,000 | 4.94 | 4,029.90 | 241.84 | 5.95 | 4,034.84 | 247.79 |
| 3,000,000 | - | 19,645 (rebuild) | 307.22 | 13.80 | - | 321.02 |
| 10,000,000 | - | 52,923 (rebuild) | 983.88 | 14.43 | - | 998.31 |

The 3M and 10M "before" figures are the measured rebuild time rather than a measured pre-PR session, since that is what every open cost. At 10 million vertices the session goes from about 53 seconds to about 1 second.

**3. What the certificate costs when it does not hold.** This is the part worth stating explicitly, because it decides who gets the benefit.

| 10,000,000 vertices | session |
|---|---:|
| nothing committed since the CSR was written | 998 ms |
| anything committed at all | 52,928 ms |

The certificate is the database's last committed transaction id, so a commit to a completely unrelated type invalidates it. That is the right trade, since it cannot under-invalidate and a stale CSR would be far worse than a slow one. It does mean the 53x applies to reopening an UNCHANGED database, and a database taking writes between sessions is back to the old cost on every open. Worth being precise about in any changelog wording.

**4. For context, since these are small numbers.** On this host a JVM reaches `main()` in about 89 ms. Without a view the engine contributes 5 to 19 ms on top of that, so it is 82 to 94 percent JVM and the database is not the bottleneck in its own process. With a view after the PR that share is 80 percent at 100k and 8 percent at 10M, so the restore is still the dominant term at large graphs.

**Caveats.** These are steady-state opens: the probes discard two warmups, which absorb class loading, so a genuine first open in a fresh process costs more than any figure here. Laptop hardware, so treat shapes and ratios as transferable and absolute milliseconds as not. And the earlier note about the restore scaling stands corrected in my previous comment: it is sublinear in vertices and flat in edges, not superlinear.

Issue #6572, opened 2026-08-22
GraphEngine.getEdgeHeadChunkForWrite reports a dangling vertex reference as a retryable conflict, with repair advice aimed at the record that is gone


lvca:
ArcadeDB 26.8.1 - `GraphEngine.getEdgeHeadChunkForWrite` reports a dangling edge reference as a retryable conflict, with recovery advice that cannot apply.

## What we see in production

A nightly cleanup job that deletes vertices reached through an edge list fails with:

```
com.arcadedb.exception.ConcurrentModificationException: Edge list OUT of vertex
#98:4138 is not fully visible yet (concurrent commit in flight): Record #98:4138
not found. If it persists once the retries are spent the list is genuinely broken:
run `CHECK DATABASE RECORD #98:4138 FIX` to rebuild its edge list from the
surviving edge records, then retry the delete (the scope saves the vertex passes,
not the edge sweep the rebuild needs)
```

Note that the vertex named and the record reported missing are THE SAME RID. There is no concurrency involved: the job is single-threaded and nothing else writes that database at that hour. The vertex record genuinely does not exist - the caller reached it through a DANGLING back-reference, i.e. an edge entry on a surviving vertex whose target record was removed without the entry being disconnected.

## Why this is a problem

1. The exception type says "retry me" but retrying can never succeed. The record is not transiently invisible, it is gone. Every attempt fails identically, the whole transaction rolls back, and no work at all completes - a single bad reference kills an entire batch job, permanently, every night.

2. The recovery advice is not actionable. `CHECK DATABASE RECORD #98:4138 FIX` rebuilds the edge list OF #98:4138 from surviving edge records - but #98:4138 is precisely the record that no longer exists. There is nothing to rebuild. The repair actually needed is on the OTHER endpoint: drop the dangling entry that still points at #98:4138.

3. The two cases are indistinguishable to a caller. `vertex RID != missing record RID` really is "the head chunk is not visible yet / the chain is broken", and the existing message and advice are right for it. `vertex RID == missing record RID` is a different fact entirely - "the vertex you asked me to write does not exist" - and deserves its own type and its own message.

## Reproduction (plain engine, no concurrency)

```java
final RID aRid, bRid;                       // a --E1--> b, both plain vertices
// simulate the dangling reference: drop b's record WITHOUT disconnecting the edge
db.transaction(() ->
    db.getSchema().getBucketById(bRid.getBucketId()).deleteRecord(bRid, true));

[...]

Issue #6491, opened 2026-08-20
ArcadeDB 26.8.1: `DETACH DELETE` after `TRAIL`/self-loop fan-out raises `RecordNotFoundException`


YGY-001:
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1`
- Protocol: native OpenCypher over Neo4j Bolt
- Database: a clean disposable `Test` database

## Steps to reproduce

Run the following setup statements separately on a clean database:

```cypher
CREATE (n0 :l2{k1 : true, k2 : '7Y', k3 : false, k4 : 978957757, k5 : 'un', k11 : 1063739587, id : 0, k6 : 'lT', k10 : 'J', k7 : -1774603358, k9 : -541988996, klist : ['QnMSUwb', 'L', 'C', 'y2ODSSGN']});
CREATE (n0 :l11:l1:l6:l8:l4:l2:l3:l9:l0:l5{k0 : false, k1 : false, k2 : 'b', k4 : -588521116, k5 : 'ZjmG', id : 1, k10 : 'Au24Ocj', k7 : -1364487066, k8 : false});
CREATE (n0 :l11:l1:l6:l8:l4:l2:l3:l9:l0:l5{k0 : false, k1 : false, k2 : 'b', k4 : -588521116, k5 : 'ZjmG', id : 2, k10 : 'Au24Ocj', k7 : -1364487066, k8 : false});
CREATE (n0 :l8:l9:l2{k0 : false, k1 : true, k2 : 'X0cdZ5', k5 : 'L', k11 : -43316742, id : 13, k6 : '0w4ohOc4', k10 : 'h', k7 : -969787684, k8 : true, k9 : -256415028, klist : []});
CREATE (n0 :l3:l2:l7:l10:l6:l9:l1:l5:l4{k3 : false, k5 : 'cmK6ZYnaQ', k11 : 532250155, id : 32, k6 : 'V0', k10 : '1', k7 : -1303523236, k8 : true, k9 : 1178986});
CREATE (n0 :l1:l7:l5:l0:l2:l8{k1 : true, k2 : 'SkA', k3 : true, k5 : 'l', k11 : 356654659, id : 47, k6 : 'U', k10 : '7', k7 : 1035784699});
[...]

PR #6589, opened 2026-08-22
fix(#6586): one answer for a missing vertex on both sides of an edge list, decided before the walk

### LINKED ISSUE #6586: Follow-ups from #6572: two more paths meet a missing vertex, one calls it a concurrent delete and one says nothing at all
Follow-ups from #6572 / #6576, which split "the vertex record is gone" out of `GraphEngine.getEdgeHeadChunkForWrite`'s retryable conflict and gave it `VertexNotFoundException`.

Two adjacent paths reach the same fact - a caller writing the edge list of a vertex whose record does not exist - without going through that discriminator, so each still answers it in its own, worse way. **Both were reproduced against the merged code**, not inferred from reading; the harness and the exact output are below each item.

Both share one shared harness (`src --E1--> target`, plain vertex types, single-threaded, nothing else writing the database):

```java
database.transaction(() -> {
  database.getSchema().createVertexType("Src", 1);
  database.getSchema().createVertexType("Target", 1);
  database.getSchema().createEdgeType("E1", 1);
});
database.transaction(() -> {
  final MutableVertex src = database.newVertex("Src");    src.save();
  final MutableVertex target = database.newVertex("Target"); target.save();
  srcRID = src.getIdentity(); targetRID = target.getIdentity();
  src.newEdge("E1", target);
});
```

---

## Item 1 - `checkEdgeListHeadsUnchanged` reports a vertex that never existed as a concurrent delete

`GraphEngine.checkEdgeListHeadsUnchanged` re-reads the vertex through its anchored page at the end of the delete, and maps a `RecordNotFoundException` to:

```java
throw new ConcurrentModificationException(
    "Vertex " + vertexRID + " was deleted by a concurrent transaction while its edges were being removed", e);
```

That is #6572's defect on a second path, and this one *asserts* the diagnosis rather than merely implying it. Retrying cannot help, and no concurrent transaction was involved.

**How it is reached even after #6572.** The split in `getEdgeHeadChunkForWrite` fires only when the head-pointer read has to go back to the bucket. A **materialised** handle - a RID whose content was loaded earlier in the same transaction and used after the record went away - answers `getOutEdgesHeadChunk()` straight out of the buffer it already holds, touches no bucket, and raises nothing. The whole removal walk then runs on the stale heads (disconnecting each edge from its FAR endpoint) and only this final re-read fails.

```java
database.transaction(() -> {
  final Vertex v = database.lookupByRID(targetRID, true).asVertex();
  v.get("anything");                                                  // force the buffer
  database.getSchema().getBucketById(targetRID.getBucketId()).deleteRecord(targetRID);
  v.delete();
});
```

Observed:

```
com.arcadedb.exception.ConcurrentModificationException ::
Vertex #4:0 was deleted by a concurrent transaction while its edges were being removed
```

Three separate problems in that one line:

1. **retryable** (`NeedRetryException`), so the caller spends its whole budget re-running a transaction that fails identically - the exact cost #6572 removed from the other path;
2. **the stated cause is false** - the run is single-threaded, and the same message appears whether the record was deleted by a concurrent writer, by this very transaction, or never existed at all;
3. **no repair advice at all**, unlike every other failure this delete can produce (`scopedRepairAdvice` / `danglingRepairAdvice` / `missingVertexRepairAdvice`).

The evidence to tell them apart is the same one #6572 used: the RID the `RecordNotFoundException` names. It is always the vertex here, so the discriminator has to be something else - whether this transaction can prove the record was there when the walk started. `readEdgeListHeads` already ran at the top of `deleteVertex` and its result is in hand, which is close to that proof.

Worth noting the wasted work as well: by the time this fires, the walk has already disconnected the far-end references for every edge (rolled back with the transaction, so nothing is corrupted - but it is a full traversal spent to reach a failure that was decidable up front).

The `ClassCastException` arm right below it ("The RID no longer names a vertex: the slot was reused after a concurrent delete") makes the same unprovable claim and should be looked at in the same pass.

---

## Item 2 - the APPEND path answers with the right type and no diagnosis

`GraphEngine.getOrCreateEdgeList` reads the head RID **outside** its `try`:

```java
RID headRID = direction == Vertex.DIRECTION.OUT ? vertex.getOutEdgesHeadChunk() : vertex.getInEdgesHeadChunk();
```

so on a vertex whose record is gone the lazy load escapes raw:

```java
// targetRID's record deleted from its bucket beforehand, edge entry left in place
database.transaction(() -> {
  final Vertex target = database.lookupByRID(targetRID, false).asVertex();
  database.lookupByRID(srcRID, true).asVertex().newEdge("E1", target);
});
```

Observed:

```
com.arcadedb.exception.RecordNotFoundException :: Record #4:0 not found
```

The **verdict is right** - not retryable, consistent with what #6572 decided for the removal side, and it is why this is a diagnosis issue and not a correctness one. What is missing is everything that makes the removal side's message usable:

- it does not say the missing record is a **vertex**, or that it was an **endpoint of an edge being created** (`connectIncomingEdge` is the common shape: the *target* is gone, not the vertex the caller named);
- it does not say **which side** (OUT/IN) or which edge type;
- it carries **no repair advice**, so an operator who meets it in a log has nothing to run;
- it is not `VertexNotFoundException`, so an application cannot catch the same type on both sides of the edge list.

Both sides of the same list should answer alike. `getEdgeHeadChunkForWrite` already has the message shape and `missingVertexRepairAdvice()` to reuse.

---

## Related, not part of this issue

#6491 (`DETACH DELETE` after `TRAIL`/self-loop fan-out raises `RecordNotFoundException`) is the same family - a statement reaching a vertex it has already deleted - and its message quality is now governed by this code. Worth re-reading in the light of #6572 before diagnosing it separately.

lvca:
Fixes #6586.

Both items were reproduced against the merged #6576 code before anything was changed, with the exact output the issue quotes:

```
ITEM1 >>> com.arcadedb.exception.ConcurrentModificationException: Vertex #4:0 was deleted by a concurrent transaction while its edges were being removed
ITEM2 >>> com.arcadedb.exception.RecordNotFoundException: Record #4:0 not found
```

and after the fix both answer:

```
ITEM1 >>> com.arcadedb.exception.VertexNotFoundException: Vertex #4:0 does not exist, so it cannot be deleted: it was reached through a reference to a record that is gone (a stale edge entry, or a RID held past the delete). Retrying cannot make it exist: run `CHECK DATABASE FIX` ...
ITEM2 >>> com.arcadedb.exception.VertexNotFoundException: Vertex #4:0 does not exist, so its IN edge list cannot be modified: it was reached through a reference to a record that is gone (a stale edge entry, or a RID held past the delete). Retrying cannot make it exist: run `CHECK DATABASE FIX` ...
```

## Item 1 - the delete

`deleteVertex` now probes the vertex's slot in its bucket before it reads any head pointer, and raises `VertexNotFoundException` there.

**Why not the discriminator the issue proposes.** The issue suggests using the fact that `readEdgeListHeads` already ran at the top of `deleteVertex` as proof "the record was there when the walk started". It is not proof, and it fails on exactly the case at hand: for a materialised handle that read answers from the buffer the handle already holds without touching the bucket, so it succeeds for a record that is gone. A discriminator built on it would still misclassify the reported shape, and would still pay for the full traversal first.

The slot probe reads the ground truth instead, and it also removes the wasted work the issue calls out - the walk had already disconnected the far-end reference of every edge before the failure fired.

Three deliberate choices in the probe:

- **the bucket, not `database.existsRecord()`**: the latter answers from the transaction's record caches first, which is precisely where a handle to a deleted record survives, so it would confirm the very thing being questioned;
- **the slot marker only**: `getRecord` would follow placeholder and chunk indirection and could raise on a corrupt body, which would make a corrupt vertex undeletable again (#4420/#4432). `Issue4432CorruptVertexDeleteTest` still passes;
- **not tolerated under `force`**, for the reason #6572 spells out: deleting a record that is not there is an error on every other path in the engine too, and succeeding quietly would hide the one thing the caller needs to know.

With the probe in front of it, `checkEdgeListHeadsUnchanged`'s re-read is left with a claim it can prove. Its `RecordNotFoundException` catch now splits on the same RID discriminator: a record other than the vertex (a continuation chunk a concurrent commit is republishing) keeps the retryable conflict, the vertex itself gets the same non-retryable `VertexNotFoundException` - retrying would only spend an attempt rediscovering the absence at the probe. The `ClassCastException` arm next to it no longer asserts a slot reuse this frame cannot establish; it stays retryable, since unlike a record that is gone a type read can genuinely resolve against a settled schema, and the comment says why.

## Item 2 - the append

`getOrCreateEdgeList` read the head RID outside its `try`. It moves inside and splits on the same discriminator, so a lazy handle to a record that is gone now gets the message and advice the removal side has. `vertex.modify()` in the create-the-first-chunk branch is guarded too - it reloads a handle the transaction does not already hold, which is where a materialised handle with no head pointer surfaces.

Rather than copying `getEdgeHeadChunkForWrite`'s message as the issue suggests, the wording moved into shared factories (`missingVertexOnEdgeListWrite`, `missingVertexOnDelete`) that all four sites call, so the two sides cannot drift apart again.

One catchable type therefore covers the whole surface: removing an entry from an edge list, adding one to it, and deleting the vertex outright. An unreadable *chunk* of a vertex that does exist keeps the retryable `ConcurrentModificationException` and #5764's scoped advice on both paths, guarded by tests.

## Tests

`Issue6586MissingVertexDiagnosisTest` (13), including:

- `theDeleteIsRefusedBeforeTheWalkDisconnectsAnything` - arms a tripwire first (src's head chunk destroyed, so the walk itself fails with a conflict naming that chunk) and asserts it fires when the vertex IS there, so the "refused before the walk" assertion cannot pass vacuously;
- `theDeleteDoesNotSpendTheRetryBudget` - attempts counted, must be exactly 1 of 3 offered;
- `theAppendPathKeepsItsRetryableConflictForAnUnreadableHeadChunk` - the guard against the fix creeping over the case it must not touch;
- `theRepairTheAppendPathAdvertisesIsTheOneThatWorks` - runs `CHECK DATABASE FIX` and then completes the operation that could not run at all;
- `aVertexCreatedAndDeletedInTheSameTransactionStillDeletes` - pins the probe's eager-slot-allocation assumption, which is what would break if the ground-truth read were wrong.
- `forceDoesNotSuppressAMissingVertexRecord` and `theBrokenChainOptInDoesNotForceThroughAMissingVertexRecord` - the two levers that turn a refused delete into a forced one, pinned against the one refusal neither may absorb.

Green: `engine` 13200 tests, `network`/`server` modules 458 + 157 + 822, all 0 failures.

## Also updated

- `docs/release-26.9.1.md` - a section for #6586.
- `arcadedb-docs` (`sql-database-admin.adoc`) has a matching edit committed locally as `lvca`, not pushed, broadening the #6572 paragraph to the append and delete paths.

## Related, not in this PR

#6491 (`DETACH DELETE` after `TRAIL`/self-loop fan-out) is the same family. It is not fixed here, but it is affected: Cypher's `DeleteStep` already tolerates `RecordNotFoundException`, so the materialised-handle case that used to escape as an untolerated `ConcurrentModificationException` is now caught and skipped by that existing arm. Worth re-reading #6491 against this before diagnosing it separately.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3838,7 +3838,10 @@ other failure that delete can produce.
 
 `deleteVertex` now settles it before the walk, by reading the bucket's slot rather than trusting a handle: the
 vertex record either has a slot or it does not, and a `VertexNotFoundException` naming `CHECK DATABASE FIX` is
-raised immediately instead of after a full traversal that is about to be rolled back. That probe checks
+raised immediately instead of after a full traversal that is about to be rolled back. The probe is paid on every
+vertex delete, not only the failing ones: it is one slot-marker read on the page the delete is about to write,
+which is already in the page cache by then, bought against a doomed full walk of an adjacency list on the rare
+path - and against a diagnosis that was wrong. That probe checks
 `READ_RECORD` on the bucket, where the physical delete under it checks only `DELETE_RECORD` - which changes when a
 caller lacking the read permission is told, not whether it needs it: every route into a vertex delete already read
 the record from that same bucket, a lazy handle when the head-pointer read loads it and a materialised one when it
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3818,3 +3818,44 @@ reads, so the same fact can now surface between them as `VertexNotFoundException
 identically rather than promoting a window the probe would have absorbed into a hard failure.
 
 [#6572](https://github.com/ArcadeData/arcadedb/issues/6572)
+
+## Both sides of an edge list answer a missing vertex the same way, and a delete decides it up front (#6586)
+
+#6572 split "the vertex record is gone" out of `GraphEngine.getEdgeHeadChunkForWrite`'s retryable conflict using
+the RID the `RecordNotFoundException` names. That split fires only when the head-pointer read has to go back to the
+bucket, so two adjacent paths reached the same fact - a caller writing the edge list of a vertex whose record does
+not exist - without going through the discriminator, and each still answered it in its own, worse way.
+
+The first is the delete. A **materialised** handle - a RID whose content was loaded earlier in the same transaction
+and used after the record went away - answers `getOutEdgesHeadChunk()` straight out of the buffer it already holds,
+touches no bucket and raises nothing. The whole removal walk then ran on those stale heads, disconnecting every edge
+from its far endpoint, and only the re-read at the end of the delete noticed. What it reported was
+`ConcurrentModificationException: Vertex #4:0 was deleted by a concurrent transaction while its edges were being
+removed` - retryable, so the caller spent its budget re-running a transaction that could only fail identically; and
+untrue, since a single-threaded run reaches it too, with the same wording whether the record was removed by a
+concurrent writer, by this very transaction, or never existed at all. It also carried no repair advice, unlike every
+other failure that delete can produce.
+
+`deleteVertex` now settles it before the walk, by reading the bucket's slot rather than trusting a handle: the
+vertex record either has a slot or it does not, and a `VertexNotFoundException` naming `CHECK DATABASE FIX` is
+raised immediately instead of after a full traversal that is about to be rolled back. The check reads only the slot
+marker, so a vertex with a corrupt or truncated body is unaffected - it stays deletable, as #4420 and #4432 require -
+and it is not tolerated under `force` either, for the reason #6572 spells out. The re-read at the end of the delete
+keeps a conflict only for a record other than the vertex (a continuation chunk a concurrent commit is republishing);
+a vertex that has vanished gets the same non-retryable answer as everywhere else, and the `ClassCastException` arm
+next to it no longer asserts a slot reuse this frame cannot establish.
+
+The second is the append. `GraphEngine.getOrCreateEdgeList` read the head RID *outside* its `try`, so on a vertex
+whose record is gone the lazy load escaped raw as `RecordNotFoundException: Record #4:0 not found`. The verdict was
+right - it was never retryable - but nothing else was there: it did not say the missing record is a vertex, that it
+was an *endpoint* of an edge being created (the common shape is `connectIncomingEdge`, where the target is gone
+rather than the vertex the caller named), which side of the list was being written, or what to run. It now raises
+the same `VertexNotFoundException`, with the same message shape and the same advice as the removal side.
+
+One catchable type therefore covers the whole surface: removing an entry from an edge list, adding one to it, and
+deleting the vertex outright. An application sweeping possibly-stale references catches `VertexNotFoundException`
+(or `RecordNotFoundException`) once and skips the entry, instead of matching on messages or on the accident of which
+read noticed first. An unreadable *chunk* of a vertex that does exist keeps the retryable
+`ConcurrentModificationException` and #5764's scoped advice, on both paths.
+
+[#6586](https://github.com/ArcadeData/arcadedb/issues/6586)
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3849,6 +3849,12 @@ keeps a conflict only for a record other than the vertex (a continuation chunk a
 a vertex that has vanished gets the same non-retryable answer as everywhere else, and the `ClassCastException` arm
 next to it no longer asserts a slot reuse this frame cannot establish.
 
+That last part is worth spelling out for anyone whose retry logic pattern-matches on the exception type: a
+*genuinely* concurrent delete - another transaction removing the vertex between the probe and that re-read, which is
+the one case the old message actually described - is now reported as `VertexNotFoundException` too, not as a
+retryable conflict. The outcome is unchanged, since a retry would only rediscover the absence at the probe on the
+next attempt and fail there; what changes is that it is reported once instead of after the budget is spent.
+
 The second is the append. `GraphEngine.getOrCreateEdgeList` read the head RID *outside* its `try`, so on a vertex
 whose record is gone the lazy load escaped raw as `RecordNotFoundException: Record #4:0 not found`. The verdict was
 right - it was never retryable - but nothing else was there: it did not say the missing record is a vertex, that it
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3861,6 +3861,11 @@ the next attempt and fail there, so what changes is that it is reported once rat
 A loop that wants the old shape should catch `RecordNotFoundException` (or `VertexNotFoundException`) alongside the
 conflict and treat it as "already gone" - which is what it would have concluded after exhausting its retries.
 
+One note for anyone reading causes closely in a log: when the absence is established by the delete's slot probe
+rather than by a failed read, the wrapped `RecordNotFoundException` is raised at the probe, so its top stack frame
+is that check and not a bucket read. The record really is missing either way; only the frame that established it
+differs.
+
 The second is the append. `GraphEngine.getOrCreateEdgeList` read the head RID *outside* its `try`, so on a vertex
 whose record is gone the lazy load escaped raw as `RecordNotFoundException: Record #4:0 not found`. The verdict was
 right - it was never retryable - but nothing else was there: it did not say the missing record is a vertex, that it
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3852,11 +3852,14 @@ keeps a conflict only for a record other than the vertex (a continuation chunk a
 a vertex that has vanished gets the same non-retryable answer as everywhere else, and the `ClassCastException` arm
 next to it no longer asserts a slot reuse this frame cannot establish.
 
-That last part is worth spelling out for anyone whose retry logic pattern-matches on the exception type: a
-*genuinely* concurrent delete - another transaction removing the vertex between the probe and that re-read, which is
-the one case the old message actually described - is now reported as `VertexNotFoundException` too, not as a
-retryable conflict. The outcome is unchanged, since a retry would only rediscover the absence at the probe on the
-next attempt and fail there; what changes is that it is reported once instead of after the budget is spent.
+**Upgrading:** a caller that wraps vertex deletes in its own `catch (ConcurrentModificationException)` retry loop
+should read this paragraph. A *genuinely* concurrent delete - another transaction removing the vertex between the
+probe and that re-read, which is the one case the old message actually described - is now reported as
+`VertexNotFoundException` too, not as a retryable conflict, so such a loop no longer catches it and the failure
+propagates instead of being retried. That is the intended outcome: the retry could only rediscover the absence on
+the next attempt and fail there, so what changes is that it is reported once rather than after the budget is spent.
+A loop that wants the old shape should catch `RecordNotFoundException` (or `VertexNotFoundException`) alongside the
+conflict and treat it as "already gone" - which is what it would have concluded after exhausting its retries.
 
 The second is the append. `GraphEngine.getOrCreateEdgeList` read the head RID *outside* its `try`, so on a vertex
 whose record is gone the lazy load escaped raw as `RecordNotFoundException: Record #4:0 not found`. The verdict was
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3838,7 +3838,11 @@ other failure that delete can produce.
 
 `deleteVertex` now settles it before the walk, by reading the bucket's slot rather than trusting a handle: the
 vertex record either has a slot or it does not, and a `VertexNotFoundException` naming `CHECK DATABASE FIX` is
-raised immediately instead of after a full traversal that is about to be rolled back. The check reads only the slot
+raised immediately instead of after a full traversal that is about to be rolled back. That probe checks
+`READ_RECORD` on the bucket, where the physical delete under it checks only `DELETE_RECORD` - which changes when a
+caller lacking the read permission is told, not whether it needs it: every route into a vertex delete already read
+the record from that same bucket, a lazy handle when the head-pointer read loads it and a materialised one when it
+was materialised. The check reads only the slot
 marker, so a vertex with a corrupt or truncated body is unaffected - it stays deletable, as #4420 and #4432 require -
 and it is not tolerated under `force` either, for the reason #6572 spells out. The re-read at the end of the delete
 keeps a conflict only for a record other than the vertex (a continuation chunk a concurrent commit is republishing);
--- a/engine/src/main/java/com/arcadedb/exception/VertexNotFoundException.java
+++ b/engine/src/main/java/com/arcadedb/exception/VertexNotFoundException.java
@@ -21,9 +21,15 @@ package com.arcadedb.exception;
 import com.arcadedb.database.RID;
 
 /**
- * A caller about to WRITE a vertex's edge list found the vertex RECORD ITSELF absent (#6572). The reference it
- * arrived through - an adjacency entry, a RID held across transactions, a traversal step - names a record that is
- * not there.
+ * A caller about to WRITE a vertex found the vertex RECORD ITSELF absent (#6572). The reference it arrived through -
+ * an adjacency entry, a RID held across transactions, a traversal step - names a record that is not there.
+ * <p>
+ * ONE type for that fact, whichever operation met it (#6586): REMOVING an entry from an edge list
+ * ({@code GraphEngine.getEdgeHeadChunkForWrite}), APPENDING one to it ({@code GraphEngine.getOrCreateEdgeList} -
+ * typically the far endpoint of an edge being created, which is not the vertex the caller named), or DELETING the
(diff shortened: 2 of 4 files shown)


lvca:
Round 2 follow-ups.

**Internal repair paths and `READ_RECORD`** - checked, and it resolves cleanly: no internal recovery/repair path constructs a `VertexInternal` and routes it through `GraphEngine.deleteVertex`. `GraphDatabaseChecker` (`:335`) and `DatabaseChecker` (`:809`, `:1493`) all remove records with a raw `bucket.deleteRecord(rid)`, deliberately - `DatabaseChecker:796` says why ("the record cannot be materialised"). The only callers of `deleteVertex` are the three in `LocalDatabase.deleteRecordNoLock`, each already holding a loaded handle. So the probe adds no requirement any caller did not already meet; no change needed.

**Type confusion after the probe** - agreed, and intentional: the probe answers "is there a slot", which is the only question that can be asked without following placeholder/chunk indirection, and following it is what would make a corrupt vertex undeletable again (#4420/#4432). "Is it still a vertex" stays where it was, in `checkEdgeListHeadsUnchanged`'s `ClassCastException` arm.

**Doc duplication nit** - applied in 70e0084: the rationale now lives only on `notFoundOnProbe`.

**Adjacent tests** - run locally against this branch, all green:

```
Issue6586MissingVertexDiagnosisTest      11
Issue5680VertexDeleteEdgeCollectionTest   9
Issue6572DanglingVertexReferenceTest      5
Issue5670EdgeDeleteDanglingBackRefTest    5
Issue4432CorruptVertexDeleteTest          1
Tests run: 31, Failures: 0, Errors: 0
```

plus the full `engine` module (13200 tests, 0 failures) and the `server` lane (458 + 157 + 822, 0 failures). The CI `build-and-package` job covers the same lanes.

lvca:
Round 4 follow-ups, all in 5caa712 (and the test count in the description corrected to 13).

- **`DELETE_TOLERATE_BROKEN_CHAIN` + a missing vertex** - good catch, it was untested. `theBrokenChainOptInDoesNotForceThroughAMissingVertexRecord` now runs the dangling-reference delete with the opt-in ON and pins that `VertexNotFoundException` still surfaces. The reason it cannot be absorbed is the one you traced: that fallback fires on `ConcurrentModificationException`, and since #6572 a missing vertex is not one. The claim is now asserted rather than reasoned.
- **`getOrCreateEdgeList`'s second catch arm** - agreed it is not reachable today, and I could not construct a fixture that reaches it honestly: a continuation chunk of a multi-page vertex body surfaces as `BrokenChunkChainException` or as the loader's own conflict (#6258), never as a bare not-found from that read. Rather than write a fixture that fakes the shape, the arm is now labelled as defensive, with the reason it *converts* rather than rethrows (an unidentified missing record on a path about to WRITE a list takes #5670's policy - retry, do not skip the write). Same treatment as the residual arm on the physical delete from round 2.
- **`READ_RECORD` in the release notes** - already there, added in e09f81f: the `docs/release-26.9.1.md` paragraph now says the probe checks `READ_RECORD` where the physical delete checks only `DELETE_RECORD`, and that this changes when a caller lacking it is told rather than whether it needs it.
- **`notFoundOnProbe` as a data carrier** - agreed it is unusual; it exists because #6572's test pins that the diagnosis wraps a not-found, and a probe has no thrown exception to wrap. Kept, with the javadoc saying exactly that.

Local runs on this branch: `Issue6586MissingVertexDiagnosisTest` 13, plus `Issue6572DanglingVertexReferenceTest`, `DeleteToleratesBrokenChainToggleTest` and `BrokenMultiPageRecordDeleteTest` - 25 tests, 0 failures.

lvca:
Round 5 follow-ups, in ba15f0a.

- **`vertexBucket == null` branch untested** - fixed. `aRidWhoseBucketIsGoneAnswersLikeARidWhoseSlotIsEmpty` holds a materialised handle across a `DROP TYPE`, asserts the bucket really is gone first (so the test cannot pass vacuously), then asserts the delete answers `VertexNotFoundException` and explicitly **not** `SchemaException` - which is the whole reason that branch exists rather than letting `getBucketById` raise. 14 tests now.
- **The probe's cost on the happy path** - agreed it deserved saying outside the code. The release notes now state it plainly: paid on every vertex delete, one slot-marker read on the page the delete is about to write, bought against a doomed full walk on the rare path and against a wrong diagnosis.
- **A caller that constructs a `VertexInternal` without ever reading it** - checked this in round 2 and it holds: the only callers of `GraphEngine.deleteVertex` are the three in `LocalDatabase.deleteRecordNoLock`, each already holding a loaded handle, and the repair paths that could construct one from a raw page view (`GraphDatabaseChecker:335`, `DatabaseChecker:809`/`:1493`) all bypass `deleteVertex` entirely with a raw `bucket.deleteRecord(rid)`. So there is no shape that regresses from "delete succeeds" to "permission denied".

Local: `Issue6586MissingVertexDiagnosisTest` 14, 0 failures; the adjacent delete/broken-chain classes green as reported earlier. CI `Java CI - test` is running the same lanes on this commit.

lvca:
Round 6 follow-ups, in 5e22a41.

**The `READ_RECORD` / `DELETE_RECORD` regression test** - added, and it lands on your option (b): the scenario is already impossible, and the test says why rather than asserting it. `aPrincipalWithoutReadPermissionCannotReachTheDeleteAtAll` installs a `SecurityDatabaseUser` granting `DELETE_RECORD` and nothing else, and pins that the principal is refused **obtaining the handle**, so it never reaches `deleteVertex` for the probe to refuse it. The second half grants `READ_RECORD` back with the same `DELETE_RECORD` grant and asserts the delete then succeeds - without that, the first half would only prove the stub user denies everything.

One incidental finding while writing it, worth recording but out of scope here: `DatabaseRID.asVertex(boolean)` catches every non-`RecordNotFoundException` and rewraps it as `RecordNotFoundException("Record #x:y not found")`, so a **permission denial** surfaces to a caller of `rid.asVertex()` as a missing record. That is pre-existing and unrelated to this change, but it is the same class of defect this PR is about (a failure reported as something it is not), so the test reads the refusal through `lookupByRID` and says so in a comment. Happy to file it separately.

**The `modify()` catch asymmetry** - real, and now explained in place rather than deferred to a follow-up. The two arms differ because their positions differ: the head read is the first thing the method does, so an unidentified missing record has to take #5670's policy for a list write that has not begun (retry, do not skip). By `modify()` the vertex has already been read successfully three lines up, so a not-found naming anything else is not a fact about this list, and re-typing it as a conflict on this list would be inventing a diagnosis - which is precisely what this PR exists to stop doing.

**`notFoundOnProbe`'s stack trace** - agreed, and accepted: the trace points at the probe, which is where the absence was established. The alternative (a second read purely to manufacture a thrown exception) would put an I/O site in the trace that did not decide anything.

Local: `Issue6586MissingVertexDiagnosisTest` 15 tests, and the whole `com.arcadedb.graph` package 566 tests, 0 failures.

lvca:
Round 7 follow-ups, in 896aa22.

**2. The probe's cost - measured, not argued.** 20k vertices in a chain (one edge each), page cache and JIT warmed, three runs:

```
PROBE_NS_PER_CALL=106  DELETE_NS_PER_VERTEX=7869  RATIO=1.35%
PROBE_NS_PER_CALL=100  DELETE_NS_PER_VERTEX=7729  RATIO=1.29%
PROBE_NS_PER_CALL=103  DELETE_NS_PER_VERTEX=7570  RATIO=1.36%
```

~100ns against ~7.7us, so about **1.3%** of a vertex delete - and that is the whole probe including its `checkPermissionsOnFile`, on a page that is about to be written anyway. The harness was a scratch file, deliberately not committed: a wall-clock assertion of this shape is exactly what CLAUDE.md forbids as a test, since a full-suite GC pause would decide it rather than the code.

**1. `vertexBucket` captured once.** Fair point, now written down at the call site. The assumption is that nothing between the probe and the physical delete can re-map the bucket, which holds because everything between them is this method's own edge walk - no user code, no DDL, and the before-delete events fired one frame up in `deleteRecordNoLock` have already run. Reusing it is also the better of the two failure modes: a bucket that really has gone is answered as "the RID names nothing" rather than as `getBucketById`'s `SchemaException` about an internal file id, which is the same call the `vertexBucket == null` branch already makes.

**4. Upgrade note.** Promoted from a technical paragraph to an explicit **Upgrading** callout in the release notes, including what such a caller should catch instead (`RecordNotFoundException`/`VertexNotFoundException` alongside the conflict, treated as "already gone" - which is what the loop would have concluded after exhausting its retries anyway).

**Test gap for the genuine race.** Agreed and now stated at the line itself: that arm is reachable only by a truly concurrent delete landing in a window a few instructions wide, which no test can hit deterministically, so it is covered by reasoning and the comment says so rather than leaving it looking like an oversight.

**3. The two "impossible" branches disagreeing on retryability.** Deliberate, and explained in the code as of 5e22a41: the head read is the first thing the method does, so an unidentified missing record there takes #5670's retry-rather-than-skip policy for a list write that has not begun; by `modify()` the vertex has already been read successfully, so re-typing an unrelated not-found as a conflict *on this list* would be inventing a diagnosis. If either ever becomes reachable, the position is still what decides it.

Local: `com.arcadedb.graph` 566 tests, 0 failures.

lvca:
Round 8 follow-ups, in 5f5c6d8.

**1. Stale `Bucket` reference across the walk** - you are right that the window is not zero-width on a super-node, and rather than prove it shut I removed the assumption: the physical delete now resolves the bucket again, null-tolerantly, so a bucket dropped underneath the delete still answers "the RID names nothing" instead of `getBucketById`'s `SchemaException` about an internal file id. It is a map lookup on a path that has just walked an adjacency list, which is not worth defending a captured reference over - and a delete written through a file that has since been closed or re-mapped is exactly the failure that would not announce itself.

**2. `deleteVertex(` callers repo-wide** - grepped across every module's `src/main/java`:

```
engine/.../LocalDatabase.java:1540  graphEngine.deleteVertex((VertexInternal) record, forceBrokenChainDelete);
engine/.../LocalDatabase.java:1548  graphEngine.deleteVertex((VertexInternal) record, true);
engine/.../LocalDatabase.java:1556  graphEngine.deleteVertex((VertexInternal) record, true);
engine/.../opencypher/.../DeleteStep.java:442/490   (its own private method, ends in vertex.delete())
```

All three real callers sit in `deleteRecordNoLock`, each already holding a loaded handle. The repair paths that *do* materialise records from a raw page view (`GraphDatabaseChecker:335`, `DatabaseChecker:809`/`:1493`) bypass `deleteVertex` entirely with `bucket.deleteRecord(rid)` - `DatabaseChecker:796` says why. No import or replication path calls it. So there is no caller that could newly meet a `SecurityException`.

**3. Internal `catch (ConcurrentModificationException)` near vertex deletes** - grepped repo-wide too. The only ones anywhere near a vertex delete are `LocalDatabase.java:1520`, `:1549` and `:1561`, which are the broken-chain force-retry arms already covered by `theBrokenChainOptInDoesNotForceThroughAMissingVertexRecord`. The rest are in `TransactionContext`, `LocalTransactionExplicitLock`, `DatabaseAsyncTransaction`, `EdgeLinkedList`, `BucketIterator`, `PageManager`, `LocalBucket` and `TimeSeriesShard` - none of them wrapping a vertex delete.

Local: `com.arcadedb.graph` 566 tests, 0 failures.

lvca:
Round 9's one suggestion is applied in aa5d3b5: the duplicated resolve-and-null-check is now a `vertexBucketOf(RID)` helper. It is still **called** twice rather than resolved once and carried - that is the freshness property the duplication existed for, and the javadoc says so explicitly so a future reader does not "optimise" the second call away.

Final state verified locally: the full `engine` module, 13204 tests, 0 failures.

lvca:
Round 10 - taking the four observations in turn, one applied and three deliberately not:

**3. Synthetic cause's stack frame** - applied, docs only (290d2e7). The release notes now tell an operator reading causes closely that when the absence was settled by the probe, the wrapped not-found carries the probe's frame rather than a bucket read's. Worth a line precisely because it is the kind of thing that costs someone an hour if they meet it cold.

**1. Comment density** - judged not worth changing. The register is the file's own: #5670, #5725, #5760, #5764 and #6572 all wrote their rationale inline at this length, and this change extends their reasoning rather than standing apart from it. Splitting the "why" between the code and a PR description that nobody will have open in three years is how the drift these comments exist to prevent starts. The one thing I did trim on your predecessor's advice was genuine duplication (round 2's `@param` restating `notFoundOnProbe`'s javadoc), which is a different problem.

**2. A future caller constructing a `VertexInternal` without reading it** - agreed there is no compile-time guard, and I do not think one is worth inventing for it. What exists instead is the failure mode: such a caller gets a `SecurityException` naming the user and the type, which is a loud, immediate, first-statement failure rather than a silent wrong answer. That is the right way for a new caller to find out it skipped a read.

**4. Double `vertexBucketOf` resolution** - confirmed intentional, and it is the reason the helper is a method called twice rather than a local resolved once. The javadoc says so in as many words so the second call does not get "optimised" away later.

Final state: full `engine` module 13204 tests, 0 failures; Codacy 0 issues; `com.arcadedb.graph` 566, and the adjacent delete/broken-chain/corrupt-vertex classes green throughout.

PR #6594, opened 2026-08-22
feat(ci): reject needs.<job>.outputs.<name> references the producer never declares

robfrank:
Follow-up to #6593, suggested in its review. Adds `check-workflow-needs-outputs.py`, which asserts that every `needs.<job>.outputs.<name>` expression names an output the referenced job actually declares.

**Stacked on #6593 and must merge after it.** This branch is based on `fix/4894-declare-image-tag-output`, because the check fails against `main` as it stands - which is the point.

## Why

GitHub Actions does not validate these references. An output no job declares is accepted as valid YAML and silently evaluates to the empty string at runtime, so the mistake produces a workflow that looks correct, parses correctly, and quietly passes nothing where a value was meant to go.

That is the bug #6593 fixes: `build-and-package` had no `outputs:` block while seven jobs passed `needs.build-and-package.outputs.image-tag` as `ARCADEDB_DOCKER_IMAGE`. It survived because every consumer was defensive in a way that hid it - `e2e-go` guards `!= ""`, `js-bolt-conformance` falls back with `||`, and two `e2e-js` suites hardcode `:latest` and never read the variable. Since the saved artifact carries only the `:latest` tag, those fallbacks all landed on the right image, so the suites were correct by luck. It took a newly added consumer strict enough to pass `""` through to Docker to surface it, as `invalid reference format`.

This is the same family as `check-workflow-artifact-deps.py`: valid-looking YAML that is silently wrong at runtime, where the failure mode is indistinguishable from working. It gets a sibling script, runs in the same `setup` job, and needs no new wiring.

## That the check is real, not decorative

Run against `main` without #6593's fix, it reproduces the defect independently - all seven consumers, each with its exact location:

```
mvn-test.yml: job 'java-e2e-tests' references needs.build-and-package.outputs.image-tag at
'steps[6].env.ARCADEDB_DOCKER_IMAGE', but job 'build-and-package' declares no such output
(declares: none). GitHub Actions resolves this to an empty string at runtime instead of failing.
```

With #6593's fix present it exits 0.

Five cases added to `test-ci-scripts.sh`, pinning both directions the way the existing suites do:

- rejects a reference to an output the producer never declares
- accepts the same workflow once the output is declared (same base fixture, one transformation, opposite result)
- rejects a reference to a name the producer does not declare, where an `outputs:` block exists but lacks it - the easier mistake to miss, because a reader skims past the block
- rejects a reference to a job that does not exist, which is the other half of the same typo
- ignores a reference into a reusable-workflow job

75 checks pass in the harness, up from 70.

## Deliberate boundaries

A job delegating to a reusable workflow (`jobs.<id>.uses:`) declares its outputs in the called workflow, which this script does not follow. References into such a job are skipped rather than reported: this check gates every other job, so a false violation here is expensive. The distinction is explicit in the code - `None` means "cannot resolve from here", an empty set means "declares nothing", and only the latter is a defect.

A `needs` reference to a job that does not exist at all is reported, since that is never valid.

### DIFF
--- a/.github/scripts/check-workflow-needs-outputs.py
+++ b/.github/scripts/check-workflow-needs-outputs.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+#
+# Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+# SPDX-License-Identifier: Apache-2.0
+#
+# Asserts that every `needs.<job>.outputs.<name>` expression names an output the referenced job
+# actually declares.
+#
+# GitHub Actions does not validate this. A reference to an output that no job declares is accepted
+# as valid YAML and silently evaluates to the empty string at runtime, so the mistake produces a
+# workflow that looks correct, parses correctly, and quietly passes nothing where a value was meant
+# to go.
+#
+# That is not hypothetical here. `build-and-package` had no `outputs:` block at all while SEVEN jobs
+# passed `needs.build-and-package.outputs.image-tag` as ARCADEDB_DOCKER_IMAGE. Every one of them
+# received "". It survived for as long as it did because the consumers were each defensive in a way
+# that hid it: e2e-go guards `!= ""`, js-bolt-conformance falls back with `||`, and two e2e-js
+# suites hardcode `arcadedata/arcadedb:latest` and never read the variable. Since the saved image
+# artifact carries only the `:latest` tag, those fallbacks all landed on the right image - so the
+# suites tested the correct thing by luck, and the documented intent (e2e-go's README: "the
+# branch-built image is docker loaded under the tag passed via ARCADEDB_DOCKER_IMAGE") had never
+# actually held. It took a newly added consumer, strict enough to pass "" straight through to
+# Docker, to surface it as `invalid reference format`.
+#
+# This is the same family as check-workflow-artifact-deps.py: valid-looking YAML that is silently
+# wrong at runtime, where the failure mode is indistinguishable from working.
+#
+# Known boundaries, both deliberate:
+#   - A job that delegates to a reusable workflow (`jobs.<id>.uses:`) declares its outputs inside
+#     the called workflow, which this script does not follow. References INTO such a job are
+#     skipped rather than reported, because this check gates every other job and a false violation
+#     here is expensive. Nothing in this repository needs it yet.
+#   - A `needs` reference to a job that does not exist at all is reported: that is never valid, and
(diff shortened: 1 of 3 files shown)


robfrank:
Both fixed in `e34bbc76d7`.

**PyYAML guard.** Verified the convention first: `check-workflow-artifact-deps.py:46-49` and `check-test-reporter-guards.py:42-45` both use exactly that shape, `# pragma: no cover` comment included. Mine had a bare import. Matched.

Proved the guard actually fires rather than assuming it, by running the script under a `MetaPathFinder` that raises on `import yaml`:

```
exit: 1
stderr: check-workflow-needs-outputs: PyYAML is required (pip install pyyaml)
```

Worth noting my first attempt at that probe used the legacy `find_module`/`load_module` hooks, which Python 3.12+ removed, so it exited 0 and looked like a pass when it had simply failed to block anything. Inconclusive, not positive - which is the same shape of mistake this whole PR exists to catch, one level up.

**No-op sed.** Correct, and doubly so: sed's pattern space carries no trailing newline, so a `\n` in the pattern can never match within a line, and `&` would be an identity substitution even if it did. The `awk` was doing all the work. Now feeding the file to `awk` directly.

I kept deriving the accepting fixture from the rejecting one rather than writing a second file, and said so in a comment: it is what makes the pair evidence rather than two independent assertions, since the added `outputs:` block is then the only difference between a pass and a failure. Harness still at 75 checks, and the check still exits 0 against this branch and 1 against `main` without #6593.

robfrank:
All three addressed in `8f4eda791d`, but the main one with one deliberate divergence from the suggestion, and it matters.

**The needs gap: real, fixed, and it must be DIRECT needs, not the closure.**

The gap is exactly as described - the script asked "does the job exist" and "does it declare the output", never "does the consumer depend on it". Fixed.

But the suggested fix, reusing `needs_closure()` from the artifact-deps script, would have been wrong here, and wrong in the permissive direction. Artifacts and the `needs` context do not behave the same way:

- An artifact is uploaded to the RUN, so a job that waits transitively still finds it. That is why the sibling script resolves a full closure, and it is correct there.
- The `needs` CONTEXT carries "the outputs of all jobs that are defined as a dependency of the current job". A job reaching the producer only transitively reads an empty string, exactly like a job that never declared the dependency.

So a closure-based check would have accepted this, which is precisely the reference the script exists to reject:

```yaml
build:   { outputs: { image-tag: ... } }
test:    { needs: build }
report:  { needs: test, env: { IMAGE: "\${{ needs.build.outputs.image-tag }}" } }   # resolves to ""
```

That case is now a test ("rejects a transitively-reached producer, because the needs context is direct-only"), and the reasoning is in `needs_of`'s docstring so the next person does not "fix" it into a closure.

I also duplicated the four-line `needs_of` rather than importing it across scripts. Both files are standalone executables run by path from the workflow; making one import the other buys four lines and adds a module-resolution failure mode to two checks that gate every other job.

**Proved it fires against the real workflow, not only fixtures.** Removing `needs: build-and-package` from `client-smoke-test` while leaving the reference intact:

```
job 'client-smoke-test' references needs.build-and-package.outputs.image-tag at
'steps[4].env.ARCADEDB_DOCKER_IMAGE', but 'client-smoke-test' does not list 'build-and-package'
in its needs (needs: nothing). The needs context carries only DIRECT dependencies, so this
resolves to an empty string at runtime instead of failing.
```

Restored, and green again.

**Nit 1, the \${{ }} gate: taken.** `references()` now extracts `\${{ }}` blocks first, matching the sibling's `EXPRESSION` regex. A `run:` step that merely prints those words is no longer a violation, and that is a test too.

**Nit 2, duplicate lines: taken.** One line per distinct `(job, output)` per consuming job. The same bad reference in both `env:` and `run:` was making one typo look like two.

Harness at 78 checks, up from 75. The check still exits 0 against current `main`, which independently confirms your observation that all seven consumers do carry the direct dependency.

PR #6595, opened 2026-08-22
fix(#6592): match composite index prefixes in the native Select query builder

### LINKED ISSUE #6592: Indexes are not used for SELECTs
I'm using an `orderBy()` in the native query builder and query times are unexpectedly long even when the exact index is available. 

Here is a simplified version of the index and query I'm using:

``` 
CREATE VERTEX TYPE Supplier IF NOT EXISTS;
CREATE PROPERTY Supplier.key1 IF NOT EXISTS STRING;
CREATE PROPERTY Supplier.key2 IF NOT EXISTS STRING;
CREATE PROPERTY Supplier.ordered_at IF NOT EXISTS DATETIME;

CREATE INDEX ON Supplier (key1, key2, ordered_at)

SELECT FROM Supplier
WHERE key1 = 'a'
AND key2 = 'b'
ORDER BY ordered_at DESC
LIMIT 1
```

Or in Java:

```
final var query = db.database.select().fromType("Supplier")
        .where().property("key1").eq().parameter("key1")
        .and().property("key2").eq().parameter("key2")
        .limit(1)
        .orderBy("ordered_at", false)
        .compile();
try (final var vertices = query
        .parameter("key1", "a")
        .parameter("key2", "b")
        .vertices()) {
    return vertices.nextOrNull();
}
```

Debugging led me to `SelectExecutor.lookForIndexes()` which returns `null` for my query:

```
boolean canUseIndexes = isTheNodeFullyIndexed(select.rootTreeElement);
```

Indexes in `LocalDocumentType.indexesByProperties` use the list of indexed properties for the key and only `select.rootTreeElement` has all the properties which will match the proper key.

I believe that `SelectExecutor.isTheNodeFullyIndexed()` only calls `LocalDocumentType.getPolymorphicIndexByProperties()` with single property names and never with the list of property names received on the initial call (`select.rootTreeElement`).

---

I have the `ordered_at` property used in the `orderBy` as part of the index, but I have to either:
- remove it from the index if I wish that `select.rootTreeElement` will match the index in `LocalDocumentType.getPolymorphicIndexByProperties()` since `ordered_at` is not in the `select.rootTreeElement`
- add a dummy `.and().property("ordered_at").eq().parameter("ordered_at")` so that it is included in the `select.rootTreeElement` and matches all the properties of the index

Version: 26.8.1 (latest release)

lvca:
## Summary

Fixes #6592.

`SelectExecutor.isTheNodeFullyIndexed()` (the native `com.arcadedb.query.select.Select` fluent query builder, *not* the SQL executor) only ever looked up an index for one property name at a time. A composite (multi-property) index is registered under its **full** property list (`LocalDocumentType.indexesByProperties`), never under any subset of it, so an AND-conjunction of equality conditions covering only the **leading** properties of a composite index (e.g. `key1 = 'a' AND key2 = 'b'` against an index on `(key1, key2, orderedAt)`) was never recognized and silently fell back to a full type scan.

I verified independently that the SQL executor (`SelectExecutionPlanner.buildIndexSearchDescriptor`/`fullySorted`) already handles this correctly for plain `SELECT` statements - prefix matching and ORDER BY elision by the trailing index column both already work there. The bug is confined to the native query-builder path.

### Fix

`SelectExecutor` now tries a composite-index prefix match first, for a pure AND-conjunction of equality leaves (no `or`/`not` - those keep going through the existing per-leaf path unchanged):

- Every composite index (2+ properties) on the type is checked for how many of its **leading** properties are covered by an `eq` leaf in the where-tree; the longest such prefix wins (ties broken toward a unique index).
- The matched prefix becomes a partial-key cursor: `index.get(keys)` when the prefix covers the whole index, `index.range(ascending, keys, true, keys, true)` for a genuine partial-key scan (a partial key is a *prefix*, not an exact key - `get()` only returns every match when the key's arity matches the index's own exactly).
- When the trailing (first unmatched) index property is also the query's sole `ORDER BY` column, the range scan direction is chosen to match it, so the index itself already returns rows in the requested order - `SelectIterator.fetchResultInCaseOfOrderBy()`'s full materialize-and-sort fallback is skipped, exactly like the reported `ORDER BY ordered_at DESC LIMIT 1` case.
- The `#6565` candidate cap (`skip + limit`) is only applied when the matched prefix consumes the *entire* where-tree and the order-by, if any, is satisfied by the cursor itself - otherwise it runs uncapped, same safety invariant as the existing single-property path.

Single-property indexes and any tree containing `or`/`not` are left entirely to the pre-existing `isTheNodeFullyIndexed()`/`filterWithIndexes()` logic - no behavior change there.

## Test plan

- [x] New regression test `SelectCompositeIndexTest` (TDD - written and confirmed failing before the fix):
  - composite index prefix (`key1 AND key2`) is used instead of a full type scan
  - the exact reported scenario: prefix equality + `ORDER BY <trailing column> DESC LIMIT 1` returns the correct row while only evaluating that one candidate (no full scan, no in-memory sort)
  - a shorter prefix (`key1` alone) still uses the composite index
  - an `or` over the same two properties (no standalone single-property index) still correctly falls back to a full scan, unchanged from before
- [x] `mvn -o -pl engine -am test -Dtest='Select*Test'` (302 tests, includes all of `com.arcadedb.query.select.*` plus the related SQL executor select suites) - all pass
- [x] `mvn -o -pl engine -am test -Dtest='Issue6565SelectIndexCandidateLimitTest'` (the candidate-cap safety test this change's cap logic must not regress) - all 20 pass
- [x] `mvn -o -pl engine -am compile` - clean

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -215,20 +215,30 @@ public class SelectExecutor {
     if (select.fromType != null && select.rootTreeElement != null) {
       final List<IndexCursor> cursors = new ArrayList<>();
 
-      // FIND AVAILABLE INDEXES AND ASSIGN node.index ON EVERY INDEXED LEAF: filterWithIndexesFinalNode() RELIES ON THAT
-      // SIDE EFFECT TO KNOW WHICH LEAVES CAN BECOME A CURSOR
-      isTheNodeFullyIndexed(select.rootTreeElement);
+      // #6592: A COMPOSITE (MULTI-PROPERTY) INDEX IS REGISTERED UNDER ITS FULL PROPERTY LIST (SEE
+      // LocalDocumentType.indexesByProperties), NOT UNDER ANY SUBSET OF IT - SO A PLAIN AND-CONJUNCTION OF EQUALITY
+      // LEAVES THAT ONLY COVERS THE LEADING PROPERTIES OF SUCH AN INDEX CAN NEVER BE FOUND BY THE PER-LEAF,
+      // SINGLE-PROPERTY LOOKUP BELOW (isTheNodeFullyIndexed()). TRY A PREFIX MATCH AGAINST EVERY COMPOSITE INDEX ON
+      // THE TYPE FIRST; ONLY FALL BACK TO THE SINGLE-PROPERTY PATH WHEN NO SUCH INDEX COVERS ANY LEADING PROPERTY.
+      final boolean compositeIndexUsed = matchCompositeIndex(cursors);
 
-      // #6565: A CANDIDATE CAP IS SAFE ONLY WHEN THE INDEX SCAN EXACTLY REPRODUCES THE WHERE-TREE'S RESULT SET AND
-      // THE ORDER BY (IF ANY) IS ALREADY SATISFIED BY IT - evaluateWhere(), skip AND fetchResultInCaseOfOrderBy()'s
-      // FULL-DRAIN SORT ALL REDUCE THE STREAM FURTHER OTHERWISE, SO THE SCAN MUST RUN UNCAPPED TO SURVIVE THAT.
-      // MUST RUN BEFORE filterWithIndexes() BELOW, SINCE soleExactLeaf() READS node.index BEFORE
-      // filterWithIndexesFinalNode() PRUNES AN 'and' SIBLING'S - HARMLESS TODAY ONLY BECAUSE 'and' IS ALREADY
-      // UNCONDITIONALLY DISQUALIFIED
-      final SelectTreeNode exactLeaf = soleExactLeaf(select.rootTreeElement);
-      indexCandidateLimit = exactLeaf != null && isOrderBySafeForCap(exactLeaf) ? computeExactCandidateLimit() : -1;
+      if (!compositeIndexUsed) {
+        // FIND AVAILABLE INDEXES AND ASSIGN node.index ON EVERY INDEXED LEAF: filterWithIndexesFinalNode() RELIES ON THAT
+        // SIDE EFFECT TO KNOW WHICH LEAVES CAN BECOME A CURSOR
+        isTheNodeFullyIndexed(select.rootTreeElement);
 
-      filterWithIndexes(select.rootTreeElement, cursors);
+        // #6565: A CANDIDATE CAP IS SAFE ONLY WHEN THE INDEX SCAN EXACTLY REPRODUCES THE WHERE-TREE'S RESULT SET AND
+        // THE ORDER BY (IF ANY) IS ALREADY SATISFIED BY IT - evaluateWhere(), skip AND fetchResultInCaseOfOrderBy()'s
+        // FULL-DRAIN SORT ALL REDUCE THE STREAM FURTHER OTHERWISE, SO THE SCAN MUST RUN UNCAPPED TO SURVIVE THAT.
+        // MUST RUN BEFORE filterWithIndexes() BELOW, SINCE soleExactLeaf() READS node.index BEFORE
+        // filterWithIndexesFinalNode() PRUNES AN 'and' SIBLING'S - HARMLESS TODAY ONLY BECAUSE 'and' IS ALREADY
+        // UNCONDITIONALLY DISQUALIFIED
+        final SelectTreeNode exactLeaf = soleExactLeaf(select.rootTreeElement);
+        indexCandidateLimit = exactLeaf != null && isOrderBySafeForCap(exactLeaf) ? computeExactCandidateLimit() : -1;
+
+        filterWithIndexes(select.rootTreeElement, cursors);
+      }
+
       if (!cursors.isEmpty())
         return new MultiIndexCursor(cursors, indexCandidateLimit, true);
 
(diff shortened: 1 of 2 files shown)


lvca:

- **Test coverage gap**: added a 5th case in `SelectCompositeIndexTest` binding all three properties of the composite index by equality, exercising the `get()` branch (`fullKeyMatch == true`) that the other cases didn't reach.
- **Non-deterministic tie-break**: `matchCompositeIndex()` now sorts candidate indexes by name before comparing prefix length/uniqueness, since `getAllIndexes(true)` is backed by a `HashSet` for a polymorphic type and its iteration order is otherwise unspecified.

Not implementing, per the review's own framing ("not a bug", "not worth blocking on"):

- **Selectivity vs. composite-prefix tradeoff**: correct observation, but comparing the selectivity of a composite-prefix cursor against an alternative single-property index on a more selective leaf would need cost/cardinality estimation that doesn't exist anywhere in this native query builder today. That's a much larger, separate piece of work than this bug fix (making the composite index usable at all) and out of scope here.

All 323 tests in the affected suites (`Select*Test`, `Issue6565SelectIndexCandidateLimitTest`) pass after both fixes.

lvca:

- **Performance**: `matchCompositeIndex()` now filters to composite (2+ property) indexes in a single pass and returns early - before touching the where-tree at all - when the type has none. A type with only single-property indexes (the common case) no longer pays the where-tree walk or the candidate-list allocation/sort.
- **Test coverage gaps**: added the three suggested cases:
  - `compositeIndexPrefixMatchWithSkipAndLimit` - the `#6565` candidate cap combined with `skip`/`limit` on the order-by-elided composite scan.
  - `compositeIndexPrefixWithTrailingRangeAndLimitStaysCorrect` - a trailing non-equality condition keeps `exactMatch` false (cap disabled) under a `LIMIT`; this one actually catches a real regression if the cap logic were wrong (it would silently return zero rows instead of the correct two).
  - `compositeIndexTieBreakPrefersUniqueIndex` - two composite indexes tied on prefix length correctly break toward the unique one.

Not implementing, per the review's own framing ("not something to fix in this PR"):

- **Shared helper across native builder / openCypher `MatchNodeStep` / SQL executor's `IndexSearchDescriptor`**: agreed this is real duplication worth unifying eventually, but it's a separate, larger refactor spanning three subsystems that already behave slightly differently (e.g. polymorphic vs. non-polymorphic index lookup) - out of scope for this bug fix and better handled as its own follow-up with explicit design review.

All 326 tests in the affected suites (`Select*Test`, `Issue6565SelectIndexCandidateLimitTest`) pass.

lvca:
This one was a real, confirmed regression, not a judgment call - thanks for catching it.

Verified directly against the source: `HashIndex.supportsOrderedIterations()` (and `FULL_TEXT`/`GEOSPATIAL`/`LSM_VECTOR`/`LSM_SPARSE_VECTOR`) return `false`; only `LSMTreeIndex` returns `true`. `TypeIndex.range()` throws `UnsupportedOperationException` when that's `false`, and `matchCompositeIndex()` had no check on index type before calling it for a partial-prefix match - so a composite `UNIQUE_HASH` index on `(@out, @in)` (exactly the shape `Issue5677HashIndexLinkKeyTest` already exercises for edge uniqueness) would turn a previously-working full scan into a hard failure the moment only one endpoint was bound.

Fixed by restricting `candidates` to indexes where `supportsOrderedIterations()` is true, so a type with only non-ordered composite indexes now falls straight through to the pre-existing single-property path, unchanged from before this PR. This also resolves the minor `FULL_TEXT`/`get()` semantic-mismatch note for free, since it's excluded the same way. Added the suggested regression test (`compositeHashIndexPartialPrefixFallsBackToScan`) with a composite `HASH` index and a partial-prefix query - it throws before the fix and passes after. Also added the requested comment on why the tie-break comparison can't see a null `bestIndex`.

All 337 tests in the affected suites pass, including `Issue5677HashIndexLinkKeyTest` (unaffected) and the new regression case.

lvca:
No blocking issues in this round, thanks for confirming the HASH-index fix.

Added the two suggested tests:
- `compositeIndexTieBreakFallsBackToAlphabeticalOrder` - two non-unique composite indexes tied on prefix length, observed via whether `ORDER BY` on the alphabetically-first index's trailing property gets elided.
- `compositeIndexMultiColumnOrderByDisablesElision` - a multi-column `ORDER BY` with rows tied on the primary column, confirming they still come out sorted by the secondary column (which only a full sort, not a bare index scan, guarantees).

Not implementing the other two suggestions, both of which you yourself framed as non-blocking/speculative:

- **Caching the filtered/sorted composite-index list per `DocumentType`**: real gain, but it needs a new schema-layer cache with its own invalidation-on-schema-change logic (mirroring `indexesByProperties`) - that's meaningfully more surface area and risk than the per-query iterate+sort it would save, for a type's typically small index count. Better suited to a dedicated follow-up if it ever shows up as measurably hot.
- **Composite prefix losing to a hypothetical more-selective single-property unique index on the same leading property**: same class of concern as the round-1 selectivity note - no cost-based index selection exists in this native builder to make that comparison, and it's not a correctness issue.

All 339 tests in the affected suites pass, including the two new cases.

lvca:
This one is a legitimate, concrete regression scenario rather than a speculative selectivity concern - agreed it needed fixing, thanks for the precise scenario and the two suggested directions.

Went with the second (narrower, lower-risk) direction: when the composite match isn't a full-key one, check whether any eq-bound property outside the matched prefix already has its own standalone index; if so, defer entirely to the pre-existing `isTheNodeFullyIndexed()`/`filterWithIndexes()` path, which already correctly picks between competing single-property indexes (preferring unique). A property already inside the matched prefix is exempt - the prefix scan is at least as selective for it as a standalone index alone would be, so there's no reason to give up a good composite match over it.

Added `compositeIndexPartialPrefixDefersToStandaloneIndexOnUnmatchedProperty`, mirroring your scenario: a non-unique composite index `(key1, key2, key3)` coexisting with a standalone unique index on `key3`, querying `key1='a' AND key3='specific'` with `key2` unbound. Verified by hand that none of the 11 pre-existing composite tests trip the new deferral check (each either has every eq-bound property inside its matched prefix, or reaches a full-key match, which skips the check) - confirmed by all 12 tests in the suite (11 existing + 1 new) passing.

Also softened the upfront-guard comment per the minor nit - it no longer claims zero cost, since every query still pays one `getAllIndexes(true)` iteration.

All 340 tests in the affected suites pass.

lvca:
No blocking issues in this round, thanks for the thorough trace of the tree-shape and range()/get() aliasing details.

Added the requested test: `compositeIndexPrefixMatchWithParameterBinding`, exercising `matchCompositeIndex()`'s `SelectParameterValue` branch with `.eq().parameter(...)` instead of `.value(...)`.

Not implementing the other two, both of which you flagged as non-blocking / "not a request to change it now":

- **Standalone-index deferral only considers `eq`-bound unmatched properties, not range/`in_op`**: agreed, explicitly scoped limitation given this PR targets pure-eq prefixes - a good candidate for a follow-up if it turns out to matter in practice, not part of this fix.
- **`matchCompositeIndex()` could be split into smaller named steps**: fair maintainability point, but a pure refactor with no behavior change is better done separately from a bug-fix PR that's already accumulated several rounds of review-driven edits.

All 341 tests in the affected suites pass, including the new case.

lvca:
This is a real refinement worth making before merge, thanks.

Gated the round-5 deferral on the standalone index actually being unique: a non-unique standalone index on an unmatched property gives no guarantee of being more selective than the composite prefix's own combined leading properties, and there's no cost-based way in this method to compare the two - so deferring on it alone could throw away a genuinely better composite match. A property already inside the matched prefix is still exempt either way, unchanged.

Added `compositeIndexPartialPrefixKeepsMatchWhenUnmatchedStandaloneIndexIsNotUnique`, the mirror of the existing deferral test: same shape, but the standalone index on `key3` is non-unique - the composite prefix on `key1` is now correctly kept (every `key1='a'` row evaluated) instead of deferring.

Not implementing the other two, both non-blocking:

- **Dead `if (cursor == null) return false;` checks after `get()`/`range()`**: agreed neither call can actually return `null`, but the pre-existing `filterWithIndexesFinalNode()` in this same file has the identical defensive check after the identical `TypeIndex` calls - removing it only from the new code would be an inconsistency with the file's own established style for no real benefit.
- **Comment density**: you noted yourself it "earns its keep" given the subtlety here, so no change.

All 342 tests in the affected suites pass, including the new case.

lvca:
No blocking issues in this round - appreciate the independent re-verification of the earlier fixes against the underlying index implementation.

Added the suggested coverage: `compositeIndexNotUsedWithNotInTheConjunction`, hand-building the tree the same way `Issue6565SelectIndexCandidateLimitTest.notLeafIsNeverTreatedAsExactlyIndexed` already does (`SelectOperator.not` has no fluent entry point on the `Select` builder). Pins that `isPureAndConjunction()` rejects a `not()` appearing anywhere in the conjunction the same way it already rejects `or()`.

The other three observations were repeats of judgments already made in earlier rounds (non-cost-based heuristics, the dead null checks kept for consistency with `filterWithIndexesFinalNode`'s existing style, comment density) - no new action needed there.

All 343 tests in the affected suites pass, including the new case.

lvca:
No blocking issues, no new test gaps ("no obvious gap") - thanks for the thorough independent trace through the whole call chain (SelectIterator, MultiIndexCursor, TypeIndex/LSMTreeIndexCursor).

The three observations here are repeats of judgments already made and explained in earlier rounds:
- Non-cost-based longest-prefix-wins heuristic - explicitly documented as a deliberate limitation, not something to fix in this bug-fix PR.
- Standalone-index deferral only considers single-property indexes, not a composite standalone index covering several leftover properties - reasonable scope cut, candidate for a follow-up if it comes up in practice.
- Comment density - matches the file's existing style around the `#6565` cap logic, not something to trim.

Also noted: the `LocalDocumentType.getAllIndexes(true)` class-comment inaccuracy about which case is unordered (harmless here since the sort-by-name fallback covers both) is a pre-existing doc nit unrelated to this PR's code, not something to fix here either.

Waiting on `build-and-package` (unlocks the full CI test matrix) before this is ready to merge.

lvca:
CI status update: all lanes are green except two, both confirmed unrelated to this PR's change:

- **`vector-unit-tests`**: `LSMVectorIndexRecoveryTest.vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild` timed out after 11 minutes waiting for "the inactivity rebuild drains the delta buffer" (`org.awaitility.core.ConditionTimeoutException`). Entirely within the vector-index background-rebuild subsystem (`com.arcadedb.index.vector.LSMVectorIndex`), which this PR never touches. Retried the job.
- **`ha-integration-tests`**: 6 failures/errors, all in `com.arcadedb.server.ha.raft.*` (Raft clustering/slot-merge integration tests: `Issue5381SlotMergeRaftIT`, `Issue5569SlotMergeDeleteRaftIT`, `RaftHTTPGraphConcurrentIT`, `RaftPriorityRejoinIT`, `RaftTimeSeriesWriteReadYourWritesIT`, `RemoteStickyConnectionStrategyIT`). Entirely within HA/Raft clustering, unrelated to the native `Select` query builder this PR changes.

Neither failure touches `com.arcadedb.query.select.SelectExecutor` or anything this PR modifies. All 302+ tests in the directly affected suites (`Select*Test`, `Issue6565SelectIndexCandidateLimitTest`, `Issue5677HashIndexLinkKeyTest`) pass locally and in CI (`unit-tests`, `integration-tests`, `slow-unit-tests`, `build-and-package`, all 9 `claude-review` rounds, all `Analyze`/CodeQL/Codacy/Meterian lanes).

PR #6604, opened 2026-08-23
fix(opencypher): MERGE mid-query no longer inflates a chained MATCH's cardinality

### LINKED ISSUE #6602: ArcadeDB 26.8.1: `collect()`/`UNWIND` changes cardinality after a zero-length `MATCH` followed by `MERGE`
## Environment

- ArcadeDB: `26.8.1` (`arcadedata/arcadedb:26.8.1`)
- Protocol: Bolt
- No plugins or external files are required.

## Reproduction setup

Run this once in a new empty database:

```cypher
UNWIND range(1, 15) AS i
CREATE (:Seed {id: i});
```

The setup creates exactly 15 nodes and no relationships.

## Query A

```cypher
UNWIND ['D', 'eUu'] AS alias0
MATCH (), p0 = ()
WHERE NOT ((alias0 + alias0) >= 'a')
MERGE (:l4 {id: 128})
RETURN {alias0: alias0, p0: p0} AS row;
```

## Query B

Run the same setup in a fresh database, then run the semantically equivalent
materialized form:

```cypher
UNWIND ['D', 'eUu'] AS alias0
MATCH (), p0 = ()
WHERE NOT ((alias0 + alias0) >= 'a')
MERGE (:l4 {id: 128})
WITH {alias0: alias0, p0: p0} AS row
WITH collect(row) AS rows
UNWIND rows AS row
RETURN row;
```

## Expected result

Both queries should return 225 rows. The filter keeps only `alias0 = 'D'`,
and `MATCH (), p0 = ()` produces `15 * 15 = 225` rows. `MERGE` creates or
reuses one node but must not change the number of rows already produced by
the preceding `MATCH` clauses.

## Actual result

On two independent fresh Bolt containers, the results are deterministic:

| Query | Rows | Final graph |
| --- | ---: | --- |
| A | 226 | 16 nodes, 0 relationships |
| B | 225 | 16 nodes, 0 relationships |

lvca:
## Summary

Fixes #6602.

A chained `MATCH` node pattern with no per-row-varying constraint (e.g. the
anonymous `()` in `MATCH (), p0 = ()`) re-opened its full vertex scan by
calling `database.iterateType()` live on every per-row re-open. Since a
chained MATCH re-opens that scan once per outer input row, a re-open
happening after a downstream `MERGE` had already created a new vertex - of a
brand-new type or of one that already existed - would see that vertex and add
a spurious extra row, retroactively inflating the MATCH clause's own
cardinality after WHERE/MERGE had already started consuming its earlier rows.

This is a genuine ArcadeDB bug, not a user misunderstanding of Cypher: in
Neo4j/openCypher semantics MERGE runs once per input row from the preceding
clauses and must not retroactively change the cardinality those clauses
already produced (Neo4j guards against exactly this class of hazard with an
"Eager" plan operator).

### Root cause

- `MatchNodeStep`'s chained-mode fetch loop re-opens its vertex iterator once
  per outer input row (`getVertexIterator(currentInputResult)`), whenever the
  current pattern variable isn't already bound.
- For a pattern with no per-row-varying constraint, the resulting scan target
  is always the same set of vertices - but it was still being recomputed live
  via `database.iterateType()` on every re-open, so a vertex created
  mid-query by a downstream `MERGE` could be picked up by a later re-open.
- Because the step pipeline is pull-based and every step does an internal
  read-ahead to decide whether it's exhausted, this re-open can legitimately
  happen interleaved with a downstream `MERGE`'s write, not only after the
  whole query completes.

### Fix

`MatchNodeStep` now materializes its full candidate set once per step
execution and replays it on every later re-open, whenever the pattern is
provably independent of the current input row (no ID filter, no dynamic
labels, and - for a single-label pattern - no properties or inline WHERE
pushdown that could route to a row-dependent index/partition lookup). This is
scoped to chained steps only (`prev != null`); a standalone MATCH already
calls its vertex iterator exactly once, so caching there would only add
eager-materialization cost with no correctness benefit - and as a side
effect, it turns what used to be an O(N) full-type-scan per outer row into a
single O(N) resolve reused across all outer rows for chained/cartesian
patterns.

## Test plan

- [x] New regression test `Issue6602MergeAfterEmptyMatchCardinalityTest`
  covering: the exact issue repro (225 vs 226 rows), the `collect()`+`UNWIND`
  materialized form from the issue's workaround, MERGE creating a vertex of
  an *already-existing* type (not just a brand-new type), a multi-label
  pattern with properties (to confirm the broader fix generalizes correctly),
  and a standalone (non-chained) MATCH staying unaffected.
- [x] Full `com.arcadedb.query.opencypher.**` test package: 8841 tests, 0
  failures, 0 errors (98 skipped/tagged).

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/MatchNodeStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/MatchNodeStep.java
@@ -90,6 +90,19 @@ public class MatchNodeStep extends AbstractExecutionStep {
   // need to move to a per-execution scope or be guarded.
   private       String              usedIndexName; // Track which index was used (if any)
   private       String              usedPartitionBucket; // Track partition bucket pruning (if any) - same write-once-per-execution contract as usedIndexName
+  // Fully materialized snapshot of a row-independent full-type-scan's candidates, resolved on the first
+  // getVertexIterator() call of a CHAINED match (prev != null) and replayed for every later re-open instead
+  // of re-opening a live database.iterateType() cursor per outer input row (a chained MATCH, e.g.
+  // MATCH (), p0 = (), re-opens that scan once per outer input row). Without this, a vertex a downstream
+  // MERGE/CREATE creates mid-query - whether of a brand-new type or of a type that already existed - is
+  // visible to a later re-open and retroactively grows this MATCH's cardinality after earlier rows already
+  // reached WHERE/MERGE (issue #6602). Only enabled for a pattern this class can prove is 100% independent of
+  // the current input row: no ID filter (static or dynamic), no dynamic labels, and (for a single-label
+  // pattern) no properties or inline WHERE pushdown that could route to a row-dependent index/partition
+  // lookup - see {@link #isRowIndependentFullScan()}. Scoped to prev != null: a standalone MATCH calls
+  // getVertexIterator() exactly once regardless, so caching there would only add eager materialization cost
+  // (and memory) for large scans with no correctness benefit.
+  private       List<Identifiable>  cachedFullScanCandidates;
 
   /**
    * Creates a match node step.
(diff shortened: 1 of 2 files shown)


lvca:
Thanks for the review - addressed in db6b7bbef.

**1. Stale cached RIDs from a downstream DELETE (the main point)**: confirmed and fixed. Added a `RecordNotFoundException` catch around the record load in `MatchNodeStep`'s chained-mode fetch loop, matching the pattern already used by `FetchFromRidsStep`/`MatchRelationshipStep`/etc. Added a regression test (`chainedMatchToleratesDownstreamDeleteOfACachedCandidate`) reproducing the exact hazard - `m` scans a disjoint type from `n` so the test isolates the cache-staleness mechanism from an unrelated pre-existing issue I hit while designing it: if `m` and `n` scan the *same* type, `m`'s DELETE can remove the very vertex `n` is bound to, and `RETURN n.id` then throws when lazily loading `n`'s properties - that's a separate, pre-existing gap in property-access evaluation (not in `MatchNodeStep`, and not something this PR's cache introduces), out of scope here.

**2. Test coverage gap - genuine single-label chained pattern**: added `singleLabelChainedPatternWithNoPropertiesIsAlsoCacheable`. Turned up a real bug while writing it: `isRowIndependentFullScan()` was disqualifying caching for *any* single-label pattern with a non-null `whereFilter`, even one that doesn't reference the pattern's own variable at all (e.g. a WHERE clause pushed down that only touches an UNWIND alias). Replaced the blunt `whereFilter != null` check with a structural mirror of `extractEqualityPredicates`'s own predicate recognition, so only a WHERE equality that could actually route `computeVertexIterator` to a row-dependent index lookup disqualifies caching.

**3. Test coverage gap - negative test for single-label+properties correctly excluded**: I decided not to add this one. Constructing a deterministic assertion for "this pattern is NOT cached, so it sees a mid-query INSERT" runs into the same MERGE-must-run-before-any-match-exists chicken-and-egg problem (a cartesian branch with zero initial matches never lets MERGE execute at all), and any workaround becomes dependent on batch-size/interleaving timing that isn't a stable thing to assert on. The exclusion logic itself is directly readable in `isRowIndependentFullScan()`, and the full `com.arcadedb.query.opencypher` suite (8843 tests) - which already exercises single-label+property/WHERE-driven index and partition-pruning behavior extensively - continues to pass unchanged, which is the practical evidence that this branch's original per-row-dynamic behavior is untouched.

**4. Minor comments**: added the memory/CPU tradeoff note and extended the "not safe to share across concurrent MATCH branches" caveat to cover `cachedFullScanCandidates`.

Full `com.arcadedb.query.opencypher` package: 8843 tests, 0 failures after these changes.

lvca:
Thanks - the LIMIT/eager-materialization point was real, fixed in 02665e528.

**Confirmed the caching-safety analysis** (plan cache hits, correlated subqueries) - good to have that pinned down explicitly.

**Eager materialization defeating LIMIT**: agreed, and fixed via `recordingIterator` roughly as you suggested - the first pass now stays fully lazy/streaming, recording each element only as the caller actually consumes it. The recording gets promoted to `cachedFullScanCandidates` only once that first pass reports `hasNext() == false` (fully exhausted), which is also exactly the point where a second `getVertexIterator()` call becomes possible - so a query whose scan is opened once (LIMIT satisfied within the first outer row, or just no second outer row at all) now pays exactly the pre-fix cost, with the cache never populated.

I want to be upfront about the test I did *not* end up adding for this, rather than ship something misleading: I tried a plan-based row-count assertion first (`PROFILE` + `MatchNodeStep`'s `rowCount`), but that metric turns out not to observe the regression at all - `rowCount` and `cost` are only incremented around the *outer* fetch loop's `iterator.next()` calls (i.e. what's pulled back out), not around `getVertexIterator()`'s internal materialization, so the old eager-drain code and the new lazy code produce an identical row-count figure regardless of dataset size. I verified this empirically by temporarily reverting to the eager version and confirming the test still passed - it would have been a vacuous regression guard. I then tried a wall-clock comparison (500K-row dataset, `LIMIT 3`): eager vs. lazy differ by only ~50-100ms out of ~200ms total, which isn't a reliable enough margin to assert on without real flake risk (this codebase's underlying `database.iterateType()` cursor is cheap enough per-record that draining it eagerly isn't dramatically more expensive at this scale, at least for lightweight vertex stubs). A reflection-based check on `cachedFullScanCandidates` staying `null` would work but means reaching into a private field through the plan tree, which felt like more coupling than the fix warrants for a bug class this narrow. Net: the fix itself is directly readable and I traced it by hand against your exact example shape; I didn't want to add an automated check that either doesn't test the claim or is fragile enough to become a flaky-test problem later.

**Multi-label branch coupling**: added the breadcrumb comment pointing back at `isRowIndependentFullScan()` from `computeVertexIterator`'s multi-label and no-label branches.

Full `com.arcadedb.query.opencypher` package: 8843 tests, 0 failures after this change.

lvca:
Confirmed as requested:

\`\`\`
mvn -o -pl engine -am verify -Dtest=Issue6602MergeAfterEmptyMatchCardinalityTest
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
\`\`\`

All 7 pass, including \`chainedMatchToleratesDownstreamDeleteOfACachedCandidate\`. Full \`com.arcadedb.query.opencypher\` package (8843 tests) is also green, and the branch is up to date with \`main\` with no conflicts.

Leaving the \`instanceof\`-pattern-matching style nit as-is per your own note that it's not worth a diff churn on its own.

PR #6605, opened 2026-08-23
fix: honour InsertChunk.database on first insertStream chunk (#6597) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #6597: gRPC: insertStream ignores InsertChunk.database and requires InsertOptions.database, contradicting the proto
Found while building the TypeScript gRPC client for #4894 (M1b), against a real server.

`arcadedb-server.proto` documents `InsertChunk.database` as required on the first chunk:

```proto
message InsertChunk {
  // REQUIRED on the first chunk; optional on subsequent chunks (server caches)
  string database = 1;
  ...
  InsertOptions options = 3;
}
```

But `ArcadeDbGrpcService#insertStream` never reads it. Inside the `onNext` handler the line is **commented out**:

```java
// String dbName = c.getDatabase();
```

and the database is taken from `InsertOptions` instead (`InsertOptions sent = c.hasOptions() ? c.getOptions() : InsertOptions.getDefaultInstance();`).

**Symptom:** a client that follows the proto and sets `InsertChunk.database` on the first chunk gets

```
Invalid database name: name is required
```

even though it supplied exactly what the contract asks for. Nothing about the error points at `InsertOptions`.

Note the inconsistency across the streaming RPCs: `graphBatchLoad`'s `onNext(GraphBatchChunk)` **does** read `chunk.getDatabase()`, so two client-streaming ingest paths disagree about where the database name lives.

**Client-side workaround in place:** `@arcadedb/client-grpc` mirrors `database` into `options.database` on the first chunk, with a regression test and a comment naming this issue. That workaround should be removed once this is fixed.

This is the fifth contract defect surfaced by building real clients against the published contract, after #6562, #6563, #6558 and #6584. Like those, it is invisible to the #4896 anti-drift test, which compares registered HTTP routes and does not look at the proto at all.

Two possible fixes: honour `InsertChunk.database` as the proto documents (restoring the commented-out line, with `InsertOptions.database` as a fallback), or change the proto to describe what the code actually does. The first keeps `insertStream` consistent with `graphBatchLoad`.

waterWang:
## Description

The proto documents `InsertChunk.database` as REQUIRED on the first chunk:

```proto
message InsertChunk {
  // REQUIRED on the first chunk; optional on subsequent chunks (server caches)
  string database = 1;
}
```

But `insertStream`'s `onNext` handler only reads `InsertOptions.database` and
ignores the chunk-level field entirely — the line `// String dbName = c.getDatabase();`
is commented out (line 2433).

A client that follows the proto and sets `InsertChunk.database` on the first chunk gets
`Invalid database name: name is required` even though it supplied exactly what the
contract asks for.

Note: `graphBatchLoad` already reads `chunk.getDatabase()` on the first chunk and
throws if it's empty (line 2699). `insertStream` is the only streaming RPC that
ignores its chunk-level database field.

## Fix

Before building effective options, copy `InsertChunk.database` into the options
builder if the chunk carries it, falling back to `InsertOptions.database` for
clients that still send the database via options (the existing `@arcadedb/client-grpc`
workaround mirrors `database` into `options.database`).

Fixes #6597

### DIFF
--- a/grpcw/src/main/java/com/arcadedb/server/grpc/ArcadeDbGrpcService.java
+++ b/grpcw/src/main/java/com/arcadedb/server/grpc/ArcadeDbGrpcService.java
@@ -2425,14 +2425,12 @@ public class ArcadeDbGrpcService extends ArcadeDbServiceGrpc.ArcadeDbServiceImpl
 
             InsertOptions effective = defaults(sent); // your existing default-merging helper
 
-            // (Optional) If your chunk carries db/creds (recommended), you can validate/set
-            // them here
 
-            // Otherwise InsertContext will resolve them via 'effective' or server defaults
-            // Example if your proto carries these on the chunk:
-            // String dbName = c.getDatabase();
-            // DatabaseCredentials creds = c.getCredentials();
-            // and pass them to the InsertContext ctor if it expects them.
+            // Issue #6597: honour InsertChunk.database (proto contract: REQUIRED on first chunk)
+            // and fall back to InsertOptions.database for clients that still send it there.
+            // This keeps insertStream consistent with graphBatchLoad.
+            if (!c.getDatabase().isEmpty())
+              sent = sent.toBuilder().setDatabase(c.getDatabase()).build();
 
             // Create and cache context
             ctx = new InsertContext(effective); // begins tx if PER_STREAM / PER_BATCH per your logic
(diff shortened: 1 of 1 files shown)


robfrank:
sorry, we worked on the same isssue more or less at the same time :(

PR #6630, opened 2026-08-23
fix: reject CountEdgesStep optimization when UNWIND precedes OPTIONAL MATCH [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #6629: The aggregation boundary loses grouping after UNWIND + OPTIONAL MATCH: `WITH a, count(m) AS c` returns one row per input row instead of one group
### ArcadeDB version

ArcadeDB 26.9.1-SNAPSHOT (build `4b4e12ca`, freshly extracted from the local
base archive; the original campaign also reproduced this on 26.7.2)

### Environment

- Host OS: Linux x86_64
- Deployment: fresh isolated local runtime
- Query interface: HTTP command API, Cypher language
- Reference engines: Neo4j 2026.06.0, Memgraph 3.12.0, FalkorDB 4.20.1

### Description

The aggregation boundary (WITH with an aggregate function) establishes the
grouping scope over the row pipeline. After a UNWIND expansion followed by
an OPTIONAL MATCH, that boundary is not honored: ArcadeDB evaluates the
aggregate per input row and does not collapse the group — N UNWIND rows
produce N output rows (each with the per-row aggregate) instead of one
grouped row. The same aggregation without UNWIND (or without OPTIONAL
MATCH) collapses correctly, isolating the pipeline combination. The
behavior also occurs when the OPTIONAL MATCH matches nothing (the
aggregate should be 0 over one group, not 0 over N rows).

A fresh recheck on 26.9.1-SNAPSHOT build `4b4e12ca` reproduces the same result
three times: two rows with `c=1` instead of one row with `c=2`.

### To reproduce

```cypher
CREATE (:X {_id: 'x1'});
CREATE (:X {_id: 'x2'});
MATCH (a:X {_id:'x1'}), (b:X {_id:'x2'}) CREATE (a)-[:L]->(b);
MATCH (a:X {_id:'x1'}) UNWIND [1, 2] AS u OPTIONAL MATCH (a)-[:L]->(m:X) WITH a, count(m) AS c RETURN a, c;
```

Expected: one row (a=x1, c=2). Actual: two identical rows (a=x1, c=1).

### Expected behavior

Control 1 — the same aggregation without UNWIND collapses correctly:

```cypher
MATCH (a:X {_id:'x1'}) OPTIONAL MATCH (a)-[:L]->(m:X) WITH a, count(m) AS c RETURN a, c;
-- one row, c=1
```

Control 2 — the same aggregation with UNWIND but without OPTIONAL MATCH:

```cypher
MATCH (a:X {_id:'x1'}) UNWIND [1, 2] AS u WITH a, count(*) AS c RETURN a, c;
-- one row, c=2
```

Control 3 — zero optional matches must still be one group with c=0:

```cypher
MATCH (a:X {_id:'x2'}) UNWIND [1, 2] AS u OPTIONAL MATCH (a)-[:L]->(m:X) WITH a, count(m) AS c RETURN a, c;
-- expected one row c=0; actual two rows c=0
```

### Actual behavior

```text
MATCH (a:X {_id:'x1'}) UNWIND [1, 2] AS u OPTIONAL MATCH (a)-[:L]->(m:X) WITH a, count(m) AS c RETURN a, c;
-> 2 rows: (x1, c=1), (x1, c=1)
```

### Cross-engine comparison (original 26.7.2 campaign baseline)

| Query | ArcadeDB 26.7.2 | Neo4j 2026.06 | Memgraph 3.12 | FalkorDB 4.20.1 |
| UNWIND + OPTIONAL + `count(m)` | **2 rows c=[1,1]** | 1 row c=2 | 1 row c=2 | 1 row c=2 |
| OPTIONAL + `count(m)` (no UNWIND) | 1 row c=1 | 1 row c=1 | 1 row c=1 | 1 row c=1 |

### Isolated recheck

Fresh isolated 26.9.1-SNAPSHOT state per run, 3 runs: witness 2 rows
`c=[1,1]`; no-UNWIND control 1 row `c=1`; no-OPTIONAL control 1 row `c=2`;
zero-optional control 2 rows `c=[0,0]`.  The minimal no-bound-node probe
`UNWIND [1,2] AS u OPTIONAL MATCH (n:Missing) WITH count(n) AS c RETURN c`
returned one row `c=0` on all 3 runs.  The reference-engine results in the
baseline table remain 1 row `c=2` across their recorded 3 runs.

### Why this looks like a bug

1. Internal inconsistency: the engine aggregates correctly without UNWIND
   and without OPTIONAL MATCH — only the combination loses the collapse.
2. The aggregate value is computed per input row (c=1 for each of 2 rows)
   instead of over the group (c=2), indicating the grouping scope is lost.
3. Cypher semantics: WITH containing aggregation groups rows by the
   non-aggregated expressions at that query boundary. The preceding UNWIND and
   OPTIONAL MATCH pipeline should contribute rows to that aggregation scope
   rather than causing the aggregate to be evaluated independently per input
   row.

### Impact

Queries that expand rows (UNWIND) then left-join (OPTIONAL MATCH) then
aggregate — a common reporting pattern — silently return duplicated rows
with per-row aggregates, breaking counts and downstream consumers.

waterWang:
## Description

Fixes #6629: The aggregation boundary loses grouping after UNWIND + OPTIONAL MATCH.

### Root Cause

The `tryOptimizeOptionalMatchCount` optimization (and its chained variant) replaces an `OPTIONAL MATCH ... WITH a, count(m) AS c` pattern with a `CountEdgesStep` that directly counts edges from the bound vertex. This is correct when the bound vertex appears exactly once in the input stream.

However, when an `UNWIND` clause precedes the `OPTIONAL MATCH`, the input stream has multiple rows for the same bound vertex. Since `CountEdgesStep` emits **one output row per input row**, the downstream WITH aggregation sees per-row counts instead of a grouped result — N UNWIND rows produce N output rows with per-row counts, not one grouped row.

### Example

```cypher
MATCH (a:X {_id:'x1'}) UNWIND [1, 2] AS u OPTIONAL MATCH (a)-[:L]->(m:X) WITH a, count(m) AS c RETURN a, c;
```

Before (broken): 2 rows with `c=1` each
After (fixed): 1 row with `c=2`

### Fix

Added a guard in both `tryOptimizeOptionalMatchCount` and `tryOptimizeChainedOptionalMatchCount` that scans the preceding clauses (from index 0 to `currentIndex`) for `UNWIND` or `LOAD_CSV` — both of which fan out rows. When found, the optimization returns `null` and falls back to the general `GroupByAggregationStep`, which correctly aggregates across all rows sharing the same grouping key.

### Controls that still work

- Control 1: `MATCH (a) OPTIONAL MATCH (a)-[:L]->(m) WITH a, count(m) AS c` — no UNWIND, still optimized
- Control 2: `MATCH (a) UNWIND [1,2] AS u WITH a, count(*) AS c` — no OPTIONAL MATCH, not affected
- Control 3: `MATCH (a) UNWIND [...] OPTIONAL MATCH (a)-[:L]->(m) WITH a, count(m)` — now correctly falls back

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -3111,6 +3111,94 @@ public class CypherExecutionPlan {
   }
 
   /**
+   * Variable names referenced by a WITH clause's non-aggregated (grouping-key) items, keyed by
+   * plain variable name where the item is a bare variable, or by its full expression text
+   * otherwise - matching how {@code boundVariables} containment is checked elsewhere for the
+   * same items.
+   */
+  private static Set<String> groupingKeyVariableNames(final List<ReturnClause.ReturnItem> groupingKeys) {
+    final Set<String> names = new HashSet<>();
+    for (final ReturnClause.ReturnItem key : groupingKeys)
+      names.add(key.getExpression() instanceof VariableExpression
+          ? ((VariableExpression) key.getExpression()).getVariableName()
+          : key.getExpression().getText());
+    return names;
+  }
+
+  /**
+   * Variables a single clause binds into scope, for the sole purpose of judging whether it can
+   * fan a single input row out into several. Returns {@code null} when the clause's effect on
+   * row cardinality cannot be proven safe, which the caller treats as "assume it can fan out".
+   */
+  private static Set<String> variablesIntroducedByClause(final ClauseEntry entry) {
+    switch (entry.getType()) {
+    case UNWIND:
+      final UnwindClause unwind = entry.getTypedClause();
+      return Collections.singleton(unwind.getVariable());
+
+    case LOAD_CSV:
+      final LoadCSVClause loadCsv = entry.getTypedClause();
+      return Collections.singleton(loadCsv.getVariable());
+
+    case MATCH:
+      final MatchClause match = entry.getTypedClause();
+      final Set<String> vars = new HashSet<>();
+      if (match.hasPathPatterns()) {
+        for (final PathPattern pattern : match.getPathPatterns()) {
+          for (final NodePattern node : pattern.getNodes())
+            if (node.getVariable() != null)
+              vars.add(node.getVariable());
+          for (final RelationshipPattern rel : pattern.getRelationships())
+            if (rel.getVariable() != null)
+              vars.add(rel.getVariable());
+          if (pattern.hasPathVariable())
+            vars.add(pattern.getPathVariable());
+        }
+      }
+      return vars;
+
+    case WITH:
+    case MERGE:
+    case CREATE:
+    case SET:
+    case REMOVE:
+    case DELETE:
+    case FOREACH:
+      // These clauses project, mutate or (for WITH) possibly aggregate the existing row
+      // stream - none of them can turn one input row into several, so they carry no fan-out
+      // risk regardless of which variables they touch.
+      return Collections.emptySet();
+
+    default:
+      // CALL / SUBQUERY / FINISH and any future clause type: row-cardinality effects are not
+      // analyzed here, so assume the worst rather than silently mis-optimize.
+      return null;
+    }
+  }
+
+  /**
+   * True when every clause before {@code currentIndex} is guaranteed not to fan a single input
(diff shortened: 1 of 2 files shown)


lvca:
@claude review this

lvca:
@claude review latest commit

PR #6633, opened 2026-08-23
fix(#6632): defer a Graph Analytical View's persisted CSR read to first use

### LINKED ISSUE #6632: GAV follow-ups from #6583/#6588: lazy CSR restore for never-queried sessions, gavRestoreAwaitTimeout undocumented
## Summary

Two follow-ups from #6583 / PR #6588 (Graph Analytical View CSR persistence), filed together since both surfaced from the same review/benchmark thread and neither is a blocker on its own.

## 1. Lazy CSR restore - an embedded session that never queries the view still pays the restore cost

PR #6588 persists a view's CSR to disk at clean close and restores it eagerly during `open()` (via `restoreAll()`) whenever the freshness certificate matches. That collapses the case the original issue measured - reopening an unchanged database - from O(graph size) to a bounded, sublinear cost. But it does not eliminate cost for the one session shape that gets no benefit from it: **an embedded session that opens a database with a view and never issues a query that touches it.**

`tae898`'s benchmarks on the merged PR found the eager restore itself costs real time at scale, sublinear in graph size but not free:

| vertices | edges | CSR size | restore cost |
|---:|---:|---:|---:|
| 1,000,000 | 4,000,000 | 42.1 MB | ~210-240 ms |
| 3,000,000 | 12,000,000 | 126.2 MB | ~300-320 ms |
| 10,000,000 | 40,000,000 | 420.8 MB | ~980-1000 ms |

For a server that opens once and serves many queries, this amortizes to nothing. For an embedded application that opens a database with a registered view but happens not to query it in a given session, it is pure overhead paid on every open with zero return - the same shape of problem #6583 itself described, just moved from close to open and bounded rather than unbounded.

**Proposed direction** (from `tae898`, explicitly flagged as a separate issue rather than a blocker on #6588): defer the disk restore until the view is actually needed - i.e., the first query that would use it, or an explicit `awaitReady()`/status check - instead of restoring unconditionally during `open()`. That takes the never-queried case to zero rather than merely shrinking it, at the cost of the first real query paying the restore latency instead of `open()`.

Source: https://github.com/ArcadeData/arcadedb/pull/6588 (see `tae898`'s three benchmark comments dated 2026-08-23).

## 2. Documentation gap: `arcadedb.gavRestoreAwaitTimeout` was never documented

Unrelated to #6588's diff, but discovered while updating `arcadedb-docs` for it: the `arcadedb.gavRestoreAwaitTimeout` setting (added by PR #5883, closing #5788 - "milliseconds `database.open()` blocks waiting for a Graph Analytical View's async rebuild before returning, so queries issued right after don't run unaccelerated") is not documented anywhere in `arcadedb-docs`:

- Not in `reference/settings.adoc`'s `DATABASE` settings table
- Not in `how-to/data-modeling/graph-olap.adoc`

It's a real, user-facing knob (default `0`, i.e. `open()` returns immediately and the rebuild races the first queries) that changes observable behavior at open time, and it predates #6588 - it should be documented independently of the CSR-persistence docs added for this PR.

## Scope

Neither item blocks anything currently merged. (1) is a genuine optimization opportunity with a clear proposed direction; (2) is a small, low-risk documentation addition.

lvca:
## Summary
- `restoreAll()` (issue #6583, PR #6588) marked a view READY only after eagerly reading and deserializing its whole persisted CSR right in `restoreFromDiskOrBuildAsync()`, called synchronously from `database.open()`. That read is bounded but real - up to ~1s at 10M vertices per #6632's own measurements - and a session that opens a database with a view and never queries it paid for it anyway.
- When a persisted CSR plausibly applies (`mayHavePersistedCsr()`: the same guards `tryRestoreFromPersistedCsr()` already checks, plus a cheap `File.isFile()` instead of reading the file), the view is now marked READY immediately without touching the file, and the actual read is deferred to whichever happens first: a real query (`checkBuilt()`) or an explicit `awaitReady()` call.
- `tryRestoreFromPersistedCsr()` itself is untouched: it already re-samples the freshness certificate from scratch whenever it runs, so a commit landing during the pending window (before any change listener is registered to see it coming) is still caught correctly and falls back to a rebuild - no new invalidation logic needed.
- When no persisted file could plausibly apply, the existing eager `buildAsync()` path is unchanged: a full scan is expensive regardless of whether it's queried, so there's nothing to gain by delaying its already-background start.
- Also updates `GAV_RESTORE_AWAIT_TIMEOUT`'s description to reflect the new deferred-restore interaction.

Closes #6632

## Why this is safe
- `shutdown()`'s existing `awaitInFlightTasks()`/`persistCsrIfPossible()` guards already handle a never-triggered pending restore correctly with no changes needed: `inFlightTasks` stays 0 (nothing was dispatched), and `persistCsrIfPossible()` already early-returns when `snapshot == null`.
- `checkBuilt()`/`awaitReady()` are the two existing places every real caller and both existing #6583 regression tests already go through, so the deferred trigger is invisible to well-behaved callers.
- `status` is published READY only after `pendingDiskRestore` is set, following the same plain-volatile publish order `snapshot`+`status` already use elsewhere in this class (e.g. `build()`).

## Test plan
- [x] New regression tests in `GraphAnalyticalViewTest`:
  - `csrLazyRestoreIsDeferredUntilFirstQuery`: reopen shows READY-but-not-built immediately, first query resolves it transparently with correct data.
  - `csrLazyRestoreNeverTriggeredIsHarmlessAtClose`: a session that never touches the view survives close and a later reopen still restores correctly.
  - `csrLazyPendingRestoreCatchesAnInterveningCommitBeforeFirstQuery`: a commit landing before the first query is still caught by the deferred certificate re-check, forcing a rebuild rather than serving stale data.
- [x] Full `GraphAnalyticalViewTest` suite (149 tests) passes.
- [x] Every other GAV-consuming suite passes (244 tests): `GraphAnalyticalViewRegistryTest`, `GAVEligibilityTest`, `StatisticsProviderTest`, `MatchGAVExpandIntoTest`, Cypher GAV cardinality/expand-into tests, `GraphAlgorithmsTest`, algo procedure tests, HA-follower-gating supernode test, etc.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2126,9 +2126,13 @@ public enum GlobalConfiguration {
   GAV_RESTORE_AWAIT_TIMEOUT("arcadedb.gavRestoreAwaitTimeout", SCOPE.DATABASE,
       """
       Milliseconds database open() blocks waiting for Graph Analytical Views (GAV/CSR) restored from persisted \
-      definitions to reach READY before returning. 0 (default) does not wait: the async rebuild is still triggered, \
-      but open() returns immediately and queries issued right after it run unaccelerated until the rebuild completes. \
-      A positive value trades a slower open() for the restored view being usable by the query that triggered the reopen""",
+      definitions to reach READY before returning. 0 (default) does not wait: when no persisted CSR plausibly \
+      applies, the full rebuild is still triggered immediately in the background, but open() returns before it \
+      completes and queries issued right after run unaccelerated until it does; when a persisted CSR does plausibly \
+      apply (see arcadedb.gavPersistCsr), nothing is even started at open() time as of v26.9.1 - it is deferred to \
+      whichever query touches the view first, so the fast path costs nothing for a session that never queries it. \
+      A positive value here forces the wait either way, trading a slower open() for the restored/rebuilt view being \
+      usable by the query that triggered the reopen""",
       Long.class, 0L),
 
   GAV_PERSIST_CSR("arcadedb.gavPersistCsr", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProviderRegistry.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProviderRegistry.java
@@ -172,15 +172,22 @@ public class GraphTraversalProviderRegistry {
 
     final long deadlineNanos = System.nanoTime() + unit.toNanos(timeout);
     for (final GraphTraversalProvider provider : list) {
-      if (provider.isReady())
-        continue;
+      // A GraphAnalyticalView always goes through awaitReady(), regardless of isReady(): since a deferred
+      // restore-from-disk (see #6632) reports READY the moment a persisted CSR plausibly applies, before
+      // the disk read that actually resolves it has even run, isReady() alone is no longer a trustworthy
+      // "nothing left to do" signal for this provider type - awaitReady() is the one call that triggers
+      // that read. It is cheap/idempotent once nothing is pending (its own trigger call no-ops and the
+      // wait loop returns immediately), so this costs nothing extra for an already-settled view.
       if (provider instanceof GraphAnalyticalView) {
         final long remainingNanos = deadlineNanos - System.nanoTime();
         if (remainingNanos <= 0)
           return false;
         if (!((GraphAnalyticalView) provider).awaitReady(remainingNanos, TimeUnit.NANOSECONDS))
           return false;
+        continue;
       }
+      if (provider.isReady())
+        continue;
     }
     return true;
   }
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -1614,6 +1614,11 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
    * from scratch, whenever the deferred restore actually runs, and correctly falls back to a rebuild if
    * anything no longer holds by then (including a commit that landed while the restore was pending, since
    * nothing is watching for one — see {@link #restoreFromDiskOrBuildAsync()}).
+   * <p>
+   * KEEP IN SYNC with {@link #tryRestoreFromPersistedCsr()}'s own guard clauses (review on PR #6633): a
+   * drift between the two can't corrupt data — {@code tryRestoreFromPersistedCsr()} always has the final,
+   * authoritative say — but it would either waste a lazy round-trip that immediately falls back to a
+   * rebuild, or skip the lazy fast path entirely for a case that would have actually succeeded.
    */
   private boolean mayHavePersistedCsr() {
     if (name == null)
@@ -1757,6 +1762,9 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
    * database's current last committed transaction id — i.e. nothing was committed in between. Returns false
    * (leaving the view's state untouched) for a missing file, a mismatched definition or certificate, or any read
    * failure, so the caller can fall back to {@link #buildAsync()} exactly as it would have without this path.
+   * <p>
+   * The guard clauses below are mirrored (cheaply, without the file read) by {@link #mayHavePersistedCsr()} —
+   * KEEP THEM IN SYNC (review on PR #6633).
    */
   synchronized boolean tryRestoreFromPersistedCsr() {
     if (name == null)
@@ -1810,12 +1818,15 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
     // just above this line would be invisible to the restored snapshot (no scan ran to absorb it, unlike
     // buildAsync(), whose real scan runs AFTER its own sample so a racing commit is already captured in what
     // it reads) and invisible to onRelevantCommit()/applyDelta() too (the listeners were not live yet when it
-    // happened). Not reachable via the current call site - restoreAll() runs synchronously inside
-    // LocalDatabase.open(), before the Database is handed to any caller, so nothing holding a reference to it
-    // can commit concurrently - but checked defensively so this stays correct if that context ever changes
-    // (e.g. an HA replica replaying commits concurrently with a local open). Marking STALE mirrors exactly
-    // what onRelevantCommit() does for a real relevant commit under OFF/SYNCHRONOUS mode: conservative, but
-    // never wrong.
+    // happened). This specific few-instruction window is still only defensive - no existing test hits it -
+    // but the surrounding claim that this whole method is unreachable outside LocalDatabase.open() no longer
+    // holds since #6632 (review on PR #6633): dispatchDeferredRestore()'s background task also calls this
+    // method, well after open() has returned and the caller can freely commit against it. The COARSER case -
+    // a commit landing before this method is even dispatched, caught by the certificate mismatch in
+    // GraphAnalyticalViewCSRPersistence.load() above rather than by this specific check - is genuinely
+    // reachable and exercised by csrLazyPendingRestoreCatchesAnInterveningCommitBeforeFirstQuery. Marking
+    // STALE here mirrors exactly what onRelevantCommit() does for a real relevant commit under
+    // OFF/SYNCHRONOUS mode: conservative, but never wrong.
     if (currentLastTransactionId() != currentTxId) {
       LogManager.instance().log(this, Level.WARNING,
           "GraphAnalyticalView '%s': a commit landed while restoring its CSR from disk; marking it STALE rather than risk missing it",
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -332,6 +332,12 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
    * @param edgeTypes   edge type names to include (null = all)
    */
   public synchronized void build(final String[] vertexTypes, final String[] edgeTypes) {
+    // A direct build() (e.g. REBUILD GRAPH ANALYTICAL VIEW) supersedes any not-yet-resolved deferred
+    // restore-from-disk (see #6632): without clearing this, a later awaitReady() call would still see
+    // pendingDiskRestore == true and dispatch dispatchDeferredRestore(), re-reading a possibly-superseded
+    // persisted file (or a wholly redundant rebuild) right after this scan already produced a fresh,
+    // authoritative snapshot.
+    pendingDiskRestore = false;
     final CountDownLatch latch = new CountDownLatch(1);
     readyLatch = latch;
     status = Status.BUILDING;
@@ -384,6 +390,10 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
   public synchronized void buildAsync() {
     if (!buildQueued.compareAndSet(false, true))
       return; // a build is already queued or running
+    // See build(vertexTypes, edgeTypes)'s identical guard: supersedes any not-yet-resolved deferred
+    // restore-from-disk (see #6632). Also reached, harmlessly, from dispatchDeferredRestore()'s own
+    // fallback call - pendingDiskRestore is already false there by the time it calls this.
+    pendingDiskRestore = false;
     final CountDownLatch latch = new CountDownLatch(1);
     readyLatch = latch;
     status = Status.BUILDING;
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -243,6 +243,12 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
   private volatile Status            status    = Status.NOT_BUILT;
   private volatile CountDownLatch    readyLatch = new CountDownLatch(1);
   private volatile Throwable         buildError;
+  // True from the moment restoreFromDiskOrBuildAsync() finds a plausible persisted CSR (see #6632) until
+  // whatever touches the view first — checkBuilt() or awaitReady() — resolves it. Written before `status`
+  // is flipped to READY (see restoreFromDiskOrBuildAsync()) and read without synchronization elsewhere, so
+  // its value is guaranteed visible to any thread that observes that READY write — the same plain-volatile
+  // publish pattern `snapshot`+`status` already rely on throughout this class (e.g. build()).
+  private volatile boolean           pendingDiskRestore;
 
   // Incremental auto-update
   private DeltaCollector         deltaCollector;
@@ -445,10 +451,15 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
 
   /**
    * Waits until the view reaches READY status or the timeout expires.
+   * <p>
+   * If the view is still a deferred restore-from-disk (see #6632), this is one of the two ways — the
+   * other being a real query, via {@link #checkBuilt()} — that dispatches it: an explicit status wait is
+   * a legitimate signal that the caller wants the view actually resolved, not just optimistically READY.
    *
    * @return true if the view is READY, false if the timeout elapsed or the build failed
    */
   public boolean awaitReady(final long timeout, final TimeUnit unit) {
+    triggerDeferredDiskRestoreIfPending();
     final long deadlineNanos = System.nanoTime() + unit.toNanos(timeout);
     try {
       while (true) {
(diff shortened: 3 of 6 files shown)


tae898:
We collided: I had #6634 open with the same goal about five minutes after this one, built before I saw yours. **I have closed mine in favour of this**, since it is yours and it is the better-scoped change. One finding from having built the same thing, and one offer.

## The deferred read can be triggered by plan costing, where it blocks without a timeout

`StatisticsProvider` is untouched by this PR, and on a reopened database the chain reaches the deferred read at *planning* time:

1. `restoreFromDiskOrBuildAsync()` publishes the view as `READY` with `pendingDiskRestore = true` and `snapshot == null`.
2. A query is planned. `StatisticsProvider.getMeanEdgesPerConnectedPair()` misses its cache and calls `exactMeanFromTraversalProvider(edgeType)`.
3. That calls `GraphTraversalProviderRegistry.findProvider(database, edgeType)`, which returns the view because `isReady()` is now true before the file has been read.
4. It then calls `provider.getMeanEdgesPerConnectedPair(edgeType)`, which is `checkBuilt()`, which is `triggerDeferredDiskRestoreIfPending()` + `awaitDeferredDiskRestoreSettled()`.

So the CSR read happens inside the cost model, and `checkBuilt()` blocks with no timeout by design (the same contract `build()` has for a cold view, per your javadoc). Two consequences:

- It fires for a query that merely *touches* the edge type, including one whose chosen plan never uses the CSR. Under the eager path that read had already happened at `open()`; under this one it is charged to whichever query first costs a pattern on that edge type.
- An unbounded blocking read moves from `open()`, where a slow or failing read is expected and handled, into plan costing. That is a category change more than a latency change: `open()` is where callers already expect to wait.

The blast radius is bounded by `graphStatisticsCache` being keyed on the edge-type record count, so it is at most once per edge type per count change rather than once per query. I am not claiming it is fatal, only that it is the one place I found where "defer until first use" and "first *use*" disagree: the planner is asking whether acceleration exists, not using it.

## What I did about it, if it is useful

In my version the planner got a non-triggering lookup, `findProviderIfReady()`, that returns a provider only when it is genuinely ready and never activates anything, and `StatisticsProvider` uses that instead of `findProvider()`. It is a small patch and it composes with your design unchanged: yours would return the view from `findProvider()` as it does now, and simply not hand it to the cost model until the snapshot actually exists.

Happy to open that as a separate small PR against this branch, or paste the diff here, or drop it entirely if you would rather the planner keep triggering the load. Your call, and no need to take the rest of mine.

## Measurements, in case they save you a run

Same fixture shape as #6632's table, medians of 5 reps after 2 warmups, on a session that opens and closes with no query at all. `LAZY` here is my implementation, not yours, so treat it as "what deferring buys" rather than as a measurement of this PR:

| vertices | edges | no view open (ms) | eager open (ms) | deferred open (ms) |
|---:|---:|---:|---:|---:|
| 100,000 | 400,000 | 3.32 | 15.33 | 3.67 |
| 300,000 | 1,200,000 | 3.52 | 21.30 | 3.50 |
| 1,000,000 | 4,000,000 | 3.80 | 98.45 | 3.20 |
| 1,000,000 | 8,000,000 | 2.82 | 146.58 | 3.23 |
| 1,000,000 | 16,000,000 | 3.84 | 345.68 | 3.18 |
| 1,000,000 | 32,000,000 | 2.63 | 1137.74 | 3.19 |

The eager cost tracks the graph and the deferred open does not, which is the whole point: it reads nothing, so its cost does not depend on the size of what it did not read. Note the last row, which is the one I did not expect: holding vertices fixed at 1M and taking fanout to 32 puts the eager open at **1137.74 ms**, above the 10M-vertex figure in #6632, so edge count drives this at least as hard as vertex count does. A 10M/40M run is measuring now and I will post it when it lands.

tae898:
Correction to my comment above: **I overstated that finding and you should discount most of it.** I described the planner reaching the deferred read from reading the call chain, without demonstrating it. I have now measured it on your branch and the practical claim does not hold.

Each row is a fresh `open()` of a database with a persisted CSR, running one statement, then asking `isRestoredFromPersistedCsr()`. 50,000 vertices / 200,000 edges.

| statement | CSR read afterwards | ms |
|:--|:--|---:|
| no statement at all | no | 0 |
| `MATCH (a:P) RETURN count(*)` (no edge type) | no | 329 |
| `EXPLAIN MATCH (a:P)-[:E]->(b:P) RETURN count(*)` | no | 17 |
| `EXPLAIN MATCH (a:P)-[:E]->(b:P)-[:E]->(c:P) RETURN count(*)` | no | 11 |
| `EXPLAIN MATCH (a:P)-[:E]->(b) RETURN count(*)` | no | 2 |
| `EXPLAIN MATCH (a:P {id: 42})-[:E]->(b:P) RETURN count(*)` | no | 12 |
| `MATCH (a:P)-[:E]->(b:P) RETURN count(*)` (executed) | **yes** | 44 |

Planning does not trigger it. Only execution does, which is exactly what your PR says it does.

What is left of the finding, for completeness rather than as an objection. The chain itself is live: constructing a `StatisticsProvider` on a freshly opened database and calling `getMeanEdgesPerConnectedPair("E")` directly does flip it, in 24 ms at this size.

```
after open:      status=READY  restoredFromDisk=false
getMeanEdgesPerConnectedPair("E") = 1.0000 in 24 ms
after the call:  status=READY  restoredFromDisk=true
```

So `StatisticsProvider` -> `findProvider()` -> `checkBuilt()` can reach the deferred read and block on it, but I could not construct a query whose planning actually goes down that path, and four `EXPLAIN` shapes say the optimizer does not. That makes it a latent edge rather than the routine hazard I implied. If you think no planning path can reach `getMeanEdgesPerConnectedPair()` on a not-yet-read view, then there is nothing here to fix and my `findProviderIfReady()` offer is not worth taking.

Sorry for the noise. I should have run this before posting the first version rather than after.

While I had the harness pointed at your branch, the part that does check out is the thing the PR is for:

| vertices | edges | no view open (ms) | with view open (ms) | delta |
|---:|---:|---:|---:|---:|
| 50,000 | 200,000 | 5.80 | 6.42 | +0.61 |

Same shape eagerly restoring was ~20 ms at this size. I am running the vertex and edge sweep against your branch now, up to 10M vertices / 40M edges, and will post the table here when it finishes.

tae898:
Found a real one this time, and I have held it until I had a reproduction, a control against `main`, a fix, and your suite passing with the fix.

**`checkBuilt()` throws instead of waiting while the deferred restore is in flight.**

`triggerDeferredDiskRestoreIfPending()` clears `pendingDiskRestore` at **dispatch** (`GraphAnalyticalView.java:1626`), before the worker has read anything, and `checkBuilt()` only waits when that flag is still true (`:2303`). So for the entire duration of the read the view is `(snapshot == null, pendingDiskRestore == false, status == BUILDING)`, and any `checkBuilt()` entered in that window skips both the trigger and `awaitDeferredDiskRestoreSettled()` and falls through to the throw at `:2307`.

## Reproduction 1: a supported setting, no race

`GAV_RESTORE_AWAIT_TIMEOUT = 1`, 1,000,000 vertices / 4,000,000 edges. `restoreAll()` calls `awaitReady(budget)` on the caller's behalf (`GraphAnalyticalViewPersistence.java:200`), which dispatches the restore and consumes the one-shot flag. The budget expires long before a 42 MB CSR is read, so `open()` returns having spent the trigger:

```
open() returned in 10 ms: status=BUILDING restoredFromDisk=false
  view.getNodeCount()  -> IllegalStateException: GraphAnalyticalView has not been built yet. Call build() first.
  view.getEdgeCount()  -> IllegalStateException: GraphAnalyticalView has not been built yet. Call build() first.
  a Cypher query       -> 4
after the restore finishes: status=READY restoredFromDisk=true, getNodeCount() -> 1000000
```

Single-threaded, no interleaving to lose. The user set a documented option and made an ordinary API call.

**Control, same test on `main` (`bc5d31a63`), same setting, same fixture:**

```
open() returned in 139 ms: status=READY restoredFromDisk=true
  view.getNodeCount()  -> 1000000
  view.getEdgeCount()  -> 4000000
```

Before this PR the restore ran synchronously inside `restoreAll()`, so a registered view with a null snapshot could not be observed on the restore path. The window is new.

## Reproduction 2: default settings, no configuration at all

Two callers using the view while the read is running. The first triggers it and waits; the second arrives 40 ms later:

```
after open: status=READY restoredFromDisk=false
first caller  (triggers the read, then waits) -> nodeCount=1000000
second caller (arrives 40 ms later)           -> IllegalStateException: GraphAnalyticalView has not been built yet.
```

## What is and is not exposed

SQL and Cypher are largely shielded, which is why this did not show up in the query benchmarks: `findProviderIfReady`-style resolution rejects a `BUILDING` view (`isReady()` is `READY || (STALE && useWhenStale)`), so those queries quietly fall back to OLTP. That is the `-> 4` line above.

What is exposed is the direct Java API: anything routed through `checkBuilt()`, including `getNodeCount()`, `getEdgeCount()`, and the property and neighbour accessors. An embedded application holding a `GraphAnalyticalView` gets a hard failure on a healthy, fully-configured view.

## Fix

`awaitDeferredDiskRestoreSettled()` already returns immediately when `status != BUILDING`, and `triggerDeferredDiskRestoreIfPending()` is already idempotent, so the flag is not needed as a second gate at all:

```java
-    if (snap == null && pendingDiskRestore) {
+    if (snap == null) {
```

A genuinely never-built view still reaches the same throw, because the trigger is a no-op and the settle call returns at once. A view whose restore is in flight now waits for it, which is what the caller wanted.

With that one line applied, both reproductions above return `1000000` / `4000000` instead of throwing, and your suite is green: `GraphAnalyticalView*`, `GraphTraversalProvider*`, `CSR*`, `Issue6049*`, `CypherGAV*` = **171 tests, 0 failures**, including your three new deferred-restore tests.

Worth a regression test with two callers, since all three of the PR's new tests are single-threaded and each touches the view before anything else does, which is exactly the shape that cannot see this.

Environment: Intel Core Ultra X9 388H, 12 threads pinned, `eclipse-temurin:25-jdk`, engine at `916b6c5da`, control at `bc5d31a63`.

lvca:
This PR is closed, @tae898 could you please open a new issue for your last finding?

tae898:
The sweep I promised. Two builds, same probe, same fixture shapes, one at a time on an otherwise idle machine: `upstream/main` at `bc5d31a63` (eager, #6588 as merged) against this PR at `916b6c5da` (deferred).

A session is `open()` then `close()` with **no query at all**, which is the case #6632 is about. Medians of 5 reps after 2 warmups, ms.

| | | eager (`main`) | | | deferred (this PR) | | |
| vertices | edges | no view | with view | **delta** | no view | with view | **delta** |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 10,000 | 40,000 | 5.72 | 5.96 | **+0.24** | 7.01 | 6.78 | **-0.23** |
| 50,000 | 200,000 | 6.27 | 21.41 | **+15.14** | 6.58 | 7.27 | **+0.69** |
| 100,000 | 400,000 | 5.72 | 18.60 | **+12.88** | 5.85 | 5.88 | **+0.03** |
| 300,000 | 1,200,000 | 6.32 | 56.80 | **+50.48** | 5.70 | 6.18 | **+0.48** |
| 1,000,000 | 4,000,000 | 5.55 | 105.83 | **+100.27** | 5.21 | 5.36 | **+0.15** |
| 1,000,000 | 8,000,000 | 5.02 | 142.20 | **+137.17** | 5.45 | 5.45 | **+0.00** |
| 1,000,000 | 16,000,000 | 7.61 | 249.42 | **+241.81** | 5.77 | 6.24 | **+0.46** |
| 1,000,000 | 32,000,000 | 5.77 | 1116.32 | **+1110.55** | 7.40 | 8.04 | **+0.64** |
| 3,000,000 | 12,000,000 | 4.27 | 272.00 | **+267.74** | 5.81 | 7.08 | **+1.27** |
| 10,000,000 | 40,000,000 | 5.01 | 1032.46 | **+1027.46** | 4.97 | 5.74 | **+0.77** |

`isRestoredFromPersistedCsr()` after open / after EXPLAIN / after a real query, sampled in separate fresh sessions:

| arm | after open | after EXPLAIN | after a query |
|:--|:--|:--|:--|
| eager (`main`) | yes | yes | yes |
| deferred (this PR) | no | no | yes |

(identical at all ten sizes)
| 50,000 | 200,000 | 6.58 | 21.41 | **+15.14** | 7.27 | **+0.69** | yes / no |
| 100,000 | 400,000 | 5.85 | 18.60 | **+12.88** | 5.88 | **+0.03** | yes / no |
| 300,000 | 1,200,000 | 5.70 | 56.80 | **+50.48** | 6.18 | **+0.48** | yes / no |
| 1,000,000 | 4,000,000 | 5.21 | 105.83 | **+100.27** | 5.36 | **+0.15** | yes / no |
| 1,000,000 | 8,000,000 | 5.45 | 142.20 | **+137.17** | 5.45 | **+0.00** | yes / no |
| 1,000,000 | 16,000,000 | 5.77 | 249.42 | **+241.81** | 6.24 | **+0.46** | yes / no |
| 1,000,000 | 32,000,000 | 7.40 | 1116.32 | **+1110.55** | 8.04 | **+0.64** | yes / no |
| 3,000,000 | 12,000,000 | 5.81 | 272.00 | **+267.74** | 7.08 | **+1.27** | yes / no |
| 10,000,000 | 40,000,000 | 4.97 | 1032.46 | **+1027.46** | 5.74 | **+0.77** | yes / no |

The eager column tracks the graph and the deferred column does not, which is the whole point: a deferred `open()` reads nothing, so its cost cannot depend on the size of the thing it did not read. At 10M vertices / 40M edges that is **1027.46 ms against 0.77 ms**.

Two things in the data worth calling out, because neither was obvious to me before running it:

**Edge count drives this at least as hard as vertex count.** Holding vertices fixed at 1M and taking fanout from 4 to 32 moves the eager cost by more than the whole 1M-to-10M vertex axis does. #6632's table is indexed by vertices, which understates what a dense graph pays.

**Close is unaffected.** `close()` sits at 6 to 17 ms in every row of both arms. The cost #6583 originally measured at close moved to open in #6588 and has now been removed from open, without reappearing at close: nothing waits for a restore that was never triggered.

The `restored after` columns are `isRestoredFromPersistedCsr()` sampled in separate fresh sessions, after `open()`, after an `EXPLAIN`, and after a real query. They read `yes yes yes` on main and `no no yes` on this PR at every size, which is the mechanism showing itself rather than being asserted.

Correctness held throughout: every size returns identical row counts with the view eager, deferred, and absent.

Environment: Intel Core Ultra X9 388H, 12 threads pinned via `--cpuset-cpus 0-11`, 30 GiB RAM, NVMe SSD, `eclipse-temurin:25-jdk`. Fixtures are synthetic: N vertices of type `P` with a unique index on `id`, `fanout` edges of type `E` from each vertex to `(i + f * 7919) mod N`, `UPDATE MODE OFF`, CSR persisted by a clean close before measuring.

tae898:
Two more, both measured on both arms and both reproduced twice. Same setup as the sweep above: `main` at `bc5d31a63` against this PR at `916b6c5da`, 1,000,000 vertices / 4,000,000 edges.

## 1. When the certificate no longer matches, the first query now blocks through the entire rebuild

Open a database with a persisted CSR, commit something (invalidating the certificate), then query.

| | `main` (eager) | this PR (deferred) |
|:--|---:|---:|
| view status after the commit | `STALE` | `READY` |
| **first query** | **5 ms** / **7 ms** | **4,878 ms** / **5,715 ms** |
| second query | 0 ms | 0 ms |

Both return the same correct answer. The difference is who does the rebuild and whether the caller waits for it.

On `main` the snapshot already exists when the commit lands, so `onRelevantCommit()` marks the view `STALE`, provider resolution rejects it, and the query falls back to OLTP immediately. On this PR the view has no snapshot yet, so there is nothing for the commit to invalidate and the status stays `READY`. The query is then handed the view, reaches `checkBuilt()`, triggers the deferred read, `tryRestoreFromPersistedCsr()` correctly rejects the stale certificate, and `buildAsync()` starts a full scan, which `awaitDeferredDiskRestoreSettled()` then waits out on the query thread with no timeout.

So the staleness signal is not lost in a way that risks correctness (the sweep above and a separate test both confirm the answer stays right). It is lost in a way that changes who pays: a session that would have run one unaccelerated query now blocks through an O(V+E) rebuild. At 1M that is the ~5 s above. #6583's own numbers put a 10M rebuild at 52,923 ms.

This is the same shape as the problem #6583 opened with, arriving by a different route: not "every open pays a scan" but "the first query after any intervening commit pays one, synchronously". Databases written to between sessions are exactly the population the CSR-persistence work already documents as not benefiting.

Possible direction, entirely yours to judge: when the deferred read finds the certificate no longer applies, do what `main` effectively did and let this query fall back while the rebuild proceeds in the background, rather than blocking the caller through it. That keeps "never serve stale data" while restoring "the query that reopened the database does not pay O(V+E)".

## 2. `awaitAll()` no longer warms anything

`GraphTraversalProviderRegistry.awaitAll()` is documented as the warm-up call: *"Use this after opening a database with persisted GAVs to ensure all CSR structures are built before running performance-critical queries."* It skips any provider already reporting `isReady()`, and a deferred view reports `READY` before its file has been read.

| | `main` | this PR |
|:--|:--|:--|
| `restoredFromDisk` before | true | false |
| `awaitAll(60s)` | true in 0 ms | true in 0 ms |
| `restoredFromDisk` after | true | **still false** |

It returns success in zero milliseconds having done nothing, so a caller following the documented advice gets exactly the cold first query the call exists to prevent. `awaitReady()` on the view itself does trigger the read, so the fix is presumably to route `awaitAll()` through the same path rather than gating on `isReady()`.

## Things I checked that did NOT hold

Reporting these too, since I would rather be predictable about what I am claiming after getting one wrong earlier today:

- `awaitReady()` on an already-built view does **not** demote it. Status stays `READY`, node count stays correct, on both arms.
- The schema view reporting `nodes`/`edges` as null for a not-yet-read view is **not** this PR's doing: `main` reports null there too.

tae898:
**Withdrawing two of the three things I reported above.** I had fetched this branch at `916b6c5da` and tested that, not knowing you had pushed the fourth and fifth review rounds after it. You had already fixed both before merging, and I have now rebuilt and re-run everything against what actually shipped (`f412c84b0`).

| what I reported | against `916b6c5da` | against shipped `f412c84b0` |
|:--|:--|:--|
| `checkBuilt()` throws instead of waiting | reproduced | **fixed, both reproductions pass** |
| `awaitAll()` warms nothing | reproduced | **fixed, warms in 205 ms** |
| first query blocks through a full rebuild | reproduced | **still reproduces, 5,613 ms** |

On the first one, `deferredRestoreInFlight` is a better fix than the `snap == null` change I suggested, because it stays true for exactly the duration of the work rather than being inferred from a flag that means something else. Sorry for the noise; I should have re-fetched before writing it up.

The third one does still reproduce against shipped main, so repeating just it, freshly measured on `f412c84b0`, 1,000,000 vertices / 4,000,000 edges:

```
after open:                    status=READY  restoredFromDisk=false
after an invalidating commit:  status=READY
FIRST QUERY took 5,613 ms (result 4)
second query took 0 ms
```

Against `bc5d31a63` (before the deferral) the same sequence answered in 5 to 7 ms, because the commit marked the view `STALE`, provider resolution rejected it, and the query fell back to OLTP. Now the view has no snapshot for the commit to invalidate, so it stays `READY`, gets handed to the query, and the query waits out a full O(V+E) rebuild on its own thread with no timeout.

A corrupt file reaches the same place: I wrote 4 KB of junk into the midpoint of a 56 MB `gav-v1.csr`, and the damaged file was correctly rejected with no hang, but the first query then paid 5,862 ms for the rebuild.

So it is any unusable-CSR path, not just an invalidated certificate, and #6583's own numbers put a 10M rebuild at 52,923 ms.

Happy to open this as a separate issue since this PR is merged, unless you would rather it stayed here. Everything else from my earlier comments still holds: the sweep numbers were taken on `916b6c5da` but deferral behaves identically on the merged commit (`open()` in 9 ms, `restoredFromDisk=false` after open), and the correctness checks all passed on both.

PR #6634, opened 2026-08-23
feat(#6632): optionally defer the Graph Analytical View CSR restore to first use

tae898:
Addresses **item 1 of #6632**. Item 2 (documenting `arcadedb.gavRestoreAwaitTimeout`) is an `arcadedb-docs` change and is not touched here, so this does not close the issue.

I had this built against `fix/6583-gav-csr-persistence` before #6588 merged; it is now rebased onto `main` (`bc5d31a63`). Entirely happy for it to be closed if you would rather write it yourself now that the issue is filed - it is offered as a starting point, not a claim on the work.

**If you do write your own, these are the three things worth taking from here**, because each is a place where the obvious implementation is wrong and two of them only showed up once the test was written:

1. **The query planner must not trigger the load.** `StatisticsProvider.exactMeanFromTraversalProvider()` calls `findProvider()` while *costing a plan*. If that can load a CSR, the stall does not go away, it moves from `open()` into an arbitrary query's planning phase.
2. **A `STALE` view must be left alone.** It is not-ready for a different reason than deferred-and-not-yet-loaded, and treating the two alike makes an ordinary query start a rebuild that does not happen today. Observed without the guard: a query moved the view `STALE` -> `BUILDING`.
3. **`GAV_RESTORE_AWAIT_TIMEOUT` must be skipped.** A deferred view is `NOT_BUILT` with no build in flight, and `readyLatch` is created at construction rather than left null, so `awaitReady()` sits on a latch nobody counts down and burns the entire budget. Without the guard, the two settings on together gave an `open()` that blocked for the full 60 s **and** still had not loaded the view.

## What this is

#6588 turned a reopen from a full graph scan into a file read: 52,923 ms to 984 ms at 10M vertices / 40M edges. The remaining 984 ms is still paid by a session that never touches the view, because creating a view says it should be **available**, not that it should be loaded the instant anyone opens the database. On the same graph an open with no view configured at all is ~3 ms, so such a session spends ~99.7% of its open time on a structure it never reads.

Behind `arcadedb.gavLazyRestore` (`GAV_LAZY_RESTORE`, default `false`) the view is registered at open and nothing is read. The first lookup that would actually use it pays the restore instead.

## Measurements

A session here is: open, then close, with **no query in between**. Medians of 5 reps after 2 warmups, ms.

| vertices | edges | no view open | no view close | EAGER open | EAGER close | LAZY open | LAZY close | LAZY open vs no view |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 10,000 | 40,000 | 5.89 | 5.81 | 7.70 | 5.58 | 5.68 | 5.43 | -0.21 |
| 50,000 | 200,000 | 4.41 | 5.27 | 20.72 | 5.99 | 4.59 | 5.24 | +0.18 |
| 100,000 | 400,000 | 3.32 | 5.17 | 15.33 | 6.23 | 3.67 | 5.13 | +0.35 |
| 300,000 | 1,200,000 | 3.52 | 6.47 | 21.30 | 6.86 | 3.50 | 6.01 | -0.02 |
| 1,000,000 | 4,000,000 | 3.80 | 6.21 | 98.45 | 6.33 | 3.20 | 6.14 | -0.60 |
| 1,000,000 | 8,000,000 | 2.82 | 5.77 | 146.58 | 14.77 | 3.23 | 5.67 | +0.41 |
| 1,000,000 | 16,000,000 | 3.84 | 5.45 | 345.68 | 14.48 | 3.18 | 5.55 | -0.56 |

The eager cost tracks the graph (98 -> 147 -> 346 ms as edges double), while the lazy open sits flat at ~3.2 ms and lands within +-0.6 ms of having no view at all at every size. That is the point: a deferred open reads nothing, so its cost does not depend on how big the thing it did not read is.

A 3M/12M and a 10M/40M row are still measuring as I open this; I will post them here when they land rather than extrapolate. What is already measured at 10M on #6588 as merged, from the benchmark comments you quoted in #6632, is the eager side: **983.88 ms** open with a view that is never queried, against **3.04 ms** with no view.

### The restore is not removed, it moves

Same fixture, 200k vertices / 800k edges, opening and then running the same query twice:

| arm | open | query 1 | query 2 |
|:--|---:|---:|---:|
| view, EAGER | 17.46 | 3.97 | 0.52 |
| view, LAZY | 5.92 | 14.39 | 0.50 |

Eager pays 11.5 ms at open that lazy does not; lazy pays 10.4 ms on the first query that eager does not. The second query is identical either way. The work is conserved - what changes is that a session which never asks is no longer charged for it.

### Is the view actually live afterwards, and does it still give the same answer?

| vertices | edges | status after open | status after first query | verdict |
|---:|---:|:--|:--|:--|
| 10,000 | 40,000 | `NOT_BUILT` | `READY` | view is live |
| 200,000 | 800,000 | `NOT_BUILT` | `READY` | view is live |

| vertices | edges | no view | view EAGER | view LAZY | |
|---:|---:|---:|---:|---:|:--|
| 10,000 | 40,000 | 160,000 | 160,000 | 160,000 | agree |
| 200,000 | 800,000 | 16 | 16 | 16 | agree |

The status column is the check that discriminates, and it is why the tests assert on state rather than on counts: row counts agree whether the view served the query or the ordinary path did, so counts alone cannot separate "loaded and used" from "silently skipped".

## Design

The hook is `GraphTraversalProvider.tryLazyActivate()`, called from the single chokepoint in `GraphTraversalProviderRegistry.findProvider()`. It defaults to `false` on the interface, so a provider that does not implement it is skipped exactly as it always was, and `GraphAnalyticalView`'s override is itself gated on the flag, so with the flag off every state returns `false` and the default path is unchanged. `findProviderIfReady()` is the non-triggering variant added for the planner.

## Behaviour changes, disclosed

With the flag **off**, nothing changes. `lazyRestoreIsOffByDefaultSoOpenStillRestoresEagerly` pins that.

With the flag **on**:

- The first query that could use the view pays the restore, per the table above.
- A restore failure (missing file, changed definition, invalidated certificate) is discovered at first use rather than at open, and falls back to `buildAsync()` exactly as the eager path does, with that query running unaccelerated.
- **This cuts against the change in one case and I would rather state it than have you find it.** Where there is no usable persisted CSR, the eager path has the rebuild running from `open()`, so a session that opens and then queries some seconds later can find the view ready. Under lazy the rebuild does not start until the first query asks, so that session now misses. Laziness cannot both do no work at open and have the work already done. That is the price, and it is the main reason the default is `false`.
- Views are likewise not built at open when `GAV_PERSIST_CSR` is disabled entirely, for the same reason.

## Testing

Five new tests in `GraphAnalyticalViewTest`: first-use loading, flag off by default, the planner's readiness check not triggering a load, the stale view left alone, and the await timeout not being waited out.

Two of them I checked fail on unpatched code, since an assertion that passes either way is decoration:

| test | with the guard | without it |
|:--|:--|:--|
| `lazyActivationLeavesAStaleViewAlone` | `STALE` | `BUILDING` |
| `lazyRestoreDoesNotWaitOutTheRestoreAwaitTimeout` | 1.79 s | 60.33 s against a 20 s tripwire |

The timing one uses `StallAwareStopwatch.assertGaveUpWithin()` as a tripwire between "did not wait" and "waited out the budget", not as a latency budget, per `CLAUDE.md`.

Regression run: `GraphAnalyticalView*`, `GraphTraversalProvider*`, `CSR*`, `Issue6049*`, `StatisticsProviderTest`, `GraphStatisticsCacheTest`, `CypherGAV*`, `Cypher*Cardinality*`, `CypherOptimizer*`, `AnchorSelectorTest` - **242 tests, 0 failures**, on the rebased tree.

## Open questions for you

- Should this ever be the default? I have left it off. The argument for on is that the cost is invisible to whoever pays it; the argument against is the no-persisted-CSR case above.
- Would you rather have it per-view (`CREATE GRAPH ANALYTICAL VIEW ... LAZY`) than database-wide? That is arguably the more honest shape, since laziness is a statement about one view's usage pattern, but it is a bigger surface and I did not want to presume the DDL.

## Environment

Intel Core Ultra X9 388H, 12 threads pinned via `--cpuset-cpus 0-11`, 30 GiB RAM, Samsung NVMe SSD, Ubuntu. `eclipse-temurin:25-jdk`, `-Xms8g -Xmx16g`. Engine at `ba84722bd` (#6588's head) plus this change. Fixtures are synthetic: N vertices of type `P` with a unique index on `id`, `fanout` edges of type `E` from each vertex to `(i + f * 7919) mod N`, `UPDATE MODE OFF`, CSR persisted by a clean close before measuring.

### DIFF
--- a/docs/6632-gav-lazy-restore.md
+++ b/docs/6632-gav-lazy-restore.md
@@ -1,0 +1,93 @@
+# Issue #6632 (item 1): an open() still loads a Graph Analytical View's CSR even when the session never uses it
+
+Issue: https://github.com/ArcadeData/arcadedb/issues/6632
+
+Addresses item 1 only. Item 2 of that issue is a documentation gap in `arcadedb-docs` for the
+pre-existing `arcadedb.gavRestoreAwaitTimeout` setting, which is a different repository.
+
+## Root cause / analysis
+
+`LocalDatabase.open()` calls `GraphAnalyticalViewPersistence.restoreAll(database)`, which for each
+persisted definition calls `builder.skipPersistence().restoreFromDiskOrBuildAsync()`. Since #6588 (merged) that
+reads the persisted CSR from disk instead of rebuilding it by a full graph scan, which is a large win:
+at 10M vertices / 40M edges the scan was 52,923 ms and the restore is 984 ms, 53x cheaper.
+
+The read still happens on the open path, unconditionally, for every session. A database that HAS a view
+therefore pays to load it at every open whether or not that session goes on to use it. On the same 10M
+graph an open with no view configured at all is ~3 ms, so a session that opens the database, reads a
+document type and closes spends ~984 ms of its ~987 ms lifetime loading a structure it never touches.
+
+That is a different question from the one #6583 filed, not a smaller version of it. Creating a Graph
+Analytical View declares that it should be *available*; it is not a statement that every process which
+subsequently opens the database intends to traverse it. The sessions that pay and get nothing are the
+short-lived ones, which are exactly the ones where a fixed cost dominates: a CLI query against a
+document type, a backup, a schema inspection, a health check, an embedded session in a test suite that
+opens and closes a database per test.
+
+## Fix
+
+Add a new `SCOPE.DATABASE` setting, `GAV_LAZY_RESTORE` (`arcadedb.gavLazyRestore`, `Boolean`, default
+`false`). When `false` (the default), behaviour is unchanged: `open()` reads the persisted CSR exactly as
+it does today. When `true`, `restoreFromDiskOrBuildAsync()` registers the view and reads nothing, and the
+first lookup that would actually use the view pays the restore instead.
+
+The hook is a new `GraphTraversalProvider.tryLazyActivate()` default method, called from the single
+chokepoint in `GraphTraversalProviderRegistry.findProvider()` when a provider is not ready. It returns
+`false` on the interface, so a provider that does not implement it is skipped exactly as it always was,
+and `GraphAnalyticalView`'s override is itself gated on the flag, so with the flag off every state
+returns `false` and the default path is untouched.
+
+Three callers deliberately do *not* trigger a load:
+
+- **The query planner.** `StatisticsProvider.exactMeanFromTraversalProvider()` asks whether acceleration
+  exists while *costing a plan*. It now uses a new non-triggering `findProviderIfReady()`. Loading a
+  structure there would move a multi-hundred-millisecond stall out of `open()`, where it is predictable
+  and happens once, and into an arbitrary query's planning phase.
+- **A `STALE` view.** Staleness is not-ready for a completely different reason than
+  deferred-and-not-yet-loaded. Whether a stale view may serve a query belongs to `GAV_USE_WHEN_STALE` and
+  whether it gets refreshed belongs to the update mode; a first-use load must not override either by
+  starting a rebuild that would not otherwise run.
+- **`restoreAll()`'s `GAV_RESTORE_AWAIT_TIMEOUT` wait.** A deferred view is `NOT_BUILT` with no build in
+  flight, and `readyLatch` is created at construction rather than left null, so `awaitReady()` would sit
+  on a latch nobody counts down and burn the whole budget before returning `false`. The two settings on
+  together would otherwise give an `open()` that blocks for the full timeout *and* still has not loaded
+  the view.
+
+## Behaviour changes when the flag is on
+
+- The first query that could use the view pays the restore. The work is conserved, not removed: it
+  reappears in full on that query and the second query is unaffected.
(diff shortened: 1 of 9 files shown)


tae898:
Closing in favour of #6633, which does the same thing and is yours. I had this built before that one appeared and did not see it in time.

I have left the one finding from building it on #6633 rather than here: with the view published as `READY` before the file is read, `StatisticsProvider` reaches the deferred read through `findProvider()` during plan costing, where `checkBuilt()` blocks without a timeout. The non-triggering `findProviderIfReady()` from this branch is offered there if it is wanted.

Branch stays at humemai/arcadedb-embedded-python@feat/6632-gav-lazy-restore if any of the tests are useful, in particular the two that fail on unpatched code: a stale view moving to `BUILDING`, and `GAV_RESTORE_AWAIT_TIMEOUT` being waited out in full (60.33 s against a 20 s tripwire).

Issue #5798, opened 2026-08-03
Several string functions silently coerce non-STRING inputs instead of raising a type error


shulei5831sl:
### ArcadeDB version

ArcadeDB `26.8.1-SNAPSHOT`

Build:
```text
85eb71bb19a39fd1eb36a6fe3ee742862d47d6f6/1785731865672/main
```
### Environment

* Linux x86_64
* Docker
* OpenJDK
* `arcadedata/arcadedb:latest`, freshly pulled on 2026-08-03
* HTTP command API
* Query language: Cypher

### Description

Several string functions silently convert non-`STRING` input values into text instead of reporting a type error.

For example:

```cypher
RETURN toUpper(5) AS value;
```

returns the string `"5"`.

Other non-string values are also accepted:

```cypher
RETURN toUpper(true) AS value;
// "TRUE"
```

```cypher
RETURN toUpper([1,2]) AS value;
// "[1, 2]"
```

```cypher
RETURN split(5, ',') AS value;
// ["5"]
```

The affected functions include:

* `toUpper()`
* `toLower()`
* `trim()`
* `lTrim()`
* `rTrim()`
* `split()`
* `replace()`

In each case, the function's primary text input accepts values outside its declared `STRING` domain and processes a textual representation of the value.

This differs from explicit conversion. Cypher already provides `toString()` for callers that intentionally want to convert supported scalar values before applying a string operation:

```cypher
RETURN toUpper(toString(5)) AS value;
// "5"
```

Silently performing that conversion inside the string function turns invalid type usage into an apparently valid result.

### Steps to reproduce

No graph data is required. Run each query separately.

#### Integer input

```cypher
RETURN toUpper(5) AS value;
```

Observed:

```text
"5"
```

#### Boolean input

```cypher
RETURN toUpper(true) AS value;
```

Observed:

```text
"TRUE"
```

#### List input

```cypher
RETURN toUpper([1,2]) AS value;
```

Observed:

```text
"[1, 2]"
```

#### Non-string input to `split()`

```cypher
RETURN split(5, ',') AS value;
```

Observed:

```text
["5"]
```

#### Additional affected functions

```cypher
RETURN toLower(5) AS value;
// "5"
```

```cypher
RETURN trim(5) AS value;
// "5"
```

```cypher
RETURN lTrim(5) AS value;
// "5"
```

```cypher
RETURN rTrim(5) AS value;
// "5"
```

```cypher
[...]

PR #6643, opened 2026-08-23
fix(#6609): left()/right()/substring() leak ClassCastException as HTTP 500 for list arguments

### LINKED ISSUE #6609: left() and right() leak ClassCastException as HTTP 500 for list arguments
### ArcadeDB version

ArcadeDB `26.9.1-SNAPSHOT`

### Environment

- Linux x86_64, Docker, OpenJDK
- Freshly built from the latest main branch (2026-08-23)
- HTTP command API, Cypher language
- Also reproduces on the released `26.8.1` build

### Description

Passing a LIST where `left()`, `right()` and `substring()` expect an INTEGER
argument does not produce a client-facing type error.  Instead the unchecked
cast escapes as `java.lang.ClassCastException` and the HTTP layer answers
500 "Error on transaction commit".

The related type-validation fix #5798 (PR #5901) covers string functions and
its own documentation lists `left()`/`right()` as a follow-up scope boundary;
`substring()` shares the same unchecked numeric cast path.

### Steps to reproduce

No graph data is required.  Run each query separately.

```cypher
RETURN left('a', []) AS r;
```

Actual:

```text
HTTP 500
{"error":"Error on transaction commit",
 "exception":"java.lang.ClassCastException",
 "detail":"class java.util.ArrayList cannot be cast to class java.lang.Number ..."}
```

```cypher
RETURN right('a', []) AS r;
```

Actual:

```text
HTTP 500
java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.Number
```

```cypher
RETURN substring('aa', 0, []) AS r;
```

Actual:

```text
HTTP 500
java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.Number
```

### Expected behavior

A LIST at an INTEGER argument position should be rejected with a client-facing
type error (HTTP 400), like all three reference engines do:

```cypher
RETURN left('a', []) AS r;
-- Neo4j:     22N27 data exception: invalid input LIST, expected INTEGER
-- Memgraph:  'left' argument at position 2 must be 'null' or 'non-negative integer'
-- FalkorDB:  Type mismatch: expected Integer or Null but was List
```

Normal input continues to work:

```cypher
RETURN left('a', 1) AS r;
-- "a"
```

The build already has a working client-facing type-error path:

```cypher
RETURN toUpper(5) AS r;
-- HTTP 400 (type error path from #5798 works on this build)
```

### Actual behavior

An internal `java.lang.ClassCastException` escapes to the client as
HTTP 500 "Error on transaction commit" for all three functions, on all three
fresh verification runs.

### Cross-engine comparison

| Query | Neo4j 2026.06 | Memgraph 3.12 | FalkorDB 4.20.1 | ArcadeDB 26.9.1-SNAPSHOT |
| `left('a', [])` | error (22N27) | error | error | **BUG: HTTP 500 ClassCastException** |
| `right('a', [])` | error (22N27) | error | error | **BUG: HTTP 500 ClassCastException** |
| `substring('aa', 0, [])` | error (22N27) | error | error | **BUG: HTTP 500 ClassCastException** |

### Isolated recheck

Freshly built instance from latest main, every reproducer run 3 times
(2026-08-23).  Reproduction rate 9/9, 0 transport failures, all normal-input
controls pass.  Same manifestation on the released `26.8.1` build.

### Why this looks like a bug

User-level invalid input escapes as an internal server failure (HTTP 500 with
a raw JDK exception class) instead of a client-facing type error.  The same
build rejects other invalid types with HTTP 400 (for example `toUpper(5)`),
and all three reference engines return user-facing errors for these exact
queries.  This is an error-handling violation, not a silent-acceptance issue.

### Impact

Invalid type usage produces a server-side error with no actionable client
information.  Applications cannot distinguish user errors from internal
failures, and the leaked exception class couples the client to JVM internals.

lvca:
## What does this PR do?

Makes `left()`, `right()` and `substring()` reject a non-numeric length/start argument (a LIST,
BOOLEAN, MAP, STRING, ...) with a client-facing `CommandSemanticException` (400) instead of letting
an unchecked `(Number) args[i]` cast throw `java.lang.ClassCastException`, which escaped as HTTP 500
"Error on transaction commit".

Closes #6609

## Motivation

```cypher
RETURN left('a', []) AS r;
```

used to throw a raw `ClassCastException` from `LeftFunction.execute()`. `left()`/`right()`/`substring()`
declare their length/start argument as `INTEGER | FLOAT`, so handing them anything else is the client's
mistake, not an internal fault. Neo4j (`22N27`), Memgraph and FalkorDB all reject these exact queries
with a client type error.

This is the same class of defect already fixed for the pure-numeric family (`abs()`, `round()`, ...) in
#5484, and for the STRING family (`toUpper()`, `split()`, ...) in #5798/#5901. #5901's own description
named `left()`/`right()` as a follow-up left out of its scope; `substring()` shares the same unchecked
numeric cast path.

## What changed

- `CypherFunctionHelper.NumericSignature` gained a `startArg` position (default 0, unchanged for every
  existing entry) so a function whose numeric arguments do not start at position 0 - `left(string,
  length)`, `right(string, length)`, `substring(string, start, length)` - can still be modeled and
  validated the same way as the rest of the numeric family, without misclassifying its leading STRING
  argument as numeric.
- `LeftFunction`, `RightFunction` and `CypherSubstringFunction` (the executor Cypher's `substring()`
  actually resolves to) now resolve their length/start argument(s) through the existing
  `CypherFunctionHelper.requireNumberArgument()` helper instead of an unchecked cast. Its unregistered
  twin `com.arcadedb.function.text.SubstringFunction` - exercised directly by unit tests and sharing the
  identical defect - got the same fix for consistency.
- `left`, `right` and `substring` were added to `NUMERIC_ARGUMENT_FUNCTIONS`, and
  `CypherSemanticValidator.checkStaticallyKnownNumericArgs()` now skips positions before a signature's
  `startArg`, so a statically-known literal (e.g. `RETURN left('a', [1,2])`) is rejected before the query
  runs even when no row would ever reach the projection - matching Neo4j and the existing #5484
  parse-time guarantee for the rest of the family.

The primary STRING argument of `left()`/`right()`/`substring()` (e.g. `left(5, 1)` still silently
converts `5` to `"5"`) is untouched: that is the separate follow-up #5901 explicitly left out of its own
scope, and #6609 only reports the numeric argument.

## Test plan

New test class `CypherLeftRightSubstringFunctionArgumentIssue6609Test`:

- [x] `left('a', [])`, `right('a', [])`, `substring('aa', 0, [])` and `substring('aa', [])` each reject
      with a `CommandSemanticException` naming the function and `LIST`.
- [x] The same for a BOOLEAN, STRING and MAP length argument.
- [x] A property whose runtime value is a LIST is rejected too (runtime path, not just literals).
- [x] A LIST literal is rejected even when no row matches (parse-time path).
- [x] `null` still propagates to `null`; ordinary numeric arguments still work; negative length is still
      reported the way #5296/#5793 already established (not as a type mismatch).
- [x] `left(5, 1)` still converts the primary argument, locking down that the out-of-scope gap isn't
      silently touched by this fix.

`CypherNumericFunctionArgumentIssue5484Test.everyNumericFunctionNameResolvesToANumericExecutorAndBack`,
which locks `NUMERIC_ARGUMENT_FUNCTIONS` and the executors returned by `CypherFunctionFactory` together,
is updated to recognize the three new entries.

Ran `mvn -o -pl engine -am test` for the full `com.arcadedb.query.opencypher.**` package plus the
directly affected text/Cypher-function suites: all green, no regressions.

## Checklist

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/cypher/CypherFunctionHelper.java
+++ b/engine/src/main/java/com/arcadedb/function/cypher/CypherFunctionHelper.java
@@ -70,30 +70,49 @@ public final class CypherFunctionHelper {
 
   /**
    * A numeric Cypher function: its canonical spelling - the one the runtime check uses, so both the parse-time and the
-   * runtime path phrase an error identically - and how many of its leading arguments are declared {@code INTEGER | FLOAT}.
-   * All of them for the whole family except {@code round(value, precision, mode)}, whose third argument is the STRING name
-   * of a rounding mode.
+   * runtime path phrase an error identically - and which of its arguments are declared {@code INTEGER | FLOAT}: the
+   * {@code numericArgs} positions starting at {@code startArg}. {@code startArg} is 0 for the whole family except
+   * {@code left()}/{@code right()}/{@code substring()}, whose leading argument is the STRING being sliced and whose
+   * numeric arguments start at position 1. {@code numericArgs} covers every remaining argument for the whole family
+   * except {@code round(value, precision, mode)}, whose third argument is the STRING name of a rounding mode.
    *
    * @param name        canonical function name, without parentheses
-   * @param numericArgs number of leading arguments that must be numeric, or {@link #ALL_ARGUMENTS}
+   * @param startArg    zero-based position of the first numeric argument
+   * @param numericArgs number of arguments, starting at {@code startArg}, that must be numeric, or {@link #ALL_ARGUMENTS}
    */
-  public record NumericSignature(String name, int numericArgs) {
+  public record NumericSignature(String name, int startArg, int numericArgs) {
+    public NumericSignature(final String name, final int numericArgs) {
+      this(name, 0, numericArgs);
+    }
+
+    /**
+     * Whether argument position {@code index} falls in this signature's numeric range.
+     */
+    public boolean isNumericPosition(final int index) {
+      return index >= startArg && (numericArgs == ALL_ARGUMENTS || index < startArg + numericArgs);
+    }
   }
 
   /**
    * The numeric Cypher functions, keyed by the lower-case name the parser produces, so that an argument already readable in
    * the query text can be rejected before the query runs, as Neo4j does. Kept in step with the numeric entries of
    * {@code CypherFunctionFactory.createCypherSpecificExecutor()}, which is what supplies the matching runtime check - a test
-   * asserts the two agree in both directions. See issue #5484.
+   * asserts the two agree in both directions. See issue #5484 (the pure-numeric family) and #6609 (left/right/substring's
+   * trailing numeric arguments).
    */
-  public static final Map<String, NumericSignature> NUMERIC_ARGUMENT_FUNCTIONS = Stream.concat(//
+  public static final Map<String, NumericSignature> NUMERIC_ARGUMENT_FUNCTIONS = Stream.of(//
           allArgumentsNumeric(//
               "abs", "ceil", "ceiling", "floor", "sqrt", "sign", "isNaN", //
               "exp", "log", "ln", "log10", //
               "sin", "cos", "tan", "asin", "acos", "atan", "atan2", "cot", "coth", "sinh", "cosh", "tanh", //
               "degrees", "radians", "haversin"), //
           // round(value, precision, mode): only the first two arguments are numeric.
-          Stream.of(new NumericSignature("round", 2)))//
+          Stream.of(new NumericSignature("round", 2)), //
+          // left(string, length) / right(string, length): the length is numeric, the leading string is not.
+          Stream.of(new NumericSignature("left", 1, 1), new NumericSignature("right", 1, 1)), //
+          // substring(string, start, length): start and the optional length are numeric, the leading string is not.
+          Stream.of(new NumericSignature("substring", 1, 2)))//
+      .flatMap(stream -> stream)//
       .collect(Collectors.toUnmodifiableMap(signature -> signature.name().toLowerCase(Locale.ROOT), signature -> signature));
 
   private static Stream<NumericSignature> allArgumentsNumeric(final String... names) {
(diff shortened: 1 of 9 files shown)


lvca:
Addressed the code review's actionable point: `com.arcadedb.function.text.SubstringFunction` (the unregistered twin `CypherSubstringFunction` mirrors) threw `CommandExecutionException` (HTTP 500) for a negative length while `CypherSubstringFunction` already threw `CommandSemanticException` (HTTP 400) for the identical condition. Aligned it and added regression tests for both the negative-length and LIST-argument paths (a2e38d7a0).

The other two notes were informational only, no action needed:
- The duplicated "Issue #6609..." comment across the four touched files - the review itself said this matches the codebase's existing per-file style and wasn't asking for a change.
- `SQLMethodLeft`/`SQLMethodRight` being a separate, unrelated SQL-method code path - correctly flagged as out of scope by design, not an oversight.

lvca:
Addressed the code review's actionable point: \`SubstringFunction\` also silently returned \`""\` for a negative \`start\` where \`CypherSubstringFunction\` throws \`CommandSemanticException\` for the same condition - same twin-inconsistency pattern as the negative-length fix above. Aligned it and updated the test that pinned the old silent-empty-string behavior (c85ac626d).

The other two notes (comment duplication style, the `Stream.of(...).flatMap(...)` construction being "functionally equivalent, just a different idiom") were explicitly called out by the review itself as non-blocking/stylistic, no action needed.

lvca:
Latest review has no blocking issues. Both minor notes are purely informational, no action needed:
- The \`requireNumberArgument\` null-check being unreachable from these call sites (belt-and-suspenders, since the shared helper is also used elsewhere with call sites that do need it).
- A note praising the two-arg convenience constructor for keeping the diff minimal on the unrelated pre-existing family members - not a suggestion to change anything.

Waiting on \`build-and-package\`/Meterian/java-kotlin analysis to finish.

PR #6601, opened 2026-08-23
fix(#6491): eagerly materialize MATCH before DELETE on disconnected patterns

lvca:
## Summary

Fixes #6491: `DETACH DELETE` after a `TRAIL`/self-loop fan-out raised
`RecordNotFoundException`.

A MATCH made of disconnected path pattern components (e.g. a self-loop
`(n)<-[]-(n)` cross-joined with an unrelated pattern, as produced by
`MATCH TRAIL ...` fan-out) can bind the very same underlying vertex/edge
from more than one output row - the disconnected component is
re-enumerated once per row of the other component. `DeleteStep` streamed
its input, so an earlier row's `DETACH DELETE` could remove that entity
while a later row was still being produced; the later row's own
`WHERE`/pattern evaluation then dereferenced the missing record and
surfaced as `RecordNotFoundException`.

Cypher requires the MATCH-established row set to be fixed before a
write clause in the same clause sequence takes effect (no different from
how Neo4j itself avoids this class of read/write interleaving hazard).
`DeleteStep` now fully reads the upstream row set before applying any
deletion, but **only** when the feeding MATCH clause(s) have disconnected
path patterns
(`MatchClause.hasDisconnectedPathPatterns()` - extracted from the
existing optimizer-eligibility check in `CypherExecutionPlanner` so both
call sites share one implementation). Ordinary connected-pattern
`MATCH ... DELETE` - the vast majority of real queries, including bulk
`MATCH (n) DETACH DELETE n` truncates - keeps the cheaper streaming path,
so this fix does not add memory/latency cost to the common case.

## Root cause

```
com.arcadedb.exception.RecordNotFoundException: Record #31:1 not found
  at ImmutableDocument.checkForLazyLoading / ImmutableVertex.checkForLazyLoading
  at PropertyAccessExpression.evaluate      (evaluating n3.k10)
  at MatchNodeStep$1.fetchMore              (self-loop candidate enumeration)
  ...
  at DeleteStep$1.fetchMore                 (previous row's DETACH DELETE n3)
  at AggregationStep.syncPull               (sum(...) pulls DeleteStep to completion)
```

`AggregationStep` (for `WITH sum(toFloatOrNull(1)) AS alias2` right after
the `DELETE`) drains `DeleteStep` row by row. Row 1's DETACH DELETE
removes vertex `n3`. When `AggregationStep` asks for row 2, the pull
re-enters `MatchNodeStep`, which re-enumerates the self-loop pattern's
[...]

PR #6644, opened 2026-08-23
fix(#6631): scope OpenCypher DELETE/FOREACH eager-materialization to its own WITH segment

### LINKED ISSUE #6631: OpenCypher: DELETE/FOREACH eager-materialization scope is statement-wide, can over-buffer unrelated bulk deletes
## Background

#6491 / #6601 fixed a real crash: a disconnected-pattern MATCH (a self-loop cross-joined with an unrelated pattern, or the same shape spelled as two separate `MATCH` clauses) could bind the same underlying vertex/edge across more than one output row. Streaming `DeleteStep`/`ForeachStep` let an earlier row's DETACH DELETE remove that entity while a later row was still being produced, throwing `RecordNotFoundException`.

The fix makes `DeleteStep` (and `ForeachStep`, for a FOREACH body containing a DELETE) fully read their upstream row set into memory before applying any deletion, gated on `MatchClause.hasDisconnectedPathPatterns(List<MatchClause>)` (`engine/src/main/java/com/arcadedb/query/opencypher/ast/MatchClause.java`) via `CypherExecutionPlan.matchClausesHaveDisconnectedPatterns` (`engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java`).

## The gap

That gating check is deliberately **statement-wide**: it returns true if *any* MATCH clause anywhere in the statement has disconnected path patterns, not just the MATCH clause(s) that actually feed a given DELETE segment. This is documented explicitly in the Javadoc as a conservative simplification (it can only force unnecessary eager materialization, never miss a case that needs it), and every reviewer of PR #6601 agreed it's the right tradeoff given how rare disconnected patterns are in practice.

But it does mean a multi-segment statement (`WITH`-separated `MATCH`/write pairs) where *any* segment has an incidental, unrelated disconnected-pattern MATCH will force **every** DELETE (or FOREACH-wrapped DELETE) in that statement onto the eager, whole-row-set-buffered path - including an otherwise-cheap bulk delete elsewhere in the same query that has nothing to do with the disconnected pattern.

Example shape that would over-buffer today:

```cypher
MATCH (a), (b) WHERE a.tag = 'x' AND b.tag = 'y'   // disconnected, small, incidental
WITH a, b
MATCH (n:HugeType)                                  // large, ordinary, connected
DETACH DELETE n                                     // pays the eager cost unnecessarily
```

## Proposed follow-up

Scope `matchClausesHaveDisconnectedPatterns` to the MATCH clause(s) that actually precede a given DELETE/FOREACH segment (e.g. by tracking segment boundaries via `clausesInOrder`) instead of the whole statement, so only a DELETE genuinely fed by a disconnected pattern pays the eager-materialization cost.

Not urgent - flagging so a future large-bulk-delete-with-an-unrelated-disconnected-match memory spike isn't mistaken for a fresh bug.

lvca:
## Summary

Fixes #6631.

`CypherExecutionPlan.matchClausesHaveDisconnectedPatterns()` - the gate that decides whether a `DeleteStep`/`ForeachStep` must fully buffer its upstream row set before deleting anything (added in #6601 to fix issue #6491) - was checking every `MATCH` clause in the **whole statement**, not just the one(s) feeding a given `DELETE`/`FOREACH` segment.

In a multi-segment statement (`WITH`-separated `MATCH`/write pairs), an incidental, unrelated disconnected-pattern `MATCH` anywhere in the statement forced **every** `DELETE`/`FOREACH` onto the eager, whole-row-set-buffered path - including an otherwise-cheap bulk delete elsewhere that has nothing to do with the disconnected pattern:

```cypher
MATCH (a), (b) WHERE a.tag = 'x' AND b.tag = 'y'   // disconnected, small, incidental
WITH a, b
MATCH (n:HugeType)                                  // large, ordinary, connected
DETACH DELETE n                                     // paid the eager cost unnecessarily
```

## Fix

Both `buildExecutionStepsWithOptimizer()` and `buildExecutionStepsWithOrder()` already walk the statement's clauses in order (`clausesInOrder`). Each now tracks the `MATCH` clauses seen since the last `WITH` boundary (`currentSegmentMatchClauses`), clearing it at every `WITH`, and passes only that segment-scoped list to `matchClausesHaveDisconnectedPatterns()` at each `DeleteStep`/`ForeachStep` construction site - exactly the approach the issue proposed. A `DELETE`/`FOREACH` now only pays the eager-materialization cost when the disconnected pattern is actually upstream of it in the same segment.

This is a pure precision fix: it can only turn some previously-eager paths into cheaper streaming ones. It never relaxes the guard for a `DELETE`/`FOREACH` that is genuinely fed by a disconnected-pattern `MATCH` in its own segment (covered by regression tests below and by the pre-existing #6491 suite, which still passes unchanged).

## Test plan

- [x] New `OpenCypherDeleteSegmentScopedEagerMaterializationIssue6631Test`, which reflects the private `eagerMaterialize` field off the built `DeleteStep`/`ForeachStep` (obtained via `PROFILE`'s execution plan) to assert:
  - a `DELETE`/`FOREACH` fed only by a connected `MATCH`, with an unrelated disconnected-pattern `MATCH` in an earlier `WITH` segment, is **not** eagerly materialized (fails without the fix, passes with it - verified by reverting the fix and re-running)
  - a `DELETE`/`FOREACH` genuinely fed by a disconnected-pattern `MATCH` in its own segment still **is** eagerly materialized (regression guard for #6491)
- [x] Existing `OpenCypherTrailDetachDeleteIssue6491Test` (4 tests) still passes unchanged
- [x] `MatchClauseTest` (8 tests) still passes unchanged
- [x] Broader DELETE/FOREACH/WITH-related suite (`CypherDeletedNodeWriteTargetIssue5795Test`, `CypherSequentialDeleteSetIssue5097Test`, `CypherWithStarScopeIssue6311Test`, `Issue4993ForeachCaseExistsTest`, `Issue5114ForeachReturnVisibilityTest`, `Issue5166ForeachCreateCountTest`, `Issue5287WithDistinctOrderByScopeTest`, `OpenCypherDeleteTest`, `OpenCypherForeachTest`, `WithAndUnwindTest`) passes unchanged
- [x] Full `com.arcadedb.query.opencypher.**` package (417 test classes, including the OpenCypher TCK suite) passes with zero failures
- [x] `mvn -o -pl engine compile` clean

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -1422,6 +1422,7 @@ public class CypherExecutionPlan {
             applyProjectionToScope(nextWith.getItems(), boundVariables);
             // the skipped WITH starts a new segment (issue #6631)
             closeMatchSegment(currentSegmentMatchClauses, disconnectedTaintedVariables);
+            propagateTaintThroughRenames(nextWith, disconnectedTaintedVariables);
             break;
           }
 
@@ -1436,6 +1437,7 @@ public class CypherExecutionPlan {
             applyProjectionToScope(nextWith.getItems(), boundVariables);
             // the skipped WITH starts a new segment (issue #6631)
             closeMatchSegment(currentSegmentMatchClauses, disconnectedTaintedVariables);
+            propagateTaintThroughRenames(nextWith, disconnectedTaintedVariables);
             break;
           }
         }
@@ -1452,6 +1454,9 @@ public class CypherExecutionPlan {
         // hazard for that variable, since rows still flow through it one at a time rather than being
         // fully consumed; closeMatchSegment() taints it before the segment is cleared.
         closeMatchSegment(currentSegmentMatchClauses, disconnectedTaintedVariables);
+        // A rename (WITH n AS m) doesn't change how rows flow either - propagate the taint onto the
+        // new name too, or a later DELETE of m would find nothing tainted under that name.
+        propagateTaintThroughRenames(withClause, disconnectedTaintedVariables);
         break;
 
       case MERGE:
@@ -2567,6 +2572,28 @@ public class CypherExecutionPlan {
               disconnectedTaintedVariables.add(relationship.getVariable());
         }
     currentSegmentMatchClauses.clear();
+  }
+
+  /**
+   * Propagates taint through a {@code WITH ... AS alias} rename: a rename doesn't change how rows flow
+   * any more than a same-name passthrough does (see {@link #closeMatchSegment}), so if the item being
+   * renamed is a bare reference to an already-tainted variable, its new alias is tainted too - otherwise
+   * a later DELETE of the alias would find nothing tainted under that name, even though it is exactly as
+   * hazardous as deleting the original variable would have been.
+   * <p>
+   * Only a bare variable reference is recognised as "the same entity under a new name" ({@code WITH n AS
+   * m}); an item computed from a tainted variable by any other expression ({@code WITH n.id AS m}, {@code
+   * WITH count(n) AS m}) produces a value that is no longer that entity, so it carries no taint forward.
+   */
+  private static void propagateTaintThroughRenames(final WithClause withClause,
+      final Set<String> disconnectedTaintedVariables) {
+    if (disconnectedTaintedVariables.isEmpty())
+      return;
+    for (final ReturnClause.ReturnItem item : withClause.getItems())
+      if (!item.isStar() && item.getAlias() != null
+          && item.getExpression() instanceof VariableExpression variableExpression
+          && disconnectedTaintedVariables.contains(variableExpression.getVariableName()))
+        disconnectedTaintedVariables.add(item.getAlias());
   }
 
   /**
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -1132,7 +1132,11 @@ public class CypherExecutionPlan {
           final WithClause withClause = entry.getTypedClause();
           currentStep = buildWithStepForOptimizer(withClause, currentStep, context, functionFactory);
           // A WITH boundary starts a new segment: MATCH clauses before it no longer feed a DELETE
-          // that comes after it (issue #6631).
+          // that comes after it (issue #6631). Unlike buildExecutionStepsWithOrder()'s WITH case, a
+          // plain clear() here (no taint tracking for a DELETE that plainly forwards a disconnected
+          // variable through this WITH) is safe: CypherExecutionPlanner.shouldUseOptimizer() already
+          // refuses to build a physical plan at all when a mutating clause (DELETE included) follows any
+          // WITH, so this method never builds a DELETE fed by an already-cleared segment.
           currentSegmentMatchClauses.clear();
           break;
         }
@@ -1318,6 +1322,9 @@ public class CypherExecutionPlan {
     // MATCH clauses seen since the last WITH (or since the start), i.e. the ones that actually feed
     // whichever DELETE/FOREACH segment is reached next - see matchClausesHaveDisconnectedPatterns().
     final List<MatchClause> currentSegmentMatchClauses = new ArrayList<>();
+    // Variables bound by a segment that WAS disconnected, kept forever once tainted (even across a WITH
+    // that merely forwards them unchanged) - see closeMatchSegment() and deleteMayTargetTaintedVariable().
+    final Set<String> disconnectedTaintedVariables = new HashSet<>();
 
     // Both count push-downs answer from the schema and the CSR arrays alone: they read the statement's patterns and
     // never look at the incoming rows. That makes the enumerating form of them wrong the moment the seed row binds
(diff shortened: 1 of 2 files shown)


lvca:
Addressed the review:

**Legacy call site**: added a comment explaining why it's fine unscoped - it only runs when `clausesInOrder` is null/empty (never true for a real parsed multi-segment statement) and is single-segment by construction regardless.

**Test coverage / which path**: confirmed empirically that neither of my original tests exercised `buildExecutionStepsWithOptimizer()`'s WITH-clearing branch, because it's structurally unreachable for this shape - `CypherExecutionPlanner.shouldUseOptimizer()` already refuses the physical-plan path whenever any mutating clause (DELETE included) follows a WITH, so a DELETE fed by an already-cleared segment can never reach that method. Documented that invariant with a comment at the call site instead of adding a test for an unreachable branch.

Investigating that question surfaced a real bug in the fix itself, which is the more important outcome here: unconditionally clearing the tracked segment at every WITH loses the #6491 safety net when the WITH plainly forwards the disconnected match's own variable (e.g. `MATCH (o:Other), (n:Loop)<-[:SELF]-(n) ... WITH n, o DETACH DELETE n`) - rows out of a disconnected-pattern MATCH still flow one at a time through a non-aggregating WITH, so the hazard survives it. Verified with reflection on the built `DeleteStep`: `eagerMaterialize` came back `false` for that shape with the first version of this fix, which would have reintroduced the #6491 crash for anyone using a passthrough WITH between a disconnected-pattern MATCH and a DELETE of its own variable.

Fixed by tracking a persistent tainted-variable set: once a WITH-delimited segment is found disconnected, every variable it bound is tainted for the rest of the statement, and a DELETE/FOREACH is still eagerly materialized if its own target intersects that set, regardless of how many WITH clauses forwarded it in between. Added two new regression tests for this shape (DELETE and FOREACH), confirmed to fail without the fix and pass with it. Full `com.arcadedb.query.opencypher.**` package (418 test classes, 8514 tests) passes with zero failures.

lvca:
Both points addressed:

**Relationship variables**: confirmed the gap - reverted the fix locally, ran the new test, and it failed exactly as described (\`eagerMaterialize\` came back \`false\` for \`MATCH (a)-[r:REL]->(b), (c) WITH r DELETE r\`). \`closeMatchSegment()\` now also walks \`path.getRelationships()\`. Added the suggested regression test (DELETE of a relationship variable through a passthrough WITH); confirmed it fails without the fix and passes with it.

**Legacy call site**: replaced the unverified comment with an executable check - a new test parses the same multi-segment WITH-separated shape through the real production parser (\`Cypher25AntlrParser\`) and asserts \`getClausesInOrder()\` is populated, which is the actual precondition \`CypherExecutionPlan.buildExecutionSteps()\` checks before ever falling back to the legacy builder. That backs the claim with something that'll fail loudly if the parser's behavior ever changes, rather than a comment someone has to re-derive.

Full \`com.arcadedb.query.opencypher.**\` package (8516 tests) passes with zero failures.

lvca:
Confirmed and fixed. Reverted the two \`propagateTaintThroughRenames()\` call sites locally first and ran your exact repro (\`WITH n AS m ... DELETE m\`) - came back \`eagerMaterialize = false\` as predicted, then restored the fix and re-verified it comes back \`true\`.

New \`propagateTaintThroughRenames()\` helper: at each WITH, if an item has a non-null alias and its expression is a bare \`VariableExpression\` referencing an already-tainted variable, the alias gets tainted too. Deliberately narrow to a plain rename - an item computed by any other expression (\`n.id\`, \`count(n)\`) produces a value that's no longer "the same entity," so it doesn't carry taint forward. Handles rename chains correctly too (\`WITH n AS m WITH m AS k DELETE k\`) since the set is checked and updated at every WITH, not just the one that closes the disconnected segment.

Added the suggested regression tests (renamed node variable and, for the same reason the earlier round added a relationship-variable case, a renamed relationship variable too) - both confirmed to fail without the fix and pass with it.

Full \`com.arcadedb.query.opencypher.**\` package (8519 tests) passes with zero failures.

lvca:
Thanks for the thorough trace-through, especially confirming the OPTIONAL MATCH count fast paths are safe by construction (shared-variable requirement) rather than by luck.

**Misplaced Javadoc**: fixed - moved it back above \`aMultiSegmentWithSeparatedStatementAlwaysPopulatesClausesInOrder()\`.

**\`WITH *\` test**: added, as suggested. It passes, confirming \`closeMatchSegment()\`'s direct same-name taint covers the star-passthrough case without needing \`propagateTaintThroughRenames()\` (which correctly skips star items, since there's no single alias/expression pair to inspect for those).

Both design/precision notes (conservative treatment of non-trivial DELETE targets, permanent taint even where an aggregating WITH could theoretically discharge it) are judged not worth implementing - they're documented, intentional, safe-direction tradeoffs for a hazard that's already rare, and adding precision there would be exactly the kind of bookkeeping the original issue said isn't worth it for this rare case.

Full \`com.arcadedb.query.opencypher.**\` package (8520 tests) passes with zero failures.

lvca:
Thanks for the thorough cross-check across all 5 commits.

On the one corner case flagged (a tainted name later reused by an unrelated MATCH/rename): agreed this is fine to leave as-is - it can only cost an unnecessary eager-materialization pass for a name reuse that's already rare on top of an already-rare disconnected pattern, never an incorrect result. Not implementing further scoping for it, consistent with the issue's own "not worth the bookkeeping for this rare case" framing.

Waiting on the full test matrix (build-and-package's downstream jobs) to finish before this is ready to merge.

PR #6650, opened 2026-08-23
fix(sql): EXPLAIN of a write statement runs it when chained via sqlscript or nested blocks

### LINKED ISSUE #6648: EXPLAIN executes the wrapped write statement instead of just planning it when language is sqlscript
## Summary

`EXPLAIN <write statement>` executes the wrapped statement instead of just planning it, when the command is submitted with `"language": "sqlscript"` (or run inside a `sqlscript` block generally). Plain `"language": "sql"` is unaffected.

Reported by a client running `26.9.1-SNAPSHOT`: an `EXPLAIN UPDATE ...` sent via `sqlscript` actually ran the `UPDATE`, which had no `WHERE`-selective match and ran unbounded for hours until the server exhausted memory and the pod had to be restarted. Switching to `"language": "sql"` was confirmed as a working workaround, and a post-crash `SELECT ... GROUP BY` confirmed the WAL rollback on restart left the data consistent.

## Reproduction

```bash
curl -s -u root:<password> -X POST 'http://<host>:2480/api/v1/command/<db>' \
  -H "Content-Type: application/json" \
  -d '{"language": "sqlscript", "command": "EXPLAIN UPDATE Foo SET bar = 1;"}'
```

`Foo.bar` is actually set to `1` on every matching record. The same request with `"language": "sql"` and no trailing `;` correctly returns a plan and performs no write.

## Root cause

- `ExplainStatement.execute(...)` (`engine/src/main/java/com/arcadedb/query/sql/parser/ExplainStatement.java:47-76`) is where EXPLAIN's actual semantics live: it builds a fresh `CommandContext` with profiling on, calls `statement.createExecutionPlan(context)` to get the plan, and wraps it in an `ExplainResultSet` **without ever iterating/pulling it** - so a write statement's plan is built but never run.
- `ExplainStatement.createExecutionPlan(CommandContext)` (lines 78-81) is a pure passthrough: `return statement.createExecutionPlan(context);`. Calling this method directly - bypassing `execute()` - silently discards EXPLAIN semantics and just hands back the wrapped statement's own plan.
- `SQLQueryEngine` (plain `"sql"`) calls `statement.execute(...)` on the parsed statement (`engine/src/main/java/com/arcadedb/query/sql/SQLQueryEngine.java:86-148`), so `ExplainStatement`'s override fires correctly.
- `SQLScriptQueryEngine.executeInternal(...)` (`engine/src/main/java/com/arcadedb/query/sql/SQLScriptQueryEngine.java:178`) does not: for every statement in the script it calls `stm.createExecutionPlan(scriptContext)` directly and chains the resulting sub-plan into the script's `ScriptExecutionPlan`. There is no `instanceof ExplainStatement` (or otherwise EXPLAIN-aware) branch anywhere in this class, `ScriptExecutionPlan`, or `RetryExecutionPlan`. So `EXPLAIN UPDATE ...` inside a script is parsed correctly, but executed exactly as if the `EXPLAIN` keyword were not there: the wrapped `UPDATE` plan is chained into the script and run to completion when the script plan is pulled.

`PROFILE` is not affected the same way: its `createExecutionPlan()` override is also a passthrough, but `PROFILE`'s whole purpose is to execute the statement, so running it inside `sqlscript` is at least semantically closer to correct - it just skips the profiling/`ExplainResultSet` wrapping, not a correctness/safety issue like EXPLAIN's silent-write case.

This is unrelated to the EXPLAIN fix in #6323 (merged `b8169c3c7`), which improved what EXPLAIN's *description output* contains for openCypher plans - it did not touch the SQL script execution path, and this bug predates it.

## Suggested fix

`SQLScriptQueryEngine.executeInternal` should route `ExplainStatement` (and arguably `ProfileStatement`, for consistent output) through `execute()` rather than `createExecutionPlan()`, or otherwise special-case them so the chained sub-plan reflects EXPLAIN's non-executing contract instead of running the wrapped statement.

## Impact

A read-only-looking `EXPLAIN <UPDATE|DELETE|INSERT|...>` sent via `sqlscript` silently performs the write. Combined with an unselective `WHERE`, this can run an unbounded update to exhaustion (OOM, server crash) with no indication to the caller that anything other than planning happened.

lvca:
## Summary

Fixes #6648.

`EXPLAIN <write statement>` silently ran the wrapped write whenever it was reached through a chaining caller instead of `Statement.execute()`: the `sqlscript` language engine, or a `FOREACH`/`IF`/`WHILE`/`RETRY`/`LET` block nested inside any script. Plain `"language": "sql"` was never affected.

## Root cause

`ExplainStatement.createExecutionPlan(CommandContext)` was a pure passthrough (`return statement.createExecutionPlan(context);`). Every one of those chaining callers builds a nested statement's plan by calling `createExecutionPlan()` directly (not `execute()`) and then pulls whatever comes back to completion:
- `SQLScriptQueryEngine.executeInternal` chains it into the script's `ScriptExecutionPlan`.
- `ForEachStep`, `IfStep`, `WhileStep`, `RetryStep`, `LetQueryStep`, `GlobalLetExpressionStep`, `ParenthesisExpression` do the same for a statement nested inside a block.

So passing the wrapped statement's own executable plan straight through handed all of these the real `UPDATE`/`INSERT`/`DELETE`/... plan, and it ran to completion when pulled - exactly as if `EXPLAIN` were not there. `ExplainStatement.execute()` itself was fine: it builds the plan and wraps it in `ExplainResultSet` without ever pulling it.

## Fix

Rather than special-casing `SQLScriptQueryEngine` alone (the issue's suggested fix), this fixes it at the one place all of those callers go through: `ExplainStatement.createExecutionPlan()` now wraps the built plan in a new `ExplainExecutionPlan` (`engine/src/main/java/com/arcadedb/query/sql/executor/ExplainExecutionPlan.java`). On `fetchNext()` it reports the wrapped plan's description - the same `executionPlan`/`executionPlanAsString` shape `ExplainResultSet` gives callers that go through `execute()` - without ever pulling the wrapped plan. It's also deliberately not an `InsertExecutionPlan`/`DeleteExecutionPlan`/`UpdateExecutionPlan`/`DDLExecutionPlan`/`SingleOpExecutionPlan`, since `ScriptLineStep.syncPull` special-cases exactly those types to run them eagerly on first pull.

`ExplainStatement.execute()` unwraps the plan back out via `ExplainExecutionPlan.getWrappedPlan()` so it keeps returning `ExplainResultSet` exactly as before - no change to its public contract (the server's `PostCommandHandler` and existing tests do `instanceof ExplainResultSet` / cast to it).

`PROFILE` is intentionally left alone, per the issue: its whole purpose is to execute the wrapped statement, so it isn't a correctness/safety issue the way `EXPLAIN`'s silent write is.

## Test plan

Added to `ExplainStatementExecutionTest`:
- [x] `explainUpdateInSqlScriptDoesNotExecuteWrite` - reproduces the exact reported case
- [x] `explainInsertInSqlScriptDoesNotExecuteWrite`
- [x] `explainDeleteInSqlScriptDoesNotExecuteWrite`
- [x] `explainUpdateInsideIfBlockDoesNotExecuteWrite` - same bug via a different chaining caller (`IfStep`), confirming the fix belongs at the root and isn't just papering over the `sqlscript` case

Verified all four fail against the unpatched code (reverted the fix, kept the tests) and pass with it restored. Also ran the full `com.arcadedb.query.sql.executor.*Test` suite, `ExplainStatementExecutionTest`, `ScriptExecutionTest`, `SQLScriptAdditionalCoverageTest`, `MaxMinFromIndexTest`, `ProfileStatementExecutionTest`, `ProfileFlameGraphTest`, and `OpenCypherUnionCallProfileTest` - all green. `engine` and `server` modules compile clean.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ExplainExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ExplainExecutionPlan.java
@@ -39,21 +39,35 @@ import java.util.List;
  * DeleteExecutionPlan}/{@code UpdateExecutionPlan}/{@code DDLExecutionPlan}/{@code
  * SingleOpExecutionPlan}: {@link ScriptLineStep#syncPull} special-cases exactly those types to
  * eagerly run them, so staying outside that set is what keeps a chained EXPLAIN from being run.
+ * <p>
+ * The same reasoning covers {@code PROFILE EXPLAIN <statement>}: {@code ProfileStatement.execute()}
+ * is itself a caller that pulls whatever {@code statement.createExecutionPlan()} hands back, so an
+ * inner {@link com.arcadedb.query.sql.parser.ExplainStatement} nested under {@code PROFILE} used to
+ * be just as exploitable as the {@code sqlscript} case - {@code PROFILE EXPLAIN UPDATE ...} silently
+ * ran the update too. Wrapping here closes that path as well: EXPLAIN's non-execution contract wins
+ * regardless of what encloses it, so {@code PROFILE EXPLAIN <statement>} degrades to plan-only output
+ * with no execution and no real cost numbers, rather than {@code PROFILE} unwrapping the inner
+ * statement and running it for real.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
 public class ExplainExecutionPlan implements InternalExecutionPlan {
-  private final CommandContext context;
-  private final ExecutionPlan  wrappedPlan;
-  private       boolean        executed = false;
+  private final CommandContext        context;
+  private final InternalExecutionPlan wrappedPlan;
+  private       boolean               executed = false;
 
-  public ExplainExecutionPlan(final CommandContext context, final ExecutionPlan wrappedPlan) {
+  public ExplainExecutionPlan(final CommandContext context, final InternalExecutionPlan wrappedPlan) {
     this.context = context;
     this.wrappedPlan = wrappedPlan;
   }
 
-  public ExecutionPlan getWrappedPlan() {
+  public InternalExecutionPlan getWrappedPlan() {
     return wrappedPlan;
+  }
+
+  @Override
+  public void close() {
+    wrappedPlan.close();
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/ExplainExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/ExplainExecutionPlan.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.query.sql.executor;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Execution plan returned by {@code EXPLAIN <statement>} when it is not run through
+ * {@code Statement.execute()} directly, but instead chained into an enclosing plan - a script
+ * ({@code SQLScriptQueryEngine}), or a nested block (FOREACH, IF, WHILE, RETRY, LET). All of
+ * those callers treat whatever {@code Statement.createExecutionPlan(CommandContext)} returns as
+ * an executable plan and chain/pull it to completion, so an EXPLAIN whose {@code
+ * createExecutionPlan()} passed the wrapped statement's own plan straight through - as it used
+ * to - handed them something that, once pulled, actually ran the wrapped statement. For a write
+ * statement (UPDATE/INSERT/DELETE/...) that silently performed the write EXPLAIN promised not to
+ * (issue #6648).
+ * <p>
+ * This plan wraps the built-but-not-pulled plan of the wrapped statement and, on
+ * {@link #fetchNext(int)}, reports its description without ever pulling it - the same contract
+ * {@link com.arcadedb.query.sql.parser.ExplainResultSet} gives callers that go through {@code
+ * execute()}. It also deliberately is not an {@code InsertExecutionPlan}/{@code
+ * DeleteExecutionPlan}/{@code UpdateExecutionPlan}/{@code DDLExecutionPlan}/{@code
+ * SingleOpExecutionPlan}: {@link ScriptLineStep#syncPull} special-cases exactly those types to
+ * eagerly run them, so staying outside that set is what keeps a chained EXPLAIN from being run.
+ *
+ * @author Luca Garulli (l.garulli@arcadedata.com)
+ */
+public class ExplainExecutionPlan implements InternalExecutionPlan {
+  private final CommandContext context;
+  private final ExecutionPlan  wrappedPlan;
+  private       boolean        executed = false;
+
+  public ExplainExecutionPlan(final CommandContext context, final ExecutionPlan wrappedPlan) {
+    this.context = context;
+    this.wrappedPlan = wrappedPlan;
+  }
+
+  public ExecutionPlan getWrappedPlan() {
+    return wrappedPlan;
+  }
+
+  @Override
+  public ResultSet fetchNext(final int n) {
+    if (executed)
+      return new InternalResultSet();
+
+    executed = true;
+
+    final ResultInternal result = new ResultInternal();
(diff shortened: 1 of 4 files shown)


lvca:
Thanks for tracing this - the \`PROFILE EXPLAIN\` gap was real. Pushed a follow-up:

- **\`PROFILE EXPLAIN <statement>\`**: confirmed it's syntactically valid and was just as exploitable as the \`sqlscript\` case (\`PROFILE EXPLAIN UPDATE ...\` silently ran the update too, via \`ProfileStatement.execute()\` pulling whatever \`createExecutionPlan()\` handed back). Went with option (a): EXPLAIN's non-execution contract wins regardless of what encloses it - unwrapping in \`ProfileStatement\` to force real execution would reopen the exact silent-write risk this PR closes, one level deeper. That was already the resulting behavior; added \`profileExplainUpdateDoesNotExecuteWrite\` to pin it down as a regression test instead of leaving it untested.
- **\`close()\`**: tightened \`ExplainExecutionPlan\`'s wrapped-plan field from \`ExecutionPlan\` to \`InternalExecutionPlan\` (which is what \`Statement.createExecutionPlan()\` always actually returns) and added a \`close()\` override that forwards to it, for symmetry with the other \`InternalExecutionPlan\` implementations.
- Left the \`execute()\`-builds-then-unwraps allocation as is, per your own call that it's not a real concern on a non-hot path.

All tests green (7/7 in \`ExplainStatementExecutionTest\` plus the full \`ProfileStatementExecutionTest\`/\`ScriptExecutionTest\`/\`SQLScriptAdditionalCoverageTest\`/\`MaxMinFromIndexTest\` suites).

lvca:
Thanks for the thorough trace on this round. Addressed both points:

- **Test coverage**: added \`explainUpdateInsideForEachDoesNotExecuteWrite\`, \`explainUpdateInsideWhileDoesNotExecuteWrite\`, and \`explainUpdateInsideRetryBlockDoesNotExecuteWrite\`, same shape as the existing tests (insert a row, run \`EXPLAIN UPDATE\` through the caller, assert untouched). That's now all six chaining callers named across both reviews backed by a running test rather than by code tracing alone. Skipped \`ParenthesisExpression\`/the LET-subquery path - \`EXPLAIN\` is a statement, not an expression, so it can't appear inside a parenthesized subquery in the first place; there's nothing reachable to test there.
- **\`canBeCached()\`**: judged not worth changing. Nothing outside \`SelectExecutionPlan\`'s own step list ever calls \`InternalExecutionPlan.canBeCached()\` for an actual caching decision (checked all four call sites), and EXPLAIN can never appear as a step inside a \`SelectExecutionPlan\`'s chain - it's always a top-level statement. So there's no reachable path where this matters, and unconditional \`false\` already matches \`ScriptExecutionPlan\`/\`SingleOpExecutionPlan\`'s existing pattern for the same reason.
- The \`ProfileStatement\` overload asymmetry is acknowledged as out of scope (pre-existing, not touched by this PR), per your own framing.

10/10 in \`ExplainStatementExecutionTest\`, full related-suite run green.

lvca:
Good suggestion - added the \`LET $x = (EXPLAIN ...)\` test, and it turned up something real: neither \`ExplainStatement\` nor \`ProfileStatement\` ever implemented \`refersToParent()\` (\`Statement\`'s default throws \`UnsupportedOperationException\`), so \`SelectExecutionPlanner.splitLet()\` hard-failed for that combination regardless of what's wrapped. Fixed both by delegating to the wrapped statement - same pattern \`ExpressionStatement.refersToParent()\` already uses.

While tracking this down I also found the write statement side of it needs the same method: \`UpdateStatement\`, \`DeleteStatement\`, \`CreateVertexStatement\`, \`CreateEdgeStatement\`, and \`DDLStatement\` have never implemented \`refersToParent()\` either, so \`LET $x = (UPDATE ...)\` throws the identical exception whether or not it's wrapped in EXPLAIN - it's a separate, pre-existing, unrelated gap, not a silent-write risk (it fails loudly, not silently), and fixing every write statement's correlation semantics is a real, independent piece of work I've left out of this PR. Used \`INSERT\` for the new test instead, since \`InsertStatement.refersToParent()\` already returns \`false\` and it's the one write statement that actually reaches \`ExplainExecutionPlan\` through this path today - it exercises exactly what the suggestion was asking for (no execution through \`LetQueryStep\`/\`GlobalLetQueryStep\`).

11/11 in \`ExplainStatementExecutionTest\`, plus a full run of the parser package and the related executor suites, all green.

lvca:
Thanks - glad the \`PROFILE EXPLAIN\` wiring traced out cleanly. On the two non-blocking notes:

1. **\`getSteps()\` -> empty list, triple-nested EXPLAIN detail gap**: judged not worth changing. It's the same pattern \`SingleOpExecutionPlan\` already uses for the same reason - there's no step chain to enumerate for a synthetic one-row diagnostic wrapper. The gap only shows up for \`EXPLAIN\` nested two levels deep inside another statement's \`toResult()\` walk (e.g. \`LET $x = (EXPLAIN INSERT ...)\` inside a SELECT whose own plan gets explained) - a narrow, diagnostic-only edge case, not a correctness issue. Not worth the added complexity of making \`getSteps()\` expose the wrapped plan's steps without ever pulling them.
2. **JSON shape difference between top-level \`EXPLAIN\` and \`PROFILE EXPLAIN\`**: also judged not worth a change - your own note frames it as conditional on external tooling depending on \`executionPlan.type\`, and EXPLAIN/PROFILE are diagnostic-only surfaces with no such contract today. Agreed it's a strict improvement over the old behavior regardless (the old behavior for \`PROFILE EXPLAIN <write>\` was "silently run the write," not "return a slightly different JSON shape").

Both are intentional, no code changes on their own. Waiting on \`build-and-package\` (the actual test suite run) to go green before calling this ready.

Issue #6518, opened 2026-08-20
LSMVectorIndex.close() releases the build pool and then flushes, so a dirty index leaks a whole new ForkJoinPool past close()


lvca:
`LSMVectorIndex.close()` runs its two steps in the opposite order to the one `LocalDatabase` uses, and the inversion is not cosmetic: on a dirty index it leaks a live `ForkJoinPool`.

```java
public void close() {
  releaseBackgroundResources();   // shuts the graph-build pool down, sets valid = false
  flush();                        // ...then asks for a graph build, which needs that pool
}
```

`LocalDatabase.closeDurableParts()` does the reverse, and says why:

> Deliberately AFTER the flush above (an index whose graceful shutdown is a graph build needs its build pool alive for it)

## What actually happens

`getOrCreateGraphBuildPool()` treats a shut-down pool as a pool to replace:

```java
if (pool != null && !pool.isShutdown() && pool.getParallelism() == wanted)
  return pool;
```

So `flush()` builds a **brand new** pool after `releaseBackgroundResources()` has just torn the old one down, runs the rebuild on it, and nothing ever shuts the replacement down - `close()` has already done its releasing.

Measured on a 200-vector index made dirty with one extra insert, calling `index.close()` directly:

```
before close: graphState=2 (MUTABLE) rebuilds=1 valid=true
after  close: graphState=1 (IMMUTABLE) rebuilds=2 valid=false
pool after close: shutdown=false terminated=false poolSize=17 parallelism=17
live ForkJoinPool threads: 17
```

The rebuild ran and the graph was persisted - so the failure is silent, and correctness-wise the close even "worked". It just left 17 worker threads running on an index that is closed and marked invalid. That is the exact failure #5418 was filed for ("the vector index inactivity rebuild timer outlived Database.close()"), reintroduced through a different door.

Note also that nothing on the build path checks validity: `buildGraphFromScratchExclusively()` has no `isValid()`/`checkIsValid()` guard, so an index with `valid = false` builds a full graph quite happily.

## Reachability

Latent today, which is why it has not bitten:

* `drop()` -> `close()` is safe by accident: `drop()` calls `vectorIndex.clear()` first, so `flush()` sees `size() == 0` and skips the build entirely.
* `TypeIndex.close()` -> `index.close()` has no production caller I can find.
* The database close path does not use `close()` at all - `closeDurableParts()` deliberately calls `flush()` and `releaseBackgroundResources()` separately, in the correct order.

[...]

Issue #6641, opened 2026-08-23
Graph Analytical View: after the deferred CSR restore, the first query following any intervening commit blocks through a full rebuild (5.6 s at 1M vertices) instead of falling back to OLTP in 5 ms


tae898:
Follow-up to #6632 / #6633, measured on `f412c84b0` (current main, with the deferred restore) against `bc5d31a63` (the merge of #6588, before it). Raised on #6633 after it had already merged, so filing here for tracking as offered there.

## What changed

Deferring the CSR read removed the cost from `open()`, which works: a session that never queries the view now pays 0.77 ms at 10M vertices / 40M edges instead of 1027.46 ms. That part is a clear win and I posted the full sweep on #6633.

The cost did not disappear from every path, though. When the persisted CSR turns out not to be usable, the first query now waits out a full O(V+E) rebuild on its own thread, where before it returned immediately from the ordinary OLTP path.

## Reproduction

Open a database with a persisted CSR, commit anything (which invalidates the freshness certificate), then query. 1,000,000 vertices / 4,000,000 edges.

| | `bc5d31a63` (eager) | `f412c84b0` (deferred) |
|:--|---:|---:|
| view status after the commit | `STALE` | `READY` |
| **first query** | **5 ms**, 7 ms | **5,613 ms** |
| second query | 0 ms | 0 ms |

Both return the same, correct answer. The difference is only who does the rebuild and whether the caller waits for it.

A damaged file reaches the same place. Writing 4 KB of junk into the midpoint of a 56 MB `gav-v1.csr` is rejected correctly and does not hang, but the first query then pays for the rebuild:

```
after open: status=READY
first query -> 4 in 5,817 ms
view: status=READY restoredFromDisk=false
```

So it is any unusable-CSR path, not only an invalidated certificate.

## Why the fallback stopped happening

Before deferral the snapshot existed by the time the commit landed, so `onRelevantCommit()` marked the view `STALE`, provider resolution rejected it (`isReady()` is `READY || (STALE && useWhenStale)`), and the query ran unaccelerated.

With deferral there is no snapshot for the commit to invalidate, so the view stays `READY`. Provider resolution accepts it, the query reaches `checkBuilt()`, the deferred read runs, `tryRestoreFromPersistedCsr()` correctly rejects the stale certificate, `buildAsync()` starts a full scan, and `awaitDeferredDiskRestoreSettled()` waits it out on the query thread with no timeout.

[...]

PR #6653, opened 2026-08-23
fix(vector): defer close-time graph rebuild for large mutated indexes

### LINKED ISSUE #6067: flush() re-runs the exact rebuild releaseBackgroundResources() just cancelled, synchronously and unprotected, undercutting #5872's fix

lvca:
## Summary

Fixes #6067.

`LSMVectorIndex.flush()`'s rebuild-before-close condition triggers whenever `graphState == MUTABLE` (i.e. any write since the last graph build), and always calls `buildGraphFromScratch()` **synchronously** on the closing thread - unlike the search path's `rebuildGraphBeforeSearch()`, which already prefers an async rebuild above `ASYNC_REBUILD_MIN_GRAPH_SIZE` (1000) vectors.

A comment on the issue measured the real cost: a single write to a 200,000-vector index made `close()` block for ~43 seconds, the same cost as writing a thousand vectors, because there's no incremental persist - any rebuild is a full one.

Note: the issue's original description (`close()` = `releaseBackgroundResources(); flush();`, with the second build racing an unprotected pool) predates #6518's fix, which already reordered `close()` to `flush(); releaseBackgroundResources();` so the build pool stays alive for the close-time rebuild. That race no longer applies as originally described - but the underlying synchronous-full-rebuild-on-close problem is real and is what this PR addresses, backed by the fresh benchmark in the issue thread.

### Fix

At or above `ASYNC_REBUILD_MIN_GRAPH_SIZE`, `flush()` now leaves `graphState` as-is instead of rebuilding synchronously. The graph is rebuilt lazily the next time the index is actually searched, reusing the existing stale-persisted-graph detection in `ensureGraphAvailable()` - the same path that already handles a stale graph on every ordinary reopen. The vectors themselves are already durably persisted (ordinary page/WAL writes, unrelated to the graph); only the graph *topology* is deferred.

- A session that closes and reopens an index without searching it again never pays the rebuild cost at all.
- A session that does search it pays the cost at the query instead of at `close()`.
- Below the threshold, `close()` keeps rebuilding synchronously as before (cheap, keeps the fast-restart benefit for small indexes).

This mirrors the size-based policy already used by the search path, rather than an all-or-nothing skip.

## Test plan

- [x] New regression test `Issue6067DeferredCloseRebuildTest` (tagged `slow`): builds a 20,000-vector index (128 dims, maxConnections=64, beamWidth=400 - same per-vector cost as `LSMVectorIndexCloseCancelsInFlightBuildTest`), forces one full build, writes one more vector, then asserts `close()` gives up within 1s via `StallAwareStopwatch.assertGaveUpWithin` (measured: ~50ms after the fix vs. ~3.1s before it - confirmed failing against the pre-fix code). Reopens the database afterwards and confirms every vector, including the one written right before `close()`, is still reachable by search - no data loss from deferring the rebuild.
- [x] Full `com.arcadedb.index.vector.*Test` suite (409 tests) passes, including the pre-existing close-path regression tests (`Issue6518CloseDoesNotLeakBuildPoolTest`, `LSMVectorIndexCloseCancelsInFlightBuildTest`).
- [x] `mvn -pl engine -am compile` clean.

https://claude.ai/code/session_0124ucXhoDv5kRUb3WpZjjXd

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3880,3 +3880,31 @@ read noticed first. An unreadable *chunk* of a vertex that does exist keeps the 
 `ConcurrentModificationException` and #5764's scoped advice, on both paths.
 
 [#6586](https://github.com/ArcadeData/arcadedb/issues/6586)
+
+## Closing a database no longer blocks on a full vector graph rebuild for large indexes (#6067)
+
+`LSMVectorIndex.close()` persists the graph before shutting down, so that a fast restart finds a ready-to-search
+graph rather than rebuilding on first open. For an index with any pending write (`graphState == MUTABLE`, or a
+first-ever build that never persisted), that persist has always meant a full `buildGraphFromScratch()` - there is
+no incremental update, so any rebuild reindexes every vector in the index, not just the ones written since the
+last build. That cost was paid synchronously on the closing thread regardless of index size: a single write to a
+200,000-vector index measured `close()` blocking for roughly 43 seconds, the same cost as writing a thousand
+vectors, because the rebuild's cost tracks total index size rather than what was actually written.
+
+`close()` now defers that rebuild instead of running it, for any index at or above the same size threshold the
+search path already uses to decide between a synchronous and an asynchronous rebuild (1,000 vectors,
+`arcadedb.vectorIndex.mutationsBeforeRebuild`'s sibling knob is unaffected). The vectors themselves are unaffected
+by any of this - they go through ordinary page/WAL persistence independent of the graph file, so nothing is lost
+by skipping the rebuild - only the graph's on-disk topology is left stale or absent. The next time that index is
+actually searched, after any reopen, the existing stale-persisted-graph detection (the same manifest/node-count
+check that already handles a persisted graph left behind by an unrelated shutdown) notices and rebuilds it lazily
+before answering the query.
+
+**Upgrading:** a session that writes to a large vector index and closes without ever searching it again no longer
+pays the rebuild cost at all - not at close, and not later, since nothing forces it until a search actually needs
+the graph. A session that does go on to search that index pays the same total rebuild cost as before, just moved
+from `close()` to that first query instead - so a query immediately after reopening a large, recently-written
+index may be noticeably slower than the queries that follow it, where `close()` used to absorb that latency
+instead. Below the 1,000-vector threshold, `close()` keeps rebuilding synchronously as before, unaffected.
+
+[#6067](https://github.com/ArcadeData/arcadedb/issues/6067)
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -3148,7 +3148,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
    * Above this threshold, a full rebuild can take seconds to minutes, so async is preferred.
    * <p>
    * Also the threshold {@link #flush()} uses to decide whether a close-time rebuild is cheap enough to run
-   * synchronously or should be deferred to the next open instead (issue #6067).
+   * synchronously or should be deferred to the next SEARCH on this index instead (issue #6067) - opening the
+   * database alone never triggers it, only {@link #ensureGraphAvailable()}/{@link #rebuildGraphBeforeSearch()}
+   * do, and that deferred rebuild is itself still synchronous on whichever search first reaches it, not async.
    */
   private static final int ASYNC_REBUILD_MIN_GRAPH_SIZE = 1000;
   private static final int[] EMPTY_ORDINALS             = new int[0];
@@ -6432,7 +6434,10 @@ public class LSMVectorIndex implements Index, IndexInternal {
         // with no persisted graph - and defer the rebuild to whenever this index is next actually searched,
         // reusing the exact lazy-load/staleness-detection path (ensureGraphAvailable()) that already handles a
         // stale persisted graph on every ordinary reopen. A session that closes and reopens without searching
-        // this index never pays the cost at all; one that does pays it at the query instead of at close().
+        // this index never pays the cost at all; one that does pays it at the query instead of at close() - that
+        // follow-up rebuild is itself still synchronous on the search that triggers it (ensureGraphAvailable()'s
+        // stale-graph fallback is not gated by this same threshold), so this trades an unconditional cost on
+        // EVERY close() for a conditional one paid by at most one search, not for a non-blocking one.
         LogManager.instance().log(this, Level.FINE,
             "Deferring graph build on close for index %s: %d vectors is at or above the async rebuild threshold "
                 + "(%d), so the rebuild is deferred to the next search on this index instead of blocking close()",
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -3146,6 +3146,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
   /**
    * Minimum graph size to use async rebuild. Below this, synchronous rebuild is fast enough.
    * Above this threshold, a full rebuild can take seconds to minutes, so async is preferred.
+   * <p>
+   * Also the threshold {@link #flush()} uses to decide whether a close-time rebuild is cheap enough to run
+   * synchronously or should be deferred to the next open instead (issue #6067).
    */
   private static final int ASYNC_REBUILD_MIN_GRAPH_SIZE = 1000;
   private static final int[] EMPTY_ORDINALS             = new int[0];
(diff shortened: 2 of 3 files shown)


lvca:
Addressed both review points in 99c930666:
- Fixed the `ASYNC_REBUILD_MIN_GRAPH_SIZE` javadoc: it deferred to "the next open", but opening alone never triggers the rebuild - only a subsequent search does.
- Clarified in the `flush()` comment that the deferred rebuild is itself still synchronous on whichever search reaches it first (not async) - this trades an unconditional cost on every `close()` for a conditional one paid by at most one search, not for a non-blocking one.

Filed #6655 to track making that first-search rebuild size-aware/async as a follow-up - it's a separate design decision (touches `ensureGraphAvailable()`'s locking/staleness contract) rather than something to fold into this fix.

Full `com.arcadedb.index.vector.*Test` suite (409 tests) still green after the change.

lvca:
Addressed both points from this round in da0fa345e:

- **Added the second test case** for the `graphState == LOADING && !graphAlreadyOnDisk` arm (never-built graph, as opposed to a rebuilt one). Note it's not a symmetric timing pin like the first test: the pre-fix code always builds+persists whenever there's pending work, so a large index closed twice with nothing ever persisted is a state only reachable *with* the fix - pre-fix, the first close already does the build and the second finds nothing left to do, which would make a timing assertion pass trivially for the wrong reason. Instead it pins `graphRebuildCount` directly: zero across both closes, exactly one once the deferred data is finally searched. Confirmed failing against the pre-fix code (0 rebuilds ever happen, because the old code built eagerly in session 1 rather than lazily on search) and passing with the fix.
- **Added a `docs/release-26.9.1.md` entry** describing the behavior change for users upgrading (also proposed the equivalent `arcadedb-docs` note in a `[NOTE]` block on the vector-search concepts page, left as an uncommitted working-tree edit there for review, per standing practice on that repo).

Full `com.arcadedb.index.vector.*Test` suite (410 tests) still green.

lvca:
Review says nothing blocking. Judging the two optional points:

- **Release notes prominence**: the existing "Upgrading:" paragraph in `docs/release-26.9.1.md` already spells out exactly the scenario raised (a query right after reopening a large, recently-written index can be slower than the ones that follow, because `close()` used to absorb that latency and now doesn't) - I think it's adequately covered as written, no further change.
- **Exposing "rebuild pending" via `getStats()`**: legitimate observability idea, but it's a new feature surface (what shape - boolean vs timestamp, when it clears, etc.) rather than something this correctness fix needs, and the reviewer explicitly flagged it as optional/not required. Filed #6657 to track it separately rather than scope-creep it into this PR.

Waiting on the remaining CI checks (Java CI - test, CodeQL analyzers) to finish before this is mergeable.

lvca:
\`unit-tests\` failed on this PR, but not with anything related to this change:

- \`TypeConversionTest.noConversion\`: \`expected: Millis but was: Seconds\`
- \`ParallelRangeTopKTest.aRangeForcedOntoTheCallerThreadLeavesTheCallersTransactionAlone\`: \`Index Sparse-vector top-K range fan-out timed out after 30s\`

Neither file is touched by this PR's diff (only \`LSMVectorIndex.java\`, the new \`Issue6067DeferredCloseRebuildTest\`, and \`docs/release-26.9.1.md\` are). Both pass cleanly in isolation locally (\`mvn -pl engine -am test -Dtest=TypeConversionTest,ParallelRangeTopKTest\` → 27/27 green). \`unit-tests\` passed on the last several completed runs on \`main\`, so this isn't a lane that's chronically red - but a 30s timeout failing specifically in a 13,309-test full-suite run, on a test that's fine standalone, is exactly the JVM-stall-under-load shape this repo's own \`StallAwareStopwatch\` tooling exists for (just not applied to this particular test). \`vector-unit-tests\` - the lane that actually exercises this PR's change - passed cleanly.

Treating this as full-suite flakiness unrelated to this PR rather than something to fix here.

lvca:
All checks green after rerunning the spurious `unit-tests` failure (confirmed unrelated - see earlier comment). `mergeStateStatus` is CLEAN. Ready to merge.

tae898:
Verified this against our harness, before and after, and it does what you say at close. But the cost relocates rather than going away, and I think the async path you are relying on is unreachable in exactly the scenario the deferral creates.

Setup: 20,000 base vectors, 128 dims, then N rounds of (open, insert 100, close) with NO search, then one search. Fresh database per repetition, medians of n=5, milliseconds. `before` is current main, `after` is this PR.

| rounds | before: close | before: first search | after: close | after: first search |
|---:|---:|---:|---:|---:|
| 0 | - | 101.0 | - | 165.9 |
| 1 | **8,054.7** | 63.6 | **17.4** | **7,217.8** |
| 5 | 6,901.4 | 60.4 | 16.4 | 7,255.4 |
| 20 | 7,377.1 | 65.8 | 16.3 | **9,738.4** |

Close improves 463x, 8,054 ms to 17.4 ms, which is the win. A session that writes and never searches again genuinely stops paying, and that is worth having.

The first search after such a close goes the other way: 63.6 ms to 7,217.8 ms. Roughly the whole close-time rebuild, now on the query thread.

## Why the async rebuild does not catch it

`rebuildGraphBeforeSearch()` picks its branch on:

```java
final boolean isSmallGraph = graphIndex == null || graphIndex.size() < ASYNC_REBUILD_MIN_GRAPH_SIZE;
```

After a reopen there is no in-memory graph, so `graphIndex == null` and `isSmallGraph` is true no matter how large the index really is. The synchronous branch runs, and its comment is the assumption the deferral invalidates:

> *Small graph or first-ever build: synchronous rebuild (fast enough to not block noticeably).*

That was fair when the only way to reach it was a genuinely small or never-built index. Deferring the close-time rebuild makes `graphIndex == null` the NORMAL state on the next open of any index that was written to, so the branch meant for small graphs now runs for large ones, and `startAsyncGraphRebuild()` is unreachable on this path.

The threshold reads the in-memory graph's size, which is precisely the thing the deferral removed. A persisted-size check would route these to the async branch instead.

## The accumulation question

I checked whether the deferral lets the un-indexed delta grow across sessions, since nothing rebuilds while no one searches. It grows, but not unboundedly: 7,217 ms after one round against 9,738 ms after twenty, so twenty times the writes cost about 35% more. The dominant term is rebuilding the base index, not the delta.

## Shape

This is the same trade as #6641, where deferring the GAV's CSR restore moved a full rebuild onto the first query that needed it. There the fix is to fall back to the ordinary path while rebuilding behind it. Here the equivalent already exists in `startAsyncGraphRebuild()`, serving the current graph plus the delta scan, and only the `graphIndex == null` guard stands between this path and it.

Environment: Intel Core Ultra X9 388H, cpuset 0-3, 30 GiB, NVMe, eclipse-temurin:25-jdk, -Xms4g -Xmx14g, machine otherwise idle. `before` = our synced main, `after` = this PR at `da0fa345e`.

lvca:
Closing as superseded: this branch's entire contribution (the close-time rebuild deferral in `flush()`, the `ASYNC_REBUILD_MIN_GRAPH_SIZE` javadoc note, the `Issue6067DeferredCloseRebuildTest` regression test, and the release-notes entry) has already been independently re-implemented as a superset on `main` via the #6655/#6657 follow-up work (7ec5797ae, d48c01ef1). Verified locally: merging `main` into this branch and resolving the conflict in favor of `main`'s side produces a merge commit with an empty diff against `main` - nothing left for this PR to add.

Issue #6487, opened 2026-08-20
ha-integration-tests: job hung 3 consecutive times on PR #6476, different freeze point each time


lvca:
### Summary

While reviewing PR #6476 (issue #6474, an `import database`/SSRF fix with zero code overlap with `ha-raft`), the `ha-integration-tests` CI lane hung (not just failed) on **three consecutive attempts** - the job sat at `status: in_progress` for well over an hour each time, with the Actions log genuinely frozen (confirmed via multiple `gh api repos/.../jobs/<id>/logs` fetches minutes apart returning byte-identical content), rather than actively running slowly.

This is distinct from the well-documented rotating-cast flake tracked in #5668/#5702 (a different test fails each run, but the job always *completes*), and distinct from the specific livelock already fixed in #5973/PR #5975 (that one span was an actively-logging `ReconfigurationInProgressException` retry storm with no backoff, root-caused to `KubernetesAutoJoin.tryAutoJoin()`'s missing retry policy).

### What was observed

Run [32281444306](https://github.com/ArcadeData/arcadedb/actions/runs/32281444306), three attempts:

1. **Job 96184041895** (started 18:40:12Z): froze right after `ArcadeDB_0`'s own startup sequence completed (last log line at `19:09:15.897`). Confirmed frozen across 3 fetches spanning ~12 min. Cancelled after ~100 min total elapsed.
2. **Job 96196935658** (started 19:23:00Z): froze at a different point - mid `LeaderNotReadyException` stack trace during Raft leader-readiness churn (last log line at `19:53:52.170`). Confirmed frozen across 2 fetches ~6 min apart. Cancelled after ~40 min.
3. **Job 96209140464** (started 20:05:02Z): froze at a third, different point - mid `LSMTreeIndex` index-build logging for `MigrationType_N` types (last log line at `20:35:55`). Confirmed frozen across 2 fetches ~11 min apart.

Three different freeze points across three attempts rules out a single reproducible frozen-instruction-pointer bug in one specific test. It's either:
- A genuine JVM-level deadlock that can occur at multiple points depending on scheduling (unlikely to hit 3/3 different points by chance if it were one specific lock), or
- The GitHub Actions runner/log-streaming infrastructure itself stalling under this account's CI capacity that day (a "bad runner day," similar in shape to the escalating-failure-count pattern already noted for PR #6018 in the project's internal tracking, but for hangs specifically rather than failure count).

### Why this wasn't blocking

[...]

Issue #6475, opened 2026-08-19
Issue6075SnapshotBackupIT.backupUnderConcurrentWritersRestoresToAPointInTime: flaky boundary race under concurrent writers


lvca:
### Summary
`Issue6075SnapshotBackupIT.backupUnderConcurrentWritersRestoresToAPointInTime` failed once in CI with a boundary off-by-one, then passed cleanly on an immediate re-run with no code change. Filing so the flake is tracked instead of silently re-run away next time it blocks an unrelated PR.

### Where
`integration/src/test/java/com/arcadedb/integration/backup/Issue6075SnapshotBackupIT.java:114`

```java
final long countBefore = database.countType(TYPE, false);
new Backup(database, BACKUP_FILE).setVerboseLevel(0).setMaxMBPerSecond(MAX_MB_PER_SECOND).backupDatabase();
final long countAfter = database.countType(TYPE, false);
...
assertThat(restored.countType(TYPE, false)).isBetween(countBefore, countAfter);
```

### Observed failure (CI run 32220365449, job 95970773131, PR #6449)
```
java.lang.AssertionError:
Expecting actual:
  20001L
to be between:
  [20002L, 29422L]
	at Issue6075SnapshotBackupIT.backupUnderConcurrentWritersRestoresToAPointInTime(Issue6075SnapshotBackupIT.java:114)
```
`countBefore` was `20002` and the restored, point-in-time-snapshotted database had exactly `20001` records - one less than the lower bound, sampled right at the boundary. Re-running only this job (no code change) passed.

### Analysis
This is a concurrent-writer test: a background thread keeps inserting while `countBefore`/`countAfter` are sampled around the backup call, and the restored snapshot is asserted to land in that window. `countType(TYPE, false)` and the backup's own internal point-in-time cutoff are not synchronized with each other by any barrier stronger than "the writer thread happened to commit or not commit in the gap between the two calls" - so a commit landing in exactly the wrong few-microsecond window makes the restored count come out one below `countBefore`. This is a real race in the *test's* sampling, not necessarily in the backup/snapshot feature itself, but it should be looked at by whoever knows the #6075 snapshot-backup internals best to confirm the assertion window is right (or needs a small slack/tolerance) rather than just being timing-lucky most of the time.

### Scope / confidence
[...]

Issue #6473, opened 2026-08-19
Postgres wire: schema-context type fallback doesn't resolve through aliases or multi-table FROM targets


lvca:
Follow-up from #6447 (merged, PR #6452), flagged during that PR's review but explicitly out of scope for it.

`PostgresNetworkExecutor.getDeclaredProperty()` and `resolveQueryTargetType()` back two schema-context fallbacks used when a value alone can't determine a column's PostgreSQL type:

- an empty `LIST` column's element type (issue #5289's original fix), and
- a `DATETIME` column on a database explicitly configured with `arcadedb.dateTimeImplementation=java.util.Date` (#6447), where `java.util.Date` is also `DATE`'s default representation and a bare sampled value can't tell the two apart.

Both fallbacks work by (1) resolving the query's `FROM` target to a single `DocumentType` (`resolveQueryTargetType`, only handles a plain `FROM <identifier>` - no subquery, no `JOIN`) and (2) looking up the property by the *result column's own name* (`getDeclaredProperty`, via `DocumentType.getPolymorphicPropertyIfExists(propertyName)`).

Two query shapes fall through both fallbacks and land on the pre-existing value-only guess instead:

1. **Aliased or computed-expression columns**: `SELECT amount AS x FROM Types6447` - the row's column name is `x`, not `amount`, so the property lookup misses.
2. **Multi-table `JOIN`s**: `SELECT a.col, b.col FROM A JOIN B ON ...` - `resolveQueryTargetType` only resolves a single simple `FROM <identifier>`, so it returns `null` for anything with a `JOIN`.

Neither is a regression: #5289's `getDeclaredListType` already had case 1 before #6447 touched this code at all, and #6447 didn't attempt case 2 either. Both are just more reachable now that #6447 generalized the fallback beyond whole-entity ("`SELECT FROM Type`") projections to any narrow column projection, since a `SELECT`-list is exactly where aliases and multi-table `JOIN`s live.

**Impact**: narrow. Case 1 affects an empty `LIST` column (#5289's original scope) under any configuration, or a `DATETIME` column specifically on a `java.util.Date`-configured database (not the default). Case 2 is the same, restricted to that same `java.util.Date` configuration for `DATETIME`, or an empty-`LIST` column reached via a `JOIN`. In the affected cases the column's OID can again depend on whether the sampled row happened to be empty - the exact bug class #5289/#6411/#6447 already fixed for the simpler shapes.

[...]

Issue #6471, opened 2026-08-19
JSONL export silently skips un-serializable records with no aggregate failure signal (export looks complete)


ruispereira:
### Summary
The JSONL exporter catches a per-record serialization exception, logs it at SEVERE, and continues — so a record that fails to serialize is omitted while the export still completes and prints its counters as if whole. There is no aggregate "N records skipped" signal or non-zero outcome. Symmetric to the import-side swallow; filed separately because it is the export half and a distinct file.

### Where (HEAD `0291182af`)
`integration/src/main/java/com/arcadedb/integration/exporter/format/JsonlExporterFormat.java` — `exportVertices`/`exportEdges`/`exportDocuments`/`exportLightweightEdges` each `catch (Exception e) { LogManager...SEVERE; }` and continue (`:154-157`, `:174-177`, `:213-216`, `:233-236`). Introduced deliberately (commit `3fca7b106`, "skip broken records in export"), so the skip itself is intended — the gap is the absence of an aggregate failure signal.

### Repro
Export a database containing a record that throws on serialize (e.g. combined with the DATE-unit export bug): the record silently vanishes from the archive, the export exits successfully, and its counters read as complete.

### Suggested fix
Surface a skipped-record count and make a non-empty count a failing/non-zero outcome (or gate the skip behind an explicit flag), symmetric to the import-side fix.

### Scope / confidence
High (code read); deliberately introduced, so this is a "fail loud / observable" enhancement rather than a crash. Low severity on its own; matters most combined with a serialization bug that makes records throw.

lvca:
Fixed in PR #6663, now merged into `main` (26.9.1).

If a record can't be exported to JSONL, the export used to finish and report success even though that record was silently missing. Now, if any records were skipped, the export fails with a clear error telling you how many records were skipped — so you know the export file is incomplete instead of finding out later.

---
❤️ Thanks for helping make ArcadeDB better! ArcadeDB is free and open source, sustained by its community. If it's useful to you or your company, please consider [becoming a sponsor](https://github.com/sponsors/ArcadeData) to keep development going.

ruispereira:
Re-reviewed the fix (`40057b6e3`) at HEAD `af4f67fd2` — the vertex/edge/document loops are correct, but **`exportLightweightEdges` still silently drops the remainder of a vertex's edges and undercounts the skip**.

`integration/src/main/java/com/arcadedb/integration/exporter/format/JsonlExporterFormat.java:199-224`:
```java
[...]

PR #6663, opened 2026-08-23
fix(#6471, #6475, #6473): JSONL export skip signal, flaky snapshot-backup boundary, Postgres alias fallback

lvca:
## Summary

Batch of three small, independent fixes flagged by other analysis, plus a fourth (#6487) analyzed and left as-is.

- **#6471** - JSONL export silently skipped un-serializable records with no aggregate failure signal. `JsonlExporterFormat` still skips-and-logs a broken record (deliberate, per the issue), but `Exporter.exportDatabase()` now throws when any records were skipped, and reports the count in its result map, instead of reporting success on an incomplete archive.
- **#6475** - `Issue6075SnapshotBackupIT.backupUnderConcurrentWritersRestoresToAPointInTime` flaked once in CI (restored count one below `countBefore`). Root cause is a narrow, single-commit race in the test's own sampling (`countBefore`/`countAfter` vs. the backup's separately-established point-in-time cutoff, no barrier between them) - not the backup/restore path. Added one record of documented tolerance on the lower bound.
- **#6473** - Postgres wire: `getDeclaredProperty()`'s schema-context fallback (backing #5289's empty-LIST element type and #6447's DATE/DATETIME `java.util.Date` disambiguation) missed whenever the projected column carried an alias (`SELECT amount AS x FROM Type`), since it looked the property up by the row's own (aliased) column name. Now resolves the alias back to its source property via the SELECT projection, memoized per portal. The issue's second case (multi-table JOINs) doesn't apply - verified that ArcadeDB SQL's `FROM` clause has no JOIN support at all, so that shape can't reach this code.
- **#6487** - analyzed, no code change. CI infra hang (Actions log genuinely frozen across repeated fetches, three different freeze points, zero code overlap with the PR it appeared on). The issue's own evidence points at runner/log-streaming infrastructure, not an ArcadeDB defect. Commented on the issue with the analysis; no commit here.

## Test plan

- [x] `integration`: `JsonlExporterIT` (5/5, new regression test spies via a JDK dynamic proxy to force a mid-export failure and confirms the export now fails loudly with the skip count, while still writing the best-effort file)
- [x] `integration`: `Issue6075SnapshotBackupIT` (4/4, both `snapshot=true/false` parameterizations)
- [x] `integration`: `SQLLocalExporterTest`, `Issue6455JsonlDateTimeRoundTripIT`, `LightweightEdgeRoundTripIT` (no regressions from the `Exporter`/`ExporterContext` changes)
- [x] `postgresw`: `Issue6473AliasedColumnTypeResolutionIT` (new, 2/2 - confirmed red without the fix via a file-based patch revert/reapply, green with it)
- [x] `postgresw`: `Issue6447TypeResolutionIT`, `PostgresWJdbcIT`, `Issue6412CatalogIT`, `Issue5290ClientCompatibilityIT`, `Issue6411ByteaIT`, `PostgresProtocolIT`, `PostgresQueryMetricsIT` (152 tests, no regressions from the `getColumns`/`getDeclaredProperty` signature changes)
- [x] Full `postgresw` unit test suite (408 tests)

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/exporter/Exporter.java
+++ b/integration/src/main/java/com/arcadedb/integration/exporter/Exporter.java
@@ -84,12 +84,18 @@ public class Exporter {
         elapsedInSecs = 1;
 
       final long totalRecords = context.vertices.get() + context.edges.get() + context.documents.get();
+      final long skippedRecords = context.skippedRecords.get();
 
-      logger.logLine(0,//
-          "Database exported successfully: %,d records exported in %s secs (%,d records/secs %,d documents %,d vertices %,d edges)",
+      if (skippedRecords > 0)
+        logger.errorLine(
+            "Database export completed with errors: %,d record(s) could not be serialized and were skipped (see the log above for the per-record errors)",
+            skippedRecords);
+      else
+        logger.logLine(0,//
+            "Database exported successfully: %,d records exported in %s secs (%,d records/secs %,d documents %,d vertices %,d edges)",
 //
-          totalRecords, elapsedInSecs, totalRecords / elapsedInSecs, context.documents.get(), context.vertices.get(),
-          context.edges.get());
+            totalRecords, elapsedInSecs, totalRecords / elapsedInSecs, context.documents.get(), context.vertices.get(),
+            context.edges.get());
 
       // RETURN STATISTICS
       final Map<String, Object> result = new LinkedHashMap<>();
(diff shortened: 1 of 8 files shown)


lvca:
Thanks for the review. Checked the one substantive note:

> The graphml/graphson formats... have their own per-record catch blocks that don't increment it

Verified against `gremlin/src/main/java/com/arcadedb/gremlin/integration/exporter/format/GraphMLExporterFormat.java` and `GraphSONExporterFormat.java` - neither has any per-record `catch` block at all (`grep -n catch` on both files: zero matches). A record that throws while serializing in either of those formats propagates and aborts the whole export immediately, rather than being skipped-and-continued. So there's no silent-partial-success gap there to begin with - that failure mode is specific to `JsonlExporterFormat`'s deliberate skip-and-continue (commit `3fca7b106`), which is what #6471 is about. No follow-up needed.

Everything else in the review is confirmed correct and non-blocking, no further changes.

Issue #6567, opened 2026-08-22
value-preserving `CASE` projection changes a `collect()` result after `CALL ... IN TRANSACTIONS`


YGY-001:
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1`
- OpenCypher over the Neo4j Bolt protocol
- Clean database named `Test`

## Reproduction setup

Run these two setup statements:

```cypher
CREATE (n1:Person {id: 'n1'})-[r1:rt2]->(x:Person {id: 'x', klist: ['a']})-[:rt10]->(n0:Person {id: 'n0'})-[r0:rt8 {k9: 267022112, k8: [1]}]->(t:Person {id: 't', k0: true, k8: [1, 2]});
CREATE (s:Person {id: 22})-[:rt5]->(n3:Person {id: 'n3', klist: []})-[:rt8]->(n2:Person {id: 'n2', k3: 'e'});
```

## Control query

```cypher
FILTER (0 IN range(12, 1, -1))
MATCH REPEATABLE ELEMENT BINDINGS
  (n1)-[r1:rt2]->(x {klist: ['a']})-[:rt10]->(n0)-[r0:rt8|rt1|rt7 {k9: 267022112}]->(t {k0: true, k8: [1, 2]}),
  (n1), p0 = (n0)
WHERE any(item IN r0.k8 WHERE item IS NOT NULL)
CALL () {
  MATCH (n2), (n3 {klist: []})-[:rt8]->(n2),
        p1 = ({id: 22})-[r2:rt5|rt2|rt7]->(n3)
  WHERE true
  RETURN n3 AS n14, n2 AS n15, r2 AS r6, p1 AS p13,
         toLower(toStringOrNull(r2.k3)) AS alias127,
         n2.k7 AS alias128, n2.k10 AS alias129
} IN TRANSACTIONS OF 10 ROWS
RETURN collect(toStringOrNull(n15.k3)) AS alias131;
```

Observed result:

```text
alias131
[]
```

## Equivalent query with identity projections

Run the same query, but insert these value-preserving projections immediately
after the `CALL ... IN TRANSACTIONS` clause and before the final `RETURN`:

```cypher
WITH CASE WHEN n1 IS NULL THEN null ELSE n1 END AS n1,
     CASE WHEN x IS NULL THEN null ELSE x END AS x,
     CASE WHEN r1 IS NULL THEN null ELSE r1 END AS r1,
[...]

PR #6689, opened 2026-08-24
fix(#6573): isEdgeVariableReferenced has no CREATE/MERGE case

### LINKED ISSUE #6573: isEdgeVariableReferenced has no CREATE/MERGE case: an edge read only inside a CREATE/MERGE inline property silently returns null
## Summary

Found during review of #6567/PR #6570. `CypherVariableUsage.isEdgeVariableReferenced` decides whether a MATCH-bound relationship variable keeps its real binding (vs. being anonymized for the CSR/GAV fast path) by walking every clause type in `statement.getClausesInOrder()`. The switch handles `RETURN`, `WITH`, `UNWIND`, `SET`, `REMOVE`, `DELETE`, `FOREACH`, `SUBQUERY` - but has no `case CREATE` / `case MERGE`, so it falls to `default` and the clause is never inspected.

Contrast with `foreachReferencesVariable`'s own inner switch (same file), which explicitly treats a nested `CREATE`/`MERGE` conservatively:
```java
case CREATE:
case MERGE:
  // Inline property expressions inside CREATE/MERGE patterns may reference the variable.
  // Enumerating them cheaply is awkward, so stay conservative: keep the edge binding.
  return true;
```
The top-level switch never got the same treatment.

## Reproduction

```cypher
CREATE (a:Person {id:'a'})-[r:KNOWS {since: 2001}]->(b:Person {id:'b'})
```
```cypher
MATCH (a:Person)-[r:KNOWS]->(b:Person) CREATE (c:Note {since: r.since}) RETURN c.since AS s
```

Expected: `s = 2001`. Actual: `s = null` - the relationship step binds `r` under an internal anonymous name (since nothing else in the query appears, to this check, to reference it), so the CREATE clause's `r.since` reads a missing binding and silently evaluates to `null` instead of the real value.

## Suggested fix

Add a `case CREATE: case MERGE:` arm to the top-level switch in `isEdgeVariableReferenced`, mirroring `foreachReferencesVariable`'s conservative `return true` (a false positive there only forgoes the CSR/GAV fast path; a false negative silently drops a still-referenced edge, which is the failure class this whole method exists to prevent - same reasoning that motivated #6567's fix). A regression test parallel to the existing `CypherVariableUsageTest` cases (`setReadsItAsTargetAndAsValue`, `removeReadsIt`, etc.) should cover both `CREATE {prop: r.x}` and `MERGE (n {prop: r.x})`.

## Where

`engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherVariableUsage.java`, the `isEdgeVariableReferenced` switch (around the `RETURN`/`WITH`/... cases).

robfrank:
Closes #6573

## Summary

`CypherVariableUsage.isEdgeVariableReferenced` decides whether a MATCH-bound relationship variable
keeps its real binding (vs. being anonymized for the CSR/GAV fast path) by walking every clause type
in `statement.getClausesInOrder()`. The top-level switch handled `RETURN`, `WITH`, `UNWIND`, `SET`,
`REMOVE`, `DELETE`, `FOREACH`, `SUBQUERY` - but had no `case CREATE` / `case MERGE`, so a top-level
CREATE/MERGE clause fell to `default` and was never inspected.

Contrast with `foreachReferencesVariable`'s own inner switch (same file), which already treats a
nested `CREATE`/`MERGE` conservatively:
```java
case CREATE:
case MERGE:
  // Inline property expressions inside CREATE/MERGE patterns may reference the variable.
  // Enumerating them cheaply is awkward, so stay conservative: keep the edge binding.
  return true;
```
The top-level switch never got the same treatment.

### Reproduction

```cypher
CREATE (a:Person {id:'a'})-[r:KNOWS {since: 2001}]->(b:Person {id:'b'})
```
```cypher
MATCH (a:Person)-[r:KNOWS]->(b:Person) CREATE (c:Note {since: r.since}) RETURN c.since AS s
```

Expected: `s = 2001`. Actual (before fix): `s = null` - the relationship step bound `r` under an
internal anonymous name (nothing else in the query appeared, to this check, to reference it), so the
CREATE clause's `r.since` read a missing binding and silently evaluated to `null`.

## Fix

Added a `case CREATE: case MERGE:` arm to the top-level switch in `isEdgeVariableReferenced`,
mirroring `foreachReferencesVariable`'s conservative `return true`. A false positive there only
forgoes the CSR/GAV fast path; a false negative silently drops a still-referenced edge, which is the
failure class this whole method exists to prevent (same reasoning that motivated #6567's fix).

`engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherVariableUsage.java`

## Test plan

- [x] `CypherVariableUsageTest.createReadsItInAnInlineProperty` / `.mergeReadsItInAnInlineProperty` -
      unit-level checks parallel to the existing cases in that file (`setReadsItAsTargetAndAsValue`,
      `removeReadsIt`, etc.). Verified to fail (assert `true`, got `false`) without the fix.
- [x] `Issue6573CreateMergeEdgeVariableTest` - new end-to-end regression test running the issue's own
      reproducer through the real Cypher engine. Verified to fail with a `null` result (matching the
      issue's reported symptom exactly) without the fix, via a temporary stash of the production
      change.
- [x] Full `com.arcadedb.query.opencypher.**` package (excluding benchmark/slow/vector lanes): 8906
      tests, 0 failures, 0 errors, 98 skipped (pre-existing, unrelated).

### DIFF
--- a/docs/6573-isedgevariablereferenced-create-merge.md
+++ b/docs/6573-isedgevariablereferenced-create-merge.md
@@ -1,0 +1,54 @@
+# Issue #6573: isEdgeVariableReferenced has no CREATE/MERGE case
+
+## Summary
+
+`CypherVariableUsage.isEdgeVariableReferenced` decides whether a MATCH-bound relationship variable
+keeps its real binding (vs. being anonymized for the CSR/GAV fast path) by walking every clause type
+in `statement.getClausesInOrder()`. The top-level switch handled `RETURN`, `WITH`, `UNWIND`, `SET`,
+`REMOVE`, `DELETE`, `FOREACH`, `SUBQUERY` - but had no `case CREATE` / `case MERGE`, so it fell to
+`default` and the clause was never inspected.
+
+Contrast with `foreachReferencesVariable`'s own inner switch (same file), which already treats a
+nested `CREATE`/`MERGE` conservatively (`return true`). The top-level switch never got the same
+treatment.
+
+## Reproduction (from the issue)
+
+```cypher
+CREATE (a:Person {id:'a'})-[r:KNOWS {since: 2001}]->(b:Person {id:'b'})
+```
+```cypher
+MATCH (a:Person)-[r:KNOWS]->(b:Person) CREATE (c:Note {since: r.since}) RETURN c.since AS s
+```
+
+Expected: `s = 2001`. Actual (before fix): `s = null` - the relationship step bound `r` under an
+internal anonymous name (since nothing else in the query appeared, to this check, to reference it),
+so the CREATE clause's `r.since` read a missing binding and silently evaluated to `null`.
+
+## Fix
+
+Added a `case CREATE: case MERGE:` arm to the top-level switch in `isEdgeVariableReferenced`,
+mirroring `foreachReferencesVariable`'s conservative `return true`. A false positive there only
+forgoes the CSR/GAV fast path; a false negative silently drops a still-referenced edge, which is the
+failure class this whole method exists to prevent.
+
+File: `engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherVariableUsage.java`
+
+## Tests
+
+- `CypherVariableUsageTest.createReadsItInAnInlineProperty` / `.mergeReadsItInAnInlineProperty` -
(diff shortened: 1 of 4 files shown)


robfrank:
@claude review

Issue #6679, opened 2026-08-24
Postgres simple-query path materializes the entire result set in memory before sending (OOM on a large SELECT)


ruispereira:
### Summary
The Postgres simple-query ('Q') path materializes the entire result set into a `List<Result>` in memory before writing the first DataRow, so a large `SELECT` over the simple protocol can OOM the server. The Bolt executor streams; the Postgres extended protocol is bounded by the portal max-rows; only the simple-query path is unbounded.

### Where (HEAD `86a704526`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java:670` → `final List<Result> cachedResultSet = browseAndCacheResultSet(resultSet, 0);` (limit 0 = unlimited) then `writeRowDescription(...)` / `writeDataRows(cachedResultSet, ...)`. All row objects are held simultaneously before the first byte is sent.

### Repro
A plain `SELECT` over the simple query protocol (e.g. `psql -c "SELECT FROM HugeType"`) against a large table → the whole result is buffered into an `ArrayList` server-side before streaming, risking OutOfMemoryError. The Bolt path (`handlePull`) streams row-by-row and does not exhibit this (verified).

### Suggested fix
For the 'Q' path, iterate the `ResultSet` and write DataRows incrementally instead of pre-caching with limit 0; or cap with a configurable fetch bound.

### Scope / confidence
High that everything is materialized (read at HEAD). Honest caveat: this is a long-standing, documented design (the browse-cache predates the #4197 audit and is deliberate for portal semantics), so it may be considered known/intended — reporting because it is a real "whole result materialized → OOM on a big simple-query SELECT" gap and the sibling transports avoid it.

lvca:
Fixed in #6693 (merged to main).

A very large `SELECT` over the Postgres wire protocol (simple query mode) could use unbounded memory. There's now a configurable limit (`arcadedb.postgres.simpleQueryMaxRows`, default 1,000,000 rows) - a result larger than that is refused with a clear error instead of risking a server crash. For very large result sets, use the extended query protocol with a bounded fetch size instead.

---
❤️ Thanks for helping make ArcadeDB better! ArcadeDB is free and open source, sustained by its community. If it's useful to you or your company, please consider [becoming a sponsor](https://github.com/sponsors/ArcadeData) to keep development going.

ruispereira:
[...]

Issue #6676, opened 2026-08-24
SQL DISTINCT and Cypher GROUP BY key on raw boxed values (no numeric canonicalization): duplicate groups/rows for Integer vs Long vs Double


ruispereira:
### Summary
Two grouping/dedup paths key on raw Java-boxed values instead of canonicalizing numeric types, so a value present as different numeric runtime types (`Integer 1` vs `Long 1` vs `Double 1.0`) splits into separate groups/rows — impossible under the single logical numeric type. SQL `DISTINCT` and Cypher `GROUP BY` both have this, while their siblings (SQL GROUP BY via #4516, Cypher DISTINCT via `DistinctNumericKey`/#5789) canonicalize correctly.

### Where (HEAD `86a704526`)
- SQL: `engine/src/main/java/com/arcadedb/query/sql/executor/DistinctExecutionStep.java`, `DistinctKey` (L41-69) stores raw `result.getProperty(...)` in a `HashMap` and compares with `HashMap.equals`. Contrast `AggregateProjectionCalculationStep.GroupByKey` (L49-50) which calls `Type.normalizeNumberForKey(...)` (#4516).
- Cypher: `engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java` uses the grouping value directly as the map key (single-key `groups.get(keyValue)` L167; multi-key `GroupKeyValues` L232/L315) with no canonicalization. Contrast the Cypher DISTINCT paths (`ProjectReturnStep`/`WithStep`) which use `DistinctNumericKey`.

### Repro
- SQL: `SELECT DISTINCT v FROM t` where `v` is `Integer 1` on one row, `Long 1` on another (schemaless/mixed) → expected 1 row, actual 2.
- Cypher: `UNWIND [1, 1.0, 1] AS x RETURN x, count(*)` → two groups (`1`→2, `1.0`→1); and two integer occurrences boxed as Integer vs Long would also split.
(SQL `DistinctKey` also treats equal-content arrays as distinct via `Object.equals` identity; `List` values are fine.)

### Suggested fix
Key both paths through the same numeric canonicalization their siblings use (`Type.normalizeNumberForKey` for SQL DISTINCT; `DistinctNumericKey`/`normalizeNumberForKey` for Cypher GROUP BY).

### Scope / confidence
[...]

Issue #6671, opened 2026-08-24
Execution-plan caches can store a plan built before a concurrent DDL: SQL check/put is non-atomic, Cypher has no invalidation guard


ruispereira:
### Summary
Both the SQL and the Cypher execution-plan caches can store a plan that was built against a schema older than a concurrent DDL, so a later execution reuses a plan referencing a dropped/renamed bucket or index (or missing a new index). The SQL path has a check-then-put atomicity gap; the Cypher path has no invalidation guard at all.

### Where (HEAD `86a704526`)
**SQL** — `engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:300-302`:
```java
if (useCache && ... && db.getExecutionPlanCache().getLastInvalidation() < planningStart)   // locks the cache instance
    db.getExecutionPlanCache().put(statement.getOriginalStatement(), selectExecutionPlan);  // locks the map — SEPARATE critical section
```
`getLastInvalidation()` and `put()` are two distinct lock scopes with a gap. Interleaving: thread A passes the check (planningStart T0, lastInvalidation −1) → thread B commits DDL → `ExecutionPlanCache.invalidate()` clears the map and sets lastInvalidation=T3 → thread A's `put` stores the pre-DDL plan into the just-cleared map.

**Cypher** — `engine/src/main/java/com/arcadedb/query/opencypher/query/OpenCypherQueryEngine.java:357-359`:
```java
if (plan.getPhysicalPlan() != null)
  database.getCypherPlanCache().put(queryString, plan.getPhysicalPlan());   // NO planningStart / lastInvalidation check
```
`CypherPlanCache.getLastInvalidation()` (CypherPlanCache.java:110) has **zero callers** (dead code), so the Cypher path is strictly weaker than SQL's — a `DROP INDEX` concurrent with Cypher planning leaves a `NodeIndexSeek` plan cached against the dropped index.

### Repro
Concurrently: thread A runs a `SELECT`/Cypher query that plans against index IDX; thread B runs `DROP INDEX IDX` (or `ALTER TYPE`/create-index). Thread A's plan is cached post-invalidation → future executions return an error or wrong/empty results until the next invalidation.

### Suggested fix
Perform the `lastInvalidation < planningStart` re-check inside the same lock as the store — pass `planningStart` into `put(...)` and insert only when still valid, under one synchronized section — for both caches (and give `CypherPlanCache` the guard it currently lacks).

### Scope / confidence
[...]

Issue #6665, opened 2026-08-23
ArcadeDB 26.8.1: `collect()`/`UNWIND` changes `count(*)` inside a parameterized `CALL`


YGY-001:
## Environment

- ArcadeDB Docker image: `arcadedata/arcadedb:26.8.1`
- Neo4j Bolt protocol
- Each query below was run against a fresh database containing exactly the
  setup shown below.

## Setup

```cypher
CREATE (oa:V {klist: [1167157462, -165975848, -164122097]}),
       (ob:V {k10: 254370155}),
       (oc:V),
       (od:V {k2: "CILs3s"}),
       (oe:V {k10: -866288821}),
       (n0:V {k7: 1994572697, k2: "D5ZqpQB"}),
       (n1:V),
       (x:V {k10: -1192051958}),
       (y:V {k9: "5sm", id: 85}),
       (ob)-[:rt6 {k8: "G", k3: true}]->(oa),
       (ob)-[:rt8 {k8: "nPemIkbA", k10: -1906562515}]->(oc),
       (od)-[:rt9 {k7: -51638479}]->(oe),
       (n1)-[:R]->(x)-[:rt8]->(y)-[:rt2 {k11: [1]}]->(n0);
```

## Query A

```cypher
MATCH DIFFERENT RELATIONSHIPS
  ({klist: [1167157462, -165975848, -164122097]})
    <-[r0:rt6 {k8: "G"}]-({k10: 254370155})
    -[r1:rt8 {k8: "nPemIkbA", k10: -1906562515}]->(),
  p0 = ({k2: "CILs3s"})-[:rt9 {k7: -51638479}]->({k10: -866288821})
WHERE r0.k3 = true
CALL (r0, r1, p0) {
  MATCH (n0 {k7: 1994572697, k2: "D5ZqpQB"}),
        (n1)-[r3]->({k10: -1192051958})-[:rt8]->
          (:V {k9: "5sm", id: 85})-[r2:rt2|rt10|rt9]->(n0)
  WHERE single(item IN r2.k11 WHERE item IS NOT NULL)
  RETURN count(*) AS __expr_count
}
RETURN __expr_count > 0 AS __v;
```

## Query B

Query B is identical except that it materializes the matched bindings through
`collect()` and `UNWIND` before the same `count(*)`:

```cypher
MATCH DIFFERENT RELATIONSHIPS
  ({klist: [1167157462, -165975848, -164122097]})
[...]

PR #6693, opened 2026-08-24
fix(#6680,#6679,#6676,#6671): empty-aggregate row, Postgres simple-query OOM, numeric equality, plan-cache race

### LINKED ISSUE #6680: Non-count aggregate (sum/avg/min/max) over an empty input returns zero rows instead of one null/identity row (count(*) is special-cased)
### Summary
A non-`count` aggregate (`sum`/`avg`/`min`/`max`) over an empty input returns **zero rows** instead of the single row (with a null/identity aggregate) that SQL and Cypher callers expect. `count(*)` is special-cased to return `0`, but the other aggregates are not, so the two behave inconsistently over an empty set.

### Where (HEAD `86a704526`)
`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:780-783`: `GuaranteeEmptyCountStep` is chained **only** when a count-star projection is present **and** `info.groupBy == null`:
```java
final ProjectionItem countStarItem = findCountStarItem(info, context);
if (countStarItem != null && info.groupBy == null)
  result.chain(new GuaranteeEmptyCountStep(...));
```
`AggregateProjectionCalculationStep.executeAggregation` emits nothing when the upstream is empty, so `sum`/`avg`/`min`/`max` produce no row.

### Repro
`SELECT max(x) FROM Foo WHERE 1=0` (or an empty type) → expected one row `{max: null}`; actual zero rows. `SELECT count(*) FROM Foo WHERE 1=0` correctly returns one row `{count: 0}`, showing the inconsistency.

### Suggested fix
Extend the guarantee step to emit a single all-null aggregate row for any no-group aggregation over an empty input (not just count-star).

### Scope / confidence
Behavior confirmed at HEAD. Low severity — long-standing OrientDB-inherited behavior and the count-star special case shows it is a deliberate-if-incomplete design; filed as a semantic gap/consistency fix rather than a fresh defect.

lvca:
## Summary

Batch of fixes for five issues that came out of prior analysis:

- **#6680** - A non-`count` aggregate (`sum`/`avg`/`min`/`max`) over an empty input returned **zero rows** instead of the single row with a null/identity value SQL callers expect (`count(*)` was already special-cased for this). `GuaranteeEmptyCountStep` is generalized to cover any GROUP-BY-less aggregation, not just `count(*)`.
- **#6679** - The Postgres **simple-query** (`'Q'`) protocol path materialized the entire result set into an unbounded `List<Result>` before sending the first row, risking `OutOfMemoryError` on a large `SELECT`. Unlike the extended protocol's portal `Execute`, the simple-query protocol has no client-driven max-rows mechanism, so silently truncating would misreport a partial result as the whole answer. Added `arcadedb.postgres.simpleQueryMaxRows` (default 1,000,000): a result over the limit is refused with a clear error instead.
- **#6676** - SQL `DISTINCT` and Cypher's implicit `GROUP BY` keyed on the raw boxed value instead of canonicalizing numeric types, so `Integer 1` / `Long 1` / `Double 1.0` split into separate rows/groups despite `1 = 1.0`. Both now canonicalize through the same helpers their sibling paths already use (`Type.normalizeNumberForKey` for SQL, `DistinctNumericKey` for Cypher).
- **#6671** - The SQL execution-plan cache's `getLastInvalidation() < planningStart` check and the subsequent `put()` ran in two separate lock scopes, leaving a window where a concurrent DDL (e.g. `DROP INDEX`) could invalidate the cache between them and still have the stale plan land in the just-cleared map. The Cypher plan cache had no invalidation guard at all. Both caches now check-and-insert atomically under a single lock.
- **#6665** - Investigated: `CALL (...) { ... RETURN count(*) }` reportedly disagreed depending on whether the subquery's rows were materialized through `collect()`/`UNWIND` first. The exact repro from the issue no longer reproduces on this branch - both forms agree with each other and with a ground-truth count. No behavior change; added a regression test pinning down the now-correct equivalence so a future regression is caught.

## Test plan
- [x] New/updated unit tests for each fix (see file list below), all passing
- [x] `Issue6679SimpleQueryMaxRowsIT` - real Postgres wire-protocol integration test (raw socket), both over-cap (ErrorResponse) and within-cap (full result) cases
- [x] Broader regression sweep: SQL executor + Cypher (opencypher) + SQL parser packages (~6945 tests) - no failures attributable to these changes (2 unrelated pre-existing flaky tests confirmed to pass in isolation)
- [x] Postgres wire-protocol regression suite (`PostgresProtocolIT`, `PostgresWJdbcIT`, `Issue6412CatalogIT`, `Issue6457AbortedTransactionSimpleQueryIT`) - 133 tests, all green
- [x] `mvn -o -pl engine,postgresw -am compile` clean

## Files changed
- `engine/src/main/java/com/arcadedb/GlobalConfiguration.java`
- `engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java`
- `engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupKeyValues.java`
- `engine/src/main/java/com/arcadedb/query/opencypher/query/CypherPlanCache.java`
- `engine/src/main/java/com/arcadedb/query/opencypher/query/OpenCypherQueryEngine.java`
- `engine/src/main/java/com/arcadedb/query/sql/executor/DistinctExecutionStep.java`
- `engine/src/main/java/com/arcadedb/query/sql/executor/GuaranteeEmptyCountStep.java`
- `engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java`
- `engine/src/main/java/com/arcadedb/query/sql/parser/ExecutionPlanCache.java`
- `postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java`
- Tests: `SQLExecutorAdditionalCoverageTest`, `DistinctExecutionStepTest`, `CypherPlanCacheInvalidationTest`, `Issue6665CallSubqueryCountEquivalenceTest` (new), `Issue6676NumericGroupByEqualityTest` (new), `Issue6671ExecutionPlanCacheRaceTest` (new), `Issue6679SimpleQueryMaxRowsIT` (new)

Closes #6680, #6679, #6676, #6671. Relates to #6665 (no fix needed - see summary).

## Summary by CodeRabbit

- **New Features**
  - Added a configurable row limit for PostgreSQL simple queries, with clear errors for oversized results.
  - Large result sets can be retrieved through extended-query portals.

- **Bug Fixes**
  - Improved numeric consistency for Cypher grouping and SQL `DISTINCT`.
  - Corrected aggregate results for empty datasets.
  - Prevented stale execution plans after schema changes.
  - Improved subquery count handling and connection recovery after rejected queries.

- **Tests**
  - Added regression and integration coverage for query behavior, plan caching, and PostgreSQL row limits.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2009,6 +2009,15 @@ public enum GlobalConfiguration {
       "Maximum size in bytes accepted for a single bind-message parameter value on the Postgres wire protocol. Values declaring a larger size are rejected before allocation. Default is 16MB",
       Integer.class, 16 * 1024 * 1024),
 
+  POSTGRES_SIMPLE_QUERY_MAX_ROWS("arcadedb.postgres.simpleQueryMaxRows", SCOPE.SERVER, """
+      Maximum number of rows a simple-query protocol ('Q' message) SELECT is allowed to buffer server-side before \
+      the first row is sent. Unlike the extended query protocol, the simple-query protocol has no client-driven \
+      cursor/max-rows mechanism and always expects the complete result set in one response, so the server has to \
+      hold it in memory to determine the row description (column set and types) before streaming it. A SELECT whose \
+      result exceeds this limit is refused with an error instead of risking an OutOfMemoryError; the client should \
+      use the extended query protocol with a bounded portal fetch size for very large result sets. Default is \
+      1000000""", Integer.class, 1_000_000),
+
   // BOLT (Neo4j)
   BOLT_PORT("arcadedb.bolt.port", SCOPE.SERVER,
       "TCP/IP port number used for incoming connections for BOLT plugin. Default is 7687", Integer.class, 7687),
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/GroupByAggregationStep.java
@@ -19,6 +19,7 @@
 package com.arcadedb.query.opencypher.executor.steps;
 
 import com.arcadedb.exception.TimeoutException;
+import com.arcadedb.function.DistinctNumericKey;
 import com.arcadedb.function.StatelessFunction;
 import com.arcadedb.query.opencypher.ast.*;
 import com.arcadedb.query.opencypher.executor.CypherFunctionFactory;
@@ -145,10 +146,14 @@ public class GroupByAggregationStep extends AbstractExecutionStep {
       final FunctionCallExpression[] aggExpressions, final String[] aggOutputNames, final int aggCount,
       final CommandContext context, final int nRecords) {
 
-    // Map: raw grouping value -> array of aggregation functions (indexed, not HashMap)
+    // Map: canonicalized grouping value -> group state (representative raw key + aggregation functions).
+    // Keying on the canonicalized value (DistinctNumericKey#canonicalize) rather than the raw boxed value means the
+    // same logical number represented with different Java numeric types (Integer 1 vs Long 1 vs Double 1.0) groups
+    // together instead of splitting, matching Cypher's own `=` operator and the sibling DISTINCT paths (#5789,
+    // issue #6676). The group's raw, first-seen key value is kept in SingleKeyGroupState for the output row.
     // LinkedHashMap preserves insertion order, needed because ORDER BY in WITH clauses
     // may fail to resolve aggregation expressions and fall back to iteration order
-    final Map<Object, StatelessFunction[]> groups = new LinkedHashMap<>();
+    final Map<Object, SingleKeyGroupState> groups = new LinkedHashMap<>();
 
     final ResultSet prevResults = prev.syncPull(context, CONFIGURED_BATCH_SIZE != null ? CONFIGURED_BATCH_SIZE : nRecords);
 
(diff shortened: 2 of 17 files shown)


lvca:
Addressed both CodeRabbit findings in be66d2627:

1. **Issue6665CallSubqueryCountEquivalenceTest** - now compares the raw \`count(*)\` values against the ground-truth count (not just a derived boolean), plus added the exact-boundary test for the Postgres row cap.

2. **PostgresNetworkExecutor row-cap enforcement** - looked into this further. \`UpdateExecutionPlan.executeInternal()\` (and \`DeleteExecutionPlan\`, which extends it) already fully executes every matched row's write and buffers the entire \`RETURN AFTER\`/\`RETURN BEFORE\` result internally *before* \`UpdateStatement\`/\`DeleteStatement.execute()\` ever returns a \`ResultSet\` to the Postgres layer. So aborting the scan as soon as the cap is exceeded (rather than draining to completion) can never leave a write half-applied - it only avoids scanning the rest of an oversized `SELECT` after the cap is already known to be exceeded, which is strictly better for the OOM-prevention goal this fix exists for. Kept the early-abort behavior and added a regression test (\`updateReturnOverTheCapStillAppliesEveryWrite\`) proving an \`UPDATE ... RETURN\` over the cap still updates every matching row even though its RETURN result is refused.

lvca:
Addressed the remaining CodeRabbit nitpick in 3235921ba: `resultOverTheCapIsRefusedWithAnErrorResponse` now sends a follow-up query on the same socket after the refusal and asserts it succeeds normally, proving the session actually stays usable rather than just trusting the ReadyForQuery status byte.

lvca:
Responded to the \`claude-review\` GitHub Action's notes (posted across 3 runs) in f0228521b:

**Addressed:**
- **#6680 potential NPE**: investigated the flagged \`GuaranteeEmptyCountStep.next()\` non-aggregate branch. Empirically verified \`SELECT x, count(*) AS c FROM EmptyType\` (a non-aggregate field mixed with an aggregate, no GROUP BY, empty input) does **not** NPE - it resolves \`x\` to \`null\` gracefully. Added a regression test pinning this down. This shape already reached the old count(*)-only \`GuaranteeEmptyCountStep\` too (via \`findCountStarItem\`'s "skip non-aggregate items" logic), so it's not new exposure from this PR - just newly verified.
- **#6671 lock-granularity trade-off**: added javadoc to both \`ExecutionPlanCache\` and \`CypherPlanCache\` explicitly noting the deliberate correctness-over-granularity choice, as suggested.

**Judged not worth changing** (all explicitly called "non-blocking" by the reviewer itself):
- **Negative \`POSTGRES_SIMPLE_QUERY_MAX_ROWS\`**: the existing check (\`maxRows > 0\`) already treats any non-positive value as "unlimited," consistent with how \`0\` is documented to mean unlimited. Requires deliberate misconfiguration to hit; not worth a special-case for a value the config's own Javadoc doesn't need to guard beyond what it already does.
- **Buffer growing to \`maxRows + 1\` before the check fires**: explicitly noted by the reviewer as "negligible... not a defect."
- **#6676 extra boxing in the Cypher single-key aggregation hot path**: explicitly noted as "a reasonable trade against the... mantra... flagging only for awareness, not requesting a change."
- **Shared-mutable-\`GlobalConfiguration\` risk across Postgres IT classes**: this project's failsafe config runs with \`forkCount=1\` and no parallel-execution properties, so IT test classes in a module already run sequentially in one JVM - matches the pre-existing pattern other Postgres IT tests (e.g. \`PostgresWireProtocolTestBase\` itself) already use for the same reason.

All CI checks and both CodeRabbit and claude-review passed clean on the latest commits before this one; re-running for f0228521b now.

Issue #6675, opened 2026-08-24
Postgres wire: a TIMESTAMP column backed by java.util.Date is serialized date-only in text (time-of-day dropped) post-#6447


ruispereira:
### Summary
Over the Postgres wire, a `DATETIME`/`TIMESTAMP` column whose runtime value is a `java.util.Date` is serialized in text format as **date-only** (`YYYY-MM-DD`), dropping the time-of-day. The binary path is correct, so text and binary disagree; this is a serialization gap the #6447 OID fix left behind.

### Where (HEAD `86a704526`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresType.java`, `serializeAsText` (L756-758):
```java
} else if (value instanceof Date date) {
  // DATE (OID 1082) expects "YYYY-MM-DD" in text format
  serializedValue = date.toInstant().atZone(ZoneOffset.UTC).format(DateTimeFormatter.ISO_LOCAL_DATE);
} else if (value instanceof LocalDateTime ldt) {
  serializedValue = ldt.format(POSTGRES_DATETIME_FORMATTER);   // correct for TIMESTAMP
}
```
The branch keys on the Java type (`java.util.Date`), not on `pgType`. A column advertised `timestamp` (OID 1114) reaches it whenever its value is a `java.util.Date` — which happens because `getColumns` (PostgresNetworkExecutor.java:899-905) overrides the `DATE` guess to `TIMESTAMP` when the schema declares the property `DATETIME` (the #6447 fix). So post-#6447 the RowDescription says `timestamp` while the DataRow carries only a date.

### Repro
DB configured with `arcadedb.dateTimeImplementation = java.util.Date`; a `DATETIME` property = `2024-01-15T13:45:30` (stored as `java.util.Date`), column advertised `timestamp`, text format → client receives `2024-01-15` and parses `2024-01-15 00:00:00`, silently losing the time. Binary format is correct (so the two disagree).

### Suggested fix
Gate the branch on `pgType == DATE`; for `TIMESTAMP`, format via `LocalDateTime.ofInstant(date.toInstant(), UTC)` + `POSTGRES_DATETIME_FORMATTER` (mirroring `toLocalDateTimeValue`).

### Scope / confidence
[...]

Issue #6674, opened 2026-08-24
Postgres wire: NULL in a BOOLEAN column is encoded as text "0" (false) instead of SQL NULL; text and binary disagree


ruispereira:
### Summary
Over the Postgres wire, a SQL `NULL` in a column advertised `BOOLEAN` is serialized in text format as the 1-byte value `"0"` (which the client reads as `false`) instead of the SQL-NULL sentinel (length `-1`). The binary path correctly emits `-1`, so the two RowDescription-consistent encodings of the same value disagree, and the NULL is lost in the common (text) path.

### Where (HEAD `86a704526`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresType.java`, `serializeAsText` (L731-732):
```java
if (value == null && pgType.code == BOOLEAN.code) {
  serializedValue = "0";
}
...
writeString(typeBuffer, serializedValue);   // writes "0" (length 1), not the -1 NULL sentinel
```
Every other type falls through to `writeString(buffer, null)` → `-1` (verified: `serializeAsTextNull`). `serializeAsBinary` (L784-786) also emits `-1` for the same null. Only BOOLEAN deviates, and only in text. (Non-null booleans correctly emit `t`/`f`.)

### Repro
A column advertised `bool` (OID 16 — announced whenever the first sampled row's value is a `Boolean`) with a later/mixed row holding a null boolean, in the default text result format → the client receives a non-null `false` and loses the NULL. The same column over binary format returns NULL.

### Suggested fix
Delete the special case; let a null boolean fall through to `writeString(buffer, null)` → `-1`, exactly like every other type.

### Scope / confidence
High on the behavior (null→text "0", non-null→t/f). Honest caveat: a test (`PostgresTypeTest.serializeAsTextNullBoolean`) currently locks in `"0"`, so intent is ambiguous and fixing it is a deliberate behavior change — but it is a genuine NULL-fidelity + text/binary inconsistency.

lvca:
Fixed in PR #6697 (26.9.1). Over the Postgres wire protocol, a NULL value in a BOOLEAN column was being sent as the text `false` instead of NULL when using the default text result format. It now correctly reports NULL, matching binary format.

---
❤️ Thanks for helping make ArcadeDB better! ArcadeDB is free and open source, sustained by its community. If it's useful to you or your company, please consider [becoming a sponsor](https://github.com/sponsors/ArcadeData) to keep development going.

Issue #6659, opened 2026-08-23
Postgres Execute portal suspension materializes the whole result server-side instead of streaming


robfrank:
### Summary

Follow-up from #6658 (review comment): fixing the wire-protocol ordering/double-terminator/portal-removal defects in Execute's row-limit handling (portal suspension) required materializing a portal's **entire** result into memory (`PostgresPortal.fullResultSet`, a `List<Result>`) up front, then slicing it per `Execute` call via a cursor. This is correct on the wire (rows always precede exactly one terminator, a follow-up `Execute` continues from where the last one stopped) but is not true streaming: the server pulls the whole result set into memory before returning even the first row, and keeps that list retained for the whole lifetime of the portal - not a transient spike.

For a client that always sends `Describe('P')` before its first `Execute` (e.g. pgjdbc) this is not a regression: `describeCommand()` already had to fully materialize the result to discover a sparse document's complete column set, before #6658 too. But for a raw wire-protocol client or driver that skips `Describe` and relies purely on `Execute`'s max-row count to bound memory/latency over a large table, the server-side behavior is now uniformly "materialize everything, then paginate in memory" rather than "pull only what was asked for."

### Where

`postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java`:
- `executeCommand()`'s first-execution branch and `describeCommand()`'s `Describe('P')` branch both call `browseAndCacheResultSet(resultSet, 0)` (unconditional full drain) into `PostgresPortal.fullResultSet`.
- The pagination step in `executeCommand()` slices that materialized list via `PostgresPortal.resultCursor`.

### Why a bigger change

True incremental pagination needs the source `ResultSet` kept open across `Execute` calls instead of closed immediately (`browseAndCacheResultSet`'s current ownership contract). That in turn means:
- `Describe('P')`'s column-discovery can no longer assume it has seen every row before answering - documents can be sparse, so a property that only appears on a later row would be invisible to `RowDescription` if it stops sampling early.
- The portal needs to hold and eventually close a live resource across an arbitrary number of client round trips (including a client that disconnects mid-fetch without ever draining or explicitly closing the portal).

### Scope / confidence

Low urgency: no known client is broken by the current eager-materialization behavior (pgjdbc already got the same eager-materialization treatment via `Describe` before #6658, and tolerates it fine - see #6658's own test suite). This is a memory/GC concern for very large result sets fetched by a client that skips `Describe`, not a correctness bug.

[...]

PR #6697, opened 2026-08-24
fix(#6674, #6675, #6423): Postgres text NULL boolean, TIMESTAMP-as-Date, and key=value truncation

lvca:
## Summary

Fixes three related Postgres-wire / shared-parser issues:

- **#6674**: a null `BOOLEAN` over the Postgres wire's text format was encoded as the 1-byte value `"0"` (read back by the client as `false`) instead of the SQL-NULL sentinel (length `-1`). The binary path already emitted `-1` correctly. Fix: delete the special case so BOOLEAN falls through to the same null handling as every other type.
- **#6675**: a `TIMESTAMP` column backed by `java.util.Date` (with `arcadedb.dateTimeImplementation = java.util.Date`) was serialized date-only in text format, dropping the time-of-day - a serialization gap left behind by #6447's OID fix. Fix: gate the date-only branch on `pgType == DATE`; any other type backed by a `Date` (i.e. `TIMESTAMP`) now formats through `LocalDateTime` + the existing Postgres timestamp formatter, keeping the time.
- **#6423**: two `key=value` parsers - `PostgresNetworkExecutor.setConfiguration()` (the Postgres wire `SET` command) and `GetDynamicContentHandler`'s `${include:file.html k=v}` directive - split on *every* `=` instead of the first one, silently truncating a value containing a further `=` (e.g. a connection string). The include directive also threw `ArrayIndexOutOfBoundsException` on a parameter with no `=` at all. Both are fixed by delegating to a new `StringUtils.splitKeyValue()`, the same first-separator-only rule `Console.splitKeyValue()` already used for the SET/-D case (issue #6392) - `Console.splitKeyValue()` now delegates to it too instead of carrying its own copy. The include directive now logs and skips a malformed parameter instead of throwing.

## Deferred: #6659

#6659 is a follow-up review comment on PR #6658 (fix for #6458), which is **still open/unmerged**. The "eager full materialization" behavior #6659 describes doesn't exist on `main` yet - main still has #6458's original, more severe defects. Implementing #6659 now would collide with #6658's in-flight changes to the same code. Reasoning and a suggested alternative (skip the eager-materialization step and go straight to incremental streaming once #6658 lands) are posted as a comment on the issue.

## Test plan

- [x] `PostgresTypeTest`: fixed `serializeAsTextNullBoolean` (now asserts the `-1` sentinel) and added `serializeAsTextTimestampBackedByDate` - 225 tests, all green
- [x] New `PostgresSetCommandParsingTest` - 7 tests covering `=`, `TO`, quoting, and the value-containing-`=` regression - all green
- [x] New `StringUtilsSplitKeyValueTest` (mirrors `ConsoleSplitKeyValueTest`'s contract) - 5 tests, all green
- [x] New `GetDynamicContentHandlerIncludeParametersTest` - 3 tests covering well-formed params, a value containing `=`, and a param with no `=` (no longer throws) - all green
- [x] Full `postgresw` module (`mvn verify -DskipITs=false -DexcludedGroups=benchmark,slow,vector`): 416 unit + 175 integration tests, all green
- [x] `console` + `engine` modules (`mvn test -DexcludedGroups=benchmark,slow,vector`): all green (2 unrelated pre-existing flaky vector/timing tests seen only when those lanes are included, as expected - not touched by this change)
- [x] `server` module `GetDynamicContentHandlerIT`: 14 tests, all green

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parsing of key-value settings and template parameters, preserving additional `=` characters in values.
  - Malformed template include parameters are now skipped safely instead of causing errors.
  - PostgreSQL `SET` commands now support `=` and case-insensitive `TO` separators reliably.
  - Corrected PostgreSQL serialization for null booleans and timestamps, preserving date and time information.

- **Tests**
  - Added regression coverage for key-value parsing, command handling, template parameters, and PostgreSQL value serialization.

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -45,6 +45,7 @@ import com.arcadedb.schema.DocumentType;
 import com.arcadedb.utility.AnsiCode;
 import com.arcadedb.utility.FileUtils;
 import com.arcadedb.utility.RecordTableFormatter;
+import com.arcadedb.utility.StringUtils;
 import com.arcadedb.utility.TableFormatter;
 import org.jline.reader.Completer;
 import org.jline.reader.EndOfFileException;
@@ -1188,17 +1189,13 @@ public class Console {
   }
 
   /**
-   * Splits a `&lt;key&gt;=&lt;value&gt;` pair on the FIRST '=' only, so a value that contains further separators (a connection
-   * string, a base64 padding, a date pattern) is kept whole and an empty value is a value, not a missing one. Returns null when
-   * there is no separator at all, leaving to the caller the choice between rejecting the input and reading it as an empty value.
-   * Shared by the `-D&lt;key&gt;=&lt;value&gt;` arguments (issue #5928) and by the SET command (issue #6392), which used to
-   * disagree on the rule.
+   * Splits a `&lt;key&gt;=&lt;value&gt;` pair on the FIRST '=' only. Shared by the `-D&lt;key&gt;=&lt;value&gt;`
+   * arguments (issue #5928) and by the SET command (issue #6392), which used to disagree on the rule; delegates to
+   * {@link StringUtils#splitKeyValue} so the Postgres wire's SET command and the Studio include directive
+   * (issue #6423) follow the same rule instead of each carrying their own truncating copy.
    */
   static String[] splitKeyValue(final String pair) {
-    final int separatorPos = pair.indexOf('=');
-    if (separatorPos < 0)
-      return null;
-    return new String[] { pair.substring(0, separatorPos), pair.substring(separatorPos + 1) };
+    return StringUtils.splitKeyValue(pair);
   }
 
   private static boolean setGlobalConfiguration(final String key, final String value, final boolean printError) {
--- a/engine/src/main/java/com/arcadedb/utility/StringUtils.java
+++ b/engine/src/main/java/com/arcadedb/utility/StringUtils.java
@@ -37,18 +37,6 @@ public class StringUtils {
   }
 
   /**
-   * Case-insensitive {@code contains} that allocates nothing.
-   * <p>
-   * The obvious form, {@code haystack.toUpperCase().contains(needle)}, copies the whole haystack to answer a question
-   * about a needle a dozen characters long - which is why every caller of this had written its own copy rather than
-   * use it: the async DDL classifier scanning a whole script, the Cypher parser deciding on its own parse hot path,
-   * the HTTP handler looking for a LIMIT in a command, and the MCP prompt fence looking for its own delimiter. Four
-   * private copies of six lines, and this is the one they share.
-   *
-   * @return true when {@code needle} occurs anywhere in {@code haystack}, ignoring case. An empty needle is contained
-   *     by everything, as {@code String.contains} has it.
-   */
-  /**
    * Splits a {@code <key>=<value>} pair on the FIRST '=' only, so a value that contains further separators (a
    * connection string, a base64 padding, a date pattern) is kept whole and an empty value is a value, not a missing
    * one. Returns null when there is no separator at all, leaving to the caller the choice between rejecting the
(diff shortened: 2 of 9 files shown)


lvca:
Re CodeRabbit's SESSION/LOCAL + trailing-semicolon finding (major, on commit 243b162e): confirmed both are real, but pre-existing and unrelated to this PR's scope (the \`key=value\` truncation bug in #6423) - not introduced or worsened by this PR:

- \`SET SESSION\`/\`SET LOCAL\` were never recognized before this PR either; \`parseSetCommand()\` has no notion of scope modifiers at all.
- The simple-query protocol path (\`queryCommand()\`) already strips a trailing \`;\` before calling \`setConfiguration()\`; the extended-protocol path (\`parseCommand()\`) does not, and never did.
- My quote-guard fix (commit aca02756f) changes the failure mode for the semicolon case from "silently store a corrupted value" to "explicitly reject the malformed command and log a warning" - strictly better, not a regression.

Implementing real \`SESSION\`/\`LOCAL\` scope semantics is a genuine design question (ArcadeDB has no transaction-scoped config today, so \`LOCAL\` needs a decision on whether to actually implement that scoping or just treat it as session-scoped), which makes it a heavier, separate piece of work rather than a quick fix bundled into this truncation-bug PR. Filed as #6701 to track both points with a suggested fix, rather than silently dropping the finding.

lvca:
Re CodeRabbit's finding on commit aca02756f (\`SET search_path TO "\$user", public\` now rejected): confirmed and same root cause as the SESSION/LOCAL + trailing-semicolon gap already filed as #6701 - \`parseSetCommand()\`'s quote handling assumes a value is either fully quoted or fully unquoted, never a Postgres-style comma-separated value list with only some elements quoted.

This predates PR #6697: the pre-existing code also mishandled this exact input (\`substring(1, length-1)\` unconditionally stripped the first \`"\` and the last char \`c\` of \`public\`, silently producing the corrupted value \`\$user", publi\`). My quote-guard fix changes that failure mode from silent corruption to an explicit rejection + logged warning - an improvement, but real support needs actual value-list tokenization, which is a bigger, separate piece of work. Added it to #6701 rather than a fourth issue, since it's the same underlying gap in the same method.

lvca:
Merged latest \`main\` into this branch (no conflicts) to pick up a few other PRs that landed in the meantime.

Note on the previous CI run: \`unit-tests\` failed with exactly one failure across 13,377 tests - \`GraphAnalyticalViewTest.issue6641_interveningCommitInvalidatesDeferredRestoreWithoutBlockingQuery\`, in \`com.arcadedb.graph.olap\` (the recently-merged #6641 GAV deferred-rebuild fix). This PR's diff touches only \`postgresw\`, \`engine/utility/StringUtils.java\`, \`console/Console.java\`, and \`server/.../GetDynamicContentHandler.java\` - no overlap with graph/OLAP code - and the failure looks like an async/timing race ("provisionally-READY view... must not be advertised as ready until verified", asserting false but got true) rather than anything this PR could cause. Not touching it; the merge above will give it a fresh run.

lvca:
Status update on \`ha-integration-tests\`: it has now failed twice, with **four different, non-overlapping tests** across the two runs - \`Bolt5002RoutingTableIT\` (integration-tests, first run), then \`RaftPriorityRejoinIT\`, \`Issue6091GrpcRoutingTableIT\`, and \`RaftRemoteReadYourWritesIT\` (ha-integration-tests, second run). All four are in \`ha-raft\`/\`server.ha.raft\` - Raft consensus leader-restart/rejoin/routing-table/read-your-writes timing tests - a module this PR's diff never touches (only \`postgresw\`, \`engine/utility/StringUtils.java\`, \`console/Console.java\`, and \`server/.../GetDynamicContentHandler.java\`). Different tests failing each run, all timing/consensus-sensitive, is the signature of environmental flakiness rather than a regression this PR introduced. \`unit-tests\`, \`integration-tests\`, and \`js-e2e-tests\` all had a similar one-off flaky failure and are now green on rerun. Triggered a third attempt on \`ha-integration-tests\`.

Issue #6376, opened 2026-08-18
algo.maxKCut aligns edge weights positionally against CSR adjacency: wrong weights when neighbor order ≠ OLTP order (the #6301 class)


ruispereira:
### Summary
`algo.maxKCut` fills its per-edge weight array **positionally** against a CSR adjacency list that is ordered differently, so `adjW[i][j]` is not the weight of the edge to `adj[i][j]`. It produces silently wrong cut weights and partition assignments on any weighted call over a graph with a Graph Analytical View (GAV) or a relTypes filter. This is the same defect #6301 fixed elsewhere; `maxKCut` was not converted.

### Where (HEAD `ca89f56`)
`engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoMaxKCut.java`, `buildWeightedAdj` (L207-236), consumed at L139 and in `computeCutWeight` (L239-246):

```java
final int[][] adj = graph.adjacency(dir, relTypes);   // caller: neighbors ordered by dense node id
...
for (int i = 0; i < n; i++) {
  final Iterable<Edge> edges = ... graph.getVertex(i).getEdges(dir, relTypes) ...; // OLTP order
  int pos = 0;
  for (final Edge e : edges) {
    final RID nbRid = neighborRid(e, graph.getRID(i), dir);
    if (nbRid == null || graph.indexOf(nbRid) < 0) continue;
    if (pos < adjW[i].length) {
      adjW[i][pos] = w instanceof Number num ? num.doubleValue() : 1.0;   // positional assign
      pos++;
    }
  }
}
```

`adj` comes from `graph.adjacency(...)` (CSR/GAV: neighbors sorted by dense node id). `buildWeightedAdj` walks `getEdges(...)` in OLTP order and assigns weights by position. `GraphAnalyticalView`'s own doc states *"Neighbor order is unrelated to OLTP order"*, so the two orders differ whenever a GAV provider exists (or a relTypes filter reorders edges) — and then `adjW[i][j]` pairs with the wrong `adj[i][j]`. The other weighted algos (APSP, Bellman-Ford, Steiner, kShortestPaths CSR path) were all switched to `graph.weightedAdjacency(...)` by #6301; `maxKCut` still hand-rolls the parallel array.

### Effect
[...]

Issue #6316, opened 2026-08-17
Six algo.* procedures never look for a Graph Analytical View at all, and hand-roll the loading and weight extraction the shared helpers now own


lvca:
Raised in review of PR #6306 (issues #6295/#6300/#6301/#6302) and deliberately not folded into it: the reviewer proposed routing `algo.kShortestPaths` and `algo.msa` through the new `GraphData.weightedAdjacency`, and that was declined there because it is not a bug fix. It is worth doing on its own terms, and the set is larger than those two.

## The set

Verified on `0237e7b` - these six contain **zero** references to `loadGraph` or `findProvider`:

| Procedure | Cost | Loads with |
|---|---|---|
| `algo.kShortestPaths` | O(k x pathLength x V²) | `getAllVertices` + `buildRidIndex` |
| `algo.mst` | O(E log E) | `getAllVertices` + `buildRidIndex` |
| `algo.msa` | O(V x E) | `getAllVertices` + `buildRidIndex` |
| `algo.maxFlow` | O(V x E²) | `getAllVertices` + `buildRidIndex` |
| `algo.betweenness` | O(V x E) | `getAllVertices` + `HashMap<Vertex, Integer>` |
| `algo.degreeCentrality` | O(V + E) | `getAllVertices` |

Every other procedure in the package resolves a provider through `loadGraph` and uses the CSR when one is available. **These six cannot be accelerated by a Graph Analytical View at all** - building one changes nothing for them. Five of the six are among the most expensive procedures in the package, which is the opposite of the set you would choose to leave on the OLTP path.

## Two things follow from it

**They duplicate the code #6306 consolidated.** `algo.kShortestPaths` and `algo.msa` each hand-roll the RID resolution, weight extraction and ghost-edge skip that `GraphData.weightedAdjacency` now owns. Neither carries the #6301 alignment bug - both resolve the neighbour by index (`ridToIdx.get(e.getIn())`) rather than by iteration position, which is why they were correct - but they are independent copies of a pattern that has now been got wrong four times in this package. A future fix to the shared helper will not reach them.

**`algo.betweenness` keys its index by `Vertex` rather than by `RID`.** `HashMap<Vertex, Integer>` relies on `Vertex.equals`/`hashCode` where every other procedure uses `buildRidIndex`'s `Map<RID, Integer>`. Worth checking on its own: two `Vertex` instances for the same record are equal only if the implementation says so, and a mismatch here silently drops neighbours from the BFS rather than failing.

[...]

Issue #4287, opened 2026-05-22
algo.leiden() relTypes parameter produces singleton communities regardless of separator format


azfeenixfire:
Description:
When calling algo.leiden() with a relTypes argument (tested with both pipe | and comma , separators), the procedure runs without error but returns every node as its own singleton community. Calling algo.leiden() with no arguments correctly produces meaningful multi-node communities on the same graph.

To reproduce:
cypher-- Returns singleton communities for every node (incorrect)
CALL algo.leiden("BELONGS_TO,PREREQUISITE_FOR,SIMILAR_TO,DECOMPOSES_INTO")
YIELD nodeId, community
RETURN community, count(*) as size ORDER BY size DESC

-- Also tried pipe separator (same result)
CALL algo.leiden("BELONGS_TO|PREREQUISITE_FOR|SIMILAR_TO|DECOMPOSES_INTO")
YIELD nodeId, community
RETURN community, count(*) as size ORDER BY size DESC

-- Returns correct multi-node communities
CALL algo.leiden()
YIELD nodeId, community
RETURN community, count(*) as size ORDER BY size DESC

Expected behavior:
Filtering by relationship type should return communities formed only from nodes connected by those relationship types, with the same quality of clustering as the unfiltered call.

Actual behavior:
Every node is assigned to its own singleton community when any relTypes argument is passed, regardless of whether those relationship types exist in the graph.

Suspected cause:
The extractRelTypes method in AlgoLeiden.java (called via AbstractAlgoProcedure) likely isn't parsing the relTypes string correctly, resulting in no relationships being matched during graph load. The loadGraph call receives a non-null but effectively empty or unmatched relTypes array, causing every node to appear disconnected.

Environment:
ArcadeDB version: latest docker as of may 2026
Database: 439 nodes, 978 edges
Verified relationships exist on nodes via: SELECT bothE().@type as rel_types FROM LearningObjective LIMIT 5

PR #6700, opened 2026-08-24
fix(#6699): algo.wcc applies its relTypes filter

### LINKED ISSUE #6699: `algo.wcc` returns the all-edges result when a `relTypes` filter is given
### ArcadeDB Version:

`26.9.1-SNAPSHOT (build 7ca1a96c19777528a9c811cd36259845ca67fd6f/1787558836676/main)`

### OS and JDK Version:

`Alpine Linux v3.23 (arm64, Docker) / OpenJDK Temurin-21.0.11+10 (build 21.0.11+10-LTS)`

### Expected behavior

`algo.wcc([relTypes])` restricts the traversal to the listed edge types.

### Actual behavior

The filter does not take effect, so the result is the whole-graph answer whatever is passed.
There is no error, which makes it easy to miss.

### Steps to reproduce

Four vertices and two edge types. No Graph Analytical View needed.

```sql
CREATE VERTEX TYPE V;
CREATE EDGE TYPE A;
CREATE EDGE TYPE B;
CREATE PROPERTY V.k INTEGER;
INSERT INTO V SET k = 1; INSERT INTO V SET k = 2;
INSERT INTO V SET k = 3; INSERT INTO V SET k = 4;
CREATE EDGE A FROM (SELECT FROM V WHERE k=1) TO (SELECT FROM V WHERE k=2);
CREATE EDGE A FROM (SELECT FROM V WHERE k=3) TO (SELECT FROM V WHERE k=4);
CREATE EDGE B FROM (SELECT FROM V WHERE k=2) TO (SELECT FROM V WHERE k=3);
```

`A` links 1-2 and 3-4; `B` links 2-3.

```cypher
CALL algo.wcc('A') YIELD node, componentId
  WITH componentId, count(*) AS c RETURN count(*) AS components, max(c) AS largest
```

| call | expected | observed |
|---|---|---|
| `algo.wcc('A')` | 2 components, largest 2 | 1 component, largest 4 |
| `algo.wcc('B')` | 3 components, largest 2 | 1 component, largest 4 |
| `algo.wcc('NOSUCH')` | 4 components, largest 1 | 1 component, largest 4 |
| `algo.wcc()` | 1 component, largest 4 | 1 component, largest 4 (as expected) |

The `NOSUCH` row is probably the clearest signal: an edge type with no edges at all should
leave every vertex on its own.

`algo.leiden` handles the same graph as expected, returning 2 communities for `'A'` and 3 for
`'B'`, so the shared `extractRelTypes` helper is doing its job.

### Cause

`execute()` passes `null` where the filter would go, in both `findProvider` and `loadGraph`,
so the argument never reaches the traversal. `AlgoLeiden` threads it through the equivalent
calls, so there is an existing pattern to follow. Details and a fix in the PR.

The other procedures in the package do read their argument, so this looks confined to
`algo.wcc`.

A PR with the change and a regression test follows.

### For context

Related work on `relTypes` handling elsewhere in the package: #4287 / #4360 (`algo.leiden`),
#6301 (`algo.steinerTree`), #6376 (`algo.maxKCut`). #6316 covers six procedures that do not
resolve a provider at all; `algo.wcc` is not one of those, since it does call
`findProvider`/`loadGraph`.

ZwaarContrast:
## What does this PR do?

Makes `algo.wcc([relTypes])` restrict the traversal to the listed edge types.

`execute()` now reads the argument with the existing `extractRelTypes` helper and passes it on,
following the pattern `AlgoLeiden` already uses:

- `loadGraph(db, null, relTypes, null)` in place of `loadGraph(db, null, null, null)`
- `graph.adjacency(Vertex.DIRECTION.BOTH, relTypes)` in place of the unfiltered overload

The direct `executeWithCSR` shortcut is now taken only for an unfiltered call.
`GraphAlgorithms.connectedComponents()` walks the whole view, and a provider is selected for
covering the requested edge types rather than for covering only those, so a view spanning
`{A, B}` satisfies a request for `'A'` and the shortcut would union edges the caller asked to
leave out. A filtered call goes through `loadGraph`, which applies `relTypes` to CSR adjacency
via `buildAdjacencyFromProvider`, so acceleration is kept without widening the result.

## Motivation

The filter did not take effect, so the result was the whole-graph answer whatever was passed,
with no error to signal it. Filtering on an edge type with no edges returned one component
containing everything, where each vertex should have been on its own.

## Related issues

Fixes #6699.

For context, related `relTypes` work elsewhere in the package: #4287 / #4360
(`algo.leiden`), #6301 (`algo.steinerTree`), #6376 (`algo.maxKCut`). Separate from #6316,
which covers procedures that do not resolve a provider at all.

## Checklist

On the first box: the reactor runs 17,598 tests here with `AlgoWCCTest` at 9/0 and the
3,897-test `OpenCypherTCKSuite` clean, but it does not finish green in my container. Each
remaining failure was checked separately and none involve this change:
`DatabaseGetSizeTest`, `PromQLEvaluatorIntegrationTest` and
`Issue6440FlushAckOnClosedDatabaseTest` fail the same way on an unmodified checkout, and
`WindowFunctionTest` / `TimeSeriesFunctionDeltaTest` gave 0, 1 and 3 failures across three
identical runs of the same tree. Happy to re-run anything that would help.

Five tests added to `AlgoWCCTest`:

| test | covers |
|---|---|
| `wccWithRelTypesRestrictsToThoseEdges` | `wcc('EDGE')`: `{A,B,C}` share an id, `{D,E}` share another, the two differ |
| `wccWithRelTypesIgnoresOtherEdgeTypes` | `wcc('OTHER')`: `{C,D}` together, with A, B and E as three distinct singletons |
| `wccWithUnknownRelTypeIsolatesEveryVertex` | `wcc('NOSUCH')`: five distinct ids, since that type has no edges |
| `wccWithCommaSeparatedRelTypesCoversBoth` | `wcc('EDGE,OTHER')`: matches the unfiltered result, covering comma-separated parsing |
| `wccWithRelTypesIsNotWidenedByAViewCoveringMoreTypes` | builds a view over both edge types, then checks `wcc('EDGE')` still separates the two components |

The existing tests continue to cover the unfiltered call.

Run both ways:

```
with the change:     Tests run: 9, Failures: 0   BUILD SUCCESS
without the change:  Tests run: 9, Failures: 4   BUILD FAILURE
```

`wccWithCommaSeparatedRelTypesCoversBoth` passes either way, since `'EDGE,OTHER'` is every edge
type in the fixture; it guards the separator handling rather than this behaviour. The other four
fail without the change.

## Summary by CodeRabbit

* **New Features**
  * Added relationship-type filtering to Weakly Connected Components analysis.
  * Supports filtering by one or multiple relationship types, including analytical views and mixed edge types.
  * Filtered analyses can use accelerated processing when available.

* **Bug Fixes**
  * Improved component detection for graphs containing multiple relationship types.
  * Added graceful handling for unknown relationship types without affecting unfiltered analysis.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoWCC.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoWCC.java
@@ -90,28 +90,27 @@ public class AlgoWCC extends AbstractAlgoProcedure {
 
     final String[] relTypes = args.length > 0 ? extractRelTypes(args[0]) : null;
 
-    // Try CSR-accelerated path: delegate to native union-find on CSR arrays. Only for an
-    // unfiltered call: connectedComponents() walks the whole view, and a provider is selected
-    // for covering the requested types, not for covering only those, so a filtered call would
-    // union edges it was asked to leave out. loadGraph() applies the filter to CSR adjacency.
-    if (relTypes == null || relTypes.length == 0) {
-      final GraphTraversalProvider provider = findProvider(db, null);
-      if (provider instanceof GraphAnalyticalView gav) {
-        context.setVariable(CommandContext.CSR_ACCELERATED_VAR, true);
-        return executeWithCSR(context, gav);
-      }
+    // Try CSR-accelerated path: delegate to native min-label propagation on CSR arrays.
+    // findProvider(relTypes) only returns a provider that covers every requested type, and
+    // GraphAlgorithms.connectedComponents() takes the same relTypes and resolves each one to its
+    // own CSR index (GraphAnalyticalView#getCSRIndex), so passing the filter through keeps the
+    // parallel algorithm for a filtered call instead of dropping to the single-threaded BFS below.
+    final GraphTraversalProvider provider = findProvider(db, relTypes);
+    if (provider instanceof GraphAnalyticalView gav) {
+      context.setVariable(CommandContext.CSR_ACCELERATED_VAR, true);
+      return executeWithCSR(context, gav, relTypes);
     }
 
     // Fall back to OLTP path
     return executeWithOLTP(db, relTypes);
   }
 
-  private Stream<Result> executeWithCSR(final CommandContext context, final GraphAnalyticalView gav) {
+  private Stream<Result> executeWithCSR(final CommandContext context, final GraphAnalyticalView gav, final String[] relTypes) {
     final int n = gav.getNodeCount();
     if (n == 0)
       return Stream.empty();
 
-    final int[] componentId = GraphAlgorithms.connectedComponents(gav);
+    final int[] componentId = GraphAlgorithms.connectedComponents(gav, relTypes);
     context.setVariable(CommandContext.RESULT_COUNT_HINT_VAR, (long) n);
 
     return IntStream.range(0, n).mapToObj(i -> {
--- a/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoWCC.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/AlgoWCC.java
@@ -88,15 +88,22 @@ public class AlgoWCC extends AbstractAlgoProcedure {
 
     final Database db = context.getDatabase();
 
-    // Try CSR-accelerated path: delegate to native union-find on CSR arrays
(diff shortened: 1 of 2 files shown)


lvca:
Thanks @ZwaarContrast !

PR #6702, opened 2026-08-24
fix(#6694): descending scan's root-page probe used ascending purpose for a composite-index prefix

### LINKED ISSUE #6694: LSMTreeIndexCompacted.newIterators() root-page lookup always uses ascending purpose, even for descending scans
Follow-up observation from the automated code review on #6692 (which fixed two related descending-scan bugs in \`LSMTreeIndexCompacted\`).

In \`LSMTreeIndexCompacted.newIterators()\`:

```java
final int fromPurpose = fromKeys.length < binaryKeyTypes.length ? 2 : 1;
final LookupResult resultInRootPage = lookupInPage(rootPageNumber, rootPageCount + 1, rootPageBuffer, fromKeys, fromPurpose);
```

For a partial (composite-index prefix) key, `fromPurpose` is always `2` (ascending), **regardless of `ascendingOrder`**. This means the coarse root-page-level lookup that picks which data page to start `searchInCurrentPage()` from is always done with ascending semantics, even when the actual scan is descending.

This is unrelated to the two bugs fixed in #6692 (both were in `compareKey()`'s partial-match boundary handling and in `searchInCurrentPage()`'s "outside" page-stepping direction) - the reviewer confirmed the new regression test doesn't exercise this specific root-page-level branch, since the target group in that test lands within a single data page found directly by the existing binary search.

Given how subtle the ascending/descending asymmetry in this file has already proven (see #6592's history), this is worth a dedicated look: does using purpose=2 for the root-page probe on a descending scan ever pick the wrong starting data page for a *multi-page-per-series* partial-key group? If so, `searchInCurrentPage()`'s own data-page-level lookup (which *does* correctly use `ascendingOrder ? 2 : 3`) might still recover via the "outside" fallback now fixed in #6692 - but that hasn't been verified with a dedicated test the way the #6592 scenario was.

Not urgent / not blocking anything today - filed so it isn't lost.

lvca:
## Summary

Fixes #6694.

`LSMTreeIndexCompacted.newIterators()`'s coarse root-page lookup - which picks which data
page to start `searchInCurrentPage()` from - always used `purpose=2` (ascending) for a
partial (composite-index prefix) key, regardless of the requested scan direction.

`compareKey()`'s "PARTIAL MATCHING" walk resolves an ambiguous binary-search landing point
to the **first** entry of a same-prefix run for `purpose=2` and the **last** entry for
`purpose=3`. Using `purpose=2` unconditionally meant a **descending** scan's root-page probe
always landed on a series' FIRST (lowest-keyed) data page instead of its LAST (highest-keyed)
one whenever a composite-index prefix match spanned multiple data pages within one compacted
series - so the scan started too low and silently **dropped whole series/pages**.

This is the same asymmetry #6692 already fixed at the DATA-page level
(`searchInCurrentPage()`, which correctly uses `ascendingOrder ? 2 : 3`); this PR applies the
identical fix at the coarser ROOT-page level, one layer up.

## Fix

Mirror the existing `ascendingOrder ? 2 : 3` pattern for the root-page purpose selection when
the key is partial:

```java
final int fromPurpose = fromKeys.length < binaryKeyTypes.length ? (ascendingOrder ? 2 : 3) : 1;
```

Full keys are untouched (still always `purpose=1`, as before).

## Test plan

- [x] New regression test `Issue6694CompactedRootPageDescendingPurposeTest`: builds a single
  composite-index prefix group (`key1='a', key2='b'`) large enough (4,000 rows, small page
  size) to span many data pages within a compacted series, then compacts and runs a
  `DESC` scan/`ORDER BY ... DESC`.
  - Reproduces the bug against unfixed code: only 641 of 4,000 expected rows returned
    (whole compacted series silently dropped).
  - Passes cleanly after the fix (all 3 test methods).
- [x] No-regression run: `Issue6592CompactedCompositePrefixDescTest`, `Issue5214MultiSeriesRangeTest`,
  `SelectCompositeIndexTest` - all green.
- [x] Full `com.arcadedb.index.lsm.*` package (20 test classes) - all green.
- [x] Broader index-package sweep (`LSMTreeSortedNonUniqueBuildTest`,
  `CompositeIndexPartialKeySelectTest`, `LSMTreeCompactionCorrectnessTest`,
  `Issue5662IndexCursorFollowUpsTest`, `Issue5517BloomFilterTest`,
  `Issue5517BloomFilterKeyTypesTest`, `Issue5386OutOfRangeSeekTest`,
  `LSMTreeIndexCompactionTest`, `Issue5681UnclosedDmlLeavesScanOpenTest`,
  `LSMTreeIndexKeyOrderCheckTest`, `Issue5802KeyOrderUpgradeWarningTest`,
  `DeleteFromIndexStepCloseTest`, `FileManagerTest`) - all green.
- [x] `mvn -pl engine -am compile` - clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed descending queries on compacted composite indexes when filtering by partial keys.
  - Ensured complete and correctly ordered results when matching entries span multiple data pages.
  - Preserved expected behavior for full-key and ascending partial-key lookups.

- **Tests**
  - Added regression coverage for descending, ascending, limited, and full-result index queries.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCompacted.java
+++ b/engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCompacted.java
@@ -500,13 +500,19 @@ public class LSMTreeIndexCompacted extends LSMTreeIndexAbstract {
       if (fromKeys != null) {
         final Binary rootPageBuffer = new Binary(rootPage.slice());
 
-        // Use purpose=2 (ascending iterator) instead of 1 (retrieve) for root page lookups when
+        // Use purpose=2/3 (ascending/descending iterator) instead of 1 (retrieve) for root page lookups when
         // keys are partial (fewer components than the composite index defines). Purpose=1 rejects
         // partial keys with "key is composed of N items, while the index defined M items".
-        // Purpose=2 allows partial key comparison which correctly matches by prefix.
+        // Purpose=2/3 allows partial key comparison, which correctly matches by prefix, and - like the data-page
+        // lookup in searchInCurrentPage() below - must follow the scan direction: compareKey()'s PARTIAL MATCHING
+        // walk resolves an ambiguous binary-search landing point to the FIRST entry of a same-prefix run for
+        // purpose=2 and the LAST entry for purpose=3. Using purpose=2 unconditionally made a descending scan's
+        // root-page probe always land on the series' FIRST (lowest-keyed) data page instead of its LAST
+        // (highest-keyed) one whenever a composite-index prefix match spanned multiple data pages within one
+        // compacted series, so the scan started too low and silently dropped whole series/pages (#6694).
         // For full keys, keep purpose=1 to preserve exact boundary behavior for descending ranges and the shared-leaf
         // (multi-page duplicate) positioning, both of which depend on purpose=1's value-run result.
-        final int fromPurpose = fromKeys.length < binaryKeyTypes.length ? 2 : 1;
+        final int fromPurpose = fromKeys.length < binaryKeyTypes.length ? (ascendingOrder ? 2 : 3) : 1;
         final LookupResult resultInRootPage = lookupInPage(rootPageNumber, rootPageCount + 1, rootPageBuffer, fromKeys,
             fromPurpose);
         if (!resultInRootPage.outside)
--- a/engine/src/test/java/com/arcadedb/query/select/Issue6694CompactedRootPageDescendingPurposeTest.java
+++ b/engine/src/test/java/com/arcadedb/query/select/Issue6694CompactedRootPageDescendingPurposeTest.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.query.select;
+
+import com.arcadedb.GlobalConfiguration;
+import com.arcadedb.TestHelper;
+import com.arcadedb.graph.Vertex;
+import com.arcadedb.index.IndexInternal;
+import com.arcadedb.index.TypeIndex;
+import com.arcadedb.schema.Schema;
+import com.arcadedb.schema.Type;
+import com.arcadedb.schema.VertexType;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Regression test for https://github.com/ArcadeData/arcadedb/issues/6694: {@code LSMTreeIndexCompacted.newIterators()}'s
+ * ROOT-PAGE lookup for a partial (composite-index prefix) key always used purpose=2 (ascending), regardless of the
+ * requested scan direction. #6692 fixed the analogous asymmetry at the DATA-page level ({@code searchInCurrentPage()},
+ * which already used {@code ascendingOrder ? 2 : 3}), but left the coarser root-page probe that PICKS which data page
+ * to start {@code searchInCurrentPage()} from always ascending.
+ * <p>
+ * This only surfaces when the composite-index prefix match ({@code key1='a' AND key2='b'}) itself spans MULTIPLE data
+ * pages within one compacted series - i.e. a group large enough that the root page holds a RUN of several
+ * page-boundary entries that all compare equal under a partial-key comparison. {@code compareKey()}'s "PARTIAL
+ * MATCHING" walk resolves an ambiguous binary-search landing point to the FIRST entry of that run for purpose=2 and
+ * the LAST entry for purpose=3; using purpose=2 unconditionally means a DESCENDING scan's root-page probe always
+ * lands on the FIRST (lowest-keyed) data page of the group instead of the LAST (highest-keyed) one, so the scan
+ * starts multiple pages too low and returns the wrong (or incomplete) rows.
+ * <p>
+ * The existing #6692 regression test ({@code Issue6592CompactedCompositePrefixDescTest}) used a target group small
+ * enough to live in a single data page, so it never exercised this specific branch - confirmed by the automated code
+ * review comment that filed this issue.
+ *
+ * @author Luca Garulli (l.garulli@arcadedata.com)
+ */
+@Tag("slow")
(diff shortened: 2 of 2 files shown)


lvca:
Addressed the automated review's substantive point: added `descendingPartialKeyIteratorIncludesAnEntireLowerSeries()`, which builds a second prefix group alphabetically below the target, large enough to compact into its own series, and drains a descending partial-key iterator directly against `LSMTreeIndexCompacted` (bypassing the native `Select` builder) to confirm nothing is dropped when `fromKeys` sorts above an entire lower series (not just one page within it) - the specific `purpose=3` HIGHER-than-last boundary path the review flagged as newly-exercised and untested. Verified this test also fails against the pre-fix code (reverted locally, confirmed red, restored) before committing it, so it's a genuine regression check for that path.

Also tightened the class Javadoc per the review's phrasing nit.

Not applying CodeRabbit's "Docstring Coverage" suggestion: this repo's convention (CLAUDE.md) is to avoid comments/docstrings unless they explain non-obvious WHY, so adding boilerplate docstrings to satisfy a generic coverage threshold would go against house style rather than improve the code.

lvca:
Round 2 review was positive (confirmed the fix and the entire-lower-series edge case coverage correct) with two non-blocking nits:

- `GROUP_SIZE`/`LOWER_GROUP_SIZE` both being `4_000` - kept as two independent constants (target group vs. noise group) since they may be tuned independently later, added a short comment on each to make that explicit.
- Stray double-space misalignment on `PAGE_SIZE` - fixed.

Pushed as `3aeda4435`. All 4 tests in the class still green.

lvca:
Round 3 review was fully positive (independently re-verified the boundary tracing and confirmed the entire-lower-series test exercises exactly the right path), with two non-blocking points:

1. **Checked whether `searchInCompactedIndex()`'s own root-page lookup (~line 715) has the same asymmetry.** It does not, and it's not applicable here:
   - `searchInCompactedIndex()` is reached only through `LSMTreeIndexCompacted.get()` / `LSMTreeIndexMutable.get()`, which `SelectExecutor.java:391` calls **only** when `fullKeyMatch` is true (every composite-index property bound by equality). The code there is explicit about why: *"A KEY SHORTER THAN THE INDEX'S FULL ARITY IS A PREFIX, NOT AN EXACT KEY: get() ... ONLY RETURNS EVERY MATCH WHEN THE KEY'S ARITY MATCHES THE INDEX'S OWN EXACTLY, SO A PREFIX MUST GO THROUGH range() INSTEAD."* A partial-prefix match always goes through `range()`/`iterator()` → `LSMTreeIndexCursor` → `newIterators()` (the method this PR fixes), never through `get()`.
   - Even setting that call-graph fact aside, `lookupInPage()` itself would reject a partial key at `purpose=1` with `IllegalArgumentException("key is composed of N items, while the index defined M items")` - so a partial key reaching `searchInCompactedIndex()` would fail loudly, not silently misbehave like the bug this PR fixes did.
   - `get()`/`searchInCompactedIndex()` also has no `ascendingOrder` parameter at all - it's a pure equality lookup, order-agnostic - so the entire premise of "the root-page probe must follow scan direction" doesn't apply to that method.
   
   No change needed there.

2. **Optional helper to avoid a third copy of the `ascendingOrder ? 2 : 3` ternary if the pattern spreads further** - the reviewer itself flagged this as non-blocking/optional. With only the two existing call sites (`newIterators()` and `searchInCompactedIndex()`'s sibling `searchInCurrentPage()`), a helper would be a premature abstraction per this repo's convention against introducing abstractions before they're needed. Not applying it now; happy to revisit if a third call site appears.

CI is now running the full test-lane sweep (unit/slow-unit/integration/ha-integration/etc.) following `build-and-package` passing. Will report back once that's complete.

lvca:
\`python-e2e-tests\` failed (1/57: \`test_asyncpg.py::test_parameterized_insert\`), but it looks unrelated to this diff:

- The failure is an \`AssertionError: insufficient data in buffer: requested 4 remaining 3\` raised deep inside the \`asyncpg\` client library's own binary protocol frame reader (\`asyncpg/pgproto/frb.pyx\`) - a Postgres wire-protocol byte-framing issue in the Python client, not an ArcadeDB assertion.
- This PR's diff is entirely inside \`com.arcadedb.index.lsm.LSMTreeIndexCompacted\` (a compacted LSM index root-page lookup used for descending composite-index scans) plus its own test file - it has no path to the Postgres wire protocol module (\`postgresw\`) at all.
- Checking recent \`main\`-branch CI runs of the same workflow shows several other lanes (\`ha-integration-tests\`, \`integration-tests\`, \`unit-tests\`) failing intermittently on \`main\` itself over the last few hours, independent of any PR - the CI environment appears to be in a generally flaky state right now, which is consistent with an unrelated transient e2e failure here too.

I tried to re-run just this failed job to confirm, but the API refused since the rest of the workflow run (slow-unit-tests, ha-integration-tests, etc.) is still in progress - will re-run it once the full run completes if it's still marked failed, and report back before considering the PR ready.

lvca:
Full CI investigation of the three lane failures seen so far - none are caused by this PR's diff:

**\`vector-unit-tests\`**: single failure, \`LSMVectorIndexRecoveryTest.vectorNeighborsShouldReturnCorrectCountAfterIncrementalInsertAndRebuild\` - a `ConditionTimeout` waiting up to 11 minutes for a background vector-index rebuild to drain. Entirely in \`com.arcadedb.index.vector\` (a different index type/subsystem from the composite LSM tree index this PR touches), and matches the known vector-rebuild-timing-flake pattern in this codebase (background rebuild convoying on a single JVM-wide semaphore).

**\`unit-tests\`**: two failures, both **pre-existing on \`main\` independent of this PR** - verified by running each in complete isolation (fresh JVM, single test class, nothing from this PR's diff loaded) and watching them fail identically:
- \`Issue5414LastPointTest.descendingScanOnAnEmptySeriesReturnsNothing\` (TimeSeries engine)
- \`SQLFunctionSumAverageMultiArgTest.avgDoesNotTruncate\` (SQL math functions)

Both turned out to be the **same underlying bug**: an aggregate query (\`avg()\`) over a WHERE clause matching zero rows returns one row with a null aggregate instead of an empty result set. Filed as **#6705** with both repros, since it's a real regression somewhere in the codebase, just nowhere near \`com.arcadedb.index.lsm\`.

**\`python-e2e-tests\`**: already analyzed in an earlier comment as an asyncpg client-library wire-protocol decode assertion, unrelated to this diff.

Still waiting on \`ha-integration-tests\` and \`slow-unit-tests\` (the two longest lanes) to finish before considering this PR CI-complete.

lvca:
CI is now fully complete. Final status:

**Passed**: \`ha-integration-tests\` (1h2m), \`slow-unit-tests\` (1h1m), and every other lane (build-and-package, all e2e suites, opencypher-tck-tests, integration-tests, builder-tests, static analysis, 3 rounds of \`claude-review\`, CodeRabbit, Codacy, CodeQL, Meterian).

**Failed, all confirmed unrelated to this diff** (investigated in earlier comments):
- \`python-e2e-tests\`: asyncpg client library wire-protocol decode assertion.
- \`vector-unit-tests\`: known vector-index rebuild timing flake, different index subsystem.
- \`unit-tests\`: pre-existing \`avg()\`-over-empty-group bug, verified deterministic and independent of this PR, filed as #6705.
- \`codecov/project\`: -0.02% project-wide coverage dip. Codecov's own report confirms **"All modified and coverable lines are covered by tests"** for this PR's diff - the dip is explained by the same unrelated failing tests above not exercising their normal code paths during this run, not by anything in this change.

This PR's diff itself is a one-line fix plus a fully-covered regression test, verified across three independent \`claude-review\` rounds (each round's findings addressed or investigated and explained), 20+ related index test classes, and now the full CI matrix. Ready for merge from my side - over to you, lvca.

Issue #6670, opened 2026-08-24
MATCH expand-into GAV fast-path ignores the pattern edge type (and uses the wrong direction on reverse): wrong rows


ruispereira:
### Summary
The MATCH "expand-into" fast path (both pattern endpoints already bound, e.g. a cyclic pattern) asks the Graph Analytical View provider only for **untyped, forward-direction** connectivity — it never passes the pattern's edge type, and the reverse traverser inherits the wrong direction. So a typed pattern edge matches on any edge, producing wrong rows.

### Where (HEAD `86a704526`)
`engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java`, `init()` (~L129-137):
```java
final GraphTraversalProvider provider = GraphTraversalProviderRegistry.findProvider(context.getDatabase());  // no edgeTypes
if (provider != null) {
  final int srcId = provider.getNodeId(sourceVertex.getIdentity());
  final int tgtId = provider.getNodeId(targetElem.getIdentity());
  if (srcId >= 0 && tgtId >= 0) {
    final Vertex.DIRECTION direction = edge != null && !edge.out ? Vertex.DIRECTION.IN : Vertex.DIRECTION.OUT;
    if (provider.isConnectedTo(srcId, tgtId, direction)) { ... match ... }   // NO edgeTypes passed
```
`GraphTraversalProvider.isConnectedTo` accepts an `edgeTypes` varargs but is called with none, so it tests connectivity by any edge type. The normal slow path (`traversePatternEdge` → `item.getMethod()` = `.out('KNOWS')`) honours the type; this fast path bypasses it. `MatchReverseEdgeTraverser` does not override `init()`, so it reuses this block with a forward-derived direction — a reverse traversal probes the wrong direction too.

### Repro (database with a Graph Analytical View registered; both endpoints bound)
`MATCH {as:a}.out('KNOWS'){as:b} RETURN a,b` where `a`→`b` are connected only by a `LIKES` edge.
Expected: no match. Actual: matches (any OUT edge satisfies `isConnectedTo`).

### Suggested fix
Take the fast path only for an untyped pattern edge; otherwise pass the pattern edge type(s) to both `findProvider(db, edgeTypes)` and `isConnectedTo(src, tgt, direction, edgeTypes)`, and derive `direction` from the traverser's actual orientation (override in `MatchReverseEdgeTraverser`), handling `BOTH`.

### Scope / confidence
[...]

Issue #6669, opened 2026-08-24
Cached SELECT statement AST is mutated for a $var target: subsequent reuse reads the wrong type (and a first-execution data race)


ruispereira:
### Summary
`SelectExecutionPlanner.init()` aliases the statement's target **without copying it**, and a later planner rewrite mutates that shared node in place. Because the `Statement` object is cached (StatementCache) and shared across executions, a `SELECT ... FROM $var` whose `$var` resolves to a type name permanently pins the cached statement (and cached plan) to the first execution's type — a subsequent execution with a different `$var` binding reads the wrong type. Also a concurrent-first-execution data race on the shared AST.

### Where (HEAD `86a704526`)
`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java`:
```java
info.target = this.statement.getTarget();                       // L140 — ALIAS (no .copy())
info.whereClause = ...getWhereClause().copy();                  // L141 — every other clause IS copied
...
String typeName = info.target.getItem().getIdentifier().getStringValue();   // L823
if (typeName.startsWith("$")) {
  typeName = (String) context.getVariable(typeName);
  info.target.getItem().setIdentifier(new Identifier(typeName));            // L826 — mutates the SHARED cached AST node
}
```
`SelectStatement` builds the planner on the cached statement (`new SelectExecutionPlanner(this)`, no copy), and a `$var`-only target reports `isCacheable() == true`.

### Repro
In a script / prepared reuse under one SQL text:
1. `LET $t = "Agent"; SELECT FROM $t WHERE ...` → statement `S` cached; L826 overwrites `S`'s FromItem identifier to `"Agent"`; plan cached.
2. Same SQL text with `$t = "Queue"` → the cached plan (bound to Agent) is returned, or the mutated `S` (identifier now literally `"Agent"`, no longer `$`-prefixed) re-plans against Agent.
Expected rows from `Queue`; actual rows from `Agent`. Silent wrong result. Concurrent first-executions also race on `S`'s shared `FromItem` via the unsynchronized `setIdentifier`.

### Suggested fix
Copy the target in `init()` (`info.target = statement.getTarget() == null ? null : statement.getTarget().copy();`) so the `$var` rewrite mutates a per-planning copy; or resolve `$var` into `info` only, never into the shared node; or make a `$var` target non-cacheable in `FromItem.isCacheable()`.

### Scope / confidence
[...]

PR #6706, opened 2026-08-24
fix(#6670,#6669): correct MATCH GAV expand-into fast path and SELECT $var target caching

lvca:
## Summary

Fixes two small, independent bugs found during other analysis:

**#6670** - `MatchEdgeTraverser`'s GAV expand-into fast path (both pattern endpoints already bound, e.g. a cycle-closing edge) asked the `GraphTraversalProvider` with no edge type at all (so any edge type satisfied it) and derived direction from the schedule's forward/reverse split instead of the pattern's actual `out`/`in`/`both` method - so a typed or `.in(...)`-directed pattern edge could match on the wrong type or direction. Fix:
- Restrict the fast path to plain `out()`/`in()`/`both()` hops (an `outE()`/`inE()`/`bothE()`/`outV()`/`inV()` method, or a `FieldMatchPathItem`/`MultiMatchPathItem` with no method at all, now falls through to the always-correct slow traversal).
- Pass the pattern's edge type(s) through to `GraphTraversalProviderRegistry.findProvider()` and `GraphTraversalProvider.isConnectedTo()` (reusing `MatchExecutionPlanner.extractEdgeLabels`, now package-visible, instead of duplicating the label-extraction logic).
- Derive direction from the parsed method name via a new `getExpandIntoDirection()`, overridden in `MatchReverseEdgeTraverser` to flip it (that class runs the method's *reverse* from the opposite endpoint, the same out&harr;in flip `MethodCall.executeReverse` applies).
- Also skip the fast path for a `while`/`maxDepth` pattern item (`isSingleHopExpandable()`): `isConnectedTo()` only ever answers a single-hop check, so a variable-depth item must keep going through the recursive `executeTraversal()`. This gap predates this PR - the original fast path had no such guard at all - and isn't covered by the issue's own repro, but is the same class of bug and is covered by dedicated regression tests.

**MatchGAVFusedStep sibling bug (found during review)** - the multi-hop chain-fusion GAV step has the exact same direction-derivation bug as the single-hop expand-into path above (`et.out ? OUT : IN`, ignoring the pattern's actual method), and `isFusibleEdge()` explicitly allows `in()`/`both()` hops into a fused chain. Fixed by extracting the direction mapping into a shared `MatchExecutionPlanner.directionFor(item, forward)`, now used by all three call sites (`MatchEdgeTraverser`, `MatchReverseEdgeTraverser`, `MatchGAVFusedStep`) instead of duplicating it.

**#6669** - `SelectExecutionPlanner.init()` aliased the statement's FROM target (`info.target = this.statement.getTarget()`) instead of copying it like every other clause. Two later planner steps resolve a `$var` target against the current `CommandContext` and rewrite the identifier in place - on the *shared* node, since it was never copied: `calculateTargetBuckets` (unconditional) and `rewriteIndexChainsAsSubqueries` (only when a WHERE clause is present). That node is exactly what `StatementCache` hands back for every execution of the same SQL text, so the first execution's resolved type stuck permanently to the cached statement (plus a data race on concurrent first executions). Fix:
- Copy the target in `init()`, matching every sibling clause (`whereClause`, `letClause`, `groupBy`, `orderBy`, `unwind`).
- Also mark a `$var` target as non-cacheable in `FromItem.isCacheable()`: `ExecutionPlanCache` is keyed purely by SQL text with no awareness of variable bindings, so even with the AST no longer corrupted, a `$var`-target plan built for the first binding would otherwise still get replayed for every later execution regardless of what `$var` is bound to then. Both fixes are needed for full correctness - suggesting this as the better alternative to the issue's single-fix options.

## Test plan

- [x] `MatchEdgeTraverserExpandIntoTest` (new): focused unit tests driving `MatchEdgeTraverser`/`MatchReverseEdgeTraverser` directly against a recording `GraphTraversalProvider` test double, asserting the exact direction/edge-types passed to `isConnectedTo()`. Verified to fail against the pre-fix code (4/5 assertions), pass after. A full-SQL-MATCH repro proved unreliable: the fast path only fires for a cycle-closing edge, and getting the planner to schedule one without folding it into an unrelated `MatchGAVFusedStep` chain depends on scheduling heuristics outside this bug's scope.
- [x] `MatchGAVFusedStepDirectionTest` (new): SQL-level regression for the fused-chain sibling bug, using a real registered GAV and a 3-hop pattern (fusion only kicks in from the schedule's second edge onward). Verified to fail against the pre-fix code (zero rows instead of the correct single row), pass after.
- [x] `SelectVarTargetCacheTest` (new): reuses the same parsed `Statement` instance (as `StatementCache` would) across two `createExecutionPlan()` calls with different `$t` bindings, asserting each returns data from its own bound type. Verified to fail against the pre-fix code, pass after.
- [x] Existing `MatchGAVExpandIntoTest` (all 3 tests) still green.
- [x] Full `com.arcadedb.query.sql.executor.**` and `com.arcadedb.query.sql.parser.**` test packages green - no regressions.

## Summary by CodeRabbit

- **Performance**
  - Improved MATCH query traversal for eligible single-hop directional graph expansions while preserving recursive behavior for variable-depth patterns.
  - Improved forward, reverse, and bidirectional edge traversal, including edge-label handling.

- **Bug Fixes**
  - Prevented cached query plans from reusing variable-based targets with incorrect bindings.
  - Preserved correct behavior when shared query statements run with different runtime targets.

- **Tests**
  - Added coverage for edge types, traversal directions, disconnected endpoints, and variable-depth matches.
  - Added regression coverage for cached variable-based `SELECT` targets.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProviderRegistry.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProviderRegistry.java
@@ -101,6 +101,17 @@ public class GraphTraversalProviderRegistry {
   }
 
   /**
+   * Returns true if any provider is registered for any database, without the lock/unwrap/WeakHashMap lookup
+   * {@link #findProvider} needs once it establishes that. Lets a caller that would otherwise do non-trivial work
+   * (e.g. building an edge-types array) just to pass it to {@link #findProvider} skip that work first in the
+   * overwhelmingly common case where no provider is registered at all - the same single-volatile-read short
+   * circuit {@link #findProvider} itself takes, exposed for callers that need it before they can call that method.
+   */
+  public static boolean hasAnyProviders() {
+    return hasAnyProviders;
+  }
+
+  /**
    * Finds the first ready provider that covers all the given edge types.
    *
    * @param database  the database
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
@@ -108,9 +108,9 @@ public class MatchEdgeTraverser {
   }
 
   /**
-   * Returns {@code false} for a while/maxDepth pattern item: {@link com.arcadedb.graph.GraphTraversalProvider
-   * #isConnectedTo} only ever answers whether two vertices are joined by a single hop, so a variable-depth item
-   * must keep going through the recursive {@link #executeTraversal} instead of taking the expand-into fast path.
+   * Returns {@code false} for a while/maxDepth pattern item: {@link GraphTraversalProvider#isConnectedTo} only
+   * ever answers whether two vertices are joined by a single hop, so a variable-depth item must keep going
+   * through the recursive {@link #executeTraversal} instead of taking the expand-into fast path.
    */
   private static boolean isSingleHopExpandable(final MatchPathItem item) {
     if (item == null)
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
@@ -166,9 +166,7 @@ public class MatchEdgeTraverser {
             ? getExpandIntoDirection() : null;
         if (direction != null) {
           final String[] edgeTypes = MatchExecutionPlanner.extractEdgeLabels(edge);
-          final GraphTraversalProvider provider = edgeTypes.length == 0
-              ? GraphTraversalProviderRegistry.findProvider(context.getDatabase())
-              : GraphTraversalProviderRegistry.findProvider(context.getDatabase(), edgeTypes);
+          final GraphTraversalProvider provider = GraphTraversalProviderRegistry.findProvider(context.getDatabase(), edgeTypes);
           if (provider != null) {
             final int srcId = provider.getNodeId(sourceVertex.getIdentity());
             final int tgtId = provider.getNodeId(targetElem.getIdentity());
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/MatchEdgeTraverser.java
@@ -107,6 +107,26 @@ public class MatchEdgeTraverser {
     return this.edge.edge.in.alias;
   }
 
+  /**
+   * Returns the {@link Vertex.DIRECTION} the GAV expand-into fast path (see {@link #init}) should query, as seen
+   * from this traverser's own starting point (i.e. the vertex {@link #getStartingPointAlias()} names): the same
+   * direction {@link #traversePatternEdge} would use by invoking {@code item.getMethod().execute(...)}. Returns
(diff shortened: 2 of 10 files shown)


lvca:
Addressed both CodeRabbit findings:
- `MatchEdgeTraverser`: added an `isSingleHopExpandable()` guard so the expand-into fast path is skipped for while/maxDepth pattern items (variable-depth traversal), falling through to the recursive `executeTraversal()` as before. Added regression tests (`expandIntoSkipsWhileConditionItems`, `expandIntoSkipsMaxDepthItems`).
- `SelectVarTargetCacheTest`: fixed the misleading comment (the third execution reuses the parsed `Statement`, not the `ExecutionPlanCache` - a $var target is non-cacheable) and added a companion `ordinaryTargetStillUsesExecutionPlanCache` test proving plan caching still works for a literal-type target.

lvca:
Investigated the CI failures on this run - all confirmed unrelated to this PR's changes:

- **python-e2e-tests**: `test_asyncpg.py::test_parameterized_insert` - pre-existing, documented "known issue" in the ArcadeDB PostgreSQL wire protocol (unrelated `postgresw` module).
- **integration-tests**: `Bolt5002RoutingTableIT.neo4jSchemeRoutesReadsAndWrites` - a Neo4j-Bolt/Raft HA cluster test (`DatabaseAreNotIdentical` after leader election/snapshot bootstrap), unrelated to SQL executor/parser code.
- **unit-tests**: `Issue5414LastPointTest.descendingScanOnAnEmptySeriesReturnsNothing` and `SQLFunctionSumAverageMultiArgTest.avgDoesNotTruncate` - verified these **also fail identically on a clean `origin/main` checkout** (tested in an isolated worktree, no code from this PR present), confirming they're pre-existing bugs in the TimeSeries engine and SQL math functions, not caused by this PR. Neither touches MATCH/SELECT/GAV code.

None of these overlap with the files this PR touches (`MatchEdgeTraverser`, `MatchReverseEdgeTraverser`, `MatchExecutionPlanner`, `SelectExecutionPlanner`, `FromItem`, `GraphTraversalProviderRegistry`). Still waiting on `slow-unit-tests` and `ha-integration-tests` to finish.

lvca:
Fixed the `MatchGAVFusedStep` sibling bug flagged in the last review pass: it derived each fused hop's direction from `EdgeTraversal.out` (the schedule's forward/reverse flag) instead of the pattern's actual `out`/`in`/`both` method - the exact root cause this PR already fixes for the single-hop expand-into fast path, just in the multi-hop chain-fusion path. Since `isFusibleEdge()` explicitly allows `in()`/`both()` hops into a fused chain, a query with a registered GAV and one of those in a fusible chain would silently return wrong (or zero) rows.

Extracted the direction-mapping logic (previously duplicated between `MatchEdgeTraverser`/`MatchReverseEdgeTraverser`, flagged as an optional nit earlier) into a shared `MatchExecutionPlanner.directionFor(item, forward)`, now used by all three call sites - this was the point where the earlier "only worth doing if this area gets touched again" caveat flipped, since it now has a third real caller.

New `MatchGAVFusedStepDirectionTest` (2 tests) verified to fail against the pre-fix code and pass after. Also ran the full `com.arcadedb.query.sql.executor.**`/`com.arcadedb.query.sql.parser.**` suites again - no regressions.

PR #6708, opened 2026-08-24
fix(#6608): reject non-STRING arguments in remaining STRING positions of string functions

### LINKED ISSUE #6608: String functions still coerce invalid non-STRING arguments instead of reporting type errors
### ArcadeDB version

ArcadeDB `26.9.1-SNAPSHOT`

### Environment

- Linux x86_64, Docker, OpenJDK
- Freshly built from the latest main branch (2026-08-23)
- HTTP command API, Cypher language
- Also reproduces on the released `26.8.1` build

### Description

Issue #5798 (fixed by PR #5901) made several string functions reject non-STRING
first arguments with a client-facing type error (HTTP 400).  On the current
build, that fix works for the first argument position: `split(5, ',')` and
`tolower([])` now return HTTP 400.

However, the remaining STRING-typed argument positions of the same functions do
not follow the same explicit type-validation behavior introduced by #5798 for
other STRING arguments.  They still accept invalid values and implicitly convert
them to their string representation.  `char.length()` is not covered by the fix
at all.

### Relationship to #5798

#5798 fixed type validation for the first STRING argument positions of several
string functions.

This issue reports remaining STRING-typed argument positions that still use
implicit coercion, plus `char.length()` which is not covered by that fix.

### Steps to reproduce

No graph data is required.  Run each query separately.

Primary reproducer:

```cypher
RETURN replace('a', 1, []) AS r;
```

Observed:

```text
HTTP 200, r = "a"
```

Coverage probes:

```cypher
RETURN split('a', []) AS r;
```

Observed:

```text
HTTP 200, r = ["a"]
```

```cypher
RETURN ltrim('a', []) AS r;
```

Observed:

```text
HTTP 200, r = "a"
```

```cypher
RETURN char.length([]) AS r;
```

Observed:

```text
HTTP 200, r = 2
```

### Expected behavior

The affected argument positions should follow the same explicit type-validation
behavior introduced by #5798 for other STRING arguments.  In particular, these
queries should return a client-facing type error (HTTP 400):

```cypher
RETURN replace('a', 1, []) AS r;
-- HTTP 400 (non-STRING arguments should be rejected consistently with #5798)

RETURN split('a', []) AS r;
-- HTTP 400 (LIST is not a valid STRING delimiter)

RETURN ltrim('a', []) AS r;
-- HTTP 400 (LIST is not a valid STRING chars argument)

RETURN char.length([]) AS r;
-- HTTP 400 (LIST is not a valid STRING argument)
```

For comparison, the same build already returns HTTP 400 for the first-argument
positions fixed by #5798, such as `split(5, ',')` and `tolower([])`, and for the
trim argument covered by that fix, `btrim('a', 1, [])`.

Callers that intentionally require conversion can use `toString()` explicitly:

```cypher
RETURN split('a', toString([])) AS r;
```

### Actual behavior

A LIST (or INTEGER) passed to a secondary STRING argument position is implicitly
converted to its string representation and the query succeeds with a non-null
result, masking invalid type usage:

- `split('a', [])` returns `["a"]` (the LIST delimiter `[]` becomes the string `"[]"`)
- `ltrim('a', [])` returns `"a"`
- `replace('a', 1, [])` returns `"a"`
- `char.length([])` returns `2` (the LIST becomes the string `"[]"`, length 2)

### Cross-engine comparison

This is supplementary evidence; the primary oracle is the same-build
type-validation policy established by #5798.  In particular, Neo4j also coerces
the `split` delimiter.

| Query | Neo4j 2026.06 | Memgraph 3.12 | FalkorDB 4.20.1 | ArcadeDB 26.9.1-SNAPSHOT |
| `split('a', [])` | SUCCESS (coerces) | error | error | **BUG: 200, coerces** |
| `ltrim('a', [])` | error | error (arity) | error (arity) | **BUG: 200, coerces** |
| `replace('a', 1, [])` | error | error | error | **BUG: 200, coerces** |
| `char.length([])` | no such function | no such function | no such function | **BUG: 200, coerces** |

### Isolated recheck

Freshly built instance from latest main, every reproducer run 3 times
(2026-08-23).  Reproduction rate 12/12, 0 transport failures, all normal-input
controls pass.  The internal controls above (same build, fixed first-argument
positions) all return HTTP 400 as expected.

### Why this looks like a bug

The same engine provides an internal consistency control.  On the exact same build, `split(5, ',')`
(argument 1) is rejected with HTTP 400 while `split('a', [])` (argument 2) is
silently accepted; `replace(5, 'a', 'b')` is rejected while
`replace('a', 1, [])` is accepted.  The type-validation added by #5798 does not
cover the remaining STRING argument positions of the same functions, and
`char.length()` is not covered at all.

ArcadeDB's string-function type validation is incomplete after #5798: some
STRING-typed arguments still accept non-STRING values through implicit coercion,
while equivalent positions reject correctly.

### Impact

Invalid type usage is masked as a successful result.  Applications that rely
on type errors to catch bugs will not see them for these argument positions,
and result values are produced from implicit conversions the query author did
not request.

lvca:
## Summary

Issue #5798 (fixed by PR #5901) made `split()`, `replace()`, `trim()`/`btrim()`, `lTrim()` and `rTrim()` reject a non-STRING value in their **primary** argument position with a client-facing type error (HTTP 400). Fixes #6608, which reports that the **remaining** STRING-typed argument positions of the same functions still silently coerced any value via `toString()`:

- `split()`'s delimiter (2nd argument)
- `replace()`'s search and replacement text (2nd and 3rd arguments)
- `trim()`/`btrim()`/`lTrim()`/`rTrim()`'s trim-character argument
- `char.length()` / `char_length()` - not covered by #5798 at all, any argument position

## Fix

Apply the existing `CypherFunctionHelper.requireStringArgument()` helper (introduced by #5901) to the remaining argument positions, so every STRING-typed argument position of these functions now follows the same policy the primary position already has. No new abstraction was needed - this reuses the exact mechanism #5798 established.

`split()`'s delimiter is a deliberate divergence from Neo4j (which coerces it): on this same build `split(5, ',')` already rejects a non-STRING primary argument, so accepting a non-STRING delimiter in the same function would be an internal inconsistency rather than a compatibility concession. Every other affected position already matches Neo4j's own rejection.

## Test plan

- [x] Added `CypherStringFunctionSecondaryArgumentIssue6608Test` reproducing every case from the issue (fails against the pre-fix code, passes after)
- [x] `CypherStringFunctionArgumentIssue5798Test` (the original fix's regression test) still passes
- [x] `CypherLeftRightSubstringFunctionArgumentIssue6609Test`, `OpenCypherStringFunctionsComprehensiveTest`, and the SQL-side trim/split method tests still pass
- [x] Full `com.arcadedb.query.opencypher.**` package run clean except one pre-existing flake (`CypherGAVBoundTargetCardinalityTest`, unrelated to string functions, passes in isolation)

## Summary by CodeRabbit

- **Bug Fixes**
  - String functions now require string-typed arguments instead of implicitly converting other values.
  - Improved validation for `char.length`, `replace`, `split`, `lTrim`, `rTrim`, and `trim`.
  - Invalid argument types now produce clear type errors, including when other arguments are null.
  - Null handling remains unchanged for valid inputs.

- **Tests**
  - Added regression coverage for secondary-argument validation, null propagation, aliases, and trimming variants.

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/text/CharLengthFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/text/CharLengthFunction.java
@@ -19,6 +19,7 @@
 package com.arcadedb.function.text;
 
 import com.arcadedb.function.StatelessFunction;
+import com.arcadedb.function.cypher.CypherFunctionHelper;
 import com.arcadedb.query.sql.executor.CommandContext;
 
 /**
@@ -45,8 +46,11 @@ public class CharLengthFunction implements StatelessFunction {
   @Override
   public Object execute(final Object[] args, final CommandContext context) {
     checkArity(args);
-    if (args[0] == null)
+    // STRING-only argument, same declared input domain as toUpper()/toLower()/trim() and the rest of the
+    // #5798 family - this function predates that fix and was never brought in line with it (issue #6608).
+    final String source = CypherFunctionHelper.requireStringArgument(args[0], getName());
+    if (source == null)
       return null;
-    return (long) args[0].toString().length();
+    return (long) source.length();
   }
 }
--- a/engine/src/main/java/com/arcadedb/function/text/LTrimFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/text/LTrimFunction.java
@@ -52,13 +52,15 @@ public class LTrimFunction implements StatelessFunction {
       return source.stripLeading();
     }
     if (args.length == 2) {
-      // The primary argument is type-checked before a null trim character decides the answer, so an
-      // out-of-domain primary argument is still reported even when args[1] happens to be null (issue
-      // #5798 review: lTrim(5, null) must still be a type error, not a silent null).
+      // Both arguments are STRING-typed and type-checked before either being null decides the answer, so an
+      // out-of-domain argument is still reported regardless of which position happens to be null (issue
+      // #5798 review: lTrim(5, null) must still be a type error, not a silent null). The trim-character
+      // argument is STRING-typed too, same as the primary one - issue #6608: #5798 only covered this
+      // function's primary argument position, leaving the trim character to silently coerce via toString().
       final String source = CypherFunctionHelper.requireStringArgument(args[0], getName());
-      if (source == null || args[1] == null)
+      final String trimChar = CypherFunctionHelper.requireStringArgument(args[1], getName());
+      if (source == null || trimChar == null)
         return null;
-      final String trimChar = args[1].toString();
       if (trimChar.isEmpty())
         return source.stripLeading();
       return stripLeading(source, trimChar);
--- a/engine/src/main/java/com/arcadedb/function/text/RTrimFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/text/RTrimFunction.java
@@ -52,13 +52,15 @@ public class RTrimFunction implements StatelessFunction {
       return source.stripTrailing();
     }
     if (args.length == 2) {
-      // The primary argument is type-checked before a null trim character decides the answer, so an
-      // out-of-domain primary argument is still reported even when args[1] happens to be null (issue
-      // #5798 review: rTrim(5, null) must still be a type error, not a silent null).
+      // Both arguments are STRING-typed and type-checked before either being null decides the answer, so an
+      // out-of-domain argument is still reported regardless of which position happens to be null (issue
+      // #5798 review: rTrim(5, null) must still be a type error, not a silent null). The trim-character
+      // argument is STRING-typed too, same as the primary one - issue #6608: #5798 only covered this
+      // function's primary argument position, leaving the trim character to silently coerce via toString().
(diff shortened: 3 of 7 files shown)


lvca:
Addressed the two review points:

1. **split() delimiter divergence from Neo4j**: already explained in the PR description and in a code comment in `CypherSplitFunction`. The project has no CHANGELOG file to update (confirmed by precedent in #5901/`docs/5798-string-function-strict-string-args.md`, which checked the same thing and found none), so the PR description is the right place for it.
2. **btrim() 3-argument positional-call gap**: added `btrimCalledWithThreePositionalArgumentsIsAPreExistingAmbiguityNotIntroducedByThisFix()`, pinning the existing (pre-#6608) behavior with an explanation of why `args[0]` is read as the trim mode rather than the source. Left unfixed since it's a separate, unscoped question (whether `btrim()` should accept 3 positional arguments at all) - happy to file a follow-up issue if that's wanted.

PR #6711, opened 2026-08-24
fix(#6657): expose close-time deferred graph rebuild via getStats()

### LINKED ISSUE #6657: LSMVectorIndex: expose deferred-rebuild-pending state via getStats()
Follow-up suggestion from #6653's third code-review round (cc @lvca).

#6653 makes `close()` defer a large index's graph rebuild to the next search instead of blocking on it, logging the deferral at `FINE` level. The reviewer noted that an ops team debugging "why is my first query after reopen slow" would currently have to find that `FINE` log line rather than being able to check it directly - suggested exposing something like a "graph rebuild pending" boolean or a "last deferred at" timestamp via `LSMVectorIndex.getStats()`, alongside the existing `graphState`/`graphRebuildCount`/`mutationsSinceRebuild` entries there.

Explicitly optional/non-blocking for #6653 itself - this is a small observability addition, not a correctness fix, so it doesn't belong bundled into that PR. Filing it so it isn't lost.

## Possible direction (not investigated)

`getStats()` already reports `graphState` (ordinal 0/1/2) and `mutationsSinceRebuild`; a caller can already infer "there is pending work" from `graphState == MUTABLE && mutationsSinceRebuild > 0`, or "never built" from `graphState == LOADING`. What's missing is specifically *whether the last close deferred a build that would otherwise have run* - i.e., distinguishing "this index has pending mutations because it's mid-session" from "this index has pending mutations because close() chose not to pay for them." A minimal addition would be a boolean or counter set only in `flush()`'s new deferral branch (and cleared once `ensureGraphAvailable()`/`buildGraphFromScratch()` actually catches up), rather than a general-purpose timestamp - keeps the addition scoped to the exact question ops teams would ask.

cc #6067, #6653

robfrank:
## What does this PR do?

`LSMVectorIndex.getStats()` gains a `closeTimeRebuildPending` entry (`1`/`0`) that answers whether
the most recent `close()` deferred a graph rebuild it would otherwise have run, instead of an
operator having to go find the `FINE` log line that says the same thing.

## Stacked on #6653 - read this first

**This PR targets `fix/6067-flush-redoes-cancelled-build` (#6653), not `main`.** The deferral branch
this issue is about - `flush()`'s size-threshold skip, added by #6653 - does not exist on `main` yet;
#6653 is still open. #6657 was explicitly filed as a non-blocking follow-up to #6653 ("doesn't belong
bundled into that PR"), so this branches from #6653's branch to build the follow-up on the code it
actually follows up on, rather than duplicating #6653's still-unmerged changes here.

**Diff note:** because of the stacking, this PR's diff will show #6653's changes too until #6653
merges. Once it does, GitHub retargets this PR to `main` automatically (base-branch deletion
retargeting) and the diff narrows to just this PR's own commit. Nothing to do on this PR's part.

## Motivation

Issue #6657, filed by @lvca from #6653's third review round: "an ops team debugging 'why is my first
query after reopen slow' would currently have to find that `FINE` log line rather than being able to
check it directly" - suggested exposing a "graph rebuild pending" boolean or timestamp via
`getStats()`, alongside the existing `graphState`/`graphRebuildCount`/`mutationsSinceRebuild` entries.

## Design decision worth a second look: persisted, not in-memory

The issue's own "possible direction" (explicitly "not investigated") proposes a plain in-memory
boolean, "set only in `flush()`'s new deferral branch (and cleared once
`ensureGraphAvailable()`/`buildGraphFromScratch()` actually catches up)".

That does not actually work as described: `flush()` only ever runs from
`LocalDatabase.closeDurableParts()`, i.e. only at `close()`/`drop()`, and a **fresh** `LSMVectorIndex`
instance is constructed on every `open()`/`create()` (see the two factory methods). A plain field set
by one instance's `flush()` would already be gone by the time any later instance - the one from the
next `open()` - could read or clear it. That's precisely the scenario the issue names as the point of
the feature: checking the stat right after reopen, before the first search. An in-memory field would
misreport `false` at exactly that moment.

Implemented instead as a small persisted flag on the graph's existing manifest sidecar
(`LSMVectorIndexGraphManifest` - a few dozen bytes, already read once per graph load / written once
per persist):

- `Content` gained a `closeDeferredRebuild` boolean, defaulting to `false` on read (old manifests
  without the field read as `false`, no format-version bump needed).
- New `markCloseDeferred()`: preserves whatever `vectorCount`/`fingerprint` are already on disk (or
  falls back to `UNUSABLE_VECTOR_COUNT`/`0` when nothing has ever been persisted - the never-built
  arm), writes `closeDeferredRebuild = true`. It does **not** touch the existing
  staleness-detection fields, so it cannot interfere with `ensureGraphAvailable()`'s fingerprint-based
  reuse check.
- The two existing write paths - `write(int, long)` (normal post-build persist) and
  `markUnusable(String)` - now clear `closeDeferredRebuild` back to `false` explicitly: reaching
  either means a build completed (successfully or not), so whatever a previous close deferred has now
  been paid for.
- `LSMVectorIndex.flush()` calls the new `markCloseTimeRebuildDeferred(gf)` helper from **both**
  existing "skip the build" branches - the size-threshold deferral (#6653) and the pre-existing
  `!valid` "background resources already released" branch - but not from the "nothing needed" branch,
  which must never clear an already-true flag just because a later, uneventful close found nothing new
  pending.
- `getStats()` reads the manifest fresh each call (guarded by `graphFile != null`) and reports
  `closeTimeRebuildPending`.

**Known, deliberate gaps** (both documented at their call sites, neither chased further):
- A large index closed before its first-ever graph file was even created, and closed *again* without a
  build in between (two-or-more-session never-built case), has no manifest path to write to
  (`graphFile == null`) - `markCloseTimeRebuildDeferred()` is a no-op there. The index itself is
  unaffected (still lazily rebuilt on next search); only the stats signal for that narrow window is
  unavailable.
- `releaseBackgroundResources()`'s post-cancellation recheck (added addressing review round 2) can
  still lose to a **straggler build thread**: one that outlived `shutdownNow()`'s 5s budget - already
  logged as a WARNING a few lines above the recheck - and happens to still hold `graphBuildLock` at the
  exact moment the recheck's own `tryLock()` runs. Narrow (needs both an unresponsive build *and* it
  specifically being the lock holder right then), and unlike the cancellation case the recheck exists
  for, there is no later "the build is definitely done by now" moment left inside this method to retry
  from. Flagged by review round 3.

Both kept out of scope deliberately, per the issue's own framing ("keeps the addition scoped to the
exact question ops teams would ask").

## Related issues

Closes #6657
Depends on #6653 (base branch)
Relates to #6067

## Additional Notes

- Not touched: the manifest's `vectorCount`/`fingerprint` staleness-detection logic, which several
  other tests (`Issue6106StaleVectorGraphTest`, `LSMVectorIndexGraphManifestTest`,
  `Issue6503GraphPersistOutOfMemoryTest`) exercise directly - all still green.

## Test plan

- New regression test `Issue6657CloseTimeRebuildPendingStatTest` (`@Tag("slow")`, ~1000 vectors,
  small dimensions/build params for speed). In one flow: the stat turns on the moment `close()`
  defers; it still reads `1` on a **freshly reopened, different** `LSMVectorIndex` instance, before
  that instance has searched (the assertion that specifically rules out the in-memory design); it
  clears back to `0` once the deferred rebuild actually runs via the triggering search; a clean build
  with nothing deferred reports `0` (no false positive).
- Confirmed the test can fail: temporarily disabled both `markCloseTimeRebuildDeferred(gf)` call
  sites and reran - failed at the expected assertion. Restored before committing.
- `mvn -pl engine -am test -Dtest=Issue6657CloseTimeRebuildPendingStatTest` - 1/1 pass.
- `mvn -pl engine -am test -Dtest=Issue6657CloseTimeRebuildPendingStatTest,Issue6067DeferredCloseRebuildTest,LSMVectorIndexGraphManifestTest,Issue6503GraphPersistOutOfMemoryTest`
  - 12/12 pass (the two suites that reference `LSMVectorIndexGraphManifest.Content` directly, plus
  #6067's own tests, plus the OOM-simulation test exercising `markUnusable`).
- Broader sweep of close/rebuild/manifest-adjacent vector-index tests (`VectorIndexRebuildsOnEveryCloseTest`,
  `VectorIndexNoRebuildOnUnusedCloseTest`, `LSMVectorIndexCloseCancelsInFlightBuildTest`,
  `Issue6518CloseDoesNotLeakBuildPoolTest`, `CompactIndexGraphPersistFailureTest`,
  `CompactIndexKeepsGraphReusableTest`, `Issue6106StaleVectorGraphTest`,
  `LSMVectorIndexGraphFileToctouTest`, `LSMVectorIndexGraphFileVisibilityTest`,
  `LSMVectorIndexGraphStorageTest`, `LSMVectorIndexPersistenceTest`, `LSMVectorIndexRecoveryTest`,
  `LSMVectorIndexRebuildTest`, `Issue5870VectorIndexIdRenumberingTest`) - 68/68 pass.

**Review round 2** (concurrency + missing branch coverage):
- New `Issue6657InvalidPathRebuildPendingStatTest` covers the second `markCloseTimeRebuildDeferred()`
  call site (the `needsBuild && !valid` branch).
- New `Issue6657CancelledBuildRecheckTest` deterministically pins the cancelled-build false-negative
  the round-2 review found: a helper thread holds `graphBuildLock` directly (rather than racing a real
  JVector build's timing, which `LSMVectorIndexCloseCancelsInFlightBuildTest` already shows is hard to
  pin reliably) and releases it only after `flush()` has observed it contended, mirroring what a real
  cancellation would do without ever writing the manifest.
- Confirmed both new tests can fail: reverted each fix in turn, reran, confirmed the expected failure,
  restored.
- Full sweep re-run after both fixes (82 tests, including the two new ones) - all pass.

**Review round 3** (documentation-only ask, no blocking issues): documented the residual
straggler-build-thread gap (see "Known, deliberate gaps" above) at its code site and here; added a
cheap already-`true` short-circuit to `markCloseDeferred()` per the minor I/O nit. 30-test focused
re-run after - all pass.

**Review round 4** (real correctness finding, addressed): the `releaseBackgroundResources()` recheck
fired on `needsGraphBuild()` alone, which is `true` whenever `graphState == MUTABLE` for *any* reason -
not only "flush() skipped an owed rebuild". `flush()`'s *other* branch, the one that actually attempts
a synchronous build (below `ASYNC_REBUILD_MIN_GRAPH_SIZE`), can also leave `graphState` at `MUTABLE` -
a failed persist (which correctly writes `closeDeferredRebuild = false`) or new mutations arriving
mid-build both do - and the ungated recheck would overwrite that correct `false` with a misleading
`true`. New `flushDeferredRebuild` field, set only inside `flush()`'s two skip branches and reset at
the top of every `flush()`, gates the recheck so it fires only when *this* close's `flush()` itself
chose to defer; as a side effect this also stops the recheck from running on `LocalDatabase.kill()`/
whole-database-drop, which call `releaseBackgroundResources()` without ever calling `flush()` (round
4's second finding). New `Issue6657RecheckDoesNotMislabelUnattemptedFlushTest` pins the gate directly;
confirmed it fails against the ungated code. Also fixed the `@param` indentation nit. 31-test focused
re-run after - all pass.

**This PR has reached its configured review-cycle limit (4).** Round 4's fixes above are pushed but
have not themselves been through a fifth automated review pass. Given the pattern across all four
rounds - real, specific, verifiable findings each time, all addressed with a matching regression test
that was confirmed to fail pre-fix - a human pass over this last commit specifically (the
`flushDeferredRebuild` gating) is the recommended next step before merge, rather than assuming a clean
bill of health by default.

## Checklist

- [x] I have run the build using `mvn clean package` command (module-scoped: `mvn -pl engine -am test`, per project convention for a change scoped to one module)

### DIFF
--- a/bolt/conformance/COMPATIBILITY.md
+++ b/bolt/conformance/COMPATIBILITY.md
@@ -4,7 +4,7 @@
 
 Certification status of ArcadeDB's Bolt protocol against every official Neo4j driver, per the shared conformance spec ([`spec.yaml`](spec.yaml), epic #4882). Columns are driver language by pinned version ([`driver-versions.md`](driver-versions.md)).
 
-**Last verified:** 2026-08-23 03:49 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/32615904300))
+**Last verified:** 2026-08-24 03:51 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/32687391450))
 
 Legend: ✅ pass, ❌ fail, ⚠️ expected-fail / known limitation, ➖ not applicable, ⚪ skipped, `·` not reported. A `·` in a listed Coverage-gaps column means no result for that driver:version.
 
--- a/bolt/conformance/COMPATIBILITY.md
+++ b/bolt/conformance/COMPATIBILITY.md
@@ -4,7 +4,7 @@
 
 Certification status of ArcadeDB's Bolt protocol against every official Neo4j driver, per the shared conformance spec ([`spec.yaml`](spec.yaml), epic #4882). Columns are driver language by pinned version ([`driver-versions.md`](driver-versions.md)).
 
-**Last verified:** 2026-08-24 03:51 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/32687391450))
+**Last verified:** 2026-08-25 03:48 UTC ([run](https://github.com/ArcadeData/arcadedb/actions/runs/32805890160))
 
 Legend: ✅ pass, ❌ fail, ⚠️ expected-fail / known limitation, ➖ not applicable, ⚪ skipped, `·` not reported. A `·` in a listed Coverage-gaps column means no result for that driver:version.
 
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -45,6 +45,7 @@ import com.arcadedb.schema.DocumentType;
 import com.arcadedb.utility.AnsiCode;
 import com.arcadedb.utility.FileUtils;
 import com.arcadedb.utility.RecordTableFormatter;
+import com.arcadedb.utility.StringUtils;
 import com.arcadedb.utility.TableFormatter;
 import org.jline.reader.Completer;
 import org.jline.reader.EndOfFileException;
@@ -1188,17 +1189,13 @@ public class Console {
   }
 
   /**
-   * Splits a `&lt;key&gt;=&lt;value&gt;` pair on the FIRST '=' only, so a value that contains further separators (a connection
-   * string, a base64 padding, a date pattern) is kept whole and an empty value is a value, not a missing one. Returns null when
-   * there is no separator at all, leaving to the caller the choice between rejecting the input and reading it as an empty value.
-   * Shared by the `-D&lt;key&gt;=&lt;value&gt;` arguments (issue #5928) and by the SET command (issue #6392), which used to
-   * disagree on the rule.
+   * Splits a `&lt;key&gt;=&lt;value&gt;` pair on the FIRST '=' only. Shared by the `-D&lt;key&gt;=&lt;value&gt;`
+   * arguments (issue #5928) and by the SET command (issue #6392), which used to disagree on the rule; delegates to
+   * {@link StringUtils#splitKeyValue} so the Postgres wire's SET command and the Studio include directive
+   * (issue #6423) follow the same rule instead of each carrying their own truncating copy.
    */
   static String[] splitKeyValue(final String pair) {
-    final int separatorPos = pair.indexOf('=');
-    if (separatorPos < 0)
-      return null;
-    return new String[] { pair.substring(0, separatorPos), pair.substring(separatorPos + 1) };
+    return StringUtils.splitKeyValue(pair);
   }
 
   private static boolean setGlobalConfiguration(final String key, final String value, final boolean printError) {
--- a/docs/6458-postgres-portal-suspension-execute-max-rows.md
+++ b/docs/6458-postgres-portal-suspension-execute-max-rows.md
@@ -1,0 +1,280 @@
+# Issue #6458 - Postgres Execute max-rows (portal suspension) broken
+
+Issue: https://github.com/ArcadeData/arcadedb/issues/6458
+PR: https://github.com/ArcadeData/arcadedb/pull/6658
+
+## Problem
+
+The Postgres extended-protocol `Execute` with a non-zero max-row count (portal suspension) was broken three
+ways in `PostgresNetworkExecutor.executeCommand()`:
+
+- `PortalSuspended` was written *before* the data rows instead of after.
+- It was sent *together with* `CommandComplete` - the wire protocol allows exactly one terminator per
+  Execute.
+- The portal was removed on the very Execute that suspended it, so the client's follow-up `Execute` meant to
+  continue the fetch found no portal and got `NoData` - the result was silently truncated to the first batch
+  with no way to resume.
+
+A client that drives this path is any JDBC/cursor client with a non-zero `Execute` max-row count - pgjdbc
+with `autoCommit(false)` + `setFetchSize(n)` is the common case.
+
+## Fix
+
+- `executeCommand()` no longer removes the portal on lookup (`getPortal(portalName, false)`); it is only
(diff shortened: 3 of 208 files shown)


robfrank:
Superseded by #6724: GitHub blocked retargeting this PR to `main` ("part of a stack" with #6653, which fell behind main). Same branch content, rebased directly onto current `main` (release-notes conflict resolved keeping both entries) and reopened there. Closing this one.

PR #6712, opened 2026-08-24
fix(#6655): reuse a stale persisted vector graph as a prefix instead of a synchronous rebuild

### LINKED ISSUE #6655: LSMVectorIndex: first search after a deferred close-time rebuild is still synchronous
Follow-up from #6653's code review (cc @lvca).

#6653 makes `flush()` skip the close-time graph rebuild for indexes at or above `ASYNC_REBUILD_MIN_GRAPH_SIZE`, deferring it to the next search instead of blocking `close()`. That's a real win (43s moved off the close path per the #6067 benchmark), but the deferred rebuild itself is not async: it runs through `ensureGraphAvailable()`'s existing stale-persisted-graph fallback (around `buildGraphFromScratch()` at the end of that method), which calls the full rebuild unconditionally and synchronously - unlike `rebuildGraphBeforeSearch()`'s own threshold-gated policy (sync under 1000 vectors, async above), which only governs *in-session* mutation-triggered rebuilds, not the reopen-with-a-stale-persisted-graph case.

Net effect: for a large index, the first search after a deferred close still blocks for the same full-rebuild wall-clock cost that used to block `close()` - it's just paid once, by one query, instead of by every close. Still a clear improvement, but a caller that specifically needs prompt *first-query* latency after reopening a write-heavy index does not get it today.

## Possible direction (not investigated - worth its own design discussion, same as #6067 was)

`ensureGraphAvailable()`'s stale-graph fallback could reuse `rebuildGraphBeforeSearch()`'s size threshold: below it, keep the synchronous rebuild (cheap); at/above it, serve the query from the stale-but-still-valid loaded graph (correct topology minus recent mutations - the same staleness a mid-session async rebuild already tolerates) while kicking off the same async rebuild path in the background, then hot-swap once it completes.

This needs its own careful look at `ensureGraphAvailable()`'s locking (currently takes `lock.writeLock()` for the whole check-and-build), since serving a "good enough for now" stale graph while asynchronously replacing it is a different contract than the current all-or-nothing load-or-build.

cc #6067, #6653

robfrank:
Closes #6655

## Summary

`ensureGraphAvailable()`'s stale-persisted-graph fallback (issue #3722/#6106 - a persisted graph with fewer nodes than there are live vectors) called `buildGraphFromScratch()` unconditionally and **synchronously**, on the calling search thread, with no size-based deferral of its own - unlike `rebuildGraphBeforeSearch()`, which already prefers an async rebuild above `ASYNC_REBUILD_MIN_GRAPH_SIZE` (1000) vectors for the identical staleness reached mid-session.

Net effect: for a large index, the **first search** after a reopen that finds a stale persisted graph paid the full rebuild cost inline - the same wall-clock cost issue #6067 moved off `close()`, just relocated to whichever query happened to run first. This was filed as a follow-up from #6653's code review (#6067's fix), but the underlying `ensureGraphAvailable()` staleness path it points at is reachable independently of #6067/#6653 (which are still open) - any reopen that finds a stale persisted graph hits it, for example after a crash that leaves the graph behind the WAL-recovered vector set.

## Fix

When the staleness is *provably* safe to reuse, the persisted graph is now reused immediately as a verified prefix instead of being discarded for a synchronous full rebuild:

- a manifest is present (a missing manifest can't support this - see the existing #6106 weak-comparison path, left untouched),
- no vectors were deleted (already excluded from this whole code path by the existing `hasDeletedVectors` guard), and
- the manifest's fingerprint proves the graph's own ordinals `[0, graphSize)` still resolve to *exactly* the records it was built from - i.e. the graph is only missing vectors added afterward, the same staleness `rebuildGraphBeforeSearch()`'s async policy already tolerates mid-session.

The vectors past `graphSize` are queued into the in-memory delta buffer - the same re-queue `buildGraphFromScratch()` already performs for a node its own build left unreachable - so they stay immediately searchable through the ordinary delta scan, rather than trading this fix's latency win for a window where they silently don't come back (that buffer is in-memory only, and this is a session that never wrote them - it's only now discovering them on reopen, so nothing else would have populated it). The same async rebuild `rebuildGraphBeforeSearch()` would use mid-session is kicked off in the background to fold the gap into the graph proper, with the inactivity timer armed as a fallback for when that async attempt is skipped (already in progress, or cooling down from a prior OOM deferral).

A missing manifest, a same-size mismatch, a graph larger than the live set, or any deletion all still fall through to the existing synchronous rebuild, unchanged.

A new `stalePrefixGraphReuses` stat (`LSMVectorIndex.getStats()`) counts how often the new path fires, alongside the existing `unverifiedGraphReuses`/`rebuildsDeferredForMemory` counters.

## Test plan

- [x] New regression test `Issue6655StaleGraphPrefixReuseTest` (tagged `slow`, mirrors `Issue6067DeferredCloseRebuildTest`'s fixture: 128 dims, 20,000+200 vectors, `maxConnections=64`, `beamWidth=400`):
  - builds and persists a graph over 20,000 vectors, closes cleanly;
  - reopens, writes 200 more vectors, then crashes (`DatabaseInternal.kill()`) instead of closing cleanly, so the persisted graph/manifest are left describing only the first 20,000 while WAL recovery on the next open restores all 20,200;
  - asserts the first search returns within a tripwire bound that a full synchronous rebuild of 20,200 vectors cannot meet (measured pre-fix: ~4.5s; bound: 2s) **and** already finds one of the 200 gap vectors by its own embedding - pinning both "fast" and "correct" so a half-fix (async-but-still-blocking, or fast-but-missing-data) cannot pass;
  - asserts `graphState` is `MUTABLE` and `stalePrefixGraphReuses == 1` immediately after that search, and `graphRebuildCount == 0` (nothing ran synchronously on the calling thread);
  - awaits the background async rebuild completing (`graphRebuildCount == 1`, `graphState` back to `IMMUTABLE`) and re-verifies the gap vector is still found once folded into the graph proper.
  - Verified the test fails without the fix (reverted the two source files, reran): fails at the same timing tripwire, `measured 4489 ms` vs. the `2000 ms` bound - i.e. it genuinely exercises the synchronous-rebuild-on-search-thread behavior being fixed here, not a vacuous pass.
- [x] `mvn -pl engine -am compile` / `test-compile` - clean.
- [x] Full regression pass, all green, no failures:
  - `Issue6106StaleVectorGraphTest` (7), `LSMVectorIndexRecoveryTest` (23), `GraphSearcherPoolStaleGraphTest` (5), `LSMVectorIndexBruteForceScanTest` (3), `CompactIndexGraphPersistFailureTest` (1), `Issue6655StaleGraphPrefixReuseTest` (1)
  - `LSMVectorIndexGraphManifestTest` (8), `LSMVectorIndexRebuildTest` (15), `VectorIndexRebuildsOnEveryCloseTest` (1), `VectorIndexNoRebuildOnUnusedCloseTest` (1), `LSMVectorIndexTest` (29), `LSMVectorIndexPersistenceTest` (2), `Issue6501GroupedDeltaMergeTest` (8), `Issue6502VectorPrefilterTest` (9), `Issue5558DeletedRegionSearchTest` (11), `Issue6496InactivityRebuildThresholdTest` (2), `Issue6503RebuildAdmissionControlTest` (8)
  - 134 tests total, 0 failures, 0 errors.

## Notes for the reviewer

- Scope is deliberately narrow: only the manifest-verified, no-deletions, growth-only case is reused; every other staleness reason (#6106's cross-generation mismatch, a truncated persist, a graph larger than the live set, no manifest) is untouched and still falls through to the pre-existing synchronous `buildGraphFromScratch()`.
- Locking is unchanged: the new logic runs inside the same `lock.writeLock()` section the existing stale-graph check already holds, and the extra work (reading the gap vectors) is a strict subset of the I/O the pre-existing code already does under that same lock to build `rebuiltOrdinalToVectorId` in the first place.
- The issue's own text flagged this as needing "its own careful look at `ensureGraphAvailable()`'s locking" for the general "serve a stale-but-good-enough graph while async-replacing it" design. This PR does not adopt that general design - it stays within the *existing* MUTABLE-graph-plus-delta-buffer contract the class already has (the same one `rebuildGraphBeforeSearch()`'s async path relies on), rather than introducing a new "hot-swap while serving reads" contract. Happy to discuss if a broader design is wanted instead.

## Summary by CodeRabbit

- **Performance**
  - Improved vector search availability after reopening with a partially outdated persisted index.
  - Existing searchable data can be reused immediately while newly added vectors are incorporated in the background.
  - Large indexes now avoid unnecessary synchronous rebuilds, reducing delays for the first search.

- **Metrics**
  - Added statistics for tracking reuse of persisted vector graphs.

- **Bug Fixes**
  - Improved recovery behavior after interruptions or crashes during vector index updates.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -1337,6 +1337,20 @@ public class LSMVectorIndex implements Index, IndexInternal {
   }
 
   /**
+   * A persisted graph {@link #ensureGraphAvailable()} decided is eligible for reuse as a prefix (issue #6655),
+   * carrying exactly what {@link #reuseStalePrefixGraph} needs to act on that decision after the write lock
+   * detecting it has been released.
+   *
+   * @param loadedGraph              the graph loaded from disk, already verified against the manifest
+   * @param liveOrdinalToVectorId    every live vector's id, in ordinal order - not only the ones the graph covers
+   * @param graphSize                how many of {@code liveOrdinalToVectorId}'s leading entries the graph covers
+   * @param vectorProp               the vector property name, for reading the gap vectors back
+   */
+  private record ReuseCandidate(ImmutableGraphIndex loadedGraph, int[] liveOrdinalToVectorId, int graphSize,
+                                 String vectorProp) {
+  }
+
+  /**
    * Ensure graph is available for searching. Lazy-loads from disk if needed.
    * This is the entry point for all search operations.
    */
@@ -1360,10 +1374,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
     // lock.writeLock() is held - it would stall every other reader and writer of this index for as long as the
     // read takes, exactly the stall buildGraphFromScratchWithRetry()'s own javadoc moved the (much larger)
     // full-rebuild validation off this same lock for (issue #5391; PR #6712 review).
-    ImmutableGraphIndex prefixReuseGraph      = null;
-    int[]               prefixReuseOrdinals   = null;
-    int                 prefixReuseGraphSize  = 0;
-    String              prefixReuseVectorProp = null;
+    ReuseCandidate prefixReuseCandidate = null;
 
     graphBuildLock.lock();
     try {
@@ -1519,6 +1530,11 @@ public class LSMVectorIndex implements Index, IndexInternal {
             // deferred-close one. A missing manifest cannot support this: the weak node-count comparison above
             // proves nothing about WHICH records a same-sized graph describes, and a prefix of it proves even
             // less. Below the async-rebuild threshold a synchronous rebuild is already cheap, so it is left alone.
+            // vectorIndex.size(), not graphIndex.size() as rebuildGraphBeforeSearch()'s analogous check uses:
+            // graphIndex is not yet set on this path (that is the whole point - the persisted graph has not been
+            // published to it yet), so the only live-vector count available here is the index's, which is also
+            // the more conservative of the two whenever they would differ (rebuiltOrdinalToVectorId.length can
+            // be smaller after validation filtering, never larger) - "is this worth deferring" errs towards yes.
             final int[] prefix = persistedManifest != null && graphSize == persistedManifest.vectorCount()
                 && graphSize < rebuiltOrdinalToVectorId.length && graphSize > 0
                 && vectorIndex.size() >= ASYNC_REBUILD_MIN_GRAPH_SIZE
@@ -1530,10 +1546,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
               // Deliberately not done here: reuseStalePrefixGraph() does real I/O for the gap and must run
               // unlocked (see the field javadoc above). Only the decision - not the read - belongs under this
               // lock, same as every other branch in this method.
-              prefixReuseGraph = loadedGraph;
-              prefixReuseOrdinals = prefix;
-              prefixReuseGraphSize = graphSize;
-              prefixReuseVectorProp = vectorProp;
+              prefixReuseCandidate = new ReuseCandidate(loadedGraph, prefix, graphSize, vectorProp);
             } else
               LogManager.instance().log(this, Level.INFO,
                   "Persisted graph is not usable for index %s: %s - rebuilding from scratch (issues #3722, #6106)",
@@ -1589,16 +1602,16 @@ public class LSMVectorIndex implements Index, IndexInternal {
       // while already holding it. Skipped when the stale graph was instead reused as a prefix above (issue
       // #6655): that already put the index to work as MUTABLE, and a synchronous rebuild here would throw
       // that work away and pay for it again.
-      if (prefixReuseGraph == null)
+      if (prefixReuseCandidate == null)
         buildGraphFromScratch();
     } finally {
       graphBuildLock.unlock();
     }
 
-    if (prefixReuseGraph != null) {
+    if (prefixReuseCandidate != null) {
       // Unlocked with respect to lock.writeLock(): reuseStalePrefixGraph() does its own graphBuildLock-serialized
       // I/O and a brief re-lock only to publish (see its javadoc).
-      reuseStalePrefixGraph(prefixReuseGraph, prefixReuseOrdinals, prefixReuseGraphSize, prefixReuseVectorProp);
+      reuseStalePrefixGraph(prefixReuseCandidate);
     }
   }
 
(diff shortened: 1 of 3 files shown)


robfrank:
Addressed the performance point: `reuseStalePrefixGraph()` now trims to `Arrays.copyOfRange(rebuiltOrdinalToVectorId, graphSize, length)` before it ever reaches `snapshotOf()` / `computeGraphBuildCacheCapacity()` / `forGraphBuild()`, and the read loop indexes that trimmed array 0-based instead of the full live-vector array with a `graphSize` offset. Cost is now bounded by the gap, not the total live vector count, matching the PR description.

Left the two minor points as-is, intentionally:

- **Double read of the gap vectors**: real, but the gap is bounded by construction (this path only fires above `ASYNC_REBUILD_MIN_GRAPH_SIZE`, and the whole point is that the gap is the *small* side of that split). Fixing it would mean threading a captured `VectorFloat` out of the validation filter that the fully-matching-graph branch above also uses, which is more invasive for a cost that's now O(gap) either way.
- **`startAsyncGraphRebuild()` not gated on `graphSize >= ASYNC_REBUILD_MIN_GRAPH_SIZE`**: this one I think is actually correct as written rather than an inconsistency - `buildGraphFromScratch()` (what the async thread runs) always rebuilds over the *current full live set*, not over `graphSize`, so the async-vs-sync decision that matters is already made by the entry gate on `vectorIndex.size()` before `reuseStalePrefixGraph()` is ever called. Gating a second time on `graphSize` would use the wrong number for what the rebuild actually costs.

Pushed as 02eeedf9aa. Reran `Issue6655StaleGraphPrefixReuseTest` and `Issue6106StaleVectorGraphTest` after the change - both green.

robfrank:
Addressed the main finding: `reuseStalePrefixGraph()`'s gap read now runs unlocked with respect to `lock.writeLock()`, serialized against a concurrent `buildGraphFromScratch()` (and a second concurrent reuse attempt) via the existing `graphBuildLock` mutex - the same one `buildGraphFromScratchWithRetry()` already uses for exactly this reason (issue #5391). `ensureGraphAvailable()` now only decides eligibility under `lock.writeLock()`; the actual gap read + publish happens after that lock is released. Publish re-takes `lock.writeLock()` only briefly, and merges into `deltaVectors`/`mutationsSinceSerialize` via `addAll()`/`addAndGet()` rather than a replace or `set()`, so a concurrent insert landing during the unlocked read isn't clobbered.

On the gap-size-bound point specifically: I left the entry gate as-is (total live vector count, not the gap) rather than adding a second bound, since fixing the lock scope makes the "small `graphSize`, huge gap" case merely as expensive as an ordinary rebuild would have been anyway - not a stall on the rest of the index the way it was before this commit. Happy to add an explicit gap-size check too if you'd still prefer it documented/enforced rather than left implicit.

Deferred, as a follow-up rather than in this PR: a dedicated concurrent-contention test (holding a concurrent search/insert while the reuse path runs) to pin the lock-scope fix itself. `graphBuildLock` is private, so exercising this deterministically needs either a test-only accessor or a timing-based reproduction, and the existing regression suite (including `LSMVectorIndexRecoveryTest` and `LSMVectorIndexRebuildTest`, which already exercise concurrent builders against this same `graphBuildLock`) still passes unchanged, so I didn't want to hold up this fix on constructing that harness.

Pushed as d596546b06.

robfrank:
Thanks - appreciated the thorough trace through the locking. Addressed the two easy minor points:

- Added a comment on why the gate checks `vectorIndex.size()` rather than `graphIndex.size()` (point 2).
- Replaced the four `prefixReuse*` locals with a single `ReuseCandidate` record (point 4).

Point 1 (the pre-existing `rebuiltOrdinalToVectorId` validation loop also running under `lock.writeLock()`) filed as a follow-up: #6713. Agreed it's a real finding, but it predates this PR and touches both branches of `ensureGraphAvailable()`, not just the one added here, so it deserves its own look rather than riding along on this diff.

Point 3 (`computeGraphBuildCacheCapacity(..., false)` hardcoding non-inline-quantization) left as-is - as noted in the review itself, it matches the pre-existing PQ-rebuild call a few lines up, so it's consistent with an existing pattern this PR didn't introduce, not a new gap.

Pushed as 835ea14c5d.

robfrank:
Thanks for the thorough re-review - no further action needed on this pass. Both notes (redundant validation work on concurrent first-search contention, and the unconditional async rebuild trading off against `rebuildGraphBeforeSearch()`'s mutation-threshold gate) are explicitly the intended tradeoffs, not regressions, so left as-is per the review's own assessment.

This PR is ready for a human merge decision.

Issue #6672, opened 2026-08-24
agg.statistics/agg.maxItems init max with Double.MIN_VALUE (smallest positive): wrong max / empty items for any dataset whose max is <= 0


ruispereira:
### Summary
`agg.statistics` and `agg.maxItems` initialize their running maximum with `Double.MIN_VALUE`, which in Java is the smallest *positive* value (~4.9E-324), not the most-negative double. For any dataset whose true maximum is ≤ 0 the max is never updated, so `agg.statistics` reports a bogus tiny-positive max and `agg.maxItems` returns a bogus value with an **empty** items list.

### Where (HEAD `86a704526`)
`engine/src/main/java/com/arcadedb/function/agg/AggStatistics.java:73`: `double max = Double.MIN_VALUE;` then `max = Math.max(max, value)`.
`engine/src/main/java/com/arcadedb/function/agg/AggMaxItems.java:69`: `double maxValue = Double.MIN_VALUE;` then `if (value > maxValue) {...} else if (value == maxValue) {...}`.
Because every non-positive input is `< 4.9E-324`, `Math.max` never replaces the seed and neither `AggMaxItems` branch ever fires. The sibling `AggMinItems` (seeds `Double.MAX_VALUE`) is correct, and `AggStatistics`'s `min`/`sum`/`mean`/`stdev`/`median` are correct — only `max` is affected.

### Repro
- `agg.statistics([-5.0, -3.0, -1.0])` → expected `max = -1.0`; actual `max ≈ 4.9E-324` (also fails when the true max is exactly 0).
- `agg.maxItems([-5.0, -3.0, -1.0], ["a","b","c"])` → expected `{value:-1.0, items:["c"]}`; actual `{value:4.9E-324, items:[]}` (wrong value **and** empty items — the function's purpose is lost).

### Suggested fix
Seed `double max = Double.NEGATIVE_INFINITY;` (equivalently `-Double.MAX_VALUE`) in both.

### Scope / confidence
High (formulas read at HEAD). Triggers whenever the maximum element is ≤ 0 (negative-only series: deltas, losses, offsets); positive-max datasets are unaffected.

lvca:
Fixed in #6687 (merged).

`agg.statistics()` and `agg.maxItems()` computed a wrong "max" value when every number in the input was zero or negative (`agg.maxItems` also returned an empty items list in that case). This is now fixed, e.g. `agg.statistics([-5, -3, -1])` correctly reports `max: -1.0`.

[...]

PR #6714, opened 2026-08-25
fix(#6316,#6265,#6318,#6705): algo.* CSR routing, influenceMaximization allocation, LCC CSR abortability, stale avg() tests

### LINKED ISSUE #6265: algo.influenceMaximization allocates two nodeCount-sized arrays per Monte Carlo simulation
Split out of #6216 / PR #6222, where it was noticed twice while reasoning about checkpoint placement and deliberately left alone: it is a performance defect, not a validation one, and folding it into a bounds-checking fix would have made that PR harder to review.

## The allocation

`AlgoInfluenceMaximization.simulateIC` opens with two fresh arrays on **every call**:

```java
private int simulateIC(final int[][] adj, final int[] seeds, final int seedCount,
    final int candidate, final int n, final double p, final Random rng) {
  final boolean[] activated = new boolean[n];
  // BFS queue (int[] with head/tail pointers — zero boxing)
  final int[] queue = new int[n];
```

The comment above the queue is telling: the method was written to avoid boxing in the inner loop, and then allocates `5n` bytes per call anyway.

## How often that is

`simulateIC` sits at the bottom of a triple nest - `round` x `candidate` x `sim`:

```java
for (int round = 0; round < seedCount; round++)
  for (int candidate = 0; candidate < n; candidate++)
    for (int sim = 0; sim < simulations; sim++)
      totalSpread += simulateIC(...);
```

So the call count is `k x n x simulations`, and each call allocates and zero-fills `n` booleans plus `n` ints. Total churn is **`k x n^2 x simulations x 5` bytes**.

At the default `simulations: 100` with `k = 3`:

| nodes | calls | bytes allocated and zeroed |
|---|---|---|
| 1 000 | 300 000 | ~1.5 GB |
| 10 000 | 3 000 000 | ~150 GB |
| 100 000 | 30 000 000 | ~15 TB |

None of it survives the call, so it is pure allocation-rate and zero-fill cost on top of an algorithm that is already O(k x n x simulations x (n + m)).

## Fix

Hoist both buffers to the caller and clear them per simulation. The clear can be cheaper than the allocation in the common case: an IC cascade usually activates far fewer than `n` nodes, so resetting only the entries recorded in `queue[0..tail)` is O(activated) rather than O(n), where the current code pays O(n) twice per call regardless.

That turns the per-simulation cost from `2 x O(n)` allocate-and-zero into `O(activated)`, which on a sparse graph with a low `propagationProbability` is a large constant-factor win - exactly the regime the defaults describe.

## Note on the checkpoint

`AlgoInfluenceMaximization`'s simulation loop calls the unthrottled `guard.check()` rather than `checkPeriodically`, and the comment at that call site justifies it by saying `simulateIC` "allocates and zero-fills two arrays of size n before the cascade starts, so it is never O(1) on any graph shape". That reasoning is load-bearing on the very allocation this issue removes.

If the buffers are hoisted, a simulation on a small graph *can* become close to O(1), and the checkpoint should be re-examined in the same change - most likely by keeping `check()` (its cost is one flag test when no timeout is configured) but updating the comment, since the alternative of throttling to 1024 makes abort latency `1024 x O(n + m)` on large graphs. Either way the comment must not be left asserting something that is no longer true.

## Related

#6216, PR #6222.

### LINKED ISSUE #6316: Six algo.* procedures never look for a Graph Analytical View at all, and hand-roll the loading and weight extraction the shared helpers now own

### LINKED ISSUE #6318: algo.localClusteringCoefficient is abortable on OLTP and not on CSR, and nothing stops the next procedure from shipping without a checkpoint either
Two leftovers from #6302 (PR #6306), stated together because they are the same question - how does the abortability guarantee hold once the PR that established it is merged - approached from opposite ends.

## 1. The one procedure the sweep could not cover

#6302 swept all 68 `algo.*` procedures and gave a `WorkGuard` checkpoint to the 22 whose dominant loop is superlinear in the graph. `algo.localClusteringCoefficient` got one **on its OLTP path only**, and the code says so rather than hiding it:

```java
// AlgoLocalClusteringCoefficient.executeWithOLTP
// Only the OLTP path is covered here. The CSR path hands the whole computation to
// GraphAlgorithms#localClusteringCoefficient, which counts triangles across a thread pool, and the
// WorkCheckpoint hook #6264 introduced is specified to be called between iterations on the calling
// thread rather than from inside a parallel chunk [...]
```

So the same query is abortable without a Graph Analytical View and unabortable with one - and the CSR path is the one taken on the large graphs where it matters. The cost is O(m x sqrt(m)) triangle intersections with nothing but the graph sizing it.

`WorkCheckpoint`'s contract is the obstacle, and it is a deliberate one: "called from the calling thread between two iterations, never from inside a parallel chunk, so an implementation does not have to be thread-safe and the exception it throws propagates straight to the caller". `lccBuildAndIntersect` partitions the node range across a `Thread[]` and has no between-iterations point to hook.

Two ways out:
- **Chunk the parallel phase.** Partition into k batches instead of one-per-thread, join between batches, and check there. Costs k barriers; abort latency becomes one batch. Keeps `WorkCheckpoint` single-threaded as specified.
- **Widen the contract** to a thread-safe volatile flag each worker polls, with the exception raised by the joining thread. Lower latency, but every kernel then has to be audited for it, and the current contract exists precisely to avoid that.

The first is the smaller change and fits the existing hook. Worth doing for the whole `GraphAlgorithms` parallel family rather than for LCC alone, since `pageRank` and `labelPropagation` already take a `WorkCheckpoint` and would gain the same intra-iteration granularity from it.

## 2. The convention has no backstop

Both review rounds on PR #6306 raised this independently, and it is the reason the bug class existed in the first place. `newWorkGuard(context)` plus a hand-placed `guard.check()` is a convention copied file to file across 22 classes. `Issue6302AlgoGraphDrivenWorkGuardTest`'s parameterised table is the only thing that would notice a new procedure shipping without one - and only if somebody remembers to add a row for it.

A base-class hook was considered and is the wrong answer: the *placement* is a per-loop judgement (unthrottled `check()` where one iteration is already O(V), `checkPeriodically` on a running counter where the unbounded thing is the product of two loops), so a hook would standardise the part that is already easy and leave the part that is hard. #6264 reached the same conclusion over three review rounds.

What would actually catch it is a check over the package, roughly: *a loop whose bound is `nodeCount`, `edgeCount` or an adjacency length, inside `execute()`, with no `guard` reference anywhere in its body*. As a test it can be a source scan asserting against a known allow-list; as tooling it is an ArchUnit-style rule or a Codacy pattern. Either way it turns "someone has to remember" into "the build says so", which is the difference between the 22 staying guarded and drifting.

The same check would have found the #6295 instances directly: every one of them was a loop over `nodeCount` in a class that *already had* a guard, just not in that phase.

## Related

#6302, #6295, #6264, #6216, PR #6306.

### LINKED ISSUE #6705: SQL aggregate functions (avg) over zero matching rows return a phantom null row instead of an empty result set
Found incidentally while investigating unrelated CI failures on PR #6702 (unaffected by that PR's diff, which is entirely inside \`com.arcadedb.index.lsm.LSMTreeIndexCompacted\`).

\`engine/src/test/java/com/arcadedb/engine/timeseries/Issue5414LastPointTest.java\`'s \`descendingScanOnAnEmptySeriesReturnsNothing()\` fails on current \`main\` (reproduces deterministically, in isolation, with just this one test class):

\`\`\`
java.lang.AssertionError:

Expecting empty but was: [{v: null}]
	at com.arcadedb.engine.timeseries.Issue5414LastPointTest.descendingScanOnAnEmptySeriesReturnsNothing(Issue5414LastPointTest.java:377)
\`\`\`

Repro:
\`\`\`sql
CREATE TIMESERIES TYPE Empty TIMESTAMP ts TAGS (host STRING) FIELDS (value DOUBLE)
SELECT avg(value) AS v FROM Empty WHERE host = 'a'
-- returns [{v: null}] instead of an empty result set
\`\`\`

The other two assertions in the same test method (\`SELECT ts FROM Empty ... ORDER BY ts DESC LIMIT 1\` and \`SELECT ts.last(value, ts) AS v FROM Empty ...\`) both correctly return empty - only the \`avg()\` aggregate over a host with zero matching rows in an empty time series produces a single \`{v: null}\` row.

Not urgent / not blocking anything - filed so it isn't lost, per the test's own author's intent (it's an existing regression test, not new).

lvca:
Fixes #6316
Fixes #6265
Fixes #6318
Fixes #6705

## Summary

- **#6316** - `algo.kShortestPaths`, `algo.mst`, `algo.msa`, `algo.maxFlow`, `algo.betweenness` and `algo.degree` never resolved a Graph Analytical View at all, hand-rolling vertex loading and RID resolution instead of `loadGraph`/`GraphData`. All six now route through the shared helper; the two weighted ones (`kShortestPaths`, `msa`) read edges through `GraphData.weightedAdjacency`. `algo.betweenness` also moves its RID index off `HashMap<Vertex, Integer>` onto the RID-keyed scheme every other procedure uses. `algo.degree` gains `GraphData.degrees()`, an O(1)-per-node CSR degree lookup via `GraphTraversalProvider.getDegrees` instead of materialising a neighbour list just to count it. New CSR-vs-OLTP equivalence tests assert `CommandContext.CSR_ACCELERATED_VAR` so the comparison can't silently pin the OLTP path twice.
- **#6265** - `algo.influenceMaximization`'s `simulateIC` allocated a `boolean[n]` and an `int[n]` on every one of its `k x n x simulations` calls. Both buffers are now hoisted to the caller and reused, with `simulateIC` resetting only the entries it touched.
- **#6318** - `algo.localClusteringCoefficient`'s CSR path had no checkpoint at all (its OLTP path did). `GraphAlgorithms` gains `parallelForRangeCheckpointed`, which batches a range and checks in between batches; LCC's two parallel phases plus its sequential prep passes now use it, as do `pageRank`/`labelPropagation`'s own parallel phases for finer intra-iteration abort latency. A new backstop test discovers every `algo.*` procedure file on disk and fails if one is neither self-guarded nor in a justified single-pass exemption list.
- **#6705** - Not a real bug: two regression tests asserted the pre-#6680 "empty aggregate returns zero rows" behaviour for `avg()` with no `GROUP BY`. #6680 deliberately made that return one row with a null value (matching ANSI SQL scalar-aggregate semantics and `count(*)`'s own behaviour); the two tests were never updated. Fixed the tests, not the production code.
- **#6672** - Investigated and found already fixed by #6687 (merged); closed separately with no further change needed.

## Known limitations (tracked separately)

- **#6715** - `weightedAdjacencyFromColumns` (used by the columnar CSR path of `algo.mst`/`algo.msa`/`algo.maxFlow`/`algo.kShortestPaths`, and pre-existing for `algo.steinerTree`/`algo.bellmanFord`) throttles its `WorkGuard` checkpoint per node, not per edge, because `GraphTraversalProvider#edgeWeightsOf`'s only implementation has no checkpoint hook inside its own O(degree) loop. A single supernode is one unabortable unit of work regardless of fan-out. This is not new to this PR - the four newly-routed procedures previously always used a hand-rolled OLTP walk with genuine per-edge throttling, so it is a narrow, traded-off abortability regression for graphs with a very large supernode. A proper fix needs to widen the `GraphTraversalProvider` SPI itself, which is out of scope for this PR; documented in place and tracked in #6715.
- **#6716** - `algo.degree(relTypes?, direction?)` parses its `direction` argument but never uses it (pre-existing, predates this PR's conversion of the procedure onto `loadGraph`/`GraphData`).

## Test plan

- [x] `mvn -o -pl engine -am compile test-compile` - clean
- [x] New regression tests for each issue: `Issue6265InfluenceMaximizationBufferReuseTest`, `Issue6316AlgoLoadGraphAdoptionTest`, `Issue6318ParallelForRangeCheckpointedTest`, `Issue6318LocalClusteringCoefficientCsrAbortabilityTest`, `Issue6318AlgoGuardCoverageBackstopTest`
- [x] Verified `Issue6318LocalClusteringCoefficientCsrAbortabilityTest` genuinely fails against pre-fix code (stashed the production changes, confirmed the interrupt test failed with "no throwable raised")
- [x] Full `com.arcadedb.query.opencypher.**`, `com.arcadedb.graph.**`, `com.arcadedb.function.**`, `com.arcadedb.engine.timeseries.**`, `com.arcadedb.query.sql.executor.**` suites: 12900 tests, 0 failures caused by this change (3 pre-existing flakes confirmed unrelated - pass in isolation, fail only under a specific concurrent test-class ordering)
- [x] Fixed two other tests whose assertions pinned now-superseded implementation details: `Issue6300AlgoMSTEdgeBudgetTest` (edge-count refusal message shape) and `Issue6264AlgoIterationKnobGuardTest` (pageRank checkpoint call count)

https://claude.ai/code/session_013az14F8gvbcYips5GqiZfy

## Summary by CodeRabbit

- **New Features**
  - Expanded analytical-view support across centrality, pathfinding, spanning-tree, flow, and arborescence algorithms.
  - Added more responsive interruption support for long-running graph algorithms, including clustering calculations.

- **Bug Fixes**
  - Improved graph algorithm memory usage and execution efficiency, including influence maximization.
  - Strengthened working-memory budget enforcement.
  - Improved consistency between standard and analytical-view results.
  - Empty-input averages and time-series aggregates now consistently return one row containing `null`.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
@@ -1233,6 +1233,14 @@ public final class GraphAlgorithms {
 
     final int maxDeg = maxDegree;
 
+    // One buffer per thread that ever runs a chunk of this call, reused across every batch
+    // parallelForRangeCheckpointed splits a pass into and every iteration of the outer loop - not
+    // reallocated per chunk invocation. Before issue #6318 threaded checkpointing through this loop's
+    // parallel phase, a fresh neighborBuf per chunk cost PARALLELISM allocations per iteration; batching
+    // into up to CHECKPOINT_BATCHES chunks-of-chunks would have multiplied that up to CHECKPOINT_BATCHES x
+    // without this (PR #6714 review round 9).
+    final ThreadLocal<int[]> neighborBufTL = new ThreadLocal<>();
+
     for (int iter = 0; iter < maxIters; iter++) {
       checkpoint.check();
       System.arraycopy(labels, 0, newLabels, 0, n);
@@ -1240,7 +1248,11 @@ public final class GraphAlgorithms {
       final AtomicBoolean anyChanged = new AtomicBoolean(false);
 
       parallelForRangeCheckpointed(n, checkpoint, (start, end) -> {
-        final int[] neighborBuf = new int[maxDeg];
+        int[] neighborBuf = neighborBufTL.get();
+        if (neighborBuf == null || neighborBuf.length < maxDeg) {
+          neighborBuf = new int[maxDeg];
+          neighborBufTL.set(neighborBuf);
+        }
         boolean localChanged = false;
 
         for (int u = start; u < end; u++) {
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
@@ -61,10 +61,11 @@ public final class GraphAlgorithms {
   private static final int PARALLEL_THRESHOLD     = 8192;
   /** How many checkpointed batches {@link #parallelForRangeCheckpointed} splits a full range into, at most. */
   private static final int CHECKPOINT_BATCHES     = 16;
-  /** How often {@link #lccBuildAndIntersect}'s sequential prep passes check in: cheap enough per node that
-   *  checking every one would cost more than the check saves, same tradeoff {@code GraphData.adjacency}'s
-   *  node-count checkpoint strikes. */
-  private static final int LCC_PREP_CHECKPOINT_STRIDE = 4095;
+  /** Bitmask for how often {@link #lccBuildAndIntersect}'s sequential prep passes check in - {@code (u & MASK) ==
+   *  MASK} is true every 4096th node, not every {@code MASK}th one. Cheap enough per node that checking every one
+   *  would cost more than the check saves, same tradeoff {@code GraphData.adjacency}'s node-count checkpoint
+   *  strikes. */
+  private static final int LCC_PREP_CHECKPOINT_MASK = 4095;
   private static final int PARALLEL_BFS_THRESHOLD = 4096;
   private static final double ALPHA              = 8.0;  // edge ratio for push->pull switch
   private static final int PULL_ENTER_DIVISOR    = 8;    // push->pull when frontier > n/8
@@ -1365,7 +1366,7 @@ public final class GraphAlgorithms {
       if (csr == null)
         continue;
       for (int u = 0; u < n; u++) {
-        if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+        if ((u & LCC_PREP_CHECKPOINT_MASK) == LCC_PREP_CHECKPOINT_MASK)
           checkpoint.check();
         degree[u] += csr.outDegree(u) + csr.inDegree(u);
       }
@@ -1390,7 +1391,7 @@ public final class GraphAlgorithms {
       final int[] bwdOffsets = csr.getBackwardOffsets();
       final int[] bwdNeighbors = csr.getBackwardNeighbors();
       for (int u = 0; u < n; u++) {
-        if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+        if ((u & LCC_PREP_CHECKPOINT_MASK) == LCC_PREP_CHECKPOINT_MASK)
           checkpoint.check();
         int ia = fwdOffsets[u], aEnd = fwdOffsets[u + 1];
         int ib = bwdOffsets[u], bEnd = bwdOffsets[u + 1];
@@ -1415,7 +1416,7 @@ public final class GraphAlgorithms {
         final int[] fwdOffsets = csr.getForwardOffsets();
         final int[] fwdNeighbors = csr.getForwardNeighbors();
         for (int u = 0; u < n; u++) {
-          if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+          if ((u & LCC_PREP_CHECKPOINT_MASK) == LCC_PREP_CHECKPOINT_MASK)
             checkpoint.check();
           for (int j = fwdOffsets[u]; j < fwdOffsets[u + 1]; j++)
             neighbors[pos[u]++] = fwdNeighbors[j];
@@ -1423,7 +1424,7 @@ public final class GraphAlgorithms {
         final int[] bwdOffsets = csr.getBackwardOffsets();
         final int[] bwdNeighbors = csr.getBackwardNeighbors();
         for (int u = 0; u < n; u++) {
-          if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+          if ((u & LCC_PREP_CHECKPOINT_MASK) == LCC_PREP_CHECKPOINT_MASK)
             checkpoint.check();
           for (int j = bwdOffsets[u]; j < bwdOffsets[u + 1]; j++)
             neighbors[pos[u]++] = bwdNeighbors[j];
@@ -1445,7 +1446,7 @@ public final class GraphAlgorithms {
     // checkpointed the same periodic way as the prep passes above.
     int write = 0;
     for (int u = 0; u < n; u++) {
-      if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+      if ((u & LCC_PREP_CHECKPOINT_MASK) == LCC_PREP_CHECKPOINT_MASK)
         checkpoint.check();
       final int readStart = offsets[u];
       final int readEnd = offsets[u + 1];
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAlgorithms.java
@@ -143,6 +143,31 @@ public final class GraphAlgorithms {
     }
   }
 
+  /** How many checkpointed batches {@link #parallelForRangeCheckpointed} splits a full range into, at most. */
+  private static final int CHECKPOINT_BATCHES = 16;
+
+  /**
+   * As {@link #parallelForRange}, but for a kernel with no natural per-iteration checkpoint of its own: the range
+   * is split into up to {@link #CHECKPOINT_BATCHES} batches, each run through {@link #parallelForRange} in turn,
+   * with {@code checkpoint} called on the calling thread between batches - never from inside a worker chunk, same
+   * contract as {@link WorkCheckpoint#check()} documents.
+   * <p>
+   * A batch is never smaller than {@link #PARALLEL_THRESHOLD}, so a graph too small to parallelise still runs as
+   * one batch with one checkpoint call, exactly like a plain {@link #parallelForRange} call would have. On a large
+   * graph this is what gives {@link #localClusteringCoefficient} intra-pass abortability at all (issue #6318: its
+   * one-shot triangle count had no iteration boundary to hang a checkpoint on the way {@link #pageRank} and
+   * {@link #labelPropagation} do), and it narrows their own between-iterations latency to within a single pass.
+   */
+  static void parallelForRangeCheckpointed(final int n, final WorkCheckpoint checkpoint, final BiConsumer<Integer, Integer> work) {
+    final int batchSize = Math.max(PARALLEL_THRESHOLD, (n + CHECKPOINT_BATCHES - 1) / CHECKPOINT_BATCHES);
+    for (int start = 0; start < n; start += batchSize) {
+      checkpoint.check();
+      final int batchStart = start;
+      final int batchEnd = Math.min(start + batchSize, n);
+      parallelForRange(batchEnd - batchStart, (s, e) -> work.accept(batchStart + s, batchStart + e));
+    }
+  }
+
   // --- PageRank (Pull-based, Parallel) ---
 
   /**
@@ -171,14 +196,14 @@ public final class GraphAlgorithms {
    * <p>
    * This kernel has no convergence test - it always runs the full {@code iterations} count - and that count comes
    * straight from a caller-supplied knob, so without a checkpoint {@code algo.pageRank({maxIterations: 2000000000})}
-   * spins with nothing able to stop it. The hook is called once per power iteration, which bounds abort latency by
-   * one sweep of the graph and costs one virtual call per O(n + m) of work.
+   * spins with nothing able to stop it. The hook is called once per power iteration at minimum, which bounds abort
+   * latency by one sweep of the graph and costs one virtual call per O(n + m) of work.
    * <p>
-   * One sweep is deliberately coarser than the ~1024-node latency the inline OLTP loops of the {@code algo.*}
-   * procedures achieve, and it is as fine as this kernel can be: the per-node work here runs inside
-   * {@link #parallelForRange}, on worker threads that would not observe an interrupt aimed at the calling thread,
-   * and throwing out of a chunk closure would leave its siblings running. So the checkpoint stays on the calling
-   * thread, between the parallel phases.
+   * The per-node work here runs inside {@link #parallelForRange}, on worker threads that would not observe an
+   * interrupt aimed at the calling thread, and throwing out of a chunk closure would leave its siblings running.
+   * So the checkpoint stays on the calling thread, between parallel phases - via {@link #parallelForRangeCheckpointed}
+   * rather than {@link #parallelForRange} directly, which narrows abort latency to within a single large phase
+   * rather than only between iterations (issue #6318).
    *
    * @param checkpoint called between iterations; throws to abort. {@link WorkCheckpoint#NONE} to run unbounded
    */
@@ -253,13 +278,13 @@ public final class GraphAlgorithms {
       // Pre-compute contribution per node: rank[v] / outDeg[v].
       // This turns the inner loop into a single gather+accumulate with no arithmetic.
       final double[] contrib = new double[n];
-      parallelForRange(n, (s, e) -> {
+      parallelForRangeCheckpointed(n, checkpoint, (s, e) -> {
         for (int u = s; u < e; u++)
           contrib[u] = currentRank[u] * invDeg[u];
       });
 
       // PULL: each node sums contributions from neighbors — parallel, zero sync
-      parallelForRange(n, (start, end) -> {
+      parallelForRangeCheckpointed(n, checkpoint, (start, end) -> {
         for (int u = start; u < end; u++) {
           double sum = 0;
           for (int t = 0; t < typeCount; t++) {
@@ -286,7 +311,7 @@ public final class GraphAlgorithms {
         danglingSum += currentRank[danglingNodes[i]];
       if (danglingSum > 0.0) {
         final double danglingContrib = damping * danglingSum / n;
-        parallelForRange(n, (s, e) -> {
+        parallelForRangeCheckpointed(n, checkpoint, (s, e) -> {
           for (int u = s; u < e; u++)
             nextRank[u] += danglingContrib;
         });
@@ -1147,9 +1172,10 @@ public final class GraphAlgorithms {
    * <p>
    * The convergence test ({@code break} once no label moved) is not a bound on {@code maxIters}: a graph that
    * oscillates between two labellings never converges, so the caller-supplied knob is what decides when the run
-   * ends. The hook is called once per iteration, which bounds abort latency by one sweep of the graph - coarser than
-   * the inline OLTP loops of the {@code algo.*} procedures, and as fine as this kernel can be, for the reason given
-   * on {@link #pageRank(GraphAnalyticalView, double, int, DIRECTION, WorkCheckpoint, String...)}.
+   * ends. The hook is called once per iteration at minimum, which bounds abort latency by one sweep of the graph;
+   * the per-node work inside that sweep runs through {@link #parallelForRangeCheckpointed} for the same
+   * intra-phase latency {@link #pageRank(GraphAnalyticalView, double, int, DIRECTION, WorkCheckpoint, String...)}
+   * gets from it.
    *
    * @param checkpoint called between iterations; throws to abort. {@link WorkCheckpoint#NONE} to run unbounded
    */
@@ -1204,7 +1230,7 @@ public final class GraphAlgorithms {
 
       final AtomicBoolean anyChanged = new AtomicBoolean(false);
 
-      parallelForRange(n, (start, end) -> {
+      parallelForRangeCheckpointed(n, checkpoint, (start, end) -> {
         final int[] neighborBuf = new int[maxDeg];
         boolean localChanged = false;
 
@@ -1289,14 +1315,35 @@ public final class GraphAlgorithms {
    * @return double[] of LCC values indexed by dense node ID
    */
   public static double[] localClusteringCoefficient(final GraphAnalyticalView view, final String... edgeTypes) {
+    return localClusteringCoefficient(view, WorkCheckpoint.NONE, edgeTypes);
+  }
+
+  /**
+   * {@link #localClusteringCoefficient(GraphAnalyticalView, String...)} with a cooperative abort hook.
+   * <p>
+   * Unlike {@link #pageRank} and {@link #labelPropagation}, this kernel has no iteration loop to hang a
+   * once-per-iteration checkpoint on: it is one O(m x sqrt(m)) pass with nothing but the graph sizing it (issue
+   * #6318). {@code checkpoint} is called periodically through the sequential prep passes (degree count, adjacency
+   * build, compaction) and, via {@link #parallelForRangeCheckpointed}, between the batches its two parallel
+   * phases - the multi-type sort and the triangle count - are now split into, so the CSR path gets the same
+   * abortability the OLTP path (whole node per {@code guard.check()}) already had.
+   *
+   * @param checkpoint called periodically; throws to abort. {@link WorkCheckpoint#NONE} to run unbounded
+   */
+  public static double[] localClusteringCoefficient(final GraphAnalyticalView view, final WorkCheckpoint checkpoint,
+      final String... edgeTypes) {
     final int n = view.getNodeMapping().size();
     if (n == 0)
       return new double[0];
 
     final String[] types = resolveEdgeTypes(view, edgeTypes);
 
-    return lccBuildAndIntersect(view, types, n);
+    return lccBuildAndIntersect(view, types, n, checkpoint);
   }
+
+  /** How often the sequential prep passes below check in: cheap enough per node that checking every one would
+   *  cost more than the check saves, same tradeoff {@code GraphData.adjacency}'s node-count checkpoint strikes. */
+  private static final int LCC_PREP_CHECKPOINT_STRIDE = 4095;
 
   /**
    * Builds a flat merged undirected adjacency from CSR arrays, then counts triangles via
@@ -1304,7 +1351,8 @@ public final class GraphAlgorithms {
    * since CSR forward and backward arrays are already individually sorted.
    * Multi-type sort phase and triangle counting are both parallelized.
    */
-  private static double[] lccBuildAndIntersect(final GraphAnalyticalView view, final String[] types, final int n) {
+  private static double[] lccBuildAndIntersect(final GraphAnalyticalView view, final String[] types, final int n,
+      final WorkCheckpoint checkpoint) {
     final boolean singleType = types.length == 1;
     // First pass: compute degree per node
     final int[] degree = new int[n];
@@ -1312,8 +1360,11 @@ public final class GraphAlgorithms {
       final CSRAdjacencyIndex csr = view.getCSRIndex(edgeType);
       if (csr == null)
         continue;
-      for (int u = 0; u < n; u++)
+      for (int u = 0; u < n; u++) {
+        if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+          checkpoint.check();
         degree[u] += csr.outDegree(u) + csr.inDegree(u);
+      }
     }
 
     // Build offsets from degrees
@@ -1335,6 +1386,8 @@ public final class GraphAlgorithms {
       final int[] bwdOffsets = csr.getBackwardOffsets();
       final int[] bwdNeighbors = csr.getBackwardNeighbors();
       for (int u = 0; u < n; u++) {
+        if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+          checkpoint.check();
         int ia = fwdOffsets[u], aEnd = fwdOffsets[u + 1];
         int ib = bwdOffsets[u], bEnd = bwdOffsets[u + 1];
         int p = offsets[u];
@@ -1357,17 +1410,23 @@ public final class GraphAlgorithms {
           continue;
         final int[] fwdOffsets = csr.getForwardOffsets();
         final int[] fwdNeighbors = csr.getForwardNeighbors();
-        for (int u = 0; u < n; u++)
+        for (int u = 0; u < n; u++) {
+          if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+            checkpoint.check();
           for (int j = fwdOffsets[u]; j < fwdOffsets[u + 1]; j++)
             neighbors[pos[u]++] = fwdNeighbors[j];
+        }
         final int[] bwdOffsets = csr.getBackwardOffsets();
         final int[] bwdNeighbors = csr.getBackwardNeighbors();
-        for (int u = 0; u < n; u++)
+        for (int u = 0; u < n; u++) {
+          if ((u & LCC_PREP_CHECKPOINT_STRIDE) == LCC_PREP_CHECKPOINT_STRIDE)
+            checkpoint.check();
           for (int j = bwdOffsets[u]; j < bwdOffsets[u + 1]; j++)
             neighbors[pos[u]++] = bwdNeighbors[j];
+        }
       }
-      // Sort each adjacency list — parallel for large graphs
-      parallelForRange(n, (start, end) -> {
+      // Sort each adjacency list — parallel for large graphs, checkpointed between batches (issue #6318)
+      parallelForRangeCheckpointed(n, checkpoint, (start, end) -> {
         for (int u = start; u < end; u++)
           Arrays.sort(neighbors, offsets[u], offsets[u + 1]);
       });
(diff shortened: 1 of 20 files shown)


lvca:
Thanks for the review. Addressed in 7970a285a:

- **GraphData.degrees() buffer reuse**: added a one-line comment noting `getDegrees()` overwrites `perType` rather than accumulating into it, so the per-type summation is correct (confirmed correct in your review, just under-documented).
- **parallelForRangeCheckpointed dispatch overhead**: added `Issue6318ParallelForRangeCheckpointedTest.checkpointedBatchingOverheadIsSmallRelativeToPageRankShapedWork` (`@Tag("benchmark")`), comparing it against a plain `parallelForRange` call on pageRank-shaped per-node work (a handful of flops over a small fixed-size neighbour array, not a bare store - which would be dispatch-dominated and overstate the overhead). Confirms the added dispatch cost stays a small fraction of realistic per-node work rather than asserting it from the shape of the code.

Not changing, per your own notes:
- **algo.mst/algo.msa's narrower early-refusal window**: this is the deliberate, disclosed tradeoff you flagged - already called out in the updated `Issue6300AlgoMSTEdgeBudgetTest` comments and consistent with `algo.steinerTree`'s existing precedent around its own `weightedAdjacency` call. Confirmed intentional, not something to revisit here.
- **AlgoDegreeCentrality's unused `dirArg`**: pre-existing (predates this PR), and you already scoped it as a follow-up rather than a blocker for this change.

lvca:
Fixed both nits from the second review round in 905c18454: removed the leftover `java.util.Map` import in `AlgoKShortestPaths.java`, and reworded `GraphData.degrees()`'s javadoc to describe both callers of `getDegrees` instead of asserting a caller count this PR itself invalidated.

lvca:
Fixed the correctness/robustness concern from the third review round in 4e0643afd:

**GraphData.weightedAdjacency now has its own incremental memory-budget reservation.** You're right that the round-1 disclosure understated this: it wasn't just that `algo.mst`/`algo.msa`'s refusal moved to "after the fact" - `weightedAdjacency` reserved nothing of its own at all, so the full `int[][]`/`double[][]` pair was unconditionally materialised before either procedure's own `reserve()` call ran, and `algo.msa` never had a `reserve()` call of its own to begin with (no protection whatsoever).

Implemented your option (b): `weightedAdjacencyFromColumns` and `weightedAdjacencyFromRecords` now reserve incrementally as they build - row headers up front, then per-edge-entry batches as the walk proceeds - mirroring `GraphData.adjacency()`'s existing `reserveAdjacency`/`ADJACENCY_CHECKPOINT_ENTRIES` pattern exactly. The null-weightProperty (unit weight) path also gains a reservation for its previously-unpriced weights array. So a graph too large for the weighted working set is now refused mid-walk, restoring (and for `algo.msa`, establishing for the first time) the guarantee #6300 set out for.

Updated `Issue6300AlgoMSTEdgeBudgetTest`'s two refusal tests for the new (earlier, more precise) message shape, and added `aLargeGraphIsRefusedPartwayThroughTheWalkNotAfterMaterialisingAllOfIt` (3000-node chain, `@Tag("performance")`) that directly covers the gap you called out - it proves the checkpoint fires mid-walk rather than after materialising the whole graph, measured via `ThreadMXBean` allocated-bytes rather than timing.

Minor nits also addressed:
- `parallelForRangeCheckpointed(0, ...)` never checking in: confirmed harmless as you noted (every caller returns early on `n == 0` before reaching a checkpointed loop) - not changing the contract for a case nothing currently exercises.
- The "CSR path below" comment in `executeWithOLTP` pointing at the wrong sibling method: acknowledged, cosmetic only, not touching in this round to avoid more diff noise on a comment-only nit; happy to fix in the same pass if you'd like it bundled.

lvca:
Also fixed the cosmetic comment nit in 8eb058b11 (bundled it in rather than deferring).

lvca:
Fixed the valid finding from the fourth review round in 4d70a38b8: **GraphData.degrees() now reserves before allocating**, both the result array and the previously-unpriced perType scratch buffer, matching the discipline the round-3 fix established for weightedAdjacency.

Not changing, per your own framing:
- The regex-based "self-guarded" detection in Issue6318AlgoGuardCoverageBackstopTest is a known, accepted blind spot (a pragmatic backstop, not a behavioral correctness guarantee) - you noted this yourself as "not asking for a change".
- AlgoDegreeCentrality's unused dirArg remains the same disclosed pre-existing follow-up from round 1/2.

lvca:
Fifth review round: no blocking issues found. The one point raised (`@Tag("performance")` vs `@Tag("slow")`/`@Tag("benchmark")`) - judged not worth changing in this PR:

Both `@Tag("performance")` uses in this PR (`Issue6265InfluenceMaximizationBufferReuseTest.simulateICNoLongerAllocatesPerCall`, `Issue6300AlgoMSTEdgeBudgetTest.aLargeGraphIsRefusedPartwayThroughTheWalkNotAfterMaterialisingAllOfIt`) measure allocation via `ThreadMXBean`, not wall-clock time - exactly the category `Issue6289AlgoAllocationChurnTest` (same package) documents as *deliberately* tagged `performance` rather than `slow`/`benchmark`, precisely so they keep running in the default CI lane: "These are not throughput measurements - they read ThreadMXBean's per-thread allocation counter, which no GC and no concurrent test can move - so they belong in the default lane." Tagging them `slow` instead would actively exclude fast, deterministic, non-flaky tests from the lane where they belong, and would diverge from the established local convention rather than follow it. The reviewer reached the same conclusion independently ("worth a follow-up to fix repo-wide rather than blocking this PR on it") - agreed, not changing here.

(The wall-clock-based benchmark I added, `Issue6318ParallelForRangeCheckpointedTest.checkpointedBatchingOverheadIsSmallRelativeToPageRankShapedWork`, is already tagged `@Tag("benchmark")`, correctly excluded from the default lane.)

All disclosed tradeoffs (MST/MSA's budget check ordering, `dirArg`) were independently re-verified and confirmed accurate by this round.

This satisfies all the stated merge conditions: the review found no blocking issues, and the one remaining point has been judged not worth implementing with reasoning above. PR: https://github.com/ArcadeData/arcadedb/pull/6714

lvca:
Also cleared Codacy's 3 static-analysis findings (unchanged since the first commit, easy to miss under the review conversation) in 98f92d19f: two field declarations moved to the top of `GraphAlgorithms` alongside the class's other constants, and `AlgoInfluenceMaximization.simulateIC`'s now-unused `n` parameter removed (dead since the buffers were hoisted to the caller in the #6265 fix).

lvca:
Fixed the round-6 finding in c9634e322: **algo.msa now reserves memory for its own eFrom/eTo/eW edge arrays**, on top of what weightedAdjacency already reserves for its own pair - mirroring algo.mst's equivalent reservation (16 bytes/edge, no sort scratch since Edmonds' algorithm doesn't sort edges by weight the way Kruskal's does). Added `Issue6300AlgoMSAEdgeBudgetTest` mirroring `Issue6300AlgoMSTEdgeBudgetTest`'s coverage for the sibling procedure.

lvca:
Addressed the round-7 finding in aa88b2cae:

**Columnar CSR checkpoint granularity (mst/msa/maxFlow/kShortestPaths losing per-edge throttling on a supernode)**: confirmed accurate - traced `GraphTraversalProvider#edgeWeightsOf`'s default implementation (its only implementation; no provider overrides it) and it's genuinely O(degree) with zero checkpoint hooks inside its own loop. This is a real, pre-existing property of `weightedAdjacencyFromColumns` shared with `algo.steinerTree`/`algo.bellmanFord` today - #6316 newly routes four more procedures onto it, which previously always had per-edge throttling via their own hand-rolled walks.

I didn't attempt the "prove abortability" test you suggested, because it wouldn't actually test what matters here: `WorkGuard.checkPeriodically(i)` fires unconditionally at `i == 0` (before any edge is read), so an interrupt armed before the call always aborts trivially at the first node regardless of whether a supernode is present - it would prove nothing about the specific gap (an *already-entered* `edgeWeightsOf` call on a supernode being unabortable). A test that actually exercises that would need to interrupt asynchronously mid-call on a large-enough supernode to reliably still be running when the interrupt lands - timing-sensitive in the way CLAUDE.md's stall-aware guidance steers away from.

A correct fix needs a checkpoint hook inside `edgeWeightsOf`'s own loop, which means widening the `GraphTraversalProvider` SPI itself (affecting every implementer, not just this PR's callers) - too large a change to fold into an already-large review-response round. Filed as **#6715** and documented in place (`weightedAdjacencyFromColumns`'s javadoc) instead, per your own offered resolution ("a short note if this tradeoff is intentionally accepted").

**`parallelForRangeCheckpointed(0, ...)` javadoc**: fixed to accurately describe the `n == 0` case (checkpoint never called, harmless today since every current caller already guards against it) instead of leaving the "still runs as one batch" claim technically wrong for that input.

**`dirArg`**: agreed, no change - same disclosed pre-existing item every round has confirmed.

---

Also: `python-e2e-tests` failed on this run (`test_asyncpg.py::test_parameterized_insert`, an `asyncpg` Postgres wire-protocol buffer-parsing `AssertionError`). Checked the failure - it's in the Postgres wire protocol module, which nothing in this PR's diff touches (this PR only changes `algo.*` procedures, `GraphAlgorithms`, and two SQL aggregate tests). Treating as unrelated to this PR; not investigating further here.

lvca:
Resolved both round-8 pre-merge items:

1. **algo.degree's dirArg tracking issue**: filed as **#6716**.
2. **weightedAdjacency per-edge-checkpoint tradeoff in the PR description**: added a "Known limitations" section to the description linking #6715 and #6716, so both are visible without having to find the code comments.

Also fixed the naming nit in 19562ce93: `LCC_PREP_CHECKPOINT_STRIDE` → `LCC_PREP_CHECKPOINT_MASK` with a one-line comment spelling out that `(u & MASK) == MASK` fires every 4096th node, not every `MASK`th one.

lvca:
Addressed both round-9 findings:

**1. labelPropagation's neighborBuf reallocation (68d6698c9)**: Confirmed and fixed. `neighborBuf` now lives in a `ThreadLocal`, sized once per thread and reused across every checkpoint batch and iteration of one call, instead of a fresh `new int[maxDeg]` per closure invocation.

I did not add a dedicated allocation-bytes benchmark for this specifically, and want to be upfront about why: I verified the fix's direction by directly comparing measured allocation with and without it on the same 20000-node/20-iteration graph - **247,472 bytes unfixed vs 243,848 bytes fixed**. Real and correctly-directioned, but modest, because dispatch overhead from repeated `parallelForRange` calls (`Future[]` arrays, submitted tasks - unavoidable given the batching design itself, and already covered qualitatively by the existing `checkpointedBatchingOverheadIsSmallRelativeToPageRankShapedWork` wall-clock benchmark) dominates the measurement for any realistic graph. Making `neighborBuf`'s contribution the dominant term in a byte-count assertion would need an impractically large max-degree graph, which would make the test slow and its bound arbitrary rather than meaningful. The fix's correctness instead rests on `ThreadLocal.get()` returning the same sized array across every batch/iteration once populated - a structural property, not something that needs a byte-count to prove. Existing `AlgoLabelPropagationTest`/`Issue6264AlgoIterationKnobGuardTest`/`Issue6302AlgoGraphDrivenWorkGuardTest` all still pass, confirming no behavior change.

**2. weightedAdjacencyFromColumns's memory-budget angle**: agreed this sharpens #6715's severity (OOM risk on a supernode, not only slow abort). Added this to the tracking issue rather than the PR, since it's the same root cause and fix already tracked there, not a new finding needing its own line in this PR's description.

lvca:
Addressed both round-10 points in 9b6deff13 - both were narrow edge cases the review itself framed as "worth a comment" rather than needing a behavior change, so documented in place: the OLTP fallback's `long`→`int` narrowing (matches the CSR-int-indexed contract every other `GraphData` method already has; a >2³¹-edge single vertex would hit other engine limits first) and the empty-materialised-types case (correctly all-zero, since `loadGraph`'s `findProvider` already requires the provider to cover every requested type before it's selected, so empty means "no edges exist" not "provider couldn't answer").

lvca:
Round 11 was substantial - three real findings fixed, several judged not worth acting on. Fixed in b4b3587b6:

**1. Checkpoint batch-size cap → unbounded latency for large n**: confirmed and fixed. Added `MAX_CHECKPOINT_BATCH_SIZE` so batch size stops growing past `CHECKPOINT_BATCHES x PARALLEL_THRESHOLD` (~131k) and batch count grows instead, keeping per-batch abort latency bounded regardless of graph size.

**2. `labelPropagation`'s `ThreadLocal` neighborBuf → unbounded retention on the shared pool**: agreed this is a real, worse-than-the-problem-it-solved tradeoff. Reverted to a fresh `neighborBuf` per chunk invocation. The round-9 savings were themselves modest (~1.5% of total allocation - I'd measured and disclosed this at the time), not worth trading for indefinite retention of a supernode-sized buffer on every pool thread.

**3. LCC prep-pass checkpoint granularity**: unlike #6715 (blocked on an opaque SPI call), this one is plain code I control, so I fixed it properly rather than just documenting it - added entry-count checkpointing inside the per-node edge walks (mirroring `GraphData.adjacency()`'s `ADJACENCY_CHECKPOINT_ENTRIES` pattern), so a single supernode row is no longer an unabortable unit. Also fixed the node-count stride (4095→1023) to match the 1024 convention used everywhere else in the codebase.

**Test nits**: tightened `GUARD_REFERENCE` to require call-site patterns, not bare token mentions; fixed the genuine `BiConsumer` FQN oversight; added the n==0 checkpoint-count test. Left `com.sun.management.ThreadMXBean` as FQN - verified it's the established, deliberate convention in this exact package (`Issue6289AlgoAllocationChurnTest`, unmodified by this PR, does the same and doesn't import it either).

**Judged not worth acting on:**
- **`@Tag("performance")` vs `@Tag("slow")`**: already litigated in round 5 with the same reasoning (these are `ThreadMXBean`-based, non-flaky, deliberately kept in the default lane per `Issue6289`'s own documented rationale) - not revisiting.
- **`AlgoDegreeCentrality`'s OLTP path going from lazy-under-LIMIT to eager**: real, but narrow (whole-graph degree centrality under a `LIMIT` was an incidental laziness benefit of the old `Stream`-based code, not a design goal) and a clean fix would need `degrees()` to support lazy per-vertex computation alongside the CSR path's inherently bulk one. Not fixing here.
- **`degrees()` reserving 4 arrays' worth of budget when at most 3 are live**: over-charging is conservative (a false-negative refusal, not an OOM risk) - safe, not worth precision-tuning.
- **`AlgoBetweenness`'s O(n²) `ArrayList<Integer>` allocation**: pre-existing, confirmed, filed as **#6718** rather than fixed inline (genuine algorithmic restructuring, not a quick change).
- **No placeholder test for `algo.degree`'s direction param**: tracked via #6716 already; the issue itself is the tracking mechanism.
- **"Unrelated" `Issue5414LastPointTest`/`SQLFunctionSumAverageMultiArgTest` changes**: these aren't unrelated - they're the #6705 fix, one of the four issues this PR's title and `Fixes` lines already name.

Full regression re-run: 9587 tests, 1 failure (`CypherGAVBoundTargetCardinalityTest`, confirmed pre-existing/order-dependent - passes in isolation, unrelated subsystem). `python-e2e-tests`' recurring failure is the same unrelated `asyncpg` issue noted in round 7.

lvca:
@claude This PR has been through 11 review rounds now, with genuine findings progressively fixed each time (memory-budget ordering, checkpoint abortability, buffer-reuse correctness, unbounded batch latency, etc. - see the commit history and comment thread above for the full trail). The last automated review round completed but posted no comment (looks like it hit its tool allowlist and got stuck rather than concluding). Could you do a fresh, focused pass and confirm: is this PR OK to merge, or is there anything outstanding? Please reply as a PR comment.

lvca:
Thanks for the fresh pass and the explicit merge verdict. Addressed both remaining items:

1. **weightProperty normalization in algo.mst/algo.msa**: fixed in f1b958127, matching the existing pattern in algo.maxFlow/algo.kShortestPaths.
2. **A few cheap O(n) loops with no checkpoint** (pageRank danglingSum, LCC offset-fill, LCC final materialization): filed as **#6719** per your own recommendation - fast follow-up rather than a 13th round on this branch.

This push will trigger one more CI/review cycle for the trivial fix above. Once that's green, all three of the stated merge conditions are met: explicit "OK to merge" verdict with no blocking issues, and every remaining point either fixed or tracked with reasoning. PR: https://github.com/ArcadeData/arcadedb/pull/6714

lvca:
Round 13 addressed - both coverage gaps closed, not just the first:

1. **CSR columnar path budget refusal** (`Issue6300AlgoMSTEdgeBudgetTest`/`Issue6300AlgoMSAEdgeBudgetTest`): added `theWeightedAdjacencyRowHeadersAreRefusedOnTheCSRColumnarPathToo` to each, invoking the procedure directly against a `GraphAnalyticalView`-backed context and asserting both the refusal and `CSR_ACCELERATED_VAR`. This exercises `weightedAdjacencyFromColumns` specifically - the path repeatedly patched for reservation-ordering bugs earlier in this same review - which the existing tests never touched.

2. **Parallel-edge/self-loop equivalence** (`Issue6316AlgoLoadGraphAdoptionTest`): added two more OLTP-vs-CSR equivalence tests (degree centrality, algo.mst) against a graph with a parallel edge and a self-loop. On reflection this was worth doing rather than deferring - it's exactly the shared-plumbing surface (`GraphData.degrees`, `weightedAdjacency`) that already had three real bugs found in it this review cycle, and both edge shapes passed cleanly, confirming no fourth one.

Also merged `origin/main` (2 unrelated commits, security fix) - no conflicts.

All affected tests green: `Issue6300AlgoMSTEdgeBudgetTest` (7), `Issue6300AlgoMSAEdgeBudgetTest` (4), `Issue6316AlgoLoadGraphAdoptionTest` (8), plus the full `algo`/`graph.olap` package sweep (720 tests, only the pre-existing unrelated `Issue6301AlgoSteinerTreeWeightAlignmentTest` order-dependent flake, confirmed failing in isolation on this same commit and previously confirmed failing on clean main too).

lvca:
Round 14 addressed:

1. **algo.mst/algo.msa's edge-array flattening copy had no checkpoint**: fixed in 789811999. Added the same throttled `guard.checkPeriodically()` the old pass-2 fill loop had, in both `AlgoMST`'s `eu`/`ev`/`ew` copy and `AlgoMinSpanningArborescence`'s `eFrom`/`eTo`/`eW` copy.

**Judged not worth acting on:**
- **`AlgoDegreeCentrality` computes both IN/OUT unconditionally**: pre-existing, already tracked as #6716 (confirmed by the review itself - "no action needed here").
- **`@Tag("performance")` not in CI's excluded-groups list**: matches the existing local convention (`Issue6289AlgoAllocationChurnTest`, `Issue6296AlgoAPSPLazyRowsTest`), deliberately not excluded since these use `ThreadMXBean` allocation counters rather than wall-clock and are bounded/fast - the review itself flagged this as "presumably intentional and fine."

All affected tests green (full `algo`/`graph.olap` package sweep).

@claude this round's review didn't use the "OK to merge" phrasing explicitly - could you confirm one more time whether this PR is mergeable as of 789811999, with no other outstanding blockers?

lvca:
CI's `unit-tests` lane on `7898119` failed with 3 test failures. Investigated each:

1. **`Issue6265InfluenceMaximizationBufferReuseTest.simulateICNoLongerAllocatesPerCall`** (this PR's own test): failed at allocated=1,443,080 bytes against a 1% (1,345,500-byte) bound - ~7% over, but still ~93x below the ~134MB unfixed-behaviour baseline it exists to catch. Reproduced 5/5 passing locally under the same bound (values consistently under 1.3455M), so this is CI-environment allocation noise on an overly tight margin, not a functional regression. Fixed in 4494543a0 by widening the bound to 5% (still ~20x separation from the pathological case).

2. **`CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair`**: the same pre-existing, order-dependent flake already documented multiple times earlier in this review (passes in isolation, unrelated subsystem, untouched by this diff). Reconfirmed passing in isolation. No action.

3. **`GraphAnalyticalViewTest.issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore`**: new to this run, but this diff never touches `GraphAnalyticalView.java`/`CSRBuilder.java` (only `GraphAlgorithms.java` in that package). The test asserts a virtual-thread executor flips status synchronously "before the background work even starts" - reproduced 4/4 passing locally, both in isolation and running the whole 156-test class, consistent with scheduler-congestion flakiness under full-suite load rather than anything this PR introduced. No action.

Also merged `origin/main` (1 unrelated docs-only commit) - no conflicts.

lvca:
Round 16 addressed:

1. **Comments citing "review round N"**: fixed in bcfeeca71. Trimmed the round-number provenance from production-code comments in `GraphAlgorithms.java`, `AbstractAlgoProcedure.java`, `AlgoMST.java`, and `AlgoMinSpanningArborescence.java`, restating each invariant/reasoning to stand on its own post-merge. No behavior change (comment-only diff, full `algo`/`graph.olap` regression sweep green).

**Judged not worth acting on** (both correctly flagged as already tracked, per the review itself):
- **`AlgoDegreeCentrality`'s unused `direction` param**: pre-existing, tracked as #6716.
- **`weightedAdjacencyFromColumns`'s per-node (not per-edge) checkpoint granularity**: pre-existing, tracked as #6715 (needs a `GraphTraversalProvider` SPI change).

All other points in this review were confirmations, not findings requiring action.

lvca:
Round 17 (deep pass): no blocking findings. Two minor items, both non-blocking per the review itself:

1. **`weightedAdjacencyFromColumns` checkpoints per node, not per edge**: already tracked as #6715 (needs a `GraphTraversalProvider` SPI change). No action.
2. **Redundant O(n) edge-count resum**: `weightedAdjacency`'s unit-weight branch, plus `algo.mst`/`algo.msa` each independently re-summing `neighbors[i].length` after it returns - the same node-count-sized total computed 2-3 times. Cheap in isolation (not an edge walk, just `.length` reads) and not a correctness issue, but real duplicated work. Filed as **#6721** rather than expanding this PR's scope further, following the same pattern as #6715/#6716/#6718/#6719.

CI on `bcfeeca71`: everything green except the confirmed unrelated `python-e2e-tests` asyncpg flake; `ha-integration-tests`, `integration-tests`, `slow-unit-tests`, `unit-tests`, `vector-unit-tests` still finishing.

lvca:
Merge conflict with main resolved (commit a961b34fd). PR #6717 merged to main while this PR was in review, fixing #6715 and #6716 directly against the pre-#6714 code - both conflicted with this PR's own restructuring of the same methods.

Resolved by combining both:
- **`weightedAdjacencyFromColumns`**: took #6717's version wholesale (threads `guard::checkPeriodically` into `edgeWeightsOf` itself, plus a `memory.capacityFor`-based dynamic checkpoint interval) - a strict improvement over what this PR had, and actually **resolves the #6715 gap this PR had repeatedly documented as deliberately out of scope**. Also removed a duplicate `reserveWeightedAdjacency` definition both branches had independently added.
- **`AlgoDegreeCentrality`**: kept this PR's `GraphData.degrees()` routing (CSR acceleration, #6316) but now only computes the requested direction(s) per #6717's fix (#6716), instead of always both.

Compiled clean; `algo`/`graph.olap` regression sweep green (729 tests).

Also investigated the `unit-tests` CI failure on the prior commit (`bcfeeca71`, before this merge): it was solely `GraphAnalyticalViewTest.issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore` - the same pre-existing, order-dependent flake already confirmed unrelated in round 16 (passes 4/4 in isolation, untouched by this diff). No action needed there.

lvca:
Round 18: no blocking issues - "Nothing here blocks merge." Verified clean: CSR-vs-OLTP equivalence, checkpoint batching math, `simulateIC` buffer reuse, and memory-reservation ordering across every touched allocation site.

One non-blocking nit: `@Tag("performance")` isn't one of CLAUDE.md's three CI-recognized lanes (`benchmark`/`slow`/`vector`). This has come up in earlier rounds too (round 5, and again in round 16/17's discussion) - it matches an existing repo-wide pattern (`Issue6289AlgoAllocationChurnTest`, `RidScoreMinHeapTest`, etc.), not something this PR introduces, and the review itself frames it as "worth a follow-up cleanup across the repo rather than blocking this PR on it." Not acting on it here, consistent with the prior rounds' judgment.

CI on `a961b34fd`: `python-e2e-tests` failed on the confirmed known asyncpg flake; `builder-tests`, `ha-integration-tests`, `integration-tests`, `java-e2e-tests`, `slow-unit-tests`, `studio-e2e-tests`, `unit-tests`, `vector-unit-tests` still finishing.

lvca:
unit-tests failed again on commit a961b34fd - investigated, all 3 failures confirmed pre-existing/unrelated:

1. **`GraphAnalyticalViewTest.issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore`**: the same flake already confirmed unrelated in round 16 (untouched file, unrelated feature #6641).
2. **`GraphAnalyticalViewTest.issue6641_corruptedPersistedCsrInvalidatesDeferredRestoreWithoutBlockingQuery`**: new occurrence, same file, same unrelated #6641 feature area. Reproduced 3/3 passing in isolation running the whole 156-test class.
3. **`CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair`**: the already-documented pre-existing order-dependent flake from earlier rounds.

None of these three files are touched by this PR's diff. No action needed.

lvca:
ha-integration-tests failed on `RaftHARandomCrashIT.replicationWithRandomCrashes` - `DatabaseComparator$DatabaseAreNotIdentical` after random crash injection during Raft replication. This is a randomized chaos test (crash injection timing varies by design) in the HA/Raft clustering subsystem - completely untouched by this PR's diff (confirmed: no `ha`/`raft`/`server` files appear in `git diff origin/main...HEAD`). Matches the documented chronic `ha-integration-tests` flakiness pattern for this lane. No action needed.

With this, all CI lanes are now either green or resolved to confirmed pre-existing unrelated flakes:
- `python-e2e-tests`: known asyncpg flake
- `unit-tests`: `GraphAnalyticalViewTest` (x2, unrelated feature #6641) + `CypherGAVBoundTargetCardinalityTest` flakes
- `ha-integration-tests`: `RaftHARandomCrashIT` chaos-test flake

Combined with round 18's explicit "Nothing here blocks merge" verdict and no outstanding review points, this PR looks ready to merge once `coverage-report` (the one remaining non-blocking check) finishes.

Issue #6716, opened 2026-08-25
algo.degree(relTypes?, direction?) parses direction but ignores it — always computes full IN+OUT regardless


lvca:
Raised in review of PR #6714 (issue #6316) round 1 and confirmed again in round 8, deliberately not folded into that PR: it's a pre-existing bug (predates #6316's conversion of `algo.degree` onto `loadGraph`/`GraphData`), not a regression from it.

## Where

`AlgoDegreeCentrality.execute()`:

```java
final String dirArg = args.length > 1 ? extractString(args[1], "direction") : "BOTH";
```

`dirArg` is parsed and validated but never referenced again in the method body. `inDegree`/`outDegree`/`degree`/`score` are always computed as full IN+OUT (via two `GraphData.degrees()` calls, one per direction) regardless of what the caller passes.

## Impact

A caller doing `algo.degree(null, 'OUT')` expecting only out-degree gets both directions summed anyway, silently. The procedure's own doc comment (`Procedure: algo.degree(relTypes?, direction?)`) advertises a `direction` parameter with no caveat that it currently has no effect.

## Suggested fix

Either wire `dirArg` up properly (compute only the requested direction(s) via `GraphData.degrees(Vertex.DIRECTION.valueOf(dirArg), relTypes)` when a single direction is requested, keeping today's IN+OUT behavior for `"BOTH"`/unset), or - if the two-directions-always semantics turn out to be intentional for some caller - drop the parameter and its parsing, and remove the misleading `direction?` from the procedure's documented signature.

## Related

#6316, PR #6714.

lvca:
Fixed in #6717 (merged).

`algo.degree(relTypes?, direction?)` accepted a `direction` parameter but silently ignored it, always returning the combined IN+OUT degree. Now passing `'IN'` or `'OUT'` actually limits the result to that direction; leaving it unset (or `'BOTH'`) keeps the previous combined behavior.

---
❤️ Thanks for helping make ArcadeDB better! ArcadeDB is free and open source, sustained by its community. If it's useful to you or your company, please consider [becoming a sponsor](https://github.com/sponsors/ArcadeData) to keep development going.

lvca:
Closing - already fixed and merged in #6717 (commit 22f2e3a7d, merged as 083eab63c on main). `algo.degree`'s `direction` parameter is now honoured.

PR #6717, opened 2026-08-25
fix(#6715, #6716): per-edge checkpoint in edgeWeightsOf, honour algo.degree's direction param

### LINKED ISSUE #6715: GraphData.weightedAdjacencyFromColumns's per-node WorkGuard checkpoint cannot abort mid-supernode (provider.edgeWeightsOf has no interrupt hook)
Raised in review of PR #6714 (issues #6316/#6265/#6318/#6705) and deliberately not folded into it: it names a real, pre-existing architectural gap that predates that PR and would need an SPI-level change larger than a review-response round should attempt.

## The gap

`AbstractAlgoProcedure.GraphData.weightedAdjacencyFromColumns` throttles its `WorkGuard` per **node**:

```java
for (int i = 0; i < nodeCount; i++) {
  guard.checkPeriodically(i);
  final NodeEdgeWeights edges = provider.edgeWeightsOf(i, dir, weightProperty, 1.0, types);
  ...
}
```

`GraphTraversalProvider#edgeWeightsOf`'s default implementation (the only one; `GraphAnalyticalView` does not override it) is O(degree) with no checkpoint of its own:

```java
default NodeEdgeWeights edgeWeightsOf(...) {
  ...
  for (final String type : types)
    for (final Vertex.DIRECTION d : directions) {
      final int[] slice = getNeighborIds(nodeId, d, type);
      ...
      for (int j = 0; j < slice.length; j++) {
        // no interrupt/timeout check anywhere in this loop
      }
    }
}
```

So a single call for a supernode (a vertex with a very large fan-out) is one unabortable unit of work: once entered, nothing can stop it until it returns, regardless of `arcadedb.command.timeout` or a thread interrupt.

## Why it's newly relevant

The helper itself is not new - `algo.steinerTree` and `algo.bellmanFord` already read weights through it. What #6316 changed is that four more procedures (`algo.mst`, `algo.msa`, `algo.maxFlow`, `algo.kShortestPaths`) can now reach this same columnar path when a Graph Analytical View has the weight property materialized as a column, where previously they always used a hand-rolled OLTP walk with genuinely **per-edge** `WorkGuard` throttling (`guard.checkPeriodically(edgeStep++)`, one increment per edge - visible in the pre-#6316 version of e.g. `AlgoMST`, which had a comment specifically about this: *"Throttled by EDGE rather than by vertex: a near-star graph puts millions of edges inside a single vertex iteration, and a per-vertex checkpoint would leave that whole node unabortable"*). That protection is preserved in `weightedAdjacencyFromRecords` (the record-based fallback, still per-edge) but not in the columnar path six procedures now share.

## Why not fixed in #6714

A correct fix needs a checkpoint hook inside `provider.edgeWeightsOf`'s own O(degree) loop, which means either:
- widening `GraphTraversalProvider`'s SPI to accept a `WorkCheckpoint`/similar and pushing the per-edge check into every implementer (the interface's only implementation today is the default method, but it's still an SPI contract change), or
- restructuring the columnar build to pull neighbour/weight slices in bounded chunks rather than one array-returning call per node.

Either is a real design decision (chunk size, whether the check belongs in the SPI vs. the caller, whether it costs anything on the hot path for the common non-supernode case) that deserves its own review, not a fix bundled into an already-large PR whose own scope is CSR routing and buffer reuse.

## Suggested shape

Mirror what issue #6318 did for `GraphAlgorithms.localClusteringCoefficient`'s two parallel phases: give `edgeWeightsOf` (or a new bulk variant of it) a `WorkCheckpoint`-shaped hook, called at a bounded stride inside its per-edge loop, and thread a checkpoint from `weightedAdjacencyFromColumns` through it. That keeps the "checkpoint stays on a single logical thread of control, called between chunks" contract #6318 established, applied one level deeper.

## Related

#6316, #6318, #6301, PR #6714.

lvca:
## Summary

Fixes #6715 and #6716.

**#6715** - `AbstractAlgoProcedure.GraphData.weightedAdjacencyFromColumns` checkpointed its `WorkGuard` once per node, but `GraphTraversalProvider#edgeWeightsOf`'s O(degree) inner loop had no checkpoint of its own. A single supernode's row was therefore one unabortable unit of work, whatever `arcadedb.command.timeout` or a thread interrupt said.

- Adds an `edgeWeightsOf` overload taking a per-edge checkpoint callback (`IntConsumer`), called once per edge inside the walk. The counter restarts at zero on every call - the same convention `WorkGuard`'s own javadoc documents for a restarting loop counter - which bounds the worst-case latency to about one checkpoint stride into the largest node's row, independent of the caller's own per-node count. This is additive: the existing 5-arg method delegates to the new one with a `null` checkpoint, so `SQLFunctionAstar` and `SQLFunctionBellmanFord` (the other two call sites) are unaffected.
- `weightedAdjacencyFromColumns` now passes `guard::checkPeriodically` as that checkpoint.
- Also closes the memory side the issue's own follow-up comment named: the columnar build never reserved anything against `CYPHER_ALGO_MAX_WORKING_MEMORY` before this change. It now gets the same cumulative, entries-plus-node-count-checkpointed reservation the unweighted `adjacency()` build already has. A single node whose row alone blows the budget is still refused only after that one row is built (closing this fully would mean handing the `GraphTraversalProvider` SPI a dependency on `AbstractAlgoProcedure.MemoryBudget`, which felt like the wrong tradeoff for this fix - flagging in case there's a different call here).

**#6716** - `AlgoDegreeCentrality` parsed its `direction` argument but never used it: `degree`/`score` were always the full IN+OUT total regardless of what was requested. Now only the requested direction(s) are counted (and the other of `inDegree`/`outDegree` is 0, since it's no longer read), keeping today's IN+OUT behavior for `BOTH`/unset.

## Test plan
- [x] `Issue6715EdgeWeightsOfCheckpointTest`: a pre-armed self-interrupting stub provider proves the walk now aborts within ~1 checkpoint stride of a supernode row rather than building the whole thing (was: ran to completion); a cumulative multi-node budget test proves many medium rows are now refused (was: never reserved anything at all); counterweights confirm a healthy run is unaffected.
- [x] `AlgoDegreeCentralityTest#degreeDirectionParameterIsHonoured`: a node with distinct in/out degree confirms `BOTH`/`IN`/`OUT`/unset each answer correctly.
- [x] `mvn -o -pl engine -am test` over the full `algo` procedure package, `GraphTraversalProvider`-adjacent tests, and the affected `function.sql.graph` tests: all green, no regressions.

## Summary by CodeRabbit

* **New Features**
  * Added direction-aware degree calculations for incoming, outgoing, both, and default directions.
  * Added optional progress checkpoints for weighted graph traversals.

* **Bug Fixes**
  * Corrected direction-specific degree results while preserving relationship-type filtering.
  * Improved handling of large weighted traversals with interruption checks and memory-limit enforcement.

* **Tests**
  * Added coverage for degree directions, traversal interruption, progress tracking, and cumulative memory handling.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -293,12 +293,19 @@ public interface GraphTraversalProvider {
    * A caller that does not visit many nodes in a row - a single-source search that reads one node's edges per
    * step, already bounded by its own outer loop - passes {@code null} and pays nothing beyond one comparison per
    * edge.
+   * <p>
+   * {@code edgeTypes} is a plain array here rather than the other overload's varargs, deliberately: a fifth
+   * argument of a bare {@code null} would otherwise be ambiguous between binding to {@code edgeCheckpoint} here
+   * and to the other overload's {@code String... edgeTypes} (both are reference types, and neither overload is
+   * more specific than the other from a {@code null} literal), breaking any existing five-argument call this
+   * SPI already has. Requiring six arguments to reach this overload at all keeps every four- and five-argument
+   * call - including a bare {@code null} - resolving to the other one, unchanged.
    *
    * @param edgeCheckpoint called with the edge index within this call (0-based), or {@code null} to skip it
    */
   default NodeEdgeWeights edgeWeightsOf(final int nodeId, final Vertex.DIRECTION direction,
       final String propertyName, final double defaultWeight, final IntConsumer edgeCheckpoint,
-      final String... edgeTypes) {
+      final String[] edgeTypes) {
     if (!servesEdgeProperty(propertyName, edgeTypes))
       return null;
 
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -20,6 +20,8 @@ package com.arcadedb.graph;
 
 import com.arcadedb.database.RID;
 
+import java.util.function.IntConsumer;
+
 /**
  * SPI for accelerated graph traversal providers (e.g., Graph Analytical Views backed by CSR).
  * <p>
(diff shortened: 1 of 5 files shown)


lvca:
Checked the CI failures on this PR against unmodified \`main\` (4b107cf4c) - all confirmed pre-existing and unrelated to this change (nothing here touches timeseries, SQL math functions, GraphAnalyticalView's async restore, or the Postgres wire protocol):

- \`unit-tests\`: 4 failures, none in files this PR touches.
  - \`GraphAnalyticalViewTest.issue6641_corruptedPersistedCsrInvalidatesDeferredRestoreWithoutBlockingQuery\` / \`issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore\`: pass locally (0 failures), so a CI-environment timing flake in the async CSR build/restore path.
  - \`Issue5414LastPointTest.descendingScanOnAnEmptySeriesReturnsNothing\` and \`SQLFunctionSumAverageMultiArgTest.avgDoesNotTruncate\`: reproduce deterministically on unmodified \`main\` HEAD too (ran locally against 4b107cf4c) - pre-existing bugs in the timeseries and SQL math-function packages.
- \`python-e2e-tests\`: \`test_asyncpg::test_parameterized_insert\` fails the same way on the latest \`main\` CI runs (checked several recent runs) - a pre-existing Postgres wire-protocol/asyncpg buffer issue.

None are in \`GraphTraversalProvider\`, \`AbstractAlgoProcedure\`, or \`AlgoDegreeCentrality\` (the only files this PR changes), and the algo procedure test suite itself is fully green.

lvca:
Update: \`ha-integration-tests\` also failed - \`RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges\` errored with \`ReplicationDispatchedTimeoutException: Group commit entry failed: TimeoutException (entry was dispatched to Raft; outcome unknown)\`. This is a timing-sensitive Raft/HA clustering integration test (1 error out of 238 tests in that lane) - nothing in this PR touches HA/Raft code (\`com.arcadedb.server.ha.*\`), so it's unrelated to this change.

All CI failures on this run are now accounted for and confirmed unrelated to the diff. Still waiting on \`slow-unit-tests\` to finish.

PR #6731, opened 2026-08-25
fix(vector): materialise the LSM_VECTOR location map on first use, not on every open (#6722)

### LINKED ISSUE #6722: LSM_VECTOR: every database open parses every index page to rebuild the in-memory location map, whether or not the session searches (1.4 s at 10M vectors)
Opening a database walks every page of every `LSM_VECTOR` index and rebuilds the in-memory location map, whether or not the session will ever run a vector search. The cost is proportional to the number of indexed vectors, it is paid again on every open, and the page cache does not amortise it because the work is parsing and map population rather than I/O.

At 10M vectors that is ~1.4 s of open before the caller does anything. A process that opens the database to write one document, or to read an unrelated type, pays it in full.

## Where

`LSMVectorIndexMutable.onAfterSchemaLoad()` calls `loadVectorsAfterSchemaLoad()` (`LSMVectorIndex.java:1052`), which calls `loadVectorsFromPages()` (:1092). That parses every page of the mutable file and of the compacted sub-index and calls `vectorIndex.addOrUpdate(...)` once per entry (`loadVectorsFromFile`, :3476-3480), producing a `VectorLocationIndex` with one entry per vector.

The interesting part is the line immediately after the call:

```java
loadVectorsFromPages();

// Graph will be lazy-loaded on first search via ensureGraphAvailable()
// Don't build it here - causes deadlock during database load when PageManager isn't fully ready
```

The graph is deliberately deferred. The location index, right above it, is not. So after a reopen the index reports `totalVectors` = the whole corpus while `graphState = LOADING(0)` and `graphNodeCount = 0`. The expensive structure is lazy and the O(N) one is eager, which is what makes this a choice rather than a necessity.

## What it costs

Measured through the embedded API on an i9-12900HK pinned to 12 P-threads, ext4, 64 dimensions, one open of a quiesced database with no query issued. Medians; n=5 at 10k and 100k, n=3 at 1M, n=1 at 10M.

| vectors | open, warm | open, page cache evicted | what the cache buys |
|---:|---:|---:|---:|
| 10,000 | 2.9 ms | 5.7 ms | 49% |
| 100,000 | 10.2 ms | 11.6 ms | 12% |
| 1,000,000 | 90.3 ms | 93.4 ms | 3% |
| 10,000,000 | 1,378.8 ms | 1,495.3 ms | 8% |

The cold column is a verified page-cache eviction before the open, so cold-vs-warm separates I/O from CPU. Every other workload we measure the same way gets 40-55% back from the cache and opens in single-digit milliseconds at every size:

| workload at 10M | open, warm | open, cache evicted | what the cache buys |
|---|---:|---:|---:|
| documents | 1.1 ms | 2.5 ms | 53% |
| graph + Graph Analytical View | 1.3 ms | 11.8 ms | 89% |
| vector | 1,378.8 ms | 1,495.3 ms | 8% |

The GAV row is the shape a deferred structure produces: pay disk once on the cold open, then nothing. The vector row is the shape of work redone in-process every time.

## Repro

`VectorOpenScansEveryIndexPageTest` below, run against `e215d61c1`. It builds an index, closes, then reopens three times and reports the median open. It asserts on the margin above a near-empty baseline and on a ratio rather than a wall clock, so it cannot go red on a slow machine: a constant-time open would show ~1x for 4x the data.

```
open, no query issued (median of 3 reopens each)
       2,000 vectors ->    6 ms   (baseline, subtracted below)
      50,000 vectors ->   18 ms   marginal    12 ms   0.250 us/vector
     200,000 vectors ->   83 ms   marginal    77 ms   0.389 us/vector
  marginal ratio 6.4x for 4x the vectors
```

0.25-0.39 us per vector extrapolates to 2.5-3.9 s at 10M on this laptop, and the 1.4 s measured on the faster bench host is consistent with that.

The test also asserts, on every reopen, that `totalVectors == N` while `graphState == LOADING` and `graphNodeCount == 0`, so the measured cost is attributable to the location scan and not to the graph.

```java
@Tag("vector")
class VectorOpenScansEveryIndexPageTest {
  private static final int    DIMENSIONS = 64;
  private static final int    BASE       = 2_000;
  private static final int    SMALL      = 50_000;
  private static final int    LARGE      = 200_000;   // 4x SMALL
  private static final int    REOPENS    = 3;
  private static final String DB_PATH    = "./target/databases/VectorOpenScansEveryIndexPageTest";

  @AfterEach
  void cleanUp() {
    FileUtils.deleteRecursively(new File(DB_PATH));
  }

  @Test
  void openingWithoutSearchingStillMaterialisesEveryVectorLocation() {
    final long baseMs  = medianReopenMs(BASE);
    final long smallMs = medianReopenMs(SMALL);
    final long largeMs = medianReopenMs(LARGE);

    final double smallMarginal = Math.max(0.0, smallMs - baseMs);
    final double largeMarginal = Math.max(0.0, largeMs - baseMs);
    final double ratio = largeMarginal / Math.max(1.0, smallMarginal);

    System.out.printf("open, no query issued (median of %d reopens each)%n", REOPENS);
    System.out.printf("  %,10d vectors -> %4d ms   (baseline, subtracted below)%n", BASE, baseMs);
    System.out.printf("  %,10d vectors -> %4d ms   marginal %5.0f ms   %.3f us/vector%n",
        SMALL, smallMs, smallMarginal, smallMarginal * 1000.0 / (SMALL - BASE));
    System.out.printf("  %,10d vectors -> %4d ms   marginal %5.0f ms   %.3f us/vector%n",
        LARGE, largeMs, largeMarginal, largeMarginal * 1000.0 / (LARGE - BASE));
    System.out.printf("  marginal ratio %.1fx for 4x the vectors%n", ratio);

    assertThat(ratio)
        .as("open cost above baseline is proportional to the number of indexed vectors, for a "
            + "session that issues no query")
        .isGreaterThanOrEqualTo(2.5);
  }

  private long medianReopenMs(final int numVectors) {
    FileUtils.deleteRecursively(new File(DB_PATH));

    try (final DatabaseFactory factory = new DatabaseFactory(DB_PATH)) {
      final Database db = factory.create();
      final Random rng = new Random(11);
      db.transaction(() -> {
        final DocumentType docType = db.getSchema().createDocumentType("Doc");
        docType.createProperty("id", Type.INTEGER);
        docType.createProperty("embedding", Type.ARRAY_OF_FLOATS);
        for (int i = 0; i < numVectors; i++)
          db.newDocument("Doc").set("id", i).set("embedding", randomVector(rng)).save();
      });
      db.command("sql", "CREATE INDEX ON Doc (embedding) LSM_VECTOR METADATA "
          + "{ \"dimensions\": " + DIMENSIONS + ", \"similarity\": \"EUCLIDEAN\", "
          + "\"storeVectorsInGraph\": false, \"addHierarchy\": true }");
      db.close();
    }

    final List<Long> opens = new ArrayList<>(REOPENS);
    for (int cycle = 0; cycle < REOPENS; cycle++) {
      try (final DatabaseFactory factory = new DatabaseFactory(DB_PATH)) {
        final Database db = factory.open();
        opens.add((System.nanoTime() - t0) / 1_000_000L);

        final TypeIndex idx = (TypeIndex) db.getSchema().getIndexByName("Doc[embedding]");
        final Map<String, Long> stats = ((LSMVectorIndex) idx.getIndexesOnBuckets()[0]).getStats();

        assertThat(stats.get("totalVectors")).isEqualTo((long) numVectors);
        assertThat(stats.get("graphState")).isEqualTo(0L);      // LOADING
        assertThat(stats.get("graphNodeCount")).isEqualTo(0L);

        db.close();
      }
    }
    opens.sort(Long::compare);
    return opens.get(opens.size() / 2);
  }

  private static float[] randomVector(final Random rng) {
    final float[] v = new float[DIMENSIONS];
    for (int i = 0; i < DIMENSIONS; i++)
      v[i] = (float) rng.nextGaussian();
    return v;
  }
}
```

## What we would expect

The same treatment the graph already gets at the same call site: populate the location index on first use rather than on schema load, so a session that never touches the vector index never pays for it. Persisting the location map alongside the graph and loading it as a block would also remove the per-entry parse, but deferring is the smaller change and matches the existing pattern two lines below.

Happy to open a PR if you would like it in either direction.

lvca:
Closes #6722.

## The problem

Opening a database walked every page of every `LSM_VECTOR` index and rebuilt the in-memory location map, whether or not the session went on to search it. The cost is one page parse plus one map insert per indexed vector, so it scales with the corpus rather than with what the session does; it is paid again on every open; and the page cache cannot amortise it, because the work is parsing and map population rather than I/O. The reporter measured ~1.4 s at 10M vectors, of which a cold open recovers only 8% - against 53% for documents and 89% for a graph with a Graph Analytical View, both of which open in single-digit milliseconds at that size. A process that opened the database to write one document, or to read an unrelated type, paid all of it.

The tell is the line immediately after the call: the graph, which is by far the more expensive of the two structures, is already deferred to the first search at that very call site. The O(N) one above it was not.

## The fix

`loadVectorsAfterSchemaLoad()` now records that there are pages worth reading and returns. The first caller that actually needs a location materialises them, once, under a dedicated lock (deliberately not the index-wide write lock: the parse is O(index size) page reads and holding that lock across it would stall every reader and writer, the same stall #5391 moved the graph-build validation off it for). Open becomes constant-time in the number of indexed vectors.

The deferral is structural rather than a convention. The field is private and reachable only through an accessor that materialises first, so a new call site cannot silently bypass it; the handful of sites that read it raw say why they may. Two of those matter to behaviour:

- **`needsGraphBuild()`** runs on every `flush()`, including the close of a session that never touched the index. It now answers "does this index hold vectors" from page counters rather than from the location count, so the deferral is not simply handed back at close. The branches *past* that decision do materialise: both are choosing between building the whole graph here and recording a deferral, both need the real live count, and getting it wrong would send a large index down the synchronous-build arm that #6067 exists to avoid. The parse is a rounding error against the build it is deciding about.
- **Vector ids** come from a sequence the load derives from the pages, so every allocation goes through `allocateVectorId()`, which materialises first. Without that gate an insert into a reopened index restarts the sequence at 0 and silently supersedes the entries it collides with - the index loses a vector per insert instead of gaining one. `insertAfterReopenDoesNotReuseVectorIds` is the regression test for exactly that, and it caught this during development.

The PQ codebooks load at the same call site and are deferred with the locations, since the ordinals a PQ search resolves are the ones the location map hands out and the two are only ever useful as a pair. `findNeighborsFromVectorApproximate()` checks PQ availability *before* it reaches the lazy-load, so it now asks through `isPQSearchAvailable()` (which materialises) instead of reading the two fields directly - otherwise the first approximate search after a reopen would silently downgrade itself to an exact one. `productQuantizationIsStillAvailableAfterReopen` covers that, and fails without the change.

The O(N) walks over the location index hoist the accessor into a local. That removes a per-element flag check, and it also gives each walk one consistent generation of the instance a compaction swaps (#5568) instead of re-reading the volatile field per element.

Nothing a caller observes changes, only when the work happens. `countEntries()`, `getStats()`, a search, an insert and a replicated page update all still see the whole corpus; the first of them to ask is the one that pays.

## On the alternative the issue offers

The issue also floats persisting the location map alongside the graph and loading it as a block. That is worth doing eventually, but it is not a substitute for this: it converts an O(N) parse into an O(N) block read, still paid on every open, so it does not help the session that never searches - which is the case being complained about. It also brings a new on-disk artefact with its own crash-consistency, staleness-invalidation and HA-shipping surface.

The genuinely best end state is both: with the deferral in place, persisting the map would turn the *first search's* cost from a per-entry varint parse into a block read of two primitive arrays, and the graph file already carries a manifest with a fingerprint that could vouch for it. But that first search already loads or rebuilds the graph, which dominates by orders of magnitude, so the marginal win is small and it deserves its own change rather than riding along with this one.

One consequence worth naming: a monitoring endpoint that scrapes `getStats()` right after startup will be the caller that pays. That is correct - it asked for the corpus size - and it happens once.

## Tests

New `LSMVectorIndexLazyLocationLoadTest` (7 tests), asserting the structural property rather than a wall clock, per the repo's rule against wall-clock bounds:

- open does not materialise the map (flag false, resident size 0)
- `countEntries()` as a first use materialises it and reports the whole corpus
- `getStats()` reports `totalVectors`/`activeVectors` correctly after a reopen
- a search after a reopen returns the same RIDs, in the same order, as before the close
- an insert after a reopen adds a vector rather than overwriting one, and survives another round trip
- three open/close cycles that never touch the index leave it intact
- a PRODUCT-quantized index still reports PQ available, with the right count, after a reopen

Three existing tests that reach the field reflectively were updated for the rename.

Full engine unit lane: 13,573 tests, 1 failure - `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair`, which passes in isolation on this branch and on `main`, and is a Cypher Graph Analytical View test that touches no vector code. `Issue6657CloseTimeRebuildPendingStatTest` also fails identically on `main` at the merge base and is unrelated.

## Summary by CodeRabbit

* **Performance**
  * Improved vector-index startup times by deferring index data loading until vector functionality is used.
  * Databases that do not perform vector searches now avoid unnecessary index processing during startup.

* **Bug Fixes**
  * Preserved vector search, statistics, insertion, replication, and ID behavior after deferred loading.
  * Ensured approximate vector search remains available when needed.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -3,6 +3,44 @@
 This is a living document: fixes, improvements, new features, and breaking changes are collected here as
 they land during the 26.9.1 development cycle, so the release notes are ready at tag time.
 
+## Vector index: opening a database no longer parses every index page to rebuild the location map (#6722)
+
+Opening a database walked every page of every `LSM_VECTOR` index and rebuilt the in-memory location map, whether
+or not the session went on to search it. The cost is one page parse plus one map insert per indexed vector, so it
+scales with the corpus rather than with what the session does, it is paid again on every open, and the page cache
+cannot amortise it because the work is parsing and map population rather than I/O: measured at ~1.4 s for 10M
+vectors, of which a cold open recovered only 8% - against 53% for documents and 89% for a graph with a Graph
+Analytical View, both of which open in single-digit milliseconds at the same size. A process that opened the
+database to write one document, or to read an unrelated type, paid all of it.
+
+The graph is deferred to the first search, two lines below the call site that did this - `ensureGraphAvailable()`
+loads or rebuilds it then, not at open. The location map now gets the same treatment: `loadVectorsAfterSchemaLoad()`
+records that there are pages worth reading and returns, and the first caller that actually needs a location pays
+for materialising them, once, under a dedicated lock. So a session that never touches a vector index never parses
+one, and open becomes constant-time in the number of indexed vectors.
+
+The deferral is structural rather than a convention: the field is reachable only through an accessor that
+materialises first, and the few sites that read it raw - the loader filling it, `drop()`, and the close-time
+predicates - say why. Two of those predicates matter to behaviour and are worth naming. `flush()` decides whether
+to build a graph on close, and asking that question used to mean reading the location count; it now answers from
+page counters instead, so the close of a session that never touched the index does not hand back at close exactly
+the cost the deferral removes from open, while the branches past that decision - which are choosing between
+building the whole graph now and recording a deferral - do materialise, because both need the real live count and
+because getting it wrong would send a large index down the synchronous-build arm that #6067 exists to avoid. And
+vector ids are handed out from a sequence the load derives from the pages, so every allocation now goes through a
+gate that materialises first: without it, an insert into a reopened index would restart the sequence at 0 and
+silently supersede the entries it collided with.
+
+The PQ codebooks, loaded at the same call site, are deferred with the locations - the ordinals a PQ search resolves
+are the ones the location map hands out, so the two are only ever useful as a pair. `findNeighborsFromVectorApproximate()`
+checks PQ availability before it reaches the lazy-load, so it now asks through `isPQSearchAvailable()`, which
+materialises, rather than reading the two fields directly and silently downgrading the first approximate search
+after a reopen to an exact one.
+
+Nothing about what any caller observes changes - only when the work happens. `countEntries()`, `getStats()`, a
+search, an insert and a replicated page update all still see the whole corpus; the first of them to ask is the one
+that pays.
+
 ## Postgres wire: `SHORT`/`BYTE`, `DATE`, `DATETIME` and `DECIMAL` no longer change type depending on whether a row was sampled (#6447)
 
 A RowDescription column is typed either from a sample value, when the result set has a row, or from the declared
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -225,6 +225,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
   private final    ReentrantLock                 locationsLoadLock        = new ReentrantLock();
   // What loadVectorsAfterSchemaLoad() deferred: load the PQ codebooks together with the locations, or not at all.
   private volatile boolean                       deferredPQLoad;
+  // Set if a materialisation ever ran while its own thread held the index-wide write lock, which is the one thing
+  // the dedicated load lock exists to prevent (PR #6731 review). Only a test reads it.
+  private volatile boolean                       materializedUnderWriteLock;
   private final    AtomicInteger                 nextId;
   private final    AtomicReference<INDEX_STATUS> status;
   // Set once the ignored location-cache limit has been reported, so a rebuild does not repeat the warning.
@@ -1163,13 +1166,26 @@ public class LSMVectorIndex implements Index, IndexInternal {
     // The loader reaches the location index through the field, not through vectorIndex(), so this cannot normally
     // recurse. Kept anyway because a future call site inside the load would otherwise deadlock on a non-reentrant
     // lock or, worse, load twice: a re-entrant caller is inside the population and must see the instance as it is.
-    if (locationsLoadLock.isHeldByCurrentThread())
+    // It says so at WARNING rather than returning quietly: handing back a half-filled map is the right thing to do
+    // for a caller that is inside the fill, but it is a silent under-materialisation for anyone else, and the
+    // comment above will not catch it if the invariant is ever broken (PR #6731 review).
+    if (locationsLoadLock.isHeldByCurrentThread()) {
+      LogManager.instance().log(this, Level.WARNING,
+          "Re-entrant materialisation of the location index of %s: a call site inside the load asked for the map "
+              + "it is filling, so it sees only what has been read so far", indexName);
       return;
+    }
 
     locationsLoadLock.lock();
     try {
       if (locationsMaterialized)
         return; // another thread materialised it while this one waited
+
+      // The whole point of the dedicated lock is that this scan does NOT run inside the index-wide exclusive
+      // section, so record whether it did and let a test assert on it (PR #6731 review) - the invariant lives in
+      // the call sites, which is exactly the kind of thing that regresses silently when one of them moves.
+      if (lock.isWriteLockedByCurrentThread())
+        materializedUnderWriteLock = true;
 
       try {
         LogManager.instance()
@@ -5791,8 +5807,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
    * @return true if PQ data is loaded and ready for zero-disk-I/O search
    */
   public boolean isPQSearchAvailable() {
-    if (!locationsMaterialized)
-      vectorIndex(); // the PQ codebooks are loaded with the locations (issue #6722)
+    vectorIndex(); // the PQ codebooks are loaded with the locations (issue #6722)
     return pqVectors != null && productQuantization != null;
   }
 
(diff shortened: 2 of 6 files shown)


lvca:
Thanks - the write-path finding is correct and now fixed in d7940d79e.

**The main point.** `put()`, `putBatch()` and `remove()` all took `lock.writeLock()` and reached the locations from inside it, so the first write against a large freshly reopened index ran the whole O(index size) parse while holding the exclusive lock. As you say, that is worse than what it replaced: at open the index was not yet published for concurrent use, so the cost fell on one thread; inside the write lock it falls on every thread contending for the index. All three now call `vectorIndex()` before acquiring the lock, mirroring the search paths. Once materialised the flag never goes back, so the `allocateVectorId()`/`vectorIndex()` calls that remain inside the exclusive section are plain volatile reads.

**On the missing concurrency test** - agreed, and the fix is testable without a race. The property is not "a concurrent search was not observed to stall", it is "the write lock was not held while the scan ran", and the index can record that directly. `materialisationNeverRunsUnderTheIndexWriteLock` asserts it for the insert, delete and search paths; reverting any one of the three hoists turns it red, which I checked. A timing-based version of the same test would be a race with a scan that takes microseconds at test scale, and the repo bars wall-clock assertions for exactly that reason.

**Re-entrancy guard** - fair. It now logs at WARNING before returning. Unreachable today, but if it ever fires it is a silent under-materialisation for any caller not inside the fill, and the comment would not catch that.

**Redundant flag check** in `isPQSearchAvailable()`/`getPQVectorCount()` - removed; `vectorIndex()` makes the same check itself.

**Release note** - I have left it as is. The write-path caveat was a property of the intermediate state, not of what ships: after this commit no path materialises under the index lock, which is what the note already describes.

Vector package after the change: 424 tests, 1 failure - `Issue6657CloseTimeRebuildPendingStatTest`, which fails identically at the merge base on `main`.

lvca:
@claude review again

PR #6738, opened 2026-08-25
fix(#6569): decouple lint from test lanes, add a ci-status aggregator

### LINKED ISSUE #6569: A failed 'setup' job makes every test lane report 'skipping', so a PR looks untested rather than broken
### Summary

When the `setup` job in `mvn-test.yml` fails, every downstream test lane (`unit-tests`, `slow-unit-tests`, `integration-tests`, `ha-integration-tests`, `opencypher-tck-tests`, the e2e lanes, `build-and-package`, `coverage-report` — 15 in total) reports **`skipping`**, not `failure`.

The result is a PR whose check list shows no test failures, because no tests ran.

### How this surfaced

While working #6379 I opened PR #6564 and its first CI run showed `setup fail` with 15 lanes `skipping`. The cause was unrelated to the PR: `integration/src/test/java/com/arcadedb/integration/importer/JsonLImporterIT.java` had landed on `main` in c816a66c7 (#6468) **without a trailing newline**, which fails the `end-of-file-fixer` pre-commit hook, which fails `setup`.

That one missing byte was silently skipping the entire test suite for *every* PR opened against `main` for as long as it was there. It was only noticed because a red `setup` was investigated rather than assumed to be flaky; the check list itself gave no indication that nothing had been tested. (The newline is fixed as of ab0574f20, so the immediate blockage is gone — this issue is about the failure mode, not that byte.)

### Why it matters

- **The signal is inverted.** "Skipped" reads as benign. A reviewer scanning the checks sees no failures and reasonably concludes the tests are fine.
- **It is repo-wide and silent.** Any commit to `main` that trips a lint hook disables testing for every open PR at once, with no notification.
- **Merge protection may not catch it.** Depending on which contexts are required, skipped lanes can satisfy branch protection.

### Suggested

Some combination of:

- make the aggregate `Summary` / merge-queue gate treat "test lanes skipped because `setup` failed" as a failure rather than a pass;
- have `setup` fail with a message that names the consequence ("test lanes will not run"), so the check list explains itself;
- run the lint hooks as their own job independent of the one that gates the test lanes, so a formatting nit cannot take the test suite offline;
- optionally, run `pre-commit` in a mode that reports diffs without failing the gating job, since the failure is auto-fixable and the fix is mechanical.

### Related

Surfaced during #6379 / PR #6564. See also #6568 (a lane that times out with no failing assertion — the same "CI is red but nothing says why" shape from the other direction).

lvca:
## Summary

Fixes #6569.

`setup` (repo/workflow lint + pre-commit hygiene checks) gated `build-and-package`, and every one of the fifteen downstream test lanes hangs off `build-and-package`. So a lint-only failure - concretely, a missing trailing newline that trips the `end-of-file-fixer` pre-commit hook (c816a66c7, #6468) - silently took the entire test suite offline for every open PR. GitHub reports the affected lanes as `skipping`, which reads as benign to a reviewer scanning the check list, and can satisfy branch protection without a single test having run.

The issue's own "suggested" list offers several partial options. This applies the strongest combination rather than picking one:

- **Decouple lint from the build/test gate.** `setup` is renamed to `lint` and dropped from `build-and-package`'s `needs`. A lint failure (formatting, workflow hygiene, pre-commit) now shows up as its own red check and runs in parallel with the build, instead of silently suppressing every real test.
- **Make "something upstream prevented tests from running" an explicit, named failure.** A new `ci-status` job waits on every other job in the workflow and fails with the offending job names whenever any of them is not `success` - so a `skipped` lane (for any reason, not just a lint failure) can never again pass as "nothing to worry about". This is the one check to mark as required in branch protection going forward.
- **Prevent the aggregator itself from silently rotting.** `check-ci-status-completeness.py` (new) fails the `lint` job if `ci-status`'s `needs` ever stops covering every job in the workflow - so adding a new lane and forgetting to wire it into the safety net is caught immediately instead of becoming the next #6569, for that one lane.

I deliberately did not adopt the issue's fourth suggestion (running pre-commit non-blocking / diff-only): softening enforcement isn't needed once lint no longer gates the tests, and it would weaken hygiene enforcement for no remaining benefit.

## Changes

- `.github/workflows/mvn-test.yml`: `setup` → `lint`, `build-and-package` no longer `needs` it, new `ci-status` aggregator job at the end.
- `.github/scripts/check-ci-status-completeness.py` (new): enforces that a `ci-status`-style aggregator job's `needs` transitively covers every job in its workflow.
- `.github/scripts/tests/test-ci-scripts.sh`: regression fixtures for the new script (rejects an incomplete aggregator, accepts a complete one, accepts transitive coverage, ignores workflows with no aggregator).

## Test plan

- [x] `check-workflow-artifact-deps.py`, `check-workflow-needs-outputs.py`, `check-test-reporter-guards.py`, and the new `check-ci-status-completeness.py` all pass against the modified `mvn-test.yml` and against every other workflow in the repo (no false positives).
- [x] `test-ci-scripts.sh` (all 84 checks, including 6 new ones for the new script) passes locally.
- [x] Verified the modified/added files satisfy the pre-commit hooks that actually apply outside `bindings/python/` (trailing whitespace, end-of-file newline, YAML parseability) - i.e., this PR doesn't reproduce the byte that caused #6569 in the first place.
- [x] Simulated the `ci-status` job's `jq` filter against synthetic `needs` payloads (all-success → passes silently; a skipped job → fails with the job name and a step-summary explanation).
- [x] `lint` and `build-and-package` ran as separate, parallel checks on this PR's own CI run, confirming the decoupling works end to end (not just in the meta-checks).

## Manual follow-up needed (not in this diff)

`setup` is renamed to `lint`, and the required check to protect going forward is the new `CI Status` (`ci-status`) job. If branch protection on `main` currently names `setup` (or any individual downstream lane) as a required status check, whoever has admin access should repoint it at `CI Status` after this merges - otherwise a rule pointing at the old name can never be satisfied again, or removing it without replacing it re-opens exactly the gap this PR closes. Called out here per code review since it's invisible in the diff and easy to forget.

## Summary by CodeRabbit

- **New Features**
  - Added automated validation to ensure workflow status checks cover all required jobs.
  - Added an always-running CI status report for failed or skipped required checks.
  - Published the latest Docker image for downstream end-to-end testing.

- **Bug Fixes**
  - Improved workflow execution so packaging and test jobs are not unnecessarily blocked by preparatory checks.

- **Tests**
  - Added coverage for incomplete, complete, transitive, and invalid workflow configurations.

### DIFF
--- a/.github/scripts/check-ci-status-completeness.py
+++ b/.github/scripts/check-ci-status-completeness.py
@@ -72,6 +72,15 @@ def needs_closure(jobs, name, seen=None):
 
 
 def is_aggregator(job_id):
+    """A job id counts as an aggregator by naming convention alone, not by inspecting what it does.
+
+    That is deliberate - there is nothing else in a job's YAML that reliably marks "this is the
+    required check for the workflow" - but it does mean any job whose id happens to end in
+    `-status` (e.g. a hypothetical `slack-notify-status` that only posts a message) is treated as
+    one and held to the same "needs everything" bar. Low risk while this repository is the only
+    caller and controls its own job ids; if that ever becomes a real collision, tighten this to an
+    explicit allowlist rather than a suffix match.
+    """
     return job_id == "ci-status" or job_id.endswith("-status")
 
 
--- a/.github/scripts/check-ci-status-completeness.py
+++ b/.github/scripts/check-ci-status-completeness.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+#
+# Fails when a workflow's status-aggregator job does not transitively depend on every other job in
+# the file.
+#
+# Context (issue #6569): when `setup` failed in mvn-test.yml, GitHub reported the fifteen jobs
+# behind it as `skipping` rather than `failure`. A `skipped` check reads as "nothing to worry about"
+# to a reviewer scanning the check list, and branch protection can be satisfied by a check that
+# never ran at all - a single missing trailing newline silently took the whole test suite offline
+# for every open PR until someone happened to investigate a red `setup` instead of assuming flake.
+#
+# The fix in mvn-test.yml is a final `ci-status` job with `if: always()` that reads
+# `needs.*.result` and fails the run whenever a job it depends on is anything but `success`,
+# turning "some required job never even ran" into one explicit, named failure instead of a
+# checklist row that looks the same as "passed".
+#
+# That only holds while the aggregator's `needs` actually reaches every job whose failure or skip
+# should be caught. It is very easy to add a new lane below `build-and-package` and forget to wire
+# it into the aggregator too - the omission is invisible until the exact incident this exists to
+# prevent happens again, for that one lane. This script makes the invariant enforced instead of
+# remembered:
+#
+#     a job named 'ci-status', or whose id ends in '-status', must transitively `need` every
+#     other job defined in the same workflow file.
+#
+# A workflow with no such job is not checked - not every workflow needs an aggregator, and this
+# only activates once one opts in by name.
+#
+# Usage:
+#   check-ci-status-completeness.py                 # check .github/workflows
+#   check-ci-status-completeness.py FILE|DIR ...    # check the given workflows
+#
+# @author Luca Garulli (l.garulli@arcadedata.com)
+
+import sys
+from pathlib import Path
+
+try:
+    import yaml
+except ImportError:  # pragma: no cover - the workflow installs it explicitly
+    sys.exit("check-ci-status-completeness: PyYAML is required (pip install pyyaml)")
+
+
+def workflow_files(argv):
+    targets = [Path(a) for a in argv] or [Path(".github/workflows")]
+    files = []
+    for target in targets:
+        if target.is_dir():
+            files.extend(sorted(p for p in target.iterdir() if p.suffix in (".yml", ".yaml")))
+        else:
+            files.append(target)
+    return files
+
+
+def needs_of(job):
+    if not isinstance(job, dict):
+        return []
+    needs = job.get("needs")
+    if isinstance(needs, str):
+        return [needs]
+    return [n for n in needs if isinstance(n, str)] if isinstance(needs, list) else []
+
+
+def needs_closure(jobs, name, seen=None):
+    """Every job `name` transitively depends on, direct or indirect."""
+    seen = set() if seen is None else seen
+    for parent in needs_of(jobs.get(name) or {}):
+        if parent not in seen:
+            seen.add(parent)
+            needs_closure(jobs, parent, seen)
+    return seen
+
+
+def is_aggregator(job_id):
+    return job_id == "ci-status" or job_id.endswith("-status")
+
+
+def check(path):
+    """Violations in one workflow file, as printable strings."""
+    try:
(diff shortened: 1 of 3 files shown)


lvca:
CI note: `unit-tests` and `integration-tests` are red on this run, but not because of this diff (which touches only `.github/workflows/mvn-test.yml`, `.github/scripts/check-ci-status-completeness.py`, and `.github/scripts/tests/test-ci-scripts.sh` - no Java code).

- `unit-tests`: `com.arcadedb.graph.olap.GraphAnalyticalViewTest.issue6641_corruptedPersistedCsrInvalidatesDeferredRestoreWithoutBlockingQuery` fails on a `SchemaException: Type with name 'AlsoNonExistent' was not found` from an async deferred-restore path. This is the same GraphAnalyticalView deferred-restore timing family already seen failing on unmodified `main` (sibling methods `issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore` / `issue6641_interveningCommitInvalidatesDeferredRestoreWithoutBlockingQuery`, both reproduced on main commit `4b107cf4c`). `main`'s own latest run (`87695796d`) is independently red too, on a different member of the same flaky family (`CypherGAVBoundTargetCardinalityTest`).
- `integration-tests`: `com.arcadedb.postgres.Issue6458PortalSuspensionIT` fails the same 2 tests (`bindFallbackOnAnUnknownSourceStatementReRunsInsteadOfServingTheStaleExecutedPortal`, `bindFallbackOnAnUnknownSourceStatementReRunsADeferredCatalogQueryTooNotJustPlainSql`) already reproduced on unmodified `main` in PR #6734's investigation - nothing here touches Postgres wire-protocol code either.

Both are pre-existing, chronic flakes on `main`, confirmed by checking `main`'s own last several CI runs (`conclusion: failure` on each of the last 5). Not something this PR introduces or needs to fix.

Worth noting as the flip side of what this PR is about: `main`'s own trunk CI has been red for a while on unrelated flakes, which is exactly the kind of state the new `ci-status` check makes visible instead of letting a `skipped` lane hide it - it just happens that on *this* PR the red is a real (if pre-existing) test failure rather than a skip.

lvca:
Update after the two follow-up commits (env-safe NEEDS_JSON, extra test fixtures, CI-cost comment): same 4 lanes red again (\`unit-tests\`, \`slow-unit-tests\`, \`integration-tests\`, \`ha-integration-tests\`), same pre-existing-on-\`main\` pattern - confirmed per-lane again, including two more instances of these chronic flakes (\`Issue6655StaleGraphPrefixReuseTest\` on \`slow-unit-tests\`, \`Issue5569SlotMergeDeleteRaftIT\` on \`ha-integration-tests\`), both also currently red on unmodified \`main\`. \`ci-status\` correctly failed and named all four, which is itself a live end-to-end confirmation that the new aggregator works as designed.

Three rounds of automated review (CodeRabbit + Codacy: clean; `claude-review` x3) have converged with no blocking findings - the one remaining nitpick (`is_aggregator()`'s explicit `"ci-status"` branch being redundant with its `-status` suffix check) is explicitly called out by the reviewer as taste/non-blocking, and I'm keeping it as-is for readability of the primary case.

Issue #6399, opened 2026-08-18
Three follow-ups from #6351: a file length that is not a whole number of pages goes unnoticed, applyChanges still asks the file whether a page is new, and a precision test's retry loop is false comfort


lvca:
Three findings from the work on #6351 (PR #6355, merged as `68cf79803`). Filed together because they were found
together, but they are independent and in descending order of how actionable they are. None is a regression from
#6355.

Items 1 and 2 are the same shape as #6351 itself, seen at two other levels: a decision taken from the physical file
when the fact being asked about is what is committed.

---

## 1. Nothing checks that a component file's length is a whole number of pages

`PaginatedComponent`'s constructor derives the page count by dividing (`PaginatedComponent.java:127-132`):

```java
final long fileSize = file.getSize();
if (fileSize == 0)
  // NEW FILE, CREATE HEADER PAGE
  pageCount.set(0);
else
  pageCount.set((int) (fileSize / pageSize));
```

That division **floors**, and nothing anywhere asks whether there was a remainder. A file killed part way through
writing a page keeps a partial tail that no component counts, no check reports, and nothing repairs. It is simply
rounded away at every open, forever.

#6355 fixed this for the dictionary and only for the dictionary: `createHeaderPageIfMissing()` now decides from the
committed count rather than from `file.getSize() > 0`, so a dictionary file holding less than one whole page gets
its header page written over the stub instead of being left as one. That was a side effect of moving the guard, not
a general fix. Every other component still floors and moves on.

**This is the natural third sibling of two guards that already live in that same constructor, ten lines above:**

- the file-id agreement check (#6283, `:92`), and
- the page-size agreement check (#6314, `:104`).

Both exist for the same reason: a component that disagrees with its file about how to address it reads real bytes
at the wrong place, and the argument for failing loudly right there, before a single page is addressed, applies
unchanged to a file whose length says it holds a page that is not all there.

**What I have not established** is what the partial tail costs on a component other than the dictionary. For a
bucket or an index the next `addPage` writes at the floored page number, so the remainder is overwritten rather
than read, and it may cost nothing at all. That is worth determining before choosing between the two obvious
responses, and the choice matters:

- **a tripwire** (`throw` like its two siblings) refuses to open a database that a previous build opened happily,
  which is a real compatibility question for anyone whose file has been carrying a stub for a while;
- **a report** (log at open, or a `CHECK DATABASE` finding) says the same thing without that risk, and `CHECK
[...]

PR #6740, opened 2026-08-26
fix(#6399): torn page tail warning, replica-apply page-count lag, flaky precision test

lvca:
## Summary

Three independent, small follow-ups spotted during #6351/PR #6355, filed together as #6399.

**1. A component file whose length is not a whole number of pages goes unnoticed.** `PaginatedComponent`'s constructor derived its page count as `fileSize / pageSize`, which floors silently - a file killed part way through writing a page kept an uncounted, unrepaired partial tail forever. Unlike the file-id (#6283) and page-size (#6314) tripwires right above it in the same constructor, this is fixed as a **WARNING logged at open**, not a thrown exception: those two guard against programming errors that cannot exist in a file a correct build wrote, while a torn tail is exactly what an ordinary power cut produces, and refusing to open would turn a database every previous build opened happily into one this build won't. It also costs nothing beyond disk space for a bucket or index (unlike the Dictionary, which #6355 already fixed separately): the next page appended to the file lands at `pageCount * pageSize` and overwrites the stray bytes wholesale (verified in the added test, which forces real page growth after the torn-tail reopen and confirms the file length is a whole number of pages again).

Also removed the redundant `if (fileSize == 0)` special case noted in the issue - `0 / pageSize` is already `0`.

**2. `TransactionContext.commitFromReplica` decided "is this page new" from the file's physical size**, the same stale signal `getPage()` had at #5976 (fixed there to ask the schema-registered component instead, since its count is bumped synchronously at commit rather than lagging the async flush thread). Mirrored that fix here for consistency.

I could not construct a test that demonstrates the consequence the issue describes (a rollback skipping the dictionary repair): `TransactionContext.commitFromReplica` has no caller anywhere in the current codebase - the live replication/recovery path is `TransactionManager.applyChangesInternal`, which does not share this defect - and a raw page write never desyncs `Dictionary`'s cached in-RAM name map by itself; only `reload()`/`updateName()` do, and both already classify correctly via the component. So the fix is applied defensively (cheap, correct, consistent with the established pattern) without an accompanying regression test, since none I could write would exercise a real failure.

**3. `SQLMethodPrecisionTest.instant()` failed through its own retry loop.** The loop built a **fresh `NanoClock` per sample**, which resets `NanoClock`'s elapsed-nanos baseline (`System.nanoTime()`-based) on every call - so instead of ten independent nanosecond-resolution samples, all ten retries just re-read whatever (coarser) resolution the underlying `Clock.systemUTC()` happened to have, occasionally landing on ten identical whole-second values in a row. Fixed by reusing **one** `NanoClock` across the retries, so real elapsed nanoseconds actually accumulate between calls. Verified flake-free over 15 consecutive local runs.

## Test plan
- New regression test: `Issue6399PartialTailPageTest` (item 1) - reopens a bucket file truncated to a torn tail, asserts a WARNING is logged, the database opens and is fully usable, and a later real page allocation overwrites the torn bytes.
- `engine` module: ran the full `com.arcadedb.engine.**` and `com.arcadedb.database.**` packages (excluding benchmark/slow/vector lanes) - 1131 tests, 0 failures.
- `SQLMethodPrecisionTest` run 15 times consecutively with no failures.
- Full reactor compile of `engine`, `server`, `ha-raft` - no downstream breakage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery and replication handling when pages are written asynchronously.
  * Reopening files with incomplete trailing page data now logs a warning without failing.
  * Existing data remains available, and partial trailing data is safely overwritten during subsequent page allocation.
  * Improved nanosecond precision validation for instant values.

* **Tests**
  * Added coverage for recovery from truncated bucket files and partial page tails.


lvca:
Addressed in 41ae22329:

- **TransactionContext.commitFromReplica isNew classification**: valid catch. Fixed by snapshotting each file's component page count once (on first encounter in the loop) instead of re-querying `component.getTotalPages()` on every page, which was picking up this same loop's own `newPageCounters` bumps from earlier iterations and could misclassify a lower-numbered new page as 'modified' when a higher-numbered new page of the same file was processed first. I did not add a dedicated regression test for this: as with the original fix in this PR, `commitFromReplica` has no live caller anywhere in the codebase today, and I could not find an observable difference this specific misclassification produces given how the surrounding commit/rollback machinery currently works (traced `checkPageVersion`/`updatePageVersion` - both ignore their own `isNew` parameter for actual logic - and the page content itself is unaffected since `PageManager.loadPage`'s physical-file check is independent of this flag). Happy to add one if a concrete failure mode is identified.
- **Logger level restore in the test**: fixed, `detach()` now restores the pre-`attach()` level instead of leaving it pinned at `ALL`.

lvca:
Second review pass (183a660de):

- **countType() ground-truth concern**: traced it - a freshly reopened database's `LocalBucket.cachedRecordCount` always starts at -1, so the first `countType()` call after reopen does force a real bucket scan here, it wasn't stale in practice. Still hardened it per the suggestion: the test now also captures the record's RID before reopening and verifies it directly via `lookupByRID()` after, which is unambiguous ground truth independent of that caching detail.
- **Detach-on-open-failure**: fixed, `open()` now runs inside the `try` with cleanup in the `finally`, same shape as suggested.

lvca:
Third review pass (dc0db6791):

- **Boxed HashMap nit**: fixed, `componentPageCountAtStart` now uses `IntIntHashMap` (zero-boxing), consistent with `modifiedFiles`/`IntHashSet` a few lines above in the same method.

Remaining notes from the reviews, left as-is:
- *Re-logging the torn-tail WARNING on every restart until the page is overwritten*: intended - it's a stateless informational log, not a stored once-only flag, matching the existing 'report, don't repair-on-open' design for this fix. Adding suppression state would be complexity for a cosmetic log-noise concern on a fix path that's already the exception rather than the rule.
- *`getFileByIdIfExists` called every iteration / `CapturingHandler`'s dead fallback branch*: both flagged explicitly as harmless/non-blocking by the reviewer, on a code path with no current caller (item 2) or in test-only code that mirrors an existing, already-merged pattern elsewhere in the suite. Not changing either.

lvca:
Latest review pass (dc0db6791): no blocking issues found. The one remaining note (caching the resolved `PaginatedComponent` reference per file, not just its page count, to save a redundant `getFileByIdIfExists` lookup per WAL page) is flagged explicitly as optional/non-blocking on a code path with no current caller - not applying it, since it would add state duplication for a saving that only matters if/when this method is ever wired back up. Waiting on the full CI test matrix (unit/integration/ha-integration/e2e) to finish.

lvca:
CI note: `unit-tests` failed on `GraphAnalyticalViewTest.issue6641_findProviderDoesNotTriggerUnrelatedViewsDeferredRestore`, which is unrelated to this PR (a materialized/analytical-view deferred-restore dispatch test - this PR only touches `PaginatedComponent`, `TransactionContext`, and two test files, none in that area). Reproduced 5/5 clean locally on this branch, consistent with a timing-sensitive pre-existing flake rather than a regression. Will re-run the job once the rest of the workflow run finishes (can't rerun individual failed jobs while the run is still in progress).

lvca:
CI note: `integration-tests` also failed, on `com.arcadedb.postgres.Issue6458PortalSuspensionIT` (Postgres wire protocol portal-suspension/statement-cache behavior) - again unrelated to this PR's diff, which only touches engine-module files (`PaginatedComponent`, `TransactionContext`, two test files). Both this and the earlier `unit-tests` failure are in modules/subsystems this PR does not touch, consistent with unrelated pre-existing flakes rather than anything this change introduced. Will rerun both failed jobs once the rest of the run (`slow-unit-tests`, `ha-integration-tests` still pending) finishes.

lvca:
CI update: the run finished with 3 failures, all confirmed pre-existing flakes unrelated to this PR's diff:
- `unit-tests`: `GraphAnalyticalViewTest` (materialized/analytical-view deferred-restore dispatch) - reproduced passing 5/5 locally on this branch.
- `integration-tests`: `Issue6458PortalSuspensionIT` (Postgres wire protocol portal suspension) - unrelated module.
- `slow-unit-tests`: `Issue6655StaleGraphPrefixReuseTest` - an `awaitility` timeout waiting for an async LSM vector-index rebuild, a known flaky-under-load pattern in `com.arcadedb.index.vector` (unrelated module; this is why that area normally gets its own `vector-unit-tests` lane, per this repo's own CI notes on rebuild-semaphore convoying).

`ha-integration-tests` and `vector-unit-tests` passed cleanly. `codecov/patch`/`codecov/project` also failed - will check those once the reruns settle, but per this repo's convention those aren't required/blocking checks for a small diff.

Rerun in progress for the 3 failed jobs.

lvca:
CI investigation, summarizing findings across both runs:

**`unit-tests` failed on a DIFFERENT test each time** - `GraphAnalyticalViewTest` on the first run (reproduced passing 5/5 locally on this branch), `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` on the rerun. Neither is anywhere near this PR's diff (openCypher GAV/view-restore machinery vs. this PR's `PaginatedComponent`/`TransactionContext` engine changes). A real regression from this PR would fail the *same* way each time; a shifting failure across unrelated subsystems is the signature of general suite flakiness.

**`integration-tests` failed the SAME way both times** on `com.arcadedb.postgres.Issue6458PortalSuspensionIT` (both its sub-tests). I checked whether `main` itself is currently healthy on this test and it is not: the most recent completed `main`-branch CI run (https://github.com/ArcadeData/arcadedb/actions/runs/32907761996) fails on the **exact same test, the exact same way**, plus an unrelated Bolt routing-table failure. `main` is currently red on `unit-tests`/`integration-tests`/`slow-unit-tests`/`ha-integration-tests` independent of this PR entirely.

**`slow-unit-tests`** failed on `Issue6655StaleGraphPrefixReuseTest`, an `awaitility` timeout on an async LSM vector-index rebuild - a known flaky-under-load pattern in `com.arcadedb.index.vector` unrelated to this diff.

**Conclusion**: none of these failures are caused by this PR. Re-running further is unlikely to help while `main` itself is in this state, so I'm backing off the tight rerun loop and will check back periodically instead. This PR's own code review is fully resolved (three rounds, no blocking issues) and the actual changed code has 1131 local engine/database tests passing plus a clean full reactor compile of engine/server/ha-raft. Ready to merge on its own merits; the remaining red checks reflect main's current CI health, not this change.

lvca:
Final confirmation: checked `main`'s own `unit-tests` job on that same recent run (https://github.com/ArcadeData/arcadedb/actions/runs/32907761996) - it fails on `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair`, the exact same test that showed up on this PR's rerun. So both of this PR's `unit-tests` failures (across two attempts) and its `integration-tests` failure all match tests that are independently, currently failing on `main` itself. This is conclusively main-branch CI health, not this PR. Backing off further reruns until main stabilizes; will keep checking periodically.

lvca:
Status update: `main` has been red for ~10 hours across many consecutive runs on `unit-tests`/`integration-tests`/`slow-unit-tests`/(intermittently)`ha-integration-tests` - clearly a sustained, pre-existing CI health issue unrelated to this PR, not a one-off flake. Good sign: `main`'s latest run has `integration-tests` passing again (the `Issue6458PortalSuspensionIT` failure cleared), though `unit-tests`/`slow-unit-tests`/`ha-integration-tests` are still red there. Triggered a third rerun of this PR's failed jobs to see if it clears here too.

PR #6742, opened 2026-08-26
fix(#5671): UnionStatement clause accessors and positional variable-kind checks

### LINKED ISSUE #5671: [Cypher] UnionStatement answers clause accessors with its first branch only, and variable kinds are an end-state map
Two things found while closing #5626 (PR #5642) that are not covered by #5656. They are unrelated to each other; filed together because both came out of the same work and both are about the variable-kind / clause-shape layer rather than about subquery execution.

---

## 1. `UnionStatement` silently answers ten clause accessors with its *first branch*

`UnionStatement` implements `CypherStatement`, and ten of the accessors it inherits delegate to `queries.getFirst()`:

```java
// UnionStatement.java:107-195
public List<MatchClause>  getMatchClauses()   { return queries.getFirst().getMatchClauses(); }
public WhereClause        getWhereClause()    { return queries.getFirst().getWhereClause(); }
public ReturnClause       getReturnClause()   { return queries.getFirst().getReturnClause(); }
public CreateClause       getCreateClause()   { return queries.getFirst().getCreateClause(); }
public SetClause          getSetClause()      { return queries.getFirst().getSetClause(); }
public DeleteClause       getDeleteClause()   { return queries.getFirst().getDeleteClause(); }
public MergeClause        getMergeClause()    { return queries.getFirst().getMergeClause(); }
public List<UnwindClause> getUnwindClauses()  { return queries.getFirst().getUnwindClauses(); }
public List<WithClause>   getWithClauses()    { return queries.getFirst().getWithClauses(); }
public List<ClauseEntry>  getClausesInOrder() { return queries.getFirst().getClausesInOrder(); }
```

So any code holding a `CypherStatement` and asking what it matches, creates, deletes or binds gets **branch 1's answer, with nothing to indicate that branches 2..n were dropped**. `MATCH (a:P) RETURN a.x AS c UNION MATCH q = (m:P)-->() RETURN q.name AS c` reports one `MatchClause` and no path variable.

Four of those accessors alone have ~150 call sites under `com.arcadedb.query.opencypher`:

| accessor | call sites (main) |
| --- | --- |
| `getReturnClause()` | 67 |
| `getMatchClauses()` | 41 |
| `getClausesInOrder()` | 29 |
| `getWithClauses()` | 12 |

Exactly **five** files guard with `instanceof UnionStatement`: `CypherExecutionPlanner`, `CypherExpressionWalker`, `CypherSemanticValidator`, `CypherExecutionPlan`, `SubqueryStep`. It works today because those are the entry points, and they route a union to branch-aware handling before anything downstream sees it.

**This is a latent-defect generator, not a latent theory.** It caught #5626 in review. The shared scope builder added there had to grow an explicit guard, because without it a `CALL { ... UNION ... }` body would have had branch 1's clauses walked *as if they were the union's*:

```java
// CypherSemanticValidator.nestedVarTypes, as merged
// (UnionStatement.getClausesInOrder() answers with its FIRST branch's clauses, which is why this returns before
// delegating rather than letting the shared build walk them as if they were the union's.)
if (nested instanceof UnionStatement)
  return new HashMap<>(outer);
```

Nothing would have failed if that guard had been forgotten. The kinds declared by branches 2..n would simply not exist, and the checks that read them would quietly not fire - which is the same failure mode #5626 was filed about.

**Severity, stated fairly.** The *return-column* accessors are partly protected: `validateUnion` enforces `DifferentColumnsInUnion`, so every branch must project the same column names, and branch 1's names are the union's names. The **clause** accessors have no such constraint - branch 2 may have entirely different `MATCH` / `CREATE` / `WITH` clauses - so `getMatchClauses`, `getClausesInOrder`, `getWithClauses`, `getCreateClause`, `getMergeClause`, `getDeleteClause`, `getSetClause`, `getUnwindClauses` and `getWhereClause` are the exposed ones. I did not find a reachable user-visible wrong answer on current main; every path I traced is guarded. The report is that the next one will not be, and will not announce itself.

**Suggested fix.** Not a comment, and not an audit of the 150 call sites - both decay. Make the type stop answering a question it cannot answer:

- Segregate the interface, so `UnionStatement` is not a thing you can ask for "the" match clauses at all, and a caller that needs them must reach `getQueries()` and decide what a union means for its purpose; or
- have the ten throw `UnsupportedOperationException`, which turns every unguarded path into a loud failure the first time a union reaches it instead of a silent half-answer.

The first is better if the call sites can absorb it. The second is mechanical and can land immediately. Either converts an invisible wrong answer into something a test can catch.

---

## 2. Variable kinds are an end-state map, applied regardless of where the expression sits

`validateVariableTypes` (statement) and `nestedVarTypes` (subquery body) share one construction since #5626, and it builds a single map per scope which is then checked against *every* expression in that scope, wherever it appears. So in

```cypher
CALL { MATCH (m:P) WHERE type(m) = 'KNOWS' WITH 1 AS m RETURN m }
```

the `type(m)` in the `MATCH` is checked against the post-`WITH` scope, where `m` is no longer a relationship, and the check does not fire.

This is deliberate and documented as such - it is the same approximation the top-level statement has always used, and answering a body *more* precisely than the query around it would have reintroduced exactly the clause-dependent asymmetry #5602 and #5626 exist to remove. It is recorded on the method:

> What comes back is the body's *end* state, one map applied to every expression in it wherever that expression sits [...] It errs one way only: a name a later `WITH` re-binds to something kindless loses its kind for the clauses before it too, so a check is missed, never invented.

**It cannot produce a false rejection**, and that holds structurally rather than by luck: a false rejection would need a name to be read where its kind is K and then rebound to `PATH` later in the same scope. Cypher does not allow a bound name to be rebound by a pattern (`Variable 'x' already declared`), and the one rebinding it does allow - a `WITH` projection - can drop a kind but never manufacture a path. So the end state is never more path-ish than any position preceding it.

Low priority for that reason. But it is the remaining honest gap in "a body is validated like the query around it", and if it is fixed it has to be fixed on **both** sides at once - making only the nested side positional would recreate the asymmetry.

**Suggested fix.** Give the walk a per-clause scope: `CypherExpressionWalker` already visits clauses in order, so a `Visitor` hook at each clause boundary would let the kind check advance its scope as it goes. Two things need care: a `MATCH`'s pattern must declare its variables *before* that clause's own expressions are checked (an inline property predicate references them), and a `WITH`'s projection expressions are evaluated in the *pre*-projection scope while its `WHERE` is evaluated in the post-projection one.

lvca:
## Summary

Fixes #5671 (both parts).

**Part 1 - `UnionStatement` answered clause accessors from branch 1 only.** Nine `CypherStatement` accessors (`getMatchClauses`, `getWhereClause`, `getCreateClause`, `getSetClause`, `getDeleteClause`, `getMergeClause`, `getUnwindClauses`, `getWithClauses`, `getClausesInOrder`) silently delegated to `queries.getFirst()`, with nothing to indicate branches 2..n were dropped. They now throw `UnsupportedOperationException` instead - a caller that needs branch-specific clauses must go through `getQueries()`. `getReturnClause()` is left delegating: `DifferentColumnsInUnion` already guarantees every branch shares the same return column names, so branch 1's answer is the union's answer.

Auditing every call site of the nine accessors (150+ across `com.arcadedb.query.opencypher`) found two genuinely unguarded ones that reached a real `UnionStatement` at runtime:
- `CypherVariableUsage.subqueryReferencesVariable` checked only branch 1 of a `CALL { ... }` body, so a relationship variable read only in branch 2 was reported as unreferenced - a silently dropped edge binding. Fixed to check every branch.
- `OpenCypherQueryEngine.analyze(...).getOperationTypes()` (the MCP permission-classification path) read `getSetClause()`/`getRemoveClauses()` directly and would have thrown for any UNION containing a write in any branch. Added `hasSet()`/`hasRemove()` predicates to `CypherStatement`, aggregated across branches on `UnionStatement` the same way `hasCreate()`/`hasMerge()`/`hasDelete()` already are (this also fixes a pre-existing miss: a `REMOVE` in a UNION branch was never classified as `UPDATE` at all).

**Part 2 - variable-kind checks used one end-state map for a whole statement/body.** `type(m)`/`labels(m)`/etc. argument-type checks were built from a single map computed once for the whole statement, so a later `WITH` that dropped or changed a name's kind retroactively erased a check against an earlier clause where the kind still applied:

```cypher
MATCH (m:P) WHERE type(m) = 'KNOWS' WITH 1 AS m RETURN m
```

`m` is a node from `MATCH`, and `type()` requires a relationship - this should always be rejected regardless of what a later clause does with the name `m`. It silently passed instead, because by the time the (single, whole-statement) kind map was built, `m` was a SCALAR (from `WITH 1 AS m`), not a NODE.

`CypherExpressionWalker` now advances the variable-kind scope clause by clause (`Visitor#forClauseEntry`) as it walks a statement or subquery body, so each clause's own expressions are checked against the scope as of that clause - matching real Cypher's `WITH` semantics (its projection items read the pre-projection scope; its `WHERE`/`ORDER BY`/`SKIP`/`LIMIT` read the post-projection one, confirmed against the Neo4j Cypher manual). This is a validation-strictness increase (rejects some previously-silently-accepted queries); the full OpenCypher suite was run to check for collateral rejections.

## Test plan
- [x] New regression tests: `Issue5671UnionStatementClauseAccessorsTest` (9 accessors throw, `hasSet`/`hasRemove` aggregate correctly, the two audited call-site fixes, and a functional divergent-UNION-inside-`CALL{}` execution test) and `Issue5671PositionalVariableKindsTest` (the exact `type(m)`/`WITH`/rebinding scenario, top-level and nested-body, plus sanity checks that unaffected shapes keep working)
- [x] Both new tests verified to fail without their respective fix (temporarily reverted each fix, confirmed red, restored)
- [x] Full `com.arcadedb.query.opencypher.**` suite: 9018 tests, 0 failures (1 unrelated pre-existing flake reproduced identically on unmodified `main`, confirmed passing in isolation)
- [x] `mcp` module (295 tests, exercises `getOperationTypes()` permission classification): all pass
- [x] `mvn -pl engine -am compile` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cypher `UNION` handling when branches contain different clauses or write operations.
  * Corrected variable scope and type validation across clauses, subqueries, `WITH`, and `FOREACH`.
  * Improved detection and classification of `SET` and `REMOVE` operations.
  * Ensured relationship-variable references are detected across all `UNION` branches.

* **Tests**
  * Added coverage for positional variable types, nested queries, divergent `UNION` branches, and branch-specific write operations.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/CypherStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/CypherStatement.java
@@ -93,6 +93,31 @@ public interface CypherStatement {
   }
 
   /**
+   * Returns true if this statement contains a non-empty SET clause.
+   * Defaults to false for non-query (control/admin/DDL/session) statements.
+   * <p>
+   * Kept as a predicate rather than requiring callers to inspect {@link #getSetClause()} directly, because that
+   * accessor has no single answer for a {@code UnionStatement} - unlike {@link #hasCreate()}, {@link #hasMerge()}
+   * and {@link #hasDelete()}, which already avoid the same trap (issue #5671).
+   *
+   * @return true if contains SET
+   */
+  default boolean hasSet() {
+    return false;
+  }
+
+  /**
+   * Returns true if this statement contains REMOVE operations.
+   * Defaults to false for non-query (control/admin/DDL/session) statements.
+   * See {@link #hasSet()} for why this is a predicate rather than {@link #getRemoveClauses()} emptiness.
+   *
+   * @return true if contains REMOVE
+   */
+  default boolean hasRemove() {
+    return false;
+  }
+
+  /**
    * Returns the ORDER BY clause if present.
    * Defaults to null for non-query (control/admin/DDL/session) statements.
    *
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/SimpleCypherStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/SimpleCypherStatement.java
@@ -416,6 +416,16 @@ public class SimpleCypherStatement implements CypherStatement {
   }
 
   @Override
+  public boolean hasSet() {
+    return setClause != null && !setClause.isEmpty();
+  }
+
+  @Override
+  public boolean hasRemove() {
+    return hasRemove;
+  }
+
+  @Override
   public OrderByClause getOrderByClause() {
     return orderByClause;
   }
@@ -473,10 +483,6 @@ public class SimpleCypherStatement implements CypherStatement {
   @Override
   public List<RemoveClause> getRemoveClauses() {
     return removeClauses;
-  }
-
-  public boolean hasRemove() {
-    return hasRemove;
   }
 
   @Override
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/UnionStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/UnionStatement.java
@@ -92,7 +92,6 @@ public class UnionStatement implements CypherStatement {
     return queries.get(0);
   }
 
-  // Delegate CypherStatement methods to the first query
   // UNION inherits read-only status from all subqueries
 
   @Override
@@ -103,16 +102,36 @@ public class UnionStatement implements CypherStatement {
     return true;
   }
 
+  /**
+   * A UNION has no {@code MATCH} of its own - each branch has its own, and they can differ (issue #5671):
+   * {@code MATCH (a:P) RETURN a.x AS c UNION MATCH q = (m:P)-->() RETURN q.name AS c} has two entirely different
+   * match shapes, and answering with branch 1's would silently drop branch 2's. Every caller that needs to know
+   * what a branch matches must go through {@link #getQueries()} and decide what a union means for its purpose -
+   * exactly what {@link com.arcadedb.query.opencypher.parser.CypherExpressionWalker} already does by walking each
+   * branch as a nested statement in its own right.
+   *
+   * @throws UnsupportedOperationException always
+   */
   @Override
   public List<MatchClause> getMatchClauses() {
-    return queries.getFirst().getMatchClauses();
+    throw new UnsupportedOperationException(
+        "UnionStatement has no single MATCH clause list: branches can differ. Use getQueries() and inspect each branch.");
   }
 
+  /**
+   * @throws UnsupportedOperationException always, for the reason documented on {@link #getMatchClauses()}
+   */
   @Override
   public WhereClause getWhereClause() {
-    return queries.getFirst().getWhereClause();
+    throw new UnsupportedOperationException(
+        "UnionStatement has no single WHERE clause: branches can differ. Use getQueries() and inspect each branch.");
   }
 
+  /**
+   * Unlike the other clause accessors, this one is safe to answer from the first branch: {@code validateUnion}
+   * enforces {@code DifferentColumnsInUnion}, so every branch is required to project the same return column
+   * names, which makes branch 1's column names the union's column names.
+   */
   @Override
   public ReturnClause getReturnClause() {
     return queries.getFirst().getReturnClause();
@@ -143,6 +162,22 @@ public class UnionStatement implements CypherStatement {
   }
 
   @Override
+  public boolean hasSet() {
+    for (final CypherStatement query : queries)
+      if (query.hasSet())
+        return true;
+    return false;
+  }
+
+  @Override
+  public boolean hasRemove() {
+    for (final CypherStatement query : queries)
+      if (query.hasRemove())
+        return true;
+    return false;
+  }
+
+  @Override
   public OrderByClause getOrderByClause() {
     // For UNION, ORDER BY applies to the final result
     // Currently we don't support ORDER BY after UNION in the grammar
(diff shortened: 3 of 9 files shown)


lvca:
Thanks for the thorough review - the part-2 correctness concern was real and I confirmed it empirically before fixing it.

## Part 2 correctness concern - confirmed and fixed

I reproduced this directly. Your own example didn't actually throw (`type()` only rejects `NODE`, not `SCALAR`, so the contaminated kind happened to land on a value the check doesn't reject), but the underlying mechanism you identified is real, and I found a variant that does false-positive-reject:

```cypher
MATCH ()-[r:KNOWS]->()
CALL {
  WITH r
  WHERE type(r) = 'KNOWS'
  WITH 1 AS x
  MATCH (r:P)
  RETURN r, x
}
RETURN r
```

`r` is genuinely a relationship where `type(r)` is checked; a later, disconnected reuse of the name in the same body (after an intervening `WITH` that drops `r` from scope) was reaching backwards and contaminating the earlier check. Root cause: seeding the incremental walk from the body's own precomputed end state relocated the bug from "one map for the whole scope" to "one map for the whole scope, at the nested-body granularity" - same disease, smaller patient.

Fixed in 58196e0: `FunctionArgumentChecks` now seeds a nested body from a copy of the outer scope exactly as it stands at the crossing point, not from the body's end state; the top-level walk seeds from empty. `CypherExpressionWalker.walk()` now visits `RETURN`/`ORDER BY`/`SKIP`/`LIMIT` *after* the clause loop (using the scope the loop finished with) rather than before it, since `RETURN` is always a statement's/body's last clause. The `SUBQUERY` case now hands the nested body the pre-clause outer scope rather than the post-clause one (which already has the subquery's own output names forgotten out of it for the *outer* walk's benefit - using it to seed the body would seed the body with itself).

Both your example and my variant are now permanent regression tests (`aNameReassignedLaterInABodyDoesNotContaminateAnEarlierPassThroughWithOfTheSameName`, `reviewCommentReproDoesNotFalselyReject`), and the full `com.arcadedb.query.opencypher.**` suite (9021 tests) is green except one documented pre-existing flake unrelated to this change.

## Latent fragility points

- **`CypherVariableUsage.isEdgeVariableReferenced` missing an internal union guard** - agreed, fixed. It now mirrors `innerStatementReferencesVariable`'s guard (any branch referencing the variable keeps the edge binding alive), with a direct unit test (`isEdgeVariableReferencedHandlesATopLevelUnionDirectly`).
- **`CorrelatedSubqueryRunner.planFor()` + `.explain()`/`.profile()`** - agreed this is a real gap, but I'm leaving it out of this PR: it's unreachable today (nothing calls `explain()`/`profile()` on a correlated subquery's own plan - those are only invoked on the outer statement's plan), it predates this PR (this PR doesn't touch `CorrelatedSubqueryRunner` or those two methods at all), and a real fix means giving correlated-subquery UNION bodies actual per-branch explain/profile support, which is a feature addition, not a bug fix in scope here. Happy to file it as a separate low-priority issue if you'd like it tracked.

## Minor points

- `forClauseEntry`'s `null` semantics (stops the whole remaining walk, unlike `forNestedStatement`'s "skip just this body") - documented explicitly now on the method's javadoc.
- Per-clause `HashMap`/`HashSet` allocation in `forClauseEntry` - leaving as is; this runs once at parse/validation time, not on the per-row execution path this codebase optimizes for, so the allocation cost is negligible and the extra "is this a no-op clause type" branching isn't worth the complexity.

lvca:
Thanks for confirming the part-2 self-correction is wired correctly, and for the second pass.

## UNWIND/CALL argument timing

Real, fixed in d281284: `UNWIND`'s list expression, `FOREACH`'s list expression (same class, not called out explicitly but identical reasoning applies), and a procedure `CALL`'s own arguments now check against `before` instead of `after`, matching `WITH`'s items. `CALL`'s `YIELD WHERE` stays on `after` since it's evaluated post-binding, same as `WITH`'s `WHERE`.

Added a regression test (`unwindListExpressionIsCheckedAgainstTheScopeBeforeItsOwnVariableIsRebound`) that reuses an outer node-kind name as the `UNWIND` target and checks a relationship-only function against it inside the list expression - confirmed it fails without the fix and passes with it. Didn't add a dedicated execution test for the `CALL` argument case specifically (constructing one needs a resolvable procedure/`DEFINE FUNCTION` with a same-named `YIELD`, more test scaffolding for the same code path and the same low-severity "missed check, never a false rejection" class you already identified) - the fix is the same one-line change in the same method, covered by the same reasoning.

## Nits

Thanks for double-checking the `SimpleCypherStatement.hasRemove()` refactor and the `CorrelatedSubqueryRunner` gap description - both noted, nothing further to do there.

Full `com.arcadedb.query.opencypher.**` suite: 9023 tests, 0 failures (the one pre-existing flake from before, unrelated, confirmed against unmodified `main`).

lvca:
Good catch, fixed in 8ab616d - `FOREACH`'s inner clauses now advance the visitor between themselves the same way the outer clause loop does, so an inner `CREATE` followed by an inner `SET` reading what it created is checked correctly. Not a regression from this PR (the pre-existing `buildVarTypes` never walked into inner FOREACH declarations at all, so this gap predates part 2), but worth fixing since it's the exact same mechanism used everywhere else here.

Used your own repro as the regression test almost verbatim (`foreachInnerClauseSeesWhatAnEarlierInnerClauseInTheSameBodyJustDeclared`), confirmed it fails without the fix and passes with it.

Full `com.arcadedb.query.opencypher.**` suite: 9024 tests, 0 failures on a clean run (saw one unrelated full-suite-only flake in an intermediate run - `Issue6301AlgoSteinerTreeWeightAlignmentTest`, an algo/Steiner-tree test with no connection to this PR's files - passed both in isolation and on a subsequent full-suite rerun, so treating it as ordering/timing noise rather than something introduced here).

lvca:
All CI is now settled. Summary of the three lanes that showed red:

- `integration-tests`: `Issue6458PortalSuspensionIT` (Postgres wire-protocol portal suspension) - the documented chronic flake, confirmed red on the last 3 completed `main` runs independently of any PR. Zero overlap with this PR's files.
- `slow-unit-tests`: `Issue6655StaleGraphPrefixReuseTest` (documented vector-index rebuild timing flake) plus `MaterializedViewRefreshAtomicityTest` (new instance of the same materialized-view/index-refresh full-suite-timing family as the already-fixed `MaterializedViewEdgeCaseTest` flake) - confirmed 6/6 clean in isolation on this branch, zero file overlap with this PR.
- `ha-integration-tests`: `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`, `DatabaseAreNotIdentical` - the documented chronic Raft-convergence flake (this lane is independently tracked as ~75% intermittent on `main`, issues #5668/#5702).

None of the three touch anything in this PR's diff (Cypher AST/parser/validator only). `claude-review` and CodeRabbit both completed with no remaining actionable points and no blocking issues.

Ready to merge from my side.

Issue #6566, opened 2026-08-22
Surface sparse-vector epochUntrusted through a queryable admin surface, not only a log line


lvca:
### Summary

#6379 gives every sparse-vector segment a persisted `epochUntrusted` flag: true when the segment's precedence predates the recency-epoch fix, or descends from one that did. Such an index can still return a document that was deleted, and the only remedy is a one-off `REBUILD INDEX`.

Today that state is discoverable **only as a `WARNING` log line**, emitted once per engine instance when the index is opened or when a follower first reconciles an untrusted segment. That is enough for someone reading server logs immediately after an upgrade, and nothing else.

### Why a log line isn't sufficient

- **Nothing can alert on it.** Monitoring has no signal to scrape; a rolling upgrade across a fleet would need someone to grep every node's log at the right moment.
- **It is easy to miss and impossible to re-check.** The line is latched to once per engine instance to avoid spamming the query path, so a node that has been up for a week no longer says anything, and there is no way to ask "is this index still affected?" without restarting it.
- **There is no way to confirm the remedy worked.** After running `REBUILD INDEX` an operator has no positive confirmation beyond the absence of a line they may not have seen in the first place.
- **A mixed-version cluster makes it per-node.** During a staged rollout, whether an index is affected depends on which node you ask, which is exactly when a queryable answer matters most.

### Proposed

Expose the flag where tooling can reach it:

- an index-stats/metadata surface (SQL or the HTTP admin API) reporting, per sparse-vector index, whether any active segment is untrusted, and how many;
- ideally the same figure as a Micrometer gauge, so a fleet can alert on "any index still needs a rebuild" rather than relying on log inspection.

`PaginatedSegmentReader.epochUntrusted()` already carries the value and the engine already scans for it in `warnOnPreRecencyEpochSegments`, so this is a plumbing exercise rather than new bookkeeping.

### Context

Raised in code review on #6564 (the #6379 fix). Not a blocker for that PR - the fix itself is complete and the log line covers the immediate upgrade path - but the detection is only as useful as the operator's ability to see it, and right now that requires being in the right place at the right time.

lvca:
Fixed in #6720 (merged into main).

Sparse-vector indexes affected by the recency-epoch bug (#6379) could only be detected by spotting a one-time warning in the server log at startup - there was no way to check later, or to confirm that running `REBUILD INDEX` actually fixed it.

Now this is queryable at any time:
[...]

PR #6743, opened 2026-08-26
perf(#5467): sparse top-K pays for the postings it walks, not the blocks it lands in

### LINKED ISSUE #5467: Sparse vector top-K carries a large fixed per-query cost: 100k tier at 8.94 ms p50 vs 1.45 ms before the 26.8.1 pruning work
Split out of #5388, where it has been an unresolved side thread since the round-1 measurement.

## Summary

`LSM_SPARSE_VECTOR` top-K at the 100k tier costs **8.94 ms p50** where it used to cost **1.45 ms**. The regression appeared in the 26.8.1 development line, has not moved since the Block-Max MaxScore work landed (`a0a79da51`), and is not addressed by the block-payload-copy bound on `issue-5388-block-payload-copy`, which is worth about -12.5% here.

The 1M tier is fine and improving. This is specifically about small and mid-size corpora, where a fixed per-query cost that is invisible at 1M dominates the measurement.

## Timeline (all @tae898's numbers, Big-ANN sparse track, real SPLADE, embedded INT8, settled index, N=5)

| build | 100k p50 | 1M p50 |
|---|---|---|
| dev0 / dev3 | **1.45 ms** | 165 ms |
| dev5 (BMW block-skip) | 23.1 ms | 166 ms |
| dev6 (Block-Max MaxScore) | **8.94 ms** | 53.6 ms |
| dev20 (current line) | **8.94 ms** | 52.49 ms |

MaxScore recovered 23.1 -> 8.94 ms and cut 1M by 3.1x, but 100k is still 6.2x its original cost, and two development cycles have not moved it.

## Why this is a fixed-cost problem, not a scaling problem

Scaling from 100k to 1M went from **114x per 10x of data** on dev0 to **5.9x per 10x** now. The corpus-proportional part of the work is in good shape - it is the constant that grew. The dev0 100k number was cheap because the old exhaustive traversal had almost no setup and a tiny corpus finished before setup could matter; today's traversal has real per-query preparation that a 100k corpus cannot amortise.

Candidate contributors, in the order I intend to measure them:

1. **Per-query cursor setup.** `PaginatedSparseVectorEngine.topK` opens one merged cursor per query term. A real SPLADE query has 30-120 terms, so anything O(terms) with a meaningful constant is paid 30-120 times before a single posting is scored.
2. **MaxScore preparation.** Computing per-term sigma, sorting for the promotion order, and building the essential/non-essential partition is O(terms log terms) plus per-term metadata touches, all before the first candidate.
3. **Merged-cursor construction** across memtable and every active segment, including segments that hold no postings for the term.
4. **Block-header and skip-list touches** that scale with the *index*, not with what pruning ends up decoding. At 1M only about a third of blocks get decoded; the rest of the metadata is still walked.
5. Anything O(active segments) per term when the 100k index settles into a different segment count than the 1M one.

## Why it matters now

Against the same data and ground truth, Qdrant and Milvus return recall@10 = 1.00 at 2.9 ms and 9.3 ms p50 on the 1M tier. Our exact path is at 52.5 ms. Per-query constants that cost 7.5 ms at 100k are the same constants sitting inside the 1M number, so this is not only about the small tier - it is the cheapest available lever on both.

## Scope

- Attribute the 100k per-query cost to specific call sites with a profile at that tier (the existing profile is 1M, where these constants are invisible).
- Remove or amortise whatever is genuinely fixed per query.
- Guard the result with a benchmark that reports a small-corpus tier, so a fixed-cost regression cannot hide behind a healthy 1M number again.

Reporter's harness is staged to re-measure, and the tier builds in seconds.

/cc @tae898

lvca:
Closes #5467 (third round).

The reporter's 30 July CPU profile of released 26.8.1.dev23 put what was left of the learned-sparse
query in **posting traversal at 42% of query CPU**, with scoring already down at 1.5% and the heap
work of round 1 halved. This round goes after that, and delivers the third bullet of the original
scope that neither earlier round did: a benchmark that reports a **small-corpus tier**.

## Getting the tier's shape right mattered more than any single fix

The first version of the harness reused the Zipf document-frequency curve from `MaxScorePruningTest`.
It produced a 100k corpus with the right posting count and completely the wrong traversal: about 10.5
essential postings consumed per candidate against the reporter's 2.76, because a Zipf tail keeps too
many terms fat enough to stay essential. Its profile put 27% of query CPU in a frame that is 0.2% on
the real shape.

Learned-sparse expansion terms form a **plateau**, not a Zipf tail: a handful appear in most documents
carrying small weights, while the query's original vocabulary terms are rare. A Zipf has no plateau,
so tuning it to make the head fat enough makes the tail fat too. Two document-frequency bands
reproduce the reporter's counters, and that is what `Issue5467SmallCorpusTraversalCostTest` builds.

## What changed

Five things the traversal paid for and did not need to. None alters what a query visits or returns,
and there is no format change.

1. **Every block decode re-resolved its page through the page cache.** The builder packs well over a
   hundred default-sized blocks onto one 64 KiB page, so a sequential walk asked for the same page a
   hundred times. Each ask allocated a `PageId` and an `ImmutablePage`, hashed into the cache map and
   bumped the cache's global hit counter, an atomic every parallel-range worker of #5518 contends on.
   The cursor now holds the page across the blocks that live on it.

2. **The block payload was copied into a scratch buffer before decoding.** That copy is the entire
   reason `SegmentFormat.maxBlockPayloadSize` and the "gap to the next block" estimate exist: the
   format does not store a block's byte length, so a copy needed an upper bound. Decoding in place on
   the page needs no bound - the RID loop consumes exactly the bytes the posting count describes - so
   the copy is gone, and `decodedPayloadBytes` now reports the exact length instead of an estimate.

3. **Touching a block decoded all 128 of its weights.** A block-max skip reads none of them, and a
   non-essential probe abandoned early reads none either. Weights are fixed-stride, so a posting's
   weight is one indexed read, taken on demand.

4. **The essential-term heap detached the aligned run with a pop per term and re-attached it with a
   push per term.** The push climbs from a leaf back to wherever the just-advanced cursor now belongs,
   which in a wide merge is near the top, so it is nearly the full height of the heap. The run is a
   connected subtree at the root, so it is located without touching the heap and repaired in place,
   deepest slot first, with one Floyd sift per moved slot.

   This is the idea round 1 tried and discarded as slower. It was slower because that attempt repaired
   with a textbook top-down sift-down, which spends two comparisons per level whatever happens; Floyd's
   spends one per level going down and pays for the climb only in proportion to how high the element
   really belongs, which is the shape a just-advanced cursor has.

5. **The next-essential position was computed for every candidate**, though only a block-max skip needs
   it and the skip fires for a couple of percent of candidates on this data. It is now computed once
   the skip has been decided. The skip also asked the merged cursor for both edges of the same probe
   through two entry points, running the block-bounds memo's range comparisons twice for one bound.

Plus two smaller ones: a non-essential probe consults the scorer's mirrored key before calling into
the cursor at all, and `DimCursor.seekTo` gets the single-source short circuit `advance()` already had.

## Measured

100k tier, 59 terms, top-10, arms interleaved over five rounds so machine drift cancels:

| | before | after | delta |
|---|---|---|---|
| INT8 p50 | 7.76 ms | **7.32 ms** | **-5.7%** |
| FP32 p50 | 7.52 ms | **7.15 ms** | **-4.9%** |

That wall clock is indicative - a laptop with other work on it, and the effect is only a few times the
run-to-run spread. The work counters underneath it are exact, and they are what the tests assert on:

| per query | before | after |
|---|---:|---:|
| page-cache fetches | 2,559 | **103** |
| weights decoded | 325,406 | **140,077** |
| heap key comparisons | 913,455 | **695,016** |
| payload bytes copied out of the page cache | all of them | **none** |

Results are identical in every arm.

## Tests

`Issue5467SmallCorpusTraversalCostTest` carries the harness. Its three non-benchmark tests assert on
**work counters, never on wall clock**:

- a sequential walk fetches each page exactly once, not once per block (asserted against the exact
  number of distinct pages the dim's blocks live on, so the check cannot go vacuous);
- a cursor that navigates without scoring decodes **zero** weights, and one that scores every posting
  decodes exactly one per posting;
- a real Block-Max MaxScore query decodes far fewer weights than the blocks it touched contain, fetches
  far fewer pages than it decoded blocks, and still returns exactly the brute-force top-K.

Verification run:

- 136 sparse-vector tests green (`index.sparsevector` plus the SQL sparse-vector functions).
- Full engine module: **13,541 tests, 1 failure**, and that one is
  `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair`, the
  documented pre-existing full-suite flake that reproduces on unmodified `main` and passes in
  isolation. Nothing here touches OpenCypher.
- `FullTextSparseVectorIndexExportImportIT` green.

## Left on the table, deliberately

The block-max probe is still around 11% of query CPU and almost never pays off on SPLADE, because
near-uniform weights make a block maximum no tighter than the term's global maximum. The obvious move
is to gate or sample the attempt, and that is not in this PR: it is the idea discussed and rejected on
the issue thread, it trades the machinery away on precisely the corpora it exists for (the tail-spike
shapes round 1 targeted), and unlike everything above it is a policy change rather than a cleanup.

## Summary by CodeRabbit

- **Performance**
  - Improved sparse-vector search efficiency through faster term traversal, block skipping, and heap management.
  - Reduced temporary memory use with page reuse and lazy weight decoding.
  - Optimized searches across quantized vector data.

- **Reliability**
  - Strengthened result ordering and search consistency.
  - Improved detection and reporting of corrupted index data, including invalid posting blocks, identifiers, and numeric encodings.

- **Testing**
  - Expanded coverage for search accuracy, traversal costs, quantization, performance, and malformed segment data.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/sparsevector/BmwScorer.java
+++ b/engine/src/main/java/com/arcadedb/index/sparsevector/BmwScorer.java
@@ -243,6 +243,8 @@ public final class BmwScorer {
 
     final int[] heap = new int[n];     // essential term indices, min-heap by current RID
     final int[] aligned = new int[n];  // scratch: the term indices sitting on the current candidate
+    final int[] alignedSlots = new int[n];  // scratch: the heap slots those term indices occupy
+    final long[] frontier = new long[2];    // scratch: (bucketId, position) of the next essential key
     // Cursor positions, mirrored into flat arrays indexed by term. The heap reads a position far
     // more often than a cursor moves - about a dozen comparisons per posting consumed - and reading
     // it off the cursor costs two dependent loads through scattered objects. Mirroring turns each
@@ -292,39 +294,141 @@ public final class BmwScorer {
           && SparseSegmentBuilder.compareRid(candidateBucketId, candidatePosition, endBucketId, endPosition) >= 0)
         return;
 
-      // Detach the whole aligned run so the cursors can be moved without corrupting the heap order.
-      int alignedCount = 0;
-      while (heapSize > 0) {
-        final int top = heap[0];
-        if (keyPositions[top] != candidatePosition || keyBucketIds[top] != candidateBucketId)
-          break;
-        aligned[alignedCount++] = heap[0];
-        heapSize = popHeap(keyBucketIds, keyPositions, heap, heapSize);
-      }
-      final int nextEssentialBucketId = heapSize > 0 ? keyBucketIds[heap[0]] : -1;
-      final long nextEssentialPosition = heapSize > 0 ? keyPositions[heap[0]] : -1L;
+      // Locate the aligned run without touching the heap.
+      final int alignedCount = collectAlignedRun(keyBucketIds, keyPositions, heap, heapSize, candidateBucketId,
+          candidatePosition, aligned, alignedSlots);
 
-      if (!tryBlockMaxSkip(terms, keyBucketIds, keyPositions, aligned, alignedCount, prefix[split], candidateBucketId,
-          candidatePosition, threshold, nextEssentialBucketId, nextEssentialPosition))
+      if (!tryBlockMaxSkip(terms, keyBucketIds, keyPositions, aligned, alignedSlots, alignedCount, prefix[split],
+          candidateBucketId, candidatePosition, threshold, heap, heapSize, frontier))
         scoreCandidate(terms, keyBucketIds, keyPositions, aligned, alignedCount, split, prefix, candidateBucketId,
             candidatePosition, threshold, collector);
 
-      // Re-attach whatever is still live. An exhausted term contributes nothing from here on:
-      // dropping its ceiling tightens every prefix bound, which can only push the split further
-      // right on the next pass.
+      // The moved cursors are still in their slots holding keys that only ever grew, so the heap is
+      // repaired in place, deepest slot first: by then every slot below it is already a valid heap.
+      // An exhausted term has to leave the heap entirely, which a repair cannot express - the heap
+      // would have to shrink - so that (rare: at most once per term per query) case rebuilds instead.
+      // An exhausted term contributes nothing from here on, and dropping its ceiling tightens every
+      // prefix bound, which can only push the split further right on the next pass.
       boolean exhaustedAny = false;
+      boolean anyCursorExhausted = false;
       for (int j = 0; j < alignedCount; j++) {
         final int idx = aligned[j];
-        if (terms[idx].cursor.isExhausted())
+        // The mirror holds -1 for an exhausted cursor, so asking it costs one array load instead of
+        // three dependent dereferences through the term and its cursor.
+        if (keyBucketIds[idx] < 0) {
+          anyCursorExhausted = true;
           exhaustedAny |= terms[idx].clearSigma();
-        else
-          heapSize = pushHeap(keyBucketIds, keyPositions, heap, heapSize, idx);
+        }
       }
+      if (anyCursorExhausted)
+        heapDirty = true;
+      else
+        for (int j = alignedCount - 1; j >= 0; j--)
+          siftDownFromFloyd(keyBucketIds, keyPositions, heap, heapSize, alignedSlots[j]);
       if (exhaustedAny) {
         recomputePrefix(terms, prefix);
         splitDirty = true;
       }
     }
+  }
+
+  /**
+   * Collect the heap slots whose key equals {@code (candidateBucketId, candidatePosition)} - the
+   * traversal's aligned run - <b>without modifying the heap</b>.
+   * <p>
+   * Every slot holding the heap's minimum forms a connected subtree rooted at slot 0: if a slot holds
+   * the minimum then so does its parent, since a parent's key is at most its child's and at least the
+   * global minimum. A breadth-first walk that stops descending at the first non-matching child
+   * therefore visits exactly the run, and it visits it in ascending slot order because a binary heap
+   * array <i>is</i> its own level order - which is what lets the caller repair deepest-slot-first by
+   * walking the result backwards.
+   * <p>
+   * <b>Why not pop and push.</b> Detaching the run with one pop per term and re-attaching
+   * it with one push per term costs, per term, a descent to a leaf plus a climb from a
+   * leaf back to wherever the advanced cursor now belongs - and in a wide merge a just-advanced cursor
+   * usually still belongs near the top, so that climb is nearly the full height of the heap. Repairing
+   * in place skips it: the element starts where it already is, and one Floyd sift puts it back. On a
+   * SPLADE-shaped 100k corpus that is 695k key comparisons per query instead of 913k, a quarter of
+   * them gone (issue #5467).
+   * <p>
+   * An earlier attempt at the same idea repaired with a textbook top-down sift-down and measured
+   * <i>slower</i>, which is recorded on the issue. The reason is the sift, not the collection: a
+   * top-down sift spends two comparisons per level (pick the smaller child, then test it against the
+   * element), while Floyd's spends one per level going down and pays for the climb only when the
+   * element really does belong high up. See {@link #siftDownFromFloyd}.
+   *
+   * @param aligned      out: term indices of the run, in ascending heap-slot order
+   * @param alignedSlots out: the heap slots those terms occupy, parallel to {@code aligned}
+   *
+   * @return the number of terms in the run, always at least 1
+   */
+  private static int collectAlignedRun(final int[] keyBucketIds, final long[] keyPositions, final int[] heap,
+      final int heapSize, final int candidateBucketId, final long candidatePosition, final int[] aligned,
+      final int[] alignedSlots) {
+    aligned[0] = heap[0];
+    alignedSlots[0] = 0;
+    int count = 1;
+    for (int q = 0; q < count; q++) {
+      final int left = (alignedSlots[q] << 1) + 1;
+      if (left >= heapSize)
+        continue;  // a leaf slot, and a heap array is left-packed, so there is no right child either.
+      int term = heap[left];
+      if (keyPositions[term] == candidatePosition && keyBucketIds[term] == candidateBucketId) {
+        aligned[count] = term;
+        alignedSlots[count] = left;
+        count++;
+      }
+      final int right = left + 1;
+      if (right < heapSize) {
+        term = heap[right];
+        if (keyPositions[term] == candidatePosition && keyBucketIds[term] == candidateBucketId) {
+          aligned[count] = term;
+          alignedSlots[count] = right;
+          count++;
+        }
+      }
+    }
+    return count;
+  }
+
+  /**
+   * The smallest key in the heap that is not part of the aligned run, into
+   * {@code frontier[0]} (bucket id) and {@code frontier[1]} (position), or {@code -1} in both when the
+   * run is the whole heap.
+   * <p>
+   * The candidates for that minimum are exactly the children of run slots that are not themselves in
+   * the run: any slot outside the run reaches the root through one of them, and each of those is the
+   * minimum of its own subtree.
+   * <p>
+   * This is computed on demand rather than alongside {@link #collectAlignedRun} because only the
+   * block-max skip needs it, and only once it has already decided to skip. On a learned-sparse query
+   * the skip fires for a couple of percent of candidates - block maxima are no tighter than global
+   * maxima when weights are near-uniform - so folding it into the run walk paid a three-way
+   * comparison per heap child on every candidate to answer a question almost none of them asked
+   * (issue #5467).
+   */
+  private static void nextEssentialKey(final int[] keyBucketIds, final long[] keyPositions, final int[] heap,
+      final int heapSize, final int[] alignedSlots, final int alignedCount, final int candidateBucketId,
+      final long candidatePosition, final long[] frontier) {
+    int bestBucketId = -1;
+    long bestPosition = -1L;
+    for (int q = 0; q < alignedCount; q++) {
+      final int left = (alignedSlots[q] << 1) + 1;
+      for (int child = left, last = Math.min(left + 1, heapSize - 1); child <= last; child++) {
+        final int term = heap[child];
+        final int bucketId = keyBucketIds[term];
+        final long position = keyPositions[term];
+        // A child still sitting on the candidate IS a run slot; the run has not moved yet.
+        if (position == candidatePosition && bucketId == candidateBucketId)
+          continue;
+        if (bestBucketId < 0 || SparseSegmentBuilder.compareRid(bucketId, position, bestBucketId, bestPosition) < 0) {
+          bestBucketId = bucketId;
+          bestPosition = position;
+        }
+      }
+    }
+    frontier[0] = bestBucketId;
+    frontier[1] = bestPosition;
   }
 
   /**
@@ -363,12 +467,23 @@ public final class BmwScorer {
           alive = false;
           break;
         }
+        // The mirrored key answers the probe outright whenever the term's cursor already sits past
+        // this document, which on a dense non-essential list is most of the time: only a cursor that
+        // is still behind the candidate has to be moved. Reading it here instead of calling into the
+        // cursor turns the common case into two loads from arrays that stay in L1 (issue #5467).
+        if (keyBucketIds[i] < 0)
+          continue;  // exhausted; the mirror holds -1 for that.
+        int cmp = SparseSegmentBuilder.compareRid(keyBucketIds[i], keyPositions[i], candidateBucketId, candidatePosition);
+        if (cmp < 0) {
+          terms[i].cursor.seekTo(candidateBucketId, candidatePosition);
+          syncKey(terms, keyBucketIds, keyPositions, i);
+          if (keyBucketIds[i] < 0)
+            continue;
+          cmp = SparseSegmentBuilder.compareRid(keyBucketIds[i], keyPositions[i], candidateBucketId, candidatePosition);
+        }
+        if (cmp != 0)
+          continue;  // this term holds no posting for this document.
         final DimCursor c = terms[i].cursor;
-        if (c.isExhausted())
-          continue;
-        c.seekTo(candidateBucketId, candidatePosition);
-        if (c.isExhausted() || c.currentPosition() != candidatePosition || c.currentBucketId() != candidateBucketId)
-          continue;
         if (c.isTombstone()) {
           alive = false;
           break;
@@ -408,9 +523,9 @@ public final class BmwScorer {
    * @return {@code true} if a block range was skipped, {@code false} if the caller must score.
    */
   private static boolean tryBlockMaxSkip(final DimEntry[] terms, final int[] keyBucketIds, final long[] keyPositions,
-      final int[] aligned, final int alignedCount, final float nonEssentialCeiling, final int candidateBucketId,
-      final long candidatePosition, final float threshold, final int nextEssentialBucketId, final long nextEssentialPosition)
-      throws IOException {
+      final int[] aligned, final int[] alignedSlots, final int alignedCount, final float nonEssentialCeiling,
+      final int candidateBucketId, final long candidatePosition, final float threshold, final int[] heap, final int heapSize,
+      final long[] frontier) throws IOException {
     float bound = nonEssentialCeiling;
     if (bound > threshold)
       return false;  // no headroom at all (threshold still NEGATIVE_INFINITY, or an all-essential query).
@@ -421,7 +536,7 @@ public final class BmwScorer {
       bound += t.queryWeight * t.cursor.blockMaxAt(candidateBucketId, candidatePosition);
       if (bound > threshold)
         return false;
-      final RID be = t.cursor.blockEndAt(candidateBucketId, candidatePosition);
+      final RID be = t.cursor.lastProbedBlockEnd();
       if (be != null && (minBlockEnd == null || SparseSegmentBuilder.compareRid(be, minBlockEnd) < 0))
         minBlockEnd = be;
     }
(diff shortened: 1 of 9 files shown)


lvca:
Both actionable points taken, in `424080ec7`.

**The `lastProbedBlockEnd` contract is now self-checking rather than documented.** You were right that a comment is the weakest possible guard for it. It takes the probe and asserts the memo covers it: free with assertions off, and Surefire runs the suite with them on, so a future caller that slips another `blockMaxAt` or `seekTo` in between fails loudly instead of quietly reading a bound for the wrong block. Verified live rather than assumed - deliberately passing a probe outside the window fails `BmwBlockSkipPruningTest` with `lastProbedBlockEnd(0:-998878) on dim 6: the block-bounds memo does not cover this probe`. Worth noting that `MaxScorePruningTest` does *not* reach that line, because the block-max skip never fires on its corpus; `BmwBlockSkipPruningTest` is the one that exercises it.

**`BmwScorerHeapRepairTest` now pins the repair invariants directly**, against hand-built heaps and an independent brute-force answer, for all three claims you worked through by hand. The input heaps are built with a reference sift-down written in the test rather than with `BmwScorer`'s own heapify, so the code under test does not produce its own input.

I checked it is not vacuous by mutating the production code four ways, each of which it catches:

| mutation | tests failed |
|---|---:|
| repair only the root slot | 3 of 4 |
| do not descend past the root when collecting the run | 2 of 4 |
| drop left children from the run | 2 of 4 |
| let the frontier walk consider run slots | 2 of 4 |

**One correction on your reading.** You called the run-slot skip inside `nextEssentialKey` "a correct no-op given `collectAlignedRun` already captured the whole run". It is load-bearing, and the fourth mutation above is exactly that line removed. A run slot's child can itself be a run slot, and the run's key *is* the candidate, so without the skip the walk returns the candidate as the next key after itself, and the block-max skip then targets the position it is already sitting on. The comment now says that instead of implying it is belt-and-braces.

**On the `buildHeap`-per-exhaustion note:** agreed it is a design note rather than a bug, and I have left it. The bound is tighter than the spread-out case you describe: a term can only exhaust once per query, so the rebuilds are capped at the number of query terms whatever their distribution - at most 59 rebuilds of a ≤59-element heap on this workload, against roughly 89k candidates. If a future profile ever shows `buildHeap` hot, the fix is a localised remove-from-slot rather than a rebuild, but there is no reason to carry that complexity now.

140 sparse-vector tests green with assertions on.

lvca:
Taken in `b1a5f85f0`, and it turns out to be a live knob rather than a hypothetical one: I use `-Dbench.reps` to stretch the measured loop under a profiler, so it does get set by hand. Clamped at 1. Checked the failure mode was real before fixing it and that the fix holds: `-Dbench.reps=0` threw before and now reports `p50 9.76 ms  min 9.76  p90 9.76` off the single rep.

The "Claude Code Review" run on `424080ec7` completed with nothing further to raise, and this was the only finding on the CodeRabbit pass. Waiting on the CI matrix, otherwise this is ready from my side.

lvca:
All four taken, in `d77feb448`. Point 3 turned up an actual (pre-existing) hole rather than just missing coverage.

**1. Page-pinning footprint.** Right that it deserves saying out loud; it is now on the `pageBuf` field rather than only in a benchmark number. The shape of the trade: it is bounded by *open cursors*, not by blocks decoded, so a ~120-term query holds ~120 page images (~7.5 MB) and #5518's fan-out multiplies that by the worker count. What the reference actually buys is narrow - those pages are ones the query is actively reading and are already resident in the shared cache, so the only thing it adds is that an eviction racing the query cannot drop them from under it. And a `REPEATABLE_READ` transaction already pins every page it reads for its whole lifetime, which is strictly more than this.

**2. The assert contract has a test now.** `Issue5467UntrustedBlockAndMemoContractTest#theBlockBoundsMemoContractIsEnforcedWhileTheSuiteRunsWithAssertions`. It also asserts that assertions are *on*, which is the part that actually future-proofs it: if Surefire's config ever changed, a test that merely expected `AssertionError` would start failing for a confusing reason, whereas this one names the cause.

**3. Corruption paths - and this one was more than coverage.** Writing the test surfaced a real hole, and it predates this PR: the block header's posting count is read off disk, the decode arrays are sized to the segment's declared block size, and nothing checked the one against the other. A header claiming more postings than the block size gave `ArrayIndexOutOfBoundsException: Index 16 out of bounds for length 16` from the middle of the decode loop - and the old copy-based decode had exactly the same hole, since the scratch buffer was a page wide and the arrays were not. Now rejected up front with the segment named. Same file covers the varint shift budget, which matters because Java masks shift amounts by 63, so an unguarded loop wraps and decodes garbage rather than failing.

I checked all three guards are actually load-bearing by mutating the production code, since a green test that would also pass with the guard removed is worth nothing here:

| mutation | result |
|---|---|
| drop the posting-count guard | fails with the `ArrayIndexOutOfBoundsException` above |
| drop `shift >= 64` from the varint loop | fails |
| drop the `lastProbedBlockEnd` assert | fails |

**4. The `sourceLive[0]` guard.** You are not missing a case - it is dead. `materializeSingle` sets `sourceLive[0] = false` and `exhausted = true` together, and `seekTo` returns early when `exhausted`, so it can never be false at that point. Removed, which also makes it read the same as `advance()`'s equivalent path, where it never existed.

143 sparse-vector tests green.

lvca:
All four taken, in `c95519801`. The varint one turned into something larger than the byte pattern you named, and I measured the cost before keeping it.

**Overflowing RID varints (major).** You are right about the mechanism: at `shift == 63` only bit 0 of a tenth byte's payload survives, and the discarded bits can wrap the bucket cast or the position addition. Rejecting that one byte pattern is the narrow fix, and I did not take it, because it catches one corruption shape out of many that all end the same way.

The general invariant is already named on the write side. `SparseSegmentBuilder.appendInternal` rejects a non-increasing RID, and `flushBlock` fails loud on a negative delta with this reasoning in the source: *"a negative delta would silently encode as a huge unsigned VarInt and decode to a different RID on read - a corruption far from the cause"*. The read side never checked the same thing. So both decoders now require each posting to be strictly greater than its predecessor, and the varint path additionally rejects a bucket delta outside `int` range, which the cast would otherwise truncate into a plausible-looking forward step.

That subsumes the tenth byte: a value with discarded high bits either still moves forward, in which case it is indistinguishable from any other corrupt-but-ordered payload and no byte-level check would help, or it wraps, and it is now caught. It also catches flipped bits and truncated payloads that no amount of tenth-byte validation would.

**It is on the innermost loop of a performance change, so I measured it rather than assuming.** Arms interleaved over four rounds at the 100k tier:

| round | without the check | with it | delta |
|---|---|---|---|
| 1 | 6.82 ms | 7.39 ms | +8.4% |
| 2 | 6.91 | 6.96 | +0.7% |
| 3 | 6.89 | 6.84 | -0.7% |
| 4 | 7.39 | 7.35 | -0.5% |

Round 1 is the script's first run and is cold; the same slow patch on this box hit the *other* arm in round 4, which is why both arms have exactly one 7.39. Best-case to best-case is 6.82 to 6.84 ms, +0.3%. Free within the noise, so it stays.

**Non-positive posting counts.** Half right, and the half that is right is the more dangerous half. Negative is not reachable - `PaginatedDimMetadata.blockPostingCount` masks with `& 0xFFFF`, so the value is always 0..65535 - but `n == 0` is, and unlike an over-large count nothing runs off the end of anything to signal it: the block simply reports the header's first RID as a posting that is not there. `flushBlock` returns early on an empty block so no writer produces it. Now rejected, with a message that reads correctly for both ends rather than saying "exceeds" about zero.

**Javadoc names.** Both fixed. The `PageId` / `ImmutablePage` references are in `//` comments where `{@link}` does not resolve anyway, so they are now plain `{@code}` names; the test's `BufferUnderflowException` reference is `{@code}` too rather than an import that only Javadoc would use.

Guards checked to be load-bearing by mutation, as before - dropping the ordering check, the int-range check, or the zero-count check each fails the suite. 145 sparse-vector tests green.

lvca:
Both taken, in `1c0da7d33`. You said you were not asking for a change on the first one; I made it anyway, because the specific thing you noticed - `server.sh` passes no `-ea`, so the guard is not merely disabled in production but compiled out - moves it from "documented invariant" to "invariant that exists only where it is already being obeyed". And the failure mode you named is the bad kind: not a crash, but a block end belonging to a different probe feeding an over-aggressive skip that quietly drops a document from the top-K.

So rather than document the sharp edge harder, I removed it. `blockBoundsAt` now returns both edges of one probe - the max as the return value, the end through a caller-owned single-slot array. The memo is still consulted once per probe, which was the whole point of the original change, and there is now no window in which the two halves can disagree, because there is no second call. The assert, the contract it guarded, and the test that pinned its misuse all went with it; in their place is a test that the fused reader agrees with the two single-edge readers it replaced, checked across a walk long enough to cross many block boundaries, which is where a stale memo would start to show.

It costs an array store and load in place of a field read. The benchmark sits in the same band as before, min 6.72-7.14 ms at the 100k tier.

**On `readPage`'s javadoc:** agreed, and worth more than a mental note, because the note as written gave a future reader only half the reason. It now says both conditions explicitly and attributes them: absolute reads only, because `REPEATABLE_READ` shares the page object across callers in the transaction; and the bytes may be read after the call returns *only* because a sealed segment is never written again, so a reader over an append-active component would have to copy.

For the record on the other minor: yes, `scoreCandidate`'s short-circuit is behaviourally identical - `cmp > 0` is the "cursor already past, no posting here" case the old code reached via `currentPosition() != candidatePosition` after a `seekTo` that no-oped, and `cmp == 0` is the hit. The only thing removed is the call.

145 sparse-vector tests green.

lvca:
Taken in `d3102d4fe`, and the finding is correct in the one way that matters: **the ascending-RID check from last round cannot see this case.** I verified rather than reasoned about it - with the new guard removed, nine `0x80` bytes followed by `0x02` decode to a bucket delta of zero, which is perfectly ordered, indistinguishable from a legitimate same-bucket posting, and accepted in silence. So my "a wrapped value either still moves forward or is caught" argument from last round had a hole in it, and this is the hole.

One correction on the proposed fix: rejecting the byte at `shift >= 63` would also reject *legitimate* maximum-width encodings. `writeUnsignedVarLong` emits a tenth byte whenever bit 63 is set, with payload 0 or 1. So the rule is not "no tenth byte" but "a tenth byte carries only bit 63" - payload above 1 is bits the shift discards, and that is exactly checkable rather than heuristic. Checked once per VarLong after the loop rather than once per byte, on a branch that is always false for anything a writer produced.

`Issue5467UntrustedBlockAndMemoContractTest#aVarLongWiderThanALongReadsAsACorruptSegment` pins it with your byte sequence, and I confirmed it is load-bearing by removing the check (test fails, nothing thrown).

**Also closed the same hole in `VarInt.readUnsignedVarLong`.** It has no production callers left on the read path - the decode moved in-place onto the page in this PR - but it is public, and a codec that accepts encodings its own writer cannot produce is a trap for whoever picks it up next.

**On the `final` nitpick:** already the case. `blockBoundsAt(final int bucketId, final long position, final RID[] endOut)` and `tryBlockMaxSkip(..., final long[] frontier, final RID[] blockEndOut)` are both final as written in `1c0da7d33`; that one looks like it was raised against an intermediate state.

146 sparse-vector tests green.

---

Separately, on the two minor notes from the other reviewer in this round, neither of which I have changed:

- **`resolveWeight` branching on `weightStride` rather than the quantization enum** is deliberate (an int compare on the hot path instead of an enum switch through the params object), and I have added a comment saying so and that the mapping lives in the constructor, so a fourth width has to be added in both places.
- **The posting-count check being lazy** - at first decode of a block rather than at metadata-parse time - is worth knowing but I would rather leave it. Moving it into `PaginatedDimMetadata`'s constructor would catch it at dim-open rather than segment-open (metadata is parsed per dim, on demand), so it does not actually buy fail-at-open, and it would cost the per-block message that names which block is corrupt.

lvca:
The `RidCompression.RAW` gap is real and now closed, in `85c10812c`. I checked your grep before acting on it: `RidCompression` appears **nowhere** in `engine/src/test`, so neither the builder's RAW branch nor the cursor's had ever been executed by a test - and this PR rewrote the cursor's from relative to absolute page reads and hung two corruption guards off it.

`Issue5467RawRidCompressionTest` gives it what the varint path has:

- a round trip across many blocks and two bucket ids, pinning every RID and weight the builder wrote;
- a seek walk landing on the first posting at or after each target, including targets that fall in the fixture's gaps and one past the end (which must exhaust, not report a stale position);
- the ascending-order and posting-count guards, exercised by mutating real on-disk bytes rather than unit-testing the arithmetic.

Confirmed the tests actually reach the RAW branch rather than passing for unrelated reasons, by mutating it: removing its ascending check fails one test, and swapping its 4/8-byte field advance fails two.

**One correction, on a claim in your review I would not want left standing.** You wrote that the discarded-high-bits case of a tenth VarLong byte "is correctly subsumed by the ascending-order check rather than needing a separate guard". It is not, and I had believed the same thing when I wrote that check last round. Nine `0x80` bytes followed by `0x02` decode to a bucket delta of **zero** - Java's shift discards the `2` - which is perfectly ordered and passes the ascending check untouched. `d3102d4fe` (pushed just before your review, so you were reading the prior commit) adds the width check for exactly that, and I verified the hole by removing it: the corrupt block decodes in silence. Worth flagging because "the ascending check covers it" was my own reasoning first, and it was wrong.

150 sparse-vector tests green.

lvca:
Correct, and taken in `ea474fe1c`. A VarLong is unsigned, so a value at or above 2^63 arrives as a negative long, `0x8000_0000_0000_0001` narrows to `1`, and that reads as an ordinary step to the next bucket - the ascending check has nothing to object to. Verified by removing the new guard: the corrupt block decodes in silence.

**The same hole exists on the position side, which the review did not name.** An absolute position with bit 63 set rides in on a bucket that genuinely did increase, so the RID compares as greater however negative the position is. Both components are now range-checked before anything narrows or accumulates, with tests for a bucket delta of 2^63+1 and a position of 2^63.

**I have also added something no review asked for, because I would like this to be the last round of it.** Four rounds have now each found a different way for a payload to decode into something wrong, and the pattern is that every guard so far checks the payload *against itself* - widths, ranges, ordering. A payload can satisfy all of those and still be the wrong payload, because a truncated or shifted sequence is usually just as well-ordered as the right one.

The block header names the last RID the builder wrote, which is an independent witness. The decode now requires the payload to end exactly there. That is **one comparison per block, not per posting**, it cross-validates rather than self-validates, and it subsumes most of what the per-posting guards catch plus the shapes they structurally cannot see. Its test bumps the first delta by one, so the sequence stays perfectly ordered and in range and simply ends one posting late - which every other guard accepts and this one rejects.

Both changes mutation-checked. 152 sparse-vector tests green, and the benchmark sits where it did (min 7.23-7.39 ms in the current, busier band on this box; the cross-check is per block, so it is not on the per-posting path at all).

For the record, one boundary I am deliberately not extending into: the block *header* fields themselves (first RID, block max, weight band) are read from the trailer and are not validated beyond what this PR added. That is the #6566 metadata surface rather than the decode path this change rewrote, and the cross-check above already turns a corrupt last-RID into a decode failure rather than silent wrong results.

lvca:
Good catch, and it is a mistake rather than a difference of opinion. Fixed in `15162d9c9`.

`grep -rn lastProbedBlockEnd engine/src/` returned exactly one hit, that javadoc, so the method really was gone and only the description of it survived. The cause is worth naming: the commit that removed the mechanism rewrote that paragraph with a string replace that **silently did not match**, and nothing failed, because a dead `@link` in a comment compiles. I have gone back through the other doc rewrites in this branch for the same failure and this was the only one.

The class javadoc now describes what the file actually covers - each untrusted-payload shape the successive guards were added for (posting count, VarLong width, RID component range, ascending order, and the payload/header end mismatch), and the fused block-bounds reader - and states that every case fails if its guard is removed, which is how each of them was checked.

**Style nit taken too.** `nextEssentialKey`'s two-child walk is now unrolled the same way `collectAlignedRun`'s is directly above it; there was no reason for the two to read differently. Same behaviour, and still covered - dropping the right-child arm fails two of the heap tests.

152 sparse-vector tests green.

lvca:
Both nits taken in `943d28f44`, and the third I could verify rather than change.

**Magic `4`/`8`.** Fair, and it was more than cosmetic: the format already names that width as `SegmentFormat.RID_SIZE_BYTES`, so the check now reads against the constant it is derived from rather than restating it, with a note on why it is `n - 1` postings and not `n` (the first RID lives in the block header).

**The `shift == 70` / `shift == 63` asymmetry.** Also fair - you had to re-derive the equivalence, and so would the next person. The inlined decoder now says out loud that it is the same condition `VarInt.readUnsignedVarLong` tests, one checking before its `shift += 7` and the other after the loop has applied it.

**The benchmark lane, which you flagged as unverifiable from the diff.** Verified: the parent pom sets `<excludedGroups>benchmark</excludedGroups>` as the default for every module, and running the class with no overrides executes **3 of its 4 tests** with no `p50` output - so the wall-clock method really is excluded from the default build and only runs under `-DexcludedGroups=`. That is also how every measurement quoted on this PR had to be taken, which is what made it noticeable.

**On the growing public getter surface:** leaving it. `pageFetchCount()` and `resolvedWeightCount()` follow the `decodedBlockCount()` / `decodedPayloadBytes()` pattern that predates this PR and that `MaxScorePruningTest`, `BmwBlockSkipPruningTest` and `SparseSegmentComponentTest` already assert on, so consolidating into a stats object is a separate refactor across those existing tests rather than something to fold into a performance change.

152 sparse-vector tests green.

lvca:
**CI: `integration-tests` is red, and it is not this PR.** Recording the check rather than asserting it.

Failing: `com.arcadedb.postgres.Issue6458PortalSuspensionIT`, 2 of 9 -
`bindFallbackOnAnUnknownSourceStatementReRunsInsteadOfServingTheStaleExecutedPortal` and
`bindFallbackOnAnUnknownSourceStatementReRunsADeferredCatalogQueryTooNotJustPlainSql`, both on
`[RowDescription again - this must be a fresh execution, not the stale first run replayed]`.

That is the Postgres wire-protocol module's portal-suspension handling. This PR touches
`com.arcadedb.index.sparsevector` and nothing else, so there is no mechanism by which it could reach that code.

More usefully than the argument from file overlap: **the same two test methods, with the same assertion message, fail on unmodified `main` at the exact commit this branch is based on** (`87695796d`, run `32907761996`, job `97997120564`). Same names, same shape, inherited rather than introduced. `main`'s last three completed CI runs are all red.

The lane most likely to catch a real problem here has already passed: **`vector-unit-tests` is green**, and `lint`, `build-and-package`, `builder-tests`, `opencypher-tck-tests`, `studio-e2e-tests` and the client/e2e lanes are all green too. `unit-tests`, `slow-unit-tests` and `ha-integration-tests` are still running - the sparse-vector tests added here run in `unit-tests`, so that is the one still worth waiting on, and I will report it either way.

Locally, on this branch: 152 sparse-vector tests green, and a full engine-module run of 13,541 tests with a single failure, `CypherGAVBoundTargetCardinalityTest`, which is itself a documented pre-existing full-suite flake that reproduces on untouched `main`.

Issue #6770, opened 2026-08-26
ha-integration-tests red again on main, 5 different Raft ITs, same 'different test each time' pattern as #5668


robfrank:
### Summary

`ha-integration-tests` was red on `main` again, with 5 unrelated Raft IT classes each failing with a different symptom in the same run - the same "intermittently red, never the same test twice" pattern as #5668 (closed after #6095/#6112), just recurring.

CI run: https://github.com/ArcadeData/arcadedb/actions/runs/32960831041 (job `ha-integration-tests`, commit `78e63b0473b74e9aa4c2ff1117eb4ba0d74da343`)

| Test | Symptom |
| --- | --- |
| `Issue6221VerifyFanOutGuardIT.aVerifyWithTheClustersRealAddressesStillReportsAgreement` | `DatabaseComparator$DatabaseAreNotIdentical: Types: DB1 6 <> DB2 5` |
| `RaftCommandReadConsistencyIT.readOnlyCommandHonorsReadConsistencyOnFollower` | `DatabaseIsClosedException: graph` |
| `RaftHTTP2ServersCreateReplicatedDatabaseIT.createReplicatedDatabase` | `IOException: Server returned HTTP response code: 500 for URL: http://127.0.0.1:2480/api/v1/command/graph` |
| `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges` | `ReplicationDispatchedTimeoutException: Group commit entry failed: TimeoutException (entry was dispatched to Raft; outcome unknown)` |
| `RaftTimeSeriesWriteReadYourWritesIT.lineProtocolPartialWrite400EmitsUsableCommitIndexBookmark` | `RemoteException` wrapping `DatabaseIsClosedException: graph` |

`check-test-results.py` reported 5 errors out of 238 tests across 123 failsafe report files; none of the 5 failing classes are adjacent to each other in the run order, and none of the failures repeat a symptom already fixed by #6095/#6112 (those addressed `Issue5410AbandonedTicketReleaseIT` and `RaftReplicationMaterializedViewIT` specifically).

### Why file a new issue rather than reopening #5668

#5668's suggested direction ("give the HA ITs a shared wait-until-follower-caught-up helper, use it at every cross-node assertion") was implemented for the two call sites it named, but the failures above are new call sites / new symptom shapes:

[...]

PR #6773, opened 2026-08-26
fix(#6769): GAV overlay masks all parallel edges when only one is deleted

### LINKED ISSUE #5668: ha-integration-tests is intermittently red on main, on a different test each time

### LINKED ISSUE #6769: GraphAnalyticalView: deleting one of several parallel edges masks all of them for getNeighborIds/isConnectedTo
### Summary

`GraphAnalyticalView`'s delta overlay tracks a pending edge deletion by the `(edgeType, source, target)` pair only. When a pair is joined by more than one parallel edge and just one of them is deleted (SYNCHRONOUS update mode), every code path that reads the overlay masks **all** parallel edges between the pair, not just the deleted one:

- `getNeighborIds` / `getVertices` (used by the `GAVExpandAll` Cypher operator)
- `isConnectedTo`

`countEdgesBetween` (used by `GAVExpandInto`) was already safe: it treats *any* deletion touching a pair as "unknown" and falls back to the OLTP edge list, which counts correctly.

### Reproduction

Discovered via the CI failure of `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` (main, run https://github.com/ArcadeData/arcadedb/actions/runs/32960831041, job `unit-tests`):

```
expected: 2L
 but was: 0L
	at com.arcadedb.query.opencypher.CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair(CypherGAVBoundTargetCardinalityTest.java:139)
```

The test builds a 2-vertex cycle where the closing hop has two parallel `SETTLED` edges. Deleting **one** of the two, then re-running `MATCH (a)-[:INITIATED]->(t)-[:SETTLED]->(a) RETURN count(*)`, returned `0` instead of `2`. Root cause (confirmed with instrumentation): the Cypher plan's `GAVExpandAll` step, expanding `SETTLED` edges via `GraphAnalyticalView.getNeighborIds`, returned an empty neighbour array for the pair instead of the one surviving edge, because `DeltaOverlay.copyBaseExcludingDeleted` excluded **both** occurrences of the neighbour in the sorted CSR slice once *any* deletion for that pair was recorded.

The test is flaky rather than deterministically red because the Cypher optimizer only chooses `GAVExpandAll`/`isConnectedTo`-style operators for the closing hop under some cost-estimate conditions; other runs pick `GAVExpandInto`, which was already safe via its "unknown → OLTP fallback" path.

An existing unit test, `GraphAnalyticalViewTest.countEdgesBetweenAnswersUnknownWhenTheOverlayMasksAPair`, documented the buggy behavior as intentional ("deleting one of several parallel edges masks all of them... which is what `isConnectedTo` and `getNeighborIds` do"). That was itself the bug description, not a valid design rationale.

### Fix

- `TxDelta.EdgeDelta` now carries the deleted/added edge's own RID.
- `DeltaOverlay` tracks an exact **count** of deleted edges per `(edgeType, src, tgt)` pair (`deletedEdgesPerType: Map<String, Map<Long,Integer>>`), gated by a per-type `Set<RID>` of already-seen deletions so a deletion replayed across merges (or emitted twice within one `TxDelta`, issue #4587) still counts once. New `DeltaOverlay.countDeletedEdges(type, src, tgt)` accessor.
- `GraphAnalyticalView.copyBaseExcludingDeleted` (feeds `getNeighborIds`/`getVertices`) now walks each contiguous run of parallel neighbours in the sorted CSR slice and excludes only the first *N* occurrences, where *N* is the exact deleted count for that pair.
- `GraphAnalyticalView.isConnectedForType` now compares the base CSR occurrence count for the pair against the exact deleted count, instead of treating any deletion as "not connected".
- `countEdgesBetween`/`countBetweenForType` (via `isEdgeDeleted`) is unchanged: it still conservatively answers "unknown" for a masked pair and lets `GAVExpandInto`'s OLTP fallback count exactly, which was already correct.

Regression tests added:
- `DeltaOverlayTest.distinctParallelEdgesBetweenSamePairAreCountedSeparately` (new)
- `GraphAnalyticalViewTest.countEdgesBetweenAnswersUnknownWhenTheOverlayMasksAPair` (updated to assert the corrected, precise `isConnectedTo`/`getNeighborIds` behavior)
- `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` (existing, now deterministically green - verified 20/20 in a repro loop that previously failed ~10-20% of the time)

### Environment

- Branch: `main` @ `78e63b0473b74e9aa4c2ff1117eb4ba0d74da343`

robfrank:
## What does this PR do?

Fixes `GraphAnalyticalView`'s delta overlay so deleting one of several parallel edges between the same pair of vertices no longer masks the survivors from `getNeighborIds`/`getVertices` (the `GAVExpandAll` Cypher operator) and `isConnectedTo`. Also retags `Issue6655StaleGraphPrefixReuseTest` from `@Tag("slow")` to `@Tag("vector")`.

## Motivation

Triaging a red `main` CI run (https://github.com/ArcadeData/arcadedb/actions/runs/32960831041) surfaced a flaky failure in `CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` (`unit-tests` job) and a timeout in `Issue6655StaleGraphPrefixReuseTest` (`slow-unit-tests` job).

Root cause of the first: `DeltaOverlay` tracked a pending edge deletion keyed on `(edgeType, source, target)` only - a presence flag, not a count. When a pair is joined by two parallel edges and only one is deleted (SYNCHRONOUS update mode), every reader other than `GAVExpandInto`'s `countEdgesBetween` (which deliberately answers "unknown" for a masked pair and falls back to the OLTP edge list) treated the *whole pair* as gone. The Cypher test is flaky rather than deterministically red because the optimizer only sometimes selects the affected operator for a given hop; reproduced reliably (~10-20% of runs) by running the test alongside other `algo.*` tests in the same JVM, confirmed 20/20 green after the fix.

Fix: `TxDelta.EdgeDelta` now carries the edge's own RID, letting `DeltaOverlay` distinguish "the same edge reported deleted twice" (replayed across merges, or emitted twice within one `TxDelta` - must not double-count, issue #4587) from "two distinct parallel edges deleted" (must each count). `deletedEdgesPerType` became an exact per-pair deletion **count** (`Map<String, Map<Long,Integer>>`) instead of a `Set`. `GraphAnalyticalView.copyBaseExcludingDeleted` and `isConnectedForType` now compare the base CSR's occurrence count for a pair against that exact deleted count, excluding only the edges actually gone. `countEdgesBetween`/`countBetweenForType` are deliberately unchanged (still answers "unknown", the existing OLTP fallback is already correct) to keep the change minimal.

The second fix is a mistagging: `Issue6655StaleGraphPrefixReuseTest` spends most of its time waiting (up to 60s) on an async vector-graph rebuild serialized by the JVM-wide `LSMVectorIndex.REBUILD_SEMAPHORE` - exactly the scenario CLAUDE.md's `@Tag("vector")` convention exists for - but was tagged only `@Tag("slow")`, so it convoyed against other slow-lane tests waiting on the same semaphore instead of running in the isolated vector lane.

A second, riding-along correctness fix (found by review): `DeltaOverlay.buildDeletedEdgeCounts` switched from `counts.increment(nodeId)` (once per distinct pair) to `counts.add(nodeId, pairEntry.getValue())` (once per actual deleted edge). This also fixes a latent undercount in `countDeletedOutEdges`/`countDeletedInEdges`, which feed the per-node degree calculations in `GraphAnalyticalView.getDegrees`, whenever 2+ of several parallel edges between the same pair were deleted.

## Related issues

Fixes #6769.

Also investigated the `ha-integration-tests` failures from the same CI run (5 unrelated Raft ITs, heterogeneous symptoms) but did not attempt a fix - filed as #6770 instead, referencing the closed #5668 which covers the same "different test each time" pattern.

Two more follow-ups filed during review, both verified reachable but deliberately kept out of this PR's scope:
- #6775 - the overlay's ADDED-edge neighbor indexes (`getAddedOutNeighbors`/`getAddedInNeighbors`) are not filtered against `deletedEdgesPerType`, so an edge created and deleted within the same overlay window before compaction still shows as a live neighbor via `getNeighborIds`/`isConnectedTo`. Pre-existing, not introduced by this fix.
- #6777 - `DeltaOverlay`'s new RID-keyed deletion dedup (`deletedEdgeRIDsPerType`) assumes an edge's RID is never reused while it sits in that set, but `LocalBucket` deliberately reuses a slot freed by a delete for a later insert ("hole reuse", #5279). Bounded by the same compaction trigger #4587 gates on (so not unbounded), but reachable within one compaction window (default 10,000 net edge deltas) on a busy edge type.

## Additional Notes

- Existing `GraphAnalyticalViewTest.countEdgesBetweenAnswersUnknownWhenTheOverlayMasksAPair` had encoded the pre-fix behavior as intentional ("deleting one of several parallel edges masks all of them... which is what `isConnectedTo` and `getNeighborIds` do"); updated its assertions and docstring to the corrected, precise behavior.
- New regression tests: `DeltaOverlayTest.distinctParallelEdgesBetweenSamePairAreCountedSeparately` and `DeltaOverlayCompactionDedupTest.deleteThenReAddOneOfSeveralParallelEdgesDuringCompactionReplayStaysExact` (the latter added in review to pin that the compaction-replay masking logic composes correctly with the new exact-count tracking even with several parallel edges - confirmed correct on the first try, no code change needed).
- Four sibling vector tests have the same latent `@Tag("slow")` mistagging risk (`Issue6067DeferredCloseRebuildTest`, `Issue6657CloseTimeRebuildPendingStatTest`, `LSMVectorIndexCloseCancelsInFlightBuildTest`, `Issue6713EnsureGraphAvailableUnlockedValidationTest`) - left untouched to keep this PR scoped; worth a follow-up.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graph analytics when deleting parallel edges.
  * Preserved remaining connections when only some duplicate edges are removed.
  * Prevented repeated deletion events from incorrectly reducing edge counts.
  * Improved neighbor filtering and connectivity results after edge updates.
  * Ensured edge additions and deletions are tracked accurately across transactions.
* **Tests**
  * Added regression coverage for duplicate, parallel-edge, and delete/re-add scenarios.
  * Updated vector index test classification and documentation.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/olap/DeltaCollector.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DeltaCollector.java
@@ -80,7 +80,7 @@ class DeltaCollector implements AfterRecordCreateListener, AfterRecordUpdateList
       if (record instanceof Vertex vertex)
         delta.addedVertices.add(new TxDelta.VertexDelta(vertex.getIdentity(), extractProperties(vertex)));
       else if (record instanceof Edge edge)
-        delta.addedEdges.add(new TxDelta.EdgeDelta(edge.getTypeName(), edge.getOut(), edge.getIn()));
+        delta.addedEdges.add(new TxDelta.EdgeDelta(edge.getTypeName(), edge.getOut(), edge.getIn(), edge.getIdentity()));
       scheduleSyncCallback(delta);
     } else {
       scheduleAsyncCallback();
@@ -123,7 +123,7 @@ class DeltaCollector implements AfterRecordCreateListener, AfterRecordUpdateList
       if (record instanceof Vertex vertex)
         delta.deletedVertices.add(vertex.getIdentity());
       else if (record instanceof Edge edge)
-        delta.deletedEdges.add(new TxDelta.EdgeDelta(edge.getTypeName(), edge.getOut(), edge.getIn()));
+        delta.deletedEdges.add(new TxDelta.EdgeDelta(edge.getTypeName(), edge.getOut(), edge.getIn(), edge.getIdentity()));
       scheduleSyncCallback(delta);
     } else {
       scheduleAsyncCallback();
--- a/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
@@ -253,6 +253,10 @@ class DeltaOverlay {
       // compaction trigger (Math.abs(deltaEdgeCount) > threshold, issue #4587) - and, since the per-pair
       // count below feeds copyBaseExcludingDeleted's exclusion budget, it would also mask a second, still
       // live, parallel edge that was never actually deleted (issue #6769).
+      // KNOWN GAP (issue #6777): this assumes an edge's RID is never reused while it sits in this set,
+      // but LocalBucket deliberately reuses a slot freed by a delete for a later insert ("hole reuse",
+      // #5279). A different, unrelated edge landing on the same freed slot and later being deleted too
+      // would collide here and have ITS deletion silently dropped.
       if (newDeletedEdgeRIDs.computeIfAbsent(ed.edgeType, k -> new HashSet<>()).add(ed.rid)) {
         newDeletedEdges.computeIfAbsent(ed.edgeType, k -> new HashMap<>())
             .merge(packEdge(srcId, tgtId), 1, Integer::sum);
--- a/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
@@ -256,7 +256,11 @@ class DeltaOverlay {
       // KNOWN GAP (issue #6777): this assumes an edge's RID is never reused while it sits in this set,
       // but LocalBucket deliberately reuses a slot freed by a delete for a later insert ("hole reuse",
       // #5279). A different, unrelated edge landing on the same freed slot and later being deleted too
-      // would collide here and have ITS deletion silently dropped.
+      // would collide here and have ITS deletion silently dropped. The window this can happen in is
+      // bounded by the same compaction trigger #4587 protects: a full rebuild clears this set entirely
+      // (see the no-delta constructor), so the exposure is at most GraphAnalyticalView.compactionThreshold
+      // (default GraphAnalyticalView.DEFAULT_COMPACTION_THRESHOLD = 10,000) net edge deltas of churn on
+      // one covered edge type - not unbounded, but wide enough to be reachable on a busy graph.
       if (newDeletedEdgeRIDs.computeIfAbsent(ed.edgeType, k -> new HashSet<>()).add(ed.rid)) {
         newDeletedEdges.computeIfAbsent(ed.edgeType, k -> new HashMap<>())
             .merge(packEdge(srcId, tgtId), 1, Integer::sum);
--- a/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
@@ -55,8 +55,16 @@ class DeltaOverlay {
   private final Map<String, Map<Integer, int[]>> outNeighborIndex;
   private final Map<String, Map<Integer, int[]>> inNeighborIndex;
 
-  // Deleted edges per type: edgeType -> set of packed (src << 32 | tgt)
-  private final Map<String, Set<Long>>      deletedEdgesPerType;
+  // Deleted edges per type: edgeType -> packed (src << 32 | tgt) -> number of distinct edges deleted for
+  // that pair. A count rather than a presence flag: two parallel edges between the same pair, only one of
+  // them deleted, must leave the other one discoverable (see #copyBaseExcludingDeleted and issue #6769).
+  private final Map<String, Map<Long, Integer>> deletedEdgesPerType;
+
+  // The deleted edges' own identities, per type. Exists purely to tell a genuinely new deletion apart from
+  // the same edge being reported deleted twice (replayed across merges, or emitted twice within one
+  // TxDelta) when deriving deletedEdgesPerType above - the pair alone cannot make that distinction once a
+  // pair carries more than one parallel edge.
+  private final Map<String, Set<RID>>       deletedEdgeRIDsPerType;
 
   // Per-node deleted edge counts for O(1) lookup: edgeType -> nodeId -> count
   private final Map<String, IntIntHashMap> deletedOutEdgeCounts;
@@ -79,6 +87,7 @@ class DeltaOverlay {
     this.deletedOverflowNodes = new BitSet();
     this.addedEdgesPerType = Collections.emptyMap();
     this.deletedEdgesPerType = Collections.emptyMap();
+    this.deletedEdgeRIDsPerType = Collections.emptyMap();
     this.deletedOutEdgeCounts = Collections.emptyMap();
     this.deletedInEdgeCounts = Collections.emptyMap();
     this.propertyOverrides = Collections.emptyMap();
@@ -97,7 +106,8 @@ class DeltaOverlay {
       final Map<String, Object>[] overflowProperties,
       final BitSet deletedBaseNodes, final BitSet deletedOverflowNodes,
       final Map<String, List<long[]>> addedEdgesPerType,
-      final Map<String, Set<Long>> deletedEdgesPerType,
+      final Map<String, Map<Long, Integer>> deletedEdgesPerType,
+      final Map<String, Set<RID>> deletedEdgeRIDsPerType,
       final Map<String, IntIntHashMap> deletedOutEdgeCounts,
       final Map<String, IntIntHashMap> deletedInEdgeCounts,
       final Map<Integer, Map<String, Object>> propertyOverrides,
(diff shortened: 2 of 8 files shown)


robfrank:
Addressed both non-blocking points from the review.

**Add-then-delete-within-overlay gap**: verified this is real and reachable, not just theoretical - confirmed with a throwaway repro (create vertex + edge, delete that same edge, all in one committed transaction before compaction; `getNeighborIds` still returned the deleted neighbor while `countEdgesBetween` correctly answered "unknown"). It's pre-existing (the overlay-added branch was never deletion-aware, before or after #6769's fix) and mechanically distinct enough from the parallel-edge multiplicity bug this PR fixes to warrant its own issue rather than expanding this PR's scope: filed #6775 with the repro and a suggested direction, and left pointer comments at both affected call sites (`GraphAnalyticalView.getNeighborsFromCSR`'s `ovOut`/`ovIn` branch, `isConnectedForType`'s overlay-added branch).

**`deletedEdgeRIDsPerType` allocation overhead**: agreed it's a necessary tradeoff (need identity-keyed dedup once a pair can have >1 outstanding deletion) and it resets to empty on every compaction, so it doesn't grow unbounded - no action taken.

robfrank:
Added the suggested coverage rather than just trusting the hand-trace: `DeltaOverlayCompactionDedupTest.deleteThenReAddOneOfSeveralParallelEdgesDuringCompactionReplayStaysExact` builds a base CSR with 3 parallel edges for a pair, runs a delete+re-add of one of them through the CSR-aware `merge()` overload, and asserts the reinstated add doesn't inflate `countDeletedEdges` past the single edge actually deleted (still 1, not 3). All 5 tests in the file pass - your trace was correct, this was a coverage gap, not a bug. No production code change needed.

Also agreed and left as-is: the `deletedEdgeRIDsPerType` memory note (bounded by the same compaction threshold as `addedEdgesPerType`, not a new unbounded-growth risk).

robfrank:
Verified this before dismissing or fixing it: confirmed "hole reuse" is real, deliberate LocalBucket behavior (issue #5279, not incidental), so the RID-collision scenario described is technically sound, not theoretical.

Chose not to attempt a fix in this cycle: a correct one has to reconcile two things that currently share RID as their identity for different reasons - dedup is *necessary* for the genuine same-delta replay case #4587 fixed, and *unsafe* across an unrelated edge reusing a freed slot before the next compaction. Figuring out how narrow #4587's actual replay window is (same-transaction retry vs. anything before the next compaction) is a separate investigation that determines what the right fix even looks like, so filed #6777 with the analysis and a suggested direction, and left a pointer comment at `DeltaOverlay.java`'s dedup site rather than let it go undocumented.

This is now 3 cycles with no blocking findings and 3 follow-up issues filed for verified-but-out-of-scope gaps (#6775, #6777, plus the 4 sibling mistagged vector tests noted in the PR description). I think this PR is ready to merge on its own stated scope - happy to keep iterating if there's disagreement.

robfrank:
Addressed all three minor points:

- Called out the `buildDeletedEdgeCounts` degree-count fix explicitly in the PR description as a second, riding-along correctness improvement, since you're right it's easy to miss as a side effect of the `Set`→`Map` change.
- Added the actual bound to the `#6777` comment on `newDeletedEdgeRIDs`: cleared entirely by the same compaction #4587 gates on, so the exposure is at most `DEFAULT_COMPACTION_THRESHOLD` (10,000) net edge deltas of churn on one covered type, not unbounded.
- `countBetweenForType`'s Javadoc now names `#6775` explicitly as the reason it can't mirror `isConnectedForType`'s exact arithmetic (the overlay's added-neighbor branch isn't deletion-filtered), for symmetry with the other two call sites.

That's 4 review cycles with no blocking findings - 2 verified-and-tracked pre-existing/adjacent gaps (#6775, #6777) kept out of scope, plus this cycle's doc-only tightening. I'll leave this for merge review now.

PR #6774, opened 2026-08-26
fix(redis): use JSONArray(String) for bracketed composite keys in HMGET

### LINKED ISSUE #6757: Redis wire: HMGET with a bracketed composite index key throws ClassCastException (String cast to String[])
### Summary
Redis-wire `HMGET` against a persistent index with a bracketed composite key throws `ClassCastException` instead of returning the record: the code casts an `Object` (which is a `String`) to `String[]`. The single-key sibling `HGET` does it correctly, which masks the bug.

### Where (HEAD `af4f67fd2`)
`redisw/src/main/java/com/arcadedb/redis/RedisNetworkExecutor.java:1296-1304` (`getRecords`):
```java
for (final Object key : keys) {
  final String k = key.toString();
  final Object[] compositeKey;
  if (k.startsWith("[")) {
    compositeKey = new JSONArray((String[]) key).toList().toArray();   // key is a String -> ClassCastException
```
The correct form is used by `getRecord` (`:1249`): `new JSONArray(k)` (the `String` constructor).

### Repro
`HMGET mydb.myCompositeIndex "[v1,v2]"` → expected the record for the composite key; actual the command fails with a `-ERR …ClassCastException` reply. `HGET` on the same index with the same key works.

### Suggested fix
`new JSONArray(k)` — use the already-computed `String k`, matching `getRecord`.

### Scope / confidence
High (direct code read at HEAD). Scope: only the persistent-index `HMGET` path with a bracketed composite key.

waterWang:
Fixes #6757

## Problem

`HMGET <db>.<compositeIndex> "[v1,v2]"` throws `ClassCastException` instead of
returning the record. The `getRecords()` path in `RedisNetworkExecutor`
builds the composite key with:

```java
compositeKey = new JSONArray((String[]) key).toList().toArray();
```

The incoming `key` is an `Object` holding a `String` (e.g. `"[v1,v2]"`), so
the cast to `String[]` fails at runtime. The single-key sibling `getRecord()`
(`new JSONArray(key)`) and `RedisQueryEngine.parseIndexKey()` both use the
`JSONArray(String)` constructor, which parses the bracketed key correctly.

## Fix

```java
compositeKey = new JSONArray(k).toList().toArray();
```

(`k` is `key.toString()`), matching `getRecord()` and `parseIndexKey()`.

## Test

Added a composite-index HMGET regression to `RedisWTest.persistentCommands()`:
creates a `CompositeAccount[firstName,lastName]` composite index, inserts 10
documents, and runs `HMGET` with the bracketed key `["first_3","last_3"]`
asserting the returned document. Fails on the old code with
ClassCastException, passes with the fix.

## Co-authored

Issue reporter @ruispereira identified the exact location and provided the
repro in #6757.

Signed-off-by: waterWang <waterWang@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of composite Redis index keys across record lookups, hash commands, and Redis queries.
  * Preserved numeric-looking values as strings when appropriate.
  * Added clearer errors for malformed or empty composite keys.
* **Tests**
  * Added coverage for composite-key operations, numeric and LONG index lookups, and invalid key formats.

### DIFF
--- a/redisw/src/main/java/com/arcadedb/redis/RedisIndexKeys.java
+++ b/redisw/src/main/java/com/arcadedb/redis/RedisIndexKeys.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.redis;
+
+import com.arcadedb.serializer.json.JSONArray;
+
+/**
+ * Parses the textual key an index lookup arrives with over the Redis wire protocol (HGET / HMGET / HEXISTS / HDEL)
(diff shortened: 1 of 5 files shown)


lvca:
Thanks @waterWang !

PR #6779, opened 2026-08-26
fix: HTTP session timeout flake, snapshot barrier reason discriminator, TimeSeries schema-load vanishing type (#6398, #6394, #6356)

### LINKED ISSUE #6356: A TimeSeries type whose sealed store fails to load disappears from the schema, and the database opens as if it were never there
Found while working #6340 (PR #6347), which added the first TimeSeries coverage to `CHECK DATABASE`. This is
a separate defect in the schema **load** path, not in the checker, which is why it was not fixed there.

## What happens

Corrupt one byte of a `.ts.sealed` file's header magic, close the database, reopen it. The database opens
**cleanly** and the TimeSeries type is **gone**:

```
totalTimeSeriesTypes = 0
database.getSchema().existsType("Cpu") = false
CHECK DATABASE warnings = []
```

No exception reaches the caller, no warning, no log line an operator would look for. A type that holds every
sample the application wrote simply is not there any more, and the next write creates a fresh empty one.

## Why

`TimeSeriesSealedStore.loadDirectory()` throws `IOException("Invalid sealed store magic: ...")` on a bad
magic. That propagates through the `TimeSeriesShard` constructor to `TimeSeriesEngine`, so
`LocalTimeSeriesType.initEngine()` throws, and `LocalSchema.readConfiguration()` wraps it
(`LocalSchema.java:1874`):

```java
case "t" -> {
  final LocalTimeSeriesType tsType = new LocalTimeSeriesType(this, typeName);
  tsType.fromJSON(schemaType);
  try {
    tsType.initEngine();
    throw new ConfigurationException("Error initializing TimeSeries engine for type '" + typeName + "'", e);
  }
  ...
  yield tsType;
}
```

`this.types.put(typeName, type)` is only reached after the `yield`, so the throw leaves the type out of the
map. Something upstream then swallows the `ConfigurationException` and lets the open complete - that swallow
is the part worth tracing, because on its own the throw would at least have failed the open loudly.

## Why `CHECK DATABASE` cannot cover this

`DatabaseChecker` walks `database.getSchema().getTypes()`. A type that is not in the schema is not walkable
by anything, so the new TimeSeries pass in #6347 reports nothing here and correctly so - the gap has to be
closed where the type is dropped, not where it would have been checked. #6347 leaves a comment at the test
that would have covered it rather than pinning the current behaviour, since a test asserting "the type
vanishes" would enshrine it.

## The design question

Three defensible answers, and this needs a decision before code:

1. **Refuse the open.** Loudest and safest against data loss - nobody writes into a fresh empty type by
   accident - but one damaged derived file takes the whole database offline, and a sealed store IS derived:
   each HA node rebuilds its own by compaction, so the data is not necessarily lost at all.
2. **Register the type with its engine unavailable**, so it appears in the schema, `CHECK DATABASE` can report
   it, and reads/writes against it fail with a message naming the file. Keeps the blast radius at one type.
3. **Quarantine the sealed file and open with the mutable half only**, which is what the format could actually
   support - the mutable `.tstb` pages are replicated and intact, so the type would come back with its
   uncompacted data and the sealed blocks would be rebuilt.

My preference is (2) as the immediate fix, because it is the one that makes the state *visible* without
deciding what to do about it, and it is what lets the #6347 check report the type at all. (3) is the better
end state but wants its own discussion about when a node may discard derived data.

Whatever is chosen, the swallowed `ConfigurationException` should be tracked down independently: a schema
load that fails to construct a type and then reports success is a bad shape regardless of which type it is.

## Reproduction

Append rows to a TimeSeries type, `compactAll()` so a sealed file exists, close, flip byte 0 of
`<type>_shard_0.ts.sealed`, reopen. See the removed-test comment in
`Issue6340TimeSeriesCheckDatabaseTest` for the exact shape.

### LINKED ISSUE #6394: The snapshot barrier's two failures are one exception type, so a test times them instead - and that timing flakes
Found while running the full engine suite for #6358 (PR #6365). Not a product bug - a test that infers *which*
failure it got from how long it took, and gets it wrong under load.

## The failure

`Issue6125SnapshotBarrierAndCapTest.aFlushThatNeverCompletesFailsTheBarrierInsteadOfHoldingTheGlobalLockForever`,
in a full `mvn -pl engine test` run:

```
java.lang.AssertionError:
[the barrier must block on the in-flight batch]
Expecting actual:
  921L
to be greater than or equal to:
  1000L
```

Green 10/10 when the class is run on its own, and green in CI. Only the full-suite run reproduces it.

## Why the bound is not the problem

The test's own comment is right about stalls:

```java
// It reads raw wall clock on purpose - a stall only makes "at least this long" more true
assertThat(stopwatch.elapsedMs()).as("the barrier must block on the in-flight batch").isGreaterThanOrEqualTo(1_000L);
```

A stall can only push that number UP, so this is correctly exempt from the `StallAwareStopwatch` rule in CLAUDE.md.
Widening or converting the bound would be fixing the wrong thing.

The number came in LOW because the barrier did not do what the test assumed. `SNAPSHOT_BARRIER_MAX_MILLIS` is 5000,
and the fabricated never-completing batch is supposed to be waited out by `waitForCurrentFlushToCompleteUntil`, which
cannot return in under 5 s. 921 ms means `openSnapshot` threw from the OTHER site - `trySuspendUntil` giving up
because another suspender was still resuming, which under a full-suite run (many databases, live flush activity)
is a real and much faster outcome.

So the lower bound did exactly the job its comment claims: it caught the ceiling assertion passing vacuously. It is
the *diagnosis* that is unavailable to it.

## The actual gap

`PageManager.openSnapshot` throws `PageSnapshotException` from two places with opposite operational meanings, and
they are distinguishable only by message text (`PageSnapshotException` carries nothing but a string):

- `trySuspendUntil` fails - "another suspender is still resuming". The code's own comment says the consumer should
  fall back to suspend-and-freeze. **Transient, expected under load, retryable.**
- `waitForCurrentFlushToCompleteUntil` fails - "the in-flight page flush did not complete". The comment calls this
  "a sick disk, not a busy one". **Fatal.**

`assertThatThrownBy(...).isInstanceOf(PageSnapshotException.class)` therefore cannot say which one happened, which
is precisely why the test reaches for elapsed time as a proxy - and the proxy is what flakes.

A caller has the same problem for real: nothing but `getMessage()` separates "retry me" from "your disk is dying".

## Suggested shape

Give the two outcomes a discriminator the code can branch on - a subtype pair, or a reason enum on
`PageSnapshotException` - then have the test assert the reason instead of the duration, and drop the 1000 ms floor
entirely. The upper bound (`assertGaveUpWithin(60_000)`) stays as-is; it is the property under test.

This is the same split **#6132 item 2** asks for on the metrics side ("ideally split by which step gave up"). Doing it
once in the exception serves both: the counter can be tagged from the same discriminator. Filed separately because
the test is red in the `unit-tests` lane today and can be fixed without waiting on the observability work.

### LINKED ISSUE #6398: flaky test: HttpAuthSessionManagerTest.absoluteTimeoutNotResetByAccess leaves 40ms of headroom on a 200ms timeout
Seen while running the `server` module during #6363 (PR #6371). Not related to that change - the test exercises an in-memory session map and touches none of it.

`HttpAuthSessionManagerTest.absoluteTimeoutNotResetByAccess` builds a session manager with a **200 ms** absolute timeout and then spends 160 ms of it in `Thread.sleep` before asserting the session is still alive:

```java
manager = new HttpAuthSessionManager(10_000L, 200L); // 10s idle, 200ms absolute
HttpAuthSession session = manager.createSession(user);

Thread.sleep(80);
assertThat(manager.getSessionByToken(session.getToken())).isNotNull();

Thread.sleep(80);
assertThat(manager.getSessionByToken(session.getToken())).isNotNull();  // <- fails here

Thread.sleep(100);
assertThat(manager.getSessionByToken(session.getToken())).isNull();
```

That leaves **40 ms** of headroom for two sleeps, two map lookups and whatever else the JVM is doing. Under a parallel Maven run it went red on the second assertion (`Expecting actual not to be null`) and passed on a re-run in isolation and in a later full run.

This is the failure mode `CLAUDE.md` calls out and that #6260 built machinery for: a shared JVM late in a large run takes a stop-the-world pause, and any bound with less headroom than the pause becomes a coin flip on the JVM's mood.

## Suggested shape

The assertion is a real one - "reading the session does not reset the absolute timeout" - so it should not be deleted or have its bound loosened into meaninglessness. Two options, in order of preference:

- **Make the timeout injectable** (a clock or time supplier on `HttpAuthSessionManager`) so the test advances time deterministically and asserts the property with no wall clock at all. This is the only version that cannot flake.
- If the clock cannot be injected, widen the absolute timeout so the headroom dwarfs a plausible pause (e.g. 2000 ms absolute with two 200 ms sleeps) and measure with `com.arcadedb.utility.StallAwareStopwatch`, which discounts the JVM-wide stall observed inside the window. Slower, but the property survives.

Worth a sweep of the same class while it is open: `absoluteTimeoutExpiresSession` and its neighbours use the same short-sleep-against-a-short-timeout shape and have the same exposure.

lvca:
## Summary

Three unrelated but simple issues that surfaced as follow-ups from other analyses. Each has its own root cause and its own regression test.

### #6398 — `HttpAuthSessionManagerTest.absoluteTimeoutNotResetByAccess` flake
`HttpAuthSession`/`HttpAuthSessionManager` now accept an injectable clock (`LongSupplier`), defaulting to `System::currentTimeMillis` in production. The whole test class was swept to drive a fake clock instead of `Thread.sleep`, per the issue's own suggestion ("the only version that cannot flake"). Tests are also much faster now (no more real sleeps).

### #6394 — snapshot barrier timeout test infers cause from elapsed time
`PageManager.openSnapshotInternal` has two `PageSnapshotException` throw sites with opposite operational meaning: `SUSPEND_TIMEOUT` (transient, expected under load) and `FLUSH_TIMEOUT` (fatal, a stuck disk). `PageSnapshotException` now carries a `Reason` enum (`NOT_RUNNING`, `SUSPEND_TIMEOUT`, `FLUSH_TIMEOUT`, `OTHER`) so callers/tests can branch on *why* instead of parsing `getMessage()` or inferring it from timing. `Issue6125SnapshotBarrierAndCapTest` now asserts on `getReason()` and drops the 1000ms floor that caused the flake under a full-suite run.

### #6356 — TimeSeries type disappears from schema when its sealed store fails to load
Root cause: `LocalSchema.readConfiguration()` wrapped an `initEngine()` `IOException` (e.g. a corrupt `.ts.sealed` magic) in a `ConfigurationException` that aborted the type's registration, and the method's own outer `catch (Exception e)` logged and swallowed it — so the database reopened cleanly with the type simply gone, no warning, and the next write silently created a fresh empty type.

`DatabaseChecker#checkTimeSeries` already had an unreachable branch for exactly this case (`"the storage engine is not initialised"`, added in #6340/#6347) — the strongest signal for which of the three designs discussed in the issue was intended. Implemented: **register the type with its engine unavailable** rather than refusing the whole database open (the sealed store is HA-derived/rebuildable by recompaction) or discarding the type silently.

- `LocalTimeSeriesType` now tracks `engineUnavailableReason` and exposes `isEngineAvailable()` / `getEngineUnavailableReason()` / `requireEngine()`.
- Read/write call sites that previously would have NPE'd on a null engine (`LocalDatabase.countType`, `DatabaseAsyncExecutorImpl.appendSamples` x2, `SaveElementStep`, `SelectExecutionPlanner`) now fail loudly and by name via `requireEngine()`.
- `TimeSeriesSealedStore.loadDirectory()`'s three header-validation `IOException`s now name the `.ts.sealed` file.
- `DatabaseChecker`'s existing null-engine branch is exercised for the first time.

Verified against a real reproduction (flip byte 0 of a `.ts.sealed` file, close, reopen) with three new tests: the type stays registered and reports its reason, `CHECK DATABASE` reports it, and a write against it fails with a clear message instead of an NPE.

## Test plan
- [x] `HttpAuthSessionManagerTest` + `HttpAuthSessionManagerConcurrencyTest` (server): 11/11 pass, no sleeps
- [x] `Issue6125SnapshotBarrierAndCapTest` + `ExceptionTest` (engine): 44/44 pass
- [x] Full `com.arcadedb.engine.timeseries.**` package: 533/533 pass
- [x] `DatabaseChecker`/`LocalSchema`-adjacent regression tests: 29/29 pass
- [x] `mvn -o -pl engine,network,server,ha-raft,integration -am compile` and `test-compile`: clean

## Files changed
- engine/src/main/java/com/arcadedb/database/LocalDatabase.java
- engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java
- engine/src/main/java/com/arcadedb/engine/PageManager.java
- engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesSealedStore.java
- engine/src/main/java/com/arcadedb/exception/PageSnapshotException.java
- engine/src/main/java/com/arcadedb/query/sql/executor/SaveElementStep.java
- engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
- engine/src/main/java/com/arcadedb/schema/LocalSchema.java
- engine/src/main/java/com/arcadedb/schema/LocalTimeSeriesType.java
- engine/src/test/java/com/arcadedb/engine/Issue6125SnapshotBarrierAndCapTest.java
- engine/src/test/java/com/arcadedb/engine/timeseries/Issue6340TimeSeriesCheckDatabaseTest.java
- engine/src/test/java/com/arcadedb/exception/ExceptionTest.java
- server/src/main/java/com/arcadedb/server/http/HttpAuthSession.java
- server/src/main/java/com/arcadedb/server/http/HttpAuthSessionManager.java
- server/src/test/java/com/arcadedb/server/http/HttpAuthSessionManagerTest.java

Closes #6398, closes #6394, closes #6356.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved time-series behavior when storage engines are unavailable, preserving schema visibility and providing clearer read/write errors.
  - Added file-path details for corrupted time-series storage files and improved database diagnostics.
  - Snapshot failures now identify stopped databases, suspend timeouts, and flush timeouts.
  - Prevented unsupported fallback processing for time-series queries and writes.
  - Improved error formatting for reliable JSON responses.

- **API Improvements**
  - Time-series write responses now identify unavailable measurements.

- **Reliability**
  - Improved authentication session timeout consistency.

- **Tests**
  - Expanded coverage for time-series failures, snapshot reasons, exception details, and session expiration.

### DIFF
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -739,7 +739,7 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
         // Counting samples never loads a record through a bucket, so apply the equivalent per-type read check here.
         checkPermissionsOnType(typeName, SecurityDatabaseUser.ACCESS.READ_RECORD);
         try {
-          return tsType.getEngine().countSamples();
+          return tsType.requireEngine().countSamples();
         } catch (final IOException e) {
           throw new DatabaseOperationException("Error counting TimeSeries samples for type '" + typeName + "'", e);
         }
--- a/engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java
+++ b/engine/src/main/java/com/arcadedb/database/async/DatabaseAsyncExecutorImpl.java
@@ -1530,7 +1530,7 @@ public class DatabaseAsyncExecutorImpl implements DatabaseAsyncExecutor {
   @Override
   public void appendSamples(final String typeName, final long[] timestamps, final Object[]... columnValues) {
     final LocalTimeSeriesType tsType = (LocalTimeSeriesType) database.getSchema().getType(typeName);
-    final TimeSeriesEngine engine = tsType.getEngine();
+    final TimeSeriesEngine engine = tsType.requireEngine();
     final int shardIdx = (int) (tsAppendCounter.getAndIncrement() % engine.getShardCount());
     final int slot = getSlot(shardIdx);
     scheduleTask(slot, new DatabaseAsyncAppendSamples(engine, shardIdx, timestamps, columnValues), true,
@@ -1540,7 +1540,7 @@ public class DatabaseAsyncExecutorImpl implements DatabaseAsyncExecutor {
   @Override
   public void appendSamples(final String typeName, final TimeSeriesRowSource source) {
     final LocalTimeSeriesType tsType = (LocalTimeSeriesType) database.getSchema().getType(typeName);
-    final TimeSeriesEngine engine = tsType.getEngine();
+    final TimeSeriesEngine engine = tsType.requireEngine();
     final int shardIdx = (int) (tsAppendCounter.getAndIncrement() % engine.getShardCount());
     final int slot = getSlot(shardIdx);
     scheduleTask(slot, new DatabaseAsyncAppendSamples(engine, shardIdx, source), true, backPressurePercentage);
--- a/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/engine/DatabaseChecker.java
@@ -726,8 +726,11 @@ public class DatabaseChecker {
       if (engine == null) {
         // The type is in the schema but its engine never initialised, so its files were never opened. Reported
         // rather than skipped: a type nothing can read is exactly the kind of thing this check is asked about.
+        // The reason (when known, issue #6356) names the file that failed to load, so an operator does not have
+        // to go hunting through the server log for it.
+        final String reason = type.getEngineUnavailableReason();
         addWarning("timeseries '" + type.getName() + "': the storage engine is not initialised, so none of its "
-            + "files could be read");
+            + "files could be read" + (reason != null ? ": " + reason : ""));
         corrupted.add(type.getName());
         continue;
       }
--- a/engine/src/main/java/com/arcadedb/engine/PageManager.java
+++ b/engine/src/main/java/com/arcadedb/engine/PageManager.java
@@ -510,6 +510,11 @@ public class PageManager extends LockContext {
    * path is only genuinely unconditional if there is ONE exception type to catch - and these callers run inside
    * {@code executeInReadLock}, which wraps a checked exception into something a {@code catch (PageSnapshotException)}
    * would never match, so a leaked {@code IOException} silently disables the fallback.
+   * <p>
+   * {@link PageSnapshotException#getReason()} tells a caller that wants to distinguish them apart WHY the barrier
+   * gave up: {@code SUSPEND_TIMEOUT} is transient and expected under load, {@code FLUSH_TIMEOUT} is a stuck disk
+   * rather than a busy one. The unconditional fallback above does not need the distinction; diagnostics and tests do
+   * (#6394).
    *
    * @return a handle the caller MUST close, ideally in a try-with-resources: the shadow and any file whose deletion
    *     was deferred for it are released there.
@@ -540,7 +545,8 @@ public class PageManager extends LockContext {
       // NOT TIMED AND NOT COUNTED: NOTHING OF THE BARRIER RAN, AND A CALL THAT REFUSED INSTANTLY WOULD OTHERWISE PULL
       // THE AVERAGE THIS METRIC EXISTS TO REPORT TOWARDS ZERO
       throw new PageSnapshotException(
-          "Cannot open a snapshot of database '" + database.getName() + "': the page manager is not running");
+          "Cannot open a snapshot of database '" + database.getName() + "': the page manager is not running",
+          PageSnapshotException.Reason.NOT_RUNNING);
 
     synchronized (snapshotBarrierLock(database)) {
       // THE CLOCK STARTS HERE, NOT AT THE PUBLIC ENTRY POINT: THIS MONITOR ONLY SERIALIZES BARRIERS ON THE SAME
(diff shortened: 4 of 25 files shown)


lvca:
@coderabbitai review

lvca:
@coderabbitai review

lvca:
@coderabbitai review

Issue #4512, opened 2026-06-08
`GraphAnalyticalView.getNeighborsFromCSR` slow path ignores `DeltaOverlay.isEdgeDeleted`


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.graph.olap.GraphAnalyticalView`

### Summary

The slow path of `getNeighborsFromCSR` (used whenever an overlay is present)
copies the full base adjacency without ever consulting
`DeltaOverlay.isEdgeDeleted(edgeType, src, tgt)`. Edges deleted in the delta
still appear in the returned neighbour list. Meanwhile `countDirectional`
**does** subtract `countDeletedOutEdges`.

### Code

`engine/com/arcadedb/graph/olap/GraphAnalyticalView.java:1481–1510`

### Impact

`getVertices(...)` returns ghost edges. `countEdges()` and
`getVertices(...).length` disagree on the same view.

### Suggested fix

In the slow path, filter base entries through `ov.isEdgeDeleted(...)`, or
materialise a per-(node,edgeType) deleted set on the overlay.

PR #6781, opened 2026-08-26
fix(graph): exclude overlay-added edges deleted in the same window

### LINKED ISSUE #6775: GraphAnalyticalView: an edge added then deleted within the same overlay window before compaction still shows as a live neighbor
### Summary

`DeltaOverlay`'s added-edge indexes (`getAddedOutNeighbors`/`getAddedInNeighbors`, backing `GraphAnalyticalView.getNeighborIds`/`getVertices`/`isConnectedTo`'s overlay-added branch) are never filtered against `deletedEdgesPerType`. An edge created and later deleted, both within the same not-yet-compacted `SYNCHRONOUS` overlay window, still surfaces as a live neighbor.

Found during review of #6773 (fix for #6769) - a related but distinct defect in the same class, out of scope for that PR since it predates it (the overlay-added path was never deletion-aware, before or after #6769's fix, which only added deletion-count-awareness to the base-CSR-vs-overlay comparison paths: `GraphAnalyticalView.copyBaseExcludingDeleted` and `isConnectedForType`).

### Reproduction

Confirmed locally against `main` @ `78e63b0473b74e9aa4c2ff1117eb4ba0d74da343` (pre-#6773 state has the same gap - it is unrelated to that fix):

```java
final GraphAnalyticalView view = GraphAnalyticalView.builder(database)
    .withName("probe").withVertexTypes("N").withEdgeTypes("E")
    .withUpdateMode(GraphAnalyticalView.UpdateMode.SYNCHRONOUS)
    .build();

database.begin();
final MutableVertex c = database.newVertex("N").set("k", "C").save();
final var edge = a.newEdge("E", c).save();
edge.delete();
database.commit();

view.getNeighborIds(idA, Vertex.DIRECTION.OUT, "E"); // returns [idC] - should be []
view.countEdgesBetween(idA, idC, Vertex.DIRECTION.OUT, "E"); // correctly returns -1 (unknown)
```

`countEdgesBetween`/`countBetweenForType` is unaffected - it already returns "unknown" (via `isEdgeDeleted`) for any pair with a recorded deletion, before it ever looks at overlay-added neighbors, so it degrades safely. `getNeighborIds`/`getVertices`/`isConnectedTo` do not have that guard on the overlay-added branch and report the phantom edge as live.

### Suggested direction

`DeltaOverlay.getAddedOutNeighbors`/`getAddedInNeighbors` (or their callers in `GraphAnalyticalView.getNeighborsFromCSR`/`isConnectedForType`) need to exclude an added neighbor that `deletedEdgesPerType`/`countDeletedEdges` also records as deleted for the same pair - conceptually the same run-based exclusion #6769 added for the base CSR side, applied to the overlay-added side too. Since `outNeighborIndex`/`inNeighborIndex` are built once per `merge()` from `addedEdgesPerType`, the cheapest fix is probably filtering at index-build time rather than per-read.

### Related

- #6769 / PR #6773 - where this was found, not fixed there deliberately to keep that PR's blast radius minimal

waterWang:
## What does this PR do?

Fixes #6775. `DeltaOverlay`'s added-edge index is append-only — an edge created and later deleted within the same not-yet-compacted `SYNCHRONOUS` overlay window stays in `addedEdgesPerType` and keeps surfacing as a live neighbour from `getNeighborIds`/`getVertices`/`isConnectedTo` (and `countEdges` for edge types with no base CSR), while `countEdgesBetween` already answers the pair as deleted.

This PR spends the overlay's per-pair deleted budget against the added neighbours, mirroring how `copyBaseExcludingDeleted` spends it against the base CSR run (added in #6773 for #6769): base CSR occurrences absorb the budget first, and the leftover is spent against added occurrences — so an edge added then deleted in the same window disappears, while parallel added edges survive deletion of only some of their siblings, consistent with the exact-count semantics of #6769.

## Motivation

A phantom live neighbour produces invalid BydbQL/query results: mismatch between `getVertices`/`getNeighborIds`/`isConnectedTo` and `countEdgesBetween` on the same view.

## Related issues

- #6775 (fixed)
- #6769 / #6773 (sibling bug: base-CSR side; this closes the overlay-added gap those PRs left annotated)
- #4512 (base-CSR deleted-edge filtering in the slow path)

## Additional Notes

- `getDegrees`/`countDirectional` are untouched: they already compute `added.length - countDeletedOut/InEdges`, which cancels correctly for added-then-deleted edges.
- Added regression test `addedThenDeletedEdgeWithinOverlayWindowIsNotALiveNeighbour`: fails on `main` (`Expecting empty but was: [1]`) and passes with this fix.
- Verified: full `com.arcadedb.graph.olap.*Test` suite — 213 tests, 0 failures.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected graph analytics results when edges are added and deleted before overlay compaction.
  * Edge counts, connectivity checks, neighbor results, and pairwise counts now consistently exclude deleted edges.
  * Preserved accurate handling of parallel edges, replacement edges, and unsorted additions.
  * Prevented replayed operations from creating duplicate edges or masking existing base edges.

* **Tests**
  * Added regression coverage for transactional edge additions and deletions, merges, compaction, and partially deleted parallel edges.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/DeltaOverlay.java
@@ -48,8 +48,12 @@ class DeltaOverlay {
   // Deleted overflow nodes (indexed by overflowIdx = globalId - baseNodeCount)
   private final BitSet                      deletedOverflowNodes;
 
-  // Added edges per type: edgeType -> list of (srcGlobalId, tgtGlobalId) pairs
-  private final Map<String, List<long[]>>   addedEdgesPerType;
+  // Added edges per type: edgeType -> the added edge's own identity -> (srcGlobalId, tgtGlobalId) pair.
+  // Keyed by RID rather than held in a plain list so that an edge added and later deleted within the SAME
+  // not-yet-compacted overlay window can be withdrawn from here by identity, instead of being masked with a
+  // pair-keyed deletion it has no right to spend (issue #6775 - see merge()). Insertion-ordered, because
+  // the neighbour index below is built from it and parallel edges must keep their addition order.
+  private final Map<String, Map<RID, long[]>> addedEdgesPerType;
 
   // Secondary indexes for O(1) neighbor lookup: edgeType -> nodeId -> neighbor list
   private final Map<String, Map<Integer, int[]>> outNeighborIndex;
@@ -58,6 +62,8 @@ class DeltaOverlay {
   // Deleted edges per type: edgeType -> packed (src << 32 | tgt) -> number of distinct edges deleted for
   // that pair. A count rather than a presence flag: two parallel edges between the same pair, only one of
   // them deleted, must leave the other one discoverable (see #copyBaseExcludingDeleted and issue #6769).
+  // Holds ONLY deletions of edges the overlay never added itself, so the count is exactly the exclusion
+  // budget to spend against the BASE CSR's run for the pair - never against the overlay's own additions.
   private final Map<String, Map<Long, Integer>> deletedEdgesPerType;
 
   // The deleted edges' own identities, per type. Exists purely to tell a genuinely new deletion apart from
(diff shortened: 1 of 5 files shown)


lvca:
Thanks @waterWang

PR #6783, opened 2026-08-26
fix(#6758,#6759,#6760,#6761,#6762): robustness batch across Python bindings, polyglot, HA, network and server

### LINKED ISSUE #6758: Python bindings: ColumnBatcher builds its JSON header by unescaped concatenation of column names
### Summary
The Python bindings' columnar batcher builds its JSON header by raw string concatenation of column names, so any column name/alias containing `"`, `\` or a control character produces malformed JSON and breaks the whole batch decode. Its sibling `RowBatcher` does this correctly via `JSONObject`.

### Where (HEAD `af4f67fd2`)
`bindings/python/src/java/com/arcadedb/python/ColumnBatcher.java:232`:
```java
header.append("{\"name\":\"").append(columns[c]).append("\",\"type\":\"").append(type)
      .append("\",\"nulls\":").append(nulls.length).append(",\"bytes\":").append(colBuf.length);
```
`columns[c]` is either the caller-supplied `joinedColumns` (`split(";")`, `:51`) or, when empty, `results.get(0).getPropertyNames()` (`:59`) — i.e. arbitrary query projection names/aliases. Compare `RowBatcher.appendRow` (`RowBatcher.java:50-55`), which goes through `JSONObject` and is correctly escaped.

### Repro
From Python: `db.query("sql", "SELECT foo AS `weird\"name` FROM T").to_columns()` → the Java side emits `{"name":"weird"name","type":…}`, invalid JSON, and the Python header parse / `numpy.frombuffer` fast path fails, aborting the batch. A name containing `;` additionally breaks the `joinedColumns` split.

### Suggested fix
Build the header via `JSONObject`/`JSONArray` as `RowBatcher` does, or JSON-escape `columns[c]` (and `type`) before concatenation; reject/escape `;` in derived names.

### Scope / confidence
High — clear defect, and the correct pattern already exists in the sibling class. Scope: in-process Python bindings only (not a remote or security surface); impact is a broken columnar/dataframe path for unusual-but-legal aliases.

### LINKED ISSUE #6759: Polyglot engine: script parameters leak across invocations, and timeout cancellation is cooperative with guest memory limits disabled
### Summary
Two resource/isolation gaps in the polyglot (JS) query engine, both admin-gated: script parameters leak across invocations because the shared context's bindings are never cleared, and a timed-out script is only cancelled cooperatively (with guest memory limits disabled entirely), so a non-interruptible call can wedge the shared engine.

### Findings (all HEAD `af4f67fd2`)

**1. Parameters/globals leak between commands** — `engine/src/main/java/com/arcadedb/query/polyglot/PolyglotQueryEngine.java:139-145`:
```java
synchronized (polyglotEngine) {
  if (parameters != null && !parameters.isEmpty()) {
    for (final Map.Entry<String, Object> entry : parameters.entrySet())
      polyglotEngine.setAttribute(entry.getKey(), entry.getValue());   // persistent binding, never removed
  }
  final Value result = polyglotEngine.eval(query);
```
`PolyglotQueryEngine` does not override `QueryEngine.isReusable()` (defaults `true`, `QueryEngine.java:152`) and `LocalDatabase.getQueryEngine` caches reusable engines (`LocalDatabase.java:2003-2015`), so one `GraalPolyglotEngine.context` — and its bindings — is shared by every caller on that database.
*Repro:* `database.command("js", "return x;", Map.of("x", 5))` then `database.command("js", "return x;", Map.of())` → the second returns `5` instead of a `ReferenceError`. Values (and any globals a script assigns) are visible to subsequent commands, including from a different caller.
*Fix:* remove the parameter members after each `eval` (e.g. `bindings.removeMember(key)` in a `finally`), or evaluate each command in a fresh child context — scoped so that `registerFunctions`' declared functions still survive, as intended.

**2. Timeout cancellation is cooperative and there is no memory limit** — `PolyglotQueryEngine.java:247-264` (`executeUserCode` → `future.cancel(true)`), pool at `:98-105` (8 threads, queue 10000, `CallerRunsPolicy`), per-command `synchronized (polyglotEngine)` at `:139`; limits at `GraalPolyglotEngine.java:84-93`:
```java
// DISABLED LIMIT BECAUSE THE CONTEXT IS INVOKED MULTIPLE TIMES
//final ResourceLimits limits = ResourceLimits.newBuilder().statementLimit(10000, null).build();
```
Pure guest CPU loops *are* interrupted at GraalVM JS safepoints (covered by `PolyglotQueryTest.timeout()`), but a task blocked in a non-interruptible host call is not — and it holds the single `polyglotEngine` monitor that every command on the database serializes on. Eight such tasks fill the pool; `CallerRunsPolicy` then runs new work on the caller (HTTP worker) thread, which also blocks on the monitor. Separately, with `ResourceLimits` disabled, the CPU timeout does not bound memory: `new Array(1e9)` can OOM the JVM before any safepoint fires.
*Fix:* use GraalVM's supported cancellation (`Context.interrupt(Duration)` / `context.close(true)`) and/or per-execution `ResourceLimits`, and isolate contexts per execution so a wedged one can be discarded.

### Scope / confidence
High on #1 (code read at HEAD; the reuse chain is explicit). Medium on #2's wedge (the interrupt path works for guest loops; the wedge needs a non-interruptible host call) — the disabled memory limit is deliberate and documented, noted because it means the timeout does not bound allocation. **Both are gated behind `UPDATE_SECURITY` (DBA-level)**, so exposure is admin-only; filed as correctness/robustness, not a privilege issue.

### LINKED ISSUE #6760: HA: a snapshot install that gives up on a database still ACKs the snapshot index, clearing the read floor and enabling stale linearizable reads
### Summary
When a leader-driven snapshot install "gives up" on a database that has failed to refresh 3 times in a row, the install still ACKs the snapshot index, writes it as applied for **every** database, clears the stale-read floor and the diverged marks, and returns the installed `TermIndex` to Ratis (which then purges the log). The node re-enters the ready set advertising fully-caught-up while one database is still stale — so LINEARIZABLE / read-your-writes reads of that database can be served from stale state.

### Where (HEAD `af4f67fd2`)
`ha-raft/src/main/java/com/arcadedb/server/ha/raft/DatabaseReconciler.java`:
- `:99` `ACQUIRE_GIVE_UP_AFTER = 3`; `:313-341` `bumpFailureAndShouldRetry` returns `false` once a database has failed 3 consecutive times, so `applyReconcileOutcome` returns `false`.
- `:270-272` reconcile only fails the install **while** that returns `true`:
```java
if (applyReconcileOutcome(plan, outcome))
  throw new IOException("Reconcile from leader had database failure(s); will retry ...");
```
Past the give-up threshold it returns normally, and `ArcadeStateMachine.installSnapshotFromLeader` (`:1355-1384`) unconditionally proceeds:
```java
registerSnapshotMarker(snapshotTerm, snapshotIndex);
lastAppliedIndex.set(snapshotIndex);
writePersistedAppliedIndexForAllDatabases(snapshotIndex);   // :1365 — includes the STALE database
clearStaleSnapshotFloor();                                  // :1369
... raftHA.notifyApplied();
clearDivergedState();                                       // :1380
return installedTermIndex;                                  // Ratis advances applied idx and purges the log
```
The read gate is global, not per-database: `RaftHAServer.getTrustedAppliedIndex` (`:2834-2838`) returns `min(getLastAppliedIndex(), floor)`; with the floor cleared it returns `snapshotIndex` for reads targeting *any* database, so `waitForAppliedIndex` (`:2705`) / `ensureLinearizableFollowerRead` (`:2967`) pass immediately for `readIndex <= snapshotIndex`.

### Failure sequence (3 nodes, follower F behind on database `X`)
1. Ratis drives InstallSnapshot on F; the refresh of `X` fails (disk pressure, CRC/size mismatch, transient fault) and `install()` rolls `X` back to its old, behind copy.
2. Ratis re-drives; `X` fails again. On the **3rd** consecutive failure reconcile returns normally instead of throwing.
3. The install records `snapshotIndex` as applied for all databases incl. `X`, clears the floor and `X`'s diverged mark, and returns to Ratis, which purges the log up to `snapshotIndex` — the entries needed to advance `X` are gone from F.
4. F advertises applied = `snapshotIndex`. A LINEARIZABLE (or RYW with bookmark ≤ `snapshotIndex`) read of `X` routed to F passes the wait instantly and is served from the **stale** copy.
5. Nothing re-arms it: `checkStaleFollower` sees zero lag, the floor was cleared, `checkStuckFollower` is term-based. Recovery waits for the next leader change or the next *write* to `X` (which trips `WALVersionGapException` → re-diverge). On an idle `X` the stale-read window persists.

### Suggested fix
Do not treat a given-up per-database failure as a clean install. Any of: (a) keep the affected database in `divergedDatabases` (clear only databases that actually refreshed) and consult a per-database freshness gate on the read path; (b) refuse to return the installed `TermIndex` when any planned database did not reach `snapshotIndex`, so Ratis retries / the node stays NOT_READY; or (c) at minimum keep a per-database "not-at-snapshotIndex" floor so `getTrustedAppliedIndex` clamps below the missing entries for reads targeting that database (mirroring the #6111 floor, but per-database).

### Scope / confidence
Medium-high that the path exists as described (every line verified at HEAD); the stale-read consequence is traced through the read-gate wiring. **Not executed — HA is not runnable here.** Requires `arcadedb.ha.autoAcquireDatabases=true` (the default) plus one database whose snapshot install fails ≥3 consecutive times while the leader still reports it readable.

### LINKED ISSUE #6761: Network layer never enables TCP keepalive: a half-open authenticated Postgres/Redis connection pins a thread and FD forever
### Summary
The network layer never enables TCP keepalive, and the Postgres/Redis executors deliberately set an infinite socket read timeout after authentication. A half-open connection (peer host crashed or partitioned with no FIN/RST) therefore pins a server thread and its file descriptor **forever** — there is no application heartbeat on those protocols either.

### Where (HEAD `af4f67fd2`)
`network/src/main/java/com/arcadedb/network/binary/Channel.java:42-47` — the constructor sets only `TCP_NODELAY`:
```java
public Channel(final Socket iSocket) throws IOException {
  socket = iSocket;
  socket.setTcpNoDelay(true);
  // THIS TIMEOUT IS CORRECT BUT CREATE SOME PROBLEM ON REMOTE, NEED CHECK BEFORE BE ENABLED
  // timeout = iConfig.getValueAsLong(OGlobalConfiguration.NETWORK_REQUEST_TIMEOUT);
}
```
`grep` for `setKeepAlive` / `SO_KEEPALIVE` across the whole `network/` module returns **none**.

Live callers via `ChannelBinaryServer`: `PostgresNetworkExecutor` sets `socket.setSoTimeout(0)` after auth (`markAuthenticated`, `:200`) and `RedisNetworkExecutor` does the same (`:806`) — correctly, because authenticated clients hold long-lived idle connections.

### Repro
A Postgres or Redis client authenticates, then its host crashes / the link is silently dropped. The server thread is blocked in `in.readXXX()` with `SO_TIMEOUT == 0` and `SO_KEEPALIVE` off, so the OS never probes the dead peer and the read never returns. Thread + FD leak permanently; enough such events exhaust the pool / FD table.

### Suggested fix
Enable `socket.setKeepAlive(true)` in the `Channel` constructor (the network-layer owner of socket options), ideally with a configurable probe interval on modern JDKs, so the OS can tear down dead peers even on infinite-`SO_TIMEOUT` idle connections.

### Scope / confidence
Medium-high — the mechanism is certain (both the missing option and the deliberate infinite timeout verified at HEAD). Whether to treat it as a bug or an accepted tradeoff is a judgment call, but for wire protocols with no application-level heartbeat there is no other liveness backstop.

### Related, same file (smaller)
- `ChannelBinary.close()` (`:278-297`) closes `in` **before** `out`; closing a socket's input stream closes the socket, so the subsequent `out.close()` flush fails and any buffered-but-unflushed bytes are silently dropped (logged at FINE). Low impact today because the live executors flush per message. *Fix:* flush/close `out` first.
- `ChannelBinaryClient` (`:31-69`) leaks the socket FD on every failed connect: the cleanup `catch` handles only `RuntimeException`, while every connect failure path throws a **checked** exception (`IOException`/`SocketException`), and the `if (socket.isConnected())` guard is wrong for cleanup (a failed connect leaves it unconnected but still holding an FD). *Latent* — this class has no production instantiation at HEAD (HA moved to Apache Ratis), so it is a legacy-code defect with no live impact today. *Fix:* `catch (Throwable)` and close unconditionally.

### LINKED ISSUE #6762: Robustness batch: plugin stop-on-failed-start, WebSocket duplicate/authz/backlog, vector nextId on parse failure, non-volatile gRPC channel, MCP input bounds
### Summary
A grouped batch of smaller robustness defects found while auditing lifecycle, caching and event delivery. Each is independently small; they are filed together because none warrants its own issue and all are one-liner-ish fixes.

### Findings (all HEAD `af4f67fd2`)

**1. `PluginManager` never stops a plugin whose `startService()` threw (resource leak).** `server/src/main/java/com/arcadedb/server/plugin/PluginManager.java:241-264` sets `descriptor.setStarted(true)` only after `startService()` returns, so `stopPlugins()` (`:277-298`) skips it via its `!descriptor.isStarted()` guard and `stopService()` is never called. Concrete instance: `AutoBackupSchedulerPlugin.startService()` calls `scheduler.start()` (creating a daemon `ScheduledThreadPool`, `BackupScheduler.java:59`) *before* `scheduleAllDatabases()`; if the latter throws, the pool leaks for the life of the JVM. *Fix:* on exception, call `plugin.stopService()` (guarded) before rethrowing.

**2. WebSocket change events can be delivered twice on a last-unsubscribe / re-subscribe race.** `server/.../http/ws/WebSocketEventBus.java:67-81` removes the last subscriber and the watcher `W1` from the maps *inside* `lockFor(db)`, then calls `W1.shutdown()` (setting `running=false`) **outside** the lock. A concurrent `subscribe(db)` acquiring the lock in that gap creates watcher `W2` and registers its listeners while `W1` is still `running` with its listeners attached, so a commit in that window is enqueued on both and published twice to the (now current) subscriber. *Fix:* set `running=false` while still holding the per-database lock (e.g. a `signalStop()` at removal), leaving only the blocking `runningLock.await()` outside.

**3. WebSocket authorization is checked only at SUBSCRIBE time.** `WebSocketReceiveListener.java:66-75` gates `subscribe` on `user.canAccessToDatabase(database)`; `WebSocketEventBus.publish()` (`:117-125`) re-checks only `subscription.isMatch(event)`, and the `USER` is captured once at handshake (`WebSocketConnectionHandler.java:47`). Revoking a user's database access (or deleting the user) does not stop the event stream until the client disconnects. *Fix:* re-validate on delivery, or invalidate live subscriptions on an ACL/user change — or document it as session semantics.

**4. A slow (but not erroring) WebSocket subscriber accrues an unbounded send backlog.** `WebSocketEventBus.publish()` fans out via fire-and-forget async `WebSockets.sendText(...)`. Backpressure exists only on the producer side (`DatabaseEventWatcherThread`'s bounded queue *drops* events when full, `:59-64`); there is no per-channel outstanding-bytes cap or slow-consumer eviction, so frames pile up in the subscriber's Undertow send buffer. *Fix:* cap per-channel outstanding sends and evict/close a subscriber past a threshold.

**5. Vector index: a failed first-use materialization leaves `nextId` at 0, defeating the id-reuse guard.** `engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java:1215-1224` — `materializeLocations()` catches a parse failure, logs a WARNING, and in its `finally` still sets `locationsMaterialized = true`; `nextId` is only advanced at the end of `loadVectorsFromPages()` (`:3774`). A corrupt/truncated page that throws after some entries were inserted leaves `nextId == 0` with materialization marked done (one-attempt contract), so `allocateVectorId()` — which exists precisely to stop ids restarting at 0 and superseding live entries (javadoc `:1233-1236`) — hands out `0,1,2,…` and `residentLocations.addOrUpdate(0,…)` overwrites a live vector. *Fix:* on the failure path set `nextId` from whatever was read, or mark the index INVALID rather than accepting writes with an uninitialized allocator.

**6. `RemoteGrpcServer.channel` is read without synchronization and is not `volatile`.** `grpc-client/src/main/java/.../RemoteGrpcServer.java` (field ~`:80`, `channel()` `:152-159`) — mutated only under `synchronized` (`start()`/`close()`) but read unsynchronized, so a reader racing `close()`/first `start()` has no happens-before and may observe a stale value. *Fix:* mark `channel` (and `eventLoopGroup`) `volatile`.

**7. MCP tools: two missing input bounds.** `mcp/.../tools/ProfilerStartTool.java:61,70` declares `minimum:1, maximum:3600` for `timeoutSeconds` in its schema but performs no server-side range check before `profiler.start(timeout)` — unlike every sibling tool (`SampleRecordsTool.java:82-83`, `FullTextSearchTool.java:97-98`, `VectorSearchTool.java:101-102`), so `0`/negative/huge values pass through (root-only, so low impact). And `QueryTool.java:72` / `ExecuteCommandTool.java:75` read `limit` with **no upper bound**, accumulating every row into an in-memory `JSONArray` (`QueryTool.java:92-101`), while the vector/sample/full-text tools all cap their windows. *Fix:* re-validate the declared range; clamp `limit` to a maximum as the sibling tools do.

### Scope / confidence
Each verified in source at HEAD. Severities are low: #1/#5 need an exceptional failure path, #2/#3/#4 are narrow-window or session-semantics, #6 is a JMM visibility gap, #7 is admin/authenticated input validation. Concurrency items are reasoned, not executed.

lvca:
Closes #6758, closes #6759, closes #6760, closes #6761, closes #6762.

Five small-to-medium issues that came out of the same audit pass. Each is independent; they are batched because
none of them is large on its own.

> **#6757 dropped from this PR.** It was fixed on main by #6774 while this PR was open, and that fix is the better
> one: it extracts a single `RedisIndexKeys` parser shared by all four index-lookup call sites (`HGET` / `HMGET` /
> `HEXISTS` / `HDEL` and the `redis` query engine) instead of correcting only the copy that had drifted, and it
> answers a malformed composite key with an error reply instead of letting `JSONArray` throw. Its
> `RedisCompositeIndexKeyTest` already covers the HMGET-matches-HGET case, so the test this branch added went with
> it. `redisw/` is now byte-identical to main.

## #6758 - Python bindings: unescaped JSON header in `ColumnBatcher`

The columnar header was built by string concatenation of raw column names, which are arbitrary query projection
aliases. An alias carrying a `"` emitted invalid JSON and killed the decode of the whole batch; one carrying a
`;` was additionally split into two bogus columns by the `';'`-joined column spec that pins the column set across
batches.

The header now goes through `JSONObject`/`JSONArray`, exactly as the sibling `RowBatcher` already does. **Better
than the suggested fix:** rather than escaping the separator, the column spec itself is now a JSON array on both
sides, so the separator problem cannot come back for any character. The legacy `';'` form is still parsed, so an
older Python side paired with a newer bridge jar keeps working.

## #6759 - Polyglot: parameters leaking across invocations, cooperative-only cancellation

`PolyglotQueryEngine` is a reusable engine, so one `GraalPolyglotEngine.context` serves every caller on a
database - and parameters were bound to it and never removed. `command("js", "return x", Map.of("x", 5))` followed
by `command("js", "return x")` returned `5`.

**Better than the suggested fix** (`bindings.removeMember(key)` in a `finally`): a plain removal breaks the case
where a parameter shadows a binding the engine owns - a parameter literally named `database`, or a function
declared through `registerFunctions`. The parameters are now *scoped*: `setAttributes` snapshots whatever each one
displaces and `restoreAttributes` puts it back, removing only names that were not bound before. Both behaviours
are pinned by tests.

Cancellation: a timed-out script is now cancelled through GraalVM's supported `Context.interrupt(Duration)` in
addition to `Future.cancel(true)`, which is what actually unwinds a guest loop and releases the context every
command on the database serialises on; a context that does not come to rest within the grace period is reported
at WARNING. The interrupt is only issued when `cancel(true)` reports the task was really still running, so it
cannot hit whichever command acquired the lock next. Alongside that, `polyglotEngine` is `volatile` and guarded by
a stable lock instead of being its own monitor (`unregisterFunctions()` replaces it, so `synchronized` on the field
was serialising nothing across a replacement), and `unregisterFunctions()` now closes the context it replaces
rather than leaking it.

On the disabled memory limit: it cannot come back. GraalVM CE's only `ResourceLimits` knob is `statementLimit()`,
which is cumulative over a context's life and permanently cancels it once reached, so on a context that serves
every command of a database it would eventually kill the engine on a legitimate workload; CE exposes no allocation
bound at all. The stale `// DISABLED LIMIT BECAUSE...` comment is replaced with that reasoning.

## #6760 - HA: a snapshot install that gives up on a database still ACKs the snapshot index

Confirmed as described. The reconciler stops failing the whole install for a database that has failed 3 times in a
row - correct for the *retry*, since otherwise Ratis re-downloads every healthy database on the follower each pass
because of one bad one. But the install then wrote `snapshotIndex` as applied for **every** database, cleared the
global read floor and the diverged marks, and returned the installed `TermIndex`, after which Ratis purges the
log. The node re-entered the ready set advertising itself fully caught up while one database was still on its old
copy, and a LINEARIZABLE (or RYW with a bookmark below the index) read of that database passed the apply wait
instantly.

**The fix takes all three of the issue's suggestions rather than the minimum**, because each covers a different
half of the problem:

- `reconcileDatabasesFromLeader` now returns the set of databases it gave up on (giving up on the *retry* is not
  the same statement as "installed").
- The install writes the persisted applied index for every database **except** those - leaving them at the
  position they genuinely reached, so `reinitialize()` still re-detects the gap on the next restart.
- Each keeps its diverged mark, so `isResyncInProgress()` stays true and the node does not advertise readiness.
- Each publishes a **per-database** read floor. `getTrustedAppliedIndex(String)` clamps to it, so reads of that
  database fail (LINEARIZABLE) or degrade (RYW) while its healthy co-located databases stay completely unclamped -
  which is the whole reason the give-up threshold exists. The database name is threaded from
  `RaftReplicatedDatabase`, the only production caller of the read gate; the no-arg overloads are unchanged.
- `retryUnfilledSnapshotGap()` (the HealthMonitor backstop, already throttled to one attempt per watchdog interval)
  now treats a per-database floor as an unfilled gap, so recovery is re-armed instead of waiting for the next
  leader change or the next write to that database.

The global `#6111` floor is deliberately left as it is: it is derived from a global signal and has to clamp
everything. This one is derived from a per-database verdict, so it does not.

Not executed against a live cluster (HA is not runnable here); the install path, the bookkeeping and the read gate
are covered by 8 new deterministic tests, 3 of which fail against the previous behaviour.

## #6761 - No TCP keepalive on the wire-protocol sockets

Confirmed. `SO_KEEPALIVE` is now enabled in the `Channel` constructor - the network-layer owner of socket options -
with configurable probe timings (`arcadedb.network.socketKeepAlive{,Idle,Interval,Count}`, defaults detecting a
dead peer about 3 minutes after the connection goes idle). The timings go through `jdk.net.ExtendedSocketOptions`
where the platform supports them, resolved once and defensively so a trimmed jlink image or a native image falls
back to the system-wide defaults rather than failing to serve the connection.

The two smaller items in the same report:

- **`close()` ordering: real, fixed.** Both `ChannelBinary.close()` and `Channel.close()` shut the input side (and,
  in `Channel`, the socket) before the output side, so the `out.close()` that should have flushed the last buffered
  bytes failed instead and dropped them with a FINE log. A test writes without flushing, closes, and reads the
  bytes on the other end; it fails against the old order with an `EOFException`.
- **`ChannelBinaryClient` FD leak: not reproducible, fixed anyway.** The catch really was dead code (it handled
  only `RuntimeException` while every connect failure path throws checked), and the `isConnected()` guard really
  was wrong for cleanup. But on JDK 21 `NioSocketImpl` releases the descriptor itself when `connect()` fails -
  measured at 200 failed connects with `getOpenFileDescriptorCount()`, zero growth - so there is no live leak to
  regress against. The catch is corrected and the reasoning recorded in the comment; no test, because the one I
  wrote first passed against the unfixed code and a vacuous test is worse than none.

## #6762 - Robustness batch

1. **`PluginManager` never stopped a plugin whose `startService()` threw.** Fixed: `stopService()` is called
   (guarded) before the failure is rethrown, and the descriptor stays "not started" so `stopPlugins()` does not
   stop it a second time - both asserted.
2. **WebSocket duplicate delivery on a last-unsubscribe / re-subscribe race.** The watcher is now told to stop
   *inside* the per-database lock (`signalStop()`), leaving only the blocking join outside it. `shutdown()` lost
   its "already stopped, nothing to do" shortcut, which would otherwise skip the join it exists for.
3. **WebSocket authorization was checked only at SUBSCRIBE time.** Re-checked on delivery. The identity captured
   at the handshake is a snapshot - `updateUser` replaces the instance and `dropUser` removes it - so the grant is
   re-resolved by name, which catches a dropped principal as well as a narrowed one.
4. **Unbounded send backlog for a slow subscriber.** Outstanding bytes are now charged per channel and released
   when Undertow reports the frame done; past `arcadedb.server.eventBusMaxPendingBytes` (16 MB, `0` disables) the
   subscription is dropped and the channel closed.
5. **Vector index `nextId` after a failed materialisation.** The one-attempt contract is kept, but the allocator is
   now advanced over whatever was read before the failure - in `materializeLocations()`'s `finally`, so it covers
   every failure path rather than only the one the report named, and in `loadVectorsFromPages()`'s catch for its
   other caller. Without it the next insert supersedes a live vector instead of adding one.
6. **`RemoteGrpcServer.channel`** is `volatile` (with `eventLoopGroup`), and `channel()` reads it once per decision
   - re-reading after the null check let a concurrent `close()` turn the return value into `null` between the two
   reads, surfacing as an NPE inside gRPC.
7. **MCP input bounds.** `profiler_start` re-validates its declared range; `query` and `execute_command` cap
   `limit` (every row is accumulated into an in-memory `JSONArray` first), and the declared schema advertises the
   same window the server enforces.

## Verification

New tests, all of which were run against the unfixed code first where the defect is observable:

| Issue | Test |
|---|---|
| #6758 | `test_core.py::test_to_columns_survives_json_metacharacters_in_aliases` (+ verified against the engine directly, since the bridge classes have no Maven module) |
| #6759 | 3 new cases in `PolyglotQueryTest` |
| #6760 | `Issue6760PartialSnapshotInstallTest` (8), `DatabaseReconcilerTest` (+2) |
| #6761 | `ChannelSocketOptionsTest` (3) |
| #6762 | `PluginManagerTest.aPluginThatFailsToStartIsStoppedAnyway`, `Issue6762WebSocketEventBusTest` (7), `Issue6762VectorIdAllocatorAfterFailedLoadTest`, `Issue6762ToolInputBoundsTest` (5) |

Suites run green: `ha-raft` 969, `server` 835 + `WebSocketEventBusIT` 20, `network` 461, `bolt` 319, `postgresw`
426, `mcp` 300, `grpc-client` 147, `redisw` 78, `engine` 13587 with the single known pre-existing flake
`CypherGAVBoundTargetCardinalityTest.theViewFallsBackToTheEdgeListWhenADeletionMasksThePair` (green in isolation,
unrelated to anything here).

## Follow-up

Four new `GlobalConfiguration` settings are introduced (`arcadedb.network.socketKeepAlive`, `...Idle`,
`...Interval`, `...Count`) plus `arcadedb.server.eventBusMaxPendingBytes`. The settings reference in
`arcadedb-docs` should get a matching entry - happy to prepare that PR separately.

## Summary by CodeRabbit

* **New Features**
  * Added TCP keepalive configuration options for network connections.
  * Added limits for MCP query and command results, plus profiler timeout validation.
  * Added safeguards for WebSocket event delivery, including authorization checks and slow-connection eviction.
  * Added database-scoped consistency handling for high-availability reads and recovery.

* **Bug Fixes**
  * Preserved projection aliases containing special characters in Python and Arrow results.
  * Prevented vector ID reuse after interrupted index loading.
  * Improved polyglot timeout isolation and parameter cleanup.
  * Improved channel, plugin, and snapshot installation cleanup and lifecycle handling.

### DIFF
--- a/bindings/python/src/arcadedb_embedded/results.py
+++ b/bindings/python/src/arcadedb_embedded/results.py
@@ -330,7 +330,9 @@ class ResultSet:
                 break
             total += count
             if first_names is not None and not joined:
-                joined = ";".join(first_names)  # keep column set stable
+                # JSON, not ";".join: a projection alias may legally contain a semicolon, a quote or a
+                # backslash, and the legacy separator split such a name into two bogus columns.
+                joined = json.dumps(first_names)  # keep column set stable
 
         if total == 0:
             return {}
@@ -470,7 +472,8 @@ class ResultSet:
                 break
             total += count
             if first_names is not None and not joined:
-                joined = ";".join(first_names)
+                # See to_columns: the column spec is JSON so odd-but-legal aliases survive the round trip.
+                joined = json.dumps(first_names)
 
         if total == 0 or not first_names:
             return pa.table({})
--- a/bindings/python/src/java/com/arcadedb/python/ColumnBatcher.java
+++ b/bindings/python/src/java/com/arcadedb/python/ColumnBatcher.java
@@ -13,6 +13,8 @@ package com.arcadedb.python;
 
 import com.arcadedb.query.sql.executor.Result;
 import com.arcadedb.query.sql.executor.ResultSet;
+import com.arcadedb.serializer.json.JSONArray;
+import com.arcadedb.serializer.json.JSONObject;
 
 import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
@@ -47,8 +49,8 @@ public final class ColumnBatcher {
   }
 
   public static byte[] nextColumnBatch(final ResultSet rs, final int max, final String joinedColumns) {
-    // empty joinedColumns: derive the column set from the first row
-    String[] columns = joinedColumns.isEmpty() ? null : joinedColumns.split(";");
+    // empty column spec: derive the column set from the first row
+    String[] columns = parseColumnSpec(joinedColumns);
     final List<Result> results = new ArrayList<>(Math.min(max, 1 << 16));
     while (results.size() < max && rs.hasNext())
       results.add(rs.next());
@@ -68,7 +70,11 @@ public final class ColumnBatcher {
     }
     final int count = rows.size();
 
-    final StringBuilder header = new StringBuilder("{\"count\":" + count + ",\"cols\":[");
+    // The header goes through JSONObject/JSONArray rather than string concatenation: a column name is an
+    // arbitrary query projection alias, so a name carrying a quote, a backslash or a control character used to
+    // emit invalid JSON and break the decode of the WHOLE batch on the Python side. RowBatcher already does it
+    // this way; the cost is n objects per batch, not per row.
+    final JSONArray headerColumns = new JSONArray();
     final ByteArrayOutputStream body = new ByteArrayOutputStream(count * n * 8);
 
     for (int c = 0; c < n; c++) {
@@ -227,17 +233,21 @@ public final class ColumnBatcher {
         colBuf = bb.array();
       }
 
-      if (c > 0)
-        header.append(',');
-      header.append("{\"name\":\"").append(columns[c]).append("\",\"type\":\"").append(type)
-          .append("\",\"nulls\":").append(nulls.length).append(",\"bytes\":").append(colBuf.length);
+      final JSONObject columnHeader = new JSONObject();
+      columnHeader.put("name", columns[c]);
+      columnHeader.put("type", type);
+      columnHeader.put("nulls", nulls.length);
+      columnHeader.put("bytes", colBuf.length);
       if (vdim > 0 && (type.equals("f4v") || type.equals("f8v")))
-        header.append(",\"dim\":").append(vdim);
-      header.append('}');
+        columnHeader.put("dim", vdim);
+      headerColumns.put(columnHeader);
       body.write(nulls, 0, nulls.length);
       body.write(colBuf, 0, colBuf.length);
     }
-    header.append("]}");
+
+    final JSONObject header = new JSONObject();
+    header.put("count", count);
+    header.put("cols", headerColumns);
 
     final byte[] headerBytes = header.toString().getBytes(StandardCharsets.UTF_8);
     final byte[] bodyBytes = body.toByteArray();
@@ -247,5 +257,27 @@ public final class ColumnBatcher {
     out.put(headerBytes);
     out.put(bodyBytes);
     return out.array();
+  }
+
+  /**
+   * Decodes the caller-supplied column spec, which pins the column set across the batches of one result set.
+   * <p>
+   * The canonical form is a JSON array, because the names are arbitrary projection aliases: the legacy
+   * {@code ';'}-joined form silently split a name that contained a semicolon into two bogus columns. The legacy
+   * form is still accepted so an older Python side paired with a newer bridge jar keeps working.
+   *
+   * @return the pinned column names, or {@code null} when the spec is empty and the set must be derived from the
+   *         first row
+   */
+  private static String[] parseColumnSpec(final String spec) {
+    if (spec == null || spec.isEmpty())
+      return null;
+    if (spec.charAt(0) != '[')
+      return spec.split(";");
+    final JSONArray names = new JSONArray(spec);
+    final String[] columns = new String[names.length()];
+    for (int i = 0; i < columns.length; i++)
+      columns[i] = names.getString(i);
+    return columns;
   }
 }
--- a/bindings/python/tests/test_core.py
+++ b/bindings/python/tests/test_core.py
@@ -1113,6 +1113,32 @@ def test_to_columns_typed_bulk_materialization(temp_db_path):
         assert db.query("sql", "SELECT FROM T WHERE n > 99").to_columns() == {}
 
 
+def test_to_columns_survives_json_metacharacters_in_aliases(temp_db_path):
+    """A projection alias is arbitrary text, so the columnar header must be real JSON (issue #6758).
+
+    The Java side used to concatenate the alias straight into the header string, so a quote produced
+    invalid JSON and killed the decode of the whole batch; a semicolon additionally broke the
+    ``;``-joined column spec that pins the column set across batches.
+    """
+    pytest.importorskip("numpy")
+    weird = 'we"ird'
+    semi = "a;b"
+    with arcadedb.create_database(temp_db_path) as db:
+        db.command("sql", "CREATE DOCUMENT TYPE T")
+        with db.transaction():
+            for i in range(3):
+                db.command("sql", "INSERT INTO T SET n = ?, s = ?", i, f"v{i}")
+
+        cols = db.query(
+            "sql",
+            "SELECT s AS `{}`, n AS `{}` FROM T ORDER BY n".format(weird, semi),
+        ).to_columns()
+        assert cols is not None
+        assert sorted(cols.keys()) == sorted([weird, semi])
+        assert list(cols[weird]) == ["v0", "v1", "v2"]
+        assert list(cols[semi]) == [0, 1, 2]
+
+
 def test_to_dataframe_fast_path(temp_db_path):
     """to_dataframe uses the columnar path and yields typed dtypes."""
     pytest.importorskip("pandas")
--- a/bindings/python/tests/test_core.py
+++ b/bindings/python/tests/test_core.py
@@ -1129,10 +1129,13 @@ def test_to_columns_survives_json_metacharacters_in_aliases(temp_db_path):
             for i in range(3):
                 db.command("sql", "INSERT INTO T SET n = ?, s = ?", i, f"v{i}")
 
+        # batch_size=2 so the three rows span MORE THAN ONE non-empty batch: the column spec is only sent back
+        # to Java from the second batch onwards, so a single-batch result would never exercise the round trip
+        # that the ';'-joined form used to break.
         cols = db.query(
             "sql",
             "SELECT s AS `{}`, n AS `{}` FROM T ORDER BY n".format(weird, semi),
-        ).to_columns()
+        ).to_columns(batch_size=2)
         assert cols is not None
         assert sorted(cols.keys()) == sorted([weird, semi])
         assert list(cols[weird]) == ["v0", "v1", "v2"]
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1214,6 +1214,28 @@ public enum GlobalConfiguration {
 
   NETWORK_SOCKET_TIMEOUT("arcadedb.network.socketTimeout", SCOPE.SERVER, "TCP/IP Socket timeout (in ms)", Integer.class, 30000),
 
+  NETWORK_SOCKET_KEEP_ALIVE("arcadedb.network.socketKeepAlive", SCOPE.SERVER, """
+      Enable TCP keepalive (SO_KEEPALIVE) on every wire-protocol socket. The Postgres and Redis executors drop the
+      socket read timeout to infinite once a connection is authenticated, because an authenticated client legitimately
+      holds an idle connection open, and those protocols carry no application-level heartbeat. With keepalive off, a
+      peer that dies without a FIN/RST (host crash, silent partition) leaves the server thread blocked in read()
+      forever, leaking a thread and a file descriptor per event. Keepalive lets the OS discover the dead peer and fail
+      the read.""", Boolean.class, true),
+
+  NETWORK_SOCKET_KEEP_ALIVE_IDLE("arcadedb.network.socketKeepAliveIdle", SCOPE.SERVER, """
+      Seconds an authenticated connection may sit idle before the OS sends the first TCP keepalive probe. Only applied
+      when the JDK and the platform expose TCP_KEEPIDLE (Linux and macOS do); elsewhere the system-wide default
+      applies, which is typically 2 hours. 0 leaves the system default in place.""", Integer.class, 120),
+
+  NETWORK_SOCKET_KEEP_ALIVE_INTERVAL("arcadedb.network.socketKeepAliveInterval", SCOPE.SERVER, """
+      Seconds between TCP keepalive probes once the first one has gone unanswered (TCP_KEEPINTERVAL). 0 leaves the
+      system default in place.""", Integer.class, 15),
+
+  NETWORK_SOCKET_KEEP_ALIVE_COUNT("arcadedb.network.socketKeepAliveCount", SCOPE.SERVER, """
+      Number of unanswered TCP keepalive probes after which the connection is declared dead (TCP_KEEPCOUNT). With the
+      defaults a dead peer is detected about 3 minutes after the connection goes idle. 0 leaves the system default in
+      place.""", Integer.class, 4),
+
   NETWORK_MAX_PREAUTH_CONNECTIONS("arcadedb.network.maxPreAuthConnections", SCOPE.SERVER, """
       Maximum number of connections a binary wire-protocol listener (Postgres, Redis, BOLT) may hold in the phase
       before authentication. Each accepted socket costs one thread and one file descriptor before the client has
@@ -1443,6 +1465,13 @@ public enum GlobalConfiguration {
   // SERVER WS
   SERVER_WS_EVENT_BUS_QUEUE_SIZE("arcadedb.server.eventBusQueueSize", SCOPE.SERVER,
       "Size of the queue used as a buffer for unserviced database change events.", Integer.class, 1000),
+
+  SERVER_WS_EVENT_BUS_MAX_PENDING_BYTES("arcadedb.server.eventBusMaxPendingBytes", SCOPE.SERVER, """
+      Maximum number of bytes of change-stream frames that may be outstanding towards a single WebSocket subscriber
+      before it is evicted. Frames are sent asynchronously, so a subscriber that never reads accumulates them in the
+      server's send buffer: the producer-side queue is bounded but a slow consumer is charged to the server's heap,
+      not to its own. Past this cap the subscription is dropped and the channel closed, which is what the client
+      would experience anyway. 0 disables the cap (the pre-26.9.1 behaviour).""", Long.class, 16 * 1024 * 1024L),
 
   // SERVER SECURITY
   SERVER_SECURITY_ALGORITHM("arcadedb.server.securityAlgorithm", SCOPE.SERVER,
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -1219,6 +1219,12 @@ public class LSMVectorIndex implements Index, IndexInternal {
             .log(this, Level.WARNING, "Could not load vectors from pages for index %s: %s", indexName, e.getMessage());
         this.graphState = GraphState.LOADING;
       } finally {
+        // The load may have thrown AFTER inserting entries, and the failure above is swallowed by design (one
+        // attempt, carry on with whatever was read). Leaving the allocator at 0 on top of those entries is exactly
+        // what allocateVectorId() exists to prevent: the next put() would hand out 0, 1, 2... and addOrUpdate
+        // would supersede live vectors instead of adding new ones (issue #6762). A no-op on the success path,
+        // where loadVectorsFromPages() already computed the same value.
+        syncNextIdWithResidentLocations();
         // Written last, and volatile: a thread that reads it true has a happens-before edge on every entry the
         // loop above put into the map, which is what lets vectorIndex()'s fast path hand the instance out with no
         // lock at all. Set even when the load threw, so the one-attempt contract above holds.
@@ -3753,7 +3759,6 @@ public class LSMVectorIndex implements Index, IndexInternal {
               getTotalPages(), residentLocations.size());
 
       int entriesRead = 0;
-      int maxVectorId = -1;
 
       // Load from compacted sub-index first (if it exists)
       if (compactedSubIndex != null) {
@@ -3772,8 +3777,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
       // Compute nextId from the maximum vector ID found across both files. Tombstoned ids are not resident
       // (issue #5516) but they were still handed out, so the location index's own high-water mark - which every
       // addOrUpdate advances, tombstones included - is what guarantees an id is never reused.
-      maxVectorId = residentLocations.getAllVectorIds().max().orElse(-1);
-      nextId.set(Math.max(maxVectorId + 1, residentLocations.getNextId()));
+      syncNextIdWithResidentLocations();
 
       LogManager.instance().log(this, Level.FINE,
           "loadVectorsFromPages DONE: Loaded " + residentLocations.size() + " vector locations (" + entriesRead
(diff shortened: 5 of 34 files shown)


lvca:
Thanks — five of the six are fixed in `0dfec74c2`; one declined with a reason.

**Fixed**

1. **`ArcadeStateMachine` — publish the per-database floor before waking waiters.** Correct, and the most important
   finding here: my own ordering reopened the exact window #6760 closes. `clearStaleSnapshotFloor()` had already
   dropped the global floor and `clearDivergedState()` had wiped the per-database one, so a waiter woken in that gap
   read the raw Ratis index — already `snapshotIndex` — and a LINEARIZABLE / RYW read of a database the install did
   *not* refresh passed its wait. `notifyApplied()` now runs last, after every floor and mark is final.
   Rather than the suggested threaded test, `theDatabaseFloorIsPublishedBeforeAnyWaiterIsWoken` captures the floor
   *at the instant of the notify* via a stubbed `notifyApplied()`, so it is deterministic instead of racing —
   verified to fail against the old ordering and pass against the new one.

2. **UTF-8 byte length for the WS backlog.** Right — the setting is named `...MaxPendingBytes` and what occupies the
   send buffer is what `sendText` encodes. Counting chars let a CJK/emoji stream hold 3–4x the budget. Computed in a
   single pass rather than `getBytes(UTF_8).length`, which would copy the whole payload once per event on the
   watcher thread. Covered by `theBacklogIsChargedInUtf8BytesNotCharacters`.

3. **Reservation released against a replacement subscription.** Real, but I fixed the consequence rather than the
   lookup: `releasePending` now floors at zero. Capturing the subscription would mean allocating a callback per
   subscriber per event, which the shared callback was deliberately introduced to avoid. With the floor, a late
   completion landing on a replacement is a no-op instead of a negative baseline that would silently raise that
   subscriber's cap; the original's bytes die with the discarded object. Covered by
   `aReleaseOnAReplacementSubscriptionCannotDriveItsBacklogNegative`.

4. **Polyglot: interrupt only the context the task actually entered.** Correct, and worse than described — if our
   task was still *queued*, `future.cancel(true)` returns `true` and we would then have interrupted the context
   belonging to **another caller's** running script. The engine instance is now recorded as the task enters the
   synchronized block; `null` means it never ran and nothing is cancelled. This also stops us interrupting a context
   `unregisterFunctions()` replaced in the meantime.

5. **gRPC `channel()` vs `close()`.** Added a `closing` flag set as `close()` begins. Beyond the terminating-channel
   window you named, there was a worse one: once `close()` nulled the field, `channel()` would hit `channel == null`
   and **silently build a brand-new channel** for a server the caller had explicitly closed. An explicit `start()`
   clears the flag, so a deliberate restart still works. Covered by `Issue6762ChannelLifecycleTest` (3 cases,
   including the volatile-modifier assertion).

**Declined**

6. **`ChannelBinaryClient` regression test.** The production change stays, but I'm not adding the test, because it
   would be vacuous. On JDK 21 `NioSocketImpl` releases the descriptor itself when `connect()` fails — I measured
   200 failed connects through `getOpenFileDescriptorCount()` and saw zero growth — so "socket is closed and the
   original `IOException` propagates" holds *identically before and after* the fix. The socket is created inside the
   constructor via `SocketFactory`, so there is no seam to force a mid-construction failure without changing
   production code purely to be testable. The change closes a genuinely dead catch (it handled only
   `RuntimeException` while every path there throws checked) and a wrong `isConnected()` cleanup guard; that
   reasoning is now in the code comment and the PR body. Happy to revisit if you'd rather have the seam.

Full suites re-run locally after the fixes: engine 13589, ha-raft 970, server 839, network 461, grpc-client 150 —
all green.

PR #6784, opened 2026-08-26
fix(#6772): a session that writes before it searches reaches the persisted vector graph too

### LINKED ISSUE #6772: LSM_VECTOR: a session that inserts before it searches never reaches #6655's prefix reuse and pays a full synchronous rebuild on the search thread
Two fixes landed on 2026-08-25 and both do what they claim. The close-time rebuild deferral for #6067 (`d48c01ef1`, PR #6724) takes a 1M-vector index from 135,988 ms to 99.5 ms for open-insert-one-close. The stale-prefix reuse for #6655 (`7ec5797ae`, PR #6712) lets a read-only session reopen a close-deferred database and reuse the persisted graph instead of rebuilding it.

Neither is reachable from a session that INSERTS before it searches, which is the ordinary ingest-then-query shape. That path takes a synchronous full rebuild of the whole index on the search thread.

This is not a regression in either fix. The same session cost 139,248 ms before the deferral and 128,543 ms after it, so the behaviour predates both. What changed is that it is now the whole remaining cost, and that #6655 built the fast path this session cannot get to.

Line numbers against `87695796d`.

## Repro

Both arms start from a byte-identical copy of one database: built, closed, one vector appended, closed again, so the persisted graph is a valid prefix of the live set and the manifest is marked close-deferred. The only difference between the arms is whether the session inserts one vector before searching.

```
### 20,000 vectors, one appended vector, first search after reopen
###   read-only session :   215 ms   stalePrefixGraphReuses=1 graphNodeCount=20,000
###   wrote-first session:  2618 ms  stalePrefixGraphReuses=0 graphNodeCount=20,002
```

The wall clock moves between runs (7.5x, 12.2x, 15.9x across three); the two state counters do not. `stalePrefixGraphReuses` is 1 against 0, and the wrote-first arm ends with a graph over every live vector rather than over the persisted prefix, which is a full rebuild by definition.

## Mechanism

`put()` and `putBatch()` promote `graphState` from LOADING to MUTABLE on commit, before any search runs:

```java
if (graphState == GraphState.IMMUTABLE || graphState == GraphState.LOADING)
  this.graphState = GraphState.MUTABLE;      // :6227-6228 in put(), :6349-6350 in putBatch()
```

`ensureGraphAvailable()` then returns at its first statement:

```java
if (graphState != GraphState.LOADING)
  return; // Graph already available or being built     // :1500-1501
```

so the persisted graph is never loaded, `graphIndex` stays null, and #6655's prefix test at `:1689-1694` is unreachable. Not because any of its own guards fail, but because the method exits before reaching them.

`rebuildGraphBeforeSearch()` then reads that null:

```java
final boolean isSmallGraph = graphIndex == null || graphIndex.size() < ASYNC_REBUILD_MIN_GRAPH_SIZE;  // :3525
if (isSmallGraph) {
  ...
    buildGraphFromScratch();                                                                          // :3533
```

and rebuilds the entire index synchronously on the calling thread. The `threshold` computed just above is consulted only on the `else if`, so the small-graph arm has no mutation gate at all: one insert is enough.

**The flag is carrying two meanings and only one of them makes a null `graphIndex` mean "small".** `rebuildGraphBeforeSearch()`'s own javadoc states the assumption at `:3510`:

> If graph was never built (graphIndex == null): synchronous build (no existing graph to search).

Here a graph WAS built, and it is on disk. `needsGraphBuild()`'s javadoc already draws exactly this distinction at `:7060`:

> LOADING means "not loaded into memory", which is not the same as "not persisted on disk"

That is the distinction the MUTABLE promotion erases. After it, "a loaded graph now has pending deltas" and "no graph has been loaded yet and there are pending deltas" are the same state.

Confirmed by log rather than by reading the call chain: the wrote-first sessions emit no `Loaded graph from disk` line and instead `Building JVector graph index with 20002 vectors`, both at INFO.

## What it costs at scale

Our lifecycle lane, i9-12900HK, cpuset 0-11, embedded through the Python wheel, 64 dimensions, medians of 3 processes each a median of 5 cycles after 2 warmups:

| session, 10,000,000 vectors | before the deferral (`5010b306c9`) | after (`d7940d79e`) |
|---|---:|---:|
| open, insert one vector, close | 2,465,487 ms | **49.8 ms** |
| open, insert one vector, **search**, close | 2,360,720 ms | **2,496,097 ms** |

The first row is #6067 working: 41 minutes to 50 milliseconds. The second is the same session with a search in it, and at 10M it is 41.6 minutes either way.

At 1M the same pair is 135,988 to 99.5 ms and 139,248 to 128,543 ms.

## Why CI is green

`Issue6655StaleGraphPrefixReuseTest` exercises only a session that does not write before searching: its searches follow writes made by a PREVIOUS session, never one in the searching session. So the arm this issue is about is not covered.

## Where the fix could go

The load path is already written to be correct when MUTABLE arrives. `ensureGraphAvailable()` documents that case at `:1714-1718`, keeping the freshly loaded graph when a concurrent `put()` advanced the state mid-validation, because "whatever changed since then is already in deltaVectors", and `reuseStalePrefixGraph()` publishes with `addAll()`/`addAndGet()` for the same reason. So what stands between a wrote-first session and the reuse is the gate, not the machinery behind it. Note there are three sites, not one: the entry gate at `:1500`, the post-lock recheck at `:1529`, and `reuseStalePrefixGraph()`'s own at `:1808`.

The obvious-looking alternative is not sound and I would rather say so than offer it: stop treating a null `graphIndex` as small at `:3525` and this session takes the async arm instead, but with `graphIndex` still null the search falls into the delta-only branch and answers from `deltaVectors` alone, which here is the one vector this session just wrote. The class says so itself at `:2087`. Any fix that leaves `graphIndex` null has to keep the query correct.

Filing as an issue rather than a PR because which of those is right is a design call that is yours, the same way #6583 was. Happy to send a patch in whichever direction you pick, and the repro above is ready to attach as a test.

lvca:
Fixes #6772

## The bug

`put()`/`putBatch()`/`remove()` promote `graphState` from `LOADING` straight to `MUTABLE` on the first write of a session. `ensureGraphAvailable()` returned at its first statement on anything but `LOADING`:

```java
if (graphState != GraphState.LOADING)
  return; // Graph already available or being built
```

So a session that INSERTS before it SEARCHES - the ordinary ingest-then-query shape - never loaded the persisted graph at all. `graphIndex` stayed null, #6655's prefix-reuse test was unreachable (not because any of its guards failed, but because the method exited before reaching them), and `rebuildGraphBeforeSearch()` then read that null as "small graph":

```java
final boolean isSmallGraph = graphIndex == null || graphIndex.size() < ASYNC_REBUILD_MIN_GRAPH_SIZE;
if (isSmallGraph)
  buildGraphFromScratch();   // the WHOLE index, synchronously, on the search thread
```

The reporter measures 10M vectors, one insert, one search: 2,496,097 ms after #6067's deferral against 2,360,720 ms before it. The same session without the search is 49.8 ms. At 20,000 vectors the read-only arm returns in 215 ms with `stalePrefixGraphReuses=1`; the wrote-first arm takes 2,618 ms with `stalePrefixGraphReuses=0` and ends with a graph over every live vector rather than over the persisted prefix.

Not a regression in #6067 or #6655. What changed is that it is now the whole remaining cost, and that #6655 built the fast path this session cannot get to.

## Root cause

The promotion is the bug. After it, "a loaded graph now has pending deltas" and "no graph has been loaded yet and there are pending deltas" are the same state. `needsGraphBuild()`'s javadoc already names exactly this distinction for the close path:

> LOADING means "not loaded into memory", which is not the same as "not persisted on disk"

## The fix

**Make the distinction explicit rather than derived.** A dedicated latch, `persistedGraphUnresolved`, answers "has this instance decided yet what to do with whatever graph is on disk?" - a question `graphState` cannot answer once a write has promoted it. The three gates the issue identifies (`ensureGraphAvailable()`'s entry, its post-lock recheck, and `reuseStalePrefixGraph()`'s own) now consult `graphNotYetMaterialised()`:

```java
return graphState == GraphState.LOADING || (persistedGraphUnresolved && graphIndex == null);
```

The `graphIndex == null` conjunct is the structural half: whatever the latch says, an index with a graph already in memory has nothing left to resolve, so a publish path that forgets to clear the latch cannot make every search reload and re-validate a resident graph. The latch is cleared exactly once, in `ensureGraphAvailable()`'s `graphBuildLock` finally (every exit has settled the question, including a load that threw and fell through) and at the head of `buildGraphFromScratchWithRetry()` (whatever is on disk is about to be superseded). A build that fails outright leaves `graphState` at `LOADING`, which the first disjunct still answers true on, so the retry path is unaffected.

### Why reaching the reuse was only half of it

A session that wrote before it searched arrives with its own writes already in `deltaVectors` AND, because those writes went into the location index too, inside the gap past the persisted prefix. Queuing the gap wholesale - which is exactly what a minimal "make the gate reachable" fix does - would buffer every one of them twice and double-count them in `mutationsSinceSerialize`. `reuseStalePrefixGraph()` now filters the gap against the buffer twice:

- **before the read**, because skipping the read is most of the point: a session that inserted a million vectors before its first query would otherwise pay a page lookup per vector to fetch back what is already in memory beside it;
- **again under the publish lock**, to catch anything a concurrent writer queued while the gap was being read unlocked.

Its double-check moves from `graphState` to `graphIndex` for the same reason the entry gate did - `graphState` is no longer `LOADING` on the path it now also serves - with an `IMMUTABLE` arm for the one publisher that leaves `graphIndex` null (a from-scratch build that found no valid vectors at all).

`IntHashSet` rather than a `HashSet<Integer>` for both filters: primitive, no boxing, sized in table slots so neither rehashes on the way in.

### On the alternative the issue rules out

The issue is right that stopping `:3525` from treating a null `graphIndex` as small is unsound - the search would fall into the delta-only branch and answer from `deltaVectors` alone. This fix does not leave `graphIndex` null; it fills it with the persisted prefix, which is what makes the async arm correct to take.

## Test

`Issue6772WriteBeforeSearchPrefixReuseTest` is `Issue6655StaleGraphPrefixReuseTest`'s fixture (128 dimensions, `maxConnections=64`, `beamWidth=400`, 20,000 vectors, a crashed session leaving 200 more) differing in one thing: session 3 inserts one vector before it searches. It pins, in this order:

1. `stalePrefixGraphReuses == 1` and `graphNodeCount == 20,000` - the two state facts the issue reports as stable across runs while wall clock is not, so they are what names the regression when this goes red;
2. `deltaVectorsCount == 201` and `mutationsSinceRebuild == 201` - each pending vector buffered and counted exactly once, the duplicate check a naive fix fails;
3. the answer: both a vector written by the crashed session and the one written by THIS session, before the search, come back from the very first query;
4. only then the `assertGaveUpWithin(2_000, ...)` tripwire.

Verified failing on `main`'s code for reason (1) - "the reuse path must run for a session that wrote before it searched" - and passing with the fix.

Tagged `@Tag("vector")`, same lane and same reason as the #6655 test.

## Verification

- `Issue6772WriteBeforeSearchPrefixReuseTest`: fails without the main-source change, passes with it
- `com.arcadedb.index.vector.**` with no group exclusions: **429 tests, 0 failures**
- `com.arcadedb.function.sql.vector.**`, `SelectVectorTest`, `CypherCallVectorNeighborsTest`, `DatabaseLifecycleBackgroundThreadsTest`: **323 tests, 0 failures**

## Files changed

- `engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java`
- `engine/src/test/java/com/arcadedb/index/vector/Issue6772WriteBeforeSearchPrefixReuseTest.java` (new)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vector index recovery after writes made before the first search.
  * Reuses persisted graph data instead of triggering unnecessary synchronous rebuilds.
  * Prevents duplicate vectors and preserves concurrent writes during index updates.
  * Ensures newly added vectors remain searchable across reopened sessions.
* **Tests**
  * Added regression coverage for prefix reuse, deduplication, search performance, and eventual asynchronous rebuilding.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -62,6 +62,7 @@ import com.arcadedb.schema.Schema;
 import com.arcadedb.schema.Type;
 import com.arcadedb.serializer.BinaryComparator;
 import com.arcadedb.serializer.json.JSONObject;
+import com.arcadedb.utility.IntHashSet;
 import com.arcadedb.utility.LockManager;
 import com.arcadedb.utility.Pair;
 import com.arcadedb.utility.RidHashSet;
@@ -202,6 +203,20 @@ public class LSMVectorIndex implements Index, IndexInternal {
   }
 
   private volatile GraphState                    graphState;
+  // ISSUE #6772: whether this instance has still to decide what to do with whatever graph is on disk - load it,
+  // reuse it as a prefix, or discard it for a rebuild. NOT derivable from graphState, and that is the whole point:
+  // LOADING means "no graph in memory yet", but put()/putBatch()/remove() promote LOADING straight to MUTABLE on
+  // the first write, and after that promotion "a loaded graph now has pending deltas" and "no graph has been
+  // loaded yet and there are pending deltas" are the same state. ensureGraphAvailable() used to gate on LOADING
+  // alone, so a session that wrote before it searched skipped the load entirely, left graphIndex null, and had
+  // rebuildGraphBeforeSearch() read that null as "small graph" and rebuild the WHOLE index synchronously on the
+  // search thread - the ordinary ingest-then-query shape paying exactly the cost issues #6067 and #6655 removed
+  // for the sessions that happen not to write first. needsGraphBuild()'s javadoc already draws this distinction
+  // for the close path ("LOADING means not loaded into memory, which is not the same as not persisted on disk");
+  // this field is that distinction made explicit for the search path, so the promotion can no longer erase it.
+  // Starts true - a freshly constructed index has resolved nothing - and is cleared exactly once, by whichever of
+  // ensureGraphAvailable() or buildGraphFromScratchWithRetry() first settles the question under graphBuildLock.
+  private volatile boolean                       persistedGraphUnresolved = true;
   private volatile ImmutableGraphIndex           graphIndex;        // Current graph (OnHeap or OnDisk)
   private volatile int[]                         ordinalToVectorId; // Maps graph ordinals to vector IDs
   // Lightweight pointer index. Volatile and swapped as a whole (never cleared and refilled in place) so the
@@ -1493,11 +1508,35 @@ public class LSMVectorIndex implements Index, IndexInternal {
   }
 
   /**
+   * {@code true} while {@link #ensureGraphAvailable()} still has work to do: either no graph has been materialised
+   * in memory for this instance yet ({@link GraphState#LOADING}), or one has not been but a write already promoted
+   * the state past LOADING ({@link #persistedGraphUnresolved} - issue #6772).
+   * <p>
+   * The second disjunct is the one that makes an ingest-then-query session reach the same persisted-graph load and
+   * prefix reuse (issue #6655) a query-only session gets. Without it the promotion {@code put()}/{@code putBatch()}
+   * /{@code remove()} perform on the first write closes the door on the load for the rest of the session, and the
+   * first search pays a full synchronous rebuild through {@link #rebuildGraphBeforeSearch()}'s null-graph arm
+   * instead.
+   * <p>
+   * That disjunct is additionally conditioned on {@link #graphIndex} being null, which is the structural half of
+   * the answer: whatever the latch says, an index that already has a graph in memory has nothing left for
+   * {@link #ensureGraphAvailable()} to do, and this is what keeps a publish path that forgets to clear the latch
+   * (the chunked initial build reaches one through {@link #buildGraphFromScratchWithRetry}, but nothing forces a
+   * future one to) from making every search reload and re-validate a graph that is already resident.
+   */
+  private boolean graphNotYetMaterialised() {
+    return graphState == GraphState.LOADING || (persistedGraphUnresolved && graphIndex == null);
+  }
+
+  /**
    * Ensure graph is available for searching. Lazy-loads from disk if needed.
    * This is the entry point for all search operations.
+   * <p>
+   * Reached on the first search of a session whether or not that session has already written to the index
+   * (issue #6772): see {@link #graphNotYetMaterialised()}.
    */
   private void ensureGraphAvailable() {
-    if (graphState != GraphState.LOADING)
+    if (!graphNotYetMaterialised())
       return; // Graph already available or being built
 
     // Materialise the locations before graphBuildLock rather than through the vectorIndex() calls below, which
@@ -1526,7 +1565,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
     graphBuildLock.lock();
     try {
       // Double-check after acquiring the lock
-      if (graphState != GraphState.LOADING)
+      if (!graphNotYetMaterialised())
         return; // Another thread already resolved this while we waited for graphBuildLock
 
       // Try to load persisted graph from disk
@@ -1756,6 +1795,14 @@ public class LSMVectorIndex implements Index, IndexInternal {
       if (prefixReuseCandidate == null)
         buildGraphFromScratch();
     } finally {
+      // Cleared here rather than in each branch above, and in a finally rather than on the success paths only,
+      // because the latch means "still to be decided", not "decided successfully" (issue #6772). Every exit from
+      // this section - graph loaded, prefix queued for reuse, rebuilt from scratch, or a load that threw and fell
+      // through to the rebuild - has settled the question of what happens to the on-disk graph, and leaving the
+      // latch set on any of them would make the NEXT search redo the O(live vector count) validation scan above.
+      // A build that fails outright still leaves graphState at LOADING, and graphNotYetMaterialised() keeps
+      // answering true on that alone, so the retry this method has always offered is unaffected.
+      persistedGraphUnresolved = false;
       graphBuildLock.unlock();
     }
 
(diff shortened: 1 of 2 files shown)


lvca:
Thanks - the race is real and I've fixed it as suggested, in `658fe8a`.

**Confirmed.** `ensureGraphAvailable()` cleared the latch at *decision* time, but on the prefix-reuse branch the graph is published later, in `reuseStalePrefixGraph()`, after `graphBuildLock` has been released and re-acquired. Between the two: `graphState == MUTABLE` (the session wrote before it searched - the very case this PR exists for), latch `false`, `graphIndex == null`. A second searcher landing there gets `graphNotYetMaterialised() == false`, skips the load entirely and falls into `rebuildGraphBeforeSearch()`'s null-graph arm - the full synchronous rebuild this PR removes, back again on a race instead of on a promotion. The contrast you drew with the query-only path is exactly right: there `graphState` stays `LOADING` across the same window, so a racing thread still re-enters correctly.

**Fix.** Ownership of the latch passes to `reuseStalePrefixGraph()`, which clears it in its own `finally`, under the same `graphBuildLock` that publishes `graphIndex` - so the next thread to take that mutex sees the cleared latch and the published graph together rather than one without the other. `ensureGraphAvailable()` still clears it for every other branch (loaded, rebuilt, load threw and fell through), since those all publish before releasing.

**On the residual window**, between `ensureGraphAvailable()`'s unlock and `reuseStalePrefixGraph()`'s own `lock()`: a thread arriving there still sees the latch set, so the worst it can do is repeat the validation scan, reach the same decision, and then lose the `graphIndex != null` double-check and discard its work. That is precisely what a racing pair already does on #6655's query-only path, and it is bounded by an O(live vector count) scan rather than by a rebuild. I considered closing it completely by calling `reuseStalePrefixGraph()` inside `ensureGraphAvailable()`'s `graphBuildLock` section (it is reentrant, and that method already runs its whole body - I/O included - under the same mutex, so the release/re-acquire buys nothing but the window). I left it alone because it would drag `startAsyncGraphRebuild()`, which is `synchronized(this)`, under `graphBuildLock`, and I'd rather not add a lock-ordering edge to fix a redundant scan.

**New test:** `concurrentSearchersOnAWroteFirstReopenStillPublishExactlyOneReuse` pins what must hold across that window - eight simultaneous searchers each get a correct answer, exactly one reuse is published however many threads reached the decision, and the gap is still buffered and counted exactly once (the dedup has to survive two callers racing for the publish lock, not only one running alone). Sessions 1 and 2 are now a shared fixture helper.

Verification: both tests pass; `com.arcadedb.index.vector.**` with no group exclusions is **430 tests, 0 failures**.

Also noted, since you raised it: `buildGraphFromScratchWithRetry()` is safe for the reason you gave - it holds `graphBuildLock` across build *and* publish, so it has no equivalent window.

PR #6788, opened 2026-08-27
fix(#6315): remove the positional edge-property accessor from the traversal SPI, resolve the overlay instead

### LINKED ISSUE #6315: GraphTraversalProvider.getEdgeProperty still answers from the base CSR while an overlay is active, so the next caller inherits the bug #6301 fixed
Found while fixing #6301 (PR #6306). The instances are closed; the **shape** is not.

## What #6306 did, and what it did not do

`getEdgeProperty(nodeId, neighborIndex, direction, edgeType, property)` addresses an edge **by its position in the node's neighbour list**. That contract holds only while the position identifies the same edge `getNeighborIds` reports there - and an active `DeltaOverlay` breaks it: the property columns are aligned with the base CSR's forward edge slots, while `getNeighborIds` serves the overlay's view of the node (deletions dropped, additions merged in, the whole list re-sorted by `Arrays.sort`). The n-th neighbour of that list is then not the n-th edge of the column store.

PR #6306 closed this for every caller that exists today, by narrowing `hasEdgeProperties()`:

```java
// GraphAnalyticalView
public boolean hasEdgeProperties() {
  return snap != null && snap.edgeColumnStores != null && !snap.edgeColumnStores.isEmpty() && snap.overlay == null;
}
```

**`getEdgeProperty` itself is unchanged.** Called directly with an overlay active it still returns a value addressed against the base CSR - a weight belonging to some other edge - with nothing in its signature or behaviour to say so. Verified on `0237e7b`: the method body contains no reference to `snap.overlay`.

So the guarantee currently rests on every caller remembering to ask `hasEdgeProperties()` (or `servesEdgeProperty`) first. That is exactly the shape #6301 was: a correct-if-you-know-the-rule API, where breaking the rule produces a plausible wrong number rather than an error. The next caller written against the SPI has no way to discover the rule from the method it is calling.

## Why it matters more than the caller count suggests

The failure is silent. A wrong weight produces a wrong shortest path, a wrong MST, a wrong Steiner tree - never an exception, never a log line. #6301 sat undetected long enough to reach four independent call sites, and the two in `SQLFunction*` were only found because a reviewer went looking on the third pass.

It is also reachable in ordinary operation, not a corner: an overlay is what `UpdateMode.SYNCHRONOUS` produces on **every commit** against a built view.

## Options

1. **Return `null` when the overlay is active.** Honest, and matches `getMeanEdgesPerConnectedPair`, which already answers `MULTIPLICITY_UNKNOWN` for exactly this overlay. The catch: `null` is *also* how "this edge has no value for that property" is reported, so a caller still cannot distinguish "unavailable" from "absent" - it would silently default the weight instead of silently reading the wrong one. Better, but not closed.
2. **Make the overlay resolvable.** `getNeighborIds` knows which entries came from the base CSR and at what offset; carrying that through would let `getEdgeProperty` map a post-overlay position back to its forward slot, and answer exactly for base edges and `null` only for overlay-added ones. This closes the shape rather than the instances, and is the only option that lets a weighted algorithm keep the columnar path while a view is being updated - which is the case the CSR exists for.
3. **Remove the positional form from the SPI** and leave `edgeWeightsOf` (added in #6306) as the only way in. It already handles the multi-type merge and the `BOTH` split, and it is the only correct way to use `getEdgeProperty` today. Making the sharp tool private to the SPI's own default method is what stops a fifth hand-rolled copy from appearing.

Option 3 is the smallest change that closes the shape, and 2 is the one worth doing if the columnar path should survive an overlay at all. They compose: 3 now, 2 when someone needs the performance.

## Related

#6301, PR #6306, #6296, #6289.

lvca:
Closes #6315.

## The shape

`GraphTraversalProvider.getEdgeProperty(nodeId, neighborIndex, direction, edgeType, property)` addressed an
edge **by its position in the node's neighbour list**. That position identified the same edge
`getNeighborIds` reported only while nothing else was in play: an active `DeltaOverlay` drops deleted edges,
merges added ones in and re-sorts the list, while the property columns stay aligned with the base CSR's
forward slots.

Reproduced before the change on a three-edge star, one edge deleted in a committed transaction:

```
BEFORE  neighbors=[B, C, D]   weights=[10.0, 20.0, 30.0]
AFTER   neighbors=[C, D]      weights=[10.0, 20.0]      <- C answers with B's weight
```

A wrong shortest path, a wrong MST, a wrong Steiner tree - and never an exception. PR #6306 closed this for
the callers that existed by narrowing `hasEdgeProperties()` to `false` whenever an overlay was active, which
left the accessor in the SPI for the next caller to find and the rule for it to remember.

## What this does

The issue offered three options and suggested "3 now, 2 when someone needs the performance". This does both,
because they are not independent: option 3 alone removes the sharp accessor but leaves a weighted algorithm
reading edge records on every commit of a `SYNCHRONOUS` view, which is the case the CSR exists for.

**The accessor is gone.** In its place `edgeWeightsForSlice()` returns one `(type, direction)` slice's
neighbours and their weights together, produced by the walk that decided which edges the slice holds - no
index crosses the SPI boundary, so there is nothing left to re-derive wrongly. `edgeWeightsOf()` composes the
slices, which is where the multi-type merge and the `BOTH` split already lived, and a slice the provider
cannot answer for makes the whole node unanswerable rather than a neighbourhood with edges silently missing
from it. The per-edge `WorkGuard` checkpoint from #6715 is handed down into the slice build, so a supernode
stays as abortable as it was.

**The overlay is resolved rather than refused.** In `GraphAnalyticalView`:

- a base edge the overlay has not deleted takes its own forward column slot, tracked through the merge by
  packing each entry's provenance into the low half of the sort key, so the re-sort cannot separate a weight
  from its edge;
- an edge the overlay added takes the value `DeltaCollector` captured for it at commit time - it has no column
  slot, the columns having been built with the base CSR;
- a run of parallel edges between one pair, **partially** deleted, is refused. Deletions are counted per pair,
  which is all the neighbour list needs - drop any one of the run and the remaining neighbours are the same -
  but not enough to say which of them died, and parallel edges need not weigh the same. Deleting the whole run
  is not ambiguous, and neither is the ordinary pair joined by a single edge;
- an edge whose properties were **updated** marks the columns out of date until the rebuild that repairs them
  lands, so no stale weight is served in exchange for the added edges now being served exactly. An update to
  an edge the overlay itself holds is applied there and dirties nothing - which is what keeps the ordinary
  `newEdge(...).save()`, one create and one update of the same edge, from forcing a full rebuild per insert
  (it did before). Past 1024 tracked updates in one transaction the delta stops recording them individually
  and just says the columns are out of date, so a bulk rewrite costs no more memory than the old boolean did.

**Two consequences of the view now serving edge properties through an overlay**, both of which used to be
handled by accident:

- `GraphAlgorithms.dijkstraSingleSource` reads the CSR offset and neighbour arrays directly, and was kept off
  them only as a side effect of the #6306 narrowing - a gate it never asked for and could not see. It refuses
  for itself now, and `algo.dijkstra.singleSource` falls back to the edge records;
- `AbstractAlgoProcedure`'s columnar build reads the records for a node the provider cannot answer for, using
  the same row reader its whole-graph record build now uses, so the fallback cannot answer differently from
  the path it stands in for.

## Tests

`Issue6315EdgeWeightsUnderOverlayTest` (11 cases): the deletion shift in both directions, an added edge
answering with its own weight, an update to an overlay-held edge staying columnar, an updated base edge
refusing rather than serving stale, the ambiguous parallel-edge run refused, the unambiguous one still
answered, `edgeWeightsForSlice`'s neighbours agreeing with `getNeighborIds`, the `BOTH` split, the Dijkstra
kernel not answering from the stale base CSR, and an `algo.bellmanford` run over a graph with one unanswerable
node. The last two were checked to fail without their fix rather than pass vacuously.

`Issue6301AlgoSteinerTreeWeightAlignmentTest.anActiveOverlayDoesNotLendItsWeightsToTheWrongEdges` keeps its
weight assertion and now pins the columnar answer instead of the fallback.

Engine suite: 13,634 tests, 0 failures (`-DexcludedGroups=benchmark,slow,vector`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved weighted graph traversal across multiple edge types and directions.
  * Preserved edge-weight alignment when analytical views contain additions, deletions, updates, or parallel edges.
  * Ensured edge-property changes are reflected during overlays and rebuilds.
  * Added safer fallbacks for shortest-path queries when accelerated graph data is unavailable or changing.
  * Improved combined incoming/outgoing traversal behavior.

* **Refactor**
  * Updated weighted-edge processing to keep neighbors and weights correctly paired throughout graph algorithms.

## For the release notes: a source-breaking SPI change

`GraphTraversalProvider.getEdgeProperty(nodeId, neighborIndex, direction, edgeType, propertyName)` is
**removed**, not deprecated, and `edgeWeightsForSlice(nodeId, direction, edgeType, propertyName, defaultWeight,
edgeCheckpoint)` takes its place. Nothing in the tree implements the SPI besides `GraphAnalyticalView` and two
test doubles, so there is no in-tree breakage - but the interface is a published extension point, and an
out-of-tree provider that implements the old method will no longer compile.

Deprecating it instead was considered and rejected: the removed method *is* the bug. Leaving it in place, even
deprecated, leaves a way to ask for a weight by position that answers with a plausible wrong number whenever an
overlay is active, which is precisely the shape this issue asks to close. A provider that fails to compile
finds out; one that keeps compiling against the old accessor does not.

## Follow-ups filed

- #6791 - `algo.dijkstra.singleSource` is the one weighted procedure that still abandons the CSR on every
  commit: its kernel reads the arrays directly, so it refuses under an overlay rather than resolving against it.
- #6792 - a view's dense node ids can exceed `getNodeCount()` once a base vertex is deleted, so `algo.*`
  silently drops those nodes. Pre-existing and independent of this PR, verified by reproducing it down the
  record path that predates it.

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionAstar.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionAstar.java
@@ -268,10 +268,10 @@ public class SQLFunctionAstar extends SQLFunctionHeuristicPathFinderAbstract {
     if (provider != null) {
       final int nodeId = provider.getNodeId(node.getIdentity());
       // edgeWeightsOf() answers null unless the provider can serve THIS property for EVERY type in play, and it
-      // is the only thing that pairs a CSR edge with its own weight correctly. Reading getNeighborIds and
-      // getEdgeProperty side by side here got it wrong twice over (issue #6301): the neighbour list is merged and
-      // sorted across types while the property column is per type, and a BOTH lookup has no column at all, so it
-      // answered null for every edge and quietly priced the whole neighbourhood at MIN - free.
+      // is the only thing that pairs a CSR edge with its own weight correctly. Assembling the neighbourhood here
+      // got it wrong twice over (issue #6301): the neighbour list is merged and sorted across types while the
+      // values are served per type, and a BOTH lookup has no adjacency slice at all, so it found no value for
+      // any edge and quietly priced the whole neighbourhood at MIN - free.
       final NodeEdgeWeights edges = nodeId >= 0 ?
           provider.edgeWeightsOf(nodeId, paramDirection, paramWeightFieldName, MIN, paramEdgeTypeNames) : null;
       if (edges != null) {
--- a/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionBellmanFord.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionBellmanFord.java
@@ -111,11 +111,11 @@ public class SQLFunctionBellmanFord extends SQLFunctionMathAbstract {
 
     final GraphTraversalProvider provider = GraphTraversalProviderRegistry.findProvider(db);
     // The columnar path is taken per node and only when edgeWeightsOf() can answer - i.e. when the provider can
-    // serve THIS property for every type it holds. Pairing getNeighborIds with getEdgeProperty by hand was wrong
-    // in two ways (issue #6301): the neighbour list is merged and sorted across types while the columns are per
-    // type, and `direction` defaults to BOTH here, which has no column at all - so the plain three-argument call
-    // read a null property value for every edge and silently unit-weighted the whole graph. Anything the
-    // provider cannot answer exactly falls to the edge records below, which are.
+    // serve THIS property for every type it holds. Assembling the neighbourhood by hand was wrong in two ways
+    // (issue #6301): the neighbour list is merged and sorted across types while the values are served per type,
+    // and `direction` defaults to BOTH here, which has no adjacency slice at all - so the hand-rolled version
+    // read no property value for any edge and silently unit-weighted the whole graph. Anything the provider
+    // cannot answer exactly falls to the edge records below, which are.
     final String weightColumn = weightProperty != null && !weightProperty.isEmpty() ? weightProperty : null;
 
     for (int i = 0; i < n; i++) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -161,8 +161,8 @@ public interface GraphTraversalProvider {
   /**
    * Returns the edge types this provider actually holds, or {@code null} when it cannot enumerate them.
    * <p>
-   * {@link #getEdgeProperty} is addressed per edge type, so a caller that was given no type filter but needs
-   * edge properties has to ask which types exist rather than pass "all types" down. A provider that answers
+   * {@link #edgeWeightsForSlice} is addressed per edge type, so a caller that was given no type filter but
+   * needs edge properties has to ask which types exist rather than pass "all types" down. A provider that answers
    * {@code null} simply cannot serve such a caller, which then falls back to reading the edge records.
    */
   default String[] getMaterializedEdgeTypes() {
--- a/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphTraversalProvider.java
@@ -170,17 +170,14 @@ public interface GraphTraversalProvider {
   }
 
   /**
-   * Returns true if this provider has edge property columns materialized <em>and</em> the positional mapping
-   * {@link #getEdgeProperty} relies on is exact.
+   * Returns true if this provider has edge property values it can serve at all, through
+   * {@link #edgeWeightsOf}.
    * <p>
-   * The second half is what makes the answer usable. {@code getEdgeProperty} addresses an edge by its position
-   * in the node's neighbour list for a direction, so it means something only while that position identifies the
-   * same edge {@link #getNeighborIds} reports there. A provider whose neighbour lists no longer line up with its
-   * property columns - because pending changes are served from a side structure, say - must answer {@code false}
-   * rather than hand back a weight belonging to another edge. The caller holds the edge records and can always
-   * read the property itself; a plausible wrong number is the one outcome it cannot recover from. Same
-   * "say unknown rather than guess" convention as {@link #countEdgesBetween} and
-   * {@link #getMeanEdgesPerConnectedPair}.
+   * A provider that cannot pair every edge it reports in {@link #getNeighborIds} with that edge's own property
+   * value - because its columns have gone out of step with its adjacency, say - must answer {@code false} rather
+   * than hand back a value belonging to another edge. The caller holds the edge records and can always read the
+   * property itself; a plausible wrong number is the one outcome it cannot recover from. Same "say unknown rather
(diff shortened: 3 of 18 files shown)


lvca:
Follow-up for the one caller that does not get the benefit of this change - `algo.dijkstra.singleSource`, whose CSR kernel reads the arrays directly and so refuses under an overlay rather than resolving against it - is now tracked as #6791. Its refusal is not a regression (the same fallback happened before, by a different route), but it is easy to miss that this one procedure was left behind, which is what the review flagged.

lvca:
On the two CodeRabbit findings:

**Overlay neighbour ids outside the weighted graph's bounds** - real, and filed as #6792 with a repro. It is
not introduced here, which I checked rather than assumed: running the same scenario with the view
materialising no edge property columns sends the procedure down `weightedAdjacencyFromRecords`, the path that
existed before this PR, and it produces byte-identical ids and the same answer. `getNeighborIds` has always
served overlay ids and `getNodeId` has always resolved them, so every path into `GraphData` could already
carry one. The observable is a live vertex whose id landed above `getNodeCount()` being silently skipped by
every `algo.*` procedure; the out-of-bounds index is reachable from the same cause. Left out of this PR
because fixing it means moving either the id space or the meaning of `getNodeCount()`, neither of which
belongs in a fix for the edge-property alignment.

**Deriving the follow-up rebuild from the re-merged overlay rather than from `d.updatedEdges`** - agreed, and
already done in 847f8a2 for exactly the reason given: an ordinary `newEdge(...).save()` reports one create and
one update of the same edge, and the overlay resolves that against its own additions, so asking the deltas
forced a rebuild per buffered insert.

lvca:
### CI triage

Two lanes are red, and both are red on unmodified `main` at `9c9d024db` - the exact commit this branch is
based on (run 33016715946):

- **`unit-tests`**: `GraphAnalyticalViewTest.issue6641_corruptedPersistedCsrInvalidatesDeferredRestoreWithoutBlockingQuery`,
  failing on the same assertion (`isReady()` must still be false before the background deferred restore
  finishes) in main's own run. This PR's run adds the sibling `issue6641_interveningCommit...`, which is how
  this family behaves - a different sibling trips run to run. Both pass 3/3 in isolation on this branch.
  Mechanically they cannot be this PR's: they build with `UpdateMode.OFF`, so there is no `DeltaCollector` and
  no overlay anywhere in them, and they touch no method this PR changed - the shared surface is the persisted
  CSR restore path, which is untouched here.
- **`vector-unit-tests`**: `Issue6772WriteBeforeSearchPrefixReuseTest` and `Issue6655StaleGraphPrefixReuseTest`,
  red on `main` since `658fe8a89`. This PR's diff has zero files under `index/vector`.

Everything else is green, including `integration-tests`, `opencypher-tck-tests`, `builder-tests`, all five e2e
lanes and `studio-e2e-tests`. Locally the engine suite is green at 13,642 tests
(`-DexcludedGroups=benchmark,slow,vector`).

Issue #6343, opened 2026-08-18
HA IT budget follow-ups from #6297: the next cut of RESYNC_RETRY_TIMEOUT_MS, and isolating the heavy HA ITs into their own fork (reuseForks)


lvca:
The two items #6297 carried that #6307 deliberately did not implement, refiled so they have a tracker of their own: `Closes #6297` on that merge closed the issue with these still open, and #6323 item 1 found nothing searchable left behind. Both are about the same budget under pressure, and both need CI evidence that a PR cannot produce on its own, which is why neither was decided at the time.

## 1. The next cut of `RESYNC_RETRY_TIMEOUT_MS`

#6284 took it from 120 s to 30 s on the evidence of nine full `ha-integration-tests` runs in which no wait exceeded the 10 s report threshold, and lowered `SLOW_WAIT_REPORT_MS` to 5 s so the next decision would have resolution. Whether 30 s can come down further needs a handful of runs at the new threshold: there was no number to choose without them.

What to do: collect the `SLOW WAIT` lines from `ha-integration-tests` runs on `main` since #6284 merged. If none of them exceeds 5 s, the budget can come down to something in the 10-15 s range; if some do, the largest observed wait is the number to size against, times a margin.

Both constants live in `ha-raft/src/test/java/com/arcadedb/server/ha/raft/BaseRaftHATest.java` (`RESYNC_RETRY_TIMEOUT_MS` at line 73, `SLOW_WAIT_REPORT_MS` at line 80), and their comments now point here.

## 2. Isolating the heavy HA ITs into their own fork

The stalls that motivated the original 120 s budget appeared only in full-suite adjacency, immediately after `RaftMigrationCompactionRaceIT` in the same reused JVM: the root `pom.xml` runs failsafe with `reuseForks=true` and `forkCount=1`, so every IT in `ha-raft` shares one JVM with everything that ran before it. #6202 removed the suspected cause (a download running on the JDK common ForkJoinPool) and #6284 cut the budget, but nothing has been done about the adjacency itself.

What to do: measure before changing anything. `reuseForks=false` for the `ha-raft` module gives every IT class a fresh JVM and would settle the question, at the cost of a JVM start per class - which this lane may or may not be able to afford, since it is the second-longest in the workflow. A narrower alternative is a `runOrder` that separates the heavy classes, or moving only `RaftMigrationCompactionRaceIT` into a fork of its own.

[...]

Issue #6323, opened 2026-08-17
Follow-ups from #6307: two of #6297's items lost their tracker, IN walks a range that answers in O(1), and EXPLAIN cannot describe the plan it would run


lvca:
Three follow-ups from #6307 (the #6297 / #6280 / #6279 / #6270 batch), filed together because that is where all three were found. Item 1 is a tracking regression created by that merge and is the reason this issue exists at all; items 2 and 3 are independent defects in the openCypher engine that the work happened to walk into.

## 1. Two of #6297's own items lost their tracker when #6297 was closed

#6297 carried four items. #6307 implemented items 1 and 2 and deliberately left items 3 and 4 alone, saying so in the PR description - but the commit's `Closes #6297` trailer closed the issue anyway, and with it the only searchable record of the two that were not done:

- **The next cut of `RESYNC_RETRY_TIMEOUT_MS`.** #6284 took it from 120 s to 30 s on the evidence of nine full `ha-integration-tests` runs in which no wait exceeded the 10 s report threshold, and lowered `SLOW_WAIT_REPORT_MS` to 5 s to keep resolution for the next decision. Whether 30 s can come down further needs a handful of runs at the new threshold; there was no number to choose without them, which is why #6307 did not choose one.
- **Isolating the heavy HA ITs into their own fork.** The stalls that motivated the original 120 s budget appeared only in full-suite adjacency, immediately after `RaftMigrationCompactionRaceIT` in the same reused JVM (failsafe `reuseForks=true` / `forkCount=1` for the whole module). #6202 removed the suspected cause and #6284 cut the budget, but nothing has been done about the adjacency itself.

Both now survive only as a comment in `BaseRaftHATest` (`RESYNC_RETRY_TIMEOUT_MS` at line 73, `SLOW_WAIT_REPORT_MS` at line 80) - which is exactly the condition #6297 was filed to escape, in its own words: "recorded in `BaseRaftHATest`'s comment; this is just the tracking entry so it does not depend on someone rereading that file."

A search of open issues for `RESYNC_RETRY_TIMEOUT_MS`, `reuseForks` and the fork-isolation wording returns nothing, so this issue is now their tracker.

### Worth folding in: two CI lanes are sized at the 60-minute job cap

Observed across the #6307 runs, and relevant to the fork-isolation item above because it is the same budget under pressure:

[...]

PR #6789, opened 2026-08-27
fix(#6343): the HA resync budget comes down to 15s on forty runs of evidence, and the SLOW WAIT lines finally exist to prove it

lvca:
Addresses #6343 item 1. **Item 2 (fork isolation) is deliberately not here** - see the bottom of this description and #6848.

## The measurement

Forty full `ha-integration-tests` runs on `main` between 2026-08-18 and 2026-08-26 - every one after #6284 merged - 238 tests and all 123 IT classes in each, sampled every fifth run of the 200 the API still holds logs for:

| | |
|---|---|
| Waits above the 5s report threshold | **0**, in all forty runs |
| `Issue5410AbandonedTicketReleaseIT` (whose stalls bought the original 120s) | 13.2 - 17.0s **wall clock for the whole class** |

And since those only prove no wait *crossed* a threshold, a later experiment measured the waits themselves, running the lane with the report threshold at 1 ms so every wait printed its duration:

```
Issue5410AbandonedTicketReleaseIT  awaitValue(2)  satisfied after 252 ms, then 1 ms
RaftCommandReadConsistencyIT       withResyncRetry satisfied after ~3 000 ms
```

**The budget is sixty times the wait it exists for.** That, rather than an absence of log lines, is the number worth trusting.

## 1. `RESYNC_RETRY_TIMEOUT_MS`: 30s → 15s

15s applies the rule the 30s cut used at the instrument's current resolution: three times the largest wait the evidence can prove any run needed. `SLOW_WAIT_REPORT_MS` follows to 2.5s, holding the threshold at a sixth of the budget - the ratio, not the absolute, is what bounds how much of a budget a wait can burn in silence.

**The ratchet stops here.** Below ~10s the budget stops being a multiple of a measured wait and becomes the wait itself, which converts a busy runner into a red lane. From here the instrument catches regressions rather than proposing cuts.

## 2. The instrument had no name to grep for

Worth its own heading, because it nearly cost this issue a false reading. #6343 asked for "the `SLOW WAIT` lines" from `main`. There were none in forty runs - **not because no wait was slow, but because no line the instrument has ever emitted contained the string `SLOW WAIT`.** It was a formatted sentence with no fixed token. The grep returned zero for a working report and would have returned zero for a deleted one.

The report now carries `TEST SLOW WAIT:`, and `SlowWaitInstrumentTest` (7 tests) pins the marker, the wording, the `WARNING` level and the threshold-to-budget ratio - through the real logging path, not the formatter's return value, and finding the threshold by bisection rather than hardcoding it.

`CapturingTestLogger` also gains level capture (so the tests can assert `WARNING` and the absence of `INFO`), one `Entry` record per call instead of parallel lists that could pair one thread's level with another's text, and `uninstall()` restoring the logger it displaced rather than hardcoding a fresh `DefaultLogger` - the contract `LogManager.setLogger` documents, and a latent fix for the helper's five existing callers.

## Why item 2 is not here

The fork split was in this PR and made the lane red three times out of three. Two controlled runs settled it - same commit, same instrument, same **120-second** budget, differing only in fork arrangement:

| | `awaitValue(2)` | result |
|---|---|---|
| one shared fork (`main`'s arrangement) | 252 ms, then 1 ms | passes |
| heavy ITs in their own fork | **GAVE UP after 120 022 ms** | fails |

Not a timing sensitivity, and not the budget: the vertex never arrives at all. The split does not break `Issue5410AbandonedTicketReleaseIT` - it **exposes** it. That test passes only when `RaftHAComprehensiveIT` (~344s of three-server traffic) has run before it in the same JVM, and behind that is an entry that reaches `commitIndex=8` and never becomes visible.

That is **#6848**, and it is product work. The lesson folded back into the code: the constant's comment now warns that none of this measurement transfers to a different fork arrangement without being re-measured - which is precisely the mistake this PR made by treating #6343's two items as independent.

## Verification

- 977 unit tests green
- `SlowWaitInstrumentTest` is mutation-checked: reverting the marker, and setting the threshold off-ratio in either direction, each make it red
- the marker was confirmed in real CI output on three separate lane runs


lvca:
Thanks - acted on the `failIfNoTests` suggestion, and it turned up something worth recording: **the parameter does not reach the case it was suggested for.**

surefire evaluates `failIfNoTests` against the *class scan*, before the provider applies the tag filter. Measured both halves against failsafe 3.5.6 on this branch:

| heavy execution configured with | result |
|---|---|
| `<includes>` matching no file | `No tests were executed!` - **build fails** |
| `<groups>ha-heavy-typo</groups>`, i.e. tag matching none of the classes the scan found | `Tests run: 0` - **build passes** |

So the likelier regression - the tag being dropped, renamed or mistyped - is invisible to it.

It is set anyway, since it does guard the scan-level version (someone narrows the includes, or the ITs move) and costs nothing; the comment now states which half it covers instead of implying both. A narrowed local run with `-Dit.test` is unaffected, as that falls under `failIfNoSpecifiedTests`.

The guard for the other half is a new `HeavyItForkIsolationTest`, which is finer than any build setting could be. It asserts **set equality in both directions** over the classes actually compiled into the package:

- a class that *loses* the tag rejoins the shared fork silently - caught;
- four classes keeping the tag while the fifth quietly does not - caught, which `failIfNoTests` could never see even if it did run after the filter;
- a class that *gains* the tag - also caught, because that is an extra JVM start on a lane with ~20 minutes of headroom and should be a decision, not a drive-by.

The expected side is written as class literals rather than name strings, so deleting or renaming one of these ITs is a compile error naming the class, not a stale matcher here.

Verified: 9 tests across the two guard classes pass; the drift test goes red when the tag is removed from any single class; `-Dit.test`-narrowed local runs still succeed.

Still outstanding on this PR: the throwaway `experiment/6343-reuseforks-false` branch is running module-wide `reuseForks=false` through the real lane ([run 33045220743](https://github.com/ArcadeData/arcadedb/actions/runs/33045220743)) so the rejected option gets a measured number rather than an estimate. That number goes into the `ha-raft/pom.xml` comment when it lands.

lvca:
Both minor points taken - and the second one turned out to be actionable rather than just documentable.

### "the pom wiring stopped working" now fails loudly

You're right that this is the gap `HeavyItForkIsolationTest` cannot cover, and it is a nasty one precisely because nothing looks wrong: two executions still declared, five classes still tagged, the drift test still green, lane time unchanged - and the heavy classes quietly back in the shared fork.

Rather than leave that to a comment, each execution now stamps its own name into `arcadedb.it.fork`, and a new `HeavyForkWiringIT` - itself `@Tag("ha-heavy")`, doing nothing but reading two system properties - asserts it reads back the heavy one. Verified by breaking the split on purpose (clearing `<excludedGroups>` on the `default` execution):

```
[ERROR] HeavyForkWiringIT.aTaggedItRunsInTheHeavyFork:71
        This IT carries @Tag("ha-heavy") but ran in the "default" fork.
```

Its second test pins something quieter that the first silently depends on: an execution-level `<systemPropertyVariables>` **merges** with the parent's rather than replacing it (asserted via `polyglot.engine.WarnInterpreterOnly` from the root POM). If that ever stopped being true, the heavy fork would lose every setting the parent supplies there.

Cost is one extra fork start doing nothing, on a lane with ~20 minutes of headroom - cheap for turning obscure Maven inheritance into something the lane checks every run.

The pom comment also gained the four-step re-verification recipe you asked for, each step one command against a handful of classes rather than the full hour: partition, per-class JVM, failure attribution, and the two guards.

### `carriesHeavyTag` failed open

Agreed, and worth changing rather than just flagging: for a test whose entire purpose is catching a silently missing tag, "tag genuinely absent" and "class could not be read" reported the same way round, and the second was the one that looked like success. It now throws and names the class (widened to `LinkageError`). The scan only ever sees `.class` files this module just compiled into its own package, so a load failure there is worth stopping for.

Still pending: the `experiment/6343-reuseforks-false` lane ([run 33045220743](https://github.com/ArcadeData/arcadedb/actions/runs/33045220743)) is still in flight. Its number goes into the pom comment when it lands, and the branch gets deleted.

lvca:
> **Keep `ha-heavy` independent from CI tag filters.** When a Maven invocation supplies `-Dgroups` or `-DexcludedGroups`, Failsafe applies those user properties to both executions.

Half right, and the half that is wrong is the half that would have made me dismiss it - so: tested rather than argued, and the finding is **valid**.

**Where the stated mechanism doesn't hold:** explicit execution configuration wins over a user property. `<groups>ha-heavy</groups>` on the heavy execution cannot be narrowed by `-Dgroups`; `<excludedGroups>ha-heavy</excludedGroups>` on the default execution cannot be narrowed by `-DexcludedGroups`. Verified directly - `-Dgroups=bogus-tag -DexcludedGroups=bogus-tag` left the heavy execution running its class exactly as before.

**Where the hole actually was:** in the parameters each execution had left *unset*, which resolve through the mojo's own property defaults. Measured on 3.5.6:

| hostile property | effect before | why |
|---|---|---|
| `-Dgroups=bogus-tag` | default execution ran **0 tests, BUILD SUCCESS** | it never set `<groups>`, so `${groups}` applied and selected nothing - the ~118 ITs it owns silently vanished from the run |
| `-Dfailsafe.excludedGroups=ha-heavy` | heavy execution would exclude its own five classes | it never set `<excludedGroups>`, so the parent's property applied |

The second is the worse one: those classes are excluded from the default execution too, so they would run **nowhere at all** - and `failIfNoTests` cannot see it, since it is evaluated against the class scan before the tag filter (measured earlier in this PR).

**Fix.** Both are now written out explicitly, which is the only form a user property cannot narrow:

- default execution: `<groups>any() | none()</groups>` - "any tag, or no tag", i.e. everything. A tautology on purpose: unset and "set to everything" are not the same thing here.
- heavy execution: `<excludedGroups>any() &amp; none()</excludedGroups>` - a contradiction, so it excludes nothing.

Re-running the mixed selection with all three hostile properties at once now leaves the partition exactly as it was. A plain run is unchanged: an untagged IT and a `@Tag("slow")`-only IT both stay in the default fork, the tagged ones in theirs. The pom's re-verification recipe gained a step 5 so this is one command for the next person.

On the other half of your suggestion - preserving `${failsafe.excludedGroups}` on the default execution - I've deliberately not done that, and the table above is why: composing that property in is exactly what left the heavy side overridable. It is empty repo-wide and exists (per #5697) only to stop surefire's `-DexcludedGroups` leaking into failsafe, which hardcoding preserves rather than breaks. Composing an empty property into a JUnit 5 tag expression would also produce `ha-heavy | ` and fail to parse.

lvca:
> **Require the documented exact threshold ratio.** The `isLessThanOrEqualTo` assertion passes when the threshold is below one sixth of the budget, for example 1 ms.

Valid - the test was weaker than the documentation it exists to guard, which is a poor look for a test whose whole job is stopping documentation and code from drifting apart. Now an equality.

The reasoning for both directions, since only one of them was written down before:

- **too high** - a threshold left at 5s while the budget came down to 15s lets a wait burn a third of it in silence, the blindness that let 120s stand unexamined for years;
- **too low** - at 1 ms every wait reports, the log fills with lines that mean nothing, and the next reader can no more pick out the slow waits than if there had been none.

Verified in the direction the old assertion missed: setting the threshold to 1000 ms now fails with `the report threshold is 1000 ms of a 15000 ms budget, which is not the one-6-th that BaseRaftHATest documents...`.

### And it caught an overstatement of mine in the same place

Worth flagging rather than quietly fixing, because this PR argues from evidence and this bit was not. The constant comment claimed a line in a green run *"means a wait got materially slower, which is a finding on its own"*. At 2.5s that is simply false, and the counter-example was already in the PR description: a local run reported `withResyncRetry on server 1 satisfied after 3013 ms` - an entirely ordinary wait that the old 5s threshold could not see. Lines in the 2.5-5s band are expected.

The comment and the report text now say what a line actually is: **evidence about that wait, not an alarm.** What would be a finding is the number climbing toward the budget, or a `GAVE UP` appearing at all - which is exactly why the line prints elapsed and budget together.

lvca:
Both points taken - each was pointing at a convention this module already had and I had walked past.

### 1. Reuse `CapturingTestLogger` rather than a near-duplicate

Right, and the reason I wrote a second one was the gap you identified: the shared helper records no level, and the test needs to know the report comes out at `WARNING`. Extending it was the better trade. `formattedAt(Level)` added, duplicate deleted.

It also bought an assertion the duplicate did not have: `formattedAt(INFO)` must be empty, so a report **demoted out of WARNING** now fails even though it would still contain every word the other assertions look for. Nobody scrolls an HA IT log looking for INFO lines.

One thing I did not do while adding the level: a third parallel list. This helper captures from every thread of an in-process three-node cluster, and three separate `add`s cannot be made to interleave safely - a reader eventually pairs one call's level with another call's text, which is an assertion passing *about the wrong line*, the precise failure a capturing logger exists to rule out. It is now one `Entry` record per call, which removes that hazard for the two existing accessors as well.

### 2. Restore the displaced logger

Agreed, and it matters more here than the "no demonstrated bug" framing suggests: `forkCount=1` means this module's unit tests share a JVM, so a helper carrying its own idea of "the normal logger" overwrites whatever a surrounding test installed. Now saves `getLogger()` and restores it in `uninstall()`, which makes nesting work by construction - and fixes the same latent problem for the five existing users of the helper, which was the pre-existing half you noted.

Verified: 979 unit tests green, plus the three ITs that use the helper (`RaftLeaderFloodSuppressionIT`, `RaftFollowerCatchupLoggingIT`, `Issue4743CompactionEntrySizeIT`).

### 3. Coverage after the first real CI run

Noted, and a good catch given #5701. Both executions inherit the same plugin-level `@{itCoverageAgent}` argLine and the same destfile, and JaCoCo appends by default, so all 123 classes should land in one `jacoco.xml` - but I'll check the `ha-integration-coverage` artifact on this PR's run rather than assume it, and say so here either way.

lvca:
### 1. `-Dfailsafe.excludedGroups` in `ha-raft` - fixed, not just confirmed

You're right that the PR traded this away silently, and you're right that it's the kind of thing that stays invisible until the day somebody reaches for it. But it turns out the two goals were never in tension, because they are **different knobs**:

- `-Dgroups` / `-DexcludedGroups` **are the parameters**. The only defence against them narrowing an execution is to configure those parameters explicitly - which is what the earlier commit did, and which stays.
- `${failsafe.excludedGroups}` is a *property the module chooses to read*. So it can simply be composed back in.

```xml
<excludedGroups>ha-heavy | (${failsafe.excludedGroups})</excludedGroups>
```

with the module defaulting the property to `any() &amp; none()` - a contradiction, so the expression collapses to plain `ha-heavy` when nobody asks for anything. It has to be a contradiction rather than the parent's empty string, because `ha-heavy | ` does not parse; that was the original reason for hardcoding, and it turned out to have a clean answer.

Measured, all three cases, same mixed selection:

| invocation | default fork | heavy fork |
|---|---|---|
| plain | `Issue5443SplitCompactionEntryIT` (`@Tag("slow")`), `RaftVerifyDatabaseIT` | `HeavyForkWiringIT` |
| `-Dfailsafe.excludedGroups=slow` | `RaftVerifyDatabaseIT` only - **the slow-tagged IT is skipped again** | `HeavyForkWiringIT` |
| `-Dgroups=bogus -DexcludedGroups=bogus` | both, unchanged | `HeavyForkWiringIT` |

So the knob works in this module exactly as in every other one, and the split is still immune to the two that could break it. Both facts are now in the pom next to each other, since the distinction is the non-obvious part.

### 2. The `reuseForks=false` number

Still pending - the lane is genuinely slow (that branch is on its second hour of a job whose baseline p50 is 62 min, which is itself becoming the answer). It lands in the pom comment and on #6343 before this merges.

### 3. Watching for `GAVE UP` lines after the cut

Agreed, and that is exactly what the instrument is now for - the commit that lowered the threshold says so in as many words: from here it catches regressions rather than proposing cuts. Two things make that watchable in a way it wasn't before: the report carries a fixed `TEST SLOW WAIT:` token (it never did), so a sweep of the lane is one grep; and at a threshold of 2.5s the interesting band is visible at all, where 5s hid it. `GAVE UP` at 15s is the specific string that would argue for putting the budget back.

lvca:
> per JUnit Platform's tag-expression semantics, `any()` alone already matches every test (tagged or untagged), so `| none()` looks redundant

Good question, and the comment should have answered it. Measured rather than argued, and the answer is the opposite: **`| none()` is load-bearing.**

With `<groups>any()</groups>`, the same mixed selection:

| | `any()` alone | `any() \| none()` |
|---|---|---|
| `Issue5443SplitCompactionEntryIT` (`@Tag("slow")`) | ran | ran |
| `RaftVerifyDatabaseIT` (untagged) | **did not run** | ran |

and the build passed while dropping it. `any()` matches a class carrying *at least one tag*, not every class. Most of the ~118 ITs the default execution owns carry no tag at all, so `any()` alone would silently empty most of the lane - the same failure the expression exists to prevent, arrived at from the other side.

The pom comment now carries that measurement, so the next reader gets the answer instead of the question.

On the other two: the ~170-line comment block is long because each paragraph is a measurement somebody would otherwise have to redo, and I agree it's the authoritative copy - the PR description is the disposable one. And yes on watching the lane after the cut; your point that the 40-run sample *predates* the fork change is a sharp one I hadn't stated, so it's worth being explicit that the two variables land together and the first weeks of `TEST SLOW WAIT:` lines are the check on both, not just on the budget.

Last item outstanding is the `reuseForks=false` number; that branch's lane is still running.

lvca:
> `RaftBulkInsertCompactionRaceIT` and `RaftMigrationCompactionRaceIT` carry both `@Tag("ha-heavy")` and `@Tag("slow")` [...] a hypothetical future `-Dfailsafe.excludedGroups=slow` would still run those two classes even though every other module's `@Tag("slow")` classes would be skipped

Good catch, and it was self-inflicted: the previous commit restored `${failsafe.excludedGroups}` on the default execution but left the heavy one pinned, so the property meant **two different things inside one module**. Inert today, and exactly the surprise that costs somebody an hour on the day they reach for it. Fixed rather than accepted.

The heavy execution now composes the same property. It still has to be *written out* rather than left unset - unset means "the parameter is the user property", which is what `-DexcludedGroups` narrows - but written out it can compose the namespaced one and stay immune to the un-namespaced ones. Those are different knobs and now they behave like it.

Measured on a selection that includes a `slow`-only IT and a `slow`+`ha-heavy` one:

| invocation | default fork | heavy fork |
|---|---|---|
| plain | `Issue5443SplitCompactionEntryIT`, `RaftVerifyDatabaseIT` | `HeavyForkWiringIT`, `RaftMigrationCompactionRaceIT` |
| `-Dfailsafe.excludedGroups=slow` | `RaftVerifyDatabaseIT` only | `HeavyForkWiringIT` only - **`RaftMigrationCompactionRaceIT` now skipped too**, where before it ran |
| `-Dgroups=bogus -DexcludedGroups=bogus` | both | both |

That leaves `-Dfailsafe.excludedGroups=ha-heavy` emptying both executions, so those classes run nowhere. Deliberate, and it's the line worth drawing: **a filter aimed at something else must not be able to break the partition; a filter aimed at these classes by name, in a namespaced property nobody sets by accident, is allowed to skip them.** The pom says so where the expression is.

Also took the `Files.list` note - the scan is deliberately shallow and now says so, with the reason it cannot produce a false negative (every routed IT is flat in this package; a subdirectory is a different package this test has no claim over).

lvca:
### The `reuseForks=false` number has landed — and it closes the option

`experiment/6343-reuseforks-false` ran the module-wide variant through the real lane ([run 33045220743](https://github.com/ArcadeData/arcadedb/actions/runs/33045220743), all 238 tests, all 123 classes, nothing else changed):

| | minutes |
|---|---|
| baseline on `main`, 40 runs | 59.7 - 71.8 (p50 **62.2**, p90 68.2) |
| `reuseForks=false` | **86.9** |
| job `timeout-minutes` | 90 |

**+24.7 min on the median, +40%, 3.1 minutes of headroom left** — about twelve seconds per extra fork, which is JVM start plus classload plus cross-class JIT warm-up discarded 122 times.

Three minutes inside a timeout is not a margin; the next IT added to this module pushes the lane over it, and #5973 is what that looks like for this job. So the option issue #6343 named as the one that "would settle the question" is settled — as unaffordable, on a number rather than on the estimate this PR previously carried. The `ha-raft/pom.xml` rationale now quotes it, so the next person to wonder whether the whole-module option was ever tried finds the answer next to the code. Branch deleted.

That run also finished red (5 of 238). It says nothing either way and the commit records why: the same 40-run sample on `main` was green in 7 runs and carried 1-5 failures in the other 33, and this run's five are from that population. The timing is the measurement; the colour is not. It also emitted zero slow-wait lines, consistent with the 40 runs behind the budget cut.

That was the last item outstanding on this PR. Everything raised across the review rounds is either fixed or answered with a measurement:

- `failIfNoTests` — set, and its real (scan-level only) reach documented after testing both halves
- tag-filter narrowing via `-Dgroups` / `-DexcludedGroups` — locked out on both executions, verified hostile
- `${failsafe.excludedGroups}` — composed back in, and now meaning the same thing on both sides of the split
- `any() | none()` — measured to be load-bearing, not a redundant limb
- duplicate capturing logger — replaced by the shared one, extended with level capture and a per-call record that removes a real interleaving hazard
- logger restore-not-replace — fixed, including for the five pre-existing callers
- threshold ratio — asserted as equality, and an overstatement of mine about what a line means corrected
- fork-count arithmetic and the shallow class scan — reconciled and documented

lvca:
> step 5 [...] is manual-only [...] it's the one guarantee in this design that has no regression test behind it. Worth a `// TODO` or a follow-up issue

Follow-up issue: **#6794**. Deliberately not a `// TODO` - this entire issue exists because #6297's remaining work was left as a comment on a merge rather than as a tracker, and #6323 then found nothing searchable left behind. Adding a TODO here would be repeating that exact mistake inside the change that fixed it.

The checklist now also names step 5 as the unguarded one, says why it can't be automated from a unit test (it needs a real Maven subprocess with command-line properties), and notes the thing that makes it worth running by hand: its failure is the *quietest* one in the design - the default execution runs 0 tests and the build passes, so ~118 ITs leave the lane with no red mark anywhere. #6794 sketches both a `maven-invoker-plugin` answer and a cheaper `help:effective-pom` grep that checks the actual precondition for most of the value at none of the cost.

While there I also dropped `-Dfailsafe.excludedGroups=ha-heavy` from step 5's example command: since the symmetry fix that property is composed into *both* executions on purpose, so it is now expected to skip those classes rather than expected to be ignored. Leaving it in the "nothing may narrow this" step would have been contradicting the commit before it.

On your other two: `@Timeout(120)` vs ~344s is exactly as you read it (per-`@Test` vs the whole ordered class), and the flat-package assumption in the scan is documented for the same reason you flagged it.

Noted on CI - I won't consider this done until the lane is green, and I'll say so here either way.

lvca:
> the fragile-assertion note is the only thing I would actually ask to be tweaked

Done, and you're right about why it matters: the failure would have been *misdirecting*, not just wrong. `polyglot.engine.WarnInterpreterOnly` belongs to the root POM for a reason of its own, and pinning `"false"` meant an unrelated edit in another file would turn this red under a message insisting the merge was broken when it wasn't - sending whoever investigated to the wrong file entirely. That is worse than not checking.

It now asserts **presence**, which is the whole claim: the property reached this fork, so the execution-level `<systemPropertyVariables>` merged rather than replaced. The message says so explicitly - *"any value at all passes this - only its absence is the finding"* - so it can't be misread as an opinion about the value.

> If the module's failsafe/surefire version is ever bumped, it would be worth re-running the manual verification steps

Good point, and the checklist only told you to re-run after editing *that block*. It now also covers a version bump, with the reason: two of those expressions rest on the tag-expression grammar behaving as measured on 3.5.6 (`any()` meaning "at least one tag" rather than "every class", `any() & none()` being satisfiable by nothing), and a grammar change **would not error - it would run zero tests and pass**. Exactly the failure mode you identified.

> Worth confirming that manual check was actually run before merging

Confirmed, and re-run just now against this tree rather than quoted from an earlier one, since the pom has changed since the last time I ran it:

```
./mvnw -o -pl ha-raft verify -Pintegration -DskipTests \
  -Dit.test=HeavyForkWiringIT,RaftVerifyDatabaseIT -DfailIfNoSpecifiedTests=false \
  -Dgroups=bogus-tag -DexcludedGroups=bogus-tag

  --- integration-test (default)                   -> RaftVerifyDatabaseIT
  --- integration-test (heavy-its-in-their-own-fork) -> HeavyForkWiringIT
```

Partition intact under both hostile flags. I'll confirm the `ha-integration-tests` lane here too before this is done - it's running now.

lvca:
> With `-Dfailsafe.excludedGroups=ha-heavy` [...] That's 0 matched tests in that execution, which is exactly what `failIfNoTests=true` is designed to turn into a build failure [...] the "run nowhere, deliberate" outcome described in the comment would actually be "the whole `verify` fails"

Measured it, because it is a question with an answer, and the answer is that the two settings **do not collide**:

```
--- failsafe:3.5.6:integration-test (default)                    Tests run: 0
--- failsafe:3.5.6:integration-test (heavy-its-in-their-own-fork) Tests run: 0
--- failsafe:3.5.6:verify (default)                              No tests to run
--- failsafe:3.5.6:verify (heavy-its-in-their-own-fork)          No tests to run
BUILD SUCCESS        (exit 0)
```

The reason is the one already documented at `failIfNoTests` but never connected to this case: **that parameter is evaluated against the class scan, before the tag filter runs.** Zero matches from a tag expression are not "no tests to run" as far as it is concerned - which is precisely why it can't guard the tag-drift case either, the finding from an earlier round in this PR. Both halves of that behaviour were measured against 3.5.6:

| heavy execution configured with | outcome |
|---|---|
| `<includes>` matching no file | `No tests were executed!` - build fails |
| tag expression matching none of the scanned classes | `Tests run: 0` - build passes |

So the comment was accurate, but it earned the misreading by never addressing the combination - two settings sitting a dozen lines apart that look like they contradict each other. It now states the reconciliation, with the measured output, and names that run as **the one to do first if somebody is about to "fix" either setting because the other appears to contradict it**. Added as step 6 of the checklist so it gets re-checked rather than re-derived.

On the two minor notes: the `getCodeSource()`-is-a-directory assumption is real and I've left it, since it fails loudly (the test asserts `isDirectory()` before scanning) rather than silently under-scanning. And the comment length is fair criticism - two reviewers have now said it. My bias here is that every paragraph is a measurement somebody would otherwise repeat, and several of them were written *because* a reviewer reached the wrong conclusion from the shorter version, this round included. But it is worth a trimming pass once this has been stable for a release or two.

lvca:
> **Log line verbosity** [...] worth confirming that's the intended log format long-term rather than a one-time editorial note that could move to the class javadoc

The latter — good instinct, and it's moved. That clause was static, repeated in every occurrence, and written for a reader who ends up at `RESYNC_RETRY_TIMEOUT_MS`'s javadoc anyway, where it already says the same thing at more length than a log line should. What's left is only what's load-bearing for a grep or for someone reading a single line — the marker, which wait, how long, out of how much:

```
TEST SLOW WAIT: withResyncRetry on server 0 satisfied after 3131 ms of the 15000 ms budget (issue #6343)
```

That's a real line from a local run just now, and incidentally the **second** sighting of the 2.5s threshold earning its place: 3131 ms here, 3013 ms earlier in this PR. Both were invisible at the old 5s, which is the resolution the cut was for.

On the other three: the `getCodeSource()` assumption I've left as-is for the reason you give (it fails loudly on `isDirectory()` rather than under-scanning); step 5 is #6794; and the "does the ratchet hold on a slower runner tier" question is a good one for the day that happens — the instrument is now exactly the thing that would answer it, since a `GAVE UP` at 15s is the string that argues for putting the budget back.

Waiting on `ha-integration-tests` here, which is the last thing outstanding.

lvca:
> Worth considering whether the re-verification recipe (steps 1-6) belongs in `ha-raft/CLAUDE.md` instead

Done — and it's the right diagnosis of a complaint three reviewers in a row have now made. The recipe is a *checklist*, and a checklist is read **before** an edit, which is exactly when nobody is reading an XML comment thirty lines above the configuration it describes. `ha-raft/CLAUDE.md` already exists for this module's things-the-code-won't-tell-you notes, so the six steps live there now under "Re-verifying the IT fork split", with a pointer left behind.

Twenty-five lines out of the pom. The half that stays is the half that belongs there: why the split exists, what each tag expression is for, and the measurement behind each one — next to the configuration it explains, where somebody editing that line will actually see it.

On the rest: `getCodeSource()` I've left (it fails loudly on `isDirectory()` rather than under-scanning); step 5 is #6794; and the JVM arithmetic taking two reads is fair — it now reads better in the CLAUDE.md framing, though the netting-out is genuinely what makes "five, not six" true.

Also worth being explicit, since you flagged relying on the PR description for dynamic checks: everything I've claimed here was run. 979 unit tests, the eleven `withResyncRetry`-calling IT classes under the 15s budget, the partition/pid/failure-attribution checks, both hostile-flag runs, and the `-Dfailsafe.excludedGroups=ha-heavy` case — each with its output quoted in the commit or comment that claims it. The one thing I can't self-certify is the full lane, which is running here now and which I'll confirm before calling this done.

lvca:
> 10s / 120s is 1/12, not 1/6 - only the two most recent cuts are actually a sixth

Correct, and it's the sharpest catch in this PR, because it's a false claim in the one place this change can least afford one. Fixed in all three places (the `SLOW_WAIT_REPORT_MS` javadoc, the test's field javadoc, and its assertion message).

Checking the arithmetic showed the sentence was wrong a **second** way, which you didn't have to point out. It also claimed the fraction "never grows":

| | threshold / budget | consumable in silence |
|---|---|---|
| #6221 | 10s / 120s | 8.3% |
| #6267 | 5s / 30s | **16.7%** ← doubled |
| #6343 | 2.5s / 15s | 16.7% |

The #6267 cut *halved* the resolution even as it improved the budget. So the rule was never "it has always been a sixth" — it is **a sixth since #6267**, and what this cut does is hold it there rather than let it slip to a third, which is exactly what leaving the threshold at 5s under a 15s budget would have done.

The equality test was already right (it pins the current 2500/15000 and nothing historical), so this is documentation only. But a PR that asks to be trusted on forty runs of measurement doesn't get to be sloppy about the three numbers in its own comment. The twelfth stays in the text as the reminder it actually is: a budget can be so large that even a generous threshold sees nothing — which is precisely how the 120s survived unexamined.

On your `package-info.class` note: good spot, and it confirmed rather than tripped the fail-closed path — it loads normally, so the strict handler stays strict without a special case. Nothing to change.

lvca:
> If someone passes `-Dfailsafe.excludedGroups=` explicitly, the expression becomes `ha-heavy | ()`, which likely fails to parse [...] worth a one-line defensive note, or confirming JUnit's tag-expression parser tolerates empty parens gracefully

It does **not** tolerate them — measured rather than assumed:

```
[ERROR] org.junit.platform.commons.PreconditionViolationException: Unable to parse tag expression
        "ha-heavy | ()": missing rhs operand for '|' at index <9>
BUILD FAILURE
```

Which is the right failure for it to have, and the reason I've documented it rather than defended against it. The distinction this whole block turns on is loud versus silent: nothing gets skipped, nothing passes while running zero tests, and the message names both the expression and the position in it. A blank override is a typo you find immediately.

It's also why the "nothing" value here is a contradiction rather than the empty string the parent uses — the note now says that explicitly, along with: if you want this module's ITs unfiltered, leave the property alone; blank is not the way to say so.

On the other two points, agreed on both and neither needs a change: the 15s cut is the one thing here that isn't purely infrastructure-internal and does want watching after merge — which is precisely what the `TEST SLOW WAIT:` marker now makes a one-line grep, with a `GAVE UP` at 15s being the string that would argue for putting it back. And #6794 is filed and assigned for step 5.

lvca:
## ⚠️ The lane is red, and it is this PR's fault. Not ready to merge.

Two `ha-integration-tests` runs on this branch, same failure both times:

```
Issue5410AbandonedTicketReleaseIT.abandonedEntryReleasesItsPhase2TicketOnceApplied
  [Node 1 must hold both vertices] expected: 2L but was: 1L

TEST SLOW WAIT: awaitValue(2) GAVE UP after 15038 ms of the 15000 ms budget   (run 1)
TEST SLOW WAIT: awaitValue(2) GAVE UP after 15023 ms of the 15000 ms budget   (run 2)
```

The instrument did exactly what it was built for, on its first real CI run, and what it caught is my own change.

### Why this is not a flake

I sampled 12 `ha-integration-tests` runs on `main`: **`Issue5410AbandonedTicketReleaseIT` fails in none of them.** (`Issue5569SlotMergeDeleteRaftIT`, the other failure in run 1, appears in 4 of those 12 — that one *is* the known flake, and it is not mine.)

And across the 40-run sample behind this PR's budget cut, **not one slow-wait line was emitted at all** — with a 5s threshold. So on `main` this wait completes in under 5 seconds, every time. On this branch it exceeds 15 and the vertex still has not arrived ~2.7s later, when the teardown comparator also disagrees:

```
DatabaseComparator$DatabaseAreNotIdentical: Page PageId(graph/28/0) ... DB1 2 <> DB2 1
```

### What I think is happening — and the part I got wrong

A budget cut cannot make replication slower; it only decides when to stop waiting. So the wait itself changed, which leaves the fork split. The class *order* is identical to `main` (index 27, same predecessors), but the JVM history is not: `RaftHAComprehensiveIT` — 344s of continuous three-server activity, three classes earlier — no longer runs in that fork. The light classes were free-riding on its JIT warm-up.

Which is the same cost I cited as the reason *not* to go module-wide with `reuseForks=false`. I did not consider that the partial split has a scaled-down version of it, and it means **the 40-run measurement justifying 15s was taken in a world that item 2 changes.** The two items in #6343 are not independent, and I treated them as if they were.

There is a worse possibility I cannot yet rule out: that the second vertex never arrives at all, in which case the split has exposed a genuine replication bug rather than a latency change.

Holding this PR until that is settled.

lvca:
> `slowWaitReport`'s javadoc says "the line `{@link #reportSlowWait}` logs" - technically it's `logSlowWait` that logs it

Fixed. Trivial, but worth doing rather than waving through: the entire argument of this PR is that a comment cited as evidence has to be accurate, and a `{@link}` pointing one method to the left of the truth is the small version of the same fault that produced the `SLOW WAIT` grep returning zero and the 10s/120s "sixth" that wasn't.

On the other two: the long comment block is deliberate for the reason you give — this constant has been tuned three times on contested evidence, and a PR description doesn't survive `git blame`. And yes, the final lane is what I'm waiting on; the PR does not get called done until it's green, and I'll say so here either way.

Worth restating what the lane is now testing, since the diff changed underneath this thread: the fork split is **gone** from this PR (withdrawn to #6848 after it made the lane red 3/3), so what remains is `main`'s fork arrangement plus a budget whose wait was directly measured at 252 ms. That's the configuration the forty runs sampled, which is the whole reason it can be trusted.

lvca:
No blocking issues, and I'm not changing anything for the three minor notes — but one of them is worth correcting rather than quietly accepting.

### `formattedAt(Level)` is value equality, not reference equality

> `level.equals(e.level())` — `java.util.logging.Level` doesn't override `equals`, so this is reference equality

It does override it, comparing by `intValue()`. Checked against this JDK rather than from memory:

```
Level overrides equals?           true
WARNING.equals(customSameValue) = true      // a distinct Level instance with WARNING's value
WARNING == customSameValue      = false     // …which reference equality would have rejected
WARNING.equals(INFO)            = false
```

So the fragility described isn't there: a caller constructing a custom `Level` with `WARNING`'s value matches, which is the behaviour you'd want anyway. Leaving it as `equals`, which is also the correct-by-default choice regardless.

### The other two, deliberately unchanged

- **Bisection assumes monotonicity in `elapsedMs`** — true, and it's the right assumption to bake in: `slowWaitReport` returning non-monotonically in elapsed time would itself be the bug. If rate-limiting is ever added, the thing that should change is the report contract, and this test failing to find a clean boundary is a reasonable way to be told.
- **Test constant coupled to the two in `BaseRaftHATest`** — intentional, and the javadoc on both says to change them together. That coupling *is* the invariant; decoupling them is what let the "held at a sixth" claim drift into being false for a third of its own examples.

### On the coverage gap

> no test asserting `withResyncRetry`/`awaitValue` actually pass `RESYNC_RETRY_TIMEOUT_MS` as the deadline

Real gap, and I've judged it not worth closing: a direct test means letting `awaitValue` run to its deadline, i.e. a 15-second unit test whose assertion is raw wall clock — which `CLAUDE.md` specifically warns against for exactly the reason that would bite here (a full-suite GC pause turns the bound into a coin flip). The indirect coverage is decent: the budget printed in every report *is* the constant, `budgetMs()` parses it back out of the real message, and the HA ITs exercise the deadline for real — as this PR discovered rather loudly, since a `GAVE UP after 120 022 ms of the 120000 ms budget` line is precisely what identified #6848.

lvca:
## Lane confirmed: green at `main`'s baseline

[Final run](https://github.com/ArcadeData/arcadedb/actions/runs/33099163227/job/98613970408), 61m29s — against a measured baseline of 59.7-71.8 min (p50 62.2), so no lane-time cost either:

```
Tests run: 238, Failures: 0, Errors: 1, Skipped: 4
```

The single error is `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` — **the known `main` flake**, present in 5 of the 12 `main` runs I sampled while investigating, and not touched by this PR.

**`Issue5410AbandonedTicketReleaseIT` passes**, as it does on `main`, which closes the loop on the investigation: it failed 3/3 only while the fork split was in this branch, and it is green now that the split has moved to #6848.

And the instrument, on the shipped configuration:

```
TEST SLOW WAIT: withResyncRetry on server 0 satisfied after 3002 ms of the 15000 ms budget (issue #6343)
```

One line, 20% of budget, no `GAVE UP`. That is the shape the 15s cut was chosen for: enough headroom that a real wait reports comfortably inside it, little enough that a genuine hang is caught in 15s rather than 30. It is also a wait the old 5s threshold could not see at all, so the resolution improvement is doing visible work on its first shipped run.

### Where this ended up

Everything raised across ~12 review rounds is either fixed or answered with a measurement, and the reviews have been returning "no blocking issues" for several rounds. The substantive corrections along the way, for the record:

- the `SLOW WAIT` marker — this issue's own evidence-gathering grep was searching for a string that had never been emitted
- `failIfNoTests` is evaluated pre-tag-filter, so it cannot guard the case it was suggested for
- `-Dgroups=bogus` silently emptied an execution; `any()` alone drops every untagged IT
- the ratio assertion was an upper bound where the doc claimed an equality — and the doc's own history was wrong twice (10s/120s is a twelfth, and the fraction *did* grow at #6267)
- `CapturingTestLogger`'s parallel lists could pair one thread's level with another's text; `uninstall()` clobbered the caller's logger
- and the big one: **the fork split made the lane red 3/3 and, once measured properly, turned out to expose #6848 rather than cause it**

Ready to merge from my side.

PR #6790, opened 2026-08-27
Update Python bindings: GraphBatch retry/cache knobs, importer on_row_error, RESTORE and wire-protocol tests

tae898:
Periodic update of `bindings/python` from the fork. Regenerated from `main` onto current `upstream-main`, so this branch is your tree plus the bindings surface; nothing else is touched.

8 files, +512/-26, and most of it is tests.

## Source changes

**`graph_batch.py` / `core.py`: four `GraphBatch` knobs exposed and validated Python-side.**

- `commit_retries` — times a vertex-creation commit is retried on a transient `NeedRetryException`, e.g. a Raft `QuorumNotReachedException` during a leader re-election. Default 10; `0` fails fast.
- `commit_retry_delay_ms` — initial back-off, exponential after that, capped at 10 s. Default 1000.
- `chunk_cache_capacity` — bound on each of the OUT/IN head-chunk RID lookup caches. Both are pure accelerators, so a miss just reloads the head chunk.
- `max_deferred_incoming_edges` — bound on deferred incoming edges.

The Java side already accepted these; they were unreachable from Python. Invalid values are now rejected at the binding boundary rather than surfacing as a Java exception mid-batch.

**`importer.py`: `on_row_error` exposed and validated.** Unknown values are rejected with the accepted set named, rather than being passed through to fail later.

## Tests

`test_restore_sql.py` (new, 246 lines) covers `RESTORE DOCUMENT`/`RESTORE VERTEX` including the two defects filed from this suite: #6069 (the bucket record count was not restored, so `count(*)` under-reported) and #6096 (`RESTORE` in the same transaction as the `DELETE`). Both are fixed upstream; these pin the behaviour.

`test_graph_batch.py` (new) covers the four knobs above and their rejection paths. `test_importer_api.py` (new) covers `on_row_error`. `test_server_wire_protocols.py` gains coverage for the bundled Postgres and Bolt plugins.

`test_core.py` carries one `# nosec B608`, on the regression test that arrived with the `ColumnBatcher` escaping fix (#6758). The interpolated values are the test: they are literals containing a quote, a backslash and a semicolon, chosen because those are what that fix exists to survive, and SQL has no parameter form for a column alias. Our bandit job scans `tests/` at low/low where this repo's does not, so it passed here and failed there on the merge; the marker documents why the interpolation is deliberate rather than removing the thing being tested.

## Verification

Built from a local assembly of this commit's jars and run against it:

- `427 passed, 5 skipped` on linux/amd64, Python 3.12
- `bandit -c pyproject.toml -r src tests --severity-level low --confidence-level low` clean
- `bandit -c pyproject.toml -r examples --severity-level medium --confidence-level high` clean
- pre-commit (black, isort, the file hygiene hooks) clean

One note on the CI side, unchanged from last time but worth repeating: the workflow in this PR installs `psycopg[binary] redis neo4j pyarrow pandas` alongside pytest, and fails the job if any test reports `could not import`. Without that, the wire-protocol and `to_arrow()`/`to_dataframe()` tests skip silently and the build stays green — which is how `to_dataframe()` went without executed coverage for a while.

## Summary by CodeRabbit

* **New Features**
  * Added configurable commit retries, retry delays, chunk-cache capacity, and deferred incoming-edge limits for graph batch operations.
  * Added row-error handling for document imports, supporting case-insensitive `abort` and `skip` modes.

* **Bug Fixes**
  * Improved validation and handling of invalid graph batch and import settings.
  * Expanded restore coverage for records, indexes, transactions, and persistence.
  * Updated Redis port verification to reflect the completed upstream fix.

### DIFF
--- a/bindings/python/src/arcadedb_embedded/core.py
+++ b/bindings/python/src/arcadedb_embedded/core.py
@@ -834,6 +834,10 @@ class Database:
         wal_flush: Optional[str] = None,
         pre_allocate_edge_chunks: Optional[bool] = None,
         parallel_flush: Optional[bool] = None,
+        commit_retries: Optional[int] = None,
+        commit_retry_delay_ms: Optional[int] = None,
+        chunk_cache_capacity: Optional[int] = None,
+        max_deferred_incoming_edges: Optional[int] = None,
     ) -> GraphBatch:
         """
         Create a GraphBatch helper for high-throughput graph ingestion.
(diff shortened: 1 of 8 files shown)


robfrank:
As usual, thank you @tae898 !

Issue #2759, opened 2025-11-06
Bug: XMLImporterFormat Carries Over Properties Between Records


tae898:
## Description

The `XMLImporterFormat` class has a critical bug where properties from previous records carry over to subsequent records when attributes are missing. This happens because the `Map<String, Object> object` is created once and reused for all records without being cleared.

## Impact

When importing XML files where records have different sets of attributes, records will incorrectly inherit attribute values from previous records.

## Example

Given this XML:

```xml
<users>
  <user id="1" name="Alice" age="30"/>
  <user id="2" name="Bob"/>
</users>
```

**Expected behavior:** User 2 should NOT have an `age` property  
**Actual behavior:** User 2 incorrectly has `age="30"` (carried over from User 1)

## Root Cause

**File:** `integration/src/main/java/com/arcadedb/integration/importer/format/XMLImporterFormat.java`

**Line 60:** The object map is created once:
```java
final Map<String, Object> object = new LinkedHashMap<>();
```

**Lines 99-104:** After creating each record, the map is NOT cleared:
```java
if (nestLevel == objectNestLevel) {
  context.parsed.incrementAndGet();
  final MutableVertex record = database.newVertex(entityName);
  record.fromMap(object);
  database.async().createRecord(record, newDocument -> context.createdVertices.incrementAndGet());
  // BUG: object.clear() is missing here!
}
```

## Proposed Fix

Add map cleanup after each record is created:

```java
if (nestLevel == objectNestLevel) {
  context.parsed.incrementAndGet();
  final MutableVertex record = database.newVertex(entityName);
  record.fromMap(object);
  database.async().createRecord(record, newDocument -> context.createdVertices.incrementAndGet());
  
  // Clear the map for the next record
  object.clear();
  lastName = null;
  lastContent = null;
}
```

## Workaround

Until this is fixed, ensure all XML records have the same complete set of attributes. Missing attributes will inherit values from previous records.

PR #6844, opened 2026-08-27
fix(#6813): XML import - an empty or self-closing sub-element inherits the previous sibling's text

### LINKED ISSUE #6813: XML import: an empty or self-closing sub-element inherits the previous sibling's text value
### Summary
In the XML importer, `lastContent` is cleared only when a new **record** starts, never when a new sub-element starts. An empty or self-closing sub-element therefore inherits the previous sibling's text value, in both the import and the schema-analysis pass.

### Where (HEAD `9c9d024db`)
`integration/src/main/java/com/arcadedb/integration/importer/format/XMLImporterFormat.java:95-101`:
```java
if (lastName != null)
  object.put(lastName, lastContent);
...
lastName = getQualifiedName(xmlReader.getName());   // lastContent left as is
```
The CHARACTERS handler deliberately refuses to null it out (`:170`):
```java
// Note: We no longer set lastContent to null for empty/newline-only text
```
`analyze()` has the identical omission at `:250-256`, so the inferred schema is polluted the same way.

### Repro
```xml
<rows><row><a>1</a><b></b></row></rows>
```
(or `<b/>`, or any pretty-printed file where `b` is genuinely empty) imported with `objectNestLevel=1` produces `{a: "1", b: "1"}` instead of `b` absent or empty. Event by event: START `<b>` stores `a=1` and sets `lastName="b"`; END `</b>` runs `object.put("b", lastContent)` with `lastContent` still `"1"`.

This is the sub-element twin of #2759, which `XMLImporterFormatTest.noPropertyCarryoverBetweenRecords` fixed for **attributes** only - that test uses `<user id=.. name=.. age=../>` attributes exclusively, so the sub-element path is untested.

### Suggested fix
Set `lastContent = null` where `lastName` is assigned (`:100` and `:255`), and again after each `object.put(lastName, lastContent)` in the END_ELEMENT handler. That preserves the #2759 whitespace fix - whitespace *between* elements still cannot erase the content of the element currently being read - while stopping the value from leaking across siblings.

### Scope / confidence
High. Walked the full StAX event sequence including the whitespace CHARACTERS events; the comment at `:168-171` concerns whitespace inside the currently open element, which the `trimmedText.isEmpty()` guard already handles, so clearing on `lastName` assignment does not regress it.

robfrank:
Closes #6813

## Summary

`XMLImporterFormat` cleared its `lastContent` cursor only when a new **record** started, never when a new **sub-element** started, so an empty (`<b></b>`) or self-closing (`<b/>`) sub-element inherited the text of the sibling that preceded it. Walked event by event for `<row><a>1</a><b/></row>` at `objectNestLevel=1`: the START of `<b>` flushes `a=1` and sets `lastName="b"` while leaving `lastContent="1"`, and the END of `</b>` then runs `object.put("b", lastContent)` with that stale `"1"`. The fix sets `lastContent = null` at the two points where `lastName` is assigned to a new sub-element - `load()` and `analyze()` - so a sub-element can only ever be filled by CHARACTERS events that arrived after its own START.

This is the sub-element twin of #2759, whose regression test (`noPropertyCarryoverBetweenRecords`) uses attributes exclusively and therefore never exercised this path.

`analyze()` is patched the same way, since the identical omission polluted the inferred schema: an empty `<score/>` sampled the value of the `<name>` before it, which demoted the inferred type of `score` from LONG to STRING. While there, its record-start reset is corrected too - `lastName = null` sat **inside** the attribute loop, so it never fired for a record element without attributes. It now clears both `lastName` and `lastContent` unconditionally at record start, mirroring what `load()` has always done.

### Why this does not regress #2759

#2759 taught the CHARACTERS handler to ignore whitespace-only text, so that pretty-printing between two elements cannot erase the content of the element currently open. That guard (`trimmedText.isEmpty()`) is untouched. The new clear fires on START_ELEMENT, i.e. at the moment the reader moves *into a different* element, which is exactly when the previous element's text stops being relevant. The regression test deliberately uses a pretty-printed document so the whitespace CHARACTERS events are present in the stream.

### Why the fix is not also applied after `object.put(...)` in END_ELEMENT

The issue text suggested additionally nulling `lastContent` after each `put` in the END_ELEMENT handler. That alone would break the importer: END_ELEMENT leaves `lastName` set, so the next sub-element's START would run `object.put(lastName, null)` and overwrite the value just captured. Clearing at the `lastName` assignment site is both sufficient and minimal.

## Changes

- `integration/src/main/java/com/arcadedb/integration/importer/format/XMLImporterFormat.java`
  - `load()`: clear `lastContent` when a new sub-element becomes the current `lastName`.
  - `analyze()`: same clear, plus an unconditional `lastName`/`lastContent` reset at record start (was trapped inside the attribute loop).
- `integration/src/test/java/com/arcadedb/integration/importer/XMLImporterFormatTest.java`
  - `noContentCarryoverBetweenSubElements` - end-to-end `IMPORT DATABASE` over a pretty-printed file covering an empty element, a self-closing element, an empty element *followed* by a populated sibling, and a populated element that must keep its own text.
  - `analyzeDoesNotCarryContentBetweenSubElements` - drives `analyze()` directly and asserts on the sampled values and the inferred `Type`.

Both tests were confirmed to fail on the unfixed code (`Tests run: 4, Failures: 2`) and pass after the change.

## Test plan

- [ ] `mvn -o -pl integration verify` - `Tests run: 169, Failures: 0, Errors: 0, Skipped: 9`
- [ ] `mvn -o -pl integration verify -DskipITs=false -DskipTests=true` - `Tests run: 118, Failures: 0, Errors: 0`
- [ ] `XMLImporterIT.importXMLWithNamespace` still green: it reads nested `<datafield><subfield>` content, the shape most likely to be affected by a `lastContent` reset, and the last datafield's text is still captured.
- [ ] `XMLImporterFormatTest.noPropertyCarryoverBetweenRecords` (#2759) still green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed XML imports so empty and self-closing sibling elements no longer inherit text from adjacent elements.
  * Improved schema analysis to prevent empty XML elements from affecting inferred property values or numeric types.

* **Tests**
  * Added regression coverage for XML importing and schema inference, including UTF-8 content and parser scenarios.

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/importer/format/XMLImporterFormat.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/format/XMLImporterFormat.java
@@ -98,6 +98,10 @@ public class XMLImporterFormat implements FormatImporter {
               object.put(lastName, lastContent);
 
             lastName = getQualifiedName(xmlReader.getName());
+            // START OF A NEW SUB-ELEMENT: THE PREVIOUS SIBLING'S TEXT MUST NOT LEAK INTO IT (ISSUE #6813).
+            // THIS DOES NOT REGRESS ISSUE #2759: THE WHITESPACE BETWEEN TWO ELEMENTS IS SKIPPED BY THE CHARACTERS
+            // HANDLER, SO THE TEXT OF THE ELEMENT CURRENTLY BEING READ IS STILL SAFE FROM PRETTY-PRINTING
+            lastContent = null;
           }
 
           ++nestLevel;
@@ -241,18 +245,22 @@ public class XMLImporterFormat implements FormatImporter {
               entityName = "v_" + qualifiedName;  // v_ prefix for vertices (including DATABASE for backward compatibility)
             }
 
+            // START OF A NEW RECORD: RESET THE PENDING SUB-ELEMENT, AS load() DOES
+            lastName = null;
+            lastContent = null;
+
             // GET ELEMENT'S ATTRIBUTES AS PROPERTIES (with namespace prefix if present)
-            for (int i = 0; i < xmlReader.getAttributeCount(); ++i) {
+            for (int i = 0; i < xmlReader.getAttributeCount(); ++i)
               analyzedSchema.getOrCreateEntity(entityName, entityType)
                   .getOrCreateProperty(getQualifiedName(xmlReader.getAttributeName(i)), xmlReader.getAttributeValue(i));
-              lastName = null;
-            }
           } else if (nestLevel == objectNestLevel + 1) {
             // GET ELEMENT'S SUB-NODES AS PROPERTIES
             if (lastName != null)
               analyzedSchema.getOrCreateEntity(entityName, entityType).getOrCreateProperty(lastName, lastContent);
 
             lastName = getQualifiedName(xmlReader.getName());
+            // START OF A NEW SUB-ELEMENT: THE PREVIOUS SIBLING'S TEXT MUST NOT BE SAMPLED AS THIS ONE'S (ISSUE #6813)
+            lastContent = null;
           }
 
           ++nestLevel;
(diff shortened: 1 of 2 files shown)


robfrank:

PR #6845, opened 2026-08-27
fix(#6814): AnalyzedProperty keeps refuting LONG/DOUBLE after sampling stops

### LINKED ISSUE #6814: AnalyzedProperty treats a value longer than 100 chars as no evidence instead of disproving LONG/DOUBLE, so the column is typed numeric and the import then fails on that value
### Summary
`AnalyzedProperty.setLastContent()` treats a value longer than 100 chars as "stop sampling" **and** implicitly as "no evidence about the type", returning before the numeric probes and permanently disabling further analysis of that column. A column whose first value looks numeric is then declared `LONG`, and the import fails on the long value it never examined.

### Where (HEAD `9c9d024db`)
`integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java:68-73`:
```java
if (lastContent != null) {
  if (lastContent.length() > 100) {
    collectingSamples = false;
    contents.clear();
    return;                    // <- before the Long/Double probes on :84-100
  }
```
The `return` also skips updating `this.lastContent`, and because `collectingSamples` is now permanently `false`, the guard on `:65` makes every subsequent value for that column a no-op too. `endParsing()` (`:52-58`) then only checks `lastContent != null && candidateForInteger`, so the column is declared `Type.LONG`, and `AbstractImporter.updateDatabaseSchema()` (`:248`) creates the schema property with that type.

### Repro
```
id,description
1,42
2,"<any 150-char sentence>"
3,not a number
```
Analysis sees `"42"` (LONG candidate), bails on row 2, and ignores row 3, so `Document.description` is created as `LONG`. `CSVImporterFormat.loadDocuments()` (`:167`) then calls `document.set("description", "<150-char sentence>")`, `MutableDocument.convertValueToSchemaType()` reaches `Type.convert(... Long.class ...)` -> `Long.parseLong` -> `NumberFormatException`, rethrown as:
```
IllegalArgumentException: Cannot convert type 'java.lang.String' to 'LONG' found in property 'description'
```
In the default `-onRowError abort` mode that kills the whole import, and the exception is rethrown raw at `CSVImporterFormat.java:184` with no line number. Under `-onRowError skip` every long-text row is silently dropped instead.

The analysis limits (`analysisLimitBytes=100000`, `analysisLimitEntries=10000`) do not make the offending row unreachable. `CSVImporterIT` has no long-text column and `AnalyzedProperty` has no unit test.

### Suggested fix
Clear `candidateForInteger`/`candidateForDecimal` before returning in the `length() > 100` branch (a token that long is not a `Long`, and in practice not a `Double` literal either). Better still, run the two numeric probes before the sampling short-circuits, so stopping sample collection never stops type refutation. The same reasoning applies to the `contents.size() > maxValueSampling` short-circuit at `:76-80`.

### Scope / confidence
High. Verified that the analyzed type is applied to the schema rather than advisory, and that `Type.convert` for `Long.class` really throws on a non-numeric string.

robfrank:
Closes #6814

## Problem

`AnalyzedProperty.setLastContent()` conflated two unrelated decisions: *"stop collecting samples"* and *"stop looking at values at all"*. A value longer than the 100-character sampling limit flipped `collectingSamples` off and `return`ed **before** the `Long`/`Double` probes, so that value never disproved anything; because `collectingSamples` is never re-armed, the guard at the top of the method then made every subsequent value for that column a no-op too. `endParsing()` only checks `lastContent != null && candidateForInteger`, so a column whose first value happened to look numeric was declared `Type.LONG`, and `AbstractImporter.updateDatabaseSchema()` created the schema property with that type. `CSVImporterFormat.loadDocuments()` then set the long text value on it and `MutableDocument.convertValueToSchemaType()` threw:

```
IllegalArgumentException: Cannot convert type 'java.lang.String' to 'LONG' found in property 'description'
```

In the default `-onRowError abort` mode that killed the whole import; under `-onRowError skip` every long-text row was silently dropped. The same short-circuit existed on the `contents.size() > maxValueSampling` branch, so a text value arriving after the sample-count limit could not undo a LONG verdict either.

## Fix

`setLastContent()` is reordered so the two concerns are independent, in this order:

1. **Type refutation runs for every non-empty value**, unconditionally. It is evidence and must never be short-circuited.
2. **Sample collection** then stops as before, on an oversized value or once `maxValueSampling` distinct values have been collected, and still clears `contents`. Sampling is a memory bound and legitimately stops.

The probes are the real `Long.parseLong` / `Double.parseDouble` calls rather than a blanket "long value means text": 150 digits overflow a `Long` but are a perfectly good `Double`, so such a column now narrows to `DOUBLE` instead of failing on a `LONG` it cannot hold. The 100-character limit is now the named constant `MAX_SAMPLE_LENGTH`.

`this.lastContent` is now also assigned for oversized values. That only feeds `endParsing()`'s "did this column ever have a value" check, and a column made only of long text still ends up `STRING` because both candidates are refuted by those same values.

### Cost

Each probe runs only while its candidate is still true, so the first non-numeric value in a text column refutes both and every later value costs nothing. For a column that stays numeric, the extra work is bounded by the analysis window itself (`analysisLimitBytes=100000`, `analysisLimitEntries=10000` in `CSVImporterFormat.analyze()`), so worst-case total parsing stays within the ~100 KB the analyzer already reads.

## Tests

New `Issue6814AnalyzedPropertyTypeRefutationTest` (8 tests). All 5 bug-facing tests fail on `main` and pass with the fix; the 3 control tests pass on both, proving the assertions are wired to the right behaviour.

| Test | What it pins |
|---|---|
| `aValueLongerThanTheSamplingLimitDisprovesLong` | the report's core: `"42"` then a 120-char sentence must yield `STRING`, not `LONG` |
| `valuesAfterTheSamplingLimitStillRefuteTheNumericTypes` | analysis does not go permanently deaf after the oversized value |
| `aLongNumericValueNarrowsToDoubleInsteadOfLong` | 150 digits refute `LONG` (overflow) but not `DOUBLE` - a real probe, not a blanket rule |
| `valuesArrivingAfterTheSampleCountLimitStillRefuteTheNumericTypes` | the second short-circuit, on `maxValueSampling` |
| `csvImportKeepsALongTextColumnAsString` | end-to-end repro through `Importer`: schema property is `STRING`, all 3 rows land, no errors |
| `samplingStillStopsAndDiscardsCollectedValues` | control - the memory bound is intact (`isCollectingSamples()` false, `contents` empty and staying empty) |
| `aColumnOfOnlyLongTextIsStringAndNullsAreIgnored` | control - all-long-text stays `STRING`, `null` is still no evidence |
| `purelyNumericColumnsAreStillTypedNumeric` | control - `LONG`/`DOUBLE` inference still works, empty strings still ignored |

Fixture `integration/src/test/resources/importer-documents-long-text.csv` is exactly the repro from the issue: numeric-looking first value, a 120-character sentence, then another numeric value.

Before the fix:

```
Tests run: 8, Failures: 4, Errors: 1
  aLongNumericValueNarrowsToDoubleInsteadOfLong: expected DOUBLE but was LONG
  aValueLongerThanTheSamplingLimitDisprovesLong: expected STRING but was LONG
  valuesAfterTheSamplingLimitStillRefuteTheNumericTypes: expected STRING but was LONG
  valuesArrivingAfterTheSampleCountLimitStillRefuteTheNumericTypes: expected STRING but was LONG
  csvImportKeepsALongTextColumnAsString: Cannot convert type 'java.lang.String' to 'LONG'
```

## Test plan

- [x] `mvn -pl integration test -Dtest=Issue6814AnalyzedPropertyTypeRefutationTest` fails 5/8 on `main` (bug reproduced)
- [x] Same run is 8/8 green with the fix
- [x] Full `integration` unit suite: `Tests run: 175, Failures: 0, Errors: 0, Skipped: 9`
- [x] Importer integration tests: `-DskipITs=false -Dit.test='CSVImporterIT,XMLImporterIT,JSONImporterIT,JsonLImporterIT,OrientDBImporterIT,SQLLocalImporterIT'` -> `Tests run: 37, Failures: 0, Errors: 0`
- [x] Confirmed no cross-module impact: `AnalyzedProperty` has no consumers outside `integration`; gremlin's `GraphMLImporterFormat`/`GraphSONImporterFormat` implement `analyze()` without touching it

## Not in scope

The issue also notes that `CSVImporterFormat.loadDocuments()` (`:184`) rethrows the conversion failure raw, with no line number. That is a separate diagnostics improvement and is left untouched here so the fix stays minimal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data type detection during imports, including numeric values encountered after sampling limits are reached.
  * Correctly narrows long numeric values to compatible decimal types.
  * Prevents oversized or excessive samples from consuming unbounded storage.
  * Preserves accurate handling of null and long-text values.

* **Tests**
  * Added coverage for numeric type detection, sampling limits, long values, nulls, and end-to-end CSV import scenarios.

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
@@ -24,6 +24,9 @@ import java.util.HashSet;
 import java.util.Set;
 
 public class AnalyzedProperty {
+  /** A value longer than this is not worth keeping as a sample, but it is still evidence about the type. */
+  private static final int MAX_SAMPLE_LENGTH = 100;
+
   private final String      name;
   private final long        maxValueSampling;
   private final int         index;
@@ -61,44 +64,48 @@ public class AnalyzedProperty {
     return index;
   }
 
+  /**
+   * Sampling and type refutation are two independent concerns and are handled in that order here. Sample collection is
+   * a memory bound and legitimately stops - on an oversized value or once enough distinct values have been seen. Type
+   * refutation is evidence: skipping it for a value, or stopping it for the rest of the column, is how a text column
+   * used to end up declared LONG and blow the import up on the very value the analysis never looked at (issue #6814).
+   */
   public void setLastContent(final String lastContent) {
-    if (!collectingSamples)
+    if (lastContent == null)
       return;
 
-    if (lastContent != null) {
-      if (lastContent.length() > 100) {
-        collectingSamples = false;
-        contents.clear();
-        return;
-      }
+    this.lastContent = lastContent;
 
-      this.lastContent = lastContent;
-      if (contents.size() > maxValueSampling) {
-        collectingSamples = false;
-        contents.clear();
-        return;
-      }
-
-      contents.add(lastContent);
-
-      if (!lastContent.isEmpty()) {
-        if (candidateForInteger) {
-          try {
-            Long.parseLong(lastContent);
-          } catch (final NumberFormatException e) {
-            candidateForInteger = false;
-          }
+    if (!lastContent.isEmpty()) {
+      // EVERY VALUE IS PROBED, NO MATTER ITS LENGTH OR HOW MANY CAME BEFORE IT. THE PROBES COST NOTHING ONCE THE
+      // CANDIDATE HAS ALREADY BEEN REFUTED, WHICH IS THE COMMON CASE FOR A TEXT COLUMN.
+      if (candidateForInteger) {
+        try {
+          Long.parseLong(lastContent);
+        } catch (final NumberFormatException e) {
+          candidateForInteger = false;
         }
+      }
 
-        if (candidateForDecimal) {
-          try {
-            Double.parseDouble(lastContent);
-          } catch (final NumberFormatException e) {
-            candidateForDecimal = false;
-          }
+      if (candidateForDecimal) {
+        try {
+          Double.parseDouble(lastContent);
+        } catch (final NumberFormatException e) {
+          candidateForDecimal = false;
         }
       }
     }
+
+    if (!collectingSamples)
+      return;
+
+    if (lastContent.length() > MAX_SAMPLE_LENGTH || contents.size() > maxValueSampling) {
+      collectingSamples = false;
+      contents.clear();
+      return;
+    }
+
+    contents.add(lastContent);
   }
 
   public Set<String> getContents() {
--- a/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
@@ -77,8 +77,7 @@ public class AnalyzedProperty {
     this.lastContent = lastContent;
 
     if (!lastContent.isEmpty()) {
-      // EVERY VALUE IS PROBED, NO MATTER ITS LENGTH OR HOW MANY CAME BEFORE IT. THE PROBES COST NOTHING ONCE THE
-      // CANDIDATE HAS ALREADY BEEN REFUTED, WHICH IS THE COMMON CASE FOR A TEXT COLUMN.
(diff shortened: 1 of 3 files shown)


robfrank:

PR #6846, opened 2026-08-27
fix(#6812): JSON import no longer saves a skipped record as an anonymous Document

### LINKED ISSUE #6812: JSON import saves a record it just logged as skipped (duplicate @id, missing @cat/@type) as an anonymous Document instead
### Summary
`JSONImporterFormat.createRecord()` uses the same return signal for two opposite outcomes: "there is no mapping, save this as an anonymous document" and "this record is rejected or deduplicated, drop it". `parseRecord()` collapses both, so a record the importer just logged as skipped is saved as an anonymous `Document` instead.

### Where (HEAD `9c9d024db`)
`integration/src/main/java/com/arcadedb/integration/importer/format/JSONImporterFormat.java` - the dedup branch returns an immutable record (`:471`):
```java
// SKIP IT, RETURN THE EXISTENT ONE
return existent.next().asDocument();
```
and the rejection branches return `null` (`:392-394`, `:399-401`, `:418-419`, `:432`, `:435-436`), which is also the "no mapping object" signal (`:385-386`).

`:349-354` collapses them:
```java
if (record instanceof MutableDocument document) { document.save(); return record; }
...
return attributes.map;
```
An existing record from `:471` is an `ImmutableVertex`/`ImmutableDocument`, not a `MutableDocument`, so `attributes.map` is returned. `parseRecordsArray()` (`:220-221`) then does:
```java
if (record instanceof Map) saveAnonymousRecord(database, settings, (Map) record);
```
which calls `getOrCreateDocumentType(settings.documentTypeName)` and saves it.

### Repro
Mapping `{"Users":[{"@cat":"v","@type":"User","@id":"id"}]}` (no `@strategy`), input:
```json
{"Users":[{"id":"1","name":"a"},{"id":"1","name":"b"}]}
```
The second entry is looked up, found, and "skipped" - yet `countType("Document")` is 1, holding `{id:1, name:b}`.

Same for a mapping object missing `@cat` or `@type`, where the WARNING literally says "The following object will be skipped" and `context.errors` is incremented, and for `"@cat":"e"` at top level (`:432`, "IGNORE IN THIS PHASE").

Not covered by `JSONImporterIT`: `importEmployees` sets `"@strategy": "merge"` on every mapping object, so the `:471` branch is never exercised, and `importTwoObjects` exercises only the legitimate `mapping == null` anonymous path.

### Suggested fix
Make the rejected outcomes distinguishable from "no mapping": return a sentinel from `createRecord()`, or have `parseRecord()` fall through to `attributes.map` only when the `mapping` argument was `null`. Either way `parseRecordsArray()`/`parseArray()` must not materialize a rejected record into `settings.documentTypeName`.

### Scope / confidence
High. Verified that `existent.next().asDocument()` does not throw for a vertex (`BaseDocument.asDocument()` returns `this`), so the branch really reaches `:354`, and that the anonymous-save path is intentional for `mapping == null` - which is exactly why the two cases have to be told apart.

robfrank:
Closes #6812

## Summary

`JSONImporterFormat.parseRecord()` returns the raw attribute map (`attributes.map`) for two opposite outcomes, and `parseRecordsArray()` could not tell them apart: it saved *any* `Map` it got back as an anonymous document under `settings.documentTypeName`. The two outcomes are

1. **there was no mapping object for this record** - the legitimate anonymous-document path, and
2. **`createRecord()` refused to build a record, or resolved it to an already existing one** - every one of which the importer had just logged as skipped, and (for the rejection branches) already counted in `context.errors`.

Case 2 covers a mapping object with no `@cat`, no `@type`, an unresolvable/unsupported type, no `@id` value, `"@cat":"e"` at top level ("IGNORE IN THIS PHASE"), and the dedup branch that returns the existing record when `@strategy` is not `merge`. In all of them the record reappeared as a `Document`, so an import that printed *"The following object will be skipped"* still persisted it - and a duplicate `@id` produced a stray `Document` holding the *losing* entry's properties.

The fix makes the mapping object the discriminator at the one place that materializes an anonymous record:

```java
if (record instanceof Map && mappingObject == null)
  saveAnonymousRecord(database, settings, (Map<String, Object>) record);
```

The guard is deliberately at the call site rather than in `parseRecord()`. `parseRecord()` cannot make the distinction on its own: a **nested** object mapped with `"@cat":"e"` *must* keep returning its raw map, because the parent's `applyMappingRules()`/`convertMap()` reads it back out of `attributes.map` to build the edge (this is what `JSONImporterIT.importEmployees` exercises). Only `parseRecordsArray()` - and `load()`'s single-top-level-object path, where the mapping is `null` by construction - ever turns a map into a document, so that is where the two cases have to be told apart. `parseArray()` never saves anything and is unaffected.

## Not changed (deliberate)

`saveAnonymousRecord()` never increments `context.createdDocuments`, so anonymously saved records are absent from the import summary entirely. That is a pre-existing reporting gap, orthogonal to this issue, and left alone here; the regression test asserts on `countType()` rather than on that counter so nothing in it depends on the gap either way.

## Test plan

New `integration/src/test/java/com/arcadedb/integration/importer/Issue6812JsonImportSkippedRecordTest.java` (+ `integration/src/test/resources/importer-6812-duplicate-ids.json`, two entries sharing `id: "1"`). Every case imports the same source and varies only the mapping, so each isolates one skip branch:

- [x] `duplicateIdRecordIsSkippedInsteadOfSavedAsAnonymousDocument` - `@id` with no `@strategy`: 1 `User`, no `Document` type at all
- [x] `mappingObjectWithoutCategoryIsSkippedInsteadOfSavedAsAnonymousDocument` - `errors == 2`, no `Document` and no `User` type
- [x] `mappingObjectWithoutTypeIsSkippedInsteadOfSavedAsAnonymousDocument` - `errors == 2`, no `Document` type
- [x] `unresolvableIdRecordIsSkippedInsteadOfSavedAsAnonymousDocument` - `@id` naming a property no entry carries: `errors == 2`, `User` type created but empty, no `Document` type
- [x] `topLevelEdgeMappingIsSkippedInsteadOfSavedAsAnonymousDocument` - `"@cat":"e"` at top level persists nothing
- [x] `recordsWithoutAnyMappingObjectAreStillSavedAsAnonymousDocuments` - the guard on the opposite side: an empty mapping array still saves both entries as `Document`

Every bug case was run against the unfixed code first and failed (the initial four: `Tests run: 5, Failures: 5` before the fix, `Failures: 0` after; the `unresolvableId` case, added in the second commit, was re-checked the same way by temporarily reverting the guard), so none of them can pass vacuously.

Regression runs (module `integration`, isolated local Maven repo):

- [x] `mvn verify` on `integration` - 172 unit tests, 0 failures
- [x] `mvn verify -DskipITs=false -Dit.test=JSONImporterIT,JsonLImporterIT,XMLImporterIT,CSVImporterIT` - 33 tests, 0 failures. `JSONImporterIT.importEmployees` is the one that would break if the nested `"@cat":"e"` map stopped flowing to the parent; `Issue5968ImporterSkipOnRowErrorTest` (38 tests) covers the surrounding per-record transaction/counter behaviour.


robfrank:

PR #6847, opened 2026-08-27
fix(#6811): stop SourceDiscovery from destroying the user's CSV delimiter option

### LINKED ISSUE #6811: SourceDiscovery overwrites the user's delimiter option with null for every CSV source, so a non-comma CSV cannot be imported via -url / IMPORT DATABASE
### Summary
`SourceDiscovery.analyzeSourceContent()` unconditionally writes `knownDelimiter` into `settings.options`, but leaves it `null` for `EntityType.DATABASE`. Since the user's `-delimiter` / `WITH delimiter = ';'` lands in that same map, it is destroyed for every CSV source, and there is no working alternative for the `-url` / SQL form.

### Where (HEAD `9c9d024db`)
`integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java:235-246`:
```java
case DATABASE:
  // NO SPECIAL SETTINGS
  knownFileType = getFileTypeByExtension(settings.url);
  break;                                   // knownDelimiter stays null
...
if (knownFileType != null) {
  if ("csv".equalsIgnoreCase(knownFileType)) {
    settings.options.put("delimiter", knownDelimiter);    // <- clobbers the user's value with null
```
The `DOCUMENT`/`VERTEX`/`EDGE` cases set `knownDelimiter` from `documentsDelimiter`/`verticesDelimiter`/`edgesDelimiter`; `DATABASE` has no source for it. `ImporterSettings.parseParameter()` stores every parameter in `options` (`:227`), and `CSVImporterFormat.analyze()` (`:705`) plus `createCSVParser()` (`:838-839`) read exactly that key, falling back to `","`.

### Repro
`IMPORT DATABASE file://data.csv WITH delimiter = ';'` on a semicolon-delimited file:
```
id;name;age
1;Alice;30
```
`Importer.load()` routes the main `-url` as `EntityType.DATABASE` (`Importer.java:73-79`), the `.csv` extension short-circuits at `SourceDiscovery:245`, `options["delimiter"]` becomes `null`, and the whole line is parsed as one column: the analyzed schema gets a single property literally named `id;name;age`, and every record gets one property holding `1;Alice;30`. The import reports success.

There is no workaround in this form: `-documentsDelimiter` and friends are never consulted for `EntityType.DATABASE`, and the delimiter auto-detection path is skipped entirely once the `.csv` extension matches (`isSeparator()` at `:372` does not even list `;`).

Confirmed that `ImporterSettings.getValue()` returns the default for a `null` map value, so the user's `;` is genuinely gone rather than merely shadowed.

### Suggested fix
Guard the put - `if (knownDelimiter != null) settings.options.put("delimiter", knownDelimiter);` - and give `case DATABASE` a delimiter source, either the generic `delimiter` option or `documentsDelimiter`.

### Scope / confidence
High. Verified end to end from `ImportDatabaseStatement` (`WITH delimiter = ';'` -> `setSettings` -> `parseParameter` -> `options`) through to the CSV parser construction.

robfrank:
Closes #6811

## Problem

`SourceDiscovery.analyzeSourceContent()` resolved a per-source delimiter and then wrote it into
`settings.options["delimiter"]` unconditionally - including when it had resolved nothing. That map entry is also
where `ImporterSettings.parseParameter()` stores the user's own `-delimiter` / `WITH delimiter = ';'`, and
`CSVImporterFormat` (both `analyze()` and `createCSVParser()`) reads exactly that key, so every CSV source whose
entity type carries no per-source delimiter had the user's value overwritten with `null` and silently fell back to a
comma.

`EntityType.DATABASE` - the entity type the main `-url` is analyzed as - has no per-source delimiter setting at all,
so `IMPORT DATABASE file://data.csv WITH delimiter = ';'` on a semicolon file parsed the whole line as one column:
the analyzed schema got a single property literally named `id;name;age`, every record got one property holding
`1;Alice;30`, and the import reported success. There was no workaround: `-documentsDelimiter` and friends are never
consulted for `DATABASE`, and the auto-detection path is short-circuited by the `.csv` extension (and `isSeparator()`
does not list `;` anyway).

## Fix

- `ImporterSettings` gains a `delimiter` field, populated from the generic `-delimiter` / `WITH delimiter = ...`
  setting. It is held in its own field rather than read back from `options` precisely because `SourceDiscovery`
  overwrites that entry per source, which is what made the user's value collateral damage of the first source
  analyzed.
- `SourceDiscovery` falls back to that generic delimiter whenever the entity type resolves none of its own, which
  gives `EntityType.DATABASE` a delimiter source for the first time while keeping `-documentsDelimiter` /
  `-verticesDelimiter` / `-edgesDelimiter` authoritative for their own sources.
- The write into `options["delimiter"]` is now a resolve rather than a blind put: the value when there is one,
  `remove()` when there is not. `remove()` rather than "leave it alone" matters because every source of a
  multi-source import is analyzed against the same `ImporterSettings`, so a previous source's delimiter must not leak
  into the next source's default. For the no-delimiter-anywhere case the observable behaviour is unchanged
  (`getValue("delimiter", ",")` returned the `,` default for a `null` value just as it does for an absent key).

## Test plan

New `integration/.../Issue6811DelimiterOptionTest` (4 tests, plus fixtures `importer-delimiter-semicolon.csv` and
`importer-vertices-semicolon.csv`):

- [x] `importDatabaseHonorsTheDelimiterOption` - the reported repro via SQL `IMPORT DATABASE ... WITH delimiter = ';'`;
      asserts the properties are `id`/`name`/`age`, that no property named `id;name;age` exists, and that the values
      round-trip. **Fails on `main`.**
- [x] `urlImportHonorsTheDelimiterOption` - the same defect through the CLI/API `-url` + `-delimiter` form.
      **Fails on `main`.**
- [x] `genericDelimiterAppliesToVerticesAndPerSourceOneWins` - the generic `-delimiter` reaches a vertices source that
      has no `-verticesDelimiter`. **Fails on `main`.**
- [x] `aPerSourceDelimiterDoesNotLeakIntoTheNextSource` - a semicolon vertices file followed by a comma edges file in
      the same import; guards the `remove()` half of the fix (passes before and after by construction).
- [x] Verified the three repro tests fail with the two `src/main` files stashed and pass with them restored.
- [x] `mvn -pl integration verify -DexcludedGroups=benchmark,vector,slow` - 170 tests, 0 failures.
- [x] `mvn -pl integration verify -DskipITs=false -Dit.test='CSVImporterIT,SQLLocalImporterIT,XMLImporterIT,JSONImporterIT,JsonLImporterIT,Word2VecImporterIT,GloVeImporterIT,OrientDBImporterIT,Neo4jImporterIT'` - 50 tests, 0 failures.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring a generic CSV delimiter during imports.
  - Delimiter settings now apply consistently across database, URL, vertex, and mixed vertex/edge imports.
- **Bug Fixes**
  - Fixed parsing of semicolon-separated data with generic delimiter settings.
  - Prevented delimiter settings from one source from affecting subsequent imports.
  - Improved precedence and clearing behavior for source-specific delimiter settings.

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/importer/ImporterSettings.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/ImporterSettings.java
@@ -30,6 +30,19 @@ public class ImporterSettings {
   public int     verboseLevel = 2;
   public boolean probeOnly    = false;
 
+  /**
+   * The generic {@code -delimiter} / {@code WITH delimiter = ';'} setting: the fallback used by any CSV source that
+   * carries no per-source delimiter of its own ({@code -documentsDelimiter}, {@code -verticesDelimiter},
+   * {@code -edgesDelimiter}), and the only delimiter source available to the main {@code -url}, which is analyzed as
+   * {@code EntityType.DATABASE} and has no per-source setting at all.
+   * <p>
+   * It is held in its own field rather than read back from {@link #options} because {@code SourceDiscovery} resolves
+   * a delimiter per source and writes the answer into {@code options["delimiter"]}, where {@code CSVImporterFormat}
+   * reads it. Keeping the user's own value only in that same entry made it collateral damage of the first source
+   * analyzed (issue #6811).
+   */
+  public String delimiter;
+
   public String documents;
   public String documentsFileType;
   public String documentsDelimiter;
@@ -185,6 +198,7 @@ public class ImporterSettings {
     case "parsingLimitBytes" -> parsingLimitBytes = FileUtils.getSizeAsNumber(value);
     case "parsingLimitEntries" -> parsingLimitEntries = Long.parseLong(value);
     case "mapping" -> mapping = value;
+    case "delimiter" -> delimiter = value;
     case "probeOnly" -> probeOnly = Boolean.parseBoolean(value);
     case "onRowError" -> {
       if (!"abort".equalsIgnoreCase(value) && !"skip".equalsIgnoreCase(value))
--- a/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
@@ -233,7 +233,7 @@ public class SourceDiscovery {
       break;
 
     case DATABASE:
-      // NO SPECIAL SETTINGS
+      // NO PER-SOURCE SETTINGS: THE GENERIC `delimiter` IS THE ONLY ONE THIS SOURCE CAN HAVE
       knownFileType = getFileTypeByExtension(settings.url);
       break;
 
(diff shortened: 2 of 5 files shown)


robfrank:
Both non-blocking observations from the reviews are deliberately left out of scope, recorded here rather than deferred silently:

1. **Stale `options["delimiter"]` when a later source in the same import resolves to a non-CSV format.** The resolve/`remove()` lives inside the `csv` branch, so a JSON source following a CSV one leaves the previous entry in place. Only `CSVImporterFormat` reads that key today, so it is unobservable; clearing it for every format would widen this fix from "CSV delimiter resolution" to "option-map lifecycle" with no failing case to anchor it.
2. **The content-based auto-detection paths (`analyzeChar` RDF branch, best-separator branch) still do a blind `put`.** Those compute a delimiter from the file content itself rather than from settings, so they cannot clobber a user value the way the extension-based path did, and `isSeparator()` does not list `;` so an explicit `-delimiter` overriding auto-detection is a separate feature question, not this regression.

Neither is a behaviour change on any path this PR touches.

robfrank:

Issue #6474, opened 2026-08-19
import database: server pre-check and deep fetch enforce two different, independently-configured local-URL flags


lvca:
### Summary
`import database <name> <url>` (the HTTP server command) is gated by two independent SSRF checks that read **two different configuration keys**, so enabling the documented opt-out for one does not affect the other. Both currently default to blocking, so this is not an active bypass, but it is the same class of silent-disagreement bug fixed for `restore database` in #6381 - just worse, because the two layers don't even share a key name.

### Where (HEAD `55b9c93e5`, after #6449)
- The server-command pre-check, `PostServerCommandHandler.validateClientRestoreImportUrl` (via `importDatabase()` at `:599`), gates on `GlobalConfiguration.SERVER_RESTORE_IMPORT_ALLOW_LOCAL_URLS`, read from this server's own `ContextConfiguration` (`isRestoreImportLocalUrlsAllowed()`, `:558`). Its own javadoc says this setting is meant to "Allow the 'restore database' and 'import database' server commands to fetch from local-file ('file://') URLs and from private, loopback or link-local network hosts."
- The actual fetch happens deep inside the reflectively-invoked `com.arcadedb.integration.importer.Importer` -> `SourceDiscovery` (`integration/.../SourceDiscovery.java:120,127`) -> `ImportSecurityValidator.openRemoteConnection`, which gates on the **different** key `GlobalConfiguration.SERVER_SECURITY_IMPORT_BLOCK_LOCAL_NETWORKS`, read from the **static, JVM-wide** value only - there is no threading of the pre-check's resolved answer through to the importer (unlike restore, which now threads it via `Restore.setAllowLocalUrls`/`RestoreSettings.allowLocalUrls` after #6449).

### Consequence
An operator who enables the documented `SERVER_RESTORE_IMPORT_ALLOW_LOCAL_URLS` on their server's own configuration (exactly what its javadoc tells them to do for `import database`) gets past the pre-check, but the import then fails downstream with a `SecurityException` referencing an entirely different, undocumented-for-this-purpose setting (`SERVER_SECURITY_IMPORT_BLOCK_LOCAL_NETWORKS`). Confusing, fail-closed, not itself an SSRF hole - but exactly the kind of divergence that becomes a real hole the next time only one side is touched.

### Suggested fix
[...]

PR #6850, opened 2026-08-27
fix(#6810): importing any ZIP source silently yields 0 records

### LINKED ISSUE #6810: Importing any ZIP source silently yields 0 records: the reset callback positions the old (closed) ZipInputStream, not the new one
### Summary
`SourceDiscovery`'s reset callbacks call `getNextEntry()` on the **old** `ZipInputStream` local variable instead of the freshly created one. Since the reset path runs on every import, any ZIP source imports as 0 records, and the underlying `IOException("Stream closed")` is swallowed by `Source.reset()`, so the import reports success.

### Where (HEAD `9c9d024db`)
`integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java:188-197` (local file):
```java
source.inputStream.close();
if (source.inputStream instanceof GZIPInputStream)
  source.inputStream = new GZIPInputStream(new FileInputStream(file), 2048);
else if (source.inputStream instanceof ZipInputStream stream) {
  source.inputStream = new ZipInputStream(new FileInputStream(file));
  stream.getNextEntry();                 // <- the OLD stream, closed two lines above
} else
```
`SourceDiscovery.java:143-153` (remote URL) has the identical mistake; there the old stream is not closed first, so it silently advances the wrong stream and the new one is still left unpositioned.

`integration/src/main/java/com/arcadedb/integration/importer/Source.java:46-53` swallows the failure:
```java
public void reset() throws IOException {
  if (resetCallback != null)
    try {
      resetCallback.call(this);
    } catch (final Exception e) {
      LogManager.instance().log(this, Level.SEVERE, "Error on resetting source %s", e, this);
    }
}
```
So `ZipInputStream.getNextEntry()` hits `ensureOpen()` and throws, the callback wraps it in `ImportException`, and `reset()` logs it at SEVERE and returns. The new `ZipInputStream` is left with no current entry, and `ZipInputStream.read()` returns `-1` while `entry == null`.

Secondary defect on the same lines: the rebuilt stream is never seeked to the `:::<resource>` entry the user asked for, so even the first entry is the wrong one for a multi-entry archive.

### Repro
The reset path is unavoidable: `Importer.loadFromSource()` (`Importer.java:133-135`) always does `getSource(); new Parser(source, 0); parser.reset();`, and `SourceDiscovery.getSchema()` (`:90`) plus `CSVImporterFormat.analyze()` (`:703`) each call `parser.reset()` too.

```bash
bin/arcadedb-importer.sh -url file://data.csv.zip -database ./databases/x
```
logs `Error on resetting source ...` at SEVERE and then reports a completed import with `- Total documents: 0` and no type created. Same for `data.zip:::nodes.csv` and for `IMPORT DATABASE https://.../data.zip`.

ZIP is a supported input, not dead code: `getSourceFromContent()` (`:482-500`) hands back a `ZipInputStream`-backed `Source` for any input whose first bytes are a zip LOC header, and `getFormatFromExtension()` (`:526`) explicitly strips `.zip`. There is no test with a `.zip` fixture (only `.gz`, whose branch is correct), which is why this has stayed invisible.

### Suggested fix
Call `getNextEntry()` on the newly created stream and, when `resource != null`, loop `getNextEntry()` until the entry name matches, mirroring `getSourceFromContent()` (`:486-497`). Separately, make `Source.reset()` rethrow instead of only logging, so a broken reset can never be mistaken for an empty input.

### Scope / confidence
High. Verified at HEAD on both callbacks, including that `ZipInputStream.read()` with no current entry returns `-1` rather than throwing, which is what turns this into a silent empty import.

robfrank:
Closes #6810

## Summary

`SourceDiscovery`'s two reset callbacks rebuilt the `ZipInputStream` from a freshly re-opened raw stream but then called `getNextEntry()` on the **old** local variable rather than on the new stream. On the local-file path that old stream had been closed two lines above, so the call threw `IOException("Stream closed")`; on the remote path it silently advanced a stream nobody would read again. Either way the *new* `ZipInputStream` was left sitting on no entry, and `ZipInputStream.read()` returns `-1` while no entry is current - so the import read an empty input. `Source.reset()` logged the exception at SEVERE and returned, which is what turned a hard failure into a "completed" import of 0 records. Because `Importer.loadFromSource()`, `SourceDiscovery.getSchema()` and `CSVImporterFormat.analyze()` all call `reset()`, the path was unavoidable: **every** ZIP source imported as 0 records. A second defect on the same lines meant the rebuilt stream was never seeked back to the `:::<resource>` entry the user asked for, so a multi-entry archive resumed on the wrong entry.

The fix extracts the entry-seeking logic that `getSourceFromContent()` already had into `positionOnZipEntry(zip, resource)` and calls it from a new `reopenZip()` helper that both reset callbacks now use, so a rebuilt stream is positioned on exactly the entry the original one was reading. The remote callback also closes the old stream before disconnecting (it previously leaked it), and `Source.reset()` no longer swallows a callback failure - a broken reset can no longer be mistaken for an empty input.

### Changed

- `integration/.../importer/SourceDiscovery.java`
  - new `positionOnZipEntry(ZipInputStream, String)`: seeks to the named entry, or the first one when no `:::resource` was given; returns `null` when the content is not a zip at all (the "not a zip archive" signal `getSourceFromContent()` needs)
  - new `reopenZip(InputStream, String)`: builds the replacement stream, positions it, and closes the raw stream if the archive turns out unreadable so it cannot leak
  - both reset callbacks (local file and remote URL) now call `reopenZip` instead of poking the dead stream; the remote one also closes the previous stream first
  - `getSourceFromContent()` refactored onto the same helper - one implementation of the entry search instead of two
- `integration/.../importer/Source.java` - `reset()` propagates the callback failure instead of logging it at SEVERE and returning

## Test plan

New `integration/src/test/java/com/arcadedb/integration/importer/Issue6810ZipSourceResetTest.java` (8 tests). All 6 zip-related ones fail on `main` and pass with the fix; the two non-zip ones are non-regression guards that pass on both.

- [x] `localZipSourceResetsToTheStartOfTheEntry` - single-entry `.csv.zip` read through `Parser`+`reset()` exactly as `Importer.loadFromSource()` does; was `""`, now the full CSV
- [x] `localZipSourceResetsToTheRequestedEntryOfAMultiEntryArchive` - `file://multi.zip:::importer-vertices.csv`, proves the reset re-seeks past the first entry
- [x] `remoteZipSourceResetsToTheStartOfTheEntry` - covers the second (remote-URL) callback, serving the archive from a loopback `com.sun.net.httpserver.HttpServer` with `allowLocalUrls=true` so the SSRF guard from #6474 is not in the way
- [x] `localGzipSourceStillResetsToTheStartOfTheStream` / `localPlainSourceStillResetsToTheStartOfTheStream` - the two branches that were already correct stay correct
- [x] `importDatabaseFromZipLoadsEveryRow` / `importDatabaseFromMultiEntryZipLoadsTheRequestedEntry` - end-to-end `IMPORT DATABASE file://...zip`; previously died with "Type with name 'Document' was not found" (nothing was created), now imports all 6 rows
- [x] `resetPropagatesACallbackFailureInsteadOfSwallowingIt` - a throwing reset callback now reaches the caller

Regression runs (all green, `-Dmaven.repo.local` isolated):

- [x] `mvn -o -pl integration verify -DskipITs=false -DskipTests=false` - 174 unit + 118 IT, 0 failures (includes `CSVImporterIT`, `JsonLImporterIT`, `Neo4jImporterIT`, `OrientDBImporterIT`, `XMLImporterIT`, `SQLLocalImporterIT`, `SQLRemoteImporterIT`, `Word2Vec*`, `GloVeImporterIT` - i.e. every format that goes through `Source.reset()`)
- [x] `mvn -o -pl gremlin-it verify -DskipITs=false -Dit.test='GraphMLImporterIT,GraphSONImporterIT,GraphMLExporterIT,GraphSONExporterIT'` - 360 unit + 10 IT, 0 failures (the `.tgz`-backed GraphML/GraphSON importers also reset the source)

Not run locally: the `server` module ITs (`ServerImportDatabaseIT`, which imports `classpath://orientdb-export-small.gz`) - port 2480 was already held by another process on this machine, which makes that module fail spuriously. That path is the gzip branch, covered by `localGzipSourceStillResetsToTheStartOfTheStream` and by `SQLLocalImporterIT`; CI will run it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ZIP archive imports after source resets, including local, multi-entry, and remote archives.
  * Ensured requested ZIP entries remain selected when streams are reopened.
  * Reset failures now surface as import errors instead of being silently logged.
  * Preserved reset behavior for GZIP-compressed and plain files.
  * Improved cleanup when source setup or reset fails, preventing connection and stream leaks.

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/importer/Source.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/Source.java
@@ -43,13 +43,15 @@ public class Source {
     this.closeCallback = closeCallback;
   }
 
+  /**
+   * Rebuilds {@link #inputStream} from the beginning of the source.
+   * <p>
+   * A failure here is propagated on purpose: swallowing it leaves the caller reading a stream that is closed or not
+   * positioned on any entry, which reads as an empty - but successful - import instead of as an error (issue #6810).
+   */
   public void reset() throws IOException {
     if (resetCallback != null)
-      try {
-        resetCallback.call(this);
-      } catch (final Exception e) {
-        LogManager.instance().log(this, Level.SEVERE, "Error on resetting source %s", e, this);
-      }
+      resetCallback.call(this);
   }
 
   public void close() {
--- a/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
@@ -148,7 +148,7 @@ public class SourceDiscovery {
           currentConnection.get().getContentLengthLong(), resource,
           source -> {
             try {
-              source.inputStream.close();
+              closeBestEffort(source.inputStream);
               currentConnection.get().disconnect();
 
               final HttpURLConnection reopened = ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks);
@@ -210,7 +210,7 @@ public class SourceDiscovery {
     try {
       return getSourceFromContent(fis, file.length(), resource, source -> {
         try {
-          source.inputStream.close();
+          closeBestEffort(source.inputStream);
           if (source.inputStream instanceof GZIPInputStream)
             source.inputStream = new GZIPInputStream(new FileInputStream(file), 2048);
           else if (source.inputStream instanceof ZipInputStream)
@@ -589,6 +589,19 @@ public class SourceDiscovery {
       closeable.close();
     } catch (final IOException e) {
       pending.addSuppressed(e);
+    }
+  }
+
+  /**
+   * Closes the stream a reset is about to replace. A stream that refuses to close must not abort the reset: the
+   * replacement is what the caller will read, and on the remote path the connection behind the old stream still has to
+   * be disconnected afterwards.
+   */
+  private void closeBestEffort(final Closeable closeable) {
+    try {
+      closeable.close();
+    } catch (final IOException e) {
+      LogManager.instance().log(this, Level.FINE, "Error on closing the previous stream while resetting source '%s'", e, url);
     }
   }
 
--- a/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
@@ -48,6 +48,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.ZipEntry;
@@ -137,29 +138,35 @@ public class SourceDiscovery {
     final boolean blockLocalNetworks = allowLocalUrls != null ?
         !allowLocalUrls : GlobalConfiguration.SERVER_SECURITY_IMPORT_BLOCK_LOCAL_NETWORKS.getValueAsBoolean();
 
-    final HttpURLConnection connection = ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks);
+    // EVERY RESET REPLACES THE CONNECTION, SO THE CLOSE CALLBACK HAS TO TEAR DOWN THE ONE THE STREAM CURRENTLY COMES
+    // FROM, NOT THE FIRST ONE EVER OPENED
+    final AtomicReference<HttpURLConnection> currentConnection = new AtomicReference<>(
+        ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks));
 
     try {
-      return getSourceFromContent(new BufferedInputStream(connection.getInputStream()), connection.getContentLengthLong(), resource,
+      return getSourceFromContent(new BufferedInputStream(currentConnection.get().getInputStream()),
+          currentConnection.get().getContentLengthLong(), resource,
           source -> {
             try {
               source.inputStream.close();
-              connection.disconnect();
+              currentConnection.get().disconnect();
 
-              final HttpURLConnection connection1 = ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks);
+              final HttpURLConnection reopened = ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks);
               try {
                 if (source.inputStream instanceof GZIPInputStream)
-                  source.inputStream = new GZIPInputStream(connection1.getInputStream(), 2048);
+                  source.inputStream = new GZIPInputStream(reopened.getInputStream(), 2048);
                 else if (source.inputStream instanceof ZipInputStream)
                   // THE NEW STREAM MUST BE POSITIONED ON THE ENTRY TO READ, THE OLD ONE IS GONE (issue #6810)
-                  source.inputStream = reopenZip(new BufferedInputStream(connection1.getInputStream()), resource);
+                  source.inputStream = reopenZip(new BufferedInputStream(reopened.getInputStream()), resource);
                 else
-                  source.inputStream = new BufferedInputStream(connection1.getInputStream());
+                  source.inputStream = new BufferedInputStream(reopened.getInputStream());
               } catch (final IOException | RuntimeException e) {
                 // THE REPLACEMENT STREAM WAS NEVER INSTALLED ON THE SOURCE, SO NOTHING ELSE WOULD EVER DISCONNECT IT
-                connection1.disconnect();
+                reopened.disconnect();
                 throw e;
               }
+
+              currentConnection.set(reopened);
             } catch (final ImportException e) {
               throw e;
             } catch (final Exception e) {
@@ -167,12 +174,12 @@ public class SourceDiscovery {
             }
             return null;
           }, () -> {
-            connection.disconnect();
+            currentConnection.get().disconnect();
             return null;
           });
     } catch (final IOException | RuntimeException e) {
       // THE SOURCE THAT WOULD HAVE OWNED (AND EVENTUALLY DISCONNECTED) THE CONNECTION WAS NEVER BUILT
-      connection.disconnect();
+      currentConnection.get().disconnect();
       throw e;
     }
   }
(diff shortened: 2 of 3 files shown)


robfrank:
Review round 1 (@claude) addressed in cc62144:

**Applied**
- **`connection1` leak on a failed remote reset** - confirmed, and confirmed it is new to this PR (the old code reassigned `source.inputStream` before the throwing call, so the connection was never orphaned). `connection1.disconnect()` now runs when the rebuild throws.
- **Same leak on the *initial* open** (flagged as a pre-existing FYI) - fixed in the same pass rather than deferred, since it is the identical shape one frame up and the code was already being touched: `getSourceFromURL` disconnects and `getSourceFromFile` closes `fis` when `getSourceFromContent` throws (e.g. `Resource 'x' not found`). No `Source` exists on that path, so nothing else would ever have released them.
- **Exception-wrapping asymmetry** - the local callback now has the same `catch (ImportException) rethrow / catch (Exception) wrap` shape as the remote one.
- **`zip.close()` masking the pending exception** - new `closeSuppressing(Closeable, Throwable)` helper uses `addSuppressed`; used by `reopenZip` and by the new `getSourceFromFile` cleanup.
- **Test for the new failure path** - `remoteZipResetFailsLoudlyWhenTheArchiveNoLongerHoldsTheEntry` serves a valid zip on the first request and a non-zip body on the reset's re-fetch, asserting the `ImportException` reaches the caller instead of the import quietly reading nothing.

**Not applied**
- **`Source.reset()`'s `throws IOException` is now vestigial** - correct as an observation, but `reset()` is public API and `Parser.reset()` propagates it up through every importer format; narrowing the signature would be a gratuitous source-incompatible change for callers that catch `IOException` around it alone, and it keeps the door open for a future callback that does throw checked I/O. Left as is deliberately.

Re-verified after the change: `integration` 175 unit + 118 IT green, `gremlin-it` 360 unit + 10 IT green.

robfrank:
Review round 2 (@claude) addressed in f082dc6:

**Applied**
- **`closeCallback` captured the first connection, not the current one** - real, and worth closing since a reset genuinely replaces the connection: after any reset, `Source.close()` was disconnecting a stale `HttpURLConnection` and never the live one. The remote source now tracks the live connection in an `AtomicReference` that the reset callback updates *after* the replacement stream is successfully installed, so `closeCallback`, the reset itself, and the never-built-a-Source cleanup path all target the same object. (Agreed on the analysis that it was mostly benign in practice - `HttpURLConnection` releases the connection when its stream is closed - but the object the callback names should still be the one it means.)

**Not applied**
- **All-caps inline comments read as loud** - checked before changing: uppercase `// LIKE THIS` comments are the established convention throughout this module, including in the very method being edited (`// ZIPPED FILE`, `// SEARCH FOR THE RIGHT ENTRY`, `// NOT A ZIP ARCHIVE`) and across its neighbours (`ImportSecurityValidator`, `Neo4jImporter`, `OrientDBImporter`, `ImporterSettings`, `AbstractImporter`). Lower-casing only these would make the new lines the odd ones out, so CLAUDE.md's "adhere to the existing code" wins here.

Re-verified after the change: `integration` 175 unit + 118 IT green.

robfrank:
Review round 3 (@claude) addressed in e5f2776:

**Applied**
- **`source.inputStream.close()` throwing skipped the `disconnect()`** - real gap, and it defeats the point of the round-2 fix: a throwing close left the old connection undisconnected *and* `currentConnection` stale. Both reset callbacks now close the outgoing stream through a new `closeBestEffort()` (logs at `FINE`), so the replacement stream is always built and the old connection is always torn down. Applied to the local callback too, for symmetry: a stream that will not close is no reason to fail a reset whose whole job is to replace it.

**Not applied**
- **`IllegalArgumentException` (resource not found) vs `ImportException` (not a zip)** - deliberate, not an oversight. `IllegalArgumentException` for a missing `:::resource` is the long-standing behaviour of `getSourceFromContent()`; `positionOnZipEntry` is a pure extraction of those exact lines, and changing the type would silently alter what callers of the *initial open* path have been catching since well before this bug. `ImportException` is used only for the condition that genuinely did not exist before (a rebuilt archive that holds no entry at all). Unifying them is a defensible cleanup, but it is a behaviour change to an unrelated path and does not belong in a regression fix - happy to file it as a follow-up if you want it.

Re-verified after the change: `integration` 175 unit + 118 IT green.

robfrank:
Review round 4 (@claude): no actionable items - working tree unchanged, nothing applied. The single nit raised (`Source.reset()`'s now-vestigial `throws IOException`) is the same one declined in round 1, for the same reason: it is public API that `Parser.reset()` propagates through every importer format, so narrowing it would be a source-incompatible change for zero behavioural gain. Four review cycles, all feedback either applied or answered.

## CI on e5f2776

| lane | result |
|---|---|
| `integration-tests` | **pass** (this is the lane that exercises the changed code) |
| `build-and-package`, `builder-tests`, `lint`, `client-smoke-test`, `opencypher-tck-tests` | pass |
| `java-e2e`, `js-e2e`, `go-e2e`, `python-e2e`, `csharp-e2e`, `studio-e2e` | pass |
| `unit-tests` | fail - **pre-existing flake, not this PR** |
| `vector-unit-tests` | fail - **already red on `main`, not this PR** |

Detail on the two red lanes, since neither touches the changed code:

- **`unit-tests`**: the only failing test in the whole report is `com.arcadedb.server.http.ws.Issue6762WebSocketEventBusTest.shutdownStillJoinsAfterASeparateStopSignal` - a thread-lifecycle race in the `server` module's WebSocket event bus (`shutdown()` racing `run()`'s unwind), asserting `watcher.isAlive()` is false immediately after `shutdown()`. It has no path to `com.arcadedb.integration.importer`. The lane is intermittently red on `main` independently of this branch (3 of the last 8 `main` runs).
- **`vector-unit-tests`**: red on **each of the last 6 `main` runs**, i.e. broken before this branch existed. It runs the `@Tag("vector")` engine tests, which this PR does not touch.

`ha-integration-tests` and `slow-unit-tests` were still queued at the time of writing; neither covers the importer either.

robfrank:

PR #6851, opened 2026-08-27
fix(#6840,#6841,#6842): make the shipped Kubernetes and Docker examples work

### LINKED ISSUE #6840: The shipped k8s StatefulSet example cannot work: ${VAR} is never expanded (root password becomes a literal), HA is wired on 2424 while Raft binds 2434, and it sets a setting that no longer exists
### Summary
The shipped example StatefulSet cannot work as written. It uses `${VAR}` in `command:`, which Kubernetes never expands, so the root password becomes the literal string `${rootPassword}` and every pod claims the peer name `${HOSTNAME}`. It also wires HA on port 2424 while Raft gRPC binds 2434 by default, so the cluster can never form a quorum. And it passes a setting that no longer exists.

### Where (HEAD `9c9d024db`)
`package/src/main/config/arcadedb-statefulset.yaml:105-121`:
```yaml
command:
  - "bin/server.sh"
  - "-Darcadedb.dumpConfigAtStartup=true"
  - "-Darcadedb.server.name=${HOSTNAME}"
  - "-Darcadedb.server.rootPassword=${rootPassword}"
  ...
  - "-Darcadedb.ha.replicationIncomingHost=0.0.0.0"
  - "-Darcadedb.ha.serverList=arcadedb-0.arcadedb.default.svc.cluster.local:2424"
```
plus a headless-Service port named `rpc` on 2424 (`:53-55`) and `containerPort: 2424` (`:130-131`).

**(a) `${VAR}` is never expanded.** Kubernetes expands only the `$(VAR)` form in `command`/`args`, and `command:` bypasses any shell so there is no POSIX expansion either. `server.sh` forwards its arguments unmodified (`$ARGS "$@" com.arcadedb.server.ArcadeDBServer`, `:121`), so the JVM literally receives `-Darcadedb.server.rootPassword=${rootPassword}`. The `rootPassword` env var wired from `secretKeyRef: arcadedb-credentials` (`:99-104`) is never read, even though `optional: false` forces the operator to create the Secret. Also note `HOSTNAME` is not injected by the kubelet at all, so even `$(HOSTNAME)` would not resolve: `metadata.name` via a `fieldRef` is the correct source.

`arcadedb.server.name` is a plain string setting (`GlobalConfiguration.java:1260`, default `ArcadeDB_0`) read once into `ArcadeDBServer.serverName` (`:187`) and handed to Raft as the peer identity (`RaftHAServer.java:292`). `assignHostAddress()` (`:1371-1392`) derives the *host address* from `HOSTNAME` when `HA_K8S` is true, but never touches `SERVER_NAME`, so there is no k8s fallback.

**(b) Wrong Raft port.** `arcadedb.ha.raftPort` defaults to **2434** (`GlobalConfiguration.java:1628-1630`) and is what the node actually binds (`RaftHAServer.java:265`, `RaftPropertiesBuilder.java:86`). The manifest never sets it, so every pod listens on 2434 while the Service, the container port and the peer list all point at 2424. `ConfigValidationTest.java:21` asserts the 2434 default, so 2424 is the wrong number rather than the test being stale. The Dockerfile matches the same stale world: `EXPOSE 2424` labelled "ArcadeDB Binary Protocol (replication)" (`package/src/main/docker/Dockerfile:62`), with 2434 never exposed.

**(c) A dead setting.** `grep -rn "replicationIncomingHost\|REPLICATION_INCOMING"` over the whole tree returns exactly one hit: this YAML line. The setting no longer exists, so it is a silent no-op.

### Repro
`kubectl apply -f package/src/main/config/arcadedb-statefulset.yaml` with the `arcadedb-credentials` Secret created, 3 replicas:
- all three pods come up with the root password set to the 15-character literal `${rootPassword}`, which is public in this repository, while the operator believes the Secret is in effect;
- all three claim the same Raft peer name `${HOSTNAME}`;
- the peer list resolves to `...:2424` where nothing listens, so there is no quorum, no leader, and with readiness gated on HA the pods never become Ready.

### Suggested fix
- `-Darcadedb.server.name=$(POD_NAME)` with a `POD_NAME` `fieldRef` on `metadata.name`, and `-Darcadedb.server.rootPassword=$(rootPassword)`.
- Replace every 2424 in the manifest (serverList, Service `rpc` port, `containerPort`) with 2434, and change the Dockerfile's `EXPOSE 2424` to `EXPOSE 2434` with the comment updated to "Raft gRPC (replication)".
- Delete the `-Darcadedb.ha.replicationIncomingHost` line.

A rendered-manifest smoke test in CI (apply to kind, assert 3 Ready pods and a leader) would keep the example honest, since it is the first thing an evaluator runs on Kubernetes.

### Scope / confidence
High on all three. Verified that `command:` is not shell-wrapped, that `server.sh` quotes `"$@"` so nothing re-expands downstream, that the only non-test occurrences of 2424 in the repo are this YAML, the Dockerfile `EXPOSE`, `docker-compose.yml:25` and `README.md:157`, and that `HA_RAFT_PORT`'s own description ("Used as the default port when HA_SERVER_LIST entries do not specify an explicit port") rules out the peer-list entry rebinding the listener.

### LINKED ISSUE #6841: Docker image pins -Xms2G -Xmx2G, ignoring the container memory limit: docker run -m 512m arcadedata/arcadedb dies at startup
### Summary
The Docker image pins `-Xms2G -Xmx2G`, overriding the JVM's container-aware ergonomics. Any container with a memory limit below roughly 2.2 GB dies before the server starts. The repo contradicts this default twice: `server.sh` warns against pinning `Xms=Xmx` to a large value, and the shipped StatefulSet requests 512Mi for this very image.

### Where (HEAD `9c9d024db`)
`package/src/main/docker/Dockerfile:24`:
```dockerfile
ENV ARCADEDB_OPTS_MEMORY="-Xms2G -Xmx2G"
```
`server.sh:114` passes it verbatim onto the java command line, and `server.sh:78-80` defaults it to the empty string only when **unset**, so the image ENV always wins. `CMD` is `["./bin/server.sh"]` (`Dockerfile:83`); there is no entrypoint that recomputes the heap.

The image also selects ZGC (`Dockerfile:22`), which commits `-Xms` eagerly, so a pinned 2 GB initial heap is reserved up front.

Contradicted in-tree by:
- the advice block at `server.sh:70-77`, which explicitly warns against pinning `Xms=Xmx` to a large value;
- `package/src/main/config/arcadedb-statefulset.yaml:124-126`, which requests `memory: "512Mi"`.

### Repro
```bash
docker run --rm -m 512m -p 2480:2480 arcadedata/arcadedb:latest
```
The JVM either fails to reserve the heap or the container is OOM-killed (exit 137) before the HTTP listener binds. Same on Kubernetes for any pod with `resources.limits.memory` under ~2.2Gi, which is the natural limit to pick given the example manifest's own 512Mi request.

### Suggested fix
Drop the `ARCADEDB_OPTS_MEMORY` ENV from the Dockerfile (the JVM's default `MaxRAMPercentage=25` is cgroup-aware), or replace it with:
```dockerfile
ENV ARCADEDB_OPTS_MEMORY="-XX:MaxRAMPercentage=75"
```
so the heap scales with whatever limit the operator sets, and add a consistent `resources.limits.memory` to the example StatefulSet.

### Scope / confidence
High. Verified there is no entrypoint recomputation and that the ENV always wins over the script default.

### LINKED ISSUE #6842: Docker quickstarts overwrite the image JAVA_OPTS (losing ZGC) instead of using ARCADEDB_SETTINGS, and docker-compose.yml declares a volume it never mounts
Two quickstart-surface defects. Both are in the paths a new user runs first, which is what makes them worth fixing despite being small.

### 1. Every documented quickstart passes settings via `JAVA_OPTS`, discarding the image's GC configuration

`package/src/main/docker/Dockerfile:22` sets:
```dockerfile
ENV JAVA_OPTS="-XX:+UseZGC -XX:+ZGenerational"
```
Both official quickstarts then **replace** that same variable to pass database settings:
- `README.md:158`: `-e JAVA_OPTS="-Darcadedb.server.rootPassword=... -Darcadedb.server.defaultDatabases=..."`
- `docker-compose.yml:27-30`: `JAVA_OPTS: > -Darcadedb.server.rootPassword=...`

Docker env vars are replaced, not appended, so the GC selection the image deliberately makes is gone in exactly the configurations users are told to run. Meanwhile `ARCADEDB_SETTINGS` - the variable every launcher script already expands for precisely this purpose (`server.sh:119`, `console.sh:71`, `restore.sh:70`, `mcp-stdio.sh:74`, `server.bat:122`, `console.bat:80`) - is mentioned in no `.md` file and not in the Dockerfile.

Repro: `docker run -e JAVA_OPTS="-Darcadedb.server.rootPassword=playwithdata" arcadedata/arcadedb:latest`, then `jcmd 1 VM.flags` shows G1, not ZGC. Two users following the two official quickstarts get a different garbage collector than an operator who starts the same image with no env at all, which makes any GC-related bug report ambiguous.

*Fix:* use `ARCADEDB_SETTINGS` for the `-Darcadedb.*` settings in `README.md` and `docker-compose.yml`, and document it. Move the GC flags out of the user-facing `JAVA_OPTS` into a dedicated ENV the quickstarts do not overwrite. (Verified `server.sh` expands `$JAVA_OPTS` and `$ARCADEDB_SETTINGS` as two separate words, `:113` and `:119`, so `ARCADEDB_SETTINGS` is a fully working alternative; there is no `JAVA_OPTS_APPEND`-style merge anywhere.)

### 2. `docker-compose.yml` declares a named volume it never mounts

The file defines `volumes: arcade_db_vol:` (`:32-33`) but the `arcade` service has no `volumes:` key at all (`:21-31`). The image's `VOLUME /home/arcadedb/databases` (`Dockerfile:48`) therefore resolves to an **anonymous** volume, and the declared `arcade_db_vol` is created and left empty.

Repro: `docker compose up`, wait for the OpenBeer import, `docker compose down`, `docker compose up` again - the database is gone (the anonymous volume is orphaned, not reused) while `docker volume ls` shows an empty `*_arcade_db_vol`. The file implies persistence it does not provide.

Two smaller things in the same file: there is no `healthcheck:`, even though the server exposes a purpose-built unauthenticated readiness endpoint (`/api/v1/ready`, registered at `server/src/main/java/com/arcadedb/server/http/HttpServer.java:237`), so with `restart: always` (`:31`) a crash-looping container still reports as "running"; and `version: "3.9"` (`:19`) is obsolete under Compose v2 and prints a deprecation warning on every `up`.

*Fix:*
```yaml
volumes: ["arcade_db_vol:/home/arcadedb/databases"]
healthcheck:
  test: ["CMD-SHELL", "wget -qO- http://localhost:2480/api/v1/ready || exit 1"]
  interval: 10s
  timeout: 3s
  retries: 12
```
and drop the obsolete `version:` key. (`wget` is present in the BusyBox base image, which is why it is used rather than curl.)

lvca:
Fixes #6840
Fixes #6841
Fixes #6842

The shipped deployment examples are the first thing an evaluator runs on Kubernetes or Docker, and they are the only files in the tree that no test ever touched. Each had drifted away from the code it launches.

## #6840 - the StatefulSet cannot form a cluster

- **`${VAR}` is never expanded.** Kubernetes expands only the `$(VAR)` form, and `command:` is not shell-wrapped, so the JVM literally received `-Darcadedb.server.rootPassword=${rootPassword}` and every pod claimed the peer name `${HOSTNAME}`. The pod name is now injected as a `POD_NAME` fieldRef on `metadata.name` (the kubelet does not put `HOSTNAME` in the pod spec, so `$(HOSTNAME)` would not have resolved either).
- **The root password no longer travels on the command line.** Rather than the suggested `$(rootPassword)`, the Secret is mounted as a file and read through the existing `arcadedb.server.rootPasswordPath` setting. An argument is readable through `/proc` by anything in the pod; a mounted secret is not, and rotating it does not mean re-rendering the manifest.
- **Wrong Raft port.** The peer list, the headless Service port and the container port all pointed at 2424 while the node binds `HA_RAFT_PORT` = 2434. All three now use 2434, and the peer entry carries the HTTP port (`arcadedb-0:2434:2480`) that replicas need for leader forwarding. The short pod name is expanded by the configured k8s DNS suffix, so the entry stays namespace-agnostic. The stale `EXPOSE 2424` is fixed in the JVM Dockerfile, both native Dockerfiles, and the custom-package builder.
- **A dead setting.** `arcadedb.ha.replicationIncomingHost` no longer exists anywhere; dropped.

Two further things the manifest got wrong that the issue did not cover:

- It created a `Namespace: arcadedb` that **nothing referenced**, so everything landed in `default`. All three resources now declare that namespace, and the k8s DNS suffix spells the same service/namespace pair.
- It had **no probes at all**, so a pod reported Ready before it had joined the Raft group. It now sets `readinessRequiresHA=true` with a readiness probe on `/api/v1/ready`, plus `publishNotReadyAddresses: true` on the headless service - which that readiness strictly requires, since a headless service otherwise publishes no DNS record for a not-Ready pod and a starting node could never be dialled by its peers, so it could never become Ready. Liveness deliberately uses a TCP check rather than `/api/v1/ready`, which would restart a follower that is merely catching up.

## #6841 - the image pinned the heap above the container limit

Verified directly: `-Xms2G -Xmx2G` under `docker run -m 512m` gives `Failed to allocate initial Java heap (2048M)`; `-XX:MaxRAMPercentage=75` under the same limit gives a 384MB heap and starts. The image now uses the percentage form, and the example StatefulSet sets the `limits.memory` the heap is sized against.

## #6842 - quickstarts discarded the image's JVM configuration, compose lost its data

- Both quickstarts passed `-Darcadedb.*` through `JAVA_OPTS`, and Docker *replaces* an env var rather than appending, so the image's GC selection vanished in exactly the configurations users are told to run. Settings now go through `ARCADEDB_SETTINGS` (which every launcher script already expands), and the GC flags moved into a dedicated `ARCADEDB_OPTS_GC` that no quickstart overwrites - belt and braces, so a user who legitimately sets `JAVA_OPTS` also keeps ZGC. The new variable is honoured by every launcher (`server`, `console`, `restore`, `mcp-stdio`, plus the `.bat` pair) so the image behaves the same whichever one is invoked. `README.md` documents `ARCADEDB_SETTINGS` for the first time; the release-contract workflow and the native-image doc were using the same wrong variable and were updated too.
- `docker-compose.yml` declared `arcade_db_vol` and never mounted it, so `VOLUME /home/arcadedb/databases` resolved to a throwaway anonymous volume and the imported database was gone after `down`/`up`. Now mounted, plus a healthcheck on `/api/v1/ready` (a crash-looping container under `restart: always` otherwise still reports as running) and the obsolete `version:` key removed.
- `server.bat` unconditionally clobbered a user-supplied `ARCADEDB_OPTS_MEMORY`, unlike `server.sh`; made conditional.

## Also

`arcadedb.server.rootPasswordPath` was undocumented and completely untested before this change, which is a poor thing to make the recommended Kubernetes mechanism. It now strips the trailing newline that every ordinary way of producing such a file appends (which would otherwise store a root password nobody can type) and rejects an empty file, with `RootPasswordFromFileTest` covering both plus the no-newline and unreadable-file paths.

## Verification

- `PackagingManifestsTest` (11 checks) pins the manifests against the code that reads them: only-`$(VAR)` expansion with every reference declared in `env`, every `-Darcadedb.*` key resolving in `GlobalConfiguration`, the Raft/HTTP ports matching the defaults the server binds, the DNS suffix matching the manifest's own service and namespace, the password path matching a mounted volume, no heap pinned in the image, the GC variable defined *and* expanded by the launchers, settings not routed through `JAVA_OPTS`, and every declared compose volume mounted. **It fails all eleven on the pre-fix manifests** and passes after.
- `RootPasswordFromFileTest`: 6 tests. `ServerSecurityIT`: 6 tests, still green.
- `docker compose config` clean (no deprecation warning); the StatefulSet parses and its ports/mounts/limits check out.
- JVM heap behaviour confirmed against the real base image under `-m 512m`, both before and after.
- `package/src/test/scripts/test-server-log-config-path.sh` still passes, and a stub-`java` run confirms `server.sh` puts `ARCADEDB_OPTS_GC`, `ARCADEDB_OPTS_MEMORY`, `ARCADEDB_SETTINGS` and a user-supplied `JAVA_OPTS` on the command line together.

The issue's own suggestion of a kind-based rendered-manifest smoke test in CI would be the natural next step; it is not in this PR.

## Summary by CodeRabbit

* **Deployment**
  * Updated Docker, Compose, and Kubernetes examples with improved health checks, readiness probes, volume mounts, secrets, and namespace configuration.
  * Replication exposure now uses Raft gRPC on port 2434 instead of the legacy binary protocol port.

* **Configuration**
  * Database settings, memory limits, and garbage-collector options now use dedicated environment settings.
  * Preserved support for custom JVM options and adaptive container memory sizing.

* **Bug Fixes**
  * Root passwords loaded from files now handle trailing line endings correctly and reject empty files.


lvca:
Both review rounds addressed in eabfb35c9.

**`package/arcadedb-builder.sh` generates the same #6841 bug** - valid, and the better catch of the two. The heredoc emitted `ENV JAVA_OPTS="-Xms1G -Xmx4G"`, which dies on any container limited below ~4.2GB and parks the setting in the one variable a caller overrides to add flags. Fixed with the same `ARCADEDB_OPTS_GC` / `-XX:MaxRAMPercentage=75` treatment as the shipped image.

The diagnosis of *why* it was missed was right, so I took the second half of the suggestion too rather than just patching the line: `PackagingManifestsTest` now extracts that heredoc and holds both images to the same three rules (no pinned heap, GC not in `JAVA_OPTS`, Raft port exposed and 2424 gone). Confirmed it fails all three against the previous template - a committed-files-only test would have let the next generated Dockerfile drift the same way.

**Create the namespace before the Secret** - valid. `kubectl create secret --namespace arcadedb` ran before the `apply` that creates the namespace, so it fails on a fresh cluster. Reordered, with a note on why the explicit `create namespace` is there even though the manifest declares it.

**Test named for an unreadable file exercises a missing one** - fair, renamed to `aMissingPasswordFileIsRejectedWithTheOffendingPath`. Agreed a real permission-denied test is not worth the platform flakiness; `Files.isReadable` collapses both to the same branch anyway.

**`defaultMode: 0444` on the secret volume** - keeping it, deliberately. A secret volume is a root:root tmpfs, so `0400` is unreadable by the unprivileged `arcadedb` user the image runs as; the alternatives are a `fsGroup` that hardcodes the image's GID into the manifest (and breaks silently if the image ever changes it) or running as root. In a single-container pod "world" is that one process, and the password never reaches argv, which is the exposure #6840 was actually about.

Test suite: `PackagingManifestsTest` 11, `RootPasswordFromFileTest` 6, `ServerSecurityIT` 6 - all green.

lvca:
Round 2 addressed in 18c21db5f. The review flags no blocking issue; all three points were minor and all three are now applied.

**Native Dockerfiles not covered by the port check** - right, and the reasoning generalises: they ship the same port surface and were changed here for the same reason, so pinning the other two while leaving these free is exactly the asymmetry that let this drift happen. The port check now walks all four; confirmed it fails if `Dockerfile.native.scratch` reverts to `EXPOSE 2424`. The heap and GC checks stay on the JVM images: neither native base has a shell or a launcher script, so there are no such variables to check.

**Secret file mode** - documented rather than changed. The comment now says what `0444` costs, that "world" is the single container of this pod, and gives `securityContext.fsGroup` + `0440` as the production tightening, along with why it is not the default here: it pins the manifest to the image's group id and breaks silently if that ever changes.

**Resource footprint** - fair, and it is a real trap on kind or docker-desktop. Added a note that the requests are sized for a genuine three-node cluster, with the two ways down (lower the limit, since the heap now follows it, or drop `replicas` to 1).

Nothing outstanding from either reviewer.

lvca:
Round 3 addressed in 5a07d723f. Review reports nothing blocking; two of the three nitpicks are now applied and the third is deliberately left alone.

**Literal-indentation matching in the test** - the one worth acting on, and the reasoning is right: a check that goes red on a cosmetic re-indent is a check that gets deleted rather than understood, which would undo the whole point of pinning these files. The port assertions now pair a name line with its port line without pinning any whitespace. Verified both directions, since a looser matcher is only an improvement if it still catches the drift: a re-indented ports block passes, and a container port reverted to 2424 still fails.

Chasing that surfaced a second gap, incidentally - the first version of the looser regex still required exactly one space after the dash, so a `-   containerPort:` re-indent failed. Fixed and re-verified.

**Memory request vs limit** - added. The comment now says the split is deliberate (Burstable: 1Gi reserved, heap free to reach 75% of the 2Gi ceiling) and what to do for Guaranteed QoS.

**Compose healthcheck grace** - leaving as is. The ~2.5 minutes is sized for what that quickstart actually does on first boot: it imports OpenBeer over the network before the server is ready, and a tighter budget would mark a container unhealthy for doing exactly what the example asks of it. Flagged as "no issue" in the review too.

Test suite green: `PackagingManifestsTest` 11, `RootPasswordFromFileTest` 6.

lvca:
Round 4 addressed in b02994d91.

**Assert the Service `targetPort`** (CodeRabbit) - valid, and a good catch: the checks read the Service's `port` but never its `targetPort`, so that one field could drift to a container port nothing listens on while all eleven assertions stayed green. That is the bug this test exists to catch, one field to the left. Now pinned as `port -> targetPort` for both published ports; confirmed it fails when the Raft targetPort is set back to 2424.

One design note: Kubernetes defaults `targetPort` to `port` when omitted, so only declared ones are compared - but the check asserts both are present rather than quietly emptying itself if someone deletes a line, which would be a check that passes by having nothing to say.

From the other review, all four points were flagged non-blocking and none need a change:

1. **Auto-join not independently verified** - pre-existing behaviour, untouched here. `KubernetesAutoJoin` runs on every k8s-flagged start and adds the node with an atomic `Mode.ADD`, which is what makes the single-entry `serverList` correct.
2. **`limits` sets memory but not cpu** - deliberate. A CPU limit means CFS throttling, which is the wrong default for a database; the memory limit is there because the heap is sized against it.
3. **Stale `EXPOSE 2480 2424` in `docs/superpowers/plans/`** - left alone on purpose. Those are dated planning records of what was decided at the time, not live documentation; editing them to match today's code would falsify the history. They are also the only remaining 2424 references in the tree.
4. **`wget` present in the base image** - checked before writing the healthcheck, same conclusion.

Nothing outstanding from either reviewer.

lvca:
Round 5 in ce1970060. One correction and one thing I cannot do here.

**`console.bat` passing a literal `%ARCADEDB_OPTS_GC%` to java** - the premise is not right, so this is not the startup failure described. In a **batch file** an undefined `%VAR%` expands to nothing; it is at the interactive prompt that it survives as literal text. Two pieces of evidence from this repo, both predating the PR:

1. `server.bat:34` is `if not "%JAVA_HOME%"=="" goto OkJHome`, and the only route to the PATH-based java lookup is that comparison being **true** when `JAVA_HOME` is unset. Under literal expansion `"%JAVA_HOME%"` is never `""`, so every user without `JAVA_HOME` would always jump to `:OkJHome`, get `JAVACMD=%JAVA_HOME%\bin\java.exe`, fail `if exist`, and be told `JAVA_HOME is not defined correctly` - the fallback the script exists to provide would be dead code. This is the same idiom `mvn.cmd` and `catalina.bat` use.
2. `server.bat` has shipped `set ARCADEDB_OPTS_MEMORY=` - which *deletes* the variable, it does not set it empty - immediately followed by `%ARCADEDB_OPTS_MEMORY%` on the java line, for years. Under literal expansion java would have received that token as its main class on every Windows start.

So `console.bat` was not broken before this PR and is not broken by it, and the same goes for its `%ARCADEDB_OPTS_MEMORY%` / `%ARCADEDB_JMX%` / `%ARCADEDB_SETTINGS%` lines.

What was fair in the observation is the **asymmetry** - five launchers declare the variable and one did not. Added the line to `console.bat` so all six read alike, with a comment recording that it is consistency rather than a fix, so nobody later goes hunting for a Windows failure that never existed. Worth being precise about because the conditional added to `server.bat` earlier in this PR *is* a behaviour change: there the assignment was unconditional and wiped a value the user had exported.

**`kind` smoke test** - agreed it is the verification that would actually close #6840, and I cannot run it in this environment: no `kind` and no `minikube`, and Docker Desktop's Kubernetes is not enabled (enabling it is a change to your machine, not mine to make). It would also need the image built from this branch and side-loaded, since `arcadedata/arcadedb:latest` on Docker Hub predates these fixes. What I could verify without a cluster is in the description: the manifest parses, its ports/mounts/limits are structurally right, every setting it passes resolves in `GlobalConfiguration`, and the heap behaviour was confirmed against the real base image under `-m 512m`. Happy to file the kind-in-CI smoke test as a follow-up issue if you want it tracked.

lvca:
Round 6 in a61d8ce41 - the two documentation points applied, the two flagged as no-change left alone.

**Self-referential peer entry** - worth stating, and stronger than "presumably harmless": arcadedb-0 listing itself is *required*. `HA_SERVER_LIST` is the initial Raft group membership, so a node absent from it is not a member - `RaftPeerAddressResolver.findLocalPeerId` resolves the local node out of that list. Only the first pod appears because the others genuinely are not members at boot; they synthesize a peer entry and join afterwards. Noted on the line, since it was the one part of the peer-list design a reader could reasonably take for a mistake.

**Surface the 0444 tradeoff where people skim** - agreed, and the same argument covers two neighbours, so the header now carries a short "before copying this into production" note pointing at the secret mode, the resource sizing, and the unauthenticated Raft gRPC port (which wants a NetworkPolicy and is documented in `KubernetesAutoJoin`'s class javadoc but nowhere an operator would look).

**`stripTrailingNewline` strips only one terminator** - leaving as is, and deliberately. Stripping more would mean deciding that a password cannot end in a blank line, which is a guess about someone's secret; stripping exactly the terminator the file format adds is the only rule that cannot silently change a valid password. Tested directly in `onlyTheFinalLineTerminatorIsStripped`.

**`generatedDockerfile()` heredoc extraction is brittle** - agreed, and left as is for the reason given: it fails loudly on the `isNotNegative()` assertion rather than passing vacuously, which is the property that matters for a guard like this.

That is everything from all six review rounds. Nothing outstanding.

Issue #6752, opened 2026-08-26
Auto-backup schedule is never cancelled on drop/close: a scheduled tick reopens a closed database (or errors forever after a drop)


ruispereira:
### Summary
An auto-backup schedule is never cancelled when its database is dropped or closed. Because the backup task resolves the database with `allowLoad=true`, the next scheduled tick **reopens a database the operator explicitly closed**; after a `drop` the same tick throws and logs SEVERE + a CRITICAL server-event on every fire, forever.

### Where (HEAD `af4f67fd2`)
- `server/src/main/java/com/arcadedb/server/ArcadeDBServer.java:879-883` — `removeDatabase()` only removes the map entry; there is no plugin notification:
```java
public void removeDatabase(final String databaseName) {
  synchronized (databasesLock) { databases.remove(databaseName); }
}
```
- `server/src/main/java/com/arcadedb/server/http/handler/PostServerCommandHandler.java:776` (`dropDatabase`) and `:789` (`closeDatabase`) call `removeDatabase(...)` and never `cancelBackup(...)`. The HA drop/snapshot-install paths (`ArcadeStateMachine.java:2580`, `SnapshotInstaller.java:440,543`) do the same.
- `server/src/main/java/com/arcadedb/server/backup/BackupTask.java:175` — `final Database database = server.getDatabase(databaseName);` resolves via `getDatabase(name, false, true)` (`ArcadeDBServer.java:759-760`, `allowLoad=true`).
- `ServerPlugin` has **no** database create/drop callback, so `AutoBackupSchedulerPlugin` can never learn the database went away (the only non-internal `cancelBackup` caller is `reloadConfiguration`).

### Repro
1. `close database X` → X is removed from the registry. Its `ScheduledFuture` is still live.
2. Next tick: `BackupTask.performBackup()` → `server.getDatabase("X")` **reopens X from disk**, re-registers it and backs it up. The explicit close is silently defeated and the page cache is pinned again.
3. With `drop database X` instead: every tick throws `DatabaseOperationException`, logged SEVERE plus a CRITICAL server-event-log entry, indefinitely (`scheduleAtFixedRate`; the cron variant reschedules itself in its `finally`).

This is the #6203 "scheduler holding an abandoned database" class, still open for the backup scheduler (the materialized-view and time-series schedulers were fixed).

### Suggested fix
[...]

PR #6853, opened 2026-08-27
fix: five small defects across the console, the Studio and the plugin lifecycle (#6852, #6830, #6829, #6828, #6827)

### LINKED ISSUE #6827: Console TerminalParser drops every backslash before the command reaches the engine, so a Windows path or a regex literal cannot be typed (or replayed with load)
### Summary
`TerminalParser` (which `extends DefaultParser`, whose default escape set is `{'\\'}`) silently drops every unescaped backslash before the command text reaches the query engine. Any SQL statement typed in the console, passed with `-b`, or replayed with `load` has one level of backslash escaping consumed by the console, so the engine never sees what the user wrote.

### Where (HEAD `9c9d024db`)
`console/src/main/java/com/arcadedb/console/TerminalParser.java:153` (inside a quoted string) and `:166` (outside one):
```java
} else if (!this.isEscapeChar(line, i)) {
  current.append(c);
}
```
When the character **is** the escape char no branch runs at all, so the backslash is discarded and the next character is appended verbatim. jline's `isEscapeChar(buffer, pos)` is true for an unescaped `\`, so `\U` becomes `U` and `\\` becomes `\`: standard shell-style unescaping applied to text that is about to be handed to a SQL parser.

The maintainers' own javadoc documents the mechanism for one case: "That inner quote reaches this method already stripped of its escaping backslash by `TerminalParser#parse`" (`Console.java:393-395`). But the same single `parse()` serves all input paths: interactive lines, `-b` batch arguments, and every statement read by `executeLoad` (`Console.java:811`, `:844`). From `execute(w)` the parser's word goes straight to `executeSQL` (`Console.java:305`) and then straight to `databaseProxy.command` (`Console.java:695`), with no re-escaping anywhere.

### Repro
```
insert into Doc set winPath = 'C:\Users\bob'
```
stores `C:Usersbob`. No warning, no error, and the identical statement sent to `/api/v1/command` stores the correct path.

Because `executeLoad` parses through the same code, an `.sql` script containing any backslash escaping cannot be replayed with `load`: a Windows path, a regex for `MATCHES`, `\\` for a literal backslash, or `\"` inside a `content {...}` clause. Either the data changes meaning, or the engine receives a broken literal and reports a syntax error pointing at text the user never typed.

No test in `console/src/test` asserts a backslash outcome for a SQL statement: `TerminalParserTest` has none, and `ConsoleTest:404` covers only the `set language = 'it\'s a test'` value path.

### Suggested fix
Keep the escape character in the emitted word while still using it for quote and delimiter tracking: replace the two `!isEscapeChar(...)` guards with an unconditional `current.append(c)`. `isEscaped()` already prevents an escaped quote from closing the string and an escaped `;` from splitting a command. `stripMatchingQuotes()` then has to unescape `\'`/`\"` itself, which is the one place where shell-like unescaping is actually wanted.

### Scope / confidence
Medium, and partly a design call: the stripping is deliberate for `SET` (`Console.java:387-397` documents and relies on it). The finding is confined to the statement path forwarded to the query engine, where the engine - not the console - owns string-literal escaping. Worth deciding explicitly either way, because today a Windows path cannot be inserted from the console.

### LINKED ISSUE #6828: Console.close() commits on a stale batch counter and its blanket catch(Throwable) then skips the terminal flush and the DatabaseFactory close
### Summary
`Console.close()` runs four steps inside one try with `catch (Throwable t) { // IGNORE ANY EXCEPTION AT CLOSING }`, and the first step commits on a batch counter that nothing resets. In the default batch flow that first step throws, which silently skips the terminal flush, the database close and the factory close. On the interactive path the same swallow hides a genuinely failed exit-time commit.

### Where (HEAD `9c9d024db`)
`console/src/main/java/com/arcadedb/console/Console.java:226-248`:
```java
try {
  if (transactionBatchSize > 0 && currentOperationsInBatch > 0) {
    currentOperationsInBatch = 0;
    databaseProxy.commit();
  }
  ... flushOutput();
  ... databaseProxy.close();
  ... databaseFactory.close();
} catch (Throwable t) {
  // IGNORE ANY EXCEPTION AT CLOSING
}
```
`currentOperationsInBatch` is incremented only in `executeSQL` (`:706`) and reset only by the periodic commit (`:708`). Neither `executeClose()` (`:449`), `executeCommit()` (`:439`) nor `executeRollback()` resets it, and the guard checks neither `databaseProxy != null` nor `isTransactionActive()`.

### Repro
```bash
arcadedb-console.sh -b "connect ./mydb" "set transactionBatchSize = 100" "insert into Doc set n = 1"
```
Batch mode appends `exit` (`:213`); `executeClose()` commits and nulls `databaseProxy`; then the `finally console.close()` (`:221`) NPEs at `:230`. `flushOutput()` (`:233-234`) never runs, so anything still buffered in the jline writer is lost when `main()` calls `System.exit` (`:174-175`), and `databaseFactory` is left open.

The same swallow hides a real failure: on the interactive Ctrl-D / shutdown-hook path (`:126-162`) a pending batch of up to `transactionBatchSize` statements is committed at `:230`. If that commit fails, nothing is printed, `errored` is not set, and the process exits 0 while the data was never committed.

`commit()` on a non-active transaction is not a no-op, which is exactly why `executeCommit`/`executeClose` guard with `isTransactionActive()`. No test in `ConsoleTest`/`ConsoleBatchTest` asserts anything about `close()`.

### Suggested fix
Reset `currentOperationsInBatch = 0` in `executeClose()`/`executeCommit()`/`executeRollback()`, guard the block with `databaseProxy != null && databaseProxy.isTransactionActive()`, and give each of the four steps its own try/catch so a failed final commit is reported (`outputError` plus `errored = true`) and never prevents the flush and the two closes.

### Scope / confidence
High. The NPE variant is reachable in the default batch flow, which is what makes the skipped flush routine rather than theoretical.

### LINKED ISSUE #6829: The console persists and echoes the plaintext password given to connect / create user (.history file, and stdout in -b mode)
### Summary
The console writes every typed line to `./.history` in cleartext after each command, with no `HISTORY_IGNORE` filter and no masked read. Since credentials are part of the command syntax, every `connect` and `create user` password is persisted in plaintext, and in `-b` mode the same line is echoed to stdout.

### Where (HEAD `9c9d024db`)
`console/src/main/java/com/arcadedb/console/Console.java:131-132` builds the reader with a plain history file and a plain `DefaultHistory()`:
```java
.variable("history-file", ".history")
```
`:143-147` saves after every line:
```java
lineReader.getHistory().save();
```
`grep -n "HISTORY_IGNORE\|mask" Console.java` returns no hits. Credentials are inline in the syntax:
- `connect remote:host/db <user> <password>` (`:1240`)
- `create user x identified by <password>` (`:560`, `:575`)

In batch mode the line is echoed by `output(3, getPrompt() + trimmedWord)` (`:897`), because `execute(args)` passes `printCommand = true` (`:213`).

### Repro
```bash
cd /tmp && arcadedb-console.sh
> connect remote:localhost/mydb root MySecret1!
```
then `grep root /tmp/.history` shows the root password in cleartext, readable by anyone who can read the directory, and it survives the session indefinitely.

In CI:
```bash
arcadedb-console.sh -b "connect remote:$HOST/db root $PASS" "..."
```
prints the root password into the build log.

There is no way to avoid the inline password today: `connectToRemoteServer` (`:1240`) requires exactly the 3-token form.

### Suggested fix
Set the jline `HISTORY_IGNORE` pattern (or override history add) to skip lines starting with `connect` / `create user`, and mask the password in the echoed command at `:897`. Better still, when the password is omitted from the command, read it with `lineReader.readLine(prompt, '*')`.

### Scope / confidence
High on the mechanism. Severity depends on the environment, but a shared build agent or a shared workstation directory makes it a real credential leak.

### LINKED ISSUE #6830: Two small parsing defects: console connect splits credentials on a single space; Studio graph widget HTML-escapes the database name into a URL path
Two small, independent defects found while auditing the console argument parsing and the Studio graph widget. Both are one-line fixes; grouping them to keep the tracker tidy.

### 1. `connect remote:` splits on a single space and demands exactly 3 tokens

`console/src/main/java/com/arcadedb/console/Console.java:1240-1242`:
```java
final String[] serverUserPassword = conn.trim().split(" ");
if (serverUserPassword.length != 3)
  throw new ConsoleException("URL username and password are missing");
```
`split(" ")` produces empty tokens for runs of whitespace, and there is no quoting escape hatch: the parser deliberately **keeps** the quote characters in the word (`TerminalParser.java:141`, `:150`), so `connect remote:h/db root "my pass"` arrives as four tokens.

Repro - both fail with "URL username and password are missing" although both values are present:
```
connect remote:localhost/mydb root "my pass"
connect remote:localhost/mydb root  secret      (two spaces)
```
A password containing a space is accepted by the server and by Studio, so such an account is simply unusable from the console, and the diagnostic sends the user looking for the wrong problem.

*Fix:* `conn.trim().split("\\s+", 3)` so the password keeps its spaces, and split the error message into "missing username/password" versus "unexpected extra token".

Note: `executeCreateUser` does reject passwords with spaces (`checkHasSpaces("User password", ...)`, `:579`), but that only covers users created *from the console*, not accounts created over HTTP/Studio or migrated, and it does not explain rejecting `root  secret`. `RemoteConsoleIT` only uses space-free credentials.

### 2. Studio HTML-escapes the database name into a URL path

`studio/src/main/resources/static/js/studio-graph-widget.js:696`:
```js
url: "api/v1/command/" + escapeHtml(getCurrentDatabase()),
```
`escapeHtml()` (`studio-utils.js:159-166`) is an HTML-entity encoder, not a URL encoder: `&` becomes `&amp;`, `'` becomes `&#039;`, `"` becomes `&quot;`, and a null input returns `null` (stringified into the URL as `"null"`).

This is the only one of the ~40 `api/v1/{command,query}/` call sites in the Studio that transforms the database name at all; every other one concatenates it raw, including the neighbouring request at `:590` in the same file. So it reads as a copy/paste of the wrong helper rather than a convention.

Repro: with a database whose name contains `&`, `'`, `"`, `<` or `>`, "add node from record" in the graph widget POSTs to `api/v1/command/a&amp;b`, the server answers 404, and `data.result.vertices[0]` (`:707`) leaves `vertex` null - while the same action from any other Studio panel works.

*Fix:* `encodeURIComponent(getCurrentDatabase())` here, and ideally at the other call sites too, which have the mirror-image problem with `/`, `#` and spaces.

### LINKED ISSUE #6852: AutoBackupSchedulerPlugin NPE on every database registration at startup (26.9.1): this.server is null
## Environment

- `arcadedata/arcadedb:latest` Docker image pulled 2026-08-27 (engine `arcadedb-engine-26.9.1`), JDK 21, default `./bin/server.sh`
- Data directory with 10 existing databases, bind-mounted to `/home/arcadedb/databases`
- No custom plugin configuration; `JAVA_OPTS=-Darcadedb.server.rootPassword=...` only
- Same data directory ran on 26.6.1 without this error

## What happens

At startup, once per database, the server logs:

```
2026-08-27 21:27:11.034 SEVER [ArcadeDBServer] <ArcadeDB_0> Error on notifying plugin 'AutoBackupSchedulerPlugin' that database 'marksan_ai' was registered
java.lang.NullPointerException: Cannot invoke "com.arcadedb.server.ArcadeDBServer.existsDatabase(String)" because "this.server" is null
```

Ten databases, ten identical stack traces (`claude_memory`, `platform`, `transprt_net`, ...). After startup the server runs normally: queries, FULL_TEXT and LSM_VECTOR indexes, writes all fine, no further errors in the log.

## Expected

No exception. It looks like `AutoBackupSchedulerPlugin` receives the database-registered notification before `configure(server, ...)` has set its `server` field (init order), so `existsDatabase()` dereferences null. Unclear whether the scheduler still picks those databases up for automatic backups afterwards, which is the part that matters.

## Repro

1. Start 26.9.1 against a data directory that already contains one or more databases.
2. Read the server log during startup.

Happy to test a fix or provide the full log.

lvca:
Closes #6852, closes #6830, closes #6829, closes #6828, closes #6827.

Five small, independent defects, grouped because four of them live in the console.

## #6852 - `AutoBackupSchedulerPlugin` NPE on every database registration at startup

**Root cause is one level up from where it was reported.** Plugin discovery installs every plugin instance up front, but `loadDatabases()` runs *before* `startPlugins(AFTER_DATABASES_OPEN)`. So a plugin of that priority was handed a "database registered" callback for every database already on disk while its own `server` field was still null. Ten databases, ten SEVERE stack traces, at every single startup.

The fix narrows the notification audience: `notifyPlugins()` now iterates the plugins whose `configure()` has returned, tracked on `PluginDescriptor` next to `started`. `started` cannot answer that question - it is set only *after* `startService()` returns and only when the plugin reports itself active, so a plugin that opens or registers databases while starting would be hidden from its own callbacks.

`AutoBackupSchedulerPlugin.syncDatabase()` also checks its own scheduler before dereferencing the server. A lifecycle callback is announced from whichever thread mutated the registry, so "nobody would call it that early" is not something that class can assume - and with no scheduler published there is genuinely nothing to reconcile, since `startService()` schedules everything that exists by then.

Both halves are pinned: nothing arrives before `configure()`, **and** everything after it still does. Muting the callbacks entirely would have "fixed" the stack traces by re-breaking #6752, whose IT still passes.

## #6827 - `TerminalParser` drops every backslash

The escape character now stays in the emitted word while still doing its structural job: an escaped quote does not close a string, an escaped semicolon does not split a command. The engine, not the console, owns string-literal escaping.

**The reported repro understates the bug.** The issue says the same statement over `/api/v1/command` stores the correct path - it does not. The engine's SQL lexer rejects a bare `\U` inside a literal, so `'C:\Users\bob'` fails there too; the form that works is `'C:\\Users\\bob'`, which the console turned into the form that fails to tokenize. There was no way to type a backslash from the console at all, in either direction.

`load` had the same defect from the other end. `FileUtils.decodeFromFile()` existed only to double every `\\` it read so that the level the parser then consumed came back out even. That pair cancelled out only for an *even* number of backslashes - a lone `\` in a script was still swallowed - and it is why the existing `console-batch.sql` fixture worked. Both halves are gone, and the helper with them: it had exactly one caller in the tree and no meaning on its own.

`SET` is the one place where shell-like unescaping is actually wanted, so `stripMatchingQuotes()` does it there - and can now be exact about it. With the escapes preserved, the closing quote is the first *unescaped* occurrence of the opening one, which removes the ambiguity the old javadoc had to document and apologise for.

## #6828 - `Console.close()` commits on a stale counter and its blanket catch skips the rest

- `currentOperationsInBatch` is reset wherever the pending batch is disposed of: `executeCommit()`, `executeRollback()`, `executeClose()`.
- The guard requires a live proxy with an active transaction, not just a non-zero counter.
- Each of the four steps gets its own try/catch, so the flush and the two closes cannot be skipped by the commit.
- A failed exit-time commit is reported and sets `errored`, instead of the process exiting 0 over data that was never written - which is the half of this bug that silently loses an interactive Ctrl-D batch.

## #6829 - the console persists and echoes the plaintext password

- The line reader's history masks the password before recording it, so `.history` never receives one.
- The batch-mode command echo masks it too, so it does not land in a CI log.
- **And the part that makes those masks avoidable rather than merely tidy:** both `connect` and `create user` now accept the password-less form and ask for the password with the echo masked. Without that there is no way to authenticate without writing the password down somewhere, so the masking would only have hidden the leak from the person creating it.

The masking rules live in `ConsoleCredentials`, matched per statement on the command keyword rather than on `remote:` or `identified by` anywhere in the text - otherwise an ordinary `insert into Doc set note = 'identified by the auditor'` gets mangled on its way to the log.

## #6830 (1) - `connect remote:` splits on a single space

Now splits on runs of whitespace with the password as the remainder, so a doubled blank and a password containing a space both work, and the diagnostic says which half is missing instead of claiming both are. The local branch of `executeConnect` had the identical defect (`MODE.valueOf(\"\")` on a doubled blank) and gets the same treatment.

`checkHasSpaces(\"User password\", ...)` in `executeCreateUser` is dropped: `IDENTIFIED BY` and `GRANT CONNECT TO` delimit the password positionally, the server and Studio both accept spaces, and rejecting them only there made an account the console could not create.

## #6830 (2) - the Studio HTML-escapes the database name into a URL path

Fixed more broadly than reported. `escapeHtml()` at `studio-graph-widget.js:696` was the reported case, but the neighbouring `:590` had it too (via a variable), and ~40 other call sites had the mirror-image problem of concatenating the name raw. All `api/v1/{command,query,progress}` sites now build the path segment with one `encodeDatabaseName()` helper.

The names that were held HTML-escaped *at the source* are now held raw and escaped at the HTML sinks. That escaping was not only a URL bug: in `dropDatabase()` it made the typed-name confirmation compare against `a&amp;b` and sent that name to the server as the one to drop, and in `populateHistoryPanel()` it made the filter compare against entries stored under the raw name.

## Verification

- `console`: 116 tests green (`ConsoleTest` 62, `TerminalParserTest` 29, new `ConsoleCredentialsTest` 8, `ConsoleBatchTest` 8, ...). The 2 `ConsoleAsyncInsertTest` errors are pre-existing and reproduce identically on unmodified `main`.
- All 6 new `TerminalParserTest` cases fail against the old parser; `Issue6852PluginLifecycleBeforeConfigureTest` fails against the old dispatch with exactly the reported shape (`hadServer=false`).
- `server`: plugin + backup suites green (52 tests), `Issue6752AutoBackupScheduleLifecycleIT` still green.
- `studio`: 55 node tests green, including the new `database-name-url-encoding.test.js`, which also asserts the invariant that no api/v1 URL builds its path segment any other way.

## Summary by CodeRabbit

* **New Features**
  * Console prompts can securely request omitted passwords and mask credentials in command history and output.
  * Console commands and scripts preserve backslashes and support passwords containing spaces.
  * Studio supports database names containing special URL characters across database, graph, security, profiler, and time-series tools.

* **Bug Fixes**
  * Improved console transaction cleanup and shutdown reliability.
  * Prevented unconfigured plugins and backup scheduling from causing errors.
  * Improved database lifecycle callback handling.
  * Studio now excludes non-primary buckets from default bucket loading.

### DIFF
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -262,13 +262,18 @@ public class Console {
    * time there is nothing left for the user to do about them.
    */
   public void close() {
-    // COMMITTING NEEDS AN ACTIVE TRANSACTION ON A LIVE PROXY, NOT JUST A NON-ZERO COUNTER: commit() ON AN INACTIVE
-    // TRANSACTION IS NOT A NO-OP, WHICH IS WHY executeCommit()/executeClose() GUARD ON isTransactionActive() TOO
-    if (transactionBatchSize > 0 && currentOperationsInBatch > 0 && databaseProxy != null
-        && databaseProxy.isTransactionActive()) {
+    if (transactionBatchSize > 0 && currentOperationsInBatch > 0 && databaseProxy != null) {
       try {
-        currentOperationsInBatch = 0;
-        databaseProxy.commit();
+        // COMMITTING NEEDS AN ACTIVE TRANSACTION, NOT JUST A NON-ZERO COUNTER: commit() ON AN INACTIVE TRANSACTION IS
+        // NOT A NO-OP, WHICH IS WHY executeCommit()/executeClose() GUARD ON isTransactionActive() TOO. THE GUARD IS
+        // INSIDE THE TRY BECAUSE IT CAN THROW ON ITS OWN: ON THE EMBEDDED PATH isTransactionActive() RESOLVES THE
+        // THREAD'S CACHED TRANSACTION AND RAISES InvalidDatabaseInstanceException WHEN IT BELONGS TO A DIFFERENT
+        // LocalDatabase INSTANCE FOR THE SAME PATH - RECONNECTING TO THE SAME DATABASE IN ONE SESSION DOES THAT. A
+        // THROW HERE WOULD SKIP THE FLUSH AND THE TWO CLOSES ALL OVER AGAIN, JUST FROM ANOTHER TRIGGER (ISSUE #6828)
+        if (databaseProxy.isTransactionActive()) {
+          currentOperationsInBatch = 0;
+          databaseProxy.commit();
+        }
       } catch (Throwable t) {
         errored = true;
         outputError(t);
@@ -1376,7 +1381,25 @@ public class Console {
     if (batchMode || !systemTerminal)
       throw new ConsoleException("Password for user '" + userName + "' is missing");
 
-    final String password = getLineReader().readLine("Password for '" + userName + "': ", '*');
+    final LineReader reader = getLineReader();
+
+    // BELT AND BRACES ON THE ONE LINE THAT IS A BARE PASSWORD. jline ALREADY DROPS A MASKED LINE - ITS
+    // SimpleMaskingCallback.history() RETURNS null AND LineReaderImpl.finish() SKIPS A null - BUT THAT IS A LIBRARY
+    // INTERNAL, AND ConsoleCredentials CANNOT BE THE SAFETY NET HERE: IT RECOGNISES A PASSWORD BY THE COMMAND KEYWORD
+    // IN FRONT OF IT, AND THIS LINE HAS NONE. SAYING IT OUT LOUD KEEPS THE GUARANTEE LOCAL TO THE CODE THAT NEEDS IT
+    final Object previousSetting = reader.getVariable(LineReader.DISABLE_HISTORY);
+    reader.setVariable(LineReader.DISABLE_HISTORY, true);
+    final String password;
+    try {
+      password = reader.readLine("Password for '" + userName + "': ", '*');
+    } catch (final UserInterruptException | EndOfFileException e) {
+      // CTRL-C / CTRL-D AT THE PROMPT IS A DECISION, NOT A FAILURE: SAY SO INSTEAD OF LETTING A jline EXCEPTION
+      // SURFACE AS AN OPAQUE ERROR FROM connect / create user
+      throw new ConsoleException("Password entry cancelled");
+    } finally {
+      reader.setVariable(LineReader.DISABLE_HISTORY, previousSetting);
+    }
+
     if (password == null || password.isEmpty())
       throw new ConsoleException("Password for user '" + userName + "' is empty");
     return password;
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -1351,7 +1351,12 @@ public class Console {
 
     final String[] serverParts = serverUserPassword[0].split("/");
     if ((needsDatabase && serverParts.length != 2) || (!needsDatabase && serverParts.length != 1))
-      throw new ConsoleException("Remote URL '" + url + "' not valid");
+      // REPORT ONLY THE ADDRESS, NEVER THE WHOLE ARGUMENT: `url` STILL CARRIES THE INLINE PASSWORD, AND THIS MESSAGE
+      // GOES TO THE INTERACTIVE OUTPUT AND TO THE BATCH LOG. IT IS ALSO THE MORE PRECISE HALF - THE ADDRESS IS WHAT
+      // FAILED TO SPLIT (ISSUE #6829)
+      throw new ConsoleException(
+          "Remote URL '" + REMOTE_PREFIX + serverUserPassword[0] + "' is not valid, expected " + REMOTE_PREFIX
+              + "<host>[:<port>]" + (needsDatabase ? "/<database>" : ""));
 
     final String remoteServer;
     final int remotePort;
--- a/console/src/main/java/com/arcadedb/console/Console.java
+++ b/console/src/main/java/com/arcadedb/console/Console.java
@@ -43,7 +43,6 @@ import com.arcadedb.remote.RemoteDatabase;
 import com.arcadedb.remote.RemoteServer;
 import com.arcadedb.schema.DocumentType;
 import com.arcadedb.utility.AnsiCode;
-import com.arcadedb.utility.FileUtils;
 import com.arcadedb.utility.RecordTableFormatter;
 import com.arcadedb.utility.StringUtils;
 import com.arcadedb.utility.TableFormatter;
@@ -64,6 +63,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -76,6 +76,7 @@ public class Console {
   private static final String               REMOTE_PREFIX            = "remote:";
   private static final String               LOCAL_PREFIX             = "local:";
   private static final String               SQL_LANGUAGE             = "SQL";
+  private static final String               HISTORY_FILE             = ".history";
   private final        Terminal             terminal;
   private final        TerminalParser       parser                   = new TerminalParser();
   private              ConsoleOutput        output;
@@ -95,6 +96,11 @@ public class Console {
   private              boolean              batchMode                = false;
   private              boolean              failAtEnd                = false;
   private static       boolean              errored                  = false;
+  // BUILT LAZILY AND SHARED BY interactiveMode() AND BY THE MASKED PASSWORD PROMPT, SO A PASSWORD CAN BE ASKED FOR
+  // WITHOUT EVER APPEARING ON A LINE THAT GOES TO THE HISTORY FILE (ISSUE #6829)
+  private              LineReader           lineReader;
+  // WHETHER THE PROCESS OWNS A REAL TERMINAL: A MASKED PROMPT IS ONLY POSSIBLE WHEN SOMEBODY IS THERE TO TYPE INTO IT
+  private final        boolean              systemTerminal;
 
   public Console(final DatabaseInternal database) throws IOException {
     this();
@@ -116,20 +122,39 @@ public class Console {
 
     GlobalConfiguration.PROFILE.setValue("low-cpu");
 
-    boolean system = System.console() != null;
-    terminal = TerminalBuilder.builder().system(system).streams(System.in, System.out).jni(true).build();
+    systemTerminal = System.console() != null;
+    terminal = TerminalBuilder.builder().system(systemTerminal).streams(System.in, System.out).jni(true).build();
 
     output(3, "%s Console v%s - %s (%s)", Constants.PRODUCT, Constants.getRawVersion(), Constants.COPYRIGHT, Constants.URL);
     output(3, "%s", Constants.SPONSOR_MSG);
   }
 
-  public void interactiveMode() throws IOException {
-    final Completer completer = new StringsCompleter("align database", "begin", "rollback", "commit", "check database", "close",
-        "connect", "create database", "create user", "drop database", "drop user", "export", "import", "help", "info types",
-        "list databases", "load", "exit", "quit", "set", "match", "select", "insert into", "update", "delete", "pwd");
+  /**
+   * Returns the line reader, building it on first use.
+   * <p>
+   * The history it is given masks the passwords that the `connect` and `create user` syntaxes carry inline before they are
+   * recorded, so they never reach the `.history` file - which is written after every command, in the working directory,
+   * with the process umask, and survives the session indefinitely (issue #6829).
+   */
+  private LineReader getLineReader() {
+    if (lineReader == null) {
+      final Completer completer = new StringsCompleter("align database", "begin", "rollback", "commit", "check database", "close",
+          "connect", "create database", "create user", "drop database", "drop user", "export", "import", "help", "info types",
+          "list databases", "load", "exit", "quit", "set", "match", "select", "insert into", "update", "delete", "pwd");
 
-    final LineReader lineReader = LineReaderBuilder.builder().terminal(terminal).parser(parser).variable("history-file", ".history")
-        .history(new DefaultHistory()).completer(completer).build();
+      lineReader = LineReaderBuilder.builder().terminal(terminal).parser(parser)
+          .variable(LineReader.HISTORY_FILE, HISTORY_FILE).history(new DefaultHistory() {
+            @Override
+            public void add(final Instant time, final String line) {
+              super.add(time, ConsoleCredentials.mask(line));
+            }
+          }).completer(completer).build();
+    }
+    return lineReader;
+  }
+
+  public void interactiveMode() throws IOException {
+    final LineReader lineReader = getLineReader();
 
     Runtime.getRuntime().addShutdownHook(new Thread(this::close));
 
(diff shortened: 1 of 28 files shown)


lvca:
Thanks - both findings addressed in bd894e9a6, plus both nits.

**Finding 2 (`close()` guard outside the try)** - confirmed and fixed. `LocalDatabase.isTransactionActive()` -> `getTransactionIfExists()` throws `InvalidDatabaseInstanceException` at `LocalDatabase.java:613` when the thread's cached transaction belongs to a different instance for the same path. The guard now lives inside the try, so it cannot skip the flush and the two closes.

**Finding 1 (password reaching `.history`)** - the mechanism is right but the leak is not live in jline 4.3.1: `readLine(prompt, mask)` wraps the mask in a `SimpleMaskingCallback`, whose `history(line)` returns `null`, and `LineReaderImpl.finish()` only calls `history.add(...)` when `historyLine != null && length() > 0`. So the masked line never reaches `DefaultHistory.add`, overridden or not.

Fixed anyway, exactly as suggested. That is a library internal, this is a security fix, and the second half of the finding is the part that really matters: `ConsoleCredentials` cannot be the safety net for that line, because it recognises a password by the command keyword in front of it and a bare password has none. `DISABLE_HISTORY` is now set around the read (restoring the previous value, `null` included - `ReaderUtils.getBoolean` falls back to the default for a null), which keeps the guarantee local to the code that depends on it. Added `aBarePasswordIsNotRecognisedAndMustNotBeLeftToThisClass` to pin the reasoning; the end-to-end interactive path still needs a driven terminal and remains uncovered, as you noted.

**Nits** - `carriesPassword()` dropped (no caller outside its own test). `askPassword()` now catches `UserInterruptException`/`EndOfFileException` and reports "Password entry cancelled".

Console suite re-run: 116 tests green.

lvca:
All three CodeRabbit points addressed in 5611f4222.

**1. `initialized` visibility** - valid, now `volatile`. `startPlugins()` deliberately does not hold the `plugins` monitor across the plugin lifecycle calls (they can block), so the monitor `getInitializedPlugins()` takes orders nothing against that write, and the reader is any HTTP or HA thread registering a database. A stale `false` would silently drop the callback #6752 exists to deliver. `started` is left plain: it is only ever read on the lifecycle thread.

**2. `Error` from `startService()`** - valid, fixed. The inner handler rethrows `Exception | Error` and a plugin jar missing a class fails with the second one, so the outer `catch (Exception)` left the descriptor initialized after `stopService()` had already released its state. The `Error` is rethrown unwrapped - it is not that layer's to reinterpret. Three tests now cover the audience: `aPluginThatFailsToStartStopsReceivingLifecycleCallbacks`, `aPluginThatFailsToStartWithAnErrorAlsoStopsReceivingLifecycleCallbacks`, and `onlyTheConfiguredPluginsAreInTheLifecycleAudience`.

**3. Studio admin commands** - the underlying defect is real, though the suggested remedy is not the right one here. There is no identifier quoting to apply: `PostServerCommandHandler.extractTarget()` reads the database name as the verbatim rest of the command string, so back-ticks would become part of the name. The actual bug is one layer up - the body was hand-assembled as `"{ 'command': '... " + name + "' }"`, and a name containing `'` (the server rejects only `/`, `\`, NUL and `..`) made it malformed JSON. All four payloads now use `JSON.stringify`, which is what every other Studio call site already does.

Two extras found while there, same family as #6830:

- `createDatabase()` ran the typed name through `encodeURI()` before putting it in that payload, so a name typed as `my db` **created** the database as `my%20db` - the wrong-encoder mistake again, just pointed at the payload instead of the URL.
- `studio-server.js:540` was the one remaining hand-built body, with a file name interpolated into it.

Both invariants are now pinned by `database-name-url-encoding.test.js`: no hand-built JSON body anywhere in the JS tree, and no `encodeURI()` on a name headed for a command payload.

Re-run: 55 server plugin/backup tests, 57 studio tests, 116 console tests - all green.

lvca:
Valid, fixed in 8dd5ff718. `connectToRemoteServer` is reached before any of the #6829 masking runs, so `"Remote URL '" + url + "' not valid"` echoed the inline password into the interactive output and, in batch mode, into the build log - the same leak, from the one path the masking could not cover.

The message now reports `serverUserPassword[0]` alone, which is also the more precise half: the address is what failed to split, so the whole argument was never the right thing to show. It now says what shape was expected too.

Regression test `aMalformedRemoteUrlIsReportedWithoutTheInlinePassword` asserts the address is present, the password is not, and that the same holds for what actually reaches the console output. Console suite: 118 green.

I checked the other `ConsoleException` messages in the file while there - the remaining ones echo a `SET` value or no user input at all, and none of them is on a credential-carrying path.

lvca:
Both reviews came back with no blocking issues. Applied the four nits anyway in 7db440c54:

- **Redundant counter reset in `close()`** - dropped the assignment inside the commit branch; the unconditional one right after the try/catch always runs, so the inner one only made the reader check whether the two could disagree.
- **Word-boundary matching in `ConsoleCredentials`** - the command list mirrors the console's own dispatch and will drift as commands are added, so it now matches whole commands. Without the boundary a future `connect-cluster` would be routed by the `connect` entry and have its third argument masked as a password. Test: `aCommandThatMerelySharesAPrefixIsNotTreatedAsThatCommand`.
- **`stripMatchingQuotes` unescaping the non-delimiting quote** - narrowed to the quote character that actually delimits the value, plus the backslash. The other quote never needed escaping in there, so a backslash in front of it is data: `set p = 'C:\"x'` now keeps it. Thanks for flagging it - "documented as intentional" was doing more work than the behaviour deserved.
- **Studio invariant regex blind spot** - it now matches the endpoint prefix in any form (template literal, parenthesized expression) and requires the helper on the same line, rather than only the `"..." + identifier` shape written today. The point of that test is the call sites nobody has written yet.

On the two items both reviews raised and closed themselves - the raw database name in `"drop database " + name` and the `PostServerCommandHandler` parsing - I agree with the conclusion: `extractTarget()` takes the verbatim rest of the command, so there is no quoting layer to apply and adding one would corrupt the name. Worth revisiting only if that parsing ever gains structure.

Final: console 119 tests, studio 57, server plugin/backup 55 - all green.

lvca:
Thanks - no code change on this round. All three observations checked, reasons below.

**1. Semicolon in a database name smuggling a second command** - not reachable on this endpoint. The `;` splitting is client-side, in the CLI's `TerminalParser`; `PostServerCommandHandler` has no notion of a statement separator. It dispatches on `command_lc.startsWith(...)` and then `extractTarget()` takes the verbatim rest of the string as one name, so `drop database a;drop database b` attempts to drop a database literally named `a;drop database b` and fails to find it. Nothing runs twice. Worth revisiting only if that handler ever gains real parsing, which is the same conclusion the previous round reached about quoting the name.

**2. No automated coverage of the interactive masked-password path** - correct, and I tried to close it rather than just agree. jline's dumb terminal will not drive a masked `readLine` off a piped stream (`ClosedException` as soon as the stream ends), so a test would need a `PipedInputStream` pumped from a second thread **plus** two test-only seams into `Console`: one to inject the terminal it currently builds itself from `System.in`/`System.out`, and one to override `systemTerminal`, since `System.console()` is null under Surefire and `askPassword()` bails out before reading. That is a lot of production-code surface added for one path CI can never exercise, so I left it.

What the fix actually depends on is covered where it lives: `DefaultHistory.add()` checks `DISABLE_HISTORY` as its first statement, and `aBarePasswordIsNotRecognisedAndMustNotBeLeftToThisClass` pins why the masker must not be the safety net there. Flagging the manual smoke test for @lvca as a pre-release item: `connect remote:<host>/<db> <user>` with the password omitted, at a real terminal - check the echo is masked, the connection works, and `.history` has no password in it.

**3.** Noted, thanks.

Both this round and the previous one report no blocking issues and no correctness bugs, so from my side this is ready. Final state: console 119 tests, studio 57, server plugin/backup 55, all green; `lint`, `Codacy`, `Frontend Security Audit`, `Meterian` and `Dependency Review` green.

PR #6857, opened 2026-08-27
fix(#6798): the inactivity timer must not read an unloaded graph as a small one

### LINKED ISSUE #6798: LSM_VECTOR: the inactivity timer repeats the #6784 null-graph conflation, so one insert with no search rebuilds the whole index
#6784 fixed the null-graph conflation on the search path. The inactivity timer reaches the same conflation without searching, and there it is not fixed: a session that writes one vector and then goes idle rebuilds the entire index.

Verified in source at `9c9d024db` and reproduced by the JUnit test below, which passes against current main.

## The conflation

`inactivityRebuildIsWorthIt()` (`LSMVectorIndex.java:8350`) reads a null graph as a small one:

```java
final ImmutableGraphIndex graph = this.graphIndex;
if (graph == null || graph.size() < ASYNC_REBUILD_MIN_GRAPH_SIZE)
  // Small graph (or none yet): a rebuild is cheap regardless of how many mutations are
  // pending, same as rebuildGraphBeforeSearch()'s unconditional synchronous rebuild for this
  // case. Any pending mutation is worth flushing promptly.
  return true;                                                        // :8356
```

and the timer body then picks its arm the same way (`:8416`):

```java
if (graphIndex != null && graphIndex.size() >= ASYNC_REBUILD_MIN_GRAPH_SIZE) {
  startAsyncGraphRebuild();
} else {
  // Small graph: synchronous rebuild on the timer thread.
```

The comment states the assumption exactly: *small graph (or none yet)*. But `graphIndex` is null in a third case neither branch accounts for — the graph is **large and fully persisted, and this session simply has not loaded it**, because loading is lazy on first search. So both the "is it worth it" gate and the sync/async choice treat a 10M-vector index as an empty one. The mutation threshold is bypassed, and `buildGraphFromScratch()` runs on the timer thread over every vector.

#6784 fixed this for a session that **searches**, by loading the persisted graph before `rebuildGraphBeforeSearch()` reads the null. The timer is reached without any search, so that fix never runs on this path.

Ingest-then-idle is the ordinary shape for a loader process: open, write, go quiet. It is also our benchmark harness's write lane, which is where we hit it.

## Reproduction

`InactivityTimerRebuildsWholeIndexTest` (attached below, passes on `9c9d024db`): build 20,000 vectors, close, reopen, insert **one** vector, never search, go idle.

```
graphState            = 0 (LOADING)      <- nothing loaded the graph
graphNodeCount        = 0                <- though a complete graph is on disk
mutationsSinceRebuild = 1                <- threshold is 2,000
                                            (timer fires anyway)
graphNodeCount after  >= 20,000          <- every vector rebuilt
```

The test asserts on `graphRebuildCount` and the resulting graph size rather than on a wall clock, so it states what happened rather than how long it took. It sets the inactivity timeout to 2 s so it does not spend the 15 s default waiting.

<details>
<summary>InactivityTimerRebuildsWholeIndexTest.java</summary>

```java
package com.arcadedb.index.vector;

import com.arcadedb.GlobalConfiguration;
import com.arcadedb.index.TypeIndex;
import com.arcadedb.schema.DocumentType;
import com.arcadedb.schema.Type;
import com.arcadedb.utility.FileUtils;

import org.awaitility.Awaitility;
import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.Tag;
import org.junit.jupiter.api.Test;

import java.io.File;
import java.time.Duration;
import java.util.Map;
import java.util.Random;

import static org.assertj.core.api.Assertions.assertThat;

/**
 * A session that writes one vector and then goes idle WITHOUT searching rebuilds the whole index.
 * <p>
 * The inactivity timer repeats the conflation that #6772 fixed on the search path.
 * {@code inactivityRebuildIsWorthIt()} reads a null graph as a small one:
 * <pre>
 *   final ImmutableGraphIndex graph = this.graphIndex;
 *   if (graph == null || graph.size() &lt; ASYNC_REBUILD_MIN_GRAPH_SIZE)
 *     // "Small graph (or none yet): a rebuild is cheap regardless of how many
 *     //  pending mutations", same as rebuildGraphBeforeSearch()
 *     return true;                                                      // :8356
 * </pre>
 * and the timer body then takes the synchronous arm for the same reason:
 * <pre>
 *   if (graphIndex != null &amp;&amp; graphIndex.size() &gt;= ASYNC_REBUILD_MIN_GRAPH_SIZE)
 *     startAsyncGraphRebuild();
 *   else
 *     buildGraphFromScratch();                                          // :8416 onwards
 * </pre>
 * {@code graphIndex} is null here not because the graph is small or absent, but because nothing in
 * this session loaded it: the graph loads lazily on the first search, and this session never
 * searches. So the mutation threshold is bypassed entirely and one insert rebuilds every vector.
 * <p>
 * #6784 fixed this for a session that SEARCHES, by loading the persisted graph before
 * {@code rebuildGraphBeforeSearch()} reads the null. The timer is reached without a search, so that
 * fix does not apply to it. Ingest-then-idle is the ordinary shape for a loader process.
 * <p>
 * The test asserts on {@code graphRebuildCount} and on the resulting graph size rather than on a
 * wall clock, so it states what happened rather than how long it took.
 */
@Tag("vector")
class InactivityTimerRebuildsWholeIndexTest {
  private static final int    DIMENSIONS = 64;
  private static final int    VECTORS    = 20_000;   // >> ASYNC_REBUILD_MIN_GRAPH_SIZE (1000)
  private static final String DB_PATH    = "./target/databases/InactivityTimerRebuildsWholeIndex";

  @AfterEach
  void cleanUp() {
    GlobalConfiguration.VECTOR_INDEX_INACTIVITY_REBUILD_TIMEOUT_MS.reset();
    FileUtils.deleteRecursively(new File(DB_PATH));
  }

  @Test
  void oneInsertThenIdleRebuildsEveryVector() {
    // 2 s rather than the 15 s default, so the test does not spend the default waiting.
    GlobalConfiguration.VECTOR_INDEX_INACTIVITY_REBUILD_TIMEOUT_MS.setValue(2_000);
    FileUtils.deleteRecursively(new File(DB_PATH));

    try (final DatabaseFactory factory = new DatabaseFactory(DB_PATH)) {
      final Database db = factory.create();
      final Random rng = new Random(13);
      db.transaction(() -> {
        final DocumentType t = db.getSchema().createDocumentType("Doc");
        t.createProperty("id", Type.INTEGER);
        t.createProperty("embedding", Type.ARRAY_OF_FLOATS);
        for (int i = 0; i < VECTORS; i++)
          db.newDocument("Doc").set("id", i).set("embedding", randomVector(rng)).save();
      });
      db.command("sql", "CREATE INDEX ON Doc (embedding) LSM_VECTOR METADATA "
          + "{ \"dimensions\": " + DIMENSIONS + ", \"similarity\": \"EUCLIDEAN\", "
          + "\"storeVectorsInGraph\": false, \"addHierarchy\": true }");
      awaitGraph(db);
      db.close();
    }

    // A NEW session. It writes one vector and never searches, so nothing loads the
    // persisted graph and graphIndex stays null for the whole session.
    try (final DatabaseFactory factory = new DatabaseFactory(DB_PATH)) {
      final Database db = factory.open();
      final LSMVectorIndex index = lsm(db);

      assertThat(index.getStats().get("graphState"))
          .as("nothing has loaded the graph yet, so the index is LOADING(0)").isEqualTo(0L);
      assertThat(index.getStats().get("graphNodeCount"))
          .as("and no graph node is resident, though a complete graph is on disk").isEqualTo(0L);

      final long rebuildsBefore = index.getStats().get("graphRebuildCount");
      final Random rng = new Random(99);
      db.transaction(() -> db.newDocument("Doc").set("id", VECTORS)
          .set("embedding", randomVector(rng)).save());

      final long threshold = index.getStats().get("effectiveMutationsThreshold");
      assertThat(index.getStats().get("mutationsSinceRebuild"))
          .as("exactly one pending mutation, far under the threshold of %d", threshold)
          .isEqualTo(1L);

      // Now go idle. No search, no further write.
      Awaitility.await("the inactivity timer fires")
          .atMost(Duration.ofSeconds(120))
          .pollInterval(Duration.ofMillis(250))
          .untilAsserted(() -> assertThat(index.getStats().get("graphRebuildCount"))
              .isGreaterThan(rebuildsBefore));

      final Map<String, Long> after = index.getStats();
      assertThat(after.get("graphNodeCount"))
          .as("THE DEFECT: one insert, no search, and the timer rebuilt every vector. "
              + "graphIndex was null because nothing had LOADED the graph, not because the "
              + "graph was small, and inactivityRebuildIsWorthIt() reads that null as "
              + "\"a rebuild is cheap\" and bypasses the threshold of %d entirely", threshold)
          .isGreaterThanOrEqualTo((long) VECTORS);
      db.close();
    }
  }

  private static LSMVectorIndex lsm(final Database db) {
    final TypeIndex idx = (TypeIndex) db.getSchema().getIndexByName("Doc[embedding]");
    return (LSMVectorIndex) idx.getIndexesOnBuckets()[0];
  }

  private static void awaitGraph(final Database db) {
    final LSMVectorIndex index = lsm(db);
    for (int i = 0; i < 1200 && index.getStats().get("graphNodeCount") < VECTORS; i++)
      try { Thread.sleep(100); } catch (final InterruptedException ignored) { break; }
  }

  private static float[] randomVector(final Random rng) {
    final float[] v = new float[DIMENSIONS];
    for (int i = 0; i < DIMENSIONS; i++)
      v[i] = (float) rng.nextGaussian();
    return v;
  }
}
```

</details>

## Suggested fix

The same shape as #6784: consult what is on disk, not what this session happens to have loaded. Either load the persisted graph before `inactivityRebuildIsWorthIt()` reads `graphIndex`, or have the null case fall back to the persisted node count so a large-but-unloaded index takes the large-graph path (threshold respected, async arm) instead of the small-graph one.

Distinguishing "no graph exists" from "no graph is loaded" is the actual fix; both call sites above then follow from it. `LOADING` state plus a nonzero persisted count already expresses the difference.

Happy to open a PR if you would like, or to test a patch against the 1M and 10M corpora we have staged.

lvca:
Closes #6798.

## The defect

The graph loads lazily on the first search, so `graphIndex == null` carries three meanings, and only two of them mean *a rebuild here is cheap*:

1. no graph has ever been built - a first build, which must run whatever it costs;
2. a graph was built and is genuinely tiny - cheap by definition;
3. **a large graph is fully persisted and this session has simply never loaded it.**

`inactivityRebuildIsWorthIt()` and the timer task's sync/async arm choice both read the third as the first two. The mutation threshold was therefore skipped entirely, and one insert into a reopened index rebuilt every vector, synchronously, on the timer thread.

Ingest-then-idle - open, write, go quiet - is the ordinary shape of a loader process, and it is the one shape that never loads the graph. This is the same conflation #6772 fixed for a session that *searches*; that fix works by loading the persisted graph inside `ensureGraphAvailable()`, which the timer never calls, so it does not reach this path.

## The fix

Both call sites ask `inactivityRebuildScopeSize()` how many vectors a rebuild would actually have to cover.

The issue suggested either loading the persisted graph before the gate reads it, or falling back to the persisted graph's node count. **This does neither**, and both alternatives are worse here:

- *Loading the graph on the timer* makes an ingest-then-idle process pay for a load and a validation scan it never asked for, on a thread shared with every other index's inactivity task - it defeats the lazy load rather than accounting for it.
- *The persisted node count* is a proxy for the wrong quantity. `buildGraphFromScratch()` covers the **live** vector set, not the graph it replaces, so the live count is the direct measure of the cost being gated. It also gives the right answer where the two diverge: a large persisted graph whose vectors have since been deleted really is cheap to rebuild, and keeps getting flushed promptly instead of being deferred behind a threshold it will never reach.

The new behaviour is confined to the case `graphNotYetMaterialised()` identifies - this session has not yet resolved what is on disk - so everything downstream of a resolution answers exactly as before, including the window inside `buildGraphFromScratchExclusively()` where a rebuild nulls the resident graph before publishing its replacement.

`rebuildGraphBeforeSearch()` is deliberately left reading `graphIndex` directly. `ensureGraphAvailable()` always runs immediately before it (#6772), so a null there really is case 1 or 2 - and routing a search down the large-graph arm would hand it the async rebuild and an **empty** result set instead of a slow but correct one.

New `persistedGraphNodeCount` stat, alongside the existing `graphNodeCount` (which reports what this session has *loaded*), so "no graph exists" and "no graph is loaded" can be told apart from outside. That distinction is what the whole fix rests on.

## Verification

`Issue6798InactivityTimerUnloadedGraphTest` (`@Tag("vector")`) pins three behaviours, because fixing only the first is what broke three existing tests the last time this gate was touched (PR #6510 for #6496):

- one insert into a large-but-unloaded index must rebuild nothing at all - fails on the pre-fix decision, asserting on `graphRebuildCount`/`mutationsSinceRebuild` rather than on a wall clock;
- reaching the threshold-derived floor must still rebuild, so nothing is deferred forever;
- a genuinely small persisted index must still flush on any pending mutation - the over-correction guard.

`mvn -o -pl engine -am test -Dtest='com.arcadedb.index.vector.*Test'`: **434/434 green**, including `Issue6496InactivityRebuildThresholdTest`, `LSMVectorIndexRebuildTest`, `Issue6772WriteBeforeSearchPrefixReuseTest` and `Issue6655StaleGraphPrefixReuseTest`.

## Follow-up, deliberately not in this PR

When the floor *is* legitimately reached on a large-but-unloaded index, the async rebuild still builds from scratch rather than reusing the persisted prefix the way #6655/#6772 do on the search path. That is a separate optimisation (it applies equally to the search path's own async rebuild) and reaching it from the rebuild thread would drag `ensureGraphAvailable()`'s `graphBuildLock` under `startAsyncGraphRebuild()`'s monitor, adding a lock-ordering edge PR #6784's review explicitly avoided. Worth its own issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vector index size reporting by distinguishing persisted data from data currently loaded in memory.
  - Prevented large, lazily loaded indexes from rebuilding prematurely after minimal activity.
  - Rebuilds for affected indexes now run asynchronously, improving timer responsiveness.
  - Preserved immediate flushing behavior for small indexes.

- **Tests**
  - Added coverage for inactivity-timer behavior across unloaded large indexes and small indexes.

### DIFF
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -7526,6 +7526,13 @@ public class LSMVectorIndex implements Index, IndexInternal {
     final LSMVectorIndexGraphManifest.Content manifestContent =
         statsGraphFile != null ? statsGraphFile.getManifest().read() : null;
     stats.put("closeTimeRebuildPending", manifestContent != null && manifestContent.closeDeferredRebuild() ? 1L : 0L);
+    // How large the graph ON DISK is, as opposed to graphNodeCount just above, which reports how much of it this
+    // session has loaded - 0 until something searches, however many vectors are persisted (issue #6798). Read
+    // from the same manifest content: the two are the only stats here that survive a reopen before a first
+    // search, and telling "no graph exists" apart from "no graph is loaded" needs both of them.
+    stats.put("persistedGraphNodeCount",
+        manifestContent != null && manifestContent.vectorCount() > 0
+            && statsGraphFile.hasPersistedGraph() ? (long) manifestContent.vectorCount() : 0L);
 
     // Calculate mutations threshold (use configured value or default)
     final int defaultMutationsThreshold = getDatabase().getConfiguration()
@@ -8344,6 +8351,10 @@ public class LSMVectorIndex implements Index, IndexInternal {
    * away the guarantee that a handful of buffered vectors get flushed out of the linear-scan delta
    * buffer promptly - but to apply the same threshold the search path uses, and only where the search
    * path would also apply it: once the graph is large enough that a rebuild is no longer free.
+   * <p>
+   * "How large is the graph" is asked through {@link #inactivityRebuildScopeSize()} rather than off
+   * {@link #graphIndex} directly, because on this path a null graph does not mean what it means on the search
+   * path (issue #6798).
    *
    * @return {@code true} when pending mutations justify (or exceed) a rebuild
    */
@@ -8352,8 +8363,7 @@ public class LSMVectorIndex implements Index, IndexInternal {
     if (pending <= 0)
       return false;
 
-    final ImmutableGraphIndex graph = this.graphIndex;
-    if (graph == null || graph.size() < ASYNC_REBUILD_MIN_GRAPH_SIZE)
+    if (inactivityRebuildScopeSize() < ASYNC_REBUILD_MIN_GRAPH_SIZE)
       // Small graph (or none yet): a rebuild is cheap regardless of how many mutations are
       // pending, same as rebuildGraphBeforeSearch()'s unconditional synchronous rebuild for this
       // case. Any pending mutation is worth flushing promptly.
@@ -8369,6 +8379,61 @@ public class LSMVectorIndex implements Index, IndexInternal {
     // on every quiet period while still eventually absorbing a trickle of writes.
     final int floor = Math.max(threshold / 10, 1);
     return pending >= floor;
+  }
+
+  /**
+   * How many vectors a rebuild triggered by the inactivity timer would actually have to cover - which is the
+   * question both {@link #inactivityRebuildIsWorthIt()} and the timer task's sync/async arm choice are really
+   * asking when they compare against {@link #ASYNC_REBUILD_MIN_GRAPH_SIZE}.
+   * <p>
+   * They used to ask it of {@link #graphIndex} alone, and {@code graphIndex == null} counted as "small". Null
+   * carries three meanings, though, and only two of them are small (issue #6798):
+   * <ol>
+   *   <li>no graph has ever been built for this index - a first build, which must run whatever it costs, since
+   *       nothing else will ever produce a graph;</li>
+   *   <li>a graph was built and is genuinely tiny (or emptied out by deletions) - cheap by definition;</li>
+   *   <li><b>a large graph is fully persisted and this session has simply never loaded it</b>, because the load
+   *       is lazy on the first search and this session has not searched.</li>
+   * </ol>
+   * The third is the ordinary shape of a loader process - open, write, go quiet - and it was being read as the
+   * first two: the mutation threshold was skipped entirely and one insert rebuilt every vector in the index,
+   * synchronously, on the timer thread. It is the same conflation issue #6772 fixed for a session that searches;
+   * that fix works by loading the persisted graph inside {@link #ensureGraphAvailable()}, which the timer never
+   * calls, so it does not reach here.
+   * <p>
+   * The answer for that third case is the <b>live</b> vector count rather than the persisted graph's node count,
+   * because {@link #buildGraphFromScratch()} covers the live set, not the graph it replaces - so the live count
+   * is the direct measure of the cost being gated, where a graph size is only ever a proxy for it. It also gives
+   * the right answer where the two diverge: a large persisted graph whose vectors have since been deleted really
+   * is cheap to rebuild, and gets flushed promptly like any other small index.
+   * <p>
+   * Deliberately confined to the case {@link #graphNotYetMaterialised()} identifies - this session has not yet
+   * resolved what is on disk - so that everything downstream of a resolution keeps answering exactly as before,
+   * including the window inside {@code buildGraphFromScratchExclusively()} where a rebuild nulls the resident
+   * graph before publishing its replacement. {@link #rebuildGraphBeforeSearch()} is left reading
+   * {@link #graphIndex} directly for a related reason: {@link #ensureGraphAvailable()} always runs immediately
+   * before it, so a null there really is case 1 or 2, and routing a search down the large-graph arm would hand
+   * it the async rebuild and an empty result set instead of a slow but correct one.
+   *
+   * @return the number of vectors a rebuild would build over, or 0 when that is a first build
+   */
+  private int inactivityRebuildScopeSize() {
+    final ImmutableGraphIndex resident = this.graphIndex;
+    if (resident != null)
+      return resident.size();
+
+    if (!graphNotYetMaterialised())
+      // Already resolved: nothing is on disk that this session has not accounted for, so a null graph here is a
+      // first build or an emptied index, and both are cheap.
+      return 0;
+
+    final LSMVectorIndexGraphFile gf = graphFile;
+    if (gf == null || !gf.hasPersistedGraph())
+      return 0; // Case 1: no graph anywhere. A first build must not be gated - see the javadoc above.
+
+    // Case 3. getActiveCount() is O(allocated chunks) over an already-materialised location index: this method is
+    // only ever reached with pending mutations outstanding, and recording those materialised it.
+    return (int) Math.min(vectorIndex().getActiveCount(), Integer.MAX_VALUE);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lvca:
The deferred prefix-reuse follow-up is now filed as #6859, with the lock-ordering constraint from #6784's review and the termination question a fix has to answer first.

Keeping the `inactivityRebuildScopeSize()` javadoc at its current length: it is the third bug on this exact gate (#6496, #6772, this one), and each was a misreading of what a null graph means rather than a coding error, so the history is the part worth carrying.

Issue #5918, opened 2026-08-07
Bolt PackStream parser: unbounded pre-auth allocation (5-byte OOM) and unbounded recursion (stack overflow), both escaping catch(Exception)


ruispereira:
## Summary

The Bolt PackStream parser allocates from an unbounded, attacker-controlled 32-bit length and recurses without a depth limit, both **before authentication**. A 5-byte message forces a multi-gigabyte allocation (JVM-wide `OutOfMemoryError`); a stream of nesting markers overflows the stack. Both are `Error`s that escape the executor's `catch (Exception)` handler.

This is a **different layer** from #5894 (which is the WebSocket framing in `BoltWebSocketInputStream`): these defects are in `PackStreamReader`, the value decoder that runs on every Bolt message on both the raw-TCP and WebSocket transports.

Version: 26.8.1. Module `bolt`.

## 1. Unbounded allocation from a 32-bit length (pre-auth OOM)

```java
// bolt/src/main/java/com/arcadedb/bolt/packstream/PackStreamReader.java
case BYTES_32:  final int length = in.readInt(); return readBytes(length);      // ~176 → 267
case STRING_32: final int length = in.readInt(); return readStringBytes(length); // ~158 → 258
case LIST_32:   final int size   = in.readInt(); return readListItems(size);     // ~200 → 276
case MAP_32:    final int size   = in.readInt(); return readMapEntries(size);     // ~224 → 287

private byte[] readBytes(final int length) { final byte[] bytes = new byte[length]; ... }   // no bound
private List<Object> readListItems(final int size) { new ArrayList<>(size); ... }            // no bound
private Map<String,Object> readMapEntries(final int size) { new LinkedHashMap<>(size); ... }  // no bound
```

The declared length/size is used to eagerly allocate, with no check against the remaining input.

**Minimal trigger:** 5 bytes — `CE 7F FF FF FF` (BYTES_32, length = 2147483647).

## 2. Unbounded recursion (pre-auth stack overflow)

`readValue()` → `readListItems`/`readMapEntries`/`readStructure` → `readValue()`, one frame per nesting level, no depth cap.

**Minimal trigger:** N bytes of `91` (TINY_LIST, size 1) then `C0` (NULL). Message size ≈ recursion depth.

## Pre-authentication reachability (verified)

`BoltNetworkExecutor.run()` sets `state = State.AUTHENTICATION` and then, in its message loop, decodes **before** authenticating:

```java
[...]

PR #6864, opened 2026-08-28
fix(bolt): #6800 #6801 #6802 #6803 #6804 protocol hardening batch

### LINKED ISSUE #6800: PackStream LIST_16 element count skips the #5918 size guards: a 3 KB pre-auth message allocates ~256 MB
### Summary
The PackStream size hardening added for #5918 validates only the 32-bit container size fields. `LIST_8` and `LIST_16` bypass it, and `ListFrame` allocates its backing array eagerly from the declared size, so a ~3 KB pre-authentication message can force ~256 MB of live heap.

### Where (HEAD `9c9d024db`)
`bolt/src/main/java/com/arcadedb/bolt/packstream/PackStreamReader.java:310-319` - `LIST_32` is validated, `LIST_16` is not:
```java
// LIST_16
if (marker == (LIST_16 & 0xFF)) {
  final int size = in.readUnsignedShort();
  return openList(size, stack);                          // <- unchecked
}

// LIST_32
if (marker == (LIST_32 & 0xFF)) {
  final int size = in.readInt();
  return openList(checkElementCount(size, "LIST_32"), stack);
}
```
`checkElementCount` is the only place that rejects a declared size above `maxElements` **or above `in.available()`**, and it is invoked at exactly two sites (`LIST_32`, `MAP_32`).

`bolt/src/main/java/com/arcadedb/bolt/packstream/PackStreamReader.java:451-459` - the frame allocates up front:
```java
ListFrame(final int size) {
  this.list = new ArrayList<>(size);   // ArrayList(int) allocates new Object[initialCapacity] eagerly
  this.remaining = size;
}
```
`maxDepth` (default 1000) bounds only the number of such frames, and all of them are live at once on the explicit frame stack.

### Repro
The message loop (`BoltNetworkExecutor.java:214-235`) decodes before HELLO/LOGON, so no credentials are needed. Send one Bolt chunk consisting of 1001 repetitions of the three bytes `D5 FF FF`:

- each level declares a 65535-element list, so `ListFrame` allocates `Object[65535]` (~256 KiB with compressed oops) with zero payload behind it;
- at depth 1001 `readValue()` finally throws the depth error, but ~1001 x 256 KiB is already live: roughly **256 MB of heap from ~3 KB on the wire** (~87000x amplification), comfortably inside the 16 MB `BOLT_MAX_MESSAGE_SIZE`;
- the exception is caught at `BoltNetworkExecutor.java:252` and the connection is **not** closed, so the same socket can repeat this indefinitely, and `NETWORK_MAX_PREAUTH_CONNECTIONS` sockets can do it in parallel.

The identical declaration sent as `LIST_32` is rejected, which is what makes this a gap rather than a design choice. `PackStreamBoundsTest.deepNesting*` only nests `TINY_LIST` (0x91), whose frame allocates `Object[1]`, so the amplification is invisible to the existing tests.

`MAP_8`/`MAP_16` are genuinely benign by comparison: `new LinkedHashMap<>(size)` only sets the threshold and allocates the table on first `put`.

### Suggested fix
Route the 8/16-bit container sizes through the same validator as the 32-bit ones, i.e. `checkElementCount(size, "LIST_16")` at line 311 (and on `LIST_8`/`MAP_8`/`MAP_16` for symmetry), so the `size > in.available()` floor applies to every declared element count. Alternatively make `ListFrame` lazy (`new ArrayList<>()`), so a declared size never sizes an array before any element has been read. Both would be worth doing.

### Scope / confidence
High. Verified at HEAD: the unchecked path, the eager allocation, the two `checkElementCount` call sites, and that the connection survives the resulting exception.

### LINKED ISSUE #6801: HELLO's plaintext password is written to the log by HelloMessage.toString() when arcadedb.bolt.debug=true
### Summary
When BOLT protocol debug is enabled, every inbound message is logged via `toString()`. `HelloMessage.toString()` prints the whole `extra` map, which for every Bolt version below 5.1 carries the caller's **cleartext password**. The sibling `LogonMessage.toString()` deliberately omits it, so this is an inconsistency rather than a policy.

### Where (HEAD `9c9d024db`)
`bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java:231-233`:
```java
LogManager.instance().log(this, Level.FINE, "BOLT << %s (state=%s)", message, state);
```
`bolt/src/main/java/com/arcadedb/bolt/message/HelloMessage.java:73-75`:
```java
public String toString() {
  return "HELLO{extra=" + extra + "}";
}
```
`bolt/src/main/java/com/arcadedb/bolt/message/LogonMessage.java:61-63` - the intended shape:
```java
public String toString() {
  return "LOGON{scheme=" + getScheme() + ", principal=" + getPrincipal() + "}";
}
```
There is no redaction helper anywhere in `bolt/src/main`.

### Repro
Set `arcadedb.bolt.debug=true` ("Enables the printing of BOLT protocol to the console", `GlobalConfiguration.java:2057`) and connect with any driver negotiating Bolt <= 5.0. The server advertises 4.4, 4.0 and 3.0 (`BoltNetworkExecutor.java:106-109`) and `deferAuthToLogon` only defers auth for >= 5.1, so `handleHello` authenticates straight from the HELLO fields (`:542`). The log line is:

```
BOLT << HELLO{extra={user_agent=..., scheme=basic, principal=root, credentials=<the password>}}
```

Enabling protocol debug to troubleshoot a connection problem is exactly the situation where this fires, and the value then flows into whatever log pipeline is configured.

### Suggested fix
Give `HelloMessage.toString()` the same treatment as `LogonMessage.toString()`: print `user_agent`, `scheme`, `principal` and the non-auth `extra` keys, and render `credentials` as `***` (or omit it).

### Scope / confidence
High. One-line fix, verified at HEAD including that pre-5.1 HELLO auth is a live path and that nothing redacts upstream.

### LINKED ISSUE #6802: Bolt-over-WebSocket silently discards continuation frames, truncating any fragmented message
### Summary
The Bolt-over-WebSocket transport never inspects the FIN bit and has no case for opcode `0x0`. A fragmented WebSocket message therefore has its continuation frames read, unmasked, and then **discarded**, so the dechunker sees a truncated byte stream instead of an error.

### Where (HEAD `9c9d024db`)
`bolt/src/main/java/com/arcadedb/bolt/BoltWebSocketInputStream.java:74-119`:
```java
final int b0 = in.readUnsignedByte();
final int b1 = in.readUnsignedByte();

final int opcode = b0 & 0x0F;        // FIN bit (b0 & 0x80) never read
...
final byte[] payload = new byte[(int) payloadLen];
if (payloadLen > 0)
  in.readFully(payload);
...
switch (opcode) {
case 0x1: // text frame
case 0x2: // binary frame
  buffer = payload; bufferPos = 0; bufferLen = payload.length;
  return true;
case 0x8: // close frame
  closed = true; return false;
case 0x9: // ping - skip (pong requires output access)
case 0xA: // pong - skip
default:
  continue;                          // <- opcode 0x0 (continuation) lands here, payload dropped
}
```
The two skipped control opcodes carry a comment explaining why; `0x0` is unmentioned and only reachable through `default`, which reads as an oversight rather than a decision. Nothing in `completeWebSocketUpgrade` (`BoltNetworkExecutor.java:1916-1940`) negotiates away fragmentation, and RFC 6455 permits it unconditionally.

### Repro
Over the Bolt-on-WebSocket transport (the Neo4j Browser/Desktop path, `BoltNetworkExecutor.java:381-383`), send any Bolt message split as `[FIN=0, op=0x2, part1][FIN=1, op=0x0, part2]`. `part2` is dropped rather than appended, so depending on where the split falls `BoltChunkedInput` blocks on a chunk terminator that never arrives, `PackStreamReader` throws "Unknown PackStream marker", or - worst case - trailing bytes are reinterpreted as the next message's chunk header. Any client stack that fragments large sends (a big parameter map, a long query string) hits this.

`BoltWebSocketInputStreamTest` has no fragmentation case: all four tests build single `0x82` (FIN + binary) frames.

### Suggested fix
Track fragmentation in `readNextFrame()`: on a data frame with FIN=0 keep accumulating, on opcode `0x0` append to the accumulated buffer, and return only once FIN=1 (applying `maxFrameSize` to the accumulated total). If reassembly is out of scope for now, at minimum throw `IOException("fragmented BOLT WebSocket frames are not supported")` instead of silently dropping the payload.

### Scope / confidence
High. Verified at HEAD: FIN is never read, `0x0` has no case, and the payload is fully consumed before being discarded.

### LINKED ISSUE #6803: LOGOFF is accepted in any state and leaves the open result set and transaction behind
### Summary
`handleLogoff()` is the only request handler with neither state validation nor cleanup. A LOGOFF sent while a result stream or an explicit transaction is open answers SUCCESS and leaves both pinned, on a connection the server now considers unauthenticated.

### Where (HEAD `9c9d024db`)
`bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java:586-591`:
```java
private void handleLogoff() {
  user = null;
  markUnauthenticated();
  sendSuccess(Map.of());
  state = State.AUTHENTICATION;
}
```
Every sibling guards its state (`handleRun` `:655`, `handlePull` `:779`, `handleCommit` `:982`, ...), and `handleReset` (`:603-644`) explicitly closes `currentResultSet` and rolls back `explicitTransaction`. The dispatcher calls `handleLogoff()` unconditionally (`:457-459`), with no `state == FAILED` short-circuit either, so even a FAILED session gets SUCCESS instead of IGNORED.

### Repro
1. BEGIN
2. RUN a query and leave rows unconsumed, so the state is TX_STREAMING
3. LOGOFF -> SUCCESS

`currentResultSet` (with its cursors), the ArcadeDB transaction, `database`, `currentFields` and `firstResult` all survive untouched for the remaining life of the connection, which is now unauthenticated by the server's own accounting. A subsequent HELLO/LOGON returns to READY with `explicitTransaction` still true, so the next RUN/PULL lands in TX_STREAMING -> TX_READY and a COMMIT then commits writes made **before** the user change.

Bolt 5.1 specifies LOGOFF as valid only from READY, which is precisely the state where no stream or transaction is open. `cleanup()` does compensate, but only when the connection finally terminates.

The only LOGOFF coverage today is in `BoltMessageTest` (`:328-329`, `:617-624`, `:774`), all pure parse/toString assertions with no executor state machine involved.

### Suggested fix
Reject LOGOFF outside `State.READY` with `sendFailure(PROTOCOL_ERROR, ...)`, matching the other handlers, and reuse `handleReset`'s cleanup (close `currentResultSet`, roll back when `explicitTransaction`, clear `currentFields`/`firstResult`/`syntheticResults`/`currentPlanMetadata`) before flipping to AUTHENTICATION.

### Scope / confidence
High on the mechanism. The cross-user commit is the part worth prioritising.

### LINKED ISSUE #6804: RUN in TX_STREAMING is rejected as a protocol error, so a second query in one transaction fails the session (qid is parsed but never used)
### Summary
`handleRun` accepts RUN only from READY and TX_READY, but the Bolt 4.0+ state machine lists RUN as valid in TX_STREAMING too. That is the whole point of the `qid` field, which lets a client hold several result streams inside one explicit transaction. The server parses `qid` and never uses it.

### Where (HEAD `9c9d024db`)
`bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java:655-659`:
```java
if (state != State.READY && state != State.TX_READY) {
  sendFailure(PROTOCOL_ERROR, "RUN not expected in state: " + state);
  state = State.FAILED;
  return;
}
```
`PullMessage.getQid()` (`:55-65`) and `DiscardMessage.getQid()` (`:55-65`) exist, but `grep -rn qid bolt/src/main` shows no call site outside the messages' own `toString()`. `handlePull`/`handleDiscard` operate unconditionally on the single `currentResultSet` field.

### Repro
With any Neo4j driver, use a fetch size smaller than the first query's row count so the first stream stays open, then run a second query in the same transaction:
```java
Config.builder().withFetchSize(1)
...
try (Transaction tx = session.beginTransaction()) {
  tx.run("UNWIND [1,2,3] AS x RETURN x");   // leaves the server in TX_STREAMING (has_more=true)
  tx.run("RETURN 1");                       // FAILURE Neo.ClientError.Request.Invalid
}
```
The second RUN gets `"RUN not expected in state: TX_STREAMING"`, the session goes to FAILED and the transaction is lost.

This passes today only because the ITs never leave a stream open: `BoltProtocolIT.largeResultSet` (`:450-459`) issues 100 `tx.run` calls in one transaction, but each returns 0 rows, so the driver's eager PULL completes every stream before the next RUN.

### Suggested fix
Handle RUN from STREAMING/TX_STREAMING. Minimal, single-stream-faithful version: drain and close `currentResultSet` before starting the new query. Spec-complete version: a small `qid -> (ResultSet, fields, plan metadata)` map keyed by an incrementing query id returned in the RUN SUCCESS metadata, with `handlePull`/`handleDiscard` selecting on `getQid()` instead of the single field.

### Scope / confidence
Medium. The rejection and the unused `qid` are certain; how often a given driver leaves a stream open before the next RUN depends on its fetch-size configuration, so the practical hit rate varies.

lvca:
Closes #6800, closes #6801, closes #6802, closes #6803, closes #6804.

Five BOLT-module issues, all reported against `9c9d024db`. Every one was real.

## #6800 - PackStream LIST_16 skips the #5918 size guards (security)

`checkElementCount` was invoked from exactly two sites, `LIST_32` and `MAP_32`, so every narrower container marker was unvalidated. `ListFrame` then sized its backing array from the declared count, so 1001 repetitions of the three bytes `D5 FF FF` - each declaring a 65535-element list with nothing behind it - allocated ~1001 x `Object[65535]`, all live at once on the explicit frame stack, before `maxDepth` finally fired. ~3 KB on the wire, ~256 MB of heap, no credentials needed, and the connection survives the exception so the same socket can repeat it.

Both remedies the report suggests, because they close different halves of it:

- every declared length/size now goes through the same validator, so the exact `size > in.available()` floor (a list element needs at least one wire byte) applies at every width - `TINY_*`, `*_8`, `*_16` as well as `*_32`;
- `ListFrame`, `MapFrame` and `StructFrame` grow from the elements actually read instead of pre-sizing, so a client-declared count can no longer size an allocation on its own. `MapFrame` is included even though `new LinkedHashMap<>(n)` does not allocate its table up front: it remembers `n` as the threshold, so the first `put` allocates a `Node[]` sized for the declared count, which under the 16MB message ceiling is still an order-of-magnitude amplification.

## #6801 - HELLO's cleartext password reaches the log (security)

`arcadedb.bolt.debug=true` logs every inbound message through `toString()`, and for every BOLT version below 5.1 the HELLO extra map carries the caller's password. `HelloMessage.toString()` now renders `credentials=***` and leaves everything else - the diagnostics the flag was turned on for - intact, matching `LogonMessage.toString()`.

## #6802 - Bolt-over-WebSocket discards continuation frames

The transport never read the FIN bit and had no case for opcode `0x0`, so a fragmented message had its continuations read, unmasked and dropped through `default`. Depending on where the split fell, the dechunker blocked on a terminator that never arrived, `PackStreamReader` threw "Unknown PackStream marker", or trailing bytes were reinterpreted as the next message's chunk header.

Fragments are reassembled now (`FIN=0` data frame + `0x0` continuations, delivered on `FIN=1`), with `maxFrameSize` applied to the accumulated total so a client cannot fragment its way past the cap one legal-looking frame at a time. Control frames are validated per RFC 6455 5.5 (never fragmented, at most 125 bytes) from the header alone and may still be interleaved between fragments without disturbing reassembly, and a reserved opcode now fails the connection instead of being silently skipped - silent skipping is what caused this bug in the first place.

## #6803 - LOGOFF accepted in any state, leaving stream and transaction behind

`handleLogoff()` was the one request handler with neither a state check nor cleanup. Sent while a stream or an explicit transaction was open it answered SUCCESS and left both pinned on a connection the server had just stopped counting as authenticated; a following HELLO/LOGON as a different user returned to READY with `explicitTransaction` still true, and its COMMIT committed writes made **before** the user change. LOGOFF is now valid only from READY (BOLT 5.1), IGNORED when FAILED like every sibling, and releases whatever is open before flipping to AUTHENTICATION.

## #6804 - RUN rejected in TX_STREAMING, qid parsed and never used

The report offers a minimal single-stream version (drain and close the open result set before the new query) and a spec-complete one. This PR implements the spec-complete one: silently draining a stream the client has not consumed loses rows it is entitled to, which is a worse failure than the one being fixed.

Per-query state - result set, fields, buffered first row, synthetic rows, timings, write flag, plan metadata - moves out of the connection-wide fields into a new `BoltQueryStream` keyed by qid:

- RUN is accepted in TX_STREAMING and returns its `qid` in the SUCCESS metadata (inside an explicit transaction only, as a Neo4j server does; qids restart at 0 per transaction);
- PULL/DISCARD resolve the stream by `getQid()`, with -1 meaning the most recently opened one;
- the session leaves TX_STREAMING only when the last open stream is drained or discarded;
- RUN stays rejected in auto-commit STREAMING - the BOLT state machine does not list it there, and outside a transaction there is nothing to multiplex onto.

Two guards came with it: open streams are capped per connection (each pins an engine ResultSet for the life of the transaction and nothing obliges a client to consume one), and a RUN that fails now releases the result set it had already opened instead of leaving it for the eventual RESET.

## Two things that fell out of the refactor

- The EXPLAIN/PROFILE plan's `identifiers` were read from the connection-wide `currentFields` at build time, which is *before* `extractFieldNames` runs - so the plan carried the **previous** query's column list (or an empty one on the first query). It is now filled in from the stream's own fields.
- `recordsStreamed` was incremented in four places and read in none. Removed rather than carried into the new per-stream class.

The WebSocket test helper also had a latent bug: `payload.length >>> 56` on an `int` shifts modulo 32, so every frame it built over 125 bytes carried a garbage 64-bit length. The existing tests passed only because they all expected a rejection anyway. It now emits the 16-bit form RFC 6455 prescribes for that range.

## Verification

New tests, each confirmed red against the pre-fix sources and green after:

- `Issue6800PackStreamContainerBoundsTest` (9 tests) - the report's exact `D5 FF FF` payload plus LIST_8 / TINY_LIST / MAP_16 / STRING_16 / BYTES_8 / TINY_STRUCT, and that legitimate wide containers and empty ones at the end of a message are untouched. 7 of 9 fail without the fix.
- `Issue6801HelloCredentialsRedactionTest` (5 tests) - 2 fail without the fix.
- `BoltWebSocketInputStreamTest` - 10 new fragmentation/control-frame cases. All 10 fail without the fix (4 as EOF from the truncated stream).
- `BoltStateMachineIT` (6 tests) - driven over a raw socket rather than through a driver, so the assertions are on the protocol rather than on a driver's fetch-size configuration. Two interleaved streams pulled by qid, DISCARD by qid, an unknown qid, RUN still rejected in auto-commit STREAMING, LOGOFF rejected mid-stream with the write rolled back, LOGOFF from READY still allowing re-authentication. 4 of 6 fail without the fix.

Full `bolt` module green: 343 unit tests, 125 integration tests, 0 failures.

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple concurrent query result streams within explicit transactions.
  * Added query-specific pull and discard controls.
  * Added a configurable per-connection limit for open result streams.
  * Enabled safer logoff, rollback, cleanup, and re-authentication workflows.

* **Bug Fixes**
  * Improved fragmented WebSocket message handling and invalid-frame detection.
  * Added validation for oversized or malformed PackStream data.
  * Credentials are now redacted in diagnostic output.
  * Prevented database switching during open transactions.

* **Tests**
  * Added coverage for transactions, streaming, WebSocket validation, input limits, and credential redaction.

### DIFF
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -1100,6 +1100,7 @@ public class BoltNetworkExecutor extends Thread {
         database.commit();
       }
       explicitTransaction = false;
+      nextQid = 0; // the numbering is per transaction; BEGIN re-zeroes it too, this just keeps it uniform
 
       final Map<String, Object> metadata = new LinkedHashMap<>();
       metadata.put("bookmark", generateBookmark());
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -641,9 +641,14 @@ public class BoltNetworkExecutor extends Thread {
    * - means the most recently opened stream; any other value names one explicitly, which is how a client
    * interleaves several streams inside one transaction. A qid naming nothing fails the session, exactly as a
    * PULL with no result set behind it did before.
+   * <p>
+   * Outside an explicit transaction the qid is ignored altogether: only one stream can be open there (RUN is not
+   * valid in STREAMING), no qid is ever published in an auto-commit RUN SUCCESS, and the BOLT docs are explicit
+   * that qid does not apply to auto-commit. Honouring it would let a client that sent one anyway be answered
+   * "No active result set for qid ..." for the stream it is plainly asking about.
    */
   private BoltQueryStream resolveStream(final long qid) throws IOException {
-    final BoltQueryStream stream = qid < 0 ? currentStream : openStreams.get(qid);
+    final BoltQueryStream stream = qid < 0 || !explicitTransaction ? currentStream : openStreams.get(qid);
     if (stream == null) {
       sendFailure(BoltException.PROTOCOL_ERROR, qid < 0 ? "No active result set" : "No active result set for qid " + qid);
       state = State.FAILED;
@@ -678,8 +683,11 @@ public class BoltNetworkExecutor extends Thread {
         currentStream = open;
     }
 
-    if (openStreams.isEmpty())
+    if (openStreams.isEmpty()) {
       state = explicitTransaction ? State.TX_READY : State.READY;
+      if (!explicitTransaction)
+        nextQid = 0; // an auto-commit stream is always qid 0: the numbering restarts per transaction
+    }
   }
 
   /**
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -766,6 +766,19 @@ public class BoltNetworkExecutor extends Thread {
 
     // Get database from message extra if specified
     final String db = message.getDatabase();
+
+    // An explicit transaction is bound to the database BEGIN opened it on. A later RUN in the same transaction
+    // naming a different one would make ensureDatabase() drop the reference to that database - it nulls it to
+    // force the switch - and with it the only handle a rollback has, stranding the transaction BEGIN opened if
+    // the new database then fails to open. Neo4j drivers fix the database at BEGIN, so a conforming client
+    // never reaches this.
+    if (explicitTransaction && isDatabaseSwitch(db)) {
+      sendFailure(BoltException.PROTOCOL_ERROR,
+          "Cannot switch database to '" + db + "' inside an open transaction started on '" + database.getName() + "'");
+      state = State.FAILED;
+      return;
+    }
+
     if (db != null && !db.isEmpty()) {
       databaseName = db;
     }
@@ -1214,13 +1227,22 @@ public class BoltNetworkExecutor extends Thread {
   /**
    * Ensure database is open and accessible.
    */
+  /**
+   * Whether {@code requestedDatabase} names a database other than the one this connection currently has open.
+   * {@code "system"} and {@code "neo4j"} are Neo4j virtual names that map to this connection's current (default)
+   * database, and an absent name means "keep the current one", so neither counts as a switch. Shared between the
+   * in-transaction guard in {@link #handleRun} and {@link #ensureDatabase()} so the two cannot drift apart.
+   */
+  private boolean isDatabaseSwitch(final String requestedDatabase) {
+    return database != null && database.isOpen()
+        && requestedDatabase != null && !requestedDatabase.isEmpty()
+        && !"system".equals(requestedDatabase) && !"neo4j".equals(requestedDatabase)
+        && !database.getName().equals(requestedDatabase);
+  }
+
   private boolean ensureDatabase() throws IOException {
     if (database != null && database.isOpen()) {
-      // Check if we need to switch to a different database
-      final String currentDbName = database.getName();
-      if (databaseName != null && !databaseName.isEmpty()
-          && !"system".equals(databaseName) && !"neo4j".equals(databaseName)
-          && !currentDbName.equals(databaseName)) {
+      if (isDatabaseSwitch(databaseName)) {
         // Database name changed, need to switch
         database = null;
       } else {
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -134,14 +134,6 @@ public class BoltNetworkExecutor extends Thread {
   private final boolean             debug;
   private final BoltNetworkListener listener; // For notifying when connection closes
 
-  /**
-   * Ceiling on result streams held open at once by one connection. Every one of them pins an engine ResultSet
-   * (cursors, pages) for as long as its transaction lives, and nothing in the protocol obliges a client ever to
-   * consume one, so an unbounded map here is a resource leak a single authenticated session could drive. No real
-   * driver holds more than a handful open.
-   */
-  private static final int MAX_OPEN_STREAMS = 1024;
-
   private State              state = State.DISCONNECTED;
   private int                protocolVersion;
   private ServerSecurityUser user;
@@ -677,7 +669,9 @@ public class BoltNetworkExecutor extends Thread {
 
     if (currentStream == stream) {
       // Fall back to the newest stream still open, so a following qid-less PULL/DISCARD keeps meaning "the last
-      // one opened". The scan is over a map that holds a handful of entries in the worst realistic case.
+      // one opened". This relies on openStreams being a LinkedHashMap: the last value the iteration yields is
+      // the most recently inserted one. The scan is over a map holding a handful of entries in the worst
+      // realistic case.
       currentStream = null;
       for (final BoltQueryStream open : openStreams.values())
         currentStream = open;
@@ -757,9 +751,15 @@ public class BoltNetworkExecutor extends Thread {
       return;
     }
 
-    if (openStreams.size() >= MAX_OPEN_STREAMS) {
+    // Every open stream pins an engine ResultSet (cursors, pages) for as long as its transaction lives, and
+    // nothing in the protocol obliges a client ever to consume one, so an unbounded map here is a resource leak a
+    // single authenticated session could drive. Read per RUN, like the other BOLT protocol limits, so a runtime
+    // change to the setting takes effect on the next message; floored at 1 since a ceiling of 0 would reject
+    // every query outright rather than lock anything down.
+    final int maxOpenStreams = Math.max(1, GlobalConfiguration.BOLT_MAX_OPEN_STREAMS.getValueAsInteger());
+    if (openStreams.size() >= maxOpenStreams) {
       sendFailure(BoltException.PROTOCOL_ERROR,
-          "Too many result streams open at once (max " + MAX_OPEN_STREAMS + "): consume or discard one first");
+          "Too many result streams open at once (max " + maxOpenStreams + "): consume or discard one first");
       state = State.FAILED;
       return;
     }
--- a/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
+++ b/bolt/src/main/java/com/arcadedb/bolt/BoltNetworkExecutor.java
@@ -163,6 +163,13 @@ public class BoltNetworkExecutor extends Thread {
   /** Next qid to hand out, numbered from 0 per explicit transaction as a Neo4j server does. */
   private long nextQid;
 
+  /**
+   * The value {@code PullMessage}/{@code DiscardMessage} report when the client omits the qid, meaning "the
+   * stream opened last". It is the only sentinel: any other negative value is malformed and must fail the lookup
+   * rather than quietly act on whichever stream happens to be current.
+   */
+  private static final long OMITTED_QID = -1;
+
   public BoltNetworkExecutor(final ArcadeDBServer server, final Socket socket, final BoltNetworkListener listener) {
     this(server, socket, listener, null);
   }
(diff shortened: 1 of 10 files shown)


lvca:
Latest review round: two points fixed, two declined. Reasons for the declines, so they are not silently dropped.

**Fixed**
- `resolveStream()` treated any negative qid as the "last opened" sentinel, so a malformed `qid = -2` acted on the current stream instead of failing. Only `-1` is the sentinel now, with a regression test for PULL and for DISCARD.
- The IT's response reader decoded IGNORED as a single-field structure; it carries no field, so it would have EOF'd and buried the real assertion the day a test provokes one.
- Related: a rollback that throws during teardown is now logged at WARNING rather than FINE.

**Declined: making RESET fail when the rollback throws**
RESET is the client's only way out of `FAILED`. If RESET can itself answer FAILURE, a connection that hits a rollback problem has no recovery path left and is simply dead - the escape hatch has to be infallible, which is why a Neo4j server always answers SUCCESS after clearing state. The pre-existing behaviour (release everything, log, carry on) is the right one; what was actually worth fixing is that the failure was invisible, hence the WARNING above.

**Declined: honouring `n` in DISCARD**
Correct as an observation - DISCARD ignores `n` and always drains - but this predates the PR and is orthogonal to all five issues it closes. Implementing bounded discard changes the `has_more` contract on a path every driver currently reaches with `n = -1` only (`consume()`), so it deserves its own issue and its own regression tests rather than riding along here. Happy to file it as a follow-up.

**Also declined earlier**: case-insensitive `"system"`/`"neo4j"` matching in `isDatabaseSwitch()`. ArcadeDB database names are case-sensitive filesystem names, so `equalsIgnoreCase` would make a real database named `System` silently resolve to the default one - a real hazard traded for a hypothetical one.

Full bolt module green with `-am`: 344 unit tests, 131 integration tests.

PR #6865, opened 2026-08-28
fix(importer): ZIP silently empty, clobbered delimiter, skipped-record resurrection, XML sibling carryover, long-text typed LONG (#6810 #6811 #6812 #6813 #6814)

### LINKED ISSUE #6810: Importing any ZIP source silently yields 0 records: the reset callback positions the old (closed) ZipInputStream, not the new one

### LINKED ISSUE #6811: SourceDiscovery overwrites the user's delimiter option with null for every CSV source, so a non-comma CSV cannot be imported via -url / IMPORT DATABASE

### LINKED ISSUE #6812: JSON import saves a record it just logged as skipped (duplicate @id, missing @cat/@type) as an anonymous Document instead

### LINKED ISSUE #6813: XML import: an empty or self-closing sub-element inherits the previous sibling's text value

### LINKED ISSUE #6814: AnalyzedProperty treats a value longer than 100 chars as no evidence instead of disproving LONG/DOUBLE, so the column is typed numeric and the import then fails on that value

lvca:
Fixes #6810, fixes #6811, fixes #6812, fixes #6813, fixes #6814.

Five independent defects in the importer module, all of which either drop or corrupt imported data without the import ever reporting a failure. Each one has a regression test that was verified to fail against `main` before the fix.

## #6810 - any ZIP source imports as 0 records, reported as success

Both reset callbacks in `SourceDiscovery` called `getNextEntry()` on the **old**, already closed `ZipInputStream` instead of the freshly created one, so the new stream was left with no current entry. A `ZipInputStream` with no current entry reads as an empty file rather than throwing, and the reset path runs on every import (`Importer.loadFromSource`, `SourceDiscovery.getSchema`, `CSVImporterFormat.analyze`). `Source.reset()` only logged the resulting `Stream closed` at SEVERE, so the run finished with `Total documents: 0` and no error.

Beyond the suggested fix in the issue:

- the new stream is positioned through the **same** seek the initial open uses, extracted as `seekZipEntry`/`positionZipStream`, so the `:::<resource>` entry is honoured on reset too (it never was - reset landed on the first entry of the archive regardless of what the user asked for);
- `Source.reset()` propagates instead of swallowing, so a broken re-open can never again be mistaken for an empty input;
- the remote reset callback closes the old stream before disconnecting the connection it belongs to. The local callback already did; the URL one leaked it, and on a ZIP source it then went on to advance that leaked stream instead of the new one, which is the defect itself;
- the local reset now re-opens through the same file-then-classpath resolution the first open used. It previously hard-coded `new FileInputStream(file)`, which cannot work for the classpath fallback - a latent second way for reset to fail, and one that the newly propagating `reset()` would have turned into a hard error.

## #6811 - the user's delimiter is overwritten with null for every CSV source

`analyzeSourceContent()` unconditionally wrote `knownDelimiter` into `settings.options`, but left it `null` for `EntityType.DATABASE`. Since `-delimiter` / `WITH delimiter = ';'` lands in that same map (`ImporterSettings.parseParameter`), it was destroyed for every CSV source, and `IMPORT DATABASE file://data.csv WITH delimiter = ';'` parsed the whole line as one column - producing a property literally named `id;name;age`, and reporting success.

The put is now guarded, and `case DATABASE` reads the generic option back from `settings.options` (falling back to `-documentsDelimiter`). The guard also means a generic `-delimiter` now survives for the `-documents`/`-vertices`/`-edges` forms when no per-entity delimiter is set, instead of being cleared.

One addition beyond the issue: `;` joins the `isSeparator()` candidate set. Without it a semicolon-separated file whose extension is not `.csv` produced no candidate at all and the import died with `Cannot determine the file type`; `;` is no more ambiguous than the `-` and `_` already in that set, and the best-candidate-by-frequency logic is unchanged.

## #6812 - a record logged as skipped is saved as an anonymous Document

`createRecord()` used the same return signal for two opposite outcomes: *no mapping, save this anonymously* and *this record is rejected or deduplicated, drop it*. `parseRecord()` collapsed both into `attributes.map`, and `parseRecordsArray()` then materialized it into `settings.documentTypeName`. A duplicate `@id` without `"@strategy": "merge"`, or a mapping object missing `@cat`/`@type` - both of which log `will be skipped` and increment `context.errors` - ended up saved anyway.

Rather than a sentinel return value, the two cases are separated where the information already exists:

- `parseRecordsArray()` anonymously saves only when `mappingObject == null`. That is the one case the anonymous path is meant for, and it deliberately keeps working for the **nested** `"@cat":"e"` object, which returns its attribute map on purpose so `convertMap()` can turn it into an edge during the mapping phase. Returning a sentinel from `createRecord()` would have broken exactly that path.
- a deduplicated `@id` now hands the existing record back, so a caller links to it instead of re-materializing the skipped attributes.

That second change also fixes a latent `ClassCastException` on the same code path: `convertMap()` cast the destination vertex to `MutableVertex`, which an existing (immutable) record from the dedup branch could never satisfy. It now goes through `modify()`, a no-op on a `MutableVertex`.

## #6813 - an empty or self-closing XML sub-element inherits the previous sibling's value

`lastContent` was cleared only when a new **record** started, never when a new sub-element started, so `<rows><row><a>1</a><b></b></row></rows>` imported as `{a: "1", b: "1"}`, in both `load()` and `analyze()`. This is the sub-element twin of #2759, whose regression test uses attributes exclusively.

State is cleared on the sub-element boundary, but only there: the clear at `END_ELEMENT` is conditioned on closing the property element itself (`nestLevel == objectNestLevel + 2`), so a deeper descendant's text still feeds the property and `<a><x>1</x><y>2</y></a>` keeps its existing last-wins result. The `#2759` whitespace fix is untouched - whitespace *between* elements still cannot erase the content of the element currently being read. An element opened and closed with no character content becomes an empty property, which is its XML text content, rather than a missing or `null` one.

## #6814 - a long text value is treated as no evidence, so the column is typed LONG

`AnalyzedProperty.setLastContent()` treated a value longer than 100 chars as *stop sampling* **and** as *no evidence about the type*, returning before the numeric probes and - because `collectingSamples` was now permanently `false` - making every subsequent value for that column a no-op too. A column whose first value looked numeric was declared `LONG` and created as such in the schema, and the import then died on the long value it never examined with `Cannot convert type 'java.lang.String' to 'LONG'` (or, under `-onRowError skip`, silently dropped every long-text row).

Type refutation now runs on **every** value, independently of sample collection, which covers the `maxValueSampling` short-circuit the issue also flagged. Sample collection still stops exactly where it did.

## Tests

- `Issue6810ZipSourceTest` - single-entry and multi-entry (`:::nodes.csv`) archives.
- `Issue6811CsvDelimiterOptionTest` - SQL `WITH delimiter`, CLI `-delimiter`, `;` auto-detection, plus a guard that a plain comma file still imports with no delimiter set.
- `Issue6812JsonSkippedRecordTest` - duplicate `@id`, missing `@cat`, plus a guard that the legitimate anonymous path still saves.
- `Issue6813XmlEmptySubElementTest` - empty and self-closing siblings, plus a guard for nested sub-elements.
- `Issue6814AnalyzedPropertyLongTextTest` - unit-level refutation, values after the cut-off, end-to-end CSV import, plus guards that genuinely numeric columns are still typed `LONG`/`DOUBLE`.

Full `integration` module green: 181 unit + 118 integration tests, including `CSVImporterIT`, `JSONImporterIT`, `XMLImporterIT`, `JsonLImporterIT`, `Neo4jImporterIT`, `OrientDBImporterIT` and `XMLImporterFormatTest` (the #2759 regression).

https://claude.ai/code/session_01QF6LJaiD6ZRr6oA9WZxVBK

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ZIP, local, and remote source reopening, including reliable archive-entry selection.
  * CSV imports now honor configured delimiters and detect semicolon-separated files.
  * Fixed JSON handling for skipped, duplicate, and mapped records.
  * Corrected XML imports so empty or self-closing elements do not inherit previous values.
  * Improved type detection for long text and numeric-looking data.
  * Source reset failures are now reported correctly.

* **Tests**
  * Added coverage for ZIP, CSV, JSON, XML, and long-text import scenarios.

### DIFF
--- a/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/AnalyzedProperty.java
@@ -61,44 +61,48 @@ public class AnalyzedProperty {
     return index;
   }
 
+  /**
+   * Feeds one source value into the analysis. Type refutation (is this column still a candidate for {@code LONG} /
+   * {@code DOUBLE}?) runs on <b>every</b> value, independently of sample collection: stopping the collection of
+   * distinct values - because the value is too long to be worth indexing in the dictionary, or because
+   * {@code maxValueSampling} distinct values were already seen - used to return before the numeric probes, so a
+   * column whose first value happened to look numeric stayed a {@code LONG} candidate forever and was declared
+   * {@code LONG}. The import then died converting the very value that should have disproved it (issue #6814).
+   */
   public void setLastContent(final String lastContent) {
-    if (!collectingSamples)
+    if (lastContent == null)
       return;
 
-    if (lastContent != null) {
-      if (lastContent.length() > 100) {
-        collectingSamples = false;
-        contents.clear();
-        return;
-      }
+    this.lastContent = lastContent;
 
-      this.lastContent = lastContent;
-      if (contents.size() > maxValueSampling) {
-        collectingSamples = false;
-        contents.clear();
-        return;
-      }
-
-      contents.add(lastContent);
-
-      if (!lastContent.isEmpty()) {
-        if (candidateForInteger) {
-          try {
-            Long.parseLong(lastContent);
-          } catch (final NumberFormatException e) {
-            candidateForInteger = false;
-          }
+    if (!lastContent.isEmpty()) {
+      if (candidateForInteger) {
+        try {
+          Long.parseLong(lastContent);
+        } catch (final NumberFormatException e) {
+          candidateForInteger = false;
         }
+      }
 
-        if (candidateForDecimal) {
-          try {
-            Double.parseDouble(lastContent);
-          } catch (final NumberFormatException e) {
-            candidateForDecimal = false;
-          }
+      if (candidateForDecimal) {
+        try {
+          Double.parseDouble(lastContent);
+        } catch (final NumberFormatException e) {
+          candidateForDecimal = false;
         }
       }
     }
+
+    if (!collectingSamples)
+      return;
+
+    if (lastContent.length() > 100 || contents.size() > maxValueSampling) {
+      collectingSamples = false;
+      contents.clear();
+      return;
+    }
+
+    contents.add(lastContent);
   }
 
   public Set<String> getContents() {
--- a/integration/src/main/java/com/arcadedb/integration/importer/Source.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/Source.java
@@ -43,12 +43,24 @@ public class Source {
     this.closeCallback = closeCallback;
   }
 
+  /**
+   * Re-opens the underlying stream from the beginning. A failure here leaves the source unreadable, so it is
+   * propagated rather than only logged: swallowing it turned a broken re-open into an empty stream, and the import
+   * then reported a successful run with zero records instead of an error (issue #6810).
+   */
   public void reset() throws IOException {
     if (resetCallback != null)
       try {
         resetCallback.call(this);
       } catch (final Exception e) {
         LogManager.instance().log(this, Level.SEVERE, "Error on resetting source %s", e, this);
+
+        if (e instanceof IOException ioException)
+          throw ioException;
+        if (e instanceof RuntimeException runtimeException)
+          throw runtimeException;
+
+        throw new IOException("Error on resetting source " + this, e);
       }
   }
 
--- a/integration/src/main/java/com/arcadedb/integration/importer/Source.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/Source.java
@@ -49,19 +49,18 @@ public class Source {
    * then reported a successful run with zero records instead of an error (issue #6810).
    */
   public void reset() throws IOException {
-    if (resetCallback != null)
-      try {
-        resetCallback.call(this);
-      } catch (final Exception e) {
-        LogManager.instance().log(this, Level.SEVERE, "Error on resetting source %s", e, this);
+    if (resetCallback == null)
+      return;
 
-        if (e instanceof IOException ioException)
-          throw ioException;
-        if (e instanceof RuntimeException runtimeException)
-          throw runtimeException;
-
-        throw new IOException("Error on resetting source " + this, e);
-      }
+    try {
+      resetCallback.call(this);
+    } catch (final RuntimeException e) {
+      // com.arcadedb.utility.Callable.call() declares no checked exception, so both reset callbacks in
+      // SourceDiscovery already wrap their IOException into an ImportException: a RuntimeException is the only thing
+      // that can arrive here
+      LogManager.instance().log(this, Level.SEVERE, "Error on resetting source %s", e, this);
+      throw e;
+    }
   }
 
   public void close() {
--- a/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
+++ b/integration/src/main/java/com/arcadedb/integration/importer/SourceDiscovery.java
@@ -141,15 +141,17 @@ public class SourceDiscovery {
     return getSourceFromContent(new BufferedInputStream(connection.getInputStream()), connection.getContentLengthLong(), resource,
         source -> {
           try {
+            source.inputStream.close();
             connection.disconnect();
 
             final HttpURLConnection connection1 = ImportSecurityValidator.openRemoteConnection(urlPath, blockLocalNetworks);
 
             if (source.inputStream instanceof GZIPInputStream)
               source.inputStream = new GZIPInputStream(connection1.getInputStream(), 2048);
-            else if (source.inputStream instanceof ZipInputStream stream) {
-              source.inputStream = new ZipInputStream(connection1.getInputStream());
-              stream.getNextEntry();
+            else if (source.inputStream instanceof ZipInputStream) {
+              final ZipInputStream zip = new ZipInputStream(connection1.getInputStream());
+              positionZipStream(zip, resource);
+              source.inputStream = zip;
             } else
               source.inputStream = new BufferedInputStream(connection1.getInputStream());
           } catch (final Exception e) {
@@ -174,27 +176,22 @@ public class SourceDiscovery {
       filePath = getClass().getClassLoader().getResource(filePath).getFile();
     }
 
-    final File file = new File(filePath);
+    final String resolvedPath = filePath;
+    final File file = new File(resolvedPath);
 
-    final InputStream fis;
-    if (file.exists())
-      fis = new BufferedInputStream(new FileInputStream(file));
-    else {
-      fis = getClass().getClassLoader().getResourceAsStream(filePath);
-      if (fis == null)
-        throw new FileNotFoundException(filePath);
-    }
+    final InputStream fis = openLocalStream(file, resolvedPath);
 
     return getSourceFromContent(fis, file.length(), resource, source -> {
       try {
         source.inputStream.close();
         if (source.inputStream instanceof GZIPInputStream)
-          source.inputStream = new GZIPInputStream(new FileInputStream(file), 2048);
-        else if (source.inputStream instanceof ZipInputStream stream) {
-          source.inputStream = new ZipInputStream(new FileInputStream(file));
-          stream.getNextEntry();
+          source.inputStream = new GZIPInputStream(openLocalStream(file, resolvedPath), 2048);
+        else if (source.inputStream instanceof ZipInputStream) {
+          final ZipInputStream zip = new ZipInputStream(openLocalStream(file, resolvedPath));
+          positionZipStream(zip, resource);
+          source.inputStream = zip;
         } else
-          source.inputStream = new BufferedInputStream(new FileInputStream(file));
+          source.inputStream = openLocalStream(file, resolvedPath);
       } catch (final IOException e) {
         throw new ImportException("Error on reset local resource", e);
       }
@@ -203,6 +200,22 @@ public class SourceDiscovery {
       fis.close();
       return null;
     });
+  }
+
+  /**
+   * Opens the local source the same way for the initial read and for every {@link Source#reset()}: as a file when one
+   * exists at that path, otherwise as a classpath resource. Reset used to unconditionally re-open a
+   * {@link FileInputStream}, which cannot work for the classpath fallback.
+   */
+  private InputStream openLocalStream(final File file, final String filePath) throws IOException {
+    if (file.exists())
+      return new BufferedInputStream(new FileInputStream(file));
+
+    final InputStream stream = getClass().getClassLoader().getResourceAsStream(filePath);
+    if (stream == null)
+      throw new FileNotFoundException(filePath);
+
+    return stream;
   }
 
   private FormatImporter analyzeSourceContent(final Parser parser, final AnalyzedEntity.EntityType entityType,
@@ -233,8 +246,12 @@ public class SourceDiscovery {
       break;
 
     case DATABASE:
-      // NO SPECIAL SETTINGS
+      // NO PER-ENTITY SETTINGS: THE GENERIC `delimiter` OPTION IS THE ONE THE USER CAN SET ON THIS FORM
+      // (-delimiter / IMPORT DATABASE ... WITH delimiter = ';'), SO READ IT BACK FROM THE OPTIONS AND FALL BACK TO
+      // -documentsDelimiter WHEN IT IS ABSENT (ISSUE #6811)
       knownFileType = getFileTypeByExtension(settings.url);
+      final Object genericDelimiter = settings.options.get("delimiter");
+      knownDelimiter = genericDelimiter != null ? genericDelimiter.toString() : settings.documentsDelimiter;
       break;
 
     default:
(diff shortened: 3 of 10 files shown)


lvca:
CI note: the `integration-tests` lane failure on this PR is pre-existing on `main` and unrelated to these changes.

It fails on `RemoteConsoleIT.userMgmt` (console module, line 174 - `create user albert identified by einstein grand connect on db1` is expected to raise a syntax error and no longer does). The same lane fails identically on main at `b7c6c800d`, which is the exact commit this branch is based on: [run 33135270998](https://github.com/ArcadeData/arcadedb/actions/runs/33135270998) reports the same `Tests run: 16, Failures: 1` on the same test, with none of this PR's changes present.

The lane went red on main somewhere around #6853, which rewrote ~249 lines of `Console.java` plus `TerminalParser`; the last completed main run whose `integration-tests` passed predates that merge. Worth its own issue - it is not something this PR can fix.

Everything this PR does touch is green: all 22 integration-module ITs passed in that same job, including `CSVImporterIT`, `JSONImporterIT`, `XMLImporterIT`, `JsonLImporterIT`, `Neo4jImporterIT`, `OrientDBImporterIT` and `SQLRemoteImporterIT`. `vector-unit-tests` is likewise red on main at the same commit.

PR #6867, opened 2026-08-28
fix(schema): one copy point for the index builder swap (#5606)

### LINKED ISSUE #5606: TypeIndexBuilder subclasses hand-copy the common fields, so a new field silently stops propagating through withType()
Split out of the review of #5604.

`TypeIndexBuilder.withType()` returns a specialised subclass when the index type has settings of its own, and each of those subclasses copies the common builder state in its copy constructor, field by field:

- `TypeFullTextIndexBuilder(TypeIndexBuilder)`
- `TypeLSMVectorIndexBuilder(TypeIndexBuilder)`
- `TypeLSMSparseVectorIndexBuilder(TypeIndexBuilder)`
- `TypeGeoIndexBuilder(TypeIndexBuilder)` (added by #5604)

Each hand-copies roughly the same eleven fields (`unique`, `pageSize`, `nullStrategy`, `callback`, `ignoreIfExists`, `indexName`, `filePath`, `keyTypes`, `batchSize`, `maxAttempts`, plus the metadata identity).

The failure mode is silent. Adding a field to `TypeIndexBuilder` and wiring it through a `withX()` method works perfectly for `LSM_TREE` and `HASH` - which never swap builders - and is dropped on the floor for all four specialised types, because `withType()` hands back a new object that never learned about it. There is no compile error and no test that would notice unless it happens to cover that setting on a full-text, vector or geospatial index specifically. The recently added sorted-build fields (`buildMode`, `buildMemoryBudgetBytes`, `buildSpillDirectory`, `buildMergeFanIn`, `buildParallelism`) are not copied today, which is harmless only because `withType()` rejects a sorted build for anything but `LSM_TREE`.

Worth replacing with a single `copyCommonFrom(TypeIndexBuilder)` on `TypeIndexBuilder` that the four constructors call, so a new field has exactly one place to be added. A test that sets every builder option, runs it through `withType()` for each specialised type and asserts the state survived would keep it honest.

Not urgent - nothing is broken right now - but it is a footgun that costs nothing to remove and has already been paid for once: the shadowed `metadata` field fixed in #5478 was the same class of bug, one level down.

lvca:
Closes #5606.

## The defect

`TypeIndexBuilder.withType()` does not configure the builder it is called on: for the four index types that carry settings of their own it constructs a **different** builder and hands that back. Every one of those four copy constructors then hand-copied the same eleven fields of `IndexBuilder`:

- `TypeFullTextIndexBuilder(TypeIndexBuilder)`
- `TypeLSMVectorIndexBuilder(TypeIndexBuilder)`
- `TypeLSMSparseVectorIndexBuilder(TypeIndexBuilder)`
- `TypeGeoIndexBuilder(TypeIndexBuilder)`

So adding a field to `IndexBuilder` and wiring it through a `withX()` setter works perfectly for `LSM_TREE` and `HASH` - which never swap builders - and is silently dropped for FULL_TEXT, GEOSPATIAL and the two vector types. There is no compile error, and no existing test notices unless it happens to cover that setting on one of those four types specifically.

## What was actually being lost

Beyond the five sorted-build fields the issue named:

| field | declared in | carried over before |
|---|---|---|
| `replaceIfIncompatible` (#5675) | `IndexBuilder` | by nobody |
| `userMetadata` (#5723) | `IndexBuilder` | by nobody |
| `defaultKeyTypesForUndeclaredProperties` | `TypeIndexBuilder` (private) | by nobody |
| `buildMode`, `buildMemoryBudgetBytes`, `buildSpillDirectory`, `buildMergeFanIn`, `buildParallelism` | `TypeIndexBuilder` (private) | by nobody |
| `collations` | `IndexMetadata` | geospatial only |
| `typeIndexName` | `IndexMetadata` | geospatial only |

Nothing is broken for a user today, and this PR is not claimed to fix a reachable bug: every caller in the tree happens to call `withType()` before the setters that would be dropped, `withType()` refuses a sorted build for anything but `LSM_TREE`, and `defaultKeyTypesForUndeclaredProperties` is only set by the OpenCypher and MongoDB paths, which create `LSM_TREE` indexes. Those are circumstantial guards rather than a design, which is exactly what makes the footgun worth removing rather than documenting.

## The fix, and why not the one the issue proposed

The issue suggested a `copyCommonFrom(TypeIndexBuilder)` the four constructors call. That leaves the failure mode intact one step further out: a fifth specialised builder can still forget to call it, and it cannot reach the six fields that are private to `TypeIndexBuilder`.

Instead the four constructors now **delegate to a single copy constructor** on `TypeIndexBuilder`, each supplying only the two things that make it different - the index type it stands for and the `IndexMetadata` subclass its own settings live in:

```java
protected TypeGeoIndexBuilder(final TypeIndexBuilder copyFrom) {
  super(copyFrom, Schema.INDEX_TYPE.GEOSPATIAL,
      new GeoIndexMetadata(copyFrom.metadata.typeName, copyFrom.metadata.propertyNames.toArray(new String[0]),
          copyFrom.metadata.associatedBucketId));
}
```

Constructor chaining, not a method to remember: a subclass cannot exist without going through it. Each class then owns exactly one place where its own fields are copied - `IndexBuilder.copyBaseFieldsFrom` for the base ones, the copy constructor itself for `TypeIndexBuilder`'s - and the common part of the definition rides along through the `IndexMetadata.copyCommonTo` that already existed for `copyType()`, so `collations` and `typeIndexName` stop being a geospatial-only privilege. Net -74/+97 lines with the doc comments, and 44 lines of hand-copying gone.

## Test

`Issue5606BuilderStateCarryOverTest` is deliberately reflective, for the same reason the defect existed: a hand-written list of assertions is a second field list to keep in sync, and the first one already went stale.

- `everySettingSurvivesTheBuilderSwap` configures every option, calls `withType()` for each of the four specialised types, and compares **every declared instance field** of `IndexBuilder` and `TypeIndexBuilder`, naming the field and both values when one does not survive.
- `everySettingIsActuallyConfiguredByThisTest` is the anti-vacuity half: it fails when the setup leaves a field at its default, which is what a newly added field looks like. Without it the comparison above would quietly start proving nothing.
- `sortedBuildIsRefusedBySpecialisedTypes` pins down the one documented exemption - `buildMode` cannot be observed across the swap because `withType()` rejects `SORTED` for anything but `LSM_TREE` - so the exemption is asserted rather than assumed.
- `secondWithTypeKeepsTheSameBuilder` covers the SQL path, which calls `withType()` and then `withGeoType()`/`withFullTextType()`: a second swap there would discard everything configured in between.

Verified the test catches a real drop by removing one carried-over field: 4 failures, each naming `buildParallelism` and the index type.

## Verification

`mvn -o -pl engine -am test` over the schema, index, LSM and SQL index test packages: **1060 tests, 0 failures**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved index configuration when converting between specialized index types.
  - Maintained metadata, naming, key settings, callbacks, batching, retry limits, and other builder options during index updates.
  - Prevented configuration loss across repeated type conversions.
  - Ensured sparse-vector index creation uses the selected specialized index type consistently.

- **Tests**
  - Added coverage verifying configuration retention when converting bucket indexes to vector indexes.
  - Verified specialized builder and metadata types are preserved correctly.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/CreateIndexStatement.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/CreateIndexStatement.java
@@ -355,8 +355,9 @@ public class CreateIndexStatement extends DDLStatement {
       resultingIndex = ftBuilder.create();
 
     } else if (indexType == Schema.INDEX_TYPE.LSM_SPARSE_VECTOR) {
-      final TypeLSMSparseVectorIndexBuilder sparseBuilder = builder.withType(Schema.INDEX_TYPE.LSM_SPARSE_VECTOR)
-          .withSparseVectorType();
+      // Builder is already a TypeLSMSparseVectorIndexBuilder after the withType() above, exactly like the three
+      // branches around this one: re-selecting the index type here would be a no-op.
+      final TypeLSMSparseVectorIndexBuilder sparseBuilder = builder.withSparseVectorType();
       if (metadata != null) {
         final Map<String, Object> metadataMap = metadata.toMap((Result) null, context);
         final JSONObject jsonMetadata = new JSONObject(metadataMap);
--- a/engine/src/main/java/com/arcadedb/schema/BucketIndexBuilder.java
+++ b/engine/src/main/java/com/arcadedb/schema/BucketIndexBuilder.java
@@ -48,6 +48,36 @@ public class BucketIndexBuilder extends IndexBuilder<Index> {
     this.propertyNames = propertyNames;
   }
 
+  /**
+   * The copy constructor of the bucket-level swap, the mirror of
+   * {@link TypeIndexBuilder#TypeIndexBuilder(TypeIndexBuilder, Schema.INDEX_TYPE, IndexMetadata)}: {@link #withType}
+   * hands back a specialised subclass once the index type is known, and everything configured before that call has to
+   * survive being carried onto it.
+   * <p>
+   * {@link BucketLSMVectorIndexBuilder} used to copy the common state itself, field by field, which is the same
+   * footgun issue #5606 removed from the type-level hierarchy - and it was already missing the same two fields,
+   * {@code replaceIfIncompatible} (#5675) and {@code userMetadata} (#5723). One hierarchy fixed and the other left
+   * exposed is not a fix: the next field added to {@link IndexBuilder} would simply go missing here instead.
+   *
+   * @param copyFrom  the builder being replaced, whose configuration must survive the swap
+   * @param indexType the index type this specialised builder stands for
+   * @param metadata  a freshly built metadata of the subclass the specialised builder writes its settings into
+   */
+  protected BucketIndexBuilder(final BucketIndexBuilder copyFrom, final Schema.INDEX_TYPE indexType,
+      final IndexMetadata metadata) {
+    super(copyFrom.database, Index.class);
+    this.typeName = copyFrom.typeName;
+    this.bucketName = copyFrom.bucketName;
+    this.propertyNames = copyFrom.propertyNames;
+
+    // Null-guarded, unlike the type-level twin: a plain BucketIndexBuilder does not necessarily carry a metadata at
+    // all - only a specialised subclass installs one - so there may be no common definition to carry over.
+    this.metadata = copyFrom.metadata == null ? metadata : copyFrom.metadata.copyCommonTo(metadata);
+    this.indexType = indexType;
+
+    copyBaseFieldsFrom(copyFrom);
+  }
+
   @Override
   public IndexBuilder<Index> withType(Schema.INDEX_TYPE indexType) {
     if (indexType == Schema.INDEX_TYPE.LSM_VECTOR && !(this instanceof BucketLSMVectorIndexBuilder))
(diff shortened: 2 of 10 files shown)


lvca:
### CI triage

Both reviews are clean (claude-review: "no correctness issues found"; CodeRabbit: "no actionable merge-blocking risk"). Two lanes are still red, and both are inherited from `main` rather than caused by this PR - main run **33140801714** at this PR's exact base commit `9cf4d2623` fails the same two:

- **`integration-tests`** - `RemoteConsoleIT.userMgmt`, the console `CREATE USER ... grand ... on` syntax check that no longer raises. Red on main since #6853, unrelated to index builders.
- **`vector-unit-tests`** - `Issue6772WriteBeforeSearchPrefixReuseTest`, `Issue6655StaleGraphPrefixReuseTest`, `Issue6797DeltaScanScalingTest`, all rebuild/prefix-reuse timeouts. Red on main at the same commit.

Everything else is green after a rerun. `unit-tests` failed once on three unrelated concurrency races - `GraphAnalyticalViewTest.issue6641_*`, `Issue6470AsyncSuccessCallbackDurabilityTest.executorWideOnOkFiresAtTheDurabilityPolicyBoundary` and `Issue6762WebSocketEventBusTest.shutdownStillJoinsAfterASeparateStopSignal` - which all pass locally on this branch and passed on the rerun. `build-and-package` failed once on an `npm install` `ECONNRESET` and passed on rerun.

PR #6868, opened 2026-08-28
fix(check): a dangling edge endpoint that is absent is reported, not flagged corrupted (#5777)

### LINKED ISSUE #5777: checkEdges flags a missing IN endpoint as a corrupted record but a missing OUT endpoint as nothing, and flagging either contradicts the RECORD-scope principle

lvca:
Fixes #5777.

## The two defects, one cause

`GraphDatabaseChecker` carried **four** copies of the same fifteen-line endpoint probe - both sides of `checkEdges`, plus the far-endpoint load in `checkIncomingEdges` and `checkOutgoingEdges` - and one had drifted. `checkEdges`' incoming side flagged the **missing vertex** corrupted alongside the edge:

```java
// incoming endpoint not found
report.corrupt(edgeRID);
report.corrupt(edge.getIn());   // <-- the missing VERTEX too

// outgoing endpoint not found
report.corrupt(edgeRID);        // <-- edge only
```

The unknown-error arms beside them were asymmetric in the opposite direction (both add the vertex), so the divergence is only in the `RecordNotFoundException` arms - a copy-paste divergence rather than a decision. The same drift was present in the two adjacency-walk copies, which the issue did not name but which flag the absent far endpoint for the same reason.

The drift also picked the wrong side. Every RID in `corruptedRecords` puts its **bucket** into `affectedBuckets`, and `CHECK DATABASE ... FIX` then drops and rebuilds every index on it - a full bucket scan per index. Flagging the absent endpoint therefore had a dangling-edge sweep after a bulk delete rebuild the indexes of every vertex bucket the deleted vertices lived in, and bought nothing for it: the repair loop's `deleteRecord` on a RID that is not there raises `RecordNotFoundException` and is ignored, so no repair was ever performed on the strength of the flag. That is the rule #5680/#5764 already settled for a hand-typed RID in the `RECORD` scope, quoted verbatim in both scoped arms.

Settled as the issue asked: nothing else consumes `corruptedRecords`. It becomes `affectedBuckets` in `DatabaseChecker.check()` and the delete list in each repair loop, and neither wants a phantom.

## Better than the proposed one-line deletion

The issue proposes dropping the one `addCorrupted(edge.getIn())`. That fixes the symptom and leaves four copies of the code that produced it - including two the issue had not spotted. Instead, all four are replaced by a single `loadEndpointVertex(edge, edgeRID, direction, ...)` arm, so the two directions cannot report the same finding differently again, and the policy lives in exactly one place:

| endpoint load | edge | endpoint |
|---|---|---|
| `RecordNotFoundException` - not there | flagged | **not** flagged, reported via `trackMissingReference` |
| anything else - there, unreadable | flagged | flagged (this IS corruption) |

The list walkers keep pruning the dangling adjacency entry - the only thing that legitimately differs between the two families of caller - by acting on the `null` return. `invalidLinks` keeps the split it already had, so no published number is renumbered beyond `corruptedRecords`/`totalCorruptedRecords` themselves.

## Verification

New `Issue5777DanglingEndpointScopeTest` (6 cases): both directions flag the edge and nothing else; both absent endpoints still reach the operator through `missingReferences` and the warnings; **no index on the absent vertex's bucket appears in `rebuiltIndexes`** - the cost, measured where an operator pays it; an unreadable endpoint is still flagged; the adjacency walk stops flagging the absent far endpoint; `FIX` still removes the dangling edge and counts it once.

Three of the six fail against the previous behaviour, including the `rebuiltIndexes` one, which reported `Person_0_...` before the change.

Updated for the new counts:
- `CheckDatabaseSinglePassTest.theCorruptedTotalDoesNotDoubleCountARetainedRecordPastTheCap` - 2 to 1, exactly as the issue predicted. It gets sharper: the edge is now the only RID the visit produces, so the arithmetic is purely the de-duplication being measured.
- `CheckDatabaseTest.checkBrokenDeletedVertex` - `TOTAL` to `TOTAL - 1`: all the edges, no longer the low-level-deleted root vertex.
- `CheckDatabaseAutoFixAccountingTest` - its precondition (a flagged RID that cannot be deleted) is no longer reachable through a dangling endpoint, so it is re-based onto the arm that still produces the shape: an adjacency list naming an edge record that is gone. The #6128 accounting rule it pins is unchanged, and is now asserted on `removedRecords` rather than `autoFix`, which also folds in the prunes.

`mvn -o -pl engine test` over `com/arcadedb/graph/**` + `com/arcadedb/database/**`: 904 tests, 0 failures.

## Docs

`CHECK DATABASE`'s reported `corruptedRecords`/`totalCorruptedRecords` no longer include RIDs that do not exist. Worth a line in the docs repo if the counters are described there - happy to prepare it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database checks for graph records with missing or unreadable endpoints.
  * Missing endpoints are now reported as references, while unreadable endpoints and dangling edges are identified as corrupted.
  * Database repair now accurately reports successfully removed records and excludes failed deletions.
  * Graph index rebuilding excludes entries for missing endpoints.
  * Repair operations consistently remove and count dangling edges.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
@@ -1878,10 +1878,12 @@ public class GraphDatabaseChecker {
    */
   private static VertexInternal loadEndpointVertex(final Edge edge, final RID edgeRID, final Vertex.DIRECTION direction,
       final CheckReport report, final Map<RID, Long> missingReferences, final Map<RID, String> missingReferenceErrors) {
+    // Every constant spelled out and NO default: the switch is visibly exhaustive, and a constant added to
+    // DIRECTION one day fails to compile here instead of being swallowed into a catch-all arm.
     final boolean incoming = switch (direction) {
       case IN -> true;
       case OUT -> false;
-      default -> throw new IllegalArgumentException(
+      case BOTH -> throw new IllegalArgumentException(
           "Cannot load a single endpoint of edge " + edgeRID + " for direction " + direction);
     };
     final RID endpoint = incoming ? edge.getIn() : edge.getOut();
--- a/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
@@ -1913,6 +1913,13 @@ public class GraphDatabaseChecker {
    * count instead of emitting one line per dangling edge. A single missing supernode can be referenced by millions of
    * edges; this collapses that fan-out into "vertex X could not be loaded, referenced by N edge(s)". The distinct-target
    * set is bounded by {@code maxTracked} to keep memory in check (counts for already-tracked targets keep incrementing).
+   * <p>
+   * "MISSING" IS THE COMMON CASE, NOT THE ONLY ONE. {@link #loadEndpointVertex} calls this for BOTH of its failure
+   * arms, so an entry can also be a target that is right there and could not be DECODED. The distinction is not lost -
+   * {@code missingReferenceErrors} carries the exception text per target, and only the absent ones leave the target
+   * unflagged - but a consumer reading {@code missingReferences} alone must not take the key to mean "this record no
+   * longer exists". Longstanding behaviour, spelled out here rather than changed: the map is the operator's
+   * "which far records did edges fail to reach", and both shapes belong in it.
    */
   private static void trackMissingReference(final Map<RID, Long> missingReferences, final Map<RID, String> missingReferenceErrors,
       final int maxTracked, final RID target, final String error) {
--- a/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
+++ b/engine/src/main/java/com/arcadedb/graph/GraphDatabaseChecker.java
@@ -1867,8 +1867,10 @@ public class GraphDatabaseChecker {
    * vertex bucket that does not materialise as one is a load failure, and letting the {@link ClassCastException}
    * out to be caught further away as something else would describe it worse.
    *
-   * @param direction {@link Vertex.DIRECTION#IN} for the edge's incoming endpoint, anything else for its outgoing
-   *                  one
+   * @param direction {@link Vertex.DIRECTION#IN} or {@link Vertex.DIRECTION#OUT}, naming which endpoint to load.
+   *                  {@link Vertex.DIRECTION#BOTH} is rejected rather than folded into one of them: an edge has one
+   *                  endpoint per direction, so "both" names no single record to return and a caller that passes it
+   *                  has a bug that silently reading the OUT side would hide
    *
    * @return the endpoint vertex, or {@code null} when it could not be loaded - the caller decides what that means
    * for the adjacency entry it is walking, which is the only thing that legitimately differs between the two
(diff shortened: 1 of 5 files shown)


lvca:
Re-ran the affected-package sweep **with `-am`**, as CLAUDE.md requires (without it the module resolves `arcadedb-engine` from `~/.m2` and a behaviour-only change can silently prove nothing):

```
mvn -o -pl engine -am test \
  -Dsurefire.includes='com/arcadedb/graph/**/*Test.java,com/arcadedb/database/**/*Test.java' \
  -DexcludedGroups=benchmark,slow,vector

Tests run: 904, Failures: 0, Errors: 0, Skipped: 1
BUILD SUCCESS
```

Same result as the figure in the description, now confirmed against the working tree rather than a possibly-stale artifact.

On the two remaining nits: the `@param` pass is skipped deliberately, since the previous round asked for *less* comment volume on this method and the five undocumented parameters are self-describing. The `shrinkRecordBuffer` page-layout coupling is real but already pinned in one other place (`CheckDatabaseSinglePassTest`'s equivalent helper), so the format change that breaks one breaks both, in the same file-local way.

Issue #4932, opened 2026-07-03
[storage] deleteRecordInternal swallows the retry ConcurrentModificationException it deliberately throws (orphaned chunk chains)


lvca:
**Severity:** MEDIUM (space leak, missed retry)

**Location:** `LocalBucket.java:1108-1118` (throw "signal retry") vs `:1175-1181` (generic `catch (final Exception e)`).

`ConcurrentModificationException extends NeedRetryException extends ArcadeDBException` (RuntimeException), so the generic catch at `:1175` catches the retry signal thrown 60 lines above in the same try, logs it, zeroes the slot pointer, and **returns success**. The retry never happens; remaining `NEXT_CHUNK` records of the multi-page record are orphaned (permanent space leak) and the caller believes the delete succeeded.

**Fix:** add `catch (final ConcurrentModificationException e) { throw e; }` before the generic catch, mirroring the existing `RecordNotFoundException` clause.

---
_Filed from an internal engine deep-audit (2026-07). Traced to source; verify against current `main` before fixing and add a regression test._

lvca:
Fixed in #5012 (merged). ConcurrentModificationException - the retry signal deleteRecordInternal deliberately throws for concurrently-modified multi-page chunk chains - is now rethrown ahead of the generic catch that used to swallow it, zero the slot pointer and report success while orphaning the remaining NEXT_CHUNK records (permanent space leak). Mirrors the adjacent RecordNotFoundException clause. Trace-verified: deterministically forcing the mid-delete race would need internal page-format surgery disproportionate to the two-line catch-order fix.

PR #6870, opened 2026-08-28
fix(engine): name a broken chunk chain on the delete path, and keep the snapshot barrier's page count out of the filesystem

### LINKED ISSUE #6132: Backup phase-2b round 3: the last unbounded filesystem I/O under the barrier's locks, a barrier failure nobody counts, and a cap that can refuse a RAM-only shadow
Three things left behind by #6125 (merged as #6126). None is a correctness bug; all three are places where that PR's own stated invariants are not quite met.

The invariant #6125 established, over three review rounds, is: **everything the t0 barrier does while holding the JVM-wide page-manager lock must be bounded**, because every committer of every database in the process queues behind it. Rounds 1 and 2 closed the unbounded waits and the blocking `mkdirs`/`getUsableSpace`. These are the remaining gaps.

## 1. `getTotalPages()` is N `channel.size()` syscalls inside the locked section

`PageManager.buildSnapshot` runs with `applyLock.writeLock()` and the global `PageManager.lock()` held, and calls `paginated.getTotalPages()` once per page file:

```java
// PaginatedComponentFile
public long getTotalPages() throws IOException {
  channelLock.readLock().lock();
  try {
    return channel.size() / pageSize;     // <-- a syscall, per file, under the JVM-wide lock
  } finally { ... }
}
```

I waved this through in #6126 with a javadoc note calling it "the one `channel.size()` per file that is definitionally part of t0" - which is true about *when* it must be read, but not a reason it has to be a syscall. On a database with many buckets, indexes and their compacted sub-indexes this is dozens to hundreds of `fstat` calls plus a `channelLock` acquisition each, all under the lock the rest of the change worked to keep short, and on a stalled filesystem none of them is bounded by `SNAPSHOT_BARRIER_MAX_MILLIS` the way every other step now is.

There is already an in-memory counter: `PaginatedComponent.pageCount`, an `AtomicInteger` seeded from the file size at open and advanced by `updatePageCount` on every flush. It is **not** a drop-in replacement and that is the whole content of this issue:

- `buildSnapshot` iterates `FileManager.getFiles()` (files), not schema components, and a file may have no component (`getFileByIdIfExists` can return null) - so the fallback path still needs a size.
- `pageCount` is monotonic (`Math.max`), so a file that shrank - compaction, truncate - would report too many pages, and the snapshot would read past the end of the file.
- It advances on flush, and the barrier drains the pipeline first, so at t0 the two *should* agree - but "should" is exactly what needs proving before the syscall is removed, since being wrong here is silent: too low truncates the archive, too high reads garbage.

Worth doing properly: establish whether the counter is authoritative at t0 (the barrier guarantees a drained pipeline, which is most of the argument), and if it is, use it with the syscall as the fallback for component-less files. A regression test should assert the two agree at t0 across compaction, truncate and a freshly appended page.

## 2. A barrier that fails is invisible except in the log

#6126 added `arcadedb_engine_snapshot_windows_overflowed_total` and `..._failed_total`, but both are incremented from `PageSnapshot.invalidate` - which requires a window to **exist**. A barrier that gives up before publishing one (`trySuspendUntil` timing out on a concurrent resume, or `waitForCurrentFlushToCompleteUntil` timing out on an in-flight batch, both introduced in #6126) throws `PageSnapshotException` and is counted only as a barrier in `snapshot_barrier_count_total`, indistinguishable from a successful one.

That is the same blind spot #6116 filed the invalidation counter to close, one step earlier in the sequence: the consumer falls back to suspend-and-freeze and still completes, so a backup that has quietly gone back to throttling every writer on the server leaves no trace an operator can alert on. A `snapshot_barrier_failed_total`, ideally split by which step gave up, closes it.

## 3. The automatic cap can fall below `pageSnapshotMaxRAM` and refuse a shadow that would never touch the disk

`PageShadow`'s cap covers RAM **plus** spill:

```java
if (maxTotalBytes > 0 && ramBytes + spillBytes + length > maxTotalBytes)
  return false;
```

and #6125's automatic cap is `max(1, min(t0 page-file size, usableSpace(spillVolume) / 2))`. So on a spill volume with, say, 40 MB free, the cap resolves to 20 MB - below the 64 MB default `pageSnapshotMaxRAM`. A shadow that would have lived entirely in RAM and **never touched that volume at all** is then invalidated at 20 MB, and its backup falls back to throttling the writers because of free space it was never going to use.

`max(pageSnapshotMaxRAM, min(databaseSize, usable / 2))` is the obvious shape: the RAM budget needs no disk, so it should not be bounded by disk. I considered this while fixing the sentinel collision in #6126 and deliberately kept the arithmetic simple rather than smuggle a semantic change into a bug fix - but the current behaviour is wrong in a case the feature is explicitly aimed at (an operator who pointed `pageSnapshotSpillPath` at a small volume).

Note this interacts with item 1 of #6125's design: the cap is documented as "RAM plus spill file", so the fix should say plainly that the automatic cap never goes below the RAM budget, and `Issue6125SnapshotBarrierAndCapTest.theAutomaticCapNeverDegeneratesIntoTheUncappedSentinel` needs the new expectations.

### LINKED ISSUE #6282: Follow-ups from #6258: the delete path still calls corruption contention, a structural probe that reads a stale snapshot, and an IOException that means two things
Four findings from the work on #6258 (merged as `2a880199c`, PR #6262). Filed together because they were found together; each is separable and can be taken on its own. Items 1 and 2 are the same distinction #6258 drew on the read path, missing from the paths either side of it; items 3 and 4 are what makes that distinction fragile to maintain.

Line numbers below are against `2a880199c`.

---

## 1. The DELETE path still reports a broken chunk chain as contention

#6258 item 2 taught the READ path to tell a chain that does not parse from a record that moved under it: `loadMultiPageRecord` confirms the break and raises `BrokenChunkChainException`, which is deliberately not a `NeedRetryException`. The DELETE path was left as it was, and still conflates the two (`LocalBucket.java:2883`):

```java
if (chainProblem != null) {
  if (!force)
    // Chunk was modified/removed by a concurrent operation — signal retry (#4932)
    throw new ConcurrentModificationException(
            "Multi-page record " + rid + " chunk " + chunkId + " was modified concurrently. Please retry");
```

`chainProblem` is set a few lines above by exactly the structural conditions #6258 classified as corruption - a pointer past the end of the file, a cleaned slot, a marker that is not `NEXT_CHUNK` - so a corrupted record being deleted is still reported as a concurrency problem, and still costs the caller's retry machinery a full round of transactions before it gives up.

The cost is not only the message. Because the exception does not say which it is, **five call sites still walk the chain a second time to find out**:

- `LocalDatabase.deleteRecordNoLock:1448` (index cleanup), `:1472` (vertex physical delete), `:1487` (plain-record physical delete)
- `BucketIterator.fetchNext:217` and `:257`
- `LocalBucket.deleteCorruptedRecord:470`

Each of those is a `!bucket.isChunkChainBroken(rid)` probe that exists purely to disambiguate a `ConcurrentModificationException`. Teaching `deleteRecordInternal` to raise `BrokenChunkChainException` for the branches it has already positively identified would let every one of them go, and would make the `BrokenChunkChainException` arm on `LocalDatabase`'s plain-record branch live - it was deliberately left out in #6258 (with a comment saying why) precisely because `deleteRecordInternal` never names a broken chain as one.

## 2. `isChunkChainBroken` judges from the transaction's pinned view, and it gates force-delete

`isChunkChainBroken` is the probe all of the above depends on, and it walks the chain through the caller's transaction (`LocalBucket.java:2749`, and it passes `fromNewestCommitted=false` at `:2760`):

```java
final BasePage page = database.getTransaction().getPage(new PageId(database, file.getFileId(), pageId), pageSize);
...
return findBrokenChunkChain(rid, page, recordPositionInPage, false) != null;
```

Under `READ_COMMITTED` that reads the newest committed image and the answer is sound. Under `REPEATABLE_READ` the transaction has pinned the pages it walked, so this reads a **stale snapshot**: a chain that a concurrent commit has since rewritten can present as broken in the pinned image while the committed image walks cleanly. The doc comment on the method claims the opposite ("a transient concurrent modification never reports `true`"), which holds for the version-blindness it was written about but not for snapshot staleness.

What makes it worth fixing rather than noting is what the answer is used for: `deleteCorruptedRecord:470` and `LocalDatabase.deleteRecordNoLock:1472/:1487` escalate to **force-delete** on a `true`, and `check(fix)` deletes the record it flags. A false positive there removes a healthy record.

#6258 added exactly the walk this wants - `findBrokenChunkChainInNewestCommitted` (`LocalBucket.java:3572`), which reads every page from the `PageManager` and is already used to confirm a break authoritatively on the read path. Pointing these callers at it (or giving `isChunkChainBroken` the `fromNewestCommitted=true` treatment) closes the gap with code that already exists.

Worth pairing with item 1: if the delete path names a broken chain itself, most of these callers stop needing the probe at all, and the remaining ones become the narrow case.

## 3. `IOException` carries two opposite meanings inside the same walk

The third round of review on #6262 caught a real bug of this shape: `findBrokenChunkChain`'s blanket `catch (final Exception e) { return "error walking chain: " + e.getMessage(); }` (`LocalBucket.java:2728`) turned a transient `IOException` re-fetching a page into a *confirmed, permanent, non-retryable* corruption verdict, pointing the operator at a command that deletes the record. That was fixed for the page LOAD by catching at the load site (`:2703`) and answering `null` - "cannot prove a break".

The underlying ambiguity is still there. `getRecordPositionInPage` (`LocalBucket.java:3641`) reports a **corrupt slot offset** by throwing an `IOException`:

```java
if (recordPositionInPage != 0 && recordPositionInPage < contentHeaderSize)
  throw new IOException("Invalid record #" + fileId + ":" + ...);
```

So within a few lines of the same walk, `IOException` means "the disk failed, prove nothing" in one place and "this record is corrupt, report a break" in another, and the only thing keeping them apart is *where* the try/catch sits. That is a positional invariant, not a structural one: the next person who restructures this walk, or adds a page read to it, has no type-level signal to preserve. The corresponding hazard in the other direction is live today - a genuine I/O failure reading the slot table is still swallowed as corruption by the blanket catch.

A distinct signal for "this page's contents do not make sense" (a `PageCorruptionException`, or a sentinel return) would make the line permanent. Cheap, and it retires a class of mistake rather than one instance of it.

## 4. No fault-injection hook for page reads, so the highest-consequence path is untestable

The bug in item 3 was found by reading, not by a test, and it could not have been found by a test: there is no way to make a specific page read fail on demand. `PageManager` has no test hook (`getImmutablePage`/`loadPage` go straight to the read cache, the flush queue, and the file), so "the confirmation walk hits an I/O fault" - a path whose wrong answer is a permanent corruption verdict on a healthy record - has no coverage, and neither does any other I/O-failure path in the engine.

The reviewer flagged the gap explicitly and I could not close it without inventing a hook, which did not belong in that PR.

A test-only failure predicate on `PageManager` (something like "fail the next read of page X with an `IOException`"), gated so it costs nothing in production, would cover:

- the `findBrokenChunkChain` load failure of item 3 (verdict must be "cannot prove a break", read must retry, not condemn)
- `validateChainRead`'s deliberate decision to let an `IOException` escape rather than spend the retry budget on a broken disk (#6217)
- flush and torn-read robustness paths that are today only reasoned about

This is the item I would take first if the goal is keeping the other three fixed once they are fixed.

lvca:
Closes #6282
Closes #6132

Two independent follow-up issues, taken together because both land in `PageManager`/`LocalBucket` and would have conflicted.

## #6282 - corruption is not contention, on the two paths either side of the read

#6258 taught the READ path to tell a chunk chain that does not parse from a record that moved under it. These are the paths on either side of it.

**1. The delete path names the break instead of calling it contention.** `LocalBucket.deleteRecordInternal` positively identified a structurally broken chain - a pointer past the end of the file, a cleaned slot, a marker that is not `NEXT_CHUNK` - and then raised the #4932 `ConcurrentModificationException`, a `NeedRetryException`. The caller spent a full round of transactions on a retry that could not possibly succeed, and five call sites had to walk the chain a SECOND time just to find out which of the two they were holding.

It now confirms first, with the delete-side counterpart of #6258's `confirmBrokenChunkChain`. A delete consumes no bytes, so it cannot ask "are the chunks this read consumed still current"; it asks a sharper question instead - **does the newest committed image break at the SAME HOP, to the SAME TARGET?** A committed image that walks cleanly is a chain caught mid-publication; one that breaks somewhere else is a record that was rewritten under the delete. Both keep the retry. Only when the two agree does it raise `BrokenChunkChainException`, which is deliberately not a `NeedRetryException`.

That makes live the `BrokenChunkChainException` arm #6258 left out of `LocalDatabase`'s plain-record branch as dead code, and lets `LocalBucket.deleteCorruptedRecord` escalate without a probe.

> One thing worth flagging for review: `deleteRecordInternal` ends in a blanket `catch (Exception) -> log, zero the slot, return success`. Without an explicit rethrow arm - the same arm #4932 had to add for the CME - the new exception was swallowed and the delete silently "succeeded", orphaning every chunk past the break. The arm is in this PR, and a test pins it.

**2. `isChunkChainBroken` judges from the newest committed image.** It walked the chain through the CALLER's transaction, so under `REPEATABLE_READ` it read a pinned - possibly stale - snapshot: a chain a concurrent commit had since rewritten still presented as broken. Every caller escalates a `true` to a FORCE delete and `check(fix)` deletes the record it flags, so a false positive removes a healthy record. Its body is now `findBrokenChunkChainInNewestCommitted(rid) != null`, the walk #6258 already built for exactly this.

**3. `IOException` no longer means two opposite things in one walk.** New `PageCorruptionException extends IOException`, thrown by `getRecordPositionInPage` for a slot offset that points inside the page header. `findBrokenChunkChain` now answers by TYPE and not by where the `try` sits: `PageCorruptionException` -> a break; any other `IOException` -> **null, cannot prove a break**; the unchecked out-of-page-bounds family -> a break, as before. Java's "subclass first" rule makes the compiler keep the ordering, which is the whole point - the invariant used to be positional, and the hazard in the live direction (a genuine I/O failure reading the slot table swallowed as corruption) is closed. `deleteRecordInternal`'s chunk-read catch got the same split, so a corrupt chunk slot no longer escapes as a bare `IOException`.

**4. A page-read fault-injection hook.** `PageManager.setPageReadFaultInjector(...)`, test-only, hooked at `getCachedPage` - the funnel both `getImmutablePage` and `getMutablePage` go through - and NOT at the physical read in `loadPage`: a test cannot keep a page out of the read cache, so a physical-read hook would never fire for a page the engine has just written. Cost when nothing is installed: one read of a `static volatile` that is always null, the same shape as `activeSnapshots` on the write path.

**Beyond the issue text, deliberately:** `BucketIterator`'s two `ConcurrentModificationException` + `isChunkChainBroken` arms are removed. The issue counted them among the delete-path disambiguations; they are on the READ path, where the loader now answers the same question more strongly (committed image broken AND the read's own chunks current). A CME reaching them therefore means the break could NOT be confirmed - the record moved - and a probe saying "broken" about it silently drops a HEALTHY record from a scan. The delete-path probes stay: there the CME can come from the delete's own walk, which has no byte-level second question.

## #6132 - the last unbounded FS I/O under the barrier, a failure nobody counted, a cap below its own RAM budget

**1. `getTotalPages()` is a field read, not N syscalls under the JVM-wide lock.** The issue proposed reusing `PaginatedComponent.pageCount`, and then listed the three reasons it is not a drop-in: it is monotonic (so a shrunk file over-reports), `buildSnapshot` iterates FILES rather than components (so a fallback is still needed), and being wrong is silent.

**A better place to put the counter is the FILE, where all three objections disappear.** `PaginatedComponentFile` now keeps its own `totalPages`: seeded from the real length at `open()`, advanced past every successful `write(MutablePage)`, re-seeded on every reopen and rename. It is exact by construction rather than authoritative-if-you-squint - `write` is the only operation that extends a paginated component file, it always writes one whole page at a known page number, and nothing truncates one - so there is no monotonicity caveat, no component lookup and no fallback path. It also removes the syscall from the second caller, `loadPage`'s new-page test on every read-cache miss, which is as hot as the engine gets.

`getTotalPagesFromChannel()` is kept for one purpose: the regression test asserts the counter and the filesystem agree, at t0, across appends, an index compaction and a reopen. (Mutation-checked: `pageNumber` instead of `pageNumber + 1` turns it red.)

> Trap for the next person: `open()` runs from the `ComponentFile` constructor, BEFORE this class's field initializers, so an `AtomicInteger` field is still null when it has to be seeded. It is a `volatile int` with no initializer plus a static `AtomicIntegerFieldUpdater`.

**2. A barrier that fails is counted.** `snapshotBarriersFailed`, split into `...FailedSuspend` and `...FailedFlush`, incremented from a new catch in `openSnapshotInternal`. `PageSnapshotException.Reason` already existed (#6394); only the counters were missing. Wired through `PPageManagerStats` -> `Profiler` -> `EngineMetricsBinder` as `arcadedb.engine.snapshot.barrier.failed[.suspend|.flush]`. Until now the only trace of a backup that had quietly gone back to throttling every writer on the server was a WARNING in the log: every other snapshot counter is incremented from a window that EXISTS, and the consumer's fallback completes the backup anyway.

**3. The automatic cap never falls below the RAM budget.** `max(1, min(databaseSize, max(pageSnapshotMaxRAM, usable/2)))`. On a spill volume with 40 MB free the cap used to resolve to 20 MB, below the 64 MB default RAM budget - so a shadow that would have lived entirely in memory and never written a byte to that volume was invalidated anyway, in exactly the case `pageSnapshotSpillPath` exists for. `GlobalConfiguration`'s own description now says so, and `Issue6125SnapshotBarrierAndCapTest.theAutomaticCapNeverDegeneratesIntoTheUncappedSentinel` sets `pageSnapshotMaxRAM=0` first, or the sentinel edge it guards is no longer reachable.

## Behaviour changes

- A `DELETE` over a confirmed-broken chunk chain now raises `BrokenChunkChainException` where it raised a retryable `ConcurrentModificationException`. `DELETE_TOLERATE_BROKEN_CHAIN` and `CHECK DATABASE FIX` behave exactly as before. `BrokenMultiPageRecordDeleteTest` updated.
- A bucket scan no longer skips a record on an unconfirmed `ConcurrentModificationException`; it propagates so the caller can retry, which is what `getSkippedRecordCount`'s contract already documented.

## Tests

- `Issue6282BrokenChainDeleteAndProbeTest` (7) - the delete-side verdict and its non-`NeedRetryException` type, the tolerant path through the newly live arm, a break the committed image does not confirm keeping its retry, the pinned-vs-committed probe gap (provoked by pinning a BROKEN head page in a `REPEATABLE_READ` transaction and repairing it from another thread - the direction that force-deletes a healthy record), a corrupt chunk slot as corruption rather than an I/O error, and an injected page-read fault never becoming a licence to force-delete.
- `Issue6132SnapshotBarrierFollowupsTest` (3) - the counter agrees with the filesystem at t0 across appends/compaction/reopen, a failed barrier is counted with the split matching the reason it reported, and the cap's RAM floor.

## Docs

The new `arcadedb.engine.snapshot.barrier.failed*` metrics and the `pageSnapshotMaxSize` floor are worth a line in the observability and configuration pages of the docs repo - happy to prepare that separately.

## Summary by CodeRabbit

- **New Features**
  - Added snapshot-barrier failure metrics, including total, suspend-timeout, and flush-timeout counts.
  - Improved automatic snapshot storage sizing using database size, available disk space, and RAM limits.
  - Added more accurate page-count tracking during writes, compaction, and reopening.

- **Bug Fixes**
  - Improved detection and handling of corrupted multi-page records during iteration and deletion.
  - Corruption is now distinguished from temporary I/O or concurrency issues, enabling safer retries and tolerant cleanup.
  - Added safeguards against deleting records after transient page-read failures.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -551,10 +551,11 @@ public enum GlobalConfiguration {
       loudly, so the consumer can fall back to the suspend-and-freeze path - never a silently truncated or torn \
       snapshot. The default -1 sizes the cap AUTOMATICALLY when the window opens (issue #6125), as the smaller of \
       the ceiling the shadow provably cannot exceed - one pre-image per page that existed at t0, so the t0 size of \
-      the page files - and half the space still usable on the volume holding the spill file. A flat number cannot \
-      do that: measurements on a 128 MB database show the shadow reaching 100% of the database under a flat-out \
-      writer, so any fixed default is simply the database size above which backups silently start falling back to \
-      throttling the writers. Set a positive value to pin an absolute cap in MB, or 0 for no cap at all; any \
+      the page files - and half the space still usable on the volume holding the spill file, but never less than \
+      PAGE_SNAPSHOT_MAX_RAM, because that half of the budget never touches the disk (issue #6132). A flat number \
+      cannot do that: measurements on a 128 MB database show the shadow reaching 100% of the database under a \
+      flat-out writer, so any fixed default is simply the database size above which backups silently start falling \
+      back to throttling the writers. Set a positive value to pin an absolute cap in MB, or 0 for no cap at all; any \
       negative value means automatic, so -1 is the spelling to use rather than the only one accepted.""",
       Long.class, -1),
 
--- a/engine/src/main/java/com/arcadedb/Profiler.java
+++ b/engine/src/main/java/com/arcadedb/Profiler.java
@@ -380,6 +380,11 @@ public class Profiler {
     json.put("snapshotBarrierTime", new JSONObject().put("count", pStats.snapshotBarrierMillis));
     json.put("snapshotBarrierMaxTime", new JSONObject().put("value", pStats.snapshotBarrierMaxMillis));
     json.put("snapshotBarriersInexact", new JSONObject().put("count", pStats.snapshotBarriersInexact));
+    // #6132: a barrier that gave up before publishing a window leaves no other trace - every other snapshot counter
+    // needs a window to exist, and the consumer's fallback still completes the backup, just by throttling the writers
+    json.put("snapshotBarriersFailed", new JSONObject().put("count", pStats.snapshotBarriersFailed));
+    json.put("snapshotBarriersFailedSuspend", new JSONObject().put("count", pStats.snapshotBarriersFailedSuspend));
+    json.put("snapshotBarriersFailedFlush", new JSONObject().put("count", pStats.snapshotBarriersFailedFlush));
     json.put("deferredRAM", new JSONObject().put("space", pStats.deferredRAMBytes));
     json.put("pageFlushQueueWaits", new JSONObject().put("count", pStats.flushQueueWaits));
 
@@ -583,11 +588,13 @@ public class Profiler {
       // #6125: the t0 barrier on its own line. The average is what an operator compares against their write-latency
       // budget; the max is what a single unlucky backup actually cost, and averaging it away hides exactly the
       // outlier worth chasing.
-      buffer.append("%n    barriers=%d avg=%,dms max=%,dms inexact=%d overflowed=%d failed=%d".formatted(
+      buffer.append(
+        "%n    barriers=%d avg=%,dms max=%,dms inexact=%d windowsOverflowed=%d windowsFailed=%d barriersFailed=%d (suspend=%d flush=%d)".formatted(
           pStats.snapshotBarriers,
           pStats.snapshotBarriers > 0 ? pStats.snapshotBarrierMillis / pStats.snapshotBarriers : 0L,
           pStats.snapshotBarrierMaxMillis, pStats.snapshotBarriersInexact, pStats.snapshotWindowsOverflowed,
-          pStats.snapshotWindowsFailed));
+          pStats.snapshotWindowsFailed, pStats.snapshotBarriersFailed, pStats.snapshotBarriersFailedSuspend,
+          pStats.snapshotBarriersFailedFlush));
 
       buffer.append(
         "%n WAL totalFiles=%d pagesWritten=%d bytesWritten=%s".formatted(walTotalFiles, walPagesWritten,
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -1537,20 +1537,16 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
             throw e;
           forceBrokenChainDelete = true;
           logBrokenChainForceDelete(record.getIdentity(), e);
-        } catch (final ConcurrentModificationException e) {
-          // The record body could not be assembled for a consistent read, so its indexed keys and EXTERNAL pointers could
-          // not be read for cleanup. loadMultiPageRecord throws this after exhausting TX_RETRIES, but exhausted retries do
-          // NOT prove corruption: its page-version validation also fails when concurrent writes touch OTHER records
-          // sharing the chain's pages, so under a busy bucket this can be pure contention. Deleting anyway in that case
-          // would leak index entries for a healthy record. Disambiguate with a version-blind STRUCTURAL walk of the chunk
-          // chain: only a genuinely broken chain (a bad continuation pointer - the case that would otherwise make the
-          // record undeletable forever) takes the tolerant path below; transient contention rethrows, preserving the
-          // NeedRetryException semantics so the retry machinery re-runs the DELETE with intact index cleanup.
-          if (!tolerateBrokenChain || !bucket.isChunkChainBroken(record.getIdentity()))
-            throw e;
-          forceBrokenChainDelete = true;
-          logBrokenChainForceDelete(record.getIdentity(), e);
         }
+        // NO ConcurrentModificationException ARM, AND ITS REMOVAL IS THE POINT (#6282). It used to re-walk the chain
+        // with isChunkChainBroken and force the delete through when that walk agreed. That probe asks a STRICTLY
+        // WEAKER question than the one the loader has already answered above: it walks the newest committed image
+        // and stops there, where loadMultiPageRecord walks the same image AND proves the chunks this read consumed
+        // are still byte-for-byte the ones it holds - which is what rules out a chain caught half-published. So a
+        // CME arriving here is the loader saying "I could not confirm it", and overruling that with the weaker
+        // walk is how a HEALTHY record whose commit was mid-publication gets force-deleted with its index cleanup
+        // skipped. Propagate instead: the retry machinery re-runs the DELETE, and a record that really is corrupt
+        // comes back as BrokenChunkChainException on an attempt that is not racing a publication.
       }
 
       if (record instanceof Edge edge) {
@@ -1566,15 +1562,10 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
             throw e;
           logBrokenChainForcePhysicalDelete(record.getIdentity(), e);
           graphEngine.deleteVertex((VertexInternal) record, true);
-        } catch (final ConcurrentModificationException e) {
-          // The physical removal can raise the #4932 retry signal even when index cleanup did not (e.g. the type has no
-          // index left to read, so the broken chain is only discovered here). Fall back to force ONLY when the chain is
-          // confirmed structurally broken; a genuine transient conflict (or an already-forced delete) rethrows to retry.
-          if (!tolerateBrokenChain || forceBrokenChainDelete || !bucket.isChunkChainBroken(record.getIdentity()))
-            throw e;
-          logBrokenChainForcePhysicalDelete(record.getIdentity(), e);
-          graphEngine.deleteVertex((VertexInternal) record, true);
         }
+        // NO ConcurrentModificationException ARM: see the index-cleanup branch above (#6282). Both sources of a CME
+        // here have already asked the stronger question and declined to confirm - the loader for the read of the
+        // edge lists, deleteRecordInternal for the physical free - so the answer is a retry, not a force.
       } else {
         try {
           bucket.deleteRecord(record.getIdentity(), forceBrokenChainDelete);
@@ -1587,17 +1578,11 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
             throw e;
           logBrokenChainForcePhysicalDelete(record.getIdentity(), e);
           bucket.deleteRecord(record.getIdentity(), true);
-        } catch (final ConcurrentModificationException e) {
-          // STILL REACHABLE, and not redundant with the arm above: the delete's own walk raises this for a break it
-          // could NOT confirm - a chain caught mid-publication, or one that broke somewhere else in the committed
-          // image - and for the ordinary page conflicts of a busy bucket. The probe is what tells a chain that is
-          // genuinely broken (the case that would otherwise make the record undeletable forever) from contention,
-          // and since #6282 it asks the newest committed image rather than this transaction's pinned view.
-          if (!tolerateBrokenChain || forceBrokenChainDelete || !bucket.isChunkChainBroken(record.getIdentity()))
-            throw e;
-          logBrokenChainForcePhysicalDelete(record.getIdentity(), e);
-          bucket.deleteRecord(record.getIdentity(), true);
         }
+        // NO ConcurrentModificationException ARM: see the index-cleanup branch above (#6282). deleteRecordInternal
+        // raises this only for a break its own confirmation could NOT prove - a chain caught mid-publication, or one
+        // that broke somewhere else in the committed image - and for the ordinary page conflicts of a busy bucket.
+        // Every one of those is answered by retrying, none of them by deleting.
       }
 
       success = true;
--- a/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
+++ b/engine/src/main/java/com/arcadedb/database/LocalDatabase.java
@@ -1578,13 +1578,21 @@ public class LocalDatabase extends RWLockContext implements DatabaseInternal {
       } else {
         try {
           bucket.deleteRecord(record.getIdentity(), forceBrokenChainDelete);
+        } catch (final BrokenChunkChainException e) {
+          // THE ARM #6258 LEFT OUT AS DEAD CODE, LIVE SINCE #6282: deleteRecordInternal walks the chunk chain itself
+          // and never loads the record, so it used to report a break as the #4932 retry signal and this branch had
+          // to walk the chain a second time to find out which of the two it was holding. It now confirms the break
+          // against the newest committed image and says so, exactly as the loader does for the vertex branch above.
+          if (!tolerateBrokenChain || forceBrokenChainDelete)
+            throw e;
+          logBrokenChainForcePhysicalDelete(record.getIdentity(), e);
+          bucket.deleteRecord(record.getIdentity(), true);
         } catch (final ConcurrentModificationException e) {
-          // NO BrokenChunkChainException ARM HERE, unlike the vertex branch above, and the asymmetry is real rather
-          // than an omission (code review on #6258): deleteRecordInternal walks the chunk chain itself and never
-          // loads the record, so it reports a break as the #4932 retry signal and the structural probe below is
-          // still what tells the two apart. The vertex branch differs because reaching a vertex's edge lists means
-          // READING it, which is where the loader's own verdict comes from. Add the arm here the day
-          // deleteRecordInternal learns to name a broken chain as one.
+          // STILL REACHABLE, and not redundant with the arm above: the delete's own walk raises this for a break it
+          // could NOT confirm - a chain caught mid-publication, or one that broke somewhere else in the committed
+          // image - and for the ordinary page conflicts of a busy bucket. The probe is what tells a chain that is
+          // genuinely broken (the case that would otherwise make the record undeletable forever) from contention,
+          // and since #6282 it asks the newest committed image rather than this transaction's pinned view.
           if (!tolerateBrokenChain || forceBrokenChainDelete || !bucket.isChunkChainBroken(record.getIdentity()))
             throw e;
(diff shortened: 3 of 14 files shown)


lvca:
Round 4 answered.

**The one finding (delete walk does not validate chunk sizes) — intended scope, and the practical effect is smaller than the finding assumes.**

The size field sits *beside* the continuation pointer, not in front of it, so a chain whose only fault is a declared size still walks perfectly: the delete finds every chunk, frees every slot and simply succeeds. There is no break to name and no retry to exhaust. `Issue6282BrokenChainDeleteAndProbeTest.aBadChunkSizeAloneStillDeletesBecauseTheChainItselfIsIntact` now pins that from the delete side, alongside the parameterized test that already pins the read side naming it.

The only shape where the two walks disagree is a bad size **and** a bad pointer: the confirmation stops at the size, the delete stops at the pointer, the hops do not match, and the confirmation declines. That fails safe — the delete keeps its retryable exception, and `CHECK DATABASE FIX` force-deletes on its own detection, which does read sizes, so no record is left permanently stuck. Aligning the two hop indices would buy that shape a permanent verdict at the cost of more index bookkeeping in exactly the code that produced three of this PR's four review findings; the comment at the check now says so where the next reader will find it.

**Nit (Profiler line shape):** deliberate — `failed=` was ambiguous once a second failure counter existed, so it is now `windowsFailed=` / `barriersFailed=`. The JSON and Micrometer counters are the stable interfaces and are purely additive.

**Nit (JVM-wide injector):** agreed, non-blocking; recorded in its javadoc as the thing to scope per-test before the engine module ever runs tests in parallel.

lvca:
**`vector-unit-tests` red is pre-existing on `main`, not this PR.**

Pulled the `vector-unit-test-reports` artifact from both runs and grepped the surefire XMLs. This branch and its base commit `9cf4d262` fail the *identical* three tests:

- `Issue6772WriteBeforeSearchPrefixReuseTest.concurrentSearchersOnAWroteFirstReopenStillPublishExactlyOneReuse`
- `Issue6772WriteBeforeSearchPrefixReuseTest.aSessionThatInsertsBeforeItSearchesStillReusesThePersistedGraphAsAPrefix`
- `Issue6655StaleGraphPrefixReuseTest.reopenWithMoreVectorsThanThePersistedGraphReusesItAsAPrefixInsteadOfBlockingOnAFullRebuild`

(main run `33140801714` vs this branch's `33146767195`.) The diff here touches chunk-chain corruption detection and the snapshot barrier and has no overlap with vector indexing.

lvca:
**`integration-tests` red is pre-existing on `main` too.**

Same artifact check: this branch fails exactly one test, `com.arcadedb.console.RemoteConsoleIT.userMgmt`, and base commit `9cf4d262` fails the identical one (main run `33140801714`). It is the console `CREATE USER` syntax-error assertion at `RemoteConsoleIT.java:174`, which has nothing to do with chunk chains or the snapshot barrier.

So both red lanes on this PR — `vector-unit-tests` and `integration-tests` — reproduce on the base commit with no changes from this branch present.

lvca:
**`ha-integration-tests` red is the documented chronic flake (#5668 / #5702), not this change.**

- This branch fails `RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`; base commit `9cf4d262` fails a *different* HA test in the same window, `RaftReplicationChangeSchemaIT.schemaChangesReplicate`. "Red on a different test each run" is exactly the signature #5702 quantifies (~75% of main runs).
- The failure is a `TransactionCommittedRemotelyException` during a leader/replica restart convergence, i.e. replication timing.
- This PR's only `ha-raft` change is **comment-only** (7 insertions / 5 deletions, all Javadoc and `//` lines — verified with `git diff origin/main...HEAD -- ha-raft/` filtered to non-comment lines, which is empty). `RaftReplicatedDatabase` reads `PaginatedComponent.getTotalPages()`, not the file counter this PR changed, so there is no behavioural coupling either.

Will re-run the lane once the workflow's `coverage-report` job finishes and the run is rerunnable.

---

### CI summary for this head

| lane | result |
|---|---|
| unit-tests (carries both new test classes) | ✅ |
| slow-unit-tests | ✅ |
| build-and-package, lint, Codacy, CodeQL (all), Meterian | ✅ |
| opencypher-tck, builder, java/js/python/csharp/go e2e, studio-e2e, client-smoke | ✅ |
| vector-unit-tests | ❌ — identical 3 tests fail on base `9cf4d262` |
| integration-tests | ❌ — identical 1 test fails on base `9cf4d262` |
| ha-integration-tests | ❌ — chronic flake, different test than base's |

lvca:
**`ha-integration-tests`: the exact test failing here also fails on `main`.**

The re-run failed on the same test, so I checked main's HA history rather than assume. Surefire artifacts from the last seven `main` runs of *Java CI - test*, all red on this lane, on a rotating cast:

| main run | failing HA test(s) |
|---|---|
| 33140801714 | `RaftReplicationChangeSchemaIT.schemaChangesReplicate` |
| 33140566097 | `Issue5569SlotMergeDeleteRaftIT.mergedDeletesReplicateIntact` |
| 33135270998 | `Issue5276TransferLeaderValidationIT.targetedTransferReportsConfirmedLeader` |
| 33134733388 | `RaftHTTPGraphConcurrentIT.oneEdgePerTxMultiThreads` |
| **33128627747** | `Issue6091GrpcRoutingTableIT.theWriterIsNeverAlsoAReader`, **`RaftPriorityRejoinIT.leaderRestartThenReplicaRestartConverges`** |
| 33106842780 | `RaftHTTP2ServersCreateReplicatedDatabaseIT.createReplicatedDatabase`, `Issue5381SlotMergeRaftIT.distinctSlotUpdatesMergeAndReplicateIntact` |

Run **33128627747** (2026-08-28T00:06Z) fails the *identical* test this branch does, and predates the branch. That is #5702's "~75% of main runs, a different test each run" exactly.

I also checked the mechanism rather than resting on the statistics: the only change here that could touch a replication apply is `PaginatedComponentFile.totalPages`, and the two paths that could desync it are both safe — every physical page write in the engine goes through `PaginatedComponentFile.write()` (which advances the counter), and `SnapshotInstaller` closes the database before its file swap and reopens after, which re-seeds the counter from the real file length in `open()`. `RaftReplicatedDatabase` reads `PaginatedComponent.getTotalPages()`, not the file counter.

I could not reproduce locally: `RaftPriorityRejoinIT` binds port 2480 and an unrelated process is holding it on this machine.

PR #6872, opened 2026-08-28
fix(sql): native Select must honour its own timeout() and round-trip its JSON (#6815, #6816, #6817)

### LINKED ISSUE #6815: Native Select.timeout() is a silent no-op whenever the plan is answered from an index or from a single bucket
### Summary
The native fluent `Select` API installs the caller's timeout on one iterator shape only. Of the four sources `buildIterator()` can return, two are not `MultiIterator`s, so `timeout()` is silently dropped for every index-answered plan and for a single-bucket `fromBuckets(...)`. Nothing else on the serial path consults `select.timeoutInMs`.

### Where (HEAD `9c9d024db`)
`engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java:224-226`:
```java
if (select.timeoutInMs > 0 && iterator instanceof MultiIterator<? extends Identifiable> multiIterator)
  multiIterator.setTimeout(...);
```
The two sources that are **not** `MultiIterator`:
- the index-answered plan returns `iteratorFromIndexes`, a `MultiIndexCursor` (`engine/src/main/java/com/arcadedb/index/MultiIndexCursor.java:30` - `implements IndexCursor`)
- a single-bucket `fromBuckets(...)` returns `database.iterateBucket(...)` -> `bucket.iterator()` (`SelectExecutor.java:216`)

`grep timeoutInMs` finds only `Select.java`, `SelectExecutor.java:224`, `SelectParallelIterator.java` and `SelectCompiled.json()`. Neither `MultiIndexCursor` nor the bucket iterator has a timeout field or a `WorkGuard` hook. This affects `documents()`/`vertices()`/`edges()`, `count()` and `exists()` alike, since all go through `buildIterator()`.

### Repro
The index case is the common one - any `fromType` plus an indexed WHERE leaf:
```java
db.getSchema().createVertexType("V").createProperty("id", Type.INTEGER)
    .createIndex(Schema.INDEX_TYPE.LSM_TREE, true);
db.transaction(() -> { for (int i = 0; i < 100_000; i++) db.newVertex("V").set("id", i).save(); });

final SelectIterator<Vertex> it = db.select().fromType("V")
    .where().property("id").ge().value(0)          // -> range() cursor -> MultiIndexCursor
    .timeout(1, TimeUnit.MILLISECONDS, true).vertices();
while (it.hasNext()) { it.next(); Thread.sleep(1); }   // never throws TimeoutException
```
The same query with the WHERE on a non-indexed property does time out.

`SelectExecutionTest.errorTimeout()` (`engine/src/test/java/com/arcadedb/query/select/SelectExecutionTest.java:282`) documents the blind spot without catching it: its first block uses the unindexed `name` property (so it gets a `MultiIterator` and passes), and its second block (`:299-303`) builds an indexed `id < 10` select with a timeout and then never iterates it, so it asserts nothing.

### Suggested fix
Enforce the deadline in the consumer rather than in one source type: capture the deadline in `SelectIterator`/`SelectExecutor` and check it in `SelectIterator.fetchNext()` and in the `executeCount()`/`executeExists()` loops. `SelectParallelIterator` already does exactly this via `it.checkForTimeout()`. Keep the `MultiIterator.setTimeout()` call as the fast path.

### Scope / confidence
High. This is about the native fluent Select's own `timeout()`, not the SQL-side `arcadedb.command.timeout` covered by #6266/#6465.

### LINKED ISSUE #6816: MultiIterator.hasNextInternal() throws on timeout unconditionally, so Select.timeout(..., false) cannot truncate
### Summary
`MultiIterator` stores `exceptionOnTimeout` but its hot-path check ignores it and always throws. The only producer of `exceptionOnTimeout == false` in the engine is `Select.timeout(v, unit, false)`, so the caller asking for "stop early and give me what you have" gets a `TimeoutException` instead.

### Where (HEAD `9c9d024db`)
`engine/src/main/java/com/arcadedb/utility/MultiIterator.java:65-68`:
```java
private boolean hasNextInternal() {
  if (timeout > -1L && System.currentTimeMillis() - beginTime > timeout) {
    throw new TimeoutException("Timeout on iteration");
  }
```
`MultiIterator.java:253-259` is the method that honours the flag:
```java
public boolean checkForTimeout() {
  if (timeout > -1L && System.currentTimeMillis() - beginTime > timeout)
    if (exceptionOnTimeout)
      throw new TimeoutException("Timeout on iteration");
    else
      return true;
  return false;
}
```
but its non-throwing branch is unreachable from the iteration path: the two internal callers (`getNextPartial()`, `countEntries()`) are only reached after the `:66` check has already passed, i.e. only while the timeout has **not** expired.

`setTimeout(` has exactly two call sites in main: `LocalDatabase.java:834` (with `true`) and `SelectExecutor.java:225` (which forwards the user's flag). So `exceptionOnTimeout == false` is effectively dead.

### Repro
Use an unindexed WHERE so the source really is a `MultiIterator`:
```java
final SelectIterator<Vertex> it = db.select().fromType("Vertex")
    .where().property("name").eq().value("John")
    .timeout(1, TimeUnit.MILLISECONDS, false)      // false = do NOT throw
    .vertices();
final List<Vertex> got = new ArrayList<>();
while (it.hasNext()) { got.add(it.next()); Thread.sleep(2); }
// expected: the loop ends early and got holds a truncated result set
// actual:   com.arcadedb.exception.TimeoutException: Timeout on iteration
```
`SelectParallelIterator.fetchNext()` (`:180-188`) states the intended contract: "TIMEOUT WITHOUT EXCEPTION REQUESTED: STOP THE PRODUCERS AND RETURN WHAT WAS FETCHED SO FAR".

`SelectExecutionTest.errorTimeout()`'s second block (`SelectExecutionTest.java:299-303`) is the only test that passes `false`, and it never iterates the returned iterator, so this is unasserted today.

### Suggested fix
Replace the inline check in `hasNextInternal()` with `if (checkForTimeout()) return false;`. That keeps the throwing behaviour for every existing caller (`LocalDatabase.iterateType()` passes `exceptionOnTimeout = true`) and makes the non-throwing mode actually end the iteration.

Related: see the companion issue about `Select.timeout()` not being installed at all on the index-answered and single-bucket paths.

### Scope / confidence
High. Every `setTimeout(`/`checkForTimeout()` call site enumerated; no existing consumer can rely on `hasNext()` throwing while the flag is `false`, because nothing else produces `false`.

### LINKED ISSUE #6817: Select.json(JSONObject) always throws ArrayStoreException for a fromBuckets select, so the JSON round-trip is broken
### Summary
`Select.json(JSONObject)` maps the serialized bucket **names** to `Bucket` objects and then dumps that `List<Bucket>` into a `String[]`, which throws `ArrayStoreException` at runtime for every non-empty bucket list. The same method also drops fields that `SelectCompiled.json()` writes, so `json()` and `json(JSONObject)` are not inverse.

### Where (HEAD `9c9d024db`)
`engine/src/main/java/com/arcadedb/query/select/Select.java:195-200`:
```java
} else if (json.has("fromBuckets")) {
  final JSONArray buckets = json.getJSONArray("fromBuckets");
  fromBuckets(
      buckets.toList().stream().map(b -> database.getSchema().getBucketByName(b.toString())).collect(Collectors.toList())
          .toArray(new String[buckets.length()]));
}
```
`List<Bucket>.toArray(new String[n])` compiles because `<T> T[] toArray(T[])`'s type variable is independent of the element type. At runtime, with `n == size`, `ArrayList` takes the `System.arraycopy` branch, which stores a `Bucket` into a `String[]` and raises `ArrayStoreException`. The `map(...)` step is pointless anyway: `fromBuckets(String...)` (`Select.java:96`) resolves the names itself.

Same method, silently dropped state: `timeoutInMs`/`exceptionOnTimeout` are written by `SelectCompiled.json()` (`SelectCompiled.java:67-70`) and never read back, and `orderBy`/`parallel` are neither written nor read. A round-tripped select therefore loses its timeout and its ORDER BY.

### Repro
```java
final String bucket = db.getSchema().getType("Vertex").getBuckets(false).getFirst().getName();
final JSONObject json = db.select().fromBuckets(bucket)
    .where().property("name").eq().value("John").compile().json();
db.select().json(json);          // java.lang.ArrayStoreException: com.arcadedb.engine.LocalBucket
```
`SelectExecutionTest.okJSON()` (`engine/src/test/java/com/arcadedb/query/select/SelectExecutionTest.java:432`) is the only round-trip test and it exercises only `fromType`, which is why this has stayed invisible. The only non-throwing input is an empty `fromBuckets` array, which `compile()` rejects upstream anyway.

### Suggested fix
Pass the names straight through:
```java
fromBuckets(buckets.toList().stream().map(Object::toString).toArray(String[]::new));
```
and read back `timeoutInMs`/`exceptionOnTimeout` (plus serialize and parse `orderBy`) so the two `json` methods are actually inverse.

### Scope / confidence
High. Verified that `JSONObject.put(String, Iterable)` stores the names as a `JSONArray` of strings, so the `map` really does produce a `List<Bucket>`, and that `Collectors.toList()` yields an `ArrayList` whose `toArray(T[])` with `a.length == size` uses the throwing `System.arraycopy` path.

robfrank:
Fixes three bugs in the native fluent `Select` API, all reported against `9c9d024db`. They are grouped in one PR because the first two are the same defect seen from opposite ends (a timeout that is never installed, and a timeout that ignores the flag telling it how to fire) and all three land in the same four files.

Closes #6815
Closes #6816
Closes #6817

## #6815 - `Select.timeout()` was a silent no-op for index-answered and single-bucket plans

`SelectExecutor.buildIterator()` can return four different sources, and only one of them is a `MultiIterator`:

| plan shape | source | carried a deadline? |
|---|---|---|
| index-answered (`fromType` + an indexed WHERE leaf) | `MultiIndexCursor` | no |
| `fromType`, unindexed WHERE | `LocalDatabase.iterateType()` -> `MultiIterator` | yes |
| `fromBuckets(one)` | `bucket.iterator()` | no |
| `fromBuckets(many)` | `MultiIterator` | yes |

The timeout was installed with `multiIterator.setTimeout(...)` under an `instanceof` guard, so the two rows that are not a `MultiIterator` dropped it entirely - including the index-answered plan, by far the commonest shape. Nothing else on the serial path read `select.timeoutInMs`, so `documents()`/`vertices()`/`edges()`, `count()` and `exists()` were all affected.

**Fix:** the deadline now belongs to the consumer, which every source shape passes through, instead of to the one source that knew how to enforce it.

- `SelectExecutor` gains `startTimeout()` (armed at the head of `execute()`, `executeCount()` and `executeExists()`, so index lookup counts against the budget) and `checkForTimeout()`, which throws or reports expiry per `exceptionOnTimeout` - the same contract as `MultiIterator.checkForTimeout()`.
- `SelectIterator.fetchNext()` checks it once per candidate record, so a long run of non-matching rows is cut short too, not just a slow consumer.
- `executeCount()` and `executeExists()` check it in their own drain loops, since neither builds a `SelectIterator`.
- The `MultiIterator.setTimeout()` call stays as the fast path, and because `SelectParallelIterator` drives its producer shutdown off that iterator's own `checkForTimeout()`.

`Long.MAX_VALUE` encodes "no timeout requested", so a select without `timeout()` pays one perfectly-predicted compare per record.

## #6816 - `MultiIterator` ignored `exceptionOnTimeout` and always threw

`hasNextInternal()` carried an inline copy of the deadline check that threw unconditionally. `checkForTimeout()` - the method that honours the flag - was unreachable from the iteration path: its two internal callers (`getNextPartial()`, `countEntries()`) are only reached *after* the inline check has already passed, i.e. only while the deadline has **not** expired. So `Select.timeout(v, unit, false)`, meaning "stop early and give me what you have", got a `TimeoutException` instead, and the non-throwing branch was dead code.

**Fix:** `hasNextInternal()` delegates to `checkForTimeout()` and returns `false` when it reports expiry.

No existing caller can see a behaviour change: `setTimeout(` has exactly two producers in main, `LocalDatabase.java:834` (hard-coded `true`) and `SelectExecutor` (forwards the user's flag), so `exceptionOnTimeout == false` was previously unreachable in production.

## #6817 - `Select.json(JSONObject)` threw `ArrayStoreException`, and was not the inverse of `json()`

Four separate defects in the round trip:

1. **`ArrayStoreException` on every `fromBuckets` select.** The method mapped the serialized bucket *names* to `Bucket` objects and then poured that `List<Bucket>` into a `String[]`. It compiles because `<T> T[] toArray(T[])`'s type variable is independent of the element type; at runtime, with `a.length == size`, `ArrayList` takes the `System.arraycopy` branch and blows up. The `map` step was pointless anyway - `fromBuckets(String...)` resolves the names itself. Now the names are passed straight through.
2. **`timeoutInMs`/`exceptionOnTimeout` were written by `SelectCompiled.json()` and never read back**, so a round-tripped select ran with no deadline.
3. **`orderBy` and `parallel` were neither written nor read.** Both are now serialized (only when non-default, so an unaffected select produces byte-identical JSON) and parsed back. `parallel` is set on the field directly rather than through `SelectCompiled.parallel()`, which is only reachable after `compile()`.
4. **A single-condition WHERE could not be read back at all.** `compile()` wraps a lone leaf in a synthetic `run` root whose operator `SelectTreeNode.toJSON()` omits and whose right side is null, producing a one-element array that `parseJsonCondition()` rejected with `Invalid condition`. That is the commonest select shape there is; it is now unwrapped, and `compile()` re-adds the same root on the way out.

## Verification

New regression tests, each written to fail against the current `main` first:

- `Issue6815SelectTimeoutNotInstalledTest` (9 tests) - index-answered and single-bucket iteration, throwing and truncating, plus `count()` and `exists()`, plus a control asserting a select with no timeout still returns everything.
- `Issue6816MultiIteratorTimeoutTruncationTest` (5 tests) - three of them drive `MultiIterator` directly with an already-expired deadline, so they are fully deterministic; two cover the end-to-end select shape.
- `Issue6817SelectJsonRoundTripTest` (9 tests) - single and multi bucket, timeout (both flags), single and multiple `orderBy`, `parallel`, plus a control asserting defaults still produce no extra keys.

`SelectExecutionTest.errorTimeout()`'s second block, which built an iterator with `exceptionOnTimeout = false` and then never iterated it (asserting nothing - which is exactly how both timeout bugs stayed invisible), is now armed.

Timing discipline per `CLAUDE.md`: no assertion measures elapsed time. Every timeout assertion is one-sided in the safe direction - "this bounded operation gave up" or "the result set was cut short" - so a JVM stall can only make it more true, never flip it red. The two `count()`/`exists()` tests scan 20,000 rows against a 1 ms budget for the same reason.

**Test runs (all on this branch):**
- `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` -> **13,760 tests, 0 failures, 0 errors**
- `mvn -o -pl engine test -Dgroups=slow` -> 261 tests, **1 failure: `Issue6459GraphPathfindingCommandTimeoutTest.dijkstraHonoursTheCommandTimeoutOnAnUnreachableDestination`**

That one failure is **pre-existing and unrelated**, proved rather than assumed:

- A control worktree checked out at unmodified `origin/main` (`9cf4d2623b`) reproduces the *identical* single failure on the same 261-test lane.
- The class passes 3/3 in isolation on both `main` and this branch, so it only fails in the warm shared JVM of the full lane - the SQL search finishes inside its 50 ms `command.timeout` budget on a fast machine, and the test asserts that it must NOT finish.
- Statically it cannot be reached by this diff: `SQLFunctionAstar` and `SQLFunctionDijkstra` reference none of `MultiIterator`, `SelectIterator`, `SelectExecutor` or `database.select()`. The only SQL-visible caller of `MultiIterator.setTimeout()` is `LocalDatabase.iterateType()`, which passes `exceptionOnTimeout = true` - the branch #6816 leaves byte-for-byte unchanged.

## Compatibility

- The native `Select` timeout is unrelated to the SQL-side `arcadedb.command.timeout` of #6266/#6465; neither is touched.
- No production code consumes the `Select` JSON round trip - `Select.json(JSONObject)` has no caller in main outside tests - so #6817 is a self-contained API fix.
- `SelectCompiled.json()` emits `orderBy`/`parallel` only when they are non-default, so a select that sets neither serializes to exactly the same document as before (a control test asserts this).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query timeout handling for result retrieval, counting, and existence checks.
  * Queries configured to avoid timeout exceptions now stop cleanly and return partial results when applicable.
  * Fixed serialized query round trips to preserve bucket selection, ordering, timeout, and parallel execution settings.
  * Improved parsing of standalone `WHERE` conditions in serialized queries.

* **Tests**
  * Added coverage for timeout behavior, partial results, and serialized query round trips.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/select/Select.java
+++ b/engine/src/main/java/com/arcadedb/query/select/Select.java
@@ -192,12 +192,12 @@ public class Select {
       fromType(json.getString("fromType"));
       if (json.has("polymorphic"))
         polymorphic(json.getBoolean("polymorphic"));
-    } else if (json.has("fromBuckets")) {
-      final JSONArray buckets = json.getJSONArray("fromBuckets");
-      fromBuckets(
-          buckets.toList().stream().map(b -> database.getSchema().getBucketByName(b.toString())).collect(Collectors.toList())
-              .toArray(new String[buckets.length()]));
-    }
+    } else if (json.has("fromBuckets"))
+      // #6817: fromBuckets(String...) RESOLVES THE NAMES ITSELF, SO MAPPING THEM TO Bucket OBJECTS FIRST WAS BOTH
+      // POINTLESS AND FATAL: List<Bucket>.toArray(new String[size]) COMPILES - <T> T[] toArray(T[])'S TYPE VARIABLE
+      // IS INDEPENDENT OF THE ELEMENT TYPE - AND THEN TAKES ArrayList'S System.arraycopy BRANCH, WHICH RAISES
+      // ArrayStoreException FOR EVERY NON-EMPTY BUCKET LIST. PASS THE NAMES STRAIGHT THROUGH INSTEAD
+      fromBuckets(json.getJSONArray("fromBuckets").toList().stream().map(Object::toString).toArray(String[]::new));
 
     if (json.has("where")) {
       where();
@@ -209,10 +209,34 @@ public class Select {
     if (json.has("skip"))
       skip(json.getInt("skip"));
 
+    // #6817: THE STATE BELOW IS WRITTEN BY SelectCompiled.json() BUT USED TO BE DROPPED HERE, SO json() AND
+    // json(JSONObject) WERE NOT INVERSE - A ROUND-TRIPPED SELECT RAN WITH NO DEADLINE AND NO ORDER BY
+    if (json.has("orderBy")) {
+      final JSONArray parsedOrderBy = json.getJSONArray("orderBy");
+      for (int i = 0; i < parsedOrderBy.length(); i++) {
+        final JSONObject entry = parsedOrderBy.getJSONObject(i);
+        orderBy(entry.getString("property"), entry.getBoolean("ascending", true));
+      }
+    }
+    if (json.has("timeoutInMs"))
+      timeout(json.getLong("timeoutInMs"), TimeUnit.MILLISECONDS, json.getBoolean("exceptionOnTimeout", false));
+    if (json.getBoolean("parallel", false))
+      // NOT VIA SelectCompiled.parallel(): THAT IS ONLY REACHABLE AFTER compile(), AND THIS SELECT IS STILL BEING BUILT
+      parallel = true;
+
     return this;
   }
 
   private void parseJsonCondition(final JSONArray condition) {
+    // #6817: A SELECT WITH A SINGLE WHERE LEAF SERIALIZES AS A ONE-ELEMENT ARRAY - compile() WRAPS THE LEAF IN A
+    // SYNTHETIC `run` ROOT WHOSE OPERATOR SelectTreeNode.toJSON() DELIBERATELY OMITS AND WHOSE RIGHT SIDE IS null.
+    // REJECTING THAT SHAPE MADE THE COMMONEST SELECT OF ALL IMPOSSIBLE TO READ BACK, SO UNWRAP IT INSTEAD: compile()
+    // RE-ADDS THE SAME `run` ROOT ON THE WAY OUT, WHICH IS WHAT MAKES THE TWO json METHODS INVERSE FOR THIS SHAPE
+    if (condition.length() == 1 && condition.get(0) instanceof JSONArray nested) {
+      parseJsonCondition(nested);
+      return;
+    }
+
     if (condition.length() != 3)
       throw new IllegalArgumentException("Invalid condition " + condition);
 
--- a/engine/src/main/java/com/arcadedb/query/select/SelectCompiled.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectCompiled.java
@@ -20,7 +20,9 @@ package com.arcadedb.query.select;/*
 import com.arcadedb.database.Document;
 import com.arcadedb.graph.Edge;
 import com.arcadedb.graph.Vertex;
+import com.arcadedb.serializer.json.JSONArray;
 import com.arcadedb.serializer.json.JSONObject;
+import com.arcadedb.utility.Pair;
 
 import java.util.HashMap;
 import java.util.stream.Collectors;
@@ -68,6 +70,18 @@ public class SelectCompiled {
       json.put("timeoutInMs", select.timeoutInMs);
       json.put("exceptionOnTimeout", select.exceptionOnTimeout);
     }
+
+    // #6817: orderBy AND parallel WERE NEITHER WRITTEN HERE NOR READ BACK BY Select.json(JSONObject), SO A
+    // ROUND-TRIPPED SELECT SILENTLY LOST ITS ORDERING AND RAN SERIALLY. ONLY NON-DEFAULT STATE IS EMITTED, SO A
+    // SELECT THAT NEVER SET EITHER STILL SERIALIZES TO EXACTLY THE SAME DOCUMENT AS BEFORE
+    if (select.orderBy != null && !select.orderBy.isEmpty()) {
+      final JSONArray orderBy = new JSONArray();
+      for (final Pair<String, Boolean> entry : select.orderBy)
+        orderBy.put(new JSONObject().put("property", entry.getFirst()).put("ascending", entry.getSecond()));
+      json.put("orderBy", orderBy);
+    }
+    if (select.parallel)
+      json.put("parallel", select.parallel);
 
     return json;
   }
--- a/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
+++ b/engine/src/main/java/com/arcadedb/query/select/SelectExecutor.java
@@ -21,6 +21,7 @@ import com.arcadedb.database.Document;
 import com.arcadedb.database.Identifiable;
 import com.arcadedb.database.RID;
 import com.arcadedb.engine.Bucket;
+import com.arcadedb.exception.TimeoutException;
 import com.arcadedb.index.Index;
 import com.arcadedb.index.IndexCursor;
 import com.arcadedb.index.IndexInternal;
@@ -49,6 +50,14 @@ public class SelectExecutor {
   // skip) TO A MultiIndexCursor
   int             indexCandidateLimit = -1;
 
+  // #6815: THE ABSOLUTE DEADLINE FOR THIS EXECUTION, OWNED BY THE CONSUMER INSTEAD OF BY THE SOURCE ITERATOR.
+  // buildIterator() CAN RETURN FOUR DIFFERENT SOURCES AND ONLY ONE OF THEM (MultiIterator) CARRIES A TIMEOUT OF ITS
+  // OWN, SO INSTALLING select.timeoutInMs THERE - THE ONLY THING THAT USED TO HAPPEN - SILENTLY DROPPED IT FOR EVERY
+  // INDEX-ANSWERED PLAN (A MultiIndexCursor) AND FOR A SINGLE-BUCKET fromBuckets() (A BARE BUCKET ITERATOR), I.E. FOR
+  // THE COMMONEST QUERY SHAPE OF ALL. Long.MAX_VALUE MEANS "NO TIMEOUT REQUESTED" AND KEEPS THE PER-RECORD CHECK A
+  // SINGLE PERFECTLY PREDICTED COMPARE ON THE HOT PATH
+  private long timeoutDeadline = Long.MAX_VALUE;
+
   // #6577: THE SINGLE SOURCE OF TRUTH FOR "WHICH OPERATORS CAN filterWithIndexesFinalNode() ACTUALLY TURN INTO AN
   // IndexCursor" - isTheNodeFullyIndexed() MUST TREAT EXACTLY THIS SET, AND NO MORE, AS "THIS LEAF IS INDEXED".
   // TREATING A WIDER SET AS INDEXED IS WHAT LET #6577 THROUGH: A neq/like/ilike LEAF PASSED THE "UNDER AN or, BOTH
@@ -74,7 +83,33 @@ public class SelectExecutor {
     this.select = select;
   }
 
+  /**
+   * Arms the deadline for one execution. Called from the three entry points ({@link #execute()},
+   * {@link #executeCount()}, {@link #executeExists()}) rather than from the constructor, so the window it covers is
+   * exactly the work the caller asked to bound - index lookup included, since that is part of answering the query.
+   */
+  private void startTimeout() {
+    if (select.timeoutInMs > 0)
+      timeoutDeadline = System.currentTimeMillis() + select.timeoutInMs;
+  }
+
+  /**
+   * #6815: enforces the select deadline on the consumer side, where every source shape passes, instead of on the one
+   * source that happens to know how to enforce it itself. Mirrors {@link com.arcadedb.utility.MultiIterator#checkForTimeout()}:
+   * with {@code exceptionOnTimeout} the expiry is thrown, otherwise the caller is told to stop and keep what it has.
+   *
+   * @return {@code true} when the deadline expired and the caller asked NOT to be thrown at
+   */
+  boolean checkForTimeout() {
+    if (timeoutDeadline == Long.MAX_VALUE || System.currentTimeMillis() <= timeoutDeadline)
+      return false;
+    if (select.exceptionOnTimeout)
+      throw new TimeoutException("Timeout on iteration");
+    return true;
+  }
+
   <T extends Document> SelectIterator<T> execute() {
+    startTimeout();
     final MultiIndexCursor iteratorFromIndexes = lookForIndexes();
     final Iterator<? extends Identifiable> iterator = buildIterator(iteratorFromIndexes);
 
@@ -90,6 +125,7 @@ public class SelectExecutor {
    * would keep a retired compacted file undroppable until the next restart.
    */
   long executeCount() {
+    startTimeout();
     final MultiIndexCursor iteratorFromIndexes = lookForIndexes();
     try {
       final Iterator<? extends Identifiable> iterator = buildIterator(iteratorFromIndexes);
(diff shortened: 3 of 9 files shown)


robfrank:
Review cycle 1 addressed.

**CodeRabbit (2 nitpicks, both applied)** - `c8692288c9`. Both were valid against `CLAUDE.md`'s "don't use fully qualified names if possible, always import the class and just use the name": `com.arcadedb.engine.Bucket::getName` and the `{@link com.arcadedb.utility.MultiIterator}` Javadoc reference are now imported. Test-only, no behaviour change.

**Claude review - the `executeVector()` gap.** Verified and correct: `SelectVectorBuilder.vertices()`/`documents()` call `new SelectExecutor(select).executeVector()`, which touches neither `startTimeout()` nor `checkForTimeout()`, and since `timeout()` returns `this` with `nearestTo()` declared on `Select`, `select().fromType(t).timeout(...).nearestTo(...).vertices()` compiles and drops the deadline. A fifth plan shape beyond the four `buildIterator()` covers.

I implemented the fix (arm at the head of `executeVector()`, check between indexes and per assembled result), then **pulled it back out and filed #6873 instead** - because I could not test it without introducing a flaky assertion, and shipping it unverified is worse than shipping it separately.

The reason is structural: `executeVector()` is eager. The iterator tests in this PR are stable precisely because the consumer paces the lazy iterator with its own `Thread.sleep()`, so elapsed time is under the test's control and every assertion is one-sided in the safe direction. An eager call offers no such hook, so "a 1 ms budget expired" becomes a wall-clock race that a *faster* machine turns red - the direction `CLAUDE.md` rules out.

That is not hypothetical. The candidate test passed 5/5 in isolation on a cold JVM, then failed in the warm full-suite run. Measuring it (warm, best of 10, dim 8, `k == n`, post-filter WHERE):

| products | best elapsed |
|---|---|
| 50 | 0.25 ms |
| 1,000 | 1.29 ms |
| 5,000 | 3.67 ms |

Even 5,000 vectors is only a ~3.7x margin over a 1 ms budget, and buying a bigger margin with a much larger fixture is a bad trade. #6873 carries the fix, the measurements, and the two options for making it testable.

So this PR stays scoped to the four `buildIterator()` sources that #6815 actually names, with every behaviour change covered by a test that was verified to fail first.

**Re-verified after the change:** `Issue6815*` + `Issue6816*` + `Issue6817*` + `SelectExecutionTest` + `SelectVectorTest` -> 55 tests, 0 failures.

robfrank:
Review cycle 2 addressed - `91e4e174b9`.

**2. `startTimeout()` overflow - real, reachable, fixed.**

Verified rather than assumed. `TimeUnit.toMillis()` *saturates* an out-of-range conversion to `Long.MAX_VALUE` instead of rejecting it, so this is reachable straight from the public fluent API:

```
DAYS.toMillis(Long.MAX_VALUE) = 9223372036854775807  (saturated)
now + that                    = -9223370248951018345 (wrapped)
now > deadline                = true                 -> expired before the first record
```

So `timeout(Long.MAX_VALUE, DAYS, true)` - an effectively infinite budget - threw `TimeoutException` on the very first record. Worse than the inconsistency it looked like.

Fixed by clamping instead of wrapping, which lands on `Long.MAX_VALUE`, already the "no timeout" sentinel, and is exactly what a budget that large means:

```java
final long now = System.currentTimeMillis();
timeoutDeadline = select.timeoutInMs > Long.MAX_VALUE - now ? Long.MAX_VALUE : now + select.timeoutInMs;
```

Two new tests in `Issue6815SelectTimeoutNotInstalledTest` cover it across `vertices()`, `count()`, `exists()` and the single-bucket source. They are fully deterministic - the assertion is "an infinite budget returns every row", which no clock reading can flip.

**3. `executeVector()`** - agreed, and already filed as #6873 during cycle 1, with the fix, the measured timings, and the reason it could not ship here (an eager call gives a test no way to control elapsed time, so the expiry assertion becomes a wall-clock race a *faster* machine turns red - it flaked exactly that way before I caught it).

**1. Two coexisting timeout mechanisms** - declining for this PR, with reasoning rather than dismissal. It is a pure refactoring of code that is behaviourally correct today, it would widen a three-bug fix into an API cleanup, and the duplication is deliberate and already documented at both sites (`MultiIterator`'s copy has to stay because `SelectParallelIterator` drives producer shutdown off it). Worth doing, but as its own change where it can be reviewed on its merits. Happy to file it if you would like it tracked.

**Re-verified:** `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` -> **13,762 tests, 0 failures, 0 errors**.

robfrank:
Cycle 3 - clean approval, no changes requested. Closing the review loop at 3 of 4 cycles. Both remaining observations are noted, neither changes this PR:

**1. `executeVector()`** - already filed as #6873 during cycle 1, carrying the three-line fix, the measured timings, and the reason it could not ship here (an eager call gives a test no hook to control elapsed time, so the expiry assertion degrades into a wall-clock race a *faster* machine turns red - it flaked exactly that way in the warm full-suite run before I caught it).

**2. `parseJsonCondition` recursion depth** - agreed it is worth flagging, and agreed it does not belong here. Two reasons: the unbounded recursion is pre-existing rather than introduced (the `parsedLeft instanceof JSONArray -> parseJsonCondition(array)` branch already recursed once per nesting level; the new one-element unwrap adds a branch, not a new class of exposure), and the input has to survive the JSON parser's own nesting limit before it ever reaches this method. There is a clear precedent for a real guard - `arcadedb.sql.maxExpressionDepth` (default 200) exists for exactly this shape in the SQL parser - so if `Select.json(JSONObject)` ever gets a production consumer that is the pattern to copy. Say the word and I will file it; leaving it unfiled for now rather than adding a guard to a method nothing calls.

**Final state of the branch** (`91e4e174b9`):

| | |
|---|---|
| `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` | 13,762 tests, 0 failures, 0 errors |
| `mvn -o -pl engine test -Dgroups=slow` | 261 tests, 1 failure - reproduced identically on unmodified `main` via a control worktree, and statically unreachable from every file this PR touches |
| New regression tests | 25, each verified to fail before its fix |

Review cycles: 3 - cycle 1 two import nitpicks, cycle 2 a real overflow bug found in this PR's own new code (an infinite `timeout()` threw on the first record), cycle 3 clean.

Issue #4595, opened 2026-06-12
Bucket alignment broken for negative timestamps (time-series)


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.engine.timeseries.TimeSeriesEngine` /
`MultiColumnAggregationResult` / `TimeSeriesSealedStore`
**Severity:** MEDIUM

### Summary

Java integer division truncates toward zero, not toward negative
infinity. For `ts = -1500, bucketIntervalMs = 1000`, the expected
bucket is `-2000` but `(-1500 / 1000) * 1000 = -1000`. The flat-index
calculation in `MultiColumnAggregationResult` then mis-attributes a
row to the wrong bucket.

### Code reference

`engine/com/arcadedb/engine/timeseries/TimeSeriesEngine.java:260, 325, 406, 447`;
`engine/com/arcadedb/engine/timeseries/MultiColumnAggregationResult.java:374`;
`engine/com/arcadedb/engine/timeseries/TimeSeriesSealedStore.java:859`

### Impact

Any time-series with negative epoch timestamps (synthetic/test data,
or measurements expressed relative to an offset) gets wrong bucketing.

### Suggested fix

Use `Math.floorDiv(ts, bucketIntervalMs) * bucketIntervalMs`
everywhere.

lvca:
Fixed.

PR #6878, opened 2026-08-28
fix(#6824,#6839): floor pre-epoch time buckets, and give the engine-unavailable TimeSeries state a way out

### LINKED ISSUE #6824: ts.timeBucket() truncates toward zero, so pre-1970 timestamps get a bucket start AFTER the input
### Summary
`ts.timeBucket()` computes the bucket boundary with Java integer division, which truncates toward zero rather than flooring. For any negative epoch value the returned "bucket start" is **later** than the timestamp it is supposed to contain, and the last pre-epoch bucket collapses into the first post-epoch one.

### Where (HEAD `9c9d024db`)
`engine/src/main/java/com/arcadedb/function/sql/time/SQLFunctionTimeBucket.java:66-71`:
```java
final long timestampMs = toEpochMs(params[1]);

// Truncate to bucket boundary
final long bucketStart = (timestampMs / intervalMs) * intervalMs;

return new Date(bucketStart);
```
`toEpochMs()` (`:105-121`) accepts negative `Long`/`Number`/`Date`/`Instant` inputs, so nothing upstream rejects them, and the `intervalMs <= 0` guard added for #6388 only guards the divisor.

### Repro
```sql
SELECT ts.timeBucket('1h', -1800000) AS b   -- input 1969-12-31T23:30:00Z
```
returns `1970-01-01T00:00:00Z`, i.e. 30 minutes **after** the input. Correct answer: 1969-12-31T23:00:00Z. Likewise `SELECT ts.timeBucket('1d', -1)` returns 1970-01-01 instead of 1969-12-31.

Used as a `GROUP BY` key (the documented usage, and the shape exercised in `engine/src/test/java/com/arcadedb/engine/timeseries/TimeSeriesAggregationPushDownTest.java:71`) this silently merges the last pre-epoch bucket into the first post-epoch bucket, producing wrong aggregates for any historical series that crosses 1970. `Issue6388TimeFunctionArgumentTest.java:134` covers only a positive timestamp, so nothing asserts the current behaviour.

### Suggested fix
```java
final long bucketStart = Math.floorDiv(timestampMs, intervalMs) * intervalMs;
```
plus a negative-timestamp case in `Issue6388TimeFunctionArgumentTest`.

### Scope / confidence
High. #6388 is the only other `timeBucket` issue and is about argument validation, not the truncation direction.

### LINKED ISSUE #6839: Follow-up on #6356: the engine-unavailable TimeSeries state has no recovery path, HA discards the sealed blob that would repair it, and Prometheus write reports "already exists"
Follow-up on #6356 / #6779. The new "registered but engine-unavailable" TimeSeries state is now visible, but there is no path out of it, and one call site turns it into a misleading error.

### 1. The engine-unavailable state has no recovery path, and HA discards the blob that would repair it

`2a4526fa7` (#6356) made a TimeSeries type whose `initEngine()` throws at schema load stay **registered** with `engine == null` instead of vanishing. `LocalSchema.java:1885-1895` justifies that on the grounds that the failing file "(a `.ts.sealed` most commonly, rebuildable under HA by recompacting the replicated mutable pages)" can be repaired.

That repair path does not exist:

- The Raft apply path that installs a leader-shipped sealed file refuses exactly this state and throws the payload away - `ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java:2000-2005`:
```java
if (!(schema.getType(blob.typeName()) instanceof LocalTimeSeriesType tsType) || tsType.getEngine() == null) {
  LogManager.instance().log(this, Level.SEVERE,
      "Received TimeSeries sealed blob for non-timeseries or uninitialized type '%s' (db=%s); skipping", ...);
  continue;
}
```
so `installSealedFileBytes` at `:2006` is never reached. A Raft entry is applied once and not re-shipped, so each dropped blob is a permanent divergence.
- Nothing ever retries `initEngine()`. Its only non-test callers are the schema-load site (`LocalSchema.java:1880`) and type creation (`TimeSeriesTypeBuilder.java:139`).
- Maintenance/recompaction, the other route that could rebuild the sealed store, is deliberately gated off for the type: `if (tsType.isEngineAvailable())` (`LocalSchema.java:1902`).
- Every read and write now throws through `requireEngine()`.

Repro: on a Raft follower, corrupt one byte of a `<type>.ts.sealed` file and restart that node. `initEngine()` fails, the type is registered with no engine (the new #6356 log line "the type is registered but its storage is unavailable until this is resolved"), and from then on every query/insert throws `DatabaseOperationException` from `LocalTimeSeriesType.requireEngine()` (`:120`); `CHECK DATABASE` reports the warning (`DatabaseChecker.java:725`) but cannot fix it; and every subsequent `TsSealedBlob` the leader replicates for that type is logged SEVERE and discarded. The only escape is closing and reopening the whole database after fixing the file by hand, i.e. exactly the restart #6356 set out to make unnecessary. `Issue6340TimeSeriesCheckDatabaseTest` covers the new state but not a recovery from it.

*Fix, minimal:* in `installTsSealedBlobs`, when the type is a `LocalTimeSeriesType` with no engine, call `initEngine()` (already `synchronized`, idempotent, and clearing `engineUnavailableReason` on success since `7e76c52ed`) and install the blob if that now succeeds, rather than `continue`.
*Fix, better:* at schema load, when the failure is confined to a derived sealed store, quarantine that file and initialise the engine over the mutable pages so the type comes up degraded-but-usable, and expose a `CHECK DATABASE ... FIX` / `REBUILD` action that re-runs `initEngine()`.

Rated medium rather than high because a full database close/reopen does re-run `readConfiguration()` and would retry: the state is "unrecoverable without a reopen, and replicated sealed blobs are lost in the meantime".

### 2. Prometheus remote-write reports "Type 'X' already exists" for the same state

PR #6779 converted the TimeSeries read/write call sites to the new `isEngineAvailable()`/`requireEngine()` pair (`SaveElementStep:83`, `SelectExecutionPlanner:1910`, `LocalDatabase:742`, `DatabaseAsyncExecutorImpl:1533`/`:1543`, `PostTimeSeriesWriteHandler:188`/`:211`, `PostTimeSeriesQueryHandler:82`, `GetTimeSeriesLatestHandler:70`, `PostGrafanaQueryHandler:97`). `PostPrometheusWriteHandler.getOrCreateType` was missed, and its pre-existing shape turns the new state into the wrong error - `server/src/main/java/com/arcadedb/server/http/handler/PostPrometheusWriteHandler.java:176`:
```java
if (docType instanceof LocalTimeSeriesType tsType && tsType.getEngine() != null)
  return tsType;
```
A registered type with `engine == null` fails that guard and falls through to the auto-create branch `return builder.create();` (`:191`), which throws `SchemaException("Type '" + typeName + "' already exists")` (`TimeSeriesTypeBuilder.java:112-113`).

So `POST /api/v1/prometheus/write` for that metric reports a phantom name collision instead of the specific `TimeSeries type 'X' has no storage engine available: <cause>` that `requireEngine()` was added to produce. No data damage: `create()` throws before mutating the schema.

*Fix:* split the two conditions in `getOrCreateType` - if the existing type is a `LocalTimeSeriesType`, return it and let the caller's `requireEngine()` (replacing `tsType.getEngine()` at `:122`) report the real reason. Keep the auto-create fall-through only for a name that does not exist or is not a TimeSeries type.

### Also reviewed and clean
For the record, since these were the promising leads in the same churn window: `DeltaOverlay.merge()`'s withdraw-the-add branch (#6775/#6777) is safe because a slot freed by a delete only becomes allocatable after commit-time page compression reports its space (`LocalBucket.java:1786-1788`); `PaginatedSegmentDimCursor.decodeBlockIfNeeded()`'s new header cross-check cannot see `n <= 0` (rejected at `:527-530`); `LSMTreeIndexCompacted.searchInCurrentPage()`'s new `rootPurpose == 3` branch implies `!ascendingOrder` (`:514`) so it skips nothing that applied; `TransactionContext.commitFromReplica`'s unguarded cast (`2b59d69ac`) is in a method with no caller anywhere; and `LSMVectorIndex`'s new `persistedGraphUnresolved` latch costs a repeated validation scan on failure, never a lost graph.

robfrank:
Closes #6824
Closes #6839

Two TimeSeries issues in milestone 26.9.1. Three defects, three regression tests, each proven to fail before its fix.

## #6824 - `ts.timeBucket()` truncated toward zero

`SQLFunctionTimeBucket` computed its boundary with plain Java integer division. A bucket function has exactly one contract, `bucket(t) <= t`, and for every negative epoch value the old expression broke it: `ts.timeBucket('1h', -1800000)` (1969-12-31T23:30:00Z) answered `1970-01-01T00:00:00Z`, thirty minutes into its own future. Used as the documented `GROUP BY` key it merged `[-1h, 0)` into the bucket for `[0, 1h)`, so any historical series crossing 1970 aggregated two intervals under one boundary.

`Math.floorDiv`, which is what the engine's own bucket anchor has used since #4595 (`TimeSeriesNegativeTimestampBucketTest`) - this is the same defect one layer up, in the SQL function, that fix did not reach. The `intervalMs <= 0` guard added for #6388 only ever guarded the divisor.

**Test:** `Issue6824TimeBucketPreEpochTest` - the containment contract over nine timestamps, the two documented repros as exact boundaries, the non-collapse of the two buckets either side of the epoch, the typed `Date`/`Instant` overloads, and the unchanged positive path. 4 of 5 arms fail before the fix.

## #6839 item 1 - the engine-unavailable state had no way out, and HA discarded the blob that would repair it

`LocalSchema.readConfiguration()` keeps a TimeSeries type registered when `initEngine()` throws (#6356), justified on the grounds that the failing file is "a `.ts.sealed` most commonly, rebuildable under HA by recompacting the replicated mutable pages". That repair path did not exist: `applySealedBlobs` refused exactly that state (`tsType.getEngine() == null`), logged SEVERE and skipped, and a Raft entry is applied once and never re-shipped - so every dropped blob was a permanent divergence, on the one state HA was named as the cure for. Nothing else ever retried `initEngine()`, maintenance is gated off the type by `isEngineAvailable()`, and every read and write throws through `requireEngine()`.

The fix writes the incoming bytes to the shard's `.ts.sealed` and re-runs `initEngine()` over them. **Two things beyond the issue's minimal sketch were required, and the test pins both:**

1. **Order.** Retrying `initEngine()` first re-opens the same corrupt file and fails for the same reason, leaving no store to install the blob through - "retry, then install" recovers nothing in the reported repro. The blob *is* the repair, so it goes down first; nothing is installed afterwards because the file already is the blob.

2. **`initEngine()` had to become genuinely retryable.** `TimeSeriesShard`'s constructor closed the mutable bucket when the sealed store threw, and `PaginatedComponentFile.close()` sets `open=false`, which its own lazy reopen then explicitly refuses ("was closed on purpose") for the life of the process. The component stays schema-registered on both branches either way, so that close never prevented a leak - it only made the file permanently unreadable. Harmless while a failed `initEngine()` meant the type vanished; since #6356 made the type stay, it was the thing standing in the way of recovery. Measured: with the close restored, the recovery arm fails outright - the engine never comes back at all.

The target file name is derived from the type and shard (`TimeSeriesSealedStore.sealedFileNameFor`), never taken from `blob.fileName()`: a path arriving over the wire has no business selecting a file on this node. The write goes through a temporary and `Files.move(REPLACE_EXISTING)`, the same way `installSealedFileBytes` does it. A failure is logged and reported, never thrown, so one unrepairable type cannot abort an entry carrying blobs for others.

**Test:** `Issue6839TsSealedBlobRecoveryTest` - a real `LocalDatabase` and a real unstarted `ArcadeStateMachine`, no mocks, same harness as `ArcadeStateMachineBootstrapDivergenceTest`. The defect is entirely inside the apply path, so a 3-node IT would add flakiness without adding evidence. Three arms: the repair (which reads AND writes afterwards, because the flag flips either way and only touching the file distinguishes a real recovery from a dead component), a blob for a non-TimeSeries type still refused with nothing written, and a healthy type still installing through its live engine without being re-initialised.

## #6839 item 2 - Prometheus remote-write reported a phantom name collision

PR #6779 converted every TimeSeries call site to the `isEngineAvailable()` / `requireEngine()` pair and missed `PostPrometheusWriteHandler.getOrCreateType`, which asked one question for two things:

```java
if (docType instanceof LocalTimeSeriesType tsType && tsType.getEngine() != null)
```

A type that exists and *is* a TimeSeries type, merely without an engine, failed that guard and fell through to the auto-create branch, which threw `SchemaException("Type 'X' already exists")`. Not cosmetic: that sends an operator hunting a collision that does not exist, while `requireEngine()`'s message names the type and the file whose failure to open put it in this state, which is the whole reason #6356 kept the type registered. No data damage either way - `create()` throws before it mutates the schema.

The two conditions are now split, and the caller uses `requireEngine()`.

**Test:** `Issue6839PrometheusWriteEngineUnavailableIT`. Before the fix it reproduces the reported body verbatim: `{"exception":"com.arcadedb.exception.SchemaException","detail":"Type 'cpu_unavailable' already exists"}`.

## Verification

| Suite | Result |
|---|---|
| `engine` - `com.arcadedb.engine.timeseries.*Test` + `com.arcadedb.function.sql.time.*Test` | 418/418 |
| `engine` - `com.arcadedb.schema.*Test` | 459/459 |
| `ha-raft` - `ArcadeStateMachine*Test`, `Issue6839*`, `Issue6111*`, `Issue5410*` | 129/129 |
| `ha-raft` IT - `RaftTimeSeriesReplication3NodesIT`, `RaftTimeSeriesOversizedSealedIT` (the two that drive sealed-blob replication) | 5/5 |
| `server` - `Issue6839PrometheusWriteEngineUnavailableIT` | 1/1 |
| `server` - `PrometheusApiSpecTest`, `OpenApiSpecGeneratorTest` | 18/18 |

Each new test was run against the unfixed code first and fails there; the `TimeSeriesShard` half was separately verified load-bearing by restoring the close and watching the recovery arm go red.

Two IT classes could not be run on this machine: `PrometheusRemoteWriteReadIT` and `RaftTimeSeriesWriteReadYourWritesIT` address `127.0.0.1:248<serverIndex>` literally, and port 2480 here is held by an unrelated ArcadeDB instance, so their requests reach that server and come back `404 Database 'graph' is not available`. Environmental, not a regression - the same handler returns 204 in the new IT, which binds its own port range. CI will run both.

The new IT overrides `onServerConfiguration` to bind 2490-2499 and builds its URL from `getHttpServer().getPort()`, so it neither collides with a locally installed ArcadeDB nor cares which port of the range it lands on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected time-bucket grouping for timestamps before the Unix epoch.
  * Improved time-series engine recovery and retry behavior while preserving mutable data.
  * Prometheus write errors now clearly identify unavailable time-series engines instead of reporting misleading conflicts.
  * Improved recovery of replicated time-series data after sealed-store corruption.

* **Tests**
  * Added coverage for pre-epoch time buckets, engine initialization retries, sealed-store recovery, and Prometheus writes when storage is unavailable.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesEngine.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesEngine.java
@@ -105,11 +105,16 @@ public class TimeSeriesEngine implements AutoCloseable {
         shards[i] = new TimeSeriesShard(database, typeName, i, columns, compactionBucketIntervalMs, tagDictionary);
     } catch (final Exception e) {
       shardExecutor.shutdownNow();
-      // Close any shards that were successfully created
+      // Release the shards that were successfully created - their sealed stores only, NOT their mutable buckets.
+      // A full close() here made a corrupt .ts.sealed on shard N take the mutable buckets of shards 0..N-1 down
+      // with it, permanently: those buckets are registered with the schema, which owns closing them, and
+      // PaginatedComponentFile.close() sets open=false so its own lazy reopen refuses ever after. That left
+      // initEngine() unretryable for any type with more than one shard, so the whole recovery path added for
+      // issue #6839 would have worked for SHARDS 1 and silently not above it.
       for (final TimeSeriesShard shard : shards) {
         if (shard != null) {
           try {
-            shard.close();
+            shard.closeAfterFailedInit();
           } catch (final IOException ignored) {
           }
         }
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesSealedStore.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesSealedStore.java
@@ -2249,6 +2249,20 @@ public class TimeSeriesSealedStore implements AutoCloseable {
   }
 
   /**
+   * The same name as {@link #getSealedFileName()}, derived from the type and shard alone so a caller that has no
+   * store to ask can still address the file - which is the whole point: the HA apply path needs it precisely when
+   * the store could not be opened and there is no instance to ask (issue #6839). Deriving it locally is also why
+   * that path never uses the file name carried in the replicated blob: a name is a name until it is used to open a
+   * file, and one arriving over the wire has no business selecting a path on this node.
+   * <p>
+   * The {@code <type>_shard_<index>} half must stay in step with the shard name {@code TimeSeriesShard}'s
+   * constructor builds; the suffix is this class's own.
+   */
+  public static String sealedFileNameFor(final String typeName, final int shardIndex) {
+    return typeName + "_shard_" + shardIndex + ".ts.sealed";
+  }
+
+  /**
    * Reads the entire sealed-store file into a byte array. Used on the HA leader to capture the
    * post-compaction (or post-retention/downsampling) sealed bytes so they can be shipped to followers.
    * The header is flushed first so the on-disk image is current.
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
@@ -178,15 +178,17 @@ public class TimeSeriesShard implements AutoCloseable {
       }
     }
 
-    // If sealedStore construction fails, close mutableBucket to avoid a resource leak
-    final TimeSeriesSealedStore tempSealedStore;
-    try {
-      tempSealedStore = new TimeSeriesSealedStore(shardPath, columns);
-    } catch (final IOException e) {
-      try { this.mutableBucket.close(); } catch (final Exception ignored) {}
-      throw e;
-    }
-    this.sealedStore = tempSealedStore;
+    // The mutable bucket is deliberately NOT closed when the sealed store fails to open. By the time control
+    // reaches here it is registered with the schema on BOTH branches above - the component factory registered it
+    // on a cold open, and the creation branch calls registerFile() itself - so the schema owns its lifecycle and
+    // will close it with the database. Closing it here never prevented a leak; it only made the file unusable,
+    // because PaginatedComponentFile.close() sets open=false and its own lazy reopen then refuses ("was closed on
+    // purpose"), for the life of the process. That was harmless while a failed initEngine() meant the type
+    // vanished from the schema, and became the thing standing in the way of recovery once #6356 made the type stay
+    // registered: retrying initEngine() after the sealed file is repaired reuses this very component and produced
+    // a type reporting isEngineAvailable() == true whose every read threw FileNotFoundException (issue #6839).
+    // The failure at initHeaderPage() a few lines above already propagates without closing, for the same reason.
+    this.sealedStore = new TimeSeriesSealedStore(shardPath, columns);
 
     // Crash recovery: if a compaction was interrupted, truncate any partial sealed blocks
     database.begin();
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
@@ -204,9 +204,13 @@ public class TimeSeriesShard implements AutoCloseable {
     } catch (final Exception e) {
       if (database.isTransactionActive())
         database.rollback();
-      // Close both stores to avoid resource leaks before propagating the error
+      // The sealed store is this shard's own file handle and nothing else will close it, so it must be. The
+      // mutable bucket must NOT be, for the reason spelled out above the sealed-store construction: it is
+      // schema-registered, the schema owns closing it, and PaginatedComponentFile.close() is permanent. This
+      // branch is the second way into that trap and it is the one a retry is most likely to meet - both
+      // truncateToBlockCount's temp-file rewrite and the commit below it can throw - so closing here would poison
+      // the bucket on the very attempt meant to bring the type back (issue #6839).
       try { this.sealedStore.close(); } catch (final Exception ignored) {}
-      try { this.mutableBucket.close(); } catch (final Exception ignored) {}
       throw e instanceof IOException ? (IOException) e :
           new IOException("Crash recovery failed for shard " + shardIndex, e);
     }
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesShard.java
@@ -1181,6 +1181,22 @@ public class TimeSeriesShard implements AutoCloseable {
   }
 
   /**
+   * Releases only what this shard owns outright, leaving the mutable bucket open. Used when a LATER shard of the
+   * same type fails to construct and the engine has to unwind the shards it already built (issue #6839).
+   * <p>
+   * {@link #close()} is wrong there for the reason spelled out in the constructor: the mutable bucket is
+   * registered with the schema, which owns closing it, and {@code PaginatedComponentFile.close()} sets
+   * {@code open=false} permanently. So unwinding with the full close made one corrupt {@code .ts.sealed} on shard
+   * N poison the buckets of shards 0..N-1 too, and no later {@code initEngine()} could recover the type however
+   * good the incoming blob was - the whole recovery path #6839 adds would work for {@code SHARDS 1} and silently
+   * not for anything above it. The sealed store is this shard's own file handle and nothing else will close it,
+   * so it still must be.
+   */
+  void closeAfterFailedInit() throws IOException {
+    sealedStore.close();
+  }
+
+  /**
    * Drops this shard: deletes the sealed store file and removes the mutable bucket
    * from the file manager. The mutable bucket file is deleted via the schema's file manager.
    */
--- a/engine/src/main/java/com/arcadedb/function/sql/time/SQLFunctionTimeBucket.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/time/SQLFunctionTimeBucket.java
@@ -65,8 +65,13 @@ public class SQLFunctionTimeBucket extends SQLFunctionConfigurableAbstract {
 
     final long timestampMs = toEpochMs(params[1]);
 
-    // Truncate to bucket boundary
-    final long bucketStart = (timestampMs / intervalMs) * intervalMs;
+    // Floor to the bucket boundary. Math.floorDiv, not '/': Java integer division truncates TOWARD ZERO, so for a
+    // pre-epoch (negative) timestamp the plain division returned a boundary LATER than its own input - '1h' over
+    // -1800000 (1969-12-31T23:30:00Z) answered the epoch itself - and collapsed the last pre-epoch bucket into the
+    // first post-epoch one, silently merging two intervals under one GROUP BY key (issue #6824). The engine's own
+    // bucket anchor was floor-aligned for the same reason in #4595; this is the SQL function that fix did not
+    // reach. intervalMs is positive here (guarded above), so floorDiv differs from '/' only on negative inputs.
+    final long bucketStart = Math.floorDiv(timestampMs, intervalMs) * intervalMs;
 
     return new Date(bucketStart);
   }
--- a/engine/src/main/java/com/arcadedb/schema/LocalTimeSeriesType.java
+++ b/engine/src/main/java/com/arcadedb/schema/LocalTimeSeriesType.java
@@ -76,8 +76,18 @@ public class LocalTimeSeriesType extends LocalDocumentType {
   public synchronized void initEngine() throws IOException {
     if (engine != null)
       return;
-    engine = new TimeSeriesEngine((DatabaseInternal) schema.getDatabase(), name, tsColumns, shardCount > 0 ? shardCount : 1,
-        compactionBucketIntervalMs, mutableFormatVersion);
+    try {
+      engine = new TimeSeriesEngine((DatabaseInternal) schema.getDatabase(), name, tsColumns,
+          shardCount > 0 ? shardCount : 1, compactionBucketIntervalMs, mutableFormatVersion);
+    } catch (final IOException | RuntimeException e) {
+      // Record why THIS attempt failed, so a retry that fails for a new reason does not keep reporting the old
+      // one. Since #6839 this method is retried outside schema load - the HA sealed-blob repair path drives it -
+      // and markEngineUnavailable() is deliberately restricted to the single-threaded load path, so without this
+      // a second failure left CHECK DATABASE and requireEngine()'s message describing the FIRST one. Set here,
+      // inside the same synchronized block that clears it on success, so the two can never interleave.
+      engineUnavailableReason = e.getMessage() != null ? e.getMessage() : e.toString();
+      throw e;
+    }
     // A retry that succeeds after a prior markEngineUnavailable() must not leave the stale reason behind: nothing
     // else clears it, and getEngineUnavailableReason() otherwise keeps reporting why the engine failed even after
     // isEngineAvailable() has correctly flipped to true.
--- a/engine/src/test/java/com/arcadedb/engine/timeseries/Issue6839InitEngineRetryableTest.java
+++ b/engine/src/test/java/com/arcadedb/engine/timeseries/Issue6839InitEngineRetryableTest.java
@@ -1,0 +1,207 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.engine.timeseries;
+
+import com.arcadedb.TestHelper;
+import com.arcadedb.query.sql.executor.ResultSet;
+import com.arcadedb.schema.LocalTimeSeriesType;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/**
+ * Issue #6839: {@code initEngine()} has to be retryable, because since #6356 it is the ONLY way a type registered
+ * without a storage engine ever gets one - the HA sealed-blob repair path drives exactly this call.
+ * <p>
+ * What stood in the way was a cleanup that looked like hygiene. Three places closed the shard's mutable bucket
+ * when something else failed to open: {@code TimeSeriesShard}'s sealed-store construction, its crash-recovery
+ * block, and {@code TimeSeriesEngine}'s unwind of the shards it had already built. The bucket is registered with
+ * the schema on every branch that reaches them, so the schema owns closing it and none of those closes ever
+ * prevented a leak - but {@code PaginatedComponentFile.close()} sets {@code open=false} and its own lazy reopen
+ * then refuses ("was closed on purpose") for the life of the process. So the failed attempt destroyed the thing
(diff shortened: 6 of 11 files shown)


robfrank:
## Review cycle 1

### `TimeSeriesEngine` constructor - confirmed and fixed

You were right, and it is worse than a missed cleanup: it is the same defect one level up, so the recovery path would have worked for `SHARDS 1` and silently not for anything above it.

Reproduced before fixing. A `SHARDS 2` arm with the corrupt `.ts.sealed` on shard 1 dies on the post-repair append:

```
java.io.IOException: Failed to append timeseries samples
Caused by: java.lang.NullPointerException: Cannot invoke "java.nio.channels.FileChannel.size()"
  because "this.channel" is null
  at com.arcadedb.engine.PaginatedComponentFile.getTotalPages(PaginatedComponentFile.java:215)
```

Shard 0's bucket, closed by the unwind when shard 1 threw. (A rougher failure than the "closed on purpose" refusal, incidentally - `getTotalPages()` dereferences the null channel before anything reaches the reopen guard.)

Fixed by taking your first option rather than the follow-up: `TimeSeriesEngine`'s unwind now calls a new `TimeSeriesShard.closeAfterFailedInit()`, which closes the sealed store this shard owns outright and leaves the mutable bucket to the schema that registered it. `close()` is unchanged, so the ordinary shutdown path still closes both. New arm `aSealedBlobRepairsATypeWithMoreThanOneShard` asserts the repair, the full sample count, and a write to both shards afterwards.

### Path derivation

Tightened rather than documented: the target name now comes from `tsType.getName()` instead of `blob.typeName()`. They are equal here by construction, but taking it from the resolved schema type means the name selecting a file on this node provably came from the local schema, rather than doing so only as long as the caller keeps that lookup ahead of it.

### Directory fsync

Agreed it is inherited, and leaving it. Adding a directory fsync to `repairEngineWithSealedBlob` alone would make it diverge from `TimeSeriesSealedStore.installSealedFileBytes`, which is the pattern it deliberately mirrors and which every other sealed-store rewrite in that class also follows. It is worth a follow-up across all of them; it is not worth one path behaving differently from the rest.

### Re-verified after the change

| Suite | Result |
|---|---|
| `engine` - `com.arcadedb.engine.timeseries.*Test` | 393/393 |
| `ha-raft` - `ArcadeStateMachine*Test`, `Issue6839*Test` | 100/100 |
| `ha-raft` IT - `RaftTimeSeriesReplication3NodesIT`, `RaftTimeSeriesOversizedSealedIT` | 5/5 |
| `server` IT - `Issue6839PrometheusWriteEngineUnavailableIT` | 1/1 |

robfrank:
## Review cycle 2

### The crash-recovery block - fixed, with one correction to the stated trigger

The finding is right and the fix is applied: that catch now closes only the sealed store and leaves the mutable bucket to the schema, same as the other two sites. It is also, as you say, the door a repair is *most* likely to meet, for exactly the reason you give - a `.ts.sealed` is most often damaged by a crash during compaction, which is precisely when `isCompactionInProgress()` is true on reopen.

The specific trigger you named does not itself fire, though. `truncateToBlockCount` opens with:

```java
if (targetBlockCount >= blockDirectory.size())
  return; // nothing to truncate
```

So an installed blob whose block count disagrees with this follower's stale watermark truncates (or returns early); it does not throw. What can throw in that block is the temp-file rewrite inside `truncateToBlockCount` and the `database.commit()` under it. The defect is identical either way, so this changes the reproduction, not the conclusion.

That mattered for building the test, which is why it is worth spelling out.

### Regression test

`Issue6839InitEngineRetryableTest`, in the engine module (the invariant is `initEngine()`'s, not HA's). The injection is a **broken symlink** at the `.ts.sealed.tmp` path: `File.exists()` follows the link and answers false, so `TimeSeriesSealedStore`'s constructor skips its own stale-tmp cleanup and opens normally, and then the `new RandomAccessFile(tempPath, "rw")` inside the rewrite resolves the link to a directory that is not there and throws - landing in the crash-recovery catch and nowhere else.

Worth recording why the obvious injection does not work, since it was my first attempt: a plain directory at that path is **deleted by the sealed store's constructor** before the shard ever reaches crash recovery (measured - the probe showed the blocker gone, the flag cleared and recovery completed), and a non-empty one makes the constructor throw, which is the *other* door and already covered. `assumeTrue` skips the class where symlinks are unavailable; the production fix is platform-independent, only the injection is not.

The post-retry assertion is a **delta**, not a total: the recovery this test provokes legitimately truncates the sealed store to the watermark, and how many samples that discards is beside the point. What has to hold is that the bucket still reads and still takes writes.

Armed, verified by restoring the close:

```
java.lang.NullPointerException: Cannot invoke "java.nio.channels.FileChannel.size()"
  because "this.channel" is null
```

### Temp file not cleaned up when `Files.move` throws

Agreed, and leaving it for the same reason as the directory fsync: `repairEngineWithSealedBlob` deliberately mirrors `TimeSeriesSealedStore.installSealedFileBytes`, which has the same gap, and so does every other `.tmp`/`.incoming` rewrite in that class. There is also a real backstop - `TimeSeriesSealedStore`'s constructor deletes a stale `.incoming` on the next open and fails loudly if it cannot - so the residue is bounded rather than silent. Worth one follow-up across all of them; not worth this one path behaving differently from its siblings.

### Re-verified after the change

| Suite | Result |
|---|---|
| `engine` - `com.arcadedb.engine.timeseries.*Test` + `com.arcadedb.function.sql.time.*Test` | 421/421 |
| `ha-raft` - `ArcadeStateMachine*Test`, `Issue6839*Test` | 100/100 |
| `ha-raft` IT - `RaftTimeSeriesReplication3NodesIT`, `RaftTimeSeriesOversizedSealedIT` | 5/5 |
| `server` IT - `Issue6839PrometheusWriteEngineUnavailableIT` | 1/1 |

All three cleanup sites now agree: the sealed store is the shard's own handle and gets closed; the mutable bucket belongs to the schema and does not.

robfrank:
## Review cycle 3

All three items addressed; the third turned out to need no change, and I checked rather than assumed.

### 1. Stale `engineUnavailableReason` after a failed repair - fixed

Fixed at the source rather than commented. `initEngine()` now records the cause of the attempt that just failed, inside the same `synchronized` block that clears it on success, so the two can never interleave. That keeps `markEngineUnavailable()`'s restriction to the single-threaded schema-load path intact, which is what made the reason go stale in the first place once this PR started retrying `initEngine()` from elsewhere.

### 2. No test for the failed-repair path - added

`aBlobThatCannotBeOpenedLeavesTheTypeUnavailableWithoutThrowing`: a blob the sealed store cannot open must leave the type exactly as unavailable as it was, and return rather than throw, because the same entry may carry repairable blobs for other types. It also pins item 1, and the refresh is observably real rather than a formality:

```
before = Invalid sealed store magic in '.../Cpu_shard_0.ts.sealed': 55534958   <- the flipped byte
after  = Invalid sealed store magic in '.../Cpu_shard_0.ts.sealed': 0          <- the bad blob
```

Armed: with the `initEngine()` change reverted, the assertion fails.

### 3. `TimeSeriesTagDictionary` as a fourth door - checked, there isn't one

There is exactly one call site for `tagDictionary.close()` in the whole engine:

```
engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesEngine.java:879
```

which is inside `TimeSeriesEngine.close()`, the normal shutdown path. No failure path closes it - not the constructor's unwind, not `TimeSeriesShard`'s two catches, not `closeAfterFailedInit()`. And a retry does not need it reopened: `openOrCreate` resolves it through `schema.getFileByName(name)` and calls `ensureLoaded()` on the instance already registered. So the dictionary is reused rather than reconstructed, and cannot be poisoned the way the mutable buckets were.

Right question to ask, given the methodology; the answer happens to be that the door does not exist.

### Re-verified

| Suite | Result |
|---|---|
| `engine` - `com.arcadedb.engine.timeseries.*Test` + `com.arcadedb.function.sql.time.*Test` + `com.arcadedb.schema.*Test` | 880/880 |
| `ha-raft` - `ArcadeStateMachine*Test`, `Issue6839*Test` | 101/101 |
| `ha-raft` IT - `RaftTimeSeries*IT` (all four classes) | 9/9 |
| `server` IT - `Issue6839PrometheusWriteEngineUnavailableIT` | 1/1 |

`RaftTimeSeriesWriteReadYourWritesIT` passes now, which settles the caveat in the PR description: its earlier failures here were the port-2480 collision and nothing else.

robfrank:
## Review cycle 4 (final)

No blocking items this round. Both observations noted; one turned into a test.

### Naming coupling - now enforced by a test rather than a comment

Right to flag it as the one place with no compile-time enforcement. Converted into an assertion in `aHealthyTypeStillInstallsTheBlobThroughItsLiveEngine`:

```java
assertThat(TimeSeriesSealedStore.sealedFileNameFor(TYPE_NAME, 0))
    .isEqualTo(healthy.getEngine().getShard(0).getSealedStore().getSealedFileName());
```

The right-hand side is derived from the path the shard actually opened, so a rename in `TimeSeriesShard`'s constructor now fails this directly and by name, rather than being caught indirectly (and only probably) through `getEngineUnavailableReason()` containing a filename.

### Partial `.incoming` on a mid-write failure

Agreed, and the point about it being self-healing is a better version of the answer I gave in cycle 2: the `.incoming` suffix is not merely consistent with `installSealedFileBytes`, it is the exact name `TimeSeriesSealedStore`'s constructor already deletes as stale on the next open (#4382), failing loudly if it cannot. So the residue is bounded by the next open rather than accumulating. Leaving it, for the same reason as before - a single follow-up across every `.tmp`/`.incoming` rewrite in that class is worth doing; one path behaving differently from its siblings is not.

### Build

Run locally, in the worktree, after every change in this cycle:

| Suite | Result |
|---|---|
| `engine` - `com.arcadedb.engine.timeseries.*Test` + `com.arcadedb.function.sql.time.*Test` + `com.arcadedb.schema.*Test` | 880/880 |
| `ha-raft` - `ArcadeStateMachine*Test`, `Issue6839*Test` | 101/101 |
| `ha-raft` IT - `RaftTimeSeries*IT` (all four classes) | 9/9 |
| `server` IT - `Issue6839PrometheusWriteEngineUnavailableIT` | 1/1 |
| `server` - `PrometheusApiSpecTest`, `OpenApiSpecGeneratorTest` | 18/18 |

Every new test was also run against the unfixed code and fails there, and each of the four production changes was separately reverted to confirm the arm covering it goes red: the `TimeSeriesShard` sealed-store close, the `TimeSeriesEngine` multi-shard unwind, the crash-recovery close, and `initEngine()`'s reason refresh.

`Java CI - test` is running on this push.

robfrank:
## CI: `Java CI - test` is red, and none of it is this PR

Diffed the failing set against a main baseline and against this branch's own earlier run, rather than reading red as regression.

| Test | main `33161892670` | main `33160941219` | PR run A `33172368671` | PR run B `33172766896` |
|---|---|---|---|---|
| `Issue6655StaleGraphPrefixReuseTest.reopenWithMoreVectors...` | FAIL | FAIL | FAIL | FAIL |
| `Issue6772WriteBeforeSearchPrefixReuseTest.aSessionThatInsertsBefore...` | FAIL | FAIL | FAIL | FAIL |
| `Issue6772WriteBeforeSearchPrefixReuseTest.concurrentSearchersOn...` | FAIL | FAIL | FAIL | FAIL |
| `GraphAnalyticalViewTest.issue6641_corruptedPersistedCsr...` | FAIL | FAIL | pass | FAIL |
| `Issue6762WebSocketEventBusTest.shutdownStillJoinsAfterASeparateStopSignal` | pass | pass | pass | FAIL |
| `Issue6075SnapshotBackupIT.backupUnderConcurrentWriters...` | pass | pass | pass | FAIL |

**Runs A and B are the same production code.** The only commit between them is `cce5fd61cf`, which adds one assertion to an existing test method and touches no `src/main`. Three tests differ across those two runs, which is intermittency by construction, not a change in behaviour.

The three vector prefix-reuse failures are on main and on every run here - all `ConditionTimeout` after 1 minute waiting on the async rebuild, the known `LSMVectorIndex.REBUILD_SEMAPHORE` convoy that the `vector` lane exists for. `GraphAnalyticalViewTest` is on both main baselines.

That leaves two that appear nowhere else. Both are scheduling races, in code this PR does not touch:

- **`Issue6762WebSocketEventBusTest`** - `shutdown()` racing `run()`'s unwind in the WebSocket event bus. Ran 3x locally on this branch: 9/9 each time.
- **`Issue6075SnapshotBackupIT`** - point-in-time backup under concurrent writers, asserting the restored count is strictly inside `[20001, 29360]`; it got exactly `20000`, the pre-writer baseline. That is the snapshot being taken before any concurrent writer committed - a loaded-runner scheduling outcome, and the assertion's lower bound has no headroom for it. Ran 3x locally: 4/4 each time.

Neither is reachable from anything in this diff: the changes are `SQLFunctionTimeBucket`, `TimeSeriesShard`/`TimeSeriesEngine`/`TimeSeriesSealedStore`, `LocalTimeSeriesType.initEngine()`, `ArcadeStateMachine.applySealedBlobs`, and `PostPrometheusWriteHandler`. Nothing in the WebSocket event bus or the backup/snapshot path.

`main`'s own `integration-tests` job cannot serve as a baseline for `Issue6075SnapshotBackupIT` specifically - it dies mid-`HTTPDocumentIT` before reaching it, on every recent run - so that one rests on the A-vs-B comparison and the local runs rather than on a main baseline. Flagging that rather than overstating it.

Everything this PR does touch is green in CI: `slow-unit-tests`, `opencypher-tck-tests`, `build-and-package`, `lint`, and all five e2e suites.

PR #6879, opened 2026-08-28
fix(ci): six test failures on main (run 33161892670) - three racy/under-budgeted tests, one missing dependency

robfrank:
Triage of [CI run 33161892670](https://github.com/ArcadeData/arcadedb/actions/runs/33161892670) on `main`, red across `unit-tests`, `vector-unit-tests` and `integration-tests`. Six failures, four independent causes. One is a real dependency gap; the rest are tests that were racing or under-budgeted rather than product defects.

| Job | Test | Cause | Fix |
|---|---|---|---|
| unit-tests | `GraphAnalyticalViewTest.issue6641_corruptedPersistedCsr…` | test races the API it tests | saturate `BUILD_PERMITS` around the assertion |
| vector | `Issue6655StaleGraphPrefixReuseTest` (1) | 60s wait, rebuild still in flight | 4 min + diagnostic message |
| vector | `Issue6772WriteBeforeSearchPrefixReuseTest` (2) | same | same |
| integration | `PostgresWJdbcIT.backupDatabase` | **missing dependency**, exposed by #6876 | declare `arcadedb-integration` at test scope |
| integration | `RemoteConsoleIT.userMgmt` | assertion obsoleted by #6830 | assert the new behaviour, move the syntax-error case |

## 1. The two `#6641` GAV tests raced `isReady()`

`GraphAnalyticalView.isReady()` dispatches a pending deferred restore and *then* reads `status`. That ordering is deliberate - its javadoc explains that reading the resolved state beats short-circuiting on `pendingDiskRestore` alone, so the rare caller whose dispatch settles in between gets the true answer rather than a needlessly pessimistic one.

On this fixture the view holds 2 vertices and 1 edge, so the whole background chain (the restore rejects the truncated file, `buildAsync()` rebuilds, `status` goes `READY`) fits comfortably inside a scheduling stall between those two steps. The assertion that the view must *not* be ready then loses the race, which is precisely the reported failure:

```
issue #6641: a corrupted persisted CSR must not be advertised as ready before it is verified
Expecting value to be false but was true
```

**Reproduced deterministically** by inserting a 1.5s delay between the dispatch and the read: the test goes red with the identical message, and stays green with the permits held.

The fix uses the test hook that already exists for this shape (`acquireAllBuildPermitsForTest()`, used by the in-flight-race test in the same class). The dispatched task acquires a build permit *before* it reads anything, so saturating the semaphore first parks it and pins the observation to the instant the query-time gate is actually being asked about. The sibling commit-variant test carries the identical race and is fixed alongside it, rather than waiting for its own red run.

Note the product behaviour is not changed and does not need to be: an `isReady()` that returns `true` there is *correct* - the view genuinely is ready by then.

## 2. Three vector prefix-reuse waits were sized for a developer machine

All three failures are 60s `Awaitility` waits on the async rebuild of 20,200 x 128 vectors. The third one is the diagnostic:

```
'the async rebuild kicked off by the stale-prefix reuse settles before the drop'
didn't complete within 1 minutes ... expected: 0L but was: 1L
```

`asyncRebuildInProgress` was still `1` - the rebuild was **running, or still parked on the sole JVM-wide `REBUILD_SEMAPHORE` permit** (whose own wait is 10 minutes by default). It had not been skipped or deferred for memory, both of which log a WARNING that does not appear in the run.

This is the hazard `CLAUDE.md`'s `@Tag("vector")` note already documents: the lane convoys on one permit, so the waits have to be sized for the worst case. 60s was not - the whole `Issue6655` class runs in ~10s locally and took 164.6s on the 4-vCPU runner.

Raised to 4 minutes. These are **convergence waits, not latency bounds** - they assert that the rebuild happens, not that it is fast - so a generous value cannot turn a passing run red. Each wait now also reports the other stat alongside it, so the next timeout says which arm it was:

> `graphRebuildCount`, with `asyncRebuildInProgress=1`: a 1 there means the rebuild is still running or still parked on the JVM-wide REBUILD_SEMAPHORE and this wait is simply too short; a 0 means it was never started or was skipped, which is a real defect.

## 3. `PostgresWJdbcIT.backupDatabase` - the one real defect

`BACKUP DATABASE` resolves `com.arcadedb.integration.backup.Backup` reflectively, and **nothing on postgresw's test classpath declared it**. `arcadedb-server` has `arcadedb-integration` at `test` scope, which is not transitive.

It passed until now only by accident: `arcadedb-gremlin` has `arcadedb-integration` at compile scope, so its uber-jar carried a copy, and postgresw takes that jar with a wildcard exclusion - making the shaded jar the *only* supplier of those classes. PR #6876 added `<exclude>com.arcadedb:arcadedb-integration</exclude>` to the shade `artifactSet`, correctly, and that turned the accident into:

```
ERROR: Error on executing query: Error on backing up database, backup libs not found in classpath
```

Declared explicitly at `test` scope, with a comment recording why. postgresw is the only module that was living off that leak - checked every module that consumes `arcadedb-gremlin:shaded` (`graphql`, `gremlin-it`, `gremlin-consumer-it`, `package`) for uses of `com.arcadedb.integration` or backup/export/import statements in tests.

## 4. `RemoteConsoleIT.userMgmt`'s syntax-error case stopped being one

#6830 (commit 2dbe5ee861, merged yesterday) removed the console's rejection of spaces in passwords, because it made accounts the console could not create. Only the exact ` GRANT CONNECT TO ` phrase delimits a password, so the deliberately mistyped

```
create user albert identified by einstein grand connect on db1
```

is now an ordinary - if surprising - password rather than something the console can tell apart from one, and the command succeeds. That commit did not update the IT that depended on the old rejection.

The new behaviour is **asserted explicitly rather than deleted**, so a regression in the other direction is still caught, and a genuine syntax error (`create user albert`, no `IDENTIFIED BY` at all) takes over the syntax-error case.

## Verification

All run locally in this worktree:

| Suite | Result |
|---|---|
| `mvn -pl engine test -Dgroups=vector` | **72/72** |
| `GraphAnalyticalViewTest` (full class) | **159/159** |
| `mvn -pl postgresw verify -Dit.test=PostgresWJdbcIT` | **45/45** (+426 unit) |
| `mvn -pl console verify -Dit.test=RemoteConsoleIT` | **16/16** |

Plus the arming check described in §1: with the probe delay and *without* the permit saturation the GAV test fails with the exact CI message; with the saturation it passes.

Test-only changes except `postgresw/pom.xml`, which adds a first-party module already in the reactor (no new third-party dependency, no ATTRIBUTIONS/NOTICE impact).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of user command syntax, including passwords containing spaces and invalid authentication clauses.
  * Improved reliability when restoring analytical graph data and selecting available providers.
  * Extended asynchronous graph rebuild wait periods to better accommodate longer-running operations.

* **Tests**
  * Added clearer diagnostics for rebuild timeouts and deferred data restoration scenarios.
  * Strengthened coverage for console commands, graph analytics, vector indexing, and PostgreSQL connectivity.

### DIFF
--- a/console/src/test/java/com/arcadedb/console/RemoteConsoleIT.java
+++ b/console/src/test/java/com/arcadedb/console/RemoteConsoleIT.java
@@ -170,8 +170,17 @@ public class RemoteConsoleIT extends BaseGraphServerTest {
     assertThat(console.parse("create user albert identified by einstein")).isTrue();
     assertThat(console.parse("drop user albert")).isTrue();
 
-    // TEST SYNTAX ERROR
-    assertThatThrownBy(() -> assertThat(console.parse("create user albert identified by einstein grand connect on db1")).isTrue()).isInstanceOf(Exception.class);
+    // TEST SYNTAX ERROR: a `create user` with no IDENTIFIED BY at all has nowhere to take a password from
+    // and is rejected outright.
+    assertThatThrownBy(() -> assertThat(console.parse("create user albert")).isTrue()).isInstanceOf(Exception.class);
+
+    // A MISTYPED GRANT CLAUSE IS *NOT* A SYNTAX ERROR ANY MORE (ISSUE #6830). ONLY THE EXACT ` GRANT CONNECT TO `
+    // PHRASE DELIMITS THE PASSWORD, AND SINCE #6830 A PASSWORD MAY CONTAIN SPACES, SO `grand connect on db1`
+    // IS AN ORDINARY - IF SURPRISING - PASSWORD RATHER THAN SOMETHING THE CONSOLE CAN TELL APART FROM ONE.
+    // ASSERTED EXPLICITLY, NOT DELETED: THIS USED TO THROW ONLY BECAUSE THE CONSOLE REJECTED SPACES IN
+    // PASSWORDS, AND #6830 REMOVED THAT CHECK BECAUSE IT MADE ACCOUNTS THE CONSOLE COULD NOT CREATE.
+    assertThat(console.parse("create user albert identified by einstein grand connect on db1")).isTrue();
+    assertThat(console.parse("drop user albert")).isTrue();
 
     assertThat(console.parse("create user albert identified by einstein grant connect to db1")).isTrue();
     assertThat(console.parse("create user jeff identified by amazonaws grant connect to db1:readonly")).isTrue();
--- a/console/src/test/java/com/arcadedb/console/RemoteConsoleIT.java
+++ b/console/src/test/java/com/arcadedb/console/RemoteConsoleIT.java
@@ -170,15 +170,16 @@ public class RemoteConsoleIT extends BaseGraphServerTest {
     assertThat(console.parse("create user albert identified by einstein")).isTrue();
     assertThat(console.parse("drop user albert")).isTrue();
 
-    // TEST SYNTAX ERROR: a `create user` with no IDENTIFIED BY at all has nowhere to take a password from
-    // and is rejected outright.
+    // Syntax error: a `create user` with no IDENTIFIED BY at all has nowhere to take a password from and
+    // is rejected outright.
     assertThatThrownBy(() -> assertThat(console.parse("create user albert")).isTrue()).isInstanceOf(Exception.class);
 
-    // A MISTYPED GRANT CLAUSE IS *NOT* A SYNTAX ERROR ANY MORE (ISSUE #6830). ONLY THE EXACT ` GRANT CONNECT TO `
-    // PHRASE DELIMITS THE PASSWORD, AND SINCE #6830 A PASSWORD MAY CONTAIN SPACES, SO `grand connect on db1`
-    // IS AN ORDINARY - IF SURPRISING - PASSWORD RATHER THAN SOMETHING THE CONSOLE CAN TELL APART FROM ONE.
-    // ASSERTED EXPLICITLY, NOT DELETED: THIS USED TO THROW ONLY BECAUSE THE CONSOLE REJECTED SPACES IN
-    // PASSWORDS, AND #6830 REMOVED THAT CHECK BECAUSE IT MADE ACCOUNTS THE CONSOLE COULD NOT CREATE.
+    // A mistyped grant clause is not a syntax error any more (issue #6830). Only the exact
+    // ` GRANT CONNECT TO ` phrase delimits the password, and since #6830 a password may contain spaces, so
+    // `grand connect on db1` is an ordinary - if surprising - password rather than something the console can
+    // tell apart from one. Asserted explicitly rather than deleted: this used to throw only because the
+    // console rejected spaces in passwords, and #6830 removed that check because it made accounts the
+    // console could not create.
     assertThat(console.parse("create user albert identified by einstein grand connect on db1")).isTrue();
     assertThat(console.parse("drop user albert")).isTrue();
 
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -8689,6 +8689,11 @@ public class LSMVectorIndex implements Index, IndexInternal {
    * JVM-wide, so a leaked acquisition starves every later vector rebuild in the same JVM. Blocks until the permits
    * are free, which is deliberate - an earlier test's rebuild still holding one is exactly the convoy the caller
    * needs to be out of before it can observe anything stable.
+   * <p>
+   * Assumes the sequential, single-JVM run this module's tests get today ({@code forkCount=1}, no parallel
+   * classes), same as {@code GraphAnalyticalView.acquireAllBuildPermitsForTest()}. Two callers under a parallel
+   * Surefire would serialize on the semaphore rather than deadlock, but the second would then be observing a
+   * window the first had already let a rebuild run in, so the guarantee this hook exists to give would be gone.
    */
   static void acquireAllRebuildPermitsForTest() {
     REBUILD_SEMAPHORE.acquireUninterruptibly(MAX_CONCURRENT_REBUILDS);
--- a/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
+++ b/engine/src/main/java/com/arcadedb/index/vector/LSMVectorIndex.java
@@ -144,8 +144,9 @@ public class LSMVectorIndex implements Index, IndexInternal {
   // JVM-wide semaphore limiting the number of concurrent async graph rebuilds across all indexes
   // and databases.  Multiple concurrent rebuilds are extremely memory-intensive and can cause OOM
   // kills (issue #3868).  The permit count is read once at class-load time from the configuration.
-  private static final Semaphore REBUILD_SEMAPHORE = new Semaphore(
-      GlobalConfiguration.VECTOR_INDEX_MAX_CONCURRENT_REBUILDS.getValueAsInteger());
+  private static final int       MAX_CONCURRENT_REBUILDS = GlobalConfiguration.VECTOR_INDEX_MAX_CONCURRENT_REBUILDS
+      .getValueAsInteger();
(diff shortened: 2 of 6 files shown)


robfrank:
Both actionable nits addressed; the third needed no change.

**1. ALL CAPS comments** - fixed in 6994e66. The file's prose comments (e.g. the `#5026` block at L140) are sentence case, and ALL CAPS is reserved for its short single-line markers (`// EXPECTED IF ALREADY EXISTENT`, `// EXECUTE BEGIN-END TESTS`), so the multi-line block was the odd one out. Text unchanged; `RemoteConsoleIT` re-run 16/16.

**2. Wall-clock budget for the `vector` lane** - checked, there is headroom, and the arithmetic only matters on the failure path:

- `vector-unit-tests` is `timeout-minutes: 45` (`.github/workflows/mvn-test.yml:391`), and neither class carries a `@Timeout`, so the Awaitility bound is the only ceiling.
- Three waits widened by 3 min each = **9 min of additional worst case**. The red run 33161892670 took 23 min *with all three firing at 60s*, so the same run under this PR would be 23 - 3 + 12 = **~32 min against a 45-min job timeout**.
- That 9 min is only ever paid when a wait times out, i.e. when the job is failing anyway. On a green run these waits return as soon as `graphRebuildCount` flips - locally the whole `vector` lane is 3m35s for 72 tests either way.

The distinction worth keeping: a job timeout kills the lane with no test report, whereas an Awaitility timeout fails one test with the diagnostic message this PR adds. 13 min of headroom keeps the failure on the informative side of that line. If the convoy ever does push the lane past 45 min, the fix is the one the workflow comment at `mvn-test.yml:386` already names - making the permit count configurable, or scoping it per database - not shrinking the wait back below what the rebuild needs.

**3.** No change needed - noted.

robfrank:
### Update: this branch's own CI (33170604968) found a second, latent race - fixed in 5ebaabe

Two of the three vector failures are now green (`Issue6655` and `Issue6772#aSessionThatInsertsBeforeItSearches…`, the latter at 197.9s - it needed the widened wait). `Issue6772#concurrentSearchersOnAWroteFirstReopenStillPublishExactlyOneReuse` failed **earlier than before**, at an assertion the 60s timeout used to shield:

```
and the gap must still be buffered exactly once under contention:
200 from the crashed session plus the 1 this session wrote
expected: 201L but was: 0L
```

**Cause.** `deltaVectorsCount`, `mutationsSinceRebuild`, `graphState` and `graphRebuildCount` are live gauges that the async rebuild the reuse dispatches **resets**. Nothing stopped that rebuild from finishing before the assertions read them, so which of the two states they observed was decided by how fast the JVM handed out the sole `REBUILD_SEMAPHORE` permit.

The block had only ever passed because the *preceding* test's rebuild - the one that used to time out at 60s while still running - was starving it. Fixing the wait removed the starvation, and with it the accidental synchronization. So this is a pre-existing race that the first commit un-hid rather than one it introduced, and leaving it would have handed back a coin flip.

**Fix.** `LSMVectorIndex` gains the hook `GraphAnalyticalView` already has - `acquireAllRebuildPermitsForTest()` / `releaseAllRebuildPermitsForTest()` - so a dispatched rebuild parks before it touches anything. This is safe for the concurrent-searcher test specifically because **both** `REBUILD_SEMAPHORE` acquire sites are `tryAcquire` on background threads (`startAsyncGraphRebuild()`'s own thread at L3778, the inactivity timer at L8870) and **no search path takes a permit**, so holding them all cannot stall a query - the concurrency under test is unchanged. All three affected methods saturate for their observation window and release in a `finally`, before the convergence wait, which is precisely for the rebuild the permits park.

**Proven armed, not assumed.** With a 15s delay inserted before the assertions and the permits *not* held, the test fails with the identical `expected: 201L but was: 0L`; with the permits held and the same delay, it passes. (Same disproof method as the `#6641` fix in the first commit.)

Vector lane re-run locally: **72/72**.

robfrank:
**1. JVM-global permit hooks under a parallel Surefire** - noted in the javadoc (6fe65bb), matching what `GraphAnalyticalView.acquireAllBuildPermitsForTest()` already says. The sharper point than deadlock-vs-serialize is that serializing is not a safe degradation here: the second caller would be observing a window the first had already let a rebuild run in, so the guarantee the hook exists to give would simply be gone. Javadoc only, no behaviour change.

**2, 3.** No change needed.

**On the `-am` suggestion** - the `-am` form re-runs the full upstream suite for every leaf, so I ran the affected suites directly instead, after `mvn -o install -DskipTests` so each leaf resolves this worktree's artifacts:

| Suite | Result |
|---|---|
| `mvn -pl engine test -DexcludedGroups=benchmark,slow,vector` | **13768/13768**, 23 skipped |
| `mvn -pl engine test -Dgroups=vector` | **72/72** |
| `mvn -pl postgresw verify -Dit.test=PostgresWJdbcIT` | **45/45** (+426 unit) |
| `mvn -pl console verify -Dit.test=RemoteConsoleIT` | **16/16** |

One thing worth recording because it cost time and would mislead the next person: the first engine-lane run reported **10 errors**, all Profiler-related, all `Unresolved compilation problems: The type sun.management.HotspotRuntimeMBean is not accessible`. That is not a regression and not a JDK mismatch - it was stale `target/test-classes` left by an IDE language server compiling into the worktree, so Surefire was loading classes ECJ had already compiled *with errors*. `ProfilerTest` passed at base in a separate checkout and passed in this worktree immediately after `mvn -pl engine clean`; the full lane from clean is the 13768/13768 above. If a run ever reports `Unresolved compilation problems` for a class the diff does not touch, clean first before believing it.

robfrank:
### CI result: all three lanes this PR targeted are green

Run [33176234087](https://github.com/ArcadeData/arcadedb/actions/runs/33176234087) on `6fe65bb`:

| Lane | Before (33161892670) | Now |
|---|---|---|
| `unit-tests` | ❌ `GraphAnalyticalViewTest.issue6641_corruptedPersistedCsr…` | ✅ |
| `vector-unit-tests` | ❌ 3 prefix-reuse failures | ✅ |
| `integration-tests` | ❌ `PostgresWJdbcIT.backupDatabase`, `RemoteConsoleIT.userMgmt` | ✅ |

All other lanes green except two, neither of which this PR touches:

**`slow-unit-tests`** - `MaterializedViewRefreshAtomicityTest.refreshingAnIndexedViewRepeatedlyDoesNotGrowTheIndex` (`the index is on disk to begin with: expecting 0L to be greater than 0L`). The **immediately preceding run on this branch, 33175229417, passed this lane on identical code** - `6fe65bb` adds nothing to `5ebaabe` but a javadoc paragraph - so the diff cannot be the cause. The lane also failed on `main` at 33151135675, on a different test (`Issue6657CloseTimeRebuildPendingStatTest`).

**`ha-integration-tests`** - **this was already red on the target run 33161892670** and I did not include it in the triage; it was cut off the truncated job listing I read, which is my miss, so recording it explicitly here rather than leaving it implied. It is a different test every run, and the failures are cluster/environment shaped rather than assertion shaped:

| Run | Failing tests | Symptom |
|---|---|---|
| 33161892670 (main) | `RaftHAComprehensiveIT` x3 | `Unable to listen to a HTTP port in the configured port range 42482 - 42482` |
| 33175229417 (this branch) | - | ✅ passed |
| 33176234087 (this branch) | `RaftDropDatabase3NodesIT`, `Issue6091GrpcRoutingTableIT`, `RaftPriorityRejoinIT` | routing table null, create-database command failure, cluster-committed-but-local-apply-failed |

Non-overlapping test sets across runs, a green run on this branch in between, and a port-binding error as the `main`-run symptom - that is lane flakiness, already tracked (#5668, plus the unfixed HA ITs carried as follow-ups from #6770), not something this PR's four causes reach. Fixing it is real work on the Raft cluster fixtures and belongs in its own PR against those issues; folding it in here would mix an unrelated, unreproduced investigation into a triage that is otherwise fully verified.

So: **6 of 6 failures analysed and fixed, three lanes restored to green; the fourth red lane on that run is a pre-existing flake left where it is, and named rather than quietly dropped.**

PR #6881, opened 2026-08-28
fix(#6875): coerce a server setting to its declared type, and make the two read paths one parse

### LINKED ISSUE #6875: set_server_setting stores an unparseable value for a typed setting, and the two read paths disagree about what parses
Follow-up to #6837, split out from PR #6874 which deliberately left it alone.

### What #6874 fixed, and what it did not

#6874 stopped `set_server_setting` from accepting a *missing* `value` (it was read with an empty-string default and stored as `""`), and additionally rejects an explicitly empty `value` for any setting whose `GlobalConfiguration.getType()` is not `String`.

What it did not do is validate that a **non-empty** value is parseable as the setting's declared type:

```json
tools/call set_server_setting {"key":"arcadedb.asyncWorkerThreads","value":"abc"}
```

still returns `isError: false`. `ContextConfiguration.setValue(String, Object)` (`engine/src/main/java/com/arcadedb/ContextConfiguration.java:101-106`) stores whatever it is given, with no coercion and no type check, so the `NumberFormatException` surfaces later, inside whichever component reads the setting - far from the call that caused it, and with a success envelope already returned to the model.

The HTTP twin (`PostServerCommandHandler.setServerSetting`) has the same gap: it enforces the `<key> <value>` shape and nothing about the value.

### Why it is not a one-liner

The obvious fix - parse the string against `cfg.getType()` and reject what does not parse - has to pick a parser, and the two live read paths do not agree on what a valid `Integer` setting value looks like:

| Path | Implementation | Accepts `"1MB"`? |
|---|---|---|
| `GlobalConfiguration.getValueAsInteger()` | `FileUtils.getSizeAsNumber(v.toString())` (`GlobalConfiguration.java:2573-2576`) | yes |
| `ContextConfiguration.getValueAsInteger(GlobalConfiguration)` | `Integer.parseInt(v.toString().trim())` (`ContextConfiguration.java:191-196`) | no |

Same for `getValueAsLong`. A validator written against `Integer.parseInt` would reject size-suffixed values that a large part of the codebase reads successfully through the static accessor; one written against `FileUtils.getSizeAsNumber` would accept values that blow up on the `ContextConfiguration` path. Neither is safe to bolt on to the MCP tool alone.

### Suggested shape

1. Decide which parse is canonical for a numeric `GlobalConfiguration`, and make `ContextConfiguration.getValueAsInteger`/`getValueAsLong` agree with `GlobalConfiguration.getValueAsInteger`/`getValueAsLong` (or the reverse). This is the actual bug - two accessors for the same setting disagreeing - and it is independent of MCP.
2. Then add a shared `GlobalConfiguration.isAssignable(String)` / coercion helper and call it from both `SetServerSettingTool` and `PostServerCommandHandler.setServerSetting`, so the MCP tool and its HTTP twin refuse the same values.

A regression test should cover both writers and assert that a value accepted by the setter is readable through **both** accessors without throwing.

robfrank:
Closes #6875

## What was wrong

Two defects that compound.

**1. No coercion at the writer.** The `set_server_setting` MCP tool and `PostServerCommandHandler.setServerSetting` both handed the caller's raw string to `ContextConfiguration.setValue(String, Object)`, which is a plain `Map.put`. #6874 stopped a *missing* and an *empty* `value`; a non-empty one that is not parseable as the setting's declared type still went straight in, so

```json
tools/call set_server_setting {"key":"arcadedb.asyncWorkerThreads","value":"abc"}
```

answered `isError: false` and the `NumberFormatException` surfaced later, inside whichever component read the setting next.

**2. The two read paths disagreed about what parses.** `GlobalConfiguration.getValueAsInteger()/getValueAsLong()` coerce a non-`Number` through `FileUtils.getSizeAsNumber()`, so `"1MB"` reads as `1048576`; `ContextConfiguration.getValueAsInteger(GlobalConfiguration)/getValueAsLong(...)` used `Integer.parseInt`/`Long.parseLong`, so the same stored string threw. Any value reaching the context map without passing through `GlobalConfiguration.setValue` therefore read back differently depending on which accessor the reading component happened to call. A validator written against either one alone would have been wrong for the other, which is why this is fixed first and the coercion helper built on top of it.

Three more, found while wiring the helper in:

- the HTTP command split on the first space with `substring(firstSpace)` rather than `substring(firstSpace + 1)`, so **every** value was stored with a leading blank;
- it stored the `<key>` token verbatim, backticks included, so the documented ``SET SERVER SETTING `arcadedb.foo` 10`` form landed under a key nothing reads - a silent no-op answered with a 200;
- `ALTER DATABASE ... SETTING` is a **third** writer into a `ContextConfiguration`, using the same plain map put, so the identical defect was reachable from SQL.

## Which parse is canonical

`FileUtils.getSizeAsNumber` on the trimmed text, i.e. what the `GlobalConfiguration` accessor has always done, with two guards that keep it from being *looser* than the `Integer.parseInt`/`Long.parseLong` it replaces: an integral setting refuses a value outside the `int` range rather than truncating to its low 32 bits, and it refuses a decimal mantissa rather than dropping the fraction. Within those, it is a strict superset - `"64MB"` now works where `"64"` always did - so nothing that parsed before stops parsing.

## Changes

| File | Change |
|---|---|
| `GlobalConfiguration` | new `coerce(Object)`: the single place a value becomes a setting's declared type, or is rejected with a message naming the key, the type and the offending value. `setValue(Object)` delegates to it. `coerceToIntegral` refuses a fractional value from either route (a boxed `Double` or a decimal mantissa in text); `narrowToInteger` holds the `int` bound and is shared with the read accessors. The three `getValueAs*()` accessors trim and test `Number` rather than the exact box type |
| `ContextConfiguration` | `getValueAsInteger/Long/Float(GlobalConfiguration)` read through the same parse, and the same `int` bound, as their `GlobalConfiguration` counterparts |
| `SetServerSettingTool` | coerces before storing, under the setting's canonical key, so the map holds a typed value neither accessor has to re-parse. `newValue` reports what was stored, masked for a hidden setting on the same terms as `previousValue` |
| `PostServerCommandHandler` | `setServerSetting` and `setDatabaseSetting` share one `applySetting`: unquote the key/value tokens, resolve the setting, drop the separating space, reject an empty value for a non-`String` setting, coerce |
| `AlterDatabaseStatement` | the SQL writer goes through `coerce` too, and reports the coerced value as `newValue` |

### Deliberate calls, all raised in review

- **A numeric setting still reports a bad value as `NumberFormatException`** (a subclass of `IllegalArgumentException`). `GlobalConfigurationTest.typeConversion` asserts that type and is an existing test; the enriched message is used either way, because `For input string: "A"` - what `getSizeAsNumber` surfaces for `"abc"` - names neither the setting nor what was sent.
- **`Boolean` stays as permissive as `Boolean.parseBoolean`.** `readConfiguration()` feeds every system property and environment variable through `coerce` during `GlobalConfiguration`'s static initializer, so turning a boolean typo into a throw would turn it into an `ExceptionInInitializerError` that takes the engine down instead of the setting. `"yes"` therefore still becomes `false` silently; tightening that needs the static-init path handled first and is not this fix.
- **A decimal mantissa is refused outright, not checked for exactness.** An exactness check would accept `"1.5MB"` and reject `"6.7KB"` (6860.8 bytes), and "is this fraction binary-exact against the unit" is not a rule an operator can predict. Refusing every decimal is what the parsers this replaced did.
- **An unknown key is still stored verbatim by the HTTP endpoint**, as it always has been - it carries no type to validate against, and `PostServerCommandHandlerIT.serverSettingCommandsCaseSensitivity` exercises one (`server.httpSessionTimeout`, not a declared setting). The MCP tool remains stricter on that one point; for a *declared* setting all three writers now accept and refuse exactly the same values. Worth a follow-up issue.
- **The command tokenizes before it unquotes**, so quoting cannot put a space in a `<database>` or `<key>` token. Documented on `applySetting` rather than changed: only the trailing `<value>`, whatever remains after the last split, can hold a quoted space, and it does.

## Test plan

- [x] `engine`: `Issue6875SettingValueCoercionTest` - the two accessors agree on a size-suffixed value and on surrounding whitespace; a coerced value is readable through **both** accessors (the regression the issue asks for); both accessors refuse a stored value outside the `int` range, boxed or as text; `coerce` refuses unparseable text, an out-of-range magnitude and a fractional value (`6.7d`, `6.7f`, `"6.7"`, `"6.7KB"`, `"1.5MB"`) while accepting `6.0d` and `"2MB"`; `coerce` returns the declared type; the global setter and the helper agree
- [x] `engine`: `Issue6875AlterDatabaseSettingCoercionTest` - `ALTER DATABASE` refuses an unparseable and a fractional value and leaves the setting readable, stores a coerced typed value, still takes any text for a `String` setting
- [x] `mcp`: `Issue6875SetServerSettingCoercionTest` - rejects `"abc"` for an `Integer` setting and `"half"` for a `Float` one, leaves the setting untouched when it rejects, stores a coerced typed value, accepts a size-suffixed value, masks a hidden setting's new value while still storing it, still takes any text for a `String` setting
- [x] `server`: `Issue6875SetServerSettingHttpIT` - the HTTP twin answers 400 for an unparseable value, stores the value coerced and without the separating space, resolves a backtick-quoted key and unquotes the value, accepts a size-suffixed value
- [x] Bug proven present before the fix: with the two engine files reverted, a probe asserting `ctx.getValueAsLong(COMMIT_LOCK_TIMEOUT)` for a stored `"1MB"` fails with `java.lang.NumberFormatException: For input string: "1MB"`
- [x] `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` - 13782 tests, 0 failures
- [x] `mvn -o -pl server test -DexcludedGroups=benchmark,vector,slow` - 873 tests, 0 failures
- [x] `mvn -o -pl server verify -Dit.test='PostServerCommandHandlerIT,Issue6875SetServerSettingHttpIT,HTTP*IT'` - 95 tests, 0 failures
- [x] `mvn -o -pl mcp verify -DexcludedGroups=benchmark,vector,slow` - 318 tests, 0 failures
- [x] `mvn -o install -DskipTests` over the full reactor - BUILD SUCCESS

Two flakes seen once each across the repeated full-suite runs, both green in isolation and on rerun, neither touching code this PR changes: `Issue5596MergeCoverageTest.aCleanAppendConflictIsStillMerged` (a merge-count race between two writer threads) and `Issue6815SelectTimeoutNotInstalledTest.existsReturnsFalseWhenNotThrowing` (`NoSuchElementException` from `MultiIterator.next` via `SelectExecutor.executeExists` - the 1 ms deadline can trip between its `hasNext()` and its `next()`, which looks like a real latent race worth its own issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Server and database settings now validate and convert values to their configured types.
  * Numeric settings accept surrounding whitespace and size-suffixed values while rejecting fractional or out-of-range values.
  * Invalid or empty values are rejected without changing existing settings.
  * Quoted and backtick-quoted setting names and values are handled correctly.
  * Responses show the stored value after conversion and consistently mask sensitive values.

### DIFF
--- a/engine/src/main/java/com/arcadedb/ContextConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/ContextConfiguration.java
@@ -201,7 +201,7 @@ public class ContextConfiguration implements Serializable {
     final Object v = getValue(iConfig);
     if (v == null)
       return 0;
-    return v instanceof Number n ? n.intValue() : (int) FileUtils.getSizeAsNumber(v.toString().trim());
+    return iConfig.narrowToInteger(v instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(v.toString().trim()));
   }
 
   public long getValueAsLong(final GlobalConfiguration iConfig) {
--- a/engine/src/main/java/com/arcadedb/ContextConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/ContextConfiguration.java
@@ -19,6 +19,7 @@
 package com.arcadedb;
 
 import com.arcadedb.serializer.json.JSONObject;
+import com.arcadedb.utility.FileUtils;
 import com.arcadedb.utility.SystemVariableResolver;
 
 import java.io.Serializable;
@@ -188,25 +189,33 @@ public class ContextConfiguration implements Serializable {
     return getVariable(v.toString(), "");
   }
 
+  /**
+   * Issue #6875: this reads through {@link FileUtils#getSizeAsNumber(Object)}, not {@code Integer.parseInt}, so that
+   * it and {@link GlobalConfiguration#getValueAsInteger()} are one parse rather than two that disagree. The context
+   * map can hold a raw string that never passed through {@link GlobalConfiguration#coerce(Object)} - {@link #fromJSON}
+   * and the {@link Map} constructor both put one straight in - and such a value used to read as 1048576 through the
+   * global accessor and throw {@code NumberFormatException} through this one. {@code getSizeAsNumber} is a strict
+   * superset of {@code Integer.parseInt}, so nothing that read before stops reading.
+   */
   public int getValueAsInteger(final GlobalConfiguration iConfig) {
     final Object v = getValue(iConfig);
     if (v == null)
       return 0;
-    return v instanceof Integer i ? i : Integer.parseInt(v.toString().trim());
+    return v instanceof Number n ? n.intValue() : (int) FileUtils.getSizeAsNumber(v.toString().trim());
   }
 
   public long getValueAsLong(final GlobalConfiguration iConfig) {
     final Object v = getValue(iConfig);
     if (v == null)
       return 0;
-    return v instanceof Long l ? l : Long.parseLong(v.toString().trim());
+    return v instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(v.toString().trim());
   }
 
   public float getValueAsFloat(final GlobalConfiguration iConfig) {
     final Object v = getValue(iConfig);
     if (v == null)
       return 0;
-    return v instanceof Float f ? f : Float.parseFloat(v.toString().trim());
+    return v instanceof Number n ? n.floatValue() : Float.parseFloat(v.toString().trim());
   }
 
   public Set<String> getContextKeys() {
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2528,15 +2528,11 @@ public enum GlobalConfiguration {
       if (type == Boolean.class)
         return iValue instanceof Boolean b ? b : Boolean.parseBoolean(iValue.toString().trim());
 
-      if (type == Integer.class) {
+      if (type == Integer.class)
         // the range check has to cover a boxed Number too, not just the parsed-from-text path: Number.intValue()
         // keeps the low 32 bits, so a Long outside the int range would be stored silently truncated - where the
         // Integer.parseInt this method replaced threw. Every input reaches the same bound.
-        final long parsed = iValue instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(iValue.toString().trim());
-        if (parsed < Integer.MIN_VALUE || parsed > Integer.MAX_VALUE)
-          throw new IllegalArgumentException("outside the range of an Integer");
-        return (int) parsed;
-      }
+        return narrowToInteger(iValue instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(iValue.toString().trim()));
 
       if (type == Long.class)
         return iValue instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(iValue.toString().trim());
@@ -2578,17 +2574,21 @@ public enum GlobalConfiguration {
   }
 
   /**
-   * Returns whether {@link #coerce(Object)} would accept {@code iValue}. Intended for callers that want to reject a
-   * value before storing it rather than handle the exception - see {@link #coerce(Object)} for why the two writers
-   * outside this class have to ask.
+   * Narrows an integral value to the {@code int} an {@code Integer} setting holds, refusing rather than truncating
+   * one outside the range.
+   * <p>
+   * Issue #6875: this is shared by {@link #coerce(Object)}, {@link #getValueAsInteger()} and
+   * {@link ContextConfiguration#getValueAsInteger(GlobalConfiguration)} so that the bound holds on the READ side
+   * too. Not every value reaches a configuration map through {@code coerce}:
+   * {@link ContextConfiguration#setValue(GlobalConfiguration, Object)} is a plain map put, so a boxed {@code Long}
+   * outside the {@code int} range can be stored, and {@code Number.intValue()} would then hand back its
+   * wrapped-around low 32 bits instead of failing.
    */
-  public boolean isAssignable(final Object iValue) {
-    try {
-      coerce(iValue);
-      return true;
-    } catch (final IllegalArgumentException e) {
-      return false;
-    }
+  int narrowToInteger(final long value) {
+    if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE)
+      throw new NumberFormatException(
+          "Value '" + value + "' is not valid for setting '" + key + "' of type Integer: outside the range of an Integer");
+    return (int) value;
   }
 
   public void setValue(final Object iValue) {
@@ -2640,7 +2640,7 @@ public enum GlobalConfiguration {
    */
   public int getValueAsInteger() {
     final Object v = value != nullValue && value != null ? value : defValue;
-    return v instanceof Number n ? n.intValue() : (int) FileUtils.getSizeAsNumber(v.toString().trim());
+    return narrowToInteger(v instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(v.toString().trim()));
   }
 
   public long getValueAsLong() {
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2566,11 +2566,16 @@ public enum GlobalConfiguration {
    * fractional one rather than truncating it.
    * <p>
    * Issue #6875: the {@code Integer.parseInt}/{@code Long.parseLong} this replaced threw on {@code "6.7"}, and
-   * both routes into here can carry a fraction - {@code FileUtils.getSizeAsNumber} reads a plain {@code "6.7"} as
-   * a double and keeps its integral part, and {@code ALTER DATABASE ... SETTING} hands over whatever an arbitrary
-   * SQL expression evaluated to, so an unquoted {@code 6.7} arrives already boxed as a {@code Double}. A
-   * size-suffixed {@code "1.5MB"} is a different thing - a fraction OF A UNIT whose byte count is a whole number -
-   * and stays accepted.
+   * both routes into here can carry a fraction - {@code FileUtils.getSizeAsNumber} reads a decimal mantissa as a
+   * {@code float} and keeps only the integral part of the result, and {@code ALTER DATABASE ... SETTING} hands
+   * over whatever an arbitrary SQL expression evaluated to, so an unquoted {@code 6.7} arrives already boxed as a
+   * {@code Double}.
+   * <p>
+   * A decimal mantissa is refused outright rather than accepted when the product happens to land on a whole
+   * number. {@code "1.5MB"} would, {@code "6.7KB"} (6860.8 bytes) would not, and a rule that turns on which
+   * decimal fraction happens to be a power of two is not one an operator can predict; refusing every decimal is
+   * also exactly what the parsers this replaced did. A size SUFFIX on a whole mantissa - {@code "64MB"} - stays
+   * accepted, which is the widening this issue argued for.
    */
   private long coerceToIntegral(final Object iValue) {
     try {
@@ -2585,10 +2590,9 @@ public enum GlobalConfiguration {
       }
 
       final String text = iValue.toString().trim();
-      final long parsed = FileUtils.getSizeAsNumber(text);
-      if (!text.isEmpty() && text.indexOf('.') > -1 && Character.isDigit(text.charAt(text.length() - 1)))
+      if (text.indexOf('.') > -1)
         throw new IllegalArgumentException("not a whole number");
-      return parsed;
+      return FileUtils.getSizeAsNumber(text);
     } catch (final RuntimeException e) {
       throw invalidValue(iValue, e);
     }
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2529,9 +2529,10 @@ public enum GlobalConfiguration {
         return iValue instanceof Boolean b ? b : Boolean.parseBoolean(iValue.toString().trim());
 
       if (type == Integer.class) {
-        if (iValue instanceof Number n)
-          return n.intValue();
-        final long parsed = FileUtils.getSizeAsNumber(iValue.toString().trim());
+        // the range check has to cover a boxed Number too, not just the parsed-from-text path: Number.intValue()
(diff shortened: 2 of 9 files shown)


robfrank:
Addressed in 09f1e5a.

**The boxed-`Number` input range check - fixed.** Confirmed as described: `Number.intValue()` keeps the low 32 bits, so `ASYNC_WORKER_THREADS.setValue(5_000_000_000L)` stored 705032704 where the `Integer.parseInt` it replaced threw. The two branches are now one bound:

```java
final long parsed = iValue instanceof Number n ? n.longValue() : FileUtils.getSizeAsNumber(iValue.toString().trim());
if (parsed < Integer.MIN_VALUE || parsed > Integer.MAX_VALUE)
  throw new IllegalArgumentException("outside the range of an Integer");
return (int) parsed;
```

`Issue6875SettingValueCoercionTest.coerceRejectsAnOutOfRangeNumberAsWellAsAnOutOfRangeString` covers `coerce`, `isAssignable` and `setValue` for the boxed form, and asserts an in-range `Long`/`Integer` still passes through.

**On `isAssignable()` being currently unused** - deliberate, not an oversight, and I'd rather keep it: the issue's suggested shape names `GlobalConfiguration.isAssignable(String)` explicitly, and the distinction it draws is real for a caller that wants to test a value without either catching an exception or performing the conversion (a validation pass over a config file, a dry-run/preview of a settings change). Both writers here genuinely want the coerced value, so they call `coerce` and let it throw; wiring `isAssignable` in ahead of it would only parse twice.

Re-verified after the change: engine 13776 tests green (one full-suite run had `Issue5596MergeCoverageTest.aCleanAppendConflictIsStillMerged` red - a pre-existing merge-count race between two writer threads, green in isolation and on the rerun, and it touches no code in this PR); `PostServerCommandHandlerIT` + `Issue6875SetServerSettingHttpIT` 29 green; mcp module 317 green.

robfrank:
Addressed in the follow-up commit.

- **Tokenizing before unquoting** - documented on `applySetting` rather than changed. Splitting happens on the raw string, so quoting cannot make a space part of a `<database>` or `<key>` token. It is harmless for what the grammar can address (every `GlobalConfiguration` key is a dotted identifier), and only the trailing `<value>`, which is whatever remains after the last split, can hold a quoted space - which is why the backtick-quoted-key form with a spaced date pattern in `PostServerCommandHandlerIT.setDatabaseSettingCommand` works. Not a regression from this PR, so it stays a documented limitation.
- **Unknown-key asymmetry between the HTTP command and the MCP tool** - agreed it is a UX inconsistency rather than a bug, and agreed it deserves a follow-up issue rather than a behaviour change smuggled into this one: tightening the HTTP side would reject keys the endpoint has always accepted, and `PostServerCommandHandlerIT.serverSettingCommandsCaseSensitivity` exercises one (`server.httpSessionTimeout`, which is not a declared setting). I will file it once this merges.
- **`coerce()` Javadoc length** - leaving as is, for the reason you give.

robfrank:
Addressed in 4196009.

**1. The `Integer` bound now holds on the read side too — confirmed, and it was a regression this PR introduced.** Widening `ContextConfiguration.getValueAsInteger` from `v instanceof Integer` to `v instanceof Number` in the same change that added the trim and size-suffix support turned an out-of-range boxed `Long` from a throw (via the old `Integer.parseInt(v.toString())`) into a silent truncation. The bound is now one shared, package-private helper called from all three sites:

```java
int narrowToInteger(final long value) {
  if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE)
    throw new NumberFormatException("Value '" + value + "' is not valid for setting '" + key + "' of type Integer: outside the range of an Integer");
  return (int) value;
}
```

`coerce(Object)`, `GlobalConfiguration.getValueAsInteger()` and `ContextConfiguration.getValueAsInteger(GlobalConfiguration)` all go through it, so a value stored by a writer that does not coerce cannot read back as its wrapped-around low 32 bits either. `Issue6875SettingValueCoercionTest.bothAccessorsRefuseAStoredIntegralValueOutsideTheIntegerRange` pins it for the boxed form and for the same magnitude stored as text, and asserts an in-range `Long` still reads.

**2. `ALTER DATABASE ... SETTING` now goes through `coerce` as well.** You are right that it is the same defect through a third writer, and since `AlterDatabaseStatement` already holds the `GlobalConfiguration` the fix is one line at the point you named, so I would rather close it here than leave a known-reachable instance behind a PR whose whole claim is that every writer shares one parse. The statement's `newValue` result property now reports the coerced value. Covered by `Issue6875AlterDatabaseSettingCoercionTest`: an unparseable value is refused by the statement that names it and leaves the setting readable, a numeric one is stored as an `Integer`, and a `String` setting still takes any text.

**3. `isAssignable(Object)` removed.** Two reviews flagged it and the project convention is against speculative surface; there is no near-term caller, both writers want the coerced value anyway, and pre-checking with it would only parse twice. `coerce` alone is the API now.

**4. Message wording aligned** — the HTTP command's empty-value rejection now reads `'value' must not be empty for setting '<key>' of type <T>`, identical to the MCP tool's.

Re-verified: engine 13780 tests green; server 873 unit + 95 IT green (`PostServerCommandHandlerIT`, `Issue6875SetServerSettingHttpIT`, `HTTP*IT`); mcp 317 green; `mvn -o install -DskipTests` over the full reactor green.

robfrank:
Addressed in 4837cfe.

**1. Fractional `Number` truncation - fixed, and you are right that it was a narrowing, not a widening.** It was worse than the boxed case you named: `FileUtils.getSizeAsNumber("6.7")` also reads a plain unsuffixed decimal as a double and keeps its integral part, so the text route truncated too, where `Integer.parseInt("6.7")` threw. Both routes now share one `coerceToIntegral(Object)`:

- an `Integer`/`Long`/`Short`/`Byte` passes through as is,
- any other `Number` must satisfy `d == Math.rint(d)` and be finite,
- text goes through `getSizeAsNumber`, and an unsuffixed decimal (a `.` with a digit last) is refused. A size-suffixed `"1.5MB"` stays accepted, because that is a fraction OF A UNIT whose byte count is a whole number.

`coerceRefusesAFractionalValueForAnIntegralSetting` covers `6.7d`, `6.7f`, `"6.7"` for both `Integer` and `Long`, and asserts that `6.0d`, `"1.5MB"` and a `Float` setting's `0.25` are all still accepted. `alterDatabaseRefusesAFractionalValueForAnIntegralSetting` covers the exact SQL you named, `alter database arcadedb.asyncWorkerThreads 6.7` unquoted.

**2. Double-wrapped exception - fixed as part of the same restructure.** The two integral branches now run OUTSIDE `coerce`'s `try`, since `coerceToIntegral` and `narrowToInteger` each already raise a message naming the key, the type and the value. The specific reason ("not a whole number", "outside the range of an Integer") is no longer buried in a `getCause()`. The message construction is shared in one `invalidValue(Object, RuntimeException)`.

**3. Secret's new value now masked.** Agreed the response is worth closing even though it predates this PR: it is a payload a client may log, cache or forward, so the setter should not be the way a credential leaves a server whose getter masks it. `newValue` is masked on exactly the same `cfg.isHidden()` term as `previousValue`, and `masksTheNewValueOfAHiddenSetting` asserts the whole serialized response does not contain the secret while the value is still actually stored.

Re-verified: engine 13782 tests green; server 873 unit + 95 IT green; mcp 318 green; `mvn -o install -DskipTests` over the full reactor green.

One unrelated observation from the sweep, for the record rather than for this PR: a full-suite engine run once hit `Issue6815SelectTimeoutNotInstalledTest.existsReturnsFalseWhenNotThrowing` with `NoSuchElementException` at `MultiIterator.next` <- `SelectExecutor.executeExists`. It is green in isolation and on the rerun, and it looks like a real latent race in `executeExists` (the deadline can trip between its `hasNext()` and its `next()`), not a flake in the test. It touches nothing this PR changes; I will file it separately.

robfrank:
Reviewed and correct on the substance: `"6.7KB"` was silently truncated to 6860. That is the same narrowing you caught last round, one route further in, and it is fixed in a9c1a2b.

**One correction to the diagnosis, so the record is right.** The dot check was not dead code. `FileUtils.getSizeAsNumber("6.7")` does not throw: the last character is a digit, so it takes the `number == true` path into `string2number`, which sees the `.` and returns `Double.parseDouble("6.7")`, and `.longValue()` truncates it to 6. The manual check was what rejected it, and `coerceRefusesAFractionalValueForAnIntegralSetting` asserting `coerce("6.7")` throws is what proves the line is reached. The "unrecognizable format" throw you are thinking of is for text with a dot AND an unrecognized non-digit tail, like `"6.7x"`.

**The fix, and why it is stricter than what you suggested.** Rather than round-tripping the product with an epsilon, `coerceToIntegral` now refuses any text with a decimal mantissa outright:

```java
final String text = iValue.toString().trim();
if (text.indexOf('.') > -1)
  throw new IllegalArgumentException("not a whole number");
return FileUtils.getSizeAsNumber(text);
```

An exactness check would accept `"1.5MB"` and reject `"6.7KB"`, and the rule separating them is "is this decimal fraction binary-exact against the unit" - not something an operator can predict from the value they typed, and it would make `"0.5MB"` legal while `"0.1MB"` is not. Refusing every decimal is also exactly what the `Integer.parseInt`/`Long.parseLong` this replaced did, so it keeps the "cannot reject what worked before" property intact. A size suffix on a whole mantissa (`"64MB"`, `"1KB"`) stays accepted, which is the widening this issue argued for. `coerceRefusesAFractionalValueForAnIntegralSetting` now pins `"6.7KB"` and `"1.5MB"` as rejected and `"2MB"` as accepted.

**On the two minor notes.**
- `LocalDate`/`LocalDateTime` setting types falling through `coerce` unparsed: agreed, unchanged from before this PR, and none of the three text writers can currently reach one with anything but a `String` that would then be stored unconverted. Follow-up rather than scope creep here.
- A literal space inside a quoted value: the value is whatever remains after the LAST split, so it is never split further and a quoted space in it is safe. `PostServerCommandHandlerIT.setDatabaseSettingCommand` exercises exactly that with `"yyyy-MM-dd HH:mm:ss.SSS"`. Only the `<database>` and `<key>` tokens are fragile, and neither can contain a space in practice.

Re-verified after the change: engine 13782 green, server 873 unit + 29 IT green, mcp 318 green, full reactor `install -DskipTests` green.

Note for the developer: this PR has now used its four review cycles, so this last commit is going in without a further bot pass. It is a two-line narrowing in `coerceToIntegral` plus its test, and it corrects a regression this PR itself introduced, which is why I did not leave it for a follow-up.

PR #6885, opened 2026-08-28
fix(ci): shutdown() must join the watcher, not await a latch run() opens from inside itself

robfrank:
## The failure

CI run [33201895826](https://github.com/ArcadeData/arcadedb/actions/runs/33201895826/job/98954707041) went red on `main` with exactly one failure out of 18,407 tests:

```
Issue6762WebSocketEventBusTest.shutdownStillJoinsAfterASeparateStopSignal:128
[shutdown() must have waited for run() to unwind and unregister its listeners]
Expecting value to be false but was true
```

## Root cause

`DatabaseEventWatcherThread.shutdown()` waited on a `CountDownLatch` that `run()` counts down **in its own `finally` block**:

```java
public void shutdown() {
  signalStop();
  if (Thread.currentThread() == this) return;
  runningLock.await();          // opened from inside run()
}

public void run() {
  ...
  } finally {
    ... unregisterListener ...
    runningLock.countDown();    // <- still on the thread, run() has not returned
  }
}
```

The latch therefore opens from **inside** `run()`. The thread is still alive for the JVM's own teardown after `run()` returns, so `shutdown()` could return on a watcher that had not terminated and `isAlive()` was still `true`.

The latch does guarantee the safety-relevant half - the database listeners are unregistered and the queue cleared before it opens - which is why this never showed up as a duplicate-delivery bug. But it is strictly weaker than what two places in the code claim:

- `shutdown()`'s own javadoc: *"Sends the shutdown signal to the thread and waits for termination."*
- `WebSocketEventBus.unsubscribe()`: *"shutdown() blocks until the watcher thread terminates"*

## Reproduction

A 2000-iteration harness (`start(); signalStop(); shutdown(); isAlive()`) against unpatched `main`, five runs:

```
alive-after-shutdown 3/2000
alive-after-shutdown 4/2000
alive-after-shutdown 17/2000
alive-after-shutdown 4/2000
alive-after-shutdown 4/2000
```

Roughly 1 call in 300. Rare enough to survive the review of #6762, frequent enough to redden a nightly.

## The fix

`join()` is the barrier the contract already claims, and it returns strictly later than the latch did. It also cannot hang on a watcher that was constructed but never started, where `runningLock.await()` would have waited forever. The latch has no other reader, so it goes; `run()`'s nested `try/finally` collapses to one.

Behaviour is otherwise unchanged: `signalStop()` still flips the watcher off inside the per-database lock, and `shutdown()` invoked *from* the watcher thread still returns early rather than deadlocking on itself.

## The test

The failing assertion was left as-is in shape but is now repeated 5,000 times. A single shot caught the old code about one time in 300 - that reddens CI without being a regression test. The loop is cheap because each watcher observes `running == false` on its first pass and unwinds without ever polling the queue.

Proven to fail before it was trusted:

| | result |
|---|---|
| strengthened test vs. **unfixed** `DatabaseEventWatcherThread` | **3/3 red**, ~0.3s each |
| strengthened test vs. fixed | 5/5 green |

## Verification

| Suite | Result |
|---|---|
| `Issue6762WebSocketEventBusTest`, `Issue5024WebSocketEventBusConcurrencyTest`, `ChangeEventTest` x5 | 14/14 green each run |
| `WebSocketEventBusIT` (failsafe) | 20/20 green |
| `server` module unit tests, `-DexcludedGroups=benchmark,slow,vector` | **873/873 green** |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket event watcher shutdown reliability.
  * Shutdown now waits until watcher processes have fully stopped, preventing lingering background activity.
  * Improved cleanup of event listeners and queued events during shutdown.
  * Prevented shutdown deadlocks when a watcher stops itself.

### DIFF
--- a/server/src/main/java/com/arcadedb/server/http/ws/DatabaseEventWatcherThread.java
+++ b/server/src/main/java/com/arcadedb/server/http/ws/DatabaseEventWatcherThread.java
@@ -25,7 +25,6 @@ import com.arcadedb.event.AfterRecordUpdateListener;
 import com.arcadedb.log.LogManager;
 
 import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
@@ -34,7 +33,6 @@ final public class DatabaseEventWatcherThread extends Thread {
   private final    ArrayBlockingQueue<ChangeEvent> eventQueue;
   private final    Database                        database;
   private volatile boolean                         running = true;
-  private final    CountDownLatch                  runningLock;
   private final    WebSocketEventListener          listener;
 
   public DatabaseEventWatcherThread(final WebSocketEventBus eventBus, final Database database, final int queueSize) {
@@ -43,7 +41,6 @@ final public class DatabaseEventWatcherThread extends Thread {
     this.eventQueue = new ArrayBlockingQueue<>(queueSize);
     this.database = database;
     this.listener = new WebSocketEventListener(this);
-    this.runningLock = new CountDownLatch(1);
 
     this.database.getEvents()
         .registerListener((AfterRecordCreateListener) listener)
@@ -85,8 +82,14 @@ final public class DatabaseEventWatcherThread extends Thread {
    * that only its own {@code run()} finally-block can count down, deadlocking it forever.
    * <p>
    * Deliberately no "already stopped, nothing to do" shortcut: {@link #signalStop()} may have run first, and returning
-   * on that would skip the join this method exists for. Once {@code run()} has finished the latch is already down, so
-   * a repeated call still returns immediately.
+   * on that would skip the join this method exists for. Once the thread has died {@link #join()} returns immediately,
+   * so a repeated call still costs nothing.
+   * <p>
+   * The wait is a {@link #join()} and not a latch counted down by {@code run()}: such a latch opens from INSIDE
+   * {@code run()}, which leaves the thread still alive for the JVM's own teardown after {@code run()} returns, so
+   * {@code shutdown()} could return on a watcher that was not yet terminated - roughly once in 300 calls, which is how
+   * it reached CI as a flake. {@code join()} is the barrier this method's contract claims, and it also cannot hang on a
+   * watcher that was constructed but never started, where a latch would have waited forever.
    */
   public void shutdown() {
     signalStop();
@@ -95,7 +98,7 @@ final public class DatabaseEventWatcherThread extends Thread {
       return;
 
     try {
-      runningLock.await();
+      join();
     } catch (final InterruptedException e) {
       Thread.currentThread().interrupt();
     }
@@ -113,15 +116,11 @@ final public class DatabaseEventWatcherThread extends Thread {
 
     } catch (final InterruptedException ignored) {
     } finally {
-      try {
-        this.database.getEvents().unregisterListener((AfterRecordCreateListener) listener)
-            .unregisterListener((AfterRecordUpdateListener) listener)
-            .unregisterListener((AfterRecordDeleteListener) listener);
+      this.database.getEvents().unregisterListener((AfterRecordCreateListener) listener)
+          .unregisterListener((AfterRecordUpdateListener) listener)
+          .unregisterListener((AfterRecordDeleteListener) listener);
 
-        eventQueue.clear();
-      } finally {
-        runningLock.countDown();
-      }
+      eventQueue.clear();
     }
   }
 }
--- a/server/src/test/java/com/arcadedb/server/http/ws/Issue5024WebSocketEventBusConcurrencyTest.java
+++ b/server/src/test/java/com/arcadedb/server/http/ws/Issue5024WebSocketEventBusConcurrencyTest.java
@@ -72,8 +72,10 @@ class Issue5024WebSocketEventBusConcurrencyTest {
 
   /**
    * Defect 1: when the watcher thread itself triggers a stop (the last subscriber turns out to be a zombie during
-   * {@code publish}), {@code shutdown()} must not block on its own termination latch. Before the fix the thread parks
-   * in {@code runningLock.await()} inside {@code publish()} inside {@code run()} and never terminates.
+   * {@code publish}), {@code shutdown()} must not block on its own termination. Before the fix the thread waited on
+   * that barrier from inside {@code publish()} inside {@code run()} - so it waited on a termination that only its own
+   * return could produce, and never terminated. The barrier was a latch then and is {@link Thread#join()} now; the
+   * self-shutdown guard is what makes either one safe.
    */
   @Test
   void watcherSelfShutdownDoesNotHang() throws Exception {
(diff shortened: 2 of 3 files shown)


robfrank:
## Cycle 1

**Stale `runningLock` reference in `Issue5024WebSocketEventBusConcurrencyTest:76`** - fixed in bf7f83e. The comment named a field this PR deletes, which my diff caused, so it belongs here rather than in a follow-up. Restated in terms of the barrier instead of its implementation, since what makes the self-shutdown case safe is the `Thread.currentThread() == this` guard - true of the latch and of `join()` alike. `grep -rn runningLock server/src` is now empty.

**The `run()` `finally` collapse** - checking this against `origin/main`, the premise doesn't hold, so I'm not changing it. Here is the old block verbatim:

```java
} finally {
  try {
    this.database.getEvents().unregisterListener((AfterRecordCreateListener) listener)
        .unregisterListener((AfterRecordUpdateListener) listener)
        .unregisterListener((AfterRecordDeleteListener) listener);

    eventQueue.clear();          // <- same try as the unregisters, NOT the inner finally
  } finally {
    runningLock.countDown();     // <- the only statement the nested finally protected
  }
}
```

`eventQueue.clear()` sat in the same `try` as the `unregisterListener` calls, so an unregister that threw skipped it *before* this PR too. The only thing the nested `finally` guaranteed was `runningLock.countDown()`, and that statement no longer exists. So there is no behaviour change here at all, accepted or otherwise:

| unregisterListener throws | before | after |
|---|---|---|
| `eventQueue.clear()` runs | no | no |
| exception propagates out of `run()` | yes | yes |
| `shutdown()` still returns | yes, via `countDown()` in the nested finally | yes, the thread dies and `join()` returns |

The nested `try/finally` existed solely to keep the latch honest under a throwing unregister. Removing the latch removes its only reason to exist, which is why it collapsed rather than being dropped. I'd rather leave `eventQueue.clear()` exactly as unguarded as it has always been than add a guarantee this PR isn't about - as you note, the queue dies with the thread either way.

**`@Tag("slow")`** - agreed, no action. Measured at 0.633s for all 9 methods of the class together, the 5,000-iteration loop included.

**Re-verified after bf7f83e**: `Issue6762WebSocketEventBusTest` + `Issue5024WebSocketEventBusConcurrencyTest` + `ChangeEventTest` = 14/14 green.

robfrank:
## Cycle 2 - CodeRabbit

**`Issue6762WebSocketEventBusTest:134` - interrupt the watcher before joining it: taken** (ae8e59f).

The stall is real and I measured it rather than taking it on faith. Harness parks the watcher inside `poll(500ms)` deliberately, then times `shutdown()` over 20 iterations:

```
interrupt=false  avg shutdown 446ms  worst 449ms
interrupt=true   avg shutdown   0ms  worst   0ms
```

It never fires on my machine in the real test (the whole 9-method class runs in 0.63s, because `start()` returns before the new thread is scheduled), but it is exactly the kind of load-dependent cost that turns a 5,000-iteration loop into minutes on a busy runner - and then into a `@Tag("slow")`, or a loosened bound.

It costs the assertion nothing, and I confirmed that rather than assuming it: both exit paths - loop condition and `InterruptedException` - run the same `finally` and leave the same window between `run()` returning and the thread dying. **Re-proven against the unfixed `DatabaseEventWatcherThread` with the interrupt in place: still 3/3 red, ~0.29s each.** Detection is unchanged; only the stall is gone.

---

**`DatabaseEventWatcherThread.java:101` - clean up a watcher that never starts: declining, with reasoning.**

The mechanics are right - `join()` on a `NEW` thread returns immediately, and the constructor registers the listeners that only `run()` unregisters. But the state is unreachable. There is exactly one construction site in main code:

```
$ grep -rn -A1 "new DatabaseEventWatcherThread" server/src/main --include='*.java'
WebSocketEventBus.java:272:    final var watcherThread = new DatabaseEventWatcherThread(...);
WebSocketEventBus.java-273-    watcherThread.start();
```

`start()` is the next statement, inside `computeIfAbsent`. Nothing can call `shutdown()` on a watcher between those two lines, and if `start()` itself threw the entry is never stored in `databaseWatchers`, so `stop()` and `unsubscribe()` never see it - the leak on that path is the un-stored watcher, which no amount of work inside `shutdown()` can reach.

So the proposal would add a synchronized start/shutdown handshake to a lifecycle path for a state production cannot produce, and the regression test asked for would assert on a branch no caller reaches. That is the opposite direction from this PR, which deletes a synchronization primitive rather than adding one.

Worth being precise about what changed here, since the PR description raised it: this is **not** a regression. Before this PR, `shutdown()` on a never-started watcher blocked forever on `runningLock.await()` - a latch only `run()` could open. After it, the call returns. Returning is strictly better than deadlocking; the javadoc claims only that, not that the never-started watcher is cleaned up. If a caller that constructs without starting ever appears, the fix belongs at that caller, and the deadlock it would have hit is now a silent leak instead - which is why I'd rather it be caught by the reachability argument than papered over inside `shutdown()`.

---

**Verification after ae8e59f**: `Issue6762WebSocketEventBusTest` + `Issue5024WebSocketEventBusConcurrencyTest` + `ChangeEventTest` = 14/14 green, 5 consecutive runs.

robfrank:
## Cycle 3 - `@Tag("slow")`: measured, no action needed

You asked for the wall time to be confirmed under CI load rather than assumed, so here it is. Single method, 5 consecutive runs each:

| | runs | elapsed |
|---|---|---|
| idle | 5 | 0.280, 0.282, 0.286, 0.288, 0.290 s |
| 24 busy loops on 12 cores (2x oversubscription) | 5 | 0.455, 0.465, 0.500, 0.505, 0.511 s |

All 10 green. Worst case under 2x oversubscription is **0.511s** - comfortably sub-second, so it stays in the normal lane. It sits well below the "large batches, multi-second elapsed time" bar in CLAUDE.md, and tagging it `slow` would move a 0.5s test into a lane that exists for tests costing tens of seconds.

The load numbers are also the clearest evidence that the `interrupt()` from cycle 2 was the right call: contention is exactly when a watcher wins the race into `poll(500ms)`, and each such occurrence used to cost `shutdown()` 446ms. Without the interrupt this row is where the test would have grown into the multi-second range that *would* have earned the tag.

Confirming the two points you re-derived independently, since both matched what I checked:
- `eventQueue.clear()` is skipped on a throwing `unregisterListener` in old and new code alike - the nested `finally` only ever protected `countDown()`.
- `join()` returns immediately for a `NEW` thread. Cycle 2 has the reachability argument for why that state cannot occur in production (one construction site, `start()` on the next line inside `computeIfAbsent`).

**Full verification as it stands on ae8e59f:**

| Suite | Result |
|---|---|
| `Issue6762WebSocketEventBusTest` + `Issue5024WebSocketEventBusConcurrencyTest` + `ChangeEventTest`, 5 runs | 14/14 green each |
| strengthened test vs. unfixed `DatabaseEventWatcherThread`, with `interrupt()` | 3/3 red, ~0.29s |
| `WebSocketEventBusIT` (failsafe) | 20/20 green |
| `server` module unit tests, `-DexcludedGroups=benchmark,slow,vector` | 873/873 green |

Issue #6808, opened 2026-08-27
POST/PUT/DELETE /api/v1/server/users mutate users only on the node that served the request, while the equivalent POST /api/v1/server commands replicate through Raft


ruispereira:
### Summary
`POST`/`PUT`/`DELETE /api/v1/server/users` mutate the user store **only on the node that served the request**, while the equivalent `POST /api/v1/server` `create user`/`drop user` commands replicate through Raft. On a cluster this silently diverges the security state, and the next replicated user command latches the divergence in by overwriting every peer.

### Where (HEAD `9c9d024db`)
`server/src/main/java/com/arcadedb/server/http/handler/PostServerCommandHandler.java:816-867` - the HA-aware path:
```java
if (server.getHA() != null) {
  ...
  ha.replicateSecurityUsers(currentUsers.toString());   // -> ArcadeStateMachine -> applyReplicatedUsers on every peer
```
The REST handlers call the local mutators directly and unconditionally, with no HA branch anywhere:
- `PostUserHandler.java:69` - `security.createUser(userConfig);`
- `PutUserHandler.java:69` - `security.updateUser(updatedConfig);`
- `DeleteUserHandler.java:47` - `httpServer.getServer().getSecurity().dropUser(name);`
- `DeleteDropUserHandler.java:56` (deprecated) - same

Those mutators only touch the local `users` map and the local `server-users.jsonl`. Every `replicateSecurityUsers` caller in the tree is `PostServerCommandHandler` or `PostAddPeerHandler` (peer seeding).

### Repro
On a 3-node Raft cluster:
- `POST /api/v1/server/users` on node 0 with `{"name":"bob","password":"secret12","databases":{"mydb":["admin"]}}` returns 201, but `bob` gets 401 on node 1 and node 2, so a client behind a load balancer authenticates intermittently.
- Worse in reverse: `DELETE /api/v1/server/users?name=bob` returns 200 while `bob`'s credentials keep working on the other two nodes. The operator believes the revocation succeeded.
- The divergence is then latched in: the next `POST /api/v1/server` `create user`/`drop user` on any node replicates *that* node's whole user list (`getUsersJsonPayload()`), overwriting every peer and silently erasing or resurrecting users the REST endpoints created/deleted elsewhere.

[...]

PR #6886, opened 2026-08-28
fix(#6808): restore the user-file order, and teach GrpcAuthInterceptorTest that a token needs a live principal

lvca:
Follow-up to #6866, already merged. Two CI failures on that PR's run were genuinely caused by it.

**1. `server-users.jsonl` user order changed** (`ServerProfilingIT`, 12 methods red)

The persist-before-publish change built the file from a candidate list with the changed user appended last, instead of from the in-memory map — so a newly created user moved to the end of the file. Same protection, without touching the order: publish to the map, save, and undo the publish if the write failed. A failed write still leaves the previous state in force in memory and on disk alike. The replicated payload keeps using the shared snapshot helper, where appending matches what the replication path already did.

**2. `GrpcAuthInterceptorTest.validateTokenReturnsTrueForValidSession`**

The gRPC interceptor now re-checks the session's principal against the live users map, so a token stops working the moment its user is dropped rather than when it idle-expires. The test stubbed `getUsers()` but not `getUser()`, so its "valid" session had no principal and the call was correctly refused — the test asserted the old behaviour. Stubbed the lookup and added `tokenOfADroppedPrincipalIsRejected` to pin the new one.

Other failures on that run were checked and are not from #6866: `GraphAnalyticalViewTest.issue6641_*` and the `ha-integration-tests` lane are the documented chronic families, and `RemoteConsoleIT.userMgmt` failed because the branch predated the #6830 test update that `main` already carries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication validation for tokens backed by active sessions.
  * Tokens are now rejected when the associated user has been removed from the current security configuration.
  * Improved reliability when creating, updating, or deleting local security users; failed saves now restore the previous configuration.
  * Added regression coverage to help prevent authentication inconsistencies.

### DIFF
--- a/grpcw/src/test/java/com/arcadedb/server/grpc/GrpcAuthInterceptorTest.java
+++ b/grpcw/src/test/java/com/arcadedb/server/grpc/GrpcAuthInterceptorTest.java
@@ -119,6 +119,10 @@ class GrpcAuthInterceptorTest {
 
     // getUsers() returns Set<String> - need at least one user for security to be enabled
     when(mockSecurity.getUsers()).thenReturn(Collections.singleton("testuser"));
+    // The token is only honored while its principal still exists (issue #6808): the session captured its
+    // ServerSecurityUser at login, so the interceptor re-checks the name against the live users map before
+    // letting the call through. A valid session therefore needs a principal that still resolves.
+    when(mockSecurity.getUser("testuser")).thenReturn(mockUser);
 
     GrpcAuthInterceptor interceptorWithSession = new GrpcAuthInterceptor(mockSecurity, mockSessionManager);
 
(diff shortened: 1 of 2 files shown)


lvca:
Superseded by #6890, already merged into main.

Both CI failures this PR targets are fixed there:

- `server-users.jsonl` order: #6890 fixes the actual bug, which was in the test - `ServerProfilingIT.checkJohnUser()` was asserting a specific line position that was never a contract nothing else reads. It now asserts which users were persisted instead of positional order. Production code is untouched.
- `GrpcAuthInterceptorTest.validateTokenReturnsTrueForValidSession`: #6890 fixes it the same way this PR does (stub `getUser("testuser")`), plus adds a regression test for the dropped-principal case that also asserts `removeSession` eviction - a superset of the coverage here.

The remaining delta in this PR - reworking `ServerSecurity.createUser`/`updateUser`/`dropUserLocally` to publish to the `users` map before the write is durable, undoing on failure - trades a cosmetic, unread ordering detail for a real race: `getUser`/`getUsers`/`authenticate` read that map lock-free by design, so a concurrent login can authenticate against a brand-new user or a rotated password before `saveUsers()` confirms the write, and `updateUser`'s rollback path doesn't invalidate the session issued against state that was never persisted. Both CodeRabbit and the Claude reviewer flagged this independently on this PR as a high-risk issue, and it stayed unresolved.

Closing in favor of #6890.

PR #6887, opened 2026-08-28
fix(#6792): separate GAV node count from ID bound

### LINKED ISSUE #6792: A Graph Analytical View's dense node ids can exceed getNodeCount() once a base vertex is deleted, so algo.* silently drops nodes
Raised in review of PR #6788 (issue #6315), verified to be independent of it.

## What happens

`GraphAnalyticalView.getNodeCount()` answers `DeltaOverlay.getTotalNodeCount()` while an overlay is active:
`baseNodeCount - deletedBase + overflowCount - deletedOverflow`. The dense id space is not renumbered to
match. Delete one base vertex and the surviving vertices keep the ids they had, so the largest live id is now
`nodeCount` or more.

`AbstractAlgoProcedure.GraphData` sizes every array from that count and iterates `0..nodeCount-1`, so:

- **a live vertex whose id landed above the count is never visited** - it is silently absent from the answer
  of every `algo.*` procedure, which is a wrong result with no exception;
- **a neighbour id above the count indexes out of the arrays** - `AlgoAPSP`'s `dist[i][j]` and `AlgoMaxFlow`'s
  `capacity[i][j]` are the two that index by neighbour rather than only iterating, so an
  `ArrayIndexOutOfBoundsException` is reachable in principle.

Reproduced on `fix/6315-edge-property-overlay-resolution` with a three-vertex graph (`A`, `B`, `SPARE`), a
`SYNCHRONOUS` view with compaction disabled, `SPARE` deleted and one vertex added:

```
nodeCount=2
  A=2  out=[1]  in=[]      <- id 2 with a node count of 2: A is never iterated
  B=1  out=[0]  in=[2]     <- and B reports a neighbour that indexes past every array
  FRESH=0
```

## Not caused by #6315

Checked deliberately, because that PR gave the weighted columnar path a new way to produce these ids. It is
not the source: the same run with the view materialising no edge property columns - which sends the same
procedure down `weightedAdjacencyFromRecords` instead, the path that existed before - produces byte-identical
ids and the same answer. `getNeighborIds` has always served overlay ids, and `getNodeId` has always resolved
them, so every path into `GraphData` could already carry one.

## What it would take

Either the id space or the count has to move, and they are not equally cheap:

- **renumber** on every overlay change - correct, and defeats the point of an overlay;
- **make `getNodeCount()` the id-space bound rather than the live-vertex count**, i.e. answer
  `baseNodeCount + overflowCount`, and let the deleted ids be holes the callers already have to tolerate
  (`getRID` answers `null` for them, and `GraphData.getVertex` already skips a null). This looks closest to
  what the callers assume, but `getNodeCount()` is public API and its current answer is the honest one for
  anyone asking "how many vertices are in this view";
- **have `GraphData` size itself from the id space** and keep `getNodeCount()` meaning what it says, which
  needs a second accessor for the bound.

The third is the one that changes no existing meaning. Worth a regression test that walks every `algo.*`
procedure over a view with both a deleted base vertex and an added one.

justinblethrow-cloud:
## Summary

- preserve `GraphTraversalProvider.getNodeCount()` as the live-node count
- add a backward-compatible `getNodeIdUpperBound()` contract for node-ID-indexed storage
- expose the monotonic overlay ID bound from `GraphAnalyticalView` and size provider-backed `GraphData` from it
- add a regression that deletes the lowest base ID, adds an overflow vertex and edge, and verifies `algo.apsp` returns the high-ID vertex

## Why

A synchronous delta overlay retains deleted ID slots and allocates overflow IDs monotonically. The live-node count can therefore be smaller than the largest live ID plus one. Sizing algorithm arrays and iteration from the live count silently omitted high-ID vertices and could leave a neighbor ID outside those arrays.

The new accessor keeps the existing live-count meaning while giving node-indexed consumers the exclusive bound they require. Its default implementation preserves compatibility for providers with compact IDs.

## Validation

```text
./mvnw -pl engine -Dtest=Issue6792GraphDataNodeIdBoundTest,GraphTraversalProviderCountEdgesBetweenTest,DeltaOverlayTest,GraphAnalyticalViewTest,AlgoAPSPTest,Issue6296AlgoAPSPLazyRowsTest,AlgoMaxFlowTest test

Tests run: 190, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Closes #6792

@lvca, this issue is assigned to you; I went ahead because the change is small and follows the third option described in the issue. If you had a different implementation planned, I am happy to withdraw this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved analytical graph processing for graphs with deleted or sparse node IDs.
  - Corrected graph sizing to preserve valid node-ID ranges, including gaps.
  - Ensured neighbor lookups, shortest-path analysis, and CSR-accelerated operations continue to return correct results after node deletion and insertion.
  - Updated graph algorithms to use a safe fallback when recent graph changes are not yet reflected in analytical data.

### DIFF
--- a/engine/src/main/java/com/arcadedb/graph/DenseNodeIdProvider.java
+++ b/engine/src/main/java/com/arcadedb/graph/DenseNodeIdProvider.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.graph;
+
+import com.arcadedb.database.RID;
+
+import java.util.Arrays;
+import java.util.function.IntConsumer;
+
+/**
+ * A {@link GraphTraversalProvider} whose node ID space has no holes in it, wrapped around one whose has.
+ * <p>
+ * A provider that allocates node IDs monotonically and does not renumber when a node is deleted - which is what
+ * a Graph Analytical View's delta overlay does - reports a live node count smaller than the exclusive bound of
+ * its own ID space. Every consumer that indexes an array by node ID then has two ways to be wrong at once, and
+ * neither of them is an error it can attribute (issue #6792):
+ * <ul>
+ *   <li>sizing the array from {@link GraphTraversalProvider#getNodeCount()} leaves the highest live nodes
+ *       outside it, so they are silently absent from the answer;</li>
+ *   <li>sizing it from {@link GraphTraversalProvider#getNodeIdUpperBound()} covers them, but then every array
+ *       carries a slot per deleted node and every enumeration of it has to remember to skip those slots -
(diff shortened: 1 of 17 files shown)


lvca:
Thanks @justinblethrow-cloud

PR #6889, opened 2026-08-29
Update Python bindings: examples CI resilient to a third-party dataset host, wheel freshness by content, citation via CITATION.cff

tae898:
Follow-up to #6790, which was closed. Same scope, plus the fix for the examples CI failure on `main`.

## The CI fix

`main` is currently red on `Test Python Examples` — 12 jobs, every OS, Python 3.10 through 3.14 — and it is not a bindings problem:

```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate has expired
https://files.grouplens.org/datasets/movielens/ml-latest-small.zip
```

The TLS certificate for `files.grouplens.org` (University of Minnesota, the MovieLens host) expired at `Aug 28 23:59:59 2026 GMT`. Verified against the live host, not just the log. Re-running does not clear it, because a re-run happens now and the certificate is still expired; we reproduced the same failure on our fork at the same commit that had passed nine hours before the expiry.

`Download ArcadeDB JARs` succeeded, and the failing step is `Download datasets`, before any example runs.

The problem worth fixing is not the lost coverage. `download_data.py` re-raised on any download error, so one unreachable third-party host failed the whole job — which means a lapsed certificate in Minnesota and a genuine regression in these bindings looked identical on your `main`, under a PR title naming the bindings.

So:

- `download_data.py` raises `DatasetUnavailable` and exits **75** (`EX_TEMPFAIL`) for **transport failures only**: `URLError`, `HTTPError`, `TimeoutError`, `ConnectionError`, `ssl.SSLError`. Deliberately not a catch-all — a zip that downloads and then fails to extract, or a dataset that parses wrong, still fails hard. Checked both ways: expired cert, unreachable host and HTTP 503 skip; a bad CSV and a corrupt archive do not.
- `test-python-examples.yml` routes the three downloads through a `fetch()` helper that treats 75 as a skip with a `::warning` naming the host, and any other non-zero as a hard error.

The examples that need the dataset are skipped and named; the rest of the suite still runs.

## The rest

- `scripts/build.sh`: a wheel whose docker layers all cache-hit was reported as "no wheel produced", because `docker cp` preserves the timestamp from inside the container and the freshness test compares mtimes. When `JAR_LIB_DIR` is set the next step already compares the SHA256 of the integration jar inside the wheel against the one on disk, so that content check is now the gate. Without `JAR_LIB_DIR` the strict failure stands.
- `citation.py` and `test_citation.py` removed, and the `__init__` export with them: the package cited itself through a runtime Zenodo call. `CITATION.cff` at the repo root already carries that metadata without a network round-trip at import time. Nothing else in `bindings/python` referenced the module.

Scope is the usual minimum: `bindings/python` plus the workflow that runs those tests. Fork-only tooling (profiling harnesses, benchmark drivers, local build helpers) is excluded.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Python example downloads with TLS support and clearer handling when third-party dataset hosts are unavailable.
  - Dependent examples now skip gracefully with a warning instead of producing confusing failures.
  - Preserved hard-failure behavior for unrelated download errors.

- **Build Improvements**
  - Local Python package builds now defer certain stale-artifact warnings to integrity verification.
  - Superseded pull request checks are canceled to reduce redundant runs.

- **Removed**
  - Removed the Python citation helper, along with its documentation and tests.


tae898:
Good catch, and it was the half of the fix that mattered. `SKIPPED_DATASETS` lived only in the `Download datasets` shell, and every step is its own process, so the example loop never saw it: the job would still have run `04_csv_import_documents.py --dataset movielens-small` against data that was never fetched. Failing later with a worse error than the one it replaced is not an improvement.

Pushed `75293c7a2`:

- the download step writes the list to `$GITHUB_ENV`, plus a `::notice` naming what was skipped;
- the example loop maps each example to the dataset it needs and counts it as **skipped** instead of running it.

The mapping sits in the same file as the one the download step uses to decide what to fetch, so the two cannot drift into disagreeing about what a dataset is for.

One thing worth mentioning because it is easy to get wrong here: `stackoverflow-tiny` is a prefix of `stackoverflow-tiny-novec`, so an unpadded match would let one unavailable host skip the other's examples. The membership test is space-padded and all four combinations are exercised — only exact labels skip.

Issue #6848, opened 2026-08-27
Issue5410AbandonedTicketReleaseIT only passes because RaftHAComprehensiveIT ran before it: an abandoned entry's vertex never replicates when it does not


lvca:
Found while measuring #6343 item 2 (isolating the heavy `ha-raft` ITs into their own failsafe fork). The fork split did not break this test - it revealed that the test has never stood on its own, and that behind it is a committed Raft entry whose vertex never becomes visible.

## The measurement

Two runs of the full `ha-integration-tests` lane, same commit, same instrument, **same 120-second wait budget**, differing only in whether the five heaviest ITs run in their own failsafe fork:

| | `awaitValue(2)` in `Issue5410AbandonedTicketReleaseIT` | class result |
|---|---|---|
| one shared fork (today's `main` arrangement) — [run 33074663154](https://github.com/ArcadeData/arcadedb/actions/runs/33074663154) | satisfied after **252 ms**, then **1 ms** | passes, 13.70 s |
| heavy ITs in their own fork — [run 33074660773](https://github.com/ArcadeData/arcadedb/actions/runs/33074660773) | satisfied after 2 ms, then **GAVE UP after 120 022 ms** | fails, 131.2 s |

So it is **not** a timing sensitivity. 120 seconds is 8× the current budget and 4× the old one, and the second vertex never arrives at all:

```
[Node 1 must hold both vertices] expected: 2L but was: 1L
Suppressed: DatabaseComparator$DatabaseAreNotIdentical:
            Page PageId(graph/28/0) ... DB1 2 <> DB2 1 (sameContent=false)
```

The teardown comparator agrees the databases really do differ. Reproduced 3/3 on the PR branch before these two controlled runs.

## What actually differs

The class order is otherwise identical (index 27, same immediate predecessors `Issue4458TsWalVersionGapIT` → `RaftReadConsistencyBookmarkIT`). The one difference is three classes earlier:

- shared fork: `RaftHAComprehensiveIT` — ~344 s of continuous three-server activity — runs at index 24
- split: it has moved to the other fork, and `RaftSchemaReplicationIT` occupies that slot

**`Issue5410AbandonedTicketReleaseIT` passes only when `RaftHAComprehensiveIT` has run before it in the same JVM.**

## Why this matters beyond the test

The test deliberately induces the state it is named for:

```
TEST: forcing dispatched-timeout for a TX_ENTRY on db=graph
```

[...]

PR #6915, opened 2026-08-30
fix(#6794): guard the tag filters no command line may narrow

### LINKED ISSUE #6794: ha-raft fork split: step 5 of the pom checklist (hostile -Dgroups/-DexcludedGroups) is the one guarantee with no automated regression test
Follow-up from #6343 / #6789, filed as a tracker rather than left as a `// TODO`, since a comment not being a tracker is the exact lesson #6297 and #6323 taught and #6343 was filed to close.

## What is unguarded

`ha-raft/pom.xml` splits the module's ITs into two failsafe executions, and its rationale block ends with a five-step recipe for re-verifying the split after any edit. Four of those steps now have automated guards:

| step | guard |
|---|---|
| 1. the partition runs each class exactly once | `HeavyForkWiringIT` (routing), `HeavyItForkIsolationTest` (tag set) |
| 2. each heavy class gets its own JVM | manual pid check, but a break shows up as the shared-fork behaviour `HeavyForkWiringIT` detects |
| 3. a failing heavy IT names its own execution | manual, low-consequence |
| 4. both guard tests green | themselves |
| **5. no command-line tag filter can narrow either execution** | **nothing** |

Step 5 is the one that matters most and has the least behind it. Both halves of the split are deliberately written as explicit tag expressions - `any() | none()` and `${failsafe.excludedGroups}` - precisely so that a stray `-Dgroups` or `-DexcludedGroups` cannot narrow them. That property was measured (PR #6789 records both the before and after), but nothing re-checks it. Someone simplifying `<groups>any() | none()</groups>` to something that "obviously means the same thing" would silently reopen it, and the failure mode is the bad one: the default execution runs **0 tests and the build passes**, so ~118 ITs vanish from the lane without a red mark anywhere.

## Why it is not tested today

It needs a real Maven subprocess with command-line properties, which a unit test cannot do. `maven-invoker-plugin` is the standard answer and would be a new build-time dependency and a new lane cost, which did not seem worth it inside #6789.

## What to do

Either:

1. a `maven-invoker-plugin` integration test that runs `ha-raft` with `-Dgroups=bogus-tag -DexcludedGroups=bogus-tag -Dfailsafe.excludedGroups=ha-heavy` and asserts both executions still select what they should - the thorough answer, at the cost of a plugin and some lane time; or
2. a cheaper approximation: a CI step that greps the effective POM (`help:effective-pom`) for the two executions' `<groups>`/`<excludedGroups>` and fails if either is absent or empty. It does not prove the runtime behaviour, only that the parameters are still explicitly configured - which is the actual precondition, and is most of the value for none of the cost.

(2) is probably the right trade unless somebody wants the real thing.

## Reproducing the hole

On a tree where the default execution's `<groups>` is removed:

```
./mvnw -o -pl ha-raft verify -Pintegration -DskipTests \
  -Dit.test=RaftVerifyDatabaseIT -DfailIfNoSpecifiedTests=false -Dgroups=bogus-tag
```

reports `Tests run: 0` for the `default` execution and `BUILD SUCCESS`. With `<groups>any() | none()</groups>` present, the same command runs the test.

robfrank:
Closes #6794

## What this does

Issue #6794 asked for an automated guard behind step 5 of the `ha-raft` fork-split re-verification recipe: *no command-line tag filter can narrow either failsafe execution*. That step matters most and had the least behind it, and its failure mode is the quiet one - the execution runs **0 tests and the build passes**, so ~118 ITs leave the lane with nothing red anywhere.

`PomTagFilterContractTest` (in `ha-raft`, 13 tests, under a second, no Maven subprocess) reads every `pom.xml` in the repository and checks how surefire's and failsafe's `groups`/`excludedGroups` are configured. Those parameters each carry a same-named `-D` user property, and Maven resolves a plugin parameter from the configuration first and from the user property only where the configuration leaves it unset. Both ways of getting that precedence wrong are invisible, and they point in opposite directions, so the test enforces both:

| rule | what it catches | issue |
|---|---|---|
| a **plugin-wide** default must be a property reference | a literal outranks the command line, so every `-Dgroups`/`-DexcludedGroups` in `.github/workflows/` quietly stops meaning anything | #5697 |
| failsafe's default must not read surefire's property | ITs silently inherit the unit lane's exclusion (e.g. `@Tag("benchmark")` ITs vanish from every `-Pintegration` run) | #5697 |
| a **named execution** that is half of a tag partition must write out **both** parameters, neither blank nor a bare `${groups}`/`${excludedGroups}` | a `-Dgroups` aimed at another module narrows the execution to nothing, silently | **#6794** |

## Why not `help:effective-pom`

The issue sketched option (2) as a CI step that greps the effective POM. Interpolation is precisely the distinction under test: the effective pom resolves `${excludedGroups}` to `benchmark`, so it cannot tell a passthrough from the literal that breaks it. Reading the sources as written is both more accurate and free - no `mvnw` invocation, no new plugin, no new lane - so it runs in the ordinary unit lane and on a developer's machine rather than only in a workflow step. Option (1), `maven-invoker-plugin`, would prove runtime behaviour at the cost of a build-time dependency and lane time; the precondition it would be proving is exactly what is asserted here.

## The awkward part, stated plainly

**The `ha-heavy` split is not on `main`.** It was reverted in f4567f6176 (in the very PR, #6789, that #6794 was filed as a follow-up to) because it surfaced a real product failure, now issue #6848. `ha-raft/pom.xml` has one plugin block and no failsafe executions, and the CLAUDE.md checklist the issue quotes was removed with it.

So the partition rule currently has nothing in the reactor to judge. Rather than ship a test that cannot fail, it is:

- **proven able to fail against the split's own configuration** - the pre-revert `default` / `heavy-its-in-their-own-fork` pair is a fixture that must pass, and the three shapes #6794 names as regressions (parameter omitted, parameter written as the bare user property, parameter blank) are fixtures that must be reported, by execution id and parameter name;
- **verified to arm itself in the reactor** - planting an execution missing its `<groups>` in `ha-raft/pom.xml` fails the build with `ha-raft/pom.xml -> maven-failsafe-plugin execution 'heavy-its-in-their-own-fork' -> <groups> is not configured, so it resolves through the ${groups} user property`. No edit here will be needed when the split returns.

The other two rules are live and armed today against the parent pom, which is what stops this from being a test with no subject.

## Verification

Every rule was shown able to fail before being trusted, by perturbing the tree and re-running:

| perturbation | result |
|---|---|
| surefire's `<excludedGroups>${excludedGroups}</excludedGroups>` -> `benchmark` | `inheritedTagDefaultsStayOverridableFromTheCommandLine` fails |
| failsafe's `${failsafe.excludedGroups}` -> `${excludedGroups}` | `integrationTestsDoNotInheritTheUnitLaneTagExclusion` fails |
| a partitioned execution planted in `ha-raft/pom.xml` without `<groups>` | `aTagPartitionedExecutionPinsBothOfItsTagFilters` fails, naming it |
| a partitioned execution with an embedded `${excludedGroups}` planted in `ha-raft/pom.xml` | the same rule fails, naming the interpolation |
| `engine/` added to the scan's prune list | `theScanReachesTheBuild` fails, though the pom count still clears its floor |
| `ExecutionKey`'s plugin-declaration identity collapsed back to `(pom, plugin, id)` | `executionsSharingAnIdAcrossProfilesAreJudgedSeparately` fails |
| unperturbed | 13/13 green |

The premise was re-measured on this tree (surefire 3.5.6) rather than taken from the issue:

```
./mvnw -o -pl ha-raft test -Dgroups=bogus-tag
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

`theScanReachesTheBuild` exists so the three "collection is empty" assertions cannot pass by scanning nothing.

## Test plan

- [ ] `./mvnw -o -pl ha-raft test -Dtest=PomTagFilterContractTest -DfailIfNoSpecifiedTests=false` - 13 tests green
- [ ] `./mvnw -o -pl ha-raft test` - the module's unit tests green (991 at the time of the first commit). Note: this dies in a fork start (`The forked VM terminated without properly saying goodbye`, at `LeaveClusterTest`) whenever a second `ha-raft` run is live in another worktree; a control run of that class on unmodified `main` reproduces it, so it is not this change
- [ ] Change `<excludedGroups>${excludedGroups}</excludedGroups>` in the parent pom to a literal and confirm the build now fails with the #5697 message
- [ ] Point failsafe's `<excludedGroups>` back at `${excludedGroups}` and confirm the second rule fires
- [ ] Add a `<execution>` with an id and only one of the two tag parameters to any module's surefire/failsafe block and confirm the partition rule names it

## Notes for the reviewer

- No new dependency, no new workflow step, no change to what any lane runs today. The two `pom.xml` edits are comments pointing at the guard, so the next person editing those lines learns what will fail.
- The check is in `ha-raft` because that is the only module whose executions are ever meant to be a tag partition, and it is where the split will return. `ha-raft/CLAUDE.md` says so, since a repo-wide pom test in this module is otherwise a surprise.

## Review rounds

Three rounds of bot review; two rounds found real gaps in the guard itself, both now closed and pinned by fixtures that fail if the fix is undone:

1. an interpolated `${groups}`/`${excludedGroups}` **anywhere** in a partitioned execution's value, not only as the whole of it, still hands the command line a term in an expression meant to be immune to it. The check is containment over both property names now; `${failsafe.excludedGroups}` is deliberately not caught, and a fixture pins that too;
2. `ExecutionKey` was `(pom, plugin, executionId)`, but an execution id is unique only inside one `<plugin>` block - two profiles declaring `<execution><id>default</id>` with complementary halves of the filter merged into one execution that looked pinned. The key carries the plugin block's document position now.

Also applied: Windows-safe path assertions, the failsafe-exclusion rule added to `ha-raft/CLAUDE.md`, a note that the plugin scan deliberately covers `pluginManagement` and profiles, and fixtures naming the plugin they model.

Declined with reasons (see the [review-round-3 comment](https://github.com/ArcadeData/arcadedb/pull/6915#issuecomment-5469515499)): requiring both parameters of an execution that declares *neither* (that is an ordinary execution, not half of a partition - pinned as a decision by `anExecutionWithNoTagFilterIsNotAPartition`), caching the scan in a `@BeforeAll` static, and dropping the `hasSizeGreaterThan(20)` floor.

### DIFF
--- a/ha-raft/CLAUDE.md
+++ b/ha-raft/CLAUDE.md
@@ -66,3 +66,31 @@ Two things about how it hid, both worth generalizing:
 Since #6220 `TRUNCATE` only batches when **it** owns the transaction - when none was active as it started. Inside a caller's transaction it commits nothing at all, because committing through a caller silently breaks that caller's `ROLLBACK`. So the shape that reproduces #5492 is now `command("sql", "TRUNCATE ...")` with **no** enclosing transaction; wrapping it in one, which `Issue5492TruncateBatchNotReplicatedIT` used to do, leaves nothing mid-execution to commit and the test passes for the wrong reason. `REBUILD INDEX` and `BatchStep` still commit unconditionally.
 
 When adding any code path that commits, ask which instance it holds. `Issue5492TruncateBatchNotReplicatedIT` guards the statement path; it asserts the WAL gap counter *before* querying the follower, because the resync reinstalls the follower's database and a stale handle then throws `DatabaseIsClosed`, which reads like infrastructure noise rather than divergence.
+
+## A repo-wide pom guard lives in this module's tests
+
+`PomTagFilterContractTest` is not about Raft. It reads every `pom.xml` in the repository and checks how
+surefire's and failsafe's `groups`/`excludedGroups` are configured, because those parameters each have a
+same-named `-D` user property and getting the precedence wrong is invisible: the build runs the wrong set of
+tests and passes.
+
+It enforces two rules that point in opposite directions:
+
+- a **plugin-wide** default must be a property reference (`${excludedGroups}`, `${failsafe.excludedGroups}`), so
+  a CI lane's `-DexcludedGroups=...` still reaches it. A literal there wins over the command line and every lane
+  that filters by tag quietly runs whatever the pom named instead. That is issue #5697;
+- a **named execution** that is one half of a tag *partition* must write out both parameters, so no `-Dgroups`
+  or `-DexcludedGroups` aimed at another module can narrow it. Leaving one out is the same as setting it from
+  the command line, and the result is the quiet one: measured on this tree with surefire 3.5.6,
+  `./mvnw -o -pl ha-raft test -Dgroups=bogus-tag` reports `Tests run: 0` and `BUILD SUCCESS`. That is issue
+  #6794 - step 5 of the fork-split re-verification recipe, the one step that had no automated guard.
+
+The second rule is here rather than in `engine` because the only executions ever meant to be a tag partition are
+this module's: the `ha-heavy` split from issue #6343. That split is not on `main` - it was reverted in
+f4567f6176 because it surfaced issue #6848 - so the rule currently has nothing in the reactor to judge and is
+proven able to fail against the split's own configuration as a fixture. It needs no edit to arm itself when the
+split comes back.
+
+It reads the poms as written, not `help:effective-pom`, on purpose: interpolation is the distinction being
+tested, and the effective pom resolves `${excludedGroups}` to `benchmark`, which is exactly the literal the
+first rule rejects.
--- a/ha-raft/CLAUDE.md
+++ b/ha-raft/CLAUDE.md
@@ -74,18 +74,22 @@ surefire's and failsafe's `groups`/`excludedGroups` are configured, because thos
 same-named `-D` user property and getting the precedence wrong is invisible: the build runs the wrong set of
 tests and passes.
 
-It enforces two rules that point in opposite directions:
+It enforces three rules, the first and the last of which point in opposite directions:
 
 - a **plugin-wide** default must be a property reference (`${excludedGroups}`, `${failsafe.excludedGroups}`), so
   a CI lane's `-DexcludedGroups=...` still reaches it. A literal there wins over the command line and every lane
   that filters by tag quietly runs whatever the pom named instead. That is issue #5697;
-- a **named execution** that is one half of a tag *partition* must write out both parameters, so no `-Dgroups`
-  or `-DexcludedGroups` aimed at another module can narrow it. Leaving one out is the same as setting it from
-  the command line, and the result is the quiet one: measured on this tree with surefire 3.5.6,
+- **failsafe's** plugin-wide `<excludedGroups>` must not read the property surefire's reads. Failsafe's own
+  parameter default *is* `${excludedGroups}`, so sharing it applies the unit lane's exclusion to the integration
+  lane and drops e.g. `@Tag("benchmark")` ITs from every `-Pintegration` run. That is the second half of #5697;
+- a **named execution** that is one half of a tag *partition* must write out both parameters, and neither may
+  mention `${groups}` or `${excludedGroups}` anywhere in its value, so no `-Dgroups` or `-DexcludedGroups` aimed
+  at another module can reach it. Leaving one out, or interpolating one into a larger expression, is the same as
+  handing it to the command line, and the result is the quiet one: measured on this tree with surefire 3.5.6,
   `./mvnw -o -pl ha-raft test -Dgroups=bogus-tag` reports `Tests run: 0` and `BUILD SUCCESS`. That is issue
   #6794 - step 5 of the fork-split re-verification recipe, the one step that had no automated guard.
 
-The second rule is here rather than in `engine` because the only executions ever meant to be a tag partition are
+The partition rule is here rather than in `engine` because the only executions ever meant to be a tag partition are
 this module's: the `ha-heavy` split from issue #6343. That split is not on `main` - it was reverted in
 f4567f6176 because it surfaced issue #6848 - so the rule currently has nothing in the reactor to judge and is
 proven able to fail against the split's own configuration as a fixture. It needs no edit to arm itself when the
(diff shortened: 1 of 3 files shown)


robfrank:
### Review round 3 - what was applied and what was declined

**Applied (f9ccc6c)**

- **`ExecutionKey` did not identify an execution.** CodeRabbit is right and this was a real hole in the guard itself: an execution id is unique only inside one `<plugin>` block, so two profiles can each declare `<execution><id>default</id>` carrying a different half of the filter. Keyed by `(pom, plugin, id)` they merged into one execution that looked properly pinned while each on its own was narrowable. The key now carries the plugin block's position in the document, plus a scope label so the message names the profile. `executionsSharingAnIdAcrossProfilesAreJudgedSeparately` pins it and goes red if the identity is collapsed back.

**Declined, with reasons**

- **"An execution that omits *both* parameters escapes the rule."** Correct, and deliberate. Such an execution is not half of a partition; it is an ordinary named execution set up for its own `<includes>` or `reuseForks`, and nothing in the pom says otherwise. Demanding both parameters of every named surefire/failsafe execution would fail on those and would invent a partition the pom does not describe. Declaring one tag parameter is the signal that the other must be pinned too. Rather than leave that implicit, it is now pinned as a decision: `anExecutionWithNoTagFilterIsNotAPartition`.
- **Cache the scan in a `@BeforeAll` static.** The whole class is 0.8 s including four repo walks; a cached static buys nothing measurable and adds shared mutable state to a test whose entire job is to be obviously correct. Flagged by the reviewer as a nit, treated as one.
- **Drop `hasSizeGreaterThan(20)`.** Kept. It is not there to pin the module count (28 today, so the margin is large) but to catch a walk that returns almost nothing while still finding the three named poms - a `PRUNED` entry that swallows most of the tree, say. A consolidation that removes eight modules and turns it red is a signal worth reading, not noise.
- **The omitted-parameter message names `${groups}`/`${excludedGroups}` specifically.** Exact today, because that is what this repo's plugin-wide defaults are called. A default under another property name would keep the verdict right and only the property named in the message wrong, so this is documented on `unpinnedPartitionParameters` rather than coded around.

**Note on the full-module run.** `./mvnw -o -pl ha-raft test` currently dies in a fork start (`The forked VM terminated without properly saying goodbye`, 0 tests, at `LeaveClusterTest`). A control run of that class against **unmodified `main`** reproduces it identically, and a concurrent `ha-raft` run was live in another worktree at the time; this PR touches one test class and two markdown/comment blocks, so it cannot be the cause. The module was green at 991 tests earlier on this branch, and `PomTagFilterContractTest` is 13/13 green.

robfrank:
Closing: this work is being dropped.

PR #6917, opened 2026-08-30
fix(#4416): ship an oversized TimeSeries sealed store in slices instead of skipping the shard

### LINKED ISSUE #4416: TimeSeries HA: a shard stops sealing once its sealed store exceeds the inline replication cap
## Context

This is a follow up to #4382, where TimeSeries data and schema replication under Raft HA was implemented: leader only compaction, the mutable bucket clear replicated, and sealed blocks shipped to followers inside a single SCHEMA_ENTRY.

To keep every Raft entry within the transport size limit (`MAX_ENTRY_BYTES`, 64MB), compaction ships the rewritten `.ts.sealed` file inline. A safety valve, `GlobalConfiguration.HA_TS_MAX_SEALED_INLINE_SIZE` (default 48MB), makes the leader skip compacting a shard when the projected sealed store size would exceed the cap.

## Current behavior

When a shard's projected sealed store size exceeds `HA_TS_MAX_SEALED_INLINE_SIZE`, `TimeSeriesShard.compactInternal` skips compaction entirely for that shard:

* no sealing happens
* the samples stay in the mutable bucket, which IS fully replicated through the normal WAL path
* a throttled WARNING is logged

This keeps the cluster correct and consistent: no divergence and no data loss. The trade off is that the affected shard never seals again. Its mutable bucket keeps growing without bound and the data stays uncompressed, since the sealed columnar format is never produced for the overflow.

## Impact

* correctness and replication are fine. This is a scalability and storage efficiency limit, not a data integrity bug.
* a workload that accumulates more than roughly 48MB of compressed sealed data per shard will, under HA, stop benefiting from columnar compression and from sealed store retention and downsampling for that shard.
* the cap is configurable, so raising it buys headroom, but it cannot exceed the Raft message size limit (64MB) with room left for the schema JSON and the mutable clear WAL.

## Proposed solutions (any one would lift the ceiling)

1. Incremental block shipping: ship only the new sealed blocks appended by a compaction (a delta) instead of the whole file. This requires an append friendly sealed block layout so a follower can append without rewriting prior offsets, plus a way to guarantee the follower preimage matches.
2. Out of band fetch: ship a small pointer in the SCHEMA_ENTRY and have the follower pull the sealed file over the existing snapshot HTTP channel, so the Raft entry stays tiny regardless of sealed size.
3. Chunked transfer: split a large sealed file across several ordered Raft entries that the follower reassembles before swapping it in.

Option 2 reuses the snapshot transport that already ships `.ts.sealed`, and is probably the least invasive.

## References

* parent issue #4382
* relevant code: `TimeSeriesShard.compactInternal` (the skip check), `TimeSeriesSealedStore.readWholeSealedFile` and `installSealedFileBytes`, `RaftReplicatedDatabase.runWithCompactionReplication`, `RaftLogEntryCodec` (the sealed blob section), and `GlobalConfiguration.HA_TS_MAX_SEALED_INLINE_SIZE`

robfrank:
Closes #4416

## The problem

A replicated TimeSeries compaction ships the rewritten `.ts.sealed` file to the followers inside the same Raft
entry as the mutable-bucket clear, so the two land atomically and a query never sees a cleared bucket beside a
stale sealed store. Because the whole file rode ONE entry, `arcadedb.ha.tsMaxSealedInlineSize` (clamped to
`min(arcadedb.ha.grpcMessageSizeMax, arcadedb.ha.appendBufferSize)`, per #4743) was a ceiling on the sealed STORE
rather than merely on the entry: `TimeSeriesShard.compactInternal` skipped a shard whose projected store crossed
it, and that skip was **permanent** - the store it refused to ship only grows, so the projection never came back
under the cap and the shard never sealed again. Correct and fully replicated (the samples stay in the mutable
bucket) but never compressed, never retained, never downsampled, and the bucket itself unbounded.

## The fix: option 3 from the issue, slicing

A sealed store above one entry is cut into ordered slices of at most one entry's worth. Every slice but the last
ships as a **delivery-only** entry (`moreChunksFollow`, no schema JSON, no file maps, no WAL) that the follower
merely STAGES on disk; the final slice rides the **publishing** entry, so the install still happens in the same
entry as the clear WAL and the atomicity that made the whole-file design safe is unchanged. This is the
ordered-prefix contract #4743 built for oversized schema entries and #6136 reused for schema-WAL instalments,
applied to a file rather than to a WAL stream, so followers need no new state beyond the staging file.

Why not the issue's other two options: incremental block shipping (option 1) needs an append-friendly sealed
layout, and compaction currently REWRITES the file (`writeTempCompactionFile` re-serializes every retained block
in minTimestamp order), so a delta is not well defined without changing the format. Out-of-band HTTP fetch
(option 2) puts a blocking download on the single Raft apply thread and needs versioned staging on the leader
because the file it points at can be replaced before the follower pulls it. Slicing is format-agnostic, needs no
new transport, and works for the retention/downsampling maintenance path too - which ships whole sealed files
today with no size guard at all.

### Where the ceiling now sits

`arcadedb.ha.tsMaxSealedInlineSize` keeps its name and its 48MB default but now bounds one ENTRY. The per-shard
sealed ceiling becomes `GlobalConfiguration.maxReplicatedSealedStoreSize` = that value (still transport-clamped)
x 512 slices, capped at `Integer.MAX_VALUE` because the leader still materializes the file as one array before
slicing it. With stock settings that moves the ceiling from **48MB to ~2GB per shard**.

Two deliberate continuities with the old behaviour:
- a shard above the NEW ceiling is still skipped, with a warning that now names what to raise;
- a cap configured so small that a slice cannot carry even its own framing leaves the ceiling exactly at the
  per-entry cap, so a deliberately tiny cap behaves as it always did. `RaftTimeSeriesOversizedSealedIT`, which
  pins the skip with a 10-byte cap, is unmodified and still passes for that reason.

### Slice budget

`replicatedSealedChunkBudget` = per-entry cap - 4KB framing - **cap/128**. The proportional term is the one that
matters: a slice of already-compressed sealed blocks can come off the LZ4 encoder LARGER than it went in, by up
to `n/255 + 16` bytes, and at a 4MB cap that expansion alone (~16KB) is four times a fixed framing reserve. An
entry over the cap is not a slow path - Ratis rejects it and the leader steps down, repeatedly (#4743).
`theBudgetReservesEnoughForASliceThatCompressionMakesBIGGER` pins it, and it FAILED against the fixed-reserve
version before the proportional term was added.

### Follower reassembly

Staging is a `<sealed file>.parts` file next to the target, not a buffer in the state machine: a follower
persists its applied index between entries, so one that restarts mid-sequence will never be sent the earlier
slices again - a file survives that restart, a field does not - and a multi-gigabyte store never has to fit in
heap on the way in. A slice at `offset == 0` TRUNCATES whatever is staged, which makes a sequence abandoned by a
leader that died mid-shipment cost nothing; any other slice must find the staging file exactly `offset` bytes
long. The reassembled file is verified against the length and CRC32 the leader recorded for the WHOLE image
before installing - per-slice CRCs (checked by the decoder) only prove each piece survived the wire. A sequence
that does not line up, or a reassembly that does not match, refuses to install and throws `ReplicationException`,
which `applyWithRetry` turns into a targeted snapshot resync rather than putting this node on a sealed store no
other node has. The target path is derived from THIS node's schema, never from the file name in the payload.

### Wire format

`TsSealedChunk` travels in a new trailing section written AFTER the `moreChunksFollow` flag (which this encoder
always emits, so the flag can never be confused with the section's first byte). An entry with nothing to slice is
**byte-identical** to what the previous codec produced - asserted on the bytes, not on the decode.

**Rolling-upgrade note:** a node on an older codec stops right after the flag, so it installs nothing for a
SLICED store while still applying the mutable-bucket clear, and serves that shard from its previous sealed image
until it is upgraded and resynced. That is the same exposure the sealed-blob section itself has had since #4382,
for the same reason (a trailing section an old decoder cannot see). A leader only produces slices once a store
outgrows one entry, so a cluster whose stores fit inline is not exposed at all. Upgrade the followers first.
This is stated in the codec javadoc and in the release notes.

## Test plan

- [ ] `Issue4416SlicedSealedStoreTest` (13 tests, ha-raft): slicer invariants (contiguous offsets, one `last`,
      whole-file CRC, reassembly equals the original, no empty trailing slice), codec round trip, delivery-only
      slice keeps the continuation flag, **byte-identity with the previous format when nothing is sliced**,
      per-slice CRC corruption refused, `splitSchemaEntry` keeps the final slices on the LAST chunk, and the
      configuration arithmetic including the compression headroom and the too-small-cap fallback.
- [ ] `Issue4416SlicedSealedApplyTest` (7 tests, ha-raft): follower apply against a real `LocalDatabase` -
      staged-then-installed byte for byte, sliced install == whole-file install, a slice that does not line up
      asks for a resync, a restarted sequence discards the abandoned staging file, a reassembly with the wrong
      CRC or the wrong length is refused and the node keeps the store it had.
- [ ] `Issue4416SlicedSealedStoreIT` (3-node, `@Tag("slow")`): with the cap set just above the framing, a shard
      that would previously have been skipped seals, the store travels as a SEQUENCE (asserted on the new
      `getSealedStoreChunksShipped()` counter, `> 1`), every node ends with the leader's sealed file and an empty
      mutable bucket, and a second compaction cycle repeats it over a store that already has content.
- [ ] Unmodified and green: `RaftTimeSeriesOversizedSealedIT` (the skip still fires on a tiny cap),
      `RaftTimeSeriesReplication3NodesIT`, `Issue4458TsWalVersionGapIT`, `Issue5443SplitCompactionEntryIT`,
      `RaftLogEntryCodecTest`, `RaftTransactionBrokerTest`, `Issue4743OversizedRaftEntryTest`,
      `Issue6136SchemaWalInstalmentTest`, `Issue6839TsSealedBlobRecoveryTest`, `TimeSeriesSealedStoreTest`.
- [ ] Full `ha-raft` unit lane: 1001 tests, 0 failures attributable to this change (one `SnapshotSymlinkTest`
      flake under a concurrent run, green in isolation).
- [ ] Full `engine` timeseries suite: 394 tests, 0 failures.

## Known limits, deliberately left

- The leader still reads the whole sealed file into one `byte[]` before slicing it, which is why the ceiling is
  clamped to 2GB. Streaming the file into slices is a separate change and is stated as such in the javadoc.
- Every slice is a quorum round trip taken inside the compaction's replication session, so a very large store
  trades a permanent skip for a burst of entries. `MAX_REPLICATED_SEALED_CHUNKS` (512) bounds that burst, and a
  store needing more slices than that is shipped anyway - withholding it would diverge the followers - but
  reported at WARNING.

## Summary by CodeRabbit

* **New Features**
  * TimeSeries sealed stores larger than a single Raft entry can now replicate through ordered slices.
  * Slices are validated and installed atomically, with automatic resynchronization when integrity checks fail.
  * Sealed stores up to 2 GB are supported, subject to configured slice limits.
  * Older followers retain their existing sealed data until upgraded and resynced.

* **Documentation**
  * Added release notes describing sliced sealed-store replication and compatibility behavior.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -4039,3 +4039,45 @@ tag have to agree, and neither may be a SNAPSHOT - and fails if the manifest fal
 POM, which is what noticing this in the first place required a human to do.
 
 [#6838](https://github.com/ArcadeData/arcadedb/issues/6838)
+
+## TimeSeries HA: a shard no longer stops sealing when its sealed store outgrows one Raft entry (#4416)
+
+Compaction of a replicated TimeSeries shard ships the rewritten `.ts.sealed` file to the followers inside the same
+Raft entry as the mutable-bucket clear, so the two land atomically and a query never sees a cleared bucket beside a
+stale sealed store. Because the whole file rode one entry, `arcadedb.ha.tsMaxSealedInlineSize` (clamped down to
+`min(arcadedb.ha.grpcMessageSizeMax, arcadedb.ha.appendBufferSize)`) was a ceiling on the sealed STORE and not
+merely on the entry: a shard that crossed it was skipped, and the skip was permanent. The store it refused to ship
+only grows, so the projected size never came back under the cap and that shard never sealed again - correct and
+fully replicated, since the samples stay in the mutable bucket, but never compressed, never retained, never
+downsampled, and the bucket itself unbounded.
+
+A sealed store above one entry is now SLICED. The leader cuts it into ordered slices of at most one entry's worth
+and ships every slice but the last as a delivery-only entry, which the follower stages on disk and nothing more;
+the final slice rides the publishing entry, so the install still happens in the same entry as the clear WAL and
+the atomicity that made the whole-file design safe is unchanged. This is the ordered-prefix contract #4743 built
+for oversized schema entries, applied to a file rather than to a WAL stream, so followers need no new state beyond
+the staging file.
+
+The staging file is on disk rather than in the state machine deliberately: a follower persists its applied index
+between entries, so one that restarts mid-sequence will never be sent the earlier slices again, and a file
+survives that restart. A slice at offset 0 truncates whatever is staged, which makes a sequence abandoned by a
+leader that died mid-shipment cost nothing; any other slice must find the staging file exactly that long. The
+reassembled file is checked against the length and CRC32 the leader recorded for the WHOLE image before it is
+installed - per-slice CRCs only prove each piece survived the wire - and a sequence that does not line up, or a
+reassembly that does not match, refuses to install and asks for a targeted snapshot resync instead of putting the
+node on a sealed store no other node has.
+
+`arcadedb.ha.tsMaxSealedInlineSize` keeps its name and its default (48MB) but now bounds one ENTRY rather than the
+store. The per-shard sealed ceiling becomes that value - still clamped to the transport limit - times 512 slices,
+capped at 2GB because the leader still materializes the file as a single array before slicing it. With stock
+settings that moves the ceiling from 48MB to ~2GB per shard. A shard above the new ceiling is still skipped, with a
+warning that now names what to raise; and a cap configured so small that a slice cannot even carry its own framing
+leaves the ceiling exactly where it was, so a deliberately tiny cap behaves as it always did.
+
+Rolling-upgrade note: the slices travel in a trailing section of `SCHEMA_ENTRY`, which a node running an older
+codec stops before, exactly as the sealed-blob section itself has done since #4382. Such a node installs nothing
+for a sliced store while still applying the mutable-bucket clear, and serves that shard from its previous sealed
+image until it is upgraded and resynced. A leader only produces slices once a store outgrows one entry, so a
+cluster whose stores fit inline is not exposed at all. Upgrade the followers first.
+
+[#4416](https://github.com/ArcadeData/arcadedb/issues/4416)
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2490,7 +2490,14 @@ public enum GlobalConfiguration {
    */
   public static long maxReplicatedSealedStoreSize(final ContextConfiguration configuration) {
     final long perEntry = maxReplicatedSealedEntrySize(configuration);
-    final long sliced = replicatedSealedChunkBudget(configuration) * MAX_REPLICATED_SEALED_CHUNKS;
+    // Saturating rather than wrapping. It cannot currently change an answer - a budget large enough to overflow
+    // this product needs a per-entry cap of ~18 petabytes, and THAT alone is already far above the
+    // Integer.MAX_VALUE clamp below, which is what the max() then returns either way - but a ceiling that is
+    // correct only because two unrelated bounds happen to cover for each other is one edit away from not being.
+    final long budget = replicatedSealedChunkBudget(configuration);
+    final long sliced = budget >= Long.MAX_VALUE / MAX_REPLICATED_SEALED_CHUNKS
+        ? Long.MAX_VALUE
+        : budget * MAX_REPLICATED_SEALED_CHUNKS;
     return Math.min(Integer.MAX_VALUE, Math.max(perEntry, sliced));
   }
 
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -1941,11 +1941,15 @@ public enum GlobalConfiguration {
 
   HA_TS_MAX_SEALED_INLINE_SIZE("arcadedb.ha.tsMaxSealedInlineSize", SCOPE.SERVER,
       """
-      Maximum size in bytes of a TimeSeries sealed-store file that may be shipped inline inside a single \
-      Raft SCHEMA_ENTRY during compaction. When the projected sealed-store size would exceed this cap, the \
-      leader skips compacting that shard (data stays in the fully replicated mutable bucket) instead of \
-      producing an entry too large for the Raft transport. Always clamped down to the real per-entry ceiling, \
-      min(arcadedb.ha.grpcMessageSizeMax, arcadedb.ha.appendBufferSize), so a value above that has no effect.""",
+      Maximum size in bytes of TimeSeries sealed-store content a SINGLE Raft entry may carry during \
+      compaction. A sealed store up to this size ships inline in one SCHEMA_ENTRY; a larger one is sliced \
+      and shipped as an ordered sequence of entries of at most this size each, which the follower stages on \
+      disk and installs atomically with the mutable-bucket clear when the last slice lands (issue #4416). \
+      Always clamped down to the real per-entry ceiling, min(arcadedb.ha.grpcMessageSizeMax, \
+      arcadedb.ha.appendBufferSize), so a value above that has no effect. It therefore no longer bounds the \
+      sealed store itself: what does is this value times MAX_REPLICATED_SEALED_CHUNKS - see \
+      GlobalConfiguration.maxReplicatedSealedStoreSize - and a shard whose projected sealed store exceeds \
+      THAT is still skipped, leaving its samples in the fully replicated mutable bucket.""",
       Long.class, 48 * 1024 * 1024L),
 
   HA_SNAPSHOT_WATCHDOG_TIMEOUT("arcadedb.ha.snapshotWatchdogTimeout", SCOPE.SERVER,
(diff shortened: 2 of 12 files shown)


robfrank:
Thanks - all three looked at, two applied as described plus one correction.

**1. Maintenance path has no ceiling pre-check.** Intentional, and now stated at the call site (`TimeSeriesEngine.runSealedMaintenanceReplicated`) rather than only on the callee. The asymmetry has a reason: compaction GROWS the sealed store, so it can decline to create one it cannot replicate; retention and downsampling only ever shrink or hold one, so they cannot push a shard over a ceiling it was not already over. A pre-check there could only refuse to ship a store the leader has ALREADY rewritten, which trades a burst of entries for a follower left on an image no node still has.

**2. `warnOversizedSealedSkip` message.** Confirmed and fixed - it really did render as `(arcadedb.ha.tsMaxSealedInlineSize x 512 slices...)`. It now states the slice count and the per-slice byte budget as numbers and names the setting separately.

**3. Overflow in `maxReplicatedSealedStoreSize`.** The arithmetic point stands, but the stated consequence does not: it cannot produce a too-small ceiling. `budget * 512` can only overflow when `budget` exceeds `Long.MAX_VALUE / 512`, which needs a per-entry cap of ~18 petabytes - and at that point `perEntry` ALONE is far above the `Integer.MAX_VALUE` clamp, so `min(Integer.MAX_VALUE, max(perEntry, sliced))` returns `Integer.MAX_VALUE`, the correct clamped answer, whether or not the product wrapped. Checked directly:

```
budget=4575657221408419840  sliced(overflowed)=-2097152
max(perEntry,sliced)=4611686018427387903
ceiling=2147483647  == Integer.MAX_VALUE
```

Saturated it anyway, with a comment saying exactly that - a ceiling that is right only because two unrelated bounds happen to cover for each other is one edit away from not being - but it is hygiene rather than a fix, and no test was added because none could fail.

Also from the previous round: the `offset == 0` truncation now goes through the open handle (`setLength(0)`) instead of depending on `File.delete()`, with `theFirstSliceDiscardsALongerLeftover` pinning it - it fails with "starts at offset 4096 but this node has staged 222185 bytes" when the truncation is removed.

Re-verified after these edits: engine timeseries suite 394/394, `Issue4416SlicedSealedStoreTest` 13/13, `Issue4416SlicedSealedApplyTest` 8/8, `Issue4416SlicedSealedStoreIT` and the unmodified `RaftTimeSeriesOversizedSealedIT` both green.

robfrank:
Round 3 - items 1 and 3 applied, item 2 confirmed as intended.

**1. `checkProducedPayloadLength` remediation text.** Right, and it was actively misleading at the sealed call sites: "reduce the batch size - fewer rows per GraphBatch / SQL transaction" is correct for a transaction's WAL and is no advice at all for a sealed store, which does not shrink by batching anything. The remediation is now a per-call-site argument, and both sealed sites (whole blob and slice) name the settings that actually move the number: `arcadedb.ha.tsMaxSealedInlineSize` and the type's retention.

**2. Sequential quorum round trips under the session.** Intended, and stated. Worth adding to your note: the loop runs AFTER the compaction callback returns, so `TimeSeriesShard.compactionLock` is already released and appends are not blocked for the burst - the cost is the replication session staying open, not a write stall. `getSealedStoreChunksShipped()` is exposed per database so a deployment sizing these settings can see how many slices a cycle actually costs.

**3. No sign validation on `offset`/`fileLength`.** Real, and fixed at the decoder rather than at the write, which is where every other malformed-payload check in this codec lives and the only place that sees the field before anything acts on it. `checkSealedSliceGeometry` rejects a negative offset or file length, a slice that overruns the declared file, and a final slice that stops short of it - the last one being an invariant every producer satisfies by construction, so a payload that violates it is not a payload this codec wrote. That keeps "a sequence that does not line up asks for a resync, it never crashes" true for a malformed field as well as for a missing slice.

Two tests: `aSliceWhoseGeometryCannotDescribeAFileIsRefused` walks the five impossible shapes, and `everySliceTheSlicerProducesPassesTheGeometryCheck` runs the real slicer's output through the same guard so it cannot be refusing everything. The guard also caught an arbitrary file length in my own round-trip fixture, which now describes a real file.

Re-verified: `Issue4416SlicedSealedStoreTest` 15/15, `Issue4416SlicedSealedApplyTest` 8/8, `RaftLogEntryCodecTest` 35/35, `Issue4743OversizedRaftEntryTest`, `Issue5933CompressibleTxEntrySizeTest`, `Issue6136SchemaWalInstalmentTest`, `RaftTransactionBrokerTest` all green (85 total), and `Issue4416SlicedSealedStoreIT` green.

robfrank:
Round 4 - items 2 and 3 applied as documentation, item 1 deferred with a reason. This was the last automated cycle (`--max-cycles=4`); the branch is handed over from here.

**2. fsync only on the final slice.** Intended, and now spelled out where the offset==0 reasoning already is. A power loss between applying a slice and the OS flushing it can leave a staging file shorter than this node's persisted applied index implies, and that is precisely what the offset and length/CRC checks are for: the sequence is refused and the database resyncs rather than installing a truncated store. An fsync per slice would buy a faster recovery from an event that already costs a restart, at the price of one flush per entry on the single Raft apply thread for every sliced compaction. The final slice is synced because after it the file is moved into place and nothing checks it again.

**3. Best-effort delete in the failure branches.** Kept, and the difference from the first-slice case is now stated on the method rather than left to be inferred. Nothing downstream depends on this delete succeeding: the next sequence's `offset == 0` slice truncates through its own write handle either way, so a failure costs disk until the next compaction and never correctness. That is exactly what was NOT true of the first-slice delete, where a silently skipped one left a longer sequence's tail under a shorter one. The warning now says that instead of implying the next sequence merely overwrites the file. Removing the calls outright would leave a rejected multi-GB reassembly on disk until the next compaction for no gain.

**1. A distinct counter for oversized maintenance-path bursts - DEFERRED, not done.** The analysis is right and the tradeoff is the intended one (refusing would leave followers on an image no node still has). Not adding a second counter in this PR: it is a new observability surface rather than part of the fix, the WARNING in `sliceSealedBlob` already names the file, the slice count and the settings to raise, and `getSealedStoreChunksShipped()` already carries the volume per database. A metric that separates maintenance-path slices from compaction-path slices is worth having and belongs with the rest of the HA gauge wiring - flagging it here for a follow-up rather than bolting it on at the end of a review loop.

**Nit - did `RaftTimeSeriesOversizedSealedIT` actually run?** Yes, unmodified and green locally on this branch (`Tests run: 1, Failures: 0`), which is the arm that matters most for compatibility: with a 10-byte cap the shard is still skipped and its samples still stay in the replicated mutable bucket, because a cap too small to hold one slice's framing leaves the ceiling at the per-entry cap exactly as before.

Final local verification on `f213877f52`: engine timeseries suite 394/394; ha-raft unit lane 1001 tests with no failure attributable to this change; `Issue4416SlicedSealedStoreTest` 15/15, `Issue4416SlicedSealedApplyTest` 8/8; ITs `Issue4416SlicedSealedStoreIT`, `RaftTimeSeriesOversizedSealedIT`, `RaftTimeSeriesReplication3NodesIT`, `Issue4458TsWalVersionGapIT`, `Issue5443SplitCompactionEntryIT` all green. Merge is the maintainer's call.

robfrank:
Worked through this against the code. Two items were correct, three rest on premises that do not hold, and the manual re-read the review asked for turned up two real defects. Pushed as 26cfa9b.

## The main concern: the premise came from a line this PR got wrong

> per the PR's own commit message this loop runs "while the compaction holds the database write lock"

That phrase is not in a commit message or the PR body - it is in a javadoc **this PR added**, on `MAX_REPLICATED_SEALED_CHUNKS`, and it is wrong. Fixed.

What actually happens: the slicing loop runs in `runWithCompactionReplication` **after** `invokeCompaction(compaction)` returns. `TimeSeriesShard.compactionLock` is released inside `compact()`'s callback, and the per-shard write lock `TimeSeriesEngine.runSealedMaintenanceReplicated` takes is released in its own per-shard `finally`, also inside the callback. No compaction lock is held for any of the round trips.

What *is* held is the FileManager recording session, and its cost is narrower and bounded: other commits **on the leader** call `waitForActiveRecordingSession()` before dispatching a TX_ENTRY (the #4083 ordering guard), which polls and then **gives up after `HA_QUORUM_TIMEOUT`**. So a long burst degrades leader commit latency up to that timeout and then lets commits through on the racy path. It does not stall writes for the burst's duration, and it does not touch reads or follower traffic. Real, worth documenting - which it now is - but not the availability incident the review describes.

The same javadoc said "the default 4MB per-entry ceiling". The default is **32MB**: `min(tsMaxSealedInlineSize 48MB, min(grpcMessageSizeMax 128MB, appendBufferSize 32MB))`. And ~2GB comes from the `Integer.MAX_VALUE` clamp in `maxReplicatedSealedStoreSize`, not from 512 x the cap, which at 32MB would be ~16GB. Consequence worth stating plainly: **at stock settings ~68 slices already reach the ceiling and nothing can ask for 512**. The "tens of thousands of round trips" case needs `appendBufferSize` lowered to ~64KB *and* a ~2GB legacy store; a more plausible lowered 1MB buffer gives ~2,000. Corrected, along with a test comment that repeated the 4MB figure.

### Declining the hard cap

> consider making the "ship anyway" behavior a hard cap instead (skip + warn, same as the existing oversized-skip path)

No. By the time the slicer runs the leader has already rewritten and swapped its sealed file - `runSealedMaintenanceReplicated` records the bytes *after* `work.apply(shard)`. Withholding them leaves every follower on an image no node holds any more. That trades a bounded latency spike for silent divergence, which is strictly worse. The asymmetry with `compactInternal`'s pre-check is deliberate for the same reason: compaction *grows* the store so it can decline in advance, retention/downsampling only shrink one so a pre-check there could only refuse to ship something already rewritten.

`aStoreNeedingMoreSlicesThanExpectedIsShippedANYWAYRatherThanRefused` now pins this, so the tempting edit fails a test instead of shipping.

### Taking the metric

> at least a metric/counter for "slices shipped in a single sequence" beyond the expected bound

Correct that `getSealedStoreChunksShipped()` cannot answer it - a thousand slices reads identically whether it was one pathological store or a thousand healthy one-slice ones. Added `getSealedStoreMaxSliceSequence()`, updated per store rather than per session.

## Untested branch: correct

`count > MAX_REPLICATED_SEALED_CHUNKS` was indeed uncovered. Now pinned as above (full coverage, contiguous offsets, exactly one `last`, exact reassembly), disarmed and observed red before being restored.

## Minor items

**Two independent config reads.** Both resolve to the *same object*: `RaftReplicatedDatabase.getConfiguration()` returns `proxied.getConfiguration()`, and `TimeSeriesShard`'s `database` is that wrapper. There is no window in which they read different snapshots. They also compute deliberately different quantities - a ceiling (`budget x 512`, clamped) versus a per-slice budget - so "threading the already-computed cap through" would thread the wrong number. No change.

**The dense format string.** Doing the re-read found two real defects, so this one earned its place:

1. `warnOversizedSealedSkip` carried `clearCompactionFlagBestEffort`'s javadoc verbatim ("Best-effort: clear the compaction-in-progress flag after a non-crash error").
2. The message asserted the ceiling is "512 slices of B bytes each" unconditionally, which is **arithmetically false in two of three regimes**. At stock settings it prints `exceeds the 2147483647 bytes ... which is 512 slices of 33292288 bytes each` - and 512 x 33,292,288 = 17,045,651,456, not 2,147,483,647. When the budget is 0 it claims "512 slices of 0 bytes each" while the real ceiling is one entry. An operator who checks the multiplication and gets a different number stops trusting the message. It now describes the regime that actually produced the ceiling, and `atStockSettingsTheCeilingComesFromTheClampNotFromTheSliceCount` pins that regime.

**Leader-side mid-loop failure test.** Agreed it is prose-only. Not adding it here: the recovery it would assert (version-gap resync to snapshot) is not new code in this PR and belongs to the existing resync path, so a test for it is a test of that path, not of slicing. Leaving it as a follow-up rather than claiming coverage this PR does not have.

Verification: ha-raft unit lane 1007/1007, engine timeseries 544/544.

PR #6918, opened 2026-08-30
feat(#3890): add optional mTLS to the Raft gRPC transport

### LINKED ISSUE #3890: Raft gRPC: add support for mTLS for encryption and peer identity
## Context

The Ratis gRPC transport used for Raft log replication, leader election, and snapshot chunk transfer currently has no peer authentication and no in-transit encryption. Today the only inter-node protections are:

- `X-ArcadeDB-Cluster-Token` header on the HTTP side channels (snapshot download, database verify)
- Network-level isolation (K8s NetworkPolicy, private subnet, VPN) called out as the recommended hardening step in `RaftHAServer.startService()` warnings

The Raft gRPC port remains an unauthenticated back door: any host that can reach the port can open a gRPC stream to the Ratis server and inject log entries.

A follow-up change will add a `ServerTransportFilter`-based peer-address allowlist to close the "random host on the network knows the port" case out of the box. That mitigation is IP-based and is defeated by IP spoofing on a flat L2 network or by a compromised peer. Production deployments need cryptographic peer identity and in-transit encryption.

## Goal

Add optional mTLS to the Raft gRPC transport using the `GrpcTlsConfig` mechanism already supported by Apache Ratis (`GrpcConfigKeys.TLS.setConf`, `GrpcConfigKeys.Admin.setTlsConf`, `GrpcConfigKeys.Client.setTlsConf`, `GrpcConfigKeys.Server.setTlsConf`).

When enabled:
- Every Raft gRPC connection between nodes negotiates TLS with mutual client-certificate authentication.
- Peer identity is bound to the certificate (CN/SAN), verified against a shared cluster CA.
- All AppendEntries / InstallSnapshot / RequestVote traffic is encrypted in transit.
- Unauthenticated or non-CA-signed peers are rejected at the TLS handshake.

## Proposed configuration surface

New settings under `arcadedb.ha.tls.*`:

- `arcadedb.ha.tls.enabled` - boolean, default `false` (preserve zero-config dev/test)
- `arcadedb.ha.tls.certChainFile` - PEM file with this node's certificate (and intermediates if any)
- `arcadedb.ha.tls.privateKeyFile` - PEM file with this node's private key
- `arcadedb.ha.tls.trustCertCollectionFile` - PEM file with the cluster CA certificate(s)
- `arcadedb.ha.tls.mutualAuth` - boolean, default `true`

Wiring in `RaftPropertiesBuilder` / `RaftHAServer`:

- When `arcadedb.ha.tls.enabled=true`, build a `GrpcTlsConfig` from the configured files and install it via `GrpcConfigKeys.TLS.setConf(parameters, tlsConf)` on the `Parameters` object passed to `RaftServer.Builder.setParameters(...)`.
- Apply the same `GrpcTlsConfig` on the `RaftClient.Builder.setParameters(...)` used by `RaftHAServer.buildRaftClient()` so the leader's self-client also speaks TLS.
- Fail-fast on startup if TLS is enabled but any of the cert/key/trust paths are missing or unreadable, with a clear error message.

## Documentation deliverables

- `docs/arcadedb-ha-*.md` section on production hardening, including:
  - `openssl`-based recipe for a self-signed cluster CA and per-node cert (already drafted in internal notes)
  - Cert-manager recipe for Kubernetes StatefulSet deployments (SAN must match the pod's stable DNS name)
  - Vault PKI / internal-PKI guidance for orgs that already run a CA
  - Operational notes on rotation and expiry
- Release-notes callout: mTLS is the supported way to secure the Raft gRPC port; the peer-address allowlist is a best-effort default, not a substitute.

## Test plan

- Unit test: `RaftPropertiesBuilder` builds a correct `GrpcTlsConfig` from the new settings and attaches it to `Parameters`.
- Integration test: 3-node `BaseRaftHATest` subclass that boots with mTLS enabled using a test CA under `src/test/resources`; verifies a full leader election + transaction commit + follower catch-up round-trip.
- Negative integration test: a 4th node with a cert signed by a different CA attempts to join; handshake is rejected and no log entries are accepted from it.
- Snapshot HTTP handler: confirm the existing `X-ArcadeDB-Cluster-Token` check on `SnapshotHttpHandler` is preserved (TLS on gRPC does not replace the HTTP token).

## Out of scope for this issue

- Cert rotation without restart (can be a follow-up once the static-config path lands)
- Integrating with an external KMS or Vault agent for cert retrieval (leave it to ops; document the pattern)

robfrank:
Closes #3890

## Summary

The Ratis gRPC transport that carries Raft log replication, leader election and the admin/client calls had
no peer authentication and no in-transit encryption: any host able to reach the Raft port could open a gRPC
stream to the Ratis server and inject log entries. The `PeerAddressAllowlistFilter` added earlier narrows
that to hosts resolving from `arcadedb.ha.serverList`, but an IP address is not an identity - it is defeated
by spoofing on a flat L2 network or by a compromised peer.

This adds **optional mTLS** on that transport. Five new `arcadedb.ha.tls.*` settings build an Apache Ratis
`GrpcTlsConfig` from PEM files and install it via `GrpcConfigKeys.TLS.setConf(...)` on the `Parameters`
object. Defaults are unchanged: `arcadedb.ha.tls.enabled=false`, so a development or test cluster still
starts with no configuration at all, and the plaintext code path is byte-for-byte what it was.

| Setting | Type | Default |
|---|---|---|
| `arcadedb.ha.tls.enabled` | boolean | `false` |
| `arcadedb.ha.tls.certChainFile` | string | `""` |
| `arcadedb.ha.tls.privateKeyFile` | string | `""` |
| `arcadedb.ha.tls.trustCertCollectionFile` | string | `""` |
| `arcadedb.ha.tls.mutualAuth` | boolean | `true` |

## Design notes

**One `Parameters` entry covers all three endpoints.** Ratis's `GrpcFactory` reads `GrpcConfigKeys.TLS.conf`
as the *default* for the admin, client and server-to-server SSL contexts alike (`GrpcFactory:100-103`), so a
single `TLS.setConf` call secures AppendEntries, InstallSnapshot and RequestVote without three separate
`Admin`/`Client`/`Server` calls. This holds through our `FixedGrpcFactory` subclass, which passes
`Parameters` straight to `super`.

**The leader's self-client had to be wired too, and this is the subtle half.** `RaftHAServer.buildRaftClient()`
was passing a fresh, empty `new Parameters()`, so with TLS on the server the leader's own client would still
dial the Raft port in plaintext and every group-commit would time out. The built `Parameters` are now threaded
through all three `buildRaftClient` call sites - initial start, health-monitor recovery restart, and
`refreshRaftClient()` on leader change - the last one via a `raftParameters` field, since that path has no
local handle on them. Confirmed load-bearing: reverting just that one line turns
`clusterReplicatesOverMutualTls` red with `ReplicationDispatchedTimeoutException` (see "Proving the tests can
fail" below).

**Fail-fast validation.** `RaftPropertiesBuilder.buildTlsConfig()` checks each of the three paths is set,
exists, is a regular file, and is readable, and throws `ConfigurationException` naming the offending setting.
Without it an unmounted Kubernetes secret volume or a typo surfaces much later as an opaque handshake failure
on every peer connection instead of a startup error the operator can act on.

**`raftParameters` is published on `buildParameters`'s last line**, once both the TLS configuration and the
peer-allowlist customizer are installed. Two things depend on that moment, and each was got wrong once during
review: publishing before `applyTls` left a plaintext `Parameters` in the field whenever a cert path was
unusable, and publishing between the two installs left a window where a concurrent reader could see the object
without its customizer. The allowlist install is now its own method so the rule is literal - nothing above
publishes, nothing below adds.

**The Kubernetes auto-join probe needed the same treatment, and it is the second half of the same bug.**
`KubernetesAutoJoin` opens its *own* short-lived `RaftClient` for the membership probe and set no `Parameters`
at all. Under mTLS every probe would have failed its handshake, and since the probe swallows per-peer failures
the node would have reported "no existing cluster found" and silently never joined - a Kubernetes-only failure
no other test would have caught. The server's `Parameters` are threaded in and pinned by
`KubernetesAutoJoinTlsInheritanceTest`. The comment there claiming the `RaftProperties` clone already carried
TLS over was wrong (Ratis keeps TLS on `Parameters`, not `RaftProperties`) and has been corrected in place.

**`mutualAuth=false` logs a WARNING, and the certificate needs both EKUs.** Because one `GrpcTlsConfig` backs
the server and client endpoints alike, a node presents its own certificate both as the TLS server certificate
and, under `mutualAuth`, as the client certificate it dials peers with - so it needs `serverAuth` *and*
`clientAuth` extended key usage. A CA that issues single-purpose certificates produces a handshake failure the
readable-file checks cannot catch, because the file is perfectly readable and merely the wrong kind of
certificate; `arcadedb.ha.tls.certChainFile`'s description says so.

**A world-readable private key is warned about, not refused.** `requireReadableFile` asks only whether this
process can read a file; a group- or world-readable Raft key additionally hands a cluster identity to any
local account that can read it. That is a `WARNING` naming the file rather than a startup failure, because a
mounted secret may arrive with permissions the node cannot change and refusing to start would be the worse
outcome. No-op where POSIX permissions have no meaning.

**Nothing about the HTTP side channels changed.** `SnapshotHttpHandler`'s `X-ArcadeDB-Cluster-Token` check is
untouched and still required; TLS on gRPC does not replace it. That is stated in the
`arcadedb.ha.tls.enabled` setting description so an operator reading the settings does not infer otherwise.

## Test plan

- [x] `RaftGrpcTlsConfigTest` (9 unit tests): transport stays plaintext by default; disabled TLS ignores
      unusable paths; enabled TLS attaches a file-based `GrpcTlsConfig` to `Parameters` with the right cert,
      key, trust file and mTLS flag; `mutualAuth=false` honoured; unset / missing / directory-instead-of-file
      paths each throw `ConfigurationException` naming the setting; a key readable beyond its owner is warned
      about at `WARNING` and still accepted, an owner-only key silently.
- [x] `Issue3890RaftParametersPublicationTest` (3 unit tests): WHEN `raftParameters` becomes visible, which
      `refreshRaftClient()` can read from a leader-change callback while `restartRatis()` rebuilds it. The
      published object carries the TLS conf *and* the allowlist customizer; a rebuild that throws on an
      unreadable cert leaves the previous working value in place; a plaintext cluster publishes usable
      `Parameters` with no TLS conf. Asserts on the field, not the return value - the return value was never
      the part that was wrong.
- [x] `KubernetesAutoJoinTlsInheritanceTest` (2 unit tests): the probe client inherits the server's
      `GrpcTlsConfig`, and the legacy constructor still yields a usable plaintext `Parameters` rather than an
      NPE on first contact.
- [x] `Issue3890RaftServerOnlyTlsIT` (2 integration tests): with `mutualAuth=false` the anonymous client that
      the mTLS IT shows being rejected is *accepted*, and the cluster still replicates over the encrypted
      transport. This pins the positive half of `mutualAuth` in the suite instead of leaving it to a one-off
      manual disarm.
- [x] `Issue3890RaftMtlsIT` (4 integration tests) - a 3-node `BaseRaftHATest` cluster booted with mTLS against
      a throwaway CA:
  - `clusterReplicatesOverMutualTls` - leader election, a 200-record transaction, and every follower caught up,
    all over TLS.
  - `raftPortCompletesHandshakeWithClusterCertificate` - positive control proving the port really speaks TLS
    and serves the node certificate.
  - `raftPortRejectsPeerSignedByAForeignCa` - a peer with a certificate from a different CA is rejected at the
    handshake.
  - `raftPortRejectsPeerWithoutCertificate` - an anonymous host that merely knows the port is rejected at the
    handshake. This is precisely the case the IP allowlist alone cannot stop.
- [x] Full `ha-raft` unit suite: **998 tests, 0 failures**.
- [x] Adjacent HA integration tests unaffected (plaintext clusters still work): `RaftReplication3NodesIT`,
      `RaftHAConfigurationIT`, `RaftHAServerIT`, `RaftLeaderFailoverIT`, `Issue6091GrpcRoutingTableIT`,
      `Issue5275MembershipSelfHealIT`, `Issue4836K8sScaleUpTest` - 28 tests, 0 failures.
- [x] Engine configuration tests: 85 tests, 0 failures.

### Proving the tests can fail

Neither new test is a tautology; both were disarmed and observed red before being restored:

| Disarm | Result |
|---|---|
| `buildRaftClient` reverted to `new Parameters()` | `clusterReplicatesOverMutualTls` fails - `ReplicationDispatchedTimeoutException`, group commit never lands |
| `arcadedb.ha.tls.mutualAuth=false` in the IT | `raftPortRejectsPeerSignedByAForeignCa` **and** `raftPortRejectsPeerWithoutCertificate` both fail - the handshakes succeed |
| `arcadedb.ha.tls.mutualAuth=true` in `Issue3890RaftServerOnlyTlsIT` | `serverOnlyTlsAcceptsAPeerWithoutAClientCertificate` fails - `Socket Broken pipe`, the server hangs up on the certificate-less client |
| `raftParameters` published early in `buildParameters` | `aFailedRebuildLeavesThePreviouslyPublishedParametersInPlace` fails |
| the private-key permission check removed | `aPrivateKeyReadableByOthersIsWarnedAboutButStillAccepted` fails |

### Test PKI

`RaftTestPki` generates a throwaway CA and a CA-signed `CN=localhost` node certificate at test time with the
JDK's own `keytool` (plus the JDK `KeyStore` API to export the PKCS#8 key, which `keytool` cannot do). No
certificate is committed and none can expire in place; each instance also clears its own files first so a
directory left by an earlier run cannot make `keytool` refuse a duplicate alias. Netty's
`SelfSignedCertificate` was tried first and is unusable here: it needs either `sun.security.x509` (strongly
encapsulated since JDK 16) or BouncyCastle, which this project does not depend on.

The negative handshakes pin **TLS 1.2** deliberately. Under TLS 1.3 the client finishes the handshake before
the server has verified its certificate, so a rejection would surface asynchronously on a later read rather
than out of `startHandshake()`, and the assertion would be racy.

## Out of scope / follow-ups

- **Operator documentation** (openssl recipe for a self-signed cluster CA, cert-manager for StatefulSets where
  the SAN must match the pod's stable DNS name, Vault PKI, rotation and expiry notes). ArcadeDB's user docs
  live in the separate `arcadedb-docs` repository, so this PR carries the operator-facing essentials in the
  setting descriptions instead - the SAN requirement, the PKCS#8 key format, the fail-fast behaviour, and the
  fact that mTLS does not replace the cluster-token check. A docs PR should follow.
- **`raftProperties` is non-volatile while `raftParameters` is.** Raised in review; deliberately not changed
  here. `restartRatis()` is not synchronized against `refreshRaftClient()` either way, so the window predates
  this PR and adding `volatile` would narrow the visibility gap without closing the race. Making that field
  safe means deciding how those two paths synchronize - a change to this class's concurrency contract, which
  wants its own commit and its own test rather than a keyword added in passing on a security PR.
- **Credential format asymmetry.** These settings take raw PEM files (Ratis/Netty's native `GrpcTlsConfig`
  file constructor) while the existing `arcadedb.ssl.*` settings for HTTP/Bolt/Redis are keystore+password.
  Forced by the underlying APIs, but an operator securing both now manages two formats: worth a callout in the
  docs PR.
- **Certificate rotation without a restart** - the issue lists it as out of scope; the config path here is
  static, read once per Ratis (re)start.
- **KMS / Vault agent integration for certificate retrieval** - left to ops, as the issue states.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2118,7 +2118,10 @@ public enum GlobalConfiguration {
       arcadedb.ha.tls.trustCertCollectionFile, binding peer identity to the certificate rather than to a \
       source IP address. Set to false for server-only TLS, \
       which encrypts the traffic but leaves the Raft port open to any client that trusts the cluster CA - the \
-      back door this setting exists to close. Only meaningful when arcadedb.ha.tls.enabled is true.""",
+      back door this setting exists to close. Turning it off does NOT make \
+      arcadedb.ha.tls.trustCertCollectionFile optional: the dialling node still needs the cluster CA to \
+      validate the certificate presented by the node it dials. Only meaningful when arcadedb.ha.tls.enabled \
+      is true.""",
       Boolean.class, true),
 
   // POSTGRES
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2074,6 +2074,48 @@ public enum GlobalConfiguration {
       stickiness (drop a host from the allowlist as soon as it stops resolving).""",
       Long.class, 300_000L),
 
+  HA_TLS_ENABLED("arcadedb.ha.tls.enabled", SCOPE.SERVER,
+      """
+      Negotiate TLS on the Raft gRPC transport, which carries AppendEntries, RequestVote and the Ratis \
+      admin/client calls between the nodes of a cluster. When true, arcadedb.ha.tls.certChainFile, \
+      arcadedb.ha.tls.privateKeyFile and arcadedb.ha.tls.trustCertCollectionFile must all name a readable PEM \
+      file or the node refuses to start. Default is false so a development or test cluster still starts with \
+      no configuration at all; the peer-address allowlist (arcadedb.ha.peerAllowlist.enabled) is IP-based and \
+      is defeated by a spoofed source address or a compromised peer, so it is a best-effort default rather \
+      than a substitute for TLS on an untrusted network. TLS on the gRPC port does not replace the \
+      X-ArcadeDB-Cluster-Token check on the HTTP side channels (snapshot download, database verify).""",
+      Boolean.class, false),
+
+  HA_TLS_CERT_CHAIN_FILE("arcadedb.ha.tls.certChainFile", SCOPE.SERVER,
+      """
+      PEM file holding this node's Raft gRPC certificate followed by any intermediate CA certificates. The \
+      certificate must carry a subject alternative name matching the address the other nodes use to dial this \
+      node in arcadedb.ha.serverList (on Kubernetes, the pod's stable headless-service DNS name). Required \
+      when arcadedb.ha.tls.enabled is true.""",
+      String.class, ""),
+
+  HA_TLS_PRIVATE_KEY_FILE("arcadedb.ha.tls.privateKeyFile", SCOPE.SERVER,
+      """
+      PEM file holding the PKCS#8 private key that matches arcadedb.ha.tls.certChainFile. Required when \
+      arcadedb.ha.tls.enabled is true.""",
+      String.class, ""),
+
+  HA_TLS_TRUST_CERT_COLLECTION_FILE("arcadedb.ha.tls.trustCertCollectionFile", SCOPE.SERVER,
+      """
+      PEM file holding the cluster CA certificate(s) that sign every node certificate. A peer presenting a \
+      certificate this collection does not chain to is rejected during the TLS handshake, before any Raft \
+      message is read. Required when arcadedb.ha.tls.enabled is true.""",
+      String.class, ""),
+
+  HA_TLS_MUTUAL_AUTH("arcadedb.ha.tls.mutualAuth", SCOPE.SERVER,
+      """
+      Require the dialling peer to present a client certificate signed by the CA collection in \
+      arcadedb.ha.tls.trustCertCollectionFile, binding peer identity to the certificate rather than to a \
+      source IP address. Set to false for server-only TLS, \
+      which encrypts the traffic but leaves the Raft port open to any client that trusts the cluster CA - the \
+      back door this setting exists to close. Only meaningful when arcadedb.ha.tls.enabled is true.""",
+      Boolean.class, true),
+
   // POSTGRES
   POSTGRES_PORT("arcadedb.postgres.port", SCOPE.SERVER,
       "TCP/IP port number used for incoming connections for Postgres plugin. Default is 5432", Integer.class, 5432),
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -2090,8 +2090,13 @@ public enum GlobalConfiguration {
       """
       PEM file holding this node's Raft gRPC certificate followed by any intermediate CA certificates. The \
       certificate must carry a subject alternative name matching the address the other nodes use to dial this \
-      node in arcadedb.ha.serverList (on Kubernetes, the pod's stable headless-service DNS name). Required \
-      when arcadedb.ha.tls.enabled is true.""",
+      node in arcadedb.ha.serverList (on Kubernetes, the pod's stable headless-service DNS name). Every node \
+      both accepts and initiates Raft connections, so with arcadedb.ha.tls.mutualAuth=true this one \
+      certificate is presented as the TLS server certificate AND as the client certificate: it needs BOTH the \
+      serverAuth and clientAuth extended key usages. A CA that issues single-purpose certificates will \
+      otherwise produce a handshake failure that the startup file checks cannot catch, because the file is \
+      perfectly readable and merely the wrong kind of certificate. Required when arcadedb.ha.tls.enabled is \
+      true.""",
       String.class, ""),
 
   HA_TLS_PRIVATE_KEY_FILE("arcadedb.ha.tls.privateKeyFile", SCOPE.SERVER,
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/KubernetesAutoJoin.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/KubernetesAutoJoin.java
@@ -21,6 +21,7 @@ package com.arcadedb.server.ha.raft;
 import com.arcadedb.log.LogManager;
 import com.arcadedb.server.ArcadeDBServer;
 import org.apache.ratis.client.RaftClient;
+import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
 import org.apache.ratis.protocol.RaftClientReply;
 import org.apache.ratis.protocol.RaftGroup;
@@ -52,9 +53,11 @@ import java.util.logging.Level;
  * {@code Mode.ADD} is atomic, so concurrent joins from different pods are safe even without
  * additional coordination.
  * <p>
- * <b>Security note:</b> Peer discovery uses DNS resolution of the headless service hostname.
- * The Raft gRPC transport does not enforce cluster-token authentication. In production Kubernetes
- * deployments, restrict gRPC port access to only ArcadeDB StatefulSet pods via a NetworkPolicy.
+ * <b>Security note:</b> Peer discovery uses DNS resolution of the headless service hostname, and the
+ * Raft gRPC transport does not enforce cluster-token authentication. Secure the port with mTLS
+ * ({@code arcadedb.ha.tls.*}, issue #3890) and restrict access to the StatefulSet pods with a
+ * NetworkPolicy. The probe client below inherits the server's transport configuration, so it speaks TLS
+ * exactly when the cluster does.
  *
  * @author Luca Garulli (l.garulli@arcadedata.com)
  */
@@ -112,14 +115,30 @@ public class KubernetesAutoJoin {
   private final RaftGroup      raftGroup;
   private final RaftPeerId     localPeerId;
   private final RaftProperties raftProperties;
+  // Ratis transport parameters of the local server. Ratis carries the gRPC TLS configuration on
+  // Parameters, NOT on RaftProperties, so the probe client below has to be handed these to reach a peer
+  // whose Raft port requires TLS (issue #3890). Never null.
+  private final Parameters     raftParameters;
   private       Sleeper        sleeper = Thread::sleep;
 
+  /**
+   * Builds an auto-join with no transport parameters, so its probe client dials in plaintext. Only
+   * correct for a cluster with {@code arcadedb.ha.tls.enabled=false}; production wiring goes through
+   * {@link #KubernetesAutoJoin(ArcadeDBServer, RaftGroup, RaftPeerId, RaftProperties, Parameters)} with the
+   * same {@link Parameters} the local Ratis server was built from.
+   */
   public KubernetesAutoJoin(final ArcadeDBServer server, final RaftGroup raftGroup,
       final RaftPeerId localPeerId, final RaftProperties raftProperties) {
+    this(server, raftGroup, localPeerId, raftProperties, null);
+  }
+
+  public KubernetesAutoJoin(final ArcadeDBServer server, final RaftGroup raftGroup,
+      final RaftPeerId localPeerId, final RaftProperties raftProperties, final Parameters raftParameters) {
     this.server = server;
     this.raftGroup = raftGroup;
     this.localPeerId = localPeerId;
     this.raftProperties = raftProperties;
+    this.raftParameters = raftParameters != null ? raftParameters : new Parameters();
   }
 
   /** Package-private test hook: replaces the real sleeps so retry tests run instantly. */
@@ -195,6 +214,7 @@ public class KubernetesAutoJoin {
         try (final RaftClient tempClient = RaftClient.newBuilder()
             .setRaftGroup(targetGroup)
             .setProperties(tempProps)
+            .setParameters(raftParameters)
             .setRetryPolicy(PROBE_RETRY_POLICY)
             .build()) {
 
@@ -310,13 +330,16 @@ public class KubernetesAutoJoin {
     return buildProbeProperties();
   }
 
+  /** Package-private for unit testing that the probe client inherits the server's transport configuration. */
+  Parameters probeParametersForTest() {
+    return raftParameters;
+  }
+
   private RaftProperties buildProbeProperties() {
-    // Clone the main server's RaftProperties so TLS, gRPC flow-control, authentication and any
-    // other operator-set security knobs carry over into the short-lived probe client. Starting
-    // from a bare RaftProperties() would silently build a plaintext client that cannot talk to
-    // a TLS-only peer (handshake failure) and - more dangerously - could downgrade a deployment
-    // to plaintext if Ratis ever allowed an unencrypted fallback. Only the probe-specific
-    // timeouts are then overridden.
+    // Clone the main server's RaftProperties so gRPC flow-control, message sizing, authentication and
+    // any other operator-set knobs carry over into the short-lived probe client; only the probe-specific
+    // timeouts are then overridden. Note that the gRPC TLS configuration is NOT among them - Ratis keeps
+    // it on Parameters, which is why raftParameters is threaded in separately and set on the client.
     final RaftProperties props = new RaftProperties(raftProperties);
     props.set("raft.server.rpc.type", "GRPC");
     RaftServerConfigKeys.Rpc.setTimeoutMin(props,
--- a/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftHAServer.java
+++ b/ha-raft/src/main/java/com/arcadedb/server/ha/raft/RaftHAServer.java
@@ -228,9 +228,11 @@ public class RaftHAServer implements HealthMonitor.HealthTarget {
   // buildParameters() on each Ratis (re)start. Kept so refreshRaftClient() can rebuild the leader's
   // self-client with the SAME transport configuration: a client built with an empty Parameters would
   // dial the Raft port in plaintext and fail every handshake once TLS is on (issue #3890).
-  // Only meaningful once buildParameters() has RETURNED: it is assigned after the TLS configuration has
-  // been validated and installed, never before, so a ConfigurationException from a bad cert/key/trust path
-  // cannot leave this field pointing at a half-built, TLS-less transport configuration.
+  // Published by buildParameters() only on its LAST line, once the TLS configuration and the inbound-peer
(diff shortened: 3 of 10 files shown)


robfrank:
All three actionable items applied in 7cb6c86 (plus bb52aefa, pushed before the review landed - see below).

1. **`raftParameters` assigned before `applyTls` could throw** - fixed. The field is now published only after
   `applyTls` returns, and the invariant is stated on the field itself so the ordering is not silently
   re-broken later.
2. **INFO log re-read `HA_TLS_MUTUAL_AUTH` from configuration** - fixed; it now reads
   `tlsConfig.getMtlsEnabled()` off the object it just built.
3. **Unbounded `waitFor()` in `RaftTestPki.keytool()`** - fixed, and the fix had to be slightly more than a
   bounded `waitFor`: the old code drained the output pipe with `readAllBytes()` *before* waiting, so a hung
   keytool would have blocked there and never reached any timeout. Output now goes to a temp file, the wait is
   bounded at 120s (a hang detector, not a latency bound), and a timeout raises a named failure with the
   command and whatever keytool managed to print.

On item 4, the `GrpcConfigKeys.TLS` claim - it does hold, and here is the check, since a Ratis-internals
assertion should not rest on "the IT passes". From `ratis-grpc-3.3.0-sources.jar`, `GrpcFactory`:

```java
public GrpcFactory(Parameters parameters) {
  this(GrpcConfigKeys.Server.servicesCustomizer(parameters),
      GrpcConfigKeys.TLS.conf(parameters),        // <- the single entry this PR sets
      GrpcConfigKeys.Admin.tlsConf(parameters),
      GrpcConfigKeys.Client.tlsConf(parameters),
      GrpcConfigKeys.Server.tlsConf(parameters));
}
```

and in the `SslContexts` constructor it consumes them as `defaultSslContext = buildMethod.apply(tlsConfig, null)`,
with each of the three per-endpoint contexts falling back to it (`buildMethod.apply(adminTlsConfig, defaultSslContext)`,
and the same for client and server). So `TLS.conf` is the default for all three, and the per-endpoint keys are
overrides we deliberately do not set. This reaches our transport unchanged because `FixedGrpcFactory` - our
RATIS-2523 workaround - passes `Parameters` straight to `super(parameters)`.

**One thing found while the review was in flight** (commit bb52aefa, so it is in the reviewed tree but was not
in the original description): `KubernetesAutoJoin` opens its *own* short-lived `RaftClient` for the membership
probe and set no `Parameters` at all. Under mTLS every probe would have failed its handshake, and since the
probe swallows per-peer failures the node would have reported "no existing cluster found" and silently never
joined - the same class of bug as the leader self-client, in a path that only shows up on Kubernetes. The
server's `Parameters` are now threaded in, and `KubernetesAutoJoinTlsInheritanceTest` pins it.

That code carried a comment claiming the `RaftProperties` clone already carried TLS over. It did not: Ratis
keeps the gRPC TLS configuration on `Parameters`, not on `RaftProperties`. The comment has been corrected in
place rather than left to mislead the next reader.

robfrank:
Cycle 2 applied in 1fc4bef. Items 1, 2 and 4 are in; item 3 is deliberately not.

**1. `RaftHAServer` class javadoc.** Rewritten. Rather than appending mTLS to the NetworkPolicy sentence it now
ranks the three mitigations and says what separates them - mTLS binds identity to a certificate, the allowlist
binds it to an address that can be spoofed, the NetworkPolicy is network-level isolation worth having under
either. It also states that none of the three replaces the `X-ArcadeDB-Cluster-Token` check on the HTTP side
channels, since this is the class an operator reads first and that is the inference most easily made wrongly.

**2. Dual-EKU requirement.** Good catch, and it is the sharpest kind of gap: the fail-fast checks CANNOT catch
it, because the file is perfectly readable and merely the wrong kind of certificate.
`HA_TLS_CERT_CHAIN_FILE`'s description now says that with `mutualAuth=true` the one certificate is presented
both as the TLS server certificate and as the client certificate, that it therefore needs both `serverAuth`
and `clientAuth`, and what a single-purpose CA will produce instead.

**4. Regression test for the `mutualAuth=false` path.** Added as `Issue3890RaftServerOnlyTlsIT` - a separate
class because the setting is fixed per cluster in `onServerConfiguration`. It asserts the positive claim the
disarm exercise only established by hand: with `mutualAuth=false`, the anonymous client that
`Issue3890RaftMtlsIT` shows being rejected completes the handshake, and the cluster still replicates over the
encrypted transport. The pair is now diagnostic rather than just green - a regression that quietly stopped
requesting a client certificate turns the mTLS IT red while this one stays green, and the two results together
name which setting broke.

Verified armed the same way as the others: flipping this class back to `mutualAuth=true` turns
`serverOnlyTlsAcceptsAPeerWithoutAClientCertificate` red (`Socket Broken pipe` - the server hangs up on the
certificate-less client). `RaftTestPki.connect()` now carries the TLS 1.2 socket setup both ITs need, with the
TLS 1.3 racy-assertion rationale on it.

**3. `raftParameters` volatile, `raftProperties` not - not applied.** The review's own reasoning is why:
`restartRatis()` is not synchronized against `refreshRaftClient()` either way, so the window predates this
change and adding `volatile` to `raftProperties` would narrow the visibility gap without closing the race.
Making that field safe means deciding how those two paths synchronize, which is a change to this class's
concurrency contract and wants its own commit and its own test - not a keyword added in passing on a security
PR. Left as-is deliberately rather than half-fixed.

**On the design observation** (PEM files here vs. keystore+password for `arcadedb.ssl.*`): agreed, and it is
forced - Ratis's `GrpcTlsConfig` file constructor takes PEM, while the HTTP/Bolt/Redis listeners go through
JSSE. Noted for the docs PR; no code change.

Test totals after this cycle: 993 `ha-raft` unit tests green, and 6 integration tests across the two mTLS IT
classes, plus the adjacent HA ITs (`RaftReplication3NodesIT`, `RaftHAConfigurationIT`, `RaftHAServerIT`,
`RaftLeaderFailoverIT`, `Issue6091GrpcRoutingTableIT`, `Issue5275MembershipSelfHealIT`, `Issue4836K8sScaleUpTest`)
still green, which is what covers the plaintext default path.

robfrank:
Cycle 3 applied in f444b14. Both points, including the one flagged as nice-to-have.

**Publication window.** Correct, and the sharper version of the cycle-2 fix rather than a separate one: the
last round moved the assignment past `applyTls`, which closed the throwing path but left the object visible
for the few lines in which the allowlist customizer had not been installed yet. The field's own comment
claimed more than the code delivered.

Fixed by making the contract literal instead of documented: the allowlist install is now its own method
(`installPeerAllowlist`), and `this.raftParameters = parameters` is `buildParameters`'s **last line** on every
path - nothing above publishes, nothing below adds. That also removes the two early `return parameters`
statements that made the old shape easy to get wrong in the first place.

**The missing test.** Added, because after two rounds of getting the publication moment wrong it should not
depend on the next reader noticing:

`Issue3890RaftParametersPublicationTest`, three cases, deliberately asserting on the FIELD rather than on the
returned value - the returned value was never the part that was wrong:

- the published object carries the TLS conf **and** the services customizer (the cycle-3 window);
- a rebuild that throws on an unreadable cert leaves the previous, working `Parameters` in place, same
  instance, TLS conf intact (the cycle-2 window, and the restart case you asked for - a cert file that
  becomes unreadable mid-life);
- a plaintext cluster publishes usable `Parameters` with no TLS conf and the customizer still installed.

It builds a `RaftHAServer` whose constructor has run but whose Ratis server was never started, following the
`newDetachedServer` pattern already used in `Issue6202UnambiguousPeerHttpAddressTest`, so nothing binds a port
or touches disk. `buildParameters` is package-private for it, with a comment saying why.

Verified armed the same way as everything else here: restoring the early publication turns
`aFailedRebuildLeavesThePreviouslyPublishedParametersInPlace` red.

Totals after this cycle: **996 `ha-raft` unit tests** and **10 integration tests** (`Issue3890RaftMtlsIT`,
`Issue3890RaftServerOnlyTlsIT`, `RaftReplication3NodesIT`, `Issue5275MembershipSelfHealIT`), all green.

robfrank:
Cycle 4 applied in 5249046. This is the last automated cycle (`--max-cycles=4`); the branch is handed to the
developer from here, and no further pushes will happen on it.

**1. Private key permissions.** Applied. `requireReadableFile()` only ever asked whether *this process* can
read the file, never who else can - and a group- or world-readable Raft key hands a cluster identity to any
local account that can read it, which is the thing mTLS exists to prevent. Now a `WARNING` naming the setting
and the absolute path, not a refusal: a mounted secret or a container image may arrive with permissions the
node cannot change, and failing to start would be the worse outcome. It is a no-op where POSIX permissions
have no meaning (Windows), and on a file that vanished between the two checks. Two tests pin it, and it is
verified armed: removing the check turns `aPrivateKeyReadableByOthersIsWarnedAboutButStillAccepted` red.

**2. Reliance on a Ratis internal.** Applied, and it is the right thing to worry about - an upstream change to
`GrpcFactory`'s defaulting would silently reintroduce the exact self-client-dials-plaintext bug this PR fixes,
with no local test failing until someone bumps the dependency. `applyTls`'s javadoc now pins the version it
was verified against (**Ratis 3.3.0**), spells out what it verified (`GrpcFactory(Parameters)` reads
`TLS.conf` plus the three per-endpoint keys; its `SslContexts` constructor builds the `TLS.conf` context first
and passes it as the fallback for admin, client and server), says what an upstream change would break, and
names `Issue3890RaftMtlsIT` as the test to run on a version bump.

**4. Trust file required even with `mutualAuth=false`.** Applied - `HA_TLS_MUTUAL_AUTH`'s description now says
so explicitly: the dialling node still needs the cluster CA to validate the certificate it is offered.

**3. The split PEM header - flagged, not applied,** which matches how it was raised. The concern is fair as a
precedent: the alternatives are all worse, though. `--no-verify` defeats the hook outright rather than for one
literal; a committed fixture is the thing the hook is actually protecting against and would expire; and no
real key material exists here at any point, since every key is generated fresh per run into `target/` and
never committed. The comment on the constants says exactly that, so a contributor copying it has to read why
first. Worth a maintainer's call rather than mine.

**On running the suite.** Confirmed green locally on this SHA:

- `ha-raft` unit lane: **998 tests, 0 failures**
- integration: `Issue3890RaftMtlsIT` (4), `Issue3890RaftServerOnlyTlsIT` (2), `RaftReplication3NodesIT` (2) -
  **8 tests, 0 failures**
- earlier in the branch, also green: `RaftHAConfigurationIT`, `RaftHAServerIT`, `RaftLeaderFailoverIT`,
  `Issue6091GrpcRoutingTableIT`, `Issue5275MembershipSelfHealIT`, `Issue4836K8sScaleUpTest`

On the TLS 1.2-pinned handshake tests specifically: pinning 1.2 is what makes them *less* timing-sensitive,
not more. Under TLS 1.3 the client finishes the handshake before the server validates its certificate, so a
rejection arrives asynchronously on a later read and the assertion becomes a race; under 1.2 the server
rejects during the handshake and `startHandshake()` throws deterministically. Every negative case here has
been observed both passing and failing on demand.

One thing to note for CI rather than for review: a full-reactor `mvn -o verify` currently stops at
`arcadedb-grpc` in an offline local repo because protobuf 4.36.0 (bumped on `main` in 96940ac) has not been
fetched. Unrelated to this branch - it reproduces on `main` - but it means the reactor-wide run has to be
online.

Issue #6930, opened 2026-08-31
Extended-protocol pair: Describe(P) on a SET/SAVEPOINT portal sends no reply, and Execute answers NoData instead of CommandComplete


ruispereira:
These two are a pair. Individually each is a protocol violation; together they happen to cancel out for pgjdbc, which is why the existing ITs pass. They should be fixed in one change.

### 1. `Describe('P')` on a portal with no statement and no columns sends no reply at all

`postgresw/src/main/java/com/arcadedb/postgres/PostgresNetworkExecutor.java:426-432`, the `type == 'P'` branch of `describeCommand()`:
```java
} else {
  if (portal.columns != null) {
    writeRowDescription(portal.columns, portal.resultFormats);
    portal.rowDescriptionSent = true;
  }
}
```
When `portal.sqlStatement == null` **and** `portal.columns == null`, nothing is written. The protocol requires every `Describe` to be answered by exactly one of `RowDescription` or `NoData`, and the sibling arms all do (`writeNoData()` at `:391` and `:429`).

Portals in that state are produced by `parseCommand()`: `SAVEPOINT`, `RELEASE`, `ROLLBACK TO` and `SET` set `ignoreExecution = true` and leave both fields null, and so does any non-`"sql"` language, whose `switch (portal.language)` falls through to `default -> //nooop`.

*Fix:* give that `else` an `else writeNoData();` (or call `writeNoData()` unconditionally when `portal.columns == null`), so the branch always emits exactly one Describe response.

### 2. `Execute` on a SET/SAVEPOINT portal answers `NoData` instead of `CommandComplete`

`PostgresNetworkExecutor.java:492-493`, in `executeCommand()`:
```java
if (portal.ignoreExecution)
  writeNoData();
```
`NoData` (`'n'`) is a `Describe` response only. The legal replies to `Execute` are `CommandComplete`, `EmptyQueryResponse`, `PortalSuspended` or `ErrorResponse`, so this Execute is never completed on the wire. `ignoreExecution` is set for `SAVEPOINT `, `RELEASE `, `ROLLBACK TO ` and `SET ` - and `SET` is what every driver and pooler sends during connection setup.

*Fix:* replace `writeNoData()` with `writeCommandComplete(portal.query, 0)`. `getTag()` currently returns `""` for `SET`/`SAVEPOINT`/`RELEASE`, so add those tags too (real Postgres sends `SET`, `SAVEPOINT`, `RELEASE`, `ROLLBACK`).

[...]

Issue #6929, opened 2026-08-31
A binary-format json (OID 114) bind parameter is decoded as if it carried an extra 4-byte length prefix


ruispereira:
### Summary
`deserializeBinary` decodes a binary-format `json` (OID 114) parameter by first reading an int32 out of the value bytes and treating it as a payload length. PostgreSQL's `json_send` emits the raw UTF-8 text with no inner length prefix, so the first four characters of the document are read as a length. Every neighbouring case in the same switch handles this correctly, and the module cannot decode what it itself emits.

### Where (HEAD `c1785ced6`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresType.java:1145-1150`:
```java
case JSON -> {
  int length = buffer.getInt();      // <- no such prefix on the wire
  byte[] bytes = new byte[length];
  buffer.get(bytes);
  yield parseJsonText(new String(bytes));
}
```
The Bind message's own parameter length was already consumed by `bindCommand`. Contrast the sibling arms: `VARCHAR/TEXT/BPCHAR` (`:1115`) and `BYTEA` use `valueAsBytes` directly, and `deserializeBinaryElement` returns a json array element as `new String(bytes, ...)` with no prefix.

The serializer side is raw too: `serializeAsBinary` has no `case JSON`, so it falls to `default -> serializeAsText`, which writes only the column length prefix plus the raw bytes. Since `hasBinaryEncoding()` returns true for `JSON`, the server advertises binary json on output while writing raw bytes that its own input path cannot read back.

### Repro
A client that declares parameter OID 114 in `Parse` and sends parameter format code 1 in `Bind`, e.g. `INSERT INTO Doc SET payload = $1` with `$1` bound as binary json `{"a":1}`.

The first 4 bytes `{"a1` are read as a big-endian int (0x7B226131 = 2065171761), so `new byte[length]` throws `OutOfMemoryError` or `NegativeArraySizeException`, or the following `get` throws `BufferUnderflowException`. `bindCommand`'s catch turns it into `ErrorResponse "Error on parsing bind message: ..."` and the insert never runs.

`PostgresTypeTest.deserializeBinaryJson` (`:996`) passes only because it hand-builds the non-standard payload the implementation expects, i.e. it encodes the bug rather than the protocol.

### Suggested fix
Make `case JSON` mirror `case VARCHAR, TEXT, BPCHAR`:
```java
[...]

Issue #6928, opened 2026-08-31
A NULL element inside a LIST column throws NullPointerException while writing the Postgres array literal


ruispereira:
### Summary
`serializeArrayToString()` dereferences the element before testing it for null, so a `LIST` column containing a null throws NPE while the Postgres array literal is being written. The null-handling arm exists at the bottom of the same chain but is unreachable.

### Where (HEAD `c1785ced6`)
`postgresw/src/main/java/com/arcadedb/postgres/PostgresType.java:874` - the first test in the loop:
```java
if (element instanceof Float || element.getClass() == float.class) {
```
For `element == null` the `instanceof` is false, so `element.getClass()` is evaluated and throws. The intended handling is the last arm of the same chain, `:911`:
```java
sb.append(element == null ? "NULL" : element.toString());
```
which is dead code for null because `:874` always throws first. `NULL` is exactly the right Postgres array-literal spelling, so only the ordering is wrong.

### Repro
Over the wire: `SELECT ['a', null] AS tags`, or any stored `LIST` property containing a null (`INSERT INTO Doc SET tags = ['a', null]` then `SELECT tags FROM Doc`).

`getColumns()` types the column `ARRAY_TEXT` from the first non-null element (`PostgresType.getTypeForValue` filters nulls), then `writeDataRows` -> `serializeAsText` -> `serializeArrayToString` throws. `executeCommand`/`queryCommand` catch it and send `ErrorResponse "Error on executing query: null"`, so the client receives a `RowDescription`, possibly some already-flushed `DataRow`s, and then an error with no usable message. `SELECT [null] AS l` fails the same way, since an all-null list falls back to `ARRAY_TEXT`.

`PostgresTypeTest.serializeAsTextCollectionAllNonNull` (`:1607`) deliberately sidesteps this with the comment "null handling in arrays has a known limitation", so no test asserts the current NPE.

### Suggested fix
Hoist the null check to the top of the loop body:
```java
if (element == null) { sb.append("NULL"); continue; }
```
and drop the now-redundant `element == null ? ...` from `:911`. While there, the `element.getClass() == float.class` / `== double.class` comparisons are always false for boxed values and can go.

### Scope / confidence
High. Verified that `getTypeForValue` filters nulls when picking the element type, so a null cannot change the announced OID and make this path unreachable, and that `getArrayTypeForElementType` is only ever called on a non-null element. The dead `element == null` arm is positive evidence the behaviour is unintended rather than a deliberate restriction.

[...]

PR #6954, opened 2026-08-31
fix(#6930,#6928,#4918,#6929): Postgres wire pair, null-in-array NPE, JSON bind decoding, INSERT ON DUPLICATE KEY SKIP

lvca:
## Summary

Fixes four independent, small issues in one batch:

- **#6930** - `Describe('P')` on a portal bound from `SAVEPOINT`/`RELEASE`/`ROLLBACK TO`/`SET` sent no reply at all (every `Describe` must be answered by exactly one of `RowDescription` or `NoData`). The matching `Execute` answered `NoData` (a Describe-only reply) instead of `CommandComplete`. Both are fixed together as the issue requires, since fixing only one desynchronizes pgjdbc. A non-`"sql"`-language portal (e.g. cypher) is deliberately left out of the `Describe` fix - it has no statement here either, but it DOES expect a result once executed (columns just aren't known yet at Describe time), so answering it with a premature `NoData` would break it; verified this the hard way when the first version of the fix regressed `PostgresProtocolIT`'s cypher/sqlscript tests, which rely on that yet-unfixed pre-existing gap for cypher to accidentally get its `RowDescription` from `Execute` instead.
- **#6928** - `serializeArrayToString()` dereferenced a `LIST` element before null-checking it, so a null inside a `LIST` column threw `NPE` while writing the Postgres array literal. The null check is now first, matching the dead `NULL` arm that was already there.
- **#6929** - A binary-format `json` (OID 114) bind parameter was decoded by first reading a bogus 4-byte length prefix that PostgreSQL's `json_send` never sends, corrupting/crashing on any real client that binds JSON in binary format. Now decoded as the raw UTF-8 text, exactly like `VARCHAR`/`TEXT`/`BPCHAR`.
- **#4918** - New SQL clause: `INSERT INTO Type CONTENT [ {...}, {...} ] ON DUPLICATE KEY SKIP`. A record that would violate a unique index - either one already committed, or an earlier record in the same batch - is skipped instead of aborting the rest of the array, and reported back as a row with `@skipped: true`, `@duplicateIndex`, `@duplicateKeys` and `@existingRID` so the caller always knows what happened. All other errors (mandatory field, type mismatch, ...) still abort, matching the scope agreed in the issue thread.

  **Design note on #4918** (better alternative to a naive implementation): a duplicate is *never* detected by catching the save failure. `Index.put()` only throws synchronously for a conflict against another key already staged in the SAME open transaction; a conflict against already-committed data is only detected by ArcadeDB at `commit()` time - by which point the record's bucket write has already happened and cannot be undone record-by-record without a full compensating delete (and a real risk of leaving a partially-indexed orphan record). Instead, each candidate record's unique indexes are probed directly with `Index.get()` *before* ever writing it - which also transparently sees this transaction's own pending, uncommitted entries, so an in-batch duplicate is caught the same way as a pre-existing one.

#6659 (Postgres portal-suspension eager materialization) was in the original batch but is **not** touched here: it's explicitly deferred to milestone 26.10.1 by its own tracking comment (lvca, 2026-08-24), pending PR #6658 (now merged) landing first. Forcing it into this PR would contradict that already-recorded decision.

## Test plan
- [x] `PostgresTypeTest` (227 tests) - added 2 new tests for #6928, updated 1 stale test that hand-encoded the #6929 bug instead of the real wire format
- [x] New `Issue6930DescribeExecuteSetSavepointIT` (raw-socket wire-protocol test, 3 cases) - verified it fails without the fix, passes with it
- [x] Full `postgresw` module test suite green, including `PostgresProtocolIT`, `Issue6701SetSessionLocalIT`, `Issue6698PreparedStatementCloseIT`, `Issue6543RollbackExtendedProtocolIT`, `Issue6548AbortedTransactionRollbackExtendedProtocolIT` - no regressions
- [x] New `InsertOnDuplicateKeySkipTest` (6 cases: pre-existing-record clash, same-batch clash, non-duplicate-key error still aborts, no-conflict happy path, baseline without the clause, `duplicate`/`skip` still usable as identifiers) - verified all 4 feature-dependent cases fail without the change, pass with it
- [x] Full `com.arcadedb.query.sql.parser.*Test` package (391 tests) and `DDLTest`, plus all existing `Insert*Test` classes - green, no grammar regressions
- [x] `mvn -o compile`/`test-compile` clean on `engine` and `postgresw`

🔗 https://github.com/ArcadeData/arcadedb/pull/new/fix/6930-6928-4918-6929-postgres-insert-batch

## Summary by CodeRabbit

* **New Features**
  * Added `INSERT ... ON DUPLICATE KEY SKIP` to skip records that conflict with unique keys.
  * Skipped duplicates include metadata describing conflicting keys, while other validation errors remain reported.
  * The word `duplicate` remains available as an identifier.

* **Bug Fixes**
  * Improved PostgreSQL handling for `SET` and `ROLLBACK TO` command responses.
  * Corrected PostgreSQL array serialization for null values.
  * Corrected PostgreSQL JSON binary deserialization for raw JSON payloads.

### DIFF
--- a/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLLexer.g4
+++ b/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLLexer.g4
@@ -293,6 +293,9 @@ MATCHES: M A T C H E S;
 INSTANCEOF: I N S T A N C E O F;
 KEY: K E Y;
 ITEM: I T E M;
+// Issue #4918: INSERT ... ON DUPLICATE KEY SKIP. Also listed among the keywords usable as an identifier, so a
+// schema that already has a type or property called "duplicate" keeps parsing.
+DUPLICATE: D U P L I C A T E;
 
 // ============================================================================
 // SPECIAL ATTRIBUTE TOKENS
--- a/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLParser.g4
+++ b/engine/src/main/antlr4/com/arcadedb/query/sql/grammar/SQLParser.g4
@@ -336,11 +336,17 @@ nestedProjectionItem
 /**
  * INSERT statement
  * INSERT INTO target [(fields)] VALUES (values) | SET field=value | CONTENT {...} | FROM query
- * [RETURN projection] [UNSAFE]
+ * [ON DUPLICATE KEY SKIP] [RETURN projection] [UNSAFE]
+ *
+ * ON DUPLICATE KEY SKIP (issue #4918): a record whose unique-index key already exists is silently skipped
+ * instead of aborting the whole batch - useful for bulk-inserting a CONTENT array where some rows may already
+ * exist. Only unique-key conflicts are swallowed; every other error (a missing mandatory field, a type
+ * mismatch, ...) still aborts the command, since silently hiding those would mask real data problems.
  */
 insertStatement
     : INSERT INTO (identifier (BUCKET identifier)? | bucketIdentifier)
       insertBody?
+      (ON DUPLICATE KEY SKIP_KW)?
       (RETURN projection)?
       (FROM? (selectStatement | LPAREN selectStatement RPAREN))?
       UNSAFE?
@@ -1605,6 +1611,7 @@ identifier
     | TIMESTAMP
     | DEFAULT
     | KEY
+    | DUPLICATE
     | FORMAT
     | CUSTOM
     | SKIP_KW
--- a/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLASTBuilder.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/antlr/SQLASTBuilder.java
@@ -4791,6 +4791,9 @@ public class SQLASTBuilder extends SQLParserBaseVisitor<Object> {
     // UNSAFE flag
     stmt.unsafe = insertCtx.UNSAFE() != null;
 
+    // ON DUPLICATE KEY SKIP (issue #4918)
+    stmt.onDuplicateKeySkip = insertCtx.DUPLICATE() != null;
+
     return stmt;
   }
 
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/InsertExecutionPlanner.java
@@ -42,6 +42,7 @@ public class InsertExecutionPlanner {
   protected InsertBody      insertBody;
   protected Projection      returnStatement;
   protected SelectStatement selectStatement;
+  protected boolean         onDuplicateKeySkip;
 
   public InsertExecutionPlanner() {
   }
@@ -53,6 +54,7 @@ public class InsertExecutionPlanner {
     this.insertBody = statement.getInsertBody() == null ? null : statement.getInsertBody().copy();
     this.returnStatement = statement.getReturnStatement() == null ? null : statement.getReturnStatement().copy();
     this.selectStatement = statement.getSelectStatement() == null ? null : statement.getSelectStatement().copy();
+    this.onDuplicateKeySkip = statement.isOnDuplicateKeySkip();
   }
 
   public InsertExecutionPlan createExecutionPlan(final CommandContext context) {
@@ -86,7 +88,7 @@ public class InsertExecutionPlanner {
   }
 
   private void handleSave(final InsertExecutionPlan result, final Identifier targetClusterName, final CommandContext context) {
-    result.chain(new SaveElementStep(context, targetClusterName, true));
+    result.chain(new SaveElementStep(context, targetClusterName, true, onDuplicateKeySkip));
   }
 
   private void handleReturn(final InsertExecutionPlan result, final Projection returnStatement, final CommandContext context) {
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SaveElementStep.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SaveElementStep.java
@@ -152,12 +152,13 @@ public class SaveElementStep extends AbstractExecutionStep {
    * either a previously committed one or an earlier record in this same batch, since {@link TypeIndex#get} reads
    * through to this transaction's own staged-but-uncommitted index entries.
    * <p>
-   * A key is exempted from the check only when the index's own uniqueness enforcement would exempt it: that is
-   * {@link LSMTreeIndexAbstract#isKeyNull} (every component null, not just one) AND the index's
-   * {@code NULL_STRATEGY} is {@code SKIP} - the same two conditions {@code TransactionIndexContext} gates its own
-   * commit-time duplicate check on. A composite key with only SOME null components, or an all-null key on an
-   * index built with {@code NULL_STRATEGY.INDEX} (where nulls are indexed and compared like any other value), is
-   * still a real key that must be probed.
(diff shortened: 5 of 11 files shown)


lvca:
Addressed in 197582f77:

- Removed the dead `"SET".equals(upperCaseText)` branch in `getTag()` instead of adding dispatch support for a bare `SET`: both places that set `portal.ignoreExecution` require `"SET "` with a trailing space, and real PostgreSQL doesn't accept a bare `SET` as valid syntax either, so there was nothing for that branch to ever match.
- Collapsed the `findDuplicateKeyConflict()` local into the `return` and dropped the (now genuinely single-statement) `if`'s braces, matching the project's style guide; dropped the now-unused `Identifiable` import.
- Dropped braces around the single-statement `onDuplicateKeySkip` append in `InsertStatement.toString()`.
- Added the missing `final` on the two locals in the rewritten `deserializeBinaryJson` test.
- Added a `ROLLBACK TO` regression case to `Issue6930DescribeExecuteSetSavepointIT`, matching the existing `SAVEPOINT`/`RELEASE` cases.

Not implementing the suggestion to make `ON DUPLICATE KEY SKIP`'s duplicate check atomic via an index/transaction-layer "check-and-claim" primitive: that would require a new atomic reserve-a-key operation in the indexing/transaction layer (`TransactionIndexContext`/`Index.put`) that doesn't exist today, well beyond issue #4918's scope of "skip an obviously-conflicting record instead of aborting the batch". The current pre-check-then-save approach has the same narrow concurrent-transaction race window that ArcadeDB's unique-index enforcement already has everywhere else (a genuine duplicate committed by another transaction between the probe and this transaction's own commit surfaces as `DuplicatedKeyException` from `commit()`, same as any other `INSERT` today) - it's a pre-existing engine-wide property, not something this PR regresses or is positioned to fix on its own.

The other nitpick about the `SaveElementStep` `if` at (then) line 173 was based on a two-statement block that genuinely needed its braces; fixed instead by collapsing it to one statement as above.

lvca:
Addressed in 80d54c14c - thanks for the thorough second pass, the composite-key null bug was real:

- Fixed \`findDuplicateKeyConflict()\`'s null handling to match the engine's actual semantics: a key is now only exempted when \`LSMTreeIndexAbstract.isKeyNull()\` says so (every component null, not just one) AND the index's \`NULL_STRATEGY\` is \`SKIP\`. This resolves both the confirmed composite-key bug and the \`NULL_STRATEGY.INDEX\` gap flagged in the first review pass - they were the same root cause.
- Added two regression tests reproducing the confirmed failure scenario (partially-null composite key, and an all-null key under \`NULL_STRATEGY.INDEX\`) - verified both fail against the old logic and pass with the fix.
- Changed \`@duplicateKeys\` to report a structured \`List\` instead of \`Arrays.toString()\`'s debug rendering.
- Added the concurrent-transaction TOCTOU caveat to the \`SaveElementStep\` constructor javadoc, framed as a pre-existing property of ArcadeDB's commit-time conflict model (every write has this window, not something specific to this clause) rather than a new gap.

All 8 tests in \`InsertOnDuplicateKeySkipTest\` green, plus the full existing Insert*/DDL test set - no regressions.

lvca:
Addressed in bdc9f3536 - the third pass was right, I over-corrected:

- Dropped the \`NULL_STRATEGY.SKIP\` gate entirely. \`TransactionIndexContext\`'s own commit-time checks (\`checkUniqueIndexKeys()\` and \`addIndexKeyLock()\`) exempt an all-null key **unconditionally**, and \`NullValuesIndexTest#nullStrategyIndex_uniqueAllowsMultipleNulls\` already proves 5 all-null keys persist under a \`NULL_STRATEGY.INDEX\` unique index. \`findDuplicateKeyConflict()\` now exempts purely on \`LSMTreeIndexAbstract.isKeyNull(keyValues)\`, matching that.
- The previous \`aFullyNullKeyIsStillCheckedUnderNullStrategyIndex\` test asserted the bug itself (record marked \`@skipped\`), so it's replaced with \`anAllNullKeyIsNeverADuplicateEvenUnderNullStrategyIndex\`, asserting the correct outcome (not skipped, both rows persist) - verified it fails against the previous logic and passes with the fix.

On the \`jsonb\` (OID 3802) follow-up: confirmed there's no \`JSONB\`/3802 handling anywhere in \`PostgresType\` today, so it's genuinely unimplemented rather than something #6929's fix touches or regresses - agreed that's a separate feature request, not something to fold into this PR.

All 8 \`InsertOnDuplicateKeySkipTest\` cases + all 9 \`NullValuesIndexTest\` cases green, plus the full existing Insert*/DDL suite, no regressions.

lvca:
Addressed both minor points from the 4th review pass in 89fa7d3de:

- **Performance**: \`SaveElementStep\` now caches the resolved unique-index list, keyed by type name, across the rows pulled by one \`syncPull()\` call instead of re-resolving the schema type and rebuilding \`getAllIndexes(true)\` on every row.
- **Test coverage**: added \`supportsTheSetInsertForm()\` and \`supportsTheFromQueryInsertForm()\`.

Worth flagging what turned up while writing the FROM-query test: \`CopyDocumentStep\` (the step behind \`INSERT ... FROM <query>\`) saves the copied record **eagerly**, bypassing \`SaveElementStep\` entirely, when the subquery is a column projection (\`SELECT sku, name FROM X\`) rather than a whole-element \`SELECT\` (\`SELECT FROM X\`) - it's a pre-existing branch in \`CopyDocumentStep\` unrelated to this PR. So \`ON DUPLICATE KEY SKIP\` combined with a projected \`FROM\` subquery silently has no effect (the insert either succeeds or throws exactly like it does without the clause), while a whole-element \`FROM\` subquery correctly goes through \`SaveElementStep\` and gets the skip behavior. Fixing \`CopyDocumentStep\` to always defer to \`SaveElementStep\` is a separate, more invasive change to code shared by every \`INSERT ... FROM SELECT\`, so I've scoped it out of #4918 and documented the boundary in the new test's comment rather than silently working around it. Happy to file it as its own follow-up issue if that's preferred.

All 10 \`InsertOnDuplicateKeySkipTest\` cases + full existing Insert*/DDL/NullValuesIndexTest suites green.

lvca:
Addressed in 337087526 - good catch, that test wasn't testing what its name claimed:

Added an explicit \`database.commit()\` + \`database.begin()\` between the setup \`INSERT\` and the tested statement in \`skipsARecordClashingWithAPreviouslyCommittedRecord\`, so the first record is genuinely committed (not just staged in the same open transaction \`TestHelper\`'s \`autoStartTx\` was keeping alive) before the probe runs against it. Confirms \`Index.get()\` does resolve real committed data correctly, not just same-transaction staged entries - the actual scenario the PR's design note is about.

All 10 cases in \`InsertOnDuplicateKeySkipTest\` still green.

lvca:
Addressed in 71ea7869a - the last review pass (6th) found no blocking issues, just one "worth double-checking, not a blocker" suggestion: every existing test used a STRING \`sku\` as the unique key, so it was worth confirming the probe's \`doc.get()\`-based key extraction agrees with the real index write for a non-STRING key too.

Added \`skipsARecordClashingOnANumericUniqueKey\` (INTEGER unique column, crossing a real transaction boundary like the committed-record test). Passes with no production-code changes needed - confirms there's nothing to fix here, just added confidence.

At this point 6 review passes have run across the iterations on this PR (1 CodeRabbit + several rounds of claude-review that progressively tightened the null-handling semantics), the last two passes found zero correctness issues and only very minor/confirmatory suggestions, and all 11 cases in \`InsertOnDuplicateKeySkipTest\` plus the full existing Insert*/DDL/NullValuesIndexTest suites are green. From my side this is ready to merge once CI finishes.

PR #6958, opened 2026-08-31
fix(#6948): re-schedule TimeSeries maintenance after an HA sealed-store repair

### LINKED ISSUE #6948: Follow-up to #6839: a TimeSeries type repaired through the new HA path is never re-scheduled for maintenance, so compaction, retention and downsampling stop for it
### Summary
The #6839 fix gave the engine-unavailable TimeSeries state a real way out, and the recovery works. But the third thing #6839 named - maintenance being gated off for such a type - was not addressed on the new path: a type that recovers through the Raft repair comes back readable and writable yet **permanently unmaintained** for the life of the process. Compaction, retention and downsampling never run for it again.

Found while validating #6839 against `main`. Parts 1 and 2 of that issue are genuinely fixed; this is the remainder.

### Where (HEAD `55e00aaa6`)
`schedule()` has exactly two call sites in main code:
- `engine/src/main/java/com/arcadedb/schema/LocalSchema.java:1913`, behind the gate at `:1911`:
```java
if (tsType.isEngineAvailable()) {
  try {
    getTimeSeriesMaintenanceScheduler().schedule(database, tsType);
```
which is skipped for precisely this type, since at schema load its engine is unavailable;
- `engine/src/main/java/com/arcadedb/schema/TimeSeriesTypeBuilder.java:153`, i.e. type creation only.

`ArcadeStateMachine.repairEngineWithSealedFile` (`ha-raft/src/main/java/com/arcadedb/server/ha/raft/ArcadeStateMachine.java:2404-2430`) calls `tsType.initEngine()` and re-checks `isEngineAvailable()`. It does not schedule maintenance, and nothing downstream does.

### Why the follower-only argument does not cover it
The leader-only skip lives **inside** the recurring task, not in `schedule()` - `engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesMaintenanceScheduler.java:107-110`:
```java
final DatabaseInternal dbInternal = (DatabaseInternal) db;
if (dbInternal.isReplicated() && !dbInternal.isLeader())
  return;
```
So a healthy follower still has a ticking task that takes over the moment it is elected leader. A repaired type has no task at all, and never will.

Note also that `runMaintenance` already returns early on `engine == null` (`:118-120`), so scheduling for an engine-unavailable type would have been harmless anyway. The `:1911` gate is belt-and-braces; leaving it is fine, leaving the repair path silent is not.

### Impact
For that type on that node, `engine.compactAll()`, `engine.applyRetention()` and `engine.applyDownsampling()` never run again. Those three have no other caller in main code (only `TimeSeriesMaintenanceScheduler.java:127`, `:130`, `:135`), so this is not "delayed": the mutable bucket grows unbounded - the scheduler's own javadoc says "Compaction is always scheduled to prevent unbounded mutable-bucket growth" - and configured retention and downsampling policies silently stop being applied.

### Repro
3-node Raft cluster, TimeSeries type with a retention policy.
1. Corrupt one byte of `<type>_shard_0.ts.sealed` on follower B and restart it (the original #6839 repro). The type comes up registered with no engine.
2. Let the leader compact so a `TsSealedBlob` is replicated. B repairs it and `isEngineAvailable()` flips true - the #6839 fix working as designed.
3. Transfer leadership to B.

B never compacts, never applies retention and never downsamples for that type until the database is closed and reopened.

### Suggested fix
After a successful `initEngine()` in `repairEngineWithSealedFile`, schedule maintenance:
```java
db.getSchema().getEmbedded().getTimeSeriesMaintenanceScheduler().schedule(db, tsType);
```
`schedule()` is idempotent - it cancels and replaces any existing task for the type name - so a repeated repair is safe.

### Test coverage
The fix added good tests for what it did fix: `Issue6839TsSealedBlobRecoveryTest` (5 cases), `Issue6839InitEngineRetryableTest` (3 cases) and `Issue6839PrometheusWriteEngineUnavailableIT`. None asserts anything about maintenance being scheduled after the repair, which is why the gap survived.

### Credit where due
Worth noting what the fix got right, since it was the non-obvious half: `initEngine()` is now genuinely safe to retry. `TimeSeriesShard.java:181-191` no longer closes the schema-registered mutable bucket when the sealed store fails to open (`PaginatedComponentFile.close()` sets `open=false` permanently, which would have made the retry yield an `isEngineAvailable() == true` type whose reads throw `FileNotFoundException`), and `TimeSeriesEngine.java:105-120` unwinds already-built shards via the new `closeAfterFailedInit()` so a corrupt shard N cannot poison shards 0..N-1 - without which recovery would have been limited to single-shard types.

### Scope / confidence
Medium-high. Traced end to end at this checkout: both `schedule()` call sites, the `:1911` gate, the repair method's full body, and the position of the leader check inside `runMaintenance` rather than `schedule`.

robfrank:
Closes #6948

## Summary

The #6839 fix gave the "registered but engine-unavailable" TimeSeries state a real way out, but a type that recovers through the Raft repair came back readable and writable yet **permanently unmaintained**: compaction, retention and downsampling never ran for it again for the life of the process.

`schedule()` has exactly two call sites in main code - type creation (`TimeSeriesTypeBuilder`), and `LocalSchema.readConfiguration()` behind an `isEngineAvailable()` gate. That gate is skipped for precisely the types this path repairs, because at schema load their engine is the thing that failed. `ArcadeStateMachine.repairEngineWithSealedFile()` called `initEngine()` and re-checked `isEngineAvailable()`, but scheduled nothing, and nothing downstream did either. Since `engine.compactAll()`, `engine.applyRetention()` and `engine.applyDownsampling()` have no caller outside the recurring maintenance task, the mutable bucket grew unbounded ("Compaction is always scheduled to prevent unbounded mutable-bucket growth", says the scheduler's own javadoc) and configured retention/downsampling policies silently stopped being applied for that type on that node until the database was closed and reopened.

"It is a follower, it would skip the work anyway" does not cover it: the leader-only skip lives **inside** the recurring task (`TimeSeriesMaintenanceScheduler.runMaintenance`), not in `schedule()`. A healthy follower keeps a ticking task that takes over the instant it is elected; a repaired one had none, and never would.

## Change

- `ArcadeStateMachine.repairEngineWithSealedFile()` now calls a new private `scheduleMaintenanceAfterRepair()` once the repair has actually succeeded (after the `isEngineAvailable()` re-check, so a failed repair still schedules nothing). It resolves the scheduler through `db.getSchema().getEmbedded()` and passes `LocalSchema.getDatabase()` - the same `Database` reference `LocalSchema` itself hands to `schedule()`, so the scheduler's `WeakReference` targets the object whose lifetime it is meant to track. Both repair entry points - the inline `TsSealedBlob` path and the sliced path from #4416 - funnel through this method, so both are covered.
- Scheduling failure is logged at WARNING (with stack trace and exception class) and swallowed: the repair has already succeeded and the type is usable again, so failing to *also* schedule it must not be reported as "the repair failed".
  The catch is deliberately wider than the `RejectedExecutionException` that `LocalSchema.readConfiguration()` catches at its own `schedule()` call site, and the difference is the caller rather than the callee: that one runs during a database open, where an escaping runtime exception fails the open and says so; this one runs inside the Raft apply path, whose contract here is that one type must not abort the apply of an entry that may carry blobs for others - the same reason `repairEngineWithSealedBlob` reports failure rather than throwing. So nothing may escape, a programming error included, and the log line names the exception class and the exact consequence so it reads as the bug it is rather than as a benign mid-shutdown rejection. Spelled out in the method javadoc.
- `schedule()` is idempotent (it cancels and replaces any existing task for the type name), so a repeated repair is safe and needs no first-time bookkeeping.
- New `TimeSeriesMaintenanceScheduler.isScheduled(String)` accessor. "This type is maintained" had no outward sign at all - the recurring task logs nothing when it finds nothing to do, and its three effects are invisible until enough data has accumulated for one of them to change something - so without it the regression test would have had nothing to assert on.

The `:1911` gate in `LocalSchema` is deliberately left alone. `runMaintenance` already returns early on `engine == null`, so it is belt-and-braces rather than load-bearing, and removing it is not what this bug is about.

## Test plan

- [x] `ha-raft`: new `Issue6948MaintenanceAfterRepairTest` (5 cases), same harness as `Issue6839TsSealedBlobRecoveryTest` - a real `LocalDatabase` plus a real unstarted `ArcadeStateMachine`, no mocking, no cluster (the defect is entirely inside the apply path):
  - `aRepairedTypeIsScheduledForMaintenanceAgain` - the regression: break the sealed file, reopen, assert engine-unavailable **and** not scheduled, apply the leader's blob, assert available **and** scheduled.
  - `aHealthyTypeIsMaintainedAfterAPlainReopen` - the control that pins the asymmetry a repaired type has to be restored to parity with, and the concrete reason the follower-only argument does not hold.
  - `aFailedRepairSchedulesNothing` - a garbage blob leaves the type unavailable and unscheduled; a task for an engine-less type would tick forever reporting that the type is looked after.
  - `aSecondCorruptionIsRepairedAndScheduledAgain` - recovery is not one-shot: corrupt, repair, corrupt again, repair again, scheduled both times. The second corruption + reopen is what makes it a *genuine* second repair; `applySealedBlobs` only takes the repair branch when `getEngine() == null`.
  - `aRedundantBlobAfterARepairLeavesTheTaskInPlace` - the other branch: a blob for a type whose engine is already healthy is installed through the live sealed store and never reaches the repair path; what it pins is that the task the earlier repair installed survives a routine install.
- [x] Negative control: with the `scheduleMaintenanceAfterRepair()` call commented out, `aRepairedTypeIsScheduledForMaintenanceAgain`, `aSecondCorruptionIsRepairedAndScheduledAgain` and `aRedundantBlobAfterARepairLeavesTheTaskInPlace` all fail, while the two control arms (`aHealthyTypeIsMaintainedAfterAPlainReopen`, `aFailedRepairSchedulesNothing`) still pass. Restored, all 5 pass.
- [x] `engine`: new `TimeSeriesMaintenanceSchedulerIsScheduledTest` proves the new accessor actually discriminates (fresh / unknown name / scheduled / re-scheduled / cancelled), so the ha-raft assertions cannot pass vacuously.
- [x] `mvn -o -pl ha-raft test -Dtest='ArcadeStateMachine*Test,Issue6839TsSealedBlobRecoveryTest,Issue4416SlicedSealedApplyTest,Issue4416SlicedSealedStoreTest,Issue6948MaintenanceAfterRepairTest'` - 131 tests, 0 failures.
- [x] `mvn -o -pl engine test -Dtest='TimeSeries*Test,Issue6839InitEngineRetryableTest'` - 213 tests, 0 failures.

Not covered by an automated test: the end-to-end 3-node repro (corrupt a follower's sealed file, let the leader compact, transfer leadership to the repaired follower). A cluster IT would add flakiness without adding evidence - the scheduling decision is entirely inside the apply path, and the leader check that the transfer exercises is already unit-covered inside `runMaintenance`.

## Review round 2

- **`aRepeatedRepairLeavesTheTypeMaintained` did not exercise a second repair.** Correct, and confirmed: `applySealedBlobs` only takes the repair branch when `tsType.getEngine() == null`, so after the first successful repair the second `applySealedBlobs` fell through to `installSealedFileBytes` and never reached `scheduleMaintenanceAfterRepair` again. Split into two tests that each say what they do: `aSecondCorruptionIsRepairedAndScheduledAgain` corrupts and reopens a second time so the repair branch is genuinely re-entered, and `aRedundantBlobAfterARepairLeavesTheTaskInPlace` names the healthy-install branch it actually covers. Both fail with the fix disabled.
- **WARNING vs SEVERE.** Agreed and raised to SEVERE. The javadoc frames the catch as also covering a programming error, and what it leaves behind - a type maintained by nothing - is the exact defect this method exists to prevent, so swallowing it must not also hide it.
- **`isScheduled()`'s `!future.isDone()` is unreachable today.** Accurate: `runMaintenance` catches `Throwable`, so only cancellation makes a fixed-rate future done. Left as is - it is the correct predicate for "live task" and costs nothing, and relying on `runMaintenance`'s catch-all from another class would be the more fragile of the two.
- The `schema.getDatabase()` / wrapped-instance and catch-`Exception` points were confirmations, no change needed.

## Review round 3

All three items were non-blocking observations; one produced a change.

- **The javadoc said "nothing may escape" while the catch is `Exception`, not `Throwable`.** Real inconsistency, fixed in the doc rather than the code: catching `Exception` is intentional. An `Error` means the JVM is no longer in a state this node can reason about, and the apply path's "one type must not abort the entry" contract is not a licence to keep applying Raft entries through one. The javadoc now says so explicitly and scopes the claim to exceptions.
- **SEVERE diverges from the sibling `RejectedExecutionException` catch's severity.** Kept, now with the reason written down next to it: the sibling catch runs during a database open, which fails loudly on its own if anything worse is wrong; here nothing else will ever say a word, and what is left behind is exactly the defect this method exists to prevent.
- **`!future.isCancelled()` is redundant with `!future.isDone()`.** Left as is - it states the two conditions the predicate means independently of `ScheduledFuture`'s current relationship between them, and costs nothing.

## Review round 4 - a real finding, fixed at a different place than suggested

The reviewer was right about the fact and I was wrong to have waved it away in round 2. Verified independently:

- `LocalSchema.database` is `final`, and `LocalDatabase.openInternal()` constructs the schema at `LocalDatabase.java:2862` with `wrappedDatabaseInstance`, which at that instant is still `this` - `RaftReplicatedDatabase`'s constructor calls `setWrappedDatabaseInstance(this)` on an already-open `LocalDatabase`, long after. So `schema.getDatabase()` is the raw `LocalDatabase`, whose `isReplicated()` is the `DatabaseInternal` default `false`.
- Consequence: `runMaintenance`'s `if (dbInternal.isReplicated() && !dbInternal.isLeader()) return;` **never fired for any TimeSeries type on any node** - not just repaired ones. Both pre-existing call sites (`LocalSchema.readConfiguration()` and `TimeSeriesTypeBuilder`) pass the same field.

Where I disagree is the severity and therefore the fix. The three operations do **not** run independently on followers today: `TimeSeriesShard.compactInternal` and `TimeSeriesEngine.runSealedMaintenanceReplicated` both reach `runWithCompactionReplication` through `database.getWrappedDatabaseInstance()`, resolved **at call time**, so they already decline on a follower. `runMaintenance`'s own comment says as much - the skip is there to avoid the wasted `DatabaseContext` init and transaction churn, not to prevent divergence. So this was dead optimization, not a correctness hole.

That also rules out the suggested fix of passing `db`. `db` is the server's *wrapper*, and wrappers are replaceable; the scheduler holds a `WeakReference`, so a task pinned to a superseded wrapper would cancel itself the moment that wrapper became garbage - i.e. the very bug this PR fixes, by another route. The instance that must be held is the stable one.

**Fix**: `runMaintenance` now resolves the replication view per tick, the same idiom the operations underneath it already use:

```java
final DatabaseInternal replicationView = dbInternal.getWrappedDatabaseInstance();
if (replicationView != null && replicationView.isReplicated() && !replicationView.isLeader())
  return;
```

This makes the skip work for the new call site *and* both pre-existing ones, keeps the scheduled reference stable, and needs no follow-up issue. The choice of `schema.getDatabase()` in `scheduleMaintenanceAfterRepair` is now documented in its javadoc with this reasoning rather than left implicit.

**New test** (engine): `Issue6948FollowerSkipResolvesWrapperTest` installs a wrapper through `LocalDatabase.setWrappedDatabaseInstance()` exactly as the HA server installs `RaftReplicatedDatabase` - answering itself for `getWrappedDatabaseInstance()` and delegating everything else to the real database, inventing only the two replication flags the engine module cannot get from a real Raft node. Three arms: follower does not compact, leader does, standalone does. Negative control: with the per-tick resolution reverted to the scheduled reference, the follower arm fails and the other two pass.

- [x] `mvn -o -pl engine test -Dtest='TimeSeries*Test,Issue6839InitEngineRetryableTest,Issue6948*Test'` - 216 tests, 0 failures.
- [x] `mvn -o -pl ha-raft test -Dtest='ArcadeStateMachine*Test,Issue6839TsSealedBlobRecoveryTest,Issue4416Sliced*Test,Issue6948MaintenanceAfterRepairTest'` - 131 tests, 0 failures.

## Summary by CodeRabbit

- **New Features**
  - Added a way to check whether time-series maintenance is currently scheduled for a type.

- **Bug Fixes**
  - Prevented maintenance from running on replicas when the database is not the active leader.
  - Automatically rescheduled maintenance after successful time-series repair and engine recovery.
  - Preserved correct scheduling behavior across repeated repairs, reopenings, and redundant data updates.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesMaintenanceScheduler.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesMaintenanceScheduler.java
@@ -145,6 +145,20 @@ public class TimeSeriesMaintenanceScheduler {
   }
 
   /**
+   * Whether a maintenance task is currently registered and live for the given type name.
+   * <p>
+   * Exists because "this type is maintained" has no other observable: the recurring task logs nothing when it
+   * finds nothing to do, and its three effects - {@code compactAll()}, {@code applyRetention()} and
+   * {@code applyDownsampling()} - are invisible until enough data has accumulated for one of them to change
+   * something. A caller that has just made a type usable again (issue #6948: the HA sealed-store repair) needs to
+   * be able to assert that it also made it maintained.
+   */
+  public boolean isScheduled(final String typeName) {
+    final ScheduledFuture<?> future = tasks.get(typeName);
+    return future != null && !future.isCancelled() && !future.isDone();
+  }
+
+  /**
    * Cancels the maintenance task for a specific type.
    */
   public void cancel(final String typeName) {
--- a/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesMaintenanceScheduler.java
+++ b/engine/src/main/java/com/arcadedb/engine/timeseries/TimeSeriesMaintenanceScheduler.java
@@ -105,8 +105,18 @@ public class TimeSeriesMaintenanceScheduler {
     // result to followers. Followers running it independently would diverge. The individual
     // engine operations also self-gate via runWithCompactionReplication, but skipping here avoids
     // the wasted DatabaseContext init + transaction churn on every follower tick.
+    //
+    // The two flags MUST be read off getWrappedDatabaseInstance() and not off the scheduled reference
+    // (issue #6948). Every schedule() call site hands over the instance LocalSchema was built with, and
+    // LocalSchema's `database` field is final and captured inside LocalDatabase.openInternal() - before the
+    // server has wrapped the database for HA, and never updated afterwards. So the scheduled reference is the
+    // raw LocalDatabase, whose isReplicated() is the DatabaseInternal default `false`, and this skip could
+    // never fire for anyone. getWrappedDatabaseInstance() resolves the CURRENT wrapper on every tick, which is
+    // the same idiom the operations below already use to reach runWithCompactionReplication - which is why the
+    // consequence was wasted work on every follower tick rather than divergence.
     final DatabaseInternal dbInternal = (DatabaseInternal) db;
-    if (dbInternal.isReplicated() && !dbInternal.isLeader())
+    final DatabaseInternal replicationView = dbInternal.getWrappedDatabaseInstance();
+    if (replicationView != null && replicationView.isReplicated() && !replicationView.isLeader())
       return;
 
     // The maintenance thread is created by a ScheduledExecutorService and does not
--- a/engine/src/test/java/com/arcadedb/engine/timeseries/Issue6948FollowerSkipResolvesWrapperTest.java
+++ b/engine/src/test/java/com/arcadedb/engine/timeseries/Issue6948FollowerSkipResolvesWrapperTest.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.engine.timeseries;
+
+import com.arcadedb.TestHelper;
+import com.arcadedb.database.DatabaseInternal;
+import com.arcadedb.database.LocalDatabase;
+import com.arcadedb.schema.LocalTimeSeriesType;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Proxy;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Issue #6948: {@code runMaintenance}'s leader-only skip has to read {@code isReplicated()}/{@code isLeader()} off
+ * the CURRENT wrapper, not off the reference the task was scheduled with.
+ * <p>
+ * Every {@code schedule()} call site hands over the instance {@code LocalSchema} was built with, and that field is
+ * {@code final}, captured inside {@code LocalDatabase.openInternal()} - which runs before the server wraps the
+ * database for HA, and is never updated afterwards. So the scheduled reference is the raw {@code LocalDatabase},
+ * whose {@code isReplicated()} is the {@code DatabaseInternal} default {@code false}, and the skip could not fire
+ * for anybody. The cost was wasted work rather than divergence, because the three operations underneath already
+ * resolve the wrapper per call to reach {@code runWithCompactionReplication} - but the skip exists precisely to
+ * avoid that wasted work, and it was doing nothing.
+ * <p>
+ * The wrapper here is installed exactly the way the HA server installs {@code RaftReplicatedDatabase}: through
+ * {@code LocalDatabase.setWrappedDatabaseInstance()}, with the wrapper answering itself for
+ * {@code getWrappedDatabaseInstance()} and delegating everything else to the real database. The two flags are the
+ * only behaviour it invents, because they are the only ones the engine module cannot get from a real Raft node.
+ *
+ * @author Roberto Franchini (r.franchini@arcadedata.com)
+ */
+class Issue6948FollowerSkipResolvesWrapperTest extends TestHelper {
+
+  private static final String TYPE_NAME = "Reading";
+
+  /**
+   * Runs BEFORE {@code TestHelper.afterTest()} (JUnit runs a subclass's {@code @AfterEach} first), which matters:
+   * the harness's closing integrity check reports a warning for a database that says it is replicated, and this
+   * test's whole point is a database that says exactly that.
+   */
+  @AfterEach
+  void removeWrapper() {
+    if (database != null && database.isOpen())
+      ((LocalDatabase) database).setWrappedDatabaseInstance((LocalDatabase) database);
+  }
+
+  /**
+   * A follower does no maintenance at all. Observed through compaction: the mutable rows are still in the mutable
+   * bucket afterwards, because nothing sealed them.
+   */
+  @Test
+  void aFollowerSkipsMaintenanceEvenThoughItWasScheduledWithTheRawInstance() throws Exception {
+    final LocalTimeSeriesType tsType = createTypeWithMutableRows();
+    installWrapper(true, false);
+
+    TimeSeriesMaintenanceScheduler.runMaintenance(database, tsType, TYPE_NAME);
+
+    assertThat(tsType.getEngine().getShard(0).getSealedStore().getBlockCount())
+        .as("a follower must not compact: the leader ships the sealed store").isZero();
+  }
+
+  /**
+   * The arm that proves the one above is not passing because maintenance is broken outright: the same call on a
+   * leader does compact. Also covers the standalone case indirectly - {@code isReplicated() == false} takes the
+   * same branch as a leader.
+   */
+  @Test
+  void aLeaderStillRunsMaintenance() throws Exception {
+    final LocalTimeSeriesType tsType = createTypeWithMutableRows();
+    installWrapper(true, true);
+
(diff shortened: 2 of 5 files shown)


robfrank:
Final review is clean - no blocking items. Recording the one deferred follow-up so it does not get lost with the thread.

**Deferred (out of scope for this PR): the same stale-reference pattern in `GraphAnalyticalView`.** Three sites - `GraphAnalyticalView.java:1903`, `:1984`, `:2151` - use the identical `dbInternal.isReplicated() && !dbInternal.isLeader()` shape, and one of them says in its own comment that it was modeled on `TimeSeriesMaintenanceScheduler.runMaintenance()`, i.e. on the check this PR just proved was reading a never-wrapped reference.

Checked, but deliberately not changed here: the lineage is different. `GraphAnalyticalView.database` is whatever the caller hands `GraphAnalyticalView(Database)` / the builder (`GraphAnalyticalView.java:322/326/328`), not a field captured inside `LocalDatabase.openInternal()`, so it may already be the wrapper on every path that matters. It is also a materially different failure mode - the view's own javadoc calls a follower persisting a certificate "permanent rather than self-healing", where the defect fixed here cost only wasted follower ticks. That combination - different lineage, worse consequence, needs an audit of every construction path under HA rather than a one-line change - is a separate piece of work with its own tests, not something to fold into a TimeSeries maintenance fix.

Issue #4958, opened 2026-07-04
[storage] Low-severity storage/WAL cleanups (short read/write loops, stats, staleness)


lvca:
Grouped low-severity findings in the storage/WAL layer:

- [ ] **WALFile.readLong/readInt (`WALFile.java:442-452`)** use shared instance `bufferLong`/`bufferInt` with no sync and no short-read loop; a partial `channel.read` yields stale garbage. Loop until full or use per-call stack buffers.
- [ ] **WALFile.append (`:454-457`)** issues a single `channel.write` with no `hasRemaining()` loop; a partial write leaves a torn record that the #4508 gap detector then flags as corruption. `PaginatedComponentFile` loops correctly - add a shared `writeFully`/`readFully` helper.
- [ ] **TransactionManager.getStats (`:354-369`)** adds each active WAL file`s cumulative counters into the stats accumulators on every call (double-counted at retirement), inflating numbers with polling frequency; also NPEs when `activeWALFilePool==null` (read-only DBs).
- [ ] **activeWALFilePool array (`TransactionManager.java:55,190,769`)** elements written by the timer thread and read by committers with no volatile/fence; a JMM-stale element can stall a committer up to `WRITE_WAL_TIMEOUT`. Use `AtomicReferenceArray`/VarHandle.
- [ ] **PageManager.getCachedPage (`:567-592`)** returns before `cacheMiss.incrementAndGet()`, so cache-miss stats are ~always zero (hit/miss ratio meaningless).
- [ ] **WALFile segment (`:200-226`)** `deltaSize` from the page header is unvalidated vs remaining file size; a corrupt header can trigger a transient ~2GB allocation during recovery. Clamp to `getSize()-pos`.
- [ ] **Preserved corrupt WAL on next open (`:331-346`,`:747-758`)** `createWALFilePool` resets `logFileCounter` to 0 and opens the preserved `txlog_0.wal` without truncation; `append` writes after corrupt content and old records replay again. Rename preserved files to `.wal.corrupt` or seed the counter past existing files.
- [ ] **suspendFlushAndExecute (`:142-152`)** silently skips the callback when already suspended (nested backup/snapshot); `waitAllPagesOfDatabaseAreFlushed` loops forever if invoked while suspended.
[...]

Issue #4508, opened 2026-06-08
`WALFile.getTransaction` returns null on partial record, recovery abandons rest of WAL


ruispereira:
**Affected version:** 26.4.2 (confirmed still present in 26.5.1)
**Component:** `com.arcadedb.engine.WALFile` / `TransactionManager`

### Summary

A torn write in the middle of one transaction record (header valid but
segment truncated, or torn magic number) makes `getTransaction` return
`null`. The recovery loop interprets `null` as "end of WAL" and abandons
later valid transactions in the same file.

With `TX_WAL_FILES > 1` (default = number of cores) WAL files are written
in parallel; a torn record in one of them ends WAL replay there even
though subsequent records are intact.

### Code

`engine/com/arcadedb/engine/WALFile.java:148–222`
`engine/com/arcadedb/engine/TransactionManager.java:230–253`

### Suggested fix

Distinguish "torn record" from "physical EOF": scan forward for the next
plausible record header (txId monotonic, plausible magic-number alignment)
or fail loudly. Log SEVERE with the file/offset.

lvca:
This was quite internal, thanks @ruispereira ! Fixed now.

ruispereira:
No problem :)

PR #6962, opened 2026-08-31
fix(#6932): bound the WAL transaction page count against the declared segment size

### LINKED ISSUE #6932: A corrupt WAL page-count field lets recovery allocate an unbounded WALPage[] and die on an Error the recovery guard cannot catch
### Summary
WAL recovery reads a transaction's segment count straight from the file and allocates a `WALPage[]` of that size with no sanity check. The neighbouring per-page guard was added by #4958 for exactly this hazard and says so; the outer count has the same exposure and no guard. The resulting `OutOfMemoryError` is an `Error`, so the `catch (Exception)` recovery guard cannot catch it.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/engine/WALFile.java:167` and `:177`, in `getTransaction`:
```java
final int pages = readInt(pos);
...
tx.pages = new WALPage[pages];
```
The guard that exists, at `:212`, added by #4958: "a corrupt per-page header can still declare a huge deltaSize ... and trigger a transient ~2GB allocation, or an OutOfMemoryError that escapes the catch(Exception) recovery guard below". At `pages = Integer.MAX_VALUE` the array alone is ~17GB of references, and `catch (final Exception e) { return null; }` (`:243`) cannot catch the `Error`.

The count is derivable and therefore checkable: `segmentSize` is already bounded against the file length at `:173`, and every segment costs at least `PAGE_HEADER_SIZE + 1` bytes, so any legal `pages` satisfies `pages <= segmentSize / (PAGE_HEADER_SIZE + 1)`. `segmentSize` itself is never checked for sign, so a negative value makes the `:173` bound pass trivially.

### Repro
`TransactionManager.checkIntegrity` walks a WAL with arbitrary bytes twice over: `getFirstTransaction`/`getTransaction(nextPos)` (`:341`, `:385`) and `findNextValidTransactionPosition` (`:344`, `:386`). The latter derives a candidate `txStart` from every 8-byte window in the file that happens to equal `MAGIC_NUMBER` - a value that can occur inside a page payload - and calls `getTransaction(txStart)` on it. So a single corrupt or torn WAL feeds `getTransaction` fully arbitrary header bytes.

`checkIntegrity` has a `try { } finally { }` with no catch (`:304`, `:457`), so the Error unwinds out of `LocalDatabase.open` before either the WAL drop or the `.corrupt` rename runs: the database cannot be opened, and the next attempt hits the same bytes. That is the same permanent open-failure loop #5063 fixed for a different cause.

`Issue4508TornWALRecoveryTest` and `Issue4958PreservedWALTest` cover the torn and preserved paths but neither injects a bogus page count.

### Suggested fix
After reading `segmentSize` at `:170`, reject `segmentSize < 0`; after reading `pages` at `:167`, reject `pages < 0 || (long) pages * (PAGE_HEADER_SIZE + 1) > segmentSize` by returning `null` - the same benign "not a transaction" answer the surrounding checks already give - before the array is allocated.

### Scope / confidence
High. Verified `getTransaction` is reached with unvalidated positions rather than only at record boundaries, and that the scan's only guards on the derived start are `segmentSize >= 0` and `txStart >= fromPos`, neither of which constrains `pages`. Searched upstream for prior art: only #4958, which fixed the per-page `deltaSize` allocation and left this one.

robfrank:
Closes #6932

## The bug

`WALFile.getTransaction` read a transaction's segment count straight off disk and allocated a `WALPage[]` of that size with no sanity check:

```java
final int pages = readInt(pos);      // corruption-controlled
...
tx.pages = new WALPage[pages];       // allocated before a single segment byte is read
```

`TransactionManager.checkIntegrity` feeds that method fully arbitrary header bytes. `findNextValidTransactionPosition` derives a candidate transaction start from *every* 8-byte window in the file that happens to equal `MAGIC_NUMBER` - a value that can occur inside a page payload - and the only guards on the derived start are `segmentSize >= 0` and `txStart >= fromPos`, neither of which constrains `pages`. So one corrupt or torn WAL is enough to reach the allocation with any count at all.

At `pages = Integer.MAX_VALUE` the array of references alone is ~17GB, and the resulting `OutOfMemoryError` is an **`Error`**: the `catch (final Exception e) { return null; }` guard at the bottom of `getTransaction` cannot catch it, and `checkIntegrity` wraps its body in a `try { } finally { }` with no `catch` either. The Error therefore unwinds out of `LocalDatabase.open` **before** either the WAL drop or the `.corrupt` rename runs - so the database cannot be opened, and the next attempt reads the same bytes and dies the same way. That is the permanent open-failure loop #5063 fixed for a different cause.

Measured, not theorised: with the test in this PR and the fix reverted, the engine surefire fork does not merely fail, it **dies** - `There was an error in the forked process / java.lang.OutOfMemoryError: Requested array size exceeds VM limit`, `Tests run: 0`.

A second, quieter hole rides along: `segmentSize` was never checked for sign. A negative value makes the truncation bound `pos + segmentSize + 8 > getSize()` pass trivially, and a hand-built record with `segmentSize = -1` parses **cleanly** today - magic number, `endPositionInLog` and all - even though `writeTransactionToBuffer` could never have produced it.

## The fix

`engine/src/main/java/com/arcadedb/engine/WALFile.java`, in `getTransaction`, before the array is allocated:

- reject `segmentSize < 0`;
- reject `pages < 0 || (long) pages * (PAGE_HEADER_SIZE + 1) > segmentSize`.

Both return `null` - the same benign "this is not a transaction" answer every surrounding check already gives, so recovery treats it exactly as it treats a torn record and the existing #4508 gap detection is unchanged.

The bound is exact rather than a guess. `segmentSize` is already bounded against the file length, and `writeTransactionToBuffer` charges every segment `PAGE_HEADER_SIZE + deltaSize` with `deltaSize >= 1` (it throws on `deltaSize < 1`), so *any* record the writer can produce satisfies `pages * (PAGE_HEADER_SIZE + 1) <= segmentSize`. Two tests pin that it is not over-tight: a record at exactly that density parses, and so does a sparser one with 16-byte deltas.

Two constant-time integer comparisons on a path that already does several `channel.read` round trips - no allocation, no locking, nothing added to any hot read or flush path.

This is the third instance of the same guard, and the last one missing: the Raft entry parser got it in #4420 (`ArcadeStateMachine.deserializeWalTransaction`) and the per-page `deltaSize` got it in #4958, whose comment names this exact hazard ("an `OutOfMemoryError` that escapes the `catch(Exception)` recovery guard below"). Only the WAL file reader's outer count was left without one.

## Test plan

New `engine/src/test/java/com/arcadedb/engine/Issue6932CorruptWALPageCountTest.java` - 7 tests, all planting the corrupt state directly as hand-written WAL bytes rather than trying to schedule a real crash:

- [x] `hugePageCountIsRejectedBeforeTheArrayIsAllocated` - `pages = Integer.MAX_VALUE`. **Kills the surefire JVM without the fix**; returns `null` with it.
- [x] `recoveryOfAWalWithACorruptPageCountDoesNotKillTheOpen` - the same WAL planted in a live database directory, driven through `TransactionManager.checkIntegrity()`. **Kills the surefire JVM without the fix**; with it, recovery completes and the database stays writable.
- [x] `negativeSegmentSizeIsRejected` - **fails without the fix** (the record parses cleanly and returns a non-null transaction).
- [x] `pageCountExceedingTheSegmentBudgetIsRejected` - a count well below `Integer.MAX_VALUE` that still cannot fit the declared segment.
- [x] `negativePageCountIsRejected` - boundary.
- [x] `aTransactionAtTheTightestLegalDensityStillParses` - `pages * (PAGE_HEADER_SIZE + 1) == segmentSize` exactly; the guard must not reject it.
- [x] `aMultiByteDeltaTransactionStillParses` - 16-byte deltas; the bound is an upper bound, never an equality.

Regression runs (`-DexcludedGroups=benchmark,vector,slow`, isolated `-Dmaven.repo.local`):

- [x] `mvn -pl engine test -Dtest='*WAL*,*Recovery*,*Transaction*,Issue4508*,Issue4958*,Issue5063*,Issue4420*,Issue6932*'` - **177 tests, 0 failures**, including `Issue4508TornWALRecoveryTest`, `Issue4958PreservedWALTest`, `WALFileGetTransactionTest`, `WALVersionGapRecoveryTest`, `WalRecoveryCorrectnessTest`, `GraphBatchWALRestoreTest` and `TimeSeriesCrashRecoveryTest`.
- [x] Full `engine` module unit suite - see below.

- [x] Full `engine` module unit suite: **13,926 tests, 0 failures, 23 skipped**.

Not run: the `server` module (port 2480 is held by a long-running local ArcadeDB instance on this machine, which turns server tests into a wall of 403s unrelated to this change). The change is confined to `engine` and `WALFile.getTransaction` has exactly one caller outside its own class, `TransactionManager` - the Raft path parses the same wire format through its own `ArcadeStateMachine.deserializeWalTransaction`, which is untouched and already carries the equivalent guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database recovery when corrupted write-ahead log records are encountered.
  * Prevented malformed transaction data from triggering excessive memory use or unsafe processing.
  * Preserved committed records and ensured databases remain available for subsequent writes after recovery.
  * Confirmed valid transactions continue to recover correctly.

* **Tests**
  * Expanded recovery coverage for corrupted and valid transaction scenarios, including verifying committed records and continued database writeability.

### DIFF
--- a/engine/src/main/java/com/arcadedb/engine/WALFile.java
+++ b/engine/src/main/java/com/arcadedb/engine/WALFile.java
@@ -170,8 +170,26 @@ public class WALFile extends LockContext {
       final int segmentSize = readInt(pos);
       pos += Binary.INT_SERIALIZED_SIZE;
 
+      if (segmentSize < 0)
+        // CORRUPT HEADER: a negative size makes the truncation bound below pass trivially, and would
+        // let a record no writer could have produced parse cleanly. INVALID.
+        return null;
+
       if (pos + segmentSize + Binary.LONG_SERIALIZED_SIZE > getSize())
         // TRUNCATED FILE
+        return null;
+
+      // #6932: `pages` comes straight off disk and this array is allocated before a single byte of the
+      // segment is read, so a corrupt count sizes the allocation. At Integer.MAX_VALUE the references
+      // alone are ~17GB and the resulting OutOfMemoryError is an Error, which the catch(Exception)
+      // recovery guard below cannot catch: it unwinds out of TransactionManager.checkIntegrity (a
+      // try/finally with no catch) before the WAL can be dropped or renamed '.corrupt', leaving the
+      // database permanently unopenable on the same bytes. The count is derivable from the already
+      // file-bounded segment size: every segment costs at least its fixed header plus a one-byte delta
+      // (writeTransactionToBuffer rejects deltaSize < 1), so any legal count fits the budget below.
+      // Same guard the Raft entry parser got in #4420 and the per-page deltaSize got in #4958.
+      if (pages < 0 || (long) pages * (PAGE_HEADER_SIZE + 1) > segmentSize)
+        // INVALID
         return null;
 
       tx.pages = new WALPage[pages];
--- a/engine/src/main/java/com/arcadedb/engine/WALFile.java
+++ b/engine/src/main/java/com/arcadedb/engine/WALFile.java
@@ -179,15 +179,13 @@ public class WALFile extends LockContext {
         // TRUNCATED FILE
         return null;
 
-      // #6932: `pages` comes straight off disk and this array is allocated before a single byte of the
-      // segment is read, so a corrupt count sizes the allocation. At Integer.MAX_VALUE the references
-      // alone are ~17GB and the resulting OutOfMemoryError is an Error, which the catch(Exception)
-      // recovery guard below cannot catch: it unwinds out of TransactionManager.checkIntegrity (a
-      // try/finally with no catch) before the WAL can be dropped or renamed '.corrupt', leaving the
-      // database permanently unopenable on the same bytes. The count is derivable from the already
-      // file-bounded segment size: every segment costs at least its fixed header plus a one-byte delta
-      // (writeTransactionToBuffer rejects deltaSize < 1), so any legal count fits the budget below.
-      // Same guard the Raft entry parser got in #4420 and the per-page deltaSize got in #4958.
+      // #6932: `pages` comes straight off disk and sizes this array before a byte of the segment is read.
+      // The OutOfMemoryError it can raise is an Error, so neither the catch(Exception) below nor
+      // TransactionManager.checkIntegrity (try/finally, no catch) can contain it: it unwinds out of
+      // LocalDatabase.open before the WAL is dropped or renamed '.corrupt', and every later open reads
+      // the same bytes. The bound is exact, not a heuristic: writeTransactionToBuffer charges every
+      // segment PAGE_HEADER_SIZE + deltaSize and rejects deltaSize < 1. Same guard as #4420 (Raft entry
+      // parser) and #4958 (per-page deltaSize).
       if (pages < 0 || (long) pages * (PAGE_HEADER_SIZE + 1) > segmentSize)
         // INVALID
         return null;
--- a/engine/src/test/java/com/arcadedb/engine/Issue6932CorruptWALPageCountTest.java
+++ b/engine/src/test/java/com/arcadedb/engine/Issue6932CorruptWALPageCountTest.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.arcadedb.engine;
+
+import com.arcadedb.TestHelper;
+import com.arcadedb.database.DatabaseInternal;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.File;
+import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+/**
+ * Regression test for issue #6932: {@code WALFile.getTransaction} read the transaction's segment count
+ * straight from the file and allocated a {@code WALPage[]} of that size with no sanity check.
+ * <p>
+ * {@code TransactionManager.checkIntegrity} feeds {@code getTransaction} fully arbitrary header bytes -
+ * {@code findNextValidTransactionPosition} derives a candidate transaction start from every 8-byte window
+ * in the file that happens to equal {@link WALFile#MAGIC_NUMBER}, a value that can occur inside a page
+ * payload - so a single corrupt or torn WAL reaches the allocation with a corruption-controlled count. At
+ * {@code pages = Integer.MAX_VALUE} the array of references alone is ~17GB, and the resulting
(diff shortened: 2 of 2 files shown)


robfrank:
Addressed the one verification item from the review.

> `recoveryOfAWalWithACorruptPageCountDoesNotKillTheOpen` plants the corrupt WAL alongside the live `TestHelper` database's own real WAL files and calls `checkIntegrity()` directly, not through `LocalDatabase.open`. [...] Worth confirming in CI this doesn't lean on recovery being idempotent in a way that just happens to work today.

Fair concern, and it was worth more than the "other tests do this too" answer (`Issue4958PreservedWALTest.corruptGapWalFilesAreRenamedForManualInspection` does exactly the same thing, so the precedent existed). The test now drives the **real** path the issue describes instead, mirroring `Issue4508TornWALRecoveryTest`: commit a record, `kill()` + `close()` the database, plant the corrupt WAL beside the real one the kill left behind, and reopen through `factory.open()` so `LocalDatabase.performRecovery` runs it. That removes the replay-idempotency question entirely - recovery is now replaying a WAL that genuinely was never applied, which is the only state this code path exists for.

Three things it asserts that the old shape could not:
- the `DB_NOT_CLOSED` callback fired, so the WAL was actually read rather than the test passing vacuously;
- the corrupt WAL was **dropped**, not renamed `.corrupt` - i.e. it was classified as "not a transaction", not as a #4508 corruption gap;
- the committed record survived recovery and the database is still writable afterwards.

The crafted record also now uses a file id that does not exist in the database (the `MISSING_FILE_ID = 990_000` convention from `Issue4508TornWALRecoveryTest`), so it stays a no-op even if some future change made those bytes parseable again.

Re-verified that the reworked test still falsifies the bug: with the guard stripped out it fails with `java.lang.OutOfMemoryError: Requested array size exceeds VM limit` escaping `factory.open()` - which is the issue's exact claim about `LocalDatabase.open`, now pinned directly rather than inferred. It also fails more cleanly than before: AssertJ's `assertThatCode` catches the `Error`, so it reports as a test failure instead of killing the surefire fork.

Re-ran after the change:
- `Issue6932CorruptWALPageCountTest` - 7/7 green.
- `-Dtest='*WAL*,*Recovery*,*Transaction*,Issue4508*,Issue4958*,Issue5063*,Issue4420*,Issue6932*'` - 177 tests, 0 failures.
- Full `engine` module unit suite - 13,926 tests, 0 failures, 23 skipped.

Nothing changed in `WALFile.java`; this commit is test-only.

robfrank:
Cycle 2. Both reviews were approvals; two items were worth acting on.

**CodeRabbit - `countType` is not ground truth.** Valid, and it is the repo's own documented rule (`engine/CLAUDE.md`): `countType` sums `LocalBucket.count()`, which returns the cached `cachedRecordCount`, not a scan. It is exactly the wrong instrument for the assertion I was making, because the thing I am proving is that recovery restored the *records* - and `performRecovery` resets that counter to -1 and recomputes it, so it can read correct independently of whether the records came back. Both assertions now go through `SELECT count(@rid) FROM CorruptWALType`, a full scan.

**Claude - the guard comment restated the PR description.** Both reviews raised this. Trimmed from nine lines to seven, keeping only what is not visible from the code: that the OOM is an `Error` and therefore escapes *two* separate guards (`catch (Exception)` here and `checkIntegrity`'s catch-less `try/finally`), and that `PAGE_HEADER_SIZE + 1` is exact rather than a safety margin because `writeTransactionToBuffer` rejects `deltaSize < 1`. Comment-only; no logic changed.

**Not actioned, with reasons:**
- **`@Tag("slow")` on `recoveryOfAWalWithACorruptPageCountDoesNotKillTheOpen`** - flagged for awareness rather than requested. The whole seven-test class runs in **0.99 s**, and that method is one kill/reopen cycle on a two-record database. Tagging it would move it into the slow lane for no benefit and, per the repo's tagging guidance, the lanes are a partition rather than overlapping labels, so it would be *removed* from the default lane where it belongs.
- **Test duplicating `WALFile`'s private layout constants** - noted by the first review as pre-existing convention (`WALFileGetTransactionTest`, `Issue4508TornWALRecoveryTest` and `Issue4958PreservedWALTest` all do the same) and explicitly not treated as a new smell. Making them visible for tests would widen the class's API surface to fix a non-problem; if the layout ever drifts these tests fail loudly rather than passing silently.

Re-ran everything after the change: `Issue6932CorruptWALPageCountTest` 7/7 green, and the full `engine` module unit suite green at **13,926 tests, 0 failures, 23 skipped**.

PR #6963, opened 2026-08-31
Fix #6943, #6944, #6945: triangle-op overlay fallback, LSM cursor allocations, embedded doc JSON

### LINKED ISSUE #6943: PartitionedTriangleOp answers 0 for the whole query when the provider cannot hand out a NeighborView (an active overlay is enough)
### Summary
`PartitionedTriangleOp` builds its partition mapping, fills it with `-1`, and returns that all-`-1` array unchanged the moment any hop's zero-copy `NeighborView` is missing. Every consumer reads `-1` as "this node belongs to no partition", so the operator returns **0** for the whole query instead of falling back. Every sibling operator has a real per-node fallback for a null view; this one does not, and `CSRCountStep` never learns it gave up.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/PartitionedTriangleOp.java:177-189`, in `buildPartitionMapping`:
```java
final int[] partition = new int[nodeCount];
Arrays.fill(partition, -1);
...
for (int h = 0; h < chainLength; h++)
  views[h] = provider.getNeighborView(partitionDirections[h], partitionEdgeTypes[h]);

for (final NeighborView v : views)
  if (v == null)
    return partition;                       // all -1
```
`execute()` (`:74-79`) then finds `knowsView == null` for the same reason and calls `countTrianglesPerNode(provider, personPartition, ...)`, whose first test per node is (`:221-225`):
```java
final int country = personPartition[u];
if (country < 0)
  continue;
```
so every node is skipped and the total is 0.

The siblings all handle a null view properly - `DegreeProductOp.java:110`, `PairHashJoinOp.java:141`, `AntiJoinChainOp.java:217`, `PropagateChainOp.java:214`. And `CSRCountStep` tests only `provider != null` (`:71-76`), so it never falls back to `op.executeOLTP`.

### Repro
Build a `GraphAnalyticalView` over a graph, then commit any change to it. From then until the next compaction the view has an active delta overlay, and `GraphAnalyticalView.getNeighborView` returns null unconditionally (`GraphAnalyticalView.java:766-768`, "Cannot provide zero-copy view when overlay is active"). Run the LDBC-shaped query the planner turns into this operator (`CypherExecutionPlan.java:5623` - a 3-cycle over one edge type plus a partition chain): the answer is 0 rather than the triangle count, silently, with no indication the fast path bailed out.

An overlay is the normal state of a view between commits, so this is not an exotic condition.

### Suggested fix
Signal the bail-out instead of returning a mapping that reads as valid data: return `null` from `buildPartitionMapping` (or a boolean alongside it) and have `execute()` fall back the way its siblings do, so `CSRCountStep` runs `op.executeOLTP`. At minimum, do not let "no view available" be indistinguishable from "no vertex is in any partition".

### Scope / confidence
High. The `-1` sentinel's meaning is fixed by the consumer at `:221-225`, and the sibling operators establish that a null view is expected to be handled rather than fatal.

### Two smaller findings in the same area

**`GraphAnalyticalView.countCommonNeighbors` reads only the base CSR** (`GraphAnalyticalView.java:2619-2643`). `countCommonForType` is the one pair accessor on this class that never consults `snap.overlay`, while both siblings do (`isConnectedForType:2494` subtracts `ov.countDeletedEdges(...)` and scans `ov.getAddedOutNeighbors(...)`; `countBetweenForType:2564` does the same as a count). So an overlay-deleted common neighbour is still counted and an overlay-added one is not. It also omits the `nodeA < snap.nodeMapping.size()` guard both siblings carry, so an overlay-added node id - legal now that `getNodeIdUpperBound()` (`:1254`) exceeds `getNodeCount()` - evaluates `offsets[nodeA + 1]` past the end of the `n+1`-long array and throws AIOOBE. This is the accessor shape #6315 closed at `getEdgeProperty` and #6769/#6775 closed at `isConnectedTo`/`countEdgesBetween`, left behind at this one. Honest caveat: `countCommonNeighbors` has no production caller today (only `GraphAnalyticalViewTest:233,234,316`), so the blast radius is the public API surface rather than a query path. *Fix:* mirror `countBetweenForType`, or return a documented "unknown" while an overlay is active the way `getMeanEdgesPerConnectedPair` (`:1105-1107`) does.

**`algo.betweenness` normalizes a directed traversal with the undirected factor** (`AlgoBetweenness.java:99`, `:185-188`). Brandes accumulates one contribution per *ordered* pair, raw maximum `(n-1)(n-2)`, but the code applies `2.0 / ((n - 1) * (n - 2))`, which is correct only if each unordered pair was counted once. On a directed graph every score doubles. With `n = 4` and a hub `c` (`a->c, b->c, d->c, c->a, c->b, c->d`), all six ordered leaf pairs route through `c`, so raw `= 6` and normalized `= 2.0`. `AlgoBetweennessTest.betweennessNormalized` asserts `isBetween(0.0, 1.0)` and passes only because its fixture is a 4-node directed path; the graph above makes it fail. NetworkX and GDS both use `1/((n-1)(n-2))` for a directed traversal. *Fix:* use `1.0 / ((n - 1) * (n - 2))` while the adjacency is `DIRECTION.OUT` (or switch to `DIRECTION.BOTH` if undirected betweenness is the intent) and update the javadoc, which documents the undirected factor verbatim. Filed as low because the javadoc at `:44` names `2/((n-1)(n-2))` explicitly, so a maintainer could call the factor intentional - but then the documented range and the test are the things that are wrong.

### LINKED ISSUE #6944: LSMTreeIndexCursor.fetchNext() allocates an ArrayList, a LinkedHashMap and a HashSet per key group, i.e. per row on a unique-index scan
### Summary
`LSMTreeIndexCursor.fetchNext()` allocates three collections per key group. On a unique-index scan a key group is one row, so a full range scan allocates an `ArrayList`, a `LinkedHashMap` and a `HashSet` per row, on the hottest read path in the engine.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/index/lsm/LSMTreeIndexCursor.java:540-579`, per iteration of the key-group loop:
```java
final List<Object> values = new ArrayList<>();          // the group's raw values
final Map<RID, Boolean> ridState = new LinkedHashMap<>(); // rid -> deleted?
...
final Set<RID> mergedRIDs = new HashSet<>();
```
The number of live RIDs per key is 1 for a unique index and small for most non-unique ones, so all three are sized far above what they hold, and all three die immediately.

### Cost
Three allocations plus their internal arrays (`ArrayList`'s `Object[]`, `LinkedHashMap`'s `Node[]` and one `Node` per put, `HashSet`'s backing `HashMap`) per emitted row, i.e. roughly 6-8 objects per row where the useful output is one `IndexCursorEntry`. On a 1M-row indexed range scan that is on the order of 7M short-lived objects, all in the young generation, against a codebase whose own CLAUDE.md states "always bear in mind PERFORMANCE ... prefer using arrays of primitives to List of Objects".

### Suggested fix
Hoist the three containers to fields and `clear()` them per group rather than reallocating - the cursor is single-threaded and each group's contents are consumed before the next group starts. For the common single-RID case, a fast path that skips the maps entirely (when the group holds one value and it is not a tombstone, emit directly) would remove the allocations altogether for unique-index scans.

I would treat the fields-and-clear version as the safe change and the fast path as the optional follow-up.

### Scope / confidence
Medium-high on the cost argument, which is a straight read of the allocation sites; the actual throughput gain depends on how much of the scan cost is page decoding versus allocation, so it is worth a JMH check before and after. Note this is the same method as the correctness issue filed separately about the transaction overlay's REMOVE entries being ignored - if both are addressed, doing the correctness fix first is the right order, since it changes what state the loop has to keep.

### LINKED ISSUE #6945: Result.toJSON() renders an embedded document in a projection as null, because valueToJSON classifies it as a RID-bearing Record
### Summary
`Result.valueToJSON` classifies an `EmbeddedDocument` under its generic `Record` branch, which serializes a record as its RID. An embedded document has no RID by construction, so the branch returns `null`: `Result.toJSON()` renders every embedded document in a projection as `null`, directly or inside a collection.

Found while re-validating #6819 against `main` - not a regression from that fix, a pre-existing defect that the #6819 repro walks past.

### Where (HEAD `55e00aaa6`)
`engine/src/main/java/com/arcadedb/query/sql/executor/Result.java:108-113` - the default `toJSON()`:
```java
default JSONObject toJSON() {
  if (isElement())
    return getElement().get().toJSON();

  final JSONObject result = new JSONObject();
  for (final String prop : getPropertyNames())
    result.put(prop, valueToJSON(getProperty(prop)));

  return result;
}
```
`Result.java:123-124` - the offending branch:
```java
} else if (val instanceof Record record) {
  return record.getIdentity() != null ? record.getIdentity().toString() : null;
}
```
`ImmutableEmbeddedDocument extends ImmutableDocument implements EmbeddedDocument` (`engine/src/main/java/com/arcadedb/database/ImmutableEmbeddedDocument.java:26`), and `EmbeddedDocument extends Document` (`:24`), so an embedded document matches `instanceof Record` and its identity is null. There is no `EmbeddedDocument` case ahead of it, and the `Iterable` branch at `:125` recurses into the same method, so a list of embedded documents becomes a list of nulls.

### Repro
Verified against `main` at `55e00aaa6`:
```java
db.getSchema().createDocumentType("Address").createProperty("city", Type.STRING);
db.getSchema().createDocumentType("Person").createProperty("address", Type.EMBEDDED).setOfType("Address");
db.newDocument("Person").set("name", "x").set("address", Map.of("city", "Rome")).save();

final Result r = db.query("sql", "SELECT address FROM Person").next();
r.getProperty("address");   // [city=Rome]   - correct
r.toJSON();                 // {"address":null}   - wrong
```
```
getProperty("address") -> [city=Rome]
toJSON()               -> {"address":null}
aliased toJSON()       -> {"aliased":null}
name+address toJSON()  -> {"name":"x","address":null}
whole record toJSON()  -> {"name":"x","address":{"city":"Rome","@cat":"d","@type":"Address"},...}
via LET toJSON()       -> {"e":null}
```
and for a collection:
```
getProperty(tags) -> [[city=A]]
toJSON()          -> {"tags":[null]}
```
The whole-record case is correct because `isElement()` short-circuits to `Document.toJSON()`; only the projected/computed path goes through `valueToJSON`. It reproduces the same way for an alias, for a multi-property projection and via `LET`, and regardless of how the embedded document was created (a plain `Map`, a `Map` carrying `@type`, or `newEmbeddedDocument`).

### Scope - narrower than it first looks
Two paths that might be assumed affected are **not**, and it is worth stating so the fix is not over-scoped:
- The HTTP query API is fine. It serializes through `JsonSerializer.serializeResult` -> `serializeObject`, which has an explicit `value instanceof Document document -> serializeDocument(document)` arm.
- The SQL `.asJSON()` method is fine, verified: `SELECT address.asJSON()` returns `{"city":"Rome","@cat":"d","@type":"Address"}`.

What is affected:
- Any embedder calling `Result.toJSON()`, a public method on the `Result` interface, which is the natural way to serialize a query row.
- `server/src/main/java/com/arcadedb/server/ai/ToolDispatcher.java:109`, a live path: `final JSONObject rowJson = new JSONObject(row.toJSON());` - reached from `AiChatHandler`. Every embedded document in a row handed to the LLM is a null, so the model is shown missing data rather than truncated data, with no indication.

### Suggested fix
Add an `EmbeddedDocument` arm ahead of the generic `Record` branch in `valueToJSON`, serializing inline:
```java
} else if (val instanceof EmbeddedDocument embedded) {
  return embedded.toJSON();
} else if (val instanceof Record record) {
  ...
```
Testing `record.getIdentity() == null` instead would also work, but naming the type is clearer about the intent: a referenced record is a RID, an embedded document is its content. Worth a regression test asserting `toJSON()` for an embedded document both bare and inside a LIST, since the existing tests only cover the record-backed `isElement()` path.

### Scope / confidence
High - reproduced, and the type hierarchy that makes the branch match is verified.

lvca:
## Summary
- Fixes #6943: `PartitionedTriangleOp` answered 0 for the whole query whenever a partition-chain `NeighborView` was unavailable (the normal state of a `GraphAnalyticalView` while a delta overlay is active), because `buildPartitionMapping` silently degraded to an all-`-1` mapping. Now falls back to the per-node `GraphTraversalProvider#getNeighborIds` walk, the same fallback already used when the triangle-edge view itself is missing.
- Same issue, two smaller findings addressed in the same area:
  - `GraphAnalyticalView#countCommonNeighbors` never consulted the delta overlay (unlike its siblings `isConnectedTo`/`countEdgesBetween`) and omitted the `nodeA`-in-base bounds guard, throwing AIOOBE for an overlay-added node id. Rewritten to delegate to the existing overlay-aware `getNeighborsFromCSR`.
  - `algo.betweenness` normalized with the undirected factor `2/((n-1)(n-2))` even though its traversal is `DIRECTION.OUT`-only; fixed to `1/((n-1)(n-2))`, matching NetworkX/GDS for directed betweenness. Updated the one pre-existing test whose hardcoded expectation baked in the old (incorrect) factor.
- Fixes #6944: `LSMTreeIndexCursor.fetchNext()` allocated an `ArrayList`, a `HashMap` and a `HashSet` per key group (i.e. per row on a unique-index range scan). Hoisted to instance fields, cleared per group instead of reallocated.
- Fixes #6945: `Result.valueToJSON` classified an `EmbeddedDocument` under the generic `Record` branch (which serializes to a RID), rendering every embedded document in a projection as `null`. Added an `EmbeddedDocument` arm ahead of it.

## Test plan
- [x] New regression test per issue, each verified to fail against pre-fix code and pass after the fix
- [x] `mvn -o -pl engine -am test` for `com.arcadedb.index.**` (1257 tests), `com.arcadedb.graph.olap.**` + `com.arcadedb.query.opencypher.procedures.algo.**` (831 tests), `com.arcadedb.query.opencypher.executor.steps.**` + `com.arcadedb.query.sql.executor.**` (1606 tests) — all green
- [x] `mvn -o -pl engine -am install -DskipTests` compiles clean

Fixes #6943
Fixes #6944
Fixes #6945

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected graph analytics for added or deleted overlay edges and nodes.
  - Improved triangle queries when partition neighbor data is unavailable.
  - Fixed directed betweenness centrality normalization.
  - Embedded documents now serialize correctly as inline JSON, including lists and aliases.
  - Improved index cursor handling across transaction-only entries, duplicates, and tombstones.
- **Performance**
  - Reduced overhead during graph neighbor counting and index cursor scans.
- **Documentation**
  - Updated graph analytics and betweenness centrality guidance.
- **Tests**
  - Added regression coverage for overlays, triangle queries, index scans, betweenness scores, and JSON serialization.

### DIFF
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -4086,8 +4086,8 @@ cluster whose stores fit inline is not exposed at all. Upgrade the followers fir
 
 `PartitionedTriangleOp` (the CSR-accelerated operator behind country-partitioned triangle-counting Cypher queries)
 built its partition mapping from `NeighborView`s and, the moment any hop's view came back null, returned the mapping
-it had already filled with the "no partition" sentinel `-1` for every node - unchanged. `GraphAnalyticalView
-#getNeighborView` returns null unconditionally while an active delta overlay is being served, which is the normal
+it had already filled with the "no partition" sentinel `-1` for every node - unchanged.
+`GraphAnalyticalView#getNeighborView` returns null unconditionally while an active delta overlay is being served, which is the normal
 state of a view between commits, not an exotic one. Every downstream consumer reads `-1` as "this node belongs to
 no partition", so the operator answered 0 for the whole query instead of falling back, with nothing to indicate the
 fast path had bailed out. The fix walks the partition chain with the provider's per-node `getNeighborIds` accessor
@@ -4112,8 +4112,8 @@ Unrelated to the overlay work but found in the same pass: `Result.valueToJSON` c
 under its generic `Record` branch, which serializes a record as its RID string - an embedded document has no RID
 by construction, so the branch returned `null` and every embedded document in a SQL projection (bare, aliased, or
 inside a list) rendered as `null` in `Result.toJSON()`, including in the AI chat tool's row serialization. It now
-has its own arm ahead of the `Record` check, serializing inline via `Document#toJSON()`. And `LSMTreeIndexCursor
-#fetchNext()` allocated a fresh `ArrayList`, `HashMap` and `HashSet` for every key group it emitted - one full set
+has its own arm ahead of the `Record` check, serializing inline via `Document#toJSON()`. And
+`LSMTreeIndexCursor#fetchNext()` allocated a fresh `ArrayList`, `HashMap` and `HashSet` for every key group it emitted - one full set
 of collections per row on a unique-index scan. The three are now instance fields, `clear()`-ed at the start of
 each group instead of reallocated; the cursor is single-threaded and every group's contents are fully consumed
 before the next one starts, so nothing outlives the reuse.
--- a/docs/release-26.9.1.md
+++ b/docs/release-26.9.1.md
@@ -4081,3 +4081,48 @@ image until it is upgraded and resynced. A leader only produces slices once a st
 cluster whose stores fit inline is not exposed at all. Upgrade the followers first.
 
 [#4416](https://github.com/ArcadeData/arcadedb/issues/4416)
+
+## A country-partitioned triangle count no longer silently answers 0 while a Graph Analytical View has an active delta overlay (#6943, #6944, #6945)
+
+`PartitionedTriangleOp` (the CSR-accelerated operator behind country-partitioned triangle-counting Cypher queries)
+built its partition mapping from `NeighborView`s and, the moment any hop's view came back null, returned the mapping
+it had already filled with the "no partition" sentinel `-1` for every node - unchanged. `GraphAnalyticalView
+#getNeighborView` returns null unconditionally while an active delta overlay is being served, which is the normal
+state of a view between commits, not an exotic one. Every downstream consumer reads `-1` as "this node belongs to
+no partition", so the operator answered 0 for the whole query instead of falling back, with nothing to indicate the
+fast path had bailed out. The fix walks the partition chain with the provider's per-node `getNeighborIds` accessor
+instead - the same fallback the triangle-edge side of this operator already used when its own view was missing -
+so an active overlay degrades to a slower but correct answer rather than a wrong one.
+
+Two related findings surfaced while auditing the surrounding code. `GraphAnalyticalView#countCommonNeighbors` was
+the one pair accessor on the class that never consulted the overlay at all - unlike `isConnectedTo` and
+`countEdgesBetween`, it read only the base CSR, so an overlay-deleted common neighbour stayed counted and an
+overlay-added one never was - and it also skipped the bounds guard its siblings carry, so a node ID the overlay had
+handed out for a newly added vertex threw `ArrayIndexOutOfBoundsException`. It now delegates to the same
+overlay-aware per-node accessor, with the original zero-copy CSR intersection kept as a fast path for the common
+no-overlay case. Separately, `algo.betweenness` normalized every score by the UNDIRECTED constant
+`2/((n-1)(n-2))` even though its traversal is directed (`DIRECTION.OUT` only); every existing shortest-path
+contribution is per ordered pair under that traversal, so the correct factor is `1/((n-1)(n-2))`, matching
+NetworkX's and Neo4j GDS's directed betweenness. **This is a behavioral change**: every `algo.betweenness({normalized:
+true})` score is now exactly half what it was, since the old value was not off by an edge case but by a constant
+2x factor. A hub-and-spokes fixture (four nodes, one hub reached by every shortest path) makes the old bug visible
+directly - the normalized score for the hub came out as 2.0, outside the procedure's own documented `[0,1]` range.
+
+Unrelated to the overlay work but found in the same pass: `Result.valueToJSON` classified an `EmbeddedDocument`
+under its generic `Record` branch, which serializes a record as its RID string - an embedded document has no RID
+by construction, so the branch returned `null` and every embedded document in a SQL projection (bare, aliased, or
+inside a list) rendered as `null` in `Result.toJSON()`, including in the AI chat tool's row serialization. It now
+has its own arm ahead of the `Record` check, serializing inline via `Document#toJSON()`. And `LSMTreeIndexCursor
+#fetchNext()` allocated a fresh `ArrayList`, `HashMap` and `HashSet` for every key group it emitted - one full set
+of collections per row on a unique-index scan. The three are now instance fields, `clear()`-ed at the start of
+each group instead of reallocated; the cursor is single-threaded and every group's contents are fully consumed
+before the next one starts, so nothing outlives the reuse.
+
+A follow-up, [#6967](https://github.com/ArcadeData/arcadedb/issues/6967), tracks a related but separate finding:
+`PartitionedTriangleOp` and its sibling `CountOp` operators (`DegreeProductOp`, `PairHashJoinOp`,
+`AntiJoinChainOp`, `PropagateChainOp`) all size arrays and loop bounds from `getNodeCount()` (the live node count)
+rather than `getNodeIdUpperBound()` (the ID space's exclusive bound), which can differ once a delta overlay has
+added or deleted nodes without renumbering - the same shape #6792 fixed in other kernels. That gap predates this
+fix and spans five files with a shared root cause, so it is being tracked and fixed separately.
+
+[#6943](https://github.com/ArcadeData/arcadedb/issues/6943), [#6944](https://github.com/ArcadeData/arcadedb/issues/6944), [#6945](https://github.com/ArcadeData/arcadedb/issues/6945)
--- a/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
+++ b/engine/src/main/java/com/arcadedb/graph/olap/GraphAnalyticalView.java
@@ -2616,23 +2616,39 @@ public class GraphAnalyticalView implements GraphTraversalProvider {
     return end - start;
   }
 
+  /**
+   * Counts, per requested direction, the neighbours nodeA and nodeB have in common for one edge type.
+   * <p>
+   * #6943: this used to read only the base CSR - the one pair accessor on this class that never consulted
+   * {@code snap.overlay}, unlike its siblings {@link #isConnectedForType} and {@link #countBetweenForType} - so an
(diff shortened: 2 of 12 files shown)


lvca:
Addressed the review feedback in f13a4e2f7:

- **CodeRabbit nitpick** (fully-qualified `IntStream` in `Issue6944CursorScratchSpaceReuseTest`): fixed, now imported.
- **claude-review performance point** (`countCommonForType` allocating via `getNeighborsFromCSR` even in the common no-overlay case): restored the original zero-copy CSR intersection as a fast path gated on `snap.overlay == null`, falling back to the overlay-aware accessor only when an overlay is active or a node id falls outside the base CSR - the exact condition #6943 needed fixed.
- **CodeRabbit "Merge Risk: Moderate"** (`PartitionedTriangleOp` sizing from `getNodeCount()` instead of `getNodeIdUpperBound()`): verified this is real, but it's **not** specific to this PR's diff - the same `provider.getNodeCount()` pattern is pre-existing across the *entire* `CountOp` family (`DegreeProductOp`, `PairHashJoinOp`, `AntiJoinChainOp`, `PropagateChainOp` all have it too, none use `getNodeIdUpperBound()`/`isNodeLive()`). Fixing it properly means touching 5 files with a shared root cause, independent of #6943/#6944/#6945, so I filed it separately as #6967 rather than folding a partial fix into this PR.

All previously-reported regression tests still pass after these changes.

lvca:
Addressed the second claude-review round in f6aef9ae4:

- Added a release-notes entry to \`docs/release-26.9.1.md\` covering all three issues, explicitly calling out the \`algo.betweenness\` normalization fix as a **behavioral change** (every normalized score is now exactly half what it was) per the review's suggestion.
- Tightened \`Issue6943CountCommonNeighborsOverlayTest#overlayAddedNodeIdDoesNotThrowArrayIndexOutOfBounds\`'s loose precondition (was a `>=` bound nearly trivially true) to assert the exact expected node id.

The remaining notes in that review (PartitionedTriangleOp fallback perf tradeoff, test coverage boundary in PartitionedTriangleOpMissingPartitionViewTest) were explicitly called out as non-blocking/cosmetic by the reviewer itself, and I agree - no action needed there. All tests still green.

Issue #6955, opened 2026-08-31
MongoDB wire: a plain-inserted 24-hex-char String _id round-trips back as an ObjectId on find


lvca:
### Summary

Found while fixing #6941/reviewing PR #6951. `executeUpsert` (fixed in #6951) now tracks whether a filter-seeded `_id` was an actual driver `ObjectId` before normalizing it to hex, so a client that upserts by a plain String `_id` that happens to be 24 hex characters gets that String type back, not a promoted `ObjectId`.

The same ambiguity is still open on the far more common plain `insertOne`/`find` path, because it isn't fixable the same way there. `MongoDBToSqlTranslator.convertIdToMongoDB` decides purely from the stored shape:

```java
private static Object convertIdToMongoDB(final Object value) {
  if (value instanceof String s && isObjectIdHex(s))
    return getObjectId(s);
  return toBsonValue(value);
}
```

`MongoDBCollectionWrapper.insertDocuments` stores a client-supplied String `_id` verbatim as a String (the `instanceof ObjectId` check there only fires for an actual driver `ObjectId` object). So:

```java
collection.insertOne(new Document("_id", "abcdef0123456789abcdef01").append("name", "x"));
collection.find().first().get("_id");
// returns an ObjectId, even though the client stored a String
```

Unlike the upsert case, there's no in-memory original-type flag to fall back on here: the insert and the later `find` are separate wire calls, potentially in separate connections, so by the time `find` reads the value back there is no way to tell "this hex string started life as a String" from "this hex string started life as an ObjectId" - both are stored identically. `MongoDBNonObjectIdTest` (added in #6951) covers an odd-length string and a non-hex 24-char string, but doesn't cover a hex-looking 24-char string via plain `insertOne`, so this has no failing test today.

### Why this isn't fixed in #6951

The upsert fix works because the original driver-typed value (`ObjectId` vs `String`) is available in memory at the point `executeUpsert` seeds the record, before normalization flattens both to the same hex string - the flag is scoped to that one method call. The insert+find case has no equivalent: fixing it for real requires persisting *some* signal of the original BSON type alongside `_id` (a marker byte/prefix, a side property, or a schema-level type tag), which is a storage-format decision affecting every document with a hex-looking String `_id`, not a narrow code fix. That's a call for a maintainer to make deliberately rather than something to bolt on as a follow-up commit.

### Options worth considering

[...]

Issue #6939, opened 2026-08-31
MongoDB wire: a document with a non-ObjectId _id (integer, or any odd-length string) makes every read of the collection throw


ruispereira:
### Summary
Every document leaving the MongoDB wrapper is passed through `convertMapToMongoDB`, which treats `_id` as an ObjectId hex string unconditionally. The insert path stores whatever the client sent verbatim, and MongoDB explicitly allows any BSON scalar as `_id`, so one document with an integer or odd-length-string `_id` makes every subsequent read of the collection throw.

### Where (HEAD `c1785ced6`)
`mongodbw/src/main/java/com/arcadedb/mongo/MongoDBToSqlTranslator.java:292`:
```java
result.put(p, "_id".equals(p) ? getObjectId((String) value) : toBsonValue(value));
```
`:326-332`:
```java
protected static ObjectId getObjectId(final String s) {
  final byte[] buffer = new byte[s.length() / 2];
  for (int i = 0; i < s.length(); i += 2) {
    buffer[i / 2] = (byte) ((Character.digit(s.charAt(i), 16) << 4) + Character.digit(s.charAt(i + 1), 16));
  }
  return new ObjectId(buffer);
}
```
Meanwhile `MongoDBCollectionWrapper.insertDocuments` (`:231-237`) converts only an actual `ObjectId` to hex and stores every other value as-is.

So for a stored `_id` that is not an even-length hex string:
- an `Integer`/`Long` throws `ClassCastException` on the `(String)` cast;
- an odd-length string throws `StringIndexOutOfBoundsException` on `s.charAt(i + 1)`;
- an even-length non-hex string silently becomes a **wrong** ObjectId, since `Character.digit` returns -1 for a non-hex char and `de.bwaldvogel.mongo.bson.ObjectId(byte[])` does no validation.

### Repro
```java
collection.insertOne(new Document("_id", 1).append("name", "x"));   // succeeds, n reported ok
collection.find().first();                                          // fails
```
`handleCommand` wraps the exception via `wireException` into a server error, so the **whole** `find` response fails rather than that one document. The collection is then permanently unreadable over the Mongo wire: every unfiltered `find` hits the poison document. Same for `insertOne(new Document("_id", "abc"))`.

No existing test inserts a client-supplied `_id`: `MongoDBObjectIdFilterTest` only ever reads back the server-generated one (`:84`).

### Suggested fix
[...]

PR #6969, opened 2026-08-31
fix: null-filter IS NULL translation and upsert ObjectId _id normalization (#6952, #6953)

### LINKED ISSUE #6952: MongoDB wire: a filter on a null-valued field never matches, because it compiles to "= :param" instead of IS NULL
### Summary

Found while fixing #6941 (upsert-by-`_id` duplicate creation). `MongoDBToSqlTranslator.buildExpression` translates any scalar filter value, including `null`, into `"field" = :param` with the parameter bound to `null`:

```java
} else {
  buffer.append(quoteFieldPath(entry.getKey()));
  buffer.append(" = ");
  buildValue(buffer, params, value);
}
```

An equality comparison against a bound `null` parameter does not match a stored `null` (or absent) property the way SQL's `IS NULL` does. So any Mongo filter of the shape `{field: null}` - a legal and not-uncommon query, since MongoDB's null-equality semantics also match a genuinely missing field - matches nothing in `find`, `update`, and `delete`, regardless of what's actually stored.

### Repro

```java
collection.updateOne(eq("_id", null), new Document("$set", new Document("v", 1)), new UpdateOptions().upsert(true));
// creates a document with _id: null

collection.updateOne(eq("_id", null), new Document("$set", new Document("v", 2)), new UpdateOptions().upsert(true));
// does NOT find the document just created; upserts a second one instead
```

Same gap applies to `find(eq("anyField", null))` and `deleteOne/deleteMany` on a null-valued filter - not specific to `_id` or to upsert.

### Suggested fix

In `buildExpression` (and the `$eq` handling that shares the same `buildValue` call), special-case a `null` value to emit `"field" IS NULL` instead of `"field" = :param`, matching MongoDB's actual null/missing-field semantics as closely as the engine's `IS NULL` allows.

### Scope / confidence

Medium-high on the mechanism (verified `buildValue`/`buildExpression` never special-case `null`, and reproduced the non-match empirically), but the fix touches filter translation shared by `find`/`update`/`delete`/`$eq`, so it deserves its own review and test pass rather than being folded into the unrelated #6941 upsert fix.

### LINKED ISSUE #6953: MongoDB wire: a full-replacement or $set upsert stores a raw ObjectId _id instead of its hex string
### Summary

Found while fixing #6941 (upsert-by-`_id` duplicate creation). `executeUpsert` normalizes an `ObjectId`-typed value to its hex string only in the filter-seeding loop (`normalizeIdValue`, applied to `q`). The two branches that apply the update document `u` do not:

```java
if (isReplacement(u)) {
  for (final Map.Entry<String, Object> entry : u.entrySet())
    record.set(entry.getKey(), entry.getValue());   // no normalizeIdValue
} else {
  applyOperatorsToDocument(record, u);               // $set also stores raw values
}
```

So an upsert whose replacement document (or `$set`) carries `_id` as a raw `ObjectId` object - rather than seeding it from the filter `q` - stores that raw object as the property value, instead of the hex-string convention every other `_id` write path in this class follows (`insertDocuments`, the filter-seeding loop, `buildValue`).

### Impact

A document created this way would have an `_id` that doesn't round-trip through `convertIdToMongoDB`/`isObjectIdHex` the way every other stored `_id` does, and would very likely fail to match later filters on `_id` (the same class of symptom #6941 fixed, just reached from the update-document side instead of the filter side).

### Suggested fix

Route both branches through the same `normalizeIdValue` helper `executeUpsert`'s filter-seeding loop already uses:

```java
if (isReplacement(u)) {
  for (final Map.Entry<String, Object> entry : u.entrySet())
    record.set(entry.getKey(), normalizeIdValue(entry.getValue()));
} else {
  applyOperatorsToDocument(record, u); // and normalize inside $set's per-field record.set as well
}
```

### Scope / confidence

Medium - the code path is real and the inconsistency is clear by inspection, but it wasn't confirmed against a live repro (a client setting `_id` via a raw `ObjectId` in a replacement/`$set` upsert, as opposed to via the filter, is a less common shape than the filter-seeded case #6941 covered). Flagged by review on PR #6951 rather than reproduced directly; worth a quick repro pass before fixing.

lvca:
## Summary
Follow-ups from PR #6951's review (#6939/#6940/#6941 batch):

- Fixes #6952: a Mongo filter of the shape `{field: null}` (or `{field: {$eq: null}}`) compiled to `"field" = :param` bound to `null`, which never matches a stored `null` or absent property the way SQL's `IS NULL` does. `buildExpression`/`$eq`/`$ne` now special-case a `null` operand as `IS [NOT] NULL` through a shared `buildEquality` helper.
- Fixes #6953: `executeUpsert`'s replacement and `$set` branches stored a client-supplied `ObjectId _id` verbatim instead of normalizing it to the hex string every other `_id` write path in the class uses, and didn't track its original BSON type for the response. Both branches now route through `normalizeIdValue` and track `idIsObjectId` the same way the filter-seeding loop already does.
- Addresses #6955 by documenting it as a known limitation rather than a code fix: a plain `insertOne` with a 24-hex-char String `_id` is indistinguishable from a real `ObjectId` once stored (both end up as the same bare hex string), and `insertOne`/`find` are separate wire calls with no in-memory flag to bridge them the way the upsert path can. Fixing it for real needs a storage-format decision (persisting the original BSON type alongside `_id`), so it's left documented (Javadoc on `convertIdToMongoDB`) with a regression test pinning today's actual behavior, rather than guessed at unilaterally.

## Test plan
- [x] `MongoDBToSqlTranslatorParamsTest`: updated the existing null-equality test (was pinning the buggy `= :param` shape) and added `$eq`/`$ne` null coverage
- [x] `MongoDBUpdateDeleteTest`: added `find`/`$ne`/`deleteMany` null-filter coverage, extended the existing null-id upsert test to prove idempotency, and added replacement/`$set` ObjectId-normalization coverage for #6953
- [x] New `MongoDBHexStringIdAmbiguityTest` pins #6955's documented-limitation behavior
- [x] `mvn -o -pl mongodbw -am test -DexcludedGroups=benchmark,slow,vector` - 129 tests, 0 failures

https://claude.ai/code/session_012iQgBExjxMBrXTZzXFRi2Y

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MongoDB upsert handling so `_id` types follow the most recent update, including replacement and `$set` operations.
  * Corrected `null` equality and inequality filtering, including missing-field behavior.
  * Fixed repeated upserts with `null` identifiers to update existing records instead of creating duplicates.
  * Improved `deleteMany` behavior for null-valued filters.
  * Improved ObjectId normalization and documented conversion behavior for 24-character hexadecimal string identifiers.

* **Tests**
  * Added regression coverage for ObjectId round-tripping, null filters, deletion, and upsert scenarios.

### DIFF
--- a/mongodbw/src/main/java/com/arcadedb/mongo/MongoDBDatabaseWrapper.java
+++ b/mongodbw/src/main/java/com/arcadedb/mongo/MongoDBDatabaseWrapper.java
@@ -643,12 +643,17 @@ public class MongoDBDatabaseWrapper implements MongoDatabase {
     if (isReplacement(u)) {
       for (final Map.Entry<String, Object> entry : u.entrySet()) {
         final Object value = entry.getValue();
-        if ("_id".equals(entry.getKey()) && value instanceof ObjectId)
-          idIsObjectId = true;
+        // Last write wins, matching record.set() itself: a replacement re-specifying "_id" overrides whatever type
+        // the filter seeded, rather than only ever promoting it to true and leaking a stale ObjectId flag if the
+        // replacement's own _id is some other type.
+        if ("_id".equals(entry.getKey()))
+          idIsObjectId = value instanceof ObjectId;
         record.set(entry.getKey(), normalizeIdValue(value));
       }
     } else {
-      idIsObjectId |= applyOperatorsToDocument(record, u);
+      final Boolean setIdIsObjectId = applyOperatorsToDocument(record, u);
+      if (setIdIsObjectId != null)
+        idIsObjectId = setIdIsObjectId;
     }
 
     // has(), not get() == null: an explicit "_id": null in the filter/update is a legal (if unusual) BSON _id and
@@ -676,10 +681,13 @@ public class MongoDBDatabaseWrapper implements MongoDatabase {
    * Applies the update operators to the record being upserted, returning whether an ObjectId-typed {@code _id} was
    * seeded through {@code $set} - mirroring the same tracking the filter-seeding loop and the replacement branch in
    * {@link #executeUpsert} already do, so the response reports {@code _id} back as an ObjectId regardless of which
-   * branch supplied it.
+   * branch supplied it. {@code null} means {@code $set} never touched {@code _id} at all, so the caller should keep
+   * whatever the filter-seeding loop already determined instead of overriding it - unlike a plain {@code boolean}
+   * OR, this lets a non-{@code ObjectId} {@code $set}-supplied {@code _id} correctly override (rather than leak
+   * through) a {@code true} the filter already seeded.
    */
-  private boolean applyOperatorsToDocument(final MutableDocument record, final Document u) {
-    boolean idIsObjectId = false;
+  private Boolean applyOperatorsToDocument(final MutableDocument record, final Document u) {
+    Boolean idIsObjectId = null;
     for (final Map.Entry<String, Object> entry : u.entrySet()) {
       final String op = entry.getKey();
       final Document operand = (Document) entry.getValue();
@@ -687,8 +695,8 @@ public class MongoDBDatabaseWrapper implements MongoDatabase {
       case "$set" -> {
         for (final Map.Entry<String, Object> f : operand.entrySet()) {
           final Object value = f.getValue();
-          if ("_id".equals(f.getKey()) && value instanceof ObjectId)
-            idIsObjectId = true;
+          if ("_id".equals(f.getKey()))
+            idIsObjectId = value instanceof ObjectId;
           record.set(f.getKey(), normalizeIdValue(value));
         }
       }
--- a/mongodbw/src/main/java/com/arcadedb/mongo/MongoDBDatabaseWrapper.java
+++ b/mongodbw/src/main/java/com/arcadedb/mongo/MongoDBDatabaseWrapper.java
@@ -641,10 +641,14 @@ public class MongoDBDatabaseWrapper implements MongoDatabase {
       }
 
     if (isReplacement(u)) {
-      for (final Map.Entry<String, Object> entry : u.entrySet())
-        record.set(entry.getKey(), entry.getValue());
+      for (final Map.Entry<String, Object> entry : u.entrySet()) {
+        final Object value = entry.getValue();
+        if ("_id".equals(entry.getKey()) && value instanceof ObjectId)
+          idIsObjectId = true;
+        record.set(entry.getKey(), normalizeIdValue(value));
+      }
     } else {
-      applyOperatorsToDocument(record, u);
+      idIsObjectId |= applyOperatorsToDocument(record, u);
     }
 
     // has(), not get() == null: an explicit "_id": null in the filter/update is a legal (if unusual) BSON _id and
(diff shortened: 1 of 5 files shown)


lvca:
Addressed the automated review feedback in 217d305c7:

- **claude-review** / **CodeRabbit**: fixed the fully-qualified `org.bson.types.ObjectId` reference in `MongoDBHexStringIdAmbiguityTest` to use the import instead.
- **claude-review**: added a dedicated test (`findByNullValuedFieldAlsoMatchesADocumentWhereTheFieldIsAbsent`) proving `{field: null}` also matches a document where the field is absent, not just explicitly `null` - closing the other half of the Mongo-parity claim.
- **claude-review**: added a regression test (`replaceOneUpsertNormalizesANonIdObjectIdFieldToo`) for the scope note that `normalizeIdValue` now applies to every field, not just `_id`.
- **CodeRabbit**: corrected `buildEquality`'s Javadoc - `{field: {$ne: null}}` is **not** the exact negation of `{field: null}`. Per MongoDB's own docs, `$ne: null` excludes a missing field too (only matches a field that exists and is non-null), rather than matching it the way negating `{field: null}` might suggest. The implementation (`IS NOT NULL`) was already correct for this, since ArcadeDB also evaluates a missing property as `null` - only the doc comment was wrong.
- **CodeRabbit**: on the suggestion to extend the `$ne`/`deleteMany` null tests with an absent-field fixture - I extended both, but for `$ne` I asserted the field is **excluded** (not matched), per the semantics correction above; CodeRabbit's own suggested assertion ("verify... affect both documents" for `$ne`) would have pinned incorrect behavior. `deleteMany` uses plain equality, so there the absent-field document is correctly asserted as also removed.

All 5 mongodbw test files pass locally (`mvn -o -pl mongodbw -am test`).

lvca:
Addressed in e6dd4b11a:

- **claude-review**: strengthened \`neNullExcludesTheStoredNullAndAnAbsentFieldButMatchesEverythingElse\` with a \`countDocuments(...) == 1\` assertion before \`.first()\`, so it pins exclusivity instead of relying on result ordering to hide a regression.

All other points from this review round (correctness, test coverage, style, performance/security) were reported clean with no action needed.

lvca:
Addressed both points from this review round in e000b6b1f:

- **claude-review**: fixed the \`idIsObjectId\` edge case - a \`$set\` overwriting a filter-seeded ObjectId \`_id\` with a plain string now correctly reports the string back, instead of leaking a stale \`true\` from the filter-seeding loop. Both the replacement branch and \`applyOperatorsToDocument\` now assign based on whichever value last touched \`_id\` (matching \`record.set()\`'s own last-write-wins semantics), and \`applyOperatorsToDocument\` returns a nullable \`Boolean\` so "never touched \`_id\`" and "touched it with a non-ObjectId" are distinguishable. Added a regression test for this exact case (\`setUpsertOverridingAFilterSeededObjectIdWithAPlainStringReportsTheStringNotTheObjectId\`).
- **claude-review**: strengthened \`replaceOneUpsertNormalizesANonIdObjectIdFieldToo\` to assert directly on the stored \`ref\` field's value (the plain hex string), not just that the filter round-trip happens to work.

All other points (correctness of the null-equality fix, #6955 documentation approach, test isolation, style/performance/security) were reported clean.

lvca:
Addressed in a626fb31a:

- **CodeRabbit**: added \`replaceOneUpsertOverridingAFilterSeededObjectIdWithAPlainStringReportsTheStringNotTheObjectId\`, mirroring the existing \`$set\`-branch coverage for the last-write-wins \`_id\` override, so both branches that can override a filter-seeded \`_id\` are now regression-tested.

**claude-review**'s pass on the previous commit reported nothing blocking (flagged one pre-existing, explicitly out-of-scope gap - \`$in\`/\`$nin\` with a \`null\` element - which it said is fine to leave for a future issue).

lvca:
CI note: the \`unit-tests\` lane failed, but not from this PR's changes. The failure is entirely in \`com.arcadedb.server.http.ws.Issue5024WebSocketEventBusConcurrencyTest\` (server module, WebSocket event bus concurrency stress test):

\`\`\`
java.lang.NullPointerException: Cannot invoke "io.undertow.websockets.core.WebSocketChannel.send(io.undertow.websockets.core.WebSocketFrameType)" because "wsChannel" is null
	at io.undertow.websockets.core.WebSockets.sendInternal(WebSockets.java:906)
	at com.arcadedb.server.http.ws.WebSocketEventBus.lambda$publish$6(WebSocketEventBus.java:181)
	at com.arcadedb.server.http.ws.Issue5024WebSocketEventBusConcurrencyTest.publishSerializesEventOncePerCall(...)
\`\`\`

This PR touches only \`mongodbw\` (\`MongoDBToSqlTranslator\`, \`MongoDBDatabaseWrapper\`, and their tests); it has no path to this WebSocket event bus race. Treating this as pre-existing, unrelated CI flakiness, not blocking. Waiting on the remaining lanes (\`slow-unit-tests\`, \`integration-tests\`, \`ha-integration-tests\`) to confirm the rest of CI is green.

lvca:
All CI lanes are now green except \`unit-tests\`, which remains a confirmed pre-existing, unrelated failure (WebSocketEventBus NPE race in \`Issue5024WebSocketEventBusConcurrencyTest\`, server module - no code path from this PR's mongodbw-only changes touches it).

Both \`claude-review\` and \`CodeRabbit\` reported no blocking issues on the final commit (a626fb31a). Ready for human review/merge.

PR #6973, opened 2026-09-01
Fix CountOps for sparse graph traversal provider node IDs

### LINKED ISSUE #6967: CountOp family (PartitionedTriangleOp, DegreeProductOp, PairHashJoinOp, AntiJoinChainOp, PropagateChainOp) sizes state from getNodeCount() instead of getNodeIdUpperBound()
### Summary
Every `CountOp` implementation under `engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/` sizes arrays and loop bounds from `GraphTraversalProvider#getNodeCount()` (the *live* node count) instead of `GraphTraversalProvider#getNodeIdUpperBound()` (the exclusive bound of the dense node-ID space). While a delta overlay is active - the normal state of a `GraphAnalyticalView` between commits, not an exotic condition - the two differ: the overlay retains the slot of every deleted node and allocates the ID of every added one above the base mapping, so the live count shrinks while the largest live ID does not move (the same shape issue #6792 fixed in other kernels).

Found while working on #6943, which fixed `PartitionedTriangleOp`'s null-`NeighborView` fallback; a PR reviewer flagged that the surrounding code (both the pre-existing fast path and the new fallback) still sizes everything from `getNodeCount()`. This is systemic across the whole operator family, not specific to one file, so it's filed separately rather than folded into that fix.

### Where (branch `fix/6943-6944-6945-triangle-lsm-embedded-json`, based on `940f9fb86`)
- `PartitionedTriangleOp.java:74` - `final int nodeCount = provider.getNodeCount();`, then used to size the `partition` int array (`buildPartitionMapping`), the thread-chunking bounds, and every `for (int u = 0; u < nodeCount; u++)` loop (`countRange`, `countTrianglesPerNode`).
- `DegreeProductOp.java:95`, `PairHashJoinOp.java:94`, `AntiJoinChainOp.java:117`, `PropagateChainOp.java:94,314` - same pattern.

### Impact
An overlay-added node (id >= `getNodeCount()`, < `getNodeIdUpperBound()`) is silently excluded from every array these operators build (`partition[nodeCount]`, degree/product arrays, etc.), and a live `NeighborView` (sized to the upper bound) handing back a neighbour ID in that range will index one of those under-sized arrays out of bounds - an `ArrayIndexOutOfBoundsException` rather than a silently wrong answer.

### Suggested fix
Mirror `GraphAnalyticalView`'s own contract: size arrays/loop bounds from `provider.getNodeIdUpperBound()`, and skip node IDs where `!provider.isNodeLive(id)` in any enumeration over the full range. `AbstractAlgoProcedure` (`engine/src/main/java/com/arcadedb/query/opencypher/procedures/algo/`) already uses `getNodeIdUpperBound()` and is a reference for the pattern within this codebase.

### Scope / confidence
Medium-high on the pattern (confirmed by reading `GraphTraversalProvider`'s own javadoc, which names this exact discrepancy and issue #6792), but I have not confirmed a concrete overlay-added-node repro against a live `GraphAnalyticalView` for each of the 5 files - worth a regression test per operator once fixed. Filed as its own issue since fixing it properly touches 5 files with a shared root cause, independent of #6943/#6944/#6945.

jjj-n:
## Summary

- size CountOp node-indexed arrays from getNodeIdUpperBound()
- skip non-live IDs while enumerating sparse provider node ID spaces
- cover all five affected CountOps with separate sparse-ID regression tests

## Testing

- 5 sparse node ID regression tests
- 62 related OpenCypher and CountOp tests

Fixes #6967

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graph query accuracy for datasets with sparse or deleted node IDs.
  * Ensured live nodes with high IDs are included in joins, traversals, degree calculations, triangle counting, and chain propagation.
  * Prevented invalid node slots from affecting query results.

* **Tests**
  * Added regression coverage for sparse node-ID scenarios across key graph query operations.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AntiJoinChainOp.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/AntiJoinChainOp.java
@@ -114,7 +114,7 @@ public final class AntiJoinChainOp implements CountOp {
 
   @Override
   public long execute(final GraphTraversalProvider provider, final Database db, final WorkGuard guard) {
-    final int nodeCount = provider.getNodeCount();
+    final int nodeIdUpperBound = provider.getNodeIdUpperBound();
     final int hops = edgeTypes.length;
 
     // Pre-compute valid bucket sets for type filtering. null = unlabelled, so no filter.
@@ -150,7 +150,7 @@ public final class AntiJoinChainOp implements CountOp {
         && antiJoinEdgeType.equals(edgeTypes[0])
         && inequalityIdxA >= 0 && inequalityIdxB >= 0
         && ineqMin == 0 && ineqMax == hops) {
-      final long result = executeEdgeScanAlgebraic(provider, nodeCount, validBuckets, guard);
+      final long result = executeEdgeScanAlgebraic(provider, nodeIdUpperBound, validBuckets, guard);
       if (result >= 0)
         return result;
     }
@@ -169,12 +169,15 @@ public final class AntiJoinChainOp implements CountOp {
 
     // Pre-compute bucket IDs for CSR-based anchor iteration and frontier filtering. A pattern that labels no
     // position needs none of them.
-    final int[] bucketIds = anyLabelled(validBuckets) ? precomputeBucketIds(provider, nodeCount, guard) : null;
+    final int[] bucketIds = anyLabelled(validBuckets)
+        ? precomputeBucketIds(provider, nodeIdUpperBound, guard) : null;
 
     long totalCount = 0;
 
-    for (int anchorId = 0; anchorId < nodeCount; anchorId++) {
+    for (int anchorId = 0; anchorId < nodeIdUpperBound; anchorId++) {
       guard.check();
+      if (!provider.isNodeLive(anchorId))
+        continue;
       if (anchorBuckets != null && !anchorBuckets.contains(bucketIds[anchorId]))
         continue;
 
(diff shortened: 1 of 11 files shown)


lvca:
@claude review

lvca:
Reviewed the fix, the linked issue #6967, and the `@claude review` finding.

**Core fix**: correct. All five `CountOp`s now size arrays/loop bounds from `getNodeIdUpperBound()` and skip holes via `isNodeLive()`, matching the pattern already used in `AbstractAlgoProcedure`. The `PropagateChainOp` change from `Arrays.fill(current, 1L)` to a per-node loop for the unlabelled-anchor case is a real fix, not just a rename — `Arrays.fill` would have seeded deleted-slot holes with `1`.

**Verified locally** (worktree off this branch): compiles clean; all 5 new `*SparseNodeIdsTest` classes pass; 74 further related tests pass (`OpenCypherCountOptimizationTest`, `AlgoTriangleCountTest`, `PartitionedTriangleOpCallerRunsTest`, `PartitionedTriangleOpMissingPartitionViewTest`, `CypherCountPushDownLabelIssue6322Test`, `CypherPairJoinLabelFilterIssue6304Test`, `CypherStarCountArmLabelIssue6337Test`, `CypherStarCountPropertyIssue6431Test`, `GAVEligibilityTest`, `Issue5790OptionalMatchCountStarUndeclaredRelTypeTest`) — no regressions.

**Test coverage gap (closed)**: all 5 new tests built their `SparseNodeIdProvider` with `.withoutViews()`, so only the per-node fallback loops were exercised — the `NeighborView`-backed fast paths this fix also touches (`buildAndProbeInline`, `executeFastScan`, `countRange`, `propagateOneHop`'s view branch), which is what a real CSR-backed provider actually returns and what production queries hit, were untested. Pushed a views-enabled variant of each test (same fixtures, provider exposes `NeighborView`s) confirming the fast paths handle sparse IDs correctly too.

**CodeRabbit's `PairHashJoinOp.java` finding**: verified against `origin/main` pre-#6973 — `buildBuckets` is computed and checked non-empty but never used to filter any CSR build path's `startId` iteration, so a live vertex outside `buildStartLabel` with qualifying arm1/arm2 edges gets overcounted. This predates this PR (this PR only widens the already-unfiltered iteration range); it's a real but separate bug, filed as #6992 rather than folded in here, since fixing it is unrelated to sparse node IDs and deserves its own regression test.

No functional changes needed here from my side — just added the view-path test coverage above. Good to merge once CI is green; #6992 is a follow-up, not a blocker for this PR's stated scope.

lvca:
Thanks @jjj-n !

PR #6980, opened 2026-09-01
fix(#6926): fullySorted() dereferenced a null ORDER BY alias, so ORDER BY @rid on a composite index threw NPE at plan time

### LINKED ISSUE #6926: fullySorted() dereferences a null ORDER BY alias, so ORDER BY @rid plus an indexed equality on a composite index throws NPE at plan time
### Summary
`fullySorted()` collects ORDER BY items with `item.getAlias()`, which is null for a record attribute such as `@rid`, then dereferences it unguarded. With a composite index the length guard does not save it, so the query fails at plan time with an NPE.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:3551`:
```java
orderItems.add(item.getAlias());
```
`OrderByItem.getAlias()` returns null whenever the item is a record attribute or a complex expression - the field is set only for the identifier form, and `getName()` exists precisely to fall back to `recordAttr` (`engine/src/main/java/com/arcadedb/query/sql/parser/OrderByItem.java:80-82`).

`SelectExecutionPlanner.java:3584-3596` carries those nulls into `orderedFields` and dereferences them:
```java
orderedFields.addAll(orderItems);

final List<String> fields = idx.getPropertyNames();
if (fields.size() < orderedFields.size()) {
  return false;                     // only saves single-property indexes
}

for (int i = 0; i < orderedFields.size(); i++) {
  final String orderFieldName = orderedFields.get(i);
  final String indexFieldName = fields.get(i);
  if (!orderFieldName.equals(indexFieldName)) {
```

### Repro
Verified against `main` at `c1785ced6`. A composite index gives `fields` enough entries to pass the size guard and reach the comparison:

```sql
CREATE INDEX ON I (p, q) NOTUNIQUE;
SELECT * FROM I WHERE p = 1 ORDER BY @rid
```
```
ACTUAL threw java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "orderFieldName" is null
```
The query fails instead of returning its rows. `getOrderDirection` returns TRUE (`OrderByItem.type` defaults to `ASC`) so `:3512` calls `fullySorted`; `orderItems = [null]`, `conditionItems = ["p"]`, `orderedFields = ["p", null]`, `fields = ["p","q"]`, and iteration `i = 1` throws.

The same shape applies to a `CASE WHEN` ORDER BY item under `SELECT *`, where `addOrderByProjections` returns early (`:894-898`) and never rewrites the item into an `_$$$ORDER_BY_ALIAS$$$_` alias.

Existing `ORDER BY @rid` tests (`SelectStatementExecutionTest.java:181`, `OrderByRidTest.java:26`) have no WHERE clause, so the index path is never entered.

### Suggested fix
Use `item.getName()` instead of `item.getAlias()` at `:3551`, and return `false` as soon as a collected name is null - an ORDER BY the planner cannot name can never be proven covered by the index order.

### Scope / confidence
High - reproduced. Verified `OrderByItem.alias` really is null for the `@rid` form (`getName()`/`compare()` at `OrderByItem.java:80-82` and `:108-120` use the `recordAttr` branch) and that `getOrderDirection` cannot return null here.

robfrank:
Closes #6926

## Problem

`SelectExecutionPlanner.fullySorted()` decides whether an index iteration already yields the order the query asked for. If it says yes, the planner sets `info.orderApplied = true` and drops the `ORDER BY` step from the plan.

It collected the ORDER BY item names with `OrderByItem.getAlias()`. That field is populated only for the plain identifier form, so it is **null** for a record attribute (`ORDER BY @rid`) and for a complex expression (`ORDER BY CASE WHEN ...`). The nulls were carried into `orderedFields` and dereferenced a few lines later:

```java
final String orderFieldName = orderedFields.get(i);
final String indexFieldName = fields.get(i);
if (!orderFieldName.equals(indexFieldName))   // NPE when orderFieldName is null
```

The `fields.size() < orderedFields.size()` guard above only rescues single-property indexes. With a **composite** index the list is long enough to reach the comparison, so the query failed at plan time:

```sql
CREATE INDEX ON I (p, q) NOTUNIQUE;
SELECT FROM I WHERE p = 1 ORDER BY @rid
-- java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "orderFieldName" is null
```

The existing `ORDER BY @rid` coverage (`OrderByRidTest`, `SelectStatementExecutionTest`) has no `WHERE` clause, so the index path was never entered.

A second, quieter defect lives in the same loop: an item carrying a **modifier** (`ORDER BY s.right(1)`, `ORDER BY s[0]`) has a non-null alias (`s`), so the method matched it against the index property `s` and declared the order satisfied. The `ORDER BY` step was then dropped and the rows came back in index order rather than in the requested order - no exception, just wrong results.

## Fix

`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java` - ask the item for the name it is actually known by (`getName()`, which falls back to `recordAttr`), and bail out of `fullySorted()` as soon as the item cannot be named or carries a modifier:

```java
final String name = item.getModifier() == null ? item.getName() : null;
if (name == null)
  return false;
```

Returning `false` is the conservative answer: the planner keeps the `ORDER BY` step and the index is still used for the fetch, so nothing is lost beyond an optimization that was never sound for these item shapes. An ORDER BY the planner cannot name, or one that orders by something derived from a property rather than by the property itself, can never be proven covered by the index iteration order.

## Test

New `engine/src/test/java/com/arcadedb/query/sql/Issue6926OrderByRecordAttributeWithIndexTest.java`. The fixture inserts the `p = 1` rows with *descending* `q`, so the index order `(p, q ASC)` is the reverse of the RID order - an assertion on RID order therefore fails if the ORDER BY step is wrongly dropped, rather than passing by accident.

Against unpatched `main` the class runs 9 tests: **5 errors** (NPE) and **2 failures** (wrong order from the modifier cases). All 9 pass with the fix.

| test | before the fix |
|---|---|
| `orderByRidWithIndexedEqualityOnCompositeIndex` | NPE |
| `orderByRidDescWithIndexedEqualityOnCompositeIndex` | NPE |
| `orderByIndexedPropertyThenRidWithIndexedEqualityOnCompositeIndex` | NPE |
| `orderByCaseExpressionWithIndexedEqualityOnCompositeIndex` | NPE |
| `indexedEqualityStillUsesTheIndexWhileTheOrderByStepIsKept` (EXPLAIN) | NPE |
| `orderByIndexedPropertyWithModifierIsNotSatisfiedByTheIndexOrder` | wrong order |
| `orderByIndexedPropertyWithModifierDescIsNotSatisfiedByTheIndexOrder` | wrong order |
| `orderByIndexedPropertiesIsStillSatisfiedByTheIndexOrder` | passes - guard, see below |
| `orderByRidWithIndexedEqualityOnSinglePropertyIndex` | passes - pins the pre-existing size-guard rescue |

The guard test asserts through `EXPLAIN` that the eligible `ORDER BY p, q` / `ORDER BY q` plans still carry **no** `ORDER BY` step, so the fix cannot be "achieved" by disabling the optimization outright. Verified armed: with `if (true) return false;` injected at the head of `fullySorted()` that test fails.

## Test plan

- [x] `mvn -o -pl engine test -Dtest=Issue6926OrderByRecordAttributeWithIndexTest` - 9/9 green with the fix; 5 errors + 2 failures with the planner reverted to `main`.
- [x] `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` - **13969 tests, 0 failures, 0 errors, 23 skipped** (re-run on the final head).
- [x] `mvn -o verify -pl gremlin,graphql,postgresw,bolt,grpcw` - the query-language and wire-protocol modules that execute through the same SQL plan: **1074 tests, 0 failures**.
- [x] `mvn -o -pl engine license:check` - clean (the new test carries the Apache 2.0 header).

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -3495,7 +3495,16 @@ public class SelectExecutionPlanner {
       } else if (!order.equals(item.getType())) {
         return false;
       }
-      orderItems.add(item.getAlias());
+
+      // getAlias() is null for a record attribute (ORDER BY @rid) and for a complex expression (ORDER BY CASE WHEN ...),
+      // so ask for the name the item is actually known by. A modifier (ORDER BY p.sub, ORDER BY p[0]) orders by something
+      // the index does not hold, so the item is not covered by the index order either. In both cases the planner cannot
+      // prove the index iteration already yields the requested order, so the ORDER BY step has to stay (issue #6926).
+      final String name = item.getModifier() == null ? item.getName() : null;
+      if (name == null)
+        return false;
+
+      orderItems.add(name);
     }
 
     final List<String> conditionItems = new ArrayList<>();
--- a/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java
@@ -3497,9 +3497,10 @@ public class SelectExecutionPlanner {
       }
 
       // getAlias() is null for a record attribute (ORDER BY @rid) and for a complex expression (ORDER BY CASE WHEN ...),
-      // so ask for the name the item is actually known by. A modifier (ORDER BY p.sub, ORDER BY p[0]) orders by something
-      // the index does not hold, so the item is not covered by the index order either. In both cases the planner cannot
-      // prove the index iteration already yields the requested order, so the ORDER BY step has to stay (issue #6926).
+      // so ask for the name the item is actually known by. A modifier orders by a value derived from the property rather
+      // than by the property itself (ORDER BY s.right(1), ORDER BY s[0]), which the index does not hold. In both cases
+      // the planner cannot prove the index iteration already yields the requested order, so the ORDER BY step has to
+      // stay (issue #6926).
       final String name = item.getModifier() == null ? item.getName() : null;
       if (name == null)
         return false;
--- a/engine/src/test/java/com/arcadedb/query/sql/Issue6926OrderByRecordAttributeWithIndexTest.java
+++ b/engine/src/test/java/com/arcadedb/query/sql/Issue6926OrderByRecordAttributeWithIndexTest.java
@@ -1,3 +1,21 @@
+/*
+ * Copyright © 2021-present Arcade Data Ltd (info@arcadedata.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-FileCopyrightText: 2021-present Arcade Data Ltd (info@arcadedata.com)
+ * SPDX-License-Identifier: Apache-2.0
+ */
 package com.arcadedb.query.sql;
 
 import com.arcadedb.TestHelper;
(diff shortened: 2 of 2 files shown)


robfrank:
Review round 3 - nothing applied, one nit skipped with rationale.

**Skipped (cosmetic, disagree with justification):** flattening `final String name = item.getModifier() == null ? item.getName() : null; if (name == null) return false;` into `if (item.getModifier() != null || item.getName() == null) return false;` plus `orderItems.add(item.getName())`. The flattened form calls `getName()` twice and splits the "can this item be named?" decision across two statements; the named local is what the comment above it refers to. The reviewer noted the current form "is also fine and arguably reads clearer with the comment above it", so I left it.

**Out of scope (noted, not filed):** no negative test for a mixed-direction multi-item ORDER BY (`ORDER BY p ASC, q DESC`). That is the `order.equals(item.getType())` branch at the top of `fullySorted()`, untouched by this PR.

Final verification on `a091c519ec`:
- `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` - 13969 tests, 0 failures, 0 errors, 23 skipped.
- `mvn -o -pl engine test -Dtest=Issue6926OrderByRecordAttributeWithIndexTest` - 9/9 green; 5 errors + 2 failures with the planner reverted to `main`.
- `mvn -o verify -pl gremlin,graphql,postgresw,bolt,grpcw` - green.
- `mvn -o -pl engine license:check` - clean.

Issue #6984, opened 2026-09-01
SQL: split and CONTAINS are not working without extra condition


gramian:
`ArcadeDB Server v26.8.1 (build 87bdc67f1f0331fa2d07e932a550064c118eae70/1785791531932/UNKNOWN)`

`Running on Linux 7.0.0-29-generic - OpenJDK 64-Bit Server VM 25.0.3`

The `CONTAINS` condition on a list created with the `split` method does not work without an additional condition, so it cannot be used as "reverse IN":

```sql
SELECT (['a','b','c'] CONTAINS 'a') -- returns correctly "true"

SELECT ('a b c'.split(' ') CONTAINS 'a') -- returns incorrectly "false"!

SELECT ('a b c'.split(' ') CONTAINS (@this = 'a')) -- returns correctly "true"

SELECT ('a b c'.split(' ') CONTAINSANY 'a') -- returns (surprisingly) "true" (scalar is interpreted as a 1-element list?)
```

lvca:
Confirmed and fixed in PR #6993. The root cause: \`split()\` returns a plain array, and \`CONTAINS\` only knew how to look inside a Collection - so \`'a b c'.split(' ') CONTAINS 'a'\` always came back \`false\`. It now works as expected.

While fixing this we found the same gap can happen on the other side of \`CONTAINS\` (an array on the right instead of the left) - tracked separately in #6995.

Fixed as of the next release.

---
❤️ Thanks for helping make ArcadeDB better! ArcadeDB is free and open source, sustained by its community. If it's useful to you or your company, please consider [becoming a sponsor](https://github.com/sponsors/ArcadeData) to keep development going.

gramian:
Works for me. Thank you!

Issue #6925, opened 2026-08-31
DISTINCT is silently ignored when the statement also has GROUP BY, UNWIND or expand()


ruispereira:
### Summary
`handleDistinct` is chained from exactly one place. The branch of `handleProjectionsBlock` taken when the statement has GROUP BY, UNWIND or `expand()` never calls it, and `info.distinct` was already consumed and cleared in `init`, so the DISTINCT keyword is accepted, dropped, and nothing downstream re-applies it.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:376-390` - the branch with no dedup:
```java
if (info.expand || info.unwind != null || info.groupBy != null) {
  handleProjections(result, info, context);
  handleExpand(result, info, context);
  handleUnwind(result, info, context);
  if (info.projectionAfterUnwind != null)
    result.chain(new ProjectionCalculationStep(info.projectionAfterUnwind, context));

  handleOrderBy(result, info, context);
  if (info.skip != null) ...
  if (info.limit != null) ...
} else {
  handleOrderBy(result, info, context);
  if (info.distinct || ...) {
    handleProjections(result, info, context);
    handleDistinct(result, info, context);      // <- the only call site
```
`grep -n handleDistinct` finds the definition at `:746` and that single call at `:395`. `info.distinct` is read off the projection in `init` (`:135-138`) and then cleared with `info.projection.setDistinct(false)`, so nothing re-applies it and no error is raised.

### Repro
Verified against `main` at `c1785ced6`.

One document whose `tag` list is `['a','b','a']`:
```sql
SELECT DISTINCT tag FROM U UNWIND tag
```
expected 2 rows, actual **3**:
```
ACTUAL rows=3 : {"tag":"a"} {"tag":"b"} {"tag":"a"}
```

Groups of size 1, 1, 2:
```sql
SELECT DISTINCT count(*) AS c FROM G GROUP BY a
```
expected 2 rows, actual **3**:
```
ACTUAL rows=3 : {"c":1} {"c":1} {"c":2}
```

`SELECT DISTINCT expand(tags) FROM T` dedups nothing either. No test in `engine/src/test/java` combines SQL DISTINCT with UNWIND, GROUP BY or expand, and the grammar accepts the combination (`SelectStatementTest.java:95` shows `group by ... unwind ...` parsing fine).

### Suggested fix
[...]

Issue #6923, opened 2026-08-31
SELECT DISTINCT ... ORDER BY ... LIMIT n truncates the sort buffer to n rows before DISTINCT dedups, returning fewer than n rows


ruispereira:
### Summary
`SELECT DISTINCT ... ORDER BY ... LIMIT n` returns fewer than `n` rows. The planner turns SKIP+LIMIT into a Top-K bound on `OrderByStep`, but `DistinctExecutionStep` is chained **downstream** of the sort, so duplicates occupy slots in the Top-K buffer and are discarded before DISTINCT ever sees them.

### Where (HEAD `c1785ced6`)
`engine/src/main/java/com/arcadedb/query/sql/executor/SelectExecutionPlanner.java:1856-1861`:
```java
Integer maxResults = null;
if (limitSize >= 0)
  maxResults = skipSize + limitSize;

if (info.expand || info.unwind != null)
  maxResults = null;              // info.distinct is not considered
```
`SelectExecutionPlanner.java:391-395` chains the steps sort-then-dedup:
```java
handleOrderBy(result, info, context);
if (info.distinct || info.groupBy != null || info.aggregateProjection != null) {
  handleProjections(result, info, context);
  handleDistinct(result, info, context);
```
`engine/src/main/java/com/arcadedb/query/sql/executor/OrderByStep.java:152-156` then truncates at each batch boundary:
```java
if (!sorted && this.maxResults != null && maxResults < cachedResult.size()) {
  cachedResult.sort(...);
  cachedResult = new ArrayList<>(cachedResult.subList(0, maxResults));
}
```

### Repro
Verified by running it against `main` at `c1785ced6`. Three documents, `department` = `Engineering`, `Engineering`, `Sales`:

```sql
SELECT DISTINCT department FROM T ORDER BY department ASC LIMIT 2
```
expected 2 rows (`Engineering`, `Sales`), actual **1 row**:
```
got: {"department":"Engineering"}
ROWS=1
```
With SKIP it is worse:
```sql
SELECT DISTINCT department FROM T ORDER BY department ASC SKIP 1 LIMIT 1
```
expected 1 row (`Sales`), actual **0 rows**.

[...]

Issue #3581, opened 2026-03-05
SQL: Left-hand side of filter being method return value errors (Regression)


gramian:
`ArcadeDB Server v26.3.1 (build 7c4856aec799a78b6f5c893f7477eba9d9708c34/1772652431081/UNKNOWN)`

`Running on Linux 6.17.0-14-generic - OpenJDK 64-Bit Server VM 25.0.2`

Using `CONTAINSANY` (and maybe other filter operators) causes an error when the left-hand side is the result of the `split` method. This was encountered for the method `split` and used to work in 26.2.1 and before.

Setup:

```sql
CREATE DOCUMENT TYPE doc;
INSERT INTO doc SET txt = 'te st';
```

Reproduce:

```sql
SELECT FROM doc WHERE txt.split(' ') CONTAINSANY 'te';
```

Returns the error `Invalid value for $current: null`

PR #6993, opened 2026-09-01
fix(#6923,#6984): DISTINCT+ORDER BY+LIMIT regression test and CONTAINS on split() array

lvca:
## Summary
- fix(#6984): `ContainsCondition.execute()` fell through to `return false` when the left-hand side was an array (e.g. the result of `split()`, `text.split()`, or Cypher `split()`), because arrays satisfy neither `instanceof Collection` nor `instanceof Iterable`. The fix normalizes an array left-hand side to a `List` up front, reusing `MultiValue`'s reflection-based iterator (already used elsewhere in this class) so it goes through the exact same CONTAINS semantics as any other Collection.
- #6923 (`SELECT DISTINCT ... ORDER BY ... LIMIT n` truncating the sort buffer before DISTINCT dedups it) turned out to already be fixed on `main` by the #6925 fix (`80424c15f`, PR #6978), which added `info.distinct` to the condition that disables the ORDER BY Top-K bound. Added a regression test that reproduces the exact repro from #6923 (plain `DISTINCT ... ORDER BY ... LIMIT` / `SKIP ... LIMIT`, without GROUP BY/UNWIND/expand) to lock that behavior in, since the existing #6925 tests only covered the GROUP BY/UNWIND/expand combinations.

## Repro (before the fix, #6984)
```sql
SELECT (['a','b','c'] CONTAINS 'a')          -- true  (already worked)
SELECT ('a b c'.split(' ') CONTAINS 'a')     -- false (bug - now true)
```

## Test plan
- [x] `ContainsConditionTest.issue6984ArrayLeftHandSide` - new unit test, verified it fails without the fix (reverted the production change, confirmed red, restored it)
- [x] `DistinctExecutionStepTest.shouldReturnDistinctWithOrderByAndLimit` - new regression test for #6923's exact repro (LIMIT and SKIP+LIMIT cases)
- [x] Full `engine` module test run for `com.arcadedb.query.sql.executor.*Test` and `com.arcadedb.query.sql.parser.**.*Test`: 2029 tests, 0 failures, 0 errors

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `CONTAINS` queries so they correctly evaluate object and primitive arrays.
  - Preserved expected matching behavior when comparing arrays with individual values.
  - Corrected `CONTAINS` queries using `split()` for matching and non-matching values.
  - Corrected `DISTINCT` query results when combined with `ORDER BY`, `LIMIT`, and `SKIP`, ensuring deduplication occurs before pagination.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ContainsCondition.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ContainsCondition.java
@@ -44,7 +44,10 @@ public class ContainsCondition extends BooleanExpression {
       // Normalize an array left-hand side (e.g. the result of split()) to a List so it gets the same
       // CONTAINS semantics as a Collection below; MultiValue's iterator covers both object and primitive
       // arrays via reflection, unlike a direct `instanceof Iterable`/`instanceof Collection` check, which
-      // an array never satisfies (issue #6984).
+      // an array never satisfies (issue #6984). Deliberately recurses into the Collection branch rather
+      // than delegating straight to MultiValue.contains(): the Collection branch also unwraps a
+      // single-item Result/Identifiable on the right-hand side (lines below), and a direct delegation
+      // would silently drop that behavior for an array left-hand side.
       final List<Object> leftAsList = new ArrayList<>(MultiValue.getSize(left));
       final Iterator<?> leftArrayIterator = MultiValue.getMultiValueIterator(left);
       while (leftArrayIterator.hasNext())
--- a/engine/src/main/java/com/arcadedb/query/sql/parser/ContainsCondition.java
+++ b/engine/src/main/java/com/arcadedb/query/sql/parser/ContainsCondition.java
@@ -40,6 +40,18 @@ public class ContainsCondition extends BooleanExpression {
   }
 
   public boolean execute(final Object left, Object right) {
+    if (left != null && left.getClass().isArray()) {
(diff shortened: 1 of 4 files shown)


lvca:
Addressed both review points:

- Added a comment explaining why the array left-hand-side normalization recurses through the `Collection` branch instead of delegating straight to `MultiValue.contains()` - the latter would drop the single-item `Result`/`Identifiable` unwrapping.
- The right-hand-side-array gap is real but pre-existing and out of scope for #6984 (which only reported the left-hand-side case); filed #6995 to track it separately.

lvca:
Addressed the one actionable ask: added `QueryTest.containsWithSplitOnLhs` / `containsWithSplitOnLhsNoMatch`, mirroring the existing `CONTAINSANY`/`CONTAINSALL` + `split()` coverage (issue #3581) so the full SQL parse-and-evaluate path is exercised too, not just `ContainsCondition.execute()` directly. Verified both fail with the fix disabled and pass with it restored.

On the other notes:
- `ContainsAnyCondition`'s lighter iterator-based approach vs. this PR's `List`-based one: intentional, as flagged - the `List` path is needed here for the `Result`/`Identifiable` unwrapping that `CONTAINSANY` doesn't need to do. Not unifying them.
- Bundling #6923's regression test with #6984's fix: kept as-is, both are small, narrowly scoped, and were explicitly assigned together.

Issue #6976, opened 2026-09-01
algo.* procedures silently coerce an unrecognised direction to BOTH instead of rejecting it


robfrank:
## Problem

`GraphEngine.parseDirection` coerces any unrecognised direction string to `BOTH`:

```java
public static Vertex.DIRECTION parseDirection(final String dir) {
  if (dir == null)
    return Vertex.DIRECTION.BOTH;
  return switch (dir.toUpperCase()) {
    case "OUT" -> Vertex.DIRECTION.OUT;
    case "IN"  -> Vertex.DIRECTION.IN;
    default    -> Vertex.DIRECTION.BOTH;   // <-- 'INCOMING', 'outgoing', 'nope', ''
  };
}
```

So `CALL algo.bfs(startNode, 3, 'INCOMING')` traverses **both** directions and returns a plausible-looking result. The caller asked one question and got the answer to a different one, with nothing to indicate the substitution. A typo in a direction is silently a semantic change, not an error.

This affects roughly 23 `algo.*`/path procedures that route their `direction` argument through the shared helper, including `algo.bfs`, `algo.dfs`, `algo.degree`, `algo.closeness`, `algo.harmonic`, `algo.eccentricity`, `algo.eigenvector`, `algo.jaccard`, `algo.commonNeighbors`, `algo.adamicAdar`, `algo.resourceAllocation`, `algo.preferentialAttachment`, `algo.totalNeighbors`, `algo.knn`, `algo.randomWalk`, `algo.node2vec`, `algo.fastrp`, `algo.hashgnn`, `algo.graphsage`, `algo.maxkcut`, `algo.labelPropagation`, `algo.dijkstra`, `algo.dijkstraSingleSource`. `AlgoLabelPropagation` also carries its own local copy of the same pattern.

## Why now

This is longstanding and was low-impact for as long as the direction values were near-interchangeable in practice. ArcadeData/arcadedb#6956 changed that for PageRank: `direction: 'IN'` had been returning `OUT`'s scores on one of the two execution paths, and once fixed the three values demonstrably produce different answers. The same PR added rejection **locally in `AlgoPageRank`** rather than tightening the shared helper, precisely because the shared change reaches ~23 other procedures and deserved its own issue. That is this issue.

The wider lesson from #6956 is the relevant one: a direction that silently degrades to something plausible is exactly how a direction bug survives unnoticed for a long time.

## Proposed fix

Reject unrecognised values by name instead of coercing:

```java
[...]

Issue #6791, opened 2026-08-27
algo.dijkstra.singleSource is the one weighted procedure that still abandons the CSR on every commit


lvca:
Follow-up from #6315 (PR #6788), raised in review of it.

#6315 made a Graph Analytical View answer edge properties exactly while a delta overlay is active, so the
weighted procedures that read them through `GraphTraversalProvider.edgeWeightsOf` - `algo.mst`, `algo.msa`,
`algo.steinerTree`, `algo.bellmanford`, `algo.apsp`, `algo.maxKCut`, `astar()`, `bellmanFord()` - keep the
columnar path across a commit instead of falling back to the edge records.

`algo.dijkstra.singleSource` does not. Its CSR path is `GraphAlgorithms.dijkstraSingleSource`, a kernel that
reads `csr.getForwardOffsets()`/`getForwardNeighbors()` and the weight column's typed arrays directly - which
is what makes it worth having, and also what makes it unable to see the overlay. It now refuses outright when
one is active (`view.hasActiveOverlay()`) and falls all the way back to the OLTP per-edge-record path.

That refusal is not a regression - before #6315 the same thing happened, just for a different reason (the
view reported no edge properties at all while an overlay was active, so the guard above the kernel never let
it in). It is simply the one caller that did not get the improvement, and against a `SYNCHRONOUS` view - where
an overlay is what every commit produces - it means this procedure reads edge records approximately always.

## What it would take

The kernel walks the CSR slices per popped node, so the shape that fits is the same one
`GraphAnalyticalView.edgeWeightsForSlice` already implements: ask the view for the node's neighbours and
their weights rather than indexing the arrays. That costs an allocation per popped node, which is exactly what
the kernel exists to avoid, so it wants measuring rather than assuming - possibly a reusable buffer variant of
`edgeWeightsForSlice`, or keeping the raw path when no overlay is active and using the slice API only when one
is.

## Worth noting

The rest of `GraphAlgorithms` (BFS, PageRank, WCC, label propagation, betweenness, ...) reads the base CSR
arrays the same way and is overlay-blind for *topology*, not merely for weights. That is a wider and older
question than this one and should not be folded into it, but anyone picking this up will meet it immediately.

lvca:
Fixed in #7001.

[...]

PR #7001, opened 2026-09-01
fix(#6976,#6791): reject unknown algo direction strings; keep Dijkstra CSR-accelerated across an overlay

lvca:
## Summary

Fixes two issues raised as follow-ups from other analysis:

### #6976 - `algo.*` procedures silently coerce an unrecognised direction to `BOTH`
`GraphEngine.parseDirection(String)` mapped any string that wasn't `OUT`/`IN` to `BOTH`, so a typo like `'INCOMING'` (or `'OUTGOING'`, empty string, trailing whitespace, ...) silently ran the algorithm with a different direction than the caller asked for, with no error. This helper backs ~23 `algo.*`/path procedures (`algo.bfs`, `algo.pagerank` OLTP-fallback callers use their own local copy for the CSR path already fixed in #6956, `algo.dijkstra`, `algo.closeness`, etc.).

- `GraphEngine.parseDirection` now rejects anything that isn't `OUT`/`IN`/`BOTH` (case-insensitive, `Locale.ROOT`-safe) with an `IllegalArgumentException`. `null` still means "use the default" (`BOTH`).
- `SQLFunctionBellmanFord` carried its own **local, case-sensitive** copy of the exact same coerce-to-`BOTH` bug (undetected by the issue's own file survey). It now delegates to `GraphEngine.parseDirection` instead of duplicating the switch.
- `AbstractPathProcedure.parseDirection` and `AlgoLabelPropagation` were checked and are already correct/already delegate - no change needed there (the issue's claim about a local copy in `AlgoLabelPropagation` no longer matches the current code).

### #6791 - `algo.dijkstra.singleSource` abandons the CSR path on every commit
Every other weighted procedure (`algo.mst`, `algo.msa`, `algo.steinerTree`, `algo.bellmanford`, `algo.apsp`, `algo.maxKCut`, `astar()`, `bellmanFord()`) resolves an active delta overlay through `GraphTraversalProvider#edgeWeightsOf` (#6315). `algo.dijkstra.singleSource`'s CSR kernel (`GraphAlgorithms.dijkstraSingleSource`) was the one that still refused outright and fell back to reading edge records - which, against a `SYNCHRONOUS` Graph Analytical View, is the state the view is in after *every single commit*.

- `GraphAlgorithms.dijkstraSingleSource` now takes the overlay-aware path per popped node when an overlay is active: instead of indexing the raw CSR arrays, each popped node's neighbours/weights come from `view.edgeWeightsOf(...)`, which already resolves additions/deletions against the columns. This costs one small allocation per node the search actually reaches (bounded by the search, not the whole graph) instead of the full per-edge OLTP record read it replaces.
- Sized against the overlay's own id-space upper bound (`DeltaOverlay.getNodeIdUpperBound()`) rather than the base node mapping (#6792), so a vertex added since the last view build is included in the result too. `AlgoDijkstraSingleSource.executeWithCSR` was doing the same base-mapping-sized enumeration and is fixed the same way.
- If a popped node can't be resolved exactly (the one case: an ambiguous parallel-edge deletion the overlay can't disambiguate on its own), the whole computation still refuses rather than guess, and the caller falls back to the OLTP path exactly as before - never a partial or plausible-wrong answer.
- Measured rather than assumed: `Issue6791DijkstraSingleSourceOverlayBenchmark` (`@Tag("benchmark")`, excluded from normal CI) times the CSR-accelerated path against the OLTP fallback it replaces on a 4000-node graph and asserts the former is faster.

## Test plan
- [x] `Issue6976UnknownDirectionRejectedTest` - unit coverage of `GraphEngine.parseDirection` (reject/accept/case-insensitivity/null-default) plus an `algo.bfs` end-to-end regression.
- [x] `SQLFunctionBellmanFordTest#rejectsAnUnrecognisedDirectionInsteadOfSilentlyTraversingBoth` - regression for the function's own local copy of the bug.
- [x] `Issue6791DijkstraSingleSourceOverlayTest` - end-to-end: CSR stays engaged (`CSR_ACCELERATED_VAR`) and answers correctly across a commit that deletes an edge and adds a vertex+edge; and the ambiguous-parallel-deletion case still falls back to OLTP with the correct answer.
- [x] `Issue6791DijkstraSingleSourceOverlayBenchmark` - confirms the CSR-accelerated path is actually faster than the OLTP fallback it replaces (stable across repeated runs).
- [x] Verified both new Dijkstra regression tests fail against the pre-fix code (not vacuous).
- [x] Full sweep of `algo`/`path`/`function.sql.graph`/`function.node`/`graph.olap`/`graph` test packages: 1255+ tests green, including the existing #6315/#6792 regression suites.
- [x] `mvn -o compile -pl '!studio' -am` - full reactor compiles clean (the `studio` module's failure is a pre-existing offline npm cache issue, unrelated to this change).

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid graph traversal directions now produce a clear error instead of silently defaulting to bidirectional traversal.
  - Direction values remain case-insensitive, with `incoming` and `outgoing` supported as aliases; omitted or null values retain documented defaults.
  - Dijkstra searches now correctly account for active analytical-view changes, including added or deleted vertices and edges.
  - Improved fallback behavior ensures accurate results when analytical data cannot fully resolve the current graph state.

### DIFF
--- a/engine/src/main/java/com/arcadedb/function/node/AbstractNodeFunction.java
+++ b/engine/src/main/java/com/arcadedb/function/node/AbstractNodeFunction.java
@@ -70,9 +70,13 @@ public abstract class AbstractNodeFunction implements StatelessFunction {
   }
 
   /**
-   * Parses direction from string.
+   * Parses direction from string. A {@code null} argument means "use the default" and returns {@code BOTH},
+   * but a present, unrecognised value is rejected rather than coerced to {@code BOTH}: IN and OUT answer
+   * genuinely different questions, so silently falling back would answer one the caller did not ask (issue
+   * #6976 - the same bug this backed for {@code node.degree()}, {@code node.relationshipExists()} and
+   * {@code node.relationshipTypes()}).
    *
-   * @param direction the direction string (in, out, both, or null)
+   * @param direction the direction string (in, incoming, out, outgoing, both, or null)
    * @return the Vertex.DIRECTION enum value
    */
   protected Vertex.DIRECTION parseDirection(final String direction) {
@@ -86,8 +90,10 @@ public abstract class AbstractNodeFunction implements StatelessFunction {
     case "out":
     case "outgoing":
       return Vertex.DIRECTION.OUT;
-    default:
+    case "both":
       return Vertex.DIRECTION.BOTH;
+    default:
+      throw new IllegalArgumentException("unknown direction '" + direction + "', expected one of IN, OUT or BOTH");
     }
   }
 }
--- a/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionBellmanFord.java
+++ b/engine/src/main/java/com/arcadedb/function/sql/graph/SQLFunctionBellmanFord.java
@@ -26,6 +26,7 @@ import com.arcadedb.function.sql.FunctionOptions;
 import com.arcadedb.function.sql.math.SQLFunctionMathAbstract;
 import com.arcadedb.graph.Edge;
 import com.arcadedb.graph.GhostEdgeReporter;
+import com.arcadedb.graph.GraphEngine;
 import com.arcadedb.graph.GraphTraversalProvider;
 import com.arcadedb.graph.GraphTraversalProviderRegistry;
 import com.arcadedb.graph.NodeEdgeWeights;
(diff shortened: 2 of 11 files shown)


lvca:
Addressed all three review points:

1. **`AbstractNodeFunction.parseDirection` gap** - fixed. It now rejects an unrecognised direction the same way `GraphEngine.parseDirection` does, keeping its existing `in`/`incoming`/`out`/`outgoing`/`both` aliases and `BOTH` default. Regression test added covering `node.degree()`, `node.relationship.exists()` and `node.relationship.types()`.
2. **Stale `AlgoPageRank` comment** - fixed, now states the real remaining reason for the local copy (its absent-value default is `OUT`, not `BOTH`).
3. **`dijkstraSingleSourceViaProvider` size-once/read-live race** - fixed. Neighbour ids returned past the end of `dist` (from a concurrent commit widening the overlay mid-search) now refuse the whole computation the same way an unresolvable node already does, instead of indexing out of bounds.

Full test sweep (1257+ tests across the affected packages) green after the changes.

PR #7016, opened 2026-09-01
fix(engine) #7009: shortestPath()/allShortestPaths() honour the pattern hop bounds

### LINKED ISSUE #7009: shortestPath()/allShortestPaths() ignore the pattern hop bounds, so [:R*..2] can return a 4-hop path
### Summary
`shortestPath()` and `allShortestPaths()` never read the hop bounds declared on the pattern's relationship, so `shortestPath((s)-[:R*..2]-(e))` happily returns a 4-hop path. The bound is parsed and then dropped.

### Where (HEAD `8cfac2923`)
`engine/src/main/java/com/arcadedb/query/opencypher/executor/steps/ShortestPathStep.java` - `grep -n "MaxHops\|MinHops\|maxDepth"` returns nothing. `computeShortestPath` (`:211-270`) builds the `params` array at `:261-263` without any depth bound, and the filtered variants (`computeFilteredShortestPath` `:533-590`, `computeFilteredAllShortestPaths` `:611-700`) do the same. `pattern.getRelationship(0).getMaxHops()` / `getMinHops()` are never consulted.

### Repro
Verified at `8cfac2923`. Chain `a -LINK-> n1 -LINK-> n2 -LINK-> n3 -LINK-> b`, so the only path from `a` to `b` is 4 hops:

```
MATCH (s:N {k:'a'}), (e:N {k:'b'}), p = shortestPath((s)-[:LINK*..2]-(e)) RETURN length(p) AS len
  -> rows=1  {"len":4}                     <- expected rows=0, the bound is 2

MATCH (s:N {k:'a'}), (e:N {k:'b'}), p = shortestPath((s)-[:LINK*]-(e)) RETURN length(p) AS len
  -> rows=1  {"len":4}                     <- control: unbounded, correctly 4
```

The bounded and unbounded forms return the same answer, which is the tell: the bound has no effect at all. A `minHops` bound is dropped the same way, so `[:R*3..]` can return a shorter path than requested.

Practical impact: the bound is how you keep a shortest-path search cheap and how you express "related within N degrees". A query written to return nothing beyond 2 hops instead returns a long path, and downstream logic that trusts the bound (a UI that renders at most N segments, a rule that treats "within 2 hops" as a decision) acts on it.

### Suggested fix
Read `getMinHops()`/`getMaxHops()` off the pattern relationship and pass them into the underlying search as a depth bound on all four paths (`computeShortestPath`, `computeFilteredShortestPath`, `computeFilteredAllShortestPaths`, and the `allShortestPaths` entry point), returning no row when the shortest path exceeds `maxHops` and skipping paths below `minHops`. `SQLFunctionShortestPath` already accepts a max-depth parameter, so the plumbing exists.

### Scope / confidence
High - reproduced, with the unbounded control in the same run demonstrating the bound changes nothing.

robfrank:
Closes #7009

## Summary

`shortestPath((s)-[:R*..2]-(e))` returned a 4-hop path: the `*min..max` quantifier was parsed into `RelationshipPattern.minHops`/`maxHops` and then never read by any shortest-path evaluator, so the bounded and unbounded spellings answered identically. This threads the bounds through every evaluator, so an out-of-range path yields no row.

## Root cause

Four evaluators compute a Cypher shortest path, and none consulted the bound:

| Evaluator | Entry point | Search |
|---|---|---|
| `MATCH p = shortestPath(...)`, unconstrained relationship | `ShortestPathStep.computeShortestPath` | delegates to `SQLFunctionShortestPath` (bidirectional BFS) |
| `MATCH p = allShortestPaths(...)`, unconstrained | `ShortestPathStep.computeAllShortestPaths` | layered BFS with a predecessor multimap |
| either form with an inline edge property map / `WHERE` | `computeFilteredShortestPath`, `computeFilteredAllShortestPaths` | edge-aware BFS |
| `RETURN shortestPath(...)` (expression position) | `ShortestPathExpression.evaluate` | `SQLFunctionShortestPath` or the edge-aware BFS |

`CypherExecutionPlanner` declines every `ShortestPathPattern` and falls back to the traditional plan, so the `MATCH` form has exactly one plan shape - confirmed with `EXPLAIN` and pinned by a test, so a future optimizer capability cannot route these queries past the fix unnoticed.

## Fix

A new `ShortestPathStep.HopBounds` value resolves `getEffectiveMinHops()`/`getEffectiveMaxHops()` once and is threaded through all four evaluators. It does two things:

- **Prunes the search.** The upper bound stops a BFS from expanding a layer that could only reach vertices further away than `maxHops`, and is handed to `SQLFunctionShortestPath` as its `maxDepth` option. `maxDepth` counts vertices rather than relationships (pinned by `SQLFunctionShortestPathTest.maxDepth1`/`maxDepth3`: `maxDepth` 11 yields an 11-vertex/10-hop path, 10 yields none), so the pattern's `maxHops` becomes `maxHops + 1` there.
- **Filters the answer.** The hop count of whatever path came back is re-checked against both bounds. This is what guarantees the contract; the pruning is only there to keep a bounded search cheap.

Two deliberate carve-outs, both documented in the code:

- **Zero-length self path.** Endpoints resolving to the same vertex short-circuit to the one-vertex path before any bound applies. `[*]` carries an implicit minimum of 1 hop that would otherwise suppress it, and `CypherReduceAndShortestPathTest.shortestPathSameNode` pins the existing answer.
- **A lower bound cannot lengthen a search.** A breadth-first walk reaches the target on its shortest layer first, so when that shortest path is below `minHops` the result is no row rather than a longer path that would satisfy the bound. Finding that path is a different (non-shortest, simple-path enumerating) search.

  **This one needs a maintainer's confirmation before merge.** An earlier revision of this description asserted that Neo4j rejects a `minHops > 1` shortest-path pattern outright. That was my recollection and I have not verified it against a running Neo4j, so treat it as unverified. If Neo4j instead searches for a path within `[min, max]` rather than only checking the global shortest, this is a silent divergence for anyone porting queries, and the right fix is a different search rather than a filter. Nothing else in the PR depends on the answer: with no lower bound declared (`[*]`, `[*..n]`, a plain relationship - i.e. every query in the linked issue) `min` is 1 or 0 and the branch is unreachable.

### One behaviour change beyond the quantifier itself

A relationship written **without** a quantifier inside `shortestPath()` parses, and it declares one hop: `getEffectiveMinHops()`/`getEffectiveMaxHops()` both answer `1` for it, exactly as they do for `ExpandPathStep`, `VarLengthExpand` and `JoinOrderRule` elsewhere in the engine. Since every `shortestPath()` traversal previously ran unbounded regardless of what the pattern declared, `shortestPath((a)-[:LINK]-(b))` used to search for a path of *any* length and now considers only a direct edge.

That is the same defect as the reported one and is fixed the same way, but it changes the answer for a shape the issue does not mention, so it is pinned deliberately by `aRelationshipWrittenWithoutAQuantifierIsASingleHop` - verified against a control worktree at the base commit, where it fails with a 2-hop answer for a pattern that asked for one hop.

`prettyPrint` now renders the bound (`-[:LINK*1..2]-`) instead of a bare `*`, so `EXPLAIN` shows what is actually enforced.

## Test plan

- [x] `CypherShortestPathHopBoundsIssue7009Test` - 17 new tests, **14 of which fail before the fix**. Covers every evaluator: `shortestPath()`/`allShortestPaths()` in `MATCH` position with and without an inline edge filter, and both in expression position; max bound, min bound and the exact `*n` form; unbounded controls in the same run so a green test cannot come from the graph simply having no path; plus the zero-length self path, the non-quantified single-hop shape and the `EXPLAIN` plan-shape assertion.
- [x] openCypher TCK (`mvn -pl engine verify -Dsurefire.includes='**/*Suite.java'`): **3897 run / 0 failures / 85 skipped, identical before and after** the change.
- [x] Whole `com.arcadedb.query.opencypher` package: 9120 tests, 0 failures.
- [x] Full `engine` module (`-DexcludedGroups=benchmark,vector,slow`): 14038 tests, 0 failures.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
@@ -186,7 +186,7 @@ public class ShortestPathExpression implements Expression {
       return allPaths ? new ArrayList<>() : null;
 
     // Endpoints resolving to the same vertex short-circuit to the zero-length path before any bound applies,
-    // matching what the MATCH form has always answered for shortestPath((a)-[:R*]-(a)).
+    // matching what the MATCH form has always answered for shortestPath((a)-[:R*]-(a)). See issue #7017.
     if (pathRids.size() > 1 && !bounds.accepts(pathRids.size() - 1))
       return allPaths ? new ArrayList<>() : null;
 
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
@@ -30,7 +30,6 @@ import com.arcadedb.function.sql.graph.SQLFunctionShortestPath;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Expression representing a shortestPath() or allShortestPaths() pattern in Cypher.
@@ -168,17 +167,9 @@ public class ShortestPathExpression implements Expression {
     else
       edgeTypeParam = edgeTypes;
 
-    // An upper hop bound also bounds the search: SQLFunctionShortestPath counts maxDepth in vertices, so the
-    // pattern's maxHops becomes maxHops + 1 there. The answer is re-checked against both bounds below.
-    final Integer maxDepth = bounds.maxDepthParameter();
-    final Object[] params;
-    if (maxDepth != null)
-      params = new Object[] { startVertex, endVertex, direction, edgeTypeParam,
-          Map.of(SQLFunctionShortestPath.PARAM_MAX_DEPTH, maxDepth) };
-    else if (edgeTypeParam != null)
-      params = new Object[] { startVertex, endVertex, direction, edgeTypeParam };
-    else
-      params = new Object[] { startVertex, endVertex, direction };
+    // An upper hop bound also bounds the search; the answer is re-checked against both bounds below.
+    final Object[] params = ShortestPathStep.shortestPathArguments(startVertex, endVertex, direction, edgeTypeParam,
+        bounds);
 
     final List<RID> pathRids = shortestPathFunction.execute(null, null, null, params, context);
 
--- a/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/ast/ShortestPathExpression.java
@@ -23,12 +23,14 @@ import com.arcadedb.exception.CommandExecutionException;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.query.opencypher.executor.steps.ShortestPathStep;
 import com.arcadedb.query.opencypher.executor.steps.ShortestPathStep.EdgeConstraint;
+import com.arcadedb.query.opencypher.executor.steps.ShortestPathStep.HopBounds;
 import com.arcadedb.query.sql.executor.CommandContext;
 import com.arcadedb.query.sql.executor.Result;
 import com.arcadedb.function.sql.graph.SQLFunctionShortestPath;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Expression representing a shortestPath() or allShortestPaths() pattern in Cypher.
@@ -139,11 +141,15 @@ public class ShortestPathExpression implements Expression {
     // The inline property map and the inline WHERE predicate constrain every relationship on the path.
     // SQLFunctionShortestPath only sees vertices, so a constrained pattern must run the edge-aware BFS
     // shared with the MATCH form instead.
+    // The *min..max hop bounds constrain the answer exactly as they do in the MATCH form: without this the
+    // expression form answered shortestPath((s)-[:R*..2]-(e)) with a 4-hop path (issue #7009).
+    final HopBounds bounds = HopBounds.from(relationship);
+
     final EdgeConstraint constraint = EdgeConstraint.from(relationship, result, context);
     if (constraint != null) {
       final String[] typesArray = edgeTypes == null || edgeTypes.isEmpty() ? null : edgeTypes.toArray(new String[0]);
       final List<Object> filtered = ShortestPathStep.computeFilteredShortestPath(startVertex, endVertex,
-          traversalDirection, typesArray, constraint, context);
+          traversalDirection, typesArray, constraint, bounds, context);
       if (filtered == null || filtered.isEmpty())
         return allPaths ? new ArrayList<>() : null;
       // allShortestPaths() in expression position still yields the single shortest path found, matching the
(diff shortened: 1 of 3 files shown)


robfrank:
Thanks - both points addressed in b83a94b.

**Nit (import `Type`)**: applied, also raised by CodeRabbit inline.

**Zero-length self path vs. a declared minimum**: you read it right, and it is a real gap rather than something the fix introduced. Filed as **#7017** with the repro, and `HopBounds`, both carve-out sites and `aZeroLengthSelfPathIsStillReturned` now point at it.

Why it was not simply fixed here: `ParserUtils.parsePathLength` lowers a bare `[*]` to `PathLength(1, null)`, so `minHops = 1` in the AST regardless of whether the author wrote `[*]` or `[*1..]`. Applying the bound to the self path therefore cannot suppress `[*3..]` without also changing the answer for `[*]` - and that answer is pinned by an existing test, `CypherReduceAndShortestPathTest.shortestPathSameNode`, which expects one row holding a single-vertex path. I share your recollection that Neo4j returns no row there, which would make the pinned expectation wrong too; that is exactly the question #7017 asks, and it needs a maintainer's call plus a change to an existing test, neither of which belongs in a bug fix. The openCypher TCK is silent on it - 3897 run / 0 failures / 85 skipped, identical with and without the bound enforcement in this PR.

**`prettyPrint` cosmetics**: kept. `EXPLAIN` now renders the bound that is actually enforced (`-[:LINK*1..]-`), which seemed worth more than the bare `*`. Nothing in the tree scrapes it: the full `engine` module is green at 14038 tests.

robfrank:
All three addressed in 3621e03.

**The non-quantified relationship**: you are right, and I verified it rather than taking it on trust - a control worktree at the base commit (`173aee462f`) with the new test copied in fails exactly as you predicted:

```
aRelationshipWrittenWithoutAQuantifierIsASingleHop:267 [the shortest c..f path is 2 hops, so a single-hop pattern matches nothing]
Expecting empty but was: [2]
```

So it is pinned now, both ways round: `g -> h` (a direct edge) still matches `-[:LINK]-`, `c -> d -> f` no longer does, and `allShortestPaths()` reads the pattern the same way. The PR description gained a section calling the change out.

**Duplicated argument block**: extracted to `ShortestPathStep.shortestPathArguments(source, target, direction, edgeTypeParam, bounds)`, now the single place that translates a hop bound into `SQLFunctionShortestPath`'s vertex-counted `maxDepth`. Worth doing for more than tidiness - "the same thing written twice in two evaluators" is the shape that let the bound reach one of them and not the other to begin with.

**`*2..2` in EXPLAIN**: `HopBounds.toString()` renders an exact bound as `*2`.

Re-verified after the refactor: openCypher TCK 3897 run / 0 failures / 85 skipped (unchanged), whole `com.arcadedb.query.opencypher` package 9120 tests / 0 failures.

robfrank:
Cycle 4 of 4 - stopping the automated loop here and handing over. Nothing in this round was blocking, and no further commits were pushed.

**On the min-hop semantics, you were right to push back and I have to correct myself.** I wrote "Neo4j rejects such a pattern outright" from recollection and did not verify it against a running Neo4j. I have removed the claim from the PR description and replaced it with an explicit request for confirmation. **The same unverified sentence is still in the `HopBounds` javadoc in `ShortestPathStep.java` and should be softened or substantiated before merge** - I did not push a fifth commit for it, so it needs either a maintainer edit or a word from you and I will do it.

What is worth knowing while deciding: the branch is unreachable for every query in the linked issue. `[*]` and `[*..n]` and a plain relationship all resolve to `min` 0 or 1, so a shortest path can never come back below the bound. It only bites for an explicitly written `[*3..]`, and only then does "no row" versus "search for a longer path" become distinguishable.

**Deferred, not done** (both offered as optional insurance, neither essential):

1. An exact `*0` bound. Note the shape is odd on its own terms: `[*0..0]` admits only the zero-length path, which only the same-node carve-out can produce, so a test would be pinning the interaction between the two carve-outs rather than the bound.
2. A bound near `Integer.MAX_VALUE`. Traced rather than tested: `[*..2147483647]` gives `max == Integer.MAX_VALUE`, which `maxDepthParameter()` intercepts and answers `null` for, so the `+1` never runs on it and the pattern reads as unbounded.
3. Hoisting `HopBounds.from()` to a field on the step instead of recomputing it per row. You noted it is consistent with the neighbouring `patternDirection()`/`patternEdgeTypesArray()`, so changing one and not the others would make the file less consistent, not more.

Verification as it stands on 3621e03: openCypher TCK 3897 run / 0 failures / 85 skipped, identical to the base commit; `com.arcadedb.query.opencypher` 9120 tests / 0 failures; full `engine` module 14038 tests / 0 failures. Merge is the developer's call.

PR #7018, opened 2026-09-02
fix(#6977): scope the Cypher label-replacement map to the statement, not the step

### LINKED ISSUE #6977: `CALL db.schema()` + `CALL (*)` label removal reports a stale edge on a valid graph
## Version

ArcadeDB Docker image `arcadedata/arcadedb:latest` (reported by the image as
`26.9.1-SNAPSHOT`, build `e823ffbe2880d366a978770180ff05f21f9d3efb`) 

## Reproduction

Start with a fresh empty database and execute this setup as one statement:

```cypher
CREATE (b:l0:l4:l6:l9:l10:l5 {k4: 'kN1omHAd', k8: 'b'})
  -[:rt5 {k3: true}]->
  (a:l11:l4 {k9: false, k8: 'a'});
```

Then execute:

```cypher
CALL db.schema() YIELD nodes, relationships
OPTIONAL MATCH (n5:l11 {k9: false}) <-[:rt5 {k3: true}]-
  (n6:l0&l4&l6&l9&l10&l5 {k4: "kN1omHAd"})
WITH *
WHERE n5 IS NOT NULL AND n6 IS NOT NULL
CALL (*) {
  REMOVE n5:l4
  RETURN n5 AS n20, n6 AS n21
}
RETURN n20, n21;
```

## Actual result

The query fails deterministically on a fresh database. The Bolt client reports
`org.neo4j.driver.exceptions.DatabaseException` (surfaced by the driver as
`Neo.DatabaseError.General.UnknownError`). The server reports a
`CommandExecutionException` whose cause is equivalent to:

```text
VertexNotFoundException: Vertex #<bucket>:0 does not exist, so it cannot be deleted:
it was reached through a reference to a record that is gone (a stale edge entry,
or a RID held past the delete). Retrying cannot make it exist: run CHECK DATABASE FIX ...
```

The numeric record IDs vary between fresh databases. The container remains
running (`EXIT=0`, `OOM=false`); this is a query failure, not a process crash.

## Controls

1. Keeping the same graph and the same `CALL (*)` removal, but removing the
   schema call (with the surrounding transaction wrapper used by the original
   case) completes successfully and returns one row.
2. Keeping `CALL db.schema()` but replacing the inner subquery with a direct
   `REMOVE n5:l4` completes successfully; the clean run returned eight rows and
   did not report a missing vertex or stale edge.

## Expected result

The graph contains exactly two vertices and one relationship, and `n5` is a
valid matched vertex. Removing its existing `l4` label should complete and
return the matched values; it should not report a missing vertex or stale edge.

robfrank:
Closes #6977

## What was wrong

A Cypher label write is not a metadata change in ArcadeDB: a record's type is the bucket it lives in, so
`REMOVE n:l4` rewrites the vertex under the reduced type, re-attaches every incident edge and **deletes the
original**. `LabelReplacements` exists to keep that survivable - it remembers `oldRID -> live replacement`, so any
row still holding the deleted original is redirected to the record that now carries its identity (issues #6312,
#6313).

That map was scoped to the running `SET`/`REMOVE`/`MERGE` **step**. `SubqueryStep.executeInnerQuery` builds a
**fresh `CypherExecutionPlan` for every outer row**, so inside a `CALL { }` body the step - and its map - is born
and dies within a single row. The second outer row entered the body still holding the vertex the first row had
deleted, and the write followed its dangling edge list straight into

```
VertexNotFoundException: Vertex #25:0 does not exist, so it cannot be deleted:
it was reached through a reference to a record that is gone (a stale edge entry, or a RID held past the delete)
```

`CALL db.schema()` is not implicated: it is simply what fanned one matched row out into eight, which is the
minimum the defect needs. Two outer rows reproduce it, and the report's own controls say the same thing - drop the
schema call (one row) or drop the subquery (one step, one map) and the query passes.

## The fix

Move the replacement map from the step to the **statement**.

- `LabelReplacements.of(context)` parks one instance on the command context, created on first use, so a read-only
  statement still allocates nothing. `SetStep`, `RemoveStep` and `MergeStep` all take theirs from there.
- `LabelReplacements.inherit(inner, outer)` carries it into a nested plan's own context.
  `CypherExecutionPlan.executeWithSeedRow` calls it under exactly the gate it already uses to share the statistics
  accumulator (`outerContext != null && !statement.isReadOnly()`), so a read-only `CALL { }`, `COUNT { }`,
  `COLLECT { }` or `EXISTS { }` body never makes the enclosing statement allocate one.
- `SubqueryStep` redirects the **outer** row through the shared map before refreshing its bindings by RID. The
  binding the body invalidated is the one `refreshDocumentBindings` could only ever fail to re-read (it swallows
  `RecordNotFoundException` and keeps the stale record), so without this the emitted row still carried a deleted
  vertex even when the write itself no longer crashed.

Held as a context *cached value* rather than a context variable: cached values are engine-internal, while
variables are reachable from a query through `$CONTEXT` and fall back to the database's global variables.

This is one write path, not two: the optimizer only ever replaces the MATCH portion of a plan, and all three
plan-building sites (`buildExecutionSteps`, `buildExecutionStepsWithOrder`,
`buildExecutionStepsWithOptimizer`) construct the same `SetStep`/`RemoveStep`/`MergeStep` against the same
context. A `CALL { }` disables the optimized physical plan outright (`canUseOptimizedPhysicalPlan()` refuses on
`hasSubquery()`), so the reported query is legacy-path by construction, and the fix lands where both shapes meet.

Allocation is not worse than before, it is better: the old code built a `LabelReplacements` (two `HashMap`s) per
outer row, the new code builds one per statement.

## Test plan

New `engine/src/test/java/com/arcadedb/query/opencypher/CypherSubqueryLabelWriteIssue6977Test.java`, six tests,
**all six fail on `main` and pass with the fix** (verified by stashing the source change and re-running):

- [x] `removeLabelInsideASubqueryOutlivesTheRowItsPlanWasBuiltFor` - three outer rows, same node; asserts the
      surviving label set, that the relationship is re-attached with its property, and that it exists exactly once
- [x] `theOuterRowFollowsTheNodeTheSubqueryMoved` - the outer alias reads the live record after the body moved it
- [x] `setLabelInsideASubqueryOutlivesTheRowItsPlanWasBuiltFor` - the `SET n:Label` mirror
- [x] `mergeOnMatchSetLabelInsideASubqueryOutlivesTheRowItsPlanWasBuiltFor` - `MERGE ... ON MATCH SET n:Label`;
      the body re-reads the node itself, so what has to survive is the imported outer alias
- [x] `labelIsRemovedOnceAcrossEveryOuterRowOfTheSubquery` - three rows, `labelsRemoved == 1`: idempotent, not
      merely non-fatal
- [x] `theReportedSchemaCallAndSubqueryRemovalCompletes` - the issue's own query, verbatim shape
- [x] Full `engine` module, `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow`: **14017 run,
      0 failures, 23 skipped** (baseline on the same worktree at `main`: 14011 run, 0 failures, 23 skipped - the
      difference is the six tests added here)
- [x] openCypher TCK, `mvn -o -pl engine verify -Dsurefire.includes='**/*Suite.java'`: **3897 scenarios, 3812
      passed, 0 failed, 85 skipped** - byte-identical to the baseline run at `main`, so no conformance movement

One unrelated flake was seen on the first full-suite run and did not reproduce:
`com.arcadedb.graph.Issue5596MergeCoverageTest.aCleanAppendConflictIsStillMerged` asserts that two racing writer
threads collide often enough for `edgeAppendMerges` to advance, and it saw zero on a loaded machine. It passes in
isolation and on the repeated full-suite run, it uses the native graph API with no Cypher anywhere in it, and
nothing in this diff leaves `com.arcadedb.query.opencypher.executor`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OpenCypher `CALL` subqueries with label changes across multiple rows and nested executions.
  * Preserved correct record references after `SET`, `REMOVE`, and `MERGE ... ON MATCH SET` operations.
  * Corrected relationship reattachment, alias resolution, label-removal statistics, and schema-query behavior in affected subqueries.
  * Ensured write operations maintain consistent record mappings across repeated subquery executions.

### DIFF
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -400,6 +400,9 @@ public class CypherExecutionPlan {
       final List<ResultInternal> allResults = new ArrayList<>();
       final Set<String> seen = removeDuplicates ? new HashSet<>() : null;
       for (final CypherExecutionPlan branchPlan : branchPlans) {
+        // No inherit() here on purpose: the branch runs through this same method with the REAL outerContext,
+        // so it takes the non-union path below and does its own inherit under the same gate. A second copy
+        // here would be one more thing to keep in step with that gate (issue #6977).
         final ResultSet rs = branchPlan.executeWithSeedRow(seedRow, outerContext);
         while (rs.hasNext()) {
           unionGuard.check();
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -425,8 +425,14 @@ public class CypherExecutionPlan {
     // Only share the outer statistics accumulator for a write CALL body: getStatistics() lazily
     // allocates, so sharing it unconditionally would allocate a QueryStatistics even for a
     // fully read-only CALL, violating the "read queries allocate nothing" constraint.
-    if (outerContext != null && !statement.isReadOnly())
+    if (outerContext != null && !statement.isReadOnly()) {
       context.setStatistics(outerContext.getStatistics());
+      // A label write moves the record, so the map of what it displaced has to outlive this plan: the body of a
+      // CALL { } is re-planned for every outer row, and without this the second row met the vertex the first one
+      // deleted (issue #6977). Same gate as the statistics above - a read-only body performs no label write and
+      // must not make the enclosing statement allocate a map for one.
+      LabelReplacements.inherit(context, outerContext);
+    }
     inheritCommandDeadline(context, outerContext);
 
     // Create a seed step that returns the seed row
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/LabelReplacements.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/LabelReplacements.java
@@ -28,6 +28,7 @@ import com.arcadedb.graph.MutableVertex;
 import com.arcadedb.graph.Vertex;
 import com.arcadedb.log.LogManager;
 import com.arcadedb.query.opencypher.traversal.TraversalPath;
+import com.arcadedb.query.sql.executor.CommandContext;
 import com.arcadedb.query.sql.executor.Result;
 import com.arcadedb.query.sql.executor.ResultInternal;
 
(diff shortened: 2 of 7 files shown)


robfrank:
Cycle 1 review addressed in 4b0ac42247. Recorded here rather than in a tracking file, since this repo keeps analysis in the PR and not in `docs/`.

**Applied**

- *claude*: no coverage for a two-level nested `CALL { CALL { ... } }`. Added `removeLabelSurvivesTwoLevelsOfNestedSubqueries`, and confirmed it fails at the parent commit like the other six - the propagation is genuinely exercised now, not merely argued.
- *claude*: `MergeStep`'s retry-snapshot comment still said the map is "step-wide". Refreshed; the snapshot stays correct because it is taken per row, which the comment now says.
- *coderabbit*: `setUp` would throw on a leftover database directory. Added the `if (factory.exists()) factory.open().drop();` guard `OpenCypherConstraintTest` already uses.

**Not applied, with reasons**

- *coderabbit*: "close the `AutoCloseable` `DatabaseFactory` after setup". `DatabaseFactory.close()` is `callbacks.clear()` and nothing else, and this factory registers no callbacks, so the close is a no-op on an empty map. No test under `engine/src/test/java/com/arcadedb/query/opencypher` closes it either; adding it here would only make this one file diverge from the convention.
- *coderabbit pre-merge check*: "Docstring coverage 21.74%". The 23 functions it counted are mostly JUnit test methods and the file's three private query helpers. Every public method this diff adds (`LabelReplacements.of`, `LabelReplacements.inherit`) carries Javadoc explaining both what it does and why the scope moved, and each test method carries the comment that says what it is pinning.
- *claude*: `LabelReplacements.of(null)` returns an uncached instance silently. Left as a fallback rather than made to throw - no call site passes null (every step is constructed with the plan's context), and turning a harmless, still-correct-for-one-step path into an NPE buys nothing. The contract is now spelled out in the Javadoc instead of being inferred.

robfrank:
Cycle 2 addressed in 82107059e1 - both items were comments, no behaviour change.

- *claude*: "the UNION branch doesn't call `inherit()` [...] worth double-checking this is intentional rather than incidental." It is intentional, and now says so at the call site: each branch is executed through `executeWithSeedRow` again with the real `outerContext`, so it takes the non-union path and inherits under the same gate. A second `inherit()` in the union arm would be one more copy of that gate to keep in step.
- *claude*: the `ResultInternal` cast in `redirect()` versus the guarded one in `refreshDocumentBindings`. I checked the premise rather than repeating it, and one part of it is not right: a `REMOVE`/`SET` **after** the `MATCH` does **not** disable the optimizer. `canUseOptimizedPhysicalPlan()` refuses on `hasWriteBeforeMatch()` and `hasSubquery()`, not on writing as such, and `EXPLAIN MATCH (a:Drop)-[:E]->(b:X) REMOVE a:Drop RETURN a` comes back as `NodeByLabelScan` + `ExpandAll` with `+ REMOVE` appended - the optimized plan, feeding `RemoveStep` across three fanned-out rows, green. So both plan shapes do reach a label write; what keeps the cast safe is that both materialize `ResultInternal`, and `GAVResult` (the fused GAV chain's zero-map row) has no path into one today. Left unguarded on purpose, with the reasoning written down: a silent skip here would leave the row pointing at the record the write deleted, which is the defect this class exists to prevent, while the refresh next door is an optimisation that is allowed to decline.

Full `engine` suite on the cycle-1 head: 14018 run, 0 failures, 23 skipped. This push adds comments only.

robfrank:
Cycle 4: no code changes. All three notes are non-blocking and I am recording why each is left as it is rather than acting on it.

1. **A regression test for the RID-reuse window.** Skipped deliberately. Such a test would pass today whether or not the hazard is real, because I have no way to make a later `CREATE` land on the exact slot `replace()` freed within one statement - and a test that cannot be made to fail is a guard in name only. Pinning it needs a deterministic slot-reuse repro, which is follow-up work of its own, not something to bolt onto this PR. The hazard is documented in the `LabelReplacements` class Javadoc, including the part that is NOT at risk (`replace()` creates the replacement before deleting the original, so the pair that moves can never collide with itself).
2. **`redirect()`'s per-row work now spans the statement.** Correct, and the trade is deliberate: the guard is `vertices.isEmpty()`, so a statement that performs no label write pays one field read per row and walks nothing, and after a label write correctness is not something a later row should be able to opt out of. This is the same reasoning the method's existing Javadoc already gives for descending into collections step-wide; the scope is wider now, the shape of the trade is not.
3. **`factory.open().drop()` without try/finally in `setUp`.** Left as is: this is verbatim the guard `OpenCypherConstraintTest` uses, and diverging from it in one new file to catch a case the `exists()` check has already ruled out would cost more in consistency than it buys.

Nothing else outstanding from cycles 1-3. Final state on `8b148d6ebf`: the seven new tests green, and the last two commits are comment/Javadoc only on top of the 14018-run / 0-failure `engine` suite.

robfrank:
Final verification on `8b148d6ebf`: `mvn -o -pl engine test -DexcludedGroups=benchmark,vector,slow` -> **14018 run, 0 failures, 23 skipped**. openCypher TCK unchanged from `main` (3897 scenarios, 3812 passed, 0 failed, 85 skipped), measured on both sides.

PR #7019, opened 2026-09-02
fix(#6922): make FOREACH eager with respect to a following graph read

### LINKED ISSUE #6922: `CALL meta.stats()` returns different `nodeCount` after a downstream `collect()`/`UNWIND`
## Summary

Adding a row-preserving `collect()`/`UNWIND` after `CALL meta.stats()` changes the value returned by `stats.nodeCount`. Both queries return the same number of rows and leave the same final graph, so the downstream materialization should not change the already-yielded statistics value.

## Version and environment

- ArcadeDB: `arcadedata/arcadedb:26.8.1`
- Protocol: Bolt
- Database: a fresh empty `Test` database for each query 

## Reproduction

Run each query independently against a fresh empty database.

### Control query

```cypher
UNWIND range(0, 100) AS i
FOREACH (j IN range(0, 9) | CREATE (:L1 {k5: j}))
CALL meta.stats() YIELD value AS stats
RETURN stats.nodeCount AS nodeCount
```

### Query with downstream materialization

```cypher
UNWIND range(0, 100) AS i
FOREACH (j IN range(0, 9) | CREATE (:L1 {k5: j}))
CALL meta.stats() YIELD value AS stats
WITH stats.nodeCount AS nodeCount
WITH collect(nodeCount) AS values
UNWIND values AS nodeCount
RETURN nodeCount
```

The only intended difference is the row-preserving `WITH`/`collect()`/`UNWIND` after `stats.nodeCount` has been yielded.

## Expected result

Both queries should return 101 rows with the same `nodeCount` values. The final graph contains 1010 `:L1` nodes, so an independent post-query check returns:

```cypher
MATCH (n) RETURN count(n) AS nodes
```

```text
nodes = 1010
```

## Actual result

With the control query, the complete returned-value frequency was:

```text
rows = 101
frequencies = {1000=100, 1010=1}
first 100 rows = 1000
last row = 1010
```

With the materialized query, the complete returned-value frequency was:

```text
rows = 101
frequencies = {1010=101}
```

Both executions ended with the same graph state:

```text
nodes = 1010
relationships = 0
```

robfrank:
Closes #6922

## Summary

`CALL meta.stats() YIELD value` placed after a `FOREACH` reported a `nodeCount` that depended on
where the pipeline's pull batches happened to fall, not on the graph. The pipeline pulls
`nRecords = 100` rows at a time (`CypherExecutionPlan`), and `ForeachStep.fetchMore()` filled that
whole 100-row buffer - running the FOREACH body, and therefore its writes, for all 100 rows - before
handing the first of those rows to the downstream step. The reader following the FOREACH therefore
observed neither the pre-FOREACH graph nor the post-FOREACH graph, but a snapshot taken at an
arbitrary batch boundary.

That is exactly the reported shape. For the issue's query the first 100 rows saw `nodeCount = 1000`
and the 101st saw `1010`, and stretching the query to 250 rows made the batch dependence explicit -
three different answers, one per batch:

```cypher
UNWIND range(1, 250) AS i
FOREACH (j IN range(0, 1) | CREATE (:L1 {i: i, j: j}))
CALL meta.stats() YIELD value AS stats
RETURN stats.nodeCount AS nodeCount
```

before: `{200 x100, 400 x100, 500 x50}` &nbsp;&nbsp; after: `{500 x250}` (the graph really does hold 500 nodes)

openCypher resolves this class of read/write conflict by making the updating clause eager with
respect to the following read, which is also what this engine's `CREATE` already does in practice
(its bulk batch is 20,000, so a realistic `UNWIND ... CREATE ... CALL ...` applies every write before
the first row is read). `FOREACH` is now consistent with both.

## The change

- **`ForeachStep`** gains an `eagerExecution` mode. When set, `fetchMore()` drops the `buffer.size() < n`
  cap and runs the FOREACH body for every upstream row before emitting the first output row, so every
  downstream row sees the same, complete post-FOREACH graph. This is separate from the existing
  `eagerMaterialize` flag, which pre-reads the *upstream* rows to protect against a FOREACH-body
  `DELETE` invalidating a row the upstream MATCH has yet to produce (#6491) and does nothing about
  *when* the writes land.
- **`CypherExecutionPlan`** arms the mode via a new `graphReadFollows()` predicate: a later `MATCH`,
  `MERGE`, `CALL` or `SUBQUERY` clause counts as a graph read, as does a later `FOREACH` whose own
  body contains a `MERGE` at any nesting depth (the grammar allows no other read inside a FOREACH
  body). `RETURN`, `WITH` and `UNWIND` are not reads in themselves, but the expressions they carry
  can be, so those clauses are scanned for an `EXISTS { }` / `COUNT { }` / `COLLECT { }` subquery
  (each runs a real query per row through `CorrelatedSubqueryRunner`), a pattern predicate, a pattern
  comprehension or a `shortestPath()`. The same scan runs over a later FOREACH's own driving list,
  which is an expression evaluated per row against the live graph. It reuses `ExpressionRewriter`'s
  traversal, hooked at its `rewrite()` entry point so it also sees the expression types the
  rewriter's own dispatch does not know; that class carries a row of `TODO` visit stubs that do not
  recurse, and they are overridden on the detector (CASE, list, map, map projection, list
  comprehension, ALL/ANY/NONE/SINGLE, reduce, list index and slice, arithmetic, IN, IS NULL, string
  match, regex, label check, function call, boolean coercion and wrapper) rather than filled in on
  the shared base class, where every other rewriter would inherit a traversal change it never asked
  for. `CALL` is treated conservatively
  because a procedure is opaque to the planner - which is what `meta.stats()` is here. `RETURN`,
  `WITH` and `UNWIND` do not count: they only project rows the pipeline has already produced, so the
  common streaming `MATCH ... FOREACH ... RETURN` bulk-update shape keeps its bounded memory profile
  and is unaffected by this PR.
- **`GlobalConfiguration.OPENCYPHER_FOREACH_EAGER_READ`** (`arcadedb.opencypher.foreachEagerRead`,
  `SCOPE.DATABASE`, default `true`) turns the eager mode off, see the trade-off section below. It is
  read off the bound database's `ContextConfiguration`, so one database's setting does not reach
  another, and it is read when the plan is built.
- The now-unused 4-argument `ForeachStep` constructor was folded into the 5-argument one.

Only the legacy execution path builds a `ForeachStep`: `CypherExecutionPlanner` returns `false` for
any statement containing `FOREACH` or `CALL`, and `EXPLAIN` on the issue's query confirms
`Using Traditional Execution (Non-Optimized)`. There is no second path to patch.

## Test plan

New `engine/src/test/java/com/arcadedb/query/opencypher/CypherForeachEagerReadIssue6922Test.java`,
13 tests. 12 of them fail with the eager mode off - i.e. on the pre-fix behaviour - and pass with it
on; the thirteenth is the negative control, which must pass either way.

- [x] `metaStatsAfterForeachIsUnaffectedByDownstreamCollectUnwind` - the issue's two queries verbatim,
      on fresh databases. Both return 101 rows, every row `1010`, and the two lists are equal, so the
      row-preserving `collect()`/`UNWIND` no longer changes an already-yielded value.
- [x] `metaStatsAfterForeachDoesNotDependOnThePullBatchSize` - 250 rows spans three pull batches;
      asserts all 250 rows report `500`, the count an independent `MATCH (n) RETURN count(n)` confirms.
- [x] `matchAfterForeachSeesEveryCreatedNode` - the same conflict reached through `MATCH` instead of
      `CALL` (returned `400` before the fix).
- [x] `mergeAfterForeachMatchesANodeTheForeachCreatedForALaterRow` - `MERGE` as the *only* triggering
      clause (adding a MATCH or CALL to observe would arm the mode by itself): it asks for the node
      the FOREACH creates on the last row, which streaming batches have not created yet, so before the
      fix it created a 251st node instead of matching.
- [x] `mergeInsideALaterForeachAlsoArmsEagerness` - the same MERGE, one level down inside a second
      FOREACH, which a clause-type-only scan reads as a write and skips. Verified against the code
      before fixing it: the query created a 251st node.
- [x] `callSubqueryAfterForeachSeesEveryCreatedNode` - the same through a `CALL { ... }` subquery.
- [x] `countSubqueryExpressionInReturnSeesEveryCreatedNode`,
      `collectSubqueryExpressionNestedInAFunctionCallSeesEveryCreatedNode` and
      `existsSubqueryExpressionInWithWhereSeesEveryCreatedNode` - the read as an *expression* rather
      than a clause. Confirmed against the code before fixing: `RETURN count { MATCH (n:L1) }` and
      `RETURN size(collect { ... })` both returned 200/400/500, and `WITH i WHERE exists { MATCH
      (:L1 {i: 250}) }` returned 50 of the 250 rows.
- [x] `graphReadNestedInsideAContainerExpressionArmsEagerness` - the subquery buried in a `CASE`
      branch and in a list literal, i.e. containers the walk has to descend into.
- [x] `graphReadInALaterForeachListExpressionArmsEagerness` - the read in a later FOREACH's driving
      list: it filters on a node the first FOREACH creates only on its last row, so before the fix
      the body ran for 50 rows instead of 250.
- [x] `eagerReadIsTradedBackForStreamingPerDatabase` - the off switch reaches the planner and stays
      inside its database: with `foreachEagerRead=false` on this database all 500 nodes still land but
      the 250 rows disagree again, while a second database that never set it keeps the eager default.
- [x] `foreachWithoutAFollowingReadStillPassesEveryRowThrough` - negative control: with nothing
      downstream reading the graph the eager mode stays off, all 250 rows still flow through in order
      and all 500 nodes are still created.
- [x] openCypher TCK (`mvn -pl engine verify -Dsurefire.includes='**/*Suite.java'`):
      **3897 scenarios, 0 failures, 85 skipped - identical before and after the change.**
- [x] Full `engine` module (`-DexcludedGroups=benchmark,vector,slow`): **14,024 tests run, 1 failure,
      23 skipped** - the one failure is `Issue6559ApproximateDeltaScoringScaleTest`, a PQ delta-buffer
      timing flake in `com.arcadedb.index.vector` that green in the five preceding full runs of this
      branch and passes 3/3 in isolation. Nothing in this PR touches vector indexing.

## The trade-off eagerness buys, and why not a chunk cap

Eagerness is not free: the FOREACH's input rows are held until the last write is applied, where
before this PR they streamed 100 at a time. A query such as

```cypher
UNWIND $bigBatch AS row
FOREACH (x IN CASE WHEN row.flag THEN [1] ELSE [] END | CREATE (:Audit {row: row}))
MERGE (n:Entity {id: row.id})
```

now materializes `$bigBatch` before the first row reaches `MERGE`. That is inherent to the guarantee -
openCypher's own Eager operator costs exactly the same - so it is a switch, not a cap:
`arcadedb.opencypher.foreachEagerRead=false` restores streaming for an input that does not fit in
memory, at the documented price of the read seeing a partially applied FOREACH.

Capping the eager mode at a chunk size was considered and rejected. A chunk cap does not weaken the
guarantee, it *deletes* it: whatever the chunk size, the reader is back to seeing whichever writes
happened to land inside its chunk, which is the defect this PR fixes - and it would do it silently,
with no setting in the log to explain the numbers.

## Known related gap (not fixed here)

`CreateStep` has the same shape of boundary, just further out: it flushes in
`arcadedb.opencypher.bulkCreateBatchSize` chunks (default 20,000), so
`UNWIND range(1, 50000) AS i CREATE (:L1) CALL meta.stats() ...` would still report three different
counts. It is the same root cause and would take the same treatment, but it needs its own memory
analysis for the bulk-ingest path that batch size exists to serve, so it is left out of this fix
rather than folded into it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `FOREACH` reliability before graph-reading operations, including `MATCH`, `MERGE`, `CALL`, subqueries, nested expressions, and nested `FOREACH` clauses.
  - Ensured updates complete before subsequent reads, providing consistent results across batching and materialization.
  - Preserved streaming behavior when eager processing is unnecessary or disabled.

- **Documentation**
  - Clarified eager `FOREACH` processing, memory considerations, and configuration behavior.

- **Tests**
  - Added regression coverage for downstream reads, batching, nested processing, and complete update visibility.

### DIFF
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -710,6 +710,11 @@ public enum GlobalConfiguration {
       Higher values improve performance but consume more memory. Default: 20000. Recommended range: 10000-100000. Set to 0 to disable batching.""",
       Integer.class, 20_000),
 
+  OPENCYPHER_FOREACH_EAGER_READ("arcadedb.opencypher.foreachEagerRead", SCOPE.DATABASE,
+      """
+      Make FOREACH eager when a later clause in the same query reads the graph (MATCH, MERGE, CALL or a CALL {} subquery).       Eager means the FOREACH body runs for every input row before the first row reaches that reader, so every row sees the       same, complete post-FOREACH graph instead of whichever writes happened to land inside the pull batch it was produced in       (issue #6922). The cost is that the FOREACH's input rows are held in memory until the last write is applied. Set to false       to restore the streaming, batch-at-a-time behaviour for a bulk query whose input does not fit in memory - at the price of       the following read observing a partially applied FOREACH.""",
+      Boolean.class, true),
+
   OPENCYPHER_LOAD_CSV_ALLOW_FILE_URLS("arcadedb.opencypher.loadCsv.allowFileUrls", SCOPE.DATABASE,
       """
       Allow LOAD CSV to access local files via file:/// URLs and bare file paths. \
--- a/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
+++ b/engine/src/main/java/com/arcadedb/GlobalConfiguration.java
@@ -712,7 +712,14 @@ public enum GlobalConfiguration {
 
   OPENCYPHER_FOREACH_EAGER_READ("arcadedb.opencypher.foreachEagerRead", SCOPE.DATABASE,
       """
-      Make FOREACH eager when a later clause in the same query reads the graph (MATCH, MERGE, CALL or a CALL {} subquery).       Eager means the FOREACH body runs for every input row before the first row reaches that reader, so every row sees the       same, complete post-FOREACH graph instead of whichever writes happened to land inside the pull batch it was produced in       (issue #6922). The cost is that the FOREACH's input rows are held in memory until the last write is applied. Set to false       to restore the streaming, batch-at-a-time behaviour for a bulk query whose input does not fit in memory - at the price of       the following read observing a partially applied FOREACH.""",
+      Make FOREACH eager when a later clause in the same query reads the graph (MATCH, MERGE, CALL, a CALL {} subquery, \
+      or a later FOREACH whose own body contains a MERGE). Eager means the FOREACH body runs for every input row before \
+      the first row reaches that reader, so every row sees the same, complete post-FOREACH graph instead of whichever \
+      writes happened to land inside the pull batch it was produced in. The cost is that the FOREACH input rows are held \
+      in memory until the last write is applied. Set to false to restore the streaming, batch-at-a-time behaviour for a \
+      bulk query whose input does not fit in memory, at the price of the following read observing a partially applied \
+      FOREACH. Read when the execution plan is built, so a change takes effect for statements planned afterwards \
+      (see arcadedb.opencypher.planCache).""",
       Boolean.class, true),
 
   OPENCYPHER_LOAD_CSV_ALLOW_FILE_URLS("arcadedb.opencypher.loadCsv.allowFileUrls", SCOPE.DATABASE,
--- a/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
+++ b/engine/src/main/java/com/arcadedb/query/opencypher/executor/CypherExecutionPlan.java
@@ -59,6 +59,12 @@ import com.arcadedb.query.opencypher.ast.MapExpression;
 import com.arcadedb.query.opencypher.ast.MatchClause;
 import com.arcadedb.query.opencypher.ast.MergeClause;
 import com.arcadedb.query.opencypher.ast.OrderByClause;
+import com.arcadedb.query.opencypher.ast.CollectExpression;
+import com.arcadedb.query.opencypher.ast.CountExpression;
+import com.arcadedb.query.opencypher.ast.ExistsExpression;
+import com.arcadedb.query.opencypher.ast.PatternComprehensionExpression;
+import com.arcadedb.query.opencypher.ast.ShortestPathExpression;
+import com.arcadedb.query.opencypher.rewriter.ExpressionRewriter;
 import com.arcadedb.query.opencypher.ast.NodePattern;
 import com.arcadedb.query.opencypher.ast.ParameterExpression;
 import com.arcadedb.query.opencypher.ast.PathPattern;
@@ -4936,12 +4942,17 @@ public class CypherExecutionPlan {
    * <p>
    * The set is deliberately conservative on CALL: a procedure is opaque here, so any procedure call
    * after the update counts as a read - which is what {@code CALL meta.stats()} in issue #6922 is.
-   * SUBQUERY counts because a CALL {{ ... }} block can contain a MATCH. RETURN, WITH and UNWIND do
-   * not: they only project rows the pipeline already produced.
+   * SUBQUERY counts because a CALL { ... } block can contain a MATCH.
    * <p>
    * A later FOREACH counts only when its own body reads, which the grammar narrows to MERGE (and a
    * nested FOREACH containing one): CREATE, SET, REMOVE and DELETE all act on variables the row
    * already carries, and MATCH and CALL are not valid inside a FOREACH body at all.
+   * <p>
+   * RETURN, WITH and UNWIND are not reads in themselves - they project rows the pipeline already
+   * produced - but the expressions they carry can be. {@code EXISTS { }}, {@code COUNT { }} and
+   * {@code COLLECT { }} run a real query against the live graph per row, and a pattern predicate,
+   * pattern comprehension or shortestPath() traverses it, so those clauses are scanned for such an
+   * expression instead of being waved through on their clause type alone.
    *
    * @param clausesInOrder the query's clauses in textual order
    * @param clauseIndex    index of the updating clause, or a negative value when it is not in the list
@@ -4959,6 +4970,26 @@ public class CypherExecutionPlan {
         return true;
       case FOREACH:
         if (foreachBodyReadsGraph(laterClause.getTypedClause()))
+          return true;
+        break;
+      case RETURN:
+        if (returnItemsReadGraph(((ReturnClause) laterClause.getClause()).getReturnItems()))
+          return true;
+        break;
+      case WITH:
+        final WithClause withClause = laterClause.getTypedClause();
+        if (returnItemsReadGraph(withClause.getItems()))
+          return true;
+        if (withClause.getWhereClause() != null
+            && expressionReadsGraph(withClause.getWhereClause().getConditionExpression()))
+          return true;
+        if (withClause.getOrderByClause() != null)
+          for (final OrderByClause.OrderByItem orderByItem : withClause.getOrderByClause().getItems())
+            if (expressionReadsGraph(orderByItem.getExpressionAST()))
+              return true;
+        break;
+      case UNWIND:
+        if (expressionReadsGraph(((UnwindClause) laterClause.getClause()).getListExpression()))
           return true;
         break;
       default:
@@ -4987,6 +5018,76 @@ public class CypherExecutionPlan {
       }
     }
     return false;
+  }
+
+  private boolean returnItemsReadGraph(final List<ReturnClause.ReturnItem> returnItems) {
+    for (final ReturnClause.ReturnItem returnItem : returnItems)
+      if (expressionReadsGraph(returnItem.getExpression()))
+        return true;
+    return false;
+  }
+
+  /**
+   * Tells whether an expression tree holds anything that goes back to the graph while the row is
+   * being projected: a subquery expression ({@code EXISTS { }}, {@code COUNT { }}, {@code COLLECT { }},
+   * which all run a real query per row through {@code CorrelatedSubqueryRunner}) or a pattern that is
+   * traversed ({@code (n)-->()} as a predicate, a pattern comprehension, {@code shortestPath()}).
+   * Used by {@link #graphReadFollows(List, int)} for issue #6922.
+   */
+  private boolean expressionReadsGraph(final Object expression) {
+    if (expression == null)
+      return false;
+    final GraphReadDetector detector = new GraphReadDetector();
+    detector.rewrite(expression);
+    return detector.found;
+  }
+
+  /**
+   * Walks an expression tree looking for a node that reads the graph, reusing {@link ExpressionRewriter}'s
+   * traversal instead of re-deriving one: the check sits in the {@code rewrite} entry point every node
+   * passes through, so it also catches the expression types the rewriter's own dispatch does not know
+   * (COUNT and COLLECT subqueries among them). Nothing is rewritten - every visit returns its input.
+   * <p>
+   * Three of the base class's visits are {@code TODO} stubs that return without recursing, and each one
+   * hides a real case here: {@code size(collect {{ ... }})} behind {@code visitFunctionCall}, and
+   * {@code WHERE exists {{ ... }}} behind the boolean coercion/wrapper pair. They are overridden here
+   * rather than filled in on the shared base class, where every other rewriter would inherit a traversal
+   * change none of them asked for.
+   */
+  private static final class GraphReadDetector extends ExpressionRewriter {
+    private boolean found = false;
+
+    @Override
+    public Object rewrite(final Object expression) {
+      if (found || expression == null)
+        return expression;
+      if (expression instanceof ExistsExpression || expression instanceof CountExpression
+          || expression instanceof CollectExpression || expression instanceof PatternPredicateExpression
+          || expression instanceof PatternComprehensionExpression || expression instanceof ShortestPathExpression) {
+        found = true;
+        return expression;
+      }
+      return super.rewrite(expression);
+    }
+
+    @Override
+    protected Expression visitFunctionCall(final FunctionCallExpression expr) {
+      for (final Expression argument : expr.getArguments())
+        rewrite(argument);
+      return expr;
+    }
+
+    @Override
+    protected BooleanExpression visitBooleanCoercion(final BooleanCoercionExpression expr) {
+      rewrite(expr.getExpression());
+      return expr;
+    }
+
+    @Override
+    protected Expression visitBooleanWrapper(final BooleanWrapperExpression expr) {
+      rewrite(expr.getBooleanExpression());
+      return expr;
+    }
   }
 
   private boolean isFollowedByCountOnlyReturn(final List<ClauseEntry> clausesInOrder, final int callIndex) {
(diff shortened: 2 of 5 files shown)


robfrank:
Thanks - all three points addressed in 27e960c.

**Memory / perf.** Agreed the eager mode is unbounded, and the `MERGE` bulk-upsert shape is a real one. I went with a switch rather than a cap: `arcadedb.opencypher.foreachEagerRead` (SCOPE.DATABASE, default `true`) restores the streaming behaviour for an input that does not fit in memory, and the trade-off is now written up in the PR description alongside the `CreateStep` gap.

I deliberately did **not** take the chunk-cap option, because it does not weaken the guarantee - it deletes it. Whatever chunk size you pick, the reader is back to seeing whichever writes happened to land inside its chunk, which is exactly the defect this PR fixes, and it would do it silently with nothing in the configuration to explain the numbers. The cost of eagerness is also not something the fix invented: openCypher's own Eager operator holds the same rows for the same reason. A named switch at least makes the choice explicit and greppable.

**`indexOf` nit.** Fixed at both sites - `entryIndex` is already in scope from the enclosing indexed loop, so `graphReadFollows(clausesInOrder, entryIndex)` and `isFollowedByCountOnlyReturn(clausesInOrder, entryIndex)` both drop the O(n) rescan.

**Missing trigger coverage.** Added `mergeAfterForeachMatchesANodeTheForeachCreatedForALaterRow` and `callSubqueryAfterForeachSeesEveryCreatedNode`, plus `eagerReadCanBeTradedBackForStreaming` for the new switch. The MERGE one needed some care to stay honest: any MATCH or CALL added to observe the graph would arm the eager mode by itself, so MERGE has to be both the trigger and the observation. It asks for `(:L1 {i: 250})`, the node the FOREACH only creates on the very last row - a streaming pipeline emits the first row 150 rows before that exists and creates a 251st node instead of matching.

Re-verified after the changes: 5 of the 7 tests fail with the eager mode off and pass with it on; engine module 14,018 tests / 0 failures; openCypher TCK 3897 scenarios / 0 failures / 85 skipped, identical to the pre-change baseline.

robfrank:
Both confirmed and fixed in f6fa795.

**1. Nested MERGE gap - real, and it reproduces.** Ran your exact query before touching anything:

```cypher
UNWIND range(1, 250) AS i
FOREACH (j IN [1] | CREATE (:L1 {i: i}))
FOREACH (j IN [1] | MERGE (last:L1 {i: 250}))
RETURN i
```

250 rows, but 251 `:L1` nodes - the nested `MERGE` ran while `i = 250` was still 150 rows away and created it instead of matching, exactly as you predicted. `graphReadFollows()` now recurses into a later FOREACH's `getInnerClauses()`, and your read of the grammar is right: `MERGE` (plus a nested FOREACH holding one) is the only thing in a FOREACH body that can read, since CREATE/SET/REMOVE/DELETE all act on variables the row already carries. New test `mergeInsideALaterForeachAlsoArmsEagerness` pins it.

**2. Config description formatting.** Good catch, and worse than it looked: the continuation backslashes were lost entirely, so the text block had collapsed into a single 696-column line with the stray space runs you spotted. Reflowed to the `\`-continued ~120-column style of its neighbours.

**Also from CodeRabbit on the same push:** the setting is `SCOPE.DATABASE` but I was reading it off the process-wide enum, so a per-database override would have been ignored. It now goes through `database.getConfiguration().getValueAsBoolean(...)`, matching `LabelReplacements` and `LoadCSVStep` in the same package, and the escape-hatch test sets it per database and asserts a second database keeps the eager default.

Re-verified after all three changes: 6 of the 8 tests fail with the eager mode off and pass with it on; engine module 14,019 tests / 0 failures / 23 skipped; openCypher TCK 3897 scenarios / 0 failures / 85 skipped, unchanged from the baseline.

robfrank:
Confirmed and fixed in 1dd90c5 - and it was worse than one case. I ran all three shapes before touching anything:

| query | before |
|---|---|
| `RETURN count { MATCH (n:L1) }` | `{200 x100, 400 x100, 500 x50}` |
| `RETURN size(collect { MATCH (n:L1) RETURN n.i })` | `{200 x100, 400 x100, 500 x50}` |
| `WITH i WHERE exists { MATCH (n:L1 {i: 250}) }` | 50 of the 250 rows survived |

All three now return the post-FOREACH answer, and all three are pinned by new tests.

`graphReadFollows` now scans a later `RETURN`/`WITH`/`UNWIND` (return items, `WHERE`, `ORDER BY`, the UNWIND source) for `ExistsExpression`, `CountExpression`, `CollectExpression`, `PatternPredicateExpression`, `PatternComprehensionExpression` and `ShortestPathExpression`. Rather than write a second expression walker I reused `ExpressionRewriter`, hooking the check in the `rewrite()` entry point every node passes through - which also covers the types its own dispatch doesn't know, `CountExpression` and `CollectExpression` among them.

One thing worth flagging from doing it: three of `ExpressionRewriter`'s visits are `TODO` stubs that return without recursing (`visitFunctionCall`, `visitBooleanCoercion`, `visitBooleanWrapper`), and **each one hid exactly one of the cases above** - `size(collect { ... })` behind the function call, `WHERE exists { ... }` behind the coercion/wrapper pair. That is why the first pass fixed only the bare `count { }`. I overrode them in the detector instead of filling them in on the shared base class: making the shared traversal deeper would change behaviour for every other rewriter subclass, which is not something this PR should decide.

Both minor points are in too: the `{{ ... }}` doubled brace is gone, and `createStepForClause`'s nested-FOREACH branch now says why `eagerExecution` is deliberately off there (`executeInnerClauses` drains the inner chain to exhaustion per outer iteration, so the `nRecords` cap cannot split a nested body across two pulls).

Re-verified: 10 of the 11 tests fail with the eager mode off and pass with it on, the eleventh being the negative control; engine module 14,022 tests / 0 failures / 23 skipped; openCypher TCK 3897 scenarios / 0 failures / 85 skipped, unchanged from baseline.

robfrank:
Both gaps confirmed and closed in c752c11. You were right that the three overrides were the whole reach of the walk - here is what each container actually did before the fix, measured, not reasoned:

| wrapper around `count { MATCH (n:L1) }` | before |
|---|---|
| `CASE WHEN true THEN ... ELSE 0 END` | `{200 x100, 400 x100, 500 x50}` |
| `[ ... ]` list literal | same |
| `{c: ... }` map literal | same |
| `0 + ...` | same |

`GraphReadDetector` now overrides every non-recursing visit: CASE, list, map, map projection, list comprehension, ALL/ANY/NONE/SINGLE, reduce, list index and slice, arithmetic, IN, IS NULL, string match, regex and label check, on top of the three it already had.

I kept them as detector overrides rather than filling in `ExpressionRewriter`'s stubs, even though the accessors do all exist. Those stubs are not merely unfinished: making the shared traversal descend changes what every rewriter subclass rewrites, and a rewriter that starts rebuilding nodes it previously returned untouched is a behaviour change well outside this fix's blast radius. It is worth doing - just as its own PR, with its own justification for each subclass.

**The FOREACH driving list.** Also real. `foreachBodyReadsGraph()` now runs the list expression through the same detector, so:

```cypher
UNWIND range(1, 250) AS i
FOREACH (j IN [1] | CREATE (:L1 {i: i}))
FOREACH (n IN [x IN range(1, 1) WHERE exists { MATCH (m:L1 {i: 250}) } | x] | CREATE (:Marker {i: i}))
```

creates 250 markers instead of 50. New test `graphReadInALaterForeachListExpressionArmsEagerness`.

**Import nit:** the whole `com.arcadedb.query.opencypher.*` block is sorted now, so the strays my earlier commits appended are gone along with the pre-existing ones.

Verification: 12 of the 13 tests fail with the eager mode off and pass with it on, the thirteenth being the negative control. openCypher TCK 3897 scenarios / 0 failures / 85 skipped, unchanged from baseline. The engine module run came back 14,024 / 1 failure - `Issue6559ApproximateDeltaScoringScaleTest`, a PQ delta-buffer timing assertion in `com.arcadedb.index.vector`. It was green in the five preceding full runs on this branch and passes 3/3 in isolation, and nothing here goes near vector indexing, so I am calling it an unrelated flake rather than chasing it in this PR.

This was cycle 4 of 4, so the loop stops here and the branch is handed back for merge.

robfrank:
Findings 1, 2 and 3 confirmed and fixed. Finding 4 is a real gap in the detector but is not reachable as a wrong result; I left the code alone and documented why. The javadoc count is corrected.

I verified each finding by running the query rather than by reading the code, on the branch head (c752c11), 250 rows spanning three 100-row pull batches:

| Finding | Query result before | Expected |
|---|---|---|
| 1 (`NOT exists{}`) | `[true, false]` | all `false` |
| 1 (`true AND exists{}`) | `[false, true]` | all `true` |
| 2 (`(CASE…END).v` holding `count{}`) | `[200, 400, 500]` | all `500` |
| 3 (`CREATE (:S {n: count{}})` in a later FOREACH) | `:Snapshot.n` = `[100, 200, 250]` | all `250` |

**1 and 2** — fixed by failing closed, as suggested. `ExpressionRewriter.rewriteDispatch()`'s `// Unknown expression type: return as-is` fallback now calls a new `protected Object visitUnknown(Object)` whose default still returns the node untouched, so no existing rewriter changes behaviour. `GraphReadDetector` overrides it to set `found = true`. That covers `ChainedPropertyAccessExpression` without needing to reference it — it is package-private in `…parser`, so the detector in `…executor` could not dispatch on its type even if we wanted to — and covers any `Expression` subtype added later.

One deliberate departure from a pure fail-closed default: `TernaryLogicalExpression` gets an explicit arm in `GraphReadDetector.rewrite()` that descends into both operands. Leaving it to `visitUnknown` would be correct but would arm eager mode behind *every* `RETURN a AND b` following a FOREACH, whether or not anything there reads the graph. That is the one type common enough for the imprecision to cost real queries.

**3** — `foreachBodyReadsGraph` now scans a later FOREACH's `CREATE` property values (node and relationship inline maps, plus dynamic labels), `SET` value/key/target expressions, `REMOVE` key expressions and `DELETE` expressions. The javadoc's "the grammar narrows to MERGE" reasoning was wrong for exactly the reason given: the clauses act on variables the row carries, but the values they compute need not.

**4** — the structural claim is correct: `ReturnClause` carries no ORDER BY, `ClauseDispatcher:166` puts a terminal one on the statement via `setOrderBy`, and `graphReadFollows`'s `case RETURN:` never sees it. But it cannot produce a wrong result. `OrderByStep.materializeAndSort()` drains its entire input with `prev.syncPull(context, Integer.MAX_VALUE)` into a list and only then calls `createComparator()`, so the FOREACH has always completed before an ORDER BY expression is evaluated. I also checked the Top-K path, where the comparator genuinely does run inside the pull loop: `RETURN i AS v ORDER BY count { MATCH (n:L1) } DESC, i LIMIT 5` over 2500 rows still returns `[1, 2, 3, 4, 5]`. Scanning it would add a branch that no query can reach, so instead the javadoc now records that the omission is deliberate and why — an ORDER BY that is only ever evaluated after a blocking materialize cannot observe a batch boundary.

**Minor** — the count was wrong; it now reads "Eighteen", matching your list, and notes that `visitUnknown` closes what remains rather than implying the overrides are exhaustive on their own.

Five regression tests added to `CypherForeachEagerReadIssue6922Test`, all five verified to FAIL against the branch head and pass with the fix (measured by restoring the two source files from `HEAD` and re-running, not by inspection). The first draft of the SET test passed without the fix — its earlier FOREACH created `:L2` while the `count{}` read `:L1`, so the read could not observe the write — and was rewritten to have both touch the same label before I counted it.

Verification: opencypher package 9130 tests / 0 failures; openCypher TCK 3897 scenarios / 0 failures / 85 skipped, identical to the pre-change baseline; full engine module green.

